PR #950, opened 2013-11-03
Log socket accept errors when forking tests

jroper:
If an exception is thrown when accepting a connection from a forked test agent, currently I'm seeing that all that happens is SBT hangs with no output.  Thread dumps show that the main process is waiting for the agent to return, while the agent is waiting for the server to send it something.

This change logs the exception, so that at least the error can be googled. It also cleans up the server socket.


harrah:
Thanks as always @jroper.  Merged in 191737d35b6433192bd167970f5e98188ebb0348.

PR #954, opened 2013-11-05
fixed incorrect `fork in test`

backuitist:
Should be `fork in Test` (configuration not task). Added tips for figuring out this problem.


harrah:
This sounds like a bug if you are saying `fork in test` doesn't work- it used to and it should.  The reason is that you might not want `test:run` to fork, for example.

backuitist:
Throughout the doc you can read both syntax: `fork in test` and `fork in Test`. At the moment only the latest works.

harrah:
That's a bug then.  Please open an issue.

PR #955, opened 2013-11-05
added '.idea' and '.idea_modules' to gitignore

backuitist:


harrah:
These should go in your global .gitignore, not the project's.

PR #956, opened 2013-11-06
Fix #944 Add option preserveExecutable to IO.copyFile and IO.copyDirectory

### LINKED ISSUE #944: Add option preserveExecutable to IO.copyFile and IO.copyDirectory
As I understand correctly, they have option `preserveLastModified`, but no `preserveExecutable`.

I can modify and send a pull request if you want. Please let me know.

Use case:

I am creating this SBT plugin:
https://github.com/ngocdaothanh/xitrum-package

For example, it allows users to copy `bin/start.sh` to `target/xitrum/bin/start.sh`. But because there's no `preserveExecutable` option, the copied `start.sh` will not be execute.

ngocdaothanh:
preserveExecutable is true by default.

### DIFF
--- a/util/io/src/main/scala/sbt/IO.scala
+++ b/util/io/src/main/scala/sbt/IO.scala
@@ -542,12 +542,13 @@ object IO
 	* If the source is a directory, the corresponding directory is created.
 	*
 	* If `preserveLastModified` is `true`, the last modified times are transferred as well.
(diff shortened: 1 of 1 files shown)


harrah:
Merged to the 0.14 branch (because of binary incompatibility) in 1e61b9acf34b7d89a54a5adc021cc37cd3ecd0ef.  Thanks for preserving source compatibility!

PR #960, opened 2013-11-06
Add in the ability to launch non xsbti.AppMain classes.

jsuereth:
- Deprecate old mainClass method on appProvider
- Create entryPoint method which represnets class used to launch instead.
- Create PlainApplication to wrap static `main` methods.  Can return
  either Int, Exit or Unit.
- Detect supported 'plain' classes via reflection.
- Add new unit tests appropriate to the feature.

Relatively self-contained.  Probably needs more supported 'raw' interfaces (like passing classpath), but clean enough we can just include in future sbt.

Review by @harrah, cc> @adriaanm @havocp

### DIFF
--- a/launch/interface/src/main/java/xsbti/AppProvider.java
+++ b/launch/interface/src/main/java/xsbti/AppProvider.java
@@ -8,13 +8,27 @@ public interface AppProvider
 	public ScalaProvider scalaProvider();
 	/** The ID of the application that will be created by 'newMain' or 'mainClass'.*/
 	public ApplicationID id();
-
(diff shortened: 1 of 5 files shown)


jsuereth:
Ok, this should be ready for merge (IMHO)

harrah:
squash and I'll merge, thanks.

jsuereth:
squashed.

harrah:
Merged in b0f6f9483918f7e7aeca321e0ff44f745fd5e167, thanks!

PR #961, opened 2013-11-07
Allow specifying sbt.boot.properties as a URI

jroper:
This allows straight forward system independent generation of the property, since URIs are well specified and only support one file separator.

Play was relying on this behaviour in SBT 0.12, which worked by accident.  This makes it explicitly supported, tested and documented.


harrah:
Merged in 7d27fc226a9d059996ba36409831bffaff7c8d9b, thanks!

PR #967, opened 2013-11-08
Added a test reporter for JUnit XML

backuitist:
Taken from https://github.com/chenkelmann/junit_xml_listener and slightly improved.
Activating is now just a matter of setting `testReportJUnitXml` to true.
This not only brings a very handy feature (hudson/jenkins support that syntax),
but gives more options for testing the SBT test infrastructure.
For instance checking that the test output is well segregated during a parallel
test run is just a matter of checking the report output.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -349,6 +349,7 @@ object Defaults extends BuildCommon
 			Seq(ScalaCheck, Specs2, Specs, ScalaTest, JUnit)
 		},
 		testListeners :== Nil,
+		testReportJUnitXml :== false,
 		testOptions :== Nil,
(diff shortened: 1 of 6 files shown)


jsuereth:
Nice.  One question though:

Plugins have more flexibility to change/release than sbt itself. Do you think it makes sense just to include that plugin in sbt's own build so we can improve our testing or do you see it as something fundamental sbt should provide?

backuitist:
I think it is fundamental enough to provide it. About the change/release thing, I don't think the JUnit XML syntax is gonna change very much but on the other hand the `TestsListener` trait (and everything that comes with it) might. It is also quite lightweight.
Improving the testing comes as a bonus :)

backuitist:
Closing as this is superseded by https://github.com/sbt/sbt/pull/993

PR #969, opened 2013-11-10
Resolved bugs reported in Issue 947.

awhitford:
name and version had some duplication in the source that didn't seem to appear anywhere -- looked safe to remove.
For version, I think the initial advice should be to use a SNAPSHOT version (as a non-SNAPSHOT version should be limited to "cutting a release.")
For organization, it looks like the initial advice was being hidden because of a missing double colon, and then the later advice for organizationName was incorrectly stating organization in its example.
The organizationHomepage and license urls needed to be url values -- not strings.

### DIFF
--- a/src/sphinx/Howto/metadata.rst
+++ b/src/sphinx/Howto/metadata.rst
@@ -9,8 +9,6 @@ A project should define :key:`name` and :key:`version`.  These will be used in v
    :type: setting
    :title: Set the project name
 
-   name := "demo"
-
 ::
 
     name := "Your project name"
@@ -22,16 +20,16 @@ For published projects, this name is normalized to be suitable for use as an art
    :type: setting
    :title: Set the project version
 
-   version := "1.0"
-
 ::
 
-   version := "1.0"
+   version := "1.0-SNAPSHOT"
 
 .. howto::
    :id: organization
    :type: setting
    :title: Set the project organization
+
+::
 
    organization := "org.example"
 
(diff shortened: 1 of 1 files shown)


harrah:
The "duplication" is used by the index with examples: http://www.scala-sbt.org/release/docs/Howto/Howto.html#project-metadata.  @jsuereth might have an opinion about using `-SNAPSHOT`.  Looks good otherwise.

jsuereth:
yeah, given that we use ivy AND all the issues around -SNAPSHOT in generally, it's best to avoid that convention.  I have a talk about it in effective sbt, and I know a lot of those of us who work with maven/ivy dependency management agree on this.  I can go into more detail, but I highly recommend not using -SNAPSHOT (see the sbt-git plugin's option here for sha-versioned releases).  I'd prefer not to encourage this in our HowTo, because people will copy-paste those example and then wonder why resolutoins continues to get slower every time they add dependencies, and why changes aren't showing up when they compile.... (or too many changes).

Also -SNAPSHOT completely ruins tools like `git bisect`, which are bread and butter for support/bug analysis tasks...

jsuereth:
Oh, other than -SNAPSHOT bits, this looks great!!!! Thanks for taking time to improve our docs.  If you're using -SNAPSHOT with sbt (and not because you integrate with existing maven builds), then I'd encourage you to look at the alternatives to this.

harrah:
@awhitford Can you update the pull request according to the feedback?

awhitford:
Hope this helps.

harrah:
Merged in e7d5cd1bd25aa24abb6a1ead3a2701bc451ff724, thanks!

PR #972, opened 2013-11-11
Fix type error in example code

### LINKED ISSUE #971: The "Generating files" HOWTO is broken for 0.13.0
According to http://www.scala-sbt.org/0.13.0/docs/Howto/generatefiles.html:

> ``` scala
> sourceGenerators in Compile += <your Task[Seq[File]] here>
> ```

But this simply doesn't work. It's obvious why when comparing with [the 0.12.4 version](http://www.scala-sbt.org/0.12.4/docs/Howto/generatefiles.html), which gives:

> ``` scala
> sourceGenerators in Compile <+= <your Task[Seq[File]] here>
> ```

Now, that's just _not_ the kind of change which happened between 0.12 and 0.13. _Edit_: I spoke a bit too soon; you can replace <+= with += if the type on the right changes, and loses the `Initialize` part, but it's not the case here, probably because it's a setting containing a task.

This is visible in the example. Example given for 0.12.4:

> ``` scala
> sourceGenerators in Compile <+= sourceManaged in Compile map { dir =>
>   val file = dir / "demo" / "Test.scala"
>   IO.write(file, """object Test extends App { println("Hi") }""")
>   Seq(file)
> }
> ```

Example given for 0.13.0, but non-working:

> ``` scala
> sourceGenerators in Compile += Def.task {
>   val file = (sourceManaged in Compile).value / "demo" / "Test.scala"
>   IO.write(file, """object Test extends App { println("Hi") }""")
>   Seq(file)
> }
> ```

After replacing `+=` with `<+=`, we get a example that works:

``` scala
sourceGenerators in Compile <+= Def.task {
  val file = (sourceManaged in Compile).value / "demo" / "Test.scala"
  IO.write(file, """object Test extends App { println("Hi") }""")
  Seq(file)
}
```

Error from the given 0.13.0 code (not very self-explanatory, I'd say):

```
> reload
[info] Loading project definition from /Users/pgiarrusso/Documents/Research/PhD/Teaching/SLE2013/public-material-repo/lectures/sbt-sample/project
/Users/pgiarrusso/Documents/Research/PhD/Teaching/SLE2013/public-material-repo/lectures/sbt-sample/build.sbt:8: error: No implicit for Append.Value[Seq[sbt.Task[Seq[java.io.File]]], sbt.std.FullInstance.M[Seq[java.io.File]]] found,
  so sbt.std.FullInstance.M[Seq[java.io.File]] cannot be appended to Seq[sbt.Task[Seq[java.io.File]]]
sourceGenerators in Compile += Def.task {
                            ^
```

I'd open a further bug saying "example code in documentation is not tested", but I don't know if you'd care and/or if the technology for fixing it exists.

Blaisorblade:
Fix #971.

Note that this is no complete fix, because this solution uses an operator which is not any more described in the guide for 0.13.

Also note that I didn't do any further testing compared to the example I've shown in #971.

### DIFF
--- a/src/sphinx/Howto/generatefiles.rst
+++ b/src/sphinx/Howto/generatefiles.rst
@@ -9,19 +9,19 @@ sbt provides standard hooks for adding source or resource generation tasks.
    :title: Generate sources
    :type: setting
 
-   sourceGenerators in Compile += <your Task[Seq[File]] here>
+   sourceGenerators in Compile <+= <your Task[Seq[File]] here>
 
 A source generation task should generate sources in a subdirectory of :key:`sourceManaged` and return a sequence of files generated.  The key to add the task to is called :key:`sourceGenerators`.  It should be scoped according to whether the generated files are main (`Compile`) or test (`Test`) sources.  This basic structure looks like:
 
 ::
 
-    sourceGenerators in Compile += <your Task[Seq[File]] here>
+    sourceGenerators in Compile <+= <your Task[Seq[File]] here>
 
 For example, assuming a method `def makeSomeSources(base: File): Seq[File]`,
 
 ::
 
-    sourceGenerators in Compile +=
+    sourceGenerators in Compile <+=
       Def.task { makeSomeSources( (sourceManaged in Compile).value / "demo" ) }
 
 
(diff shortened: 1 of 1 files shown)


harrah:
Merged in 09fc6c1fd6279ce0251749262cfe0edb4a94ca21, thanks!

PR #976, opened 2013-11-12
Make recompilation on macro definition optional

gkossakowski:
Introduce a new incremental compiler option that controls
incremental compiler's treatment of macro definitions and their clients.
The current strategy is that whenever a source file containing a macro
definition is touched it will cause recompilation of all direct
dependencies of that file.

That strategy has proven to be too conservative for some projects like
Scala compiler of specs2 leading to too many source files being recompiled.
We make this behavior optional by introducing a new option
`recompileOnMacroDef` in `IncOptions` class. The default value is set to
`true` which preserves the previous behavior.

---

The f79292e and 513f56c are needed due to problem with case classes and binary compatibility. See [this](https://groups.google.com/d/topic/sbt-dev/3VJEnuO9ktM/discussion) discussion for details.

### DIFF
--- a/compile/inc/src/main/scala/sbt/inc/IncOptions.scala
+++ b/compile/inc/src/main/scala/sbt/inc/IncOptions.scala
@@ -49,8 +49,43 @@ final class IncOptions(
 	val newClassfileManager: () => ClassfileManager
 ) extends Product with Serializable {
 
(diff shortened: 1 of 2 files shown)


harrah:
Once all of the changes are made, this LGTM so just merge it when you are done.

gkossakowski:
@harrah: I incorporated all of your feedback. Thanks a lot!

PR #980, opened 2013-11-13
completions command

hvesalai:
The completions command is meant for dump terminals that cannot use
the default tab completion. It has been built for use by the emacs
sbt-mode (see https://github.com/hvesalai/sbt-mode), but is equally
useful for other code editors that can integrate with sbt.

### DIFF
--- a/main/command/src/main/scala/sbt/BasicCommandStrings.scala
+++ b/main/command/src/main/scala/sbt/BasicCommandStrings.scala
@@ -12,6 +12,7 @@ import Path._
 object BasicCommandStrings
 {
 	val HelpCommand = "help"
+	val CompletionsCommand = "completions"
 	val Exit = "exit"
 	val Quit = "quit"
 
@@ -31,6 +32,9 @@ object BasicCommandStrings
 
 	Searches the help according to the provided regular expression.
 """
+
+  	def CompletionsDetailed = "Displays a list of completions for the given argument string (run 'completions <string>')."
+  	def CompletionsBrief = (CompletionsCommand, CompletionsDetailed)
 
 	def HistoryHelpBrief = (HistoryCommands.Start -> "History command help.  Lists and describes all history commands.")
 	def historyHelp = Help(Nil, (HistoryHelpBrief +: HistoryCommands.descriptions).toMap, Set(HistoryCommands.Start))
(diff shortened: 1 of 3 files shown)


jsuereth:
Neat.  This is the kind of thing we want in the Server API at some point I think...

harrah:
@jsuereth see the above discussion.  This provides a limited view of the information that would not be sufficient for the Server API.  The Server API requires both the display and append members.  Perhaps that work can be mostly deferred if this pull request provides support for different output formats via flags?

hvesalai:
What's this talk about a Server API? Can I read more about it somewhere?

hvesalai:
@harrah is there something that you want me to change in the pull request or is it good as is?

harrah:
Mailing list post: https://groups.google.com/forum/#!topic/sbt-dev/1TvYrLF4ExU
Wiki page: https://github.com/sbt/sbt/wiki/Client-server-split

As for the pull request, I'm checking with Josh to see if he thinks we should at least ensure the command supports different output formats, perhaps by switches.  Since this implementation does everything you need, I'd like to merge it but not be tied to the output format if it needs to be changed.

jsuereth:
Hey man.  Sorry I was on travel and just unable to cope with the incoming email.  Still fighting it down this week.

In any case, here's what I think:
1. The mechanism you're using here to get completions looks fine for now.  It's the fastest way to good completions.
2. I may actually duplicate that code in the "shim" we use for the sbt remote-control protocol.   This would throw a more formal "typed" API on asking for, and receiving autocompletion potentials.
3. For code re-use, i'd like to see `def runCompletions(state: State)(input: Option[String]): State` become `def runCompletions(state: State)(input: Option[String]): (State, Seq[String])`.  That is, it returns the completions.  Then I can call your method directly inside my "shim" to grab available completions and return them in an alternative format.

How's that sound?  I can even do #3 myself as needed before the next cut of sbt, unless this is planned to go in 0.13.2, in which case we should make that change now.

Otherwise LGTM!!!!!  Thanks much for the contribution.  Looking forward to the emacs tooling!

harrah:
We won't worry about reusing this command for the build server then.  The runCompletions implementation is fairly specific to the output and there isn't much to it, so if you will do a typed API, you can just do your own thing there.

@hvesalai looks like it is just the trailing space problem and then this can be merged.

hvesalai:
@harrah The above commit adds support for quoted strings. From my side this is now ready to be merged.

harrah:
Merged in 9989e5631ff83853bd7b84b0539913973f7450d6, thanks!

PR #981, opened 2013-11-13
added test-output segregation in forked-tests

backuitist:
I've copied jumi `fi.jumi.core.stdout` package into the test agent.
It needs cglib.
While doing this I simplified the test report API. A TestEvent is now a
SuiteReport and it is made of TestReport. A TestReport contains the test
stdout capture. I've successfully tested it on a test project,
but I'm going to add a scripted test for it.

### DIFF
--- a/testing/agent/src/main/java/sbt/NullOutputStream.java
+++ b/testing/agent/src/main/java/sbt/NullOutputStream.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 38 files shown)


testing/agent/src/main/java/sbt/OutputStringBuilder.java
```diff
@@ -0,0 +1,23 @@
+package sbt;
+
+public class OutputStringBuilder implements OutputListener {
+	private StringBuilder outBuilder = new StringBuilder();
```
luontola: OutputListener may be called from multiple threads (if a test spaws more threads), but StringBuilder is not thread-safe. In this case, using the old StringBuffer is preferred. Additionally, for full thread-safety, getOutAndReset() should synchronize on the StringBuffer instance.

backuitist: Thanks! I've fixed it.

testing/agent/src/main/java/sbt/NullOutputStream.java
```diff
@@ -0,0 +1,67 @@
+/*
```
jsuereth, line 1: We can't just pull in code from Apache license without also updating the licensing and documentation of sbt.  Yes, I believe BSD + Apache are compatible, but you still have to show the license in the code.

I'm guess it's not practical to re-implement the idea to avoid the licensing pains?

harrah, line 1: jsuereth missed the mailing list discussion where this was marked as a proof of concept and won't be merged- no need to act on this.

backuitist, line 1: hehe, i was just about to say that, and also point you to this: https://groups.google.com/forum/#!topic/sbt-dev/E7gPr8iu9b4

jsuereth, line 1: Thanks guys.  I'm still ~ 1 week behind.

backuitist:
Closing as this is superseded by #993

PR #982, opened 2013-11-13
execute scripted test without publishing SBT first

backuitist:
Saves you some time when only your test has changed.

### DIFF
--- a/project/Sbt.scala
+++ b/project/Sbt.scala
@@ -151,9 +151,7 @@ object Sbt extends Build
 		normalizedName := "sbt"
 	)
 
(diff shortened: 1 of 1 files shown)


harrah:
Cool, how much time is saved?  I know publishLocal takes about 2 s.  Is there additional savings when the test itself is run, perhaps due to less work done by sbt startup since the last modified times don't change?

harrah:
Merged in e4dd71c08c794d3e2075a1fd3f554313797e5937, thanks!

backuitist:
If I run `publishAll` twice, the second time, on my machine (which I believe is a good reference), it takes 4 seconds. Perhaps as you said, it might even speed up sbt startup time thanks to unchanged last-modified?
It's not a lifesaver but considering how slow are scripted tests, it's a bit of a relief :)

PR #983, opened 2013-11-13
i found the colons that delimited the item with the explanation to be ve...

drdamour:
...ry confusing since the definitions included colons.  some lines omitted the colon so i went with that


drdamour:
Also, the wiki markup has :key: in it...but this doesn't appear in the online docs, is this a bug?

harrah:
The `:key:` is a custom role that generates links to the highlighted source.   Your patch looks good, but there appears to be a spurious word in there.

drdamour:
ok let me resubmit

Issue #979, opened 2013-11-13
OutOfMemoryError when mixing usage of 0.13.0 and 0.13.1-RC2


paulp:
As described here: https://groups.google.com/forum/#!topic/sbt-dev/NPb0kZ2rpsY

Ideally you can reproduce it with
`git clone https://github.com/paulp/sbtbug && cd sbtbug && ./runner.sh`
However there is something contributing non-determinism, and sometimes it works fine.

Okay, I found the likely source of non-determinism. "sbt run" is behaving very differently in 0.13.1-RC2. It doesn't wait for main to finish. If I use this main body:

``` scala
  def main(args: Array[String]): Unit = {
    println("a")
    Thread.sleep(10000);
    println("b")
  }
```

Under 0.13.0 it prints a, sleeps 10 seconds, prints b. Under 0.13.1-RC2 it prints a and then exits.

0.13.0:

```
[info] Running bippy.Main
a
b
[success] Total time: 12 s, completed Nov 12, 2013 7:29:42 PM
```

0.13.1-RC2:

```
[info] Running bippy.Main
a
[success] Total time: 1 s, completed Nov 12, 2013 7:29:58 PM
```

harrah:
I appreciate you tracking down that non-determinism, since that's a regression that will also need to get fixed.

paulp:
It becomes slipperier when I try to isolate that. But I do have an attachment which should make it easy to at least see the memory exhaustion.

paulp:
Or maybe not, it seems I can't attach a tgz. But I'll keep it in case reproducing it is hard.

harrah:
Ok, thanks.  I have a guess that I'll try tomorrow.

harrah:
I can't reproduce either issue.  I ran that main several times using 0.13.1-RC2 and it always waited for `run` to finish.  I don't know if it matters, but I use a plain startup path and not sbt-extras.

harrah:
Ah, I can reproduce the awt and the not waiting for `run` issues on a mac, though.

SethTisue:
I just recently started getting mysterious OutOfMemoryErrors when starting sbt, and I also recently started testing 0.13.1-RC2, so perhaps I've been experiencing this too. if there's anything I should do next time it happens, let me know. (On the group @harrah mentioned something about a cache? What cache do I try clearing to see if the error goes away?)

harrah:
[...]

PR #985, opened 2013-11-14
Replace binary Analysis format with a text-based one.

benjyw:
Reads/writes are a little faster with the text format,
and it's far more useful. E.g., it allows external manipulation
and inspection of the analysis.

We don't gzip the output. It does greatly shrink the files,
however it makes reads and writes 1.5x-2x slower, and we're
optimizing for speed over compactness.


gkossakowski:
Hi Benjy!

Thanks for submitting the patch! I'm reviewing it right now.

W dniu czwartek, 14 listopada 2013 użytkownik Benjy Weinberger napisał:

> Reads/writes are a little faster with the text format,
> and it's far more useful. E.g., it allows external manipulation
> and inspection of the analysis.
> 
> We don't gzip the output. It does greatly shrink the files,
> however it makes reads and writes 1.5x-2x slower, and we're
> 
> ## optimizing for speed over compactness.
> 
> You can merge this Pull Request by running
> 
>   git pull https://github.com/benjyw/sbt text_analysis_file
> 
> Or view, comment on, or merge it at:
> 
>   https://github.com/sbt/sbt/pull/985
> Commit Summary
> - Replace binary Analysis format with a text-based one.
> 
> File Changes
> - _M_
>   compile/integration/src/main/scala/sbt/compiler/AggressiveCompile.scalahttps://github.com/sbt/sbt/pull/985/files#diff-0(2)
> - _M_
>   compile/integration/src/main/scala/sbt/compiler/IncrementalCompiler.scalahttps://github.com/sbt/sbt/pull/985/files#diff-1(11)
> - _M_ compile/persist/src/main/scala/sbt/inc/AnalysisFormats.scalahttps://github.com/sbt/sbt/pull/985/files#diff-2(2)
> - _M_ compile/persist/src/main/scala/sbt/inc/FileBasedStore.scalahttps://github.com/sbt/sbt/pull/985/files#diff-3(16)
> - _D_ compile/persist/src/main/scala/sbt/inc/InternPool.scalahttps://github.com/sbt/sbt/pull/985/files#diff-4(82)
> - _D_
>   compile/persist/src/main/scala/sbt/inc/InternedAnalysisFormats.scalahttps://github.com/sbt/sbt/pull/985/files#diff-5(98)
> - _A_ compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scalahttps://github.com/sbt/sbt/pull/985/files#diff-6(312)
> 
> Patch Links:
> - https://github.com/sbt/sbt/pull/985.patch
> - https://github.com/sbt/sbt/pull/985.diff

## 

Grzegorz Kossakowski
Scalac hacker at Typesafe http://www.typesafe.com/
twitter: @gkossakowski http://twitter.com/gkossakowski
github: @gkossakowski http://github.com/gkossakowski

compile/integration/src/main/scala/sbt/compiler/IncrementalCompiler.scala
```diff
@@ -40,10 +40,5 @@ object IC extends IncrementalCompiler[Analysis, AnalyzingCompiler]
 		try { readCacheUncaught(file)._1 } catch { case _: Exception => Analysis.Empty }
 
 	def readCacheUncaught(file: File): (Analysis, CompileSetup) =
-	{
-		import sbinary.DefaultProtocol.{immutableMapFormat, immutableSetFormat, StringFormat, tuple2Format}
-		val formats = new sbt.inc.InternedAnalysisFormats()
-		import formats._
-		sbt.IO.gzipFileIn(file)( in => sbinary.Operations.read[(Analysis, CompileSetup)](in) )
-	}
+		sbt.IO.gzipFileIn(file)( in => TextAnalysisFormat.read(new BufferedReader(new InputStreamReader(in))) )
```
gkossakowski: You mention in commit message that gzip is not used anymore. Why is it used here?

benjyw: Ooops, missed that. Just pushed a commit to fix it.

compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
```diff
@@ -0,0 +1,312 @@
...
+
+	def time[T](key: String)(f: => T) = {
+		val ret = aggregate(key)(f)
+		close(key)
+		ret
+	}
+
+	def close(key: String) {
+		if (printTimings) {
+			println("[%s] %dms".format(key, timers.getOrElse(key, 0L) / 1000000))
```
gkossakowski, line 32: I was thinking whether we would like to use proper logging API here but I'm guessing that just for printing timings this should be ok.

compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
```diff
@@ -0,0 +1,312 @@
...
+		val infos = FormatTimer.time("read sourceinfos") { SourceInfosF.read(in) }
+		val compilations = FormatTimer.time("read compilations") { CompilationsF.read(in) }
+		val setup = FormatTimer.time("read setup") { CompileSetupF.read(in) }
+
+		(Analysis.Empty.copy(stamps, apis, relations, infos, compilations), setup)
+	}
+
+	object RelationsF {
+		object Headers {
+			val srcProd = "products"
```
gkossakowski, line 108: Do we want this to be public?

compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
```diff
@@ -0,0 +1,312 @@
...
+		val stamps = FormatTimer.time("read stamps") { StampsF.read(in) }
+		val apis = FormatTimer.time("read apis") { APIsF.read(in) }
+		val infos = FormatTimer.time("read sourceinfos") { SourceInfosF.read(in) }
+		val compilations = FormatTimer.time("read compilations") { CompilationsF.read(in) }
+		val setup = FormatTimer.time("read setup") { CompileSetupF.read(in) }
+
+		(Analysis.Empty.copy(stamps, apis, relations, infos, compilations), setup)
+	}
+
+	object RelationsF {
```
gkossakowski: @benjyw, do you intentionally made this object and all inner objects public?

compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
benjyw, line 32: Well, this only prints when that sysprop is set, and I want to get rid of
this anyway, once we've convinced ourselves that the code is reasonably
optimal. I had used these timers to squeeze performance gains out anywhere
I could, but the main remaining issue is the sbinary serialization of the
API objects...

compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
benjyw: No, they could all be private.

gkossakowski:
Overall, looks very good to me. Once questions related to access of newly introduced apis are addressed I think we can merge it.

I tested it on a few projects (scalacheck, scala compiler, scalatest) and it worked well everywhere.

Great work @benjyw! I wanted text-based format for Analysis for a long time!

benjyw:
Pushed a commit that makes all the inner objects private[this].

harrah:
The last comment I have is that a proposed fix for #979 is to version the incremental compiler cache.  Do you have an opinion on how to version it now that we are changing to a text format?  A version header or include it in the file name?

benjyw:
I like that idea. Happy to add a version header inside the file. I don't
like relying on file names for such things.

harrah:
Ok, thanks.  That + Stamps.toString/fromString and we are set.

benjyw:
What behavior to you want when the version in the file isn't compatible
with the one the format code can handle? Throw a ReadException?

harrah:
Yes, I think that should work.  There used to be code that would trap any exception while reading from the cache and return an empty Analysis.  As long as that code is still triggered, a ReadException is good.

benjyw:
Ya, that code is still there in FileBasedStore.

OK, pushed a commit with both those changes. I think I've addressed
everything?

harrah:
Yep, LGTM.  Thanks!

gkossakowski:
I looked at all new commits and LGTM as well.

I was wondering, since we have `TestCaseGenerators` that define scalacheck generators for Analysis object would it make sense to write a simple scalacheck test that checks whether for `a: Analysis` you have `read(write(a)) == a`? This could come as a separate PR. @benjyw, wdyt?

For now, let's squash all commits into one and I think we are ready to merge.

benjyw:
Closed in favor of #988.

benjyw:
Sent a new pull request with all the squashes. And yes - unit test for
load/save would be a good idea.

PR #996, opened 2013-11-21
Extract dependencies by tree walking

gkossakowski:
_This is a preview of a PR that is coming soon. It's not finished yet but I decided to send it so I can get @harrah's feedback early_

This PR introduces new `memberRef` and `inheritance` relations. Also, the logic that populates those relations by tree walking is provided.

I'd love to hear if this is going in the right direction. Most of commit messages has been updated so they are up-to-date with patches I submit here. They could help with answering some of the questions that might arise during the review.


harrah:
This fixes some pending tests, right?  Will any existing tests now fail?   Should there be new pending tests to handle new cases where it will fail to track dependencies?

gkossakowski:
@harrah: it will fix some of functional tests, e.g. `source-dependencies/import-class`. However, it will happen only when the rest of name-hashing branch is merged so new dependency relations are used in invalidation logic.

There's one more complication that I don't know how to solve: how to use scripted tests with both `nameHashing` flag (that will be introduced to `IncOptions`) enabled and disabled? We could use `set` command in `test` file but then it's not really easy to say whether the text should be marked as passing or pending. What's your opinion on this? Do you have better ideas?

Also, if you look cc279ad you'll realize that it tests similar scernarios as `source-dependencies/inherited-dependencies`. I made it to be an unit test so I have code completion in Eclipse and much faster turn-around time. However, another way of looking at cc279ad is that it simply duplicates `source-dependencies/inherited-dependencies` and allows me to test new relations independently. Are you ok with this?

Answering your question about new failing tests with name-hashing: yes, there will be some (e.g. `constants`) due to imprecision of tracking dependencies and used names. Tracking dependencies could be work-arounded by using `CompilationUnit.depends` but there's nothing we can do about missing used names.

Ultimately, this will have to be fixed on scala side and I was hoping to do that in 2.11.0-M8 but I'm running short of time so it might need to be pushed to 2.12, unfortunately.

Let me know if this answers all of your questions.

harrah:
Thanks for the clarifications.

I agree that if `source-dependencies/inherited-dependencies` mainly tests that inherited dependencies are properly tracked, it doesn't make sense to keep it in addition to the unit test.

Do you think both nameHashing and the existing algorithm need to be tested simultaneously?  It is a transition, so perhaps as long as you run the scripted tests with nameHashing on, it will be ok until 0.14.0 when we remove non-nameHashing.  Not ideal, but I guess the alternative is to extend scripted to somehow run with multiple configurations.  (You could add procedures, but...)

As long as the tests that fail are marked pending and new tests are introduced to mark new problems, that's fine.  If this isn't yet the time to do that, then when the rest of name-hashing is introduced works.

gkossakowski:
> I agree that if source-dependencies/inherited-dependencies mainly tests that inherited dependencies are properly tracked, it doesn't make sense to keep it in addition to the unit test.

My plan was to keep it as long as the old dependency tracking logic is in place. In the new unit test, I would test the same functionality but the new dependency tracking is exercised. Does this sound ok?

I agree with you that since this is a transition we don't want to spend too much time on engineering the test framework. Actually, I like your idea of running all tests with nameHashing enabled, identifying the tests for which results differ and document those cases by introducing new tests that specifically target name hashing algorithm.

Also, I should mention that my general strategy is that I'll be unit testing each component individually as I introduce them and only once the whole algorithm is merged I'll focus on functional (scripted) tests because only then I have all pieces that I can test together in place. As a reminder, the pieces I'm talking about are:
- new dependency tracking
- used names tracking
- hashing of names (more precisely of all definitions grouped by simple names)
- changes to invalidation algorithm which leverage all information from above components

For the first 3 I have unit tests that I'll be submitting along with PRs that introduce those components. Does this strategy sound good to you?

gkossakowski:
Closing this PR. The follow up PR that is ready for merge (after being reviewed of course) is coming soon.

PR #1002, opened 2013-11-24
Extract dependencies by tree walking

gkossakowski:
_This is follow up on #996. I consider this PR to be ready for review._

At very high level this PR introduces a new source dependency tracking mechanism based on tree walking and collecting symbols. Here are two commit messages which describe the best this PR in detail:

---
#### Introduce `memberRef` and `inheritance` relations.

Introduce an alternative source dependency tracking mechanism that is needed by upcoming name hashing algorithm. This new mechanism is implemented by introducing two new source dependency relations called `memberRef` and `inheritance`.

Those relations are very similar to existing `direct` and `publicInherited` relations in some subtle ways. Those differences will be highlighted in the description below.

Dependencies between source files are tracked in two distinct categories:
- dependencies introduced by inheriting from a class/trait defined in other source file
- dependencies introduced by referring (selecting) a member defined in other source file (that covers all other kinds of dependencies)

Due to invalidation algorithm implementation details sbt would need to track inheritance dependencies of public classes only. Thus, we had relation called `publicInherited`. The name hashing algorithm which improves invalidation logic will need more precise information about dependencies introduced by inheritance including dependencies of non-public classes. That's one difference between `inheritance` and `publicInherited` relations.

One surprising (to me) thing about `publicInherited` is that it includes all base classes of a given class and not just parents. In that sense `publicInherited` is transitive. This is a bit irregular because everything else in Relations doesn't include transitive dependencies.

Since we are introducing new relations we have an excellent chance to make things more regular. Therefore `inheritance` relation is non-transitive and includes only extracted parent classes.

The access to `direct`, `publicInherited`, `memberRef` and `inheritance` relations is dependent upon the value of `memberRefAndInheritanceDeps` flag. Check documentation of that flag for details.

The two alternatives for source dependency tracking are implemented by introduction of two subclasses that implement Relations trait and one abstract class that contains some common logic shared between those two subclasses. The two new subclasses are needed for the time being when we are slowly migrating to the name hashing algorithm which requires subtle changes to dependency tracking as explained above. For some time we plan to keep both algorithms side-by-side and have a runtime switch which allows to pick one. So we need logic for both old and new dependency tracking to be available. That's exactly what two subclasses of MRelationsCommon implement. Once name hashing is proven to be stable and reliable we'll phase out the old algorithm and the old dependency tracking logic.

---
#### Extract source code dependencies by tree walking.

Previously incremental compiler was extracting source code dependencies by inspecting `CompilationUnit.depends` set. This set is constructed by Scala compiler and it contains all symbols that given compilation unit refers or even saw (in case of implicit search). There are a few problems with this approach:
- The contract for `CompilationUnit.depend` is not clearly defined in Scala compiler and there are no tests around it. Read: it's not an official, maintained API.
- Improvements to incremental compiler require more context information about given dependency. For example, we want to distinguish between dependency on a class when you just select members from it or inherit from it. The other example is that we might want to know dependencies of a given class instead of the whole compilation unit to make the invalidation logic more precise.

That led to the idea of pushing dependency extracting logic to incremental compiler side so it can evolve indepedently from Scala compiler releases and can be refined as needed. We extract dependencies of a compilation unit by walking a type-checked tree and gathering symbols attached to them.

Specifically, the tree walk is implemented as a separate phase that runs after pickler and extracts symbols from following tree nodes:
- `Import` so we can track dependencies on unused imports
- `Select` which is used for selecting all terms
- `Ident` used for referring to local terms, package-local terms and top-level packages
- `TypeTree` which is used for referring to all types

Note that we do not extract just a single symbol assigned to `TypeTree` node because it might represent a complex type that mentions several symbols. We collect all those symbols by traversing the type with CollectTypeTraverser. The implementation of the traverser is inspired by `CollectTypeCollector` from Scala 2.10. The `source-dependencies/typeref-only` test covers a scenario where the dependency is introduced through a TypeRef only.


gkossakowski:
I forgot to mention that all code review suggestions  from #996 has been addressed.

harrah:
Everything addressed => LGTM.

gkossakowski:
For the record, the 7579704 has been reworked since #996 so it correctly protects the access to source dependency relations depending on `memberRefAndInheritanceDeps` flag's value. That cause some ugliness in the code because I had to rename constructor arguments in order to be able to override getters for those relations. All that ugly code will go away in 0.14 so maybe it's not that bad.

Other than that you might want to check how I implemented reading/writing Analysis using text format. Again, the code is a big ugly but will be cleaned up once `direct` and `publicInherited` relations are phased out. While writing this comment I realized that I should have increased version number of the text format... I'll do that shortly.

The last thing I wanted to highlight is 8a6b914 where I noted that there's a bug in current dependency tracking mechanism. This reaffirmed my belief that adding test cases that document _current_ behavior and show the new behavior as a diff is very beneficial.

harrah:
Thanks for pointing out the pieces where additional review was useful.  I'm ok with these changes and I agree the format version should be bumped.  Let's have @benjyw review 7579704 if he doesn't mind.

benjyw:
I'll take a look.

benjyw:
[Sorry, previously replied to the wrong email]

A couple of general comments/questions:
- You mention that publicInherited includes all transitive bases. I had
  been assuming that publicInherited is a strict subset of direct, because
  the comments (e.g.,
  https://github.com/sbt/sbt/blob/0.13/compile/inc/src/main/scala/sbt/inc/Relations.scala#L65)
  strongly imply this. I will have to revisit some of my code to see if
  anything is broken under this assumption. Probably a good idea to change
  the comments appropriately?
- It's not clear from the commit message, but are memberRef and inheritance
  replacing direct and publicInherited, or augmenting them?

gkossakowski:
> - You mention that publicInherited includes all transitive bases. I had been assuming that publicInherited is a strict subset of direct, because the comments (e.g., https://github.com/sbt/sbt/blob/0.13/compile/inc/src/main/scala/sbt/inc/Relations.scala#L65) strongly imply this. I will have to revisit some of my code to see if anything is broken under this assumption. Probably a good idea to change the comments appropriately?

The comment and contract still holds modulo bugs (see commit message in 8a6b914). The `direct` relation is not very well specified (`memberRef` tries a little harder but is still far from being perfect) but if you look at how that relation is populated you'll see that it should include all base classes as well.

> - It's not clear from the commit message, but are memberRef and inheritance replacing direct and publicInherited, or augmenting them?

Eventually, they will replace `direct` and `publicInherited` relations once the rest of my work (which relies on new dependency tracking relations) is merged and battle tested. For now, we'll have a flag which controls which algorithm (and which dependency tracking logic) is being used. The default will be to use the old code. Therefore I shouldn't break any existing clients of incremental compiler.

The fact that merging/grouping is not implemented for new dependency tracking mechanism is not set in stone. I just had to reduce the scope of my work to something manageable so I can get the whole algorithm merged fairly soon and we can start testing it. There will be enough of time for you to catch up with changes to Relations before all of my patches arrive.

Let me know if that addresses your concerns.

gkossakowski:
@benjyw: I addressed all of your review feedback. I also reworded description of b837169 so it (hopefully) better communicates the intent and rationale behind those changes.

Also, I added a test case in 331fffb as we discussed.

gkossakowski:
@harrah: The b837169 looks significantly different from what you saw previously that you might want to have another (even cursory) look. I don't think there's anything controversial in there, though.

gkossakowski:
I talked to @harrah and he gave me green light. Merging!

PR #1005, opened 2013-11-25
Work harder on generating unique values in generator for Analysis.

gkossakowski:
The TestCaseGenerators uses global set for ensuring that certain generated
values are unique. This is not the best design because the more properties
you check the harder is to generate new sample inputs because of already
accumulated values. This results in:

[info] + Analysis.Simple Merge and Split: OK, proved property.
[info] ! Analysis.Complex Merge and Split: Gave up after only 8 passed tests. 93 tests were discarded.

I don't have an ambition to reduce the scope of this global set but at
least I wanted to make generators to work a bit harder on generating
samples.

Instead of using `suchThat` method for filtering out non-unique samples
we use `retryUntil` that never gives up (therefore it might not
terminate).

Also, I overridden the `identifier` to delegate to original
`Gen.identifier` but with minimal size set to be to '3'. This means,
the generated identifier will be of size 3 or larger which is needed in
order to avoid collisions.

### DIFF
--- a/compile/inc/src/test/scala/sbt/inc/TestCaseGenerators.scala
+++ b/compile/inc/src/test/scala/sbt/inc/TestCaseGenerators.scala
@@ -27,7 +27,12 @@ object TestCaseGenerators {
 	// Ensure that we generate unique class names and file paths every time.
 	// Using repeated strings may lead to all sorts of undesirable interactions.
 	val used = scala.collection.mutable.Set.empty[String]
-	def unique[T](g: Gen[T]) = g suchThat { o: T => used.add(o.toString) }
+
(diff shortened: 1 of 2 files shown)


gkossakowski:
Review by @harrah.
//cc @benjyw

gkossakowski:
@harrah: I upgraded to the latest release of scalacheck and updated the code accordingly.

I hope nothing will break as a side effect of this upgrade, though.

harrah:
Thanks.  Upgrading is indeed always a risk.  LGTM.

adriaanm:
I worked with Ricky on the upgrade to 1.11.1 from 1.11.0 to fix a bug discovered by running the scala test suite on 1.11, I think it's purely a couple of bug fixes.

PR #1006, opened 2013-11-25
Tweak build to include appropriate fixes/adaptations for Scala 2.11.0-M7

jsuereth:
- Limit M7 hackery to one location in the build
- Migrate scala-compiler dependency declarations to use workaround for M7
- fix libModular to work with new module organizations.
- Bump versions of modules used for 2.11 build.

Review by @harrah  cc @adriaanm

Note: This does not include the type fixes in the 0.13-2.11 branch, but includes different build hackery for 2.11.  I'd like to standardize on this mechanism for dbuild (changing scalaVersion in ThisBuild fixes all deps appropriately).

### DIFF
--- a/project/Sbt.scala
+++ b/project/Sbt.scala
@@ -261,7 +261,7 @@ object Sbt extends Build
 		scalacOptions := Nil,
 		ivyScala ~= { _.map(_.copy(checkExplicit = false, overrideScalaVersion = false)) },
 		exportedProducts in Compile := Nil,
-		libraryDependencies <+= scalaVersion( "org.scala-lang" % "scala-compiler" % _ % "provided")
+    libraryDependencies += scalaCompilerInit.value % "provided"
 	)
 	//
 	def compileInterfaceSettings: Seq[Setting[_]] = precompiledSettings ++ Seq[Setting[_]](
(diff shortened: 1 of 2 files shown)


adriaanm:
LGTM -- could we also integrate the type inference fixes from the 0.13-2.11 branch (those that are still required -- would be good to figure out if these are 2.11 regressions, as they imply a lack of source compatibility)?

harrah:
Some fixes were internal apis: Global seems to have acquired a Range type, for example.  (This isn't in 0.13 because I didn't immediately see an easy/brief way to hide that in a 2.10 compatible way.)  I recall 5fe35be7a32c506c29cb830a2c77a6fc20f9e35e being required due to a correctness fix.  (This change is included in 0.13.)  e6437fdcc7073b8a4cd0a3e65c0ea96999e7ee25 looks like it has a singleton inference workaround and maybe a workaround for an existential inference (not a fancy existential or anything).  (These are not in the 0.13 branch.)

adriaanm:
Happy to clean up things on our end to enable consolidating these branches. Ultimately, that's what integration is about to me -- discovering these issues and fixing them where appropriate. If we added unnecessary bloat to Global, we should trim it. Especially if it breaks sbt.

adriaanm:
@JamesIry is responsible for source compatibility between 2.10, and 2.11, so I'll let him investigate further.

jsuereth:
This is out of date now.  I'm going to close for now until we can consolidate 2.11 fixes in one spot and canonicalize the build hooks between regular 0.13/0.14/1.0 and the various scala verisons used.

Issue #1010, opened 2013-11-27
Merge name hashing branch that improves incremental compiler


gkossakowski:
~~This is a meta-issue that allows people to get quick overview of where we are with merging [`name-hashing`](https://github.com/gkossakowski/sbt/compare/sbt:0.13...name-hashing) branch back to sbt code base.~~

Name hashing has been merged to sbt and released with sbt 0.13.2-M1!
### Current status

The name hashing branch consists of 6 major components/changesets:
- [x] more fine-grained logging which helps to understand why files get invalidated (#941)
- [x] an option to control incremental compiler's behavior related to macro definitions (#976)
- [x] new source dependency tracking logic that is based on tree walking. The tree walk extracts dependencies introduced through member reference and inheritance (#1002)
- [x] tracking of used names in Scala source files (#1013)
- [x] hashing of public definitions grouped by simple name (#1026)
- [x] changes to incremental compiler algorithm that makes use of all data described above and improve the invalidation logic (#1042)

Only once the last bullet is merged you'll be able to see (and test) changes to incremental compiler behavior.
### Targeted sbt version

At the moment we are targeting shipping improved incremental compiler with sbt 0.13.2. It's realistic we can do that but obviously there's no commitment on our side that we'll make it. Plans can always change.
### Why would I care?

If you have large enough Scala project of certain (quite common) structure then you might suffer from the problem that sbt's incremental compiler recompiles too much when you modify your project. A good example are Play projects. [Here's](https://gist.github.com/jroper/6374383) comparison of incremental compiler behavior in different sbt versions for some sample Play project and typical code modification scenarios. As you can see, in some cases the improvement can be quite dramatic.

NOTE: Once name hashing is merged not only sbt will benefit. Zinc will see the same improvements. Scala IDE (Eclipse plugin) and IntelliJ embed zinc so they will have better incremental compilation as well.
### What's the basic idea?

I wrote some [documentation](https://github.com/gkossakowski/sbt/wiki/Incremental-compiler-notes) that explains current incremental compiler design and challenges related to incremental compilation in Scala. There's also a [chapter](https://github.com/gkossakowski/sbt/wiki/Incremental-compiler-notes#name-hashing-algorithm) that explains the basic idea behind the name hashing algorithm.

retronym:
(edited ticket to link your WIP branch for the curious)

adriaanm:
dundundun! the end is extremely nigh!

gkossakowski:
[...]

PR #1013, opened 2013-11-28
Used names extraction logic

gkossakowski:
This PR introduces another piece of name hashing algorithm as mentioned in #1010.

The first two (88444f2 and 40dc5cd) are purely cleanups/refactorings and do not change any behavior. The description of the third is copied below for convenience.

---

Tracking of used names is a component needed by the name hashing algorithm. The extraction and storage of used names is active only when `AnalysisCallback.nameHashing` flag is enabled and it's disabled by default.

This change constists of two parts:
1. Modification of Relations to include a new `names` relation that allows us to track used names in Scala source files
2. Implementation of logic that extracts used names from Scala compilation units (that correspond to Scala source files)

The first part is straightforward: add standard set of methods in Relations (along with their implementation) and update the logic which serializes and deserializes Relations.

The second part is implemented as tree walk that collects all symbols associated with trees. For each symbol we extract a simple, decoded name and add it to a set of extracted names. Check documentation of `ExtractUsedNames` for discussion of implementation details.

The `ExtractUsedNames` comes with unit tests grouped in `ExtractUsedNamesSpecification`. Check that class for details.

Given the fact that we fork while running tests in `compiler-interface` subproject and tests are ran in parallel which involves allocating multiple Scala compiler instances we had to bump the default memory limit.

This commit contains fixes for gkossakowski/sbt#3, gkossakowski/sbt#5 and gkossakowski/sbt#6 issues.

### DIFF
--- a/compile/inc/src/main/scala/sbt/inc/Analysis.scala
+++ b/compile/inc/src/main/scala/sbt/inc/Analysis.scala
@@ -59,7 +59,7 @@ object Analysis
 
 	/** Merge multiple analysis objects into one. Deps will be internalized as needed. */
 	def merge(analyses: Traversable[Analysis]): Analysis = {
-		if (analyses.exists(_.relations.memberRefAndInheritanceDeps))
+		if (analyses.exists(_.relations.nameHashing))
 			throw new IllegalArgumentException("Merging of Analyses that have" +
 				"`relations.memberRefAndInheritanceDeps` set to `true` is not supported.")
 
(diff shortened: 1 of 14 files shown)


harrah:
LGTM overall- just a couple of questions.

harrah:
Ok, perhaps open a ticket on your fork for the isSynthetic problem and let's merge this?

gkossakowski:
Good idea. I filed a ticket at gkossakowski/sbt#9.

Also, I pushed a new commit which addresses other two minor comments. Merging.

PR #1018, opened 2013-11-29
Documentation : use isSnapshot setting instead of "version.value.trim.endsWith("SNAPSHOT")"

pdalpra:
Since SBT provides the `isSnapshot` setting to check whether the current version is a snapshot version or not, I thought it could be nice to use it here instead of reimplementing it :)


harrah:
Makes sense.  Merged in 621f9dc568302a9de8c9a1c0c104128a9d7fd3a4, thanks!

PR #1019, opened 2013-11-30
Make incremental compiler compatible with Scala 2.11.

gkossakowski:
The scala/scala@2d4f0f1859b957b744f9b9f222dec8e8c478a4a8 removes the
`toplevelClass` method. The recent change from
aac19fd02be94f4ef6ba98187c9cbbc2b66a60f9 introduces dependency on that
method. Combination of both changes makes incremental compiler incompatible
with Scala 2.11.

This change introduces a compatibility hack that brings back source
compatibility of incremental compiler with Scala 2.8, 2.9, 2.10 and 2.11.

The compatibility hack is making clever use implicit conversions that
can provide dummy method definitions for methods removed from Scala
compiler.

Also, the code that depends on `enclosingTopLevelClass` has been refactored
so the dependency is more centralized.

### DIFF
--- a/compile/interface/src/main/scala/xsbt/Compat.scala
+++ b/compile/interface/src/main/scala/xsbt/Compat.scala
@@ -4,8 +4,39 @@ import scala.tools.nsc.Global
 import scala.tools.nsc.symtab.Flags
 
 /**
- *  Collection of hacks that make it possible for the compiler interface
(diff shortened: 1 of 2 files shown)


gkossakowski:
@harrah: feedback addressed.

harrah:
Thanks, you are all set on this one. LGTM.

PR #1021, opened 2013-12-01
Merge 0.13 to 0.13-2.11

retronym:
And then bump to Scala 2.11.0-M7

Review by @harrah

/cc @cunei


retronym:
My motivation is to get dbuild for 2.11 working, and zinc master [doesn't build](https://jenkins-dbuild.typesafe.com:8499/job/Community-2.11.x/39/consoleFull) against `0.13-2.11`.

This is just a straight up merge with only trivial conflicts.

But it seems to have broken somethings:

Before the merge, I couldn't build at all

```
[trace] Stack trace suppressed: run last cache/*:update for the full output.
[trace] Stack trace suppressed: run last process/*:update for the full output.
[error] (cache/*:update) sbt.ResolveException: unresolved dependency: org.scala-lang.modules#scala-xml_2.11;1.0-RC2: not found
[error] (process/*:update) sbt.ResolveException: unresolved dependency: org.scala-lang.modules#scala-xml_2.11;1.0-RC2: not found
```

After the merge:

```
[error] (completion/test:compile) Compilation failed
[error] (actions/test:test) sbt.TestsFailedException: Tests unsuccessful
[error] (collections/test:compile) Compilation failed
[error] (incremental-compiler/test:compile) Compilation failed
[error] (main/test:compile) Compilation failed
[error] (task-system/test:test) sbt.TestsFailedException: Tests unsuccessful
[error] (io/test:compile) Compilation failed
[error] (logging/test:compile) Compilation failed
[error] (classpath/test:compile) Compilation failed
[error] Total time: 247 s, completed Dec 2, 2013 12:08:39 AM
```

Dbuild wants to build:

``` json
{
    name:   "sbt",
    uri:    "git://github.com/sbt/sbt.git#0.13-2.11"
    extra: {
      projects: ["compiler-interface",
                 "classpath","logging","io","control","classfile",
                 "process","relation","interface","persist","api",
                 "compiler-integration","incremental-compiler","compile",
                 "launcher-interface"
                ],
      run-tests: false
    }
  }
```

@adriaanm @cunei Thoughts on this branch / merge?

harrah:
I would prefer seeing the relevant changes from 0.13-2.11 rebased on a stable branch, such as 0.13.1 when it gets released as final.  It would be best to use a stable version of zinc to go with that as well.  If it is necessary to do it sooner than 0.13.1 final, then 0.13.1-RC5 should be the base.

adriaanm:
The difference is essentially one commit (as seen here: https://github.com/adriaanm/sbt/compare/sbt:0.13.1...0.13-2.11), so I agree rebasing is probably the way to go. In fact, e6437fdcc707 seems to indicate regressions in source compatibility.

adriaanm:
I believe @jamesiry was going to look into this last week.

retronym:
> I would prefer seeing the relevant changes from 0.13-2.11 rebased on a stable branch, such as 0.13.1

Do you mean that we should submit the changes back to the 0.13 branch, and retire 0.13-2.11? Having a single codebase/build that supports 2.10 and 2.11 does seem like the best outcome.

retronym:
One thing I noticed is that `sbinary` is published against `2.11.0-M4` only: http://repo.typesafe.com/typesafe/ivy-releases/org.scala-tools.sbinary/

retronym:
I've just pushed one more commit to bump to 2.11.0-M7, using the old sbinary, and working around SI-8023.

Please let me know what you need with regards to rebasing or retargetting these changes.

retronym:
BTW, I don't see why we would rebase then `push -f` _this_ branch. Merging seems more natural to me. I would see the value in retiring this branch and rebasing the changes onto a new PR against `0.13`, however.

retronym:
>  In fact, e6437fd seems to indicate regressions in source compatibility.

Mark reported the regression that required explicit type arguments of `[c.type]`. https://issues.scala-lang.org/browse/SI-7688. It is since been fixed. I've just pushed a commit that reverts that workaround.

I'll look at the other changes now.

harrah:
I agree a single codebase would be nice, but it is harder to handle the temporary workarounds.  This way, you can make the workaround on the 2.11 branch and not worry about making it compatible with 2.10 as well.

The rebase was requested because the merge generated a lot of noise on the pull request.  The rebase could be pushed to a new branch, but this branch is used just for integration and not development so I think a new branch would mainly just make us change branch names in integration configurations.

It is nice to see the previous workarounds are no longer necessary.  For the new workaround, do you think it will get fixed for 2.11 or will it be a necessary regression?

retronym:
I've just pushed a cast-free workaround. It looks like a deep issue to fix correctly in the compiler (cycle avoidance), so I can't promise a timeline for a fix.

harrah:
That's a fancy workaround.  I was just wondering whether it would survive for a while or not.  I understand these things can be tricky to resolve and that the fix that exposed the problem might be more important than this situation.  Thanks for this improved hack.

retronym:
@harrah We've fixed the underlying bug in scalac now. So after M8 (or sooner, under dbuild), we should need purely build changes to support 2.11. Those build changes are easy enough to express conditionally on `scalaVersion.value.startsWith("2.11")`, so ideally we can just retire this branch altogether and make that change on `0.13`. WDYT?

harrah:
I think the first step is to update this branch on say 0.13.1-RC5 (should be final early next week) and see how it looks.

jsuereth:
@harrah I think we need to update/merget this soon.   Our 2.11 is way behind at this point.

jsuereth:
CLosing this in favor of @gkossakowski's patch.   No longer need the gymnastics in types thanks to @retronym's bug squashing.

retronym:
Yay for fewer branches!

PR #1029, opened 2013-12-05
Access `Relations.names` only when name hashing is enabled

gkossakowski:
Check individual commits for details.

This PR fixes broken scripted tests. I'm sorry for the regression!


harrah:
No problem, thanks for the fix.

PR #1039, opened 2013-12-13
fix plugin documentation with current example

opyate:
The portions of [0.13.1 Plugin documentation](http://www.scala-sbt.org/0.13.1/docs/Extending/Plugins.html) that I removed in my commit doesn't work for me, and I have a reproducible test here:   https://gist.github.com/opyate/7943389

### DIFF
--- a/src/sphinx/Extending/Plugins.rst
+++ b/src/sphinx/Extending/Plugins.rst
@@ -95,110 +95,39 @@ By Example
 Using a library in a build definition
 -------------------------------------
 
-As an example, we'll add the Grizzled Scala library as a plugin.
-Although this does not provide sbt-specific functionality, it
-demonstrates how to declare plugins.
+As an example, we'll add the `xsbt-web-plugin
+<https://github.com/JamesEarlDouglas/xsbt-web-plugin>`_.
 
-1a) Manually managed
-~~~~~~~~~~~~~~~~~~~~
-
-1. Download the jar manually from
-   https://oss.sonatype.org/content/repositories/releases/org/clapper/grizzled-scala\_2.8.1/1.0.4/grizzled-scala\_2.8.1-1.0.4.jar
-2. Put it in `project/lib/`
-
-1b) Automatically managed: direct editing approach
+1) Automatically managed: direct editing approach
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Edit `project/plugins.sbt` to contain:
 
 ::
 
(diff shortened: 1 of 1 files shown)


src/sphinx/Extending/Plugins.rst
```diff
@@ -95,110 +95,39 @@ By Example
...
 Edit `project/plugins.sbt` to contain:
 
 ::
 
-    libraryDependencies += "org.clapper" %% "grizzled-scala" % "1.0.4"
+    addSbtPlugin("com.earldouglas" % "xsbt-web-plugin" % "0.5.0")
 
 If sbt is running, do `reload`.
 
-1c) Automatically managed: command line approach
```
jsuereth, line 120: This is still a valid mechanism to configure plugins.  Do you really think we should drop it?

src/sphinx/Extending/Plugins.rst
```diff
@@ -95,110 +95,39 @@ By Example
...
-    ...
-
-To switch back to the main project:
-
-.. code-block:: console
-
-    > reload return
-    [info] Set current project to root (in build file:/Users/harrah/demo2/)
-
-1d) Project dependency
```
jsuereth, line 153: I also use this a bit, so I'd like to keep this doc around.

jsuereth:
Very much appreciate the pull request to improve docs.  IF you leave the sections you've removed around, I'll update them with more up-to-date information.

Thanks!

src/sphinx/Extending/Plugins.rst
opyate, line 120: Neither `set libraryDependencies += "com.earldouglas" % "xsbt-web-plugin" % "0.5.0"` nor `addSbtPlugin("com.earldouglas" % "xsbt-web-plugin" % "0.5.0")` works from the CLI for me (even after `reload plugins`).

Also, not only does `libraryDependencies += "com.earldouglas" % "xsbt-web-plugin" % "0.5.0"` not work from the CLI -- it doesn't work from project/myplugin.sbt.

I tried all this on a clean Linux install, as per https://gist.github.com/opyate/7943389. I.e. no ~/.sbt, no .sbtopts, etc.

Please provide an example here and I'll incorporate it into the pull request.

jsuereth, line 120: That would be 

```
reload plugins

set addSbtPlugin("com.earldouglas" % "xsbt-web-plugin" % "0.5.0")

session save

reload return
```

eed3si9n:
I'm closing this due to inactivity

Issue #847, opened 2013-08-14
upgrade to ivy 2.3.0 (final)


jdolson:
sbt 0.13.0 is still using ivy 2.3.0-rc1 (released Apr. 2012). The final was [released](http://ant.apache.org/ivy/history/2.3.0/release-notes.html) in Jan. 2013 and includes several fixes not in rc1 (or even rc2). The second RC broke two public APIs sbt uses and so the fix is nontrivial. It is too late to fix this for 0.13 but it could be fixed for 0.14.

zavg:
The sbt 0.13.0 is still depends on ivy 2.3.0-rc1, which, for example, consists critical bug [[IVY-1378]](https://issues.apache.org/jira/browse/IVY-1378) which does not allow run sbt on Windows if the name of current user contains some non-ascii characters (Umlauts, cyrillic symbols, etc.)

peter-svensson:
IVY 2.3.0-rc1 also contains a bug that handles transitive dependency versions wrong. 

Dependency:

```
libraryDependencies ++= Seq(
  "org.glassfish.jersey.core" % "jersey-client" % "2.4.1"
)
```

When resolving the transitive dependencies of that jar I get errors:

```
org.glassfish.hk2#hk2-utils;2.4.1: not found
org.glassfish.hk2#hk2-locator;2.4.1: not found
```

Looking at the dependency tree I can see that I get the following:

```
[info]   +-org.glassfish.jersey.core:jersey-client:2.4.1
[info]     +-com.google.guava:guava:14.0.1
[info]     +-javax.ws.rs:javax.ws.rs-api:2.0
[info]     +-org.glassfish.hk2.external:javax.inject:2.2.0-b21
[info]     +-org.glassfish.hk2:hk2-api:2.2.0-b21
[info]     | +-org.glassfish.hk2:hk2-utils:2.4.1 (error: not found)
[info]     | 
[info]     +-org.glassfish.hk2:hk2-locator:2.4.1 (error: not found) 
```

Looking in the pom.file for hk2-api (which is the file that requires hk2-utils) [http://repo1.maven.org/maven2/org/glassfish/hk2/hk2-api/2.2.0-b21/hk2-api-2.2.0-b21.pom] we see that:

```
<dependencies>
...
<dependency>
<groupId>org.glassfish.hk2</groupId>
<artifactId>hk2-utils</artifactId>
<version>${project.version}</version>
</dependency>
</dependencies>
```

[...]

PR #1040, opened 2013-12-13
Support Ivy 2.3.0-final.

willb:
This entailed modifying ResolutionCache and the CustomPomParser
to reflect changes to the ResolutionCacheManager interface and
DefaultExtendsDescriptor class between Ivy 2.3.0-rc1 and
2.3.0-rc2. Specifically,
1. ResolutionCacheManager now includes two additional methods
   that needed implementations in ResolutionCache:
   getResolvedModuleDescriptor(mrid: ModuleRevisionId) and
   saveResolvedModuleDescriptor(md: ModuleDescriptor). I adapted
   the implementations for these (which are expressed primarily in
   terms of other interface methods) from Ivy 2.3.0's
   DefaultResolutionCacheManager.
2. Instead of taking a ModuleRevisionIdentifier and a resolved
   ModuleRevisionIdentifier as its first two arguments, the
   DefaultExtendsDescriptor constructor now takes a
   ModuleDescriptor.

Note that ResolutionCache.getResolvedModuleDescriptor does not
appear to be used by Ivy as sbt uses Ivy and there is thus no
test coverage for its implementation. Also note that the
DefaultResolutionCacheManager object created in
Update.configureResolutionCache now requires a reference to an
IvySettings object; DRCM expects this to be non-null.

### DIFF
--- a/ivy/src/main/scala/sbt/CustomPomParser.scala
+++ b/ivy/src/main/scala/sbt/CustomPomParser.scala
@@ -203,7 +203,7 @@ object CustomPomParser
 		val unique = IvySbt.mergeDuplicateDefinitions(withExtra)
 		unique foreach dmd.addDependency
 
-		for( ed <- md.getInheritedDescriptors) dmd.addInheritedDescriptor( new DefaultExtendsDescriptor( mrid, resolvedMrid, ed.getLocation, ed.getExtendsTypes) )
+		for( ed <- md.getInheritedDescriptors) dmd.addInheritedDescriptor( new DefaultExtendsDescriptor( md, ed.getLocation, ed.getExtendsTypes) )
 		for( conf <- md.getConfigurations) {
 			dmd.addConfiguration(conf)
 			for(art <- md.getArtifacts(conf.getName)) {
--- a/ivy/src/main/scala/sbt/Ivy.scala
+++ b/ivy/src/main/scala/sbt/Ivy.scala
@@ -310,7 +310,7 @@ private object IvySbt
 	private[this] def configureResolutionCache(settings: IvySettings, localOnly: Boolean, resCacheDir: Option[File])
 	{
 		val base = resCacheDir getOrElse settings.getDefaultResolutionCacheBasedir
-		settings.setResolutionCacheManager(new ResolutionCache(base))
+		settings.setResolutionCacheManager(new ResolutionCache(base, settings))
 	}
 	// set the artifact resolver to be the main resolver.
 	// this is because sometimes the artifact resolver saved in the cache is not correct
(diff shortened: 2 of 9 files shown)


harrah:
LGTM, thank you!

I had previously thought this should go into 0.14 due to the risk of binary incompatible changes.  It looks like the sbt side is ok, but I don't know about Ivy itself.  Problems would occur if a) Ivy has some incompatibilities and b) a plugin uses Ivy directly and is affected by them.

@jsuereth @eed3si9n (and anyone watching #847) any thoughts?

As a follow up commit, perhaps do a similar doc refresh as was done in https://github.com/xuwei-k/xsbt/commit/d5d4d8377e3f6030163bbca0755ebd0492208aa5 ?

jsuereth:
Wow nice!  I think we may need to push a milestone of 0.13.2 with this and get teh community/plugins to experiment and see what happens....

Still, Really looking forward to reviewing this.

eed3si9n:
There's no signature changes to public classes or methods, right?
I can't speak for all the plugins, but I personally can't think of any situation I had to drive Ivy directly.

A minor point but the code seems to be using whitespace as opposed to tab. (I wish this was automated)

jsuereth:
There's only one using internal Ivy I know of - sbt-license-report.  IT's fine to break it though, I own that one.

harrah:
@eed3si9n No signature changes to sbt side.  I don't know about Ivy.

@jsuereth That means that sbt-license-report users have to know to bump the sbt-license-report version if they go to 0.13.2, though.

eed3si9n:
There's like [one project](https://github.com/search?q=sbt-license-report+extension%3Asbt+extension%3Ascala&type=Code&ref=advsearch&l=) using sbt-license-report on github, and it's activator. Not saying it's not widely popular among the closed source community, but still.

harrah:
@eed3si9n Compatibility changes need to take into account plugins that we may not know about.  In this case it is indeed @jsuereth's plugin used at Typesafe, but that actually makes it less problematic as he indicates.  I don't have a strong opinion either way, but if there are real plugins using Ivy directly that would be affected, we should consider them.  From the diff, it looks like a rather central class in Ivy was changed.  However, perhaps no one constructs it and they only consume it.

I agree we should publish a milestone and get people to try it out.  Plugins may use Ivy, but it may be that they aren't generally affected.

harrah:
So, I guess that means I'll merge to the 0.13 branch.

harrah:
Sorry, I didn't realize this targeted 0.13.1, which has been released as final.  Can you target the 0.13 branch instead?

willb:
Sorry about that.  I've rebased my patches atop 0.13 and force-updated my branch.  I re-ran the test suite as a sanity check and everything still passes.

harrah:
No problem, thanks for the quick response.  Merged to 0.13 in 5df24e5eb2e44d8f12d978960454942fc44b1ecd.

PR #1042, opened 2013-12-15
Implement name hashing algorithm in incremental compiler

gkossakowski:
This PR implements the name hashing algorithm.

It starts with number of commits that are refactorings of existing functionality. Those changes structure the core of incremental compiler in a way that is makes it easier to plug an alternative (name hashing) implementation of the invalidation logic. Then we have the main commit 729cc41 which provides the implementation (see the description below). Finally, we have two commits that show what's the difference in behavior between the new and old algorithm when it comes to existing test suite. Here comes description of the main commit:

---

Provide implementation of invalidation logic that takes computed name hashes into account. The implementation is spread amongst two classes:
1. `IncrementalNameHashing` which implements a variant of incremental compilation algorithm that computes modified
    names and delegates to `MemberReferenceInvalidationStrategy` when invalidating member reference dependencies
2. `MemberReferenceInvalidationStrategy` which implements the core logic of dealing with dependencies introduced by member reference. See documentation of that class for details.

The name hashing optimization is applied when invalidating source files having both internal and external dependencies (in initial iteration), check `invalidateByExternal` and `invalidateSource` methods for details.

As seen in implementation of `MemberReferenceInvalidationStrategy` the name hashing optimization is not applied when implicit members change.

NOTE: All functionality introduced in this commit is enabled only when `IncOptions.nameHashing` flag is set to true.

The `source-dependencies/transitive-memberRef` test has been changed to test name hashing variant of incremental compilation. The change to invalidated files reflects the difference between the old and the new algorithm.

Also, there a few new tests added that cover issues previously found while testing name hashing algorithm and are fixed in this commit. Each paragraph describes a single test.

Add a test case which shows that detect properly changes to type aliases in the name hashing algorithm. See gkossakowski/sbt#6 for details.

Add test covering bug with use of symbolic names (issue gkossakowski/sbt#5).

Add a test which covers the case where we refer to a name that is declared in the same file. See issue gkossakowski/sbt#3 for details.

### DIFF
--- a/compile/inc/src/main/scala/sbt/inc/Analysis.scala
+++ b/compile/inc/src/main/scala/sbt/inc/Analysis.scala
@@ -56,6 +56,8 @@ trait Analysis
 object Analysis
 {
 	lazy val Empty: Analysis = new MAnalysis(Stamps.empty, APIs.empty, Relations.empty, SourceInfos.empty, Compilations.empty)
+	private[sbt] def empty(nameHashing: Boolean): Analysis = new MAnalysis(Stamps.empty, APIs.empty,
+		Relations.empty(nameHashing = nameHashing), SourceInfos.empty, Compilations.empty)
 
 	/** Merge multiple analysis objects into one. Deps will be internalized as needed. */
 	def merge(analyses: Traversable[Analysis]): Analysis = {
--- a/compile/inc/src/main/scala/sbt/inc/Changes.scala
+++ b/compile/inc/src/main/scala/sbt/inc/Changes.scala
@@ -22,14 +22,14 @@ sealed abstract class APIChange[T](val modified: T)
  * api has changed. The reason is that there's no way to determine if changes to macros implementation
  * are affecting its users or not. Therefore we err on the side of caution.
  */
-case class APIChangeDueToMacroDefinition[T](modified0: T) extends APIChange(modified0)
-case class SourceAPIChange[T](modified0: T) extends APIChange(modified0)
+final case class APIChangeDueToMacroDefinition[T](modified0: T) extends APIChange(modified0)
+final case class SourceAPIChange[T](modified0: T) extends APIChange(modified0)
 /**
  * An APIChange that carries information about modified names.
  *
  * This class is used only when name hashing algorithm is enabled.
  */
-case class NamesChange[T](modified0: T, modifiedNames: ModifiedNames) extends APIChange(modified0)
+final case class NamesChange[T](modified0: T, modifiedNames: ModifiedNames) extends APIChange(modified0)
 
 /**
  * ModifiedNames are determined by comparing name hashes in two versions of an API representation.
@@ -39,7 +39,7 @@ case class NamesChange[T](modified0: T, modifiedNames: ModifiedNames) extends AP
  * on whether modified name is implicit or not. Implicit names are much more difficult to handle
  * due to difficulty of reasoning about the implicit scope.
  */
-case class ModifiedNames(regularNames: Set[String], implicitNames: Set[String]) {
+final case class ModifiedNames(regularNames: Set[String], implicitNames: Set[String]) {
 	override def toString: String =
 		s"ModifiedNames(regularNames = ${regularNames mkString ", "}, implicitNames = ${implicitNames mkString ", "})"
 }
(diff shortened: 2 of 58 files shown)


harrah:
A general comment: doing transitive traversals on individual source files means the time complexity goes up.  For example, if I have `A -> B -> [subgraph G]` and `C -> D -> [subgraph G]` and I ask for `transitive(A)` and then `transitive(C)`, I process all of G each time.  When doing `transitive([A,C])`, I'd only do it once.

I don't necessarily mean that there is a better alternative or that it is a problem in practice, but it might be something to keep in mind.

harrah:
That's all for the first pass.  I'll review it again tomorrow paying more attention to the algorithm and less on the implementation details.

gkossakowski:
> A general comment: doing transitive traversals on individual source files means the time complexity goes up.

Yes, I was aware of that. That's why I said I the end result is the same but I didn't comment on the time complexity. However, when I was benchmarking incremental compiler, the time spent in incremental was usually negligible and it was I/O bound (checking files' meta data).

However, we do some logging while invalidating source files so we'll get more logging output due to my refactoring. I think this is acceptable. In general, I err on too verbose than not verbose enough when it comes to debug logs. It's much easier to filter something out than to modify the code and ship a new binaries in case logging turns out to be not sufficient enough.

gkossakowski:
Thanks! Meanwhile, I'll address your review feedback.

harrah:
@gkossakowski I'm ok merging this when the existing feedback is addressed.  I realized there is no point in me reviewing the algorithm- we've already done that and I'm not going to add anything new without a lot more work and I'm only delaying getting it merged.  I'm otherwise happy with the implementation and I see no reason to keep it out.

gkossakowski:
> For the memberRef one, I mean things like memberRefInvalidationStrategy. Or, are you referring to that as well?

Yes, I mean the same. The `memberRefInvalidationStrategy` uses `memberRef` relation and it's important it uses that relation and not `direct`. I'd like to keep the two implementation (where it's relevant) as separate as possible. I know it's already confusing to have two variant of the algorithm living in parallel so at least I want to be very careful with names.

> For MacroDefChange, MacroUseChange then?

I like MacroUseChange. It conveys what I wanted to convey really well. I'll rename it accordingly.

gkossakowski:
> I like MacroUseChange. It conveys what I wanted to convey really well. I'll rename it accordingly.

Sorry, I was too fast about it. I forgot that the api change is caused by macro definition and not use. Other files (dependent files) are being invalidated due to macro use (or dependency on anything else from the file where macro is defined). Therefore, I don't think it's a good name to describe what we want to describe. Especially because it's not unambiguous.

I'm still open to suggestion of a better name but I still can't think of anything better myself.

gkossakowski:
@harrah: I just pushed a bunch of commit which address most of comments.

When I worked on this PR I went a bit too far with long, descriptive names. The reason was that I really wanted to be careful at naming things correctly due to inherent complexity of the logic we are dealing with here. I didn't want ambiguous or confusing names to obfuscate the algorithm itself. I agree I went a little bit too far with this :)

Let me know if this look good to you. If it does, I'll squash commits and we should be ready to go. I'm planning to keep 115feb5 separate because it fixes something that was introduced before this PR.

harrah:
I think you might want to change `MemberReferenceInvalidator` to `MemberRefInvalidator`, since `MemberRef` is a consistently used abbreviation elsewhere (and of course shorter).  I understand your point with `Macro[Def|Use]Change`.  I think the much shorter name is worth the small loss in accuracy.  You can probably at least drop the leading `API` and shorten `MacroDefinition` to `MacroDef`.

Anyway, it looks fine, merge whenever you are happy.

gkossakowski:
Ok. For the record, I also spent some time comparing performance of the clean builds between new and old algorithm. Depending on the project, there's between 2% and 10% performance hit when name hashing is enabled. I observed 2% performance hit when compiling some Play projects. The 10% performance hit (the largest I could find) was observed when compiling Scala compiler. Almost all additional time is spent calculating name hashes. The time spent in additional phase that extracts dependencies was negligible.

Given the fact that incremental compiler already incurs non-trivial overhead on clean builds we probably want to look into this soon. However, I don't think this should be a blocker for merging this PR. I'll squash commits and go ahead with merging.

harrah:
Agree.  Not a blocker since it is not on by default.

gkossakowski:
Ok. Merging! Thanks a lot for all feedback and help!

PR #1043, opened 2013-12-17
Make analysis file format portable.

benjyw:
binary-serialized object.

Also fixes some indentation nits (why, oh why, does SBT code use tabs?)

### DIFF
--- a/compile/inc/src/main/scala/sbt/CompileSetup.scala
+++ b/compile/inc/src/main/scala/sbt/CompileSetup.scala
@@ -27,7 +27,7 @@ object CompileSetup
 		def equiv(a: File, b: File) = a.getAbsoluteFile == b.getAbsoluteFile
 	}
 	implicit val equivOutput: Equiv[APIOutput] = new Equiv[APIOutput] {
-		implicit val outputGroupsOrdering = Ordering.by({ og: MultipleOutput.OutputGroup => og.sourceDirectory })
+		implicit val outputGroupsOrdering = Ordering.by((og: MultipleOutput.OutputGroup) => og.sourceDirectory)
 		def equiv(out1: APIOutput, out2: APIOutput) = (out1, out2) match {
 			case (m1: MultipleOutput, m2: MultipleOutput) =>
 				(m1.outputGroups.length == m2.outputGroups.length) &&
--- a/compile/inc/src/main/scala/sbt/CompileSetup.scala
+++ b/compile/inc/src/main/scala/sbt/CompileSetup.scala
@@ -27,12 +27,14 @@ object CompileSetup
 		def equiv(a: File, b: File) = a.getAbsoluteFile == b.getAbsoluteFile
 	}
 	implicit val equivOutput: Equiv[APIOutput] = new Equiv[APIOutput] {
+		implicit val outputGroupsOrdering = Ordering.by({ og: MultipleOutput.OutputGroup => og.sourceDirectory })
 		def equiv(out1: APIOutput, out2: APIOutput) = (out1, out2) match {
 			case (m1: MultipleOutput, m2: MultipleOutput) =>
-				m1.outputGroups zip (m2.outputGroups) forall {
+				(m1.outputGroups.length == m2.outputGroups.length) &&
+				(m1.outputGroups.sorted zip m2.outputGroups.sorted forall {
 					case (a,b) => 
 						equivFile.equiv(a.sourceDirectory, b.sourceDirectory) && equivFile.equiv(a.outputDirectory, b.outputDirectory)
-				}
+				})
 			case (s1: SingleOutput, s2: SingleOutput) => equivFile.equiv(s1.outputDirectory, s2.outputDirectory)
 			case _ => false
 		}
--- a/compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
+++ b/compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
@@ -354,13 +354,13 @@ object TextAnalysisFormat {
 			val output = outputDirMode match {
 				case Some(s) => s match {
 					case `singleOutputMode` => new SingleOutput {
-						def outputDirectory = outputAsMap(singleOutputKey)
+						val outputDirectory = outputAsMap(singleOutputKey)
 					}
 					case `multipleOutputMode` => new MultipleOutput {
 						def outputGroups = outputAsMap.toArray.map {
 							case (src: File, out: File) => new MultipleOutput.OutputGroup {
-								def sourceDirectory = src
-								def outputDirectory = out
+								val sourceDirectory = src
+								val outputDirectory = out
 							}
 						}
 					}
--- a/compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
+++ b/compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
@@ -4,6 +4,7 @@ package inc
 import java.io._
 import sbt.{CompileSetup, Relation}
 import xsbti.api.{Compilation, Source}
+import xsbti.compile.{MultipleOutput, SingleOutput}
 import javax.xml.bind.DatatypeConverter
 
 
@@ -55,7 +56,7 @@ object TextAnalysisFormat {
 	implicit val compilationF = xsbt.api.CompilationFormat
 
 	def write(out: Writer, analysis: Analysis, setup: CompileSetup) {
-    VersionF.write(out)
+		VersionF.write(out)
 		// We start with relations because that's the part of greatest interest to external readers,
 		// who can abort reading early once they're read them.
 		FormatTimer.time("write relations") { RelationsF.write(out, analysis.relations) }
@@ -68,7 +69,7 @@ object TextAnalysisFormat {
 	}
 
 	def read(in: BufferedReader): (Analysis, CompileSetup) = {
-    VersionF.read(in)
+		VersionF.read(in)
 		val relations = FormatTimer.time("read relations") { RelationsF.read(in) }
 		val stamps = FormatTimer.time("read stamps") { StampsF.read(in) }
 		val apis = FormatTimer.time("read apis") { APIsF.read(in) }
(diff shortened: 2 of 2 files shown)


harrah:
Review by @gkossakowski (not sure if you get notified when I assign to you).

benjyw:
Ping?

gkossakowski:
Hi @benjyw. Sorry for the delay. I'm offline at the moment but I'll review this by Friday.

compile/inc/src/main/scala/sbt/CompileSetup.scala
```diff
@@ -27,12 +27,14 @@ object CompileSetup
 		def equiv(a: File, b: File) = a.getAbsoluteFile == b.getAbsoluteFile
 	}
 	implicit val equivOutput: Equiv[APIOutput] = new Equiv[APIOutput] {
+		implicit val outputGroupsOrdering = Ordering.by({ og: MultipleOutput.OutputGroup => og.sourceDirectory })
```
gkossakowski: I noticed that you wrap all lambdas in curly braces. Is there any reason for that? I haven't seen a styling guide recommending this technique, AFAIR.

Also, in this case you wrap in both parentheses and curly braces which is not necessary.

compile/inc/src/main/scala/sbt/CompileSetup.scala
```diff
@@ -27,12 +27,14 @@ object CompileSetup
...
 		def equiv(out1: APIOutput, out2: APIOutput) = (out1, out2) match {
 			case (m1: MultipleOutput, m2: MultipleOutput) =>
-				m1.outputGroups zip (m2.outputGroups) forall {
+				(m1.outputGroups.length == m2.outputGroups.length) &&
+				(m1.outputGroups.sorted zip m2.outputGroups.sorted forall {
```
gkossakowski, line 34: The new code has a slightly different semantic. Is this intentional change? See:

```
scala> List(1,2,3,4) zip List('a,'b,'c)
res1: List[(Int, Symbol)] = List((1,'a), (2,'b), (3,'c))
```

Note that zip returns non-empty result even if Lists are of different length.

compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
```diff
@@ -298,31 +299,73 @@ object TextAnalysisFormat {
...
+			val outputDir = "output directories"
+			val compileOptions = "compile options"
+			val javacOptions = "javac options"
+			val compilerVersion = "compiler version"
+			val compileOrder = "compile order"
+		}
+
+		// Dummy src file for when there's only a single output, so we can
+		// uniformly represent both SingleOutput and MultipleOutput as maps.
+		private[this] val singleOutput = new File("singleOutput")
```
gkossakowski: Is this really safe? Usually magic values cause troubles. What if user has source folder called "singleOutput"? You'll have `MultipleOutput` with one of the entries set to "singleOutput". The logic a few lines below will be broken by that, no?

compile/inc/src/main/scala/sbt/CompileSetup.scala
benjyw: Usually that's because the IntelliJ Scala plugin fails to grok it otherwise. In this case it looks like the curlies are actually necessary:

[error] /Users/benjy/src/sbt/compile/inc/src/main/scala/sbt/CompileSetup.scala:30: not found: value og
[error]         implicit val outputGroupsOrdering = Ordering.by(og: MultipleOutput.OutputGroup => og.sourceDirectory)
[error]                                                                                           ^

I can remove the parens if you think it's important. I like them for clarity.

compile/inc/src/main/scala/sbt/CompileSetup.scala
benjyw, line 34: I changed the semantics deliberately. Previously (A, B) and (A, B, C) would have been considered equivalent, which seems wrong no?

compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
benjyw: Yes, this is susceptible to the problem you mention. I could change that string to something that is not a valid filename, such as ":single source:". I had originally tried to do this in a more principled way, but it made the code quite laborious.

benjyw: However I've pushed a change onto this branch that makes this more principled.

compile/inc/src/main/scala/sbt/CompileSetup.scala
harrah: The underlying problem is probably that a parameter with an explicit type should be in parentheses.  Perhaps you can get away without them in some situations, but I usually see `{}` for making a function body a block or for pattern matching anonymous functions.

compile/inc/src/main/scala/sbt/CompileSetup.scala
gkossakowski, line 34: Makes sense. I just asked because it wasn't mentioned in the commit message.

compile/inc/src/main/scala/sbt/CompileSetup.scala
gkossakowski: Indeed, in this case you need to provide the parameter type of the lambda because it can't be inferred. In such case you need to use `()` to delimit the parameter list. Your trick with using `{}` is neat but it does not scale to lambdas with multiple parameters:

```
scala> val f = { x: Int => x + 1 }
f: Int => Int = <function1>

scala> val g = { x: Int, y: Int => x + y }
<console>:1: error: ';' expected but ',' found.
       val g = { x: Int, y: Int => x + y }
```

Actually, I'm surprised this trick with using `{}` around anonymous function with explicit parameter type works at all. I couldn't find relevant excerpt in Scala spec which allows it.

You mentioned that you like the for clarity. What kind of clarity do you have in mind? As @harrah mentioned, you normally use `{}` for delimiting a block or pattern matching anonymous function.

harrah: Yep, it's surprising.  Here, `{}` defines a block and the syntax for the result expression of a block allows omitting the parentheses for a single argument as a special case.  See the `Block` and `ResultExpr` productions in the spec.

gkossakowski: Oh, makes sense, now! Normally, this is being used with by-name parameters and some larger pieces of code passed as an argument, right?

benjyw: This is all beyond me, I just added parens and curlies until things worked... I have no personal preference, as long as it compiles and IntelliJ groks it.

gkossakowski: Sorry to hang up on this little syntactic issue. In the case above you can just do:

```
Ordering.by((og: MultipleOutput.OutputGroup) => og.sourceDirectory)
```

For other cases in this file you should be able to just replace `{}` with `()`.

compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
```diff
@@ -298,31 +299,79 @@ object TextAnalysisFormat {
...
+			val compilerVersion = readSeq(in)(Headers.compilerVersion, identity[String]).head
+			val compileOrder = readSeq(in)(Headers.compileOrder, identity[String]).head
+
+			val output = outputDirMode match {
+				case Some(s) => s match {
+					case `singleOutputMode` => new SingleOutput {
+						def outputDirectory = outputAsMap(singleOutputKey)
+					}
+					case `multipleOutputMode` => new MultipleOutput {
+						def outputGroups = outputAsMap.toArray.map {
```
gkossakowski, line 360: Wouldn't it make sense to declare this as `lazy val` (or just val) so mapping is done only once?

compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
```diff
@@ -298,31 +299,79 @@ object TextAnalysisFormat {
...
+				case m: MultipleOutput => (multipleOutputMode, (m.outputGroups map { x => x.sourceDirectory -> x.outputDirectory }).toMap)
+			}
 
-		val stringToSetup = ObjectStringifier.stringToObj[CompileSetup] _
-		val setupToString = ObjectStringifier.objToString[CompileSetup] _
+			writeSeq(out)(Headers.outputMode, mode :: Nil, identity[String])
+			writeMap(out)(Headers.outputDir, outputAsMap, { f: File => f.getPath })
+			writeSeq(out)(Headers.compileOptions, setup.options.options, identity[String])
+			writeSeq(out)(Headers.javacOptions, setup.options.javacOptions, identity[String])
+			writeSeq(out)(Headers.compilerVersion, setup.compilerVersion :: Nil, identity[String])
```
gkossakowski, line 341: Not sure if creating a Seq (a List) with single element in order to write is the best way to do it but it's fine with me. In the end, we don't intend Analysis to be a format widely used.

gkossakowski:
@benjyw: please remove curly braces around lambdas and check whether it makes sense to change `def` to `val` as pointed out above. Other than that this is good to go.

Oh and I share your sentiment about tabs for indentation. Sbt is the only code base I work with that uses tabs and causes problems from time to time. In ideal world, IDEs would be just fine with it but in practices they are not...

compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
benjyw, line 341: It's a bit wonky, but it enables me to not have a special case for scalars, and this turns out to be very helpful in my python code that parses and manipulates this file. It allows that code to just treat everything as a map, since it mostly doesn't care about the meaning of any of it.

compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
benjyw, line 360: Yes, that would be nicer. Changed to val, and also below in the multiple case. That also had the odd side-effect of making IntelliJ not choke on the anonymous class definition, so win-win!

benjyw:
Addressed the comments and pushed another commit to the pull request. PTAL. Thanks!

compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
gkossakowski, line 341: Makes sense. It's interesting to see how idioms from the other language (Python) creep in here.

gkossakowski:
@benjyw: the latest iteration of this change looks good to me. Please squash all commits into one and we should be ready to go with this.

benjyw:
Closed in favor of a squashed version at https://github.com/sbt/sbt/pull/1068 .

benjyw:
Done: https://github.com/sbt/sbt/pull/1068

PR #1050, opened 2013-12-23
Unify warning to other schemes and use string interpolation

jaceklaskowski:


harrah:
Thanks!  Merged in a5b8dffc74b46ef21555ce71514b0c76d83bf8ce.

PR #1060, opened 2014-01-04
Follow deprecation - @deprecated says to use pair and so it's used in the body

jaceklaskowski:


util/io/src/main/scala/sbt/Path.scala
```diff
@@ -118,7 +118,7 @@ sealed abstract class PathFinder
 	final def \ (literal: String): PathFinder = this / literal
 
 	@deprecated("Use pair.", "0.13.1")
-	def x_![T](mapper: File => Option[T]): Traversable[(File,T)] = x(mapper, false)
+	def x_![T](mapper: File => Option[T]): Traversable[(File,T)] = pair(mapper, false)
```
jsuereth, line 121: SO, since the `pair` method is still calling `x`, would you mind (while you're cleaning up) also moving the implementation of the `x` method to `pair` and having the `x` method also delegate to `pair` ?

Thanks.

jsuereth:
Closing due to inactivity.

PR #1064, opened 2014-01-06
Javadoc changes

jaceklaskowski:


launch/interface/src/main/java/xsbti/MainResult.java
```diff
@@ -1,8 +1,12 @@
...
-* has competed.  This interface should be treated as 'sealed', with Exit and Reboot the only
-* direct subtypes.
-*/
+/**
+ * A launched application should return an instance of this from its 'run' method
+ * to communicate to the launcher what should be done now that the application
+ * has completed.  This interface should be treated as 'sealed', with Exit and Reboot the only
+ * direct subtypes.
+ *
+ * @see xsbti.Exit
```
jsuereth, line 9: Hmm, I thought Javadoc automatically generated "known subclasses" for this....

jsuereth:
Would like to wait on this PR until after the "big" launcher PR is merged, just to avoid conflicts.

Also, sbt has never defined a style preference for Javadoc comments. Currently 99% of them are:

```
/** Comments start on the first line
 * and havecontinue down here.
 */
```

My Java style is rusty. is the format you're using the general convention now?

launch/interface/src/main/java/xsbti/MainResult.java
jaceklaskowski, line 9: I can see it in http://www.scala-sbt.org/release/api/index.html#xsbti.MainResult, but thought javadoc wouldn't handle it this way. I may have been mistaken.

jaceklaskowski:
As to waiting - no worries, I/it can wait.

As for the javadoc, I have never seen the beginning of a javadoc in the same line as the opening tag, but it doesn't have to mean I'm correct or follow the general convention. See http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/1a3de3cdc684/make/data/swingbeaninfo/javax/swing/SwingBeanInfoBase.java for a sample.

jsuereth:
Ah, seems like it is a convention then.

launch/interface/src/main/java/xsbti/MainResult.java
jsuereth, line 9: Yeah, we only generate scaladoc for these though.  Think you may want to additonally `@see` the reboot exception in xsbti here too.

jsuereth:
Cool, sorry for the long delay.  One comment and then I think these improvements are ok to go in. Thanks for helping with our docs!

PR #1065, opened 2014-01-06
Use string interpolation and replace deprecated methods

jaceklaskowski:


main/src/main/scala/sbt/CommandStrings.scala
```diff
@@ -38,7 +38,7 @@ s"""$multiTaskSyntax
 	def multiTaskBrief = """Executes all of the specified tasks concurrently."""
 
 
-	def showHelp = Help(ShowCommand, (ShowCommand + " <key>", actBrief), actDetailed)
+	def showHelp = Help(ShowCommand, (s"$ShowCommand <key>", showBrief), showDetailed)
```
jsuereth, line 41: Why the switch to show over act?   Was this a bug?

PR #1068, opened 2014-01-07
Make analysis file portable.

benjyw:
binary-serialized object.

A squashed version of https://github.com/sbt/sbt/pull/1043

### DIFF
--- a/compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
+++ b/compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
@@ -4,6 +4,7 @@ package inc
 import java.io._
 import sbt.{CompileSetup, Relation}
 import xsbti.api.{Compilation, Source}
+import xsbti.compile.{MultipleOutput, SingleOutput}
 import javax.xml.bind.DatatypeConverter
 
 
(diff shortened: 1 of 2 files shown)


compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
```diff
@@ -298,31 +299,79 @@ object TextAnalysisFormat {
...
+			val compilerVersion = readSeq(in)(Headers.compilerVersion, identity[String]).head
+			val compileOrder = readSeq(in)(Headers.compileOrder, identity[String]).head
+
+			val output = outputDirMode match {
+				case Some(s) => s match {
+					case `singleOutputMode` => new SingleOutput {
+						val outputDirectory = outputAsMap(singleOutputKey)
+					}
+					case `multipleOutputMode` => new MultipleOutput {
+						def outputGroups = outputAsMap.toArray.map {
```
gkossakowski, line 360: Sorry, I was too fast say this is fine. I missed the fact that you forgot to change `def outputGroups` to `val outputGroups` so `outputAsMap.toArray.map` logic is not rerun on every invocation.

benjyw:
Closed in favor of https://github.com/sbt/sbt/pull/1069, which fixes a nit.

PR #1069, opened 2014-01-07
Make analysis file portable.

benjyw:
binary-serialized object.

### DIFF
--- a/compile/inc/src/main/scala/sbt/CompileSetup.scala
+++ b/compile/inc/src/main/scala/sbt/CompileSetup.scala
@@ -27,12 +27,14 @@ object CompileSetup
 		def equiv(a: File, b: File) = a.getAbsoluteFile == b.getAbsoluteFile
 	}
 	implicit val equivOutput: Equiv[APIOutput] = new Equiv[APIOutput] {
+		implicit val outputGroupsOrdering = Ordering.by((og: MultipleOutput.OutputGroup) => og.sourceDirectory)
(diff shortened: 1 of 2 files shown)


gkossakowski:
LGTM.

Thanks!

jsuereth:
Is there something issing from here?  I'm getting this now when compiling:

```
[error] /home/jsuereth/projects/sbt/sbt/compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala:360: overriding method outputGroups in trait MultipleOutput of type ()Array[xsbti.compile.MultipleOutput.OutputGroup];
[error]  value outputGroups has incompatible type
[error]                         val outputGroups = outputAsMap.toArray.map {
[error]                             ^
```

benjyw:
taking a look

jsuereth:
If you add the type annotation on that line, it succeeds.   Seems like an odd "inferred something more specific" issue where Scala's bytecode would use Object because the type is a crazy `AnyRef with XYZ with { random methods}`.   I've started noticing this issue in Scala 2.10.3 but haven't been able to isolate it.    (@retronym does this sound familiar?  I'm not in tune with the ticket DB).

benjyw:
Can you fix directly, so it doesn't have to go through a pull request from
me? Or would you rather have the pull request?

jsuereth:
Doesn't matter. I think I'll issue a quick fix to unblock everyone.

jsuereth:
done.

PR #1076, opened 2014-01-13
Bring back and deprecate `Incremental.incDebugProp`.

gkossakowski:
The ae15eccd9c7aea2b4336ea454d974aed66d5ec16 accidentally removed
`Incremental.incDebugProp` which broke Scala IDE build that relies on it.

We bring back that val but at the same time we deprecate it because we
have better mechanism for configuring incremental compiler now.

I also added a little comment with the history of `incDebugProp` which
explains proper migration path.


jsuereth:
This looks fine to me.  Feel free to merge at your convenience! Thanks for the fast fix :)

gkossakowski:
Thank for the quick review!

PR #1082, opened 2014-01-14
Fixes to the internal API hooks for the sbt server.

jsuereth:
- Alter the TaskProgress listener key to be `State => TaskProgress` so it
  can be instantiated from the current server/sbt state.
- Expose the xsbti.Reporter interface for compilation through to sbt builds.


harrah:
LGTM

PR #1099, opened 2014-01-27
Add link to plugin xitrum-package (collects dependency .jar files for standalone Scala programs)

ngocdaothanh:
I also move the section "Release plugins" to below the section "One jar plugins", because they are very related.


jsuereth:
Looks good. Thanks for the fixes!

PR #1101, opened 2014-01-28
Avoid compiler crash for naked `key.value` calls.

### LINKED ISSUE #1031: Compiler crash when .value, etc... used outside of task macro
This was caused by the removal of `resetLocalAttrs`.  When `.value, etc...` are used properly, the task/setting macros remove them as part of macro expansion.  When they are not removed, they survive beyond typers.  Therefore, problems with untyped trees from the `.value`-like macros are only exposed when used in the wrong context.

retronym:
Untyped trees underneath typed trees makes Jack and sad boy.
And they make superaccessors a sad phase.

The recent refactoring to retain original types in the trees
representing the argument to the task macro meant that the `value`
macro also was changed to try to avoid this untyped-under-typed
problem. However, it didn't go deep enough, and left the child
trees of the placeholder tree `InputWrapper.wrap[T](key)` untyped.

This commit uses `c.typeCheck` to locally typeheck that tree fully
instead.

Fixes #1031 

Review by @harrah

### DIFF
--- a/main/settings/src/main/scala/sbt/std/InputWrapper.scala
+++ b/main/settings/src/main/scala/sbt/std/InputWrapper.scala
@@ -75,8 +75,16 @@ object InputWrapper
 		sel.setPos(pos) // need to set the position on Select, because that is where the compileTimeOnly check looks
 		val tree = ApplyTree(TypeApply(sel, TypeTree(tpe) :: Nil), ts.tree :: Nil)
 		tree.setPos(ts.tree.pos)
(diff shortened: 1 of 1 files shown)


jsuereth:
LGTM

jsuereth:
Thanks much for fixing this @retronym ! Really good to see someone with expertise on this.

retronym:
Just pushed without the accidentally-added changes.

harrah:
LGTM

PR #1105, opened 2014-02-06
Fix formatting of inline code samples which are pluralized

suzannehamilton:
The docs contain some inline code samples whose formatting is broken, for example on the [Build Definition page](http://www.scala-sbt.org/release/docs/Getting-Started/Basic-Def.html):

screen shot 2014-02-06 at 01 11 25

I'm pretty sure this is because [Sphinx requires inline markup to be surrounded by spaces](http://sphinx-doc.org/rest.html?highlight=backquote#inline-markup), which can be worked around by adding an escaped space. In the example above, the plural terms like "`class`es" don't have this escaped space.

I used `\ `[^`]+`[A-Za-z]` to find any other occurrences and fixed those as well.


harrah:
Thanks!  Merged in 4ab8074753695665300b692a4fee26635b1ed118 and 76df88646445121b4f8dfb1933e163fb65eb68fc.

PR #1112, opened 2014-02-12
Multi VM Tests (without Jumi capture)

backuitist:
Allow execution of tests in multiple forked VM. Within a VM tests are executed sequentially, making possible output capture while having multiple tests run concurrently. This works well together with testHideSuccessfulOutput.
To activate this feature, forking must be activated.
The number of forked VM is set by testNumberForkedJvm. When set to more than one, intra-vm parallel execution is disabled.

This is being discussed in

```
https://groups.google.com/forum/#!topic/sbt-dev/E7gPr8iu9b4
https://groups.google.com/d/msg/sbt-dev/w7uli_-gQe8/FyhFYPhLX7sJ
```


backuitist:
Actually we discussed it in https://github.com/sbt/sbt/pull/967.

This pull request originated in the need for segregated test output. 

Currently SBT is unable to run tests in parallel while providing a meaningful trace: test outputs are interleaved. This is especially annoying when a test pass on your machine but breaks on a continuous integration server.

On a continuous integration server you can lookup the tests output in 2 ways, either your crawl through the standard output or you tell the server which test produced what.
SBT provides a facility for introspecting test reports in the form of `TestReportListener`. As it stands right now this facility is lacking (no stdout) but worse, difficult to use (if you take a look at the existing JUnit XML report listener it has to resort to ugly hacks). In my opinion this poor state of affair is due to a missing reference implementation.
Quick question: how will you test the `TestReportListener` without this reference implementation?

JUnit XML is the defacto standard to provide test reports to the continuous integration server. Having it as part of this PR is not as off topic as it seems.
I know that SBT modular design dictates to bundle this as a plugin but there's a limit to this approach. How bad is it to bundle 100 lines of code into SBT as opposed to require every user that uses a continuous integration server to lookup the plugin list and download it as a separate jar?
Lastly, JUnit XML is not going to change and even if it did it wouldn't break backward compatibility.

If I haven't convinced you, then I'm happy that you take a look at it, and rewrite the test that checks the segregation of test outputs.

jsuereth:
I don't have an issue dumping test output to a standard format.  If JUnit XML is that standard, that's ok too for Defaults.scala.   What I'd like, though, is for a separate PR (that this one would depend on) where we can evaluate/merge JUnit test report separately.  While they are dependent issues, they don't have to be treated as one.  We could do the JUnitReport XML evaluation/merge a lot quicker if it was in isolation.  The parallelization will require a bit more work to look into.

backuitist:
The JUnit XML report comes with a refactoring of the TestReportListener. If you want a separate PR for the JUnit XML then it means removing all the parallelization code from this PR while retaining the TestReportListener refactoring. While this is surely doable it will take some time, in my opinion more time than what you'll save by having 2 PR instead of one. You might want to take a closer look at the PR.

backuitist:
I've rebased and squashed the commits into https://github.com/sbt/sbt/pull/1119

PR #1113, opened 2014-02-12
Scala 2.11 compatibility

retronym:
Review by @gkossakowski

PR #1114, opened 2014-02-12
Remove work-arounds for Scala 2.11 problematic dependencies

gkossakowski:
The ff0fd6eec658502f276f89c46f4aba0e0e268ddc introduced some exclusions
that were necessary for getting sbt to resolve dependencies properly
against Scala 2.11.0-M7.

Scala 2.11.0-M8 fixed its dependency structure so we can get rid of those
exclusions now.

### DIFF
--- a/project/Sbt.scala
+++ b/project/Sbt.scala
@@ -276,7 +276,7 @@ object Sbt extends Build
 		artifact in (Compile, packageSrc) := Artifact(srcID).copy(configurations = Compile :: Nil).extra("e:component" -> srcID)
 	)
 	def compilerSettings = Seq(
-		libraryDependencies <+= scalaVersion( "org.scala-lang" % "scala-compiler" % _ % "test" excludeAll(ExclusionRule(organization = "org.scala-lang.modules"))),
+		libraryDependencies <+= scalaVersion("org.scala-lang" % "scala-compiler" % _ % "test"),
 		unmanagedJars in Test <<= (packageSrc in compileInterfaceSub in Compile).map(x => Seq(x).classpath)
 	)
 	def precompiled(scalav: String): Project = baseProject(compilePath / "interface", "Precompiled " + scalav.replace('.', '_')) dependsOn(interfaceSub) settings(precompiledSettings : _*) settings(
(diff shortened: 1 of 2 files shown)


gkossakowski:
Review by @retronym

jsuereth:
LGTM - was this tested with the dbuild scripts?

gkossakowski:
I tested it by building the whole sbt myself. See #1113.

I'm happy to run a dbuild job as well. Can you give me a pointer which one specifically I should run?

jsuereth:
This one: https://github.com/typesafehub/sbt-builds-for-ide/blob/wip/cleanup-build-mechanism/sbt-on-2.11.x

retronym:
That build doesn't compile all of SBT:

```
          projects: ["compiler-interface",
                     "classpath","logging","io","control","classfile",
                     "process","relation","interface","persist","api",
                     "compiler-integration","incremental-compiler","compile","launcher-interface"
```

Can we crank up the coverage?

retronym:
LGTM. I used this patch locally earlier today.

jsuereth:
Ah, right.  We used to not build all of it because of 2.11 incompatibilities with macros.  We can crank up the coverage once we figure out why it's not working inside the pr-validation automation.

gkossakowski:
@jsuereth: as noted in #1113 I've built entire sbt on top of Scala 2.11.0-M8 locally so I think it should work now.

jsuereth:
Just because it builds locally doesn't mean it won't break all the crazy integration scripts we have between projects.  particularly anything that alters dependencies _may_ cause issues in the dbuild script.  I'd prefer if we don't merge these types of changes until we can validate that the IDE integration still works, thanks.

retronym:
Sorry. That stuff is only a problem if you're using M7 or earlier.

Incidentally, I'm working on dbuild at the moment, resurrecting the Community 2.11 build.

jsuereth:
Cool.  I just want to make sure it doesn't break our existing dbuild scripts though.  Our version.properties is still on M7 for our nightlies, and my PR is still hanging in the air to update that.

PR #1115, opened 2014-02-12
Added basic documentation for DefinableSetting, inferring behavior from implementation.

metasim:
Method doc accuracy should be checked, as some behavior was inferred.

Is there an SBT-specific ScalaDoc syntax for linking to the same version of the published Sphinx generated docs? It would be nice to be able to cross-link back to that documentation when a broader exposition is available.

### DIFF
--- a/main/settings/src/main/scala/sbt/Structure.scala
+++ b/main/settings/src/main/scala/sbt/Structure.scala
@@ -27,7 +27,7 @@ sealed trait ScopedTaskable[T] extends Scoped {
 
 /** Identifies a setting.  It consists of three parts: the scope, the name, and the type of a value associated with this key.
 * The scope is represented by a value of type Scope.
-* The name and the type are represented by a value of type AttributeKey[T].
+* The name and the type are represented by a value of type `AttributeKey[T]`.
 * Instances are constructed using the companion object. */
 sealed abstract class SettingKey[T] extends ScopedTaskable[T] with KeyedInitialize[T] with Scoped.ScopingSetting[SettingKey[T]] with Scoped.DefinableSetting[T]
 {
@@ -52,7 +52,7 @@ sealed abstract class SettingKey[T] extends ScopedTaskable[T] with KeyedInitiali
 
 /** Identifies a task.  It consists of three parts: the scope, the name, and the type of the value computed by a task associated with this key.
 * The scope is represented by a value of type Scope.
-* The name and the type are represented by a value of type AttributeKey[Task[T]].
+* The name and the type are represented by a value of type `AttributeKey[Task[T]]`.
 * Instances are constructed using the companion object. */
 sealed abstract class TaskKey[T] extends ScopedTaskable[T] with KeyedInitialize[Task[T]] with Scoped.ScopingSetting[TaskKey[T]] with Scoped.DefinableTask[T]
 {
@@ -76,7 +76,7 @@ sealed abstract class TaskKey[T] extends ScopedTaskable[T] with KeyedInitialize[
 /** Identifies an input task.  An input task parses input and produces a task to run.
 * It consists of three parts: the scope, the name, and the type of the value produced by an input task associated with this key.
 * The scope is represented by a value of type Scope.
-* The name and the type are represented by a value of type AttributeKey[InputTask[T]].
+* The name and the type are represented by a value of type `AttributeKey[InputTask[T]]`.
 * Instances are constructed using the companion object. */
 sealed trait InputKey[T] extends Scoped with KeyedInitialize[InputTask[T]] with Scoped.ScopingSetting[InputKey[T]] with Scoped.DefinableSetting[InputTask[T]]
 {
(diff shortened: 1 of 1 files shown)


main/settings/src/main/scala/sbt/Structure.scala
```diff
@@ -95,6 +95,21 @@ object Scoped
...
+      * Allows specification of the Scope or part of the [[Scope]] of a setting being referenced. 
+      * @example 
+      *  {{{
+      *  name in Global := "hello Global scope"
+      * 
+      *  name in (Compile, packageBin) := "hello Compile scope packageBin"
+      *  
+      *  name in Compile := "hello Compile scope"
+
+      *  name.in(Compile).:=("hello ugly syntax")
```
jsuereth, line 109: nice

main/settings/src/main/scala/sbt/Structure.scala
```diff
@@ -113,16 +128,25 @@ object Scoped
...
 		def scopedKey: ScopedKey[S]
 
 		private[sbt] final def :==(app: S): Setting[S]  =  macro std.TaskMacro.settingAssignPure[S]
+        /** Binds a single value to this. A new `Setting` is defined using the value(s) of `app`. */
 		final def <<= (app: Initialize[S]): Setting[S]  =  macro std.TaskMacro.settingAssignPosition[S]
+        /** Internally used function for setting a value along with the `.sbt` file location where it is defined. */
 		final def set (app: Initialize[S], source: SourcePosition): Setting[S]  =  setting(scopedKey, app, source)
+        /** Setting accessor with explicit Scope specification. */
 		final def get(settings: Settings[Scope]): Option[S] = settings.get(scopedKey.scope, scopedKey.key)
+        /** Lift this into an Option. */
```
jsuereth: Not quite.  This creates an `Initialize` which is `None` if there was no previous definition of this key and `Some(value)` if there was.  This allows you to use the _existence_ of a setting to define another setting.

Help trying to make that more normal english welcome :)

main/settings/src/main/scala/sbt/Structure.scala
```diff
@@ -113,16 +128,25 @@ object Scoped
...
 		final def ? : Initialize[Option[S]] = Def.optional(scopedKey)(idFun)
+        /** Lift this into an Option, and then call `getOrElse` using `i` as the fallback value. */
```
jsuereth: Uses the value defined at the current key _or_ the value created by the passed in `Initialize[T]` if no value exists at the current key.

main/settings/src/main/scala/sbt/Structure.scala
```diff
@@ -113,16 +128,25 @@ object Scoped
...
 		final def or[T >: S](i: Initialize[T]): Initialize[T] = (this.?, i)(_ getOrElse _ )
+        /** Lift this into an Option, and then call `getOrElse`, evaluating `or` if needed. */
```
jsuereth: Like `?` but work on a lazy value rather than at the `Initialize` level.

I.e. Use the value defined at the current key _or_ the passed in value if no value exists at the current key.

jsuereth:
Thanks at ton for taking a crack at this!  I tried to help comment on the wording as I could.   Let me know if it's human-readable or not.

jsuereth:
Also there is no cross-linking ability currently.  That's a good idea though, something to look into.

metasim:
I'll roll in your corrections and resubmit. Sorry it wasn't closer to correct and more comprehensive, but maybe that's a helpful way of going about it for you.

jsuereth:
@metasim no problems!  This is far easier.  It's a lot harder for me to go back to before I knew what things meant and figure out the key wordings someone who _doesn't_ know needs in order to learn.  When writing books, e.g. I like having someone who knows nothing about the subject as I find I get a lot more valuable feedback on whether I'm teaching well.

Plus, anytime you want to try to improve docs via pull requests, I'm more than happy to review/help! Thanks again.

main/settings/src/main/scala/sbt/Structure.scala
metasim: Regarding this:

>  ...and `Some(value)` if there was.

Where does `value` come from since `?` takes no parameters?

Is it effectively whatever is returned by `get`?

jsuereth: yes.  It'll try to access the previous value defined at the key and give you the value in a `Some` if it exists.

Since these are all late-bound, the setting you create will look at the previous setting definitions.   In sbt all settings have a strict ordering they're evaluated in.  This is your means of checking to see if someone in a higher ordering touched a setting you need to tweak.

metasim:
Hopefully this is a bit better. I expounded a bit, so don't hesitate to nit pick! :-)

jsuereth:
It looks good to me.  Maybe  a quick glance from @harrah or @cunei then good to merge.

metasim:
Gracias!

PR #1119, opened 2014-02-14
Multi VM forked test & test output segregation

backuitist:
Currently SBT is unable to run tests in parallel while providing a meaningful trace:
test outputs are interleaved. This is especially annoying when a test pass on
your machine but breaks on a continuous integration server.

This commit allows running forked tests in multiple VM, enabling test output segregation.
In order to report test outputs a refactoring of `TestReportListener` was necessary,
breaking compatibility with 0.13.
In addition SBT provides now a reference implementation of the new `TestReportListener`,
formatting test reports in JUnit XML.

Those features are disabled by default and are activated with the following settings:
- `testNumberForkedJvm` : controls how many VM are used to run the tests
                          Note: this is reduced if there are fewer tests than vm
                          Note: forking must be enabled (`fork in test := true`)
- `testHideSuccessfulOutput` : when set to true, do not print tests output on the console
                               Note: this does not work in non-forked mode
- `testReportJUnitXml` : when set to true produce reports in the JUnit XML format
                         Note: in non-forked mode test outputs are not captured

Meaningful (debug level) logs are produced for each of these features.


eed3si9n:
Hi Bruno, thanks for the contribution and sorry for the late response. I'm excited to bring this in, but we'd have to take some time to review the changes for bincompat, and possible ask to break up the commits.

eed3si9n:
We didn't have time to review this in 0.13.5, so I'm putting this in a branchhttps://github.com/sbt/sbt/tree/wip/multi-vm-tests, but close this pull req for major code formatting that's about happen.

PR #1121, opened 2014-02-15
migrates to 2.11.0-RC1

xeno-by:
Migrates sbt/sbt to https://github.com/scala/scala/pull/3452 that is planned to be merged into trunk before 2.11.0-RC1. No more casts to Global. No more powerContext. /cc @retronym @harrah

### DIFF
--- a/main/settings/src/main/scala/sbt/std/InputWrapper.scala
+++ b/main/settings/src/main/scala/sbt/std/InputWrapper.scala
@@ -67,6 +67,7 @@ object InputWrapper
 	def wrapImpl[T: c.WeakTypeTag, S <: AnyRef with Singleton](c: Context, s: S, wrapName: String)(ts: c.Expr[Any], pos: c.Position)(implicit it: c.TypeTag[s.type]): c.Expr[T] =
 	{
 			import c.universe.{Apply=>ApplyTree,_}
+			import compat._
 		val util = new ContextUtil[c.type](c)
 		val iw = util.singleton(s)
 		val tpe = c.weakTypeOf[T]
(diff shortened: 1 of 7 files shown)


xeno-by:
Unfortunately I haven't yet been able to test these changes as I've run into troubles described in https://groups.google.com/forum/#!topic/sbt-dev/giUaFvwA_ww. It all compiles, though, that's for sure.

If anyone could assist me with testing, that'd be greatly appreciated. You don't even need to build my Scala branch, just use already published artifacts from http://private-repo.typesafe.com/typesafe/scala-pr-validation-snapshots/org/scala-lang/scala-compiler/2.11.0-2fc0164-SNAPSHOT/. I tried that and failed: https://github.com/scala/async/pull/61#issuecomment-35156526.

jsuereth:
I think this has to be targeted at the 0.14 (soon to be 1.0) branch.   sbt 0.13 is targeted at scala 2.10.x so if this breaks using 2.10.x, we'll have to keep it in a different branch.

retronym:
Yep, that makes sense.

The idea is to see if the proposed extensions to the "hard-core" parts of the macro APIs will allow SBT and async to move away from using non-binary-compatible parts of the compiler, which both do at the moment by casting.

jsuereth:
@retronym Cool, thanks for the response.  I'll leave this open for now until we can move the 0.14 branch to be 1.0, then we can merge into there.

Question though:  For 2.11 incremental compiler would this put us on a branched tree or will 0.13 still work against 2.11?

retronym:
Right now, this is just intented as a POC to help us review the proposed API changes to Scala.

Assuming they go through, we'll submit a real PR against SBT. If we could test this patch easily, it provides us with a stronger case. I might be able to do that with dbuild.

I'll close this PR for now to clear it from the queue.

xeno-by:
Thank you!

jsuereth:
I can help dbuilding next week if needed.

Thanks for helping keep us up to date!

PR #1122, opened 2014-02-17
Include value of `nameHashing` flag in `CompileSetup`.

### LINKED ISSUE #1071: assertion failed: When `nameHashing` is disabled `names` relation should be empty: $names
using 0.13.2-M1 with incOptions := incOptions.value.withNameHashing(true)
i get the error on incremental compilation
last doesn't show  stacktrace for some reason

### LINKED ISSUE #1081: Restart of sbt is required after changing the value of the name hashing flag
If you change the value of the name hashing flag then simply running `clean` followed by `compile` is not enough to switch to different implementation of incremental compilation algorithm. It turns out, a full restart of sbt is required because Analysis objects are cached in memory by incremental compiler.

The caching was introduced in 135609e5b0656a05eb9e6d4f14e0ca2165edb81e. The commit message said it was temporary hack but it never got reverted.

We'll need to investigate whether the hack is needed at all and if so, we'll need to clarify caching strategy.

gkossakowski:
The CompileSetup class is being used to detect changes to arguments of
incremental compiler that affect result of compilation and trigger
recompilation. Examples of such arguments include, the target (output)
directory, Scala compiler options, Scala compiler version, etc.

By adding `nameHashing` to CompileSetup we have a chance to handle change
to that flag smoothly by throwing away old Analysis object and starting
with an empty one. That's implemented in AggressiveComile by extending
the logic that was responsible for detection of changes to CompileSetup
values. Thanks to this change we fix #1081.

Analysis formats has been updated to support persisting of newly added
value in CompileSetup. We used to not store the value of `nameHashing`
flag in persisted Analysis file and infer it from contents of relations
but that leads to issue #1071 when empty relations are involved. Given
the fact that CompileSetup stores `nameHashing` value now, we can just
use it when reading relations and fix #1071. This requires reading/writing
compile setup before reading relations. I decided to make that change even
if there's a comment saying that reading/writing relations first was done
intentionally.

### DIFF
--- a/compile/inc/src/main/scala/sbt/CompileSetup.scala
+++ b/compile/inc/src/main/scala/sbt/CompileSetup.scala
@@ -11,7 +11,8 @@ package sbt
 //    because complexity(Equiv[Seq[String]]) > complexity(Equiv[CompileSetup])
 //     (6 > 4)
 final class CompileOptions(val options: Seq[String], val javacOptions: Seq[String])
-final class CompileSetup(val output: APIOutput, val options: CompileOptions, val compilerVersion: String, val order: CompileOrder)
+final class CompileSetup(val output: APIOutput, val options: CompileOptions, val compilerVersion: String,
+	val order: CompileOrder, val nameHashing: Boolean)
 
 object CompileSetup
 {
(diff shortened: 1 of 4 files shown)


gkossakowski:
//cc @benjy that might care about changes to Analysis format

jsuereth:
I'm not 100% on the BC details myself, but this LGTM.

gkossakowski:
Scala IDE doesn't refer to `CompileSetup`: https://github.com/scala-ide/scala-ide/search?q=CompileSetup&type=Code
Zinc does but by just passing it around: https://github.com/typesafehub/zinc/search?q=CompileSetup&source=cc

Therefore none of those two should be affected by the new field. Zinc should stop depending on `CompileSetup` in the future.

gkossakowski:
@jsuereth: let me know if my little searches convince you. If yes, I'd like to merge it.

jsuereth:
@gkossakowski Ah, no. I trust you was just making a note that I'm relying on your knowledge there ;).  Already gave a LGTM, just noting what I couldn't review myself.  Thanks for the search.

PR #1125, opened 2014-02-18
Make change to CompileSetup backwards compatible.

gkossakowski:
The c7f435026ff8591085f427b8ee4d192e1f10f783 introduced a new parameter
to the constructor of `CompileSetup` but it turns out that this class
is being used in zinc. Introduce an overloaded variant of that constructor
that preserves backwards compatibility.


gkossakowski:
Review by @retronym.

gkossakowski:
This is follow up to #1122 that fixes breakage in zinc. I'll try to be more diligent when checking whether downstream projects are affected.

retronym:
LGTM, merging to unbreak our community build.

PR #1126, opened 2014-02-19
explicitly disapprove plugins using default package

jozic:


jsuereth:
Thanks for the fix!  Totally agree.

PR #1127, opened 2014-02-19
using compat._ to plug source compatibility breakages

retronym:
This commit makes the code source compatible across Scala 2.10.3
and https://github.com/scala/scala/pull/3452, which is proposed
for inclusion in Scala 2.11.0-RC1.

We only strictly need the incremental compiler to build on Scala
2.11, as that is integrated into the IDE. But we gain valuable
insight into compiler regressions by building _all_ of SBT with
2.11.

We only got there recently (the 0.13 branch of SBT now fully cross
compiles with 2.10.3 and 2.11.0-SNAPSHOT), and this aims to keep
things that way.

Once 2.10 support is dropped, SBT macros will be able to exploit
the new reflection APIs in 2.11 to avoid the need for casting
to compiler internals, which aren't governed by binary compatibility.
This has been prototyped by @xeno-by: https://github.com/sbt/sbt/pull/1121

Review by @jsuereth

This merge doesn't have to be coordinated with the merge to Scala:
- the zinc/IDE build chain is unaffected.
- we can use the branch in our community build that builds _all_ of SBT

But in any case, it will be nice to restore our community build to the mainline sooner rather than later.

This branch was built in against https://github.com/scala/scala/pull/3452 here: https://jenkins-dbuild.typesafe.com:8499/job/Community-2.11.x-retronym-pr-3452/8/console

### DIFF
--- a/main/settings/src/main/scala/sbt/std/InputWrapper.scala
+++ b/main/settings/src/main/scala/sbt/std/InputWrapper.scala
@@ -59,6 +59,9 @@ object InputWrapper
 	private[std] def wrapPrevious[T: c.WeakTypeTag](c: Context)(ts: c.Expr[Any], pos: c.Position): c.Expr[Option[T]] =
 		wrapImpl[Option[T],InputWrapper.type](c, InputWrapper, WrapPreviousName)(ts, pos)
 
+	// TODO 2.11 Remove this after dropping 2.10.x support.
+	private object HasCompat { val compat = ??? }; import HasCompat._
+
 	/** Wraps an arbitrary Tree in a call to the `<s>.<wrapName>` method of this module for later processing by an enclosing macro.
 	* The resulting Tree is the manually constructed version of:
 	*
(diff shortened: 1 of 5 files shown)


main/settings/src/main/scala/sbt/std/InputWrapper.scala
```diff
@@ -59,6 +59,9 @@ object InputWrapper
 	private[std] def wrapPrevious[T: c.WeakTypeTag](c: Context)(ts: c.Expr[Any], pos: c.Position): c.Expr[Option[T]] =
 		wrapImpl[Option[T],InputWrapper.type](c, InputWrapper, WrapPreviousName)(ts, pos)
 
+	// TODO 2.11 Remove this after dropping 2.10.x support.
+	private object HasCompat { val compat = ??? }; import HasCompat._
```
jsuereth, line 63: So this is shaded by the actual "compat" object in c.universe?

retronym, line 63: Exactly. We're going do document this trick in a migration guide.

jsuereth:
SO, if I  understand the workaround correctly, LGTM.  Feel free to merge at your convenience.

gkossakowski:
LGTM.

I think we should have a snippet in migration docs that explains this technique. It's really non-trivial for regular Scala users to spot what's going on.

xeno-by:
We will.

gkossakowski:
What I meant by that: this link should be included as a comment in every instance compatibility hack people paste into their code so they can easily look it up later. For sbt it doesn't matter much but for other code bases it does.

PR #1137, opened 2014-02-26
update doc: javaHome is not File but Option[File].

gourlaysama:
It's all in the title :-)


jsuereth:
Thanks much for the fix!

PR #1139, opened 2014-02-27
Completed a full example of forked tests with grouping

lucamilanesio:
The previous example of how to fork group of tests in different
JVMs was incomplete and not fully working.

a) The "testGrouping in Test" is needed, otherwise the grouping is
   not known to which phase to apply
b) Added a more complete Project definition to explain what the curly
   brackets section refers to and where the settings need to be included

The new complete example works out of the box :-)

### DIFF
--- a/src/sphinx/Detailed-Topics/Testing.rst
+++ b/src/sphinx/Detailed-Topics/Testing.rst
@@ -204,7 +204,7 @@ available with :key:`testGrouping` key. For example:
       def groupByFirst(tests: Seq[TestDefinition]) =
         tests groupBy (_.name(0)) map {
           case (letter, tests) => new Group(letter.toString, tests, SubProcess(Seq("-Dfirst.letter"+letter)))
-        } toSeq;
+        } toSeq
 
       lazy val root = Project("root", file("."), settings = defaultSettings ++  Seq(
         scalaVersion := "2.10.3",
--- a/src/sphinx/Detailed-Topics/Testing.rst
+++ b/src/sphinx/Detailed-Topics/Testing.rst
@@ -191,26 +191,19 @@ The setting:
 specifies that all tests will be executed in a single external JVM. See
 :doc:`Forking` for configuring standard options for forking. More control
 over how tests are assigned to JVMs and what options to pass to those is
-available with :key:`testGrouping` key. For example:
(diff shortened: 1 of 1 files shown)


jsuereth:
The previous example was supposed to be in `build.sbt`, while this one is in `Build.scala`.  Does the previous one not work in a `build.sbt` file?

lucamilanesio:
mmm ... the example would never work in build.sbt and it actually didn't :-(

In order for the Fork to work in Groups, you have to define your grouping logic in the Build.scala AFAIK and from what I've seen in the sbt test code base.

Is there another way to define Groups in build.sbt ?

jsuereth:
it wouldn't work in sbt 0.12.-, but it should work in sbt 0.13.+.  What sbt are you using?  Our docs should be versioned appropriately....

lucamilanesio:
I am using sbt 1.13.1 (latest and greatest).

I tried to insert into the build.sbt and it does not work :-(

See below the code:
  import Tests._

  libraryDependencies += "org.scalatest" % "scalatest_2.10" % "2.0" % "test"

  fork in Test := true

  {
    def groupByFirst(tests: Seq[TestDefinition]) =
      tests groupBy (_.name(0)) map {
        case (letter, tests) => new Group(letter.toString, tests, SubProcess(Seq("-Dfirst.letter"+letter)))
      } toSeq;
    testGrouping <<= definedTests in Test map groupByFirst
  }

There are three problems:
1) the fork in Test := true is not revelevant and is not used 
2) It would be better to make a complete example so that people can actually "copy&paste" it to their project.
3) The "testGrouping" setting does not specify to which phase applies: does it work for you ? It does not for me unless you include "in Test"

I will amend the pull request to make the example specific to build.sbt and working.

Does it make sense ?

Luca.

lucamilanesio:
Pushed new commit: let me know if that works for you.

P.S. Without the "in Test" after testGrouping, does not work for me. Let me know if you had the same behaviour.

Luca.

jsuereth:
interesting that it needs `<<=` vs `:=`.  Will have to look into that.  Also, thank you very much for the fixes.   Our goal is that the `.sbt` format _be_ the copy-pastable format. It's supposed to have less finicky behaviors than `.scala`, but perhaps that's not the case right now.

Again, thanks for the fixes.  Much appreciated.

PR #1141, opened 2014-02-28
Extra tweak for making the forkGroup work on build.sbt

lucamilanesio:
When the fork grouping definition is on build.sbt, extra rules
apply, such as the line taken as separator between settings.
It seems as well that the ';' to terminate the def function is needed
to isolate it from the setting definition itself.

### DIFF
--- a/src/sphinx/Detailed-Topics/Testing.rst
+++ b/src/sphinx/Detailed-Topics/Testing.rst
@@ -201,9 +201,8 @@ available with :key:`testGrouping` key. For example in build.sbt:
       def groupByFirst(tests: Seq[TestDefinition]) =
         tests groupBy (_.name(0)) map {
           case (letter, tests) => new Group(letter.toString, tests, SubProcess(Seq("-Dfirst.letter"+letter)))
-        } toSeq
-
-        testGrouping in Test <<= groupByFirst( (definedTests in Test).value )
+        } toSeq;
+        testGrouping in Test := groupByFirst( (definedTests in Test).value )
     }
 
 The tests in a single group are run sequentially. Control the number


src/sphinx/Detailed-Topics/Testing.rst
```diff
@@ -201,9 +201,8 @@ available with :key:`testGrouping` key. For example in build.sbt:
       def groupByFirst(tests: Seq[TestDefinition]) =
         tests groupBy (_.name(0)) map {
           case (letter, tests) => new Group(letter.toString, tests, SubProcess(Seq("-Dfirst.letter"+letter)))
-        } toSeq
-
-        testGrouping in Test <<= groupByFirst( (definedTests in Test).value )
+        } toSeq;
```
jsuereth, line 204: Hmm, we should probably just not have postFix notation here....

``` scala

(tests gorupBy (_.name(0)) map {
  case (letter, tests) => new Group(letter.toString, tests, SubProcess(Seq("-Dfirst.letter"+letter)))
}).toSeq
// maybe a comment then here to make sure the line breaks...
```

lucamilanesio:
Why don't we make it even simpler then, so the sample would be straightforward:

``` scala
testGrouping in Test := (definedTests in Test).value map {
  case test => new Group(test.name, Seq(test), SubProcess(Seq("-Dtest.name="+test.name)))
} toSeq
```

In this case we just run one JVM per test suite.

jsuereth:
Right, but removing the postfix operator, so:

``` scala
testGrouping in Test := (definedTests in Test).value.map {
  case test => new Group(test.name, Seq(test), SubProcess(Seq("-Dtest.name="+test.name)))
}.toSeq
```

I agree that's much simpler.  I think the previous format was never fully updated for 0.13 syntax.  What you suggest is a lot better, would love to take that in.

eed3si9n:
I follow Coding in Style talk by @jsuereth: infix `map`, but use dot-notation for `toSeq` with parens:

``` scala
testGrouping in Test := ((definedTests in Test).value map {
  case test => new Group(test.name, Seq(test), SubProcess(Seq("-Dtest.name=" + test.name)))
}).toSeq
```

His original commit comment reflects the style, but not the comment above.

jsuereth:
Just wanted to check status of this.  Are we going to fix the style in this PR or should I merge and fix in follow on?

lucamilanesio:
Yes, no problem ... will change the style in order to be merged.

eed3si9n:
closing this due to inactivity.

PR #1147, opened 2014-03-04
Fixes #429 - Add better error messages when credentials are not found.

### LINKED ISSUE #429: Report authentication failures when resolving dependencies
Currently if a repository requires authentication and it's not in the credentials file SBT simply reports it was not found and gives the "tried" url. There should be a mention of authentication failure in that error.

jsuereth:
- Create hackery to inspect registered credentials in the IvyCredentialStore.
- Create a new authenticator which inserts itself _after_ the ivy authenticator.
  - Will issue an error message detailing host/realm required if credentials are
    not found.
  - Also lists out configured Realms with a 'is misspelled' message.
  - Ignores proxy-related authentication errors, for now.

Fixes #429 

Review by @cunei or @eed3si9n


eed3si9n:
+1 for better error handling.

PR #1151, opened 2014-03-05
Update generatefiles.rst

schmmd:
Wrapped the resource generation with `Def.task`.  Otherwise this example does not compile.


jsuereth:
Thanks much for the fix!

PR #1153, opened 2014-03-06
Override certain SecurityManager methods to avoid filesystem performance hit

harrah:
Further explanation in the comment in the code.  This was discovered while investigating a major performance reduction using sbt-assembly 0.9.2 with sbt 0.12 and 0.13.  When used with sbt 0.13, the `assembly` time was much slower.  This patch will probably speed up other tasks that touch many files, like `clean`.  (To be clear, it will probably restore them to 0.12 speeds.)


jsuereth:
LGTM.  Thanks!

Issue #1156, opened 2014-03-07
publishM2 in sbt 0.13.2-M2 don't overwrite


patriknw:
Changed behaviour compared to 0.13.1.

See [log](https://jenkins.akka.io:8498/job/pr-validator-per-commit/2008/consoleFull) from this [pull request](https://github.com/akka/akka/pull/2060) build

```
java.io.IOException: file copy not done from /localhome/jenkinsakka/workspace/pr-validator-per-commit/akka-persistence/target/akka-persistence-experimental_2.10-2.3-SNAPSHOT.pom to /localhome/jenkinsakka/.m2/repository/com/typesafe/akka/akka-persistence-experimental_2.10/2.3-SNAPSHOT/akka-persistence-experimental_2.10-2.3-SNAPSHOT.pom: destination already exists and overwrite is false
    at org.apache.ivy.plugins.repository.file.FileRepository.copy(FileRepository.java:78)
    at org.apache.ivy.plugins.repository.file.FileRepository.put(FileRepository.java:58)
    at org.apache.ivy.plugins.repository.AbstractRepository.put(AbstractRepository.java:130)
    at org.apache.ivy.plugins.resolver.RepositoryResolver.put(RepositoryResolver.java:234)
    at org.apache.ivy.plugins.resolver.RepositoryResolver.publish(RepositoryResolver.java:216)
    at sbt.IvyActions$$anonfun$publish$3.apply(IvyActions.scala:262)
    at sbt.IvyActions$$anonfun$publish$3.apply(IvyActions.scala:261)
    at scala.collection.TraversableLike$WithFilter$$anonfun$foreach$1.apply(TraversableLike.scala:772)
    at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
    at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:47)
    at scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:771)
    at sbt.IvyActions$.publish(IvyActions.scala:261)
    at sbt.IvyActions$$anonfun$publish$1$$anonfun$apply$1.apply$mcV$sp(IvyActions.scala:97)
    at sbt.IvyActions$$anonfun$publish$1$$anonfun$apply$1.apply(IvyActions.scala:97)
    at sbt.IvyActions$$anonfun$publish$1$$anonfun$apply$1.apply(IvyActions.scala:97)
[...]

PR #1160, opened 2014-03-08
Don't fail publishing on overwrite, but issue a warning.

jsuereth:
Workaround for #1156.
- Creates a new FileRepository that will stil allow local files
  to be transfered if overwrite is true (non-snapshot module),
  but will issue a warning about deprecated behavior.
- Ensure warning is long enough to annoy people into asking
  what it's about.

Review by @dansanduleac (if you have time)


jsuereth:
Build failure appears to be spurious. I'll have to investigate that one.

jsuereth:
So this time the test passed that broke: https://travis-ci.org/sbt/sbt/builds/20360357

Merging this as is.

PR #1162, opened 2014-03-08
Removing deprecated "re-publish release artifacts" behavior from tests.

jsuereth:
In sbt 1.0 republishing released artifacts will be removed, causing these tests
to fail.  We migrate to SNAPSHOT version now, to prevent surprises.

Review by @dansanduleac

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,17 @@
 language: scala
 script:
-  - sbt scripted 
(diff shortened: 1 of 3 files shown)


jsuereth:
Note: This is the first PR we're attempting a TravisCI build for.  We'll see how it goes.

PR #1163, opened 2014-03-08
Record dependencies on macro arguments

Duhemm:
> Macros take arguments as trees and return some other trees; both of them have dependencies but we see trees only after expansion and recorded only those dependencies.

(from [this discussion in sbt-dev](https://groups.google.com/d/msg/sbt-dev/zDOIFDSNRrk/_YTVn7bRJb0J)).

This pull request solves this problem by looking into the attachments of the trees that are supposed to contain originals of macro expansions and recording dependencies of the macro before its expansion.

### DIFF
--- a/compile/interface/src/main/scala/xsbt/Compat.scala
+++ b/compile/interface/src/main/scala/xsbt/Compat.scala
@@ -91,4 +91,42 @@ abstract class Compat
 	private[this] def sourceCompatibilityOnly: Nothing = throw new RuntimeException("For source compatibility only: should not get here.")
 
 	private[this] final implicit def miscCompat(n: AnyRef): MiscCompat = new MiscCompat
+
+	object MacroExpansionOf {
+		def unapply(tree: Tree): Option[Tree] = {
+
+			// MacroExpansionAttachment (MEA) compatibility for 2.8.x and 2.9.x
+			object Compat {
+				class MacroExpansionAttachment(val original: Tree)
+
+				// Trees have no attachments in 2.8.x and 2.9.x
+				implicit def withAttachments(tree: Tree): WithAttachments = new WithAttachments(tree)
+				class WithAttachments(val tree: Tree) {
+					object EmptyAttachments {
+						def all = Set.empty[Any]
+					}
+					val attachments = EmptyAttachments
+				}
+			}
+			import Compat._
+
+			locally {
+				// Wildcard imports are necessary since 2.8.x and 2.9.x don't have `MacroExpansionAttachment` at all
+				import global._ // this is where MEA lives in 2.10.x
+
+				// `original` has been renamed to `expandee` in 2.11.x
+				implicit def withExpandee(att: MacroExpansionAttachment): WithExpandee = new WithExpandee(att)
+				class WithExpandee(att: MacroExpansionAttachment) {
+					def expandee: Tree = att.original
+				}
+
+				locally {
+					import analyzer._ // this is where MEA lives in 2.11.x
+					tree.attachments.all.collect {
+						case att: MacroExpansionAttachment => att.expandee
+					} headOption
+				}
+			}
+		}
+	}
 }
--- a/compile/interface/src/main/scala/xsbt/ExtractUsedNames.scala
+++ b/compile/interface/src/main/scala/xsbt/ExtractUsedNames.scala
@@ -38,7 +38,7 @@ import scala.tools.nsc._
  * The tree walking algorithm walks into TypeTree.original explicitly.
  *
  */
-class ExtractUsedNames[GlobalType <: CallbackGlobal](val global: GlobalType) {
+class ExtractUsedNames[GlobalType <: CallbackGlobal](val global: GlobalType) extends Compat {
 	import global._
 
 	def extract(unit: CompilationUnit): Set[String] = {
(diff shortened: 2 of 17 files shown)


sbt/src/sbt-test/source-dependencies/macro-arg-dep/project/build.scala
```diff
@@ -0,0 +1,29 @@
...
+	val defaultSettings = Seq(
+		libraryDependencies <+= scalaVersion("org.scala-lang" % "scala-reflect" % _ ),
+		incOptions := incOptions.value.withNameHashing(true)
+	)
+
+	lazy val root = Project(
+	   base = file("."),
+	   id = "macro",
+	   aggregate = Seq(macroProvider, macroClient),
+	   settings = Defaults.defaultSettings ++ defaultSettings
```
jsuereth, line 14: Note:  Usage of Defaults.defaultSettings is going to be deprecated in 0.13.2

jsuereth:
So this looks extra super cool!  I'm going to have to delay this until 0.13.3 though. I have way too many issues to resolve for 0.13.2 right now (with auto-plugins), but I'll get back to you next week after the 0.13.2-RC1 is out and 0.13 is ok to merge into again.

Duhemm:
(ping @xeno-by)

compile/interface/src/main/scala/xsbt/Compat.scala
```diff
@@ -91,4 +91,39 @@ abstract class Compat
...
+					object EmptyAttachments {
+						def all = Set.empty[Any]
+					}
+					val attachments = EmptyAttachments
+				}
+			}
+			import Compat._
+
+			locally {
+				import global._ // this is where MEA lives in 2.10.x
```
gkossakowski, line 115: Can we import just MEA from Global? Global has more than 1600 members, we don't need to pull all of them here.

compile/interface/src/main/scala/xsbt/Compat.scala
```diff
@@ -91,4 +91,39 @@ abstract class Compat
...
+				implicit def withExpandee(att: MacroExpansionAttachment): WithExpandee = new WithExpandee(att)
+				class WithExpandee(att: MacroExpansionAttachment) {
+					def expandee: Tree = att.original
+				}
+
+				locally {
+					import analyzer._ // this is where MEA lives in 2.11.x
```
gkossakowski, line 124: Same as above, let's just import MEA, not the whole Analyzer.

compile/interface/src/main/scala/xsbt/Compat.scala
```diff
@@ -91,4 +91,39 @@ abstract class Compat
...
+					}
+					val attachments = EmptyAttachments
+				}
+			}
+			import Compat._
+
+			locally {
+				import global._ // this is where MEA lives in 2.10.x
+				implicit def withExpandee(att: MacroExpansionAttachment): WithExpandee = new WithExpandee(att)
+				class WithExpandee(att: MacroExpansionAttachment) {
```
gkossakowski, line 119: Why do we need that implicit class? Is it because expandee is defined in 2.11, or what?

compile/interface/src/main/scala/xsbt/Compat.scala
Duhemm, line 115: Unfortunately this won't compile with Scala 2.8 and 2.9 since `MacroExpansionAttachment` appeared in 2.10.

compile/interface/src/main/scala/xsbt/Compat.scala
Duhemm, line 119: Exactly, it has been renamed from `original` to `expandee` in 2.11.

xeno-by:
Great job, @Duhemm!

compile/interface/src/main/scala/xsbt/Compat.scala
gkossakowski, line 115: Ah, yes. This is the way to go then. Could you just expand the comment in the code explaining why do we need a wildcard import?

compile/interface/src/main/scala/xsbt/Compat.scala
gkossakowski, line 119: Ok. Let's document it in the code.

gkossakowski:
@jsuereth: don't worry about this PR, I'll take care of it and schedule for sbt 0.13.3 release.

gkossakowski:
Indeed. This looks very promising! I'm glad you managed to figure out the maze of compatibility hacks we have here. :thumbsup:

Have you considered expanding the `macro-arg-dep` test to cover nested macro invocations? I have something like that in mind:

``` scala
package macro

object ClientNested {
    Provider.printTree(Provider.printTree(Foo.str))
}
```

Also, given the fact that you modify both `Dependency` and `ExtractUsedNames`, we should expand their units tests:

https://github.com/sbt/sbt/blob/0.13/compile/interface/src/test/scala/xsbt/DependencySpecification.scala
https://github.com/sbt/sbt/blob/0.13/compile/interface/src/test/scala/xsbt/ExtractUsedNamesSpecification.scala

The code samples you are going to use in those tests should look similar to what you have in the `macro-arg-dep`.

Just in case you it's not clear what's the point of testing (seemingly) the same thing twice: scripted tests are [functional tests](http://en.wikipedia.org/wiki/Functional_testing) and the specifications for `Dependency` and `ExtractUsedNames` are [unit tests](http://en.wikipedia.org/wiki/Unit_testing). Here's a concise description of the difference: http://www.softwaretestingtricks.com/2007/01/unit-testing-versus-functional-tests.html

Duhemm:
Hi guys,

Does anybody know what happened with the last build ? It seems to have timed out, but I don't see how the changes between 5f6b06a and d76b4ea may have caused this ?

I've had to modify `ScalaCompilerForUnitTesting` so that one can compile separately multiple files (this was required in order to compile macro expansions). What do you think ? I will look into the unit tests for `ExtractUsedNames` shortly !

jsuereth:
Yeah, it's transient to you, see the error message:

```
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
```

It's quite possible there's a rare deadlock in the test listener, which would have caused this.  It's the first time I've seen this message, but something I can pay attention to.

compile/interface/src/main/scala/xsbt/Compat.scala
xeno-by, line 124: Just to summarize previous discussion. `MEA` can't be imported here, because that would be incompatible with versions prior to 2.11.

compile/interface/src/main/scala/xsbt/ExtractUsedNames.scala
```diff
@@ -53,30 +53,44 @@ class ExtractUsedNames[GlobalType <: CallbackGlobal](val global: GlobalType) {
...
-			// to types but that might be a bad thing because it might expand aliases eagerly which
-			// not what we need
-			case t: TypeTree if t.original != null =>
-				t.original.foreach(handleTreeNode)
-			case t if t.hasSymbol && eligibleAsUsedName(t.symbol) =>
-				addSymbol(t.symbol)
-			case _ => ()
+
+		def handleTreeNode(node: Tree): Unit = {
+			def handleMacroExpansion(original: Tree): Unit = original.foreach(handleTreeNode)
```
xeno-by, line 58: Can we have a test of a nested macro expansion, i.e. a macro application that expands into another macro application?

sbt/src/sbt-test/source-dependencies/macro-arg-dep/test
```diff
@@ -0,0 +1,15 @@
+> compile
+
+# remove `Foo.str` which is an argument to a macro that incremental compiler doesn't see in
+# Client.scala because macro has been already expanded
+
+$ copy-file macro-client/changes/Foo.scala macro-client/Foo.scala
+
+# we should recompile Foo.scala first and then fail to compile Client.scala due to missing
+# `Foo.str`; however recompilation of Client.scala is never triggered due to missing
+# dependency
```
gkossakowski: We fix this problem in this PR so let's update the comment to say that it does trigger recompilation because dependencies are properly recorded.

gkossakowski:
Ok. I think we just missing test for nested macro applications we'll be done with the code. Once the last patch (for nested macro applications) is submitted we should look into squashing most of the commits. E.g. 14ff4b0 and c9be404 squashed reduce to empty commit as the latter one cancels the former. All small commits that fix typos, reduce use of reflection, etc. should be squashed as well.

Duhemm:
Hi,

I just added this test and updated the comments you mentioned, is everything ok ?

gkossakowski:
Yes, looks good. Could you rebase the branch and squash most of the commits? I think we would like to keep commits for the following events:
- add pending test as in 2bd07bd
- record depedencies on macro arguments as in 8d1e409
- extend unit testing infrastructure in separate commit and then use it, it would mean that you break up 5f6b06a into two (this is optional step if you know how to split commits in Git)
- add tests for nested macros

That would give us 5 commits in total.

Duhemm:
It looks like I managed to split the commits ! Thank you for your help !

gkossakowski:
@jsuereth: since this PR is shaping very nicely I'd like to get it in sbt 0.13.2. It's a very nice piece of work and improves significantly macro handling in name hashing. It's safe because it affects name hashing only. The old incremental compilation algorithm is unaffected. WDYT?

gkossakowski:
@Duhemm: you need to update one more commit :)

gkossakowski:
LGTM

gkossakowski:
@jsuereth: ping on merging this into 0.13.2

xeno-by:
\o/

eed3si9n:
I'm merging this into 0.13.2. @gkossakowski Could you add an entry into https://github.com/sbt/sbt/blob/0.13/src/sphinx/Community/Changes.rst to summarize this enhancement, plz?

PR #1169, opened 2014-03-10
Fix overwrite default to true in deprecated method calls.

jsuereth:
We cannot break existing users, but we can deprecate the improper usage.
This is part #2 of the workaround for #1156.  This ensures that
users will stop using the legacy methods after 0.13.2 is out.

Review by @eed3si9n


jsuereth:
The test passes locally, I'll investigate why it struggles on TravisCI, but I think this change is ok.

PR #1182, opened 2014-03-15
Documentation default paths filtering resources use unmanagedResources

flurdy:
Not entirely sure but does it not make sense that filtering resources should use unmanagedResources and not unmanagedSources?


jsuereth:
Nice catch!   Thanks for the fix.  Note:  Doc fixes need to be targeted at 0.13.1 to be automatically synced.

flurdy:
Yeah sorry, forgot to create a new pull request with the correct target branch for this one as well. Will do now :)

flurdy:
Created https://github.com/sbt/sbt/pull/1235. Closing this one

PR #1189, opened 2014-03-18
Allow end users to add Plugins, not just AutoPlugins

jroper:
This allows plugins to define a Plugins instance that captures both the plugin and its required dependencies.

Also fixed up some scaladocs that were wrong.


jsuereth:
Nice fix.!  I like how it preserves exisitng semantics.

Thanks much.  Once travis reports green, can be merged.

jroper:
Travis is green.

jsuereth:
note: @eed3si9n I'm going to merge this.  We can follow on with cleanups against this patch.

PR #1190, opened 2014-03-18
Allow plugins to be declaritively added in build.sbt

jroper:
This allows you to add:

``` scala
val plugins = Web && Javascript
```

to your `build.sbt`, and these plugins will be auto added to the project associated with the `build.sbt`.

### DIFF
--- a/main/actions/src/main/scala/sbt/compiler/Eval.scala
+++ b/main/actions/src/main/scala/sbt/compiler/Eval.scala
@@ -219,7 +219,7 @@ final class Eval(optionsNoncp: Seq[String], classpath: Seq[File], mkReporter: Se
 		private[this] var vals = List[String]()
 		def getVals(t: Tree): List[String] = { vals = Nil; traverse(t); vals }
 		override def traverse(tree: Tree): Unit = tree match {
-			case ValDef(_, n, actualTpe, _) if isTopLevelModule(tree.symbol.owner) && types.exists(_ <:< actualTpe.tpe) =>
+			case ValDef(_, n, actualTpe, _) if isTopLevelModule(tree.symbol.owner) && types.exists(actualTpe.tpe.<:<) =>
 				vals ::= nme.localToGetter(n).encoded
 			case _ => super.traverse(tree)
 		}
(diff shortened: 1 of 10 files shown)


main/actions/src/main/scala/sbt/compiler/Eval.scala
```diff
@@ -219,7 +219,7 @@ final class Eval(optionsNoncp: Seq[String], classpath: Seq[File], mkReporter: Se
 		private[this] var vals = List[String]()
 		def getVals(t: Tree): List[String] = { vals = Nil; traverse(t); vals }
 		override def traverse(tree: Tree): Unit = tree match {
-			case ValDef(_, n, actualTpe, _) if isTopLevelModule(tree.symbol.owner) && types.exists(_ <:< actualTpe.tpe) =>
+			case ValDef(_, n, actualTpe, _) if isTopLevelModule(tree.symbol.owner) && types.exists(actualTpe.tpe.<:<) =>
```
jroper, line 222: Took me a _long_ time to track down this bug.

main/src/main/scala/sbt/EvaluateConfigurations.scala
```diff
@@ -44,11 +44,20 @@ object EvaluateConfigurations
...
+				val definitonValues = definitions.values(loader)
+				val projs = definitonValues.collect {
+					case p: Project => resolveBase(file.getParentFile, p)
+				}
+				val plugs = definitonValues.collect {
+					case pl: Plugins => pl
+				}.foldLeft(Plugins.empty)(Plugins.and)
+				(projs, plugs)
+			}
+			(imp, projPlugs)
```
jroper, line 60: Modified to return projects and plugins since we are now extracting two things from the definitions.

jroper:
I've discussed this with Christopher, and we think we have a better way of doing this, the idea is to not force you to do:

``` scala
lazy val root = project.in(file(".")).addPlugins(foo)
```

Just to add a plugin.  An alternative would be:

``` scala
lazy val root = project.root.addPlugins(foo)
```

Where `root` is a method on `project` that looks like this:

``` scala
def root = in(file("."))
```

eed3si9n:
This got closed automatically when I deleted 0.13.2 branch. Please resend it against 0.13 branch.

jroper:
I'm not convinced that it was the best way to do things.

PR #1197, opened 2014-03-21
fixes #1196

### LINKED ISSUE #1196: SBT throws an exception instead of info message if project override testGrouping in Test and contains no tests.
The same problem as in #777 if project override testGrouping in Test

sbt 0.13.1

The check if the results sequence is empty in foldTasks method (Tests.scala) is missing on the 0.13.x 

``` scala
    testGrouping in Test <<= (definedTests in Test) map { tests =>
      tests map { test =>
        new Tests.Group(
          name = test.name,
          tests = Seq(test),
          runPolicy = Tests.SubProcess(javaOptions = Seq.empty[String]))
      }
    }
```

sandreev:
Added missing fork for empty results sequence in foldTasks method, Tests.scala


jsuereth:
Ah, I see the pull request.  Odd it didn't show up on the bug.

Thanks for the fix!

PR #1198, opened 2014-03-21
Unifies AutoPlugin and AutoImport. Fixes #1188

### LINKED ISSUE #1188: sbt-0.13.2-M3 AutoImport of settings
It appears that extending AutoPlugin doesn't automatically bring in AutoImport and thus settings have to be imported explicitly.

eed3si9n:
See #1204


huntc:
I'm happy with the approach proposed here. I'll leave it to the sbt gods to review the detail. Thanks.

Issue #1142, opened 2014-03-01
Disable macro heuristic for inherited macros


gkossakowski:
At the moment, any change (even whitespace change) to a file containing a macro will trigger invalidation of files that depend on it. The reason is that we don't have any means to detect change to implementation of a macro that could affects clients that call it so we err on conservative side.

However, the definition of "containing a macro" here means both defines a macro directly in source file or _inherits_ it. Obviously, if we just inherit a macro then the heuristic doesn't need to kick in because we know that macro hasn't been modified.

The way invalidation logic was implemented internally made it difficult to distinguish between defining a macro in source file or inheriting it. The name hashing algorithm is much more amenable to such optimizations so we should consider doing it.

It would help with unexpected recompiles due whitespace change as observed in [scala compiler](https://groups.google.com/forum/#!topic/scala-internals/STmaXmH2RQg/discussion] or in Scala IDE (that inherits from Global).

gkossakowski:
Actually, I realized that the changes to name hashing are not related to this problem. Both old and new algorithms can be improved when it comes to handling of inherited macros. The decision whether given file (API in sbt's terminology) is made here in [HasMacro](https://github.com/sbt/sbt/blob/0.13/compile/api/src/main/scala/xsbt/api/APIUtil.scala#L21) API visitor.

The fix would consist of changing that visitor to not visit inherited members. The rest of the incremental compilation algorithm should work ok with restricted definition of "contains a macro".

When it comes to testing for this issue, I believe that scripted test would be enough. We don't really have unit tests related to `HasMacro` functionality.

Duhemm:
Hi Greg,

Can you confirm that [the test I wrote](https://github.com/Duhemm/sbt/commit/3fcf91709d01ac2b713c2af38a1609e56c88336c) is relevant to this problem ? 

Are you sure that fixing the API visitor should be enough ? I've been able to fix it so that a class that do not explicitly define a macro is not considered as having a macro (just like `Client` in my test), but unfortunately this doesn't seem to be enough. ([Here are the changes](https://github.com/Duhemm/sbt/commit/6625dec6c61173537e1d22d5a5e1c9a93cfd8f29)). 

I will continue investigating this week, but if you have any idea why recompilations are still triggered, please let me know !

(Do you know how I can turn this issue into a pull request ? It would be easier to track changes and discuss it. Should I create a new PR ?)

gkossakowski:
[...]

PR #1202, opened 2014-03-21
Disable macro heuristic for inherited macros (sbt/sbt#1142)

Duhemm:
> At the moment, any change (even whitespace change) to a file containing a macro will trigger invalidation of files that depend on it. The reason is that we don't have any means to detect change to implementation of a macro that could affects clients that call it so we err on conservative side.

This pull request fixes this behavior. Classes that inherit a macro are no longer considered as having a macro.

Please see [this discussion](https://github.com/sbt/sbt/issues/1142).


xeno-by:
Hi guys! What do you think about this pull request?

jsuereth:
Up to @gkossakowski .   It looks good to me, but missed the 0.13.2-RC1 by a day.  0.13.5-M2 is coming soon though (0.13.5 probably out around April), so I'm targetting it at that.

xeno-by:
Is there are you are going to have RC2? This is such a huge improvement to usability of SBT with macros!

gkossakowski:
I'll have a look either tonight or tomorrow.

jsuereth:
IF a critical issue opens in the next 2 weeks, we'll have an RC2.  We normally don't allow non-regression fixes once something is in RC and 0.13.2 has been too long delayed.  However 0.13.5 will be out like a month after 0.13.2 anyway, so you shouldn't have to wait long. 

IN fact, 0.13.5 may hit RC almost immediately after 0.13.2 is out, we'll see.

jsuereth:
Build failure appears to be a transient issue.

gkossakowski:
Hi guys, I'm sorry for being behind on this. I got busy with some other tasks. I'll definitively look into this on Wednesday.

Duhemm:
Thank you, I just made the changes. It looks like your comments disappeared. They were [here for the first commit](https://github.com/Duhemm/sbt/commit/1677ac86bf2be7f7d0783550b19d6af3de951802) and [there for the second](https://github.com/Duhemm/sbt/commit/6902e712785b62c7c6ebf897ef56e533339a0b94)

gkossakowski:
LGTM

Excellent work. Thanks!

gkossakowski:
This will be shipped as part of sbt 0.13.5.

PR #1195, opened 2014-03-20
Unifies AutoPlugin and RootPlugin

eed3si9n:
See #1203.

PR #1203, opened 2014-03-21
Unifies AutoPlugin and RootPlugin (resend)

eed3si9n:
I'm resending #1195 since target branch needs to be changed to 0.13.
- Separates `def select` into `def trigger` and `def requires`.
- Setting `trigger = noTrigger` and `requires = empty` makes a plugin an opt-in "root".
- This now allows an opt-in plugin that lists required plugins that it can depend on.

For proof of concept (for non-breakage and the new feature) see the scripted test:
- [Q.scala](https://github.com/eed3si9n/sbt/blob/f8bedf4012a5c44cb159abe2c816702767219d0b/sbt/src/sbt-test/project/auto-plugins/project/Q.scala)
- [build.sbt](https://github.com/eed3si9n/sbt/blob/f8bedf4012a5c44cb159abe2c816702767219d0b/sbt/src/sbt-test/project/auto-plugins/build.sbt)

This adds an opt-in plugin with requirements named `S`:

``` scala
// This is an opt-in plugin with a requirement
// Unless explicitly loaded by the build user, this will not be activated.
object S extends AutoPlugin
{
    def requires = Q
    def trigger = noTrigger

    override def projectSettings = Seq(
        del in q += " S"
    )
}
```

The build user just needs to bring in `S` to grab its dependency `Q`:

``` scala
// with S selected, Q is loaded automatically, which in turn selects R
lazy val projE = project.addPlugins(S)

...

    val optInValue = (del in projE in q).value
    same(optInValue, " Q S R", "del in projE in q")
```

bringing in `Q` transitively "enables" `R`. As expected, all three plugins are topologically sorted (required first).


jsuereth:
SO, I just caused some logical inconsistencies with this patch.  Particularly, the following setup:

_plugin_

``` scala
    import sbt._

object AI extends AutoImport
{
    trait EmptyAutoPlugin extends AutoPlugin {
        def requires = empty
        def trigger = noTrigger
    }
    object A extends EmptyAutoPlugin {
      val a = settingKey[String]("")
      override def projectSettings = Seq(a := "a")
    }
    object B extends EmptyAutoPlugin {
      val b = settingKey[String]("")
      override def projectSettings = Seq(b := "b")
    }

    lazy val check = settingKey[Unit]("Verifies settings are as they should be.")
}

    import AI._

object Q extends AutoPlugin
{
    def requires: Plugins = A && B
    def trigger = allRequirements
    val q = settingKey[String]("")
    override def projectSettings = Seq(q := "q")
}

object R extends AutoPlugin
{
    def requires = Q
    def trigger = allRequirements
    val r = settingKey[String]("")
    override def projectSettings = Seq(r := "r")
}

// This is an opt-in plugin with a requirement
// Unless explicitly loaded by the build user, this will not be activated.
object S extends AutoPlugin
{
    def requires = Q && !R
    def trigger = noTrigger
    val s = settingKey[String]("")
    override def projectSettings = Seq(s := "s")
}



// This is an opt-in plugin with a requirement
// Unless explicitly loaded by the build user, this will not be activated.
object T extends AutoPlugin
{
    def requires = A && !Q
    def trigger = noTrigger

    val t = settingKey[String]("")
    override def projectSettings = Seq(t := "T")
}
```

_build_

```
// with S selected, Q is loaded automatically, which in turn selects R
lazy val projA = project.addPlugins(S)

// S and T have direct conflicts of dependent plugins.
lazy val projB = project.addPlugins(S, T)

check := {
    val t = (T.t in projB).value    
    assert(t == "T", "Failed to load T plugin correctly.")
    assert((S.s in projB).value == "S", "Failed to load S plugin correctly")
    assert((Q.q in projB).value == "Q", "Failed to load Q plugin correctly")
    assert((AI.B.b in projB).value == "B", "Failed to load B plugin correctly")
    assert((AI.A.a in projB).value == "A", "Failed to load A plugin correctly")
}
```

All the excludes are getting ignored somehow, which is totlaly wrong.   The plugin-loading should immediately fail due to inconsistent versions.   It's even questionable if we want to allow users to get into this scenario.

This is the main reason I was advocating AGAINST this style of API and the previous one.   When a user wants a plugin, you don't want anything to prevent it.  This means if a plugin is automatically enabled upon dependencies it _cannot_ be specified by a user _OR_ we have to deal with this kind of crazy conflict issue.

I'll leave the test for you to dig into, but We need an answer to this issue (which will arise) before we can merge this.

jsuereth:
BTW - The check in the above does NOT denote what I think should happen. I think the abve build.sbt should fail on load with a helpful error message about competing dependencies, if this API is the one we want (which I still argue it's not really).

eed3si9n:
> All the excludes are getting ignored somehow, which is totlaly wrong. The plugin-loading should immediately fail due to inconsistent versions. 

I added your example into scripted and was able to correctly get the error:

```
[info] [error] sbt.AutoPluginException: Error determining plugins for project 'projB' in /private/var/folders/7g/l96y5q310h90xsz109qqjpmc0000gn/T/sbt_3bf32038/auto-plugins-conflict2/projB:
[info] [error] Contradiction in selected plugins.  These plguins were both included and excluded: Q, R
```

Note `test` says `-> check` but if you flip it to `> check` you should be able to see the error too.

jsuereth:
That error message isn't particularly user friendly.  How would I fix it?

The reason for the logic clas is to aide users in fixing issues.

We should get a tesr that checks the error message and the error should be
able to give me resolution options

eed3si9n:
Can `scripted` actually check for an error that happens at build level? I'm surprised even `-> check` functions correctly. I could reverse lookup the requirement and exclusion sources and inform the build user what's pulling in `Q` and `R` if you want.

jsuereth:
No... you'll have to do a unit test.

I think the error messages here are critical though.  We can't leave them
cryptic, as we used to.

That's why preventing this situarin the first place via design is use full.

eed3si9n:
I added more friendlier error message:

```
[error]    'Contradiction in enabled plugins:
[error]      - requested: sbt.AI$T && sbt.AI$U
[error]      - enabled: sbt.AI$U, sbt.AI$T, sbt.AI$A, sbt.AI$Q, sbt.AI$R, sbt.AI$B
[error]      - conflict: sbt.AI$Q is enabled by sbt.AI$A && sbt.AI$B; required by sbt.AI$T, sbt.AI$R; excluded by sbt.AI$U
[error]      - conflict: sbt.AI$R is enabled by sbt.AI$Q; excluded by sbt.AI$T'
```

Added unit tests too.

eed3si9n:
`unary_!` is now hidden from API.

PR #1204, opened 2014-03-21
Unifies AutoPlugin and AutoImport. Fixes #1188 (resend)

### LINKED ISSUE #1188: sbt-0.13.2-M3 AutoImport of settings

eed3si9n:
I'm resending #1198 since target branch needs to be changed to 0.13.

See #1188 for details. This pull req implements @jroper's `def autoImport: Any` idea. 
- `trait AutoImport` is subsumed by `def autoImport` method under `AutoPlugin`.
- When `def autoImport` is overridden by a `lazy val` or a `val`, `*.sbt` automatically imports `autoImport._`.

### DIFF
--- a/main/src/main/scala/sbt/Plugins.scala
+++ b/main/src/main/scala/sbt/Plugins.scala
@@ -19,14 +19,19 @@ Steps for plugin authors:
 1. Determine if the AutoPlugin should automatically be activated when all requirements are met, or should be opt-in.
 2. Determine the [[AutoPlugins]]s that, when present (or absent), act as the requirements for the AutoPlugin.
 3. Determine the settings/configurations to that the AutoPlugin injects when activated.
+4. Determine the keys and other names to be automatically imported to *.sbt scripts.
 
 For example, the following will automatically add the settings in `projectSettings`
   to a project that has both the `Web` and `Javascript` plugins enabled.
 
-    object MyPlugin extends AutoPlugin {
+    object Plugin extends sbt.AutoPlugin {
         def requires = Web && Javascript
         def trigger = allRequirements
         override def projectSettings = Seq(...)
+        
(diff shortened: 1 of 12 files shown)


main/src/main/scala/sbt/Plugins.scala
```diff
@@ -55,6 +52,10 @@ abstract class AutoPlugin extends Plugins.Basic
 
 	val label: String = getClass.getName.stripSuffix("$")
 
+	/** When this method is overridden with a val or a lazy val, `autoImport._` is automatically
```
jsuereth: can they override this with an `object` ?

main/src/main/scala/sbt/Plugins.scala
```diff
@@ -55,6 +52,10 @@ abstract class AutoPlugin extends Plugins.Basic
...
+	 * imported to *.sbt scripts. */
+	def autoImport: Any = ()
```
jsuereth: The "sbt way" in the past was to actually just see if something called 'autoImport' exists on the class and then write the import value, rather than having an abstract method.   Not sure how I feel about that vs. this.

jsuereth:
LGTM caveat being able to write `object` for the override. At worst, we remove the `def autoImport: Any`, at best it works as is.

eed3si9n:
I missed something with `val autoImport`.

## steps

``` scala
package sbtfoo
import sbt._

object FooPlugin extends AutoPlugin {
}
```

## problem

The name `FooPlugin` is no longer visible to `build.sbt`.

## expectation

`FooPlugin` is visible regardless of the existence of underlying `autoImport`.

Issue #1191, opened 2014-03-18
CustomPomParser produces unparseable ivy descriptor


dansanduleac:
_Edit: the issue lies differently than where I initially thought, so I explanded below to explain._

CustomPomParser adds this tag named `sbtTransformHash` in the `<info>` of ivy descriptors that are imported from a POM.
This seems to work fine when the imported ivy is in the ivy cache, but if I:
- `ivy.install` it to some ivy repository, say "local"
- do a `resolver.getDependency` on the module and force ivy to ignore the cache by setting `changing=true`, where `resolver` is the local ivy resolver

I get a parse error (the module in this case is `scala-swing;2.10.3`):

```
[debug] downloading N:\externals\org.scala-lang\scala-swing\2.10.3\ivys\ivy.xml ...
[debug]         ext-copies: downloading N:\externals\org.scala-lang\scala-swing\2.10.3\ivys\ivy.xml
[debug]         ext-copies: downloading N:\externals\org.scala-lang\scala-swing\2.10.3\ivys\ivy.xml.sha1
[debug] sha1 OK for N:\externals\org.scala-lang\scala-swing\2.10.3\ivys\ivy.xml
[debug]         [SUCCESSFUL ] org.scala-lang#scala-swing;2.10.3!ivy.xml(ivy.original) (32ms)

java.text.ParseException: [xml parsing: ivy-2.10.3.xml.original:16:21: cvc-complex-type.2.4.a: Invalid content was found starting with element 'sbtTransformHash'. One of '{WC[##other:""]}' is expected. in file:/C:/Users/dsandul/.ivy2/cache/org.scala-lang/scala-swing/ivy-2.10.3.xml.original
[error] , unknown tag sbtTransformHash in file:/C:/Users/dsandul/.ivy2/cache/org.scala-lang/scala-swing/ivy-2.10.3.xml.original
        at org.apache.ivy.plugins.parser.AbstractModuleDescriptorParser$AbstractParser.checkErrors(AbstractModuleDescriptorParser.java:89)
        at org.apache.ivy.plugins.parser.AbstractModuleDescriptorParser$AbstractParser.getModuleDescriptor(AbstractModuleDescriptorParser.java:344)
        at org.apache.ivy.plugins.parser.xml.XmlModuleDescriptorParser.parseDescriptor(XmlModuleDescriptorParser.java:118)
[...]

PR #1205, opened 2014-03-22
CustomPomParser compatibility with 0.13.1. #1191 #1192

eed3si9n:
I merged in #1192 but @jsuereth pointed out that it could break compat with 0.13.x.

@dansanduleac Is there a test you can add?


jsuereth:
LGTM

PR #1212, opened 2014-03-25
Bump Scala version to 2.10.4.

ijuma:
Targetting 0.13 branch because I assume it's too late for 0.13.2.


jsuereth:
Thanks for the fix man!

jsuereth:
Yeah. Unless 2.10.4 fixed an sbt regression in 0.13.1->0.13.2.   Also, 2.10.4 has a known regression in macros, so targetting 0.13.5 is the right way to go.

eed3si9n:
+1 on putting it in 0.13.5.

PR #1215, opened 2014-03-26
Deprecate autoSettings, Rename ProjectSettings => BuildScalaFiles

eed3si9n:
- `autoSettings` is renamed to `settingSets`. Since this was part of 0.13.0 we need to deprecate it first.
- `ProjectSettings` is not out yet, so we can rename this to `BuildScalaFiles`


jsuereth:
LGTM. Once travis says "ok" feel free to merge.

eed3si9n:
One of the travis job failed due to timeout, so I'm merging this in.

PR #1183, opened 2014-03-16
Documentation page title width overlaps search box

flurdy:
The documentation page title is set via css to 900 px. This overlaps the search box so that you cannot click on it to select it. When the screen is maxed out the last quarter to the right of the search box is selectable.

This change changes the css to 600px to reduce the title width.

Please note I do not know if there are other reasons for the width being originally set to 900px, nor do I know enough about sphinx to be sure this works as intended.

PR #1222, opened 2014-03-28
Documentation page title width overlaps search box (0.13.1)

flurdy:
The documentation page title is set via css to 900 px. This overlaps the search box so that you cannot click on it to select it. When the screen is maxed out the last quarter to the right of the search box is selectable.

This change changes the css to 600px to reduce the title width.

Please note I do not know if there are other reasons for the width being originally set to 900px, nor do I know enough about sphinx to be sure this works as intended.

(Migrated to target 0.13.1 branch as requested in https://github.com/sbt/sbt/pull/1183)


jsuereth:
Great!  Thanks for the fix!

PR #1225, opened 2014-03-31
Added setting 'testResultLogger' which allows customisation of test reporting.

### LINKED ISSUE #1185: Be silent when there are no tests to run
In a multi-module project, when I run `test` in the root which aggregates everything else, I want to be able to hide all the noise about modules without tests.

This is the kind of noise I want to opt-out of.

```
[info] No tests to run for blah-server-schema/test:test
[info] No tests to run for blah-server-logic/test:test
[info] No tests to run for blah-api/test:test
[info] No tests to run for blah-server-impl/test:test
[info] No tests to run for blah-server/test:test
[info] No tests to run for blah/test:test
```

I think a normal scoped setting would be enough. Does that sound good? Should I start writing a patch?

japgolly:
Allows users to customise the logging that accompanies tests when they finish.

Users can now silence test output on modules without tests, using this setting:

``` scala
testResultLogger := Tests.ResultLogger.silentWhenNoTests
```

The default logic is unchanged. _(Fixes #1185)_

### DIFF
--- a/main/actions/src/main/scala/sbt/Tests.scala
+++ b/main/actions/src/main/scala/sbt/Tests.scala
@@ -264,78 +264,10 @@ object Tests
 		(tests, mains.toSet)
 	}
 
+	@deprecated("Tests.showResults() has been superseded with TestResultLogger and setting 'testResultLogger'.", "0.13.5")
 	def showResults(log: Logger, results: Output, noTestsMessage: =>String): Unit =
-	{
-		val multipleFrameworks = results.summaries.size > 1
-		def printSummary(name: String, message: String) 
-		{
-			if(message.isEmpty)
-				log.debug("Summary for " + name + " not available.")
-			else
-			{
-				if(multipleFrameworks) log.info(name)
-				log.info(message)
-			}
-		}
-
-		for (Summary(name, messages) <- results.summaries)
-			printSummary(name, messages)
-		val noSummary = results.summaries.headOption.forall(_.summaryText.size == 0)
-		val printStandard = multipleFrameworks || noSummary
-		// Print the standard one-liner statistic if no framework summary is defined, or when > 1 framework is in used.
-		if (printStandard)
-		{
-			val (skippedCount, errorsCount, passedCount, failuresCount, ignoredCount, canceledCount, pendingCount) = 
-				results.events.foldLeft((0, 0, 0, 0, 0, 0, 0)) { case ((skippedAcc, errorAcc, passedAcc, failureAcc, ignoredAcc, canceledAcc, pendingAcc), (name, testEvent)) =>
-					(skippedAcc + testEvent.skippedCount, errorAcc + testEvent.errorCount, passedAcc + testEvent.passedCount, failureAcc + testEvent.failureCount, 
-					 ignoredAcc + testEvent.ignoredCount, canceledAcc + testEvent.canceledCount, pendingAcc + testEvent.pendingCount)
-				}
-			val totalCount = failuresCount + errorsCount + skippedCount + passedCount
-			val base = s"Total $totalCount, Failed $failuresCount, Errors $errorsCount, Passed $passedCount"
-
-			val otherCounts = Seq("Skipped" -> skippedCount, "Ignored" -> ignoredCount, "Canceled" -> canceledCount, "Pending" -> pendingCount)
-			val extra = otherCounts.filter(_._2 > 0).map{case(label,count) => s", $label $count" }
-
-			val postfix = base + extra.mkString
-			results.overall match {
-				case TestResult.Error => log.error("Error: " + postfix)
-				case TestResult.Passed => log.info("Passed: " + postfix)
-				case TestResult.Failed => log.error("Failed: " + postfix)
-			}
-		}
-		// Let's always print out Failed tests for now
-		if (results.events.isEmpty)
-			log.info(noTestsMessage)
-		else {
-			import TestResult.{Error, Failed, Passed}
-			import scala.reflect.NameTransformer.decode
-
-			def select(resultTpe: TestResult.Value) = results.events collect {
-				case (name, tpe) if tpe.result == resultTpe =>
-					decode(name)
-			}
-
-			val failures = select(Failed)
-			val errors = select(Error)
-			val passed = select(Passed)
-
-			def show(label: String, level: Level.Value, tests: Iterable[String]): Unit =
-				if(!tests.isEmpty)
-				{
-					log.log(level, label)
-					log.log(level, tests.mkString("\t", "\n\t", ""))
-				}
-
-			show("Passed tests:", Level.Debug, passed )
-			show("Failed tests:", Level.Error, failures)
-			show("Error during tests:", Level.Error, errors)
-		}
-
-		results.overall match {
-			case TestResult.Error | TestResult.Failed => throw new TestsFailedException
-			case TestResult.Passed => 
-		}
-	}
+    TestResultLogger.Default.copy(printNoTests = TestResultLogger.const(_ info noTestsMessage))
+      .run(log, results, "")
 }
 
 final class TestsFailedException extends RuntimeException("Tests unsuccessful") with FeedbackProvidedException
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -371,6 +371,7 @@ object Defaults extends BuildCommon
 		},
 		testListeners :== Nil,
 		testOptions :== Nil,
+		testResultLogger :== TestResultLogger.Default,
 		testFilter in testOnly :== (selectedFilter _)
 	))
 	lazy val testTasks: Seq[Setting[_]] = testTaskOptions(test) ++ testTaskOptions(testOnly) ++ testTaskOptions(testQuick) ++ testDefaults ++ Seq(
@@ -380,16 +381,15 @@ object Defaults extends BuildCommon
 		definedTestNames <<= definedTests map ( _.map(_.name).distinct) storeAs definedTestNames triggeredBy compile,
 		testFilter in testQuick <<= testQuickFilter,
 		executeTests <<= (streams in test, loadedTestFrameworks, testLoader, testGrouping in test, testExecution in test, fullClasspath in test, javaHome in test, testForkedParallel) flatMap allTestGroupsTask,
+		testResultLogger in (Test, test) :== TestResultLogger.SilentWhenNoTests, // https://github.com/sbt/sbt/issues/1185
 		test := {
-			implicit val display = Project.showContextKey(state.value)
-			Tests.showResults(streams.value.log, executeTests.value, noTestsMessage(resolvedScoped.value))
+			val trl = (testResultLogger in (Test, test)).value
+			val taskName = Project.showContextKey(state.value)(resolvedScoped.value)
+			trl.run(streams.value.log, executeTests.value, taskName)
 		},
 		testOnly <<= inputTests(testOnly),
 		testQuick <<= inputTests(testQuick)
 	)
-	private[this] def noTestsMessage(scoped: ScopedKey[_])(implicit display: Show[ScopedKey[_]]): String =
-		"No tests to run for " + display(scoped)
-
 	lazy val TaskGlobal: Scope = ThisScope.copy(task = Global)
 	lazy val ConfigGlobal: Scope = ThisScope.copy(config = Global)
 	def testTaskOptions(key: Scoped): Seq[Setting[_]] = inTask(key)( Seq(
@@ -488,9 +488,9 @@ object Defaults extends BuildCommon
 			val modifiedOpts = Tests.Filters(filter(selected)) +: Tests.Argument(frameworkOptions : _*) +: config.options
 			val newConfig = config.copy(options = modifiedOpts)
 			val output = allTestGroupsTask(s, loadedTestFrameworks.value, testLoader.value, testGrouping.value, newConfig, fullClasspath.value, javaHome.value, testForkedParallel.value)
-			val processed =
-				for(out <- output) yield
-					Tests.showResults(s.log, out, noTestsMessage(resolvedScoped.value))
+			val taskName = display(resolvedScoped.value)
+			val trl = testResultLogger.value
+			val processed = output.map(out => trl.run(s.log, out, taskName))
 			Def.value(processed)
 		}
 	}
--- a/main/actions/src/main/scala/sbt/TestResultLogger.scala
+++ b/main/actions/src/main/scala/sbt/TestResultLogger.scala
@@ -1,0 +1,159 @@
+package sbt
+
+import sbt.Tests.{Output, Summary}
+
+/**
+ * Logs information about tests after they finish.
+ *
+ * Log output can be customised by providing a specialised instance of this
+ * trait via the `testTestResultLogger` setting.
+ *
+ * @since 0.13.5
+ */
+trait TestResultLogger {
+
+  /**
+   * Perform logging.
+   *
+   * @param log The target logger to write output to.
+   * @param results The test results about which to log.
+   * @param taskName The task about which we are logging. Eg. "my-module-b/test:test"
+   */
+  def run(log: Logger, results: Output, taskName: String): Unit
+
+  /** Only allow invocation if certain criteria is met, else use another `TestResultLogger` (defaulting to nothing) . */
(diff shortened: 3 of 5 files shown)


main/actions/src/main/scala/sbt/Tests.scala
```diff
@@ -264,78 +264,105 @@ object Tests
 		(tests, mains.toSet)
 	}
 
-	def showResults(log: Logger, results: Output, noTestsMessage: =>String): Unit =
```
jsuereth, line 267: Since this is public, it can't be removed (for binary compatible reasons).  You should make sure it's still here, but delegates to the new code.  You can deprecate it, too.

main/actions/src/main/scala/sbt/Tests.scala
```diff
@@ -264,78 +264,105 @@ object Tests
...
-		{
-			if(message.isEmpty)
-				log.debug("Summary for " + name + " not available.")
-			else
-			{
-				if(multipleFrameworks) log.info(name)
-				log.info(message)
-			}
-		}
+  type ResultLogger = (Output, String, Logger) => Unit
```
jsuereth: I think I'd prefer for this to be an abstract class (or trait), so we can evolve it in a binary compatible way if needed.

jsuereth:
It's a nice fix!  

You can either target the 1.0 branch, but probably best to just fix up the binary compatibility issues so we can target the 0.13 series with this.

japgolly:
Hey Josh, I incorporated your feedback. How's it looking now?

main/actions/src/main/scala/sbt/Tests.scala
```diff
@@ -264,78 +264,130 @@ object Tests
...
-		results.overall match {
-			case TestResult.Error | TestResult.Failed => throw new TestsFailedException
-			case TestResult.Passed => 
-		}
-	}
+  @deprecated("Tests.showResults() has been superseded with Tests.ResultLogger and setting 'testResultLogger'.", "0.13.5")
+  def showResults(log: Logger, results: Output, noTestsMessage: =>String): Unit =
+    ResultLogger.default.copy(printNoTests = ResultLogger.const(_ info noTestsMessage))
+
+  trait ResultLogger {
```
jsuereth: This guy definitely needs some docs.

main/actions/src/main/scala/sbt/Tests.scala
```diff
@@ -264,78 +264,130 @@ object Tests
...
+    def apply(log: Logger, results: Output, taskName: String): Unit
+
+    final def onlyIf(f: (Output,String) => Boolean, otherwise: ResultLogger = ResultLogger.nop) =
```
jsuereth: I couldn't see if these two methods are used.   From my perspective, it's better to hide things like this in `private[sbt]` unless you _know_ you want to expose them, or you see a need.  Evolving APIs becomes rather tricky if you expose a lot more than is necessary.

jsuereth:
The feature looks good, as I said before.   Last step is Scaladoc comments for folks re-reading the code, and one question on some exposed methods.

japgolly:
Alrighty, doco added. I left those two methods you mentioned public because I think they're useful enough; I plan to use them once this feature's in.

If there's anything else you'd like me to change just give me a shout.

jsuereth:
Ok, LGTM.   cc> @eed3si9n

main/actions/src/main/scala/sbt/Tests.scala
```diff
@@ -264,78 +264,162 @@ object Tests
...
+  object ResultLogger {
+
+    /** Creates a `ResultLogger` using a given function. */
+    def apply(f: (Logger, Output, String) => Unit): ResultLogger =
+      new ResultLogger{
+        override def apply(log: Logger, results: Output, taskName: String) = f(log, results, taskName)
+      }
+
+    /** Creates a `ResultLogger` that ignores its input and always performs the same logging. */
+    def const(f: Logger => Unit) = apply((l,_,_) => f(l))
```
eed3si9n: Re: `const`, `nop`, and `silentWhenNoTests`, I think we should follow sbt's unwritten rule of no batteries included, but allow enough extension points, at least in the initial iteration. Further feature can be provided as a plugin or in [Opts.scala](https://github.com/sbt/sbt/blob/0.13/main/src/main/scala/sbt/Opts.scala).

main/actions/src/main/scala/sbt/Tests.scala
```diff
@@ -264,78 +264,162 @@ object Tests
...
+
+    /**
+      * Perform logging.
+      *
+      * @param log The target logger to write output to.
+      * @param results The test results about which to log.
+      * @param taskName The name of the task about which we are logging.
+      *                 Eg. "my-module-b/test:test"
+      */
+    def apply(log: Logger, results: Output, taskName: String): Unit
```
eed3si9n: imho capturing `taskName: String` as public API feels wrong. At this level we should be capturing as typed info as possible like `ScopedKey[_]` and let each imple worry about turning them into `String`.

main/actions/src/main/scala/sbt/Tests.scala
japgolly: It makes sense to move `silentWhenNoTests` to `Opts`. _(Sorry, didn't know that existed; first time working on SBT code.)_

I think `const` and `nop` should stay where they are because they're useful when one wants to write their own `ResultLogger`. I don't see we'd want to treat them differently than `apply` and `choose`.

eed3si9n: If you want `nop` to stay, I'd say rename to `Null` following [Logger.scala](https://github.com/sbt/sbt/blob/0.13/util/log/src/main/scala/sbt/Logger.scala).

main/actions/src/main/scala/sbt/Tests.scala
japgolly: Like I said, I don't know very much about SBT's internals but before this patch,
1) The key wasn't exposed. _(meh)_
2) Going from key to string isn't just `key.value`, its `Project.showContextKey(state.value)(key.value)` which means you need to expose `state` as well and have special knowledge that I wouldn't have had to get a string out.

In light of that, what do you think?

main/actions/src/main/scala/sbt/Tests.scala
japgolly: I just started moving `silentWhenNoTests` to `Opts`. Are you sure this is a good idea? It feels pretty wrong to me. Everything relating to `ResultLogger` used to be in one place, now with this change I'm putting half here, half there, creating dependencies...  As a user wanting to customise existing, write my own, or change `ResultLogger` in anyway, I'd expect to find everything I needed in the `ResultLogger` object.

Is this **really** necessary?

main/actions/src/main/scala/sbt/Tests.scala
jsuereth: There's an implicit "Show[_]" you could expect for tasks.  I think you're opening up some can of worms we may need to think about a bit here.   I agree with eugene in principle that we should try to use `ScopedKey[_]`, but we may need to look at the details of this to see why it's not being passed around.

main/actions/src/main/scala/sbt/Tests.scala
jsuereth: Well, at minimum, I think maybe `ResultLogger` => `TestResultLogger` and put it in its own file, including renaming ResultLoggers => TestResultLogger.

eed3si9n: From the maintenance point of view, I would like to minimize the lines of code sbt needs to maintain. So if it were up to me, I would move everything except `apply` and `default` into a plugin like sbt-test-logging. Is that really necessary? Not really. All this is nice to have.

main/actions/src/main/scala/sbt/Tests.scala
eed3si9n: Also I think it'd make sense more if `apply` was called `logResult`.

``` scala
def logResult(log: Logger, results: Ouput, key: ScopedKey[_])(implicit display: Show[ScopedKey[_]]): Unit = ...
```

japgolly:
Ok, I'm feeling like this has turned into a bike-shed exercise. Let's reset and start again. Let's confirm the following (and anything else you'd like to add) before I make more changes.
- The reason I raised this was because test results were needlessly noisy when there were no tests. I've made it extensible but should I also change the default behaviour? Currently I've left the default behaviour intact cos I didn't want to rock the boat but after this much back-and-forth already, the boat has taken water and my feet are already annoyingly cold and wet. Initial reaction by Josh seemed favourable towards changing the default behaviour. Is that still the case?
- Rename `Test.ResultLogger` to `TestResultLogger` and put in its own file. Np.
- Change `taskName` param to be a `ScopedKey[_]` and pass around an implicit `Show` instance. Gocha, np.
- Rename `nop` to `Null`. Personally I think that's silly but, sure, ok.
- Rename `apply` to `logResult`. I disagree. It's a `Test.ResultLogger`, it only does one thing. It's just a typed function. If _anything_ I'd prefer `run` or similar, `logResult` is redundant.
- Anything else I missed?

eed3si9n:
> Ok, I'm feeling like this has turned into a bike-shed exercise.

Haha. Sorry if it felt like nitpicking/bikesedding because it _is_ exactly what's going on. I'm generally ready to accept this patch, and I'm not trying give you hard time to reject it. I figured I'd give my two cents now instead of changing it later, so you have a fair chance of rebuttal. That's fair right?

> I've made it extensible but should I also change the default behaviour? 

I don't think it should be completely silent by default. If I'm in a particular subproject running `testOnly` I want the shell to tell me if I've made a typo. If I'm in root project of that aggregates a bunch of sub projects, I do find it noisy too, but I think the core philosophy of aggregate is that it just aggregates and doesn't add extra behavior.

> Rename `apply` to `logResult`. I disagree. It's a Test.ResultLogger, it only does one thing. It's just a typed function.

When I first saw the code, I honestly thought `apply` is a factory method to create a logger, which doesn't make sense because it's returning `Unit`. See usage code in Defaults:

``` scala
testResultLogger.value(streams.value.log, executeTests.value, display(resolvedScoped.value))
```

It's really not clear what it's doing. But it could be:

``` scala
val trl = testResultLogger.value
trl.logResult(streams.value.log, executeTests.value, resolvedScoped.value)
```

japgolly:
Ah cool, no worries; that's totally fair. In that case I'll keep coming back with my own, hmmm, well I've only 1 cent, well, my own one cent, and then if it gets too tedious just call veto and I'll drop it. (I know I don't have the big-picture view of SBT that you guys do.)

> I don't think it should be completely silent by default. If I'm in a particular subproject running testOnly I want the shell to tell me if I've made a typo. If I'm in root project of that aggregates a bunch of sub projects, I do find it noisy too, but I think the core philosophy of aggregate is that it just aggregates and doesn't add extra behavior.

Makes perfect sense. What about if, when no tests exist, we made it silent when `test` was run and maintained the current behaviour when `test-only`?

Re: `apply` grievances, ok I see your point. How about `run` then? I think `testResultLogger.value.run(...)` addresses both of our viewpoints.

eed3si9n:
> What about if, when no tests exist, we made it silent when `test` was run and maintained the current behaviour when `test-only`?

I'm ok with that.

> Re: `apply` grievances, ok I see your point. How about `run` then?

`run` is fine.

main/actions/src/main/scala/sbt/Tests.scala
jsuereth: That's true. Odd are if it's in sbt, your ability to change it is very low.

However, given we have _no_ way to configure it now, I think `TestResultLogger` (the minimum) definitely needs to go in sbt.

jsuereth:
Yep, looking forward to the cleanup on this one!  Will be a nice improvement.  Thanks.

japgolly:
Awesome, well, I think there's just one question left then. When providing different behaviour for `test` vs `test-only` I see two possible ways of doing it, I'd like your advice.

1) Is it possible to have something like `testResultLogger in Test := ...; testResultLogger in TestOnly := ...`. As I understand `testOnly` is a task which means it can't be used to scope, right? Is there a similar way?

2) If I'm to use the `taskName` key, I could simply check if it ends in `:test` but that seems a bit fickle to me. The `Show` typeclass doesn't guarantee any consistency here (not is it meant to). Secondly, the param would be a `ScopedKey[_]`. Is there a reason it's type param isn't specified? I wonder if I could just use `.value` directly to inspect it.

eed3si9n:
> As I understand `testOnly` is a task which means it can't be used to scope, right?

Not only there's task scoping, it's encouraged for any internal settings like this to be scoped into tasks like `jarName in (Compile, assembly)`. I'm actually not sure about `ScopedKey[_]`. If it's possible to use narrower types, certainly that's better.

jsuereth:
Yeah, you basically want the following:

``` scala
testLogger in (Test, test) := testLogger in Test
testLogger in (Test, testOnly) := testLogger in Test
```

And then have the individual tasks depend on their task-specific testLogger key.  Users can directly override it as appropriate.

japgolly:
Hi guys, major problem on that point of using `task: SettingKey` instead of `taskName: String`. I now understand why the current implementation doesn't do this either.

**Background:**
1. `SettingKey` isn't a type, it's a dependent type constructor. I have to use `Init[_]#SettingKey[_]`.
2. Good 'ole `Show` is invariant and can be contravariant at best.

**Problem:**
In `Defaults.inputTests0` it attempts to log test results using a `Show` instance for `Def.SettingKey[_]`. The compiler fails because it cannot prove that a `Show[Def.SettingKey[_]]` is a `Show[Init[_]#SettingKey[_]]` which is actually correct.

**Solutions I can think of:**
1. Change TestResultLogger to use `Def.SettingKey[_]` would be mean putting `sbt.Def` on its classpath.
2. Change `Project.showContextKey` to return a `Show[Init[_]#SettingKey[_]]` and ensure that whatever logic therein works with any kind of `SettingKey[_]`.
3. Use string.
4. Rework that odd looking type hierarchy. _(I know it's not an option given SBT's adoption, just let me dream!)_

eed3si9n:
Well that sucks. Sorry you had to run around for that. Let's just use `String`.

japgolly:
That's life. `String` it shall be!

japgolly:
Sorry for being so noisy but I have another question. I haven't been able to get SBT to provide a different `testResultLogger` setting for `test`.

My attempts to do this are changing `test::testResultLogger` but `inspect test` shows that it depends on `test:testResultLogger`. :confused:

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -376,16 +377,16 @@ object Defaults extends BuildCommon
 		definedTestNames <<= definedTests map ( _.map(_.name).distinct) storeAs definedTestNames triggeredBy compile,
 		testFilter in testQuick <<= testQuickFilter,
 		executeTests <<= (streams in test, loadedTestFrameworks, testLoader, testGrouping in test, testExecution in test, fullClasspath in test, javaHome in test, testForkedParallel) flatMap allTestGroupsTask,
+    //testResultLogger in test :== TestResultLogger.SilentWhenNoTests, // https://github.com/sbt/sbt/pull/1225
+    testResultLogger in (Test, test) :== TestResultLogger.const(_ info "TODO 123"),
+    testResultLogger in test         :== TestResultLogger.const(_ info "TODO 234"),
```
japgolly: This is how I was trying to override the default but isn't working.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -376,16 +377,16 @@ object Defaults extends BuildCommon
...
 		test := {
-			implicit val display = Project.showContextKey(state.value)
-			Tests.showResults(streams.value.log, executeTests.value, noTestsMessage(resolvedScoped.value))
+			val taskName = Project.showContextKey(state.value)(resolvedScoped.value)
+			testResultLogger.value.run(streams.value.log, executeTests.value, taskName)
```
eed3si9n: Basically instead of calling `testResultLogger.value` you do

``` scala
val trl = (testResultLogger in (Test, test)).value
trl.run(...)
```

`testTaskOptions` defines test-related options bound to a given key.

japgolly:
Thanks for that, Eugene. It fixed it.

I think we're good to go. Please take another look at this patch when you have time. Cheers!

eed3si9n:
LGTM. Could you squash your commits and rebase it to the head? I'll merge this tomorrow unless there're objections.

japgolly:
np, done

eed3si9n:
Merged. Thanks for your contribution!

japgolly:
Thank you for helping me with my first SBT patch. :smile:

PR #1227, opened 2014-04-02
Autorun GC and finalizers after each task run.

### LINKED ISSUE #1223: Reducing PermGen usage in sbt

```
```

jsuereth:
- Add `sbt.task.autogc` flag to disable
- Add hook to automatically GC/finalizer after running tasks or
  cancelling tasks

This is to workaround too-lazy GC leaving too many classloaders in
memory and causing PermGen OOMs far before necessary.

Fixes #1223 

Review by @cunei  and @eed3si9n 

Discussion for @ijuma ->  What do you think of the current mechanism to disable this feature?   Until I can break binary compatibility (and put the flag on the execution config object), this was the best solution I had, which is why it's `private[sbt]` (so we can modify it in 1.0 for a better solution).

### DIFF
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -14,7 +14,11 @@ package sbt
 	import std.Transform.{DummyTaskMap,TaskAndValue}
 	import TaskName._
 
-final case class EvaluateConfig(cancelable: Boolean, restrictions: Seq[Tags.Rule], checkCycles: Boolean = false, progress: ExecuteProgress[Task] = EvaluateTask.defaultProgress)
+final case class EvaluateConfig(
+	cancelable: Boolean, 
+	restrictions: Seq[Tags.Rule], 
+	checkCycles: Boolean = false, 
+	progress: ExecuteProgress[Task] = EvaluateTask.defaultProgress)
 final case class PluginData(dependencyClasspath: Seq[Attributed[File]], definitionClasspath: Seq[Attributed[File]], resolvers: Option[Seq[Resolver]], report: Option[UpdateReport], scalacOptions: Seq[String])
 {
 	val classpath: Seq[Attributed[File]] = definitionClasspath ++ dependencyClasspath
@@ -37,6 +41,19 @@ object EvaluateTask
 
 	private[sbt] def defaultProgress: ExecuteProgress[Task] =
 		if(java.lang.Boolean.getBoolean("sbt.task.timings")) new TaskTimings else ExecuteProgress.empty[Task]
+
+
+	// Default is true, but can be disabled.
+	private[sbt] val isAutoGcEnabled =
+		System.getProperty("sbt.task.autogc") == null ||
+		java.lang.Boolean.getBoolean("sbt.task.autogc")
+	private[sbt] val isPoorlyActingGcOnJvm: Boolean = {
+		// TODO - Not so hacky a mechanism here.
+		val javaVersion = System.getProperty("java.version")
+		javaVersion.startsWith("1.5") ||
+		javaVersion.startsWith("1.6") ||
+		javaVersion.startsWith("1.7")
+	}
 
 	val SystemProcessors = Runtime.getRuntime.availableProcessors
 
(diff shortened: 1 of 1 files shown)


ijuma:
System property seems good enough to me.

avakhrenev:
With sbt 0.13.5-M2 I'm observing following issue: when `exportJars := true` after executing `testQuick` the jar is not deleted on `clean` task. It can be deleted only after project reload.
This change seems to fix that.

jsuereth:
Hmm, that somewhat makes sense, depending on your OS.   Can you tell me which OS you were on?   I know with windows, if a file-handle is open, the delete will not succeed.  On linux they just assume you know you're asking for applications to fail if you rm.  I'd be curious if other OSes have that same model.

It seems some of our files are left open by URLClassloaders and aren't closed until the full finalization runs, which is how this fixes that issue.  Now, I just have to figure out why the build is failing.

avakhrenev:
I'm on Windows 7 x64 - seems like it doesn't allow concurrently modify opened files. Another effect of this classloading issue: jars in `unmanagedBase` can't be modified if they picked by SBT.

Thank you for your work!

main/src/main/scala/sbt/EvaluateTask.scala
```diff
@@ -38,6 +42,19 @@ object EvaluateTask
 	private[sbt] def defaultProgress: ExecuteProgress[Task] =
 		if(java.lang.Boolean.getBoolean("sbt.task.timings")) new TaskTimings else ExecuteProgress.empty[Task]
 
+
+	// Default is true, but can be disabled.
+	private[sbt] val isAutoGcEnabled =
+		System.getProperty("sbt.task.autogc") == null ||
+		java.lang.Boolean.getBoolean("sbt.task.autogc")
+	private[sbt] val isPoorlyActingGcOnJvm: Boolean = {
```
jsuereth, line 50: From @cunei -  Recommends also running for Java 8 because even though the gc is better, it can't be relied upon completely to clean things up.

ijuma, line 50: By the way, what GCs were tested and with which options?

main/src/main/scala/sbt/EvaluateTask.scala
```diff
@@ -185,6 +202,22 @@ object EvaluateTask
 		val tags = tagged[Task[_]](_.info get tagsKey getOrElse Map.empty, Tags.predicate(config.restrictions))
 		val (service, shutdown) = completionService[Task[_], Completed](tags, (s: String) => log.warn(s))
 
+		// FIX - We create a cleanup method here which ensures the thread pool is closed down
+		//       and we attempt to force the JVM < 8 to release classloaders, due to issues
+		//       with WeakHashMaps/WeakHashSets and lazy GCs not running finalizers.
+		def shutdownAndCleanup(): Unit = {
```
jsuereth, line 208: from @eed3si9n - Expose the gc/finalizer/gc method somewhere like (in sbt.IO) with the enablement flag so we can re-use it, e.g. in sbt reload, where we want Classloaders to be cleaned but aren't executing a task.

main/src/main/scala/sbt/EvaluateTask.scala
jsuereth, line 50: OpenJDKs + Oracle JDK8 I believe.  I think the options were the defaults, but maybe @cunei can chime in with more info.

cunei, line 50: That is correct: OpenJDK 1.6 & 1.7 (the most recent versions available on Ubuntu 13.10) and Oracle's JDK 1.8.0-b132. I tested a variety of options, including CMSClassUnloadingEnabled with UseConcMarkSweepGC, varying amounts of PermSize, MaxPermSize, MinPermHeapExpansion, MaxPermHeapExpansion, changes in SoftRefLRUPolicyMSPerMB, of course DoEscapeAnalysis (which, even if it reduces PermGen occupation somewhat, does not solve the leak). The results of the investigation is that openjdk1.6/1.7 are rather lazy with finalization. JDK1.8 was more prompt, but because of the nature of finalization (which runs using a separate thread) it cannot be relied upon to take place when needed. Hence I would personally recommend the explicit runFinalization() with JDK1.8 as well.

jsuereth:
Alright, I think I'm going to modify this PR slightly to use: http://docs.oracle.com/javase/7/docs/api/java/lang/management/MemoryMXBean.html

With this, we can get an approximate number of "objects requiring finalization".   We can set some kind of configurable threshold (or never) that users can use to control when we attempt to force GC/Finalization after task execution.

I'm not sure what the heuristic should be, but woudl love to hear some ideas :)

jsuereth:
Closing until refactored.

PR #1231, opened 2014-04-04
add reference to sbt-git-stamp plugin

pkaeding:


jsuereth:
Great! Thanks for the plugin.

PR #1235, opened 2014-04-05
Documentation default paths filtering resources use unmanagedResources (0.13.1)

flurdy:
Correct target branch for previous pull request https://github.com/sbt/sbt/pull/1182

In the documentation filtering resources should use unmanagedResources and not unmanagedSources.


jsuereth:
Thanks again!

PR #1236, opened 2014-04-05
Implemented the fileParser and added DynamicExamples.

urbas:
Main additions:
- Implemented `Parsers.fileParser` (removes a TODO)
- Added `Parser.DynamicExamples` (lazily generates/regenerates completions; especially useful for file completions if there are a lot of files in subdirectories)
- Added `Parser.SourceOfExamples` (an abstraction of the set of examples; allows lazy listing of examples; the only alternative before this was `Set[String]`, which cannot be lazy; `SourceOfExamples` is used by `DynamicExamples`)
- Added `Parsers.FileExamples` (searches for files/directories recursively; it should be fairly efficient, as it lazily traverses only the path with the user's prefix; if there's no prefix, then only a limited number of files are displayed&mdash;this avoids long recursive traversals)

### DIFF
--- a/util/complete/src/main/scala/sbt/complete/ExampleSource.scala
+++ b/util/complete/src/main/scala/sbt/complete/ExampleSource.scala
@@ -1,0 +1,49 @@
+package sbt.complete
+
+import java.io.File
+
+/**
+ * These sources of examples are used in parsers for user input completion. An example of such a source is the
+ * [[sbt.complete.FileExamples]] class, which provides a list of suggested files to the user as they press the
+ * TAB key in the console.
+ */
+trait ExampleSource
+{
+  /**
+   * @return a (possibly lazy) list of completion example strings. These strings are continuations of user's input. The
+   *         user's input is incremented with calls to [[withAddedPrefix]].
+   */
+  def apply(): Iterable[String]
+
+  /**
+   * @param addedPrefix a string that just typed in by the user.
+   * @return a new source of only those examples that start with the string typed by the user so far (with addition of
+   *         the just added prefix).
+   */
+  def withAddedPrefix(addedPrefix: String): ExampleSource
+}
+
+/**
+ * Provides path completion examples based on files in the base directory.
+ * @param base the directory within which this class will search for completion examples.
+ * @param prefix the part of the path already written by the user.
+ */
+class FileExamples(base: File, prefix: String = "") extends ExampleSource {
+  private val relativizedPrefix: String = "." + File.separator + prefix
+
+  override def apply(): Iterable[String] = files(base).map(_.toString.substring(relativizedPrefix.length))
+
+  override def withAddedPrefix(addedPrefix: String): FileExamples = new FileExamples(base, prefix + addedPrefix)
+
+  protected def fileStartsWithPrefix(path: File): Boolean = path.toString.startsWith(relativizedPrefix)
+
+  protected def directoryStartsWithPrefix(path: File): Boolean = {
+    val pathString = path.toString
+    pathString.startsWith(relativizedPrefix) || relativizedPrefix.startsWith(pathString)
+  }
+
+  protected def files(directory: File): Iterable[File] = {
+    val (subDirectories, filesOnly) = directory.listFiles().toStream.partition(_.isDirectory)
+    filesOnly.filter(fileStartsWithPrefix) ++ subDirectories.filter(directoryStartsWithPrefix).flatMap(files)
+  }
+}
--- a/util/complete/src/main/scala/sbt/complete/ExampleSource.scala
+++ b/util/complete/src/main/scala/sbt/complete/ExampleSource.scala
@@ -24,7 +24,8 @@ trait ExampleSource
 }
 
 /**
- * @param examples the source of examples that will be displayed to the user when they press the TAB key.
+ * A convenience example source that wraps any collection of strings into a source of examples.
+ * @param examples the examples that will be displayed to the user when they press the TAB key.
  */
 sealed case class FixedSetExamples(examples: Iterable[String]) extends ExampleSource {
 	override def withAddedPrefix(addedPrefix: String): ExampleSource = FixedSetExamples(examplesWithRemovedPrefix(addedPrefix))
--- a/util/complete/src/main/scala/sbt/complete/ExampleSource.scala
+++ b/util/complete/src/main/scala/sbt/complete/ExampleSource.scala
@@ -1,6 +1,7 @@
 package sbt.complete
 
 import java.io.File
+import sbt.IO._
 
 /**
  * These sources of examples are used in parsers for user input completion. An example of such a source is the
@@ -27,12 +28,15 @@ trait ExampleSource
  * A convenience example source that wraps any collection of strings into a source of examples.
  * @param examples the examples that will be displayed to the user when they press the TAB key.
  */
-sealed case class FixedSetExamples(examples: Iterable[String]) extends ExampleSource {
+sealed case class FixedSetExamples(examples: Iterable[String]) extends ExampleSource
+{
 	override def withAddedPrefix(addedPrefix: String): ExampleSource = FixedSetExamples(examplesWithRemovedPrefix(addedPrefix))
 
 	override def apply(): Iterable[String] = examples
 
-	private def examplesWithRemovedPrefix(prefix: String) = examples.collect { case example if example startsWith prefix => example substring prefix.length }
+	private def examplesWithRemovedPrefix(prefix: String) = examples.collect {
+		case example if example startsWith prefix => example substring prefix.length
+	}
 }
 
 /**
@@ -40,22 +44,19 @@ sealed case class FixedSetExamples(examples: Iterable[String]) extends ExampleSo
  * @param base the directory within which this class will search for completion examples.
  * @param prefix the part of the path already written by the user.
  */
-class FileExamples(base: File, prefix: String = "") extends ExampleSource {
-	private val relativizedPrefix: String = "." + File.separator + prefix
-
-	override def apply(): Iterable[String] = files(base).map(_.toString.substring(relativizedPrefix.length))
+class FileExamples(base: File, prefix: String = "") extends ExampleSource
+{
+	override def apply(): Stream[String] = files(base).map(_ substring prefix.length)
 
 	override def withAddedPrefix(addedPrefix: String): FileExamples = new FileExamples(base, prefix + addedPrefix)
 
-	protected def fileStartsWithPrefix(path: File): Boolean = path.toString.startsWith(relativizedPrefix)
-
-	protected def directoryStartsWithPrefix(path: File): Boolean = {
-		val pathString = path.toString
-		pathString.startsWith(relativizedPrefix) || relativizedPrefix.startsWith(pathString)
+	protected def files(directory: File): Stream[String] = {
+		val childPaths = directory.listFiles().toStream
+		val prefixedDirectChildPaths = childPaths.map(relativize(base, _).get).filter(_ startsWith prefix)
+		val dirsToRecurseInto = childPaths.filter(_.isDirectory).map(relativize(base, _).get).filter(dirStartsWithPrefix)
+		prefixedDirectChildPaths append dirsToRecurseInto.flatMap(dir => files(new File(base, dir)))
 	}
 
-	protected def files(directory: File): Iterable[File] = {
-		val (subDirectories, filesOnly) = directory.listFiles().toStream.partition(_.isDirectory)
-		filesOnly.filter(fileStartsWithPrefix) ++ subDirectories.filter(directoryStartsWithPrefix).flatMap(files)
-	}
+	private def dirStartsWithPrefix(relativizedPath: String): Boolean =
+		(relativizedPath startsWith prefix) || (prefix startsWith relativizedPath)
 }
--- a/util/complete/src/main/scala/sbt/complete/ExampleSource.scala
+++ b/util/complete/src/main/scala/sbt/complete/ExampleSource.scala
@@ -9,18 +9,29 @@ import java.io.File
  */
 trait ExampleSource
 {
-  /**
-   * @return a (possibly lazy) list of completion example strings. These strings are continuations of user's input. The
-   *         user's input is incremented with calls to [[withAddedPrefix]].
-   */
-  def apply(): Iterable[String]
+	/**
+	 * @return a (possibly lazy) list of completion example strings. These strings are continuations of user's input. The
+	 *         user's input is incremented with calls to [[withAddedPrefix]].
+	 */
+	def apply(): Iterable[String]
 
-  /**
-   * @param addedPrefix a string that just typed in by the user.
-   * @return a new source of only those examples that start with the string typed by the user so far (with addition of
-   *         the just added prefix).
-   */
-  def withAddedPrefix(addedPrefix: String): ExampleSource
+	/**
+	 * @param addedPrefix a string that just typed in by the user.
+	 * @return a new source of only those examples that start with the string typed by the user so far (with addition of
+	 *         the just added prefix).
+	 */
+	def withAddedPrefix(addedPrefix: String): ExampleSource
+}
+
+/**
+ * @param examples the source of examples that will be displayed to the user when they press the TAB key.
+ */
+sealed case class FixedSetExamples(examples: Iterable[String]) extends ExampleSource {
+	override def withAddedPrefix(addedPrefix: String): ExampleSource = FixedSetExamples(examplesWithRemovedPrefix(addedPrefix))
+
+	override def apply(): Iterable[String] = examples
+
+	private def examplesWithRemovedPrefix(prefix: String) = examples.collect { case example if example startsWith prefix => example substring prefix.length }
 }
 
 /**
@@ -29,21 +40,21 @@ trait ExampleSource
  * @param prefix the part of the path already written by the user.
  */
 class FileExamples(base: File, prefix: String = "") extends ExampleSource {
-  private val relativizedPrefix: String = "." + File.separator + prefix
+	private val relativizedPrefix: String = "." + File.separator + prefix
 
-  override def apply(): Iterable[String] = files(base).map(_.toString.substring(relativizedPrefix.length))
+	override def apply(): Iterable[String] = files(base).map(_.toString.substring(relativizedPrefix.length))
 
-  override def withAddedPrefix(addedPrefix: String): FileExamples = new FileExamples(base, prefix + addedPrefix)
+	override def withAddedPrefix(addedPrefix: String): FileExamples = new FileExamples(base, prefix + addedPrefix)
 
-  protected def fileStartsWithPrefix(path: File): Boolean = path.toString.startsWith(relativizedPrefix)
+	protected def fileStartsWithPrefix(path: File): Boolean = path.toString.startsWith(relativizedPrefix)
 
-  protected def directoryStartsWithPrefix(path: File): Boolean = {
-    val pathString = path.toString
-    pathString.startsWith(relativizedPrefix) || relativizedPrefix.startsWith(pathString)
-  }
+	protected def directoryStartsWithPrefix(path: File): Boolean = {
+		val pathString = path.toString
+		pathString.startsWith(relativizedPrefix) || relativizedPrefix.startsWith(pathString)
+	}
 
-  protected def files(directory: File): Iterable[File] = {
-    val (subDirectories, filesOnly) = directory.listFiles().toStream.partition(_.isDirectory)
-    filesOnly.filter(fileStartsWithPrefix) ++ subDirectories.filter(directoryStartsWithPrefix).flatMap(files)
-  }
+	protected def files(directory: File): Iterable[File] = {
+		val (subDirectories, filesOnly) = directory.listFiles().toStream.partition(_.isDirectory)
+		filesOnly.filter(fileStartsWithPrefix) ++ subDirectories.filter(directoryStartsWithPrefix).flatMap(files)
+	}
 }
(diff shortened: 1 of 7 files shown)


eed3si9n:
Thanks for the contribution.

util/complete/src/main/scala/sbt/complete/Parsers.scala
```diff
@@ -128,8 +128,30 @@ trait Parsers
...
+			pathString.startsWith(relativizedPrefix) || relativizedPrefix.startsWith(pathString)
+		}
+
+		protected def files(directory: File): Iterable[File] = {
+			val (subDirectories, filesOnly) = directory.listFiles().toStream.partition(_.isDirectory)
+			filesOnly.filter(fileStartsWithPrefix) ++ subDirectories.filter(directoryStartsWithPrefix).flatMap(files)
+		}
+	}
+
+	def fileParser(base: File, maxNumberOfExamples: Int = 25): Parser[File] =
```
eed3si9n: Adding an extra param even if you supply default value breaks binary compatibility iirc. You might have to provide overloads.

jsuereth: Yeah, overloads is the only safe way to do this, unfortunately.

jsuereth:
Yeah, great work here!  Can't wait to use this.  Tried to comment on places where I saw some fixes.  Will follow up more tomorrow.

util/complete/src/main/scala/sbt/complete/Parsers.scala
urbas: Makes sense. Will create an overload.

util/complete/src/main/scala/sbt/complete/ExampleSource.scala
```diff
@@ -0,0 +1,60 @@
+package sbt.complete
+
+import java.io.File
+
+/**
+ * These sources of examples are used in parsers for user input completion. An example of such a source is the
+ * [[sbt.complete.FileExamples]] class, which provides a list of suggested files to the user as they press the
+ * TAB key in the console.
+ */
+trait ExampleSource
```
jsuereth, line 11: This looks nice.

util/complete/src/main/scala/sbt/complete/Parser.scala
```diff
@@ -87,6 +87,24 @@ sealed trait RichParser[A]
...
+	 * @return a new parser with a new source of completions.
+	 */
+	def examples(exampleSource: ExampleSource, maxNumberOfExamples: Int, removeInvalidExamples: Boolean): Parser[A]
+
+	/**
+	 * @param exampleSource the source of examples when displaying completions to the user.
+	 * @return a new parser with a new source of completions. It displays at most 25 completion examples and does not
+	 *         remove invalid examples.
+	 */
+	def examples(exampleSource: ExampleSource): Parser[A] = examples(exampleSource, maxNumberOfExamples = 25, removeInvalidExamples = false)
```
jsuereth, line 106: Note:  When it comes to which examples to show, I guess we're relegating the filtering mechanism to the ExampleSource ?

util/complete/src/main/scala/sbt/complete/Parser.scala
```diff
@@ -703,6 +744,32 @@ private final class Examples[T](delegate: Parser[T], fixed: Set[String]) extends
 			Completions(fixed map(f => Completion.suggestion(f)) )
 	override def toString = "examples(" + delegate + ", " + fixed.take(2) + ")"
 }
+private final class DynamicExamples[T](delegate: Parser[T], exampleSource: ExampleSource, maxNumberOfExamples: Int, removeInvalidExamples: Boolean) extends ValidParser[T]
```
jsuereth: So, I think this guy needs some docs too (for maintainers).  Primarily, as I've thought about it, this is what this guy is about:

His `maxNumberOfExamples` is more about limiting the amount of gunk we keep around in memory.  Keeping  a ton of File/String objects for traversing a filesystem is actually kind of bad.   "Dynamic" here in the sense that we wind up with a TON of possible completions and it's expensive to compute them all up front.  I'd make a note of that in the description.

The filtering isn't about UI, that should be done in the JLineReader or other autocompletion library things.  The maxNumberOfExamples is because it's expensive to compute the full transitive closure of completions for a filesystem (and other related things).

Given that, I'm ok with this going in as-is.

jsuereth:
Nice, that's way cleaner, great work!  One last comment and I think it's good to merge.

util/complete/src/main/scala/sbt/complete/Parser.scala
urbas, line 106: Yes, I didn't manage to finish this. I wanted to port the two old methods:

``` scala
    /** Explicitly defines the completions for the original Parser.*/
    def examples(s: String*): Parser[A]

    /** Explicitly defines the completions for the original Parser.*/
    def examples(s: Set[String], check: Boolean = false): Parser[A]
```

to the new `examples(ExampleSource)` method, delete the `Examples` class, and rename `DynamicExamples` to `Examples`. I'll prepare an example on another branch (to test whether it works at all), and then we can decide whether to scrap the idea.

util/complete/src/main/scala/sbt/complete/Parser.scala
```diff
@@ -703,6 +744,32 @@ private final class Examples[T](delegate: Parser[T], fixed: Set[String]) extends
...
+		if(exampleSource().isEmpty)
+			if(resultEmpty.isValid) Completions.nil else Completions.empty
+		else {
+			val examplesBasedOnTheResult = filteredExamples.take(maxNumberOfExamples).toSet
+			Completions(examplesBasedOnTheResult.map(ex => Completion.suggestion(ex)))
+		}
+	}
+	override def toString = "examples(" + delegate + ", " + exampleSource().take(2).toList + ")"
+
+	private def filteredExamples: Iterable[String] = {
```
urbas, line 762: A note of warning: this filter behaves differently to the `checkMatches` method (which I am thinking of replacing it with). The latter will throw an exception if no valid examples remain after filtering. IMHO, having zero valid examples is not really an exception case. After all, it is only completions we are talking about, nothing wrong with having no completions. :)

util/complete/src/main/scala/sbt/complete/Parser.scala
urbas: Will write up some docs then.

util/complete/src/main/scala/sbt/complete/Parser.scala
jsuereth, line 762: So, I think the behavior is around these assumptions:
1. There is always one example
2. The example returned is always a complete string.

SO, failure to find _one_ example means someone typed in malformed input.  I agree with "dynamic" completion, it's possible that the first assumption is bad, so we should revisit that.

urbas:
Here's the proposed port of the set-based `examples` to `ExampleSource`:

>   https://github.com/urbas/sbt/commit/5b40b0c73aec41a1ce2e60391ae87b22150efc63

It seems to be working (I tested it in the sbt console with `testOnly`--I just kept pressing TAB to autocomplete class paths). Should I test with something else?

In any case, please, tell me whether I should include this commit in this pull request.

Oh, btw, should I maybe write some tests for the newly introduced classes? I've looked at `ParserTest.scala`, but the `ParserExample` object seems to just print some stuff? Well, I am not sure what it's doing. Is this throw-away code? I'd gladly get rid of it, if that's so. :smile: 

Well, in any case, just give me some commands, and I shall execute.

jsuereth:
The new commit looks like the right direction for me.  

Also, for testing, I think you may want to write some.   I'm not sure what kind of testing we do besides dorking around on the sbt console right now, can check with @harrah.   If you had time to add specific tests just to capture the _design_ of the feature you want, that'd be great too.

Great work!  Can't wait to get this in for 0.13.5.

urbas:
Sure, I'll add some unit tests. -Btw, when's 0.13.5 due?-

Update: Found the info on google groups. Feature freeze is on Wednesday, right? Should be doable.

jsuereth:
Cool.  Thanks!

util/complete/src/test/scala/sbt/complete/FixedSetExamplesTest.scala
```diff
@@ -0,0 +1,26 @@
+package sbt.complete
+
+import org.specs2.mutable.Specification
+import org.specs2.specification.Scope
+
+class FixedSetExamplesTest extends Specification {
```
urbas, line 6: I didn't write tests for the `FileExamples` class because the following options were not ideal:
-  I could mock the `File` class and fake a directory structure, but I couldn't determine whether SBT has a dependency on a mock framework (such as Mockito). If you're okay with adding a mocking framework to SBT I can go ahead with this. (Btw, I suggest using Mockito. I have some experience using Mockito in ScalaTest, and it's something I would recommend. Having mocks would also help in the other tests I wrote.)
- I could subclass `File` and create fake files and with a fake directory structure. This could work, but I'd have to create boolean flags or counters to verify that certain paths have not been traversed (when limiting the number of examples).
- I could simply use SBT's own directory structure and files in the test. However, I would recommend strongly against this for a number of reasons: tests now depend on external state that has nothing to do with the tests, hidden dependency, no way to verify whether a directory has been traversed or not (unlike the two approaches above).

What to do?

util/complete/src/test/scala/sbt/complete/ParserWithExamplesTest.scala
```diff
@@ -0,0 +1,93 @@
...
+		val colorParser = "blue" | "green" | "black" | "red"
+		val parserWithExamples = new ParserWithExamples[String](
+			colorParser,
+			FixedSetExamples(examples),
+			maxNumberOfExamples,
+			removeInvalidExamples
+		)
+	}
+
+	case class GrowableSourceOfExamples() extends Iterable[String] {
```
urbas, line 73: I could mock this instead of writing all this logic if we had a mock framework. Just saying :)

jsuereth, line 73: If you find a mock framework without classloader leaks, feel free to include it.

util/complete/src/test/scala/sbt/complete/FixedSetExamplesTest.scala
jsuereth, line 6: - Mockito has issues with leaking classloaders.  I'd prefer not to use it.  
- Subclassing file also seems risky

Why not just generate a temporary directory (`IO.withTemporaryDirectory`) and create the files you want in the test?   When it comes to testing your fileParser and reading the filesystem, I think it's ok to test against the filesystem itself, because running, e.g. on Windows could be enlightening for finding issues in cross platfrom assumptions that you'd never get with something like Mockito.

urbas, line 6: Oh, nice, will add this.

jsuereth:
So one quick question, otherwise LGTM and I think it's ok to merge for the upcoming sbt 0.13.5-M4 release.

@eed3si9n any thoughts?

util/complete/src/main/scala/sbt/complete/Parser.scala
```diff
@@ -156,16 +174,9 @@ object Parser extends ParserMain
 	def mkFailures(errors: => Seq[String], definitive: Boolean = false): Failure = new Failure(errors.distinct, definitive)
 	def mkFailure(error: => String, definitive: Boolean = false): Failure = new Failure(error :: Nil, definitive)
 
-	def checkMatches(a: Parser[_], completions: Seq[String])
```
eed3si9n, line 159: You're under arrest for bincompat violation against 0.13.0.

jsuereth, line 159: OH NOES! The compatibility police have caught you.

This method has to remain, but you can deprecate it.

util/complete/src/main/scala/sbt/complete/Parser.scala
```diff
@@ -418,14 +430,24 @@ trait ParserMain
 		// The x Completions.empty removes any trailing token completions where append.isEmpty
 		apply(p)(s).completions(level) x Completions.empty
 
-	def examples[A](a: Parser[A], completions: Set[String], check: Boolean = false): Parser[A] =
```
jsuereth, line 421: In other bin-compat failures, you need to keep this method around and delegate to the new one.

util/complete/src/main/scala/sbt/complete/Parser.scala
```diff
@@ -418,14 +430,24 @@ trait ParserMain
...
+	 *          only [[Parser.completions]] is modified.
+	 * @param completions the source of examples when displaying completions to the user.
+	 * @param maxNumberOfExamples limits the number of examples that the source of examples should return. This can
+	 *                            prevent lengthy pauses and avoids bad interactive user experience.
+	 * @param removeInvalidExamples indicates whether completion examples should be checked for validity (against the given parser). An
+	 *              exception is thrown if the example source contains no valid completion suggestions.
+	 * @tparam A the type of values that are returned by the parser.
+	 * @return
+	 */
+	def examples[A](a: Parser[A], completions: ExampleSource, maxNumberOfExamples: Int, removeInvalidExamples: Boolean): Parser[A] =
```
eed3si9n, line 454: bincompat violation. To make sure all the plugins published against sbt 0.13.0 works against 0.13.5, please provide an override def example with the same signature as the old one.

jsuereth:
New bot we need:

urbas:
Oh noes! Sad times.

Okay, officer kitteh... I'll scrubs the bugs away now

jsuereth:
Ok, this definitely LGTM now, cc> @eed3si9n for one last check.

eed3si9n:
I'm getting ready for 0.13.5-RC2 and the several tests are failing from file examples:

```
[info] listing all files in an absolute base directory should
[error] ! produce the entire base directory's contents
[error]     NoSuchElementException: None.get (Option.scala:313)
```

urbas:
I ran the test locally (a Linux machine) and they all passed. What
operating system do you use? Maybe the code/test is not OS-agnostic.

eed3si9n:
I'm using Mac.

urbas:
Could you please investigate the exact difference between
`fileExamples().toList` and `allRelativizedPaths`? Maybe just paste the
contents of the lists here?

eed3si9n:
``` scala
    "listing all files in an absolute base directory" should {
        "produce the entire base directory's contents" in new directoryStructure {
            println("fileExamples: " + fileExamples().toList.toString)
            try {
                println("allRelativizedPaths: " + allRelativizedPaths.toString)
            } catch {
                case e: Exception => e.printStackTrace
            }
            fileExamples().toList should containTheSameElementsAs(allRelativizedPaths)
        }
    }
```

This prints out:

```
fileExamples: List(bar, bazaar, far, foo, moo, far/bardir2, far/barfile2, far/fardir1, far/farfile1)
java.util.NoSuchElementException: None.get
    at scala.None$.get(Option.scala:313)
    at scala.None$.get(Option.scala:311)
    at sbt.complete.FileExamplesTest$directoryStructure$$anonfun$allRelativizedPaths$1.apply(FileExamplesTest.scala:68)
    at sbt.complete.FileExamplesTest$directoryStructure$$anonfun$allRelativizedPaths$1.apply(FileExamplesTest.scala:68)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
    at scala.collection.immutable.List.foreach(List.scala:318)
    at scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
    at scala.collection.AbstractTraversable.map(Traversable.scala:105)
    at sbt.complete.FileExamplesTest$directoryStructure.allRelativizedPaths(FileExamplesTest.scala:68)
    at sbt.complete.FileExamplesTest$$anonfun$1$$anonfun$apply$1$$anon$1$delayedInit$body.apply(FileExamplesTest.scala:16)
    at scala.Function0$class.apply$mcV$sp(Function0.scala:40)
    at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:12)
    at sbt.complete.FileExamplesTest$directoryStructure$$anonfun$delayedInit$1.apply(FileExamplesTest.scala:78)
    at sbt.complete.FileExamplesTest$directoryStructure$$anonfun$delayedInit$1.apply(FileExamplesTest.scala:75)
    at sbt.IO$.withTemporaryDirectory(IO.scala:285)
    at sbt.complete.FileExamplesTest$directoryStructure.delayedInit(FileExamplesTest.scala:74)
    at sbt.complete.FileExamplesTest$$anonfun$1$$anonfun$apply$1$$anon$1.<init>(FileExamplesTest.scala:13)
    at sbt.complete.FileExamplesTest$$anonfun$1$$anonfun$apply$1.apply(FileExamplesTest.scala:13)
    at sbt.complete.FileExamplesTest$$anonfun$1$$anonfun$apply$1.apply(FileExamplesTest.scala:13)
    at org.specs2.specification.FragmentsBuilder$$anon$1.asResult(FragmentsBuilder.scala:134)
    at org.specs2.specification.FragmentsBuilder$$anon$1.asResult(FragmentsBuilder.scala:133)
    at org.specs2.execute.AsResult$.apply(Result.scala:236)
    at org.specs2.specification.Example$$anonfun$apply$1.apply(Fragment.scala:141)
    at org.specs2.specification.Example$$anonfun$apply$1.apply(Fragment.scala:141)
    at org.specs2.specification.Example.execute(Fragment.scala:104)
    at org.specs2.specification.FragmentExecution$$anonfun$1.apply(FragmentExecution.scala:52)
    at org.specs2.specification.FragmentExecution$$anonfun$1.apply(FragmentExecution.scala:52)
    at org.specs2.execute.ResultExecution$class.execute(ResultExecution.scala:22)
    at org.specs2.execute.ResultExecution$.execute(ResultExecution.scala:76)
    at org.specs2.specification.FragmentExecution$class.executeBody(FragmentExecution.scala:28)
    at org.specs2.runner.TestInterfaceConsoleReporter.executeBody(TestInterfaceRunner.scala:123)
    at org.specs2.specification.FragmentExecution$class.execute(FragmentExecution.scala:52)
    at org.specs2.runner.TestInterfaceConsoleReporter.execute(TestInterfaceRunner.scala:123)
    at org.specs2.specification.FragmentExecution$$anonfun$executeFragment$1$$anonfun$apply$1.apply(FragmentExecution.scala:35)
    at org.specs2.specification.FragmentExecution$$anonfun$executeFragment$1$$anonfun$apply$1.apply(FragmentExecution.scala:35)
    at org.specs2.control.Exceptions$class.catchAllOr(Exceptions.scala:53)
    at org.specs2.control.Exceptions$.catchAllOr(Exceptions.scala:108)
    at org.specs2.specification.FragmentExecution$$anonfun$executeFragment$1.apply(FragmentExecution.scala:35)
    at org.specs2.specification.FragmentExecution$$anonfun$executeFragment$1.apply(FragmentExecution.scala:35)
    at org.specs2.reporter.DefaultExecutionStrategy$class.executeWithBarrier$1(ExecutionStrategy.scala:102)
    at org.specs2.reporter.DefaultExecutionStrategy$$anonfun$executeConcurrently$1$$anonfun$apply$2.apply(ExecutionStrategy.scala:104)
    at org.specs2.reporter.DefaultExecutionStrategy$$anonfun$executeConcurrently$1$$anonfun$apply$2.apply(ExecutionStrategy.scala:104)
    at org.specs2.internal.scalaz.concurrent.Promise$Unfulfilled$.fulfill(Promise.scala:100)
    at org.specs2.internal.scalaz.concurrent.Promise$Done.eval(Promise.scala:127)
    at org.specs2.internal.scalaz.concurrent.Promise$$anonfun$1.apply(Promise.scala:15)
    at org.specs2.internal.scalaz.concurrent.Promise$$anonfun$1.apply(Promise.scala:15)
    at org.specs2.internal.scalaz.concurrent.Actor$$anonfun$2.apply(Actor.scala:31)
    at org.specs2.internal.scalaz.concurrent.Actor$$anonfun$2.apply(Actor.scala:25)
    at org.specs2.internal.scalaz.concurrent.Effects$$anon$1$$anonfun$1.apply(Effect.scala:13)
    at org.specs2.internal.scalaz.concurrent.Effects$$anon$1$$anonfun$1.apply(Effect.scala:13)
    at org.specs2.internal.scalaz.concurrent.Effect$$anonfun$$bang$1.apply$mcV$sp(Effect.scala:8)
    at org.specs2.internal.scalaz.concurrent.Effect$$anonfun$$bang$1.apply(Effect.scala:8)
    at org.specs2.internal.scalaz.concurrent.Effect$$anonfun$$bang$1.apply(Effect.scala:8)
    at org.specs2.internal.scalaz.concurrent.StrategyLow$$anon$3$$anon$4.call(Strategy.scala:35)
    at java.util.concurrent.FutureTask.run(FutureTask.java:262)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
    at java.lang.Thread.run(Thread.java:744)
```

I also have other tests failing too in similar way.

```
[info] listing all files in an absolute base directory should
[error] ! produce the entire base directory's contents
[error]     NoSuchElementException: None.get (Option.scala:313)
[info]  
[info] listing files with a prefix that matches none should
[info] + produce an empty list
[info]  
[info] listing single-character prefixed files should
[error] ! produce matching paths only
[error]     NoSuchElementException: None.get (Option.scala:313)
[info]  
[info] listing directory-prefixed files should
[error] ! produce matching paths only
[error]     NoSuchElementException: None.get (Option.scala:313)
[error] ! produce sub-dir contents only when appending a file separator to the directory
[error]     NoSuchElementException: None.get (Option.scala:313)
[info]  
[info] listing files with a sub-path prefix should
[error] ! produce matching paths only
[error]     NoSuchElementException: None.get (Option.scala:313)
```

urbas:
Great, thanks, that explains a lot. `fileExamples` seems to return the
right thing. However, `sbt.IO.relativize(File, File)` seems to be returning
something else on Mac (compared to Linux; can somebody run it on Windows
too?).
1. The easy fix would be to not use `relativize` (and simply list the paths
   by hand). I can do that from here, but the tests would have to be executed
   on your machine (I do not own a Mac).
2. Another fix would be to check why `relativize` is returning something
   else on Mac and establish whether it's intended or a bug (this I cannot do
   from here, as I don't have a Mac). If that's intended behaviour of
   `relativize` or fixing it isn't an option, then we can simply go back to 1.

PR #1238, opened 2014-04-07
Handle macros that have themselves as original tree (Fix #1237)

Duhemm:
It has been reported in sbt/sbt#1237 that stack overflows may occur during the
extraction of used names (and later of dependencies between files). This
problem has been introduced by sbt/sbt#1163, which was about recording the
dependencies of macro arguments.

When a macro is expanded, the compiler attaches the tree before expansion to
the tree representing the expanded macro. As of Scala 2.11-RC3, some macros
have themselves attached as original tree, which caused the same macro to be
inspected over and over until a stack overflow.

Issue #1237, opened 2014-04-06
Incremental compiler StackOverflowError with name hashing


xuwei-k:
## steps

```
$ git clone git@github.com:xuwei-k/scalaz.git xuweik_scalaz
$ cd xuweik_scalaz
$ git checkout c27cd30
$ echo "sbt.version=0.13.2-RC1" > project/build.properties
$ sbt ++2.11.0-RC4 compile
```

or

> You can also reproduce it in scripted test `macro-arg-dep` (and most likely in any other test that uses macro). Simply add `scalaVersion := "2.11.0-RC3"` to the `build.scala`.
## problems

```
java.lang.StackOverflowError
    at scala.collection.TraversableLike$class.collect(TraversableLike.scala:282)
    at scala.collection.AbstractTraversable.collect(Traversable.scala:104)
    at xsbt.Compat$MacroExpansionOf$.unapply(Compat.scala:125)
    at xsbt.ExtractUsedNames.xsbt$ExtractUsedNames$$handleTreeNode$1(ExtractUsedNames.scala:86)
    at xsbt.ExtractUsedNames$$anonfun$handleMacroExpansion$1$1.apply(ExtractUsedNames.scala:58)
    at xsbt.ExtractUsedNames$$anonfun$handleMacroExpansion$1$1.apply(ExtractUsedNames.scala:58)
    at scala.reflect.internal.Trees$ForeachTreeTraverser.traverse(Trees.scala:1616)
    at scala.reflect.internal.Trees$TreeContextApiImpl.foreach(Trees.scala:106)
    at xsbt.ExtractUsedNames.handleMacroExpansion$1(ExtractUsedNames.scala:58)
    at xsbt.ExtractUsedNames.xsbt$ExtractUsedNames$$handleTreeNode$1(ExtractUsedNames.scala:88)
    at xsbt.ExtractUsedNames$$anonfun$handleMacroExpansion$1$1.apply(ExtractUsedNames.scala:58)
    at xsbt.ExtractUsedNames$$anonfun$handleMacroExpansion$1$1.apply(ExtractUsedNames.scala:58)
    at scala.reflect.internal.Trees$ForeachTreeTraverser.traverse(Trees.scala:1616)
    at scala.reflect.internal.Trees$TreeContextApiImpl.foreach(Trees.scala:106)
    at xsbt.ExtractUsedNames.handleMacroExpansion$1(ExtractUsedNames.scala:58)
[...]

PR #1239, opened 2014-04-07
Revert #1163

gkossakowski:
The #1237 reported a critical regression against sbt 0.13.2-RC1. We have a fix on its way in #1238 but I'd like us to better understand [whether](https://github.com/sbt/sbt/issues/1237#issuecomment-39718977) there's a bug in Scala 2.11 and whether the fix in #1237 is complete.

Therefore, I propose we revert #1163 (the PR that introduced improved macro handling) in sbt 0.13.2 to unblock the release. We'll fix the macro handling for good in 0.13.5 which is around the corner.


gkossakowski:
//cc @Duhemm, @xeno-by

gkossakowski:
Review by @jsuereth.

jsuereth:
LGTM. I think this is the right fix.  most likely 0.13.5 will be a few weeks after 0.13.2 final anyway.

jsuereth:
After TravisCI reports success, I'll merge and release. 

Thanks for staying on top of this @gkossakowski and @duhemm

gkossakowski:
@jsuereth: I'd like to draw your attention to 4cc5e4a. I hope I did the right thing there.

jsuereth:
Yeah, that's the right fix.  I think with 0.13.5 we're going to use sbt-git for release versioning off of tags rather than this gunk with these files all the time.

gkossakowski:
Ok. Great to hear that the process is being constantly improved!

jsuereth:
GO GO Travis CI!

retronym:
Even if we fix the bug in Scala 2.10.5 / 2.11.1, we'll still need the workaround in SBT, so I don't think we should rule out merging the other PR.

jsuereth:
@retronym Well, the macro-support will be delayed until 0.13.5 due to this (which is only a few weeks away now anyway.

gkossakowski:
@retronym: the behavior in 2.10.x is correct. It's regression in Scala 2.11.

jsuereth:
Ok, this passed the incremental compiler tests in travis. In the last few tests fail, we may blame the slow resolution issue facing all travis CI builds now.

PR #1244, opened 2014-04-09
Upgrade sbt plugins and version with sbt-git.

jsuereth:
- Version number now includes git SHA for snapshots, or the tag.
- Version base string now encoded in `version.sbt`
- Upgrade to latest sbt-site, sbt-git, sbt-ghpages.
- Show deprecation warnings when building sbt.

Review by @eed3si9n


eed3si9n:
Does that mean I have to check things in to test things locally?

jsuereth:
I think it failed because it's integrating against the broken trunk I caused before.  let me rebase/re-push

PR #1245, opened 2014-04-09
Fix infinite loop of death.

jsuereth:
Turns out travis CI was right about my PR causing nothing to run.

review by @eed3si9n and sorry for not having tested that last minute refactoring beyond compiling!


jsuereth:
Ok, merging before we see more failures.

gkossakowski:
There's another infinite loop that I found while testing against a very simple sbt project. I'm on the phone right now so I cannot file a ticket. I'll do it in 2 hours.

gkossakowski:
For future reference see #1251.

PR #1249, opened 2014-04-10
`DefaultOptions` for resolvers should not resets existing values.

indrajitr:
`DefaultOptions.addResolvers` and `DefaultOptions.addPluginResolvers`
should not reset the existing value while adding the new ones. The
names are prefixed with _add_ afterall.


jsuereth:
LGTM

PR #1252, opened 2014-04-10
Pick the right overload in EvaluateTask.runTask

### LINKED ISSUE #1251: Sbt hangs at loading global plugins
Sbt 0.13.5-SNAPSHOT hangs while loading global plugins. An empty project is enough to reproduce it:

```
Grzegorzs-MacBook-Pro:sbt-empty grek$ echo "sbt.version=0.13.5-SNAPSHOT" > project/build.properties
Grzegorzs-MacBook-Pro:sbt-empty grek$ sbt
[info] Loading global plugins from /Users/grek/.sbt/0.13/plugins
// hangs here
```

I attached the profiler and noticed that it hangs in `EvaluateTask.runTask` method. It recurses by accident: wrong overload is chosen. The fix is on its way.

gkossakowski:
The deprecated method should forward to the other overloaded alternative
but it recursed instead.

This kind of mistake would be easily caught by linter warning about
unused `newConfig` local variable. I hope we'll get there some day.

Fixes #1251

### DIFF
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -292,7 +292,7 @@ object EvaluateTask
     def runTask[T](root: Task[T], state: State, streams: Streams, triggers: Triggers[Task], config: EvaluateConfig)(implicit taskToNode: NodeView[Task]): (State, Result[T]) =
     {
     	val newConfig = EvaluateTaskConfig(config)
-    	runTask(root, state, streams, triggers, config)(taskToNode)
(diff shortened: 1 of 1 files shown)


gkossakowski:
Review by @jsuereth 

I'll leave improvements to test coverage to people more experienced with that part of sbt.

jsuereth:
Huh, do you have the call stack for this?  I turned on `@deprecated` warnings and was flooded with so much info I may have missed that call!.

In any case, thanks for the fix. I'm merging now.

gkossakowski:
Here's the stacktrace:

evaluatetask-hang

PR #1253, opened 2014-04-10
FIx configuration ordering issues in sbt 0.13.5

jsuereth:
The autoplugin configurations represent the new "default project" so they must come first.   The previous mechanism of specifying configurations can override autoplugins, but must be injected _after_ the default configurations.

Review by @eed3si9n or @gkossakowski cc> @jroper

### DIFF
--- a/main/src/main/scala/sbt/Project.scala
+++ b/main/src/main/scala/sbt/Project.scala
@@ -109,6 +109,11 @@ sealed trait Project extends ProjectDefinition[ProjectReference]
 	/** Adds configurations to this project.  Added configurations replace existing configurations with the same name.*/
 	def overrideConfigs(cs: Configuration*): Project = copy(configurations = Defaults.overrideConfigs(cs : _*)(configurations))
 
+    /** Adds configuration at the *start* of the configuration list for this rpoject.  Prevous configurations replace this prefix
+     * list with the same name. 
+     */
+    def prefixConfigs(cs: Configuration*): Project = copy(configurations = Defaults.overrideConfigs(configurations : _*)(cs))
+
 	/** Adds new configurations directly to this project.  To override an existing configuration, use `overrideConfigs`. */
 	def configs(cs: Configuration*): Project = copy(configurations = configurations ++ cs)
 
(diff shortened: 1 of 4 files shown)


sbt/src/sbt-test/actions/configuration-delegation/build.sbt
```diff
@@ -0,0 +1,17 @@
...
+
+def makeFoo(config: Configuration): Setting[_] = 
+  foo in config := IO.write(file(s"${config.name}-foo"), "foo")
+
+lazy val PerformanceTest = (config("pt") extend Test)
+
+lazy val root = (
+  (project in file("."))
+  .configs(PerformanceTest)
+  .settings(Seq(Compile, Test, Runtime, PerformanceTest).map(makeFoo) :_*)
```
gkossakowski, line 13: Why do you need `foo` in all those configurations when you test the compile configuration only?

jsuereth, line 13: The issue before was that the default configurations (Compile/Test/Runtime) were coming _after_ PerformanceTest.  So if you typed `foo` on the command line, then the PerformanceTest version was run instead of `Compile` as expected.

I should make this more rigorous with a task in just Test/Runtime/PT and Runtime/PT to make sure the _default_ order is correct and delegation handled appropriately.

main/src/main/scala/sbt/Project.scala
```diff
@@ -109,6 +109,11 @@ sealed trait Project extends ProjectDefinition[ProjectReference]
 	/** Adds configurations to this project.  Added configurations replace existing configurations with the same name.*/
 	def overrideConfigs(cs: Configuration*): Project = copy(configurations = Defaults.overrideConfigs(cs : _*)(configurations))
 
+    /** Adds configuration at the *start* of the configuration list for this rpoject.  Prevous configurations replace this prefix
+     * list with the same name. 
+     */
+    def prefixConfigs(cs: Configuration*): Project = copy(configurations = Defaults.overrideConfigs(configurations : _*)(cs))
```
eed3si9n: Would `prependConfigs` be a better name?

jsuereth: perhaps.  TBH we may want to evaluate this a bit more as well.  I just wanted a fix in place for 0.13.5.  Going to add a card to do a quick survey of AutoPlugins vs. Defaults to see if any more residual ordering things remain.

jsuereth: Oh, I can mark it as `private[sbt]`

Issue #1184, opened 2014-03-16
ClassfileManager crashes when name hashing is enabled


gkossakowski:
@aloiscochard reported on sbt-dev a crash in ClassfileManager when name hashing is enabled:

```
java.lang.IllegalArgumentException: requirement failed: Source file '/home/alois/timeout/jvm/target/classes.bak/sbt3697008749647665633.class' does not exist.
        at scala.Predef$.require(Predef.scala:233)
        at sbt.IO$.copyFile(IO.scala:580)
        at sbt.IO$.move(IO.scala:764)
        at sbt.inc.ClassfileManager$$anonfun$transactional$1$$anon$2$$anonfun$complete$2.apply(ClassfileManager.scala:53)
        at sbt.inc.ClassfileManager$$anonfun$transactional$1$$anon$2$$anonfun$complete$2.apply(ClassfileManager.scala:53)
        at scala.collection.TraversableLike$WithFilter$$anonfun$foreach$1.apply(TraversableLike.scala:772)
        at scala.collection.mutable.HashMap$$anonfun$foreach$1.apply(HashMap.scala:98)
        at scala.collection.mutable.HashMap$$anonfun$foreach$1.apply(HashMap.scala:98)
        at scala.collection.mutable.HashTable$class.foreachEntry(HashTable.scala:226)
        at scala.collection.mutable.HashMap.foreachEntry(HashMap.scala:39)
        at scala.collection.mutable.HashMap.foreach(HashMap.scala:98)
        at scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:771)
        at sbt.inc.ClassfileManager$$anonfun$transactional$1$$anon$2.complete(ClassfileManager.scala:53)
        at sbt.inc.Incremental$.manageClassfiles(Incremental.scala:66)
        at sbt.inc.Incremental$.compile(Incremental.scala:37)
```

The full compilation log is available here: http://pastie.org/8916318

Unfortunately, we don't log activity of ClassfileManager so it's hard to tell why the class file is missing. Also, we do not log the name of the class the ClassfileManager tried to restore. It's hard to tell whether it was the last compilation iteration or earlier compilation that caused removal of the class file.

aloiscochard:
@gkossakowski next time it would be nice if you could update that ticket if you expect me to do an action (like giving more debug information).

I'm not following sbt itself but just that ticket, so there was no way I would have guess you did that change without -by random chance- got notified on twitter.

[...]

PR #1254, opened 2014-04-10
Add debug logging in ClassfileManager

gkossakowski:
Add logging of various operations the transactional class file manager is
doing. You can pass logger to be used by the transactional class file
manager by using overloaded definition of `ClassfileManager.transactional`
method. The old overload has been deprecated.

The factory methods for class file manager in IncOptions companion object
has been deprecated in favor of using ClassfileManager companion object
directly. The code in Defaults.scala has been updated to use non-deprecated
methods.

The canonical way of enabling transactional class file manager in sbt
project is:

``` scala
incOptions := incOptions.value.withNewClassfileManager(
  sbt.inc.ClassfileManager.transactional(
    crossTarget.value / "classes.bak",
    (streams in (Compile, compile)).value.log
  )
)
```

It's a bit verbose which shows that the api for this is not the best.
However, I don't expect sbt users to need this code very often.

This patch should help debug the problem described in #1184

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -215,7 +215,8 @@ object Defaults extends BuildCommon
 	)
 
 	def compileBase = inTask(console)(compilersSetting :: Nil) ++ compileBaseGlobal ++ Seq(
-		incOptions := IncOptions.setTransactional(incOptions.value, crossTarget.value / "classes.bak"),
+		incOptions := incOptions.value.withNewClassfileManager(
+			sbt.inc.ClassfileManager.transactional(crossTarget.value / "classes.bak", sbt.Logger.Null)),
 		scalaInstance <<= scalaInstanceTask,
 		crossVersion := (if(crossPaths.value) CrossVersion.binary else CrossVersion.Disabled),
 		crossTarget := makeCrossTarget(target.value, scalaBinaryVersion.value, sbtBinaryVersion.value, sbtPlugin.value, crossPaths.value)
--- a/compile/inc/src/main/scala/sbt/inc/ClassfileManager.scala
+++ b/compile/inc/src/main/scala/sbt/inc/ClassfileManager.scala
@@ -13,7 +13,7 @@ trait ClassfileManager
 	* Any empty ancestor directories of deleted files must not exist either.*/
 	def delete(classes: Iterable[File]): Unit
 
-	/** Called once per compilation step with the class files generated during that step.*/	
+	/** Called once per compilation step with the class files generated during that step.*/
 	def generated(classes: Iterable[File]): Unit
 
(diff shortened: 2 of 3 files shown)


gkossakowski:
Review by @jsuereth

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -215,7 +215,8 @@ object Defaults extends BuildCommon
 	)
 
 	def compileBase = inTask(console)(compilersSetting :: Nil) ++ compileBaseGlobal ++ Seq(
-		incOptions := IncOptions.setTransactional(incOptions.value, crossTarget.value / "classes.bak"),
+		incOptions := incOptions.value.withNewClassfileManager(
+			sbt.inc.ClassfileManager.transactional(crossTarget.value / "classes.bak", sbt.Logger.Null)),
```
jsuereth, line 219: should this one be `streams.value.log` instead of `sbt.Logger.Null` ?

jsuereth:
your example may want to show:

``` scala
(streams in (compile, Compile)).value.log
```

Or some such variant if you want the log to align with the `compile in Compile` task.   The stream-logging stuff is a bit... fun right now, and needs some cleanup.  But just curious if the output locations is:

a. where you want it
b. in the same file as the `compile in Compile` logs.

jsuereth:
Oh, otherwise LGTM

main/src/main/scala/sbt/Defaults.scala
gkossakowski, line 219: Logging is opt-in. I didn't want to clutter the debug log too much. Most of the time you don't really need to know what ClassfileManager does even if you are debugging incremental compiler problems.

gkossakowski:
Clarified in commit message that logging is off by default and fixed the code snippet according to your suggestion.

I agree that logging in sbt could receive some love.

aloiscochard:
@gkossakowski the instruction you provided did not work.

I had to do:

```
    incOptions := incOptions.value.withNewClassfileManager(
      sbt.inc.ClassfileManager.transactional(
        crossTarget.value / "classes.bak",
        (streams in Compile).value.log
      )
    )
```

I prefer not to know why it should be like that, but I thought it was worth mentioning if anyone else try to use this.

gkossakowski:
@jsuereth: any ideas on why the original snippet would not work?

jsuereth:
@gkossakowski I sent you an email about it elsewhere.   You have the wrong order for `in` so you confuse the hell out of scala looking for implicits on the overloaded method.

Your code _should_ have been:

```
incOptions := incOptions.value.withNewClassfileManager(
  sbt.inc.ClassfileManager.transactional(
    crossTarget.value / "classes.bak",
    (streams in (Compile, compile)).value.log
  )
)
```

See me note above.

gkossakowski:
@jsuereth: I updated the description in this issue. Thanks.

PR #1266, opened 2014-04-16
Use the correct state in buildPluginDefinition()

cunei:
Within buildPluginDefinition(), the call to setProject() can
(and usually will) return a modified structure together with
the new state.

The subsequent call to evalPluginDef() should use the updated
structure, rather the old stucture that was present before
the setProject() ("pluginDef"); if that is not the case,
the code called by evalPluginDef() will find an inconsistent
structure/state combination, and behave in bizarre ways as
a result.

More in general, it is a bit dangerous to pass to routines
in parallel the two separate state and structure, as the
two may easily inadvertently fall out of alignment, as in this
case.

This patch should be applied to both the 0.13 branch as well
as to a future 0.12.5 release (the corresponding file there
is ./main/Load.scala).

### DIFF
--- a/main/src/main/scala/sbt/Load.scala
+++ b/main/src/main/scala/sbt/Load.scala
@@ -612,7 +612,7 @@ object Load
 	{
 		val (eval,pluginDef) = apply(dir, s, config)
 		val pluginState = Project.setProject(Load.initialSession(pluginDef, eval), pluginDef, s)
-		config.evalPluginDef(pluginDef, pluginState)
+		config.evalPluginDef(Project.structure(pluginState), pluginState)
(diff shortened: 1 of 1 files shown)


jsuereth:
This LGTM.  NOTE to whoever merges (after TravisCI finished) we need to backport to 0.12 branch.

jsuereth:
@cunei / @eed3si9n I'm running a test build of this patch against a new 0.12 branch from 0.12.4.  WE have no travis CI for this version of sbt, so I'm doing it lcoally.  Will push the new 0.12 branch for 0.12.5 once it's done

PR #1276, opened 2014-04-21
Automatically run GC/finalization after each task execution.

### LINKED ISSUE #1223: Reducing PermGen usage in sbt

jsuereth:
Fixes #1223.
- Add a new key to disable forcing the garbage collector to run
  after each task-executioin
- Add a new flag to disable forcing the garbage collector to run
  after each task-exeuction
- Add a hook into EvalauteTask to run System.gc/System.runFinalization
  after each task execution

Review by @eed3si9n

### DIFF
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -80,8 +80,16 @@ sealed trait EvaluateTaskConfig {
 	def checkCycles: Boolean
 	def progressReporter: ExecuteProgress[Task]
 	def cancelStrategy: TaskCancellationStrategy
+	/** If true, we force a finalizer/gc run (or two) after task execution completes.
+	 * This helps in instances where 
+	 */
+	def forceGarbageCollection: Boolean
 }
 final object EvaluateTaskConfig {
+	// Returns the default force garbage collection flag,
+	// as specified by system properties.
+	private[sbt] def defaultForceGarbageCollection: Boolean =
+	  sys.props.get("sbt.task.forcegc").map(java.lang.Boolean.parseBoolean).getOrElse(true)
 	/** Pulls in the old configuration format. */
 	def apply(old: EvaluateConfig): EvaluateTaskConfig = {
 	  object AdaptedTaskConfig extends EvaluateTaskConfig {
@@ -91,6 +99,7 @@ final object EvaluateTaskConfig {
         def cancelStrategy: TaskCancellationStrategy =
           if(old.cancelable) TaskCancellationStrategy.Signal
           else TaskCancellationStrategy.Null
+        def forceGarbageCollection = defaultForceGarbageCollection
 	  }
 	  AdaptedTaskConfig
 	}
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/EvaluateTask.scala
```diff
@@ -198,6 +211,9 @@ object EvaluateTask
 		val maker: State => Keys.TaskProgress = getSetting(Keys.executeProgress, const(new Keys.TaskProgress(defaultProgress)), extracted, structure)
 		maker(state).progress
    }
+   // TODO - Should this pull from Global or from the project itself?
+   private[sbt] def forcegc(extracted: Extracted, structure: BuildStructure): Boolean =
+      getSetting(Keys.forcegc in Global, EvaluateTaskConfig.defaultForceGarbageCollection, extracted, structure)
```
eed3si9n, line 216: Can it be scoped to the root task so it could be overridden at the task level? Not sure if that's useful.

main/src/main/scala/sbt/EvaluateTask.scala
```diff
@@ -301,8 +317,21 @@ object EvaluateTask
 		val log = state.log
 		log.debug("Running task... Cancel: " + config.cancelStrategy + ", check cycles: " + config.checkCycles)
 		val tags = tagged[Task[_]](_.info get tagsKey getOrElse Map.empty, Tags.predicate(config.restrictions))
-		val (service, shutdown) = completionService[Task[_], Completed](tags, (s: String) => log.warn(s))
-
+		val (service, shutdownThreads) = completionService[Task[_], Completed](tags, (s: String) => log.warn(s))
+
+		def shutdown(): Unit = {
+		  // First ensure that all threads are stopped for task execution.
+		  shutdownThreads()
```
eed3si9n, line 324: When did threads come into the picture? Is this part of the GC deal?

jsuereth, line 324: completionService is an Executor service.  They were always in the picture, just trying to name correctly what the "hook" returned does, so it's clear.

main/src/main/scala/sbt/EvaluateTask.scala
jsuereth, line 216: No.  There can be multiple roots as of sbt 0.13.2 _and_ you're not guaranteed to have a key relating to the task you run.  

Also, I don't think it's particularly useful, but a good suggestion :)

eed3si9n:
LGTM

main/src/main/scala/sbt/EvaluateTask.scala
```diff
@@ -301,8 +317,21 @@ object EvaluateTask
...
+		  // Now we run the gc cleanup to force finalizers to clear out file handles (yay GC!)
+		  if(config.forceGarbageCollection) {
+		  	// Force the detection of finalizers for scala.reflect weakhashsets
+ 			System.gc()
+ 			// Force finalizers to run.
+ 			System.runFinalization()
+ 			// Force actually cleaning the weak hash maps.
+ 			System.gc()
```
viktorklang, line 332: Just FYI: All of these are advisory and are no-ops on some JVMs.

PR #1278, opened 2014-04-22
Migrate JUnitXmlReporter into sbt as autoplugin

jsuereth:
Fragmented from #1112 

This adds a `JUnitXmlReportListener` to the sbt testing library which can generate junit-xml reports using the test-listener interface.   It also creates a plugin that can be enabled to add the listener to sbt builds.

Review by @eed3si9n  cc @backuitist


eed3si9n:
The build failure seems unrelated, so I'm merging this in.

PR #1291, opened 2014-04-26
Def.setting needed for externalIvyFile

jaceklaskowski:
Sorry for the recent pull request that introduced an error since `externalIvyFile` expects `sbt.Def.Initialize[java.io.File]` rather than `java.io.File` itself.


jsuereth:
Hmm, I feel we could probably make that nicer, but still LGTM. Thanks for the fix!

PR #1295, opened 2014-04-28
add sbt-one-log into Community Plugins

zavakid:
add a sbt plugin : sbt-one-log link into the community plugins.


jsuereth:
Thanks for the patch!  Note: the 0.13.2 branch hits the doc site immediately, but 0.13.5-RC1 will come out wed, so I can merge into 0.13. which will hit the doc site sometime after wed, or you can retarget 0.13.2 if you want it out immediately.

zavakid:
thanks for the reply!
it's OK, to merge into 0.13, thanks!

PR #1296, opened 2014-04-28
Revert "add precompiled 2.11.0"

jsuereth:
This reverts commit 36ec1032741874aec9ce97c4d886c3208af7ce5a.

This 2.11.0 dependency was causing issues with older sbt launchers
attempting to use the 2.11.0 scala release for sbt's classloader,
leading to binary incompatibility issues.

Review by @eed3si9n


jsuereth:
cc> @xuwei-k An unexpected side effect, but just so you know.  We'll add the precompiled stuff after we figure out the launcher issue.   You can target the 1.0 branch as well, allthough we'll be merging into that one soon.

PR #1304, opened 2014-05-01
scalariform

eed3si9n:
## what this changes

the code is now formatted automatically for source and build files.
review @jsuereth


jsuereth:
LGTM, but I do want to wait for travis on this one.

eed3si9n:
travis says:

> I'm sorry but your test run exceeded 50.0 minutes.

jsuereth:
Yeha, so dependency management + source-dependencies tests are on the edge of running under 50 minutes usually.  We should probably fragment them down into two halves so we can fit it under the travis window (or start paying for travis).

In any case, LGTM.

Issue #1224, opened 2014-03-29
Add a SettingKey to define the initial project on (re)load


jaceklaskowski:
From https://groups.google.com/d/msg/sbt-dev/dB1TmFN8pvw/4JF_3CzPAJYJ:

_"In either case, feel free to open an enhancement ticket about this if it's causing signfiicant issues.  We can probably create a setting the `reload` command will read when setting the initial project, rather than always using "root"."_

jaceklaskowski:
Submitted https://github.com/sbt/sbt/pull/1305 to get this going. Please review and guide me to make it complete. I call myself stuck on trying to understand the sbt sources and the way projects are loaded.

dwijnand:
Copy (and tweaked) workaround from #2405:

```scala
onLoad in Global ~= (_ andThen ("project foo" :: _))
```

jaceklaskowski:
It could've been easier, but the workaround by @dwijnand makes a lot of sense to me and so I'm closing  the issue.

dwijnand:
I agree it could and should be easier. Do you mind if we use this issue to track the addition of a more formal way to define the initial project?

PR #1305, opened 2014-05-01
Reload default project

jaceklaskowski:
It's my attempt at #1224 - failed unfortunately and I badly need help. Please advise how to turn the debug-heavy code into something usable. I can't seem to find a way to return to the main directory of a multi-project build and select one (as if `project [projectName]` were called).


main/src/main/scala/sbt/Keys.scala
```diff
@@ -53,6 +53,7 @@ object Keys
 	val onUnload = SettingKey[State => State]("on-unload", "Transformation to apply to the build state when the build is unloaded.", DSetting)
 	val onLoadMessage = SettingKey[String]("on-load-message", "Message to display when the project is loaded.", DSetting)
 	val transformState = AttributeKey[State => State]("transform-state", "State transformation to apply after tasks run.", DSetting)
+  val defaultProject = SettingKey[Option[Project]]("default-project", "Default project reload will use as the project to return to.", BSetting)
```
jsuereth: this should be a `ProjectRef`

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -150,7 +150,8 @@ object Defaults extends BuildCommon
 		aggregate :== true,
 		maxErrors :== 100,
 		fork :== false,
-		initialize :== {}
+		initialize :== {},
+    defaultProject := None
```
jsuereth, line 153: We can inject this if it doesn't exist using our existing "default project" rules, or you can just leave it unspecified.  You'll have to read it when creating the initial `BuildStructure` for default project, so you can handle existence/absence in your own code (similar to how we handle `logLevel`)

main/src/main/scala/sbt/Project.scala
```diff
@@ -492,13 +493,33 @@ object Project extends ProjectExtra
...
 			projectReturn(s) match { case Nil => (setProjectReturn(s, base :: Nil), base); case x :: xs => (s, x) }
-		case Plugins =>
+		case LoadAction.Plugins =>
 			val (newBase, oldStack) = if(Project.isProjectLoaded(s))
 					(Project.extract(s).currentUnit.unit.plugins.base, projectReturn(s))
 				else // support changing to the definition project if it fails to load
 					(BuildPaths.projectStandard(s.baseDir), s.baseDir :: Nil)
 			val newS = setProjectReturn(s, newBase :: oldStack)
 			(newS, newBase)
+    case DefaultProject =>
```
jsuereth, line 522: I think this may be the wrong approach.  When we "load" a project, we should set the `currentProject` to whichever matche sthe ref it exists.

jsuereth:
The approach of having the setting in the build is cool.  However, you may need to attack the generation of BuildStructure when we load a project. You'd want to look at the `project` command.

If we have a default project setting, it should be the default project when we first load.

In fact, what should happen is we deprecate this (or just expose it in .sbt files): https://github.com/sbt/sbt/blob/0.13/main/src/main/scala/sbt/Build.scala#L19-L20

I may be reworking the DSL a bit for 0.13.6, so it could be I can expose the hooks you need in that change.  Otherwise, you need to attack a bit deeper to read your setting.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -906,6 +907,7 @@ object Defaults extends BuildCommon {
   lazy val coreDefaultSettings: Seq[Setting[_]] = projectCore ++ disableAggregation
   @deprecated("0.13.2", "Default settings split into `coreDefaultSettings` and IvyModule/JvmModule plugins.")
   lazy val defaultSettings: Seq[Setting[_]] = projectBaseSettings ++ defaultConfigs
+>>>>>>> aa4ee809e59ee119d14684bdde34d2f1e5958510
```
jsuereth, line 910: Artifact of merging

main/src/main/scala/sbt/Project.scala
```diff
@@ -191,6 +191,404 @@ sealed trait ClasspathDep[PR <: ProjectReference] { def project: PR; def configu
 final case class ResolvedClasspathDependency(project: ProjectRef, configuration: Option[String]) extends ClasspathDep[ProjectRef]
 final case class ClasspathDependency(project: ProjectReference, configuration: Option[String]) extends ClasspathDep[ProjectReference]
 
+<<<<<<< HEAD
```
jsuereth, line 194: Merge artifact.

typesafe-tools:

jsuereth:
Closing from inactivity.

PR #1307, opened 2014-05-02
remove unnecessary var

xuwei-k:


jsuereth:
thanks for the fix! LGTM

PR #1308, opened 2014-05-02
some more source getting formatted

eed3si9n:
scalariform is catching more source to format, which for some reason it didn't catch before.


jsuereth:
Looks good to me

Issue #1299, opened 2014-04-30
addPlugins does not work when used in a build.sbt file located in a subdirectory


julienrf:
I have the following structure:

```
root/
+- project-a/
    +- build.sbt
+- project-b/
    +- build.sbt
+- build.sbt
```

The `root/build.sbt` file just aggregates the sub projects:

``` scala
lazy val `project-a` = project

lazy val `project-b` = project
```

And, in both `root/project-a/build.sbt` and `root/project-b/build.sbt` I use `addPlugins`:

``` scala
lazy val root = project.in(file(".")).addPlugins(Foo)
```

If I run sbt from the `root/` directory the sub-projects are loaded but they behave is if the `addPlugins` were missing.

If I run sbt from the `root/project-a/` or `root/project-b/` directory everything works as expected.

jsuereth:
This is a fundamental limitation in our loading system right now.  We do not merge "instances of the project class" so the first one we see wins, in this case the one defined in root/build.sbt.   The current workaround is to define the addPlugins in that build.sbt file.

jsuereth:
Also, sbt should be detecting this.  Which version of sbt are you using?  In 0.13.5-M4 I see:

```
java.lang.RuntimeException: Overlapping output directories:/home/jsuereth/projects/sbt/diagnose/auto-plugin-issue/project-a/target:
    ProjectRef(file:/home/jsuereth/projects/sbt/diagnose/auto-plugin-issue/,root)
    ProjectRef(file:/home/jsuereth/projects/sbt/diagnose/auto-plugin-issue/,project-a)
```

jsuereth:
Oh, also one final thing:  WIth sbt-multi-project you need _always_ run from the root directory.  Sbt does not support the maven-style mechanism of cd-ing into a subdirectory and building.  Internal to the sbt console you can swap the active-project via the `project` command.

julienrf:
Yeah, I’m using 0.13.5-M2.

> WIth sbt-multi-project you need _always_ run from the root directory

I’d like to! So, by adding the `addPlugins` thing in the root `build.sbt` file it is ok. But, since my root project does nothing more than aggregating the subprojects I hoped it was also possible to also run the projects from their directories… I use the root project only to run the tests of both sub-projects.

jsuereth:
So, you should see:  #1213 .    We plan to resolve this issue and that one at the same time.

Basically, inside the lower build.sbt you'll be able to do manipulations to the project object inline, like:

``` scala
[...]

PR #1312, opened 2014-05-06
Expand the DSL abilities with Project manipulations

### LINKED ISSUE #1213: Circumventing the common declaration of addPlugins
Given the proposed AutoPlugin mechanism where plugins are required to be added, it'd be nice to have a short cut to the following common expressions in build.sbt:

``` scala
import com.typesafe.sbt.web.SbtWeb

lazy val root = project.in(file(".")).addPlugins(SbtWeb)
```

For example, if the project is a single module then may be sbt's processing of build.sbt can be magical and automatically perform the above. Why else would you call `addSbtPlugin` from `plugins.sbt` if your single module project wasn't going to use it?

Further to the above, and orthogonal, perhaps a short cut to the expression could be provided e.g.:

``` scala
lazy val root = project.current.addPlugins(SbtWeb)
```

My major concern though is that all build.sbt files will have at least 3 extra lines over what they used to have. It isn't a massive concern, but if we are able to address it then I think it'll make a good UX.

jsuereth:
This does the following:
- Split loading of configuration into two pieces: Discovery and Resolution
- Discovery is when we find all projects defined in a config file _and_ load all Setting/Manipulation/Definitions
- Resolution is when we take _all_ the .sbt + .scala files associated with a project, do the AutoPlugin resolution and finalize the `Project` instance
- Migrates the DSL of `.sbt` files to use `sbt.internals.DslEntry`.
- Adds `ProjectManipulation` as a DSL option instead of just `Seq[Setting[_]]`
- Applies project manipulation functions before finalizing projects
- Add `enablePlugin` and `disablePlugin` to the DSL.
- Update tests.

Note: @benburton informs me we should upgrade to using Fiber instead of a DSL.

Fixes #1213 and #1299

Also, I expect this one may take some time for review, but we should still be prompt.   Review by @eed3si9n cc> @cunei


jsuereth:
Oh, also still waiting on full test suite.  DO NOT MERGE until we have two verified computers with success.  This one is a too big not to be cautious if Travis fails.

jsuereth:
Note on failures:

```
[trace] Stack trace suppressed: run last root/*:scripted for the full output.
[error] (root/*:scripted) Failed tests:
[error]     project / derived
[error] Total time: 5342 s, completed May 6, 2014 6:27:19 PM
```

I'll have to investigate.  Maybe @dansanduleac has ideas on why we're failing now.

jsuereth:
Ok, ignore the derived setting test failures, they're failing against _MASTER_ for me, so I think this PR is good to go. I'm going to investigate why we lost "success" for the derived settings tests.  Looks like a commit is missing or something.

eed3si9n:
A scripted test combined with auto plugin would be interesting.

In `a.sbt`:

``` scala
enablePlugins(FooPlugin) // Suppose FooPlugin requires BarPlugin, and bring in setting for foo

check := {
  println(foo.value) // foo.value fails here?
}
```

In `b.sbt`:

``` scala
disablePlugins(BarPlugin) 
```

eed3si9n:
Also what's the ordering if there is `c.sbt`?

``` scala
val root = (project in file(".")).enablePlugins(BarPlugin)
```

Add docs for this behavior?

jsuereth:
@eed3si9n BarPlugin is enabled first if you define a root project there, an oddity of the implementation.

Basically, here's your order:
1. `Project` instances are loaded first
2. .sbt file manipulations in alphabentical order (or whatever SettingSets defines them to be)

Not a ton we can do about that at this point, without some MAJOR refactorings to the DSL.

eed3si9n:
Is there a document that describes this DSL shuffling? To be clear, `enablePlugins(FooPlugin)` in `a.sbt` is applied to the same instance defined in `c.sbt` as `root` correct?

jsuereth:
It's the same shuffling that occurs with regular-old settings in build.sbt, so by "document" I think the way we already describe build.sbt and multi-modules has documentation on this.  If not, then we're lacking in general around load and _now_ (with more clearly commented code) we could improve those docs if needed

jsuereth:
Oh, and the answer to the question is yes.

jsuereth:
@eed3si9n Check out  f579e90 for the test showcasing what you asked for.

jsuereth:
So it looks like besides derived setting failures, travisCI is passing.  Giving it one more go with better memory settings this time.

jsuereth:
Ok, so enough tests passed you should be able to review/merge now.

jsuereth:
COOL!  Travis passes everything but `project/derived` which is failing in 0.13 branch anyway.

PR #1313, opened 2014-05-07
Sort setting key names in the inspect tree view

2m:
This little add-on helps quite a lot when comparing two dependency trees, i.e. after and before a change in the build.

### DIFF
--- a/main/src/main/scala/sbt/SettingGraph.scala
+++ b/main/src/main/scala/sbt/SettingGraph.scala
@@ -45,7 +45,7 @@ case class SettingGraph(name: String,
     } getOrElse { "" }
 
   def dependsAscii: String = Graph.toAscii(this,
-    (x: SettingGraph) => x.depends.toSeq,
+    (x: SettingGraph) => x.depends.toSeq.sortBy(_.name),
     (x: SettingGraph) => "%s = %s" format (x.definedIn getOrElse { "" }, x.dataString))
 }
 


jsuereth:
Thanks for the submission!  This LGTM once tests report back success or timeout.

2m:
Yea. I tried to run tests on my machine, but it seems to hang at some places.

jsuereth:
`scripted` timed out or `test`?   We have a legitimate deadlock somewhere in the process code, but I can't actually remove the commit which introduced it due to binary compatibility restrictions and I haven't been able to fix it. (Note:  Don't use #> or <# in the process lib)

2m:
It would just stop running tests after a while, and would not drop back to sbt console.

```
[info] + Locks.Lock in nonexisting directory: OK, passed 100 tests.
[info] + Locks.Uncontested re-entrant lock: OK, passed 100 tests.
[info] + Locks.Uncontested double lock: OK, passed 100 tests.
[info] + Locks.Contested single lock: OK, passed 100 tests.
[info] Passed: Total 56, Failed 0, Errors 0, Passed 56
^C
```

Another try.

```
[info] Total for specification ScalaProviderTest
[info] Finished in 2 ms
[info] 9 examples, 0 failure, 0 error
[info] 
[info] Passed: Total 56, Failed 0, Errors 0, Passed 56
^C
```

jsuereth:
Right.  It's the `scripted` test which matter, generally.  And after the lock tests is the Process test, which has the deadlock.  In any case, Travis is reporting the usual timeouts I see, so this is good to go.

PR #1316, opened 2014-05-07
This doesn't quite fix the build, but gets closer

jsuereth:
- Expand travisCI to also run unit tests (except io library that currently has a deadlock)
- Fragment scripted tests into "pages" to keep long-running tests within the travisCI timeout
- Apply scalariform reformating to test code (missed previously).

Oh, review by @eed3si9n


jsuereth:
Danger @eed3si9n, Danger ->  Travis hasn't even finished running!

PR #1317, opened 2014-05-08
Add Ant-style incremental compilation mode

gkossakowski:
This PR implements an Ant-style incremental compilation mode. This mode
emulates what Ant's scalac command does. It recompiles just changed source
files and does not perform any invalidation of dependencies.
This is a very naive mode of incremental compilation that very often leads
to broken binaries.

The Ant-style mode is being introduced because Scala team needs it for
migration of Scala compiler to sbt. The name hashing algorithm doesn't
work well with Scala compiler sources due to deep inheritance chains.
There's a plan to refactor compiler's code to use more composition instead
of inheritance.

Once Scala compiler sources are refactored to work well with name hashing
algorithm, Ant-style mode will be deleted immediately.


gkossakowski:
@jsuereth: the failure in Travis seems to be completely unrelated to my changes. Can you have a look at it?

jsuereth:
Yeah, `project/derived` is known to have issues (found out earlier this week).  Dan was looking into that (based on his recent change).  I'll be backing out the change which caused the breakage monday if it's not fixed by then.

compile/inc/src/main/scala/sbt/inc/IncOptions.scala
```diff
@@ -65,7 +65,22 @@ final class IncOptions(
      *   3. Hashing of public names is enabled. See `sbt.inc.AnalysisCallback` for details.
      *
      */
-    val nameHashing: Boolean) extends Product with Serializable {
+    val nameHashing: Boolean,
+    /**
+     * THE `antStyle` OPTION IS UNSUPPORTED, MAY GO AWAY AT ANY POINT.
```
jsuereth, line 70: :)  `private[sbt]` is the way to make it really evil to use.  Then just put an "exposer" in your build, like:

`project/exposer.scala`

```
package sbt
object Exposer {
  def doSomethingEvil(): IncOptions =....
}
```

gkossakowski:
Ok. @jsuereth: are you ok reviewing this PR? We could also ask @eed3si9n to look at it as he probably haven't had chance to dive into incremental compiler yet.

compile/inc/src/main/scala/sbt/inc/IncOptions.scala
```diff
@@ -74,52 +89,60 @@ final class IncOptions(
...
+      apiDumpDirectory, newClassfileManager, recompileOnMacroDef, nameHashing, antStyle)
   }
 
   def withNameHashing(nameHashing: Boolean): IncOptions = {
     new IncOptions(transitiveStep, recompileAllFraction, relationsDebug, apiDebug, apiDiffContextSize,
-      apiDumpDirectory, newClassfileManager, recompileOnMacroDef, nameHashing)
+      apiDumpDirectory, newClassfileManager, recompileOnMacroDef, nameHashing, antStyle)
+  }
+
+  def withAntStyle(antStyle: Boolean): IncOptions = {
```
jsuereth, line 143: Ah, this would need `private[sbt]` as well.

compile/inc/src/main/scala/sbt/inc/IncOptions.scala
gkossakowski, line 70: Yes, that's possibility but I hope putting a fat warning in docs is enough. I hope this won't survive longer than a few months until we get our code refactored on Scala compiler side.

jsuereth:
So this looks good to me. the `private[sbt]` suggestion was just a possibility to ensure folks don't use it without an enforced warning.    In any case, good to merge!

PR #1318, opened 2014-05-09
Check for object autoImport in auto plugins. 

### LINKED ISSUE #1314: Auto import feature of auto plugins does not work with objects
# Steps
1. Create an `AutoPlugin` that defines a nested `autoImport` object: 
   
   ``` scala
   package foo
   
   import sbt._
   import Keys._
   
   object MyPlugin extends AutoPlugin {
   
    object autoImport {
      val scalatest = "org.scalatest" %% "scalatest" % "2.1.5"
    }
   }
   ```
2. Add that autoplugin to a new project and try to reference a member of `MyPlugin.autoImport` without explicitly importing it:
   `scala
   libraryDependencies += scalatest
   `
3. Observe SBT error:
   
   ``` scala
   [info] Done updating.
   /.../autoimport/tester/build.sbt:1: error: not found: value scalatest
   libraryDependencies += scalatest
                         ^
   ```
4. Extract an object and declare `autoImport` as a val that references that object:
   
   ``` scala
   package foo
   
   import sbt._
   import Keys._
   
   object MyPlugin extends AutoPlugin {
   
    val autoImport = Deps
    object Deps {
      val scalatest = "org.scalatest" %% "scalatest" % "2.1.5"
    }
   }
   ```
5. Observe that auto import works correctly.
# Problems

Documentation on `autoImport` claims that both vals and objects are supported but objects are not currently supported. Either docs should be updated or objects should be supported.
# Notes

Bug appears to be in `sbt.Plugins.hasAutoImportGetter` -- invoking this function from the REPL on each of the two variants above returns `false` and `true` respectively.

eed3si9n:
Fixes #1314 
## what this changes

Apparently when I was coding this I had intended to work only for `val` and `lazy val`. Later we included `object autoImport` as a use case but apparently did not check. Now we have the support for `object autoImport` and test to confirm it.

### DIFF
--- a/main/src/main/scala/sbt/Plugins.scala
+++ b/main/src/main/scala/sbt/Plugins.scala
@@ -319,9 +319,9 @@ ${listConflicts(conflicting)}""")
 		val hasGetterOpt = catching(classOf[ScalaReflectionException]) opt {
 			im.symbol.asType.toType.declaration(ru.newTermName("autoImport")) match {
 				case ru.NoSymbol => false
-				case sym => sym.asTerm.isGetter
+				case sym => sym.asTerm.isGetter || sym.asTerm.isModule
 			}
 		}
 		hasGetterOpt getOrElse false
-	} 
+	}
 }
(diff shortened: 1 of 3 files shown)


jsuereth:
LGTM.

PR #1319, opened 2014-05-09
Enable name hashing algorithm in sbt build.

gkossakowski:
The name hashing seems to be stable enough for sbt to use it by default
now. It also greatly improves incremental compilation experience for
people working on sbt sources.


jsuereth:
Thanks @gkossakowski !  This has been on my todo list for a long time.  LGTM.

gkossakowski:
Only the derived setting test failed (which is known to fail). Merging.

PR #1277, opened 2014-04-22
Separate location of analysis cache if we're cross compiling scala.

jsuereth:
- Add task to determine file-name of analysis cache
- If crossPaths := true, then add scala binary version to the analysis cache name.

This makes it possible to leverage incremental compilation with the `+` command.

Fixes #1267 

Review by @gkossakowski  cc> @eed3si9n

PR #1330, opened 2014-05-13
Use separate update caches when cross compiling scala

pvlugter:
Similar to https://github.com/sbt/sbt/pull/1277. Avoids unnecessary updates when switching scala versions.


jsuereth:
@pvlugter this LGTM. I can merge into 0.13 (0.13.6).  Do you need this fix earlier?  Unfortunately, 0.13.5 is in RC mode...

pvlugter:
0.13.6 is fine. Thanks.

Issue #1329, opened 2014-05-13
Order of magnitude performance regression in watch mode


jroper:
The performance of SBT's watch mode seems to have regressed across multiple releases, with a significant regression occurring since 0.13.1.

I've used the Play code base as a benchmark, on OSX, with 4 CPU cores.

I've used `find . -path '*/src/main/*' | wc -l` from Play's `framework` directory to estimate the number of source files being watched.

I've used `~compile` as the test to run, and then monitored the `java` process CPU usage in `top`.

On SBT 0.12.4 (Play 2.2.x branch, 482 source files), `java` uses a fairly steady 20% CPU.

On SBT 0.13.1 (Not on any branch, but at the ad306013fe6b2566f328bff3f41ff325d5a269a7 commit, 521 source files), `java` uses approx 40% CPU, fluctuating between 23% and 70%.

On SBT 0.13.5-RC1 (Play master, 574 source files), I'm seeing `java` use between 150% and 250% CPU, seemingly averaging around 200%.

Either adding more files for SBT to watch increases the amount of CPU consumed exponentially, or there's been some very serious regressions.

This has been reported by other users (at least one user on Linux):

https://github.com/playframework/playframework/issues/2793

This makes me very wary of ever giving control of watching files over to SBT.

Related: #1223

jroper:
I've captured some profiling snapshots between SBT 0.13.1 and 0.13.5-RC1, but the data from the profiling is a mess - I suspect that YourKit filters out by default many of the method calls that we're most interested in profiling.  One sampling snapshot that I got showed that SBT 0.13.5-RC1 was calling File.isHidden a lot, it wasn't being called in 0.13.1.  The tracing snapshots reveal nothing, the only methods they show taking any serious is the main call heirarchy.  If anyone knows how to better read these than me, then I'll be happy to send the data.

retronym:
I'd suggest to use sampling mode, rather than tracing mode, as the first pass.

When you view the CPU snapshot, you can configure which methods are filtered out of the view with:

![image](https://cloud.githubusercontent.com/assets/65551/2954117/0fd6fca0-da6c-11e3-8e3c-d1a2f6ecd3b9.png)

gaydenko:
I just want to add I haven't met the issue with 0.13.5-M2, a problem has appeared with RC1.

retronym:
I wonder if e076bfa970e is in play here. Do you see the same issue with `sbt -Dsbt.task.forcegc=false ...`

[...]

PR #1332, opened 2014-05-13
Default force gc to false

jsuereth:
Workaround for #1329 
Related #1223

Review by @eed3si9n


eed3si9n:
Could you add a documentation in `src/sphinx/Community/Changes.rst` for adding the force GC, and now defaulting to `false`?

jsuereth:
@eed3si9n done.  Also removed mention of RC1->RC2.   That kind of info isn't really needed in Changes, which is more about major release changes.

ijuma:
Good to see this. Forcing GCs so frequently is too dangerous.

jsuereth:
@ijuma sadly, we'll have to have some kind of an alternative to try to request one on misbehaving machines.  We're getting way to many preventable permgen errors on jenkins.

Issue #1240, opened 2014-04-07
javaOptions in test doesn't seem to work


ijuma:
The forking documentation page[1] includes the following line:

`javaOptions in test += "-Xmx8G"`

This did not have the desired effect in our build. Changing it slightly made it work:

`javaOptions in Test += "-Xmx8G"`

It looks like either code or documentation need to be changed. There is another `in test` instance on that same page. Maybe that needs updating too?

[1] http://www.scala-sbt.org/0.13.1/docs/Detailed-Topics/Forking.html

jsuereth:
This is a duplicate of #975

jsuereth:
Also, thanks for pointing out the docs, we should get those updated.

Issue #975, opened 2013-11-11
regression: fork in test doesn't work anymore


backuitist:
## steps

``` scala
fork in test := true
```
## problem

Test does not fork
## expectations

Test forks.
## notes

 `fork in Test` works. Also see [Enable forking](http://www.scala-sbt.org/0.13.2/docs/Detailed-Topics/Forking.html#enable-forking):

> To enable forking all test tasks only, set fork to true in the test scope:
> 
> ```
> fork in test := true
> ```

eed3si9n:
Here's what I have so far. This was introduced by sbt/sbt@0bb460c03ed9940c67c9048d353dfd151a5bfdd0 around line 377 of Defaults.scala:

```
 -      testExecution <<= testExecutionTask(key),
 -      testGrouping <<= testGrouping or singleTestGroup(key)
 -  ) )
 +      testExecution <<= testExecutionTask(key)
 +  ) ) ++ Seq(
 +      derive(testGrouping <<= singleTestGroupDefault)
 +  )
```

The commit message says it fixes #539. The actual fix I think is coming from the fact that it's now a derived setting, and the reported behavior is coming from the fact that `singleTestGroup(key)` was changed to `singleTestGroupDefault`, which is no longer scoped to the passed in `key` (e.g. `test` task).

``` scala
    def singleTestGroup(key: Scoped): Initialize[Task[Seq[Tests.Group]]] = inTask(key, singleTestGroupDefault)
    def singleTestGroupDefault: Initialize[Task[Seq[Tests.Group]]] = Def.task {
        val tests = definedTests.value
        val fk = fork.value
        val opts = forkOptions.value
        Seq(new Tests.Group("<default>", tests, if(fk) Tests.SubProcess(opts) else Tests.InProcess))
    }
```

Currently `Tests.Group` carries the knowledge of both the sequence of `TestDefinition` and `TestRunPolicy`, which controls forking.
So why was `singleTestGroup(key)` changed to `singleTestGroupDefault`? Here's what happens if you change it back:

```
[info] java.lang.RuntimeException: Scope cannot be defined for dependency baseDirectory of derived setting testGrouping defined at (sbt.Defaults) Defaults.scala:405
[info]  at scala.sys.package$.error(package.scala:27)
[info]  at scala.Predef$.error(Predef.scala:142)
[info]  at sbt.Init$$anonfun$derive$1.apply(Settings.scala:92)
[...]

Issue #708, opened 2013-03-13
Sbt ignores scalaVersion when source dependency added


espringe:
Checkout: https://github.com/espringe/scoobi

and sbt will happily uses the correct (2.10) version of scala. However, add a source dependency (e.g. uncomment this line):

https://github.com/espringe/scoobi/blob/9d0f0e5696/project/build.scala#L53

and sbt ignores the scalaVersion and uses scala 2.9.2 for everything

(And even the source project is a 2.10 project)

I'm using sbt 0.12.2

harrah:
I assume by "uses 2.9.2 for everything", you mainly mean dependencies and cross-versioning.  This is indeed a fundamental problem to be fixed for 0.13.  You can work around it by explicitly setting `scalaBinaryVersion := "2.10"` for the scoobi and ascii-graphs projects.  (If you didn't control `ascii-graphs`, you could do it from the scoobi project with the appropriate scope.)

Issue #539, opened 2012-09-01
When using multiple test configurations without testGrouping, the wrong tests are run


szeiger:
I am using this sbt build with the highlighted workaround for setting a dummy test grouping for both, the standard Test configuration and my own DocTest configuration: https://github.com/typesafehub/doc-example/blob/02836ed9901c7ebf5880ac7eae98b933f51238b9/project/Build.scala#L15-L18

Without these explicit testGroupings, the doctest:test command runs the same tests as the standard test command, instead of its own doctest tests.

vigdorchik:
I think when you just say `testGrouping <<= definedTests in Test map partitionTests`
you override testGrouping for all configurations to be a result of partitioning `definedTests in Tests`.  If you want the same function to apply to all configurations, just say `testGrouping <<= definedTests map partitionTests`. Otherwise you can detail the target on the left hand-side.

szeiger:
But even without any testGrouping for any of the configurations, the wrong tests are run.

vigdorchik:
Can you make a public repository that demonstrates the problem?

szeiger:
Ah, sorry, the original repo is private. You should be able to access it now. But the problem can also be reproduced with https://github.com/slick/slick. Just run doctest:test -> runs the correct tests for "doctest", then comment out both testGrouping lines in Build.scala, reload and run again -> now it runs the tests for "test" instead of "doctest".

vigdorchik:
Actually you only need the second setting `testGrouping in DocTest <<= definedTests in DocTest map partitionTests` but it's true, without it the code in sbt
https://github.com/harrah/xsbt/blob/0.13/main/Defaults.scala#L313
groups the wrong tests. I'm not sure about the fix though.

PR #1333, opened 2014-05-13
Fixed project/derived test to pass

jsuereth:
Review by @eed3si9n cc @dansanduleac

Basically, the type of derived setting test which is now failing is the following:

``` scala
Def.derive(customC := s"${organization.value}-${customC.value}-${version.value}")
```

Also, derived settings must _explicitly_ be placed in the global scope to be able to derive them into projects now.   All the default sbt derived settings had this in place, it's only the tests that needed to be updated.

In general, @dansanduleac's changes bring more "stability" and regularity to derived settings, the issue remains that I have no idea what the above setting _should_ do.  For now, we'll leave the behavior unspecified, as all tests are passing and we don't really document/recommend derived settings for users yet.

### DIFF
--- a/main/src/main/scala/sbt/Load.scala
+++ b/main/src/main/scala/sbt/Load.scala
@@ -536,7 +536,7 @@ object Load {
             case DiscoveredProjects(Some(root), discovered, files) =>
               log.debug(s"[Loading] Found root project ${root.id} w/ remaining ${discovered.map(_.id).mkString(",")}")
               val finalRoot = finalizeProject(root, files)
-              loadTransitive(discovered, buildBase, plugins, eval, injectSettings, acc :+ finalRoot, memoSettings, log, false, buildUri, context)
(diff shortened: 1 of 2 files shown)


eed3si9n:
> In general, @dansanduleac's changes bring more "stability" and regularity to derived settings, the issue remains that I have no idea what the above setting should do.

Since you brought up a general discussion.. I'm squarely on the fence about whether derived settings should exist in sbt given the number of times it's used in `Defaults.scala` and the issues it continues to cause like sbt/sbt#975 (aka #1240). I think we should start a thread on sbt-dev list, write up the problem sbt/sbt@0bb460c03ed9940c67c9048d353dfd151a5bfdd0 claims to solve (#539, #708, and #391), and re-evaluate if the solution is valid.

jsuereth:
@eed3si9n Actually the `in Test` vs. `in test` thing I think is another super confusing area of sbt ->  How delegation works in the presence of task-scope + configuratoin scope when both task and configuration share the same name...

Feel free to start up the argument.  I think we have a few bigger fish to fry right now before we'll have time to re-evaluate derived settings.

PR #1339, opened 2014-05-15
fix the "fork in [T]est" documentation

fwbrasil:
"fork in test" doesn't work with sbt 0.13.2

### DIFF
--- a/src/sphinx/Detailed-Topics/Forking.rst
+++ b/src/sphinx/Detailed-Topics/Forking.rst
@@ -48,7 +48,7 @@ the :key:`test` scope:
 
 ::
 
-    fork in test := true
+    fork in Test := true
 
 See :doc:`Testing` for more control over how tests are assigned to JVMs and
 what options to pass to each group.


jsuereth:
Thanks for the fix.  This should probably be issued against the 0.13.2b branch as 0.13 doens't auto-update the docs yet (0.13 branch has docs for upcoming 0.13.6)

eed3si9n:
Should we fix the doc or the bug?

jsuereth:
The doc.  I think `fork in test` is unnecessarily confusing.

jsuereth:
i.e it should either be `fork in Test` for ANY test task, or `fork in (Test, test)` for just the "test everything" task vs. `fork in (Test, testOnly)` for "test one thing".

As it stands as "<setting> in <task>" it really highlights some of the confusion we can create around setting delegation.

eed3si9n:
What about `fork in run`?

jsuereth:
Personally?  Yeah, I think it should be `fork in Runtime` (which doesn't work because, silly enough, the run task is implemented in the `Compile` scope, so with that caveat, `fork in (Compile, run)`).  I think task-scope only makes sense for a _particular_ task.   `fork in test` vs `fork in Test` -> Which one wins for `test:test`.  Is that something we want users to care about?

While we have the flexibility in the model to do these things, it doesn't mean we should.   I think limiting our own use of setting delegation to ease understanding is important.

eed3si9n:
It's conceivable to assert that `fork in run` automatically translates into `fork in (Compile, run)` since typing `run` from the shell delegates to `compile:run`, that type of semantic change needs to wait till 1.0. 
Both `fork in run` and `fork in test` has worked the same way since 0.12 (or even earlier).

jsuereth:
I'm not sure I agree with you here that it automatically translates. The reality of sbt is the opposite happens:

It looks in previous scopes for things.   The more I think about it, the `fork in test` on derived settings may actually just be a bug in how derived settings look for delegated setting values.  Perhaps we should investigate.

jsuereth:
Closing as website has moved to sbt/website project.

PR #1340, opened 2014-05-16
Abstract over dependency kind in Analysis

Duhemm:
Work in progress to make it easier to add new kinds of dependency.

Please see this discussion : https://groups.google.com/forum/#!topic/sbt-dev/clIbzCp9VX8


compile/inc/src/main/scala/sbt/inc/Analysis.scala
```diff
@@ -155,14 +159,35 @@ private class MAnalysis(val stamps: Stamps, val apis: APIs, val relations: Relat
...
+    val inherited = dep.context match {
+      case DependencyByInheritance => true
+      case _                       => false
+    }
+    copy(stamps, apis.markExternalAPI(toClassName, classApi), relations.addExternalDep(fromSrc, dep))
+  }
+
+  def addExternalDep(src: File, dep: String, depAPI: Source, inherited: Boolean): Analysis = {
+    val dependency = Dependency(ExternalDependencyEdge(src, dep, depAPI), if (inherited) DependencyByInheritance else DependencyByMemberRef)
+    copy(stamps, apis.markExternalAPI(dep, depAPI), relations.addExternalDep(src, dependency), infos)
```
gkossakowski: I think we can just forward to the other method to not duplicate the same code:

``` scala
addExternalDep(dependency)
```

compile/inc/src/main/scala/sbt/inc/Relations.scala
```diff
@@ -334,19 +346,55 @@ private class MRelationsDefaultImpl(srcProd: Relation[File, File], binaryDep: Re
     new MRelationsDefaultImpl(srcProd + (src, prod), binaryDep, direct = direct,
       publicInherited = publicInherited, classes + (src, name))
 
+  def addExternalDep(src: File, dependency: Dependency): Relations = {
```
gkossakowski: Did you consider dropping `src` parameter? That information is encoded in DependencyEdge already. Once we remove that parameter, we can also remove asserts that check for consistency between those two sources of the same information.

Duhemm: That's a good idea, thanks ! Btw, why is the counterpart of `addExternalDep` named `addInternalSrcDeps` (note the plural) ? Wouldn't it be better to have simply `addExternalDep` and `addInternal(Src)?Dep` ? I'm pretty sure you said ultimately we could have only one function to add dependencies, but cannot find back where this message is.

Duhemm:
How about binary dependencies ? How would you represent them ?

compile/inc/src/main/scala/sbt/inc/Relations.scala
gkossakowski: Yes, I think having one function for adding dependencies is the way to go. I think we should have just `addSrcDeps` that handles both external and internal dependencies. It would simply have signature `addSrcDeps(dependencies: Iterable[Dependency])`.

gkossakowski:
I think binary dependency is another kind of dependency edge. You would have

``` scala
/**
 * Represents a dependency edge between a source file and a binary file (a class file).
 */
final case class BinaryDependencyEdge(fromSrc: File, toBinary: File) extends DependencyEdge
```

You would have to verify what happens when we refer to a class file from a jar. Do we get a path to just a jar file or something else?

In a separate comment I'll outline what I think remains to be done.

gkossakowski:
Analysis have three methods that manipulate dependencies: `addSource`, `addBinaryDep`, `addExternalDep`. The `addSource` takes internal dependencies as parameter. I think we could just have `addSource` take dependencies of all kinds and deprecate the rest. I don't see any use case for adding dependencies after a source has been already added.

In other words, `addSource` would become central method that takes the following information about single src:
- src location (File)
- api representation
- stamp (which lets one to check if source has changed since the stamp has been)
- all dependencies of given source

Once refactoring is done, the only thing that is not tracked by `addSource` would be `products`. I don't see any reason why they shouldn't be tracked by `addSource` so we could add a TODO for that.

It would be `addSource` that is responsible for dispatching dependencies to right places (e.g. marking external apis as it's done in `addExternalDep` at the moment.

I think the same story would apply to `Relations` where we would unify all methods manipulating dependencies into just one method. As I mentioned above, I think binary dependencies should be refactored as well especially because macros can introduce binary dependencies too.

To sum up, I think we should have one method for manipulation dependencies both in `Analysis` and in `Relations`.

Let me know if that makes sense.

Duhemm:
This sounds good ! I'll find some time this afternoon to start implementing that and come back to you in case of problems.

xeno-by:
Hi guys! What's the status of this pull request? Would that be something that might be included in 0.13.5?

I also heard that @gkossakowski will be away for some time, so @jsuereth, @eed3si9n, maybe you would be able to review this PR?

gkossakowski:
The 0.13.5 is in RC mode already. The PR itself needs a bit more work before it can be merged.

Duhemm:
To summarize what has been done so far :
- Introduced a new representation of the dependencies in [Dependency.scala](https://github.com/Duhemm/sbt/blob/abstract-over-dependency-kind/compile/inc/src/main/scala/sbt/inc/Dependency.scala). Dependencies now have a "Dependency edge" that indicates the kind of dependency (internal, external, binary, product) and a "Dependency context" that indicates what introduced the dependency (member reference or inheritance, no context for binary dependencies and products).
- Previously there were many methods to add and manipulate dependencies (typically one per kind of dependency (`addProduct`, `addBinary`, and so on) in both [Analysis.scala](https://github.com/Duhemm/sbt/blob/abstract-over-dependency-kind/compile/inc/src/main/scala/sbt/inc/Analysis.scala) and [Relations.scala](https://github.com/Duhemm/sbt/blob/abstract-over-dependency-kind/compile/inc/src/main/scala/sbt/inc/Relations.scala). I refactored this so that there is only `addSource` that is not deprecated. Moreover, `addSource` expects to receive all the dependencies of a source file at once, whereas previously the dependencies were added one after the other.

(I don't think the failure of 3a0aab116ce37567a1d603f4f511de38d2896110 is due to my changes)

xeno-by:
@gkossakowski Could you elaborate on what needs to be done to get this merged? Our semester project report is due at Jun 6, and it would be nice to say there that Martin did this, this, and this, and also merged this, this, and this into a production version of SBT.

gkossakowski:
I restarted Travis build to clear up the failure of 3a0aab1.

Duhemm:
It looks like the only test that is currently failing is `sbt.SettingsTest`. I just ran a few times `test-only sbt.SettingsTest` and it always passes... Do you know how the changes I made could impact this part of SBT ?

It would be great if you could tell me what you feel comes next =)

jsuereth:
So those failures were from a flaky test.  Basically, we were accidentally testing derived settings with empty keys, and we break on empty keys (which is fine, we don't allow empty keys).   The test was updated in master, so to fix you can rebase.

So, this change is good to go from me, but I would like to see that test go green with a final rebase on master.

This change is definitely in time for 0.13.6, which should be RC/out july-ish.

gkossakowski:
I don't think merging products into dependencies is a good idea. After all, products (class files coming out of source files) are not dependencies of source files.

When I made a comment before saying `addSource` should handle `products` I meant it should have an additional parameter for taking list of products. Martin, could you update your PR accordingly?

There's also an interesting problem with the fact that all dependencies are marked as `private[inc]`. It's done intentionally so we refactor code in the future without worrying about the fact that they are part of private API. However, _public_ methods in `Analysis` take classes with restricted visibility. Those public methods (like addSource) can't be called from outside of `inc` package. I don't think this is a big deal because most people would just read what is in an Analysis and not try to manipulate it. However, we have to clearly document that new methods for manipulation of Analysis can't be called from outside of `inc` package. If somebody comes up with a use case of manipulating Analysis then we'll have to put a bit more work into this issue.

jsuereth:
The good news is, your tests are passing :)

gkossakowski:
The bad news is that Travis is not testing all of the commits so we don't know which commits have to be squashed :)

jsuereth:
@gkossakowski that's by design on Travis IIRC.

gkossakowski:
FYI: I've spent more time thinking about what should be done exactly. I'll write down my thoughts tomorrow.

gkossakowski:
Hi, unfortunately I didn't make it to get to this PR and write what I had in mind.

I'd suggest dropping abstracting over different dependency edges (their set should be fixed) and just introduce a way to abstract over dependency contexts (we expect those to expand in number).

Duhemm:
I've just opened a new PR in sbt/sbt#1387 for abstracting over dependency contexts instead of over dependency kinds. 

Please ping me whenever you feel like the original idea should be implemented !

Duhemm:
It looks like the build errored because it ran out of memory (https://travis-ci.org/sbt/sbt/jobs/29583345#L2139). It passed in the previous build.

typesafe-tools:

xeno-by:
ping @gkossakowski

Duhemm:
@gkossakowski Hi Greg ! I just rebased over 0.13. I'll try to find a nice way to squash / reorder commits while on the train.

gkossakowski:
Excellent! Please do! Let me know once you are done. I'll try to squeeze some time today to offer some comments.

Duhemm:
Hi Greg ! Sorry I didn't have time to do it this afternoon, I hope you've not changed your schedule to review this...

I ended up with 4 commits (that I'm pushing incrementally to this branch to make sure everything stays coherent) : https://github.com/Duhemm/sbt/compare/abstract-over-dependency-kind-clean

What do you think ? Actually I'm thinking that it would also make sense to keep only your first commit and then squash everything.

Duhemm:
All commits are here and the tests pass \o/

compile/inc/src/main/scala/sbt/inc/Analysis.scala
```diff
@@ -155,17 +165,49 @@ private class MAnalysis(val stamps: Stamps, val apis: APIs, val relations: Relat
...
+    val newAPIs = externalDeps.foldLeft(apis.markInternalSource(src, api)) {
+      case (tmpApis, ExternalDependency(_, toClassName, classApi, _)) => tmpApis.markExternalAPI(toClassName, classApi)
+    }
+
+    val newRelations = relations.addSource(src, products, internalDeps, externalDeps, binaryDeps)
+
+    copy(newStamps, newAPIs, newRelations, infos.add(src, info))
+  }
+
+  def addSource(src: File, api: Source, stamp: Stamp, directInternal: Iterable[File], inheritedInternal: Iterable[File], info: SourceInfo): Analysis = {
```
gkossakowski, line 193: How about rewriting this method so it forwards to the new implementation we have above?

compile/inc/src/main/scala/sbt/inc/Analysis.scala
```diff
@@ -51,9 +51,19 @@ trait Analysis {
   def copy(stamps: Stamps = stamps, apis: APIs = apis, relations: Relations = relations, infos: SourceInfos = infos,
     compilations: Compilations = compilations): Analysis
 
+  def addSource(src: File, api: Source, stamp: Stamp, info: SourceInfo,
+    products: Iterable[(File, String, Stamp)],
+    internalDeps: Iterable[InternalDependency],
+    externalDeps: Iterable[ExternalDependency],
+    binaryDeps: Iterable[(File, String, Stamp)]): Analysis
+
+  @deprecated("Register all products and dependencies in addSource.", "0.13.6")
```
gkossakowski: Sbt 0.13.6 is out already so this needs to be changed to 0.13.7.

compile/inc/src/main/scala/sbt/inc/Compile.scala
```diff
@@ -161,9 +161,13 @@ private final class AnalysisCallback(internalMap: File => Option[File], external
 
   def nameHashing: Boolean = options.nameHashing
 
-  def get: Analysis = addUsedNames(addCompilation(addExternals(addBinaries(addProducts(addSources(Analysis.empty(nameHashing = nameHashing)))))))
+  def get: Analysis = addUsedNames(addCompilation(addProductsAndDeps(Analysis.empty(nameHashing = nameHashing))))
+
+  @deprecated("Register products and dependencies using addProductsAndDeps", "0.13.6")
```
gkossakowski: Both `AnalysisCallback` the class and `AnalysisCallback` the interface can be treated as sbt-private so you don't need to worry about deprecations and just refactor the code directly.

compile/inc/src/main/scala/sbt/inc/Compile.scala
```diff
@@ -192,4 +200,30 @@ private final class AnalysisCallback(internalMap: File => Option[File], external
...
+      case (a, (src, api)) =>
+        val stamp = current.internalSource(src)
+        val hash = stamp match { case h: Hash => h.value; case _ => new Array[Byte](0) }
+        // TODO store this in Relations, rather than Source.
+        val hasMacro: Boolean = macroSources.contains(src)
+        val s = new xsbti.api.Source(compilation, hash, api._2, api._1, publicNameHashes(src), hasMacro)
+        val info = SourceInfos.makeInfo(getOrNil(reporteds, src), getOrNil(unreporteds, src))
+        val direct = sourceDeps.getOrElse(src, Nil: Iterable[File])
+        val publicInherited = inheritedSourceDeps.getOrElse(src, Nil: Iterable[File])
+        val (externalInherited, externalDirect) = extSrcDeps.filter(_._1 == src).partition(_._4)
```
gkossakowski: The `filter` here is linear scan. This will hurt performance in larger projects. You'll need to change `extSrcDeps` to be a Map. While you are at it, I'd change `extSrcDeps` to be `Map[File, Iterable[ExternalDependency]]`.

compile/inc/src/main/scala/sbt/inc/Relations.scala
```diff
@@ -78,8 +82,40 @@ trait Relations {
    * and inheritance dependencies on `inheritedDependsOn`.  Everything in `inheritedDependsOn` must be included in `directDependsOn`;
    * this method does not automatically record direct dependencies like `addExternalDep` does.
    */
+  @deprecated("Record all internal dependencies using `addInternalDeps`.", "0.13.6")
```
gkossakowski: Why different name instead of just introducing an overload?

compile/inc/src/main/scala/sbt/inc/Relations.scala
```diff
@@ -515,42 +649,59 @@ private class MRelationsNameHashing(srcProd: Relation[File, File], binaryDep: Re
...
+    new MRelationsNameHashing(srcProd + (src, prod), binaryDep, internalDependencies = internalDependencies,
+      externalDependencies = externalDependencies, classes + (src, name), names = names)
 
-  def addExternalDep(src: File, dependsOn: String, inherited: Boolean): Relations = {
-    val newIH = if (inherited) inheritance.addExternal(src, dependsOn) else inheritance
-    val newMR = memberRef.addExternal(src, dependsOn)
-    new MRelationsNameHashing(srcProd, binaryDep, memberRef = newMR, inheritance = newIH, classes,
-      names = names)
-  }
+  def addProducts(src: File, products: Iterable[(File, String, Stamp)]): Relations =
```
gkossakowski: The third element of the tuple is not used. Why take it as an argument to this method?

gkossakowski:
Let me summarize my feedback:
- [x] changes to `Analysis.scala` look good.
- [ ] in `Relations.scala` it would be good if could check if some names could be reduced to just overloads of existing methods
- [ ] you need to fix linear scan problem in `Compile.scala`

Other than that `Compile.scala` doesn't abstract over external dependencies properly (yet). However, this can be addressed in a separate PR. You could try to add better macro support to incremental compiler to see which places in the incremental compiler need more work in order to properly abstract over dependency contexts.

Duhemm:
Hi @gkossakowski ! Sorry for the timeout and thank you very much for your review. I've addressed the points you mentioned in your previous comments, and fixed the linear scan in `Compile.scala`. Could you please elaborate about the second checkbox ? I'm afraid some words are missing.

gkossakowski:
I edited the comment. Somehow, a few word got lost. I apologize for that.

compile/inc/src/main/scala/sbt/inc/Compile.scala
```diff
@@ -106,7 +105,10 @@ private final class AnalysisCallback(internalMap: File => Option[File], external
     binaryClassName.put(binary, className)
     add(binaryDeps, source, binary)
   }
-  def externalSourceDependency(t4: (File, String, Source, Boolean)) = extSrcDeps += t4
+
+  // The type corresponds to : (internal source, external source depended on, API of external dependency, true if an inheritance dependency)
+  def externalSourceDependency(t4: (File, String, Source, Boolean)) =
+    extSrcDeps += t4._1 -> List(ExternalDependency(t4._1, t4._2, t4._3, if (t4._4) DependencyByInheritance else DependencyByMemberRef))
```
gkossakowski: What happens when this method is called with the same file (first element of the tuple) multiple times?

Duhemm: The old values will be lost... Thanks.

Duhemm:
Regarding the new names in `Relations.scala`, most of the new (`private`) names are the plural form of `addProduct`, `addBinaryDep`, and `addExternalDep`. I think it makes more sense to keep these new names rather than overloading the old ones, since they get all the dependencies at once and not one by one.

The only new names that are public are `addSource`, `internalDependencies` and `externalDependencies`.

I had also deprecated `addInternalSrcDeps` in favor or `addInternalDeps`, but I've finally only introduced an overload of `addInternalSrcDeps` and removed `addInternalDeps`.

Duhemm:
Hi @gkossakowski ! Any update on the status of this PR ?

gkossakowski:
Hi! Thanks for pinging me! I'm at Typesafe eng meeting but I've allocated some time for inc compiler. This is on top of my list.

compile/inc/src/main/scala/sbt/inc/Analysis.scala
gkossakowski, line 193: @Duhemm, I still think my suggestion makes sense but we don't need to block the whole PR on this. You could do this in a separate PR once this one is merged.

compile/inc/src/main/scala/sbt/inc/Compile.scala
```diff
@@ -106,7 +105,12 @@ private final class AnalysisCallback(internalMap: File => Option[File], external
     binaryClassName.put(binary, className)
     add(binaryDeps, source, binary)
   }
-  def externalSourceDependency(t4: (File, String, Source, Boolean)) = extSrcDeps += t4
+
+  // The type corresponds to : (internal source, external source depended on, API of external dependency, true if an inheritance dependency)
+  def externalSourceDependency(t4: (File, String, Source, Boolean)) = {
```
gkossakowski, line 110: This method is called only from this file. How about marking it private and refactoring it to take regular method arguments? That would clean up the implementation too.

compile/inc/src/main/scala/sbt/inc/Compile.scala
```diff
@@ -161,30 +165,17 @@ private final class AnalysisCallback(internalMap: File => Option[File], external
...
   def getOrNil[A, B](m: collection.Map[A, Seq[B]], a: A): Seq[B] = m.get(a).toList.flatten
-  def addExternals(base: Analysis): Analysis = (base /: extSrcDeps) { case (a, (source, name, api, inherited)) => a.addExternalDep(source, name, api, inherited) }
   def addCompilation(base: Analysis): Analysis = base.copy(compilations = base.compilations.add(compilation))
   def addUsedNames(base: Analysis): Analysis = (base /: usedNames) {
     case (a, (src, names)) =>
       (a /: names) { case (a, name) => a.copy(relations = a.relations.addUsedName(src, name)) }
   }
 
+  // This is no longer used by the new implementation relative to Dependency contexts
+  // See https://github.com/sbt/sbt/issues/1340
```
gkossakowski, line 178: This method is internal to this class and it's not part of public API. You can just remove it.

compile/inc/src/main/scala/sbt/inc/Relations.scala
```diff
@@ -227,9 +269,11 @@ object Relations {
         assert(nameHashing || (memberRefSrcDeps == emptySourceDependencies), "When name hashing is disabled the `memberRef` relation should be empty.")
         assert(!nameHashing || (directSrcDeps == emptySource), "When name hashing is enabled the `direct` relation should be empty.")
 
-        if (nameHashing)
-          Relations.make(srcProd, binaryDep, memberRefSrcDeps, inheritanceSrcDeps, classes, names)
-        else {
+        if (nameHashing) {
+          val internal = InternalDependencies(Map(DependencyByMemberRef -> mri.asInstanceOf[Relation[File, File]], DependencyByInheritance -> ii.asInstanceOf[Relation[File, File]]))
```
gkossakowski, line 273: I think `mri.asInstanceOf[Relation[File, File]]` should be extracted to a separate variable. Same goes for all other casts. This would also make thos lines much shorter and easier to read.

compile/inc/src/main/scala/sbt/inc/Analysis.scala
```diff
@@ -51,9 +51,19 @@ trait Analysis {
   def copy(stamps: Stamps = stamps, apis: APIs = apis, relations: Relations = relations, infos: SourceInfos = infos,
     compilations: Compilations = compilations): Analysis
 
+  def addSource(src: File, api: Source, stamp: Stamp, info: SourceInfo,
+    products: Iterable[(File, String, Stamp)],
+    internalDeps: Iterable[InternalDependency],
+    externalDeps: Iterable[ExternalDependency],
+    binaryDeps: Iterable[(File, String, Stamp)]): Analysis
+
+  @deprecated("Register all products and dependencies in addSource.", "0.13.7")
```
gkossakowski: Sbt 0.13.7 has been released already. All deprecations need to be changed to have "0.13.8" as a version number.

gkossakowski:
Hi @Duhemm. I'm done reviewing this. I'd like all of my comments to be addressed but not all of them need to be done in this PR. It's been sitting open for too long already. The only thing that I'd like to be addressed before we merge is changing deprecations to point at "0.13.8" version number. Everything else can be addressed in a new PR.

Please ping me once you change those deprecations and we'll be ready to merge.

lightbend-cla-validator:
Hi @Duhemm,

Duhemm:
Hi @gkossakowski and thank you for the review ! I have just changed the deprecations. Should I squash this last commit ?

gkossakowski:
Yes, please squash it.

Duhemm:
It's done. If you could stop [this build](https://travis-ci.org/sbt/sbt/builds/41263657) (1112), it would be awesome. Thanks !

gkossakowski:
Hey! I just noticed that you squashed deprecation changes to the last commit. That's a mistake. This commit should touch just AnalysisTest. Could you move those changes to the commit that introduced deprecations? I think it's 078dced749d3e4cf6feccf97c003b8a0b95ac21e that should be your target.

You may need to use `git rebase -i` to achieve that. See http://stackoverflow.com/questions/6217156/break-a-previous-commit-into-multiple-commits and http://emmanuelbernard.com/blog/2014/04/14/split-a-commit-in-two-with-git/

Duhemm:
Oops, sorry about that ! I've fixed it.

gkossakowski:
LGTM!

Merging! Congratulations on a really good work.

gkossakowski:
@Duhemm, having said that, could you address my other comments in a separate PR?

compile/inc/src/main/scala/sbt/inc/Analysis.scala
Duhemm, line 193: @gkossakowski I'm not sure I understand. Isn't it what it does ? I think I changed the body of this method since September 15th but not its signature, thus your comment is not on an "outdated diff".

compile/inc/src/main/scala/sbt/inc/Compile.scala
Duhemm, line 110: Addressed in #1736.

compile/inc/src/main/scala/sbt/inc/Compile.scala
Duhemm, line 178: Addressed in #1736.

compile/inc/src/main/scala/sbt/inc/Analysis.scala
gkossakowski, line 193: Yes, you are right. I didn't notice you addressed my comment already. Thanks!

PR #1346, opened 2014-05-17
Ensure sequences in analysis files are read in order.

benjyw:
For various reasons, we serialize sequences as:

0 -> foo
1 -> bar
...

Until now we were implicitly relying on the sequences being in order.
However external code may end up (due to bugs or otherwise) messing
with the ordering:

1 -> bar
0 -> foo
...

This change ensures that we don't get confused by that. Although
it's best if external code doesn't mess up the ordering, it's still
a good idea to be defensive about this.

Note that the sequences we serialize are short, so the extra sort
is not a performance concern.

### DIFF
--- a/compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
+++ b/compile/persist/src/main/scala/sbt/inc/TextAnalysisFormat.scala
@@ -436,7 +436,7 @@ object TextAnalysisFormat {
   }
 
   private[this] def readSeq[T](in: BufferedReader)(expectedHeader: String, s2t: String => T): Seq[T] =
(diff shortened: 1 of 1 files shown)


jsuereth:
So, for the Travis failure, I rebooted the failing 2.11 job just to see if it was a transient "concurrency" test failure.  reviewing shortly.

gkossakowski:
@jsuereth: I'm happy to review anything incremental compiler related.

gkossakowski:
Actually, it's one-liner. Hence LGTM.

jsuereth:
Hmm, looks like we have some sort of unrelated 2.11 concurrency issue here making the tests fail.   I'm ok merging this, but we shouldn't let anything big through until we figure out that issue.

benjyw:
Thanks for the quick turnaround!

PR #1349, opened 2014-05-19
Remove line of dead code.

paulp:
Unmoored expression in constructor does nothing except pack
on a few micros.

### DIFF
--- a/main/command/src/main/scala/sbt/State.scala
+++ b/main/command/src/main/scala/sbt/State.scala
@@ -183,7 +183,6 @@ object State {
           log.debug(s"> $x")
           f(x, s.copy(remainingCommands = xs, history = x :: s.history))
       }
-    s.copy(remainingCommands = s.remainingCommands.drop(1))
     def :::(newCommands: Seq[String]): State = s.copy(remainingCommands = newCommands ++ s.remainingCommands)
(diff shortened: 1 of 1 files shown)


jsuereth:
LGTM.  Once travis reports success, will merge.

eed3si9n:
Here's from https://github.com/sbt/sbt/commit/d7b66458f20c7a71d8a063058ad3e3d18e0fdeb4#diff-26ff6a5cf7642f39621297cb1c094748R32 in 2010:

``` scala
 +  implicit def stateOps(s: State): StateOps = new StateOps {
 +      def process(f: (String, State) => State): State =
 +          s.commands match {
 +              case x :: xs => f(x, s.copy()(commands = xs))
 +              case Nil => exit(true)
 +          }
 +          s.copy()(commands = s.commands.drop(1))
```

the line in question was indented as if it's part of `def process`. And then scalariform unindented it (correctly) and it looks like this now:

``` scala
    def process(f: (String, State) => State): State =
      s.remainingCommands match {
        case Seq() => exit(true)
        case Seq(x, xs @ _*) =>
          log.debug(s"> $x")
          f(x, s.copy(remainingCommands = xs, history = x :: s.history))
      }
    s.copy(remainingCommands = s.remainingCommands.drop(1))
```

Issue #1356, opened 2014-05-20
[0.13.5] transitive dependency resolution error


eed3si9n:
Originally reported to:
-  sbt-dev as [transitive dependency resolution error in 0.13.5](https://groups.google.com/d/msg/sbt-dev/g_keXZOhmkw/FVh75SHVmkoJ)
- Stackoverflow as [sbt 0.13.2-M3 to 0.13.5-RC3 issue](http://stackoverflow.com/questions/23701209/sbt-0-13-2-m3-to-0-13-5-rc3-issue)
  by Gregory Jablonski 

> I'm having a Build that used to work using sbt version 0.13.2-M3 stop working with 0.13.5-RC2 (and -RC3).  Some transitive dependencies are failing to resolve, it's looking for weird "versions" that look like "working@<hostname>"
> 
> It only happens when the needed dependencies aren't already in the local ivy cache.
> 
> I put a question on stack overflow here http://stackoverflow.com/questions/23701209/sbt-0-13-2-m3-to-0-13-5-rc3-issue with a repro, but since this actually looks like a regression thought I'd mention it here too.
## Notes

Michael Pilquist noted:
- https://issues.apache.org/jira/browse/IVY-1457
- https://issues.apache.org/jira/browse/IVY-1467

0.13.2-M3 the version of sbt that's said to work uses Ivy 2.3.0:
- https://github.com/sbt/sbt/blob/v0.13.5-M2/project/Util.scala#L171

PR #1357, opened 2014-05-20
Roll back Ivy to 2.3.0 for 0.13.5 for #1356

eed3si9n:
This is a fix for #1356.

Due to test failures at the last minute
- https://travis-ci.org/sbt/ivy/builds/25567318
- https://travis-ci.org/sbt/sbt/jobs/25567288

we might need to roll back Ivy back to 2.3.0 final. 
The reported version of sbt that worked was 0.13.2-M3, which uses Ivy 2.3.0.


jsuereth:
Lgtm.  We can fix ivy perf updates for 0.13.6, but that does shift our current sprint

PR #1364, opened 2014-05-22
Accumulate Test Count

cheeseng:
Changed to combine SuiteResult with same name, instead of replacing with latest one. This solves the test result count problem when a Suite is executed more than 1 time due to being returned as nested suites (sub-task).

This solves incorrect test count problem when a suite in ScalaTest is executed more than one time as nested suites, current code simply replace the old count with the new count, this changes will add them up instead.

The following is an issue reported in ScalaTest that drive this request:

https://github.com/scalatest/scalatest/issues/167


eed3si9n:
LGTM. @jsuereth?

jsuereth:
LGTM - Test failure looks to be the 'empty key string' failure that Dan fixed in another PR.

Thanks again @cheeseng for the fixes!

cheeseng:
Thanks Eugene and Josh for pulling this in!

PR #1362, opened 2014-05-27
Allow bytecode enhancement to update analysis before it's saved

jroper:
See https://github.com/playframework/playframework/pull/2237 for a more in depth discussion on this.

Essentially, it should be possible for byte code enhancement tasks to do their work and update the analysis before it's saved.  Currently, the compile task saves the analysis to disk before it completes, so any byte code enhancement that happens after that, and hence needs to update the analysis (to prevent incremental compilation from invalidating the enhanced classes), needs to save the analysis again.

@gkossakowski suggested that we have the following tasks:

```
readAnalysis <- compileIncremental <- saveAnalysis <- compile
```

This would allow byte code enhancers to override the compileIncremental task.

A similar approach might be:

```
readAnalysis <- compileIncremental <- byteCodeEnhance <- saveAnalysis <- compile
```

Where `byteCodeEnhance` is a task that takes a sequence of byte code enhancers, and runs them in sequence.  This would allow multiple byte code enhancer plugins to automatically plug themselves in, the order being random based on plugin load order, but then also allow the user to override that, explicitly defining the order of that sequence if it matters.


jroper:
I've attached a PR to this.

Note, I've got no idea what the right breakdown of what should go where, where API's should or shouldn't live etc - this part of the code looks like it's had 20 iterations of standards half applied to it, so I've got no idea what the current standard is.  So, tell me how to rearrange the code according to the current standard, and I'll happily oblige.

gkossakowski:
Hi, unfortunately, I don't there's a clearly defined standard when it comes to incremental compiler's API. It's all mostly driven by what sbt, Scala IDE or zinc happen to use.

For example, Scala IDE [is using](https://github.com/scala-ide/scala-ide/blob/master/org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/EclipseSbtBuildManager.scala#L237) the `AggressiveCompile` class so we can't just change it's signature. We'll have to prepare all changes in source backward-compatible way.

I'll provide some comments in line.

compile/integration/src/main/scala/sbt/compiler/AggressiveCompile.scala
```diff
@@ -33,17 +33,20 @@ class AggressiveCompile(cacheFile: File) {
...
     javacOptions: Seq[String] = Nil,
+    previousAnalysis: Analysis,
+    previousSetup: Option[CompileSetup],
     analysisMap: File => Option[Analysis] = { _ => None },
     definesClass: DefinesClass = Locate.definesClass _,
     reporter: Reporter,
     compileOrder: CompileOrder = Mixed,
     skip: Boolean = false,
-    incrementalCompilerOptions: IncOptions)(implicit log: Logger): Analysis =
+    incrementalCompilerOptions: IncOptions)(implicit log: Logger): (Analysis, CompileSetup, Boolean) =
```
gkossakowski, line 43: Why not return `AnalysisResult` here?

main/actions/src/main/scala/sbt/Compiler.scala
```diff
@@ -14,16 +14,21 @@ import java.io.File
 object Compiler {
   val DefaultMaxErrors = 100
 
-  final case class Inputs(compilers: Compilers, config: Options, incSetup: IncSetup)
+  final case class Inputs(compilers: Compilers, config: Options, incSetup: IncSetup, previousAnalysis: PreviousAnalysis)
   final case class Options(classpath: Seq[File], sources: Seq[File], classesDirectory: File, options: Seq[String], javacOptions: Seq[String], maxErrors: Int, sourcePositionMapper: Position => Position, order: CompileOrder)
   final case class IncSetup(analysisMap: File => Option[Analysis], definesClass: DefinesClass, skip: Boolean, cacheFile: File, cache: GlobalsCache, incOptions: IncOptions)
   final case class Compilers(scalac: AnalyzingCompiler, javac: JavaTool)
+  final case class PreviousAnalysis(analysis: Analysis, setup: Option[CompileSetup])
```
gkossakowski, line 21: Why is setup optional here? It would be great to add a few lines comments to this class even if there's no documentation in the neighborhood. You've got to start somewhere.

main/actions/src/main/scala/sbt/Compiler.scala
```diff
@@ -57,21 +62,22 @@ object Compiler {
...
+  def apply(in: Inputs, log: Logger, reporter: xsbti.Reporter): AnalysisResult =
     {
       import in.compilers._
       import in.config._
       import in.incSetup._
-      val agg = new AggressiveCompile(cacheFile)
-      agg(scalac, javac, sources, classpath, CompileOutput(classesDirectory), cache, None, options, javacOptions,
-        analysisMap, definesClass, reporter, order, skip, incOptions)(log)
+      val agg = new AggressiveCompile
+      AnalysisResult.tupled(agg(scalac, javac, sources, classpath, CompileOutput(classesDirectory), cache, None, options, javacOptions,
```
gkossakowski, line 78: I think I'd be more happy to pack less in one line than more.

compile/integration/src/main/scala/sbt/compiler/AggressiveCompile.scala
```diff
@@ -24,7 +24,7 @@ final class CompileConfiguration(val sources: Seq[File], val classpath: Seq[File
   val previousAnalysis: Analysis, val previousSetup: Option[CompileSetup], val currentSetup: CompileSetup, val progress: Option[CompileProgress], val getAnalysis: File => Option[Analysis], val definesClass: DefinesClass,
   val reporter: Reporter, val compiler: AnalyzingCompiler, val javac: xsbti.compile.JavaCompiler, val cache: GlobalsCache, val incOptions: IncOptions)
 
-class AggressiveCompile(cacheFile: File) {
+class AggressiveCompile {
```
gkossakowski, line 27: This is part of API that Scala-IDE is using, we'll have to cook changes in backwards compatible way. See:
https://github.com/scala-ide/scala-ide/blob/master/org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/EclipseSbtBuildManager.scala#L237

compile/integration/src/main/scala/sbt/compiler/AggressiveCompile.scala
jroper, line 43: All the case classes used in the SBT "interface" (ie, types used in SBT keys) were in the Compilers class, so it felt natural that AnalysisResult also go there.  AggressiveCompile is in a different sub project that can't see that, so it therefore was excluded from using that.  Happy to move AnalysisResult, but would like some guidance as to exactly where it should go.

main/actions/src/main/scala/sbt/Compiler.scala
jroper, line 21: I didn't dig deep into `AggressiveCompile`, but its methods want an `Analysis` and `Option[CompileSetup]`.  We could make it take an `Option[(Analysis, CompileSetup)]`, which would mean for this we could make `setup` not an `Option`, but in order to create the empty analysis, I think it needs to know the hashing strategy, and I'm not sure where that gets passed to - it may be no problem, but I'm just not sure.

compile/integration/src/main/scala/sbt/compiler/AggressiveCompile.scala
jroper, line 27: One thing we could do is leave `AggressiveCompile` alone, and provide a new API.  In fact, we could put everything that SBT needs into one method on the companion object, and have the instance just do cache reading/writing, and delegate to the companion object.

main/actions/src/main/scala/sbt/Compiler.scala
gkossakowski, line 21: I'll look into that tomorrow to understand the issue exactly.

compile/integration/src/main/scala/sbt/compiler/AggressiveCompile.scala
gkossakowski, line 27: That sounds good to me.

Let's make the new API `private[sbt]` if we are not 100% that it's final iteration.

compile/integration/src/main/scala/sbt/compiler/AggressiveCompile.scala
gkossakowski, line 43: The original commit (502f355a65456ec50d7bea29ac5f1f92a9a547bb) that introduced those case classes said it's just a convenience. I don't really understand why convenience methods related to incremental compiler do not live in a subproject related to incremental compiler.

For that reason, I think we should not add more code there and just grow what's in incremental compiler so other projects like zinc and Scala IDE can benefit from the same API.

jroper, line 43: So should I move it to AggressiveCompile.scala as a top level class, or as class in the AggressiveCompile object, or something else?

gkossakowski, line 43: Actually, after sleeping over it I think it doesn't really matter where AnalysisResult lives. The whole code needs more work so figuring this out for just one class doesn't really buy us anything.

I'm fine keeping this as-is.

main/actions/src/main/scala/sbt/Compiler.scala
gkossakowski, line 21: I poked around the code and I didn't really see why we would need an Option at all. I think you could just pass `(Analysis, CompileSetup)`. When it comes to name hashing flag, it's stored in IncOptions that you might need to thread through.

gkossakowski:
I'll be going offline for next at least two weeks. @jroper, you'll need find some other reviewer for this PR but I think the core of the change is fine. You just need to make sure that you are not breaking scalaide and zinc.

jsuereth:
I can take over from here then, @gkossakowski.  THanks for doing what you could and good luck with the surgery.

@jroper Give me a bit to dig in and ping me if you've addressed all greg's concerns or need help doing so.

main/actions/src/main/scala/sbt/Compiler.scala
```diff
@@ -14,16 +14,21 @@ import java.io.File
 object Compiler {
   val DefaultMaxErrors = 100
 
-  final case class Inputs(compilers: Compilers, config: Options, incSetup: IncSetup)
+  final case class Inputs(compilers: Compilers, config: Options, incSetup: IncSetup, previousAnalysis: PreviousAnalysis)
```
jsuereth, line 17: Unfortunately, even those this is final, this breaks binary compatibility of the unapply method of Inputs.

jsuereth, line 17: You can cheat with:

``` scala
final case class Inputs(compilers: Compilers, config: Options, incSetup: IncSetup)(val previousAnalysis: PreviousAnalysis)
```

I believe, although this doesn't fix the copy-method, it may be ok.

main/src/main/scala/sbt/Keys.scala
```diff
@@ -140,6 +140,9 @@ object Keys {
   val consoleQuick = TaskKey[Unit]("console-quick", "Starts the Scala interpreter with the project dependencies on the classpath.", ATask, console)
   val consoleProject = TaskKey[Unit]("console-project", "Starts the Scala interpreter with the sbt and the build definition on the classpath and useful imports.", AMinusTask)
   val compile = TaskKey[Analysis]("compile", "Compiles sources.", APlusTask)
+  val compileIncremental = TaskKey[Compiler.AnalysisResult]("compileIncremental", "Actually runs the incremental compliation", DTask)
```
jsuereth, line 143: I like this split of tasks.

jsuereth:
SO generally, LGTM but a few important BC issues to look into, particularly modifying the case classes.

typesafe-tools:

jsuereth:
@jroper i'm revisiting this patch this week. Is the primary goal the creation of 'save analysis' and 'load analysis' tasks so that Play can hook 'save analysis' or 'load analysis' to take into account its own auto-generated code?

jsuereth:
@jroper @gkossakowski I've moved this into a branch  (wip/bytecode-enhancement).  I'll be cleaning it up on there.

PR #1383, opened 2014-05-30
use custom scriptedParser instead of spaceDelimited parser

xuwei-k:
https://github.com/sbt/sbt/blob/v0.13.5/project/Sbt.scala#L179-L192


jsuereth:
OOOH, nice.   However, did you see the improved scripted parser I added in 0.13 branch (not 0.13.5)?

I've meant to get around to do this so KUDOS.  If you could take a look at my additons (mostly about fragmenting test suites for travis CI) let me know if you think they belong in the plugin.

Otherwise LGTM.

PR #1385, opened 2014-06-01
Comment out all site generation on 0.13.5 branch

eed3si9n:
Since the sphinx part of the site is migrated to sbt/website, and API and SXR is done for 0.13.5, Docs.settings can be `Nil` for 0.13.5 branch to prevent Jenkins from overwriting the new site. Once we merge this to 0.13 we can uncomment the API and SXR part.

### DIFF
--- a/project/Docs.scala
+++ b/project/Docs.scala
@@ -12,7 +12,6 @@
 
 object Docs
 {
-	val rootFiles = SettingKey[Seq[File]]("root-files", "Location of file that will be copied to the website root.")
 	val latestRelease = SettingKey[Boolean]("latest-release")
 
 	val siteExcludes = Set(".buildinfo", "objects.inv")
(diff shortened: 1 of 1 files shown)


project/Docs.scala
```diff
@@ -27,27 +26,18 @@ object Docs
 	val HomeHtml = "home.html"
 	val VersionPattern = """(\d+)\.(\d+)\.(\d+)(-.+)?""".r.pattern
 
-	def settings: Seq[Setting[_]] =
-		site.settings ++
-		site.sphinxSupport(DocsPath) ++
-		site.includeScaladoc("api") ++
```
jsuereth, line 33: I'm of the opinion that we should still let the API docs be created on this branch, so users can improve them during the 0.13.5 release.   0.13.6 API docs will be different.

Also, I think most of what we want for pushing API/sxr docs from here is in this code, so I think maybe we shouldn't gut all of this.

jsuereth:
Take a look at my altered branch and let me know what you think.

project/Docs.scala
eed3si9n, line 33: I'm still thinking about this, but more leaning towards treating Scaladoc comment as part of code. We should just cut a release and let it go. Any improvement should just wait till 0.13.6. There are several benefits for this:
- Documentation maintenance of 0.13.5 doesn't cut into the development time of 0.13.6.
- Having the clear deadline should force us to review API docs as part of milestone process.
- When we ship zipped version of the API, what we shipped on the release date is the same thing we show on the website.

I only deleted the Sphnix related code. site.settings is just commented out.

jsuereth, line 33: See my pr.  We can disable autopush, but when we do release doc generation
would still work.  I.e. second part of your todo.

Regarding whether we always push javadocs I see your point and xanelive
with that.

Maybe we need a "upcoming" javadoc list too.

eed3si9n:
I'm going to close this so we can continue the discussion in #1389.

PR #1387, opened 2014-06-01
Abstract over dependency contexts

Duhemm:
The current implementation of Analysis and Relations expect to receive
information about the context in which a dependency has been collected
as a boolean value simply indicating whether the dependency comes from
an inheritance relationship or from a member reference.

Unfortunately, adding a new kind of dependency context (for instance,
dependencies coming from a macro expansion) requires to change the
signature of many methods in these classes, and can lead to
inconsistencies.

Since the number of dependency contexts is expected to grow,
abstracting over the dependency contexts permits to stabilize the
signature of these methods.

This pull request is a follow up from sbt/sbt#1340, which came from this discussion : https://groups.google.com/forum/#!topic/sbt-dev/clIbzCp9VX8

This implement the idea from @gkossakowski that we should only abstract over dependency contexts instead of over the whole dependency kind for now.

/cc @xeno-by


Duhemm:
Close this & reopen #1340 as suggested by @gkossakowski

Issue #1390, opened 2014-06-02
ClassLoaderCache failure


eed3si9n:
```
[info] ! ClassLoaderCache concurrent access.Same class loader for same classpaths concurrently processed: Falsified after 18 passed tests.
[info] > ARG_0: List("ykmoyw", "PjebuueXivwvuvcO")
[info] > ARG_1: 932
```
- https://travis-ci.org/sbt/sbt/jobs/26525290#L2004

PR #1391, opened 2014-06-02
Improve test failure message for #1390.

jsuereth:
Remove flaky test from sbt.

What was happening is memory pressure was causing the classloader to get evicted in the SCala 2.11 code.   I expect this is because parallel collections in Scala 2.11 use weakreferences as well during execution, so for the .par method, there was no strong reference to any classloader.

Still this is super odd, but if the classloader _IS_ collected, it's not the end of the world, nor is it a bug, so the test can disappear.


jsuereth:
Review by @eed3si9n

PR #1393, opened 2014-06-04
One more fix to release script.

jsuereth:
- Ensure completion is appropriately cross versioned
- The publish launcher task should not try to publish
  the useless root project.

Review by @eed3si9n cc> @havocp


eed3si9n:
Travis failed with the following:

``` scala
Reference to undefined setting: 
  root/*:deployLauncher from root/*:publishLauncher (/home/travis/build/sbt/sbt/project/Sbt.scala:307)
```

jsuereth:
@eed3si9n Just caught that.  I'm undoing the freaking `CredentialsFile.exists` gunk in our release script.   I'll create a command to catch missing credentials and we'll update the publish command to handle that.

jsuereth:
@eed3si9n this passes except for the flaky 2.11 test.

PR #1396, opened 2014-06-05
Adds support for Scala 2.12

skyluc:
Extends the tweaks to the configuration for 2.11 to 2.12

It likely requires harrah/sbinary#14

### DIFF
--- a/project/Util.scala
+++ b/project/Util.scala
@@ -26,7 +26,7 @@ object Util {
   def noRemotePublish(in: Seq[Setting[_]]) = in filterNot { _.key.key == publish.key }
 
   def nightlySettings = Seq(
-    nightly211 <<= scalaVersion(_.startsWith("2.11.")),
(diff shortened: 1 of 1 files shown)


jsuereth:
Thanks for the patch! LGTM after travis reports success.

skyluc:
I have a hard time to see what are the failures. It looks like some network problems plus some tests failing.

jsuereth:
Yeah, maven central failures with junk like this:

```
[info] [error] SERVER ERROR: Connection timed out url=http://repo1.maven.org/maven2/org/scala-lang/scala-compiler/2.10.0-M7/scala-compiler-2.10.0-M7.pom
```

I've restarted.  You also ran into the memory pressure bug in the ClassLoaderCache test.  I have another PR where i'm removing that test (for lacking utility and being flaky).  However I want to see the rest of the 2.11 tests pass if we can.

PR #1400, opened 2014-06-05
Add 'scala-jar' to the list of jar artifacts.

dpratt:
The CustomPomParser has a hard-coded list of 'blessed' Maven POM packaging types
that are actually packaged as jar files, but Ivy incorrectly maps the file extension.
This patch allows artifacts published with the 'scala-jar' packaging to be properly
resolved and used in an SBT project.


jsuereth:
This LGTM, but it'd be nice to have a test as well.

dpratt:
Sure - I can cook up a test.

FYI - this is to support some stuff we use but is also public on github and maven central. Basically, it's a custom packaging scheme in maven that automatically pulls in some sensible defaults for a scala project (integration with surefire, auto importing scala-maven-plugin, etc).

Unfortunately, the best way to do this is with a maven package type, which causes Ivy to attempt to think the extension of the published artifacts is 'scala-jar'.
—

jsuereth:
yep!  I totally understand the maven package type, had debated doing that a few years back but avoided due to the issue you're trying to fix now (and didn't now how to fix back then).

The improvement is fine, would just love a test :)

dpratt:
Ask and ye shall receive.

jsuereth:
does the jar file have to be legitimate or can we just "touch" it to avoid having something larger stick around the repo?

dpratt:
Nope - you could just touch it. The PR has a jar with a basic "Hello World" class in it, but there's no reason it couldn't be empty.

jsuereth:
Yeah, I mean if we look for the class in the file, that'd be one thing.  However, we've been trying to find ways to speed up our tests.  Dealing with a 0 KB file may be better, but it's not a big deal.

I rebooted your TravisCI tests. Looks like you ran into a DerivedSetting test failure.  We just did a round of removing flaky tests, we may have missed a bug here, so don't worry.  Looks like your test was successful :)

dpratt:
Looks like the travis build failed again - it doesn't look like this relates to my PR, but if it does, please let me know and I'll fix it ASAP.

jsuereth:
I just rebooted the travis tests.  We can merge once htey're good.  Sorry for taking so long getting to this :)

PR #1401, opened 2014-06-08
Triggered Execution should be aware of rename or move file

xuwei-k:
The details of the pull req is described in [this Japanese blog post](http://d.hatena.ne.jp/xuwei/20140608/1402167335) @xuwei-k wrote yesterday. 
## steps

Set up a project structure as follows:

``` scala
val a = project

val b = project.dependsOn(a)

val root = project.aggregate(a, b)
```

From sbt shell run `~compile` at the root level, which aggregates both `a` and `b`.

While the triggered execution is still in play, move some source files from `a` project to `b` project.
## problem

Currently recompilation is not triggered.
## expectation

Recompilation is triggered.
## notes

@eed3s9n provided feedback during development to compare the performance on:
- comparing two `Set[Strings]`s
- comparing two sorted `List[String]`s
- comparing sha1 of sorted `List[String]`s that's `mkString`ed

There was no clear advantage among the three. Using Play2, Scalaz, or sbt's code base the most it took during comparison seemed to be around 20~30ms.

### DIFF
--- a/util/io/src/main/scala/sbt/SourceModificationWatch.scala
+++ b/util/io/src/main/scala/sbt/SourceModificationWatch.scala
@@ -10,13 +10,14 @@ object SourceModificationWatch {
     {
       import state._
 
-      def sourceFiles: Iterable[java.io.File] = sourcesFinder.get
-      val (lastModifiedTime, fileCount) =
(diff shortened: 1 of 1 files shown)


eed3si9n:
LGTM. @jsuereth Is there anything we can do in term of testing triggered execution?

jsuereth:
LGTM

PR #1405, opened 2014-06-12
BintrayMavenRepository added

evgeny-goldin:
This PR needs some additional work. Right now trying to use sbt 0.13.6-SNAPSHOT I'm getting 

```
java.lang.NoSuchFieldError: JCenter
    at sbt.Classpaths$.sbt$Classpaths$$bootRepository(Defaults.scala:1555)
    at sbt.Classpaths$$anonfun$appRepositories$1.apply(Defaults.scala:1527)
    at sbt.Classpaths$$anonfun$appRepositories$1.apply(Defaults.scala:1527)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
    at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
    at scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:34)
    at scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
    at scala.collection.AbstractTraversable.map(Traversable.scala:105)
    at sbt.Classpaths$.appRepositories(Defaults.scala:1527)
    at sbt.Classpaths$$anonfun$40.apply(Defaults.scala:1041)
    at sbt.Classpaths$$anonfun$40.apply(Defaults.scala:1041)
    at sbt.EvaluateSettings$MixedNode.evaluate0(INode.scala:175)
    at sbt.EvaluateSettings$INode.evaluate(INode.scala:135)
    at sbt.EvaluateSettings$$anonfun$sbt$EvaluateSettings$$submitEvaluate$1.apply$mcV$sp(INode.scala:69)
    at sbt.EvaluateSettings.sbt$EvaluateSettings$$run0(INode.scala:78)
    at sbt.EvaluateSettings$$anon$3.run(INode.scala:74)
[error] java.lang.NoSuchFieldError: JCenter
```

### DIFF
--- a/ivy/src/main/scala/sbt/Resolver.scala
+++ b/ivy/src/main/scala/sbt/Resolver.scala
@@ -20,6 +20,10 @@ sealed case class MavenRepository(name: String, root: String) extends Resolver {
   override def toString = name + ": " + root
 }
 
+sealed class JCenter(isSecure: Boolean = false) extends MavenRepository("jcenter", s"http${if (isSecure) "s" else ""}://jcenter.bintray.com/")
+
+sealed class BintrayMavenRepository(subject: String, repo: String, isSecure: Boolean = false) extends MavenRepository(s"bintray$subject/$repo/", s"http${if (isSecure) "s" else ""}://dl.bintray.com/$subject/$repo/")
+
 final class Patterns(val ivyPatterns: Seq[String], val artifactPatterns: Seq[String], val isMavenCompatible: Boolean, val descriptorOptional: Boolean, val skipConsistencyCheck: Boolean) {
   private[sbt] def mavenStyle(): Patterns = Patterns(ivyPatterns, artifactPatterns, true)
   private[sbt] def withDescriptorOptional(): Patterns = Patterns(ivyPatterns, artifactPatterns, isMavenCompatible, true, skipConsistencyCheck)
@@ -135,6 +139,7 @@ final case class SftpRepository(name: String, connection: SshConnection, pattern
 
 import Resolver._
 
+object JCenter extends JCenter(false)
 object DefaultMavenRepository extends MavenRepository("public", IBiblioResolver.DEFAULT_M2_ROOT)
 object JavaNet2Repository extends MavenRepository(JavaNet2RepositoryName, JavaNet2RepositoryRoot)
 object JavaNet1Repository extends JavaNet1Repository


.gitignore
```diff
@@ -1,2 +1,3 @@
+.idea
```
jsuereth: We prefer to have these in your git global ignore config file so idea-related thigns remain a user-machine concern.

ivy/src/main/scala/sbt/Resolver.scala
```diff
@@ -20,6 +20,22 @@ sealed case class MavenRepository(name: String, root: String) extends Resolver {
   override def toString = name + ": " + root
 }
 
+sealed class BintrayMavenRepository(subject: String = "", repo: String = "", isSecure: Boolean = false) extends MavenRepository("", "") {
```
jsuereth: So, having "magic strings" is generally not recommended.   For JCenter, I'd recommend somethign like:

``` scala
class JCenter(isSecure: Boolean = false) extends MavenRepository { ..}
class BintrayMavenRepository(subject: String, repo: String) extends MavenRepository { ... }
```

Would be a bit more ideal.

ivy/src/main/scala/sbt/Resolver.scala
```diff
@@ -136,6 +152,7 @@ final case class SftpRepository(name: String, connection: SshConnection, pattern
 import Resolver._
 
 object DefaultMavenRepository extends MavenRepository("public", IBiblioResolver.DEFAULT_M2_ROOT)
+object BintrayMavenRepository extends MavenRepository("bintray", "http://jcenter.bintray.com/")
```
jsuereth: I feel this should be called JCenter.

indrajitr:
:heart:

jsuereth:
So BTW generally, love this patch.  The Launcher still needs some work to support the bintray string `launch/src/main/scala`.

My comment about JCenter (and the magic strings) still holds as well.

evgeny-goldin:
Thanks, Josh. I'm now working on addressing your comments.

evgeny-goldin:
Hi,

I updated the PR. It addresses the comments above and also updates JCenter to be a default Maven repo. Some new issues:
- Attempting to use secure JCenter fails with what seems to be a problem on sbt side and JCenter SSL certificate:
  
  ```
  externalResolvers   := Resolver.withDefaultResolvers( Seq(new JCenter(true)), jcenter = false )
  ```
  
  causes
  
  ```
  [info] downloading https://jcenter.bintray.com/org/springframework/spring-core/4.0.5.RELEASE/spring-core-4.0.5.RELEASE.jar ...
  [warn]  [FAILED     ] org.springframework#spring-core;4.0.5.RELEASE!spring-core.jar: The HTTP response code for https://jcenter.bintray.com/org/springframework/spring-core/4.0.5.RELEASE/spring-core-4.0.5.RELEASE.jar did not indicate a success. See log for more detail. (51ms)
  [warn]  [FAILED     ] org.springframework#spring-core;4.0.5.RELEASE!spring-core.jar: The HTTP response code for https://jcenter.bintray.com/org/springframework/spring-core/4.0.5.RELEASE/spring-core-4.0.5.RELEASE.jar did not indicate a success. See log for more detail. (51ms)
  [warn] ==== jcenter: tried
  [warn]   https://jcenter.bintray.com/org/springframework/spring-core/4.0.5.RELEASE/spring-core-4.0.5.RELEASE.jar
  ```
  
  even though
  
  ```
  $ curl -I https://jcenter.bintray.com/org/springframework/spring-core/4.0.5.RELEASE/spring-core-4.0.5.RELEASE.jar
  HTTP/1.1 200 OK
  Server: nginx
  Date: Sun, 15 Jun 2014 02:49:35 GMT
  Content-Type: application/octet-stream
  Content-Length: 976350
  ...
  ```
- I had to comment out `case Predefined.JCenter` in `main/src/main/scala/sbt/Defaults.scala` as it causes `NoSuchFieldError` to be thrown, mentioned in my first comment to this PR.
- The default Maven repo is made to be JCenter. So this makes the following list of built-in reolvers:
  - `DefaultMavenRepository` = http://jcenter.bintray.com/, enabled by default
  - `CentralMavenRepository` = http://repo1.maven.org/maven2/
  - `JCenter` = http://jcenter.bintray.com/
  - `BintrayMavenRepository` = http://dl.bintray.com/[subject]/[repo]/
- After enabling JCenter as the default Maven repo - chaining to Typesafe repos when sbt is launched has stopped working. I remove `~/.ivy2/cache/` and start sbt, then I see all artifacts are resolved from JCenter and resolving `org.scala-tools.sbinary:sbinary_2.10:0.4.2` fails as it is only available at [repo.typesafe.com](http://repo.typesafe.com/typesafe/ivy-releases/org.scala-tools.sbinary/sbinary_2.10/0.4.2/)
- `ivy/src/main/scala/sbt/Resolver.scala`: in `withDefaultResolvers` I changed the name of the `boolean` argument from `mavenCentral` to `jcenter`. This actually introduces backward-incompatible change as someone might have referred to this argument by name in `build.sbt`, as in my first bullet above.

jsuereth:
Note:  This test looks like it needs to be fixed: https://travis-ci.org/sbt/sbt/jobs/27596517

I'll look at the details of the rest as soon as I can.

launch/src/main/scala/xsbt/boot/Update.scala
```diff
@@ -315,6 +315,7 @@ final class Update(config: UpdateConfiguration) {
           case Local                                      => localResolver(settings.getDefaultIvyUserDir.getAbsolutePath)
           case MavenLocal                                 => mavenLocal
           case MavenCentral                               => mavenMainResolver
+          case JCenter                                    => mavenResolver("JCenter", "http://jcenter.bintray.com")
```
jsuereth: This "JCenter" should be lower case ("jcenter") to match the other JCenter object.

launch/src/main/scala/xsbt/boot/Update.scala
```diff
@@ -348,14 +349,15 @@ final class Update(config: UpdateConfiguration) {
       resolver.setRoot(root)
       resolver
     }
-  /** Creates a resolver for Maven Central.*/
-  private def mavenMainResolver = defaultMavenResolver("Maven Central")
+  /** Creates a resolver for JCenter.*/
+  private def mavenMainResolver = defaultMavenResolver("JCenter")
```
jsuereth: This should not be overriden, as we still want the old `maven-central` string to work against the old maven-central.

launch/src/main/scala/xsbt/boot/Update.scala
```diff
@@ -348,14 +349,15 @@ final class Update(config: UpdateConfiguration) {
...
   /** Creates a maven-style resolver with the default root.*/
   private def defaultMavenResolver(name: String) =
     {
       val resolver = new IBiblioResolver
       resolver.setName(name)
       resolver.setM2compatible(true)
+      resolver.setRoot("http://jcenter.bintray.com/")
```
jsuereth: Similarly, this shouldn't be here.

launch/src/main/scala/xsbt/boot/Update.scala
evgeny-goldin: Fixed

launch/src/main/scala/xsbt/boot/Update.scala
evgeny-goldin: Removed

launch/src/main/scala/xsbt/boot/Update.scala
evgeny-goldin: Removed

evgeny-goldin:
Will look into failing test

evgeny-goldin:
Hello Josh, I'll need your input regarding the failing tests. Hope you don't mind.
- `sbt "scripted dependency-management/pom-advanced"` - this test fails due to new JCenter resolver found in the generated POM. My understanding is previously when Central was the default repo - it wasn't written out to the POM as Maven uses it anyway. Now with JCenter being the default repo - it's written out to the POM and it fails an assumption in a test. This can be fixed be changing "`sbt/src/sbt-test/dependency-management/pom-advanced/test`" as shown below, but is it the correct fix?
  
  ```
  # write the default pom.  The only repositories should be Scala Tools Releases and Snapshots
  > check-pom http://jcenter.bintray.com/ https://oss.sonatype.org/content/repositories/releases/ https://oss.sonatype.org/content/repositories/snapshots/
  
  # include file:// repositories.  The generated repositories section should include the local Maven repository as well
  $ touch repo.all
  > check-pom http://jcenter.bintray.com/ https://oss.sonatype.org/content/repositories/releases/ https://oss.sonatype.org/content/repositories/snapshots/ file://*.m2/repository/
  
  # include file:// repositories.  The generated repositories section should include the local Maven repository as well
  $ delete repo.all
  $ touch repo.none
  > check-pom
  ```
- Both `sbt "safeUnitTests"` and `sbt "checkBuildScala211"` fail with `NoSuchFieldError: : JCenter  (Update.scala:318)` which I couldn't understand - `Predefined` enum does include this (`launch/interface/src/main/java/xsbti/Predefined.java`), is there anything else missing?

Once we resolved the failing tests - we have an issue of sbt not delegating to Typesafe repos (one of the bullets in my message above) when the default repo has switched to JCenter.

Thanks!

jsuereth:
Ok, so:
- `sbt "scripted dependency-management/pom-advanced"`  - You need to update the test for your change.   Personally though, I think making bintray the DEFAULT repository should be either a separate commit or PR, where you would have to update that test data to reflect the new default.
- `sbt/src/sbt-test/dependency-management/pom-advanced/test` - Yes the fix is correct, but again making bintray the default should be a separate PR.
- have you run those tests on your own machine (after a clean)?  I think it could be an issue with using a previous launcher.jar file to run those tests rather than the most recently compiled version.  I may have to look into that if you don't see any obvious code issues.

ivy/src/main/scala/sbt/Resolver.scala
```diff
@@ -135,7 +139,9 @@ final case class SftpRepository(name: String, connection: SshConnection, pattern
 
 import Resolver._
 
-object DefaultMavenRepository extends MavenRepository("public", IBiblioResolver.DEFAULT_M2_ROOT)
+object JCenter extends JCenter(false)
+object DefaultMavenRepository extends JCenter(false)
```
jsuereth: This should not be changed yet.   People may still want to point to the previous one.

ivy/src/main/scala/sbt/Resolver.scala
```diff
@@ -168,12 +174,12 @@ object Resolver {
     withDefaultResolvers(userResolvers, true)
   /**
    * Add the local Ivy repository to the user repositories.
-   * If `mavenCentral` is true, add the Maven Central repository.
+   * If `jcenter` is true, add the JCenter repository.
    */
-  def withDefaultResolvers(userResolvers: Seq[Resolver], mavenCentral: Boolean): Seq[Resolver] =
+  def withDefaultResolvers(userResolvers: Seq[Resolver], jcenter: Boolean): Seq[Resolver] =
```
jsuereth: You actually can't change the names of arguments in a safe way without an sbt version bump.  Therefore this still needs to be called mavenCentral, thanks to named-arguments in scala.  Also, migrating to jcenter as default should be a separate PR.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1552,6 +1552,7 @@ object Classpaths {
         case p: xsbti.PredefinedRepository => p.id match {
           case Predefined.Local                => Resolver.defaultLocal
           case Predefined.MavenLocal           => Resolver.mavenLocal
+          //          case Predefined.JCenter              => new JCenter
```
jsuereth: Why is this commented out?  Also, why not use the object `JCenter`?

jsuereth:
I think this is coming  a LOT closer to mergeable!   However, I'd like to split it in two:
1. A PR which adds bintray DSL support throughout sbt
2. A PR which migrates from Maven Central -> Bintray as the default repository.

The second PR we'll need to ping sbt-dev for a community discussion.  That shouldn't block 90% of the code from making it though.

evgeny-goldin:
Thanks a lot, Josh. I'll address it all next week.

evgeny-goldin:
I've cleaned up this PR from all changes related to making JCenter the default repo, it now only includes changes for new `JCenter` and `BintrayMavenRepository` resolvers. 

I see one of the tests [fails](https://travis-ci.org/sbt/sbt/builds/30369497). Running `sbt safeUnitTests` locally produces:

```
[info] listing all files in an absolute base directory should
[info] ! produce the entire base directory's contents
[error]    NoSuchElementException: : None.get  (file:1)
```

Travis  [gives](https://travis-ci.org/sbt/sbt/jobs/30369513):

```
[info] > Exception: java.util.NoSuchElementException: null
[info] + settings.Derived setting(s) replace DerivedSetting in the Seq[Setting[_]]: OK, proved property.
[info] + settings.DerivedSetting in ThisBuild scopes derived settings under projects thus allowing safe +=: OK, passed 100 tests.
[error] Error: Total 11, Failed 0, Errors 1, Passed 10
[error] Error during tests:
[error]     sbt.SettingsTest
```

Could it be caused by merging with recent branch changes?

typesafe-tools:

evgeny-goldin:
Hi @jsuereth, could you have a look on this PR? This time the scope is only reduced to adding two custom resolvers. And we have one test failing which I couldn't quite see why. Thank you!

jsuereth:
Hmm, these are two different tests.  The failing local test is a file-system ordering issue with OSX vs. linux.  As i develop on linux I've asked some OSX developers to look into the issue and fix the test, but Travis works as does my machine.

As for the derived settings test, it may just legitimately be flaky, or it's catching a bug in derived settings which is mostly uncaught.   I'll have to look in detail at the error, but don't worry about those.

Travis CI has been failing lately due to other reasons.  I'll look at the patch now.

launch/interface/src/main/java/xsbti/Predefined.java
```diff
@@ -4,6 +4,7 @@
 {
 	Local("local"),
 	MavenLocal("maven-local"),
+	JCenter("jcenter"),
```
jsuereth: So, looks like the launcher support itself is missing here.  I.e you have the "internal to sbt" ivy support, but the sbt-launcher is actually a different project that shares classes with sbt itself + has its own ivy code.  I believe you had hooks for that at one point, but if we're only adding the Resolvers to sbt, then this launcher piece should be dropped.

src/sphinx/Detailed-Topics/Resolvers.rst
```diff
@@ -22,6 +22,10 @@ A few predefined repositories are available and are listed below
 
 -  `DefaultMavenRepository` This is the main Maven repository at
    http://repo1.maven.org/maven2/ and is included by default
+-  `JCenter` This is the Bintray JCenter repository at
+   http://jcenter.bintray.com/
+-  `BintrayMavenRepository` This is the Bintray repository at
```
jsuereth: Docs look good!

jsuereth:
One comment about the launcher again, otherwise LGTM.  Also, ping me if you want to chat (live) about this PR.   The launcher architecture/design may not be apparent from the code layout.  I've debated many times if it belongs as a separate project or not.  In any case, you can either remove anything in xsbti, or we can discuss what can go in the launcher as a V1 patch.

evgeny-goldin:
Right, removed the change from `Predefined.java`, it was a leftover from the bigger PR.

evgeny-goldin:
After merging this PR with `"0.13"` branch I see tests fail with unresolved Scala dependency.

jsuereth:
@evgeny-goldin Yeah, that's related to a maven-central issue we're working on.

ok to test  mr. jenkins.

jsuereth:
Would you mind doing a `git rebase -i <origin?sbt?insert your sbt/sbt remote here>/0.13` and squashing all unnecessary commits as well as updating against current master?  There was a borked patch on 0.13, as well as this PR no longer merging cleanly.

Thanks for the hard work!  Almost there :)

evgeny-goldin:
Done! But I see *.rst files were removed 2 commits ago?

jsuereth:
@evgeny-goldin Yeah, we moved the website to http://github.com/sbt/website  to make it easier to contribute to.  The only "site" this project generates now are API + SXR pages.

evgeny-goldin:
Awesome, thanks! Opened https://github.com/sbt/website/pull/42 for resolvers doc update.

eed3si9n:
When @jsuereth gave me a chance to review this pull req before he merged, I only glanced over it and gave it an ok, but I now that I'm looking at the documentation, I feel like the place `class JCenter` and `class BintrayMavenRepository` is added is inconsistent with all other hard-coded repository builders we provide in the file like:

``` scala
object Resolver {
  ....
   def typesafeRepo(status: String) = new MavenRepository("typesafe-" + status, TypesafeRepositoryRoot + "/" + status)
   def typesafeIvyRepo(status: String) = url("typesafe-ivy-" + status, new URL(TypesafeRepositoryRoot + "/ivy-" + status + "/"))(ivyStylePatterns)
   def sbtPluginRepo(status: String) = url("sbt-plugin-" + status, new URL(SbtPluginRepositoryRoot + "/sbt-plugin-" + status + "/"))(ivyStylePatterns)
   def sonatypeRepo(status: String) = new MavenRepository("sonatype-" + status, SonatypeRepositoryRoot + "/" + status)
  ....
}
```

Shouldn't they also be just a method under `Resolver`?

PR #1409, opened 2014-06-14
make it possible to reference a branch of a local git repository

vn971:
Reasoning on why this change is wanted can be found
on the "sbt-dev" mailing list,
topic [problem with referencing a local git repository](https://groups.google.com/forum/#!topic/sbt-dev/uWBMBOII3GM).

```
file:///home/vasya/dev/repo.git#branch
```

### DIFF
--- a/util/io/src/main/scala/sbt/IO.scala
+++ b/util/io/src/main/scala/sbt/IO.scala
@@ -832,7 +832,7 @@ object IO {
     {
       if (!uri.isAbsolute) return uri; //assertAbsolute(uri)
       val str = uri.toASCIIString
-      val dirStr = if (str.endsWith("/") || uri.getScheme != FileScheme || uri.getRawFragment != null) str else str + "/"
+      val dirStr = if (str.endsWith("/") || uri.getScheme != FileScheme || Option(uri.getRawFragment).isDefined) str else str + "/"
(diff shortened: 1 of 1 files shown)


jsuereth:
LGTM

util/io/src/main/scala/sbt/IO.scala
```diff
@@ -832,7 +832,7 @@ object IO {
     {
       if (!uri.isAbsolute) return uri; //assertAbsolute(uri)
       val str = uri.toASCIIString
-      val dirStr = if (str.endsWith("/") || uri.getScheme != FileScheme) str else str + "/"
+      val dirStr = if (str.endsWith("/") || uri.getScheme != FileScheme || uri.getRawFragment != null) str else str + "/"
```
eed3si9n: This is a personal preference, but I prefer `Option(uri.getRawFragment).isDefined` over `null` in the code.

vn971: Should I change?

eed3si9n: Yea, could you? Unless you have some good reason to keep `null` in there. Otherwise, I'm ready to merge.

eed3si9n:
Merged. Thanks for the contribution.

PR #1419, opened 2014-06-23
Fixed scripted newer command

jroper:
The scripted newer command was effectively a noop, it always passed because it returned false instead of throwing an exception when it failed.

Implemented specs for most of the scripted file commands as well.


jsuereth:
LGTM.  Thanks for also adding a decent set of tests here.

PR #1438, opened 2014-07-08
Update broken links on the Community Plugins documentation page

ollyw:
A couple of basic fixes to the documentation:
- The sbt-ghpages repo has moved and the old one is no longer accessible, so the link has been updated.
- The Schoir plugin is not longer accessible in the typesafe github area, and it seems to have disappeared without trace, so I have removed it.


jsuereth:
Note: the documentation has moved to http://github.com/sbt/website

PR #1450, opened 2014-07-14
Select the first test fingerprint for a test name for forked tests

pvlugter:
This takes the 'distinct by name' filtering which is applied to in-process tests and makes it available to forked tests (by moving it to Tests.processOptions which is used in both in-process and forked).

The test from the original commit for the filtering (https://github.com/sbt/sbt/commit/a0e301e) is extended to also test when forked, and this test fails without the other changes.

This resolves junit tests being run twice as reported in https://github.com/sbt/junit-interface/issues/14, for at least some of the test samples.

### DIFF
--- a/main/actions/src/main/scala/sbt/Tests.scala
+++ b/main/actions/src/main/scala/sbt/Tests.scala
@@ -150,7 +150,14 @@ object Tests {
       def includeTest(test: TestDefinition) = !excludeTestsSet.contains(test.name) && testFilters.forall(filter => filter(test.name))
       val filtered0 = discovered.filter(includeTest).toList.distinct
       val tests = if (orderedFilters.isEmpty) filtered0 else orderedFilters.flatMap(f => filtered0.filter(d => f(d.name))).toList.distinct
-      new ProcessedOptions(tests, setup.toList, cleanup.toList, testListeners.toList)
+      val uniqueTests = distinctBy(tests)(_.name)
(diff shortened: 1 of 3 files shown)


main/actions/src/main/scala/sbt/Tests.scala
```diff
@@ -150,7 +150,14 @@ object Tests {
       def includeTest(test: TestDefinition) = !excludeTestsSet.contains(test.name) && testFilters.forall(filter => filter(test.name))
       val filtered0 = discovered.filter(includeTest).toList.distinct
       val tests = if (orderedFilters.isEmpty) filtered0 else orderedFilters.flatMap(f => filtered0.filter(d => f(d.name))).toList.distinct
-      new ProcessedOptions(tests, setup.toList, cleanup.toList, testListeners.toList)
+      val uniqueTests = distinctBy(tests)(_.name)
+      new ProcessedOptions(uniqueTests, setup.toList, cleanup.toList, testListeners.toList)
+    }
+
+  private[this] def distinctBy[T, K](in: Seq[T])(f: T => K): Seq[T] =
```
jsuereth, line 157: This seems like a SUPER useful method for the collections API.

jsuereth:
LGTM

benmccann:
Woohoo! Thanks! I'm so glad you tracked that down. I've seen complaints about tests running twice on and off for the past couple of years

benmccann:
I keep finding more and more bugs that I thought were unrelated to this, but are actually solved by this fix. Thank you sooo much!

Issue #413, opened 2012-03-28
Artifact resolution is unreasonably slow for multi-module projects


jwebb:
### note from sbt team

This issue is now solved using [cached resolution](http://www.scala-sbt.org/0.13/docs/Cached-Resolution.html) feature. 

``` scala
updateOptions := updateOptions.value.withConsolidatedResolution(true)
```

If anyone has further issues with dependency resolution performance, please provide more context and open a new Github issue.
### original report

So I have a multi-module project - dozens of modules - in which many of the modules have very similar (transitive) dependencies. Each day, sbt (v0.11.2) decides it needs to re-check all those dependencies, and it does so independently for each module. So if I have, say, 20 modules depending on slf4j, sbt appears to hit the repository 20 times to re-download exactly the same pom.xml.  Multiply that by a hundred-odd dependencies and the whole process is taking a couple of minutes. Seems like some caching or up-front reconciliation would be helpful here.

nau:
Confirm, we have the same problem. It's the most annoying thing using sbt. We spend unreasonable amount of time just waiting for dependencies resolution. Any suggestion on how to improve this would be really appreciated.

manuelbernhardt:
Same issue here. It takes 5-10 minutes to resolve artifacts, and it is really quite frustrating sometimes.

manuelbernhardt:
I made a screen capture to illustrate how slow it is: http://vimeo.com/61423682

It often is much slower, when on a slower network connection. This is SBT 0.12.2

everson:
I have the same issue. It is consistently slow, no matter how much times I run it. It takes more time resolving than building itself, even on a fast network connection. Using SBT 0.12.4.

nau:
Guys, any progress on this? We spend horrifying amount of time just resolving same hundred dependencies for each of 20 sub-projects. It can take 10 minutes to resolve the same artifacts that didn't change. I tired of hearing "Maven does it better" and hearing suggestions to move back to Maven.
Currently, for us, this is the most annoying, hence the most important issue. Please, help :)

mtgto:
I affected same problem in my project using sbt multi-modules (more than 10).
To solve this problem, I traced how HTTP requests called by sbt.
I found below remarkable situations:
- When I use **_SNAPSHOT VERSION library**_, sbt might access to the resolver _per every building of sub-projects_.
- sbt might not memory any HTTP response (include `404 Not Found`) on library resolution of another sub-project.

[...]

PR #1454, opened 2014-07-17
Consolidated resolution

eed3si9n:
Fixes/improves #413

This adds a new setting key called `updateOptions`, which can enable consolidated resolution for update task.
The consolidated resolution automatically generates an artificial module descriptor based on the SHA-1 of all external dependencies (all direct, and transitive via internal dependencies). This consolidates the Ivy resolution of similar Ivy dependency graph across multiple subprojects.

This is how it's enabled:

``` scala
updateOptions := updateOptions.value.withConsolidatedResolution(true)
```

/review @jsuereth

### DIFF
--- a/ivy/src/main/scala/sbt/Ivy.scala
+++ b/ivy/src/main/scala/sbt/Ivy.scala
@@ -4,6 +4,7 @@
 package sbt
 
 import Resolver.PluginPattern
+import ivyint.{ ConsolidatedResolveEngine, ConsolidatedResolveCache }
 
 import java.io.File
 import java.net.URI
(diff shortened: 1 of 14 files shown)


jsuereth:
Failures look related: https://travis-ci.org/sbt/sbt/jobs/30140249

jsuereth:
LGTM, once tests pass.

eed3si9n:
After some travis job restarts, all tests are passing.

jsuereth:
If cache is in memory why is it using a directory?

Use a temporary directory if it's truly transient

eed3si9n:
The cache dir is used to save a marker file to indicate cache is available. Running `clean` on any of the participating subproject of the artificial graph or the on the root project cleans the marker file, and thus forces uncached `update` next time.

jsuereth:
I may have missed the clean hook.

It's ok to merge as-is, but I feel like the cache file is "off" and we'll hit bugs round it.   However, since it's opt in, that's ok for v1 so we can discover what the issues may be.

I would like to see a test with a projectref on a separare build and make sure the cache isn't borked

eed3si9n:
I might be able to get rid of the folder and make this 100% in-memory, if `clean` just deleted items from the mutable map. I might have to switch to using `TrieMap` then for concurrency guard?
Also I want to still work on SHA-1 calculation of the Ivy graph.

jsuereth:
LGTM

PR #1456, opened 2014-07-17
Export build.sbt values inside sbt.

jsuereth:
- Expose the values PAST the Eval/sbt.compiler package.
- Find projects using the name API rather than finding htem and dropping all values immediately.
- Adds a test to make sure the .sbt values are discovered and set-able
- Expose .sbt values in Set command and inside BuildUnit methods.
- Ensure `consoleProject` can see build.sbt values.
- Add notes for where we can look in the build if we want to expose .sbt values between files.

Review by @eed3si9n


main/actions/src/main/scala/sbt/compiler/Eval.scala
```diff
@@ -88,7 +88,7 @@ final class Eval(optionsNoncp: Seq[String], classpath: Seq[File], mkReporter: Se
       val value = (cl: ClassLoader) => getValue[Any](i.enclosingModule, i.loader(cl))
       new EvalResult(i.extra, value, i.generated, i.enclosingModule)
     }
-  def evalDefinitions(definitions: Seq[(String, scala.Range)], imports: EvalImports, srcName: String, valTypes: Seq[String]): EvalDefinitions =
+  def evalDefinitions(definitions: Seq[(String, scala.Range)], imports: EvalImports, srcName: String): EvalDefinitions =
```
eed3si9n: bincompat?

main/src/main/scala/sbt/BuildStructure.scala
```diff
@@ -50,15 +50,15 @@ final class LoadedBuildUnit(val unit: BuildUnit, val defined: Map[String, Resolv
...
-  def classpath: Seq[File] = unit.definitions.target ++ unit.plugins.classpath
+  def classpath: Seq[File] = unit.definitions.target ++ unit.plugins.classpath ++ unit.definitions.dslDefinitions.classpath
 
   /**
    * The class loader to use for this build unit's publicly visible code.
-   * It includes build definition and plugin classes, but not classes for .sbt file statements and expressions.
+   * It includes build definition and plugin classes and classes for .sbt file statements and expressions.
    */
-  def loader = unit.definitions.loader
+  def loader = unit.definitions.dslDefinitions.classloader(unit.definitions.loader)
```
eed3si9n, line 61: How frequently would this be called?

main/actions/src/main/scala/sbt/compiler/Eval.scala
jsuereth: Not an issue. None of this is exposed in a way that bincompat could be an issue (but I did think through it carefully)

main/src/main/scala/sbt/BuildStructure.scala
jsuereth, line 61: I dunno.  I'd make it a lazy-val, but that's risky.  I think we need to figure out caching behind this (which may already exist).

AFAICT it's only called from `set` and/or `consoleProject`.

eed3si9n, line 61: If it's only called from `set` then it's not a big deal. I am just hoping it's not creating a new instance of classloader per task execution, like during `~compile`.

main/src/main/scala/sbt/Main.scala
```diff
@@ -234,7 +234,19 @@ object BuiltinCommands {
     case (s, (all, arg)) =>
       val extracted = Project extract s
       import extracted._
-      val settings = EvaluateConfigurations.evaluateSetting(session.currentEval(), "<set>", imports(extracted), arg, LineRange(0, 0))(currentLoader)
+      val dslVals = extracted.currentUnit.unit.definitions.dslDefinitions
+      // TODO - This is horribly inefficient.  We should try to only attach the
+      // classloader + imports NEEDED to compile the set command.
+      System.err.println(s"DSL imports: ${dslVals.imports mkString "\n"}")
```
eed3si9n: println?

jsuereth: DOH, good catch

jsuereth:
@eed3si9n Fixes made and tests now pass.

PR #1467, opened 2014-07-27
Eviction warnings (Fixes #1200)

### LINKED ISSUE #1200: shell should display all evicted libraries on start up
This is inspired by [In SBT 0.13, does scalaVersion still control the version of scala used for compile, run and test?](http://stackoverflow.com/questions/22551430/), [sbt-lock](https://github.com/tkawachi/sbt-lock), and nescala talks around efforts in the field to control the deps pulled in by Ivy.
## problems

Currently the build user is at the mercy of conflict management by Ivy. So when it evicts Scala library or any other libraries, our usual response is.

THIS IS IVY

(THIS IS IVY)

## what we could do

We have the knowledge of when eviction happens, so why not add a task to display the list of evicted libraries including scala-library.jar on the shell, and call it at starts ups and/or the on completion of `update`.

eed3si9n:
Fixes #1200.

This implements all stories from https://github.com/sbt/sbt/wiki/User-Stories%3A-Conflict-Warning.
sbt will display eviction warnings when it resolves your project's managed dependencies via `update` task. Currently the eviction warnings are categorized into three layers: `scalaVersion` eviction, direct evictions, and transitive evictions. By default eviction warning on `update` task will display only `scalaVersion` eviction and direct evictions.

`scalaVersion` eviction warns you when `scalaVersion` is no longer effective. This happens when one of your dependency depends on a newer release of scala-library than your `scalaVersion`.

```
[warn] Scala version was updated by one of library dependencies:
[warn]     * org.scala-lang:scala-library:2.10.2 -> 2.10.3
[warn] Run 'evicted' to see detailed eviction warnings
```

Direct evictions are evictions related to your direct dependencies. Warnings are displayed only when API incompatibility is suspected. For Java libraries, Semantic Versioning is used for guessing, and for Scala libraries Second Segment versioning (second segment bump makes API incompatible) is used.

```
[warn] There may be incompatibilities among your library dependencies.
[warn] Here are some of the libraries that were evicted:
[warn]   * commons-io:commons-io:1.4 -> 2.4
[warn] Run 'evicted' to see detailed eviction warnings
```

To display all eviction warnings with caller information, run `evicted` task.

```
[warn] There may be incompatibilities among your library dependencies.
[warn] Here are some of the libraries that were evicted:
[warn]     * com.typesafe.akka:akka-actor_2.10:2.1.4 -> 2.3.4 (caller: com.typesafe.akka:akka-remote_2.10:2.3.4,
org.w3:banana-sesame_2.10:0.4, org.w3:banana-rdf_2.10:0.4)
```

/review @jsuereth

### DIFF
--- a/ivy/src/main/scala/sbt/EvictionWarning.scala
+++ b/ivy/src/main/scala/sbt/EvictionWarning.scala
@@ -1,0 +1,44 @@
+package sbt
+
+import collection.mutable
+
+final class EvictionWarningOptions(
+    val configurations: Seq[String],
+    val level: Level.Value) {
+}
+object EvictionWarningOptions {
+  def apply(): EvictionWarningOptions =
+    new EvictionWarningOptions(Vector("compile"), Level.Warn)
+}
+
+object EvictionWarning {
+  def apply(options: EvictionWarningOptions, report: UpdateReport, log: Logger): Unit = {
+    val evictions = buildEvictions(options, report)
+    processEvictions(evictions, log)
+  }
+
+  private[sbt] def buildEvictions(options: EvictionWarningOptions, report: UpdateReport): Seq[ModuleDetailReport] = {
+    val buffer: mutable.ListBuffer[ModuleDetailReport] = mutable.ListBuffer()
+    val confs = report.configurations filter { x => options.configurations contains x.configuration }
+    confs flatMap { confReport =>
+      confReport.details map { detail =>
+        if ((detail.modules exists { _.evicted }) &&
+          !(buffer exists { x => (x.organization == detail.organization) && (x.name == detail.name) })) {
+          buffer += detail
+        }
+      }
+    }
+    buffer.toList.toVector
+  }
+
+  private[sbt] def processEvictions(evictions: Seq[ModuleDetailReport], log: Logger): Unit = {
+    if (!evictions.isEmpty) {
+      log.warn("Some dependencies were evicted:")
+      evictions foreach { detail =>
+        val revs = detail.modules filter { _.evicted } map { _.module.revision }
+        val winner = (detail.modules filterNot { _.evicted } map { _.module.revision }).headOption map { " -> " + _ } getOrElse ""
+        log.warn(s"\t* ${detail.organization}:${detail.name} (${revs.mkString(", ")})$winner")
+      }
+    }
+  }
+}
--- a/ivy/src/main/scala/sbt/EvictionWarning.scala
+++ b/ivy/src/main/scala/sbt/EvictionWarning.scala
@@ -1,25 +1,120 @@
 package sbt
 
 import collection.mutable
+import Configurations.Compile
 
-final class EvictionWarningOptions(
-    val configurations: Seq[String],
-    val level: Level.Value) {
+final class EvictionWarningOptions private[sbt] (
+    val configurations: Seq[Configuration],
+    val warnScalaVersionEviction: Boolean,
+    val warnDirectEvictions: Boolean,
+    val warnTransitiveEvictions: Boolean,
+    val showCallers: Boolean,
+    val guessCompatible: Function1[(ModuleID, Option[ModuleID], Option[IvyScala]), Boolean]) {
+  private[sbt] def configStrings = configurations map { _.name }
+
+  def withConfigurations(configurations: Seq[Configuration]): EvictionWarningOptions =
+    copy(configurations = configurations)
+  def withWarnScalaVersionEviction(warnScalaVersionEviction: Boolean): EvictionWarningOptions =
+    copy(warnScalaVersionEviction = warnScalaVersionEviction)
+  def withWarnDirectEvictions(warnDirectEvictions: Boolean): EvictionWarningOptions =
+    copy(warnDirectEvictions = warnDirectEvictions)
+  def withWarnTransitiveEvictions(warnTransitiveEvictions: Boolean): EvictionWarningOptions =
+    copy(warnTransitiveEvictions = warnTransitiveEvictions)
+  def withShowCallers(showCallers: Boolean): EvictionWarningOptions =
+    copy(showCallers = showCallers)
+  def withGuessCompatible(guessCompatible: Function1[(ModuleID, Option[ModuleID], Option[IvyScala]), Boolean]): EvictionWarningOptions =
+    copy(guessCompatible = guessCompatible)
+
+  private[sbt] def copy(configurations: Seq[Configuration] = configurations,
+    warnScalaVersionEviction: Boolean = warnScalaVersionEviction,
+    warnDirectEvictions: Boolean = warnDirectEvictions,
+    warnTransitiveEvictions: Boolean = warnTransitiveEvictions,
+    showCallers: Boolean = showCallers,
+    guessCompatible: Function1[(ModuleID, Option[ModuleID], Option[IvyScala]), Boolean] = guessCompatible): EvictionWarningOptions =
+    new EvictionWarningOptions(configurations = configurations,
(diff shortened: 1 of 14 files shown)


ivy/src/main/scala/sbt/EvictionWarning.scala
```diff
@@ -0,0 +1,200 @@
+package sbt
+
+import collection.mutable
+import Configurations.Compile
+
+final class EvictionWarningOptions private[sbt] (
+    val configurations: Seq[Configuration],
```
jsuereth, line 7: Nice! so it's opt-in then?

ivy/src/main/scala/sbt/EvictionWarning.scala
```diff
@@ -0,0 +1,200 @@
...
+      warnScalaVersionEviction = warnScalaVersionEviction,
+      warnDirectEvictions = warnDirectEvictions,
+      warnTransitiveEvictions = warnTransitiveEvictions,
+      showCallers = showCallers,
+      guessCompatible = guessCompatible)
+}
+
+object EvictionWarningOptions {
+  def default: EvictionWarningOptions =
+    new EvictionWarningOptions(Vector(Compile), true, true, false, false, defaultGuess)
```
jsuereth, line 44: Should we check "test" too?

eed3si9n, line 44: @jsuereth This is the default.

eed3si9n, line 44: I was thinking maybe I should define `test:evicted`. I need to probably prevent `Compile` eviction from showing up twice if `Test` and `Compile` are both in there.

ivy/src/main/scala/sbt/UpdateReport.scala
```diff
@@ -50,25 +63,125 @@ final class ConfigurationReport(val configuration: String, val modules: Seq[Modu
   private[this] def addConfiguration(mod: ModuleID): ModuleID = if (mod.configurations.isEmpty) mod.copy(configurations = Some(configuration)) else mod
 
   def retrieve(f: (String, ModuleID, Artifact, File) => File): ConfigurationReport =
-    new ConfigurationReport(configuration, modules map { _.retrieve((mid, art, file) => f(configuration, mid, art, file)) }, evicted)
+    new ConfigurationReport(configuration, modules map { _.retrieve((mid, art, file) => f(configuration, mid, art, file)) }, details, evicted)
+}
+
+/**
+ * In sbt's terminology, "module" consists of organization, name, and version.
+ * In Ivy's, "module" means just organization and name, and the one including version numbers
```
jsuereth, line 73: This doesn't really describe what the detail report is or what it does or contains.  I still don't understand why it's necessary after readin ghtis.

ivy/src/main/scala/sbt/VersionNumber.scala
```diff
@@ -0,0 +1,147 @@
+package sbt
+
+final class VersionNumber private[sbt] (
+    val numbers: Seq[Long],
+    val tags: Seq[String],
+    val extras: Seq[String]) {
+  def _1: Option[Long] = get(0)
```
jsuereth, line 7: Well, of course it should be called _1....

ivy/src/test/scala/EvictionWarningSpec.scala
```diff
@@ -0,0 +1,224 @@
...
+
+import org.specs2._
+
+class EvictionWarningSpec extends BaseIvySpecification {
+  def is = s2"""
+
+  This is a specification to check the eviction warnings
+
+  Eviction of scala-library whose scalaVersion should
+    be detected                                                 $scalaVersionWarn1
```
jsuereth, line 11: Oh my, real man specs!

main/actions/src/main/scala/sbt/CacheIvy.scala
```diff
@@ -62,16 +63,23 @@ object CacheIvy {
...
-  implicit def confReportFormat(implicit m: Format[String], mr: Format[Seq[ModuleReport]], mi: Format[Seq[ModuleID]]): Format[ConfigurationReport] =
-    wrap[ConfigurationReport, (String, Seq[ModuleReport], Seq[ModuleID])](r => (r.configuration, r.modules, r.evicted), { case (c, m, v) => new ConfigurationReport(c, m, v) })
-  implicit def moduleReportFormat(implicit ff: Format[File]): Format[ModuleReport] =
-    wrap[ModuleReport, (ModuleID, Seq[(Artifact, File)], Seq[Artifact])](m => (m.module, m.artifacts, m.missingArtifacts), { case (m, as, ms) => new ModuleReport(m, as, ms) })
+  implicit def confReportFormat(implicit m: Format[String], mr: Format[Seq[ModuleReport]], mdr: Format[Seq[ModuleDetailReport]], mi: Format[Seq[ModuleID]]): Format[ConfigurationReport] =
+    wrap[ConfigurationReport, (String, Seq[ModuleReport], Seq[ModuleDetailReport], Seq[ModuleID])](r => (r.configuration, r.modules, r.details, r.evicted), { case (c, m, d, v) => new ConfigurationReport(c, m, d, v) })
+  implicit def moduleReportFormat(implicit cf: Format[Seq[Caller]], ff: Format[File]): Format[ModuleReport] = {
```
jsuereth, line 68: The only thing I really really want to check here is what happens when you bump to sbt 0.13.6-SNAPSHOT on an existing project and whether or not we seeamlessy handle the new binary format for the updateReport cache.  I _think_ it should be ok, but would love to know you've tried it.

jsuereth:
This LGTM - minus the failing travis tests.

main/actions/src/main/scala/sbt/CacheIvy.scala
eed3si9n, line 68: I've tested on existing projects like large-graph for testing, and it seemed to work fine. Maybe sbinary can detect content type change?

ivy/src/main/scala/sbt/UpdateReport.scala
eed3si9n, line 73: A sequence of `ModuleDetailReport` called `details` is newly added to `ConfigurationReport`, replacing `modules` and `evicted`. (Note old `evicted` is just a seq of `ModuleID`s).

`ModuleDetailReport` groups the `ModuleReport` of both winners and evicted reports by their `organization` and `name` so I can calculate the eviction warning etc easier later. This should also allow us to report on modules that are completely evicted via exclude rules etc.

ivy/src/main/scala/sbt/VersionNumber.scala
eed3si9n, line 7: Yea calling it `_1` is least unambiguous when you are talking semantics of various version number scheme rather than trying to say major, minor, patch, release, build number etc.

ivy/src/main/scala/sbt/UpdateReport.scala
jsuereth, line 73: Right, can you put that into the scaladoc comment?

ivy/src/main/scala/sbt/VersionNumber.scala
jsuereth, line 7: Yeah, it's ok, but a bit unexpected :)

main/actions/src/main/scala/sbt/CacheIvy.scala
jsuereth, line 68: I think it's more a `try Some(f) catch { case _ => None }` going on...

jsuereth:
So, tests pass. If you modify the javadoc with what you wrote, I'll merge ASAP.

adriaanm:
:+1:

PR #1473, opened 2014-07-29
Improved consolidated resolution testing. #1454

eed3si9n:
Added exclude and override rules, and checks for exact match between project A (with consolidated resolution) and B (without).


jsuereth:
LGTM

PR #1483, opened 2014-07-31
Remove second invalid URL reference

joscha:
There was a duplicate entry for the Google Groups URL and it had whitespace at the end.
This is related to #1482


eed3si9n:
Yea. I noticed it too and fixed it already.

PR #1490, opened 2014-08-01
Fixes NullPointerException during update. Fixes #1484

### LINKED ISSUE #1484: 0.13.6-M1 NullPointerException during update
While trying out the 0.13.6-M1 for #1477 I found a `NullPointerException`. After some debugging, I narrowed down the problem to the following two lines:

``` scala
conflictManager := ConflictManager.strict.copy(organization = "^(?!org\\.scala-lang).*$")

libraryDependencies += "org.specs2" %% "specs2" % "2.3.10-scalaz-7.1.0-M6" % "test"
```

I estimate that we might be the only company in the world with a configuration anything similar to this.

eed3si9n:


ivy/src/main/scala/sbt/IvyRetrieve.scala
```diff
@@ -71,25 +78,29 @@ object IvyRetrieve {
...
-    val branch = Option(revId.getBranch)
+    val branch = nonEmptyString(revId.getBranch)
     val (status, publicationDate, resolver, artifactResolver) = dep.isLoaded match {
       case true =>
-        (Option(dep.getDescriptor.getStatus),
+        (nonEmptyString(dep.getDescriptor.getStatus),
           Some(new ju.Date(dep.getPublication)),
-          Option(dep.getModuleRevision.getResolver.getName),
-          Option(dep.getModuleRevision.getArtifactResolver.getName))
+          nonEmptyString(dep.getModuleRevision.getResolver.getName),
```
jsuereth, line 86: dang, that's a lot of guarding.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1286,7 +1286,16 @@ object Classpaths {
             case (_, Some(out)) if uptodate(inChanged, out) => out
             case _ => work(in)
           }
-          outCache(in)
+          try {
+            outCache(in)
+          } catch {
+            case e: NullPointerException =>
```
jsuereth, line 1292: You just want to catch null pointer exception?  Also, maybe you should log the exception.

jsuereth:
Minor comment on exception handling, otherwise LGTM

main/src/main/scala/sbt/Defaults.scala
eed3si9n, line 1292: amended the commit and -f pushed with log.trace. See line 1299.

jsuereth:
LGTM

PR #1493, opened 2014-08-03
Fixes #1384. Fixes Scope.parseScopedKey

### LINKED ISSUE #1384: scoped key parsing
The regex used in Scope.parseScopedKey

```
  val ScopedKeyRegex = """((\w+)\/)?((\w+)\:)?([\w\-]+)""".r
```

does not parse scoped keys. \w+ is far too restrictive to match URIs, as the string is [defined to contain](http://www.scala-sbt.org/0.13.5/docs/Getting-Started/Scopes.html) - specifically "sbt displays (and parses) scoped keys like this" - and as sbt displays them.

```
scala> Scope parseScopedKey "{file:/mirror/r/java-stream-wrappers/}plot/runtime:classpathConfiguration"
scala.MatchError: {file:/mirror/r/java-stream-wrappers/}plot/runtime:classpathConfiguration (of class java.lang.String)
    at sbt.Scope$.parseScopedKey(Scope.scala:127)

scala> Scope parseScopedKey "plot/runtime:classpathConfiguration"
res1: (sbt.Scope, String) = (Scope(Select(LocalProject(plot)),Select(ConfigKey(runtime)),This,This),classpathConfiguration)
```

eed3si9n:
Fixes #1384.

`Scope.parseScopedKey` now supports full range of legal keys described in the documentation including {.} and other notation for `ProjectRef`, `BuildRef`, and `ThisBuild`.
Just in case any plugin is using `val ScopedKeyRegex = """((\w+)\/)?((\w+)\:)?([\w\-]+)""".r`, I'm leaving that old regex as is and added `ScopedKeyRegex2` in `private[sbt]`.

### DIFF
--- a/main/settings/src/main/scala/sbt/Scope.scala
+++ b/main/settings/src/main/scala/sbt/Scope.scala
@@ -124,12 +124,32 @@ object Scope {
 
   def parseScopedKey(command: String): (Scope, String) =
     {
-      val ScopedKeyRegex(_, projectID, _, config, key) = command
-      val pref = if (projectID eq null) This else Select(LocalProject(projectID))
-      val conf = if (config eq null) This else Select(ConfigKey(config))
-      (Scope(pref, conf, This, This), transformTaskName(key))
+      val ScopedKeyRegex2(_, uriOrNull, _, projectIdOrNull, _, configOrNull, _, inTaskOrNull, key) = command
+      val uriOpt = Option(uriOrNull) map { new URI(_) }
+      val projectIdOpt = Option(projectIdOrNull)
+      val configOpt = Option(configOrNull)
+      val inTaskOpt = Option(inTaskOrNull)
+      val DotURI = new URI(".")
+      val GlobalStr = "*"
+      val scope = (uriOpt, projectIdOpt, configOpt, inTaskOpt) match {
+        case (None, None, Some(GlobalStr), None) => GlobalScope
+        case _ =>
+          val projScope = (uriOpt, projectIdOpt) match {
+            case (Some(DotURI), Some("")) => Select(ThisBuild)
+            case (Some(uri), Some(""))    => Select(BuildRef(uri))
+            case (Some(uri), Some(p))     => Select(ProjectRef(uri, p))
+            case (None, Some(p))          => Select(LocalProject(p))
+            case _                        => This
+          }
+          val configScope = configOpt map { case c if c != GlobalStr => Select(ConfigKey(c)) } getOrElse This
+          val inTaskScope = inTaskOpt map { t => Select(AttributeKey(t)) } getOrElse This
+          Scope(projScope, configScope, inTaskScope, This)
+      }
+      (scope, transformTaskName(key))
     }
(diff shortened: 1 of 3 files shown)


main/settings/src/main/scala/sbt/Scope.scala
```diff
@@ -124,12 +124,32 @@ object Scope {
...
+          }
+          val configScope = configOpt map { case c if c != GlobalStr => Select(ConfigKey(c)) } getOrElse This
+          val inTaskScope = inTaskOpt map { t => Select(AttributeKey(t)) } getOrElse This
+          Scope(projScope, configScope, inTaskScope, This)
+      }
+      (scope, transformTaskName(key))
     }
+  // TODO: Change to ScopedKeyRegex2 on sbt 1.0
   val ScopedKeyRegex = """((\w+)\/)?((\w+)\:)?([\w\-]+)""".r
+  private[sbt] val ScopedKeyRegex2 = """([{](.*?)[}])?((\w*)\/)?(([\w\*]+)\:)?(([\w\-]+)\:\:)?([\w\-]+)""".r
```
jsuereth: This looks like gobbledygook to me.  Any reason we don't just make it a format parser with less crazy?

I'm not 100% happy with the error messages you get if you f up the syntax already, but I'm not sure I'd be able to maintain this regex.

main/settings/src/main/scala/sbt/Scope.scala
```diff
@@ -124,12 +124,32 @@ object Scope {
...
+            case _                        => This
+          }
+          val configScope = configOpt map { case c if c != GlobalStr => Select(ConfigKey(c)) } getOrElse This
+          val inTaskScope = inTaskOpt map { t => Select(AttributeKey(t)) } getOrElse This
+          Scope(projScope, configScope, inTaskScope, This)
+      }
+      (scope, transformTaskName(key))
     }
+  // TODO: Change to ScopedKeyRegex2 on sbt 1.0
   val ScopedKeyRegex = """((\w+)\/)?((\w+)\:)?([\w\-]+)""".r
```
jsuereth, line 153: Mark this deprecated.

main/settings/src/test/scala/ScopedKeySpec.scala
```diff
@@ -0,0 +1,48 @@
...
+class ScopedKeySpec extends Specification {
+  def is = s2"""
+
+  This is a specification to check the scoped key parsing.
+
+  fullClasspath should
+    ${beParsedAs("fullClasspath", ThisScope, "fullClasspath")}
+
+  test:fullClasspath should
+    ${beParsedAs("test:fullClasspath", ThisScope in ConfigKey("test"), "fullClasspath")}
```
jsuereth, line 19: Love these tests.

jsuereth:
The idea LGTM, but I think it's time to drop the Regexes and only have 1 problem.

eed3si9n:
Now with deprecation.

jsuereth:
LGTM, can you rebase so we can see a green CI?

PR #1497, opened 2014-08-04
0.13.5 merge

eed3si9n:
I'm seeing some doc updates in pull req lately.


jsuereth:
That would be nice.

THis LGTM.

PR #1516, opened 2014-08-08
Allow root plugins to be disabled.

### LINKED ISSUE #1455: auto-enabled AutoPlugins with no requires cannot be disabled
If I create an auto-plugin as follows:

``` scala
object BadPlugin extends AutoPlugin {
   def trigger = allRequirements
}
```

Then I cannot disable in `build.sbt`:

``` scala
disablePlugins(BadPlugin)
```

Disabling in the DSL should override the automatic nature of the plugin.

jsuereth:
Fixes #1455
- Add a mechanism to detect if a plugin clause includes/excludes
  a particular plugin.
- Add a filter for our "enabling" clauses so that user-excluded
  root plugins do not show up.
- Add a test to ensure this succeed.

Review by @eed3si9n

### DIFF
--- a/main/src/main/scala/sbt/Plugins.scala
+++ b/main/src/main/scala/sbt/Plugins.scala
@@ -156,8 +156,13 @@ object Plugins extends PluginsFunctions
 			// circular dependencies in the logic.
 			val allRequirementsClause = defined.filterNot(_.isRoot).flatMap(d => asRequirementsClauses(d))
 			val allEnabledByClause = defined.filterNot(_.isRoot).flatMap(d => asEnabledByClauses(d))
+
+      // Note: Here is where the function begins.  We're given a list of plugins now.
 			(requestedPlugins, log) => {
-				val alwaysEnabled: List[AutoPlugin] = defined.filter(_.isAlwaysEnabled)
+        def explicitlyDisabled(p: AutoPlugin): Boolean = hasExclude(requestedPlugins, p)
+				val alwaysEnabled: List[AutoPlugin] = defined.filter(_.isAlwaysEnabled).filterNot(explicitlyDisabled)
+        System.err.println(s"Always Enabled Plugins = ${alwaysEnabled.mkString(", ")}")
+        System.err.println(s"Requested = $requestedPlugins")
 				val knowlege0: Set[Atom] = ((flatten(requestedPlugins) ++ alwaysEnabled) collect {
 					case x: AutoPlugin => Atom(x.label)
 				}).toSet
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/Plugins.scala
```diff
@@ -156,8 +156,13 @@ object Plugins extends PluginsFunctions
 			// circular dependencies in the logic.
 			val allRequirementsClause = defined.filterNot(_.isRoot).flatMap(d => asRequirementsClauses(d))
 			val allEnabledByClause = defined.filterNot(_.isRoot).flatMap(d => asEnabledByClauses(d))
+
+      // Note: Here is where the function begins.  We're given a list of plugins now.
```
eed3si9n, line 154: What kind of indentation setting are you using with IntelliJ?

main/src/main/scala/sbt/Plugins.scala
```diff
@@ -156,8 +156,13 @@ object Plugins extends PluginsFunctions
...
 			(requestedPlugins, log) => {
-				val alwaysEnabled: List[AutoPlugin] = defined.filter(_.isAlwaysEnabled)
+        def explicitlyDisabled(p: AutoPlugin): Boolean = hasExclude(requestedPlugins, p)
+				val alwaysEnabled: List[AutoPlugin] = defined.filter(_.isAlwaysEnabled).filterNot(explicitlyDisabled)
+        System.err.println(s"Always Enabled Plugins = ${alwaysEnabled.mkString(", ")}")
+        System.err.println(s"Requested = $requestedPlugins")
```
eed3si9n: err print not.

main/src/main/scala/sbt/Plugins.scala
eed3si9n, line 154: The rest of the file is using TABs.

main/src/main/scala/sbt/Plugins.scala
```diff
@@ -50,278 +50,296 @@ will activate `MyPlugin` defined above and have its settings automatically added
...
+      // Avoids the requirement for pure Nature strings *and* possible
+      // circular dependencies in the logic.
+      val allRequirementsClause = defined.filterNot(_.isRoot).flatMap(d => asRequirementsClauses(d))
+      val allEnabledByClause = defined.filterNot(_.isRoot).flatMap(d => asEnabledByClauses(d))
+
+      // Note: Here is where the function begins.  We're given a list of plugins now.
+      (requestedPlugins, log) => {
+        def explicitlyDisabled(p: AutoPlugin): Boolean = hasExclude(requestedPlugins, p)
+        val alwaysEnabled: List[AutoPlugin] = defined.filter(_.isAlwaysEnabled).filterNot(explicitlyDisabled)
+        System.err.println(s"Always Enabled Plugins = ${alwaysEnabled.mkString(", ")}")
```
eed3si9n: So it's hard to review now, but here's your println statement.

jsuereth:
@eed3si9n fixed

Issue #321, opened 2012-01-03
SBT doesn't see new SNAPSHOT dependency versions in local maven repos 


tomdz:
I'm using SBT with a local maven repo cache (e.g. `~/.m2/repository`) for a Scala project that depends on a snapshot version of Java library that I develop at the same time. E.g. I add/change something in the java library and use maven to build a new snapshot version of the library. However, no combination of sbt commands (`update`, `clean`, etc.) will make it see the new snapshot version in the local repo - I have to delete the library from the ivy cache every time. It would be great if sbt/ivy would consider the timestamp of the snapshot version and update accordingly.

earldouglas:
I have a similar issue; possibly a duplicate.  https://github.com/harrah/xsbt/issues/335

wpraet:
Hi, any news on this one? Is it fixed in sbt-0.12.x?

absurdhero:
This does not seem to be a duplicate of 335 since running clean does not work as a workaround.
This behavior also remains even if you set the dependencies that are in your local maven repo as changing().

Reproduction Info:

I added the local maven repo and dependencies to sbt with the following lines in build.sbt:

resolvers += "Local Maven Repository" at Path.userHome.asFile.toURI.toURL + ".m2/repository"

libraryDependencies += "com.example.local" % "localproject" % "1.0-SNAPSHOT" changing()

Running "maven clean install" on localproject installs it to ~/.m2/repository

Code changes are not reflected in my sbt project and the ivy cache located at play/repository/cache/com.example.local/localproject/  (since I am using Play)
contains the old copy of the installed package.

snopoke:
I'm also having the same problem and find it very frustrating having to manually clear out the cache to get ivy to see the changes. I'm still using 0.11.2.

jlc:
Hi everyone,

Does anyone has any news about this issue?

I also have it and the only solution which I could find is to change the version number. As you might guess, this is extremely annoying.
Moreover, I have completely removed the directory ~/.ivy2, unfortunately this didn't have any positive effects.

This is just to know if a fix to this issue is currently in progress.

Thanks for your work!

Jeanluc

forceoflight:
Hello,

this issue is still visible in Play Framework 2.1.0 with sbt version 0.12.2. Has someone from the sbt team studied what could cause this issue?

Regards,
Markku

ghost:
Same issue here with Play2 and sbt

David-Noble-at-work:
[...]

PR #1520, opened 2014-08-10
Fixes #1514, #321. Fixes -SNAPSHOT issue by re-implemeting ChainResolver

### LINKED ISSUE #1514: sbt doesn't see new SNAPSHOT dependency versions
This is a general case for #321. 
## steps
1. Create two repositories RepoA and RepoB (or use `Resolver.defaultShared.nonlocal()`)
2. Publish SNAPSHOT to RepoA.
3. Try to use the dependency.
4. Publish SNAPSHOT to RepoB.
5. Try to use the dependency.
## problem

You get the SNAPSHOT from RepoA.
## expectation

You always get the fresh one.
## Notes
- https://github.com/sbt/sbt/tree/fix/issue-321-snapshot-multiple-repo
- [sbt-dirty-money](https://github.com/sbt/sbt-dirty-money)

eed3si9n:
Fixes #1514, #321

Adds `lastestSnapshots` flag to `updateOptions`, which controls the behavior of the chained resolver. Up until 0.13.6, sbt was picking the first `-SNAPSHOT` revision it found along the chain. When  `lastestSnapshots` is enabled (default: `true`), it will look into all resolvers on the chain, and compare them using the publish date.
The tradeoff is probably a longer resolution time if you have many remote repositories on the build or you live away from the severs. So here's how to disable it:

``` scala
updateOptions := updateOptions.value.withLatestSnapshots(false)
```

Ivy by default uses latest-revision as the latest strategy. This strategy I don't think takes in account for the possibility that a changing revision may exist in multiple repositories/resolvers with having identical version number like 0.1.0-SNAPSHOT.
The implementation is a bit hacky, but I think it attacks the core of this problem.


ivy/src/main/scala/sbt/Ivy.scala
```diff
@@ -285,8 +294,187 @@ private object IvySbt {
...
+                Message.verbose(getName + ": Checking cache for: " + dd)
+                Option(findModuleInCache(dd, data, true)) map { mr =>
+                  Message.verbose(getName + ": module revision found in cache: " + mr.getId)
+                  forcedRevision(mr)
+                }
+              }
+            var temp: Option[ResolvedModuleRevision] =
+              if (useLatest) None
+              else resolvedOrCached
+            val resolvers = getResolvers.toArray.toVector collect { case x: DependencyResolver => x }
```
jsuereth, line 326: Why the filter?  What else could be in there?

ivy/src/main/scala/sbt/Ivy.scala
```diff
@@ -285,8 +294,187 @@ private object IvySbt {
...
+                }
+              case None if errors.size > 1 =>
+                val err = (errors.toList map { IvyStringUtils.getErrorMessage }).mkString("\n\t", "\n\t", "\n")
+                throw new RuntimeException(s"several problems occurred while resolving $dd:$err")
+              case _ =>
+                if (resolved == mrOpt) resolved.orNull
+                else (mrOpt map { resolvedRevision }).orNull
+            }
+          }
+        // Ivy seem to not want to use the module descriptor found at the latest resolver
```
jsuereth, line 396: That' the fun stale cache thing I found IIRC.  It tries to keep MDs in memory.

ivy/src/main/scala/sbt/UpdateOptions.scala
```diff
@@ -9,19 +9,26 @@ import java.io.File
  *
  * See also UpdateConfiguration in IvyActions.scala.
  */
-final class UpdateOptions(
+final class UpdateOptions private[sbt] (
+    /** If set to true, check all resolvers for snapshots. */
+    val latestSnapshots: Boolean,
```
jsuereth, line 14: Maybe this should instead be an option of which chain resolver you pick?

I guess that means users wouldn't be able to disable it...

jsuereth:
LGTM!!!!

ivy/src/main/scala/sbt/Ivy.scala
eed3si9n, line 326: Ivy doesn't use Java 5 generics, so we get like `Collection` all the time.

PR #1525, opened 2014-08-12
Fix Eval issues

jsuereth:
Fixes both:
- Hash conflicts when build.sbt files are too similar, leading to incomprehensible error messages
- Littering the `project/target/config-classes` directory with tons of unnecessary junk.

Review by @eed3si9n


main/src/main/scala/sbt/EvaluateConfigurations.scala
```diff
@@ -23,6 +23,11 @@ import scala.annotation.tailrec
  *
  */
 object EvaluateConfigurations {
+
+  type LazyClassLoaded[T] = ClassLoader => T
+
+  private[sbt] case class TrackedEvalResult[T](generated: Seq[File], result: LazyClassLoaded[T])
```
eed3si9n, line 29: Not sure if I like the name `generated`. It sounds like an artificially generated source. `classFiles`?

jsuereth, line 29: Actually, `classFiles` would be a lie as well because it contains a cached object file of data as well.  See my most recent fix commit.

jsuereth:
@eed3si9n This is now passing all scripted tests for me.

main/src/main/scala/sbt/Load.scala
```diff
@@ -238,8 +247,36 @@ object Load {
...
+    // the files aren't properly returned, even though they should be.
+    // TODO - figure out where the caching of whether or not to generate classfiles occurs, and
+    // put cleanups there, perhaps.
+    if (!keepSet.isEmpty) {
+      def keepFile(f: File) = keepSet(f.getCanonicalPath)
+      import Path._
+      val existing = (baseTarget.***.get).filterNot(_.isDirectory)
+      val toDelete = existing.filterNot(keepFile)
+      if (!toDelete.isEmpty) {
+        System.err.println(s"Number of generated files: ${keepSet.toSeq.length}")
```
eed3si9n: println

main/src/main/scala/sbt/Load.scala
```diff
@@ -75,7 +75,16 @@ object Load {
...
       val imports = BuildUtil.baseImports ++ BuildUtil.importAllRoot(config.globalPluginNames)
-      loader => EvaluateConfigurations(eval, files, imports)(loader).settings
+      loader => {
+        val loaded = EvaluateConfigurations(eval, files, imports)(loader)
+        // TODO - Is GC-ing here correct?  Could this break other sbt instances running against old
+        // global classes?  Perhaps...
+        //  In the future we should figure out a way to GC the global directory.  For now,
+        // we leave this as a TODO thanks to possibly mutli-process instabilities we're not ready to solve in
+        // the 0.13 series.
+        //cleanEvalClasses(base, loaded.generatedFiles)
```
eed3si9n: Maybe remove this entire block of comments followed by a commented out code.

main/src/main/scala/sbt/Load.scala
```diff
@@ -238,8 +247,36 @@ object Load {
   def mkEval(classpath: Seq[File], base: File, options: Seq[String]): Eval =
     new Eval(options, classpath, s => new ConsoleReporter(s), Some(evalOutputDirectory(base)))
 
+  /**
+   * This will clean up left-over files in the config-classes directory if they are no longer used.
+   *
+   * @param base  The base directory for the build, should match the one passed into `mkEval` method.
+   */
+  def cleanEvalClasses(base: File, keep: Seq[File]): Unit = {
+    // TODO - Remove printlns.
```
eed3si9n: TODO. Remove TODO comments.

eed3si9n:
Notes please. Other than cosmetic things, LGTM.

main/src/main/scala/sbt/Load.scala
jsuereth: I think we may need to do that in the future.  Wanted to keep note.  We still leak in ~/.sbt

jsuereth:
@eed3si9n Merge conflict resolved, and comments fixed.  I rebased for this.

PR #1526, opened 2014-08-12
HTTPS: Use HTTPS when downloading JARs from Typesafe repo and repo.scala-sbt.org

mxswd:
By using secure URLs, people on your network are not (trivially) able to inject executable code into your dependencies to run on your computer or your build server.

Unfortunately `JavaNet2RepositoryRoot` has no `https` mirror. Is there an alternative mirror or a Scala annotation to warn users (and transitive users) that anyone could execute arbitrary code on their computers?

### DIFF
--- a/ivy/src/main/scala/sbt/Resolver.scala
+++ b/ivy/src/main/scala/sbt/Resolver.scala
@@ -146,11 +146,11 @@ sealed trait JavaNet1Repository extends Resolver {
 object Resolver {
   private[sbt] def useSecureResolvers = sys.props.get("sbt.repository.secure") map { _.toLowerCase == "true" } getOrElse true
 
-  val TypesafeRepositoryRoot = "http://repo.typesafe.com/typesafe"
-  val SbtPluginRepositoryRoot = "http://repo.scala-sbt.org/scalasbt"
+  val TypesafeRepositoryRoot = "https://repo.typesafe.com/typesafe"
+  val SbtPluginRepositoryRoot = "https://repo.scala-sbt.org/scalasbt"
   val SonatypeRepositoryRoot = "https://oss.sonatype.org/content/repositories"
   val JavaNet2RepositoryName = "java.net Maven2 Repository"
-  val JavaNet2RepositoryRoot = "http://download.java.net/maven/2"
+  // val JavaNet2RepositoryRoot = "http://download.java.net/maven/2" // DANGER: unavailable via https.
   val JCenterRepositoryName = "jcenter"
   val JCenterRepositoryRoot = "https://jcenter.bintray.com/"
   val DefaultMavenRepositoryRoot = "https://repo1.maven.org/maven2/"


typesafe-tools:

ivy/src/main/scala/sbt/Resolver.scala
```diff
@@ -146,11 +146,11 @@ sealed trait JavaNet1Repository extends Resolver {
 object Resolver {
   private[sbt] def useSecureResolvers = sys.props.get("sbt.repository.secure") map { _.toLowerCase == "true" } getOrElse true
 
-  val TypesafeRepositoryRoot = "http://repo.typesafe.com/typesafe"
-  val SbtPluginRepositoryRoot = "http://repo.scala-sbt.org/scalasbt"
+  val TypesafeRepositoryRoot = "https://repo.typesafe.com/typesafe"
+  val SbtPluginRepositoryRoot = "https://repo.scala-sbt.org/scalasbt"
```
jsuereth, line 150: We actually don't have a legitimate SSL certificate here yet, which breaks on some JDKs, unfortunately.  We're still waiting on a scala-sbt.org SSL cert, but once that's here we could merge this.

mxswd, line 150: Thanks, sounds good to me. Also the `JavaNet2RepositoryRoot` is referenced in that file, so that also needs to get fixed.

ivy/src/main/scala/sbt/Resolver.scala
```diff
@@ -146,11 +146,11 @@ sealed trait JavaNet1Repository extends Resolver {
...
   val SonatypeRepositoryRoot = "https://oss.sonatype.org/content/repositories"
   val JavaNet2RepositoryName = "java.net Maven2 Repository"
-  val JavaNet2RepositoryRoot = "http://download.java.net/maven/2"
+  // val JavaNet2RepositoryRoot = "http://download.java.net/maven/2" // DANGER: unavailable via https.
```
eed3si9n, line 153: Commenting this out would break binary compatibility with sbt 0.13.0.

ivy/src/main/scala/sbt/Resolver.scala
eed3si9n, line 150: Typesafe repo is rekeyed. I am now waiting on repo.scala-sbt.org.
I am all for using HTTPS by default, but I would also like to provide an escape hatch to revert back to HTTP in case something wrong happens. See sbt/sbt@f3d71f98e486f5b5ae42fbddfb8cb2b6cd46f0e3. Could you please add `private[sbt]` method `typesafeRepositoryRoot` and `sbtPluginRepositoryRoot` similar to `centralRepositoryRoot`?

eed3si9n:
Closing this since I've implemented #1548 and fix/1541-scala-sbt branch.

ivy/src/main/scala/sbt/Resolver.scala
mxswd, line 153: Could we deprecate it this release then remove it next?

mxswd:
@eed3si9n cool! Well done, looks great!

PR #1529, opened 2014-08-13
An attempt to speed up travis CI by unziping a pregenerated ~/.ivy2/cach...

jsuereth:
...e file.

Also includes the script to generate a new tar.bz2 which we should do anytime
we bump revisions significantly.

Review by @eed3si9n


jsuereth:
This wont' work as my dropbox url was shutdown.

PR #1532, opened 2014-08-15
Fix NPE during update again

eed3si9n:
Ivy gives an array that contains null for caller configurations.
sbinary barfs when it sees null. Curiously two of the sbt plugins that
hit this bug happens to be from Typesafe:

```
addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.3.2")
addSbtPlugin("com.typesafe.sbt" % "sbt-native-packager"  % "0.7.3")
```


jsuereth:
Lgtm

PR #1536, opened 2014-08-16
HTTPS: Fetch typesafe artifacts using https

benmccann:
See http://blog.ontoillogical.com/blog/2014/07/28/how-to-take-over-any-java-developer/

Seems some people are having problems with typesafe's ssl cert, so may want to fix that before comitting this. Sounds like [the solution](http://stackoverflow.com/questions/18746565/godaddy-ssl-cert-not-working-with-java/21117993#21117993) for Java not supporting GoDaddy G2 certs is for typesafe to rekey it's SSL cert using SHA-1
http://support.godaddy.com/help/article/4976/re-keying-an-ssl-certificate

Hopefully that's not too much trouble. If it ends up being tricky, could also just buy an SSL cert from someplace else. They start at $4.99/yr from ssls.com

### DIFF
--- a/launch/src/main/resources/sbt.boot.properties0.10.0
+++ b/launch/src/main/resources/sbt.boot.properties0.10.0
@@ -11,7 +11,7 @@
 
 [repositories]
   local
-  typesafe-ivy-releases: http://repo.typesafe.com/typesafe/ivy-releases/, [organization]/[module]/[revision]/[type]s/[artifact](-[classifier]).[ext]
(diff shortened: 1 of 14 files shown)


typesafe-tools:

edwardcallahan:
We can rekey.

edwardcallahan:
SHA-1 cert for repo.typesafe.com has been installed. Cheers to Alex @ JFrog for the quick turnaround on the install of the updated cert.

benmccann:
Awesome. Thanks @edwardcallahan !

PR #1547, opened 2014-08-22
Use the new Docker queue on Travis

joshk:
faster, better, stronger

vms boot in seconds, network is more stable

also cleaned up the .travis.yml file a little :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,3 +35,4 @@ script:
 notifications:
   email:
     - qbranch@typesafe.com
+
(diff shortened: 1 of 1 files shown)


typesafe-tools:

eed3si9n:
sounds fun! thanks for this. can't wait to see https://travis-ci.org/sbt/sbt/builds/33257634 go green.

eed3si9n:
Thus far bunch of jobs are dying 4~5 minutes in.

joshk:
yeah, this is an interesting one, looks like it might be ram/memory related.

will dive in deeper soon.

PR #1561, opened 2014-08-31
Fix url to sbt-dev group

jamesward:
There is a unicode char at the end of that line.


typesafe-tools:

PR #1408, opened 2014-06-14
Never inspect twice the same macro application, fix (again) #1237

Duhemm:
Consider the following macro :

``` scala
def foo(a: Any): Any = macro impl
def impl(c: Context)(a: c.Expr[Any]): c.Expr[Any] = a
```

In Scala 2.10.4 (I've not observed this with any other version of Scala), an application of this macro such as :

``` scala
foo(someVal)
```

will expand to `someVal`. If we inspect this expansion, we will find that `someVal` is attached the original tree `foo(someVal)`, as expected. However, if we inspect the components of this tree, we will find a cyclic chain of `MacroExpansionAttachment` : The same original tree is attached to `someVal`, and so on, until the stack overflows.

The solution used here consists to filter out the nodes that point to the current node in the `original` tree (I think that the diff may be clearer than my sentence).

Thanks to @dwijnand for reporting the bug !

/cc @xeno-by

Fixes #1237 (again).

PR #1563, opened 2014-09-01
Fix #1544 (SO in dependencies extraction with macros and name hashing)

### LINKED ISSUE #1544: Name hashing: StackOverflow with macros
I found another instance of problem reported in #1237. If you enable name hashing in sbt's `source-dependencies/macro` scripted test then you get:

```
[info] java.lang.StackOverflowError
[info]  at scala.reflect.internal.Trees$class.itraverse(Trees.scala:1098)
[info]  at scala.reflect.internal.SymbolTable.itraverse(SymbolTable.scala:13)
[info]  at scala.reflect.internal.SymbolTable.itraverse(SymbolTable.scala:13)
[info]  at scala.reflect.api.Trees$Traverser.traverse(Trees.scala:2825)
[info]  at xsbt.Dependency$ExtractDependenciesByMemberRefTraverser.traverse(Dependency.scala:149)
[info]  at xsbt.Dependency$ExtractDependenciesByMemberRefTraverser.traverse(Dependency.scala:104)
[info]  at scala.reflect.internal.Trees$class.itraverse(Trees.scala:1174)
[info]  at scala.reflect.internal.SymbolTable.itraverse(SymbolTable.scala:13)
[info]  at scala.reflect.internal.SymbolTable.itraverse(SymbolTable.scala:13)
[info]  at scala.reflect.api.Trees$Traverser.traverse(Trees.scala:2825)
[info]  at xsbt.Dependency$ExtractDependenciesByMemberRefTraverser.traverse(Dependency.scala:149)
[info]  at xsbt.Dependency$ExtractDependenciesByMemberRefTraverser.traverse(Dependency.scala:146)
[info]  at xsbt.Dependency$ExtractDependenciesByMemberRefTraverser.traverse(Dependency.scala:104)
[info]  at scala.reflect.api.Trees$Traverser.traverseTrees(Trees.scala:2829)
[info]  at scala.reflect.internal.Trees$class.itraverse(Trees.scala:1174)
[info]  at scala.reflect.internal.SymbolTable.itraverse(SymbolTable.scala:13)
[info]  at scala.reflect.internal.SymbolTable.itraverse(SymbolTable.scala:13)
[info]  at scala.reflect.api.Trees$Traverser.traverse(Trees.scala:2825)
[...]
```

Duhemm:
This PR addresses these bugs :
- #1237 
- #1408 
- #1544 

The fix that was applied to `compile/interface/src/main/scala/xsbt/Dependency.scala` was unfortunately not sufficient, and sbt could still recurse infinitely over some nodes that had a cyclic chain of original-expanded trees.

To fix this problem, I ported the fix that was applied to `compile/interface/src/main/scala/xsbt/ExtractUsedNames.scala`, which required me to slightly change how dependencies are extracted, so that we can perform a more aggressive filter on nodes that we have already inspected.


typesafe-tools:

gkossakowski:
Hi Martin!

Thanks for the fix! What's the rationale for introducing `extractByTreeWalk` method? Why using `traverse` method doesn't work?

Please squash both commits into one. The general practice is to change tests and code in single commit. This way it's easier to see how changes to code affect test status.

Duhemm:
Hi ! Commits are squashed ;)

The problem with using `traverse` is that we can't filter out all the children of a node that are equal to the node we are currently inspecting.

This sentence is most likely unclear. I'll comment directly on the diff, it will be easier.

compile/interface/src/main/scala/xsbt/Dependency.scala
```diff
@@ -92,68 +92,85 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile {
...
+
+      def handleTreeNode(node: Tree): Unit = {
+
+        def handleMacroExpansion(original: Tree): Unit = {
+          // Some macros seem to be their own orignal tree, or appear in the children of their
+          // original tree. To prevent infinite loops, we need to filter out nodes that we already
+          // handled.
+          // This is only relevant for Scala 2.10.4
+          // See https://issues.scala-lang.org/browse/SI-8486
+          original.filter(_ ne node).foreach(handleTreeNode)
```
Duhemm: Here we need to keep only the children that are not equal to the node that we are inspecting.

Imagine a macro expansion `#1` that had an original tree `#1`. This problem was successfully solved by the previous fix, because we had `#1 == #1`, and we wouldn't inspect the original tree any further.

Imagine now a macro expansion `#1` that had an original tree `#2`, that had an original tree `#1`. With the previous fix, we have `#1 != #2`, so we start `traverse`-ing `#2`, and we have no reference to `#1` in scope, which means that we won't be able to detect it if we end up inspecting it again (which will happen, since the original of `#2` is `#1`).

Using this fix, when we see that `#1` has an original tree `#2`, we filter out all the subtrees of `#2` that are equal to `#1`, and we don't end up looping over the same part of the tree.

I hope this is clearer.

gkossakowski: Ok. That explains why using `filter` (that is recursive) makes sense. However, I still don't understand what was wrong with using `traverse` and filtering the nodes that we would recurse on.

Duhemm: In the scripted test `source-dependencies/macro`, we get the following macro expansion : `0` (let's call it `exp`). Seeing that this is a macro expansion, we get that its original tree `orig` is `Provider.tree(0)`.

Now, we can `filter` from `orig` all trees that equal to `exp`. We get the following trees :
- `Provider.tree(0)` (no original tree attached, not considered a macro expansion)
- `Provider.tree` (no original tree attached, not considered a macro expansion)
- `Provider` (no original tree attached, not considered a macro expansion)
- ~~`0` (considered a macro expansion from `Provider.tree(0)`)~~ **filtered out**

If we use `traverseTrees` on this list, not only will we traverse multiple times some nodes, but we will also end up traversing again `0` (because it appears in `Provider.tree(0)`, which will be seen as a macro expansion and make us recurse infinitely.

Am I missing something ?

gkossakowski: Well, what I meant is that instead of dropping traverse one can just record (in a private, mutable set) already visited originals and not process them again. This is what we tend to do when dealing with cyclic data structures in sbt. See: https://github.com/sbt/sbt/blob/0.13/compile/api/src/main/scala/xsbt/api/HashAPI.scala#L52

The HashAPI is overly generic because it has multiple different ways to cycle. For us, a simple mutable set is enough.

I find the mutable set + simple guard a bit easier to reason about because you know that there won't be looping no matter how long cycle chains are. For example, I can't see why your patch would handle `Provider.tree(Provider.tree(0))` properly and not loop.

gkossakowski:
Ping @Duhemm. What do you think about my last comment?

Duhemm:
@gkossakowski Yes, this solution is definitely simpler and easier to understand, thanks ! Is everything else all right ?

gkossakowski:
LGTM!
Excellent comments and I like the smaller diff! :ok_hand:

PR #1573, opened 2014-09-04
Allow the "-bin" Scala version suffix to specify a bincompat version

cunei:
There is sometimes the need to use a test version of Scala that is
intended to be binary compatible with a standard release version.

At this time, due to the particular logic implemented within sbt,
a non-numeric suffix will never have the same binaryScalaVersion
of a release version ending in ".0", like for instance "2.11.0".

This commit allows developers to use as suffix any string that
begins with "-bin", for instance "2.11.0-bin-compat-test-1".
Such a suffix will have a binaryScalaVersion of "2.11", being
therefore considered binary compatible with release versions.


jsuereth:
LGTM

puffnfresh:
:+1:

PR #1574, opened 2014-09-04
Natural whitespace handling for SBT configuration parser AKA 'no more blankies'

ajozwik:
The idea is to remove limitation with blank lines in sbt file.

For backward compatibility only one method is modified -  EvaluateConfigurations.splitExpressions.
Most pull request contains test and sbt files.
The father of idea is  @gkossakowski, solution was implemented by @WarsawScala members in project https://github.com/WarsawScala/sbtchallenge


typesafe-tools:

eed3si9n:
This is huge (in terms of impact)! Thanks @gkossakowski and @WarsawScala!

main/src/main/scala/sbt/SplitExpressionsNoBlankies.scala
```diff
@@ -0,0 +1,385 @@
...
+    import scala.compat.Platform.EOL
+    import BugInParser._
+    import XmlContent._
+
+    val mirror = universe.runtimeMirror(this.getClass.getClassLoader)
+    val toolbox = mirror.mkToolBox(options = "-Yrangepos")
+    val indexedLines = lines.toIndexedSeq
+    val original = indexedLines.mkString(END_OF_LINE)
+    val merged = handleXmlContent(original)
+    val fileName = if (file == null) "Here should be file name" else file.getAbsolutePath
```
jsuereth: Do you know when null is passed in?

I'd prefer to see Option[File] is this can be null, and additionally, I'd prefer if we could pass in a fake filename.

main/src/main/scala/sbt/SplitExpressionsNoBlankies.scala
```diff
@@ -0,0 +1,385 @@
...
+
+    val parsed =
+      try {
+        toolbox.parse(merged)
+      } catch {
+        case e: ToolBoxError =>
+          val seq = toolbox.frontEnd.infos.map { i =>
+            s"""[$fileName]:${i.pos.line}: ${i.msg}"""
+          }
+          throw new MessageOnlyException(
```
jsuereth, line 42: Any reason why you drop the stacktrace?  Sometimes we need those when debugging an issue between sbt/scalac.

main/src/main/scala/sbt/SplitExpressionsNoBlankies.scala
```diff
@@ -0,0 +1,385 @@
...
+          case util.Failure(th) =>
+            val missingText = tryWithNextStatement(merged, t.pos.end, t.pos.line, fileName, th)
+            originalStatement + missingText
+          case _ =>
+            originalStatement
+        }
+        val numberLines = statement.count(c => c == '\n')
+        Some((statement, LineRange(t.pos.line - 1, t.pos.line + numberLines)))
+      } else {
+        None
```
jsuereth, line 87: what does it mean if you have a statement with position?

main/src/main/scala/sbt/SplitExpressionsNoBlankies.scala
```diff
@@ -0,0 +1,385 @@
...
+      case _         => false
+    }
+
+    def convertImport(t: Tree): (String, Int) =
+      (merged.substring(t.pos.start, t.pos.end), t.pos.line - 1)
+
+    def convertStatement(t: Tree): Option[(String, LineRange)] =
+      if (t.pos.isDefined) {
+        val originalStatement = merged.substring(t.pos.start, t.pos.end)
+        val statement = util.Try(toolbox.parse(originalStatement)) match {
```
jsuereth: What situation is this trying to solve?  You're already parsing previously, can you add a comment to the effect of what this additional try is doing?

main/src/main/scala/sbt/SplitExpressionsNoBlankies.scala
```diff
@@ -0,0 +1,385 @@
...
+      }
+
+    (imports map convertImport, statements flatMap convertStatement)
+  }
+
+}
+
+/**
+ * Scala parser cuts last bracket -
+ * @see http://stackoverflow.com/questions/25547149/scala-parser-cuts-last-bracket
```
jsuereth, line 96: Ah, I'd like to see this note above as well, which is where folks will be looking when the start debugging before they reach here.

jsuereth:
All in all, nice work!   I have a questions/clarifications.  Particularly, want to see how this works with `set` and `session save` commands.

main/src/main/scala/sbt/SplitExpressionsNoBlankies.scala
ajozwik, line 42: - Original ToolBoxError contains less information than we need (information about line number is missing). See [scala.tools.reflect.ToolBoxFactory](https://github.com/scala/scala/blob/2.10.x/src/compiler/scala/tools/reflect/ToolBoxFactory.scala)

``` scala
 def throwIfErrors() = {
        if (frontEnd.hasErrors) {
          var msg = "reflective compilation has failed: " + EOL + EOL
          msg += frontEnd.infos map (_.msg) mkString EOL
          throw ToolBoxError(msg)
        }
      }
```
-  MessageOnlyException does not have constructor with Throwable.
-  I can print the error. Which logger should I use? `ConsoleLogger(System.err).trace(e)`?

main/src/main/scala/sbt/SplitExpressionsNoBlankies.scala
ajozwik: I will remove checking null. File always is set

main/src/main/scala/sbt/SplitExpressionsNoBlankies.scala
ajozwik, line 87: Please clarification. 
Object t:Tree can be or statement (or definition) with position define, or  or statement (or definition) with NoPostion (then t.pos.start throw exception).

main/src/main/scala/sbt/SplitExpressionsNoBlankies.scala
ajozwik: See next comment. There is a bug in parser end position

main/src/main/scala/sbt/SplitExpressionsNoBlankies.scala
ajozwik, line 96: It is work-around

som-snytt, line 96: When I saw that comment, I went, Hey I read that SO question!  I also tried/failed to come up with a helpful suggestion.  In the past, folks have joked about adding a char of text, reparsing until something changes. Anyway, it's fun to see SO in the wild.

ajozwik:
For line

``` scala
scalaVersion := "2.11.1"; name := "ala"
```

when you type:

``` scala
set name  := "newName"
```

And call `session save` you will receive

``` scala
 name  := "newName"
```

scalaVersion is dropped.

jsuereth:
@ajozwik We'll have to look into fixing this then....   Possibly by having session use position more-directly.   Because of this, I'll talk to @eed3si9n.  We may want to pull this into a branch, work out the issues with "session save" and any `sbt-release` plugin things that may arise, then merge into 0.13.

Again, great work!  I'm super super excited about this parser.

jsuereth:
@ajozwik I created an sbt/sbt branch (wip/no-blankies) for continuing work on the `set` and `session save` command here.   I think we may need to expose `Position` through `Setting[_]` to fix issues around here, or some other workaround.  We can either leave this PR open, or we can continue work on that branch.

ajozwik:
Could you send us the acceptance tests? What is the expected behaviour? For session save we have to rewrite only the value/definition - not all line. I think only save/save-all have to be adapted.

jsuereth:
@ajozwik I'll see if I can write them tomorrow for you and put them in the wip/no-blankies branch on sbt/sbt.

jsuereth:
Sorry, it's taking me a while.

Basically, I'd like to see session save work.  Again, we can close those PR if you'd like as I've already merged it into a "wip" branch on sbt/sbt and we (the sbt team) will make sure this feature makes it into sbt 0.13.7.  Kudos on some really great work.   

And yes, session save/save-all needs to be checked, and whatever we do to fix it, we'll want to verify works with the sbt-release plugin.  If you're looking into it, let us know so we don't schedule the work for next sprint (starts next week).

Thanks!

ajozwik:
@jsuereth I changed "session save" implementation. Lines 116 - 124 from  SessionSettings.scala has been replaced with: SessionSettingsNoBlankies.oldLinesToNew method. Code seems to work. I tested mainly by sbt console.
I create sbt.SessionSettingsSpec . This test scans session-settings folder. For example there is:
- 1.sbt.txt - input sbt file
- 1.sbt.txt_1/1.xml - input setting - here `name := "alaMaKota"` from LineRange( 1, 2)
- 1.sbt.txt_1/1.xml.result - expected result

You can add your filed to this folder with conventions 
- <file_name>.sbt.txt - input sbt file
- <file_name>.sbt.txt_1/<input_xml_name>.xml - input setting - here `name := "alaMaKota"` from LineRange( 1, 2)
- <file_name>.sbt.txt_1/<input_xml_name>.xml.result - expected result

I will add more test files. To quick test run:

```
sbt
project main
~testOnly sbt.SessionSettingsSpec
```

main/src/main/scala/sbt/SplitExpressionsNoBlankies.scala
```diff
@@ -0,0 +1,401 @@
...
+    } else {
+      val c = content.charAt(index)
+      if (c == '/' && content.size > index + 1) {
+        val nextChar = content.charAt(index + 1)
+        if (nextChar == '/') {
+          val endOfLine = content.indexOf('\n', index)
+          findFirstNotBlankNotCommentedIndex(content, endOfLine)
+        } else {
+          //if (nextChar == '*')
+          val endOfCommented = content.indexOf("*/", index + 1)
```
dragos, line 136: I wonder if this code handles Scala nested comments properly. Why not use the Scala tokenizer instead of parsing comments?

ajozwik, line 136: My mistake - this comments are in code - which can not be parsed. You right - this method can be written using tokenizer. Could you send me, what is `Scala tokenizer` ?  I can use just split("""//""") for // comments.

ajozwik:
In travis build failed. But it is problem with travis and communication with repository. I think the code is ready to test (with `session save`).

ajozwik:
Create new pull request on branch wip/no-blankies #1604

main/src/main/scala/sbt/SplitExpressionsNoBlankies.scala
ajozwik, line 136: @dragos, please send me example of code with comments. I open #1606 with minor tasks. I can test with scala parser - but I not sure - to test all cases. How scala parse nested comments? Maybe I can reuse existing function?

dragos, line 136: I'm sorry, I missed the comment. The Scala tokenizer is actually called Scanner, so again it is my bad :)

Have a look at [Scanners.skipNestedComments](https://github.com/scala/scala/blob/2.11.x/src/compiler/scala/tools/nsc/ast/parser/Scanners.scala#L131) to see how they do it. Since you have a compiler around, you can easily get all the tokens from a source file. For example, we do that in the IDE [here](https://github.com/scala-ide/scala-ide/blob/master/org.scala-ide.sdt.core/src/org/scalaide/core/compiler/CompilerApiExtensions.scala#L63)

ajozwik, line 136: Thank you, I will look at it. The problem is (I think not so big) that sbt uses scala 2.10.4 ...
I tried to correct scala parser. I will create pull request - if tests pass. Only for scala 2.11.3. There are big changes between versions. If it will be needed I will try to correct 2.10.5 also (maybe sbt 0.13.7 will use it).

ajozwik, line 136: @dragos I made work around. We have knowledge that statement is correct - it was parsed. Only bracket is missing. So now I search for "}" or ")", see https://github.com/sbt/sbt/blob/wip/no-blankies/main/src/main/scala/sbt/internals/parser/SplitExpressionsNoBlankies.scala#L111 line 111 and related.

PR #1581, opened 2014-09-07
Fix issue where generated config-classes was not propogated.

jsuereth:
Review by @eed3si9n 

There are no addition notes because this is a regression caused by the fix for config file classes.


eed3si9n:
Merged!

PR #1584, opened 2014-09-09
Fix issues with specifying scalaHome/scalaInstance and running tests

jsuereth:
Fixes an issue where the classpath filter was blocking loading Scala classes inside test-runners, thanks to an invalid classpath list.

cc> @paulp Any chance you've seen this issue?  I just discovered it the other day, but given your usage of sbt, I think you may be the only one I know who would have hit this, may be able to validate it outside the Akka build/test case.

Review by @cunei cc> @eed3si9n @rkuhn


paulp:
Don't know. With sbt I almost never know if something is a bug or my inability to understand how it's supposed to work. You're right this is the sort of thing I would likely have seen, but if I can trial-and-error my way to something which does what I want I discard the failed paths.

sbt/src/sbt-test/tests/scala-instance-classloader/src/test/scala/Test.scala
```diff
@@ -0,0 +1,21 @@
...
+class BadTest {
+
+  @Test
+  def testCpIssue(): Unit = {
+    // TODO - This is merely the laziest way to run the test.  What we want to do:
+    // * Load something from our own classloader that's INSIDE the scala library
+    // * Try to load that same something from the THREAD CONTEXT classloader.
+    // * Ensure we can do both, i.e. the second used to be filtered and broken.
+    val current = Thread.currentThread.getContextClassLoader
+    val mine = this.getClass.getClassLoader
```
rkuhn, line 14: I guess these two are left-overs from debugging? (and the comment also seems to be outdated)

rkuhn:
LGTM (but I am not an sbt expert); in any case I applaud the `toString` methods on the classloaders.

sbt/src/sbt-test/tests/scala-instance-classloader/src/test/scala/Test.scala
jsuereth, line 14: Basically.   Eventually I want to NOT use akka for this test, which is why the comment is there.  "Yours" vs. "Mine" was to try to make that happen directly (i.e. without akka).

cunei:
The definition of `isManagedVersion` is a a bit fragile, but within the current restrictions LGTM. Cheers!

jsuereth:
Thanks @cunei !

PR #1586, opened 2014-09-09
Fix ivy-mvn cross-wiring issues

jsuereth:
This allows more interplay between ivy/maven than was otherwise available.  Also fixed a potentially epic hole in transitive maven dependencies where more artifacts than are strictly desired get loaded.

Review by @eed3si9n


jsuereth:
@eed3si9n ready for another look.

eed3si9n:
Code looks good to me. Notes plz.

jsuereth:
@eed3si9n Can you fix the notes thing?

jsuereth:
@eed3si9n I added notes, as well as a note about adding notes.  I didn't follow what we did in 0.13.6 because I can't stand constant conflicts.   Instead, I think we work on a utility to pull notes of this style into a combined notes when issuing RCs.

jsuereth:
Note this also fixes: #1592

PR #1590, opened 2014-09-11
Add developer key

jedesah:
The main motivation for this change is to get rid of the mandatory pomExtra that everyone is forced to use to publish to maven central.

As an added benefit, it sould be easy enough to implement a plugin that depends on the git plugin to allow to fill in the developers list using the git commit log.

Originated from this stackoverflow question: http://stackoverflow.com/questions/25630281/is-there-a-reason-why-sbt-does-not-support-the-developer-key
Figured this should be easy enough and I could do it myself instead of just doing a ticket.


typesafe-tools:

jsuereth:
Very much a fan of adding this.
1.  Is this hooked into the Pom generation in sbt?  If not, we'll need to do that
2.  Binary compatibility.    If you alter a constructor, you actually need to provide an overridden construct which delegates from the previous one to the new one. (Yeah, it's annoying.... if you need help I can show you where/how).

Otherwise, good work!

ivy/src/main/scala/sbt/MakePom.scala
```diff
@@ -100,6 +101,21 @@ class MakePom(val log: Logger) {
         case _ => NodeSeq.Empty
       }
     }
+  def makeDeveloperInfo(moduleInfo: ModuleInfo): NodeSeq =
```
jsuereth, line 104: Ah, nevermind, I missed this.

ivy/src/main/scala/sbt/IvyInterface.scala
```diff
@@ -10,7 +10,7 @@ import org.apache.ivy.plugins.resolver.{ DependencyResolver, IBiblioResolver }
 import org.apache.ivy.util.url.CredentialsStore
 
 /** Additional information about a project module */
-final case class ModuleInfo(nameFormal: String, description: String = "", homepage: Option[URL] = None, startYear: Option[Int] = None, licenses: Seq[(String, URL)] = Nil, organizationName: String = "", organizationHomepage: Option[URL] = None, scmInfo: Option[ScmInfo] = None) {
+final case class ModuleInfo(nameFormal: String, description: String = "", homepage: Option[URL] = None, startYear: Option[Int] = None, licenses: Seq[(String, URL)] = Nil, organizationName: String = "", organizationHomepage: Option[URL] = None, scmInfo: Option[ScmInfo] = None, developers: Seq[Developer] = Seq()) {
```
jsuereth, line 13: Here is where we need to provide a new `def this(... everythign but devleopers)` method to retain binary compatibility, otherwise this can't hit until sbt 1.0

ivy/src/main/scala/sbt/IvyInterface.scala
```diff
@@ -20,6 +22,8 @@ final case class ModuleInfo(nameFormal: String, description: String = "", homepa
 /** Basic SCM information for a project module */
 final case class ScmInfo(browseUrl: URL, connection: String, devConnection: Option[String] = None)
 
+final case class Developer(id: String, name: String, email: String, url: URL)
```
jedesah, line 25: Is this the right place to put this? I am not even sure if Ivy has a concept of Developer?

jsuereth, line 25: No, but this is OUR interface, so this is fine.  We may hide ivy even further in the future, in which case this location becomes "canon" for dependnecy management in sbt, so yes this is right.

jedesah, line 25: Okay, sounds good!

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1042,7 +1042,7 @@ object Classpaths {
     organization <<= organization or normalizedName,
     organizationName <<= organizationName or organization,
     organizationHomepage <<= organizationHomepage or homepage,
-    projectInfo <<= (name, description, homepage, startYear, licenses, organizationName, organizationHomepage, scmInfo) apply ModuleInfo,
+    projectInfo <<= (name, description, homepage, startYear, licenses, organizationName, organizationHomepage, scmInfo, developers) apply ModuleInfo,
```
jsuereth, line 1046: So the test failures are legitimate, right now.  I think you want a setitng like:

```
developers <<= developers ?? Nil
```

in here as well, as right now the default build does not define the developers setting at all, so all the tests explode.

eed3si9n:
Could you add a release note entry in `notes/0.13.7`?

ivy/src/main/scala/sbt/IvyInterface.scala
```diff
@@ -10,7 +10,9 @@ import org.apache.ivy.plugins.resolver.{ DependencyResolver, IBiblioResolver }
 import org.apache.ivy.util.url.CredentialsStore
 
 /** Additional information about a project module */
-final case class ModuleInfo(nameFormal: String, description: String = "", homepage: Option[URL] = None, startYear: Option[Int] = None, licenses: Seq[(String, URL)] = Nil, organizationName: String = "", organizationHomepage: Option[URL] = None, scmInfo: Option[ScmInfo] = None) {
+final case class ModuleInfo(nameFormal: String, description: String = "", homepage: Option[URL] = None, startYear: Option[Int] = None, licenses: Seq[(String, URL)] = Nil, organizationName: String = "", organizationHomepage: Option[URL] = None, scmInfo: Option[ScmInfo] = None, developers: Seq[Developer] = Seq()) {
+  def this(nameFormal: String, description: String = "", homepage: Option[URL] = None, startYear: Option[Int] = None, licenses: Seq[(String, URL)] = Nil, organizationName: String = "", organizationHomepage: Option[URL] = None, scmInfo: Option[ScmInfo] = None) =
```
jsuereth: You don't want the default parameters in this version.  That's why you're getting compilation failures, it's a  pretty obtuse error message though:

```
[error] /home/travis/build/sbt/sbt/ivy/src/main/scala/sbt/IvyInterface.scala:14: double definition:
[error] method <init>$default$8:=> Option[sbt.ScmInfo] and
[error] method <init>$default$8:=> Option[sbt.ScmInfo] at line 13
[error] have same type
[error]   def this(nameFormal: String, description: String = "", homepage: Option[URL] = None, startYear: Option[Int] = None, licenses: Seq[(String, URL)] = Nil, organizationName: String = "", organizationHomepage: Option[URL] = None, scmInfo: Option[ScmInfo] = None) =
[error]                                                                                                                                                                                                                                    ^
[error] /home/travis/build/sbt/sbt/ivy/src/main/scala/sbt/IvyInterface.scala:14: double definition:
[error] method <init>$default$7:=> Option[java.net.URL] and
[error] method <init>$default$7:=> Option[java.net.URL] at line 13
[error] have same type
[error]   def this(nameFormal: String, description: String = "", homepage: Option[URL] = None, startYear: Option[Int] = None, licenses: Seq[(String, URL)] = Nil, organizationName: String = "", organizationHomepage: Option[URL] = None, scmInfo: Option[ScmInfo] = None) =
[error]                                                                                                                                                                                          ^
```

Basically, this method is _only_ for binary compatibility, so you don't need the syntactic convenience of the default parameters anyway.  We're just trying to make sure binary calls to the old constructor don't explode at runtime :)

jsuereth:
SOO close, sorry for the obtuse error message.  You find a nice corner of scala (default parameters + overridding => bad times).  If only the overridding wasnt' necessary for binary compatibility :(

jedesah:
Hopefully this does it. @jsuereth Thx so much for all the help getting this to work!!

jsuereth:
LGTM!  Thanks much for the fix!!!

benhutchison:
To be honest, while Im 100% with the intent of this feature, I think this PR has a few problems:
- the setting key is defined `List[Developer]` instead of `Seq[Developer]`
- Needed better release notes, should include simple example how to use the setting. I couldnt find a single person using this setting "in the wild" and maybe insufficient docs is the reason.
- The `email` is marked mandatory. I dont like revealing my email for spam reasons, and havent ever included it, and [Maven Central seems fine with that](https://repo1.maven.org/maven2/com/github/benhutchison/prickle_2.12/1.1.13/prickle_2.12-1.1.13.pom). Better if it were optional...

At a more meta level, I cant help noticing there still alot of duplicated info between `homepage`, `scmInfo` and `developers`. When SBT `name` == github `name`, all that'd be strictly required to generate the pom is the github account name, plus any contributor IDs, eg 

```
github := GithubInfo("benhutchison", contributor = Seq("cmyers"))
```

PS @OlegIlyenko Cant reproduce your report, [this build.sbt](https://github.com/benhutchison/mouse/blob/106cc7f59ced321b8add0aab5d821ee239b86644/build.sbt#L20) generates:

```
    <developers>
        <developer>
            <id>benhutchison</id>
            <name>Ben Hutchison</name>
            <email>brhutchison@gmail.com</email>
            <url>https://github.com/benhutchison</url>
        </developer>
    </developers>
```

OlegIlyenko:
@benhutchison it was already fixed in some other PR a while ago.

Issue #1430, opened 2014-07-01
sbt project: `set every version` does not replace anything


dansanduleac:
The sbt project itself sets a build-wide version in `project/Sbt.scala`.
If I `set version in ThisBuild := "test"`, that replaces it. Good.

However, if I `set every version := "test"`, that overrides nothing:

```
> set every version := "test"
[info] Defining */*:version
[info] The new value will be used by no settings or tasks.
[info] Reapplying settings...
[info] Set current project to root (in build file:/D:/work/ext/sbt/)
```

Is this expected behaviour?

jsuereth:
no, I think we just ran into this again.

cunei:
Unless I messed up the bisection, "set every" started working erratically after commit ~~a3d89dc8656d7ee5fb2656dae79b5154ebe9a5e2~~ 322f6de6551665cade7d56b532348ea5dc3d54db

cunei:
The definition of setAll(), which implements 'set every' uses `._1s`, which just grabs the keySet from the relation's forwardMap. Hence, having a mapping from something to an empty set, or not having the mapping, is /not/ equivalent when using `_1s`. I am issuing a pull request to revert the optimization.

dwijnand:
With #1591 merged (and released as a part of 0.13.7) this issue is resolved and can therefore be closed.

PR #1591, opened 2014-09-11
Fix for #1430

cunei:
In Relation, `_1s` is defined as "Returns the set of all `_1`s such that `(_1, _2)` is in this relation.". Which is false: it is implemented as fwd.keySet, but it should be fwd.filterNot(_._2.isEmpty).keySet. Except setAll() depends on the incorrect semantics, and the apparently innocuous optimization in 322f6de broke it as a result. This pull request reverts the change, pending an implementation of setAll which does not depend on `_1s` (after which `_1s` can probably be fixed). See #1430.

### DIFF
--- a/main/src/main/scala/sbt/SettingCompletions.scala
+++ b/main/src/main/scala/sbt/SettingCompletions.scala
@@ -27,7 +27,7 @@ private[sbt] object SettingCompletions {
     {
       import extracted._
       val r = relation(extracted.structure, true)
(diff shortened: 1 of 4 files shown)


jsuereth:
Hmm, look like this breaks the logic unit tests, so it appears we have something relying on the NEW behavior too.   I'll investigate.

jsuereth:
LGTM, after travis is green we can merge!

Issue #1589, opened 2014-09-10
Fix #485 to respect actual Maven repo location


briantopping:
https://github.com/sbt/sbt/commit/2f13b7a8c726b51f5765830a5ee810c9906760bd creates the "publish-m2" command, but if the repo is not at the default location, everything breaks and there's no configuration to change it.

I put a note on the offending line in that commit.

jsuereth:
@topping You're very correct.  PR welcome if you have a chance.

briantopping:
Thanks Josh, getting this in could solve a lot of problems locally. Will report back.

briantopping:
Resolving this issue with PR merged to master.

PR #1600, opened 2014-09-15
PR for #1589

briantopping:
Please review this PR for #1589.  As I have not been able to run SBT from my local build, I have not been able to test this PR. As mentioned in the notes for #1589, using the command there after "sbt package" does not run with the code changes that are apparently part of the build.

Happy to test this if someone can point me to some RTFM!


typesafe-tools:

ivy/src/main/scala/sbt/Resolver.scala
```diff
@@ -300,8 +300,19 @@ object Resolver {
   def localBasePattern = "[organisation]/[module]/" + PluginPattern + "[revision]/[type]s/[artifact](-[classifier]).[ext]"
   def defaultRetrievePattern = "[type]s/[organisation]/[module]/" + PluginPattern + "[artifact](-[revision])(-[classifier]).[ext]"
   final val PluginPattern = "(scala_[scalaVersion]/)(sbt_[sbtVersion]/)"
-
-  private[this] def mavenLocalDir = new File(Path.userHome, ".m2/repository/")
+  private[this] def mavenLocalDir: File =
```
jsuereth: Minor style comment:  A helper method cleans this up a bit.

``` scala
private[this] def mavenLocalDir: File = {
  def loadHomeFromSettings(f: File): Option[File] =
       (XML.loadFile(f) \ "settings" \ "localRepository")  match {
         case scala.xml.Text(loc) => Some(new File(loc))
         case _ => None
       }
  loadHomeFromSettings(new File(Path.userHome, ".m2/settings.xml")) orElse
  loadHomeFromSettings(new File(Path.fileProperty("M2_HOME"), "conf/settings.xml")) getOrElse
  new File(Path.userHome, ".m2/repository")
}
```

jsuereth:
IF this passes the test suite, I'm ok merging.

briantopping:
Ok, now that I understand how to test (thanks @eed3si9n!), I find that it doesn't work as it should. Let me get it fixed up first.

jsuereth:
Looking good!  tests pass, I'm about to click the merge button, but one comment first :)

briantopping:
Oh good idea! I'll add that thanks!!

eed3si9n:
Could you add release note entry in `notes/0.13.7`?

briantopping:
Hi guys, I am working on this but found some problem with my XML selection logic that I thought was working.  Hope to have a final commit tomorrow!

jsuereth:
Looking really good!  I reset the travis test that failed, because I think it was a transient.

briantopping:
Thanks Josh! I was wondering why the test failed for me and not others.

That's a good plan that it should warn on bad XML in a Maven settings file instead of fail since Maven is "non-critical" to SBT.  I'll get that worked up today! It's been taking a while because I have been trying to get my Windoze workstation going, but it's loaded with profiles and security, so it's been very slow going.

briantopping:
Hi Josh, as I get in to look at this (on my Mac, lol), I've got the exception for the bad XML but Resolver has no reference to a Logger.  I assume there's dozens of hacks and one way it's usually done.  Any suggestion?

**EDIT:** The more I look through the code, the less convinced I am that it is going to be possible to log a warning from a Resolver without a cascade of interface changes. The Resolver singleton is the only two clients of the mavenLocalDir method that was changed, but the Resolver does not have a Logger, and because it's a singleton, there's no clear responsibility for initialization with a Logger.  It also seems like the owner of Loggers are TaskStreams, but it's unclear to me how to obtain the current TaskStream.  As a last resort, I set a breakpoint at the instantiation of the singleton and dug around in the call chain to see if I could find anything resembling a Logger or a TaskStream and could find neither.

WDYT?

jsuereth:
@topping There _is_ a default logger you can use in an escape hatch, and also System.err.

It's an interesting conundrum here, since it's in a static method, and we have binary compatibility restrictions.   I guess I'm ok with either of these options:
- Logging the warning on System.err
- Not logging a warning.

The escape hatch logger is here: https://github.com/sbt/sbt/blob/0.13/main/src/main/scala/sbt/Main.scala#L57-L58, but we've recently been thinking about an overhaul to sbt's logging infrastructure due to issues.

briantopping:
Thanks Josh, I'll code to the escape hatch and if it needs to be recoded later, it's only one extra call.

jsuereth:
@topping not sure you can use the escape hatch, because it's in a downstream library.... you may be stuck with System.err....

ivy/src/main/scala/sbt/Resolver.scala
```diff
@@ -300,8 +301,23 @@ object Resolver {
...
+      try {
+        val file = XML.loadFile(f())
+        (file \ "localRepository").text match {
+          case ""    => None
+          case e @ _ => Some(new File(e))
+        }
+      } catch {
+        // Occurs inside File constructor when property or environment variable does not exist
+        case _: NullPointerException => None
+        case e: SAXParseException    => System.err.println(s"WARNING: Problem parsing ${f().getAbsolutePath}, ${e.getMessage}"); None
```
jsuereth, line 315: Awesome!

jsuereth:
LGTM

briantopping:
Awesome, thanks!

PR #1601, opened 2014-09-16
Expose Ivy's circular dependency check via UpdateOptions

eed3si9n:
See http://ant.apache.org/ivy/history/2.3.0/concept.html#circular

By default circular dependencies are warned, but they do not halt the dependency resolution. Using the following setting, circular dependencies can be treated as an error.

``` scala
updateOptions := updateOptions.value.withCircularDependencyLevel(CircularDependencyLevel.Error)
```

### DIFF
--- a/ivy/src/main/scala/sbt/CircularDependencyLevel.scala
+++ b/ivy/src/main/scala/sbt/CircularDependencyLevel.scala
@@ -1,0 +1,27 @@
+package sbt
+
+import org.apache.ivy.plugins.circular.{ CircularDependencyStrategy, WarnCircularDependencyStrategy, IgnoreCircularDependencyStrategy, ErrorCircularDependencyStrategy }
+
+/**
+ * Wrapper around circular dependency strategy.
+ */
+sealed trait CircularDependencyLevel {
+  private[sbt] def ivyStrategy: CircularDependencyStrategy
+  private[sbt] def name: String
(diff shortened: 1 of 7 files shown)


ivy/src/main/scala/sbt/CircularDependencyLevel.scala
```diff
@@ -0,0 +1,27 @@
+package sbt
+
+import org.apache.ivy.plugins.circular.{ CircularDependencyStrategy, WarnCircularDependencyStrategy, IgnoreCircularDependencyStrategy, ErrorCircularDependencyStrategy }
+
+/**
+ * Wrapper around circular dependency strategy.
+ */
+trait CircularDependencyLevel {
+  private[sbt] def ivyStrategy: CircularDependencyStrategy
```
jsuereth, line 9: Do you know the ramifications of a publicly extendable class where bits are marked `private[sbt]`?  Would extenders of this trait need to be in the sbt package?

jsuereth:
LGTM

ivy/src/main/scala/sbt/CircularDependencyLevel.scala
eed3si9n, line 9: I guess I should seal that sucker.

eed3si9n:
sealed the `CircularDependencyLevel` trait.

PR #1602, opened 2014-09-16
Bump to a version of ivy that will not swap in httpclient URL handler.

jsuereth:
When ivy finds the httpclient on the classpath, it will silently change
its url handler, and the httpclient in ivy 2.3.x has a few known
issues.

Review by @eed3si9n


eed3si9n:
Notes plz.

jsuereth:
Notes added.

Issue #1617, opened 2014-09-24
NPE in ClassToAPI


gkossakowski:
As [reported](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/scala-ide-user/3-vDAKsqu-M/_FHp8G1UWHMJ) by Scala IDE user, we have a NPE in ClassToAPI:

```
scala.MatchError: null
    at sbt.ClassToAPI$.reference(ClassToAPI.scala:255)
    at sbt.ClassToAPI$.defLike(ClassToAPI.scala:154)
    at sbt.ClassToAPI$.methodToDef(ClassToAPI.scala:143)
    at sbt.ClassToAPI$$anonfun$7.apply(ClassToAPI.scala:73)
    at sbt.ClassToAPI$$anonfun$7.apply(ClassToAPI.scala:73)
    at sbt.ClassToAPI$$anonfun$mergeMap$1.apply(ClassToAPI.scala:175)
    at sbt.ClassToAPI$$anonfun$mergeMap$1.apply(ClassToAPI.scala:175)
```

@dragos's hypothesis is that we are getting null from Java Reflection.

gkossakowski:
I believe this got closed by #1619.

PR #1619, opened 2014-09-24
Use Option.apply instead of Some.apply in ClassToAPI

gkossakowski:
When dealing with Java APIs (Java reflection in case of ClassToAPI), one
should always go through Option.apply that performs null check.

This is supposed to fix NPE reported in #1617. We don't have a reproduction
so this is just an educated guess.


gkossakowski:
Review by @dragos.

typesafe-tools:

dragos:
LGTM! Thanks, @gkossakowski!

PR #1621, opened 2014-09-25
propagate pattern descriptorOptional and skipConsistencyCheck

tmandke:
propagate pattern descriptorOptional and skipConsistencyCheck when `Resolver.url(name, url)(patterns)` is called.


typesafe-tools:

jsuereth:
LGTM!  Thanks much for the fix.

PR #1626, opened 2014-09-26
Cut valid text corrected

ajozwik:
Solution for https://github.com/sbt/sbt/commit/30b9cfc843077267fd3f0e0fa38d7c234c2d791d#sbt-src-sbt-test-project-session-save-test-P13. Problem was in `cutExpression` function. Now scala parser is used to split first line in sbt expression.

### DIFF
--- a/main/src/main/scala/sbt/SessionSettingsNoBlankies.scala
+++ b/main/src/main/scala/sbt/SessionSettingsNoBlankies.scala
@@ -1,13 +1,12 @@
 package sbt
 
-import java.io.File
-
 import scala.collection.immutable.SortedMap
 import scala.reflect.runtime.universe._
 
 object SessionSettingsNoBlankies {
 
-  private val FAKE_FILE = new File("fake")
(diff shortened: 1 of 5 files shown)


typesafe-tools:

jsuereth:
Cool!!!  Lgtm

jsuereth:
@ajozwik Did you see my refactoring commits for:
1. Keeping binary compatibility
2. Moving the new parser into an internals package (where in sbt 1.0 we want to keep a lot of internal details, so e.g. adding argumetns to parser methods is ok since it won't be publicly exposed).

I'm going to merge this is and I'll rebase my branch, but I'd like you to take a look at it.  I want to migrate the SessionSettingsNoBlankeis out of the sbt package and into internals, additionally, I'd also like to give that guy a newer/better name.

ADDITIONALLY, there's a pre-existing issue in `session save` around how it picks the `.sbt` file and the ordering of loading `.sbt` files which we may "correct" or at least "make consistent" as we merge this.

PR #1628, opened 2014-09-29
Corrections for xml handling and session save

ajozwik:
See comment https://github.com/sbt/sbt/commit/e8218faa7fb777ab99e7615fa0fecd1e6580264b#commitcomment-7954217

Still can be problem with white spaces in tests.
I received:

```
Caused by: java.lang.RuntimeException: Test script error: File contents are different:
k1 := {error("k1")}

k2 <<= k1 map identity
And:
k1 := {error("k1")}

k2 <<= k1 map identity

    at scala.sys.package$.error(package.scala:27)
```

But I have no idea - what is wrong? End of line?

Issue #1606, opened 2014-09-17
Natural whitespace handling for SBT configuration parser AKA 'no more blankies'


ajozwik:
Open list to implement/check:
- [x] Add tests around the new no-blank-lines files and session-save
- [ ] Possibly expand the notion of "position" we have in Setting[_] to include start/stop so we can more accurately alter a build.sbt file without changing it.
- [x] Investigate @dragos [comments about tokenizer/parser](https://github.com/sbt/sbt/pull/1574#discussion-diff-17551216).
- [ ] Current SBT parser cuts line by line. If we have:

``` scala
val a = <a/><b/>

val r = 4
```

SBT parser split it to: `val a = <a/><b/>` and `val r = 4`. Scala parser will compile one by one separate.

``` scala
scala> val a = <a/><b/>
a: scala.xml.NodeBuffer = ArrayBuffer(<a/>, <b/>)
```

But this code will not compile.

``` scala
scala> :pa
// Entering paste mode (ctrl-D to finish)

val a = <a/><b/>

val r = 4

// Exiting paste mode, now interpreting.

<console>:3: error: ';' expected but 'val' found.
val r = 4
^
```

For backward compatibility we have to find, where is xml code - and at least add ; to the end (now I will add brackets ()). The code to add bracket can be rewritten.
- [x] Write acceptance tests.
- [x] I think for all sbt part - how to split code to atomic parts - it is easier to test. 
  In #noblankies we extracted `EvaluateConfigurations.splitExpressions` and  `SessionSettingsNoBlankies.oldLinesToNew` for `session save`. Both methods can be tested by unit tests - without running whole sbt.
- [x] Scala parser should be corrected. See [Scala parser cuts last bracket](http://stackoverflow.com/q/25547149/651140)

jsuereth:
@ajozwik Let me know if you have time for a skype sometime this week to discuss the TODOs on here and what each of us can to do make sure this gets merged soon.  I'd like to have a 0.13.7-M3 for this as soon as we can to flush out more issues.

ajozwik:
@jsuereth
Point 4 and 7 are related to scala parser. They are out of scope. We have work - around.
Only point 2: `Possibly expand the notion of "position" we have in Setting[_] to include start/stop so we can more accurately alter a build.sbt file without changing it` is still open.

eed3si9n:
For easier code review, I rebased wip/no-blankies on top of 0.13, and made no-blankies-rebased.
[...]

PR #1630, opened 2014-09-29
Merge  #1628 and build again

ajozwik:
Rebase wip/no-blankies to current branch


typesafe-tools:

main/src/main/scala/sbt/SessionSettingsNoBlankies.scala
```diff
@@ -1,10 +1,7 @@
 package sbt
 
-import java.io.File
-
-import sbt.internals.parser.{ XmlContent, SplitExpressionsNoBlankies }
-import scala.collection.immutable.SortedMap
 import scala.reflect.runtime.universe._
+import sbt.internals.parser.{ XmlContent, SplitExpressionsNoBlankies }
 
 object SessionSettingsNoBlankies {
```
jsuereth: I am thinking we need to do some cleanup here....
1. We have .sbt refactoring code in `SessionSettings` that feels like it's somewhat unrelated to that class _specifically_.  We also have this new `SessionSettingsNoBlankies` with specific refactoring utilitites.
2. Most of the `.sbt` Parsing/Compiling/Refactoring code should get moved into an internals package (if it's new) so that we can limit what we expose (in .sbt 1.0, a lot of these "public" APIs will become private, so it's not so painful to modify behavior when making improvements).

That said, I think I'd like to rename this file to "SbtRefactorings" or "ConfgurationRefactorings" and slide more of the code out of the SessiongSettings class and into here (albeit without breaking binary compatibillity.

@ajozwik Do you think the method names would be altered if we viewed this as a generic SbtRefactorings utility object.

ajozwik: I will move the code to internals package. But with binary compatibile - the problem is with empty new line. Old session save replace whole line and add new line at end. New replace statement with new statement - which have only one line without new line.

ajozwik:
Related to #1606

main/src/main/scala/sbt/internals/parser/SbtRefactorings.scala
```diff
@@ -0,0 +1,76 @@
...
+import scala.reflect.runtime.universe._
+
+private[sbt] object SbtRefactorings {
+
+  import sbt.internals.parser.SplitExpressionsNoBlankies.{ END_OF_LINE, FAKE_FILE }
+
+  val EMPTY_STRING = ""
+  val REVERSE_ORDERING_INT = Ordering[Int].reverse
+
+  def applyStatements(lines: List[String], commands: List[List[String]]): List[String] = {
```
jsuereth: Would you mind adding some Scaladoc comments on what the intended input/output for this is?

jsuereth: Something like:

```
/**
 * Refactors an `.sbt` file so that the new settings are used instead of any existing settings.
 * @param lines  The lines of an sbt file as a List[String] where each string is one line
 * @param commands A List of settings (space separate) that should be inserted into the current file.
 *                                  If the settings replaces a value, it will replace the orginal line in the .sbt file.
 * @return a list of new lines which represent the contents of the refactored .sbt file.
 */
```

As a potential.   By writing the docs, this will highlight where our APIs in sbt are a bit 'odd'.    For example, I think I may go clean up this interface at some point to just take in:
  `(SbtConfigFile, Seq[SessionSetting]) => SbtConfigFile`

Or some other set of types which have more meanings than `List[String]` and `List[List[String]]`.

For now, this API is fine, but I'd like to clean up (either yourself or my team) before we get into sbt master.

cc @eed3si9n

main/src/test/scala/sbt/internals/parser/SessionSettingsSpec.scala
```diff
@@ -35,11 +32,15 @@ abstract class AbstractSessionSettingsSpec(folder: String, printDetails: Boolean
       file =>
         val originalLines = Source.fromFile(file).getLines().toList
         foreach(expectedResultAndMap(file)) {
-          case (expectedResultList, map) =>
-            val resultList = SessionSettingsNoBlankies.oldLinesToNew(originalLines, map)
+          case (expectedResultList, commands) =>
+            val resultList = SbtRefactorings.applyStatements(originalLines, commands.map(List(_)))
```
jsuereth: Yeah, code like this makes me think the exposed refactoring API may be off... (`List[List[String]]` instead of just `List[String]` seems an artifact of sbt's implementation)

jsuereth:
Caveat scaladoc/API comments, this looks good!  

I can either merge this PR and do the refactoring on my end issuing a PR for you to review, or you can update the PR on your end.   Up to you!  

Again, thanks much for all the hard work.  This is rapidly approaching something we can PR against 0.13

main/src/main/scala/sbt/internals/parser/SbtRefactorings.scala
ajozwik: You right. Now we do not need a content of file as List[String]. Inside `splitExpression` the new content is created - and again split.
I will correct the code tonight - tomorrow fill free to update the code.

ajozwik:
@jsuereth I finished - so feel free to adapt the API.

jsuereth:
@ajozwik AWESOME, thanks!  this looks good to merge once TravisCI reports back.

ajozwik:
It is some magic work outside the applySessionSettings:

``` scala
201    val adjusted = if (!newSettings.isEmpty && needsTrailingBlank(exist)) exist :+ "" else exist
202   val lines = adjusted ++ newSettings.flatMap(x => x._2 :+ "")
```

But then the newSettings has to be new parameter to `applySessionSettings` function and unit tests begin to be more complicated (you have to create `Setting` object).

PR #1631, opened 2014-09-29
Cached resolution (minigraph caching)

eed3si9n:
[Consolidated resolution](https://github.com/sbt/sbt/pull/1454) proved that avoiding work by caching is going to be faster than trying to do all the work efficiently.

cached-resolution
(consolidated resolution)

One drawback of this approach is that it requires most subprojects to have identical dependencies graph to take advantage of the speedup.

Instead of creating an artificial graph for unique subprojects, cached resolution creates an artificial dependency graph for each direct dependency of a subproject. I'm calling them a minigraph. Minigraphs can expand to dozens of modules using stock resolution algorithm, and then saved to individual json files. Finally, these minigraphs are stitched together using the conflict resolver built into Ivy.

cached_resolution
(cached resolution)

To make sure that dynamic revisions (such as -SNAPSHOT) are handled correctly, sbt will keep an internal logical clock, which will be incremented when the user issues a command. Also these minigraphs are shared across different builds.


ivy/src/main/scala/sbt/IvyActions.scala
```diff
@@ -141,7 +142,7 @@ object IvyActions {
    */
   @deprecated("Use updateEither instead.", "0.13.6")
```
jsuereth: Deprecate as removed if you're making updateEither private sbt, something like "This is no longer public"

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1262,25 +1264,28 @@ object Classpaths {
     val transform: UpdateReport => UpdateReport = r => substituteScalaFiles(scalaOrganization.value, r)(subScalaJars)
     val uwConfig = (unresolvedWarningConfiguration in update).value
     val show = Reference.display(thisProjectRef.value)
+    val st = state.value
+    val logicalClock = LogicalClock(Hash.toHex(intToByteArray(st.hashCode)))
```
jsuereth, line 1268: interesting notion of "clock"

jsuereth, line 1268: I need to think through whether the invaliation provided by this matches what we want....   can you elucidate your thoughts here?

eed3si9n, line 1268: The notion of logical clock comes from [Lamport's time](http://en.wikipedia.org/wiki/Lamport_timestamps). Since the minigraphs are shared across different builds, I need to coordinate them somehow. The clock acts as a point in time that I can freeze the dynamic dependencies during `update` on multiple-project build. State's hash would remain constant during command execution, so this avoids re-resolving dynamic dependencies when `update` is delegated. The next time the user types anything into the shell, the local clock will tick forward. If that's too trigger happy, we can later slow it down to synchronize with a wall clock (e.g. 15min) when State's hash is different.

ivy/src/main/scala/sbt/ivyint/ConsolidatedResolveEngine.scala
```diff
@@ -1,120 +0,0 @@
...
-import core.{ IvyPatternHelper, LogOptions }
-import org.apache.ivy.util.Message
-
-private[sbt] object ConsolidatedResolveCache {
-  def createID(organization: String, name: String, revision: String) =
-    ModuleRevisionId.newInstance(organization, name, revision)
-  def sbtOrgTemp = "org.scala-sbt.temp"
-}
-
-private[sbt] class ConsolidatedResolveCache() {
```
jsuereth, line 20: Very glad this was private sbt

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -0,0 +1,411 @@
...
+import core.module.descriptor.{ DefaultModuleDescriptor, ModuleDescriptor, DependencyDescriptor, Configuration => IvyConfiguration }
+import core.{ IvyPatternHelper, LogOptions }
+import org.apache.ivy.util.Message
+import org.apache.ivy.plugins.latest.{ ArtifactInfo => IvyArtifactInfo }
+import org.json4s._
+
+private[sbt] object CachedResolutionResolveCache {
+  def createID(organization: String, name: String, revision: String) =
+    ModuleRevisionId.newInstance(organization, name, revision)
+  def sbtOrgTemp = "org.scala-sbt.temp"
```
jsuereth, line 22: So, BASICALLY, we can NEVER EVER EVER use this org?

eed3si9n, line 22: You _could_ as long as the name doesn't collide with `"temp-resolve-" + sha1`.

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -0,0 +1,411 @@
...
+}
+
+class CachedResolutionResolveCache() {
+  import CachedResolutionResolveCache._
+  val updateReportCache: concurrent.Map[ModuleRevisionId, Either[ResolveException, UpdateReport]] = concurrent.TrieMap()
```
jsuereth, line 27: is this multi-threaded?

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -0,0 +1,411 @@
...
+  val maxConflictCacheSize: Int = 10000
+
+  def clean(md0: ModuleDescriptor, prOpt: Option[ProjectResolver]): Unit = {
+    val mrid0 = md0.getModuleRevisionId
+    val mds =
+      if (mrid0.getOrganisation == sbtOrgTemp) Vector(md0)
+      else buildArtificialModuleDescriptors(md0, prOpt) map { _._1 }
+    mds foreach { md =>
+      updateReportCache.remove(md.getModuleRevisionId)
+      // todo: backport this
```
jsuereth: What does this mean?  backport where?

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -0,0 +1,411 @@
...
+      directDependencyCache.remove(md.getModuleRevisionId)
+    }
+  }
+  def directDependencies(md0: ModuleDescriptor): Vector[DependencyDescriptor] =
+    directDependencyCache.getOrElseUpdate(md0.getModuleRevisionId, md0.getDependencies.toVector)
```
jsuereth, line 45: It's amazing how my initial impression was "why the cache/getOrElse", then I forgot `getDependencies` is super side-effecting/slow.

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -0,0 +1,411 @@
...
+            case Some(internal) => directDependencies(internal) flatMap expandInternalDeps
+            case _              => Vector(dep)
+          }
+        } getOrElse Vector(dep)
+      val expanded = directDependencies(md0) flatMap expandInternalDeps
+      val rootModuleConfigs = md0.getConfigurations.toVector
+      expanded map { buildArtificialModuleDescriptor(_, rootModuleConfigs, prOpt) }
+    }
+  def isDependencyChanging(dd: DependencyDescriptor): Boolean =
+    dd.isChanging || IvySbt.isChanging(dd.getDependencyRevisionId)
```
jsuereth: hmm, do we need to unify this concept in ivy in a different mechanism?  Hopefully we're not always dragging this around.

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -0,0 +1,411 @@
...
+  def buildArtificialModuleDescriptor(dd: DependencyDescriptor, rootModuleConfigs: Vector[IvyConfiguration], prOpt: Option[ProjectResolver]): (DefaultModuleDescriptor, Boolean) =
+    {
+      val mrid = dd.getDependencyRevisionId
+      val confMap = (dd.getModuleConfigurations map { conf =>
+        conf + "->(" + dd.getDependencyConfigurations(conf).mkString(",") + ")"
```
jsuereth, line 64: hmm, what's this doing?

jsuereth, line 64: Nevermind, got it.  You're caching the configuration dependency aspect of the current dependency, right?

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -0,0 +1,411 @@
...
+      val depsString = mrid.toString + ";" + confMap.mkString(";")
+      val sha1 = Hash.toHex(Hash(depsString))
+      val md1 = new DefaultModuleDescriptor(createID(sbtOrgTemp, "temp-resolve-" + sha1, "1.0"), "release", null, false) with ArtificialModuleDescriptor {
+        def targetModuleRevisionId: ModuleRevisionId = mrid
+      }
+      for {
+        conf <- rootModuleConfigs
+      } yield md1.addConfiguration(conf)
+      md1.addDependency(dd)
+      (md1, isDependencyChanging(dd))
```
jsuereth: I'm having trouble seeing how isDependencyChanging is used here / how -SNAPSHOTS gets evicted from the cache.

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -0,0 +1,411 @@
...
+        case x: ArtificialModuleDescriptor =>
+          val tmrid = x.targetModuleRevisionId
+          (tmrid.getOrganisation, tmrid.getName, tmrid.getRevision + "_" + mrid.getName)
+        case _ =>
+          (mrid.getOrganisation, mrid.getName, mrid.getRevision)
+      }
+      val staticGraphDirectory = miniGraphPath / "static"
+      val dynamicGraphDirectory = miniGraphPath / "dynamic"
+      val staticGraphPath = staticGraphDirectory / pathOrg / pathName / pathRevision / "graphs" / "graph.json"
+      val dynamicGraphPath = dynamicGraphDirectory / logicalClock.toString / pathOrg / pathName / pathRevision / "graphs" / "graph.json"
```
jsuereth, line 91: So you save/load in a different location if it's changing (i.e. -SNAPSHOT)?

jsuereth, line 91: I think this means we may need to do some sort of hackery, like mark all "inter-project" dependencies as changing....

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
eed3si9n, line 27: concurrent.TrieMap should be right?

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -0,0 +1,411 @@
...
+              val (surviving, evicted, mgr) = f
+              if (conflictCache.size > maxConflictCacheSize) {
+                conflictCache.remove(conflictCache.head._1)
+              }
+              conflictCache((cf0, cf1)) = (surviving map { _.module }, evicted map { _.module }, mgr)
+              (surviving, evicted)
+          }
+      }
+    }
+  class FileSerializer extends CustomSerializer[File](format => (
```
jsuereth: I feel like these shoudl probably be in a differnet file.

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -0,0 +1,411 @@
...
+          Nil)
+    }
+  ))
+}
+
+private[sbt] trait ArtificialModuleDescriptor { self: DefaultModuleDescriptor =>
+  def targetModuleRevisionId: ModuleRevisionId
+}
+
+private[sbt] case class UpdateReportLite(configurations: Seq[ConfigurationReportLite])
```
jsuereth: Same here, maybe but this in a file.

jsuereth:
Wow.  Generally looks good.  A few minor comments (inter-project being the major concern), and then add project notes and good to merge!

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
eed3si9n: I was thinking to backport to consolidated resolution since it could explain one of the behaviors reported on issue or ML. It probably won't be needed if we're merging this.

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
eed3si9n: I could refactor that into `IvySbt`.

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
eed3si9n, line 64: Yea. This is the hacky part of this code. I'm calculating SHA1 of a "dependency" by making a string representation by listing out configurations that each rootModuleConf depends on.

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
eed3si9n, line 91: Yea, so if either you're -SNAPSHOT or you include a SNAPSHOT, you're on a different path.

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
eed3si9n: Eventually `changing` influences line 118 when I write out the minigraph into json:

``` scala
              writeUpdateReport(ur,
                if (changing) dynamicGraphPath
                else staticGraphPath)
```

eed3si9n:
Added a few more commits to address the points raised in the review. Also added notes trumpeting the upsides.

jsuereth:
Ok, LGTM

omnilinguist:
@eed3si9n @jsuereth How significant of a performance improvement, generally/roughly speaking, is this expected to have on "update" tasks, and is this at all related to/dependent on sbt/ivy#4? Is this available starting from sbt 0.13.7?

cc: @wchen9911

jsuereth:
Yeah, so YMMV, BUT if you're using few -SNAPSHOTs, we've seen second-resolves improve dramatically (>50%).   It is related to sbt/ivy#4 _and_ it'll be available sbt 0.13.7 onward.   You should try it out in sbt 0.13.7-RC5 and see what you find.  @eed3si9n is workign on the documentation for it now, if you'd like to read up: https://github.com/sbt/website/pull/72

eed3si9n:
@omnilinguist This feature came out of LinkedIn Tools team consulting to improve resolution performance. For projects with 10+ subprojects we saw 10x improvements running `update` task.
One project in particular had heterogenous dependencies, and it improved from 269s to 25±5s with cached resolution.

omnilinguist:
@eed3si9n @jsuereth Yes, that is who I am representing. :) Not sure how soon we will be able to test with 0.13.7\* though, unless it is compatible with the Play 2.2.x branch (we're at approximately 2.2.4 and sbt 0.13.1) - do you know anything about this?

eed3si9n:
@omnilinguist sbt 0.13.7 maintains binary compatibility with sbt 0.13, so it should work with Play 2.2.x.

PR #1633, opened 2014-09-30
Update to current 0.13.7-SNAPSHOT

ajozwik:
$HOME/.ivy2/local/org.scala-sbt/sbt-launch/ instead of temporary target folder


typesafe-tools:

jsuereth:
Thanks for the update!  LGTM

PR #1637, opened 2014-10-01
Documentation and renaming of "blankies" into somethign a bit easier to ...

jsuereth:
...find.
- Rename SplitExpression\* to `SbtParser` denoting that is parses .sbt files
- Adds a few todos
- Document APIs for internal usage.

Review by @ajozwik  cc> @eed3si9n

### DIFF
--- a/main/src/main/scala/sbt/EvaluateConfigurations.scala
+++ b/main/src/main/scala/sbt/EvaluateConfigurations.scala
@@ -8,7 +8,7 @@ import compiler.{ Eval, EvalImports }
 import complete.DefaultParsers.validID
 import Def.{ ScopedKey, Setting }
 import Scope.GlobalScope
-import sbt.internals.parser.SplitExpressionsNoBlankies
+import sbt.internals.parser.SbtParser
 
 import scala.annotation.tailrec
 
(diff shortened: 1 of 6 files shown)


main/src/main/scala/sbt/EvaluateConfigurations.scala
```diff
@@ -218,7 +218,9 @@ object EvaluateConfigurations {
    */
   private[sbt] def splitExpressions(file: File, lines: Seq[String]): (Seq[(String, Int)], Seq[(String, LineRange)]) =
```
ajozwik, line 219: Why we need lines? File is enough - we can read in `SplitExpressionsNoBlankies`. But I think it is for future changes.

ajozwik:
For current changes - no comments. For future - we want to have sbt file - the lines are deprecated.

main/src/main/scala/sbt/EvaluateConfigurations.scala
jsuereth, line 219: I agree, sort of.   It's useful to have a method which just takes the raw streams or "virtual file" so we can do in-memory settings (like `set` command in sbt), where multiple settings could be in a string.   Right now we do not do that so you are correct.

I agree we should work on simplifying API now that the feature is pretty stable.   I can add a TODO here about removing the need for lines and reading in splitExpressions.  For now, as it's a private API, it's ok to refactor this at any time.  I"m more concerned with the exposed public APIs _and_ ensuring this area is documented for the next user group that wants to work on it :)

ajozwik, line 219: If I find some time - I can fight with it. But I not promise.
Thank you for your help.

Issue #1640, opened 2014-10-03
Name hashing heap explosion with shapeless master


milessabin:
Building shapeless master with SBT 0.13.6 results in an OutOfMemoryError with default 2G heap (configured via .jvmopts) and continued to fail with the heap limit increased to 3 and 4G.

Reverting to 0.13.5 or disabling name hashing (with `incOptions := incOptions.value.withNameHashing(false)`) allows the build to complete successfully in 2G.

gkossakowski:
Thanks for the report!  This looks like duplicate of #1610 but with advantage of publicly available project where OOM happens. I'll have a quick look now.

gkossakowski:
The problem is with macro expansion handling. The `ExtractUsedNames.handleMacroExpansion` appears to go into an infinite loop and leaks memory at the same time.

gkossakowski:
//cc @Duhemm who worked on improved macro handling recently.

Duhemm:
My first guess is that the set containing the list of original trees (trees before expansion, `inspectedOriginalTrees`) introduced in #1544 ends up consuming all the memory... What would be a better solution to the problem of cyclic structures in macro originals ?

gkossakowski:
I think the solution in #1544 was fine but we might have missed some small detail. @Duhemm, can you reproduce the problem on your machine?

Duhemm:
Actually I've not been able to reproduce the OOM on milessabin/shapeless. Are there other instructions than `compile`-ing the project with sbt 0.13.6 to reproduce it ?

You're right, I've been looking a little too fast at your message (I'm helping the students with their software engineering class at the moment) and you said it was located in `ExtractUsedNames.handleMacroExpansion`.

Do you think the `filter` in https://github.com/sbt/sbt/blob/0.13/compile/interface/src/main/scala/xsbt/ExtractUsedNames.scala#L64 is responsible for this problem ?

gkossakowski:
No, it should work out of the box. Are you sure you have the latest (master) shapeless checked out? I used 5ec04acd101367cbf3aba0b1a08973192b0e1de8 commit from shapeless. Apply

``` patch
diff --git a/project/build.properties b/project/build.properties
index be6c454..64abd37 100644
--- a/project/build.properties
+++ b/project/build.properties
@@ -1 +1 @@
-sbt.version=0.13.5
+sbt.version=0.13.6
```

[...]

Issue #1610, opened 2014-09-22
0.13.6 - java.util.concurrent.ExecutionException: java.lang.OutOfMemoryError Regression


mdedetrich:
When upgrading to SBT 0.13.6, building our code makes SBT to run out of memory. Switching to SBT v0.13.5 immediately fixed the issue, so there is definitely a regression in this area.

This is private company code, so it may be hard to diagnose, but hopefully this stack trace can help

The code makes heavy use of implicits, so this may help in solving the regression (although I have no real idea if that is what caused the issue)

https://gist.github.com/mdedetrich/7dfa12e8c9e584b4d6a2

gkossakowski:
Hi!

You can keep sbt 0.13.6 but turn off name hashing incremental compilation algorithm with `incOptions := incOptions.value.withNameHashing(false)`. How much memory does sbt process have? Does increasing memory limit help?

Could you take a memory dump and analyze what occupies most of the memory?

mdedetrich:
The SBT process had the following parameters

```
SBT_OPTS="-XX:+UseConcMarkSweepGC -XX:+CMSClassUnloadingEnabled -Xms14g -Xmx14g"
```

This is on the latest and fully upgraded macbook pro. I tried different options (Xms/xmx used to be 8g) and it made no difference)

Regarding memory dump, do you have any instructions for this?

gkossakowski:
Yes, take a look at:

http://stackoverflow.com/questions/542979/using-heapdumponoutofmemoryerror-parameter-for-heap-dump-for-jboss

One you have memory dump file created you can browse it using tools like YourKit or VisualVM. It appears like there's a big memory leak so it should be fairly easy to identify what's going on just by inspecting the heap dump.

gkossakowski:
This has been fixed by #1642. See #1640.

PR #1642, opened 2014-10-03
Port fix for #1544 from Dependency to ExtractUsedNames

Duhemm:
The fix for sbt/sbt#1237 was unfortunately not completely correct,
and infinite loops could still occur during the extraction of used
names.

In sbt/sbt#1544, a fix that was robuster and easier to understand
was applied to `/compile/interface/src/main/scala/xsbt/Dependency.scala`
in a similar situation (cyclic chains of original trees in macro
expansions).

This commit ports this fix to `ExtractUsedNames.scala`.

Closes sbt/sbt#1640, sbt/sbt#1610.

### DIFF
--- a/compile/interface/src/main/scala/xsbt/ExtractUsedNames.scala
+++ b/compile/interface/src/main/scala/xsbt/ExtractUsedNames.scala
@@ -49,6 +49,16 @@ class ExtractUsedNames[GlobalType <: CallbackGlobal](val global: GlobalType) ext
 
   private def extractByTreeWalk(tree: Tree): Set[String] = {
     val namesBuffer = collection.mutable.ListBuffer.empty[String]
+
+    /*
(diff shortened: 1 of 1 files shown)


typesafe-tools:

jsuereth:
review by @gkossakowski

jsuereth:
also, having seen the discussion, the patch LGTM.  Would like to merge today/tomorrow as we want to issue a 0.13.7-M3 with the new "no blank lines" parser

gkossakowski:
LGTM.
It would be great to have a test case covering the issue, though. @Duhemm, did you manage to figure out the shape of macro definition and application that was causing the infinite loop?

Duhemm:
@gkossakowski I've not tried yet, I'll find some time this week to investigate.

PR #1643, opened 2014-10-03
.gitignore cleaned

ajozwik:
.gitignore restored according to @eed3si9n comment


typesafe-tools:

Issue #1639, opened 2014-10-03
force not respected during cached resolution


eed3si9n:
## steps

``` scala
lazy val check = taskKey[Unit]("Runs the check")

def commonSettings: Seq[Def.Setting[_]] =
  Seq(
    ivyPaths := new IvyPaths( (baseDirectory in ThisBuild).value, Some((target in LocalRootProject).value / "ivy-cache")),
    libraryDependencies := Seq(
      "org.springframework" % "spring-core" % "3.2.2.RELEASE" force() exclude("org.springframework", "spring-asm"),
      "org.springframework" % "spring-context" % "4.0.3.RELEASE" exclude("org.springframework", "spring-asm")
    ),
    scalaVersion := "2.10.4"
  )

def consolidatedResolutionSettings: Seq[Def.Setting[_]] =
  commonSettings ++ Seq(
    updateOptions := updateOptions.value.withConsolidatedResolution(true)
  )

lazy val a = project.
  settings(consolidatedResolutionSettings: _*)

lazy val b = project.
  settings(commonSettings: _*)

lazy val root = (project in file(".")).
  settings(
    organization in ThisBuild := "org.example",
    version in ThisBuild := "1.0",
    check := {
      val acp = (externalDependencyClasspath in Compile in a).value
      val bcp = (externalDependencyClasspath in Compile in b).value
      if (acp == bcp) ()
      else sys.error("Different classpaths are found:\n" +
        "consolidated: " + acp.toString + "\n" +
        "plain:        " + bcp.toString)
    }
  )
```

From shell run `check` task.
## problem

consolidated resolution is picking up `Attributed(/xxx/target/ivy-cache/cache/org.springframework/spring-context/jars/spring-context-4.0.3.RELEASE.jar)`:

```
[...]

PR #1644, opened 2014-10-05
[0.13.7] Fixes json serialization turning license into {}.

eed3si9n:
While trying to reproduce #1639 with 0.13.7's cached resolution, I ran into several issues:
- minigraph (json4s serialization) persists license `(Some("name of license"), None)` into `{}`, which is unserializable.
- cached resolution in scripted test bleeding out to global location.

The temporary workaround for the first issue is to put `Some("http://localhost")` in place for `None`.
For the second issue I introduced `dependencyCacheDirectory` key, which is set to 

``` scala
dependencyCacheDirectory := (baseDirectory in LocalRootProject).value / "dependency"
```

in scripted tests.

/review @jsuereth


ivy/src/main/scala/sbt/IvyRetrieve.scala
```diff
@@ -128,7 +128,8 @@ object IvyRetrieve {
     val licenses: Seq[(String, Option[String])] = mdOpt match {
       case Some(md) => md.getLicenses.toArray.toVector collect {
         case lic: IvyLicense if Option(lic.getName).isDefined =>
-          (lic.getName, nonEmptyString(lic.getUrl))
+          val temporaryURL = "http://localhost"
+          (lic.getName, nonEmptyString(lic.getUrl) orElse { Some(temporaryURL) })
```
jsuereth, line 132: so license url is non-optional?

jsuereth:
LGTM - I think...

ivy/src/main/scala/sbt/IvyRetrieve.scala
eed3si9n, line 132: Normally it _is_ optional, but our json serialization is not smart about it (not gonna get into serialization debate here) as described in the description, so this is a total workaround.

PR #1645, opened 2014-10-06
Natural whitespace handling for SBT configuration parser AKA 'no more blankies'

### LINKED ISSUE #1606: Natural whitespace handling for SBT configuration parser AKA 'no more blankies'

eed3si9n:
This PR fixes #1606. We should get this in as is for a milestone and fix whatever that needs fixing later.


jsuereth:
+1   I'm ok with this as-is for M3/master.   Kudos for all the cleanups and contributions, this is a very exciting contributed PR.

eed3si9n:
We do need to start writing a note if we wanna release this out though. What do you think about the following?:

## whitespace handling improvements

Starting sbt 0.13.7, build.sbt will be parsed using a customized Scala parser. This eliminates the requirement to use blank line as the delimiter between each settings, and also allows blank lines to be inserted at arbitrary position within a block.

This feature was contributed by [Andrzej Jozwik @ajozwik](https://github.com/ajozwik) inspired by Typesafe's @gkossakowski organizing multiple meetups with Warszaw Scala on [how to patch sbt specifically with the focus on blank line issue](http://blog.japila.pl/2014/07/gkossakowski-on-warszawscala-about-how-to-patch-scalasbt/).

gkossakowski:
I can't find any mention of being able to fallback to the old `.sbt` parser. Is there a way to configure sbt to do that?

eed3si9n:
@gkossakowski +1. we did talk about that internally but haven't done it. let's get that in by RC.

ajozwik:
For fallback to old `.sbt` parser I think we must to set property on start for example `-Dold.sbt=true`. I do not know if you can put it to sbt file - it is parsed before setting is red.

PR #1647, opened 2014-10-07
Replace @WarszawScala by @WarsawScala

ajozwik:
Organization name changed to more readable. Old name does not match any organization.


typesafe-tools:

PR #1652, opened 2014-10-08
Add escape hatch to old parser in case we run into problems.

jsuereth:
Just make sure users have the means to still build in case there's any
issues we missed.  This code should be removed in sbt 1.0.

Missed this in the earlier wip / Merge.

Review by @eed3si9n  cc @gkossakowski


eed3si9n:
LGTM. Let's wait for Travis.

PR #1653, opened 2014-10-08
Fixes #1649. Exclusion rules and other cached resolution fixes

### LINKED ISSUE #1649: Cached resolution does not respect excludes
In cached resolution, minigraph keys do not consider excluded dependencies.  If one project excludes a transitive dependency, and another does not, it becomes non-deterministic which minigraph will be used under cached resolution.

Minimal repro:
- Clone https://github.com/rossabaker/sbt-cached-resolution-exclusion
- Run `sbt clean test`

Expectation:
- Both specs should pass: two projects should be able to independently prune their transitive dependencies.

Result:
- One will not.  For me it's ExcludeSpec, but I suspect this may be non-deterministic.

Real world motivation:
- hadoop-core brings in old servlet-apis under the `org.mortbay.jetty` organization.  This is not evicted by the Java Servlet API from `javax.servlet`, resulting in two conflicting versions of `javax.servlet` on the classpath.  If the user has compiled another Hadoop project that does not exclude the servlet-api, the cache is polluted for the project that does need to exclude it.

Extra credit:
- Works with `excludeAll`
- Works with `ivyXML`

eed3si9n:
Fixes #1649.

### DIFF
--- a/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
+++ b/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
@@ -112,11 +112,19 @@ private[sbt] class CachedResolutionResolveCache() {
       def loadMiniGraphFromFile: Option[Either[ResolveException, UpdateReport]] =
         (if (staticGraphPath.exists) Some(staticGraphPath)
         else if (dynamicGraphPath.exists) Some(dynamicGraphPath)
-        else None) map { path =>
-          log.debug(s"parsing ${path.getAbsolutePath.toString}")
-          val ur = JsonUtil.parseUpdateReport(md, path, cachedDescriptor, log)
-          updateReportCache(md.getModuleRevisionId) = Right(ur)
-          Right(ur)
+        else None) match {
+          case Some(path) =>
+            log.debug(s"parsing ${path.getAbsolutePath.toString}")
+            val ur = JsonUtil.parseUpdateReport(md, path, cachedDescriptor, log)
+            if (ur.allFiles forall { _.exists }) {
+              updateReportCache(md.getModuleRevisionId) = Right(ur)
+              Some(Right(ur))
+            } else {
+              log.debug(s"some files are missing from the cache, so invalidating the minigraph")
+              IO.delete(path)
+              None
+            }
+          case _ => None
         }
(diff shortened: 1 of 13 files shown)


lightbend-cla-validator:
Hi @eed3si9n,

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -59,11 +59,28 @@ private[sbt] class CachedResolutionResolveCache() {
     }
   def buildArtificialModuleDescriptor(dd: DependencyDescriptor, rootModuleConfigs: Vector[IvyConfiguration], prOpt: Option[ProjectResolver]): (DefaultModuleDescriptor, Boolean) =
     {
+      def excludeRuleString(rule: ExcludeRule): String =
+        s"""Exclude(${rule.getId},${rule.getConfigurations.mkString(",")},${rule.getMatcher})"""
+      def includeRuleString(rule: IncludeRule): String =
+        s"""Include(${rule.getId},${rule.getConfigurations.mkString(",")},${rule.getMatcher})"""
```
jsuereth, line 68: Hmm..... this feels less consistent....   Are the toStrings of rule.getMatcher pretty consistent in general?

jsuereth:
Generally LGTM

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
eed3si9n, line 68: Yes. Ivy, like many Java libraries, have nice `toString()` normally. :neutral_face:

eed3si9n:
Since this was not merged yet, I'm stacking another bug fix on top.

jsuereth:
LGTM, again

eed3si9n:
More fixes were added yesterday. Tests failing, but not sure if it's me or Travis.

eed3si9n:
@jsuereth so you were right about Travis error being legit. Good job, Travis/scripted!
`updateClassifiers` sets a flag called `missingOk`, which allows Ivy to explorer around to grab JARs, but not let the unresolved error propagate as error. I needed to port the behavior for cached resolution.

jsuereth:
LGTM

PR #1663, opened 2014-10-10
enable -deprecation for Scala 2.10

eed3si9n:
Enable `-deprecation` flag to catch old code being use when we migrate things.
In this commit I moved error to sys.error.


jsuereth:
lgtm

PR #1662, opened 2014-10-10
Disable pattern matching in val syntax in sbt 0.13.7.

jsuereth:
Fixes #1661 or at least works around it for now.

Review by @eed3si9n cc> @jaceklaskowski

PR #1665, opened 2014-10-12
Remove DefDef and check rhs

eed3si9n:
#1662
- lhs pattern matching occurs only for vals.
- def also breaks due to proc notation.
- instead of pretty printing, we can just check rhs.


eed3si9n:
@jsuereth I'm getting multiple Travis 10 min errors with `SCRIPTED_TEST="checkBuildScala211"`.

jsuereth:
Does this catch lazy vals?

jsuereth:
Merging without scala 2.11 build for now, we'll handle that on my pr.  I think we may have an issue with the 2.11 parser and the new parsrer unit tests.  Glad it's caught, but curious....

PR #1669, opened 2014-10-13
Fixes #1648 - warn about multiple or no main classes

### LINKED ISSUE #1648: selectPackageMain should always warn about ambiguous main class
In

https://github.com/sbt/sbt/blob/0.13/main/src/main/scala/sbt/Defaults.scala#L668

sbt guesses main class for a package, but do not warn if there is more than one.

sbt should use some strategy to warn about the fact, like two lines above for selectRunMain

kretes:
Fixes #1648 

Works, but doesn't look like the best way to handle that.

My PR would benefit from some hints on how to handle logging in sbt itself.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -264,9 +264,9 @@ object Defaults extends BuildCommon {
     definedSbtPlugins <<= discoverPlugins,
     discoveredSbtPlugins <<= discoverSbtPluginNames,
     inTask(run)(runnerTask :: Nil).head,
-    selectMainClass := mainClass.value orElse selectRunMain(discoveredMainClasses.value),
+    selectMainClass := pickMainClass(discoveredMainClasses.value) orElse askForMainClass(discoveredMainClasses.value),
     mainClass in run := (selectMainClass in run).value,
-    mainClass := selectPackageMain(discoveredMainClasses.value),
+    mainClass := pickMainClassOrWarn(discoveredMainClasses.value, streams.value.log),
     run <<= runTask(fullClasspath, mainClass in run, runner in run),
     runMain <<= runMainTask(fullClasspath, runner in run),
     copyResources <<= copyResourcesTask
@@ -662,10 +662,22 @@ object Defaults extends BuildCommon {
       new Package.Configuration(srcs, path, options)
     }
 
-  def selectRunMain(classes: Seq[String]): Option[String] =
+  @deprecated("use Defaults.askForMainClass", "0.13.7")
+  def selectRunMain(classes: Seq[String]): Option[String] = askForMainClass(classes)
+  @deprecated("use Defaults.pickMainClass", "0.13.7")
+  def selectPackageMain(classes: Seq[String]): Option[String] = pickMainClass(classes)
+  def askForMainClass(classes: Seq[String]): Option[String] =
     sbt.SelectMainClass(Some(SimpleReader readLine _), classes)
-  def selectPackageMain(classes: Seq[String]): Option[String] =
+  def pickMainClass(classes: Seq[String]): Option[String] =
     sbt.SelectMainClass(None, classes)
+  private def pickMainClassOrWarn(classes: Seq[String], logger: Logger): Option[String] = {
+    classes match {
+      case Nil                           => logger.warn("No main class detected")
+      case multiple if multiple.size > 1 => logger.warn("Multiple main classes detected.  Run 'show discoveredMainClasses' to see the list")
+      case _                             =>
+    }
+    pickMainClass(classes)
+  }
 
   def doClean(clean: Seq[File], preserve: Seq[File]): Unit =
     IO.withTemporaryDirectory { temp =>
--- a/notes/0.13.7.markdown
+++ b/notes/0.13.7.markdown
@@ -11,6 +11,7 @@
   [@tmandke]: https://github.com/tmandke
   [@topping]: https://github.com/topping
   [@WarsawScala]: https://github.com/WarsawScala
+  [@kretes]: https://github.com/kretes
   [1237]: https://github.com/sbt/sbt/issues/1237
   [1430]: https://github.com/sbt/sbt/issues/1430
   [1544]: https://github.com/sbt/sbt/issues/1544
@@ -33,6 +34,7 @@
   [1631]: https://github.com/sbt/sbt/pull/1631
   [1642]: https://github.com/sbt/sbt/pull/1642
   [1683]: https://github.com/sbt/sbt/pull/1683
+  [1648]: https://github.com/sbt/sbt/pull/1648
 
 ### Fixes with compatibility implications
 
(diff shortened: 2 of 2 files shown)


lightbend-cla-validator:
Hi @kretes,

typesafe-tools:

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -662,10 +662,17 @@ object Defaults extends BuildCommon {
       new Package.Configuration(srcs, path, options)
     }
 
-  def selectRunMain(classes: Seq[String]): Option[String] =
+  def askForMainClass(classes: Seq[String]): Option[String] =
```
jsuereth, line 669: So, we can't actually change the name of this method without breaking binary compatibility.  I agree that your name is better, so you can keep the old one and deprecate it if you want.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -662,10 +662,17 @@ object Defaults extends BuildCommon {
...
     sbt.SelectMainClass(Some(SimpleReader readLine _), classes)
-  def selectPackageMain(classes: Seq[String]): Option[String] =
+  def pickMainClass(classes: Seq[String]): Option[String] =
```
jsuereth, line 671: ditto on new name here.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -662,10 +662,17 @@ object Defaults extends BuildCommon {
...
     sbt.SelectMainClass(None, classes)
+  def warnAboutNoMainClass(classes: Seq[String]): Option[String] = {
```
jsuereth: have this guy take in a `sbt.Logger`and do "log.warn"

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -264,9 +264,9 @@ object Defaults extends BuildCommon {
     definedSbtPlugins <<= discoverPlugins,
     discoveredSbtPlugins <<= discoverSbtPluginNames,
     inTask(run)(runnerTask :: Nil).head,
-    selectMainClass := mainClass.value orElse selectRunMain(discoveredMainClasses.value),
+    selectMainClass := pickMainClass(discoveredMainClasses.value) orElse askForMainClass(discoveredMainClasses.value),
     mainClass in run := (selectMainClass in run).value,
-    mainClass := selectPackageMain(discoveredMainClasses.value),
+    mainClass := pickMainClass(discoveredMainClasses.value) orElse warnAboutNoMainClass(discoveredMainClasses.value),
```
jsuereth: You can pass in a logger with:

``` scala
warnAboutNoMainClass(discoveredMainClasses.value, streams.value.log)
```

jsuereth:
Commented as I could.  Thanks for adding this! Looking forward to the new warning.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -662,10 +662,21 @@ object Defaults extends BuildCommon {
...
     sbt.SelectMainClass(Some(SimpleReader readLine _), classes)
-  def selectPackageMain(classes: Seq[String]): Option[String] =
+  def pickMainClass(classes: Seq[String]): Option[String] =
     sbt.SelectMainClass(None, classes)
+  private def warnAboutNoMainClass(classes: Seq[String], logger: Logger): Option[String] = {
+    classes match {
+      case Nil                           => logger.warn("No main class detected")
+      case multiple if multiple.size > 1 => logger.warn("Multiple main classes detected.  Run 'show discoveredMainClasses' to see the list")
+    }
+    None
```
jsuereth: shouldn't this be `classes.headOption` ?

jsuereth:
One question, otherwise LGTM

main/src/main/scala/sbt/Defaults.scala
kretes: warnAboutNoMainClass is only used as an argument to 'orElse' when sbt.SelectMainClass(None, classes) yields None. That's why it returns None, as a "pure side-effect" ;-) function.

returning classes.headOption would change the way sbt resolves main class for ambiguous results. 

alternative way of expressing that would be to rewrite warnAboutNoMainClass as:

```
private def pickMainClassOrWarn(classes: Seq[String], logger: Logger): Option[String] = {
    classes match {
      case Nil                           => logger.warn("No main class detected")
      case multiple if multiple.size > 1 => logger.warn("Multiple main classes detected.  Run 'show discoveredMainClasses' to see the list")
    }
    sbt.SelectMainClass(None, classes)
  }
```

and use it directly `mainClass := pickMainClassOrWarn(discoveredMainClasses.value, streams.value.log)`

jsuereth: Yeah, I think I'd like the above option better, or make "warnAboutNoMainClass as its own standalone method returning "Unit" and just call it from the right location.

In any case, generally LGTM, but we need you to rebase as we've drifted from master.

kretes:
Rebased and hopefully ready.

WRT to depracation I wonder if this is really the case to keep methods as deprecated - I searched github, noone is using the methods to calculate main class.

jsuereth:
Regarding deprecation, we like to be cautious.  Ideally the new methods become `private[sbt]` so it's no longer a concern in the future.  You may notice I do a lot of `deprecated("No longer public")` in the future for things like this.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -662,10 +662,21 @@ object Defaults extends BuildCommon {
...
+  @deprecated("use Defaults.pickMainClass", "0.13.7")
+  def selectPackageMain(classes: Seq[String]): Option[String] = pickMainClass(classes)
+  def askForMainClass(classes: Seq[String]): Option[String] =
     sbt.SelectMainClass(Some(SimpleReader readLine _), classes)
-  def selectPackageMain(classes: Seq[String]): Option[String] =
+  def pickMainClass(classes: Seq[String]): Option[String] =
     sbt.SelectMainClass(None, classes)
+  private def pickMainClassOrWarn(classes: Seq[String], logger: Logger): Option[String] = {
+    classes match {
+      case Nil                           => logger.warn("No main class detected")
```
jsuereth, line 675: I think this needs to be `Seq()`

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -662,10 +662,21 @@ object Defaults extends BuildCommon {
...
+      case multiple if multiple.size > 1 => logger.warn("Multiple main classes detected.  Run 'show discoveredMainClasses' to see the list")
+    }
```
jsuereth, line 678: You need a default case that does nothing, otherwise this will throw MatchError, see: https://travis-ci.org/sbt/sbt/jobs/38639951#L4395

jsuereth:
The tests caught something, commented on the error.

main/src/main/scala/sbt/Defaults.scala
kretes, line 675: why?
Currently it works, and is the same as in https://github.com/sbt/sbt/blob/0.13/run/src/main/scala/sbt/SelectMainClass.scala#L11

main/src/main/scala/sbt/Defaults.scala
kretes, line 678: right. only now I noticed how are the 'integration' tests for sbt projects done. Is there a way to verify some log output in those 'scripted' tests ? I could add some test for that

main/src/main/scala/sbt/Defaults.scala
jsuereth, line 675: Ah, nevermind.  That's my bad.

PR #1672, opened 2014-10-14
Local ivy repository defined in sbt launcher configuration doesn't respect useOrigin

wchen9911:
Sbt sets useOrigin=true for the default ivy cache, however, that does not work for any ivy repository defined in sbt launcher configuration files. This is because ivy's useOrigin only applies to FileSystemResolver, but not URLResolver. 

A patch was proposed, but doesn't seem be taken: http://ant.1045680.n5.nabble.com/PATCH-Ivy-useOrigin-true-support-for-URLResolver-with-file-URLs-td5714566.html

This pull request will create FileSystemResolver if the ivy repository defined in the sbt launcher configuration file is a local file system one.

Also a unit test is added to this. The risk should be minimal.

### DIFF
--- a/launch/src/main/scala/xsbt/boot/Update.scala
+++ b/launch/src/main/scala/xsbt/boot/Update.scala
@@ -305,12 +305,12 @@ final class Update(config: UpdateConfiguration) {
     settings.addRepositoryCacheManager(manager)
     settings.setDefaultRepositoryCacheManager(manager)
   }
-  private def toIvyRepository(settings: IvySettings, repo: xsbti.Repository) =
+  private[boot] def toIvyRepository(settings: IvySettings, repo: xsbti.Repository) =
     {
       import xsbti.Predefined._
       repo match {
         case m: xsbti.MavenRepository => mavenResolver(m.id, m.url.toString)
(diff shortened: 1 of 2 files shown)


typesafe-tools:

launch/src/main/scala/xsbt/boot/Update.scala
```diff
@@ -327,12 +327,13 @@ final class Update(config: UpdateConfiguration) {
     }
   }
   /** Uses the pattern defined in BuildConfiguration to download sbt from Google code.*/
-  private def urlResolver(id: String, base: String, ivyPattern: String, artifactPattern: String, mavenCompatible: Boolean, descriptorOptional: Boolean, skipConsistencyCheck: Boolean) =
+  private def ivyResolver(id: String, base: String, ivyPattern: String, artifactPattern: String, mavenCompatible: Boolean, descriptorOptional: Boolean, skipConsistencyCheck: Boolean) =
     {
-      val resolver = new URLResolver
+      val (resolver, newBase) = if (base.startsWith("file:")) (new FileSystemResolver, base.substring(5)) else (new URLResolver, base)
+      println(base)
```
wchen9911, line 333: Forgot to delete this.

PR #1683, opened 2014-10-20
Revert chain resolver  by default, for older behavior.

jsuereth:
The issue comes into play where we cannot accurately get a publication date from Maven artifacts, leading to the current
mechanism having undefined behavior and causing other bugs to pop up in resolution.

Review by @eed3si9n 

Ref #1520, #1514, #321


eed3si9n:
LGTM. waiting for Travis.

PR #1686, opened 2014-10-21
Fixes #1684. cached resolution applies overrides on conflict

### LINKED ISSUE #1684: [0.13.7-M3] cached resolution does not respect dependencyOverrides
## steps
1. Make a build that uses `"net.databinder" %% "unfiltered-uploads" % "0.8.0"` (this brings in commons-io 1.4) with cached resolution enabled, and resolve it.
2. Make another build that uses `"net.databinder" %% "unfiltered-uploads" % "0.8.0"` with cached resolution enabled and the following overrides:

``` scala
    dependencyOverrides += "commons-io" % "commons-io" % "2.0"
```
## problems

Resulting classpath includes commons-io 1.4.
## expectation

When there are overrides, commons-io 2.0 is used.
## original report

withCachedResolution(true) resolves dependencies differently than withCachedResolution(false) and I would expect them to resolve the same way, except for one is cached and one is not.  I am getting several builds with odd behavior, only to find out the dependencies have been affected by trying to cache resolution to speed up the resolution.   

I want to use the cache because I have approximately 1000 modules defined in dependencyOverrides so that I can lock down versions.  Those don't seem to be honored if resolution cache is used.

I have attached a diff of just one application to show the different libraries that were included. 

If you need specific SBT output, please tell me what you need and how to get it.

The diff patch file can be found here:

https://github.com/jasonchaffee/sbt-1684/blob/master/dependency-diff.patch

eed3si9n:
#1684

This change brings over dependency overrides to artificial graph.
However, it seems forced might win, so I need to take overrides logic
in account during conflict resolution.

### DIFF
--- a/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
+++ b/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
@@ -12,9 +12,11 @@ import core.resolve._
 import core.module.id.{ ModuleRevisionId, ModuleId => IvyModuleId }
 import core.report.{ ResolveReport, ConfigurationResolveReport, DownloadReport }
 import core.module.descriptor.{ DefaultModuleDescriptor, ModuleDescriptor, DependencyDescriptor, Configuration => IvyConfiguration, ExcludeRule, IncludeRule }
+import core.module.descriptor.OverrideDependencyDescriptorMediator
 import core.{ IvyPatternHelper, LogOptions }
 import org.apache.ivy.util.Message
 import org.apache.ivy.plugins.latest.{ ArtifactInfo => IvyArtifactInfo }
+import org.apache.ivy.plugins.matcher.{ MapMatcher, PatternMatcher }
 
 private[sbt] object CachedResolutionResolveCache {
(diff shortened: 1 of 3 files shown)


ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -82,8 +85,10 @@ private[sbt] class CachedResolutionResolveCache() {
           case rules    => Some(conf + "->(" + (rules map includeRuleString).mkString(",") + ")")
         }
       })
+      val os = extractOverrides(parent)
+      val moduleLevel = s"""dependencyOverrides=${os.mkString(",")}"""
```
jsuereth, line 89: so bascially, we're just toString-ing overrides into the hash?

jsuereth:
LGTM

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
eed3si9n, line 89: That and actually porting the overrides over to mini graphs at line 101.

PR #1692, opened 2014-10-25
Events reported by forked tests should be sequence of (String, SuiteResult)

fiadliel:
Currently events reported by forked tests are a `Map[String, SuiteResult]`. With some configurations (in particular, nested suites in ScalaTest), the suite name is the same for all nested suites. Using a Map causes loss of reporting information for all but one tests with the same name.

Changing this to a Seq allows all results to be reported.

This is related to https://github.com/scalatest/scalatest/issues/167
Also related to https://github.com/sbt/sbt/pull/1364
But this fixes the issue when `fork in Test := true`.

Example failure at https://gist.github.com/fiadliel/4071e61b3fd9331a81d4

### DIFF
--- a/main/actions/src/main/scala/sbt/Tests.scala
+++ b/main/actions/src/main/scala/sbt/Tests.scala
@@ -25,7 +25,7 @@ object Tests {
    * @param events The result of each test group (suite) executed during this test run.
    * @param summaries Explicit summaries directly provided by test frameworks.  This may be empty, in which case a default summary will be generated.
    */
-  final case class Output(overall: TestResult.Value, events: Map[String, SuiteResult], summaries: Iterable[Summary])
+  final case class Output(overall: TestResult.Value, events: Seq[(String, SuiteResult)], summaries: Iterable[Summary])
 
   /**
    * Summarizes a test run.
@@ -246,10 +246,10 @@ object Tests {
     }
 
   def processResults(results: Iterable[(String, SuiteResult)]): Output =
-    Output(overall(results.map(_._2.result)), results.toMap, Iterable.empty)
+    Output(overall(results.map(_._2.result)), results.toSeq, Iterable.empty)
   def foldTasks(results: Seq[Task[Output]], parallel: Boolean): Task[Output] =
     if (results.isEmpty)
-      task { Output(TestResult.Passed, Map.empty, Nil) }
+      task { Output(TestResult.Passed, collection.mutable.ListBuffer.empty, Nil) }
     else if (parallel)
       reduced(results.toIndexedSeq, {
         case (Output(v1, m1, _), Output(v2, m2, _)) => Output(if (v1.id < v2.id) v2 else v1, m1 ++ m2, Iterable.empty)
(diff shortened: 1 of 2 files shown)


typesafe-tools:

main/actions/src/main/scala/sbt/Tests.scala
```diff
@@ -25,7 +25,7 @@ object Tests {
    * @param events The result of each test group (suite) executed during this test run.
    * @param summaries Explicit summaries directly provided by test frameworks.  This may be empty, in which case a default summary will be generated.
    */
-  final case class Output(overall: TestResult.Value, events: Map[String, SuiteResult], summaries: Iterable[Summary])
+  final case class Output(overall: TestResult.Value, events: Seq[(String, SuiteResult)], summaries: Iterable[Summary])
```
jsuereth, line 28: Unfortunately, this breaks binary compatibility :(.  I think what you've done is correct, but we cannot accept this in the 0.13.x series, unfortunately, due to breakages.

What you could do:

``` scala
final case class Output(overall: TestResult.Value, eventSequence: Seq[(String, SuiteResult)], summaries: Iterable[Summary]) {
  @deprecated("0.13.8", "Use the event sequence version instead")
  def this(overall: TestResult.Value, events: Map[String, SuiteResult], summaries: Iterable[Summary]) =
     this(overall, events.toSeq, summaries)
  @deprecated("0.13.8", "use eventSequence instead")
  final def events: Map[String, SuiteResult] = eventSequence.toMap
  // Override copy and unapply methods..... (lots of ugly work)
}
```

Option #2 is to create a new type "Output2" or some such while deprecating this entire case class.

Option #3, of course, is to wait for sbt 1.0.

fiadliel:
Option 1: while I'm willing to do the work, and I think it almost compiles, the resulting API seems quite dissonant --- Output.unapply providing a different signature to the main constructor feels wrong.

Option 2: seems ugly but only needs to be kept until 1.0. I think it ends up better for the user.

The issue is not urgent for us, since it's a known issue (now), and it's obvious what scalatest features to avoid. But I think a fix is worth adding, even if the API is ugly, because the bug could be an unexploded landmine for a lot of people...

WDYT?

jsuereth:
Yeah, I agree the API gets super ugly/bad.   Let try Option 2.  I don't like leaving landmines either. If there's anyway we can help here, let me know.  Pretty soon, we'll be refreshing the 1.0 branch for breaking changes, and we can try to cross-post this one.

jsuereth:
@fiadliel Would you be ok if I moved this PR into a "feature/test-improvements" branch for us to tweak this around?  You can submit further PRs there until it's ready to go into 0.13 or 1.0 branches.

fiadliel:
Yes, that's fine. Sorry I didn't get back to this yet this week.

dwijnand:
@fiadliel you still interested in this?

eed3si9n:
Closing this for now due to inaction.
Please reopen this against 1.0.x branch if you're still interested.

PR #1694, opened 2014-10-25
Fix the run logic so the mainClass setting is actually used when set.

jsuereth:
This is one of those super hard to test scenarios where the "main class" being run is a from a different JAR (i.e. we'll never discover one).    More importantly, the run task wasn't using the configured main class due to, what appears to be, a refactoring mixup we didn't catch during code review of the new select-main-class warning.

Review by @eed3si9n cc @fommil

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -264,7 +264,7 @@ object Defaults extends BuildCommon {
     definedSbtPlugins <<= discoverPlugins,
     discoveredSbtPlugins <<= discoverSbtPluginNames,
     inTask(run)(runnerTask :: Nil).head,
-    selectMainClass := pickMainClass(discoveredMainClasses.value) orElse askForMainClass(discoveredMainClasses.value),
+    selectMainClass := mainClass.value orElse askForMainClass(discoveredMainClasses.value),
     mainClass in run := (selectMainClass in run).value,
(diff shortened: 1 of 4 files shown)


eed3si9n:
#1669
This should be reviewed by @kretes

kretes:
This will not change what was introduced in #1669 .
Pity there are no test for those behaviours. Could you point to some test that for given sbt project verifies what the actual value of some sbt setting is? I guess it's needed here to avoid regression in the future

jsuereth:
I was actually used a custom publishe dproject for the test.  I'll see if I can expand the regression test suite with such a test, albeit I'm hesitant to do so, as anythign which requries publishing to a local artifactory is cause of flakiness in travis (it seems).

jsuereth:
Ok guys, I added an automated test.

kretes:
Yes, that works. However this scripted test scenario is not really easy to follow.
I would imagine an assertion like 'ensure mainClass is "Test"' instead of specifying an sbt job to run.
I should dig into scripted plugin more to understand how to use it better

jsuereth:
@jaceklaskowski The inter-project resolver is how we fake ivy into understanding internal project dependencies within an sbt build.  What sbt does for "dependsOn" is two-fold.
1. We set up a dependency in IVY so any generated poms/ivy files are correct in specifying the dependencies
2. We use inter-project resolver and sbt-task dependencies to figure out what the classpath should be and ensure the sub-project is compiled before the dependent project.

So inter-project comes from sbt-itself.

Issue #1696, opened 2014-10-25
Resolvers doc out of sync: bintrayRepo and jcenterRepo don't exist


erikkaplun:
The doc at http://www.scala-sbt.org/0.13/docs/Resolvers.html has:

> Resolver.bintrayRepo("owner", "repo") This is the Bintray repository at https://dl.bintray.com/[owner]/[repo]/
> Resolver.jcenterRepo This is the Bintray JCenter repository at https://jcenter.bintray.com/

However, these do not seem to exist in reality:

> error: value jcenterRepo is not a member of object sbt.Resolver
> error: value bintrayRepo is not a member of object sbt.Resolver

I haven't checked the rest of the helpers.

jsuereth:
Which version of sbt are you using?

erikkaplun:
0.13.6 (sorry, should've mentioned)

jsuereth:
Can yu try sbt 0.13.7-RC1

erikkaplun:
Actually before it even gets to the `build.sbt` part, I get:

```
java.lang.NoSuchMethodError: sbt.Def$.derive(Lsbt/Init$Setting;ZLscala/Function1;Lscala/Function1;)Lsbt/Init$Setting;
    at wartremover.package$.wartremoverSettings$lzycompute(package.scala:20)
    at wartremover.package$.wartremoverSettings(package.scala:11)
    at wartremover.WartRemover$.projectSettings$lzycompute(WartRemover.scala:16)
    at wartremover.WartRemover$.projectSettings(WartRemover.scala:16)
```

jsuereth:
Hmm, this looks like a binary compatibility issue with 0.13.7-RC1 and the wart remover plugin.  Let me investigate that one.

jsuereth:
Hmm, that incompatibility existed in 0.13.6, so I'm curious how you're only hitting it now....

dwijnand:
Doh, having just fixed it.. oh well :)

https://github.com/typelevel/wartremover/issues/123

avdv:
Hi.

I also get this error:

```
build.sbt:8: error: value bintrayRepo is not a member of object sbt.Resolver
resolvers += Resolver.bintrayRepo("scalaz", "releases")
                      ^
[error] Type error in expression
```

sbt about:

```
[info] This is sbt 0.13.7
[info] No project is currently loaded
[info] sbt, sbt plugins, and build definitions are using Scala 2.10.4
```

[...]

PR #1698, opened 2014-10-28
Fix BC issue discovered in #1696.

jsuereth:
Def.derive has a new parameter, so we add an override which delegates down to the new method.

Review by @eed3si9n


eed3si9n:
LGTM

PR #1700, opened 2014-10-28
Fixes #1699. ignore transitive force during cached resolution

### LINKED ISSUE #1699: cached resolution is picking non-latest winners on conflict
originally reported by @fommil in sbt-dev [user caching of resolution results?](https://groups.google.com/d/msg/sbt-dev/XEPDaXVug7Q/32ymVIDycfsJ)
## steps

https://github.com/fommil/ensime-server/tree/sbt-speedup
From sbt, run `debug`,  `update` once, then, `show update` for the second time.
## problem

```
[warn] Scala version was updated by one of library dependencies:
[warn]  * org.scala-lang:scala-library:(2.11.2, 2.11.1) -> 2.11.0
[warn] To force scalaVersion, add the following:
[warn]  ivyScala := ivyScala.value map { _.copy(overrideScalaVersion = true) }
```

See https://gist.github.com/eed3si9n/1a1eac83da82037a963d for full log.
## expectation

2.11.2 is selected since it's specified as `scalaVersion` and it's the latest.
## note

In the first run I see

```
[debug] - conflict in compile:org.scala-lang:scala-library (org.scala-lang:scala-library:2.11.0 (), org.scala-lang:scala-library:2.11.2 ())
[debug] - forced dependency:    org.scala-lang:scala-library:2.11.0 (): (Artifact(scala-library,jar,jar,None,ArraySeq(),None,Map()),/Users/eugene/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.11.0.jar)
```

This indicates that scala-library 2.11.0 was chosen because it is (supposedly) a transitively forced dependency.

eed3si9n:
#1699

When Ivy translates pom to ivy.xml, it adds force=“true”.
So when both non-Maven dependencies and Maven dependencies are mixed, Maven dependencies always wins, which is the case for scala-library dependency added by the user.


sbt/src/sbt-test/dependency-management/cached-resolution/test
```diff
@@ -11,5 +11,3 @@
 > c/clean
 
 > check
-
-> check2
```
jsuereth, line 15: no more check2?

jsuereth:
LGTM.  Could we also attackt he POM parser so that force isn't always enabled to on?

I guess the issue there is it wouldn't actually fix existing users cache....

sbt/src/sbt-test/dependency-management/cached-resolution/test
eed3si9n, line 15: check2 is now cached-resolution4.

eed3si9n:
Travis caught legit test failure, which was relying on old behavior. Added a commit with modified test.

jsuereth:
We also need to document the order of eviction rules here....

Essentially "cachedResolution = true" is becoming _our_ resolution engine....

eed3si9n:
> We also need to document the order of eviction rules here

Added comments. See 9482509b321e4860bbcf97589ed9a8669a7e2930

jsuereth:
LGTM.  Going to merge this as it's a doc-only change.

jsuereth:
meant to say - doc-only change since trvis passed.

Issue #1679, opened 2014-10-16
0.13.7-M3 + Play 2.3.5 + withCachedResolution


ghost:
Not sure if I should post here or at the playframework project.

Sbt: 0.13.7-M3
Scala: 2.11.2
Play: 2.3.5

``` scala
import play.PlayScala
import sbt._
import sbt.Keys._

object ApplicationBuild extends Build {

  val appDependenciesResolvers = Seq(
    "Typesafe Repository" at "http://repo.typesafe.com/typesafe/releases/",
    "Sonatype Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots/"
  )

  val main = Project("main", file("."))
    .enablePlugins(PlayScala)
    .settings(
      resolvers ++= appDependenciesResolvers,
      updateOptions ~= {_.withCachedResolution(true)},
      scalaVersion := "2.11.2",
      doc in Compile <<= target.map(_ / "none")
    )

}
```

Throws exception;

```
[main] $ run
[info] Updating {file:/Users/mark/Projects/test-project/}main...
[info] :: resolving dependencies :: org.scala-sbt.temp#temp-resolve-4eaae06905cddc4c510c340abfce50cc215a9998;1.0
[info]  confs: [compile, runtime, test, provided, optional, compile-internal, runtime-internal, test-internal, plugin, sources, docs, pom, scala-tool]
[info] Resolving org.reactivemongo#reactivemongo_2.11;0.10.5.akka23-SNAPSHOT ...
[info]  found org.reactivemongo#reactivemongo_2.11;0.10.5.akka23-SNAPSHOT in sbt-chain
[info] Resolving org.scala-lang#scala-library;2.11.1 ...
[info]  found org.scala-lang#scala-library;2.11.1 in sbt-chain
[info]  [2.11.1] org.scala-lang#scala-library;2.11.1
[info] Resolving org.reactivemongo#reactivemongo-bson-macros_2.11;0.10.5.akka23-SNAPSHOT ...
[info]  found org.reactivemongo#reactivemongo-bson-macros_2.11;0.10.5.akka23-SNAPSHOT in sbt-chain
[info] Resolving org.scala-lang#scala-compiler;2.11.1 ...
[info]  found org.scala-lang#scala-compiler;2.11.1 in sbt-chain
[info] Resolving org.scala-lang#scala-reflect;2.11.1 ...
[info]  found org.scala-lang#scala-reflect;2.11.1 in sbt-chain
[...]

PR #1713, opened 2014-11-02
Fixes cached resolution mutual eviction (drops slf4j)

### LINKED ISSUE #1710: force from a module can mutually evict all versions in cached resolution
## steps

define a project X1 that uses library Y1 and Y2 and use cached resolution.
- both Y1 and Y2 includes dependencies that includes slf4j-api 1.6.6 and 1.7.5 transitively.
- in Y1 add slf4j-api to 1.6.6 with force to mark 1.7.5 as evicted.
- Y1 naturally marks 1.6.6 as evicted.

``` scala
lazy val X1 = project.
  settings(cachedResolutionSettings: _*).
  settings(
    libraryDependencies ++= Seq(
      "com.example" %% "y1" % "0.1.0" % "compile->compile;runtime->runtime",
      "com.example" %% "y2" % "0.1.0" % "compile->compile;runtime->runtime")
  )

lazy val Y1 = project.
  settings(commonSettings: _*).
  settings(
    name := "y1",
    libraryDependencies ++= Seq(
      // this includes slf4j 1.7.5
      "com.ning" % "async-http-client" % "1.8.14",
      // this includes slf4j 1.6.6
      "com.twitter" % "summingbird-core_2.10" % "0.5.0",
      "org.slf4j" % "slf4j-api" % "1.6.6" force()
    )
  )

lazy val Y2 = project.
  settings(commonSettings: _*).
  settings(
    name := "y2",
    libraryDependencies ++= Seq(
      // this includes slf4j 1.6.6
      "com.twitter" % "summingbird-core_2.10" % "0.5.0",
      // this includes slf4j 1.7.5
      "com.ning" % "async-http-client" % "1.8.14")
  )
```
## problem

none of the slf4j gets included.
## expectation

ignore the overrides from either side, and pick a winner.

eed3si9n:
Fixes #1710, #1679

When stitching the minigraphs together only exclude the artifacts that were evicted in _all_ graphs, instead of some graphs.

Consider the following scenario:
- Y1 evicts slf4j-api 1.6.6 and picks 1.7.5
- Y2 evicts slf4j-api 1.7.5 and picks 1.6.6

At the root level, we need to use our own judgement and pick 1.7.5.
The fix is one-liner change:

``` scala
else Vector(xs.head.copy(evicted = xs forall { _.evicted }, callers = xs flatMap { _.callers }))
```

/cc @jsuereth


jsuereth:
LGTM

PR #1714, opened 2014-11-04
Expose mechanism whereby bytecode enhancement can be run *before* saving incremental compiler hashes.

jsuereth:
(post dated notes by @eed3si9n) After code review the structure was change such that:
1. `compile` task depends on `manipulateBytecode` task.
2. By default `maniputateBytecode` task delegates to `compileIncremental` task.
3. `compileIncremental` task depends on `previousCompile` task.
4. `previousCompile` task reads `Analysis` instance from the cache, if any.
5. `compileIncremental` task does the actual compilation.
6. After `manipulateBytecode` is done, `compile` task saves the `Analysis` into the cache.
## original description

Originally submitted as #1362. Related playframework/playframework#2237.

Splits the compile task into three components:
1. readAnalysis
2. compileIncremental
3. saveAnalysis

Byte code enhancers can inject themselves into 2, although we may want to directly expose an API for bytecode enhancers in the future.

Additionally, this PR attempts to clean up some names throughout the incremental compiler (and add toString for debugging).   Please open the bikeshed arguments on naming, as I'd like to isolate what these components should be called.

Review by @gkossakowski and @eed3si9n cc> @jroper


compile/integration/src/main/scala/sbt/compiler/MixedAnalyzingCompiler.scala
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+  def withBootclasspath(args: CompilerArguments, classpath: Seq[File]): Seq[File] =
+    args.bootClasspathFor(classpath) ++ args.extClasspath ++ args.finishClasspath(classpath)
+
+  /**
+   * The first level of compilation.  This checks to see if we need to skip compiling because of the skip flag,
+   * IF we're not skipping, this setups up the `CompileConfiguration` and delegates to `compile2`
+   */
+  private def compile1(sources: Seq[File],
```
eed3si9n: What's the point of having this as a method? It's a wrapper around if-expression that's called only by `analyzingCompile`.

compile/integration/src/main/scala/sbt/compiler/MixedAnalyzingCompiler.scala
```diff
@@ -0,0 +1,233 @@
...
+import sbt.inc._
+import sbt.inc.Locate
+import xsbti.{ AnalysisCallback, Reporter }
+import xsbti.api.Source
+import xsbti.compile.CompileOrder._
+import xsbti.compile._
+
+/**
+ * This is a compiler that mixes the `sbt.compiler.AnalyzingCompiler` for Scala incremental compilation
+ * with a `xsbti.JavaCompiler`, allowing cross-compilation of mixed Java/Scala projects with analysis output.
```
eed3si9n, line 95: > This is a compiler that mixes the `sbt.compiler.AnalyzingCompiler` ... with a `xsbti.JavaCompiler`

Is there a situation in which that is not desirable? Why does this warrant a creation of a new object instead of just adding methods to existing `IC` or something?

eed3si9n, line 95: Ok. So this is `AggressiveCompile` without `cacheFile: File` parameter. I feel like the name should reflect the caching aspect rather than the mixing.

compile/integration/src/main/scala/sbt/compiler/MixedAnalyzingCompiler.scala
jsuereth: This is copy-pasted. I just tried to document what it does.   Think of it as separation of concerns, as the implementation is pretty big of the actual compile method.

compile/integration/src/main/scala/sbt/compiler/MixedAnalyzingCompiler.scala
jsuereth, line 95: It's actually exactly the same as AggressiveCompile.  Just AggressiveCompile was  bad name.   This is bascially an Analyzing compiler that handles both Javac + Scalac, while the AnalyzingCompiler trait only does Scala.

If you have a better idea for a name, let me know!

main/src/main/scala/sbt/Keys.scala
```diff
@@ -141,6 +142,9 @@ object Keys {
   val consoleQuick = TaskKey[Unit]("console-quick", "Starts the Scala interpreter with the project dependencies on the classpath.", ATask, console)
   val consoleProject = TaskKey[Unit]("console-project", "Starts the Scala interpreter with the sbt and the build definition on the classpath and useful imports.", AMinusTask)
   val compile = TaskKey[Analysis]("compile", "Compiles sources.", APlusTask)
+  val compileIncremental = TaskKey[Compiler.CompileResult]("compileIncremental", "Actually runs the incremental compliation", DTask)
+  val readAnalysis = TaskKey[Compiler.PreviousAnalysis]("readAnalysis", "Read the incremental compiler analysis from disk", DTask)
```
eed3si9n: `previousAnalysis` maybe?

main/src/main/scala/sbt/Keys.scala
```diff
@@ -141,6 +142,9 @@ object Keys {
...
+  val saveAnalysis = TaskKey[Analysis]("saveAnalysis", "Save the incremental compiler analysis to disk", DTask)
```
eed3si9n: `currentAnalysis` to be symmetric with `previousAnalysis`. It's hard to come up with a good name because I don't think in sbt we ever explicitly mention the fact that we save things in keys, or try to make break up a post-step.

jsuereth: I think currentAnalysis is kind of a misnomer. IT's like the "saveAnalysisAfterCompileAndReturnIt" task.

eed3si9n: If it were for any other task, "saveAnalysisAfterCompileAndReturnIt" would just be called `compile` like we don't call `update` "saveUpdateReportUsingSbinaryAfterUpdateAndReturnIt." The explicit mentioning of "saving" sticks out.

jsuereth: What if we hid it inside of the "compile" task scope?  compile:::saveAnalysis and compile::incrementalCompile, e.g.

gkossakowski:
Hi guys. I just skimmed over this very quickly so this wasn't detailed review. I was mostly checking what's overall theme of those patches. I agree with the direction expressed in commit messages.

jsuereth:
@eed3si9n fixed your comments, this one should be ready to roll.

eed3si9n:
@jsuereth Do we have release notes on this feature?

jsuereth:
you mean something other than the `notes/` markdown?

jsuereth:
hmm, actually no, we dont'.  I'll create a PR shortly

Issue #1704, opened 2014-10-30
Custom ivyXML not respected with cached resolution


dpratt:
A couple of my builds use a custom ivyXML setting to do explicit global excludes for unwanted transitive dependencies. For example, I do a global exclude of log4j, commons-logging, and a few other things not needed by my project. It looks like this isn't being respected when cached resolution is used. I'm not terribly surprised by this, since adding custom stuff to the ivy descriptor is certainly an inelegant hack. Is there a way to accomplish this same goal without hunting down each library that has the unwanted transitives and manually adding entries to each libraryDependency element? Here's an example from one of my builds that fails to properly exclude them when cached resolution is enabled:

``` scala
ivyXML :=
  <dependencies>
    <exclude module="log4j"/>
    <exclude module="slf4j-log4j12"/>
    <exclude module="commons-logging"/>
    <exclude module="wstx-asl"/>
    <exclude module="jta"/>
    <exclude module="solr-solrj"/>
    <exclude org="org.eclipse.jetty"/>
    <exclude org="org.sonatype.sisu.inject"/>
    <exclude org="com.sun.jersey"/>
    <exclude org="com.google.caliper"/>
    <exclude org="com.google.code.findbugs"/>
    <exclude org="junit"/>
  </dependencies>
```

PR #1715, opened 2014-11-05
Fixes #1711 and #1704. Cached resolution fixes

### LINKED ISSUE #1711: [0.13.7-RC2] cached resolution pulls in % "test" dependencies from other subproj

eed3si9n:
## fix for #1711

Cached resolution flattens external dependencies coming from subproject dependencies.
That process needs to ignore non-compile configurations, otherwise I'd end up with test dependencies from subprojects.
## fix for #1704

Exclude rules at the project level (using `ivyXML`) are now ported over to artificial graph so it's respected.

### DIFF
--- a/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
+++ b/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
@@ -87,8 +87,10 @@ private[sbt] class CachedResolutionResolveCache() {
           case rules    => Some(conf + "->(" + (rules map includeRuleString).mkString(",") + ")")
         }
       })
+      val mes = parent.getAllExcludeRules.toVector
+      val mesStr = (mes map excludeRuleString).mkString(",")
       val os = extractOverrides(parent)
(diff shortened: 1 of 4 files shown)


ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -54,8 +54,10 @@ private[sbt] class CachedResolutionResolveCache() {
       def expandInternalDeps(dep: DependencyDescriptor): Vector[DependencyDescriptor] =
         prOpt map {
           _.getModuleDescriptor(dep.getDependencyRevisionId) match {
-            case Some(internal) => directDependencies(internal) flatMap expandInternalDeps
-            case _              => Vector(dep)
+            case Some(internal) => directDependencies(internal) filter { dd =>
+              !dd.getDependencyConfigurations("compile").isEmpty
```
jsuereth: Why is this specific to compile?

jsuereth:
LGTM

Issue #1719, opened 2014-11-09
cached resolution not respecting explicit artifacts


eed3si9n:
## steps

``` scala
lazy val classifierTest = project.
  settings(commonSettings: _*).
  settings(
    artifact in (Compile, packageBin) := Artifact("demo"),
    libraryDependencies := Seq(
      "net.sf.json-lib" % "json-lib" % "2.4" classifier "jdk15" intransitive()
    )
  )

lazy val a = project.
  dependsOn(classifierTest).
  settings(commonSettings: _*).
  settings(
    updateOptions := updateOptions.value.withCachedResolution(true)
  )
```
## problem

```
[info] sbt.ResolveException: download failed: net.sf.json-lib#json-lib;2.4!json-lib.jar
```
## expectation

works because there's a jar under "jdk15" classifier.
## note

artificial graph generation is not picking up classifier under some circumstances.

Issue #1716, opened 2014-11-06
[0.13.7-RC3] "test->test" does not work when cached resolution enabled


xuwei-k:
https://github.com/xuwei-k/sbt-0.13.7-RC3-bug
### Step

`b/src/test/scala/Test.scala`

``` scalaa
import org.scalatest.FunSpec
```

`project/build.properties`

```
sbt.version=0.13.7-RC3
```

`build.sbt`

``` scala
val commonSettings = Seq(
  scalaVersion := "2.11.4",
  updateOptions := updateOptions.value.withCachedResolution(true)
)

lazy val a = project.settings(commonSettings: _*).settings(
  libraryDependencies += "org.scalatest" %% "scalatest" % "2.2.2" % "test"
)

lazy val b = project.settings(commonSettings: _*).dependsOn(
  a,
  a % "test->test"
)
```

and

`sbt test:compile`
### actual

```
[info]   Compilation completed in 14.753 s
[error] /home/travis/build/xuwei-k/sbt-0.13.7-RC3-bug/failure/b/src/test/scala/Test.scala:1: object scalatest is not a member of package org
[error] import org.scalatest.FunSpec
[error]            ^
[error] one error found
[error] (b/test:compile) Compilation failed
[error] Total time: 23 s, completed Nov 6, 2014 7:57:44 PM
```
### expect

`sbt test:compile` should be success

eed3si9n:
The fix I made for #1711 wasn't correct.

xuwei-k:
I confirmed that correction with v0.13.7-RC4. thanks!

PR #1720, opened 2014-11-10
cached resolution: Fixes #1711, #1716, #1719

### LINKED ISSUE #1711: [0.13.7-RC2] cached resolution pulls in % "test" dependencies from other subproj

eed3si9n:
To fix #1711, cached resolution started creating artificial `DependencyDescriptor`s recently (like 4 days ago) af9e520e2817f7bf38cbc3bb474bce4a310a04af, and went out as RC-3.
There are some issues with the implementation, which are reported as #1716 and #1719.
This PR adds scripted test that reproduces those issues and fixes for them, hopefully fixing all three issues.


ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -88,32 +93,44 @@ private[sbt] class CachedResolutionResolveCache() {
...
         {
+          log.debug(s""":::: remapping configuration of ${dd0} with ${confMap.toList map { case (k, v) => (k, v.toList) }}""")
           val dd = new DefaultDependencyDescriptor(md0, dd0.getDependencyRevisionId, dd0.getDynamicConstraintDependencyRevisionId,
             dd0.isForce, dd0.isChanging, dd0.isTransitive)
+          val moduleConfigurations = dd0.getModuleConfigurations.toVector
           for {
-            moduleConf <- dd0.getModuleConfigurations
-            (rootModuleConf, vs) <- confMap
+            moduleConf <- moduleConfigurations
+            (rootModuleConf, vs) <- confMap.toSeq
```
jsuereth, line 104: Does this handle transitive `test->compile->runtime` style mappings?

sbt/src/sbt-test/dependency-management/cached-resolution-classifier/multi.sbt
```diff
@@ -8,31 +8,45 @@ def commonSettings: Seq[Def.Setting[_]] =
...
       "net.databinder" %% "unfiltered-uploads" % "0.8.0",
       "commons-io" % "commons-io" % "1.3",
-      "com.typesafe" % "config" % "0.4.9-SNAPSHOT",
-      "junit" % "junit" % "4.11" % "test"
+      "com.typesafe" % "config" % "0.4.9-SNAPSHOT"
     )
   )
 
 lazy val b = project.
+  dependsOn(classifierTest, transitiveTest % "test->test").
```
jsuereth, line 42: For giggles, why not do `*->*` here (to denote that all your configs depend on all his configs).

jsuereth:
You may need to walk me through this one.

jsuereth:
LGTM

PR #1572, opened 2014-09-03
Don't hardcode existing relations in TextAnalysisFormat

Duhemm:
This follows my work on making it easier to add new dependency kinds in sbt.

> The previous implementation of TextAnalysisFormat contained the list
> of all the existing relations that sbt knew of, and used this
> information to write to and read from the disk the persisted analyses.
> 
> In this knew implementation, TextAnalysisFormat gets from the
> Relations object what are the existing relations, and then persists
> them to disk.
> 
> The previous situation was not optimal since it meant that, in order
> to add a new dependency kind, one had to modify both the Relations
> and TextAnalysisFormat.
> 
> Using this new implementation, no change to TextAnalysisFormat is
> required whenever a new dependency kind is added.

Thanks @gkossakowski for the help !

/cc @xeno-by

PR #1736, opened 2014-11-19
Abstract over dependency kind in Compile

Duhemm:
This follows up on my work to make it easier to add new dependency kind in sbt.

This is the last part that needs to be adapted to take advantage of the new infrastructure that has been introduced by :
- Abstract over dependency kind in Analysis (#1340)
- Don't hardcode relations in TextAnalysisFormat (#1572)

/cc @xeno-by


typesafe-tools:

compile/inc/src/main/scala/sbt/inc/Analysis.scala
```diff
@@ -5,6 +5,7 @@ package sbt
 package inc
 
 import xsbti.api.Source
+import xsbti.DependencyContext._
```
gkossakowski, line 8: Why is this needed? If nothing else is modified in the file I'm guessing this import is a dead code.

Duhemm, line 8: This file uses `DependencyByMemberRef` and `DependencyByInheritance` that were previously defined in the same package as `Analysis`. `DependencyByMemberRef` and `DependencyByInheritance` have been moved to an `enum` in the package `xsbti`, because the interface `AnalysisCallback` uses them.

Duhemm:
Hey,
Any update about the status of this pull request ? Thanks !

jsuereth:
@gkossakowski is this one good to go?

gkossakowski:
LGTM

Yes, merging.

PR #1754, opened 2014-11-29
Check for null type trees in dependency extraction

### LINKED ISSUE #1655: 0.13.6 throws null pointer with macro paradise
clone this https://github.com/MasseGuillaume/sbt-example-paradise

git@github.com:MasseGuillaume/sbt-example-paradise.git

sbt "core/run"

Duhemm:
In some cases the dependency extraction may encounter a null `TypeTree`
(eg. arguments of macro annotations that are untyped). In such cases,
we simply ignore the node.

This solves all the crashes that I've been able to reproduce (https://github.com/sbt/sbt/issues/1655#issue-45323426, julianpeeters/avro-scala-macro-annotations#4, [Duhemm/macro-annotation](https://github.com/Duhemm/macro-annotation))

/cc @xeno-by @dragos

Fixes #1655

### DIFF
--- a/sbt/src/sbt-test/source-dependencies/macro-annotation/project/Build.scala
+++ b/sbt/src/sbt-test/source-dependencies/macro-annotation/project/Build.scala
@@ -1,0 +1,45 @@
+import sbt._
+import Keys._
+
+object BuildSettings {
+  val paradiseVersion = "2.0.1"
+  val buildSettings = Defaults.defaultSettings ++ Seq(
+    version := "1.0.0",
+    scalacOptions ++= Seq(""),
+    scalaVersion := "2.11.4",
+    resolvers += Resolver.sonatypeRepo("snapshots"),
+    resolvers += Resolver.sonatypeRepo("releases"),
+    addCompilerPlugin("org.scalamacros" % "paradise" % paradiseVersion cross CrossVersion.full),
+    incOptions := incOptions.value.withNameHashing(true)
+  )
+}
+
+object MyBuild extends Build {
+  import BuildSettings._
+
+  lazy val root: Project = Project(
+    "root",
+    file("."),
+    settings = buildSettings ++ Seq(
+      run <<= run in Compile in core
+    )
+  ) aggregate(macros, core)
+
+  lazy val macros: Project = Project(
+    "macros",
+    file("macros"),
+    settings = buildSettings ++ Seq(
+      libraryDependencies <+= (scalaVersion)("org.scala-lang" % "scala-reflect" % _),
+      libraryDependencies ++= (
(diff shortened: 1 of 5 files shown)


typesafe-tools:

jsuereth:
Ok, so we can't add the dependency because it's null.  Does this mean if you use a macro annotation that no dependency is ever recorded?

dragos:
LGTM

jsuereth:
Also, is there any way to get a test around this?

Duhemm:
I just added a scripted test that uses a macro annotation. Would you prefer it to be introduced before as a pending test, and then mark it as fixed ?

dragos:
I think it's fine like this (test + fix) in one go.

jsuereth:
this looks great.  will merge once tests pass

Duhemm:
Great, thanks !
Looks like something went wrong with the build: https://travis-ci.org/sbt/sbt/jobs/42855282#L6955
Could you please restart this job ?

sbt/src/sbt-test/source-dependencies/macro-annotation/project/Build.scala
```diff
@@ -0,0 +1,45 @@
+import sbt._
+import Keys._
+
+object BuildSettings {
+  val paradiseVersion = "2.0.1"
+  val buildSettings = Defaults.defaultSettings ++ Seq(
+    version := "1.0.0",
+    scalacOptions ++= Seq(""),
```
jaceklaskowski, line 8: Why is this needed?!

sbt/src/sbt-test/source-dependencies/macro-annotation/project/Build.scala
```diff
@@ -0,0 +1,45 @@
...
+    scalaVersion := "2.11.4",
```
jaceklaskowski, line 9: Shouldn't it be `2.11.5` since the version's out?

sbt/src/sbt-test/source-dependencies/macro-annotation/project/Build.scala
```diff
@@ -0,0 +1,45 @@
...
+    settings = buildSettings ++ Seq(
+      run <<= run in Compile in core
+    )
+  ) aggregate(macros, core)
+
+  lazy val macros: Project = Project(
+    "macros",
+    file("macros"),
+    settings = buildSettings ++ Seq(
+      libraryDependencies <+= (scalaVersion)("org.scala-lang" % "scala-reflect" % _),
```
jaceklaskowski, line 32: I thought that `<+=` got deprecated (or at least new code should use it with care) since `scalaVersion.value` works and looks so much "better".

PR #1757, opened 2014-12-03
Registering dependencies on auxiliary (non-scala) files

Duhemm:
Hi !

I have been working on implementing a technique that would allow sbt to record dependencies on auxiliary files from macro-client.

The idea is to ask the author of the macro to attach to the expanded macro the list of the files that have been used to produce the expansion. Using this information, sbt would be able to invalidate macro-clients whenever a file used to produce a macro expansion is modified. This information would be given to sbt using a `java.util.HashMap[String, Any]` : The macro author would simply have to add a mapping from `touchedFiles` to a `List[java.io.File]` that contains the list of files that have been used to produce the expansion.

The scripted test gives an example of case where we encounter this kind of dependency.

The parts of this work that, I guess, may be the most surprising are :
- Since we do not know where auxiliary files are in general, there's no directory that we can scan and get them all. We only see them if the compiled file advertises them. Therefore, we may not see them at all for a while, even if they have not be deleted. We need to pass them from a previous analysis to the next, because otherwise sbt would consider that they have been deleted.  [Here](https://github.com/sbt/sbt/pull/1757/files#diff-a2b129460f05548355e1345594eafcd8R24)
- Adding a method `allAuxiliary` to `ReadStamps`. The methods of this kind (`allBinary`, ...) are all in the trait `Stamps` that extends `ReadStamps`. This method needs to be here because from method `changedInitial` in `IncrementalCommon.scala` we have an instance of `ReadStamps`. However, we need to check all the stamps on auxiliary dependencies to detect it if they have been changed. [Here](https://github.com/sbt/sbt/pull/1757/files#diff-016e49d59d4b8642540746430ccb7da9R21)

---
#### Consider a project composed of 2 Scala sources and one auxiliary file:
- `Provider.scala`: This files defines a macro. The implementation of this macro uses an external file `util.txt`.

``` scala
object Provider {
  def macroWithAuxFile: String = macro macroWithAuxFileImpl
  def macroWithAuxFileImpl(c: Context): c.Tree = {
    import c.universe._, compat._

    val file = new java.io.File("util/util.txt")
    val lines = scala.io.Source.fromFile(file).mkString
    val tree = q"$lines"
  }
}
```

``````
 - `Client.scala` which uses the macro `Provider.hello`.
```scala
object Client {
  println(Provider.macroWithAuxFile) // Will expand to println("Hello")
}
``````
- `util.txt`, a simple text file

```
Hello
```

This example features a macro that will, during expansion, open the file **util.txt** and expand to the content of the file. Changing the content of the file should thus invalidate all expansions of this macro.
#### The auxiliary file `util.txt` gets modified

Imagine now that the user modifies the content of `util.txt`:

```
Bonjour
```
#### What gets recompiled without this PR
- Nothing
#### What should be recompiled
- `Client.scala`, because the expansion of the macro is outdated (it should now be `println("Bonjour")`)
#### What gets recompiled with this PR + a few lines to tell sbt about the auxiliary file
- `Client.scala`, because it uses an auxiliary file that has been modified.
#### What's the underlying problem?

During the expansion of `Provider.macroWithAuxFile`, the file `util.txt` is opened and read. Its content becomes the expansion of the macro. However, sbt cannot "guess" that `Client.scala` used this auxiliary file. We propose to add an attachment to the expansion of the macro to tell sbt what auxiliary files have been used during expansion. sbt would be able to inspect this attachment and register the dependencies between macro clients and auxiliary files.
#### How is this problem fixed ?

Macro authors would have to add a few lines of code to their macros, as shown in the following example:

``` scala
object Provider {
  def macroWithAuxFile: String = macro macroWithAuxFileImpl
  def macroWithAuxFileImpl(c: Context): c.Tree = {
    import c.universe._, compat._

    val file = new java.io.File("util/util.txt")
    val lines = scala.io.Source.fromFile(file).mkString
    val map = new java.util.HashMap[String, Any] 
    map.put("touchedFiles", file :: Nil)
    val tree = q"$lines"
    tree.updateAttachment(map)
  }
}
```

(No changes are required to the macro clients).

During the dependency extraction phase, sbt would extract the list of `touched files` from trees that result from the expansion of a macro that use auxiliary files. sbt would therefore be able to understand that a specific tree depends on an auxiliary file.

Once the dependency relationship is established, the invalidation logic is very simple: Whenever an auxiliary file is modified, all the files that depend on it get invalidated.
#### Passing stamps for auxiliary files from a previous Analysis to the next

If I remember correctly, sbt uses `Stamp`s to know whether a file has been modified since the last time it has been compiled. Whenever the user issues a `compile` command in the sbt console, sbt puts aside its old collection of `Stamp`s and constructs a new **empty** instance of `Stamp`s. sbt will take a look at all the Scala sources, compute the new `Stamp` for each of these sources. Finally, sbt will compare the old `Stamp`s with the new ones it just computed. Whenever two `Stamp`s are different, sbt will recompile the file that has been modified. If a file has been deleted, it won't have a new `Stamp` and sbt will know that this file has been deleted. (This is also true for binaries and products)

The key idea here is that sbt is able to do this, because it knows in advance **where to look to find the sources** (in `src/` or whatever has been configured). However, sbt cannot do this for auxiliary files, because there's no standard place to put them. sbt will not be able to perform its active scanning for auxiliary files, which means that the new instance of `Stamps` won't contain any `Stamp` for auxiliary file! Sbt would then consider that these auxiliary files have been deleted, and recompile the macro clients that used them.

The solution to this problem is to pass, from the previous instance of `Stamps` to the new and empty one, all the `Stamp` that were associated to auxiliary files.


typesafe-tools:

jsuereth:
hmm, this is a really interesting idea.  May need to chat some more, but additionally, we may want to merge this when we canbreak binary compatibility.

Duhemm:
Thank you for your comments @jsuereth ! Do you think that introducing overloads could fix this problem, or would you prefer to leave this pull request for later ?

xeno-by:
Hi guys! What do you think about this one? Looks useful to me.

xeno-by:
Ping

sbt/src/sbt-test/source-dependencies/macro-auxiliary-files/test
```diff
@@ -0,0 +1,10 @@
+> compile
+
+# Modify the auxiliary file
+$ copy-file changes/util.txt util/util.txt
+
+$ copy-file changes/Bar.scala macro-provider/src/main/scala/Bar.scala
+
+> macro-client/compile
+
+> macro-client/checkRecompiled Client.scala
```
gkossakowski, line 10: Where `checkRecompiled` is defined?

Duhemm, line 10: Hello Greg! It's defined here: https://github.com/sbt/sbt/pull/1757/files#diff-2097d5b5a7f0292a54187e2fb573c989R2, in  `sbt/src/sbt-test/source-dependencies/macro-auxiliary-files/macro-client/build.sbt`

gkossakowski:
Could you guys give pointers to macros in the wild that need this kind of functionality?

Also, I don't follow this:

> Since we do not know where auxiliary files are in general, there's no directory that we can scan and get them all. We only see them if the compiled file advertises them. Therefore, we may not see them at all for a while, even if they have not be deleted. We need to pass them from a previous analysis to the next, because otherwise sbt would consider that they have been deleted.

Can you give a specific example where this happens? I thought that every time you recompile the file, you invoke macros and get their auxiliary dependencies collected.

Is using bare tree attachments a long-term plan or are you planning to have a dedicated API for macros authors?

Duhemm:
> I don't follow this: (...)

What I meant is that in the case of traditional source files, sbt **actively** looks for source files to know whether they have been changed, because it knows in advance where these files will be located.

However with auxiliary files, sbt can only passively wait for the macros to tell it what auxiliary files they use, because they may be anywhere on disk. Assume that a macro provider `A.scala` uses an auxiliary file `util.txt` to perform its expansion. If, during a certain compilation run, no macro clients of `A.scala` need to be recompiled, sbt won't hear at all from `util.txt`. Therefore it will consider that this file no longer exists. The problem occurs when we don't invoke the macros.

If I remember correctly, @xeno-by mentioned a future resources API for scala.meta, maybe he can elaborate?

xeno-by:
@gkossakowski Virtually all type providers would benefit from this.

Yes, in scala.meta we have a dedicated API for macro authors to access resource files of the project. Calls to this API will be intercepted and then communicated to SBT by the macro engine, completely transparently for the users.

sbt/src/sbt-test/source-dependencies/macro-auxiliary-files/test
gkossakowski, line 10: Ok. Thanks.

gkossakowski:
@xeno-by: again, do you guys plan to bring this over to reflection api? If this is a common need it seems it should be part of API macro authors work with.

@Duhemm: I still don't follow what you said. Could you rewrite your comment in a format we used in #1836? That style made everything more clear to me.

Duhemm:
@gkossakowski I've updated the description of the PR.

xeno-by:
@gkossakowski My understanding is that we can't change the reflection API in both 2.11 and 2.12, which makes my desire irrelevant :)

dwijnand:

PR #1748, opened 2014-12-03
excludeDependencies: excludeAll at the project level

eed3si9n:
``` scala
excludeDependencies += "org.apache.logging.log4j"
excludeDependencies += "com.example" %% "foo"
```

In the first example, all artifacts from the organization `"org.apache.logging.log4j"` are excluded from the managed dependency.
In the second example, artifacts with the organization `"com.example"` and the name `"foo"` cross versioned to the current `scalaVersion` are excluded.


eed3si9n:
Attached a PR on it. /review @jsuereth

ivy/src/main/scala/sbt/CrossVersion.scala
```diff
@@ -78,6 +78,14 @@ object CrossVersion {
   def apply(module: ModuleID, is: Option[IvyScala]): Option[String => String] =
     is flatMap { i => apply(module, i) }
 
+  /** Constructs the cross-version function defined by `exclude` and `is`, if one is configured. */
+  private[sbt] def apply(exclude: SbtExclusionRule, is: IvyScala): Option[String => String] =
```
jsuereth: this is, most definitely, the worst api we have.

jsuereth: can we not propogate `Option[String=>String]`?

eed3si9n: I was somewhat torn between being consistent with the `ModuleID` logic and picking horrible `apply` naming. In the end, I went for consistency. I can get rid of `apply` here since it's only used internally for `substituteCross`.

main/actions/src/main/scala/sbt/CacheIvy.scala
```diff
@@ -130,18 +130,22 @@ object CacheIvy {
...
   implicit def inlineIvyIC: InputCache[InlineIvyConfiguration] = wrapIn
   implicit def moduleSettingsIC: InputCache[ModuleSettings] =
-    unionInputCache[ModuleSettings, PomConfiguration :+: InlineConfiguration :+: EmptyConfiguration :+: IvyFileConfiguration :+: HNil]
+    unionInputCache[ModuleSettings, PomConfiguration :+: InlineConfiguration :+: InlineConfigurationWithExcludes :+: EmptyConfiguration :+: IvyFileConfiguration :+: HNil]
 
   implicit def ivyConfigurationIC: InputCache[IvyConfiguration] =
     unionInputCache[IvyConfiguration, InlineIvyConfiguration :+: ExternalIvyConfiguration :+: HNil]
 
   object L4 {
+    implicit val inlineWithExcludesToHL = (c: InlineConfigurationWithExcludes) =>
```
jsuereth, line 139: oh man...... my favorite part of our code :)

eed3si9n, line 139: It's artisanal New England style pickling.

ivy/src/main/scala/sbt/CrossVersion.scala
jsuereth: yeah... i'd instead go the other way.  Let's see how much of that API we can clean up internally and instead unify on something that's easier to grok when you read it.

Basically, i'd rather go consistent when it's not a choice between confusion and understanding. In that case, let's start moving the confusing code into something less confusing.

jsuereth:
Feature looks great.   Sad we had to make a new inline configuration, but that's the life of binary compatibility.

eed3si9n:
Trying to keep bincompat in the face of case classes and default arguments was literally the hardest part of this PR. I really should make pseudo case class generator.

jsuereth:
ok, lgtm, but needs to be rebase on your other ivy fix.

eed3si9n:
Rebased and squashed.

sslavic:
Will this feature be reflected in generated pom? Either way, would be nice to have that fact documented. Since pom 4.0 model doesn't directly support project level exclusion, I guess every declared dependency in pom would have to have exclusions section added.

Maven guys are against project level exclusions as can be read on http://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html :

"This is mainly done to be sure the dependency graph is predictable, and to keep inheritance effects from excluding a dependency that should not be excluded. If you get to the method of last resort and have to put in an exclusion, you should be absolutely certain which of your dependencies is bringing in that unwanted transitive dependency."

eed3si9n:
This was an oversight on my part. I'd probably need to translate it to dependency-level exclusion for poms. Identical exclusion rules (for example for log libraries) added to all dependencies can add unnecessary resolution time for large dependency graph, so I think it makes sense to support this concept.

PR #1759, opened 2014-12-03
Minor code cleanup

jedesah:
I was inspecting the codebase to figure out something when I saw a few IntelliJ suggestions. Eventually, I ran the inspection tool and went through the enormous list of possible issues. I think these are changes for the better. Please let me know, if you do not agree with some of the types of changes, I can modify the PR to make everyone happy.

### DIFF
--- a/interface/src/main/java/xsbti/api/Modifiers.java
+++ b/interface/src/main/java/xsbti/api/Modifiers.java
@@ -10,7 +10,7 @@ public final class Modifiers implements java.io.Serializable
 	private static final int LazyBit = 5;
 	private static final int MacroBit = 6;
 
-	private static final int flag(boolean set, int bit)
+	private static int flag(boolean set, int bit)
 	{
 		return set ? (1 << bit) : 0;
 	}
@@ -30,7 +30,7 @@ public final class Modifiers implements java.io.Serializable
 
 	private final byte flags;
 
-	private final boolean flag(int bit)
+	private boolean flag(int bit)
 	{
 		return (flags & (1 << bit)) != 0;
 	}
--- a/project/Util.scala
+++ b/project/Util.scala
@@ -124,12 +124,12 @@ object Util {
   def cleanPom(pomNode: scala.xml.Node) =
     {
       import scala.xml._
-      def cleanNodes(nodes: Seq[Node]): Seq[Node] = nodes flatMap (_ match {
-        case elem @ Elem(prefix, "dependency", attributes, scope, children @ _*) if excludePomDependency(elem) =>
+      def cleanNodes(nodes: Seq[Node]): Seq[Node] = nodes flatMap ({
+        case elem@Elem(prefix, "dependency", attributes, scope, children@_*) if excludePomDependency(elem) =>
           NodeSeq.Empty
-        case Elem(prefix, "classifier", attributes, scope, children @ _*) =>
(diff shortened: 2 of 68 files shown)


typesafe-tools:

jsuereth:
thanks for doing this.  I'll try to look through them all, but it may be a while given the extent of files touched.

interface/src/main/java/xsbti/api/Modifiers.java
```diff
@@ -10,7 +10,7 @@
 	private static final int LazyBit = 5;
 	private static final int MacroBit = 6;
 
-	private static final int flag(boolean set, int bit)
```
jsuereth, line 13: what was the reasoning here?

project/Util.scala
```diff
@@ -124,12 +124,12 @@ object Util {
   def cleanPom(pomNode: scala.xml.Node) =
     {
       import scala.xml._
-      def cleanNodes(nodes: Seq[Node]): Seq[Node] = nodes flatMap (_ match {
-        case elem @ Elem(prefix, "dependency", attributes, scope, children @ _*) if excludePomDependency(elem) =>
+      def cleanNodes(nodes: Seq[Node]): Seq[Node] = nodes flatMap ({
+        case elem@Elem(prefix, "dependency", attributes, scope, children@_*) if excludePomDependency(elem) =>
```
jsuereth: This one may be intellij fighting with our scalariform style....

jsuereth:
The nonEmpty changes are fine.  One question on why the change, and then for pattern matching with binding (`@`) I think we should leave it as it was.

interface/src/main/java/xsbti/api/Modifiers.java
jedesah, line 13: Well I just followed IntelliJ's hint saying it's not possible already to override a private field which sort of makes sense. Basically it's saying that final is redundant.Maybe the author was trying to signal that it should be not be overridden even if it one day becomes protected or private. Another explanation could be some bizarre optimization rule I am unaware of.

project/Util.scala
jedesah: So I think the `_ match` should be removed but Indeed IntelliJ modified the style below under my nose without me realizing it.

eed3si9n:
@jsuereth What's the status on this?

interface/src/main/java/xsbti/api/Modifiers.java
jsuereth, line 13: I'd lean on the side of not touching anything in xsbti right now.   This code is used in unexpected ways (intellij may not be aware) and it DOES do dirty things.....

jsuereth:
@eed3si9n My bad.  Responded to one comment.  Basically I'm terrified of mucking in xsbti, but the tests ARE passing, so it's your call if we're good to go.

interface/src/main/java/xsbti/api/Modifiers.java
eed3si9n, line 13: In Java `final` method means that you can't [override or hide](http://www.coderanch.com/how-to/java/OverridingVsHiding) it in the subclass. That concern is moot because `Modifier` is a final class. In addition, `flag` method being a private static class means it's used for its own implementation, but it's not externally observable. I'd say this is ok as is.

eed3si9n:
@jedesah Merged. Thanks for your contribution!

PR #1752, opened 2014-12-04
Cached resolution is too verbose

eed3si9n:
The cached resolution is really great, but it's way too verbose. I'm not including the output here, I can do that if needed.

It would be great, if by default the output would be comparable to the non-cached resolution (in terms of logged lines).

Is there a workaround, can I set the logLevel for the resolution to Warn? I already tried "logLevel in update := Level.Warn", but this didn't do the trick (while "logLevel := Level.Warn" worked, which is not what I want).


eed3si9n:
Attached a PR on this issue. /review @jsuereth 
- Fixes cached resolution being too verbose
- Adds new `UpdateLogging` named `Default`
- When global `logLevel` or `logLevel in update` is `Debug`, `Default` will bump up to `Full` `UpdateLogging`.

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -297,6 +297,27 @@ private[sbt] trait CachedResolutionResolveEngine extends ResolveEngine {
...
+      val ivy = makeInstance
+      ivy.pushContext()
+      ivy.getLoggerEngine.pushLogger(log)
+      try { f(ivy) }
+      finally {
+        ivy.getLoggerEngine.popLogger()
+        ivy.popContext()
+      }
+    }
+  def withDefaultLogger[A](log: MessageLogger)(f: => A): A =
```
jsuereth, line 313: nice...

jsuereth:
LGTM

PR #1671, opened 2014-10-13
Improve xml handling

ajozwik:
Fixed #1666. Previous parser does not work for multi line statements with `//`

PR #1765, opened 2014-12-05
Rollback xml parsing work-around

### LINKED ISSUE #1738: Parentheses are added to XML even inside String literals (0.13.7)
```
TaskKey[Unit]("crashlytics-gen") <<= (collectResources in Android) map {
  case (_,res) =>
  val resource = """<string name="com.crashlytics.android.build_id">%s</string>""" format java.util.UUID.randomUUID.toString
  val lines = """<?xml version="1.0" encoding="utf-8"?>""" ::
    "<resources>" :: resource :: "</resources>" :: Nil
  println("wtf: " + lines.mkString(""))
  lines foreach println
}
```

Produces this:

```
> crashlytics-gen
wtf: <?xml version="1.0" encoding="utf-8"?>(<resources>(<string name="com.crashlytics.android.build_id">f7a7b0d5-9ac9-4c1c-97c7-2fb852fa5bfc</string>)</resources>)
<?xml version="1.0" encoding="utf-8"?>
(<resources>
(<string name="com.crashlytics.android.build_id">f7a7b0d5-9ac9-4c1c-97c7-2fb852fa5bfc</string>)
</resources>)
```

ajozwik:
Fixes #1738
https://github.com/scala/scala/issues/4186 solves bug in scala compiler.

### DIFF
--- a/main/src/test/scala/sbt/internals/parser/ErrorSpec.scala
+++ b/main/src/test/scala/sbt/internals/parser/ErrorSpec.scala
@@ -39,6 +39,21 @@ class ErrorSpec extends AbstractSpec with ScalaCheck {
         """.stripMargin
       MissingBracketHandler.findMissingText(buildSbt, buildSbt.length, 2, "fake.txt", new MessageOnlyException("fake")) must throwA[MessageOnlyException]
     }
+
+    "handle xml error " in {
+      val buildSbt =
+        """
+          |val a = <a/><b/>
+          |val s = '
+        """.stripMargin
+      SbtParser(SbtParser.FAKE_FILE, buildSbt.lines.toSeq) must throwA[MessageOnlyException].like {
+        case exp =>
+          val message = exp.getMessage
+          println(s"${exp.getMessage}")
+          message must contain(SbtParser.FAKE_FILE.getName)
+      }
+    }
+
   }
 
   private def containsLineNumber(buildSbt: String) = {
--- a/main/src/main/scala/sbt/EvaluateConfigurations.scala
+++ b/main/src/main/scala/sbt/EvaluateConfigurations.scala
@@ -93,10 +93,9 @@ object EvaluateConfigurations {
    *
    * @param eval    The evaluating scala compiler instance we use to handle evaluating scala configuration.
    * @param file    The file we've parsed
-   * @param lines   The lines of the configurtion we'd like to evaluate.
    * @param imports The default imports to use in this .sbt configuration.
    *
-   * @return A function which can take an sbt classloader and return the raw types/configuratoin
+   * @return A function which can take an sbt classloader and return the raw types/configuration
    *         which was compiled/parsed for the given file.
    */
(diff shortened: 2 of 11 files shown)


typesafe-tools:

eed3si9n:
Given SI-9027, I think it's better to roll back #1671 and ask the user to use semicolon or `xml:group` rather than rolling out our own lexers.

```
scala> val xs: xml.NodeSeq = <xml:group><x/><y/></xml:group>
xs: scala.xml.NodeSeq = <x/><y/>
```

ajozwik:
``` scala
scala> :pa
// Entering paste mode (ctrl-D to finish)

val g = <xml:group><a/><b/></xml:group>
val b = <a/><b/>;
val c = (<a/><b/>)
val a = <a/><b/>


// Exiting paste mode, now interpreting.

g: scala.xml.Group = <a/><b/>
b: scala.xml.NodeBuffer = ArrayBuffer(<a/>, <b/>)
c: scala.xml.NodeBuffer = ArrayBuffer(<a/>, <b/>)
a: scala.xml.NodeBuffer = ArrayBuffer(<a/>, <b/>)


scala> a == b
res4: Boolean = true

scala> a == c
res5: Boolean = true

scala> a == g
res6: Boolean = false
```

`Group` is not the same as `NodeBuffer`.

Second problem is that how to detect xml compilation problem?

See

``` scala
scala> :pa
// Entering paste mode (ctrl-D to finish)

val b = <a/><b/>
val s = 3

// Exiting paste mode, now interpreting.

<console>:2: error: ';' expected but 'val' found.
val s = 3
^
```

So the problem will be how to detect, that compilation problem is xml group compilation error?

ajozwik:
@jsuereth @eed3si9n - we need use 2.10.4 in current sbt by at least 4 months. See https://github.com/scala/scala/pull/4186#issuecomment-66162506

eed3si9n:
My preference is to tell user to put semicolon or parens when they see  `error: ';' expected but 'val' found.` rather than trying to implement our own lexer. As #1738 has shown, workarounds can cause more issues. Even though it is several months away, a definitive fix is coming up anyway.

As per `NodeBuffer` vs `Group`, if they behave the same semantically as `NodeSeq`, I am ok with them not being equal, but just to be safe we can recommend semicolon or parens instead of `xml:group`.

ajozwik:
OK, so I have to change some tests as well.

main/src/test/scala/sbt/internals/parser/ErrorSpec.scala
```diff
@@ -39,6 +39,21 @@ class ErrorSpec extends AbstractSpec with ScalaCheck {
         """.stripMargin
       MissingBracketHandler.findMissingText(buildSbt, buildSbt.length, 2, "fake.txt", new MessageOnlyException("fake")) must throwA[MessageOnlyException]
     }
+
+    "handle xml error " in {
+      val buildSbt =
+        """
+          |val a = <a/><b/>
+          |val s = '
```
ajozwik, line 47: This test will pass even when scala parser will be corrected. The second statement `val s = '` is incorrect

PR #1620, opened 2014-12-08
sbt resolves dependencies every compile when using %% with dependencyOverrides

eed3si9n:
The issue is not caused when % is used instead of %%.

Following is reproducible build.sbt.
https://gist.github.com/tkawachi/125cdcad0b0e014f1f44

ref. https://groups.google.com/d/msg/sbt-dev/cUOCbLVXwEQ/O9lzi0atI_kJ

### DIFF
--- a/ivy/src/main/scala/sbt/IvyConfigurations.scala
+++ b/ivy/src/main/scala/sbt/IvyConfigurations.scala
@@ -34,12 +34,42 @@ final class InlineIvyConfiguration(val paths: IvyPaths, val resolvers: Seq[Resol
     checksums: Seq[String], resolutionCacheDir: Option[File], log: Logger) =
     this(paths, resolvers, otherResolvers, moduleConfigurations, localOnly, lock, checksums, resolutionCacheDir, UpdateOptions(), log)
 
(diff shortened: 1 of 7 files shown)


eed3si9n:
Attached PR on this. /review @jsuereth

ivy/src/main/scala/sbt/IvyConfigurations.scala
```diff
@@ -108,18 +138,18 @@ object InlineConfiguration {
...
+    val configurations: Seq[Configuration],
+    val defaultConfiguration: Option[Configuration],
+    val ivyScala: Option[IvyScala],
+    val validate: Boolean,
+    val conflictManager: ConflictManager) extends ModuleSettings {
   def withConfigurations(configurations: Seq[Configuration]) = copy(configurations = configurations)
   def noScala = copy(ivyScala = None)
 
-  def copy(module: ModuleID = this.module,
+  private[sbt] def copy(module: ModuleID = this.module,
```
jsuereth, line 152: nice Binary compat trick here :)

jsuereth:
LGTM

PR #1771, opened 2014-12-11
Fix logger not overwriting the previous line in alternate shells

dwickern:
When running the 'update' task in bash, the output is all collapsed onto one line.
On Windows, even using an ANSI capable shell, running 'update' spams the console.

Tested with dash in Ubuntu; ANSICON, Console2 and ConsoleZ in Windows.

### DIFF
--- a/util/log/src/main/scala/sbt/ConsoleOut.scala
+++ b/util/log/src/main/scala/sbt/ConsoleOut.scala
@@ -16,7 +16,7 @@ object ConsoleOut {
     cur.contains(s) && prev.contains(s)
 
   /** Move to beginning of previous line and clear the line. */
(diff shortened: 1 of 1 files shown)


typesafe-tools:

eed3si9n:
LGTM. @jsuereth?

eed3si9n:
[Terminals FAQ](http://www.vt100.net/terminals_faq) says:

> "ANSI" either refers to:
> - the standard ANSI X3.64, which defines a structure for defining escape sequences for terminals and printers, or
> - ANSI.SYS, a display driver found on PCs running DOS.

Here's [ANSI.SYS](http://msdn.microsoft.com/en-us/library/cc722862.aspx)

PR #1766, opened 2014-12-13
Lots of "No main class detected" warnings since 0.13.7

eed3si9n:
In 0.13.7 a new feature was added - a check that a project has a main class. However, not every project needs one: for example, our build consists of a few dozens of projects, only several of which are runnable. Consequently SBT prints

```
[warn] No main class detected
```

for _every_ project in the build which does not have a main class (and there is a majority of them).

Discussed [here](https://groups.google.com/forum/#!topic/sbt-dev/-5zD6fbHcHc)


eed3si9n:
This warning was added in #1669 /cc @kretes. I should be able to just remove the warning without losing the feature he added.

eed3si9n:
Attached PR. /review @jsuereth

PR #1773, opened 2014-12-14
Fixes #1223. Throttle force GC

### LINKED ISSUE #1223: Reducing PermGen usage in sbt

eed3si9n:
This is take 2 of the force GC. This will run GC on a set interval or before cross building.

@cunei @jsuereth Do you have some test case that I can use to see if this fixes #1223 and that it won't blow up with `~compile`?


main/src/main/scala/sbt/EvaluateTask.scala
```diff
@@ -322,12 +346,20 @@ object EvaluateTask {
...
-          // Force finalizers to run.
-          System.runFinalization()
-          // Force actually cleaning the weak hash maps.
-          System.gc()
+          try {
+            val opfc = memoryMxBean.getObjectPendingFinalizationCount
+            if (opfc > config.maxObjectPendingFinalization) {
+              log.info(s"Forcing garbage collection... Objects pending finalization: $opfc")
+              // Force the detection of finalizers for scala.reflect weakhashsets
+              System.gc()
```
jsuereth: You may need to run this before the check willbe true...

eed3si9n:
Not sure why I close this, but the issue #1223 is still alive and kicking.
Maybe I need some throttling as mentioned in #1329.

jsuereth:
besides my one minor comment, I think this LGTM now.

jsuereth:
Also, please rebase.

dwijnand:
Why not use `scala.concurrent.duration.Duration`/`FiniteDuration` instead of `Int`, which needs to be documented as "(in seconds)" everywhere?

Among other things it would also match your notes: `(default: 60s)`.

main/src/main/scala/sbt/GCUtil.scala
```diff
@@ -0,0 +1,36 @@
...
+  // Returns the default force garbage collection flag,
+  // as specified by system properties.
+  val defaultForceGarbageCollection: Boolean = true
+  val defaultMinForcegcInterval: Int = 60
+  val lastGcCheck: AtomicLong = new AtomicLong(0L)
+
+  def forceGcWithInterval(minForcegcInterval: Int, log: Logger): Unit =
+    {
+      val now = System.currentTimeMillis
+      val last = lastGcCheck.get
```
jsuereth: lazySet for performance, YO

main/src/main/scala/sbt/GCUtil.scala
```diff
@@ -0,0 +1,36 @@
...
+      // This throttles System.gc calls to interval
+      if (now - last > minForcegcInterval * 1000) {
+        lastGcCheck.set(now)
```
jsuereth: lazySet for performance, YO

dwijnand:
Nice! Thank you.

eed3si9n:
I'm seeing

```
[error] (sbtRoot/*:scripted) Failed tests:
[error]     source-dependencies / macro-arg-dep-2-11
[error]     source-dependencies / macro-arg-dep-nested-2-11
[error] Total time: 723 s, completed May 6, 2015 1:30:32 AM
```

on Travis, but I can't reproduce it locally, so I'm going to merge this.

Issue #1729, opened 2014-11-13
Status of support for macros in incremental compiler


Duhemm:
This is a meta issue that gives an overview of the current state of the support for macro-enabled programs in the incremental compiler.
### Support for macros
- [x] Record dependencies on macro arguments (#1163)
- [x] Disable macro heuristic for inherited macros (#1142, #1202)
- [ ] Hooking Macrotracker in sbt (#1778, [discussion](https://groups.google.com/d/msg/sbt-dev/clIbzCp9VX8/z4IiYEZwoNEJ), [plugin](https://github.com/scalamacros/macrotracker))
- [ ] Recompile macro clients if any transitive dependency of the macro is updated ( #1836, related issue : #399, [discussion](https://groups.google.com/forum/#!topic/sbt-dev/Ub7asIeMVJA))
- [ ] Allow macro clients to depend on auxiliary (non-scala) files (#1757, [discussion](https://groups.google.com/d/msg/sbt-dev/f7fKIQSIx20/BcWJaBsY2VkJ))
### Falicities that help to support macros

These facilities make it easier to extend the kind of dependency relations that sbt knows of, and thus make it easier to implement better support for macros (or anything else that requires adding a new dependency kind).
- [x] Abstract over dependency kind in Analysis and Relations (#1340)
- [x] Don't hardcode existing relations in TextAnalysisFormats (#1572)
- [x] Abstract over dependency kind in Compile (#1736)
#### Some more words on...
- [Macrotracker](https://github.com/scalamacros/macrotracker) is a compiler plugin that registers calls to the reflection API and attaches the list of inspected ("touched") symbols to the expanded macros. These symbols can then be extracted by sbt and registered as dependencies of the macro client.
- I have a prototype of a solution for the problem of transitive dependencies of macros. It works great when the macro provider has no external dependencies. I have a solution to this problem, but I'm afraid this would hurt performance.
- Macro providers can add an attachment to the expanded macro to tell sbt that the macro client depends on an auxiliary file by giving its path. An implementation of this proposition can be found [here](https://github.com/Duhemm/sbt/tree/deps-auxiliary-files). [Example of project that uses this feature](https://github.com/Duhemm/macro-external-files)
#### Collecting the dependencies of a macro

Consider the following macro : 

``` scala
object Provider {
  def testMacro(arg: String): String = macro testMacroImpl
  def testMacroImpl(c: Context)(arg: c.Tree): c.Tree = {
    import c.universe._
    val foo = Foo.bar
[...]

PR #1778, opened 2014-12-15
Macrotracker support

Duhemm:
Hi !

> MacroTracker is a compiler plugin for scalac 2.11.0. Its job is to register all the symbols that are touched during the expansion of a macro, and attach them to the expanded macro. Using this information, we can deduct the dependencies between the expanded macro and everything that has been involved in its expansion. 

(From [this thread in sbt-dev](https://groups.google.com/forum/#!topic/sbt-dev/clIbzCp9VX8))

MacroTracker helps to track the dependencies that are introduced using the reflection API. Whenever a symbol is accessed using the reflection API, MacroTracker registers it and finally attaches all the touched symbols to the expanded macro.

It can help greatly in cases such as the ones that are exposed in the [accompanied scripted test](https://github.com/Duhemm/sbt/tree/macrotracker-support/sbt/src/sbt-test/source-dependencies/macro-macrotracker).

MacroTracker can be found here : [scalamacros/macrotracker](https://github.com/scalamacros/macrotracker).

### DIFF
--- a/compile/inc/src/main/scala/sbt/inc/IncrementalNameHashing.scala
+++ b/compile/inc/src/main/scala/sbt/inc/IncrementalNameHashing.scala
@@ -45,6 +45,12 @@ private final class IncrementalNameHashing(log: Logger, options: IncOptions) ext
     val transitiveInheritance = byExternalInheritance flatMap { file =>
       invalidateByInheritance(relations, file)
     }
+    val externalMacroExpansionR = relations.macroExpansion.external
+    val byExternalMacroExpansion = externalMacroExpansionR.reverse(modified)
+    log.debug(s"Files invalidated by macro expansion from (external) $modified: $byExternalMacroExpansion; now invalidating by macro expansion (internally).")
+    val transitiveMacroExpansion = byExternalMacroExpansion flatMap { file =>
+      invalidateByMacroExpansion(relations, file)
+    }
     val memberRefInvalidationInternal = memberRefInvalidator.get(relations.memberRef.internal,
       relations.names, externalAPIChange)
     val memberRefInvalidationExternal = memberRefInvalidator.get(relations.memberRef.external,
@@ -57,7 +63,7 @@ private final class IncrementalNameHashing(log: Logger, options: IncOptions) ext
     // This includes non-inheritance dependencies and is not transitive.
     log.debug(s"Getting sources that directly depend on (external) $modified.")
     val memberRefB = memberRefInvalidationExternal(modified)
-    transitiveInheritance ++ memberRefA ++ memberRefB
+    transitiveInheritance ++ transitiveMacroExpansion ++ memberRefA ++ memberRefB
   }
 
   private def invalidateByInheritance(relations: Relations, modified: File): Set[File] = {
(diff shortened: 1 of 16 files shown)


typesafe-tools:

xeno-by:
Looks like the test failure is spurious. Would it be possible to rerun the test suite? 

Also, wdyt about this patch? From what I can see, it allows achieving much better precision than what was possible before + brings us closer to closing https://github.com/sbt/sbt/issues/1729.

eed3si9n:
I'm rerunning the tests.

eed3si9n:
@jsuereth and I were both busy during December with Typesafe related things, but hopefully we can review open PRs including this one after we come back from the break next week.

xeno-by:
Hello guys! Any news on this one?

gkossakowski:
I have looked at this PR from very high-level point of view. It's great to see that work done in #1340 is paying off: major feature is being added without too many changes to incremental compiler sources.

I have a few concerns:
- this PR introduces a dependency on a compiler plugin that is of SNAPSHOT version. SNAPSHOTs are not guaranteed to stay around but sbt tests should run for a long time. We'll need a stable release at the very least.
- touched symbols are stored as tree attachment at the moment. Is this how you guys this envision long term?
- we run tree traverser that looks for attached symbols on all tree nodes in every file, no matter if there's any macro expansion or not; what kind of performance hit are we expecting?

Also, do you see your compiler plugin merged back into Scala 2.11/2.12? It would really help to understand long-term plans around this feature.

xeno-by:
1) I can push a stable release if necessary.

2) I think so, yes. This is domain-specific metadata (much like e.g. `MacroExpansionAttachment`), so I don't have other ideas how to associate it with trees except using attachments. Do you see any problems that might result from this?

3) I didn't give any thought to backporting this functionality into scala/scala. If you find it useful and if the Scala team finds it reasonable, I can submit a pull request that embeds tracking of touchedSymbols into Scala 2.11. /cc @retronym

retronym:
I would favour a minmial change to SBT that enabled the meat of this PR to be factored out as a SBT plugin. But I fear that this will prove a difficult task.

Failing that, we need another way to be sure that the new code is only run when users opt-in to the macro tracking compiler plugin. Perhaps you can check whether that plugin is enabled. We don't want other users exposed to performance penalties of extra traversals, nor to potential bugs in this code.

I took a look at the internals of macro-tracker. It looks to be rather difficult to maintain. How can we reason that all overrides are provided? I noticed that the inherent difficulty, and the solution, runtime reflection synchronization. I suppose the macro tracker has a lower bar, we shouldn't reject a solution that 99% of dependencies just because we can't prove it tracks 100%. But I would love to hear ideas for how we could implement this more cleanly if it were done in the compiler proper.

gkossakowski:
Making incremental fully extensible might be rather challenging. It might be more work than providing support for a few special features.
I think structuring the code so this logic is clearly separated and making the whole feature opt-in is a good approach.
I'd suggest to add logging of timings so users can easily understand the performance hit they are opting-in.

xeno-by:
@retronym The only approach that I can think of involves macros, which might or might not be desireable in the compiler (like the experience with runtime reflection has shown).

`Perhaps you can check whether that plugin is enabled`
I find this be a really great idea!

`performance penalties of extra traversals`
Can we fuse this traversal into the other ones that are already performed?

`if it were done in the compiler proper`
Does this mean that you'd be interested in having something like macrotracker in scala/scala?

retronym:
> Does this mean that you'd be interested in having something like macrotracker in scala/scala?

Take it as meaning: we could start talking about that if implementation in the compiler is drastically simpler than the implementation as a plugin :)

xeno-by:
@retronym I could write a macro that generates a proxy that mechanically calls wrap/unwrap and lets compiler developers to provide custom implementations of wrap/unwrap for the types they are interested in. Would that qualify as being drastically simpler?

retronym:
The macro itself will still take some work to understand. And we don't have great tools for debugging code involving macros, so I'd be a little wary about that.

But it does seem like a worthwhile thing to try if you have the time. I guess the refactoring to use a macro rather than boilerplate could be done in the compiler plugin as in the core compiler, though.

dwijnand:

PR #1779, opened 2014-12-15
Remove javac options from doc tool.

jedesah:
The java documentation tool does not take all the options that the javac tool accepts so I believe it makes sense to have two separate keys for them in sbt

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -729,7 +729,7 @@ object Defaults extends BuildCommon {
       val srcs = sources.value
       val out = target.value
       val sOpts = scalacOptions.value
-      val jOpts = javacOptions.value
+      val jOpts = javaDocOptions.value
       val xapis = apiMappings.value
       val hasScala = srcs.exists(_.name.endsWith(".scala"))
       val hasJava = srcs.exists(_.name.endsWith(".java"))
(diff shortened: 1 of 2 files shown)


typesafe-tools:

jsuereth:
Just a note, this doesn't actually create javadoc options that are fed into the build, hence the failures.

Additionally, right now `javacArgs in (Compile, doc)` is how you reference javadoc only arguments.

I agree with you that not all are the same, and only some are the same.   sbt's _attempt_ here is that those which are shared go in "javacAgs", those which are note get relegated to "javacArgs in (Compile,compile)" or "javacArgs in (Compile,doc)".

I'd be ok having a different key, but we'd have to tease out the areas in the build where we expect settings to be shared and they would no longer be.  If you need help working this through sbt, I can give you pointers/guidance.

thanks for opening the PR!

jedesah:
@jsuereth Just to be clear, when you say `javacArgs`, do you mean `javacOptions` or do you mean something else I am not finding?

jedesah:
You know, `javacOptions in (Compile, compile)` sort of solves my problem at hand for now and even though it's a little bit weird to me that javacOptions applies to the javadoc tool, if this is a complicated change, I am not sure I have time to figure this stuff out right now.

I think I will go ahead and close this PR for now. Thx for the help diagnosing my problem!

jsuereth:
@jedesah No problem. Can you opena ticket abouta less complicated mechanism for controlling arguments between javac/javadoc? (or at least one with less confusion).

jedesah:
Done!

PR #1787, opened 2014-12-24
specify the "--depth" parameter for efficient when git clone

xuwei-k:


typesafe-tools:

jsuereth:
LGTM . Any chance we can get a test for this?

eed3si9n:
1. +1 on `--depth 1`. Shallow clones should be fine for read-only usage. Also see git/git@82fba2b9d39163a0c9b7a3a2f35964cbc039e1a.
2. As per testing, we have scripted [project/source-plugins](https://github.com/sbt/sbt/tree/0.13/sbt/src/sbt-test/project/source-plugins), but we need to isolate `sbt.global.base` so we have fresh checkouts, if it's not done so already.
3. What's more interesting to me is how we document this undocumented feature.

[Dispatch's documentation](http://dispatch-classic.databinder.net/Project+Setup.html) talks more about source dependency than ours.
The only reference I could find in our current doc on source dependencies is for [plugins](http://www.scala-sbt.org/0.13/docs/Plugins.html#1d%29+Project+dependency):

``` scala
lazy val assemblyPlugin = uri("git://github.com/sbt/sbt-assembly")
```

And as mentioned above scripted test demonstrates the use of source dependency for sbt plugin (using build.scala in meta-meta build; this should be updated to `project/plugin.sbt`).

PR #1793, opened 2015-01-07
sbt-maven-resolver: Use Aether for resolving maven dependencies

jsuereth:
sbt 0.13.8 adds an extension point in the dependency resolution to customize Maven resolvers.
This allows us to write sbt-maven-resolver auto plugin, which internally uses Eclipse Aether
to resolve Maven dependencies instead of Apache Ivy.

To enable this plugin, add the following to `project/maven.sbt` (or `project/plugin.sbt` the file name doesn't matter):

``` scala
libraryDependencies += Defaults.sbtPluginExtra("org.scala-sbt" % "sbt-maven-resolver" % sbtVersion.value,
  sbtBinaryVersion.value, scalaBinaryVersion.value)
```

This will create a new `~/.ivy2/maven-cache` directory, which contains the Aether cache of files.
You may notice some file will be re-downloaded for the new cache layout.
Additionally, sbt will now be able to fully construct
`maven-metadata.xml` files when publishing to remote repositories or when publishing to the local `~/.m2/repository`.
This should help erase many of the deficiencies encountered when using Maven and sbt together.

**Notes and known limitations**:
- sbt-maven-resolver requires sbt 0.13.8 and above.
- The current implementation does not support Ivy-style dynamic revisions, such as "2.10.+" or "latest.snapshot".  This
  is a fixable situation, but the version range query and Ivy -> Maven version range translation code has not been migrated.
## original case

This change adds the usage of Aether to replace Ivy's default maven dependency handling.
- Creates a new type of Ivy dependency resolver which resolves dependencies using Aether
- Wires Aether's artifact download mechanism to use Ivy's download mechanisms
- Adds configuration/adaptation to Aether so it can successfully resolve sbt plugins
  <s>\- Creates a new `UpdateOptions` flag which enables/disables Aether resolution. (Right now defaults to true so we can see if TravisCI passe swith Aether resolution, then we'll add a commit to default to false to ensure we didn't break previous behavior).</s>
- Updates tests which rely on Aether behavior for correctness to default to aether resolution.
## Known bugs fixed:

Fixes #1322
Fixes #321
Fixes #647
Fixes #1616
## Possible bugs fixed (hoping to verify soon):
#1676
#679

Review by @eed3si9n 

NOTE: Please do not merge until this notice is taken down, which implies that Aether resolution is no longer the default (keeping existing behavior and having it opt-in) as well as ensuring that the plugin deduplication concern has a test in MavenResolutionSpec.

PR #1795, opened 2015-01-08
Turn Aether integration into sbt-maven-resolver

eed3si9n:
/review @jsuereth 
PR on top of #1793


ivy/src/main/scala/sbt/MakePom.scala
```diff
@@ -330,7 +331,7 @@ class MakePom(val log: Logger) {
       val repositories = if (includeAll) allResolvers(settings) else resolvers(settings.getDefaultResolver)
       val mavenRepositories =
         repositories.flatMap {
-          case m: org.apache.ivy.plugins.resolver.MavenRemoteRepositoryResolver if m.repo.root != DefaultMavenRepository.root =>
+          case m: CustomRemoteMavenResolver if m.repo.root != DefaultMavenRepository.root =>
```
jsuereth: This will have to support IBiblio too now.

jsuereth:
Looks pretty good to me.  Can you update the notes file for this?

jsuereth:
Oh, also this appears to have broken the unit tests so they don't even run....

jsuereth:
Additionally, check the test failures.  A bunch of scripted tests that used to pass (due ot using aether) now fail because the old, incorrect, behavior is exhibited.

jsuereth:
I think we need to duplicate a few of the tricksy maven tests if we go this route.

E.g. `project/transitive-plugins` needs to be duplicated to check the plugin here, as well as most of the dependency-management tests.

ivy/src/main/scala/sbt/ConvertResolver.scala
```diff
@@ -94,37 +96,34 @@ private[sbt] object ConvertResolver {
...
+    case (r, settings, log) =>
       r match {
         case repo: MavenRepository =>
           {
-            if (isUseAetherForResolution(settings)) {
-              repo match {
-                case cache: MavenCache => new org.apache.ivy.plugins.resolver.MavenCacheRepositoryResolver(cache, settings)
-                case _                 => new org.apache.ivy.plugins.resolver.MavenRemoteRepositoryResolver(repo, settings)
+            val pattern = Collections.singletonList(Resolver.resolvePattern(repo.root, Resolver.mavenStyleBasePattern))
+            final class PluginCapableResolver extends IBiblioResolver with ChecksumFriendlyURLResolver with DescriptorRequired {
```
jsuereth: This needs the CustomMavenRepo tag or MakePom will start injected maven central into pom.xml files silently.

sbt/src/sbt-test/dependency-management/pom-advanced/project/maven.sbt
```diff
@@ -0,0 +1,2 @@
+libraryDependencies += Defaults.sbtPluginExtra("org.scala-sbt" % "sbt-maven-resolver" % sbtVersion.value,
```
jsuereth, line 1: This we need to test without AND with the maven plugin.

jsuereth:
I forgot to cc @hamnis on this PR.   Would be good (even for a post-mortem review) of Aether code and any suggestions.  

Also, @hamnis, this new sbt-maven-resolver plugin duplicates a lot of the functionality of sbt-aether-deploy, but with the hooks to also resolve.    Would be awesome if you could do a review of the publishing code here and see if we can merge the two plugins in the future.   AFAIK, if you use the sbt-maven-resolver plugin you should see all the maven-metadata.xml show up on any publish/publishM2.

Also, Kudos to @hamnis for sbt-aether-deploy.   I learned enough about layouts from your code to be able to get resolution working, although I couldn't just directly re-use your code, due to the hackery to ensure Aether won't cache two sbt-versioned plugin jars in the same location.
- Josh

Issue #1730, opened 2014-11-14
Cached resolution doesn't handle test-dependency in multi-project build


fredriv:
I have a multi-project build where one subproject contains common test-code, which the other subprojects depend on in their test configuration, e.g.:

```
lazy val foo = project.in(file("foo"))

lazy val bar = project.in(file("bar")).dependsOn(foo % Test)
```

This build fails in SBT 0.13.7-RC5 if I enable cached resolution (it works fine otherwise). Basically, it does not include the compile dependencies from "foo" in the test configuration in "bar".

I've created a minimal test project that reproduces the bug at https://github.com/fredriv/sbt-cached-resolution-bug

Not sure if this is related to other cached resolution issues such as #1723 or if this is specific to multi-project builds.

eed3si9n:
Probably same as #1711.

fredriv:
I ran the test using SBT 0.13.7-RC5 which was released on Nov. 12th. https://repo.typesafe.com/typesafe/ivy-releases/org.scala-sbt/sbt-launch/0.13.7-RC5/

It seems the fixes for #1711 were merged on Nov. 10th?

eed3si9n:
> I've created a minimal test project that reproduces the bug at https://github.com/fredriv/sbt-cached-resolution-bug

Thanks for the repro project!
I tested on sbt 0.13.8-M6 and it seems to work now.

PR #1797, opened 2015-01-11
Fixes #1711, #1730. Cached resolution: fixes internal project

### LINKED ISSUE #1711: [0.13.7-RC2] cached resolution pulls in % "test" dependencies from other subproj

eed3si9n:
Fixes #1711, #1730

### DIFF
--- a/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
+++ b/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
@@ -58,18 +58,27 @@ private[sbt] class CachedResolutionResolveCache() {
       log.debug(s"::: expanding internal dependencies of module descriptor ${md0.getModuleRevisionId}")
       val rootModuleConfigs = md0.getConfigurations.toArray.toVector
       val rootNode = new IvyNode(data, md0)
+      // Initially confMap contains rootModuleConf mapped to itself.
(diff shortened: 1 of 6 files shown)


ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -82,17 +91,19 @@ private[sbt] class CachedResolutionResolveCache() {
       def nextConfMap(dd: DependencyDescriptor, previous: Map[String, Array[String]]): Map[String, Array[String]] =
         previous map {
           case (rootModuleConf, vs) =>
-            rootModuleConf -> (vs flatMap { conf =>
+            rootModuleConf -> (vs flatMap { conf0 =>
+              val conf = if (conf0 == "default(compile)") "compile"
```
jsuereth, line 95: Why are we special casing default(compile)?

jsuereth:
I think this LGTM, just want to understand why we have to special case.

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
eed3si9n, line 95: This is a hack on traversing Ivy modules across different configurations. `"default(compile)"` is treated specially because it happens to show up often.

eed3si9n:
Closed in favor of #1873

PR #1798, opened 2015-01-11
Adjust 'projectMacroImpl' to work for camelCased project names.

indrajitr:
Currently `project` macro allows us to do this:

``` scala
lazy val util = project // Project('util', file('util'))
```

This patch allows `project` macro to expands in backward compatible way allowing camelCased project names to be hyphenated like so without resorting to `backtick-workaround`:

``` scala
lazy val projUtil = project // Project('proj-util', file('proj-util'))
```

### DIFF
--- a/main/src/main/scala/sbt/Project.scala
+++ b/main/src/main/scala/sbt/Project.scala
@@ -567,7 +567,10 @@ object Project extends ProjectExtra {
       import c.universe._
       val enclosingValName = std.KeyMacro.definingValName(c, methodName => s"""$methodName must be directly assigned to a val, such as `val x = $methodName`.""")
       val name = c.Expr[String](Literal(Constant(enclosingValName)))
-      reify { Project(name.splice, new File(name.splice)) }
+      reify {
+        val hyphenated = Util.camelToHypen(name.splice)
+        Project(hyphenated, new File(hyphenated))
+      }
     }
 }
 


typesafe-tools:

eed3si9n:
Hi Indrajit. Nice seeing you here :)

This at the first this looks like an interesting PR, but I have a few concerns.
1. This breaks the semantics of `project`. In other words, if there were some build file, that relied on older behavior like `val projA = project`, this will break the build.
2. Even though people do sometimes use hyphen-concatenated lowercase, I don't think it's a strong enough convention for sbt 0.13. For instance, key names have started to migrate to camel case (to unify towards Scala syntax) in sbt 0.13. We frequently get criticism on sbt shell discrepancies from build.sbt, and we are likely to move towards `test in Test` instead of `test:test`.
3. In general, I am already not a big fan of `project` assuming things from its LHS because it feels like magic and because it needs to be explained to new users. Imagine writing documentation on this behavior. "`project` will take the name of `val`, and it turns it into hyphen-concatenated name."

indrajitr:
@eed3si9n, great points!

My thoughts:
1. Acknowledged! This didn't come to my mind. So this actually ends up being a breaking change after all -- that in itself trumps over any other point it might have had in favor.
2. Hmm, I thought we usually make the project id of the format `proj-lib`, `proj-core` etc. so that the `name` and thus the `moduleName` ends up being of that format (although I see that you don't follow that in your configs :smile:). In fact, I did come across quite a few cases where the developer resorted to `backtick` to achieve this and got me wondering how to persuade people to actually move _towards_ camel-case. [This](https://github.com/softprops/unisockets/blob/master/build.sbt) is the most recent one I saw today that triggered me to create this pull request.
3. Point taken. The fact that just writing `val x = project` does lot more with the name `x` feels like too much magic anyway. Too much implicit convention and it becomes crazy magikz, too much explicit expression, and it becomes crazy verbose :)

The benefit is very marginal (and maybe cosmetic) and the breaking change might not be worth it. Let's kill this. Alternately, this can be moved under `DefaultOptions` in case people chose to opt-in.

jsuereth:
Yeah, I'm in agreement with Eugene. We should strive for _consistency_ here.  Translating CamelCase to database-case everywhere is inconsistent in sbt.   Additionally, in sbt 1.0 we may even deprecate the old key db-style syntax bridge we've been using and turn all keys into the camel-case variant....

eed3si9n:
I am closing this PR. This could be a fun plugin though.

indrajitr:
Agreed, you guys got me convinced that this magic isn't a good idea -- even as a plugin.

Individual project builds might have this for convenience though.

PR #1799, opened 2015-01-12
Enable cross-version support for Scala sources.

indrajitr:
This allows attaching unmanaged sources on `per-ScalaVersion` basis.

For example, with `scalaBinaryVersion` set to `2.11`, the files in
`src/{main,test}/scala-2.11` as well `src/{main,test}/scala` would be
available in the `unmanagedSources` list.

This makes `unmanagedSources` consistent with `managedSources` (which is allowed to be scala version specific).

This is one of the older techniques which I initially felt would get irrelevant with time. But I am finding myself still using this repeatedly :)

Consider this more of a bouncing off idea with @jsuereth and @eed3si9n . We first need to see if this makes sense OOTB, and if it does, should we just append `scalaBinaryVersion` to `scalaSource` with no user control (like in the case of `managedSources`) or should we allow users some control via, say `crossSource` key (yet another Key!).

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -181,7 +181,7 @@ object Defaults extends BuildCommon {
     sourceManaged <<= configSrcSub(sourceManaged),
     scalaSource := sourceDirectory.value / "scala",
     javaSource := sourceDirectory.value / "java",
-    unmanagedSourceDirectories := Seq(scalaSource.value, javaSource.value),
+    unmanagedSourceDirectories := makeCrossSources(scalaSource.value, javaSource.value, scalaBinaryVersion.value, crossPaths.value),
     unmanagedSources <<= collectFiles(unmanagedSourceDirectories, includeFilter in unmanagedSources, excludeFilter in unmanagedSources),
     watchSources in ConfigGlobal <++= unmanagedSources,
     managedSourceDirectories := Seq(sourceManaged.value),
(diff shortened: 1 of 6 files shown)


typesafe-tools:

jsuereth:
I like the idea, and I agree it becomes something which should be default.

Ideally we avoid having a flag to enable this, and instead just silently add `src/main/scala-<binaryVersion>` _and_ `src/main/scala-<fullVersion>` to the list of sources.

However, for compatibility, it probably makes sense to have a flag (or autoplugin) which enables this.

eed3si9n:
1. +1. I think it makes sense to have this one out of the box.
2. I think we should be symmetric with the `crossTarget` and only add `src/main/scala-<binaryVersion>`. The reuse of `crossPaths` is fine with me.

eed3si9n:
`scripted source-dependencies/relative-source-error` is failing on Travis. I restarted it and failing on the same test.

indrajitr:
1. Having flag is something I am wanting to avoid now -- instead users can override `unmanagedSourceDirectories` completely.
2. Indeed, using `scalaBinaryVersion` instead of of full `scalaVersion` is to be consistent with `crossTarget`. This is also to stay consistent with sbt's treatment of `scalaBinaryVersion` pre-2.10 and 2.10 onwards (see the updated commit message for more). The unintended side effect of this is that `src/{main,test}/scala-2.10` would be included but `src/{main,test}/scala-2.9` _would not_ be included (`src/{main,test}/scala-<scalaVersion>` would be included instead). We have to choose between consistency with `crossTarget`'s behavior and element of surprise (why `src/{main,test}/scala-2.9` is not selected would be un-obvious).
3. `scripted source-dependencies/relative-source-error` could have been due to docker `tmp` mount point issue on Travis (have had this earlier). This didn't occur on my local env or even on the Travis build of my fork.
4. Regardless, have restructured a bit and added a small test to test the functionality.

eed3si9n:
LGTM

PR #1808, opened 2015-01-18
Addional form of  to allow better syntax for bundled plugins.

indrajitr:
Again, one of those random ideas to open conversation :)

Since this hurts: 

``` scala
libraryDependencies += Defaults.sbtPluginExtra("org.scala-sbt" % "sbt-maven-resolver" % sbtVersion.value, sbtBinaryVersion.value, scalaBinaryVersion.value)
```

and this isn't allowed:

``` scala
addSbtPlugin("org.scala-sbt" % "sbt-maven-resolver" % sbtVersion.value)
```

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -948,7 +948,7 @@ object Defaults extends BuildCommon {
   @deprecated("Settings now split into AutoPlugins.", "0.13.2")
   lazy val projectBaseSettings: Seq[Setting[_]] = projectCore ++ runnerSettings ++ paths ++ baseClasspaths ++ baseTasks ++ compileBase ++ disableAggregation
 
-  // These are project level settings that MUST be on every project.  
+  // These are project level settings that MUST be on every project.
   lazy val coreDefaultSettings: Seq[Setting[_]] =
     projectCore ++ disableAggregation ++ Seq(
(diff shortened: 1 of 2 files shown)


typesafe-tools:

jsuereth:
I like it!  good from me, cc. @eed3si9n for a second.

eed3si9n:
I think the idea is ok, but I don't like the name `addSbtPlugin`. When you have overloaded functions, you need to keep the semantics consistent, otherwise we get into _a lot_ of trouble. (Mostly people getting mad on twitter)

Why not make a new method?:

``` scala
addMavenResolverPlugin
```

indrajitr:
The intention is to have a simple way of activating bundled (but optional) `AutoPlugins` which would have same version as hosting sbt. Do we want to have new method just for a specific plugin?

eed3si9n:
If it's truly bundled like `JUnitXmlReportPlugin`, we don't need to put anything in `project/plugin.sbt`.
In this case hard-coded method with single purpose is more typesafe compared to allowing String-based parameters.

If you want to provide a function, I would name it to something like:

``` scala
addBuiltinPlugin("sbt-maven-resolver")
```

eed3si9n:
I implemented this in f42419bdd4cd6d9237e2603a248583fbabb6109d

indrajitr:
:+1:

PR #1816, opened 2015-01-19
Add javaScalaCompiler task

pauldraper:
I want to add a distributed cache of compiled targets and analyses, via an SBT plugin.

For that, I need to be able to change out the incremental compile implementation. That term is _very_ overloaded...I mean the step where we compile these X files to targets and report the relationships.

---

It seems like I should hook into the `compileIncremental` task. But I can't seem to do that without trying to (1) duplicate lots of functionality, and (2) access lots of private SBT classes.

Currently, compilation looks like this:

```
Keys.compileIncremental (Keys.scala)
Defaults.compileIncrementalTask (Defaults.scala)
Defaults.compileIncrementalTaskImpl (Defaults.scala)
Compiler.Compile (Compiler.scala)
IC.incrementalCompile (IncrementalCompiler.scala)
IC.compileInternal (IncrementalCompiler.scala)
```

which passes `MixedAnalyzingCompiler#compile` to

```
IncrementalCompile.apply (Compile.scala)
Incremental.compile (Incremental.scala)
```

which then uses one of the `IncrementalCommon` implementations, and does some more stuff.

---

I'm a little confused by the architecture (e.g. what belongs in `sbt.compiler.IC.incrementalCompile` vs `sbt.inc.IncrementalCompile` vs `sbt.inc.Incremental.compile`).

Currently, I hook into the `IC.compileInternal` -> `IncrementalCompile.apply` step.

I welcome suggestions how this can be better achieved.

---

Usage:

```
javaScalaCompiler := { compiler: MixedAnalyzingCompiler =>
  val analysis = previousCompile.value.analysis
  val underlying = javaScalaCompiler.value(compiler)
  (file: Set[File], changes: DependencyChanges, callback: AnalysisCallback) => {
    println(analysis)
    underlying(file, changes, callback)
  }
}
```


typesafe-tools:

main/actions/src/main/scala/sbt/Compiler.scala
```diff
@@ -106,14 +106,14 @@ object Compiler {
         (Analysis.empty(nameHashing = incOptions.nameHashing), None)
       }
     }
-    compile(InputsWithPrevious(in, PreviousAnalysis(previousAnalysis, previousSetup)), log, reporter).analysis
+    compile(InputsWithPrevious(in, PreviousAnalysis(previousAnalysis, previousSetup)), log, reporter, _.compile).analysis
   }
-  def compile(in: InputsWithPrevious, log: Logger): CompileResult =
+  def compile(in: InputsWithPrevious, log: Logger, compile_ : MixedAnalyzingCompiler => (Set[File], DependencyChanges, xsbti.AnalysisCallback) => Unit = _.compile): CompileResult =
```
jsuereth, line 111: As you may have noticed, there's a TON of overridden methods in here.   This is because we can't break binary compatibilty, which is why the API is a bit clunky (and why we have too many layers and can't remove/consolidate them).

Do you have time to do an in-person chat about this feature? I'd like to see more what you're trying to do, and look into an alternative design (as an iterim).

Note: in 1.0 we hope to get 99% of this API int' the `actoins/Compile.scala` so you can have your hook way up there.   Additionally IC vs. IncrementalCompiler distinction should disappear.

jsuereth:
Question:  Why not just replace the AnalyzingCompiler + JavaAnalyzingCompiler with your own that reports status?  I understand the API is rough (I was just in there trying to clenaup without breaking binary compatibility).

I love the feature you want to provide. Let's see if we can get a good hook that's binarycompatible.   I _think_ we may be able to sneak it into Inputs and you can wrap the Analyzing compilers, rather than trying to force a function all the way down the chain...

pauldraper:
@jsuereth , I had the same thought.

I thought about the [`compilers`](https://github.com/sbt/sbt/blob/0.13/main/src/main/scala/sbt/Keys.scala#L148) task key.

However, `AnalyzingCompiler` is a final class, and also has a fair number of methods.

`JavaTool` is a trait, but deprecated in 0.13.8, and it isn't clear to me what the replacement is, or what is going on there.

> This is because we can't break binary compatibilty

Ah, I wasn't aware that SBT had any binary compatibility requirements around this area.

> Do you have time to do an in-person chat about this feature?

Yeah, if that would help, I'd be very happy to. I have Google (paulddraper@gmail.com).

jsuereth:
Ok, so I've just refreshed myself through all that code.

`JavaTool` is deprecated in favor of `JavaTool`, i.e. this new one: https://github.com/sbt/sbt/blob/0.13/compile/src/main/scala/sbt/compiler/javac/JavaCompiler.scala#L98-L117.  The old interface was a bit wonky, the new one is a "plain old interface".

Now, regarindg binary compatibility, yes we care.   We have a lot of plugin that depend on us, in addition to Zinc.  Anything we do in this codepath, being publicly exposed, is dangerous.  I'd like to migrate most of it into something a bit more private w/ a nice public interface, but that's a task for later.

In any case, just curious, given the recent work we did for bytecodeEnhancers, can't you just do the following:

```
compileIncremental in Compile := {
   val result = (compileIncremental in Compile).value
   publishToMyServer(result)
   result
}
```

I'd like to better understand the use case.  Also, sent you an email, let me know if you didn't see it.  Sorry for taking so long to get to this!

dwijnand:
What do we want to do here? It's been 6 months, is this code worth salvaging and getting it merged in?

pauldraper:
Sorry, I dropped this.

Given the slow compile times for Scala, I think there is something good that could be done with caching. But I haven't the time currently, so I'm closing this.

PR #1817, opened 2015-01-20
Def.sequential. Fixes #1001

### LINKED ISSUE #1001: Make it easier to control the sequencing of Tasks
As Eugene [also discussed](https://groups.google.com/forum/#!topic/simple-build-tool/1c7unObqjJ8), it's awkward to control the sequencing of tasks in sbt builds.  Let's see if we can find a simple control structure.

One common use case for task sequencing is for integration testing:
  Task 1: Start a test server / wait for it to come on line
  Task 2: run integration tests
  Task 3: stop servers

Controlling remote servers is another use case for task sequencing, and described in #1000.

I wonder if we could imitate the Future apis for Tasks (e.g. Future.sequence), and possibly use the scala async macro..

eed3si9n:
Adds `Def.sequential` based on Mark's implementation provided in [#1001](https://github.com/sbt/sbt/issues/1001#issuecomment-29259967). This one does more type wrangling so it's easier to use.

This runs tasks under semi-sequential semantics. Here's an example usage:

``` scala
lazy val root = project.
  settings(
    testFile := target.value / "test.txt",
    sideEffect0 := {
      val t = testFile.value
      IO.append(t, "0")
      t
    },
    sideEffect1 := {
      val t = testFile.value
      IO.append(t, "1")
      t
    },
    foo := Def.sequential(compile in Compile, sideEffect0, sideEffect1, test in Test).value
  )
```

Normally sbt's task engine will reorder tasks based on the dependencies among the tasks, and run as many tasks in parallel (See [Custom settings and tasks](http://www.scala-sbt.org/0.13/tutorial/Custom-Settings.html) for more details on this).
`Def.sequential` instead tries to run the tasks in the specified order.
However, the task engine will still deduplicate tasks. For instance, when `foo` is executed, it will only compile once, even though `test in Test` depends on compile.


sbt/src/sbt-test/project/sequential/build.sbt
```diff
@@ -0,0 +1,37 @@
...
+    sideEffect0 := {
+      Thread.sleep(100)
+      val t = testFile.value
+      IO.append(t, "0")
+    },
+    sideEffect1 := {
+      val t = testFile.value
+      IO.append(t, "1")
+    },
+    sideEffect2 := {
```
jsuereth, line 31: Can we add a dependency from sideEffect2 to sideeffect1 to ensure 1 is not re-run?

jsuereth:
Minor test checks, otherwise LGTM

sbt/src/sbt-test/project/sequential/build.sbt
eed3si9n, line 31: Updated the test

PR #1831, opened 2015-01-29
Add support for external apidoc URLs for both Scala and Java projects.

indrajitr:
We now support attaching external apidoc URL for both Scala and Java projects via explicit keys `scalaApiMappings` and `javaApiMappings` respectively.
The old key `apiMappings` is thus deprecated in favor of more explicitly named `scalaApiMappings`.

The new set POM properties preserving the apidoc URLs are `info.scalaApiURL` and `info.javaApiURL` for Scala apidoc and Java apidoc respectively.


typesafe-tools:

ivy/src/main/java/sbt/mavenint/SbtPomExtraProperties.java
```diff
@@ -12,7 +12,11 @@
     public static final String POM_INFO_KEY_PREFIX = "info.";
     public static final String POM_SCALA_VERSION = "scalaVersion";
     public static final String POM_SBT_VERSION = "sbtVersion";
+    /** @deprecated Use {@link #POM_SCALA_API_KEY} instead */
+    @Deprecated
```
eed3si9n, line 16: I don't know if deprecating `info.apiURL` given that there are existing poms out there is worth the change for consistency.
1. We would have to support both `info.apiURL` and `info.scalaApiURL`.
2. If someone had written a tool on using `info.apiURL`, we'll break it.

@jsuereth  Am I being too paranoid?

indrajitr, line 16: FWIW, we _do_ intend to honor `info.apiURL` as fallback to `info.scalaApiURL` for pre-existing poms. So we are good on that front.

However, there could potentially be case where you publish an upstream lib with sbt 0.13.8 (and thus have scalaApiURL), however downstream library might have been using sbt 0.13.6 which won't see apiURL.

So, the point @eed3si9n is making is valid and we can stay with `info.apiURL` if we _must_ at the cost of adding to the overall non-intuitiveness :)

jsuereth, line 16: Well, to resolve this, here's what I'd say:
1. We do a survey of recently released scala libs and figure out the % which are actively using the apiURL feature.  Given its low press, I'm assuming it's not heavily used yet.
2. If we find apiURL isn't heavily used (and given its recent addition) we deprecate it in favor of scalaApiURL and javaApiUrl. Basically, if people have trouble seeing the new libs, they should be freely able to upgrade to sbt 0.13.8, since it's bin-compat.   We need to make sure this is always true.
3. After this fix, we can start advertising the feature more, possibly recommending the OSS community us it.   I'll let you, @indrajitr, take the lead on pushing its use and ensuring libraries across teh OSS ecosystem can cross-linke docs.

Sound likea  good plan?

jsuereth:
Ok, so the ivy related code looks fine.  You may also need to hit here:

The POM reader in the sbt-eether-resolver plugin: https://github.com/sbt/sbt/blob/0.13/sbt-maven-resolver/src/main/java/org/apache/maven/repository/internal/SbtArtifactDescriptorReader.java#L274-L289

Then, after the POM properties are read into Aether, they are translated into Ivy Properties here:

https://github.com/sbt/sbt/blob/0.13/sbt-maven-resolver/src/main/scala/sbt/mavenint/MavenRepositoryResolver.scala#L146-L148

We need this path to work as well, as I expect at some point we may just stop relying on Ivy's maven-integration in favor of Aether handling its own mess.

jsuereth:
@indrajitr We just had a discussion that opened some cans of worms here.  Besides Aether-integration, we kind of have to figure out what tools may be using info.apiURL before we remove it.   @eed3si9n has convinced me of the rightness of sticking with info.apiURL for scaladoc (and possibly adding info.javadocURL, for javadoc).

I wish we had done this in a cleaner fashion immediately.  Also, relating to info.apiURL, we should try to push this convention on thw hole maven ecosystem and all toolign aroudn maven artifacts.   Otherwise it's an obscure sbt feature.   Ideas here welcome.

eed3si9n:
@jsuereth wrote:

> 1. We do a survey of recently released scala libs and figure out the % which are actively using the apiURL feature. Given its low press, I'm assuming it's not heavily used yet.

Let's start from the most widely used OSS Scala library, which is named scala-library. Here's from [scala-library-2.11.6.pom](https://repo1.maven.org/maven2/org/scala-lang/scala-library/2.11.6/scala-library-2.11.6.pom):

``` xml
<properties>
<info.apiURL>http://www.scala-lang.org/api/2.11.6/</info.apiURL>
</properties>
```

I don't think we should now deprecate this. As it stands, this PR is not ready for 0.13.9.

indrajitr:
I'll refresh the change shortly. Sorry about the radio silence.

I just need to squash my changes and update this request (for 0.13.10 maybe?).

Few points though:
- Existing POMs are never in danger because the code was intended to _always_ honor `info.apiURL` as fallback to `info.scalaApiURL`.
- That said, I agree that the keeping `info.apiURL` around is safer.
- So we should just have `info.apiURL` and `info.javaApiURL` (and have necessary comments in the code so that our future self remembers the decision).

Fair?

dwijnand:
@indrajitr Are you still interested in this?

PR #1836, opened 2015-02-01
Invalidate macro providers whose transitive dependencies have been recompiled

Duhemm:
Hi !

This pull request introduces the invalidation of macro providers whose transitive dependencies have been recompiled more recently than the macro provider.

The scripted test gives a pretty good example of why and when this invalidation should occur to keep consistant macro expansions.

This pull request also adds a new option to `IncOptions` to disable this behaviour, because it may be annoying for some projects that use a lot of macros (e.g. shapeless).

---
#### Consider a project composed of 3 Scala sources:
- `Provider.scala`: This files defines a macro. The implementation of this macro depends on `Helper.something()`.

``` scala
object Provider {
  def impl(c: Context): c.Tree = {
    import c.universe._
    val result = Helper.hello()
    q"$result"
  }
  def hello = macro impl
}
```
- `Helper.scala`, which provides a function that is used by the macro implementation

``` scala
object Helper {
  def hello(): String = "Hello"
}
```
- `Client.scala` which uses the macro `Provider.hello`.

``` scala
object Client {
  println(Provider.hello) // Will expand to println("Hello")
}
```
#### The code gets modified

Imagine now that the user modifies the implementation (but not the signature) of the method `Helper.hello`:

``` scala
object Helper {
  def hello(): String = "Bonjour"
}
```
#### What gets recompiled without this PR
- `Helper.scala` because it has been modified. Nothing else is recompiled.
#### What should be recompiled
- `Helper.scala`, because it has been modified.
- `Client.scala`, because its expansion is outdated (it should now be `println("Bonjour")`)
#### What gets recompiled with this PR
- `Helper.scala`, because it has been modified.
- `Provider.scala`, because **it has a macro implementation that depends on something that has been recompiled**
- `Client.scala`, because it is a client of a macro provider that has been recompiled.
  The reason why `Provider.scala` is recompiled is because this is the easiest way to make sure that all the clients of its macro will be recompiled.
#### What's the underlying problem?

Dependencies of macro implementation cannot be treated like classical dependencies. With classical dependencies, sbt reasons about changes to the public API of a file to know what other files may be impacted by the changes. If a file has been modified, but its API has remained unchanged, then there's generally no need to recompile anything else.

Because macros are expanded at compile-time, we cannot simply rely on API changes to know whether or not a change will affect potential expansions of a macro. In a normal Scala function, the byte code of this function will be invoked whenever the function is applied. With macros, it will be called only once at compile time. Further changes to the implementation of this function should however trigger the recompilation of macros that use it, because expanding again the macro could yield something different than what we got with the previous implementation of the macro.

In this example, the public API of `Helper.scala` has not been changed, but we should however recompile `Client.scala`, because it uses the macro `Provider.hello`, and this macro has dependencies that have been recompiled. If we do not perform re-expansion of the macro `Provider.hello`, we stay with the previous expansion which was `"Hello"`. Recompiling `Client.scala` will yield a new expansion `"Bonjour"`. Clearly, if `Provider.hello` was not a macro there would be no need to recompile anything else than `Helper.scala`.

In the previous scenario we only considered the case where `Helper.scala` gets modified. We also have to consider **all** (not only `DependencyFromMacroImpl`!) the transitive dependencies of `Helper.scala`: A change to the implementation of a function that `Helper.scala` calls could change the return value of `Helper.hello`, which may change the expansion of `Provider.hello`).
#### How is this problem fixed ?

This PR introduces a new dependency context called `DependencyFromMacroImpl`. The goal of this dependency context is to track the dependencies that have been introduced inside the body of a macro implementation. In the present example, sbt would register that `Provider.scala` has a dependency with context = `DependencyFromMacroImpl` on `Helper.scala`.
##### What will sbt do with these dependencies?

In our previous example, before the code gets changed, sbt will know that all the files have been compiled for the last time on time:

```
CTime(Helper.scala) = CTime(Provider.scala) = CTime(Client.scala) = t
```

Then, `Helper.scala` is modified, it will therefore be recompiled, and we get in this configuration:

```
CTime(Helper.scala) = t1 > CTime(Provider.scala) = CTime(Client.scala) = t
```

But `Provider.scala` depends with `DependencyFromMacroImpl` on `Helper.scala`. It should therefore be recompiled every time that `Provider.scala` gets recompiled. sbt will recompile `Provider.scala`:

```
CTime(Provider.scala) = t2 > CTime(Helper.scala) = t1 > CTime(Client.scala)
```

Because `Provider.scala` is a macro provider, all the files that depend on it will be recompiled (this behaviour was already implemented and has not been changed in this pull request):

```
CTime(Client.scala) = t3 > CTime(Provider.scala) = t2 > CTime(Helper.scala) = t1
```
##### How is invalidation of macro providers performed?

For all macro provider `m`, sbt will get the time `Tm` of the last compilation of `m`. Because `m` is a macro provider, it has dependencies with context = `DependencyFromMacroImpl`. sbt will get all the files `d` that `m` depends on in this dependency context.
- If any of these files have a last compilation time that is greater than `Tm`, then sbt will recompile `m`.
- Otherwise, sbt will consider **all the dependencies** of `d`, and try to find one of its transitive dependency that has a last compilation time that is greater than `Tm`. If it finds one, then `m` will be recompiled, otherwise it will not.


typesafe-tools:

Duhemm:
Hello!
Are there any updates on the status of this PR?
It looks like the failures were due to unresolved dependencies, and are not related to the changes brought by this pull request.

gkossakowski:
I think the description of this PR is unclear. Could you give an example of a macro and recompilation scenario as it's done in current incremental compiler (without merging this PR). Then, could you point out what's the specific problem (e.g. which files are not recompiled and why) and then explain how this PR fixes it? The template would be:

> You have macros defined like:
> [here's code of macros]
> There's also code that uses macros:
> [here's user code]
> Let's imagine a user modifies his code this way
> [show changes]
> the following files will get recompiled in the following iterations:
> [explain which files get recompiled and why]
> This is not ideal because we would expect the following files to be recompiled:
> [list files you would like to be recompiled and why]
> The underlying problem is...
> This PR addresses the underlying problem by...

The whole description should be understandable to somebody who knows how macros in Scala work but doesn't need to know details of incremental compiler implementation to follow your description.

Once this is done, I'll start digging into code. Let me know if what I'm asking for is clear to you.

Duhemm:
@gkossakowski I've updated the description of the pull request. Please tell me if anything is still unclear.

gkossakowski:
The new PR description reads great. What would happen if `Hello.scala` would look like this:

``` scala
object Helper {
  def hello(): String = Foo.msg
}
```

with `Foo.scala` defined as:

``` scala
object Foo {
  def msg: String = "hello"
}
```

Will recompilation of `Foo.scala` trigger invalidation of macro provider?

Duhemm:
Yes. The compilation time of all the transitive dependencies of macro implementations are considered, both internal and external.
I created a sample repo when I worked on this : https://github.com/Duhemm/macro-transitive-dependencies (Actually, I think this is the same example as the scripted test in this PR. The readme explains what happens in this example project)

dwijnand:

PR #1841, opened 2015-02-03
Various changes I made prior to pushing out M3

eed3si9n:


build.sbt
```diff
@@ -578,10 +593,16 @@ def customCommands: Seq[Setting[_]] = Seq(
     "clean" ::
       "checkCredentials" ::
       "conscript-configs" ::
+      "interfaceProj/compile" :: // Java project needs to compile first
```
jsuereth: Ok, this seems pretty broken then.  Not really a huge fan of sbt-doge if it does this.

eed3si9n: This may or may not be needed with the latest sbt-doge. I needed to get a build out first then fix doge.

eed3si9n:
The build failures seems transient:

```
downloading https://oss.sonatype.org/content/repositories/releases/org/scala-lang/scala-compiler/2.9.0/scala-compiler-2.9.0.jar ...
:: problems summary ::
:::: WARNINGS
        module not found: org.scala-lang#scala-library;2.9.0
```

`Launcher> testOnly xsbt.boot.ScalaProviderTest` passes locally.

build.sbt
```diff
@@ -568,20 +576,47 @@ def customCommands: Seq[Setting[_]] = Seq(
...
+   * Because of the way ++ (and its improved version wow) is implemented
+   * precompiled compiler briges are handled outside of doge aggregation on root.
+   * `so compile` handles 2.10.x/2.11.x cross building. 
+   */
   commands += Command.command("release-sbt") { state =>
     // TODO - Any sort of validation
     "clean" ::
       "checkCredentials" ::
       "conscript-configs" ::
+      "precompiled-2_8_2/compile" ::
```
jsuereth: This is still bad.  Why can't doge handle this?

PR #1842, opened 2015-02-03
Build fix

eed3si9n:
Continuation of #1841


build.sbt
```diff
@@ -460,7 +477,7 @@ def otherRootSettings = Seq(
   Scripted.scriptedUnpublished <<= scriptedUnpublishedTask,
   Scripted.scriptedSource <<= (sourceDirectory in sbtProj) / "sbt-test",
   publishAll := {
-    (publishLocal).all(ScopeFilter(inAnyProject)).value
+    val _ = (publishLocal).all(ScopeFilter(inAnyProject)).value
```
jsuereth, line 511: why this change?

jsuereth:
I think this now LGTM.

build.sbt
eed3si9n, line 511: There was a warning saying the expression's not doing anything or something to that effect.

eed3si9n:
Added two more commits related to the build.

jsuereth:
LGTM

Issue #1368, opened 2014-05-23
sbt classloaders should not return null from java.lang.ClassLoader#getResources


OlegYch:
this breaks general contract of the method
in particular i'm having an issue with NullLoader returning null at https://github.com/hazelcast/hazelcast/blob/db8721eab91961d61ef4957682e71ce9dc3fdcf7/hazelcast/src/main/java/com/hazelcast/util/ServiceLoader.java#L169
using sbt 0.13.5-RC5

hcoles:
Just hit the same issue, also with hazelcast.

@OlegYch Did you find a workaround?

PR #1849, opened 2015-02-09
Classloader::getResource() should never return null

jerrinot:
fixing #1368

### DIFF
--- a/util/classpath/src/main/scala/sbt/classpath/ClassLoaders.scala
+++ b/util/classpath/src/main/scala/sbt/classpath/ClassLoaders.scala
@@ -6,6 +6,7 @@ package classpath
 
 import java.io.File
 import java.net.{ URL, URLClassLoader }
+import java.util.Collections
 import annotation.tailrec
 
 /**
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:
Hi @jerrinot,

typesafe-tools:

jerrinot:
I just signed the individual CLA

hcoles:
This doesn't compile as it uses the java 7 only Collections.emptyEnumeration.

There are also some other areas of the code where null might be returned e.g in NullLoader. I am unclear if this fix would also catch these.

I'm unable to try this out locally due to this problem -> http://stackoverflow.com/questions/28431934/publish-precompiled-x-binaries-locally-for-sbt

I'd be happy to produce a clean tested PR if some-one can help me through the issue above.

hcoles:
@jerrinot I've put together an alternate PR that is confirmed to work and does not rely on java 7. It might be worth closing this one to avoid confusion.

jerrinot:
@hcoles: great, thanks for info!
I'm closing to PR in favour for #1853

PR #1853, opened 2015-02-11
Fix #1368 - getResources should never return null

### LINKED ISSUE #1368: sbt classloaders should not return null from java.lang.ClassLoader#getResources

hcoles:
This has been tested manually and confirmed to fix the original issue in which the return null broke builds that used hazelcast.

I've checked for any code within the sbt repo relying on a null return from getResources, but could find none apart from the updated lines and ClassLoaders.scala.

I've left the null check and subsequent null return in place on line 89 of ClassLoaders.scala as it does no harm and might conceivably be relied on by third party code with it's own non compliant classloader.

### DIFF
--- a/util/classpath/src/main/scala/sbt/classpath/DualLoader.scala
+++ b/util/classpath/src/main/scala/sbt/classpath/DualLoader.scala
@@ -6,12 +6,13 @@ package classpath
 
 import java.net.URL
 import java.util.Enumeration
+import java.util.Collections
 
 /** A class loader that always fails to load classes and resources. */
 final class NullLoader extends ClassLoader {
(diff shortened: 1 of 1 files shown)


typesafe-tools:

eed3si9n:
LGTM pending CI.

jsuereth:
LGTM as well

PR #1858, opened 2015-02-13
Adds an hook treating InlineConfigurationWithExcludes structurally

eed3si9n:
This allows us to treat newly added `InlineConfigurationWithExcludes` via sbt plugins using structural subtyping.

/review @havocp, @jsuereth


jsuereth:
LGTM

jsuereth:
So It hink perhaps master has failing builds, see:

https://travis-ci.org/sbt/sbt/jobs/50697890#L4388-L4391

eed3si9n:
I'm embarrassed by the fact that our build history is too bloody to tell what's real and unreal any more - https://travis-ci.org/sbt/sbt/builds
First thing I do on Monday should be to rectify this.

eed3si9n:
The CI failures are coming from unresolved precompiled interfaces, which were removed in #1842.
`scripted project/transitive-plugins` passes locally, and I'm able to launch sbt 0.13.8-M4 with a build containing plugins fine. I'm restarting failed Travis jobs to see what happens.

eed3si9n:
Travis is right. After `cleanLocal` (with sbt-dirty-money), I can reproduce this error.

PR #1784, opened 2014-12-18
Multi-project build.sbt

eed3si9n:

PR #1859, opened 2015-02-15
Fix nightly publishing location by demoting publish-status to ThisBuild

eed3si9n:
`stamp-version` is a command used to substitute `-SNAPSHOT` with the current timestamp. This is called by Jenkins job for the nightlies.
## steps
1. run `stamp-version` from sbt shell
2. run `publishStatus` from sbt shell
## problem

`publishStatus` currently returns "releases", causing nightly releases to get published on the wrong repository by the Jenkins job.
## expectations

`publishStatus` returns "snapshots".
## analysis

Nightlies broke in general when I migrated the build in #1784 so we didn't notice this for a while.
Another thing I broke there was promoting `publishStatus` setting from build level to project level in
https://github.com/sbt/sbt/pull/1784/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR27.
This overrides `(publishStatus in ThisBuild := status)`, which is part of the `stamp-version` command.
A few days ago #1842 was merged, making the nightlies to build again, and now this bug is surfaced.

The fix is to demote `publishStatus` back as a build-level setting by making it an auto plugin.

/review @jsuereth, @havocp

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,7 +30,7 @@ def commonSettings: Seq[Setting[_]] = Seq(
 )
 
 def minimalSettings: Seq[Setting[_]] =
-  commonSettings ++ customCommands ++ Status.settings ++
+  commonSettings ++ customCommands ++
   publishPomSettings ++ Release.javaVersionCheckSettings
 
 def baseSettings: Seq[Setting[_]] =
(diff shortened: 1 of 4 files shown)


jsuereth:
lgtm

PR #1871, opened 2015-02-22
Disable publishing artificial root.

### LINKED ISSUE #1869: artificial root should not publish by default
Suggested by @softprops at nescala. Please comment on this issue, if someone wants to send PR for this.

dwijnand:
Fixes #1869.

First off, this is my first proper venture into the non-Keys/Defaults parts of
sbt, and first pull request on sbt..

My manual test shows this to fix it, I'm not sure I picked the best way (key)
or place to fix this, but I thought I'd share what I've found work.

I'm happy to iterate on this.


typesafe-tools:

eed3si9n:
Thanks for the PR. 

I think we should add some test. Are you familiar with the [scripted test](http://eed3si9n.com/testing-sbt-plugins)?
Ours are under [sbt/src/sbt-test](https://github.com/sbt/sbt/tree/0.13/sbt/src/sbt-test). Could you add a scripted test with an isolated `ivyPaths`

```
    ivyPaths := new IvyPaths((baseDirectory in LocalRootProject).value, Some((target in LocalRootProject).value / "ivy-cache")),
```

and locally publish the artificial root and check that nothing gets published (including ivy.xml)?

dwijnand:
No ivy files are being published, I'll look into those.

I know scripted by name, have run them, but never created any so I'll try and see if I can figure it out here.

eed3si9n:
I'm seeing this locally:

```
[N:/var/folders/7g/l96y5q310h90xsz109qqjpmc0000gn/T/sbt_dcd48716/default-root/target/ivy-cache/local]$ ls **/**.xml
com.example/app_2.10/0.1.0/ivys/ivy.xml           com.example/default-root_2.10/0.1.0/ivys/ivy.xml
```

Here's my scripted test `build.sbt`:

``` scala
val commonSettings = Seq(
  organization := "com.example",
  version := "0.1.0",
  ivyPaths := new IvyPaths((baseDirectory in LocalRootProject).value, Some((target in LocalRootProject).value / "ivy-cache"))
)

lazy val app = (project in file("app")).
  settings(commonSettings: _*)

commonSettings

lazy val check = taskKey[Unit]("check")
check := {

}
```

Haven't implemented the `check` yet, so you can take that if you want and add the `check` part.

dwijnand:
Yeah, but completely overriding publish/publishLocal fixed the ivy file, that's one done.
Thanks for the test, I can try and finish it.

dwijnand:
I wasn't sure how to do it with a check task :-/

eed3si9n:
`check` or not doesn't matter as long as it's tested :)
LGTM pending Travis.

eed3si9n:
Actually I just thought about one more thing.
What if you're using [bare build.sbt](http://www.scala-sbt.org/0.13/tutorial/Bare-Def.html)? Would you lose `publish`? This could be trickier than I thought. Maybe we have to check on some factor (existence of sources?) to turn off publishing.

dwijnand:
Indeed that's broken, looking into it.

eed3si9n:
Here are some of the failures from [Travis](https://travis-ci.org/sbt/sbt/jobs/51722380):

```
[error] x dependency-management / publish-local
[error]    {line 1}  Command failed: publish-local failed
[error] x dependency-management / snapshot-resolution [PENDING]
[error]    {line 17}  Command failed: dependent/compile failed
```

Assuming these failures are legitimate (sometimes Travis is flaky), the change as it stands would end up breaking existing builds in the wild. Could you find out why they are failing? At a quick glance, publish-local test seem to have an explicit root project, except it's defined in `project/MultiPublishTest.scala`.

We might need to add additional checks to remove `publish`:
1. not having any source in `Compile` AND
2. not having any source in `Test` AND
3. not having any library dependencies

But I guess it will depend on why the tests are failing.

dwijnand:
So I saw those and ran those block of scripted tests locally and they didn't fail..?.. I'm a little unsure because they take so long to run I started doing other things and now I'm not sure. Either way, yeah I agree we should check again. I see if I can get to them tomorrow.

dwijnand:
I re-ran both sets of tests (`scripted dependency-management/*2of2` and `mavenResolverPluginTest:scripted dependency-management/*2of2`) and only had `dependency-management / snapshot-resolution` fail for me (in both sets), but just as much as it's failing for me on the current head of 0.13.

Does `scripted dependency-management/snapshot-resolution` on the 0.13 branch, work for you locally? I looked into it (`$ pause`) and the relevant jars are published in the right places, but snapshot resolution is pulling in the wrong/bad version for some reason.

Either way it looks to me like it's unrelated, and the current tests are passing otherwise (despite Travis). Do you want to block this pull request to explore additional reasons for avoid publishing root, or do you want to do it as another pull request?

eed3si9n:
I'm going to clean up the failing builds tomorrow. My own PR is blocked as well since we can't get clean build.

dwijnand:
OK cool. Thanks man.

eed3si9n:
I fixed some Travis setting, and the build is looking better. Could you rebase on 0.13 branch?

eed3si9n:
Closing this in favor of #1878

Issue #1811, opened 2015-01-18
Cached resolution issues


eed3si9n:
This issue is a summary of cached resolution issues:
- <s>#1711 Cached resolution: pulls in % "test" dependencies from other subproj</s>
- <s>#1730 Cached resolution doesn't handle test-dependency in multi-project build</s> 
- <s>#1758 Cached resolution interferes with exclusions and custom configs</s> 
- <s>#1760 cached resolution: transitive dependency survives even after its caller is evicted</s>
- <s>#1761 Modules were resolved with conflicting cross-version suffixes only when cached resolution is on </s>
- <s>#1721/#1763 Cached resolution: OOM when compiling</s>
- <s>#1776 Cached resolution: updateClassifiers does not download sources</s>
- #1868 Cached resolution: integration tests do not see libraries from dependencies with caching
- #1930 Cached resolution: updateClassifiers with cached resolution is really slow
- <s>#2014 Cached resolution: Minigraph files for dynamic dependencies are cleaned up</s>
- <s>#2046 Cached resolution falsely evicts netty (spark + cassandra)</s>
- <s>#2100 withConsolidatedResolution(true) - update gets slower and slower as more time is spent parsing and saving graph.json</s>
- #2105 cached resolution: remove transitive dependencies from callers array when using consolidated resolution

PR #1873, opened 2015-02-23
Fixes #1711, #1730. Cached resolution: fixes internal project

### LINKED ISSUE #1711: [0.13.7-RC2] cached resolution pulls in % "test" dependencies from other subproj

### LINKED ISSUE #1776: cached resolution: updateClassifiers does not download sources
## steps

eg with this config and sbt 0.13.7

```
scalaVersion := "2.11.4"
//fork := true
libraryDependencies += "org.scalaz.stream" %% "scalaz-stream" % "snapshot-0.6a"
resolvers += "Scalaz Bintray Repo" at "http://dl.bintray.com/scalaz/releases"
transitiveClassifiers in updateClassifiers := List("sources")
updateOptions := updateOptions.value.withCachedResolution(true).withLatestSnapshots(false)
```
## problem

updateClassifiers will only download "bundles"
## expectation

without cached resolution it will also download "jars" and "srcs" (not sure why it downloads both bundles and jars, though)

eed3si9n:
Fixes #1711, #1730

Re-fixes cached resolution's internal dependency issue by recursively calling customResolve instead of including the transitive dependencies from internal dependencies into your own graph.

Transformation of configuration still happens, but at the level of resolved graph (UpdateReport), which is much less granular, and hopefully less error-prone.

This should also fix #1776. Needs testing.

See #1811 for the list of cached resolution issues.

### DIFF
--- a/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
+++ b/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
@@ -17,6 +17,7 @@ import core.{ IvyPatternHelper, LogOptions }
 import org.apache.ivy.util.{ Message, MessageLogger }
 import org.apache.ivy.plugins.latest.{ ArtifactInfo => IvyArtifactInfo }
 import org.apache.ivy.plugins.matcher.{ MapMatcher, PatternMatcher }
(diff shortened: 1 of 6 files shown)


jsuereth:
LGTM

eed3si9n:
Rebased to 0.13.

PR #1876, opened 2015-02-24
Fix Travis setting

eed3si9n:
Trying to see if this fixes the build.

`scripted dependency-management/publish-local` fails on Travis, but works locally.

/review @jsuereth

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,16 +4,7 @@ sudo: false
 cache:
   directories:
     - $HOME/.ivy2/cache
-
-    # At the moment, sbt 0.13.5 is preinstalled in Travis VM image,
-    # which fortunately corresponds to current scalaz settings.
-    # The line below can be used to cache a given sbt version.
-    # - $HOME/.sbt/launchers/0.13.x
-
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,16 +4,7 @@ sudo: false
...
-    # At the moment, sbt 0.13.5 is preinstalled in Travis VM image,
-    # which fortunately corresponds to current scalaz settings.
-    # The line below can be used to cache a given sbt version.
-    # - $HOME/.sbt/launchers/0.13.x
-
-    # The line below is used to cache the scala version used by the build
-    # job, as these versions might be replaced after a Travis CI build
-    # environment upgrade (e.g. scala 2.11.2 could be replaced by scala 2.11.4).
-    - $HOME/.sbt/boot/scala-$TRAVIS_SCALA_VERSION
+    - $HOME/.sbt/boot
```
jsuereth, line 7: nice.  This should be ok, although not sure our launcher tests will cause issues with this cache....

jsuereth:
LGTM

PR #1880, opened 2015-02-25
Fixing precompiled and nightly build

eed3si9n:
We noticed that -SNAPSHOT is being published as our nightly.
This is because "wow" command (or ++) does not replay version injected by stamp-version.

Another thing is that precompiled and `++` have bad interaction, so I'm pushing everything precompiled up front.

```
java.text.ParseException: inconsistent module descriptor file found in 'https://repo.typesafe.com/typesafe/ivy-snapshots/org.scala-sbt/precompiled-2_8_2/0.13.8-SNAPSHOT/ivys/ivy.xml': bad module name: expected='precompiled-2_8_2' found='precompiled-2_9_3';
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -604,11 +604,12 @@ def customCommands: Seq[Setting[_]] = Seq(
     test.all(safeProjects).value
   },
   commands += Command.command("release-sbt-local") { state =>
-    "so clean" ::
+    "clean" ::
     "allPrecompiled/clean" ::
     "allPrecompiled/compile" ::
-    "so compile" ::
     "allPrecompiled/publishLocal" ::
+    "so clean" ::
+    "so compile" ::
     "so publishLocal" ::
     "reload" ::
     state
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -631,27 +632,27 @@ def customCommands: Seq[Setting[_]] = Seq(
...
-      "so clean" ::
+      "clean" ::
       "allPrecompiled/clean" ::
       "allPrecompiled/compile" ::
-      "so compile" ::
-      "so publish" ::
       "allPrecompiled/publish" ::
-      "publishLauncher" ::
+      "compile" ::
+      "publish" ::
```
jsuereth, line 655: doesn't this mean we aren't pushing the 2.11 artifacts nightly?

jsuereth:
so, this is because "stamp-version" is injecting settings in a location no one uses/i.e. meant to be transietnt.   We literally had to fix ++ to _not_ do that in the past.  I'm not suprised stamp-version doesn't work with anything else which sets settings.

THe fix is ok for nightly, but not ideal as we don't push/test the 2.11 things.

build.sbt
eed3si9n, line 655: I don't think we were before either since Jenkins was issuing `clean stamp-version publish`.

PR #1879, opened 2015-02-25
(don't merge) Limit the Travis tests to scripted dependency-management/publish-local

eed3si9n:
This is a PR for debugging Travis failures on `scripted dependency-management/publish-local`

PR #1885, opened 2015-02-27
Fix ANSI escape sequences.

### LINKED ISSUE #1143: ansi escape sequence stripping is broken
The ansi escape sequence stripping (`ConsoleLogger.removeEscapeSequences`) is broken, especially when stripping escapes for colouring.  An example colour escape sequence is `<ESC>[33m`, only the `<ESC>[` is being stripped, leaving the `33m` behind.

According to wikipedia:

> Escape sequences start with the character ESC (ASCII decimal 27/hex 0x1B/octal 033). For two character sequences, the second character is in the range ASCII 64 to 95 (@ to _). However, most of the sequences are more than two characters, and start with the characters ESC and [ (left bracket). This sequence is called CSI for Control Sequence Introducer (or Control Sequence Initiator). The final character of these sequences is in the range ASCII 64 to 126 (@ to ~).

The problem with the current logic is that it's missing the middle bit about the second character either being ASCII 64 to 95 or 96 (`[`), and just looks for the final character in range 64 to 126, so for > 2 character escapes, it will never strip the CSI.

jsuereth:
Now we handle the CSI (ESC + [).

Fixes #1143 

Review by @eed3si9n

### DIFF
--- a/util/log/src/main/scala/sbt/ConsoleLogger.scala
+++ b/util/log/src/main/scala/sbt/ConsoleLogger.scala
@@ -31,11 +31,40 @@ object ConsoleLogger {
   /**
    * An escape terminator is a character in the range `@` (decimal value 64) to `~` (decimal value 126).
    * It is the final character in an escape sequence.
+   *
+   * cf. http://en.wikipedia.org/wiki/ANSI_escape_code#CSI_codes
    */
+  @deprecated("No longer public.", "0.13.8")
   def isEscapeTerminator(c: Char): Boolean =
     c >= '@' && c <= '~'
 
-  /** Returns true if the string contains the ESC character. */
+  /**
+   * Test if the character AFTER an ESC is the ANSI CSI.
+   *
+   * see: http://en.wikipedia.org/wiki/ANSI_escape_code
+   *
+   * The CSI (control sequence instruction) codes start with ESC + '['.   This is for testing the second character.
+   *
+   * There is an additional CSI (one character) that we could test for, but is not frequnetly used, and we don't
+   * check for it.
+   *
+   * cf. http://en.wikipedia.org/wiki/ANSI_escape_code#CSI_codes
+   */
+  private def isCSI(c: Char): Boolean = c == '['
+
+  /**
+   * Tests whether or not a character needs to immediately terminate the ANSI sequence.
+   *
+   * c.f. http://en.wikipedia.org/wiki/ANSI_escape_code#Sequence_elements
+   */
+  private def isAnsiTwoCharacterTerminator(c: Char): Boolean =
+    (c >= '@') && (c <= '_')
+
+  /**
+   * Returns true if the string contains the ESC character.
+   *
+   * TODO - this should handle raw CSI (not used much)
+   */
   def hasEscapeSequence(s: String): Boolean =
     s.indexOf(ESC) >= 0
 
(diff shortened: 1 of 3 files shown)


util/log/src/main/scala/sbt/ConsoleLogger.scala
```diff
@@ -35,6 +35,19 @@ object ConsoleLogger {
   def isEscapeTerminator(c: Char): Boolean =
     c >= '@' && c <= '~'
 
+  /**
+   * Test if the character AFTER an ESC is the ANSI CSI, which means more than two characters until terminatioin.
```
eed3si9n: See http://en.wikipedia.org/wiki/ANSI_escape_code#Sequence_elements

> There is a single-character CSI (`155`/`0x9B` /`0233`) as well. The `ESC[` two-character sequence is more often used than the single-character alternative ...
> Only the two-character sequence is recognized by devices that support just ASCII (7-bit bytes) or devices that support 8-bit bytes, but use the 0x80–0x9F control character range for other purposes. On terminals that use UTF-8 encoding, both forms take 2 bytes (CSI in UTF-8 is `0xC2`, `0x9B`) but the `ESC[` sequence is clearer.

util/log/src/main/scala/sbt/ConsoleLogger.scala
```diff
@@ -35,6 +35,19 @@ object ConsoleLogger {
...
+   */
+  def isCSI(c: Char): Boolean = c == '['
+
+  /**
+   * Tests whether or not a character needs to immediately terminate the ANSI sequence.
+   */
+  def isSimpleEscapeTerminator(c: Char): Boolean = {
+    val n = c.toInt
+    (n >= 64) && (n <= 95)
```
eed3si9n: Citation needed. http://en.wikipedia.org/wiki/ANSI_escape_code#CSI_codes

> The final byte is technically any character in the range `64`–`126` (hex `0x40`–`0x7E`, ASCII `@` to `~`)

util/log/src/main/scala/sbt/ConsoleLogger.scala
jsuereth: Right.  We never parsed that before, at all.   I also don't think we use it.

This fixes the initial bug and invalid handling if ESC[. We need to improve this anyway for the request about filtering color escapes separately from cursor control.   So I think this PR is ok as is.

util/log/src/main/scala/sbt/ConsoleLogger.scala
jsuereth: Not for 2 character codes

eed3si9n: Link?

jsuereth: The same.  Read further

eed3si9n: http://en.wikipedia.org/wiki/ANSI_escape_code#Sequence_elements

> Escape sequences start with the character ESC (ASCII decimal 27/hex 0x1B/octal 033). For two character sequences, the second character is in the range ASCII 64–95 (@ to _).

ok. So this is non-CSI case. Having `isEscapeTerminator` and `isSimpleEscapeTerminator` without any explanation is a bit confusing. Also these are all public functions?

jsuereth: Also github mobile interface stinks.

Your link is just CSI escape codes.  Check above in the doc for non CSI escapes

eed3si9n: Also would it make sense to write `'@'` instead of `64` to be consistent with the `isEscapeTerminator` above?

eed3si9n: > Check above in the doc for non CSI escapes

Yea. I now understand the code is valid. Now just nitpicking on readability.

jsuereth: Yeah, all good points. I'll add the references and more description of why.

TBH - i'm really not sure what the "known escapes" are from in the tests, still.  They're definitely not ANSI, but there must be a reason we try to escape them.

eed3si9n:
Travis is reporting failure on `dependency-management/metadata-only-resolver`.

eed3si9n:
Notes plz

jsuereth:
@eed3si9n notes added, cleanups made.  Need to see if travis succeeds on unrelated tests....

jsuereth:
@eed3si9n The failure is disconcerting.  We're testing whether we find the JAR correctly to resolve the pom.  Maven-central only has a pom, but anotehr repo has the JAR, so we need to fail to resolve until the other resolver is in place.  But that repo may be gone:

```
 Server access Error: Connection timed out url=http://maven.nuxeo.org/nexus/content/groups/public/com/sun/jmx/jmxri/1.2.1/jmxri-1.2.1.jar
[info] [warn]   [
```

jsuereth:
@eed3si9n failure is due to publishLocal.  I know you just fixed this, so am I not up-to-date with master?  I just did a rebase, but it still seems bad.  Ideas welcome.

eed3si9n:
It seems like the test fails at some probability. Local test and #1879 show that on their own these tests are passing ok (I just checked running https://travis-ci.org/sbt/sbt/builds/52092661 now again).

PR #1887, opened 2015-02-28
Fix metadataonly tests.

jsuereth:
Turns out the repository we were relying on existing disappeared.  Using a fake jar/pom does the same trick.

review by @eed3si9n


eed3si9n:
Let's see if this passes.

PR #1893, opened 2015-03-01
Fixes #1881. Clone SystemProperties defending against ConcurrentModificationException

### LINKED ISSUE #1881: Sporadic ConcurrentModificationException from JUnitXmlTestsListener
In sbt 0.13.7 we are seeing a sporadic `ConcurrentModificationException`:

```
java.util.ConcurrentModificationException
    at java.util.Hashtable$Enumerator.next(Hashtable.java:1367)
    at sbt.JUnitXmlTestsListener.<init>(JUnitXmlTestsListener.scala:33)
    at sbt.plugins.JUnitXmlReportPlugin$$anonfun$projectSettings$1.apply(JUnitXmlReportPlugin.scala:27)
    at sbt.plugins.JUnitXmlReportPlugin$$anonfun$projectSettings$1.apply(JUnitXmlReportPlugin.scala:27)
    at sbt.$tilde$greater$$anonfun$$u2219$1.apply(TypeFunctions.scala:40)
    at sbt.std.Transform$$anon$4.work(System.scala:63)
    at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:226)
    at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:226)
    at sbt.ErrorHandling$.wideConvert(ErrorHandling.scala:17)
    at sbt.Execute.work(Execute.scala:235)
    at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:226)
    at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:226)
    at sbt.ConcurrentRestrictions$$anon$4$$anonfun$1.apply(ConcurrentRestrictions.scala:159)
    at sbt.CompletionService$$anon$2.call(CompletionService.scala:28)
    at java.util.concurrent.FutureTask.run(FutureTask.java:266)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
    at java.util.concurrent.FutureTask.run(FutureTask.java:266)
```

Examining the code in JUnitXmlTestsListener, this seems likely to be genuine concurrent access from another test.

It could be fixed by creating a defensive clone of the system properties.  The safest way I can think of to do this would be:

``` scala
diff --git a/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala b/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
index 2611486..5524e2f 100644
--- a/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
+++ b/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
@@ -27,7 +27,8 @@ class JUnitXmlTestsListener(val outputDir: String) extends TestsListener {
   val properties =
     <properties>
       {
-        val iter = System.getProperties.entrySet.iterator
+        // create a clone, defending against [[ConcurrentModificationException]]
+        val iter = System.getProperties.clone.asInstanceOf[util.Hashtable[AnyRef, AnyRef]].entrySet.iterator
         val props: ListBuffer[XNode] = new ListBuffer()
         while (iter.hasNext) {
           val next = iter.next
```

Note that using `sys.props.toList` or any variant of this will suffer from the same issue as the current code, specifically that it will use an iterator on the underlying (mutable) entry set.  The clone method on the other hand is synchronized (if clunky looking).

aerskine:


lightbend-cla-validator:
Hi @aerskine,

typesafe-tools:

jsuereth:
LGTM

som-snytt:
Alternatively, 

https://github.com/scala/scala/blob/v2.11.5/src/compiler/scala/tools/reflect/WrappedProperties.scala#L34

```
props.stringPropertyNames().asScala.toList.map(k => (k, props.get(k).asInstanceOf[String]))
```

aerskine:
The version in the Scala compiler could use a fix too -- it is robust against ConcurrentModificationException, but not against concurrent modification, as the property could be removed before the call to `props.get(k)`

som-snytt:
Apparently, it doesn't care about null values. Actually, let me check:

https://github.com/scala/scala/blob/2.11.x/src/compiler/scala/tools/util/PathResolver.scala#L55

and it doesn't convert the null...

I'd forgotten that I'd noted a code comment on the PR:

```
  // Todo: bring some sanity to the intersection of system properties aka "mutable
  // state shared by everyone and everything" and the reality that there is no other
  // mechanism for accomplishing some things on the jvm.
```

aerskine:
Yes, I'd noted some other places where SI-7269 was cited.  There's no great solution as long as sys.props models the mutability of the underlying properties.

eed3si9n:
Could you add a note in format similar to https://raw.githubusercontent.com/sbt/sbt/0.13/notes/0.13.8/ansi-escapes.markdown?
Otherwise LGTM.

aerskine:
Added.

PR #1899, opened 2015-03-04
Exempt scala-actors-migration from version checks.

### LINKED ISSUE #1818: sbt fails to correctly determine binary compatibility for scala-actors-migration
I am trying to get rid of a warning I am getting when using the SBT plugin in Intellij.

```
[warn] Binary version (1.1.0) for dependency org.scala-lang#scala-actors-migration_2.11;1.1.0
[warn] in org.company.package#rd$sources_javadoc_2.11;1.0 differs from Scala binary version in project (2.11).
```

I am still researching the problem, but I just wanted to put this up quickly and see if anyone has any idea what's going on. Judging by the patterns `CrossversionUtil.scalaApiVersion` it seems that SBT always expects the version to contain the full semantic version as a substring of the dependency name but in this case we seem to oddly have `2.11;1.1.0`. Any ideas?

dwijnand:
Fixes #1818.


typesafe-tools:

eed3si9n:
I feel like we should flip the logic and check only for scala-compiler, scala-library, etc - http://mvnrepository.com/artifact/org.scala-lang

dwijnand:
Sure, but isn't false positive warnings better than no warning and repercussion failures? I'm not sure, but I'm happy to do it the other way.

jsuereth:
yeah, it's safer to have false positives. e.g. scala-reflect showed up one day and had to be included.   THe scala team is now aware of our convention (or at least were at some point) hence everything moving into org.scala-lang.modules.  I'd rather special case the outlier to the rule than prevent flexibility in design in the future.

jsuereth:
this LGTM

eed3si9n:
If we're not flipping the logic, could we also add the "scala-pickling" to the known bad list (even though the latest Pickling has moved to org.scala-lang.modules)?

dwijnand:
Absolutely.

PR #1902, opened 2015-03-06
Define Project.settings with SettingsDefinition.

dwijnand:
This maintains old, existing ways of configurating project settings,
while adding the possibility to:
- Define add a sequence of settings, between other individual settings
- Pass a Seq[Setting[_]] to Project.settings without having to do
  "varargs expansion", ie. ": _*"

I haven't checked if this is binary backwards-compatible, but I also don't know
what sbt requirements on binary-compatibility are, but it seems to be
source-compatible with existing ways of adding settings to a project.

Please let me know if you know of other common ways I missed.

Also, modulo Travis highlighting anything.


typesafe-tools:

dwijnand:
Oh.. notes. I guess I'll wait for Travis and feedback, and if it looks good I'll push up some notes for review. My bad.

eed3si9n:
The first rule of sbt pull request is that you can't break binary compatibility with the first release within the series, in this case 0.13.0. This rule is often shorted to "bincompat."
Your change might maintain source compatibility for builds, but if any plugin relied on `def settings(ss: Setting[_]*)`, even indirectly, it will be broken.

This is why I mentioned overloading `settings(any: Any)` since you can't have two varargs, but you _might_ have a better luck with `Any` backed with a macro.

dwijnand:
Hmm, I thought of settings in plugins, but a part of me thought that Project.settings isn't typically invoked by plugins, but I guess that's too fast and loose.

I'll try the macro route, and see if I can manage. Thanks for your comments Eugene.

dwijnand:
I don't know the full ins and outs of bincompat, but I tried to simulate what this change would mean to a project using a "Project.settings-using-plugin" and, to my amateurial attempt maybe, it seems to hold... I pushed up my experiment here: https://github.com/dwijnand/sbt8bincompat/commits/master
- [a995929](https://github.com/dwijnand/sbt8bincompat/commit/a995929c6accd8790bc13a8353fe82af4410121f) Create a 0.13.7 plugin (`sbt8bincompat-plugin`) which uses `Project.settings` in different ways, setting `description`
- [2f6b60a](https://github.com/dwijnand/sbt8bincompat/commit/2f6b60af5ab9cef495de47daa3e7fe50d97eb50d) Create a 0.13.7 test project (`sbt8bincompat-testit`) which uses `sbt8bincompat-plugin`\- now the plugin uses `Project.settings` to define `description`
- (Separately) `publishLocal` this pull-request branch (commit [a058a38](https://github.com/dwijnand/sbt/commit/a058a38b7b7a4b3959be085b3718ff7afc58bbb3)) as 0.13.8-SNAPSHOT, and `rm -r ~/.sbt/boot`
- [98c0e91](https://github.com/dwijnand/sbt8bincompat/commit/98c0e91bfbe4dec5d279bca0a1c4392ee153591b) Switch `sbt8bincompat-testit` to 0.13.8-SNAPSHOT, launch with

```
sbt -sbt-jar ~/.ivy2/local/org.scala-sbt/sbt-launch/0.13.8-SNAPSHOT/jars/sbt-launch.jar
```

   it loads successfull!! and `description` is still set by the 0.13.7-built plugin!!
- [2401581](https://github.com/dwijnand/sbt8bincompat/commit/2401581d4542f3425d6c11c0948989711400268f) As a sanity check, check that Project.settings can take `Seq[Setting[_]]` directly, like the scripted test, by setting `startYear` is all sorts of ways, and it compiles and loads and returns the last winning declaration (`Some(2014)`).

It must be how varargs are encoded but it seems to survive switching from `Setting[_]*` to `SettingsDefinition*`. I feel like I've given this a fair chance to fail, but can't find a way. Please can you sanity check if I've done anything wrong anywhere?

If despite this you still would rather it be a macro, that's fine. I just wanted to post my findings in case it changed your mind.

jsuereth:
Project/build.properties determines the sbt version used, not just the launcher.  Was there one in your test project?

dwijnand:
Yep [switched to 0.13.8-SNAPSHOT](https://github.com/dwijnand/sbt8bincompat/commit/98c0e91bfbe4dec5d279bca0a1c4392ee153591b)

dwijnand:
I guess I could've just used the 0.13.7 launcher, but anyways..

eed3si9n:
ok. I stand corrected on the binary compatibility. `Setting[T]` actually extends `SettingsDefinition`:

``` scala
  sealed class Setting[T] private[Init] (val key: ScopedKey[T], val init: Initialize[T], val pos: SourcePosition) extends SettingsDefinition {
    def settings = this :: Nil
    ...
```

So even if the plugins pass in a `Seq` of `Setting`, it should continue to work as a `Seq` of `SettingsDefinition`. This is looking good to me. @jsuereth?

jsuereth:
Ooh, i do kind of like the trick you're using.  It _is_ BC thanks to how the JVM can migrate methods up/down the class hierarchy when linking.  My only concern is odd interactions w/ type inference, but given that all our test suite passes, I'm willing to give this a shot.

dwijnand:
I can make a point of testing the RC1 on the suite of sbt builds in the company I work for and see if anything flags or fails.

eed3si9n:
Merged.

Issue #1900, opened 2015-03-06
More robustness in help and completion system


DavidPerezIngeniero:
It's a multi-module project with a `build.sbt` and some custom `project/*.scala` files.
Version of SBT: 0.13.7.

```
> help
java.util.NoSuchElementException: key not found: -capture-settings
        at scala.collection.MapLike$class.default(MapLike.scala:228)
        at scala.collection.AbstractMap.default(Map.scala:58)
        at scala.collection.MapLike$class.apply(MapLike.scala:141)
        at scala.collection.AbstractMap.apply(Map.scala:58)
        at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
        at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
        at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
        at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:47)
        at scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
        at scala.collection.AbstractTraversable.map(Traversable.scala:105)
        at sbt.BuiltinCommands$.allTaskAndSettingKeys(Main.scala:180)
        at sbt.BuiltinCommands$.keysHelp(Main.scala:386)
        at sbt.BuiltinCommands$$anonfun$actHelp$1.apply(Main.scala:383)
        at sbt.BuiltinCommands$$anonfun$actHelp$1.apply(Main.scala:383)
        at sbt.BasicCommands$$anonfun$1.apply(BasicCommands.scala:30)
        at sbt.BasicCommands$$anonfun$1.apply(BasicCommands.scala:30)
        at scala.collection.LinearSeqOptimized$class.foldLeft(LinearSeqOptimized.scala:111)
        at scala.collection.immutable.List.foldLeft(List.scala:84)
        at scala.collection.TraversableOnce$class.$div$colon(TraversableOnce.scala:138)
        at scala.collection.AbstractTraversable.$div$colon(Traversable.scala:105)
        at sbt.BasicCommands$.helpParser(BasicCommands.scala:30)
        at sbt.BasicCommands$$anonfun$help$1.apply(BasicCommands.scala:26)
        at sbt.BasicCommands$$anonfun$help$1.apply(BasicCommands.scala:26)
        at sbt.Command$$anonfun$sbt$Command$$argParser$1.apply(Command.scala:77)
[...]

PR #1916, opened 2015-03-12
Issue 1900, more robustness in help system

DavidPerezIngeniero:
If some ill-behaved command throws an exception when calling the brief method, the help system continues working.
https://github.com/sbt/sbt/issues/1900

### DIFF
--- a/main/command/src/main/scala/sbt/BasicCommands.scala
+++ b/main/command/src/main/scala/sbt/BasicCommands.scala
@@ -13,6 +13,8 @@ import BasicKeys._
 
 import java.io.File
 
+import scala.util.control.NonFatal
+
 object BasicCommands {
   lazy val allBasicCommands = Seq(nop, ignore, help, completionsCommand, multi, ifLast, append, setOnFailure, clearOnFailure, stashOnFailure, popOnFailure, reboot, call, early, exit, continuous, history, shell, read, alias) ++ compatCommands
 
@@ -29,7 +31,7 @@ object BasicCommands {
     {
       val h = (Help.empty /: s.definedCommands) { (a, b) =>
         a ++
-          (try b.help(s) catch { case ex: Throwable => Help.empty })
+          (try b.help(s) catch { case NonFatal(ex) => Help.empty })
       }
       val helpCommands = h.detail.keySet
       val spacedArg = singleArgument(helpCommands).?
@@ -41,7 +43,7 @@ object BasicCommands {
       val message = try
         Help.message(h, arg)
       catch {
-        case ex: Throwable =>
+        case NonFatal(ex) =>
           ex.toString
       }
       System.out.println(message)
--- a/main/command/src/main/scala/sbt/BasicCommands.scala
+++ b/main/command/src/main/scala/sbt/BasicCommands.scala
@@ -27,7 +27,7 @@ object BasicCommands {
 
   def helpParser(s: State) =
     {
-      val h = (Help.empty /: s.definedCommands) { (a, b) ⇒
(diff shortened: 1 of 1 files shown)


typesafe-tools:

DavidPerezIngeniero:
It can be extended also to the completion system.

jsuereth:
Nice, I like the change!

jsuereth:
LGTM cc> @eed3si9n

main/command/src/main/scala/sbt/BasicCommands.scala
```diff
@@ -27,7 +27,10 @@ object BasicCommands {
 
   def helpParser(s: State) =
     {
-      val h = (Help.empty /: s.definedCommands)(_ ++ _.help(s))
+      val h = (Help.empty /: s.definedCommands) { (a, b) ⇒
+        a ++
+          (try b.help(s) catch { case ex: Throwable ⇒ Help.empty })
```
pauldraper: Maybe this should be `ex: Exception` or `NotFatal(e)` instead?

DavidPerezIngeniero:
And also:

``` scala
(try b.help(s) catch { case ex: Throwable ⇒ Help.empty })
```

The message of the exception should be shown, instead of silent..

main/command/src/main/scala/sbt/BasicCommands.scala
jsuereth: yeah, maybe use `NonFatal(e)` so you don't catch OOM/Interrupted errors that would have to be handled differently.

main/command/src/main/scala/sbt/BasicCommands.scala
```diff
@@ -27,7 +27,10 @@ object BasicCommands {
 
   def helpParser(s: State) =
     {
-      val h = (Help.empty /: s.definedCommands)(_ ++ _.help(s))
+      val h = (Help.empty /: s.definedCommands) { (a, b) ⇒
```
eed3si9n: We don't use Unicode arrow in our code base. Could you switch to `=>`?

main/command/src/main/scala/sbt/BasicCommands.scala
eed3si9n: Any update on this?

DavidPerezIngeniero:
Removed unicode arrow.  Do I have to repeat the pull request?

jsuereth:
no, it showed up.

DavidPerezIngeniero:
Removed more unicode arrows...

eed3si9n:
Is catching `Throwable` vs [`NonFatal`](http://www.scala-lang.org/api/2.10.4/index.html#scala.util.control.NonFatal$) still not settled? I think we're worried things like `InterruptedException` getting caught without proper propagation.

DavidPerezIngeniero:
Done.  Thanks to your comment I've learnt about `NonFatal`.

eed3si9n:
Awesome. The code LGTM.
Could you rebase against 0.13, squash all commits, and add release note entry to https://github.com/sbt/sbt/blob/0.13/notes/0.13.9.markdown please? Then I'll merge it once Travis goes through. Thanks for your contribution.

eed3si9n:
I just noticed that this PR is going against `sbt:0.13.7`. Please open one against 0.13.

DavidPerezIngeniero:
Created another pull request, based on 0.13.

eed3si9n:
Closing this in favor of #1940.
Also good news is that Travis tests passed - https://travis-ci.org/sbt/sbt/builds/55801662

PR #1923, opened 2015-03-17
Warn when ignoring previous analysis for various reasons.

benjyw:
This can confuse users and is hard to debug. It seems better to be
explicit about why we're choosing not to do an incremental compile.

### DIFF
--- a/compile/integration/src/main/scala/sbt/compiler/AggressiveCompile.scala
+++ b/compile/integration/src/main/scala/sbt/compiler/AggressiveCompile.scala
@@ -149,9 +149,12 @@ class AggressiveCompile(cacheFile: File) {
           // previous Analysis completely and start with empty Analysis object
           // that supports the particular value of the `nameHashing` flag.
           // Otherwise we'll be getting UnsupportedOperationExceptions
+          log.warn("Ignoring previous analysis due to incompatible nameHashing setting.")
           Analysis.empty(currentSetup.nameHashing)
         case Some(previous) if equiv.equiv(previous, currentSetup) => previousAnalysis
-        case _ => Incremental.prune(sourcesSet, previousAnalysis)
+        case _ =>
+          log.warn("Pruning sources from previous analysis, due to incompatible CompileSetup.")
+          Incremental.prune(sourcesSet, previousAnalysis)
       }
       IncrementalCompile(sourcesSet, entry, compile0, analysis, getAnalysis, output, log, incOptions)
     }


typesafe-tools:

gkossakowski:
It looks to me that the failure is spurious. I restarted the build.

compile/integration/src/main/scala/sbt/compiler/AggressiveCompile.scala
```diff
@@ -149,9 +149,12 @@ class AggressiveCompile(cacheFile: File) {
           // previous Analysis completely and start with empty Analysis object
           // that supports the particular value of the `nameHashing` flag.
           // Otherwise we'll be getting UnsupportedOperationExceptions
+          log.warn("Ignoring previous analysis due to incompatible nameHashing setting.")
```
gkossakowski, line 152: Agreed that this is not expected to happen commonly so we should warn.

compile/integration/src/main/scala/sbt/compiler/AggressiveCompile.scala
```diff
@@ -149,9 +149,12 @@ class AggressiveCompile(cacheFile: File) {
...
           Analysis.empty(currentSetup.nameHashing)
         case Some(previous) if equiv.equiv(previous, currentSetup) => previousAnalysis
-        case _ => Incremental.prune(sourcesSet, previousAnalysis)
+        case _ =>
+          log.warn("Pruning sources from previous analysis, due to incompatible CompileSetup.")
+          Incremental.prune(sourcesSet, previousAnalysis)
```
gkossakowski, line 157: However, I'm not sure about that one. Did you check that changes to `CompileSetup` are always triggered by explicit user action or some other type of infrequent event?

benjyw, line 157: This is actually the one that bites us the most. Changes to CompileSetup happen e.g., when we upgrade a scalac plugin or change memory options. It's not at all obvious that this will invalidate your analysis and trigger a full compile, so we think it's worth a warning.

gkossakowski, line 157: I see. I wonder if this should be info/debug level message. It feels like warning is a bit too strong for an action that is not dangerous and works as designed. WDYT?

benjyw, line 157: In our case it can be dangerous because of older analysis files from cached artifacts. But "works as designed" is debatable: it really depends on what the compiler arguments are. For example, why should changing the heap size invalidate your analysis? That's extremely non-obvious, I think?

gkossakowski, line 157: Agreed.

gkossakowski:
LGTM

I'll let sbt guys click the merge button.

jsuereth:
LGTM too.

benjyw:
Thanks all!

PR #1929, opened 2015-03-20
Make Initialize[T] covariant

pauldraper:
Allows for build definitions like this:

```
import sbt._

class Foo

class Bar extends Foo

object AppBuild extends Build {
    private val fooKey = SettingKey[Foo]("foo")

    private val fooSetting = Def.setting(new Bar)

    lazy val root = (project in file(".")).settings(
        fooKey <<= fooSetting
    )
}
```

which currently does not compile.

### DIFF
--- a/util/collection/src/main/scala/sbt/Settings.scala
+++ b/util/collection/src/main/scala/sbt/Settings.scala
@@ -410,7 +410,7 @@ trait Init[Scope] {
       allDefs flatMap { case d: DerivedSetting[_] => (derivedToStruct get d map (_.outputs)).toStream.flatten; case s => Stream(s) }
     }
 
-  sealed trait Initialize[T] {
+  sealed trait Initialize[+T] {
     def dependencies: Seq[ScopedKey[_]]
     def apply[S](g: T => S): Initialize[S]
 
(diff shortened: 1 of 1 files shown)


typesafe-tools:

util/collection/src/main/scala/sbt/Settings.scala
```diff
@@ -589,7 +589,7 @@ trait Init[Scope] {
       case Some(i) => Right(new Optional(i.validateKeyReferenced(g).right.toOption, f))
     }
     def mapConstant(g: MapConstant): Initialize[T] = new Optional(a map mapConstantT(g).fn, f)
-    def evaluate(ss: Settings[Scope]): T = f(a.flatMap(i => trapBadRef(evaluateT(ss)(i))))
+    def evaluate(ss: Settings[Scope]): T = f(a.flatMap(i => trapBadRef(evaluateT(ss)(i))): Option[Id[S]])
```
pauldraper, line 592: This was necessary for compilation (though I'm not quite sure why).

jsuereth:
Could you add some tests exercising the new behavior?

Also, the `<<=` isn't something we'd really like to continue to propagate, I think, but cc'ing @eed3si9n for his thoughts.

pauldraper:
Ah, well yes, if `<<=` is on its way out, then this is unnecessary.

(Personally, I've always preferred `<<=` over `:=` because I don't have to keep track of whether I'm in the right context, or using dynamic reference, etc. which can be pretty confusing. But I know I'm mostly alone in that preference. And fewer symbols is always nice.)

eed3si9n:
Yea, as we reduce the number of DSL operators, `<<=` would likely be on consideration for getting chopped. It can fully be replaced using `:= x.value` right?
Also I'd be really careful of adding any variance, since it expands the area of what we have to reason about in our head. If it's just for type inference, I don't think it's worth it.

pauldraper:
> If it's just for type inference, I don't think it's worth it.

It's not for type inference per se; it's for type checking.

For example, if I wanted to use `fooSetting` for two keys:

``` scala
val fooSetting = Def.setting[Bar](new Bar)
...
SettingKey[Foo]("foo") <<= fooSetting
SettingKey[Bar]("bar") <<= fooSetting
```

Variance is _really_ nice; you'd notice very quickly if `Option` or `Function2` were not variant.

---

But yes `:= x.value` is (AFAIK) a full replacement for `<<= x`. Deprecating `<<=` is a compelling reason to close this.

PR #1944, opened 2015-03-26
Wip/fix aether range query

jsuereth:
Use the VersionRangeQuery API on aether to handle any versions that are ranges (like `1.+`, `[2.0,3.0)`).  Note:  Some ivy ranges are _not_ acceptible aether ranges.  We may be able to smooth that out later, for now we disallow them completely.


jsuereth:
REview by @eed3si9n

eed3si9n:
This is a plugin, so if the test passes, I'm fine with merging it.

PR #1946, opened 2015-03-27
Rename root project to sbtRoot.

dwijnand:
This is mostly for IntelliJ IDEA.

Currently IntelliJ IDEA's Scala (and SBT) plugin defines:
- the project name (as seen in the window title and in the "open recent
  project" list) from `name`
- the root module (as seen in the project view and in project structure)
  from `id`
- doesn't use `moduleName` at all

After this change the sbt project is no longer identified as "root".

I was undecided between `sbtRoot` and `sbtRootProj`, and went with the
shorter option. I'm happy to revise this decision.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -39,7 +39,7 @@ def baseSettings: Seq[Setting[_]] =
 def testedBaseSettings: Seq[Setting[_]] =
   baseSettings ++ testDependencies
 
-lazy val root: Project = (project in file(".")).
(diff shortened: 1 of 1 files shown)


dwijnand:
Oh! Now that I think of it, as the sbt module is defined as `sbtProj` I guess `sbt` is also an option..

I've submitted now, so I'll leave it as is and wait for feedback.

typesafe-tools:

eed3si9n:
LGTM

Issue #1924, opened 2015-03-18
Allow test frameworks implementations to participate in test discovery


michaelpigg:
The current approach to test discovery makes it difficult to use a test framework like partest from sbt. Partest defines test names with a combination of directory and file names rather than with type features of the test class.

My initial thought is to add a `enumerateTests` method to [Framework](https://github.com/sbt/test-interface/blob/master/src/main/java/sbt/testing/Framework.java) that would be called during discovery to allow the implementation to provide a set of test names.

dwijnand:
Closing. See #1949.

PR #1949, opened 2015-03-30
Allow test framework to provide test names

michaelpigg:
This is a working approach to enable #1924. It redefines the discovery process in `Tests` to also call a method `enumerateTests` on the `Framework` if it is implemented. The framework returns an `Array[String]` of test names when it implements this method. Each returned name is used to create a corresponding `TestDefinition`.

Implementing this change in scala-partest-interface (https://github.com/scala/scala-partest-interface/pull/11) allows me to
- Provide partest test categories as test names
- Take advantage of testForkedParallel to run test categories in parallel
- Specify a subset of test categories with `test-only`

This doesn't fix the related issue that many bits of code in the test framework assume that only `AnnotatedFingerprint` and `SubclassFingerprint` are the only known types. However, that is probably a separate issue.

### DIFF
--- a/main/actions/src/main/scala/sbt/Tests.scala
+++ b/main/actions/src/main/scala/sbt/Tests.scala
@@ -266,8 +266,21 @@ object Tests {
     }
   def overall(results: Iterable[TestResult.Value]): TestResult.Value =
     (TestResult.Passed /: results) { (acc, result) => if (acc.id < result.id) result else acc }
-  def discover(frameworks: Seq[Framework], analysis: Analysis, log: Logger): (Seq[TestDefinition], Set[String]) =
-    discover(frameworks flatMap TestFramework.getFingerprints, allDefs(analysis), log)
+  def discover(frameworks: Seq[Framework], analysis: Analysis, log: Logger): (Seq[TestDefinition], Set[String]) = {
+    val (discoveredTestDefinitions, mains) = discover(frameworks flatMap TestFramework.getFingerprints, allDefs(analysis), log)
+    val enumeratedTestDefinitions = enumerate(frameworks)
+    (discoveredTestDefinitions ++ enumeratedTestDefinitions, mains)
+  }
+
+  def enumerate(frameworks: Seq[Framework]): Seq[TestDefinition] = {
+    val enumeratingFrameworks = frameworks.map(f => (f, scala.util.Try(f.getClass.getMethod("enumerateTests")))).filter(frameworkAndMethod => frameworkAndMethod._2.isSuccess)
+    enumeratingFrameworks.flatMap(frameworkAndMethod => frameworkAndMethod._2.get.invoke(frameworkAndMethod._1) match {
+      case names: Array[String] => names.map(name => {
+        new TestDefinition(name, frameworkAndMethod._1.fingerprints.head, true, Array(new SuiteSelector))
+      }).toSeq
+      case _ => Seq.empty[TestDefinition]
+    })
+  }
 
   def allDefs(analysis: Analysis) = analysis.apis.internal.values.flatMap(_.api.definitions).toSeq
   def discover(fingerprints: Seq[Fingerprint], definitions: Seq[Definition], log: Logger): (Seq[TestDefinition], Set[String]) =


lightbend-cla-validator:
Hi @michaelpigg,

typesafe-tools:

main/actions/src/main/scala/sbt/Tests.scala
```diff
@@ -266,8 +266,21 @@ object Tests {
...
-  def discover(frameworks: Seq[Framework], analysis: Analysis, log: Logger): (Seq[TestDefinition], Set[String]) =
-    discover(frameworks flatMap TestFramework.getFingerprints, allDefs(analysis), log)
+  def discover(frameworks: Seq[Framework], analysis: Analysis, log: Logger): (Seq[TestDefinition], Set[String]) = {
+    val (discoveredTestDefinitions, mains) = discover(frameworks flatMap TestFramework.getFingerprints, allDefs(analysis), log)
+    val enumeratedTestDefinitions = enumerate(frameworks)
+    (discoveredTestDefinitions ++ enumeratedTestDefinitions, mains)
+  }
+
+  def enumerate(frameworks: Seq[Framework]): Seq[TestDefinition] = {
+    val enumeratingFrameworks = frameworks.map(f => (f, scala.util.Try(f.getClass.getMethod("enumerateTests")))).filter(frameworkAndMethod => frameworkAndMethod._2.isSuccess)
```
jsuereth, line 276: a second option here is to first update the test-interface JAr to have this method, and then invoke it here and catch the JVM link exception (so existing clients still work) rather than using reflection.

eed3si9n:
Are we adding a direct call in sbt to a specific test framework using reflection? If we accepted that as an option, what would prevent ScalaTest or Specs2 to add some other similar hooks?
Is there any way of evolving `test-interface` in a binary compatible way instead?

main/actions/src/main/scala/sbt/Tests.scala
michaelpigg, line 276: Ah, this is probably what you suggested when we discussed this at ScalaDays, but I did not fully understand. I will work on that option.

michaelpigg:
@eed3si9n This isn't intended to be a call to a specific test framework, but it is unique to a specific category of frameworks (those that can't use class-based mechanisms to enumerate tests).

eed3si9n:
The following is the signature of `Fingerprint`:

``` java
package sbt.testing;

/**
 * A way to identify test classes and/or modules that should
 * be discovered when the client performs discovery.
 */
public interface Fingerprint {}
```

It on its own doesn't say anything about the test being a file or a class, I think. Am I missing something?

jsuereth:
@eed3si9n Yeah.   A good exercise would be to go try to reimplement scripted so that it honors the test-interface.  We need a new hook in there so we can enumerate tests without `.class`

michaelpigg:
You're right that Fingerprint should allow for file-based tests. The issue is more that the code in sbt assumes that annotation and class fingerprints are the only types and fails with any other Fingerprint. I initially tried a new Fingerprint and quickly ran into that code. But maybe the better solution is to fix the code that cant handle arbitrary Fingerprints.

michaelpigg:
Do you think the Fingerprint should carry the test enumeration? Or should it serve more as a marker like the existing Fingerprints?

eed3si9n:
> You're right that Fingerprint should allow for file-based tests. The issue is more that the code in sbt assumes that annotation and class fingerprints are the only types and fails with any other Fingerprint.

I'm totally open to dealing with that part of the implementation. We can add a new kind of Fingerprint without breaking existing test frameworks.

> Do you think the Fingerprint should carry the test enumeration? Or should it serve more as a marker like the existing Fingerprints?

I'm not a test expert but I think the point of Fingerprint is to enumerate the tests:
1. The end user from sbt tries to hit tab on `testOnly`
2. Framework Foo returns the list of Fingerprints for the current project
3. sbt figures out how to print the names of the Fingerprints
4. The user picks one of the names
5.  Framework Foo interprets the Fingerprint as an actual test and runs it

It's just an incidental detail that traditional JUnit-like frameworks had a granularity being at the class level.

The `$kind/$testName` structure of scala-partest can be abstracted further to `$testGroup/$tesName` like scripted as @jsuereth suggested. We could either:
- Have two additional types of Fingerprint for testGroup and testName, and let the user type in either `neg` or `neg/foo`.
- Or, have just one for testName, but allow match by startsWith so `neg` matches all negative tests.

michaelpigg:
@eed3si9n It's not so clear to me what `Fingerprint` represents, but my vision is clouded by the current implementation. I do like the approach that you described, and it would certainly be cleaner. I had tried defining an `EnumerationFingerprint` which would simply supply an array of test names. Had I not run into the issues with new `Fingerprint`s, then I would not have made the reflection call to `enumerateTests` and would have used names provided in the fingerprint instead.

```
package sbt.testing;

public interface EnumerationFingerprint extends Fingerprint {
  public String[] testNames;
}
```

I think you're suggesting something more like the following, where a fingerprint is returned for each individual test?

```
package sbt.testing;

public interface TestNameFingerprint extends Fingerprint {
  public String testName;
}
```

eed3si9n:
Yes. I would add `public String testGroup` in there too.

michaelpigg:
Looking again, I think there's actually only one place that doesn't handle unknown Fingerprints well, which is ForkTests.scala line 109:

```
  private[this] def forkFingerprint(f: Fingerprint): Fingerprint with Serializable =
    f match {
      case s: SubclassFingerprint  => new ForkMain.SubclassFingerscan(s)
      case a: AnnotatedFingerprint => new ForkMain.AnnotatedFingerscan(a)
      case _                       => sys.error("Unknown fingerprint type: " + f.getClass)
    }
```

The other places where a `Fingerprint` is pattern matched appear to have reasonable defaults (at least don't throw an exception).

dwijnand:
@michaelpigg Are you still interested in this?

michaelpigg:
@dwijnand No, not interested now.

dwijnand:
Ok, not a problem. If there's any interest in the future this PR could inform them of the work you did.

Meanwhile I'll close this.

PR #1950, opened 2015-03-31
Add retrieveManagedSync

ajsquared:
Adds retrieveManagedSync key that, when set to true, enables synchronizing retrieved to the current build by removed unneeded files.

### DIFF
--- a/ivy/src/main/scala/sbt/IvyActions.scala
+++ b/ivy/src/main/scala/sbt/IvyActions.scala
@@ -31,7 +31,9 @@ final class UpdateConfiguration(val retrieve: Option[RetrieveConfiguration], val
     logging: UpdateLogging.Value = this.logging): UpdateConfiguration =
     new UpdateConfiguration(retrieve, missingOk, logging)
 }
-final class RetrieveConfiguration(val retrieveDirectory: File, val outputPattern: String, val sync: Boolean)
+final class RetrieveConfiguration(val retrieveDirectory: File, val outputPattern: String, val sync: Boolean) {
+  def this(retrieveDirectory: File, outputPattern: String) = this(retrieveDirectory, outputPattern, false)
+}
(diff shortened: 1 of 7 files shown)


typesafe-tools:

ajsquared:
Pushed a fix for the test that did not compile

ajsquared:
I'm trying to reproduce the failures from CI, but am having difficulty doing so.

ivy/src/main/scala/sbt/IvyActions.scala
```diff
@@ -31,7 +31,7 @@ final class UpdateConfiguration(val retrieve: Option[RetrieveConfiguration], val
     logging: UpdateLogging.Value = this.logging): UpdateConfiguration =
     new UpdateConfiguration(retrieve, missingOk, logging)
 }
-final class RetrieveConfiguration(val retrieveDirectory: File, val outputPattern: String)
+final class RetrieveConfiguration(val retrieveDirectory: File, val outputPattern: String, val sync: Boolean)
```
jsuereth: Given this is a binary incompatible change, you'll ahve to add a new constructor with the old patterns and instantiates a default sync value for boolean.

ajsquared: Okay, I will add that in.

eed3si9n:
I've been restarting Travis CI jobs to clear out the transient failures. There's still one more pending but we should get green soon.
Otherwise LGTM.

jsuereth:
Also, looks good to me.

ajsquared:
@eed3si9n It looks like the `dependency-management/eviction-warning` scripted test failed again in `mavenResolverPluginTest:scripted dependency-management/*1of2 project/transitive-plugins`.  I'm not able to reproduce this failure.

PR #1959, opened 2015-04-08
Only enable cross sources when multiple versions.

### LINKED ISSUE #1957: Auto-scala-bin-compat source folder probably doesn't make sense in Play projects
Since the upgrade to sbt version 0.13.8 an empty folder "app-2.11" is appearing in the root of my Play 2.3.8 application when I work with it in dev ("run") mode.

dwijnand:
The intent of the original feature/improvement was for projects with
multiple scala versions (`crossScalaVersions`) to automatically support
cross-version source directories (via `scalaBinaryVersion`).

Unfortunately for all cross-built projects configured for only one version
(2.11-only for instance) there were some noticeable effects with this
change, such as an extra `app-2.11` directory being created in a Play app.

As of this commit, only projects with multiple Scala versions will
support cross-version source directories, maintaining the above intent
and resolving the above effect.

The scripted test from the original pull request never explicitly
declared its `crossScalaVersions` and instead switched versions from the
shell to pickup sources at the correct path. It's now explicit about
which Scala versions it supports.
## 

Fixes #1957

cc @indrajitr as he authored the original change (https://github.com/sbt/sbt/pull/1799)

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -181,7 +181,7 @@ object Defaults extends BuildCommon {
     sourceManaged <<= configSrcSub(sourceManaged),
     scalaSource := sourceDirectory.value / "scala",
     javaSource := sourceDirectory.value / "java",
-    unmanagedSourceDirectories := makeCrossSources(scalaSource.value, javaSource.value, scalaBinaryVersion.value, crossPaths.value),
+    unmanagedSourceDirectories := makeCrossSources(scalaSource.value, javaSource.value, scalaBinaryVersion.value, crossPaths.value, crossScalaVersions.value),
     unmanagedSources <<= collectFiles(unmanagedSourceDirectories, includeFilter in unmanagedSources, excludeFilter in unmanagedSources),
     watchSources in ConfigGlobal <++= unmanagedSources,
     managedSourceDirectories := Seq(sourceManaged.value),
@@ -242,8 +242,8 @@ object Defaults extends BuildCommon {
     derive(scalaBinaryVersion := binaryScalaVersion(scalaVersion.value))
   ))
 
-  def makeCrossSources(scalaSrcDir: File, javaSrcDir: File, sv: String, cross: Boolean): Seq[File] = {
-    if (cross)
+  def makeCrossSources(scalaSrcDir: File, javaSrcDir: File, sv: String, cross: Boolean, svs: Seq[String]): Seq[File] = {
+    if (cross && svs.length > 1)
       Seq(scalaSrcDir.getParentFile / s"${scalaSrcDir.name}-$sv", scalaSrcDir, javaSrcDir)
     else
       Seq(scalaSrcDir, javaSrcDir)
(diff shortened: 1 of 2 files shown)


typesafe-tools:

dwijnand:
Let me know if this requires a test- it felt like a lot of work to setup a scripted test to show it _not_ supporting a `src/main/scala-2.11` directory. But I'll do it if you prefer.

Also, as I haven't talked this through with anyone yet I didn't create notes for it yet. If this is good / has iterated to the point of being good / etc, I'd be happy to create notes.

eed3si9n:
In my opinion this change is not needed because if you want to back out of the cross sources you should set `crossPaths` to `false`, which is symmetric to `target`.

dwijnand:
I don't have a personal issue with the problem itself, just noticed all the noise about it.

But switching off `crossPaths` can't be the way to back out of cross sources as it also switches off cross building, which is a bit more serious than just removing some confusing empty directories.

Given `crossPaths` has been default on for a long time, I was just making the default of cross sources as of 0.13.8 a little less surprising.

I'm still on the fence on whether the "issue" lies with sbt or with the tooling around it (IDE support, plugins, etc) as I'm finding it hard to judge. But changing it here hopefully means it fixes it for all. But (...) it does mean more special logic within sbt, which might be best to avoid as much as possible..

I'm happy for you to close this if you still think it isn't needed.

dwijnand:
Ah, I also see how I confused things a little in my description:

> Unfortunately for all non-cross built project (2.11-only for instance) there were some noticeable effects with this change..

I meant to say:

> Unfortunately for all cross-built projects configured for only one version (2.11-only for instance) there were some noticeable effects with this change..

I'm happy to re-write the commit message to make it clearer. And I've edited the original above for the same reason.

eed3si9n:
Also see playframework/playframework#4127. Part of the issue with Play was that it was `mkdirs`ing the source dirs when it didn't exist. OP of the issue says:

> I hope that Play with SBT 0.13.8+ will continue to accept app-2.xx and test-2.xx directories

and there's a fix already to not do `mkdirs` that's been merged - playframework/playframework#4201

dwijnand:
That's good.

But I was also thinking in IntelliJ (perhaps ScalaIDE has something similar) if you enable "Create directories for empty content roots automatically" while importing an sbt project (I never do, but people do) it gives you:

Screenshot from IntelliJ

dwijnand:
But I guess that's only on the first import, and typically only for the start of a project.

So it would just mean deleting unneeded "scala-X" directories just as much as probably most people delete the "java" directories and/or "resources" directories.

eed3si9n:
Plus if someone had already started using `scala-2.10` directory for whatever reason, and if we backed it out in 0.13.9, that's going to be a messier issue.

indrajitr:
Sorry, bit late into the conversation :)
@dwijnand, in case setting `crossPaths` to `false` is not an option (which would have been ideal), will it be possible to consider resetting `unmanagedSourceDirectories` as so?

```
unmanagedSourceDirectories := Seq(scalaSource.value, javaSource.value)
```

This allows us to cover for most cases by default (and override specific ones on case by case basis).

dwijnand:
Yes, that looks perfect. The reason I don't want to disable `crossPaths` is so that the artifact remains scala version specific, so what you propose would be how I'd disable cross sources. Thanks.

PR #1960, opened 2015-04-08
Honor overwrite flag when publishing locally.

asflierl:
It seems since SBT 0.13.8, publishM2 does not overwrite files in the local Maven repo anymore (even though isSnapshot and publishM2Configuration.overwrite are both actually 'true'). I've tracked it down to a 'put' method in ConvertResolver that does not pass along the 'overwrite' flag to the actual file copy method. 
I was unsure how to best proceed with the issue so I just created this pull request for simplicity's sake. The change seems rather uncontroversial but then again: who knows. :)

### DIFF
--- a/ivy/src/main/scala/sbt/ConvertResolver.scala
+++ b/ivy/src/main/scala/sbt/ConvertResolver.scala
@@ -253,7 +253,7 @@ private[sbt] object ConvertResolver {
           if (totalLength > 0) {
             progress.setTotalLength(totalLength);
           }
-          FileUtil.copy(source, new java.io.File(url.toURI), progress)
+          FileUtil.copy(source, new java.io.File(url.toURI), progress, overwrite)
         } catch {
           case ex: IOException =>
             fireTransferError(ex)
--- a/notes/0.13.9/overwrite.markdown
+++ b/notes/0.13.9/overwrite.markdown
@@ -1,0 +1,11 @@
+
+  [@asflierl]: http://github.com/asflierl
+  [1960]: https://github.com/sbt/sbt/pull/1960
+
+### Fixes with compatibility implications
+
+### Improvements
+
+### Bug fixes
+
+- Honor overwrite flag when publishing locally. [#1960][1960] by [@asflierl][@asflierl]


lightbend-cla-validator:
Hi @asflierl,

We see that you have signed the Typesafe Contributors License Agreement before, however, the CLA has changed since you last signed it.

typesafe-tools:

eed3si9n:
LGTM. Could you add a release note in this style? - https://raw.githubusercontent.com/sbt/sbt/0.13/notes/0.13.9/help.markdown

asflierl:
Sure, can do that… but I'm confused. Should I add a note to 0.13.9.markdown or add a fresh file e.g. "overwrite.markdown" with only my note to the 0.13.9 dir? Sorry for the noob-ish question. :)

eed3si9n:
Fresh file like `0.13.9/overwrite.markdown` would be preferable to avoid PR conflicts. We are happy to have more contributors so feel free to ask.

ivy/src/main/scala/sbt/ConvertResolver.scala
```diff
@@ -253,7 +253,7 @@ private[sbt] object ConvertResolver {
           if (totalLength > 0) {
             progress.setTotalLength(totalLength);
           }
-          FileUtil.copy(source, new java.io.File(url.toURI), progress)
+          FileUtil.copy(source, new java.io.File(url.toURI), progress, overwrite)
```
jsuereth, line 256: I think this may not issue a failure if the file already exists...

asflierl:
Done. :)

jsuereth:
The change is good, but Ivy generally throws an exception/error if you try to write to a file that already exists.  You should mimic this if you can.  Right now, I think that utility will just silently not copy.

asflierl:
But isn't this exactly what the overwrite flag is for? To not throw an error if the file exists? 5 lines above my change, there is this:

``` scala
if (!overwrite && resource.exists()) {
  throw new IOException("destination file exists and overwrite == false");
}
```

My use case here is publishing snapshots with publishM2.

jsuereth:
I forget why I made the overwrite commment, something i may have been misunderstanding....  This LGTM

Issue #1958, opened 2015-04-08
Flaky tests on Travis


eed3si9n:
The last time we had a merge was around 11 days ago for 289162a301c64fd8e5a931d7434c91ac125d50c8. For a while I though we reached some level of stability.

Now scripted tests on Travis CI is flaky again. Here are some the tests that are failing. Note some of the Travis CI jobs have been restarted so the failure may not exist currently:

From https://travis-ci.org/sbt/sbt/builds/56452142:
- mavenResolverPluginTest:scripted dependency-management/eviction-warning
- mavenResolverPluginTest:scripted dependency-management/transitive-version-range

From https://travis-ci.org/sbt/sbt/builds/57554505:
- scripted dependency-management/eviction-warning
  - `[error]    {line 1}  Command failed`
- scripted dependency-management/mvn-local
  -  `[error] Looking for ivy file ref, method not implemented!  MavenRepositoryResolver(Maven2 Local) will always return null.`
  -  <pre>[error] /tmp/sbt_d0966a0b/mvn-local/main/B.scala:1: not found: value A
    [error] object B { val y = A.x }
    [error]                    ^
    [error] one error found
    [error](main/compile:compileIncremental) Compilation failed
    [error] Total time: 1 s, completed Apr 8, 2015 12:56:59 AM</pre>
- mavenResolverPluginTest:scripted dependency-management/cached-resolution-classifier
  - `[error] Failed to resolve artifacts from jcenter, The following artifacts could not be resolved: net.sf.json-lib:json-lib:jar:sources:2.4, net.sf.json-lib:json-lib:jar:javadoc:2.4, net.sf.json-lib:json-lib:jar:2.4: Could not find artifact net.sf.json-lib:json-lib:jar:sources:2.4 in jcenter (https://jcenter.bintray.com/)`
- mavenResolverPluginTest:scripted dependency-management/eviction-warning
- mavenResolverPluginTest:scripted dependency-management/mvn-local
  - <pre>[error] /tmp/sbt_8c108aba/mvn-local/main/B.scala:1: object junit is not a member of package org
    [error] import org.junit.Test
    [error]            ^
    [error] one error found
    [debug] Compilation failed (CompilerInterface)
    [error](main/compile:compileIncremental) Compilation failed
    [error] Total time: 0 s, completed Apr 8, 2015 1:01:23 AM</pre>
- scripted tests/fork2
[...]

Issue #1921, opened 2015-03-16
the Aether Maven resolver does not handle transitive dependencies specified with a maven version range expression


dpratt:
In the example below, it looks like Ivy is trying to resolve joda-time:joda-time:[2.2,). The version in the moduleid appears to be a legal Ivy version range expression, but Ivy still chokes on it.

**build.sbt**

```sbt
organization := "com.test"
name := "maven-version-range-bug"
version := "1.0.0-SNAPSHOT"
scalaVersion := "2.11.6"
libraryDependencies += "com.amazonaws" % "aws-java-sdk" % "1.7.8.1"
```

**project/plugins.sbt**
```sbt
addMavenResolverPlugin
```

Running the above project gives the following output - 

```
info] Loading global plugins from /Users/dpratt/.sbt/0.13/plugins
[info] Loading project definition from /Users/dpratt/sbt-resolve-bug/project
[info] Set current project to maven-version-range-bug (in build file:/Users/dpratt/sbt-resolve-bug/)
[info] Updating {file:/Users/dpratt/sbt-resolve-bug/}sbt-resolve-bug...
[info] Resolving jline#jline;2.12.1 ...
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: joda-time#joda-time;[2.2,): Illegal character in path at index 20: joda-time/joda-time/[2.2,)/joda-time-[2.2,).pom
```

PR #1964, opened 2015-04-10
Various workarounds to get Travis CI less flaky

eed3si9n:
#1958.

Our build on Travis CI is not reliable currently.
This PR is an attempt to get it more stable so we can merge the open PRs.

/review @jsuereth


dwijnand:
Such patience. Much respect. (unintended doge meme :-/)

eed3si9n:
Got green on the first try on the `continuous-integration/travis-ci/pr` side:

flaky

 `continuous-integration/travis-ci/push` wasn't so lucky:

```
[error]     tests / do-not-discover
[error]     tests / fork2
```

.travis.yml
```diff
@@ -35,6 +33,8 @@ env:
     - SCRIPTED_TEST="scripted project-load/*"
     - SCRIPTED_TEST="safeUnitTests"
     - SCRIPTED_TEST="checkBuildScala211"
+    # - SCRIPTED_TEST="mavenResolverPluginTest:scripted dependency-management/*1of2 project/transitive-plugins"
```
jsuereth, line 36: why are these disabled?

sbt/src/sbt-test/dependency-management/mvn-local/test
```diff
@@ -43,4 +43,5 @@ $ copy-file changes/libDeps.sbt library/build.sbt
 > library/publishM2
 
 # should succeed now that the dependency is there
+$ sleep 1000
```
jsuereth, line 46: so this is the specific fix?  odd.....

.travis.yml
dwijnand, line 36: https://github.com/sbt/sbt/commit/38b0e57cab7b06d25e311be0e7de5b349035cb47

eed3si9n:
See #1958 for sampling of the flaky tests.

sbt/src/sbt-test/dependency-management/mvn-local/test
eed3si9n, line 46: Scripted error was `{line 46} Command failed`, which is `compile`. If compile is failing because it's not finding the right JAR, I figured incrementing 1s could nudge Ivy to look harder. Not sure if this theory holds.

jsuereth, line 46: it points at a concurrency race that doesn't put me at ease.

.travis.yml
eed3si9n, line 36: as @dwijnand pointed to my commit message, 50% of Travis CI failures are coming from `mavenResolverPluginTest:scripted`, and it seems like many of them are flaky (goes away after repeatedly restarting the build jobs). We certainly should figure out why that is and fix them.
#1921 likely needs refixing, since I'm seeing local failures. Also adding Maven cache files into Travis cache may help.

At the moment, however, it's very labor intensive to validate the PRs because of the high false positive rate. This would result to either us ignoring Travis failures and missing legitimate bugs, or not merging the PRs in timely manner. Neither of which are good tradeoff with keeping flaky tests in our build.

jsuereth:
Ok, let's merge this for now and I'll fix aether resolver in another PR.

PR #1970, opened 2015-04-14
Test exclude

matthewfarwell:
Added exclusions to testOnly and testQuick command line, indicated by a '-' prefix.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -566,11 +566,18 @@ object Defaults extends BuildCommon {
 
   def selectedFilter(args: Seq[String]): Seq[String => Boolean] =
     {
-      val filters = args map GlobFilter.apply
-      if (filters.isEmpty)
+      val (excludeArgs, includeArgs) = args.partition(s => s.startsWith("-"))
+
+      val includeFilters = includeArgs map GlobFilter.apply
+      val excludeFilters = excludeArgs.map(_.substring(1)).map(GlobFilter.apply)
+
+      if (includeFilters.isEmpty && excludeArgs.isEmpty)
         Seq(const(true))
-      else
-        filters.map { f => (s: String) => f accept s }
+      else if (includeFilters.isEmpty) {
+        excludeFilters.map { f => (s: String) => !f.accept(s) }
+      } else {
+        includeFilters.map { f => (s: String) => (f.accept(s) && !excludeFilters.exists(x => x.accept(s))) }
+      }
     }
   def detectTests: Initialize[Task[Seq[TestDefinition]]] = (loadedTestFrameworks, compile, streams) map { (frameworkMap, analysis, s) =>
     Tests.discover(frameworkMap.values.toList, analysis, s.log)._1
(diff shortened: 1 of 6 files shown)


typesafe-tools:

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -566,11 +566,18 @@ object Defaults extends BuildCommon {
...
+
+      val includeFilters = includeArgs map GlobFilter.apply
+      val excludeFilters = excludeArgs.map(_.substring(1)).map(GlobFilter.apply)
+
+      if (includeFilters.isEmpty && excludeArgs.isEmpty)
         Seq(const(true))
-      else
-        filters.map { f => (s: String) => f accept s }
+      else if (includeFilters.isEmpty) {
+        excludeFilters.map { f => (s: String) => !f.accept(s) }
```
jsuereth: I'm not sure this works the way you want it to.  Are all these filters "summed" up at the end, or "productized"?

i.e. you may want a single function that takes the string, and then runs through all the excludes, like:

``` scala
Seq( { (s: String) => excludeFilters forall (f => !(f accept s)) })
```

sbt/src/sbt-test/tests/test-exclude/test
```diff
@@ -0,0 +1,55 @@
...
+# with only exclusion
+> test-only -Test1
+-$ exists target/Test1.run
+$ exists target/Test2.run
+
+$ delete target/Test2.run
+
+
+# with only glob exclusion
+> test-only -Test*
```
jsuereth, line 53: Here, we should have an additional test with some arbitrary string, like -Blah to see if my intuitation about joining filters is correct.

jsuereth:
One minor comment on how we join pure exclusion filters, but otherwise LGTM.

cc> @eed3si9n ?

sbt/src/sbt-test/tests/test-exclude/test
matthewfarwell, line 53: Done.

matthewfarwell:
You are right. I corrected the logic and added more tests (unit & scripted).

DavidPerezIngeniero:
Which version of SBT will have this available?
Maybe 0.13.9?

jsuereth:
Cool, LGTM.  We can include in sbt 0.13.9, I _think_ our tests are fixed, so let's get this through :)

jsuereth:
Oh, this is  based on the flaky maven-test base, so we can merge as-is.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -892,7 +901,7 @@ object Defaults extends BuildCommon {
       selectTests ~ options
     }
 
-  def distinctParser(exs: Set[String], raw: Boolean): Parser[Seq[String]] =
+  private def distinctParser(exs: Set[String], raw: Boolean): Parser[Seq[String]] =
```
dwijnand, line 904: We missed this. This is a breaking change, see https://travis-ci.org/typesafehub/zinc/builds/79476798

```
[error] /home/travis/build/typesafehub/zinc/project/Scriptit.scala:56: method distinctParser in object Defaults cannot be accessed in object sbt.Defaults
[error]     Defaults.distinctParser(scriptitTests(scriptitBase, script).toSet, raw = false)
[error]              ^
```

Issue #1666, opened 2014-10-12
[0.13.7] Natural whitespace parsing breaks XML


eed3si9n:
Found using community-build - https://jenkins-dbuild.typesafe.com:8499/job/Community-2.11.x-manual/56/console
## steps

```
[scala-stm] [error] name := "scala-stm"
[scala-stm] [error] 
[scala-stm] [error] organization := "org.scala-stm"
[scala-stm] [error] 
[scala-stm] [error] version := "0.7"
[scala-stm] [error] 
[scala-stm] [error] scalaVersion := "2.11.0"
[scala-stm] [error] 
[scala-stm] [error] //scalaBinaryVersion := "2.11.0-M8"
[scala-stm] [error] 
[scala-stm] [error] //crossScalaVersions := Seq("2.10.1-RC2", "2.10.0", "2.9.3", "2.9.2")
[scala-stm] [error] 
[scala-stm] [error] //libraryDependencies += ("org.scalatest" % "scalatest_2.10.0-RC5" % "[1.5,)" % "test")
[scala-stm] [error] libraryDependencies += ("org.scalatest" %% "scalatest" % "[1.5,)" % "test")
[scala-stm] [error] 
[scala-stm] [error] libraryDependencies += ("junit" % "junit" % "4.5" % "test")
[scala-stm] [error] 
[scala-stm] [error] // skip exhaustive tests
[scala-stm] [error] testOptions += Tests.Argument("-l", "slow")
[scala-stm] [error] 
[scala-stm] [error] // test of TxnExecutor.transformDefault must be run by itself
[scala-stm] [error] parallelExecution in Test := false
[scala-stm] [error] 
[scala-stm] [error] // make sure Java classes use 1.6 file format
[scala-stm] [error] javacOptions in (Compile, compile) ++= Seq("-source", "1.6", "-target", "1.6")
[scala-stm] [error] 
[scala-stm] [error] ////////////////////
[scala-stm] [error] // publishing
[scala-stm] [error] 
[scala-stm] [error] pomExtra :=
[scala-stm] [error]   <url>http://nbronson.github.com/scala-stm/</url>
[scala-stm] [error]   <licenses>
[scala-stm] [error]     <license>
[scala-stm] [error]       <name>BSD</name>
[scala-stm] [error]       <url>https://github.com/nbronson/scala-stm/blob/master/LICENSE.txt</url>
[scala-stm] [error]       <distribution>repo</distribution>
[scala-stm] [error]     </license>
[...]

PR #1980, opened 2015-04-19
Update scala to version 2.10.5

ajozwik:
See #1666, scala/scala#4186 solved problem with xml groups.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,8 @@ notifications:
     - joshua.suereth@typesafe.com
 
 script:
-  - sbt -J-XX:ReservedCodeCacheSize=128m "$SCRIPTED_TEST"
+  - sbt ";publishLocal;project bundledLauncherProj;publishLocal;project precompiled-2_9_2;publishLocal;project precompiled-2_8_2;publishLocal;project precompiled-2_9_3;publishLocal"     
+  - sbt -J-XX:ReservedCodeCacheSize=128m -Dsbt.version=0.13.9-SNAPSHOT "$SCRIPTED_TEST"
 
   # Tricks to avoid unnecessary cache updates
   - find $HOME/.sbt -name "*.lock" | xargs rm
--- a/build.sbt
+++ b/build.sbt
@@ -21,7 +21,7 @@ def buildLevelSettings: Seq[Setting[_]] = Seq(
 )
 
 def commonSettings: Seq[Setting[_]] = Seq(
-  scalaVersion := "2.10.4",
+  scalaVersion := "2.10.5",
   publishArtifact in packageDoc := false,
   publishMavenStyle := false,
   componentID := None,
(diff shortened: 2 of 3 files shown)


typesafe-tools:

eed3si9n:
Can you run scripted tests locally?

eed3si9n:
The last Scala version bump happened on #1212. This didn't require change to scripted as far as I can tell.

ajozwik:
I can rollback two last commits and you will see the result.
On my local machine first I call `sbt ";publishLocal;project bundledLauncherProj;publishLocal;project precompiled-2_9_2;publishLocal;project precompiled-2_8_2;publishLocal;project precompiled-2_9_3;publishLocal"` and then `sbt -Dsbt.version=0.13.9-SNAPSHOT <scripted name>`

ajozwik:
Failed test - problem with bintray ...

My local result:

```
About to run tests: 
 * compiler-project/*

Running compiler-project / semantic-errors
Running compiler-project / inc-pickled-existential
Running compiler-project / macro-config
Running compiler-project / separate-analysis-per-scala
Running compiler-project / src-dep-plugin
Running compiler-project / error-in-invalidated
Running compiler-project / run2.8
....
```

.travis.yml
```diff
@@ -42,7 +42,8 @@ notifications:
     - joshua.suereth@typesafe.com
 
 script:
-  - sbt -J-XX:ReservedCodeCacheSize=128m "$SCRIPTED_TEST"
+  - sbt ";publishLocal;project bundledLauncherProj;publishLocal;project precompiled-2_9_2;publishLocal;project precompiled-2_8_2;publishLocal;project precompiled-2_9_3;publishLocal"
```
eed3si9n, line 45: If this is required, we should fix the build script instead of putting it in `.travis.yml` so all required JARs are published locally before the scripted tests.

.travis.yml
```diff
@@ -42,7 +42,8 @@ notifications:
     - joshua.suereth@typesafe.com
 
 script:
-  - sbt -J-XX:ReservedCodeCacheSize=128m "$SCRIPTED_TEST"
+  - sbt ";publishLocal;project bundledLauncherProj;publishLocal;project precompiled-2_9_2;publishLocal;project precompiled-2_8_2;publishLocal;project precompiled-2_9_3;publishLocal"     
+  - sbt -J-XX:ReservedCodeCacheSize=128m -Dsbt.version=0.13.9-SNAPSHOT "$SCRIPTED_TEST"
```
eed3si9n, line 46: Why is this needed all a sudden?

ajozwik, line 46: The new process need to use sbt in version 0.13.9-SNAPSHOT. If we do not define the version - it will be red from `project/build.properties`. But there is 0.13.8. And we come back to the "NoSuchMethod" problem.

.travis.yml
ajozwik, line 45: It will be required only - when 0.13.9 will be published. After it the .travis script can be restored. 
Please checkout the code beaf39a196b530e335b44806069b3686dca43f38 and run the tests. You will receive: errors, see all the builds in: https://travis-ci.org/sbt/sbt/builds/54919087:

```
....
About to run tests:

* actions/*

java.lang.NoSuchMethodException: sbt.test.ScriptedTests$.run(java.io.File, boolean, [Ljava.lang.String;, java.io.File, [Ljava.lang.String;, scala.Function1)

at java.lang.Class.getMethod(Class.java:1667)

at Scripted$.doScripted(Scripted.scala:66)

at $0e28670edea040be68f9$$anonfun$scriptedTask$2$$anonfun$apply$2$$anonfun$apply$3.apply(build.sbt:474)

at $0e28670edea040be68f9$$anonfun$scriptedTask$2$$anonfun$apply$2$$anonfun$apply$3.apply(build.sbt:471)

...
```

I guess it is problem with classloader. Mentioned in stack method exist.
As the solution the new process should be run - by scala 2.10.5 - and then all is run by the same classloader.

jsuereth, line 45: Ah, I ran into this.  The way we use scripted is a bit gunky now.  We actually no longer have it isolated, so we need to fix it.

.travis.yml
jsuereth, line 46: @eed3si9n Scripted is borked in some way.  you can't change its code in the current repoa nd see those changes.  I suggest, maybe, we just remove it from the codebase and modularize it.

jsuereth:
@eed3si9n We should figure out how to merge this before 0.13.9

jsuereth:
Ok, so we bumped to 2.11.6, what should we do with this PR?

ajozwik:
You can jest close without merging. 2.11.6 contains the same correction

eed3si9n:
@jsuereth we bumped what to what?

eed3si9n:
Closing this in favor of #2068

PR #1981, opened 2015-04-20
publish to bintray

eed3si9n:
/review @jsuereth

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,7 +11,12 @@ import Sxr.sxr
 // but can be shared across the multi projects.
 def buildLevelSettings: Seq[Setting[_]] = Seq(
   organization in ThisBuild := "org.scala-sbt",
-  version in ThisBuild := "0.13.9-SNAPSHOT"
+  version in ThisBuild := "0.13.9-SNAPSHOT",
+  // bintrayOrganization in ThisBuild := None,
+  // bintrayRepository in ThisBuild := "test-test-test",
+  bintrayOrganization in ThisBuild := Some("sbt"),
+  bintrayRepository in ThisBuild := "ivy-releases",
+  bintrayPackage in ThisBuild := "sbt"
 )
 
 def commonSettings: Seq[Setting[_]] = Seq(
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -11,7 +11,12 @@ import Sxr.sxr
 // but can be shared across the multi projects.
 def buildLevelSettings: Seq[Setting[_]] = Seq(
   organization in ThisBuild := "org.scala-sbt",
-  version in ThisBuild := "0.13.9-SNAPSHOT"
+  version in ThisBuild := "0.13.9-SNAPSHOT",
+  // bintrayOrganization in ThisBuild := None,
+  // bintrayRepository in ThisBuild := "test-test-test",
+  bintrayOrganization in ThisBuild := Some("sbt"),
```
jsuereth, line 17: this is fine, but FYI, we'll have to mirror this on the typesafe org so the URLs match the old one.

jsuereth:
LGTM

build.sbt
```diff
@@ -26,7 +31,9 @@ def commonSettings: Seq[Setting[_]] = Seq(
   testOptions += Tests.Argument(TestFrameworks.ScalaCheck, "-w", "1"),
   javacOptions in compile ++= Seq("-target", "6", "-source", "6", "-Xlint", "-Xlint:-serial"),
   incOptions := incOptions.value.withNameHashing(true),
-  crossScalaVersions := Seq(scala210)
+  crossScalaVersions := Seq(scala210),
+  bintrayPackage := (bintrayPackage in ThisBuild).value,
+  bintrayRepository := (bintrayRepository in ThisBuild).value
```
dwijnand, line 36: Why is this necessary?

eed3si9n, line 36: So I can change the value after load in one spot, and all subproject would listen to it.

dwijnand, line 36: I'm still not getting it :-/

Is this overriding some other setting?
Or is it a stub for this changing in the near future?

It just looks like a noop to me..

eed3si9n, line 36: It is rewiring what bintray-sbt adds to each project.
It's not a noop since `commonSettings` gets appended to all subprojects as a project-level setting whereas `(bintrayRepository in ThisBuild).value` is a build-level setting.

dwijnand, line 36: Ah-ha, ok got it! Thank you, Eugene.

PR #1989, opened 2015-04-25
Add inThisBuild, similar to inConfig. Fixes #1847

### LINKED ISSUE #1847: inThisBuild function
I'm using `in ThisBuild` more lately when I port `Build.scala` to `build.sbt`.
It would be nice to have `inThisBuild` similar to `inConfig`.

dwijnand:
Fixes #1847

Also added the

```
private[sbt] def inThisBuild[T](i: Initialize[T]): Initialize[T]
```

variant, to match the others, but I'm happy to drop that if it's superfluous.


typesafe-tools:

eed3si9n:
Awesome! Thanks for the contribution.

PR #1990, opened 2015-04-26
Add help message for `inspect actual`. Fixes #1651.

### LINKED ISSUE #1651: help inspect doesn't mention inspect actual
Just found out about `inspect actual` as described in http://www.scala-sbt.org/release/docs/Inspecting-Settings.html#Actual+Dependencies:

> `inspect actual <scoped-key>` shows the actual dependency used.

Executing `help inspect` doesn't show it in the help, tough:

```
> help inspect
inspect <key>

    For a plain setting, the value bound to the key argument is displayed using its toString method.
    Otherwise, the type of task ("Task" or "Input task") is displayed.

    "Dependencies" shows the settings that this setting depends on.

    "Reverse dependencies" shows the settings that depend on this setting.

    When a key is resolved to a value, it may not actually be defined in the requested scope.
    In this case, there is a defined search sequence.
    "Delegates" shows the scopes that are searched for the key.
    "Provided by" shows the scope that contained the value returned for the key.

    "Related" shows all of the scopes in which the key is defined.

inspect tree <key>

    Displays `key` and its dependencies in a tree structure.
    For settings, the value bound to the setting is displayed and for tasks, the type of the task is shown.

inspect uses <key>

    Displays the settings and tasks that directly depend on `key`.

inspect definitions <key>

    Displays the scopes in which `key` is defined.

> about
[info] This is sbt 0.13.7-M3
[info] The current project is {file:/Users/jacek/sandbox/resourceGenerators/}resourcegenerators 1.0
[info] The current project is built against Scala 2.11.2
[info] Available Plugins: sbt.plugins.IvyPlugin, sbt.plugins.JvmPlugin, sbt.plugins.CorePlugin, sbt.plugins.JUnitXmlReportPlugin, com.typesafe.sbteclipse.plugin.EclipsePlugin, net.virtualvoid.sbt.graph.Plugin, com.github.tototoshi.sbt.musical.Plugin, com.timushev.sbt.updates.UpdatesPlugin
[info] sbt, sbt plugins, and build definitions are using Scala 2.10.4
```

dwijnand:
Fixes #1651

### DIFF
--- a/main/src/main/scala/sbt/CommandStrings.scala
+++ b/main/src/main/scala/sbt/CommandStrings.scala
@@ -95,39 +95,36 @@ $ShowCommand <task>
 
   val InspectCommand = "inspect"
   val inspectBrief = (InspectCommand + " [uses|tree|definitions] <key>", "Prints the value for 'key', the defining scope, delegates, related definitions, and dependencies.")
-  val inspectDetailed =
-    InspectCommand + """ <key>
-
-	For a plain setting, the value bound to the key argument is displayed using its toString method.
-	Otherwise, the type of task ("Task" or "Input task") is displayed.
-
-	"Dependencies" shows the settings that this setting depends on.
-
-	"Reverse dependencies" shows the settings that depend on this setting.
-
-	When a key is resolved to a value, it may not actually be defined in the requested scope.
-	In this case, there is a defined search sequence.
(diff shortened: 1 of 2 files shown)


typesafe-tools:

eed3si9n:
LGTM

PR #1991, opened 2015-04-26
Add quit command, an alias of exit. Fixes #1374.

### LINKED ISSUE #1374: quit vs. exit - unify the way how to leave any session/tool
Hi, wouldn't be better to make quitting console and sbt with the same command?
## steps

Now I need
If I want to leave console session:

```
:quit
```

If I want to leave sbt:

```
exit
```
### problem

?
### expectation

I suppose 'exit' and ':exit' would save many errors ...

dwijnand:
Fixes #1374.
Implemented so that it matches the Scala REPL's :quit.

### DIFF
--- a/main/command/src/main/scala/sbt/BasicCommands.scala
+++ b/main/command/src/main/scala/sbt/BasicCommands.scala
@@ -16,7 +16,7 @@ import java.io.File
 import scala.util.control.NonFatal
 
 object BasicCommands {
-  lazy val allBasicCommands = Seq(nop, ignore, help, completionsCommand, multi, ifLast, append, setOnFailure, clearOnFailure, stashOnFailure, popOnFailure, reboot, call, early, exit, continuous, history, shell, read, alias) ++ compatCommands
+  lazy val allBasicCommands = Seq(nop, ignore, help, completionsCommand, multi, ifLast, append, setOnFailure, clearOnFailure, stashOnFailure, popOnFailure, reboot, call, early, exit, quit, continuous, history, shell, read, alias) ++ compatCommands
 
   def nop = Command.custom(s => success(() => s))
   def ignore = Command.command(FailureWall)(idFun)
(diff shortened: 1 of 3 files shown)


typesafe-tools:

eed3si9n:
Wouldn't it make sense for the unified command to be vim quit? (e.g. `:quit` and `:q`)

dwijnand:
Given the typos/inversions in the original ticket I assumed `quit` was intended.

I'm happy with whatever, I use ctl-D personally.
Would you want to drop `quit` from this PR (while adding `:quit` and `:q`)?

eed3si9n:
We can do all three:
- `quit`
- `:quit`
- `:q`

jsuereth:
Can we also have `kthxbai` as an option?  I've always wanted that to be supported too.

/the peanut gallery will shut up now.

dwijnand:
Of course `kthxbai` , how could I forget?

dwijnand:
I can obviously whip up 2 more commands (and probably will if no one comments back before) but I'm wondering, should these perhaps be encoded as aliases? i.e. using [BasicCommands.newAlias](https://github.com/sbt/sbt/blob/0.13.8/main/command/src/main/scala/sbt/BasicCommands.scala#L261-L262)?

eed3si9n:
Alias would probably be better.

dwijnand:
Hmm thank god I checked:

```
java.lang.AssertionError: assertion failed: ':quit' is not a valid command name.
    at scala.Predef$.assert(Predef.scala:179)
    at sbt.SimpleCommand.<init>(Command.scala:17)
    at sbt.Command$.make(Command.scala:33)
    at sbt.Command$.make(Command.scala:32)
    at sbt.BasicCommands$.newAlias(BasicCommands.scala:275)
```

:-/ Thoughts?

eed3si9n:
I don't think there's a point in having `quit` if it doesn't work on Scala REPL.

One way to "solve" this is to define a plugin that does something along the lines of:

```
initialCommands += "\nimport sys.exit"
```

For Scala versions that defines `sys.exit`. That will allow `exit` to work on both REPL and sbt shell.

dwijnand:
I'm not sure, perhaps having `quit` available would also make sbt a little more beginner-friendly (I'm thinking of Ctrl-C in vim :D). If so I'd just need to tweak the note, and I'll still make the command an alias.

WRT the original issue, I'll update that ticket with what we've discovered here and your suggestion, as it would obviously mean work elsewhere. Maybe I'll still do it, as it could be my gateway drug into writing and maintaining an actual sbt plugin.

dwijnand:
Let me know if a `quit` command is still of interest.

PR #1992, opened 2015-04-26
Fix crossScalaVersions default, derived from scalaVersion. Fixes #1828.

### LINKED ISSUE #1828: Better crossScalaVersions default behavior
**build.sbt**

``` scala
scalaVersion := "2.11.4"
```

**project/build.properties**

```
sbt.version=0.12.4
```

Then

``` bash
$ sbt 'show scalaVersion'
[info] 2.11.4
$ sbt 'show crossScalaVersions'
[info] List(2.11.4)
```

---

**build.sbt**

``` scala
scalaVersion := "2.11.4"
```

**project/build.properties**

```
sbt.version=0.13.7
```

Then

``` bash
$ sbt 'show scalaVersion'
[info] 2.11.4
$ sbt 'show crossScalaVersions'
[info] List(2.10.4)
```

---

It seems that the default behavior for `crossScalaVersions` was better in 0.12. Is there a deliberate reason this changed?

dwijnand:
Fixes #1828
Using derive, which I don't know well, so let me know if this is wrong.


typesafe-tools:

eed3si9n:
The use of `derive` makes sense here, but this is no simple bug fix as it will change the way `crossScalaVersions` is defined by default. The release note should move to "Fixes with compatibility implications" and explain how it's been broken during 0.13.(0-8).

dwijnand:
Very true. I'll try and find time later today to amend that. Thanks Eugene.

dwijnand:
I spent a long time trying to word it right that now I'm second guessing myself. Have a look, what do you think?

notes/0.13.9/cross-scala-versions-default.markdown
```diff
@@ -0,0 +1,22 @@
+
+  [@dwijnand]: http://github.com/dwijnand
+  [1828]: https://github.com/sbt/sbt/issues/1828
+  [1992]: https://github.com/sbt/sbt/pull/1992
+
+### Fixes with compatibility implications
+
+- New `crossScalaVersions` default value, now correctly derives from what `scalaVersion` is set to. See below for more info.
```
eed3si9n: "New" and "now" are both relative. How about:

> Starting 0.13.9, `crossScalaVersions` default value is fixed back to the older 0.12.x behavior. See below for details.

notes/0.13.9/cross-scala-versions-default.markdown
```diff
@@ -0,0 +1,22 @@
...
+
+### Improvements
+
+### Bug fixes
+
+### New `crossScalaVersions` default value
```
eed3si9n: Remove "New"

notes/0.13.9/cross-scala-versions-default.markdown
```diff
@@ -0,0 +1,22 @@
...
+
+As of this fix `crossScalaVersions` returns to the behaviour present in `0.12.4` whereby it defaults to what `scalaVersion`
+is set to, for example if `scalaVersion` is set to `"2.11.6"`, `crossScalaVersions` now defaults to `Seq("2.11.6")`.
+
+Therefore when upgrading from any version between `0.13.0` and `0.13.8` please be aware of this new default if your build
```
eed3si9n: Remove "please".

dwijnand:
Done, thanks for the feedback!

eed3si9n:
Since we have a green checkbox on the code change, I'm going to merge. Don't try this at home.

PR #1994, opened 2015-04-27
Add Ivy `branch` support to ModuleID

oxbowlakes:
Allow an Ivy branch to be set on the project's module (such that Ivy is aware of it) , as well as on its dependencies.
This change:
- adds the `branchName: Option[String]` to `ModuleID` while preserving backwards-compatibility.
- changes the resolution cache's artifact pattern to include the `([branch]/)` after the module name
- changes `defaultIvyPatterns` as well as the local ivy repository pattern to include the `([branch]/)` after the module name
- changes `ModuleReport` generation to set the branch onto the `ModuleID` (on top of saving it inside the `ModuleReport`)
- includes `branchName` in `Format[ModuleID]` and adds a `CacheIvyTest` that ensures a cached-then-read `ModuleID` remains equal to the original (save for `CrossVersion`, which is not being cached perfectly)

Usage: 

``` scala
libraryDependencies += "org.something" % "module" % "version" branch "myBranch"
```

or

``` scala
libraryDependencies += "org.something" % "module" % "version" branch "myBranch"
```

### DIFF
--- a/ivy/src/main/scala/sbt/Ivy.scala
+++ b/ivy/src/main/scala/sbt/Ivy.scala
@@ -496,7 +496,8 @@ private[sbt] object IvySbt {
     }
   private[this] def defaultInfo(module: ModuleID): scala.xml.Elem = {
     import module._
-    <info organisation={ organization } module={ name } revision={ revision }/>
+    val base = <info organisation={ organization } module={ name } revision={ revision }/>
+    branch.fold(base) { br => base % new scala.xml.UnprefixedAttribute("branch", br, scala.xml.Null) }
   }
   private[this] def addExtraAttributes(elem: scala.xml.Elem, extra: Map[String, String]): scala.xml.Elem =
     (elem /: extra) { case (e, (key, value)) => e % new scala.xml.UnprefixedAttribute(key, value, scala.xml.Null) }
--- a/ivy/src/main/scala/sbt/Ivy.scala
+++ b/ivy/src/main/scala/sbt/Ivy.scala
@@ -427,7 +427,7 @@ private[sbt] object IvySbt {
   def toID(m: ModuleID) =
     {
       import m._
-      ModuleRevisionId.newInstance(organization, name, revision, javaMap(extraAttributes))
+      ModuleRevisionId.newInstance(organization, name, branch.orNull, revision, javaMap(extraAttributes))
     }
 
   private def substituteCross(m: ModuleSettings): ModuleSettings =
--- a/ivy/src/main/scala/sbt/Ivy.scala
+++ b/ivy/src/main/scala/sbt/Ivy.scala
@@ -427,7 +427,7 @@ private[sbt] object IvySbt {
   def toID(m: ModuleID) =
     {
       import m._
-      ModuleRevisionId.newInstance(organization, name, branch.orNull, revision, javaMap(extraAttributes))
+      ModuleRevisionId.newInstance(organization, name, branchName.orNull, revision, javaMap(extraAttributes))
     }
 
   private def substituteCross(m: ModuleSettings): ModuleSettings =
@@ -497,7 +497,7 @@ private[sbt] object IvySbt {
   private[this] def defaultInfo(module: ModuleID): scala.xml.Elem = {
     import module._
     val base = <info organisation={ organization } module={ name } revision={ revision }/>
-    branch.fold(base) { br => base % new scala.xml.UnprefixedAttribute("branch", br, scala.xml.Null) }
+    branchName.fold(base) { br => base % new scala.xml.UnprefixedAttribute("branch", br, scala.xml.Null) }
   }
   private[this] def addExtraAttributes(elem: scala.xml.Elem, extra: Map[String, String]): scala.xml.Elem =
     (elem /: extra) { case (e, (key, value)) => e % new scala.xml.UnprefixedAttribute(key, value, scala.xml.Null) }
--- a/ivy/src/main/scala/sbt/IvyActions.scala
+++ b/ivy/src/main/scala/sbt/IvyActions.scala
@@ -313,7 +313,7 @@ object IvyActions {
     {
       val mextra = IvySbt.javaMap(mid.extraAttributes, true)
       val aextra = IvySbt.extra(art, true)
-      IvyPatternHelper.substitute(pattern, mid.organization, mid.name, mid.branch.orNull, mid.revision, art.name, art.`type`, art.extension, conf, null, mextra, aextra)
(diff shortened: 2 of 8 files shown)


lightbend-cla-validator:
Hi @dansanduleac,

typesafe-tools:

oxbowlakes:
My name is Chris Marshall and I work for GSA Capital, where the original author of these commits, Dan Sanduleac, also works.

I can confirm that GSA have downloaded and signed the CLA but wish to clarify that this applies _only_ to PRs submitted to SBT via my github account (I have gone through step 2: "confirmation of signing via on github"), not directly from Dan's. Our reason for this is that the agreement contains no clause protecting GSA in the event that Dan uploads IP that we did not wish him to. Hence as an internal control, we asked that Dan's contributions go via a _trusted_ employee (that would be me).

If you have any questions about this, please contact me @oxbow_lakes

jsuereth:
@oxbowlakes The CLA looks fine. The automation goes by commit author.

jsuereth:
So, @oxbowlakes and @dansanduleac -  Dan has done GREAT work here.  

My main concern here is around our ability to drop Ivy in the future.  This feature would be very hard to replicate in Aether (which already has to contort severely to support ivy-metadata).

Are you guys using this feature at work?  If so can you give us the use case?  Once we get the unit tests passing, I think we could possibly merge this work for now, I'm just hesitant that it will hinder our ability to migrate to a better resolution logic engine in the future (which we know we need to do.  Even cached resolution is not enough for usability).

S11001001:
@jsuereth Surrounding [ermine-language](https://github.com/ermine-language/) development at McGraw Hill Financial, we have sbt projects with multiple lines of active development, e.g. "v1" and "v2", and upstream and downstream non-sbt projects with similar lines of development.

So that those working downstream — within and without our set of sbt projects — don't have to constantly rebuild the sbt projects as they change, which requires quite a different setup from what they're commonly using, our CI deploys to a shared Ivy repo with the Ivy branch marked accordingly, "v1" or "v2".  The downstream Ivy files reference projects by those Ivy branches, `latest.integration` so changes upstream don't necessitate updating the version identifiers downstream.  Our sbt projects do the same to interdepend and depend on upstream non-sbt projects in the build.

When we introduce another VCS branch, we update the Ivy branch identifiers, and start CI building for that branch as well.  Thus, development proceeds on the parallel lines, the binary set of each branch never stepping on those of the others.

Lack of sbt support for Ivy branches has long obliged us to use `ivy.xml` files and `externalIvyFile` in place of `libraryDependencies` in our sbt projects.  We would love to drop the XML.

oxbowlakes:
@jsuereth - I'll leave some of the technical details to Dan but, yes, we are using a forked SBT with this PR as a central plank of our build system. We might have multiple branches in git for a number of reasons (none of these seem particularly unusual):
- we develop new functionality on "feature" or "bugfix" branches
- we have multiple "live" branches being built at any one time, corresponding to the latest patch of a `major.minor` version pair, as we have many library consumers not necessarily running the latest code
- we might have bespoke branches for particular consumers

We have well over a hundred individual libraries/applications (each in their own git repositories), with many interdependencies; adding new features may involve cross-cutting concerns across many such libraries. Hence we need to be able to declare that any one given library is dependent on a specific branch of another.

Whilst having our fork is workable - we wish to use vanilla SBT as much as possible, primarily because we might expect IDE integration to work well (currently this is one of my major gripes about SBT) and hence want the PR to become part of the codebase.

Obviously we can see your concerns about the constraints this might place on the future direction of SBT.

OlegYch:
please excuse my ignorance, but how is this different from embedding branch name in artifact name or version?

jsuereth:
@OlegYch The big difference is in conflict resolution.  Ivy doesn't know the modules are the same if you embedd branches inside the artifact name.  This is an issue with how we encode scala in maven, as it leads to possible BROKEN resolutions with two incompatible JARs on the same classpath, whereas in Ivy we correctly deconstruct the module name for correct conflict resolution.

@oxbowlakes  @S11001001 Ok, I hear you guys.   I'd agree we should add this feature then, but I may need some of @dansanduleac's help if we try to encode this same semantic inside of another dependency management system later!  In any case, this concern was why we held off on the branch feature for so long.

Could I ask for a few things?
1. A `notes/<feature>.md` file describing the change
2. Help updating our documentation (particularly around @S11001001's use case) about this feature?

dansanduleac:
@jsuereth Sorry about the delay, I'll prepare a document describing the change as you requested, and have a look at the documentation surrounding this. 

I might need your help though with fixing the pickling issue which is causing the tests to fail - I couldn't quite figure out how to create picklers that would work for both `ModuleID` and `ModuleIDWithBranch`.
I've tried adding `@directSubclasses(Array(classOf[ModuleIDWithBranch]))` to `ModuleID`, however now I'm getting this error 

```
 /data/home/dsandul/ext/sbt/ivy/src/main/scala/sbt/ModuleID.scala:127: Cannot generate a pickler for sbt.ModuleIDWithBranch. Recompile with -Xlog-implicits for details
   implicit val pickler: Pickler[ModuleID] with Unpickler[ModuleID] = PicklerUnpickler.generate[ModuleID]
                                                                                               ^
```

and compiling with `-Xlog-implicits` doesn't seem to give me any meaningful information.
I might be missing something obvious, but don't have much experience with the pickling library, so not sure what it is that it's getting tripped on.

jsuereth:
@dansanduleac I think we'll have to write a custom pickler.
1. Write a custom pickler at the ModuleID level
2. Always return MOduleID w/ Branch from this pickler.

This should work as we want...

kdeggelman:
Hi all - is there currently a workaround for specifying a branch name within sbt? I'd like to avoid downloading the JAR if possible. Thanks!

dwijnand:
This needs a rebase.

dansanduleac:
This also needs working Pickler / Unpickler implementations for the
proposed ModuleID hierarchy, as @jsuereth pointed out (only a temporary
issue, while we need to maintain backward compatibility with 0.13,
otherwise we could just change ModuleID directly). Unfortunately, I haven't
had the time to do that for months now. Maybe I will get a chance to
revisit it soon in the interest of finally merging this.

All I can say for now is, even without that, sbt 0.13.8 still compiles and
works with this fix alone (note: ignoring the fact that some tests don't
compile, because they expect the Pickler to exist). But the main code
compiles.. I suspect this is because sbinary is still being used for
caching, and a transition to the picking stuff will take place soon?

jsuereth:
@dansanduleac Exactly.  We're also migrating this code into a module.   You should follow up with @eed3si9n on the details.

eed3si9n:
I'm closing this since this is now part of sbt 1.0.x code base. https://github.com/sbt/librarymanagement/pull/26

PR #1995, opened 2015-04-27
Add POM props missing from ScmInfo/Developer

pdalpra:


typesafe-tools:

ivy/src/main/scala/sbt/IvyInterface.scala
```diff
@@ -21,9 +22,9 @@ final case class ModuleInfo(nameFormal: String, description: String = "", homepa
 }
 
 /** Basic SCM information for a project module */
-final case class ScmInfo(browseUrl: URL, connection: String, devConnection: Option[String] = None)
+final case class ScmInfo(browseUrl: URL, connection: String, devConnection: Option[String] = None, tag: Option[String] = None)
```
jsuereth, line 25: Unfortunately this isn't a binary compatible change.   

You'll need to rip the apply/unapply methods into a companion object and change it to a regular class.  Then have two constructors, one with the existing options and one with the extra.  It's really gunky.

An alternative MAY be to have the tag encoded in the URL as a "#" or something.  Not the best., but it would work.

ivy/src/main/scala/sbt/IvyInterface.scala
```diff
@@ -21,9 +22,9 @@ final case class ModuleInfo(nameFormal: String, description: String = "", homepa
...
 
-final case class Developer(id: String, name: String, email: String, url: URL)
+final case class Developer(id: String, name: String, email: String, url: URL, organization: Option[String] = None, organizationUrl: Option[URL] = None, roles: Seq[String] = Nil, timezone: Option[TimeZone] = None, properties: Map[String, String] = Map.empty)
```
jsuereth, line 27: Wow, I didn't realize this existed.   Again, the same issue with binary compatibility.

Let me chat with @eed3si9n to see if we come up with a clever way to add these without breaking BC.

jsuereth:
Looks great, except we need to find a binary compatible way to do this.

ivy/src/main/scala/sbt/IvyInterface.scala
pdalpra, line 25: If you're okay with it and there isn't another solution, breaking it up in regular class + two apply/unapply methods would be the best course IMO (cleaner, no 'clever encoding'...)

pdalpra:
@jsuereth Any ideas or should I go with your previous suggestion ?

jsuereth:
Let's try the previous suggestion.

pdalpra:
Sure :)
Just a suggestion : maybe we can keep those classes as case classes and only override `apply`/`unapply` methods on the companion or would it break binary comp ?
This would allow to keep the default impl of `equals`/`hashCode`/`toString`.

jsuereth:
@pdalpra That sounds like an interesting idea, can you frame it so I understand it better?

pdalpra:
Right away :)
This is basically what I have in mind :

``` scala
object Developer {
 // There is already an apply method that takes all properties, new ones included

 // Let's add an apply method that takes only the properties that existed before this PR
 def apply(id: String, name: String, email: String, url: URL) = …

 // Same goes for unapply, there is already one extracting all the properties, 
 // let's add an another one that extracts the previously existing properties
 def unapply(dev: Developer): Option[(String, String, String, URL)] = …
}
case class Developer(current properties, new properties without default values)
```

Does this seem an acceptable solution which does not break binary compatibilty ?

pdalpra:
@jsuereth any news ? :)

eed3si9n:
My vote is to kick this can forward to 1.0. At least it's not ready for 0.13.9.

pdalpra:
Sure, if 1.0 allows to break binary compatibility, this PR is ready as-is.

jsuereth:
yeah, let's defer this to 1.0.  @eed3si9n We should probably figure out how we start cutting 1.0.  We need to first extract the lib-mgt module or do we need to merge this into a 1.0 branch then extract a lib-mgt module?

eed3si9n:
@jsuereth Probably creative git-filter-branch like we did with some other projects (sbt-remote-control?) to extract parts of code base out when the time comes.

eed3si9n:
Please resend this PR to sbt/librarymanagement.

PR #1996, opened 2015-04-28
Fix issue with aether resolver and bad URLs.

jsuereth:
It appears like the change to use the new MRID when checking for JAR was not applied before merge.

Review by @eed3si9n.  This should fix the unit tests.


eed3si9n:
#1958 catalogs some of the flaky tests that was failing on Travis intermittently. Is URL thing one of them?

jsuereth:
@eed3si9n Yes.  It actually looks like a merge conflict deleted a good portion of the code I had, somehow.   We're missing commits/fixes.  I'm just going through them one at a time.

PR #2006, opened 2015-05-07
Added Extracted.runInputTask

jroper:
This provides a convenience function for running an input task from the extracted state. This is particularly useful for commands, release steps etc that may want to run input tasks, like scripted.

I haven't added tests, because I'm not sure where the right place to test this is - I don't think any of the specs are applicable here unless there's any easy way to fire up a fake sbt project, so maybe a scripted test? But should it be an existing scripted test? If there is a right place, let me know, and I'll add a test there.

### DIFF
--- a/main/src/main/scala/sbt/Extracted.scala
+++ b/main/src/main/scala/sbt/Extracted.scala
@@ -3,6 +3,7 @@ package sbt
 import Project._
 import Scope.GlobalScope
 import Def.{ ScopedKey, Setting }
+import sbt.complete.Parser
 import std.Transform.DummyTaskMap
 
 final case class Extracted(structure: BuildStructure, session: SessionSettings, currentRef: ProjectRef)(implicit val showKey: Show[ScopedKey[_]]) {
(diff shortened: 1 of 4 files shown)


typesafe-tools:

jsuereth:
Nice!  yes we should get a test.  The right place is probably be in the "api" section of scripted tests.  There is nothing specific to `Project.extract` at this point.

jsuereth:
Also, the code itself LGTM

jsuereth:
@jroper woudl you mind rebasing this against master? Should clean up the PR validation failures.

jroper:
Rebased, tests written.

jsuereth:
I give a grade I+

dwijnand:
Either now that this has just kicked off a Travis build, or if you want when it's done, notes plz! :D

eed3si9n:
LGTM pending release note (e.g. https://github.com/sbt/sbt/blob/8aca4e8fa6d4da9f31b267f2a8e503c147be237b/notes/0.13.9/force-gc.markdown)

jroper:
Release note added.

dwijnand:
LGTM. Nice tests.

PR #2012, opened 2015-05-12
Fix/1856

### LINKED ISSUE #1856: CustomPomParser produces unparsable ivy descriptor when POM has a parent
#1191 is closed, so I'm opening this because it's still an issue.

If a child POM has a parent section, the transformed child will have an e:sbtTransformHash tag without the corresponding namespace declaration.

Here's why:
- The parent POM is processed and addExtra is used.  This adds e:sbtTransformHash to its extraInfo and adds the extra namespace
- Ivy's PomModuleDescriptorParser adds the parent's extra info to the child before finishing parseDescriptor
- CustomPomParser begins transforming the child module, but it already has the sbtTransformHash tag, so it skips addExtra

Because addExtra is never run on the child module, the child is left with the extraInfo sbtTransformHash tag but never gets the addExtraNamespace.

jsuereth:
Fixes #1856 

REview by @eed3si9n

### DIFF
--- a/ivy/src/main/scala/sbt/CustomPomParser.scala
+++ b/ivy/src/main/scala/sbt/CustomPomParser.scala
@@ -74,14 +74,10 @@ object CustomPomParser {
       val oldTransformedHashKey = "sbtTransformHash"
       val extraInfo = md.getExtraInfo
       val MyHash = MakeTransformHash(md)
-      val h = MyHash
       // sbt 0.13.1 used "sbtTransformHash" instead of "e:sbtTransformHash" until #1192 so read both
       Option(extraInfo).isDefined &&
         ((Option(extraInfo get TransformedHashKey) orElse Option(extraInfo get oldTransformedHashKey)) match {
-          case x @ Some(MyHash) =>
-            true
-          case Some(other) =>
-            false
+          case Some(MyHash) => true
           case _ => false
         })
     }
--- a/ivy/src/main/scala/sbt/CustomPomParser.scala
+++ b/ivy/src/main/scala/sbt/CustomPomParser.scala
@@ -51,7 +51,15 @@ object CustomPomParser {
   private[this] val TransformedHashKey = "e:sbtTransformHash"
   // A hash of the parameters transformation is based on.
   // If a descriptor has a different hash, we need to retransform it.
-  private[this] val TransformHash: String = hash((unqualifiedKeys ++ JarPackagings).toSeq.sorted)
+  private[this] def makeCoords(mrid: ModuleRevisionId): String = s"${mrid.getOrganisation}:${mrid.getName}:${mrid.getRevision}"
+
+  // We now include the ModuleID in a hash, to ensure that parent-pom transformations don't corrupt child poms.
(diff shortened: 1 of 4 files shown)


eed3si9n:
This hopefully fixes news but do you need to do anything about the existing/cached poms that broke?

jsuereth:
@eed3si9n Comments accounted for, and rebased.

ivy/src/main/scala/sbt/CustomPomParser.scala
```diff
@@ -95,10 +108,10 @@ object CustomPomParser {
       val mergeDuplicates = IvySbt.hasDuplicateDependencies(md.getDependencies)
 
       val unqualify = toUnqualify(filtered)
-      if (unqualify.isEmpty && extraDepAttributes.isEmpty && !convertArtifacts && !mergeDuplicates)
```
dwijnand, line 98: Should we kill `mergeDuplicates`, `convertArtifacts` and the `artifactExtIncorrect` method as a part of this PR?

ivy/src/main/scala/sbt/CustomPomParser.scala
```diff
@@ -65,11 +73,16 @@ object CustomPomParser {
...
+      val h = MyHash
       // sbt 0.13.1 used "sbtTransformHash" instead of "e:sbtTransformHash" until #1192 so read both
       Option(extraInfo).isDefined &&
         ((Option(extraInfo get TransformedHashKey) orElse Option(extraInfo get oldTransformedHashKey)) match {
-          case Some(TransformHash) => true
-          case _                   => false
+          case x @ Some(MyHash) =>
+            true
+          case Some(other) =>
+            false
```
dwijnand: redundant?

ivy/src/main/scala/sbt/CustomPomParser.scala
```diff
@@ -65,11 +73,16 @@ object CustomPomParser {
     {
       val oldTransformedHashKey = "sbtTransformHash"
       val extraInfo = md.getExtraInfo
+      val MyHash = MakeTransformHash(md)
+      val h = MyHash
```
dwijnand: redundant?

dwijnand:
I remember trying to figure out how to fix this. Nice work.

ivy/src/main/scala/sbt/CustomPomParser.scala
jsuereth: ah, wow.  Yeah, leaving my debug mess hanging around.

ivy/src/main/scala/sbt/CustomPomParser.scala
jsuereth, line 98: not in this PR.  mergeDuplicates has some work it still does which is useful, we just need to figure otu either an alternative mechanism _or_ make that mechanism work.

eed3si9n:
LGTM. notes plz.

jsuereth:
Notes there.  You can wait for the next build, or merge directly.

eed3si9n:
Merged!

PR #2017, opened 2015-05-13
Add forceUpdatePeriod

ajsquared:
Adds `forceUpdatePeriod` key, that takes values of `Option[FiniteDuration]`.  If set, a full `update` will occur after that amount of time without needing to explicitly run the `update` task.

The use case is to simplify the workflow a fit for projects with SNAPSHOT/`changing()` dependencies.  By setting `forceUpdatePeriod` you can ensure you will get new SNAPSHOT versions periodically even without explicitly running `update`.


typesafe-tools:

ajsquared:
The `dependency-management/mvn-local` test seems to be flaky for me locally.  Its failing intermittently both with and without the changes in this PR.

`source-dependencies/macro-arg-dep-2-11` and `source-dependencies/macro-arg-dep-nested-2-11` are both working consistently for me locally with these changes.

dwijnand:
Why not use `scala.concurrent.duration.Duration`/`FiniteDuration` instead of a Long?

ajsquared:
@dwijnand Good call, I'll make that change

ajsquared:
`source-dependencies/macro-arg-dep-2-11` and `source-dependencies/macro-arg-dep-nested-2-11` are still passing for me locally.

The `reporter/*` failure looks spurious as well.

jsuereth:
LGTM

Issue #1562, opened 2014-09-01
Infinite loop when typing !!!!


mauhiz:
So I had this console where I usually type `!!` (previous command)... then typed `!!` again (that makes 4 of them).. and SBT started an infinite loop of displaying `!!!!`

Crazy !!!! See #1562 for recursion issues.

Also, there was no breaking out of it short of killing the console.

(sbt 0.13.5 on Windows 7 console)

dwijnand:
This isn't 0.13.5 or Windows specific, and by that I mean I can reproduce it with 0.13.8 on Mac (in sbt's build FWIW).

kzys:
Fixed due to #2025? I couldn't reproduce the issue on sbt 0.13.9 @ OS X.

dwijnand:
Looks like it did, thanks @kzys.

PR #2025, opened 2015-05-18
Prevent history command(s) from going into an infinite loop

PanAeon:
Preventing execution of history commands as result of another history command, thus fixing infinite loops. Ticket  [#1562][1562]


typesafe-tools:

jsuereth:
This LGTM

PR #2027, opened 2015-05-20
Do not emit in the pom dependencies that are only sources or docs

### LINKED ISSUE #2001: withSources() flag incorrectly exports source code
The sbt `withSources()` flag incorrectly exposes source code as a classpath dependency of the artifact.
## Steps
1. git clone https://github.com/htmldoug/sbt-withSources-bug1.git
2. sbt publishLocal
3. open ~/.ivy2/local/default/sbt-withsources-bug1_2.10/0.1-SNAPSHOT/poms/sbt-withsources-bug1_2.10.pom
4. observe that sources are exported as classpath dependencies

``` scala
<dependency>
    <groupId>com.typesafe</groupId>
    <artifactId>config</artifactId>
    <version>1.2.1</version>
    <classifier>sources</classifier>
</dependency>
```
## Problems

Using withSources()...

```
libraryDependencies += "com.typesafe" % "config" % "1.2.1" withSources()
```

... causes the sources to be **exported in the pom**.

``` xml
<dependency>
    <groupId>com.typesafe</groupId>
    <artifactId>config</artifactId>
    <version>1.2.1</version>
    <classifier>sources</classifier>
</dependency>
```
1. This behavior is undocumented, surprising, and a security risk. In our play projects, sbt-native-packager happily gobbles up these classpath dependencies and **deploys our source code to our prod servers!** 
2. In downstream dependencies, I've seen sbt evict the classes dependency, e.g. `"com.typesafe" % "config" % "1.2.1"`, in favor of the sources jar! Example not included here (since sources shouldn't be exported in the first place), but I can create one if needed.
## Expectations

From [Download Sources](http://www.scala-sbt.org/0.13/docs/Library-Management.html#Download+Sources):

> To have sbt download the dependency’s sources without using an IDE plugin, add withSources() to the dependency definition.

There is no mention of exposing the source code as a classpath dependency. I expect that this is not the correct behavior.
## Notes

Example here: https://github.com/htmldoug/sbt-withSources-bug1

cunei:
The pom generation code tries its best to map Ivy's configurations
to Maven scopes; however, sources and javadoc artifacts cannot be
properly mapped and they currently are emitted as dependencies in
the default scope (compile). That may lead to the source/doc jars
being erroneously processed like regular jars by automated tools. 

Arguably, the source/docs jars should not be included in the pom
file as dependencies at all. This commit filters out the
dependencies that only appear in the sources and/or javadoc Ivy
configurations, thereby preventing them from appearing in the
final pom file.

This pull request should fix #2001. Review by @eed3si9n and @jsuereth.

### DIFF
--- a/ivy/src/main/scala/sbt/MakePom.scala
+++ b/ivy/src/main/scala/sbt/MakePom.scala
@@ -233,28 +233,33 @@ class MakePom(val log: Logger) {
       val artifacts = dependency.getAllDependencyArtifacts
       val includeArtifacts = artifacts.filter(d => includeTypes(d.getType))
       if (artifacts.isEmpty) {
-        val (scope, optional) = getScopeAndOptional(dependency.getModuleConfigurations)
-        makeDependencyElem(dependency, scope, optional, None, None)
+        val configs = dependency.getModuleConfigurations
+        if (configs.filterNot(Set("sources","docs")).nonEmpty) {
+          val (scope, optional) = getScopeAndOptional(dependency.getModuleConfigurations)
+          makeDependencyElem(dependency, scope, optional, None, None)
+        } else NodeSeq.Empty
       } else if (includeArtifacts.isEmpty)
         NodeSeq.Empty
       else
-        NodeSeq.fromSeq(artifacts.map(a => makeDependencyElem(dependency, a)))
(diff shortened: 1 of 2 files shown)


jsuereth:
LGTM

jsuereth:
Can you add notes?

PR #2030, opened 2015-05-21
Fixes #1721/#1763. Cached resolution: summarize callers in graph.json

### LINKED ISSUE #1721: Cached resolution: OutOfMemoryException
On a rather complex project with many subproject and dependencies, running `update` with cached resolution fails with OutOfMemoryException even with 3GB heap (without cached resolution it definitely doesn't fail on 2GB and probably will work with 1GB).

Memory dump analysis shows that a 700MB byte array was created and held under `sbinary.JavaOutput`.

It would be nice if this array was at least split into pieces.

SBT version: 0.13.7-RC3

eed3si9n:
- On some of the builds graph.json is reaching 250MB+
- JSON parsing alone takes hours
- 97% of the content are caller info
- This change summarizes all callers into one (zero caller would have correctness issues)
  #1721/#1763

/review @jsuereth

### DIFF
--- a/ivy/src/main/scala/sbt/JsonUtil.scala
+++ b/ivy/src/main/scala/sbt/JsonUtil.scala
@@ -25,8 +25,34 @@ private[sbt] object JsonUtil {
     }
   def toLite(ur: UpdateReport): UpdateReportLite =
     UpdateReportLite(ur.configurations map { cr =>
-      ConfigurationReportLite(cr.configuration, cr.details)
+      ConfigurationReportLite(cr.configuration, cr.details map { oar =>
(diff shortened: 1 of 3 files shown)


jsuereth:
LGTM, think this need notes

eed3si9n:
Travis failures are legit, I think. I need to address the correctness issue on SNAPSHOT due to my removing the callers from the graph. Hopefully I can approximate it with a single `Boolean` that indicates that the path is dynamic.

eed3si9n:
Zero caller broke when there's `force()` etc. I updated to summarize all callers with just one. It seems to work at least for the scripted tests that failed.

jsuereth:
LGTM

PR #2035, opened 2015-05-24
Add project-level exclusions in the POM.

### LINKED ISSUE #1877: Project-level exclusion in pom
as pointed out by @sslavic in #1748, project-level exclusion needs to be translated to dependency-level in pom.

dwijnand:
Closes #1877.

---

This is probably not done, it's just the first signs of understanding on my part of how this part of the
codebase works.

So I just wanted to push up what I was going with, in hope for some early feedback.

### DIFF
--- a/ivy/src/main/scala/sbt/MakePom.scala
+++ b/ivy/src/main/scala/sbt/MakePom.scala
@@ -104,7 +104,7 @@ class MakePom(val log: Logger) {
        {
          val deps = depsInConfs(module, configurations)
          makeProperties(module, deps) ++
-           makeDependencies(deps, includeTypes)
+           makeDependencies(deps, includeTypes, module.getAllExcludeRules)
        }
        { makeRepositories(ivy.getSettings, allRepositories, filterRepositories) }
      </project>)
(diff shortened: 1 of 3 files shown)


eed3si9n:
You might want to create a scripted test to repro failure first with an isolated ivy cache and maven repo.

typesafe-tools:

dwijnand:
Will do. For now my (manual) test case is:

```
libraryDependencies += "com.typesafe.akka"          %% "akka-actor"   % "2.3.11"
libraryDependencies += "com.github.julien-truffaut" %% "monocle-core" % "1.1.1"

excludeDependencies += "com.typesafe" % "config"
```

produces:

```
<dependencies>
    <dependency>
        <groupId>org.scala-lang</groupId>
        <artifactId>scala-library</artifactId>
        <version>2.11.6</version>
        <exclusions>
            <exclusion>
                <groupId>com.typesafe</groupId>
                <artifactId>config</artifactId>
            </exclusion>
        </exclusions>
    </dependency>
    <dependency>
        <groupId>com.typesafe.akka</groupId>
        <artifactId>akka-actor_2.11</artifactId>
        <version>2.3.11</version>
        <exclusions>
            <exclusion>
                <groupId>com.typesafe</groupId>
                <artifactId>config</artifactId>
            </exclusion>
        </exclusions>
    </dependency>
    <dependency>
        <groupId>com.github.julien-truffaut</groupId>
        <artifactId>monocle-core_2.11</artifactId>
        <version>1.1.1</version>
        <exclusions>
            <exclusion>
                <groupId>com.typesafe</groupId>
                <artifactId>config</artifactId>
            </exclusion>
        </exclusions>
    </dependency>
</dependencies>
```

eed3si9n:
I'm guessing you can grab some ideas from https://github.com/sbt/sbt/tree/0.13/sbt/src/sbt-test/dependency-management/mvn-local

jsuereth:
Nice work!

dwijnand:
So I've added a test to the existing `exclude-dependencies` scripted test, but only for project `b` in the test as we don't have group-level exclusions.

Will that do? Or do we want to try and do group-wide exclusions in the pom?

[This question on Stackoverflow](http://stackoverflow.com/questions/13018625/how-to-exclude-all-artifacts-from-a-group-in-maven/13022383#13022383) seems to suggest that this is now an option in Maven.

[This](https://github.com/sbt/sbt/blob/0.13.8/ivy/src/main/scala/sbt/MakePom.scala#L320-L321) is the portion of code that we'd have to change.

jsuereth:
@dwijnand I say le'ts leave it as-is for now and open another PR about the "*" option for group-wide exclusions.  maven may support it in latest, but we'd want to make sure there's a good wanrign about it.

Also, your PR will break Binary Compatibility, but in an area that we hope people arne't relyign on BC.   Should be easy-ish to fix (I hope).

jsuereth:
awh, man.  This needs to be rebased, but LGTM

dwijnand:
Done.

PR #1938, opened 2015-03-25
Removing launcher in favor of re-wrapping sbt/launcher module

jsuereth:
Removes the launcher code from this project.
- launcher-interface is depended on directly form maven-central
- launcher-implementation is rebundled against the current version of sbt
- All tests/dependencies updated appropriately.
- sonatype-snapshots added as repository when we're using snapshot modules of launcher.
- Bintray is now looked in _first_ for all artifacts, as it's generally faster.  We still look at
  maven central, because sbt-plugins on there do NOT sync to bintray correctly.

Review by @eed3si9n

PR #2057, opened 2015-06-14
Adds bundledLauncherProj to allProj

eed3si9n:
This matters when someone tries to locally build sbt from source.

/review @dwijnand, @jsuereth


dwijnand:
:+1:

Everything's used by `sbtRoot` now:
- before: before
- after: after

eed3si9n:
@jsuereth There's a legit failure introduced by #1938 (wrapped launcher).
`launch` directory contains tests, but because `autoScalaLibrary` is set to `false`, it's unable to compile the tests.

jsuereth:
Hmm, so we should remove the tests?

Your PR looks good to me

eed3si9n:
Yea. I'll try to split out the test into another subproject.

PR #2070, opened 2015-06-22
Add gitter channels to README.

dwijnand:
Not sure if README's the best place, or if we want to use gitter's badge or how to word it, but I thought I'd start that conversation here.


typesafe-tools:

eed3si9n:
I suggest we centralize all communication channel notifications to http://www.scala-sbt.org/community.html

dwijnand:
WFM, superseeded by sbt/website#160

PR #2072, opened 2015-06-24
Reinstate SNAPSHOT test that was fixed in #1520/#1514

eed3si9n:
REF #1520/#1514
Bringing back a scripted test /review @jsuereth, @dwijnand


dwijnand:
LGTM mod Travis.

PR #2079, opened 2015-06-27
Fixes #1973. Renames distinct to distinctName

### LINKED ISSUE #1973: PathFinder.distinct is based on file name, not path
If I have the following files:
- `a/b/foo.txt`
- `a/c/foo.txt`

And I do:

``` scala
file("a").***.distinct
```

I end up with only one file, even though there are two distinct paths, because the implementation of `distinct` matches on file name, not file path.

See https://github.com/playframework/playframework/issues/4208 for an example of issues this can cause.

eed3si9n:
Fixes #1973

/review @dwijnand, @jroper

### DIFF
--- a/notes/0.13.9/distinct-rename.markdown
+++ b/notes/0.13.9/distinct-rename.markdown
@@ -1,0 +1,11 @@
+
+  [@eed3si9n]: http://github.com/eed3si9n
+  [1973]: https://github.com/sbt/sbt/issues/1973
+
+### Fixes with compatibility implications
+
+### Improvements
+
+- Renames `distinct` method on `PathFinder` to `distinctName`. [#1973][1973] by [@eed3si9n][@eed3si9n]
+
+### Bug fixes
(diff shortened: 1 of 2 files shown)


dwijnand:
The change LGTM, but it doesn't really "fix" #1973.

Perhaps we should add a `.distinctPath`, which would solve #1973 (imo).

Also I agree with @jroper about the intuitive intent of `.distinct` and would argue that the deprecated message should point to `.distinctPath` (or maybe just indicate both).

dwijnand:
For reference, how @jroper worked around current `.distinct`: https://github.com/playframework/playframework/pull/4236

eed3si9n:
What's the return type of `distinctPath`? `PathFinder`, `Seq[String`, or `Seq[File]`?

dwijnand:
`PathFinder` I would've thought.

eed3si9n:
Added it.

``` scala
scala> import sbt._
import sbt._

scala> (file("src") ***).distinctPath
res0: sbt.PathFinder = 

   src/main
   src/main/conscript/sbt
   src/main/conscript/screpl/launchconfig
   src/main/conscript/screpl
   src
   src/main/conscript/scalas/launchconfig
   src/main/conscript/sbt/launchconfig
   src/main/conscript
   src/main/conscript/scalas
```

dwijnand:
Sweet. LGTM.

Issue #2008, opened 2015-05-07
References to undefined settings at runtime.


DavidPerezIngeniero:
Sometimes SBT complains with this exception:

```
sbt.Init$RuntimeUndefined: References to undefined settings at runtime.
        at sbt.Init$class.sbt$Init$$handleUndefined(Settings.scala:506)
        at sbt.Init$$anon$4.apply(Settings.scala:511)
        at sbt.Init$$anon$4.apply(Settings.scala:511)
        at sbt.$tilde$greater$$anon$6.apply(TypeFunctions.scala:39)
        at sbt.$tilde$greater$$anon$6.apply(TypeFunctions.scala:39)
        at sbt.std.FullInstance$$anonfun$flatten$1$$anonfun$apply$3.apply(TaskMacro.scala:51)
        at sbt.std.FullInstance$$anonfun$flatten$1$$anonfun$apply$3.apply(TaskMacro.scala:51)
        at scala.Function1$$anonfun$compose$1.apply(Function1.scala:47)
        at sbt.$tilde$greater$$anonfun$$u2219$1.apply(TypeFunctions.scala:40)
        at sbt.std.Transform$$anon$4.work(System.scala:63)
        at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:226)
        at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:226)
        at sbt.ErrorHandling$.wideConvert(ErrorHandling.scala:17)
        at sbt.Execute.work(Execute.scala:235)
        at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:226)
        at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:226)
        at sbt.ConcurrentRestrictions$$anon$4$$anonfun$1.apply(ConcurrentRestrictions.scala:159)
        at sbt.CompletionService$$anon$2.call(CompletionService.scala:28)
        at java.util.concurrent.FutureTask.run(FutureTask.java:262)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
        at java.util.concurrent.FutureTask.run(FutureTask.java:262)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
        at java.lang.Thread.run(Thread.java:745)
```

[...]

PR #2080, opened 2015-06-28
Add notes for #2008/#2009.

dwijnand:


typesafe-tools:

Issue #1967, opened 2015-04-11
SBT don't recompile related java file use final field


bash99:
how to reproduce: 
#### A.java

``` java
Class A {
   public static final String SCRIPT_1 = "return 1";
}
```
#### B.java

``` java
Class B {
    public static void main(String[] args) {
        System.out.println(A.SCRIPT_1);
    }
}
```
1. change `A.java`'s constant to `"return 2"`; save and recompile.

only A.java is compiled, and run B still output
`"return 1"`;

English is not my native language, so sorry for any mistakes.

stuhood:
This appears to be due to http://stackoverflow.com/questions/3524150/is-it-possible-to-disable-javacs-inlining-of-static-final-variables

eed3si9n:
So the workaround would be to change to 

``` java
Class A {
   public static final String SCRIPT_1 = "return 1".toString();
}
```

stuhood:
@eed3si9n : Yea. I'm going to see whether it is feasible to include the primitive content of the static-final-primitive field in the API of the class, so that this is considered to be an API change.

stuhood:
It looks like it should be possible to do this via the `sbt.classfile.ClassFile` API, which is already used for java compilation.

One question for current SBT committers: would you be opposed to adding usage of that classfile parser to `sbt.ClassToAPI`? It looks like it currently uses only information that is available via reflection, but in this case we'd need to look in (at least) the constant pool.

eed3si9n:
I don't think I'm opposed to the idea of it. Just not sure what the implementation entails, and feasibility of it.

stuhood:
@eed3si9n @jsuereth : Any thoughts on how to represent a final-static field in SourceAPI?

I've gotten the constant value from `sbt.classfile.ClassFile`, and now the change seems to be a matter of exposing the constant as part of the API of the class. The most obvious backwards-compatible approach seems to be to represent it as an `xsbti.api.Singleton`, but it's not immediately obvious what a good unique `xsbti.api.Path` would be for the type. Perhaps a name-mangling of the field?

eed3si9n:
I could be off base, but there's an effort going on around incremental compiler to generalize the notion of dependencies, like #1757, so I'm wondering if we could track relationship between the Java sources without mangling. /cc @Duhemm

stuhood:
@eed3si9n : No, I don't think so: we want field/api-level dependency tracking, and #1757 provides file level tracking.

jsuereth:
[...]

PR #2085, opened 2015-07-02
Encode static-final constant fields as Singletons

stuhood:
... so that when their values change, the APIs change and downstream consumers are invalidated. See #1967 

There are 3 TODOs in the code currently:
1. We re-parse the classfile for every final field we encounter; should instead plumb through a lazy val for the parsed representation
2. Could skip parsing for non-static fields, since javac won't inline them
3. Should add a negative test, split out a `JavacForUnitTests` class, and move the new test to the appropriate package


lightbend-cla-validator:
Hi @stuhood,

typesafe-tools:

stuhood:
@jsuereth, @eed3si9n : I believe this one is ready for review... would you mind taking a look?

jsuereth:
@stuhood on it now, sorry for delay.

compile/api/src/main/scala/sbt/ClassToAPI.scala
```diff
@@ -67,21 +68,26 @@ object ClassToAPI {
...
-      val parentJavaTypes = allSuperTypes(c)
-      if (!Modifier.isPrivate(c.getModifiers))
-        cmap.inherited ++= parentJavaTypes.collect { case c: Class[_] => c }
-      val parentTypes = types(parentJavaTypes)
-      val instanceStructure = new api.Structure(lzyS(parentTypes.toArray), lzyS(all.declared.toArray), lzyS(all.inherited.toArray))
-      val staticStructure = new api.Structure(lzyEmptyTpeArray, lzyS(all.staticDeclared.toArray), lzyS(all.staticInherited.toArray))
-      (staticStructure, instanceStructure)
+  def structure(c: Class[_], enclPkg: Option[String], cmap: ClassMap): (api.Structure, api.Structure) = {
+    lazy val cf = {
+      // TODO: need better way to get access to classfile-location
```
jsuereth: I think maybe the TODO shoudl be about reading as much of the APi as possible from a classfile parser rather than reflection.  Reflection locks us into a specific JDK we have to be running on, where as parsing frees that.

compile/api/src/main/scala/sbt/ClassToAPI.scala
```diff
@@ -127,14 +133,54 @@ object ClassToAPI {
   def upperBounds(ts: Array[Type]): api.Type =
     new api.Structure(lzy(types(ts)), lzyEmptyDefArray, lzyEmptyDefArray)
 
-  def fieldToDef(enclPkg: Option[String])(f: Field): api.FieldLike =
```
jsuereth, line 130: This would be a binary compatibility breakage.  May be ok, but would prefer to see this deprecated and an overload/diferent name instead.

compile/api/src/main/scala/sbt/ClassToAPI.scala
```diff
@@ -127,14 +133,54 @@ object ClassToAPI {
...
+  /** Parses the constant value represented by the given ConstantValue AttributeInfo. */
+  private def constantPoolConstantValue(cf: ClassFile, ai: classfile.AttributeInfo): AnyRef = {
+    assert(ai.name.exists(_ == "ConstantValue"), s"Non-ConstantValue attribute not supported: ${ai}")
+    import classfile.Constants._
+    cf.constantPool(classfile.Parser.entryIndex(ai)) match {
+      case classfile.Constant(ConstantString, nextOffset, _, _) =>
+        // follow the indirection from ConstantString to ConstantUTF8
+        val nextConstant = cf.constantPool(nextOffset)
+        nextConstant.value.getOrElse {
+          throw new RuntimeException(s"Empty UTF8 value in constant pool: $nextConstant")
```
jsuereth: Ideally I'd like some other exception which extends runtime exception here.  I'm also clueless what we should actually do if we encounter this.  Is it safe to ignore?  Does it mean a corrupted classfile?  

Ideally I'd like a string telling people what to do if we understand why it would happen.

compile/api/src/main/scala/sbt/ClassToAPI.scala
```diff
@@ -127,14 +133,54 @@ object ClassToAPI {
...
+        }
+      case constant @ classfile.Constant((ConstantFloat | ConstantLong | ConstantDouble | ConstantInteger), _, _, ref) =>
+        ref.getOrElse {
+          throw new RuntimeException(s"Empty primitive value in constant pool: $constant")
```
jsuereth: Again, ideally the error message should offer guidance for how to fix.  (the fallback in my mind is just assume corrupted class file and ask to clean).  Something like:

`s"Empty primitive value in constant pool: $constant \n  This may mean your classfiles are corrupted.  Please clean and try again"`

compile/api/src/main/scala/sbt/ClassToAPI.scala
```diff
@@ -127,14 +133,54 @@ object ClassToAPI {
...
+              c.getName.split("\\.").toSeq :+ (name + "$" + constantValue)
+          }.map { constantComponents =>
+            new api.Singleton(pathFromStrings(constantComponents))
+          }
+        } else {
+          None
+        }
+      val tpe = specificTpe.getOrElse(reference(returnType(f)))
+      if (mods.isFinal) {
+        new api.Val(tpe, name, accs, mods, annots)
```
jsuereth, line 168: Cool, so if we see a "val" with a singleton type, we can assume that this is one which Java may be performing the constant optimisation/elide on, right?

compile/src/test/scala/sbt/compiler/javac/JavaCompilerSpec.scala
```diff
@@ -14,20 +16,24 @@ object JavaCompilerSpec extends Specification {
...
+    issue errors and warnings                                       ${findsErrors(forked)}
+    yield the same errors as local javac                            $forkSameAsLocal
 
   Documenting a file with forked javadoc should
-      document a java file                  ${docWorks(forked)}
-      find errors in a java file            ${findsDocErrors(forked)}
+    document a java file                                            ${docWorks(forked)}
+    find errors in a java file                                      ${findsDocErrors(forked)}
+
+  Analyzing classes generated by javac should result in
```
jsuereth, line 32: Nice additons!!!

jsuereth:
This LGTM.  My comments are all very minor, but usability related.   Can't wait to see more of these :)

adriaanm:
Skimming on BART,  I'm wondering whether the compiler's ConstantType is available here? When type checking, a final val without an expected type is typed as a ConstantType that tracks its value, denoting it can be constant folded. In general, it's not safe to do this for a SingletonType. I'll have a deeper look at the latest when M2 is out.

jsuereth:
@adriaanm We may need to add a new subclass to denote ConstantType.   you're right that may be better as an option in the future.   Shouldn't be a difficult change to make.   Let use know after M2 :)

jsuereth:
Cleanups look great!, just waiting on @adriaanm if he has any concerns over how we've encoded the types of the constant pool

compile/api/src/main/scala/sbt/ClassToAPI.scala
```diff
@@ -127,14 +135,48 @@ object ClassToAPI {
...
+      val specificTpe: Option[api.Type] =
+        if (mods.isFinal) {
+          val cf = _cf
+          val attributeInfos = cf.fields.find(_.name.exists(_ == name)).toSeq.flatMap(_.attributes)
+          // create a singleton type ending with the ConstantValue of this field. because this type
+          // is purely synthetic, it's fine that the name might contain filename-banned characters.
+          attributeInfos.collectFirst {
+            case ai @ classfile.AttributeInfo(Some("ConstantValue"), _) =>
+              try {
+                c.getName.split("\\.").toSeq :+ (name + "$" + cf.constantValue(ai))
```
adriaanm: I think it would be useful to elaborate a bit on why these parts (class name, field name, constant value) are included in the singleton type's path. The thought process of a future incremental compiler hacker would have to go something like this. If the class name changes, but the constant does not, should we recompile? (I assume we don't.) So, does it suffice to include the value? But wait! If you change the constant's value, and not the class name, the change won't be detected, so I should include the reference to the field to encode the dependency. 

How about the type of the field (not captured here), though? Does it matter whether I go from `int` to `long`, as, erm, long as the value doesn't change?

adriaanm: Regarding my earlier comment, I think it's okay not to add ConstantType to the API, but I would suggest including a marker (:snail:, to evoke its constant nature?) in the singleton type's path to make sure we can tell the difference if we ever need to.

adriaanm:
This is great stuff, thanks! I added a suggestion or two.

compile/api/src/main/scala/sbt/ClassToAPI.scala
stuhood: That's a lot of questions, none of which I know the answer to. "baby's first sbt/scala patch", and so forth.

What would you suggest?

adriaanm: Sorry, I should've used a more specific format for that feedback. 

I think your encoding makes sense, but it does not capture the type of the field, which is unlikely not to also result in a change of value when it changes, but it could (and thus changes to the type of the field would go undetected). 

I'd also like to see some docs for the motivation of choosing these elements for the path. I think my rhetorical questions could serve as a basis for that.

compile/api/src/main/scala/sbt/ClassToAPI.scala
adriaanm, line 168: So, to also capture the field's type, you could use an intersection of `specificTpe` and `reference(returnType(f))`.

adriaanm, line 168: I guess that's a `Structure` in sbt API speak (it's been a while since I worked with Greg on this part of the compiler.)

compile/api/src/main/scala/sbt/ClassToAPI.scala
stuhood: > I think your encoding makes sense, but it does not capture the type of the field, which is unlikely not to also result in a change of value when it changes, but it could (and thus changes to the type of the field would go undetected).

Aha! Absolutely right. Will fix.

jsuereth:
Thanks @adriaanm, all good points.

dwijnand:
This needs notes as well btw.

compile/api/src/main/scala/sbt/ClassToAPI.scala
stuhood: I've added a test for the type change, and expanded the explanation around the choice of members in the Singleton type. Thanks!

stuhood:
@dwijnand , @adriaanm : I believe I've addressed your feedback. Thanks!

jsuereth:
LGTM

adriaanm:
Thank you! I love the template approach to testing compilation. Hypnotoad approves.

PR #2091, opened 2015-07-09
Don't show the stack trace on compilation error in build definition

Duhemm:
This pull request hides the stack trace on compilation error in the build definition, as suggested in https://github.com/sbt/sbt/issues/2071.

### DIFF
--- a/notes/0.13.10/hide-stacktrace.markdown
+++ b/notes/0.13.10/hide-stacktrace.markdown
@@ -1,0 +1,12 @@
+
+  [@Duhemm]: http://github.com/Duhemm
+  [2071]: https://github.com/sbt/sbt/issues/2071
+  [2091]: https://github.com/sbt/sbt/pull/2091
+
+### Fixes with compatibility implications
+
+### Improvements
+
+- Hide the stack trace on compilation error in build definition [#2071][2071]/[#2091][2091] by [@Duhemm][@Duhemm]
+
+### Bug fixes
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -478,7 +478,17 @@ object BuiltinCommands {
       val (s1, base) = Project.loadAction(SessionVar.clear(s0), action)
       IO.createDirectory(base)
       val s = if (s1 has Keys.stateCompilerCache) s1 else registerCompilerCache(s1)
-      val (eval, structure) = Load.defaultLoad(s, base, s.log, Project.inPluginProject(s), Project.extraBuilds(s))
+
+      val (eval, structure) =
+        try Load.defaultLoad(s, base, s.log, Project.inPluginProject(s), Project.extraBuilds(s))
+        catch {
+          case ex: compiler.EvalException =>
+            s0.log.debug(ex.getMessage)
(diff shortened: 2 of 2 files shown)


typesafe-tools:

eed3si9n:
How does output for `scalaVersion := "2.11.6".xx` look now?

Would it be difficult to add test for this? Could you also add notes in the style of https://raw.githubusercontent.com/sbt/sbt/0.13/notes/0.13.9/distinct-rename.markdown

dwijnand:
Catching, logging and rethrowing looks like it would log more!

eed3si9n:
Could you paste the session of what the user would see for `scalaVersion := "2.11.6".xx`?

Duhemm:
```
master… foobar ❯ xsbt                                                                                                                                                           9:15
[info] Loading global plugins from /Users/martin/.sbt/0.13/plugins
[info] Updating {file:/Users/martin/.sbt/0.13/plugins/}global-plugins...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
[info] Loading project definition from /Users/martin/Documents/Projects/Duhemm/foobar/project
[info] Updating {file:/Users/martin/Documents/Projects/Duhemm/foobar/project/}foobar-build...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
/Users/martin/Documents/Projects/Duhemm/foobar/Build.sbt:2: error: value xx is not a member of String
    scalaVersion := "2.11.3".xx,
                             ^
sbt.compiler.EvalException: Type error in expression
[error] sbt.compiler.EvalException: Type error in expression
```

If the user types `l`, then the debug log (that contains the complete stack trace) will be shown.

@dwijnand Actually the exception is re-thrown without its stack trace

dwijnand:
Ah, that's so uncommon I completely missed it, nice.

eed3si9n:
I think this LGTM. @jsuereth?

dwijnand:
This is merging into 0.13, but has release notes for 0.13.9. Do we want this to merge into 0.13.9, or alternatively should this be fore 0.13.10?

dwijnand:
And perhaps we should reference #2071 in the notes.

Duhemm:
@dwijnand I've moved the notes and added the reference to #2071, thanks!

notes/0.13.10/hide-stacktrace.markdown
```diff
@@ -0,0 +1,11 @@
+
+  [@Duhemm]: http://github.com/Duhemm
+  [2091]: https://github.com/sbt/sbt/pull/2091
+
+### Fixes with compatibility implications
+
+### Improvements
+
+- Hide the stack trace on compilation error in build definition [#2071][2071]/[#2091][2091] by [@Duhemm][@Duhemm]
```
dwijnand, line 10: Sorry 1 more thing, you need a reference to 2071 above.

Duhemm, line 10: That's fixed – Sorry for all these problems!

dwijnand:
LGTM. Thanks.

jsuereth:
Thsi LGTM too.

PR #2094, opened 2015-07-09
Make Logger.null public + minor clean up

pdalpra:
This PR makes the dummy `Logger.Null` public.
While I was at it, I also added explicit types annotation on methods where they were missing.

### DIFF
--- a/interface/src/main/java/xsbti/Logger.java
+++ b/interface/src/main/java/xsbti/Logger.java
@@ -5,9 +5,9 @@ package xsbti;
 
 public interface Logger
 {
-	public void error(F0<String> msg);
-	public void warn(F0<String> msg);
-	public void info(F0<String> msg);
-	public void debug(F0<String> msg);
-	public void trace(F0<Throwable> exception);
+	void error(F0<String> msg);
+	void warn(F0<String> msg);
+	void info(F0<String> msg);
+	void debug(F0<String> msg);
+	void trace(F0<Throwable> exception);
 }
--- a/util/log/src/main/scala/sbt/Logger.scala
+++ b/util/log/src/main/scala/sbt/Logger.scala
@@ -41,8 +41,7 @@ object Logger {
     newLog.setTrace(oldLog.getTrace)
   }
 
-  // make public in 0.13
-  private[sbt] val Null: AbstractLogger = new AbstractLogger {
+  val Null: AbstractLogger = new AbstractLogger {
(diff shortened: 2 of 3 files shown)


typesafe-tools:

pdalpra:
Shall I add notes too ?

eed3si9n:
Yes for notes. The change looks fine pending Travis.

pdalpra:
Notes added, in 0.13.10.

interface/src/main/java/xsbti/Logger.java
```diff
@@ -5,9 +5,9 @@
 
 public interface Logger
 {
-	public void error(F0<String> msg);
-	public void warn(F0<String> msg);
-	public void info(F0<String> msg);
-	public void debug(F0<String> msg);
-	public void trace(F0<Throwable> exception);
+	void error(F0<String> msg);
```
jsuereth, line 8: why no public?

util/log/src/main/scala/sbt/Logger.scala
```diff
@@ -10,19 +10,19 @@ import java.io.File
 
 abstract class AbstractLogger extends Logger {
   def getLevel: Level.Value
-  def setLevel(newLevel: Level.Value)
-  def setTrace(flag: Int)
+  def setLevel(newLevel: Level.Value): Unit
+  def setTrace(flag: Int): Unit
```
jsuereth, line 14: nice! I've always wanted to take the time to do this :)

jsuereth:
LGTM

interface/src/main/java/xsbti/Logger.java
pdalpra, line 8: In a Java interface, methods are implicitly `public`, constants are implicitly `public static final`, so public is not needed :)

eed3si9n, line 8: Found this on StackOverflow, JLS says http://docs.oracle.com/javase/specs/jls/se7/html/jls-9.html#jls-9.4

> It is permitted, but discouraged as a matter of style, to redundantly specify the public and/or abstract modifier for a method declared in an interface.

pdalpra:
Shall I amend my PR to make the `Null` Logger to an object, like you suggested in the mailing list ?

jsuereth:
@pdalpra No. Looks like it wouldn't be binary compatible then.

PR #2095, opened 2015-07-09
Do not use `ListBuffer#readOnly`

adriaanm:
It's dangerous, deprecated, and was removed in 2.12.0-M1.
See https://github.com/scala/scala/pull/4140.

`ListBuffer#toList` has equivalent performance,
except it actually returns an immutable copy(-on-write).

This turned up while trying to get the community build to [pass for 2.12.0-M2](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/33/console)


typesafe-tools:

eed3si9n:
LGTM pending Travis

Issue #1763, opened 2014-12-04
Cached resolution: OOM when compiling


ornicar:
With -Xmx1G. OOM does not happen with 0.13.7 without CachedResolution

Reproduce:

```
git clone -b cachedresolution https://github.com/ornicar/lila.git
cd lila
git submodule update --init modules/chess
./bin/build-deps.sh
sbt compile
```

```
[info] Loading global plugins from /home/thib/dotfiles/sbt/plugins
[info] Updating {file:/home/thib/.sbt/0.13/plugins/}global-plugins...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
[info] Loading project definition from /home/thib/tmp/lila/project
[info] Updating {file:/home/thib/tmp/lila/project/}lila-build...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
[info] Compiling 3 Scala sources to /home/thib/tmp/lila/project/target/scala-2.10/sbt-0.13/classes...
[warn] there were 1 deprecation warning(s); re-run with -deprecation for details
[warn] one warning found
[info] Set current project to lila (in build file:/home/thib/tmp/lila/)
[info] Updating {file:/home/thib/tmp/lila/}common...
[info] Updating {file:/home/thib/tmp/lila/}chess...
[info] Updating {file:/home/thib/tmp/lila/}swisssystem...
[info] Done updating.
[info] Done updating.
[info] Done updating.
[info] Updating {file:/home/thib/tmp/lila/}db...
[info] Updating {file:/home/thib/tmp/lila/}hub...
[info] Updating {file:/home/thib/tmp/lila/}memo...
[info] Done updating.
[info] Done updating.
[info] Done updating.
[warn] There may be incompatibilities among your library dependencies.
[info] Updating {file:/home/thib/tmp/lila/}rating...
[info] Updating {file:/home/thib/tmp/lila/}coordinate...
[warn] Here are some of the libraries that were evicted:
[warn]  * com.google.guava:guava:16.0.1 -> 18.0
[warn] Run 'evicted' to see detailed eviction warnings
[info] Updating {file:/home/thib/tmp/lila/}wiki...
[info] Compiling 22 Scala sources to /home/thib/tmp/lila/modules/common/target/scala-2.11/classes...
[info] Done updating.
[info] Updating {file:/home/thib/tmp/lila/}search...
[info] Done updating.
[info] Updating {file:/home/thib/tmp/lila/}socket...
[...]

PR #2097, opened 2015-07-10
Fixes #2046. Cached resolution: Fixes double eviction

### LINKED ISSUE #2046: cached resolution falsely evicts netty (spark + cassandra)
I ran into a failure with cached resolution, below is a minimized example. The bug doesn't occur with cachedResolution disabled.

Spark includes a dependency on netty-3.8.0.Final, and the cassandra driver includes a dependency on netty-handler-4.x. These two variants of netty don't conflict, and both need to be included. If cached resolution is enabled, however, the 3.8 version is incorrectly evicted. 

A workaround is to force the version of the cassandra driver with dependencyOverrides. That doesn't change the version of the cassandra driver selected, but it avoids the bug.

A quick way to see the problem is to copy the following build.sbt, and then:

```
 $ sbt
 > deps netty
```

then comment out the dependency override and:

```
> reload
> update
> deps netty
```

note that netty 3.8 isn't in on the list in the second case.

``` scala
val vSpark = "1.3.1"
val sparkCore             = "org.apache.spark"         %% "spark-core"                % vSpark
val sparkCassandra        = "com.datastax.spark"       %% "spark-cassandra-connector" % "1.3.0-M1"
val cassandraDriver       = "com.datastax.cassandra"    %  "cassandra-driver-core"    % "2.1.6"

import sbt.complete.DefaultParsers._

  settings(
    scalaVersion := "2.11.6",
    updateOptions := updateOptions.value.withCachedResolution(true),  // this breaks it
    dependencyOverrides += cassandraDriver,   // this fixes it
    libraryDependencies ++= Seq(
      sparkCore,
      cassandraDriver,
      sparkCassandra
    ),
    deps := {
      val matchString = (Space ~> token(StringBasic,"<search string>")).parsed
      val classpath = (fullClasspath in Compile).value
      val found = classpath.map(_.data.toString).filter(_.contains(matchString))
      found foreach println
    }
  )

lazy val deps = inputKey[Unit]("search classpath")
```

eed3si9n:
Fixes #2046

This PR is against 0.13.9 branch, intending to fix cached resolution correctness issue.
The issue is fixed by sorting the stitched graph prior to resolving any conflicts.
To sort the stitched graph, we need the caller info.

Simply bringing back the entire caller list resulted in #1763 regression. I was able to work around this by removing all but one artificial callers.

/review @jsuereth, @dwijnand

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -250,7 +250,7 @@ lazy val ivyProj = (project in file("ivy")).
   settings(
     baseSettings,
     name := "Ivy",
-    libraryDependencies ++= Seq(ivy, jsch, sbtSerialization, launcherInterface),
+    libraryDependencies ++= Seq(ivy, jsch, sbtSerialization, scalaReflect.value, launcherInterface),
     testExclusive)
 
 // Runner for uniform test interface
@@ -293,7 +293,7 @@ lazy val cacheProj = (project in cachePath).
   settings(
     baseSettings,
     name := "Cache",
-    libraryDependencies ++= Seq(sbinary, sbtSerialization) ++ scalaXml.value
+    libraryDependencies ++= Seq(sbinary, sbtSerialization, scalaReflect.value) ++ scalaXml.value
   )
 
 // Builds on cache to provide caching for filesystem-related operations
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -14,6 +14,7 @@ object Dependencies {
   lazy val sbinary = "org.scala-tools.sbinary" %% "sbinary" % "0.4.2"
   lazy val sbtSerialization = "org.scala-sbt" %% "serialization" % "0.1.1"
   lazy val scalaCompiler = Def.setting { "org.scala-lang" % "scala-compiler" % scalaVersion.value }
+  lazy val scalaReflect = Def.setting { "org.scala-lang" % "scala-reflect" % scalaVersion.value }
   lazy val testInterface = "org.scala-sbt" % "test-interface" % "1.0"
   lazy val scalaCheck = "org.scalacheck" %% "scalacheck" % "1.11.4"
   lazy val specs2 = "org.specs2" %% "specs2" % "2.3.11"
--- a/ivy/src/main/scala/sbt/JsonUtil.scala
+++ b/ivy/src/main/scala/sbt/JsonUtil.scala
@@ -8,7 +8,8 @@ import sbt.serialization._
 import java.net.{ URLEncoder, URLDecoder }
 
 private[sbt] object JsonUtil {
-  val fakeCallerOrganization = "org.scala-sbt.temp-callers"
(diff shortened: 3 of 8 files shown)


build.sbt
```diff
@@ -250,7 +250,7 @@ lazy val ivyProj = (project in file("ivy")).
   settings(
     baseSettings,
     name := "Ivy",
-    libraryDependencies ++= Seq(ivy, jsch, sbtSerialization, launcherInterface),
+    libraryDependencies ++= Seq(ivy, jsch, sbtSerialization, scalaReflect.value, launcherInterface),
```
jsuereth, line 253: uh-oh

eed3si9n, line 253: We bumped up Scala version for compiler, so reflect needs to come up together.

project/Dependencies.scala
```diff
@@ -14,6 +14,7 @@ object Dependencies {
   lazy val sbinary = "org.scala-tools.sbinary" %% "sbinary" % "0.4.2"
   lazy val sbtSerialization = "org.scala-sbt" %% "serialization" % "0.1.1"
   lazy val scalaCompiler = Def.setting { "org.scala-lang" % "scala-compiler" % scalaVersion.value }
+  lazy val scalaReflect = Def.setting { "org.scala-lang" % "scala-reflect" % scalaVersion.value }
```
jsuereth, line 17: I still don't see why this was needed....

jsuereth:
hard for me to follow the nitty-grittys, but generally LGTM.  Tests look good.

project/Dependencies.scala
eed3si9n, line 17: Pickling needs reflection because of macro.

eed3si9n:
I switched to using sbt/serialization in Defaults.scala to do cached update, but I am now thinking I should back it out since there are non-cached resolution cases that still has insane amount of callers. This is ultimately an Ivy issue.

project/Dependencies.scala
eed3si9n, line 17: In other words, Pickling will pull in the version of scala-reflect that it used to build the release. I think I last used 2.10.5 thinking sbt would need it.

https://github.com/scala/pickling/blob/581f4b55b97eb7a200ca49661dcceaa356f4003e/build.sbt#L60-L61

eed3si9n, line 17: Turns out, the version of Pickling that we use is not the latest - https://github.com/sbt/serialization/blob/master/project/Dependencies.scala

eed3si9n:
@jsuereth backing out 98d81a62f92b9d13d1c3a658e140ad9daa227261 fixed the build. (or because I deleted the Travis cache?)

jsuereth:
Probably backing out 98d81a6.... hmm...

jsuereth:
If we get two repeated success, ok to merge

eed3si9n:
It's been 2-0 with "push" and "pr".

PR #2098, opened 2015-07-10
Fix most SBT's build warnings

pdalpra:
This PR fixes most (2 remain) warnings in SBT's own build.

Here are those which remain : 

```
/Users/pierredalpra/Work/Forks/sbt/build.sbt:12: warning: postfix operator intransitive should be enabled
by making the implicit value scala.language.postfixOps visible.
This can be achieved by adding the import clause 'import scala.language.postfixOps'
or by setting the compiler option -language:postfixOps.
See the Scala docs for value scala.language.postfixOps for a discussion
why the feature should be explicitly enabled.
def buildLevelSettings: Seq[Setting[_]] = Seq(
                                         ^
```

```
/Users/pierredalpra/Work/Forks/sbt/build.sbt:555: warning: method normalize in object StringUtilities is deprecated: Different use cases require different normalization.  Use Project.normalizeModuleID or normalizeProjectID instead.
def precompiled(scalav: String): Project = Project(id = normalize("Precompiled " + scalav.replace('.', '_')), base = compilePath / "interface").
```

I'm having quite a hard time to understand where the first one comes from with my basic knowledge of SBT's internals...

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -108,8 +108,7 @@ lazy val interfaceProj = (project in file("interface")).
     watchSources <++= apiDefinitions,
     resourceGenerators in Compile <+= (version, resourceManaged, streams, compile in Compile) map generateVersionFile,
     apiDefinitions <<= baseDirectory map { base => (base / "definition") :: (base / "other") :: (base / "type") :: Nil },
-    sourceGenerators in Compile <+= (cacheDirectory,
-      apiDefinitions,
+    sourceGenerators in Compile <+= (apiDefinitions,
       fullClasspath in Compile in datatypeProj,
       sourceManaged in Compile,
       mainClass in datatypeProj in Compile,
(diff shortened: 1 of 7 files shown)


typesafe-tools:

jsuereth:
LGTM, thanks for the fixes!

jsuereth:
For the normalize bit, you want `normalizeProjectId`

pdalpra:
@jsuereth should I simply expect that the normalization could be done and get using `normalizeProjectID (...).right.get` to extract it from the `Either` ?

jsuereth:
@pdalpra yeah, for now.  The reality is we should probably just make something which is already normalized :).  I also think `precompiled` libs should be going away after 0.13.9

pdalpra:
Okay, I'll add that to my PR then :)
Any clues on the first warning ?

dwijnand:
> I also think precompiled libs should be going away after 0.13.9

@jsuereth Do you mean with @Duhemm efforts, or something else?

jsuereth:
@pdalpra No, looks liek a bug in detection somehow....

jsuereth:
@dwijnand With the @Duhemm efforts

Issue #2087, opened 2015-07-03
logLevel := Level.Debug should output javaOptions


viktorklang:
To be able to verify that javaOptions have been applied (for instance with fork := true),
the javaOptions which are used should be logged at Debug level.

dwijnand:
Closed with #2103.

viktorklang:
Thank you! :clap: :+1:

Issue #2041, opened 2015-05-28
sbt should issue a warning about ignoring javaOptions when fork is set to false in build.sbt


stammana:
The following is not being honored when fork is set to false.

javaOptions in Test += "-Dconfig.file=conf/application_test.conf"

It works fine when the fork parameter is removed from build.sbt

sbt version 0.13.5.

pdalpra:
From SBT itself: 

```
> help javaOptions
Options passed to a new JVM when forking.
```

`javaOptions` is only used when forking new JVMs.

stammana:
the fork is set to false not to true. what is the default value for fork in sbt?

pdalpra:
Again, from SBT itself : 

```
> show test:fork
[info] false
```

SBT does not fork new JVMs by default.

stammana:
> show test:fork
> [info] true

Getting it as true by default on version 0.13.5

pdalpra:
It must have been overriden someplace else in your build, possibly from a plugin, as `false` is indeed the default value : https://github.com/sbt/sbt/blob/0.13.5/main/src/main/scala/sbt/Defaults.scala#L152

stammana:
seems like you are correct, but why is it not honoring my javaOptions when fork is set to false.

pdalpra:
See my previous comment. `javaOptions` is only used when forking new JVMs. If `fork in Test` is set to `false`, `javaOptions in Test` is simply not used, since you're not forking new JVMs.

stammana:
It would have been great if it takes java options when fork is set to false. I don't need to supply it from the command line. I can simply say "sbt clean test". 

Can this be an enhancement?

jsuereth:
@stammana Not really.   By the time sbt has read your build file, the JVM is already instantiated and we can't send javaOptions to it.  The only way to send java options is to fork a new JVM where you can pass them in, you cannot set them after it's been started.

So, you can either set javaOptions to true, or you can use one of the startup script tricks of creating a .sbtopts file and putting your java options in there.

At a maximum, we'll just issue a warning when people do this.

dwijnand:
Closed with #2103

PR #2103, opened 2015-07-13
Log javaOptions/fork interactions

pdalpra:
This fixes both #2041 and #2087.
- When forking, log the javaOptions that are used
- When javaOptions are defined but fork := false, warn that javaOptions
  will be ignored

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -413,7 +413,7 @@ object Defaults extends BuildCommon {
     definedTests <<= detectTests,
     definedTestNames <<= definedTests map (_.map(_.name).distinct) storeAs definedTestNames triggeredBy compile,
     testFilter in testQuick <<= testQuickFilter,
-    executeTests <<= (streams in test, loadedTestFrameworks, testLoader, testGrouping in test, testExecution in test, fullClasspath in test, javaHome in test, testForkedParallel) flatMap allTestGroupsTask,
+    executeTests <<= (streams in test, loadedTestFrameworks, testLoader, testGrouping in test, testExecution in test, fullClasspath in test, javaHome in test, testForkedParallel, javaOptions in test) flatMap allTestGroupsTask,
     testResultLogger in (Test, test) :== TestResultLogger.SilentWhenNoTests, // https://github.com/sbt/sbt/issues/1185
     test := {
       val trl = (testResultLogger in (Test, test)).value
@@ -519,7 +519,7 @@ object Defaults extends BuildCommon {
         implicit val display = Project.showContextKey(state.value)
         val modifiedOpts = Tests.Filters(filter(selected)) +: Tests.Argument(frameworkOptions: _*) +: config.options
         val newConfig = config.copy(options = modifiedOpts)
-        val output = allTestGroupsTask(s, loadedTestFrameworks.value, testLoader.value, testGrouping.value, newConfig, fullClasspath.value, javaHome.value, testForkedParallel.value)
+        val output = allTestGroupsTask(s, loadedTestFrameworks.value, testLoader.value, testGrouping.value, newConfig, fullClasspath.value, javaHome.value, testForkedParallel.value, javaOptions.value)
         val taskName = display(resolvedScoped.value)
         val trl = testResultLogger.value
         val processed = output.map(out => trl.run(s.log, out, taskName))
@@ -542,19 +542,27 @@ object Defaults extends BuildCommon {
   }
 
   def allTestGroupsTask(s: TaskStreams, frameworks: Map[TestFramework, Framework], loader: ClassLoader, groups: Seq[Tests.Group], config: Tests.Execution, cp: Classpath, javaHome: Option[File]): Task[Tests.Output] = {
-    allTestGroupsTask(s, frameworks, loader, groups, config, cp, javaHome, forkedParallelExecution = false)
+    allTestGroupsTask(s, frameworks, loader, groups, config, cp, javaHome, forkedParallelExecution = false, javaOptions = Nil)
   }
 
   def allTestGroupsTask(s: TaskStreams, frameworks: Map[TestFramework, Framework], loader: ClassLoader, groups: Seq[Tests.Group], config: Tests.Execution, cp: Classpath, javaHome: Option[File], forkedParallelExecution: Boolean): Task[Tests.Output] = {
+    allTestGroupsTask(s, frameworks, loader, groups, config, cp, javaHome, forkedParallelExecution, javaOptions = Nil)
+  }
+
+  def allTestGroupsTask(s: TaskStreams, frameworks: Map[TestFramework, Framework], loader: ClassLoader, groups: Seq[Tests.Group], config: Tests.Execution, cp: Classpath, javaHome: Option[File], forkedParallelExecution: Boolean, javaOptions: Seq[String]): Task[Tests.Output] = {
     val runners = createTestRunners(frameworks, loader, config)
     val groupTasks = groups map {
       case Tests.Group(name, tests, runPolicy) =>
         runPolicy match {
           case Tests.SubProcess(opts) =>
+            s.log.debug(s"javaOptions: ${opts.runJVMOptions}")
             val forkedConfig = config.copy(parallel = config.parallel && forkedParallelExecution)
             s.log.debug(s"Forking tests - parallelism = ${forkedConfig.parallel}")
             ForkTests(runners, tests.toList, forkedConfig, cp.files, opts, s.log) tag Tags.ForkedTestGroup
           case Tests.InProcess =>
+            if (javaOptions.nonEmpty) {
+              s.log.warn("javaOptions will be ignored, fork is set to false")
+            }
             Tests(frameworks, loader, runners, tests, config, s.log)
         }
     }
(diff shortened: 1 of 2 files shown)


typesafe-tools:

dwijnand:
1) This requires notes.
2) You're breaking bincompat changing those public methods
3) It's interesting to note the different approach here compared to https://github.com/Duhemm/sbt/commit/fc6a3373058c6b7d7c7a492b05196ba2318b41d0, ie. logging as late as possible with what information percolates down (Martin's way), vs logging as early as possible at ever entry point (your way). It's a trade-off between information availability and making sure a log always happen. I'm interested in what other think.

pdalpra:
1) Indeed, I waited for the PR creation, since the PR number is often included in the notes
2) Argh, I'll need to see if I can get the javaOptions in another way to allGroupTestTasks (or somewhere else).
3) On the topic of logging the javaOptions that are used, @Duhemm 's solution is without hesitation a better solution, since it avoids code duplication. However, this can't work for 2041, since the issue is that we're not forking but javaOptions are still provided and ignored.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -541,20 +541,24 @@ object Defaults extends BuildCommon {
     }
   }
 
-  def allTestGroupsTask(s: TaskStreams, frameworks: Map[TestFramework, Framework], loader: ClassLoader, groups: Seq[Tests.Group], config: Tests.Execution, cp: Classpath, javaHome: Option[File]): Task[Tests.Output] = {
-    allTestGroupsTask(s, frameworks, loader, groups, config, cp, javaHome, forkedParallelExecution = false)
+  def allTestGroupsTask(s: TaskStreams, frameworks: Map[TestFramework, Framework], loader: ClassLoader, groups: Seq[Tests.Group], config: Tests.Execution, cp: Classpath, javaHome: Option[File], javaOptions: Seq[String]): Task[Tests.Output] = {
```
jsuereth: This isn't private, so we can't add an option to it without also created an overloaded method that has the original # of arguments (binary compatibility)

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -541,20 +541,24 @@ object Defaults extends BuildCommon {
...
+    allTestGroupsTask(s, frameworks, loader, groups, config, cp, javaHome, javaOptions, forkedParallelExecution = false)
   }
 
-  def allTestGroupsTask(s: TaskStreams, frameworks: Map[TestFramework, Framework], loader: ClassLoader, groups: Seq[Tests.Group], config: Tests.Execution, cp: Classpath, javaHome: Option[File], forkedParallelExecution: Boolean): Task[Tests.Output] = {
+  def allTestGroupsTask(s: TaskStreams, frameworks: Map[TestFramework, Framework], loader: ClassLoader, groups: Seq[Tests.Group], config: Tests.Execution, cp: Classpath, javaHome: Option[File], javaOptions: Seq[String], forkedParallelExecution: Boolean): Task[Tests.Output] = {
```
jsuereth: Similar here.

jsuereth: Also, can we just use `opts.runJVMOptions` to check if they are empty, rather than pass down into this function?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -739,7 +743,17 @@ object Defaults extends BuildCommon {
...
     val si = scalaInstance.value
-    if (fork.value) new ForkRun(forkOptions.value) else new Run(si, trapExit.value, tmp)
+    val s = streams.value
+    val options = javaOptions.value
+    if (fork.value) {
+      s.log.debug(s"javaOptions: $options")
+      new ForkRun(forkOptions.value)
+    } else {
+      if (options.nonEmpty) {
+        s.log.warn("javaOptions will be ignored, fork is set to false")
```
jsuereth: Thanks so much for adding this, will help a lot.

jsuereth:
@dwijnand You win for reviewing while I had this open and waiting, thereby misisng your comments completely.  Agree with your comments.

main/src/main/scala/sbt/Defaults.scala
pdalpra: See below, this is what I do when the `runPolicy` is `SubProcess`, as I have access to the `ForkOptions`.
Unfortunately, I cannot do that when the Strategy is `InProcess`...

pdalpra:
So, should I remove the `javaOptions` part and leave it to @Duhemm, or should I implement his solution by logging in `ForkJava`/`LocalJava` ?

pdalpra:
Added notes.

dwijnand:
@pdalpra Do we log the same (or "sufficiently" similar) information and in all the same cases by moving it to `ForkJava`/`LocalJava`? If so then I think as late as possible is better.

pdalpra:
@dwijnand By moving the logging to `ForkJava`/`LocalJava`, we can only fix #2087 : we're forking and we log the options that are used.
#2041 needs to be done upfront, since we're not forking.

@jsuereth @dwijnand I just pushed a fix for the bincompat issue, by addind an overload for `allGroupTestTasks`.

dwijnand:
Doesn't `LocalJava` deal with the non-forked version?

pdalpra:
`LocalJava` seems to deal only with compilation :/

pdalpra:
More precisely, `LocalJava` goal (as stated by its documentation : `Helper methods for trying to run the java toolchain out of our own classloaders.`) is to run the Javadoc tool or `javac` in process.
After all, running a main class or tests in process boils down to calling a method.

pdalpra:
BTW, what @Duhemm has done on his fork does not affects the `run` or `test` tasks, as what he has done logs the JVM options used for `javac`/`javadoc`.

main/src/main/scala/sbt/Defaults.scala
jsuereth: Also, thinking about this, it's not actually telling WHICH CONFIGURATION has javaOptions and fork set to false.  You may actually want to grab the magic "what's my current configuration" dependency so you can report which one is the problem, otherwise it'll just be popping out and giving the user no idea how to solve it.  GIve me a abit and I'll look up the syntax.

jsuereth: Ok, it looks like the best you can do is depend on the `sbt.Keys.configuration` and grab the currently active configuration.  We also need you to grab the scoped "task", ideally, so you can tell users where the bad setting is.

jsuereth:
BC changes look good, was just commenting on what the error messages look like in practice (From a quick test).

Ideally I'd like to know, if I set `fork in (Compile, run) := false` and `javaOptions in Compile := Seq(...)` I'd like to know WHICH `runner` is the problematic one.

pdalpra:
Indeed, as it is, It's nothing more than a hint that something is wrong for any not-completely-trivial build.
Will fix ASAP.

pdalpra:
I have two small questions: 
- if we are to print the configuration, this mean that we need to add yet another parameter to `allGroupTestTasks`, which is not so hard in itself. However, as we'll need to provide a default value for that parameter, which configuration should be used as the default value ?
- Is there a simple way to get the name of the Scope ? I found `Scope.display`, should I use it ?

main/src/main/scala/sbt/Defaults.scala
pdalpra: I'm still missing some info to get the scoped task, but that's where I am right now :

```
[warn] */*:javaOptions will be ignored, */*:fork is set to false
```

jsuereth:
So, as for allGroupTestTasks... hmm... Maybe let's defer that for now.  I think there may be an alternative way to get the error message on that setting.

As for Scope.display, that should be fine.   I personally get confused over the different show options EVERY TIME  I need to show something, so you should try it out and check the error messages.   if it's wrong, we'll find you the right one :)

jsuereth:
Also, thanks @pdalpra for the dedication on this one.  Very glad to see you going the extra mile on error messages :)

pdalpra:
As shown above, I managed to get something that looks quite good, except for the scopes...
Is `javaOptions.scopedKey` the right way to get the key with its scope ?

You're welcome :) 
After all, if we're going to add log messages, let's do it right, no ? ;)

jsuereth:
no, I dont think `javaOptions.scopedKey` is doing the right thing...

pdalpra:
It seems that what is done in `SettingsGraph` (used by `inspect tree`) could get me to the info I want, but it looks like it's overkill just to get scoped `run` and `javaOptions` :/

jsuereth:
@pdalpra I'll have to look. The issue is NOT that you want the scopedKey of the current key, it's that you want to find the scope of the current Setting you're a part of.

SO, for ANY key, you can do:  `key.scopedKey`.  This is because, internally, sbt uses ScopedKeys for its setting system.  A scoped key is just an `AttributeKey` and a `Scope`.  `TaskKey` and `SettingKey` and friends are the user-facing DSL for these.  They allow you to manipulate them and also keep the high level API for things.

If you just call `javaOptions.scopedKey` you'll get a `ScopeKey` with the default scope.   If you're inspecting a setting _AFTER_ setting resolution, you'll see the real scope for the key (we do translation/resolution during a build-compile step).

SO yes, `SettingGraph` is looking at the Finalized `ScopedKey` (our internal API).  What you see in your setting is _before_ this step, so the scopedKey would be wrong.  

WHat we do have in sbt, are a few "magic" keys that will be injected by the setting system.  For example, `taskStreams` is a setting which gets injected based on the scope of the key the setting is being applied to.   We expose _ONE_ of these _magic_ keys called "configuration" which you can use to grab the current configuration you're being applied to.   See:  https://github.com/sbt/sbt/blob/0.13/main/src/main/scala/sbt/Load.scala#L142-L168

TL;DR;  Try to depend on `resolvedScope` key and that should give you the right scope.

pdalpra:
Thank you for your detailed explanation, I'll try it ASAP.

pdalpra:
Got it : 

```
[warn] {file:/Users/pdalpra/test/}test/compile:run::javaOptions will be ignored, {file:/Users/pdalpra/test/}test/compile:run::fork is set to false
```

I only logged scopes on the warning, not on the debug that display the javaOptions.

Can't get that level of details to `allGroupTestTasks` however.

pdalpra:
Amended my commit so that the project axis in not shown in the warning message :

```
[warn] compile:run::javaOptions will be ignored, compile:run::fork is set to false
```

jsuereth:
@pdalpra nice!

dwijnand:
Nice indeed!

pdalpra:
BTW, thank you for your patience with me on this @jsuereth !

Issue #681, opened 2013-02-22
Add support for Scala version specific sources in compiler interface


gkossakowski:
Sbt managed to maintain one code base for the compiler interface project that was shared across all version of Scala compiler. The compiler interface is responsible for interacting with Scala compiler internals that perform dependency tracking and API extraction. It managed to maintain source compatibility by applying various trick that can be seen here:
https://github.com/sbt/sbt/blob/0.13/compile/interface/src/main/scala/xsbt/Analyzer.scala#L126

However, with the work I'm doing to track dependencies in a more fine-grained way we'll hit the wall of not being able to for example support dependency tracking for macros and supporting 2.9 compiler at the same time. See [this](https://groups.google.com/forum/?fromgroups=#!topic/scala-internals/b95Y-GbXVGA) thread on scala-internals for initial discussion about how to achieve that.

We can still share most of the code of compiler interface but we'll definitively need some custom code per Scala major version. It's worth noting that sbt already ships precompiled versions of the compiler interface so that part wouldn't be affected (apart from the fact that compilation of precompiled interface would need to chose the right version of certain sources). The more tricky problem is with artifact that stores compiler interface sources and is being used in case precompiled interface is missing for particular version of Scala compiler. We have two choices:
- Support only latest major release when it comes to `compiler-interface-src` artifact so we ship only one version like right now and it only works with latest major version of Scala at the time sbt was released;
- Include version number of major version Scala compiler in name of `compiler-interface-src` artifact as it's done with any other cross-versioned artifact

One way or another this seems like fairly involved refactoring so I'm filing a ticket for that.

gkossakowski:
@harrah: I'd love to hear your input on that. I consider this to be critical for sbt 0.13 release. Not blocker in a sense that we can ship sbt 0.13 without this feature but it will mean there's little chance we can properly support incremental compilation in case macros are involved.

Since it's fairly difficult refactoring of parts of sbt I don't understand I think I'll need you to work on this if you consider it to be important.

harrah:
Why isn't the dependency tracking code in the compiler?

gkossakowski:
[...]

PR #2106, opened 2015-07-15
Find most specific version of compiler interface sources

Duhemm:
Ref https://github.com/sbt/sbt/issues/681

This commit introduces a mechanism that allows sbt to find the most
specific version of the compiler interface sources that exists using
Ivy.

For instance, when asked for a compiler interface for Scala 2.11.8-M2,
sbt will look for sources for:
- 2.11.8-M2 ;
- 2.11.8 ;
- 2.11 ;
- the default sources.

This commit also modifies the build definition by removing the
precompiled projects and configuring the compiler-interface project so
that it publishes its source artifacts in a Maven-friendly format.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -577,6 +577,16 @@ def customCommands: Seq[Setting[_]] = Seq(
     "reload" ::
     state
   },
+  /** There are several complications with sbt's build.
+   * First is the fact that interface project is a Java-only project
+   * that uses source generator from datatype subproject in Scala 2.10.5.
+   *
+   * Second is the fact that all subprojects are released with crossPaths
+   * turned off for the sbt's Scala version 2.10.5, but some of them are also
+   * cross published against 2.11.1 with crossPaths turned on.
+   *
+   * `so compile` handles 2.10.x/2.11.x cross building.
+   */
   commands += Command.command("release-sbt") { state =>
     // TODO - Any sort of validation
     "clean" ::
--- a/build.sbt
+++ b/build.sbt
@@ -316,8 +316,7 @@ lazy val compileInterfaceProj = (project in compilePath / "interface").
     // needed because we fork tests and tests are ran in parallel so we have multiple Scala
     // compiler instances that are memory hungry
     javaOptions in Test += "-Xmx1G",
-    publishArtifact in (Compile, packageSrc) := true,
-    publishMavenStyle := true
+    publishArtifact in (Compile, packageSrc) := true
   )
 
 // Implements the core functionality of detecting and propagating changes incrementally.
--- a/build.sbt
+++ b/build.sbt
@@ -81,17 +81,6 @@ lazy val bundledLauncherProj =
     packageBin in Compile := sbtLaunchJar.value
   )
 
-
-// This is used only for command aggregation
-lazy val allPrecompiled: Project = (project in file("all-precompiled")).
-  aggregate(precompiled282, precompiled292, precompiled293).
-  settings(
-    buildLevelSettings,
-    minimalSettings,
-    publish := {},
-    publishLocal := {}
-  )
-
 /* ** subproject declarations ** */
 
 // defines Java structures used across Scala versions, such as the API structures and relationships extracted by
@@ -199,7 +188,7 @@ lazy val logProj = (project in utilPath / "log").
     testedBaseSettings,
     name := "Logging",
     libraryDependencies += jline
-  ) 
+  )
 
 // Relation
 lazy val relationProj = (project in utilPath / "relation").
@@ -317,7 +306,7 @@ lazy val compileInterfaceProj = (project in compilePath / "interface").
   dependsOn(interfaceProj % "compile;test->test", ioProj % "test->test", logProj % "test->test", /*launchProj % "test->test",*/ apiProj % "test->test").
   settings(
     baseSettings,
-    precompiledSettings,
+    libraryDependencies += scalaCompiler.value % "provided",
     name := "Compiler Interface",
     exportJars := true,
     // we need to fork because in unit tests we set usejavacp = true which means
@@ -327,12 +316,9 @@ lazy val compileInterfaceProj = (project in compilePath / "interface").
     // needed because we fork tests and tests are ran in parallel so we have multiple Scala
     // compiler instances that are memory hungry
     javaOptions in Test += "-Xmx1G",
-    artifact in (Compile, packageSrc) := Artifact(srcID).copy(configurations = Compile :: Nil).extra("e:component" -> srcID)
+    publishArtifact in (Compile, packageSrc) := true,
+    publishMavenStyle := true
   )
-
-lazy val precompiled282 = precompiled(scala282)
-lazy val precompiled292 = precompiled(scala292)
-lazy val precompiled293 = precompiled(scala293)
 
 // Implements the core functionality of detecting and propagating changes incrementally.
 //   Defines the data structures for representing file fingerprints and relationships and the overall source analysis
@@ -442,7 +428,7 @@ lazy val mainProj = (project in mainPath).
 //  technically, we need a dependency on all of mainProj's dependencies, but we don't do that since this is strictly an integration project
 //  with the sole purpose of providing certain identifiers without qualification (with a package object)
 lazy val sbtProj = (project in sbtPath).
-  dependsOn(mainProj, compileInterfaceProj, precompiled282, precompiled292, precompiled293, scriptedSbtProj % "test->test").
+  dependsOn(mainProj, compileInterfaceProj, scriptedSbtProj % "test->test").
   settings(
     baseSettings,
     name := "sbt",
@@ -512,7 +498,7 @@ def otherRootSettings = Seq(
   }
 ))
 lazy val docProjects: ScopeFilter = ScopeFilter(
-  inAnyProject -- inProjects(sbtRoot, sbtProj, scriptedBaseProj, scriptedSbtProj, scriptedPluginProj, precompiled282, precompiled292, precompiled293, mavenResolverPluginProj),
+  inAnyProject -- inProjects(sbtRoot, sbtProj, scriptedBaseProj, scriptedSbtProj, scriptedPluginProj, mavenResolverPluginProj),
   inConfigurations(Compile)
 )
 def fullDocSettings = Util.baseScalacOptions ++ Docs.settings ++ Sxr.settings ++ Seq(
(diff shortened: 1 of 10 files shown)


typesafe-tools:

compile/ivy/src/main/scala/sbt/compiler/ComponentCompiler.scala
```diff
@@ -64,4 +94,123 @@ class ComponentCompiler(compiler: RawCompiler, manager: ComponentManager) {
       manager.define(binID, Seq(targetJar))
     }
   }
-}
\ No newline at end of file
+}
+
+private[compiler] class IvyComponentCompiler(compiler: RawCompiler, manager: ComponentManager, ivyConfiguration: IvyConfiguration, log: Logger) {
```
jsuereth, line 88: Glad to see this has a name :).  Can you add a comment on what this class is meant to do?

compile/ivy/src/test/sbt/compiler/CascadingVersionsSpecifications.scala
```diff
@@ -0,0 +1,50 @@
+package sbt
+package compiler
+
+import org.junit.runner.RunWith
+import org.specs2.mutable.Specification
+import org.specs2.runner.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class CascadingVersionsSpecification extends Specification {
```
jsuereth: Did you add this to teh "safeUnitTests" list.  I realize we have been forgetting to add new unit-test projects to this.

main/actions/src/main/scala/sbt/Compiler.scala
```diff
@@ -80,6 +95,13 @@ object Compiler {
       val scalac = scalaCompiler(instance, cpOptions)
       new Compilers(scalac, javac)
     }
+  @deprecated("Deprecated in favor of new sbt.compiler.javac package.", "0.13.8")
+  def compilers(instance: ScalaInstance, cpOptions: ClasspathOptions, javac: JavaTool, ivyConfiguration: IvyConfiguration)(implicit app: AppConfiguration, log: Logger): Compilers =
```
jsuereth: Do we need to duplicate a deprecated method?

jsuereth:
Genereally, LGTM.  Great work!

main/actions/src/main/scala/sbt/Compiler.scala
Duhemm: It is used by (line 72 in this file)

``` scala
def compilers(instance: ScalaInstance, cpOptions: ClasspathOptions, javaHome: Option[File], ivyConfiguration: IvyConfiguration)(implicit app: AppConfiguration, log: Logger): Compilers
```

What is supposed to be the replacement for this method? `sbt.compiler.MixedAnalyzingCompiler`?

Duhemm:
Feedback addressed, thanks!

main/actions/src/main/scala/sbt/Compiler.scala
jsuereth: Yeah, maybe just inline the method in line 72 and let's not add a new, already deprecated, method.

compile/ivy/src/main/scala/sbt/compiler/ComponentCompiler.scala
```diff
@@ -23,12 +25,40 @@ object ComponentCompiler {
...
+   */
+  private[compiler] def cascadingSourceModuleVersions(fullVersion: String): Seq[String] = {
+
+    Try {
+      val VersionNumber(numbers @ maj +: min +: _, tags, _) = VersionNumber(fullVersion)
+      Seq(fullVersion,
+        numbers.mkString(".") + (if (tags.nonEmpty) tags.mkString("-", "-", "") else ""),
+        numbers.mkString("."),
+        s"$maj.$min"
+      ).distinct
```
eed3si9n: This part feels stringy. I wonder if it could be represented as `Vector[VersionNumber]` and fold this cascade logic also into the class.

build.sbt
```diff
@@ -616,36 +572,14 @@ def customCommands: Seq[Setting[_]] = Seq(
...
-   * the precompiled projects (that uses Scala 2.8.2 etc.)
-   * 
-   * Finally, there's the fact that all subprojects are released with crossPaths
-   * turned off for the sbt's Scala version 2.10.4, but some of them are also
-   * cross published against 2.11.1 with crossPaths turned on.
-   *
-   * Because of the way ++ (and its improved version wow) is implemented
-   * precompiled compiler briges are handled outside of doge aggregation on root.
-   * `so compile` handles 2.10.x/2.11.x cross building. 
-   */
```
eed3si9n, line 642: Aren't some babies are thrown with the bath water here? For example the note on doge still stands.

compile/ivy/src/main/scala/sbt/compiler/ComponentCompiler.scala
```diff
@@ -14,6 +15,7 @@ object ComponentCompiler {
   val compilerInterfaceSrcID = compilerInterfaceID + srcExtension
   val javaVersion = System.getProperty("java.class.version")
 
+  @deprecated("Please specify `ivyConfiguration`.", "0.13.10")
```
eed3si9n: Avoid "please" and "thank you"s. The imperative sentence is normally fine e.g. "press any keys."

compile/ivy/src/main/scala/sbt/compiler/ComponentCompiler.scala
```diff
@@ -64,4 +94,129 @@ class ComponentCompiler(compiler: RawCompiler, manager: ComponentManager) {
...
+    }
+  }
+
+  /**
+   * Returns a dummy module that depends on "org.scala-sbt" % `id` % `sbtVersion`.
+   * Note: Sbt's implementation of Ivy requires us to do this, because only the dependencies
+   *       of the specified module will be downloaded.
+   */
+  private def getModule(id: String): ivySbt.Module = {
+    val dummyID = ModuleID(xsbti.ArtifactInfo.SbtOrganization, scala.util.Random.alphanumeric take 20 mkString "", sbtVersion, Some("component"))
```
eed3si9n: I'd suggest against using the actual "org.scala-sbt", but use "org.scala-sbt.temp" instead. That's what I do for cached resolution.

eed3si9n: The name part of the `ModuleID` should also be prefixed with something like "temp-module": https://github.com/sbt/sbt/blob/v0.13.9-RC2/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala#L105

Also in this new design, how often will this random module be created? Are you checking for the target module in Ivy cache prior to generating this? If not we should use SHA1 of the target module ID so we don't accumulate junk.

build.sbt
```diff
@@ -327,13 +316,10 @@ lazy val compileInterfaceProj = (project in compilePath / "interface").
     // needed because we fork tests and tests are ran in parallel so we have multiple Scala
     // compiler instances that are memory hungry
     javaOptions in Test += "-Xmx1G",
-    artifact in (Compile, packageSrc) := Artifact(srcID).copy(configurations = Compile :: Nil).extra("e:component" -> srcID)
+    publishArtifact in (Compile, packageSrc) := true,
+    publishMavenStyle := true
```
eed3si9n: Is the Maven style necessary here?

main/actions/src/main/scala/sbt/Compiler.scala
```diff
@@ -80,6 +95,13 @@ object Compiler {
...
+    {
+      val scalac = scalaCompiler(instance, cpOptions, ivyConfiguration)
+      new Compilers(scalac, javac)
+    }
+  @deprecated("Please provide an IvyConfiguration", "0.13.10")
```
eed3si9n: The same as above.

eed3si9n:
Yea. Generally this looks pretty awesome. Some minor comments/concerns.
I think this should accompany a page in sbt/website's developer guide section and a quick description and a link in the release note.

compile/ivy/src/main/scala/sbt/compiler/ComponentCompiler.scala
eed3si9n: If you search with `@deprecated` you'll find a bunch of examples like:

``` scala
@deprecated("Use `binaryDependency(File, String, File, DependencyContext)`.", "0.13.8")
```

compile/ivy/src/main/scala/sbt/compiler/ComponentCompiler.scala
jsuereth: Ideally this only corrupts teh resolution cache, but doesn't actually show up in `~/.ivy2/cache`

build.sbt
Duhemm: I don't know if it is necessary, but the idea was to be able to publish the compiler interface in a most maven-friendly way, no?

eed3si9n: To make it Maven friendly, we should sync to Maven central, and thus far we've only taken those steps after we split subproject out to a repository, like sbt/launcher and sbt/io. So I think we think we should wait for it until we have a separate Compiler API, which hopefully we'll have soon.

compile/ivy/src/main/scala/sbt/compiler/ComponentCompiler.scala
Duhemm: I implemented it directly in `VersionNumber`, and moved the tests to `VersionNumberSpec`. The implementation is much shorter now :).

build.sbt
Duhemm, line 642: I finally kept everything that doesn't directly mention the `precompiled` projects. Do the comments that I left make sense?

compile/ivy/src/main/scala/sbt/compiler/ComponentCompiler.scala
Duhemm: Done

compile/ivy/src/main/scala/sbt/compiler/ComponentCompiler.scala
Duhemm: I changed it to do it the same way you do in the file you linked. 
The random module is created pretty often (once per tried source version). 
Do you have an example of checking the Ivy cache?

build.sbt
Duhemm: I removed the maven-style publishing.

main/actions/src/main/scala/sbt/Compiler.scala
Duhemm: Done

eed3si9n:
Thanks for the update. LGTM pending Travis and notes.

eed3si9n:
So this test is failed on Travis twice a row, so it could be legit?

```
[error] DEBUGME: Clearing errors: ArrayBuffer()
[error] DEBUGME: Logging: /tmp/sbt_a3ae04a6/semantic-errors/src/main/scala/bad.scala:2:Error: type mismatch;
[error]  found   : Boolean(false)
[error]  required: Int
[info] [success] Total time: 1 s, completed Jul 16, 2015 12:04:50 PM
[info] [info] Defining *:compileOrder
[info] [info] The new value will be used by compile:compile::compileInputs, test:compile::compileInputs
[info] [info] Reapplying settings...
[info] [info] Set current project to semantic-errors (in build file:/tmp/sbt_a3ae04a6/semantic-errors/)
[info] [info] Compiling 1 Scala source and 1 Java source to /tmp/sbt_a3ae04a6/semantic-errors/target/scala-2.10/classes...
[info] [error] /tmp/sbt_a3ae04a6/semantic-errors/src/main/java/bad.java:3: incompatible types
[info] [error] found   : int
[info] [error] required: bad
[info] [error]     public bad foo() { return 1; }
[info] [error]                               ^
[info] [error] 1 error
[info] java.lang.AssertionError: assertion failed: Failed to report any problems!
```

Duhemm:
`scripted compiler-project/semantic-errors` is fixed.
`scripted dependency-management/mvn-local` passes given that one has flushed Travis' caches before running it.

eed3si9n:
> `scripted dependency-management/mvn-local` passes given that one has flushed Travis' caches before running it.

Could you explain why that is? Do we have to change the test?

jsuereth:
@eed3si9n Martin is seeing odd stale cache issues unless he does that.  The kind of issue that doesn't make any sense (i.e. failure to find a pom.xml file that was just published).

jsuereth:
@eed3si9n I'm cool merging this if you are.

eed3si9n:
Merged

dwijnand:
Shall we merge https://github.com/sbt/website/pull/172 too then?

PR #2108, opened 2015-07-15
Line content from diagnostic classes if available

fkorotkov:
### Problem

While using local java compiler line content in errors is wrong. Only an expression is shown and not a whole line of code.
### Expectation

Get a proper line content.
### Solution

When you run javac from command line com.sun.tools.javac.util.AbstractDiagnosticFormatter#formatSourceLine is used to format an output. There diagnostic classes are used to get line content. Made a change to try to use diagnostic classes if they are available.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -573,7 +573,7 @@ lazy val safeUnitTests = taskKey[Unit]("Known working tests (for both 2.10 and 2
 lazy val safeProjects: ScopeFilter = ScopeFilter(
   inProjects(mainSettingsProj, mainProj, ivyProj, completeProj,
     actionsProj, classpathProj, collectionProj, compileIncrementalProj,
-    logProj, runProj, stdTaskProj),
+    logProj, runProj, stdTaskProj, compilerProj),
   inConfigurations(Test)
 )
 lazy val otherUnitTests = taskKey[Unit]("Unit test other projects")
--- a/compile/src/main/scala/sbt/compiler/javac/DiagnosticsReporter.scala
+++ b/compile/src/main/scala/sbt/compiler/javac/DiagnosticsReporter.scala
@@ -69,18 +69,37 @@ final class DiagnosticsReporter(reporter: Reporter) extends DiagnosticListener[J
         def startPosition: Option[Long] = checkNoPos(d.getStartPosition)
         def endPosition: Option[Long] = checkNoPos(d.getEndPosition)
         override val offset: Maybe[Integer] = Logger.o2m(checkNoPos(d.getPosition) map { x => new Integer(x.toInt) })
-        // TODO - Is this pulling contents of the line correctly?
-        // Would be ok to just return null if this version of the JDK doesn't support grabbing
-        // source lines?
-        override def lineContent: String =
-          Option(d.getSource) match {
-            case Some(source: JavaFileObject) =>
-              (Option(source.getCharContent(true)), startPosition, endPosition) match {
-                case (Some(cc), Some(start), Some(end)) => cc.subSequence(start.toInt, end.toInt).toString
-                case _                                  => ""
+        override def lineContent: String = {
+          def getDiagnosticLine: Option[String] =
+            try {
+              // See com.sun.tools.javac.api.ClientCodeWrapper.DiagnosticSourceUnwrapper
+              val diagnostic = d.getClass.getField("d").get(d)
+              // See com.sun.tools.javac.util.JCDiagnostic#getDiagnosticSource
+              val getDiagnosticSourceMethod = diagnostic.getClass.getDeclaredMethod("getDiagnosticSource")
+              Option(getDiagnosticSourceMethod.invoke(diagnostic)) match {
+                case Some(diagnosticSource) =>
+                  // See com.sun.tools.javac.util.DiagnosticSource
+                  val getLineMethod = diagnosticSource.getClass.getMethod("getLine", Integer.TYPE)
+                  Option(getLineMethod.invoke(diagnosticSource, line.get())).map(_.toString)
+                case _ => None
               }
-            case _ => ""
-          }
+            } catch {
+              case ignored: NoSuchMethodException => None
+              case ignored: NoSuchFieldException  => None
+            }
+
+          def getExpression: String =
+            Option(d.getSource) match {
+              case Some(source: JavaFileObject) =>
+                (Option(source.getCharContent(true)), startPosition, endPosition) match {
+                  case (Some(cc), Some(start), Some(end)) => cc.subSequence(start.toInt, end.toInt).toString
+                  case _                                  => ""
+                }
+              case _ => ""
+            }
+
+          getDiagnosticLine.getOrElse(getExpression)
+        }
         private val sourceUri = fixSource(d.getSource)
         override val sourcePath = Logger.o2m(sourceUri)
         override val sourceFile = Logger.o2m(sourceUri.map(new File(_)))
--- a/compile/src/main/scala/sbt/compiler/javac/DiagnosticsReporter.scala
+++ b/compile/src/main/scala/sbt/compiler/javac/DiagnosticsReporter.scala
@@ -84,6 +84,7 @@ final class DiagnosticsReporter(reporter: Reporter) extends DiagnosticListener[J
                 case _ => None
               }
             } catch {
+              // TODO - catch ReflectiveOperationException once sbt is migrated to JDK7
               case ignored: Throwable => None
             }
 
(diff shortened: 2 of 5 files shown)


lightbend-cla-validator:
Hi @fkorotkov,

typesafe-tools:

fkorotkov:
@typesafehub-validator signed.

compile/src/test/scala/sbt/compiler/javac/JavaCompilerSpec.scala
```diff
@@ -64,20 +65,28 @@ object JavaCompilerSpec extends Specification {
...
 
-  def lineMatches(p: Problem, lineno: Int): Boolean =
-    p.position.line.isDefined && (p.position.line.get == lineno)
+  def lineMatches(p: Problem, lineno: Int, lineContent: Option[String] = None): Boolean = {
+    def lineContentCheck =
+      lineContent match {
+        case Some(content) => content.equalsIgnoreCase(p.position.lineContent())
+        case _             => true
+      }
+    def lineNumberCheck = p.position.line.isDefined && (p.position.line.get == lineno)
```
dwijnand, line 117: Option.get should be avoided:

```
def lineNumberCheck = p.position.line.exists(_ == lineno)
```

notes/0.13.9/proper-line-content-from-local-javac.markdown
```diff
@@ -0,0 +1,10 @@
+
+  [@fkorotkov]: http://github.com/fkorotkov
+
+### Fixes with compatibility implications
+
+### Improvements
+
+Use diagnostic classes to get lines contents in local java compiler.
```
dwijnand: This should be in `notes/0.13.10/` and it should reference pull request 2108

compile/src/main/scala/sbt/compiler/javac/DiagnosticsReporter.scala
```diff
@@ -69,18 +69,37 @@ final class DiagnosticsReporter(reporter: Reporter) extends DiagnosticListener[J
...
+                  // See com.sun.tools.javac.util.DiagnosticSource
+                  val getLineMethod = diagnosticSource.getClass.getMethod("getLine", Integer.TYPE)
+                  Option(getLineMethod.invoke(diagnosticSource, line.get())).map(_.toString)
+                case _ => None
               }
-            case _ => ""
-          }
+            } catch {
+              case ignored: NoSuchMethodException => None
+              case ignored: NoSuchFieldException  => None
```
dwijnand: There's lots of exception that can be thrown via reflection, as well as your `Maybe#get` usage. Perhaps we should just catch any non fatal exception and fallback to `getExpression`.

build.sbt
```diff
@@ -570,7 +570,7 @@ lazy val safeUnitTests = taskKey[Unit]("Known working tests (for both 2.10 and 2
 lazy val safeProjects: ScopeFilter = ScopeFilter(
   inProjects(mainSettingsProj, mainProj, ivyProj, completeProj,
     actionsProj, classpathProj, collectionProj, compileIncrementalProj,
-    logProj, runProj, stdTaskProj),
+    logProj, runProj, stdTaskProj, compilerProj),
```
dwijnand: why?

dwijnand:
Thanks for the pull request. Looks like it won't merge cleanly, so it needs a rebase. I've also added some comments.

compile/src/main/scala/sbt/compiler/javac/DiagnosticsReporter.scala
```diff
@@ -69,18 +69,37 @@ final class DiagnosticsReporter(reporter: Reporter) extends DiagnosticListener[J
...
-          Option(d.getSource) match {
-            case Some(source: JavaFileObject) =>
-              (Option(source.getCharContent(true)), startPosition, endPosition) match {
-                case (Some(cc), Some(start), Some(end)) => cc.subSequence(start.toInt, end.toInt).toString
-                case _                                  => ""
+        override def lineContent: String = {
+          def getDiagnosticLine: Option[String] =
+            try {
+              // See com.sun.tools.javac.api.ClientCodeWrapper.DiagnosticSourceUnwrapper
+              val diagnostic = d.getClass.getField("d").get(d)
```
jsuereth, line 76: I'm really nervous casting directly to com.sun.javac.api.ClientCodeWrapper.DiagnosticSourceUnwrapper internals.

IF we do this, I think the alternative mechanism (that was here before) should be a fallback in the event these classes do not exist.   (i.e. alternative JDKs).

compile/src/main/scala/sbt/compiler/javac/DiagnosticsReporter.scala
```diff
@@ -69,18 +69,37 @@ final class DiagnosticsReporter(reporter: Reporter) extends DiagnosticListener[J
...
+                case Some(diagnosticSource) =>
+                  // See com.sun.tools.javac.util.DiagnosticSource
+                  val getLineMethod = diagnosticSource.getClass.getMethod("getLine", Integer.TYPE)
+                  Option(getLineMethod.invoke(diagnosticSource, line.get())).map(_.toString)
+                case _ => None
               }
-            case _ => ""
-          }
+            } catch {
+              case ignored: NoSuchMethodException => None
```
jsuereth: Very glad to see this caught and returning None.  I think this whole "getDiagnosticLine" mechanism should probably be its own stand-alone static utiltiy method.

compile/src/main/scala/sbt/compiler/javac/DiagnosticsReporter.scala
jsuereth: Yeah, like InvocationTargetException.  You could just catch ReflectionException and that should be good.  I don't htink we want to catch any non fatal exception though, those error shoudl bubble up.

compile/src/test/scala/sbt/compiler/javac/JavaCompilerSpec.scala
jsuereth, line 117: +1

jsuereth:
Thanks for the PR.  A few things to clean up, but having the option to go right after the JDK classes will be nice. As I said, I'd like to also use the previous mechanism as a fallback before finally not having any diagnostic info.

fkorotkov:
Thank you guys for the feedback. Addressed in https://github.com/fkorotkov/sbt/commit/c73f513016a8c2be94decfda8ea83b6e1b95601d.

@jsuereth I do fall back to the previous behavior. See https://github.com/sbt/sbt/pull/2108/files#diff-9b267fe75b6831da3dca25ceb91082e9R90

compile/src/test/scala/sbt/compiler/javac/JavaCompilerSpec.scala
fkorotkov, line 117: Unfortunately it is xsbti.Maybe and there is no exists method

jsuereth, line 117: AH, right.  Maybe is the limited interface for java-interop/classloader boundary fun.   Sorry!

jsuereth:
@fkorotkov Yeah, the previous behavior was that (casting to JavaObject).  I know that worked for SOME files.  We know your reflection will work for Oracle/OpenJDK (I think).   When it comes to these kinds of behaviors that aren't fully specified, we have to "hope for the best" and leave hooks for people who care about the non-standard JDKs to patch in.   If we have `JavaObject` as an esacpe hook, I hope this means alternative JDKs will at least have a chance to work, vs. casting.

compile/src/main/scala/sbt/compiler/javac/DiagnosticsReporter.scala
```diff
@@ -69,18 +69,37 @@ final class DiagnosticsReporter(reporter: Reporter) extends DiagnosticListener[J
...
+            Option(d.getSource) match {
+              case Some(source: JavaFileObject) =>
+                (Option(source.getCharContent(true)), startPosition, endPosition) match {
+                  case (Some(cc), Some(start), Some(end)) => cc.subSequence(start.toInt, end.toInt).toString
+                  case _                                  => ""
+                }
+              case _ => ""
+            }
+
+          getDiagnosticLine.getOrElse(getExpression)
```
jsuereth, line 101: Looks good!

jsuereth:
Thanks much @fkorotkov !   Really glad to have better tooling support on the java side of the house here!

fkorotkov:
Thank you guys!

I initially though to use source.getCharContent and split it in lines. But then I check how javac works and went that way. But now as you mentioned other JDK I think it might be a better solution. Anyway source.getCharContent is already used in the default behavior.

PR #2114, opened 2015-07-17
Add an Append.Sequence instance for List

### LINKED ISSUE #2107: Cannot use += and ++= to Keys.developers
Type of Keys.developers is `SettingKey[List[Developer]]`. And there is no `Append.Sequence` instance for List.
This seems to needs to change the type to Seq or add the instance. The former is simple fix, but maybe breaks compatibility.

pdalpra:
Fixes #2107

### DIFF
--- a/main/settings/src/test/scala/UsageTest.scala
+++ b/main/settings/src/test/scala/UsageTest.scala
@@ -13,7 +13,7 @@ import complete.{ DefaultParsers, Parsers }
 	val x = task { set.value }
 	val y = task { true }
 	val z = task { if(y.value) x.value else plain.value }
-	val a = taskDyn { 
+	val a = taskDyn {
 		if(y.value) z else x
 	}
 }*/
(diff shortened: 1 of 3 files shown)


pdalpra:
Is there a place where I could add tests for this ?

typesafe-tools:

eed3si9n:
https://github.com/sbt/sbt/blob/0.13/main/settings/src/test/scala/UsageTest.scala looks like a place.

jsuereth:
With tests, this would LGTM

eed3si9n:
And notes.

pdalpra:
Amended my commit to add notes and tests.

main/settings/src/test/scala/UsageTest.scala
```diff
@@ -100,4 +102,14 @@ object Assign {
...
+
+  // Test that Append.Sequence instances for Seq/List work and don't mess up with each other
+  seqSetting := Seq("test1")
+  seqSetting ++= Seq("test2")
+  seqSetting ++= List("test3")
+  seqSetting += "test4"
+
+  seqSetting := List("test1")
+  seqSetting ++= List("test2")
+  seqSetting += "test4"
```
dwijnand: I think these 3 should be `listSetting`.

pdalpra: Woops.

pdalpra: fixed

dwijnand:
LGTM, unless we want to also test that the view bound variant is also working.

jsuereth:
LGTM

PR #2115, opened 2015-07-18
Fixes #2105/#1763. Cached resolution: removes duplicate callers during merge

### LINKED ISSUE #2105: cached resolution: remove transitive dependencies from callers array when using consolidated resolution
@eed3si9n - I hope I'm describing the issue correctly - let me know..

When using `withConsolidatedResolution(true)` can the graph.json size be further reduced by removing transitive dependencies from the `callers` array? For small projects it's manageable, but for a large project the json files mushroom in size and the performance of consolidated resolution deteriorates.

Steps: 

``` sh
    git@github.com:edeustace/sbt-callers-sample.git
    cd sbt-callers-sample/lib-two
    sbt publish-local
    cd ../lib-one
    sbt publish-local
    cd ..
    sbt update
```

Then inspect the graph.json for module `a`. You should see something like: 

``` javascript
     [
      {
        "caller": {
          "organization": "org.scala-sbt.temp",
          "name": "temp-resolve-1fd7120a8dac7d10ffc45ff63fc42f19d0d148e1",
          "revision": "1.0",
          //...
        },
        //...
      },
      {
        "caller": {
          "organization": "com.ee",
          "name": "lib-one_2.10",
          "revision": "0.0.1-SNAPSHOT",
          //...
        },
        //...
      },
      {
        "caller": {
          "organization": "com.ee",
          "name": "lib-two_2.10",
          "revision": "0.0.1-SNAPSHOT",
          //...
        },
      }
      //...
    ]
```

The tmp org (`org.scala-sbt.temp`) is fine as is the direct dependency `lib-one`.

The issue is that `lib-two` is listed as a caller even though it's a transitive dependency of `a`.

eed3si9n:
Fixes #2105/#1763

In #2097 which became 0.13.9-RC2 I ended up regressing on OOM (#1763), which I thought I tested locally, but apparently I tested on an older SNAPSHOT.
I've gotten to the root cause of the bloating caller info this time, I think. It's my merge code that was `flatMap`ing all callers across the subprojects. So the more subproject you add the more it was duplicating the callers.

/review @jsuereth, @dwijnand

### DIFF
--- a/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
+++ b/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
@@ -27,7 +27,7 @@ private[sbt] object CachedResolutionResolveCache {
   def createID(organization: String, name: String, revision: String) =
     ModuleRevisionId.newInstance(organization, name, revision)
   def sbtOrgTemp = JsonUtil.sbtOrgTemp
-  def graphVersion = "0.13.9B"
+  def graphVersion = "0.13.9C"
   val buildStartup: Long = System.currentTimeMillis
   lazy val todayStr: String = toYyyymmdd(buildStartup)
   lazy val tomorrowStr: String = toYyyymmdd(buildStartup + (1 day).toMillis)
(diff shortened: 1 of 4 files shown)


jsuereth:
LGTM

Issue #2109, opened 2015-07-15
maven resolver plugin pulls excluded transitive dependencies


chasebradford:
chase:lib1:1.0-SNAPSHOT

```
<dependencies>
</dependencies>
```

chase:lib2:1.0-SNAPSHOT:

```
<dependencies>
  <dependency>
    <groupId>chase</groupId>
    <artifactId>lib1</artifactId>
    <version>1.0-SNAPSHOT</version>
  </dependency>
</dependencies>
```

top:

```
<dependencies>
  <dependency>
    <groupId>chase</groupId>
    <artifactId>lib2</artifactId>
    <version>1.0-SNAPSHOT</version>
    <exclusions>
      <exclusion>
        <groupId>chase</groupId>
        <artifactId>lib1</artifactId>
      </exclusion>
    </exclusions>
  </dependency>
</dependencies>
```

My test SBT project depends on top.

```
// build.sbt
resolvers += "Local Maven Repository" at "file://" + Path.userHome.absolutePath + "/.m2/repository"

libraryDependencies += "chase" % "top" % "1.0-SNAPSHOT"
```

lib1 still resolves when I run `sbt update`, even though it's a transitive dependency and should be excluded.  I hacked the SbtArtifactDescriptorReader in the sbt-maven-resolver to `System.err.println(dependency.getArtifactId() + " EXCLUDES " + exclusion.getArtifactId());` and I can see that the exclusion rule is detected while processing top's dependencies.  For some reason it still shows up in the full dependency tree:

```
#[Jul 15 14:44:30][~/TestProject]
> sbt dependencyTree
[info] Loading project definition from /Users/cbradford/TestProject/project
[info] Set current project to testproject (in build file:/Users/cbradford/TestProject/)
[info] Updating {file:/Users/cbradford/TestProject/}testproject...
[info] Resolving chase#top;1.0-SNAPSHOT ...
lib2 EXCLUDES lib1
[info] Resolving chase#lib2;1.0-SNAPSHOT ...
[info] Resolving chase#lib1;1.0-SNAPSHOT ...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
[info] default:testproject_2.10:0.1-SNAPSHOT [S]
[info]   +-chase:top:1.0-SNAPSHOT
[info]     +-chase:lib2:1.0-SNAPSHOT
[info]       +-chase:lib1:1.0-SNAPSHOT
```

chasebradford:
[...]

PR #2126, opened 2015-07-23
Fix/2109 transitive excludes

jsuereth:
Opening the PR to show the issue in Ivy itself and we need to figure out what to do with the test.

Right now, this brings maven-plugin-resolver related behavior in-line with Ivy's default for transitive excludes.  Unfortunately, I think the default behavior is wrong (see attached test, and thanks to @chasebradford for the test case).

We could mark the test as pending, leave #2109 open, and merge this while we search for a better fix, or we can continue working on #2109 during this PR.

The fix to maven-resolver-plugin _is_ done though.  It just discovered a bigger issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,8 +34,8 @@ env:
     - SCRIPTED_TEST="scripted tests/*"
     - SCRIPTED_TEST="scripted project-load/*"
     - SCRIPTED_TEST="checkBuildScala211"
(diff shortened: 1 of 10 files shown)


jsuereth:
Review/help from @eed3si9n

jsuereth:
@eed3si9n this one is ready for a look now.

eed3si9n:
LGTM

PR #2128, opened 2015-07-24
Remove reference to dead servers.

jsuereth:
Don't really think this deserves notes.

Review by @eed3si9n


eed3si9n:
LGTM pending Travis.

PR #2129, opened 2015-07-26
[0.13.9-RC3] cached resolution: stack overflow when circular dependency is found

eed3si9n:
## steps
1. use cached resolution
2. have sizable graph (for instance by using spark)
3. have a circular dependency
## problem

```
java.lang.StackOverflowError
    at scala.collection.MapLike$DefaultKeySet.contains(MapLike.scala:169)
    at scala.collection.GenSetLike$class.apply(GenSetLike.scala:43)
    at scala.collection.AbstractSet.apply(Set.scala:47)
    at sbt.ivyint.CachedResolutionResolveEngine$$anonfun$32$$anonfun$apply$15$$anonfun$apply$16$$anonfun$apply$17.apply(CachedResolutionResolveEngine.scala:420)
    at sbt.ivyint.CachedResolutionResolveEngine$$anonfun$32$$anonfun$apply$15$$anonfun$apply$16$$anonfun$apply$17.apply(CachedResolutionResolveEngine.scala:418)
    at scala.collection.LinearSeqOptimized$class.exists(LinearSeqOptimized.scala:80)
    at scala.collection.immutable.List.exists(List.scala:84)
    at sbt.ivyint.CachedResolutionResolveEngine$$anonfun$32$$anonfun$apply$15$$anonfun$apply$16.apply(CachedResolutionResolveEngine.scala:418)
    at sbt.ivyint.CachedResolutionResolveEngine$$anonfun$32$$anonfun$apply$15$$anonfun$apply$16.apply(CachedResolutionResolveEngine.scala:418)
    at scala.collection.Iterator$class.exists(Iterator.scala:753)
    at scala.collection.AbstractIterator.exists(Iterator.scala:1157)
    at scala.collection.IterableLike$class.exists(IterableLike.scala:77)
    at scala.collection.AbstractIterable.exists(Iterable.scala:54)
    at sbt.ivyint.CachedResolutionResolveEngine$$anonfun$32$$anonfun$apply$15.apply(CachedResolutionResolveEngine.scala:417)
    at sbt.ivyint.CachedResolutionResolveEngine$$anonfun$32$$anonfun$apply$15.apply(CachedResolutionResolveEngine.scala:417)
    at scala.collection.Iterator$class.exists(Iterator.scala:753)
    at scala.collection.AbstractIterator.exists(Iterator.scala:1157)
    at scala.collection.IterableLike$class.exists(IterableLike.scala:77)
    at scala.collection.AbstractIterable.exists(Iterable.scala:54)
    at sbt.ivyint.CachedResolutionResolveEngine$$anonfun$32.apply(CachedResolutionResolveEngine.scala:416)
    at sbt.ivyint.CachedResolutionResolveEngine$$anonfun$32.apply(CachedResolutionResolveEngine.scala:414)
    at scala.collection.TraversableLike$$anonfun$partition$1.apply(TraversableLike.scala:321)
    at scala.collection.TraversableLike$$anonfun$partition$1.apply(TraversableLike.scala:321)
    at scala.collection.Iterator$class.foreach(Iterator.scala:727)
    at scala.collection.AbstractIterator.foreach(Iterator.scala:1157)
    at scala.collection.IterableLike$class.foreach(IterableLike.scala:72)
    at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
    at scala.collection.TraversableLike$class.partition(TraversableLike.scala:321)
    at scala.collection.AbstractTraversable.partition(Traversable.scala:105)
    at sbt.ivyint.CachedResolutionResolveEngine$class.sortModules$1(CachedResolutionResolveEngine.scala:414)
    at sbt.ivyint.CachedResolutionResolveEngine$class.sortModules$1(CachedResolutionResolveEngine.scala:427)
    at sbt.ivyint.CachedResolutionResolveEngine$class.sortModules$1(CachedResolutionResolveEngine.scala:427)
    at sbt.ivyint.CachedResolutionResolveEngine$class.sortModules$1(CachedResolutionResolveEngine.scala:427)
    at sbt.ivyint.CachedResolutionResolveEngine$class.sortModules$1(CachedResolutionResolveEngine.scala:427)
    at sbt.ivyint.CachedResolutionResolveEngine$class.sortModules$1(CachedResolutionResolveEngine.scala:427)
```
## what this PR changes
1. First `sortModules` is now tail recursive so blowing up the stack should not occur.
2. Second, it short circuits when the algorithm stops making progress (`notCalled.isEmpty`), so the total number of iteration should be much fewer.

/review @jsuereth, @dwijnand

### DIFF
--- a/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
+++ b/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
@@ -407,9 +407,11 @@ private[sbt] trait CachedResolutionResolveEngine extends ResolveEngine {
         }: _*)
       val stackGuard = reports0.size * reports0.size * 2
       // sort the all modules such that less called modules comes earlier
-      def sortModules(cs: ListMap[(String, String), Vector[OrganizationArtifactReport]],
+      @tailrec def sortModules(cs: ListMap[(String, String), Vector[OrganizationArtifactReport]],
+        acc: ListMap[(String, String), Vector[OrganizationArtifactReport]],
         n: Int): ListMap[(String, String), Vector[OrganizationArtifactReport]] =
         {
+          println(s"sortModules: $n / $stackGuard")
           val keys = cs.keySet
           val (called, notCalled) = cs partition {
             case (k, oas) =>
@@ -422,9 +424,8 @@ private[sbt] trait CachedResolutionResolveEngine extends ResolveEngine {
                 }
               }
           }
-          notCalled ++
-            (if (called.isEmpty || n > stackGuard) called
-            else sortModules(called, n + 1))
+          (if (called.isEmpty || n > stackGuard) acc ++ notCalled ++ called
+          else sortModules(called, acc ++ notCalled, n + 1))
         }
       def resolveConflicts(cs: List[((String, String), Vector[OrganizationArtifactReport])]): List[OrganizationArtifactReport] =
         cs match {
@@ -446,7 +447,7 @@ private[sbt] trait CachedResolutionResolveEngine extends ResolveEngine {
                 x :: resolveConflicts(next)
             })
         }
-      val sorted = sortModules(allModules, 0)
+      val sorted = sortModules(allModules, ListMap(), 0)
       val result = resolveConflicts(sorted.toList)
       result.toVector
     }
--- a/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
+++ b/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
@@ -396,59 +396,68 @@ private[sbt] trait CachedResolutionResolveEngine extends ResolveEngine {
   def mergeOrganizationArtifactReports(rootModuleConf: String, reports0: Vector[OrganizationArtifactReport], os: Vector[IvyOverride], log: Logger): Vector[OrganizationArtifactReport] =
     {
       // group by takes up too much memory. trading space with time.
-      val orgNamePairs = (reports0 map { oar => (oar.organization, oar.name) }).distinct
+      val orgNamePairs: Vector[(String, String)] = (reports0 map { oar => (oar.organization, oar.name) }).distinct
       // this might take up some memory, but it's limited to a single
       val reports1 = reports0 map { filterOutCallers }
-      val allModules: ListMap[(String, String), Vector[OrganizationArtifactReport]] =
-        ListMap(orgNamePairs map {
+      val allModules0: Map[(String, String), Vector[OrganizationArtifactReport]] =
+        Map(orgNamePairs map {
           case (organization, name) =>
             val xs = reports1 filter { oar => oar.organization == organization && oar.name == name }
             ((organization, name), xs)
         }: _*)
-      val stackGuard = reports0.size * reports0.size * 2
       // sort the all modules such that less called modules comes earlier
-      @tailrec def sortModules(cs: ListMap[(String, String), Vector[OrganizationArtifactReport]],
-        acc: ListMap[(String, String), Vector[OrganizationArtifactReport]],
-        n: Int): ListMap[(String, String), Vector[OrganizationArtifactReport]] =
+      @tailrec def sortModules(cs: Vector[(String, String)],
+        acc: Vector[(String, String)],
+        n: Int, guard: Int): Vector[(String, String)] =
         {
-          println(s"sortModules: $n / $stackGuard")
-          val keys = cs.keySet
-          val (called, notCalled) = cs partition {
-            case (k, oas) =>
-              oas exists {
-                _.modules.exists {
-                  _.callers exists { caller =>
-                    val m = caller.caller
-                    keys((m.organization, m.name))
-                  }
+          // println(s"sortModules: $n / $guard")
+          val keys = cs.toSet
+          val (called, notCalled) = cs partition { k =>
+            val reports = allModules0(k)
+            reports exists {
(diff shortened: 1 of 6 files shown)


dwijnand:
LGTM. Does this need notes?

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -396,58 +396,68 @@ private[sbt] trait CachedResolutionResolveEngine extends ResolveEngine {
...
+        Map(orgNamePairs map {
           case (organization, name) =>
             val xs = reports1 filter { oar => oar.organization == organization && oar.name == name }
             ((organization, name), xs)
         }: _*)
-      val stackGuard = reports0.size * reports0.size * 2
       // sort the all modules such that less called modules comes earlier
-      def sortModules(cs: ListMap[(String, String), Vector[OrganizationArtifactReport]],
-        n: Int): ListMap[(String, String), Vector[OrganizationArtifactReport]] =
+      @tailrec def sortModules(cs: Vector[(String, String)],
```
jsuereth, line 409: Is this the key thing that was having the Stack overflow?

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -396,58 +396,68 @@ private[sbt] trait CachedResolutionResolveEngine extends ResolveEngine {
...
                 }
               }
+            }
           }
-          notCalled ++
-            (if (called.isEmpty || n > stackGuard) called
-            else sortModules(called, n + 1))
+          lazy val result0 = acc ++ notCalled ++ called
+          (if (n > guard) {
+            log.warn(s"""cached resolution detected circular dependencies: ${cs.mkString(",")}""")
```
jsuereth: hmm, that's... fun...

sbt/src/sbt-test/dependency-management/cached-resolution-circular/changes/multi.sbt
```diff
@@ -0,0 +1,57 @@
...
+      "com.linkedin.pegasus" % "data-avro" % "1.9.40",
+      "org.jboss.netty" % "netty" % "3.2.0.Final"
+    )
+  )
+
+lazy val b = project.
+  settings(commonSettings: _*).
+  settings(
+    name := "b",
+    // this adds circular dependency
```
jsuereth, line 35: I'm actually pretty certain that these should be ERRORS, not something we workaround allow.   What is the expected behavior for A to depend on B and B to depend on A?  Why are they jsut not hte same project if they have that level fo coupling?

sbt/src/sbt-test/dependency-management/circular-dependency/changes/multi.sbt
```diff
@@ -4,6 +4,7 @@ def commonSettings: Seq[Def.Setting[_]] =
   Seq(
     ivyPaths := new IvyPaths( (baseDirectory in ThisBuild).value, Some((target in LocalRootProject).value / "ivy-cache")),
     scalaVersion := "2.10.4",
+    fullResolvers := fullResolvers.value.filterNot(_.name == "inter-project"),
```
jsuereth, line 7: You need this because sbt itself will cause problems with circular dependencies right?

jsuereth:
SO, in the presence of circular dependencies we should not StackOverflow.  However, I think we need to issue a warning that you really shouldn't have them, rather than just "detected".

The fix looks fine, it's the premise of circular dependency support I question.

eed3si9n:
@dwijnand I don't think notes are needed beyond what's in there already for cached resolution. It's essentially an improvement to #2097.

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
eed3si9n, line 409: Sort of. There was a guard before called `n` to prevent infinite loop but `size * size * 2` was too generous. This implementation will dynamically shrink the guard at each step so in general it should iterate much fewer number (like `100 * 100 * 2` vs 10-ish).

sbt/src/sbt-test/dependency-management/cached-resolution-circular/changes/multi.sbt
eed3si9n, line 35: I am trying to emulate circular dependency that happens in the wild. Ivy will print warning already, and there's also a way to make it a error too (via #1601). But sometimes it happens transitively or as a result of evictions. Given enough number of dependencies on the graph, I am guessing it becomes inevitable.

sbt/src/sbt-test/dependency-management/circular-dependency/changes/multi.sbt
eed3si9n, line 7: sbt won't have problem with circular dependency of library dependencies. I am just making sure that it uses the JAR that's locally published.

dwijnand:
I reckoned as much. Perhaps a reference to the PR is customary.

eed3si9n:
I pushed out 0.13.9-M2 so I can test this prior to a public release.

sbt/src/sbt-test/dependency-management/cached-resolution-circular/changes/multi.sbt
jsuereth, line 35: I guess it's surprising to me that it would be inevitable to have a circle.  I think that hits at some other problem, but if we're handling it now already, it'd be good to fix it.

jsuereth:
Cool, LGTM

PR #2136, opened 2015-07-30
Keep roots when cleaning

pdalpra:


typesafe-tools:

notes/0.13.10/clean-leave-roots.markdown
```diff
@@ -0,0 +1,11 @@
+[@pdalpra]: http://github.com/pdalpra
+[106]: https://github.com/sbt/sbt/issues/106
+[2133]: https://github.com/sbt/sbt/pull/2133
+
+### Fixes with compatibility implications
+
+### Improvements
+
+- clean task does not delete root folders anymore, only contents [#106][106]/[#2133][2133] by [@pdalpra][pdalpra]
```
dwijnand: 2136

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -716,10 +716,15 @@ object Defaults extends BuildCommon {
 
   def doClean(clean: Seq[File], preserve: Seq[File]): Unit =
     IO.withTemporaryDirectory { temp =>
+      def cleanLeaveRoot(root: File) = {
+        val (dirs, files) = (root.*** --- root).get.partition(_.isDirectory)
+        IO.delete(files)
+        IO.delete(dirs)
```
dwijnand: why partition then delete both?

pdalpra: For an unknown reason, I thought that trying to delete an already deleted file could end up badly, but it won't.

pdalpra:
Commit amended

notes/0.13.10/clean-leave-roots.markdown
```diff
@@ -0,0 +1,11 @@
+[@pdalpra]: http://github.com/pdalpra
+[106]: https://github.com/sbt/sbt/issues/106
+[2136]: https://github.com/sbt/sbt/pull/2136
+
+### Fixes with compatibility implications
+
+### Improvements
+
+- clean task does not delete root folders anymore, only contents [#106][106]/[#2136][2136] by [@pdalpra][pdalpra]
```
jsuereth, line 9: There may be even more ticket saround this.

ALSO, did you run a performance test at all to see what the impact is?  We may already be screwed for clean on windows, but I'm curious if this is noticeable on other platforms.

pdalpra, line 9: > There may be even more tickets around this.

I'll look for other tickets.

> ALSO, did you run a performance test at all to see what the impact is? We may already be screwed for clean on windows, but I'm curious if this is noticeable on other platforms.

The very basic tests I ran showed no noticeable difference, but I'll test on 'big' projects to see its impacts in the large.

pdalpra, line 9: Run some tests (running `test:compile` then `clean` on sbt itself) and it seems it has no noticeable impact on the performance of `clean`.
Also, I did not notice other tickets linked to that issue.

jsuereth, line 9: Which platform, just curious.

pdalpra, line 9: Mac OS X (I don't have any easy access to Windows, unfortunately).

jsuereth:
Note to self: This fix looks good, just need to check windows performance when I have my windows machine tonight.

fommil:
@jsuereth @eed3si9n @dwijnand may be of interest to you, I created a Windows XP VirtualBox image for ENSIME testing. May be of interest to you? Happy to share.

I'm also keen on creating a Windows 7 one, as that's what I have to use at work.

Of course, there is also AppVeyor which we're using to test our Windows builds of ENSIME. You could probably use it to test this one performance thing by testing the time of a compile / clean for something like akka.

eed3si9n:
@fommil I think it'll be useful.

dwijnand:
@fommil and/or @eed3si9n (and/or @jsuereth) what's a reasonable way to assert this doesn't bomb windows performance?

I want to either merge or close this.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -719,7 +719,7 @@ object Defaults extends BuildCommon {
       val (dirs, files) = preserve.filter(_.exists).flatMap(_.***.get).partition(_.isDirectory)
       val mappings = files.zipWithIndex map { case (f, i) => (f, new File(temp, i.toHexString)) }
       IO.move(mappings)
-      IO.delete(clean)
+      clean.foreach(root => IO.delete((root.*** --- root).get))
```
eed3si9n, line 722: One way to mitigate potential perf regression might be to avoid `***` here and scan only the first level of items using `*`.

eed3si9n (changes requested):

fommil:
Without a windows perf test, is hard to say. One thing that could give an indicator would be to instrument every i/o call and count them. I hope the sbt 1.0 api will be nio compatible because that's the only way to improve things on windows. All the sbt apis that pass around Seq[File] are introducing areas of contention as the more efficient way to perform i/o is to have a scan thread and then a work pool.

jvican:
I think this is the root of a sbt problem I've been experiencing for a long time.
  
I'm happy to get this to the finish line. What do I have to do?

eed3si9n:
@jvican There's a perf concern on Windows - https://github.com/sbt/sbt/pull/2136/files#r88750604

dwijnand:
This will need a refresh. So let's close it for now, and re-review it if someone refreshes it.

PR #2138, opened 2015-08-03
[0.13.9-RC3] Fixes #2129. break up circular dependency loops in cached resolution

eed3si9n:
Fixes #2129

Simple remove-one method to workaround for circular dependency did not
work. This fix traverses the entire graph to detect all loops and then
breaks them up.

To debug what's going on I wrote a one-off script to analyze the looping graph visually based on `show update` output and the warning message produced by #2129, which had 100 candidates for circular dependency, but most of them were leaf nodes hanging off of the other. Among the 100, there were only 2 small loops:
https://gist.github.com/eed3si9n/91f3ceafebce75b59833

/review @jsuereth, @dwijnand

### DIFF
--- a/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
+++ b/ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
@@ -405,6 +405,77 @@ private[sbt] trait CachedResolutionResolveEngine extends ResolveEngine {
             val xs = reports1 filter { oar => oar.organization == organization && oar.name == name }
             ((organization, name), xs)
         }: _*)
+      // this returns a List of Lists of (org, name). should be deterministic
+      def detectLoops(allModules: Map[(String, String), Vector[OrganizationArtifactReport]]): List[List[(String, String)]] =
+        {
+          val loopSets: mutable.Set[Set[(String, String)]] = mutable.Set.empty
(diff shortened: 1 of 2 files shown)


ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -405,6 +405,77 @@ private[sbt] trait CachedResolutionResolveEngine extends ResolveEngine {
...
+                    val oars: Vector[OrganizationArtifactReport] = oars0 map { oar =>
+                      val mrs = oar.modules map { mr =>
+                        val callers0 = mr.callers
+                        val callers = callers0 filterNot { c => (c.caller.organization, c.caller.name) == moduleWithMostCallers }
+                        if (callers.size == callers0.size) mr
+                        else mr.copy(callers = callers)
+                      }
+                      OrganizationArtifactReport(oar.organization, oar.name, mrs)
+                    }
+                    (k, oars)
```
dwijnand, line 473: Perhaps it would be useful to debug log what's being dropped here, for future debugging help.

dwijnand:
LGTM. Nice script.

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
eed3si9n, line 473: Added a few logs.

jsuereth:
This LGTM.

PR #2140, opened 2015-08-03
Set version to 0.13.10-SNAPSHOT.

dwijnand:
Given that work for the 0.13.9 is now on the 0.13.9 branch, 0.13 should be set to 0.13.10-SNAPSHOT.


typesafe-tools:

jsuereth:
Ha! good catch!

PR #2143, opened 2015-08-05
Update scala211 to 2.11.7.

dwijnand:


typesafe-tools:

PR #2146, opened 2015-08-10
cached resolution: don't include callers from evicted modules

eed3si9n:
This is the last of cached resolution fix for 0.13.9 series that I made last week, and released 0.13.9-M4.
Unfortunately the issue that a client was having did not resolve.

I am thinking about releasing 0.13.9 with it as is without going through another RC.

/review @jsuereth, @dwijnand


ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -542,14 +556,6 @@ private[sbt] trait CachedResolutionResolveEngine extends ResolveEngine {
       val result = resolveConflicts(sorted.toList, allModules0)
```
dwijnand, line 556: Unrelated to this PR, but why is this using `allModules0` rather than `allModules2`?

dwijnand:
LGTM, shame to have a mutable Map floating around, but it's limited to the scope of the (rather large) method.

ivy/src/main/scala/sbt/ivyint/CachedResolutionResolveEngine.scala
eed3si9n, line 556: `allModules2` is the list of modules with circular dependencies removed.

PR #2147, opened 2015-08-11
Merge 0.13.9 into 0.13.

dwijnand:


typesafe-tools:

dwijnand:
After a number of restarts, this has finally passed now.

PR #2151, opened 2015-08-12
Extract Def.settings from Project#settings.

dwijnand:
This allows for the same functionality that using `SettingsDefinition` in
`Project#settings` allows (specifying either bare `Setting[_]` or a `Seq[Setting[_]]`)
to be available outside of the settings for a project, for instance when
defining a val.

In short, it allows:

``` scala
val modelSettings = Def.settings(
  sharedSettings,
  libraryDependencies += foo
)
```


typesafe-tools:

eed3si9n:
Notes please.
LGTM pending Travis.

eed3si9n:
Could you rebase this? The Merge button is gone.

dwijnand:
Done.

jsuereth:
Nice cleanups :)

PR #2158, opened 2015-08-15
Fix error when name hashing is enabled and mixed with non-enabled

stuhood:
If `MRelationsNameHashing` is in use, name-hashing is _enabled_. Looks like a copy-paste error from a similar error message in `MRelationsDefaultImpl`.


typesafe-tools:

eed3si9n:
LGTM

dwijnand:
LGTM.

Issue #2155, opened 2015-08-14
Edge case in incremental compile


danchia:
Download https://s3.amazonaws.com/coursera-misc/danchia/broken.tgz
### Steps to reproduce in sample case

```
$ sbt
> project foo
> run
```

in bar/build.sbt
comment out

``` scala
libraryDependencies += "com.typesafe.scala-logging" %% "scala-logging-slf4j" % "2.1.2"
```

and uncomment

``` scala
libraryDependencies += "com.typesafe.scala-logging" %% "scala-logging" % "3.1.0"
```

in bar/src/main/scala/org/example/bar.scala
uncomment out

``` scala
import com.typesafe.scalalogging.Logger
```

and comment

``` scala
import com.typesafe.scalalogging.slf4j.Logger
```

now back in sbt

```
> reload
> run
```
### problem

```
java.lang.AbstractMethodError: org.coursera.Foo.org$coursera$Bar$_setter_$org$coursera$Bar$$logger_$eq(Lcom/typesafe/scalalogging/Logger;)V
at org.coursera.Bar$class.$init$(bar.scala:8)
at org.coursera.Foo.<init>(foo.scala:3)
at org.coursera.Main$.main(foo.scala:10)
at org.coursera.Main.main(foo.scala)
at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.lang.reflect.Method.invoke(Method.java:497)
```

dwijnand:
@danchia Thank you for the step-by-step bug report!

Duhemm:
Actually this bug describes a known issue of the incremental compiler. A pending test case has been introduced in c6b0043336a48c713eec5202fc6ef88006c5275b.

The problem comes from `Bar` being a trait. If `Bar` is an abstract class, then this issue is not observed.
The two classes having the same name (`Logger`) has no effect on this issue.

Duhemm:
Including the private definitions in the computation of the the API hash seems to fix this issue. I tried it in duhemm/sbt@f58a96e7c436f3df9b515a7a6b8815a5f87475ea. I'm afraid always including the private definitions would greatly degrade the performances of the incremental compiler.

[...]

PR #2160, opened 2015-08-17
Include private members in API hash of traits.

Duhemm:
During compilation, scalac generates getters and setters for the fields
of traits, regardless of their access modifiers. Therefore, when a
field of a trait is modified, all its implementors must be recompiled to
take these changes into account.

Private fields of traits were not included in the API hash of traits,
and their implementors were thus not recompiled when modified.

This commit changes the way the API hash is computed for traits only, so
that the generated hash includes the private members of traits.

Fixes sbt/sbt#2155


compile/api/src/main/scala/xsbt/api/HashAPI.scala
```diff
@@ -343,14 +345,14 @@ final class HashAPI(includePrivate: Boolean, includeParamNames: Boolean, include
     hashType(a.baseType)
     hashAnnotations(a.annotations)
   }
-  final def hashStructure(structure: Structure, includeDefinitions: Boolean) =
-    visit(visitedStructures, structure)(structure => hashStructure0(structure, includeDefinitions))
-  def hashStructure0(structure: Structure, includeDefinitions: Boolean): Unit = {
+  final def hashStructure(structure: Structure, includeDefinitions: Boolean, isTrait: Boolean = false) =
+    visit(visitedStructures, structure)(structure => hashStructure0(structure, includeDefinitions, isTrait))
+  def hashStructure0(structure: Structure, includeDefinitions: Boolean, isTrait: Boolean = false): Unit = {
```
Duhemm, line 356: Does that count as a breakage of the binary compatibility?

typesafe-tools:

compile/api/src/main/scala/xsbt/api/HashAPI.scala
eed3si9n, line 356: Let's try the overload, and deprecate the old one.

Duhemm, line 356: @eed3si9n Done

eed3si9n:
LGTM

dwijnand:
Nice!

indrajitr:
:star:

Issue #1750, opened 2014-11-26
updateClassifiers doesn't download sources and docs for updated snapshot dependencies


newffy:
Hello.
sbt.version=0.13.7
we are using corporate maven repository (served by artifactory 3.4.1)
Our libraryDependencies:

``` scala
libraryDependencies ++= {
    val jettyVer = "9.2.3.v20140905"
    val akkaVersion = "2.3.7"
    val kamonVersion = "0.3.5"
    Seq(
        "javax.servlet" % "javax.servlet-api" % "3.1.0" % "provided",
        "com.simplesys" %% "akka-extender" % "1.0.2-SNAPSHOT",
        "com.simplesys" %% "isc-wrapper" % "1.0.0-SNAPSHOT",
        "com.simplesys" %% "common" % "1.0.0-SNAPSHOT",
        "com.simplesys.jdbc.drivers" % "oracle" % "11.2.0.4",
        "com.simplesys" %% "isc-components" % "1.0.0-SNAPSHOT",
        "com.simplesys" %% "core-library" % "1.0.0-SNAPSHOT",
        "com.simplesys" %% "scala-io-extender" % "1.0.0-SNAPSHOT",
        "ru.simplesys" %% "core-domains" % "1.2.1-SNAPSHOT",
        "ru.simplesys" %% "core-utils" % "1.0.1-SNAPSHOT",
        "com.simplesys" % "smartclient-js" % "10.0-v20141114",
        "com.simplesys" %% "scala-gen" % "1.0.0-SNAPSHOT" % "test",
        "com.typesafe.akka" %% "akka-actor" % akkaVersion,
        "com.typesafe.akka" %% "akka-slf4j" % akkaVersion,
        "com.typesafe.akka" %% "akka-persistence-experimental" % akkaVersion exclude("org.iq80.leveldb","leveldb"),
        "org.iq80.leveldb" % "leveldb" % "0.7",
        "com.twitter" %% "chill-akka" % "0.5.0",
        "com.github.bseibel" %% "akka-persistence-bdb" % "1.0.1",
        "com.typesafe.akka" %% "akka-testkit" % akkaVersion % "test",
        "org.eclipse.jetty" % "jetty-webapp" % jettyVer % "container",
        "org.eclipse.jetty" % "jetty-annotations" % jettyVer % "container",
        "org.eclipse.jetty" % "jetty-plus" % jettyVer % "container",
        "org.scalatest" %% "scalatest" % "2.2.2" % "test",
[...]

PR #2169, opened 2015-08-21
Publish compiler interface to an alternate local repository for Scripted tests

### LINKED ISSUE #2165: Issue with compiler-interface in scripted tests
I was creating a scripted test, one that involves compiling some Scala, and I hit this:

``` scala
sbt.InvalidComponent: Couldn't retrieve default sources: module 'compiler-interface'
```

It's probably related to the change in https://github.com/sbt/sbt/pull/2106.

The workaround, for now, is to add (as was done in https://github.com/sbt/sbt/pull/2106 for `dependency-management/mvn-local`):

``` scala
resolvers += Resolver.file("old-local", file(sys.props("user.home") + "/.ivy2/local"))(Resolver.ivyStylePatterns)
```

Duhemm:
Some scripted tests override the default local repository, which
produces errors with the compiler because sbt cannot find the sources
for the compiler interface.

This fix proposes to publish the compiler interface sources to an
alternate local repository before running the scripted tests. This
alternate repository is added to the scripted sbt's configuration, so
that sbt is finally able to find the compiler interface sources.

Fixes #2165

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -43,7 +43,7 @@ def testedBaseSettings: Seq[Setting[_]] =
 
 
 val altLocalRepoName = "alternative-local"
-val altLocalRepoPath = sys.props("user.home" + "/.ivy2/sbt-alternative")
+val altLocalRepoPath = sys.props("user.home") + "/.ivy2/sbt-alternative"
 lazy val altLocalResolver = Resolver.file(altLocalRepoName, file(sys.props("user.home") + "/.ivy2/sbt-alternative"))(Resolver.ivyStylePatterns)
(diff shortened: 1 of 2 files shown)


typesafe-tools:

dwijnand:
In terms of this fixing the issue, I can confirm it worked for me in my #1750 scripted test.

dwijnand:
What do we want to do with this? /cc @eed3si9n @Duhemm

eed3si9n:
@Duhemm Could you forward port this too?

eed3si9n:
Has this been forward ported?

dwijnand:
Yes https://github.com/sbt/zinc/pull/92

PR #2163, opened 2015-08-17
Fix SNAPSHOT srcs/docs with Ivy upgrade.

dwijnand:
Fixes #1750.

---

Updates to the fixes in sbt/ivy#17.

I realise this is missing some due diligence regression suite tests and notes,
but I wanted to push this out for feedback.

PR #2172, opened 2015-08-23
Add localIfFile to MavenRepository, & use it.

dwijnand:


dwijnand:
While #2163 [needs more work](https://github.com/sbt/sbt/pull/2163#issuecomment-133852046), this is an sbt change that I needed and is useful.

typesafe-tools:

eed3si9n:
LGTM

Issue #813, opened 2013-07-12
Running tagged tests works in 0.12.4 but not with 0.13.0-RC1


ghost:
Steps to reproduce the problem:
1. $ git clone https://github.com/oscarvarto/scalatest-tag.git
2. $ cd scalatest-tag
3. $ sbt
    >  xtests:test

sbt says 
    [info] No tests to run for xtests:test
(See https://github.com/oscarvarto/scalatest-tag/blob/master/project/build.properties#L2 --> 0.13.0-RC1 does not run tests tagged XTest)

(I am using the sbt launcher for 0.13.0-RC1 to be able to do this with 0.12 and 0.13 versions of sbt)

Now comment https://github.com/oscarvarto/scalatest-tag/blob/master/project/build.properties#L2 and uncomment https://github.com/oscarvarto/scalatest-tag/blob/master/project/build.properties#L1 so that sbt.version=0.12.4, and try again. The expected behavior is that of sbt 0.12 (being able to run only those tests tagged XTest, with xtests:test): 
$ sbt
[info] Loading project definition from /Users/oscarvarto/programacion/scala2013/scalatest-tag/project
[info] Set current project to scalatest-tag (in build file:/Users/oscarvarto/programacion/scala2013/scalatest-tag/)
    > test
[info] Updating {file:/Users/oscarvarto/programacion/scala2013/scalatest-tag/}scalatest-tag...
[info] Resolving org.scalatest#scalatest_2.10;2.0.M6-SNAP26 ...
[info] Done updating.
[info] DummyTest:
[info] - This test is tagged XTest (pending)
[info] - This is a normal test
[info] Passed: : Total 2, Failed 0, Errors 0, Passed 1, Skipped 1
[success] Total time: 1 s, completed Jul 11, 2013 8:57:55 PM
    > xtests:test
[info] DummyTest:
[info] - This test is tagged XTest (pending)
[info] Passed: : Total 1, Failed 0, Errors 0, Passed 0, Skipped 1
[success] Total time: 0 s, completed Jul 11, 2013 8:57:59 PM

mharrah on #sbt irc channel said:
[7:52pm] mharrah: oscarvarto: ok, the first problem is that the tests aren't detected due to a fixed bug.  However, you definitely followed the docs, so I'll have to see if this can be improved.
[7:52pm] mharrah: in particular, xtests doesn't contain any sources, so xtests:test doesn't think it has anything to run
[...]

PR #2178, opened 2015-08-26
Run pending tests in scripted, consider their successes as failures

Duhemm:
This will allow to detect more easily when some scripted tests are fixed and will force us to re-enable them, thus helping to find future regressions.

The following tests have had their status changed:
- dependency-management/gh-1484-npe **pending -> test**  
  It was considered as flaky in #1958, but it seems to run fine now.
- dependency-management/publish-local **pending -> disabled**  
  It was considered as flaky in #1958.
- project/session-save **pending -> disabled**  
  Unreliable on Travis.
- source-dependencies/implicit-search-companion-scope **pending -> test**  
  I can't really say why it was pending, but it looks like it runs just fine. Maybe it's redundant with other tests?
- source-dependencies/trait-super **pending -> test**  
  It appears to have been fixed by #2160 
- tests/arguments-new **pending -> test**  
  Marked as pending since #813, but runs fine now.


typesafe-tools:

dwijnand:
`dependency-management / mvn-local` is also another one that is very flaky on Travis (re-running `mavenResolverPluginTest:scripted dependency-management/*2of2` for this PR in an attempt to have it go green).

Duhemm:
Thanks @dwijnand! I've disabled it.

jsuereth:
Why is session-save disabled?

Duhemm:
I have seen spurious failures on Travis. It has been marked as pending by @eed3si9n in #1958, but I don't have much more information.

jsuereth:
k. It's our only test of .sbt refactoring code, but we can merge as is for now.

PR #2183, opened 2015-08-27
Implement `projects` for .sbt files

Duhemm:
`projects` gives access to the list of projects that have been loaded
from .sbt files, just like it does for .scala files.

This works by registering all the projects that are discovered in .sbt
files in a global variable. The list of projects is then retrieved from
this global variable.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1824,6 +1824,11 @@ trait BuildExtra extends BuildCommon with DefExtra {
   import Defaults._
 
   /**
+   * The list of all projects that have been loaded.
+   */
+  def projects: Seq[Project] = BuildExtra.knownProjects.values.toSeq.flatten
+
+  /**
    * Defines an alias given by `name` that expands to `value`.
    * This alias is defined globally after projects are loaded.
    * The alias is undefined when projects are unloaded.
@@ -1956,6 +1961,11 @@ trait BuildExtra extends BuildCommon with DefExtra {
   def filterKeys(ss: Seq[Setting[_]], transitive: Boolean = false)(f: ScopedKey[_] => Boolean): Seq[Setting[_]] =
     ss filter (s => f(s.key) && (!transitive || s.dependencies.forall(f)))
 }
+object BuildExtra {
+  import scala.collection.mutable.Map
+  private[sbt] val knownProjects: Map[String, Seq[Project]] = Map.empty
+}
+
 trait DefExtra {
   private[this] val ts: TaskSequential = new TaskSequential {}
   implicit def toTaskSequential(d: Def.type): TaskSequential = ts
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -159,7 +159,8 @@ object Defaults extends BuildCommon {
     fork :== false,
     initialize :== {},
     forcegc :== sys.props.get("sbt.task.forcegc").map(java.lang.Boolean.parseBoolean).getOrElse(GCUtil.defaultForceGarbageCollection),
-    minForcegcInterval :== GCUtil.defaultMinForcegcInterval
+    minForcegcInterval :== GCUtil.defaultMinForcegcInterval,
+    projects := { buildStructure.value.units.values.map(_.defined).flatMap(_.values).toSeq }
   ))
   def defaultTestTasks(key: Scoped): Seq[Setting[_]] = inTask(key)(Seq(
     tags := Seq(Tags.Test -> 1),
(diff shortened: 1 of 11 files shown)


typesafe-tools:

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1956,6 +1961,11 @@ trait BuildExtra extends BuildCommon with DefExtra {
   def filterKeys(ss: Seq[Setting[_]], transitive: Boolean = false)(f: ScopedKey[_] => Boolean): Seq[Setting[_]] =
     ss filter (s => f(s.key) && (!transitive || s.dependencies.forall(f)))
 }
+object BuildExtra {
+  import scala.collection.mutable.Map
```
dwijnand: mutable? :-/ I feel like we should avoid mutable and side-effecting things..

eed3si9n:
Could you add a test and notes?

Duhemm:
Done

main/src/main/scala/sbt/Defaults.scala
eed3si9n: @Duhemm Is it possible to move this info into `State` somehow so it's immutable?

Duhemm: This is what I'm trying to achieve at the moment. It's most likely possible, but I'm afraid this would require a lot of changes.

Duhemm:
@dwijnand @eed3si9n It turns out everything could be implemented in a much simpler manner using a task. WDYT?

Moreover it's more correct because it include the auto-generated project (in case there's no root project for instance).

eed3si9n:
@Duhemm `projects` are used to aggregate other builds. You can't evaluate task there.

Duhemm:
From where should we get the `State` that holds the projects then?

jsuereth:
@Duhemm The `buildStructure` you're using is coming from the current state, so you are pulling it appropriately from one of the "magic" tasks :)

I would put a BIG HUGE WARNING on this feature that it's `read only` you _CANNOT_ add to projects via this task.  I'd even go so far as to call it "readOnlyProjectList" or something to denote that.

Otherwise looks good.

dwijnand:
> I would put a BIG HUGE WARNING on this feature that it's read only you CANNOT add to projects via this task. I'd even go so far as to call it "readOnlyProjectList" or something to denote that.

TIL. Either by name or docs we should communicate this.

eed3si9n:
See https://github.com/sbt/sbt/blob/v0.13.5/project/Sbt.scala#L30-L32:

``` scala
    override def projects = super.projects.map(p => p.copy(configurations = (p.configurations.filter(_ != Provided)) :+ myProvided))
    lazy val root: Project = Project("root", file("."), aggregate = nonRoots ) settings( rootSettings : _*) configs( Sxr.sxrConf, Proguard )
    lazy val nonRoots = projects.filter(_ != root).map(p => LocalProject(p.id))
```

`projects` method is passed on to `aggregate: => Seq[ProjectReference]`. Task won't be useful there.

jsuereth:
@Duhemm What's the point of this feature?

eed3si9n:
This is intended as the replacement for 

``` scala
trait Build {
  def projects: Seq[Project] = ReflectUtilities.allVals[Project](this).values.toSeq
}
```

eed3si9n:
@jsuereth Duhemm/sbt@ef2ffcf34b9bfa731734fca350338202ab5049de originally implemented it using a global variable. @Duhemm then tried it with tasks. Do you have better ideas on where we can keep this info?

dwijnand:
The uses I've seen of `Build.projects` is to programmatically define projects. So if the current task version is read-only it won't be a replacement of that particularly use case.

eed3si9n:
@jsuereth Here's ScopeFilter usage: https://github.com/sbt/sbt/blob/v0.13.5/project/Sbt.scala#L200-L202

``` scala
    def deepTasks[T](scoped: TaskKey[Seq[T]]): Initialize[Task[Seq[T]]] = deep(scoped.task) { _.join.map(_.flatten.distinct) }
    def deep[T](scoped: SettingKey[T]): Initialize[Seq[T]] =
        Util.inAllProjects(projects filterNot Set(root, sbtSub, scriptedBaseSub, scriptedSbtSub, scriptedPluginSub) map { p => LocalProject(p.id) }, scoped)
```

Duhemm:
How should I continue with this PR? Should we use the first version that used the global variable? Basically remove the last commit of this PR?

dwijnand:
Personally I like the AutoPlugin idea..

eed3si9n:
@Duhemm I think removing the last commit is a good starting point.

dwijnand:
Replaced by https://github.com/sbt/sbt/pull/2717, which is in 0.13, I'll forward port it soon.

eed3si9n:
@dwijnand We might still need `projects` for delegation purpose.

dwijnand:
Can't we get this effect with `loadedBuild.value.allProjectRefs.map(_._1)` or `loadedBuild.value.allProjectRefs.map(_._2)`?
- http://www.scala-sbt.org/release/sxr/sbt/Keys.scala.html#sbt.Keys.loadedBuild
- http://www.scala-sbt.org/release/sxr/sbt/BuildStructure.scala.html#sbt;LoadedBuild.allProjectRefs

eed3si9n:
`.value` can't be evaluated if you want to be used in `lazy val root = (project in file(".")).aggregate(projects: _*)`

dwijnand:
Isn't not a way to setup aggregation via settings? I know you can setup dependsOn with settings.

PR #2197, opened 2015-09-04
Fixes #2196, Specify the compiler bridge module in a setting

### LINKED ISSUE #2196: Simplify specific compiler bridge sources support
REF #2106. I take responsibility for the current design described in [sbt incrementlcompiler API plans](https://docs.google.com/document/d/1cf-yhJ4f8ZoHE6zAcqAlorXo9OheDQrH48zN8L__6ug/edit#).
As it stands, I am a bit concerned about the "repeatable build" issue, also the annoyance of hitting the Internet, which may not be required. (#2174)

For example, if a new Scala 2.12.0-M3 comes out, and tomorrow a specific compiler bridge comes out for it, depending on when you started the build, the outcome of the JAR may become different.

In the document it says:

> sbt should provide a Seq[ModuleID] setting called scalaCompilerBridgeSources. In the worst case scenario if Scala (or community fork) compiler break things without a recourse, a third party can supply a bridge.

We should actually provide `ModuleID` setting named `scalaCompilerBridgeSource`, and that setting should be the one describing the cascade logic in `Defaults.scala` using pattern match on `scalaVersion` key. This way, given a particular version of sbt (like `2.10.5` or `2.12.0-M3`), the build is going to be repeatable regardless of the state of the Internet.

Duhemm:
Fixes #2196

In order to restore reproducibility of builds, we no longer cascade over
the possibly available versions of the compiler bridge sources (a
specific version of the bridge sources may not be available one day, but
exist on the next day), but rather let the build definition configure
which module to use.

Also, download the compiler bridge sources and its dependencies in a temporary directory. We don't need to keep them around, because the next time we try to compile using this specific version of the compiler bridge, we will be able to immediately find the compiled version.

### DIFF
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -137,6 +137,7 @@ object Keys {
   val sbtPlugin = SettingKey[Boolean]("sbt-plugin", "If true, enables adding sbt as a dependency and auto-generation of the plugin descriptor file.", BMinusSetting)
   val printWarnings = TaskKey[Unit]("print-warnings", "Shows warnings from compilation, including ones that weren't printed initially.", BPlusTask)
   val fileInputOptions = SettingKey[Seq[String]]("file-input-options", "Options that take file input, which may invalidate the cache.", CSetting)
+  val scalaCompilerBridgeSource = SettingKey[ModuleID]("scala-compiler-bridge-source", "Configures the module ID of the sources of the compiler bridge.", CSetting)
 
   val clean = TaskKey[Unit]("clean", "Deletes files produced by the build, such as generated sources, compiled classes, and task caches.", APlusTask)
   val console = TaskKey[Unit]("console", "Starts the Scala interpreter with the project classes on the classpath.", APlusTask)
--- a/compile/ivy/src/main/scala/sbt/compiler/ComponentCompiler.scala
+++ b/compile/ivy/src/main/scala/sbt/compiler/ComponentCompiler.scala
@@ -15,7 +15,7 @@ object ComponentCompiler {
   val compilerInterfaceSrcID = compilerInterfaceID + srcExtension
   val javaVersion = System.getProperty("java.class.version")
 
-  @deprecated("Use `interfaceProvider(ComponentManager, IvyConfiguration, File)`.", "0.13.10")
+  @deprecated("Use `interfaceProvider(ComponentManager, IvyConfiguration, ModuleID)`.", "0.13.10")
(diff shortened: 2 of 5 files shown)


typesafe-tools:

dwijnand:
From as much as I understand what's going on, this LGTM.

eed3si9n:
LGTM

main/src/main/scala/sbt/Keys.scala
```diff
@@ -137,6 +137,7 @@ object Keys {
   val sbtPlugin = SettingKey[Boolean]("sbt-plugin", "If true, enables adding sbt as a dependency and auto-generation of the plugin descriptor file.", BMinusSetting)
   val printWarnings = TaskKey[Unit]("print-warnings", "Shows warnings from compilation, including ones that weren't printed initially.", BPlusTask)
   val fileInputOptions = SettingKey[Seq[String]]("file-input-options", "Options that take file input, which may invalidate the cache.", CSetting)
+  val scalaCompilerBridgeSource = SettingKey[ModuleID]("scala-compiler-bridge-source", "Configures the module ID of the sources of the compiler bridge.", CSetting)
```
jsuereth, line 140: Think this should be a sequence, so I could have it be the "hierarchy" if I wanted.

eed3si9n, line 140: That's what the original design document said too, but I think `Seq[ModuleID]` means you still are reliant on the state of Internet, which defies the purpose of this change.

Duhemm, line 140: Actually the idea is to replace the hierarchy by specifying directly in the build definition what version of the compiler bridge is required. After the compiler bridges have been split, it will become something like:

``` scala
scalaCompilerBirdgeSource := scalaVersion.value match {
  case v if v startsWith "2.11" => "org.scala-sbt" % "compiler-interface" % sbtVersion.value
  case _  => "org.scala-sbt" % "compiler-interface-old" % sbtVersion.value
}
```

This way we keep the flexibility, but we don't rely on the hypothetical existence of some artefacts. The same build could yield totally different result depending on the availability of a specific compiler interface source at some point.

jsuereth, line 140: The issue is about forward compatibility here.  If Scala 2.12 comes out and requires specific sources, then sbt 0.13.x will not work anymore until a new release comes out.

I understand the reasoning, I just wanted to err on the side of forward compatibility and letting the Scala team release their own bridge-sources if needed for milestone releases.   You guys do not remember the pain-days of pre-scala-bridge-stability.

eed3si9n:
Ref https://github.com/sbt/sbt/issues/681

PR #1940, opened 2015-03-25
Fixes #1900, more robustness in help system

DavidPerezIngeniero:
Fixes #1900 
This time, the pull request is based in the 0.13 branch.

PR #2207, opened 2015-09-16
Improve time measuring capabilities

DavidPerezIngeniero:
Currently, by setting `-Dsbt.task.timing=true`, time execution of tasks is measured.

It has the following limitations:
- A very long number of decimals, that makes it unreadable.  Example:

```
{file:/home/d/rds/project/}rds-build/compile:resources: 0.015903999999999998 ms
```
- Some tasks like `compile` are composed of a large number of tasks.  Many of them are quite quick to execute.   Normally, we want to know the long running tasks.
- If you call in the command line `sbt task1 task2 task3`, I'd like the report to be at the end.  Otherwise, I have to read the full output, that can be very verbose, in order to study running times.

Here is my solution:
- By setting the property `-Dsbt.task.timings.divider=number`, we select the power of 10 to divide the shown number.  If it is 0, the result will be nanoseconds.  3=microseconds, 6=milliseconds, 9=seconds.
- Info is aligned for improving readability:

```
Total time: 210 sec
  {file:/home/d/rds/}servision/compile:compileIncremental: 63 sec
  {file:/home/d/rds/}servision/test:compileIncremental   : 37 sec
  {file:/home/d/rds/}servision/*:update                  : 35 sec
  {file:/home/d/rds/}servision/*:compilaJython           : 20 sec
  */*:descargaPylib                                      : 11 sec
  {file:/home/d/rds/}iText/compile:compileIncremental    : 10 sec
  {file:/home/d/rds/}ddlUtils/compile:compileIncremental :  8 sec
```
- By setting the property `-Dsbt.task.timings.threshold=number`, tasks whose duration is lower than the given number will be hidden.  The threshold will take into account the selected unit set by `sbt.task.timings.divider`.  In this way we avoid noise.
- By setting the property `-Dsbt.task.timings.shutdown=true`, the report is generated when finishing the JVM, instead of when finishing running a specified task.

**Note:**

Commit 3fe9809 belongs to pull request https://github.com/sbt/sbt/pull/1940
I forgot to push this commit when doing the pull request.

### DIFF
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -160,8 +160,15 @@ object EvaluateTask {
   import TaskExtra._
   import Keys.state
 
+  lazy private val sharedProgress = new TaskTimings(shutdown = true)
+
   private[sbt] def defaultProgress: ExecuteProgress[Task] =
-    if (java.lang.Boolean.getBoolean("sbt.task.timings")) new TaskTimings else ExecuteProgress.empty[Task]
+    if (java.lang.Boolean.getBoolean("sbt.task.timings")) {
+      if (java.lang.Boolean.getBoolean("sbt.task.timings.shutdown"))
+        sharedProgress
+      else
+        new TaskTimings(shutdown = false)
+    } else ExecuteProgress.empty[Task]
 
   val SystemProcessors = Runtime.getRuntime.availableProcessors
 
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -183,7 +183,7 @@ object BuiltinCommands {
           Some(index.keyMap(key))
         catch {
           case NonFatal(ex) =>
-            s.log error ex.getMessage
+            s.log debug ex.getMessage
             None
         }
       }.collect { case Some(s) => s }.distinct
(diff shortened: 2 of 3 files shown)


typesafe-tools:

dwijnand:
A couple of thoughts..

We've recently switched the default branch to "1.0.x", so that contributions can go there, and so we don't have to keep porting/copying changes all over.

This has very recently changed, also it might be a little shaky as we're transitioning.

So this change should be rebased and pull requested against that branch (`main/src/main/scala/sbt/EvaluateTask.scala` is still part of "sbt" proper).

---

In terms of the actual change I think it's awesome, but I don't like the flags it's using (magic numbers, values unclear from the name without the explanation you've provided).

What if instead of `-Dsbt.task.timings.divider` it were `-Dsbt.task.timings.unit`, which accepted the unit strings in [scala.concurrent.duration.Duration](https://github.com/scala/scala/blob/v2.11.7/src/library/scala/concurrent/duration/Duration.scala#L72-L94) (somehow), so `ms` or `mins` or `minutes`

`-Dsbt.task.timings.threshold` would then be the string version of `Duration`, and therefore not take into account `divider` (so `-Dsbt.task.timings.unit=seconds -Dsbt.task.timings.threshold=1second`)

`-Dsbt.task.timings.shutdown=true` perhaps could be `-Dsbt.task.timings.print.on.shutdown=true` (and similarly the name of the property in the code) to be more expressive. Or maybe `-Dsbt.task.timings.on.shutdown=true`..

---

Again, I think this is a great change, thanks.

DavidPerezIngeniero:
I've followed your suggestions and created https://github.com/sbt/sbt/pull/2210

For simplicity, thredshold remains to be a number, in the same unit that the specified with the  `-Dsbt.task.timings.unit`.  You are free to create more ellaborate logic.

So, next version of SBT will be 1.0? Won't it?

dwijnand:
Yep, 1.0 or 1.0.0 I guess. Probably a bunch of milestones and release candidates before that though.

dwijnand:
Closed in favour of #2210 then.

PR #2210, opened 2015-09-17
Improve time measuring capabilities (1.0.X)

DavidPerezIngeniero:
Currently, by setting -Dsbt.task.timing=true, time execution of tasks is measured.

It has the following limitations:
- A very long number of decimals, that makes it unreadable. Example:

```
{file:/home/d/rds/project/}rds-build/compile:resources: 0.015903999999999998 ms
```
- Some tasks like compile are composed of a large number of tasks. Many of them are quite quick to execute. Normally, we want to know the long running tasks.
- If you call in the command line `sbt task1 task2 task3`, I'd like the report to be at the end. Otherwise, I have to read the full output, that can be very verbose, in order to study running times.

Here is my solution:
- By setting the property `-Dsbt.task.timings.unit=string`, we can select the desired unit: ns (nanoseconds), us (microseconds), ms (milliseconds), s (seconds).
- Info is aligned for improving readability:

```
Total time: 210 sec
  {file:/home/d/rds/}servision/compile:compileIncremental: 63 sec
  {file:/home/d/rds/}servision/test:compileIncremental   : 37 sec
  {file:/home/d/rds/}servision/*:update                  : 35 sec
  {file:/home/d/rds/}servision/*:compilaJython           : 20 sec
  */*:descargaPylib                                      : 11 sec
  {file:/home/d/rds/}iText/compile:compileIncremental    : 10 sec
  {file:/home/d/rds/}ddlUtils/compile:compileIncremental :  8 sec
```
- By setting the property -Dsbt.task.timings.threshold=number, tasks whose duration is lower than the given number will be hidden. The threshold will take into account the selected unit set by `sbt.task.timings.unit` property. In this way we avoid noise.
- By setting the property `-Dsbt.task.timings.on.shutdown=true`, the report is generated when finishing the JVM, instead of when finishing running a specified task.  It is more appropiate when SBT is used in batch mode, instead of interactive mode.

Note:

Commit 3fe9809 belongs to pull request #1940
I forgot to push this commit when doing the pull request, that is a very trivial issue.


typesafe-tools:

dwijnand:
I'll look into how to make `-Dsbt.task.timings.threshold` take a String that Duration parses.

DavidPerezIngeniero:
I don't understand why some checks fail.

I think that most people the units for threshold and display purposes will be the same.  So there is no need to  write the logic in the threshold for units.

dwijnand:
Tests are failing because pieces of sbt are being moved around in the namespace. So it's not compiling.

It's very possible that people will generally want a threshold in the same unit (milliseconds, seconds or minutes I expect), but I still think `-Dsbt.task.timings.threshold=1` or `-Dsbt.task.timings.threshold=10` is unclear. Especially unclear when the unit property affects the threshold property.

DavidPerezIngeniero:
I've created another option activable with `-Dsbt.task.timings.omit.paths=true`, that allows to transform reports like this:

```
  {file:/home/jenkins/jobs/Servision-develop/workspace/}servision/*:testIntegracionJython      : 137 sec
  {file:/home/jenkins/jobs/Servision-develop/workspace/}workspace/*:generaDespliega            :  99 sec
  {file:/home/jenkins/jobs/Servision-develop/workspace/}servision/*:checkDuplicatedDependencies:  94 sec
  {file:/home/jenkins/jobs/Servision-develop/workspace/}workspace/*:fabricaDojo                :  92 sec
  {file:/home/jenkins/jobs/Servision-develop/workspace/}workspace/izpack:createInstaller       :  91 sec
  {file:/home/jenkins/jobs/Servision-develop/workspace/}servision/*:compilaJython              :  82 sec
```

into this cleaner and more readable report:

```
servision/*:testIntegracionJython      : 137 sec
workspace/*:generaDespliega            :  99 sec
servision/*:checkDuplicatedDependencies:  94 sec
workspace/*:fabricaDojo                :  92 sec
workspace/izpack:createInstaller       :  91 sec
servision/*:compilaJython              :  82 sec
```

DavidPerezIngeniero:
Will it take long to integrate this pull request?

main/src/main/scala/sbt/TaskTimings.scala
```diff
@@ -5,15 +5,45 @@ import sbt.internal.util.RMap
...
+      System.err.println(s"Unknown sbt.task.timings.unit: $x.\nUsing milliseconds.")
+      ("ms", 6)
+  }
 
   type S = Unit
 
-  def initial = { start = System.nanoTime }
+  if (shutdown) {
+    start = System.nanoTime
+    Runtime.getRuntime.addShutdownHook(new Thread {
```
jsuereth, line 38: This could actually cause a massive memory leak when doing `reload` in sbt.    i.e. All previosuly loaded sbt-instances/build definitions will be leaving their 'shutdown hooks" registered and then all the classes we get left over.

I'm not sure what a good solution would be here, but I'm hesitant to throw this in as is.  It _is_ hidden behind a flag, so that's good.  But ideally we'd have more visibility into the sbt lifecycle so you could dump the report when a build is "unloaded", rather than just JVM shutdown.

DavidPerezIngeniero:
In the use case of launching an automated SBT build, `reload` should never be used.  I think `reload` is used mainly in interactive sessions.  In that case, no hook is present.

eed3si9n:
Maybe this pull request was created before 1.0.x branch was passing all the tests. Could you rebase against the latest to see if the tests will pass now?

DavidPerezIngeniero:
I have already merged the latest changes from the 1.0.x branch.

Fsero:
Any reason to not merge this?

DavidPerezIngeniero:
I wonder the same.

eed3si9n:
Sorry about the inactivity on this. I rebased this PR in https://github.com/sbt/sbt/pull/2938 and merged it.

DavidPerezIngeniero:
Thanks!

PR #2214, opened 2015-09-19
More thorough classfile location detection

stuhood:
Wider scale testing of #2085 (in a branch I'm working on, so perhaps not also in master) indicated that this method would fail for system classes. Let's patch that up.


typesafe-tools:

eed3si9n:
Could you send the same PR to https://github.com/sbt/io plz?

stuhood:
Created https://github.com/sbt/io/pull/13

PR #2224, opened 2015-09-29
bumping up incrementalcompiler to the latest and using Doc object

eed3si9n:


main/actions/src/main/scala/sbt/Doc.scala
```diff
@@ -26,9 +26,9 @@ object Doc {
     scaladoc(label, cache, compiler, Seq())
   def scaladoc(label: String, cache: File, compiler: AnalyzingCompiler, fileInputOptions: Seq[String]): Gen =
     cached(cache, fileInputOptions, prepare(label + " Scala API documentation", compiler.doc))
-  def javadoc(label: String, cache: File, doc: sbt.compiler.javac.JavaTools, log: Logger, reporter: Reporter): Gen =
+  def javadoc(label: String, cache: File, doc: sbt.internal.inc.javac.JavaTools, log: Logger, reporter: Reporter): Gen =
```
jsuereth, line 29: Should we have public APIs relying on private APIs, or should some of this be public?

jsuereth:
LGTM

PR #2226, opened 2015-09-29
Make object Import to alias types and values

eed3si9n:
This is from @Duhemm's build-import branch, rebased against the latest 1.0.x.

I'm trying to use Travis do the check.
See also https://travis-ci.org/Duhemm/sbt/builds/82708108


dwijnand:
I'm happy to break Build.scala builds by requiring them to add `import Import._`, but I don't think we should be creating forwarders for values/types in the internal package, or forwarders for packages under sbt, for instance complete.

eed3si9n:
@dwijnand I agree, but let's get to the state where we know what's broken, and what we can fix in the short term.

eed3si9n:
Note sbt/util#14.

eed3si9n:
dependency-management scripted tests are failing because they can no longer see `private[sbt] def matching(f: DependencyFilter)` on `RichUpdateReport`. I need to move that into internal or something.
Eventually majority of these should move to librarymanagement repo.

eed3si9n:
Finally green from Travis. I suggest we merge first and then tweak some details later.

dwijnand:
> I suggest we merge first and then tweak some details later.

OK, merge away then. Some notes from me:
- re-enable maven resolver plugin using scripted tests
- deprecate forwarders we're providing as a migration path, for instance internal API things that we intend to no longer forward in the future (sbt.internal.inc.Analysis?)
- move types/values out of internal that we've found out (and agree) are actually API (eg. sbt.complete)
- create a migration guide in the docs

dwijnand:
Btw thanks for sorting this, that's awesome.

PR #2230, opened 2015-10-01
Always export apiURL even if autoAPIMappings is false.

jkinkead:
This is a simple change to make sbt honor the value of `apiURL` all of the time. It's not clear from reading [the documentation](http://www.scala-sbt.org/0.13/docs/Howto-Scaladoc.html) that this is totally ignored when `autoAPIMappings` is false - and it's also an unintuitive interaction, since one setting affects how you generate Scaladoc, while the other affects how you generate your pom.xml.


typesafe-tools:

jsuereth:
LGTM and please update this for the 1.0 branch :)

jkinkead:
Done! :)

dwijnand:
Closing in favour of #2262 then.

Issue #1933, opened 2015-03-22
False positive in the inconsistent duplicate warnings


dwijnand:
# steps

1) Given `sbt.version=0.13.8`, setup `build.sbt` as:

```
scalaVersion := "2.11.6"
libraryDependencies += "org.scoverage" %% "scalac-scoverage-runtime" % "1.0.4" % "provided"
```

2) Invoke `updateClassifiers`
# problem

```
[warn] Multiple dependencies with the same organization/name but different versions. To avoid conflict, pick one version:
[warn]  * org.scala-lang:scala-library:(2.11.6, 2.11.4)
```
# expectation

No warnings.
# note

Original report below:

@fommil indicates [on the mailing lists](https://groups.google.com/d/topic/sbt-dev/61J8cnQrsxU/discussion) (and @sslavic [as well](https://groups.google.com/d/msg/sbt-dev/w6dx6WWyAOE/0aJv8pbD9_gJ)) that there seems to be a false positive in the inconsistent duplicate warnings.

From Sam's post:

> Since upgrading to sbt 0.13.8 I get a _lot_ of errors of this nature (i.e. same thing repeated over and over):

```
[warn] Multiple dependencies with the same organization/name but different versions. To avoid conflict, pick one version:
[warn]  * org.scala-lang:scala-library:(2.11.6, 2.11.4)
[warn]  * org.scala-lang.modules:scala-parser-combinators_2.11:(1.0.1, 1.0.2, 1.0.3)
[warn]  * org.scala-lang.modules:scala-xml_2.11:(1.0.1, 1.0.3)
```

> I'm actually depending on explicit versions of all these libraries so I believe this is a bug in the warning code.

Running `gen-ensime` on the [ensime-server](https://github.com/ensime/ensime-server) build will reproduce the problem.

fommil:
Good point actually, this now occurs on the ensime master branch, since we upgraded sbt

ijuma:
It would be good to know if there is a workaround for this without disabling all eviction warnings or whether downgrading back to 0.13.7 is the only way to get rid of the excessive warnings.

jsuereth:
It looks like transitive eviction is leaking warnings here.

I.e. you're seeing the various evictions which take place when looking for your dependencies.   These should not all be showing up as root level, though, I think.

dwijnand:
Any ideas why this only occurs when `% "provided"`? Even `% "test"` doesn't cause it..

EECOLOR:
[...]

Issue #2256, opened 2015-10-27
"javac returned nonzero exit code" error with no stacktrace or details in SBT 0.13.8+


JoshRosen:
Consider the following invalid `javacOptions` config:

``` scala
scalacOptions ++= Seq("-target:jvm-1.6"),
javacOptions ++= Seq("-source", "1.7", "-target", "1.6"),
```

With SBT 0.13.7, this results in the following error message:

```
[error] javac: source release 1.7 requires target release 1.7
[error] (compile:compile) javac returned nonzero exit code
```

In SBT 0.13.8+, on the other hand, the error details are missing, making this hard to debug:

```
[error] (compile:compileIncremental) javac returned nonzero exit code
```

I think that SBT's handling of `javac` error output must have changed in 0.13.8 because there are reports from other users that describe similar symptoms:
- https://groups.google.com/forum/#!msg/play-framework/UqNeFuoCZ0Q/GSTjBDV6CWcJ
- https://stackoverflow.com/questions/31132664/how-to-find-out-what-happened-when-sbt-says-javac-returned-nonzero-exit-code

dwijnand:
Yes, things changed in 0.13.8 with #1702

shajra:
Just curious, do you know when this is going to hit an SBT release?  Is there a workaround?  We've got at least one module that for now needs to be pure-Java, and this problem is kind of annoying.  It feels weird pinning us to 0.13.7 just for this (though I'm not sure if we're really using any features in 0.13.8+ beyond bug fixes)

eed3si9n:
@shajra Hopefully soon. You can try the nightly from here in the meantime - https://dl.bintray.com/sbt/ivy-snapshots/org.scala-sbt/sbt-launch/0.13.10-20151229-205306/

shajra-cs:
@eed3si9n good to know.  Using a nightly is probably more awkward than just downgrading temporarily to 0.13.7, but at least now I know where the nightlies are.  Also, I have this thread of conversation to point coworkers to if they are curious about the state of things.  Thanks.

shajra-cs:
@eed3si9n actually, I shouldn't assume.  What's the best way to use that URL for the SBT 0.13.10 snapshot?  We're using Paul's "sbt-extras" script.

eed3si9n:
[...]

PR #2258, opened 2015-10-28
Fix for sbt/sbt#1933

Duhemm:
sbt was reporting warning abouts inconsistent versions of dependencies
even if these dependencies didn't have the same configuration (as in
`provided` vs `compile`).

This commit fixes this problem by comparing the dependencies by
organization, artifact name and configuration.

Fixes sbt/sbt#1933

### DIFF
--- a/ivy/src/main/scala/sbt/Ivy.scala
+++ b/ivy/src/main/scala/sbt/Ivy.scala
@@ -536,14 +536,18 @@ private[sbt] object IvySbt {
     {
       import IvyRetrieve.toModuleID
       val dds = moduleID.getDependencies
-      inconsistentDuplicateWarning(dds map { dd => toModuleID(dd.getDependencyRevisionId) })
+      val deps = dds flatMap { dd =>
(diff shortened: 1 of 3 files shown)


typesafe-tools:

dwijnand:
Nice!

Should this target 0.13? Also, would be great to have release notes for this.

Duhemm:
Good catch! I just added the notes.

I think @eed3si9n would like to release a 0.13.10 at some point

eed3si9n:
I was thinking about 0.13.10 but it may or may not happen.

eed3si9n:
@Duhemm Could you send the same PR to sbt/librarymanagement?

max-voloshyn:
Thanks for the fix! Is there any plans to release `sbt` version with this fix or ability to use this fix via `brew`?

dwijnand:
@max-voloshin See @eed3si9n's very recent answer: https://github.com/sbt/sbt/issues/2256#issuecomment-168457666

max-voloshyn:
@dwijnand thanks!

Issue #1171, opened 2014-03-11
Incremental compiler misses change to value class, and results to NoSuchMethodError at runtime


xuwei-k:
see https://github.com/xuwei-k/sbt-value-class-bug
### step

From sbt shell, `compile` A.scala (without value class)

``` scala
case class A(a: String)
```

B.scala:

``` scala
object B {
  def foo = A("a")
}
```

Change to A.scala to following:

``` scala
case class A(a: String) extends AnyVal
```

From sbt shell, `run` (without `clean` )
### problem

```
[info] [error] (run-main-0) java.lang.NoSuchMethodError: B$.foo()LA;
[info] java.lang.NoSuchMethodError: B$.foo()LA;
[info]  at C$delayedInit$body.apply(C.scala:3)
[info]  at scala.Function0$class.apply$mcV$sp(Function0.scala:40)
[info]  at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:12)
[info]  at scala.App$$anonfun$main$1.apply(App.scala:71)
[info]  at scala.App$$anonfun$main$1.apply(App.scala:71)
[info]  at scala.collection.immutable.List.foreach(List.scala:318)
[info]  at scala.collection.generic.TraversableForwarder$class.foreach(TraversableForwarder.scala:32)
[info]  at scala.App$class.main(App.scala:71)
[info]  at C$.main(C.scala:1)
[info]  at C.main(C.scala)
[info]  at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[info]  at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
[info]  at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[info]  at java.lang.reflect.Method.invoke(Method.java:606)
[info] [trace] Stack trace suppressed: run last compile:run for the full output.
[info] java.lang.RuntimeException: Nonzero exit code: 1
[info]  at scala.sys.package$.error(package.scala:27)
[info] [trace] Stack trace suppressed: run last compile:run for the full output.
[info] [error] (compile:run) Nonzero exit code: 1
[info] [error] Total time: 4 s, completed Mar 11, 2014 9:32:49 AM
[error] x value-class-bug / simple
```
### notes

`NoSuchMethodError` does not occur if `clean` is executed first.

eed3si9n:
[...]

PR #2261, opened 2015-11-04
Consider signatures of methods before and after erasure in ExtractAPI

Duhemm:
The signatures of methods that have value classes as arguments or return
type change during the erasure phase. Because we only registered
signatures before the erasure, we missed some API changes when a class
was changed to a value class (or a value class changed to a class).

This commit fixes this problem by recording the signatures of method
before and after erasure.

Fixes sbt/sbt#1171

### DIFF
--- a/compile/interface/src/main/scala/xsbt/ExtractAPI.scala
+++ b/compile/interface/src/main/scala/xsbt/ExtractAPI.scala
@@ -180,11 +180,13 @@ class ExtractAPI[GlobalType <: CallbackGlobal](val global: GlobalType,
   private def printMember(label: String, in: Symbol, t: Type) = println(label + " in " + in + " : " + t + " (debug: " + debugString(t) + " )")
   private def defDef(in: Symbol, s: Symbol): List[xsbti.api.Def] =
     {
-      def isMacro(sym: Symbol): Boolean =
-        sym.isMacro || (sym.info.members.sorted exists isMacro) || (sym.children exists isMacro)
-        //sym.isMacro || (sym.children exists isMacro) || (sym.isType && sym.asType.toType.members.sorted.exists(isMacro))
-      val inspectPostErasure = !isMacro(in.enclosingTopLevelClass)
-
+      def collectAll(acc: Set[Symbol], s: Symbol): Set[Symbol] =
+        if (acc contains s) acc
+        else
+          ((s.info.members.sorted ++ s.children) foldLeft (acc + s)) {
+            case (acc, sym) => collectAll(acc, sym)
+          }
+      val inspectPostErasure = !collectAll(Set.empty, in.enclosingTopLevelClass).exists(_.isMacro)
       def build(t: Type, typeParams: Array[xsbti.api.TypeParameter], valueParameters: List[xsbti.api.ParameterList]): List[xsbti.api.Def] =
         {
           def parameterList(syms: List[Symbol]): xsbti.api.ParameterList =
--- a/compile/interface/src/main/scala/xsbt/ExtractAPI.scala
+++ b/compile/interface/src/main/scala/xsbt/ExtractAPI.scala
@@ -178,9 +178,9 @@ class ExtractAPI[GlobalType <: CallbackGlobal](val global: GlobalType,
 
   private def viewer(s: Symbol) = (if (s.isModule) s.moduleClass else s).thisType
   private def printMember(label: String, in: Symbol, t: Type) = println(label + " in " + in + " : " + t + " (debug: " + debugString(t) + " )")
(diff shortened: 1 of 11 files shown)


typesafe-tools:

eed3si9n:
@Duhemm Could you fix the codacy failure due to the lack of return type signature on public method, plz?

dwijnand:
Very cool. Nice to see old, old bugs get fixes. Well done.

eed3si9n:
This needs to be forward ported.

compile/interface/src/main/scala/xsbt/ExtractAPI.scala
```diff
@@ -192,13 +215,56 @@ class ExtractAPI[GlobalType <: CallbackGlobal](val global: GlobalType,
...
+              //   class A(val x: Int) extends AnyVal
+              //   def foo(a: A): Int = A.x <- has type (LA)I before erasure
+              //                            <- has type (I)I after erasure
+              // If we change A from value class to normal class, we need to recompile all clients
+              // of def foo.
+              val beforeErasure =
+                build(resultType, typeParams, parameterList(params) :: valueParameters)
+              val afterErasure  =
+                if (inspectPostErasure)
+                  build(resultType, typeParams, global exitingPostErasure (parameterList(mType.params) :: valueParameters))
```
dwijnand, line 232: See https://github.com/scala/scala/commit/6dd149505d6782d35655cfd76e573e7fbeaecb77

(part of https://github.com/scala/scala/compare/2.12.x...adriaanm:sbt-api-consolidate)

/cc @adriaanm

adriaanm, line 232: Thanks for tracking those changes! I think it would be better to avoid phase travel where possible, but I'd have to think about my changes a bit more before recommending they be put into production at sbt. My sbt-api-consolidate branch is solidly in experimentation territory ;-)

dwijnand, line 232: No problem, just happened across that branch from somewhere and wanted to bring it to @Duhemm's attention.

Truth be told I have no idea what the code or you are talk about :P Learning by osmosis, diff-by-diff.

gkossakowski:
Did you guys measure performance impact of patch? I'm a bit worried about it because it doubles the size of API extracted. The API extraction is already the slowest part of the incremental compiler.

gkossakowski:
//cc @Duhemm

Duhemm:
@gkossakowski I didn't. I think that the increase of the size of the API has been somewhat mitigated by #2414. I'll try to look into it.

gkossakowski:
Oh, I didn't see #2414. It will indeed help with API structure size.

Do you need to detect whether a type in a signature just changed from a value class to a regular class or do you need the full information about what it erases to?

smarter:
It's still bigger than it needs to be, consider:

``` scala
class A(val self: Int) extends AnyVal
class B {
  def foo(a: A): A = new A(1)
}
```

Then ExtractAPI returns four signatures for `foo` but we only need twos (non-erased and fully-erased):

``` scala
def foo(a: this#A): this#A
def foo(a: this#A): scala.this#Int
def foo(a: scala.this#Int): this#A
def foo(a: scala.this#Int): scala.this#Int
```

gkossakowski:
Would it be enough to introduce a special `TypeRef` that is used to referring to value classes and then return just one signature of `foo`? The switch from AnyVal to AnyRef would be detected by change from `ValueClassTypeRef` to `TypeRef`.

smarter:
I think you still need to know what the underlying type is, consider three files `A.scala`, `B.scala`, `C.scala`:

``` scala
class A(val self: Int) extends AnyVal

object B {
  def foo: A = new A(1)
}

object C {
    B.foo
  }
}
```

If we change the underlying type of `A` from `Int` to `Double`, then `B.scala` will be recompiled. But if you don't register that the underlying type of `A` has changed, then `C.scala` won't be recompiled and will fail at runtime, because the signature of `B.foo` in `C.class` will be incorrect.

gkossakowski:
Yes, you're right. However, the issue is that there're two different ways to refer to `A` (before erasure and after erasure) so why can't we encode that in `ValueClassTypeRef` that would contain reference to unerased and erased type? This way the rest of api extraction can remain oblivious to value classes.

smarter:
>  why can't we encode that in ValueClassTypeRef that would contain reference to unerased and erased type

That seems like a nice idea, though instead of having a reference to the erased type, I'd like to have a reference to the underlying type. In Dotty, a value class can have another value class as its underlying type.

PR #2262, opened 2015-11-04
Always export apiURL even if autoAPIMappings is false.

jkinkead:
Copy of #2230, against 1.0.x.


typesafe-tools:

eed3si9n:
Re-running a few Travis jobs, but generally LGTM

dwijnand:
LGTM

eed3si9n:
Forgot about the notes. Could you send another PR with a note entry here? - https://github.com/sbt/sbt/tree/1.0.x/notes/0.13.10

Issue #2263, opened 2015-11-09
sbt 0.13.10-SNAPSHOT fails to compile akka/akka


Duhemm:
While testing #2261 on a large codebase (akka), sbt kept crashing. I first thought it was because of my changes, but it turns out that sbt cannot compile akka since #2085 and #2214 have been merged.
# Steps
1. Checkout sbt/sbt@33478132c5d0f4e109d85522ba52fd77bf470a9d
2. build and publish sbt locally.
3. Clone akka/akka and checkout akka/akka@18a24c0e5036ad97281732e87f33aea214caf8e3 (note: It looks like compiling akka requires java 8)
4. Compile akka using the freshly built sbt (change/remove `project/build.properties`).
# Problem

Compilation fails with the following stack trace:

```
java.lang.IllegalStateException: Failed to parse class java.util.AbstractMap$SimpleImmutableEntry: this may mean your classfiles are corrupted. Please clean and try again.
    at sbt.ClassToAPI$.fieldToDef(ClassToAPI.scala:162)
    at sbt.ClassToAPI$$anonfun$8.apply(ClassToAPI.scala:78)
    at sbt.ClassToAPI$$anonfun$8.apply(ClassToAPI.scala:78)
    at sbt.ClassToAPI$$anonfun$mergeMap$1.apply(ClassToAPI.scala:227)
    at sbt.ClassToAPI$$anonfun$mergeMap$1.apply(ClassToAPI.scala:227)
    at scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:251)
    at scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:251)
    at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
    at scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:34)
    at scala.collection.TraversableLike$class.flatMap(TraversableLike.scala:251)
    at scala.collection.AbstractTraversable.flatMap(Traversable.scala:105)
    at sbt.ClassToAPI$.merge(ClassToAPI.scala:233)
    at sbt.ClassToAPI$.mergeMap(ClassToAPI.scala:227)
    at sbt.ClassToAPI$.structure(ClassToAPI.scala:78)
    at sbt.ClassToAPI$.x$2$lzycompute$1(ClassToAPI.scala:66)
[...]

PR #2268, opened 2015-11-11
Fix support for parsing of classes located in runtime jars

stuhood:


typesafe-tools:

eed3si9n:
Fixes #2263?

eed3si9n:
Coincidently I haven't merged months old sbt/io/pull/13 yet. Could you update that PR too?

util/io/src/main/scala/sbt/IO.scala
```diff
@@ -36,39 +36,43 @@ object IO {
...
-      // NB: This assumes that classes without code sources are System classes, and thus located in
-      // jars. It assumes that `urlAsFile` will truncate to the containing jar file.
-      val clsfile = s"${cl.getName.replace('.', '/')}.class"
-      Option(ClassLoader.getSystemClassLoader.getResource(clsfile))
-        .flatMap {
-          urlAsFile
-        }.getOrElse {
-          sys.error("No class location for " + cl)
-        }.toURI.toURL
+  def classfileLocation(cl: Class[_]): URL = {
```
dwijnand, line 42: Bincompat?

dwijnand:
Is this a fix to an improvement for which we have notes or is it a fix that requires its own notes?

eed3si9n:
@dwijnand I think this is an improvement to #2214, which is a part of #2085, which does have a note entry.

dwijnand:
Cool thanks.

util/io/src/main/scala/sbt/IO.scala
dwijnand, line 42: @eed3si9n Do we want to do something about this binary compatibility breakage before cutting a 0.13.10 release?

eed3si9n, line 42: Yes. Could you send a PR or issue against 0.13 to fix this?

dwijnand, line 42: sure

dwijnand:
Also, this hasn't been forward-ported to 1.0, which I realise is a little hard as it's touching files in different modules..

Issue #2228, opened 2015-09-30
sbt's handling of javac errors differs between running locally and forked.


baroquebobcat:
We ran into this when javac's encoding had been accidentally defaulted to ascii. We have a java file with a comment containing utf-8 characters. When sbt compiles with a forked javac, it fails to compile. When sbt compiles with a local javac it compiles successfully, but prints the errors from javac as errors.

This is confusing because users expect compiler errors to cause a compilation failure.

The thing that's interesting is that it fails to compile as expected both locally and forked when the utf-8 character is in a char literal or a string literal.

I was able to verify this with the source version of sbt on the 0.13 branch with a change to JavaCompilerSpec, but I'm not sure how to approach fixing it.

eed3si9n:
Which version of sbt are you using? I'm guessing this is related to #1702

baroquebobcat:
0.13.9, but I was able to reproduce on the head of the 0.13 branch.

Duhemm:
Surprisingly enough, the `success` reported by sbt is actually correct and the java compiler actually creates the classfiles:

```
> compile
[info] Updating {file:/Users/martin/Desktop/tmp/}tmp...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
[info] Compiling 1 Java source to /Users/martin/Desktop/tmp/target/scala-2.10/classes...
[error] /Users/martin/Desktop/tmp/src/main/java/X.java:3: unmappable character for encoding ascii
[error] /Users/martin/Desktop/tmp/src/main/java/X.java:3: unmappable character for encoding ascii
[error] /Users/martin/Desktop/tmp/src/main/java/X.java:3: unmappable character for encoding ascii
[error] /Users/martin/Desktop/tmp/src/main/java/X.java:3: unmappable character for encoding ascii
[success] Total time: 0 s, completed Nov 2, 2015 2:37:21 PM
> run
[info] Running X 
Foobar! ����
```

I suppose all the messages reported as `ERROR` to the `DiagnosticsReporter` should be flushed as `WARNING` if the compilation actually succeeded?

eed3si9n:
> I suppose all the messages reported as ERROR to the DiagnosticsReporter should be flushed as WARNING if the compilation actually succeeded?

We have a stringly logic that determines what's an error and what's a warning. I don't know if we should change that completely. Maybe this needed to be added as a warning. /cc @jsuereth

Duhemm:
[...]

PR #2271, opened 2015-11-11
Report failed compilation if errors are registered.

Duhemm:
In some cases, the local java compiler may report compilation errors,
but succeed in compiling the source files (for instance, if there are
encoding problems in the source). However, the command line javac will
report a failed compilation on the same input.

To have the local java compiler behave like the forked java compiler, we
now report the compilation as failed if error messages have been
registered during compilation.

Fixes sbt/sbt#2228


typesafe-tools:

dwijnand:
LGTM (for what I can see, not what I know :D)

PR #2281, opened 2015-11-20
Fix search result highlighting on Windows

### LINKED ISSUE #1982: Regex messes up colors
I'm using SBT 0.13.8 on Windows. When filtering tasks or settings using a regex, the foreground color changes to dark red, and stays that way.

image

eamelink:
Use only ANSI color codes that are supported by Windows. Fixes #1982


typesafe-tools:

eed3si9n:
LGTM. Thanks for the contribution

Issue #2264, opened 2015-11-09
updateClassifiers does not respect Ivy configuration mappings


brunon:
When a dependency uses non-standard configurations, and does not have a "compile" configuration, it is possible to include it in libraryDependencies using a configuration mapping like this

``` sbt
libraryDependencies += "ossjava" % "freemarker" % "2.3.19" % "compile->runtime"
```

(This is a private library available on a filesystem resolver, so don't look for it on the Web)

This makes the "update" task work. **However, "updateClassifiers" does not seem to detect the Ivy configuration mapping and still looks for "compile" and fails**

This is the ivy.xml that is looked up by the filesystem resolver, you'll notice it has only "runtime" nothing else

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<ivy-module version="2.0" xmlns:extra="http://ant.apache.org/ivy/extra">

  <info organisation="ossjava" module="freemarker" revision="2.3.19">
  </info>

  <configurations>
    <conf name="runtime" description="Core runtime dependencies"/>
  </configurations>

  <publications>
    <artifact name="freemarker-2.3.19" type="src" ext="src.jar" conf="runtime"/>
    <artifact name="freemarker-2.3.19" type="jar" ext="jar" conf="runtime"/>
  </publications>

  <dependencies/>

</ivy-module>
```

Finally, this is the message that SBT provides:

```
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]  ::          UNRESOLVED DEPENDENCIES         ::
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]  :: ossjava#freemarker;2.3.19: configuration not found in ossjava#freemarker;2.3.19: 'default(compile)'. Missing configuration: 'compile'. It was required from default#a$sources_javadoc_2.10;0.1-SNAPSHOT test-internal
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
```

More debug output:

```
> last-grep freemarker
[debug] == resolving dependencies default#a$sources_javadoc_2.10;0.1-SNAPSHOT->ossjava#freemarker;2.3.19 [compile->default(compile)]
[info] Resolving ossjava#freemarker;2.3.19 ...
[debug] sbt-chain: module revision found in cache: ossjava#freemarker;2.3.19
[debug]         found ossjava#freemarker;2.3.19 in sbt-chain
[...]

PR #2297, opened 2015-12-02
Fixes #2264. Fixes updateClassifiers when default configuration is missing + uses listed artifacts if any

eed3si9n:
Fixes #2264 
Continues #2296

Even though it's not really used, updateClassifiers constructs
dependency graph based on the result from update.
The direct cause of #2264 came from the fact that the `allModules`
returned from ConfigurationReport did not include dependency
configurations. For example it returned "compile" instead of
"compile->runtime". I've identified that in #2264 and was fixed by
@Duhemm in f49fb33.

Martin identified that the fix still does not address the fact that
updateClassifier hardcodes the classifiers to be tried. This commit
adds the fallback behavior so for Ivy-published modules it will use the
explicit list of artifacts, and for others it will fallback to the
hardcoded list of classifiers.

PR #2300, opened 2015-12-07
(Do not merge) Fixes sbt/sbt#2264. Use explicit artifacts if any, fallback to hardcoded

Duhemm:
This is a port for sbt/sbt#2297. This won't compile until we have a new release of sbt/librarymanagement that includes both sbt/librarymanagement#9 and sbt/librarymanagement#13.


typesafe-tools:

eed3si9n:
This change is in direct conflict with the change added in #2478.
Maybe we need to add a separate task for `updateClassifiers` and the task that grabs sources and docs.

eed3si9n:
I'm going to close this PR for now.

PR #2303, opened 2015-12-07
Fixes #2302. Don't check pre and post conditions in sbt.Execute by default

### LINKED ISSUE #2302: sbt.Execute pre and post condition checking has bad run time characteristics for big graphs
The test project shown in #2266 spends about 40 seconds just validating the post conditions before doing anything. The reason is that the post condition checking in `addNew` seems to have at least O(n^2) run time in the depth of the graph. That's because, it recursively creates dependency tasks and but always checks invariants for the complete transitive hull.

jrudolph:
The checking code has bad run time characteristics and would need to be fixed for large projects with deep task dependency chains.

The code in sbt.Execute has been in production for a long time so it seems safe enough to drop the extra checks by default. To debug issues, you can set `-Dsbt.execute.extrachecks=true` to revert to the old behavior.

Fixes #2302.

### DIFF
--- a/tasks/src/main/scala/sbt/Execute.scala
+++ b/tasks/src/main/scala/sbt/Execute.scala
@@ -21,6 +21,8 @@ private[sbt] object Execute {
 
   def config(checkCycles: Boolean, overwriteNode: Incomplete => Boolean = const(false)): Config = new Config(checkCycles, overwriteNode)
   final class Config private[sbt] (val checkCycles: Boolean, val overwriteNode: Incomplete => Boolean)
+
+  final val checkPreAndPostConditions = sys.props.get("sbt.execute.extrachecks").exists(java.lang.Boolean.parseBoolean)
 }
 sealed trait Completed {
   def process(): Unit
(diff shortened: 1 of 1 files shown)


typesafe-tools:

eed3si9n:
LGTM pending Travis.

fommil:
hmm, where is your repo @jrudolph ?

fommil:
oh wow, your fork is `xsbt`. You're properly hardcore.

fommil:
not sure about "fixes" but this certainly improves things! I'm seeing an `update` in https://github.com/fommil/sbt-big-project/tree/master/src/sbt-test/sbt-big-project/clusterfsck drop from 68s down to 33s. Which is awesome.

eed3si9n:
Could you also add a release not entry in https://github.com/sbt/sbt/tree/0.13/notes/0.13.10?

jrudolph:
I now have a build that just hangs which would have failed with an assertion without this change. Maybe the assertion that notices that no tasks are runnable any more should be enabled in any case.

PR #2311, opened 2015-12-14
Bumping Scala version to 2.10.6.

eed3si9n:


dwijnand:
:+1: what about in `project/p.sbt`?

eed3si9n:
@dwijnand Good call. Fixed it.

PR #2321, opened 2015-12-16
Fix bincompat from change in #2268

dwijnand:


typesafe-tools:

dwijnand:
Travis finally passed. Ready for review.

eed3si9n:
I think you're missing the semantic difference introduced in 9fbbaab58e246bce7b21117e079206519c386b57. The old one returns directory while the new one returns the actual *.class location IIUC.

dwijnand:
Which change(s) in particular?

I'm a bit lost. My intent here is just the restore binary compatibility. Any semantic difference introduced in that commit was accepted and is already in 0.13. Do you want to revert that and restore the previous semantics?

eed3si9n:
See

```
-   * Returns a URL for the directory or jar containing the the class file `cl`.
+   * Returns a URL for the classfile containing the given class
```

I think we need to add both the implementation. /cc @stuhood

eed3si9n:
Closing this in favor of #2342

PR #2322, opened 2015-12-18
Drop sealed from the typeclasses in Append

dwijnand:


typesafe-tools:

dwijnand:
Thanks!

eed3si9n:
Could you forward port this to 1.0.x?

dwijnand:
Will do.

PR #2342, opened 2015-12-30
[0.13.10] Fix bincompat issue introduced in #2268

eed3si9n:
Continuation of #2321
Ref #2268

### DIFF
--- a/util/io/src/main/scala/sbt/IO.scala
+++ b/util/io/src/main/scala/sbt/IO.scala
@@ -77,13 +77,22 @@ object IO {
    * If the location cannot be determined or if it is not a file, an error is generated.
    * Note that Java standard library classes typically do not have a location associated with them.
    */
-  def classLocationFile(cl: Class[_]): File = {
-    val classURL =
-      Option(cl.getProtectionDomain.getCodeSource).getOrElse {
-        sys.error("No class location for " + cl)
-      }.getLocation
-    toFile(classURL)
-  }
+  def classLocationFile(cl: Class[_]): File =
+    Option(cl.getProtectionDomain.getCodeSource) match {
+      case Some(codeSource) =>
+        val classURL = codeSource.getLocation
+        toFile(classURL)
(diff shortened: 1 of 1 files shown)


util/io/src/main/scala/sbt/IO.scala
```diff
@@ -68,18 +86,29 @@ object IO {
   }
```
dwijnand: Do you want to restore the semantics that were changed here as well?

https://github.com/sbt/sbt/commit/9fbbaab58e246bce7b21117e079206519c386b57?diff=unified#diff-dd0ad8518ad4dee1e8ab8f4f1d69cbe4R61

dwijnand:
Ah I now understand what you meant in #2321 now...

Sorry for dropping the ball there and picking up the slack. Thanks, 1 comment.

util/io/src/main/scala/sbt/IO.scala
eed3si9n: I think the semantics for the `def classLocationFile(cl: Class[_]): File` remain the same.

dwijnand: Looks to me like it used to fallback to using the system classloader (via old `classLocation(cl: Class[_]): URL`) and now it doesn't.

eed3si9n: You're right. Do you want to take over porting the fixing on to this branch?

dwijnand: I'm prepping to clean re-install my laptop so I can't ATM, sorry.

I will if this is still like this when I'm setup again (maybe this weekend maybe in 10 days..)

eed3si9n: No problem. Added 068cffa57835530a91f8b798b482ef0a94200b33

dwijnand:
Excellent, thank you. LGTM pending Travis.

PR #2344, opened 2015-12-30
Make sbt aware of Dotty

smarter:
This small set of changes, together with the compiler-bridge I wrote
(https://github.com/smarter/dotty-bridge) enables us to compile code
using Dotty in sbt, see https://github.com/smarter/dotty-example-project
for an example.

It'd be great if this could go in 0.13.10 as using a custom sbt is very cumbersome, let me know if there's anything I can do to help with that!

### DIFF
--- a/compile/src/main/scala/sbt/compiler/RawCompiler.scala
+++ b/compile/src/main/scala/sbt/compiler/RawCompiler.scala
@@ -21,16 +21,29 @@ class RawCompiler(val scalaInstance: xsbti.compile.ScalaInstance, cp: ClasspathO
 
     val arguments = compilerArguments(sources, classpath, Some(outputDirectory), options)
     log.debug("Plain interface to Scala compiler " + scalaInstance.actualVersion + "  with arguments: " + arguments.mkString("\n\t", "\n\t", ""))
-    val mainClass = Class.forName("scala.tools.nsc.Main", true, scalaInstance.loader)
-    val process = mainClass.getMethod("process", classOf[Array[String]])
-    process.invoke(null, arguments.toArray)
-    checkForFailure(mainClass, arguments.toArray)
+    val args = arguments.toArray
+    val reporter =
+      if (ScalaInstance.isDotty(scalaInstance.version)) {
+        val mainClass = Class.forName("dotty.tools.dotc.Main", true, scalaInstance.loader)
+        val process = mainClass.getMethod("process", classOf[Array[String]])
+        process.invoke(null, args)
+      } else {
+        val mainClass = Class.forName("scala.tools.nsc.Main", true, scalaInstance.loader)
+        val process = mainClass.getMethod("process", classOf[Array[String]])
+        process.invoke(null, arguments.toArray)
+        mainClass.getMethod("reporter").invoke(null)
+      }
+    checkForFailure(reporter, arguments.toArray)
   }
   def compilerArguments = new CompilerArguments(scalaInstance, cp)
-  protected def checkForFailure(mainClass: Class[_], args: Array[String]): Unit = {
-    val reporter = mainClass.getMethod("reporter").invoke(null)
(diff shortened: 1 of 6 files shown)


lightbend-cla-validator:
Hi @smarter,

typesafe-tools:

compile/src/main/scala/sbt/compiler/RawCompiler.scala
```diff
@@ -21,14 +21,22 @@ class RawCompiler(val scalaInstance: xsbti.compile.ScalaInstance, cp: ClasspathO
...
+        val process = mainClass.getMethod("process", classOf[Array[String]])
+        process.invoke(null, arguments.toArray)
+        mainClass.getMethod("reporter").invoke(null)
+      }
+    checkForFailure(reporter, arguments.toArray)
   }
   def compilerArguments = new CompilerArguments(scalaInstance, cp)
-  protected def checkForFailure(mainClass: Class[_], args: Array[String]): Unit = {
-    val reporter = mainClass.getMethod("reporter").invoke(null)
+  protected def checkForFailure(reporter: AnyRef, args: Array[String]): Unit = {
```
dwijnand, line 39: This is a breaking (bincompat) change.

dwijnand:
This is kind of exciting. I'm surprised it's such a small change.

compile/src/main/scala/sbt/compiler/RawCompiler.scala
smarter, line 39: OK, will fix, I have no idea what is and isn't public API :). Is it okay if I push-force my branch or do you prefer code reviews to be addressed in separate commits?

smarter:
This would have been much harder to get in 0.13 without #2197, so thank you @Duhemm  :).

compile/src/main/scala/sbt/compiler/RawCompiler.scala
dwijnand, line 39: Part of 1.0 is making that decision and enforcing it. In these 0.13 days, anything that can be accessed, is public :-/

It's not really been decided, but I'm personally happy with history-rewrites and force pushes.

SethTisue:
perhaps instead of the `isDotty` hack, an actual `SettingKey[Boolean]` called `isDotty`?

smarter:
A setting would be slightly annoying if you want your sbt build to support both Scala and Dotty, you'd have to do `isDotty := scalaVersion.value.startsWith("0.")` yourself, but I'm not opposed to it if people like this better.

smarter:
...Unless the setting default value is `scalaVersion.value.startsWith("0.")` ? I could do this but I have no idea of the proper place to put this setting in the many dimensions of sbt.

eed3si9n:
Settings generally go into `Keys.scala` and `Defaults.scala`. Somewhere around here perhaps? https://github.com/sbt/sbt/blob/v0.13.9/main/src/main/scala/sbt/Defaults.scala#L245

smarter:
Am I allowed to add a `boolean isDotty()` method to the interface `ScalaInstance`?

smarter:
If I make `isDotty` a key then we need a way to access `isDotty` from various classes like `RawCompiler`, the obvious thing to do would be to add an `isDotty` field to the class `ScalaInstance` but this is tricky to do without breaking binary compatibility (I can't change the primary constructor of the class so I'd have to add a mutable field, ...). So I'd rather keep things as-is (no `isDotty` key) for now and revisit this for `1.0` when we'll break binary compatiblity. Anything else that needs to be done before this can be merged?

eed3si9n:
> So I'd rather keep things as-is (no isDotty key) for now and revisit this for 1.0 when we'll break binary compatibility. 

That is a very good point.

> Anything else that needs to be done before this can be merged?

I think we're good here.

PR #2127, opened 2015-07-24
Add warning to MakePom for intransitive dependencies.

jsuereth:
Intransitive does not work in Maven, and does not translate to pom.xml.

Review by @eed3si9n or @dwijnand

PR #2345, opened 2015-12-31
Maven compatibility changes (intransitive warnings and "configuration not public")

### LINKED ISSUE #1827: "configuration not public" on local test->test dependency if the same version has been resolved out of a public Maven repository
# Problem

I saw this problem on [stackoverflow](http://stackoverflow.com/questions/26235997/teamcity-cant-resolve-test-artifact-from-nexus-with-configuration-not-public) too, but it didn't seem like it had been solved, and I couldn't find an issue, so I'm posting it here.

I'm trying to do development on twitter/util, but when I try to run update on util-collection, it tells me: 

```
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]  ::          UNRESOLVED DEPENDENCIES         ::
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]  :: com.twitter#util-core_2.10;6.23.0: configuration not public in com.twitter#util-core_2.10;6.23.0: 'test'. It was required from com.twitter#util-collection_2.10;6.23.0 test
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]
[warn]  Note: Unresolved dependencies path:
[warn]      com.twitter:util-core_2.10:6.23.0
[warn]        +- com.twitter:util-collection_2.10:6.23.0
[trace] Stack trace suppressed: run last util-collection/*:update for the full output.
[error] (util-collection/*:update) sbt.ResolveException: unresolved dependency: com.twitter#util-core_2.10;6.23.0: configuration not public in com.twitter#util-core_2.10;6.23.0: 'test'. It was required from com.twitter#util-collection_2.10;6.23.0 test
[error] Total time: 1 s, completed Jan 27, 2015 1:49:54 AM
```

I tried making a demo project to repro, and I found that I could only repro when I chose the same version, organization, and artifact name.

eed3si9n:
### intransitive warning

Made a minor improvement on @jsuereth's #2127, and split the handling to its own object called `CompatibilityWarning`. This is now invoked during `update` rather waiting till `makePom`, which might be too late.
### "configuration not public": Adds `withInterProjectFirst()`

Fixes #1827.

`withInterProjectFirst()` option when set to `true` will prioritize `inter-project` resolver over all other resolvers **and** the Ivy cache. This is aimed to workaround the fact that on Maven, `Test` configuration is considered private, and thus project dependency with `test->test` configuration may not under some condition: The condition is when someone resolves `x:y:1.0` and you have a subproject named and versioned exactly that, and some other subproject tries to depend on it. (182374c71226892e7712e96e464457494233f8aa)
This does happen when a project publishes to Maven and then someone grabs the code from Github, which is the case for `twitter-util`. I've run into this myself during sbt modularization, and apparently it happens with spark development too.


ivy/src/main/scala/sbt/CompatibilityWarning.scala
```diff
@@ -0,0 +1,25 @@
...
+  def processIntransitive(module: IvySbt#Module, log: Logger): Unit = {
+    val directDependencies: Seq[ModuleID] = module.moduleSettings match {
+      case x: InlineConfiguration             => x.dependencies
+      case x: InlineConfigurationWithExcludes => x.dependencies
+      case _                                  => Seq()
+    }
+    directDependencies foreach { m =>
+      if (!m.isTransitive) {
+        log.warn(
+          s"""Found intransitive dependency ($m), but Maven repositories do not support intransitive dependencies.
```
dwijnand: Shall we try and make a reference to the fact that this is dependant on `publishMavenStyle` (as it is in the notes)?

I'm just not sure if "but Maven repositories" is sufficient to understand that this is because `publishMavenStyle` is set to `true`.

dwijnand:
A wombo combo double bugfix PR, nice!

Can I exploit this PR to ask, how come we have some settings that aren't that - settings, but rather `withInterProjectFirst` or `withLatestSnapshots`?

1 minor logging comment, other than that LGTM!

ivy/src/main/scala/sbt/CompatibilityWarning.scala
eed3si9n: I can fix that.

eed3si9n:
> Can I exploit this PR to ask, how come we have some settings that aren't that - settings, but rather withInterProjectFirst or withLatestSnapshots?

That's a good question. I first saw these flags for name hashing  https://github.com/sbt/sbt/blob/0.13/compile/inc/src/main/scala/sbt/inc/IncOptions.scala#L138-L141 `withNameHashing()` and I think these are sort of meant as implementation detail of sbt itself rather than describing the build, and perhaps eventually turned on/off by default (and/or go away). For example, name hashing will be the only mechanism to do incremental compilation soon.

huitseeker:
Kudos: tested on an sbt build of spark, which exhibits this problem through sbt-pomreader. This fixes it ! :)

PR #2349, opened 2016-01-05
Upgrade Scala version in scripted tests

gkossakowski:
Both continuations and macro-config set scalaVersion explicitly but since
sbt relies now on Scala 2.10 it's not needed anymore. In particular, we
can upgrade continuations to 2.10 which makes it easier to work with Java
8.


typesafe-tools:

eed3si9n:
LGTM pending Travis

dwijnand:
LGTM and Travis passed, merging.

Issue #2348, opened 2016-01-05
Provide a Setting that points to the (eventual) config:packageBin


fommil:
Currently the `packageBin` task is the only way to get hold of the jar output from a project/config. (actually `(artifactPath in Compile in packageBin).value` would be enough but it amounts to the same).

However, it is often useful to know where the file is going to be without necessarily invoking the compilation (e.g. projects that are compiler plugins and need to be referenced by IDEs, caching strategies).

Proposed implementation:

```
  val packageBinFile = SettingKey[File](
    "packageBinFile",
    "The (eventual) location of the packageBin."
  )

  def packageBinFileSetting = Def.setting {
    val append = configuration.value match {
      case Compile => ""
      case Test    => "-tests"
      case _       => "-" + configuration.value.name
    }
    crossTarget.value / s"${projectID.value.name}_${scalaBinaryVersion.value}-${projectID.value.revision}$append.jar"
  }
```

eed3si9n:
It's `artifactPath in packageBin`.

fommil:
I don't think you read the ticket in enough detail, that forces many ivy Tasks to run, all the way down (resulting in hundreds of IO heavy tasks) which is very slow on windows.

fommil:
(BTW configuration is needed too)

fommil:
I should really from a tasktimings trace of the current approach to show what is being triggered.

eed3si9n:
Using sbt-big-project-test on Windows XP, `compile:packageBin::artifactPath` returns immediately for me even after issuing `clean` first.

fommil:
Its possible I was getting crosstalk from other tasks or settings. Did you do a TaskTimings to trace that?

eed3si9n:
> Did you do a TaskTimings to trace that?

I did it anyway, but I didn't have to because `artifactPath` is a `settingKey`, which by design does not invoke the task engine.

What you might see in the task timing if you run sbt from the shell is the compilation of the build i.e. `project/ClusterFsckBuild.scala` itself, not the processing of `compile:packageBin::artifactPath`.

fommil:
OK, cool, I'll confirm this tonight and close the ticket if that's the case. And remove my workaround.

Issue #1209, opened 2014-03-25
Provide a mechanisms to log all tasks as they are run with a simple setting enablement.


sander24:
There are a bunch of things that play all together and produce a total effect of SBT builds being VERY hard to debug for a beginner (and probably also for intermediate) user. If you want your tool to gain more popularity, this certainly needs to be addressed.

One of the obvious things that would be easy but important to improve is how SBT prints what it is actually doing.

1) My personal opinion is that executed tasks should be printed to everybody, not just the few that type in the "--debug" key. But this is debatable, of course. I would suggest rather to provide an option to _hide_ them ("--quiet"?), rather than hiding them by default. I don't know a single other build engine that wouldn't by default at least give hints about what it is actually doing. For example, this is output of a full build of my project:

```
$ sbt clean package
Loading /media/linux-data/apps/sbt/bin/sbt-launch-lib.bash
[info] Loading project definition from /media/linux-data/dev/workspace-indigo/jr-license-server/project
[info] Set current project to license-server (in build file:/media/linux-data/dev/workspace-indigo/jr-license-server/)
[success] Total time: 0 s, completed Mar 25, 2014 1:35:23 PM

[warn] warning: [options] bootstrap class path not set in conjunction with -source 1.6
[warn] 1 warning
[success] Total time: 54 s, completed Mar 25, 2014 1:36:16 PM
```

This really looks like a "very quiet" mode, not the default mode. A developer normally wants to see what was executed, at least on a very high level.

2) The more serious problem is that, in the sbt-release context, I can't seem to be able to see _at all_ which tasks actually execute, even with "--debug". I'm talking about the case where tasks are wrapped into ReleaseSteps.

```
$ sbt --debug release
Loading /media/linux-data/apps/sbt/bin/sbt-launch-lib.bash
[debug] > boot
[debug] > reload
[debug] > sbtStashOnFailure
[debug] > onFailure load-failed
[debug] > loadp
[info] Loading project definition from /media/linux-data/dev/workspace-indigo/jr-license-server/project
[debug] Running task... Cancelable: false, check cycles: false
[info] Set current project to license-server (in build file:/media/linux-data/dev/workspace-indigo/jr-license-server/)
[debug] > sbtPopOnFailure
[debug] > resumeFromFailure
[debug] > iflast shell
[debug] > release
[debug] Running task... Cancelable: false, check cycles: false
 === Starting custom ReleaseStep ===
[debug] Running task... Cancelable: false, check cycles: false
[...]

PR #2354, opened 2016-01-07
Inter-project dependency tracking

### LINKED ISSUE #2266: update resolution extremely slow in a project with no ivy dependencies
I have a large project at work and the build definition contains explicit `unmanagedJars` instead of using ivy to manage dependencies.

Typing an `update` in the `sbt` console takes about 71 seconds, on an extremely fast computer, to do nothing. `offline := true`. It takes close to 10 minutes to run `updateClassifiers` (which also does nothing). It seems like there is a lot of wastage in the way that dependencies are calculated.

It should be trivial to create a test project that demonstrates the problem.

eed3si9n:
Fixes #2266

Adds `trackInternalDependencies` and `exportToInternal` settings. These
can be used to control whether to trigger compilation of a dependent
subprojects when you call `compile`. Both keys will take one of three
values: `TrackLevel.NoTracking`, `TrackLevel.TrackIfMissing`, and
`TrackLevel.TrackAlways`. By default they are both set to
`TrackLevel.TrackAlways`.

When `trackInternalDependencies` is set to `TrackLevel.TrackIfMissing`,
sbt will no longer try to compile internal (inter-project) dependencies
automatically, unless there are no `*.class` files (or JAR file when
`exportJars` is `true`) in the output directory. When the setting is
set to `TrackLevel.NoTracking`, the compilation of internal
dependencies will be skipped. Note that the classpath will still be
appended, and dependency graph will still show them as dependencies.
The motivation is to save the I/O overhead of checking for the changes
on a build with many subprojects during development. Here's how to set
all subprojects to `TrackIfMissing`.

``` scala
      aggregate(....).
      settings(
        inThisBuild(Seq(
          trackInternalDependencies := TrackLevel.TrackIfMissing,
          exportJars := true
        ))
      )
```

The `exportToInternal` setting allows the dependee subprojects to opt
out of the internal tracking, which might be useful if you want to
track most subprojects except for a few. The intersection of the
`trackInternalDependencies` and `exportToInternal` settings will be
used to determine the actual track level. Here's an example to opt-out
one project:

``` scala
    lazy val dontTrackMe = (project in file("dontTrackMe")).
      settings(
        exportToInternal := TrackLevel.NoTracking
      )
```

/review @gkossakowski @dwijnand, @jsuereth, @Duhemm

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -99,6 +99,8 @@ object Defaults extends BuildCommon {
       internalConfigurationMap :== Configurations.internalMap _,
       credentials :== Nil,
       exportJars :== false,
+      trackInternalDependencies :== TrackLevel.TrackAlways,
+      exportToInternal :== TrackLevel.TrackAlways,
       retrieveManaged :== false,
       retrieveManagedSync :== false,
       configurationsToRetrieve :== None,
@@ -1005,6 +1007,9 @@ object Classpaths {
     unmanagedClasspath <<= unmanagedDependencies,
     managedClasspath := managedJars(classpathConfiguration.value, classpathTypes.value, update.value),
     exportedProducts <<= exportProductsTask,
+    exportedProductsAlways <<= trackedExportedProducts(TrackLevel.TrackAlways),
+    exportedProductsIfMissing <<= trackedExportedProducts(TrackLevel.TrackIfMissing),
+    exportedProductsNoTracking <<= trackedExportedProducts(TrackLevel.NoTracking),
     unmanagedJars := findUnmanagedJars(configuration.value, unmanagedBase.value, includeFilter in unmanagedJars value, excludeFilter in unmanagedJars value)
   ).map(exportClasspath)
 
@@ -1532,6 +1537,52 @@ object Classpaths {
     val x2 = copyResources.value
     classDirectory.value :: Nil
   }
+  // This is a variant of exportProductsTask with tracking
+  private[sbt] def trackedExportedProducts(track: TrackLevel): Initialize[Task[Classpath]] = Def.task {
+    val art = (artifact in packageBin).value
+    val module = projectID.value
+    val config = configuration.value
+    for { (f, analysis) <- trackedProductsImplTask(track).value } yield APIMappings.store(analyzed(f, analysis), apiURL.value).put(artifact.key, art).put(moduleID.key, module).put(configuration.key, config)
+  }
+  private[this] def trackedProductsImplTask(track: TrackLevel): Initialize[Task[Seq[(File, Analysis)]]] =
+    Def.taskDyn {
+      val useJars = exportJars.value
+      val jar = (artifactPath in packageBin).value
+      val dirs = productDirectories.value
+      def containsClassFile(fs: List[File]): Boolean =
+        (fs exists { dir =>
+          (dir ** DirectoryFilter).get exists { d =>
+            (d * "*.class").get.nonEmpty
+          }
+        })
+      TrackLevel.intersection(track, exportToInternal.value) match {
+        case TrackLevel.TrackAlways if (useJars) =>
+          Def.task {
+            Seq((packageBin.value, compile.value))
+          }
+        case TrackLevel.TrackAlways =>
+          Def.task {
+            products.value map { (_, compile.value) }
+          }
+        case TrackLevel.TrackIfMissing if (useJars && !jar.exists) =>
+          Def.task {
+            Seq((packageBin.value, compile.value))
+          }
+        case TrackLevel.TrackIfMissing if (!useJars && !containsClassFile(dirs.toList)) =>
+          Def.task {
+            products.value map { (_, compile.value) }
+          }
+        case _ =>
+          Def.task {
+            val analysis = previousCompile.value.analysis
+            (if (useJars) Seq(jar)
+            else dirs) map {
+              (_, analysis)
+            }
+          }
+      }
+    }
+
   def exportProductsTask: Initialize[Task[Classpath]] = Def.task {
     val art = (artifact in packageBin).value
     val module = projectID.value
@@ -1547,7 +1598,7 @@ object Classpaths {
   def constructBuildDependencies: Initialize[BuildDependencies] = loadedBuild(lb => BuildUtil.dependencies(lb.units))
 
   def internalDependencies: Initialize[Task[Classpath]] =
-    (thisProjectRef, classpathConfiguration, configuration, settingsData, buildDependencies) flatMap internalDependencies0
+    (thisProjectRef, classpathConfiguration, configuration, settingsData, buildDependencies, trackInternalDependencies) flatMap internalDependencies0
   def unmanagedDependencies: Initialize[Task[Classpath]] =
     (thisProjectRef, configuration, settingsData, buildDependencies) flatMap unmanagedDependencies0
   def mkIvyConfiguration: Initialize[Task[IvyConfiguration]] =
@@ -1585,17 +1636,25 @@ object Classpaths {
       visited.toSeq
     }
   def unmanagedDependencies0(projectRef: ProjectRef, conf: Configuration, data: Settings[Scope], deps: BuildDependencies): Task[Classpath] =
-    interDependencies(projectRef, deps, conf, conf, data, true, unmanagedLibs)
+    interDependencies(projectRef, deps, conf, conf, data, TrackLevel.TrackAlways, true, unmanagedLibs0)
+  @deprecated("This is no longer public.", "0.13.10")
   def internalDependencies0(projectRef: ProjectRef, conf: Configuration, self: Configuration, data: Settings[Scope], deps: BuildDependencies): Task[Classpath] =
     interDependencies(projectRef, deps, conf, self, data, false, productsTask)
+  private[sbt] def internalDependencies0(projectRef: ProjectRef, conf: Configuration, self: Configuration, data: Settings[Scope], deps: BuildDependencies, track: TrackLevel): Task[Classpath] =
+    interDependencies(projectRef, deps, conf, self, data, track, false, productsTask0)
+  @deprecated("This is no longer public.", "0.13.10")
   def interDependencies(projectRef: ProjectRef, deps: BuildDependencies, conf: Configuration, self: Configuration, data: Settings[Scope], includeSelf: Boolean,
     f: (ProjectRef, String, Settings[Scope]) => Task[Classpath]): Task[Classpath] =
+    interDependencies(projectRef, deps, conf, self, data, TrackLevel.TrackAlways, includeSelf,
+      { (pr: ProjectRef, s: String, sc: Settings[Scope], tl: TrackLevel) => f(pr, s, sc) })
+  private[sbt] def interDependencies(projectRef: ProjectRef, deps: BuildDependencies, conf: Configuration, self: Configuration, data: Settings[Scope],
+    track: TrackLevel, includeSelf: Boolean, f: (ProjectRef, String, Settings[Scope], TrackLevel) => Task[Classpath]): Task[Classpath] =
     {
       val visited = interSort(projectRef, conf, data, deps)
       val tasks = asScalaSet(new LinkedHashSet[Task[Classpath]])
       for ((dep, c) <- visited)
         if (includeSelf || (dep != projectRef) || (conf.name != c && self.name != c))
-          tasks += f(dep, c, data)
+          tasks += f(dep, c, data, track)
 
       (tasks.toSeq.join).map(_.flatten.distinct)
     }
@@ -1639,6 +1698,14 @@ object Classpaths {
     configurations.find(_.name == conf)
   def productsTask(dep: ResolvedReference, conf: String, data: Settings[Scope]): Task[Classpath] =
     getClasspath(exportedProducts, dep, conf, data)
+  def productsTask0(dep: ResolvedReference, conf: String, data: Settings[Scope], track: TrackLevel): Task[Classpath] =
+    track match {
+      case TrackLevel.NoTracking     => getClasspath(exportedProductsNoTracking, dep, conf, data)
+      case TrackLevel.TrackIfMissing => getClasspath(exportedProductsIfMissing, dep, conf, data)
+      case TrackLevel.TrackAlways    => getClasspath(exportedProductsAlways, dep, conf, data)
+    }
+  private[sbt] def unmanagedLibs0(dep: ResolvedReference, conf: String, data: Settings[Scope], track: TrackLevel): Task[Classpath] =
+    unmanagedLibs(dep, conf, data)
   def unmanagedLibs(dep: ResolvedReference, conf: String, data: Settings[Scope]): Task[Classpath] =
     getClasspath(unmanagedJars, dep, conf, data)
   def getClasspath(key: TaskKey[Classpath], dep: ResolvedReference, conf: String, data: Settings[Scope]): Task[Classpath] =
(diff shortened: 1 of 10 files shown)


dwijnand:
I'll try and have a look when I have a bit more time to dig into it, but I thought perhaps we could cc in @fommil, given he's been spending time recently with this problem in parallel, and he opened the original issue.

Also seems codacy isn't happy.

fommil:
@dwijnand thanks, I have this on my todo list today to investigate / understand. Been wasting the whole day trying to build a Windows 7 image to test on.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1569,6 +1574,46 @@ object Classpaths {
...
+  private[sbt] def trackedExportedProducts(track: TrackLevel): Initialize[Task[Classpath]] = Def.task {
+    val art = (artifact in packageBin).value
+    val module = projectID.value
+    val config = configuration.value
+    for { (f, analysis) <- trackedProductsImplTask(track).value } yield APIMappings.store(analyzed(f, analysis), apiURL.value).put(artifact.key, art).put(moduleID.key, module).put(configuration.key, config)
+  }
+  private[this] def trackedProductsImplTask(track: TrackLevel): Initialize[Task[Seq[(File, Analysis)]]] =
+    Def.taskDyn {
+      val useJars = exportJars.value
+      val jar = (artifactPath in packageBin).value
```
fommil, line 1587: this is still going to kill the performance, as it invokes quite a lot of subtasks. Hence https://github.com/sbt/sbt/issues/2348

I'll do some graph timings with my patch to see what is invoked, but my suspicion is that this isn't going to make a massive difference to performance on Windows due to the large number of Tasks that will be created :cry:

fommil:
I think I've been reading too much of the sbt source code because I actually understand this :fearful: 

Conceptually it looks sound, but I can't be sure about which tasks are being scheduled until I run this on my test projects (I'll turn off sbt-big-project settings). I'll report back when I do that.

fommil:
Running this on my simple test project https://github.com/fommil/sbt-big-project/tree/master/src/sbt-test/sbt-big-project/simple reconditioned to look like

```
object SimpleBuild extends Build {
  override lazy val settings = super.settings ++ Seq(
    scalaVersion := "2.10.6",
    version := "v1",
    trackInternalDependencies := TrackLevel.TrackIfMissing,
    exportJars := true,
    concurrentRestrictions in Global := Seq(Tags.limitAll(1))
  )
  def simpleProject(name: String): Project = {
    BigProjectTestSupport.createSources(name)
    Project(name, file(name)).settings(
      //BigProjectPlugin.overrideProjectSettings(Compile, Test),
      BigProjectTestSupport.testInstrumentation(Compile, Test)
    )
  }
  val a = simpleProject("a")
  val b = simpleProject("b") dependsOn(a)
  val c = simpleProject("c") dependsOn(b)
  val d = simpleProject("d") dependsOn(c)
}
```

I'm still seeing a **lot** of dependent tasks in subprojects running. This doesn't look so bad for a small project, but for huge projects this is a lot of tasks. In particular, this is causing 32 tasks to run per dependent project, which is not scalable, especially for builds with 100+ projects.

Below is the output of https://github.com/sbt/sbt/issues/1209#issuecomment-168697975 with a filter to only show the tasks for the `a` subproject.

```
     16:  a/*:update: 152.999064 ms
     26:  a/compile:exportedProductsIfMissing: 94.152834 ms
     37:  a/compile:unmanagedJars: 75.065294 ms
     47:  a/*:projectDescriptors: 62.375231 ms
     54:  a/*:ivySbt: 47.900470999999996 ms
     55:  a/*:ivyConfiguration: 47.027559 ms
     58:  a/compile:internalDependencyClasspath: 43.097148 ms
     60:  a/*:ivyModule: 41.560179999999995 ms
     61:  a/compile:dependencyClasspath: 40.824545 ms
     63:  a/*:fullResolvers: 39.766501 ms
     64:  a/*:projectResolver: 39.411094999999996 ms
     67:  a/compile:externalDependencyClasspath: 36.671284 ms
     68:  a/*:moduleSettings: 36.525515 ms
     77:  a/*:allDependencies: 29.765629999999998 ms
     80:  a/compile:managedClasspath: 28.869135999999997 ms
     86:  a/*:update::unresolvedWarningConfiguration: 26.312904 ms
     87:  a/*:transitiveUpdate: 26.202226999999997 ms
     88:  a/compile:readAnalysis: 24.143079 ms
     91:  a/compile:unmanagedClasspath: 23.806255999999998 ms
     95:  a/*:projectDependencies: 23.214388 ms
     96:  a/compile:incCompileSetup: 23.113008999999998 ms
    103:  a/*:dependencyPositions: 20.61122 ms
    107:  a/*:credentials: 19.892084999999998 ms
    113:  a/compile:classpathConfiguration: 11.159374999999999 ms
    115:  a/compile:update: 10.568942999999999 ms
    128:  a/*:externalResolvers: 6.731135999999999 ms
    131:  a/*:bootResolvers: 6.668432999999999 ms
    132:  a/*:dependencyCacheDirectory: 6.656975999999999 ms
    133:  a/*:updateCacheName: 6.598299 ms
    139:  a/*:incOptions: 4.222926999999999 ms
    140:  a/compile:compileAnalysisFilename: 4.116671999999999 ms
    141:  a/compile:productDirectories: 0.990143 ms
```

and if I enable caching `updateOptions := updateOptions.value.withCachedResolution(true)` I still get these tasks.

So I honestly can't say if this improves things or not, sorry!

However, combined with sbt-big-project, this might actually improve the workflow because I've still got a few unsolved problems e.g. https://github.com/fommil/sbt-big-project/issues/11 (I'll be working on that actively, so you might want to check back)

fommil:
this might still be useful for when sbt is restarted. Consider the case where the sbt-big-project caches are all empty, so everything is going to get invoked.

Without this project tracking support, it potentially means that all projects will be recompiled. There are two things that can trigger this:

1. incremental compiler is wrong and makes a change. This is catastrophic if it happens in a low-level library.
2. the user has copied "snapshot" jars into place of the packageBins (e.g. jars obtained from some CI build for a given git hash). I potentially have the option to do this and it would dramatically speed up and simplify my workflow if it works (we're talking saving 30 minutes of setup time to cut a new dev branch).

Will this new tracking support take 2 into account? If `compile:compile` is not being called, then my logic to delete the packageBins on compilation changes won't be triggered.

Unfortunately this is extremely hard to test... does `scripted` allow for multiple sessions? (or would a `reload` be enough to trigger this?)

fommil:
(just thinking out loud... perhaps it is possible to test the behaviour for 2. It just means touching dummy .jar packageBin files as soon as the sbt session begins, to simulate snapshot jars being put there)

This certainly deserves further research. I'll try to write those tests in the next few days.

main/src/main/scala/sbt/Defaults.scala
eed3si9n, line 1587: I can certainly make lighter-weight setting or task for jar file if that's going to trigger things. I just reopened #2348.

fommil, line 1587: Thanks. Btw, I don't think this is the only thing triggering lots of upstream Tasks.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1569,6 +1574,46 @@ object Classpaths {
...
+      val dirs = productDirectories.value
+      TrackLevel.intersection(track, exportToInternal.value) match {
+        case TrackLevel.TrackAlways if (useJars) =>
+          Def.task {
+            Seq((packageBin.value, compile.value))
+          }
+        case TrackLevel.TrackAlways =>
+          Def.task {
+            products.value map { (_, compile.value) }
```
gkossakowski, line 1602: Won't this trigger multiple calls to `compile` task, one for each product?

fommil:
Sorry, I meant to look into this tonight but ran out of time.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1569,6 +1574,46 @@ object Classpaths {
...
+          }
+        case TrackLevel.TrackIfMissing if (useJars && !jar.exists) =>
+          Def.task {
+            Seq((packageBin.value, compile.value))
+          }
+        case TrackLevel.TrackIfMissing if (!useJars && (dirs flatMap { d => (d ** "*.class").get }).nonEmpty) =>
```
gkossakowski: In case you have lots of class files (which is often the case), you'll be creating rather large collection just to check if it's non empty. Wouldn't it be better to implement a custom file system traverser that return eagerly the answer?

main/src/main/scala/sbt/Defaults.scala
eed3si9n, line 1602: Within a dynamic task (`Def.taskDyn`) the last task value is deferred.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Probably better, but I was lazy to implement one.

gkossakowski:
Could you give a context when is this really needed? Is it trying to workaround issues with NFS file systems?

What incremental compiler is doing is equivalent to what `git status` is doing. For `git status` it takes 0.05s to check status of 40k files on my machine (MBP 2015). I'm guessing performance scales linearly so it would take 0.5s to check status of 400k or about 1.2s for 1m files. Does sbt need to support larger projects than with 1m files?

main/src/main/scala/sbt/Defaults.scala
gkossakowski, line 1602: Ok, good to know.

eed3si9n:
> Could you give a context when is this really needed? Is it trying to workaround issues with NFS file systems?

Primarily yes. More details are found in #2266 and https://github.com/eed3si9n/sbt-big-project-test, but on Windows 100+ subproject build seem to take 27s just running `compile` without any source change.
I was also thinking that this feature can potentially become an extension point down the line for further optimization like binary/source switching.

eed3si9n:
Here's the timing breakdown that `sbt.task.timings` collected - https://docs.google.com/spreadsheets/d/1bb3mAo0mC2LqXaomiNVs0uKBlwMG1vPActzTUSKtWjQ/edit#gid=166341631

main/src/main/scala/sbt/Defaults.scala
dwijnand: A custom file system traverser? Why not just the following?

``` scala
dirs.iterator.flatMap(d => (d ** "*.class").get).nonEmpty
```

eed3si9n: `dirs` typically contain one element, so making that lazy is not going to make things faster I think, but I could take it further and make an iterator of all subdirectories and then run `* "*.class"` on it.

dwijnand: There might be speed improvements, sure.

I was just thinking (off of Grzegorz's comment) about avoiding unnecessary allocation.

dwijnand: Sorry, understood your comment better now. I understand.

eed3si9n: Added 02c72392040cb3aef57fb49e7fa8e5864ced18c8. Interestingly, there's a slight performance penalty in terms of time when I break the iteration apart into directories and then `*.class`. Adds around 25% more time when tested with `new File(".")`, traversing all directories under sbt projects. For more realistic test traversing `new File("util/io/target/classes")`, the penalty remained: 3ms vs 2ms using the original implementation. _Hopefully_, it's doing better in memory perf, but I'm not too sure about that either without profiling.

``` scala
import sbt._
import java.io.File

def time[R](block: => R): R = {
  val t0 = System.currentTimeMillis
  val result = block
  val t1 = System.currentTimeMillis
  println("Elapsed time: " + (t1 - t0) + "ms")
  result
}

def imple1(dirs: Seq[File]) = (dirs flatMap { d => (d ** "*.class").get }).nonEmpty

def containsClassFile(fs: List[File]): Boolean =
  (fs exists { dir =>
    (dir ** DirectoryFilter).get exists { d =>
      (d * "*.class").get.nonEmpty
    }
  })

def imple3(fs: List[File]): Boolean =
  (fs exists { dir =>
    (dir ** DirectoryFilter).get.iterator exists { d =>
      (d * "*.class").get.nonEmpty
    }
  })
```

dwijnand: Interesting. On what filesystem were you testing?

gkossakowski:
Thanks for sharing details. From incremental compiler point of view this looks good to me. I'll defer review of sbt task engine performance to other people, though.

fommil:
I'm hoping to review this commit from a functional PoV tonight.

BTW, any chance you could create a single commit that cleanly applies to 0.13.9? That would be useful for a custom build I use at work.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: I did the above testing using my MBP with SSD.

fommil:
This is what I tried, using the [`simple` project](https://github.com/fommil/sbt-big-project/tree/master/src/sbt-test/sbt-big-project/simple) edited to not use the `BigProjectSettings`, and to use `TrackIfMissing`

``` scala
import sbt._
import Keys._
import Def.Initialize
import fommil._
object SimpleBuild extends Build {
  override lazy val settings = super.settings ++ Seq(
    scalaVersion := "2.10.6",
    version := "v1",
  )
  def simpleProject(name: String): Project = {
    BigProjectTestSupport.createSources(name)
    Project(name, file(name)).settings(
       trackInternalDependencies := TrackLevel.TrackIfMissing,
       exportJars := true
    )
  }
  val a = simpleProject("a")
  val b = simpleProject("b") dependsOn(a)
  val c = simpleProject("c") dependsOn(b)
  val d = simpleProject("d") dependsOn(c)
}
```
1. `d/compile`
2. check that another `d/compile` doesn't invoke `a/compile:compile`, `b/compile:compile` or `c/compile:compile`
3. note that `compile` does force compiles of everything. Interesting (probably because I added it for each project, not the `settings`, so the root project doesn't get the setting)
4. exit sbt, start sbt
5. check that `d/compile` doesn't invoke `a/compile:compile`, etc
6. make a trivial change to something under `b`
7. `d/compile` doesn't recompile `b` (this is expected behaviour, "desired" is a strong word but it is the trade off for this workflow. Hopefully I will address this is `sbt-big-project`)
8. delete `./b/target/scala-2.10/b_2.10-v1.jar`
9. `d/compile` does recompile `b` (nice! desired behaviour)
10. `rm -f b/target/scala-2.10/classes/b/*`
11. `d/compile` does not recompile `b` (very much desired behaviour, believe it or not!)
12. `b/compile` does recompile `b` (desired behaviour)
13. `b/compile` does attempt to recompile `b` (desired behaviour)

so, basically, fantastic work!

The one thing that doesn't work quite as expected is this:
1. d/compile
2. trivial change to b
3. b/compile
4. d/compile or d/runMain => uses _old_ version of b.jar

this is something I've been struggling with in sbt-big-project even without these changes, and I think the solution is to do something like this https://github.com/fommil/sbt-big-project/blob/c3d4bc3260ac5e25f63bdb94a7e0bd5c3ada6a7d/src/main/scala/BigProjectSettings.scala#L100-L117 (i.e. delete the jar if the compile produced anything). There are a couple of other sbt / zinc problems deep inside here.

AFAIK, this doesn't address the fundamental performance problems with sbt on large projects on Windows, but it gives me the levers that I need in order to implement sensible workarounds in `sbt-big-project`. Great work!

I probably need to go back to the drawing board with sbt-big-project because I don't necessarily need to cache everything that I'm caching, so I should find what the remaining bottlenecks are. I might not need to cache around `packageBin` anymore, for example.

Can you please squash and create a commit that cleanly applies against 0.13.9? (I get conflicts)

dwijnand:
Thanks @fommil for trying this out.

eed3si9n:
@fommil Thanks for the review.

> Can you please squash and create a commit that cleanly applies against 0.13.9? (I get conflicts)

See 98b26ae5d3009cbfd5a062cae108871d3ad3e466. I've squashed and backported the commits to 0.13.9, and merged that forward to 0.13.

fommil:
FYI, I'm formalising the above test scripted with a `scripted` test in `sbt-big-project`, including some aspects of the workflow that `sbt-big-project` expects (such as `b/compile` deleting `b`'s `packageBin` and `d/runMain` needing to use the latest jar)

PR #2374, opened 2016-01-16
FPORT: Fixes #2302. Don't check pre and post conditions in sbt.Execute by default.

### LINKED ISSUE #2302: sbt.Execute pre and post condition checking has bad run time characteristics for big graphs

dwijnand:
Forward-port of #2303.

The checking code has bad run time characteristics and would need to be fixed
for large projects with deep task dependency chains.

The code in sbt.Execute has been in production for a long time so it seems safe
enough to drop the extra checks by default. To debug issues, you can set
`-Dsbt.execute.extrachecks=true` to revert to the old behavior.


typesafe-tools:

PR #2379, opened 2016-01-16
FPORT: Release notes for name hashing private members fix

dwijnand:
Forward-port of #2339.


typesafe-tools:

PR #2338, opened 2015-12-29
Include sbt-ivy-snapshots to launchconfig

eed3si9n:
## what this changes

both `typesafe-ivy-releases` and `sbt-ivy-snapshots` repos are always included into the launchconfig.
## notes

Prior to this change, the launchconfig for sbt was generated using Transform's resource generator so releases would include `typesafe-ivy-releases` repo, and snapshot releases would include `typesafe-ivy-snapshots` and `sonatype-snapshots` repos in addition. This is no longer useful since nightly builds are now published to `sbt-ivy-snapshots` instead, which is backed by Bintray.

This removes the switching logic, so both `typesafe-ivy-releases` and `sbt-ivy-snapshots` repos are always included into the launchconfig. So it's no longer necessary to download a launcher just to try the nightly builds.

/review @dwijnand, @Duhemm, @jsuereth

PR #2380, opened 2016-01-16
FPORT: Include sbt-ivy-snapshots to launchconfig

dwijnand:
Forward-port of #2338.

Include sbt-ivy-snapshots to launchconfig
Prior to this change, the launchconfig for sbt was generated using
Transform's resource generator so releases would include
`typesafe-ivy-releases` repo, and snapshot releases would include
`typesafe-ivy-snapshots` and `sonatype-snapshots` repos in addition.
This is no longer useful since nightly builds are now published to
`sbt-ivy-snapshots` instead, which is backed by Bintray.

This removes the switching logic, so both `typesafe-ivy-releases` and
`sbt-ivy-snapshots` repos are always included into the launchconfig.
This removes the step of needing to download a launcher just to try the
nightly builds.

### DIFF
--- a/launch/src/main/input_resources/sbt/sbt.boot.properties
+++ b/launch/src/main/input_resources/sbt/sbt.boot.properties
@@ -13,9 +13,9 @@
 [repositories]
   local
   jcenter: https://jcenter.bintray.com/
-${{repositories}}
   maven-central
-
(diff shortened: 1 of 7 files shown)


typesafe-tools:

PR #2381, opened 2016-01-16
Add sbt-mima. Fixes #2037.

### LINKED ISSUE #2037: Consider using sbt-mima to mechanically verifying we're not breaking bincompat.
I had a look and didn't see this mentioned in any issue, pull request or mailing list thread.

Thoughts?

dwijnand:


typesafe-tools:

eed3si9n:
This is good, but since sbt 1.0 will be an aggregate of modules, we should MIMA check the publicly exposed types at the module level.

dwijnand:
Do you mean here somehow, or do you mean add mima to the modules? Either way, I agree.

eed3si9n:
We need to catch API breakage at the PR validation of the modules.

dwijnand:
Ok. I'll create a meta-ticket for the various modules.

PR #2387, opened 2016-01-17
Update README with 1.0.x

dwijnand:


typesafe-tools:

PR #2392, opened 2016-01-17
FPORT: Upgrade Scala version in scripted tests

dwijnand:
Forward port of #2349.


typesafe-tools:

PR #2393, opened 2016-01-17
FPORT: Make sbt aware of Dotty

dwijnand:
This small set of changes, together with the compiler-bridge I wrote
(https://github.com/smarter/dotty-bridge) enables us to compile code
using Dotty in sbt, see https://github.com/smarter/dotty-example-project
for an example.

Partial forward port of #2344.


typesafe-tools:

PR #2397, opened 2016-01-17
FPORT: Consider signatures of methods before and after erasure in ExtractAPI

dwijnand:
Forward port of #2261.


typesafe-tools:

dwijnand:
The scripted part of this seems to have gone in with:
- https://github.com/sbt/incrementalcompiler/pull/41, and
- https://github.com/sbt/incrementalcompiler/pull/65

The notes, I agree, we can diff/copy-paste post everything.

PR #2407, opened 2016-01-22
Adds CompatibilityWarningOptions

### LINKED ISSUE #2347: regression in 0.13.10-SNAPSHOT maven style publishing
I don't get this in 0.13.9.

We recently added a new `Configuration` to the `ensime-sbt` plugin https://github.com/ensime/ensime-sbt/blob/5f18b5aac2b6bd30dc829414e782e6ce3461d5c2/src/main/scala/EnsimePlugin.scala#L98

It is marked as `hide` and shouldn't appear anywhere in the project's `pom`.

However, when running a really simple project https://github.com/fommil/sbt-big-project/tree/master/src/sbt-test/sbt-big-project/simple with the new plugin, I see this

```
[warn] Found intransitive dependency (org.scala-lang:scala-compiler:2.10.6:ensime-internal) while publishMavenStyle is true, but Maven repositories
[warn]   do not support intransitive dependencies. Use exclusions instead so transitive dependencies
[warn]   will be correctly excluded in dependent projects.
[warn]            
```

Basically, this shouldn't even be a consideration so the warning shouldn't appear.

eed3si9n:
Fixes #2347

This addresses 0.13.10 regression, which currently warns users about
Maven incompatibility on a private configuration. This adds a config
class so the build user can control the level of the warning as well as
the target configuration to be monitored.
By default, we are only going to look at `Compile` and `Runtime`.

/review @dwijnand, @Duhemm

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1104,6 +1104,7 @@ object Classpaths {
 
   private[this] def baseGlobalDefaults = Defaults.globalDefaults(Seq(
     conflictWarning :== ConflictWarning.default("global"),
+    compatibilityWarningOptions :== CompatibilityWarningOptions.default,
     homepage :== None,
     startYear :== None,
     licenses :== Nil,
@@ -1397,6 +1398,7 @@ object Classpaths {
     val depDir = dependencyCacheDirectory.value
     val uc0 = updateConfiguration.value
     val ms = publishMavenStyle.value
+    val cw = compatibilityWarningOptions.value
     // Normally, log would capture log messages at all levels.
     // Ivy logs are treated specially using sbt.UpdateConfiguration.logging.
     // This code bumps up the sbt.UpdateConfiguration.logging to Full when logLevel is Debug.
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1412,17 +1414,18 @@ object Classpaths {
...
       uwConfig = uwConfig, logicalClock = logicalClock, depDir = Some(depDir),
-      ewo = ewo, mavenStyle = ms, log = s.log)
+      ewo = ewo, mavenStyle = ms, compatWarning = cw, log = s.log)
   }
   @deprecated("Use cachedUpdate with the variant that takes unresolvedHandler instead.", "0.13.6")
   def cachedUpdate(cacheFile: File, label: String, module: IvySbt#Module, config: UpdateConfiguration,
     transform: UpdateReport => UpdateReport, skip: Boolean, force: Boolean, depsUpdated: Boolean, log: Logger): UpdateReport =
     cachedUpdate(cacheFile, label, module, config, transform, skip, force, depsUpdated,
-      UnresolvedWarningConfiguration(), LogicalClock.unknown, None, EvictionWarningOptions.empty, true, log)
+      UnresolvedWarningConfiguration(), LogicalClock.unknown, None, EvictionWarningOptions.empty, true, CompatibilityWarningOptions.default, log)
```
dwijnand, line 1423: Is this intentionally not using the configured `compatibilityWarningOptions` value?

eed3si9n, line 1423: `cachedUpdate` is a method, not a task, so I don't know what you configured it for. This method is just for bincompat.

dwijnand, line 1423: I realise the implementation impacts. Ok, thanks, I'm happy to leave this as such.

dwijnand:
Do we want to explain this new configuration option in the notes?

Other than that LGTM.

Please also pull request against 1.0.x :)

eed3si9n:
It's not something that I foresee the build user customizing often, so I'm ok with leaving it out of the release notes on how to configure it.

PR #2410, opened 2016-01-23
Update version to 0.13.11-SNAPSHOT

dwijnand:


typesafe-tools:

dwijnand:
As we're still stabilizing 0.13.10 and we have a precendent of a 0.13.10 fix being merged to 0.13 and then 0.13 being merged back to 0.13.10 (#2418), I'm going to hold this change for a little bit.

PR #2413, opened 2016-01-23
[0.13.10] Always invalidate API when return type is a value class

smarter:
Before this commit, we did not do the invalidation for methods with
multiple parameter list, the comment above `hasValueClassAsReturnType`
said:

> Note: We only inspect the "outermost type" (i.e. no recursion) because
>   we don't need to inspect after erasure a function that would, for
>   instance, return a function that returns a subtype of AnyVal.

But this is wrong: a method with signature:

``` scala
  def foo(a: A)(b: B): C
```

is erased to:

``` scala
  def foo(a: A, b: B): C
```

and not, as the comment in the code suggest, to:

``` scala
  def foo(a: A): B => C
```

so we do need to inspect the final result type of methods, because they
can be value classes that will be erased to their underlying value.

/cc @Duhemm

### DIFF
--- a/compile/interface/src/main/scala/xsbt/ExtractAPI.scala
+++ b/compile/interface/src/main/scala/xsbt/ExtractAPI.scala
@@ -208,20 +208,14 @@ class ExtractAPI[GlobalType <: CallbackGlobal](val global: GlobalType,
         s.asMethod.paramss.flatten map (_.info) exists (t => isAnyValSubtype(t.typeSymbol))
       }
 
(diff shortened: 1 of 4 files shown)


typesafe-tools:

dwijnand:
:+1:

Also requires a forward port to sbt/incrementalcompiler

smarter:
Rebased.

eed3si9n:
LGTM pending Travis

PR #2414, opened 2016-01-23
[0.13.10] ExtractAPI: avoid unnecessary duplication of defs with primitive types

smarter:
NOTE: This PR is based on top of https://github.com/sbt/sbt/pull/2413 for convenience because they conflict with each other, but they could be made independent.

<hr>

If a method's type contains a non-primitive value class then it has two
signatures: one before erasure and one after erasure. Before this
commit, we checked if this was the case using `isAnyValSubtype`, but
this is too crude since primitive value classes are also subtypes of
`AnyVal` but do not change signature after erasure.

This commit replaces `isAnyValSubtype` by `isDerivedValueClass` which
excludes primitive value classes.

In practice, for an empty class, this reduces the size of the output of
`DefaultShowAPI` from 65 lines to 25 lines.
Before:
https://gist.github.com/smarter/cf1d6fe58efda88d6ee6#file-old-api
After:
https://gist.github.com/smarter/cf1d6fe58efda88d6ee6#file-new-api

/cc @Duhemm

### DIFF
--- a/compile/interface/src/main/scala/xsbt/Compat.scala
+++ b/compile/interface/src/main/scala/xsbt/Compat.scala
@@ -107,8 +107,8 @@ abstract class Compat {
     }
     lazy val AnyValClass = global.rootMirror.getClassIfDefined("scala.AnyVal")
 
-    def isAnyValSubtype(sym: Symbol): Boolean = sym.isNonBottomSubClass(AnyValClass)
-
+    def isDerivedValueClass(sym: Symbol): Boolean =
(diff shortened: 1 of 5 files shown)


typesafe-tools:

smarter:
Rebased.

eed3si9n:
LGTM pending Travis

dwijnand:
Scripted test "tests / nested-subproc" failed. Starting the "tests" suite again.

Duhemm:
Thanks for the fixes!

PR #2399, opened 2016-01-18
bugfix - apply autoImports for global plugins at global configuration stage

timcharper:

PR #2419, opened 2016-01-25
Fixes #2415. Hide global setting fix behind sbt.globalsettingfix

### LINKED ISSUE #2415: [0.13.10] Source compatibility breakage due to ambiguous imports
Ref #2399

@fommil wrote in https://gitter.im/sbt/sbt?at=56a3c957586242210adf1d87

> is anybody seeing this with 0.13.10-RC1?
> 
> ``` scala
>  /home/fommil/.sbt/0.13/global.sbt:8: error: reference to useGpgAgent is ambiguous;
> it is imported twice in the same scope by
> import com.typesafe.sbt.pgp.PgpKeys._
> and import _root_.com.typesafe.sbt.SbtPgp.autoImport._
> useGpgAgent := true
> ```
> 
> is there now an implicit dependency on the pgp plugin?

Us importing `autoImport._` might be breaking source compatibility.

"not doing a wildcard import" is apparently a workaround, but this is a sticky situation because global settings are used across all 0.13 builds. Maybe we need to hide this feature behind a flag?

 /cc @timcharper

eed3si9n:
We are hiding a bug fix on global setting that was not importing auto
imports because fixing this via sbt/sbt#2399 breaks the source
compatibility: sbt/sbt#2415
I've split out the relevant portion of the scripted test into
global-settings, and marked it pending.

/cc @timcharper, @dwijnand


dwijnand:
Thinking about this more I'm now in agreement with a feature flag. :+1:

Even if it's just for the remainder of 0.13, can we find a less opaque name than `sbt.globalsettingfix`? Not sure but perhaps something like `sbt.enable.autoimports.in.global`?

Also we should tweaks notes so people know can learn how to benefit from this fix.

eed3si9n:
> Not sure but perhaps something like sbt.enable.autoimports.in.global?

Using dot (`.`) delimited English phrase I don't think is consistent with other flags we do have, which are partially documented to be fair in http://www.scala-sbt.org/0.13/docs/Command-Line-Reference.html and peppered around the code via `sys.props` and `Boolean.getBoolean`.

As per tweaking notes, could we do it after the merge? I'm about to get on a flight, and if possible I'd like to get RC2 sooner without involving Travis waiting.

dwijnand:
I'm happy to tweak both the words and the formatting, just looking for something a little more informative. `sbt.global.autoimports` to match `sbt.global.base`?

Sure we can do notes later, let's just remember :) Have a safe flight and enjoy the conference.

eed3si9n:
@dwijnand Added a commit from :airplane:.

dwijnand:
Ugh, I'd merge if I could. Punching Travis into going green.

PR #2423, opened 2016-01-26
Use scripted engine from sbt/util

Duhemm:


typesafe-tools:

dwijnand:
LGTM in spirit and Travis seems happy with it. :+1:

PR #2430, opened 2016-01-31
Update bridge retrieval in `consoleProject`

Duhemm:
Fixes sbt/sbt#2428

### DIFF
--- a/main/src/main/scala/sbt/ConsoleProject.scala
+++ b/main/src/main/scala/sbt/ConsoleProject.scala
@@ -11,7 +11,13 @@ object ConsoleProject {
     val cpImports = new Imports(extracted, state)
     val bindings = ("currentState" -> state) :: ("extracted" -> extracted) :: ("cpHelpers" -> cpImports) :: Nil
     val unit = extracted.currentUnit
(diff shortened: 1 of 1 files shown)


typesafe-tools:

dwijnand:
I'm guessing there's no way to write a test for this?

dwijnand:
maybe a scripted test with `initialCommands` set to `sys exit 0` or something?

Duhemm:
Actually there already are tests for `consoleProject`: https://github.com/sbt/sbt/tree/0.13/sbt/src/sbt-test/project/console

I don't know how they passed...

dwijnand:
huh "This can't actually fail, unfortunately.  The repl traps the exception and continues normally." :)

Duhemm:
Hmmm I think this comment applies only to line 8 of the test. The exception thrown by the failed assertion  will be caught by the Scala REPL which will then continue normally. sbt will probably not even notice the exception.

However, a failure in the previous commands should be caught by sbt...

dwijnand:
This isn't a blocker to this being merged of course. I've just been thinking more about making my pull requests include tests to avoid my addition/change regressing. If you can find a way that we can catch this, that would be great. But even otherwise LGTM.

Issue #1104, opened 2014-02-06
Class-based dependency tracking in name hashing algorithm


gkossakowski:
This is a meta-issue that provides quick overview of what needs to be done in order to change dependency tracking from source file-based to class name-based.
### Tasks

Class-based dependency tracking can be implemented in two phases. The third phase is dedicated to testing on large projects.
#### Invalidate by inheritance based on class dependencies (phase 1)

~~_Tasks marked as completed are completed only in a prototype implementation. No changes has been merged to sbt yet._~~
- [x] add tracking of declared classes in source files; we want to invalidate classes but at the end we need className -> srcFile relation because we can recompile source files only
- [x] add dependency tracking of top-level (not owned by any other class) imports; we'll assign dependencies introduced by those imports to the first top level class/object/trait declared in the source file (see notes below for details)
- [x] track dependencies at class level (and use `declaredClasses` to immediately map them back to files until the rest of the algorithm is implemented)
- [x] add tracking of children of sealed parents; we should track this in `API` object corresponding to sealed parent; this will enable us to perform proper invalidation when a new case (class) introduced to a sealed hierarchy
- [x] add per (top level) class tracking of api hashes; tracking of api hashes per each (even nested) class separately will be done in phase 2
- [x] handle dependencies coming from local and anonymous classes (see discussion below)
- [x] switch invalidation of dependencies by inheritance to invalidate at class level instead of at source level (would fix #2320)
- [x] distinguish between source and binary class names. Introduce a new relation `binaryClassName: Relation[String, String]` and use it whenever the conversion is needed.
#### Track name hashes per class (phase 2)
- [x] refactor tracking of APIs to be done at class level instead of at source level
- [x] extract APIs of all classes (including inner ones) separately
- [x] track name hashes at class level instead of being at source level (see also: #2319)
- [x] implement the rest of invalidation logic (member reference, name hashing) based on class names (get rid of mapping through `declaredClasses` in almost entire algorithm) (would fix #2320)

Only once the last bullet is merged we'll see improvement to incremental compilation when big number of nested classes is involved (e.g. like in Scala compiler's case).
#### Testing, bug fixing and benchmarking (phase 3)
- [x] test with the Scalatest repo
- [x] test with the scala repo using the sbt build
- [x] test with the specs2 repo
- [x] fix handling of anonymous and local classes defined in Java (similar logic will have to be implemented as for handling local Scala classes) (https://github.com/sbt/zinc/issues/192)
[...]

PR #2432, opened 2016-02-02
Invalidate package objects transitively in name hashing

### LINKED ISSUE #2326: Name hashing breaks when using a package object hierarchy
- sbt version 0.13.9
- scala version 2.11.7

``` scala
// src/main/scala/p/impl/Foo.scala
package p.impl

trait Foo
```

``` scala
// src/main/scala/p/impl/PackageObject.scala
package p.impl

abstract class PackageObject extends Foo
```

``` scala
// src/main/scala/p/package.scala
import p.impl.PackageObject

package object p extends PackageObject
```

Any change in `Foo.scala`, even adding a space at the beginning of the file, causes:

```
[error] missing or invalid dependency detected while loading class file 'PackageObject.class'.
[error] Could not access type Foo in package p.impl,
[error] because it (or its dependencies) are missing. Check your build definition for
[error] missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
[error] A full rebuild may help if 'PackageObject.class' was compiled against an incompatible version of p.impl.
[error] one error found
[error] (compile:compileIncremental) Compilation failed
```

The current workaround is disabling name hashing:

```
incOptions ~= (_ withNameHashing false)
```

/cc @gkossakowski

[EDIT 2015-12-21: Moved `Foo` and `PackageObject` into a distinct package, and moved files into directories to follow the package structure.. to further remove doubts]

gkossakowski:
Modify `invalidatedPackageObjects` to look for package objects that
transitively inherit from an invalidated class instead of just inheriting
directly. This is the correct behavior that should have been implemented
in the first place. The bug comes from a subtle copy&paste mistake. The old
implementation used `publicInherited` relation and new one looks identical
but uses `inheritance` relation. The difference is that `publicInherited`
was a relation that was expanded transitively along inheritance chain but
`inheritance` is not. We have to perform the transitive walk in
`IncrementalNameHashing.invalidatedPackageObjects` implementation.

Mark `pkg-self` test as passing.

Fixes #2326


typesafe-tools:

compile/inc/src/main/scala/sbt/inc/IncrementalNameHashing.scala
```diff
@@ -18,7 +18,9 @@ private final class IncrementalNameHashing(log: Logger, options: IncOptions) ext
   // Package objects are fragile: if they inherit from an invalidated source, get "class file needed by package is missing" error
   //  This might be too conservative: we probably only need package objects for packages of invalidated sources.
   override protected def invalidatedPackageObjects(invalidated: Set[File], relations: Relations): Set[File] =
-    invalidated flatMap relations.inheritance.internal.reverse filter { _.getName == "package.scala" }
+    transitiveDeps(invalidated)(relations.inheritance.internal.reverse).filter {
+      _.getName == "package.scala"
+    }
```
paulp, line 23: I realize it's not new to this commit, but what is the basis for this filter? It's not supposed to be a correctness requirement that package objects be defined in files called package.scala. It is (according to the scala specification) a convention.

Duhemm:
LGTM.

@paulp's comment is right. Maybe that we should add a flag `isPackageObject`, similar to `hasMacro`, and rely on that rather than the filename?

compile/inc/src/main/scala/sbt/inc/IncrementalNameHashing.scala
gkossakowski, line 23: Yes, it was a quick hack. @Duhemm's idea would fix it for good.

gkossakowski:
@Duhemm: yes, that's a good way to implement package object handling. Would you like to give it a try?

Duhemm:
@gkossakowski Sure!

gkossakowski:
great, small suggestion for implementation: introduce `hasPackageObject` to `APIs` that checks the flag in the `Source` object. This way the particular mechanism of storing that information can be hidden and changed later on. E.g. I'll need to change it for #1104.

paulp:
Note if you haven't already that `_.getName == "package.scala"` occurs more than one place in the sbt source.

Issue #2436, opened 2016-02-02
Regression in name hashing of traits


gkossakowski:
``` scala
// A.scala
trait A {
  //def foo: Int = 12
}

// B.scala
class B(a: A)
```

When you uncomment definition of `foo` `B.scala` is recompiled but it shouldn't. The recompilation of `B.scala` caused by change of a hash of `A` caused by addition of `foo`. This regressed in 40ebc825317ed29a2ab87be07246fcdcfc6d1e3d.

I think it's large enough regression that should be considered as a showstopper for 0.13.10.
Unfortunately, existing scripted tests don't cover this scenario.

//cc @eed3si9n

gkossakowski:
@Duhemm: I assigned you to have a look into this.

Duhemm:
Fixed in #2441

dwijnand:
Btw the reason this isn't closed is because GitHub issues close when the `Fixes #XXXX` commit hits the default branch, which for sbt/sbt is 1.0.x.

PR #2437, opened 2016-02-02
Add a pending test for a modified member of a trait

gkossakowski:
The test shows that name hashing optimization is broken for members
defined in traits. The problem is that modification of a member of a trait
triggers change of a hash of trait's name.

The behavior covered by this test regressed in
40ebc825317ed29a2ab87be07246fcdcfc6d1e3d
The bug is tracked in #2436.


gkossakowski:
Review by @Duhemm

typesafe-tools:

Duhemm:
Shouldn't this test be located in `source-dependencies` rather than `compiler-project`?

Apart from that, LGTM

gkossakowski:
I put it into `compiler-project` to emphasize that this test doesn't check if dependencies are tracked correctly but whether invalidation is done correctly. However, many tests in source-dependencies test for invalidation too so probably makes sense to put everything in one place. I'll move it.

Duhemm:
@gkossakowski Could you please update your branch? It looks like GitHub forces you to have an up-to-date branch when using required status check: https://github.com/blog/2051-protected-branches-and-required-status-checks

gkossakowski:
Do you mean you cannot merge my PR? I see the merge button here (I still have merge rights to sbt).

Duhemm:
That's strange. Maybe it's because Travis has not finished running the build... I have a grayed out merge button.

screen shot 2016-02-03 at 16 52 01

dwijnand:
@gkossakowski as an owner you can override the status check requirement.

gkossakowski:
Ok, in that case I'll just merge it.

PR #2438, opened 2016-02-03
Do not rely on filename of package objects

Duhemm:
Add a new field in `Source` that indicates whether this `Source`
represents a package object. This field is used in the incrmental
compiler to perform recompilation of package objects rather than relying
on the file name being `package.scala`.

See discussion in #2432


typesafe-tools:

dwijnand:
As a side note, in scripted is a pending test run and does it fail the build if it's successful?

compile/inc/src/main/scala/sbt/inc/APIs.scala
```diff
@@ -38,6 +38,8 @@ trait APIs {
 
   def internal: Map[File, Source]
   def external: Map[String, Source]
+
+  def hasPackageObject(src: File): Boolean
```
dwijnand, line 42: bincompat?

Duhemm, line 42: Adding a new method does not break binary compatibility. People who extend this trait will have to implement `hasPackageObject`, but classes that have already been compiled should not need to be recompiled.

Duhemm:
Yes, pending scripted tests are run and a successful pending test will cause a build failure since #2178

compile/inc/src/main/scala/sbt/inc/APIs.scala
dwijnand, line 42: Is this a reported false positive then from MiMa?

```
[error]  * abstract method hasPackageObject(java.io.File)Boolean in interface t.APIs does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("t.APIs.hasPackageObject")
```

Duhemm, line 42: Oh, I see. In the past we've been adding new methods without considering it as a binary compatibility breakage. I could maybe make this method `private[sbt]`. I'm not sure what would be best.

In any case, this shouldn't break existing plugins.

dwijnand, line 42: I think it only breaks forward compatibility not backwards compatibility?

Also even if it's a breaking change that doesn't necessarily make it a blocker. sbt exposes too much and we've made concessions before. It's partially why we don't use MiMa currently in 0.13.

dwijnand:
I thought so. Excellent.

eed3si9n:
LGTM

gkossakowski:
//cc @paulp

gkossakowski:
I found out that sbt tracks package objects already with `DefinitionType` enum, see: https://github.com/sbt/sbt/blob/0.13/interface/other#L72

I'll use that in class-based dependency tracking branch.

PR #2440, opened 2016-02-04
Do not create dir, if trying to output classes to a jar

cunei:
On occasion, it may be useful to manipulate `classDirectory` so that a jar file, rather than a class directory, is used as a destination for the compiled class files (see https://github.com/scala/scala/commit/2700617052).

This tiny change prevents a directory to be created with the same name as a desired jar file, as that would prevent the jar from being correctly created in that case.

### DIFF
--- a/compile/integration/src/main/scala/sbt/compiler/MixedAnalyzingCompiler.scala
+++ b/compile/integration/src/main/scala/sbt/compiler/MixedAnalyzingCompiler.scala
@@ -37,7 +37,7 @@ final class MixedAnalyzingCompiler(
    */
   def compile(include: Set[File], changes: DependencyChanges, callback: AnalysisCallback): Unit = {
     val outputDirs = outputDirectories(output)
-    outputDirs foreach (IO.createDirectory)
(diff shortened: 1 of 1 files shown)


dwijnand:
I noticed this is targeting the 0.13.10 branch, is this fixing a regression that 0.13.10-RC1 or 0.13.10-M1 introduces?

cunei:
No, it is not a regression fix. Should I target 0.13 or 1.0.x, instead?

dwijnand:
Yes, and by yes I mean both.. please :+1:

The change for 1.0 would have to be a pull request to https://github.com/sbt/incrementalcompiler

eed3si9n:
I'm going to allow this change to 0.13.10 since it's a minor fix that enables us to do straight-to-jar compilation.

dwijnand:
Ok :+1:

dwijnand:
Ok :+1:

cunei:
The test "dependency-management/cached-resolution-classifier" failed in Travis, but it succeeds on my machine.

dwijnand:
Restarted

PR #2448, opened 2016-02-09
add documention for FileFunction.cached

timcharper:


cache/tracking/src/main/scala/sbt/Tracked.scala
```diff
@@ -231,6 +231,23 @@ class Difference(val cache: File, val style: FilesInfo.Style, val defineClean: B
 object FileFunction {
   type UpdateFunction = (ChangeReport[File], ChangeReport[File]) => Set[File]
 
+  /**
+    Incremental compilation helper. Returns a function whose input is a Set of
```
eed3si9n: Tracked API is a generic caching utility, and not specifically for incremental compilation only.

dwijnand:
LGTM thanks.

eed3si9n:
LGTM too.

PR #2453, opened 2016-02-10
Avoid CCE when scalac internally uses compileLate

retronym:
For example, when the `--sourcepath` option is provided
and the refchecks phase compiles an annotation found
on a referenced symbol from the sourcepath.

`compileLate` assumes that all non-sentinel compiler
phases can be down cast to `GlobalPhase`.

This commit changes the two phases in SBT to extend
this instead of `Phase`. This has the knock on benefit
of simplifying the phases by letting the `GlobalPhase.run`
iterator over the list of compilation units and feed them
to us one by one.

I checked that the test case failed before making each
change.

### DIFF
--- a/compile/interface/src/main/scala/xsbt/API.scala
+++ b/compile/interface/src/main/scala/xsbt/API.scala
@@ -22,16 +22,19 @@ final class API(val global: CallbackGlobal) extends Compat {
   @inline def debug(msg: => String) = if (settings.verbose.value) inform(msg)
 
   def newPhase(prev: Phase) = new ApiPhase(prev)
-  class ApiPhase(prev: Phase) extends Phase(prev) {
+  class ApiPhase(prev: Phase) extends GlobalPhase(prev) {
     override def description = "Extracts the public API from source files."
     def name = API.name
-    def run: Unit =
+    override def run(): Unit =
       {
         val start = System.currentTimeMillis
-        currentRun.units.foreach(processUnit)
+        super.run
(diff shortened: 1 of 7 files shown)


eed3si9n:
Would it make sense to change https://github.com/sbt/sbt/blob/0.13/compile/interface/src/main/scala/xsbt/Analyzer.scala while we're at it?

compile/interface/src/main/scala/xsbt/API.scala
```diff
@@ -22,16 +22,19 @@ final class API(val global: CallbackGlobal) extends Compat {
...
-  class ApiPhase(prev: Phase) extends Phase(prev) {
+  class ApiPhase(prev: Phase) extends GlobalPhase(prev) {
     override def description = "Extracts the public API from source files."
     def name = API.name
-    def run: Unit =
+    override def run(): Unit =
       {
         val start = System.currentTimeMillis
-        currentRun.units.foreach(processUnit)
+        super.run()
```
eed3si9n: `compiler-project / run2.8` failed:

```
[error] /tmp/sbt_2e562db/xsbt/API.scala:31: error: ApiPhase.super.run of type Unit does not take parameters
[error]         super.run()
[error]                  ^
[error] one error found
```

retronym:
@eed3si9n Just updated the extraction phase, too, and fixed the compile error.

eed3si9n:
LGTM

PR #2456, opened 2016-02-11
FPORT: Invalidate package objects transitively in name hashing

Duhemm:
Forward port of #2432


Duhemm:
This PR needs to target only sbt/incrementalcompiler

PR #2457, opened 2016-02-11
FPORT: Avoid CCE when scalac internally uses compileLate

Duhemm:
Forward port of #2453


dwijnand:
LGTM

dwijnand:
How is this passing given the forward port to sbt/incrementalcompiler isn't even merged, let alone released and updated to??

Duhemm:
The test was not in the correct directory for sbt 1.0, good catch!

dwijnand:
Rebased in #2487

"... and he snags the ball and runs with it..."

Issue #2460, opened 2016-02-15
No tab-completion for tasks defined in AutoPlugin's buildSettings


dotta:
To reproduce, use the following auto plugin

```
package bug

import sbt._

object MyPlugin extends AutoPlugin {
  override def trigger = noTrigger
  object autoImport {
    val myTask = taskKey[Unit]("My task")
    val otherTask = taskKey[Unit]("My other task")
  }

  import autoImport._

  override def buildSettings = super.buildSettings ++ Seq(
    myTask := println("Called my task")
  )

  override def projectSettings = Seq(
    otherTask := println("Called other task")
  )
}
```

Then, I've created a build that enables the project

```
// plugins.sbt
addSbtPlugin("foo" % "sbt-plugin" % "0.1-SNAPSHOT")
```

```
// build.sbt
enablePlugins(MyPlugin)
```

And, finally, start the sbt console on this project, and you'll see that tab-completion is available for the `otherTask` defined in the AutoPlugin projectSettings, but no tab-completion for `myTask`:

```
$ sbt
[info] Loading global plugins from /Users/mirco/.sbt/0.13/plugins
[info] Loading project definition from /Users/mirco/tmp/test-sbt-bug/project
[info] Updating {file:/Users/mirco/tmp/test-sbt-bug/project/}test-sbt-bug-build...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
[info] Set current project to test-sbt-bug (in build file:/Users/mirco/tmp/test-sbt-bug/)
> other
otherResolvers   otherTask        
> otherTask
My other task
[success] Total time: 0 s, completed Feb 15, 2016 12:27:11 PM
> my
{invalid input}   
> myTask
Called my task
[success] Total time: 0 s, completed Feb 15, 2016 12:27:19 PM
```

retronym:
Note that the completion options are available under:

```
> {.}/my<TAB>

```

Which probably comes down to: https://github.com/sbt/sbt/blob/v0.13.9/main/settings/src/main/scala/sbt/Def.scala#L26-L31

dotta:
I believe a fix for this ticket was merged. Closing?

dotta:
@eed3si9n @Duhemm should this ticket be closed? And, if yes, will the fix be available in the next 0.13.x release?

eed3si9n:
I'm guessing it's open because it hasn't been forward ported to sbt 1.0 yet.

[...]

PR #2469, opened 2016-02-22
Completion for build-level keys

Duhemm:
sbt's shell provided completion only for keys that were relative to a
defined project, but didn't provide completion for keys that belong to
the build definition only.

This commit fixes this issue by defining a new kind of `Parser` (from
which completions are generated) which runs its input simultaneously on
distinct parsers. We now define a parser for project-level keys and
another parser for build-level keys. These two parsers are eventually
combined, and we get the completions of both parsers.

Fixes sbt/sbt#2460

### DIFF
--- a/main/src/main/scala/sbt/Act.scala
+++ b/main/src/main/scala/sbt/Act.scala
@@ -48,12 +48,22 @@ object Act {
           new ParsedKey(makeScopedKey(proj, conf, task, extra, key), mask)
         }
 
-      for {
-        rawProject <- optProjectRef(index, current)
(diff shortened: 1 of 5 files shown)


dwijnand:
Cool :)

eed3si9n:
LGTM

eed3si9n:
Notes plz

Duhemm:
@eed3si9n I just added the notes.

eed3si9n:
Could you forward port this?

eed3si9n:
Started util side of FPORT https://github.com/sbt/util/pull/41

PR #2475, opened 2016-02-23
Handle source, docs artifacts correctly for Ivy [1.0.x]

oxbowlakes:
This is a re-installment of https://github.com/sbt/sbt/pull/1016 for the 1.0.x branch, sitting on top of changes already merged in https://github.com/sbt/librarymanagement/pull/25.

To reiterate, specifically in this PR we are addressing the following:
- Introduce the SettingKeys `sourceArtifactTypes` and `docArtifactTypes`: Sets which contain the artifact types that would qualify them as either source or doc artifacts.
- Use the new `ArtifactTypeFilter` introduced in https://github.com/sbt/librarymanagement/pull/25 in the `update` task (specifically, I added it to `UpdateConfiguration`) to allow "all artifact types except those found in either `sourceArtifactTypes` or`docArtifactTypes`"
- In `updateClassifiers` / `updateSbtClassifiers`, invert the filter from the `UpdateConfiguration`: it will therefore _only_ let through sources and docs
- `srcTypes` and `docTypes` now have to also be explicitly passed into `GetClassifiersConfiguration` because (since https://github.com/sbt/librarymanagement/pull/25) after finishing the update in `updateClassifiers`, it will map over all artifacts to set their classifier to either `sources` or `javadoc`, if it's not already set, according to a reverse mapping from the `(src|doc)Types` to "source" and "javadoc" respectively (well, `Artifact.SourceClassifier` and `Artifact.DocClassifier`, to be precise). This is because the artifacts retrieved from Ivy publications might well not have this attribute set, and last time I checked, IDE plugins such as IntelliJ's depended exactly on this value to classify the artifacts.

Hopefully, after this, plugins could just look at the artifact's type and see in which of the two sets contains it. That would also resolve a great deal of pain surrounding "e:classifier". For instance, I've discovered that it's impossible to use that attribute in an ivy repository if its value is not either "sources" or "javadoc" (e.g. if it's src), because plugins expect it to be one of those two values as mentioned before.

### DIFF
--- a/main/actions/src/main/scala/sbt/CacheIvy.scala
+++ b/main/actions/src/main/scala/sbt/CacheIvy.scala
@@ -115,9 +115,9 @@ object CacheIvy {
   private[this] val crossToInt = (c: CrossVersion) => c match { case Disabled => 0; case b: Binary => BinaryValue; case f: Full => FullValue }
 
   implicit def moduleIDFormat(implicit sf: Format[String], bf: Format[Boolean]): Format[ModuleID] =
-    wrap[ModuleID, ((String, String, String, Option[String]), (Boolean, Boolean, Boolean, Seq[Artifact], Seq[ExclusionRule], Map[String, String], CrossVersion))](
-      m => ((m.organization, m.name, m.revision, m.configurations), (m.isChanging, m.isTransitive, m.isForce, m.explicitArtifacts, m.exclusions, m.extraAttributes, m.crossVersion)),
-      { case ((o, n, r, cs), (ch, t, f, as, excl, x, cv)) => ModuleID(o, n, r, cs, ch, t, f, as, excl, x, cv) }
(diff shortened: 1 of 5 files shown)


lightbend-cla-validator:
Hi @dansanduleac,

oxbowlakes:
This PR is not from @dansanduleac, it's from @oxbowlakes. I believe that we've already signed the typesafe CLA

eed3si9n:
Yea you guys should be fine.

dansanduleac:
Looks like this needs to wait on https://github.com/sbt/incrementalcompiler/pull/70, and then subsequently on further changes that would get this project to compile cleanly against `incrementalcompiler`.

Duhemm:
I think this is superseded by #2478. Thanks @oxbowlakes & @dansanduleac!

PR #2478, opened 2016-02-24
Forward port "Handle source, docs artifacts correctly for Ivy [1.0.x]" + update modules

Duhemm:
This PR currently refers to a locally published version of `incrementalcompiler`, which includes both sbt/incrementalcompiler#70 and sbt/incrementalcompiler#71.

I'm thinking that maybe we should merge this PR in priority and then rebase all the forward port that were waiting for a release of `incrementalcompiler` or `librarymanagement`.

Note that this is not only adapting to changes. Because of the changes to the API of the incremental compiler I've had to make some other changes:
- Created a new key `compileOptions` that produces a `xsbti.compile.CompileOptions`,
- Created two new keys `sourceTypes` and `docTypes` that specify which artifacts must be considered source or doc artifacts.


dansanduleac:
Looks like you've replicated quite a bit of work already in https://github.com/sbt/sbt/pull/2475, which is supposed sbt/sbt's counterpart of https://github.com/sbt/librarymanagement/pull/25 (the breaking changes in `librarymanagement-0.1.0-M7`), specifically the `srcTypes` / `docTypes` settings, and the `CacheIvy` changes. :)

Though, you should be aware that it's also necessary to set a default filter for `UpdateConfiguration` that excludes src/doc artifacts, since you only want the binaries, and subsequently inverts that filter when getting classifiers  (see https://github.com/sbt/sbt/pull/2475/files#diff-6373e7f7122325e753b75fe1cc76ff5fR1268). 

Would it be possible to rebase these changes on top of that PR so that we don't get any overlap?

Duhemm:
@dansanduleac You're right. Maybe that it would be better to first merge #2475 and then rebase this one.

Duhemm:
I have rebased this PR on #2475. The problem is that #2475 upgrades the dependency on `librarymanagement` to 0.1.0-M7, but `incrementalcompiler` still uses 0.1.0-M3.

This PR now includes the changes of #2475 plus the changes that will be required once a new release of `incrementalcompiler` is cut.

dansanduleac:
@Duhemm Thanks for rebasing. I see what's going on - all these subprojects are causing dependency hell. 
Perhaps it makes sense for 
- a PR to update `incrementalcompiler` to use `librarymanagement-0.1.0-M6` first, then
- a PR here based off that, and only afterwards 
- upgrade `incrementalcompiler` to use `0.1.0-M7` & cut a new release
- merge #2475 cleanly (since it's only necessary on top of `librarymanagement-0.1.0-M7`)

I suspect your changes fall into step (2) above.

eed3si9n:
https://travis-ci.org/sbt/sbt/jobs/111582958

```
Running dependency-management / make-pom
[error] x dependency-management / make-pom [PENDING]
[error]   Mark as passing to remove this failure.

[error] x dependency-management / pom-type
[error]    {line 1}  Command failed: checkPom failed
```

A bunch of failures here too - https://travis-ci.org/sbt/sbt/jobs/111583017

eed3si9n:
Merged

PR #2395, opened 2016-01-17
FPORT: Refix snapshots

dwijnand:
Forward port of #2313. Depends on sbt/librarymanagement#20.

PR #2313, opened 2015-12-14
Fixes #1514. Enable latest SNAPSHOT option by default

eed3si9n:
Fixes #1514
Fixes #1616
#1520 originally fixed #1514 by reimplementing part of chain resolver to check all resolvers to find the latest snapshot artifacts. This behavior did not work well:
1. with Maven repositories where sbt was failing to calculate the correct publication dates.
2. with module doesn't have any artifacts (parent POMs). #1616 
#2075 should fix the publication date issue, and  we should enable this flag back again. 1bf789eb3dc48f8786da326f2ce18793f09faea5 fixes the module without artifacts.

The build user can opt out by:

``` scala
updateOptions := updateOptions.value.withLatestSnapshots(false)
```

/review @jsuereth, @dwijnand, @Duhemm

PR #2484, opened 2016-02-26
FPORT: Refix snapshots

eed3si9n:
Forward port of #2313. 
Subsumes of #2395
Continuation of #2485


eed3si9n:
It seems like there are other pull request that should go before this.

eed3si9n:
Rebased on top of 1.0.x

PR #2485, opened 2016-02-26
FPORT: Maven compatibility changes + Adds CompatibilityWarningOptions

eed3si9n:
Forward port of:
- https://github.com/sbt/sbt/pull/2345
- https://github.com/sbt/sbt/pull/2407

These need to go in together since librarymanagement has both forward ported already.


eed3si9n:
We set dependency-management/pom-type to pending in #2478 and now it says it's passing. Maybe this test is kind of flaky since I don't think the change introduced in this PR.

PR #2486, opened 2016-02-26
FPORT: Adds more memory to internal scripted task

eed3si9n:
Forward port of #2355


eed3si9n:
This looks like a trivial forward port, so I am merging this without Travis.

PR #2487, opened 2016-02-27
FPORT: Avoid CCE when scalac internally uses compileLate

dwijnand:
Forward port of #2453

@dwijnand's rebase of @Duhemm's #2457 forward port.


eed3si9n:
sbt/incrementalcompiler implements scripted test, so this should move there too.

dwijnand:
True, will move.

dwijnand:
Moved to https://github.com/sbt/incrementalcompiler/pull/73

Issue #2431, opened 2016-02-01
Scopes and artifacts in sbt


romanowski:
**Steps:**
Small repro can be found here: https://github.com/romanowski/sbt-problems/tree/artifact-resolve-problem

**Problem:**

``` scala
// following lines cannot be resolved
libraryDependencies += "a" % "b" % "1.0.0" % "compile->runtime" artifacts(Artifact("b1", "jar", "jar"))

libraryDependencies += "a" % "b" % "1.0.0" % "test->runtime" artifacts(Artifact("b1", "jar", "jar"))

// however changing this to code below fixes problem
libraryDependencies += "a" % "b" % "1.0.0" % "compile,test->runtime" artifacts(Artifact("b1", "jar", "jar"))
```

This happens only when artifacts are present. For some reason sbt tries to download "b.jar" (in general "[module].jar"

Output:

``` code
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[warn]  [FAILED     ] a#b;1.0.0!b.jar:  (0ms)
[warn] ==== local: tried
[warn]   /home/krzysiek/.ivy2/local/a/b/1.0.0/jars/b.jar
[warn] ==== public: tried
[warn]   http://repo1.maven.org/maven2/a/b/1.0.0/b-1.0.0.jar
[warn] ==== buggy: tried
[warn]   /home/krzysiek/ivy-repo/repo/a/b/1.0.0/b.jar
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]  ::              FAILED DOWNLOADS            ::
[warn]  :: ^ see resolution messages for details  ^ ::
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]  :: a#b;1.0.0!b.jar
```

**Expectations**
Sbt should handle both ways of configuring scopes (in case of artifacts are present)

Duhemm:
Fixed in #2500

PR #2500, opened 2016-03-07
Fix sbt/sbt#2431

Duhemm:
When concatenating the artifacts coming from two modules, we sometimes
attempted to create a default artifact from the organization and name of
the module.

However, this may fail because a module a % b % "1.0" may not have an
artifact named "b.jar" (see sbt/sbt#2431).

Fixes sbt/sbt#2431.


Duhemm:
It looked to me that sometimes we tried to create non existing (as in not explicitly mentioned in the `DependencyDescriptor`) artefacts. However, if the `DependencyDescriptor` comes from a `POM` file, we will have that `dd.getAllDependencyArtifacts` is empty. In such cases, it is correct to create the default artifact the way we used to.

We try to create a default artifact when trying to merge the artefacts of two `DependencyDescriptor`. We should pass to `defaultArtifact` both `DependencyDescriptor`s to see whether both have empty dependencies before creating the default artifact.

eed3si9n:
Could you forward port this?

PR #2507, opened 2016-03-08
Fix instability of self variable API representation

### LINKED ISSUE #2504: Unstable representation of self variable
Project with three source files:

``` scala
// Global.scala
class Global { 
  class Foo extends Namers
}

// Namers.scala
class Namers {
  class Namer { thisNamer => }
}

// Client.scala
class Client(global: Global)
```
## steps
1. compile
2. add a whitespace to `Global.scala` and watch `Client.scala` being recompiled
## problem

Whitespace change causes recompilation of dependent files.
## notes

The problem is unstable representation of classes that have a self variable (check `Namer`). This instability comes from not forcing info of the symbol that represent self type/self variable. This is a regression caused by 1ce11230546639fd4a06ab96d19784979a7a358c.

gkossakowski:
The reason for instability is a bit tricky so let's unpack what the previous code checking if there's self type declared was doing. It would check if `thisSym` of a class is equal to a symbol representing the class. If that's true, we know that there's no self type. If it's false, then `thisSym` represents either a self type or a self variable. The second (type test) was supposed to check whether the type of `thisSym` is different from a type of the class. However, it would always yield false because TypeRef of `thisSym` was compared to ClassInfoType of a class. So if you had a self variable the logic would see a self type (and that's what API representation would give you).

Now the tricky bit: `thisSym` is not pickled when it's representing just a self variable because self variable doesn't affect other classes referring to a class. If you looked at a type after unpickling, the
symbol equality test would yield true and we would not see self type when just a self variable was declared.

The fix is to check equality of type refs on both side of the type equality check. This makes the pending test passing.

Also, I added another test that checks if self types are represented in various combinations of declaring a self variable or/and self type.

Fixes #2504.

### DIFF
--- a/compile/interface/src/test/scala/xsbt/ExtractAPISpecification.scala
+++ b/compile/interface/src/test/scala/xsbt/ExtractAPISpecification.scala
@@ -1,7 +1,7 @@
 package xsbt
 
 import org.junit.runner.RunWith
-import xsbti.api.{ Definition, SourceAPI, ClassLike, Def }
+import xsbti.api._
 import xsbt.api.SameAPI
 import org.specs2.mutable.Specification
 import org.specs2.runner.JUnitRunner
@@ -40,11 +40,13 @@ class ExtractAPISpecification extends Specification {
   }
 
   /**
-    * Checks if representation of the inherited Namer class (with a declared self variable) in Global.Foo
-    * is stable between compiling from source and unpickling. We compare extracted APIs of Global when Global
-    * is compiled together with Namers or Namers is compiled first and then Global refers
-    * to Namers by unpickling types from class files.
-    */
+   * Checks if representation of the inherited Namer class (with a declared self variable) in Global.Foo
+   * is stable between compiling from source and unpickling. We compare extracted APIs of Global when Global
+   * is compiled together with Namers or Namers is compiled first and then Global refers
+   * to Namers by unpickling types from class files.
+   *
+   * See https://github.com/sbt/sbt/issues/2504
+   */
   "Self variable and no self type" in {
     def selectNamer(api: SourceAPI): ClassLike = {
       def selectClass(defs: Iterable[Definition], name: String): ClassLike = defs.collectFirst {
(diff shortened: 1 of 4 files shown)


gkossakowski:
Review by @adriaanm

gkossakowski:
I think this fix (once reviewed and merged) might warrant a new release. The bug is rather bad because it degrades performance of incremental compiler in a way that's hard to both detect and debug. The counterargument against expedited release would be that it only affects code bases that declare inner classes with self variables.

eed3si9n:
We can do a milestone right away so ppl can test it if they run into this.

compile/interface/src/test/scala/xsbt/ExtractAPISpecification.scala
```diff
@@ -39,4 +38,67 @@ class ExtractAPISpecification extends Specification {
...
+      val foo = selectClass(global.structure.declared, "Global.Foo")
+      selectClass(foo.structure.inherited, "Namers.Namer")
+    }
+    val src1 =
+      """|class Namers {
+         |  class Namer { thisNamer => }
+         |}
+         |""".stripMargin
+    val src2 =
+      """|class Global {
```
adriaanm, line 65: Did you end up figuring out why nesting was needed to repro?

compile/interface/src/test/scala/xsbt/ExtractAPISpecification.scala
```diff
@@ -39,4 +38,67 @@ class ExtractAPISpecification extends Specification {
...
+   * Checks if self type is properly extracted in various cases of declaring a self type
+   * with our without a self variable.
+   */
+  "Self type" in {
+    def collectFirstClass(defs: Array[Definition]): ClassLike = defs.collectFirst {
+      case c: ClassLike => c
+    }.get
+    val srcX = "trait X"
+    val srcY = "trait Y"
+    val srcC1 = "class C1 { this: C1 => }"
```
adriaanm, line 87: Lets add some variations that have a self variable but no explicit self type.
- `class Cthis { this =>  }`
- `class C_ { _ =>  }`
- `class Cself { self =>  }`

Also, instead of naming all classes `s"C$n"`, maybe have their name reflect whether they have a self type, so that the test is easy to proofread.

adriaanm:
Good stuff, @gkossakowski! Thanks for picking this up.

compile/interface/src/test/scala/xsbt/ExtractAPISpecification.scala
gkossakowski, line 65: Yes, because you need to inherit a class as a member to have its full API representation. If you just inherit from a class directly, then you're going to have API representation of its members.

adriaanm, line 65: Makes sense, thanks!

compile/interface/src/test/scala/xsbt/ExtractAPISpecification.scala
gkossakowski, line 87: I've tried longer names but it gets out of hand pretty quickly. The main value behind this test is not to be easy to read but easy to execute and produce a failure that can be diagnosed in isolation. I don't expect many people to read this ever again but if somebody comes along being confused by short names, she can refactor the code before fixing the bug this test uncovered.

I'll add the three variants you suggested and keep the class counter at single digit.

gkossakowski, line 87:
```
scala> class C7 { this => }
<console>:1: error: not a legal formal parameter.
Note: Tuples cannot be directly destructured in method or function parameters.
      Either create a single parameter accepting the Tuple1,
      or consider a pattern matching anonymous function: `{ case (param1, param1) => ... }
class C7 { this => }
```

A bug in parser?

gkossakowski:
I added more tests.

gkossakowski:
Ping

eed3si9n:
@adriaanm Can I merge it?

adriaanm:
LGTM, sorry for delay

eed3si9n:
@gkossakowski Please forward port this to sbt/zinc.

gkossakowski:
It's part of class-based branch that I'm preparing for merge.

On 18 March 2016 at 17:50, eugene yokota notifications@github.com wrote:

> @gkossakowski https://github.com/gkossakowski Please forward port this
> to sbt/zinc.
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> https://github.com/sbt/sbt/pull/2507#issuecomment-198449664

## 

Grzegorz Kossakowski
twitter: @gkossakowski http://twitter.com/gkossakowski
github: @gkossakowski http://github.com/gkossakowski

dwijnand:
Forward port in https://github.com/sbt/zinc/pull/90

PR #2510, opened 2016-03-11
Add Append & Remove instances for Option

dwijnand:


dwijnand:
@eed3si9n WDYT of this change?

eed3si9n:
LGTM

Issue #2435, opened 2016-02-02
cached resolution: Incremental update remove scoped from evicted libs


romanowski:
**Steps:**
Small repro can be found here: https://github.com/romanowski/sbt-problems/tree/incremental-update-scopes

**Problem:**
If incremental update is on when given lib is evicted all extra scopes are dropped.

``` scala
lazy val a = project settings(
    myResolver,
    updateOptions := updateOptions.value.withCachedResolution(true), //comment this line to make ws compile
    libraryDependencies += "a" % "b" % "1.0.0" % "compile->runtime",
    libraryDependencies += "a" % "b" % "1.0.0" % "compile->runtime2 "
)

lazy val b = project dependsOn(a) settings(
    myResolver,
    updateOptions := updateOptions.value.withCachedResolution(true), //comment this line to make ws compile
    libraryDependencies += "a" % "b" % "1.0.1" % "compile->runtime"
)
```

**Expectations**
Incremental update will preserve scopes declared in evicted libs.

Duhemm:
I've investigated this problem. Here are my findings:
- This project exhibits two kinds of conflicts: a conflict on versions (`1.0.0` vs `1.0.1`) and a conflict on configurations (`compile->runtime` AND `compile->runtime2` vs `compile->runtime` only).
- It looks like when cached resolution is off, sbt will select the most recent version and select all configuration (we end up with a dependency on `"a" % "b" % "1.0.1"` with configurations `compile->runtime` and `compile->runtime2`.
- When cached resolution is on, sbt will use the cached `UpdateReport` coming from project `a` (it contains the information about both configurations `runtime` and `runtime2` for its dependency `"a" % "b" % "1.0.0"`), and generate the `UpdateReport` for project `b` (it contains the information about `runtime` only for `"a" % "b" % "1.0.1"`).
- sbt will then select the most recent version, but has no idea about the configuration `runtime2`.

The goal is to eventually compile project `b` while having on the class path:
- project `a`
- `b1.jar` coming from `"a" % "b" % "1.0.1" % "compile->runtime"`
- `b2.jar` coming from `"a" % "b" % "1.0.1" % "compile->runtime"`
- `b3.jar` coming from `"a" % "b" % "1.0.1" % "compile->runtime2"`

The issue is that we will update project `b` and ask only for `"a" % "b" % "1.0.1" % "compile->runtime"`, and replace both configurations coming from project `a` by only one. 

[...]

PR #2513, opened 2016-03-14
Fix for sbt/sbt#2435

Duhemm:
Fixes sbt/sbt#2435

Consider a configuration where we have two projects, A and B.

A has a library dependency on "a" % "b" % "1.0.0" % "compile->runtime"
and "a" % "b" % "1.0.0" % "compile->runtime2"

B depends on project A, and has a library dependency on
"a" % "b" % "1.0.1" % "compile->runtime".

Note that project B depends on a more recent version of "a" % "b" than
project A, and that it depends ONLY on it's "runtime" configuration.

However, when compiling project B, we expect to have on the classpath
project A, and "a" % "b" % "1.0.1" % "compile->runtime" AND
"a" % "b" % "1.0.1" % "compile->runtime2" because it is part of the
compile configuration of project A.

This commit changes the cached resolution engine so that it behaves like
that, by first resolving dependencies on other project and then ensuring
that the dependent project specifies dependencies on the same
configurations.


Duhemm:
I still can't reproduce the failure on my machine. The failure on Travis is different from the original failure (before my fix).

Wiping the cache on Travis seems to fix the issue.

eed3si9n:
notes and forward port plz

eed3si9n:
LGTM

eed3si9n:
I'm seeing Travis CI failures on dependency-management/cached-resolution-configurations from PRs that has nothing to do with library management.

PR #2528, opened 2016-03-31
Implements scriptedBufferLog on sbt build

eed3si9n:
This adds `scriptedBufferLog` key to sbt's build. Currently this value is hardcoded to buffered `true`, but I often want to see the log realtime when I'm coding.

/review @dwijnand, @Duhemm


dwijnand:
:+1:

PR #2530, opened 2016-03-31
Deprecate the Build trait

dwijnand:


main/src/main/scala/sbt/Build.scala
```diff
@@ -8,6 +8,7 @@ import Keys.{ name, organization, thisProject, autoGeneratedProject }
 import Def.{ ScopedKey, Setting }
 
 // name is more like BuildDefinition, but that is too long
+@deprecated("Use .sbt format instead", "0.13.12")
```
eed3si9n, line 11: I think we should write a migration guide in http://www.scala-sbt.org/0.13/docs/Full-Def.html and link to it from here.

dwijnand, line 11: Right, yeah. There and in the notes as well.

I'll try and sketch one out. I'll need a little bit of your time, as your documentation-fu is stronger than mine.

eed3si9n:
I'll merge this for now.

dwijnand:
Forward ported with a breaking change in https://github.com/sbt/sbt/pull/2524

Issue #2537, opened 2016-04-04
name hashing does not track names of synthetic symbols and names from synthetic methods


eed3si9n:
https://github.com/gkossakowski/sbt/issues/9

> As discussed here the current strategy for extracting used names might miss synthetic names for case classes like `copy`, `apply`, `unapply`, etc.
> 
> This ticket is a reminder for me to investigate this issue further. Both for members of case classes and other synthetic members.
## steps

``` scala
// A.scala
case class A(a: Int)
// B.scala
class B { def test(a: A) = a.copy() }

// > compile

// A.scala
case class A(a: Int) { private def copy = ??? }

// > compile
// [info] Compiling 1 Scala source to /Users/jason/code/scratch1/target/scala-2.11/classes...
// [success]

// > ;clean;compile
// ...
// [info] Compiling 2 Scala sources to /Users/jason/code/scratch1/target/scala-2.11/classes...
// [error] /Users/jason/code/scratch1/src/main/scala/B.scala:2: method copy in class A cannot be accessed in A
// [error]   def test(a: A) = a.copy()
// [error]                      ^
```
## problem

Name hashing under compiles.
## expectation

Changes are detected.
## notes

isSynthetic is intentionally omitted.

``` scala
 private def eligibleAsUsedName(symbol: Symbol): Boolean = {
    def emptyName(name: Name): Boolean = name match {
      case nme.EMPTY | nme.EMPTY_PACKAGE_NAME | tpnme.EMPTY | tpnme.EMPTY_PACKAGE_NAME => true
      case _ => false
    }

    (symbol != NoSymbol) &&
      !symbol.isSynthetic &&
      !emptyName(symbol.name)
  }
```

jvican:
I think this was fixed in sbt/zinc#110. Can we close?

dwijnand:
That would seem to not do the "names of synthetic symbols" part.

dwijnand:
Migrated to https://github.com/sbt/zinc/issues/238 by @jvican

PR #2540, opened 2016-04-06
Adds withIncludeSynthToNameHashing.

eed3si9n:
Provides a workaround flag

``` scala
incOptions := incOptions.value.withIncludeSynthToNameHashing(true)
```

for name hashing
not including synthetic methods. This will not be enabled by default in
sbt 0.13. It can also enabled by passing `sbt.inc.include_synth=true`
to JVM. Ref #2537

/review @Duhemm, @gkossakowski, @dwijnand

### DIFF
--- a/cache/tracking/src/main/scala/sbt/Tracked.scala
+++ b/cache/tracking/src/main/scala/sbt/Tracked.scala
@@ -232,23 +232,23 @@ object FileFunction {
   type UpdateFunction = (ChangeReport[File], ChangeReport[File]) => Set[File]
 
   /**
-    Generic change-detection helper used to help build / artifact generation /
-    etc. steps detect whether or not they need to run. Returns a function whose
-    input is a Set of input files, and subsequently executes the action function
-    (which does the actual work: compiles, generates resources, etc.), returning
-    a Set of output files that it generated.
-
-    The input file and resulting output file state is cached in
-    cacheBaseDirectory. On each invocation, the state of the input and output
-    files from the previous run is compared against the cache, as is the set of
-    input files. If a change in file state / input files set is detected, the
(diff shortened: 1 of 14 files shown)


gkossakowski:
Fails with:

```
[error] /home/travis/build/sbt/sbt/interface/src/test/scala/xsbti/TestCallback.scala:8: class TestCallback needs to be abstract, since method includeSynthToNameHashing in trait AnalysisCallback of type ()Boolean is not defined
[error] class TestCallback(override val nameHashing: Boolean = false) extends AnalysisCallback
[error]       ^
[error] one error found
```

Assuming this error is fixed, this looks good to me.

@eed3si9n, I'm wondering if we could always include synthetic names. I haven't found a good reason to not to include them. What do you think?

eed3si9n:
@gkossakowski I can flip the switch and have it on by default. I thought we are concerned that changing `apply` could cascade into compiling all sources because everyone uses `apply`.

mpociecha:
So maybe does it make sense to add the possibility to specify the list of names which won't be included? It would be only the half-measure, I think, but it anyway may be worth to have it for the test purposes.

eed3si9n:
@mpociecha For now let's keep it simple on or off, and we can fine tune this more in sbt/zinc for 1.0, which will have synthetic methods included by default.

smarter:
Note that including synthetic members is not enough, you should also include synthetic classes (which are currently ignored), consider:

``` scala
case class A(x: Int)
// private object A
```

``` scala
object B {
  A(0) // should stop compiling when the private object line is uncommented
}
```

dwijnand:
Forward port in https://github.com/sbt/zinc/pull/110, with a breaking change for 1.0.

PR #2541, opened 2016-04-06
Static launcher, get bridge sources from resources

Duhemm:
This commit introduces a new "static" launcher that does not use Ivy to
gather all the artifacts that it requires, but rather expect them to be
immediately available.

To be able to use sbt without Internet access, we add a new
`ComponentCompiler` that is able to retrieve the bridge sources from the
resources on classpath and compile it.


Duhemm:
This PR still needs some work, I'll reopen it when it's ready.

PR #2543, opened 2016-04-06
[sbt 1.0] Reduce sbt.Project(...) params and make settings(...) lazy

### LINKED ISSUE #2104: make settings lazy
Making settings lazy will allow to safely make circular references between projects, that is very useful in many cases.
An example of typical usecase where it is useful is scalajs crossproject used for the whole app.
Here is an example:

``` scala
lazy val app = crossProject
        .crossType(CrossType.Full)
        .in(file("app"))
        .settings(
            name := "gsea",
            version := Versions.gsea
        )
        .jvmConfigure(p=>p.enablePlugins(SbtTwirl,SbtWeb))
        .jvmSettings(Revolver.settings:_*)
        .jvmSettings(
            libraryDependencies ++= Dependencies.akka.value ++ Dependencies.webjars.value,
            mainClass in Compile :=Some("org.denigma.preview.Main"),
            mainClass in Revolver.reStart := Some("org.denigma.preview.Main"),
            libraryDependencies ++= Dependencies.RLang.value ++ Dependencies.compilers.value,
            (managedClasspath in Runtime) += (packageBin in Assets).value
        )
        .jsSettings(
            libraryDependencies ++= Dependencies.sjsLibs.value,
            persistLauncher in Compile := true,
            persistLauncher in Test := false,
            jsDependencies += RuntimeDOM % "test"
        )

    lazy val appJS = app.js
    lazy val appJVM = app.jvm settings ( //I know it is unsafe, because settings are not lazy :((((
        resourceGenerators in Compile <+=
            (fastOptJS in Compile in appJS,    packageScalaJSLauncher in Compile in appJS) map(
                (f1, f2) => Seq(f1.data, f2.data)
                ),
        watchSources <++= (watchSources in  appJS)
        )
```

here jvm project is akka-http server to run the app
js project is scalajs frontend
shared code is in src folder

Here when akka-http project is started then it runs fastOptJS on js project and compiles javascript.

Such kind of circular references is so often done by scalajs users, that @gzm0 even had to add warning to scalajs docs http://www.scala-js.org/api/sbt-scalajs/0.6.0/index.html#org.scalajs.sbtplugin.cross.CrossProject

That means that there is clearly a demand for fixing this.

### LINKED ISSUE #2698: overzealous cycle detection
create a project containing `build.sbt`

``` scala
lazy val c = project.dependsOn(LocalProject("a") % "test")
lazy val a: Project = project.dependsOn(c)
```

start sbt, expecting a valid (if not crazy) project. Get

```
[error] Cyclic reference involving 
[error]    Project(id c, ...
[error]    Project(id a, ...
```

eed3si9n:
### sbt.Project(...)

The constructor that accepts settings are no longer supported because it breaks the auto plugins's setting ordering.
### lazy settings(...) method

Fixes #2104. The signature of `settings(...)` method is now changed to:

``` scala
def settings(ss: Eval[Def.SettingsDefinition]*): Project = ...
```

`Eval`, borrowed from Cats can wrap by-name parameter, which now allows us to make reference to other subprojects (in sbt 0.13 this causes circular reference).
### lazy dependsOn(...) method

Fixes #2698 

```
def dependsOn(deps: Eval[ClasspathDep[ProjectReference]]*): Project =
```
### caveat

The notation `settings(commonSettings: _*)` no longer works. Fortunately, `settings(commonSettings)` works in sbt 0.13.8+ (http://www.scala-sbt.org/0.13/docs/sbt-0.13-Tech-Previews.html#Nicer+ways+of+declaring+project+settings).


fommil:
Sooooo many builds use this horrible constructor. This is a good move :+1:

smarter:
The `: _*` thing is used in a bunch of place in http://github.com/sbt/website/

eed3si9n:
@smarter I am conflicted on this one. Is it laziness vs `_*`? Can you figure out implicits that can do both?

smarter:
For what it's worth, there's http://docs.scala-lang.org/sips/pending/repeated-byname.html, beyond that, no idea

eed3si9n:
I think @sjrd was able to do it in https://github.com/sbt/sbt/issues/2104#issuecomment-122513969

dwijnand:
What's the issue with : _\* no longer working?

eed3si9n:
@dwijnand It would be nice if the build just worked without changing the source for `build.sbt`. Maybe the right thing to do is provide `settingsLazy()` for ppl who needs laziness, and keep the existing signature for `settings(...)`?

dwijnand:
How do you feel about a runtime check in 0.13.12 .settings that warned about using : _\* use? (Runtime, which in sbt's case means when the build system initialises) That is provided we can correctly identify it, which I think might be possible.

I think we should expect to have a migration guide for 1.0, despite best intentions with 0.13, so let's break that build source given there's a better alternative available.

eed3si9n:
Another idea might be to write a macro that takes `A: _*` and wrap everything in `Eval`. That might give us both the source-compat and laziness.

dwijnand:
It's technically another option, but I think we should only reach for macros when the ROI is significant. Here I don't think it is. Making .settings a macro in 0.13 for the deprecation warning I'd be fine with (as it's stop-gap).

eed3si9n:
I did start a migration notes here - https://github.com/sbt/sbt/pull/2543/commits/cedade876c74b845fb345e2d3d62d6396e4b2a4f

> It's technically another option, but I think we should only reach for macros when the ROI is significant. Here I don't think it is. 

I disagree on that. Let's say there are [1M builds in the world](http://www.scala-lang.org/blog/2016/03/14/announcing-the-scala-center.html). If we count the occurrence at which we see `_*` vs nice-style, I think the result will be lopsided towards `_*`. 30% vs 1%? I don't know for sure, but the nice-style style is somewhat new, and I am guessing only hardcore fans of sbt know about it while `_*` has been documented on the tutorial as the best practice to organize the build.

dwijnand:
Right, in terms of number of breakages you're absolutely right (it also doesn't help that it's not available in sbt-scalajs's CrossProject, as a sidenote).

But I still think we shouldn't use a macro and stick it in the migration guide, given that there's a better alternative available.

Do you not like my idea of giving a deprecation warning in 0.13? Via runtime check or macro.

(I predict a very highly voted StackOverflow question with whatever the compiler error is for not migrating).

dwijnand:
Backing up a bit, I agree, we should fix the lazy issue with a special-cased `settingsLazy` method.

main/src/main/scala/sbt/Project.scala
```diff
@@ -41,18 +42,22 @@ sealed trait ProjectDefinition[PR <: ProjectReference] {
...
   def aggregate: Seq[PR]
+  private[sbt] def aggregateEval: Eval[Seq[PR]]
 
   @deprecated("Delegation between projects should be replaced by directly sharing settings.", "0.13.0")
   def delegates: Seq[PR]
+  private[sbt] def delegatesEval: Eval[Seq[PR]]
 
   /** The references to projects that are classpath dependencies of this project. */
   def dependencies: Seq[ClasspathDep[PR]]
+  def dependenciesEval: Eval[Seq[ClasspathDep[PR]]]
```
dwijnand: Could these not be in `Project` instead? And why isn't this one `private[sbt]`?

eed3si9n: That might be a mistake. I can try removing all from this trait.

eed3si9n:
Finally updated the commit on this.

dwijnand:
:+1:

eed3si9n:
Failing tests
- actions / aggregate
- actions / state
- dependency-management / delegate
- project / delegates
- project / transitive-plugins

PR #396, opened 2012-03-11
Parser for escaped and verbatim string literals

eed3si9n:
## background

see [Arguments with whitespace](https://groups.google.com/forum/#!topic/simple-build-tool/el5c3oaSgsw/discussion). Heiko wrote:

> How can I pass arguments to the run method that contain a whitespace, e.g.
> 
> ```
> run "Mark Harrah" 
> ```
## what's here

I've changed the `spaceDelimited` parser as follows:

``` scala
def spaceDelimited(display: String): Parser[Seq[String]] = (token(Space) ~> token(StringBasic, display)).* <~ SpaceClass.*
```

Instead of simple `NotSpace`, it uses `StringBasic`, which parses three different kinds of strings:

``` scala
lazy val StringBasic = StringVerbatim | StringEscapable | NotQuoted
```
### verbatim string

Verbatim string uses the same style as Scala:

```
run """foo bar C:\""" """x"""
```

Arguments are passed into the app as two arguments.
### escaped string

Escaped string uses the same style as Scala/Java. See [String Literals](http://docs.oracle.com/javase/specs/jls/se5.0/html/lexical.html#101083) for Java spec.

``` scala
run "baz\u0398\t\""
```

This passes in `bazΘ   "`.
### unquoted string

Finally, for backward compat, we have unquoted string, which begins with non-double-quotation character and ends with a space.

``` scala
run quux
```
## note

All escape sequence evaluations are done at the parser-level, so the underlying settings system does not see any escaping. Thus most of the existing tasks should be able get the benefits without changing them.
The exceptions are InputTasks and Command that currently uses `"` and `\`.

PR #2549, opened 2016-04-12
Native script support

ekrich:
Allows for a filename like `hellosbt.bat` in Windows.

```
::#!
@echo off
call xsbt -Dsbt.version=0.13.12 -Dsbt.main.class=sbt.ScriptMain --error %~nx0 %*
goto :eof
::!#

/***
name := "sbt embed"
version := "1.0"
scalaVersion := "2.11.8"
logLevel in Global := Level.Debug
*/

println("Hello, " + args.headOption.getOrElse("World") + "!")
```

Or with Unix, `hellosbt.sh`

```
#!/usr/bin/env sbt -Dsbt.version=0.13.12 -Dsbt.main.class=sbt.ScriptMain --error

/***
name := "sbt embed"
version := "1.0"
scalaVersion := "2.11.8"
logLevel in Global := Level.Debug
*/

println("Hello, " +  args.headOption.getOrElse("World") + "!")
```

### DIFF
--- a/main/src/main/scala/sbt/Script.scala
+++ b/main/src/main/scala/sbt/Script.scala
@@ -42,7 +42,7 @@ object Script {
       val append = Load.transformSettings(Load.projectScope(currentRef), currentRef.build, rootProject, scriptSettings ++ embeddedSettings)
 
       val newStructure = Load.reapply(session.original ++ append, structure)
-      val arguments = state.remainingCommands.drop(1).map(arg => "\"" + arg + "\"")
+      val arguments = state.remainingCommands.drop(1)
       val newState = arguments.mkString("run ", " ", "") :: state.copy(remainingCommands = Nil)
       Project.setProject(session, newStructure, newState)
     }
--- a/main/src/main/scala/sbt/Script.scala
+++ b/main/src/main/scala/sbt/Script.scala
@@ -13,10 +13,20 @@ object Script {
   lazy val command =
     Command.command(Name) { state =>
       val scriptArg = state.remainingCommands.headOption getOrElse sys.error("No script file specified")
-      val script = new File(scriptArg).getAbsoluteFile
-      val hash = Hash.halve(Hash.toHex(Hash(script.getAbsolutePath)))
+      val scriptFile = new File(scriptArg).getAbsoluteFile
+      val hash = Hash.halve(Hash.toHex(Hash(scriptFile.getAbsolutePath)))
       val base = new File(CommandUtil.bootDirectory(state), hash)
       IO.createDirectory(base)
+      val src = new File(base, "src_managed")
+      IO.createDirectory(src)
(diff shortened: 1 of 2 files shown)


main/src/main/scala/sbt/Script.scala
```diff
@@ -32,7 +42,7 @@ object Script {
       val append = Load.transformSettings(Load.projectScope(currentRef), currentRef.build, rootProject, scriptSettings ++ embeddedSettings)
 
       val newStructure = Load.reapply(session.original ++ append, structure)
-      val arguments = state.remainingCommands.drop(1)
+      val arguments = state.remainingCommands.drop(1).map(arg => "\"" + arg + "\"")
```
eed3si9n: What does this quoting do? Wouldn't this interfere if the existing users are already adding quotes?

ekrich: No, prior to that change, arguments with quotes and spaces did not work.
The quotes get stripped.

ekrich: I checked `run` and quoting the argument works from the sbt prompt.

```
> run "sbt World"
[info] Running Test sbt World
[info] Hello, sbt World!
```

Running from sbt requires escaping the quotes:

```
> sbt "run \"sbt World\""
[info] Running Test sbt World
[info] Hello, sbt World!
```

eed3si9n: Where are you running the second example from? If it's something like Bash, `"` gets interpreted by that, right?

eed3si9n: Here's using Zsh and sbt 0.13.9:

``` sh
$ sbt "eval \"hello\""
...
[info] ans: String = hello
```

eed3si9n: ok. So I guess this is the repro of the bug.

``` scala
#!/usr/bin/env scalas

println(args.toList)
```

Running this yields this:

```
$ script/args.scala "foo bar" "baz"
...
List(foo, bar, baz)
```

ekrich: This is on the windows command line. I was just trying different scenarios to get sbt to print an argument with a space. The `run` above is the same code as the sbt script example above. You can try seeing if you can run a script like above `hellosbt.scala` and try passing "sbt World". It should ignore the "World".

eed3si9n: I filed #2551 for this. Could you split this PR into two commits: One that adds the native file extension support, and the second that fixes the quotation issue?

ekrich: Sure, no problem thanks.

eed3si9n:
Ref #396 for the triple quoting.

eed3si9n:
Closed in favor of #2603

PR #2553, opened 2016-04-13
Simplify EvaluateSettings and INode

pauldraper:
This simplifies EvaluateSettings.

(1) Reducing states from five to four.
(2) Remove redundant or unnecessary data structures.
(3) Move task counting logic into separate class.
(4) Remove the special nature of BindNode.
(5) Liberate most nodes from unnecessary `AList` operations.
(6) Reduce the scope of synchronization.

`registerIfNew` showed up as a hotspot when I profiled with YourKit. The equivalent `start()` is now about twice as fast. Though outside of profiling, I can't see a real difference in overall times.

But if nothing else, the implementation is at least far easier to understand.

### DIFF
--- a/util/collection/src/main/scala/sbt/INode.scala
+++ b/util/collection/src/main/scala/sbt/INode.scala
@@ -1,177 +1,194 @@
 package sbt
 
-import java.lang.Runnable
-import java.util.concurrent.{ atomic, Executor, LinkedBlockingQueue }
-import atomic.{ AtomicBoolean, AtomicInteger }
-import Types.{ :+:, ConstK, Id }
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.Executor
+import Types.Id
+import scala.concurrent.duration.Duration
+import scala.concurrent.{ Await, Promise }
 
-object EvaluationState extends Enumeration {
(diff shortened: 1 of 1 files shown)


eed3si9n:
Are you keeping bincompat with sbt 0.13.0? If not this type of refactoring might be better suited for 1.0.x series.

pauldraper:
If someone extends `EvaluateSettings` or uses `EvaluationState`, binary compatibility is broken.

These are core parts of SBT implementation, and I estimate it to be very unlikely that they would be used elsewhere.

I have no problem putting this on 1.0 instead. The only issue is that any changes to this code would have to be made separately for 0.13 and 1.0.

eed3si9n:
> Though outside of profiling, I can't see a real difference in overall times.

Are you saying that the perf gain is not observable from the build user's point of view?

pauldraper:
> Are you saying that the perf gain is not observable from the build user's point of view?

I went and ran this end-to-end on a few large projects. It seems there is a difference in load times, but a small one, about 1-2%. Nothing real exciting on that front.

dwijnand:
What do either of you want to do with this PR?

eed3si9n:
The refactoring looks useful, but I think it should target 1.0.x code base.
I'm closing this for now.

Issue #2546, opened 2016-04-09
flaky compiler-project/error-in-invalidated test


eed3si9n:
## steps

Run Travis
## problem

Not sure since when but compiler-project/error-in-invalidated test seems to fail the first time it runs.

PR #2563, opened 2016-04-18
Fixes incremental compiler missing member ref from macro expansion sbt/sbt#2560

### LINKED ISSUE #2560: 0.13.11 regression: incremental compiler misses when macro expansion references another source
## steps
1. Macro expansion in `A.scala` references another source `B.scala`.
2. Change `B.scala`.
## problem

`A.scala` doesn't get compiled.
## expectation

`A.scala` gets compiled.
## notes

This can be confirmed when "member reference internal dependencies" doesn't contain the entry from `A.scala` to `B.scala`.

This was caused by 0f616294c4e713dc415f5dc3ae7aef257decb228:

``` scala
   override def traverse(tree: Tree): Unit = tree match {
     ....
      case MacroExpansionOf(original) if inspectedOriginalTrees.add(original) => traverse(original)
      case other => super.traverse(other)
    }
```

This won't traverse into an `Apply` tree, for instance if it has `MacroExpansionAttachment`. The fix is to call `super.traverse(tree)` for `MacroExpansionOf` case as well.

eed3si9n:
Fixes #2560

Unlike other scripted macro tests, the call site of the macro I added for this repro is
`Provider.tree(0)`, which itself does not introduce internal member reference.
Instead the macro expansion calls `Bar.bar(0)`. Due to #2560, the expanded
tree is not traversed, and thus the reference to `Bar` is not caught
during incremental compilation.

The fix is a one-liner change to traverse the expanded tree, which we used to do prior to 0.13.11 (0f61629).

/review @gkossakowski, @Duhemm

### DIFF
--- a/compile/interface/src/main/scala/xsbt/Dependency.scala
+++ b/compile/interface/src/main/scala/xsbt/Dependency.scala
@@ -146,7 +146,9 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile {
       // In some cases (eg. macro annotations), `typeTree.tpe` may be null. See sbt/sbt#1593 and sbt/sbt#1655.
       case typeTree: TypeTree if typeTree.tpe != null => symbolsInType(typeTree.tpe) foreach addDependency
 
-      case MacroExpansionOf(original) if inspectedOriginalTrees.add(original) => traverse(original)
+      case m @ MacroExpansionOf(original) if inspectedOriginalTrees.add(original) =>
+        traverse(original)
+        super.traverse(m)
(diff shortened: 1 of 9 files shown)


gkossakowski:
It looks like the test failure is legit.

eed3si9n:
@gkossakowski There's been failures on that test prior to this PR - #2546

gkossakowski:
LGTM

eed3si9n:
Zinc portion of forward porting: https://github.com/sbt/zinc/pull/111

Issue #2518, opened 2016-03-17
off-the-grid/offline installation of sbt


eed3si9n:
## steps

Suppose I'm behind some firewall or just no internet.
I'd like to create a file image that can launch sbt.
## problem

Prior to sbt 0.13.11 I probably could have created a file image based on all the JARs sbt will download during boot and it might have worked. With sbt 0.13.11 the compiler bridge is resolved from the repository.
## expectation

A command `install` or something that creates installation image that's runnable without hitting Ivy cache OR Ivy local.

Ideally a single JAR version of sbt that does not need to hit the Internet.

dwijnand:
Does that single JAR include the contents of sbt-launch.jar?

eed3si9n:
The single-jar might diverge too much from the current implementation, so a better option might be preloaded-local repository like that of Activator, and some way of denotation in the `launchconfig` to look at the preloaded-local, such as `preloaded-local`. See http://www.scala-sbt.org/0.13/docs/Launcher-Configuration.html

dwijnand:
That sounds like a lot of machinery...

I've been thinking recently that I'm not so sure about sbt-launcher..

In my ideal world, given an `sbt.version` in `project/build.properties` and `sbt-extras` a single uniquely-versioned fat jar of sbt is downloaded (unless already present) and then run.

fommil:
There aren't enough +1s in the world for this.

dwijnand:
Fixed in #2564

eed3si9n:
I don't think #2564 fully work.

eed3si9n:
Another idea around this is defining a local preloaded repo name "preloaded-local", "preloaded-local-ivy" or something like that on a pre-agreed location on the machine, such as `~/.sbt/preloaded-local/`.
Installer then can place sbt artifacts to the location up front.

dwijnand:
Why a new location? Why not place them in the ivy cache?

eed3si9n:
Ivy cache is opaque to us, whereas additional "preloaded" local repo is perfectly safe. See also Activator.

dwijnand:
I don't think we should duplicate artefacts more. Surely Ivy must provide some API for injecting artefacts into its cache. Either directly or indirectly.

Activator introducing yet-another-repo is a bad precedent. See ~/.ivy2/cache vs ~/.m2/repository vs ~/coursier/cache. sbt shouldn't introduce another one.

eed3si9n:
[...]

PR #2564, opened 2016-04-18
Off-the-grid installation of sbt

Duhemm:
This PR adds a new main object `sbt.Main` that implements a launcher that can start sbt without ivy or internet access.

The new command `install` creates the fat JAR that can then be launched.

Ref #2518


eed3si9n:
Notes please.
Also forward port.

k0001:
If I understand correctly, among other things this change allows me to call, for example, `sbt compile` within a SBT project (or a similar command that avoids `sbt-launcher` perhaps), and provided all the `jar`s that both `sbt` and the project itself depends on are already in place, the project will start compiling without downloading anything. Is this correct?

If this is the case: How do I tell `sbt` where to find all those `jar`s? Otherwise: Can I accomplish the example mentioned above somehow?

Sorry if this isn't the right place to ask this question!

k0001:
FYI: I'm working on improving the packaging of SBT applications in the Nix package manager (https://nixos.org/nix). Basically, I want to be able to call `sbt compile` (or any other command that would achieve the same thing), without having SBT hit the internet. One valid approach seems to be to create a local ivy cache, but that's really cumbersome to do as you have to recreate the ivy repository structure and have the `ivy.xml` files. Instead, I would like to be able to somehow list where to find all of the `jar`s both SBT and the project itself needs by hand, in a `CLASSPATH`-like fashion, and it looks like this pull request tries to solve some of that. Am I right? Thanks!

eed3si9n:
This was an experimental feature, and it hasn't been given enough attention.
I wrote https://github.com/sbt/sbt-export-repo, which lets you create a preloaded local repository correctly, by driving Ivy.

PR #2565, opened 2016-04-18
Adding sleep in scripted test to make sure timestamp bumps

### LINKED ISSUE #2546: flaky compiler-project/error-in-invalidated test

eed3si9n:
Fixes #2546. Ref #958/#965

`scripted compiler-project/error-in-invalidated` has been failing
frequently on Travis CI. It seems like incremental compiler is not
catching the change in source for `changes/A2.scala`.

Perhaps the machines are getting fast enough that we are doing three compilations within 1s?

/review @gkossakowski, @Duhemm

### DIFF
--- a/sbt/src/sbt-test/compiler-project/error-in-invalidated/build.sbt
+++ b/sbt/src/sbt-test/compiler-project/error-in-invalidated/build.sbt
@@ -1,1 +1,5 @@
-incOptions := sbt.inc.IncOptions.Default
+lazy val root = (project in file(".")).
+  settings(
+    incOptions := sbt.inc.IncOptions.Default,
+    scalaVersion := "2.11.7"
+  )
--- a/sbt/src/sbt-test/compiler-project/error-in-invalidated/test
+++ b/sbt/src/sbt-test/compiler-project/error-in-invalidated/test
@@ -1,9 +1,13 @@
 > compile
+
 # comment out `initialized` method in A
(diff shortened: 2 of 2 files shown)


smarter:
In https://github.com/sbt/sbt/pull/2523 I've been seeing flaky failures in the new test `types-in-used-names-b`, I wonder if that's the same issue. If so, it sure sounds like we need a better solution than randomly adding sleeps until the tests pass.

eed3si9n:
Assuming that the underlying issue is the change detection via last modified date, two alternatives I can think of are:
1. checking hash of files to look for changes. (what's the overhead on calculating 1000s of hashes)
2. Java 7 file watching

eed3si9n:
I wonder if we can treat timestamp == timestamp case to be invalidated. This might also require another `$ sleep 1000` for the reserve reason. For example, to test incremental compiler not doing additional compilation, we'd need to wait 1s.

gkossakowski:
I have never seen this problem on my Mac. Is time resolution on Travis filesystem lower?

gkossakowski:
@smarter, did you encounter the problem you describe on your local machine?

smarter:
I didn't manage to get `types-in-used-names-b` to fail locally, but it still fails occasionally on travis with `sleep 1000` so that's not it.

Issue #2539, opened 2016-04-05
Host machine's Ivy cache affects scripted


eed3si9n:
## steps

Remove compiled compiler bridge from my machine's Ivy cache.
## problem

scripted fails.
## expectation

scripted should not depend on my machine's Ivy cache for the compiler bridge.
## notes

This prevents me to test compiler bridge change quickly.

Compiler.scala line 99

``` scala
  @deprecated("Use `scalaCompiler(ScalaInstance, ClasspathOptions, IvyConfiguration, ModuleID)`.", "0.13.10")
  def scalaCompiler(instance: ScalaInstance, cpOptions: ClasspathOptions)(implicit app: AppConfiguration, log: Logger): AnalyzingCompiler =
    {
      val launcher = app.provider.scalaProvider.launcher
      val componentManager = new ComponentManager(launcher.globalLock, app.provider.components, Option(launcher.ivyHome), log)
      val provider = ComponentCompiler.interfaceProvider(componentManager)
      new AnalyzingCompiler(instance, provider, cpOptions)
    }
  def scalaCompiler(instance: ScalaInstance, cpOptions: ClasspathOptions, ivyConfiguration: IvyConfiguration, sourcesModule: ModuleID)(implicit app: AppConfiguration, log: Logger): AnalyzingCompiler =
    {
      val launcher = app.provider.scalaProvider.launcher
      val componentManager = new ComponentManager(launcher.globalLock, app.provider.components, Option(launcher.ivyHome), log)
      val provider = ComponentCompiler.interfaceProvider(componentManager, ivyConfiguration, sourcesModule)
      new AnalyzingCompiler(instance, provider, cpOptions)
    }
```

PR #2567, opened 2016-04-20
Change ivy home when running scripted tests

Duhemm:
This commit changes the build so that when running `scripted`:
- sbt is published to a local repository in a different ivy home. We
  publish using the ivy home `/path/to/sbt/sbt/target/scripted-ivy`
  (instead of `~/.ivy2`).
- this custom ivy home is explicitly specified at when sbt is launcher.

Fixes sbt/sbt#2539


dwijnand:
Superseded by https://github.com/sbt/zinc/pull/215.

PR #2575, opened 2016-04-25
[sbt 1.0] Rename sbt.Plugin => sbt.internal.OldPlugin

eed3si9n:
I don't think we need sbt.Plugin any more.


dwijnand:
:+1:

PR #2576, opened 2016-04-26
Support repositories override in static launcher

Duhemm:


dwijnand:
Do we still want this in 0.13? It's been merged into sbt/librarymanagement#34 for 1.0.

eed3si9n:
Probably not in sbt 0.13.

PR #2577, opened 2016-04-29
 [sbt 1.0] Bump to Scala 2.11.8

eed3si9n:

PR #2586, opened 2016-05-04
[sbt 1.0] We shouldn't cross publish as sbt_2.11 for backward compatibility

eed3si9n:
By keeping sbt artifact to be called sbt (as opposed to `sbt_2.11`), the user can try by just
changing the build properties.

Ref #2577

/review @Duhemm, @dwijnand

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -196,6 +196,7 @@ lazy val sbtProj = (project in sbtPath).
     baseSettings,
     name := "sbt",
     normalizedName := "sbt",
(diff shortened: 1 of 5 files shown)


Duhemm:
LGTM

dwijnand:
I'm kind of torn here. But LGTM. The SBT modules are/will stay binary cross versioned right?

eed3si9n:
> The SBT modules are/will stay binary cross versioned right?

Yes.

PR #2589, opened 2016-05-05
[sbt 1.0] Bumping up to Zinc 1.0.0-M1

eed3si9n:
Bumping up to new modules.
- Some changes to the DSL organization according to http://www.scala-sbt.org/0.13/docs/Coding-Guideline.html
- Minor adjustment to class discovery code - 738e8fb2f395f79364cb450e97c50d0db0331921

/review @Duhemm, @dwijnand


MIGRATION.md
```diff
@@ -3,3 +3,9 @@ Migration notes
 
 - `project/Build.scala` style based on `sbt.Build` is removed. Migrate to `build.sbt`.
 - `Project(...)` constructor is limited to just two parameters.
+
+- change import in auto plugin to:
+
+```scala
+import sbt._, syntax._, Keys._
```
Duhemm, line 10: Should we advertise the existence of `sbt.Import`? That's what we use to keep the source compatibility of build definitions, but I guess we don't want to keep it around.

eed3si9n, line 10: `Import` is now a trait, and I'm mixing that into the package object.

gkossakowski:
Zinc dropped support for Scala 2.8 and 2.9 which causes this test failure:

```
[info] [info] 'compiler-bridge_2.10' not yet compiled for Scala 2.8.1.final. Compiling...
[error] /tmp/sbt_ce7f5e4/xsbt/API.scala:43: error: ';' expected but string literal found.
[error]           s"$className:\n\t${names.mkString(", ")}"
[error]            ^
[error] /tmp/sbt_ce7f5e4/xsbt/Analyzer.scala:33: error: ')' expected but string literal found.
[error]               assert(sym.isClass, s"${sym.fullName} is not a class")
```

main/actions/src/main/scala/sbt/DotGraph.scala
```diff
@@ -30,7 +30,7 @@ object DotGraph {
   def apply(relations: Relations, outputDir: File, sourceToString: File => String, externalToString: File => String): Unit = {
     def file(name: String) = new File(outputDir, name)
     IO.createDirectory(outputDir)
-    generateGraph(file("int-source-deps"), "dependencies", relations.internalSrcDep, sourceToString, sourceToString)
+    generateGraph(file("int-source-deps"), "dependencies", relations.internalClassDep, identity[String], identity[String])
```
gkossakowski: This should be renamed to `int-class-deps`.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -508,18 +509,18 @@ object Defaults extends BuildCommon {
       (cp, s) =>
         val ans: Seq[Analysis] = cp.flatMap(_.metadata get Keys.analysis) map { case a0: Analysis => a0 }
         val succeeded = TestStatus.read(succeededFile(s.cacheDirectory))
-        val stamps = collection.mutable.Map.empty[File, Long]
+        val stamps = collection.mutable.Map.empty[String, Long]
```
dwijnand, line 512: Is this change expected? `File` isn't the most impressive of types, but it still beats `String` in my book.

gkossakowski, line 512: Stamps were tracked per source file but now are tracked per class (that is named by a String). Tracking stamps by class name is more natural when everything else is tracked by class name (as it is in zinc).

dwijnand, line 512: I see, thanks. But that's not intuitive. I'd advise to consider a ClassName type.

dwijnand:
Not sure how I feel about having to import the second half of things with `syntax`, but LGTM.

eed3si9n:
There seems to be legitimate problem around nested auto plugin discovery.. I get `java.lang.ClassNotFoundException: example.Imports.B$` when the plugin looks like

``` scala
package example

object Imports {
  object A extends AutoPlugin
}
```

while the classes directory contains `/private/var/folders/7g/l96y5q310h90xsz109qqjpmc0000gn/T/sbt_dc68f035/auto-plugins/project/target/scala-2.11/sbt-1.0.0-SNAPSHOT/classes/example/Imports$B$.class`

Ref sbt/zinc#116?

main/actions/src/main/scala/sbt/Tests.scala
```diff
@@ -273,7 +273,19 @@ object Tests {
   def discover(frameworks: Seq[Framework], analysis: CompileAnalysis, log: Logger): (Seq[TestDefinition], Set[String]) =
     discover(frameworks flatMap TestFramework.getFingerprints, allDefs(analysis), log)
 
-  def allDefs(analysis: CompileAnalysis) = analysis match { case analysis: Analysis => analysis.apis.internal.values.flatMap(_.api.definitions).toSeq }
+  def allDefs(analysis: CompileAnalysis) = analysis match {
```
gkossakowski, line 276: This method used to return top level classes and objects but now returns also nested and inherited. Is that change intentional?

Have considered using the following definition?

``` scala
def allDefs(analysis: CompileAnalysis): Seq[xsbti.api.ClassLike] = {
  def companionsApis(c: xsbti.api.Companions) = Seq(c.classApi, c.objectApi)
  analysis match {
    case analysis: Analysis =>
      analysis.apis.internal.values.flatMap(x => companionsApis(x.api)).toSeq

  }
}
```

eed3si9n, line 276: sbt 0.13 is able to detect auto plugins nested in another object.

gkossakowski, line 276: Everywhere in `PluginDiscovery` different methods explicitly talk about top-level modules. Do you have a sense how plugins declared in another object are discovered?

PR #2615, opened 2016-05-13
Add Windows script support and native file extensions on Unix platforms

ekrich:
When running a sbt script, this change lets the user on UNIX and
Windows platforms to use native file extensions like none/.sh or
.bat/.cmd. The code copies the file to the sbt boot/hash/src_managed
directory with a .scala extension.


ekrich:
Forward port from 0.13 to 1.0.x

eed3si9n:
Thanks!

PR #2621, opened 2016-05-17
tryWindowsCompatibility in implicit conversions of ProcessExtras

acdenhartog:
I have implemented a rather stupid method that tries to make commands that _do already work_ on Windows actually work without modifying the build scripts.

It attempts to find an existing .bat/.exe alternative to a given command in the same folder, if that fails it passes the command to cmd.exe which will do whatever it can; likely delegating to whatever programs are installed within Windows.

I do not claim that this method is pretty, it most assuredly is not, but I do claim that it is useful and could save time in having to deal with Windows specific issues.

The example that motivated this pull request: .sh files will run on Windows if Git is installed; the main Scala repository is currently using such as shell script in their partest, and with "cmd /c " appended to the command it just works.

(also note that ProcessExtra is a separate file in sbt 1.0.x, so in that context this horrible hack is hidden away and alone somewhere instead of looking like pollution on top of your otherwise pretty code base)

### DIFF
--- a/util/process/src/main/scala/sbt/Process.scala
+++ b/util/process/src/main/scala/sbt/Process.scala
@@ -11,14 +11,70 @@ import java.net.URL
 trait ProcessExtra {
   import Process._
   implicit def builderToProcess(builder: JProcessBuilder): ProcessBuilder = apply(builder)
-  implicit def fileToProcess(file: File): FilePartialBuilder = apply(file)
+//implicit def fileToProcess(file: File): FilePartialBuilder = apply(file)
   implicit def urlToProcess(url: URL): URLPartialBuilder = apply(url)
   @deprecated("Use string interpolation", "0.13.0")
   implicit def xmlToProcess(command: scala.xml.Elem): ProcessBuilder = apply(command)
   implicit def buildersToProcess[T](builders: Seq[T])(implicit convert: T => SourcePartialBuilder): Seq[SourcePartialBuilder] = applySeq(builders)
 
-  implicit def stringToProcess(command: String): ProcessBuilder = apply(command)
-  implicit def stringSeqToProcess(command: Seq[String]): ProcessBuilder = apply(command)
+//implicit def stringToProcess(command: String): ProcessBuilder = apply(command)
+//implicit def stringSeqToProcess(command: Seq[String]): ProcessBuilder = apply(command)
+  
+  implicit def fileToProcess(file: File): FilePartialBuilder =
+    if(onWindows) apply(new File(windowsAlternativeToFile(file).getOrElse("cmd /c "+file))) else apply(file)
+  implicit def stringToProcess(command: String): ProcessBuilder = apply(tryWindowsCompatibility(command))
+  implicit def stringSeqToProcess(command: Seq[String]): ProcessBuilder = apply(tryWindowsCompatibilitySeq(command))
+
+  lazy val onWindows = sys.props("os.name").startsWith("Windows")
+  
+  def tryWindowsCompatibilitySeq(command:Seq[String]):Seq[String] =
+    tryWindowsCompatibility(command.head) +: command.tail
+
+  /** Looks for a Windows executable with the same file name in the same folder.
+    * @param actualFile which is actually a file that exists.
+    * @return optionally path to a Windows executable.
+    */
+  def windowsAlternativeToFile(actualFile:File):Option[String] = {
+    val withoutExtension = actualFile.getName.takeWhile(_ != '.') //remove extension if there is one
+    val sameDirectory = actualFile.getParentFile
+    val extensionsOfSameNameFiles =
+      sameDirectory.listFiles()
+        .filter((f: File) => f.getName.startsWith(withoutExtension)) //only files of same name
+        .map((f: File) => f.getName.drop(withoutExtension.length)) //only their extensions
+    extensionsOfSameNameFiles.find(_ == ".bat") //batch files take precedence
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:
Hi @acdenhartog,

eed3si9n:
First, thanks for the contribution. I am a bit concerned about the changes tho.
The fact that implicit changes the behavior from the normal invocation doesn't make sense to me. If it's good enough for implicits, why not change the actual behavior of `Process(...)`.
Also, doesn't this silently change the behavior of the build when ppl upgrade to 0.13.12? As you know, Process was donated to the standard library, so I'm reluctant to diverge the behavior in 0.13.

I think a better course of action might be to experiment this first using a plugin. See for example, sbt-doge that I experimented with cross building behavior. That got merged into sbt 1.0.x after sbt and Play trying it out for a while.

acdenhartog:
@eed3si9n Hi, first of all some context: I was getting a bit annoyed by Perry on the Gitter and his dismissive way of talking about users needing to manual deal with Windows compatibility. The code in this RFC was rushed and not entirely serious; I just wanted to make the point that it is possible and saying "that is all the fault of CreateProcess" is not a sufficient argument.

One can just append `"cmd","/C "` to get Windows compatibility in many cases, and I think platform independence and time saving should be a major goal of SBT.

Unfortunately, on further reading I can understand a bit better where Perry was coming from.

First of all let us put aside that `.bin/.exe` lookup hack which is probably a bad idea anyway as it could lead to unintended behaviors.

But what I do think ought to be dealt with is the ["issues caused by changes to Runtime.exec"](https://blogs.oracle.com/thejavatutorials/entry/changes_to_runtime_exec_problems). Java used to provide more "implicit cmd.exe calls" and I really want those back if there is going to be shell scripts called in build files.

Windows operating systems have a tiny program called Run Command in their Start Menu. I expected `"command"!` to work exactly the same as Run Command. But currently that is not the case, and this leads to both confusion and needless incompatibility between platforms.

So perhaps what should be done is figure out how to make Process work as expected. I have taken a look at the actual implementation and it seems that eventually everything [gets passed to JProcessBuilder](https://github.com/scala/scala/blob/2.12.x/src/library/scala/sys/process/Process.scala#L104) which on Widows delegates to [CreateProcess](https://msdn.microsoft.com/en-us/ms682425).

On searching about this stuff a bit more I ran into this blog post from a few months ago: [Java and Command Line Injections in Windows](http://codewhitesec.blogspot.nl/2016/02/java-and-command-line-injections-in-windows.html) so clearly the underlying JProcessBuilder is not secure, even now that ambiguities from most implicit cmd calls have been removed.

Having added such easy access too the command line in all of Scala, I think it ought to be safe...

Perhaps Oracle will address these things, but the code referred to in that blog post still exists in the most recent Java API. Perhaps that method in Process of the Scala standard library should be modified such that on Windows it:
- Can add an extra layer of security to help avoid command line injections.
- Provide the implict `"cmd","/C"` so we can be more easily platform independent.

Although implicitly getting cmd to interpret things and avoiding injection do seem like contradictory goals.
_edit:_ actually looking over [this blog post](https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/#comments) referred to in the one above it seems doable.

Another possibly interesting alternative is to let programmers specify that on Windows all commands from Process are delegated to sh.exe (Git bash or similar), that adds a dependency but it should further remove compatibility barriers.

Or maybe we should all be using the Ammonite-Shell instead. Could that be an SBT plugin? [There is an SBT plugin to get the REPL](https://github.com/alexarchambault/sbt-ammonite). Maybe there needs to be one to get Ammonite code directly into build files as well. How would one approach that? I know how to import dependencies for projects, and SBT plugins that provide settings, but not generic Scala imports for SBT itself...
_edit:_ perhaps would be better to call into Ammonite scripts instead of embedding Ammonite code directly into a build file, but even then it would be nicer to just import the script instead of hitting that same problem of invocation it as a process.

Anyway, I am not sure what to conclude from all of this.

eed3si9n:
I was thinking this whole thing sounds familiar then I remembered:
- https://github.com/sbt/sbt-sh
- https://github.com/foundweekends/conscript/blob/21724b03c4615dd77b56a0d914d933a2d460fb16/src/main/scala/apply.scala#L12-L18

In other words, why not provide something that on Windows wraps using `cmd`, and on Mac and Linux uses `/bin/sh`? As per the security concern, I am not sure what the attack vector we are worried about is since this is not a web app. If someone can change the build script, you can pretty much do anything on the file system.

Can we close this PR and restart a discussion on sbt-dev list, or a more focused issue?

acdenhartog:
And conscript of course does exactly what that microsoft blog says not to:
_Do Not:
1. Simply add quotes around command line argument arguments without any further processing.
2. Allow cmd to ever see an unescaped `"` character._

> I am not sure what the attack vector we are worried about is since this is not a web app.

For SBT it is indeed not a concern. However, that Process is now part of the standard library. (although it is not compatible compatible with Scala.js; that would be something) And I suppose security of JProcessBuilder is really Oracle's responsibility. meh.

I think instead we should just discourage people form using bash/batch scripts in their builds as much as possible. The file system part of Ammonite is at least is Windows compatible now.

Also I just found out about deprecating `Build.scala`. I renamed a `build.scala` to `build.sbt` and eventually, after seeing it not working but not complaining much either, remembered how to import into SBT. :neutral_face::palm_tree:

(will submit a completely different issue regarding `Build.scala` rewrite soon-ish)

You may close this PR.

Issue #2620, opened 2016-05-17
cached resolution: Configuration regression


eed3si9n:
## steps

We don't have minimized steps yet, but in some build you run `update` task using nightly build of sbt 0.13.12.
## problem

It tries to add configuration `it-internal` into a subproject `B` that's transitively depended by subproject `A`.
## expectation

Works.
## notes

Likely something related to https://github.com/sbt/sbt/pull/2513.

Ref #2435 /cc @Duhemm

Duhemm:
Reproduction:

`build.sbt`:

``` scala
def makeProject(name: String): Project =
  Project(name, file(name))
    .enablePlugins(PlayScala)
    .settings(updateOptions := updateOptions.value.withCachedResolution(true))

lazy val p0 = makeProject("p0")
lazy val p1 = makeProject("p1")
lazy val p2 = makeProject("p2").dependsOn(p0, p1)
```

`project/play.sbt`:

``` scala
addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.4.3")
```

eed3si9n:
Try-catch workaround (#2628) is now merged to 0.13.

eed3si9n:
I have a feeling that these issues might be related to https://github.com/sbt/librarymanagement/issues/45

eed3si9n:
If try-catch works, I'm going to close this for now.

PR #2628, opened 2016-05-24
Workaround for sbt/sbt#2620

Duhemm:


eed3si9n:
LGTM
Please forward port

PR #2630, opened 2016-05-25
Loading performance

eed3si9n:
This is a result of spending a day focusing on sbt's loading performance.

A very large build that I'm looking at takes 43s to load using sbt 0.13.11, and I was able to shave off 14s off of that with simple caching. Most of the changes are putting in debug logs.

/review @Duhemm, @dwijnand


main/src/main/scala/sbt/Load.scala
```diff
@@ -288,13 +307,16 @@ object Load {
...
+      val build = (info: BuildLoader.BuildInfo) => Some(() =>
+        loadUnit(info.uri, info.base, info.state, info.config)
+      )
       val components = BuildLoader.components(resolver, build, full = BuildLoader.componentLoader)
       BuildLoader(components, fail, s, config)
     }
   def load(file: File, loaders: BuildLoader, extra: List[URI]): sbt.PartBuild = loadURI(IO.directoryURI(file), loaders, extra)
   def loadURI(uri: URI, loaders: BuildLoader, extra: List[URI]): sbt.PartBuild =
     {
+      val log = loaders.config.log
```
Duhemm: I think that `log` is unused here

main/src/main/scala/sbt/Load.scala
```diff
@@ -648,54 +676,66 @@ object Load {
...
+            for {
+              file <- files
+              config <- (memoSettings get file).toSeq
+              setting <- config.settings
+            } yield setting
+          }
+        // Expand the AddSettings instance into a real Seq[Setting[_]] we'll use on the project
+        def expandSettings(auto: AddSettings): Seq[Setting[_]] = auto match {
+          case BuildScalaFiles     => rawProject.settings
+          case User                => globalUserSettings.cachedProjectLoaded(loadedPlugins.loader)
```
Duhemm, line 724: Here the value of `loadedPlugins.loader` will only be used on the first call to `cachedProjectLoaded`. Could this be a problem?

main/src/main/scala/sbt/Plugins.scala
```diff
@@ -345,4 +347,13 @@ ${listConflicts(conflicting)}""")
...
   }
+
+  /** Debugging method to time how long it takes to run various compilation tasks. */
+  private[this] def timed[T](label: String, log: Logger)(t: => T): T = {
+    val start = System.nanoTime
+    val result = t
+    val elapsed = System.nanoTime - start
+    log.debug(label + " took " + (elapsed / 1e6) + " ms")
+    result
+  }
```
Duhemm, line 358: This brings to 5 the number of `timed` functions. Maybe that we should just define it `private[sbt]` somewhere

main/src/main/scala/sbt/Load.scala
eed3si9n, line 724: The correctness hinges on the fact that it will be ok. As I see it the only place where you see `projectLoaded` gets changed is in `loadGlobalSettings` (https://github.com/sbt/sbt/pull/2630/files/c67b3881de6287c7c970b1bf5a15c9c1739845ef#diff-1b5b6f2f3765a38e3ca6b130a15076ffL72). I am guessing that the output of `buildGlobalSettings` is stable across subprojects. Am I wrong here?

main/src/main/scala/sbt/Load.scala
```diff
@@ -933,7 +973,11 @@ object Load {
   val LoadBuildConfiguration = sbt.LoadBuildConfiguration
 
   final class EvaluatedConfigurations(val eval: Eval, val settings: Seq[Setting[_]])
-  final case class InjectSettings(global: Seq[Setting[_]], project: Seq[Setting[_]], projectLoaded: ClassLoader => Seq[Setting[_]])
+  final case class InjectSettings(global: Seq[Setting[_]], project: Seq[Setting[_]], projectLoaded: ClassLoader => Seq[Setting[_]]) {
+    private val cache: mutable.Map[Unit, Seq[Setting[_]]] = mutable.Map.empty
+    def cachedProjectLoaded(cl: ClassLoader): Seq[Setting[_]] =
+      cache.getOrElseUpdate((), projectLoaded(cl))
```
dwijnand: What @Duhemm said above: this doesn't take the input (ClassLoader) into account for subsequent calls. Is this not a problem?

main/src/main/scala/sbt/Plugins.scala
eed3si9n, line 358: We need to profile the difference, but the tricky part about `timed` is that it appears along the call graph at almost random places (megamorphic?) that JIT won't be able to inline them. It might actually be better if we turn it into a macro.

main/src/main/scala/sbt/Load.scala
eed3si9n: https://github.com/sbt/sbt/pull/2630/files/c67b3881de6287c7c970b1bf5a15c9c1739845ef#r64524986

main/src/main/scala/sbt/Load.scala
dwijnand, line 724: What if the cache were `mutable.Map[ClassLoader, Seq[Setting[_]]]`?

eed3si9n, line 724: How's this? - f9b1583251de58d87f1d0de3c898402f234b6df2

dwijnand, line 724: A few thoughts:
1. `java.net.URL`'s hashCode is (insanely) a blocking, synchronized, dns-resolving operation, making it (IMHO) a bad data-type to use in general, but especially if used as the (or a part of) the key of any hash-based collection (map/set/etc).
2. I don't think it's universally correct to define a cache of a `java.net.URLClassLoader`-using-function in terms of the hashset of its urls as I think the ordering is important.
3. I also don't think it's universally correct to define the cache solely in terms of its URLs all together, as I think there are other properties of `java.net.URLClassLoader` that affect it's behaviour.
4. Finally, this is trading memory for speed, and I wonder how it impacts multi-project builds that cross-build for Scala.js, where there are already memory struggles. (eg. shapeless, circe, cats)

I say "universally" because this is just from a tunnel vision review of the change, not in terms of sbt's use here or the real-world situation behind this change.

Also note that `URLClassLoader` doesn't define `hashCode` at all, falling back on `java.lang.Object` generic (identity-based?) hashCode.

dwijnand, line 724: One potential avenue here is to hide the use of a cache behind a feature-flag.

eed3si9n, line 724: Updated to 6a0f5dd47111ffd883c77e4c2103539eae68d443 per review by @jsuereth on Gitter - https://gitter.im/sbt/sbt-dev?at=5745d8d280352f204df2099a

dwijnand:
LGTM

main/src/main/scala/sbt/Plugins.scala
jsuereth, line 358: +1 on macro

dwijnand:
This needs forward-porting (noticed when forward-porting #2717).

PR #2633, opened 2016-05-30
Add new SBT global setting asciiGraphWidth

RomanIakovlev:
This setting controls the maximum width of the ASCII graphs printed
by commands like `inspect tree`. Default value corresponds to the
previously hardcoded value of 40 characters.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 target/
 __pycache__
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 target/
 __pycache__
+.idea
--- a/main/src/main/scala/sbt/internal/SettingGraph.scala
+++ b/main/src/main/scala/sbt/internal/SettingGraph.scala
@@ -14,7 +14,7 @@ import Predef.{ any2stringadd => _, _ }
 import sbt.io.IO
 
 object SettingGraph {
-  def apply(structure: BuildStructure, basedir: File, scoped: ScopedKey[_], generation: Int, graphMaxWidth: Int)(implicit display: Show[ScopedKey[_]]): SettingGraph =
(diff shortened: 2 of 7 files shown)


main/src/main/scala/sbt/internal/SettingGraph.scala
```diff
@@ -51,7 +53,8 @@ case class SettingGraph(name: String,
 
   def dependsAscii: String = Graph.toAscii(this,
```
dwijnand: I would pass the value here rather than in the case class constructor.

.gitignore
```diff
@@ -1,2 +1,3 @@
 target/
 __pycache__
+.idea
```
dwijnand: This is an un-related change.

main/src/main/scala/sbt/internal/SettingGraph.scala
```diff
@@ -60,8 +63,7 @@ object Graph {
   // [info]   | +-baz
   // [info]   |
   // [info]   +-quux
-  def toAscii[A](top: A, children: A => Seq[A], display: A => String): String = {
-    val defaultWidth = 40
+  def toAscii[A](top: A, children: A => Seq[A], display: A => String, defaultWidth: Int): String = {
     // TODO: Fix JLine
     val maxColumn = math.max( /*JLine.usingTerminal(_.getWidth)*/ 0, defaultWidth) - 8
```
eed3si9n, line 66: This is not your change, but we should fix this one too from c981abd295fbb8a93d712ccefc46027772f7f18a /cc @Duhemm

RomanIakovlev, line 66: @eed3si9n what exactly has to be done here for the JLine? I see this code is used in 0.13 branch, and it seems to work well there (in my limited testing).

eed3si9n: Not sure why that was commented out.

eed3si9n:
Merged in https://github.com/sbt/sbt/pull/2941

PR #2634, opened 2016-05-30
Override scala organization and version transitively at the Ivy level

milessabin:
This is a first cut at a PR intended to elicit feedback.

The primary motivation for this change is to allow an alternative Scala organization to be used in builds which have dependencies which themselves have direct dependencies on Scala distribution artefacts (eg. scala-reflect, scala-compiler). As well as supporting alternative Scala organizations I believe that this change also addresses #2286 and some of the concerns raised in the discussion on that ticket.

The change is to use an Ivy `DependencyDescriptorMediator` to rewrite all direct and transitive dependencies on scala-library, scala-compiler, scala-reflect, scala-actors,  and scalap to have the same organization (either specified by the `scalaOrganization` setting or defaulting to `org.scala-lang`) and version (as specified by `scalaVersion`).

Issues to be resolved before merging,
- Although this behaviour should be the default, it should probably be made configurable. How should such an option be exposed? If an attribute of `UpdateOptions`, how would that be made available to `IvyScala`?
- How should this be tested? Locally I have been able to verify that it behaves as intended by running the included scripted test. However this relies on a local build of Scala with the organization set to "com.milessabin". Should this be done via a mock repository? Is there an example that I could base tests on in the tree?

PR #2635, opened 2016-06-03
Override scala organization and version transitively at the Ivy level

milessabin:
This is a forward port of the parts of #2634 which aren't included in https://github.com/sbt/librarymanagement/pull/44. Please note that the scripted test `dependency-management/scala-organization` will fail until the latter is merged.


milessabin:
I think Travis should go green now that sbt/librarymanagement#44 has been merged ... could you rerun the build?

eed3si9n:
I published librarymanagement 0.1.0-M12 to Sonatype (https://oss.sonatype.org/content/repositories/public/org/scala-sbt/librarymanagement_2.11/0.1.0-M12/). When it syncs to Maven Central, could you bump the dependencies?

milessabin:
Done.

milessabin:
I don't think the remaining test failure is my doing ... thoughts?

eed3si9n:
Probably not.. I'm going to rerun the Travis job for the failing one.

Issue #2614, opened 2016-05-13
Warning on transitive macro recompilation


eed3si9n:
## steps
1. Transitive macro definition causes recompilation by default.
## problem

This can end up invalidating a lot of code.
## expectation

Print warning message during compilation, so people are aware of why suddenly incremental compiler is recompiling everything.

dwijnand:
This isn't in the default branch (1.0.x), but it is in 0.13.12, which is what the issue is targeting. So I'm going to close it given #2637 was merged, and it's in 0.13.12-RC1.

eed3si9n:
Don't we have to forward port?

dwijnand:
Yes. Added to #2352. Good shout.

PR #2637, opened 2016-06-07
Inform the build user about macro recompilation

eed3si9n:
Ref #2614

Since RecompileOnMacroDef is enabled by default, let the build user know that incremental compiler just invalidate a bunch of stuff because of macro.
### sample

```
hello-world> compile
[info] Compiling 1 Scala source to /Users/eugene/work/helloworld/target/scala-2.11/classes...
[info] Because /Users/eugene/work/helloworld/src/main/scala/A.scala contains a macro definition, the following dependencies are invalidated unconditionally:
[info]  /Users/eugene/work/helloworld/src/main/scala/D.scala
[info] Compiling 1 Scala source to /Users/eugene/work/helloworld/target/scala-2.11/classes...
```

/review @Duhemm, @gkossakowski

### DIFF
--- a/compile/inc/src/main/scala/sbt/inc/MemberRefInvalidator.scala
+++ b/compile/inc/src/main/scala/sbt/inc/MemberRefInvalidator.scala
@@ -53,7 +53,7 @@ import xsbt.api.APIUtil
 private[inc] class MemberRefInvalidator(log: Logger) {
   def get[T](memberRef: Relation[File, T], usedNames: Relation[File, String], apiChange: APIChange[_]): T => Set[File] = apiChange match {
     case _: APIChangeDueToMacroDefinition[_] =>
-      new InvalidateUnconditionally(memberRef)
+      new InvalidateDueToMacroDefinition(memberRef)
     case NamesChange(_, modifiedNames) if modifiedNames.implicitNames.nonEmpty =>
       new InvalidateUnconditionally(memberRef)
     case NamesChange(modifiedSrcFile, modifiedNames) =>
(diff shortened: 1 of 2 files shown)


gkossakowski:
The logic in this PR logs just the file that contains a macro but I thought you wanted to log all files that get recompiled due to dependency on a file that contains a macro. The latter lets you report what's the actual impact of dependency on a macro: whether this leads to big recompilations or not. Isn't it this that you want to report to users?

eed3si9n:
@gkossakowski Nice catch. I moved the logging location to `MemberRefInvalidator`, and this behaves much better.

compile/inc/src/main/scala/sbt/inc/MemberRefInvalidator.scala
```diff
@@ -79,12 +79,19 @@ private[inc] class MemberRefInvalidator(log: Logger) {
...
 
-  private class InvalidateUnconditionally[T](memberRef: Relation[File, T]) extends (T => Set[File]) {
+  private class InvalidateUnconditionally[T](memberRef: Relation[File, T], apiChange: APIChange[_]) extends (T => Set[File]) {
     def apply(from: T): Set[File] = {
       val invalidated = memberRef.reverse(from)
-      if (invalidated.nonEmpty)
-        log.debug(s"The following member ref dependencies of $from are invalidated:\n" +
-          formatInvalidated(invalidated))
+      if (invalidated.nonEmpty) {
+        apiChange match {
```
gkossakowski: Isn't it better to introduce `InvalidateDueToMacroDefinition` class and put special logging logic there? This way you can let virtual dispatch do the job instead of introducing a pattern match. Also, this way you pattern match on APIChange just once: in `MemberRefInvalidator.get` method.

eed3si9n: Sure.

eed3si9n: Added InvalidateDueToMacroDefinition

gkossakowski:
LGTM

milessabin:
This is creating huge amounts of build noise for shapeless (and presumably will for other macro defining projects as well). Can we have it off by default?

PR #2646, opened 2016-06-19
Cleanups

dwijnand:


dwijnand:
@eed3si9n Can you read Codacy? How/where am I adding +1186 complexity?

eed3si9n:
@dwijnand Under the Files tab you see a bunch of stuff there. I don't think it's you who introduced the issues, so I think we can ignore Codacy.

dwijnand:
Ok, well I'll be working on the Codacy issues too.

PR #2652, opened 2016-06-29
Default compiler plugins to fully cross versioned

dwijnand:
The compiler API is neither source nor binary compatible between minor versions, so this is always required.

Except for ancient continuations, which used to be part of scala and therefore versioned with it.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2014,7 +2014,7 @@ trait BuildExtra extends BuildCommon with DefExtra {
 
   /** Transforms `dependency` to be in the auto-compiler plugin configuration. */
   def compilerPlugin(dependency: ModuleID): ModuleID =
-    dependency.copy(configurations = Some("plugin->default(compile)"))
+    dependency.copy(configurations = Some("plugin->default(compile)")) cross CrossVersion.full
 
   /** Adds `dependency` to `libraryDependencies` in the auto-compiler plugin configuration. */
   def addCompilerPlugin(dependency: ModuleID): Setting[Seq[ModuleID]] =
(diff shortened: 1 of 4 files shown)


eed3si9n:
Doesn't this kind of depend on how the compiler plugins are published out in the wild?

dwijnand:
Yeah, but I think it's a better default to bias to fully cross versioned instead of disabled.

dwijnand:
Hmm... maybe not. From https://github.com/search?q=addCompilerPlugin&type=Code&utf8=%E2%9C%93 it looks like lots of compiler plugins aren't..

I guess given that evidence we should leave it as is.

sjrd:
Too bad ... because that would make plugin authors realize that they should fully cross-publish their compiler plugins.

dwijnand:
Yeah it would, but is it worth the cost of everything else having to migrate given the new default?

For instance every build using kind-projector would have to suddenly switch to specifying cross-binary: https://github.com/non/kind-projector#using-the-plugin

eed3si9n:
I'm generally wary of the "teaching lesson" via slightly changing the semantics of existing build. When the build no longer works because the default value changed, ppl won't learn. They just throw things at screen and tweet angry stuff about sbt.

If we want to do evangelize, it would need to be a new method or something similar to what sbt did with `%%`.

sjrd:
Sure, a "teaching lesson" change must be a deprecation, keeping the old behavior, but encouraging a style that makes it evident that it is the "wrong thing to do", and in a major version later you can actually change the default.

Here, it could be:
- 0.13.12 emits a deprecation warning if `addCompilerPlugin` is called with a `ModuleID` where no explicit `cross` was specified, with a message explaining that compiler plugins should typically have `cross CrossVersion.full`
- 1.0.0 defaults to `CrossVersion.full`

We've used that strategy a couple times in the past in Scala.js, and users have been very happy with our deprecation policy.

PR #2655, opened 2016-07-07
Fix #2611: Codacy issues

### LINKED ISSUE #2611: todo: Codacy issues on 1.0.x
Currently there are some issues reported by Codacy (https://www.codacy.com/app/eed3si9n/sbt/dashboard) on code style and error prone category.
I've excluded scripted directory, but if things get included back again, we can ignore them.

For example, Codacy says "Fatal exception is being caught" in main-command/src/main/scala/sbt/Watched.scala.

dwijnand:
There's a lot of false positives in Codacy, sadly. But I've fixed all the ones that are plausible to fix.

I actually pushed for some more dubious changes, just so they could be talked about here in review. I'm happy to rework or revert any change.

### DIFF
--- a/main-settings/src/main/scala/sbt/std/InputWrapper.scala
+++ b/main-settings/src/main/scala/sbt/std/InputWrapper.scala
@@ -90,19 +90,14 @@ object InputWrapper {
 
   def valueMacroImpl[T: c.WeakTypeTag](c: Context): c.Expr[T] =
     ContextUtil.selectMacroImpl[T](c) { (ts, pos) =>
-      val tpe = ts.tree.tpe
-      if (tpe <:< c.weakTypeOf[Initialize[Task[T]]])
-        InputWrapper.wrapInitTask[T](c)(ts, pos)
-      else if (tpe <:< c.weakTypeOf[Initialize[T]])
-        InputWrapper.wrapInit[T](c)(ts, pos)
-      else if (tpe <:< c.weakTypeOf[Task[T]])
-        InputWrapper.wrapTask[T](c)(ts, pos)
-      else if (tpe <:< c.weakTypeOf[InputTask[T]])
-        InputWrapper.wrapInputTask[T](c)(ts, pos)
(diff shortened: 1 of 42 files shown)


eed3si9n:
Travis reports and other compilation errors:

```
[error] /home/travis/build/sbt/sbt/main-settings/src/main/scala/sbt/std/InputConvert.scala:16: value WrapInitTaskName is not a member of object sbt.std.ParserInput
[error]       case ParserInput.WrapInitTaskName => Converted.Failure(in.pos, initTaskErrorMessage)
[error]                        ^
```

eed3si9n:
> I actually pushed for some more dubious changes, just so they could be talked about here in review.

I'm not particularly good at spotting bad changes. Which ones do you feel are more dubious?

dwijnand:
- the various places where I opted to use NonFatal.. are those all correct changes?
- the performance impact of using pattern matching over if/else's

I'll actually compile it locally now and get this building :)

dwijnand:
All green. I'll fixup the commits and rebase after the initial review.

main-settings/src/main/scala/sbt/std/InputWrapper.scala
```diff
@@ -182,12 +182,12 @@ object ParserInput {
       }
     }
 
-  private def wrapInputTask[T](c: Context)(tree: c.Tree, pos: c.Position) = {
+  private def wrapInputTask[T: c.WeakTypeTag](c: Context)(tree: c.Tree, pos: c.Position) = {
```
eed3si9n, line 185: So we're thread safe because 2.11? - http://docs.scala-lang.org/overviews/reflection/thread-safety.html

dwijnand, line 185: This is a fixup commit, to be folded into the original, faulty refactor. The original code already used WeakTypeTag, so there's no semantic change, see https://github.com/sbt/sbt/pull/2655/files#diff-407b2912ff3a899d0ed98050a9b91908R176.

eed3si9n:
LGTM

PR #2659, opened 2016-07-08
Fixes #2654 Silence macro recompilation info message

### LINKED ISSUE #2654: Option to turn off the new macro recompilation info message
Ref #2614 #2637

@milessabin wrote:

> With 0.13.12-RC1 I'm getting a lot of output of the form,

```
[info] Because /home/miles/projects/shapeless/core/src/main/scala/shapeless/typeable.scala contains a macro definition, the following dependencies are invalidated unconditionally:                                                                                                     
[info]  /home/miles/projects/shapeless/core/jvm/target/scala-2.12.0-M5/src_managed/main/shapeless/tupletypeables.scala                      
[info]  /home/miles/projects/shapeless/core/src/main/scala/shapeless/alacarte.scala                                                         
[info]  /home/miles/projects/shapeless/core/src/main/scala/shapeless/ops/maps.scala                                                         
[info]  /home/miles/projects/shapeless/core/src/main/scala/shapeless/ops/traversables.scala                                                 
[info]  /home/miles/projects/shapeless/core/src/main/scala/shapeless/syntax/typeable.scala
```

> I think the idea of encouraging people to move macro definitions to subprojects makes sense, but having done that, those subprojects will want to have the warning disabled.
> Otherwise it's a warning which can't be fixed and that's ... unhelpful.

dwijnand:
Fixes #2654 Silence macro recompilation info message.

As well as add a test for the macro recompilation info message feature.

(Note, targeting the 0.13.12 branch)

### DIFF
--- a/compile/inc/src/main/scala/sbt/inc/IncOptions.scala
+++ b/compile/inc/src/main/scala/sbt/inc/IncOptions.scala
@@ -84,7 +84,9 @@ final class IncOptions(
     /**
      * Include synthetic methods into the dependency tracking by name hashing.
      */
-    val includeSynthToNameHashing: Boolean) extends Product with Serializable {
+    val includeSynthToNameHashing: Boolean,
+    /** Determines whether to log information on file recompiled due to a transitive macro change */
+    val logRecompileOnMacro: Boolean) extends Product with Serializable {
 
   /**
    * Secondary constructor introduced to make IncOptions to be binary compatible with version that didn't have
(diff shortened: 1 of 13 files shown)


eed3si9n:
LGTM

milessabin:
Awesome! Thank you :-)

stewSquared:
Can someone clarify what both means?
Suppose we have `-Ywarn-unused` and `-Ywarn-macros:both`
Does it error when a val is unused either before OR after macro expansion?
Or does it error only when a val is unused before AND after?

ie., does `both` generally result in more or fewer warnings?

PR #2661, opened 2016-07-08
AutoPlugin should require JvmPlugin by default

### LINKED ISSUE #2082: Auto plugins should depend on JvmPlugin out of the box
Since the launch of AutoPlugin there's been confusions caused by auto plugins not depending on `JvmPlugin` by default, most recently #2081.
For 100% of the plugin authors, it's most natural to assume that initialization of the so-called built-in keys have already taken place, and that they can safely call `+=` on them.
Even for the plugin authors wishing to substitute the default behaviors, they should explicitly override `requires`. By adopting affordance principle (aka principle of least surprise/resistance) auto plugins should depend on `JmvPlugin` out of the box.

Likely we need to do this when we can break semantic compat in 1.0.

dwijnand:
Fixes #2082

### DIFF
--- a/main/src/main/scala/sbt/Plugins.scala
+++ b/main/src/main/scala/sbt/Plugins.scala
@@ -71,7 +71,7 @@ abstract class AutoPlugin extends Plugins.Basic with PluginsFunctions {
 
   /** This AutoPlugin requires the plugins the [[Plugins]] matcher returned by this method. See [[trigger]].
    */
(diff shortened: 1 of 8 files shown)


eed3si9n:
LGTM

eed3si9n:
Travis says:

```
java.lang.RuntimeException: Failed tests:
    project / auto-plugins
```

Issue #2171, opened 2015-08-22
Make "version" a task


dwijnand:
For 1.0 perhaps we should make `version` a `TaskKey` rather than a `SettingKey`, ie. doesn't memoise, as sometimes you want the version to be dynamic rather than static, such as make it depend on a git tag or the date/time.

WDYT? Outside of bincompat, are there serious impacts to this change?

dwijnand:
I want to mention #2059, to cross pollinate the discussions.

dwijnand:
I tried to do this but "everything" depends on version. And as @eed3si9n quite rightly pointed out, sometimes a release is done over multiple task runs, so it's important that version is stable between runs.

dwijnand:
Expressed more fully on Gitter:

> You could decide that for you the value of version is dynamic and should be defined in terms of, let's say, the current timestamp. The problem, though, is that you don't want that value to change across, say, different modules of a project, nor change across different scala version (cross-publishing), etc. So making `version` a task puts upon a strict requirement that those publishes occur within the same task execution, so that task deduplication comes in effect, i.e. that `version` is evaluated once, to one value, only. Which would never fly as `++` is a command, aka multiple task executions.

Issue #2088, opened 2015-07-06
Fail to download the first SNAPSHOT publish to Artifactory


dwijnand:
In trying to figure out #1750 I've hit this snag. Asides from it getting in the way, I wonder if it's indicative of anything, such as my setting up wrong.
## steps

1) Setup:

```
sbt.version=0.13.8
```

Dependency project, `sbt-gh1750-bippy`:

``` scala
organization := "com.dwijnand"
        name := "sbt-gh1750-bippy"
     version := "0.1.0-SNAPSHOT"
scalaVersion := "2.11.7"

val localRelease  = "local-release"  at "http://localhost:8081/artifactory/libs-release-local"
val localSnapshot = "local-snapshot" at "http://localhost:8081/artifactory/libs-snapshot-local"

publishTo := Some(if (isSnapshot.value) localSnapshot else localRelease)
credentials := Seq(Credentials("Artifactory Realm", "localhost", "admin", "password"))
```

Dependee project, `sbt-gh1750-myapp`:

``` scala
organization := "com.dwijnand"
        name := "sbt-gh1750-myapp"
     version := "0.1.0-SNAPSHOT"
scalaVersion := "2.11.7"

val localRelease  = "local-release"  at "http://localhost:8081/artifactory/libs-release-local"
val localSnapshot = "local-snapshot" at "http://localhost:8081/artifactory/libs-snapshot-local"

resolvers += localSnapshot

libraryDependencies += "com.dwijnand" %% "sbt-gh1750-bippy" % "0.1.0-SNAPSHOT"
```

2) In `sbt-gh1750-bippy`, `publish`
3) In `sbt-gh1750-myapp`, `reload` and `update`
## problem

`update` outputs:

```
[info] Updating {file:/Users/dnw/Desktop/sbt-gh1750-myapp/}sbt-gh1750-myapp...
[info] Resolving jline#jline;2.12.1 ...
[warn]  [NOT FOUND  ] com.dwijnand#sbt-gh1750-bippy_2.11;0.1.0-SNAPSHOT!sbt-gh1750-bippy_2.11.jar (4ms)
[warn] ==== local-snapshot: tried
[warn]   http://localhost:8081/artifactory/libs-snapshot-local/com/dwijnand/sbt-gh1750-bippy_2.11/0.1.0-SNAPSHOT/sbt-gh1750-bippy_2.11-0.1.0-20150706.083249-1.jar
[...]

PR #2662, opened 2016-07-08
[sbt 1.0] Make publishTo & otherResolvers TaskKey's

### LINKED ISSUE #2059: Change publishTo type from setting Key to task Key
I want to use the credentials taskKey in the definition of the publishTo resolver (S3 Resolver) but I can't.
Is there another way to do this ?

dwijnand:
Fixes #2059


dwijnand:
`dependency-management/snapshot-resolution` provides an interesting case. It currently has:

``` scala
    publishMavenStyle := (publishTo.value match {
      case Some(repo) =>
        repo match {
          case repo: PatternsBasedRepository => repo.patterns.isMavenCompatible
          case _: RawRepository => false // TODO - look deeper
          case _: MavenRepository => true
          case _: MavenCache => true
          case _ => false  // TODO - Handle chain repository?
        }
      case _ => true
    })
```

That will no longer work as `publishMavenStyle` is a setting and `publishTo` is a task with this PR.

Should I fix it for the test, or shall I make `publishMavenStyle` a task? /cc @eed3si9n

dwijnand:
For now I (somewhat) fixed it locally to that scripted test, just so Travis would go green. But let me know.

eed3si9n:
So we now know it's possible, why do we want this again?

dwijnand:
The initial feature request (#2059) said because they wanted to define `publishTo` in terms of `credentials`, which is a task.

I'm not sure I see a strong use-case there, but, without intending to sound like the devil's advocate, may I invert the question to you: why don't we want to do this?

For clarity, I don't feel strongly that we do, this was just easy pickings that was originally in the 1.0.0 milestone. So I'm happy not to, but we should close the feature request too then.

eed3si9n:
`publishTo` seems to come up occasionally, so I was wondering what your take is. Recently I've seen an issue because bintray-sbt checks for Bintray credential file, and bintray-sbt was included into an Activator template. @2m I think also brought this up.

It's a classic setting vs task debate. From a common sense point of view where you publish your artifacts should not change over the lifetime of an sbt session. Turning it into a runtime behavior seems like an incidental complexity because it might want to depend on things that are tasks. I'm leaning towards merging this, but I want to see what's going on.

dwijnand:
Sounds like you mean https://github.com/softprops/bintray-sbt/issues/88.

There's also @2m's needs-a-rework PR https://github.com/softprops/bintray-sbt/pull/69.

The argument for setting is that we feel it should be sufficient, and settings are more performant (because they're memoized).

The argument for task is that it's more flexible.

So, if you really wonder my take, I mildly bias in this case to allow for the flexibility. (For instance this same argument applied to `version` (https://github.com/sbt/sbt/issues/2171) is much different, because version is so cornerstone to a project).

But I'm happy to hold this PR for a little while, but please let's not put in on the dead-but-no-closed PRs stack - a closed PR can always be re-opened months later.

2m:
That is correct, @dwijnand. This is what I needed when working on `bintray-sbt` - to be able to make `publishTo` (currenty a setting) depend on `credentials` (currently a task), `publishTo` needs to become a task.

dwijnand:
Artifactory (see #2088) requires the publishing URL to have a `;build.timestamp=...` suffix for SNAPSHOTs, as a guide for what SNAPSHOTs artefacts belong to the same version.

By `publishTo` being a setting defined in terms of current time (eg. `... + ";build.timestamp=" + new java.util.Date().getTime`), the build user must remember to reload the build in order to have a new suffix for the next snapshot release.

Instead if `publishTo` per a task this particular issue would work correctly out the box, because `publishTo` is then read by `publishConfiguration`, which uses the same URL for all the artefacts in the same publish.

At least in terms of the sbt 0.13 codebase, the only other place `publishTo` is used is to define `otherResolvers`, which then because part of the definition of the `ivyConfiguration`.

So I think we should merge this, if TravisCI passes.

dwijnand:
Oh actually, let me know, so I'll write some notes before we merge.

dwijnand:
Thank you

PR #2663, opened 2016-07-08
Check formatting with TravisCI

### LINKED ISSUE #2657: Travis should run code formatting and fail the build if anything changed

dwijnand:
Fixes #2657

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ matrix:
 
 env:
   matrix:
+    - SBT_CMD="test:compile scalariformCheck"
     - SBT_CMD="mimaReportBinaryIssues"
     - SBT_CMD="safeUnitTests"
     - SBT_CMD="otherUnitTests"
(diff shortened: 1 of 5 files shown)


dwijnand:
Going to rework this. I like this implementation better https://github.com/scalikejdbc/scalikejdbc/commit/b1e55a76d494dc367585 (referenced from https://github.com/sbt/util/issues/7)

dwijnand:
@eed3si9n Do you mind giving me a hand here? I'm trying to define a task `scalariformCheck` that runs `scalariformFormat` (which is a task that is defined as `ScalariformKeys.format` in code and `scalariform-format` as the task name) in aggregate on all projects `sbtRoot` aggregates, and then the git diff check once.

I tried a bunch of different things, including a variant of your answer to http://stackoverflow.com/questions/27036974/sbt-run-task-after-all-subprojects, but no dice. So I'm stuck.

Thanks.

eed3si9n:
Currently `script` section is written to be:

``` yaml
 script:
    - sbt -J-XX:ReservedCodeCacheSize=128m "$SBT_CMD"
```

and you passed in 

```
test:compile scalariformCheck
```

so

```
[info] Set current project to sbtRoot (in build file:/home/travis/build/sbt/sbt/)
[error] Expected key
[error] Expected '::'
[error] Expected end of input.
[error] test:compile scalariformCheck
[error]             ^
```

I think is to be expected.. You need to define something that you can call as a single stuff.

dwijnand:
Yeah I know, that's why I'm trying to re-define `scalariformCheck` to do `scalariformFormat` for both compile and test sources, for all aggregated projects, then the git diff check.

But I can't get the aggregation to work: it never formats the aggregated projects and therefore the git diff check never fails when it should..

dwijnand:
Finished in #2676.

PR #2673, opened 2016-07-12
Don't redefine updateOptions to ignore ThisBuild

### LINKED ISSUE #2671: SBT ignores `updateOptions` defined at `ThisBuild` level
## steps

Create a build that sets `updateOptions` at `ThisBuild` level:

``` scala
inThisBuild(Seq(updateOptions := updateOptions.value.withCachedResolution(true)))

lazy val sub = project
  .settings(printCached := { println(updateOptions.value.cachedResolution) })

lazy val printCached = taskKey[Unit]("just for debugging")
```
## problem

`updateOptions` in `sub` project ignore `ThisBuild` settings.

```
> sub/printCached
false
```
## expectation

`ThisBuild` settings should be taken into account.
## notes

As a workaround one may set `updateOptions` at `Global` level instead:

``` scala
inScope(Global)(Seq(updateOptions := updateOptions.value.withCachedResolution(true)))
```

```
> sub/printCached
true
```

This is caused by SBT `Defaults` explicitly injecting `updateOptions` from `Global` into a project, bypassing `ThisBuild`. See https://github.com/sbt/sbt/blob/v0.13.11/main/src/main/scala/sbt/Defaults.scala#L1192

The line mentioned should, probably, be removed to let normal SBT inheritance work.

dwijnand:
Fixes #2671

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1245,7 +1245,6 @@ object Classpaths {
       // By default, to retrieve all types *but* these (it's assumed that everything else is binary/resource)
       new UpdateConfiguration(retrieveConfiguration.value, false, ivyLoggingLevel.value, ArtifactTypeFilter.forbid(specialArtifactTypes))
     },
-    updateOptions := (updateOptions in Global).value,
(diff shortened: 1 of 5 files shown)


eed3si9n:
Maybe we need a guideline on what goes into what scope.

PR #2678, opened 2016-07-20
Workaround for inclusion of ScalaCheck classes in scala-partest-extras.jar

milessabin:
A test for this would require a local Scala distribution directory with the offending `lib/scala-partest-extras.jar`.

A snapshot or speedy 0.13.13 release would be super helpful.

### DIFF
--- a/testing/src/main/scala/sbt/TestFramework.scala
+++ b/testing/src/main/scala/sbt/TestFramework.scala
@@ -187,7 +187,11 @@ object TestFramework {
   def createTestLoader(classpath: Seq[File], scalaInstance: ScalaInstance, tempDir: File): ClassLoader =
     {
       val interfaceJar = IO.classLocationFile(classOf[testing.Framework])
-      val interfaceFilter = (name: String) => name.startsWith("org.scalatools.testing.") || name.startsWith("sbt.testing.")
+      // We exclude scalacheck her, possibly temporarily, because a consequence of
+      // https://github.com/scala/scala/pull/5244 is that scalacheck class files are
+      // found via the scalaInstance loader, but fail resolution because the
+      // test-interface classes aren't also present.
+      val interfaceFilter = (name: String) => name.startsWith("org.scalatools.testing.") || name.startsWith("sbt.testing.") || name.startsWith("org.scalacheck.")
       val notInterfaceFilter = (name: String) => !interfaceFilter(name)
       val dual = new DualLoader(scalaInstance.loader, notInterfaceFilter, x => true, getClass.getClassLoader, interfaceFilter, x => false)
       val main = ClasspathUtilities.makeLoader(classpath, dual, scalaInstance, tempDir)


dwijnand:
For context, see https://gitter.im/scala/contributors?at=578f245defc7f7b626ec9d22, which links to the GitHub issue in scala/scala, which links to an earlier conversation in the sbt/sbt-dev Gitter room.

dwijnand:
Personally I would prefer this to be fixed in scala/scala. But (I think) I understand the situation, so I can be convinced to accept this.

dwijnand:
@milessabin do you remember where we stand on this now?

milessabin:
Blimey, I'd completely forgotten about that!

testing/src/main/scala/sbt/TestFramework.scala
```diff
@@ -187,7 +187,11 @@ object TestFramework {
   def createTestLoader(classpath: Seq[File], scalaInstance: ScalaInstance, tempDir: File): ClassLoader =
     {
       val interfaceJar = IO.classLocationFile(classOf[testing.Framework])
-      val interfaceFilter = (name: String) => name.startsWith("org.scalatools.testing.") || name.startsWith("sbt.testing.")
+      // We exclude scalacheck her, possibly temporarily, because a consequence of
+      // https://github.com/scala/scala/pull/5244 is that scalacheck class files are
+      // found via the scalaInstance loader, but fail resolution because the
+      // test-interface classes aren't also present.
+      val interfaceFilter = (name: String) => name.startsWith("org.scalatools.testing.") || name.startsWith("sbt.testing.") || name.startsWith("org.scalacheck.")
```
eed3si9n, line 194: Would it make sense to implement the fix at the ScalaInstance layer? Adding check for `"org.scalacheck."` here feels odd.

milessabin:
I think maybe it would be best to merge this. If we don't we'll always be vulnerable to upstream changes.

testing/src/main/scala/sbt/TestFramework.scala
milessabin: This is a `ClassLoader` rather than an Ivy level thing. It's ugly, but I think this _is_ the right place.

eed3si9n: `dual` filters `scalaInstance.loader`, but even as `scalaInstance.loader` scala-partest-extras.jar is not useful or potentially dangerous isn't it? I feel like the fix should be somewhere here - https://github.com/sbt/sbt/blob/0.13/util/classpath/src/main/scala/sbt/ScalaInstance.scala

eed3si9n: Isn't `allJars(scalaHome)` the problem?

eed3si9n: See also:

``` scala
  private[this] val blacklist: Set[String] = Set("scala-actors.jar", "scalacheck.jar", "scala-partest.jar", "scala-partest-javaagent.jar", "scalap.jar", "scala-swing.jar")
```

eed3si9n (changes requested):
`scala-partest-extras.jar` should be added to the blacklist here - https://github.com/sbt/sbt/blob/0.13/util/classpath/src/main/scala/sbt/ScalaInstance.scala#L110

testing/src/main/scala/sbt/TestFramework.scala
milessabin: OK, point taken ... adding `scala-partest-extras.jar` to the blacklist seems plausible.

dwijnand:
Fixed in scala/scala https://github.com/scala/scala/pull/5414

PR #2681, opened 2016-07-22
Fix #2677: Tags.ForkedTestGroup doesn't work

### LINKED ISSUE #2677: Tags.ForkedTestGroup doesn't work
http://www.scala-sbt.org/0.13/docs/Testing.html

> Control the number of forked JVMs allowed to run at the same time by setting the limit on Tags.ForkedTestGroup tag, which is 1 by default.

---

project/build.properties

```
sbt.version=0.13.12
```

build.sbt

``` scala
concurrentRestrictions in Global := Seq(
  Tags.limit(Tags.ForkedTestGroup, 2)
)

libraryDependencies += "org.specs2" %% "specs2-core" % "3.7" % Test

scalaVersion := "2.11.7"

inConfig(Test)(Seq(
  fork := true,
  testGrouping := definedTests.value.map { test =>
    new Tests.Group(test.name, Seq(test), Tests.SubProcess(
      ForkOptions(
        javaHome.value,
        outputStrategy.value,
        Nil,
        Some(baseDirectory.value),
        javaOptions.value,
        connectInput.value,
        envVars.value
      )
    ))
  }
))
```

src/test/scala/example/Test.scala

``` scala
package example

import org.specs2.mutable.Specification

trait Test extends Specification {
  Thread.sleep(10 * 1000)
}

class Test0 extends Test

class Test1 extends Test

class Test2 extends Test

class Test3 extends Test
```

---

Then 

```
$ sbt test
...
[success] Total time: 12 s, completed Jul 18, 2016 7:30:10 PM
```

It should take at least 20 seconds to finish the tests, but `ForkedTestGroup` doesn't actually limit the number of forked test groups that run at a time.

pauldraper:
Fix #2677

Move tag to correct task

### DIFF
--- a/main/actions/src/main/scala/sbt/ForkTests.scala
+++ b/main/actions/src/main/scala/sbt/ForkTests.scala
@@ -9,9 +9,10 @@ import java.net.ServerSocket
 import java.io._
 import Tests.{ Output => TestOutput, _ }
 import ForkMain._
+import sbt.ConcurrentRestrictions.Tag
 
 private[sbt] object ForkTests {
-  def apply(runners: Map[TestFramework, Runner], tests: List[TestDefinition], config: Execution, classpath: Seq[File], fork: ForkOptions, log: Logger): Task[TestOutput] = {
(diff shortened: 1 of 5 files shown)


eed3si9n:
LGTM

eed3si9n:
Could you forward port this to sbt 1.0 plz?

Issue #2686, opened 2016-07-29
updateSbtClassifiers broken with sbt 0.13.12 on Scala 2.9 projects


jastice:
## steps

Create a simple project with this in `build.sbt`:

``` scala
scalaVersion := "2.9.3"
```

run `sbt -Dsbt.version=0.13.12 updateSbtClassifiers`
## problem

Task fails with resolve error:

```
[error] (*:updateSbtClassifiers) sbt.ResolveException: unresolved dependency: org.scala-lang#scala-reflect;2.9.3: not found
```
## notes

This problem does not occur up to sbt 0.13.11

dwijnand:
I wonder if this is fallout from #2634. (/cc @milessabin).

eed3si9n:
Interesting. Because of the precompiled compiler bridges, sbt does depend on multiple Scala versions.

milessabin:
I think it probably is a consequence of #2634. I'll investigate ... feel free to assign to me.

dwijnand:
I would assign you if I could. I think you need to join the sbt org to do so. /cc @eed3si9n

eed3si9n:
@milessabin Invited you to sbt-committer team.

milessabin:
Thanks!

dwijnand:
Welcome!

drewhk:
I actually had the same issue with the Akka build, which uses Scala 2.11.8. Had to downgrade to 0.13.11 to solve the issue.

drewhk:
[error] impossible to get artifacts when data has not been loaded. IvyNode = org.scala-lang#scala-reflect;2.11.8

dwijnand:
Fixed in https://github.com/sbt/sbt/pull/2692

PR #2691, opened 2016-08-03
Fix for #2686

milessabin:
The cause of the problem reported in #2686 is that the version clamping that was introduced in #2634 could "downgrade" scala-reflect 2.10.x to scala-reflect 2.9.x ... which doesn't exist of course.

This is a fairly brute force fix for that: if the scalaVersion being clamped to is < 2.10.x then references to scala-reflect and scala-actors artefacts are left unchanged ... this roughly approximates the behaviour you would have seen in this situation prior to #2634 with respect to those two artefacts.

A better fix might work out why such artefacts are being asked for in the first place given that this is a project which specifies 2.9.3 and does nothing other than invoke `updateSbtClassifiers` ... I don't understand the meaning or purpose of `updateSbtClassifiers` sufficiently to be able to attempt that unfortunately.

### DIFF
--- a/ivy/src/main/scala/sbt/IvyScala.scala
+++ b/ivy/src/main/scala/sbt/IvyScala.scala
@@ -57,6 +57,13 @@ private object IvyScala {
       overrideScalaVersion(module, check.scalaOrganization, check.scalaFullVersion)
   }
 
+  def pre210(scalaVersion: String): Boolean =
+    CrossVersion.scalaApiVersion(scalaVersion) match {
+      case Some((2, y)) if y < 10 => true
+      case Some((x, _)) if x < 2  => true
+      case _                      => false
+    }
+
   class OverrideScalaMediator(scalaOrganization: String, scalaVersion: String) extends DependencyDescriptorMediator {
     def mediate(dd: DependencyDescriptor): DependencyDescriptor = {
       val transformer =
@@ -65,6 +72,7 @@ private object IvyScala {
             if (mrid == null) mrid
             else
               mrid.getName match {
+                case ReflectID | ActorsID if pre210(scalaVersion) => mrid
                 case name @ (CompilerID | LibraryID | ReflectID | ActorsID | ScalapID) =>
                   ModuleRevisionId.newInstance(scalaOrganization, name, mrid.getBranch, scalaVersion, mrid.getQualifiedExtraAttributes)
                 case _ => mrid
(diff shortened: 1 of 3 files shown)


dwijnand:
While I do appreciate the effort and detail, I'm afraid this is a symptom of a bigger problem, and I don't like the fix.

Have a look this from the log for `updateSbtClassifiers`:

```
[debug] dependency descriptor has been mediated: dependency: org.scala-lang#scala-library;2.10.6 {compile=[default(compile)]} => dependency: org.scala-lang#scala-library;2.9.3 {compile=[default(compile)]}
[debug] dependency descriptor has been mediated: dependency: org.scala-lang#scala-reflect;2.10.6 {compile=[default(compile)]} => dependency: org.scala-lang#scala-reflect;2.9.3 {compile=[default(compile)]}
```

That can't be right.. right?

milessabin:
I don't disagree. Advice on addressing the deeper problem would be much appreciated.

milessabin:
My vague hunch is that the Scala dependencies for the project and the meta project are mixed together here, and the version clamping, which is only supposed to apply to the project, is being applied across the board. I think this is probably a long-standing problem which has been brought to light by #2634.

eed3si9n:
We discussed on Gitter, but I'll write out some stuff here for the record Keys.scala describe `updateSbtClassifiers` as:

> Resolves and optionally retrieves classifiers, such as javadocs and sources, for sbt, transitively.

It constructs a dependency graph for sbt itself, and uses Ivy to retrieve "classifier" artifacts (Scaladocs and sources) used by IDEs. IIUC, the dependency graph for sbt is constructed as:

``` scala
    sbtDependency := {
      val app = appConfiguration.value
      val id = app.provider.id
      val scalaVersion = app.provider.scalaProvider.version
      val binVersion = binaryScalaVersion(scalaVersion)
      val cross = if (id.crossVersioned) CrossVersion.binary else CrossVersion.Disabled
      val base = ModuleID(id.groupID, id.name, sbtVersion.value, crossVersion = cross)
      CrossVersion(scalaVersion, binVersion)(base).copy(crossVersion = CrossVersion.Disabled)
    }
```

The root of the problem seem to be that the mediator is somehow instructed to use `scalaVersion.value` (2.9.3; the Scala version for the user's project) instead of the Scala version of the build definition. (2.10.6)

dwijnand:
Perhaps we should validate that the X.Y of the scala version matches before replacing it. WDYT?

eed3si9n:
@dwijnand That's still ad-hoc. We should find why this is happening before we can address it correctly.

dwijnand:
If we're able, absolutely. I was only proposing something mildly more robust than `pre210`.

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/2692

PR #2702, opened 2016-08-19
Configurable explicit list of artifacts for Scala binary version check

### LINKED ISSUE #2701: Ivy Scala component version check misfires with alternative `scalaOrganization`
The version check in `IvyScala` [here](https://github.com/sbt/sbt/blob/0.13/ivy/src/main/scala/sbt/IvyScala.scala#L98-L118) is only correct where `scalaOrganization` is actually `org.scala-lang`, and then only so long as the collection of artefacts under that organization remains fixed.

With an alternative organization (eg. "org.typelevel") _all_ other artefacts (aside from the two whitelisted explicitly in that test, if they exist) will have their binary version tested against the Scala binary version. In particular, with a build of the form,

```
scalaOrganization := "org.typelevel"

scalaVersion := "2.11.8"

libraryDependencies += "org.typelevel" %% "cats" % "0.6.1"
```

We'll see the spurious (and very confusing) warning,

```
[warn] Binary version (0.6.1) for dependency org.typelevel#cats_2.11;0.6.1
[warn]   in org.dummy#scala-organization_2.11;0.1-SNAPSHOT differs from Scala binary version in project (2.11).
```

because of a mismatch between the cats binary version (0.6.1) and the Scala binary version (2.11.8).

I think the right fix here is to explicitly list the artefact names which we intend to validate, possibly making it a configurable setting.

milessabin:
Fixes #2701.

### DIFF
--- a/ivy/src/main/scala/sbt/IvyScala.scala
+++ b/ivy/src/main/scala/sbt/IvyScala.scala
@@ -20,6 +20,7 @@ object ScalaArtifacts {
   val ReflectID = "scala-reflect"
   val ActorsID = "scala-actors"
   val ScalapID = "scalap"
+  val Artifacts = Seq(LibraryID, CompilerID, ReflectID, ActorsID, ScalapID)
   val DottyIDPrefix = "dotty"
 
   def dottyID(binaryVersion: String): String = s"${DottyIDPrefix}_${binaryVersion}"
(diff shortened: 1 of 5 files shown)


milessabin:
Happy to forward port this to 1.0.0 when you're happy with it.

ivy/src/main/scala/sbt/IvyScala.scala
```diff
@@ -44,13 +45,13 @@ object SbtArtifacts {
 
 import ScalaArtifacts._
 
-final case class IvyScala(scalaFullVersion: String, scalaBinaryVersion: String, configurations: Iterable[Configuration], checkExplicit: Boolean, filterImplicit: Boolean, overrideScalaVersion: Boolean, scalaOrganization: String = ScalaArtifacts.Organization)
+final case class IvyScala(scalaFullVersion: String, scalaBinaryVersion: String, configurations: Iterable[Configuration], checkExplicit: Boolean, filterImplicit: Boolean, overrideScalaVersion: Boolean, scalaOrganization: String = ScalaArtifacts.Organization, scalaArtifacts: Seq[String] = ScalaArtifacts.Artifacts)
```
dwijnand, line 48: This is in the public API, so it's a binary incompatible change.

milessabin, line 48: That's unfortunate.

What do you suggest? I could hard code the list for 0.13 and move this to 1.0.0.

dwijnand, line 48: Yep, that.

PR #2703, opened 2016-08-19
Explicit list of artifacts for Scala binary version check

### LINKED ISSUE #2701: Ivy Scala component version check misfires with alternative `scalaOrganization`

milessabin:
Fixes #2701.

### DIFF
--- a/ivy/src/main/scala/sbt/IvyScala.scala
+++ b/ivy/src/main/scala/sbt/IvyScala.scala
@@ -20,6 +20,7 @@ object ScalaArtifacts {
   val ReflectID = "scala-reflect"
   val ActorsID = "scala-actors"
   val ScalapID = "scalap"
+  val Artifacts = Seq(LibraryID, CompilerID, ReflectID, ActorsID, ScalapID)
   val DottyIDPrefix = "dotty"
 
   def dottyID(binaryVersion: String): String = s"${DottyIDPrefix}_${binaryVersion}"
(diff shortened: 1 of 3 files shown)


dwijnand:
LGTM

eed3si9n:
LGTM

PR #2705, opened 2016-08-22
Adds templateResolvers and `new` command

eed3si9n:
This adds `new` command, which helps create a new build definition. The
`new` command is extensible via a mechanism called the template
resolver, which evaluates the arbitrary arguments passed to the command to find
and run a template.

See https://github.com/sbt/template-resolver/blob/master/src/main/java/sbt/template/TemplateResolver.java

As a reference implementation, [Giter8](http://www.foundweekends.org/giter8/) integration is [provided](https://github.com/eed3si9n/sbt/blob/954e7444086e4345c767585574d6cec9a2d83cb3/main/src/main/scala/sbt/Giter8TemplateResolver.scala), which works as follows:

```
$ sbt new eed3si9n/hello.g8
```

This will run eed3si9n/hello.g8 using Giter8.

`templateResolvers` is just a key in the `Global` scope, so you can write your own auto plugin to extend it:

``` scala
object Giter8TemplatePlugin extends AutoPlugin {
  override def requires = CorePlugin

  override lazy val globalSettings: Seq[Setting[_]] =
    Seq(
      templateResolvers += Giter8TemplateResolver
    )
}
```

The credit for the idea of extensible template support goes to @corruptmemory, but I take the blame for implementation details :)
### update

If you want to make your own Giter8 template, see [Making your own templates](http://www.foundweekends.org/giter8/template.html). All you need is a Git repo.

If you want to extend template _resolving_ mechanism to something other than Giter8, just extend this Java class:

``` java
package sbt.template;

/** A way of specifying template resolver.
 */
public interface TemplateResolver {
  /** Returns true if this resolver can resolve the given argument.
   */
  public boolean isDefined(String[] arguments);
  /** Resolve the given argument and run the template.
   */
  public void run(String[] arguments);
}
```

and define an auto plugin that adds to `templateResolvers`. For example you might want to create Maven archetype-like template, or whatever mechanism that exists in-house.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -449,7 +449,8 @@ lazy val commandProj = (project in mainPath / "command").
   settings(
     testedBaseSettings,
     name := "Command",
-    libraryDependencies ++= Seq(launcherInterface, templateResolverApi, giter8)
+    libraryDependencies ++= Seq(launcherInterface, templateResolverApi, giter8),
+    dependencyOverrides += plexusUtils
   )
 
 // Fixes scope=Scope for Setting (core defined in collectionProj) to define the settings system used in build definitions
@@ -488,6 +489,7 @@ lazy val mavenResolverPluginProj = (project in file("sbt-maven-resolver")).
     baseSettings,
     name := "sbt-maven-resolver",
     libraryDependencies ++= aetherLibs,
+    dependencyOverrides += plexusUtils,
     sbtPlugin := true
   )
 
--- a/build.sbt
+++ b/build.sbt
@@ -449,7 +449,7 @@ lazy val commandProj = (project in mainPath / "command").
   settings(
     testedBaseSettings,
     name := "Command",
-    libraryDependencies += launcherInterface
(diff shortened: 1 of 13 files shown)


hseeberger:
Welcome back, g8 ;-)

ktoso:
Excellent, looking forward to this!

dwijnand:
LGTM

huntc:
Love it!

nafg:
You can use code like https://gist.github.com/nafg/05fafb7050ed38854184b4ca64612da7 to use Activator templates.

It seems the way IntelliJ (and I think also Activator itself) gets activator templates is through a very complex API that ultimately gives you a zip file containing a full-text search database (perhaps it was lucene?) which you can then use the client library for locally to search. Sounds like lightbend is really trying to be enterprise-ready... :trollface:

What my code does instead, is simply scrape the Activator index page for template ids. You can then download the template bundle by appending the id to `https://www.lightbend.com/activator/template/bundle/`.

It shouldn't be a big deal to write a templateResolver that takes the same shortcut.

PR #2708, opened 2016-08-24
Fix key selection for build level keys

### LINKED ISSUE #2707: Can't invoke non-build level keys
## steps

```
> show baseDirectory
[error] Ambiguous keys: {.}/*:baseDirectory, *:baseDirectory
[error] show baseDirectory
[error]                   ^
> show {.}/*:baseDirectory
[info] /s/t
> show *:baseDirectory
[error] Ambiguous keys: {.}/*:baseDirectory, *:baseDirectory
[error] show *:baseDirectory
[error]                     ^
```
## problem

sbt suggests a key that doesn't work.
## expectation

`show *:baseDirectory` should show the base directory.
## notes

This is broken as of 0.13.12 and works in 0.13.11, so I have a feeling it might be due to #2469. /cc @Duhemm

Duhemm:
PR #2469 added build keys to tab completion, with the side effect of
considering them as available candidates in key selection, thus making sbt
think that some inputs were ambiguous (e.g. `baseDirectory`): should it
apply to the current project or to the build level key?

This commit fixes this issue by improving the key selection:
- If there are no candidates, we return the default key
- If there's a single possible project level key, and zero or more
  build level keys, then we select the project level key.
- If there are zero project level keys, and a single build level key,
  then we select the build level key
- If there are multiple candidates, sbt says that the input is
  ambiguous.

Fixes #2707

### DIFF
--- a/main/src/main/scala/sbt/Act.scala
+++ b/main/src/main/scala/sbt/Act.scala
@@ -77,11 +77,16 @@ object Act {
       selectFromValid(ss filter isValid(data), default)
     }
   def selectFromValid(ss: Seq[ParsedKey], default: Parser[ParsedKey])(implicit show: Show[ScopedKey[_]]): Parser[ParsedKey] =
-    selectByTask(selectByConfig(ss)) match {
-      case Seq()       => default
-      case Seq(single) => success(single)
-      case multi       => failure("Ambiguous keys: " + showAmbiguous(keys(multi)))
+    selectByTask(selectByConfig(ss)) partition isBuildKey match {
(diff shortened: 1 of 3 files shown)


dwijnand:
Thanks @Duhemm. Mind adding a scripted test for this?

Duhemm:
@dwijnand Just added the scripted test - but I'm not sure how to make it more exciting...

eed3si9n:
LGTM

dwijnand:
Perfect. Thank you.

PR #2709, opened 2016-08-25
Deprecate .value extension method from input tasks

eed3si9n:
Calling `.value` method on an input task returns `InputTask[A]`, which
is completely unintuitive and often results to a bug.
In most cases `.evaluated` should be called, which returns `A` by
evaluating the task. Just in case `InputTask[A]` is needed,
`toInputTask` method is now provided.
### note

Implicit for `.value` is on `Initialize[T]`:

``` scala
  implicit def macroValueI[T](in: Initialize[T]): MacroValue[T] = ???
```

For instance, in case of `Initalize[InputTask[Unit]]`, `T` is `InputTask[Unit]`, so we're checking `tpe` to conform to `Initialize[InputTask[InputTask[Unit]]]`, which I don't think would ever match. Fixing the semantics of `.value` at this juncture is dangerous, so the best we can do is remove it and tell people to use `.evaluated`.

### DIFF
--- a/main/settings/src/main/scala/sbt/std/InputWrapper.scala
+++ b/main/settings/src/main/scala/sbt/std/InputWrapper.scala
@@ -98,7 +98,7 @@ object InputWrapper {
         InputWrapper.wrapInitTask[T](c)(ts, pos)
       else if (tpe <:< c.weakTypeOf[Initialize[T]]) {
         if (c.weakTypeOf[T] <:< c.weakTypeOf[InputTask[_]]) {
-          c.warning(pos, """`value` is deprecated for an input task. Use `evaluated` or `toInputTask`.
+          c.warning(pos, """`value` is deprecated for an input task. Use `evaluated` or `inputTaskValue`.
                            |See http://www.scala-sbt.org/0.13/docs/Input-Tasks.html for more details.""".stripMargin)
         }
         InputWrapper.wrapInit[T](c)(ts, pos)
@@ -112,7 +112,7 @@ object InputWrapper {
       else
         c.abort(pos, s"Internal sbt error. Unexpected type ${tpe.widen}")
     }
-  def toInputTaskMacroImpl[T: c.WeakTypeTag](c: Context): c.Expr[InputTask[T]] =
+  def inputTaskValueMacroImpl[T: c.WeakTypeTag](c: Context): c.Expr[InputTask[T]] =
     ContextUtil.selectMacroImpl[InputTask[T]](c) { (ts, pos) =>
       InputWrapper.wrapInit[InputTask[T]](c)(ts, pos)
     }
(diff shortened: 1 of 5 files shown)


main/settings/src/main/scala/sbt/std/InputWrapper.scala
```diff
@@ -96,17 +96,22 @@ object InputWrapper {
...
+      else if (tpe <:< c.weakTypeOf[Initialize[T]]) {
+        if (c.weakTypeOf[T] <:< c.weakTypeOf[InputTask[_]]) {
+          c.abort(pos, """`value` is no longer allowed for an input task. Use `evaluated` or `toInputTask`.
+                         |See http://www.scala-sbt.org/0.13/docs/Input-Tasks.html for more details.""".stripMargin)
+        } else InputWrapper.wrapInit[T](c)(ts, pos)
+      }
       else if (tpe <:< c.weakTypeOf[Task[T]])
         InputWrapper.wrapTask[T](c)(ts, pos)
-      else if (tpe <:< c.weakTypeOf[InputTask[T]])
-        InputWrapper.wrapInputTask[T](c)(ts, pos)
```
Duhemm, line 104: What about the case when `tpe <:< c.weakTypeOf[InputTask[T]]` is `true`? Why can it be removed?

eed3si9n: See Note I just added to the description of this issue.

Duhemm: This makes sense, thanks!

main/settings/src/main/scala/sbt/std/InputWrapper.scala
```diff
@@ -96,17 +96,22 @@ object InputWrapper {
...
-      else if (tpe <:< c.weakTypeOf[Initialize[InputTask[T]]])
-        InputWrapper.wrapInitInputTask[T](c)(ts, pos)
       else
         c.abort(pos, s"Internal sbt error. Unexpected type ${tpe.widen}")
     }
+  def toInputTaskMacroImpl[T: c.WeakTypeTag](c: Context): c.Expr[InputTask[T]] =
+    ContextUtil.selectMacroImpl[InputTask[T]](c) { (ts, pos) =>
+      val tpe = ts.tree.tpe
```
Duhemm: I think this is unused

dwijnand:
Why are we breaking this in 0.13? Why not deprecate with replacement in 0.13 and replace/remove in 1.0.

eed3si9n:
Because I want to break things when we can during the tech preview (0.13.5 onward).
I don't think there's harm here since plugin bincompat won't be affected.

dwijnand:
I thought we want to experiment and preview changes, without breaking backwards compatibility (until 1.0), both binary compatibility and source compatibility I assumed, and this is breaking source compatibility.

eed3si9n:
@dwijnand Changed to deprecation.

main/settings/src/main/scala/sbt/std/InputWrapper.scala
eed3si9n: I think I was wrong on the initial analysis. It doesn't match during `.value` but it does get used in some other code path. Added it back 725028553c56a151ae8360dd37ed817a8efc13a0 with a better test.

PR #2710, opened 2016-08-25
[sbt 1.0] Remove .value extension method from input tasks

eed3si9n:
This is a forward port of #2709

Calling `.value` method on an input task returns `InputTask[A]`, which
is completely unintuitive and often results to a bug.

In most cases `.evaluated` should be called, which returns `A` by
evaluating the task. Just in case `InputTask[A]` is needed,
`inputTaskValue` method is now provided.


dwijnand:
LGTM

Issue #2660, opened 2016-07-08
Create a sbt-compat as a migration tool to 1.x


dwijnand:
See https://github.com/sbt/sbt/issues/2619

dwijnand:
And/or revive [sbt-cross-building](https://github.com/jrudolph/sbt-cross-building) /cc @jrudolph.

dwijnand:
I think we'll make due with a combination of aliases/shims shipped within sbt, and [sbt-migration-rewrites](https://github.com/scalacenter/sbt-migration-rewrites). If need be we can create this.

PR #2711, opened 2016-08-26
[sbt 1.0] Remove old operators `<<=`, `<+=`, and `<++=`

eed3si9n:
`<<=`, `<+=`, and `<++=` will be removed.

The recommended migration path is sbt 0.13 style `:=` macro.

These operators are carryover from sbt 0.10 - 0.12, and haven't been on the Getting Started guide since [sbt 0.13.0](http://www.scala-sbt.org/0.13.0/docs/Getting-Started/More-About-Settings.html) came out in 2013.

This PR is on top of #2710


hseeberger:
+1

magro:
How would this be rewritten:

```
compile in Jmh <<= (compile in Jmh) dependsOn (compile in Test)
```

(from https://github.com/ktoso/sbt-jmh)

ccmtaylor:
we use `<<=` a lot to setup assembly settings like

``` scala
assemblyMergeStrategy in assembly <<= (assemblyMergeStrategy in assembly) { (old) => {
        case PathList("org", "mockito", xs @ _*) => MergeStrategy.first
        case PathList("com", "twitter", "common", xs @ _*) => MergeStrategy.first
        case x => old(x)
      }}
```

Intellij throws a fit trying to parse it, but it works :)

eed3si9n:
@magro The RHS of `<<=` is an `Initialize[_]` expression. For instance the return type of `dependsOn` method is `Initialize[Task[A]]`. And, anything that returns `Initialize[_]` can be put in a parenthesis and then put `.value` at the end as follows:

``` scala
compile in Jmh := ((compile in Jmh) dependsOn (compile in Test)).value
```

magro:
@eed3si9n Awesome, thanks for the great explanation!

dwijnand:
@ccmtaylor:

``` scala
assemblyMergeStrategy in assembly ~= (old => {
  case PathList("org", "mockito", xs @ _*)           => MergeStrategy.first
  case PathList("com", "twitter", "common", xs @ _*) => MergeStrategy.first
  case x                                             => old(x)
})
```

SethTisue:
fwiw, I still often see builds that use `<<=` in order to use `dependsOn`, so e.g.

    compile in Compile <<= (compile in Compile).dependsOn(myTask)

even when the rest of the build has moved over to `:=`

SethTisue:
the thing about `<<=` and `dependsOn` is going to be an FAQ for sure. I'd say it accounts for most of the old operator usages I still see in the wild

dwijnand:
I've not had a chance to review this yet but ideally <<='s deprecation message could detail that.

eed3si9n:
@metasim suggested on [twitter](https://twitter.com/metasim/status/769930437359796224):

> Would be great if legacy usage generated a build error with instructions on how to fix, a la "implicit not found" errors.

It might be a good idea to to this in 1.0, not just 0.13.13, so I might bring back those methods and define a macro that just aborts with an error message.

magro:
A helpful error message would indeed be great!

dwijnand:
@eed3si9n Ideally for 1.0 we could do that in sbt-compat (ref #2660).

eed3si9n:
@dwijnand I was thinking sbt-compat would be more for sbt plugin authors to backport sbt 1.0 plugins so it's source-compatible with sbt 0.13 if it's possible. Any migration error message if we're adding them should be added to sbt out of the box without additional plugin if we're targeting build users, no?

dwijnand:
Ah I see, yes.

Personally I think we should deprecate in 0.13 and remove in 1.0: they've
been "socially" deprecated for a while now.

eed3si9n:
@dwijnand I don't know if we can count on all users first trying 0.13.13 before jumping to sbt 1.0 out of some necessity. I don't think it hurts to put `def <<=` with a macro backing that immediately fails with an error message.

dwijnand:
It's just extra cruft we have to hold on to.. When would you be happy to then actually get rid of them? sbt 2.0? or sbt 1.1?

metasim:
> It's just extra cruft we have to hold on to.. When would you be happy to then actually get rid of them? sbt 2.0? or sbt 1.1?

Given the potential to save people a lot of frustration, I'd say keep them until 2.0. That said, maybe implement them in a file called `Cruft.scala` just to make the point? :-)

eed3si9n:
I talked to Dale earlier, and he's cool with error message only macro idea.

dwijnand:
LGTM

metasim:
+1 for use of "Restligeist" :-)

eed3si9n:
@metasim Spread the word - https://twitter.com/eed3si9n/status/770584274819055617

DavidPerezIngeniero:
+1

pfn:
Yuck, I use this mnemonic quite often 

```
def someTask = Def.task {
... 
} 

someTaskKey <<= someTask
```

Changing it to `someTaskKey := someTask.value` feels so ugly. The change to `dependsOn` is especially ugly and a reason I exclusively use `<<=`

PR #2715, opened 2016-08-27
Backport style changes in scripted tests and Defaults.scala

eed3si9n:
This backports the scripted tests and Defaults.scala style changes to use `build.sbt` and `:=`.

The purpose is to catch mistakes earlier if there's any, and also to prepare for the `<<=` deprecation.

/review @dwijnand, @Duhemm


dwijnand:
A little much to review, so a superficial LGTM - I'm sure it's all excellent.

dwijnand:
Not merging in case you're doing your PRs in particular order and want to rebase (and/or squash) this first.

eed3si9n:
This one is good to be merged given we agree on the general direction of `<<=` removal in sbt 1.0.

dwijnand:
Ok cool.

Also I realise now that this is required for https://github.com/sbt/sbt/pull/2716, so merging this now.

main/src/main/scala/sbt/Opts.scala
```diff
@@ -52,14 +52,14 @@ object DefaultOptions {
   def pluginResolvers(plugin: Boolean, snapshot: Boolean): Seq[Resolver] = {
     if (plugin && snapshot) Seq(Classpaths.typesafeSnapshots, Classpaths.sbtPluginSnapshots) else Nil
   }
-  def addResolvers: Setting[_] = Keys.resolvers <++= Keys.isSnapshot apply resolvers
-  def addPluginResolvers: Setting[_] = Keys.resolvers <++= (Keys.sbtPlugin, Keys.isSnapshot) apply pluginResolvers
```
dwijnand, line 56: removing this is a breaking change: https://github.com/scala/community-builds/pull/290

PR #2716, opened 2016-08-27
Deprecate the old operators `<<=`, `<+=`, and `<++=`

eed3si9n:
This PR is on top of #2715

This change deprecates the no-longer-documented operators `<<=`, `<+=`, and `<++=` in anticipation of the removal in sbt 1.0. Ref #2711

This will display the deprecation message as follows:

``` diff
+    @deprecated("""Use `key := { x.value }` or `key ~= (old => { newValue })`.
+                  |The RHS of `<<=` takes an `Initialize[_]` expression, which can be converted to `:=` style
+                  |by wrapping the expression in parenthesis, and calling `.value` at the end.
+                  |For example, `key := (key.dependsOn(compile in Test)).value`.""".stripMargin, "0.13.13")
```

/review @dwijnand, @Duhemm, @SethTisue


dwijnand:
LGTM

PR #2717, opened 2016-08-27
Add extraProjects and derivedProjects (formerly known as buildExtras and projectExtras)

### LINKED ISSUE #2532: Ability to programmatically define projects: Seq[Project]
With the new move to `.sbt` files, we will lack a key feature which is the ability to transform `Seq[Project]`.

A typical usecase for this is when the build structure is generated (perhaps translated from another tool) and the owner of the sbt build scripts can only apply transformations to the `Project` object rather than edit the `build.sbt` directly.

I feel this is a blocker for 1.0.

eed3si9n:
Fixes #2532

This adds support to generate synthetic subprojects from an auto plugin.

In addition, a method called `projectOrigin` is added to distinguish Organic, BuildExtra, ProjectExtra, and GenericRoot.

To generate subprojects, override `buildExtras`
method as follows:

``` scala
import sbt._
import Keys._

object BuildExtrasPlugin extends AutoPlugin {
  override def buildExtras: Seq[Project] =
    List("foo", "bar", "baz") map generateProject

  def generateProject(id: String): Project =
    Project(id, file(id)).
      settings(
        name := id
      )
}
```

Subprojects may also be derived from an existing subproject
by overriding `projectExtras`.

``` scala
import sbt._
import Keys._

object ProjectExtrasPlugin extends AutoPlugin {
  // Enable this plugin by default
  override def requires: Plugins = sbt.plugins.CorePlugin

  override def projectExtras(proj: ProjectDefinition[_]): Seq[Project] =
    // Make sure to exclude project extras to avoid recursive generation
    if (proj.projectOrigin != ProjectOrigin.ProjectExtra) {
      val id = proj.id + "1"
      Seq(
        Project(id, file(id)).
          enablePlugins(DatabasePlugin)
      )
    }
    else Nil
}
```

/review @dwijnand, @Duhemm

### DIFF
--- a/main/src/main/scala/sbt/Project.scala
+++ b/main/src/main/scala/sbt/Project.scala
@@ -64,6 +64,9 @@ sealed trait ProjectDefinition[PR <: ProjectReference] {
    */
   def plugins: Plugins
 
+  /** Indicates whether the project was created organically, or was generated synthetically. */
+  def projectOrigin: ProjectOrigin
+
   /** The [[AutoPlugin]]s enabled for this project.  This value is only available on a loaded Project. */
   private[sbt] def autoPlugins: Seq[AutoPlugin]
 
@@ -88,7 +91,7 @@ sealed trait Project extends ProjectDefinition[ProjectReference] {
   def copy(id: String = id, base: File = base, aggregate: => Seq[ProjectReference] = aggregate, dependencies: => Seq[ClasspathDep[ProjectReference]] = dependencies,
     delegates: => Seq[ProjectReference] = delegates, settings: => Seq[Setting[_]] = settings, configurations: Seq[Configuration] = configurations,
     auto: AddSettings = auto): Project =
-    unresolved(id, base, aggregate = aggregate, dependencies = dependencies, delegates = delegates, settings, configurations, auto, plugins, autoPlugins)
+    unresolved(id, base, aggregate = aggregate, dependencies = dependencies, delegates = delegates, settings, configurations, auto, plugins, autoPlugins, projectOrigin)
 
   def resolve(resolveRef: ProjectReference => ProjectRef): ResolvedProject =
     {
@@ -96,7 +99,7 @@ sealed trait Project extends ProjectDefinition[ProjectReference] {
       def resolveDeps(ds: Seq[ClasspathDep[ProjectReference]]) = ds map resolveDep
       def resolveDep(d: ClasspathDep[ProjectReference]) = ResolvedClasspathDependency(resolveRef(d.project), d.configuration)
       resolved(id, base, aggregate = resolveRefs(aggregate), dependencies = resolveDeps(dependencies), delegates = resolveRefs(delegates),
-        settings, configurations, auto, plugins, autoPlugins)
+        settings, configurations, auto, plugins, autoPlugins, projectOrigin)
     }
   def resolveBuild(resolveRef: ProjectReference => ProjectReference): Project =
     {
@@ -104,7 +107,7 @@ sealed trait Project extends ProjectDefinition[ProjectReference] {
       def resolveDeps(ds: Seq[ClasspathDep[ProjectReference]]) = ds map resolveDep
       def resolveDep(d: ClasspathDep[ProjectReference]) = ClasspathDependency(resolveRef(d.project), d.configuration)
       unresolved(id, base, aggregate = resolveRefs(aggregate), dependencies = resolveDeps(dependencies), delegates = resolveRefs(delegates),
-        settings, configurations, auto, plugins, autoPlugins)
+        settings, configurations, auto, plugins, autoPlugins, projectOrigin)
     }
 
   /**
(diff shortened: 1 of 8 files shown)


dwijnand:
Wow knocking them out the park this week, awesome.

I'm only concerned with the name buildExtras: for builds where all the projecrs are programmatically generated it sounds a bit weird. So I was thinking perhaps just "projects" (like we have projectSettings, not extraProjectSettings).

projectExtras I'm not sure of either, but I don't have any better suggestions.

eed3si9n:
> So I was thinking perhaps just "projects" (like we have projectSettings, not extraProjectSettings).
- If I call it `projects` that would imply both the organic one and synthetic ones.
- If I call it `syntethicProjects` or `extras` that would imply both the build-level and project-level.

`buildExtras` I think has a nice symmetry with `buildSettings` and easy to grasp what it would do.

dwijnand:
I see the organic/synthetic project argument. But what's a project-level project?

eed3si9n:
`def projectExtras(proj: ProjectDefinition[_]): Seq[Project]`.
project-level extras allow you to generate a synthetic subproject from an existing stuff.

dwijnand:
I see. Cool.

main/src/main/scala/sbt/Project.scala
```diff
@@ -174,13 +177,19 @@ sealed trait Project extends ProjectDefinition[ProjectReference] {
...
   /** Definitively set the [[AutoPlugin]]s for this project. */
   private[sbt] def setAutoPlugins(autos: Seq[AutoPlugin]): Project = {
     // TODO: for 0.14.0, use copy when it has the additional `autoPlugins` parameter
-    unresolved(id, base, aggregate = aggregate, dependencies = dependencies, delegates = delegates, settings, configurations, auto, plugins, autos)
+    unresolved(id, base, aggregate = aggregate, dependencies = dependencies, delegates = delegates, settings, configurations, auto, plugins, autos, projectOrigin)
+  }
+
+  /** Definitively set the [[ProjectOrigin]] for this project. */
+  private[sbt] def setProjectOrigin(origin: ProjectOrigin): Project = {
+    // TODO: for 0.14.0, use copy when it has the additional `autoPlugins` parameter
```
dwijnand: Is this TODO right?

main/src/main/scala/sbt/Project.scala
```diff
@@ -192,6 +201,15 @@ sealed trait ClasspathDep[PR <: ProjectReference] { def project: PR; def configu
 final case class ResolvedClasspathDependency(project: ProjectRef, configuration: Option[String]) extends ClasspathDep[ProjectRef]
 final case class ClasspathDependency(project: ProjectReference, configuration: Option[String]) extends ClasspathDep[ProjectReference]
 
+/** Indicate whether the project was created organically, or synthesized by a plugin. */
```
dwijnand: Could you explain what GenericRoot is (somehow)? Either here or on the case objects.

main/src/main/scala/sbt/Project.scala
eed3si9n: It should be additional `projectOrigin` parameter, I guess.

notes/0.13.13/extras.md
```diff
@@ -0,0 +1,43 @@
...
+    import sbt._
+    import Keys._
+
+    object ProjectExtrasPlugin extends AutoPlugin {
+      // Enable this plugin by default
+      override def requires: Plugins = sbt.plugins.CorePlugin
+      override def trigger = allRequirements
+
+      override def projectExtras(proj: ProjectDefinition[_]): Seq[Project] =
+        // Make sure to exclude project extras to avoid recursive generation
```
dwijnand, line 34: Are we not managing this sbt side because there are edge-cases where it's convenient to have (some level of) recursion? Eg. one autoplugin's projectExtras consuming a previous autoplugin's projectExtras?

If not, then can we avoid making this something the user needs to deal with?

dwijnand:
I'm really happy to see progress, thank you.

Reviewed, some minor tweaks, but generally LGTM.

notes/0.13.13/extras.md
eed3si9n: Yes. It's possible that plugin A adds Foo1, Bar1, and plugin B derives more subprojects off of that.

eed3si9n:
Pushed a commit per reviews.

mslinn:
Docs addressing motivation (why does this feature exist, what problems does it solve) would be helpful.

clhodapp:
I agree 100% with @dwijnand's first comment. The feature itself here is absolutely terrific. That said, calling these "extras" feels really under-descriptive to me. There are innumerable "extra" things that could be generated by a plugin, so, as a user, why would I assume that the "extras" would be projects? Also, @dwijnand indicates, this naming would feel kind of weird if you were in a setting where most projects were generated (maybe someone defines a plugin in the future for opinionated builds described by a JSON dependencies file, for instance). Why not just call them `generatedProjects` and `generatedSubprojects`?

clhodapp:
Or `derivedProjects` for the second one might be more accurate

eed3si9n:
> There are innumerable "extra" things that could be generated by a plugin, so, as a user, why would I assume that the "extras" would be projects?

A typical build user would not see the method name when they consume a plugin. This is mostly for advanced plugin authors. 

I want the names that can be used as noun phrase like `projectSettings` and `buildSettings` that would distinguish the build- and project-level clearly. `generatedProjects` + `derivedProjects` are not bad, but they are more subtle because you can claim that both are "generated" and/or "derived" off of something like JSON. Can we come up with something that's unambiguous?

ahjohannessen:
Why not just `subProjects` ?

clhodapp:
How about `independentProjects` and `derivedProjects`?

One thing I'll note now that I've thought about it is that sbt's model doesn't even really include the notion of "subprojects" unless you count aggregation (which I don't _think_ is in play here). There are just "projects". So isn't it a bit confusing to say that these generated projects are "subprojects"?

mslinn:
I do not believe it is accurate to say SBT does not support the concept of subprojects. See dependsOn in the [multi-project build page](http://www.scala-sbt.org/1.0/docs/Multi-Project.html).

eed3si9n:
ok. How about:

``` scala
def extraProjects: Seq[Project]
def derivedProjects(proj: ProjectDefinition[_]): Seq[Project]
```

?

clhodapp:
That combo is better than any one so far. From my point of view, the bikeshedding can stop with them ;). Thanks again for this great feature.

clhodapp:
@mslinn That usage of "subproject" means "subproject of the build" not "subproject of another project" and is roughly synonymous with just "project".

mslinn:
@clhodapp if this is going to be an ongoing source of confusion, the docs should introduce a term like "build subproject" or "subbuild" - I don't care if the b's are elided or not.

eed3si9n:
Sent PR for the rename (#2738)

PR #2721, opened 2016-08-31
bumping to zinc 1.0.0-X3

eed3si9n:


dwijnand:
TravisCI failed

``` scala
[info] java.util.NoSuchElementException: None.get
[info]  at scala.None$.get(Option.scala:347)
[info]  at scala.None$.get(Option.scala:345)
[info]  at sbt.internal.inc.TextAnalysisFormat$APIsF$$anonfun$23.apply(TextAnalysisFormat.scala:277)
```

PR #2738, opened 2016-09-15
Rename to extraProjects/derivedProjects

eed3si9n:
Per discussion in #2717, buildExtras and projectExtras are renamed to extraProjects and derivedProjects respectively.

/review @dwijnand, @Duhemm


dwijnand (approved):
Much better, thanks for following up.

Issue #1041, opened 2013-12-15
Passing commandline arguments to an SBT script doesn't work in 0.13.1


404-:
Given an executable file called 'sbt-test.scala' which contains the code below, attempting to run using SBT 0.13.1 with the following args yields errors:

```
$ ./sbt-test.scala --arg=val
[error] Expected ID character
[error] Not a valid command: arg
[error] arg=val
```

```
$ ./sbt-test.scala  --flag --arg=val
[error] Not a valid command: flag
[error] flag
```

Running the same commands but with SBT 0.13.0 completes successfully and yields expected output without any errors

``` scala
#!/bin/sh
SBT_0_13_0="/opt/tripadvisor/l/data/multilib/sbt/0.13.0/sbt-launch-0.13.0.jar"
SBT_0_13_1="/opt/tripadvisor/l/data/multilib/sbt/0.13.1/sbt-launch-0.13.1.jar"
SBT="$SBT_0_13_1"
exec java -Dsbt.main.class=sbt.ScriptMain -jar "$SBT" $0 "$@"
!#

/***
scalaVersion := "2.10.3"

libraryDependencies ++= Seq(
  "com.typesafe" % "config" % "1.0.0"
)
*/

import com.typesafe.config.{ConfigFactory, Config}

println(s"Args: ${args mkString ", "}")

val cfg = ConfigFactory.parseString(
  """
    |credentials {
    |  user = someone
    |  pass = s3cr3t
    |}
  """.stripMargin)

println(cfg getString "credentials.user")
```

dwijnand:
This must've broke with the 0.13.1 feature:

> Early commands: any command prefixed with -- gets moved before other commands on startup and doesn’t force sbt into batch mode.

[0.13.1 release notes](http://www.scala-sbt.org/0.13/docs/ChangeSummary_0.13.0.html#sbt+0.13.1)

nafg:
+1

bshlgrs:
Any updates to this?

nafg:
Have you tried on 0.13.9?

o-martin:
Pb is still there with 0.13.9 all arguments starting with "--" are interpreted by sbt and not passed to the scala scripts as arguments. Other pattern are successfully passed

SethTisue:
Seems like this would be a fix that someone out there in open-source land could reasonably tackle without having to learn tons about how sbt works? I assume that's why @jsuereth tagged it "community".

[...]

PR #2741, opened 2016-09-15
[sbt 1.0] Rename early command to `early(command)`

### LINKED ISSUE #2734: Rename early command in sbt 1.0
## problem

sbt uses `--` in a way inconsistent with Unix convention. (https://github.com/sbt/sbt/commit/e93c4450a1a06fe6c67df821c65fea302cecca84):

```
> help --
--<command>

        Schedules an early command, which will be run before other commands on the command line.
        The order is preserved between all early commands, so `sbt --a --b` executes `a` and `b` in order.
```
## expectation

I'm not sure what the purpose of the early command is. Can we remove this in sbt 1.0?
## notes
- http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html
- https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html

/cc @jsuereth, @dwijnand 

Ref #1041

eed3si9n:
Fixes #2734, Ref #1041

e93c4450a1a06fe6c67df821c65fea302cecca84 added a feature called _early command_, which uses `--` as a prefix to denote some commands that runs ahead of session loading. While the feature might be useful especially for logging, `--` is too useful just for this purpose.

In addition, this adds log level commands with single `-`, such as `-error` to treat them as early commands.

/review @dwijnand, @Duhemm

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -10,6 +10,8 @@ Migration notes
 - Removed the pre-0.13.7 *.sbt file parser (previously available under `-Dsbt.parser.simple=true`)
 - Removed old, hyphen-separated key names (use `publishLocal` instead of `publish-local`)
 - Removes no-longer-documented old operators `<<=`, `<+=`, and `<++=`.
+- Renames early command feature from `--<command>` to `early(<command>)`.
+- Log options `-error`, `-warn`, `-info`, `-debug` are added as shorthand for `"early(error)"` etc.
 
 #### Additional import required
 
--- a/main-command/src/main/scala/sbt/BasicCommandStrings.scala
+++ b/main-command/src/main/scala/sbt/BasicCommandStrings.scala
@@ -60,21 +60,18 @@ ${runEarly(level.toString)}
 		* if no other commands are passed, interactive mode is still entered
 """
 
-  def runEarly(command: String) = {
-    val sep = if (command.isEmpty || Character.isLetter(command.charAt(0))) "" else " "
-    s"$EarlyCommand$sep$command"
-  }
+  def runEarly(command: String) = s"$EarlyCommand($command)"
   private[sbt] def isEarlyCommand(s: String): Boolean = {
-    s.startsWith(EarlyCommand) && s != Compat.FailureWall && s != Compat.ClearOnFailure
+    s.startsWith(EarlyCommand + "(") && s.endsWith(")")
   }
 
-  val EarlyCommand = "--"
-  val EarlyCommandBrief = (s"$EarlyCommand<command>", "Schedules a command to run before other commands on startup.")
+  val EarlyCommand = "early"
+  val EarlyCommandBrief = (s"$EarlyCommand(<command>)", "Schedules a command to run before other commands on startup.")
   val EarlyCommandDetailed =
-    s"""$EarlyCommand<command>
+    s"""$EarlyCommand(<command>)
 
 	Schedules an early command, which will be run before other commands on the command line.
-	The order is preserved between all early commands, so `sbt --a --b` executes `a` and `b` in order.
+	The order is preserved between all early commands, so `sbt "early(a)" "early(b)"` executes `a` and `b` in order.
 """
 
   def ReadCommand = "<"
--- a/main-command/src/main/scala/sbt/BasicCommandStrings.scala
+++ b/main-command/src/main/scala/sbt/BasicCommandStrings.scala
@@ -52,7 +52,7 @@ object BasicCommandStrings {
 	This will be used as the default level for logging from commands, settings, and tasks.
 	Any explicit `logLevel` configuration in a project overrides this setting.
 
-${runEarly(level.toString)}
+-$level
 
 	Sets the global logging level as described above, but does so before any other commands are executed on startup, including project loading.
 	This is useful as a startup option:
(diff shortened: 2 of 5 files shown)


dwijnand (changes requested):
Why not keep these as `--debug`/`--warn`/`--error`? `-debug`/`-warn`/`-error` is less UNIX-y that that.

This way we remove `--` from making everything early to just special cases.

eed3si9n:
It's true that general Unix convention would be to use double hyphen for long options, but [java](https://docs.oracle.com/javase/8/docs/technotes/tools/unix/java.html), scala, and sbt have been using single hyphen for options. See https://github.com/sbt/sbt-launcher-package/blob/master/src/universal/bin/sbt

```
  -d | -debug        set sbt log level to debug
```

dwijnand:
That is true.. But what's wrong with keeping it as `--error`? It conveys the right meaning, it doesn't clash with anything, and most importantly it's backwards compatible.

eed3si9n:
I'd keep it for 0.13 for compat but I don't really see strong reason to keep it for 1.0.
`early(command)` is the generic handler, and `-error` would be a new option that's consistent with every other options. The reason for removal is the same reason for removing `<<=`, which is reducing the surface area.

dwijnand:
Ok. I think it's much better if we don't do these breaking changes in point releases, but instead documented in the upgrade/migration notes. Thanks

dwijnand (approved):

dwijnand:
For clarity, I think we should add notes here.

eed3si9n:
Added migration notes - 9783ab176521a65eaa19b9013ea906a32c6f65b3

retronym:
Over in scala/scala#6256, we found that:

```
$ sbt '--warn' 'eval ???'; echo $?
[warn] The `-` command is deprecated in favor of `onFailure` and will be removed in a later version
[error] scala.NotImplementedError: an implementation is missing
[error] 	at scala.Predef$.$qmark$qmark$qmark(Predef.scala:284)
[error] 	at $b7f279a4d8e441649156$.$sbtdef(<eval>:1)
[error] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[error] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[error] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[error] 	at java.lang.reflect.Method.invoke(Method.java:498)
[error] 	at sbt.compiler.Eval$.getValue(Eval.scala:551)
[error] 	at sbt.compiler.Eval.$anonfun$eval$1(Eval.scala:121)
[error] 	at sbt.BuiltinCommands$.$anonfun$loadedEval$1(Main.scala:383)
[error] 	at sbt.internal.util.ManagedLogger.log(ManagedLogger.scala:24)
[error] 	at sbt.util.Logger.info(Logger.scala:22)
[error] 	at sbt.BuiltinCommands$.loadedEval(Main.scala:383)
[error] 	at sbt.BuiltinCommands$.$anonfun$eval$1(Main.scala:374)
[error] 	at sbt.Command$.$anonfun$apply1$2(Command.scala:146)
[error] 	at sbt.MainLoop$.processCommand(MainLoop.scala:154)
[error] 	at sbt.MainLoop$.$anonfun$next$2(MainLoop.scala:137)
[error] 	at sbt.State$$anon$1.runCmd$1(State.scala:242)
[error] 	at sbt.State$$anon$1.process(State.scala:248)
[error] 	at sbt.MainLoop$.$anonfun$next$1(MainLoop.scala:137)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:16)
[error] 	at sbt.MainLoop$.next(MainLoop.scala:137)
[error] 	at sbt.MainLoop$.run(MainLoop.scala:130)
[error] 	at sbt.MainLoop$.$anonfun$runWithNewLog$1(MainLoop.scala:108)
[error] 	at sbt.io.Using.apply(Using.scala:22)
[error] 	at sbt.MainLoop$.runWithNewLog(MainLoop.scala:102)
[error] 	at sbt.MainLoop$.runAndClearLast(MainLoop.scala:58)
[error] 	at sbt.MainLoop$.runLoggedLoop(MainLoop.scala:43)
[error] 	at sbt.MainLoop$.runLogged(MainLoop.scala:35)
[error] 	at sbt.StandardMain$.runManaged(Main.scala:113)
[error] 	at sbt.xMain.run(Main.scala:76)
[error] 	at xsbt.boot.Launch$$anonfun$run$1.apply(Launch.scala:109)
[error] 	at xsbt.boot.Launch$.withContextLoader(Launch.scala:128)
[error] 	at xsbt.boot.Launch$.run(Launch.scala:109)
[error] 	at xsbt.boot.Launch$$anonfun$apply$1.apply(Launch.scala:35)
[error] 	at xsbt.boot.Launch$.launch(Launch.scala:117)
[error] 	at xsbt.boot.Launch$.apply(Launch.scala:18)
[error] 	at xsbt.boot.Boot$.runImpl(Boot.scala:41)
[error] 	at xsbt.boot.Boot$.main(Boot.scala:17)
[error] 	at xsbt.boot.Boot.main(Boot.scala)
[error] scala.NotImplementedError: an implementation is missing
0
```

gives a zero error code, because this is now the same as `sbt 'onFailure -warn' ...`.

I'll assume that this was unintentional. This is quite dangerous, as `sbt-extra` [lists `-w` in its help ](https://github.com/paulp/sbt-extras/blob/master/sbt#L289-L302)which it translates to `--warn`. Both sbt-extras and the official sbt launcher passes through `--warn`, which is was our scripts have been using.

It is quite easy to miss or ignore the warning, and then run into the trap of having build scripts pass when a test fails.

Thoughts? I'd recommend adding `--warn` etc as aliases, possibly deprecated.

eed3si9n:
`onFailure` kicking in was unintentional. Maybe it's time to remove `-` command too.
For compat (and also for JEP 293), I'd be happy to bring back `--warn` as special case of `early(warn)`.

eed3si9n:
Opened an issue for this https://github.com/sbt/sbt/issues/4088

PR #2752, opened 2016-09-23
[sbt 1.0] Improve `show` when key returns a `Seq`

eed3si9n:
This changes the output to:

```
> show externalDependencyClasspath
[info] Updating {file:/xxx/hello/}root...
[info] Resolving jline#jline;2.12.1 ...
[info] Done updating.
[info] * Attributed(/Users/xxx/.sbt/boot/scala-2.11.8/lib/scala-library.jar)
[info] * Attributed(/Users/xxx/.ivy2/cache/commons-io/commons-io/jars/commons-io-
2.4.jar)
```

/review @dwijnand, @Duhemm


dwijnand (approved):
Amazingly beautiful and amazingly simple.

dwijnand:
Can we backport this too?

eed3si9n:
> Can we backport this too?

I'm not sure if we count this as semantic change. If we do we might have to guard it with an opt-out flag.

dwijnand:
I'm happy with an opt-out.

PR #2760, opened 2016-09-28
FPORT: Add extraProjects adn derivedProjects. Fixes #2532

### LINKED ISSUE #2532: Ability to programmatically define projects: Seq[Project]

dwijnand:
This adds support to generate synthetic subprojects from an auto plugin.

In addition, a method called `projectOrigin` is added to distinguish Organic, BuildExtra, ProjectExtra, and GenericRoot.

Forward-port of #2717 and #2738

### DIFF
--- a/main/src/main/scala/sbt/Plugins.scala
+++ b/main/src/main/scala/sbt/Plugins.scala
@@ -93,6 +93,12 @@ abstract class AutoPlugin extends Plugins.Basic with PluginsFunctions {
 
   // TODO?: def commands: Seq[Command]
 
+  /** The [[Project]]s to add to the current build. */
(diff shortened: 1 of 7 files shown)


dwijnand:
Green! @eed3si9n your scripted test works, but do you want to glance at this, given it was a little bit of a tricky cherry-pick?

eed3si9n (approved):

PR #2769, opened 2016-10-05
Migrate to blackbox.Context

dwijnand:
Requires a release of sbt/util.

@eed3si9n should I release 0.1.0-M14? or 0.1.0-X1? or 1.0.0-X1? Thanks


dwijnand:
Oh, there is already a 0.1.0-M14 with the necessary changes (just the git tag is missing from GitHub).

PR #2771, opened 2016-10-07
Workaround #1696 & bump to JDK 1.8

dwijnand:


dwijnand:
Required to unfork https://github.com/sbt/sbt-standalone-build/blob/2d0ba76f795a2b76ffa80ca465a807fec40b9a05/sbt-standalone.dbuild#L14

Issue #2781, opened 2016-10-13
Tab completion with lots of projects with similar names is super slow


bantonsson:
## steps

Tested with 0.13.12 but the older versions are slow as well.

Set up a build sbt with lots of projects (100 to be exact) [build.sbt](https://gist.github.com/bantonsson/d2e82529980fd20991a4aaa722243f67#file-build-sbt)

Start sbt prompt do the following:

```
> pref<TAB>
```
## problem

After pressing `<TAB>` for completion you have enough time to drink your coffee while sbt is working before the common prefix of all the projects is filled in.
## expectation

After pressing `<TAB>` sbt should fill in the common prefix of the projects without working for seconds (producing a ton of garbage).
## notes

I debugged this and most of the time is being spent in `EditDistance.levenshtein` and my first reaction was to optimize that method. Even if that method got faster (but uglier) it didn't have much of an effect on the overall run time.

Here is a [stack trace](https://gist.github.com/bantonsson/d2e82529980fd20991a4aaa722243f67#file-slow-sbt-tab-completion-stack-trace-and-project) of where the distance calculation is triggered.

Turns out that the errors are eagerly evaluated 7 times (yes that is seven, and I still don't know why seven) for every project that partially matches, basically computing the Levenshtein distance 7 \* 100 (the number of projects with partial match) \* 507 (the number of commands and settings that we check the distance against) times.
## solution

I think I have a fix for the eager evaluation. Where should I open a PR?

dwijnand:
Open it here against 0.13.

I wonder if https://github.com/sbt/sbt/pull/2547 fixes it for you.

bantonsson:
Will open a PR against 0.13.

I'll also try that fix locally, but I'm not sure that is the issue here. We'll see.

bantonsson:
The `validId` fix is not related to this issue.

PR #2782, opened 2016-10-13
Lazily concatenate failed errors for completion

bantonsson:
- Fixes [sbt/sbt#2781]
- When using `<TAB>` completion the failed errors were always
  computed for mathcing projects even if there was no failure,
  leading to excessive computation of Levenshtein distances
  and a large lag (seconds) on builds with many matching
  projects.


dwijnand:
So subtle :)

Any drawbacks from lazily concatenating the failed errors?

bantonsson:
@dwijnand Not that I can come to think of. Looking at all the other code that goes out of its way to be call by name and lazy evaluation, I think that this was just a slip-up.

dwijnand (approved):
Ok, thanks. LGTM.

dwijnand:
Up to @eed3si9n if he feels like squeezing this into 0.13.13 (retargeting the PR to the 0.13.13 branch)

eed3si9n:
I'm going to merge it so I can try this Github feature :)

eed3si9n:
Could you or Dale forward port this to sbt/util plz?

PR #2785, opened 2016-10-16
Adding column number information into sbt output

VlachJosef:
Currently sbt output lacks column number information easily
accessible for cli tools.

This change is adding column number just behind line number

```
[error] /Users/me/column/src/main/scala/Column.scala:5:31: type mismatch;
```

### DIFF
--- a/compile/src/main/scala/sbt/LoggerReporter.scala
+++ b/compile/src/main/scala/sbt/LoggerReporter.scala
@@ -93,7 +93,8 @@ class LoggerReporter(maximumErrors: Int, log: Logger, sourcePositionMapper: Posi
       log(msg)
     else {
       val sourcePrefix = m2o(pos.sourcePath).getOrElse("")
-      val lineNumberString = m2o(pos.line).map(":" + _ + ":").getOrElse(":") + " "
+      val columnNumber = m2o(pos.pointer).map(_.toInt + 1).getOrElse(1)
+      val lineNumberString = m2o(pos.line).map(":" + _ + ":" + columnNumber + ":").getOrElse(":") + " "
       log(sourcePrefix + lineNumberString + msg)
       val lineContent = pos.lineContent
       if (!lineContent.isEmpty) {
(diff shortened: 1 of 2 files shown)


eed3si9n:
I don't know if people currently rely on the existing output, so it might be prudent to target sbt 1.0 for this change?

VlachJosef:
Yes, I agree... it may be a brittle change so I'm ok with putting it into 1.0 branch.

compile/src/main/scala/sbt/LoggerReporter.scala
```diff
@@ -93,7 +93,8 @@ class LoggerReporter(maximumErrors: Int, log: Logger, sourcePositionMapper: Posi
       log(msg)
     else {
       val sourcePrefix = m2o(pos.sourcePath).getOrElse("")
-      val lineNumberString = m2o(pos.line).map(":" + _ + ":").getOrElse(":") + " "
+      val columnNumber = m2o(pos.pointer).map(_.toInt + 1).getOrElse(1)
```
dwijnand, line 96: Why the `+ 1` and defaulting to 1? Shouldn't it stay 0-based like line number?

VlachJosef: This took me also by surprise, but in this way it is consistent with other tools like [the_silver_searcher](https://github.com/ggreer/the_silver_searcher) which count columns from 1, for example: 

```
$ ag --nogroup --line-number --column package .
main/src/main/scala/sbt/internal/BuildUtil.scala:1:1:package sbt
main/src/main/scala/sbt/internal/BuildUtil.scala:2:1:package internal
```

Also vim seems to count them from 1, in emacs land it seems little [inconsistent](https://github.com/bbatsov/rubocop/issues/276), but `next-error` command counts them also from 1.

dwijnand:
To move this to 1.0 you need to move this PR to sbt/zinc where LoggerReporter is:

https://github.com/sbt/zinc/blob/1.0/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/LoggerReporter.scala

VlachJosef:
PR moved into https://github.com/sbt/zinc/pull/183

fommil:
@VlachJosef can you please add a pre-emptive PR to `sbt-mode` to support this?

PR #2791, opened 2016-10-20
Add new param to javadoc.run due to interface change

peiyuwang:
This combined with https://github.com/sbt/zinc/pull/186 should fix https://jenkins.scala-sbt.org:8592/job/sbt-standalone-build/40/consoleFull

Test:
- `sbt test` succeeded, previously fail with above error.


eed3si9n:
Thanks!

eed3si9n:
We finally got around to bumping everything up to the latest in https://github.com/sbt/sbt/pull/2892 and the changes proposed here were addressed in https://github.com/sbt/sbt/commit/7fcfec8b8e2b58b33b028606ecafc8ae03f7cdd4

PR #2794, opened 2016-10-20
Cross build io (+ control) for playframework/playframework#6110

dwijnand:


eed3si9n:
I'm targeting 0.13.13

eed3si9n:
Hmm this is actually going to break `so publishSigned` because of Java 1.6 vs 1.8.

eed3si9n:
crossBuild needs to be changed as follows in `project/Util.scala`:

``` scala
  def crossBuild: Seq[Setting[_]] =
    Seq(
      crossPaths := (scalaBinaryVersion.value match {
        case "2.10" => false
        case _      => true
      })
    )
```

PR #2827, opened 2016-11-11
Apply scalaVersion enforcement to Compile related configs

### LINKED ISSUE #2786: 0.13.12 regression in resolving org.scala-lang:scala-library in custom configuration
## steps

Repo to reproduce: https://github.com/olafurpg/scalafmt/files/521826/issue485.zip
## problem

Running `scalafmt` results in errors

```
[error] Modules were resolved with conflicting cross-version suffixes in {file:/Users/heiko/projects/scala/new-in-scala-212/}new-in-scala-212:
[error]    org.scala-lang.modules:scala-xml _2.11, _2.12.0-RC1

OR

java.lang.NoClassDefFoundError: scala/Product$class
    at org.scalafmt.config.Docstrings$ScalaDoc$.<init>(Docstrings.scala:8)
    at org.scalafmt.config.Docstrings$ScalaDoc$.<clinit>(Docstrings.scala)
    at org.scalafmt.config.ScalafmtConfig$.apply$default$2(ScalafmtConfig.scala:2)
    at org.scalafmt.config.Settings$class.$init$(Settings.scala:17)
    at org.sc

```

I am unable to reproduce the issue with `sbt.version=0.13.11` in build.properties.

I tried the following with no success:

``` scala
dependencyOverrides += "org.scala-lang" % "scala-library" % org.scalafmt.Versions.scala % "scalafmt"
// and
libraryDependencies += "org.scala-lang" % "scala-library" % org.scalafmt.Versions.scala % "scalafmt" force()
```

However, changing to typelevel/scala fixes the issue (PR: https://github.com/olafurpg/scalafmt/pull/522):

``` scala
libraryDependencies += "org.typelevel" % "scala-library"     % org.scalafmt.Versions.scala   % "scalafmt"
```
## expectation

I expected the `scalafmt` task to run without error.

More specifically, I expect the 0.13.12 resolver to not evict `org.scala-lang:scala-library:2.11.8` from the `scalafmt` configuration in favor of `org.scala-lang:scala-libary:2.12.0-RC1` defined in the `default`/`compile` configuration.
## notes
- Potential cause of this issue: https://github.com/sbt/sbt/pull/2634
- Original scalafmt issue: https://github.com/olafurpg/scalafmt/issues/485#issuecomment-253978435
- Scalafmt is 2.11 only. To overcome the 2.10 requirement for SBT plugins, the scalafmt SBT plugin uses a custom configuration to classload scalafmt with 2.11.8 scala-library.
- Attempts to port scala.meta (a scalafmt dependency) to 2.10 have so far been unsuccessful: https://github.com/scalameta/scalameta/issues/295

eed3si9n:
Fixes #2786. Ref #2634.

sbt 0.13.12 added Ivy mediator that enforces scalaOrganization and
scalaVersion for Scala toolchain artifacts.
This turns out to be a bit too aggressive because Ivy configurations
can be used as an independent dependency graph that does not rely on
the scalaVersion used by Compile configuration. Enforcing
scalaVersion in those graphs causes runtime failure.

This change checks if the configuration extends Default, Compile,
Provided, or Optional before enforcing scalaVersion.

/review @dwijnand

### DIFF
--- a/ivy/src/main/scala/sbt/Configuration.scala
+++ b/ivy/src/main/scala/sbt/Configuration.scala
@@ -51,6 +51,15 @@ object Configurations {
   private[sbt] def DefaultConfiguration(mavenStyle: Boolean) = if (mavenStyle) DefaultMavenConfiguration else DefaultIvyConfiguration
   private[sbt] def defaultConfiguration(mavenStyle: Boolean) = if (mavenStyle) Configurations.Compile else Configurations.Default
   private[sbt] def removeDuplicates(configs: Iterable[Configuration]) = Set(scala.collection.mutable.Map(configs.map(config => (config.name, config)).toSeq: _*).values.toList: _*)
+
+  /** Returns true if the configuration should be under the influence of scalaVersion. */
+  private[sbt] def underScalaVersion(c: Configuration): Boolean =
+    c match {
+      case Default | Compile | IntegrationTest | Provided | Runtime | Test | Optional |
+        CompilerPlugin | CompileInternal | RuntimeInternal | TestInternal => true
+      case config =>
+        config.extendsConfigs exists underScalaVersion
+    }
 }
 /** Represents an Ivy configuration. */
 final case class Configuration(name: String, description: String, isPublic: Boolean, extendsConfigs: List[Configuration], transitive: Boolean) {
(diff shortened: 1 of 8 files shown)


dwijnand (approved):
Is there any way that that list of config can be made configurable by the build user?

dwijnand:
Also I think we should add a little test case, that emulates scalafmt's use case. To avoid this regressing.

eed3si9n:
I tried adding testing but I couldn't get it to fail without using 2.12, which requires JDK8.

I also tried exposing configurations setting but without a major hack like InlineConfigurationWithExcludes, I don't think we can easily in sbt 0.13.

milessabin:
Many thanks for dealing with this ... my apologies for not being able to find the time to get to it before you did :-(

eed3si9n:
@milessabin no problemo :)

PR #2844, opened 2016-11-23
Add props to define source dependencies on io/util/lm/zinc

dwijnand:
Use either:

* `-Dsbtio.path` / `-Dsbtutil.path` / `-Dsbtlm.path` / `-Dsbtzinc.path` on the command line, or
* `-Dsbtio.path` / `-Dsbtutil.path` / `-Dsbtlm.path` / `-Dsbtzinc.path` in project/local.properties

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -195,7 +195,7 @@ lazy val mainSettingsProj = (project in file("main-settings")).
       utilLogging, sbtIO, utilCompletion, compilerClasspath, libraryManagement)
   )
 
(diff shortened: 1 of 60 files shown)


eed3si9n:
Not a huge fan of this change as this makes our build non-idiomatic.
We can merge this for the sake of moving sbt 1 forward, but we should come back to this at some point to support source dependency as the feature, so we can remove the hacks and duplicates.

eed3si9n (approved):

PR #2852, opened 2016-11-29
Back out build-level key completion

### LINKED ISSUE #2851: Calling build-level keys at a subproject resolves to wrong scoping
Ref #2707, #2708, #2469

## steps

Using a multi-project build with sbt 0.13.13, navigate to a subproject and call `version`.

```
contraband root> project library
[info] Set current project to contraband (in build file:/Users/xxx/work/contraband/)
contraband> version
[info] library/*:version
[info] 	0.3.0-M1
[info] plugin/*:version
[info] 	0.3.0-M1
[info] {.}/*:version
[info] 	0.3.0-M1
contraband> inspect version
[info] Setting: java.lang.String = 0.3.0-M1
[info] Description:
[info] 	The version/revision of the current module.
[info] Provided by:
[info] 	{file:/Users/xxx/work/contraband/}/*:version
[info] Defined at:
[info] 	/Users/xxx/work/contraband/build.sbt:13
[info] Delegates:
[info] 	{.}/*:version
[info] 	*/*:version
[info] Related:
[info] 	*/*:version
```

## problem

The problem is that typing "version" seems to be resolving to some build level scoping, and it ends up aggregating calls to the build level and the subprojects.

## expectation

Here's the result using sbt 0.13.11:

```
contraband root> project library
[info] Set current project to contraband (in build file:/Users/xxx/work/contraband/)
contraband> version
[info] 0.3.0-M1
contraband> inspect version
[info] Setting: java.lang.String = 0.3.0-M1
[info] Description:
[info]  The version/revision of the current module.
[info] Provided by:
[info]  {file:/Users/xxx/work/contraband/}/*:version
[info] Defined at:
[info]  /Users/xxx/work/contraband/build.sbt:13
[info] Reverse dependencies:
[info]  library/*:bintraySyncMavenCentral
[info]  library/*:projectId
[info]  library/*:bintrayRemoteSign
[info]  library/*:isSnapshot
[info]  library/*:bintrayRelease
[info]  library/*:bintrayUnpublish
[info]  library/*:bintray::publishTo
[info] Delegates:
[info]  library/*:version
[info]  {.}/*:version
[info]  */*:version
[info] Related:
[info]  */*:version
[info]  {.}/*:version
```

## notes

This behavior may result to unintended side effect aggregating over the subprojects.

eed3si9n:
Fixes #2851
Ref #2707, #2708, #2469

Given the output from sbt 0.13.13's response to `version`, the logic that is used to tab-complete, and eventually resolve the build level key seems flawed.

```
contraband root> project library
[info] Set current project to contraband (in build file:/Users/xxx/work/contraband/)
contraband> version
[info] library/*:version
[info] 	0.3.0-M1
[info] plugin/*:version
[info] 	0.3.0-M1
[info] {.}/*:version
[info] 	0.3.0-M1
```

I'm guessing that the unscoped key should never result to a build-level key.
A potential fix might be to emulate the delegation logic while enumerating the project keys, or scope the resolved build-level key to the current subproject.

/cc @Duhemm

### DIFF
--- a/main/src/main/scala/sbt/Act.scala
+++ b/main/src/main/scala/sbt/Act.scala
@@ -77,16 +77,11 @@ object Act {
       selectFromValid(ss filter isValid(data), default)
     }
   def selectFromValid(ss: Seq[ParsedKey], default: Parser[ParsedKey])(implicit show: Show[ScopedKey[_]]): Parser[ParsedKey] =
-    selectByTask(selectByConfig(ss)) partition isBuildKey match {
-      case (_, Seq(single))         => success(single)
-      case (Seq(single), Seq())     => success(single)
-      case (Seq(), Seq())           => default
-      case (buildKeys, projectKeys) => failure("Ambiguous keys: " + showAmbiguous(keys(buildKeys ++ projectKeys)))
+    selectByTask(selectByConfig(ss)) match {
+      case Seq()       => default
+      case Seq(single) => success(single)
+      case multi       => failure("Ambiguous keys: " + showAmbiguous(keys(multi)))
     }
-  private def isBuildKey(parsed: ParsedKey): Boolean = parsed.key.scope.project match {
-    case Select(_: BuildReference) => true
-    case _                         => false
-  }
   private[this] def keys(ss: Seq[ParsedKey]): Seq[ScopedKey[_]] = ss.map(_.key)
   def selectByConfig(ss: Seq[ParsedKey]): Seq[ParsedKey] =
     ss match {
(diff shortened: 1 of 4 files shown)


dwijnand:
Doesn't this revert the fix for https://github.com/sbt/sbt/issues/2460?

Are you saying that @dotta's issue is "working as intended"?

eed3si9n:
#2460 is a legitimate concern, but the fixes that were implemented is causing more bugs in my opinion. We could keep this open while we try to find a better fix, and if we can't merge this.
But the general direction of allowing to resolve to a build-level key (without scoping) is probably wrong.

PR #2859, opened 2016-12-02
Start lightweight client

eed3si9n:
This is the beginning of a lightweight client, which talks to the
server over Contraband-generated JSON API. Given that the server is
started on port 5173:

```
$ cd /tmp/bogus
$ sbt client localhost:5173
...
[info] Set current project to bogus (in build file:/private/tmp/bogus/)
client on port 5173
> compile
StatusEvent(Processing, Vector(compile, server))
StatusEvent(Ready, Vector())
StatusEvent(Processing, Vector(, server))
StatusEvent(Ready, Vector())
> clean
StatusEvent(Processing, Vector(clean, server))
StatusEvent(Ready, Vector())
StatusEvent(Processing, Vector(, server))
StatusEvent(Ready, Vector())
```

It's not much, but at least it can process user input, and send it over to the server.


protocol/src/main/contraband/server.contra
```diff
@@ -0,0 +1,47 @@
...
+  commandQueue: [String]
+}
+
+# enum Status {
+#  Ready
+#  Processing
+# }
+
+## Executon event.
+type ExectionEvent {
```
dwijnand: Typo. Execution. Also in the comment above.

dwijnand (approved):

PR #2892, opened 2017-01-05
-sbinary/+sjson-new, -datatype/+contraband & upgrades

dwijnand:
* start to replace sbinary with sjson-new
* upgrade from sbt/datatype to sbt/contraband
* upgrade and migrate to new sbt modules APIs

[This is going to fail Travis CI because it needs proper released versions of sbt/lm and sbt/zinc]


eed3si9n:
Formatting issue?:

```
[error] (sbtRoot/*:scalariformCheck) Working directory is dirty!
[error] diff --git a/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala b/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
[error] index f693aa3..ca0898c 100644
[error] --- a/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
[error] +++ b/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
[error] @@ -90,7 +90,7 @@ class JUnitXmlTestsListener(val outputDir: String) extends TestsListener {
[error]                                                       case TStatus.Failure if (e.throwable.isDefined)=> <failure message={ e.throwable.get.getMessage } type={ e.throwable.get.getClass.getName }>{ trace }</failure>
[error]                                                       case TStatus.Failure=> <failure message={ "No Exception or message provided" }/>
[error]                                                       case TStatus.Ignored | TStatus.Skipped | TStatus.Pending=> <skipped/>
[error] -                                                     case _=> {}
[error] +                                                     case _    => {}
[error]                                                     }
[error]                                                   }
[error]                                                 </testcase>
```

dwijnand:
I'm confused. That's https://github.com/sbt/sbt/commit/0619f75024b6952dba8973381d3fe4fee4f9334d. I'll just drop that commit.

eed3si9n (approved):

PR #2902, opened 2017-01-10
Automatically choose proper compiler bridge for dotty

smarter:
Before this commit, using dotty in your sbt project required to add:
```scala
  scalaCompilerBridgeSource := ("ch.epfl.lamp" % "dotty-sbt-bridge" %
    scalaVersion.value % "component").sources()
```
in your build.sbt. We might as well automatically do this, this reduces
the boilerplate for using dotty in your project to:
```scala
  scalaOrganization := "ch.epfl.lamp"
  scalaVersion := "0.1.1-SNAPSHOT"
  scalaBinaryVersion := "2.11" // dotty itself is only published as a
                               // 2.11 artefact currently
```

Note: I did not add a test for this since there is no released version of dotty yet, we could have a test that depends on a nightly build of dotty (http://central.maven.org/maven2/ch/epfl/lamp/dotty_2.11/) but I'm not sure if maven keeps old nightly builds around forever.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -236,7 +236,13 @@ object Defaults extends BuildCommon {
       val _ = clean.value
       IvyActions.cleanCachedResolutionCache(ivyModule.value, streams.value.log)
     },
-    scalaCompilerBridgeSource := ModuleID(xsbti.ArtifactInfo.SbtOrganization, "compiler-interface", sbtVersion.value, Some("component")).sources()
+    scalaCompilerBridgeSource := {
+      if (ScalaInstance.isDotty(scalaVersion.value))
+        // Maintained at https://github.com/lampepfl/dotty/tree/master/sbt-bridge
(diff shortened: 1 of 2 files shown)


dwijnand (commented):
I'll defer to @eed3si9n, as he understands how compiler bridges are meant to be used and customized.

smarter:
The single test failure appears to be spurious and is unrelated to this PR:
```scala
[info] Could not create file /tmp/sbt_ecf6163e/java-mixed/target/streams/$global/ivySbt/$global/streams/outjava.io.IOException: No such file or directory
```

eed3si9n (changes requested):
I think it's generally ok, but could you add release notes in https://github.com/sbt/sbt/tree/0.13/notes/0.13.14 and clarify in there that Dotty support in sbt is an experimental feature that's controlled by EPFL, plz?

smarter:
Release notes added!

eed3si9n (approved):
Thanks!

dwijnand:
`->compile,default` here vs `->default(compile)` in #2903... which one should it be?

smarter:
Oops, that commit should not be there

smarter:
Fixed

PR #2903, opened 2017-01-10
Fix support for published dotty nightly builds

smarter:
Dotty nightly builds are published to maven, so they end up in
configuration "default", not "compile". We still need to look into
"compile" when dotty is published locally.

You can test this using https://github.com/smarter/dotty-example-project
by updating the sbt version used in project/build.properties and by
replacing "0.1.1-SNAPSHOT" by a nightly version like
"0.1.1-20170109-be64643-NIGHTLY" in build.sbt

### DIFF
--- a/ivy/src/main/scala/sbt/IvyScala.scala
+++ b/ivy/src/main/scala/sbt/IvyScala.scala
@@ -29,7 +29,7 @@ object ScalaArtifacts {
 
   private[sbt] def toolDependencies(org: String, version: String, isDotty: Boolean = false): Seq[ModuleID] =
     if (isDotty)
-      Seq(ModuleID(org, DottyIDPrefix, version, Some(Configurations.ScalaTool.name + "->compile"),
+      Seq(ModuleID(org, DottyIDPrefix, version, Some(Configurations.ScalaTool.name + "->default(compile)"),
         crossVersion = CrossVersion.binary))
     else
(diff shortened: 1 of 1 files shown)


smarter:
More spurious failures "Could not create file: ..." like in https://github.com/sbt/sbt/pull/2902 but no real failure as far as I can tell.

ivy/src/main/scala/sbt/IvyScala.scala
```diff
@@ -29,7 +29,7 @@ object ScalaArtifacts {
 
   private[sbt] def toolDependencies(org: String, version: String, isDotty: Boolean = false): Seq[ModuleID] =
     if (isDotty)
-      Seq(ModuleID(org, DottyIDPrefix, version, Some(Configurations.ScalaTool.name + "->compile"),
+      Seq(ModuleID(org, DottyIDPrefix, version, Some(Configurations.ScalaTool.name + "->compile,default"),
```
eed3si9n: Should this be `default(compile)`?

eed3si9n (changes requested):

ivy/src/main/scala/sbt/IvyScala.scala
smarter: this should work too yeah, I'll try it.

smarter: Done.

eed3si9n (approved):

PR #2907, opened 2017-01-13
Split log output per channel

### LINKED ISSUE #2800: Split terminal output per-channel
Given https://github.com/sbt/sbt/issues/2798
## steps
1. Start sbt server on terminal1 where a build is located: sbt server
2. Start sbt client on terminal2 from anywhere like ~/temp/: sbt client localhost:5001
3. On terminal1 `project echo-server`
4. On terminal2 `project echo-client`
5. On terminal1 type `hello` (let's say hello is a command that logs "hello")
6. On terminal2 type `hi` (let's say hello is a command that logs "hi")
## expectation

On terminal1 I see "hello", on terminal2 I see "hi", but I don't see each other's output.

eed3si9n:
LogManager implementation is modified to use ManagedLogger, which can swap out backing Appenders without re-creating the log instance.

The State was also changed to track `currentCommand: Option[Exec]`. `Exec` knows the origin of the command invocation, and using that we can now send the network-originated events only to the network clients.

Combined together, this implements log splitting between the sbt clients (channels).

Requires https://github.com/sbt/util/pull/68
Fixes #2800


protocol/src/main/scala/sbt/protocol/Serialization.scala
```diff
@@ -46,7 +53,44 @@ object Serialization {
   /**
    * @return A command or an invalid input description
    */
-  def deserializeEvent(bytes: Seq[Byte]): Either[String, EventMessage] =
+  def deserializeEvent(bytes: Seq[Byte]): Either[String, Any] =
```
dwijnand, line 56: Can we not do any better than Any? Or is this WIP to be fixed at a later date?

dwijnand (approved):

protocol/src/main/scala/sbt/protocol/Serialization.scala
eed3si9n: Not really. Unless we limit our events to descend from some common trait we have to use Any here.

dwijnand:
> [error] x project / transitive-plugins (out-1)
[error]    {line 4}  Command failed: a/publish failed (out-1)

eed3si9n:
Yea. It's just missing https://repo1.maven.org/maven2/org/scala-sbt/librarymanagement_2.12/1.0.0-X4 from Maven Central.

PR #2933, opened 2017-01-20
Spelling

jsoref:


protocol/src/main/contraband-scala/sbt/protocol/codec/ExectionEventFormats.scala
```diff
@@ -5,21 +5,21 @@
 // DO NOT EDIT MANUALLY
 package sbt.protocol.codec
 import _root_.sjsonnew.{ deserializationError, serializationError, Builder, JsonFormat, Unbuilder }
-trait ExectionEventFormats { self: sjsonnew.BasicJsonProtocol =>
-implicit lazy val ExectionEventFormat: JsonFormat[sbt.protocol.ExectionEvent] = new JsonFormat[sbt.protocol.ExectionEvent] {
-  override def read[J](jsOpt: Option[J], unbuilder: Unbuilder[J]): sbt.protocol.ExectionEvent = {
+trait ExecutionEventFormats { self: sjsonnew.BasicJsonProtocol =>
```
jsoref: This is an API break afaict.

And yes, I can see there's a DO NOT EDIT MANUALLY marker.
I didn't edit it manually, a script did ;-)

1. If you're intentionally misspelling things, you should include a note explaining why.
1. If you're generating things, you should indicate the source, not just the tool.

sbt/src/sbt-test/actions/set/build.sbt
```diff
@@ -1,6 +1,6 @@
 
 TaskKey[Unit]("checkName", "") := {
-	assert(name.value == "hello-world", "Name is not hello-worled, failed to set!")
+	assert(name.value == "hello-world", "Name is not hello-world, failed to set!")
```
jsoref, line 3: I'm assuming this is a typo, if it isn't, it's pretty weird.

sbt/src/sbt-test/source-dependencies/macro-annotation/macros/src/main/scala/Macros.scala
```diff
@@ -3,24 +3,24 @@ import scala.language.experimental.macros
 import scala.annotation.StaticAnnotation
 
 object HelloMacro {
-  def impl(c: Context)(annottees: c.Tree*): c.Tree = {
+  def impl(c: Context)(annotatees: c.Tree*): c.Tree = {
```
jsoref: I get that scala seems to like this invented word. But it isn't right.
Note: as each commit covers a single concept, if you don't like a change, you can trivially drop the commit and move along.

sbt/src/sbt-test/source-dependencies/macro-annotation/macros/src/main/scala/Macros.scala
```diff
@@ -3,24 +3,24 @@ import scala.language.experimental.macros
...
     import c.universe._
 
-    annottees match {
+    annotatees match {
       case (classDecl: ClassDef) :: Nil =>
-        val q"$mods class $name[..$tparams] $ctorMods(...$paramss) extends { ..$earlydefns } with ..$bases { $self => ..$body }" = classDecl
+        val q"$mods class $name[..$tparams] $ctorMods(...$params) extends { ..$earlydefns } with ..$bases { $self => ..$body }" = classDecl
         q"""
-        case class $name(...$paramss) extends ..$bases {
+        case class $name(...$params) extends ..$bases {
```
jsoref: I can't tell why you have an extra `s` here. It could be intentional, but, I'm betting it isn't.
Note: if it is intentional, you could improve readability by expanding the `s` into something more...

eed3si9n (approved):
All the changes look fine. Thanks for doing this.

jsoref (commented):
Most of the rest of the changes are comments, local variables, or private functions. But you should review each word ...

jsoref:
@eed3si9n : thanks for being responsive.

protocol/src/main/contraband-scala/sbt/protocol/codec/ExectionEventFormats.scala
eed3si9n: This is a generated code, and an outdated one. This file should be removed since there's now https://github.com/sbt/sbt/blob/1.0.x/protocol/src/main/contraband-scala/sbt/protocol/codec/ExecutionEventFormats.scala

sbt/src/sbt-test/actions/set/build.sbt
eed3si9n: Typo.

sbt/src/sbt-test/source-dependencies/macro-annotation/macros/src/main/scala/Macros.scala
dwijnand: It is intentional, it's a list of lists of parameters. And it's intentionally terse (instead of "paramsList"). It's just one of those things that comes with the domain of introspecting code and/or generating code. We should revert this one.

sbt/src/sbt-test/source-dependencies/macro-annotation/macros/src/main/scala/Macros.scala
dwijnand: We should drop this one. "annottees" is simply what these are called now..

eed3si9n: If "annottees" is the standard in macro lingo, let's just stick to it. I don't want to use sbt source as a platform of protest.

sbt/src/sbt-test/source-dependencies/macro-annotation/macros/src/main/scala/Macros.scala
eed3si9n: When I see `params` I think of `List[X]`. When I see `paramss`, I think of `List[List[X]]`. That's fairly common convention for compiler code base.

eed3si9n (changes requested):
I guess I did have some change requests.

protocol/src/main/contraband-scala/sbt/protocol/codec/ExectionEventFormats.scala
jsoref: I'd rather someone do that in a distinct PR.

eed3si9n: That's fine.

dwijnand (changes requested):
Thanks @jsoref for the fixes.

dwijnand (approved):

eed3si9n (approved):

PR #2934, opened 2017-01-20
Move some server pieces from main-command to main

dwijnand:
Given the commented out code in main-command's Command.process this probably needs attention before merging.


main-command/src/main/scala/sbt/Command.scala
```diff
@@ -91,16 +91,16 @@ object Command {
   /** This is the main function State transfer function of the sbt command processing, called by MainLoop.next, */
   def process(exec: Exec, state: State): State =
     {
-      val channelName = exec.source map { _.channelName }
-      State.exchange.publishEventMessage(ExecStatusEvent("Processing", channelName, exec.execId, Vector()))
+      //      val channelName = exec.source map { _.channelName }
+      //      State.exchange.publishEventMessage(ExecStatusEvent("Processing", channelName, exec.execId, Vector()))
```
eed3si9n: We still need this, so we might have to move `Command.process` to somewhere else.

eed3si9n (changes requested):

eed3si9n (approved):

PR #2935, opened 2017-01-20
Query settings

dwijnand:
Builds on #2934.

WIP, just to share with @eed3si9n.


main/src/main/scala/sbt/internal/CommandExchange.scala
```diff
@@ -74,7 +74,7 @@ private[sbt] final class CommandExchange {
       def onIncomingSocket(socket: Socket): Unit =
         {
           s.log.info(s"new client connected from: ${socket.getPort}")
-          val channel = new NetworkChannel(newChannelName, socket)
+          val channel = new NetworkChannel(newChannelName, socket, s)
```
eed3si9n: NetworkChannel represents a connection from some user. I don't think a state object should be owned by it since it's sort of misleading since the "state" there represents a past state.
I think a better approach is to pass in `(onQuery: Query => Unit)` here that it can call back, and handle query in the `CommandExchange`.

main/src/main/scala/sbt/internal/server/NetworkChannel.scala
```diff
@@ -71,15 +71,24 @@ final class NetworkChannel(val name: String, connection: Socket, state: State) e
...
+  }
+
   def shutdown(): Unit = {
     println("Shutting down client connection")
     running.set(false)
     out.close()
   }
 }
+
+final case class GetSettingReq[A](settingKey: SettingKey[A]) extends CommandMessage
```
eed3si9n: 1. Probably move to protocol subproject. 
2. Rename to `SettingQuery`?
3. Avoid case class.

main/src/main/scala/sbt/internal/server/NetworkChannel.scala
```diff
@@ -71,15 +71,24 @@ final class NetworkChannel(val name: String, connection: Socket, state: State) e
...
+final case class GetSettingRsp[A](value: Option[A]) extends EventMessage
```
eed3si9n: 1. Probably move to protocol subproject.
2. Rename to `SettingEvent`?
3. Avoid case class.

eed3si9n (changes requested):
A few comments.

main/src/main/scala/sbt/internal/server/NetworkChannel.scala
dwijnand: Yeah, that's the WIP part. Not sure about "SettingEvent", could be confusing.

eed3si9n: `SettingQueryResponse`?

main/src/main/scala/sbt/internal/CommandExchange.scala
dwijnand: In terms of "past state" I think using `onQuery: Query => Unit ` would amount to the same thing.

Imo, let's go with what works, and then use test cases to correct it.

eed3si9n: It might work but I see it more of a breach of responsibility.

main/src/main/scala/sbt/internal/server/NetworkChannel.scala
dwijnand: It breaks my natural vertical alignment, but sure.

main/src/main/scala/sbt/internal/server/NetworkChannel.scala
dwijnand: Because SettingKey is defined in main-settings it's not trivial to move this to the protocol subproject.

dwijnand: Also can't define this in the Contraband Schema Language: https://github.com/sbt/contraband/issues/74

dwijnand:
Looks like this using `telnet`:

```
{"type":"SettingQuery","setting":"scalaVersion"}
{"type":"SettingQueryResponse","values":["2.12.1"]}
{"type":"SettingQuery","setting":"organization"}
{"type":"SettingQueryResponse","values":["com.dwijnand"]}
{"type":"SettingQuery","setting":"homepage"}
{"type":"SettingQueryResponse","values":["None"]}
```

eed3si9n:
Why do we get multiple values back? If it's because of project aggregation, we probably should remove that and force the client to pass in fully qualified settings.

dwijnand:
At least from aggregation. Maybe something else too. Delegation is also involved. I'm not sure if that's configuration delegation or project delegation.

Ultimately I was emulating show, except with structured dara rather than println toString.

By "fully qualified" do you mean it can't depend on the "current" project? What about depending on configuration delegation?

eed3si9n:
> By "fully qualified" do you mean it can't depend on the "current" project? 

No. The notion of current project shouldn't exist in querying as it's state dependent.

> What about depending on configuration delegation?

I think that's fine as long as it's stable. (It is right?)

dwijnand:
> > By "fully qualified" do you mean it can't depend on the "current" project?
> No. The notion of current project shouldn't exist in querying as it's state dependent.

OK, then that means creating alternative parsers for the string.

> > What about depending on configuration delegation?
> I think that's fine as long as it's stable. (It is right?)

It's not stable across reloads (or reboots) I assume, FWIW.

dwijnand:
@eed3si9n:

+ [x] `NetworkChannel` no longer owns a `State`
+ [x] data types moved to protocol project
+ [x] renamed to `SettingQuery`/`SettingQueryResponse`
+ [x] avoided case class, used CSL
+ [x] Removed use of aggregation
+ [x] Removed dependency on the current/session project, allows for explicit project/ThisBuild/Global

Could you give this a quick once over?

dwijnand:
Trailing concerns:

+ stale State (e.g change a file, reload on server, query settings returns old values)
+ proper structured JSON (i.e don't toString)

dwijnand:
I rebased on latest 1.0.x and remove the upgrade of the sbt-contraband plugin as it depends on a breaking bincompat change in sjson-new that was failing the tests, which I CBA to deal with right now.

> java.lang.NoClassDefFoundError: sjsonnew/LNil$

eed3si9n (approved):

eed3si9n:
Merged!

Issue #1444, opened 2014-07-10
Macro implementation when using triggers is not working


huntc:
This works:

```scala
discoveredDist <<= (dist in ReactiveRuntime)
  .storeAs(discoveredDist in Global)
  .triggeredBy(dist in ReactiveRuntime)
```

however this doesn't:

```scala
discoveredDist := (dist in ReactiveRuntime)
  .storeAs(discoveredDist in Global)
  .triggeredBy(dist in ReactiveRuntime)
  .value
```

liff:
Minimal repro:

``` scala
// build.sbt
val hello = taskKey[Unit]("hello")
val world = taskKey[Unit]("world")

hello := println("hello")
world := println("world")

world := (world triggeredBy hello).value // prints "hello", should print "hello\nworld"

// world <<= world triggeredBy hello // works
```

Run with `sbt hello`.

cunei:
It seems this issue is only incidentally related to the macro expansion. The code that is generated when this expression is expanded:
```
world := (world triggeredBy hello).value
```
in the end happens to be similar to:
```
world <<= ((world triggeredBy hello) map ( a => a ))
```
In the latter expression, the ``triggeredBy`` for some reason gets ignored; that happens on 0.12.x as well. I'm doing some debugging now.

eed3si9n:
### old logic

If we consider a task to consists of two parts, `value` function and *an attribute list*, the existing behavior is defensible. `<<=` binds both the parts to the LHS task key whereas `:=` creates a brand new task using the given value and a blank attribute list. 

In that logic,

```scala
world := (world triggeredBy hello).value
```

is not much different from

```scala
world := {
  val x = (world triggeredBy hello).value
  123
}
```

or 

```scala
world := 123
```

### Functor law

The part that starts to break down is `map`.
The first functor law states that mapping with `identity` should not modify the original functor.
See also http://eed3si9n.com/herding-cats/Functor.html
It might not be strictly enforced, but the law matches the intuitive feel people on the standard collections too.

### new ideas?

Even though it's forgetful, focusing only on the value using `:=` is very valuable in flattening the learning curve of sbt.

One workaround might be to track triggers as a separate key:

```scala
onComplete in hello += world
```

This would move the trigger as if they are values, and at the end of the loading we can put them to the appropriate task attributes.

gabro:
Is there a workaround to this issue that will achieve the same behavior without triggering a deprecation warning in sbt?

cunei:
[...]

PR #2942, opened 2017-01-23
[sbt 1.0] Support sourceGenerators += Def.task { ... }

eed3si9n:
This is on top of #2937

During the docs review, one of the comments we got from @lihaoyi was:

> Despite writing many sourceGenerators, it's not clear what this whole `Def.task{}.taskValue` thingy is about, even after reading this snippet. Why can we not directly `+=` the result of `myGenerator`?

A quick answer is because the types don't match up. The expected type for the generator is `Seq[Task[Seq[File]]]` whereas `Def.task { ... }` returns `Def.Initialize[Task[...]]`. `Initialize[_]` is sort of the DAG node, and `Task[_]` is like a beefed up `Function0[_]`.

Since we are pretty sure what user wants, we can extract the `Task[...]` out of `Initialize[Task[...]]`.

### Append instance that extracts taskValue

This adds a macro-level hack to support += op for sourceGenerators and resourceGenerators using RHS of Initialize[Task[Seq[File]]]. When the types match up, the macro now calls `.taskValue` automatically.


dwijnand (commented):
I don't know if we're improving or not the situation by adding a macro-level hack.

eed3si9n:
At least the end-user won't have to think about the subtlety of "task" vs `Def.Initialize[Task[...]]` vs `Task[...]`. I think that's a progress.

dwijnand:
This would be like a macro-level hack so you could do `scalaVersion := Some("2.12.1")` by just unwrapping the Some.

Personally I think this would be just as easily solved with a `Def.taskValue`, but feel free to overrule and merge.

eed3si9n:
My real target is the backport to 0.13.14, so we can address both of the issues that came up in 0.13.13 regarding `<<=` deprecation (this and #1444).

dwijnand:
I think instead we should provide `Def.taskValue { .. }` as a "solution" to the criticism on `Def.task { .. }.taskValue`. And then add `-Ywarn-value-discard` by default to the meta-build `scalacOptions`, like we added `-deprecation`.

eed3si9n:
I don't see `Def.taskValue { Nil }` as an improvement over `(Def.task { Nil }).taskValue` since you're still introducing additional concept to migrate from `<+=` without apparent return to the user. Worse, this would add discrepancy with task key. The test for this PR supports:

```scala
sourceGenerators in Compile += buildInfo,
sourceGenerators in Compile += Def.task { Nil }
```

dwijnand:
> Worse, this would add discrepancy with task key.

What discrepancy? What do you mean?

eed3si9n:
>> Worse, this would add discrepancy with task key.
>
> What discrepancy? What do you mean?

In general this is where the rubber meets the road, or (*insert impedance mismatch analogy*).
We teach people about settings vs tasks, there are keys specific to settings and tasks, but the actual types around "settings" and "tasks" are more complicated.
What we colloquially call "settings" are actually `Def.Initialize[A]`. Note `SettingKey[A]` extends `KeyedInitialize[A]`.

Similarly, what we colloquially call "tasks" are `Def.Initialize[Task[A]]`. `TaskKey[A]` extends `KeyedInitialize[Task[A]]`. This is how we are able to maintain the consistent feel of custom tasks `Def.task { ... }` and built-in task (keys).

My claim is that the build user expects `sourceGenerators` to also collect reference to "tasks", and that its type is `SettingKey[Seq[Task[Seq[File]]]]` is not that important. In other words, I should be able to write:

```scala
sourceGenerators in Compile += buildInfo
```

dwijnand:
Should you also be able to write `crossScalaVersions += scalaVersion`? `crossScalaVersions` collects version strings and `scalaVersion` is a version string.

eed3si9n:
> Should you also be able to write `crossScalaVersions += scalaVersion`? `crossScalaVersions` collects version strings and `scalaVersion` is a version string.

No, I don't think so. We know `crossScalaVersions` is `SettingKey[Seq[String]]` as in a "setting of sequence of String", and `scalaVersion` a "setting of String". So it's perfectly consistent with Getting Started narrative that it requires `.value` in:

```scala
crossScalaVersions += scalaVersion.value
```

The tricky bit is that `sourceGenerators` is `SettingKey[Seq[Task[Seq[File]]]]`, a "setting of sequence of `Task`s". I'm trying to smooth out the awkwardness that `Task` is actually not "task".

dwijnand:
Let's be clear, `sourceGenerators` and `resourceGenerators` are outliers that expose sbt's underlying `Task` type. We should either (a) own this complexity and detail/document it more/better or (b) address the complexity. For instance why can't they, instead of be `SettingKey[Seq[Task[Seq[File]]]]`, be `TaskKey[Seq[File]]`? I don't think we should pick option (c): obscure it, by introducing false `Append` providers and a macro-level implicit conversion.

If it's accepted that you can't `crossScalaVersions += scalaVersion` but you must `crossScalaVersions += scalaVersion.value`, then it's equally accepted that `sourceGenerators in Compile += buildInfo` doesn't make any sense, and that if you want the underlying value (`Seq[File`) then you `buildInfo.value`, and if you want the underlying task then you `buildInfo.taskValue`.

And if adding an inline task by writing `sourceGenerators in Compile += Def.task { .. }.taskValue` is too annoying let's add `Def.taskValue { .. }`.

Or, of course, it's your prerogative to overrule me.

eed3si9n:
> Or, of course, it's your prerogative to overrule me.

I'm fairly sure this magic is the way to go.

For a reference, earlier version of sbt required `seq(...)` function to put `Seq[Setting[_]]` in `build.sbt`. This was worked around by implicit magic in f8d12c5106ffa9c0a1770d47cce2c4471706aadf that unified `Seq[Setting[_]]` and `Setting[_]` into `SettingsDefinition`. We still carry that over today - https://github.com/sbt/util/blob/5306e292900676f69dde458b9a2d7223e3a4c141/internal/util-collection/src/main/scala/sbt/internal/util/Settings.scala#L426-L434

In this case, the implicit conversion from "task" to `Task` can be considered analogous of eta expansion in Scala. This matches user expectation as we see in these comments https://github.com/sbt/website/pull/306#discussion_r94283276:

> Despite writing many `sourceGenerators`, it's not clear what this whole `Def.task{}.taskValue` thingy is about, even after reading this snippet. Why can we not directly `+=` the result of `myGenerator`?

fommil:
why have a macro at all, isn't this ideal for an implicit conversion?

dwijnand:
A few thoughts:

1. Do implicit conversions kick-in post macro-expansion?
2. This "conversion" can only happen within the context of sbt's macros, you can't generally call `Initialize[Task[T]].taskValue`, so I don't think you can define the implicit conversion
3. Having it as a macro-level hack _at least_ scopes it to only this codepath, rather than anywhere.

eed3si9n:
I am doing an implicit conversion. It's just that the RHS requires a macro, so it's stubbed with `???`.

jvican:
One thought: is the failure in CI relevant? I believe it is.

eed3si9n:
iirc I cancelled the Travis job on merge commit because there was some other thing I needed to verify at the time.

PR #2943, opened 2017-01-23
[sbt 0.13] Support sourceGenerators += Def.task { ... }

eed3si9n:
This is a backport of https://github.com/sbt/sbt/pull/2942

### Adds an Append instance that extracts taskValue

This adds a macro-level hack to support += op for sourceGenerators and resourceGenerators using RHS of Initialize[Task[Seq[File]]].
When the types match up, the macro now calls `.taskValue` automatically.

Before:

```scala
sourceGenerators in Compile += buildInfo.taskValue,
sourceGenerators in Compile += (Def.task { Nil }).taskValue
```

After:

```scala
sourceGenerators in Compile += buildInfo,
sourceGenerators in Compile += Def.task { Nil }
```

### DIFF
--- a/main/settings/src/main/scala/sbt/Append.scala
+++ b/main/settings/src/main/scala/sbt/Append.scala
@@ -1,8 +1,9 @@
 package sbt
 
 import java.io.File
-import Def.Classpath
+import Def.{ Classpath, Initialize }
 import scala.annotation.implicitNotFound
+import reflect.internal.annotations.compileTimeOnly
 
 object Append {
   @implicitNotFound(msg = "No implicit for Append.Value[${A}, ${B}] found,\n  so ${B} cannot be appended to ${A}")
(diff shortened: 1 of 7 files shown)


jvican:
Note that this only works for settings.
  
The macro code to implement this looks scary. Are we sure we want to special case this for settings and `+=`? This is IMO a big hack that can confuse people even more since they won't know where they should actually do `.taskValue` or not...
  
@eed3si9n Can you comment on this?

eed3si9n:
What we colloquially call "tasks" are `Def.Initialize[Task[A]]`, which `TaskKey[A]` also extends. Importantly, it's not just `Task[A]`. The build user expects `sourceGenerators` to also collect reference to "tasks", and the fact that its type is `SettingKey[Seq[Task[Seq[File]]]]` is not that important. It's just the matter of matching types.

See discussion in #2942 for more details.

jvican:
I'm well aware of those details.

My comment is that this only works for setting keys (source generators) but it won't work for task keys. We're still forcing users to match the types in that case.

eed3si9n:
Sure. But why would anyone wrap Task in a Task?

PR #2952, opened 2017-02-06
Touch up and add missing notes for 0.13.14

dwijnand:


dwijnand:
@eed3si9n For the jdk 9 changes I opted for the more lean option of keeping the details in the PR description rather than in the release notes. Let me know if you think we should add more details in the notes.

eed3si9n (approved):
LGTM

PR #2956, opened 2017-02-08
[sbt 1.0] Remove sbt.syntax

eed3si9n:
This removes `object sbt.syntax` that used to carry the build.sbt DSL, and folds it back into the sbt package object like sbt 0.13. 
In theory putting DSL in `syntax` may have been a good idea, but `*.scala` file breakage isn't worth it.


main/src/main/scala/sbt/internal/BuildUtil.scala
```diff
@@ -73,7 +73,7 @@ object BuildUtil {
     }
   }
 
-  def baseImports: Seq[String] = "import scala.xml.{TopScope=>$scope}" :: "import sbt._, Keys._, syntax._" :: Nil
+  def baseImports: Seq[String] = "import _root_.scala.xml.{TopScope=>$scope}" :: "import _root_.sbt._" :: "import _root_.sbt.Keys._" :: Nil
```
dwijnand, line 76: Good idea

dwijnand (approved):

Issue #2953, opened 2017-02-06
SBT cannot resolve well transitive dependencies for Birt


DavidPerezIngeniero:
(See the guidelines for contributing, linked above)

## steps

Create a project that has this dependency:

"org.eclipse.birt.runtime" % "org.eclipse.birt.runtime" % "4.6.0-20160607"

## problem

When invoking `update` here is the result:

- org.eclipse.scout.sdk.deps:org.eclipse.core.jobs:[3.8.0,3.8.1) (error: not found)
- org.eclipse.scout.sdk.deps:org.eclipse.core.jobs:[3.8.0,3.8.1) (error: not found)

As, it can be seen [in the Maven repo](https://maven-repository.com/artifact/org.eclipse.birt.runtime/org.eclipse.birt.runtime/4.6.0-20160607)
the real dependencies are:

- "org.eclipse.scout.sdk.deps"	% "org.eclipse.core.runtime" % "3.12.0.v20160606-1342"

among many others.

## expectation

Resolve a dependency that works with Maven, Gradle, ....

## notes

Coursier can resolve everything:

```
coursier fetch org.eclipse.birt.runtime:org.eclipse.birt.runtime:4.6.0-20160607
```

It must be a bug in Ivy.

DavidPerezIngeniero:
This behaviour observed in SBT 0.13.11.
SBT 0.13.13 continues to exhibit this bug.

DavidPerezIngeniero:
dependency graph gathered by SBT, by means of the plugin `"net.virtual-void" % "sbt-dependency-graph" % "0.8.2"`:

```
  +-org.eclipse.birt.runtime:org.eclipse.birt.runtime:4.6.0-20160607
  | +-org.eclipse.birt.runtime.3_7_1:com.lowagie.text:2.1.7
  | +-org.eclipse.birt.runtime.3_7_1:javax.wsdl:1.5.1
  | +-org.eclipse.birt.runtime.3_7_1:org.apache.batik.bridge:1.6.0
  | +-org.eclipse.birt.runtime.3_7_1:org.apache.batik.css:1.6.0
  | +-org.eclipse.birt.runtime.3_7_1:org.apache.batik.dom.svg:1.6.0
  | +-org.eclipse.birt.runtime.3_7_1:org.apache.batik.ext.awt:1.6.0
  | +-org.eclipse.birt.runtime.3_7_1:org.apache.batik.parser:1.6.0
[...]

Issue #2291, opened 2015-11-27
Transitive dependency causes wrong evictions


axos88:
In the example below my build.sbt depends on the angular webjar 1.4.7. My other dependency depends on angular webjar version [1.3.0). However the newest version of said jar is 1.5.0-beta.2

Expected: angular jar version 1.4.7 to be used, as it satisfies both requirement
Actual: angular 1.4.7 evicted, and 1.5.0-beta.2 used instead, which does not satisfy my main build.sbt requirements (version 1.4.7)

```
organization := "vandra.hu"

name := "sbt-bug"

version := "0.0.1"

scalaVersion := "2.11.6"

scalacOptions := Seq("-unchecked", "-deprecation", "-encoding", "utf8", "-feature")

libraryDependencies ++= Seq(
   "org.webjars.bower" % "angular" % "1.4.7"
  ,"org.webjars.bower" % "angular-bootstrap" % "0.14.2"
)
```

pom.xml of angular-bootstrap:

```
(...)
        <dependency>
            <groupId>org.webjars.bower</groupId>
            <artifactId>angular</artifactId>
            <version>[1.3.0,)</version>
        </dependency>
```

Run results:

```
akovanm0:foo avandra$ sbt sbtVersion evicted
[info] Set current project to sbt-bug (in build file:/Users/avandra/projects/foo/)
[info] 0.13.6
[info] Updating {file:/Users/avandra/projects/foo/}foo...
[info] Resolving jline#jline;2.12.1 ...
[info] Done updating.
[warn] There may be incompatibilities among your library dependencies.
[warn] Here are some of the libraries that were evicted:
[warn]  * org.webjars.bower:angular:1.4.7 -> 1.5.0-beta.2 (caller: org.webjars.bower:angular-bootstrap:0.14.2, vandra.hu:sbt-bug_2.11:0.0.1)
[success] Total time: 1 s, completed Nov 27, 2015 12:39:19 PM
```

axos88:
Worked around with force() for now, but this should be fixed

axos88:
Just reproduced in 0.13.9 as well:

```
akovanm0:foo avandra$ sbt sbtVersion evicted
[info] Loading project definition from /Users/avandra/projects/foo/project
[info] Set current project to sbt-bug (in build file:/Users/avandra/projects/foo/)
[info] 0.13.9
[info] Updating {file:/Users/avandra/projects/foo/}foo...
[info] Resolving jline#jline;2.12.1 ...
[...]

PR #2959, opened 2017-02-10
[sbt 0.13] Maven version range improvement

### LINKED ISSUE #2954: Maven version range should translate to a fixed lower bound in Ivy
This is spawn from https://github.com/sbt/sbt/issues/2291#issuecomment-270495227

Upon a dependency conflict (multiple version candidates are found for the same module), by default Ivy uses [latest-wins](http://ant.apache.org/ivy/history/2.3.0/ivyfile/conflicts.html), and Maven uses [nearest-wins](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html) logic to pick the winner. Given that most people use this default, it might make sense to simply use the lower bound instead of trying to translate Maven's [version range](http://maven.apache.org/components/enforcer/enforcer-rules/versionRanges.html) to Ivy's dynamic revision.

### case study

See for example: https://github.com/sbt/sbt/issues/2291

```scala
libraryDependencies ++= Seq(
   "org.webjars.bower" % "angular" % "1.4.7",
  "org.webjars.bower" % "angular-bootstrap" % "0.14.2"
)
```

angular-bootstrap requires:

```xml
        <dependency>
            <groupId>org.webjars.bower</groupId>
            <artifactId>angular</artifactId>
            <version>[1.3.0,)</version>
        </dependency>
```

If we translate Maven's `[1.3.0,)` to Ivy's dynamic revision, Ivy will go out to the Internet and find `1.5.0-beta.2`. This is surprising because the user has already specified `1.4.7` in the graph that satisfies the condition.

If instead as I propose here we translated it to the lower bound `1.3.0`, Ivy will evict it and pick `1.4.7` as expected.

eed3si9n:
Previously, when the dependency resolver (Ivy) encountered a Maven version range such as `[1.3.0,)` it would go out to the Internet to find the latest version. This would result to a surprising behavior where the eventual version keeps changing over time *even when there's a version of the library that satisfies the range condition*.

This changes to some Maven version ranges would be replaced with its lower bound so that when a satisfactory version is found in the dependency graph it will be used. You can disable this behavior using the JVM flag `-Dsbt.modversionrange=false`.

Fixes #2954
Ref #2291 / #2953

### DIFF
--- a/ivy/src/main/scala/sbt/VersionRange.scala
+++ b/ivy/src/main/scala/sbt/VersionRange.scala
@@ -1,0 +1,81 @@
+package sbt
+
+object VersionRange {
+  /** True if the revision is an ivy-range, not a complete revision. */
+  def isVersionRange(revision: String): Boolean = {
+    (revision endsWith "+") ||
+      (revision contains "[") ||
+      (revision contains "]") ||
+      (revision contains "(") ||
+      (revision contains ")")
+  }
+
+  // Assuming Ivy is used to resolve conflict, this removes the version range
+  // when it is open-ended to avoid dependency resolution hitting the Internet to get the latest.
+  // See https://github.com/sbt/sbt/issues/2954
+  def stripMavenVersionRange(version: String): Option[String] =
+    if (isVersionRange(version)) {
+      val noSpace = version.replaceAllLiterally(" ", "")
+      noSpace match {
+        case MavenVersionSetPattern(open1, x1, comma, x2, close1, rest) =>
+          // http://maven.apache.org/components/enforcer/enforcer-rules/versionRanges.html
+          (open1, Option(x1), Option(comma), Option(x2), close1) match {
+            case (_, None, _, Some(x2), "]") => Some(x2)
+            // a good upper bound is unknown
+            case (_, None, _, Some(x2), ")") => None
+            case (_, Some(x1), _, None, _)   => Some(x1)
+            case _                           => None
+          }
+        case _ => None
+      }
+    } else None
+
+  /** Converts Ivy revision ranges to that of Maven POM */
+  def fromIvyToMavenVersion(revision: String): String = {
+    def plusRange(s: String, shift: Int = 0) = {
+      def pow(i: Int): Int = if (i > 0) 10 * pow(i - 1) else 1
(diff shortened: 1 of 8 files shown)


ivy/src/main/scala/sbt/VersionRange.scala
```diff
@@ -0,0 +1,81 @@
...
+            case (_, None, _, Some(x2), ")") => None
+            case (_, Some(x1), _, None, _)   => Some(x1)
+            case _                           => None
+          }
+        case _ => None
+      }
+    } else None
+
+  /** Converts Ivy revision ranges to that of Maven POM */
+  def fromIvyToMavenVersion(revision: String): String = {
```
eed3si9n, line 34: This function is migrated from MakePom.scala so we have version range related functions in the same spot.

dwijnand (changes requested):
See my comment in #2954 (https://github.com/sbt/sbt/issues/2954#issuecomment-281319644). I'm not convinced the bug report is correct or the fix is correct. And I think it would be unwise to rush in such a change: we have no idea if it's correct or the impact it might have. If you _really_ want to get this merged, i.e. you _really_ need this for 0.13.14, I would *strongly* suggest you put this behind an opt-in flag, rather than behind an opt-out flag.

dwijnand (approved):
@eed3si9n is convinced this is the correct change.

We'll release a release candidate or milestone of 0.13.14, and hopefully that'll catch any regressions due to this.

daddykotex:
We just stumbled on this! We had issues resolving internal packages (hosted on a private artifactory).

Our projects extensively use version ranges. that means our `build.sbt` use version like `[1.0.0,)` for almost all of our internal dependencies. This is useful as it makes sure the latest version available for the dependency is resolved from the internal repository and that we don't have to update every single project and change the version. In our case, we want to resolve the latest available (from the internal repository) and not the latest in the graph.

This of course mean that we can't easily reproduce our builds.

The opt-out flag works as intended though.

By the way it was just to let you know, because I don't think you can do anything for us. We probably have an out of the norm approach. 

Thanks for your hard work on `sbt`.

eed3si9n:
> The opt-out flag works as intended though.

Good to know :) Thanks for the note.

PR #2961, opened 2017-02-14
Fix #2551: scripted mode ignores quotation

### LINKED ISSUE #2551: scripted mode ignores quotation
## steps

``` scala
#!/usr/bin/env scalas

println(args.toList)
```

Run this:

```
$ chmod +x args.scala
$ args.scala "foo bar" "baz"
```
## problem

```
List(foo, bar, baz)
```

The quotation is ignored.
## expectation

```
List(foo bar, baz)
```
## notes

See #2549

ekrich:
This fix is for 0.13. another PR will be opened for 1.0.x due to differences in the fix.


eed3si9n:
Forward port is https://github.com/sbt/sbt/pull/2962

PR #2955, opened 2017-02-07
Fix needed due to changes in Zinc

cunei:
sbt/zinc@1f09901 moved ComponentManager; this commit adjusts sbt correspondingly.

This PR should only be merged after the zinc dependency version has been updated.

PR #2969, opened 2017-02-21
Upgrade modules

dwijnand:
Super-seeds #2955.


dwijnand:
Or supersedes if you prefer real English.

jvican:
Superseeded, of course. :smile:

dwijnand:
This is still in a state. The commit needs to be split for its different concerns. And some changes due breaking changes in Zinc need addressing. But I pushed this so I could ask @eed3si9n if he has any thoughts on how to deal with them.

jvican:
Since you're using latest Zinc API, it would be interesting if you chime in and let me know if there's any aspect in the API that bothers you / you find annoying. I'm sure we can polish its rough edges.

dwijnand:
Breaking changes are annoying. Look at the two FIXMEs in the diff.

dwijnand:
Do you consider sbt/util part of "Zinc's API"? Because another one is some APIs now require ManagedLogger instead of Logger. How do I deal with that change, given the Logger I have on hand?

jvican:
Thanks for the pointers, first `FIXME` is severe. I don't have any idea how Zinc logging works, to be honest. I consider Zinc API whatever is defined in the `compiler-interface`. I think that change is related to https://github.com/sbt/zinc/pull/224/files, and it's sbt-only, so I would not consider it part of the Zinc API.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -595,7 +595,7 @@ object Defaults extends BuildCommon {
             rel.internalClassDeps(c).map(intlStamp(_, analysis, s + c)) ++
               rel.externalDeps(c).map(stamp) +
               (apis.internal.get(c) match {
-                case Some(x) => x.compilation.startTime
+                case Some(x) => Long.MinValue // FIXME: x.compilation.startTime
```
jvican: I think you can replace this by `x.compilationTimestamp`.

jvican (commented):
Small comment on fht first FIXME we have discussed about.

dwijnand:
I'm passing the buck on this one, to @eed3si9n.

I originally took this task because I had already dealt with some of the breaking changes in the modules in another branch of mine (unpublished).

But there are some serious breaking changes to deal with here, that I can't deal with because I don't understand what the plan is.


So this can serve as a starting point.

PR #2986, opened 2017-03-02
Fix a bunch but not all compile warnings

dwijnand:


scripted/plugin/src/main/scala/sbt/ScriptedPlugin.scala
```diff
@@ -41,11 +41,8 @@ object ScriptedPlugin extends AutoPlugin {
     scriptedClasspath := getJars(scriptedConf).value,
     scriptedTests := scriptedTestsTask.value,
     scriptedRun := scriptedRunTask.value,
-    scriptedDependencies := {
-      val analysis = (compile in Test).value
-      val pub = (publishLocal).value
-      Unit
-    },
+    scriptedDependencies := (()),
+    scriptedDependencies := scriptedDependencies.dependsOn(compile in Test, publishLocal).value,
```
eed3si9n: I find the one above to be more readable.

dwijnand: I changed it because we're getting compile errors that "analysis" and "pub" are unused. I'll find another way.

Issue #2989, opened 2017-03-03
Split dependency resolution into resolve and downloading (lock dependency)


eed3si9n:
### steps

1. Run `update` task (or `compile` task that calls `update`) on a newly created project.

### expectation

1. `update` internally calls `resolveDependencies`.
2. `resolveDependencies` performs Ivy resolution and persists the resulting dependency graph into a JSON file under `src/sbt/resolveDependencies.json`. This file might contain:
  a. SHA-1 of the input `libraryDependencies`, and SHA-1 of all `libraryDependencies` across the build ("build clock").
  b. URLs for *all artifacts*. Or an expression relative to resolvers.
  c. Timestamp for `-SNAPSHOT` resolution.
3. Using the persisted dependency graph, `update` task downloads *required artifacts*.

### steps 2

1. The user git commits `src/sbt/resolveDependencies.json`.
2. Later day, some other user Bob git clones and runs `compile` task (that internally calls `update`).

### expectation 2

4. `compile` internally calls `update`, which internally calls `resolveDependencies`.
5. `resolveDependencies` finds that there's a persisted dependency graph at `src/sbt/resolveDependencies.json`, and returns it without calling Ivy.
6.  Using the returned dependency graph, `update` task downloads *required artifacts*.

### steps 3

1. Bob adds `commons-io` 1.0 to `libraryDependencies`.
2. Bob reloads the build.
3. Bob runs `compile`.

### expectation 3

7. `compile` internally calls `update`, which internally calls `resolveDependencies`.
8. Detecting the change in `libraryDependencies`, which changes the SHA-1 of all libraryDependencies  ("build clock"), all `resolveDependencies.json` are invalidated, and performs Ivy resolution and overwrites the file.

### note

This should work similar to `Gemfile.lock`.

In addition, by using our own caching mechanism, we should be able to avoid global Ivy locking when Ivy is not invoked. Note that it will need special handling for `-SNAPSHOT`.

Ref https://github.com/sbt/sbt/issues/290

dwijnand:
Please could you describe the issue with steps/problem/expectation?

Does it mean that the global lock file in the ivy cache it will only lock the cache when it's downloading and not when it's just resolving locally?

Or is this some project-local freeze/lock file equivalent? (e.g Gemfile.lock)

eed3si9n:
@dwijnand Updated the description. The latter, but there's some impact to Ivy locking as well.

stephennancekivell:
[...]

PR #2992, opened 2017-03-04
Fix #2982: Add a parallel Ivy resolve engine

### LINKED ISSUE #2982: Parallelize and cache download of Artifacts
In light of Coursier, we should revisit the idea of parallelizing and caching download of Artifacts.

jvican:
Ivy downloads have traditionally been single-threaded.

Parallel downloads are a must for a modern build tool. This commit
builds upon the work done by Josh Suereth in the branch
sbt/ivy-parallel-download-artifact.

To avoid adding external dependencies, it uses the Scala parallel
collections. If maintainers consider that is worth it to use a more
modern and appropriate approach, like `Task`, I'm happy to reimplement
the features with it.

### DIFF
--- a/ivy/src/main/scala/sbt/Ivy.scala
+++ b/ivy/src/main/scala/sbt/Ivy.scala
@@ -4,31 +4,30 @@
 package sbt
 
 import Resolver.PluginPattern
-import ivyint.{ CachedResolutionResolveEngine, CachedResolutionResolveCache, SbtDefaultDependencyDescriptor }
-
+import ivyint.{ CachedResolutionResolveCache, CachedResolutionResolveEngine, ParallelResolveEngine, SbtDefaultDependencyDescriptor }
 import java.io.File
 import java.net.URI
 import java.text.ParseException
 import java.util.concurrent.Callable
-import java.util.{ Collection, Collections => CS, Date }
+import java.util.{ Collection, Date, Collections => CS }
 import CS.singleton
 
 import org.apache.ivy.Ivy
 import org.apache.ivy.core.report.ResolveReport
-import org.apache.ivy.core.{ IvyPatternHelper, LogOptions, IvyContext }
-import org.apache.ivy.core.cache.{ ResolutionCacheManager, CacheMetadataOptions, DefaultRepositoryCacheManager, ModuleDescriptorWriter }
+import org.apache.ivy.core.{ IvyContext, IvyPatternHelper, LogOptions }
+import org.apache.ivy.core.cache.{ CacheMetadataOptions, DefaultRepositoryCacheManager, ModuleDescriptorWriter, ResolutionCacheManager }
 import org.apache.ivy.core.event.EventManager
-import org.apache.ivy.core.module.descriptor.{ Artifact => IArtifact, DefaultArtifact, DefaultDependencyArtifactDescriptor, MDArtifact }
-import org.apache.ivy.core.module.descriptor.{ DefaultDependencyDescriptor, DefaultModuleDescriptor, DependencyDescriptor, ModuleDescriptor, License }
+import org.apache.ivy.core.module.descriptor.{ DefaultArtifact, DefaultDependencyArtifactDescriptor, MDArtifact, Artifact => IArtifact }
+import org.apache.ivy.core.module.descriptor.{ DefaultDependencyDescriptor, DefaultModuleDescriptor, DependencyDescriptor, License, ModuleDescriptor }
 import org.apache.ivy.core.module.descriptor.OverrideDependencyDescriptorMediator
 import org.apache.ivy.core.module.id.{ ArtifactId, ModuleId, ModuleRevisionId }
 import org.apache.ivy.core.resolve._
 import org.apache.ivy.core.settings.IvySettings
 import org.apache.ivy.core.sort.SortEngine
-import org.apache.ivy.plugins.latest.{ LatestStrategy, LatestRevisionStrategy, ArtifactInfo }
+import org.apache.ivy.plugins.latest.{ ArtifactInfo, LatestRevisionStrategy, LatestStrategy }
 import org.apache.ivy.plugins.matcher.PatternMatcher
-import org.apache.ivy.plugins.parser.m2.{ PomModuleDescriptorParser }
-import org.apache.ivy.plugins.resolver.{ ChainResolver, DependencyResolver, BasicResolver }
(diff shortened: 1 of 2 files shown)


main/src/main/scala/sbt/EvaluateConfigurations.scala
```diff
@@ -174,6 +174,7 @@ object EvaluateConfigurations {
     // TODO - Should we try to namespace these between.sbt files?  IF they hash to the same value, they may actually be
     // exactly the same setting, so perhaps we don't care?
     val result = try {
+      println(s"EXPRESSION $expression")
```
jvican: Oops, this has to go. Will remove soon.

eed3si9n:
Is this the resurrection of Josh's build? If so we should credit him for the work.

jvican:
@eed3si9n I've already done so. Citing from the description of my PR:
> This commit builds upon the work done by Josh Suereth in the branch sbt/ivy-parallel-download-artifact.

eed3si9n:
Why don't we rebase his original commit, and append changes on top of it?
I was also working on parallel download on one of the plane rides, and I used git cherry picking.

jvican:
> Why don't we rebase his original commit, and append changes on top of it?
  
Sure thing, I can do it. Will push in a bit.

jvican:
By the way, I'm a hard user of coursier, so if the sbt maintainers consider coursier to be a better alternative to this PR, let me know -- I can invest some time to create a PR to use coursier for dependency resolution. Otherwise, I think this PR does a good job at parallelizing the current ivy resolution.

fommil:
Is the build failure real? Anything that speeds up meta resolution is very very welcome!

eed3si9n:
I think a more conservative approach would be to first implement Gemfile lock (https://github.com/sbt/sbt/issues/2989), and then implement custom caching / parallelizing on top of the URL list (https://github.com/sbt/sbt/issues/2982) instead of re-implementing the entire dependency resolution stack.

dwijnand:
To a brief glance it looks like the codacy issue is easily solvable.

jvican:
> I think a more conservative approach would be to first implement Gemfile lock 
  
I'm already working on that. Thanks for suggesting.
  
> then implement custom caching / parallelizing on top of the URL list (#2982) instead of re-implementing the entire dependency resolution stack.
  
I disagree here. I don't see why these two depend on each other and I think that this should be merged independently of the other two features you mention. The speedup is real, you can try it out by publishing locally and removing your ivy cache.
  
Also, this commit does not reimplement the entire dependency resolution stack. It just wraps the existing resolve engine and make it parallel.

OlivierBlanvillain:
@jvican Do you have measurements / estimates of how much speed up you get this these changes? How does this perform compared to Coursier ?

jvican:
I'll give you concrete results tomorrow @OlivierBlanvillain, but I got very good results locally.

olafurpg:
@OlivierBlanvillain I ran a small experiment to compare 

- sbt-coursier
- sbt 0.13.13
- sbt 0.13.14-SNAPSHOT, published locally from this PR

The challenge was to resolve and download one module [`"org.apache.spark" % "spark-sql_2.11" % "2.1.0"`](http://search.maven.org/#artifactdetails%7Corg.apache.spark%7Cspark-sql_2.11%7C2.1.0%7Cjar), which has 134 transitive dependencies.

I created three empty sbt project with the following `build.sbt`:

```
scalaVersion in ThisBuild := "2.11.8"
libraryDependencies += "org.apache.spark" %% "spark-sql" % "2.1.0"
```

I started sbt in all three project with the same command:

```sh
export COURSIER_CACHE=$(pwd)/.coursier # for coursier project
java -jar sbt-launch.jar -Dsbt.ivy.home=$(pwd)/.ivy2
```

Once the sbt shell had loaded, I ran the command `update` and recorded the total time. Without further ado, here are the results

```
coursier: 13s
0.13.13: 51s
0.13.14-SNAPSHOT: 36s
```

I noticed that resolution was taking the longest time in the plain sbt resolver projects. I re-ran the experiment for 0.13.13 on a new clean project to record how many seconds it took before downloading started. It took 32 seconds to resolve. This indicates that parallel downloads (from this PR) took roughly 4s while synchronous downloads on 0.13.13 took 19s, ~4-5x difference.

OlivierBlanvillain:
Amazing!

I expect that people do no lose they entire cache very often, maybe a "fair" comparasion with coursier would be to first warm up the build tool cache in all three option with an empty compile (without the `libraryDependencies`), to then only mesure the time it takes to actually download spark related stuff.

alexarchambault:
How is the IO parallelized by this PR?

In coursier, the IO is made on a fixed thread pool with 6 threads by default, hence up to 6 parallel downloads. I guess the same number of parallel downloads should be allowed for Ivy here.

I haven't tried to tweak / optimize the number of parallel downloads in coursier. 6 is usually the max number of parallel downloads in the usual browsers afaik or last time I checked (chrome, firefox, ...)

eed3si9n:
For sbt 1.0 librarydependency, my plan was to use Gigahorse or something on the backend so we can download in parallel and keep the connections alive.

fommil:
@eed3si9n you said you didn't want to depend on `netty`... so if you bring that in, can't you provide a websockets sbt-server API? And besides, doesn't *shading* make your argument about binary compatibility moot?

eed3si9n:
@fommil You're right on Netty / shading. Let's create another issue to discuss that.

jvican:
@olafurpg Thank you for those results, pretty awesome!
  
@alexarchambault I don't know the details, it relies on parallel collections which are not the most efficient for IO, but as we see it brings us a good 9x speedup. Coursier beats us in resolutions, great tool :wink:.
  
I'm updating this soon.

japgolly:
The coursier result looks fantastic. May I ask then, why this PR instead of switching to coursier?

jvican:
@japgolly I've just opened https://github.com/sbt/sbt/issues/2997. There I explain why I think this should not be closed :smile:.

japgolly:
@jvican I just had a read and I understand now. Thank you for the info!

jvican:
I'll be updating this PR soon with @eed3si9n's remarks.

jvican:
I'm closing this and porting to sbt 1.0. I will not be using Eugene's advice anymore because the required changes for this have to go into `librarymanagement` and the whole dir structure has changed.

PR #2996, opened 2017-03-07
Notify & enable users to stay in the warm shell

### LINKED ISSUE #2987: Display warning when invoked in batch mode
## steps

Someone runs `$ sbt compile` from the terminal.

## expectation

Display 

```
Executing "compile" in batch mode. Note running "sbt" and executing "compile" from the shell repeatedly is going to compile much faster. (Think 2x)
```

## notes

It's been pointed out many times that batch mode is 1.3x ~ 2x slower because of Scalac JIT (and from sbt 0.13.13 classloader caching)
https://twitter.com/eed3si9n/status/837585213563801600

dwijnand:
Notify & enable users to stay in sbt's shell on the warm JVM by hitting
[ENTER] while sbt is running.

Looks like this; first I run 'sbt about', then I hit [ENTER]:

    $ sbt about
    [info] !!! Executing in batch mode !!! For better performance, hit [ENTER] to remain in the sbt shell

    [info] Loading global plugins from /Users/dnw/.dotfiles/.sbt/0.13/plugins
    [info] Loading project definition from /s/t/project
    [info] Set current project to t (in build file:/s/t/)
    [info] This is sbt 0.13.14-SNAPSHOT
    [info] The current project is {file:/s/t/}t 0.1.0-SNAPSHOT
    [info] The current project is built against Scala 2.12.1
    [info] Available Plugins: sbt.plugins.IvyPlugin, sbt.plugins.JvmPlugin, sbt.plugins.CorePlugin, sbt.plugins.JUnitXmlReportPlugin, sbt.plugins.Giter8TemplatePlugin
    [info] sbt, sbt plugins, and build definitions are using Scala 2.10.6
    >
    >

Fixes #2987 

image

### DIFF
--- a/main/command/src/main/scala/sbt/MainLoop.scala
+++ b/main/command/src/main/scala/sbt/MainLoop.scala
@@ -67,8 +67,10 @@ object MainLoop {
       val loggedState = state.copy(globalLogging = newLogging)
       def isInteractive = System.console() != null
       def hasShell = state.remainingCommands contains "shell"
-      if (isInteractive && !hasShell)
-        state.log warn "Executing in batch mode; for better performance, hit [ENTER] to remain in the sbt shell"
+      if (isInteractive && !hasShell) {
+        state.log warn "BATCH MODE: for better performance hit [ENTER] to switch to interactive mode"
+        state.log warn "  consider launching sbt without any commands, or explicitly passing 'shell'"
+      }
       try run(loggedState) finally out.close()
     }
 
--- a/main/command/src/main/scala/sbt/MainLoop.scala
+++ b/main/command/src/main/scala/sbt/MainLoop.scala
@@ -65,6 +65,10 @@ object MainLoop {
       val newLogging = state.globalLogging.newLogger(out, logBacking)
       transferLevels(state, newLogging)
       val loggedState = state.copy(globalLogging = newLogging)
+      def isInteractive = System.console() != null
+      def hasShell = state.remainingCommands contains "shell"
+      if (isInteractive && !hasShell)
+        state.log info "!!! Executing in batch mode !!! For better performance, hit [ENTER] to remain in the sbt shell"
       try run(loggedState) finally out.close()
     }
 
--- a/main/command/src/main/scala/sbt/MainLoop.scala
+++ b/main/command/src/main/scala/sbt/MainLoop.scala
@@ -68,7 +68,7 @@ object MainLoop {
       def isInteractive = System.console() != null
       def hasShell = state.remainingCommands contains "shell"
       if (isInteractive && !hasShell)
-        state.log info "!!! Executing in batch mode !!! For better performance, hit [ENTER] to remain in the sbt shell"
+        state.log warn "Executing in batch mode; for better performance, hit [ENTER] to remain in the sbt shell"
       try run(loggedState) finally out.close()
     }
 
--- a/main/command/src/main/scala/sbt/MainLoop.scala
+++ b/main/command/src/main/scala/sbt/MainLoop.scala
@@ -68,7 +68,8 @@ object MainLoop {
       def isInteractive = System.console() != null
       def hasShell = state.remainingCommands contains "shell"
       if (isInteractive && !hasShell) {
-        state.log warn "BATCH MODE: for better performance hit [ENTER] to switch to interactive mode"
+        state.log warn "Executing in batch mode."
+        state.log warn "  For better performance, hit [ENTER] to switch to interactive mode, or"
         state.log warn "  consider launching sbt without any commands, or explicitly passing 'shell'"
       }
       try run(loggedState) finally out.close()
--- a/main/command/src/main/scala/sbt/State.scala
+++ b/main/command/src/main/scala/sbt/State.scala
@@ -176,13 +176,18 @@ object State {
 
   /** Provides operations and transformations on State. */
   implicit def stateOps(s: State): StateOps = new StateOps {
-    def process(f: (String, State) => State): State =
-      s.remainingCommands match {
-        case Seq() => exit(true)
-        case Seq(x, xs @ _*) =>
-          log.debug(s"> $x")
-          f(x, s.copy(remainingCommands = xs, history = x :: s.history))
+    def process(f: (String, State) => State): State = {
+      def doX(x: String, xs: Seq[String]) = {
+        log.debug(s"> $x")
+        f(x, s.copy(remainingCommands = xs, history = x :: s.history))
       }
+      def isInteractive = System.console() != null
+      def hasInput = System.console().reader().ready()
+      s.remainingCommands match {
+        case Seq()           => if (isInteractive && hasInput) doX("shell", Nil) else exit(true)
+        case Seq(x, xs @ _*) => doX(x, xs)
+      }
+    }
     def :::(newCommands: Seq[String]): State = s.copy(remainingCommands = newCommands ++ s.remainingCommands)
     def ::(command: String): State = (command :: Nil) ::: this
     def ++(newCommands: Seq[Command]): State = s.copy(definedCommands = (s.definedCommands ++ newCommands).distinct)
--- a/notes/0.13.14/stay-in-shell.md
+++ b/notes/0.13.14/stay-in-shell.md
@@ -1,0 +1,7 @@
+### Improvements
+
+- Notifies & enables users to stay in sbt's shell on the warm JVM by hitting \[ENTER\] while sbt is running. [#2987][]/[#2996][] by [@dwijnand][]
+
+[#2987]: https://github.com/sbt/sbt/issues/2987
+[#2996]: https://github.com/sbt/sbt/pull/2996
+[@dwijnand]: https://github.com/dwijnand


fommil:
the concept is really great, I'm tired of telling people to stay in shell mode. The hint here suggests that hitting return is the only way to get into script mode, can you think of wording that captures how to start in script mode?

"BATCH MODE: for better performance use the sbt shell by not passing commands, or by pressing [ENTER] to upgrade."

main/command/src/main/scala/sbt/MainLoop.scala
```diff
@@ -65,6 +65,10 @@ object MainLoop {
       val newLogging = state.globalLogging.newLogger(out, logBacking)
       transferLevels(state, newLogging)
       val loggedState = state.copy(globalLogging = newLogging)
+      def isInteractive = System.console() != null
+      def hasShell = state.remainingCommands contains "shell"
+      if (isInteractive && !hasShell)
+        state.log info "!!! Executing in batch mode !!! For better performance, hit [ENTER] to remain in the sbt shell"
```
larsrh: _(bikeshedding)_ Should this not be a warning instead? I believe this would make it more likely for people to pay attention instead of the triple exclamation marks.

dwijnand: Much better. Thanks.

dwijnand:
@fommil I agree, having a little trouble thinking of a good, not-too-verbose way to communicate that.

There so many parts to the message:
* the current status "executing in batch mode"
* the reason for the message "for better performance"
* and the 3 different options: pass no commands, pass 'shell', or hit [ENTER]

WDYT?

```
[warn] BATCH MODE: for better performance hit [ENTER] to switch to interactive mode
```

dwijnand:
image

eed3si9n (changes requested):
I think this is a run-on is a sentence. (bike shedding)

How about:

```
Executing command(s) 'about' in batch mode.
For faster compilation, hit [ENTER] to switch to interactive mode, or
consider launching sbt without any commands.
```

dwijnand:
@eed3si9n 

* I avoided printing out the commands as I CBA to deal with quoting, etc; and I don't think it's required
* I stuck to "for better performance" because this is not just specific to compilation
* I kept the "or explicitly pass 'shell'" part, as it's the 3rd option some might be interested in

dwijnand:
image

main/command/src/main/scala/sbt/State.scala
```diff
@@ -176,13 +176,18 @@ object State {
...
   /** Provides operations and transformations on State. */
   implicit def stateOps(s: State): StateOps = new StateOps {
-    def process(f: (String, State) => State): State =
+    def process(f: (String, State) => State): State = {
+      def doX(x: String, xs: Seq[String]) = {
+        log.debug(s"> $x")
+        f(x, s.copy(remainingCommands = xs, history = x :: s.history))
+      }
+      def isInteractive = System.console() != null
+      def hasInput = System.console().reader().ready()
```
eed3si9n, line 185: What happens if someone runs `sbt run` that asks user for text input?

eed3si9n: Or more likely `sbt +publishSigned`

main/command/src/main/scala/sbt/MainLoop.scala
eed3si9n: I'm not so sure if warning is the right level here. sbt is able to reduce the log level using `--warn` (This could get rid of Resolving etc.), and I feel like "tip of the day" message sits squarely on the "info" level. Yes, your build is going to compile slower, but it's not like it's destroying your code.

main/command/src/main/scala/sbt/State.scala
dwijnand: **src/main/scala/t/T.scala**

```scala
package t

object T {
    val con = System.console()
    val x = con readLine "Give me something: "
    con printf ("You gave me: %s\n", x)
  }
}
```

->

```
$ sbt run

[info] Loading global plugins from /Users/dnw/.dotfiles/.sbt/0.13/plugins
[info] Loading project definition from /s/t/project
[info] Set current project to t (in build file:/s/t/)
[info] Running t.T
Give me something: You gave me:
[success] Total time: 0 s, completed 07-Mar-2017 14:24:09
```

---

**project/plugins.sbt**

```scala
addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.0.0")
```

->

```
$ sbt +publishLocalSigned

[info] Loading global plugins from /Users/dnw/.dotfiles/.sbt/0.13/plugins
[info] Loading project definition from /s/t/project
[info] Updating {file:/s/t/project/}t-build...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
[info] Set current project to t (in build file:/s/t/)
[info] Setting version to 2.12.1
[info] Reapplying settings...
[info] Set current project to t (in build file:/s/t/)
[info] Packaging /s/t/target/scala-2.12/t_2.12-0.1.0-SNAPSHOT-sources.jar ...
[info] Done packaging.
[info] Main Scala API documentation to /s/t/target/scala-2.12/api...
[info] Wrote /s/t/target/scala-2.12/t_2.12-0.1.0-SNAPSHOT.pom
[info] :: delivering :: com.dwijnand#t_2.12;0.1.0-SNAPSHOT :: 0.1.0-SNAPSHOT :: integration :: Tue Mar 07 14:28:17 GMT 2017
[info] 	delivering ivy file to /s/t/target/scala-2.12/ivy-0.1.0-SNAPSHOT.xml
[info] Packaging /s/t/target/scala-2.12/t_2.12-0.1.0-SNAPSHOT.jar ...
[info] Done packaging.
model contains 3 documentable templates
[info] Packaging /s/t/target/scala-2.12/t_2.12-0.1.0-SNAPSHOT-javadoc.jar ...
[info] Done packaging.
Please enter PGP passphrase (or ENTER to abort):
java.lang.RuntimeException: Empty passphrase. aborting...
	at scala.sys.package$.error(package.scala:27)
	at com.typesafe.sbt.pgp.SbtPgpCommandContext.inputPassphrase(SbtPgpCommandContext.scala:27)
	at com.typesafe.sbt.pgp.SbtPgpCommandContext$$anonfun$1$$anonfun$apply$1$$anonfun$apply$2.apply(SbtPgpCommandContext.scala:34)
	at com.typesafe.sbt.pgp.SbtPgpCommandContext$$anonfun$1$$anonfun$apply$1$$anonfun$apply$2.apply(SbtPgpCommandContext.scala:34)
	at scala.Option.getOrElse(Option.scala:120)
	at com.typesafe.sbt.pgp.SbtPgpCommandContext$$anonfun$1$$anonfun$apply$1.apply(SbtPgpCommandContext.scala:34)
	at com.typesafe.sbt.pgp.SbtPgpCommandContext$$anonfun$1$$anonfun$apply$1.apply(SbtPgpCommandContext.scala:34)
	at com.typesafe.sbt.pgp.Cache$$anonfun$withValue$1.apply(SbtHelpers.scala:39)
	at scala.Option.getOrElse(Option.scala:120)
	at com.typesafe.sbt.pgp.Cache$class.withValue(SbtHelpers.scala:38)
	at com.typesafe.sbt.pgp.PasswordCache$.withValue(SbtHelpers.scala:55)
	at com.typesafe.sbt.pgp.SbtPgpCommandContext$$anonfun$1.apply(SbtPgpCommandContext.scala:34)
	at com.typesafe.sbt.pgp.SbtPgpCommandContext.retry(SbtPgpCommandContext.scala:42)
	at com.typesafe.sbt.pgp.SbtPgpCommandContext.withPassphrase(SbtPgpCommandContext.scala:31)
	at com.typesafe.sbt.pgp.BouncyCastlePgpSigner.sign(PgpSigner.scala:37)
	at com.typesafe.sbt.pgp.PgpSettings$$anonfun$signingSettings$1$$anonfun$apply$1.apply(PgpSettings.scala:120)
	at com.typesafe.sbt.pgp.PgpSettings$$anonfun$signingSettings$1$$anonfun$apply$1.apply(PgpSettings.scala:117)
	at scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:251)
	at scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:251)
	at scala.collection.immutable.Map$Map4.foreach(Map.scala:181)
	at scala.collection.TraversableLike$class.flatMap(TraversableLike.scala:251)
	at scala.collection.AbstractTraversable.flatMap(Traversable.scala:105)
	at com.typesafe.sbt.pgp.PgpSettings$$anonfun$signingSettings$1.apply(PgpSettings.scala:117)
	at com.typesafe.sbt.pgp.PgpSettings$$anonfun$signingSettings$1.apply(PgpSettings.scala:115)
	at scala.Function4$$anonfun$tupled$1.apply(Function4.scala:35)
	at scala.Function4$$anonfun$tupled$1.apply(Function4.scala:34)
	at sbt.$tilde$greater$$anonfun$$u2219$1.apply(TypeFunctions.scala:40)
	at sbt.std.Transform$$anon$4.work(System.scala:63)
	at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:228)
	at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:228)
	at sbt.ErrorHandling$.wideConvert(ErrorHandling.scala:17)
	at sbt.Execute.work(Execute.scala:237)
	at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:228)
	at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:228)
	at sbt.ConcurrentRestrictions$$anon$4$$anonfun$1.apply(ConcurrentRestrictions.scala:159)
	at sbt.CompletionService$$anon$2.call(CompletionService.scala:28)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[error] (*:signedArtifacts) Empty passphrase. aborting...
[error] Total time: 3 s, completed 07-Mar-2017 14:28:20
```

main/command/src/main/scala/sbt/MainLoop.scala
dwijnand: I found that without my "!!!" even **I** who implemented the feature couldn't see it.

That's why I agreed with Lars that the issue was in the choice of using info. Move it to warn allowed me to drop the !!! workaround, make the message shorter, and also (as you can see from the screenshots below) benefit from `[warn]` being in yellow.

We have only 1 axis (error/warn/info/debug) on which to attribute semantics. I don't know you, but if I IRL see someone run `sbt compile` I would want to warn them.

> Yes, your build is going to compile slower, but it's not like it's destroying your code.

It's not an error, just a warning.

eed3si9n (approved):

dwijnand:
Testing the behaviour of `System.console()` on Travis CI:

```scala
package t

object T {
    def isInteractive = System.console() != null
    if (isInteractive)
      println("interactive: TRUE")
    else
      println("interactive: FALSE")
  }
}
```

```yml
language: scala
jdk: oraclejdk8
scala: 2.12.2
script:
  - sbt package
  - scala target/scala-*/*.jar
  - scala target/scala-*/*.jar < /dev/null
  - scala target/scala-*/*.jar | cat
  - scala target/scala-*/*.jar < /dev/null | cat
```

https://travis-ci.org/dwijnand/l2travis/builds/208656785

I'm expecting the output:

```
interactive: FALSE
interactive: FALSE
interactive: FALSE
interactive: FALSE
```

dwijnand:
No `scala` binary on Travis CI. Switched to sbt-assembly and java: https://travis-ci.org/dwijnand/l2travis/builds/208672634

dwijnand:
```
$ java -jar target/scala-*/*-assembly-*.jar
interactive: TRUE
```

So when Travis CI invokes things it's technically in "interactive" mode. So unless the invocation does something like `< /dev/null`/`| cat` (or use `-batch` which is a sbt-extras' feature), then they'll see these extra 3 lines of warning..

I think we should merge, and if anyone wants to make the out-the-box experience with Travis CI or similar better, they're very welcome to contribute an enhancement.

ScalaWilliam:
Wouldn't it be enough to just indicate this warning, and NOT do the Enter thing?

How can I disable this feature if I don't want to see the warning every time I'm intentionally in batch mode?

When I see a warning my alarm bells go off to get ready to investigate a problem but this non-problem is going to be appearing all the time.

dwijnand:
> Wouldn't it be enough to just indicate this warning, and NOT do the Enter thing?

You could. The Enter thing was added for convenience.

> How can I disable this feature if I don't want to see the warning every time I'm intentionally in batch mode?
>
> When I see a warning my alarm bells go off to get ready to investigate a problem but this non-problem is going to be appearing all the time.

Using sbt-extras if you can pass the `-batch` flag you indicate to sbt that you're not an interactive shell. Then it won't tell you how to switch to interactive mode.


I wonder, do you think the length of the message, i.e the fact that it's 3 lines long, part of the reason your alarm bells are going off?

eed3si9n:
Note that currently the warning is displayed incorrectly, which will be fixed soon - https://github.com/sbt/sbt/issues/3004

> When I see a warning my alarm bells go off to get ready to investigate a problem but this non-problem is going to be appearing all the time.

Are we talking about literal or figurative/mental alarm bells here?
The purpose of this is to let the user community know that using batch mode is slower, so to some degree if it's catching your attention, it's doing its job. Another idea might be to display a large ASCII art at "info" level when you run the batch mode.

btw, what would be the reason for using batch mode intentionally besides CI?

ScalaWilliam:
I find seeing this message all the time a bit redundant, would like to disable it permanently without specifying extra flags. I understood the message the first time I saw it, why keep on repeating it. So suggestion would be to have an SBT option in `~/.sbt` that helps disable this thing.

```
```

The 'warn' part captures attention very quickly, especially when it's highlighted very visibly in the terminal.

---

I just saw @eed3si9n's comment that this is actually a bug so I guess my comment above is not relevant any more.

--- 

I do use batch mode typically to execute multiple commands locally to do test & build sequences - before going to CI.

It's more convenient to just type `sbt clean stage docker:publishLocal && doSomethingElse` in one line.

We also call into SBT from Makefiles, I'm not sure how this will affect things (will it think interactive mode & also 'warn' users when they try to build via make?). Some projects use sbt as one of several build tools.

Issue #2944, opened 2017-01-26
clean up source code


tpolecat:
So, at least for me the state of the source code is a big barrier to contributing. It's poorly formatted, sparsely commented, and in places it's organized strangely. So I suggest a cleanup task to be done at some point post-1.0 when we're free to make "big" changes. In this case the changes are trivial but global and will result in a huge PR. Anyway:

- Run everything through scalafmt to make formatting consistent and introduce some whitespace between definitions.
- Add Scaladoc comments explaining at least what every class does. Ideally every public member would be documented but that's too much to ask initially.
- In some places there are a bunch of definitions in the same source file that might work better split up into multiple source files.

After this is done I think it will be easier for me (and hopefully others) to get some momentum in trying to understand how the hell sbt actually works so we can start making meaningful improvements.

eed3si9n:
I agree that sbt's code base is very difficult to tackle.

Not sure what goes to pre- and post-1.0, but my approach to solving this is to split sbt into distinct modules and clarify public API vs internal implementation (rather than trying to make everyone understand every class). I'd say nearly half of sbt 0.13 code is around incremental compilation and supporting classes. But for most contributors, incremental compiler can be a simple function that takes your source files and generates `*.class` files, and don't have know about the details.

Similarly, library management is thin wrapper around Ivy, and something people can totally get into and improve, but we should also be able to abstract it out as a simple function that takes sequence of ModuleIDs and generates `File`s.

PR #3003, opened 2017-03-11
Reformat with scalafmt

dwijnand:
Refs #2944


.scalafmt.conf
```diff
@@ -0,0 +1,16 @@
+style = defaultWithAlign
```
eed3si9n: I'm against this since it messes with the Git history.

.scalafmt.conf
```diff
@@ -0,0 +1,16 @@
...
+maxColumn = 112
+assumeStandardLibraryStripMargin = true
+
+align.openParenCallSite = false
+danglingParentheses = true
+
+docstrings = JavaDoc
```
eed3si9n: Why are we using JavaDoc style?

eed3si9n (changes requested):

.scalafmt.conf
cvogt: Do you think it makes a difference since most code is being touched anyways in such a code format PR? You guys better have no other open pull requests or branches you still plan to merge :).

Git blame has a mode to ignore whitespace btw.

dwijnand: Indeed, ignoring whitespace is supported not only in git blame, but in (afaik) all git functions. And it's supported by GitHub's UI.

I feel _strongly_ that vertically aligned code is _much_ more legible. So I'm against this decision.

But I've dropped it and updated the PR nonetheless.

.travis.yml
```diff
@@ -39,14 +29,22 @@ env:
...
     - SBT_CMD="mavenResolverPluginTest:scripted dependency-management/*1of2 project/transitive-plugins"
 
-notifications:
-  email:
-    - eugene.yokota@typesafe.com
-    - joshua.suereth@typesafe.com
+matrix:
+  fast_finish: true
+  include:
+    - env: SBT_CMD="scalafmtDiffTest"
```
jvican, line 35: Why do we use the sbt plugin while there's a bootstrap script that runs really fast in comparison with it? The sbt plugin doesn't buy us anything, it's just an alias to the command line tool.
 
I would propose to put the bootstrap script in the root project directory instead.

.scalafmt.conf
```diff
@@ -0,0 +1,20 @@
...
+align.tokens.add = [
+  { code = "%",   owner = "Infix" }
+  { code = "%%",  owner = "Infix" }
+  { code = "%%%", owner = "Infix" }
+  { code = "=>",  owner = "Case"  }
+]
+
+danglingParentheses = true
+
+rewrite.rules = [RedundantBraces, RedundantParens, SortImports]
```
jvican, line 14: Why do we need these rewrites? Why `RedundantBraces`? I don't want scalafmt to remove the curly braces in a very long if-else expression just because it is a simple expression. Scalafmt is a formatter, not a code rewriter...
  
IMO, any body that is more than one line long (two tops) has to be enclosed in curly braces. They help readability. What this rewrite is saying is that if I have an if else of more than 10 lines of code, scalafmt will remove the parens. I don't think we should do this.

.scalafmt.conf
```diff
@@ -0,0 +1,20 @@
+maxColumn = 112
```
jvican, line 1: I'm against this change.
  
Aside from the fact that I think this configuration file is too big, and we should just go for the default, which is a sentiment I've expressed several times in the Gitter channel, I think that the max column is the worst that can happen to these formatting options.
  
I suggest that we go for the gold maxcolumn standard, 80.  Having it larger than that hinders readability in other devices like mobile phones or laptops. People *do* read codebases in these devices, and even write code in them, and we should take them into account.
  
One of the main reasons why I proposed formatting the sbt sources is because reading horizontal code and scrolling left and right is less than an ideal experience. If we're pushing the limits of the max column option, why reformat sbt at all?
  
Plus, this option doesn't make sense if there's no vertical alignment.

jvican (commented):
I've just left several comments on the formatting options as a potential external contributor to the project.

.travis.yml
jvican: See https://olafurpg.github.io/scalafmt/#Coursier for more information on the script. The current approach has the disadvantage that we pay the sbt startup time to format, while the CLI tool uses nailgun to reduce startup time of the JVM.

.scalafmt.conf
jvican: @eed3si9n Eugene, is there any chance we could have vertical alignment given the fact that both git and GitHub support ignoring whitespace?
  
Perhaps we could say in the CONTRIBUTING guide that this option should be enabled locally to browse through the git history? I think this is a good compromise.

jvican: FTR, I *personally* don't like this option, but I see value in compromising on the formatting options. After all, we all have our opinions.
  
This is how you do it in Intellij http://stackoverflow.com/questions/19935317/ignore-whitespace-changes-in-intellij-changebars (see the second answer). With git, you do it with `git diff -w`.

.scalafmt.conf
eed3si9n: > I suggest that we go for the gold maxcolumn standard, 80. Having it larger than that hinders readability in other devices like mobile phones or laptops.

80 column comes from IBM punch cards - https://en.wikipedia.org/wiki/Punched_card
I often code from my laptop, but never thought "hmmm, I wish the code base had 80 column max." Changing our code base so someone can read our code in mobile phones make no sense. Is the advent of the mobile phones the reason why we are switching to Scalafmt?

Why not 132 columns (number of characters printable by IBM 1403), or 100 columns ([Google's Java Guide](https://google.github.io/styleguide/javaguide.html#s4.4-column-limit))?

jvican: The origin of this standard is not relevant for this discussion.
  
> Changing our code base so someone can read our code in mobile phones make no sense.
  
You missed the laptop part. Why are you changing the codebase at all then? Is this a real attempt to optimize for the contributors of sbt? Lots of people read code in their mobile phones. You don't think this is true? Have a look at the [Scalajs style guide](https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md), particularly at the line length section. Alike them, I *do review* code in my mobile phone. I'm sure this is true for a big chunk of developers because we spend considerable amounts of time commuting.
  
> Why not 132 columns (number of characters printable by IBM 1403) or 100 columns (Google's Java Guide)?
  
Google sticks to 80 in almost all the other programming languages they use. See, for instance, [Google's Python style guide](https://google.github.io/styleguide/pyguide.html#Line_length) or the C++ Coding Style guide. The reason to go for 100 for Java was probably because Java is a verbose language.
  
These are the strongest reasons why we should stick to 80 columns:
  
* Readable with a larger font on laptops. Sometimes it's a matter of taste, some other times it's an imperative to use larger fonts, especifically for people with eye problems (like me).
* Limits the complexity in one line. Java is a *verbose* programming language. Scala, however, is a complex and concise language that allows the programmer to increase the complexity of the logic in one line since boilerplate is almost minimal. Having shorter lines reduces the complexity of method bodies significantly and allows the programmer to think more about the code he/she's writing. Having a very long line is usually a code smell that hints that the code should be refactored.
  
Some other arguments in favor:
* Leaves space for putting two versions side by side for comparison. Lots of times, we diff and blame stuff and having long lines will make it an uncomfortable experience.
* Leaves space for navigation views in the IDE and other windows in the same screen.
* Limits indentation, which in turn limits complexity of methods/functions.
  
Another data point: designers use a column size between 50 and 80 in websites, more than that is considered bad for readability. There are usability studies that back this up.
  
One should always have good reasons to go against standard because there are always good reasons they are widely accepted.

.scalafmt.conf
cvogt: > I feel strongly that vertically aligned code is much more legible
I am with you @dwijnand ;). But I have no vote here and don't ask for one.

.travis.yml
cvogt: how about `git diff --exit-code` ;), quick, simple, efficient. After running a normal format that is.

dwijnand:
Clearly this PR is premature.

Before we end up forcing opinions and compromising values, let's take this topic back to planning.

I would advise separate issues:

+ one for line length
+ one for vertical alignment
+ one for whether or not we should use Olaf's embedded defaults or, if not, what options should we toggle

.scalafmt.conf
jozic: you can't make happy everyone, so it's better to make happier people who work with codebase more often
Eugene is a core contributor, so making him feel uncomfortable about writing/reading code because someone can read it from mobile is not fair, in my opinion. If scalajs core contributors read it on mobile, it's fine, they require it to be 80 as it's important to them
I'm not saying Eugene feels uncomfortable, he certainly can speak for himself, but my point is that most active contributors should have more weight on decisions like this

no intetion to make this a war, just a side-view :) :peace_symbol:

tpolecat:
I wish to register a belated  (╯°□°）╯︵ ┻━┻  about this. Just pick something and click go. You're never going to get everyone to agree.

But having said that, 80 chars is ridiculous. I agree 1000% with @eed3si9n

olafurpg:
My thoughts

- maxColumn 100 is totally fine. The 80 char default was chosen when I shared an office with @sjrd who convinced me 80 was a safe default (I still think it is, I personally review code on mobile). My experience with scalafmt is that most people prefer 100 or 120 (seems popular choice for IntelliJ users).
- don't use vertical alignment, it's distracting. I personally would like to move it into a separate module away from scalafmt. Vertical alignment is anyways implemented as a (fairly hacky) post-processing step on top of the formatted output.
- use default or I'm happy to discuss adding `style=sbt`, sbt is a large enough project to merit its own style. Please don't configure several random options in .scalafmt.conf, provided styles are better tested.

I've digged a bit around the sbt sources and I think it might actually look best with bin-packing enabled. There seems to be quite a lot of methods with multiple (>4) non-vararg parameters in the sbt sources, for which non-bin-packing can be wasteful

```scala
foo(
  a,
  b,
  c,
  d,
  e,
  f)
// bin-pack
foo(a, b, c,
  d, e, f)
```

 Bin-packing is currently not supported in scalafmt. I've gotten really positive results for bin-packing using paiges: https://github.com/johnynek/paiges. Here one example output from formatting pathologically nested input: https://gist.github.com/olafurpg/44ee2d0a8c5b8330702cf00d23d0978a I can't promise when bin-packing will get into scalafmt, I'm currently waiting for a 0.1 release of paiges.

ShaneDelmore:
I have no opinion on a code format but would like to see some consistency.  How about we just pick a style, and for anyone who doesn't like it, just have a localDev.conf with your preferred config, then use localDev.conf while you are working on code, and standard conf before you checkin?  May not be the most elegant solution but almost anything is better than doing nothing.

olafurpg:
@ShaneDelmore that will not work because the formatted output is a function of both token trivia along with the ast. For example, scalafmt respects line breaks in a few places
```scala
// max column             |
// original
def foo =
  2
// formatted
def foo =
  2
// even if the statement could fit within maxColumn
def foo = 2
```
If localDev.conf has maxColumn=80 and the repo.conf has maxColumn=120, then the local configuration will insert line breaks which won't be reverted with repo.conf.

For the record, even formatters like prettier that are more opinionated than scalafmt also work the same way.

PR #3014, opened 2017-03-14
Check for regular file in Package.makeJar

### LINKED ISSUE #3010: Package.makeJar deletes folder recursively
## steps

Call [`Package.makeJar`](https://github.com/sbt/sbt/blob/v0.13.13/main/actions/src/main/scala/sbt/Package.scala#L93-L99) with `jar` set to an existing _folder_.

Exhibited in a plugin (sbt/sbt-assembly#216), but can probably also be exhibited by setting the a path for `packageBin`, ... which is a directory.

## problem

Here's the relevant code:

```scala
  def makeJar(sources: Seq[(File, String)], jar: File, manifest: Manifest, log: Logger): Unit = {
    log.info("Packaging " + jar.getAbsolutePath + " ...")
    IO.delete(jar)
    log.debug(sourcesDebugString(sources))
    IO.jar(sources, jar, manifest)
    log.info("Done packaging.")
  }
```

The code assumes that `jar` is a file and deletes it. If however `jar` is a folder, it will get recursively deleted.

## expectation

My expectation is that any code which is tasked with writing out _a regular file_ should overwrite existing _regular files_ but not a _folder._ Hence, `makeJar` should error out when `jar` is anything but a regular file (or non-existing).

## notes

* Can be easily changed by adding a guard there (`if (jar.isFile() || !jar.exists())`).
* There are potentially other occurrences of the same problem.

larsrh:
Fixes #3010.

### DIFF
--- a/main/actions/src/main/scala/sbt/Package.scala
+++ b/main/actions/src/main/scala/sbt/Package.scala
@@ -91,8 +91,13 @@ object Package {
       ManifestAttributes((attribKeys zip attribVals) ++ { homepage map (h => (IMPLEMENTATION_URL, h.toString)) }: _*)
     }
   def makeJar(sources: Seq[(File, String)], jar: File, manifest: Manifest, log: Logger): Unit = {
-    log.info("Packaging " + jar.getAbsolutePath + " ...")
-    IO.delete(jar)
+    val path = jar.getAbsolutePath
+    log.info("Packaging " + path + " ...")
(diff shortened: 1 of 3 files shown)


eed3si9n (approved):
LGTM pending Travis

larsrh:
~The build failure in `safeUnitTests` doesn't seem to be caused by me.~

PR #3015, opened 2017-03-14
Fix #3013: ScriptedPlugin: Add support for paginated tests

### LINKED ISSUE #3013: ScriptedPlugin: Add support for paginated tests (scripted foo/*1of3)
This is another issue like https://github.com/sbt/sbt/issues/2601 where `ScriptedPlugin` is less powerful than the internal `Scripted` used by sbt, which supports paginated tests which are very useful when running on a CI: https://github.com/sbt/sbt/commit/5d5d7a6c0551beaeefffce61638ac0c8c29e57ec

smarter:
This was already supported in the internal Scripted used by sbt but not
in the ScriptedPlugin. This is fixed by just copy-pasting the modified
parser. We will have to wait for sbt itself to be built using an sbt
with the upgraded ScriptedPlugin to be able to avoid the code duplication.

### DIFF
--- a/notes/0.13.14/paginate_scripted_tests.md
+++ b/notes/0.13.14/paginate_scripted_tests.md
@@ -1,0 +1,12 @@
+### Improvements
+
+- ScriptedPlugin: Add the ability to paginate scripted tests.
+  It is now possible to run a subset of scripted tests in a directory at once,
+  for example:
+  ```
+  scripted source-dependencies/*1of3
+  ```
+  Will create three pages and run page 1. This is especially useful when running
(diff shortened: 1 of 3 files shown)


smarter:
I would love to get this in 0.13.14 if possible. Our scripted tests for dotty incremental compilation have never been run on our CI just because of the time they take.

eed3si9n:
Rebased to 0.13.14 branch.

eed3si9n (approved):

dwijnand:
Cool. Thanks @smarter.

PR #3019, opened 2017-03-15
Speed up startup and reload times

jvican:
These changes are optimizations that affect the following sbt subsytems:

1. Sbt parser.
2. Plugin discovery.

A concise explanation for every change is provided in the commit message.  In
short, this PR makes sure that we don't load the whole Scala reflect universe
to check whether a plugin has a `autoImport` member or to parse a sbt file.

In informal measurements, these changes affect around ~12-14% of the whole
startup time. These are the results before:

```
jvican in /data/rw/code/scala/scala
> $ time sbt "exit"
[info] Loading global plugins from /home/jvican/.sbt/0.13/plugins
[info] Loading project definition from /data/rw/code/scala/scala/project/project
[info] Loading project definition from /data/rw/code/scala/scala/project
[info] Resolving key references (11310 settings) ...
[info] *** Welcome to the sbt build definition for Scala! ***
[info] Check README.md for more information.
sbt -mem 2048 "exit"  29.66s user 0.56s system 297% cpu 10.171 total

jvican in /data/rw/code/scala/scala
> $ time sbt "exit"
[info] Loading global plugins from /home/jvican/.sbt/0.13/plugins
[info] Loading project definition from /data/rw/code/scala/scala/project/project
[info] Loading project definition from /data/rw/code/scala/scala/project
[info] Resolving key references (11310 settings) ...
[info] *** Welcome to the sbt build definition for Scala! ***
[info] Check README.md for more information.
sbt -mem 2048 "exit"  30.52s user 0.46s system 299% cpu 10.337 total

jvican in /data/rw/code/scala/scala
> $ time sbt "exit"
[info] Loading global plugins from /home/jvican/.sbt/0.13/plugins
[info] Loading project definition from /data/rw/code/scala/scala/project/project
[info] Loading project definition from /data/rw/code/scala/scala/project
[info] Resolving key references (11310 settings) ...
[info] *** Welcome to the sbt build definition for Scala! ***
[info] Check README.md for more information.
sbt -mem 2048 "exit"  28.85s user 0.43s system 293% cpu 9.988 total
```

And after:

```
jvican in /data/rw/code/scala/scala
> $ time sbt "exit"
[info] Loading global plugins from /home/jvican/.sbt/0.13/plugins
[info] Loading project definition from /data/rw/code/scala/scala/project/project
[info] Loading project definition from /data/rw/code/scala/scala/project
[info] Resolving key references (11355 settings) ...
[info] *** Welcome to the sbt build definition for Scala! ***
[info] Check README.md for more information.
sbt -mem 2048 "exit"  24.83s user 0.51s system 286% cpu 8.831 total

jvican in /data/rw/code/scala/scala
> $ time sbt "exit"
[info] Loading global plugins from /home/jvican/.sbt/0.13/plugins
[info] Loading project definition from /data/rw/code/scala/scala/project/project
[info] Loading project definition from /data/rw/code/scala/scala/project
[info] Resolving key references (11355 settings) ...
[info] *** Welcome to the sbt build definition for Scala! ***
[info] Check README.md for more information.
sbt -mem 2048 "exit"  24.46s user 0.46s system 287% cpu 8.674 total

jvican in /data/rw/code/scala/scala
> $ time sbt "exit"
[info] Loading global plugins from /home/jvican/.sbt/0.13/plugins
[info] Loading project definition from /data/rw/code/scala/scala/project/project
[info] Loading project definition from /data/rw/code/scala/scala/project
[info] Resolving key references (11355 settings) ...
[info] *** Welcome to the sbt build definition for Scala! ***
[info] Check README.md for more information.
sbt -mem 2048 "exit"  25.12s user 0.45s system 290% cpu 8.802 total
```

### DIFF
--- a/main/src/main/scala/sbt/Plugins.scala
+++ b/main/src/main/scala/sbt/Plugins.scala
@@ -340,18 +340,27 @@ ${listConflicts(conflicting)}""")
       case ap: AutoPlugin => model(ap)
     }
 
+  private val autoImport = "autoImport"
+
+  /** Determines whether a plugin has a stable autoImport member by:
+    *
+    * 1. Checking whether there exists a public field.
+    * 2. Checking whether there exists a public object.
+    *
+    * The above checks work for inherited members too.
+    *
+    * @param ap The found plugin.
+    * @param loader The plugin loader.
+    * @return True if plugin has a stable member `autoImport`, otherwise false.
+    */
   private[sbt] def hasAutoImportGetter(ap: AutoPlugin, loader: ClassLoader): Boolean = {
-    import reflect.runtime.{universe => ru}
+    val pluginClazz = ap.getClass
     import util.control.Exception.catching
-    val m = ru.runtimeMirror(loader)
-    val im = m.reflect(ap)
-    val hasGetterOpt = catching(classOf[ScalaReflectionException]) opt {
-      im.symbol.asType.toType.declaration(ru.newTermName("autoImport")) match {
-        case ru.NoSymbol => false
-        case sym => sym.asTerm.isGetter || sym.asTerm.isModule
-      }
-    }
-    hasGetterOpt getOrElse false
+    catching(classOf[NoSuchFieldException])
+      .opt(pluginClazz.getField(autoImport))
+      .orElse(catching(classOf[ClassNotFoundException]).opt(
+        Class.forName(s"${pluginClazz.getName}$autoImport$$", false, loader)))
+      .isDefined
   }
 
   /** Debugging method to time how long it takes to run various compilation tasks. */
--- a/main/src/main/scala/sbt/internals/parser/SbtParser.scala
+++ b/main/src/main/scala/sbt/internals/parser/SbtParser.scala
@@ -3,15 +3,17 @@ package internals
 package parser
 
 import java.io.File
+import java.util.concurrent.ConcurrentHashMap
 
 import sbt.internals.parser.SbtParser._
 
 import scala.compat.Platform.EOL
-import scala.reflect.internal.util.BatchSourceFile
+import scala.reflect.internal.util.{ BatchSourceFile, Position }
 import scala.reflect.io.VirtualDirectory
 import scala.tools.nsc.interactive.RangePositions
 import scala.tools.nsc.{ CompilerCommand, Global }
-import scala.tools.nsc.reporters.StoreReporter
+import scala.tools.nsc.reporters.{ Reporter, StoreReporter }
+import scala.util.Random
 
 private[sbt] object SbtParser {
   val END_OF_LINE_CHAR = '\n'
@@ -37,17 +39,49 @@ private[sbt] object SbtParser {
    * Provides the previous error reporting functionality in
    * [[scala.tools.reflect.ToolBox]].
    *
-   * This is a sign that this whole parser should be rewritten.
+   * This parser is a wrapper around a collection of reporters that are
+   * indexed by a unique key. This is used to ensure that the reports of
+   * one parser don't collide with other ones in concurrent settings.
+   *
+   * This parser is a sign that this whole parser should be rewritten.
    * There are exceptions everywhere and the logic to work around
    * the scalac parser bug heavily relies on them and it's tied
    * to the test suite. Ideally, we only want to throw exceptions
    * when we know for a fact that the user-provided snippet doesn't
    * parse.
    */
-  private[sbt] class ParserStoreReporter extends StoreReporter {
-    def throwParserErrorsIfAny(fileName: String): Unit = {
-      if (parserReporter.hasErrors) {
-        val seq = parserReporter.infos.map { info =>
+  private[sbt] class UniqueParserReporter extends Reporter {
+
+    private val reporters = new ConcurrentHashMap[String, StoreReporter]()
+
+    override def info0(pos: Position, msg: String, severity: Severity, force: Boolean): Unit = {
+      val reporter = getReporter(pos.source.file.name)
+      severity.id match {
+        case 0 => reporter.info(pos, msg, force)
+        case 1 => reporter.warning(pos, msg)
+        case 2 => reporter.error(pos, msg)
+      }
+    }
+
+    def getOrCreateReporter(uniqueFileName: String): StoreReporter = {
+      val reporter = reporters.get(uniqueFileName)
+      if (reporter == null) {
+        val newReporter = new StoreReporter
+        reporters.put(uniqueFileName, newReporter)
+        newReporter
+      } else reporter
+    }
+
+    private def getReporter(fileName: String) = {
+      val reporter = reporters.get(fileName)
+      if (reporter == null)
+        sys.error(s"Sbt parser failure: no reporter for $fileName.")
+      reporter
+    }
+
+    def throwParserErrorsIfAny(reporter: StoreReporter, fileName: String): Unit = {
+      if (reporter.hasErrors) {
+        val seq = reporter.infos.map { info =>
           s"""[$fileName]:${info.pos.line}: ${info.msg}"""
         }
         val errorMessage = seq.mkString(EOL)
@@ -56,24 +90,24 @@ private[sbt] object SbtParser {
             s"$errorMessage\n${SbtParser.XmlErrorMessage}"
           else errorMessage
         throw new MessageOnlyException(error)
-      }
+      } else ()
     }
   }
 
-  private[sbt] final val parserReporter = new ParserStoreReporter
+  private[sbt] final val globalReporter = new UniqueParserReporter
 
   private[sbt] final lazy val defaultGlobalForParser = {
     import scala.reflect.internal.util.NoPosition
     val options = "-cp" :: s"$defaultClasspath" :: "-Yrangepos" :: Nil
-    val reportError = (msg: String) => parserReporter.error(NoPosition, msg)
+    val reportError = (msg: String) => globalReporter.error(NoPosition, msg)
     val command = new CompilerCommand(options, reportError)
     val settings = command.settings
     settings.outputDirs.setSingleOutput(new VirtualDirectory("(memory)", None))
 
-    // Mixing RangePositions is necessary, otherwise global ignores -Yrangepos
-    val global = new Global(settings, parserReporter) with RangePositions
+    // Mix RangePositions, otherwise global ignores -Yrangepos
+    val global = new Global(settings, globalReporter) with RangePositions
     val run = new global.Run
-    // Necessary to have a dummy unit for initialization...
+    // Add required dummy unit for initialization...
     val initFile = new BatchSourceFile("<wrapper-init>", "")
     val _ = new global.CompilationUnit(initFile)
     global.phase = run.parserPhase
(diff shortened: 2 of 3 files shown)


main/src/main/scala/sbt/internals/parser/SbtParser.scala
```diff
@@ -6,14 +6,100 @@ import java.io.File
...
+  private[sbt] final lazy val defaultGlobalForParser = {
+    import scala.reflect.internal.util.NoPosition
+    val options = "-cp" :: s"$defaultClasspath" :: "-Yrangepos" :: Nil
+    val reportError = (msg: String) => parserReporter.error(NoPosition, msg)
+    val command = new CompilerCommand(options, reportError)
+    val settings = command.settings
+    settings.outputDirs.setSingleOutput(new VirtualDirectory("(memory)", None))
+
+    // Mixing RangePositions is necessary, otherwise global ignores -Yrangepos
+    val global = new Global(settings, parserReporter) with RangePositions
```
eed3si9n: Isn't the purpose of the toolbox to provide some abstraction instead of strongly coupling the implementation of sbt with Scala compiler? It is possible to reuse the toolbox, or would that incur initialization penalty?

/cc @adriaanm, @retronym

jvican: This approach is better than reusing a toolbox, that relies on the `ReflectGlobal`. I explain more in the commit message.
  
This part of sbt already has access to `Global` since it uses it in `Eval`. There's no reason to use the toolbox, since the toolbox is meant for applications that don't have scala compiler as dependencies.

main/src/main/scala/sbt/Plugins.scala
```diff
@@ -341,17 +341,15 @@ ${listConflicts(conflicting)}""")
...
-    val im = m.reflect(ap)
-    val hasGetterOpt = catching(classOf[ScalaReflectionException]) opt {
-      im.symbol.asType.toType.declaration(ru.newTermName("autoImport")) match {
-        case ru.NoSymbol => false
-        case sym => sym.asTerm.isGetter || sym.asTerm.isModule
-      }
+    val pluginClazz = ap.getClass
+    val autoImportExists = catching(classOf[ClassNotFoundException]).opt {
+      val autoImportClassName = s"${pluginClazz.getName}autoImport$$"
+      val fields = pluginClazz.getDeclaredFields
```
dwijnand: I wonder if "getMethod" is better than searching the fields.

main/src/main/scala/sbt/Plugins.scala
```diff
@@ -341,17 +341,15 @@ ${listConflicts(conflicting)}""")
...
+      fields.exists(_.getName == "autoImport") ||
+        {Class.forName(autoImportClassName, false, loader); true}
```
dwijnand: I think you need to handle ClassNotFoundException: [Class#forName](http://docs.oracle.com/javase/6/docs/api/java/lang/Class.html#forName(java.lang.String,%20boolean,%20java.lang.ClassLoader))

main/src/main/scala/sbt/internals/parser/SbtParser.scala
```diff
@@ -6,14 +6,100 @@ import java.io.File
...
+  private[sbt] class ParserStoreReporter extends StoreReporter {
+    def throwParserErrorsIfAny(fileName: String): Unit = {
+      if (parserReporter.hasErrors) {
+        val seq = parserReporter.infos.map { info =>
+          s"""[$fileName]:${info.pos.line}: ${info.msg}"""
+        }
+        val errorMessage = seq.mkString(EOL)
+        val error: String =
+          if (errorMessage.contains(XML_ERROR))
+            s"$errorMessage\n${SbtParser.XmlErrorMessage}"
```
dwijnand, line 90: You can drop the `SbtParser.` here (and thus the braces)

main/src/main/scala/sbt/internals/parser/SbtParser.scala
```diff
@@ -3,17 +3,144 @@ package internals
...
+    val command = new CompilerCommand(options, reportError)
+    val settings = command.settings
+    settings.outputDirs.setSingleOutput(new VirtualDirectory("(memory)", None))
+
+    // Mix RangePositions, otherwise global ignores -Yrangepos
+    val global = new Global(settings, globalReporter) with RangePositions
+    val run = new global.Run
+    // Add required dummy unit for initialization...
+    val initFile = new BatchSourceFile("<wrapper-init>", "")
+    val _ = new global.CompilationUnit(initFile)
```
dwijnand, line 112: is the LHS val necessary? Do we depend on side-effects from calling the constructor of CompilationUnit?

dwijnand (changes requested):
LGTM, apart from the ClassNotFoundException part.

The rest are thoughts, questions and minor points/issues.

dwijnand:
2 more things: (1) I restarted the failure on Travis CI, and (2) nice work!

main/src/main/scala/sbt/Plugins.scala
eed3si9n: Isn't that handled with `catching(classOf[ClassNotFoundException]).opt` a few lines above?

main/src/main/scala/sbt/Plugins.scala
jvican: It is, I'll replace.

main/src/main/scala/sbt/Plugins.scala
jvican: Indeed.

main/src/main/scala/sbt/internals/parser/SbtParser.scala
jvican: Yes, black magic of the compiler...

main/src/main/scala/sbt/Plugins.scala
dwijnand: You're right. I'm blind. But personally I'd try and wrap things as small as possible.

main/src/main/scala/sbt/Plugins.scala
dwijnand: When you do, mind the NoSuchMethodExceptions.

main/src/main/scala/sbt/internals/parser/SbtParser.scala
dwijnand: yes to both?? or just the 2nd question?

dwijnand (approved):

main/src/main/scala/sbt/internals/parser/SbtParser.scala
jvican: Just the last one, the reason I use `val _` is to make it explicit that the line is really required and it's not any dead code (some IDEs may report otherwise).

dwijnand: Some IDEs got to get with the system. We _side-effect_ in _constructors_ of course!

retronym:
I'd suggest to drop the last commit for now. Parsing is really quite cheap once you've got the compiler initialized, so I don't think future proofing this code so as not to be a point of contention for a hypothetical multi-threaded project load is worth.

retronym:
I'd also suggest starting with an initial commit that reuses the `ToolBox` based parser. If for some reason, we have to revert the change to use `Global` directly (hypothetically this might happen if the compiler API changes in a future compiler that SBT upgrades to and this code is struggling to fina a maintainer), we'll still have the main performance improvement.

main/src/main/scala/sbt/Plugins.scala
retronym: This should walk up the superclass chain in case the `AutoPlugin` inherits the `autoImport` member.

We no longer can check whether or not `autoImport` is a stable term (`val` or `object`), but I think it is fine to document that plugin authors must not declare `var autoImport`.

main/src/main/scala/sbt/internals/parser/SbtParser.scala
```diff
@@ -130,7 +179,7 @@ private[sbt] case class SbtParser(file: File, lines: Seq[String]) extends Parsed
      * @return originalStatement or originalStatement with missing bracket
      */
     def parseStatementAgain(t: Tree, originalStatement: String): String = {
-      val statement = util.Try(toolbox.parse(originalStatement)) match {
+      val statement = util.Try(parse(originalStatement, fileName)) match {
         case util.Failure(th) =>
           val missingText = findMissingText(content, t.pos.end, t.pos.line, fileName, th)
           originalStatement + missingText
```
retronym: I know this triple-parsing rigmarole bugs you, so I thought I'd leave you with the idea to ignore the end position of parsed statements, and instead  use the start position of the subsequent statement (or the end position of the enclosing tree, if there is no subsequent statement) - 1.

main/src/main/scala/sbt/Plugins.scala
jvican: You're right, this should also be done for super classes. I actually thought this was not possible in the current plugin architecture.

main/src/main/scala/sbt/internals/parser/SbtParser.scala
jvican: This is such a great idea. Thank you.

jvican:
> I'd also suggest starting with an initial commit that reuses the ToolBox based parser. If for some reason, we have to revert the change to use Global directly (hypothetically this might happen if the compiler API changes in a future compiler that SBT upgrades to and this code is struggling to fina a maintainer), we'll still have the main performance improvement.
  
The problem is that toolbox is not thread-safe. You cannot use it in the same way I'm using it now because reporter errors will clash with each other. This is the reason why I'm doing all the magic with the `reporterId`s. I actually think that we may want to parse sbt files in parallel (if sbt is not already doing this). I heard somewhere that scalac was parsing in parallel too.
    
> Parsing is really quite cheap once you've got the compiler initialized

Isn't there an overhead of using `ReflectGlobal` over just `Global`? The current approach in this commit, as I see it, avoids class loading some classes.
  
True, parsing is quite cheap. The reason I went for this solution though is because I gained 400ms. I'll repeat my measurements with the global toolbox, but we still have the issue that is not thread-safe and we cannot instrument it.
  
However, a bigger benefit with this approach is that we can reuse the parsed trees from this concrete parser and feed them into eval, instead of parsing the whole compilation unit again. This will reduce quite some code and ease maintenance. Potentially, we could even reuse the global of eval for parsing, avoiding the initialization of the extra global, which I believe it may be non-negligible. WDYT?

retronym:
The lack of thread safety is moot if we live with synchronisation. I
believe will still be a big win by avoiding unneeded init cost

I'm in favour of the end result of using global directly; but I think it's
worth an extra commit to make that a separate step.

jvican:
@retronym So what you propose is that I include a commit before my changes to parser that uses a global toolbox instance and is synchronized, no?
  
I'm happy to do that, I have that commit lying around my computer somewhere.

retronym:
👍

main/src/main/scala/sbt/Plugins.scala
jvican: I've tried this out, but it's not as simple as it seems since the tests check that the current sbt behaviour is exactly the same as the one before, and there are some mismatches in the way both parsers handle positions and new lines. I'm looking into this problem and hope to reconcile them, but we should probably stage this to another PR.

jvican:
I'm closing this because I'm reopening against 1.0.

jvican:
Hello, Jason's changes have been applied in https://github.com/sbt/sbt/pull/3115, which is rebased to latest master 1.0.

PR #3025, opened 2017-03-16
Write sbt.version to project/build.properties

### LINKED ISSUE #754: Print a warning if sbt.version is not defined in the project/build.properties
This would help creating more reproducible builds.

dwijnand:
Have sbt.version set in project/build.properties is a best practice
because it makes the build more deterministic and reproducible.

With this change sbt, after ensuring that the base directory is probably
an sbt project, writes out sbt.version in project/build.properties if it
is missing.

Fixes #754


main/src/main/scala/sbt/Main.scala
```diff
@@ -13,25 +13,55 @@ import scala.annotation.tailrec
...
       )
     }
+
+  private val sbtVersionRegex = """sbt\.version\s*=.*""".r
+  private def isSbtVersionLine(s: String) = sbtVersionRegex.pattern matcher s matches ()
+
+  private def isSbtProject(baseDir: File, projectDir: File) =
+    projectDir.exists() || (baseDir * "*.sbt").get.nonEmpty
+
+  private def setSbtVersion(baseDir: File, sbtVersion: String) = {
```
eed3si9n, line 43: We might want to put opt-out Java flag around this method.

eed3si9n (changes requested):

eed3si9n (approved):

main/src/main/scala/sbt/Main.scala
```diff
@@ -13,25 +13,56 @@ import scala.annotation.tailrec
...
+    val buildPropsLines = if (buildProps.canRead) IO.readLines(buildProps) else Nil
+
+    val sbtVersionAbsent = buildPropsLines forall (!isSbtVersionLine(_))
+
+    if (sbtVersionAbsent) {
+      val errorMessage = s"WARN: No sbt.version set in project/build.properties, base directory: $baseDir"
+      try {
+        if (isSbtProject(baseDir, projectDir)) {
+          val newBuildPropsLines = s"sbt.version=$sbtVersion" :: buildPropsLines
+          IO.writeLines(buildProps, newBuildPropsLines)
```
Duhemm: I think sbt should tell the user that it's writing the file. Software should not perform side effects silently without explicit user consent.

eed3si9n: Between not booting up at all and creating `build.properties`, creating `build.properties` I think is lesser of two bad scenarios.

Duhemm: I agree. My point is that sbt should also say "don't worry, I'll write it out for you".

eed3si9n: ok. I don't have a strong opinion on displaying warning or silently creating the file. We can go either way.

dwijnand: Oh good, because I agree with Martin.

PR #3031, opened 2017-03-18
Bump underlying modules to latest

eed3si9n:
Adjust to module updates.
Compiles locally, but it required some changes to other module source, so this is not going to compile yet.

- https://github.com/sbt/zinc/pull/271
- https://github.com/sbt/librarymanagement/pull/75


dwijnand (approved):
🤞 on a green Travis CI

PR #3035, opened 2017-03-21
Notify about shell less

dwijnand:
+ Don't notify ScriptMain users by moving the logic to xMain
+ Only trigger shell if shell is a defined command
+ Use existing Shell/BootCommand strings instead of new ones


eed3si9n:
Can this change wait till 0.13.15?
I don't know if we merge this it requires RC-3.

dwijnand:
@ekrich was keen to have this fix. I'm not fussed.

eed3si9n:
I guess from the scripting point of view it's a regression/annoyance.
My vote is to merge it, but not postpone the final date then.

ekrich:
That sounds good to me. I trust the testing and don't want to push the
date. If you do push a RC3 I can confirm the fix.

PR #3043, opened 2017-03-25
Fix race condition

benmccann:
This was exposed by Play's scripted tests in https://github.com/playframework/playframework/pull/7133

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -150,7 +150,7 @@ object Defaults extends BuildCommon {
     Previous.references :== new Previous.References,
     concurrentRestrictions := defaultRestrictions.value,
     parallelExecution :== true,
-    pollInterval :== 500,
+    pollInterval :== 1000,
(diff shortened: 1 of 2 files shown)


He-Pin:
Could we make use of NIO2 once we move to java8?

benmccann:
Does NIO2 have this functionality?

NIO does in jdk7, but it doesn't work on Mac. In Play we use NIO on Windows and Linux and jnotify/polling on Mac. See https://github.com/playframework/play-file-watch/blob/master/src/main/scala/play/dev/filewatch/FileWatchService.scala#L67

He-Pin:
Does monix's nio address this?
https://github.com/monix/monix-nio

@radusw ping?

radusw:
@hepin1989 It will pretty soon. Currently we have this PR: https://github.com/monix/monix-nio/pull/16 by @hmemcpy ( an example: https://twitter.com/hmemcpy/status/844521638431526918 )

benmccann:
I'm temporarily closing this pending further investigation. I think there might be a better fix

PR #3068, opened 2017-04-03
Kill settingsSets

### LINKED ISSUE #2617: AddSettings must not be internal._, because it is used in public APIs
Similarly to #2616, `sbt.internal.AddSettings` should not be internal, as it is used in the public method `Project.settingSets`.

dwijnand:
Fixes #2617


eed3si9n:
Plz add to https://github.com/sbt/sbt/blob/1.0.x/MIGRATION.md

eed3si9n (approved):

PR #3069, opened 2017-04-04
Switch Defaults.runnerTask back to a setting

dwijnand:
I believe this was an unintentional change in
74cfbd4a9ca162dbe9d6fc236ac19671ff16d951.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -898,7 +898,7 @@ object Defaults extends BuildCommon {
         scalaRun.value.run(mainClass, data(classpath.value), parser.parsed, streams.value.log).get
       }
     }
-  def runnerTask: Initialize[Task[ScalaRun]] = runnerInit
+  def runnerTask: Setting[Task[ScalaRun]] = runner := runnerInit.value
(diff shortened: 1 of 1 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -898,7 +898,7 @@ object Defaults extends BuildCommon {
         scalaRun.value.run(mainClass, data(classpath.value), parser.parsed, streams.value.log).get
       }
     }
-  def runnerTask: Initialize[Task[ScalaRun]] = runnerInit
+  def runnerTask: Setting[Task[ScalaRun]] = runner := runnerInit.value
```
eed3si9n, line 901: If it's a setting we should rename this to `runnerTaskSetting`.

eed3si9n (changes requested):

main/src/main/scala/sbt/Defaults.scala
dwijnand: But the only motivation for this change is to restore source-compatibility with 0.13..

eed3si9n: ok

eed3si9n (approved):

PR #2764, opened 2016-09-30
Deprecate tuple enrichments. Fixes #2763

dwijnand:
Fixes #2763

PR #3079, opened 2017-04-05
Removals: tuple enrichments, old ops, project delegation & deprecated

dwijnand:
- https://github.com/sbt/sbt/pull/2764
- Deprecate the old operators `<<=`, `<+=`, and `<++=` https://github.com/sbt/sbt/pull/2716


jvican:
Can we bring some of these decisions up to discussion?
  
I feel like tuple enrichments should not be removed. The semantics of `.value` are not the same because don't follow Scala-like evaluation semantics. Removing them pushes people to use the DSL API without a clear solution in mind.
  
The DSL API does not give users the freedom to choose how tasks should be evaluated (some times you want full control over task evaluation). This is not possible with the DSL API.

eed3si9n:
> Can we bring some of these decisions up to discussion?

Yes.

> I feel like tuple enrichments should not be removed. The semantics of `.value` are not the same because don't follow Scala-like evaluation semantics. Removing them pushes people to use the DSL API without a clear solution in mind.

In 2017, if we have to teach new users one style of DSL, it's the sbt 0.13 style `{ val k = key.value; val k2 = key2.value }`, and I would really like to reduce the amount of DSL the user has to learn to master sbt. For migration purpose, I'm open to keeping tuple enrichment as a plugin. WDYT?

main-settings/src/main/scala/sbt/Structure.scala
```diff
@@ -40,8 +40,6 @@ sealed abstract class SettingKey[T] extends ScopedTaskable[T] with KeyedInitiali
   final def :=(v: T): Setting[T] = macro std.TaskMacro.settingAssignMacroImpl[T]
   final def +=[U](v: U)(implicit a: Append.Value[T, U]): Setting[T] = macro std.TaskMacro.settingAppend1Impl[T, U]
   final def ++=[U](vs: U)(implicit a: Append.Values[T, U]): Setting[T] = macro std.TaskMacro.settingAppendNImpl[T, U]
-  final def <+=[V](v: Initialize[V])(implicit a: Append.Value[T, V]): Setting[T] = macro std.TaskMacro.fakeSettingAppend1Position[T, V]
```
eed3si9n, line 43: These are error macros (Restligeist macro).

eed3si9n (changes requested):

main-settings/src/main/scala/sbt/Structure.scala
dwijnand: Can't they be regular errors in 1.0?

dwijnand:
Oh man, looks like we were using a lot of deprecated APIs in the scripted test.

I think I might break this PR into 4 PRs for each atomic commit and deal with them independently.

main-settings/src/main/scala/sbt/Structure.scala
eed3si9n: No, the point of error macro is to fail at compile time, but with customized error message pointing to the migration guide.

dwijnand: Ok, I'll drop the commit.

jvican:
>  In 2017, if we have to teach new users one style of DSL, it's the sbt 0.13 style { val k = key.value; val k2 = key2.value }, and I would really like to reduce the amount of DSL the user has to learn to master sbt. For migration purpose, I'm open to keeping tuple enrichment as a plugin. WDYT?
  
In my opinion, it deserves built-in support in sbt. I'm okay with sbt leaning more towards the 0.13 style, but we have to acknowledge its limitations. Power users or people that don't want to deal with DSL limitations, can always use the non DSL API, and I think it's fine to leave it as it is since the maintenance burden is negligible. If we put this API in an independent plugin, then we're leaving sbt users with only one way of doing stuff, but this way is incomplete and has fundamental problems (Scala evaluation semantics are not preserved).
  
Another reason why tuple enrichments should stay is because there is no way to provide automatic rewrites for them. If we do, we change semantics, since tasks won't be evaluated in parallel anymore.
  
Until a full solution is not worked out in the DSL API, I feel that we should leave tuple enrichments intact. These things can go away in the future when the DSL behaves like Scala code.

eed3si9n:
> This way is incomplete and has fundamental problems (Scala evaluation semantics are not preserved).

The fact that tasks do not evaluate in sequential order is by design, not really a problem.
We use the DSL to implement DAG of happens-before relationship as documented in [Task graph](http://www.scala-sbt.org/0.13/docs/Task-Graph.html). If we evaluated tasks the same way we evaluate functions, then `compile` called n times in `test` would cause 5 compilation.

> Power users or people that don't want to deal with DSL limitations, can always use the non DSL API, and I think it's fine to leave it as it is since the maintenance burden is negligible.

Power user can also use sequential tasks and dynamic tasks - http://www.scala-sbt.org/0.13/docs/Howto-Sequencing.html

> These things can go away in the future when the DSL behaves like Scala code.

The tasks run in parallel, so it won't behave like Scala code. At the same time, tuple enrichment do not look anything like what's in the Standard library. (Cats has [Cartesian operations with tuples](http://typelevel.org/cats/typeclasses/applicative.html))

> Another reason why tuple enrichments should stay is because there is no way to provide automatic rewrites for them. If we do, we change semantics, since tasks won't be evaluated in parallel anymore.

I'm a bit confused here. In `{ val k = key.value; val k2 = key2.value }`, `key` and `key2` evaluate in parallel. Are you concerned about the rewrite of `map` or `flatMap`?

jvican:
> The fact that tasks do not evaluate in sequential order is by design, not really a problem.
  
It is a problem, sometimes you may want to control parallel task execution.

I would argue that this is not by design, since the non-DSL API allows you to have control over these things. The deduplication property applies only to the DSL, and I think it's good to have this distinction and keep it this way. A good benefit of sbt is how flexible it is. I believe we can keep it this way without removing core parts of the API, and removing the parts that are more painful (as you're already doing in other PRs).
  
> Power user can also use sequential tasks and dynamic tasks - http://www.scala-sbt.org/0.13/docs/Howto-Sequencing.html
  
Dynamic tasks are difficult to grasp and the consequence of a limited macro implementation, not a design decision. The fact that tasks in `if` will always be executed could be fixed with a better macro implementation, and still achieve the properties of `inspect` and `deduplication` (with some magic for the first one, though).
  
They are the reason why sbt users find sbt so difficult. We're embedding basic conditionals into a DSL construct that people have to learn to use and identify its need, while an `if` is a fundamental construct of any programming language.

> The tasks run in parallel, so it won't behave like Scala code.
  
What I meant by this is fixing some macros to make control flow have an influence on task execution (as described in my previous example).
  
> Are you concerned about the rewrite of map or flatMap?
  
Yes, I was explicitly talking about the rewrite of tuple enrichments into map or flatMap calls, which should be nested and therefore not parallel.
  
The issue is that the `.value` version also changes semantics since control flow inside bodies of tuple enrichments won't be respected anymore. This is why I meant there is no direct rewrite of tuple enrichments.

eed3si9n:
@jvican Let's continue the discussion in https://github.com/sbt/sbt/pull/3083

PR #3082, opened 2017-04-05
Remove most @deprecated

### LINKED ISSUE #3078: Remove deprecated for 1.0.0
Current status:

```
$ git grep '@deprecated' | wc -l
     121
```

HEAD is 1f02e61819ad3c1eedda73007ecd63663a0461bb.

dwijnand:
Some have become private[sbt] instead.

Additionally:
* space out terms
* remove commented-out blocks
* hand format long lines
* some general cleanup

Closes #3078 (not all are removed, but enough to close the issue)

### DIFF
--- a/main-settings/src/main/scala/sbt/Reference.scala
+++ b/main-settings/src/main/scala/sbt/Reference.scala
@@ -94,13 +94,4 @@ object Reference {
     case BuildRef(b)      => Some(b)
     case _                => None
   }
-
-  @deprecated("Explicitly wrap the URI in a call to RootProject.", "0.13.0")
-  implicit def uriToRef(u: URI): ProjectReference = RootProject(u)
-
-  @deprecated("Explicitly wrap the File in a call to RootProject.", "0.13.0")
-  implicit def fileToRef(f: File): ProjectReference = RootProject(f)
-
-  @deprecated("Explicitly wrap the String in a call to LocalProject.", "0.13.0")
-  implicit def stringToReference(s: String): ProjectReference = LocalProject(s)
 }
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -17,3 +17,8 @@ Migration notes
 - `TestResult.Value` is now `TestResult`.
 - the scripted plugin is cross-versioned now, so you must use %% when depending on it
 - Removed the methods settingsSets from Project (along with add/setSbtFiles)
+- Dropped deprecated InputTask apply method and inputTask DSL method - replace with `Def.inputTask` & `Def.spaceDelimited().parsed`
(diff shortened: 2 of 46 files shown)


eed3si9n (changes requested):
A bunch of failures on Travis. Some of them I'm sure are legit.

```
[error] 	actions / cross-multiproject
[error] 	actions / cross
[error] 	actions / depends-on
[error] 	actions / state
[error] 	actions / update-state-fail
[error] 	apiinfo / unstable-existential-names
[error] 	compiler-project / inc-pickled-existential
[error] 	compiler-project / inc-pickled-refinement
[error] 	compiler-project / src-dep-plugin
[error] 	dependency-management / ivy-settings-a
[error] 	dependency-management / ivy-settings-multi-b
[error] 	dependency-management / override
[error] 	dependency-management / override2
[error] 	dependency-management / pom-advanced
[error] 	dependency-management / pom-classpaths
[error] 	dependency-management / provided
[error] 	project / load-hooks
[error] 	project / scripted-plugin
[error] 	project / src-plugins
[error] 	source-dependencies / abstract-type-override
[error] 	tests / empty
[error] 	tests / fork
[error] 	tests / serial
```

dwijnand:
* reverted a change in BuildUtil.aggregationRelation (left a comment) which was breaking actions/cross-multiproject and actions/update-state-fail
* dropped the attempt to remove the deprecated methods in scripted because I can't figure out how to fix project/scripted-plugin
* dropped the attempt to remove the deprecated DefinableTask#task because I can't figure out how to fix actions/depends-on
* fixed all the other broken scripted tests due to the deprecation removal

If Travis CI passes I'll squash this down.

eed3si9n (approved):

main-settings/src/main/scala/sbt/Reference.scala
```diff
@@ -94,13 +94,4 @@ object Reference {
     case BuildRef(b)      => Some(b)
     case _                => None
   }
-
-  @deprecated("Explicitly wrap the URI in a call to RootProject.", "0.13.0")
-  implicit def uriToRef(u: URI): ProjectReference = RootProject(u)
```
eed3si9n, line 99: Should some of these survive as Restligeist macro?

eed3si9n:
I know we're technically just removing deprecations, but maybe some of the popularly used methods should be mentioned in the migration guide and/or Restligeist macro.

main-settings/src/main/scala/sbt/Reference.scala
dwijnand: I think 4 years is sufficient time to allow users to migrate before we drop the maintenance cost.. https://github.com/sbt/sbt/commit/0507a2a1a9be5006b6ffb4dee3318796b70fe290

PR #3083, opened 2017-04-05
[sbt 1.0] Make tuple apply/map/flatMap enrichment private[sbt]

### LINKED ISSUE #2762: Remove the tuple map enrichments

dwijnand:
Ref #2764
Fixes #2762


eed3si9n:
@jvican wrote in https://github.com/sbt/sbt/pull/3079:

> I feel like tuple enrichments should not be removed. The semantics of `.value` are not the same because don't follow Scala-like evaluation semantics. Removing them pushes people to use the DSL API without a clear solution in mind.
>
> The DSL API does not give users the freedom to choose how tasks should be evaluated (some times you want full control over task evaluation). This is not possible with the DSL API.

eed3si9n:
More in response to https://github.com/sbt/sbt/pull/3079#issuecomment-291996023:

> Another reason why tuple enrichments should stay is because there is no way to provide automatic rewrites for them. If we do, we change semantics, since tasks won't be evaluated in parallel anymore.

If this removing tuple enrichment is a hinderance to (automatic) sbt 1.0 migration, I'm ok with keeping this around.

eed3si9n:
I'm closing this for now.

PR #3098, opened 2017-04-12
[sbt 1.0] Make server command the default shell

eed3si9n:
This makes `server` command the default shell.

Note:

There used to be another commit about improving the error handling, but it's moved to another PR - https://github.com/sbt/sbt/pull/3104.

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -31,7 +31,7 @@ import java.net.URI
 import java.util.Locale
 import scala.util.control.NonFatal
 
-import BasicCommandStrings.{ Shell, Server }
+import BasicCommandStrings.{ Shell, OldShell }
 import CommandStrings.BootCommand
 
 /** This class is the entry point for sbt. */
@@ -92,7 +92,7 @@ final class xMain extends xsbti.AppMain {
   private def endsWithBoot(state: State) = state.remainingCommands.lastOption exists (_.commandLine == BootCommand)
 
   private def notifyUsersAboutShell(state: State) =
-    if (isInteractive && !hasCommand(state, Shell) && !hasCommand(state, Server) && !endsWithBoot(state)) {
+    if (isInteractive && !hasCommand(state, Shell) && !hasCommand(state, OldShell) && !endsWithBoot(state)) {
       state.log warn "Executing in batch mode."
       state.log warn "  For better performance, hit [ENTER] to switch to interactive mode, or"
       state.log warn "  consider launching sbt without any commands, or explicitly passing 'shell'"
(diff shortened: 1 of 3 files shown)


dwijnand (changes requested):
Let's find some time to talk about this together (perhaps later tonight) because I have the following concerns about making the server default:

* the recent addition about "keeping the shell warm" messaging as well as its implementation
* if perhaps we want to now fold the "server" command into shell, with possibly a "old-shell" fallback command..
* (minor) the fact that sbt-extras invokes sbt explicitly with "shell" while sbt-launcher-package doesn't

I'm happy to sign-off on your Future[Unit] change if you want to tear that out of here into a new PR.

eed3si9n:
> I'm happy to sign-off on your Future[Unit] change if you want to tear that out of here into a new PR.

ok. That's split out. https://github.com/sbt/sbt/pull/3104

eed3si9n:
After discussion with @dwijnand, the server command is now called `shell`, and the old shell command is called `oldshell` - 86342d0bdd78cc18ee280610742e46a1d434295c

main/src/main/scala/sbt/Main.scala
```diff
@@ -92,7 +92,7 @@ final class xMain extends xsbti.AppMain {
   private def endsWithBoot(state: State) = state.remainingCommands.lastOption exists (_.commandLine == BootCommand)
 
   private def notifyUsersAboutShell(state: State) =
-    if (isInteractive && !hasCommand(state, Shell) && !hasCommand(state, Server) && !endsWithBoot(state)) {
+    if (isInteractive && !hasCommand(state, Shell) && !hasCommand(state, OldShell) && !endsWithBoot(state)) {
```
dwijnand, line 95: This should be Shell

dwijnand (changes requested):

main/src/main/scala/sbt/Main.scala
eed3si9n: It has both Shell and OldShell in the if predicate.

dwijnand: Ah it also has Shell. @eed3si9n if possible could you draw attention to these details when forward-porting?

dwijnand (approved):

main/src/main/scala/sbt/Main.scala
eed3si9n: There's nothing to forward port since it's sbt 1.0 :)

eed3si9n: Oh you mean when I changed it from the 0.13?

dwijnand: I should've noticed the trailing commits in https://github.com/sbt/sbt/pull/3077

PR #3101, opened 2017-04-13
Document & cleanup commands (& misc cleanups)

dwijnand:
This was part of my branch to add the "addSbtPlugin" command, which is now on the backburner.

Asides from the cleanups I'm particularly interested in getting the scaladocs for creating commands merged in.

### DIFF
--- a/main-command/src/main/scala/sbt/Command.scala
+++ b/main-command/src/main/scala/sbt/Command.scala
@@ -8,98 +8,155 @@ import sbt.internal.util.complete.{ DefaultParsers, EditDistance, Parser }
 import sbt.internal.util.Types.const
 import sbt.internal.util.{ AttributeKey, AttributeMap, Util }
 
+/**
+ * An operation that can be executed from the sbt console.
+ *
+ * <p>The operation takes a [[sbt.State State]] as a parameter and returns a [[sbt.State State]].
+ * This means that a command can look at or modify other sbt settings, for example.
+ * Typically you would resort to a command when you need to do something that's impossible in a regular task.
+ */
 sealed trait Command {
   def help: State => Help
   def parser: State => Parser[() => State]
+
   def tags: AttributeMap
   def tag[T](key: AttributeKey[T], value: T): Command
 }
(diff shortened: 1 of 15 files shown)


main-command/src/main/scala/sbt/Command.scala
```diff
@@ -8,98 +8,155 @@ import sbt.internal.util.complete.{ DefaultParsers, EditDistance, Parser }
...
   def make(name: String, briefHelp: (String, String), detail: String)(parser: State => Parser[() => State]): Command =
     make(name, Help(name, briefHelp, detail))(parser)
-  def make(name: String, help: Help = Help.empty)(parser: State => Parser[() => State]): Command = new SimpleCommand(name, help, parser, AttributeMap.empty)
 
-  def apply[T](name: String, briefHelp: (String, String), detail: String)(parser: State => Parser[T])(effect: (State, T) => State): Command =
-    apply(name, Help(name, briefHelp, detail))(parser)(effect)
+  // General command construction
+
+  /** Construct a command with the given name, parser and effect. */
   def apply[T](name: String, help: Help = Help.empty)(parser: State => Parser[T])(effect: (State, T) => State): Command =
```
jvican, line 65: Can we add another `apply` method that does not take a parser? I know that's covered by `command`. This would make it easier for people to define commands that take no input from the user, which is a pretty common use case, at least for my sbt workflow.

dwijnand: I don't think adding another way to do the same thing just with a different name is a good idea, in general. It's especially not a good idea when there are already literally 13 different methods to create commands. Also it's out of scope for this PR to be adding features.

jvican: My comment was relevant for this PR because the previous way to do it could be deprecated/removed and replaced by this new `apply` method. Now that we're doing breaking changes, it would be great if those breaking changes optimize for the widely used scenario.

dwijnand: Please could you open a separate issue to discuss adding features and breaking changes? I appreciate the request is related, but it's not directly relevant to what I'm doing here.

jvican: Yes, as you wish.

eed3si9n (approved):

PR #3120, opened 2017-04-19
Fix sbt 1.0 in Intellij

jvican:
This commit fixes most of the red squiggles created by the lack of
analysis of Contraband-generated Scala files. It does not end with all
the red squiggles but it substantially decreases the amount of them.


eed3si9n (approved):
LGTM. Maybe this is something I should put in the contraband plugin by default.

dwijnand (approved):
Ah cool. Yeah we should at the very least document this in Contraband.

PR #3121, opened 2017-04-19
Improve `Load` in sbt

jvican:
It mainly does three things:

* Clean up the implementation, removing unused values like
  `globalPluginDefs`.
* Make the implementation more readable.
* And the big one: Remove the creation of a classloader that we were
  instantiating but whose value we were throwing away. This has an impact
  in performance, but I'm yet to benchmark how much it is.
  
The second commit reduces the complexity around `loadPluginDefinition` et al.
`pluginDefinitionLoader` is not used anywhere in sbt, so the extra definitions are removed.

Both the implementation of `loadPluginDefinition` and `pluginDefinitionLoader` are reduced to a bare minimum where the components at hand (definition classpath, dependency classpath) are properly defined.

Documentation to the three main methods has been added.

### DIFF
--- a/main/src/main/scala/sbt/internal/Load.scala
+++ b/main/src/main/scala/sbt/internal/Load.scala
@@ -34,7 +34,7 @@ import Keys.{
   thisProjectRef,
   update
 }
-import tools.nsc.reporters.ConsoleReporter
+import scala.tools.nsc.reporters.ConsoleReporter
 import sbt.internal.util.{ Attributed, Settings, ~> }
 import sbt.util.{ Eval => Ev, Show }
 import sbt.internal.util.Attributed.data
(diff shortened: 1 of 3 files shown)


jvican:
I have repurposed this PR to improve `Load.scala`, which is my incipient work to try to improve plugin classloading.

main/src/main/scala/sbt/internal/Load.scala
```diff
@@ -1012,21 +1014,21 @@ final case class LoadBuildConfiguration(
...
+  lazy val globalPluginClasspath: Def.Classpath =
+    Load.buildPluginClasspath(this, Load.globalPluginClasspath(globalPlugin))
+
+  lazy val detectedGlobalPlugins: DetectedPlugins = {
+    val pluginData: PluginData = {
+      globalPlugin.map { info =>
+        val data = info.data
+        PluginData(data.fullClasspath, data.internalClasspath,
+          Some(data.resolvers), Some(data.updateReport), Nil)
+      }.getOrElse(PluginData(globalPluginClasspath))
```
eed3si9n, line 1087: I generally write out pattern matching for `Option`. It's the fastest and most readable:

> Not only is it by far the fastest (probably at least 10 times as fast as the alternatives), it is also the clearest.

 according to [Odersky](https://www.reddit.com/r/scala/comments/449td4/fold_or_pattern_matching_with_an_option/)

main/src/main/scala/sbt/internal/Load.scala
```diff
@@ -1012,21 +1014,21 @@ final case class LoadBuildConfiguration(
...
     }
-    val baseDir = globalPlugin match {
-      case Some(x) => x.base
-      case _       => stagingDirectory
-    }
-    Load.loadPluginDefinition(baseDir, this, pluginData)
+    val baseDir = globalPlugin.map(_.base).getOrElse(stagingDirectory)
```
eed3si9n: Same here.

eed3si9n (changes requested):
Style comment.

eed3si9n:
closing this in favor of https://github.com/sbt/sbt/pull/3130

Issue #2994, opened 2017-03-06
Consider making sbt less verbose


jvican:
Current sbt is extremely verbose. Messages could be improved to be:
  
* More concrete. Straight to the point.
* Shorter. This also includes removing unnecessary messages.
* Grouped by topic (perhaps with progress bars).
  
I'm going to illustrate my points with examples.
  
Very short extract of `publishLocal` in `sbt/sbt`:
```

[info] will publish locally with javac version 1.8.0_92
[info] will publish locally with javac version 1.8.0_92
[info] :: delivering :: org.scala-sbt#test-agent;0.13.14-SNAPSHOT :: 0.13.14-SNAPSHOT :: integration :: Mon Mar 06 15:03:11 CET 2017
[info] 	delivering ivy file to /data/rw/code/scala/sbt/testing/agent/target/ivy-0.13.14-SNAPSHOT.xml
[info] :: delivering :: org.scala-sbt#collections;0.13.14-SNAPSHOT :: 0.13.14-SNAPSHOT :: integration :: Mon Mar 06 15:03:11 CET 2017
[info] 	delivering ivy file to /data/rw/code/scala/sbt/util/collection/target/ivy-0.13.14-SNAPSHOT.xml
[warn] bootstrap class path not set in conjunction with -source 1.6
[info] will publish locally with javac version 1.8.0_92
[info] :: delivering :: org.scala-sbt#cross;0.13.14-SNAPSHOT :: 0.13.14-SNAPSHOT :: integration :: Mon Mar 06 15:03:12 CET 2017
[info] 	delivering ivy file to /data/rw/code/scala/sbt/util/cross/target/ivy-0.13.14-SNAPSHOT.xml
[info] will publish locally with javac version 1.8.0_92
[info] :: delivering :: org.scala-sbt#sbt-launch;0.13.14-SNAPSHOT :: 0.13.14-SNAPSHOT :: integration :: Mon Mar 06 15:03:12 CET 2017
[info] 	delivering ivy file to /data/rw/code/scala/sbt/launch/target/ivy-0.13.14-SNAPSHOT.xml
[info] Packaging /data/rw/code/scala/sbt/util/collection/target/collections-0.13.14-SNAPSHOT.jar ...
[info] Done packaging.
[info] will publish locally with javac version 1.8.0_92
[...]

PR #2747, opened 2016-09-16
[sbt 1.0] Quieter ivyLoggingLevel for CI

eed3si9n:
`ivyLoggingLevel` is dropped to `UpdateLogging.Quiet` when Jenkins or
Travis is detected.

/review @dwijnand, @Duhemm

PR #3122, opened 2017-04-19
Change "Updating" logging from info to debug

japgolly:
The noise:value ratio of "Updating"/"Done updating" logging is very high.
This changes the log level to debug.

This would result in:

```
[info] Updating {file:/…/scalajs-react/}root...
[info] Updating {file:/…/scalajs-react/}core...
[info] Updating {file:/…/scalajs-react/}gh-pages-macros...
[info] Done updating.
[info] Compiling 2 Scala sources to /…/scalajs-react/gh-pages-macros/…
[info] Done updating.
[info] Updating {file:/…/scalajs-react/}extra...
[info] Done updating.
[info] Done updating.
[info] Updating {file:/…/scalajs-react/}cats...
[info] Updating {file:/…/scalajs-react/}scalaz72...
[info] Compiling 60 Scala sources to /…/scalajs-react/core/…
[info] Done updating.
[info] Done updating.
[info] Updating {file:/…/scalajs-react/}monocle...
[info] Done updating.
[info] Updating {file:/…/scalajs-react/}test...
[info] Updating {file:/…/scalajs-react/}gh-pages...
[info] Done updating.
[info] Done updating.
[info] Compiling 25 Scala sources to /…/scalajs-react/extra/…
[info] Compiling 4 Scala sources to /…/scalajs-react/cats/…
[info] Compiling 4 Scala sources to /…/scalajs-react/scalaz-7.2/…
[info] Compiling 10 Scala sources to /…/scalajs-react/test/…
[info] Compiling 4 Scala sources to /…/scalajs-react/monocle/…
[info] Compiling 23 Scala sources to /…/scalajs-react/gh-pages/…
[info] Compiling 39 Scala sources to /…/scalajs-react/test/…
```

becoming:

```
[info] Compiling 2 Scala sources to /…/scalajs-react/gh-pages-macros/…
[info] Compiling 60 Scala sources to /…/scalajs-react/core/…
[info] Compiling 25 Scala sources to /…/scalajs-react/extra/…
[info] Compiling 4 Scala sources to /…/scalajs-react/cats/…
[info] Compiling 4 Scala sources to /…/scalajs-react/scalaz-7.2/…
[info] Compiling 10 Scala sources to /…/scalajs-react/test/…
[info] Compiling 4 Scala sources to /…/scalajs-react/monocle/…
[info] Compiling 23 Scala sources to /…/scalajs-react/gh-pages/…
[info] Compiling 39 Scala sources to /…/scalajs-react/test/…
```

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1484,14 +1484,14 @@ object Classpaths {
       def work = (_: In) match {
         case conf :+: settings :+: config :+: HNil =>
           import ShowLines._
-          log.info("Updating " + label + "...")
+          log.debug("Updating " + label + "...")
           val r = IvyActions.updateEither(module, config, uwConfig, logicalClock, depDir, log) match {
             case Right(ur) => ur
             case Left(uw) =>
               uw.lines foreach { log.warn(_) }
               throw uw.resolveException
           }
-          log.info("Done updating.")
+          log.debug("Done updating.")
           val result = transform(r)
           val ew = EvictionWarning(module, ewo, result, log)
           ew.lines foreach { log.warn(_) }


japgolly:
Addresses a subset of #2994

jvican:
This is great, thanks @japgolly!
  
P.S. This is targeting 0.13 -- I think there will be no more 0.13.x releases. Can you rebase it on top of 1.0? We're planning to release it soon.

eed3si9n:
First, the "Updating" log is inside `def work`, so that won't get displayed every time sometime types in `compile`, correct?

When in fact sbt is going to update the dependency, I think it would make sense to write it out. Perhaps not both but at least the first one since on a larger project it might take long time. Also note that on sbt 1.0 "Resolving .." log entry will no longer display on CI environments (https://github.com/sbt/sbt/pull/2747).

japgolly:
@jvican No worries, if the change is approved in spirit I'll rebase.

@eed3si9n It gets displayed every time after a `clean`. (Which makes it feel like it gets displayed all the time.)

I really don't think it makes sense to print out at info. Info means most people are interested and/or care, or that it's providing info that's useful to the user. When SBT downloads dependencies, it logs that it's going to download those at info which is fine, and separate and I'm not affecting that. When updating takes a long time it's either assuaged by the "Resolving …" / "downloading …" info-level messages, or not at all (including by "Updating …" messages). "Updating x" makes no sense to anyone that doesn't have deep SBT knowledge. What does it mean? Checking that all dependencies are available locally? I don't know. "Update" doesn't communicate what it does. I'm talking in circles now. Need :coffee: :grinning:

japgolly:
Oh another thing, if "updating" just means downloading/locating all of dependencies, if they're all static (no SNAPSHOTs) why should it run again after a clean? Sounds like a waste of time. :grimacing:

jvican:
> Oh another thing, if "updating" just means downloading/locating all of dependencies, if they're all static (no SNAPSHOTs) why should it run again after a clean? Sounds like a waste of time. 😬
  
I agree on this somehow, and we've already had this discussion in Gitter. Have a look at this chat and scroll down to see how the conversation develops.
https://gitter.im/sbt/sbt-contrib?at=58d7ed3af22385553dda876b
  
In my opinion, update should cache more and we should change its meaning or improve the current implementation. Every time we change the scala version, the unnecessary dependencies get resolved/updated again. This may go with the dependency lock file https://github.com/sbt/sbt/issues/2989.
  
@japgolly To me, dependencies update is a general build tool concept and users may expect it to be printed out to assert that an update is indeed happening. What I'm super sure is that "Done updating" should be removed.
  
Perhaps we can also get rid of the "updating" log under a special sbt setting that is enabled per project. Either that, or we wait unti lthe dependency lock file that will not trigger dependencies update.
  
Thanks again for opening this @japgolly.

jvican:
I've thought about this and I've seen that `Updating` is shown even if no resolution takes place. This doesn't make sense because you don't update if you don't resolve again, making the whole message unnecessary.
  
Given these circumstances, I think that removing it is not a bad idea after all, **EDIT** though I think a similar message as `Updating` should appear before the results of resolution just to signal for which project sbt is resolving dependencies.

japgolly:
Yeah I was thinking the same thing. I think there are two cases:

1. It says "Updating" and then starts actually updating dependencies in which case you start seeing messages starting with "Resolving."

2. It says "Updating", determines there's nothing to do and finishes nearly immediately.

If we remove both messages as per this PR then in the first case, the users will see "Resolving" messages which is sufficient, and in the second they see nothing which is perfect.

japgolly:
> I think a similar message as Updating should appear before the results of resolution just to signal for which project sbt is resolving dependencies.

When you say "which project", this doesn't help answer that. In a multi-module project all the modules generally run in parallel and the console gets spammed in non-deterministic order. You'll likely see "Updating {A,B,C}", "Done updating" twice and then "Resolving." Still don't know if A, B or C :smile:

jvican:
Can someone restart the failing instance (not the whole build)? https://travis-ci.org/sbt/sbt/jobs/223465131
  
@japgolly @eed3si9n I think this is a useful change. Are there more comments on this PR or should we merge this?

jvican:
@japgolly Now that you're at this, can you also get rid of the "Done packaging" logs?

japgolly:
Sorry @jvican, I'm not sure what you mean. Do you want me to delete the "Done packaging" line instead of making it debug-level? I'm happy to make any change you need, that's not a problem...  But I'm not sure what the story is with this issue. My understanding is we need @eed3si9n 's blessing to proceed (?). @eed3si9n marked it as needs discussion 17 days ago but hasn't been involved in any discussion since. Does this still need further discussion? Or is this approved? I'm not sure what we're still waiting for. :confused:

eed3si9n:
In general I don't think it's a high priority change because if you run `clean`, your task cache is blown out, so you should expect more log messages. Since dependency resolution is one of long-running tasks, I believe that these log entries were added over the course of code history to address specific concerns like "sbt looks frozen for 4 minutes, what's going on?"

It would be nice if sbt didn't do that if you feed a huge graph, but the fact of the matter is that it does. And with a build with dozens of subprojects having `Updating {file:/…}foo...` helps to identify which subproject is currently resolving.

japgolly:
I'm sorry but that sounds like another *"dont wanna!"* rather than a thoughtful or technical response. Or is it me? Am I on the fringe here? Has there been something different about every multi-module project I've used, in contrast to the usual or what you normally work with, @eed3si9n? My builds seem to look quite similar to most I see in OSS so I'm really confused as to why we seem to have such different experiences.  Anyway,

> if you run clean, your task cache is blown out, so you should expect more log messages.

That's not logically entailed. If I type `compile` I expect to see `Compiling ...`. If I type `clean compile` I might expect to see `Cleaning` but then next I expect to see `Compiling ...`. Understanding SBT's internals doesn't mean that it's ideal nor externally-expected behaviour.

> Since dependency resolution is one of long-running tasks, I believe that these log entries were added over the course of code history to address specific concerns like "sbt looks frozen for 4 minutes, what's going on?"

When there's nothing to resolve, it's not a long-running task.
When there are dependencies to resolve, there are other log messages that indicate progress.
(Btw if resolving dependencies is the only thing that happens during "updating" we should just change the output to "Resolving dependencies..." instead of "Updating...")

> with a build with dozens of subprojects having Updating {file:/…}foo... helps to identify which subproject is currently resolving.

Like I said, in a non-trivially sized project, they all run in parallel and then finish anonymously. One cannot look at those logs to determine which modules have and have not completed.

> I don't think it's a high priority change

As far as triage goes, no it's not high priority, I agree. But I fail to see the relevance. Is SBT only accepting high priority changes? I'm sorry to say this but SBT has poor UX and this tiny patch was a test to see if collaboration with the SBT team has improved enough that I might start making more active contributions to incrementally improve SBT's UX. The collaboration experience has *not* improved. It's been drawn-out. It's frustrating. My enthusiasm is completely gone. Regardless of any outcome henceforth I do not see myself attempting to contribute again.

olafurpg:
I did a small experiment to test the hypothesis that "Updating X; Done updating" logging statements are helpful on a clean build with clean cache when resolution might take time.

I opened a PR in scalameta/scalameta disabling coursier so that we would use the default ivy resolver on a clean cache

- PR: https://github.com/scalameta/scalameta/pull/856
- CI logs: https://platform-ci.scala-lang.org/scalameta/scalameta/253/1

A good way to analyze the logs is to search (Ctrl+F) for "Updating" and see how many seconds pass between the "Updating" logs and the "Resolving" logs.

My observations
- 14 crossProject modules and 3 project
- not as many dependencies as in a typical Play! app, but I counted 49 "[SUCCESSFUL ]" lines from the ivy resolver. 
- 34 "Updating {" lines of which it seems all of them are followed by a "Resolving" lines within 1 second.
  - Note. The CI server has a good internet connection so the time between "Updating" and "Resolving" may be longer on other machines.
- Lines 406 - 429 are a series of "Updating/Done updating" lines

My conclusion: the "Updating/Done updating" logs are noisy and users should not experience "sbt looks frozen for 4 minutes" if the log statements get moved to debug from info. I approve this PR 👍

eed3si9n:
Thanks @olafurpg for the data.

If we start from the compilation of the actual subprojects, the build gets loaded around line 272 (44s):

```
[info] Set current project to scalametaRoot (in build file:/drone/src/github.com/scalameta/scalameta/)
```

Because `compile` depends on having the `update` completed first, we see `[info] Updating {file:/drone/src/github.com/scalameta/scalameta/}commonJS...` right away, letting us know which subproject is being worked on. Let's suppose we don't have that. What the user would see is:

```
[info] Resolving org.scala-lang#scala-compiler;2.11.11 ...
```

Given that since sbt overwrites the "Resolving .." on the same line logs basically the user will see rotating series of "Resolving ..." with occasional "downloading ..." scrolling by.

Around line 412 (56s) we would see the _first_ subproject name:

```
[info] Compiling 55 Scala sources to /drone/src/github.com/scalameta/scalameta/scalameta/common/js/target/scala-2.11/classes...
```

That's 12s of "Resolving.." but some builds have it worse. More "Resolving ..." with other compilation continues till around 73s. This demonstrates my point that if we didn't have `Updating {file:/drone/src/github.com/scalameta/scalameta/}commonJS...`, on a larger projects and running a lot of CPU cores, you might see minutes of "Resolving" without knowing what subproject you're updating.

I would suggest tying this to `ivyLoggingLevel` setting such that when it's set to `Quiet` it won't display "Updating" too or put this behind a flag. I honestly don't have any attachment on these log entries, but I can imagine that when I remove them, I get scolded by some other sets of programmers who say that sbt's UX degraded.

japgolly:
> This demonstrates my point that if we didn't have Updating {file:/drone/src/github.com/scalameta/scalameta/}commonJS..., on a larger projects and running a lot of CPU cores, you might see minutes of "Resolving" without knowing what subproject you're updating.

This is false. The snippet you referred to doesn't demonstrate your point, rather it demonstrates mine. Here are lines [274,277]

```
[info] Updating {file:/drone/src/github.com/scalameta/scalameta/}commonJS...
[info] Updating {file:/drone/src/github.com/scalameta/scalameta/}commonJVM...
[info] Resolving org.scala-lang#scala-compiler;2.11.11 ...

[info] Resolving org.scala-lang#scala-library;2.11.11 ...
```
Which subproject is causing the resolving?

*(And what's with those deliberate, silly newlines? Given the way this PR has gone I won't even start on that. Forgive me, I digress...)*

How about lines [458,465]?
```
[info] Updating {file:/drone/src/github.com/scalameta/scalameta/}transversersJS...
[info] Updating {file:/drone/src/github.com/scalameta/scalameta/}parsersJS...
[info] Updating {file:/drone/src/github.com/scalameta/scalameta/}semanticJVM...
[info] Updating {file:/drone/src/github.com/scalameta/scalameta/}quasiquotesJVM...
[info] Done updating.
[info] Done updating.
[info] Resolving com.trueaccord.scalapb#scalapb-runtime_2.11;0.6.0-pre2 ...

[info] Resolving org.scala-lang#scala-library;2.11.8 ...
```
Which subproject is causing the resolving?

It's orthogonal to this issue but if being able to tie resolution to a subproject is so important then what SBT should be logging is something similar to:
```
[info] [transversersJS] Resolving com.trueaccord.scalapb#scalapb-runtime_2.11;0.6.0-pre2 ...
[info] [semanticJVM] Resolving org.scala-lang#scala-library;2.11.8 for ...
```
(Which makes you think: maybe *all* logging in a multi-module build should be prefixed by the subproject name.)

jvican:
@japgolly I really like your proposed format for resolution logs: `Resolving org.scala-lang#scala-library;2.11.8 for XXXX` where `XXXX` is the project name. I feel that the project we're resolving for should be present in that line too. This way, we feed two birds with one scone: we keep the same conciseness than we have now by removing two frequent and unnecessary log lines. The place to modify here is here: https://github.com/sbt/librarymanagement/blob/1.0/librarymanagement/src/main/scala/sbt/internal/librarymanagement/ivyint/SbtChainResolver.scala#L80. The code to detect that log pattern and delete the last line is in sbt/sbt, though it's not relevant for this change.
  
Regarding your second suggestion, I agree sbt should prepend the name of the project before every log. The project name at the beginning of the line helps people identify which project's task is running. This is good for UX and debuggability.
  
As a matter of fact, other build tools like pants do so for every task name, thus showing the whole pipeline. I think sbt should get inspired from that approach and only do it for task names. Clarity is an important feature.
  
Would you mind opening a ticket about this @japgolly? I am happy to implement this feature.

eed3si9n:
If we are hiding logs to debug level, perhaps the ones that doesn't carry its weight is "Resolving..." log. We already started hiding it for CI (#2747), plus it exposes internal behavior of Ivy that user can't really do much about because it overwrites too quickly.

I'd be happy to discuss prefixing subproject ids on all log entries, probably in another issue.

jvican:
I think the "Resolving..." log carries meaningful information. People can know how much time does a concrete module take to resolve. It's useful in CI too, since lines are mapped to timestamps that show how much time happened between every log.
  
>  because it overwrites too quickly.
  
This is not always the case, unfortunately. That's why it's necessary.

japgolly:
Yeah no worries, just I'm a bit busy atm, let me get back to you shortly,
probably 2 days

japgolly:
Hello all, sorry I'm a bit late -- I was severely sick and in bed nearly all of last week.

So what's the outcome this PR? Do we have consensus? Can it be merged as it? Does it need changes?

@jvican What is the issue you'd like me to raise? You mean prefixing logs by module name, right?

jvican:
Yes @japgolly :smile:. Something like: "Consider prepending module name to every log". This would not only help UX but make obvious blatant errors with scoping.
  
> So what's the outcome this PR? Do we have consensus? Can it be merged as it? Does it need changes?
  
I think this is good to go when we have the prefix in all the logs. I think all Eugene's point boiled down to that. But I'll let him address your question.
  
N.B. I'm all in for this change. Thanks, @japgolly.

jvican:
@eed3si9n @dwijnand Can we revive this discussion and make it actionable, please?
  
I think @japgolly has made the case that `Done compiling` and `Done updating` and `Done whatever` does not bring anything to users when tasks are completed concurrently and logs are not ordered.
  
I think that the logs linked in the CI also show that there's no value in saying `Updating`, what we really want to see is what sbt does `Resolving`, which is the time-consuming task.
  
The discussion and data in this PR leads me to believe this is the best we can do here:
  
* Add the project prefix to the `Resolving` logs.
* Remove the `Updating` logs.
* Put all the "Done" bla changes under the debug log level, and prepend the project name.

eed3si9n:
@jvican In sbt 1.0 "Resolving" log is no longer displayed per https://github.com/sbt/librarymanagement/pull/51.

As previously stated:
1. "Updating .." is only displayed only when lm is doing actual dependency resolution, not cached. 
2. The intended audience of "Updating.." is CI users on fresh compile with potentially long resolution time. Without "Updating" they won't see any output on screen until the compilation starts.

Before jumping to the solution space, let's try to identify a problem statement first, and then brainstorm ways to tackle it.

japgolly:
> Before jumping to the solution space, let's try to identify a problem statement first, and then brainstorm ways to tackle it.

I thought we've already identified the problems. Let's get on with it.

The problems:

* Users get blasted with a lot of noise (see first comment here). I don't want to see it anymore
* Updating gets printed for everything after a clean, even though everything is resolved and non-snapshots
* In a CI it's useful to have some kind of info on the screen, so users know something is happening and presumably so that the CI VM doesn't terminate
* Most "Updating" start and finish in < 100 ms meaning it's useless noise and doesn't indicate that any long process is happening

Solution Idea:

How about we set Updating to debug unless running in CI env.

eed3si9n:
@japgolly Thanks for restating your problem statements. I think this is helpful because without starting with the common precondition, we won't agree on the solution.

### steps

1. `clean` all or many subprojects.
2. `compile` all or many subprojects.

### problems

> - Users get blasted with a lot of noise (see first comment here). I don't want to see it anymore.

I think it's a legitimate concern, but we'll come back to this one.

> - Updating gets printed for everything after a clean, even though everything is resolved and non-snapshots

Calling `clean` wipes out the task cache for `update` task. So even though the artifacts may be downloaded, the dependency graph is re-resolved using Ivy. This could take 30s or more to complete.

> - In a CI it's useful to have some kind of info on the screen, so users know something is happening and presumably so that the CI VM doesn't terminate

"something is happening" info is useful for both CI and non-CI.

> - Most "Updating" start and finish in < 100 ms meaning it's useless noise and doesn't indicate that any long process is happening

This really depends on how complex the dependency graph is.

### clean situation

> - Users get blasted with a lot of noise (see first comment here). I don't want to see it anymore.

The question here is, how much should we accommodate the use of `clean` of all subprojects?
Not much, is my leaning. sbt is at its core an interactive, incremental build tool, so if there are some more logs when someone runs `clean`, I think it's not that bad.

Also about the amount of logs. If you have 10 subprojects, you get 10 x 2 entries.
I don't think it's that bad (compared e.g., to "Resolving ..." filling up Travis CI log adding hundreds of lines of noise).

### "Updating" as debug level

> - "How about we set Updating to debug unless running in CI env."

- Pro: You don't see anything by default.
- Con: You don't see anything by default, when it's resolving things for 30s.

### solution 2: customization

Another solution I propose here is customized setting that you can try.
In your `~/.sbt/1.0/updateLog.sbt` (or `~/.sbt/0.13/updateLog.sbt`) write:

```scala
logLevel in Global in update := Level.Warn
```

That will remove the "Updating ..." log on all builds on your machine.
If you want, you could make it into a plugin and let others try it as well.

eed3si9n:
Closing this, and opening https://github.com/sbt/sbt/issues/4362.

PR #3125, opened 2017-04-19
[sbt 1.0] Reformat using Scalafmt

### LINKED ISSUE #2944: clean up source code

eed3si9n:
Fixes #2944

Taking recommendation by @olafurpg in #3003:

> - maxColumn 100 is totally fine.
> - don't use vertical alignment, it's distracting.
> - use default or I'm happy to discuss adding style=sbt, sbt is a large enough project to merit its own style.

The 4 tweakings I took from original @dwijnand PRs with my added comments:

```yaml

# http://docs.scala-lang.org/style/scaladoc.html recommends the JavaDoc style.
# scala/scala is written that way too https://github.com/scala/scala/blob/v2.12.2/src/library/scala/Predef.scala
docstrings = JavaDoc

# This also seems more idiomatic to include whitespace in import x.{ yyy }
spaces.inImportCurlyBraces = true

# This works around sequence wildcard (`_*`) turning into `_ *`
spaces.beforeSeqWildcard = true

# Vertical alignment only => for pattern matching
align.tokens.add = [
  { code = "=>",  owner = "Case"  }
]
```

Note this adds automatic formatting on `compile` task using `AutomateScalafmtPlugin` implemented by @hseeberger, and CI checking.


main-actions/src/main/scala/sbt/ForkTests.scala
```diff
@@ -93,12 +112,20 @@ private[sbt] object ForkTests {
...
+                          classOf[ForkMain].getCanonicalName,
+                          server.getLocalPort.toString)
         val ec = Fork.java(fork, options)
         val result =
           if (ec != 0)
-            TestOutput(TestResult.Error, Map("Running java with options " + options.mkString(" ") + " failed with exit code " + ec -> SuiteResult.Error), Iterable.empty)
+            TestOutput(TestResult.Error,
+                       Map(
+                         "Running java with options " + options
+                           .mkString(" ") + " failed with exit code " + ec -> SuiteResult.Error),
```
smarter, line 127: This is a pretty weird formatting.

jvican:
Can we please merge this after my PR is merged? Rebasing it on top of this will be very difficult and I prefer to spend that time on other things.

main-actions/src/main/scala/sbt/Sync.scala
```diff
@@ -26,42 +26,44 @@ import sjsonnew.{ Builder, JsonFormat, Unbuilder, deserializationError }
...
 
   def copy(source: File, target: File): Unit =
     if (source.isFile)
       IO.copyFile(source, target, true)
     else if (!target.exists) // we don't want to update the last modified time of an existing directory
-    {
-      IO.createDirectory(target)
-      IO.copyLastModified(source, target)
-    }
+      {
```
smarter, line 63: This is also weird, you want `else if (...) {` on a single line, not indenting the block as if it was a normal expression.

main-actions/src/test/scala/sbt/compiler/EvalTest.scala
```diff
@@ -62,35 +63,38 @@ val p = {
...
+    testImport("import annotation._, math._, meta._" :: Nil))
+  property("multiple imports") = forAll(
+    testImport("import annotation._" :: "import math._" :: "import meta._" :: Nil))
 
-  private[this] def testImport(imports: Seq[String]): Int => Prop = i =>
-    value(eval.eval("abs(" + i + ")", new EvalImports(imports.zipWithIndex, "imp"))) == math.abs(i)
+  private[this] def testImport(imports: Seq[String]): Int => Prop =
+    i =>
+      value(eval.eval("abs(" + i + ")", new EvalImports(imports.zipWithIndex, "imp"))) == math.abs(
+        i)
```
smarter, line 81: Also unlike any kind of style a human would use.

smarter:
@olafurpg Does scalafmt take operator precedence into account when deciding where to put newlines?

main-command/src/main/scala/sbt/State.scala
```diff
@@ -109,17 +118,22 @@ trait StateOps {
...
+
   /** Sets the value associated with `key` in the custom attributes map.*/
   def put[T](key: AttributeKey[T], value: T): State
+
   /** Removes the `key` and any associated value from the custom attributes map.*/
   def remove(key: AttributeKey[_]): State
+
   /** Sets the value associated with `key` in the custom attributes map by transforming the current value.*/
   def update[T](key: AttributeKey[T])(f: Option[T] => T): State
+
```
dwijnand, line 136: holy shit when did scalafmt start doing this newline inserting? <3 @olafurpg

main-settings/src/main/scala/sbt/Structure.scala
```diff
@@ -294,138 +473,335 @@ object Scoped {
...
-  @deprecated("The sbt 0.10 style DSL is deprecated: '(k1, k2) map { (x, y) => ... }' should now be '{ val x = k1.value; val y = k2.value }'.\nSee http://www.scala-sbt.org/0.13/docs/Migrating-from-sbt-012x.html", "0.13.13") implicit def t11ToApp11[A, B, C, D, E, F, G, H, I, J, K](t11: (Initialize[A], Initialize[B], Initialize[C], Initialize[D], Initialize[E], Initialize[F], Initialize[G], Initialize[H], Initialize[I], Initialize[J], Initialize[K])): Apply11[A, B, C, D, E, F, G, H, I, J, K] = new Apply11(t11)
+  @deprecated(
+    "The sbt 0.10 style DSL is deprecated: '(k1, k2) map { (x, y) => ... }' should now be '{ val x = k1.value; val y = k2.value }'.\nSee http://www.scala-sbt.org/0.13/docs/Migrating-from-sbt-012x.html",
+    "0.13.13"
+  ) implicit def t2ToApp2[A, B](t2: (Initialize[A], Initialize[B])): Apply2[A, B] = new Apply2(t2)
+  @deprecated(
+    "The sbt 0.10 style DSL is deprecated: '(k1, k2) map { (x, y) => ... }' should now be '{ val x = k1.value; val y = k2.value }'.\nSee http://www.scala-sbt.org/0.13/docs/Migrating-from-sbt-012x.html",
+    "0.13.13"
```
dwijnand: jesus christ. We should exclude this section and the same one above. They're exceptions.

main-actions/src/main/scala/sbt/Package.scala
```diff
@@ -41,34 +40,38 @@ object Package {
...
     val main = manifest.getMainAttributes
     for (option <- conf.options) {
       option match {
-        case JarManifest(mergeManifest)          => mergeManifests(manifest, mergeManifest)
-        case MainClass(mainClassName)            => main.put(Attributes.Name.MAIN_CLASS, mainClassName)
-        case ManifestAttributes(attributes @ _*) => main.asScala ++= attributes
-        case _                                   => log.warn("Ignored unknown package option " + option)
+        case JarManifest(mergeManifest) => mergeManifests(manifest, mergeManifest)
+        case MainClass(mainClassName) => main.put(Attributes.Name.MAIN_CLASS, mainClassName)
+        case ManifestAttributes(attributes @ _ *) => main.asScala ++= attributes
```
eed3si9n: Changing `_*` to `_ *` looks odd.  

SLS says "The last pattern in a pattern sequence may be a sequence wildcard _*." - https://www.scala-lang.org/files/archive/spec/2.12/08-pattern-matching.html#pattern-sequences

tpolecat:
The lack of blank lines between definitions is among my top complaints and this doesn't seem to affect it. See [Eval.scala](https://github.com/eed3si9n/sbt/blob/499bba8142a988ffec713bd9db0db1c117c97dd7/main-actions/src/main/scala/sbt/compiler/Eval.scala) for example. Is there an easy way to improve that?

Otherwise the formatting isn't really my taste but it's consistent and it's an improvement so 👍 from me. I'm kind of ambivalent about enforcing it as part of the build vs. just doing it once and calling it good. No strong opinion.

main/src/main/scala/sbt/Keys.scala
```diff
@@ -78,174 +85,511 @@ import org.apache.logging.log4j.core.Appender
...
-  val timingFormat = SettingKey[java.text.DateFormat]("timing-format", "The format used for displaying the completion time.", CSetting)
-  val extraLoggers = SettingKey[ScopedKey[_] => Seq[Appender]]("extra-loggers", "A function that provides additional loggers for a given setting.", DSetting)
-  val logManager = SettingKey[LogManager]("log-manager", "The log manager, which creates Loggers for different contexts.", DSetting)
-  val logBuffered = SettingKey[Boolean]("log-buffered", "True if logging should be buffered until work completes.", CSetting)
-  val sLog = SettingKey[Logger]("setting-logger", "Logger usable by settings during project loading.", CSetting)
+  val logLevel =
+    SettingKey[Level.Value]("log-level", "The amount of logging sent to the screen.", ASetting)
+  val persistLogLevel =
+    SettingKey[Level.Value]("persist-log-level",
+                            "The amount of logging sent to a file for persistence.",
```
dwijnand: uh... do we want to change this like this? I know the former is longer than 100 columns, but I thinks it's about 1000x more readable. We should exclude these blocks, maybe the whole file.

main/src/main/scala/sbt/Main.scala
```diff
@@ -160,34 +203,79 @@ import TemplateCommandUtil.templateCommand
...
+
+  def DefaultCommands: Seq[Command] =
+    Seq(
+      ignore,
+      help,
+      completionsCommand,
+      about,
+      tasks,
+      settingsCommand,
+      loadProject,
```
dwijnand, line 219: these binpacked would be much nicer..

.scalafmt.conf
```diff
@@ -0,0 +1,10 @@
+maxColumn = 100
+project.git = true
+project.excludeFilters = [ /sbt-test/, /input_sources/ ]
```
dwijnand: we should exclude `/contraband-scala/` too otherwise scalafmt and contraband are going to keep fighting each other

main-actions/src/main/scala/sbt/Package.scala
dwijnand: It's why I'd set `spaces.beforeSeqWildcard = true`. I've no idea why the default is false. scalameta/scalafmt#772

main-actions/src/main/scala/sbt/Package.scala
```diff
@@ -41,34 +40,38 @@ object Package {
...
     val manifest = new Manifest
     val main = manifest.getMainAttributes
     for (option <- conf.options) {
       option match {
-        case JarManifest(mergeManifest)          => mergeManifests(manifest, mergeManifest)
-        case MainClass(mainClassName)            => main.put(Attributes.Name.MAIN_CLASS, mainClassName)
-        case ManifestAttributes(attributes @ _*) => main.asScala ++= attributes
-        case _                                   => log.warn("Ignored unknown package option " + option)
+        case JarManifest(mergeManifest) => mergeManifests(manifest, mergeManifest)
+        case MainClass(mainClassName) => main.put(Attributes.Name.MAIN_CLASS, mainClassName)
```
dwijnand: Can we at least agree that `=>`s should be aligned? I find the former state to be so much more legible..

project/Formatting.scala
```diff
@@ -1,40 +0,0 @@
...
-      includeFilter in (BuildConfig, scalariformFormat) := ("*.scala": FileFilter),
-      includeFilter in (BuildSbtConfig, scalariformFormat) := ("*.sbt": FileFilter),
-      scalariformFormat in Compile := {
-        val x = (scalariformFormat in BuildSbtConfig).value
-        val y = (scalariformFormat in BuildConfig).value
-        (scalariformFormat in Compile).value
-      },
-      scalariformCheck := {
-        val diff = "git diff".!!
-        if (diff.nonEmpty) sys.error("Working directory is dirty!\n" + diff)
```
dwijnand, line 37: We should continue to enforce the format with CI.

dwijnand (changes requested):
Thank you for the PR, @eed3si9n. Looks generally good, I've left a few comments to discuss.

main/src/main/scala/sbt/Keys.scala
eed3si9n: Maybe we need a Perl script to lift the key description such as `"The amount of logging sent to a file for persistence."` into a variable name right above the `val`.

main-actions/src/main/scala/sbt/Package.scala
eed3si9n: Sure why not.

main-settings/src/main/scala/sbt/Structure.scala
```diff
@@ -264,27 +336,125 @@ object Scoped {
   }
   implicit def richAnyTaskSeq(in: Seq[AnyInitTask]): RichAnyTaskSeq = new RichAnyTaskSeq(in)
   final class RichAnyTaskSeq(keys: Seq[AnyInitTask]) {
-    def dependOn: Initialize[Task[Unit]] = Initialize.joinAny[Task](keys).apply(deps => nop.dependsOn(deps: _*))
+    def dependOn: Initialize[Task[Unit]] =
+      Initialize.joinAny[Task](keys).apply(deps => nop.dependsOn(deps: _*))
   }
 
   // this is the least painful arrangement I came up with
```
dwijnand: The way to exclude this block is by surrounding it with

```
// format: off
```

and then

```
// format: on
```

same for the t2ToApp2, etc block below.

I can do it in a follow up PR if you'd prefer.

dwijnand (approved):

main-actions/src/main/scala/sbt/ForkTests.scala
olafurpg: Yes, I agree. Scalafmt leaves respects line breaks around infix operators. If you put a newline between `+` and `options` the formatting would look nicer.

main-actions/src/test/scala/sbt/compiler/EvalTest.scala
olafurpg: This is a known issue https://github.com/scalameta/scalafmt/issues/667 I suspect it would look nicer wrapped in curly braces.

main-command/src/main/scala/sbt/State.scala
olafurpg: Scalafmt inserts a blank line if there's a docstring :))

main-actions/src/main/scala/sbt/Package.scala
olafurpg: I've opened https://github.com/scalameta/scalafmt/issues/890

olafurpg:
@tpolecat There is a `newlines.alwaysBeforeTopLevelStatements` setting http://scalameta.org/scalafmt/#newlines.alwaysBeforeTopLevelStatements which was contributed by @Daxten to enforce blank lines between definitions. Currently, scalafmt only enforces a blank before docstrings.

main-actions/src/main/scala/sbt/Package.scala
olafurpg: FWIW, the Scala.js coding style permits vertical alignment in pattern matching only.

You can limit vertical alignment to `=>` in pattern matching only with 

```
align.tokens = [{owner="Case", code="=>"}]
```

That will avoid unwanted alignment for lambda arrows, import rename arrows etc.

main-actions/src/test/scala/sbt/compiler/EvalTest.scala
olafurpg: Wrapping it in curly braces and inserting a newline after `==` yields this output

```scala
  private[this] def testImport(imports: Seq[String]): Int => Prop = { i =>
    value(eval.eval("abs(" + i + ")", new EvalImports(imports.zipWithIndex, "imp"))) ==
      math.abs(i)
  }
```

I tried really hard to come up with a better solution for infix operators but it always came out badly in some weird DSLs in the wild. Until scalafmt is able to handle infix operators well in all cases I thought it was best to leave those line break choices to the user.

olafurpg:
First of all, I'm really happy to see this PR 😄 A few notes

- scalafmt does not insert or remove line breaks around infix operators, so weird formatting like below can usually be fixed with a bit of manual help

```scala
foo(aaaaaa(bbbbbbbbbbbbbbbb), 333333333333) == a(
  2)
// manually insert newline after `==`
foo(aaaaaa(bbbbbbbbbbbbbbbb), 333333333333) ==
  a(2)
```
I really wanted scalafmt to be able handle infix operators well (at least common ones like == && ||) but I always hit on places where I didn't like the output. Instead of forcing bad formatted output without an escape hatch, scalafmt respects the line break choices in the original source for infix operators.

- if you enforce scalafmt in CI with `scalafmt --test`, you can customize the failure message with to help contributors figure out how to fix the problem.

```scala
onTestFailure = "To fix this error, run XXX".
```

- if you really dislike the output, please report an issue with the original code, formatted code and your expected output. Some things are super easy to fix while others are a lot harder. 

- the `docstring`/`spaces.inImportCurlyBraces`/`spaces.beforeSeqWildcard` options are totally reasonable. I've opened https://github.com/scalameta/scalafmt/issues/890 to discuss enabling those in the default style. If that meets big resistance, I'm open to add `style=sbt`.

Finally, my experience is that the formatted output tends to look a bit nicer when new code is written with scalafmt in mind (esp. max column) compared to when an existing codebase is formatted with scalafmt.

dwijnand:
@tpolecat @olafurpg I too find spaces between multi-line definitions to make code much better. The reason I didn't include `newlines.alwaysBeforeTopLevelStatements=true` was because of issues in how that works. I've taken the time to file those issues now (https://github.com/scalameta/scalafmt/issues/892, https://github.com/scalameta/scalafmt/issues/893), then we can do that formatting.

Daxten:
Will take a look this weekend!

SethTisue:
scalafmt all the things! 🎉

project/AutomateScalafmtPlugin.scala
```diff
@@ -0,0 +1,67 @@
...
+              format(_ => (), "Reformatted") // Recalculate the cache
+            }
+          )
+        )
+      }
+  }
+
+  private val scalafmtInc = taskKey[Unit]("Incrementally format modified sources")
+
+  override def requires = ScalafmtPlugin
```
eed3si9n, line 61: @dwijnand See here. It requires ScalafmtPlugin.

PR #3130, opened 2017-04-21
Improve `Load` in sbt

eed3si9n:
This is a touch up of https://github.com/sbt/sbt/pull/3121 to address a cosmetic point to clear up PR queue before applying formatter.

### original PR

It mainly does three things:

* Clean up the implementation, removing unused values like
  `globalPluginDefs`.
* Make the implementation more readable.
* And the big one: Remove the creation of a classloader that we were
  instantiating but whose value we were throwing away. This has an impact
  in performance, but I'm yet to benchmark how much it is.
  
The second commit reduces the complexity around `loadPluginDefinition` et al.
`pluginDefinitionLoader` is not used anywhere in sbt, so the extra definitions are removed.

Both the implementation of `loadPluginDefinition` and `pluginDefinitionLoader` are reduced to a bare minimum where the components at hand (definition classpath, dependency classpath) are properly defined.

Documentation to the three main methods has been added.

### DIFF
--- a/main/src/main/scala/sbt/internal/Load.scala
+++ b/main/src/main/scala/sbt/internal/Load.scala
@@ -34,7 +34,7 @@ import Keys.{
   thisProjectRef,
   update
 }
-import tools.nsc.reporters.ConsoleReporter
+import scala.tools.nsc.reporters.ConsoleReporter
 import sbt.internal.util.{ Attributed, Settings, ~> }
 import sbt.util.{ Eval => Ev, Show }
 import sbt.internal.util.Attributed.data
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/internal/Load.scala
```diff
@@ -988,19 +1045,25 @@ private[sbt] object Load {
...
+  private var timedIndentation: Int = 0
+
   /** Debugging method to time how long it takes to run various compilation tasks. */
   private[sbt] def timed[T](label: String, log: Logger)(t: => T): T = {
+    timedIndentation += 1
     val start = System.nanoTime
     val result = t
     val elapsed = System.nanoTime - start
-    log.debug(label + " took " + (elapsed / 1e6) + " ms")
+    timedIndentation -= 1
```
dwijnand, line 1057: I wish this were at least wrapped try/finally, if not also been an AtomicInteger..

jvican: Thanks for having a look at this @dwijnand. These are the reasons why I discarded those ideas:
* The execution of the sbt load is sequential and `timed` is only used within the `Load` object.
* If an exception is thrown, sbt stops, reports it and exits. Given these circumstances, there should be no corruption of the value of indentation.
  
What we could do, however, is to make this utility global and accessible to other sbt subsystems where these conditions don't hold. But I haven't had the need to compute the execution time of other sbt utils, so that's why I didn't consider it in the first place.

dwijnand: I see. Thanks, @jvican.

PR #3141, opened 2017-04-25
Replace buggy scalafmt autoplugin by hook

jvican:
Benefits of this approach:
  
* Reformatting only happens before the commit, not at every compilation.
* The git pre-commit hook is automatically set up if it doesn't exist.
* Incremental compilation works.

Notes:
  
* Those users that have local pre-commit hooks will need to copy paste the
  commit hook by hand. It is placed under the `scripts` dir.


jvican:
Please review @eed3si9n and @dwijnand. Even if `AutomateScalafmtPlugin` is working on sbt (it fails on Zinc as I've described in the ticket), I think this is a considerable improvement over the status quo and would like to merge it.

jvican:
This PR has quickly evolved and the solution I came up with is more sophisticated than the status quo. In short, the commit template shown to users is not updated with the git changes that occur in the git hooks (though internally those changes will be used for the commit). This means that if a file is formatted, users won't see it even though the latest git state will be pushed. This makes users miss lots of the formatted files. To see the true results, they have to cancel the operation and run commit again.
  
The python script in this PR works around this issue and recreates the message with the up-to-date information. This is a general solution that supports anything that the users wants (git templates, alternative git hooks, etc). The issue is that it will automatically add modified files that have been staged. This may be unexpected for lots of users, especially if they keep a dirty state before doing commits.
  
Unfortunately, this is enough of a problem so that I consider closing this PR. The solution presented here is too sophisticated. I will use these scripts myself, and with more work I believe it's possible to provide an automated way to warn users collisions between dirty files and files that have been formatted. But for now, I'm closing this and we'll see what the future brings us.

eed3si9n:
Sounds good. Perhaps a simpler improvement we can add to the existing auto plugin is disable feature that can be called from sbt shell for power users.

PR #3147, opened 2017-04-27
Improvements & bug fixes to the new startup messages

### LINKED ISSUE #3091: `--error` does not suppress new batch mode warning in 0.13.15
```text
% sbt --error compile
```

this regresses on #840

### LINKED ISSUE #3097: `sbt new` triggers batch mode warning
## steps

with sbt 0.13.15

```
> sbt new
```

## problem

`[warn]`

## expectation

no warning :)

dwijnand:
The two new startup messages introduced in sbt 0.13.15 are:

+ when writing out `sbt.version`, for build reproducability, and
+ when informing the user about sbt's shell, for the performance improvement

When writing out `sbt.version` the messaging now:

+ correctly uses a logger rather than println
+ honours the log level set, for instance by `--error`
+ never executes when sbt "new" is being run

When informing the user about sbt's shell the messaging now:

+ is a 1 line message, rather than 3
+ is at info level, rather than warn level
+ can be suppressed with `suppressSbtShellNotification := true`
+ only triggers when "compile" is being run
+ never shows when sbt "new" is being run

Fixes #3091
Fixes #3097

### DIFF
--- a/main/command/src/main/scala/sbt/State.scala
+++ b/main/command/src/main/scala/sbt/State.scala
@@ -178,16 +178,16 @@ object State {
   /** Provides operations and transformations on State. */
   implicit def stateOps(s: State): StateOps = new StateOps {
     def process(f: (String, State) => State): State = {
-      def doX(x: String, xs: Seq[String]) = {
-        log.debug(s"> $x")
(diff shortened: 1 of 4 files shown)


eed3si9n (approved):
Nice work. Thanks for addressing this.

PR #3149, opened 2017-04-28
Clean up `update` in `Defaults`

jvican:
The current implementation of `update` is complex and large.

The following commits make the implementation clearer and move the bits that
don't depend on tasks or settings defined in `Defaults` to the
`librarymanagement.internal` namespace.

For more information on the changes, please have a look at the commit messages.

### DIFF
--- a/main/src/main/scala/sbt/internal/librarymanagement/AltLibraryManagementCodec.scala
+++ b/main/src/main/scala/sbt/internal/librarymanagement/AltLibraryManagementCodec.scala
@@ -1,0 +1,113 @@
+package sbt.internal.librarymanagement
+
+import sbt.internal.util.{ HList, HNil }
+import sbt.internal.util.Types._
+import sbt.io.{ Hash, IO }
+import sbt.librarymanagement._
+import sbt.util._
+import sbt.util.CacheImplicits._
+import sjsonnew.JsonFormat
+
+object AltLibraryManagementCodec extends LibraryManagementCodec {
+  type In0 = ModuleSettings :+: UpdateConfiguration :+: HNil
+  type In = IvyConfiguration :+: In0
+
+  object NullLogger extends sbt.internal.util.BasicLogger {
+    override def control(event: sbt.util.ControlEvent.Value, message: ⇒ String): Unit = ()
+    override def log(level: Level.Value, message: ⇒ String): Unit = ()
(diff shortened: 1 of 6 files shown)


main/src/main/scala/sbt/internal/librarymanagement/AltLibraryManagementCodec.scala
```diff
@@ -0,0 +1,113 @@
+package sbt.internal.librarymanagement
```
eed3si9n: `sbt.internal.librarymanagement` package should be used exclusively by sbt/librarymanagement.
Mostly so we know where things are, but there might be JDK 9 Jigsaw rule that enforces one jar - one package (?). In any case this might be a good timing to move this code to sbt/librarymanagement. If not it should be under `sbt.internal`.

main/src/main/scala/sbt/internal/librarymanagement/DependencyResolver.scala
```diff
@@ -0,0 +1,117 @@
...
+import java.io.File
+
+import sbt.internal.util.Types._
+import sbt.internal.util.HNil
+import sbt.librarymanagement._
+import sbt.util.{ CacheStore, CacheStoreFactory, Logger, Tracked }
+import sbt.librarymanagement.syntax._
+import sbt.util.CacheImplicits._
+
+object DependencyResolver {
```
eed3si9n: Same with above. This object should be called `LibraryManagementImpl` or something like that since `Resolver` has specific meaning.

eed3si9n (changes requested):
I like the overall direction of the change, but I have a few comments on the where the code should go.

jvican:
Tests are passing and feedback is addressed too. Should we merge this?

dwijnand:
```
[error] /tmp/sbt_5b8d949e/cache-update/build.sbt:25: error: object AltLibraryManagementCodec is not a member of package sbt.internal.librarymanagement
[error]   import sbt.internal.librarymanagement.AltLibraryManagementCodec._
[error]                                         ^
[info] [error] Type error in expression
[error] x dependency-management / cache-update
```

jvican:
Can someone restart this job? https://travis-ci.org/sbt/sbt/jobs/228697659. Spurious Travis error.

jvican:
Tests are passing now.

eed3si9n (approved):

PR #1360, opened 2014-05-21
Incremental cleanup

gkossakowski:
This PR just moved code around. There are no changes to behavior.

PR #3151, opened 2017-04-28
Add parallel batch mode to scripted tests

jvican:
This pull request has two goals:
  
* Speed up scripted tests by allowing them to run in parallel.
* Allow scripted to run scripted tests by reusing the same sbt instance.
  
However, these tasks ended up not being easy. This is why this PR also addresses other issues, like the ones described in #3160.
  
The integrity of the tests has not been changed, except for one: `auto-plugins`, which still tests all the behaviour of plugins but uncovered https://github.com/sbt/sbt/issues/3164. As a result, I've done a temporary fix that accounts for the fact that `reload` in the batch mode will modify the value being tested by one.
  
This PR also deletes some deprecated methods in scripted and adjusts the code (both for scripted sbt and scripted plugin) accordingly.


dwijnand (dismissed):
LGTM.

But I expect Travis CI to fail because those deprecations are more tricky than just removing them - that's why I had left them alone for now.

You might want to just drop those hunks to get this green and merged.

dwijnand:
Oh, and well done, thank you very much for this!

eed3si9n (changes requested):
scripted forks a new instance of sbt, so I don't know if this would work on Travis CI containers due to limited memory.

jvican:
Is there a limitation in Travis that I'm now aware of? I don't see why this should fail. @eed3si9n

jvican:
> scripted forks a new instance of sbt, so I don't know if this would work on Travis CI containers due to limited memory.
  
@eed3si9n You're right. This brings me to the second improvement that I wanted to make: reusing sbt instances to run things in parallel.
  
This is a shame since tests are running much faster locally. The overhead of sbt scripted tests is mainly I/O -- this is why parallelizing them yields profitable gains.
  
Honestly, I think that Travis is not enough for sbt. I would be happy to put sbt/sbt under the Scala Center CI, where we have 64 cores and 14GB of memory. Would you be happy with it? /cc @dwijnand

jvican:
I have pushed a commit that reduces the memory for every scripted test, in the hope that this makes Travis happy... let's see.

jvican:
No luck, unfortunately. I'll try to make my next improvement to see if that helps us get out of this situation. Otherwise, I would suggest to change CIs, at this point it's clear that Travis is a blocker...

scripted/sbt/src/main/scala/sbt/test/SbtHandler.scala
```diff
@@ -80,7 +80,10 @@ final class SbtHandler(directory: File,
     val thread = new Thread() { override def run() = { p.exitValue(); server.close() } }
```
eed3si9n:
```
[error] x compiler-project / inc-pickled-refinement compiler-project / inc-pickled-refinement
[error]  Cause of test exception: {line 7}  Command failed: Remote sbt initialization failed
```

The forked sbt on the other side is failing to get up, so you might need more println around here where the Process is being spawned.

jvican: Thanks for the tip.

I think the issue here is memory since all these tests are passing locally. Can we discuss migrating CI to a more powerful machine, either Drone (free) or paying for a more powerful Travis?
  
We need to find a good way to move forward.

eed3si9n: We can potentially discuss it, but I'm relatively happy with Travis CI as we get 5 workers for free. I think that's pretty good ROI.

jvican: You get 5 workers for free but the turnaround time is at minimum 45 minutes for every commit you do, and there's usually a long queue. We need more powerful alternatives. I will not insist on using our free CI, but can you please consider paying a pro account in Travis to increase the CI resources?

Being honest here, contributing to sbt is less than an ideal experience. I've tried solving it by parallelizing the testing engine, but even if that is blocked by Travis resources, I don't know what else to do. In my view, there is no other way to move forward than increasing the resources.

jvican:
@dwijnand @eed3si9n This is ready for review. Tests are passing and description is updated. I will report on the benefits brought by this PR tomorrow.

jvican:
Note to myself: I have yet to update to latest librarymanagement, but I'll do that in another PR.

main/src/main/scala/sbt/internal/parser/SbtParser.scala
```diff
@@ -99,14 +98,15 @@ private[sbt] object SbtParser {
   }
 
   private[sbt] final val globalReporter = new UniqueParserReporter
+  private[sbt] var scalacGlobalInitReporter: Option[ConsoleReporter] = None
 
   private[sbt] final lazy val defaultGlobalForParser = {
-    import scala.reflect.internal.util.NoPosition
     val options = "-cp" :: s"$defaultClasspath" :: "-Yrangepos" :: Nil
-    val reportError = (msg: String) => globalReporter.error(NoPosition, msg)
+    val reportError = (msg: String) => System.err.println(msg)
```
eed3si9n, line 105: We get into trouble when we start printlning directly. I am guessing this is to surface scripted failures, but would this have negative impact to normal usage?

scripted/sbt/src/main/scala/sbt/test/ScriptedTests.scala
```diff
@@ -8,120 +8,277 @@ package test
...
+            val wrapHook = (file: File) => {
+              preHook(file)
+              runner.processStatement(sbtHandler.asInstanceOf[SbtHandler], statement, states)
+            }
+            commonRunTest(label, tempTestDir, wrapHook, handlers, runner, states, buffer)
+          }
+
+          // Run the test and delete files (except global that holds local scala jars)
+          val result = runOrHandleDisabled(label, tempTestDir, runTest, buffer)
+          IO.delete(tempTestDir.*("*" -- "global").get)
```
eed3si9n, line 208: This wipes out the forked instances temp directory? So you're saving time by not resolving sbt 1.0.0-SNAPSHOT from the second batch onwards?

eed3si9n (commented):
Overall looks pretty good.
Posted a few minor questions for clarification.

main/src/main/scala/sbt/internal/parser/SbtParser.scala
jvican: No, not at all. In fact, this should never happen. It happened only in this case because I was fiddling with the classpath of the compiler. But this should not affect normal users.

scripted/sbt/src/main/scala/sbt/test/ScriptedTests.scala
jvican: Yes, this wipes out all the content of the temp directory except the global dir holding the scala jars. The benefits of running scripted in batch more are several, though -- for instance, you always keep a hot sbt instance around.

jvican:
After playing around with lots of options, this PR is now finished.
  
**Improvements**:
  
* Batch mode execution.
* Parallel execution (by default 1 sbt instance because Travis doesn't allow more).
* Forked sbt compilation and scripted run.
* Tests in #1360 fixed.
  
It is a large PR (in number of commits), but line changes are small. I suggest reviewing all at once, and if there's a concrete question you can check if the commit message that introduced a change explains it.
  
As a result, and even with only one sbt instance running scripted tests, this PR speeds up the scripted tests by almost 2x.
* 36 minutes: https://travis-ci.org/scalacenter/sbt/builds/228317280.
* 61 minutes: https://travis-ci.org/sbt/sbt/builds/224407414.
  
I believe we may win a little bit more of a time when util-X11 is merged, which brings parallel ivy downloads.
  
In the future, when performance issues and memory leaks are addressed in sbt, I will come back to scripted to try to run 2 sbt instances at a time.
  
In the meanwhile, if you wanna run scripted locally, I suggest that you change the number of sbt instances [here](https://github.com/sbt/sbt/pull/3151/files#diff-68598a656c879eaa71bb207678c547cfR347). It does bring a speedup.

sbt/src/sbt-test/project/auto-plugins/build.sbt
```diff
@@ -52,9 +52,9 @@ check := {
 	same(ddel, None, "del in projD")
 //
 	val buildValue = (demo in ThisBuild).value
-	same(buildValue, "build 0", "demo in ThisBuild")
+	same(buildValue, "build 1", "demo in ThisBuild") // this is temporary, should be 0 until # is fixed
 	val globalValue = (demo in Global).value
-	same(globalValue, "global 0", "demo in Global")
+	same(globalValue, "global 1", "demo in Global") // this is temporary, should be 0 until # is fixed
```
Duhemm, line 57: Until what is fixed?

sbt/src/sbt-test/project/old-ops/test
```diff
@@ -1,17 +1,18 @@
...
--> compile
+-> reload
 
 $ copy-file changes/taskAppend1/build.sbt build.sbt
--> compile
+-> reload
 
 $ copy-file changes/taskAppendN/build.sbt build.sbt
--> compile
+-> reload
```
Duhemm, line 18: So scripted was starting up a whole new sbt between every command???

sbt/src/sbt-test/apiinfo/unstable-existential-names/build.sbt
```diff
@@ -1,12 +1,26 @@
...
+    }
+  }
+}
+
+val checkIterations = inputKey[Unit]("Verifies the accumulated number of iterations of incremental compilation.")
+
+checkIterations := {
+  val expected: Int = (Space ~> NatBasic).parsed
+  val actual: Int = ((compile in Compile).value match { case a: Analysis => a.compilations.allCompilations.size }) - CompileState.previousIterations
+  assert(expected == actual, s"Expected $expected compilations, got $actual")
```
Duhemm, line 25: Shouldn't this test be in zinc's suite?

sbt/src/sbt-test/compiler-project/inc-pickled-existential/build.sbt
```diff
@@ -1,15 +1,28 @@
...
+    }
+  }
+}
+
+val checkIterations = inputKey[Unit]("Verifies the accumulated number of iterations of incremental compilation.")
+
+checkIterations := {
+  val expected: Int = (Space ~> NatBasic).parsed
+  val actual: Int = ((compile in Compile).value match { case a: Analysis => a.compilations.allCompilations.size }) - CompileState.previousIterations
+  assert(expected == actual, s"Expected $expected compilations, got $actual")
```
Duhemm, line 27: This one should be moved to, shouldn't it? I think zinc's scripted supports this kind of assertion.

There are certainly other tests that could be moved over to zinc. Is there a reason to keep them here?

sbt/src/sbt-test/source-dependencies/trait-member-modified/build.sbt
```diff
@@ -22,7 +32,7 @@ TaskKey[Unit]("checkCompilations") := {
     val files = fileNames.map(new java.io.File(_))
     assert(recompiledFiles(iteration) == files, "%s != %s".format(recompiledFiles(iteration), files))
   }
-  assert(allCompilations.size == 2)
+  assert(allCompilations.size == 2, s"All compilations is ${allCompilations.size}")
```
Duhemm, line 35: Maybe say that we expected 2?

Duhemm (changes requested):
Awesome improvements, very impressive!

 - Why has the test `project-load/sha-conflict` moved from `pending` to `disabled`?
 - Travis CI kills a build if it doesn't produce any output for 10 minutes. It appears that people circumvent that issue by disabling log buffering (hopefully they'll never have to do that again, thanks to that PR 🎉). Would it be possible to have scripted periodically write something next to the test name, to show that it's not dead? Something like
```
Running run / hello-world .............. done!
```

sbt/src/sbt-test/project/auto-plugins/build.sbt
jvican: #1364

jvican: #3164

sbt/src/sbt-test/project/old-ops/test
jvican: What I meant is that `sbt` was initialized lazily, only when `>` statements were invoked. `compile` here implied initializing sbt.

sbt/src/sbt-test/apiinfo/unstable-existential-names/build.sbt
jvican: I have to port these improvements to Zinc, indeed! The issue is that this has to be included in every build.sbt. This is why we have to put it in here.

sbt/src/sbt-test/compiler-project/inc-pickled-existential/build.sbt
jvican: All these tests should be moved to Zinc yes. But I'd like to do that in the future, I don't have enough cycles for it now.

eed3si9n:
> It is a large PR (in number of commits), but line changes are small. I suggest reviewing all at once, and if there's a concrete question you can check if the commit message that introduced a change explains it.

Once the review is done, could we squash them into 1 or several number of commits for future debugging purpose?

jvican:
@eed3si9n Honestly, I prefer not to. They have been done this way so that they are easy to reason about, and each one implements a granular change that may be useful for the future. That being said, I would like to squash those commits that have to do with travis options and "disable parallelism".

jvican:
> Travis CI kills a build if it doesn't produce any output for 10 minutes. It appears that people circumvent that issue by disabling log buffering (hopefully they'll never have to do that again, thanks to that PR 🎉). Would it be possible to have scripted periodically write something next to the test name, to show that it's not dead?
  
This would be an excellent issue for a Scala Center spree, thanks for the idea!

jvican:
Can we try to get this in? When reviews are done, I'll squash the pertinent commits. I'd like to get this in first before making more PRs.

eed3si9n (approved):

jvican:
I'm squashing the pertinent commits and rebasing to prepare for the merge. Will report when this is ready.

jvican:
Can someone reload this failed build instance? https://travis-ci.org/sbt/sbt/jobs/230899635 The culprit is https://github.com/sbt/sbt/issues/3170.
  
The rest of the tests are passing.

jvican:
Tests are passing.

muuki88:
@jvican great work! Sbt-native-packager scripted test run for almost 40 minutes. This will be a huge development experience improvement 😍😍

PR #3152, opened 2017-04-28
Fix default scalaBinaryVersion for Dotty

smarter:
Dotty is versioned as 0.*, but `CrossVersionUtil#binaryScalaVersion`
will return the full version instead of just major.minor for all
compiler versions < 2.10, add a special case for Dotty to avoid this.

### DIFF
--- a/util/cross/src/main/input_sources/CrossVersionUtil.scala
+++ b/util/cross/src/main/input_sources/CrossVersionUtil.scala
@@ -8,7 +8,8 @@ object CrossVersionUtil
 	val noneString = "none"
 	val disabledString = "disabled"
 	val binaryString = "binary"
-	val TransitionScalaVersion = "2.10"
+	val TransitionDottyVersion = "" // Dotty always respects binary compatibility
+	val TransitionScalaVersion = "2.10" // ...but scalac doesn't until Scala 2.10
 	val TransitionSbtVersion = "0.12"
 
 	def isFull(s: String): Boolean = (s == trueString) || (s == fullString)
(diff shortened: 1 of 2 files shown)


util/cross/src/main/input_sources/CrossVersionUtil.scala
```diff
@@ -57,7 +58,15 @@ object CrossVersionUtil
 			case PartialVersion(major, minor) => Some((major.toInt, minor.toInt))
 			case _ => None
 		}
-	def binaryScalaVersion(full: String): String = binaryVersionWithApi(full, TransitionScalaVersion)(scalaApiVersion)
+	def binaryScalaVersion(full: String): String = {
+		val cutoff =
+			if (full.startsWith("0."))
```
dwijnand, line 63: This is.. unfortunate.

dwijnand (changes requested):

dwijnand (approved):

dwijnand:
oops, miss clicked "request changes".

eed3si9n (approved):

smarter:
Another spurious test failure, unrelated to the PR:
```scala
[info] [error] (*:ivyConfiguration) Could not create file /tmp/sbt_3d9e72fd/info/target/streams/$global/ivyConfiguration/$global/streams/outjava.io.IOException: No such file or directory

[info] [error] Total time: 0 s, completed Apr 28, 2017 1:05:03 PM

[error] x dependency-management / info

[error]    {line 4}  Command failed: clean failed
```

PR #3153, opened 2017-04-28
Remove the "hit [ENTER] to switch to interactive mode" feature

### LINKED ISSUE #3091: `--error` does not suppress new batch mode warning in 0.13.15

dwijnand:
In sbt 0.13.15, in addition to notifying the user about the existence of
sbt's shell, a feature was added to allow the user to switch to sbt's
shell - a more pro-active approach to just displaying a message.

Unfortunately sbt is often unintentionally invoked in shell scripts in
"interactive mode" when no interaction is expected by, for exmaple,
invoking `sbt package` instead of `sbt package < /dev/null`. In that
case hitting [ENTER] would silently trigger sbt to run its shell,
easily wrecking the script. In addition to that I was unhappy with the
implementation as it created a tight coupling between sbt's command
processing abstraction to sbt's shell command.

If you want to stay in sbt's shell after running a task like `package`
then invoke sbt like so:

    sbt package shell

Fixes #3091

### DIFF
--- a/main/command/src/main/scala/sbt/State.scala
+++ b/main/command/src/main/scala/sbt/State.scala
@@ -182,11 +182,8 @@ object State {
         log.debug(s"> $cmd")
         f(cmd, s.copy(remainingCommands = remainingCommands, history = cmd :: s.history))
       }
-      def isInteractive = System.console != null
-      def hasInput = Option(System.console) exists (_.reader.ready())
-      def hasShellCmd = s.definedCommands exists { case c: SimpleCommand => c.name == Shell; case _ => false }
       s.remainingCommands match {
-        case Seq()           => if (isInteractive && hasInput && hasShellCmd) runCmd(Shell, Nil) else exit(true)
+        case Seq()           => exit(true)
         case Seq(x, xs @ _*) => runCmd(x, xs)
       }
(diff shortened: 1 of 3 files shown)


main/command/src/main/scala/sbt/State.scala
```diff
@@ -182,11 +182,8 @@ object State {
         log.debug(s"> $cmd")
         f(cmd, s.copy(remainingCommands = remainingCommands, history = cmd :: s.history))
       }
-      def isInteractive = System.console != null
```
jvican, line 185: Actually, I don't know why this existed if we have `interactive` in `State`.

jvican (approved):
LGTM.

main/command/src/main/scala/sbt/State.scala
dwijnand: `interactive` in State is an indication as to whether sbt is in interactive mode, i.e it's in the process of running `shell`.

This "interactive" is whether the JVM is running in interactive mode: by virtue of having a stdin and stdout attached.

eed3si9n (approved):

PR #3163, opened 2017-05-02
Add duplicated settings test

jvican:
This test shows how sbt is recomputing the settings completely (even if
they have side effect) every time a `Project.extract` and settings
manipulations are performed.

This is worrying because it's:

* Inefficient.
* Surprising.
* Incorrect -- settings may produce side effects.

Unless I'm mistaken, this proves that the computation of settings is not
performed once.


jvican:
Can someone please restart the build? Tests failures are spurious.

jvican:
This is ready for review. It adds a pending test for #3164.

eed3si9n (approved):

PR #3173, opened 2017-05-05
Remove implicit formats from Defaults

jvican:
`Defaults` is not the right place to expose the implicit formats. Since
this file is usually read by people, we should keep it as clean as
possible and hide implementation details.

This commit does a small cleanup in update too:

* Uses `dependsOn` to show idiomatic ways of expressing our previous
  implementation.
* Hides the JsonKeyFormat in an `UnsafeLibraryManagementCodec` file.
  I must admit that I don't know why we have an extra format for
  ModuleID.

Regarding dependency positions, the meat of the implementation has been
moved to an independent file, as well as the implicit formats. The
implementation of `dependencyPositionsTask` is now kept to a bare
minimum. Docs have been added too.

### DIFF
--- a/main/src/main/scala/sbt/internal/librarymanagement/UnsafeLibraryManagementCodec.scala
+++ b/main/src/main/scala/sbt/internal/librarymanagement/UnsafeLibraryManagementCodec.scala
@@ -1,0 +1,17 @@
+package sbt.internal.librarymanagement
+
+import sbt.librarymanagement.{ LibraryManagementCodec, ModuleID }
+import sjsonnew.JsonFormat
+
+object UnsafeLibraryManagementCodec {
+  val moduleIdJsonKeyFormat: sjsonnew.JsonKeyFormat[ModuleID] = {
+    new sjsonnew.JsonKeyFormat[ModuleID] {
+      import sjsonnew.support.scalajson.unsafe._
+      val moduleIdFormat: JsonFormat[ModuleID] = LibraryManagementCodec.ModuleIDFormat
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/internal/librarymanagement/UnsafeLibraryManagementCodec.scala
```diff
@@ -0,0 +1,17 @@
+package sbt.internal.librarymanagement
```
eed3si9n, line 1: This should be moved to sbt.internal.

eed3si9n (changes requested):

main/src/main/scala/sbt/internal/librarymanagement/UnsafeLibraryManagementCodec.scala
jvican: I didn't quite get the reasonind behind this. To me, this utility is totally part of `librarymanagement` and should go here. The rest of the utilities in this package are very similar to this one. Can you elaborate?

eed3si9n: For the sake of code organization, sbt/sbt code should be packaged under `sbt.internal`. If it belongs to `sbt.internal.librarymanagement` it should move to sbt/lm.

dwijnand:
This conflicts with current 1.x Defaults.scala. Probably easier to re-author the change than attempt to rebase it. Closing, new PR welcome.

PR #3180, opened 2017-05-08
[sbt 1.0] Rename scope component Global to Zero (`in Global` stays the same)

### LINKED ISSUE #3178: Keep Global as scope, but rename Global as scope component to Zero
## problem

While working on the [Scopes](http://www.scala-sbt.org/0.13/docs/Scopes.html) and [Scope Delegation](http://www.scala-sbt.org/0.13/docs/Scope-Delegation.html) document, I noticed that the term `Global` in sbt is used for two different meaning.

- One meaning is the universal fallback scope component.
- Second meaning is an alias for `GlobalScope` which contains `Global` on all three axes.

The user typically only come across the second form, which makes it difficult to fully grasp the first meaning. Scoping is already one of the trickiest part of sbt, and this ambiguity doesn't help.

## proposal

Keep the `Global` scope, but rename the `Global` as scope component to `Zero: ScopeAxis[Nothing]`.

So `Global == Scope(Zero, Zero, Zero)`.

The term `Zero` matches the multi-axis intuition for `(Int, Int, Int)` tuple.

## notes

Historically sbt seems to have pretended that scoping on a single axis is able to form a scope, e.g. `in Compile`, and blurred the boundary between `sbt.ScopeAxis` and `sbt.Scope`. In the updated docs, I've broken them down to calling one _scope component_ and the other _full scope_ or a "tuple" to clarify the multi-component nature of the scope.

For compatibility, in sbt 0.13, we can provide `Zero` as simple alias for `Global`, and do the renaming in sbt 1.0. The scope component `Global` normally doesn't come up in the context of build definition, so there should be little impact to the users.

For example, before:

```scala
// same as concurrentRestrictions in (Global, Global, Global)
concurrentRestrictions in Global := Seq(
  Tags.limitAll(1)
)
```

after:

```scala
// same as concurrentRestrictions in (Zero, Zero, Zero)
concurrentRestrictions in Global := Seq(
  Tags.limitAll(1)
)
```

I feel like others may have made similar proposal in the past, so I'm not claiming originality of the change, but I can't track down the previous discussion on it so I might be imagining it too.

eed3si9n:
Fixes #3178

While working on the Scopes and Scope Delegation document, I noticed that the term Global in sbt is used for two different meaning.

1. Universal fallback scope component `*`
2. An alias for GlobalScope

This disambiguates the two by renaming ScopeAxis instance to Zero.
Since this is mostly internal to sbt code, the impact to the user should be minimal.

## notes

Historically sbt seems to have pretended that scoping on a single axis is able to form a scope, e.g. `in Compile`, and blurred the boundary between `sbt.ScopeAxis` and `sbt.Scope`. In the updated docs, I've broken them down to calling one _scope component_ and the other _full scope_ or a "tuple" to clarify the multi-component nature of the scope.

For compatibility, in sbt 0.13, we can provide `Zero` as simple alias for `Global`, and do the renaming in sbt 1.0. The scope component `Global` normally doesn't come up in the context of build definition, so there should be little impact to the users.

For example, before:

```scala
// same as concurrentRestrictions in (Global, Global, Global)
concurrentRestrictions in Global := Seq(
  Tags.limitAll(1)
)
```

after:

```scala
// same as concurrentRestrictions in (Zero, Zero, Zero)
concurrentRestrictions in Global := Seq(
  Tags.limitAll(1)
)
```

I feel like others may have made similar proposal in the past, so I'm not claiming originality of the change, but I can't track down the previous discussion on it so I might be imagining it too.

### DIFF
--- a/main-settings/src/main/scala/sbt/ScopeAxis.scala
+++ b/main-settings/src/main/scala/sbt/ScopeAxis.scala
@@ -3,26 +3,41 @@ package sbt
 import sbt.internal.util.Types.some
 
 sealed trait ScopeAxis[+S] {
-  def foldStrict[T](f: S => T, ifGlobal: T, ifThis: T): T = fold(f, ifGlobal, ifThis)
-  def fold[T](f: S => T, ifGlobal: => T, ifThis: => T): T = this match {
+  def foldStrict[T](f: S => T, ifZero: T, ifThis: T): T = fold(f, ifZero, ifThis)
+  def fold[T](f: S => T, ifZero: => T, ifThis: => T): T = this match {
     case This      => ifThis
-    case Global    => ifGlobal
+    case Zero      => ifZero
     case Select(s) => f(s)
   }
   def toOption: Option[S] = foldStrict(some.fn, None, None)
-  def map[T](f: S => T): ScopeAxis[T] = foldStrict(s => Select(f(s)), Global, This)
+  def map[T](f: S => T): ScopeAxis[T] = foldStrict(s => Select(f(s)), Zero, This)
   def isSelect: Boolean = false
 }
+
+/**
+ * This is a scope component that represents not being
+ * scoped by the user, which later could be further scoped automatically
+ * by sbt.
+ */
 case object This extends ScopeAxis[Nothing]
-case object Global extends ScopeAxis[Nothing]
+
+/**
+ * Zero is a scope component that represents not scoping.
+ * It is a universal fallback component that is strictly weaker
+ * than any other values on a scope axis.
+ */
+case object Zero extends ScopeAxis[Nothing]
+
+/**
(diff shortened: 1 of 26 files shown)


main-settings/src/main/scala/sbt/ScopeAxis.scala
```diff
@@ -3,26 +3,41 @@ package sbt
...
+/**
+ * This is a scope component that represents not being
+ * scoped by the user, which later could be further scoped automatically
+ * by sbt.
+ */
 case object This extends ScopeAxis[Nothing]
-case object Global extends ScopeAxis[Nothing]
+
+/**
+ * Zero is a scope component that represents not scoping.
```
jvican, line 25: :+1:

jvican (approved):
LGTM.

Issue #3167, opened 2017-05-02
Migrate InteractionService from sbt-core-next


eed3si9n:
## steps

Migrate sbt-pgp.

## problem

InteractionService is missing in sbt 1.0.
See https://github.com/sbt/sbt-pgp/pull/100

## expectation

InteractionService should be there.

PR #3182, opened 2017-05-09
Add InteractionService from sbt-core-next

eed3si9n:
Fixes sbt/sbt#3167


jvican (commented):
Can this be used for sbt-bintray too and thus remove the specific code used for getting input there?

eed3si9n:
Probably https://github.com/sbt/sbt-bintray/blob/b0d9e37a8b9540a679a41534a1e5865a3eb8be20/src/main/scala/Prompt.scala can be replaced with this.

Issue #2699, opened 2016-08-17
Eviction warning is unclear


chris-martin:
Example output from a compile using SBT 0.13.7:

```
[warn] There may be incompatibilities among your library dependencies.
[warn] Here are some of the libraries that were evicted:
[warn]  * com.google.code.findbugs:jsr305:2.0.1 -> 3.0.0
[warn] Run 'evicted' to see detailed eviction warnings
```

I don't understand what `2.0.1 -> 3.0.0` signifies. Which version was evicted, 2.0.1 or 3.0.0? Which version was used? What does the arrow mean?

eed3si9n:
Ivy normally picks the latest. In this case, 2.0.1 got evicted in favor of 3.0.0.

chris-martin:
I take it you disagree that the warning is insufficiently explanatory?

eed3si9n:
I'm open to alternative wording, but I personally think it's clear enough once you know what it means. Given that the user might see the message multiple times a day, this sort of messages need to err on the shorter/insufficient side.

chris-martin:
Replacing the arrow with your English explanation adds 22 characters:

```
[warn] There may be incompatibilities among your library dependencies.
[warn] Here are some of the libraries that were evicted:
[warn]  * com.google.code.findbugs:jsr305: 2.0.1 got evicted in favor of 3.0.0
[warn] Run 'evicted' to see detailed eviction warnings
```

chris-martin:
> once you know what it means

That's the tricky part. Is there a path to finding out what it means?

eed3si9n:
If someone googles "sbt eviction warning" the first hit would be the documentation http://www.scala-sbt.org/0.13/docs/Library-Management.html#Eviction+warning and the second is this blog post [Improved dependency management with sbt 0.13.7](https://www.lightbend.com/blog/improved-dependency-management-with-sbt-0137).

If someone knows what "eviction" means, I think current arrow is clear enough. Also note there might be multiple (potentially many) entries

```
[warn] There may be incompatibilities among your library dependencies.
[warn] Here are some of the libraries that were evicted:
[warn]  * com.google.code.findbugs:jsr305:2.0.1 -> 3.0.0
[warn]  * bar.foo:x:2.0.1 -> 3.0.0
[warn]  * io.commons:io.commons:2.0.1 -> 3.0.0
[warn]  * com.typesafe.akka:akka-actor_2.10:2.1.4 -> 2.3.7
[...]

PR #3204, opened 2017-05-23
Improve the eviction warning presentation.

eed3si9n:
Fixes sbt/sbt#2699

Before:

    [warn] There may be incompatibilities among your library dependencies.
    [warn] Here are some of the libraries that were evicted:
    [warn]  * com.google.code.findbugs:jsr305:2.0.1 -> 3.0.0
    [warn] Run 'evicted' to see detailed eviction warnings

After:

```
[warn] Found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[warn]
[warn] 	* org.apache.logging.log4j:log4j-api:2.8.2 is selected over 1.4
[warn] 	    +- de.heikoseeberger:akka-log4j_2.12:1.4.0            (depends on 2.8.2)
[warn] 	    +- com.example:hello-world_2.12:0.1-SNAPSHOT          (depends on 1.4)
[warn]
[warn] 	* com.typesafe.akka:akka-actor_2.12:2.5.0 is selected over 2.4.17
[warn] 	    +- de.heikoseeberger:akka-log4j_2.12:1.4.0            (depends on 2.5.0)
[warn] 	    +- com.typesafe.akka:akka-parsing_2.12:10.0.6         (depends on 2.4.17)
[warn] 	    +- com.typesafe.akka:akka-stream_2.12:2.4.17 ()       (depends on 2.4.17)
[warn]
[warn] Run 'evicted' to see detailed eviction warnings
```

### DIFF
--- a/ivy/src/main/scala/sbt/EvictionWarning.scala
+++ b/ivy/src/main/scala/sbt/EvictionWarning.scala
@@ -120,18 +120,24 @@ final class EvictionPair private[sbt] (
 object EvictionPair {
   implicit val evictionPairLines: ShowLines[EvictionPair] = ShowLines { a: EvictionPair =>
     val revs = a.evicteds map { _.module.revision }
-    val revsStr = if (revs.size <= 1) revs.mkString else "(" + revs.mkString(", ") + ")"
-    val winnerRev = (a.winner map { r =>
-      val callers: String =
-        if (a.showCallers)
-          r.callers match {
-            case Seq() => ""
-            case cs    => (cs map { _.caller.toString }).mkString(" (caller: ", ", ", ")")
-          }
-        else ""
-      r.module.revision + callers
-    }) map { " -> " + _ } getOrElse ""
-    Seq(s"\t* ${a.organization}:${a.name}:${revsStr}$winnerRev")
+    val revsStr = if (revs.size <= 1) revs.mkString else "{" + revs.mkString(", ") + "}"
+    val seen: mutable.Set[ModuleID] = mutable.Set()
+    val callers: List[String] = (a.evicteds.toList ::: a.winner.toList) flatMap { r =>
+      val rev = r.module.revision
+      r.callers.toList flatMap { caller =>
+        if (seen(caller.caller)) Nil
+        else {
+          seen += caller.caller
+          List(f"\t    +- ${caller}%-50s (depends on $rev)")
+        }
+      }
+    }
+    val winnerRev = a.winner match {
+      case Some(r) => s":${r.module.revision} is selected over ${revsStr}"
+      case _       => " is evicted completely"
+    }
+    val title = s"\t* ${a.organization}:${a.name}$winnerRev"
+    title :: (if (a.showCallers) callers.reverse else Nil) ::: List("")
   }
 }
 
@@ -227,13 +233,13 @@ object EvictionWarning {
     }
 
     if (a.directEvictions.nonEmpty || a.transitiveEvictions.nonEmpty) {
-      out += "There may be incompatibilities among your library dependencies."
-      out += "Here are some of the libraries that were evicted:"
+      out += "Found version conflict(s) in library dependencies; some are suspected to be binary incompatible:"
+      out += ""
       out ++= (a.directEvictions flatMap { _.lines })
       out ++= (a.transitiveEvictions flatMap { _.lines })
     }
 
-    if (a.allEvictions.nonEmpty && a.reportedEvictions.nonEmpty && !a.options.showCallers) {
+    if (a.allEvictions.nonEmpty && a.reportedEvictions.nonEmpty) {
       out += "Run 'evicted' to see detailed eviction warnings"
     }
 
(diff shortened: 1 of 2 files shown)


sjrd:
The warning message is definitely clearer than it used to be. But I would disagree that this "fixes #2699". There is a lot of discussion in that ticket that goes much further than just "the message is unclear".

omervk:
@eed3si9n Great work on this, this is a real step forward in debugging eviction issues and a step forward for sbt/sbt#2699.

There are a few suggestions I'd like to make:
1. There is no longer a reference to calling 'evicted', which might be useful. It also adds serendipity (e.g. teaches a novice that this is the professional term used in this context, adding power to their future googling for solutions).
2. I would very much appreciate it if you could give the entire path, instead of just one item per dependency. Here's an example: I have a dependency on `b 1.1` and `d 1.0`. I want to upgrade to `b 1.3`, causing an eviction notice of `d 1.0 -> d 2.0` when `b` does not list `d` in its list of dependencies. Seeing something like `b 1.1 -> c 1.0 -> d 1.0 :: b 1.3 -> c 2.0 -> d 2.0` would let us know that the dependency of `b` on `c` is the one that is causing this issue. We could then look for a different version of `b` (`1.2`?) which depends on `c < 2.0` by going through the list of dependencies for each version. This could be great even as a different command (for instance in 'evicted'?)

Again, greatly appreciate your work on this.

jvican:
IMO, active voice (rather than passive voice) reads better:
```
[warn] Found version mismatch(es) of library dependencies; some are suspected to be binary incompatible:
[warn] 	* com.typesafe.akka:akka-actor_2.12:2.5.0 is selected over 2.4.17
[warn] 	    +- de.heikoseeberger:akka-log4j_2.12:1.4.0            (depends on 2.5.0)
[warn] 	    +- com.typesafe.akka:akka-parsing_2.12:10.0.6         (depends on 2.4.17)
[warn] 	    +- com.typesafe.akka:akka-stream_2.12:2.4.17 ()       (depends on 2.4.17)
```
  
Also, two more things:
1. Can we use some colors to highlight the offending versions? I find that colors make sbt ouput easier to read, telling the user what's more or less important.
2. Can we remove the unnecessary whitespaces?
  
I would like to note that the fact that concrete dependencies are now blamed makes it more urgent to allow modules to define the semantic versioning scheme that should be used for every module (for instance, sjs modules). This may be a good addition for sbt 1.0.
  
Regarding what @sjrd mentions https://github.com/sbt/sbt/issues/2699/, how about we rename the title of that issue to reflect the contents of that discussion and say that this PR addresses one of the issues highlighted there, rather than all?

eed3si9n:
@sjrd

> The warning message is definitely clearer than it used to be. But I would disagree that this "fixes #2699". There is a lot of discussion in that ticket that goes much further than just "the message is unclear".

I think it addresses the core issue around messaging. Any discussions around pro/con of changing version scheme and semantics should probably branch out to new issues.

@omervk 

> There is no longer a reference to calling 'evicted', which might be useful. It also adds serendipity (e.g. teaches a novice that this is the professional term used in this context, adding power to their future googling for solutions).

Right. Now that caller is displayed by default the it's no longer hitting the condition to display the message. I can change that so it will get displayed.

> I would very much appreciate it if you could give the entire path, instead of just one item per dependency.

That would be nice (I do display full tree when a dependency is not found), but the data structure doesn't have access to the information, and I first wanted to fix the messaging. This could be further implemented in the future.

@jvican 

> [warn] Found version mismatch(es) of library dependencies; some are suspected to be binary incompatible:

How about "[warn] Found version conflict(s) in library dependencies; some are suspected to be binary incompatible:"?

> Can we use some colors to highlight the offending versions? I find that colors make sbt output easier to read, telling the user what's more or less important.

You mean put yellow or red to version numbers `2.5.0` and `2.4.17`? I don't know if it highlights beyond the vertical alignment. Also I can't tell objectively what's "offending":
- you can say that akka-stream_2.12 is depending on Akka 2.4, and it needs to upgrade.
- or, if everyone else is depending on Akka 2.4, the offending might be akka-log4j_2.12.

> Can we remove the unnecessary whitespaces?

Blank lines make it more readable imho. Especially when you have multiple of them:

```
[warn] Multiple versions of the library dependencies were found; some suspected to be binary incompatible:
[warn]
[warn] 	* org.apache.logging.log4j:log4j-api:2.8.2 is selected over 1.4
[warn] 	    +- de.heikoseeberger:akka-log4j_2.12:1.4.0            (depends on 2.8.2)
[warn] 	    +- com.example:hello-world_2.12:0.1-SNAPSHOT          (depends on 1.4)
[warn]
[warn] 	* com.typesafe.akka:akka-actor_2.12:2.5.0 is selected over 2.4.17
[warn] 	    +- de.heikoseeberger:akka-log4j_2.12:1.4.0            (depends on 2.5.0)
[warn] 	    +- com.typesafe.akka:akka-parsing_2.12:10.0.6         (depends on 2.4.17)
[warn] 	    +- com.typesafe.akka:akka-stream_2.12:2.4.17 ()       (depends on 2.4.17)
[warn]
```

eed3si9n:
I need to start forward porting 0.13 changes to 1.0, so I'm going to merge this as-is.

PR #3208, opened 2017-05-23
Move `appmacro` and `collection` to sbt/sbt

jvican:
This commit moves everything behind the old `util-appmacro` and
`util-collection` over to sbt/sbt.

The reason for this change is because the logic of these modules is
really tight to the implementation of sbt and, I would even say, define
what sbt essentially is. They don't deserve to be called `util` nor to
be in an independent repository.

I have renamed the modules because I do not think they conveyed how
important these are for the sbt project. I don't have strong opinions
here, so I'm happy to pick other names if others consider appropriate.

`appmacro` has been renamed as `core-macros`. It is indeed the project
that are then implemented by `main-settings`. Because of this strong

`util-collection` has been renamed as `core-settings`. It defines all
the core structures (like `Init`, `AttributeKey`, `Setting`) that are
the core of sbt. These data structures are used by `appmacro` and the
same argument holds to move this module here. Given the importance of
these data structures, I feel that this should be one of the major
projects of sbt.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -146,21 +146,22 @@ lazy val testAgentProj = (project in file("testing") / "agent").settings(
 
 // Basic task engine
 lazy val taskProj = (project in file("tasks"))
+.dependsOn(settingsDataProj)
   .settings(
     testedBaseSettings,
     name := "Tasks"
   )
-  .configure(addSbtUtilControl, addSbtUtilCollection)
+  .configure(addSbtUtilControl)
 
 // Standard task system.  This provides map, flatMap, join, and more on top of the basic task model.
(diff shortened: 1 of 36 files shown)


jvican:
I close this, I've realised we cannot move the old `util-collection` here because every module in `util` depends on it. I'll reopen only moving the old `util-apply`.

eed3si9n:
Could we preserve the Git history plz.

jvican:
@eed3si9n Sure. I'll look into how to do it now.

eed3si9n:
https://github.com/sbt/util/commit/ef7ff653cb251acc96eb4210f87ac8d978cedb19

PR #3209, opened 2017-05-23
Move `util-appmacro` to sbt as `core-macros`

jvican:
This commit moves everything behind the old util-appmacro over to
sbt/sbt.

appmacro has been renamed as core-macros. It is indeed the project
that are then implemented by main-settings. Because of this strong

The reason for this change is because the logic of this module is really
tight to the implementation of sbt and, I would even say, essentially
define the sbt DSL.


jvican:
This is not ready yet. Hold tight.

jvican:
Last commit should make it ready.

eed3si9n:
The PR is missing the actual (pre-name) history that took place in sbt/sbt such as https://github.com/sbt/util/commit/c95df4681bf1ae03348bf08640c62b9d25aeb47c. I am guessing that directory specification needs `util/appmacro`.

dwijnand:
> The PR is missing the actual (pre-name) history that took place in sbt/sbt such as sbt/util@c95df46.

It isn't (wasn't) because that history is already here, given this _is_ sbt/sbt: https://github.com/sbt/sbt/commit/1c22478ed

Having merged #3211 we now have multiple versions of the same history within the same repository; see 1c22's filter-branch mutilated twin: https://github.com/sbt/sbt/commit/087e386c9

jvican:
Oh wow, that's sad 😢

PR #3211, opened 2017-05-23
Move `util-appmacro` to sbt as `core-macros`

jvican:
This commit moves everything behind the old util-appmacro over to
sbt/sbt.

appmacro has been renamed as core-macros. It is indeed the project
that are then implemented by main-settings. Because of this strong

The reason for this change is because the logic of this module is really
tight to the implementation of sbt and, I would even say, essentially
define the sbt DSL.


jvican:
Hope this makes it. @eed3si9n

eed3si9n (approved):
Looks fine. Let's see if Travis passes.

PR #3216, opened 2017-05-25
[sbt 1.0] Add first round of DSL checks to sbt

jvican:
This pull request adds the necessary infrastructure to run sbt checks.  The
purpose of the PR is to allow sbt to prohibit DSL usages that are dangerous and
yield unexpected results.  One of these cases is the use of `value` inside if
expressions (though there are more).

This will make it easier both to learn sbt and to use it. I am very optimistic
about the result of these changes in the close future.

For the review, I suggest having a look at the independent commits to get an
idea of what I am doing in every one of them.


jvican:
The error messages are inspired by [those of Rust](https://blog.rust-lang.org/2016/08/10/Shape-of-errors-to-come.html) and those of Dotty.
  
In the future, for sbt DSL checks that are warnings and not errors, I'd like to add a `showError $ERROR_ID` where $ERROR_ID corresponds to an identifier shown in the message that elaborates on the reported error, the consequences, et cetera.

jvican:
An example of the sbt check:
sbt-dsl-error-if

main-settings/src/main/scala/sbt/unchecked.scala
```diff
@@ -0,0 +1,13 @@
...
+
+/** An annotation to designate that the annotated entity
+ *  should not be considered for additional sbt compiler checks.
+ *  These checks ensure that the DSL is predictable and prevents
+ *  users from doing dangerous things at the cost of a stricter
+ *  code structure.
+ *
+ *  @since 1.0.0
+ */
+class unchecked extends Annotation
```
eed3si9n: I don't know if it's a good idea to reintroduce a famous annotation like this. Could we just use the real unchecked for this purpose?

eed3si9n (changes requested):
The change looks nice!

I'm just not too sure about the unchecked stuff.

jvican:
So, after I have implemented the first two checks, I want to give a summary of what this PR is about.
  
I feel that sbt authors have a hard time writing sbt code because it's not predictable -- sometimes it does not behave like Scala code.
  
These checks are the first step to disallow dangerous code that misbehaves. The most common scenario is the use of a task invocation inside the body of an if expression. People may think the task in the body won't be executed if the predicate is false, but this is not the case -- sbt macros hoist up that value and evalute the task no matter what the if predicate is.
  
In order to address this issue, I think it's utterly important to give good feedback and explain sbt users in the spot why that construct is dangerous. This is also a good time to familiarize with the most complicated parts of sbt.
  
Feedback on the error message format is very much welcome.

main-settings/src/main/scala/sbt/unchecked.scala
jvican: The reason I preferred to create our own unchecked annotation is because reusing the Scala one can cause more confusion, and it's used for different purposes so I don't want to overload its semantics. The good thing is that `scala.annotation.unchecked` is by default not in the scope of sbt files, but `sbt.unchecked` is -- so you can have a seamless experience. Plugin files will need to import explicitly.
  
I think that keeping the same name as the Scala one is good for people to remember it.  They share the name, but not the namespace. Otherwise, it's a tedious task to remember "that special annotation that disables the sbt checks".

eed3si9n: Plugin authors regularly write `import sbt._`. So I'd be surprised if there are ambiguity of "unchecked" if I tried to use the `scala.annotation.unchecked`. I think we should a new name like "lintoff" since the meaning is also slightly different.

main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
```diff
@@ -0,0 +1,83 @@
...
+}
+
+object TaskLinterDSLFeedback {
+  private final val startBold = if (ConsoleAppender.formatEnabled) AnsiColor.BOLD else ""
+  private final val startRed = if (ConsoleAppender.formatEnabled) AnsiColor.RED else ""
+  private final val startGreen = if (ConsoleAppender.formatEnabled) AnsiColor.GREEN else ""
+  private final val reset = if (ConsoleAppender.formatEnabled) AnsiColor.RESET else ""
+
+  def useOfValueInsideIfExpression(offendingValue: String) =
+    s"""${startBold}The evaluation of `${offendingValue}` happens always inside a regular task.$reset
```
eed3si9n: I think we should say "a task" instead "a regular task". I would think most people know it's a non-dynamic variety, and even if it's it's a dynamic task, if-expressions situation still stands.

main-settings/src/main/scala/sbt/unchecked.scala
jvican: Good argument. `lintoff` sounds a little bit unclear though. What about `@sbtUnchecked`? I would like to reuse the same meaning as `scala.annotation.unchecked`.

main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
jvican: The issue is: what happens if people think that only normal tasks exist? If this is the case and we say "a task", then they will be *very confused* because that's the only construct they are familiar with.
  
It seems that we disagree in this statement: "most people know it's a non-dynamic variety".

main-settings/src/main/scala/sbt/unchecked.scala
eed3si9n: `@sbtUnchecked` sounds good.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1769,14 +1776,18 @@ object Classpaths {
...
-    publishConfiguration := publishConfig(
-      packagedArtifacts.in(publish).value,
-      if (publishMavenStyle.value) None else Some(deliver.value),
-      resolverName = getPublishTo(publishTo.value).name,
-      checksums = checksums.in(publish).value,
-      logging = ivyLoggingLevel.value,
-      overwrite = isSnapshot.value
-    ),
+    publishConfiguration := {
+      // TODO(jvican): I think this is a bug.
```
jvican, line 1780: I think this is a bug @eed3si9n. It's not clear to me that we should deliver unconditionally here. The previous if expression has the same semantics as this one, but I think the newest code uncovers the unexpected misbehaviour.

main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
jvican: @eed3si9n Can we discuss this in a separate issue and we tackle all the error message wording before RC1? I'd like to focus on other stuff for now.

jastice:
given [all the ways *not* to get a conditional task execution right](https://stackoverflow.com/questions/42915843/redefine-an-sbt-task-conditional-on-a-setting), the "use a dynamic task" warrants a link to a more detailed explanation.
(tell me where to put it and I'll make a PR for the doc)

jvican:
> given all the ways not to get a conditional task execution right, the "use a dynamic task" warrants a link to a more detailed explanation.
(tell me where to put it and I'll make a PR for the doc)
  
Totally agreed. There are several ways we could improve this. I would like to give every error an identifier, and then allow users to browse a full explanation of everything via `help`. Then we can cross-link between error messages.
  
@jastice As this is not high priority, I suggest that we synchronize on an issue on the best ways to do so before the RC1.

main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
eed3si9n: sure.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Good catch. You're probably right.

eed3si9n (approved):

eed3si9n:
Merged! Thanks for the contribution.

jvican:
Thanks @eed3si9n.

clhodapp:
Is the error here accurate? Are you not preventing the behavior that the error claims will happen with the error? Would it not be more accurate to say "regular tasks cannot conditionally depend on other tasks"?

clhodapp:
Wonderful idea, by the way

jvican:
> Is the error here accurate? Are you not preventing the behavior that the error claims will happen with the error? Would it not be more accurate to say "regular tasks cannot conditionally depend on other tasks"?
  
Which error do you mean? :smile:

clhodapp:
Sorry! The one that's called out in the screenshot:
screenshot

It states that "Regular tasks always evaluate task inside the bodies of if expressions", which has historically been true. However: The fact that a check for this is being added means that this is no longer true, I would think. Regular tasks _won't_ evaluate tasks inside the bodies of ifs because (due to this patch) putting a call to `.value` inside of an `if` is now disallowed, right?

jvican:
Ah! @clhodapp I see what you mean.
  
Yes, you're right! The reason why we still tell users that is because users can bypass this check with `@sbtUnchecked`. So they can do something like:
```scala
if (condition) foo.value: @sbtUnchecked
else ""
```
  
Another reason why that error message is useful is because it helps sbt users realise what and why they were doing wrong. If someone migrates to 1.0 and all of a sudden gets that message, he'll get a good reason why the previous code was rejected.
  
@clhodapp I'm totally open to change the error message, what do you think would be a better wording? As a user, what would you like to see?

clhodapp:
Hrm, I feel like it's kind of weird to allow it at all (even with `@sbtUnchecked`) but I understand now that the intent with this change is not to prevent things that we don't like from being possible but rather to make sure that users know and have opted into confusing behaviors. Given that, the only changes I would make would be to word things a bit more strongly and to tell the user how to make the error go away. More like "You should generally not do this because it is confusing. If you want to do it anyway, you'll need to add @sbtUnchecked".

Perhaps:
```
s"""${startBold}The evaluation of `$task` happens always inside a regular task.$reset
   |
   |${ProblemHeader}: `$task` is inside the if expression of a regular task.
   |  Regular tasks always evaluate calls to `.value` inside both of the branches of `if` expressions,
   | regardless of whether the predicate is true or false. Because this can cause confusion, calling
   | `.value` on tasks inside of `if` branches has been disabled by default.
   |
   |${SolutionHeader}:
   |  1. If you only want to evaluate it when the if predicate is true or false, use a dynamic task.
   |  2. If you are fine with `$task` being evaluated regardless of the value of the predicate, you can make this
   |     explicit by evaluating `$task` outside the `if` expression and assigning the result to a `val`.
   |     For example, you could write `val result = $task.value; if (condition) result else ""`.
   |  3. If you understand the evaluation model and would like sbt to simply allow you to call `.value` inside
   |     of the `if` branch, you can also annotate your call to `.value` with `@sbtUnchecked`. For example, you
   |     could write `if (condition) $task.value: @sbtUnchecked else ""`. Be aware that this may be confusing
   |     for others who try to read your build definition.

""".stripMargin
```

clhodapp:
Note: I edited my last comment so if you are reading it in an email, you'll have to come over to GitHub to see the updated text. Sorry!

jvican:
> Hrm, I feel like it's kind of weird to allow it at all (even with @sbtUnchecked) but I understand now that the intent with this change is not to prevent things that we don't like from being possible but rather to make sure that users know and have opted into confusing behaviors.
  
There are two reasons what explain the existence of `@sbtUnchecked`:
  
1. We, the developers of sbt, may have missed valid use case scenarios.
2. Or there's a regression in the DSL linter and valid code is recognised as invalid. To avoid keeping users awaiting a new sbt release to fix their builds, the annotation helps them continue with their life.
  
That being said, I really like the problem statement that you've done. Thank you, I will update this message before RC1! :smile:.
  
The only thing I'm not sure about is to put the use of `sbtUnchecked` in the solution header. To me, this should be kind of secret -- only available to people that really want to work around the DSL check. I would prefer to have only one or two solutions, instead of three, which may be confusing for users and forces them to read them all. `@sbtUnchecked` is an advanced use case so I feel comfortable leaving it to the docs. Does this make sense @clhodapp ?

julienrf:
I’m a bit late but why not forbid this by construction instead of letting users write what they want and then checking that it makes sense?

clhodapp:
@jvican hrm, I feel like having features like `@sbtUnchecked` exist but not be publicized is going to reinforce sbt's reputation for being confusing/having bad documentation. If we feel the need to hide part of our api from the user, it probably shouldn't exist. Further, without the solution text explaining `@sbtUnchecked` the problem text is confusing because it explains a theoretical scenario that feels impossible; It reads as "We don't let you do this but if we did here's what would happen."

PR #3225, opened 2017-05-28
[sbt 1.0] Add missing .value DSL check and other improvements

jvican:
The following PR adds a DSL check to make sure that the following snippets report an error:

```scala
val myTask = Def.task[String] {
  val _ = compile
  fooString
}
```

```scala
val myTask = Def.task[Unit] {
  compile
}
```
  
These snippets fail because the user has forgot to add `.value` on a key. I've been myself victim of these errors on the past, thinking that a task was executed when it was not. An interesting example is where the expected return type is Unit, and hence scala automatically adds `()` at the bottom of the block and typechecks the code when it should fail.

### Implementation notes

The missing .value analysis is dumb on purpose because it's expensive.

Detecting valid use cases of idents whose type is an sbt key is difficult and
dangerous because we may miss some corner cases. Instead, we report on the
easiest cases in which we are certain that the user does not want to have a
stale key reference. Those are idents in the rhs of val definitions with `_` as
name and idents in statement position inside blocks.

In the future, we can cover all val definitions no matter what their name is.
Unfortunately, doing so will come at the cost of speed: we have to run the
unused name analysis in `TypeDiagnostics` and expose it from the power context
in `ContextUtil`.

This is good enough for now. If users express interest in having a smarter
analysis, we can always consider trying the unused name analysis. I am not sure
how slow it will be -- hopefully it won't be that much.

Other improvements: handling of more complex expressions and more tests to
check annotating expressions with `@sbtUnchecked` works.

### DIFF
--- a/main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
+++ b/main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
@@ -20,7 +20,6 @@ abstract class BaseTaskLinterDSL extends LinterDSL {
       private val settingKeyType = typeOf[sbt.SettingKey[_]]
       private val inputKeyType = typeOf[sbt.InputKey[_]]
       private val uncheckedWrappers = MutableSet.empty[Tree]
-      private val identsWithValue = MutableSet.empty[Ident]
       var insideIf: Boolean = false
       var insideAnon: Boolean = false
       var disableNoValueReport: Boolean = false
@@ -61,11 +60,6 @@ abstract class BaseTaskLinterDSL extends LinterDSL {
       override def traverse(tree: ctx.universe.Tree): Unit = {
         tree match {
           case ap @ Apply(TypeApply(Select(_, nme), tpe :: Nil), qual :: Nil) =>
-            // Keep track of wrapped idents to detect missing `.value`
-            qual match {
-              case i: Ident => identsWithValue.add(i)
-              case _        => ()
-            }
             val shouldIgnore = uncheckedWrappers.contains(ap)
             val wrapperName = nme.decodedName.toString
             if (!shouldIgnore && isTask(wrapperName, tpe.tpe, qual)) {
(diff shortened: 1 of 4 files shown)


jvican:
Note that the failed commit is due to Travis cancelling the build when a new commit is pushed. This is up for review @eed3si9n.

eed3si9n:
I'm not clear on what this PR is about. Could you show me some repro steps with before and after error messages?

jvican:
The test cases show exactly what this PR does @eed3si9n. I'm not in a computer now so I cannot take screenshots.

eed3si9n:
So the following errors:

```diff
+  test("Detect a missing `.value` inside a task") {
+    expectError(TaskLinterDSLFeedback.missingValueForKey("fooNeg")) {
+      """
+        |import sbt._
+        |import sbt.Def._
+        |
+        |val fooNeg = taskKey[String]("")
+        |
+        |def avoidDCE = {println(""); ""}
+        |val bazNeg = Def.task[String] {
+        |  fooNeg
+        |  avoidDCE
+        |}
+      """.stripMargin
+    }
+  }
```

because `fooNeg` is missing `.value` invocation.

There are some cases where passing keys unevaluated is valid. An example I can think of is sbt-buildinfo:

```scala
  enablePlugins(BuildInfoPlugin).
  settings(
    buildInfoKeys := Seq[BuildInfoKey](name, version, scalaVersion, sbtVersion),
    buildInfoPackage := "hello"
  )
```

The case I was hoping it would catch is specifically when the return type is `Unit` and value discarding happens.

jvican:
Note that error is caught:

```
+  test("Detect a missing `.value` inside a task whose return type is Unit") {
		+    expectError(TaskLinterDSLFeedback.missingValueForKey("fooNeg4")) {
		+      """
		+        |import sbt._
		+        |import sbt.Def._
		+        |
		+        |val fooNeg4 = taskKey[String]("")
		+        |
		+        |val bazNeg = Def.task[Unit] {
		+        |  fooNeg4
		+        |}
		+      """.stripMargin
		+    }
```

The example that you describe is not caught because we only report on task keys that are identifiers on statement position or that appear in the this of a valdef with an underscore.

jvican:
To summarize our private discussion, these examples that @eed3si9n shows compile successfully.

eed3si9n (approved):

Issue #3226, opened 2017-05-29
[sbt 1.0.0-M6] update caching is still not working


eed3si9n:
## steps

```scala
lazy val p1 = project
  .settings(
  )

lazy val p2 = project
  .dependsOn(p1 % "compile;test->test")
  .settings(
    libraryDependencies += zinc,
  )
```

1. compile
2. compile

## problem

I am seeing the following output:

```
> compile
[info] Updating {file:/xxx/}p2...
[warn] Choosing local for org.scala-sbt#zinc_2.12;1.0.0-X16-SNAPSHOT
[warn] Choosing local for org.scala-sbt#zinc-core_2.12;1.0.0-X16-SNAPSHOT
[warn] Choosing local for org.scala-sbt#zinc-apiinfo_2.12;1.0.0-X16-SNAPSHOT
[warn] Choosing local for org.scala-sbt#compiler-interface;1.0.0-X16-SNAPSHOT
[warn] Choosing local for org.scala-sbt#zinc-classfile_2.12;1.0.0-X16-SNAPSHOT
[warn] Choosing local for org.scala-sbt#zinc-classpath_2.12;1.0.0-X16-SNAPSHOT
[warn] Choosing local for org.scala-sbt#zinc-persist_2.12;1.0.0-X16-SNAPSHOT
[warn] Choosing local for org.scala-sbt#zinc-compile-core_2.12;1.0.0-X16-SNAPSHOT
[warn] Choosing local for org.scala-sbt#zinc-ivy-integration_2.12;1.0.0-X16-SNAPSHOT
[info] Done updating.
```

This indicates that the `update` is getting triggered even though the dependency graph has *not* changed.

## expectation

Works like sbt 0.13. Transitive call to `update` task via `compile` the second time should *not* trigger actual resolution (used to be called `doWork`, now called `def resolve`), thus the log should be much less noisier.

Basically it should be:

```
> compile
[success] Total time: 1 s, completed May 28, 2017 8:19:48 PM
> compile
[success] Total time: 0 s, completed May 28, 2017 8:19:49 PM
```

## notes

Original output on sbt:

```
> compile
[info] Updating {file:/xxx/sbt/}sbtRoot...
[info] Updating {file:/xxx/sbt/}testAgentProj...
[info] Updating {file:/xxx/sbt/}bundledLauncherProj...
[info] Updating {file:/xxx/sbt/}coreMacrosProj...
[info] Updating {file:/xxx/sbt/}taskProj...
[...]

PR #3233, opened 2017-05-29
Fix cached update task

eed3si9n:
Fixes sbt/sbt#3226

One of the checks that the cached update task performs is called `depsUpdated`, which checks if the subproject dependencies have been thawed out of cache or not.
For this to function correctly, when we thaw the `UpdateReport` from JSON, we need to mark the report as cached. This was done subtley in sbt 0.13 for sbinary (https://github.com/sbt/sbt-zero-thirteen/blob/v0.13.15/main/actions/src/main/scala/sbt/CacheIvy.scala#L66-L67).


jvican:
Great job chasing this one.

PR #3237, opened 2017-05-31
add visualize task to sbt

zartstrom:
.. from scala spree in CPH ..

We leverage the existing `graphSettings` function to create a task that writes dependency graphs to .dot-files.

See also this pdf created from `declared_dependencies.dot` output (created with `dot -Tpdf declared_dependencies.dot -o declared_deps.pdf`):
``
[declared_deps.pdf](https://github.com/sbt/sbt/files/1041913/declared_deps.pdf)

### DIFF
--- a/sbt/src/sbt-test/project/visualize/test
+++ b/sbt/src/sbt-test/project/visualize/test
@@ -1,0 +1,6 @@
+# Run the visualize command and check for the output
+> visualize
+$ exists actual_dependencies.dot 
+$ exists declared_dependencies.dot 
+# $ copy-file actual_dependencies.dot /tmp
+# $ copy-file declared_dependencies.dot /tmp
--- a/sbt/src/sbt-test/project/visualize/test
+++ b/sbt/src/sbt-test/project/visualize/test
@@ -1,6 +1,4 @@
 # Run the visualize command and check for the output
 > visualize
-$ exists actual_dependencies.dot 
-$ exists declared_dependencies.dot 
-# $ copy-file actual_dependencies.dot /tmp
-# $ copy-file declared_dependencies.dot /tmp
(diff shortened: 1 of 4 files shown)


lightbend-cla-validator:
Hi @zartstrom,

sbt/src/sbt-test/project/visualize/test
```diff
@@ -0,0 +1,6 @@
+# Run the visualize command and check for the output
+> visualize
+$ exists actual_dependencies.dot 
+$ exists declared_dependencies.dot 
+# $ copy-file actual_dependencies.dot /tmp
```
jvican: Can you remove these statements? We don't need to copy here :smile:.

jvican (approved):
LGTM, thank you Philipp for this contribution! I just requested a minor change.

eed3si9n:
Perhaps this should exist first as a plugin to prove its usefulness before we merge this to sbt?
I would much rather have sbt-pgp or sbt-dependency-graph feature before generating a dot file.

jvican:
> Perhaps this should exist first as a plugin to prove its usefulness before we merge this to sbt?
  
I think this feature was always meant to be provided by sbt, and Mark put all the infrastructure there in `Project.graphSettings`. The value of this plugin is that it allows to see the connections between all your task and settings. Sbt dependency graph and others already prove this feature is useful as is.

> I would much rather have sbt-pgp or sbt-dependency-graph feature before generating a dot file.
  
Sure, you can port them if you want too. I would like to have them merged into sbt.

dwijnand:
I agree this should be spun out (and iterated on) externally to sbt first, in its own plugin. And if a useful form is found, we can reconsider folding it into sbt proper.

Issue #3143, opened 2017-04-27
Memory leak when switching version


lrytz:
## steps

```
> cat build.sbt
crossScalaVersions := (0 to 6).map(i => s"2.10.$i") ++ (0 to 11).map(i => s"2.11.$i") ++ (0 to 2).map(i => s"2.12.$i")

> cat project/build.properties
sbt.version=0.13.15

> sbt crossScalaVersions +version
```

## problem

In the beginning, sbt switches through versions quickly, but it gets slower and slower, and around 2.11.8 or 2.11.9 it's stuck.

I took a heap dump in YourKit. There's 422 MB (865 MB retained size) of `scala.collection.immutable.::`, I looked at a few objects, they seem to be lists of `sbt.Init$Setting`. There's 84 MB of `sbt.Init$Setting` (440 MB retained size).

![image](https://cloud.githubusercontent.com/assets/119636/25472046/e05185b0-2b29-11e7-9b2a-19df1fd99426.png)

## notes

See also https://github.com/typesafehub/genjavadoc/pull/103

dwijnand:
Thank you very much for the very minimised reproduction, @lrytz.

dwijnand:
Btw, we've run into this problem (pretty sure it's this root cause) ourselves in sbt/zinc: https://github.com/sbt/zinc/commit/d5d59b6f4c563a5f9aa5d095295b93b912b90cf3.

jvican:
Thank you for the report @lrytz. Indeed, I knew something was wrong here when we run into https://github.com/sbt/zinc/commit/d5d59b6f4c563a5f9aa5d095295b93b912b90cf3. I've also had several related issues locally, when abusing `so` from `sbt-doge`.
  
I'll have a look at this if someone doesn't beat me to it.

dragos:
I ran into this recently as well. Most time seems to be spent in appending lists:

<img width="1252" alt="screen shot 2017-04-20 at 11 49 57" src="https://cloud.githubusercontent.com/assets/133742/25642361/348801de-2f91-11e7-82b2-0ddc9350b0dd.png">

dragos:
As a workaround we use `reload` between `+` commands, it seems to help. In my case I ran into the problem with 4 Scala versions already.

jvican:
[...]

PR #3238, opened 2017-06-01
Fix sbt/sbt#3143: Add test to confirm leak is gone

jvican:
The following commit adds the test case used to report #3143 in a big
projects with lots of cross versions and modules. The execution of the
test is instant!

### DIFF
--- a/main/src/main/scala/sbt/Cross.scala
+++ b/main/src/main/scala/sbt/Cross.scala
@@ -27,9 +27,7 @@ object Cross {
   private[sbt] def spacedFirst(name: String) = opOrIDSpaced(name) ~ any.+
 
   private case class Switch(version: ScalaVersion, verbose: Boolean, command: Option[String])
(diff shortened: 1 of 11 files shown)


dwijnand (approved):

main/src/main/scala/sbt/Cross.scala
```diff
@@ -27,9 +27,7 @@ object Cross {
   private[sbt] def spacedFirst(name: String) = opOrIDSpaced(name) ~ any.+
 
   private case class Switch(version: ScalaVersion, verbose: Boolean, command: Option[String])
-  private trait ScalaVersion {
```
eed3si9n, line 30: Could we test without change to `main/src`?
Also formatting is failing.

eed3si9n (changes requested):

eed3si9n:
Merged as https://github.com/sbt/sbt/pull/3998

PR #3242, opened 2017-06-02
Add gitter badge to README.md

BardurArantsson:


BardurArantsson:
Looks like a spurious failure. I have no idea how an update such as this would cause build failure :)

Issue #1812, opened 2015-01-18
Unify sbt shell and build.sbt syntax


eed3si9n:
The sbt 0.13 shell syntax

```
{<build-uri>}<project-id>/config:intask::key
```

and sbt 0.13 build.sbt DSL syntax discrepancy makes learning curve steep. There has been several proposal of unifying the two. 

## idea 1: improve both shell notation and build.sbt syntax so they can be unified

```
> Test/compile
> Test/run foo
> root/IntegrationTest/testOnly x.y
> projectA/IntegrationTest/compile
```

<s>

## older idea: allow build.sbt expression in sbt shell

Things to consider are scoping and passing in input task arguments.

```
> compile in Test
> (run in Test).toTask(" foo")
```

This is not going to be usable.

## appendix: making Scala REPL the sbt shell

[consoleProject](http://www.scala-sbt.org/0.13/docs/Console-Project.html) is a useful experiment on what it's like to be to integrate REPL and sbt shell (idea 0). We will have to provide all the feature we currently provide via commands:
- [ ] Multi-project support. `project foo` is a command.
- [ ] Cross build support. `++2.11.5`
- [ ] Reload support.
- [ ] Delegation of tasks `compile` should work, not `compile in Compile`.
- [ ] `.eval` should be called automatically.
- [ ] Continuous execution support. `~compile`
- [ ] `last`
- [ ] `;foo;bar;`
</s>

jsuereth:
I really like this idea.   Does this mean the Command engine disappears?  cosnoleProject is very much not compatible with it.

jsuereth:
i.e. I think you need to add notes of `<command>` working and `<last>` etc.

eed3si9n:
I don't know if we can get there in time. All I'm saying is that if people want to experiment with unified syntax, we already have `consoleProject`.

eed3si9n:
The other direction @lihaoyi suggested is implementing REPL friendly Scala notation. I just tried implementing `:` method:

``` scala
scala> class Configuration  { def `:`(task: Any): Unit = () }
defined class Configuration

scala> val Test = new Configuration
Test: Configuration = Configuration@be633d8

scala> object compile
defined object compile

scala> Test:compile
<console>:14: error: not found: type compile
       Test:compile
            ^
```

so something other than `:`.

dwijnand:
I don't think we can simultaneously change both syntaxes..

eed3si9n:
We can use slash everywhere:

``` scala
[...]

PR #3255, opened 2017-06-07
[sbt 0.13] Unify sbt shell and build.sbt syntax (unified slash syntax)

eed3si9n:
This PR is now moved to https://github.com/sbt/sbt/pull/3434

Fixes sbt/sbt#1812

This adds unified slash syntax for both sbt shell and the build.sbt DSL. Instead of the current `<project-id>/config:intask::key`, this **adds** `<project-id>/<config-ident>/intask/key` where `<config-ident>` is the Scala identifier notation for the configurations like `Compile` and `Test`. (The old shell syntax will continue to function)

This also adds a series of implicits called `ScopeSyntax` that adds `/` operators to project references, configuration, and keys such that the same syntax works in build.sbt.

### Examples

These examples work *both* from the shell *and* in build.sbt.

```scala
Global / cancelable
ThisBuild / scalaVersion
Test / test
root / Compile / compile / scalacOptions
ProjectRef(uri("file:/xxx/helloworld/"),"root")/Compile/scalacOptions
Zero / Zero / name
```

The inspect command now outputs something that can be copy-pasted:

    > inspect compile
    [info] Task: sbt.inc.Analysis
    [info] Description:
    [info] 	Compiles sources.
    [info] Provided by:
    [info] 	ProjectRef(uri("file:/xxx/helloworld/"),"root")/Compile/compile
    [info] Defined at:
    [info] 	(sbt.Defaults) Defaults.scala:326
    [info] Dependencies:
    [info] 	Compile/manipulateBytecode
    [info] 	Compile/incCompileSetup
    [info] Reverse dependencies:
    [info] 	Compile/printWarnings
    [info] 	Compile/products
    [info] 	Compile/discoveredSbtPlugins
    [info] 	Compile/discoveredMainClasses
    [info] Delegates:
    [info] 	Compile/compile
    [info] 	compile
    [info] 	ThisBuild/Compile/compile
    [info] 	ThisBuild/compile
    [info] 	Zero/Compile/compile
    [info] 	Global/compile
    [info] Related:
    [info] 	Test/compile

### notes about "Updating ..."

This PR adds `Def.displayRelativeReference` (https://github.com/sbt/sbt/pull/3255/commits/41bdf6c8489caafee46f9cd1b34e49bce0a1e7d2), which displays just the id portion of the `ProjectRef(...)`, which will be used for the "Updating.. " log:

```
hello-world> update
[info] Updating root...
[info] Resolving jline#jline;2.14.3 ...
[info] Done updating.
```


main/settings/src/main/scala/sbt/Reference.scala
```diff
@@ -78,7 +78,7 @@ object Reference {
       case LocalRootProject    => "{<this>}<root>"
       case LocalProject(id)    => "{<this>}" + id
       case RootProject(uri)    => "{" + uri + " }<root>"
-      case ProjectRef(uri, id) => "{" + uri + "}" + id
+      case ProjectRef(uri, id) => s"""ProjectRef(uri("$uri"),"$id")"""
```
jvican, line 81: Why `ProjectRef` here? This is shown a lot by `update` et al. In my opinion, `ProjectRef` is unnecessary here and it makes the `display` output longer... the same for `uri(`.

main/settings/src/main/scala/sbt/Scope.scala
```diff
@@ -99,18 +99,69 @@ object Scope {
...
+  def display(config: ConfigKey): String = guessConfigIdent(config.name) + "/"
+
+  private[sbt] val configIdents: Map[String, String] =
+    Map(
+      "it" -> "IntegrationTest",
+      "scala-tool" -> "ScalaTool",
+      "plugin" -> "CompilerPlugin"
+    )
+  private[sbt] val configIdentsInverse: Map[String, String] =
+    configIdents map { _.swap }
```
jvican, line 111: Nitpick: why `{}` for a simple operation that can be done with `()`? I think it's nicer the other way, could you write it with `()`?

main/settings/src/main/scala/sbt/Scope.scala
```diff
@@ -99,18 +99,69 @@ object Scope {
...
+  def displayMasked(scope: Scope, sep: String, mask: ScopeMask): String =
+    displayMasked(scope, sep, showProject, mask)
+
+  def displayMasked(scope: Scope, sep: String, mask: ScopeMask, showZeroConfig: Boolean): String =
+    displayMasked(scope, sep, showProject, mask, showZeroConfig)
+
   def displayMasked(scope: Scope, sep: String, showProject: Reference => String, mask: ScopeMask): String =
+    displayMasked(scope, sep, showProject, mask, false)
+
+  // unified / style introduced in sbt 0.13.16 / sbt 1.0
```
jvican: Can we add multi-line comments and Scaladoc instead of single line comments?

notes/0.13.16/unified-shell.markdown
```diff
@@ -0,0 +1,50 @@
...
+    Zero / Zero / name
+
+The inspect command now outputs something that can be copy-pasted:
+
+    > inspect compile
+    [info] Task: sbt.inc.Analysis
+    [info] Description:
+    [info]  Compiles sources.
+    [info] Provided by:
+    [info]  ProjectRef(uri("file:/xxx/helloworld/"),"root")/Compile/compile
```
jvican, line 37: A better way IMO: `(root in file:/xxx/helloworld/)/Comple/compile`. I am just trying to find a better output than `ProjectRef` and `uri`.

sbt/src/sbt-test/dependency-management/ivy-settings-multi-b/project/Build.scala
```diff
@@ -3,6 +3,11 @@ import Keys._
 
 object Build extends Build
 {
-	lazy val a = Project("a", file(".")) settings(externalIvySettings()) dependsOn(b)
-	lazy val b = Project("b", file("b")) settings(externalIvySettings( (baseDirectory in ThisBuild) / "ivysettings.xml" ))
+	lazy val a = Project("a", file("."))
+    .settings(externalIvySettings())
```
jvican: This is awkward, can you align?

sbt/src/sbt-test/project/unified/build.sbt
```diff
@@ -0,0 +1,12 @@
+lazy val root = (project in file("."))
+  .settings(
+    Global / cancelable := true,
+    ThisBuild / scalaVersion := "2.12.2",
```
jvican, line 4: Can one use `in` if they want?

jvican (changes requested):
This is neat. it is a prime candidate for the sbt-migration-rewrites.

 As I see it, `task in (configuration, project)` should be rewritten to `project / configuration / task`. The simpler case is easier. I'm happy that this rewrite is just syntactic.
  
Yesterday I was discussing with @Duhemm about this change. We made the observation that the fact that we're expressing things with `/` only does not allow us to distinguish between a task that is inside a configuration or a project just from the syntax. This is not necessarily bad, but a worthwhile observation that may confuse users. I would like @eed3si9n to elaborate on this and the proposal, and also add docs.
  
As I see it, `in` is still allowed. We should add deprecations. If we're going to be consistent, we should force the consistency.
  
I've made several requests for clarifications, but good job. My only request would be that you would add more tests around the corner cases I've mentioned before, the interplay between configurations, projects, and `intask`.
  
The rest of the scripted tests can be migrated with a rewrite the Center will work on https://github.com/scalacenter/sbt-migration-rewrites/issues/4.

main/settings/src/main/scala/sbt/Reference.scala
eed3si9n: One of the motivations behind this change is to make the output of `inspect` command roundtrip-able. Displaying `ProjectRef(uri("file:/xxx/helloworld/"),"root")`, the user can put that into either the shell or build.sbt.

It is a bit verbose, but it removes the need to guess or memorize our symbolic conventions. From the output it's clear that the project axis uses `ProjectRef` class.

main/settings/src/main/scala/sbt/Scope.scala
eed3si9n: It's a matter of taste. I prefer using `{}` for an infix expression that accepts a lambda, as it feels like it's part of a language construct. Probably a discussion we should have in another issue.

main/settings/src/main/scala/sbt/Reference.scala
jvican: I don't think it's a common scenario to copy-paste `ProjectRef` from the shell to the DSL. I wouldn't call this consistent -- you are basically outputting scala code to the shell for something that is >95% only meant to be read by the user.
  
Also, no one would write the `uri` completely with `file`, you would probably just refer to the project definition.
  
I am not convinced by this, honestly. I think we should find a better way to show project refs, at least for task and settings references. Remember that this output is shown for every `update`.

eed3si9n:
> Yesterday I was discussing with @Duhemm about this change. We made the observation that the fact that we're expressing things with `/` only does not allow us to distinguish between a task that is inside a configuration or a project just from the syntax. 

Not sure what you mean here. They are syntactically distinguishable if we take in account of the shape of identifiers. If we can assume that all configuration identifiers are capitalized, and subproject ids start with lowercase, then we will know between those two axes:

```scala
Compile / scalacOptions
projA / scalacOptions
```

You're going to have harder time with a key scoped with another task vs a key scoped with a subproject.

```scala
ornate / compile
fancy / compile
```

If we are comparing with `in` operator in build.sbt, the situation is the same:

```scala
compile in ornate
compile in fancy
```

This on its own is not really a problem until we chain them together:

```scala
compile in ornate in fancy
```

The slash syntax is less ambiguous in this case:

```scala
fancy / ornate / compile
```

sbt/src/sbt-test/dependency-management/ivy-settings-multi-b/project/Build.scala
eed3si9n: I'm guessing there are tabs in the file. I can fix it.

jvican:
> The slash syntax is less ambiguous in this case:
  
I agree it is less ambiguous. You have already proven my point with the key inside a task vs the key inside a project. 
  
> If we can assume that all configuration identifiers are capitalized, and subproject ids start with lowercase, then we will know between those two axes
  
Note that there is no syntactic restriction on the shape of configurations and users can define lowercase configurations if they wish so. So it's still worth taking it into account. You can assume otherwise, but it's theoretically possible. If we eventually do assume that, I think we should add it to the 'configuration' docs.
  
It's true that `in` is ambiguous to express the difference between key in task vs project. Because of this, I wouldn't worry too much about it. It's fair.
  
The only thing I'm concerned about is docs. I think we should clearly document this feature in the official docs (not only the changelog) and remove the outdated bits explaining the key references (located in Scopes).

eed3si9n:
> Note that there is no syntactic restriction on the shape of configurations and users can define lowercase configurations if they wish so. So it's still worth taking it into account. You can assume otherwise, but it's theoretically possible. If we eventually do assume that, I think we should add it to the 'configuration' docs.

In sbt 1.0 I'll capture the Scala identifier via macro, and make it impossible to define a lower-case starting configurations.

However, for sbt 0.13 it's actually safe because during the runtime I have _no idea_ what the Scala identifier is actually called, so I'm capitalizing it guessing that the user did the same:

```diff
+  private[sbt] def guessConfigIdent(conf: String): String =
+    configIdents.applyOrElse(conf, (x: String) => x.capitalize)
```

This could be added as a note somewhere in the doc.

main/settings/src/main/scala/sbt/Reference.scala
eed3si9n: > Remember that this output is shown for every `update`.

Perhaps `update` should use `Def.displayRelative` or similar function, that can display `root` as opposed to the full `ProjectRef`.

sbt/src/sbt-test/project/unified/build.sbt
eed3si9n: I can deprecate the use of `in` operator. I'm still not sure if we should remove in sbt 1.0 or wait till later.

main/settings/src/main/scala/sbt/Reference.scala
jvican: I don't see why we have to do this. Isn't this useless for most of the occasions. Can you please explain a common use case for which copy-pasteability is desirable for project refs? Project references are used only in exceptional cases where direct access to project definitions is not possible.
    
If we're going to optimize for a use case, we should optimize for the project definition name. This `root / Compile / scalacOptions` instead of `ProjectRef(uri("file:/xxx/helloworld/"),"root")/Compile/scalacOptions`.

sbt/src/sbt-test/project/unified/build.sbt
jvican: You mean sbt 2.0? I think we should remove it in 1.0. My rationale is the following:
  
* If we don't remove it in sbt 1.0, we cannot remove it in sbt 1.0.x. Otherwise we break compat.
* We can provide out-of-the-box syntactic rewrites so that users don't worry about this. It's just a matter of having good docs.

main/settings/src/main/scala/sbt/Reference.scala
eed3si9n: In general, I think we are in agreement that we should try to use the Scala identifier for the references (e.g. `root`) where possible. It's just that it would require additional knowledge of where the "current" project is to be able to display the relative names.

> I really don't see why we have to do this.

Why am I doing the `ProjectRef(uri("file:/xxx/helloworld/"),"root")` is simple. The general form of the sbt shell notation is [`{<build-uri>}<project-id>/config:intask::key`][1], and I wanted to make sure that the full gamut is expressible with slash syntax.

  [1]: http://www.scala-sbt.org/0.13/docs/Scopes.html#Referring+to+scoped+keys+from+the+sbt+shell

jvican: > In general, I think we are in agreement that we should try to use the Scala identifier for the references (e.g. root) where possible.
  
Can we remove the URI from the display function? I think it does not give any valuable information for most of the sbt users. It's verbose, especially for large paths (common case for most of devs). The idea would be to use only the project references (e.g. `root`).
  
I think that it's unlikely that people want this information. If they want it, they can either do `project` (which will show the uri) or `projectInfo` that will show the information of all the projects.
  
In my opinion, we should stick with what's common. And in that case that is references to project definitions. Note that this change should also be done for `{this}` and use `ThisBuild` instead. WDYT?

sbt/src/sbt-test/project/unified/build.sbt
lrytz: > We can provide out-of-the-box syntactic rewrites so that users don't worry about this

How long do you think it would take for this to be available? It would be good to test this on a lot of projects before RC1 (assuming that RC1 means feature freeze, regression fixes only).

jvican: Half a day to code it and ship it. We already have all the infrastructure, we just need to do it (and it would be really simple). See [this](https://github.com/scalacenter/sbt-migration-rewrites).

main/settings/src/main/scala/sbt/Reference.scala
ShaneDelmore: I do prefer the more concise path syntax, but if the goal is that I can inspect a setting, then copy paste one individual line to inspect or show it further, then I think that is a great idea.

sbt/src/sbt-test/dependency-management/ivy-settings-multi-b/project/Build.scala
ShaneDelmore: Shouldn't scalafmt be handling this now?

sbt/src/sbt-test/project/unified/build.sbt
ShaneDelmore: As a non-expert sbt user, this is much more obvious to me that there is a hierarchy of settings and I am setting one in a specific place in the hierarchy 👍

main/settings/src/main/scala/sbt/Reference.scala
jvican: The thing is that you can achieve the same goals with the more concise path syntax (just references to the project names).

ShaneDelmore: Agreed, I was just adding as a end user that I do like the idea of the output being copy pastable into a show/inspect whichever format is chosen.

eed3si9n: https://github.com/sbt/sbt/pull/3255/commits/41bdf6c8489caafee46f9cd1b34e49bce0a1e7d2 adds `Def.displayRelativeReference`, which displays just the id portion of the `ProjectRef(...)` for the current build. This will be used for the "Updating.. " log. See "Updating root...":

```
hello-world> update
[info] Updating root...
[info] Resolving jline#jline;2.14.3 ...
[info] Done updating.
```

jvican: I am not interested in having a new function for this. What I would like is that the implementation of the new function becomes the implementation of display, and that it is used by **all** its current users. You can read my previous messages to understand why I'm asking that. There is no reason to show by default the scala code for project reference, and we better only show just the project reference and we achieve your primary concern: copy-paste ability.

jvican: Last comment sounds harsh. I'm gonna reword in a nicer tone:

1. Why a new function?
2. Which of my previous arguments don't you agree with?
3. Why do you still this is a good idea (when most of the builds will **never** use project references specifying the URI)?

In my opinion, I'm proposing something reasonable, that addresses your main concerns and that is objectively better: more concise, shorter, clearer and natural to end users.

eed3si9n: The current shell notation covers `{<build-uri>}<project-id>/config:intask::key` and that it needs to be transcoded _without losing information_. The fact the the users never comes across non-current `ProjectRef` is simply untrue. Just start sbt on a fresh machine, you'll see it load your global plugins, and then your meta build. These need to be addressed.

The new function was added to display the `ProjectRef()` more concisely where possible. Simply removing the URL portion is not an option. You can make builds that stitches together source dependencies with the same project ids.

jvican: > The fact the the users never comes across non-current ProjectRef is simply untrue. 
  
Can you provide examples of normal builds that make use of `ProjectRef` and whose use is not source dependencies (like sbt)? My point is that no one uses `ProjectRef` in their builds, they only use reference to project definitions. If you're so sure this is not the case, prove me wrong.
  
> Just start sbt on a fresh machine, you'll see it load your global plugins, and then your meta build. These need to be addressed.
  
When you start up sbt, you do not use `Reference.display` to show the URI from which the project is loaded. A proof:
```
jvican in /data/rw/code/scala/sbt                                     [11:37:24] 
> $ sbt                                                     [±global-caching ▴▾]
[info] Loading settings from plugins.sbt ...
[info] Loading project definition from /data/rw/code/scala/sbt/project
[info] Loading settings from build.sbt ...
[info] Set current project to sbtRoot (in build file:/data/rw/code/scala/sbt/)
[info] sbt server started at 127.0.0.1:5842
> 
```
  
>  You can make builds that stitches together source dependencies with the same project ids.
  
You can do very dangerous things with sbt. It's not that sbt disallows you from doing it. If users define source dependencies with the same project ids, it's their fault. We apply the same logic with the misuse of other sbt features, even though sbt should forbid these use cases and it does not.
 
I understand your motivation not to lose information, but I'm yet to be convinced of a use case that **requires** the users to know where a project is located and that they need to copy-paste into their builds.
  
Also, you've forgotten to replace the use of `Reference.display` for `conflictWarning`:     `conflictWarning := conflictWarning.value.copy(label = Reference.display(thisProjectRef.value))`.

sjrd: I generally agree with @jvican. I would never copy-paste `ProjectRef(uri("file:/xxx/helloworld/"),"root")/Compile/scalacOptions` inside my build. I would never *write* that in my build either.

There are two things "we" want here:

* Copy-past'ability
* No loss of information whatsoever

IMO those goals are at odds. You may choose one *or* the other. Both is unreasonable. *Yes*, *if I did* copy-paste `ProjectRef(uri("file:/xxx/helloworld/"),"root")/Compile/scalacOptions` in my build, it would *work*, but it would destroy the readability and the cleanliness of my build code. I would never do that, and I certainly don't want beginners to do it because we told them they could copy-paste from the output of some task. It will confuse even more beginners, will produce builds that are even more obscure and unreadable to the uninitiated *and* the expert alike. There is no advantage whatsoever.

sjrd:
I am not sure this is a good idea. Overall, I agree that having the same syntax in the REPL and in the .sbt files is worthwhile goal. However, there are several things that are problematic with the new syntax, and the very fact that we are *changing* something.

## The reinforced impression of "hierarchy"

The slashes imply a strong impression of *hierarchy* between the three axes of a scope. This was [even mentioned by a user](https://github.com/sbt/sbt/pull/3255#discussion_r120910341). The problem is that it *isn't* a hierarchy.

First, one can specify a scope `in (project, taskKey)`, while leaving out the configuration part. This is clearly at odds with any commonly understood notion of hierarchy.

Second, [the scope delegation chain](http://www.scala-sbt.org/0.13/docs/Scope-Delegation.html), superbly documented by @eed3si9n recently, does *not* follow any reasonable notion of hierarchy either. The only rule that seems to agree with the notion of hierarchy in project/config/task is "Rule 1: Scope axis precedence". The other rules are in complete contradiction with the idea that project/config/task is a hierarchy. In fact, they even define (actual) hierarchies in each axis instead (particularly well exemplified with the parent chain of configurations, ending in `Global`). But even Rule 1 does not really follow a reasonable notion of hierarchy, as is clearly shown in Exercises C and D (section [Scope axis precedence, again](http://www.scala-sbt.org/0.13/docs/Scope-Delegation.html#Scope+axis+precedence%2C+again)).

## More symbols

I'm not afraid of symbols, but a lot of people are, and complaints about sbt using too many symbols [are still vivid](https://www.reddit.com/r/scala/comments/6eabzd/sbt_feedback_recent_changes/diagrtt/). If we're going to add more symbols, we'd better make sure their meaning is *trivial* and in full accordance with established meaning of said symbols in the rest of the world. As I explained in the previous section, `/` is definitely not the right symbol for the notion we have here.

## The problem of *change*

You may recall the discussion on the (totally unrelated) [proposal for indentation-base syntax in Scala](https://github.com/lampepfl/dotty/issues/2491). Forget about all the pros and cons of the proposed syntax itself. But look at all those comments saying that the mere fact of *changing* the syntax will have dramatic consequences on the eco-system.

I believe we should take similar considerations into account in the case of sbt. The last time we had a change of syntax was in sbt 0.13, *4 years ago*, we and are *still* paying the price of changing that syntax, because there is so much documentation and SO answers out there using the old syntax.

A new change of syntax now will have consequences lasting even longer, now that sbt has been around so long.

If we're going to unify the syntaxes of the REPL and .sbt files, I would suggest we don't touch the syntax of .sbt files, and instead only adapt the syntax of the REPL to be the same.

eed3si9n:
@sjrd Thanks for your feedback.

> ## The reinforced impression of "hierarchy"
>
> The slashes imply a strong impression of _hierarchy_ between the three axes of a scope. This was even mentioned by a user. The problem is that it _isn't_ a hierarchy.

There is a hierarchy both conceptually and implementation-wise. The scope axis precedence, including the Exercises C and D demonstrates that if the stronger axis matches (e.g. `Compile` in `Compile` vs `console`) the weaker axis becomes useless.

Another utilitarian merit for the slash syntax is the tab completion. The key parser is constructed as a series of Monad where the earlier components feed into the next part. So from the user's point of view,

    # I first type in a subproject name
    > zincFoo[tab]

    # / get completed
    > zincFoo/

    # I type in the first few characters of the key
    > zincFoo/sca[tab][tab]

    # first tab completes "scala", next tab shows more keys
    > zincFoo/scala
    scalaBinaryVersion          scalaCompilerBridgeSource   scalaHome                   scalaInstance               scalaOrganization           scalaSource                 scalaVersion
    scalacOptions

    # I change my mind and try to find what's under Test/test
    > zincFoo/Test/test/[tab]
    fullClasspath       javaHome            javaOptions         logBuffered         parallelExecution   streams             streamsManager      tags                testExecution       testGrouping
    testListeners       testOptions         testResultLogger

> ## The problem of change
>
> I believe we should take similar considerations into account in the case of sbt. The last time we had a change of syntax was in sbt 0.13, 4 years ago, we and are still paying the price of changing that syntax, because there is so much documentation and SO answers out there using the old syntax.
>
> A new change of syntax now will have consequences lasting even longer, now that sbt has been around so long.
>
> If we're going to unify the syntaxes of the REPL and .sbt files, I would suggest we don't touch the syntax of .sbt files, and instead only adapt the syntax of the REPL to be the same.

I agree with your point about the weight of documentation, SO answers, and basically collective investment made on the knowledge of the current `build.sbt`.

Since the entire goal is to reduce the cognitive load, it probably makes sense to unify towards `build.sbt` DSL even with the tab completion tradeoff.

eed3si9n:
Side note: In previous discussions about https://github.com/sbt/sbt/issues/1812 I used to make a case that unification to build.sbt would not be feasible because of tasks like `testOnly` that takes arguments, but now that I've done this PR (slash syntax allows optional spaces), I am now more optimistic about `build.sbt` syntax.

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/3259

lihaoyi:
I'm not totally sure what the issue being described here is

> I used to make a case that unification to build.sbt would not be feasible because of tasks like testOnly that takes arguments

---------------------------------

I personally prefer the slash syntax to the build.sbt syntax. For that matter, I even prefer the slash syntax *within* the build.sbt file, if we could make that work (can we?).

My opinion is that as long as we preserve backwards compatibility, we should not be worried about the existing knowledge-base of SO answers: builds will keep working, C&P will keep working. Furthermore, if we are thinking of respecting semantic versioning, 1.0.0 *is* your chance to break backwards compatibility.

I disagree with the source of confusion for the 0.7 to 0.13 migration; I do not think the *migration* was the problem. I don't think we're paying any cost for that move. What I think is confusing is the 0.13 syntax, *standing alone*: People aren't currently confusing "0.13 syntax with 0.7 syntax", people are just confused by 0.13 syntax itself. For example, I think that `:=` and `.value` have unquestionably been a success at making SBT easy to use, despite being a big change from the original 0.13 syntax, with a significant but acceptable level of confusion given the benefit.

If we believe that the 0.13 syntax is the problem, rather than the rate-of-change of syntax, then the correct thing to do isn't to avoid further change but rather to evolve away from the problematic parts of 0.13 syntax as fast as possible. 

Philosophically, I think we have to be somewhat more aggressive at evolving SBT. I think it is relatively clear that SBT's current user-experience is confusing beyond the level that more-stackoverflow-answers-and-more-books will solve. In a way, I think SBT has somewhat plateaued at how well it's going to be understood given the huge quantity of docs and answers and tribal knowledge the community has accumulated.

To make things better, I think you will have to make some bets that will follow a J-curve: short-term negative benefit due to confusion/migration, with a positive pay-off further down the road as we re-build the communal knowledge and expertise around the new, "better" practices.

Anyway, these are just my thoughts as an outsider. As the ones actually doing the work on SBT, naturally the final decision is all yours

eed3si9n:
> I personally prefer the slash syntax to the build.sbt syntax. For that matter, I even prefer the slash syntax within the build.sbt file, if we could make that work (can we?).

The slash syntax was designed to work for both shell and build.sbt. This PR includes a test file that looks like this:

```scala
  .settings(
    Global / cancelable := true,
    ThisBuild / scalaVersion := "2.12.2",
    Test / test := (),
    console / scalacOptions += "-deprecation",
    Compile / console / scalacOptions += "-Ywarn-numeric-widen",
    projA / Compile / console / scalacOptions += "-feature",
    Zero / Zero / name := "foo"
  )

lazy val projA = (project in file("a"))
```

eed3si9n:
Starting a thread on [contributors.scala-lang.org](https://contributors.scala-lang.org/t/unification-of-sbt-shell-notation-and-build-sbt-dsl/913) on this topic.

PR #3261, opened 2017-06-12
Call the soon-to-exist Global#close when done compiling

retronym:
Scala 2.13.3 is likely to have a new method, `Global.close`.
This will let the compiler release resources, notably file
handles to JARs on the classpath.

This commit calls changes the compiler interfaces to call this
method, falling back to a no-op extension method for backwards
compatibility.

Furthermore, it changes the API to the global cache to let
the caller hand back the compiler instance. By default, this
compiler cache is a no-op (`GlobalCache.fresh`), and this is
where I've added the `close` call.

The other code path does not close the Global. AFAIK, caching
global instances (enabled with `sbt -Dsbt.resident.limit=<N>`)
a unfinished experiment and and not use in the wild. For instance,
I don't see how the current implementation ensures single threaded
access to the cached global instances, I would have expected a
check-out/check-in protocol to deal with that.

In any case, that code path is untouched under this patch.

### DIFF
--- a/compile/interface/src/main/scala/xsbt/CompilerInterface.scala
+++ b/compile/interface/src/main/scala/xsbt/CompilerInterface.scala
@@ -95,6 +95,14 @@ private final class CachedCompiler0(args: Array[String], output: Output, initial
     try { run(sources.toList, changes, callback, log, dreporter, progress) }
     finally { dreporter.dropDelegate() }
   }
+  def close(): Unit = {
+    val c = compiler
+    if (c != null) {
+      class GlobalCloseCompat(g: Global) { def close(): Unit = () }
+      implicit def GlobalCloseCompat(g: Global): GlobalCloseCompat = new GlobalCloseCompat(g)
+      c.close()
+    }
+  }
   private[this] def run(sources: List[File], changes: DependencyChanges, callback: AnalysisCallback, log: Logger, dreporter: DelegatingReporter, compileProgress: CompileProgress): Unit = {
     if (command.shouldStopWithInfo) {
       dreporter.info(null, command.getInfoMessage(compiler), true)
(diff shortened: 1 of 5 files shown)


retronym:
Marking as `on hold` so we can perform an end-to-end test after the compiler change.

See scala/bug#10295 / https://github.com/retronym/scala/tree/ticket/10295

jvican:
> The other code path does not close the Global. AFAIK, caching global instances (enabled with sbt -Dsbt.resident.limit=<N>) a unfinished experiment and and not use in the wild.
  
This has been removed in Zinc 1: https://github.com/sbt/zinc/issues/340

dwijnand:
I think we should look to improve the performance of sbt 1 rather than sbt 0.13.

As such, I think we should close this.

retronym:
@jvican 

> This has been removed in Zinc 1: [sbt/zinc#340](https://github.com/sbt/zinc/issues/340)

I think the removal was incomplete.

SBT still [has logic](https://github.com/sbt/sbt/blob/fbadfabfe2ca6f200045f69c851fbb940d298229/main/src/main/scala/sbt/Main.scala#L839-L854) around `sbt.resident.limit` and Zinc still [includes](https://github.com/sbt/zinc/blob/6ef476b25226feee78a01b68e1f4f39e2c121d20/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/CompilerCache.scala#L23-L22) the cache of Globals.

jvican:
> SBT still [has logic](https://github.com/sbt/sbt/blob/fbadfabfe2ca6f200045f69c851fbb940d298229/main/src/main/scala/sbt/Main.scala#L839-L854) around `sbt.resident.limit`

Yes, though IIRC it's disabled by default.

PR #3263, opened 2017-06-13
bump to using sbt 1.0.0-M6

eed3si9n:
We are missing the plugin for Scalafmt but I think we need to use beta2 ourselves.

- https://github.com/sbt/io/pull/46
- https://github.com/sbt/util/pull/82
- https://github.com/sbt/librarymanagement/pull/114

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ matrix:
 
 env:
   matrix:
-    - SBT_CMD=";test:compile;scalafmt::test;test:scalafmt::test;mainSettingsProj/test;safeUnitTests;otherUnitTests"
+    - SBT_CMD=";test:compile;mainSettingsProj/test;safeUnitTests;otherUnitTests"
   # - SBT_CMD="mimaReportBinaryIssues"
     - SBT_CMD="scripted actions/*"
     - SBT_CMD="scripted apiinfo/* compiler-project/* ivy-deps-management/*"
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ matrix:
 
 env:
   matrix:
-    - SBT_CMD=";test:compile;mainSettingsProj/test;safeUnitTests;otherUnitTests"
+    - SBT_CMD=";test:compile;scalafmt::test;test:scalafmt::test;mainSettingsProj/test;safeUnitTests;otherUnitTests"
   # - SBT_CMD="mimaReportBinaryIssues"
     - SBT_CMD="scripted actions/*"
     - SBT_CMD="scripted apiinfo/* compiler-project/* ivy-deps-management/*"
--- a/project/Sxr.scala
+++ b/project/Sxr.scala
@@ -26,17 +26,22 @@ object Sxr {
     target := target.in(taskGlobal).value / "browse",
     sxr in taskGlobal := sxrTask.value
   )
-  def taskGlobal = ThisScope.copy(task = Global)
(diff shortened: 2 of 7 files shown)


project/plugins.sbt
```diff
@@ -5,9 +5,10 @@ scalacOptions ++= Seq("-feature", "-language:postfixOps")
...
 // addSbtPlugin("com.typesafe.sbt" % "sbt-git" % "0.9.2")
 // addSbtPlugin("com.typesafe.sbt" % "sbt-javaversioncheck" % "0.1.0")
-//addSbtPlugin("com.geirsson" % "sbt-scalafmt" % "0.7.0-RC1")
-addSbtPlugin("com.lucidchart" % "sbt-scalafmt" % "0.3")
 // addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "1.2.0")
 addSbtPlugin("org.foundweekends" % "sbt-bintray" % "0.4.0")
 addSbtPlugin("org.scala-sbt" % "sbt-contraband" % "0.3.0-M5")
 addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.0-M1")
+
+// hopefully this comes back soon.
```
jvican: Can we wait for this? This should come soon enough and the plugin is already cross-built. Worst-case scenario we can publish it ourselves temporarily.

project/Sxr.scala
```diff
@@ -26,17 +26,22 @@ object Sxr {
     target := target.in(taskGlobal).value / "browse",
     sxr in taskGlobal := sxrTask.value
   )
-  def taskGlobal = ThisScope.copy(task = Global)
+  def taskGlobal = ThisScope.copy(task = Zero)
   def sxrTask = Def task {
```
jvican, line 30: Style note: I feel that doing `Def task` is non-idiomatic and I would opt for `Def.task`. The reason why I'm nitpicking here is because our build is the best example to illustrate good practices of sbt. We use `Def.task` and `Def.setting` consistently throughout our build and ecosystem.

jvican (changes requested):
I say let's wait two days more :smile:. If neo-scalafmt is not out by then, we merge this. Otherwise, we update.

jvican:
Okay, I took the time to cross-compile neo-sbt-scalafmt: https://github.com/lucidsoftware/neo-sbt-scalafmt/pull/11. Let's hope @pauldraper merges and releases it soon and we unblock this PR.

dwijnand (dismissed):
I agree to waiting a day or two so we can continue to enforce consistent formatting.

jvican:
@pauldraper Have you released a new version for neo-sbt-scalafmt? Last time I tried I got a resolution error even though the artifact was in Centrla.

eed3si9n:
Not sure if it's announced, but I saw sbt-scalafmt 1.3 on Maven Central, so I'm using it.

.travis.yml
```diff
@@ -16,7 +16,7 @@ matrix:
 
 env:
   matrix:
-    - SBT_CMD=";test:compile;mainSettingsProj/test;safeUnitTests;otherUnitTests"
+    - SBT_CMD=";test:compile;scalafmt::test;test:scalafmt::test;mainSettingsProj/test;safeUnitTests;otherUnitTests"
```
jvican: Can you remove `test:compile`? That's not already necessary.

eed3si9n: It's merged now, so I guess you can send a new PR on it.

jvican: Done.

PR #3267, opened 2017-06-13
[sbt 0.13] Use sudo-enabled Trusty environment for more RAM

eed3si9n:


eed3si9n:
Looks like it's unable to pass `tests / fork-parallel` even after several trials. Maybe it's to do with the availability of the core.
Set aside that, the perf gain seems inconclusive at best, some faster and others much slower.

PR #3275, opened 2017-06-18
Remove unnecessary compile in Travis

jvican:
It was needed before because formatting happened on compile. Now it's
independent.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ matrix:
 
 env:
   matrix:
-    - SBT_CMD=";test:compile;scalafmt::test;test:scalafmt::test;mainSettingsProj/test;safeUnitTests;otherUnitTests"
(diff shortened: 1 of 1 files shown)


jvican:
ping @eed3si9n :smile:

dwijnand:
I kind of like it the way it is as it ensures that everything that is aggregated by the root has its main and test sources compiled.

eed3si9n:
I'm ok either way. I don't think it adds too much to the build time.

jvican:
Okay, so I'll update this task to put the scalafmt checks before compilation so that formatting is the first thing to be checked by the CI.

dwijnand:
Yeah, I'm happy to change the order.

But yeah I advise leaning towards slightly-less-lean CI validation setup code/config. From various experiences of accidentally dropping things from being validated ("who tests the tests").

PR #3184, opened 2017-05-09
Upgrade the build to use sbt 1.0.0-M5

dwijnand:
Some plugins remain commented out, for now.

The motivation for this PR is to start using and exercising sbt 1 ourselves.

PR #3281, opened 2017-06-20
Get rid of Eval

dwijnand:


eed3si9n:
Why not keep the Eval but just change the method signatures back to strict ones, and provide lazySettings etc.? As it stands build.sbt and Build.scala are not feature parity because of this reason.

See also https://github.com/sbt/sbt/issues/2104, https://github.com/sbt/sbt/issues/2698

dwijnand:
Can you show me a legitimate build that requires it?

* The Scala.js's build doesn't require it anymore (https://github.com/scala-js/scala-js/pull/2850)
* Akka's build doesn't require it, though it doesn't look like they were using lazy settings originally: https://github.com/akka/akka/pull/22968, neither do Akka HTTP and Alpakka
* ScalaTest too doesn't need it: https://github.com/scalatest/scalatest/blob/3.1.x/project/scalatest.scala.

#2698 has nothing to do with compile-time laziness, it's the DAG checker that isn't respecting the configuration hierarchy.

eed3si9n:
Here's a quick demo of what you probably can express using sbt 0.13 + Build.scala, but no longer can with sbt 1.0.x's limited `Project` constructor due to laziness:

```scala
lazy val root: Project = (project in file("."))
  .aggregate(actorProj, agentProj)
  .settings(
    inThisBuild(List(
      scalaVersion := "2.12.2",
      organization := "com.example"
    )),
    name := "Foo Root"
  )

lazy val actorProj: Project = project
  .aggregate(agentProj)

lazy val agentProj: Project = project
  .dependsOn(actorProj)
```

See also https://github.com/akka/akka/blob/c636f3540fa93c19e6d5503fb1cdfea878010fad/build.sbt#L37-L42

Setting the right `aggregate` is important for successful cross building (see recent discussions around `+` etc), and people will quickly run into cycles with those.

dwijnand:
You can express it, but using `LocalProject`:

```scala
lazy val actorProj: Project = project
  .aggregate(LocalProject("agentProj"))

lazy val agentProj: Project = project
  .dependsOn(actorProj)
```

eed3si9n:
The LocalProject workaround is stringly and confusing.

Are there any legitimate user-facing downsides of `Eval`, since it's mostly hidden.

dwijnand:
Your example is of an advanced build, with a form of cyclical dependency; sbt provides a solution with `LocalProject`, an advanced tool generally not needed.

An alternative to using the string identifier of the project would be to define a value of type `LocalProject`, and wire the build with that. Something like this perhaps:

```scala
lazy val actorProj: Project = project
  .aggregate(agentProj)

val agentProj: LocalProject = LocalProject("agentProj")
val agentProject: Project = Project(agentProj.project, file(agentProj.project))
  .dependsOn(actorProj)
```

The downsides of `Eval` are:

* it's faulty, because there's only so much you can do with implicits, varargs, etc, which sometimes don't pan out and require workarounds
* it adds more to the surface area, by introducing a bunch of additional, alternative methods
* it's unnecessary: it introduces _another_ way of doing something that can already be done with `LocalProject`, which makes sbt more complex, not less

eed3si9n:
Let's clarify for whom the build becomes complex vs simple vs easy. Using the example,

```scala
lazy val actorProj: Project = project
  .aggregate(agentProj)

lazy val agentProj: Project = project
  .dependsOn(actorProj)
```

We can say that for sbt devs it's a bit more complex because internally we have some Eval contraption, but simple for the users.
Whereas

```scala
lazy val actorProj: Project = project
  .aggregate(agentProj)

val agentProj: LocalProject = LocalProject("agentProj")
val agentProject: Project = Project(agentProj.project, file(agentProj.project))
  .dependsOn(actorProj)
```

This might simplify sbt's internal a bit, but clearly more complex and difficult for the sbt user.
Now the user needs to be aware of `LocalProject`, which as of now isn't even documented. In the shell which name should the user use to refer to the project?

Also note that `Eval` isn't about introducing new feature to sbt. We are the ones who first took Build.scala away, and I am merely reinstating the laziness. I think we need to migrate a bunch of builds to sbt 1.0.0-M6 to be sure that project cross referencing is a rare use case.

dwijnand:
Having both Eval and a bunch of additional methods and implicit conversions trying to obscure the truth, while `LocalProject` is still an alternative option, is more complex.

For cyclical situations I say we should just recommend `LocalProject` usage, simple:

```scala
lazy val actorProj: Project = project
  .aggregate(LocalProject("agentProj"))

lazy val agentProj: Project = project
  .dependsOn(actorProj)
```

If you misspell the string you get an error message during loading:

```
java.lang.RuntimeException: No project 'agenteProj' in 'file:/s/t-2017-06-28.1125/'.
Valid project IDs: actorProj, agentProj
```

> We can say that for sbt devs it's a bit more complex because internally we have some Eval contraption, but simple for the users.

You say it's all internal and simple for the users, but it's really not: we can't even avoid a source-breaking change in our own build (#3184):

```scala
-  .aggregate(nonRoots: _*)
+  .aggregateSeq(nonRoots)
```

> Also note that Eval isn't about introducing new feature to sbt. We are the ones who first took Build.scala away, and I am merely reinstating the laziness. I think we need to migrate a bunch of builds to sbt 1.0.0-M6 to be sure that project cross referencing is a rare use case.

This has nothing to do with the removal of the `Build` trait, it has to do with the removal of the `Project.apply` with the lazy parameters. How many modern builds are there that use that? I can't find a way to GitHub search it. But I've yet to find a build that won't work with `LocalProject`.

The reason we removed the laziness was for settings (interacting with AutoPlugins, core plugins), so if you're concern is lazy dependsOn/aggregate I'm happy to re-introduce those as optional lazy parameters on `Project.apply`.

jvican:
I would like to chime in and say that I agree with @dwijnand.
  
Supporting cyclic dependencies is not something I see as desirable. `LocalProject` is more than enough for it, and the complexity that `Eval` brings is not worth it. Simpler internals -> more contributions -> simpler modifications in the future -> more progress on sbt.
  
Also, let's remember that cyclic dependencies are discouraged everywhere, even in the Scala compiler. And the compiler does not support cyclic dependencies in 100% of the cases. So I think it's legit to remove such an abstruse concept as `Eval`.

eed3si9n (approved):
We can revisit the Eval in 1.x but we can't remove once it's in, so I'll merge the removal of Eval.

jvican:
I have ported away from the use of `sbt.util.Eval` successfully in akka's build.

PR #3290, opened 2017-06-26
In-source the modules of sbt/util required in sbt/sbt

dwijnand:
The subset of sbt/util was prepared like this:

1. Get rid of sbt.util.Eval (https://github.com/sbt/util/pull/83)
2. Move sbt.internal.util.HListFormats from util-cache to util-collection to break the dependency between util-cache and util-collection, so util-collection can be moved up here.
3. Move sbt.util.ShowLines out of util-collection to its own dependency to break the dependency of util-logging on util-collection, and allow util-collection can be moved up here.
4. Move sbt.internal.util.Position out of util-collection to its own dependency to break the dependency of lm on util-collection, allowing util-collection to be moved up here.
5. Run git filter-branch:

```
git filter-branch -f --index-filter 'git rm --cached -qr -- . && git reset -q $GIT_COMMIT -- util/collection util/complete util/logic internal/util-collection internal/util-complete internal/util-logic' --prune-empty
git reset --hard && git gc --aggressive && git prune
```

The I created a temporary 'before-split' branch, pointing to the commit before the first one in sbt/util, i.e a62d0dd1676658da967aa546fce81c4db726ef5b^, and then ran this git rebase:

```
git rebase -p --onto eat-util before-split detangle
```

Finally, the build was configured, and the final Eval-related changes were incorporated (https://github.com/sbt/sbt/pull/3281).


build.sbt
```diff
@@ -124,6 +124,34 @@ lazy val bundledLauncherProj =
 
 /* ** subproject declarations ** */
 
+val collectionProj = (project in file("internal") / "util-collection")
```
jvican, line 127: Happy to see this one back. I consider it fundamental and should live here :+1:

dwijnand:
See https://github.com/sbt/util/issues/58#issuecomment-312242348.

Issue #1449, opened 2014-07-13
0.13.5: autoImports always applied even if plugin not enabled


pfn:
I have an AutoPlugin that looks like the following:

```
object AndroidPlugin extends AutoPlugin {

  override def trigger = noTrigger
  override def requires = plugins.JvmPlugin

  object autoImport extends KeysContainer {
    type ProjectLayout = android.ProjectLayout
    val ProjectLayout = android.ProjectLayout

    type ProguardCache = android.ProguardCache
    val ProguardCache = android.ProguardCache
    type ProguardInputs = android.ProguardInputs
    val ProguardInputs = android.ProguardInputs

    import Dependencies.artifacts
    def apklib(m: ModuleID, n: String): ModuleID = artifacts(m, n, "apklib")
    def apklib(m: ModuleID): ModuleID            = artifacts(m, m.name, "apklib")
    def aar(m: ModuleID, n: String): ModuleID    = artifacts(m, n, "aar")
    def aar(m: ModuleID): ModuleID               = artifacts(m, m.name, "aar")

    val AutoLibraryProject = Dependencies.AutoLibraryProject
    type AutoLibraryProject = Dependencies.AutoLibraryProject

    type LibraryProject = Dependencies.LibraryProject
    val LibraryProject = Dependencies.LibraryProject
  }
}
```

I get this error when I run my scripted tests:

```
[error] C:\Users\pfnguyen\AppData\Local\Temp\sbt_ccd486fb\lib-with-resources\build.sbt:7: error: reference to platformTarget is ambiguous;
[error] it is imported twice in the same scope by
[error] import android.Keys._
[error] and import _root_.android.AndroidPlugin.autoImport._
[error] platformTarget in Android := "android-17"
[info] [error] Type error in expression
[error] ^
```

The plugin is added via addSbtPlugin(......) and the test case can be found here https://github.com/pfn/android-sdk-plugin/tree/master/sbt-test/android-sdk-plugin/lib-with-resources

I would expect that autoImport is not applied unless enablePlugins() is called explicitly. Even if addSbtPlugin is used.

pfn:
Additionally, Keys and autoImport extend from a KeysContainer trait so that users of the plugin need not enablePlugin to use the existing, pre-0.13.5 functionality.

jsuereth:
[...]

PR #3291, opened 2017-06-26
Move the tuple enrichments to sbt.TupleSyntax

### LINKED ISSUE #2762: Remove the tuple map enrichments

dwijnand:
This undeprecates the syntax, but at the same times moves it out of
implicit scope, therefore requiring a 'import Scoped.syntax._' to opt-in to
the old syntax.

Fixes #2762

### DIFF
--- a/main-settings/src/main/scala/sbt/Structure.scala
+++ b/main-settings/src/main/scala/sbt/Structure.scala
@@ -82,8 +82,10 @@ sealed abstract class SettingKey[T]
 
   protected[this] def make[S](other: Initialize[S], source: SourcePosition)(
       f: (T, S) => T
-  ): Setting[T] =
+  ): Setting[T] = {
+    import TupleSyntax._
     set((this, other)(f), source)
+  }
 }
 
 /**
@@ -143,8 +145,10 @@ sealed abstract class TaskKey[T]
 
   private[this] def make[S](other: Initialize[Task[S]], source: SourcePosition)(
       f: (T, S) => T
-  ): Setting[Task[T]] =
+  ): Setting[Task[T]] = {
+    import TupleSyntax._
     set((this, other)((a, b) => (a, b) map f.tupled), source)
+  }
 }
 
 /**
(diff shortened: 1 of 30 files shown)


main-settings/src/main/scala/sbt/Structure.scala
```diff
@@ -393,21 +406,38 @@ object Scoped {
...
-  @deprecated("The sbt 0.10 style DSL is deprecated: '(k1, k2) map { (x, y) => ... }' should now be '{ val x = k1.value; val y = k2.value }'.\nSee http://www.scala-sbt.org/0.13/docs/Migrating-from-sbt-012x.html", "0.13.13") implicit def t11ToTable11[A, B, C, D, E, F, G, H, I, J, K](t11: (ScopedTaskable[A], ScopedTaskable[B], ScopedTaskable[C], ScopedTaskable[D], ScopedTaskable[E], ScopedTaskable[F], ScopedTaskable[G], ScopedTaskable[H], ScopedTaskable[I], ScopedTaskable[J], ScopedTaskable[K])): RichTaskable11[A, B, C, D, E, F, G, H, I, J, K] = new RichTaskable11(t11)
-
-  // format: on
+  /** The sbt 0.10 style DSP was deprecated in 0.13.13, favouring the use of the '.value' macro.
+   *
+   * See http://www.scala-sbt.org/0.13/docs/Migrating-from-sbt-012x.html for how to migrate.
+   */
+  object syntax {
```
eed3si9n: If we are making this optional, I think the proper way of doing it is to turn it into a trait called `TupleSyntax`, and then define an auto plugin that introduces `TupleSyntax` as part of `autoImport`. See `trait SlashSyntax` here - https://github.com/sbt/sbt-slash/blob/b852a828bf266226a63b569c6fa9ec6178846090/src/main/scala/sbt/sbtslash/SlashSyntax.scala

eed3si9n (changes requested):
I am fine with the overall direction, but I don't think the name `Scoped.syntax` is quite right, since `Scoped` just happens to be a common trait between setting keys etc, and has nothing to do with tuple DSL. The old implementation is there more for implicit scope reason.

jvican (changes requested):
I agree with Eugene here, the naming is my only complaint. Thanks for the work on this.

main-settings/src/main/scala/sbt/Structure.scala
dwijnand: Can't use autoImport because autoImport is always imported even if the plugin isn't enabled, which is #1449.

eed3si9n (approved):

main-settings/src/main/scala/sbt/Structure.scala
eed3si9n: I just noticed this comment. We just need to make it another jar like scripted, I think.

PR #3315, opened 2017-07-13
[sbt 1.0] Better prompt out of the box

### LINKED ISSUE #3313: Better prompt out of the box
## steps

Use sbt with a multi project build.

## problem

The prompt just says `>`.

## expectation

The prompt defaults to something that tells what your current project is.

eed3si9n:
Fixes #3313

This changes the default prompt to include "sbt:" + current project name.
See https://twitter.com/eed3si9n/status/884914670351659009

dark

light

The prompt style was created by @jvican and @japgolly in https://twitter.com/japgolly/status/884945037125013504

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1746,7 +1746,8 @@ object Classpaths {
           }
           val base = ModuleID(id.groupID, id.name, sbtVersion.value).withCrossVersion(cross)
           CrossVersion(scalaVersion, binVersion)(base).withCrossVersion(Disabled())
-        }
+        },
+        shellPrompt := shellPromptFromState
       ))
 
   val ivyBaseSettings: Seq[Setting[_]] = baseGlobalDefaults ++ sbtClassifiersTasks ++ Seq(
@@ -2871,6 +2872,14 @@ object Classpaths {
             sys.error(
               "Unknown predefined resolver '" + unknown + "'.  This resolver may only be supported in newer sbt versions.")
         }
+    }
+  }
+
+  def shellPromptFromState: State => String = { s: State =>
+    val extracted = Project.extract(s)
+    (name in extracted.currentRef).get(extracted.structure.data) match {
+      case Some(name) => s"sbt:$name" + Def.withColor("> ", Option(scala.Console.CYAN))
+      case _          => "> "
     }
   }
 }
--- a/main-settings/src/main/scala/sbt/Def.scala
+++ b/main-settings/src/main/scala/sbt/Def.scala
@@ -1,7 +1,7 @@
 package sbt
 
 import sbt.internal.util.Types.const
-import sbt.internal.util.{ Attributed, AttributeKey, Init }
+import sbt.internal.util.{ Attributed, AttributeKey, Init, ConsoleAppender }
 import sbt.util.Show
 import sbt.internal.util.complete.Parser
 import java.io.File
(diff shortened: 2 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?sbt=pull/3315/head&deploy_to_bintray=false">this link.</A>

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2873,6 +2874,14 @@ object Classpaths {
         }
     }
   }
+
+  def shellPromptFromState: State => String = { s: State =>
+    val extracted = Project.extract(s)
+    (name in extracted.currentRef).get(extracted.structure.data) match {
```
dwijnand, line 2880: Interesting, never noticed the get on `DefinableSetting`. This is equivalent to `extracted getOpt (name in extracted.currentRef)`.

dwijnand (approved):

jvican (changes requested):
This is great. I very much like it.
  
I have two requests to improve your proposal:
1. Can we put an space between "sbt:project" and ">"? I would be "sbt:project > ". I think it reads better.
2. Have you considered coloring the project name?
3. Have you considered not displaying the project name if the project is the root? I think it's a little bit redundant.

eed3si9n:
I think this is a matter of preference, so I don't pretend to make a rational argument in this reply.

> 1. Can we put an space between "sbt:project" and ">"? I would be "sbt:project > ". I think it reads better.

```
sbt:hellotest> compile
[info] Total time: 37 s, completed Jul 13, 2017 2:34:45 PM
sbt:hellotest> 


sbt:hellotest > compile
[info] Total time: 37 s, completed Jul 13, 2017 2:34:45 PM
sbt:hellotest > 

```

To me, the above feels better because of the the reason you stated. I want the prompt to clump up in a ball of alphabet so I can ignore it. Putting space forces my eyes to read each pieces.

> 2. Have you considered coloring the project name?

No, I have not. We can ship that as sbt 1.1, and I am not kidding.

> 3. Have you considered not displaying the project name if the project is the root? I think it's a little bit redundant.

So that was the original proposal by @japgolly in https://twitter.com/japgolly/status/884945037125013504, right? I kind of like consistency over being clever about single vs multi build. I think it's easier to explain to users if it always displays the project name, and also I frequently have multiple tabs of builds going at the same time, so I do want the project name even for single-project builds (often for testing purposes).

That reminds me that I should credit you both for coming up with this prompt.

jvican:
> To me, the above feels better because of the the reason you stated. I want the prompt to clump up in a ball of alphabet so I can ignore it. Putting space forces my eyes to read each pieces.
  
Fair point.
  
> No, I have not. We can ship that as sbt 1.1, and I am not kidding.
  
I figured that since we're doing this now, adding color wouldn't be too much effort.
  
>  I kind of like consistency over being clever about single vs multi build. I think it's easier to explain to users if it always displays the project name, and also I frequently have multiple tabs of builds going at the same time, so I do want the project name even for single-project builds (often for testing purposes).
  
Yes, I agree it's easier to explain. I'm just afraid of having ugly root project names in my sbt shell.

eed3si9n:
Merging as is for now.

dwijnand:
> I'm just afraid of having ugly root project names in my sbt shell.

As I said in the related issue this is the upside I'm hoping from this change - less bad root names.

PR #3351, opened 2017-07-21
Allow `.value` on SettingKey inside ifs & lambdas

### LINKED ISSUE #3299: False positive in DSL checker for setting keys
## steps

```scala
val foo = taskKey[Unit]("")
val bar = settingKey[Int]("")

bar := 1
foo := (if (true) bar.value)
```

## problem

```
/s/t-2017-06-29.1446/build.sbt:5: error: The evaluation of `bar` happens always inside a regular task.

Problem: `bar` is inside the if expression of a regular task.
  Regular tasks always evaluate task inside the bodies of if expressions.

Solution:
  1. If you only want to evaluate it when the if predicate is true or false, use a dynamic task.
  2. Otherwise, make the static evaluation explicit by evaluating `bar` outside the if expression.

foo := (if (true) bar.value)
                  ^
```

## expectation

`bar` is a `SettingKey` so there's no evaluation or side-effecting.

## notes

sbt version: 1.0.0-M6

Duhemm:
Calling `.value` on a SettingKey doesn't trigger any execution and
doesn't have any side effect, so we can safely allow calls to `.value`
inside conditionals and lambdas.

Fixes #3299

### DIFF
--- a/main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
+++ b/main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
@@ -1,5 +1,6 @@
 package sbt.std
 
+import sbt.SettingKey
 import sbt.internal.util.ConsoleAppender
 import sbt.internal.util.appmacro.{ Convert, Converted, LinterDSL }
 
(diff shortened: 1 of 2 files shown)


dwijnand (approved):
Nice one, man! Thank you.

jvican (approved):
Thanks @Duhemm, was planning to do this before RC-3 and it's good that's it's off my shoulders! :smile:. LGTM.

eed3si9n:
Awesome!

Issue #3360, opened 2017-07-23
CrossUtil should calculate sbt 1.1.0-SNAPSHOT as 1.0 compat


eed3si9n:
## steps

Publish sbt locally as `1.1.0-SNAPSHOT`, and try using it.

## problem

It's unable to find sbt-pgp because it thinks 1.1.0-SNAPSHOT is not compat with 1.0.

```
[warn] 	Note: Some unresolved dependencies have extra attributes.  Check that these dependencies exist with the requested attributes.

[warn] 		com.jsuereth:sbt-pgp:1.1.0-M1 (scalaVersion=2.12, sbtVersion=1.1.0-SNAPSHOT)
```

## expectation

binarySbtVersion should be 1.0.

PR #3361, opened 2017-07-23
[bport] Fixes sbtApiVersion logic

eed3si9n:
This backports yet to be merged https://github.com/sbt/librarymanagement/pull/140

The NonRelease pattern matcher is only checking for the third segment, but for sbt 1.x, we need to check both the second and third segment since 1.1.0-M1 would be bincompat with 1.0.

Fixes sbt/sbt#3360

### DIFF
--- a/ivy/src/test/scala/CrossVersionTest.scala
+++ b/ivy/src/test/scala/CrossVersionTest.scala
@@ -60,6 +60,36 @@ object CrossVersionTest extends Specification {
     "return binary sbt version for 0.12.1 as 0.12" in {
       CrossVersion.binarySbtVersion("0.12.1") must_== "0.12"
     }
+    "return binary sbt version for 1.0.0-M6 as 1.0.0-M6" in {
+      CrossVersion.binarySbtVersion("1.0.0-M6") must_== "1.0.0-M6"
+    }
+    "return binary sbt version for 1.0.0-RC1 as 1.0" in {
+      CrossVersion.binarySbtVersion("1.0.0-RC1") must_== "1.0"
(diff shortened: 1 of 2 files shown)


ivy/src/test/scala/CrossVersionTest.scala
```diff
@@ -60,6 +60,36 @@ object CrossVersionTest extends Specification {
...
+    "return binary sbt version for 1.0.2-M1 as 1.0" in {
+      CrossVersion.binarySbtVersion("1.0.2-M1") must_== "1.0"
+    }
+    "return binary sbt version for 1.0.2-RC1 as 1.0" in {
+      CrossVersion.binarySbtVersion("1.0.2-RC1") must_== "1.0"
+    }
+    "return binary sbt version for 1.0.2 as 1.0" in {
+      CrossVersion.binarySbtVersion("1.0.2") must_== "1.0"
+    }
+    "return binary sbt version for 1.3.0 as 1.0" in {
```
eed3si9n, line 81: Prior to this patch the following tests failed:

```
[info] x return binary sbt version for 1.3.0 as 1.0
[error]    '1.3' is not equal to '1.0' (file:1)
[info]
[info] x return binary sbt version for 1.3.0-SNAPSHOT as 1.0
[error]    '1.3.0-SNAPSHOT' is not equal to '1.0' (file:1)
[info]
[info] x return binary sbt version for 1.10.0 as 1.0
[error]    '1.10' is not equal to '1.0' (file:1)
```

dwijnand (approved):

Issue #3327, opened 2017-07-17
Move sbt.protocol to sbt.internal.protocol


dwijnand:
All the classes under sbt.protocol are for internal usage only, so they should be internal.

dwijnand:
See https://github.com/sbt/sbt/pull/3366#issuecomment-318058971

PR #3366, opened 2017-07-25
Move sbt.protocol to sbt.internal.protocol

eed3si9n:
Fixes sbt/sbt#3327


eed3si9n:
I initially pushed to sbt/sbt, and cancelled it so I got the failure. On my own branch travis/push passed fine - https://travis-ci.org/eed3si9n/sbt/builds/257165587

dwijnand (approved):

dwijnand (changes requested):
Actually, let me bump MiMa, just so we can double-check the effects of this change.

dwijnand:
I don't know what I was thinking MiMa could do for us - it's not going to do usage level analysis..

So I did a little myself, and I think this change breaks our public binary API. The example I found is from the first object in the first file: `ForkTests`. It returns the test output, and nested in it is a contraband-created TestResult. By changing the package name we're removing classes that are part of the public API.

Issue #3136, opened 2017-04-23
`skip in publish`


eed3si9n:
I propose that we implement a flag named `skip in publish` to mark a subproject as non-publishing.

## steps

```scala
skip in publish := true
```

## expectation

`publish`, `publishLocal`, and `publishSigned` will opt out of publishing action.

## note

This allows me to install sbt-pgp plugin as a global plugin, but still opt out of publishing.

olafurpg:
Note. This would also be useful for sbt plugins that use synthetic projects.

dwijnand:
How is this different to `publishArtifact`?

jvican:
> How is this different to publishArtifact?
  
Same question. I've been using `publishArtifact` quite a lot to denote that something should not be published, and I'm using that to skip the execution of publishing tasks in sbt-release-early.
  
The following [SO thread](http://stackoverflow.com/questions/8786708/how-to-disable-package-and-publish-tasks-for-root-aggregate-module-in-multi-modu/37072580#37072580) mentions a misbehaviour of `publishArtifact`: "while this works it still creates and delivers ivy.xml files (at least when doing a publish-local)". This may be a difference @dwijnand.
  
I think it would be great to make the use of this option standard and suggest avoiding to disable publish tasks via `publish := {}`.
  
However, I'd like to note that I believe `publishArtifact` is not a great name for this. I prefer `skip in publish`, especially because the "skip" setting already exists. If you decide to go this route, I can provide a rewrite for it in the sbt migration rewrites. But this would force us to define what `publishArtifact` means and what we use it for. It's not clear if the misbehaviour I mentioned before is an issue until we define what `publishArtifact` is supposed to do.

eed3si9n:
See http://www.scala-sbt.org/0.13/docs/Artifacts.html for how `publishArtifact` is used.
It's used as a flag along with configuration and task scoping to denote certain artifacts are published.

```scala
// enable publishing the jar produced by `test:package`
publishArtifact in (Test, packageBin) := true

// enable publishing the test API jar
publishArtifact in (Test, packageDoc) := true
```

Changing `publishArtifact := false` at the subproject scoping simply means by default you don't want to publish something, but its power will be nullified the moment there are more scoped setting gets introduced.

That's why today we need to do `publish := (), publishSigned := ()` etc..

```
skip in publish := true
```

would carry a clear semantics that the entire operation of publishing should be skipped.

jvican:
[...]

PR #3380, opened 2017-07-28
implement `skip in publish`

eed3si9n:
`skip in publish` when set to `true` will skip publish, publishLocal, and publishM2. (sbt-pgp already skips publishSigned).

Fixes sbt/sbt#3136

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2159,16 +2159,26 @@ object Classpaths {
         classifiers.toVector
       )
     }
+
   def deliverTask(config: TaskKey[PublishConfiguration]): Initialize[Task[File]] =
     Def.task {
       val _ = update.value
       IvyActions.deliver(ivyModule.value, config.value, streams.value.log)
     }
+
   def publishTask(config: TaskKey[PublishConfiguration],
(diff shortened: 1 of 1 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2159,16 +2159,26 @@ object Classpaths {
...
     Def.task {
       val _ = update.value
       IvyActions.deliver(ivyModule.value, config.value, streams.value.log)
     }
+
   def publishTask(config: TaskKey[PublishConfiguration],
                   deliverKey: TaskKey[_]): Initialize[Task[Unit]] =
-    Def.task {
-      IvyActions.publish(ivyModule.value, config.value, streams.value.log)
+    Def.taskDyn {
```
jvican, line 2171: Note that this will make the previous direct dependencies disappear, and introduces dynamic dependencies that are not visible to `inspect`. Every time I see this I become more convinced that we need a dynamic inspect, as LAMP proposed some time ago.

jvican (approved):
LGTM.

main/src/main/scala/sbt/Defaults.scala
ches: Wow, I was just trying to make sense of the task graph for `publish` with `inspect`, now I know why I can't…

Issue #3333, opened 2017-07-18
Classloading problem in scripted plugin


larsrh:
## steps

1. Clone [sbt-slf4j](https://github.com/larsrh/sbt-slf4j), branch `topic/sbt-1`
2. Run `sbt scripted`

## problem

`[error] SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".`

As far as I can tell, there is only one reason why this would fail, which is when the file `src/main/resources/slf4j-impl-helper` doesn't end up in the classpath. It works properly when running `sbt console` instead of `sbt scripted`.

## expectation

Works as before in 0.13.x

## notes

sbt version: 1.0.0-M6

xuwei-k:
IIRC, sbt depends on slf4j since version 1.0.

larsrh:
Interesting. Why would it work in `console`, then? But maybe this plugin is obsolete, then.

xuwei-k:
- [0.13.15](https://gist.github.com/xuwei-k/1d44ad6ae1ac42d5e547eb42a4022993)
- [1.0.0-M6](https://gist.github.com/xuwei-k/907f3a5340155b42730cc4825c8a330d)
- [1.0.0-RC2](https://gist.github.com/xuwei-k/45cf610e56bec3a0ca46cf3a7d6444b2)

larsrh:
OK, but is there a `org.slf4j.impl.StaticLoggerBinder` somewhere?

eed3si9n:
Fixed in https://github.com/sbt/sbt/pull/3382

larsrh:
@eed3si9n Am I right in assuming that my plugin (`sbt-slf4j`) can be deprecated now?

eed3si9n:
Yes. Next RC of sbt 1.0 will include a SLF4J binding.

larsrh:
Perfect, thanks!

PR #3382, opened 2017-07-28
Log4j 2 SLF4J Binding

### LINKED ISSUE #1931: Add SLF4J backend
See https://github.com/sbt/sbt-git/issues/69.

eed3si9n:
Fixes sbt/sbt#3333
Fixes #1931

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -113,4 +113,11 @@ object Dependencies {
 
   val scalaXml = scala211Module("scala-xml", "1.0.6")
   val scalaParsers = scala211Module("scala-parser-combinators", "1.0.5")
+
+  def log4jVersion = "2.8.1"
+  val log4jApi = "org.apache.logging.log4j" % "log4j-api" % log4jVersion
+  val log4jCore = "org.apache.logging.log4j" % "log4j-core" % log4jVersion
+  val log4jSlf4jImpl = "org.apache.logging.log4j" % "log4j-slf4j-impl" % log4jVersion
+  // specify all of log4j modules to prevent misalignment
+  val log4jDependencies = Vector(log4jApi, log4jCore, log4jSlf4jImpl)
 }
--- a/build.sbt
+++ b/build.sbt
@@ -356,7 +356,7 @@ lazy val mainProj = (project in file("main"))
   .settings(
     testedBaseSettings,
     name := "Main",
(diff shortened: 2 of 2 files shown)


project/Dependencies.scala
```diff
@@ -113,4 +113,11 @@ object Dependencies {
 
   val scalaXml = scala211Module("scala-xml", "1.0.6")
   val scalaParsers = scala211Module("scala-parser-combinators", "1.0.5")
+
+  def log4jVersion = "2.8.1"
+  val log4jApi = "org.apache.logging.log4j" % "log4j-api" % log4jVersion
+  val log4jCore = "org.apache.logging.log4j" % "log4j-core" % log4jVersion
+  val log4jSlf4jImpl = "org.apache.logging.log4j" % "log4j-slf4j-impl" % log4jVersion
+  // specify all of log4j modules to prevent misalignment
+  val log4jDependencies = Vector(log4jApi, log4jCore, log4jSlf4jImpl)
```
dwijnand, line 122: How does this fix #3333?

eed3si9n: This adds SLF4J binding for Log4j 2. Until now we've had Log4j 2, but not the binding. The scripted test of sbt-slf4j now picks up sbt as the logger, making the plugin obsolete as @larsrh noted.

dwijnand: (The source of my question was I thought we were already doing this in sbt/librarymanagement, but instead we declare but don't use log4j-slf4j-impl there: https://github.com/sbt/util/blob/v1.0.0-M28/project/Dependencies.scala#L54)

I still don't fully understand the failure in #3333, but I think trying to fix it like this is risky. We should check what happens with this in sbt-web/sbt-play where they prefer logback as the slf4j impl.

eed3si9n: There's been https://github.com/sbt/sbt/issues/1931 and other problems relating to us _not_ having the binding by default.

dwijnand: Hmm, I guess it does make sense that sbt is an application that defines which slf4j backend to use. This will be part of the migration process for a few plugins. IMO we should highlight it somewhat in the release notes.

dwijnand (approved):

dwijnand:
I wonder what this will do to sbt-bintray which has already been cross-built for sbt 1...

https://github.com/sbt/sbt-bintray/blob/v0.5.1/build.sbt#L40

PR #3383, opened 2017-07-28
Upgrade to Scala 2.12.3

dwijnand:


jvican:
Have you checked build times?

dwijnand:
Looking at the builds for sbt/io, the build before my change: https://travis-ci.org/sbt/io/builds/258173678 and the build after my change: https://travis-ci.org/sbt/io/builds/258595634 the 2.12 build went from 3m34 to 3m21 (I'm skipping the build for my change - https://travis-ci.org/sbt/io/builds/258444546 - because it includes the time to resolve and cache 2.12.3).

jvican:
Oh, right. I've noticed that Travis has a lot of variance in build times, so local results would probably be better. I've done an informal benchmark on my own:
## Before
`[info] Total time: 74 s, completed Jul 28, 2017 7:13:30 PM`
## After
`[info] Total time: 57 s, completed Jul 28, 2017 7:07:23 PM`
  
That's a 23% improvement in compile time. Yay.

PR #3438, opened 2017-08-14
[1.0.x] `Append` instance to add `File` to `Seq[Source]`

Duhemm:
Let's people write, for instance:

```scala
watchSources += baseDirectory.value / "some-directory"
```

Instead of

```scala
watchSources += new sbt.internal.io.Source(baseDirectory.value / "some-directory", AllPassFilter, NothingFilter)
```

It also simplifies the life of people cross-building plugins between 0.13 and 1.0.

### DIFF
--- a/main-command/src/main/scala/sbt/Watched.scala
+++ b/main-command/src/main/scala/sbt/Watched.scala
@@ -6,11 +6,12 @@ package sbt
 import BasicCommandStrings.ClearOnFailure
 import State.FailureWall
 import annotation.tailrec
+import java.io.File
 import java.nio.file.FileSystems
 
 import scala.concurrent.duration._
 
-import sbt.io.WatchService
+import sbt.io.{ AllPassFilter, NothingFilter, FileFilter, WatchService }
 import sbt.internal.io.{ Source, SourceModificationWatch, WatchState }
 import sbt.internal.util.AttributeKey
 import sbt.internal.util.Types.const
@@ -18,7 +19,7 @@ import sbt.internal.util.Types.const
 trait Watched {
 
   /** The files watched when an action is run with a preceeding ~ */
-  def watchSources(s: State): Seq[Source] = Nil
+  def watchSources(s: State): Seq[Watched.WatchSource] = Nil
   def terminateWatch(key: Int): Boolean = Watched.isEnter(key)
 
   /**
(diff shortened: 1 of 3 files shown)


dwijnand (dismissed):

eed3si9n (dismissed):
`File` can mean both source files and directories. In sbt 0.13, people passed in literally list of files, correct? See for example https://github.com/sbt/sbt-protobuf/blob/v0.5.5/src/main/scala/sbtprotobuf/ProtobufPlugin.scala#L57

```scala
    watchSources ++= ((sourceDirectory in protobufConfig).value ** "*.proto").get,
```

And sbt 1.0 requires `Source`. In my opinion, adding `+=` that passes in a _directory_ is significantly _worse_ because now old code will pass compilation with broken semantics.

eed3si9n:
Here's my suggestion:

Make an alias named `sbt.WatchSource` for `sbt.io.internal.Source`. Define `apply(dir: File)` in the companion object.

Duhemm:
@eed3si9n Sorry, I should have given an example with a file. It would work just the same. The watch service works with files *and* directories (which added quite some complexity). The example you gave works fine.

Trigger on any change inside `some-directory`:
```scala
watchSources += sourceDirectory.value / "some-directory"
```

Trigger only on changes inside any of the given files (i.e. modifying `foo.java` won't trigger)
```scala
watchSources += (sourceDirectory.value ** "*.scala").get
```

eed3si9n:
> It would work just the same. The watch service works with files and directories (which added quite some complexity).

Thanks for the clarification.

If there are no difference in efficiency, would it then make more sense to only support the file usage via `+=`, in terms of keeping the same semantics as before? That would probably match the expectation of the term "source" meaning source code, not a _source_ of source code.

eed3si9n:
For old usage (strict, same as 0.13):

```scala
watchSources ++= (sourceDirectory.value ** "*.scala").get
```

For new usage (lazy):

```scala
watchSources += WatchSource(sourceDirectory.value) // or WatchSource(sourceDirectory.value, filter, filter) 
```

(`sbt.WatchSource` would be alias of sbt.io.internal.Source)

eed3si9n:
For this PR, we should be able to retarget to 1.0.1 and merge.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?sbt=pull/3438/head&deploy_to_bintray=false">this link.</A>

Duhemm:
I've addressed your comments. I put the alias `WatchSource` in `Watched` because I thought it made sense to have it there, let me know if you'd prefer to have it somewhere else.

eed3si9n:
Sorry for the inaction here! I got a bit distracted on other tasks, and we still need https://github.com/sbt/sbt/pull/3456 to get in so we can cut 1.0.x branch.

eed3si9n:
Merging to 1.0.x.

PR #3459, opened 2017-08-17
Make developers key consistent with other definitions

marconilanna:
`developers` is the only key defined as `List[T]`, while all others were defined as `Seq[T]`. This inconsistency causes confusion, as can be seen [here](https://github.com/sbt/sbt/pull/1590#issuecomment-260228079) and [here](https://github.com/sbt/sbt/issues/2107#issue-95280890).

It is good to remind us that it is generally considered best practice to program to an interface (`Seq`), not an implementation (`List`), something I believe was overlooked when the original  #1590 PR was accepted.

Since `List[T]` is a subtype of `Seq[T]`, making the type more general does not break backward compatibility, as can be easily attested by sbt's own [build.sbt](https://github.com/sbt/sbt/blob/1.x/build.sbt#L21).

~~This PR also partially reverts #2114, given such changes aren't needed anymore. The tests that append a `List` to a `SettingKey[Seq[T]]` were left, though, to attest backward compatibility is indeed preserved.~~ (breaks binary compatibility)


lightbend-cla-validator:
Hi @marconilanna,

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?sbt=pull/3459/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
A dbuild validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/115/console">is in progress...</a>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?sbt=pull/3459/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
[//]: # (Refs: pull/3459/head-1.x-1.x-1.x-1.x-1.x)
dbuild <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/115/">has checked</A> the following projects against Scala 2.12:

Project             | Reference            | Commit
:------------------ | :------------------- | :----------------
  sbt               | <A HREF="https://github.com/sbt/sbt/pull/3459">pull/3459/head      </A> | sbt/sbt@3b1f7e8107cfd407bbcf594b6de0cf7a430b2775
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/1.x">1.x                 </A> | sbt/zinc@2c18bccc366d199f88a35c62e3f683d00168e877
  io                | <A HREF="https://github.com/sbt/io/tree/1.x">1.x                 </A> | sbt/io@539a2162e151d5e64feed135d2816ab40a808349
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/1.x">1.x                 </A> | sbt/librarymanagement@9abaa9f7407cc0f0e53bedc1348bdc1887efb888
  util              | <A HREF="https://github.com/sbt/util/tree/1.x">1.x                 </A> | sbt/util@771a52f651e0a8ea9f87fb8e389ab3684d33af95
  website           | <A HREF="https://github.com/sbt/website/tree/1.x">1.x                 </A> | 

:x: The result is: ***FAILED***
<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild?sbt=pull/3459/head&zinc=1.x&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false">(restart)</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?sbt=pull/3459/head&deploy_to_bintray=false">this link.</A>

dwijnand:
Hi @marconilanna! Welcome and thank you for opening the pull request.

While I agree it's an outlier, is there any continuing confusion that hasn't been resolved since the inclusion of the Append instances?

> It is good to remind us that it is generally considered best practice to program to an interface (Seq), not an implementation (List), something I believe was overlooked when the original #1590 PR was accepted.

This general topic is very interesting to me because it's not as clean-cut in the Scala ecosystem as it is in the Java ecosystem. One reason (e.g see Daniel Spiewak's [Collections Redesign Proposal](https://gist.github.com/djspiewak/2ae2570c8856037a7738) gist) is that you have no idea of what the performance characteristics of any particular Seq is compared to List.

Another benefit of List over Seq is that List is known to be immutable, while you have no information about Seq; which is why in sbt 1's coding guidelines we recommend to ["Use specific datatypes (Vector, List, or Array), rather than Seq"](http://www.scala-sbt.org/1.x/docs/Coding-Guideline.html#Use+specific+datatypes+%28%2C+%2C+or+%29%2C+rather+than).

Unless there are other difficulties or confusion due to the use of List I don't think we should weaken `developers` to Seq.

eed3si9n:
Keys are identified by their names and types, so I'd be super careful with messing with the type.
For example, if sbt-house-rules was compiled against 1.0.0, would it still work with 1.1.0?

dwijnand:
I'm fairly sure it would be fine. sbt-houserules would create runtime a `Setting` instance using the new definition of `developers` and everything would continue to work.

The edge-case that could arise (but who the hell would do this) is if there was code that pattern-matched `developers.value` into Nil and :: it'll start failing at runtime because now someone defined it with `Array`.

marconilanna:
hi @dwijnand,

append (`+=`, `++=`) is only part of the problem. many people (myself included, hence the motivation for this pr) will try adding `developers := Seq(Developer(...` to their build.sbt and wonder why it didn't work. granted, sbt will tell you `found   : Seq[sbt.Developer] required: List[sbt.Developer]`, but still.

while i agree with the 1.0 coding guidelines and i think it is, in general, good practice to be followed, please notice that the "Use specific datatypes" section is under "Other public *API* matters". i'm not sure whether or not it was intended to also cover key definition.

for instance, other than `Seq`, we also have keys defined with `Map` and `Set`, all of which are interfaces, not specific datatypes. together with `Option`, they are the only collection types used in Keys.scala, meaning there are currently no other keys using a specific datatype.

so, i ask you, what's the plan moving forward? should we forbid new keys being defined as `Seq` and standardize on `List` and `Vector`? (as a matter of fact, just recently we had a [pr to revert from `Vector` back to `Seq`](https://github.com/marconilanna/sbt/commit/99be5b25f49c7660c0c7e23e064fdfbd6ef1f611#diff-46751185767542b1d052bc42435916bd))

as you said, `developers` is an outlier: all other keys are defined as `Seq`, and the only reason i see for the inconsistency is that it was overlooked when the original pr was accepted. fixing it now would bring more consistency (less confusion for users) and simplicity (we could revert #2114 if we wanted)

typesafe-tools:
A dbuild validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/116/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/3459/head-1.x-1.x-1.x-1.x-1.x)
dbuild <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/116/">has checked</A> the following projects against Scala 2.12:

Project             | Reference            | Commit
:------------------ | :------------------- | :----------------
  sbt               | <A HREF="https://github.com/sbt/sbt/pull/3459">pull/3459/head      </A> | sbt/sbt@e184f7c67a4c800b93352cc023730d07684d6a84
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/1.x">1.x                 </A> | sbt/zinc@2c18bccc366d199f88a35c62e3f683d00168e877
  io                | <A HREF="https://github.com/sbt/io/tree/1.x">1.x                 </A> | sbt/io@539a2162e151d5e64feed135d2816ab40a808349
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/1.x">1.x                 </A> | sbt/librarymanagement@9abaa9f7407cc0f0e53bedc1348bdc1887efb888
  util              | <A HREF="https://github.com/sbt/util/tree/1.x">1.x                 </A> | sbt/util@771a52f651e0a8ea9f87fb8e389ab3684d33af95
  website           | <A HREF="https://github.com/sbt/website/tree/1.x">1.x                 </A> | 

:x: The result is: ***FAILED***
<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild?sbt=pull/3459/head&zinc=1.x&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false">(restart)</A>

cunei:
Please note that the validator is currently broken, see https://github.com/sbt/sbt-standalone-build/issues/11

dwijnand:
> so, i ask you, what's the plan moving forward? should we forbid new keys being defined as Seq and standardize on List and Vector?

That would be my vote, yes.

marconilanna:
is performance really so critical for a collection that typically holds just a couple, at most a few dozen items?

dwijnand:
No, not performance. But the fact that it's immutable is an improvement over Seq, imo.

eed3si9n:
@marconilanna wrote:

> while i agree with the 1.0 coding guidelines and i think it is, in general, good practice to be followed, please notice that the "Use specific datatypes" section is under "Other public API matters". i'm not sure whether or not it was intended to also cover key definition.

In a general situation, I am a "use specific datatype" guy, and at the time when I proposed the Developers guide I probably was leaning towards that for `Keys.scala` as well.

*However*, given that we've long used `Seq[A]` in build.sbt, at this point I am actually thinking the right thing to do for Keys.scala is to unify towards `Seq`. For example, a bunch of keys almost became `Vector[String]` in sbt 1, but I rolled it back (https://github.com/sbt/sbt/commit/99be5b25f49c7660c0c7e23e064fdfbd6ef1f611) because it would have created annoying amount of changes to everyone's `build.sbt` file.

marconilanna:
@eed3si9n @cunei Are there any current known issues with TravisCI? Tests pass locally:

```
$ sbt test
...
[info] Passed: Total 63, Failed 0, Errors 0, Passed 63
[success] Total time: 117 s, completed 13-Oct-2017 3:11:14 PM
```

eed3si9n:
I've restarted the jobs. Let's see how it goes.

marconilanna:
@eed3si9n tests pass :-) do you think we should consider this pr, considering your last [comment](https://github.com/sbt/sbt/pull/3459#issuecomment-324246083)?

eed3si9n:
@marconilanna My stance hasn't changed much since my first comment:

> Keys are identified by their names and types, so I'd be super careful with messing with the type.
For example, if sbt-house-rules was compiled against 1.0.0, would it still work with 1.1.0?

In other words, I don't want to break existing builds or plugins without really good reason.
Let's say if someone wrote a plugin that says:

```scala
foo := {
  (List("Special thanks to") ::: developers.value).mkString(" ")
}
```

This would break at runtime right?

marconilanna:
@eed3si9n Maybe I misunderstood what you meant by:

> given that we've long used Seq[A] in build.sbt, at this point I am actually thinking the right thing to do for Keys.scala is to unify towards Seq

eed3si9n:
So my point is:

- In general, `Seq[_]` might be a good idea in Keys, even though our developer guideline says we should prefer concrete types.
- For `developers` key specifically, changing key type might not be not a good idea unless proven to be safe (bincompat for plugins etc).

PR #3481, opened 2017-08-28
Revert "Automatically choose proper compiler bridge for dotty"

eed3si9n:
This reverts commit 2283c68031732751b2488d3571f3bf68a6694178, which I forward ported in https://github.com/sbt/sbt/pull/3070

@dotta pointed out to me that:
1. sbt 1 shouldn't use `"component"` configuration for the compiler bridge.
2. The bridge for Dotty is not yet available.

Given that the module name is used for sbt 0.13, the original forward porting was a bad idea.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?sbt=pull/3481/head&deploy_to_bintray=false">this link.</A>

PR #3513, opened 2017-09-11
Don't warn on by-name keys

### LINKED ISSUE #3299: False positive in DSL checker for setting keys

dwijnand:
.. which is apparently what the type of bare vals in *.sbt files are.

Fixes #3299, again.

Also set version to 1.0.2-SNAPSHOT and fix file warnings.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,7 +9,7 @@ def buildLevelSettings: Seq[Setting[_]] =
   inThisBuild(
     Seq(
       organization := "org.scala-sbt",
-      version := "1.0.1-SNAPSHOT",
+      version := "1.0.2-SNAPSHOT",
       description := "sbt is an interactive build tool",
       bintrayOrganization := Some("sbt"),
       bintrayRepository := {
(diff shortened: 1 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=pull/3513/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

jvican (approved):
LGTM.

eed3si9n (approved):
> .. which is apparently what the type of bare vals in *.sbt files are.

That's something new I learned :)

PR #3531, opened 2017-09-15
Added name property to Command trait

### LINKED ISSUE #3297: Expose useful information about available commands
IntelliJ Scala plugin exposes some information about settings, tasks and commands to the user.

For commands, the API offers only limited ways to give information, especially since the Command trait doesn't have a "name" field. So I hack into it with some reflection and using sbt-internal API:

https://github.com/JetBrains/sbt-structure/blob/master/extractor/src/main/scala-sbt-0.13-1.0/sbt/jetbrains/BadCitizen.scala

used in

https://github.com/JetBrains/sbt-structure/blob/8ba00a43c89eabadfc3af6f54813bf20d8633588/extractor/src/main/scala-sbt-0.13-1.0/org/jetbrains/sbt/extractors/KeysExtractor.scala#L64-L71

A way to access a name for all commands meant for users would make this feature more helpful in general.

jpastuszek:
Fixes #3297

This changes `SimpleCommand` property name from `name` to `_name` so it may brake some hacks that use it. Use added `name: Option[String]` `Command` trait property instead.

### DIFF
--- a/main-command/src/main/scala/sbt/Command.scala
+++ b/main-command/src/main/scala/sbt/Command.scala
@@ -16,6 +16,7 @@ import sbt.internal.util.{ AttributeKey, AttributeMap, Util }
  * Typically you would resort to a command when you need to do something that's impossible in a regular task.
  */
 sealed trait Command {
+  def name: Option[String]
   def help: State => Help
   def parser: State => Parser[() => State]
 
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:
Hi @jpastuszek,

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=pull/3531/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

main-command/src/main/scala/sbt/Command.scala
```diff
@@ -16,6 +16,7 @@ import sbt.internal.util.{ AttributeKey, AttributeMap, Util }
  * Typically you would resort to a command when you need to do something that's impossible in a regular task.
  */
 sealed trait Command {
+  def name: Option[String]
```
dwijnand, line 19: Let's add this with an extension method defined in the `Command` object. That way we don't have to change SimpleCommand.

dwijnand (changes requested):

main-command/src/main/scala/sbt/Command.scala
jvican: I think this is indeed a superior way to do it 👍

jvican:
@jpastuszek Thank you for doing this, I'm sure @jastice will be happy to see this PR. :tada:
  
@eed3si9n or @dwijnand: would you kindly label this PR as spree please?

jastice:
Cool, thanks!

dwijnand:
Duplicate of #3671

Issue #3309, opened 2017-07-07
Make `~` fail if command is not found


jvican:
This is a ticket for the Scala Center spree at Scala Wave (Gdansk, Poland).
  
The goal of this ticket is to fix this "behaviour" that should fail when the task is not found.

## steps
Run `~myNonExistingTask`.

## problem
The `~` command waits for user input even if `myNonExistingTask` does not exist.

## expectation
  
It should fail straight away without entering in the watch cycle. That means no user input is expected, normal error is reported to the user.

## notes

sbt version: 1.0.0 (master branch)

jvican:
@eed3si9n Can you label this as `spree`?

perbellinioa:
I've done this ticket during the Dublin Scala Spree.

eed3si9n:
See my comments in https://github.com/sbt/sbt/pull/3532#discussion_r140157900 onwards. The parser is intentionally kept relaxed so it can compose with other commands.

dwijnand:
Fixed in https://github.com/sbt/sbt/pull/4335/commits/28aa1de32a1108e8b42cf7606c962df6e5dc6f52 looks like?

eatkins:
Yes. #3884 is related but not yet fixed.

PR #3532, opened 2017-09-15
Continuous execution fail if command name is not defined

perbellinioa:
Fixes [#3309](https://github.com/sbt/sbt/issues/3309)

This changes make continuous execution fail straight away without entering in the watch cycle.

### DIFF
--- a/main-command/src/main/scala/sbt/Watched.scala
+++ b/main-command/src/main/scala/sbt/Watched.scala
@@ -124,7 +124,6 @@ object Watched {
         printIfDefined(watched triggeredMessage newWatchState)
         (ClearOnFailure :: next :: FailureWall :: repeat :: s).put(ContinuousState, newWatchState)
       } else {
-
         while (System.in.available() > 0) System.in.read()
         service.close()
         s.remove(ContinuousState)
--- a/main-command/src/main/scala/sbt/Watched.scala
+++ b/main-command/src/main/scala/sbt/Watched.scala
@@ -85,12 +85,12 @@ object Watched {
   def isEnter(key: Int): Boolean = key == 10 || key == 13
   def printIfDefined(msg: String) = if (!msg.isEmpty) System.out.println(msg)
 
-  def isCommandDefined(s: State, commandName: String) = s.definedCommands.exists {
-    case c: SimpleCommand => c.name == commandName
-    case _                => false
-  }
-
   def executeContinuously(watched: Watched, s: State, next: String, repeat: String): State = {
+    def isCommandDefined(s: State, commandName: String): Boolean = s.definedCommands.exists {
+      case c: SimpleCommand => c.name == commandName
+      case _                => false
+    }
+
     @tailrec def shouldTerminate: Boolean =
       (System.in.available > 0) && (watched.terminateWatch(System.in.read()) || shouldTerminate)
     val sources = watched.watchSources(s)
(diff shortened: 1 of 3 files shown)


lightbend-cla-validator:
Hi @perbellinioa,

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=pull/3532/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

perbellinioa:
I signed the CLA.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=pull/3532/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=pull/3532/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

jvican (approved):
👍 Good work! I'm really happy that you got this done.
  
@dwijnand Would you kindly label this PR with spree?

perbellinioa:
Can someone restart the PR the failing test seems to pass locally?
Thanks!

dwijnand (dismissed):
LGTM. Nicely done, and welcome to sbt.

main-command/src/main/scala/sbt/Watched.scala
```diff
@@ -86,6 +86,11 @@ object Watched {
   def printIfDefined(msg: String) = if (!msg.isEmpty) System.out.println(msg)
 
   def executeContinuously(watched: Watched, s: State, next: String, repeat: String): State = {
+    def isCommandDefined(s: State, commandName: String): Boolean = s.definedCommands.exists {
+      case c: SimpleCommand => c.name == commandName
+      case _                => false
```
eed3si9n, line 91: How would this work for "act" command, a command that represents all of your tasks, like `compile`?
If you run this with `~compile` what do you see?

I am seeing

```
sbt:helloworld> ~compile
[error] Not a valid command: compile (similar: completions)
```

eed3si9n (changes requested):

main-command/src/main/scala/sbt/Watched.scala
perbellinioa: Hi @eed3si9n ,
Good catch!
When I run `~compile` I see exactly the same error message than you.
Do you have any ideas of how I could check as well if the string after the `~` is a task ("act" command).

eed3si9n: I am not sure if this is possible, but the closest thing that I can think of is how `+` cross building command is implemented here - https://github.com/sbt/sbt/blob/v1.0.2/main/src/main/scala/sbt/Cross.scala#L50-L54. Compare that to how `~` is implemented in https://github.com/sbt/sbt/blob/v1.0.2/main-command/src/main/scala/sbt/BasicCommands.scala#L210-L217 as:

```scala
  def continuous: Command =
    Command(ContinuousExecutePrefix, continuousBriefHelp, continuousDetail)(otherCommandParser) {
      (s, arg) =>
        withAttribute(s, Watched.Configuration, "Continuous execution not configured.") { w =>
          val repeat = ContinuousExecutePrefix + (if (arg.startsWith(" ")) arg else " " + arg)
          Watched.executeContinuously(w, s, arg, repeat)
        }
    }
```

If you want to make `~` more strict, then what you want to do is make `otherCommandParser` more strict. Current implementation of `otherCommandParser` accepts any:

```scala
  lazy val otherCommandParser: State => Parser[String] =
    (s: State) => token(OptSpace ~> combinedLax(s, NotSpaceClass ~ any.*))
```

so you can try replacing it with imple that uses `state.combinedParser` only, but I am not sure if there are other negative implication if you did that, so you'd have to chase the code and figure it out.

eed3si9n: > but I am not sure if there are other negative implication if you did that, so you'd have to chase the code and figure it out.

Maybe it's the command precedence with other commands. My guess is that `~` uses a relaxed parser intentionally so it can be combined with multi-command (`;`) and cross command (`+`). For example, currently you can write:

```
> ~ ;+compile;show version
```

so sbt is probably saying that for tab completion, I am going to show known commands, but when the user puts in combined commands like the above, it's just going to execute it and see what happens.

jvican: Any chance you can have a look at this @perbellinioa :smile:?

PR #3551, opened 2017-09-16
`sbt.server.autostart` flag and startServer

eed3si9n:
Adds JVM flag `sbt.server.autostart` to enable/disable the automatic starting of sbt server with the sbt shell.

This also adds a new command `startServer` to manually start the server.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=pull/3551/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

dwijnand:
I don't think this was a good idea seeing as we created `oldshell` for exactly this use case.

PR #3566, opened 2017-09-21
Fix #3564: Filter scripted tests based on project/build.properties

### LINKED ISSUE #3564: Filter scripted tests based on optional project/build.properties
## steps

In a cross-built sbt plugin project, `scripted` will run all tests under `src/sbt-test`.

## problem

As an sbt plugin author, there are several situations where it would be good to be able to conditionally run a scripted tests:

 - A plugin has not been cross published for sbt 1.0 yet.
 - The plugin version is different for sbt 0.13 and sbt 1.0 (e.g. tut)
 - Certain scripted tests track the minimum required sbt versions, e.g. 0.13.9 and 1.0.0-RC2.

The scripted plugin allows to pass which tests to run, however, this requires additional configuration of the form (e.g. [sbt-site](https://github.com/sbt/sbt-site/blob/7563a35d1fedf5dcfb9b01c28a34937b3269c558/build.sbt#L64-L86))

```
TaskKey[Unit]("runScriptedTest") := Def.taskDyn {
  (sbtBinaryVersion in pluginCrossBuild).value match {
    case "0.13" => Def.task(scripted.toTask("..."))
    case "1.0"  => Def.task(scripted.toTask("..."))
  }
}.value
```

## expectation

For all scripted tests in which `project/build.properties` exist, the value of the `sbt.version` property is read. If its binary version is different from `sbtBinaryVersion in pluginCrossBuild` the test will be skipped and a message indicating this will be logged.

## notes

sbt version: 0.13.16, 1.0.2
PoC: https://github.com/scalacenter/scalajs-bundler/pull/179/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR148

jonas:
Skip scripted tests where the binary version configured in the build
does not match the binary version of sbt used for cross-building.

References: #3564 
Version against 0.13: https://github.com/jonas/sbt/tree/scripted-skip-0.13

### DIFF
--- a/sbt/src/sbt-test/project/scripted-skip-incompatible/test
+++ b/sbt/src/sbt-test/project/scripted-skip-incompatible/test
@@ -1,0 +1,17 @@
+# Setup failing test that should be skipped when tests are discovered
+$ copy-file changes/Fail.scala src/sbt-test/group/skipped/changes/Fail.scala
+$ copy-file changes/build.properties src/sbt-test/group/skipped/project/build.properties
+$ copy-file changes/test src/sbt-test/group/skipped/test
+
+# Should fail when run explicitly
+-> scripted group/skipped
+
+# Should be skipped when discovered
+> scripted
+
+# Setup the same test without a project/build.properties file
+$ copy-file changes/Fail.scala src/sbt-test/group/not-skipped/changes/Fail.scala
+$ copy-file changes/test src/sbt-test/group/not-skipped/test
+
+# Running discovered tests should fail
+-> scripted
--- a/scripted/sbt/src/main/scala/sbt/test/ScriptedTests.scala
+++ b/scripted/sbt/src/main/scala/sbt/test/ScriptedTests.scala
@@ -6,6 +6,7 @@ package sbt
 package test
 
 import java.io.File
+import java.util.Properties
 
 import scala.util.control.NonFatal
 import sbt.internal.scripted._
@@ -343,7 +344,9 @@ class ScriptedRunner {
           launchOpts: Array[String],
           prescripted: File => Unit): Unit = {
     val runner = new ScriptedTests(resourceBaseDirectory, bufferLog, bootProperties, launchOpts)
-    val allTests = get(tests, resourceBaseDirectory, logger) flatMap {
+    val sbtVersion = bootProperties.getName.dropWhile(!_.isDigit).dropRight(".jar".length)
+    val accept = isTestCompatible(resourceBaseDirectory, sbtVersion) _
+    val allTests = get(tests, resourceBaseDirectory, accept, logger) flatMap {
       case ScriptedTest(group, name) =>
         runner.singleScriptedTest(group, name, prescripted, logger)
     }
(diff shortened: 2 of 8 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=pull/3566/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

sbt/src/sbt-test/project/scripted-skip-incompatible/test
```diff
@@ -0,0 +1,17 @@
+# Setup failing test that should be skipped when tests are discovered
+$ copy-file changes/Fail.scala src/sbt-test/group/skipped/changes/Fail.scala
+$ copy-file changes/build.properties src/sbt-test/group/skipped/project/build.properties
+$ copy-file changes/test src/sbt-test/group/skipped/test
+
+# Should fail when run explicitly
+-> scripted group/skipped
```
dwijnand, line 7: A scripted test running scripted, how meta. I love it 😄

scripted/sbt/src/main/scala/sbt/test/ScriptedTests.scala
```diff
@@ -399,17 +402,44 @@ class ScriptedRunner {
     reportErrors(tests.flatMap(test => test.apply().flatten.toSeq).toList)
   }
 
+  @deprecated("No longer used", "1.0.2")
```
dwijnand: Let's make the version in deprecated here (and through out) 1.1.0.

dwijnand (approved):
Looks great to me. Thanks @jonas. I have 2 requests. 1 is the deprecation version string. The other is could you add some notes, please? See https://github.com/sbt/sbt/blob/1.x/CONTRIBUTING.md#adding-notes. Thank you.

sbt/src/sbt-test/project/scripted-skip-incompatible/test
jonas: Copied from `sbt/src/sbt-test/project/scripted-plugin`. I had to wrap my head around it a few times to understand what was going on, but I think it makes perfect sense in this case to test with and without the `build.properties` file.

scripted/sbt/src/main/scala/sbt/test/ScriptedTests.scala
jonas: 👍

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=pull/3566/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

jonas:
@dwijnand Fixed the deprecation version and added release notes. Should I create a PR for the 0.13 branch as well?

dwijnand:
If you'd like this in a 0.13 release then yes - I personally don't mind.

jonas:
Hmm, I guess I could just upgrade builds to sbt 1.0. 😁

PR #3600, opened 2017-10-03
Add Scala syntax from daltonjorge/vscode-scala

eed3si9n:
This adds Scala syntax for VS Code. The current plugin was relying on daltonjorge/vscode-scala for detecting that `*.scala` is a Scala source. This change integrates the plugin into this one.

daltonjorge/vscode-scala is licensed under MIT license.


eed3si9n:
Skipping CI since there's no code to validate here.

PR #3602, opened 2017-10-04
Add, configure & enforce file headers

dwijnand:


build.sbt
```diff
@@ -39,6 +39,13 @@ def buildLevelSettings: Seq[Setting[_]] =
 
 def commonSettings: Seq[Setting[_]] =
   Seq[SettingsDefinition](
+    headerLicense := Some(HeaderLicense.Custom(
+      """|sbt
+         |Copyright 2011 - 2017, Lightbend, Inc.
+         |Copyright 2008 - 2010, Mark Harrah
+         |Licensed under BSD-style license (see LICENSE)
```
eed3si9n: I suggest we use [SPDX id](https://spdx.org/licenses/) and say:

> Licensed under BSD-3-Clause license (see LICENSE)

since there are many variants of BSD-style.

eed3si9n (changes requested):

eed3si9n (approved):

PR #3603, opened 2017-10-05
Fixes Zero / Zero / Zero / name handling

eed3si9n:
Prior to this change `Zero / Zero / Zero / name` broke as follows:

```scala
scala> Zero / Zero / Zero / name
Zero / Zero / Zero / name
<console>:18: error: inferred type arguments [sbt.Zero.type] do not conform to method /'s type parameter bounds [K <: sbt.SlashSyntax.Key[K]]
       Zero / Zero / Zero / name
                   ^
```

This fixes the bug, and also adds `toString` so I can test it from REPL:

```scala
scala> import sbt._
scala> import Keys._

scala> Zero / Zero / Zero / name
res0: sbt.SlashSyntax.ScopeAndKey[sbt.SettingKey[String]] = Zero / Zero / Zero / name
```


dwijnand (approved):
Good catch, well done.

eed3si9n:
One of virtues of compiled documentation and actually writing docs before shipping features.

PR #3606, opened 2017-10-05
Make slash syntax constructs eligible for .value enrichments

### LINKED ISSUE #3605: (Test / bar).value
## steps

```scala

  .settings(
    foo := {
      (Test / bar).value + 1
    },
    Compile / bar := 1
  )

```

## problem

```
[error]                    ^
```

dwijnand:
The implicit conversions in Def that add the .value, .evaluated,
.taskValue, .inputTaskValue and .parsed methods enrich Initialize, which
ScopeAndKey wasn't a subtype of.

Because of the lack of, what I believe is called, "kind polymorphism" the
fix requires an implementation for each of the 3 key type kinds.

Notably missing is the .previous enrichment, which is only on TaskKeys, so
I'm deferring the fix for .previous on slash syntax to a later date.

Fixes #3605


eed3si9n (changes requested):
I was able get to rid of the thunks in https://github.com/sbt/sbt/pull/3613

dwijnand:
@eed3si9n I've incorporated your fix for .previous in #3611 here.

PR #3610, opened 2017-10-05
Introduce ScopeFilterReference

dwijnand:
This removes ThisProject as a ProjectReference, and defines an additional
sealed hierarchy of references for ScopeFilter, which includes ThisProject.

This fixes Scope's resolveProjectBuild and resolveProjectRef, that no
longer mishandle ThisProject.

Builds on #3609

### DIFF
--- a/main-settings/src/main/scala/sbt/Reference.scala
+++ b/main-settings/src/main/scala/sbt/Reference.scala
@@ -14,8 +14,10 @@ import sbt.io.IO
 
 // in all of these, the URI must be resolved and normalized before it is definitive
 
+sealed trait ScopeFilterReference
+
 /** Identifies a project or build. */
-sealed trait Reference
+sealed trait Reference extends ScopeFilterReference
 
 /** A fully resolved, unique identifier for a project or build. */
 sealed trait ResolvedReference extends Reference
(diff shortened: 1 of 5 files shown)


main-settings/src/main/scala/sbt/Reference.scala
```diff
@@ -47,7 +51,7 @@ final case class RootProject(build: URI) extends ProjectReference
 final case object LocalRootProject extends ProjectReference
 
 /** Identifies the project for the current context. */
-final case object ThisProject extends ProjectReference
+final case object ThisProject extends ScopeFilterProjectReference
```
eed3si9n, line 54: Is this binary compatible?

dwijnand: No. But what do you think of the idea?

eed3si9n: Is this trying to save `ThisProject`, or push it into scope filtering only? I guess I'm not seeing the point.

dwijnand: It's trying to fix Scope's `resolveProjectBuild` and `resolveProjectRef` to no
longer mishandle `ThisProject`.

But to do so it's removing `ThisProject` as a legitimate member of the `ProjectReference` sealed family. Thinking about it over-night I think it is a legitimate member, should remain so, and the real fix for 2.0.0 is to change resolveProjectBuild and resolveProjectRef.

PR #3596, opened 2017-10-02
Cleanup and improve the unified slash syntax

dwijnand:

PR #3611, opened 2017-10-05
Make scope path syntax eligible for `.value` and `.previous`

### LINKED ISSUE #3605: (Test / bar).value

eed3si9n:
This is a retake of https://github.com/sbt/sbt/pull/3606

Since #3606 brings back the three key types, I first reverted "Cleanup and improve the unified slash syntax" (#3596), and then moved the thunk classes into `Def` where it should probably belong. This allowed me to implement handling for `.previous` in the macro by injecting `.materialize` at compile-time.

I've also taken the liberty to rename "slash syntax" to "scope path" for documentation purpose.

Finally I've copied over the example class created by Dale.

Fixes #3605


eed3si9n:
I have a better idea that can simplify this further.

PR #3613, opened 2017-10-05
Make scope path syntax eligible for `.value` and `.previous` 

### LINKED ISSUE #3605: (Test / bar).value

eed3si9n:
This is a retake on #3606 and #3611

This simplifies the `/` implementation by removing the unnecessary thunks for scope and keys, because the keys already have scopes in them. In other words, when we get something like `Test / compile / name`, it will now treat as `(Test / compile) / name` where `Test / compile` now returns materialized `TaskKey[Analysis]`. This means no more hacks to get `.value` or `.previous` working.

Fixes #3605

### DIFF
--- a/main-settings/src/main/scala/sbt/ScopePathSyntax.scala
+++ b/main-settings/src/main/scala/sbt/ScopePathSyntax.scala
@@ -7,12 +7,11 @@
 
 package sbt
 
-import java.io.File
 import sbt.librarymanagement.Configuration
 import sbt.internal.util.AttributeKey
 
 /**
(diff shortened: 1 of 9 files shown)


dwijnand (changes requested):
Please don't revert my commit. There are a number of improvements in it that are being needlessly discarded.

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/3620

PR #3615, opened 2017-10-06
Print npm output, handle exit code in case of errors

cunei:
If npm returned an error, the marker "updated" was created anyway, preventing `update` from being executed again. Also, npm failures were silent.
This pull request handles npm failures a bit better, printing diagnostic output and catching the exit code.


eed3si9n (approved):

dwijnand (approved):

dwijnand:
Thanks Toni.

PR #3620, opened 2017-10-06
Remove thunk for slash syntax

### LINKED ISSUE #3605: (Test / bar).value

eed3si9n:
Ref #3606, #3611, and #3613

This removes unnecessary thunk for slash syntax.
The semantics using this approach is strictly better than the previous `in (ref, config, task)`. By removing the thunk, we retain `(a / b) / c == a / b / c`.

See the following example:

```scala
scala> import sbt._, Keys._
scala> val t: TaskKey[Unit] = (test in Test)
t: sbt.TaskKey[Unit] = TaskKey(This / Select(ConfigKey(test)) / This / test)

scala> ThisBuild / t
ThisBuild / t
res1: sbt.TaskKey[Unit] = TaskKey(Select(ThisBuild) / Select(ConfigKey(test)) / This / test)

scala> ThisBuild / t / name
ThisBuild / t / name
res2: sbt.SettingKey[String] = SettingKey(Select(ThisBuild) / Select(ConfigKey(test)) / Select(test) / name)
```

so far so good? Now look at this:

```scala
scala> name in (ThisBuild, t)
name in (ThisBuild, t)
res3: sbt.SettingKey[String] = SettingKey(Select(ThisBuild) / This / Select(test) / name)
```

`Test` configuration knowledge is lost! For `in (..)` maybe it was ok because mostly we don't use unscoped keys, but that's the difference between `in (..)` and `/`.

Fixes #3605

### DIFF
--- a/main-settings/src/main/scala/sbt/SlashSyntax.scala
+++ b/main-settings/src/main/scala/sbt/SlashSyntax.scala
@@ -39,8 +39,8 @@ trait SlashSyntax {
 
   implicit def sbtSlashSyntaxRichConfigKey(c: ConfigKey): RichConfiguration =
     new RichConfiguration(Scope(This, Select(c), This, This))
-  implicit def sbtSlashSyntaxRichConfiguration(c: Configuration): RichConfiguration =
-    sbtSlashSyntaxRichConfigKey(c: ConfigKey)
+
+  implicit def sbtSlashSyntaxRichConfiguration(c: Configuration): RichConfiguration = (c: ConfigKey)
 
   implicit def sbtSlashSyntaxRichScope(s: Scope): RichScope = new RichScope(s)
 
--- a/main-settings/src/main/scala/sbt/SlashSyntax.scala
+++ b/main-settings/src/main/scala/sbt/SlashSyntax.scala
@@ -39,22 +39,30 @@ trait SlashSyntax {
 
   implicit def sbtSlashSyntaxRichConfigKey(c: ConfigKey): RichConfiguration =
     new RichConfiguration(Scope(This, Select(c), This, This))
-
-  implicit def sbtSlashSyntaxRichConfiguration(c: Configuration): RichConfiguration = (c: ConfigKey)
-
-  implicit def sbtSlashSyntaxRichScopeFromScoped(t: Scoped): RichScope =
-    new RichScope(Scope(This, This, Select(t.key), This))
+  implicit def sbtSlashSyntaxRichConfiguration(c: Configuration): RichConfiguration =
+    sbtSlashSyntaxRichConfigKey(c: ConfigKey)
 
   implicit def sbtSlashSyntaxRichScope(s: Scope): RichScope = new RichScope(s)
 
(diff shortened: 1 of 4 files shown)


dwijnand:
```
scala> ThisBuild / t / name
ThisBuild / t / name
res2: sbt.SettingKey[String] = SettingKey(Select(ThisBuild) / Select(ConfigKey(test)) / Select(test) / name)
```

This means: define the `name` key, in `ThisBuild` scope and in `t` task-scope. It shouldn't be using `t` configuration scope.

eed3si9n:
If `t` it scoped to `Test` it makes sense to respect t's existing scope. "To scope" means to override with non-`This`.

dwijnand:
When `t` is used in task-scope position it should only be setting the task-scope axis.

eed3si9n:
> When `t` is used in task-scope position it should only be setting the task-scope axis.

That's not referentially transparent. Here's another example:

```scala
scala> val t: TaskKey[Unit] = Test / test
val t: TaskKey[Unit] = Test / test
t: sbt.TaskKey[Unit] = TaskKey(This / Select(ConfigKey(test)) / This / test)

scala> t / streams
t / streams
res2: sbt.TaskKey[sbt.Keys.TaskStreams] = TaskKey(This / Select(ConfigKey(test)) / Select(test) / streams)
```

This is going to be identical to when you replace `t`:

```scala
scala> Test / test / streams
Test / test / streams
res3: sbt.TaskKey[sbt.Keys.TaskStreams] = TaskKey(This / Select(ConfigKey(test)) / Select(test) / streams
```

For 99% of the times, this will not show up since most people will use just the unscoped `TaskKey`, but this implementation is an improvement over treating `t` as just a holder for an attribute key.

dwijnand:
It is referentially transparent:

```scala
scala> val t = Test / test
val t = Test / test
t: sbt.SlashSyntax.ScopeAndTaskKey[Unit] = This / Select(ConfigKey(test)) / This / test

scala> t / streams
t / streams
res0: sbt.TaskKey[sbt.Keys.TaskStreams] = TaskKey(This / Select(ConfigKey(test)) / Select(test) / streams)

scala> Test / test / streams
Test / test / streams
res1: sbt.TaskKey[sbt.Keys.TaskStreams] = TaskKey(This / Select(ConfigKey(test)) / Select(test) / streams)
```

dwijnand (changes requested):
The referential transparency/implicit conversion is concerning. Not having semantic parity "in" syntax is also concerning. I've submitted https://github.com/eed3si9n/sbt/pull/3 that adds slash support for "bare" AttributeKeys, which have parity with "in". I think that covers all concerns. At some point I'd like to add some negative tests to check that you can't, for example, call `Global / Compile / scalacOption`.

dwijnand (approved):
Some minor fixes in https://github.com/eed3si9n/sbt/pull/4. But this LGTM to merge.

PR #3621, opened 2017-10-08
implement window/logMessage

eed3si9n:
This sends sbt's log message as [window/logMessage](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#window_logMessage) event to Language Server client.

### DIFF
--- a/main/src/main/scala/sbt/internal/CommandExchange.scala
+++ b/main/src/main/scala/sbt/internal/CommandExchange.scala
@@ -23,9 +23,10 @@ import scala.util.{ Success, Failure }
 import sbt.io.syntax._
 import sbt.io.Hash
 import sbt.internal.server._
-import sbt.internal.util.{ StringEvent, ObjectEvent, ConsoleOut, MainAppender }
+import sbt.internal.langserver.{ LogMessageParams, MessageType }
+import sbt.internal.util.{ StringEvent, ObjectEvent, MainAppender }
 import sbt.internal.util.codec.JValueFormats
-import sbt.protocol.{ EventMessage, Serialization, ChannelAcceptedEvent }
+import sbt.protocol.{ EventMessage, Serialization }
 import sbt.util.{ Level, Logger, LogExchange }
 
 /**
@@ -72,7 +73,7 @@ private[sbt] final class CommandExchange {
 
   def run(s: State): State = {
     consoleChannel match {
-      case Some(x) => // do nothing
+      case Some(_) => // do nothing
       case _ =>
         val x = new ConsoleChannel("console0")
         consoleChannel = Some(x)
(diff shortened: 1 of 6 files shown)


main/src/main/scala/sbt/internal/CommandExchange.scala
```diff
@@ -167,21 +168,35 @@ private[sbt] final class CommandExchange {
...
   def publishEvent[A: JsonFormat](event: A): Unit = {
+    val broadcastStringMessage = true
     val toDel: ListBuffer[CommandChannel] = ListBuffer.empty
     event match {
       case entry: StringEvent =>
+        val params = toLogMessageParams(entry)
         channels.foreach {
           case c: ConsoleChannel =>
-            if (entry.channelName.isEmpty || entry.channelName == Some(c.name)) {
+            if (broadcastStringMessage) {
```
dwijnand, line 178: Was this for debug purposes? Is `broadcastStringMessage` meant to be wired up to something, rather than be a constant true?

laughedelic: I think the way it was before is correct: it should take into account channel name, otherwise things get messed up..

main/src/main/scala/sbt/internal/CommandExchange.scala
```diff
@@ -167,21 +168,35 @@ private[sbt] final class CommandExchange {
...
+                c.publishEvent(event)
+              }
             }
           case c: NetworkChannel =>
             try {
-              if (entry.channelName == Some(c.name)) {
-                c.publishEvent(event)
+              // Note that language server's LogMessageParams does not hold the execid,
+              // so this is weaker than the StringMessage. We might want to double-send
+              // in case we have a better client that can utilize the knowledge.
```
laughedelic, line 189: I think `execId` doesn't make much sense for LSP notifications, the `channelName` is much more important:
- notifications are not necessarily caused by some request, so they don't need `execId` (which in case of ResponseMessages is the request ID)
- it's important that notifications go to the right channel. So if something is triggered by an LSP client, console channel shouldn't get spammed by the log messages and another way round.

main/src/main/scala/sbt/internal/CommandExchange.scala
eed3si9n: This was done on purpose, so on terminal client you still see the compiler warning etc coming from the VS Code client etc.

main/src/main/scala/sbt/internal/CommandExchange.scala
eed3si9n: If there are n events coming out of m commands, you need execId to distinguish them from the other.

main/src/main/scala/sbt/internal/CommandExchange.scala
laughedelic: > compiler warning etc coming from the VS Code client

I didn't get this. What do you mean? Isn't is just about messages from sbt?

main/src/main/scala/sbt/internal/CommandExchange.scala
laughedelic: My point is that LSP doesn't distinguish notifications by exec-id. It does distinguish responses, but the problem here is that I see some `StringEvents` sent to the client with empty exec-ids, because they are sent as responses to a request that client didn't make.

PR #3597, opened 2017-10-03
Add system property to revert to old polling fs watcher

stringbean:
This adds a `sbt.watch.mode` system property that if set to `polling` will use PollingWatchService instead of WatchServiceAdapter (nio).

This is a temporary workaround for users affected by #3527

Issue #3527, opened 2017-09-13
Time lag before file watch (triggered execution) detects changes on Mac


eed3si9n:
There's been a number of reports on Gitter around this topic. For example:

@fsvehla wrote [Aug 31 03:52](https://gitter.im/sbt/sbt?at=59a7c02781cd9a5d7e9d3e98):

> SBT takes a long time (3, 4 seconds) to notice source code changes changes — with and without changing pollInterval, how can I start to debug this?
I’ve switched to having the sources watched by Guard and type 'testQuick' into the SBT console with tmux, but that’s not a real solution. I’m on macOS in case there are some OS differences with the new NIO watchers
>
> Yes, I’m on 1.0.1

This is a known issue since we don't have native NIO support for macOS. It would be good if someone could write something up that is licensed under Apache v2.

eed3si9n:
Maybe we switch internally to use https://github.com/takari/directory-watcher?

/cc @Duhemm

tpolecat:
Yeah for me it's taking about 10x longer to see file changes. I'm on sbt 1.0.2 and MacOS.

jastice:
intellij-community contains some C code for mac and linux for this purpose which is apache licensed:

https://github.com/JetBrains/intellij-community/tree/master/native/fsNotifier

spockz:
It would be nice if we could get a jar that loads a native solution so that `FileSystem.newWatchService()` returns an optimised one. I'm not sure whether the lib is that pluggable though.

AtkinsChang:
I try to write some in JNI with same interface. 
https://github.com/AtkinsChang/watch-service

Custom `FileSystem.newWatchService()` is pluggable.
Java is able to config default filesystem class with property `java.nio.file.spi.DefaultFileSystemProvider`.
But if you want to extends` MacOSXFileSystem`, the library should be load in bootclasspath with rt.jar. Otherwise you need to write your own `FileSystem`.
http://docs.oracle.com/javase/8/docs/technotes/guides/io/fsp/filesystemprovider.html

RatanRSur:
seeing the same on arch linux, sbt 1.0.0

alissapajer:
Also seeing the same on ubuntu 16.04, sbt 1.0.0.

djspiewak:
[...]

PR #3626, opened 2017-10-10
[1.0.x] Add system property to revert to old polling fs watcher

eed3si9n:
This is a backport of https://github.com/sbt/sbt/pull/3597 to 1.0.x branch

This adds a sbt.watch.mode system property that if set to 'polling' will
use PollingWatchService instead of WatchServiceAdapter (nio).

On macOS this will default to 'polling' and on all others 'nio'.

This is a temporary workaround for users affected by #3527

### DIFF
--- a/main-command/src/main/scala/sbt/Watched.scala
+++ b/main-command/src/main/scala/sbt/Watched.scala
@@ -3,18 +3,19 @@
  */
 package sbt
 
-import BasicCommandStrings.ClearOnFailure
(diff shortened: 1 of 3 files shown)


dwijnand (approved):

fsvehla:
Is this part of a release yet? This would allow me to use SBT 1.0

eed3si9n:
Not yet. Hopefully soon.

PR #3635, opened 2017-10-12
 Escape imports from sbt files

### LINKED ISSUE #3464: literal `import` as project id will cause sbt unable to set.
## steps
1. create simple build.sbt as
```
lazy val `import` = (project in file("."))
  .settings(
    name := "Hello",
    scalaVersion := "2.12.3"
  )
```
2. set some key like `set version := "1.0.0", REPL or command line argument will do

## problem
set command will fail with message:
```
[info] Set current project to Hello (in build file:/mnt/f/repo/sample/)
<set>:0: error: identifier expected but 'import' found.
import $1d63e54508e5abdf4728.import
                             ^
[error] Error parsing imports for expression.
```

## expectation
key set to new value without any error.

## notes
tested on sbt version: 0.13.12/15/16 and 1.0.0, all versions failed.

PanAeon:
Escape imports from sbt files, so if user creates a backquoted definition then task evalution will not fail.
Fixes #3464

### DIFF
--- a/main/src/main/scala/sbt/internal/LoadedSbtFile.scala
+++ b/main/src/main/scala/sbt/internal/LoadedSbtFile.scala
@@ -61,7 +61,7 @@ private[sbt] final class DefinedSbtValues(val sbtFiles: Seq[compiler.EvalDefinit
       file <- sbtFiles
       m = file.enclosingModule
       v <- file.valNames
-    } yield s"import ${m}.${v}"
+    } yield s"import ${m}.`${v}`"
   }
   def generated: Seq[File] =
     sbtFiles flatMap (_.generated)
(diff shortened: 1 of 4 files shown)


dwijnand (changes requested):
Nice @PanAeon, thank you!

Could you add a note with reference to the fix please? https://github.com/sbt/sbt/blob/1.x/CONTRIBUTING.md#adding-notes

Also could you add a little scripted test testing the fix, please?

PanAeon:
Hi, @dwijnand I have added notes and some scripted tests, could you review the changes?

dwijnand (approved):
Looks great, thank you.

PR #3639, opened 2017-10-13
[WIP] Add Scala Outline view to vscode

kiritsuku:
This adds an Outline view to the explorer view of vscode. The goal is to display semantic information about Scala code like library dependencies of a project or the structure of nested Scala declarations.

There is a reload button on the outline view to refresh the data in the outline view.

The missing feature is that right now no real data is retrieved from the server because I couldn't find out so far how to run a sbt task on the server (like `dependencyClasspath`) and get back its result in a way that it can be sent back to the client together with the request id.

Review by @eed3si9n, @dwijnand, @wpopielarski. Maybe you have an idea on how to solve the above issue?


eed3si9n:
> The missing feature is that right now no real data is retrieved from the server because I couldn't find out so far how to run a sbt task on the server (like dependencyClasspath) and get back its result in a way that it can be sent back to the client together with the request id.

`sbt/exec` is meant to be that.

kiritsuku:
Ok, thanks. `sbt/exec` works fine now. I finally figured out how to use it. ;) I also created a task that can send the data back to vscode. Looks better now. I'll update the PR soon.

dwijnand:
@kiritsuku There's now a merge conflict here. Could you look at that, please?

@eed3si9n How do you feel about this change now?

eed3si9n:
We might want to salvage UI elements later, but I am not sure if we need/want outline view.

PR #3650, opened 2017-10-20
Add and use kind-projector

dwijnand:


ShaneDelmore:
That’s some pretty nice improvements to readability 👍

internal/util-collection/src/main/scala/sbt/internal/util/INode.scala
```diff
@@ -32,27 +32,25 @@ abstract class EvaluateSettings[Scope] {
...
-      case o: Optional[s, T] @unchecked =>
-        o.a match {
-          case None    => constant(() => o.f(None))
-          case Some(i) => single[s, T](transform(i), x => o.f(Some(x)))
-        }
-      case x if x == StaticScopes =>
-        strictConstant(allScopes.asInstanceOf[T]) // can't convince scalac that StaticScopes => T == Set[Scope]
-    }
+  private[this] val transform: Initialize ~> INode = λ[Initialize ~> INode] {
+    case k: Keyed[s, A1$] @unchecked => single(getStatic(k.scopedKey), k.transform)
```
eed3si9n, line 36: What's this dollar sign mean here?

eed3si9n (changes requested):
A minor comment.

internal/util-collection/src/main/scala/sbt/internal/util/INode.scala
dwijnand: `A1$` is the name of the generated type parameter of `~>`'s apply, which was `T` before.

Basically this is what I found works that allows use to use polymorphic lambda syntax _even_ in places where we need the type param. It's more on the risky side of things in this diff. Btw there are 2 cases in InputTask.scala where I couldn't find any way to use the syntax, so it uses old school anonymous inner classes.

internal/util-collection/src/main/scala/sbt/internal/util/Attributes.scala
```diff
@@ -196,11 +196,9 @@ object AttributeMap {
   def apply(entries: AttributeEntry[_]*): AttributeMap = empty ++ entries
 
   /** Presents an `AttributeMap` as a natural transformation. */
-  implicit def toNatTrans(map: AttributeMap): AttributeKey ~> Id = new (AttributeKey ~> Id) {
-    def apply[T](key: AttributeKey[T]): T = map(key)
-  }
-
+  implicit def toNatTrans(map: AttributeMap): AttributeKey ~> Id = λ[AttributeKey ~> Id](map(_))
```
eed3si9n, line 199: I am perfectly happy to go with uber fp polyfunction _value level_ λ notation, but I am really not sure if this increases the "readability."

eed3si9n (approved):

internal/util-collection/src/main/scala/sbt/internal/util/Attributes.scala
dwijnand: Yeah.. I'm happy to see the reduction in boilerplate, but I'm still not 100% convinced.

PR #3657, opened 2017-10-20
Enable parallel execution of scripted in the plugin

jvican:
Enable parallel execution of scripted in the plugin
    
The change to enable batched and parallel execution for scripted was done
only for the scripted-sbt project. This pull request enables it for
scripted-plugin, so that all sbt plugins in 1.x. can benefit from it.
    
By default, it configures a number of parallel instances of 1 and batch
execution is disabled. Users can change the number of parallel sbt hosts
running scripted tests via the `scriptedParallelInstances` setting.
    
In some plugins scripted tests', batch execution can cause issues because
the first time `>` commands are executed they assume sbt starts up. This
error can be fixed by doing `reload` before running the `>` command.
    
Note that the current scripted plugin does not allow parallel execution
in non-batched mode.

### DIFF
--- a/scripted/plugin/src/main/scala/sbt/ScriptedPlugin.scala
+++ b/scripted/plugin/src/main/scala/sbt/ScriptedPlugin.scala
@@ -23,6 +23,11 @@ object ScriptedPlugin extends AutoPlugin {
     val scriptedBufferLog = SettingKey[Boolean]("scripted-buffer-log")
     val scriptedClasspath = TaskKey[PathFinder]("scripted-classpath")
     val scriptedTests = TaskKey[AnyRef]("scripted-tests")
+    val scriptedBatchExecution =
+      settingKey[Boolean]("Enables or disables batch execution for scripted.")
+    val scriptedParallelInstances =
+      settingKey[Int](
+        "Configures the number of scripted instances for parallel testing, only used in batch mode.")
     val scriptedRun = TaskKey[Method]("scripted-run")
     val scriptedLaunchOpts = SettingKey[Seq[String]](
       "scripted-launch-opts",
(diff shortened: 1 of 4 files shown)


jvican:
To be sure this does not cause any problem in the future, I have tested that scripted runs correctly with both batch execution enabled and disabled in `sbt-native-packager`.

jvican:
@eed3si9n Can you please review this?

eed3si9n (changes requested):
Overall LGTM, but I think this should target 1.x branch since it's a feature addition.

scripted/plugin/src/main/scala/sbt/ScriptedPlugin.scala
```diff
@@ -68,15 +75,18 @@ object ScriptedPlugin extends AutoPlugin {
...
-                                           classOf[Array[String]],
-                                           classOf[File],
-                                           classOf[Array[String]],
-                                           classOf[java.util.List[File]])
-  )
+  private[this] val fCls = classOf[File]
+  private[this] val bCls = classOf[Boolean]
+  private[this] val asCls = classOf[Array[String]]
+  private[this] val lfCls = classOf[java.util.List[File]]
+  private[this] val iCls = classOf[Integer]
```
dwijnand, line 82: Any particular reason these aren't defined inside `scriptedRunTask`?

IIUC there's no computation done for these; they're just class literals.

scripted/sbt/src/main/scala/sbt/test/ScriptedTests.scala
```diff
@@ -368,6 +368,21 @@ class ScriptedRunner {
...
 
+  // This is used by sbt-scripted sbt 1.x
+  def runInParallel(
+      baseDir: File,
+      bufferLog: Boolean,
+      tests: Array[String],
+      bootProps: File,
+      launchOpts: Array[String],
+      prescripted: java.util.List[File],
+      instances: Integer
```
dwijnand, line 379: Why Integer and not Int?

dwijnand (approved):
LGTM. Just a couple of minor questions.

dwijnand:
Rebased in #3891.

PR #3659, opened 2017-10-21
Adds "reboot dev"

### LINKED ISSUE #3590: An alternative to reboot full
This for ease of hacking on sbt:

## steps

```
> reboot full
```

## problem

This wipes out the boot directory, so I'd avoid this.

## expectation

Provide:

```
> reboot dev
```

This deletes only the directory under boot that is relevant to the current -SNAPSHOT version.

eed3si9n:
This adds a new option `dev` to the `reboot` command, which deletes the only the current sbt artifacts from the boot directory. `reboot dev` reads actively from `build.properties` instead of using the current state since `reboot` can restart into another sbt version.

In general, `reboot dev` is intended for the local development of sbt.

Fixes #3590

### DIFF
--- a/main-command/src/main/scala/sbt/BasicCommandStrings.scala
+++ b/main-command/src/main/scala/sbt/BasicCommandStrings.scala
@@ -117,14 +117,17 @@ $HelpCommand <regular expression>
 
   def RebootCommand = "reboot"
   def RebootDetailed =
-    RebootCommand + """ [full]
+    RebootCommand + """ [dev | full]
 
 	This command is equivalent to exiting sbt, restarting, and running the
 	  remaining commands with the exception that the JVM is not shut down.
 
-	If 'full' is specified, the boot directory (`~/.sbt/boot` by default)
-	  is deleted before restarting.  This forces an update of sbt and Scala
-	  and is useful when working with development versions of sbt or Scala."""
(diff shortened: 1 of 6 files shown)


main-command/src/main/scala/sbt/BasicCommandStrings.scala
```diff
@@ -117,14 +117,17 @@ $HelpCommand <regular expression>
...
+    RebootCommand + """ [dev | full]
 
 	This command is equivalent to exiting sbt, restarting, and running the
 	  remaining commands with the exception that the JVM is not shut down.
 
-	If 'full' is specified, the boot directory (`~/.sbt/boot` by default)
-	  is deleted before restarting.  This forces an update of sbt and Scala
-	  and is useful when working with development versions of sbt or Scala."""
+	If 'dev' is specified, the current sbt artifacts from the boot directory
+	  (`~/.sbt/boot` by default) is deleted before restarting.
```
dwijnand: are deleted

dwijnand (approved):
LGTM. Just a typo.

main-command/src/main/scala/sbt/BasicCommandStrings.scala
eed3si9n: Fixed.

PR #3660, opened 2017-10-22
textDocument/definition implementation for VS Code.

wpopielarski:
Trial of implementation of textDocument/definition request. Supports types only, and only in case when type is found in Zinc analysis. When source(s) is found then editor opens potential source(s). This simple implementation don't use semantic data (yet).

- [x] move `vscDefinition` task to separate file
- [x] add tests


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1387,6 +1390,91 @@ object Defaults extends BuildCommon {
     }
     analysisResult.analysis
   }
+  lazy val previousAnalysis = Def.inputKey[String]("gets previous analysis")
+  lazy val previousAnalysisParser = Def.setting {
+    Parsers.StringBasic
+  }
+  lazy val previousAnalysisCommand = Command.single("previousAnalysis") { (state, definition) =>
```
eed3si9n: Why are we adding both an input key and a command here?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1387,6 +1390,91 @@ object Defaults extends BuildCommon {
...
+    }
+    state.log.warn(s"previous analysis: $previousCompile")
+    val cmdSrc = state.currentCommand.flatMap(_.source)
+    import sbt.internal.langserver.{ Location, Position, Range }
+    val params = Location(input.textDocument.uri, Range(Position(0, 0), Position(0, 10)))
+    val cname = cmdSrc.map(_.channelName).get
+    val chnl = StandardMain.exchange.channels.collectFirst {
+      case c if c.name == cname => c
+    }
+    chnl.foreach(_.publishEvent(params, responseId))
```
eed3si9n: If possible, we should use event logging here.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1387,6 +1390,91 @@ object Defaults extends BuildCommon {
...
+        state.log.warn("and there")
+        results.map(_.value)
+      }
+      r
+    }.getOrElse(Seq.empty)
+    state.log.warn(s"aggregated analysis $aggrAnal")
+
+    val analysis = {
+      val compile = Keys.compile in Compile
+      Project.runTask(compile, updatedState, true).collect {
```
eed3si9n: Is this invoking compilation task just to get to Analysis?

eed3si9n (changes requested):

wpopielarski:
still in progress, no worries about that. I have to clean it first.

2017-10-24 6:34 GMT+02:00 eugene yokota <notifications@github.com>:

main/src/main/scala/sbt/Defaults.scala
kiritsuku: What is event logging?

eed3si9n: Calling `infoEvent` on ManagedLogger, which then gets relayed to CommandExchange.

main/src/main/scala/sbt/Defaults.scala
wpopielarski: fixed

main/src/main/scala/sbt/Defaults.scala
wpopielarski: fixed, previousCompile only is used

main/src/main/scala/sbt/Defaults.scala
wpopielarski: sorry, I don't get it.

wpopielarski:
could anybody tell me what's wrong with tests? Locally they are passing here not.

wpopielarski (commented):
ready to be checked again

laughedelic:
Hi @wpopielarski! I tried to test this branch, but couldn't make it work.

1. There is one thing that I had to change locally: to comment these two lines
  https://github.com/sbt/sbt/blob/9232a8fbc872c2f94a91ec82d797a11140a271a0/main/src/main/scala/sbt/Defaults.scala#L1412-L1413  
    This is called in the `compileIncremental` task, that I guess, is used in your `lspDefinition` in some way. The result is that servers bombards client with _empty_ `publishDiagnostics` notifications (to clean up), but this happens in response to **every** `textDocument/definition` request. On the sbt codebase it means that every request takes at least 20 seconds. And even on a small project it's ~5s.

    Commenting these lines out is not a solution, but the point is, that `textDocument/definition` shouldn't trigger diagnostics cleanup at all

    On a sidenote, I think that diagnostics cleanup should happen somewhere else. An alternative approach is to store diagnostics and every time new ones are generated, "merge" them and send only the necessary changes, this way there is no need to send lots of empty diagnostics on every compile. This is just an idea and, of course, out of scope of this PR.

2. After that I could see `textDocument/definition` requests/responses. And I couldn't get any results: responses are _always empty_. I tried on the sbt code, on some work project, on a tiny test project — it didn't work anywhere. Could you provide an example code on which it is supposed to work well?

By the way, I tested it with Atom, but I think that there is nothing VSCode-specific happening here, so it sould work the same in both editors. I should install VSCode and try it out too...

laughedelic:
No luck in VS Code either:
screen shot 2017-11-05 at 05 39 16

I have a feeling that I'm holding it from the wrong end.. 😅

wpopielarski:
Just Foo is not class, trait, object so not available in zinc analysis
file. Right now jump to definition is made from these analysis files not
scalameta-sbt plugin.

2017-11-05 5:44 GMT+01:00 Alexey Alekhin <notifications@github.com>:

wpopielarski:
Alexey,

there are not my lines and I think you should ask Eugen Yokota about the
reason of these 2 lines. Actually it is hard to say why you get empty
response. One of the reason can be the lack of zinc analysis cache files
which are used to find products of compilation. So probably you have to
compile project first and than try to jump to definition. And because type
identifiers are got from these files so if you are asking about identifier
which is not found there (even if it is scala Int, String or any other
common type) you get 'No definition found'.

2017-11-05 5:25 GMT+01:00 Alexey Alekhin <notifications@github.com>:

laughedelic:
> Just Foo is not class, trait, object so not available in zinc analysis file.

> One of the reason can be the lack of zinc analysis cache files
> which are used to find products of compilation. So probably you have to
> compile project first and than try to jump to definition. 

screen shot 2017-11-05 at 17 17 33

I definitely compiled the project and I don't get definitions on _any symbol_ in this little piece of code. Tried with defining a class, a trait, an object — nothing works..
As I said before, could you provide some example code which works for you?

> there are not my lines and I think you should ask Eugen Yokota about the reason of these 2 lines.

Sure, I wrote that more for @eed3si9n 

> Right now jump to definition is made from these analysis files not scalameta-sbt plugin.

I was actually thinking about it recently and was going to ask you. It seems that using information from the scalameta's semantic DB would be a straightforward way to get any symbol information for the language server. But I guess it's out of sbt's scope, because it requires a third-party plugin.. 
Anyway, in this light, is it still worth digging zinc analysis cache to get only partial information only for some symbols? Or would it make more sense to just implement this functionality outside of sbt, using scalameta-sbt (and interact with it through sbt server)?

wpopielarski:
Really hard to say. It's working for me. So my question is: do you get
compilation errors? This part should work as well. You should get also info
as logs about received requests by sbt server

2017-11-05 17:32 GMT+01:00 Alexey Alekhin <notifications@github.com>:

laughedelic:
> So my question is: do you get
> compilation errors? This part should work as well. You should get also info
> as logs about received requests by sbt server

It compiles well, no errors (really simple code on the screenshot above). And I do see logs about receiving requests. I also see empty responses in Atom (can I see client-server communication log somewhere in VS Code?).

wpopielarski:
about scalameta-sbt. Finding a symbol helps to find its definition by
determining its type or some local stable identifier or whatever else but
it still need work to figure out its source file and position in this file.
So analysis files are just first step in this way I guess

2017-11-05 18:00 GMT+01:00 Wieslaw Popielarski <wpopielarski@gmail.com>:

wpopielarski:
I don't mean 'correct code' :) I mean if you make failing snippet do you
get error compilation back?

2017-11-05 18:05 GMT+01:00 Alexey Alekhin <notifications@github.com>:

laughedelic:
> I don't mean 'correct code' :) I mean if you make failing snippet do you get error compilation back?

😄 I see. Yes, this part works fine, errors are reported as diagnostics.

> Finding a symbol helps to find its definition by determining its type or some local stable identifier or whatever else but it still need work to figure out its source file and position in this file. So analysis files are just first step in this way I guess

You mean that even with scalameta we will need analysis cache from zinc to map symbols to the source locations? I thought it's also stored in the SemanticDB, but now I see that it's not.

wpopielarski:
but what you can see in log from:

https://github.com/sbt/sbt/pull/3660/files#diff-82bb5e5687c35f3190dd111f8a9d93bbR166

2017-11-05 18:18 GMT+01:00 Alexey Alekhin <notifications@github.com>:

laughedelic:
Oh, I didn't see these logs. Should they appear in the sbt shell? If I don't get them, what am I doing wrong?

wpopielarski:
even this one?
https://github.com/sbt/sbt/pull/3660/files#diff-82bb5e5687c35f3190dd111f8a9d93bbR134

2017-11-05 18:39 GMT+01:00 Alexey Alekhin <notifications@github.com>:

wpopielarski:
so as I got you, you checked out vsc-definition branch and published it
locally and in your test project you set sbt version to this locally
published and it failed?

2017-11-05 18:43 GMT+01:00 Wieslaw Popielarski <wpopielarski@gmail.com>:

laughedelic:
Nope, I only see (in the console)

```
[debug] onRequestMessage: JsonRpcRequestMessage(2.0, 8, textDocument/definition, {"textDocument":{"uri":".../test.scala"},"position":{"line":13,"character":17}}})
```

And then empty `textDocument/publishDiagnostics` notifications. I also see that Atom receives an empty response:

```
Scala (sbt) rpc.sendRequest textDocument/definition sending Object ...
...
Scala (sbt) rpc.sendRequest textDocument/definition received (5673ms) []
```

But no logs from `lspDefinitionTask` anywhere.

laughedelic:
Yes, I checked your branch, published locally and I run test project with that locally published version. And I know that it's that version because I see the changes and I see logs about definition requests. And I also see `lspDefinition` task in the test project sbt shell. Btw, how can I test it directly from the shell?

wpopielarski:
can you ran task 'lspDefinition` in sbt console?

2017-11-05 18:51 GMT+01:00 Alexey Alekhin <notifications@github.com>:

laughedelic:
Yes, but I don't know what arguments to give it. I tried with the raw JSON request, but it didn't work:

```
 lspDefinition """{ "jsonrpc": "2.0", "method": "textDocument/definition", "params": {"textDocument":{"uri":"file:///.../test.scala"},"position":{"line":15,"character":13}}} }"""
[warn] Incorrect definition request Some({ "jsonrpc": "2.0", "method": "textDocument/definition", "params": {"textDocument":{"uri":"file:///.../test.scala"},"position":{"line":15,"character":13}}} })
[success] Total time: 0 s, completed Nov 5, 2017 6:54:38 PM
```

wpopielarski:
ok, first 'lspDefinition {"textDocument":{"uri":"file:///.../test.
scala"},"position":{"line":15,"character":13}' but it looks like this task
is not visible, i don't know... I take a look tomorrow at this

2017-11-05 19:08 GMT+01:00 Alexey Alekhin <notifications@github.com>:

laughedelic:
OK, I figured out that it was about log level, I changed it to `warn` in the code and now I see the logs:

```
> lspDefinition {"textDocument":{"uri":"file:///.../test.scala"},"position":{"line":15,"character":13}}
[warn] lsp-definition raw request: List({"textDocument":{"uri":"file:///.../test.scala"},"position":{"line":15,"character":13}})
[warn] lsp-definition found line:   def qux(f: Foo): String = f.b
[debug] langNotify: JsonRpcNotificationMessage(2.0, textDocument/publishDiagnostics, {"uri":"file:/.../test.scala","diagnostics":[]})
[warn] lsp-definition potentials: Set()
[warn] lsp-definition potentials: Set()
[success] Total time: 1 s, completed Nov 5, 2017 7:27:56 PM
```

wpopielarski:
but `Foo` is `type Foo` or class Foo? Sorry I scolled up and see it, it's a class. so what you get when call `show previousCompile` ? Ok eureka my bug my fault. You defined `Foo` in default package and unfortunately I don't look for classes,traits, objects there. Need to fix it. But when you started to play with it use packages please, and let me know. you earned medal for testers 👍

laughedelic:
Class `Foo` just as you see on the screenshot from https://github.com/sbt/sbt/pull/3660#issuecomment-341985989, no other `Foo` in scope:

```scala
class Foo(val a: Int, val b: String)

object Buh {
  val foo: Foo = new Foo(1, "a")

  def qux(f: Foo): String = f.b

  qux(foo)
}
```

This is the only code in the one file in the project.

wpopielarski:
see at:
https://github.com/sbt/sbt/pull/3660/files#diff-82bb5e5687c35f3190dd111f8a9d93bbR61
I need some integration test too I see

laughedelic:
Didn't notice your edit in https://github.com/sbt/sbt/pull/3660#issuecomment-341994983 at first.. 😄 

I added a package, tried putting `Foo` inside the `Buh` object, nothing helps.. Here's `previousCompile` output:

```
> show previousCompile
[info] PreviousResult(analysis: Optional[Analysis: 3 Scala sources, 5 classes, 2 binary dependencies], setup: Optional[MiniSetup(output: SingleOutput(/Users/laughedelic/dev/laughedelic/scala-adder/target/scala-2.12/classes), options: MiniOptions(classpathHash: [Lxsbti.compile.FileHash;@68d57037, scalacOptions: [Ljava.lang.String;@5f0669d9, javacOptions: [Ljava.lang.String;@1c4f3b3d), compilerVersion: 2.12.4, order: Mixed, storeApis: true, extra: [Lxsbti.T2;@211d17c)])
[success] Total time: 0 s, completed Nov 5, 2017 8:04:14 PM
```

wpopielarski:
generally I tested it on the whole `sbt` project itself and working quite cool (I even did demo for @eed3si9n  and @sschaef ). Will see tomorrow then... Sorry for that.

wpopielarski:
good news is that is is working for simple projects too. But I have some
complex configuration taken from sbt project. I try to narrow it.

2017-11-05 20:06 GMT+01:00 Alexey Alekhin <notifications@github.com>:

wpopielarski:
well, looks that extremely simple build.sbt must work. The only important
observation is that the very root project folder must be open by vs
code.Like this:

screenshot from 2017-11-06 12 56 20

wpopielarski:
and one remark: the selected definition is not found in the very first time for clean project (no compilation done). The second attempt is successful.
REMARK: I didn't test it without removed by you lines in `Defaults`

wpopielarski:
ok, I see big problems when these two lines are removed. What is interesting, I cannot drop back to previous behavior.

laughedelic:
What do you mean? I already uncommented them and I get those empty diagnostics again. But I still don't get definitions. Can you share your acme test project?

wpopielarski:
funny thing, I can see definition of `Foo` in `internal.Foo.scala` only
screenshot from 2017-11-06 15 15 48

wpopielarski:
so I don't have any idea how `previousCompile` is working
screenshot from 2017-11-06 15 19 49

laughedelic:
🙁 it's confusing..

On a sidenote, guys from scalameta are actively working on the go-to-definition implementation right now: https://github.com/scalameta/language-server/pull/8

wpopielarski:
to calm down in 'previousCompile` usage. It's ok. I suffer with toolbox problem. testing...

wpopielarski:
@laughedelic could you please, get the commit up and check if definition is now working? There was my fault. Unfortunately `ToolBox.parse` is more powerful I expected. So now to check if given token is longest valid scala identifier I ask tb to compile this snippet of code.

laughedelic:
Done 👌 It worked for me for the first time! But only with objects. Doesn't work with anything else..

screen shot 2017-11-06 at 19 06 56

(UPD: better screenshot, so you can see both files content and the project structure)

I also noticed that response contains same reference object twice.

wpopielarski:
really, not working for plain classes? or classes in objects?

2017-11-06 19:05 GMT+01:00 Alexey Alekhin <notifications@github.com>:

laughedelic:
Nope. On the screenshot above I have class `Foo`, tried putting it in or out of the object, doesn't work either way.

wpopielarski:
I'm just asking because right now is very simple (and naive) assumption
about simple class names. Only suffices .A, .A$, $A$ and $A are available. But this naive assumption is good enough besides default package

2017-11-06 19:24 GMT+01:00 Wieslaw Popielarski <wpopielarski@gmail.com>:

wpopielarski:
I don't know about classes.... Here you are:
screenshot from 2017-11-07 09 29 53

wpopielarski:
@laughedelic could you play a while?
screenshot from 2017-11-07 16 56 11

laughedelic:
@wpopielarski I'll try it out tonight 👍

laughedelic:
@wpopielarski Now it **does** work for objects, classes and traits 👍

main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
```diff
@@ -60,6 +58,12 @@ private[sbt] trait LanguageServerProtocol extends CommandChannel {
         langRespond(InitializeResult(serverCapabilities), Option(request.id))
       case "textDocument/didSave" =>
         append(Exec("compile", Some(request.id), Some(CommandSource(name))))
+      case "textDocument/definition" =>
+        import sjsonnew.support.scalajson.unsafe.CompactPrinter
+        append(
+          Exec(s"lspDefinition ${CompactPrinter(json)}",
```
eed3si9n: The query for definition, if possible, should happen concurrently with whatever that's happening in the command processing engine. For example, the user could be in the middle of `test` that takes 10 min. I think `textDocument/definition` should return while the test is running, not afterwards. So I don't think we should use task or input task here.

eed3si9n (changes requested):

wpopielarski:
so should be command used here? This part of code is using
`previousCompile` to collect analyses. Is any other way to figure out
analysis files names and locations?

2017-11-08 5:05 GMT+01:00 eugene yokota <notifications@github.com>:

eed3si9n:
I am guessing that you'd need to set up a cache of Analysis files yourself based on the known locations.

wpopielarski:
that's clear. thanks

2017-11-08 6:45 GMT+01:00 eugene yokota <notifications@github.com>:

main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
wpopielarski: looks like fixed I hope

wpopielarski:
@laughedelic @sschaef  @eed3si9n could you play with is a little please?

laughedelic:
Hi @wpopielarski! I just tried out 9152ece. It's much more stable than before! It works on all objects, classes and traits in my test project 👍 

screen shot 2017-11-14 at 04 57 40

Although I can't say that it's usable yet. There are several problems:
* When I hover on empty space it responses with a lot of `{code: -32700, message: "Symbol not found in definition request"}` errors. I think it shouldn't send an error and `ParseError` is probably meant for the cases when you cannot parse the request itself. I also checked what other language servers do: 
  - Flow server sends a response with `null` result (indicating success) and 
  - and Python sends an empty response `[]` (no symbols found).
  IMO Flow server does the right thing
* It's unclear what to expect from this functionality, which things are supposed should work and which not. I think you should clarify it. You say 
  > Supports types only, and only in case when type is found in Zinc analysis

  but I have no idea what's found in Zinc analysis. Also, does it mean that it will never work for values/methods/etc.?
* It's super slow 🐌  Some unsuccessful(!) requests take ~1.5s, and most successul ones take >1s. I see a spinning wheel on my mouse pointer 😅 I understand that this is work in progress and something could be optimized, but I think it's supposed to be several orders of magnitude faster.. 

----

Overall, I think that this is a very good experiment, but I really doubt now that it should be a part of the sbt server core. Taking into account that it's implemented through a task, it could probably be an sbt plugin? Or is it really tied to the sbt internals?

In my opinion sbt server development should be focused more on the interaction with the build, not sources analysis. There are already two (three if you count Dotty) language servers for Scala that provide this functionality and many other things: ENSIME and Scalameta (see https://github.com/scalameta/language-server/pull/23). 

@wpopielarski I don't want to discourage you from improving this, I just want everybody working on things related to Scala and LSP be more aware of each other's work.

wpopielarski:
@laughedelic class trait and object are for good start. zinc can give info about methods and class fields too. I think we will use scalameta finally. meantime I add super efficient optimization (if working :) )

wpopielarski:
well, even with simple classes is problem because names are encoded so there is no other possibility than use scalameta :). But is NameTransformer

main/src/main/scala/sbt/lsp/Definition.scala
```diff
@@ -0,0 +1,330 @@
...
+import sbt.internal.langserver.ErrorCodes
+import sbt.util.Logger
+import scala.annotation.tailrec
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.concurrent.duration.Duration.Inf
+import scala.util.matching.Regex.MatchIterator
+import java.nio.file.{ Files, Paths }
+import sbt.StandardMain
+
+object Definition {
```
dwijnand: Either this object should be `private[sbt]` or the whole `lsp` package should be at `sbt.internal.lsp`.

dwijnand (changes requested):
I left one comment about making sure that private/internal things are private.

Additional things for this PR:
1. squash the commits
2. remove the weird, side diffs from this PR (e.g `import Configurations.Test`, etc)
3. check and report what the transitive dependencies of scalacache-caffeine are.

wpopielarski:
@dwijnand 
looks like there is no dependency in compile. For simple maven project dependency:tree returns:
```
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ my-app ---
[INFO] com.mycompany.app:my-app:jar:1.0-SNAPSHOT
[INFO] \- com.github.ben-manes.caffeine:caffeine:jar:2.6.0:compile
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
```

dwijnand:
the transitive dependencies of scalacache-caffeine look fine too:

```
com.github.cb372:scalacache-caffeine_2.12:0.20.0 [S]
+-com.github.ben-manes.caffeine:caffeine:2.5.6
+-com.github.cb372:scalacache-core_2.12:0.20.0 [S]
| +-org.scala-lang:scala-reflect:2.12.3 [S]
| +-org.slf4j:slf4j-api:1.7.25
|
+-org.slf4j:slf4j-api:1.7.25
```

eed3si9n:
I'm closing this in favor of https://github.com/sbt/sbt/pull/3777

eed3si9n:
If you could give me push rights, I could put those commits here too.

wpopielarski:
@eed3si9n thanks a lot for flattening history. I did the mess. And I think if https://github.com/sbt/sbt/pull/3777 is merged so is there a need to commit it to this branch? If I got you correctly.

eed3si9n:
@wpopielarski I guess not. Thanks for the contribution :)

PR #3663, opened 2017-10-24
fix sbt help new does not explain how to use sbt new

### LINKED ISSUE #2939: sbt help new does not explain how to use sbt new
## steps

`sbt help new`

## problem

Instead of explaining how to use `sbt new`, calling `sbt help new` just renders the normal output of `sbt help`:

```
> sbt help new
[info] Loading project definition from /Users/bene/workspace/scala-user-group-dus/scala-kickstart/project
[info] Set current project to Hello (in build file:/Users/bene/workspace/scala-user-group-dus/scala-kickstart/)

  help                                    Displays this help message or prints detailed help on requested commands (run 'help <command>').
  completions                             Displays a list of completions for the given argument string (run 'completions <string>').
  about                                   Displays basic information about sbt and the build.
  tasks                                   Lists the tasks defined for the current project.
  settings                                Lists the settings defined for the current project.
  reload                                  (Re)loads the current project or changes to plugins project or returns from it.
  new                                     Creates a new sbt build.
  projects                                Lists the names of available projects or temporarily adds/removes extra builds to the session.
  project                                 Displays the current project or changes to the provided `project`.
  set [every] <setting>                   Evaluates a Setting and applies it to the current project.
  session                                 Manipulates session settings.  For details, run 'help session'.
  inspect [uses|tree|definitions] <key>   Prints the value for 'key', the defining scope, delegates, related definitions, and dependencies.
  <log-level>                             Sets the logging level to 'log-level'.  Valid levels: debug, info, warn, error
  plugins                                 Lists currently available plugins.
  ; <command> (; <command>)*              Runs the provided semicolon-separated commands.
  ~ <command>                             Executes the specified command whenever source files change.
  last                                    Displays output from a previous command or the output from a specific task.
  last-grep                               Shows lines from the last output for 'key' that match 'pattern'.
  export <tasks>+                         Executes tasks and displays the equivalent command lines.
  exit                                    Terminates the build.
  early(<command>)                        Schedules a command to run before other commands on startup.
  show <key>                              Displays the result of evaluating the setting or task associated with 'key'.
  all <task>+                             Executes all of the specified tasks concurrently.

More command help available using 'help <command>' for:
  !, +, ++, <, alias, append, apply, eval, iflast, onFailure, reboot, shell

Template not found for:
```
## expectation

`sbt help new` should render useful information, e.g.:

- how to use the `new` command
- possible options for `new`
- how to discover new templates

## notes

Nevertheless `sbt new` is awesome!

PanAeon:
Should fix #2939 
Made `help` command "consume" commands that go after it so when user types `sbt help new` it behaves similar to `sbt "help new"`.
Added basic example, and link to [giter8](http://www.foundweekends.org/giter8/) to the `sbt new` detailed help message.


main-command/src/main/scala/sbt/BasicCommands.scala
```diff
@@ -87,10 +87,20 @@ object BasicCommands {
   }
 
   def runHelp(s: State, h: Help)(arg: Option[String]): State = {
-    val message = try Help.message(h, arg)
+
+    val (extraArgs, remainingCommands) = s.remainingCommands match {
```
eed3si9n, line 120: Why is it necessary to pop a remaining command from the queue?

eed3si9n (changes requested):

main-command/src/main/scala/sbt/BasicCommands.scala
eed3si9n: Never mind you explained it above:

> Made help command "consume" commands that go after it so when user types `sbt help new` it behaves similar to `sbt "help new"`.

eed3si9n: Would there be a situation where the user might need more than one argument?

dwijnand: What's the behaviour if you run `sbt help new compile`?

PanAeon: It will print `No matches for regular expression 'new compile'.` Since `help` is kind of interactive command, I'm assuming that user wouldn't need to supply other commands after it.

eed3si9n (approved):
LGTM

eed3si9n:
CI failures are legitimate. I am resending this as https://github.com/sbt/sbt/pull/4237 and try to land this.

Issue #3583, opened 2017-09-26
sbt 1.0.2 vs 0.13.16: different test TaskDefs


cunei:
### steps
```
$ cat build.sbt 
version := "0.0.1"
name := "broken"
organization := "org.catastrophe"
scalaVersion := "2.10.6"
libraryDependencies += "org.scalacheck" %% "scalacheck" % "1.13.5" % "test"
$ cat ./src/test/scala/q/X.scala
package q
object X {
  object Y extends org.scalacheck.Properties("a")
}
$ cat project/build.properties 
sbt.version=1.0.2
#sbt.version=0.13.16
```
## problem
`sbt test` succeeds with 0.13.16, but with 1.0.2 it says:
```
[error] java.lang.ClassNotFoundException: q.X.Y$
[error]         at java.lang.ClassLoader.findClass(ClassLoader.java:530)
[error]         at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
[error]         at sbt.internal.inc.classpath.ClasspathFilter.loadClass(ClassLoaders.scala:70)
[error]         at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
[error]         at java.lang.Class.forName0(Native Method)
[error]         at java.lang.Class.forName(Class.java:348)
[error]         at org.scalacheck.Platform$.loadModule(Platform.scala:67)
[error]         at org.scalacheck.ScalaCheckRunner$BaseTask.<init>(ScalaCheckFramework.scala:49)
[error]         at org.scalacheck.ScalaCheckRunner$$anon$1.<init>(ScalaCheckFramework.scala:81)
[error]         at org.scalacheck.ScalaCheckRunner.rootTask(ScalaCheckFramework.scala:81)
```
### notes
I dug into the Scalacheck code. The code receives a bunch of `sbt.testing.TaskDef` instances (as per sbt's `test-interface`), and tries a `Class.forName(name + "$",...` on each. The list does not contain `q.X.Y`, above in 0.13.16, but it does in 1.0.2, and the `forName` fails.

0.13.16:
```
$ sbt 'show testGrouping'
[info] * Group(<default>,List(),InProcess)
```
1.0.2:
```
$ sbt 'show testGrouping'
[info] * Group(<default>,Vector(Test q.X.Y : subclass(true, org.scalacheck.Properties)),InProcess)
```
Any comments on the discrepancy? I would assume that the same scalaVersion and libraries should lead to the same testing behavior (except for intentional fixes).

[...]

PR #3669, opened 2017-10-25
Interim fix for #3583, for 1.0.3

cunei:
In 0.13.x, zinc would discover only top-level objects and classes
containing tests to the test framework. In 1.x, however, zinc can
discover also nested objects and classes; that causes the "name"
of a ClassLike to no longer be usable for reflection.

This change filters out nested objects/classes from the list,
restoring compatibility with 0.13. A zinc extension of ClassLike
will probably be introduced in 1.1 or 1.2, in order to provide
the test framework with enough information to deal with nested
classes.

This patch unblocks https://github.com/sbt/sbt-standalone-build/issues/15


eed3si9n:
It might be good to add scripted test for repro.

eed3si9n (changes requested):
Release notes, and repro tests plz.

cunei:
retest this please

eed3si9n (approved):

eed3si9n:
Closed and re-opened to wake up Travis - https://travis-ci.org/sbt/sbt/builds/293329548

PR #3671, opened 2017-10-26
Add nameOption to the Command trait

### LINKED ISSUE #3297: Expose useful information about available commands

miklos-martin:
Fixes #3297 and #3531


main-command/src/main/scala/sbt/Command.scala
```diff
@@ -187,6 +187,13 @@ object Command {
 
   def spacedC(name: String, c: Parser[Char]): Parser[String] =
     ((c & opOrIDSpaced(name)) ~ c.+) map { case (f, rem) => (f +: rem).mkString }
+
+  implicit class CommandWithName(cmd: Command) {
```
dwijnand: Could you make it a value class, so we don't have to pay any allocation, please? `extends AnyVal`

main-command/src/main/scala/sbt/Command.scala
```diff
@@ -187,6 +187,13 @@ object Command {
...
+    def commandName: Option[String] = cmd match {
```
dwijnand: I wonder if this is a good name. Should we just call it `name`, or is it that a bad idea? Maybe `getName`? 🤔

dwijnand (changes requested):

main-command/src/main/scala/sbt/Command.scala
eed3si9n: Not a fan of value class. It doesn't help perf and it has bincompat cost.

dwijnand: Value classes are excellent for extension methods, like this one.

main-command/src/main/scala/sbt/Command.scala
jvican: We cannot call it name because `SimpleCommand` already defines a `name` argument of different type: `String`.

main-command/src/main/scala/sbt/Command.scala
jvican: I think value classes are good in this scenario, they're just syntactic sugar for classes + implicit conversions, and we have lots of those in sbt/sbt.

main-command/src/main/scala/sbt/Command.scala
dwijnand: I know. So on `SimpleCommand` you get `String`, on Command you'll get an `Option[String]` because it might or might not have a name.

But again, it might be a bad idea. How about `nameOption`?

miklos-martin: `nameOption` sounds better imho

main-command/src/main/scala/sbt/Command.scala
miklos-martin: What should it be then?

dwijnand (approved):
LGTM, but I'll hold fire to see about the value class thing..

jvican:
IMHO `optionalName` is better than `nameOption`.

dwijnand:
Given `nameOption` matches `reduceOption` and `headOption` in the standard library, I think it might feel more familiar.

jvican:
Sounds reasonable, then.

main-command/src/main/scala/sbt/Command.scala
adriaanm: Agreed on using VC for extension methods (also agreed it's not great for other applications). If MiMa says it's ok, then binary compat should be fine.

eed3si9n: By "bincompat cost", I mean there will be limitation on the future extensibility of the class while keeping the binary compatibility.

Take `RichFile(val asFile: File)` for example -https://github.com/sbt/io/blob/v1.0.2/io/src/main/scala/sbt/io/Path.scala

In sbt 1.1, we support POSIX permission operations via NIO, so it would be nicer if we could expand the parameter of the class to `RichFile(val asFile: File, val asPath: Path)` so that's cached. That kind of extension would not work because value class codegens all methods under companion. As a workaround I implemented POSIX operations in https://github.com/sbt/io/pull/76 as universal trait.

dwijnand: If we want to have NIO operations around a (cached) Path we should probably create a `RichPath`. We can dedup methods with universal traits, and the JDK already has the asPath/asFile conversions.

main-command/src/main/scala/sbt/Command.scala
```diff
@@ -187,6 +187,13 @@ object Command {
 
   def spacedC(name: String, c: Parser[Char]): Parser[String] =
     ((c & opOrIDSpaced(name)) ~ c.+) map { case (f, rem) => (f +: rem).mkString }
+
+  implicit class CommandWithName(val cmd: Command) extends AnyVal {
```
dwijnand: Coming back to this.. why the hell did I suggest an extension method (in https://github.com/sbt/sbt/pull/3531#discussion_r139164742) when the target type (`Command`) is right here? Oops.

Sorry @miklos-martin. Could you move this to `Command` please?

dwijnand (changes requested):

main-command/src/main/scala/sbt/Command.scala
miklos-martin: Sure.
Should I move it in the `Command` trait exactly as it is here, or with a `None` implementation in the trait and override it in `SimpleCommand` with `Some(name)`?

miklos-martin: ping @dwijnand

dwijnand: Sorry, as it is is better (imo).

miklos-martin: Btw I'd fancy the latter, if that matters

miklos-martin: OK

dwijnand (approved):
Thanks @miklos-martin.

miklos-martin:
my pleasure

miklos-martin:
I've rebased this

PR #3672, opened 2017-10-26
Add new setting to determine if running in CI env

### LINKED ISSUE #3653: Add key to know whether we're inside CI or not
## problem
Builds often need to know where they are run. In my experience, it is very common that you want to change the definition of a key depending on where it's run; whether it's in the CI or in my computer.
  
Currently, there's no mechanism to do so.

## expectation
I propose to add a boolean setting key called `insideCI`, along with a default implementation. This will allow users implement it depending on the concrete CI they use.

## notes
sbt already has checks to know whether it runs in the CI or not. See the following definition in `Defaults.scala`:
```scala
        ivyLoggingLevel :== {
          // This will suppress "Resolving..." logs on Jenkins and Travis.
          if (sys.env.get("BUILD_NUMBER").isDefined || sys.env.get("CI").isDefined)
            UpdateLogging.Quiet
          else UpdateLogging.Default
        },
```
  
It uses the standard `CI` environment option or `BUILD_NUMBER`. I propose `sys.env.get("CI").isDefined` to be the default implementation of `insideCI`, and make the definition of `ivyLoggingLevel` a call site of this key.

sbt version: 1.x

RomanIakovlev:
Fixes #3653

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -273,6 +273,7 @@ object Defaults extends BuildCommon {
         .toHex(Hash(appConfiguration.value.baseDirectory.toString))
         .## % 1000),
       serverAuthentication := Set(ServerAuthentication.Token),
+      insideCI :== sys.env.contains("BUILD_NUMBER") || sys.env.contains("CI"),
     ))
 
   def defaultTestTasks(key: Scoped): Seq[Setting[_]] =
@@ -1749,12 +1750,10 @@ object Classpaths {
         dependencyOverrides :== Vector.empty,
         libraryDependencies :== Nil,
         excludeDependencies :== Nil,
-        ivyLoggingLevel :== {
-          // This will suppress "Resolving..." logs on Jenkins and Travis.
-          if (sys.env.get("BUILD_NUMBER").isDefined || sys.env.get("CI").isDefined)
-            UpdateLogging.Quiet
-          else UpdateLogging.Default
-        },
+        ivyLoggingLevel := (// This will suppress "Resolving..." logs on Jenkins and Travis.
+        if (insideCI.value)
+          UpdateLogging.Quiet
+        else UpdateLogging.Default),
         ivyXML :== NodeSeq.Empty,
         ivyValidate :== false,
         moduleConfigurations :== Nil,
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -513,7 +513,8 @@ object Defaults extends BuildCommon {
     Seq(
       initialCommands :== "",
       cleanupCommands :== "",
-      asciiGraphWidth :== 40
+      asciiGraphWidth :== 40,
+      insideCI :== sys.env.get("BUILD_NUMBER").isDefined || sys.env.get("CI").isDefined
```
Duhemm: `sys.env.contains` maybe?

Duhemm (commented):
The changes LGTM, but I'm wondering if there's a risk of people having troubles with timeouts in Travis (and potentially other CIs) because no output may be produced for a relatively long time. Can people remove the `BUILD_NUMBER` or `CI` environment variable when running outside of self-hosted CIs? I would suggest to:
 - If `BUILD_NUMBER` or `CI` is defined, log a line saying that logging can be explicitly enabled by setting an environment variable
 - Check if that environment variable is set, and if so, still log the resolving information.

@jvican What do you think?

jvican (approved):
I think this looks awesome! Using `sys.env.contains` is indeed better :smile:. Thanks for the hard work @RomanIakovlev

Duhemm (approved):
@jvican Duh! You didn't change that. LGTM, thanks and congrats! 🎉

main/src/main/scala/sbt/Defaults.scala
eed3si9n: So should we wait for `contains` change, append to this PR ourself, or merge as is?

dwijnand (approved):

main/src/main/scala/sbt/Defaults.scala
dwijnand: Could you define this default in `globalSbtCore` rather than `configGlobal` please?

dwijnand (changes requested):

RomanIakovlev:
Moved to `globalSbtCore`, replaced `.get(...).isDefined` with `contains`.

dwijnand (approved):
Thank you!

RomanIakovlev:
@Duhemm regarding your concern about timeouts because of no output produced, if I understood it correctly. I just wanted to point this PR doesn't change the actual behavior of sbt, because the code to check the environment variables in `ivyLoggingLevel` already used to do the very same check.

PR #3673, opened 2017-10-26
Add a check for a change in SBT version before reload

### LINKED ISSUE #1055: Prevent loading a project with the wrong version of sbt
This situation arises if someone tries to use `reload` after having changed `build.properties`. As it's probably harder/impossible to fix taking `build.properties` into account when running `reload` sbt could at least refuse to load when the sbt.version in `build.properties` doesn't match the current runtime version.

RomanIakovlev:
Fixes #1055

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -56,7 +56,8 @@ import StandardMain._
 
 import java.io.{ File, IOException }
 import java.net.URI
-import java.util.Locale
+import java.util.{ Locale, Properties }
+
 import scala.util.control.NonFatal
 import BasicCommandStrings.{ Shell, TemplateCommand }
 import CommandStrings.BootCommand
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/Main.scala
```diff
@@ -672,7 +673,26 @@ object BuiltinCommands {
...
+    val sbtVersionOpt = if (buildProps.exists) {
+      val buildProperties = new Properties()
+      IO.load(buildProperties, buildProps)
+      Option(buildProperties.getProperty("sbt.version"))
+    } else None
+
+    sbtVersionOpt.foreach(version =>
+      if (version != app.id.version()) {
+        state.log.warn(
+          "SBT version in build.properties file has changed, use 'reboot' instead of 'reload' to use the new value")
```
eed3si9n: sbt should be lowercased.

eed3si9n (changes requested):

main/src/main/scala/sbt/Main.scala
RomanIakovlev: Done.

sbt/src/sbt-test/project/sbt-version-change/build.sbt
```diff
@@ -0,0 +1,6 @@
+TaskKey[Unit]("checkSbtVersionWarning") := {
+	val state = Keys.state.value.globalLogging
+	val contents = IO.read(state.backing.file)
+	assert(contents.contains("SBT version in build.properties file has changed, use 'reboot' instead of 'reload' to use the new value"))
```
eed3si9n: Change here too?

RomanIakovlev: Oops, forgot about that one, sorry! Fixed.

eed3si9n (approved):

main/src/main/scala/sbt/Main.scala
```diff
@@ -672,7 +673,26 @@ object BuiltinCommands {
...
+    val sbtVersionOpt = if (buildProps.exists) {
+      val buildProperties = new Properties()
+      IO.load(buildProperties, buildProps)
+      Option(buildProperties.getProperty("sbt.version"))
+    } else None
+
+    sbtVersionOpt.foreach(version =>
+      if (version != app.id.version()) {
+        state.log.warn(
+          "sbt version in build.properties file has changed, use 'reboot' instead of 'reload' to use the new value")
```
dwijnand: I think it would be better to include detail on how it changed. So maybe:

```scala
s"""sbt version mismatch, current: ${app.id.version()}, in build.properties: "$version", use 'reboot' to use the new value."""
```

dwijnand (commented):
LGTM, but I have a thought.

main/src/main/scala/sbt/Main.scala
RomanIakovlev: I agree it looks nicer. I'm concerned this would make scripted test a bit tricky, because current version will be constantly changing as sbt evolves. One "easy" way is, obviously, to hardcode the current sbt version number in the scripted test and update it every time the sbt version is bumped, but it has obvious disadvantages. Is there a better way?

dwijnand: You can still access `state.value.configuration.provider.id.version()` in the scripted test.

If that fails just search-replace that part of the message to ignore it.

RomanIakovlev: Right, that works. Pushed the change.

dwijnand (approved):
Thanks @RomanIakovlev!

PR #3678, opened 2017-10-27
Fix compile clean

### LINKED ISSUE #3542: Make `compile:clean` clean only compilation artifacts
## problem
You are an sbt user. You know of `clean`. You also know that tasks can be scoped in tasks.
  
You find yourself cleaning often, but most of the times you only want to clean compilation artifacts, and you want the rest of a project to be untouched.
  
You decide to be creative, and try `compile::clean`, expecting sbt to only clean the analysis files and the class files/jars created by the previous compilation.
  
```
sbt:sbtRoot> actionsProj/compile
[success] Total time: 2 s, completed Sep 16, 2017 12:22:20 AM
sbt:sbtRoot> actionsProj/compile:clean
[success] Total time: 0 s, completed Sep 16, 2017 12:22:33 AM
sbt:sbtRoot> actionsProj/compile
[info] Updating {file:/data/rw/code/scala/sbt/}actionsProj...
[info] Done updating.
[info] Formatting 10 Scala sources in actionsProj:compile ...
[info] Compiling 10 Scala sources to /data/rw/code/scala/sbt/main-actions/target/scala-2.12/classes ...
[warn] /data/rw/code/scala/sbt/main-actions/src/main/scala/sbt/Doc.scala:41:15: parameter value doc in method javadoc is never used
[warn]               doc: JavaTools,
[warn]               ^
[warn] /data/rw/code/scala/sbt/main-actions/src/main/scala/sbt/Doc.scala:42:15: parameter value log in method javadoc is never used
[warn]               log: Logger,
[warn]               ^
[warn] /data/rw/code/scala/sbt/main-actions/src/main/scala/sbt/Doc.scala:43:15: parameter value reporter in method javadoc is never used
[warn]               reporter: Reporter,
[warn]               ^
[warn] /data/rw/code/scala/sbt/main-actions/src/main/scala/sbt/Sync.scala:70:13: pattern var target in value $anonfun is never used; `target@_' suppresses this warning
[warn]       case (target, srcs) =>
[warn]             ^
[warn] /data/rw/code/scala/sbt/main-actions/src/main/scala/sbt/Sync.scala:31:13: parameter value outStyle in method apply is never used
[warn]             outStyle: FileInfo.Style = FileInfo.exists)
[warn]             ^
[warn] /data/rw/code/scala/sbt/main-actions/src/main/scala/sbt/TestResultLogger.scala:132:18: pattern var name in value $anonfun is never used; `name@_' suppresses this warning
[warn]                 (name, testEvent)) =>
[warn]                  ^
[warn] /data/rw/code/scala/sbt/main-actions/src/main/scala/sbt/Tests.scala:156:14: pattern var a in value $anonfun is never used; `a@_' suppresses this warning
[warn]         case a: Argument              => // now handled by whatever constructs `runners`
[warn]              ^
[warn] /data/rw/code/scala/sbt/main-actions/src/main/scala/sbt/Tests.scala:295:18: parameter value tags in method makeSerial is never used
[warn]                  tags: Seq[(Tag, Int)]): Task[List[(String, SuiteResult)]] = {
[warn]                  ^
[warn] /data/rw/code/scala/sbt/main-actions/src/main/scala/sbt/ForkTests.scala:77:52: method ansiCodesSupported in class Logger is deprecated (since 1.0.0): No longer used.
[warn]             val config = new ForkConfiguration(log.ansiCodesSupported, parallel)
[warn]                                                    ^
[warn] /data/rw/code/scala/sbt/main-actions/src/main/scala/sbt/Package.scala:102:26: value IMPLEMENTATION_VENDOR_ID in object Name is deprecated: see corresponding Javadoc for more information.
[warn]                          IMPLEMENTATION_VENDOR_ID)
[warn]                          ^
[warn] /data/rw/code/scala/sbt/main-actions/src/main/scala/sbt/Package.scala:105:27: value IMPLEMENTATION_URL in object Name is deprecated: see corresponding Javadoc for more information.
[warn]       homepage map (h => (IMPLEMENTATION_URL, h.toString))
[warn]                           ^
[warn] 11 warnings found
[info] Done compiling.
[success] Total time: 21 s, completed Sep 16, 2017 12:22:57 AM
```
  
But it doesn't work: sbt resolves the module again. You expected `compile::clean` to leave the cached files in `target` by resolution untouched.

## expectation
`compile:clean` should clean only the compilation artifacts.

Benefits:
  
1. sbt is more consistent, therefore more intuitive.
2. sbt optimizes for a common use case and saves some times to the end users (resolution is expensive).

laughedelic:
Fixes #3542. I started it at the Lambda World Spree with great help from @jvican.

This is a simple ah-hoc solution, it could be improved in many ways. The main change here is that
* `compile:compile::clean` cleans up only things in `compile:classDirectory`
* `test:compile::clean` cleans up only things in `test:classDirectory`

Also I'm not sure how it should handle generated sources/resources.


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -363,7 +363,14 @@ object Defaults extends BuildCommon {
   lazy val outputConfigPaths = Seq(
     classDirectory := crossTarget.value / (prefix(configuration.value.name) + "classes"),
     target in doc := crossTarget.value / (prefix(configuration.value.name) + "api")
+  ) ++ inTask(compile)(
```
jvican, line 366: To reviewers: this is scoped in the default configs `Compile`, `Test` and `Runtime`.

sbt/src/sbt-test/actions/compile-clean/test
```diff
@@ -0,0 +1,19 @@
...
+$ exists target/scala-2.12/test-classes/B.class
+
+> compile:compile::clean
+$ exists target/scala-2.12/resolution-cache/
+$ exists target/streams/
+$ exists target/scala-2.12/test-classes/B.class
+# it should clean only class files
+$ absent target/scala-2.12/classes/A.class
+
+> test:compile::clean
```
jvican: 👍 great test

jvican (approved):
LGTM! Great work @laughedelic.

main/src/main/scala/sbt/Defaults.scala
dwijnand, line 366: No if you look at #3542 it talks about creating a `compile:clean`, not `compile::clean`, i.e a `clean` scoped to the `Compile` configuration, not to the `compile` task.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -517,9 +524,9 @@ object Defaults extends BuildCommon {
     ))
 
   lazy val projectTasks: Seq[Setting[_]] = Seq(
-    cleanFiles := cleanFilesTask.value,
+    cleanFiles := Def.taskDyn { cleanFilesTask(managedDirectory.value, target.value) }.value,
```
laughedelic, line 528: This is the same as it was before. But I'm not sure if I should add `managedDirectory` to `cleanFiles` in the `compile` task above.

sbt/src/sbt-test/actions/compile-clean/test
```diff
@@ -0,0 +1,19 @@
...
+$ exists target/scala-2.12/resolution-cache/
+$ exists target/streams/
+# it should clean both compile and test classes
+$ absent target/scala-2.12/classes/A.class
```
dwijnand: I have 2 concerns with this test: (1) I think it's over-specifying - it shouldn't make assertions about whether or not there is a `streams` directory, etc, and (2) I also think it's wrong - cleaning test classes shouldn't clean compile classes too.

dwijnand (changes requested):

main/src/main/scala/sbt/Defaults.scala
jvican: I changed that in the ticket and advised Alexey to do it this way. I think that scoping in the task makes more sense because you may be interested in deleting test class files as well.

dwijnand: > I think that scoping in the task makes more sense because you may be interested in deleting test class files as well.

I agree, and `test:clean` should do that.

sbt/src/sbt-test/actions/compile-clean/test
jvican: I agree. Ideally, only tests' class files should be removed.
  
Can you elaborate how you would implement this so that cleaning test classes shouldn't clean compile classes? I don't think there's a way to avoid that given how sbt deals with configuration dependencies.

dwijnand: I don't know, I would have to experiment.

laughedelic: > it shouldn't make assertions about whether or not there is a streams directory, etc

@dwijnand OK. What about `resolution_cache`? I just want to check that whatever else is generated by sbt is kept.

dwijnand: Same with `resolution_cache`. I agree with the intent. My idea was something along the lines of:

```
test:compile
# sanity check
exists classes
exists test-classes

compile:clean
absent classes
exists test-classes

test:compile

test:clean
exists classes
absent test-classes
```

main/src/main/scala/sbt/Defaults.scala
laughedelic: I'm fine with either way. I think scoping by task makes sense if there will be more variants of `clean`. But if it addresses only compilation task, I agree with @dwijnand that it's better to scope it by configuration. Because `compile::clean` gives a false impression that it should magically work with other tasks (cleaning their specific artifacts).

dwijnand: > Because `compile::clean` gives a false impression that it should magically work with other tasks (cleaning their specific artifacts).

Exactly.

dwijnand: Most things about sbt's scopes are all about compilation/classpath.

sbt/src/sbt-test/actions/compile-clean/test
laughedelic: OK. How about creating a file in `target/` and checking that it's not deleted? (because normal `clean` would delete it).

Also I have to check for existence of specific class files, because the root class folders are preserved.

main/src/main/scala/sbt/Defaults.scala
jvican: Hmm, `test:clean` is not the same to `test:compile::clean`. The latter is explicit and portrays better the original intent: only deleting class files. `test:clean` already has a meaning now and is not clear about what it should clean, so I strongly favor `test:compile::clean`.
  
>  Because compile::clean gives a false impression that it should magically work with other tasks (cleaning their specific artifacts).
  
Well, this already happens with `test:clean`. You would expect sbt to only delete things generated by the `Test` configuration, but it will delete things generated by `Compile` and `Test`.

sbt/src/sbt-test/actions/compile-clean/test
dwijnand: > OK. How about creating a file in `target/` and checking that it's not deleted? (because normal `clean` would delete it).

Sounds good.

> Also I have to check for existence of specific class files, because the root class folders are preserved.

Yep - I was just using short-hand.

main/src/main/scala/sbt/Defaults.scala
dwijnand: > `test:clean` already works now and is not clear about what it should clean

> Well, this already happens with `test:clean`. You would expect sbt to only delete things generated by the `Test` configuration, but it will delete things generated by `Compile` and `Test`.

I'm not sure what you mean by "works" then. Currently `test:clean` just delegates to `clean`. Thus #3542.

There is actually a sys property to disable that delegation and I was playing with it the other day: https://github.com/sbt/sbt/blob/05c2c506b2b218d8ed8befcef843951391c07be6/main/src/main/scala/sbt/internal/Act.scala#L446

It's interesting because it mean that `test:clean` wouldn't run because it wouldn't delegate to `clean`, but `test` would continue to delegate to `test:test` and therefore run. I want to revisit it for sbt 2.

jvican: > I'm not sure what you mean by "works" then. Currently test:clean just delegates to clean. Thus #3542.
  
Yup, I agree. This is why I said:
  
>  this already happens with test:clean. You would expect sbt to only delete things generated by the Test configuration, but it will delete things generated by Compile and Test.
  
That sys property is interesting, didn't know it.
  
I'm trying to think how we solve this puzzle to get the intended behaviour in `test:compile::clean`.

jvican: It turns out we've been missing that the actual implementation does only delete test class files, so we're good. Turning the `test` into the following makes the test pass:
```
> test:compile
$ exists target/scala-2.12/resolution-cache/
$ exists target/streams/
$ exists target/scala-2.12/classes/A.class
$ exists target/scala-2.12/test-classes/B.class

> test:compile::clean
$ exists target/scala-2.12/resolution-cache/
$ exists target/streams/
# it should clean only test class files
$ exists target/scala-2.12/classes/A.class
$ absent target/scala-2.12/test-classes/B.class

> compile:compile::clean
$ exists target/scala-2.12/resolution-cache/
$ exists target/streams/
# it should clean only compile class files
$ absent target/scala-2.12/classes/A.class
$ absent target/scala-2.12/test-classes/B.class
```

laughedelic: > There is actually a sys property to disable that delegation and I was playing with it the other day

Wow! I just told @jvican yesterday that delegation is confusing and I'd like sbt to fail when I mix up scopes and get a delegated value without realising it.



> It turns out we've been missing that the actual implementation does only delete test class files, so we're good.

👍 done in f2aab7849472611cad1285793c0b455e246f0024

sbt/src/sbt-test/actions/compile-clean/test
laughedelic: Addressed in 4bba70b0419aff9e50bf0460efed0b5a554ceb4d

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1098,12 +1105,10 @@ object Defaults extends BuildCommon {
   }
 
   /** Implements `cleanFiles` task. */
-  def cleanFilesTask: Initialize[Task[Vector[File]]] =
```
dwijnand, line 1101: Can't get rid of this, for binary compatibility (MiMa-enforced) reasons.

laughedelic: Does f486d9d fix it?

dwijnand: Travis CI will let us know.

laughedelic: Travis is green ✅

laughedelic:
@dwijnand @jvican 
I implemented (and tested) same change for 
```
{compile,test}:{managedSources,managedResources}::clean
``` 
to clean only generated (re)sources. 

Can I add it here or should I submit a separate PR?

main/src/main/scala/sbt/Defaults.scala
laughedelic: I should also add `cleanKeepFiles` here and fix dependency on it in `cleanFilesTask`, because otherwise user cannot override it in this particular scope.

dwijnand:
I'd hold fire on any additional feature for the moment (obviously hold onto those SHAs!): this is still not implemented as `compile:clean` and `test:clean`.

jvican:
I'm against that. I've already explained why in this PR several times. Can you elaborate on why *only* class files should be cleaned by configurations?

laughedelic:
I didn't have any strong opinion about this in the beginning, but over the couple of days at Lambda World I talked with @jvican and he basically opened my eyes on some basic things about scoping in sbt that I didn't understand (or understood wrong) before. 

I think that this should be scoped by task, because it makes sense conceptually: you clean only the products of this task (in whatever configuration scope). And the source/resource-generation tasks mentioned above are another example where `clean` could be useful and if it's not scoped by task, how do you distinguish it from the `compile` task scoped one.

Another thing is the concern from https://github.com/sbt/sbt/pull/3678#discussion_r147378758:

> Because `compile::clean` gives a false impression that it should magically work with other tasks (cleaning their specific artifacts).

 but I realised that this is something general about sbt design: because of the scope delegation user should just never have such expectation and always inspect keys he uses to see if they are actually defined in the given scope or delegated to some other.

eed3si9n (changes requested):
I also agree with @dwijnand that the scoping of additional `clean` should be configuration like `Compile` and `Test`.
After all, that's what the expectation of https://github.com/sbt/sbt/issues/3542 stated:

> `compile:clean` should clean only the compilation artifacts.

jvican:
`compile:clean` will clean not only class files but any task result produced by the configuration `Compile`. The ticket I opened in https://github.com/sbt/sbt/issues/3542 mentions only compilation artifacts, the compilation artifacts are only the class files and analysis files.

eed3si9n:
I agree it's not perfect, but compilation could happen as part of `run` or `test`, so I don't think there's a point in scoping it specific to `compile`. If we are being super specific, I'm guessing that these files are generated in all sorts of subtasks under `compile` too.

dwijnand:
See also @retronym's super simple `cleanClasses` impl: https://github.com/scala/scala/pull/6256#issuecomment-366142058

jvican:
> but compilation could happen as part of run or test
  
Wait, even if compilation happens because the user does `run` or `test`, how does that change the fact that `compile::clean` will only clean class files and analysis files? Surely we're not talking about having the same `clean` task for `run` or `test`, so I don't think the argument makes a difference.
  
Maybe someone can explain to me what's the value of having `clean` be scoped to the `Compile` configuration instead of the `compile`? I'm really missing the point. Also, what happens with any other configuration that triggers `compile`, like `IntegrationTest` or `Test` or `Jmh`?

eed3si9n:
> Maybe someone can explain to me what's the value of having `clean` be scoped to the `Compile` configuration instead of the `compile`?

For me it comes down to usability and symmetry. Configuration scoping, in general I think is easier to understand for users.

If you want to add scalacOptions in Compile configuration we do:

```
Compile / scalacOptions += "-deprecation"
```

Same goes for lots of other settings related to compilation. None of them are scoped to `compile` task. If you want your `Compile` products to be cleaned it should naturally be:

```
> Compile/clean
```

If you want to compile or clean things in `Test` scope:

```
> Test/compile
> Test/clean
```

jvican:
@eed3si9n But note that your proposal is not at odds with mine, we can still scope it in the compile task, and then delegate to it from the configuration scope. The idea is that the configuration scope, in the future, removes all products produced by other tasks scoped in `Compile`, whereas `compile::clean` is strictly charged with removing compilation products. This benefits people like me that only want the class files to be removed, and benefits other plugin authors too that may want to enrich `Compile/clean` with cleaning resources generated by their plugins.

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/4539/commits/172c8e9a0baa90749f5ceb609a7e465c2c721d13.

PR #3682, opened 2017-10-27
Don't fail loading the build if npm is missing or we are on Windows

OlegYch:
Previously loading the build would fail if npm was not installed, not it just warns.
Also made it use correct npm command on Windows.


dwijnand (commented):
LGTM, @OlegYch, thank you. But what's with making build.sbt executable?

OlegYch:
@dwijnand hm no  idea, do you want me to fix it?

dwijnand:
Yes please.

PR #3688, opened 2017-10-28
cleanKeepFiles is ignored if cleanFiles is overriden

laughedelic:
### Steps
1. `touch target/keep`
1. `build.sbt`:
    ```scala
    cleanFiles := Seq(target.value)
    cleanKeepFiles := Seq(target.value / "keep")
    ```
2. Run `sbt clean`.

### Problems

File `target/keep` had to be preserved, but it's deleted.

### Expectations

Key descriptions say:
* `cleanFiles`: The files to recursively delete during a clean.
* `cleanKeepFiles`: Files or directories to keep during a clean. Must be direct children of target.

So I would expect that running `clean` will delete everything under `cleanFiles` _except_ those in `cleanKeepFiles`. 

### Notes

The reason is that `cleanFiles` is defined through `cleanFilesTask`, which uses `cleanKeepFiles` directly, so if you override `cleanFiles`, it doesn't depend on `cleanKeepFiles` anymore. 

So I changed `cleanFilesTask` to define it in terms of the `cleanFiles` and `cleanKeepFiles` settings (their diff with the limitation that keep-files have to be direct children of the directories in `cleanFiles`).

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -517,9 +517,9 @@ object Defaults extends BuildCommon {
     ))
 
   lazy val projectTasks: Seq[Setting[_]] = Seq(
-    cleanFiles := cleanFilesTask.value,
+    cleanFiles := Seq(managedDirectory.value, target.value),
     cleanKeepFiles := historyPath.value.toVector,
-    clean := (Def.task { IO.delete(cleanFiles.value) } tag (Tags.Clean)).value,
+    clean := (Def.task { IO.delete(cleanFilesTask.value) } tag (Tags.Clean)).value,
     consoleProject := consoleProjectTask.value,
     watchTransitiveSources := watchTransitiveSourcesTask.value,
     watch := watchSetting.value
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -518,9 +518,9 @@ object Defaults extends BuildCommon {
     ))
 
   lazy val projectTasks: Seq[Setting[_]] = Seq(
-    cleanFiles := cleanFilesTask.value,
+    cleanFiles := Seq(managedDirectory.value, target.value),
     cleanKeepFiles := historyPath.value.toVector,
-    clean := (Def.task { IO.delete(cleanFiles.value) } tag (Tags.Clean)).value,
+    clean := (Def.task { IO.delete(cleanFilesTask.value) } tag (Tags.Clean)).value,
```
eed3si9n, line 523: I think the existing implementation works as designed.
I think it's more useful to have definitive task that returns the list of files to delete, that we can use as dry run.

laughedelic: I'm not sure I get you right. Do you say that this solution breaks some intentional design or that there is no issue at all? (i.e. if you override `cleanFiles` and `cleanKeepFiles` and those get deleted anyway, it's normal)

eed3si9n: Problem with this implementation is that I lose the ability to preview what's about to be cleaned.
Whoever overriding `cleanFiles` needs to be responsible for keeping the existing behavior.

PR #3689, opened 2017-10-29
Don't fail loading the build if npm is missing or we are on Windows

OlegYch:
Previously loading the build would fail if npm was not installed, now it just warns.
Also made it use correct npm command on Windows.
(resubmit of #3682 without file mode changes)


dwijnand (approved):
Thanks!

eed3si9n (approved):

Issue #3644, opened 2017-10-14
server needs to include the request id in textDocument/didSave responses


kiritsuku:
Open a Scala file and save it. After saving, a `textDocument/didSave` is sent by vscode. sbt can already handle the request but it doesn't response correctly. The response needs to include the id of the request.

Errors like these are shown in vscode (only the `jsonrpc` and `result` but not the `id` field is present):
```
Received message which is neither a response nor a notification message:
client.js:20
{
    "jsonrpc": "2.0",
    "result": {
        "type": "StringEvent",
        "level": "info",
        "message": "Compiling 1 Scala source to /home/antoras/tmp/vscode-sbt-test/target/scala-2.12/classes ...",
        "channelName": "channel-1",
        "execId": ""
    }
}
Received message which is neither a response nor a notification message:
client.js:20
{
    "jsonrpc": "2.0",
    "result": {
        "type": "StringEvent",
        "level": "info",
        "message": "Done compiling.",
        "channelName": "channel-1",
        "execId": ""
    }
}
```

laughedelic:
I got a little insight on why this is happening:
1. `textDocument/didSave` is a [notification message](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#notification-message), which doesn't have an ID:

    > A processed notification message must not send a response back. They work like events.
2. sbt and treats any incoming messages as request messages and handles it in [`onRequestMessage`](https://github.com/sbt/sbt/blob/1.x/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala#L37-L62). A request message has to have an ID, so `didSave` notification becomes a request with an empty ID:

    ```
    [debug] onRequestMessage: JsonRpcRequestMessage(2.0, , textDocument/didSave, {"textDocument":{"uri":"file:///..."}}})
    ```
    Here you can see after `2.0,` this empty `String` ID.

3. sbt sends back these log messages as response messages with the same ID, which is empty so it just doesn't add them to the JSON

P.S. I'm working on a fix.

laughedelic:
I think https://github.com/sbt/sbt/pull/3740 and https://github.com/sbt/sbt/pull/3779 fix this.

BrianHotopp:
[...]

PR #3712, opened 2017-11-04
LSP: added handler for notification messages

laughedelic:
See https://github.com/sbt/sbt/issues/3644#issuecomment-341299860.

Currently all incoming messages are treated as [`RequestMessage`](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#requestmessage)s. It's is wrong, because this way [`NotificationMessage`](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#notification-message)s are treated as requests with an _empty_ request ID. Notifications don't have an ID and don't need a response (while `RequestMessage`s do require response).

This implementation is far from perfect, so I'm ready to change it according to the feedback.


eed3si9n (approved):
LGTM
Could you rebase to the latest?

eed3si9n:
I am closing this in favor of https://github.com/sbt/sbt/pull/3779

laughedelic:
OK 👌

PR #3634, opened 2017-10-12
[1.0.x] Track sources in base directory non-recursively

dwijnand:
Using a recursive Source meant that ~ looked into target. If you have
any source generators and use ~ with anything the invokes them, like
~compile, that means that the act of generating sources triggers ~ to
re-execute compile (perhaps only on macOS where the NIO WatchService
just polls, after an initial delay).

Requires sbt/io#78

Fixes #3501

PR #3722, opened 2017-11-08
Merge 1.0.x

eed3si9n:
Merge 1.0.x branch onto 1.x.
I needed to resolve conflict for `project/Dependencies.scala` so I'm doing this from a wip branch.


dwijnand:
Something's gone very wrong:

```
[error] /home/travis/build/sbt/sbt/main/src/main/scala/sbt/Defaults.scala:330:86: too many arguments (4) for constructor Source: (base: sbt.io.syntax.File, includeFilter: sbt.io.FileFilter, excludeFilter: sbt.io.FileFilter)sbt.internal.io.Source
[error] Note that 'recursive' is not a parameter name of the invoked method.
[error]         if (sourcesInBase.value) Seq(new Source(baseDir, include, exclude, recursive = false))
[error]                                                                                      ^
```

eed3si9n:
@dwijnand I am guessing that a new IO needs to be released.

dwijnand:
Well did the release of sbt/io 1.0.3 go badly? Because it looks like it doesn't include the change I included in sbt/io 1.0.2.

This is where I upgrade to sbt/io 1.0.2 to use the change: https://github.com/sbt/sbt/pull/3634/files.

PR #3742, opened 2017-11-15
 IPC Unix domain socket / Windows named pipe

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/3723

In addition to TCP, this adds support for IPC (interprocess communication) using Unix domain socket and Windows named pipe. The Java socket implementation for IPC is lifted from facebook/nailgun, which is released under Apache v2.

The use of Unix domain socket has performance and security benefits.

Here's screenshot of this running on a Windows virtual machine:
windows_7_64bit__running_

### serverConnectionType

This introduces a new setting key called `serverConnectionType`, which is set to `ConnectionType.Local` by default:

```scala
serverConnectionType := ConnectionType.Local,
```

The other option is `ConnectionType.Tcp`.

### active.json

To communicate to the clients about the connection type, it uses URI with `local:` as the protocol.

```
{"uri":"local:///Users/xxx/.sbt/1.0/server/0845deda85cb41abdb9f/sock"}
```

### DIFF
--- a/main-command/src/main/scala/sbt/internal/server/Server.scala
+++ b/main-command/src/main/scala/sbt/internal/server/Server.scala
@@ -17,13 +17,14 @@ import java.security.SecureRandom
 import java.math.BigInteger
 import scala.concurrent.{ Future, Promise }
 import scala.util.{ Try, Success, Failure }
-import sbt.internal.util.ErrorHandling
 import sbt.internal.protocol.{ PortFile, TokenFile }
 import sbt.util.Logger
 import sbt.io.IO
 import sbt.io.syntax._
 import sjsonnew.support.scalajson.unsafe.{ Converter, CompactPrinter }
 import sbt.internal.protocol.codec._
+import sbt.internal.util.ErrorHandling
+import sbt.internal.util.Util.isWindows
 
 private[sbt] sealed trait ServerInstance {
   def shutdown(): Unit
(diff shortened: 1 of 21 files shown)


eed3si9n:
Travis passed the test. Codacy problems are coming from the Nailgun code's variable names.

dwijnand:
Why does this include a bunch of Nailgun source code?

eed3si9n:
@dwijnand The only available nailgun artifact on Maven Central is 0.9.1, and the code that I am borrowing from them is fairly new stuff for nailgun 2.x.

main-command/src/main/scala/sbt/internal/server/Server.scala
```diff
@@ -38,31 +39,38 @@ private[sbt] object Server {
...
-              new ServerSocket(port, 50, InetAddress.getByName(host))
+            ErrorHandling.translate(s"server failed to start on ${connection.shortName}. ") {
+              connection.connectionType match {
+                case ConnectionType.Local if isWindows =>
+                  new NGWin32NamedPipeServerSocket(pipeName)
+                case ConnectionType.Local =>
+                  prepareSocketfile()
+                  new NGUnixDomainServerSocket(socketfile.getAbsolutePath)
+                case ConnectionType.Tcp => new ServerSocket(port, 50, InetAddress.getByName(host))
+                case ConnectionType.Ssh => sys.error("ssh it not yet supported")
```
dwijnand: Hmm, maybe it's best to introduce this enum value when it's supported.

dwijnand (changes requested):
Please could you:

1. create an issue requesting this nailgun classes be released somewhere, and
2. add somewhere a reference to what commit sha these source files come from (for future reference)

eed3si9n:
@dwijnand https://github.com/facebook/nailgun/pull/109

dwijnand:
Is there an issue about having a binary release on Maven Central?

eed3si9n:
> Is there an issue about having a binary release on Maven Central?

Binary release for Nailgun? 0.9.1 from 2012 is published, but there hasn't been public release yet since. The interesting bits that I am after were committed in 2017. It could mean that 0.9.2 is not yet out, or ppl are releasing in-house as app.

dwijnand:
> Binary release for Nailgun?

Either of nailgun or a subset of nailgun that includes the classes we're importing here.

I see you've created https://github.com/facebook/nailgun/issues/108#issuecomment-347390157 - I'd've preferred a standalone, more easily findable, issue, but I'm not going to make that get in the way of this landing.

dwijnand (approved):

jameskoch2:
Would there be any security benefit to applying these same Sockets to ForkTests? Reading the code it seems like there's a very small window between ForkTests opening a loopback port and accepting at-most-one connections from the child process. So benefit seems minimal, at a glance, but would like to hear other opinions.

eed3si9n:
I think any inter-process communication is worth reviewing to see either Unix domain socket and/or server is applicable.

jameskoch2:
I filed https://github.com/sbt/sbt/issues/3938 after closer inspection of the NamedPipes code, because by my reading it looks like it would still allow other users on the same machine to snoop on traffic.

Issue #3543, opened 2017-09-15
Investigate scripted run/error


eed3si9n:
```
[error] x run / error 
[error]  Cause of test exception: {line 18}  Command succeeded but failure was expected
```

cunei:
The scripted test invokes: `run`. I just determined that in certain cases the jar that is packaged for later use by `scripted` does not contain the classfile of the source file that has just been compiled, but the classfile that were there prior to the invocation of `run`. For example:
Inside `target/scala-2.12/error_2.12-0.1-SNAPSHOT.jar`:
```
  public void main(java.lang.String[]);
    Code:
       0: iconst_0
       1: invokestatic  #21                 // Method java/lang/System.exit:(I)V
       4: return
```
Inside `target/scala-2.12/classes/Spawn$.class`:
```
  public void main(java.lang.String[]);
    Code:
       0: iconst_1
       1: invokestatic  #21                 // Method java/lang/System.exit:(I)V
       4: return
```
I'll do some more debugging to find out why this is happening.

Issue #3545, opened 2017-09-15
Investigate scripted tests/fork-parallel 


eed3si9n:
```
[error] x tests / fork-parallel 
[error]  Cause of test exception: {line 2}  Command succeeded but failure was expected
```

cunei:
@dwijnand, can you please point me to a failed test? I am trying to reproduce, but all seems to work for me with current 1.1.x head.

cunei:
is it possible that you have seen https://github.com/sbt/sbt/pull/3802 ?

dwijnand:
@eed3si9n mentioned failures at https://github.com/sbt/sbt/pull/3816#discussion_r157337712.

Yes I saw #3802.

cunei:
OK, I found this travis run:
https://travis-ci.org/sbt/sbt/jobs/317262269
In its log there is a failed `tests/fork-parallel`. The log refers to sbt's commit 4cae6895. The commit log is here:
https://github.com/sbt/sbt/commits/4cae6895680000bacc6aeab0cc8990a5148d5338
The list of commits includes the fix in #3802, meaning that `templateStats()` is synchronized. So, there is definitely still an issue lingering somewhere. However, I spent quite a bit of time trying to get the failure on my machine, with no luck. To make things more complicated the commit above is an unreferenced commit in GitHub, meaning that it is not attached to any branch or tag; probably a `push -f` left it dangling. It turns out that there is no simple way to fetch or to reconnect a dangling commit left in GitHub, so it's difficult to fetch the same commit locally. Related commits, locally, show success for `tests/fork-parallel`.
What I am trying now is using the GitHub API to create, via REST calls, a new remote branch attached to the dangling commit inside GitHub; that seems to be the only way to retrieve the commit, so that I can test it again using Travis. WIP.

eed3si9n:
What I originally had was the content of #3789 + https://github.com/sbt/sbt/pull/3816/commits/40b86ae41265c49e6e52fa2ac043413be82bd0cc, which failed the `tests/fork-parallel` test.

cunei:
I retrieved the branch, and restarted Travis: it failed at the very same spot again:
https://travis-ci.org/sbt/sbt/jobs/319252979
Yet, it works on my machine.

cunei:
[...]

PR #3743, opened 2017-11-15
Workaround for scala/bug#10605

cunei:
`templateStats()` is not thread-safe in 2.12.x (at least up to 2.12.4).
See scala/bug#10605


eed3si9n (approved):

dwijnand:
@cunei this resolves https://github.com/sbt/sbt/issues/3543 and https://github.com/sbt/sbt/issues/3545, right?

dwijnand (approved):

cunei:
@dwijnand I'm checking...

cunei:
@dwijnand I ran both tests 20 times, with no failures. I propose that we reinstate the tests, and take further action in case it is needed. Submitted PR https://github.com/sbt/sbt/pull/3745

PR #3744, opened 2017-11-16
Port of sbt/zinc#441 to sbt/sbt 1.0.x

cunei:
I was affected by the same log4j exception while testing using scripted, so porting the zinc scripted fix https://github.com/sbt/zinc/pull/441 over to sbt/sbt.
Note: copy/paste unavoidably leads to outdated code; we should strive to factor out common code in a separate library.


dwijnand:
Now we have this fix twice in the same repo 😆 

https://github.com/sbt/sbt/commit/0b33c195efa861d2ec83de284dbab4a56fda15e6

dwijnand (approved):

cunei:
As I was saying, factoring out libraries...

PR #3754, opened 2017-11-21
Remove 2 useless files in protocol

Duhemm:
They have been moved to Contraband since, but never removed.


dwijnand (approved):
:+1: Another case of https://github.com/sbt/contraband/issues/75.

PR #3760, opened 2017-11-22
don't require publishTo specified if publishArtifact is `false`

mpollmeier:
Even with `publishArtifact := false` the user is still forced to define a (dummy) resolver that's never used, e.g. `publishTo := { Some("publishMeNot" at "https://publish/me/not") }`

Otherwise the following error is thrown:
```
publish
[error] java.lang.RuntimeException: Repository for publishing is not specified.
[error]         at scala.sys.package$.error(package.scala:27)
[error]         at sbt.Classpaths$.$anonfun$getPublishTo$1(Defaults.scala:2436)
[error]         at scala.Option.getOrElse(Option.scala:121)
[error]         at sbt.Classpaths$.getPublishTo(Defaults.scala:2436)
[error]         at sbt.Classpaths$.$anonfun$ivyBaseSettings$48(Defaults.scala:1917)
```

Is 1.0.x the right target branch, or should it rather be 1.x?

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1935,8 +1935,10 @@ object Classpaths {
         if (isSnapshot.value) "integration" else "release",
         ivyConfigurations.value.map(c => ConfigRef(c.name)).toVector,
         packagedArtifacts.in(publish).value.toVector,
-        checksums.in(publish).value.toVector,
-        getPublishTo(publishTo.value).name,
+        checksums.in(publish).value.toVector, { //resolvername: not required if publishTo is false
+          val publishToOption = publishTo.value
+          if (publishArtifact.value) getPublishTo(publishToOption).name else "local"
+        },
         ivyLoggingLevel.value,
         isSnapshot.value
       )
--- a/notes/1.2.0/publishTo-not-required-if-not-publishing.md
+++ b/notes/1.2.0/publishTo-not-required-if-not-publishing.md
@@ -1,0 +1,12 @@
+
+[@mpollmeier]: https://github.com/mpollmeier
+
+[#3760]: https://github.com/sbt/sbt/pull/3760
+
+### Fixes with compatibility implications
+
+### Improvements
+
+### Bug fixes
+
+- Remove requirement on `publishTo` if `publishArtifact` is false. [#3760][] by [@mpollmeier][]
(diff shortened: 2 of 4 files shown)


dwijnand:
I think this is a good change. Could you add a test for this use case so we don't regress, please? Either to an existing scripted test or a new one.

Given this isn't (as far as I understand) a regression in behaviour since sbt 0.13, I think it's safer merge this in 1.x, so it participates in a release candidate phase.

2m:
FWIW If `skip in publish := true` is used to disable the pubishing, then no `publishTo` needs to be defined.

mpollmeier:
* rebased onto 1.x and changed the target branch
* added a test (and tested that it fails prior to this patch, becomes green with the patch)

mpollmeier:
the travis errors look like flukes. @dwijnand anything else stopping this to get merged?

dwijnand (approved):
Restarted the failed Travis CI jobs. This LGTM, pending Travis CI.

mpollmeier:
green enough @dwijnand ?

dwijnand (approved):
Sorry for the delay @mpollmeier. Yes.

I added notes as well (and rebased).

eed3si9n:
I actually don't understand the steps, problem, and the expectation of this PR.
As discussed in #3136 and noted by @2m, the right flag to toggle the behavior is `publish / skip`, not `publishArtifact`, which is meaningless when it's unscoped.

eed3si9n:
The description says:

```
publish
[error] java.lang.RuntimeException: Repository for publishing is not specified.
[error]         at scala.sys.package$.error(package.scala:27)
[error]         at sbt.Classpaths$.$anonfun$getPublishTo$1(Defaults.scala:2436)
[error]         at scala.Option.getOrElse(Option.scala:121)
[error]         at sbt.Classpaths$.getPublishTo(Defaults.scala:2436)
[error]         at sbt.Classpaths$.$anonfun$ivyBaseSettings$48(Defaults.scala:1917)
```

so if I reverse engineer the steps would it be:

### steps

1. don't specify `publishTo`, but set `publishArtifact` to `false`.
2. run `publish`.

### problem

```
[error] java.lang.RuntimeException: Repository for publishing is not specified.
[error]         at scala.sys.package$.error(package.scala:27)
[error]         at sbt.Classpaths$.$anonfun$getPublishTo$1(Defaults.scala:2436)
[error]         at scala.Option.getOrElse(Option.scala:121)
[error]         at sbt.Classpaths$.getPublishTo(Defaults.scala:2436)
[error]         at sbt.Classpaths$.$anonfun$ivyBaseSettings$48(Defaults.scala:1917)
```

### expectation

no errors.

Would that be right?

mpollmeier:
Your 'reverse engineer' of the problem from back then is correct, and as you said, if `publish/skip` is the setting we should actually use, that works for me. Feel free to revert this, then.

Issue #1506, opened 2014-08-07
UnknownHostException while running test on Java 7 on Mac OS X


eed3si9n:
## steps

Not sure if this happens every time. (Might need DHCP to change )

```
$ java -version
java version "1.7.0_65"
Java(TM) SE Runtime Environment (build 1.7.0_65-b17)
$ sbt
> test
```
## problem

```
[trace] Stack trace suppressed: run last ivy/*:testListeners for the full output.
[error] (ivy/*:testListeners) java.net.UnknownHostException: foo-computer: foo-computer: nodename nor servname provided, or not known
[error] Total time: 3 s, completed Aug 7, 2014 4:28:01 AM            
Ivy> last testListeners
java.net.UnknownHostException: foo-computer: foo-computer: nodename nor servname provided, or not known
        at java.net.InetAddress.getLocalHost(InetAddress.java:1473)
        at sbt.JUnitXmlTestsListener.<init>(JUnitXmlTestsListener.scala:18)
        at sbt.plugins.JUnitXmlReportPlugin$$anonfun$projectSettings$1.apply(JUnitXmlReportPlugin.scala:27)
        at sbt.plugins.JUnitXmlReportPlugin$$anonfun$projectSettings$1.apply(JUnitXmlReportPlugin.scala:27)
        at scala.Function1$$anonfun$compose$1.apply(Function1.scala:47)
        at sbt.$tilde$greater$$anonfun$$u2219$1.apply(TypeFunctions.scala:42)
        at sbt.std.Transform$$anon$4.work(System.scala:64)
        at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:237)
        at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:237)
        at sbt.ErrorHandling$.wideConvert(ErrorHandling.scala:18)
        at sbt.Execute.work(Execute.scala:244)
        at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:237)
        at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:237)
        at sbt.ConcurrentRestrictions$$anon$4$$anonfun$1.apply(ConcurrentRestrictions.scala:160)
        at sbt.CompletionService$$anon$2.call(CompletionService.scala:30)
        at java.util.concurrent.FutureTask.run(FutureTask.java:262)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
[...]

PR #3766, opened 2017-11-24
Add warning for slow hostname lookups on OS X

eatkins:
I spent a lot of time debugging why it took 5 seconds to run tests each
time. It turns out that if the hostname is not set explicitly on os x,
then getaddrinfo takes 5 seconds to try (and fail) to resolve the dns
entry for the localhostname. This is easily fixed by setting the
hostname, but it is not at all easy to figure out that a slow hostname
lookup is the reason why tests are slow to start.

I don't know if this is a common issue on other platforms, so only issue
the warning on OS X.

Related to #1506

### DIFF
--- a/main/src/main/scala/sbt/internal/Load.scala
+++ b/main/src/main/scala/sbt/internal/Load.scala
@@ -1039,7 +1039,9 @@ private[sbt] object Load {
         // Grab all the settings we already loaded from sbt files
         def settings(files: Seq[File]): Seq[Setting[_]] = {
           if (files.nonEmpty)
-            log.info(s"${files.map(_.getName).mkString(s"Loading settings for project ${p.id} from ", ",", " ...")}")
+            log.info(
+              s"${files.map(_.getName).mkString(s"Loading settings for project ${p.id} from ", ",", " ...")}"
+            )
           for {
             file <- files
             config <- (memoSettings get file).toSeq
--- a/main/src/main/scala/sbt/plugins/JUnitXmlReportPlugin.scala
+++ b/main/src/main/scala/sbt/plugins/JUnitXmlReportPlugin.scala
@@ -30,6 +30,6 @@ object JUnitXmlReportPlugin extends AutoPlugin {
   // It might be a good idea to derive this setting into specific test scopes.
   override lazy val projectSettings: Seq[Setting[_]] =
     Seq(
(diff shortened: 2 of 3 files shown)


lightbend-cla-validator:
Hi @eatkins,

eatkins:
I signed the Lightbend Contributors Licence Agreement.

dwijnand:
Given the only use of the hostname is to embed it into the JUnit results XML, what if we just avoid the whole pitfall and use "localhost"?

eatkins:
@dwijnand My concern with changing the behavior is that users with multiple build agents may rely on the hostname field to determine on which machine their tests ran. An alternative approach would be to make the machine name configurable in the project settings and switch to using localhost by default. This would require existing users to update the machine name when they upgrade sbt though.

dwijnand:
@eatkins OK! Do you want to resolve the merge conflict, or shall I?

eatkins:
Resolved.

dwijnand (approved):
Thanks @eatkins!

rtyley:
Just want to remark that on my Apple M1 Mac Mini (macOS Big Sur 11.5.2) following the advice given in the message from this PR (_"scutil --set HostName "$(scutil --get LocalHostName).local""_) unfortunately **didn't** work for me - `java.net.InetAddress.getLocalHost.getHostName` was still very slow (5 seconds) - even after restart & executing the command with sudo.

Looking around, I came across [this Stackoverflow answer](https://stackoverflow.com/a/39698914/438886), which recommends manually updating `/etc/hosts`. I checked the contents of `/etc/hosts` and the recommended lines weren't there yet, even after having executing the `scutil` commands - so I manually added them:

```
127.0.0.1   localhost roberto-mac.local
::1         localhost roberto-mac.local
```

...immediately, without a restart or further intervention, the `java.net.InetAddress.getLocalHost.getHostName` call became quick again, and the slow-lookup warning from this PR was no longer displayed.

I could open a PR to update the advice, but in the updated advice would it be better to drop `scutil` altogether, or just advise the user to do both?

eatkins:
Aah, funny that I didn't find that thread when I initially added this warning. Given that adding those lines to /etc/hosts seems to work for both old and new versions of OSX, I'd suggest that indeed dropping the scutil commands would be good. It would be nice if we could print a command that the user could copy and paste to add the necessary lines to /etc/hosts similar to the scutil example rather than just telling the user to manually add the lines.

PR #3772, opened 2017-11-27
Create RunFromSourceMain & "sbtOn" command

dwijnand:


dwijnand:
This is failing CI because the new custom "sbt" command is being used for "sbt:scalafmt::test"..

dwijnand:
Renamed to "sbtOn".

eed3si9n:
Is this 1.0.x worthy?

dwijnand:
I targeted 1.0.x so that I can use it when investigating and fixing 1.0.x hotfixes (without having to do a lot of git backflips).

Outside the large test source file, it adds and configures sbt-buildinfo (and even that in test scope) which I can't see impacting the quality and integrity of future 1.0.x hotfixes, but you're in the best position to make that judgement.

eed3si9n (approved):

PR #3774, opened 2017-11-28
Inline always true broadcastStringMessage

dwijnand:
Follow up on my comment https://github.com/sbt/sbt/pull/3621#discussion_r143405881.


main/src/main/scala/sbt/internal/CommandExchange.scala
```diff
@@ -168,34 +168,20 @@ private[sbt] final class CommandExchange {
   }
 
   def publishEvent[A: JsonFormat](event: A): Unit = {
-    val broadcastStringMessage = true
```
eed3si9n, line 171: I was thinking about potentially turning this into a client specific setting.

dwijnand: I see.

PR #3776, opened 2017-11-28
Alt scala parser sync workaround

dwijnand:
Apply retronym's suggestion at https://github.com/scala/bug/issues/10605
instead of the workaround merged in https://github.com/sbt/sbt/pull/3743.


retronym (approved):

retronym:
I doubt the external synchronization in the workaround was a bottleneck, so if you want to be ultra conservative you could leave things that way. But multi-threaded parsing is definitely exercised in Scala IDE, so I think its fine to depend on it here.

Issue #3486, opened 2017-08-30
sbt-maven-resolver no longer available in 1.0.x


dpratt:
(See the guidelines for contributing, linked above)

## steps

Add the maven resolver plugin to any project by putting 

```
addMavenResolverPlugin
```

in project/plugins.sbt

## problem

The project can't be loaded due to the fact that org.scala-sbt:sbt-maven-resolver:1.0.0 cannot be resolved.

## notes

I don't know if this plugin is supported in the 1.0.x series, but the settings hook function is still available in the global API. I propose that it should be either removed, or replaced with an implementation that just logs the fact that the plugin is not available/compatible with 1.0.x

sbt version: *insert sbt version*

dwijnand:
sbt-maven-resolver was lifted and shifted to https://github.com/sbt/sbt-maven-resolver, but hasn't gotten any attention (love) since.. much like `addMavenResolverPlugin`..

slandelle:
Then, why not drop it, or make it noop with a warning log? At least mention in the migration guide `addMavenResolverPlugin` doesn't currently work.

It took me a desk head bangings to figure out why I got this "org.scala-sbt#sbt-maven-resolver;1.0.2: not found" error.

Please tell me how I can help.

dwijnand:
It slipped through the cracks.

A straightforward PR that deprecates `addMavenResolverPlugin` and (if possible) prints some kind of warning on use would most probably be merged quickly.

slandelle:
I'm not familiar with sbt sources yet, but I'll give it a try.

slandelle:
@dwijnand I can't manage to print a warning. How do you grab a logger in a Settings (not in a Task)?

dwijnand:
Use `sLog.value` instead.

slandelle:
Thanks!

slandelle:
See #3781

jgonian:
I discovered `addMavenResolverPlugin` and `sbt-maven-resolver` in the SBT documentation while I was trying to overcome multiple integration issues between SBT and Maven during our company's migration to SBT.

For example, this is the only available way to have libraries released truly in Maven style, including updating the `maven-metadata.xml` file that many plugins are depending upon.

When I read the documentation (http://www.scala-sbt.org/1.0/docs/Combined+Pages.html#Maven+resolver+plugin and #1793) I still think that this is a useful plugin and removing it without putting a better integration with Maven in place, feels like a huge step backwards. 
 
[...]

PR #3781, opened 2017-11-29
Deprecate addMavenResolverPlugin, see #3486

slandelle:
Motivation:

sbt-maven-resolver hasn’t been maintained lately and hasn’t been
released for sbt 1.0. Without this fix, adding addMavenResolverPlugin
will make build fail on unresolved dependency.

Modification:

Deprecate and make `addMavenResolverPlugin` a noop.

Result:

Build no longer crashes on sbt 1.0 when using `addMavenResolverPlugin`.
User will get a warning though.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2981,7 +2981,7 @@ object Classpaths {
           case Predefined.SonatypeOSSSnapshots => Resolver.sonatypeRepo("snapshots")
           case unknown =>
             sys.error(
-              "Unknown predefined resolver '" + unknown + "'.  This resolver may only be supported in newer sbt versions.")
+              "Unknown predefined resolver '" + unknown + "'. This resolver may only be supported in newer sbt versions.")
         }
     }
   }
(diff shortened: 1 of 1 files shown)


slandelle:
@dwijnand Is the CI failure really related to this change?

dwijnand:
No, I restarted the failed Travis CI job.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -3016,11 +3016,16 @@ trait BuildExtra extends BuildCommon with DefExtra {
   /**
    * Adds Maven resolver plugin.
    */
+  @deprecated(
+    "Noop. sbt-maven-resolver is not currently maintained and hasn't been released for sbt 1.0.",
+    "1.0.5")
```
dwijnand, line 3021: I think a better message (where and in the log message below) is:

> The maven resolver plugin is no longer released as a part of sbt. See https://github.com/sbt/sbt-maven-resolver.

(and then we should put some better, current info in that repo's README)

Also the version should be 1.2.0 (when this PR will probably land), or maybe (retroactively) 1.0.0.

dwijnand (changes requested):

main/src/main/scala/sbt/Defaults.scala
slandelle: I guess it should be 1.0.0.

PR #3785, opened 2017-11-30
Bump vscode-sbt-scala version to 0.1.0

eed3si9n:


eed3si9n:
Cancelling Travis since there's nothing to check there.

Issue #3775, opened 2017-11-28
~ in polling mode is not triggered by IntelliJ IDEA saving files in macOS


mechkg:
## steps

1. Create a barebones SBT project, e.g. using new SBT project in IntelliJ IDEA. Add a test source file, e.g.

```
object Test extends App {
  println ("Hello SBT!")
}
```

2. Start SBT and run any command prefixed with **~**, e.g.

    ~run

3. Edit the file in IDEA and save it.

## problem

SBT does not detect file changes and does not re-run the command.

## expectation

SBT should detect file changes and re-run the command.

## notes

Changes made by other editors (e.g. TextMate, nano) do trigger SBT 1.0.4!

SBT 0.13.16 triggers correctly regardless of what program modified the file.

SBT 1.0.4 triggers correctly in nio mode (via `-Dsbt.watch.mode=nio`) albeit with a delay.

SBT version: 1.0.0 - 1.0.4
macOS version: High Sierra 10.13.1
Java version: 1.8.0_151

mechkg:
See also https://github.com/sbt/io/issues/86

xerial:
@mechkg Good catch!

dwijnand:
Upstream https://github.com/sbt/io/issues/86

eatkins:
I referenced this in #3695 but, as a temporary fix, I added an sbt plugin that replaces the WatchService for OSX with one that uses the Apple File System Events API. https://github.com/swoval/MacOSXWatchService

dwijnand:
Re-opening to track sbt/io needs releasing for inclusion in 1.1.0-RC2.

OlegYch:
@dwijnand this seems to be specific to IDEA
happens to me on windows, with sbt 1.0.0-1.0.4, with both polling and io, doesn't happen if i eg edit the file with notepad or use IO.touch

OlegYch:
the only difference i can see is that when IDEA saves file - directory last modified time is updated, but not when i use touch

eed3si9n:
New io with the fix was merged in #3816.
If the issue persists with 1.1.0-RC2, then we will reopen this case.

dwijnand:
This issue was about macOS though, where it's using polling. If there are issues with NIO and/or IntelliJ, on Windows, they should be reported.

hedefalk:
[...]

Issue #3695, opened 2017-10-30
~compile only watches some files


blast-hardcheese:
### Setup

```
mkdir -p project src/main/scala/foo
echo sbt.version=1.0.3 > project/build.properties
echo 'println("woo")' > src/main/scala/App.scala
echo 'println("woo")' > src/main/scala/foo/bar.scala
```

### Launch sbt `~compile`

```
$ sbt
sbt> ~compile
```

### Change files

1. Changes to `src/main/scala/App.scala` always trigger a recompile
1. Changes to `src/main/scala/foo/bar.scala` never trigger a recompile
1. Create a new file, `src/main/scala/foo/baz.scala`. Changes to `baz.scala` trigger a recompile until the next full sbt launch (close and open again).

## problem

`~` doesn't always track files in directories that should be covered by the file watcher

## expectation

`~` always tracks files in directories that are covered by the file watcher

## notes

sbt version: 1.0.3

blast-hardcheese:
@dwijnand This is the issue you requested of me from Gitter, please let me know if I can be of any help going forward

blast-hardcheese:
Possibly related to #3687

dwijnand:
Thanks @blast-hardcheese.

OlegYch:
for me `~` doesn't see changes in tests, but when i modify main source `~testOnly` runs twice
same for `~compile` though

dwijnand:
I can't reproduce this. I followed the given steps and it all works as intended on my box:

```scala
> ~compile
[info] Compiling 2 Scala sources to /s/t-watches/target/scala-2.12/classes ...
[error] /s/t-watches/src/main/scala/App.scala:1:1: expected class or object definition
[error] println("woo")
[error] ^
[error] /s/t-watches/src/main/scala/foo/bar.scala:1:1: expected class or object definition
[error] println("woo")
[error] ^
[error] two errors found
[error] (compile:compileIncremental) Compilation failed
[error] Total time: 2 s, completed 16-Nov-2017 16:05:04
1. Waiting for source changes... (press enter to interrupt)
[info] Compiling 2 Scala sources to /s/t-watches/target/scala-2.12/classes ...
[error] /s/t-watches/src/main/scala/App.scala:1:1: expected class or object definition
[error] println("I've changed App.scala")
[error] ^
[error] /s/t-watches/src/main/scala/foo/bar.scala:1:1: expected class or object definition
[error] println("woo")
[error] ^
[error] two errors found
[...]

PR #3790, opened 2017-12-01
Cache WatchService in continuous builds

eatkins:
I noticed that my custom WatchService was never cleaned up by sbt and
realized that after every build we were making a new WatchService. At
the same time, we were reusing the WatchState from the previous run,
which was using the original WatchService. This was particularly
problematic because it prevented us from registering any paths with the
new watch service. This may have prevented some of the file updates
from being seen by the watch service. Moreover, because we lost the
reference to the original WatchService, there was no way to clean it up,
which was a resource leak.

May be related to #3775, #3695


dwijnand (approved):
LGTM, and particularly sorry @eatkins for this PR not getting any attention for such a long time. It just slipped through the cracks - it wasn't intentional on our part.

eatkins:
Thanks. I assume best intent and recognize that everyone is busy.

PR #3794, opened 2017-12-05
update mimaPreviousArtifacts

xuwei-k:


dwijnand (approved):

build.sbt
```diff
@@ -75,10 +75,9 @@ def testedBaseSettings: Seq[Setting[_]] =
   baseSettings ++ testDependencies
 
 val mimaSettings = Def settings (
-  mimaPreviousArtifacts := Set(
-    organization.value % moduleName.value % "1.0.0"
-      cross (if (crossPaths.value) CrossVersion.binary else CrossVersion.disabled)
-  )
+  mimaPreviousArtifacts := (0 to 4).map { v =>
```
eed3si9n, line 78: What's the policy here? Do we keep adding all the hot fixes and and feature releases?
I'd say we should do both the earliest and the latest.

eed3si9n (changes requested):

build.sbt
xuwei-k: I think latest is indispensable

eed3si9n: ok. So how about `Set("1.0.0", "1.0.4")`?

dwijnand: I approved of doing this this way because it has no impact on PR validation. It looks to have bumped the time for the first Travis CI job (where we check MiMa) from about 7m30 to 8m30, where there are other jobs that take 20-30m. So re-:+1: from me.

eed3si9n: What happens when we release 1.1.0?

dwijnand: Add it to the list.

eed3si9n (approved):

PR #3795, opened 2017-12-05
Only the first session starts the server

eed3si9n:
Currently the server will try to start even if there are existing sbt sessions. This causes the second session to take over the server for Unix domain socket.

This adds a check before the server comes up and make sure that the socket is not taken.

Another thing that was happening was that even when the server didn't start, it was going through the shutdown sequence when sbt exists, deleting the portfile. This is fixed in 
2b2c1f0.

### DIFF
--- a/main-command/src/main/java/sbt/internal/NGUnixDomainSocket.java
+++ b/main-command/src/main/java/sbt/internal/NGUnixDomainSocket.java
@@ -29,6 +29,8 @@ import java.nio.ByteBuffer;
 
 import java.net.Socket;
 
+import java.util.concurrent.atomic.AtomicInteger;
+
 /**
  * Implements a {@link Socket} backed by a native Unix domain socket.
  *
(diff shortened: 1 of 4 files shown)


main-command/src/main/scala/sbt/internal/server/Server.scala
```diff
@@ -54,15 +54,17 @@ private[sbt] object Server {
...
-                case ConnectionType.Tcp => new ServerSocket(port, 50, InetAddress.getByName(host))
-              }
+            connection.connectionType match {
+              case ConnectionType.Local if isWindows =>
+                // Named pipe already has an exclusive lock.
+                addServerError(new NGWin32NamedPipeServerSocket(pipeName))
+              case ConnectionType.Local =>
+                tryClient(new NGUnixDomainSocket(socketfile.getAbsolutePath))
+                prepareSocketfile()
+                addServerError(new NGUnixDomainServerSocket(socketfile.getAbsolutePath))
```
dwijnand, line 64: For both `ConnectionType.Local ` and for `ConnectionType.Tcp`, why do the tried sockets not match the used sockets?

main-command/src/main/java/sbt/internal/NGUnixDomainSocket.java
```diff
@@ -29,6 +29,9 @@
 
 import java.net.Socket;
 
+import java.util.concurrent.atomic.AtomicInteger;
+import com.sun.jna.ptr.IntByReference;
```
dwijnand: unused?

main-command/src/main/scala/sbt/internal/server/Server.scala
eed3si9n: 1. `tryClient(...)` pretends to be a client and hits the server's URL as if it's there. Thus, if the socket fails it's a success (no server yet), and if the socket connects it's a failure (server is already up).
2. The second socket is the real server socket.

main-command/src/main/java/sbt/internal/NGUnixDomainSocket.java
eed3si9n: Yes.

main-command/src/main/scala/sbt/internal/server/Server.scala
dwijnand: Ah right, gotcha.

dwijnand (approved):

PR #3800, opened 2017-12-07
Do not try to add the "NOTICE" file if it does not exist

cunei:
The notice file was always getting added to unmanagedResources,
even when it did not actually exist.

### DIFF
--- a/project/Util.scala
+++ b/project/Util.scala
@@ -197,7 +197,7 @@ object Licensed {
 
   def settings: Seq[Setting[_]] = Seq(
     notice := (baseDirectory.value / "NOTICE"),
-    unmanagedResources in Compile ++= notice.value +: extractLicenses.value,
+    unmanagedResources in Compile ++= extractLicenses.value,
     extractLicenses := extractLicenses0((baseDirectory in ThisBuild).value,
                                         notice.value,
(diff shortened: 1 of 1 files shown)


cunei:
This pull request needs to be merged to `1.x`, so that I can complete validation for the new timestamps.

dwijnand (approved):
LGTM.

I'm confused though, what was this failing and in what circumstance?

eed3si9n (changes requested):
This requires some discussion. It's a sign that mapping became unintentionally stricter because of milli sec.

cunei:
Eugene and I just discussed the matter; I will restore the mapping semantics to "map something that may or may not exist", for the sake of compatibility. Closing the PR.

eed3si9n:
@dwijnand The new `IO.getModifiedTime(file0)` throws exception when the input doesn't exist. It alone sounds reasonable, but I don't think the strictness should propagate silently to other parts of sbt without understanding the ramifications. I think Toni ran into this while using freshly baked sbt to build sbt.

dwijnand:
Oh right. As sbt is a build tool, which uses files all the time, that's not a change in semantics we can just slip into anything but a major version change, imo.

PR #3807, opened 2017-12-12
Remove warnings

dwijnand:


main-actions/src/main/scala/sbt/Doc.scala
```diff
@@ -23,93 +19,45 @@ import sbt.internal.util.ManagedLogger
...
-    )
 
+  @deprecated("Going away", "1.1.1")
+  def javadoc(
+      label: String,
+      cacheStoreFactory: CacheStoreFactory,
+      doc: JavaTools,
+      log: Logger,
+      reporter: Reporter,
+  ): Gen = ???
```
eed3si9n, line 43: Why `???` instead of removing it or delegating it to somewhere?

main/src/test/scala/PluginsTest.scala
```diff
@@ -39,18 +39,18 @@ object PluginsTest extends Specification {
...
     "generates a detailed report on conflicting requirements" in {
-      deducePlugin(T && U, log) must throwAn[AutoPluginException](message =
-        """Contradiction in enabled plugins:
-  - requested: sbt.AI\$T && sbt.AI\$U
-  - enabled: sbt.AI\$U, sbt.AI\$T, sbt.AI\$A, sbt.AI\$Q, sbt.AI\$R, sbt.AI\$B
-  - conflict: sbt.AI\$Q is enabled by sbt.AI\$A && sbt.AI\$B; required by sbt.AI\$T, sbt.AI\$R; excluded by sbt.AI\$U
-  - conflict: sbt.AI\$R is enabled by sbt.AI\$Q; excluded by sbt.AI\$T""")
+      deducePlugin(T && U, log) must throwAn[AutoPluginException](
+        message = s"""Contradiction in enabled plugins:
+  - requested: sbt.A\\$$T && sbt.A\\$$U
```
eed3si9n: Should this be `AI\\$$T` etc?

eed3si9n (changes requested):
A few comments.

main-actions/src/main/scala/sbt/Doc.scala
dwijnand: It's public API, and I think a non-fatal runtime error, that can be handled, is better than a fatal linking error.

main/src/test/scala/PluginsTest.scala
dwijnand: Well spotted, I messed up my find/replace - fixed.

dwijnand:
This is now green, so it's ready for review.

Note that Codacy is red because of 3 cases of lines where I didn't introduce the issue (catching Exception) and 1 case of not putting a call-by-name parameter in the last position, which I think is the right decision in this case.

main-actions/src/main/scala/sbt/Doc.scala
eed3si9n: I don't think we should break this until sbt 2.x unless there's a really good reason for it.

dwijnand: It's already broken, I just made it more obviously so (and got rid of a bunch of unused warnings in the process):

https://github.com/sbt/sbt/blob/v1.1.0-RC1/main-actions/src/main/scala/sbt/Doc.scala#L63

eed3si9n (approved):

PR #3808, opened 2017-12-13
Add log4j2 configuration file

### LINKED ISSUE #3787: "ERROR StatusLogger No log4j2 configuration file found"
## steps

1. Run sbt/sbt using 1.1.0-RC1.

## problem

```
$ sbt
ERROR StatusLogger No log4j2 configuration file found. Using default configuration: logging only errors to the console. Set system property 'org.apache.logging.log4j.simplelog.StatusLogger.level' to TRACE to show Log4j2 internal initialization logging.
```

## expectation

No error message.

eed3si9n:
Fixes #3787

### DIFF
--- a/sbt/src/main/resources/log4j2.xml
+++ b/sbt/src/main/resources/log4j2.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="WARN">
+  <Appenders>
+    <Console name="Console" target="SYSTEM_OUT">
+      <PatternLayout pattern="[%t] %-5level %logger{36} - %msg%n"/>
+    </Console>
+  </Appenders>
+  <Loggers>
+    <Root level="error">
+      <AppenderRef ref="Console"/>
+    </Root>
+  </Loggers>
+</Configuration>


dwijnand:
What change caused the error to appear? We introduced log4j2 in 1.0.0, but the error is only showing now (in 1.1.0-RC1).

eed3si9n:
@dwijnand I think I know what's up. Running as `sbt -Dorg.apache.logging.log4j.simplelog.StatusLogger.level=TRACE` gives you lots of info.

Here's a stack trace I found:

```
	at org.apache.logging.slf4j.Log4jLoggerFactory.getLogger(Log4jLoggerFactory.java:29)
	at org.slf4j.LoggerFactory.getLogger(LoggerFactory.java:358)
	at scalacache.caffeine.CaffeineCache.<init>(CaffeineCache.scala:23)
	at scalacache.caffeine.CaffeineCache$.apply(CaffeineCache.scala:77)
	at scalacache.caffeine.CaffeineCache$.apply(CaffeineCache.scala:69)
	at sbt.StandardMain$.<init>(Main.scala:106)
	at sbt.StandardMain$.<clinit>(Main.scala)
```

This points to line 23 of CaffeineCache.scala. This stuff was added newly in 1.1.

```scala
  override protected final val logger =
    LoggerFactory.getLogger(getClass.getName)
```

Here, `LoggerFactory` is `org.slf4j.LoggerFactory`, and we have a binding so it's now creating a log4j logger using the default `Log4jLoggerFactory`. All this is good.

The problem is that sbt doesn't supply setting for how log4j logs should be configured, since I've been configuring everything programmatically. The added config file solves this.

dwijnand:
Ah! I proposed an alternative fix: https://github.com/sbt/sbt/pull/3814.

eed3si9n:
Sounds good. Closing this in favor of yours.

PR #3814, opened 2017-12-15
Make CaffeineCache a lazy val

### LINKED ISSUE #3787: "ERROR StatusLogger No log4j2 configuration file found"

dwijnand:
This is to avoid it initialising Log4J2 (via SLF4J), which we initialise
ourselves programmatically in LogExchange. Also there's no need to
removeAll in initialState.

Fixes #3787

My proposed alternative to #3808.

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -103,7 +103,7 @@ object StandardMain {
   private[sbt] lazy val exchange = new CommandExchange()
   import scalacache._
   import scalacache.caffeine._
-  private[sbt] val cache: Cache[Any] = CaffeineCache[Any]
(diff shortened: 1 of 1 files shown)


dwijnand:
Forgot to mention: I tested this locally and it removes the "ERROR" warning.

eed3si9n (approved):

PR #3824, opened 2017-12-19
Add Windows testing

eed3si9n:
This adds rudimentary scripted testing on Windows on AppVeyor.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -311,8 +311,7 @@ lazy val commandProj = (project in file("main-command"))
   .settings(
     testedBaseSettings,
     name := "Command",
-    libraryDependencies ++= Seq(launcherInterface, sjsonNewScalaJson.value, templateResolverApi,
-      jna, jnaPlatform),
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -310,8 +311,7 @@ lazy val commandProj = (project in file("main-command"))
   .settings(
     testedBaseSettings,
     name := "Command",
-    libraryDependencies ++= Seq(launcherInterface, sjsonNewScalaJson.value, templateResolverApi,
-      jna, jnaPlatform),
+    libraryDependencies ++= Seq(launcherInterface, sjsonNewScalaJson.value, templateResolverApi),
```
dwijnand, line 314: Why are we dropping JNA? Did we end up not using it anymore in sbt?

dwijnand (approved):
LGTM, just 1 question.

build.sbt
eed3si9n: I am dropping it here to align JNA version with the one introduced by IO.

dwijnand: 👌

PR #3828, opened 2017-12-20
warn about multiple instances once

### LINKED ISSUE #3823: Starting multiple instances of sbt gives: java.io.IOException: sbt server is already running.
## steps
In a single project directory, start multiple instances of sbt

## problem
The 2nd instance of sbt logs an error with message: `java.io.IOException: sbt server is already running.`

## expectation
No error is logged. The sbt instances should either share the sbt server or start separate instances. Either way, no error should be logged.

## notes

sbt version: 1.1.0-RC2

eed3si9n:
Fixes #3823

When you launch a second instance of sbt on a build, prior to this change it was displaying `java.io.IOException: sbt server is already running` on every command. This make it a bit less aggressive, and just display a warning once.

```
[warn] Is another instance of sbt is running on this build?
[warn] Running multiple instances is unsupported
```

**updated**:

```
[warn] sbt server could not start because there's another instance of sbt running on this build.
[warn] Running multiple instances is unsupported
```

### DIFF
--- a/main/src/main/scala/sbt/internal/CommandExchange.scala
+++ b/main/src/main/scala/sbt/internal/CommandExchange.scala
@@ -10,7 +10,7 @@ package internal
 
 import java.io.IOException
 import java.util.concurrent.ConcurrentLinkedQueue
-import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.atomic._
 import scala.collection.mutable.ListBuffer
 import scala.annotation.tailrec
 import BasicKeys.{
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/internal/CommandExchange.scala
```diff
@@ -154,6 +155,11 @@ private[sbt] final class CommandExchange {
           case Some(Success(_)) =>
             // rememeber to shutdown only when the server comes up
             server = Some(x)
+          case Some(Failure(e: AlreadyRunningException)) =>
+            s.log.warn("Is another instance of sbt is running on this build?")
```
dwijnand: I don't think I like the warning to be in question form. Feels too unsure. How about just "Another instance of sbt is running on this build".

eed3si9n: I am unsure. Is this better?

"sbt server could not start likely because there's another instance of sbt running on this build."

dwijnand: I'd be happy with that if it didn't have the "likely".

Then if down the line we find there's another reason why `AlreadyRunningException` is matched here that isn't because of multiple instances we can make the necessary changes.

dwijnand (approved):

PR #3839, opened 2017-12-26
delete "buildinfo.BuildInfo" from sbt main

xuwei-k:
sbt-buildinfo plugin have `buildInfoScopedSettings(Compile)` in default.
I think it is unnecessary. or we should set `buildinfoPackage in Compile` and `buildinfoObject in Compile`

- https://github.com/sbt/sbt-buildinfo/blob/v0.7.0/src/main/scala/sbtbuildinfo/BuildInfoPlugin.scala#L11
- https://github.com/sbt/sbt/commit/9c32c16adee85e8ae5032ebb5140c34c50664053

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -448,7 +448,6 @@ lazy val mainProj = (project in file("main"))
 //  with the sole purpose of providing certain identifiers without qualification (with a package object)
 lazy val sbtProj = (project in file("sbt"))
   .dependsOn(mainProj, scriptedSbtProj % "test->test")
-  .enablePlugins(BuildInfoPlugin)
   .settings(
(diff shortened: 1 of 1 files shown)


dwijnand (approved):
LGTM, thanks @xuwei-k. I should do the same in https://github.com/sbt/io/pull/102/files.

eed3si9n:
Not sure if this is a 1.1.0 showstopper, but certainly it's a good candidate for 1.1.1.
Ref https://twitter.com/xuwei_k/status/945560615082377216

dwijnand:
Damn, my bad :-(

PR #3848, opened 2018-01-03
Bump Scala versions for JDK 9 support

eed3si9n:
/cc @SethTisue, @retronym


dwijnand (approved):
LGTM, but looks like `dependency-management / cached-resolution-classifier` failed.

PR #3855, opened 2018-01-06
Filter auto compiler plugin to classpath JARs

### LINKED ISSUE #3784: sbt 1.0.4 puts the `-sources.jar` and `-docs.jar` in the `compile` config in `ivy.xml`
## steps

Using the following `build.sbt`:
```scala
inThisBuild(Def.settings(
  organization := "com.example",
  scalaVersion := "2.11.12"
))

lazy val compplugin = project.
  settings(
  )
```

```bash
$ echo sbt.version=0.13.16 > project/build.properties
$ sbt compplugin/clean compplugin/publishLocal
[...]
$ cp ~/.ivy2/local/com.example/compplugin_2.11/0.1-SNAPSHOT/ivys/ivy.xml ivy-from-sbt013.xml
$ echo sbt.version=1.0.4 > project/build.properties
$ sbt compplugin/clean compplugin/publishLocal
[...]
$ cp ~/.ivy2/local/com.example/compplugin_2.11/0.1-SNAPSHOT/ivys/ivy.xml ivy-from-sbt104.xml
$ diff -u ivy-from-sbt013.xml ivy-from-sbt104.xml > ivy.xml.diff
```

Results in this gist: https://gist.github.com/sjrd/39b487f3f1f3abb4a5910f583e6d0e9a

## problem

The ivy.xml published by sbt 1.0.4 declares the `src` and `doc` artifacts in the `compile` conf, instead of `sources`/`docs` as it should.

Concretely, this means that when one depends on such a project, the `-sources.jar` and `-docs.jar` are downloaded and added to the compile-time and run-time classpaths!

In the particular case of a scalac compiler plugin, since the `-sources.jar` also contains the `scalac-plugin.xml` identifying the jar as a compiler plugin, depending on it with `% "plugin"` causes *two* `-Xplugin:` options to be passed to scalac: one for `foo.jar` and one for `foo-sources.jar`. Depending on which one comes first (which I have experienced is non-deterministic), scalac will succeed or fail to load the plugin.

## expectation

I expect the same behavior as sbt 0.13, which is also documented as an example of [the Ivy documentation for `<artifact>`](http://ant.apache.org/ivy/history/2.0.0/ivyfile/artifact.html): the `src` artifact should have `conf="sources"` and the `doc` artifact should have `conf="docs"`.

## notes

In addition to the really serious issue above, sbt 1.0.4 also exposes private configurations used by sbt in the ivy.xml, such as `compile-internal`.

sbt version: 1.0.4

eed3si9n:
Fixes #3784

For Ivy repos, sbt 1.x publishes source and scaladoc JARs under `compile` configuration with types `src` and `doc` respectively. The current 0.13 implementation of `autoPlugins(...)` is not filtering them out correctly, and ends up including a source JAR as a compiler plugin.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1933,12 +1933,20 @@ object Classpaths {
 
   @deprecated("Specify the classpath that includes internal dependencies", "0.13.0")
   def autoPlugins(report: UpdateReport): Seq[String] = autoPlugins(report, Nil)
+
+  @deprecated("Specify the jarTypes", "0.13.17")
   def autoPlugins(report: UpdateReport, internalPluginClasspath: Seq[File]): Seq[String] =
-    {
-      val pluginClasspath = report.matching(configurationFilter(CompilerPlugin.name)) ++ internalPluginClasspath
-      val plugins = classpath.ClasspathUtilities.compilerPlugins(pluginClasspath)
-      plugins.map("-Xplugin:" + _.getAbsolutePath).toSeq
-    }
+    autoPlugins(report, internalPluginClasspath, Set("jar", "bundle"))
+
+  def autoPlugins(
+    report: UpdateReport,
+    internalPluginClasspath: Seq[File],
+    jarTypes: Set[String]): Seq[String] = {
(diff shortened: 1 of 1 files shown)


dwijnand (approved):
I trust this is right.

Issue #3656, opened 2017-10-20
Scripted could define default settings in Global for sbt friendliness


jvican:
## problem
The scripted plugin does not follow the good practice of defining the default values for settings in `Global`. I also think that the scopes of the scripted task keys should be revisited. Currently, everything is crammed into `projectSettings`.
  
A direct consequence of this issue is that it's not possible to disable the buffer log globally or per build. There are more examples, but you get the direction where this is going. The configuration of scripted could be more sbt-friendly in this regard, in my opinion.

## expectation
I'd like keys like `scriptedSbt`, `sbtTestDirectory`, `scriptedLaunchOpts` and `scriptedBufferLog` to be scoped globally.

ryan-williams:
Having this issue too; [the recommended way of running scripted tests](http://www.scala-sbt.org/1.x/docs/Testing-sbt-plugins.html#step+2%3A+scripted-plugin) requires settings which can't be set for all modules in a project.

A related question is whether the default value of `scriptedLaunchOpts` and `scriptedBufferLog ` should just be set to the values suggested in that tutorial… are those defaults ever *not* wanted? (if so, they could still be overridden per-project, afaict?)

ryan-williams:
I suppose there is always [this work-around that I recently learned from @2m on gitter](https://gitter.im/sbt/sbt?at=5a4f94e3232e79134dccf1fa):

`project/Scripted.scala`:

```scala
import sbt._
import sbt.Keys._
import sbt.PluginTrigger.AllRequirements
import ScriptedPlugin.autoImport._

object Scripted extends AutoPlugin {
  override def trigger = AllRequirements
  override def requires = super.requires && ScriptedPlugin
  override def projectSettings: Seq[Def.Setting[_]] =
    Seq(
      scriptedLaunchOpts := { scriptedLaunchOpts.value ++
        Seq("-Xmx1024M", "-Dplugin.version=" + version.value)
      },
      scriptedBufferLog := false
    )
}
```

which in general is a good way to set default settings in multi-module projects where some upstream plugin is setting a non-`Global` default that's not overridden by `in ThisBuild`

dwijnand:
My guess is this is the status quo because generally sbt plugins are single project builds. Happy to accept a PR to change this.

Issue #3538, opened 2017-09-15
Add scripted as library dependency for builds of sbt plugins


jvican:
## problem
Sbt plugins have only one way of being tested: scripted. Scripted is used by almost any sbt plugin in the world. Yet, every sbt plugin has to bring the scripted plugin dependency manually.
  
Shortcomings:
1. Users need to get the scripted library from the docs every time they create a new plugin:
`libraryDependencies += { "org.scala-sbt" % "scripted-plugin" % sbtVersion.value }`.
2. Users need to reload their build once the dependency has been added.

## expectation
It would be nice if `scripted` is added automatically for those builds for which `sbtPlugin := true`. I'm proposing to internalize the scripted plugin inside the sbt repo in a similar way `IvyPlugin` and others are.
  
My goal with this is to make it easier for devs to define sbt plugins. Those builds that already specify the scripted library dependency will use the specified plugin instead of the "internalized" one.
  
I cannot think of any reason why this hasn't been done before, but I believe it's a good moment to do it.

eed3si9n:
This is an interesting idea.
To do this we might need to define `SbtPlugin` plugin that will pull in both scripted and `sbtPlugin := true`.

dwijnand:
I'm for this idea, btw.

dwijnand:
If we make changes to scripted we should keep https://github.com/sbt/sbt/issues/3514 in mind too.

eed3si9n:
Note we'd essentially duplicate scripted plugin inside the SbtPlugin plugin, because we can't dynamically add dependencies to metabuild.

jvican:
I was thinking of porting my improvements from the scripted sbt suite to the scripted plugin (to get batch and parallel execution). If you say that this requires us to port the scripted plugin inside the `SbtPlugin` plugin, I can do both changes at once.
  
If you like this approach, let me know and I'll open a PR.

sjrd:
> Sbt plugins have only one way of being tested: scripted.

This is simply not true. I have a totally different way of testing sbt plugins, which results in much faster turnarounds, because it does not need to reload a new sbt instance for every single test. The idea is to have the build include the sources of the sbt plugin it defines, and have projects in the build itself that test the sbt plugin.

See for example how it's done in sbt-dynscalajs:
* [Source in the sbt plugin into the build](https://github.com/sjrd/sbt-dynscalajs/blob/b61fd2d3704faa29b3440729a58ae3044ed84a50/project/build.sbt#L15-L16)
[...]

Issue #3514, opened 2017-09-11
scripted should not be enabled by default


alexarchambault:
## steps

Start from a simple project:
```
$ cat project/build.properties
sbt.version=1.0.1

$ cat project/plugins.sbt
libraryDependencies += "org.scala-sbt" %% "scripted-plugin" % sbtVersion.value

$ cat build.sbt
scalaVersion := "2.11.11"
```

Run:
```
$ sbt update
…
[warn] 	Note: Unresolved dependencies path:
[warn] 		org.scala-sbt:scripted-sbt_2.11:1.0.1 ((sbt.ScriptedPlugin.projectSettings) ScriptedPlugin.scala#L39)
[warn] 		  +- default:coursier-sbt-1-0_2.11:0.1-SNAPSHOT
[error] sbt.librarymanagement.ResolveException: unresolved dependency: org.scala-sbt#scripted-sbt_2.11;1.0.1: not found
[error] 	at sbt.internal.librarymanagement.IvyActions$.resolveAndRetrieve(IvyActions.scala:331)
[error] 	at sbt.internal.librarymanagement.IvyActions$.$anonfun$updateEither$1(IvyActions.scala:205)
…
[error] (*:update) sbt.librarymanagement.ResolveException: unresolved dependency: org.scala-sbt#scripted-sbt_2.11;1.0.1: not found
[error] Total time: 7 s, completed 11 sept. 2017 19:10:37
```

## problem

updates don't work if scala version is set to 2.11.x.

## expectation

updates should work, the scripted plugin should not be enabled by default. It should require `enablePlugins(ScriptedPlugin)`

## notes

sbt version: 1.0.1

jvican:
Related https://github.com/sbt/sbt/issues/3538 :smile:.

bantonsson:
The `ScriptedPlugin` now being an _aggressive_ `AutoPlugin` is stopping us from upgrading our build to sbt 1.0.x. Is there a workaround for this? I only need the scripted plugin for 2 out of roughly 100 projects, but now it gets added everywhere.

jvican:
@bantonsson Adding `.disablePlugins(ScriptedPlugin)` to 98 of all your projects would do it! 😄

bantonsson:
@jvican That was the workaround I was afraid of 😉

pvlugter:
👍 for turning off the auto-trigger again

bantonsson:
Could we please get the old behaviour back for sbt 1.1.x?

[...]

Issue #2899, opened 2017-01-09
Settings should use dynamic dispatch semantics


dwijnand:
## steps

```scala
val baseDir = settingKey[File]("")
val srcDir = settingKey[File]("")

inThisBuild(Seq(
  baseDir := file("/"),
  srcDir := baseDir.value / "src"
))

val a = project settings (baseDir := file("/a"))
val b = project settings (baseDir := file("/b"))
```

## problem

```
> show baseDir
[info] {.}/*:baseDir
[info] 	/
[info] a/*:baseDir
[info] 	/a
[info] b/*:baseDir
[info] 	/b
[success] Total time: 0 s, completed 09-Jan-2017 22:58:41
> show srcDir
[info] {.}/*:srcDir
[info] 	/src
[info] a/*:srcDir
[info] 	/src
[info] b/*:srcDir
[info] 	/src
[success] Total time: 0 s, completed 09-Jan-2017 22:58:50
```

## expectation

In `ThisBuild` scope, `srcDir` is defined in terms of `baseDir`.
Later, in project `a` and project `b`, in project scope `baseDir` is redefined.
Intuitively, imho, `a/*:srcDir` should be `/a/src` not `/src`.

## notes

sbt.version=0.13.13

Also, using taskKeys instead of settingKeys changes nothing.

eed3si9n:
Your expectation doesn't match mine in terms of `inThisBuild(...)`, `inConfig(...)`, etc..
The RHS of the `srcDir := baseDir.value / "src"` should be scoped to the scope you said it should be scoped to. You might want some other feature request, but I disagree to categorize this as a bug.

dwijnand:
That's because you know the implementation details of `inThisBuild`.

Would you accept my expection if I used:

```scala
baseDir in ThisBuild := file("/")
 srcDir in ThisBuild := baseDir.value / "src"
```

or `in Global` or any other combination you can come up with?

What I'm trying to define is a default definition for `srcDir`, defined in terms of `baseDir`, as `baseDir / "src"`.
Given that universal, project-agnostic definition, I want to refine the value of `baseDir` and have `srcDir` respond appropriately.
That is what I believe is the intuitive expectations using sbt's scopes.

[...]

PR #3875, opened 2018-01-13
non-triggered scripted plugin: SbtPlugin

eed3si9n:
This PR takes care of a few scripted plugin related issues at one fell swoop.

1. Scripted plugin is no longer triggered. (#3514 suggested by @alexarchambault)
2. New `SbtPlugin` is added to the sbt mothership that depends on `sbt.ScriptedPlugin`. (#3538 suggested by @jvican) 
3. Some scripted default settings are now scoped globally. (#3656 suggested by @jvican)

The net effect is that a plugin author will be able to use `scripted` with

```scala
lazy val fooPlugin = (project in file("plugin"))
  .enablePlugins(SbtPlugin)

lazy val app = project
```

and it no longer requires disabling in other subprojects (like `app` above).

### notes about name conflict and bincompat

There is a name conflict between `sbt.test` package and `sbt.Keys.test`.

```scala
[error] /tmp/sbt_ea607a5c/build.sbt:5: error: reference to test is ambiguous;
[error] it is imported twice in the same scope by
[error] import _root_.sbt.Keys._
[error] and import _root_.sbt._
[error]     parallelExecution in test := false
[error]                          ^
```

Here are the workaround steps that I took.

1. sbt.test package is renamed to sbt.scriptedtest. This allows 1.0 plugins and builds to use `test` to mean `Keys.test`.
2. To keep binary compatibility for sbt 0.13 scripted, I am adding `sbt.test.ScriptedRunner` and `sbt.test.ScriptedTests` in `scripted-plugin` artifact.
3. Another affected user is Giter8 plugin that uses ScriptedPlugin. Since the intereactions are limited to `sbt.ScriptedPlugin.*`, we should be fine here. - https://github.com/foundweekends/giter8/blob/v0.11.0-M2/plugin/src/main/scala-sbt-1.0/giter8/SBTCompat.scala

### DIFF
--- a/main/src/main/scala/sbt/plugins/ScriptedPlugin.scala
+++ b/main/src/main/scala/sbt/plugins/ScriptedPlugin.scala
@@ -6,13 +6,21 @@
  */
 
 package sbt
+package plugins
 
+import java.io.File
 import Def.Initialize
 import Keys._
 import sbt.internal.util.complete.{ Parser, DefaultParsers }
 import sbt.internal.inc.classpath.ClasspathUtilities
 import sbt.internal.inc.ModuleUtilities
 import java.lang.reflect.Method
+import sbt.librarymanagement._
+import sbt.librarymanagement.syntax._
+import sbt.io._
+import sbt.io.syntax._
+import Project._
+import Def._
 
 object ScriptedPlugin extends AutoPlugin {
   override def requires = plugins.JvmPlugin
@@ -38,6 +46,7 @@ object ScriptedPlugin extends AutoPlugin {
     val scriptedDependencies = TaskKey[Unit]("scripted-dependencies")
     val scripted = InputKey[Unit]("scripted")
   }
+
   import autoImport._
   override lazy val projectSettings = Seq(
     ivyConfigurations ++= Seq(ScriptedConf, ScriptedLaunchConf),
@@ -66,7 +75,7 @@ object ScriptedPlugin extends AutoPlugin {
     scriptedRun := scriptedRunTask.value,
     scriptedDependencies := {
       def use[A](@deprecated("unused", "") x: A*): Unit = () // avoid unused warnings
-      val analysis = (compile in Test).value
+      val analysis = (Keys.compile in Test).value
       val pub = (publishLocal).value
       use(analysis, pub)
     },
@@ -143,7 +152,6 @@ object ScriptedPlugin extends AutoPlugin {
     //(token(Space) ~> matched(testID)).*
     (token(Space) ~> (PagedIds | testIdAsGroup)).* map (_.flatten)
   }
-
   def scriptedTask: Initialize[InputTask[Unit]] = Def.inputTask {
     val args = scriptedParser(sbtTestDirectory.value).parsed
     scriptedDependencies.value
@@ -164,6 +172,6 @@ object ScriptedPlugin extends AutoPlugin {
   }
 
   private[this] def getJars(config: Configuration): Initialize[Task[PathFinder]] = Def.task {
-    PathFinder(Classpaths.managedJars(config, classpathTypes.value, update.value).map(_.data))
+    PathFinder(Classpaths.managedJars(config, classpathTypes.value, Keys.update.value).map(_.data))
   }
 }
--- a/main/src/main/scala/sbt/plugins/ScriptedPlugin.scala
+++ b/main/src/main/scala/sbt/plugins/ScriptedPlugin.scala
@@ -48,6 +48,12 @@ object ScriptedPlugin extends AutoPlugin {
   }
 
   import autoImport._
+
+  override lazy val globalSettings = Seq(
+    scriptedBufferLog := true,
+    scriptedLaunchOpts := Seq(),
+  )
+
   override lazy val projectSettings = Seq(
     ivyConfigurations ++= Seq(ScriptedConf, ScriptedLaunchConf),
     scriptedSbt := (sbtVersion in pluginCrossBuild).value,
@@ -67,7 +73,6 @@ object ScriptedPlugin extends AutoPlugin {
       case Some((x, y)) => sys error s"Unknown sbt version ${scriptedSbt.value} ($x.$y)"
       case None         => sys error s"Unknown sbt version ${scriptedSbt.value}"
     }),
-    scriptedBufferLog := true,
     scriptedClasspath := getJars(ScriptedConf).value,
     scriptedTests := scriptedTestsTask.value,
     scriptedParallelInstances := 1,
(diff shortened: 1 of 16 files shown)


eed3si9n:
I half expected this was gonna happen, but including scripted library is causing major name conflict problems because it uses `sbt.test` package.

`sbt.test` collides with `sbt.Keys.test`, which is very commonly used in plugins and `build.sbt` that contains `import sbt._`. The workaround would be to move scripted into `sbt.scriptedtest` or something, but that would of break the bincompat with 0.13.16's scripted - https://github.com/sbt/sbt/blob/v0.13.16/scripted/plugin/src/main/scala/sbt/ScriptedPlugin.scala#L34

```scala
      ModuleUtilities.getObject("sbt.test.ScriptedTests", loader)
```

jvican:
Can you not add `import sbt.Keys._` manually for every `build.sbt` evaluation? That should make the right `test` have preference over the `sbt.test` package.

eed3si9n:
@jvican `Keys._` already is imported. The error message on Travis looks like this:

```scala
[error] /tmp/sbt_ea607a5c/build.sbt:5: error: reference to test is ambiguous;
[error] it is imported twice in the same scope by
[error] import _root_.sbt.Keys._
[error] and import _root_.sbt._
[error]     parallelExecution in test := false
[error]                          ^
```

eed3si9n:
I wonder if I can use the dark arts technique of using package protected to maintain bincompat.
https://twitter.com/eed3si9n/status/835750967949697024

main/src/main/scala/sbt/plugins/ScriptedPlugin.scala
```diff
@@ -43,9 +43,15 @@ object ScriptedPlugin extends AutoPlugin {
   }
 
   import autoImport._
+
+  override lazy val globalSettings = Seq(
+    scriptedBufferLog := true,
+    scriptedSbt := (sbtVersion in pluginCrossBuild).value,
```
dwijnand: This is a breaking semantic change because most users don't set `sbtVersion in pluginCrossBuild` in Global scope, they set it in some project scope (like the root project scope).`

(If you think this is wrong and should be fixed, please vote for #2899)

scripted/plugin/src/main/scala/sbt/ScriptedPlugin.scala
```diff
@@ -0,0 +1,10 @@
+/*
+ * sbt
+ * Copyright 2011 - 2017, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under BSD-3-Clause license (see LICENSE)
+ */
+
+package sbt
+
+// ScriptedPlugin has moved to main.
```
dwijnand: Why does this file exist?

dwijnand (changes requested):
LGTM. 1 concern

main/src/main/scala/sbt/plugins/ScriptedPlugin.scala
eed3si9n: See https://github.com/sbt/sbt/blob/v1.1.0/main/src/main/scala/sbt/PluginCross.scala#L46-L47:

```scala
        state.log.info(s"Setting `sbtVersion in pluginCrossBuild` to $version")
        val add = List(sbtVersion in GlobalScope in pluginCrossBuild :== version) ++
```

scripted/plugin/src/main/scala/sbt/ScriptedPlugin.scala
eed3si9n: This is sort of a memo to ourselves.
ScriptedPlugin itself has moved to the mothership, but I am leaving the module as is for backward compatibility of the build. In general having a subproject for the plugin, but _not_ having the plugin defined in it is highly unusual, so I'm leaving the trail blaze here.

main/src/main/scala/sbt/plugins/ScriptedPlugin.scala
dwijnand: There are 78 other users of `sbtVersion in pluginCrossBuild` (on GitHub) that probably aren't doing that though...

https://github.com/search?l=Scala&q=%22sbtVersion+in+pluginCrossBuild%22&type=Code&utf8=%E2%9C%93

scripted/plugin/src/main/scala/sbt/ScriptedPlugin.scala
dwijnand: I see what you mean.

main/src/main/scala/sbt/plugins/ScriptedPlugin.scala
eed3si9n: The majority of the users I think are just reading the value, so they should be ok, but I am happy to move `scriptedSbt` back into project scoping.

main/src/main/scala/sbt/plugins/ScriptedPlugin.scala
```diff
@@ -43,6 +43,12 @@ object ScriptedPlugin extends AutoPlugin {
...
   import autoImport._
+
+  override lazy val globalSettings = Seq(
+    scriptedBufferLog := true,
+    scriptedLaunchOpts := Seq(),
+  )
+
   override lazy val projectSettings = Seq(
     ivyConfigurations ++= Seq(ScriptedConf, ScriptedLaunchConf),
     scriptedSbt := (sbtVersion in pluginCrossBuild).value,
```
eed3si9n: Fixed `scriptedSbt`.

dwijnand:
I think the `project / scripted-skip-incompatible` failure is legit.

sjrd:
Although this is clearly better than what I was afraid of in #3538, I would suggest one improvement: switch the dependencies between `ScriptedPlugin` and `SbtPlugin`.

To me it's clear that any project enabling `ScriptedPlugin` also needs to be an sbt plugin, so `ScriptedPlugin`-depends-on-`SbtPlugin` makes sense. On the other hand, not all sbt plugins need to be tested with scripted, so the other direction is not necessary/desirable.

To further argue my point: given your current definition, can you imagine a scenario where a project would `enablePlugins(ScriptedPlugin)` but *not* `SbtPlugin`? Personally I can't.

eed3si9n:
It's currently not coded using `requires` because of cross building, but Giter8Plugin depends on ScriptedPlugin to reuse launch-sbt-for-testing behavior:
https://github.com/foundweekends/giter8/blob/c65f9f41ed18c8fc757c352a02c298059f8bb77d/plugin/src/main/scala/Giter8Plugin.scala
https://github.com/foundweekends/giter8/blob/c65f9f41ed18c8fc757c352a02c298059f8bb77d/plugin/src/main/scala-sbt-1.0/giter8/SBTCompat.scala

So we have a Venn diagram of large overlap between scripted and sbt plugins, with Giter8 and Scala.JS (and possibly others of course) as some holdouts.

sbt/src/sbt-test/project/scripted-skip-incompatible/build.sbt
```diff
@@ -0,0 +1,2 @@
+lazy val root = (project in file("."))
+  .enablePlugins(SbtPlugin)
```
eed3si9n, line 2: @dwijnand The Travis failure was legit. I had to migrate to the new way that explicitly enables the use of scripted at subproject definition.

sjrd:
OK. I do find this odd, but since you have at least one use case, that's good enough for me.

dwijnand (approved):

dwijnand:
Looks like AppVeyor won't pass. @eed3si9n want to look into that?

eed3si9n:
Yea. I'll ignore AppVeyor for now and merge, and I'll take a look at it.

PR #3876, opened 2018-01-13
Fix tab completion running `sbt console`

### LINKED ISSUE #3841: No tab completion when running `sbt console` in one go

## steps

Run `sbt console` for a project. (Note that I use [sbt-extras](https://github.com/paulp/sbt-extras) and do not have sbt installed, so it may be related to that?). Type `Indexe` and press <kbd>Tab</kbd> key.

## problem

A tab character is inserted instead of invoking the tab-completion.

## expectation

It should complete to `IndexedSeq` instead of adding a tab character.

## notes

The problem does not occur if I launch `sbt` first, and then enter into the `console`.

sbt version: 1.0.4
terminal/OS: Bash on Debian Stretch
Java: openjdk version "1.8.0_151"

eed3si9n:
Fixes #3841

This fixes console task that internally uses JLine. When `console` is started from batch mode, the tab is printed as is. This is because JLine is not initialized yet.
Calling `usingTerminal` initializes and restores the terminal afterwards.


dwijnand (changes requested):
Nice.

Notes?

eed3si9n:
Added notes in ca712ac.

dwijnand (approved):

PR #3897, opened 2018-01-23
Misc tweaks and cleanups

dwijnand:


internal/util-collection/src/main/scala/sbt/internal/util/KList.scala
```diff
@@ -10,7 +10,7 @@ package sbt.internal.util
 import Types._
 import Classes.Applicative
 
-/** Heterogeneous list with each element having type M[T] for some type T.*/
+/** A higher-kinded heterogeneous list of elements that share the same type constructor `M[_]`. */
```
eed3si9n, line 13: 👍

main/src/main/scala/sbt/internal/CommandExchange.scala
```diff
@@ -97,22 +92,11 @@ private[sbt] final class CommandExchange {
...
-    }
-    lazy val auth: Set[ServerAuthentication] = (s get serverAuthentication) match {
-      case Some(xs) => xs
-      case None     => Set(ServerAuthentication.Token)
-    }
-    lazy val connectionType = (s get serverConnectionType) match {
-      case Some(x) => x
-      case None    => ConnectionType.Tcp
-    }
+    lazy val port = s get serverPort getOrElse 5001
```
eed3si9n: Subjective: The chained infix I think is harder to read compared to `s.get.serverPort.getOrElse(5001)`.

eed3si9n (changes requested):
Minor comments. Overall LGTM

PR #3920, opened 2018-02-05
Run scripted without sbt/launcher

dwijnand:
This runs the majority of sbt/sbt scripted tests without using sbt/launcher.

Still TODO:
* investigate the effects of changing the dependency between sbt and the scripted plugin

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -261,12 +261,27 @@ lazy val runProj = (project in file("run"))
   )
   .configure(addSbtIO, addSbtUtilLogging, addSbtCompilerClasspath)
 
+val sbtProjDepsCompileScopeFilter =
+  ScopeFilter(inDependencies(LocalProject("sbtProj"), includeRoot = false), inConfigurations(Compile))
+
 lazy val scriptedSbtProj = (project in scriptedPath / "sbt")
   .dependsOn(commandProj)
   .settings(
     baseSettings,
     name := "Scripted sbt",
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -262,7 +262,7 @@ lazy val runProj = (project in file("run"))
   .configure(addSbtIO, addSbtUtilLogging, addSbtCompilerClasspath)
 
 lazy val scriptedSbtProj = (project in scriptedPath / "sbt")
-  .dependsOn(commandProj)
+  .dependsOn(commandProj, sbtProj % "compile->test")
```
eed3si9n: We should not use Test configuration since it's a mess to publish them to Maven Central.

eed3si9n (changes requested):

SethTisue:
I just want to say this is cool and I'm glad it's happening.

dwijnand:
Given this is forking a new JVM to call RunFromSourceMain the compile-time dependency isn't totally necessary. I've re-written the setup to get the classpath required.

Still TODO:
* investigate the effects of changing this scripted behaviour WRT usage of scripted outside of sbt/sbt

dwijnand:
Also TODO:
* Evaluate the performance change of this change (i.e evaluate if it's even worth it.. ;.;)

dwijnand:
@eed3si9n this is green now and re-ready for review/evaluation.

eed3si9n (approved):

PR #3922, opened 2018-02-06
Implement a suppressServer setting for high-security builds

swaldman:
### Improvements

This pull request implements a Boolean setting called `suppressServer`, whose default value is `false'.

If a build or plugin explicitly sets it to `true`, the sbt-1.x server will not start up
(exactly as if `sbt.server.autostart` were set to start).

Users may manually override this setting by running the `startServer` command at the interactive prompt.

### Motivation

Projects often encounter private information, such as deployment credentials, private keys, etc.
For such projects, it may be preferable to reduce the potential attack surface than to enjoy the
interoperability offered by sbt's server. Projects that wish to make this tradeoff can set `suppressServer`
to `true` in their build. Security-sensitive plugins can define this setting as well, modifying the
default behavior in favor of security.

(My own motivation is that I am working on a [plugin for developing Ethereum applications](https://github.com/swaldman/sbt-ethereum)
with scala and sbt. It must work with extremely sensitive private keys.)

---

See also a [recent conversation on Stack Exchange](https://stackoverflow.com/questions/48591179/can-one-disable-the-sbt-1-x-server/48593906#48593906).


lightbend-cla-validator:
Hi @swaldman,

main-command/src/main/scala/sbt/BasicKeys.scala
```diff
@@ -39,6 +39,12 @@ object BasicKeys {
                                  "The wire protocol for the server command.",
                                  10000)
 
+  val suppressServer =
```
eed3si9n: To avoid double negatives, I would say we should keep the same name `autoStartServer`.

eed3si9n (changes requested):
Thanks for the contribution!
Posted my comment.

main-command/src/main/scala/sbt/BasicKeys.scala
swaldman: Sounds good! I've reworked it from `false`-defaulting `suppressServer` to `true`-defaulting `autoStartServer`.

dwijnand (approved):
LGTM

eed3si9n (approved):
Thanks!

swaldman:
@eed3si9n Thanks, really a lot, for merging and releasing this so quickly. I was going to put in warning banners and stuff, which was going to be ugly and a pain. Now I just require sbt 1.1.1 and set `autoStartServer` to false by default. So much nicer.

eed3si9n:
This PR just happened to land on the tail end of 1.1.1. Normally it's a bit more of waiting :)

fommil:
I'd just like to say, thankyou @swaldman . This was driving me crazy.

swaldman:
@fommil thanks! i'm delighted it wasn't just me who will find this useful.

hrj:
This is a little over my head, so please bear with me.

How do I set `autoStartServer` to `false` globally? If I create a file at `~/.sbt/1.0/global.sbt` will it affect 1.0.x sbt versions too and fail? Do I need to add the setting in an `if sbt version >= 1.1` condition? Or can I add a file at `~/.sbt/1.1/global.sbt`?

Thanks!

eed3si9n:
@hrj You can write:

```scala
SettingKey[Boolean]("autoStartServer", "") := false
```

PR #3932, opened 2018-02-08
Introduce SBT_GLOBAL_SERVER_DIR env var to workaround long socket file paths on UNIX

### LINKED ISSUE #3930: server may fail to start if project's home has a long path
### steps
Check out a project in a directory with a path name longer than `UNIX_PATH_MAX`, which is normally just 108 characters (the size of the path member in `struct sockaddr_un`)

### problem
```
[error] server failed to start on local:///localhome/jenkinssbt/workspace/sbt-va
lidator/dbuild-0.9.9/bin/polenta/.dbuild/sbt-base/server/945224f82506eca65706/so
ck. java.io.IOException: Cannot fit name [/localhome/jenkinssbt/workspace/sbt-va
lidator/dbuild-0.9.9/bin/polenta/.dbuild/sbt-base/server/945224f82506eca65706/so
ck] in maximum unix domain socket length

```

### expectation
sbt should operate normally

### notes
sbt version: 1.1.0
see: https://linux.die.net/man/7/unix

dwijnand:
Fixes #3930

### DIFF
--- a/main/src/main/scala/sbt/internal/CommandExchange.scala
+++ b/main/src/main/scala/sbt/internal/CommandExchange.scala
@@ -9,7 +9,6 @@ package sbt
 package internal
 
 import java.io.IOException
-import java.nio.file.Files
 import java.util.concurrent.ConcurrentLinkedQueue
 import java.util.concurrent.atomic._
 import scala.collection.mutable.ListBuffer
@@ -24,7 +23,6 @@ import BasicKeys.{
   logLevel
 }
 import java.net.Socket
-import org.scalasbt.ipcsocket.UnixDomainSocketLibrary
 import sjsonnew.JsonFormat
 import sjsonnew.shaded.scalajson.ast.unsafe._
 import scala.concurrent.Await
@@ -34,7 +32,7 @@ import sbt.io.syntax._
 import sbt.io.{ Hash, IO }
 import sbt.internal.server._
 import sbt.internal.langserver.{ LogMessageParams, MessageType }
-import sbt.internal.util.{ StringEvent, ObjectEvent, MainAppender, Util }
+import sbt.internal.util.{ StringEvent, ObjectEvent, MainAppender }
 import sbt.internal.util.codec.JValueFormats
 import sbt.protocol.{ EventMessage, ExecStatusEvent }
 import sbt.util.{ Level, Logger, LogExchange }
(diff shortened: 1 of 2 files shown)


main/src/main/scala/sbt/internal/CommandExchange.scala
```diff
@@ -140,7 +142,18 @@ private[sbt] final class CommandExchange {
...
       val h = Hash.halfHashString(IO.toURI(portfile).toString)
       val tokenfile = BuildPaths.getGlobalBase(s) / "server" / h / "token.json"
-      val socketfile = BuildPaths.getGlobalBase(s) / "server" / h / "sock"
+      val socketfile = {
+        val socketfile = BuildPaths.getGlobalBase(s) / "server" / h / "sock"
+        connectionType match {
+          case ConnectionType.Local if !Util.isWindows =>
+            val maxSocketLength = new UnixDomainSocketLibrary.SockaddrUn().sunPath.length - 1
+            if (socketfile.absolutePath.length > maxSocketLength)
+              Files.createTempFile("sbt-server", ".sock").toFile // assuming this is short enough..
```
cunei: Assuming that there might be multiple independent sbt instances running at the same time, in which case you want the `.sock` file to be unique, wouldn't it be better to just always hash the string? Something like:
```
Files.createTempFile("sbt-server-"+socketfile.getCanonicalPath.hashCode.toString, ".sock").toFile
```
or, even better:
```
import java.security.MessageDigest

val digest = MessageDigest.getInstance("SHA-256")
d.update(socketfile.getCanonicalPath.getBytes())
Files.createTempFile("sbt-server-"+messageDigest.digest().toString, ".sock").toFile
```
Please also note that `getCanonicalPath` is preferable to the absolute path, since the latter is not unique (symbolic links are not resolved.)

dwijnand: > Assuming that there might be multiple independent sbt instances running at the same time, in which case you want the .sock file to be unique, wouldn't it be better to just always hash the string?

We do want a unique file - that's what `Files.createTempFile` does:

>      * @return  the path to the newly created file that did not exist before
>      *          this method was invoked

cunei: Right, ok. Then why not just using "sbt-server" as a base, regardless of the dir path?...

dwijnand: Isn't that what I'm doing? Using just "sbt-server" as a base regardless of the dir path?

```scala
Files.createTempFile("sbt-server", ".sock").toFile
```

dwijnand: Do you mean get rid of the ".sock" suffix and use `null` instead?

cunei: No, I mean: I see that in the revised code you are using the dir name as a base for the socket file, unless the path is too long, in which case you use "sbt-server". It there any specific advantage in using the path name, instead of always using just the "sbt-server" string in all cases? It's not very important, just curious.

dwijnand: I'm not sure what the original design decisions were for having the socket file in the global sbt directory. I'm just leaving that code path as is. Perhaps @eed3si9n can enlighten us.

eed3si9n: There are two requirements
1. sock file should be somewhere the user can secure using chown and chmod. sbt global directory is a good candidate for that.
2. You can't have two server instances running on a build. That's why `h` is calculated using the hash of portfile path.

eed3si9n (changes requested):
See the requirements I listed in the comment.

main/src/main/scala/sbt/internal/CommandExchange.scala
dwijnand: So what do you recommend if the sbt global directory's path is too long to use as the path of the socket?

eed3si9n: Maybe introduce a environment variable the user can use to specify "server" directory, and fail to start the server with a message saying either change your global directory or set the env variable?

dwijnand: ok

eed3si9n (approved):

main/src/main/scala/sbt/internal/CommandExchange.scala
vegabondx: Is there any reference on how to set this variable I believe somewhat in lines of `-Dsbt.global.server.directory'  ? I didnt find any reference on this except the release notes and this PR.

eed3si9n: Here's the first article I found on how to set an environment variable, if that's the question - https://www.twilio.com/en-us/blog/how-to-set-environment-variables-html

Issue #3773, opened 2017-11-27
Eviction spam


japgolly:
(See the guidelines for contributing, linked above)

## steps

In a multi-module project with a significant number of modules and not 100% up-to-date dependencies, (and/or with Scala.JS) run:

```
;clean ;test:compile
```

## problem

Entire screens of spam about eviction notices. Like this:
```
[warn] Found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[warn] 	* org.postgresql:postgresql:42.1.4 is selected over 9.4.1211
[warn] 	    +- org.tpolecat:doobie-postgres_2.12:0.4.1            (depends on 9.4.1211)
[warn] 	    +- com.blah.blah-db_2.12:047c51d9c07796d53c6c7cae4301f550f1ac66a1-dev-SNAPSHOT (depends on 9.4.1211)
[warn] 	    +- com.blah.blah:webapp-macro_2.12:047c51d9c07796d53c6c7cae4301f550f1ac66a1-dev-SNAPSHOT (depends on 9.4.1211)
[warn] 	* com.google.auth:google-auth-library-credentials:0.7.1 is selected over 0.4.0
[warn] 	    +- com.github.japgolly.fork.google-cloud-trace.v1:grpc-consumer:0.3.3 (depends on 0.7.1)
[warn] 	    +- com.google.auth:google-auth-library-oauth2-http:0.7.1 (depends on 0.7.1)
[warn] 	    +- io.grpc:grpc-auth:1.5.0                            (depends on 0.4.0)
[warn] 	* com.google.code.findbugs:jsr305:3.0.0 is selected over 1.3.9
[warn] 	    +- com.google.instrumentation:instrumentation-api:0.4.3 (depends on 3.0.0)
[warn] 	    +- io.grpc:grpc-core:1.5.0                            (depends on 3.0.0)
[warn] 	    +- com.google.http-client:google-http-client:1.19.0   (depends on 1.3.9)
```

"some are suspected to be binary incompatible" ← yeah well SBT, you're wrong in nearly all cases. Most libraries don't follow semver religiously if at all. Plus incompatibility depends on the API portion in use.

Since 0.13.16 (and to be clear: in 1.0.4),
* It prints every time you clean-compile
* It prints a lot

## expectation

* 0.13.15 had this behaviour right: when you first start SBT it shows you a small summary of eviction notices then it leaves you alone. I could clean-compile and still see what was going on in my console, and be able to see the compilation output.
[...]

PR #3947, opened 2018-02-15
Add the eviction warnings options to global

exoego:
Partially adresses https://github.com/sbt/sbt/issues/3773

It adds the eviction warning options to global, so that one can change the options for all sub projects at a time. 

Another PR (adding `EvictionWarningsOptions.summary`) for the issue in librarymanagement is https://github.com/sbt/librarymanagement/pull/211

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1999,7 +1999,7 @@ object Classpaths {
       ConflictWarning(conflictWarning.value, report, log)
       report
     },
-    evictionWarningOptions in update := (evictionWarningOptions in GlobalScope).value,
+    evictionWarningOptions in update := evictionWarningOptions.value,
     evictionWarningOptions in evicted := EvictionWarningOptions.full,
     evicted := {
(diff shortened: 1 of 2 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1999,6 +1999,7 @@ object Classpaths {
       ConflictWarning(conflictWarning.value, report, log)
       report
     },
+    evictionWarningOptions in update := (evictionWarningOptions in GlobalScope).value,
```
eed3si9n: RHS should be `evictionWarningOptions.value,`

eed3si9n (changes requested):

main/src/main/scala/sbt/Defaults.scala
exoego: Done in https://github.com/sbt/sbt/pull/3947/commits/dc4f93a7c3ecf2e2491aceaaecdbcec8fc8f7df9

eed3si9n (approved):
LGTM pending Travis

PR #3952, opened 2018-02-19
Speedup Parsers.validID

jrudolph:
It turned up in profiling sessions. Previously, it used parser combinators
which are somewhat slow especially when the JVM is still cold.

The grammar for ID is simple enough to afford this handwritten parser.


dwijnand:
Could we write a property-based that tests that this new impl == `matches(ID, s)`?

dwijnand (approved):
Thanks @jrudolph.

jrudolph:
Nice, thanks for adding the test @dwijnand. I wonder if it ever finds an example that would actually return true for either sides of the equality, though...

dwijnand:
@jrudolph good intuition! you were right.. expanded the tests and fixed a bug in your re-impl: https://github.com/sbt/sbt/pull/3955.

PR #3955, opened 2018-02-19
Fix validID & expand tests

dwijnand:
Make sure that we generate valid ID (according to matches(ID, s)) so
that we properly test the new validID implementation.

And that's what led to the bug fix. :)


jrudolph:
Good catch!

PR #3965, opened 2018-02-22
add scalacOptions for scaladoc

xuwei-k:


build.sbt
```diff
@@ -57,6 +57,14 @@ def commonSettings: Seq[Setting[_]] =
     testOptions in Test += Tests.Argument(TestFrameworks.ScalaCheck, "-w", "1"),
     testOptions in Test += Tests.Argument(TestFrameworks.ScalaCheck, "-verbosity", "2"),
     javacOptions in compile ++= Seq("-Xlint", "-Xlint:-serial"),
+    scalacOptions in (Compile, doc) ++= {
+      Seq(
+        "-sourcepath",
+        (baseDirectory in LocalRootProject).value.getAbsolutePath,
+        "-doc-source-url",
+        s"""https://github.com/sbt/sbt/tree/${sys.process.Process("git rev-parse HEAD").lineStream_!.head}€{FILE_PATH}.scala"""
```
dwijnand, line 65: So I found a way to make this prefer the git tag, if HEAD is on the tag:

```scala
{
  import scala.sys.process._
  val devnull = ProcessLogger(_ => ())
  val tagOrSha = ("git describe --exact-match" #|| "git rev-parse HEAD").lineStream(devnull).head
  s"https://github.com/sbt/sbt/tree/$tagOrSha€{FILE_PATH}.scala"
}
```

I also switched to `lineStream` so we don't silently output wrong source urls if the process fails.

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/4567

PR #3970, opened 2018-02-26
Use Java's redirectInput rather than sys.process's connectInput

### LINKED ISSUE #3737: Forked runs report noisy "Stream closed" exception with stack trace when process is stopped
During/after shutting down a forked process (started with runMain), often (most of the times?) an exception like this is shown:

```
Exception in thread "Thread-221" java.io.IOException: Stream closed
	at java.lang.ProcessBuilder$NullOutputStream.write(ProcessBuilder.java:433)
	at java.io.OutputStream.write(OutputStream.java:116)
	at java.io.BufferedOutputStream.flushBuffer(BufferedOutputStream.java:82)
	at java.io.BufferedOutputStream.flush(BufferedOutputStream.java:140)
	at java.io.FilterOutputStream.close(FilterOutputStream.java:158)
	at scala.sys.process.BasicIO$.$anonfun$input$1(BasicIO.scala:200)
	at scala.sys.process.BasicIO$.$anonfun$input$1$adapted(BasicIO.scala:198)
	at scala.sys.process.ProcessBuilderImpl$Simple.$anonfun$run$2(ProcessBuilderImpl.scala:75)
	at scala.sys.process.ProcessImpl$Spawn$$anon$1.run(ProcessImpl.scala:23)
```

I suspect it is related to `connectInput := true`. It can be reproduced by running any of the test main classes in akka-http / akka-http-core sub project (while also enabling forking and connectInput).

sbt 1.0.3

retronym:
Fixes #3737

### DIFF
--- a/run/src/main/scala/sbt/Fork.scala
+++ b/run/src/main/scala/sbt/Fork.scala
@@ -8,9 +8,13 @@
 package sbt
 
 import java.io.File
+import java.lang.ProcessBuilder.Redirect
+
 import scala.sys.process.Process
 import OutputStrategy._
 import sbt.internal.util.Util
+
+import java.lang.{ ProcessBuilder => JProcessBuilder }
 
 /**
  * Represents a command that can be forked.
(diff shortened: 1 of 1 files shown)


dwijnand (approved):
ok

eed3si9n (approved):
Thanks!

run/src/main/scala/sbt/Fork.scala
```diff
@@ -45,13 +49,19 @@ final class Fork(val commandName: String, val runnerClass: Option[String]) {
...
+    if (connectInput)
+      jpb.redirectInput(Redirect.INHERIT)
+    val process = Process(jpb)
 
     outputStrategy.getOrElse(StdoutOutput) match {
-      case StdoutOutput        => process.run(connectInput)
-      case out: BufferedOutput => out.logger.buffer { process.run(out.logger, connectInput) }
-      case out: LoggedOutput   => process.run(out.logger, connectInput)
-      case out: CustomOutput   => (process #> out.output).run(connectInput)
+      case StdoutOutput => process.run(connectInput = false)
```
retronym, line 60: I should note that my testing in the context of the Scala build was through this code path (`outputStrategy = StdoutOutput`). I _think_ the way the output is redirected should not have a bearing on how `connectInput` works, but if you can think of any other use cases to test for, I'd suggest doing some manual testing of this before the release.

dwijnand:
I think the `connectInput = false` change kills the `System.console() != null` check in `scala.util.Properties.coloredOutputEnabled` (i.e. no more colours).

PR #2951, opened 2017-02-05
Preliminary compatibility with JDK 9

retronym:
This pull request cleans up two small issues that stood in the way of using the soon-to-be-released Java 9 with SBT.

 - No more `java.ext.dirs`
 - `java.xml.bind` not part of the `java.base` default classpath

Scala compiler versions prior to 2.12.1 do not tolerate the absence of `rt.jar` from the JDK. But we can appease them a little manual setup.

First, we can extract an equivalent JAR with a [tool](https://github.com/retronym/java9-rt-export) I've written:

```
% git clone https://github.com/retronym/java9-rt-export
% cd java9-rt-export/
% java_use 1.8
% sbt package
% java_use 9
% mkdir -p $HOME/.sbt/0.13/java9-rt-ext; java -jar target/java9-rt-export-*.jar $HOME/.sbt/0.13/java9-rt-ext/rt.jar
% jar tf $HOME/.sbt/0.13/java9-rt-ext/rt.jar | grep java/lang/Object
```

Then, we can feed that to the compilation classpath of Scala compilers used within SBT with the `scala.ext.dirs` System property.

```
% java_use 9
% sbt -Dscala.ext.dirs=$HOME/.sbt/0.13/java9-rt-ext clean compile console
[info] Loading global plugins from /Users/jason/.sbt/0.13/plugins
[info] Compiling 1 Scala source to /Users/jason/.sbt/0.13/plugins/target/scala-2.10/sbt-0.13/classes...
[info] Loading project definition from /Users/jason/code/scratch1/project
[info] Set current project to scratch1 (in build file:/Users/jason/code/scratch1/)
[info] Defining *:scalaVersion
[info] The new value will be used by *:allDependencies, *:crossScalaVersions and 13 others.
[info] 	Run `last` for details.
[info] Reapplying settings...
[info] Set current project to scratch1 (in build file:/Users/jason/code/scratch1/)
[success] Total time: 0 s, completed 4 Feb. 2017, 10:37:59 pm
[info] Updating {file:/Users/jason/code/scratch1/}scratch1...
[info] Resolving jline#jline;2.12.1 ...
[info] Done updating.
[info] Compiling 1 Scala source and 1 Java source to /Users/jason/code/scratch1/target/scala-2.11/classes...
[info] 'compiler-interface' not yet compiled for Scala 2.11.8. Compiling...
[info]   Compilation completed in 7.315 s
[success] Total time: 9 s, completed 4 Feb. 2017, 10:38:08 pm
[...]

PR #3971, opened 2018-02-26
fix getDecoder

jilen:


eed3si9n:
Ref https://github.com/sbt/sbt/pull/2951
Ref https://github.com/sbt/zinc/pull/502

dwijnand (approved):

PR #3975, opened 2018-02-28
make sbt server extensible

### LINKED ISSUE #3890: Make it possible for sbt plugins to register server handlers for custom JSON-RPC methods
## problem
I have integrated with sbt server in my language server and managed to trigger compile on file save https://github.com/scalameta/metals/pull/170 I am very excited about the potential for using sbt server from an IDE to simplify the "Import project" step.

Would it be possible to give sbt plugins the ability to register handlers for custom JSON-RPC methods? Currently, sbt hardcodes a handful of LSP methods in https://github.com/sbt/sbt/blob/d1071002beabc3ec4d5a19453f3cf7f49ea2612f/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala#L54 like `initialize` and `textDocument/didSave`. I would love to be able to write an sbt plugin that adds handler to respond to custom methods. In particular, I am interested in implementing methods from https://github.com/scalacenter/bsp/blob/master/docs/bsp.md to simplify project import when the user opens a directory with an sbt build.

## expectation
Just brainstorming, maybe there could be a  `sbtServerMethods: SettingKey[List[JsonRpcMethodHandler]]` where
```
case class JsonRpcMethod[A:Encoder, B:Decoder](method: String)
case class JsonRpcMethodHandler[A, B](endpoint: JsonRpcMethod[A, B], handler: A => B)
```

## notes

sbt version: 1.1.0

eed3si9n:
Fixes #3890

Here's an example:

```scala
    Global / serverHandlers += ServerHandler({ callback =>
      import callback._
      import sjsonnew.BasicJsonProtocol._
      import sbt.internal.protocol.JsonRpcRequestMessage
      ServerIntent(
        {
          case r: JsonRpcRequestMessage if r.method == "lunar/helo" =>
            jsonRpcNotify("lunar/oleh", "")
            ()
        },
        PartialFunction.empty
      )
```

### DIFF
--- a/main-command/src/main/scala/sbt/ServerHandler.scala
+++ b/main-command/src/main/scala/sbt/ServerHandler.scala
@@ -26,8 +26,8 @@ object ServerHandler {
 
   lazy val fallback: ServerHandler = ServerHandler({ handler =>
     ServerIntent(
-      { case x => handler.log.debug(s"Unhandled notification received: ${x.method}") },
-      { case x => handler.log.debug(s"Unhandled request received: ${x.method}") }
+      { case x => handler.log.debug(s"Unhandled notification received: ${x.method}: $x") },
+      { case x => handler.log.debug(s"Unhandled request received: ${x.method}: $x") }
     )
   })
 }
--- a/main-command/src/main/scala/sbt/ServerHandler.scala
+++ b/main-command/src/main/scala/sbt/ServerHandler.scala
@@ -1,0 +1,67 @@
+/*
+ * sbt
+ * Copyright 2011 - 2017, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under BSD-3-Clause license (see LICENSE)
+ */
+
+package sbt
+
+import sjsonnew.JsonFormat
+import sbt.internal.protocol._
+import sbt.util.Logger
+import sbt.protocol.{ SettingQuery => Q }
+
+/**
+ * ServerHandler allows plugins to extend sbt server.
+ * It's a wrapper around curried function ServerCallback => JsonRpcRequestMessage => Unit.
+ */
+final class ServerHandler(val handler: ServerCallback => ServerIntent) {
+  override def toString: String = s"Serverhandler(...)"
+}
+
+object ServerHandler {
+  def apply(handler: ServerCallback => ServerIntent): ServerHandler =
+    new ServerHandler(handler)
+
+  lazy val fallback: ServerHandler = ServerHandler({ handler =>
+    ServerIntent(
+      { case x => handler.log.debug(s"Unhandled notification received: ${x.method}") },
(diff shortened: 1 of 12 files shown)


main-command/src/main/scala/sbt/ServerHandler.scala
```diff
@@ -0,0 +1,67 @@
+/*
+ * sbt
+ * Copyright 2011 - 2017, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under BSD-3-Clause license (see LICENSE)
+ */
+
+package sbt
+
```
dwijnand: could you make these `sbt.internal.server` please so we can break their API, please?

main-command/src/main/scala/sbt/ServerHandler.scala
```diff
@@ -0,0 +1,67 @@
...
+}
+
+object ServerHandler {
+  def apply(handler: ServerCallback => ServerIntent): ServerHandler =
+    new ServerHandler(handler)
+
+  lazy val fallback: ServerHandler = ServerHandler({ handler =>
+    ServerIntent(
+      { case x => handler.log.debug(s"Unhandled notification received: ${x.method}") },
+      { case x => handler.log.debug(s"Unhandled request received: ${x.method}") }
```
dwijnand: any particular reason not to print the whole object like before?

dwijnand (changes requested):
1 request, 1 question

rest LGTM

main-command/src/main/scala/sbt/ServerHandler.scala
eed3si9n: If we are exposing this out to plugins, then it needs to be public, no?
Are we saying that people can extend it, but the extension is not guaranteed to work down the line?

dwijnand: yes, it's exposed with 0 compatibility promises. caveat emptor

laughedelic: I think it's fine if it's presented as an experimental/beta functionality.

dwijnand (approved):

eed3si9n:
woohoo!

dwijnand:
sorry I only realised during the sbt meeting that you'd pushed the last commit 2 days ago.. :-/

olafurpg (commented):
Woot, looking forward to give this a spin! Thanks 😄

jvican:
This is a good first step to have bsp support in sbt, thanks for doing it. :+1:

jvican:
Hasn't this been released in an sbt version? This is blocking me from implementing a bsp prototype in sbt.

jvican:
@eed3si9n Would be really helpful if we can get this in a release soon so that I can work on the bsp prototype plugin ^^

eed3si9n:
Yea. We'll try to get some 1.2.0 milestone out soonish.

PR #3979, opened 2018-03-01
Optimize settings storage and delegation

jrudolph:
In my testing this improves minimum start up time (i.e. in steady state after sbt has cached everything from a previous run, testing with `time sbt exit && time sbt exit && time sbt exit` using the last result) for cinnamon by ~10 seconds both on 1.1.0 and 1.1.1.

Startup time 1.1.0 Vanilla: 56s
With the patch: 44s
Startup time 1.1.1 Vanilla: 63s
With the patch: 52s (together with the validId patch 45s)

This is quite a big improvement for such a self-contained change (and if you know what sbt else is doing startup...)

It also show that sbt 1.1.1 seems to have an unrelated performance regression that costs ~ 5s.

/cc @bantonsson

### DIFF
--- a/internal/util-collection/src/main/scala/sbt/internal/util/Settings.scala
+++ b/internal/util-collection/src/main/scala/sbt/internal/util/Settings.scala
@@ -8,14 +8,16 @@
 package sbt.internal.util
 
 import scala.language.existentials
-
 import Types._
 import sbt.util.Show
+
+import scala.annotation.tailrec
 
 sealed trait Settings[Scope] {
   def data: Map[Scope, AttributeMap]
   def keys(scope: Scope): Set[AttributeKey[_]]
   def scopes: Set[Scope]
+  def attributeKeys: Set[AttributeKey[_]]
   def definingScope(scope: Scope, key: AttributeKey[_]): Option[Scope]
   def allKeys[T](f: (Scope, AttributeKey[_]) => T): Seq[T]
   def get[T](scope: Scope, key: AttributeKey[T]): Option[T]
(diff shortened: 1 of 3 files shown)


internal/util-collection/src/main/scala/sbt/internal/util/Settings.scala
```diff
@@ -24,28 +26,73 @@ sealed trait Settings[Scope] {
...
-    data.flatMap { case (scope, map) => map.keys.map(k => f(scope, k)) }.toSeq
-
-  def get[T](scope: Scope, key: AttributeKey[T]): Option[T] =
-    delegates(scope).toStream.flatMap(sc => getDirect(sc, key)).headOption
+    keyDefs.toIterator.flatMap {
+      case (key: AttributeKey[Any], map) =>
+        map.keys.map(scope => f(scope, key))
+    }.toSeq
+
+  def get[T](scope: Scope, key: AttributeKey[T]): Option[T] = {
```
eed3si9n, line 60: Would it make sense to keep the original `get` code as comment here since it's a one-liner?

main-settings/src/main/scala/sbt/Scope.scala
```diff
@@ -278,20 +279,50 @@ object Scope {
...
+          case _            => withZeroAxis(scope.config)
+        }
+
+        val cLinIt = cLin.iterator
+        while (cLinIt.hasNext) {
+          val c = cLinIt.next()
+          val tLinIt = tLin.iterator
+          while (tLinIt.hasNext) {
+            val t = tLinIt.next()
+            val eLinIt = eLin.iterator
```
eed3si9n, line 314: We normally don't use extra axis, so in most cases this would contain `List(Zero)`.
Maybe it could checking for size == 1 make it faster?

main-settings/src/main/scala/sbt/Scope.scala
```diff
@@ -278,20 +279,50 @@ object Scope {
...
-        case Select(conf) => index.config(configProj, conf); case _ => withZeroAxis(scope.config)
-      }
+    def nonProjectScopes(resolvedProj: ResolvedReference)(
+        pLin: Seq[ScopeAxis[ResolvedReference]]): Seq[Scope] = {
       val tLin = scope.task match {
-        case t @ Select(_) => linearize(t)(taskInherit); case _ => withZeroAxis(scope.task)
+        case t @ Select(k) =>
+          if (taskInherit(k).isEmpty) t :: Zero :: Nil
+          else linearize(t)(taskInherit)
+        case _ => withZeroAxis(scope.task)
```
eed3si9n, line 288: `withZeroAxis` currently returns a List. Would using something else like Vector or Array make a difference?

eed3si9n (commented):
Thanks alot!
Some comments/questions.

internal/util-collection/src/main/scala/sbt/internal/util/Settings.scala
```diff
@@ -8,14 +8,16 @@
...
 import Types._
 import sbt.util.Show
 
+import scala.annotation.tailrec
+
 sealed trait Settings[Scope] {
   def data: Map[Scope, AttributeMap]
   def keys(scope: Scope): Set[AttributeKey[_]]
   def scopes: Set[Scope]
+  def attributeKeys: Set[AttributeKey[_]]
```
eed3si9n, line 20: This is causing mima failure

```scala
[error]  * abstract method attributeKeys()scala.collection.immutable.Set in interface sbt.internal.util.Settings is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("sbt.internal.util.Settings.attributeKeys")
```

Could you list this as an exception in build.sbt?

eed3si9n (changes requested):

jrudolph:
Thanks for the comments, @eed3si9n. Just FYI, I'm not sure when I'm able to continue on this, maybe end of the week.

main-settings/src/main/scala/sbt/Scope.scala
jrudolph: :+1: I guess there's a whole set of potential optimizations still left for optimizing walking the delegation chain. The old code tried to use `toStream` to compute the chain lazily. However, the overhead of `Stream` was much bigger than the gains. I haven't looked into how the delegation chains look exactly in general but I'm pretty sure there a some cleverer ways to walk it lazily.

eed3si9n:
ok. We are happy to take the commits and try to bring it to the finish line if you're busy.

dwijnand (changes requested):
I'm concerned that this causes regressions in correctness, as I'm not sure how much testing there is around the semantics of `Settings` and `indexedDelegates`.

Also there's a decrease in typesafety with all the `Any`s and `asInstanceOf`s introduced here.

Finally, without any kind of performance testing there are no guarantees that these speedups don't regress.

eatkins:
I tested this patch with the current development head (after resolving conflicts). I found the speedups to be a bit more modest in the akka project. It took about 18.75 seconds to load with the current development HEAD and 18.25 seconds with this patch applied.

eed3si9n:
I did apply one of his optimization in https://github.com/sbt/sbt/pull/4003, which shaved 0.7s. Akka can be used as a benchmark, but I think the real win is for builds with 100+ subprojects, which we do come across sometimes.

eatkins:
Makes sense.

jrudolph:
> I tested this patch with the current development head (after resolving conflicts). I found the speedups to be a bit more modest in the akka project. It took about 18.75 seconds to load with the current development HEAD and 18.25 seconds with this patch applied.

Yeah, that doesn't seem to be enough alone to warrant that change. I guess it would need some reprofiling at this point before deciding if it is still worth something or what needs to be done to improve it to get the speedups I measured back then.

eatkins:
I think it's time to either update this branch and possibly merge or close the PR.

eed3si9n:
It would be good to revisit, but closing this out for now.

PR #3982, opened 2018-03-03
WIP: Support non-echoing console for emacs

hvesalai:
Emacs `comint-mode` is having a hard time with sbt echoing back the commands to it. This causes sporadic freezing in emacs `sbt-mode` (of which I'm the author).

Scala 2.13 has already implemented emacs support for the REPL by way of an environment property called `env.emacs`. If the property has a non-empty value, then a "SimpleConsole" is used where there is no echo and no completions.

This PR tries to bring the same functionality to SBT.

~Special review should be placed on the `trait Implicit` changes as I do not understand the (macro?) magic there. The file used to refer to `internal.util.FullReader`, but there never was such an object (only a final class).~

I'm waiting for any feedback for changes that might be required to get this PR included into the next release of sbt.

### DIFF
--- a/internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
+++ b/internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
@@ -90,6 +90,7 @@ private[sbt] object JLine {
   //  older Scala, since it shaded classes but not the system property
   private[sbt] def fixTerminalProperty(): Unit = {
     val newValue = System.getProperty(TerminalProperty) match {
+      case _ if isEmacsShell                                => "false"
       case "jline.UnixTerminal"                             => "unix"
       case null if System.getProperty("sbt.cygwin") != null => "unix"
       case "jline.WindowsTerminal"                          => "windows"
@@ -161,6 +162,13 @@ private[sbt] object JLine {
 
   val HandleCONT =
     !java.lang.Boolean.getBoolean("sbt.disable.cont") && Signals.supported(Signals.CONT)
+
+  val isEmacsShell = {
+    import scala.util.Properties
+
+    (Properties.propOrElse("env.emacs", "") != "") || // same as scala.tools.nsc.Properties.isEmacsShell
+    Properties.envOrNone("INSIDE_EMACS").isDefined // preferred by emacs
+  }
 }
 
 private[sbt] class InputStreamWrapper(is: InputStream, val poll: Duration)


internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
```diff
@@ -191,25 +182,75 @@ trait LineReader {
...
-    val handleCONT: Boolean = JLine.HandleCONT,
-    val injectThreadSleep: Boolean = false
-) extends JLine {
-  protected[this] val reader = {
-    val cr = JLine.createReader(historyPath, in)
-    sbt.internal.util.complete.JLineCompletion.installCustomCompletor(cr, complete)
-    cr
+object LineReader {
+  // same as scala.tools.nsc.Properties.isEmacsShell
+  val isEmacsShell = sys.props.getOrElse("env.emacs", "") != ""
```
fommil: a more reliable way nowadays to check if you're running inside emacs is to look for the environment variables `INSIDE_EMACS`

hvesalai: Yeah. I just wanted to use the same method that scala REPL uses for symmetry. Maybe we should do a PR also there.

hvesalai:
The failing test is the MIMA one. Any advice how to get binary compliance to this is welcome.

eed3si9n:
I haven't looked at the details yet, but if you want to avoid breaking bincompat, I'd suggest you add more readers instead of breaking the existing ones?

hvesalai:
No, the point of the PR is that the logic for choosing the implementation of the Reader (whether it is JLine or System.in based) is in one place. Note that both SimpleReader and FullReader are both JLine based before this PR, the only difference is completion/non completion.

The alternative is that I change both SimpleReader and FullReader to internally do that decision. If bin compat is required in this case, then I would do two different PRs. This one for the next non-bin compat release and an other one that does the alternative describe in the previous sentense for this branch.

hvesalai:
The question here should be, is there any reason for having bin compat for bugs. Since `val FullReader` in `Imports` is not of type `type FullReader`, nobody could have ever used it for anything usefull. If that is the case, then nothing should break by the PR.

eed3si9n:
The reason we keep binary compatibility is for sbt plugin ecosystem.
A quick search on GitHub shows that ppl do use this class - https://github.com/search?l=Scala&q=FullReader&type=Code&utf8=%E2%9C%93

hvesalai:
Ok, I will update this PR to be bin-compat and do a separate one for the "master" branch that has the same API (in Imports), but a better implementation.

hvesalai:
Updated this PR to be API compatible with the previous implementation. To my understanding it should also be bin compat regarding all public methods.

hvesalai:
Also documented the companio object magic.

hvesalai:
After the latest change the MIMA findings are these. I don't think any of these should affect anyone.

```
[error]  * class sbt.internal.util.SimpleReader is declared final in current version
[error]    filter with: ProblemFilters.exclude[FinalClassProblem]("sbt.internal.util.SimpleReader")
[error]  * the type hierarchy of class sbt.internal.util.FullReader is different in current version. Missing types {sbt.internal.util.JLine}
[error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("sbt.internal.util.FullReader")
[error]  * the type hierarchy of object sbt.internal.util.SimpleReader is different in current version. Missing types {sbt.internal.util.JLine,sbt.internal.util.SimpleReader}
[error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("sbt.internal.util.SimpleReader$")
```

internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
```diff
@@ -191,11 +182,50 @@ trait LineReader {
...
+
+  def full(
+      historyPath: Option[File],
+      complete: Parser[_],
+      handleCONT: Boolean = HandleCONT,
+      injectThreadSleep: Boolean = false
+  ): LineReader =
+    if (haveInteractiveConsole)
+      new FullReaderJLineImpl(historyPath, complete, handleCONT, injectThreadSleep)
+    else new NonInteractiveConsoleImpl()
```
eed3si9n: The returned LineReader is expected to be interruptible when `injectThreadSleep` is `true`.

hvesalai: Ok, I'll have a look

hvesalai: I  can re-use the same InputStreamWrapper to accomplish this. I'll make a fix commit.

hvesalai: Eugene, why do you use new `FileInputStream(FileDescriptor.in)` instead of `System.in`? Is it prooven to have some difference?

eed3si9n: I am not sure. I think I copied whatever that was before it.

eed3si9n:
Mima exceptions needs to be listed out in build.sbt.

```scala
    mimaBinaryIssueFilters ++= Seq(
      // Added private[sbt] method to capture State attributes.
      exclude[ReversedMissingMethodProblem]("sbt.internal.util.AttributeMap.setCond"),
      ...
```

Could you also add release note in https://github.com/sbt/sbt/tree/1.x/notes/1.2.0 plz?

hvesalai:
@eed3si9n rewrote the PR once more so that there should be no MIMA exceptions needed and the injectThreadSleep logic is shared.

hvesalai:
And changed the target branch to be 1.1.x

eed3si9n:
Also I just wanted to let you know about `sbt -Djline.terminal=false` flag, which might do what you want already.

eed3si9n:
My vote is to target this to 1.x (ship 1.2.0).

fommil:
We're using that in Windows. Might be worth trying it in superior OS as well.

hvesalai:
@eed3si9n uh-oh. Yes. `-Djline.terminal=false` does what I want. I really did try that before making this PR, but I must have done something wrong because now I'm getting different results. Will still investigate and change this PR to do `-Djline.terminal=false` by default for emacs

fommil:
@hvesalai https://github.com/ensime/emacs-sbt-mode/blob/3e8a75a5d48724ecd6c4171107e1b581100dfce1/sbt-mode-vars.el#L17-L22

hvesalai:
@fommil yeah that is not good... it has to be one of "false" or "none". And it is also applicable to unix as well, not exceptions needed

fommil:
something must have changed, this fixed a lot of problems I was having on Windows a few years ago. And the default must have changed too.

hvesalai:
Ok, now the PR is really trivial

eed3si9n (approved):

hvesalai:
There is some controversy arround this issue. While it works as intended and removes the unwanted console echo for emacs, it also removes ANSI colors. We will discuss with @fommil what is best for the emacs community.

fommil:
I think it would be useful to have this, but it needs to be opt out even when emacs is detected with a strategy we can deploy that means we get the same behaviour in old sbt as new sbt.

```
case _ if isEmacsShell
```

what if this was

```
case null if isEmacsShell
```

We can easily set the explicit option when starting sbt, in a way that doesn't break old versions of sbt.

It seems to me that what emacs really needs is ansi colours and no echo! And no combination of options seems to get us both.

hvesalai:
I'd say let's have this on hold for a while and try to get the fix to be implemented into JLine proper. If that happens, then we can just remove all the emacs handling from sbt.

fommil:
might be good trying to get it into the jline that's used by sbt rather than jline3

hvesalai:
Both jline2 and jline3... AFAIK scala is going to jline3, so I bet sbt will follow suit at some point.

dwijnand:
yeah - in sbt 2 probably

eed3si9n (changes requested):
on hold requested by OP

hvesalai:
Ok, so happy news. I got the required fixes directly to Jline 2.14.6, so this can be closed and I can open a new PR for updating the JLine used by sbt.

hvesalai:
See #4050

PR #3985, opened 2018-03-05
Fix race condition in non-forked, parallel tests.

retronym:
Non forked tests that are run in parallel groups can call into
a single instance of TestStatusReporter concurrently.

This seems to be limited to the startGroup/endGroup/testEvent
methods called in:

  https://github.com/sbt/sbt/blob/a41727fb17bb923036f90ca2ca62897def1a893e/testing/src/main/scala/sbt/TestFramework.scala#L100-L124

Which itself is called within:

  https://github.com/sbt/sbt/blob/a41727fb17bb923036f90ca2ca62897def1a893e/testing/src/main/scala/sbt/TestFramework.scala#L203-L229

Creating the `runnables` that are run in parallel (in builds so configured):

  https://github.com/sbt/sbt/blob/a6eb1260c8162bfbfcbfb8656f152854a93d33ae/main-actions/src/main/scala/sbt/Tests.scala#L222-L230

We believe this to be the cause of the hang witnessed in
the a suite of Scalacheck-framework tests in the Scala
build:

  https://github.com/scala/scala-jenkins-infra/issues/249

This commit uses a concurrent map to support concurrent
status updates.

### DIFF
--- a/testing/src/main/scala/sbt/TestStatusReporter.scala
+++ b/testing/src/main/scala/sbt/TestStatusReporter.scala
@@ -8,14 +8,16 @@
 package sbt
 
 import java.io.File
-import sbt.io.IO
 
-import scala.collection.mutable.Map
+import sbt.io.IO
 import sbt.protocol.testing.TestResult
+import java.util.concurrent.ConcurrentHashMap
+
+import scala.collection.concurrent
 
 // Assumes exclusive ownership of the file.
 private[sbt] class TestStatusReporter(f: File) extends TestsListener {
-  private lazy val succeeded = TestStatus.read(f)
+  private lazy val succeeded: concurrent.Map[String, Long] = TestStatus.read(f)
 
   def doInit = ()
   def startGroup(name: String): Unit = { succeeded remove name }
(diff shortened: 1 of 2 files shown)


testing/src/main/scala/sbt/TestStatusReporter.scala
```diff
@@ -32,13 +34,17 @@ private[sbt] class TestStatusReporter(f: File) extends TestsListener {
 
 private[sbt] object TestStatus {
   import java.util.Properties
-  def read(f: File): Map[String, Long] = {
+  def read(f: File): concurrent.Map[String, Long] = {
     import scala.collection.JavaConverters._
     val properties = new Properties
     IO.load(properties, f)
     properties.asScala map { case (k, v) => (k, v.toLong) }
```
eed3si9n: This line should be removed?

eed3si9n (changes requested):
Thanks!
A minor comment.

testing/src/main/scala/sbt/TestStatusReporter.scala
retronym: yep, push-fed

eed3si9n (approved):

dwijnand (approved):
lgtm, but you need to tell MiMa to chillax on the `private[sbt]` changes:

```
[error]  * method write(scala.collection.mutable.Map,java.lang.String,java.io.File)Unit in object sbt.TestStatus's type is different in current version, where it is (scala.collection.Map,java.lang.String,java.io.File)Unit instead of (scala.collection.mutable.Map,java.lang.String,java.io.File)Unit
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("sbt.TestStatus.write")
[error]  * method read(java.io.File)scala.collection.mutable.Map in object sbt.TestStatus has a different result type in current version, where it is scala.collection.concurrent.Map rather than scala.collection.mutable.Map
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("sbt.TestStatus.read")
```

eed3si9n:
This is now merged as https://github.com/sbt/sbt/pull/3997

Issue #3863, opened 2018-01-08
sbt/exec should send respond to request


olafurpg:
## steps
Send `sbt/exec` command following the steps described in https://developer.lightbend.com/blog/2017-11-30-sbt-1-1-0-RC1-sbt-server/?final?final 

## problem

The sbt server never responds to request number 2.
```
Content-Length: 93^M
^M
{ "jsonrpc": "2.0", "id": 2, "method": "sbt/exec", "params": { "commandLine": "compile" } }^M

```
According to the json rpc specification, every request should have a response. By not responding the request, the client will keep an open request hanging forever.

## expectation

On command success, I expected to receive some response once the command completes execution
```
{ "jsonrpc": "2.0", "id": 2, "result": null }
```
On command failure, I expected an error object with a "Server error" code. Sbt should maybe define the categories of errors it can respond with and document which codes it returns with their meanings.

## notes

sbt version: 1.1.0

BastienClement:
I'm trying to use sbt server as a way to speed up a Jenkins pipeline that need to repetitively invoke sbt commands. A python script is called instead of sbt and forwards the command over the server socket. This script need to wait until the command is done and then exit with the appropriate code to indicate either a success or a failure.

The fact that there is no response to `sbt/exec` messages is a major pain. I'm currently using the `window/logMessage` (`Processing` / `Done`) messages to find out whether the command finished and use the presence of `type: 2` (error severity) messages as an indicator of failure, otherwise I assume a success.

This is mostly working, though a bit hackish. The whole system would be a lot more robust with an appropriate response from the server. I believe that this response should at least include whether the execution was successful or failed. This information is displayed by sbt as a `[success]` / `[error]` line, even for commands received from the server.

On the other hand, the speed gain is massive. I don't know if sbt server was ever intended as a way to script the shell instead of integrating with IDEs, but this is working wonderfully!

eed3si9n:
@galedric That sounds awesome. Could you please blog/opensource your stuff?

laughedelic:
I think this was fixed in https://github.com/sbt/sbt/pull/3991

dwijnand:
I think so too.

PR #3991, opened 2018-03-06
Add an optional exitCode to ExecStatusEvent so clients can use it

dwijnand:


main/src/main/scala/sbt/MainLoop.scala
```diff
@@ -155,11 +155,28 @@ object MainLoop {
...
+            None
+        }
+    }
     StandardMain.exchange publishEventMessage ExecStatusEvent(
       "Done",
       channelName,
       exec.execId,
-      newState.remainingCommands.toVector map (_.commandLine))
+      newState.remainingCommands.toVector map (_.commandLine),
+      exitCode,
```
eed3si9n: There's also successful event too right?
This needs some rethinking.

dwijnand: you're right.. last second change. will fix tomorrow

eed3si9n:
LSP already has the concept of Response and Error, so I'd say we should handle these in a more idiomatic way.
see also https://github.com/sbt/sbt/issues/3863

eed3si9n (changes requested):
I think we should fold this into #3863

main/src/main/scala/sbt/MainLoop.scala
```diff
@@ -155,13 +155,72 @@ object MainLoop {
...
+      StandardMain.exchange publishEventMessage doneEvent
+    }
     newState
   }
 
   def logFullException(e: Throwable, log: Logger): Unit = State.logFullException(e, log)
-}
+
+  private def determineExitCode(state: State, newState: State): Option[Int] = {
+    sealed trait ExitCode
```
eed3si9n: See https://github.com/sbt/sbt/blob/1.x/protocol/src/main/scala/sbt/internal/langserver/ErrorCodes.scala

eed3si9n (changes requested):
Getting closer.

main/src/main/scala/sbt/MainLoop.scala
dwijnand: what are you requesting is changed?

eed3si9n: You're currently making up 0 or 1 based numbering that gets sent to LSP client correct?
http://www.jsonrpc.org/specification determines the number you have to use.

dwijnand: it says:

> The remainder of the space is available for application defined errors.

I'm using 1 when there is some unknown error that was handled by onFailure, that's it.

eed3si9n: VS Code uses -32800 for cancellation. Maybe it's better to extend to some large negative integers following the convention, like `-33000`.

dwijnand: why is `-33000` better than `1` for "unknown error"?

dwijnand:
jfyi, this is green now @eed3si9n

main/src/main/scala/sbt/MainLoop.scala
eed3si9n: From the point of view of the LSP client (let's say Vim or something) they need to handle various error events. I want to fit into the ecosystem harmoniously.

eed3si9n (approved):

PR #3994, opened 2018-03-06
Add more configuration axis ScopeFilter factory methods

efemelar:


eed3si9n:
Since this is a new feature, I would target 1.x.
Otherwise LGTM

eed3si9n (approved):

efemelar:
This PR originally targeted `1.1.x` because it's amendment for lost feature.

In SBT 0.13.x one could do something like 
```
Classpaths
  .interSort(projectRef, configuration, settings, buildDependencies)
   .map { case (projectRef, configurationName) =>
     ScopeFilter(inProjects(projectRef), inConfigurations(new Configuration(configurationName)))
   }
```

In SBT 1.x `new Configuration` is now illegal but not all places dealing with configurations use `ConfigRef`'s, some still use strings.

eed3si9n:
From 1.1.1's point of view, I feel like this is a feature enhancement.
Plus hopefully this is something you can workaround if you really need it, correct?

efemelar:
Makes sense. I provided some context for changes since I forgot to add it right away. 
In my particular case I found workaround - searching through all of project configurations and filtering them by name. Should do for now.

As a side note, it feels weird to have 3 ways to reference configuration - `String`, `ConfigKey`, `ConfigRef`. If I understand correctly `ConfigRef` should be only one eventually. But I found no issues related to that. @eed3si9n do you know if there are any?

eed3si9n:
> As a side note, it feels weird to have 3 ways to reference configuration - String, ConfigKey, ConfigRef. If I understand correctly ConfigRef should be only one eventually. 

Feel free to open an issue for that. We should consolidate to `ConfigRef`.

dwijnand (approved):
lgtm. thanks @fmlrt!

PR #3995, opened 2018-03-07
Filter incompatible aggregates in cross switch commands

### LINKED ISSUE #3698: Add strict aggregation from sbt-doge to SBT 1.0
I'm trying to migrate from sbt-doge to SBT 1.0, but I can't find a way to replace [strict aggregation](https://github.com/sbt/sbt-doge#strict-aggregation) (`+++`). Using it seems to be the only way to build a project for a specific version (a thing usually done in Travis builds) in typical setups of a root project aggregating all sub-projects.

I think that the default behavior of `++` should be to do strict aggregation as I find it difficult to imagine an useful cross-project aggregation project otherwise. For compatibility purposes, one way to do this would be to add a `strictAggregation` sbt boolean setting that changed the behavior of the operator.

This issue follows from https://github.com/sbt/sbt-doge/issues/18.

cc @eed3si9n @jroper @dwijnand

ruippeixotog:
Fixes #3698.

### DIFF
--- a/main/src/main/scala/sbt/Cross.scala
+++ b/main/src/main/scala/sbt/Cross.scala
@@ -216,12 +216,30 @@ object Cross {
     Command.arb(requireSession(switchParser), switchHelp)(switchCommandImpl)
 
   private def switchCommandImpl(state: State, args: Switch): State = {
-    val switchedState = switchScalaVersion(args, state)
+    val x = Project.extract(state)
+    val (switchedState, affectedRefs) = switchScalaVersion(args, state)
 
-    args.command.toList ::: switchedState
+    val strictCmd =
+      if (args.version.force) {
+        // The Scala version was forced on the whole build, run as is
+        args.command
+      } else {
+        args.command.map { rawCmd =>
+          parseCommand(rawCmd) match {
+            case Right(_) => rawCmd // A project is specified, run as is
+            case Left(cmd) =>
+              resolveAggregates(x)
+                .intersect(affectedRefs)
+                .collect { case ProjectRef(_, proj) => s"$proj/$cmd" }
+                .mkString("all ", " ", "")
+          }
+        }
+      }
+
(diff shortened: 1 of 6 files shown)


ruippeixotog:
Review by @eed3si9n

eed3si9n (approved):
LGTM

dwijnand (approved):
thanks @ruippeixotog!

alexflav23:
@eed3si9n @dwijnand Would be nice to see an intermediary release if at all possible with this enabled. It's currently preventing me from upgrading anything to SBT 1.x +, unfortunately almost every single project I deal with is a multi aggregate behemoth with complex versioning rules and binary incompatibilities.

Is there anything I could do to help with an intermediary release from the 1.2.0 milestone you have already planned this for ?

main/src/main/scala/sbt/Cross.scala
```diff
@@ -216,12 +216,30 @@ object Cross {
...
+        args.command
+      } else {
+        args.command.map { rawCmd =>
+          parseCommand(rawCmd) match {
+            case Right(_) => rawCmd // A project is specified, run as is
+            case Left(cmd) =>
+              resolveAggregates(x)
+                .intersect(affectedRefs)
+                .collect { case ProjectRef(_, proj) => s"$proj/$cmd" }
+                .mkString("all ", " ", "")
```
ches, line 234: I _think_ this might have broken `+ -v`:

```
sbt> + -v compile
[error] Expected ';'
[error] ++ -v 2.11.12! all client/Compile/compile offline/Compile/compile example-offline/Compile/compile registry/Compile/compile online/Compile/compile
```

Filed #5506.

Thanks for working on this issue back when, `++` is behaving itself. Now sbt still seems to need work on `+`, it _maybe_ does the right thing for me with the explicit `crossScalaVersions := Nil` on the root aggregating project. Seems brittle, and esoteric…

PR #4002, opened 2018-03-10
add tests around scope delegation

eed3si9n:
Ref https://github.com/sbt/sbt/pull/3979

### DIFF
--- a/main/src/test/scala/Delegates.scala
+++ b/main/src/test/scala/Delegates.scala
@@ -16,10 +16,10 @@ import Prop._
 import Gen._
 
 object Delegates extends Properties("delegates") {
-  property("generate non-empty configs") = forAll { (c: Seq[Configuration]) =>
+  property("generate non-empty configs") = forAll { (c: Vector[Configuration]) =>
     c.nonEmpty
   }
-  property("generate non-empty tasks") = forAll { (t: Seq[Taskk]) =>
+  property("generate non-empty tasks") = forAll { (t: Vector[Taskk]) =>
     t.nonEmpty
   }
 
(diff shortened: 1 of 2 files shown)


dwijnand:
could you describe what you've looked and achieved in this PR? it's a little hard to reverse engineer from the diff.

eed3si9n:
> could you describe what you've looked and achieved in this PR? it's a little hard to reverse engineer from the diff.

1. Changes in main/src/test/scala/sbt/internal/TestBuild.scala is mostly around hard-coding the datatype to Vector instead of Seq.
2. The change in main/src/test/scala/Delegates.scala is adding two property-based testing. This the code that tests `Scope.delegates(...)`. Given a scoped key, it can return the list of delegations.
  - "Project axis delegates to BuildRef then Zero" tests that if you have `foo/Compile/x` it delegates to `ThisBuild/Compile/x` etc..
  - "Config axis delegates to parent configuration" tests that if you have `foo/Test/x` it delegates to `foo/Runtime/x` etc..

eed3si9n:
Can I merge this one?

dwijnand:
do you think it fully tests the behaviour and invariants of `Scope.delegates`?

I've never spent time understanding the state of the implementation and testing of scope delegation.

eed3si9n:
I don't think it does fully yet, but it covers some aspects.

dwijnand:
ok then I'll pick up from where you left off.

dwijnand (approved):
these tests are great!

this pr should not (I should not) blockade its merging just because test coverage could further be improved!

PR #4003, opened 2018-03-11
optimize scope delegation

eed3si9n:
This is a refactoring of  @jrudolph's https://github.com/sbt/sbt/pull/3979 on top of test PR https://github.com/sbt/sbt/pull/4002

For every setting, we need to calculate the delegates, so `indexedDelegates` is one of hot methods.

| build                      |  sbt 1.1.1     |  1.1.2 | this patch    |
| -------------------------- | -------------- | --------------  | ------------- |
| Cinnamon (125070 settings) | 35.539 ± 0.474s  |  35.469 ± 0.318s    | 32.242 ± 0.990s  |
| Akka      (30398 settings) | 17.908 ± 0.101s  | 17.780 ± 1.494s   | 17.166 ± 0.726s  |

Given that 1.1.x branch doesn't have optimization changes merged in we can ignore the difference between sbt 1.1.1 and 1.1.x. Basically it's around 35.5s within a second of waffling.
With this patch, I am getting around 32.2, which is 3s speedup.

As you can see, with Akka's build with much fewer settings there's roughly 0.6s speedup, so this patch likely won't help you much until you have _a lot_ of subprojects.

<s>In any case, 7s would be the lion's share of Johannes's ~10s speedup.</s> (In earlier version, I had user CPU time confused with the total) I focused on just the delegation expansion, so this change is a lot more contained.

### DIFF
--- a/internal/util-collection/src/main/scala/sbt/internal/util/Settings.scala
+++ b/internal/util-collection/src/main/scala/sbt/internal/util/Settings.scala
@@ -35,10 +35,10 @@ private final class Settings0[Scope](
     data.flatMap { case (scope, map) => map.keys.map(k => f(scope, k)) }.toSeq
 
   def get[T](scope: Scope, key: AttributeKey[T]): Option[T] =
-    delegates(scope).toStream.flatMap(sc => getDirect(sc, key)).headOption
+    delegates(scope).flatMap(sc => getDirect(sc, key)).headOption
 
   def definingScope(scope: Scope, key: AttributeKey[_]): Option[Scope] =
-    delegates(scope).toStream.find(sc => getDirect(sc, key).isDefined)
+    delegates(scope).find(sc => getDirect(sc, key).isDefined)
 
   def getDirect[T](scope: Scope, key: AttributeKey[T]): Option[T] =
     (data get scope).flatMap(_ get key)
--- a/main-settings/src/main/scala/sbt/Scope.scala
+++ b/main-settings/src/main/scala/sbt/Scope.scala
@@ -266,20 +266,47 @@ object Scope {
   )(rawScope: Scope): Seq[Scope] = {
     val scope = Scope.replaceThis(GlobalScope)(rawScope)
 
-    def nonProjectScopes(resolvedProj: ResolvedReference)(px: ScopeAxis[ResolvedReference]) = {
-      val p = px.toOption getOrElse resolvedProj
-      val configProj = p match {
-        case pr: ProjectRef => pr; case br: BuildRef => ProjectRef(br.build, rootProject(br.build))
-      }
-      val cLin = scope.config match {
-        case Select(conf) => index.config(configProj, conf); case _ => withZeroAxis(scope.config)
-      }
+    // This is a hot method that gets called many times
+    def exandDelegateScopes(resolvedProj: ResolvedReference)(
+        pLin: Seq[ScopeAxis[ResolvedReference]]): Vector[Scope] = {
       val tLin = scope.task match {
-        case t @ Select(_) => linearize(t)(taskInherit); case _ => withZeroAxis(scope.task)
+        case t @ Select(_) => linearize(t)(taskInherit)
+        case _             => withZeroAxis(scope.task)
       }
-      val eLin = withZeroAxis(scope.extra)
-      for (c <- cLin; t <- tLin; e <- eLin) yield Scope(px, c, t, e)
+      // val eLin = withZeroAxis(scope.extra)
+      // The following while loops handroll the nested for-expression + flatMap
+      // projAxes flatMap nonProjectScopes(resolvedProj)
+      //   ...
+      //   for (c <- cLin; t <- tLin; e <- eLin) yield Scope(px, c, t, e)
+      val res = Vector.newBuilder[Scope]
+      val pIt = pLin.iterator
+      while (pIt.hasNext) {
+        val px = pIt.next()
+        val p = px.toOption getOrElse resolvedProj
+        val configProj = p match {
+          case pr: ProjectRef => pr
+          case br: BuildRef   => ProjectRef(br.build, rootProject(br.build))
+        }
+        val cLin = scope.config match {
+          case Select(conf) => index.config(configProj, conf)
+          case _            => withZeroAxis(scope.config)
+        }
+        val cLinIt = cLin.iterator
+        while (cLinIt.hasNext) {
+          val c = cLinIt.next()
+          val tLinIt = tLin.iterator
+          while (tLinIt.hasNext) {
+            val t = tLinIt.next()
+            if (scope.extra.isSelect) {
+              res += Scope(px, c, t, scope.extra)
+            }
+            res += Scope(px, c, t, Zero)
+          }
+        }
+      }
+      res.result()
     }
+
     scope.project match {
       case Zero | This => globalProjectDelegates(scope)
       case Select(proj) =>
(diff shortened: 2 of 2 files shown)


main-settings/src/main/scala/sbt/Scope.scala
```diff
@@ -266,36 +266,65 @@ object Scope {
...
-    def nonProjectScopes(resolvedProj: ResolvedReference)(px: ScopeAxis[ResolvedReference]) = {
-      val p = px.toOption getOrElse resolvedProj
-      val configProj = p match {
-        case pr: ProjectRef => pr; case br: BuildRef => ProjectRef(br.build, rootProject(br.build))
-      }
-      val cLin = scope.config match {
-        case Select(conf) => index.config(configProj, conf); case _ => withZeroAxis(scope.config)
-      }
+    // This is a hot method that gets called many times
+    def exandDelegateScopes(resolvedProj: ResolvedReference)(
```
dwijnand: expand

internal/util-collection/src/main/scala/sbt/internal/util/Settings.scala
```diff
@@ -35,10 +35,10 @@ private final class Settings0[Scope](
     data.flatMap { case (scope, map) => map.keys.map(k => f(scope, k)) }.toSeq
 
   def get[T](scope: Scope, key: AttributeKey[T]): Option[T] =
-    delegates(scope).toStream.flatMap(sc => getDirect(sc, key)).headOption
+    delegates(scope).flatMap(sc => getDirect(sc, key)).headOption
 
   def definingScope(scope: Scope, key: AttributeKey[_]): Option[Scope] =
-    delegates(scope).toStream.find(sc => getDirect(sc, key).isDefined)
+    delegates(scope).find(sc => getDirect(sc, key).isDefined)
```
dwijnand, line 41: do they help memory consumption though, by being lazy?

main-settings/src/main/scala/sbt/Scope.scala
```diff
@@ -266,36 +266,65 @@ object Scope {
...
-        case t @ Select(_) => linearize(t)(taskInherit); case _ => withZeroAxis(scope.task)
+        case t @ Select(_) => linearize(t)(taskInherit)
+        case _             => withZeroAxis(scope.task)
+      }
+      // val eLin = withZeroAxis(scope.extra)
+      // The following while loops handroll the nested for-expression + flatMap
+      // projAxes flatMap nonProjectScopes(resolvedProj)
+      //   ...
+      //   for (c <- cLin; t <- tLin; e <- eLin) yield Scope(px, c, t, e)
+      val res = Vector.newBuilder[Scope]
```
retronym, line 281: I'm not sure `Vector` is the best choice here. `ListBuffer` would be the same as used before, it might make sense to use that first, and the separately try a different `SeqBuilder`. When I was looking at speeding up this code, I tried a pre-sized `ArrayBuffer`, I think.

retronym: But a big 👍 for getting rid of the nested `flatMap`-s, every level of nesting doubles the amount of garbage created.

retronym: The ideal structure would actually be a hand-rolled `Iterator[Scope]`, but is a huge pain to refactor into that form.

eed3si9n: I went for `Vector.newBuilder` based on https://stackoverflow.com/a/14059155/3827. @jrudolph went for the presized `val res = new ArrayBuffer[Scope](pLin.size * 2 * tLin.size * eLin.size)`, but I didn't see observable difference, so I figured that no longer became the bottleneck.

internal/util-collection/src/main/scala/sbt/internal/util/Settings.scala
eed3si9n: Assuming find returns at first hit, I am not sure if this saves either time or space.

main-settings/src/main/scala/sbt/Scope.scala
eed3si9n: > The ideal structure would actually be a hand-rolled `Iterator[Scope]`, but is a huge pain to refactor into that form.

That might be a good idea for hackathon to add an iterator version.

internal/util-collection/src/main/scala/sbt/internal/util/Settings.scala
eed3si9n: If we had an iterator version, as suggested by Jason below, it would be a different story. But I'd say that would be another PR.

eed3si9n: Probably related https://docs.scala-lang.org/tutorials/FAQ/stream-view-iterator.html

jrudolph: Delegation chain shouldn't be too long (< 100 elements usually) and the objects are short-lived, so without a bteter idea like the iterator, I'd also say, the extra memory usage doesn't matter here.

For `find` the big problem was the aggregation caching because in most cases `aggregate in X` is not defined but on the end of the delegation chain in which case using `Stream` was super expensive.

main-settings/src/main/scala/sbt/Scope.scala
jrudolph: > The ideal structure would actually be a hand-rolled Iterator[Scope], but is a huge pain to refactor into that form.

:+1:

jrudolph:
Thanks for taking that one for a spin and adding more tests, @eed3si9n!

jrudolph:
Btw. on my 5 year old computer, Akka loads in 20 seconds, so there might be something else going on on your machine with the Akka build.

eed3si9n:
> Btw. on my 5 year old computer, Akka loads in 20 seconds, so there might be something else going on on your machine with the Akka build.

Interesting. Maybe global settings causing this or global plugins (I only have `addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.0")` at the moment), but that's a big difference.

eed3si9n:
@dwijnand Can we put this into 1.1.2, or is there a blocker?

dwijnand:
I too would love to have this in 1.1.2, but given the importance of scope delegation the fact that it isn't properly tested makes me wary of merging these kinds of refactorings without even a RC.

so imo our choices are either:

* complete the retrofitting of test coverage for the scope delegation (aka #4002), or
* cross our fingers and merge this into 1.x, and then make sure we *really* test 1.2.0-RC1 to make sure we haven't broken anything

I don't, at this time, have a strong preference for or against delaying sbt 1.1.2 for this change.

main-settings/src/main/scala/sbt/Scope.scala
```diff
@@ -266,36 +266,65 @@ object Scope {
...
-        projAxes flatMap nonProjectScopes(resolvedProj)
+        expandDelegateScopes(resolvedProj)(projAxes)
     }
   }
 
+  private val zeroL = List(Zero)
   def withZeroAxis[T](base: ScopeAxis[T]): Seq[ScopeAxis[T]] =
-    if (base.isSelect) base :: Zero :: Nil
-    else Zero :: Nil
+    if (base.isSelect) List(base, Zero)
```
dwijnand, line 325: btw, this is slower than the previous implementation because this one requires an Array allocation of `List.apply`'s varargs..

retronym: Even better, the `Zero :: Nil` (also `GlobalScope :: Nil`) part can be shared, as I did in https://github.com/retronym/xsbt/commit/45b170edfef60c3500cbf7104c7b503bb61cfa27

retronym:
> so imo our choices are either:

One approach that is useful when test coverage isn't great is to include a commit in the PR that includes the old and new implementations, and asserts the results are the same. You can run that commit through some real-world builds to check the tripwire never goes off.

dwijnand:
@retronym interesting idea. I'll try that out.

eed3si9n:
> Btw. on my 5 year old computer, Akka loads in 20 seconds, so there might be something else going on on your machine with the Akka build.

Finally figured this out. noob mistake on my part. Given `sbt exit  43.68s user 4.22s system 247% cpu 19.323 total` I was reading the user CPU time.

eed3si9n:
Merging this into 1.x so we can test it out during 1.2 RC cycle.

PR #4017, opened 2018-03-15
Cleanup CommandChannel

dwijnand:


eed3si9n:
```scala
[error] /home/travis/build/sbt/sbt/main-command/src/main/scala/sbt/internal/ConsoleChannel.scala:43:7: overriding method publishEvent in class CommandChannel of type [A](event: A)(implicit evidence$2: sjsonnew.JsonFormat[A])Unit;
[error]  method publishEvent cannot override final member
[error]   def publishEvent[A: JsonFormat](event: A): Unit = ()
[error]       ^
```

eed3si9n (changes requested):
Travis says this code doesn't compile.

Issue #3861, opened 2018-01-08
sbt server responds with string id for requests with number id


olafurpg:
## steps
- Connect to sbt server
- Send request with number id

```
client --> server
{"method":"sbt/setting","params":{"setting":"a/crossScalaVersions"},"id":3,"jsonrpc":"2.0"}
client <-- server
{"jsonrpc":"2.0","id":"3","result":{"value":["2.12.4"],"contentType":"scala.collection.Seq[java.lang.String]"}}
```

## problem
Observe that the id is `3` in the request and `"3"` in the response.

## expectation
I expected the response id to be `3` like in the request. The server should respond with the same json as in the request id.

## notes
A workaround is to try and parse the response as a number and see if you sent an active request with that number. Alternatively, send requests with string ids.

sbt version: 1.1.0

PR #4025, opened 2018-03-17
Fix handling id in jsonrpc model

tiqwab:
Workaround for #3861.

This pull request fixes the behavior of sbt server to respond with number or string id judging by the type of id in requests. As suggested by @eed3si9n in ScalaMatsuri hackason, I append a prefix if id in requests is number and use it later to serialize responses.

### DIFF
--- a/protocol/src/main/scala/sbt/internal/protocol/codec/JsonRpcRequestMessageFormats.scala
+++ b/protocol/src/main/scala/sbt/internal/protocol/codec/JsonRpcRequestMessageFormats.scala
@@ -24,7 +24,10 @@ trait JsonRpcRequestMessageFormats {
             val id = try {
               unbuilder.readField[String]("id")
             } catch {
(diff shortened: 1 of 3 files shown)


dwijnand (commented):
♨️!

quite the hack, but it does work around the issue!

do you want to try writing a test for this? it would be great to have something that enforces this new behaviour.

tiqwab:
I actually tried to write a test in `sbt.ServerSpec`, but had difficulty handling setup and teardown of sbt server properly for each test case...
Or is it enough in this case to add json serialization/deserialization test? If so, I'll add it.

eed3si9n:
I just sent a PR here https://github.com/tiqwab/sbt/pull/1 with a test.

eed3si9n (approved):
LGTM

dwijnand:
thank you @tiqwab!

tiqwab:
Thank you for your review and help!

Issue #4010, opened 2018-03-13
Forked tests don't report failure for frameworks with nested tasks


retronym:
## steps

Checkout https://github.com/retronym/sbt-test-listener-bug and follow instructions the the README.

## problem

 - When `fork in Test := true`, the test suite reports success even though there was a failure
 - The test listener is called more than once with an `endGroup` with the same name.

## expectation

 - Consistent behaviour between forked and non-forked tests

## notes

Real world example: https://github.com/rickynils/scalacheck/issues/185

The README in sbt-test-listener-bug and the comments in https://github.com/rickynils/scalacheck/issues/185 contain some more analysis.

sbt version: 0.13.17, 1.1.0

PR #4028, opened 2018-03-21
Use Vector for EventHandler

eatkins:
ArrayList::add is not thread safe. I ran into cases where async tests
using utests would fail even when all of the individual tests passed.
This was because multiple threads called back into the handle method of
the handler instance variable, which just delegated to eventList::add.
When this happened, one of the events would get added to the list as a
null reference, which would manifest as an NPE upstream on the master
process. After this change, my tests stopped failing.


dwijnand (approved):
lgtm! nice fix, @eatkins!

two thoughts:

1. I think we could merge this into 1.1.x too for 1.1.2
2. would it be possible to construct a test for this? (e.g with async tests in utest)

retronym:
Yep, https://github.com/sbt/sbt/pull/3997 addresses a similar problem in non-forked tests.

retronym:
But I also recently discovered #4010 in the context of Scalacheck forked tests, which has the same symptom you're reporting: a single test failure doesn't fail the entire suite. That wasn't a race condition, but a problem of losing data by clobberring a binding in a map of results, when the test framework created many nested `Task`s for fine grained parallel execution (each test methods in a single class got a parallel `Task`, but because these shared the fully-qualified-name of the enclosing class the result aggregation code only ended up using the status of the final test)

@eatkins How sure are you that this change (alone) fixes the problem you saw?

testing/agent/src/main/java/sbt/ForkMain.java
```diff
@@ -294,7 +295,7 @@ private void runTestTasks(final ExecutorService executor, final Task[] tasks, fi
 					Task[] nestedTasks;
 					final TaskDef taskDef = task.taskDef();
 					try {
-						final List<ForkEvent> eventList = new ArrayList<ForkEvent>();
```
retronym, line 297: Alternatively:

```
final Collection<ForkEvent> eventList = new java.util.concurrent.ConcurrentLinkedDeque<ForkEvent>();
```

retronym, line 297: I agree this is racy.

eatkins:
I added a test to the PR that induces the data race that I was seeing.

@retronym I think my issue was slightly different than #4010. In my case, all of my individual tests passed, but the overall suite was failing.

testing/agent/src/main/java/sbt/ForkMain.java
eatkins: I like this if for no other reason than that it makes it very explicit that we expect that the eventList may be touched by multiple threads.

eed3si9n (changes requested):
LGTM
I'd also vote to rebase this to 1.1.x.

eed3si9n:
It seems https://github.com/sbt/sbt/pull/4030 does just that.

dwijnand:
closing in favour of #4030

PR #4030, opened 2018-03-21
Use Vector for EventHandler (1.1.x)

eatkins:
ArrayList::add is not thread safe. I ran into cases where async tests
using utests would fail even when all of the individual tests passed.
This was because multiple threads called back into the handle method of
the handler instance variable, which just delegated to eventList::add.
When this happened, one of the events would get added to the list as a
null reference, which would manifest as an NPE upstream on the master
process. After this change, my tests stopped failing.


dwijnand (approved):
thank you @eatkins

PR #4032, opened 2018-03-24
improve server testing

eed3si9n:
previously I was using separate thread in a forked test to test the server, but that is not enough isolation to run multiple server tests.
This adds `RunFromSourceMain.fork(workingDirectory: File)`, which allows us to run a fresh sbt on the given working directory. Next, I've refactored the stateful client-side buffer to a class `TestServer`.


build.sbt
```diff
@@ -506,10 +505,9 @@ lazy val sbtProj = (project in file("sbt"))
     buildInfoKeys in Test := Seq[BuildInfoKey](
       // WORKAROUND https://github.com/sbt/sbt-buildinfo/issues/117
       BuildInfoKey.map((fullClasspath in Compile).taskValue) { case (ident, cp) => ident -> cp.files },
+      classDirectory in Compile,
+      classDirectory in Test,
     ),
-    connectInput in run in Test := true,
-    outputStrategy in run in Test := Some(StdoutOutput),
-    fork in Test := true,
```
dwijnand, line 512: losing these breaks `sbtOn`, and the remote debugging was also useful when using `sbtOn`.

dwijnand (changes requested):
these changes impact `sbtOn`'s behaviour

build.sbt
eed3si9n: I made `fork` a function, so I am guessing I can write sbtOn based on that?

dwijnand: sure.

just make sure: it works, it sends its output immediately to the shell (e.g the prompt text), and it setups the remote debugging port.

eed3si9n: Fixed sbtOn.

dwijnand: Thank you!

dwijnand (approved):

PR #4036, opened 2018-03-25
Fixes -error not suppressing startup logs

### LINKED ISSUE #3849: "early(error)" and "-error" do not suppress startup logs
## steps

1. Use sbt 1.1.0-RC4
2. Run `sbt -error` or `sbt "early(error)"`

## problem

```
[info] Loading project definition from ....
```

## expectation

I do not see info level logs.

## notes

Originally #840

eed3si9n:
Fixes #3849

This brings back the 0.13 logic:

```scala
  def setGlobalLogLevel(s: State, level: Level.Value): State = {
    s.globalLogging.full match {
      case a: AbstractLogger => a.setLevel(level)
      case _                 => ()
    }
    s.put(BasicKeys.explicitGlobalLogLevels, true).put(Keys.logLevel.key, level)
  }
```

### DIFF
--- a/main/src/main/scala/sbt/internal/LogManager.scala
+++ b/main/src/main/scala/sbt/internal/LogManager.scala
@@ -229,8 +229,13 @@ object LogManager {
   //     s
   //   }
 
-  def setGlobalLogLevel(s: State, level: Level.Value): State =
-    s.put(BasicKeys.explicitGlobalLogLevels, true).put(Keys.logLevel.key, level)
+  def setGlobalLogLevel(s: State, level: Level.Value): State = {
+    val s1 = s.put(BasicKeys.explicitGlobalLogLevels, true).put(Keys.logLevel.key, level)
+    val gl = s1.globalLogging
+    LogExchange.unbindLoggerAppenders(gl.full.name)
+    LogExchange.bindLoggerAppenders(gl.full.name, (gl.backed -> level) :: Nil)
+    s1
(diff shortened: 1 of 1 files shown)


main/src/main/scala/sbt/internal/LogManager.scala
```diff
@@ -229,8 +229,13 @@ object LogManager {
   //     s
   //   }
```
dwijnand, line 230: is this commented block relevant to this?

dwijnand (commented):
is this just an idea, or was it manually tested to fixed the issue?

main/src/main/scala/sbt/internal/LogManager.scala
mrveera: I think so.It is useful when you use abstract logger type but now those are managed loggers.

eed3si9n:
> is this just an idea, or was it manually tested to fixed the issue?

Manually tested and confirmed.

main/src/main/scala/sbt/internal/LogManager.scala
eed3si9n: Not particularly to this bug. The commented out code provides fallback to the project scoping when global settings are NOT available.

```scala
   // if global logging levels are not explicitly set, set them from project settings
```

dwijnand (approved):
let's go!

PR #4037, opened 2018-03-25
speed up SNAPSHOT resolution of sbt itself

### LINKED ISSUE #49: make -SNAPSHOT resolution fast
-SNAPSHOT resolution is so slow, it's hard to use freshly baked sbt.
It should just pick them up from local or local maven, and ignore the rest.

eed3si9n:
Fixes https://github.com/sbt/launcher/issues/49

Running fresh sbt often is slow actually due to metabuild trying re-resolve SNAPSHOT sbt, not the launcher. This sets `.withLatestSnapshots(false)` for metabuild.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -171,8 +171,6 @@ object Defaults extends BuildCommon {
       fullJavaHomes := CrossJava.expandJavaHomes(discoveredJavaHomes.value ++ javaHomes.value),
       testForkedParallel :== false,
       javaOptions :== Nil,
-      sbtPlugin :== false,
-      isMetaBuild :== false,
       crossPaths :== true,
       sourcePositionMappers :== Nil,
       artifactClassifier in packageSrc :== Some(SourceClassifier),
@@ -216,7 +214,11 @@ object Defaults extends BuildCommon {
       pomPostProcess :== idFun,
       pomAllRepositories :== false,
       pomIncludeRepository :== Classpaths.defaultRepositoryFilter,
-      updateOptions := UpdateOptions(),
+      updateOptions := {
+        // This speeds up SNAPSHOT resolution for sbt itself
+        if (isMetaBuild.value) UpdateOptions().withLatestSnapshots(false)
+        else UpdateOptions()
+      },
       forceUpdatePeriod :== None
     )
 
@@ -295,6 +297,8 @@ object Defaults extends BuildCommon {
       aggregate :== true,
       maxErrors :== 100,
       fork :== false,
+      sbtPlugin :== false,
+      isMetaBuild :== false,
       initialize :== {},
       templateResolverInfos :== Nil,
       forcegc :== sys.props


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -156,7 +156,7 @@ object Defaults extends BuildCommon {
       javaHome :== None,
       testForkedParallel :== false,
       javaOptions :== Nil,
-      sbtPlugin :== false,
+      sbtPlugin := false,
```
dwijnand: isn't this redundant now?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -200,7 +200,11 @@ object Defaults extends BuildCommon {
       pomPostProcess :== idFun,
       pomAllRepositories :== false,
       pomIncludeRepository :== Classpaths.defaultRepositoryFilter,
-      updateOptions := UpdateOptions(),
+      updateOptions := {
+        // This speeds up SNAPSHOT resolution for sbt itself
+        if (sbtPlugin.value) UpdateOptions().withLatestSnapshots(false)
```
dwijnand: but `sbtPlugin` isn't only true in the metabuild

dwijnand (changes requested):
this seems to overreach

main/src/main/scala/sbt/Defaults.scala
eed3si9n: I didn't want to accidentally change the semantics of `JvmPlugin` in case there's a direct use of `projectSettings`. In reality it's probably safe to remove it.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Sure, but I think it's fine even for plugin development to have this flag on.

eed3si9n:
I should note that the intention of this PR is so it's easier for contributors to test freshly baked sbt, especially for perf comparisons.

main/src/main/scala/sbt/Defaults.scala
dwijnand: would this not affect the resolution of sbt plugins in the metabuild?

main/src/main/scala/sbt/Defaults.scala
dwijnand: I'd say let's remove it then.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Yes. Since Bintray doesn't allow SNAPSHOT, the effect should be minimal there too.

dwijnand: not for non-Bintray users..

I think some other flag for "I'm in the meta-build" should be introduced and used here, and avoiding the issue entirely.

eed3si9n: I think it's good to _not_ to make the distinction because the plugin authors would have closer emulation of the metabuild.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: done

main/src/main/scala/sbt/Defaults.scala
jvican: > I think some other flag for "I'm in the meta-build" should be introduced and used here, and avoiding the issue entirely.

This would have value on its own and could be reused for other things. I second it.

> I think it's good to not to make the distinction because the plugin authors would have closer emulation of the metabuild.

I agree with this, though.
  
I guess I want both.

eed3si9n:
I think https://github.com/sbt/sbt/pull/4443 makes this no longer needed.

PR #4040, opened 2018-03-27
Dedup keys in SlashSyntaxSpec

dwijnand:


dwijnand:
the duplication here has bugged me to no end for so long. happy to have finally nailed it.

main-settings/src/test/scala/sbt/SlashSyntaxSpec.scala
```diff
@@ -34,13 +43,13 @@ object BuildDSLInstances {
...
   implicit val arbReference: Arbitrary[Reference] = Arbitrary {
     Gen.frequency(
-      1 -> arbitrary[BuildRef],     // 96
-      100 -> ThisBuild,             // 10,271
-      3 -> LocalRootProject,        // 325
-      23 -> arbitrary[ProjectRef],  // 2,283
-      3 -> ThisProject,             // 299
-      4 -> arbitrary[LocalProject], // 436
-      11 -> arbitrary[RootProject], // 1,133
+      96 -> arbitrary[BuildRef],
```
eed3si9n, line 46: Are there meaning to these numbers?

eed3si9n (approved):
overall LGTM
A minor question

main-settings/src/test/scala/sbt/SlashSyntaxSpec.scala
dwijnand: they control the distribution of the generated values.

in order to get a good sample of data I looked on GitHub search for the frequency of these within GitHub's corpus.

originally, just out of force of habit, I took the raw numbers and simplified them (keeping the original as a comment - that broke Scalafmt somehow), but later realised the simplification is not necessary at all! xD

https://github.com/rickynils/scalacheck/blob/1.13.5/doc/UserGuide.md#generators

PR #4041, opened 2018-03-27
Give Scoped a default equals/hashCode implementation

dwijnand:


main-settings/src/main/scala/sbt/Structure.scala
```diff
@@ -17,7 +17,18 @@ import sbt.Def.{ Initialize, KeyedInitialize, ScopedKey, Setting, setting }
 import std.TaskExtra.{ task => mktask, _ }
 
 /** An abstraction on top of Settings for build configuration and task definition. */
-sealed trait Scoped { def scope: Scope; val key: AttributeKey[_] }
+sealed trait Scoped {
+  def scope: Scope
+  val key: AttributeKey[_]
+
+  override def equals(that: Any) =
```
eed3si9n: Wouldn't this create cooperative equality on Scoped subtypes?
If you want equals, I think they should be implemented at the class, not sealed trait.

eed3si9n (changes requested):

main-settings/src/main/scala/sbt/Structure.scala
dwijnand: yes! thank you! I'll change it and add a test that enforces it.

dwijnand:
Travis CI has forgotten about this PR?? closing-and-reopening to wake her up.

dwijnand:
ping?

Issue #3893, opened 2018-01-18
ParseKey failures on CI


dwijnand:
I've now seen two incidence of `ParseKey` (in mainProj) failing. Once in #3865 8 days ago, now in  #3875.

---

In #3865 it failed with:

```
[info] ! Key parser test.An explicitly specified axis is always parsed to that explicit value: Falsified after 51 passed tests.
[info] > Labels of failing property: 
[info] Key string: 'q / g'
[info] Structure: Env:
[info]     Tasks:
[info]     f (delegates: qq)
[info]     q (delegates: qq)
[info]     g (delegates: qq)
[info]     ka (delegates: qq)
[info]     someNfu (delegates: qq)
[info]     qq (delegates: )
[info] Build mfh:nam#cwhvfd :
[info]     Project q
[info]       Delegates:
[info]         ProjectRef(mfh:nam#cwhvfd,jqqdsp)
[info]       Configurations:
[info]         a
[info]         wz
[info]     Project jqqdsp
[info]       Delegates:
[info]         
[info]       Configurations:
[info]         a
[info]         wz
[info] current: ProjectRef(mfh:nam#cwhvfd,q)
[info] Settings:
[info] 	ProjectRef(uri("mfh:nam#cwhvfd"), "jqqdsp") / A / <key>
[info] 	   qq,ka,f,q,g,someNfu
[info] 
[info] 	ProjectRef(uri("mfh:nam#cwhvfd"), "jqqdsp") / A / qq / <key>
[info] 	   qq,ka,f,q,g,someNfu
[info] 
[info] 	ProjectRef(uri("mfh:nam#cwhvfd"), "jqqdsp") / A / someNfu / <key>
[info] 	   qq,ka,f,q,g,someNfu
[info] 
[info] 	ProjectRef(uri("mfh:nam#cwhvfd"), "jqqdsp") / Wz / <key>
[info] 	   qq,ka,f,q,g,someNfu
[info] 
[info] 	ProjectRef(uri("mfh:nam#cwhvfd"), "jqqdsp") / Wz / f / <key>
[info] 	   qq,ka,f,q,g,someNfu
[info] 
[info] 	ProjectRef(uri("mfh:nam#cwhvfd"), "jqqdsp") / Wz / g / <key>
[info] 	   qq,ka,f,q,g,someNfu
[info] 
[info] 	ProjectRef(uri("mfh:nam#cwhvfd"), "jqqdsp") / Wz / ka / <key>
[...]

PR #4043, opened 2018-03-27
Cleanup test/ParseKey

dwijnand:


main/src/test/scala/ParseKey.scala
```diff
@@ -8,166 +8,114 @@
...
-  final val MaxKeys = 5
-  final val MaxScopedKeys = 100
-
-  implicit val gstructure = genStructure
-
-  property("An explicitly specified axis is always parsed to that explicit value") =
-    forAllNoShrink(structureDefinedKey) { (skm: StructureKeyMask) =>
-      import skm.{ structure, key, mask => mask0 }
-
+  property("An explicitly specified axis is always parsed to that explicit value") = forAll {
```
eed3si9n, line 21: Is there a reason why this was changed to `forAll` from `forAllNoShrink `?

main/src/test/scala/ParseKey.scala
```diff
@@ -8,166 +8,114 @@
...
-      val s = displayMasked(key, mask, hasZeroConfig)
-      ("Key: " + displayPedantic(key)) |:
-        parseExpected(structure, s, expected, mask)
-    }
+      // as `Zero/name`, but could be interpreted as `Zero/Zero/name`.
+      val expected = ScopedKey(
+        Resolve(structure.extra, Select(structure.current), key.key, mask)(key.scope),
+        key.key
+      )
+      parse(structure, key, mask, hasZeroConfig)(
```
eed3si9n: Not sure `parse` is a good name here, since you are no longer passing `s: String`.
How about `withReparsedScopeKey(...)(f)`?

eed3si9n (changes requested):
A few comments

main/src/test/scala/ParseKey.scala
dwijnand: I think `forAllNoShrink` is a debug tool?

do you know why `forAllNoShrink` was committed?

I'm hoping that with this we'll eventually be able to solve https://github.com/sbt/sbt/issues/3893.

main/src/test/scala/ParseKey.scala
dwijnand: hmm, how about `parseCheck`?

or maybe just `check` seeing as the whole spec is about parsing..

eed3si9n: Either is fine with me.

main/src/test/scala/ParseKey.scala
eed3si9n: Not exactly sure why the original opted for no shrink.
I know what causes #3893 by the way. It's the ambiguous task and subproject name. In the first case, you have subproject `q` and task `q`. We just need to reject that as not legal in sbt.

PR #4044, opened 2018-03-27
Cleanup Resolve

dwijnand:


main/src/main/scala/sbt/internal/Resolve.scala
```diff
@@ -11,21 +11,23 @@ package internal
...
-            key: AttributeKey[_],
-            mask: ScopeMask): Scope => Scope = {
-    val rs =
-      resolveProject(current, mask) _ ::
-        resolveExtra(mask) _ ::
-        resolveTask(mask) _ ::
-        resolveConfig(index, key, mask) _ ::
-        Nil
+  def apply(
+      index: BuildUtil[_],
```
eed3si9n, line 15: I don't know if this additional newline makes sense here.
Let Scalafmt be.

main/src/main/scala/sbt/internal/Resolve.scala
```diff
@@ -39,16 +41,16 @@ object Resolve {
     else scope.copy(extra = Zero)
 
   def resolveConfig[P](index: BuildUtil[P], key: AttributeKey[_], mask: ScopeMask)(
-      scope: Scope): Scope =
+      scope: Scope,
```
eed3si9n, line 44: Why...

eed3si9n (changes requested):
I am a bit confused with seemingly adhoc style changes in this PR. The point of having Scalafmt is so we don't have to do this.

dwijnand:
the default that Scalafmt has of aligning things on some random column instead of at tab stops makes the code illegible to me.

ideally I would change the configuration instead.

eed3si9n:
> the default that Scalafmt has of aligning things on some random column instead of at tab stops makes the code illegible to me.

Sure it's a bit odd, but I don't personally think "illegible."

> ideally I would change the configuration instead.

Let's not spend further dev time and Git history on whitespaces, at least until it really really becomes a problem. I think there are better/bigger fish we could fry.

dwijnand:
> I think there are better/bigger fish we could fry.

because you don't think it's illegible.

personally I do this again and again while tackling the codebase and often throw it away (or leave it my workspace dirty, or throw it in a unpublished branch). it's tiresome.

eed3si9n:
Let's seek a general solution to this. Wouldn't everything be solved if we increased the `maxColumn = 100` to some large number and/or set align.openParenCallSite to false?

dwijnand:
> Let's seek a general solution to this.

happy to

> Wouldn't everything be solved if we increased the maxColumn = 100 to some large number and/or set align.openParenCallSite to false?

yeah `align.openParenCallSite = false` and `align.openParenDefnSite = false` (https://github.com/dwijnand/sbt/commit/992f6251e8642abf33e860e0bf8ac0abc55880b0) definitely help.

I'll try a few ways and come back.

dwijnand:
now builds on #4064

PR #4050, opened 2018-03-28
Update to Jline 2.14.6

hvesalai:
This version of Jline fixes three things for Emacs users:
- ANSI colors are now enabled for Emacs.
- Terminal echo is now disabled for Emacs.
- History is enabled for all dump terminals.


hvesalai:
Respective PR in scala: scala/scala#6478

dwijnand (approved):

eed3si9n (approved):
LGTM pending CI

hvesalai:
Note: we still need to figure out why newer versions of sbt don't actually use the specified jline, but instead use the jline in scala's jline.jar

PR #4054, opened 2018-03-31
Fix console, JLine issue

### LINKED ISSUE #3482: 1.0.{0,1,2} Opening console multiple times breaks escaping
## steps
1. Start an sbt shell
2. Type `console<enter>`
3. Type `ctrl-d` or `:q<enter>`
4. Type `console<enter>`
5. Type an arrow key or `ctrl-n`, `ctrl-p`, `ctrl-r`, etc.

## problem
Escapes are not interpreted correctly and sbt prints out garbage characters.

## expectation
Escapes are interpreted correctly and sbt does not print out garbage characters.

## notes
* I'm on a Mac using iterm 2.
* Bug doesn't exist on sbt 0.13.16

sbt version: *1.0.0*, *1.0.1*

eed3si9n:
Fixes #3482

This fixes running `console` multiple time messing up JLine.

### DIFF
--- a/main-actions/src/main/scala/sbt/Console.scala
+++ b/main-actions/src/main/scala/sbt/Console.scala
@@ -9,8 +9,8 @@ package sbt
 
 import java.io.File
 import sbt.internal.inc.AnalyzingCompiler
+import sbt.internal.util.JLine
 import sbt.util.Logger
-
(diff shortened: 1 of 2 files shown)


dwijnand (approved):
I don't understand the history here or what explains why this fixes it. but let's go

Issue #3042, opened 2017-03-25
Support for: lazy val foo = someCompositeProject (e.g., CrossProject)


sjrd:
## Problem

When defining a cross-compiling Scala/JVM + Scala.js project, one needs to use the following pattern:
```scala
lazy val foo = crossProject.
  settings(
    ...
  )

lazy val fooJVM = foo.jvm
lazy val fooJS = foo.js
```
When we initially designed `CrossProject`, we were hoping we would be able to do something like:
```
lazy val foo @ CrossProject(fooJVM, fooJS) = ...
```
but this cannot work because of #2290.
Now, with Scala Native coming along too, together with [a new, generic `CrossProject`](https://github.com/scala-native/sbt-crossproject), that approach wouldn't even work/scale even if we could make it work.

Nevertheless, this has been a pain point of defining Scala.js cross-projects, and it's confusing people (especially when they forget the two additional vals).

## Suggested solution

To solve this issue in a way that I think should have minimal impact on sbt, I propose the following. We add the following interface to sbt:
```scala
package sbt

trait CompositeProject {
  def componentProjects: Seq[Project]
}
```
Then, we augment the mechanism in sbt that discovers the list of all projects in a build, so that it recognizes (lazy) vals of type `<: CompositeProject` in addition to those of type `Project`. The mechanism would call the `componentProjects` methods to get a list of all the projects that need to be part of the build.

**Subtlety:** If there is a `val` of type `Project` in the build whose ID matches that of a component project, the top-level `val` of type `Project` needs to shadow/overwrite the component project. This is required for backwards compatibility, and to prevent further confusion when people do the following (which we discourage, but people do it anyway):
```scala
lazy val foo = crossProject.
  ...
lazy val fooJVM = foo.jvm.settings(moreJVMSettings)
lazy val fooJS = foo.js.settings(moreJSSettings)

// instead of the recommended
lazy val foo = crossProject.
  ...
  jvmSettings(moreJVMSettings).
  jsSettings(moreJSSettings)
lazy val fooJVM = foo.jvm
lazy val fooJS = foo.js
```

## Benefits

With the above notion of `CompositeProject` added to sbt, we can enhance Scala.js' `CrossProject` as follows:
```scala
class CrossProject(...) extends CompositeProject {
  ...
  def componentProjects: Seq[Project] = Seq(jvm, js)
}
```
and we can now simply write:
```scala
lazy val foo = crossProject.
  settings(
    ...
  )
```
and have everything work as expected.

[...]

PR #4056, opened 2018-04-02
Introduce CompositeProject

ghost:
fixes: https://github.com/sbt/sbt/issues/3042

/CC the sbt-crossproject "crew" - @sjrd @densh @MasseGuillaume


lightbend-cla-validator:
Hi @BennyHill,

ghost:
Just did the cla....please shout if I found a way to mess it up

ghost:
@sjrd Thanks for feedback...tbh, I PR'd early for such feedback.

I'll add you changes :+1:

ghost:
Also, whilst I modified `BuildDef`, I'm not entirely certain that this is in fact (still) used. I'll continue to update both....

main/src/main/scala/sbt/Project.scala
```diff
@@ -119,7 +119,13 @@ sealed trait ProjectDefinition[PR <: ProjectReference] {
     if (ts.isEmpty) Nil else s"$label: $ts" :: Nil
 }
 
-sealed trait Project extends ProjectDefinition[ProjectReference] {
+trait CompositeProject {
```
dwijnand, line 124: could you seal this please?

dwijnand:
I've retargeted this to be merged into 1.x because it's a bit more than just a straight bug fix.

however that's caused merge conflicts with unrelated changes already in 1.x - sorry!

main/src/main/scala/sbt/Project.scala
ghost: In the implementation of, say, `CrossProject`, we would want to extend this. In the sbt-test, I just create an instance, but in reality, I _think_ we would want to extend. 

Of course, I may have got the use wrong. But my impression was that by doing this, its then possible to have a `CrossProject` as a „real“ `Project`

ghost: Hm... maybe I‘m wrong, sigh. OK 👍

ghost:
Heh, I know - I went to change it before the PR....but saw the changes and as I was unsure of the right branch to use, just submitted. It was 50/50, never mind, really np 😄

main/src/main/scala/sbt/Project.scala
sjrd: Er ... @BennyHill is right, here. If it's sealed, this thing is completely useless ;) The whole point of this thing is that we can indeed have `CrossProject` implement that trait (which is a pure Java interface, btw). Otherwise, how do you propose we leverage this in sbt-crossproject?

ghost: OK.... How about that when I do the other changes, I update the tests to mimick invisaged usage. That way, it^s clear-cut

dwijnand: hmm I see your point.

what if we provide a `ProjectsList` subclass?

I'm not completely against the idea of an open interface, I just prefer to not have one.

sjrd: > what if we provide a ProjectsList subclass?

What exactly does that buy you, if that one is open, compared to having the trait open? The only thing I see is that you would be able to add a concrete field in a binary compatible way, but ... that would be weird anyway.

For things that build on top of it, it's a limitation, as they cannot extend `ProjectsList` *and* another class, but they could extend `CompositeProject` and a class, because `CompositeProject` is an interface.

dwijnand: > What exactly does that buy you, if that one is open

it wouldn't be open `final class ProjectsList` that extends `CompositeProject` I was thinking.

sjrd: Then I don't see how I can use it in `CrossProject` either, since I need `crossProject` to return something that is both a `CompositeProject` (for sbt to pick it up as such) *and* have methods that are specific to `CrossProject`s (so that I can do `.dependsOn(foo.jvm)` in some other JVM project, at least. If my only options to have a `CompositeProject` are `Project` which is sealed and `ProjectsList` which is final, it's no good.

Basically, think about how I can declare `CrossProject` so that the following typechecks and is identified as a `CompositeProject`:
```scala
lazy val foo = crossProject
  .settings(...)
  .jvmSettings(...)
  .jsSettings(...)

lazy val bar = project
  .settings(...)
  .dependsOn(foo.jvm)
```

sjrd: With an open `CompositeProject`, it's super easy. I just alter the [declaration of `CrossProject`](https://github.com/portable-scala/sbt-crossproject/blob/f1a6e1c7c18f109599174a4f8cfad60e1afb9a21/sbt-crossproject/src/main/scala/sbtcrossproject/CrossProject.scala#L10) with `extends CompositeProject` and
```scala
def componentProjects: Seq[Project] = projects.values.toSeq
```

dwijnand: got it. thanks for detailing the use case.

ghost:
just fyi, one more commit to come with a mock `CrossProject extends CompositeProject` ....

ghost:
Actually, is this OK as is?

sbt/src/sbt-test/project/sbt-composite-projects/changes/shadow.sbt
```diff
@@ -0,0 +1,32 @@
...
+lazy val bar = project
+  .dependsOn(foo.jvm)
+
+val g = taskKey[Unit]("A task in the root project")
+g := println("Hello.")
+
+
+check := {
+  val verJvm = (version in foo.jvm).?.value
+  assert (verJvm == Some("0.2.0"))
```
eed3si9n, line 26: Nice

eed3si9n (approved):
overall LGTM
could you add a release note in https://github.com/sbt/sbt/tree/1.x/notes/1.2.0 plz?

main/src/main/scala/sbt/Project.scala
```diff
@@ -121,7 +121,34 @@ sealed trait ProjectDefinition[PR <: ProjectReference] {
...
+  def componentProjects: Seq[Project]
+}
+
+object CompositeProject {
```
sjrd: This companion object can probably be `private[sbt]`. Users of sbt don't need to know about `uniqueId`.

main/src/main/scala/sbt/internal/EvaluateConfigurations.scala
```diff
@@ -151,10 +151,13 @@ private[sbt] object EvaluateConfigurations {
...
       {
-        val projects =
-          definitions.values(loader).collect {
-            case p: Project => resolveBase(file.getParentFile, p)
+        val projects = {
+          val compositeProjects = definitions.values(loader).flatMap {
+            case p: CompositeProject => p.componentProjects.map(resolveBase(file.getParentFile, _))
+            case _                   => Nil
           }
+          CompositeProject.uniqueId(compositeProjects)
```
sjrd: If you first `flatMap`, *then* call `uniqueId`, you have already lost the distinction between things that were in the original list versus those added via the `componentProjects`. That can't properly work: they will *all* be kept.

main/src/main/scala/sbt/internal/BuildDef.scala
```diff
@@ -17,7 +17,10 @@ import sbt.internal.inc.ReflectUtilities
 
 trait BuildDef {
   def projectDefinitions(@deprecated("unused", "") baseDirectory: File): Seq[Project] = projects
-  def projects: Seq[Project] = ReflectUtilities.allVals[Project](this).values.toSeq
+  def projects: Seq[Project] =
+    CompositeProject.uniqueId(
+      ReflectUtilities.allVals[CompositeProject](this).values.toSeq.flatMap(_.componentProjects))
```
sjrd: Same problem that I mentioned in the other comment, obviously.

main/src/main/scala/sbt/Project.scala
```diff
@@ -121,7 +121,34 @@ sealed trait ProjectDefinition[PR <: ProjectReference] {
...
+
+  /**
+   *  If two projects with the same id appear in `compositeProjects` and
+   *  in `compositeProjects.map(_.componentProjects)`, the one in
+   *  `compositeProjects` wins.
+   *  This is necessary for backward compatibility with the idiom:
+   *    lazy val foo = crossProject
+   *    lazy val fooJVM = foo.jvm.setting
+   */
+  def uniqueId(compositeProjects: Seq[Project]): Seq[Project] = {
```
sjrd: Basically, this method should take the original `Seq[CompositeProject]`, not a `Seq[Project]`. If it takes the `Seq[Project]`, it's too late: you don't know what were the projects that were in the initial `compositeProjects` versus those that were added through `componentProjects`.

main/src/main/scala/sbt/internal/EvaluateConfigurations.scala
sjrd: It should be possible to *first* call `CompositeProject.expand(definitions.values(loader))`, then `map` the result with `.map(resolveBase(file.getParentFile, _))`.

main/src/main/scala/sbt/Project.scala
sjrd: Hence, I would recommend calling this method `expand`, not `uniqueId`, since its job would be to *expand* the list of `CompositeProject` into the appropriate list of `Project`s.

sbt/src/sbt-test/project/sbt-composite-projects/changes/shadow.sbt
```diff
@@ -0,0 +1,32 @@
...
+
+// Based on sbt-file-projects test
+lazy val foo = new CompositeProject
+{
+  val jvm = Project.apply("jvm", new File("jvm"))
+  val js = Project.apply("js", new File("js"))
+  def componentProjects: Seq[Project] = Seq(jvm, js)
+}
+
+lazy val fooJVM = foo.jvm.settings(version := "0.2.0") // this one needs to win
```
sjrd: This is very good.

To make it even better, you can also declare the `lazy val fooJS`, but *above* the `lazy val foo` (they're lazy vals, so it will work). But that should ensure that the shadowing treatment does not accidentally work *via* the ordering (which I suspect is currently the case).

sjrd (changes requested):
I have a series of comments, but basically all related to the same root cause: I think the refactoring you applied with `uniqueId` gives the wrong answer.

eed3si9n (dismissed):
Yea. @sjrd is right about the `uniqueId`.

ghost:
Thanks @sjrd for the thorough review, feedback and suggestions. I will do these tomorrow with fresh eyes, along with the release notes

sbt/src/sbt-test/project/sbt-composite-projects/changes/shadow.sbt
ghost: >  (which I suspect is currently the case).

Indeed, adding a `lazy val fooJS` first does fail...

main/src/main/scala/sbt/Project.scala
```diff
@@ -121,7 +121,44 @@ sealed trait ProjectDefinition[PR <: ProjectReference] {
...
+   *    lazy val fooJVM = foo.jvm.settings(...)
+   *  }}}
+   *  and the rarer:
+   *  {{{
+   *    lazy val fooJS = foo.js.settings(...)
+   *    lazy val foo = crossProject
+   *    lazy val fooJVM = foo.jvm.settings(...)
+   *  }}}
+   */
+  def expand(projects: Seq[Project], compositeProjects: Seq[CompositeProject]): Seq[Project] = {
```
sjrd: You can compute `projects` inside `expand` from `compositeProjects` once and for all, since all projects are also composite projects, so that call sites don't need to know about this:
```scala
val projects = compositeProjects.collect {
  case project: Project => project
}
```
No need to pass it as an additional parameter when `expand` can compute it itself, especially since the call sites are basically `collect` ing as well to give it to `expand` anyway.

sjrd (approved):
:)

main/src/main/scala/sbt/Project.scala
ghost: Yep, so I've removed it - thx :+1:

dwijnand (approved):

PR #4059, opened 2018-04-03
make Global / cancelable true by default

### LINKED ISSUE #3252: Make `cancelable in Global := true` the default?
As far as I can tell, everyone loves this feature, so why not make it the default? The only reason I can think of is that it doesn't seem to work with `compile` (the shell says "Cancelling execution...", but nothing is canceled until the compilation completed, and you're stuck). I see two solutions to that:
1. Make compilation interruptible. There is no fundamental reason that this cannot be done.
2. Make `Ctrl+D` or some other shortcut (`Ctrl+C Ctrl+C` ?) quit sbt at any point, just like `Ctrl+C` does when `cancelable` is false.

eed3si9n:
Fixes #3252

/cc @retronym

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -217,7 +217,7 @@ object Defaults extends BuildCommon {
       settingsData := buildStructure.value.data,
       trapExit :== true,
       connectInput :== false,
-      cancelable :== false,
+      cancelable :== true,
       taskCancelStrategy := { state: State =>
         if (cancelable.value) TaskCancellationStrategy.Signal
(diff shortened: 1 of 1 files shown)


dwijnand (approved):
lgtm

but @retronym brings up some good points in https://github.com/sbt/sbt/issues/3252#issuecomment-378150961

eed3si9n:
https://github.com/sbt/sbt/issues/3252#issuecomment-389731935

> This all resulted in an experience which was practically unusuable, since it made it impossible (short of kill -9) to actually stop test or run tasks in SBT on that project.

This doesn't sound good, so I am closing this for now.

PR #4064, opened 2018-04-04
Cleanup code formatting

dwijnand:
Don't align args or params by the opening parens,
and drop the closing parens to the next line.


main-actions/src/main/scala/sbt/ForkTests.scala
```diff
@@ -84,10 +89,13 @@ private[sbt] object ForkTests {
...
-                            forkFingerprint(t.fingerprint),
-                            t.explicitlySpecified,
-                            t.selectors))
+                new TaskDef(
+                  t.name,
+                  forkFingerprint(t.fingerprint),
+                  t.explicitlySpecified,
+                  t.selectors
+              )
+            )
```
eed3si9n: Close paren `)` dangling almost like close curly bracket looks odd especially for invocation, and this doesn't even line up.

main-actions/src/main/scala/sbt/Package.scala
```diff
@@ -66,8 +68,10 @@ object Package {
     setVersion(main)
 
     val cachedMakeJar = inputChanged(cacheStoreFactory make "inputs") {
-      (inChanged,
-       inputs: Map[File, String] :+: FilesInfo[ModifiedFileInfo] :+: Manifest :+: HNil) =>
+      (
+          inChanged,
+          inputs: Map[File, String] :+: FilesInfo[ModifiedFileInfo] :+: Manifest :+: HNil
+      ) =>
```
eed3si9n: This almost makes the code look alien (non-Scala).

main-actions/src/main/scala/sbt/RawCompileLike.scala
```diff
@@ -47,16 +47,20 @@ object RawCompileLike {
...
+      cacheStoreFactory: CacheStoreFactory,
+      fileInputOpts: Seq[String],
+      doCompile: Gen
+  ): Gen =
     (sources, classpath, outputDirectory, options, maxErrors, log) => {
       type Inputs =
         FilesInfo[HashFileInfo] :+: FilesInfo[ModifiedFileInfo] :+: Seq[File] :+: File :+: Seq[
-          String] :+: Int :+: HNil
+          String
+        ] :+: Int :+: HNil
```
eed3si9n: A dangling square bracket? Is this more readable, srsly?

eed3si9n (changes requested):
Not feeling the dangling.

main-actions/src/main/scala/sbt/RawCompileLike.scala
dwijnand: this one should be rewrapped differently, agreed

main-actions/src/main/scala/sbt/ForkTests.scala
dwijnand: on it not lining up: agreed - looks like a scalafmt bug

on it looking like curly brackets: that's right, it's nice and uniform.

you don't write:

```scala
  println("Hello, World!")}
```

main-actions/src/main/scala/sbt/Package.scala
dwijnand: this code looked bad before and it's not great after scalafmt either.

but at least now you know where the parens start and end because they're symmetrical at the same column.

I would fix this manually with a local type alias.

dwijnand:
I find the dangling parentheses improve the clarity of the code.

do you want me to apply those manual fixes?

main-actions/src/main/scala/sbt/ForkTests.scala
eed3si9n: And your proposal is to turn that into JSON?

```scala
def main(
    args: Array[
        String
    ]
): Unit = {
  println(
      "Hello, World!"
  )
}
```

eed3si9n:
The switch to Scalafmt from Scalariform in https://github.com/sbt/sbt/pull/3125 was motivated by contributors complaining that sbt code base was odd looking. In #3125 I tried to synthesize what's most idiomatic given the options.

Currently we have tree options when a function invocation overflow.

### option 1: align.openParenCallSite (default: true)

```scala
new TaskDef(t.name, // align by (
            forkFingerprint(t.fingerprint),
            t.explicitlySpecified)
```

### option 2: align.openParenCallSite = false

```scala
new TaskDef(
  t.name, // no align by (
  forkFingerprint(t.fingerprint),
  t.explicitlySpecified)
```

### option 3: align.openParenCallSite = false, danglingParentheses = true

```scala
new TaskDef(
  t.name, // no align by (
  forkFingerprint(t.fingerprint),
  t.explicitlySpecified
)
```

### what human would write

Suppose if `t.explicitlySpecified` goes over the 100 character limit, what a human writer would do I think is just put a single newline and indent once as follows:

```scala
new TaskDef(t.name, forkFingerprint(t.fingerprint),
  t.explicitlySpecified)
```

I'd think option 1 or option 2 to be an acceptable compromise.

main-actions/src/main/scala/sbt/ForkTests.scala
dwijnand: if the type of `Array` and the argument to `println` are long enough yes:

```scala
def main(
  args: Array[
    String :+: String :+: String :+: String :+: String :+: String :+: String
  ]
): Unit = {
  println(
    "Hello, World! This is my super long message that requires wrappping"
  )
}
```

I'd say that's clearer than:

```scala
def main(
  args: Array[
    String :+: String :+: String :+: String :+: String :+: String :+: String]
): Unit = {
  println(
    "Hello, World! This is my super long message that requires wrappping")
}
```

dwijnand:
> what a human writer would do I think

it depends how much that human cared about the legibility of the code.

personally I would go back and introduce newlines and drop the closing paren so it was more clear with the open and close parens and the one line per argument that there are 3 arguments.

by using the formatting option `danglingParentheses = true` we get both benefits: the code is clear and you don't have to do the work.

eed3si9n:
> it depends how much that human cared about the legibility of the code.

Many people have cared about the legibility throughout the history of the mankind :) There are no major style guide that's proposed saying that we should format line wrapped function invocation with closing parenthesis on its own line. In most style guides, it's considered a bad style.

### Scala

Scala guide specifically has an entry for this [Methods with Numerous Arguments](https://docs.scala-lang.org/style/indentation.html#methods-with-numerous-arguments):

```scala
foo(
  someVeryLongFieldName,
  andAnotherVeryLongFieldName,
  "this is a string",
  3.1415)
```

### C

K&R (Linux kernel style), GNU, and Allman would all split lines one way of the other, but never close paren on its own. For example https://www.gnu.org/prep/standards/standards.html#Formatting:

```c
int
lots_of_args (int an_integer, long a_long, short a_short,
              double a_double, float a_float)
…
```

### Java

[Wrapping Lines](http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-136091.html#248)

```java
someMethod(longExpression1, longExpression2, longExpression3, 
        longExpression4, longExpression5);
 
var = someMethod1(longExpression1,
                someMethod2(longExpression2,
                        longExpression3)); 
```

### Object Pascal

[Continuation Lines](http://edn.embarcadero.com/article/10280#4.4)

```pascal
// CORRECT

function CreateWindowEx(dwExStyle: DWORD; 
  lpClassName: PChar; lpWindowName: PChar; 
  dwStyle: DWORD; X, Y, nWidth, nHeight: Integer;
  hWndParent: HWND; hMenu: HMENU; hInstance: HINST; 
  lpParam: Pointer): HWND; stdcall;

// CORRECT 

if ((X = Y) or (Y = X) or
  (Z = P) or (F = J) then
begin
  S := J;
end;
```

### Emacs Lisp

[The Emacs Lisp Style Guide](https://github.com/bbatsov/emacs-lisp-style-guide#syntax)

> Place all trailing parentheses on a single line instead of distinct lines.

```lisp
;; good; single line
(when something
  (something-else))

;; bad; distinct lines
(when something
  (something-else)
)
```

With the aid of automatic formatting (like in case of Emacs Lisp) they could chose either way, but people are picking collapsed closing parenthesis. This is because closing parens do no provide any info if the code is indented correctly.

dwijnand:
yeah a string of closing parens in lisp I know is a thing in the lisp communities.

but lisp is very regular, it's all parens. in scala parens are everywhere mixed in different contexts, with other syntax constructs.

so I'm saying it's clearer in a complex scala codebase like sbt's to space things out.

for instance:

```scala
  private[this] class ChangeOwnerAndModuleClassTraverser(
      oldowner: global.Symbol,
      newowner: global.Symbol)
      extends global.ChangeOwnerTraverser(oldowner, newowner) {
    override def traverse(tree: global.Tree): Unit = {
      tree match {
```

how long does it take you to see how many arguments and which types this class takes?

```scala
  private[this] class ChangeOwnerAndModuleClassTraverser(
      oldowner: global.Symbol,
      newowner: global.Symbol
  ) extends global.ChangeOwnerTraverser(oldowner, newowner) {
    override def traverse(tree: global.Tree): Unit = {
      tree match {
```

how about now?

let's try something else:

```scala
   * Typically, `f` is a `Select` or `Ident`.
   * The wrapper is replaced with the result of `subWrapper(<Type of T>, <Tree of v>, <wrapper Tree>)`
   */
  def transformWrappers(
      t: Tree,
      subWrapper: (String, Type, Tree, Tree) => Converted[ctx.type]): Tree = {
    // the main tree transformer that replaces calls to InputWrapper.wrap(x) with
    //  plain Idents that reference the actual input value
    object appTransformer extends Transformer {
```

how long did it take you to identify the result type of this method?

```scala
   * Typically, `f` is a `Select` or `Ident`.
   * The wrapper is replaced with the result of `subWrapper(<Type of T>, <Tree of v>, <wrapper Tree>)`
   */
  def transformWrappers(
      t: Tree,
      subWrapper: (String, Type, Tree, Tree) => Converted[ctx.type]
  ): Tree = {
    // the main tree transformer that replaces calls to InputWrapper.wrap(x) with
    //  plain Idents that reference the actual input value
    object appTransformer extends Transformer {
```

how about now?

not minutes, of course, but I'd say enough to be a drag that we should fix automatically.

eed3si9n:
The fact that each arguments / parameters are put on its own line, I think is spacing things out enough. Too much vertical spacing actually makes it harder to read, imo.

See for example:

```scala
  def sequential[A0, A1, A2, A3, A4, A5, A6, B](task0: Initialize[Task[A0]],
                                                task1: Initialize[Task[A1]],
                                                task2: Initialize[Task[A2]],
                                                task3: Initialize[Task[A3]],
                                                task4: Initialize[Task[A4]],
                                                task5: Initialize[Task[A5]],
                                                task6: Initialize[Task[A6]],
                                                last: Initialize[Task[B]]): Initialize[Task[B]] =
    sequential(List(unitTask(task0),
                    unitTask(task1),
                    unitTask(task2),
                    unitTask(task3),
                    unitTask(task4),
                    unitTask(task5),
                    unitTask(task6)),
               last)
```

I'd be ok if they were even more packed, but to me the above is acceptable.

```scala
  def sequential[A0, A1, A2, A3, A4, A5, A6, B](
      task0: Initialize[Task[A0]],
      task1: Initialize[Task[A1]],
      task2: Initialize[Task[A2]],
      task3: Initialize[Task[A3]],
      task4: Initialize[Task[A4]],
      task5: Initialize[Task[A5]],
      task6: Initialize[Task[A6]],
      last: Initialize[Task[B]]
  ): Initialize[Task[B]] =
    sequential(
      List(
        unitTask(task0),
        unitTask(task1),
        unitTask(task2),
        unitTask(task3),
        unitTask(task4),
        unitTask(task5),
        unitTask(task6)
      ),
      last
    )
```

The above is less readable to me because now everything is on column 2 and 4, and I have to track one word at a time vertically, instead of reading horizontally.

```scala
  /**
   * Typically, `f` is a `Select` or `Ident`.
   * The wrapper is replaced with the result of `subWrapper(<Type of T>, <Tree of v>, <wrapper Tree>)`
   */
  def transformWrappers(
      t: Tree,
      subWrapper: (String, Type, Tree, Tree) => Converted[ctx.type]): Tree = {
    // the main tree transformer that replaces calls to InputWrapper.wrap(x) with
    //  plain Idents that reference the actual input value
    object appTransformer extends Transformer {
```

The above is just normal looking Scala code to me. What might help would an extra space before the comment like:

```scala
  /**
   * Typically, `f` is a `Select` or `Ident`.
   * The wrapper is replaced with the result of `subWrapper(<Type of T>, <Tree of v>, <wrapper Tree>)`
   */
  def transformWrappers(
      t: Tree,
      subWrapper: (String, Type, Tree, Tree) => Converted[ctx.type]): Tree = {

    // the main tree transformer that replaces calls to InputWrapper.wrap(x) with
    //  plain Idents that reference the actual input value
    object appTransformer extends Transformer {
```

Putting newline before the `)` and making it:

```scala
...
      subWrapper: (String, Type, Tree, Tree) => Converted[ctx.type]
  ): Tree = {
...
```

to me is jarring because my eyes have been wired to look for return type around right-most edge.

We're now deeply in bikeshed weeds. At this point, I'd say we should call on Twitter to let ppl vote 👍 or 👎 for 24h and accept the result.

dwijnand:
> What might help would an extra space before the comment like:
> 
> ```scala
>   /**
>    * Typically, `f` is a `Select` or `Ident`.
>    * The wrapper is replaced with the result of `subWrapper(<Type of T>, <Tree of v>, <wrapper Tree>)`
>    */
>   def transformWrappers(
>       t: Tree,
>       subWrapper: (String, Type, Tree, Tree) => Converted[ctx.type]): Tree = {
> 
>     // the main tree transformer that replaces calls to InputWrapper.wrap(x) with
>     //  plain Idents that reference the actual input value
>     object appTransformer extends Transformer {
> ```

the problem with that is that you can't tell if the object is a part of the method or not.

dwijnand:
> At this point, I'd say we should call on Twitter to let ppl vote 👍 or 👎 for 24h and accept the result.

https://twitter.com/dwijnand/status/983979109918347264

core-macros/src/main/scala/sbt/internal/util/appmacro/TupleNBuilder.scala
```diff
@@ -47,10 +49,12 @@ object TupleNBuilder extends TupleBuilder {
       }
       def extract(param: ValDef): List[ValDef] = bindTuple(param, Nil, inputs.map(_.local), 1)
 
-      def bindTuple(param: ValDef,
-                    revBindings: List[ValDef],
-                    params: List[ValDef],
-                    i: Int): List[ValDef] =
+      def bindTuple(
+          param: ValDef,
```
ScalaWilliam, line 53: This is way better. Nice 👍

dwijnand:
image

@eed3si9n happy for me to rebase and merge this?

build.sbt
```diff
@@ -37,7 +37,7 @@ def buildLevelSettings: Seq[Setting[_]] =
       resolvers += Resolver.mavenLocal,
       scalafmtOnCompile := true,
       scalafmtOnCompile in Sbt := false,
-      scalafmtVersion := "1.3.0",
+      scalafmtVersion := "1.4.0",
```
eed3si9n, line 40: Scalafmt 1.5 is now out. Maybe we should bump it up now?

dwijnand: I chose not to because of https://twitter.com/olafurpg/status/988665279902339073.

PR #4066, opened 2018-04-05
sync reporter logging

eed3si9n:
This is an attempt to fix sbt/zinc#520


eed3si9n:
This didn't fix the issue.

PR #4067, opened 2018-04-05
perf: optimize hash for build

eed3si9n:
This addresses a hot path, which was discovered by @retronym using FlameGraph, by removing intermediate creation of byte Arrays.

`filesModifiedBytes` in particular was bad as it was going through all `*.class` files, each generating an Array. This replaces that with `fileModifiedHash` that accepts `MessageDigest`.

flamegraph-reload-baseline_svg

According to the flamegraph, evalCommon footprint reduced from 4.5% to 3.6%. See https://gist.github.com/eed3si9n/d68f3aa91be7bbe55b0c8cf77e37cf98.
Using `time sbt reload reload reload exit`, the median reduced from 41.450s to 39.467s.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -310,6 +310,10 @@ lazy val actionsProj = (project in file("main-actions"))
     name := "Actions",
     libraryDependencies += sjsonNewScalaJson.value,
     mimaSettings,
+    mimaBinaryIssueFilters ++= Seq(
+      exclude[DirectMissingMethodProblem]("sbt.compiler.Eval.filesModifiedBytes"),
+      exclude[DirectMissingMethodProblem]("sbt.compiler.Eval.fileModifiedBytes"),
+    )
   )
   .configure(
     addSbtIO,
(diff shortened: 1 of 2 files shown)


eed3si9n:
/cc @cunei 

If we focus on the remaining 3.6%, it turns out that one of the hot paths is `sbt.internal.io.MilliNative.getModifiedTime`.

evalcommon

So I tried `-Dsbt.io.jdktimestamps=true`:

```
$ time sbt -Dsbt.io.jdktimestamps=true reload reload reload exit
```

This reduced the median from 39.467s to 37.275s (with minimum measurement 35.674s). Given that build timestamp is not subject to subsecond incremental compilation, non-milisec accuracy I think is worth considering for this use case.

dwijnand (dismissed):
lgtm.

I've wanted to land these changes since October..

dwijnand (approved):

fommil:
you'll get even better timestamp performance if you use nio to walk a directory which means you get all timestamps of a directory in one OS call. I hope a change to fully nio is on the horizon for 2.0.

dwijnand:
good to know!

PR #4070, opened 2018-04-05
Add dependencyResolution scoped to updateSbtClassifiers task

### LINKED ISSUE #3432: `updateSbtClassifiers` tried wrong bintray doc and src address.

## steps
Use IDEA refresh project via sbt-shell with "Download sbt and plugin src and doc" checked.

## problem
IDEA executed `updateSbtClassifiers` and failed to download src and doc because of wrong uri:
```
https://dl.bintray.com/cuzfrog/maven/com/github/cuzfrog/sbt-tmpfs_2.12_1.0/0.3.0/sbt-tmpfs-0.3.0-sources.jar
```

and it did not try 
```
https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/com.github.cuzfrog/sbt-tmpfs/scala_2.12/sbt_1.0//0.3.0/docs/sbt-tmpfs-javadoc.jar
```
## notes

use `reload plugins` then `updateClassifiers` worked around this.

sbt version: *1.0.0*

steinybot:
Fixes #3432 

The test isn't great but it does the job. Ideally it would check that the configuration that is used when creating `IvySbt` is the same that is passed to `update` but that was a bit tricky.

The contributing guidelines say that there should be notes (https://github.com/sbt/sbt/blob/1.x/CONTRIBUTING.md#adding-notes) but it doesn't seem to match up with what was there. Let me know if I need to add this.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2168,6 +2168,7 @@ object Classpaths {
                 filterImplicit = false,
                 overrideScalaVersion = true).withScalaOrganization(scalaOrganization.value))
           },
+          dependencyResolution := IvyDependencyResolution(ivyConfiguration.value),
           updateSbtClassifiers in TaskGlobal := (Def.task {
             val lm = dependencyResolution.value
             val s = streams.value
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):

eed3si9n:
Thanks for the fix!

PR #4052, opened 2018-03-29
Use MacOSXWatchService instead of PollingWatchService

eatkins:
This watch service should be more responsive and significantly reduce
the disk overhead of the polling based service for large repos.

(See the guidelines for contributing, linked above)

PR #4071, opened 2018-04-06
Use MacOSXWatchService instead of PollingWatchService

eed3si9n:
This is a resend of https://github.com/sbt/sbt/pull/4052.

It bumps up IO to 1.1.5, and also adds new `closewatch` mode.

/cc @eatkins


eed3si9n:
Merging this since #4052 has passed the review.

Issue #4072, opened 2018-04-06
migrate last-grep to lastGrep


dwijnand:
## steps

try using `lastGrep`

## problem

```
> lastGrep error
[error] Expected ID character
[error] Not a valid command: lastGrep (similar: last-grep, last, iflast)
[error] Expected project ID
[error] Expected configuration
[error] Expected ':'
[error] Expected key
[error] Not a valid key: lastGrep (similar: streams)
[error] lastGrep error
[error]         ^
```

## expectation

```
> lastGrep error
[error] /s/t-last-grep/Main.scala:6:3: ')' expected but '}' found.
[error]   }
[error]   ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
[error] Total time: 0 s, completed 06-Apr-2018 08:47:37
```

## notes

sbt version: 1.1.2

refs https://github.com/sbt/website/pull/654

note that we should need to maintain a deprecated `last-grep` for a while.

dwijnand:
fixed in #4080, due in 1.2.0

PR #4080, opened 2018-04-08
Add lastGrep command

naferx:
Ref #4072

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -40,7 +40,6 @@ import sbt.internal.util.{
   Types
 }
 import sbt.util.{ Level, Logger, Show }
-
 import sbt.internal.util.complete.{ DefaultParsers, Parser }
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
LGTM

eed3si9n:
Thanks for this contribution.
I opened a new issue https://github.com/sbt/sbt/issues/4081 for other commands, if you're up for it.

naferx:
@eed3si9n Yeah, I'd like to keep contributing, Thanks! 
I was wondering how to inform to the user that the command is being deprecated  
I have in mind two alternatives: 
 * Include it in the (old) detailed message
 * When the user types the (old) command it raises a warning log

dwijnand:
> When the user types the (old) command it raises a warning log

that would be the more important effect, to me. I don't think there's a built in way to do this in sbt's commands.

eed3si9n:
Could we output log with warning level or something?

naferx:
I'll give it a try over the next days

PR #4090, opened 2018-04-11
Fixes linter that detects missing .value

### LINKED ISSUE #4079: warn on value discard in build code
## steps
someone declares TaskKey[Unit] and tries to define it, but doesn't know about .value or accidentally omits it
```scala
lazy val npmBuildTask = taskKey[Unit]("Execute the npm build command to build the ui")
npmBuildTask := {
  Process("npm run build", file(appPath))
}
lazy val npmInstallTask = taskKey[Unit]("Execute the npm install command to build the ui")
npmInstallTask := {
  Process("npm install", file(appPath))
}
lazy val npmTask = taskKey[Unit]("execute all npm tasks")
npmTask := {
  npmInstallTask
  npmBuildTask
}
```

## problem
nothing happens

## expectation
useful error message or warning

## notes
i suggest enabling -Ywarn-value-discard for builds by default

eed3si9n:
Fixes #4079

#3216 introduced a linter that checks against missing `.value`, but the tree only checked for `Ident`. This doesn't work because in reality the symbols of build.sbt are transformed to `$somehash.npmInstallTask` where `somehash` is the wrapper object we create. Similarly for the built-in keys, they are presented as `sbt.Keys.compile`.

With this change unused task will fail to load the build with the following message:

```
/sbt-4079/build.sbt:26: error: The key `compile` is not being invoked inside the task definition.

Problem: Keys missing `.value` are not initialized and their dependency is not registered.

Solution: Replace `compile` by `compile.value` or remove it if unused.

  compile
  ^
/sbt-4079/build.sbt:27: error: The key `npmInstallTask` is not being invoked inside the task definition.

Problem: Keys missing `.value` are not initialized and their dependency is not registered.

Solution: Replace `npmInstallTask` by `npmInstallTask.value` or remove it if unused.

  npmInstallTask
  ^
```

### DIFF
--- a/main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
+++ b/main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
@@ -24,9 +24,6 @@ abstract class BaseTaskLinterDSL extends LinterDSL {
     val isTask = convert.asPredicate(ctx)
     class traverser extends Traverser {
       private val unchecked = symbolOf[sbt.sbtUnchecked].asClass
-      private val taskKeyType = typeOf[sbt.TaskKey[_]]
-      private val settingKeyType = typeOf[sbt.SettingKey[_]]
-      private val inputKeyType = typeOf[sbt.InputKey[_]]
       private val initializeType = typeOf[sbt.Def.Initialize[_]]
       private val uncheckedWrappers = MutableSet.empty[Tree]
       var insideIf: Boolean = false
@@ -57,8 +54,8 @@ abstract class BaseTaskLinterDSL extends LinterDSL {
         }
       }
 
-      @inline def isKey(tpe: Type): Boolean =
-        tpe <:< initializeType || tpe <:< taskKeyType || tpe <:< settingKeyType || tpe <:< inputKeyType
+      @inline def isKey(tpe: Type): Boolean = isInitialize(tpe)
+      @inline def isInitialize(tpe: Type): Boolean = tpe <:< initializeType
 
       def detectAndErrorOnKeyMissingValue(i: Ident): Unit = {
         if (isKey(i.tpe)) {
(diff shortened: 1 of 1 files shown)


main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
```diff
@@ -57,7 +58,7 @@ abstract class BaseTaskLinterDSL extends LinterDSL {
       }
 
       @inline def isKey(tpe: Type): Boolean =
-        tpe <:< taskKeyType || tpe <:< settingKeyType || tpe <:< inputKeyType
+        tpe <:< initializeType || tpe <:< taskKeyType || tpe <:< settingKeyType || tpe <:< inputKeyType
```
dwijnand: this should simplify to just `tpe <:< initializeType` because keys are initializes

main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
```diff
@@ -66,6 +67,13 @@ abstract class BaseTaskLinterDSL extends LinterDSL {
         } else ()
       }
 
+      def detectAndErrorOnKeyMissingValue(s: Select): Unit = {
```
dwijnand, line 67: instead of overloading you could just make the other one take a tree

dwijnand (approved):
cool.

minor comments, but this can merge as is too

OlegYch:
does it work for something like `(Compile / compile)` ?

dwijnand:
yes, that's an Initialize too.

main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
eed3si9n: This code uses `Select#name`.

eed3si9n: I guess I can do pattern matching.

eed3si9n:
> does it work for something like `(Compile / compile)` ?

Actually it won't as it stands, because it's only handling Ident and Select. We need Apply for that.

main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
eed3si9n: Good point. Let me try that.

main-settings/src/main/scala/sbt/std/TaskLinterDSL.scala
dwijnand: oh, for some reason I thought all trees had name??

dwijnand:
Would probably be best for this to not miss the 1.2.0-RC1 train.

eed3si9n:
Added a commit.

dwijnand (approved):
LGTM, pending Travis CI.

Issue #4091, opened 2018-04-11
sbt server: responses to unknown requests


laughedelic:
See https://github.com/scalameta/metals/issues/233#issuecomment-379719556 for the context.

## steps

1. After initialization server runs some command internally, but sends its response to the client:

    ```
    13:16:41.785 ERROR s.m.m.s.SbtServer$ - Response to unknown request: Success({
      "status" : "Done",
      "channelName" : "network-3",
      "execId" : "2",
      "commandQueue" : [
        "shell"
      ],
      "exitCode" : 0
    },String("2"))
    ```

    @dwijnand pointed here:

    https://github.com/sbt/sbt/blob/314ff4643278c9f00d408ed2398c61c0ea4bd7b9/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala#L67

2. Also if I reload sbt shell while there is an open connection, the client receives some responses that weren't meant for it (wrong channel):

    ```
    13:19:18.194 ERROR s.m.m.s.SbtServer$ - Response to unknown request: Success({
      "status" : "Done",
      "channelName" : "console0",
      "execId" : "6f7b96a9-5eed-4098-acc9-2fd3a654eb4c",
      "commandQueue" : [
        "shell"
      ],
      "exitCode" : 0
    },String("6f7b96a9-5eed-4098-acc9-2fd3a654eb4c"))
    13:19:18
    13:19:20.487 ERROR s.m.m.s.SbtServer$ - Response to unknown request: Success({
      "status" : "Done",
      "channelName" : "console0",
      "execId" : "aabe6bf4-f24c-4804-97c3-0802570192a8",
      "commandQueue" : [
        "sbtStashOnFailure",
        "onFailure load-failed",
        "loadp",
        "sbtPopOnFailure",
        "resumeFromFailure",
        "shell"
      ],
      "exitCode" : 0
    },String("aabe6bf4-f24c-4804-97c3-0802570192a8"))
    ```

## problem

Client gets confused about these responses.

## expectation

Server shouldn't send responses to non-existent request and broadcast messages from the console channel.

## notes

sbt version: 1.1.2+

laughedelic:
#4093 fixed the first problem and had a comment about the second one:

> And I think the second one sits here:
> 
> https://github.com/sbt/sbt/blob/7ae83cc755e5b7f0aeb404163861c7966189688e/main/src/main/scala/sbt/internal/CommandExchange.scala#L250-L256
> 
[...]

PR #4093, opened 2018-04-11
WIP: Fix unexpected responses from the server

laughedelic:
Addresses #4091. 

85ab082 should fix the first problem. 

And I think the second one sits here:

https://github.com/sbt/sbt/blob/7ae83cc755e5b7f0aeb404163861c7966189688e/main/src/main/scala/sbt/internal/CommandExchange.scala#L250-L256

But I'm not sure how to fix it. In this case it matches any events (so we cannot check channel name here) and redirects them to the [`NetworkChannel.publishEvent`](https://github.com/sbt/sbt/blob/7ae83cc755e5b7f0aeb404163861c7966189688e/main/src/main/scala/sbt/internal/server/NetworkChannel.scala#L239) which already doesn't check if the channel name matches.

On the other hand there is also [`publishEventMessage`](https://github.com/sbt/sbt/blob/7ae83cc755e5b7f0aeb404163861c7966189688e/main/src/main/scala/sbt/internal/CommandExchange.scala#L316) method which handles `ExecStatusEvent` and checks channel name. So probably we could reuse it here. 

I'm not sure what is the general intent of this piece of code quoted above: if it's not a `StringMessage`, just broadcast is on all channels? Even if we want to do that, `publishEvent` is not a good choice, probably using `langNotify` or something similar would be better.

### DIFF
--- a/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
@@ -64,7 +64,7 @@ private[sbt] trait LanguageServerProtocol extends CommandChannel {
           else throw LangServerError(ErrorCodes.InvalidRequest, "invalid token")
         } else ()
         setInitialized(true)
-        append(Exec(s"collectAnalyses", Some(request.id), Some(CommandSource(name))))
+        append(Exec(s"collectAnalyses", None, Some(CommandSource(name))))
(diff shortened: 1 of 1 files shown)


dwijnand (approved):

dwijnand:
sadly I can't help much with the other part, because the general intent is not clear to me either.

PR #4094, opened 2018-04-11
Further optimize Eval.fileModifiedHash with nio

dwijnand:
working off of @fommil's info in https://github.com/sbt/sbt/pull/4067#issuecomment-380580693.

haven't benchmarked this

### DIFF
--- a/main-actions/src/main/scala/sbt/compiler/Eval.scala
+++ b/main-actions/src/main/scala/sbt/compiler/Eval.scala
@@ -510,9 +510,10 @@ private[sbt] object Eval {
       f.toPath,
       new SimpleFileVisitor[NioPath] {
         override def visitFile(file: NioPath, attrs: BasicFileAttributes) = {
-          if (classDirFilter accept file.toFile)
+          val f = file.toFile
+          if (classDirFilter accept f)
             digester.update(bytes(attrs.lastModifiedTime.toMillis))
-          digester.update(bytes(file.toFile.getAbsolutePath))
+          digester.update(bytes(f.getAbsolutePath))
           FileVisitResult.CONTINUE
         }
         override def postVisitDirectory(dir: NioPath, exc: IOException): FileVisitResult = {
--- a/main-actions/src/main/scala/sbt/compiler/Eval.scala
+++ b/main-actions/src/main/scala/sbt/compiler/Eval.scala
@@ -518,7 +518,7 @@ private[sbt] object Eval {
         }
         override def postVisitDirectory(dir: NioPath, exc: IOException): FileVisitResult = {
           if (exc eq null) {
-            digester.update(bytes(dir.toFile.getAbsolutePath))
+            digester.update(bytes(dir.toAbsolutePath.toString))
             FileVisitResult.CONTINUE
           } else throw exc
         }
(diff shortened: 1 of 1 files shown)


main-actions/src/main/scala/sbt/compiler/Eval.scala
```diff
@@ -503,25 +505,26 @@ private[sbt] object Eval {
...
-    if (f.isDirectory)
-      (f listFiles classDirFilter) foreach { x =>
-        fileModifiedHash(x, digester)
-      } else digester.update(bytes(getModifiedTimeOrZero(f)))
-
-    digester.update(bytes(f.getAbsolutePath))
+    Files.walkFileTree(
+      f.toPath,
+      new SimpleFileVisitor[Path] {
+        override def visitFile(file: Path, attrs: BasicFileAttributes) = {
```
fommil: from memory you can use `visitDirectory` to get the `attrs` of all the files in the dir.

main-actions/src/main/scala/sbt/compiler/Eval.scala
```diff
@@ -503,25 +505,26 @@ private[sbt] object Eval {
...
+          if (classDirFilter accept file.toFile)
+            digester.update(bytes(attrs.lastModifiedTime.toMillis))
+          digester.update(bytes(file.toFile.getAbsolutePath))
```
fommil: there should be a way to get the absolute path without going via File... see https://github.com/ensime/ensime-server/blob/3.0/api/src/main/scala/org/ensime/io/Canon.scala#L31-L42

fommil: also, avoiding two `.toFile` calls would be good here

dwijnand: is toFile expensive?

main-actions/src/main/scala/sbt/compiler/Eval.scala
dwijnand: I don't see how to do that. preVisitDirectory only has one attrs, with one lastModifiedTime for the directory itself.

main-actions/src/main/scala/sbt/compiler/Eval.scala
fommil: `Path` uses an optimised hierarchical representation, whereas `File` constructs a full `String`

when I've profiled sbt on large projects, this has been the single biggest cause of heap usage.

dwijnand: the alternative I found is `dir.toAbsolutePath.toString`

fommil: it's a shame that you need to construct the string for the callback... I guess **that** optimisation needs to wait until the API can be changed.

dwijnand: re-implemented!

dwijnand: oh you mean the other string

dwijnand:
moving to ticket https://github.com/sbt/sbt/issues/4116.

PR #4096, opened 2018-04-12
Watch improvements

eatkins:

### DIFF
--- a/main-command/src/main/scala/sbt/Watched.scala
+++ b/main-command/src/main/scala/sbt/Watched.scala
@@ -12,7 +12,7 @@ import java.nio.file.FileSystems
 
 import sbt.BasicCommandStrings.ClearOnFailure
 import sbt.State.FailureWall
-import sbt.internal.io.{ Source, SourceModificationWatch, WatchState }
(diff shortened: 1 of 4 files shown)


dwijnand (dismissed):

eatkins:
This PR was broken upstream. I'll have a fix up by tomorrow (along with a related io bugfix).

eed3si9n:
I combined this PR with https://github.com/sbt/sbt/pull/4145 and added d8fe09f00776ad45d404fe0319df237bef8cc00f that tracks the rename upstream.

eed3si9n:
@eatkins 

> (along with a related io bugfix).

Is that a show stopper? Should I hold off sbt 1.1.5?

eatkins:
Yeah. Let me make a PR now. Won't be long.

eatkins:
The bug I was referring to is arguably not show stopping, but it can make the ux annoying. It can cause the user to have to hit enter twice to exit a watched build see https://github.com/sbt/io/pull/149. Sorry I didn't make the PR earlier since fixing will require publishing another io version.

eed3si9n:
ok. Thanks for following up.

Issue #3845, opened 2017-12-30
Deleted files do not trigger ~


blast-hardcheese:
Followup to https://github.com/sbt/sbt/issues/3695

## steps

Create a file, `src/main/foo.scala` with some content (`foo.scala!` in my test case)
Watch for the recompile (and error, as our file was invalid syntax)
Delete the file
Nothing happens

## problem

Deleted files don't trigger recompile

## expectation

Deleted files modify the parent directory, which triggers recompile

## notes

MacOS 10.12.6
sbt versions: 0.13.13, 1.1.0-RC4

dwijnand:
fixed by @eatkins in https://github.com/sbt/sbt/pull/4098. coming in sbt 1.1.5.

PR #4098, opened 2018-04-13
Exclude directories instead of including files

eatkins:
The existing filter caused SourceModificationWatch.watch to ignore
deleted files because !file.exists implies !file.isFile. The intention
of the filter was to exclude directories that had a name ending in
".scala".


eatkins:
This should fix https://github.com/sbt/sbt/issues/3845

dwijnand (approved):
nice! thank you!

jrudolph:
It seems this broke triggered execution for twirl: https://github.com/playframework/twirl/issues/186

PR #4099, opened 2018-04-13
Handle managedSources writing into unmanaged source directories

eatkins:
When source generators write into the unmanaged source directory, bad
things can happen. Continuous builds will loop indefinitely and
compiling will fail because the generated sources get added to the
source list twice, causing the incremental compiler to complain about
compiling classes it has already seen. My two-pronged solution is to
de-duplicate the sources task and to filter out managed source files in
watch sources. The drawback to the latter is that it causes the source
generation task to be executed twice per compile.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -329,7 +329,18 @@ object Defaults extends BuildCommon {
       val baseDir = baseDirectory.value
       val bases = unmanagedSourceDirectories.value
       val include = (includeFilter in unmanagedSources).value
-      val exclude = (excludeFilter in unmanagedSources).value
+      val exclude = (excludeFilter in unmanagedSources).value match {
+        case e =>
+          (managedSources in ThisScope).value match {
+            case l if l.nonEmpty =>
+              e || new FileFilter {
+                private val files = l.toSet
+                override def accept(pathname: File): Boolean = files.contains(pathname)
+                override def toString = s"ManagedSourcesFilter($files)"
+              }
+            case _ => e
+          }
+      }
       val baseSources =
         if (sourcesInBase.value) Seq(new Source(baseDir, include, exclude, recursive = false))
         else Nil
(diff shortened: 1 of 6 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -342,7 +356,7 @@ object Defaults extends BuildCommon {
     sourceDirectories := Classpaths
       .concatSettings(unmanagedSourceDirectories, managedSourceDirectories)
       .value,
-    sources := Classpaths.concat(unmanagedSources, managedSources).value
+    sources := Classpaths.concat(unmanagedSources, managedSources).value.distinct
```
dwijnand: is there no way to avoid this?

dwijnand: reason I ask is `distinct` is like null-checks: you could fill your codebase with them. so I'd rather avoid it, and fix the root cause - if possible.

eatkins: Looking more at this file, it could be `sources := Classpaths.concatDistinct(unmanagedSources, managedSources).value` (probably not what you mean though...). I can't really think of a way to guarantee that there is no overlap between managed sources and unmanaged sources since there is nothing that enforces that these files live in isolated directories. I guess you could do something like
```scala
sources := {
  val unmanaged = unmanagedSources.value
  val unmanagedDirectories = unmanaged.view.map(_.toPath.getParent).toSet
  unmanaged ++ managedSources.value.filterNot(f => unmanagedDirectories.contains(p.toPath.getParent))
}
```

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -330,9 +330,23 @@ object Defaults extends BuildCommon {
...
+                override def accept(pathname: File): Boolean = files.contains(pathname)
+                override def toString = s"ManagedSourcesFilter($files)"
+              }
+            case _ => e
+          }
+      }
       val baseSources =
-        if (sourcesInBase.value) Seq(new Source(baseDir, include, exclude, recursive = false))
+        if (sourcesInBase.value)
+          Seq(new Source(baseDir, include, exclude || new SimpleFileFilter({ f =>
```
eatkins: I can remove this diff if this commit is brought in after #4096 (which makes the recursive source filter out files not in the base path).

main/src/main/scala/sbt/Defaults.scala
dwijnand: OK.. Thanks for spending the time to look into this. Let's use `Classpaths.concatDistinct` I guess.

main/src/main/scala/sbt/Defaults.scala
dwijnand: OK, let's see how things land.

eed3si9n (approved):

PR #4110, opened 2018-04-20
Remove collectAnalyses step in sbt server initialize.

olafurpg:
A client should trigger this command manually instead. Currently,
this automatic step triggers compilation in all root-aggregated
projects, which for large projects may take a long time to run.
It is better left to the client to decide what exact command to run on
startup, for example it might be desirable to scope the command by
custom project/config.

### DIFF
--- a/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
@@ -64,7 +64,6 @@ private[sbt] trait LanguageServerProtocol extends CommandChannel {
           else throw LangServerError(ErrorCodes.InvalidRequest, "invalid token")
         } else ()
         setInitialized(true)
-        append(Exec(s"collectAnalyses", Some(request.id), Some(CommandSource(name))))
         langRespond(InitializeResult(serverCapabilities), Option(request.id))
       case "textDocument/definition" =>
         import scala.concurrent.ExecutionContext.Implicits.global


olafurpg:
Discovered while using https://github.com/scalacenter/load-plugin to automatically enable sbt-metals and run `semanticdbEnable`. Currently, compilation is automatically triggered before `semanticdbEnable` which is problematic since the compiler plugin is required for definition/references to work.

Credit to @laughedelic for finding culprit in https://github.com/scalameta/metals/pull/287

olafurpg:
For example, establishing a connection to sbt server in a freshly cloned Spire repo (https://github.com/non/spire) makes the shell freeze for ~3 minutes without any indication what is going on except the CPU is spinning at 300%. This is because `collectAnalyses` is unscoped and runs for all modules in all platforms (js/jvm). In larger repos this startup time is even higher.

dwijnand:
the motivation for `collectAnalyses` is so the server is ready for `textDocument/definition` calls: https://github.com/sbt/sbt/commit/73b0034cfc7db7a03f84d6098829f08b0ec862db

what are the general expectations from LSP clients (in other languages) around `textDocument/definition`? is it acceptable for the first invocation to take a while, while some info collection is happening?

what I'm saying is I think dropping the compilation on initialise (which, btw, _do_ I agree should happen - it's a hack) should either be accompanied with a change in `textDocument/definition` to collect the analyses if not present, or with a change in sbt's lsp client to trigger the collection (at some point, either after initialise or before calling textDocument/definition).

olafurpg:
> what are the general expectations from LSP clients (in other languages) around textDocument/definition? is it acceptable for the first invocation to take a while, while some info collection is happening?

LSP supports dynamic registration of server capabilities https://microsoft.github.io/language-server-protocol/specification#client_registerCapability this allows the server to notify the client when `textDocument/definition` is ready to be used. Note that this feature may not work in all clients.

> should either be accompanied with a change in textDocument/definition to collect the analyses if not present, or with a change in sbt's lsp client to trigger the collection (at some point, either after initialise or before calling textDocument/definition).

This opens the question of whether the primary purpose sbt server is to accommodate the needs of https://marketplace.visualstudio.com/items?itemName=lightbend.vscode-sbt-scala or to enable an ecosystem of tools to programmatically communicate with sbt. I would love it to be the latter. I don't agree that tooling authors who want to integrate with sbt server for non-sbt-LSP purposes are responsible for fixing the sbt LSP implementation. I am excited about the potential for sbt server, but the current tight coupling with the sbt LSP implementation makes it practically unusable for other purposes.

olafurpg:
> makes the shell freeze for ~3 minutes without any indication what is going on except the CPU is spinning at 300%.

I'd like to stress that this is a UX issue for sbt LSP as well. I never do top-level `test:compile` in my larger projects like scalameta/scalameta or scalacenter/scalafix. In scalafix, we don't use `crossScalaVersions` so `collectAnalyses` would run compilation for the full JS/JVM + 2.11/2.12 matrix.

dwijnand:
> This opens the question of whether the primary purpose sbt server is to accommodate the needs of marketplace.visualstudio.com/items?itemName=lightbend.vscode-sbt-scala or to enable an ecosystem of tools to programmatically communicate with sbt. I would love it to be the latter.

in sbt 1.0 it was just the later, in sbt 1.1 sbt server switched to use the language server protocol and so, for instance, the "tell sbt to run this command" switched to "sbt/exec" and some initial LSP-specific features per implemented (initialise, textDocument/definition).

> I don't agree that tooling authors who want to integrate with sbt server for non-sbt-LSP purposes are responsible for fixing the sbt LSP implementation.

I agree, but I think landing this PR causes a _regression_ in sbt's LSP solution, which is why I commented.

> I am excited about the potential for sbt server, but the current tight coupling with the sbt LSP implementation makes it practically unusable for other purposes.

sure, ok, so what's a better solution? do you mean split the client out of this repo?

laughedelic:
> I agree, but I think landing this PR causes a regression in sbt's LSP solution

Do you think adding an explicit `sbt/exec(collectAnalyses)` call _from the client_ would be enough to avoid this regression?

> sure, ok, so what's a better solution? do you mean split the client out of this repo?

IMO yes, it would also make it easier to iterate on contributions. The vscode extension code is completely independent and is just one of possible client implementations, so I don't see why it should be in sbt/sbt (even if it's kind of a reference implementation).

laughedelic:
> in sbt 1.0 it was just the later, in sbt 1.1 sbt server switched to use the language server protocol

I think there are two things getting mixed together: there's sbt server as it was initially implemented in 1.0 and there is LS**P** as a _protocol_ which, I think, was chosen simply because it makes it _easier_ to integrate sbt server with other clients (not necessarily editors). Using LSP as a communication protocol for a build server doesn't make it immediately a _language server_. I might be wrong, but when you say "sbt's LSP solution" it sounds to me like "sbt language server". I think that if BSP existed a year ago, we would see sbt 1.0 with a BSP-based server. My point is that changing to LSP after 1.1 doesn't really redefine what sbt server is: a way "to enable an ecosystem of tools to programmatically communicate with sbt".

olafurpg:
> in sbt 1.0 it was just the later, in sbt 1.1 sbt server switched to use the language server protocol 

I think this was a great decision, the way I see it is that sbt server implements JSON-RPC which is flexible enough to encode both protocol specific endpoints like `textDocument/publishDiagnostics` but also extensions like `sbt/exec`.

However, `textDocument/definition` only relevant for small set of sbt clients, it makes more sense for clients who are interested in definition to trigger `sbt/exec` with `collectAnalyses` after the `initialize` handshake. The client could for example have configuration for the user to run `moreSpecificProject/collectAnalyses` to make startup snappier.

Code navigation is a hard problem, it goes beyond goto definition in project sources. It includes dependencies, java sources, classfiles, and more, see https://geirsson.com/post/2018/03/ide/#navigation It is not a good strategy IMO to have navigation implemented in sbt/sbt, which is a stable project with slow release cadence.

dwijnand:
> Do you think adding an explicit `sbt/exec(collectAnalyses)` call _from the client_ would be enough to avoid this regression?

yep

> Using LSP as a communication protocol for a build server doesn't make it immediately a _language server_. I might be wrong, but when you say "sbt's LSP solution" it sounds to me like "sbt language server".

hmm, maybe I think I mean that it switched from what it had to JSON-RPC. but by "sbt's LSP solution" I meant the (limited) language server functionality that sbt does support now.

> However, `textDocument/definition` only relevant for small set of sbt clients, it makes more sense for clients who are interested in definition to trigger `sbt/exec` with `collectAnalyses` after the `initialize` handshake.

do you know what the expected behaviour of `textDocument/definition` is if analysis info isn't readily on hand?

olafurpg:
I opened a ticket explaining our use-case and why `collectAnalyses` is problematic https://github.com/sbt/sbt/issues/4113

I hope we can find a solution for this that does not regress the behavior of the sbt language server.

olafurpg:
> do you know what the expected behaviour of textDocument/definition is if analysis info isn't readily on hand?

We also have an indexing phase in metals, which includes compilation in addition to indexing sources.jar from `updateClassifiers`. Until those analyses are complete `textDocument/definition` returns an empty response. This is similar how IntelliJ behaves.

dwijnand:
ok thanks, then we can just get the client to request collection at some point

PR #4115, opened 2018-04-24
Drop -Ywarn-unused:-params

dwijnand:
Previously we'd get in the build logs:

    [error] params cannot be negated, it enables other arguments

and lots of wawrnings.

Now we just get lots of warnings without the non-fatal error message.


eed3si9n (approved):
Yes

PR #4082, opened 2018-04-09
Re-fixing console and JLine

eed3si9n:
Fixes #3482 take 2

I thought I tested #4054 using a local build, but when I ran 1.1.3, `console` did not display anything that I typed.
Switching to `usingTerminal` which calls `terminal.restore` similar to what I had in 1.1.1 fixes `console`.

Here's a session using a locally built 1.1.4-LOCAL-M1:

```
sbt:Hello> console
[info] Starting scala interpreter...
Welcome to Scala 2.12.4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_151).
Type in expressions for evaluation. Or try :help.

scala> 1 + 1
res0: Int = 2

scala> :q

[success] Total time: 6 s, completed Apr 9, 2018 12:08:19 AM
sbt:Hello> console
[info] Starting scala interpreter...
Welcome to Scala 2.12.4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_151).
Type in expressions for evaluation. Or try :help.

// made sure up arrow still works second time
scala> :q

[success] Total time: 5 s, completed Apr 9, 2018 12:08:25 AM
sbt:Hello> :q
[info] shutting down server
```

PR #4123, opened 2018-04-25
Re-fix console and JLine bug

### LINKED ISSUE #3482: 1.0.{0,1,2} Opening console multiple times breaks escaping

eed3si9n:
Fixes #3482 take 3

There are two bugs related REPL and JLine.
1. JLine getting disabled (up arrow not working) on the second run of `console` task.
2. Typed characters not showing up even on the `console` task.

The first issue was fixed by #4054 which added `t.init`. When I noticed the second problem, I fixed it in #4082 (adds `t.restore`) but undid the first fix. This attempts to fix both the issues.

### notes

I confirmed using Scala 2.12.3, 2.12.4, and 2.12.5.

### DIFF
--- a/main-actions/src/main/scala/sbt/Console.scala
+++ b/main-actions/src/main/scala/sbt/Console.scala
@@ -41,7 +41,8 @@ final class Console(compiler: AnalyzingCompiler) {
       implicit log: Logger): Try[Unit] = {
     def console0() =
       compiler.console(classpath, options, initialCommands, cleanupCommands, log)(loader, bindings)
-    JLine.usingTerminal { _ =>
+    JLine.usingTerminal { t =>
+      t.init
       Run.executeTrapExit(console0, log)
     }
   }
--- a/build.sbt
+++ b/build.sbt
@@ -9,7 +9,7 @@ def buildLevelSettings: Seq[Setting[_]] =
   inThisBuild(
     Seq(
       organization := "org.scala-sbt",
-      version := "1.1.4-SNAPSHOT",
+      version := "1.1.5-SNAPSHOT",
       description := "sbt is an interactive build tool",
       bintrayOrganization := Some("sbt"),
       bintrayRepository := {


main-actions/src/main/scala/sbt/Console.scala
```diff
@@ -41,7 +41,8 @@ final class Console(compiler: AnalyzingCompiler) {
       implicit log: Logger): Try[Unit] = {
     def console0() =
       compiler.console(classpath, options, initialCommands, cleanupCommands, log)(loader, bindings)
-    JLine.usingTerminal { _ =>
+    JLine.usingTerminal { t =>
+      t.init
```
dwijnand, line 45: `usingTerminal` is (from https://github.com/sbt/sbt/blob/v1.1.4/internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala):

```scala
  def usingTerminal[T](f: jline.Terminal => T): T =
    withTerminal { t =>
      t.restore
      val result = f(t)
      t.restore
      result
    }
```

(`withTerminal` ensures synchronised access to the terminal)

So this does `t.restore`, `t.init`, run console, `t.restore`.

Is that right?

eed3si9n: Yea. We need both `t.restore` *and* `t.init`, it seems.

cunei: I know nothing about jline, so a cursory search led me to the canonical https://github.com/jline/jline2/issues/210 and the jruby thread. If calling both restore and init works around the problem, that looks ok to me I suppose. The API docs offer no help regarding those calls. However, nosing around the jline source I found:
```
    public void reset() throws Exception {
        restore();
        init();
    }
```
which suggests the combination is indeed the right one.
https://github.com/jline/jline2/blob/master/src/main/java/jline/TerminalSupport.java#L65

cunei (approved):

dwijnand (approved):
3rd times the charm. ☘️

PR #4126, opened 2018-04-27
upgrade Scala 2.12.4 -> 2.12.6

SethTisue:
I have no specific user-visible benefit in mind other than fewer
JARs to download for people who are on current versions of things

PR #4129, opened 2018-04-28
upgrade Scala 2.12.4 -> 2.12.6

eed3si9n:
This is a continuation of https://github.com/sbt/sbt/pull/4126

RunFromSourceMain currently assumes that Scala is already available in boot directory, which doesn't work if we tried to upgrade to another Scala version. This adds a workaround by letting Ivy resolve it like Launcher does internally.


dwijnand (approved):
LGTM. Cool.

Issue #3624, opened 2017-10-09
[vs-code_scala-sbt-plugin] Not such file or directory : active json


wilmer-daza:
(See the guidelines for contributing, linked above)

## steps
1. Install the scala sbt extension for VS CODE
2. Open the vs code editor

## problem
1. Here you have the problem:

```
fs.js:584
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^

Error: ENOENT: no such file or directory, open '/home/wilmer/Projects-git/poc-colis360/project/target/active.json'
    at Object.fs.openSync (fs.js:584:18)
    at Object.module.(anonymous function) [as openSync] (ELECTRON_ASAR.js:173:20)
    at Object.fs.readFileSync (fs.js:491:33)
    at Object.fs.readFileSync (ELECTRON_ASAR.js:505:29)
    at discoverUrl (/home/wilmer/.vscode/extensions/lightbend.vscode-sbt-scala-0.0.2/server/server.js:24:34)
    at Object.<anonymous> (/home/wilmer/.vscode/extensions/lightbend.vscode-sbt-scala-0.0.2/server/server.js:6:9)
    at Object.<anonymous> (/home/wilmer/.vscode/extensions/lightbend.vscode-sbt-scala-0.0.2/server/server.js:28:3)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
[Info  - 3:04:55 PM] Connection to server got closed. Server will restart.
```

## notes

sbt version: *insert sbt version*

eed3si9n:
Which version are you using? LSP is only available in sbt 1.1.0-M1.

wilmer-daza:
You mean the sbt version within my project/build.properties or the sbt version within my environment?

dwijnand:
The sbt version in project/build.properties.

wilmer-daza:
I just changed my sbt version to 1.1.0-M1 and still I got the same error (off course I did execute "sbt compile" on my project)


Error: ENOENT: no such file or directory, open '/home/wilmer/test/scala-test/project/target/active.json'
    at Object.fs.openSync (fs.js:584:18)
    at Object.module.(anonymous function) [as openSync] (ELECTRON_ASAR.js:173:20)
    at Object.fs.readFileSync (fs.js:491:33)
    at Object.fs.readFileSync (ELECTRON_ASAR.js:505:29)
[...]

PR #4130, opened 2018-04-29
start sbt in VS Code terminal window

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/3624

Previously VS Code extension for sbt required sbt session to be running already. I got a patch from @olofwalker over Slack that opens sbt in VS Code's terminal window, which is more natural.
It seems to work when I tested it locally.


eed3si9n:
We can tweak this further, but I am inclined to merge this and get a new version of VS Code extension out, partially because https://github.com/sbt/sbt/pull/4128 contains vulnerability fix.

PR #4132, opened 2018-04-29
Avoid postfix op without feature import

som-snytt:

### DIFF
--- a/compile/interface/src/main/scala/xsbt/Compat.scala
+++ b/compile/interface/src/main/scala/xsbt/Compat.scala
@@ -149,7 +149,7 @@ abstract class Compat {
           import analyzer._ // this is where MEA lives in 2.11.x
           tree.attachments.all.collect {
             case att: MacroExpansionAttachment => att.expandee
-          } headOption
+          }.headOption
         }
       }
     }


eed3si9n (approved):
LGTM

dwijnand (approved):
Thanks @som-snytt.

SethTisue:
as @OlegYch has noted, this makes sbt 0.13.18 work with Scala 2.13.0-RC1, it wouldn't otherwise

som-snytt:
I'm watching Jacob Bronowski's Ascent of Man with the daughter, the episodes on alchemy, Pythagoras, and Galileo.

It's one part arcana, one part discovery. I just said, you learn the Pythagorean theorem in sixth grade, but this is its significance.

I want to live my life now only in pursuit of significance, or `signifiance`. I had another moment of understanding at the keyboard with Bach. I'm no piano prodigy, does that mean I should be cut off from this experience? I hesitate to call it knowledge.

Also our experience with computer science should be a kind of personal discovery.

I don't know yet how to introduce the daughter to that.

eed3si9n:
In the movie Predestination, temporal agents travel back in time to prevent an attack in 1975 by an elusive criminal "Fizzle Bomber". I wonder if there's a term for predicting a problem and fixing it before most people notice the effect.

som-snytt:
OK, I just got back from dropping her at "art class", a Chinese class where they teach manga and similar sorts of illustration, including game art.

I was going to delete my previous comment. The fact is, I was upset because she left to use the bathroom for a half-hour. I was upset because at 3 a.m., I needed to use said facility which was clogged by one of her famously clogging poops. There is only one such facility in the apartment. There was no alternative but to fix it. Also, someone had thrown out the poop stick, the chop stick intended to break up the obstruction.

I am going to procure a new one and call it the "Fizzle Stick".

PR #4135, opened 2018-05-01
Fix VS Code extension startup

eed3si9n:
The change contributed in https://github.com/sbt/sbt/pull/4130 to start sbt within VS Code looked ok at first, but when I published the extension and started using it, I realized that it's a bit broken.
Basically there's no cleanup logic (that I could find), so simply closing VS Code would leave `project/target/active.json` behind, which lets VS Code extension to attempt connection, and it fails because there's no server running. To workaround this, I'll attempt to connect to the socket to confirm sbt server is up.

/cc @olofwalker

### DIFF
--- a/vscode-sbt-scala/client/package.json
+++ b/vscode-sbt-scala/client/package.json
@@ -1,7 +1,7 @@
 {
   "name": "vscode-sbt-scala",
   "displayName": "Scala (sbt)",
-  "version": "0.1.0",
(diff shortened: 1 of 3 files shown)


olofwalker:
Ok, this looks more solid, even if we did clean up `active.json` when closing, there are edge cases when this would not happen anyway (Extension host crashing for example)

PR #4136, opened 2018-05-01
Discovery of java home directories

2m:

This PR proposes a discovery for Java home directories implemented by scanning well known system directories for available java installations. Currently this PR only addresses Linux and MacOS. // @cunei 

This PR is a first step towards cross building across different Java versions.

Discovering Java home directories is also useful when a project needs to locate `rt.jar` of a different java installation ([for example](https://github.com/akka/akka/commit/0bc84f642622d90357ecfde5c4a2b463c53b5862#diff-4e8397c4feb85d6565399bcfe99c701aR98) when working around `-release 8` limitaions in JDK9)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1625,6 +1634,28 @@ object Defaults extends BuildCommon {
           else old
         }
       ))
+
+  private case class JavaDiscoverConf(root: String, dirRexp: Regex, innerDir: String)
+  private final val JavaDiscoverConfigs = Seq(
+    JavaDiscoverConf("/usr/lib/jvm", "java-([0-9]+)-.*".r, ""), // most linux
+    JavaDiscoverConf("/Library/Java/JavaVirtualMachines",
+                     "jdk[1\\.]*([0-9]+)\\..*".r,
```
dwijnand, line 1642: Doesn't this match, for instance, `/Library/Java/JavaVirtualMachines/jdk-10.jdk/`?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1625,6 +1634,28 @@ object Defaults extends BuildCommon {
...
+    JavaDiscoverConfigs
+      .flatMap {
+        case JavaDiscoverConf(root, dirRexp, inner) =>
+          Option(new File(root).list())
+            .getOrElse(Array.empty[String])
+            .toSeq
+            .collect {
+              case dir @ dirRexp(ver) => ver -> Paths.get(root, dir, inner).toFile
+            }
+      }(breakOut)
```
dwijnand, line 1657: With `breakOut` making an exit in Scala 2.13, best implement this with `.iterator` (after `JavaDiscoverConfigs`) and a final `.toMap`

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1625,6 +1634,28 @@ object Defaults extends BuildCommon {
...
+  )
+
+  private[sbt] def discoverJavaHomes: Initialize[Map[String, File]] = Def.setting {
+    import scala.collection.breakOut
+    JavaDiscoverConfigs
+      .flatMap {
+        case JavaDiscoverConf(root, dirRexp, inner) =>
+          Option(new File(root).list())
+            .getOrElse(Array.empty[String])
+            .toSeq
```
dwijnand, line 1653: Needn't be a seq, it can just be an iterator, with `.iterator`

main/src/main/scala/sbt/Keys.scala
```diff
@@ -268,6 +268,8 @@ object Keys {
   val outputStrategy = settingKey[Option[sbt.OutputStrategy]]("Selects how to log output when running a main class.").withRank(DSetting)
   val connectInput = settingKey[Boolean]("If true, connects standard input when running a main class forked.").withRank(CSetting)
   val javaHome = settingKey[Option[File]]("Selects the Java installation used for compiling and forking.  If None, uses the Java installation running the build.").withRank(ASetting)
+  val discoveredJavaHomes = settingKey[Map[String, File]]("Discovered Java home directories")
+  val additionalJavaHomes = settingKey[Map[String, File]]("Additional Java home directories")
```
dwijnand, line 272: Maybe call this just `javaHomes`, like `resolvers`?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -154,6 +161,8 @@ object Defaults extends BuildCommon {
       scalaHome :== None,
       apiURL := None,
       javaHome :== None,
+      discoveredJavaHomes := discoverJavaHomes.value ++ additionalJavaHomes.value,
```
dwijnand, line 164: This definition isn't intuitive to me. Discovered should only be discovered.

Maybe there should be a `fullJavaHomes` (like `fullResolvers` and `fullClasspath`) with this RHS.

dwijnand (commented):
Cool. Are these the initial pieces of your hackathon in Lisbon?

I left a few comments. Some less important than others.. 🙂

main/src/main/scala/sbt/Defaults.scala
2m: Good catch. I'll change it to `jdk(1\.|-)?([0-9]+)\..*` and extract the second group.

main/src/main/scala/sbt/Defaults.scala
2m: Sounds good. Then `fullJavaHomes = discoveredJavaHomes + javaHomes`.

2m:
Yes, this is from Lisbon and is a prequel to cross java home building. However it looks like this might be useful on its own as well.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: My vote is `fullJavaHomes` = `discoveredJavaHomes` + `unmanagedJavaHomes` as we had in Lisbon.

The reason is because the build user might want to either:
1. append to the default list (like `resolvers`) or
2. take over the entire mapping.

Note that for `resolvers`, we offer an entirely different mechanism `~/.sbt/repositories`.

dwijnand: Why `unmanagedJavaHomes` over just `javaHomes`?

And if you prefer `unmanagedJavaHomes`, why `discoveredJavaHomes` instead of `managedJavaHomes` then?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1625,6 +1634,28 @@ object Defaults extends BuildCommon {
           else old
         }
       ))
+
+  private case class JavaDiscoverConf(root: String, dirRexp: Regex, innerDir: String)
```
eed3si9n, line 1638: These should move to `CrossJava.scala`, I think.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Because `javaHomes` is not clear if it's the unmanaged part of the full.

```scala
javaHomes := Map("10" -> something)
```

I am ok with either `discoveredJavaHomes` or `managedJavaHomes` since the build user won't be interacting with them that much.

dwijnand: > Because `javaHomes` is not clear if it's the unmanaged part of the full.

It's the shorter, user-facing one, like I believe `resolvers` was designed to be.

eed3si9n: With `resolvers` the build users are somewhat aware that sbt will hit Maven Central by default and that you are supplying additional not-so-common custom resolvers for your subproject. In the case of `javaHomes`, if you _had_ to change it, that means sbt failed to detect some JDK homes automatically, and you are augmenting it with your machine-dependent info. IMO it should actually be some longer "it only works on your machine" sounding name. If not `unmanagedJavaHomes` it might be `myLittleJavaHomes`.

dwijnand: > if you _had_ to change it, that means sbt failed to detect some JDK homes automatically, and you are augmenting it with your machine-dependent info.

I see it as the same as `resolvers`. You're adding not-so-common custom java homes paths.

Also, like `resolvers`, it could have "doesn't work for everyone" concerns (e.g company-specific artefact repository URLs and company-specific JVM installation paths).

eed3si9n: `resolvers` typically go into `build.sbt` whereas your `Global / (unmanaged)JavaHomes` should go into user-defined global settings; never in `build.sbt`.

The users almost never have to debug `fullResolvers` whereas if JDK cross building doesn't work, the users will need to type `show Global/fullJavaHomes` to debug why it wasn't detected. This is analogous to managed vs unmanaged classpath or source. If there's a key named `javaHomes` already, wouldn't that be more confusing?

If we really want to use the name `javaHomes`, I would say we should get rid of `fullJavaHomes`, and just use `javaHomes` key where the user can just `+=` into it.

dwijnand: > `resolvers` typically go into build.sbt whereas your `Global / (unmanaged)JavaHomes` should go into user-defined global settings; never in `build.sbt`.

I can see both. For instance if you're in a company that uses a custom JDK/JVM installed (e.g at Twitter with GraalVM) in a specific but not-so-common path I can see it being experimented with at a project level, maybe eventually being defined in a company sbt plugin. Not necessarily in user-wide global settings.

And even if it were meant for user-wide global settings, why shouldn't it still be called `javaHomes`?

> whereas if JDK cross building doesn't work, the users will need to type `show Global/fullJavaHomes` to debug why it wasn't detected.

To debug what was discovered they would look at `discoveredJavaHomes`.

> If we really want to use the name `javaHomes`, I would say we should get rid of `fullJavaHomes`, and just use `javaHomes` key where the user can just `+=` into it.

I don't follow, but I think it's non sequitur from my earlier comments.

eed3si9n: things_and_cothings

eed3si9n: We discussed in sbt 1 meeting today. I'll take this PR and make the edits.

2m:
Cntd. in #4139

PR #4137, opened 2018-05-02
Uncomment server customization in test

eed3si9n:


dwijnand:
Spurious test failure. Reported in https://github.com/sbt/sbt/issues/4138, and job restarted.

PR #4139, opened 2018-05-03
Cross JDK forking

eed3si9n:
This is a refactoring/continuation of https://github.com/sbt/sbt/pull/4136 by @2m reflecting review comments by @dwijnand 

```
sbt:helloworld> java++ 10
[info] Reapplying settings...
sbt:helloworld> run
[info] Running (fork) Hello
[info] 10.0.1

sbt:helloworld> java++ 8
[info] Reapplying settings...

sbt:helloworld> run
[info] Running (fork) Hello
[info] 1.8.0_171
```

### original description

This PR proposes a discovery for Java home directories implemented by scanning well known system directories for available java installations. Currently this PR only addresses Linux and MacOS. // @cunei 

This PR is a first step towards cross building across different Java versions.

Discovering Java home directories is also useful when a project needs to locate `rt.jar` of a different java installation ([for example](https://github.com/akka/akka/commit/0bc84f642622d90357ecfde5c4a2b463c53b5862#diff-4e8397c4feb85d6565399bcfe99c701aR98) when working around `-release 8` limitaions in JDK9)


main/src/main/contraband/main.contra
```diff
@@ -17,3 +17,9 @@ enum PluginTrigger {
   AllRequirements
   NoTrigger
 }
+
+type JavaVersion {
```
eed3si9n, line 21: Added pseudo case class for JavaVersion, instead of using String.

main/src/main/scala/sbt/Keys.scala
```diff
@@ -268,6 +268,10 @@ object Keys {
   val outputStrategy = settingKey[Option[sbt.OutputStrategy]]("Selects how to log output when running a main class.").withRank(DSetting)
   val connectInput = settingKey[Boolean]("If true, connects standard input when running a main class forked.").withRank(CSetting)
   val javaHome = settingKey[Option[File]]("Selects the Java installation used for compiling and forking.  If None, uses the Java installation running the build.").withRank(ASetting)
+  val discoveredJavaHomes = settingKey[Map[JavaVersion, File]]("Discovered Java home directories")
+  val javaHomes = settingKey[Map[JavaVersion, File]]("The user-defined additional Java home directories")
```
eed3si9n: Custom key is named `javaHomes`.

main/src/main/scala/sbt/Keys.scala
```diff
@@ -268,6 +268,10 @@ object Keys {
...
+  val fullJavaHomes = taskKey[Map[JavaVersion, File]]("Combines discoveredJavaHomes and custom javaHomes.").withRank(CTask)
```
eed3si9n: `fullJavaHomes` is a task if someone wants to side effect. (Credit to Dale)

dwijnand:
I've a yak to shave about `JavaVersion`.. 😕

Short-term solution: can we make it an `sbt.internal` type, so we can iterate on it?

Long-term solution: one idea that I was thinking was to experiment modelling "java version" using a sum type to give a selection like jenv, where when I added Oracle JDK 1.8.0.144 I got the following selection of identifiers to choose from:

* 1.8
* 1.8.0.144
* oracle64-1.8.0.144

So the usual pitfalls need to be avoids:

* can't make `JavaVersion` a sealed trait -> means we can't add new types
* can't give `JavaVersion` or its subtypes `unapply`

So perhaps a starting point would be to make `JavaVersion` a public `package sbt` type, add different constructors on `JavaVersion` (maybe `apply` takes "1.8", `update` takes "1.8.0.144", `full` takes "oracle64-1.8.0.144") that just return opaque `JavaVersion`. But that's the creation side, I'm unsure about at the usage side. 🤔

dwijnand:
I'm overthinking it. I'm designing for use cases I don't quite have clear.

Which makes me realise: what are the motivating use cases we're looking to solve with this change? Can we explore that space before committing to a stable public API?

/cc @2m who I assume can provide the Akka use cases

main/src/main/scala/sbt/Keys.scala
eed3si9n: I had to switch this back to a setting since carrying the state around aggregation is pain in the ass when implementing a command.

eed3si9n:
Implemented `java++ 10`

eed3si9n:
@dwijnand We can iterate by changing the implementation of `JavaVersion.apply(String)`.

dwijnand:
Can we remove its other companion object applies and methods from the public API too?

eed3si9n:
I think what you're talking about can be achieved by making bunch of duplicate entries in discoveredJavaHomes like e6fa634 or jEnv without being too paranoid about JavaVersion.

dwijnand:
Can we discuss the use cases before committing to new public API?

dwijnand:
> Can we discuss the use cases before committing to new public API?

Or, let's just make `JavaVersion` `sbt.internal` and cross that bridge another day.

2m:
We have a couple of use-cases in Akka:

1. We would use the discovery of JavaHomes [to locate](https://github.com/akka/akka/commit/0bc84f642622d90357ecfde5c4a2b463c53b5862#diff-4e8397c4feb85d6565399bcfe99c701aR98) the `rt.jar` of an older version. Currently the `-release` flag is broken in JDK9 where it does not allow acceess to `Unsafe` when compiling with JDK9 and with `-release 8` flag. The workaround is to compile with `"-source", "8", "-target", "8", "-bootclasspath", "/usr/lib/jvm/java-8-openjdk/jre/lib/rt.jar"` javacFlags for which we need JavaHome discovery.

2. Cross testing with JDK8. Since we build with JDK9, we would like to test the compiled classes running Java 8 to be sure that we did not do something that is JDK8 incompatible.

eed3si9n:
Given the uses cases, the minor version and vendor handling of JDK variety should be considered out of scope of this PR.

Also given that `JavaVersion("10") -> file(...)` needs to go into build.sbt, we do need to keep bincompat, and I see no reason to put that into internal.

sbt/src/sbt-test/java/cross/test
```diff
@@ -0,0 +1,6 @@
+> java+ run
+> check 8
+
+> java++ 10!
+> run
+> check 10
```
eed3si9n, line 6: Cross JDK testing using jabba installed openjdk 10.

dwijnand:
> Also given that `JavaVersion("10") -> file(...)` needs to go into build.sbt, we do need to keep bincompat, and I see no reason to put that into internal.

We don't need to keep bincompat if we declare the whole feature experimental, which is what I'm proposing.

eed3si9n:
I was thinking how we would document this feature, and realized that we shouldn't expose `JavaVersion` to `javaHomes` key because of the compatibility with sbt 1.0, assuming we want to support `javaHomes` in user-defined global settings.

1. Keys should use `String`, but
2. on the other hand, `JavaVersion` can now be internal

2m (approved):
LGTM. Looking good. The cross building functionality is getting more and more ripe for factoring out to a more general behaviour.

main/src/main/scala/sbt/internal/CrossJava.scala
```diff
@@ -0,0 +1,365 @@
...
+  }
+
+  sealed trait JavaDiscoverConf {
+    def javaHomes: Vector[(String, File)]
+  }
+
+  object JavaDiscoverConfig {
+    val linux = new JavaDiscoverConf {
+      val base: File = file("/usr") / "lib" / "jvm"
+      val JavaHomeDir = """java-([0-9]+)-.*""".r
```
chbatey, line 314: FYI this regex won't match on a default Fedora Open JDK install. See https://github.com/akka/akka/issues/25298

PR #4145, opened 2018-05-05
IO 1.1.7, Zinc 1.1.6, dotty plugin, and watch

eed3si9n:


sbt/src/sbt-test/compiler-project/dotty-compiler-plugin/build.sbt
```diff
@@ -1,19 +1,16 @@
-lazy val dottyVersion = dottyLatestNightlyBuild
+// hardcode dottyVersion to make test deterministic
+lazy val dottyVersion = "0.8.0-bin-20180424-e77604d-NIGHTLY"
```
eed3si9n: @liufengyun I am hardcoding the version to avoid breakage caused by Dotty changes.

sbt/src/sbt-test/compiler-project/dotty-compiler-plugin/build.sbt
```diff
@@ -1,19 +1,16 @@
...
 
-lazy val pluginSetting = Seq(
-  name := "dividezero",
-  version := "0.0.1",
-  organization := "ch.epfl.lamp",
-  scalaVersion := dottyVersion,
-
-  libraryDependencies ++= Seq(
-    "ch.epfl.lamp" %% "dotty" % "provided"
```
eed3si9n: @liufengyun I wasn't sure what this is supposed to do. I don't think it's needed since setting `scalaVersion` should grab Dotty  for you.

sbt/src/sbt-test/compiler-project/dotty-compiler-plugin/build.sbt
liufengyun: It makes sense to me, thanks for the improvement.

sbt/src/sbt-test/compiler-project/dotty-compiler-plugin/build.sbt
liufengyun: I thought by default the compiler is not in compilation classpath, thanks for the fix 👍

PR #4163, opened 2018-05-15
Remove compiler warnings introduced with Scala 2.12

dwestheide:
- [x ] I've read the [CONTRIBUTING](https://github.com/sbt/sbt/blob/1.x/CONTRIBUTING.md) guidelines

 This PR removes some of the compiler warnings introduced with Scala 2.12. Specifically,  it removes warnings related to discarded non-unit values and unused parameters. 

Please note that it does _not_ remove _all_ such warnings, because a lot of them are coming from generated code or are part of SBT's public API. Removing those will have to be done in separate PRs and certainly not against the `1.1.x` branch.

### DIFF
--- a/core-macros/src/main/scala/sbt/internal/util/appmacro/ContextUtil.scala
+++ b/core-macros/src/main/scala/sbt/internal/util/appmacro/ContextUtil.scala
@@ -33,9 +33,9 @@ object ContextUtil {
       f: (c.Expr[Any], c.Position) => c.Expr[T]): c.Expr[T] = {
     import c.universe._
     c.macroApplication match {
-      case s @ Select(Apply(_, t :: Nil), tp) => f(c.Expr[Any](t), s.pos)
-      case a @ Apply(_, t :: Nil)             => f(c.Expr[Any](t), a.pos)
-      case x                                  => unexpectedTree(x)
+      case s @ Select(Apply(_, t :: Nil), _) => f(c.Expr[Any](t), s.pos)
+      case a @ Apply(_, t :: Nil)            => f(c.Expr[Any](t), a.pos)
+      case x                                 => unexpectedTree(x)
     }
   }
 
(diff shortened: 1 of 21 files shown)


core-macros/src/main/scala/sbt/internal/util/appmacro/ContextUtil.scala
```diff
@@ -33,9 +33,9 @@ object ContextUtil {
       f: (c.Expr[Any], c.Position) => c.Expr[T]): c.Expr[T] = {
     import c.universe._
     c.macroApplication match {
-      case s @ Select(Apply(_, t :: Nil), tp) => f(c.Expr[Any](t), s.pos)
```
eed3si9n, line 36: 👍

internal/util-collection/src/main/scala/sbt/internal/util/KList.scala
```diff
@@ -18,7 +18,10 @@ sealed trait KList[+M[_]] {
   def transform[N[_]](f: M ~> N): Transform[N]
 
   /** Folds this list using a function that operates on the homogeneous type of the elements of this list. */
-  def foldr[B](f: (M[_], B) => B, init: B): B = init // had trouble defining it in KNil
+  def foldr[B](f: (M[_], B) => B, init: B): B = this match {
```
eed3si9n, line 21: 👍

dwijnand:
:-/ I'd fixed these in https://github.com/sbt/sbt/pull/3807 on the 1.x branch.

This kind of confirms to me we should probably bias/favour contributors work on the 1.x branch, where almost all improvements land.

(I'll review this soon (tm))

eed3si9n (approved):
LGTM
Thanks @dwestheide

dwijnand:
The reason I I fixed this in the 1.x branch is because they're not "hotfix" changes.

I think we should close this and I would takeaway from this PR that we should bias to pointing new contributors to the 1.x branch.

Let's discuss this at the sbt meeting today (@dwestheide feel free to ping me if you'd like to join the chitchat 😄 http://everytimezone.com/#2018-5-23,180,b8jj)

dwestheide:
That's fine with me. :) Unfortunately, I won't have time to join the meeting.

dwijnand:
OK, thanks Daniel.

PR #4159, opened 2018-05-15
Migrated last command strings to camelCase

Asamsig:
Fixes issues #4081

Issue #4081, opened 2018-04-08
Migrate all kebab-case commands to camelCase


eed3si9n:
This is a generalized issue for https://github.com/sbt/sbt/issues/4072

While #4072 talks about `lastGrep`, looking at https://github.com/sbt/sbt/blob/1.x/main/src/main/scala/sbt/internal/CommandStrings.scala, there are some more kebab-cased command names. It would be good to migrate them for the sake of consistency.

Asamsig:
Looking into this!

dwijnand:
Looks like there are only 3 left:

```
11:00:02 $ rg '"[a-z]+-[a-z]+"' -- main/src/main/scala/sbt/internal/CommandStrings.scala
60:  val OldLastGrepCommand = "last-grep"
289:  def LoadCommand = "load-commands"
292:  def LoadFailed = "load-failed"
```

eed3si9n:
See also https://github.com/sbt/sbt/blob/v1.1.5/main-command/src/main/scala/sbt/BasicCommandStrings.scala#L188-L198

dwijnand:
Fixed by #4159.

naferx:
Hi @dwijnand , I know it is closed but just for curiosity 

I typed
```
sbt:test> load-commands
[error] Not a valid command: load-commands (similar: load-failed)
[error] Not a valid project ID: load-commands
[error] Expected ':'
[error] Not a valid key: load-commands (similar: commands, cleanupCommands, initialCommands)
[error] load-commands
[error]              ^
```
Is it used internally ?

dwijnand:
@naferx In working with @Asamsig on #4159 we found out that those commands had been removed a long time ago. Which is why in #4159 we just removed the strings.

naferx:
ah, good to know.
Thanks Dale

PR #4169, opened 2018-05-19
Fix kebab-case commands: notify-users-about-shell, write-sbt-version

tiqwab:
The below two commands are also targets of #4081? Applied same changes as #4159.

- `write-sbt-version`
- `notify-users-about-shell`


dwijnand (approved):
Nice find! Thank you.

PR #4194, opened 2018-06-12
Fix for #4191 (active.json should be removed on JVM shutdown)

### LINKED ISSUE #4191: active.json should be removed on JVM shutdown
## steps

```
$ sbt
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=384m; support was removed in 8.0
[info] Loading settings from idea.sbt,global-plugins.sbt ...
[info] Loading global plugins from /Users/dnw/.dotfiles/.sbt/1.0/plugins
[info] Loading settings from plugins.sbt ...
[info] Loading project definition from /d/sbt/project
[info] Loading settings from build.sbt ...
[info] Resolving key references (15599 settings) ...
[info] Set current project to sbtRoot (in build file:/d/sbt/)
[info] sbt server started at local:///Users/dnw/.sbt/1.0/server/9bea40e9748dc34f2c46/sock
> compile
[info] Updating bundledLauncherProj...
[info] Updating ...
^C
```

## problem

```
! ll project/target/
project/target/
├── [drwxr-xr-x dnw      staff     256 Mar 16 17:38]  config-classes/
├── [drwxr-xr-x dnw      staff      96 Jan 24 14:04]  scala-2.12/
├── [drwxr-xr-x dnw      staff     160 Jan 24 14:04]  streams/
└── [-rw-r--r-- dnw      staff      70 Mar 19 11:13]  active.json

3 directories, 1 file
```

## expectation

No `active.json` present after JVM shutdown.

## notes

sbt version: 1.1.1

This is a generalisation of https://github.com/sbt/sbt/issues/3862, which was about the file being removed when `exit` was executed.

# Mentoring Instructions

## Links

The port file is created on this line:

https://github.com/sbt/sbt/blob/1d7b35bd67722511b9a058113e7a1070984e6758/main-command/src/main/scala/sbt/internal/server/Server.scala#L194

and deleted on `ServerInstance` `shutdown`:

https://github.com/sbt/sbt/blob/1d7b35bd67722511b9a058113e7a1070984e6758/main-command/src/main/scala/sbt/internal/server/Server.scala#L138

However nothing guarantees that is called on JVM shutdown.

## Solution

One possible solution is to add an exit hook to the state:

https://github.com/sbt/sbt/blob/1d7b35bd67722511b9a058113e7a1070984e6758/main-command/src/main/scala/sbt/State.scala#L39

However that might not be possible or be sufficient, so perhaps adding a JVM runtime shutdown hook like:

https://github.com/sbt/sbt/blob/1d7b35bd67722511b9a058113e7a1070984e6758/main/src/main/scala/sbt/MainLoop.scala#L34

## Testing

Because this is about JVM shutdown, I think this is going to require locally publishing a custom build of sbt.

See https://github.com/sbt/sbt/blob/1d7b35bd67722511b9a058113e7a1070984e6758/CONTRIBUTING.md#instruction-to-build-just-sbt

## Motivation

The motivation here is sbt doesn't start up properly if the file is present, and also it's a marker file used by sbt clients to connect to the server.

mrveera:
Fixes #4191

    Added a shutdown hook to clean up active.json file

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -105,11 +105,18 @@ object StandardMain {
   import scalacache.caffeine._
   private[sbt] lazy val cache: Cache[Any] = CaffeineCache[Any]
 
+  private[sbt] val shutdownHook = new Thread(new Runnable {
+    def run(): Unit = {
+      exchange.shutdown
+    }
+  })
+
   def runManaged(s: State): xsbti.MainResult = {
     val previous = TrapExit.installManager()
     try {
       try {
         try {
+          Runtime.getRuntime.addShutdownHook(shutdownHook)
           MainLoop.runLogged(s)
         } finally exchange.shutdown
       } finally DefaultBackgroundJobService.backgroundJobService.shutdown()


eed3si9n:
@veera83372 Thanks for the contribution!

I've confirmed the behavior locally.


### before

```
sbt:Hello> % # I am typing Ctrl+C
[Ga:/tmp/foobuild/scala-seed-project]$ ls project/target/active.json
project/target/active.json
```

### after

```
sbt:Hello> %
[Ga:/tmp/foobuild/scala-seed-project]$ ls project/target/active.json
ls: project/target/active.json: No such file or directory
```

and I've also confirmed that this works when I close VS Code too.

eed3si9n (approved):

PR #4200, opened 2018-06-13
projectMatrix (spatial representation of cross building)

eed3si9n:
This implements basic `projectMatrix`.
See https://discuss.lightbend.com/t/spatial-representation-of-cross-building/1277

```scala
lazy val core = (projectMatrix in file("core"))
  .scalaVersions("2.12.6", "2.11.12")
  .settings(
    name := "core"
  )
  .jvmPlatform()

lazy val app = (projectMatrix in file("app"))
  .dependsOn(core)
  .scalaVersions("2.12.6", "2.11.12")
  .settings(
    name := "app"
  )
  .jvmPlatform()
```

The above will automatically generate four subprojects `coreJVM2_12`, `coreJVM2_11`, `appJVM2_12`, and `appJVM2_11` with appJVM* depending on coreJVM*.

### DIFF
--- a/main/src/main/scala/sbt/Project.scala
+++ b/main/src/main/scala/sbt/Project.scala
@@ -10,6 +10,7 @@ package sbt
 import java.io.File
 import java.net.URI
 import java.util.Locale
+import scala.collection.immutable.ListMap
 import Project._
 import BasicKeys.serverLogLevel
 import Keys.{
@@ -29,7 +30,12 @@ import Keys.{
   serverConnectionType,
   fullServerHandlers,
   logLevel,
-  watch
+  watch,
+  unmanagedSourceDirectories,
+  crossTarget,
+  sourceDirectory,
+  scalaSource,
+  target,
 }
 import Scope.{ Global, ThisScope }
 import Def.{ Flattened, Initialize, ScopedKey, Setting, SettingsDefinition }
@@ -47,6 +53,8 @@ import sbt.internal.util.Types.{ const, idFun }
 import sbt.internal.util.complete.DefaultParsers
 import sbt.internal.server.ServerHandler
 import sbt.librarymanagement.Configuration
+import sbt.librarymanagement.Configurations.{ Compile, Test }
+import sbt.librarymanagement.CrossVersion.partialVersion
 import sbt.util.{ Show, Level }
 import sjsonnew.JsonFormat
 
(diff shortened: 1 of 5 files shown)


main/src/main/scala/sbt/Project.scala
```diff
@@ -157,6 +165,288 @@ private[sbt] object CompositeProject {
...
+   * Sets the [[AutoPlugin]]s of this project.
+   * A [[AutoPlugin]] is a common label that is used by plugins to determine what settings, if any, to enable on a project.
+   */
+  def enablePlugins(ns: Plugins*): ProjectMatrix
+
+  /** Disable the given plugins on this project. */
+  def disablePlugins(ps: AutoPlugin*): ProjectMatrix
+
+  def custom(
+      idPostfix: String,
```
nafg, line 220: I think the word is suffix, not postfix? I haven't checked a dictionary but postfix is usually used for postfix operators. Appending to a string or word is usually called suffix.

nafg (changes requested):

main/src/main/scala/sbt/Project.scala
eed3si9n: Yes. You're totally right.

sjrd:
Is there a reason for this to start its life *within* sbt? It seems like it could be released as an independent sbt plugin first.

Advantage: the design can be refined in a couple iterations that would be backward incompatible. If you put it right away in sbt, you're stuck with your initial design forever. Even years after introducing `crossProject` in Scala.js, we are still refining its design with breaking changes in sbt-crossproject (see for example [*yesterday's* release](https://github.com/portable-scala/sbt-crossproject/releases/tag/v0.5.0).

eed3si9n:
> Is there a reason for this to start its life within sbt? It seems like it could be released as an independent sbt plugin first.

- For one, I think the usefulness of generalizing cross building has been shown already by  sbt-crossproject and less well-known sbt-cross, so I see this as an incorporation of the concept.
- Also `CompositeProject` (https://github.com/sbt/sbt/pull/4056) already went in, so I thought it makes sense to provide a reference implementation that solves at least one problem. The analogous situation would be the templateResolver + Giter8.
- Related, if it's part of sbt it will be announced together when 1.2 comes out, and get documented. But I guess I could do that with plugin too.

The ability to iterate is an advantage, so I am open to turning this into a plugin.

sjrd:
> For one, I think the usefulness of generalizing cross building has been shown already

I am most definitely not questioning the usefulness. Many sbt plugins are very useful.

> Also CompositeProject (#4056) already went in,

`CompositeProject` *had* to go *in*, because the feature it provides cannot be implemented in a plugin. It that had been possible, I would have implemented it years ago.

In Scala.js, we have a strict policy that "everything that *can* be outside of the core repo *must* be outside of the core repo". This significantly improves our ability to innovate in separate repos while keeping the core super stable.

IMHO sbt could use a similar policy.

eed3si9n:
ok. This is now split into a plugin https://github.com/sbt/sbt-projectmatrix

Issue #1502, opened 2014-08-06
Add plugins via a command


jsuereth:
It would be ideal, in many scenarios, to be able to add a plugin to the build via a command.

E.g. automated tooling sometimes has customization hooks that need to be added to a build.  Currently, these can use the `apply -cp <classpath> <command class>` mechanism to modify a build and add their hooks, but this leads to some less-than-ideal issues, e.g. classpath detection and cross-publishing.

Ideally, tools/users should be able to experiment with a plugin via a command like:

```
> apply-plugin "com.typesafe.sbt" % "sbt-pgp" % "1.5"
```

and have this automatically reload the build with the plugin enabled.

reference: http://stackoverflow.com/questions/25137042/how-do-i-add-a-sbt-plugin-in-to-an-sbt-session-ie-without-saving?noredirect=1#comment39130312_25137042

dwijnand:
How viable is this in nowadays sbt (AutoPlugins, sbt-server etc)? I'd very much like this but I wanted to looking for insight on it beforehand. Thanks.

jastice:
I am very interested in improving sbt integration with the IntelliJ Scala plugin for sbt v1.0. Being able to add an sbt plugin on boot or by command would make it somewhat easier. I have a vague idea how plugins are loaded, but some more concrete pointers on how to implement a command for it would be appreciated.

I do have a workaround in mind: create a custom sbt global dir with our own plugin added, and munge the user's stuff into it. Would that cause any weird issues?

jastice:
Note: For our current IDE integration purposes, it would be even sufficient to just be able to add via environment variable additional "global" directories that get loaded into the build.
Loading by command would still be neat for other purposes.

jastice:
any insights on how to accomplish this? for practical purposes #3443 is caused by not having a robust way to inject plugins when starting sbt. I am able to spend some time on this, given a little guidance.

dwijnand:
Here's an old commit of mine when I tried to accomplish this: https://github.com/sbt/sbt/commit/81a4d07b1e1657d3fb5ac23b52c63fb91bf37547

I can't even remember how well that worked or not.

jastice:
Thanks, Dale. That would work even without modifying sbt itself since it's just a sequence of commands, but the `session save` part modifies the user's build, which is something I definitely want to avoid. What I would need is a way to make a plugin sticky for the lifetime of the sbt process, but not have any effects to other instances of sbt started during or after the process runs.

[...]

PR #4211, opened 2018-06-18
addPluginSbtFile command

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/1502

This adds `--addPluginSbtFile=<file>` command, which adds the given .sbt file to the plugin build.
Using this mechanism editors or IDEs can start a build with required plugin.

```
$ cat /tmp/extra.sbt
addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.7")

$ sbt --addPluginSbtFile=/tmp/extra.sbt
...
sbt:helloworld> plugins
In file:/xxxx/hellotest/
  ...
  sbtassembly.AssemblyPlugin: enabled in root
```

/cc @jastice

### DIFF
--- a/main-command/src/main/scala/sbt/BasicCommandStrings.scala
+++ b/main-command/src/main/scala/sbt/BasicCommandStrings.scala
@@ -83,7 +83,8 @@ $HelpCommand <regular expression>
       List("-" + elem, "--" + elem)
     }
     (s.startsWith(EarlyCommand + "(") && s.endsWith(")")) ||
-    (levelOptions contains s)
+    (levelOptions contains s) ||
+    (s.startsWith("-" + AddPluginSbtFileCommand) || s.startsWith("--" + AddPluginSbtFileCommand))
   }
 
   val EarlyCommand = "early"
@@ -95,6 +96,14 @@ $HelpCommand <regular expression>
 	Schedules an early command, which will be run before other commands on the command line.
 	The order is preserved between all early commands, so `sbt "early(a)" "early(b)"` executes `a` and `b` in order.
 """
+
+  def addPluginSbtFileHelp = {
+    val brief =
+      (s"--$AddPluginSbtFileCommand=<file>", "Adds the given *.sbt file to the plugin build.")
+    Help(brief)
+  }
+
+  val AddPluginSbtFileCommand = "addPluginSbtFile"
 
   def ReadCommand = "<"
   def ReadFiles = " file1 file2 ..."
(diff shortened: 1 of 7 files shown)


smarter:
@Duhemm Looks like this could replace https://github.com/scalacenter/load-plugin for the IDE usecase.

eed3si9n:
`--addPluginSbtFile=<file>` gets to the `State` before project loading (`reload`), so it by design does not require extra reloading, and also survives `reload` command issued by the build user. 

```
sbt:helloworld> reload
[info] Loading settings for project global-plugins from vimquit.sbt,pgp.sbt ...
[info] Loading settings for project global-plugins from extra.sbt ...
```

jastice:
Nice!

nafg:
Nice. Can it be set in `.sbtopts`?

main-command/src/main/scala/sbt/BasicCommandStrings.scala
```diff
@@ -96,6 +97,14 @@ $HelpCommand <regular expression>
 	The order is preserved between all early commands, so `sbt "early(a)" "early(b)"` executes `a` and `b` in order.
 """
 
+  def addPluginSbtFileHelp = {
+    val brief =
+      (s"--$AddPluginSbtFileCommand=<file>", "Adds the given *.sbt file to the plugin build.")
```
nafg, line 102: aren't most options with a single dash (at least in the help)?

eed3si9n: The single line convention comes from Java, and since Java 9 is updating to more UNIX `--` style, I am following the cargo cult.

matt-leaguecio:
In IntelliJ i keep getting the following, it seems it doesn't support space in a folder specification.  I can't find where to change the path:  

```
"C:\Program Files\Java\jdk1.8.0_181\bin\java.exe" -agentlib:jdwp=transport=dt_socket,address=localhost:50825,suspend=n,server=y -Xdebug -server -Xmx1536M -Didea.managed=true -Dfile.encoding=UTF-8 -Dsbt.log.noformat=true -jar "C:\Users\Matt Ryan\.IntelliJIdea2019.1\config\plugins\Scala\launcher\sbt-launch.jar" "--addPluginSbtFile=C:\Users\Matt Ryan\AppData\Local\Temp\idea.sbt" "; set ideaPort in Global := 50780 ; idea-shell"
Listening for transport dt_socket at address: 50825
[error] Expected non-whitespace character
[error] --addPluginSbtFile=C:\Users\Matt Ryan\AppData\Local\Temp\idea.sbt
```

jastice:
@matt-leaguecio Looks like a bug in the Scala plugin. I've created an issue: 
https://youtrack.jetbrains.com/issue/SCL-15360

Issue #2881, opened 2016-12-28
'inspect' should consider aliases


dwijnand:
## steps

In '~/.sbtrc', define `alias c = compile`.

In 'build.sbt', define a task 'c'.

## problem

Invoking 'c' will invoke the alias (compile), while 'inspect c' will describe the task.

## expectation

Given invoking 'c' will invoke the alias I propose that inspect should consider this and communicate that to the (possibly unsuspecting) user.

## notes

I hit this issue because one of sbt scripted test (actions/depends-on) defines a 'c' task, but my 'c = compile' alias was being invoked instead, and it took my hours to discover the cause. I used 'inspect' but it didn't help me. Thus this feature request.

---

# Mentoring Instructions

## Links

* the `inspect` command is defined in https://github.com/sbt/sbt/blob/1d7b35bd67722511b9a058113e7a1070984e6758/main/src/main/scala/sbt/internal/Inspect.scala#L30-L68
* the `alias` command is defined in https://github.com/sbt/sbt/blob/1d7b35bd67722511b9a058113e7a1070984e6758/main-command/src/main/scala/sbt/BasicCommands.scala#L320-L327

## Solution

I think the starting point for a solution is to try to redefine `inspect` in terms of `State`'s `combinedParser`: https://github.com/sbt/sbt/blob/1d7b35bd67722511b9a058113e7a1070984e6758/main-command/src/main/scala/sbt/State.scala#L48 rather than `Act.scopedKeyParser`: https://github.com/sbt/sbt/blob/1d7b35bd67722511b9a058113e7a1070984e6758/main/src/main/scala/sbt/internal/Inspect.scala#L52

## Testing

At this time I think the simplest way to test this is by using [`sbtOn`](https://github.com/sbt/sbt/blob/1d7b35bd67722511b9a058113e7a1070984e6758/CONTRIBUTING.md#running-sbt-from-source---sbton) and testing manually on a minimal sbt build.

Note you don't need to use `~/.sbtrc`, you can also define sbt aliases in a build with `addCommandAlias`.

## Motivation

The motivation for this change is to bring more clarity: if you have a task `c` and an alias `c` it's very confusing that (1) running `c` will execute the alias but (2) running `inspect c` will give you info on the task!

## Background

My guess is `inspect` was built to help investigate how tasks are set up, and it's just a mistake that it doesn't take anything other than tasks into account.

PR #4221, opened 2018-06-19
Inspect supports alias

gpoirier:

Fixes https://github.com/sbt/sbt/issues/2881

### DIFF
--- a/main/src/main/scala/sbt/internal/Inspect.scala
+++ b/main/src/main/scala/sbt/internal/Inspect.scala
@@ -27,19 +27,18 @@ object Inspect {
   val Uses: Mode = UsesMode
   val Definitions: Mode = DefinitionsMode
 
-  def parser: State => Parser[(Inspect.Mode, ScopedKey[_])] =
+  def parser: State => Parser[() => String] =
     (s: State) =>
-      spacedModeParser(s) flatMap {
-        case opt @ (UsesMode | DefinitionsMode) =>
-          allKeyParser(s).map(key => (opt, Def.ScopedKey(Global, key)))
-        case opt @ (DependencyTreeMode | Details(_)) => spacedKeyParser(s).map(key => (opt, key))
+      spacedModeParser(s) flatMap { mode =>
+        commandHandler(s, mode) | keyHandler(s)(mode)
     }
-  val spacedModeParser: (State => Parser[Mode]) = (s: State) => {
+  val spacedModeParser: State => Parser[Mode] = (_: State) => {
+    val default = "-" ^^^ Details(false)
     val actual = "actual" ^^^ Details(true)
     val tree = "tree" ^^^ DependencyTree
     val uses = "uses" ^^^ Uses
     val definitions = "definitions" ^^^ Definitions
-    token(Space ~> (tree | actual | uses | definitions)) ?? Details(false)
+    token(Space ~> (default | tree | actual | uses | definitions)) ?? Details(false)
   }
 
   def allKeyParser(s: State): Parser[AttributeKey[_]] = {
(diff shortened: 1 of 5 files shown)


gpoirier:
I signed the CLA

eed3si9n:
@gpoirier Thanks for the contribution!

eed3si9n (changes requested):
Could you update `help inspect` in https://github.com/sbt/sbt/blob/1.x/main/src/main/scala/sbt/internal/CommandStrings.scala#L122-L123 too?

main/src/main/scala/sbt/internal/Inspect.scala
```diff
@@ -27,19 +27,18 @@ object Inspect {
...
-      spacedModeParser(s) flatMap {
-        case opt @ (UsesMode | DefinitionsMode) =>
-          allKeyParser(s).map(key => (opt, Def.ScopedKey(Global, key)))
-        case opt @ (DependencyTreeMode | Details(_)) => spacedKeyParser(s).map(key => (opt, key))
+      spacedModeParser(s) flatMap { mode =>
+        commandHandler(s, mode) | keyHandler(s)(mode)
     }
-  val spacedModeParser: (State => Parser[Mode]) = (s: State) => {
+  val spacedModeParser: State => Parser[Mode] = (_: State) => {
+    val default = "-" ^^^ Details(false)
```
gpoirier, line 36: @eed3si9n @dwijnand Is it a decent solution to wanting inspect the key named 'actual' with `Details(false)`? I can leave it out or renamed `-` to a different name if preferred.

gpoirier:
Good point, I missed the `help inspect` message. I'll update it, and I should also add something about the `-` mode if you guys want that in? https://github.com/gpoirier/sbt/blob/issues/2881/main/src/main/scala/sbt/internal/Inspect.scala#L36

I'll also have to figure out how to run the scripted tests locally so I don't rely on travis to know if my fix worked - because my first attempt didn't.

gpoirier:
It seems like this PR should be linked to the issue https://github.com/sbt/sbt/issues/2881, but I cannot figure out how to do that.

main/src/main/scala/sbt/internal/Inspect.scala
eed3si9n: It's a bit ironic since `-` is an actual command name in sbt 0.13 to do early processing, like `-warn`, but I think `inspect - somekey` is a decent escaping solution.

gpoirier:
I updated the help text, does it seems appropriate?

eed3si9n:
@gpoirier 

The beginning of the help text says:


>     |	For a plain setting, the value bound to the key argument is displayed using its toString method.
>     |	Otherwise, the type of task ("Task" or "Input task") is displayed.

Maybe it should be updated a bit to reflect your main change. For example:

```
For a plain setting, the value bound to the key argument is displayed using its toString method.
For an alias, the command bound to the alias is displayed.
Otherwise, the type of the key ("Task" or "Input task") is displayed.
```

gpoirier:
With the help change above, is it instead of the text I added at the bottom? The main reason I added it at the bottom was that the change apply to all of the modes.

eed3si9n:
Either works for me in terms of the text.

eed3si9n:
Overall LGTM.
Could you squash the commits please?

gpoirier:
It's squashed to a singe commit rebased on the current HEAD of 1.x.

eed3si9n:
Awesome.

eed3si9n (approved):

eed3si9n:
I am testing it locally, and seeing something like this:

```
sbt:helloworld> inspect show
[error] Not a valid project ID: show
[error] Expected ':'
[error] Not a valid key: show (similar: showTiming, sLog, ps)
[error] inspect show
[error]             ^
sbt:helloworld> inspect +
[error] Expected whitespace character
[error] Expected '-'
[error] Expected 'tree'
...
```

Is this the expected result for commands?

eed3si9n (changes requested):
See comment

gpoirier:
Those two (`+` and `show`) aren't in the list from `State#definedCommands`, which is what @dwijnand asked me to use. As far as I can tell, `show` doesn't seem to be technically be a command, or at least I can't find it by browsing the source if it's the case. Also, I can only use the `SimpleCommand`s, the `ArbitraryCommand`s doesn't expose a name.

The two error messages you listed are also different for reasons that already existed. One of the parser for the keys is looking at IDs before trying to match the result to a key. Which I'm realizing right now might be an issue, since technically keys could have names that are not valid IDs. But this PR isn't changing that behavior.

gpoirier:
@eed3si9n do you still want me to change something? If so, what should it be?

eed3si9n (approved):
ok. Let's merge this as is. Since adding support for alias was the original goal, so this PR does the job.

gpoirier:
Should a new ticket be open for the part that's not related to adding alias support then? What was your issue though, the fact that `show` and `+` weren't seen as commands or the error message? If it's the former, I don't think it can easily be solved.

eed3si9n:
I guess I was hoping that inspect somehow recognizes `+` as a command. I think it's ok as is.

We made a survey form to collect feedback on the contribution process. Please fill it out when get you get a minute. https://goo.gl/forms/LDPlTl9Wso7YwEri2

Issue #4197, opened 2018-06-13
Ctrl + C is not propagated to forked test


eed3si9n:
This is part of `Glocal / cancelable` epic.

See https://github.com/sbt/sbt/issues/3252#issuecomment-396813096

## steps

```scala
import Dependencies._

Global / cancelable := true
ThisBuild / organization := "com.example"
ThisBuild / scalaVersion := "2.12.6"
ThisBuild / version      := "0.1.0-SNAPSHOT"

lazy val root = (project in file(".")).
  settings(
    name := "Hello",
    Compile / fork := true,
    Test / fork := true,
    libraryDependencies += scalaTest % Test
  )
```

1. Run some long running test.

```scala
package example

import org.scalatest._

class HelloSpec extends FlatSpec with Matchers {
  "The Hello object" should "say hello" in {
    Hello.loop shouldEqual "hello"
  }
}

object Hello extends App {
  def loop: Unit = {
    println("hello")
    Thread.sleep(1000)
    loop
  }
}
```

2. Type in Ctrl + C into the shell.

## problem

```
sbt:Hello> Exception in thread "Thread-16" java.io.EOFException
	at java.io.ObjectInputStream$BlockDataInputStream.peekByte(ObjectInputStream.java:2958)
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1538)
	at java.io.ObjectInputStream.readObject(ObjectInputStream.java:431)
	at org.scalatest.tools.Framework$ScalaTestRunner$Skeleton$1$React.react(Framework.scala:818)
	at org.scalatest.tools.Framework$ScalaTestRunner$Skeleton$1.run(Framework.scala:807)
	at java.lang.Thread.run(Thread.java:748)
```

## expectation

Test shutsdown cleanly.

## notes

It might be an improvement to hold on to scala.sys.process.Process here https://github.com/sbt/sbt/blob/v1.1.6/main-actions/src/main/scala/sbt/ForkTests.scala#L125, and explicitly call destroy() on it.

driquelme:
@eed3si9n  ~~This might be a Scalatest issue.~~
What seems to be happening is that Scalatest spawns it's own server thread to listen for incoming events from the remote jvm tasks. 
https://github.com/scalatest/scalatest/blob/3.0.x-for-scala-2.12.x/scalatest/src/main/scala/org/scalatest/tools/Framework.scala#L849

[...]

PR #4226, opened 2018-06-22
Fix CTRL-C exception

driquelme:

ref: #4197 

Adds a shutdown hook in the forked jvm that calls `done` when a jvm shutdown is requested . 
For successful terminating tests, the hook is removed so that `done` is not called twice.


eed3si9n (approved):
LGTM
Thanks!

eed3si9n:
We made a survey form to collect feedback on the contribution process. Please fill it out when get you get a minute. https://goo.gl/forms/LDPlTl9Wso7YwEri2

Issue #4238, opened 2018-06-28
Flaky test actions/eval-is-safe-and-sound


eed3si9n:
On AppVeyor actions/eval-is-safe-and-sound fails a lot.

PR #4245, opened 2018-07-03
Select most recent JDK

raboof:
When multiple versions of, say, jdk8 are installed, pick the highest one.

### DIFF
--- a/main/src/main/scala/sbt/internal/CrossJava.scala
+++ b/main/src/main/scala/sbt/internal/CrossJava.scala
@@ -300,14 +300,15 @@ private[sbt] object CrossJava {
   }
 
   def discoverJavaHomes: ListMap[String, File] = {
-    import JavaDiscoverConfig._
-    val configs = Vector(jabba, linux, macOS)
-    ListMap(configs flatMap { _.javaHomes }: _*)
+    ListMap(JavaDiscoverConfig.configs flatMap { _.javaHomes } sortWith (versionOrder): _*)
   }
 
   sealed trait JavaDiscoverConf {
     def javaHomes: Vector[(String, File)]
   }
+
+  def versionOrder(left: (_, File), right: (_, File)): Boolean =
+    versionOrder(left._2.getName, right._2.getName)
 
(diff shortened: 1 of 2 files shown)


raboof:
(appveyor error seems to be #4238)

eed3si9n (approved):
Nice

main/src/main/scala/sbt/internal/CrossJava.scala
```diff
@@ -299,47 +300,78 @@ private[sbt] object CrossJava {
...
+            case dir @ JavaHomeDir(vendor, m, n) =>
+              val v = JavaVersion(nullBlank(m) + n).withVendor(vendor).toString
+              if ((base / dir / "Contents" / "Home").exists) v -> (base / dir / "Contents" / "Home")
+              else v -> (base / dir)
+          }
     }
+
+    val configs = Vector(
+      new JabbaDiscoverConfig,
+      new LinuxDiscoverConfig(file("/usr") / "java"),
```
eed3si9n, line 371: This adds support for Fedora. Ref https://github.com/akka/akka/issues/25298

MasseGuillaume:
This won't work in nix/nixos, you need to provide the list explicitly.

raboof:
@MasseGuillaume interesting, perhaps we should open a separate ticket about making this feature work with nix/nixos. I'm only superficially familiar with nixos, how would multiple installed JDK versions look from 'inside' a nix environment?

MasseGuillaume:
You create your environment like this:

https://github.com/scala-native/scala-native/blob/master/scripts/scala-native.nix#L10

You just need to be able to inject dependencies with an environment variable.

For example:

```
ORACLE_JDK_8 = pkgs.oraclejdk8; // available in all-packages.nix
```

https://github.com/NixOS/nixpkgs/blob/6b5c9fc197895ad2f3b63b473168a5871670eee3/pkgs/top-level/all-packages.nix#L6720

Then sbt will have `ORACLE_JDK_8` available via `sys.env`

MasseGuillaume:
https://github.com/sbt/sbt/issues/4260

PR #4253, opened 2018-07-08
use val instead of var

xuwei-k:


internal/util-collection/src/main/scala/sbt/internal/util/INode.scala
```diff
@@ -61,8 +61,8 @@ abstract class EvaluateSettings[Scope] {
     }): Seq[INode[_]]
   }
 
-  private[this] var running = new AtomicInteger
-  private[this] var cancel = new AtomicBoolean(false)
+  private[this] val running = new AtomicInteger
+  private[this] val cancel = new AtomicBoolean(false)
```
He-Pin, line 66: nice

eed3si9n (approved):

PR #4277, opened 2018-07-20
port addSbtPluginCommand to sbt 0.13

jastice:
Port #4211 to sbt 0.13 so that IntelliJ Scala plugin can use plugin injection without requiring users to upgrade their sbt builds to 1.x

edit: it actually does work as expected


eed3si9n:
I wonder if Travis is failing because it's on Java 6.

```
$ java -Xmx32m -version
java version "1.6.0_41"
```

Maybe we need to use `-Dsbt.repository.secure=false`

jastice:
I've had resolve problems with JDK7 as well. In that case adding these parameters worked:
`-Djdk.tls.client.protocols=TLSv1.2 -Dhttps.protocols=TLSv1.2`

With JDK6 that probably won't work.

jastice:
required for https://youtrack.jetbrains.com/issue/SCL-14247

eed3si9n:
Please rebase on 0.13.

Issue #4065, opened 2018-04-04
Unable to reference a key scoped to the sbt-web Assets configuration in the console


steinybot:
## steps

**build.sbt**
```
enablePlugins(SbtWeb)

val foo = taskKey[Classpath]("foo")

foo := (Assets / exportedProducts).value
```

**project/plugins.sbt**
```
addSbtPlugin("com.typesafe.sbt" %% "sbt-web" % "1.4.3")
```

## problem

I am unable to reference a key that is scoped to the `Assets` configuration from the console.
```
→ sbt
[info] Loading settings from idea.sbt ...
[info] Loading global plugins from /Users/jason/.sbt/1.0/plugins
[info] Loading settings from plugins.sbt ...
[info] Loading project definition from /source/sbt-config-key-bug/project
[info] Loading settings from build.sbt ...
[info] Set current project to sbt-config-key-bug (in build file:/source/sbt-config-key-bug/)
[info] sbt server started at local:///Users/jason/.sbt/1.0/server/3318aa2bd772e7169d56/sock
sbt:sbt-config-key-bug> inspect Assets / exportedProducts
[info] No entry for key.
[info] Description:
[info] 	Build products that go on the exported classpath.
[info] Delegates:
[info] 	Assets / exportedProducts
[info] 	exportedProducts
[info] 	ThisBuild / Assets / exportedProducts
[info] 	ThisBuild / exportedProducts
[info] 	Zero / Assets / exportedProducts
[info] 	Global / exportedProducts
[info] Related:
[info] 	Web-assets-test / exportedProducts
[info] 	Test / exportedProducts
[info] 	Runtime / exportedProducts
[info] 	Web-assets / exportedProducts
[info] 	Compile / exportedProducts
sbt:sbt-config-key-bug> inspect foo
[info] Task: scala.collection.Seq[sbt.internal.util.Attributed[java.io.File]]
[info] Description:
[info] 	foo
[info] Provided by:
[info] 	ProjectRef(uri("file:/source/sbt-config-key-bug/"), "sbt-config-key-bug") / foo
[info] Defined at:
[info] 	/source/sbt-config-key-bug/build.sbt:5
[info] Dependencies:
[info] 	Web-assets / exportedProducts
[info] Delegates:
[info] 	foo
[info] 	ThisBuild / foo
[info] 	Global / foo
sbt:sbt-config-key-bug> inspect Web-assets / exportedProducts
[error] Expected project ID
[error] Expected configuration
[...]

PR #4297, opened 2018-08-03
Fix incorrect unknown configuration warning on build settings

### LINKED ISSUE #4293: Regression: Some unknown configuration warnings are noisy or spurious

Since #4231 and sbt 1.2.0, some fairly common and innocuous setups lead to noisy warnings.

## steps

One case: use a `Test`-scoped setting in `inThisBuild` with an implicit root project:

```scala
// Entire contents of a build.sbt

inThisBuild(
  fork in Test := true
)
```

Eugene also pointed out sbt/sbt-multi-jvm#48, which at least could reasonably be addressed in that plugin.

## problem

The case above results in a warning that is unhelpful/doesn't need action:

```
[warn] The project {file:/private/tmp/4231/ }<root> references an unknown configuration "test" and was guessed to be "Test".
```

## expectation

No noisy warnings.

## notes

Workarounds: auto plugins may need to declare missing `projectConfigurations` overrides to silence warnings (in that case the warning may be positive/helpful), or `hide` a configuration, if appropriate.

sbt version: 1.2.0

steinybot:
Fixes #4293

Not too sure if it is ok to change the KeyIndex signature.

### DIFF
--- a/main/src/main/scala/sbt/internal/Load.scala
+++ b/main/src/main/scala/sbt/internal/Load.scala
@@ -328,8 +328,7 @@ private[sbt] object Load {
     val attributeKeys = Index.attributeKeys(data) ++ keys.map(_.key)
     val scopedKeys = keys ++ data.allKeys((s, k) => ScopedKey(s, k)).toVector
     val projectsMap = projects.mapValues(_.defined.keySet)
-    val configsMap: Map[String, Seq[Configuration]] =
-      projects.values.flatMap(bu => bu.defined map { case (k, v) => (k, v.configurations) }).toMap
+    val configsMap = projectConfigurations(projects)
     val keyIndex = KeyIndex(scopedKeys.toVector, projectsMap, configsMap)
     checkConfigurations(keyIndex, log)
     val aggIndex = KeyIndex.aggregate(scopedKeys.toVector, extra(keyIndex), projectsMap, configsMap)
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/internal/Load.scala
```diff
@@ -355,7 +375,9 @@ private[sbt] object Load {
           log.warn(
             s"""The project $project references an unknown configuration "$config" and was guessed to be "$guess"."""
           )
-          log.warn("This configuration should be explicitly added to the project.")
+          log.warn(
+            "This configuration should be explicitly added to the `projectConfigurations` of a plugin or the `configs` of the project."
```
eed3si9n, line 379: In the hindsight neither of the options are actually good advice. Adding a new configuration to the subproject has impact on the build beyond key lookup in the shell, so without understanding what the configuration does I don't think we can tell user to do so.

See for example https://github.com/lightbend/paradox/issues/225 which adds unwanted dependency on POM.

Since #4293, I am learning that people put config-scoped settings for all sorts of reasons, like code formatting on optionally added configuration. In those cases we can pretty much ignore those settings and not much would happen.

steinybot: I don't quite understand why adding the configuration would be bad. I would have thought that if someone is using a configuration it is safe to also put it on the project. Are there other examples?

In the paradox example it didn't look like having the `Compile` configuration on the project was the issue. Wasn't the issue there that the dependency was in the `Compile` scope? Both the `Compile` and `ParadoxTheme` configurations are on the project.

It's a shame we get the warning when it doesn't really matter.

eed3si9n: If someone publishes using Ivy layout `ivy.xml` will list configurations. So that would be another side effect, I think.

So I'd suggest the first step we should take is drop the warning at load time, and release 1.2.1.

Next, we should figure how sbt 0.13 is able to display `web-assets:` when you type `w[tab]` in the shell.

steinybot: I'm not entirely convinced that adding the configuration to the project is the issue.

Having it go in `ivy.xml` is an issue but that already exists and it doesn't mean that one shouldn't be putting it in `projectConfigurations`. It may need to go there for other reasons besides this warning such as to extend another configuration. There should be a way to prevent it going into `ivy.xml`.

The reason it works in 0.13 is because it uses the name and not the id. We can't do that with the unified slash syntax because configurations and tasks would be ambiguous. I put quite a few comments on #4065.

I didn't explore all the ramifications of not knowing the id but it seems like something that is deserving of a warning. One other problem is that errors are confusing as they use the name.

dwijnand:
Superseded by #4302, which was merged.

PR #4300, opened 2018-08-05
Set default maven configurations in pom configuration, fixes #1380

### LINKED ISSUE #1380: Publishing test and integration test dependencies in .pom problem
Dependencies like "com.typesafe" % "config" % "1.2.0" % "it" are published in default [compile] scope in pom. I think integration test dependencies should be published in test scope instead. But if "it" is replaced with "it,test", dependency is published in test scope. There is a contradiction here. If "it" dependency is published in compile scope, "it,test" should also be published in compile scope. Luckily this problem can be used as a workaround for the first one to improve generated pom.

Philippus:
This is a PR of @chEbba's commit: https://github.com/chEbba/sbt/commit/2882c3d8197cb2c953ccbf6865fa70bcfe54c781

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2031,7 +2031,8 @@ object Classpaths {
       .withExtra(pomExtra.value)
       .withProcess(pomPostProcess.value)
       .withFilterRepositories(pomIncludeRepository.value)
-      .withAllRepositories(pomAllRepositories.value),
+      .withAllRepositories(pomAllRepositories.value)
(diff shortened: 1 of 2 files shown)


eed3si9n:
@Philippus Thanks for picking this up.
The test failure on Travis CI seems legit tho.

chEbba:
@Philippus thanks for piking this up from me too, I was in the middle of relocation and had no time to PR thiss fix. Respect!

Philippus:
@eed3si9n is there a chance for this PR of being included in the `1.2.x` branch?

PR #4313, opened 2018-08-09
Support discovering JDK 11 in JAVA_HOME for CrossJava

raboof:
* Allow an optional '-' after the 'jdk' prefix in the directory name
* Allow discovering JDK's in `$JAVA_HOME`
* Allow discovering JDK's in `C://Program Files/Java`

### DIFF
--- a/main/src/main/scala/sbt/internal/CrossJava.scala
+++ b/main/src/main/scala/sbt/internal/CrossJava.scala
@@ -369,8 +369,7 @@ private[sbt] object CrossJava {
           }
     }
 
-    class WindowsDiscoverConfig extends JavaDiscoverConf {
-      val base: File = file("C://Program Files/Java")
+    class WindowsDiscoverConfig(base: File) extends JavaDiscoverConf {
 
       def candidates() = wrapNull(base.list())
 
@@ -403,7 +402,8 @@ private[sbt] object CrossJava {
       new LinuxDiscoverConfig(file("/usr") / "java"),
       new LinuxDiscoverConfig(file("/usr") / "lib" / "jvm"),
       new MacOsDiscoverConfig,
-      new WindowsDiscoverConfig,
+      new WindowsDiscoverConfig(file("C://Program Files/Java")),
+      new WindowsDiscoverConfig(file("C://Program Files (x86)/Java")),
       new JavaHomeDiscoverConfig,
     )
   }
--- a/main/src/main/scala/sbt/internal/CrossJava.scala
+++ b/main/src/main/scala/sbt/internal/CrossJava.scala
@@ -331,11 +331,13 @@ private[sbt] object CrossJava {
   }
 
   object JavaDiscoverConfig {
+    private val JavaHomeDir = """(java-|jdk-?)(1\.)?([0-9]+).*""".r
+
     class LinuxDiscoverConfig(base: File) extends JavaDiscoverConf {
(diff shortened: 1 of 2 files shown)


main/src/main/scala/sbt/internal/CrossJava.scala
```diff
@@ -367,11 +369,42 @@ private[sbt] object CrossJava {
           }
     }
 
+    class WindowsDiscoverConfig extends JavaDiscoverConf {
+      val base: File = file("C://Program Files/Java")
```
kxbmap: You can use `sys.env.get("ProgramFiles")`.

In 64-bit machine, `%ProgramFiles%` points to:
- `C:\Program Files` if running sbt with 64-bit Java
- `C:\Program Files (x86)` if running sbt with 32-bit Java

In 32-bit machine, `%ProgramFiles%` points to `C:\Program Files`.

eed3si9n (changes requested):
Could you address @kxbmap's comment on Windows please?

raboof:
Yes, hold on while I set up a windows vm to be able to test :)

muuki88 (approved):

main/src/main/scala/sbt/internal/CrossJava.scala
raboof: That variable isn't filled for me (sbt 1.2.1 on cygwin on windows8)?

Should we just look in  `C:\Program Files (x86)/Java` as well as `C:\Program Files/Java`?

eed3si9n: Looking for both makes sense.

raboof: Updated the PR to look in both.

To follow up on my last comment: the environment variable is `%PROGRAMFILES%` (case-sensitive). This always points to `C:\Program Files` on 32-bits machines. On 64-bits machines it points to `C:\Program Files` or `C:\Program Files (x86)` depending on your current process (https://stackoverflow.com/questions/17688758/using-programfilesx86-on-windows-os-32bit).

Can we invoke 32-bit Java from 64-bit sbt and vice-versa?

eed3si9n: > Can we invoke 32-bit Java from 64-bit sbt and vice-versa?

I am not sure if we tested specifically for these scenarios, but I'm guessing it's not too different from calling JDK 11 from JDK 8: You invoke java.exe in another process, and hope that serialization still hold up.

eed3si9n:
```
[error] C:\projects\sbt\main\src\test\scala\sbt\internal\CrossJavaTest.scala:55:22: not enough arguments for constructor WindowsDiscoverConfig: (base: java.io.File)sbt.internal.CrossJava.JavaDiscoverConfig.WindowsDiscoverConfig.
[error] Unspecified value parameter base.
[error]       val conf = new WindowsDiscoverConfig {
[warn] two warnings found
[error] one error found
[error] (mainProj / Test / compileIncremental) Compilation failed
```

main/src/main/scala/sbt/internal/CrossJava.scala
raboof: Right: then checking for both architectures seems to be the reasonable thing to do (even if it might then pick the 'wrong' one)

raboof:
oops

eed3si9n (approved):

PR #4318, opened 2018-08-11
Introduce conflictManagers settingKey

elbakramer:

## Introduce conflictManagers settingKey

[@elbakramer]: https://github.com/elbakramer

[4318]: https://github.com/sbt/sbt/pull/4318

### Fixes with compatibility implications

- New `conflictManagers` settingKey is introduced which can configure multiple conflict managers.
  - Original `conflictManager` can only configure a single conflict manager.
  - The `conflictManager` has remained for backward compatibility but users are encouraged to use the `conflictManagers` instead of the `conflictManager`.
  - Setting both settingKeys is not the desired case. But for now, it will use both of `conflictManager` and `conflictManagers`, with `conflictManager` value taking precedence.

### Improvements

- Can configure multiple conflict managers by setting `conflictManagers` settingKey [#4318][4318] by [@elbakramer][@elbakramer]

### Bug fixes

- Not a bug actually, but minor changes related to mixed indentation (tabs/spaces).

### Notes

- The order is important when setting `conflictManagers`. Managers on the later side of the sequence won't be considered when one of the previous managers has matching organization/module patterns and handles already.

  For example, following setting that appends managers to the existing `conflictManagers` actually does nothing since it's behind the default manager which handles every case.

  ```sbt
  conflictManagers += ConflictManager.strict // Seq(ConflictManager.default, ConflictManager.strict)
  ```

  If someone wants a different manager to handle some specific cases while falling back to the existing managers, one would prepend the manager instead of appending it.

  ```sbt
  conflictManagers := ConflictManager.strict +: conflictManagers.value
  ```


typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/932/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4318/head-1.x-1.x-pull/260/head-1.x-1.x)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/932/">has checked</A> the following projects against Scala 2.12,
tested using sbt-allsources.sh.

Project             | Reference            | Commit
:------------------ | :------------------- | :----------------
  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4318">pull/4318/head      </A> | sbt/sbt@c7a4d89c52fedff869cc688b583a4aa131244614
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/1.x">1.x                 </A> | sbt/zinc@c705199fe7387729a1c1a54b4ab2ac4e347b860a
  io                | <A HREF="https://github.com/sbt/io/tree/1.x">1.x                 </A> | sbt/io@36abd944e750c04d154b2c3218f4cc1a7e40a74a
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/pull/260">pull/260/head       </A> | sbt/librarymanagement@7dbc64d114198cbacf9105f4a1909045e558dd2e
  util              | <A HREF="https://github.com/sbt/util/tree/1.x">1.x                 </A> | sbt/util@39070c18e2280bb6532c13709bb12226259c2960
  website           | <A HREF="https://github.com/sbt/website/tree/1.x">1.x                 </A> | sbt/website@9e94e3f8ccad15543f843f2353b6998cfda5246e

:white_check_mark: The result is: ***SUCCESS***
<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4318/head&zinc=1.x&io=1.x&librarymanagement=pull/260/head&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

muuki88 (commented):
Thanks for your well done pull request 🤩

I'm not familiar with the implementation details of the conflict manager so bare with me 😀

Introducing new settings increases the maintenance overhead. Would it be possible to provide and `CompositeConflictManager` which provides the same functionality without requiring a new setting?

```scala
conflictManager := CompositeConflictManager(cm1, cm2)
```

main/src/main/scala/sbt/Keys.scala
```diff
@@ -451,6 +451,7 @@ object Keys {
   val compatibilityWarningOptions = settingKey[CompatibilityWarningOptions]("Configures warnings around Maven incompatibility.").withRank(CSetting)
   val conflictWarning = settingKey[ConflictWarning]("Configures warnings for conflicts in dependency management.").withRank(CSetting)
   val conflictManager = settingKey[ConflictManager]("Selects the conflict manager to use for dependency management.").withRank(CSetting)
+  val conflictManagers = settingKey[Seq[ConflictManager]]("Selects the conflict managers to use for dependency management.").withRank(CSetting)
```
muuki88, line 454: If we introduce this, then we should add a deprecation warning on the old setting.

I know there hasn't been any scaladoc on most of the settings, but IMHO the explanation you put in this PR would be a perfect scaladoc here.

eed3si9n (changes requested):
Note this is currently blocked on https://github.com/sbt/librarymanagement/pull/260, which breaks bincompat.

eed3si9n:
Muuki's suggestion of trying to create a composite manager makes sense to me.
I'm closing this for now.

PR #4321, opened 2018-08-14
Properly log the ClassfileManager output

smarter:


smarter:
@eed3si9n was there a reason for using Logger.Null here?

eed3si9n:
It's been `Logger.Null` ever since the logging was added by @gkossakowski in https://github.com/sbt/sbt/pull/1254. It kind of makes sense to me since the log is for debugging the behavior of sbt, Zinc, or compiler, and not normal build users.

Due to the design of sbt, even when the debug level logs are emitted to file logger, so there's a minor performance hit to these logs, so that could be another reason why the null logger was chosen here.

smarter:
But that's inconsistent with the fact that I can enable the zinc logs without changing any logger:
```scala
logLevel in Global := Level.Debug
incOptions in Global ~= { _
  .withApiDebug(true)
  .withRelationsDebug(true)
}
```

eed3si9n:
I'm not saying sbt has always been consistent on this policy, but I think the fact that you need

```scala
incOptions in Global ~= { _
  .withApiDebug(true)
  .withRelationsDebug(true)
}
```

is somewhat consistent with what I am saying, since `debug` alone still won't turn it on.

eatkins:
What about adding a TaskKey `zincLogger` (or some other name) that can be used to override the default logger? I took this approach for the logger used during a continuous build: https://github.com/sbt/sbt/blob/8df2807cd16b503826e0061abde4522bd1c880f1/main/src/main/scala/sbt/Keys.scala#L101

smarter:
@eatkins That sounds good to me, though I'd also like a way to tell sbt "just log everything you can possibly log, I'll sort it out"

eatkins:
@smarter I hear that. I actually removed the watchLogger key and replaced it with a watchLogLevel key. It allows the watch log level to be decoupled from the global debug level, so if I do `set watchLogLevel := Level.Debug`, then only the watch debug messages are printed even when the global level is set to info. If the global level is set to debug, then the watch log messages are also printed.

smarter:
Ping. I'd like to get this in since it might somehow help debug https://github.com/sbt/sbt/issues/5913, what would I need to change exactly to get this in? I'm fine with adding an extra logger as was suggested before but I'd like more details on what should be done exactly.

eed3si9n:
Yea. Given the mysterious missing directory issue, it might make sense to get this in.

eed3si9n:
Closing in favor of https://github.com/sbt/sbt/pull/5990

PR #4329, opened 2018-08-22
Allow same file to be packaged more than once

muuki88:
Allow files to be packaged more than once. This works now as expected

```scala
mappings ++= Seq(
  foo -> "foo1",
  foo -> "foo2"
)
```

### DIFF
--- a/main-actions/src/main/scala/sbt/Package.scala
+++ b/main-actions/src/main/scala/sbt/Package.scala
@@ -25,6 +25,14 @@ import sbt.util.CacheImplicits._
 import sbt.util.Tracked.{ inputChanged, outputChanged }
 
 sealed trait PackageOption
+
+/**
+  * == Package ==
+  *
+  * This module provides an API to package jar files.
+  *
+  * @see [[https://docs.oracle.com/javase/tutorial/deployment/jar/index.html]]
+  */
 object Package {
   final case class JarManifest(m: Manifest) extends PackageOption {
     assert(m != null)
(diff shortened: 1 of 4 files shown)


main-actions/src/main/scala/sbt/Package.scala
```diff
@@ -67,24 +89,30 @@ object Package {
...
           if (inChanged || outChanged) {
-            makeJar(sources.toSeq, jar.file, manifest, log)
+            makeJar(sources, jar.file, manifest, log)
             jar.file
           } else
             log.debug("Jar uptodate: " + jar.file)
         }
     }
 
-    val map = conf.sources.toMap
```
muuki88, line 84: This was the actual fix

eed3si9n (approved):
LGTM pending CI

muuki88:
Should I merge this? What's the process for integrating this into the 1.1.x and 1.2.x branches?

eed3si9n:
There's actually now a debate about whether we should merge to 1.x or 1.2.x first. Let's hold off on this for a day.

muuki88:
@eed3si9n I'll reopen this against `1.2.x` then, right?

eed3si9n:
@muuki88 I think so.

main-actions/src/main/scala/sbt/Package.scala
eed3si9n: An interesting effect the use of `Map` had was deduplicating the input mapping.
https://github.com/sbt/sbt/issues/4889

PR #4332, opened 2018-08-27
Add scalaCompilerBridgeBinaryJar task

smarter:
The end goal is to rewrite Dotty's compiler-bridge in Java (this is easy
since the zinc-specific phases are in the compiler itself) to simplify
the bootstrapping process.


smarter:
@Duhemm can you review ?

Duhemm (approved):
LGTM.

Is the future plan to use this new task in sbt-dotty primarily, or is it only for bootstrapping purposes?

If the plan is to use that in sbt-dotty, then we should make sure that the bridge is compiled with the oldest Java version that Dotty supports before distributing, but I don't see another problem.

smarter:
Yes, the plan would be to make https://github.com/lampepfl/dotty/tree/master/sbt-bridge Java-only and keep it that way. Our CI should always publish using the oldest Java version we support, otherwise things would break even now since we have some Java files.

smarter:
@eed3si9n OK to merge?

eed3si9n (approved):

eed3si9n:
Yea. Let's merge this.

PR #4334, opened 2018-08-28
Add an implicit class for StateOps

eatkins:
The State file in intellij was littered with red squiggly lines wherever
the extension methods of State called a different extension method of
State.  These went away when I switched to an implicit class, which is
the preferred way of adding extension methods since scala 2.10. As a
bonus, I was able to switch the implicit class to be a value class, so
it should not actually make a new object in most use cases.

I had to re-implement the stateOps method to delegate to the implicit
class for binary compatibility.

### DIFF
--- a/main-command/src/main/scala/sbt/State.scala
+++ b/main-command/src/main/scala/sbt/State.scala
@@ -61,7 +61,7 @@ trait Identity {
 }
 
 /** Convenience methods for State transformations and operations. */
-trait StateOps {
+trait StateOps extends Any {
   def process(f: (Exec, State) => State): State
 
   /** Schedules `commands` to be run before any remaining commands.*/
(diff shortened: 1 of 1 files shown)


main-command/src/main/scala/sbt/State.scala
```diff
@@ -233,8 +233,11 @@ object State {
     )
   }
 
+  @deprecated("Import State._ or State.StateOpsImpl to access state extension methods", "1.3.0")
+  def stateOps(s: State): StateOps = new StateOpsImpl(s)
```
eed3si9n, line 237: Does this break source compatibility if someone's build.sbt had `"compile" :: state`?

eatkins: No, because under the hood an implicit class is implemented with an implicit def. But to double check, I did add `commands += Command.command("alternative-compile") { s => "compile" :: s }` to a simple build.sbt and I had no problems loading the build with my snapshot sbt version.

dwijnand: Could've also just called the class `stateOps` 😉

eatkins: I think it's not too late!

dwijnand: Oh yeah. 😄 https://github.com/sbt/sbt/pull/4455

Issue #4179, opened 2018-06-01
sbt ~ repeat cmd: source changes outside project directory and outside home directory


Rich2:
The ~ repeat command is not responding to changes of sources outside of the build directory and outside of the home directory in Sbt 1.1.6. This does work in 1.1.4 in Kubuntu 18.10, but not in Ubuntu 18.10.

I've created a Hello world project to demonstrate: [sbtrepeat](https://github.com/Richtype/sbtrepeat.gitl)

Clone this project outside of your home directory structure.
Edit the BuildFolder/build.sbt file to hardcode the absolute path of the source directory.
run sbt from the BuildFolder
 command: ~run
 edit and save the MyMain.scala file in the src directory.

For me repeat command does not respond by rebuilding and rerunning. However if you run "show watchSources" the src directory will be listed. 

#3499 may be related. This stopped working for me when upgrading to sbt 1.1.0

sbt version: 1.1.6

dwijnand:
Lemme just /cc @eatkins, Director of ~ 😄

eed3si9n:
Does this still happen with sbt 1.3.2?

eed3si9n:
Hopefully it's now fixed. If not let us know.

PR #4335, opened 2018-08-29
Watch improvements

eatkins:

This is a large pull request that is the culmination of a number of month's work trying to improve the overall io and watch performance of sbt. For the most part, I've broken the commits up so that each one more or less stands on its own. I am doing it as one large pull request just because the overhead of making 20 partially dependent commits feels to high. It is linked with https://github.com/sbt/io/pull/177.

At a high level, this PR does two things:

1) Improve continuous builds. I made it possible to add custom trigger rules so that the user can, for example, cancel the watch if the command fails. I also fixed logging so that it should be easier to debug any bugs in the field.

2) Add a global in memory cache of the file system to improve incremental compilation start up times. This works well to reduce the startup time of incremental compilation. For a simple project with 5000 files, I've seen the compilation startup time drop from O(1 second) to O(200ms) (for small projects the startup time is O(50ms) where the 50ms is entirely spent in the update task to re-resolve the project). To achieve this speed up we combine the cache with the ExternalLookup in zinc so that we can bypass the io cost of hashing the source file contents. I also add a hybrid option that polls some directories and uses the regular os notifications for the remainder. This is to handle the case where, say, inotify isn't supported on a particular file system mount point (I think this may be the underlying issue with #4179 ). 

I repeated the experiment in [mill watch latency](http://www.lihaoyi.com/post/MillBetterScalaBuilds.html#watch--re-evaluate). With a warm jvm and if I use a resident compiler, I see O(120ms) latency between when file is modified and when _compilation_ is **complete** not when it begins. Also, in my test project with 5000 files mill burns an entire cpu polling the file system while sbt sits at about 0.3% of a cpu.

Seeing as this is a large change, I expect that there may be a good amount of discussion. Wherever possible, I've tried to keep the user visible impact of this change to a minimum. Moreover, everything I've added is configurable in such a way that a user should be able to revert back to the existing behavior with a few simple modifications if there are any new issues. It is always scary to add a global cache. In this case, I think the risk is not that bad both because I added a task to flush its contents and touching a monitored directory also triggers a full rescan. At any rate, users can choose to eschew the cache via the fileTreeViewConfig key.

I also am still working on adding some more scripted tests of continuous builds. I wanted to get this out the door, but I do expect to update a few commits. I don't expect to change much of the mainline code though unless errors are discovered writing additional scripted tests.


build.sbt
```diff
@@ -9,6 +9,7 @@ import local.Scripted
 def buildLevelSettings: Seq[Setting[_]] =
   inThisBuild(
     Seq(
+      doc in Compile := file("/dev/null"),
```
eed3si9n: What's the purpose of this?

main-command/src/main/scala/sbt/FileTreeViewConfig.scala
```diff
@@ -0,0 +1,164 @@
...
+        newMonitor(view.asInstanceOf[T], sources, logger)
+    )
+
+  /**
+   * Provides a [[FileTreeViewConfig]] with semantics as close as possible to sbt 1.2.0. This means
+   * that there is no file tree caching and the [[FileEventMonitor]] will use a [[WatchService]].
+   * @param antiEntropy the duration of the period after a path triggers a build for which it is
+   * @return a [[FileTreeViewConfig]] instance.
+   */
+  def `1.2.0`(pollingInterval: FiniteDuration, antiEntropy: FiniteDuration): FileTreeViewConfig =
```
eed3si9n: Not sure if this name is clear when it comes up as 

```scala
fileTreeViewConfig := FileTreeViewConfig.`1.2.0`(pollInterval.value, watchAntiEntropy.value)
```

Maybe `sbt1_2_compat`?

main-command/src/main/scala/sbt/Watched.scala
```diff
@@ -103,8 +112,127 @@ object Watched {
   val PollDelay: FiniteDuration = 500.milliseconds
   val AntiEntropy: FiniteDuration = 40.milliseconds
   def isEnter(key: Int): Boolean = key == 10 || key == 13
-  def printIfDefined(msg: String) = if (!msg.isEmpty) System.out.println(msg)
+  def printIfDefined(msg: String): Unit = if (!msg.isEmpty) System.out.println(msg)
+
+  type Task = () => State
```
eed3si9n: The term task mean something already so please pick some other term here.
It would be good to add Scaladoc comment on what it is.

main-command/src/main/scala/sbt/Watched.scala
```diff
@@ -103,8 +112,127 @@ object Watched {
...
+  type Setup = (State, Watched, String) => (State, Logger, Task => State)
```
eed3si9n: I tend to use the term "configuration" for these types of things.
Or maybe `WatchSetup`.

main-command/src/main/scala/sbt/Watched.scala
```diff
@@ -103,8 +112,127 @@ object Watched {
...
+   * @param state the initial state
+   * @param command the command(s) to repeatedly apply
+   * @param setup function to generate a logger and a transformation of the resultant state. The
+   *              purpose of the transformation is to preserve the logging semantics that existed
+   *              in the legacy version of this function in which the task would be run through
+   *              MainLoop.processCommand, which is unavailable in the main-command project
+   * @return the initial state if all of the input commands are valid. Otherwise, returns the
+   *         initial state with the failure transformation.
+   */
+  private[sbt] def command(
```
eed3si9n: Can we call this `executeContinuously`?

main-command/src/main/scala/sbt/Watched.scala
```diff
@@ -148,6 +276,112 @@ object Watched {
     }
   }
 
+  /**
+   * Provides a number of configuration options for continuous execution.
+   */
+  trait Config {
```
eed3si9n: If this is meant to be used by the build user, I think it's better to move it out of `object Watched`, and name it `WatchConfiguration` or something.

main-command/src/main/scala/sbt/Watched.scala
```diff
@@ -148,6 +276,112 @@ object Watched {
...
+     * A logger.
+     * @return a logger
+     */
+    def logger: Logger
+
+    /**
+     * Returns true if the continuous execution should stop.
+     * @return true if the contiuous execution should stop.
+     */
+    def shouldTerminate(): Boolean
```
eed3si9n: "should terminate" sounds like it should eventually terminate (finish calculation).
If it's for stopping, maybe flip the boolean and call it "isAlive"?

main-command/src/main/scala/sbt/Watched.scala
```diff
@@ -50,18 +62,46 @@ trait Watched {
 }
 
 object Watched {
-  val defaultWatchingMessage: WatchState => String = ws =>
-    s"${ws.count}. Waiting for source changes... (press enter to interrupt)"
+  sealed trait Action
```
eed3si9n, line 72: These probably should be generated using Contraband.

eed3si9n (changes requested):
I picked on some superficial things.

build.sbt
eatkins: https://github.com/sbt/sbt/pull/4335/commits/ba6d0fb4a7d799f6c1011be199a36370414090ac should be reverted. I was using this to speed up local development but didn't mean to include it in the PR.

main-command/src/main/scala/sbt/Watched.scala
eatkins: Good point. I'd reused the previous method name, but you're right that is unclear.

main-command/src/main/scala/sbt/Watched.scala
eatkins: Fair point. I might also be able to remove this alias anyway. I think I was mainly using it so that I could avoid the extra parens in `type Setup = (State, Watched, String) => (State, Logger, (() => State))`

main-command/src/main/scala/sbt/Watched.scala
eatkins: I'll come up with an alternate name. I didn't put a lot of thought into this to be honest.

main-command/src/main/scala/sbt/Watched.scala
eatkins: Sure.

main-command/src/main/scala/sbt/Watched.scala
eatkins: I guess for me it's a bit of a style thing. I find it intuitive to scope the kind of Configuration by putting it inside of the companion object rather than using semantic scoping with a name prefix. But I don't feel strongly enough to argue against making this change.

main-command/src/main/scala/sbt/Watched.scala
eatkins: I was going back through the PR and realized that this method ends up being removed by the end of the PR anyway. It's just a placeholder to make the git/pr history more readable.

main-command/src/main/scala/sbt/Watched.scala
eatkins: Can you expand on that? I understand that Contraband is used to evolve binary compatible apis, but I'm not exactly clear on how that would apply here.

main-command/src/test/scala/sbt/WatchedSpec.scala
```diff
@@ -0,0 +1,143 @@
...
+import sbt.io.FileEventMonitor.Event
+import sbt.io.{ IO, TypedPath }
+import sbt.util.Logger
+>>>>>>> 651478000... Refactor Watched
+
+import scala.collection.mutable
+import scala.concurrent.duration._
+
+class WatchedSpec extends FlatSpec with Matchers {
+<<<<<<< HEAD
```
eed3si9n: Git merge artifact

main-command/src/main/scala/sbt/Watched.scala
eed3si9n: I use Contraband for enums as well - https://github.com/sbt/sbt/blob/develop/main/src/main/contraband/main.contra
But I just noticed that Action is not completely sealed because of Custom, so that might not be a good fit.

main-command/src/test/scala/sbt/WatchedSpec.scala
eatkins: Whoops. Fixed.

main-command/src/main/scala/sbt/Watched.scala
eatkins: Aah, makes sense. I am not opposed to contraband. The `Custom` action is sort of speculative in the event that users might want to use ~ to potentially trigger any arbitrary state transformation. My guess is that very, very few users would ever actually do this (even if I think it's a cool capability).

eatkins:
In addition to making the requested changes from @eed3si9n, I figured out how to use jline to swap stdin to char buffered rather than line buffered which allowed me to add the feature that while in a continuous build, typing 'r' will re-run the command (see 
https://github.com/sbt/sbt/pull/4335/commits/fe64c5e7dc0d2a91ae0827b082a51073c0dddaa5) .

eed3si9n (dismissed):

eatkins:
I tacked on 
https://github.com/sbt/sbt/pull/4335/commits/0a9ae7b4b49da95bcac27db8dace663b54ff7686
though I wish that I'd just gone ahead and merged first so that it could have been a separate PR.

eed3si9n (approved):

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -351,12 +369,14 @@ object Defaults extends BuildCommon {
...
-    ).value,
-    watchSources in ConfigGlobal ++= {
+    unmanagedSources := FileManagement
+      .collectFiles(
+        unmanagedSourceDirectories,
+        includeFilter in unmanagedSources,
+        excludeFilter in unmanagedSources
+      )
+      .value,
+    watchSources in ConfigGlobal := (watchSources in ConfigGlobal).value ++ {
```
dwijnand, line 379: > Using ++= prevents these from being overridden by the user without doing
some nasty runtime manipulations.

I'm confused by this?  How is this line different than the original?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -377,6 +397,13 @@ object Defaults extends BuildCommon {
         else Nil
       bases.map(b => new Source(b, include, exclude)) ++ baseSources
     },
+    watchProjectSources in ConfigGlobal := (watchProjectSources in ConfigGlobal).value ++ {
```
dwijnand, line 400: Is it sufficient to `Global / watchProjectSources := Nil` to opt out of this?

main/src/main/scala/sbt/Defaults.scala
eatkins: Welcome back, @dwijnand :). I added this line because it was impossible for me to remove the default watchSources in the CloseWatchPlugin because of the use of `++=`. In a plugin or a build, if you do something like `watchSources in ConfigGlobal := Nil`, you will still end up with the default watch sources. I had to manually remove these in the CloseWatchPlugin in a very obscure way: https://github.com/swoval/swoval/blob/0d4882e87e0d1deb710338e3d55bed17d772bbe4/plugin/src/main/scala/com/swoval/watchservice/CloseWatchPlugin.scala#L278

main/src/main/scala/sbt/Defaults.scala
eatkins: Should be. I am planning on overhauling the documentation before the 1.3.0 release to explain all of this.

main/src/main/scala/sbt/Defaults.scala
dwijnand: OK, the reason for that is probably because in `Defaults` it's defined in `sourceConfigPaths` (i.e applied per configuration), while in `CloseWatchPlugin` you were probably trying to define it in `globalSettings`.  So it's an ordering thing of which definition in Global scope wins.

But that doesn't 100% explain why `k ++= x` transformed to `k := k.value ++ x` aren't the same.  Unless, of course, they're subtly different when sbt does it's setting re-ordering thing as it processes them (I forget where).

🤷‍♂️

main/src/main/scala/sbt/Defaults.scala
dwijnand: I was hoping to hear something like that, as I didn't notice any notes in this PR (which was never a hard requirement).  Lovely, thank you.  You've been doing some amazing stuff.

main/src/main/scala/sbt/Defaults.scala
eatkins: You know, I have to admit that I never actually tested this change, so I have no idea if it actually makes a difference. My mental model was just that `:=` always overrides any values set previously with `:=` and that `++=` always appends regardless of any previous overrides of `:=`. But I suspect that is not actually correct given your comments. The guide to scopes in the reference manual is helpful, but I wish there was a more detailed explanation of the exact key resolution algorithm.

main/src/main/scala/sbt/Defaults.scala
eatkins: Thanks for the feedback! It's always nice to hear that your work is appreciated.

main/src/main/scala/sbt/Defaults.scala
dwijnand: I think your intuition is right.  I was just grasping at straws trying to come up with plausible reasons why the change made a difference.  But if you're saying you didn't test it, then I'm more confident that it changed nothing. 😄

`:=` is actually interesting because it'll even suppress setting mistakes, for instance:

```scala
val someSources = settingKey[Seq[String]]("")

someSources ++= Seq("a", "b", "c") // or 'someSources := someSources.value ++ ..'
someSources := Seq("d")
```

results in `Seq("d")`, but without the `:=` line it results in

```scala
[error] Reference to undefined setting:
[error]
[error]   someSources from someSources (/s/t-test-order/build.sbt:4)
```

dwijnand: Oh and large parts of my understand about scopes and settings comes from experimentation 🙂

Issue #4038, opened 2018-03-26
sbt 1.x breaks forkOptions for custom test configurations


jalaziz:
## steps

```scala
import sbt._
import sbt.Keys._
import sbt.librarymanagement.Configuration

lazy val scalaTest = "org.scalatest" %% "scalatest" % "3.0.5"
lazy val SlowTests = config("slowtests").extend(Test)

lazy val root = (project in file(".")).
  configs(SlowTests).
  settings(
    inConfig(SlowTests)(Defaults.testTasks),
    inThisBuild(List(
      organization := "com.example",
      scalaVersion := "2.12.4",
      version      := "0.1.0-SNAPSHOT"
    )),
    name := "sbt-test",
    libraryDependencies += scalaTest % Test,
    run / fork := true,
    Test / fork := true,
    Test / testForkedParallel := true,
    Test / parallelExecution := true,
    SlowTests / parallelExecution := false,
    SlowTests / javaOptions += "-Djava.library.path=/path/to/lib"
  )
```

Run `sbt ";debug;slowtests:test"`.

## problem

With sbt 0.13.15, the forked JVM was correctly using the `javaOptions` set on the custom configuration.

With sbt 1.1.1, the forked JVM no longer uses the `javaOptions` of the custom test configuration and only honors the configuration from `Test`. This can be seen by inspecting `forkOptions`:
```
sbt:sbt-test> show root/slowtests:forkOptions
[info] ForkOptions(None, None, Vector(), Some(/Users/jalaziz/projects/sbt-test), Vector(), false, Map())
[success] Total time: 0 s, completed Mar 26, 2018 1:28:24 AM
```

## expectation

`forkOptions` should honor the `javaOptions` value set on the custom configuration as it did with sbt 0.13.15 or this change in behavior should be documented.

## notes

sbt version: 1.1.1

I believe the issue was introduced with 0419098d655a574cb81821786538a1955c68850a.

If I use `Defaults.testSettings` instead of `Defaults.testTasks`, the `forkOptions` setting is correct. I'm also able to work around the issue without loading all the test settings by simply adding `Defaults.testTasks ++ Seq(forkOptions := Defaults.forkOptionsTask.value)`.

subotic:
It also doesn't work with sbt version 1.2.0.

@jalaziz Thanks for the workaround.

Trying for days to get my custom test configurations working again.

[...]

PR #4336, opened 2018-08-30
sbt-4038 add regression test for fork-options

muuki88:
This pull request adds a regression test for #4038 

I noticed a couple of things

It's unlikely that 0419098d655a574cb81821786538a1955c68850a introduced this regression. The new default `forkOptions := forkOptionsTask.value` is set `Defaults.configTasks`, which isn't used by `Defaults.testTasks`. I wonder how things worked before. You should always use `inConfig(SlowTests)(Defaults.testSettings)` otherwise all the test related settings aren't set and sbt tries to resolve the values somewhere else. In the given regression test it would be the `Test` scope as `config("slowtests").extends(Test)`.

The change in [74cfbd4a9ca162dbe9d6fc236ac19671ff16d951](https://github.com/sbt/sbt/commit/74cfbd4a9ca162dbe9d6fc236ac19671ff16d951#diff-6373e7f7122325e753b75fe1cc76ff5fR1166) introduces the `forkOptions` to the `run` configuration, which are used in various places. Not sure if this could be related as well.

The regression test checks for the `javaOptions` and `runJVMOptions` to be set. This works fine when `Defaults.testSettings` are used. However the actual executed test doesn't seem to use the `javaOptions`.

- [ ] Document `inConfig(SlowTests)(Defaults.testSettings)` 
- [ ] Fix `javaOptions` not being available in the executed test

### DIFF
--- a/sbt/src/sbt-test/tests/fork-options-custom-config/test
+++ b/sbt/src/sbt-test/tests/fork-options-custom-config/test
@@ -1,3 +1,3 @@
 > check
 > test
--> slowtests:test
+-> Slowtests / test
--- a/sbt/src/sbt-test/tests/fork-options-custom-config/test
+++ b/sbt/src/sbt-test/tests/fork-options-custom-config/test
@@ -1,0 +1,3 @@
+> check
+> test
+-> slowtests:test
--- a/sbt/src/sbt-test/tests/fork-options-custom-config/build.sbt
+++ b/sbt/src/sbt-test/tests/fork-options-custom-config/build.sbt
@@ -1,0 +1,31 @@
+// test setup
+lazy val scalaTest = "org.scalatest" %% "scalatest" % "3.0.5"
+lazy val SlowTests = config("slowtests").extend(Test)
+
+configs(SlowTests)
+inConfig(SlowTests)(Defaults.testSettings)
+name := "fork-options-custom-config-test"
+
+libraryDependencies += scalaTest % Test
+
+SlowTests / javaOptions += "-Dtest.init.fail=true"
+
+// assertions
+lazy val check = taskKey[Unit]("check")
+check := {
+  val slowtestJavaOptions = (SlowTests / javaOptions).value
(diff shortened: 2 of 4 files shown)


sbt/src/sbt-test/tests/fork-options-custom-config/test
```diff
@@ -0,0 +1,3 @@
+> check
+> test
+-> Slowtests / test
```
muuki88, line 3: This doesn't fail. Either forking doesn't work properly _or_ the `javaOptions` come from some place far far away.

I used to add  these before the `check` command:

```
> show SlowTests / forkOptions
> show SlowTests / run / forkOptions
> show SlowTests / test / forkOptions
> inspect SlowTests / forkOptions
> inspect SlowTests / run / forkOptions
> inspect SlowTests / test / forkOptions
> show test / forkOptions
```

muuki88:
So I figured out that adding this configuration ([as described here](https://github.com/sbt/sbt/issues/4038#issuecomment-411393098)) indeed fixes the issue. 

It seems that somewhere along the way the classpath gets messed up by `Defaults.testSettings`.

**Without fix**
```
> inspect Slowtests / compile
[info] Provided by:
[info] 	ProjectRef(uri("file:/home/muki/dev/testing/sbt/fork-options-bug/"), "fork-options-bug") / Test / compile
[info] Defined at:
[info] 	/home/muki/dev/testing/sbt/fork-options-bug/build.sbt:6
```

**With fix**
```
[info] Provided by:
[info] 	ProjectRef(uri("file:/home/muki/dev/testing/sbt/fork-options-bug/"), "fork-options-bug") / Test / compile
[info] Defined at:
[info] 	(sbt.Defaults.configTasks) Defaults.scala:542
```

due to this, no test classes can be discovered. 

A minimal working solution is

```scala
inConfig(SlowTests)(Defaults.testSettings ++ Seq(compile := (Test / compile).value))
```

muuki88:
Closing. [See my last two comments](https://github.com/sbt/sbt/issues/4038#issuecomment-429846242)

PR #4341, opened 2018-09-03
Add `print` command that mimics `show`, but to standard output, not the logger

### LINKED ISSUE #4184: Plain print project settings
## problem
As a build master, I want to print the project settings in plain text for consumption by other scripts. For example, setting ENV vars, or setting properties in Jenkins pipeline scripts.

``` bash
export SBT_PROJECT_VERSION=$(sbt 'show version')
```

## expectation
When I run "show", I would like an output like:

>$ sbt 'show version'
0.1-SNAPSHOT

I dont want any log output. If there's an error, just print an empty string.

## notes
If changing the behaviour of `show` is not possible, then a new task like `$sbt 'print version'` would suffice.

I was steered here from a discussion on #840

Falmarri:
Fixes  #4184

- [ x ] I've read the [CONTRIBUTING](https://github.com/sbt/sbt/blob/1.x/CONTRIBUTING.md) guidelines

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -678,7 +678,10 @@ object BuiltinCommands {
   def act: Command = Command.customHelp(Act.actParser, actHelp)
 
   def actHelp: State => Help =
-    s => CommandStrings.showHelp ++ CommandStrings.multiTaskHelp ++ keysHelp(s)
+    s =>
+      CommandStrings.showHelp ++ CommandStrings.printHelp ++ CommandStrings.multiTaskHelp ++ keysHelp(
+        s
+    )
 
   def keysHelp(s: State): Help =
     if (Project.isProjectLoaded(s))
@@ -824,7 +827,7 @@ object BuiltinCommands {
         if (version != app.id.version()) {
           state.log.warn(s"""sbt version mismatch, current: ${app.id
             .version()}, in build.properties: "$version", use 'reboot' to use the new value.""")
-        }
+      }
     )
   }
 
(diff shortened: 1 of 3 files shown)


muuki88:
Thanks for your pull request :hugs: 

The usage described in the issue still wouldn't work as there are always initial logging outputs, right?
```bash
export SBT_PROJECT_VERSION=$(sbt 'show version')
```
or does it?

eed3si9n:
This would be a change in the behavior, so it might end up breaking other people's build.
Maybe we should either make a new similarly named command or implement JVM flag that we can pass in from the bash/bat script (similar to `-no-colors`)?

Falmarri:
@muuki88 If you run with `-error` it will suppress logging outputs. I hadn't considered having it override all logging on its own

@eed3si9n open to suggestions/requirements =]

muuki88:
@Falmarri since when version does `-error` work? I tried with `1.1.0` and I still see a lot of logging output.

muuki88:
@eed3si9n I like the jvm flag suggestion. Actually I search for a `--quiet` flag that silences everything aside from the `show` command. WDYT @Falmarri ?

eed3si9n:
@muuki88 `--error` works well for sbt 1.2.3:

Try the following on sbt/sbt:

```
$ sbt --error "eval println(\"hello\")"
hello
```

Falmarri:
@muuki88 the -error was fixed somewhere in the 1.1.x releases. Not sure where exactly, but it definitely works on 1.2.x

Falmarri:
@eed3si9n I pushed a change that adds this new version as the `print` command, rather than `show`. Thoughts?

eed3si9n (approved):
LGTM

muuki88 (approved):

dwijnand:
1. This isn't documented anywhere
2. It doesn't work as soon as your build contains more than one project (and you follow the custom of having the root project aggregate all modules).  Even using `ThisBuild` won't work, due to #5349...

PR #4343, opened 2018-09-04
Publish JUnit reports using a compliant filename

### LINKED ISSUE #3150: JUnitXmlTestsListener doesn't follow recommendations for filenames
JUnit XML output doesn't prepend files with `TEST-` as per JUnit recommendation.

## steps
Create a new application with at least one test class, like `com.example.HelloWorld`
Run `sbt test`

## problem
The JUnit xml file will be named `com.example.HelloWorld.xml`. This causes tools like SonarQube or Jenkins to not be able to pickup the files unless we first rename them.

The code in [JUnitXmlTestsListener.scala](https://github.com/sbt/sbt/blob/1.0.x/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala#L180)
```scala
  private def writeSuite() = {
    val file = new File(targetDir, s"${normalizeName(testSuite.value.name)}.xml").getAbsolutePath
    // TODO would be nice to have a logger and log this with level debug
    // System.err.println("Writing JUnit XML test report: " + file)
    XML.save(file, testSuite.value.stop(), "UTF-8", true, null)
  }
```

## expectation
The JUnit xml file should be named `TEST-com.example.HelloWorld.xml`

## notes
- [Maven Surefire](http://maven.apache.org/surefire/maven-surefire-plugin/) defaults to use `TEST-*.xml`
- [ANT JUnitreport](https://ant.apache.org/manual/Tasks/junitreport.html) defaults to use `TEST-*.xml`
- [Jenkins JUnit Plugin](https://github.com/jenkinsci/junit-plugin/blob/master/src/main/java/hudson/tasks/junit/SuiteResult.java#L350) is hardcoded to use `TEST-*.xml`
- [SonarQube Sonar-Java plugin](https://github.com/SonarSource/sonar-java/blob/7673a3ac725c89beaf95be0b9a93867a3a79de34/java-surefire/src/main/java/org/sonar/plugins/surefire/SurefireJavaParser.java#L76) is hardcoded to use `TEST-*.xml`

ignasi35:
Fixes #3150 by duplicating the reports generated so given a test `com.example.MyTest.java` the following will be created:

 - `com.example.MyTest.xml`
 - `TEST-com.example.MyTest.xml`

I used this approach so tooling that expects the report to not have the `TEST-` prefix still works and we could mark that filename as deprecated.

### DIFF
--- a/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
+++ b/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
@@ -226,9 +226,11 @@ class JUnitXmlTestsListener(val outputDir: String, logger: Logger) extends Tests
     d.truncatedTo(ChronoUnit.SECONDS).format(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
 
   private def writeSuite() = {
-    val file = new File(targetDir, s"${normalizeName(withTestSuite(_.name))}.xml").getAbsolutePath
+    val legacyFile = new File(targetDir, s"${normalizeName(withTestSuite(_.name))}.xml").getAbsolutePath
+    val file = new File(targetDir, s"TEST-${normalizeName(withTestSuite(_.name))}.xml").getAbsolutePath
     // TODO would be nice to have a logger and log this with level debug
     // System.err.println("Writing JUnit XML test report: " + file)
+    XML.save(legacyFile, withTestSuite(_.stop()), "UTF-8", true, null)
     XML.save(file, withTestSuite(_.stop()), "UTF-8", true, null)
     testSuite.remove()
   }


ignasi35:
I think creating more files causes some `scripted` tests to fail (because the filecount for generated files increased?)

ignasi35:
I can't reproduce the errors locally. :-(

eed3si9n (approved):
Thanks @ignasi35!

yanns:
It'd have been good to provide a way to create only the files with the new names.
Now test summaries are reporting 2 times more tests 😅

ignasi35:
@yanns I've created https://github.com/sbt/sbt/issues/4451 with your idea!

PR #4346, opened 2018-09-05
Fix second part of #4091: server responses should not be broadcasted

smarter:
Change CommandExchange#publishEvent to not broadcast ExecStatusEvent and
instead check the channelName, this matches what was already done in
CommandExchange#publishEventMessage


smarter:
@eed3si9n I've opened this against `develop` since that seems to be the workflow currently, but I'd appreciate it if this could get into the next 1.2 too.

eed3si9n (approved):

PR #2773, opened 2016-10-09
Allow single-letter aliases

kzys:
Fixes #1074.

PR #4355, opened 2018-09-14
Allow single-letter aliases

### LINKED ISSUE #1074: + fails with single-letter-named aliases
```
[sbt-0-13-1]> alias z = ; show scalaVersion ; help scalaVersion
[sbt-0-13-1]> z
[info] 2.10.4-RC1
The version of Scala used for building.
[sbt-0-13-1]> + z
[error] Expected any character
[error] + z
[error]    ^
```

Shouldn't at least `alias` command warn a user when a single-letter name is used to define an alias?

eed3si9n:
Fixes #1074

This replaces an old PR that's been around for a while (https://github.com/sbt/sbt/pull/2773)

### DIFF
--- a/sbt/src/sbt-test/actions/add-alias/test
+++ b/sbt/src/sbt-test/actions/add-alias/test
@@ -1,2 +1,3 @@
 > demo-success
 -> demo-failure
+> +z
--- a/sbt/src/sbt-test/actions/add-alias/test
+++ b/sbt/src/sbt-test/actions/add-alias/test
@@ -1,3 +1,4 @@
 > demo-success
 -> demo-failure
 > +z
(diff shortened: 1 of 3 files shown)


sbt/src/sbt-test/actions/add-alias/test
```diff
@@ -1,2 +1,3 @@
 > demo-success
 -> demo-failure
+> +z
```
muuki88, line 3: Does this work without `+`, too?

muuki88 (dismissed):
LGTM :smile: 

What puzzled me at the beginning is the `Cross` class. Maybe we could add a small scaladoc what this module provides?

PR #4356, opened 2018-09-14
Remove old warning about global dir

### LINKED ISSUE #1054: Suggested improvement for unversioned global SBT directory warning
If you do not create a ~/.sbt/0.13/ directory, SBT 0.13+ runs in compatibility mode, and issues this warning that suggests you should create the directory:

```
[warn] The global sbt directory is now versioned and is located at /Users/mslinn/.sbt/0.13.
```

The sentence actually says the opposite of what was intended. A better wording might be:

```
There is no versioned global SBT directory. If you want to make one, you could put it at ~/.sbt/0.13/
```

Also, I suggest that this be removed:

```
[warn]   The global sbt directory may be changed via the sbt.global.base system property.
```

Changing sbt.global.base leads to confusion and is rarely needed. SBT is complicated enough without encouraging users to stray from the straight and narrow path.

eed3si9n:
Fixes #1054

### DIFF
--- a/main/src/main/scala/sbt/BuildPaths.scala
+++ b/main/src/main/scala/sbt/BuildPaths.scala
@@ -44,17 +44,7 @@ object BuildPaths {
 
   def getGlobalBase(state: State): File = {
     val default = defaultVersionedGlobalBase(binarySbtVersion(state))
-    def getDefault = { checkTransition(state, default); default }
-    getFileSetting(globalBaseDirectory, GlobalBaseProperty, getDefault)(state)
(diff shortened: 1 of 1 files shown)


muuki88 (dismissed):
It removes the transition warning being emitted, but `globalDirTransitionWarning` is still present as far as I can see. We should remove that as well, right?

eed3si9n:
Yeah, you're right.

eed3si9n:
Merging since I addressed the review.

Issue #4298, opened 2018-08-04
sbt.global.base doesn't work properly with '~'


erikerlandson:
- [x ] I've read the [CONTRIBUTING](https://github.com/sbt/sbt/blob/1.x/CONTRIBUTING.md) guidelines

## steps
```sh
$ sbt -Dsbt.global.base=~/.sbt/foo
```
## problem

global sbt file(s) under `~/.sbt/foo` are not read

## expectation

sbt should read the global information under `~/.sbt/foo`

## notes

sbt version: 1.2.0

workaround: use fully qualified home directory path, e.g. `/home/user/.sbt/foo`

eed3si9n:
`~` is a feature of Bash - https://www.gnu.org/software/bash/manual/html_node/Tilde-Expansion.html

Maybe it's friendlier if sbt expanded it for you, but I'd say it's more of a nice-to-have feature than a bug.

eed3si9n:
Having said that, this might be fairly easy to add, so if @erikerlandson or someone else to contribute to it, this might be a good first issue.

The relevant code is https://github.com/sbt/sbt/blob/v1.2.1/main/src/main/scala/sbt/BuildPaths.scala#L89-L92

```scala
  def getFileProperty(name: String): Option[File] = Option(System.getProperty(name)) flatMap {
    path =>
      if (path.isEmpty) None else Some(new File(path))
  }
```

That gets used by `getGlobalBase`.

jvican:
Last time I looked at this you could not used a relative path as sbt global base, either, it had to be absolute.

erikerlandson:
possibly I just read too much into some other docs.  I've seen paths like `~/.sbt/1.0/plugins/gpg.sbt`, and assumed `~` was allowed, but wasn't in the immediate context of setting `sbt.global.base`

Thanks for explaining!

kai-chi:
this can be closed - fixed with #4367

erikerlandson:
That is great! I'll close this.

PR #4367, opened 2018-09-19
expand tilde prefix

kai-chi:
- [x] I've signed the CLA

#4298 - Expands the tilde-prefix using environmental variables.


eed3si9n (changes requested):
Thanks for the contribution.
Could you add a unit test for this please?

eed3si9n (changes requested):
Thanks!

eed3si9n (approved):

PR #4376, opened 2018-09-21
Fixes #4375

### LINKED ISSUE #4375: --addPluginSbtFile is ignored except for a global plugin project

## steps

This is based on working with IntelliJ IDEA 2018.3 EAP with Intellij Scala nightly plugin.
However, it is not necessary for reproducing this problem.
Below, I'm simply emulating what Intellij IDEA does when importing a Scala 2.12.6 / SBT 1.2.3 project.
For Intellij-specific details, see: https://youtrack.jetbrains.com/issue/SCL-14402#

```bash
mkdir tmp
cd tmp
cat > idea.sbt << EOF
// Generated by IntelliJ-IDEA Scala plugin.
// Adds settings when starting sbt from IDEA.
// Manual changes to this file will be lost.
scala.collection.Seq(
  resolvers += Resolver.file("intellij-scala-plugin", file(Path.userHome.absolutePath + "/.IntelliJIdea2018.3/config/plugins/Scala/repo"))(Resolver.ivyStylePatterns),
  addSbtPlugin("org.jetbrains" % "sbt-structure-extractor" % "2018.2.1"),
  addSbtPlugin("org.jetbrains" % "sbt-idea-shell" % "2017.2")
)
EOF
mkdir proj1
cd proj1
touch build.sbt
sbt -v -sbt-version 1.2.3 --addPluginsSbtFile=`pwd`/../idea.sbt
```

## problem

The above results in a behavior like this:

```
# Executing command line:
java
-Xms1024m
-Xmx1024m
-XX:ReservedCodeCacheSize=128m
-XX:MaxMetaspaceSize=256m
-Dsbt.version=1.2.3
-jar
/usr/share/sbt/bin/sbt-launch.jar
--addPluginSbtFile=/tmp/idea.sbt

[info] Loading project definition from /home/rouquett/tmp/proj1/project
[info] Loading settings for project proj1 from build.sbt ...
[info] Set current project to proj1 (in build file:/home/rouquett/tmp/proj1/)
[info] sbt server started at local:///home/rouquett/.sbt/1.0/server/2a7c52b9eb78c2740354/sock
sbt:proj1> plugins
In file:/home/rouquett/tmp/proj1/
	sbt.plugins.IvyPlugin: enabled in proj1
	sbt.plugins.JvmPlugin: enabled in proj1
	sbt.plugins.CorePlugin: enabled in proj1
	sbt.ScriptedPlugin
	sbt.plugins.SbtPlugin
	sbt.plugins.JUnitXmlReportPlugin: enabled in proj1
	sbt.plugins.Giter8TemplatePlugin: enabled in proj1
sbt:proj1> commands
[info] * 
```

## expectation

```
# Executing command line:
java
-Xms512m
-Xmx1536m
-Xss2m
-jar
/home/rouquett/.sbt/launchers/1.2.3/sbt-launch.jar
--addPluginSbtFile=/tmp/idea.sbt

[info] Updated file /home/rouquett/tmp/proj1/project/build.properties: set sbt.version to 1.2.3
[info] Loading settings for project global-plugins from idea.sbt ...
[info] Loading settings for project global-plugins from idea.sbt ...
[info] Loading global plugins from /home/rouquett/.sbt/1.0/plugins
[info] Updating ProjectRef(uri("file:/home/rouquett/.sbt/1.0/plugins/"), "global-plugins")...
[info] Done updating.
[info] Loading project definition from /home/rouquett/tmp/proj1/project
[info] Updating ProjectRef(uri("file:/home/rouquett/tmp/proj1/project/"), "proj1-build")...
[info] Done updating.
[info] Loading settings for project proj1 from build.sbt ...
[info] Set current project to proj1 (in build file:/home/rouquett/tmp/proj1/)
[info] sbt server started at local:///home/rouquett/.sbt/1.0/server/2a7c52b9eb78c2740354/sock
sbt:proj1> plugins
In file:/home/rouquett/tmp/proj1/
	sbt.plugins.IvyPlugin: enabled in proj1
	sbt.plugins.JvmPlugin: enabled in proj1
	sbt.plugins.CorePlugin: enabled in proj1
	sbt.ScriptedPlugin
	sbt.plugins.SbtPlugin
	sbt.plugins.JUnitXmlReportPlugin: enabled in proj1
	sbt.plugins.Giter8TemplatePlugin: enabled in proj1
	org.jetbrains.sbt.StructurePlugin: enabled in proj1
	org.jetbrains.sbt.IdeaShellPlugin: enabled in proj1
sbt:proj1> commands
[info] * SimpleCommand(idea-shell)
sbt:proj1> 
```

## notes

sbt version: 1.2.3

## observations

I ran the above with the debugger, i.e.:
```
sbt -J-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005 -sbt-version 1.2.3 --addPluginSbtFile=`pwd`/../idea.sbt
```
and used IntelliJ to attach the debugger.

In `sbt.internal.Load`, the execution reaches this point:

```
      // Load all config files AND finalize the project at the root directory, if it exists.
      // Continue loading if we find any more.
      newProjects match {
        case Seq(next, rest @ _*) =>
          log.debug(s"[Loading] Loading project ${next.id} @ ${next.base}")
...
        case Nil if makeOrDiscoverRoot =>
          log.debug(s"[Loading] Scanning directory ${buildBase}")
          val auto =
            if (context.globalPluginProject)
              AddSettings.seq(AddSettings.defaultSbtFiles, AddSettings.sbtFiles(extraSbtFiles: _*))
            else AddSettings.defaultSbtFiles
```

When evaluating the assignment to the `auto` variable,
`extraSbtFiles` is a sequence of 1 file, the argument to the `--addPluginSbtFile`.
However, since `context.globalPluginProject` is false, it is effectively ignored.
Therefore, the `extraSbtFiles` are not loaded!

NicolasRouquette:

I notice that there is not a test project to exercise the `--addPluginSbtFile` option.
I think it would make sense to add one in `sbt/src/sbt-test/project`; however, I'm not sure if the scripted machinery supports this, and if it does, how to write such a test.

### DIFF
--- a/main/src/main/scala/sbt/internal/Load.scala
+++ b/main/src/main/scala/sbt/internal/Load.scala
@@ -930,10 +930,7 @@ private[sbt] object Load {
           )
         case Nil if makeOrDiscoverRoot =>
           log.debug(s"[Loading] Scanning directory ${buildBase}")
-          val auto =
-            if (context.globalPluginProject)
-              AddSettings.seq(AddSettings.defaultSbtFiles, AddSettings.sbtFiles(extraSbtFiles: _*))
-            else AddSettings.defaultSbtFiles
+          val auto = AddSettings.seq(AddSettings.defaultSbtFiles, AddSettings.sbtFiles(extraSbtFiles: _*))
           discover(auto, buildBase) match {
             case DiscoveredProjects(Some(root), discovered, files, generated) =>
               log.debug(


main/src/main/scala/sbt/internal/Load.scala
```diff
@@ -930,10 +930,7 @@ private[sbt] object Load {
           )
         case Nil if makeOrDiscoverRoot =>
           log.debug(s"[Loading] Scanning directory ${buildBase}")
-          val auto =
-            if (context.globalPluginProject)
-              AddSettings.seq(AddSettings.defaultSbtFiles, AddSettings.sbtFiles(extraSbtFiles: _*))
-            else AddSettings.defaultSbtFiles
+          val auto = AddSettings.seq(AddSettings.defaultSbtFiles, AddSettings.sbtFiles(extraSbtFiles: _*))
```
eed3si9n, line 933: loading mechanism gets called multiple times at different level, so the above if expression is meant to load plugin files only for the metabuild.

NicolasRouquette: How many levels are there? What is the intended level for loading plugin files then if it is not the metabuild level, which I presume corresponds to `context.globalPluginProject == true`?

eed3si9n: sbt build is recursive, and the plugin loading is just build loading (build.sbt) inside `project/` directory. The directory structure can go on. In addition to the working directory, there's also machine-wide global settings, and global plugins. Each subproject can also have its own build.sbt.

NicolasRouquette: yes, i understand this principle of sbt; however, I'm still puzzled why the `--addPluginSbtFile=...` option has become ineffective. I thought, perhaps incorrectly, that I had found the reason but perhaps there's something more to it.

Bottom line is simple: when we do: `sbt -sbt-version=1.2.3 --addPluginSbtFile=foo.sbt`, when and where should this option be processed? It's obviously not happening in my environment and I'm trying to figure out why. I've tried lots of things including removing `~/.sbt` entirely and `~/.m2/repository` without luck. The only workaround I found is to move `foo.sbt` to `./project/foo.sbt`.

Issue #4333, opened 2018-08-27
Add JDK 11 testing


eed3si9n:
For all modules, we should start testing on JDK 11.

- [x] https://github.com/sbt/sbt/pull/4389
- [x] https://github.com/sbt/zinc/issues/540
- [x] https://github.com/sbt/librarymanagement/pull/263
- [x] https://github.com/sbt/util/pull/181
- [x] https://github.com/sbt/io/pull/181

He-Pin:
This is a must done now~!

eed3si9n:
All projects are now tested using AdoptOpenJDK JDK 11.

PR #4377, opened 2018-09-22
Scala 2.12.7 and linesIterator

eed3si9n:
Ref #4333

### DIFF
--- a/main-settings/src/main/scala/sbt/InputTask.scala
+++ b/main-settings/src/main/scala/sbt/InputTask.scala
@@ -27,7 +27,7 @@ final class InputTask[T] private (val parser: State => Parser[Task[T]]) {
         Parser.parse(in, parser(s)) match {
           case Right(v) => Parser.success(v)
           case Left(msg) =>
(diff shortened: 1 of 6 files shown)


main-settings/src/main/scala/sbt/InputTask.scala
```diff
@@ -27,7 +28,7 @@ final class InputTask[T] private (val parser: State => Parser[Task[T]]) {
         Parser.parse(in, parser(s)) match {
           case Right(v) => Parser.success(v)
           case Left(msg) =>
-            val indented = msg.lines.map("   " + _).mkString("\n")
+            val indented = new StringOps(msg).lines.map("   " + _).mkString("\n")
```
ritschwumm: i'd put all those repeated `new StringOps(msg).lines` in a utility function where they are easy to find  later

He-Pin: https://github.com/scala/bug/issues/11125

eed3si9n: I guess I'll just wait till 2.12.7 comes out.

Issue #4370, opened 2018-09-20
sbt/exec gives never sends a JSON-RPC Response when a command throws an exception


Duhemm:
- [x] I've read the [CONTRIBUTING](https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md) guidelines

## steps
1. Start an sbt server instance
```sh
$ mkdir project
$ echo "sbt.version=1.2.3" > project/build.properties
$ echo 'commands += Command.command("hello") { state => ??? }' >> build.sbt
$ sbt
```

2. Connect to sbt server and run `hello`:
```sh
$ tee | nc -U /Users/martin/.sbt/1.0/server/e7ba14344db70b8f0136/sock
{ "jsonrpc": "2.0", "id": 1, "method": "sbt/exec", "params": { "commandLine": "hello" } }
Content-Length: 89
Content-Type: application/vscode-jsonrpc; charset=utf-8

{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":4,"message":"Processing"}}
```

## problem
We never get a JSON-RPC `Response`.

## expectation
We get a JSON-RPC `Response` indicating that an error occurred:
```
{ "jsonrpc": "2.0", "id": 1, "method": "sbt/exec", "params": { "commandLine": "hello" } }
Content-Length: 89
Content-Type: application/vscode-jsonrpc; charset=utf-8

{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":4,"message":"Processing"}}
Content-Length: 61
Content-Type: application/vscode-jsonrpc; charset=utf-8

{"jsonrpc":"2.0","id":1,"error":{"code":-33000,"message":""}}
```

## notes
Input is processed here: https://github.com/sbt/sbt/blob/550c068b9829f8b691f046e87cd1023d87735973/main/src/main/scala/sbt/MainLoop.scala#L146-L166

If an exception is thrown in `Command.process`, we'll never get to publish the `doneEvent`.

`MainLoop.next` catches the exceptions, but it's too late to respond to the clients:
https://github.com/sbt/sbt/blob/550c068b9829f8b691f046e87cd1023d87735973/main/src/main/scala/sbt/MainLoop.scala#L138-L144

sbt version: 1.2.3

PR #4378, opened 2018-09-24
Fix: #4370

andreaTP:

This is a trivial fix for #4370 

I have still two observations:
 1. A server can crash at any point in time, I think is responsibility of the client to watch if the server is alive and behave accordingly
 2. The testing infrastructure for the LSP Server is extremely sensitive, I hacked it around to keep things working, but it will be worth to understand how we can do better.


eed3si9n:
Nice. Thanks @andreaTP!

andreaTP:
finalized and already squashed, ready for review.

eed3si9n (approved):

PR #4384, opened 2018-09-26
Implementing cancellation requests for LSP server

### LINKED ISSUE #4373: [sbt-server] support for cancellations
Started developing a thin client for Sbt (https://gitlab.com/andreaTP/sbtcli) and it turned out that there is no support for cancelling "running" tasks.

LSP provide a specification for it: https://microsoft.github.io/language-server-protocol/specification#cancelRequest

I would imagine that a CLI user would like to `Ctrl-C` a running task and it's interrupted on the server.

andreaTP:

After a number of attempts this is the best solution I've found so far, still, open to suggestions.
This will fix #4373 

(Thanks for pushing on this @TrustNoOne :-) )

### DIFF
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -14,6 +14,7 @@ import sbt.librarymanagement.{ Resolver, UpdateReport }
 
 import scala.concurrent.duration.Duration
 import java.io.File
+import java.util.concurrent.atomic.AtomicReference
 import Def.{ dummyState, ScopedKey, Setting }
 import Keys.{
   Streams,
@@ -377,6 +378,8 @@ object EvaluateTask {
       (dummyRoots, roots) :: (Def.dummyStreamsManager, streams) :: (dummyState, state) :: dummies
     )
 
+  val currentlyRunningEngine: AtomicReference[(State, RunningTaskEngine)] = new AtomicReference()
+
   def runTask[T](
       root: Task[T],
       state: State,
@@ -432,11 +435,15 @@ object EvaluateTask {
         shutdown()
       }
     }
+    currentlyRunningEngine.set((state, runningEngine))
     // Register with our cancel handler we're about to start.
     val strat = config.cancelStrategy
     val cancelState = strat.onTaskEngineStart(runningEngine)
     try run()
-    finally strat.onTaskEngineFinish(cancelState)
+    finally {
+      strat.onTaskEngineFinish(cancelState)
+      currentlyRunningEngine.set(null)
+    }
   }
 
   private[this] def storeValuesForPrevious(
--- a/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
@@ -13,12 +13,14 @@ import sjsonnew.JsonFormat
 import sjsonnew.shaded.scalajson.ast.unsafe.JValue
 import sjsonnew.support.scalajson.unsafe.Converter
 import sbt.protocol.Serialization
-import sbt.protocol.{ SettingQuery => Q }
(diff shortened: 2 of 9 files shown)


main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
```diff
@@ -80,6 +81,36 @@ private[sbt] object LanguageServerProtocol {
               import sbt.protocol.codec.JsonProtocol._
               val param = Converter.fromJson[Q](json(r)).get
               onSettingQuery(Option(r.id), param)
+            case r: JsonRpcRequestMessage if r.method == "sbt/cancelRequest" =>
+              val param = Converter.fromJson[CancelRequestParams](json(r)).get
+              try {
+                val (state, runningEngine) = EvaluateTask.currentlyRunningEngine.get
+                // remove hotspring unicode added character
+                val execId = state.currentCommand.get.execId.get.substring(1)
```
eed3si9n: Hotspring is only added for numbers.

main/src/main/scala/sbt/EvaluateTask.scala
```diff
@@ -432,6 +435,7 @@ object EvaluateTask {
         shutdown()
       }
     }
+    currentlyRunningEngine.set((state, runningEngine))
```
eed3si9n, line 438: This should be unset after run() so it doesn't hold on to the reference.

eed3si9n (changes requested):
Some minor comments, but overall looks good, I think.

main/src/main/scala/sbt/EvaluateTask.scala
andreaTP: sounds good

main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
andreaTP: got it, thanks

andreaTP:
Addressed the comments, added a test for "string ids" cancellation (as opposed to the numeric once) and already squashed commits.

andreaTP:
not sure why CI failed one test ... I closed and re-opened to re-trigger travis.

main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
```diff
@@ -80,6 +81,37 @@ private[sbt] object LanguageServerProtocol {
               import sbt.protocol.codec.JsonProtocol._
               val param = Converter.fromJson[Q](json(r)).get
               onSettingQuery(Option(r.id), param)
+            case r: JsonRpcRequestMessage if r.method == "sbt/cancelRequest" =>
+              val param = Converter.fromJson[CancelRequestParams](json(r)).get
+              try {
+                val (state, runningEngine) = EvaluateTask.currentlyRunningEngine.get
```
eed3si9n: This could return `null`, so maybe put this in `Option(EvaluateTask.currentlyRunningEngine.get)`?

eed3si9n (changes requested):

main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
andreaTP: actually this is expected, any `NonFatal` error (like NPE) that happens within the try block is catch and return to the user a proper error  answer.

andreaTP: there are more unsafe get also in the following lines, do you prefer to handle all of them one-by-one?

eed3si9n: However you want to handle it is fine, but we should handle them.

andreaTP:
Implemented a more fine-graned error handling and already squashed the commits,

eed3si9n (approved):
Thanks!

Issue #4374, opened 2018-09-21
[sbt-server] helper for completions


andreaTP:
Started building a thin client for Sbt (https://gitlab.com/andreaTP/sbtcli) and I wanted to implement some kind of completion for the interactive shell.

LSP define completion requests: https://microsoft.github.io/language-server-protocol/specification#textDocument_completion

it should be supported by the sbt-server to enable more user-friendly CLIs.

eed3si9n:
`TextDocumentPositionParams` takes `TextDocumentIdentifier`, which contains an URI.
We can make an URI convention for the shell. We can make up a URI convention for our use.
For example we can say `"sbt://shell"` means sbt shell like `chrome://about/`.

The context could be:

```json
{
  "triggerKind": 1,
  "triggerCharacter": "comp"
}
```

eed3si9n:
Fixed in https://github.com/sbt/sbt/pull/4397

PR #4385, opened 2018-09-27
[WIP] Basic completion support for LSP server

andreaTP:

This will stack on top of #4384 
I accidentally squashed everything together ...

@eed3si9n would you mind to take a look and comment on this approach?
My goal here is to have a "general" and basic support for completions that I will trigger at CLi startup and manage on my own.
More advanced stuff could come later on.

### DIFF
--- a/sbt/src/test/scala/testpkg/ServerSpec.scala
+++ b/sbt/src/test/scala/testpkg/ServerSpec.scala
@@ -47,6 +47,60 @@ class ServerSpec extends AsyncFreeSpec with Matchers {
         (s contains """"id":11""") && (s contains """"error":""")
       })
     }
+
+    "return basic completions on request" in withTestServer("events") { p =>
+      p.writeLine(
+        """{ "jsonrpc": "2.0", "id": 15, "method": "textDocument/completion", "params": { } }"""
+      )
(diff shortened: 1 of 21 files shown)


sbt/src/test/scala/testpkg/ServerSpec.scala
```diff
@@ -47,6 +47,60 @@ class ServerSpec extends AsyncFreeSpec with Matchers {
         (s contains """"id":11""") && (s contains """"error":""")
       })
     }
+
+    "return basic completions on request" in withTestServer("events") { p =>
+      p.writeLine(
+        """{ "jsonrpc": "2.0", "id": 15, "method": "textDocument/completion", "params": { } }"""
```
eed3si9n, line 53: See my comment in https://github.com/sbt/sbt/issues/4374#issuecomment-424921945.

andreaTP: I understand the ultimate goal of your proposal and make sense.
but I'm not sure we should go in that direction from day 0 ... having contextual completion is great for IDEs, not sure how much added value it has in this context considering the "lag" it will produce...

andreaTP:
Overtaken by: #4397

PR #4397, opened 2018-10-02
[sbt-server] LSP completions support

andreaTP:

This overtake #4384 . Now we do have complete support over completions as requested.
I've also added (but not used) the additional requested parameters by @eed3si9n

### DIFF
--- a/main-command/src/main/scala/sbt/State.scala
+++ b/main-command/src/main/scala/sbt/State.scala
@@ -18,7 +18,7 @@ import sbt.internal.util.{
   ExitHooks,
   GlobalLogging
 }
-import sbt.internal.util.complete.HistoryCommands
+import sbt.internal.util.complete.{ HistoryCommands, Parser }
 import sbt.internal.inc.classpath.ClassLoaderCache
 
 /**
@@ -52,6 +52,26 @@ final case class State(
       case Some(x) => x.source
       case _       => None
     }
+}
+
+/**
+ * Data structure extracted form the State Machine for safe observability purposes.
+ *
+ * @param currentExecId provide the execId extracted from the original State.
+ * @param combinedParser the parser extracted from the original State.
+ */
+private[sbt] final case class SafeState(
+    currentExecId: Option[String],
+    combinedParser: Parser[() => sbt.State]
+)
+
+private[sbt] object SafeState {
+  def apply(s: State) = {
+    new SafeState(
+      currentExecId = s.currentCommand.map(_.execId).flatten,
+      combinedParser = s.combinedParser
+    )
+  }
 }
 
 trait Identity {
--- a/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
@@ -13,7 +13,7 @@ import sjsonnew.JsonFormat
 import sjsonnew.shaded.scalajson.ast.unsafe.JValue
 import sjsonnew.support.scalajson.unsafe.Converter
 import sbt.protocol.Serialization
-import sbt.protocol.{ SettingQuery => Q, ExecStatusEvent }
+import sbt.protocol.{ SettingQuery => Q, ExecStatusEvent, CompletionParams => CP }
 import sbt.internal.protocol._
 import sbt.internal.protocol.codec._
 import sbt.internal.langserver._
@@ -135,6 +135,10 @@ private[sbt] object LanguageServerProtocol {
                 case NonFatal(e) =>
                   errorRespond("Cancel request failed")
               }
+            case r: JsonRpcRequestMessage if r.method == "sbt/completion" =>
+              import sbt.protocol.codec.JsonProtocol._
+              val param = Converter.fromJson[CP](json(r)).get
+              onCompletionRequest(Option(r.id), param)
           }
         }, {
           case n: JsonRpcNotificationMessage if n.method == "textDocument/didSave" =>
@@ -155,6 +159,7 @@ private[sbt] trait LanguageServerProtocol extends CommandChannel { self =>
   protected def setInitialized(value: Boolean): Unit
   protected def log: Logger
   protected def onSettingQuery(execId: Option[String], req: Q): Unit
+  protected def onCompletionRequest(execId: Option[String], cp: CP): Unit
 
   protected lazy val callbackImpl: ServerCallback = new ServerCallback {
     def jsonRpcRespond[A: JsonFormat](event: A, execId: Option[String]): Unit =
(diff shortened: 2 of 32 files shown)


andreaTP:
this will fix #4374

main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
```diff
@@ -135,6 +136,47 @@ private[sbt] object LanguageServerProtocol {
                 case NonFatal(e) =>
                   errorRespond("Cancel request failed")
               }
+            case r: JsonRpcRequestMessage if r.method == "textDocument/completion" =>
```
olafurpg: Is this to complete commands for the sbt shell of complete Scala source code? `textDocument/completion` is meant for completion in text documents. This is this for sbt shell completions then I think it makes sense to use a different method name like `sbt/completion`. It can still use the LSP data structures like `CompletionContext` https://microsoft.github.io/language-server-protocol/specification#textDocument_completion

olafurpg: Very cool feature btw! 😄

andreaTP: @olafurpg correct, this is for sbt interactive shell completions, I don't mind much on the `method` name.. :smile:
Sbt completions are kind of complicated since they flow down to user-code class names for example (i.e. `testOnly`).
I leave the final word to @eed3si9n

eed3si9n: On Gitter chat I wrote:

> eugene yokota eed3si9n Sep 28 05:29
> CompletionParams extends TextDocumentPositionParams
so in the request it's supposed to containt the position of which file the user's on and where
since we are using for shell purpose, I think we need to make up a non-existent URL for the shell
>  I guess textDocument: TextDocumentIdentifier; is but I don't want to assume it's for shell
since normally it's used for Scala source's completion

I think the editor interested in sbt shell completion could still use "textDocument/completion" but it should pass the URL `sbt://shell`, and sbt server should respond only for that.

main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
```diff
@@ -135,6 +136,47 @@ private[sbt] object LanguageServerProtocol {
...
+              val param = Converter.fromJson[CompletionParams](json(r)).get
+
+              try {
+                Option(EvaluateTask.lastEvaluatedState.get) match {
```
eed3si9n: Probably I should've commented this on the cancellation PR too, but generally this class should be used as a glue code to interpret LSP context, and then quickly pass it along to NetworkChannel etc.
See for example:

```scala
             case r: JsonRpcRequestMessage if r.method == "sbt/exec" =>
               val param = Converter.fromJson[SbtExecParams](json(r)).get
               appendExec(Exec(param.commandLine, Some(r.id), Some(CommandSource(name))))
               ()
```

In the same fashion, I think the parameters needed for shell completion should be put into a pseudo case class, and sent to a callback method like `appendShellCompletion(ShellCompletion(...))`. That should avoid the globally holding on to the state.

eed3si9n (changes requested):

main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
olafurpg: I think it would be clearer with a custom endpoint for this functionality only

```scala
// method: 'sbt/completion'
trait SbtCompletionParams {
  def query: String
}
trait SbtCompletionResponse {
  def items: List[String]
}
```

The way textDocument/completion method is used in PR is very different from how it's supposed to be used in LSP that I think it's more confusing than helpful.

- `sbt://shell` is not a valid URI for other requests such as `textDocument/didChange`
- `CompletionContext.triggerCharacter` is supposed to be character such as `.`
- server should return `ServerCapabilities.completionProvider = true` during the `initialize` handshake, and in this case that would mean VS Code would start sending completion requests for Scala source files as well.
- Traditionally the language server uses `TextDocumentPositionParams.position` to know the "query" of the completion request

andreaTP: I tend to agree on @olafurpg point of view, makes totally sense.

andreaTP:
ok, apart from the discussion on method-name/params to be used, just wanna check with @eed3si9n that this implementation of the functionality itself is more close to an expected one.
In case I will try to align also the cancellation request accordingly in a new PR.

The key difference that prevents me from reusing the infrastructure set-up for `SettingQuery` is that settings are "static" after sbt is loaded (and they can be listed on static knowledge) but completions (e.g. on user classes) are dynamic and heavily depends on current state machine status.

main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
eed3si9n: Yes. These are actually good points. Let's do sbt specific completion command.

andreaTP:
I improved a little the server testing infrastructure to properly timeout.
Addressed all the comments and already squashed the commits.

Ready for an (hopefully) final round of reviews. :smile:

andreaTP:
Travis is correct, it's failing because of:
```method onCompletionRequest```
not sure what I can do about this ....

File: main-command/src/main/scala/sbt/State.scala [hunk 3]
```diff
@@ -54,6 +54,20 @@ final case class State(
     }
 }
 
+final case class SafeState(
```
eed3si9n: This should be hidden to `private[sbt]`.

eed3si9n (changes requested):

main-command/src/main/scala/sbt/State.scala
eed3si9n: Also it would be good to put Scaladoc comment on the purpose of this datatype.

eed3si9n: Instead of `SafeState` should be renamed to `PreviousParser`

andreaTP: I wanted to keep here all the (state machine safe) information needed by the language Server, and reuse this class also for cancellations refactoring.

main-command/src/main/scala/sbt/State.scala
```diff
@@ -54,6 +54,26 @@ final case class State(
     }
 }
 
+/**
+ * Data structure extracted form the State Machine for safe observability purposes.
+ *
+ * @param currentExecId provide the execId extracted from the original State.
+ * @param combinedParser the parser extracted from the original State.
+ */
+private[sbt] final case class SafeState(
```
eed3si9n, line 63: Note that we're gonna be renaming this later.

eed3si9n (approved):

PR #4409, opened 2018-10-06
Include both explicit plugin resolves and boot resolvers

### LINKED ISSUE #4408: Compiler bridge cannot be resolved out of sbt-maven-snapshot
## steps

Use a nightly to compile something.

## problem

Compiler bridge is not found.

## expectation

Compiles.

## notes

Compiler bridge does not look into sbt-maven-snapshot.

Ref https://github.com/sbt/sbt/issues/304
6015d082aa5bc40dad8b8edbe1893dc060f52365

eed3si9n:
Fixes #4408

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -198,8 +198,9 @@ object Defaults extends BuildCommon {
       scalaOrganization :== ScalaArtifacts.Organization,
       scalaArtifacts :== ScalaArtifacts.Artifacts,
       sbtResolver := {
-        if (sbtVersion.value endsWith "-SNAPSHOT") Classpaths.sbtIvySnapshots
-        else Classpaths.typesafeReleases
+        val v = sbtVersion.value
+        if (v.endsWith("-SNAPSHOT") || v.contains("-bin-")) Classpaths.sbtMavenSnapshots
+        else Resolver.DefaultMavenRepository
       },
       crossVersion :== Disabled(),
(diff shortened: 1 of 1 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2289,13 +2290,20 @@ object Classpaths {
...
             val explicit = buildStructure.value
               .units(thisProjectRef.value.build)
               .unit
               .plugins
               .pluginData
               .resolvers
-            explicit orElse bootRepositories(appConfiguration.value) getOrElse externalResolvers.value
+            // https://github.com/sbt/sbt/issues/4408
+            (explicit, boot) match {
+              case (Some(x), Some(boot)) => (x.toVector ++ boot.toVector).distinct
```
eatkins: This really doesn't matter, but this would be _slightly_ more clear if `x` was renamed `explicit` since you're already shadowing `boot` anyway.

eatkins (dismissed):
LGTM with or without superficial comment.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Shadowing was unintentional. Maybe I'll rename that to b.

eatkins: In that case, I might suggest `ex` and `b`, just because `x` makes me think of a coordinate (I used to do a lot of scientific computing).

muuki88 (dismissed):

PR #4410, opened 2018-10-07
ThisBuild / semanticdbEnabled

eed3si9n:
This enables SemanticDB build-wide with a single line:

```scala
ThisBuild / scalaVersion := "2.12.7"
ThisBuild / semanticdbEnabled := true

```

SemanticDB currently requires per-subproject settings. This makes it easier to enable it build-wide for multi-project build.

Another motivation is so various tool chains that need SemanticDB to coordinate whether it is available or not.

### DIFF
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -243,6 +243,12 @@ object Keys {
   val scalaArtifacts = settingKey[Seq[String]]("Configures the list of artifacts which should match the Scala binary version").withRank(CSetting)
   val enableBinaryCompileAnalysis = settingKey[Boolean]("Writes the analysis file in binary format")
   val crossJavaVersions = settingKey[Seq[String]]("The java versions used during JDK cross testing").withRank(BPlusSetting)
+  val semanticdbEnabled = settingKey[Boolean]("Enables SemanticDB Scalac plugin").withRank(CSetting)
+  val semanticdbCompilerPlugin = settingKey[ModuleID]("SemanticDB Scalac plugin").withRank(CSetting)
+  val semanticdbVersion = settingKey[String]("SemanticDB version").withRank(CSetting)
+  val semanticdbIncludeInJar = settingKey[Boolean]("Include *.semanticdb files in published artifacts").withRank(CSetting)
+  val semanticdbTargetRoot = settingKey[File]("The output directory to produce META-INF/semanticdb/**/*.semanticdb files").withRank(CSetting)
+  val semanticdbOptions = settingKey[Seq[String]]("The Scalac options introduced for SemanticDB").withRank(CSetting)
 
   val clean = taskKey[Unit]("Deletes files produced by the build, such as generated sources, compiled classes, and task caches.").withRank(APlusTask)
   val console = taskKey[Unit]("Starts the Scala interpreter with the project classes on the classpath.").withRank(APlusTask)
--- a/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala
+++ b/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala
@@ -1,0 +1,58 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt
+package plugins
+
+import Keys._
+import sbt.librarymanagement.syntax._
+import sbt.librarymanagement.CrossVersion
+import Project.inConfig
+
+object SemanticdbPlugin extends AutoPlugin {
+  override def requires = JvmPlugin
+  override def trigger = allRequirements
+
+  override lazy val globalSettings = Seq(
+    semanticdbEnabled := false,
+    semanticdbIncludeInJar := false,
+    semanticdbOptions := List("-Yrangepos"),
+    semanticdbVersion := "4.0.0",
+    semanticdbCompilerPlugin := {
+      val v = semanticdbVersion.value
+      ("org.scalameta" % "semanticdb-scalac" % v).cross(CrossVersion.full)
+    }
+  )
+
+  override lazy val projectSettings = Seq(
+    allDependencies ++= {
+      val sdb = semanticdbEnabled.value
+      val m = semanticdbCompilerPlugin.value
+      if (sdb) List(Build0.compilerPlugin(m))
+      else Nil
+    }
+  ) ++ inConfig(Compile)(configurationSettings) ++ inConfig(Test)(configurationSettings)
+
+  lazy val configurationSettings: Seq[Def.Setting[_]] = List(
+    scalacOptions := {
+      val old = scalacOptions.value
+      val sdb = semanticdbEnabled.value
+      val sdbOptions = semanticdbOptions.value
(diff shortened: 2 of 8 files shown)


eed3si9n:
@olafurpg What do you think?

main/src/main/scala/sbt/plugins/SemanticDbPlugin.scala
```diff
@@ -0,0 +1,59 @@
...
+package sbt
+package plugins
+
+import Keys._
+import sbt.librarymanagement.syntax._
+import sbt.librarymanagement.CrossVersion
+import Project.inConfig
+import sbt.io.IO
+
+object SemanticDBPlugin extends AutoPlugin {
```
olafurpg: By convention we would write `SemanticdbPlugin` in code even if in written form it's "SemanticDB". For some reason I think `SemanticdbPlugin` reads nicer than `SemanticDBPlugin`. The same applies for setting names, `enableSemanticdb`, `semanticdbIncludeInJar`, `semanticdbVersion`, `semanticdbCompilerPlugin`

main/src/main/scala/sbt/plugins/SemanticDbPlugin.scala
```diff
@@ -0,0 +1,59 @@
...
+    semanticDBOptions := List("-Yrangepos"),
+    semanticDBVersion := "4.0.0",
+    semanticDBCompilerPlugin := {
+      val v = semanticDBVersion.value
+      ("org.scalameta" % "semanticdb-scalac" % v).cross(CrossVersion.full)
+    }
+  )
+
+  override lazy val projectSettings = Seq(
+    libraryDependencies ++= {
```
olafurpg: Does this work if users do `libraryDependencies :=` or `scalacOptions :=`? My experience with the previous sbt-scalafix version is that quite many builds do that and when it happens the user has no clue how to even debug the problem. The current sbt-scalafix installation instructions use `addCompilerPlugin` and `scalacOptions +=` and my impression so far is that people at least understand how that works so they can debug it.

olafurpg (commented):
I love the idea, thank you @eed3si9n for starting this. In fact, I was planning to discuss with you about something similar so I'm super happy you're already one step ahead of me 😄 

I really like having a boolean option to persist the semanticdb files outside the classDirectory 👍 

I also like separating the semanticdb options from scalac options so that if if semanticdb is disabled you don't include `-P:semanticdb:synthetics:on` settings.

The biggest concern for me is that users won't understand how this works so they will be confused when things fails. Any thoughts on that? For example, one risk is that user has an unsupported `scalaVersion` resulting in a resolution error for a dependency they can't see in build.sbt. We currently only publish semanticdb-scalac for the latest Scala 2.12 and 2.11 releases but I'm open to publish for more Scala versions, just need to figure out how to not make that explode our CI.

main/src/main/scala/sbt/Keys.scala
```diff
@@ -225,6 +225,12 @@ object Keys {
   val scalaArtifacts = settingKey[Seq[String]]("Configures the list of artifacts which should match the Scala binary version").withRank(CSetting)
   val enableBinaryCompileAnalysis = settingKey[Boolean]("Writes the analysis file in binary format")
   val crossJavaVersions = settingKey[Seq[String]]("The java versions used during JDK cross testing").withRank(BPlusSetting)
+  val enableSemanticDB = settingKey[Boolean]("Enables SemanticDB Scalac plugin").withRank(CSetting)
```
olafurpg: For consistency with sbt-scoverage, I propose `semanticdbEnabled := true`, https://github.com/scoverage/sbt-scoverage#how-to-use

main/src/main/scala/sbt/plugins/SemanticDbPlugin.scala
eed3si9n: Sounds good.

main/src/main/scala/sbt/plugins/SemanticDbPlugin.scala
eed3si9n: Good point. I can try `allDependencies`.

eed3si9n:
> I love the idea, thank you @eed3si9n for starting this. 

Awesome. Thanks for the review :)

> The biggest concern for me is that users won't understand how this works so they will be confused when things fails. Any thoughts on that? For example, one risk is that user has an unsupported `scalaVersion` resulting in a resolution error for a dependency they can't see in build.sbt. We currently only publish semanticdb-scalac for the latest Scala 2.12 and 2.11 releases but I'm open to publish for more Scala versions, just need to figure out how to not make that explode our CI.

I am concerned about that as well, but I think it's an issue intrinsic to the way semanticdb-scalac is currently distributed. Zinc avoids this problem by distributing the compiler bridge source JAR and compiling it on the fly. Not sure if you guys can take similar approach.

I am scoping `enableSemanticDB` (or `semanticdbEnabled`) to global, so you can do `ThisBuild / semanticdbEnabled` for simple cases, but for cross-built projects it might require something scoped to each subproject as follows:

```scala
semanticdbEnabled := { Set("2.12.7", "2.11.12")(scalaVersion.value) }
```

main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala
```diff
@@ -0,0 +1,58 @@
...
+    semanticdbOptions := List("-Yrangepos"),
+    semanticdbVersion := "4.0.0",
+    semanticdbCompilerPlugin := {
+      val v = semanticdbVersion.value
+      ("org.scalameta" % "semanticdb-scalac" % v).cross(CrossVersion.full)
+    }
+  )
+
+  override lazy val projectSettings = Seq(
+    allDependencies ++= {
```
olafurpg, line 32: I've never seen `allDependencies`, the documentation says "Inter-project and library dependencies". How does this workaround users who do `libraryDependencies :=`?

olafurpg: What about builds that do `scalacOptions :=`? I think that is more common than hardcoding library dependencies.

olafurpg (commented):
> I think it's an issue intrinsic to the way semanticdb-scalac is currently distributed. Zinc avoids this problem by distributing the compiler bridge source JAR and compiling it on the fly. Not sure if you guys can take similar approach.

That would solve the problem indeed but I don't think that could work for semanticdb-scalac since it's more difficult to build from source. Ultimately, I believe the best long-term solution would be to have builtin support from the compiler.

main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala
eed3si9n: `allDependencies` is an sbt-internal setting that we use to append other things onto `libraryDependencies`, and that's what we actually pass to the library management engine.
https://github.com/sbt/sbt/blob/v1.2.4/main/src/main/scala/sbt/Defaults.scala#L2268-L2279
So even if the build user sets the `libraryDependencies` to `Nil` it doesn't affect `allDependencies`. Normally we name keys `fullSomething`.

I can probably define something similar for `scalacOptions`.

olafurpg: I see, that's awesome! I think it would be great to have something similar for `scalacOptions`, if -Yrangepos is not defined then users get an error message during compilation.

PR #4406, opened 2018-10-05
Bump io

eatkins:
- [x] I've read the [CONTRIBUTING](https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md) guidelines
This is a prerequisite for #4335. Once we release a new io, I'll remove the "REVERT travis..." commit and change the 'Bump io' commit to go to the latest io version rather than `1.2.2-SNAPSHOT`.

PR #4415, opened 2018-10-08
Server spec

eatkins:


eatkins:
I will rebase this once #4406 is merged to get rid of the first four commits.

eatkins:
cc @andreaTP

eatkins:
Hmm...so there are still hangs which means that after the client has read the file from the server, something else hangs (I'd assumed the problem was that we didn't wait long enough for the server to boot).

PR #4417, opened 2018-10-09
fix ServerSpec flaky tests

andreaTP:

Here you have a few improvement to `ServerSpec` that would fix the observed flakiness.
I run it 20 times on Travis without a single failure.

cc. @eatkins @eed3si9n

### DIFF
--- a/sbt/src/test/scala/testpkg/ServerSpec.scala
+++ b/sbt/src/test/scala/testpkg/ServerSpec.scala
@@ -11,6 +11,7 @@ import org.scalatest._
 import scala.concurrent._
 import scala.annotation.tailrec
 import sbt.protocol.ClientSocket
+import scala.util.Try
 import TestServer.withTestServer
 import java.io.File
 import sbt.io.syntax._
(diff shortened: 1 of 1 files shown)


sbt/src/test/scala/testpkg/ServerSpec.scala
```diff
@@ -216,13 +223,24 @@ case class TestServer(baseDirectory: File)(implicit ec: ExecutionContext) {
 
   @tailrec
   final def waitForString(num: Int)(f: String => Boolean): Boolean = {
-    if (num < 0) false
-    else
-      readFrame match {
-        case Some(x) if f(x) => true
-        case _ =>
-          waitForString(num - 1)(f)
+    if (num < 0) { throw new Exception("Retries are over.") } else {
```
eatkins, line 248: I think this could be simplified:
```scala
    if (num < 0) false
    else {
      val latch = new java.util.concurrent.CountDownLatch(1)
      Future(readFrame.filter(f).foreach(_ => latch.countDown()))
      latch.await(1, TimeUnit.SECONDS) || waitForString(num - 1)(f)
    }
```

sbt/src/test/scala/testpkg/ServerSpec.scala
```diff
@@ -216,13 +223,24 @@ case class TestServer(baseDirectory: File)(implicit ec: ExecutionContext) {
...
+      import scala.concurrent.duration._
+      Try {
+        Await.result(
+          result.map(
+            _ match {
+              case Some(x) if f(x) => true
+              case _               => false
+            }
+          ),
+          num seconds
```
eatkins: Did you mean to wait less and less each iteration? In my simplification, I just used a constant timeout, but I'm not sure if this was intentional or not. Also, I prefer `num.seconds` in this kind of instance.

eatkins:
One other improvement that would be nice to add to this PR would be to rename the spec case names, e.g. `withTestServer("events")` gets copied four times which makes it tricky to tell at a glance which test is currently running.

eatkins (changes requested):

eed3si9n (dismissed):
Thanks for picking this up.
LGTM

sbt/src/test/scala/testpkg/ServerSpec.scala
eatkins: Alternatively, I'm not sure you even need a Future at all. Could this not be
```scala
if (num < 0) false
else readFrame.fold(false)(f) || { Thread.sleep(1000); waitForString(num - 1)(f) }
```
or
```scala
num >= 0 && (readFrame.fold(false)(f) || { Thread.sleep(1000); waitForString(num - 1)(f) })
```
?

sbt/src/test/scala/testpkg/ServerSpec.scala
eed3si9n: +1 on `num.seconds`

andreaTP: yup, that's better in this case

sbt/src/test/scala/testpkg/ServerSpec.scala
andreaTP: while the two latter proposals basically change the semantics of this piece of code (i.e. they do not handle anymore the timeout properly). I do agree on moving to the `CountDownLatch` proposal, that is basically the desugared javaish original version :-)
One note is for keeping the exception in `if (num < 0)` because it's fundamental to properly terminate with timeout hanging tests.

sbt/src/test/scala/testpkg/ServerSpec.scala
andreaTP: no more needed with the `CountDownLatch`

andreaTP:
ok, squashed commits and back on the original track.
This is still not perfect but flaky tests are now less than 1% .
I think we should go ahead and live with it for a while.

andreaTP:
@eatkins I rolled back the `CountDownLatch` that was causing issues but I added logs for running test for easier debug.

eed3si9n:
```
[error] /home/travis/build/sbt/sbt/sbt/src/test/scala/testpkg/ServerSpec.scala has changes after scalafmt
[error] com.lucidchart.sbt.scalafmt.ScalafmtCheckFailure: 1 Scala source not formatted in sbtProj:test
```

andreaTP:
correct @eed3si9n my bad, sorry.
now fixed and rebased on latest master.

sbt/src/test/scala/testpkg/ServerSpec.scala
eed3si9n: In the interest of making the tests less flaky, I am +1 with merging this PR as is. We can followup with more cleanup PRs later.

PR #4419, opened 2018-10-11
Fix windows reload loop

eatkins:
On windows* it was possible to get into a loop where the build would
continually restart because for some reason the build.sbt file would get
touched during test (I did not see this behavior on osx). Thankfully,
the repository keeps track of the file hash and when we detect that the
build file has been updated, we check the file hash to see if it
actually changed.

Note that had this bug shipped, it would have been fixable by overriding
the watchOnEvent task in user builds.

The loop would occur if I ran ~filesJVM/test in
https://github.com/swoval/swoval. It would not occur if I ran
test:compile, so the fact that the build file is being touched seems
to be related to the test run itself.


eatkins:
This is the last of the known issues that I've found on windows. From what I can tell, most everything works including re-run with 'r' (except on windows) and automatic reloading of the meta build.

eed3si9n (approved):

Issue #4183, opened 2018-06-05
Proposal to make launcher use a fat jar of shaded sbt dependencies


olafurpg:
- [x] I've read the [CONTRIBUTING](https://github.com/sbt/sbt/blob/1.x/CONTRIBUTING.md) guidelines

I wrote up a proposal in https://github.com/olafurpg/sbt-big on how to speed up sbt launcher and provide more stable binary compatibility for the sbt plugin ecosystem. Related discussion https://github.com/sbt/zinc/issues/546 I've inlined the sbt-big readme below:

----

This project is a proposal to create a new module in sbt/sbt called `sbt-big`
that is a fat jar with all sbt dependencies shaded under an internal namespace.
By updating the sbt launcher to use `org.scala-sbt:sbt-big` instead of
`org.scala-sbt:sbt`, we gain

- faster startup time, by making the launcher download 1 fat jar containing all
  dependencies instead of resolving and downloading 71 jars for dependencies.
  This speedup should be particularly noticable in clean CI and docker environments and from locations with high latency such as Asia or Australia.
- more stable binary compatibility, by shading sbt dependencies under the
  `sbt.internal.*` package so that sbt plugins don't conflict with the sbt boot
  classpath.
  See [sbt/zinc#546](https://github.com/sbt/zinc/issues/546) for an example issue.
  By using `sbt-big`, the sbt project is safe to upgrade to the latest versions of its
  dependencies without breaking the sbt plugin ecosystem.

To try out `sbt-big` locally,

```
git clone https://github.com/olafurpg/sbt-big.git
cd sbt-big
sbt publishM2
```

What happens during publish is that:

- sbt-assembly includes the full dependency tree of sbt
```
$ du -h $HOME/.m2/repository/org/scala-sbt/sbt-big/1.1.0/sbt-big-1.1.0.jar
 21M    /Users/ollie/.m2/repository/org/scala-sbt/sbt-big/1.1.0/sbt-big-1.1.0.jar
```

- the published artifact contains 0 dependencies
```
$ cat $HOME/.m2/repository/org/scala-sbt/sbt-big/1.1.0/sbt-big-1.1.0.pom/sbt-big-1.1.0.pom | grep dependencies
<dependencies>
  <!--the dependency that was here has been absorbed via sbt-assembly-->
  <!--the dependency that was here has been absorbed via sbt-assembly-->
</dependencies>
```

- the fastparse dependency has been shaded into the `sbt.internal.fastparse` package
```
[...]

PR #4443, opened 2018-11-09
Use the classpath obtained by the launcher for metabuild

### LINKED ISSUE #3436: [sbt 0.13] Feature request: SettingKey to check if project is metabuild
## steps
I'm writing a plugin that edits metabuilds like here

```scala
libraryDependencies ++= {
  val sbthost = "org.scalameta" % "sbthost-nsc" % "0.2.0" cross CrossVersion.full
  val isMetabuild = {
    val p = thisProject.value
    p.id.endsWith("-build") && p.base.getName == "project"
  }
  if (isMetabuild) compilerPlugin(sbthost) :: Nil
  else Nil
}
```

## problem

the `isMetabuild` heuristic seems flaky.

## expectation

I expected sbt to provide a more standard way to check from within a settingkey if a project is a metabuild.

## notes

I tried `Project.inPluginProject` but that does not work since it depends on `state.value` which is a taskKey.
sbt version: 0.13.16. I'm writing is a migration tool from sbt 0.13 to sbt 1.0 so an sbt 1.0-only feature won't help me.

### LINKED ISSUE #4437: Exclude "sbt" from the metabuild resolution (sbt artifacts are resolved twice)
## steps

1. `sbt`

## problem

In the metabuild, "sbt" is resolved again even though it's been resolved by the launcher.
This contributes to startup speed slowdown.

## expectation

Exclude "sbt" from the metabuild resolution, and instead append application classpath from the launcher.

## notes

- Effect of this is pronounced more if you use `-SNAPSHOT` build. Ref https://github.com/sbt/sbt/pull/4037
- Also this would be a must have to keep perf speedup from sbt-big effort.

eed3si9n:
Ref #4183
Fixes #3436

This implements isMetaBuild setting that is explicitly for meta build only,
unlike sbtPlugin setting which can be used for both meta build and plugin development purpose.

Fixes #4437

Until now, sbt was resolved twice once by the launcher, and the second time by the metabuild.
This excludes sbt from the metabuild graph, and instead uses app classpath from the launcher.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1851,11 +1851,19 @@ object Classpaths {
       fullClasspath := concatDistinct(exportedProducts, dependencyClasspath).value,
       internalDependencyClasspath := internalDependencies.value,
       unmanagedClasspath := unmanagedDependencies.value,
-      managedClasspath := managedJars(
-        classpathConfiguration.value,
-        classpathTypes.value,
-        update.value
-      ),
+      managedClasspath := {
+        val isMeta = isMetaBuild.value
+        val app = appConfiguration.value
+        val sbtCp0 = app.provider.mainClasspath.toList
+        val sbtCp = sbtCp0 map { Attributed.blank(_) }
+        val mjars = managedJars(
+          classpathConfiguration.value,
+          classpathTypes.value,
+          update.value
+        )
+        if (isMeta) mjars ++ sbtCp
+        else mjars
+      },
       exportedProducts := trackedExportedProducts(TrackLevel.TrackAlways).value,
       exportedProductsIfMissing := trackedExportedProducts(TrackLevel.TrackIfMissing).value,
       exportedProductsNoTracking := trackedExportedProducts(TrackLevel.NoTracking).value,
@@ -2068,6 +2076,7 @@ object Classpaths {
     allDependencies := {
       projectDependencies.value ++ libraryDependencies.value
     },
+    allExcludeDependencies := excludeDependencies.value,
     scalaModuleInfo := (scalaModuleInfo or (
       Def.setting {
         Option(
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/Keys.scala
```diff
@@ -234,6 +234,7 @@ object Keys {
   val classpathOptions = settingKey[ClasspathOptions]("Configures handling of Scala classpaths.").withRank(DSetting)
   val discoveredSbtPlugins = taskKey[PluginDiscovery.DiscoveredNames]("The names of sbt plugin-related modules (modules that extend Build, Plugin, AutoPlugin) defined by this project.").withRank(CTask)
   val sbtPlugin = settingKey[Boolean]("If true, enables adding sbt as a dependency and auto-generation of the plugin descriptor file.").withRank(BMinusSetting)
+  val isMetaBuild = settingKey[Boolean]("If true, this is a metabuild, a project inside project/ directory, and sbt is added to the classpath.").withRank(DSetting)
```
muuki88, line 237: How do you choose the ranking for settings?

muuki88 (commented):
Is there a way to integration test this? I guess we could check the `managedClasspath` for bot `isMetaBuild` `true` and `false`

main/src/main/scala/sbt/Keys.scala
eed3si9n: Rank affects how keys show up in help, I think.

muuki88: Ah, okay :laughing: So this is purely UX

eed3si9n:
@muuki88 scripted should make sure that the builds load. I've manually ran `reload plugins` and then checked `show update` and `show externalDependencyClasspath`.

muuki88:
Sounds fair :+1:

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2068,6 +2076,7 @@ object Classpaths {
     allDependencies := {
       projectDependencies.value ++ libraryDependencies.value
     },
+    allExcludeDependencies := excludeDependencies.value,
```
eed3si9n, line 2079: I just added this so `project/default-auto-plugins` passes.

eed3si9n:
Here's a preliminary benchmark running `git clean -fdx` then `time sbt exit` on Akka build:

```
sbt 1.2.6
sbt exit  149.25s user 7.21s system 359% cpu 43.491 total
sbt exit  144.66s user 6.99s system 345% cpu 43.904 total
sbt exit  150.91s user 6.95s system 349% cpu 45.174 total

sbt 1.3.0-bin-LOCAL20181109
sbt exit  137.83s user 6.93s system 354% cpu 40.823 total
sbt exit  136.84s user 6.71s system 365% cpu 39.250 total
sbt exit  133.40s user 6.78s system 353% cpu 39.680 total
```

So median improved from  43.904s to 39.680s, around **4s faster**.

**Edit**: I should note that it just happened so that for this particular benchmark it improved 4s, but by no means it will be a universal speedup. On smaller builds, the effect seems non-existent or negligible.

main/src/main/scala/sbt/Keys.scala
dwijnand: An opportunity to document that in the code (_hint hint_)

Issue #4446, opened 2018-11-10
Watch aborts if first task isn't successful


japgolly:
- [x] I've read the [CONTRIBUTING](https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md) guidelines

## steps

1. Create a build with two modules A and B such that B depends on A
1. Create a compilation error in A's main sources
2. Run `~B/test`

## problem

SBT reports the compilation error then goes back to the shell/prompt.

## expectation

It should be watching so that when I fix the compilation error, it gets automatically recompiled and then tests run. It used to work in earlier versions.

## notes

sbt version: 1.2.6

eatkins:
I have seen this before, but I'm struggling to reproduce. I have a test project here: https://github.com/eatkins/sbt-coursier-multi-project where I'm unable to get `~` to exit early if I break Foo.scala in the parent directory. Would you be able to post a simple project that causes the issue for you?

I rewrote almost the entire file watch subsystem for sbt 1.3.x. Does the problem persist if you run sbt with `-Dsbt.version=1.3.0-bin-20181106T202042`? This is a nightly build so it will probably take a few minutes for sbt to start up with this property set.

japgolly:
Sure, here is a repo with reproduction steps.

1. Check out https://github.com/japgolly/scala-graal
2. Change an `override def` to `xxx` in `core/src/main/scala/japgolly/scalagraal/Expr.scala`
3. SBT
4. Run `~extBoopickleJVM/test`

Output:

```
root> ~extBoopickleJVM/test
[info] Compiling 9 Scala sources to /home/golly/projects/public/scala-graal/core/target/scala-2.12/classes ...
[error] /home/golly/projects/public/scala-graal/core/src/main/scala/japgolly/scalagraal/Expr.scala:12:47: ';' expected but '=' found.
[error]   xxx apply(context: Context): Expr.Result[A] =
[error]                                               ^
[error] one error found
[error] (core / Compile / compileIncremental) Compilation failed
root>
```

eatkins:
Great, thanks. I can reproduce on my machine now. I'll see about a fix.

eatkins:
This should be fixed by #4447.

japgolly:
Awesome and thanks for getting a fix in so fast!!

eed3si9n:
Closing this as resolved.

PR #4447, opened 2018-11-11
Fix '~' for dependent projects with a broken parent (1.2.x)

eatkins:
In #4446, @japgolly reported that in some projects, if a parent project
was broken, then '~' would immediately exit upon startup. I tracked it
down to this managed sources filter. The idea of this filter is to avoid
getting stuck in a build loop if managedSources writes into an unmanaged
source directory. If the (managedSources in ThisScope).value line
failed, however, it would cause the watchSources, and by delegation,
watchTransitiveSources task to fail. The fix is to only create this
filter if the managedSources task succeeds.

I'm not 100% sure if we shouldn't just get rid of this filter entirely
and just document that '~' will probably loop if a build writes the
result of managedSources into an unmanaged source directory.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -361,14 +361,19 @@ object Defaults extends BuildCommon {
       val include = (includeFilter in unmanagedSources).value
       val exclude = (excludeFilter in unmanagedSources).value match {
         case e =>
-          (managedSources in ThisScope).value match {
-            case l if l.nonEmpty =>
-              e || new FileFilter {
-                private val files = l.toSet
-                override def accept(pathname: File): Boolean = files.contains(pathname)
-                override def toString = s"ManagedSourcesFilter($files)"
-              }
-            case _ => e
+          val s = state.value
+          try {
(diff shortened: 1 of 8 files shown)


eed3si9n (approved):
LGTM

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -361,14 +361,19 @@ object Defaults extends BuildCommon {
...
-            case l if l.nonEmpty =>
-              e || new FileFilter {
-                private val files = l.toSet
-                override def accept(pathname: File): Boolean = files.contains(pathname)
-                override def toString = s"ManagedSourcesFilter($files)"
-              }
-            case _ => e
+          val s = state.value
+          try {
+            Project.extract(s).runTask(managedSources in Compile in ThisScope, s) match {
```
dwijnand, line 366: Out of graph execution?  Ouch.

eatkins: Yeah, this whole thing is a mistake. I don't think we should support managedSources writing into the unmanaged source directories. You deserve an infinite loop if you set up your build that way.

PR #4454, opened 2018-11-18
Fix prompt for task progress

eed3si9n:
Some minor improvement/fixes for prompts + logging around super shell.
This should fix the behavior where the prompt overwrites the last output like for `show name` etc.

Ref #4461

### DIFF
--- a/main/src/main/scala/sbt/internal/Aggregation.scala
+++ b/main/src/main/scala/sbt/internal/Aggregation.scala
@@ -77,8 +77,6 @@ object Aggregation {
       if (show.taskValues) printSettings(r, show.print)
     }
     if (show.success) printSuccess(start, stop, extracted, success, log)
-    // wait for async logger to catch up
-    Thread.sleep(100)
   }
 
   def timedRun[T](
@@ -131,9 +129,9 @@ object Aggregation {
     if (get(showSuccess)) {
       if (get(showTiming)) {
         val msg = timingString(start, stop, structure.data, currentRef)
-        if (success) log.success(msg + "\n") else log.error(msg + "\n")
+        if (success) log.success(msg) else log.error(msg)
       } else if (success)
-        log.success("" + "\n")
+        log.success("")
     }
   }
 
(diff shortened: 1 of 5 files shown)


main/src/main/scala/sbt/internal/TaskProgress.scala
```diff
@@ -86,20 +86,24 @@ private[sbt] final class TaskProgress(currentRef: ProjectRef) extends ExecutePro
...
-    console.println("")
-    console.println("")
-    console.println("")
-    console.println("")
-    console.println("")
-    console.println("")
-    console.print(CursorUp5)
+    if (isLogReady.get) ()
+    else {
+      console.println("")
```
eatkins: This is a bit inscrutable. Maybe you could add a comment that explains why you need to print these blank lines followed by CursorUp5? I'd also find this slightly more readable if it were:
```scala
console.print(s"${"\n" * 5}$CursorUp5")
```

main/src/main/scala/sbt/internal/Aggregation.scala
```diff
@@ -77,8 +77,6 @@ object Aggregation {
       if (show.taskValues) printSettings(r, show.print)
     }
     if (show.success) printSuccess(start, stop, extracted, success, log)
-    // wait for async logger to catch up
-    Thread.sleep(100)
```
eatkins, line 81: Always great to remove a Thread.sleep.

main/src/main/scala/sbt/internal/TaskProgress.scala
```diff
@@ -86,20 +86,24 @@ private[sbt] final class TaskProgress(currentRef: ProjectRef) extends ExecutePro
 
   private[this] val console = ConsoleOut.systemOut
   private[this] def readyLog(): Unit = {
```
eatkins: Could this itself just be a lazy val and then you could get rid of the AtomicBoolean?

eatkins (dismissed):
I don't feel strongly about my suggestions so you should feel free to merge with or without them. I'm hoping these changes will make super shell more useable for me!

main/src/main/scala/sbt/internal/TaskProgress.scala
eed3si9n: That's a good idea.

main/src/main/scala/sbt/internal/TaskProgress.scala
eed3si9n: I initially added this to make space on the bottom for task progress. I am going to see if we still need this actually.

PR #4455, opened 2018-11-19
Reuse State.stateOps

dwijnand:


eatkins:
Looks like that fails MIMA:
```
[error]  * method stateOps(sbt.State)sbt.StateOps in object sbt.State has a different result type in current version, where it is sbt.State rather than sbt.StateOps
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("sbt.State.stateOps")
```
Thinking about it though, I doubt that we need to deprecate stateOps, we could just make it private[sbt]. I doubt many people are explicitly referring to it in their source code.

dwijnand:
> I doubt many people are explicitly referring to it in their source code.

You'd be surprised.

Looks like an implementation detail of how value classes desugar is making this a binary incompatibility.

Not worth pursuing.

eatkins:
Yeah. Binary compatibility is a pain.

PR #4456, opened 2018-11-19
Multi command

### LINKED ISSUE #3051: Support running several commands more intuitively
## problem
For new contributors, remembering that the sbt syntax to run several commands at once is `; myCommand; myCommand2` is hard and unintuitive.
  
Why is it unintuitive? Scala/Java programs don't follow this convention to execute several statements (**EDIT**: it could be the case that bash doesn't either, though there's a slight change in semantics).

## expectation
New users should run `myCommand; myCommand2`to execute sequentially `myCommand` and `myCommand2`. The current syntax can be deprecated or still allowed, but the new one should work out of the box.

In Scala (and also bash), `myCommand; myCommand2` is valid. This is why the proposed syntax is more intuitive and user-friendly.

eatkins:

This PR enhances the parser for multiple commands/tasks and shares the improved parser with the command parser used by Watched.watch so that there should be parity between which commands are valid on the command line and which commands are valid in a continuous build.

The two main features this PR adds is
1. proper support for string literals in a multi command -- prior to this change, if you tried to run something like
```
> ;bash "touch src/main/scala/Foo.scala; rm target/classes/Foo.class"; compile
```
it would not work correctly because the `;` in the bash command string would be interpreted as a command separator. Now the string will be interpreted as a literal value and the `;` in the bash command string is not considered a token.
2. drop the requirement that multiple commands need to be prefixed with a leading ';'. Now the command above could just be
```
> bash "touch src/main/scala/Foo.scala; rm target/classes/Foo.class"; compile
```

Also, the `~` will now use the multi command parser rather than the naive string split on `;` that it was doing before. This should make it so that any multi command that is valid on the console can be prefixed with `~` and things should work mostly as expected.

I also verified manually that tab completions still work for `;` prefixed tasks/commands.

Fixes sbt/sbt#3051

### DIFF
--- a/main-command/src/main/scala/sbt/BasicCommands.scala
+++ b/main-command/src/main/scala/sbt/BasicCommands.scala
@@ -155,13 +155,16 @@ object BasicCommands {
   }
 
   private[sbt] def multiParserImpl(state: Option[State]): Parser[List[String]] = {
-    val nonSemi = token(charClass(_ != ';', "not ';'").+, hide = const(true))
+    val nonSemi = charClass(_ != ';', "not ';'")
     val semi = token(';' ~> OptSpace)
-    def commandParser = state.map(s => (s.combinedParser & nonSemi) | nonSemi).getOrElse(nonSemi)
-    val part = semi flatMap (
-        _ => matched(commandParser) <~ token(OptSpace)
+    val nonQuote = charClass(_ != '"', label = "not '\"'")
+    val cmdPart = token(
+      ((nonSemi & nonQuote).map(_.toString) | StringEscapable.map(c => s""""$c"""")).+,
+      hide = const(true)
     )
-    (part map (_.trim)).+ map (_.toList)
+    def commandParser = state.map(s => (s.combinedParser & cmdPart) | cmdPart).getOrElse(cmdPart)
+    val part = semi.flatMap(_ => matched(commandParser) <~ token(OptSpace)).map(_.trim)
+    part.+ map (_.toList)
   }
 
   def multiParser(s: State): Parser[List[String]] = multiParserImpl(Some(s))
--- a/main-command/src/main/scala/sbt/BasicCommands.scala
+++ b/main-command/src/main/scala/sbt/BasicCommands.scala
@@ -154,14 +154,17 @@ object BasicCommands {
     state
   }
 
-  def multiParser(s: State): Parser[List[String]] = {
+  private[sbt] def multiParserImpl(state: Option[State]): Parser[List[String]] = {
     val nonSemi = token(charClass(_ != ';', "not ';'").+, hide = const(true))
     val semi = token(';' ~> OptSpace)
+    def commandParser = state.map(s => (s.combinedParser & nonSemi) | nonSemi).getOrElse(nonSemi)
     val part = semi flatMap (
-        _ => matched((s.combinedParser & nonSemi) | nonSemi) <~ token(OptSpace)
+        _ => matched(commandParser) <~ token(OptSpace)
     )
     (part map (_.trim)).+ map (_.toList)
   }
+
+  def multiParser(s: State): Parser[List[String]] = multiParserImpl(Some(s))
 
   def multiApplied(s: State): Parser[() => State] =
     Command.applyEffect(multiParser(s))(_ ::: s)
--- a/main-command/src/main/scala/sbt/BasicCommands.scala
+++ b/main-command/src/main/scala/sbt/BasicCommands.scala
@@ -39,7 +39,6 @@ object BasicCommands {
     ignore,
     help,
     completionsCommand,
-    multi,
     ifLast,
     append,
     setOnFailure,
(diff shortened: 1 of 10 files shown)


eed3si9n:
One of the commits says "Do not require leading space for multi command." This should be do not require leading semicolon?

eed3si9n:
Fixes sbt/sbt#3051

eatkins:
Yes, that is correct. It should say semicolon. I am going to have to revise this PR so I'll update the commit message when I do.

sbt/src/sbt-test/watch/watch-parser/test
```diff
@@ -0,0 +1,18 @@
+> ~; setStringValue foo; setStringValue bar
+
+> checkStringValue bar
+
+> ~;setStringValue foo;setStringValue bar; checkStringValue bar
```
eed3si9n, line 5: In this example, is the leading semicolon required, or optional?

dwijnand:
More awesome work, @eatkins!

main-command/src/main/scala/sbt/BasicCommands.scala
```diff
@@ -39,7 +39,6 @@ object BasicCommands {
     ignore,
     help,
     completionsCommand,
-    multi,
```
eed3si9n, line 42: What's the reason for removing here?

eatkins: Yeah. I meant to put this in the commit message. If multi comes after the action parser in the parser list, and the command does not have a leading `;`, then the action parser will parse it before the multi parser has a chance. I'm not sure if there is a better way to handle this.

sbt/src/sbt-test/watch/watch-parser/test
eatkins: optional. I'll duplicate that test case without the leading `;`.

main-command/src/main/scala/sbt/BasicCommands.scala
eed3si9n: In other words, we are introducing an ambiguity in the parsing semantics that's resolved by the implementation detail of which parser is listed first.

eatkins: Correct. I considered modifying the action parser, but that was a lot more work/was more scary.

eatkins: The reason I think this is ok is that the multi parser will not accept any command that doesn't contain a `;` that is not in a string literal.

dwijnand: > we are introducing an ambiguity in the parsing semantics that's resolved by the implementation detail of which parser is listed first

I think that's overly pessimistic sounding.

This change is enhancing the action parser by allowing multiple actions to be defined.

Previously multiple actions were supported with a prefix `;` that first failing to parse in the `act` parser then parsed with the `multi` parser.  Now, in order to give the `multi` parser a chance to act (no pun intended) before the action parser, we need to correct the order.

eed3si9n: @eatkins
 
> The reason I think this is ok is that the multi parser will not accept any command that doesn't contain a `;` that is not in a string literal.

That's good.

@dwijnand 

> I think that's overly pessimistic sounding.
>
> This change is enhancing the action parser by allowing multiple actions to be defined.

Yea. I am all for treating `;` as part of command syntax, but I just want to understand if we're introducing some tradeoff. If Ethan's saying that `multi` won't accept unless there's `;`, that makes me feel better since that's not ambiguous when you have command like `testOnly foo`.

eatkins: Yeah. If that case were ambiguous, I think that sbt would get caught in an infinite loop.

main-command/src/main/scala/sbt/BasicCommands.scala
```diff
@@ -154,15 +153,24 @@ object BasicCommands {
...
-        _ => matched((s.combinedParser & nonSemi) | nonSemi) <~ token(OptSpace)
+    val nonQuote = charClass(_ != '"', label = "not '\"'")
+    val cmdPart = token(
+      ((nonSemi & nonQuote).map(_.toString) | StringEscapable.map(c => s""""$c"""")).+,
+      hide = const(true)
     )
-    (part map (_.trim)).+ map (_.toList)
+    def commandParser = state.map(s => (s.combinedParser & cmdPart) | cmdPart).getOrElse(cmdPart)
+    val part = semi.flatMap(_ => matched(commandParser) <~ token(OptSpace)).map(_.trim)
+    (cmdPart.? ~ part.+).map {
```
eatkins, line 166: The `cmdPart.? ~ part.+` is what ensures that a `;` separator somewhere in the command is necessary for the parser to succeed. This is because `part` must start with a semicolon.

eed3si9n (approved):

PR #4473, opened 2018-11-30
bump util, lm, and zinc

eed3si9n:


dwijnand:
Forward-porting the TestConsoleLogger into develop as a part of #4616.

PR #4474, opened 2018-12-03
Annotate type signature of public methods and fields

eatkins:
This file was littered with intellij warnings due to public members and
fields not having their types annotated. Although in this case it didn't
really matter, it is good practice to always annotate public methods and
fields so that they can evolve in a binary compatible way.


eed3si9n (approved):
Thanks @eatkins

PR #4475, opened 2018-12-03
Optimize imports

eatkins:
It was becoming a pain to work on these files in intellij because the
auto-import feature would implicitly optimize all of the imports in
these files, leading to a large diff. I'd then have to go and manually
add the import that I care about. This change does add some wildcard
imports, which I don't always love, but these files are so unwieldy
already that I think it's worth it to have the imports follow the format
preferred by intellij.


eed3si9n (approved):
LGTM pending Travis

eatkins:
Thanks @eed3si9n. I'll merge when CI finishes (it had to run again after I updated the branch for the changes in my other PR).

PR #4476, opened 2018-12-04
Add cache of layered classloaders for test and run tasks

eatkins:
### usage

```scala
Compile / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
Compile / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.ScalaInstance
Compile / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.RuntimeDependencies

Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.ScalaInstance
Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.RuntimeDependencies
Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.ShareRuntimeDependenciesLayerWithTestDependencies
Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.TestDependencies
```

### original description

This commit adds layered classloaders to sbt so that we can cache the classloaders for the portion of the classpath that never changes. There are two benefits to doing this:
1) performance: by reusing classloaders for the dependencies, we do not have to reload all of the potentially large library jar classes. This classloading dominates the startup time of the test and run tasks.
2) memory/metaspace utilization: sbt can quickly run out of memory or meta space if any of the tasks potentially leak classloaders (usually by leaking a thread). Caching the classloaders that load most of the classes can help. Especially if the leak is in the library classes. 

Here are some gifs that show off the performance improvements:

In this example, I just illustrate how quickly the test task runs when invoked repeatedly, even though the test has to initialize an Akka actor system:
Output sample
Note that after I cleared the cache, it got slow again on the next iteration.

Here, I generated 5000 source files in the above project. In each build, zinc has to check all of the source files for changes, which can take a long time if io must be done on all 5000 files. In the following to gifs, I show one terminal triggering a continuous build and the build running in the second window. In the first gif, I use no caching:
Output sample

Here I use full caching:
Output sample


eed3si9n:
sbt currently depends on ScalaCache and [caffeine](https://github.com/ben-manes/caffeine). Do you think we could use them for this purpose?

He-Pin:
Thanks, @eatkins really great! With it already depends on caffeine, so maybe reuse of them? other things are:
1. limit the max opening files
2. when evicted, close the file.

eatkins:
@eed3si9n we could certainly use caffeine to replace LRUCache. Beyond that, I'm not completely sure that we'd get much additional value out of the caching libraries. The actual caching logic is minimal and I put everything behind interfaces that don't expose whether or not a cache is used to retrieve the result.

eatkins:
@kerr is there a specific issue that you've seen with sbt opening too many files? As far as I know, sbt shouldn't be holding handles to open files regardless of whether or not a cache is used. FWIW, all of the caches that I've added to sbt cleanup the cached resources on eviction or when the cache is explicitly closed.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -373,6 +375,7 @@ object Defaults extends BuildCommon {
         excludeFilter in unmanagedSources
       )
       .value,
+    layeringStrategy in ConfigGlobal := LayeringStrategy.Default,
```
eed3si9n: This should go to global scope so we can do `ThisBuild / layeringStrategy`

main/src/main/scala/sbt/LayeringStrategy.scala
```diff
@@ -0,0 +1,115 @@
...
+ *  val bar: Object = classLoader.loadClass("foo.Foo$").getDeclaredMethod("bar").invoke(null)
+ *  val barTyped: foo.Bar = bar.asInstanceOf[foo.Bar]
+ *  // throws ClassCastException because the thread context class loader is altClassLoader, but
+ *  // but bar was loaded by classLoader.
+ * }}}
+ *
+ * In general, this should only happen if the user explicitly overrides the thread context
+ * ClassLoader or uses reflection to manipulate classes loaded by different loaders.
+ */
+sealed trait LayeringStrategy
```
eed3si9n: Contraband?

main/src/main/scala/sbt/Keys.scala
```diff
@@ -264,6 +264,7 @@ object Keys {
   val bgRunMain = inputKey[JobHandle]("Start a provided main class as a background job")
   val fgRunMain = inputKey[Unit]("Start a provided main class as a foreground job")
   val bgCopyClasspath = settingKey[Boolean]("Copies classpath on bgRun to prevent conflict.")
+  val layeringStrategy = settingKey[LayeringStrategy]("Creates the classloader layering strategy for the particular configuration.")
```
eed3si9n: Maybe we should come up with more verbose or intuitive naming for this key and class?

main/src/main/scala/sbt/LayeringStrategy.scala
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Provides instances of [[LayeringStrategy]] that can be used to define the ClassLoader used by
+ * [[Keys.run]], [[Keys.test]] or any other task that runs java code inside of the sbt jvm.
+ */
+object LayeringStrategy {
```
eed3si9n: Also let's think more about the enum names.

eed3si9n:
It would be interesting to see how slf4j / logback behaves since they use classpath as configuration.

eatkins:
<s>I can't reproduce the travis failures either on my mac or on my linux vm using adopt jdk or just regular openjdk. It is really weird.
I fixed the travis failures by removing `sudo: false` and `dist: trusty` from the .travis.yml. Travis is removing `sudo: false` anyway. I'm not sure why we were specifying the dist key.</s>

Actually there was a bug in initializing the file tree repository that caused the travis failures.

jvican:
Nice to see this change implemented here @eatkins A layered classloader infrastructure to run tests is something it's been on my todo list for a long time, I will probably experiment and merge it in bloop soon too. It's nice to see this kind of experimentation in the sbt main repository.

eatkins:
@jvican thanks for the feedback. This work did start as a plugin, but I mentioned to @eed3si9n that I was working on the plugin and he mentioned that adding this may fix or at least mitigate some longstanding performance/memory issues in sbt so I went ahead and directly embedded the plugin into sbt.

Long term, it may make sense to move some of this code into the util project so that it can be used by other projects. Right now, it is too much of a pain to work across repos which is why I implemented everything directly in sbt itself.

main/src/main/scala/sbt/LayeringStrategy.scala
eatkins: https://github.com/sbt/sbt/pull/4476/commits/25414683740c4e57e53fec31787dc0986fafa785

eatkins:
@eed3si9n I think I addressed your review feedback and I feel comfortable with this PR now.

main/src/main/scala/sbt/Keys.scala
eatkins: I changed it to [cC]lassLoaderLayeringStrategy which is definitely verbose and hopefully more intuitive.

main/src/main/scala/sbt/internal/ClassLoaders.scala
```diff
@@ -0,0 +1,213 @@
...
+         * The first two layers may be optionally cached to reduce memory usage and improve
+         * start up latency. Because there may be mutually incompatible libraries in the runtime
+         * and test dependencies, it's important to be able to configure which layers are used.
+         */
+        val (layerDependencies, layerTestDependencies) = s match {
+          case ClassLoaderLayeringStrategy.ShareRuntimeDependenciesLayerWithTestDependencies =>
+            (true, true)
+          case ClassLoaderLayeringStrategy.ScalaInstance       => (false, false)
+          case ClassLoaderLayeringStrategy.RuntimeDependencies => (true, false)
+          case _                                               => (false, true)
```
eed3si9n: Does `Default` fall into this pattern, if so it might need a distinct name.

eatkins:
The idea behind Default was that it could be set globally and then any task that needed a layered classloader could choose what to do when Default was specified. I realized that Default really only applied at the configuration level, so it made sense to just explicitly set the classLoaderLayeringStrategy in the Compile and Test configurations. This is done in 
~~https://github.com/sbt/sbt/pull/4476/commits/87a0c58790396e97200e241a188f28425b7dabf5~~ 
~~https://github.com/sbt/sbt/pull/4476/commits/2e0dbfc7b7016786ebb4cbd135cb823b764ac40d~~ 
https://github.com/sbt/sbt/commit/ab5f14c5f03d264ff7e13cb3b4f9a81bb6ce9a1d.

eed3si9n (dismissed):
I think we are good to go on this.
Thanks for the contribution!

eatkins:
@eed3si9n I just force pushed the branch rebased against origin/develop. I need another approval to actually merge once CI finishes.

eed3si9n (approved):

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1764,24 +1760,44 @@ object Defaults extends BuildCommon {
 
   // 1. runnerSettings is added unscoped via JvmPlugin.
   // 2. In addition it's added scoped to run task.
-  lazy val runnerSettings: Seq[Setting[_]] = Seq(runnerTask, forkOptions := forkOptionsTask.value)
+  lazy val runnerSettings: Seq[Setting[_]] = {
+    val unscoped: Seq[Def.Setting[_]] =
+      Seq(runner := ClassLoaders.runner.value, forkOptions := forkOptionsTask.value)
+    inConfig(Compile)(unscoped) ++ inConfig(Test)(unscoped)
```
eed3si9n, line 1766: `runnerSettings` is part of `configSettings`, so this shouldn't be scoped to a particular configuration.
This will break some of the plugins - https://github.com/sbt/sbt-validator/issues/35

eatkins: Ok. I'll figure out how to fix this.

eatkins: I was a little worried about this change actually.

retronym:
Let's warn people about the risks of leaking static state across `run` invocations.

For example, if `runMain my.App` executed `myLibrary.staticField = 42`, this would be still seen by subsequent `run` tasks.

The static state may retain large amounts of heap (for instance a cache with no timed eviction policy).

I would suggest that SBT's `ClassLoaderCache` itself should have a timed eviction policy to counteract these problems. We would probably like the classloader cache used for the compiler to have a long timeout (because JITting scalac takes 1-2 minutes!), and a smaller timeout for third-party library caches (because we have no idea how much memory they statically retain.)

eatkins:
@retronym there is a command, clearCaches, that flushes the classloader cache.

The third party library case is difficult to handle universally. Ideally, a library should provide handles to any resources whether they are caches or threads or whatever that allow them to be cleaned up after the test run completes. In the real world, however, I am sure many projects rely on third party libraries that use singletons and leak threads. The question partially becomes to what extent do we allow badly written third party libraries control the default behavior of sbt. My instinct is to implicitly encourage users to use libraries that don’t leak resources because we can make the sbt ux so much better in those cases. I do not, however, want users to grow overly frustrated with sbt if the defaults are not reliably working in the most important use cases. All of that is to say that we are still trying to figure out what the impact of changing the default layering strategy is. Things may change between now and 1.3.0 final, but we’re gathering some very valuable data in the meantime.

eatkins:
I should add that I do 100% agree that by sbt 1.3.0 final there should be a detailed guide to all the layering strategies: the use cases each supports and the drawbacks/issues that may be encountered.

Not to belabor the point, but thread leaks are a problem even if the cache is evicted since the thread references are kept alive. sbt has always had this problem.

The only true solution is fork :)

Issue #3110, opened 2017-04-17
Def cannot be used inside a taskDyn


jvican:
## steps

Say you're implementing a dynamic task, and you want its result to be a task whose body executes a collection of tasks sequentially. An example:

```scala
    val myTask: Def.Initialize[Task[Unit]] = Def.taskDyn {
      if (foo) {
        Def.sequential(Keys.publish, Bintray.bintrayRemoteSign)
      } else Def.task(())
    }
```

## problem

When compiling the previous task, you'll get:
```
[error] Illegal dynamic reference: Def
[error] Illegal dynamic reference: Def
```

## expectation

Not getting an illegal dynamic reference.

## notes

A workaround is to define your own `TaskSequential`:

```scala
val StableDef = new TaskSequential {}
```

And redefining the previous task as:

```scala
    val myTask: Def.Initialize[Task[Unit]] = Def.taskDyn {
      if (foo) {
        StableDef.sequential(Keys.publish, Bintray.bintrayRemoteSign)
      } else Def.task(())
    }
```

jvican:
I suspect this error could be related with the implementation of `TaskSequential` via implicits. I think this was done to stay binary compatible.
 
```
trait DefExtra {
  private[this] val ts: TaskSequential = new TaskSequential {}
  implicit def toTaskSequential(d: Def.type): TaskSequential = ts
}
```
  
What could be happening is that the macro errors when it finds a function requiring an argument, since `Def.sequential` is expanded into `toTaskSequential(Def)`.

OlegYch:
another sample:

```
 lazy val osgiSettings = TaskKey[Seq[Setting[_]]]("OSGi Settings Generator")
  override def projectSettings = List(
    osgiSettings := (if (name.value.nonEmpty) seq() else seq())
  )
```
http://scastie.org/30771

dwijnand:
Can anyone provide a reproduction of this? Can't reproduce with sbt 1.1.0 and a build.sbt of:

```scala
val foo: Boolean = true

val myTask: Def.Initialize[Task[Unit]] = Def.taskDyn {
  if (foo) {
    Def.sequential(Keys.publish, clean)
  } else Def.task(())
}
```

jvican:
This issue was not easy to reproduce IIRC. You can try if https://github.com/scalacenter/sbt-release-early/ compiles with 1.1.0 if you remove the `StableDef` trick, I did it there because I got the error there for the first time. If it compiles, this ticket can be closed.

[...]

Issue #3718, opened 2017-11-06
watchSources is redundant and not precise


OlegYch:
This is more of an idea than an issue.

Recent issues with `~` like https://github.com/sbt/sbt/issues/3695 and https://github.com/sbt/sbt/issues/3501, and continuing performance problems with change detection (although slightly alleviated by NIO watching and upcoming enhancements in hashing) have led me to think that `watchSources` needs to go.

Instead ~ has to deduce the set of files consumed by each task and feed the changes to the task. 
One way to do that might be to track files accessed by each task in a bit of global mutable in-memory state - ie each task would be given a `(cache: (File => (File => T) => T), modified: Set[File])`. This state intentionally would not survive sbt restart, but imho this is ok at the moment, though it would be nice to unify that with existing caching facilities.

Perhaps there are better ways, but this was the first thing i could imagine and i'd like to receive some feedback on this

ShaneDelmore:
I use watchSources to depend on more than what is referenced during compilation.  For example, I may have test oracles files I want to compare output too, zipped files, json files, things like that, and when my code changes I want to rerun my tests, but also if any of my test oracles change I want to rerun my tests.  Isn't this what watchSources is designed for?

OlegYch:
@ShaneDelmore good point
`run` and `test` depend on `products` which reads config files, and you could make it depend on other files which are only read at runtime, just like you do with `watchSources`, so that usecase should be well covered

ShaneDelmore:
Just to make sure I am understanding this, is the proposed change to add test oracles to products instead of watchSources to enable ~test to catch changes to test resources?

OlegYch:
@ShaneDelmore any of the tasks that all of `test` `testQuick` and `testOnly` depend on, `Test / products` might be a good option
the proposed change is not limited to tests though, but rather to any task that reads files
currently `watchSources` has no idea what task it's used for, and therefore is either too eager or misses stuff and cannot be used for incremental change detection

ShaneDelmore:
In that case, I see your point but fear the outcome.  I imagine conversations like these:

NewSbtUser: I want to make ~test rerun tests when I change a file, how do I do it?
SbtExpert: Add the files to watchSources.
NewSbtUser:  That makes sense, thanks!

Vs
NewSbtUser: I want to make ~test rerun tests when I change a file, how do I do it?
SbtExpert: Add the files to any that test, testQuick, and testOnly depend on.
NewSbtUser: Huh?

[...]

PR #4485, opened 2018-12-11
Support disabling the sbt task linter

### LINKED ISSUE #3266: Reconsider failing `.value` inside lambda expression
## steps

```scala
    test in Test := {
      List(1, 2, 3) foreach { x =>
        streams.value.log.info(s"foo $x")
      }
      (test in Test).value
    }
```

## problem

```
build.sbt:17: error: The evaluation of `streams` inside an anonymous function is prohibited.

Problem: Task invocations inside anonymous functions are evaluated independently of whether the anonymous function is invoked or not.

Solution:
  1. Make `streams` evaluation explicit outside of the function body if you don't care about its evaluation.
  2. Use a dynamic task to evaluate `streams` and pass that value as a parameter to an anonymous function.

        streams.value.log.info(s"foo $x")
        ^
[error] sbt.compiler.EvalException: Type error in expression
```

## expectation

The task does exactly what I want it to do. It should work without warnings or error.

## notes

In general, an error (or even a warning) should be displayed when we are 90% certain that the build is doing something against what the user has intended. In this case, I don't think it is.

Ref https://github.com/lucidsoftware/neo-sbt-scalafmt/pull/11, https://github.com/sbt/sbt/pull/3216/commits/ca3acc4e52668c77a1f2dfa0cc247a45936cf7bd

/cc @jvican

eatkins:
This PR resolves #3266. It also does some minor clean up of the TaskLinterDSL to make it more maintainable in general.

### DIFF
--- a/main-settings/src/main/scala/sbt/dsl/LinterLevel.scala
+++ b/main-settings/src/main/scala/sbt/dsl/LinterLevel.scala
@@ -1,0 +1,68 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.dsl
+
+/**
+ * Controls the kind of linting performed during task macro expansion. The linting level is
+ * controlled by implicitly bringing an instance of LinterLevel into scope:* {{{
+ *   val foo = taskKey[Unit]("")
+ *   val bar = taskKey[Unit]("")
+ *   // This compiles because of the import in the block
+ *   val fooTask = {
+ *     import sbt.dsl.LinterLevel.Ignore
+ *     Def.task {
+ *       if (true) bar.value
+ *       else bar.value
+ *     }
+ *   }
+ *   // because the import above was in a local scope, it does not apply here, so this won't
+ *   // compile:
+ *   //val barTask = Def.task {
+ *   //  if (true) foo.value
+ *   //  else foo.value
+ *   //}
+ *
+ *   import sbt.dsl.LinterLevel.Ignore
+ *   // Both defs compile because the Ignore LinterLevel is implicitly brought into scope and
+ *   // picked up by both task defs underneath
+ *   val newFooTask = Def.task {
+ *     if (true) bar.value
+ *     else bar.value
+ *   }
+ *   val barTask = Def.task {
+ *     if (true) foo.value
+ *     else foo.value
+ *   }
+ * }}}
+ * To make this work, the instances are all defined as implicit case objects. Moreover, the
+ * the [[LinterLevel.Abort]] setting is made default by placing [[LinterLevel.Warn]] and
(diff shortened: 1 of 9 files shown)


eatkins:
cc @dwijnand @eed3si9n. I whipped this up fairly quickly and didn't give a ton of thought to the names and project layout.

dwijnand:
I reviewed it _really_ really quick and it looked good, but I also approve of fixing that issue in general.  Thanks, Ethan.

main-settings/src/main/scala/sbt/dsl/LinterLevel.scala
```diff
@@ -0,0 +1,17 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.dsl
+
+sealed trait LinterLevel
```
eatkins, line 47: I considered contraband for this, but wasn't sure if there is a way to put the (prioritized) implicit instances in the companion object.

eed3si9n: You can use https://www.scala-sbt.org/contraband/schema.html#Extra+code

```
  #xcompanioninterface CompanionInterface1
```

eed3si9n:
This is awesome. Thanks @eatkins!

main-settings/src/main/scala/sbt/dsl/LinterLevel.scala
```diff
@@ -0,0 +1,17 @@
...
+object LinterLevel extends LinterLevelLowPriority {
+  implicit case object Warn extends LinterLevel
```
eed3si9n, line 54: There should be an explanation of why this is made implicit, and how this is meant to be used.

eed3si9n (changes requested):

main-settings/src/main/scala/sbt/dsl/LinterLevel.scala
eatkins: I force pushed an update that added scaladoc to these.

eed3si9n (approved):

dwijnand:
Where did that fix come from?!  Very nice.

If you want a more common source of dynamic tasks put some projects in a seq and then map the seq, scope a key and .value that.  Users try to do that all the time.

```scala
Seq(foo, bar, baz).map(p =>
  (bark in p).value
)
```

eatkins:
Funnily enough, this is all related #3718. I'm working on trying to implement the taskSources idea in the comment at the bottom. This requires traversing the list of dependencies for each task that is being watched to extract all of the taskSources dependencies. While working on that, I realized that dynamic tasks are a huge problem because we lose the dependency information for dynamic tasks. I started working on modifying the Def.taskDyn macro to try and capture all of the dependencies for the Def.initialize instances that are referenced in the dynamic task definition. This forced me to get quite familiar with the TaskLinterDSL at which point I decided to go ahead and fix #3110 and #3266, both of which had been irritating me for a long time.

PR #4488, opened 2018-12-14
Rename TestBuild.Keys to TestBuild.TestKeys

eatkins:
It drives me crazy that in intellij when I do the go to class task that
TestBuild.Keys comes up before Keys. Given how central Keys is to sbt,
it doesn't seem like a good idea to alias that particular class name.

### DIFF
--- a/main/src/test/scala/Delegates.scala
+++ b/main/src/test/scala/Delegates.scala
@@ -23,12 +23,12 @@ object Delegates extends Properties("delegates") {
     t.nonEmpty
   }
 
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
To confirm, this is refactoring of tests, and has no effect to users, correct?

eatkins:
Yeah, this class is only used in the mainProj Delegates property test. I don't think we publish the test jar so there isn't a binary compatibility concern (and mima didn't complain about this change).

PR #4497, opened 2018-12-21
Preserve errors order when publishing diagnostics

tdroxler:
The `sbt-server` was prepending a new probem and not appending.

The result was a `textDocument/publishDiagnostics` notification
containing a inverted list of problems compare to what was show in the
sbt console.

### DIFF
--- a/main/src/main/scala/sbt/internal/server/LanguageServerReporter.scala
+++ b/main/src/main/scala/sbt/internal/server/LanguageServerReporter.scala
@@ -40,7 +40,7 @@ class LanguageServerReporter(
 ) extends ManagedLoggedReporter(maximumErrors, logger, sourcePositionMapper) {
   lazy val exchange = StandardMain.exchange
 
-  private[sbt] lazy val problemsByFile = new mutable.HashMap[File, List[Problem]]
+  private[sbt] lazy val problemsByFile = new mutable.HashMap[File, mutable.ListBuffer[Problem]]
 
   override def reset(): Unit = {
     super.reset()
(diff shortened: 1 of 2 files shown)


main/src/main/scala/sbt/internal/server/LanguageServerReporter.scala
```diff
@@ -51,7 +51,7 @@ class LanguageServerReporter(
     val pos = problem.position
     pos.sourceFile.toOption foreach { sourceFile: File =>
       problemsByFile.get(sourceFile) match {
-        case Some(xs: List[Problem]) => problemsByFile(sourceFile) = problem :: xs
+        case Some(xs: List[Problem]) => problemsByFile(sourceFile) = xs :+ problem
```
eed3si9n: As Guillaume noted on Gitter, could you replace this a better data structure for appending?

eed3si9n:
Thanks @tdroxler!

eed3si9n (changes requested):

main/src/main/scala/sbt/internal/server/LanguageServerReporter.scala
tdroxler: Back from holydays, sure I'll do it.

tdroxler: Ok I rebased and pushed a new commit using `ListBuffer`

eed3si9n (approved):
LGTM pending Travis

eed3si9n:
I hope you don't mind I added Mima exclusion commit on your branch.

tdroxler:
:+1: awesome thx

PR #4498, opened 2018-12-22
Don't guess config identifiers while constructing key index

dwijnand:


eed3si9n (dismissed):
Thanks!

dwijnand:
Either `ParserSpec` is wrong or it looks like guessing is required while constructing the key index? 🤷‍♂️

PR #4501, opened 2018-12-27
Allow trailing semicolon in multi commands

eatkins:
It was a mistake to disallow trailing semicolons for multi commands.
Firstly this was a mistake because previous versions of sbt supported a
trailing semi colon. It was also inconsistent with how commands work in
a regular shell (e.g. bash or zsh).


eed3si9n (approved):
Thanks!

PR #4508, opened 2019-01-08
Add LabeledFunctions to repo

eatkins:
I noticed that debugging settings that return functions is annoying
because often the setting is initialized as an anonymous function with a
useless toString method. To improve the debugging for users, I'm adding
a number of wrapper classes for functions that override the default
toString with a provided label.

I then used these functions to label all of the anonymous functions in
Watched.scala.

### DIFF
--- a/main-command/src/main/scala/sbt/internal/LabeledFunctions.scala
+++ b/main-command/src/main/scala/sbt/internal/LabeledFunctions.scala
@@ -1,0 +1,163 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.internal
+
+/**
+ * Provides extension methods that allow us to wrap FunctionN instances with another FunctionN that
+ * delegates to the input FunctionN, but overrides the toString method with a user specified
+ * label. Anonymous functions have the default java toString which is just the class name. Because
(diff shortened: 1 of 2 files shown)


eatkins:
For context, I have a rather massive PR with the watch dependency graph traversal stuff coming soon. This was a standalone part of the PR that I pulled out to make the upcoming review smaller.

main-command/src/main/scala/sbt/internal/LabeledFunctions.scala
```diff
@@ -0,0 +1,82 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.internal
+
+private[sbt] object LabeledFunctions {
```
eed3si9n, line 21: Could you write up the motivation for this change (what you wrote in PR description) in both commit message and Scaladoc here plz?

eed3si9n (changes requested):

main-command/src/main/scala/sbt/internal/LabeledFunctions.scala
eed3si9n: I noticed that the commit message is already there, so just the Scaladocs.

eatkins:
Ok. I added scaladoc to the non-private classes in LabeledFunctions.

eed3si9n (approved):

PR #4519, opened 2019-01-27
Implement Append for Function1

dwijnand:


main-settings/src/main/scala/sbt/Append.scala
```diff
@@ -8,109 +8,96 @@
...
+      def appendValues(a: Seq[T], b: Seq[V]): Seq[T] = a ++ b.map(x => (x: T))
       def appendValue(a: Seq[T], b: V): Seq[T] = a :+ (b: T)
     }
+
   @compileTimeOnly("This can be used in += only.")
   implicit def appendTaskValueSeq[T, V <: T]: Value[Seq[Task[T]], Initialize[Task[V]]] =
-    new Value[Seq[Task[T]], Initialize[Task[V]]] {
-      def appendValue(a: Seq[Task[T]], b: Initialize[Task[V]]): Seq[Task[T]] = ???
-    }
+    (_, _) => ???
```
eed3si9n, line 41: I think this might decrease the readability of the code.
SAM conversions are great for `onClick = _ => ()` kind of lifting, but in this case, my first reaction was confusion.
I'd ask to at least mention SAM conversion in the commit message, and maybe put a comment next to the code.

```scala
    (_, _) => ??? // lifted to Value using SAM conversion
```

eatkins: I think type annotations can help, but I do agree that sometimes SAM does make code less readable.

main-settings/src/test/scala/sbt/AppendSpec.scala
```diff
@@ -0,0 +1,31 @@
...
+  import SlashSyntax0._
+
+  def doSideEffect(): Unit = ()
+
+  Global / onLoad := (Global / onLoad).value.andThen { s =>
+    doSideEffect()
+    s
+  }
+
+  Global / onLoad += { (s: State) =>
```
eatkins, line 23: This is pretty cool.

build.sbt
```diff
@@ -547,6 +547,19 @@ lazy val mainSettingsProj = (project in file("main-settings"))
...
+      exclude[DirectMissingMethodProblem]("sbt.Append.appendSource"),
+      exclude[DirectMissingMethodProblem]("sbt.Append.appendList"),
+      exclude[DirectMissingMethodProblem]("sbt.Append.appendVectorImplicit"),
+      exclude[DirectMissingMethodProblem]("sbt.Append.appendClasspath"),
+      exclude[DirectMissingMethodProblem]("sbt.Append.appendListImplicit"),
+      exclude[DirectMissingMethodProblem]("sbt.Append.appendOption"),
+      exclude[DirectMissingMethodProblem]("sbt.Append.appendMap"),
+      exclude[DirectMissingMethodProblem]("sbt.Append.appendSeq"),
+      exclude[DirectMissingMethodProblem]("sbt.Append.appendSet"),
+      exclude[MissingClassProblem]("sbt.Append$Sequence"),
```
eatkins: Are these exclusions actually safe? Is it possible that plugins use them (implicitly)?

eatkins: I think it would be safer to leave them defined but deprecate them and remove the implicit modifier.

dwijnand: Good idea.

main-settings/src/main/scala/sbt/Append.scala
dwijnand: I'm inclined to think it's only confusing because it's a new thing we didn't use to have and so we're not used to.  And sbt being 2.12 only (that is: not still cross-building to 2.11) makes it one of the front-runners for 2.12-only features like SAM conversions.

Also, the instance you selected here is also special because it's entangled with one of sbt's rewrite macros.  If you pick some of the others I think they're comprehensible:

```scala
implicit def appendSeqN[T, V <: T]: Values[Seq[T], Seq[V]] = _ ++ _
implicit def appendList1[T, V <: T]: Value[List[T], V]     = _ :+ _
implicit def appendFunction[A, B]: Value[A => A, A => A]   = _.andThen(_)
```

dwijnand: > I'd ask to at least mention SAM conversion in the commit message, and maybe put a comment next to the code.

I'm happy to do whichever (or both) you prefer.

> I think type annotations can help, but I do agree that sometimes SAM does make code less readable.

IMO in the short-term, yes, but after a while you start seeing it as boilerplate.  But I'll add them if you prefer.

build.sbt
dwijnand: I'll defer this until we decide what to do about rewriting them to use SAM conversions.

main-settings/src/main/scala/sbt/Append.scala
eed3si9n: When the seeming types do not match up, I am inclined to assume "ok it must be implicits." Maybe eventually we'll rewire our thinking to "it must be implicits || SAM conversion." To me `_ ++ _` is equally head scratching.

In the meantime, having both commit comment and code comments would certainly benefit Ethan, me, and others who might wander into this code.

eatkins: To be clear, I am generally in favor of SAM wherever possible. I find that in roughly 10% of use cases, I find it confusing what is going on. There are also weird edge cases where SAM doesn't seem to work correctly. For example, I introduced:
```scala
trait WatchLogger { def debug(msg: => Any): Unit }
```
and I have found that if you try:
```scala
val logger: WatchLogger = m => println(m)
```
it will not do what you expect it to (it will print the name of an anonymous class). That's a specific issue with SAM and call by name, but I'm just pointing it out to illustrate where sometimes SAM can't be used.

I don't really have a horse in this specific race though.

dwijnand: > When the seeming types do not match up, I am inclined to assume "ok it must be implicits." Maybe eventually we'll rewire our thinking to "it must be implicits || SAM conversion." To me `_ ++ _` is equally head scratching.

It's the "seeming types do not match up" that'll change, with time.  You're totally cool with brain-compiling `val f: (String, String) => String = _ + _`.  With SAM conversions you'll have just expand that to also brain-compile `val f: Value[String, String] = _ + _`

eed3si9n: ok. I'm fine with the change, but since it might take a while for my old brain to rebuild plz add the comments etc.

dwijnand: > plz add the comments etc.

Will do.

main-settings/src/main/scala/sbt/Append.scala
```diff
@@ -84,4 +84,31 @@ object Append {
       b
     }
   }
+
+  // un-implicit, package-private (and JVM public) the old way of defining these
+  private[sbt] trait Sequence[A, -B, T] extends Value[A, T] with Values[A, B]
```
eatkins, line 91: Cool. I wasn't completely sure if private[sbt] was jvm public. I'm now wondering if I should change a bunch of watch related deprecated apis to package private instead so that we can turn -Xfatal-warnings on in the main project.

dwijnand: The upside of deprecated is it can include a message (and a version), while package private just says "isn't visible".

eatkins: Oh, we also might want to also deprecate these even though they're package private so we don't forget to remove them in sbt 2.

dwijnand: Done

main-settings/src/test/scala/sbt/AppendSpec.scala
dwijnand: Thanks 😄

eed3si9n:
Is it breaking on 

```scala
set resolvers ++= Nil
```

in https://github.com/sbt/sbt/blob/develop/sbt/src/sbt-test/actions/cross/test?

dwijnand:
Yeah, I can believe that.  There's something about variance and `Sequence` that's making my "simplify" not the noop refactor I wanted it to be.

I'll drop that change and resubmit with just the Function1 change.

PR #4527, opened 2019-01-31
Use custom classloader for scripted tests

### LINKED ISSUE #4514: scripted within scripted do not work with latest util version

## steps
publish "1.3.0-SNAPSHOT" in https://github.com/sbt/util
change utilVersion to "1.3.0-SNAPSHOT" in project/Dependencies.scala

## problem
scripted tests fail hard with java.lang.NoClassDefFoundError: Could not initialize class org.apache.logging.log4j.util.PropertiesUtil

## expectation
scripted works

## notes

sbt version: unreleased

eatkins:
We had previously used reflection to load the bridge class, but
continued using sbt's default classloader. This was problematic because
the metabuild could have a different classpath from that required by the
scripted tests.

Bonus: scalafmt

Fixes: #4514

### DIFF
--- a/project/Scripted.scala
+++ b/project/Scripted.scala
@@ -115,6 +115,7 @@ object Scripted {
           bootProperties: File,
           launchOpts: Array[String],
           prescripted: java.util.List[File],
+          instances: Int
       ): Unit
     }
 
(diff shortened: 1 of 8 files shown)


project/Scripted.scala
```diff
@@ -102,38 +104,43 @@ object Scripted {
...
-      bridge.runInParallel(
-        sourcePath,
-        bufferLog,
-        args.toArray,
-        launcher,
-        launchOpts.toArray,
-        callback,
-      )
-    } catch { case ite: InvocationTargetException => throw ite.getCause }
+      Thread.currentThread.setContextClassLoader(loader)
```
eed3si9n, line 124: Thanks for figuring this out.

eed3si9n:
Could you revert 67a0962b183628c962084e24e5fe8cc789bda9cf please?

project/Scripted.scala
eatkins: Necessity is the mother of invention :)

eatkins:
reverted

eatkins:
Evidently this doesn't actually fix the disabled tests. It does, however, allow me to run scripted tests on my machine using older versions of sbt.

eed3si9n (approved):

Issue #4528, opened 2019-01-31
References to undefined settings: subproject / Tut / run / runner


eed3si9n:
## steps

According to https://github.com/sbt/sbt-validator/issues/35, load cats-effect using the latest nightly.

## problem

```
microsite / Tut / run / runner from microsite / Tut / bgRun ((sbt.Defaults.mainBgRunTaskForConfig) Defaults.scala:1504)
```

## expectation

works.

## notes

runnerSettings changed in #4476. see https://github.com/sbt/sbt/pull/4476/files#r252521965
/cc @eatkins

PR #4529, opened 2019-01-31
Fix community build

eatkins:
The community build was broken for some projects because I broke builds
that relied on the unscoped definition of `runner`. To preserve legacy
behavior, I restore the old unscoped behavior and append the new scoped
runners that used the layered classloaders. This makes more sense
because the layered classloaders were specifically designed for the
Runtime and Test configurations and may not make sense in other
contexts.


eatkins:
Should hopefully fix #4528

eed3si9n (approved):
Thanks @eatkins!

PR #4512, opened 2019-01-11
Watch rewrite

eatkins:
- [x] I've read the [CONTRIBUTING](https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md) guidelines

PR #4533, opened 2019-01-31
Add meaningful toString to values in TypeFunctions

eatkins:
The kind projector syntax is confusing for beginners (e.g. myself 30
minutes before writing this). Moreover, the toString on these values
would be something like: TypeFunctions$$anon$5$aa23ra23 which made it
impossible to tell what transformer was being applied. I addressed these
issues by removing the syntactic sugar provided by the kind projector
plugin and overriding toString.

This should have no impact whatsoever on how sbt works, but should help
make debugging setting initialization easier.

This also removes all of the warning from this file in intelij, but that
wasn't actually the motivation.

### DIFF
--- a/internal/util-collection/src/main/scala/sbt/internal/util/TypeFunctions.scala
+++ b/internal/util-collection/src/main/scala/sbt/internal/util/TypeFunctions.scala
@@ -8,19 +8,22 @@
 package sbt.internal.util
 
 trait TypeFunctions {
+  import TypeFunctions._
   type Id[X] = X
   type NothingK[X] = Nothing
   sealed trait Const[A] { type Apply[B] = A }
   sealed trait ConstK[A] { type l[L[x]] = A }
   sealed trait Compose[A[_], B[_]] { type Apply[T] = A[B[T]] }
   sealed trait ∙[A[_], B[_]] { type l[T] = A[B[T]] }
+  type AnyLeft[T] = Left[T, Nothing]
+  type AnyRight[T] = Right[Nothing, T]
 
-  final val left = λ[Id ~> Left[?, Nothing]](Left(_))
-  final val right = λ[Id ~> Right[Nothing, ?]](Right(_))
-  final val some = λ[Id ~> Some](Some(_))
-  final def idFun[T]: T => T = (t: T) => t
-  final def const[A, B](b: B): A => B = _ => b
-  final def idK[M[_]]: M ~> M = λ[M ~> M](m => m)
+  final val left = λ[Id ~> AnyLeft](Left(_)).setToString("TypeFunctions.left")
+  final val right = λ[Id ~> AnyRight](Right(_)).setToString("TypeFunctions.right")
+  final val some = λ[Id ~> Some](Some(_)).setToString("TypeFunctions.some")
+  final def idFun[T]: T => T = ((t: T) => t).setToString("TypeFunctions.id")
+  final def const[A, B](b: B): A => B = ((_: A) => b).setToString(s"TypeFunctions.const($b)")
+  final def idK[M[_]]: M ~> M = λ[M ~> M](m => m).setToString("TypeFunctions.idK")
 
   def nestCon[M[_], N[_], G[_]](f: M ~> N): (M ∙ G)#l ~> (N ∙ G)#l =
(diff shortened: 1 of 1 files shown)


eatkins:
This was originally part of #4512 but I pulled it out so that it would be reviewed separately. I don't see this as a need to have if others strongly prefer the lambda syntax.

dwijnand:
FWIW, I don't like it.  If I were trying to solve this problem I'd probably spend the extra 23 lines explaining the syntax instead.

For `toString`, you can solve it for custom natural transformations but then you lose it again when you have functions:

```scala
scala> def f[A]: A => Right[Nothing, A] = (x: A) => Right(x)
f: [A]=> A => Right[Nothing,A]

scala> f[Int]
res0: Int => Right[Nothing,Int] = $$Lambda$1089/1887991591@56da7487
```

I'd probably fix this by allowing toString to be defined, maybe something like:

```scala
final val left = λ[Id ~> Left[?, Nothing]](Left(_))("TypeFunctions.left")
```

WRT IntelliJ it looks like polymorphic lambda values support has been recently fixed: https://youtrack.jetbrains.com/issue/SCL-10876

eatkins:
@dwijnand thanks for the feedback. I do see now that with the latest scala plugin for intellij that the kind projector is supported. I reworked the pr to keep the lambda syntax but added wrappers for `A => B` and `T ~> R`. If we don't want to wrap them, that's fine and I'll just drop 
https://github.com/sbt/sbt/pull/4533/commits/d98fda7a64933d911e21b5a0a1abbea0f3dcdf16 and only include 
https://github.com/sbt/sbt/pull/4533/commits/6fdac5bc9dd3d1ed5aca1e7d315eba41ed17b0d4

eatkins:
FWIW, my mild objection to the kind projector syntax is that it really isn't used that much and just adds another barrier to entry for hacking on sbt imo. I mind less now that I understand it reasonably well and intellij doesn't barf on it. I also hate the use of characters that I can't type in code, e.g. ∙ or λ.

internal/util-collection/src/main/scala/sbt/internal/util/TypeFunctions.scala
```diff
@@ -8,19 +8,22 @@
...
 
 trait TypeFunctions {
+  import TypeFunctions._
   type Id[X] = X
   type NothingK[X] = Nothing
   sealed trait Const[A] { type Apply[B] = A }
   sealed trait ConstK[A] { type l[L[x]] = A }
   sealed trait Compose[A[_], B[_]] { type Apply[T] = A[B[T]] }
   sealed trait ∙[A[_], B[_]] { type l[T] = A[B[T]] }
+  type AnyLeft[T] = Left[T, Nothing]
```
eatkins: The TypeFunctions.Ops extension class doesn't work without explicitly defining these types (which I know partially defeats the purpose of lambda syntax).

dwijnand (dismissed):
Much nicer.  I'd be down for switching to ASCII names. Also the PR title needs a rewrite.

eatkins:
@dwijnand one more look please? I decided to make the AnyLeft and AnyRight types private so they don't leak into the public api.

eatkins:
Now that I think about it, the explicit type annotations help a lot. Even if one is unfamiliar with the kind projector plugin, it's pretty obvious what is happening from the inputs and outputs.

dwijnand:
Still LGTM.

But (just for your information) not having to type annotate is one of the advantages of kind projector.  It's more relevant to when the type is more complex and the binding is implicit. But anyways it let's you read it: some is the lambda type constructor Id to type constructor Some, implemented as simply wrapping the input with Some.

dwijnand:
Oh and I agree it's good and clarifying to add type aliases.

eatkins:
That's an interesting reading that does clarify intent. I definitely don't naturally read it that way, but I also have had little exposure to programming with higher kinded types outside of sbt. I think there can be a tension between the code that is nicest for an expert and that which is nicest for a novice. I have no doubt that people who are working in the typelevel ecosystem would gag at my original pr :). In general, I have found sbt most painful when I'm working in the core applicative layer. My motivation was to help the next poor newbie who was trying to figure out what was going on.

dwijnand:
Yeah that's part of the power curve that Scala has.  It makes for such a wide range of code styles.

dwijnand (approved):

PR #4535, opened 2019-02-01
Stamped file

eatkins:
This makes a number of improvements to the api for caching and reading file stamps that I pulled out of #4512 for independent review.


main-command/src/main/scala/sbt/Stamped.scala
```diff
@@ -0,0 +1,74 @@
...
+import sbt.io.TypedPath
+import xsbti.compile.analysis.Stamp
+
+/**
+ * A File that has a compile analysis Stamp value associated with it. In general, the stamp method
+ * should be a cached value that can be read without doing any io. This can be used to improve
+ * performance anywhere where we need to check if files have changed before doing potentially
+ * expensive work.
+ */
+trait Stamped {
```
eed3si9n: I get a bit nervous when we expose the innards of Zinc.
Does this class need to be public, or can we keep this internal?

eatkins: Yeah. I agree. I'm thinking of making a class that, ignoring the name, would be something like:
```scala
trait Data { def stamp: Either[Long, String] }
object Data {
  private[sbt] def apply(stamp: Stamp): Data = new DataImpl(stamp)
  private final class DataImpl(private val s: Stamp) {
    override def stamp: Either[Long, String] =
      if (stamp.getHash.isPresent) Right(stamp.getHash.get)
      else Left(stamp.getLastModified.get())
  }
}
```

eatkins: The advantage of this too is that we can add fields to the cache in a binary compatible way if the need ever arises.

eatkins: I addressed this in 
https://github.com/sbt/sbt/commit/e8af828c7307e6fd71a0d037429aca67bf807209 (Add FileCacheEntry).

eed3si9n (approved):
Thanks @eatkins!

PR #4537, opened 2019-02-02
Allow watch loops

eatkins:

I had previously tried to prevent watch loops that could occur if managed sources write into the unmanaged source directory, but I don't think this was actually a good idea. We should not encourage such builds. In an ensuing PR, I also add the ability to explicitly exclude certain globs from triggering a build so even users who "do it wrong" imo can make their builds work.

This was pulled out of #4512. All of this code gets removed in that PR, but I still think it's a good idea to explicitly have standalone commits that document that these changes were deliberate.


eed3si9n (approved):
LGTM

eed3si9n:
Are there specific issues you can link related to watch loops?

eatkins:
The truth is that I really am not sure why I was trying to prevent watch loops. I didn't really explain why I was worried about them in https://github.com/sbt/sbt/commit/0de8345e33a40964edbb5420a0fd03a5a37e5bcb. At any rate, in the watch-rewrite PR, I add a setting watchExcludes that will allow a user to control what files are monitored which is a much better approach than special casing a certain category of loops.

Issue #923, opened 2013-10-17
Incremental compilation recompiles all the files every time


piotrga:
This is what I do:
1. clean
2. compile
3. compile

No. 3 recompiles all the sources, even though I haven't touched the files.
I am on scala 2.10.2, sbt 0.13.0 (but it happens on the 0.12.4 and 0.13.Beta2 too)

See my build.sbt:

```
name := "model_fx"

version := "1.0"

scalaVersion := "2.10.2"

dependencyClasspath in Compile ++= EclipseClasspath(Seq("lib")).classpath

scalaSource in Compile := file( "main" )

autoCompilerPlugins := true

libraryDependencies ++= Seq(
  compilerPlugin("org.scala-lang.plugins" % "continuations" % "2.10.2")
)

logLevel := Level.Debug

scalacOptions += """-Xplugin:D:\x\y\our-own-compiler-plugin.jar"""

scalacOptions ++= "-P:entity:asyncOff:v1 -target:jvm-1.6 -language:_".split(" ").toSeq
```

And this is the sbt log when I do the step 3 (so the sources are fully compiled by this time):

```
>
> compile
[debug]
[debug] Initial source changes:
[...]

PR #4539, opened 2019-02-03
Glob performance

eatkins:

This PR is built on top of #4535, #4537 and https://github.com/sbt/io/pull/216. I will update the PR as those other PRs are merged and a new io version is released.

High level, this replaces a number of file apis within sbt with glob based apis. This is good for performance because the glob based file look ups are generally routed through the file system cache. Even when the cache is not used, the performance should be good because of optimizations that I made in swoval for recursively listing a directory. Another benefit is that in watch, I can swap out the global file tree repository to determine which globs are actually used by a task. This enables watch to work even when the watch sources are unspecified or hidden from the dependency graph in a dynamic task.

I also actually implemented #3678, which allows clean to be scoped to the compile task and Compile / Test configurations. I copied some of the scripted tests from that #3678 and added a few additional scripted tests.


sbt/src/sbt-test/actions/compile-clean/test
```diff
@@ -0,0 +1,22 @@
+$ touch target/cant-touch-this
+
+> test:compile
+$ exists target/scala-2.12/classes/A.class
+$ exists target/scala-2.12/test-classes/B.class
+
+> test:compile::clean
```
eed3si9n: I think this should be `Test/clean`. In other words, `clean` should just be scoped to `Test` configuration.

eed3si9n (changes requested):

sbt/src/sbt-test/actions/compile-clean/test
eatkins: FWIW, I copied this test exactly from #3678. I think there is a valid use case for scoping both to the configuration and to the task separately. I might want to clear out the output files for just a specific task in a configuration but when i clean the configuration, I want to clear out the output files for all tasks in the configuration. Does that make sense? I don't mind dropping the task specific clean, but that was the motivation.

eatkins:
Sigh, windows is proving annoying wrt the clean task. Hopefully I'll have everything working by tomorrow afternoon.

eatkins:
~~It turns out that the java watchservice for windows is not ideal for the sbt use case. It reports way too many events. For example, if you open a file, an event is created. I need to add a native windows watch service (similar to the one i wrote for macos) that filters watch events to swoval. This shouldn't be too hard because the two apis are similar (inotify is completely different). Also, I already have the test suite written.~~ I'm less certain that I need a different watch service. It could be something else in swoval.

eatkins:
Yeah, there was a nasty performance bug in swoval: https://github.com/sbt/io/pull/222

eatkins:
I turned off the cache in this PR. I'll make it configurable, but not default, in #4512.

eed3si9n:
@eatkins Does this work for you locally?

```
[error] x source-dependencies/relative-source-error 
[error]  Cause of test exception: {line 3}  Command succeeded but failure was expected
```

sbt/src/sbt-test/actions/compile-clean/test
eed3si9n: I think task scoping here is confusing since all sorts of tasks can trigger a compilation, including `test`.

eatkins:
https://github.com/sbt/sbt/pull/4539/commits/ffa69ea5d690b5d8d548e0eefd98e83587a30de2

The commit message is arguably a little bit glib. Relative source directories are a bit tricky since sbt doesn't have a notion of current working directory so relative sources are implicitly relative to the sbt base directory. I guess marking the test PENDING is an option rather than deleting it, but it's not straightforward to make the previous semantics work with the new version of the swoval library (which automatically converts non relative paths to absolute paths internally at interface boundaries).

sbt/src/sbt-test/actions/compile-clean/test
eatkins: To play devil's advocate, what if I just want to remove the jars generated by `package`? What if I want to remove the class files but not the jars? For me, it's more confusing if clean cannot be scoped at the task level. Most users are just going to use clean at the config level but I don't see why we need to exclude it at the task level.

eed3si9n: One of the reasons is that I think task-scoped tasks are confusing concepts to users, and I want to avoid it as much as possible.

eed3si9n:
As long as https://github.com/sbt/sbt/issues/923 remains fixed, we should be ok.

sbt/src/sbt-test/actions/compile-clean/test
eatkins: I get that, but the task scoped task is only relevant for advanced users who even know that syntax. As is, the config scoped clean works as expected and most users would never even think to try the task scoped version. Personally I think the real problem is the legacy `:` based syntax. When using the slash syntax, I don't think it's that confusing. Would it bother you less if the test was written `Test/compile/clean`? That works.

eatkins: Oh, ugh. I see what you're saying a bit more now. Compile is really a special case because of the default configurations. I think my argument is more for arbitrary tasks, but there is all kinds of special casing around the compile task. That being said, I still think there is an argument that `compile/clean` should only delete the class files while `Compile/clean` would delete everything generated by the configuration.

eed3si9n: When I type `clean`, what I often want is "clear the caching behavior and start from scratch", which doesn't necessarily have to do with wiping out the `target/` directory. Having that work for one task is tricky, let alone making sure it works consistently for all tasks transitively. Config-scoped clean deleting `classes/` is much more low-tech, but addresses a use case in a consistent way.

eatkins: I disagree but I don't care enough to continue the debate so I'll just make it work for just the configuration.

eatkins: https://github.com/sbt/sbt/compare/2da32eccd6f3e242d2859fa546d76f805b55a17d..f26afe6681da174e80e2ca6db8e638e1abf732cc

eed3si9n (approved):
LGTM pending Travis

PR #4544, opened 2019-02-13
Run gc when idle

eatkins:
I often find that when I run a command it takes a long time to start up
because sbt triggers a full gc. To improve the ux, I update the command
exchange to periodically run full gc while it's waiting for a command to
run.

Bonus: optimize imports

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -884,7 +884,11 @@ object BuiltinCommands {
     val exchange = StandardMain.exchange
     val s1 = exchange run s0
     exchange publishEventMessage ConsolePromptEvent(s0)
-    val exec: Exec = exchange.blockUntilNextExec
+    val minGCInterval = Project
+      .extract(s1)
+      .getOpt(Keys.minForcegcInterval)
+      .getOrElse(GCUtil.defaultMinForcegcInterval)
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/Main.scala
```diff
@@ -884,7 +885,8 @@ object BuiltinCommands {
     val exchange = StandardMain.exchange
     val s1 = exchange run s0
     exchange publishEventMessage ConsolePromptEvent(s0)
-    val exec: Exec = exchange.blockUntilNextExec
+    val minGCInterval = Project.extract(s1).getOpt(Keys.minForcegcInterval).getOrElse(1.minute)
```
eed3si9n: We probably should use `GCUtil.defaultMinForcegcInterval` instead of `1.minute`.
We can also take this opportunity to bump the min interval.

eed3si9n (changes requested):
minor suggestion.

main/src/main/scala/sbt/Main.scala
eatkins: Good points. I re-pushed with a value of 10.minutes, but I more or less plucked that out of thin air.

eatkins: I also reworked it to defer the full GC until the user has been idle for at least one minute. It will also only force full GC at most once between tasks.

eed3si9n (approved):

Issue #4461, opened 2018-11-21
Clean up task progress


eed3si9n:
- Clean up some of the issues of it wiping out text
- Make it configurable via settings?

eed3si9n:
@eatkins I can't reproduce the text wiping issue with hello world app:

[![asciicast](https://asciinema.org/a/RdVvxs7aTb9Vor1rJL2sD1e9T.svg)](https://asciinema.org/a/RdVvxs7aTb9Vor1rJL2sD1e9T)

I'm on macOS Mojave + iTerm2.

eatkins:
Aah. I use the built in Terminal.app. I verified that if I use iterm2, then I don't see the issue that I showed you.

eatkins:
That was a great clue. If I change the Terminal.app terminal emulation mode to vt100, I don't see the issue. Previously I had it set to xterm-256color.

eatkins:
I switched back to xterm-256color and it still seems to be working, so I am not sure what my issue was.

eatkins:
I do still see a bit of weirdness even with iterm:
![Output sample](https://media.giphy.com/media/5n7VbxEzxMW3GvINqf/giphy.gif)
This is in the sbt project itself where I'm running a local version of sbt off of the latest develop branch. In my ideal world, the previous task would not be deleted and the progress tasks would be fully cleared.

eatkins:
In my ideal world, the progress tasks would be fully cleared, but my previous command would not be.

eed3si9n:
> In my ideal world, the previous task would not be deleted and the progress tasks would be fully cleared.

Agreed on both points.

eed3si9n:
Here's a recent SNAPSHOT. The progress status is cleared.

[![asciicast](https://asciinema.org/a/6xOQSpHucre4c8QHU342CFmhy.png?t=1:30)](https://asciinema.org/a/6xOQSpHucre4c8QHU342CFmhy?t=1:30)

PR #4556, opened 2019-03-08
 Make sure progress thread gets shutdown

eed3si9n:
Ref #4461
Ref https://github.com/sbt/util/pull/193


main/src/main/scala/sbt/EvaluateTask.scala
```diff
@@ -412,6 +412,7 @@ object EvaluateTask {
     def shutdown(): Unit = {
       // First ensure that all threads are stopped for task execution.
       shutdownThreads()
+      config.progressReporter.stop()
```
eed3si9n, line 415: This gets called at finally.

PR #4560, opened 2019-03-15
Unseal Parser

eatkins:
Although this is technically in the internal package, it is exposed to
users when they write a custom input task. I do not think that we should
prevent users/plugin authors from writing their own parser
implementations if there is a different library they prefer. By my
count, there are 21 implementations of this interface in sbt, so it's
unlikely that there is much benefit from a pattern matching perspective.

### DIFF
--- a/internal/util-complete/src/main/scala/sbt/internal/util/complete/Parser.scala
+++ b/internal/util-complete/src/main/scala/sbt/internal/util/complete/Parser.scala
@@ -18,7 +18,7 @@ import sbt.internal.util.Util.{ makeList, separate }
  * The methods in this trait are what must be implemented to define a new Parser implementation, but are not typically useful for common usage.
  * Instead, most useful methods for combining smaller parsers into larger parsers are implicitly added by the [[RichParser]] type.
  */
-sealed trait Parser[+T] {
+trait Parser[+T] {
   def derive(i: Char): Parser[T]
   def resultEmpty: Result[T]
   def result: Option[T]


internal/util-complete/src/main/scala/sbt/internal/util/complete/Parser.scala
```diff
@@ -18,7 +18,7 @@ import sbt.internal.util.Util.{ makeList, separate }
  * The methods in this trait are what must be implemented to define a new Parser implementation, but are not typically useful for common usage.
  * Instead, most useful methods for combining smaller parsers into larger parsers are implicitly added by the [[RichParser]] type.
  */
-sealed trait Parser[+T] {
+trait Parser[+T] {
```
eed3si9n, line 21: I was curious how long back the `sealed` goes to, and it went back all the way to the beginning.
https://github.com/sbt/sbt-zero-thirteen/commit/8e6b42180d7ab45a3f31fe44f4d73dcbf59a4186

eed3si9n:
Is this a theoretical possibility, or do you have specific use case in mind?
Not that I am totally against opening this up, but there's benefit of limiting the power.

eatkins:
It is more theoretical at this point. I have thought about how I might want to implement certain tasks that do custom parsing and while the core sbt parsers have worked fine for me, I might appreciate the option of using a different library. Moreover, the fact that sbt uses a custom parser was something that Li Haoyi specifically called out in one of his blog posts or reddit posts about sbt. I feel like this is a simple way to squash an objection to sbt. It doesn't really matter what parser sbt uses internally for its commands but unsealing the Parser allows users to implement their own however they wish via a simple wrapper around the library parser.

eatkins:
And, yeah, I also looked at the git history and unfortunately those early commits do not contain a lot of context around why certain decisions were made. I think in general it makes sense to seal anything internal but at some point the Parser leaked into the public api at which point I think it would have made sense to unseal it or provide a simpler interface rather than expose the internal interface but that ship already sailed.

eatkins:
Oh, and one more thing: I looked at `Result` and `Completions` because those are also sealed and instances must be provided to fully implement `Parser`. It looked to me like those can remain sealed without preventing the user from being able to implement `Parser`.

eed3si9n (approved):

PR #4561, opened 2019-03-18
open up ExecuteProgress, and adds a few keys

### LINKED ISSUE #4461: Clean up task progress

eed3si9n:
Fixes #4461

This opens up ExecuteProgress API that's been around under private[sbt].
Since the state passing mechanism hasn't been used, I got rid of it.

The build user can configure the build using two keys Boolean `taskProgress` and `State => Seq[TaskProgress]` `progressReports`. `taskProgress` is lightweight key on/off switch for the super shell that can be used as follows:

```scala
Global / SettingKey[Boolean]("taskProgress") := false
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -291,7 +291,10 @@ lazy val taskProj = (project in file("tasks"))
     name := "Tasks",
     mimaSettings,
     mimaBinaryIssueFilters ++= Seq(
-      exclude[ReversedMissingMethodProblem]("sbt.ExecuteProgress.stop")
+      // ok because sbt.ExecuteProgress has been under private[sbt]
+      exclude[IncompatibleResultTypeProblem]("sbt.ExecuteProgress.initial"),
+      exclude[DirectMissingMethodProblem]("sbt.ExecuteProgress.*"),
+      exclude[ReversedMissingMethodProblem]("sbt.ExecuteProgress.*"),
     )
   )
   .configure(addSbtUtilControl)
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -162,19 +162,21 @@ object EvaluateTask {
   import Keys.state
 
   lazy private val sharedProgress = new TaskTimings(reportOnShutdown = true)
-
-  // sbt-pgp calls this
-  private[sbt] def defaultProgress(): ExecuteProgress[Task] = ExecuteProgress.empty[Task]
-  private[sbt] def defaultProgress(currentRef: ProjectRef): ExecuteProgress[Task] =
+  def taskTimingProgress: Option[ExecuteProgress[Task]] =
     if (java.lang.Boolean.getBoolean("sbt.task.timings")) {
       if (java.lang.Boolean.getBoolean("sbt.task.timings.on.shutdown"))
-        sharedProgress
+        Some(sharedProgress)
       else
-        new TaskTimings(reportOnShutdown = false)
-    } else {
-      if (ConsoleAppender.showProgress) new TaskProgress(currentRef)
-      else ExecuteProgress.empty[Task]
-    }
+        Some(new TaskTimings(reportOnShutdown = false))
+    } else None
+
+  def taskProgress(state: State): ExecuteProgress[Task] = {
+    new TaskProgress(Project.extract(state).currentRef)
+  }
+
+  // sbt-pgp calls this
+  @deprecated("No longer used", "1.3.0")
+  private[sbt] def defaultProgress(): ExecuteProgress[Task] = ExecuteProgress.empty[Task]
 
   val SystemProcessors = Runtime.getRuntime.availableProcessors
 
@@ -232,13 +234,22 @@ object EvaluateTask {
       state: State
   ): ExecuteProgress[Task] = {
     import Types.const
-    val maker: State => Keys.TaskProgress = getSetting(
-      Keys.executeProgress,
-      const(new Keys.TaskProgress(defaultProgress(extracted.currentRef))),
+    val maker: State => Seq[Keys.TaskProgress] = getSetting(
+      Keys.progressReports,
+      const(Seq()),
       extracted,
       structure
     )
-    maker(state).progress
+    val reporters = maker(state) map { _.progress }
+    // configure the logger for super shell
+    ConsoleAppender.setShowProgress((reporters collect {
+      case p: TaskProgress => ()
+    }).nonEmpty)
+    reporters match {
+      case xs if xs.isEmpty   => ExecuteProgress.empty[Task]
+      case xs if xs.size == 1 => xs.head
+      case xs                 => ExecuteProgress.aggregate[Task](xs)
+    }
   }
   // TODO - Should this pull from Global or from the project itself?
   private[sbt] def forcegc(extracted: Extracted, structure: BuildStructure): Boolean =
@@ -258,7 +269,7 @@ object EvaluateTask {
       extracted: Extracted,
       structure: BuildStructure
   ): T =
-    key in extracted.currentRef get structure.data getOrElse default
+    (key in extracted.currentRef).get(structure.data).getOrElse(default)
 
   def injectSettings: Seq[Setting[_]] = Seq(
     (state in Global) ::= dummyState,
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -456,6 +456,18 @@ object Keys {
   val (executionRoots, dummyRoots) = Def.dummy[Seq[ScopedKey[_]]]("executionRoots", "The list of root tasks for this task execution.  Roots are the top-level tasks that were directly requested to be run.")
   val state = Def.stateKey
   val streamsManager = Def.streamsManagerKey
+  // wrapper to work around SI-2915
+  final class TaskProgress(val progress: ExecuteProgress[Task])
+  object TaskProgress {
+    def apply(progress: ExecuteProgress[Task]): TaskProgress = new TaskProgress(progress)
+  }
+  val useSuperShell = settingKey[Boolean]("Enables (true) or disables the super shell.")
+  // This key can be used to add custom ExecuteProgress instances
+  val progressReports = settingKey[State => Seq[TaskProgress]]("A function that returns a list of progress reporters.").withRank(DTask)
+  private[sbt] val postProgressReports = settingKey[Unit]("Internally used to modify logger.").withRank(DTask)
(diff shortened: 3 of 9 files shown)


main/src/main/scala/sbt/internal/TaskProgress.scala
```diff
@@ -167,4 +167,16 @@ private[sbt] object TaskProgress {
   def cursorUp(n: Int): String = s"\u001B[${n}A"
   def cursorDown(n: Int): String = s"\u001B[${n}B"
   final val CursorDown1 = cursorDown(1)
+
+  def isEnabled: Boolean =
+    ConsoleAppender.formatEnabledInEnv && sys.props
+      .get("sbt.progress")
+      .flatMap({ s =>
```
eatkins: Doesn't really matter but the parens are unnecessary.

tasks/src/main/scala/sbt/ExecuteProgress.scala
```diff
@@ -14,25 +14,22 @@ import sbt.internal.util.RMap
  * All methods are called from the same thread except `started` and `finished`,
  * which is called from the executing task's thread.
  * All methods should return quickly to avoid task execution overhead.
- *
- * This class is experimental and subject to binary and source incompatible changes at any time.
  */
-private[sbt] trait ExecuteProgress[F[_]] {
-  type S
```
eatkins, line 21: I like getting rid of this type.

build.sbt
```diff
@@ -291,7 +291,10 @@ lazy val taskProj = (project in file("tasks"))
     name := "Tasks",
     mimaSettings,
     mimaBinaryIssueFilters ++= Seq(
-      exclude[ReversedMissingMethodProblem]("sbt.ExecuteProgress.stop")
+      // ok because sbt.ExecuteProgress has been under private[sbt]
```
eatkins, line 294: Hopefully the community build passes. It wouldn't shock me if there are some plugins that use it given how far back it goes, but obviously a plugin author uses private[sbt] apis at their own risk.

main/src/main/scala/sbt/EvaluateTask.scala
```diff
@@ -162,19 +162,23 @@ object EvaluateTask {
   import Keys.state
 
   lazy private val sharedProgress = new TaskTimings(reportOnShutdown = true)
-
-  // sbt-pgp calls this
-  private[sbt] def defaultProgress(): ExecuteProgress[Task] = ExecuteProgress.empty[Task]
-  private[sbt] def defaultProgress(currentRef: ProjectRef): ExecuteProgress[Task] =
+  def taskTimingProgress: Option[ExecuteProgress[Task]] =
     if (java.lang.Boolean.getBoolean("sbt.task.timings")) {
```
eatkins, line 166: Not directly related to this PR, but it seems like a bummer that this is a system property since system properties aren't really discoverable. I personally would have saved myself some time if I'd known about it.

main/src/main/scala/sbt/EvaluateTask.scala
```diff
@@ -162,19 +162,23 @@ object EvaluateTask {
...
-    } else {
-      if (ConsoleAppender.showProgress) new TaskProgress(currentRef)
-      else ExecuteProgress.empty[Task]
-    }
+        Some(new TaskTimings(reportOnShutdown = false))
+    } else None
+
+  def taskProgress(state: State): ExecuteProgress[Task] = {
+    val extracted = Project.extract(state)
+    import extracted.currentRef
```
eatkins: I'm not sure I see the point of this import. I might even be tempted to just write this: `new TaskProgress(Project.extract(state).currentRef)`. Another side note: `Project.extract(state)` is so ubiquitous that it might make sense to add `def extracted = Project.extract(state)` to `StateOps`.

main/src/main/scala/sbt/EvaluateTask.scala
```diff
@@ -258,7 +271,7 @@ object EvaluateTask {
       extracted: Extracted,
       structure: BuildStructure
   ): T =
-    key in extracted.currentRef get structure.data getOrElse default
+    (key in extracted.currentRef).get(structure.data).getOrElse(default)
```
eatkins, line 272: +1

main/src/main/scala/sbt/Keys.scala
```diff
@@ -456,6 +456,17 @@ object Keys {
   val (executionRoots, dummyRoots) = Def.dummy[Seq[ScopedKey[_]]]("executionRoots", "The list of root tasks for this task execution.  Roots are the top-level tasks that were directly requested to be run.")
   val state = Def.stateKey
   val streamsManager = Def.streamsManagerKey
+  // wrapper to work around SI-2915
+  final class TaskProgress(val progress: ExecuteProgress[Task])
+  object TaskProgress {
+    def apply(progress: ExecuteProgress[Task]): TaskProgress = new TaskProgress(progress)
+  }
+  val taskProgress = settingKey[Boolean]("Enables (true) or disables the super shell.")
```
eatkins: Would it make sense to call this `useSuperShell`?

main/src/main/scala/sbt/Keys.scala
```diff
@@ -456,6 +456,17 @@ object Keys {
...
+  val progressReports = settingKey[State => Seq[TaskProgress]]("Task progress reporters.").withRank(DTask)
```
eatkins: It isn't really clear to me at a glance exactly what `progressReports` does or how it's connected to the `taskProgress` key. I wouldn't mind a comment or a more detailed key description.

main/src/main/scala/sbt/internal/TaskProgress.scala
```diff
@@ -33,9 +31,11 @@ private[sbt] final class TaskProgress(currentRef: ProjectRef) extends ExecutePro
   private[this] val isAllCompleted = new AtomicBoolean(false)
   private[this] val isStopped = new AtomicBoolean(false)
 
-  override def initial: Unit = ()
+  override def initial: Unit = {
```
eatkins: It's a little confusing that this method has no parens while the other `initial` method does.

main/src/main/scala/sbt/internal/TaskProgress.scala
```diff
@@ -167,4 +167,16 @@ private[sbt] object TaskProgress {
   def cursorUp(n: Int): String = s"\u001B[${n}A"
   def cursorDown(n: Int): String = s"\u001B[${n}B"
   final val CursorDown1 = cursorDown(1)
+
+  def isEnabled: Boolean =
+    ConsoleAppender.formatEnabledInEnv && sys.props
+      .get("sbt.progress")
+      .flatMap({ s =>
+        ConsoleAppender.parseLogOption(s) match {
```
eatkins: Admittedly superficial, but `s` is used so frequently for state throughout sbt that I assumed this was an instance of `State` not `String`.

tasks/src/main/scala/sbt/Execute.scala
```diff
@@ -153,7 +153,7 @@ private[sbt] final class Execute[F[_] <: AnyRef](
 
     results(node) = result
     state(node) = Done
-    progressState = progress.completed(progressState, node, result)
+    progress.completed(node, result)
```
eatkins: It's a little unfortunate that this is named completed rather than something like `onCompleted` because it isn't completely obvious that is being called for side effects only. But I know that this api has been around for a while.

eatkins (commented):
Most of my suggestions/comments are largely superficial. I don't feel strongly about any of them but they were just my initial reactions when reading through the changes. I don't object to merging as but am holding off on clicking approve.

main/src/main/scala/sbt/internal/TaskProgress.scala
eed3si9n: It probably should have parens here.

tasks/src/main/scala/sbt/Execute.scala
eed3si9n: Since I am getting rid of `type S`, this is an opportunity to make breaking changes.

main/src/main/scala/sbt/Keys.scala
eed3si9n: Let's do that.

main/src/main/scala/sbt/Keys.scala
eed3si9n: ok.

main/src/main/scala/sbt/internal/TaskProgress.scala
eed3si9n: I'll change to `str`.

tasks/src/main/scala/sbt/Execute.scala
eed3si9n:
```scala
trait ExecuteProgress[F[_]] {
  def initial(): Unit

  /**
   * Notifies that a `task` has been registered in the system for execution.
   * The dependencies of `task` are `allDeps` and the subset of those dependencies that
   * have not completed are `pendingDeps`.
   */
  def afterRegistered(task: F[_], allDeps: Iterable[F[_]], pendingDeps: Iterable[F[_]]): Unit

  /**
   * Notifies that all of the dependencies of `task` have completed and `task` is therefore
   * ready to run.  The task has not been scheduled on a thread yet.
   */
  def afterReady(task: F[_]): Unit

  /**
   * Notifies that the work for `task` is starting after this call returns.
   * This is called from the thread the task executes on, unlike most other methods in this callback.
   * It is called immediately before the task's work starts with minimal intervening executor overhead.
   */
  def beforeWork(task: F[_]): Unit

  /**
   * Notifies that the work for `task` work has finished.  The task may have computed the next task to
   * run, in which case `result` contains that next task wrapped in Left.  If the task produced a value
   * or terminated abnormally, `result` provides that outcome wrapped in Right.  The ultimate result of
   * a task is provided to the `completed` method.
   * This is called from the thread the task executes on, unlike most other methods in this callback.
   * It is immediately called after the task's work is complete with minimal intervening executor overhead.
   */
  def afterWork[A](task: F[A], result: Either[F[A], Result[A]]): Unit

  /**
   * Notifies that `task` has completed.
   * The task's work is done with a final `result`.
   * Any tasks called by `task` have completed.
   */
  def afterCompleted[A](task: F[A], result: Result[A]): Unit

  /** All tasks have completed with the final `results` provided. */
  def afterAllCompleted(results: RMap[F, Result]): Unit

  /** Notifies that either all tasks have finished or cancelled. */
  def stop(): Unit
}
```

eatkins: Nice. I like this.

eatkins (approved):

PR #4576, opened 2019-03-24
output Chrome traces on -Dsbt.traces=true

eed3si9n:
This is a feature implemented originally by @retronym as https://github.com/retronym/sbt/pull/2.
When `-Dsbt.traces=true`, sbt will output `target/traces/build.trace` file that can be viewed interactively in Chrome with `chrome://tracing`.

chrome___tracing

This PR also lifts out common behavior among the `ExecuteProgress[Task]` implementations.

### DIFF
--- a/main/src/main/scala/sbt/internal/AbstractTaskProgress.scala
+++ b/main/src/main/scala/sbt/internal/AbstractTaskProgress.scala
@@ -1,0 +1,81 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt
+package internal
+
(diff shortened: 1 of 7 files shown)


main/src/main/scala/sbt/internal/AbstractTaskProgress.scala
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+package sbt
+package internal
+
+import java.util.concurrent.ConcurrentHashMap
+import scala.collection.concurrent.TrieMap
+import scala.collection.JavaConverters._
+
+abstract class AbstractTaskExecuteProgress extends ExecuteProgress[Task] {
```
eatkins: Maybe private[sbt] this even though it's in the internal package? It feels like ExecuteProgress is still a moving target at this point so we should keep everything private[sbt] until 1.4.0.

eed3si9n: Yea. That probably makes sense.

PR #4577, opened 2019-03-24
use sbt 1.3.0-M2

eed3si9n:


eatkins (approved):
LGTM

eed3si9n:
Bintray is being super flakey about `curl -L --head http://repo.scala-sbt.org/scalasbt/maven-releases/org/scala-sbt/sbt-launch/1.3.0-M2/sbt-launch.jar`, which is downloaded by sbt-extras.
I had to restart some jobs multiple times to get to green.

PR #4585, opened 2019-03-28
Jline

eatkins:

### DIFF
--- a/internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
+++ b/internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
@@ -9,9 +9,11 @@ package sbt.internal.util
 
 import jline.console.ConsoleReader
 import jline.console.history.{ FileHistory, MemoryHistory }
-import java.io.{ File, InputStream, FileInputStream, FileDescriptor, FilterInputStream }
+import java.io.{ File, FileDescriptor, FileInputStream, FilterInputStream, InputStream }
+
 import complete.Parser
-import scala.concurrent.duration.Duration
+
+import scala.concurrent.duration._
 import scala.annotation.tailrec
 
 abstract class JLine extends LineReader {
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -264,15 +264,6 @@ object Defaults extends BuildCommon {
           Clean.deleteContents(tempDirectory, _ => false)
       },
       useSuperShell :== sbt.internal.TaskProgress.isEnabled,
-      progressReports := { (s: State) =>
```
eed3si9n: I consider progressReports to be somewhat stable actually.
super shell is one implementation on it.

internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
```diff
@@ -120,7 +122,7 @@ private[sbt] object JLine {
   protected[this] val originalIn = new FileInputStream(FileDescriptor.in)
 
   private[sbt] def makeInputStream(injectThreadSleep: Boolean): InputStream =
-    if (injectThreadSleep) new InputStreamWrapper(originalIn, Duration("50 ms"))
+    if (injectThreadSleep) new InputStreamWrapper(originalIn, 2.milliseconds)
```
eed3si9n, line 125: 👍

main/src/main/scala/sbt/internal/TaskProgress.scala
```diff
@@ -25,7 +25,7 @@ private[sbt] final class TaskProgress
   private[this] val isStopped = new AtomicBoolean(false)
 
   override def initial(): Unit = {
-    ConsoleAppender.setTerminalWidth(JLine.usingTerminal(_.getWidth))
+    ConsoleAppender.setTerminalWidth(JLine.terminal.getWidth)
```
eed3si9n, line 28: 👍

main/src/main/scala/sbt/Defaults.scala
eed3si9n: One thing I might change about the `progressReports` is maybe dropping `s: State` and just making it a Seq of TaskProgress.

eatkins: Aah, ok. That makes sense. I understand what the key is better now than when I first wrote the commit. It was in fact the State parameter that was especially throwing me off.

eed3si9n (changes requested):
I am happy with the first two commits, so could you split the last one out?

eatkins:
Yep. Been working on some other stuff but will re-push once I get to this.

eed3si9n (approved):

PR #4589, opened 2019-03-28
Exclude all sbt modules from metabuild resolution

### LINKED ISSUE #4582: [1.3.0-M2] Symbol 'type sbt.io.PathFinder.Combinator' is missing from the classpath
## steps

https://jenkins.scala-sbt.org:8592/job/sbt-community/641/consoleFull

```
[22:43:20] [akka-stuff] Fetching https://github.com/akka/akka.git
```

## problem

```
[error] /xxxx/akka/project/AddLogTimestamps.scala:19:68: Symbol 'type sbt.io.PathFinder.Combinator' is missing from the classpath.
[error] This symbol is required by 'method sbt.io.BaseSyntax.singlePathPathFinderCombinator'.
[error] Make sure that type Combinator is in your classpath and check for conflicting dependencies with `-Ylog-classpath`.
[error] A full rebuild may help if 'BaseSyntax.class' was compiled against an incompatible version of sbt.io.PathFinder.
[error]   val enableTimestamps: Boolean = CliOption("akka.log.timestamps", false).get
[error]                                                                    ^
[error] one error found
```

## expectation

works.

## notes

See also https://twitter.com/nafg613/status/1110328280643719168

eed3si9n:
Fixes #4582

#4443 introduced a perf enhacement of exluding sbt out of the metabuild, and instead appending the list of JARs resolved by the sbt launcher to the classpath.
This strategy worked in most cases, but it seems like some plugins are explicitly depending on IO module. In those cases, old IO would come before the new IO in the classpath ordering, resulting to "Symbol X is missing from classpath" error.
This fixes the issue by excluding all modules whose organization is `org.scala-sbt`.

As an escape hatch, I am adding a new key `reresolveSbtArtifacts`, which can be used to opt out of this behavior.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -175,6 +175,7 @@ object Defaults extends BuildCommon {
       javaOptions :== Nil,
       sbtPlugin :== false,
       isMetaBuild :== false,
+      reresolveSbtArtifacts :== false,
       crossPaths :== true,
       sourcePositionMappers :== Nil,
       artifactClassifier in packageSrc :== Some(SourceClassifier),
@@ -1897,6 +1898,7 @@ object Classpaths {
       unmanagedClasspath := unmanagedDependencies.value,
       managedClasspath := {
         val isMeta = isMetaBuild.value
+        val force = reresolveSbtArtifacts.value
         val app = appConfiguration.value
         val sbtCp0 = app.provider.mainClasspath.toList
         val sbtCp = sbtCp0 map { Attributed.blank(_) }
@@ -1905,7 +1907,7 @@ object Classpaths {
           classpathTypes.value,
           update.value
         )
-        if (isMeta) mjars ++ sbtCp
+        if (isMeta && !force) mjars ++ sbtCp
         else mjars
       },
       exportedProducts := trackedExportedProducts(TrackLevel.TrackAlways).value,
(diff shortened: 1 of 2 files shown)


eatkins (dismissed):
LGTM. Thanks!

eatkins:
Hmm...I got all kinds of errors when I published locally and tried to start up sbt with the new version. For example: [error] ContrabandConfig.scala:1:12: object contraband is not a member of package sbt  
[error] import sbt.contraband.ast._

eed3si9n:
I guess I'd have to try that too then.
I did publish a snapshot and checked with Akka build before sending in the PR.

eed3si9n:
I see it when I run sbt's build:

```
[info] Compiling 12 Scala sources to /Users/eed3si9n/work/sbt-modules/sbt/project/target/scala-2.12/sbt-1.0/classes ...
[error] /Users/eed3si9n/work/sbt-modules/sbt/project/ContrabandConfig.scala:1:12: object contraband is not a member of package sbt
[error] import sbt.contraband.ast._
```

I guess some of our plugins use org.scala-sbt organization.

eed3si9n:
I could change the organization of Contraband plugin, or exclude major sbt components like io, lm, and zinc individually.

eed3si9n:
I tested locally using both sbt/sbt and akka, so it should be ok.

PR #4590, opened 2019-03-30
Disable supershell in CI

eatkins:
The supershell output is distracting in CI. I added a system property,
sbt.ci, to explicitly set whether or not sbt is running in a ci build.
It was not at all obvious to me if the BUILD_NUMBER or CI environment
variables were set on travis or appveyor.


eatkins:
I cannot figure out why but this change works on travis, but not appveyor.

eatkins:
It does work on appveyor when I explicitly set -Dsbt.supershell=never. Again, I don't understand why using insideCI works on travis and not appveyor, but I'm not interested in investigating further with everything else on my plate.

eed3si9n (approved):

Issue #4591, opened 2019-03-31
(Compile / packageSrc) java.util.zip.ZipException: duplicate entry


eed3si9n:
## steps

1. git clone dotty
2. git checkout e4c4e0aae27c1316cae0e63278d7e4549f2890ea
3. open sbt with a recent nightly
4. sbt-dotty/publishLocal

## problem

```
sbt:dotty> sbt-dotty/publishLocal
[info] Wrote /xxx/dotty/sbt-dotty/target/scala-2.12/sbt-1.0/sbt-dotty-0.3.2-SNAPSHOT.pom
[error] java.util.zip.ZipException: duplicate entry: dotty/tools/sbtplugin/DottyPlugin.scala
[error] 	at java.util.zip.ZipOutputStream.putNextEntry(ZipOutputStream.java:232)
[error] 	at java.util.jar.JarOutputStream.putNextEntry(JarOutputStream.java:109)
[error] 	at sbt.io.IO$.addFileEntry$1(IO.scala:653)
[error] 	at sbt.io.IO$.$anonfun$writeZip$3(IO.scala:662)
[error] 	at sbt.io.IO$.$anonfun$writeZip$3$adapted(IO.scala:662)
[error] 	at scala.collection.Iterator.foreach(Iterator.scala:941)
[error] 	at scala.collection.Iterator.foreach$(Iterator.scala:941)
[error] 	at scala.collection.AbstractIterator.foreach(Iterator.scala:1429)
[error] 	at scala.collection.IterableLike.foreach(IterableLike.scala:74)
[error] 	at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
[error] 	at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
[error] 	at sbt.io.IO$.writeZip(IO.scala:662)
[error] 	at sbt.io.IO$.$anonfun$archive$1(IO.scala:616)
[error] 	at sbt.io.IO$.$anonfun$archive$1$adapted(IO.scala:613)
[error] 	at sbt.io.IO$.$anonfun$withZipOutput$1(IO.scala:698)
[error] 	at sbt.io.IO$.$anonfun$withZipOutput$1$adapted(IO.scala:687)
[error] 	at sbt.io.Using.apply(Using.scala:27)
[error] 	at sbt.io.IO$.withZipOutput(IO.scala:687)
[error] 	at sbt.io.IO$.archive(IO.scala:613)
[...]

PR #4593, opened 2019-03-31
Fix depth condition on GlobLister.aggregate

eatkins:
I wrote this check in a rush and realized that it didn't quite match the
correct glob semantics. The depth parameter is effectively the index of
the array of sorted child directories of the base. That index is
computed with getNameCount - 1, not getNameCount. It is also inclusive,
not exclusive, hence the switch from `<` to `<=`.

This change was motivated by my reviewing the initial change in the
context of the fix to https://github.com/sbt/sbt/issues/4591.


eed3si9n (approved):

eed3si9n:
Once this is merged, it would be good to run Dbuild to validate #4591.

eed3si9n:
https://jenkins.scala-sbt.org:8592/job/sbt-validator/1496/

PR #4594, opened 2019-04-01
Change default FileTree implementation

eatkins:
I have noticed on linux that the file cache updates aren't fast enough
for ExternalHooks. Say you have project b that depends on project a.
With a clean build, if you run b/compile, the file cache may not yet see
the changes to *.class files generated by project a. There are multiple
ways to fix this:
* don't use the file cache for binary products
* use the analysis results to invalidate the cache
* switch over to my hypothetical replacement file system

In the meantime, we should stop spamming users by default.


eed3si9n (dismissed):

eatkins:
Next time I'll remember to rebase before pushing :)

eed3si9n (approved):

Issue #4592, opened 2019-03-31
java.lang.IllegalAccessError: class X cannot access its superinterface Y


eed3si9n:
## steps

- git clone librarymanagement
- Run sbt using a recent nightly
- test

## problem

```
[info] sbt.internal.librarymanagement.ManagedChecksumsSpec *** ABORTED ***
[info]   java.lang.IllegalAccessError: class org.apache.ivy.plugins.parser.m2.ReplaceMavenConfigurationMappings$$anon$1 cannot access its superinterface org.apache.ivy.plugins.parser.m2.PomModuleDescriptorBuilder$ConfMapper
[info]   at java.lang.ClassLoader.defineClass1(Native Method)
[info]   at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
[info]   at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
[info]   at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
[info]   at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
[info]   at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
[info]   at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
[info]   at java.security.AccessController.doPrivileged(Native Method)
[info]   at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
[info]   at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
[info]   ...
[error] java.lang.IllegalAccessError: class org.apache.ivy.plugins.parser.m2.ReplaceMavenConfigurationMappings$$anon$1 cannot access its superinterface org.apache.ivy.plugins.parser.m2.PomModuleDescriptorBuilder$ConfMapper
[error] 	at java.lang.ClassLoader.defineClass1(Native Method)
[error] 	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
[error] 	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
[error] 	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
[error] 	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
[error] 	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
[error] 	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
[error] 	at java.security.AccessController.doPrivileged(Native Method)
[...]

Issue #4575, opened 2019-03-24
sbt-1.3.0-M2 breaks code that relies on URLClassLoader


xerial:
- [x] I've read the [CONTRIBUTING](https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md) guidelines

## steps

- Checkout airfame https://github.com/wvlet/airframe
- Run ` sbt -Dsbt.version=1.3.0-M2 logJVM/testOnly *ResourceTest`

## problem
- Some application needs to find files inside a jar file. To do so, we need to use the public method `URLClassLoader.findResourceURLs(path) : Enumeration[URL]`. ClassLoader has the same name method, but it's defined as protected, so we cannot use the base ClassLoader.
- Before sbt-1.3.0-M2, if we trace parent class loaders, we usually can find at least one URLClassLoader that can access resources inside jar files. 
    - Actually LayeredClassLoader, which is extending URLClassLoader, is found as a parent class loader, but it seems this class loader cannot find resources inside jar files.

- Upgrading airframe to use sbt-1.3.0-M2 breaks the build https://github.com/wvlet/airframe/pull/457
   - This is because no URLClassLoader that can find resources inside jar files can be found in this code: https://github.com/wvlet/airframe/blob/28a70c3fdc7a209eabf357548bdf7737c80e7b52/airframe-log/jvm/src/main/scala/wvlet/log/io/Resource.scala#L81-L90

## expectation
- sbt class loader should provide a URLClassLoader to get local resource URls.

## notes

sbt version: 1.3.0-M2

xerial:
Actually LayeredClassLoader, which is extending URLClassLoader, is found in a parent, but it seems this class loader cannot find resources inside jar files.

eatkins:
Thanks for the report @xerial. I will add this to my queue. There weren't many tests that loaded resources so that is something of a blindspot that needs to be addressed.

eatkins:
@xerial I do plan to fix the regression but I am curious if there is another way to solve this problem in airframe. Depending on the implementation details of class loading is inherently unreliable.

xerial:
@eatkins Actually simply loading class files of scala.io package is failing in sbt-1.3.0-M2. I think it should be supported by any class loaders in Scala because Scala standard libraries is always there for Scala applications.

I think this is sbt-specific issue since sbt itself is also using Scala.

xerial:
[...]

PR #4601, opened 2019-04-02
Improve layers

eatkins:

Among other things, this resolves:
* https://github.com/sbt/sbt/issues/4592 by adding a better error message (it is not possible to use a layered classloader in library management because of the use of an open package and extending a package private interface)
* https://github.com/sbt/sbt/issues/4575 by moving the sbt test-interface jar underneath the rest of the sbt metabuild classpath

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1781,7 +1781,15 @@ object Defaults extends BuildCommon {
           // by the test task layers may be cached without evicting the runtime classloader layers. The
           // cache size should be a multiple of two to support snapshot layers.
           ClassLoaderCache(4)
-        )
+        ),
+        bgCopyClasspath in bgRun := {
+          val old = (bgCopyClasspath in bgRun).value
+          old && (Test / classLoaderLayeringStrategy).value != ClassLoaderLayeringStrategy.ShareRuntimeDependenciesLayerWithTestDependencies
+        },
+        bgCopyClasspath in bgRunMain := {
+          val old = (bgCopyClasspath in bgRunMain).value
+          old && (Test / classLoaderLayeringStrategy).value != ClassLoaderLayeringStrategy.ShareRuntimeDependenciesLayerWithTestDependencies
+        },
       )
 
   lazy val testSettings: Seq[Setting[_]] = configSettings ++ testTasks ++
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1772,7 +1772,7 @@ object Defaults extends BuildCommon {
       Classpaths.addUnmanagedLibrary ++
       Vector(
         TaskRepository.proxy(
-          classLoaderCache,
+          Compile / classLoaderCache,
           // We need a cache of size four so that the subset of the runtime dependencies that are used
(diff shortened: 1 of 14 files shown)


main/src/main/scala/sbt/internal/ClassLoaders.scala
```diff
@@ -201,49 +185,42 @@ private[sbt] object ClassLoaders {
...
+        throw new IllegalStateException(s"sbt was launched with a non URLClassLoader: $cl")
+    }
+  }
+  def apply(libraryLoader: ClassLoader, fullLoader: ClassLoader): ClassLoader = {
+    val interfaceFilter: URL => Boolean = _.getFile.endsWith("test-interface-1.0.jar")
+    val (interfaceURL, rest) = fullLoader.urls.partition(interfaceFilter)
+    val interfaceLoader = new URLClassLoader(interfaceURL, libraryLoader.getParent) {
+      override def toString: String = s"SbtTestInterfaceClassLoader(${getURLs.head})"
+    }
+    val updatedLibraryLoader = new URLClassLoader(libraryLoader.urls, interfaceLoader) {
```
xerial, line 219: 👍

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1781,7 +1781,11 @@ object Defaults extends BuildCommon {
           // by the test task layers may be cached without evicting the runtime classloader layers. The
           // cache size should be a multiple of two to support snapshot layers.
           ClassLoaderCache(4)
-        )
+        ),
+        bgCopyClasspath in bgRun :=
+          (Test / classLoaderLayeringStrategy).value != ClassLoaderLayeringStrategy.ShareRuntimeDependenciesLayerWithTestDependencies,
```
eed3si9n: We might want to change this to

```scala
bgCopyClasspath in bgRun := {
  val old = (bgCopyClasspath in bgRun ).value
  old && (Test / classLoaderLayeringStrategy).value != ClassLoaderLayeringStrategy.ShareRuntimeDependenciesLayerWithTestDependencies  
}
```

This is because `bgCopyClasspath` is currently provided at `globalJvmCore`, which means that the build user can opt-out of it as:

```scala
ThisBuild / bgCopyClasspath := false
```

Without `old`, it is going to undo the wider scope.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1781,7 +1781,11 @@ object Defaults extends BuildCommon {
...
+        bgCopyClasspath in bgRunMain :=
+          (Test / classLoaderLayeringStrategy).value != ClassLoaderLayeringStrategy.ShareRuntimeDependenciesLayerWithTestDependencies,
```
eed3si9n: Same as above.

eed3si9n (changes requested):

main/src/main/scala/sbt/Defaults.scala
eatkins: Aah, good call!

eatkins: Fixed in https://github.com/sbt/sbt/pull/4601/commits/399dd920b0578609936b0d15562672d4dbfd57f3

eed3si9n (approved):

PR #4459, opened 2018-11-21
Have updateClassifiers use the dependencyResolution task

alexarchambault:
- [x] I've read the [CONTRIBUTING](https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md) guidelines

PR #4614, opened 2019-04-17
Coursier dependency resolution  integration

### LINKED ISSUE #2997: Add coursier to sbt under a flag
The default ivy resolver in sbt is slow. Anyone that has started a sbt session in CI knows. The attempts that have been made to make it faster (for example, cached resolution in 0.13.7) still fall short. [Read this discussion for more details](https://github.com/sbt/sbt/issues/413).
  
Coursier is an sbt plugin that speeds up resolution times. [In one informal benchmark](https://github.com/sbt/sbt/pull/2992), Coursier resolves spark-sql in 13 seconds while the default resolver does it in 51 seconds.
  
I suggest that we satisfy the need of fast resolution times. I propose we merge a coursier-based artifact resolution. As sbt maintainers care about stability, I propose this is done under a flag. But in the future 0.13.x version we can consider removing the default resolver if no further issues with Coursier are encountered.
  
Coursier is almost ready to hit a 1.0 release. I propose that's the one we use. Some can say that Coursier resolution is not as good as ivy. I'd like to see some examples of this if it's true, but I'd like to point out that it's the other way around too. [This issue](https://github.com/sbt/sbt/issues/2953) shows an example of an artifact that is not resolved by Ivy but it is by Coursier.
  
One issue with adding Coursier is its external dependencies. We've been discussing in https://github.com/sbt/sbt/pull/2992 that there exists the possibility to include Coursier by shading them. To the best of my knowledge, @alexarchambault has offered his help to pull this effort off.
  
I have personally settled to parallelize ivy resolution times in this PR 
 https://github.com/sbt/sbt/pull/2992. I'd like this one to be merged just for those that are forced to use ivy resolution. However, I consider coursier to be a much better alternative and would like to see it as the default one, whether now or in the future.
  
A benefit of merging this into sbt over using it as a plugin is that it allows us to resolve sbt modules faster, which is important when installing sbt for the first time. If we're to recommend sbt as the default way of getting Scala, this operation should be fast.
  
Sbt users frequently complain about resolutions. Coursier is the best solution I know to this problem. 
  
I'd like to see whether sbt maintainers are willing to make this step. I'm happy to work alongside @alexarchambault to make this a reality.
  
Related discussion: https://github.com/sbt/sbt/issues/2294.

eed3si9n:
Fixes #2997

This adopts Coursier-backed LM API implementation (lm-coursier-shaded).

As it stands, this uses Coursier by default to resolve both the metabuild and the proper build.

`ThisBuild / useCoursier := false` or `-Dsbt.coursier=false` can be used to opt-out of using Coursier for the dependency resolution.

### credits

The Coursier-backed LM API implementation was implemented by:
- @leonardehrenfried in https://github.com/sbt/librarymanagement/pull/190,
- @andreaTP in https://github.com/sbt/librarymanagement/pull/270, and
- @alexarchambault in https://github.com/coursier/sbt-coursier/pull/7 etc.


project/Dependencies.scala
```diff
@@ -115,6 +100,9 @@ object Dependencies {
   def addSbtZincCompile(p: Project): Project =
     addSbtModule(p, sbtZincPath, "zincCompile", zincCompile)
 
+  val lmCousierVersion = "1.1.0-M14"
```
eed3si9n: Note that I am adding dependency to lm-coursier 1.1.0-M14.

eed3si9n:
Here are the failing tests. Some of these are testing Ivy specific settings, we can fix them by flipping back to Ivy.

- [x] 	actions/all
- [x] 	actions/external-doc
- [x] 	dependency-management/compiler-bridge-binary
- [x] 	dependency-management/configurations-to-retrieve
- [x] 	dependency-management/artifact
- [x] 	dependency-management/cache-classifiers
- [x] 	dependency-management/cache-local
- [x] 	dependency-management/cache-resolver
- [x] 	dependency-management/cached-resolution-circular
- [x] 	dependency-management/cached-resolution-classifier
- [x] 	dependency-management/cached-resolution-conflicts
- [x] 	dependency-management/cached-resolution-exclude
- [x] 	dependency-management/circular-dependency
- [x] 	dependency-management/conflict-manager
- [x] 	dependency-management/default-artifact
- [x] 	dependency-management/dynamic-revision
- [x] 	dependency-management/ext-pom-classifier
- [x] 	dependency-management/force-update-period
- [x] 	dependency-management/deliver-artifacts
- [x] 	dependency-management/extra
- [x] 	dependency-management/info
- [x] 	dependency-management/ivy-settings-a
- [x] 	dependency-management/ivy-settings-b
- [x] 	dependency-management/ivy-settings-multi-a
- [x] 	dependency-management/pom-classpaths
- [x] 	dependency-management/ivy-settings-c
- [x] 	dependency-management/latest-local-plugin
- [x] 	dependency-management/metadata-only-resolver
- [x] 	dependency-management/retrieve-managed-sync
- [x] 	dependency-management/scala-organization
- [x] 	dependency-management/sources
- [x] 	dependency-management/pom-parent-pom
- [x] 	dependency-management/snapshot-resolution
- [x] 	dependency-management/test-artifact
- [x] 	dependency-management/update-sbt-classifiers
- [x] 	dependency-management/url
- [x] 	project/cross-plugins-defaults
- [x] 	project/default-auto-plugins
- [x] 	project/extra
- [x] 	project/update-classifiers
- [x] 	watch/dynamic-inputs
- [x] 	watch/task
- [x] 	watch/input-parser
- [x] 	watch/on-termination
- [x] 	watch/input-aggregation
- [x] 	watch/custom-config
- [x] 	classloader-cache/snapshot
- [x] repoOverrideTest:scripted dependency-management/default-resolvers

eed3si9n:
https://github.com/sbt/sbt/pull/4459 fixes `actions/all`

eed3si9n:
Reported  `actions/external-doc` failure as https://github.com/coursier/coursier/issues/1123

eed3si9n:
`dependency-management/configurations-to-retrieve` is Ivy feature.
`dependency-management/artifact` is border-line Ivy feature, meaning it could be called a bug.

`dependency-management/cache-local` is failing because it's not respecting the customized Ivy home?

```
[info] [error] (update) sbt.librarymanagement.ResolveException: Error downloading org.example:def_2.12:1.0
[info] [error]   Not found
[info] [error]   not found: /Users/eed3si9n/.ivy2/local/org.example/def_2.12/1.0/ivys/ivy.xml
[info] [error]   not found: https://repo1.maven.org/maven2/org/example/def_2.12/1.0/def_2.12-1.0.pom
```

Reported that as https://github.com/coursier/coursier/issues/1124.

`dependency-management/cached-resolution-classifier` is failing because it's missing `commons-io-1.4-sources` even though the subproject is depending on `"commons-io" % "commons-io" % "1.4" classifier "sources"`.

`dependency-management/cached-resolution-exclude` is Ivy feature.

`dependency-management/circular-dependency` is also Ivy feature, sort of. Ivy is able to throw an exception given a circular dependency.

```
[info] [error] org.example#b_2.10;1.0-SNAPSHOT->org.example#c_2.10;1.0-SNAPSHOT->...
[info] [error] org.apache.ivy.plugins.circular.CircularDependencyException: org.example#b_2.10;1.0-SNAPSHOT->org.example#c_2.10;1.0-SNAPSHOT->...
[info] [error] 	at org.apache.ivy.plugins.circular.ErrorCircularDependencyStrategy.handleCircularDependency(ErrorCircularDependencyStrategy.java:37)
```

Coursier just publishLocals without an error. Reported as https://github.com/coursier/coursier/issues/1125.

`dependency-management/default-artifact` looks like a bug.

```
[info] [error] coursier.error.FetchError$DownloadingArtifacts: Error fetching artifacts:
[info] [error] file:///private/var/folders/hg/2602nfrs2958vnshglyl3srw0000gn/T/sbt_1fe50cd/repo/a/b/1.0.0/b1.jar.jar: not found: /private/var/folders/hg/2602nfrs2958vnshglyl3srw0000gn/T/sbt_1fe50cd/repo/a/b/1.0.0/b1.jar.jar
[info] [error] file:///private/var/folders/hg/2602nfrs2958vnshglyl3srw0000gn/T/sbt_1fe50cd/repo/a/b/1.0.0/b2.jar.jar: not found: /private/var/folders/hg/2602nfrs2958vnshglyl3srw0000gn/T/sbt_1fe50cd/repo/a/b/1.0.0/b2.jar.jar
```

Note `b2.jar.jar`.

`dependency-management/dynamic-revision` is a weird one. POM says:

```xml
        <dependency>
            <groupId>org.scala-stm</groupId>
            <artifactId>scala-stm_2.10.0</artifactId>
            <version>0.6</version>
        </dependency>
```

in http://dl.bintray.com/typesafe/maven-releases/play/play_2.10/2.1.2/play_2.10-2.1.2.pom, but Coursier errors out as:

```
[info] [error] (update) sbt.librarymanagement.ResolveException: Error downloading org.scala-stm:scala-stm_2.10.6:0.6
[info] [error]   Not found
[info] [error]   not found: /Users/eed3si9n/.ivy2/local/org.scala-stm/scala-stm_2.10.6/0.6/ivys/ivy.xml
[info] [error]   not found: https://repo1.maven.org/maven2/org/scala-stm/scala-stm_2.10.6/0.6/scala-stm_2.10.6-0.6.pom
[info] [error]   not found: https://repo.typesafe.com/typesafe/releases/org/scala-stm/scala-stm_2.10.6/0.6/scala-stm_2.10.6-0.6.pom
```

It seems like it's trying to overwrite the `_2.10.0` with `_2.10.6`.

`dependency-management/latest-local-plugin` is about `latest.integration`. https://github.com/coursier/coursier/issues/209

`dependency-management/metadata-only-resolver` just runs, so I am going to disable the test.

`dependency-management/retrieve-managed-sync` is Ivy feature.

`dependency-management/scala-organization` is on sbt part of Ivy integration. For now this would require `ThisBuild / useCoursier := false`.

`dependency-management/snapshot-resolution` failure results from 24h TTL.

`dependency-management/test-artifact` failure, I'm not sure.

`dependency-management/url` seems to be some download bug. Reported https://github.com/coursier/coursier/issues/1127.

eed3si9n:
`project/cross-plugins-defaults` couldn't find sbt 0.13 due to missing Typesafe Ivy Releases repo. Adding `resolvers += Resolver.typesafeIvyRepo("releases")` lets the test pass.

`watch/*` failures is related to Coursier not excluding sbt artifacts. Reported as https://github.com/coursier/coursier/issues/1128.

`repoOverrideTest:scripted dependency-management/default-resolvers` failure seems legitimate.

```
[info] [error] (update) sbt.librarymanagement.ResolveException: Error downloading org.apache.logging.log4j:log4j-slf4j-impl:2.11.2
[info] [error]   Not found
[info] [error]   not found: /Users/eed3si9n/.ivy2/local/org.apache.logging.log4j/log4j-slf4j-impl/2.11.2/ivys/ivy.xml
[info] [error]   not found: https://oss.sonatype.org/content/repositories/public/org/apache/logging/log4j/log4j-slf4j-impl/2.11.2/log4j-slf4j-impl-2.11.2.pom
[info] [error] Error downloading org.apache.logging.log4j:log4j-api:2.11.2
[info] [error]   Not found
[info] [error]   not found: /Users/eed3si9n/.ivy2/local/org.apache.logging.log4j/log4j-api/2.11.2/ivys/ivy.xml
[info] [error]   not found: https://oss.sonatype.org/content/repositories/public/org/apache/logging/log4j/log4j-api/2.11.2/log4j-api-2.11.2.pom
[info] [error] Error downloading org.sonatype.oss:oss-parent:7
[info] [error]   Not found
[info] [error]   not found: /Users/eed3si9n/.ivy2/local/org.sonatype.oss/oss-parent/7/ivys/ivy.xml
[info] [error]   not found: https://oss.sonatype.org/content/repositories/public/org/sonatype/oss/oss-parent/7/oss-parent-7.pom
[info] [error] Error downloading com.google.protobuf:protobuf-java:3.3.1
[info] [error]   Not found
[info] [error]   not found: /Users/eed3si9n/.ivy2/local/com.google.protobuf/protobuf-java/3.3.1/ivys/ivy.xml
[info] [error]   not found: https://oss.sonatype.org/content/repositories/public/com/google/protobuf/protobuf-java/3.3.1/protobuf-java-3.3.1.pom
[info] [error] Error downloading org.sonatype.oss:oss-parent:6
[info] [error]   Not found
[info] [error]   not found: /Users/eed3si9n/.ivy2/local/org.sonatype.oss/oss-parent/6/ivys/ivy.xml
[info] [error]   not found: https://oss.sonatype.org/content/repositories/public/org/sonatype/oss/oss-parent/6/oss-parent-6.pom
[info] [error] Error downloading org.apache.logging.log4j:log4j-core:2.11.2
[info] [error]   Not found
[info] [error]   not found: /Users/eed3si9n/.ivy2/local/org.apache.logging.log4j/log4j-core/2.11.2/ivys/ivy.xml
[info] [error]   not found: https://oss.sonatype.org/content/repositories/public/org/apache/logging/log4j/log4j-core/2.11.2/log4j-core-2.11.2.pom
```

This is failing on the resolution of metabuild via Sonatype OSS.

leonardehrenfried:
While I don't think I'm up to date anymore with the technical details to add anything substantial, I still would like to say that I'm very happy that this feature is going ahead. 

Thank you, Eugene!

alexarchambault:
I'll be able to discuss more on Tuesday, but I just don't think it's the right approach.

There's fair amount of logic needed to correctly wire sbt and coursier together, even when relying on the `dependencyResolution` key. It's mostly done in the [`sbt-coursier-shared` module](https://github.com/coursier/sbt-coursier/tree/b417d4f79efa1656f7954d6a2288eb3991e5c5b4/modules/sbt-coursier-shared/src/main/scala/coursier/sbtcoursiershared). The [`sbt-lm-coursier`](https://github.com/coursier/sbt-coursier/tree/b417d4f79efa1656f7954d6a2288eb3991e5c5b4/modules/sbt-lm-coursier/src/main/scala) plugin, that relies on the `dependencyResolution` key, uses it in particular. Most of the changes done in it have [non regression tests](https://github.com/coursier/sbt-coursier/tree/b417d4f79efa1656f7954d6a2288eb3991e5c5b4/modules/sbt-coursier/src/sbt-test), that I don't think this PR would pass.

Overall, this would be quite a regression for sbt-coursier users (to the point that I would probably loudly disassociate myself from this initiative, if this were to be merged as is…).

eed3si9n:
@alexarchambault https://github.com/sbt/sbt/blob/d7b30a55f9cfcb5746f8784684dc718e2038b6b7/main/src/main/scala/sbt/internal/LMCoursier.scala is basically sbt-coursier-shared module.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2292,52 +2349,70 @@ object Classpaths {
...
+                updateConfig,
+                // scalaModule,
+                srcTypes.toVector,
+                docTypes.toVector
+              ),
+              uwConfig,
+              Vector.empty,
+              s.log
+            ) match {
+              case Left(_)   => ???
```
andreaTP, line 2403: just re-throw the exception?

andreaTP: just noticed that it was already like this ...

main/src/main/scala/sbt/internal/LibraryManagement.scala
```diff
@@ -19,6 +22,37 @@ private[sbt] object LibraryManagement {
...
+      .flatMap(
+        str =>
+          ConsoleAppender.parseLogOption(str) match {
+            case LogOption.Always => Some(true)
+            case LogOption.Never  => Some(false)
+            case _                => None
+          }
+      )
+    val notIvyOpt = ivyOpt map { !_ }
+    coursierOpt.orElse(notIvyOpt).getOrElse(true)
```
andreaTP, line 47: this default to true, are we confident enough to have it as the default choice?

main/src/main/scala/sbt/internal/LibraryManagement.scala
```diff
@@ -19,6 +22,37 @@ private[sbt] object LibraryManagement {
...
+  }
+
+  def dependencyResolutionTask: Def.Initialize[Task[DependencyResolution]] = Def.taskDyn {
```
andreaTP, line 50: Does this means that changing the from Ivy to Coursier in the shell does work already?

project/Dependencies.scala
```diff
@@ -29,22 +29,7 @@ object Dependencies {
   private val utilScripted = "org.scala-sbt" %% "util-scripted" % utilVersion
 
   private val libraryManagementCore = "org.scala-sbt" %% "librarymanagement-core" % lmVersion
-
```
andreaTP, line 32: why this configuration values got removed?
if this is the case remember that the integration tests in lm will stop to work(e.g. they should be removed as well).

andreaTP (commented):
@eed3si9n again, thanks a lot for taking over! I left a few minor comments but overall great work, can't wait to see it merged!

andreaTP:
@alexarchambault I understand your point on regression tests, but you have also to consider the gain by having coursier as a first class citizen, such as the community build that ensures some good level of confidence.
sbt-coursier will stay as an alternative if any bug got spotted in this PR (and following development).
If you are really afraid on the quality level it will be a nice contribution to add the regression tests themselves on top of this PR :+1:

project/Dependencies.scala
eed3si9n: I figured that if we adopted lm-coursier from coursier repo, we can remove one from sbt/librarymanagement.

main/src/main/scala/sbt/internal/LibraryManagement.scala
eed3si9n: I think it's ready to at least try for 1.3.0-RC-1, and run it against Scala Community Build.

main/src/main/scala/sbt/internal/LibraryManagement.scala
eed3si9n: It should.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: I'd be happy to revisit this one later, but for this PR I'll keep it as is.

leonardehrenfried:
By the way, is the sbt meta build (plugins) also resolved using coursier or is that still using ivy?

eed3si9n:
@leonardehrenfried This PR enables Coursier LM for both metabuild and proper build by default atm. Due to the difference in the way exclusions are interpreted, it might need further tweaking.

alexarchambault:
To detail a bit my position on this approach:
- this doesn't allow to fetch sbt itself with coursier, so the [coursier-based sbt launcher](https://github.com/coursier/sbt-launcher) is still be needed,
- this doesn't bring much compared to what the coursier-based launcher already does. The launcher just puts the sbt-lm-coursier plugin along sbt's classpath, which has the same effect as this PR - using coursier in the meta and proper builds. (I don't know much about the internals of the community build… is it easier to test in the community build this way? couldn't the coursier-based sbt launcher be used from there?)
- this is much less flexible than keeping coursier in a separate plugin. This ties specific versions of sbt with particular versions of coursier. So a hot fix in coursier will now have to wait for a sbt release to be used from sbt, whereas a separate plugin only needs a new release of the plugin (with scala steward automatically bumping the coursier version in the sbt-coursier project, and releases handled from the CI with [sbt-ci-release](https://github.com/coursier/sbt-coursier/blob/b417d4f79efa1656f7954d6a2288eb3991e5c5b4/project/plugins.sbt#L3), these should happen almost overnight now). Also, I'm not really looking forward testing the coursier master from sbt by building sbt from sources… Easier to just build and locally publish a plugin.
- for users already setting some coursier-specific keys in `~/.sbt/1.0/*.sbt`, this will probably require a compatibility plugin, to create keys with the same name and under the same namespace. (As the new keys have different names that the original ones in coursier, but also live in an object under the `sbt` namespace now)
- the current coursier sbt plugins will probably still be maintained a bit, for the former sbt versions at least. That makes two virtually identical code bases to maintain at the same time, which is just a loss of time.

eed3si9n:
@alexarchambault Thanks for the comment.
Here are my interests:
- overall Scala community weaning off of Ivy, and migrate to Coursier, regardless of `sbt` (script + launcher) implementation
- repeatability of the build (what author of the library get is the same as someone else who `git clone` gets)
- speedup of Scala Community Build

First I want to thank you for all the effort around Coursier work, and also want to clarify that I don't intend to prevent or compete with csbt (Coursier-based sbt launcher) or sbt-coursier to provide richer experience to many users. Coursier has improved over time, so I think it's time to discuss  drawback of having 2 major dependency resolutions. I want to tip the scale so Coursier becomes the outright majority.

To achieve that in a safe manner, I believe shipping Coursier as the out-of-box LM implementation of sbt 1.3.0 is the way to go. This will not require reinstallation of `sbt` (Bash script + launcher), the library authors just write `sbt.version=1.3.0` in `build.properties`. If for some reason the graph is not satisfactory, they can opt-out as `useCoursier := false`.

> this doesn't bring much compared to what the coursier-based launcher already does.

As noted above, what this PR aims to bring is wider distribution of Coursier and improved repeatability. If I am a library author of scopt etc, and if someone else using csbt got different dependency graph when they git cloned, that's a bug as far as sbt is concerned.

> this is much less flexible than keeping coursier in a separate plugin.

I think we should continue to allow Coursier to be bumped using the plugin. In https://github.com/sbt/sbt/pull/4614/commits/7091f0e272237a73a0eb2d1a8c80b3ca1c9316cd I had to flip the classpath ordering from `mjars ++ sbtCp` (plugin wins) to `sbtCp ++ mjars` (sbt classpath wins) due to the way exclusion works, but hopefully we can found our way.

> for users already setting some coursier-specific keys in `~/.sbt/1.0/*.sbt`, this will probably require a compatibility plugin

Yes, I tried to avoid key collision with sbt-coursier. I am open to suggestion for how we handle this.

> the current coursier sbt plugins will probably still be maintained a bit, for the former sbt versions at least. That makes two virtually identical code bases to maintain at the same time, which is just a loss of time.

I agree it's not ideal, but I think the payoff is worth it.
If you have any more concerns, I would be happy to schedule a call with you to discuss more.

alexarchambault:
@eed3si9n Thanks for your answer (and for pushing this effort further, overall).

If it's still possible to use sbt-coursier, that's a big showstopper less for me. I don't know a lot about the way classloaders are handled for plugins… Will it still be fine to load newer coursier versions from sbt-coursier (newer versions possibly breaking binary compatibility with the one sbt was built with)?

About the details here, wouldn't it be possible to keep all things related to coursier in a separate auto plugin, itself in a separate module say, rather than putting everything in `Defaults.scala`? The [`sbt` module](https://github.com/sbt/sbt/blob/4074cb32d3b1e12fddcbe0f5ce3d51d515124c19/build.sbt#L624) could depend on it, but not [`main`](https://github.com/sbt/sbt/blob/4074cb32d3b1e12fddcbe0f5ce3d51d515124c19/build.sbt#L580) say. sbt would depend on it by default, so that would make no difference at first, but that would allow to just exclude it from csbt for example.

alexarchambault:
Also, ideally, sbt plugins could depend on `main` rather than `sbt`, so that they don't unnecessarily depend on coursier itself (and don't rely on changing things in it).

alexarchambault:
To temper the issue with having sbt depend on coursier (but also on shapeless, argonaut, argonaut-shapeless, that the coursier module, with its high level API, now depends on), I'm moderately confident I can have lm-coursier not depend on coursier itself, via some shading. I'm working on it right now here: https://github.com/coursier/sbt-coursier/pull/58.

alexarchambault:
There would be two lm-coursier modules: one shading coursier (`lm-coursier-shaded`), and one with no shading. sbt-coursier would still depend on the latter, to do more advanced stuff with more refined coursier APIs.

If the plugin classpath really has a higher priority than the one of sbt, then the latter module (brought by sbt-coursier) would take over the one with shading, if ever sbt-coursier is added to a build. (_Edit_: it seems I can have the sbt-lm-coursier plugin rely only on lm-coursier, so it ought to work fine with both the shaded and non-shaded versions of lm-coursier. So we wouldn't even need this classpath take over.)

I could also picture csbt substituting the module with shading for the one with no shading from the sbt classpath, if needed.

eed3si9n:
> I don't know a lot about the way classloaders are handled for plugins… Will it still be fine to load newer coursier versions from sbt-coursier (newer versions possibly breaking binary compatibility with the one sbt was built with)?

I think we should support this. The bincompat guarantee that sbt promises should be LM API, not Coursier itself. Current blocker is https://github.com/coursier/coursier/issues/1128.

### technical details about exclusion

In sbt 1.3.x, I am trying to speed up the resolution of the metabuild by excluding `"org.scala-sbt" %  "sbt" % *`, `"org.scala-sbt" %  "zinc_2.12" % *` etc (#4443, #4589), and reuse the classpath resolved already by the launcher. This is what I'm calling `sbtCp`. When I switched metabuild resolution to Coursier, I noticed that `excludeDependencies` doesn't work the same way, and it started including submodules of sbt, so as a workaround I had to prioritize `sbtCp` over the plugin classpath. This would put Coursier 1.1.0-M14 ahead of the newer Coursier resolved by sbt-coursier. One possibility is to exclude `"org.scala-sbt" %  *`.

### source sharing

> About the details here, wouldn't it be possible to keep all things related to coursier in a separate auto plugin, itself in a separate module say, rather than putting everything in `Defaults.scala`?

This one is tricky. Having all this in Default allows Coursier to go into metabuild (or meta^2-build) because it doesn't rely on plugin injection. To share the code, but still put things into Defaults, I'd need split mainProj in half or something.

alexarchambault:
I’ll have a look at https://github.com/coursier/coursier/issues/1128 first thing tomorrow I think. (Then try to publish lm-coursier with shading.)

eatkins:
I could be mistaken but, off the cuff, I believe that the plugin classloader more or less is a URLClassLoader using the plugin classpath urls and the sbt metabuild classloader as parent. This causes the metabuild classpath to effectively leak into the plugin. One naive fix is to do something like:
```scala
class PluginClassLoader(urls: Array[URL], parent: ClassLoader) extends URLClassLoader(urls, parent) {
  override def loadClass(name: String, resolve: Boolean): Class[_] = {
    val forceParent = name.startsWith("java.") || name.startsWith("scala.") || name.startsWith("sun.") or name.startsWith("sbt.")
    if (!forceParent) {
      try {
        val clazz = findClass(name)
        if (resolve) resolveClass(clazz)
        clazz
      } catch { case _: ClassNotFoundException => super.loadClass(name, resolve) }
    } else {
      super.loadClass(name, resolve)
    }
  }
}
```
The problem with this approach is that sbt does expose some dependencies publicly. If a plugin uses the same dependencies (doesn't matter if it's a different version or not), then it will crash if tries to interop with those sbt apis that use the same dependency. This is because the plugin provided instances of the dependency classes will have been loaded with the plugin classloader but the sbt api requires an instance loaded by the sbt metabuild classloader.

Ultimately the dependencies appearing in the public api of sbt is impossible to fix for plugins in general. I do, however, think that there are some solutions that would work for most common use cases. For example, a plugin that uses a dependency internally but doesn't expose that dependency in its own api should be fine using the approach described above.

eatkins:
Looking at the sbt PR, the keys that expose coursier classes are private[sbt]. This means we could do the trick described in my comment above to ensure that the plugin classloader does not try to load classes in the `coursier` package using the metabuild classloader.

alexarchambault:
So I managed to have lm-coursier not depend on coursier anymore. It's published as `io.get-coursier::lm-coursier-shaded`. I made the sbt-lm-coursier plugin rely on it, it seems to work fine (passes all the scripted tests in particular), so I'm rather confident it can be used as a substitute to `io.get-coursier::lm-coursier` here. `1.1.0-M14+12-8bc1d583-SNAPSHOT` is available on sonatype snapshots, and `1.1.0-M14-1` should follow suit on Central.

Note that I renamed `coursier.lmcoursier` as just `lmcoursier` (to facilitate the shading of `coursier.*`), and that lm-coursier-shaded re-defines some stuff from coursier itself, under `lmcoursier.definitions` (that should be used instead of stuff from coursier, that aren't around anymore).

Once this PR uses that, I don't see major blockers AFAIC. Using csbt, I should be able to bump the lm-coursier-shaded version if needed. And one should be able to add sbt-coursier to a build, there won't be a conflicting coursier version around anymore.

If lm-coursier-shaded works fine from here, I'll probably make some extra adjustments in it, to make it easier to maintain binary compatibility in the future.

The only (minor) remaining issue to me would be code duplication between this PR, and the sbt-coursier-shared and sbt-lm-coursier modules. But it's probably worth the benefits…

alexarchambault:
@eed3si9n About https://github.com/coursier/coursier/issues/1128, replacing the occurrences of `excludeDependencies` in `LMCoursier.scala` by `allExcludeDependencies` should fix it.

eed3si9n:
Thanks for the update! I'll work on them then.

eed3si9n:
Looks like the tests passed using lm-coursier-shaded.

`allExcludeDependencies` didn't change the graph, but I think it's because I have sbt as a direct dependency. I'll remove that and see if that would work around the situation.

fernandoracca:
Excellent news. I'll try this latest setup on my builds. Thanks for the great work!

main/src/main/scala/sbt/internal/librarymanagement/IvyXml.scala
```diff
@@ -0,0 +1,225 @@
...
+      >
+        {licenseElems}
+        <description>{project.info.description}</description>
+      </info>
+    } % infoAttrs
+
+    val confElems = project.configurations.toVector.collect {
+      case (name, extends0) if !shadedConfigOpt.exists(_.value == name.value) =>
+        val extends1 = shadedConfigOpt.fold(extends0)(c => extends0.filter(_.value != c.value))
+        val n = <conf name={name.value} visibility="public" description="" />
```
mkurz, line 123: Hardcoding `visibility` to `public` is not such a good idea, see #5455

PR #4616, opened 2019-04-18
Merge 1.2.x into develop

dwijnand:
This brings the git tags into the parents of develop's HEAD.  Previously the latest 1.2.x tag for develop was v1.2.1.


eed3si9n:
What kind of git sorcery did you use to resolve the conflicts?

eed3si9n (approved):

dwijnand:
I did it manually, comparing each conflict.  Using https://sjl.bitbucket.io/splice.vim/ I've found it really manageable (and fear-removing).

There were about 5 files that conflicted and in the result the diff was only the change in #4473.

PR #4623, opened 2019-04-20
sbt-scalafmt 2.0.0

eed3si9n:
This is incremental.
Too bad that this comes with more format change.


dwijnand:
All the formatting changes look good, except for the try/finally ones.

I think you can work around them by dropping the block.

dwijnand (dismissed):

eed3si9n:
I think AppVeyor failure is a bug - https://github.com/scalameta/scalafmt/issues/1404

PR #4640, opened 2019-04-30
Use global classloader cache for ScalaInstance

### LINKED ISSUE #4639: ZincComponentCompilerSpec causes metaspace leak

## steps

Run test

## problem

java.util.concurrent.ExecutionException: java.lang.OutOfMemoryError: Metaspace    

## expectation

Tests run without running out of memory

## notes

I took a heap dump and the eclipse memory manager tool warned:
The thread java.lang.Thread @ 0x74e4452b8 pool-1294-thread-11-ScalaTest-running-ZincComponentCompilerSpec keeps local variables with total size 51,637,592 (15.15%) bytes.

sbt version: 1.3.0-SNAPSHOT

eatkins:
I noticed in a heap dump of sbt that there were many classloaders for
the scala instance. I then realized that we were making a new
classloader for the scala library on every test run. Even worse, the
ScalaInstanceLoader instance was never closed which lead to a metaspace
leak. I moved the scala instance classloader to the global classloader
cache. Not only will these be correctly cached, they will be closed if
evicted from the cache.

Fixes #4639

### DIFF
--- a/main/src/main/scala/sbt/internal/ClassLoaders.scala
+++ b/main/src/main/scala/sbt/internal/ClassLoaders.scala
@@ -40,6 +40,7 @@ private[sbt] object ClassLoaders {
       rawRuntimeDependencies =
         dependencyJars(Runtime / dependencyClasspath).value.filterNot(exclude),
       allDependencies = dependencyJars(dependencyClasspath).value.filterNot(exclude),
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Nice

PR #4644, opened 2019-05-02
Properly close a number of classloaders

eatkins:
I discovered there were a number of places where closing a ClassLoader
didn't work correctly because I was assuming it was a URLClassLoader
when it was actually a ClasspathFilter. I also incorrectly imported the
wrong kind of URLClassLoader in Run.scala. Finally, I close the
SbtMetaBuildClassLoader when xMain exits now.

### DIFF
--- a/main/src/main/scala/sbt/internal/ClassLoaders.scala
+++ b/main/src/main/scala/sbt/internal/ClassLoaders.scala
@@ -226,6 +226,11 @@ private[sbt] object SbtMetaBuildClassLoader {
     }
     new URLClassLoader(rest, updatedLibraryLoader) {
       override def toString: String = s"SbtMetaBuildClassLoader"
+      override def close(): Unit = {
+        super.close()
+        updatedLibraryLoader.close()
+        interfaceLoader.close()
+      }
     }
   }
 }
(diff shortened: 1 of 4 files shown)


eed3si9n (approved):
Nice. Thanks!

main/src/main/scala/sbt/internal/ClassLoaders.scala
```diff
@@ -226,6 +226,11 @@ private[sbt] object SbtMetaBuildClassLoader {
     }
     new URLClassLoader(rest, updatedLibraryLoader) {
       override def toString: String = s"SbtMetaBuildClassLoader"
+      override def close(): Unit = {
+        super.close()
+        updatedLibraryLoader.close()
+        interfaceLoader.close()
+      }
```
exoego, line 233: I think `closable.close` may throw `IOException`.

```scala
try {
  super.close()
} finally {
  try {
    updatedLibraryLoader.close(),
  } finally {
     interfaceLoader.close()
  }
}
```

Something like the above is safer?

eed3si9n: @exoego I think the suggestion is good, but I am ok with merging this PR as is for now, and following up later. Feel free to send us a PR or open an issue.

PR #4645, opened 2019-05-02
Unnecessary sleep

eatkins:


main/src/main/scala/sbt/internal/DefaultBackgroundJobService.scala
```diff
@@ -209,7 +209,6 @@ private[sbt] abstract class AbstractBackgroundJobService extends BackgroundJobSe
     }
     val xs = (products.toVector map { syncTo(workingDirectory / "target") }) ++
       ((full diff products) map { syncTo(serviceTempDir / "target") })
-    Thread.sleep(100)
```
eed3si9n, line 212: Looks like Thread.sleep(100) has been around since https://github.com/sbt/sbt/pull/2936.
LGTM if it passes the CI.

eed3si9n (approved):

main/src/main/scala/sbt/internal/DefaultBackgroundJobService.scala
eatkins: Yeah. I figured if anyone would know the purpose of the sleep, it'd be you :)

PR #4666, opened 2019-05-11
silence coursier log when supershell is off

eed3si9n:
CI logs are filled up with Downloaded, so I am turning the Coursier log off.

Ref https://github.com/coursier/coursier/pull/872

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -242,7 +242,7 @@ object Defaults extends BuildCommon {
       forceUpdatePeriod :== None,
       // coursier settings
       csrExtraCredentials :== Nil,
-      csrLogger :== None,
+      csrLogger := LMCoursier.coursierLoggerTask.value,
       csrCachePath :== LMCoursier.defaultCacheLocation,
       csrMavenProfiles :== Set.empty,
     )
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2438,7 +2438,7 @@ object Classpaths {
     csrConfiguration := LMCoursier.coursierConfigurationTask(false, false).value,
     csrResolvers := CoursierRepositoriesTasks.coursierResolversTask.value,
     csrRecursiveResolvers := CoursierRepositoriesTasks.coursierRecursiveResolversTask.value,
-    csrSbtResolvers := LMCoursier.coursierSbtResolversTask.value,
+    csrSbtResolvers := CoursierRepositoriesTasks.coursierSbtResolversTask.value,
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/internal/librarymanagement/CoursierRepositoriesTasks.scala
```diff
@@ -120,6 +121,36 @@ private[sbt] object CoursierRepositoriesTasks {
       }
     }
 
+  private val pluginIvySnapshotsBase = Resolver.SbtRepositoryRoot.stripSuffix("/") + "/ivy-snapshots"
+
+  def coursierSbtResolversTask: Def.Initialize[sbt.Task[Seq[Resolver]]] = Def.task {
```
eatkins, line 126: Doesn't really matter but the 'coursier' in `coursierSbtResolversTask` seems redundant in the context of CoursierRepositoriesTask.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -243,7 +243,7 @@ object Defaults extends BuildCommon {
       // coursier settings
       csrExtraCredentials :== Nil,
       csrLogger := LMCoursier.coursierLoggerTask.value,
-      csrCachePath :== LMCoursier.defaultCacheLocation,
```
eatkins, line 246: I like s/Path/Directory here.

main/src/main/scala/sbt/internal/LMCoursier.scala
```diff
@@ -15,9 +15,14 @@ import sbt.librarymanagement._
 import Keys._
 import sbt.internal.librarymanagement.{ CoursierArtifactsTasks, CoursierInputsTasks }
 import sbt.util.Logger
+import sbt.io.syntax._
 
 private[sbt] object LMCoursier {
-  def defaultCacheLocation: File = CoursierDependencyResolution.defaultCacheLocation
+  def defaultCacheLocation: File =
+    sys.props.get("sbt.coursier.home") match {
```
eatkins, line 22: What is the expected behavior if the user passes in a symlink?

eatkins (approved):
LGTM overall.

main/src/main/scala/sbt/internal/librarymanagement/CoursierRepositoriesTasks.scala
eed3si9n: In the original sbt-coursier code, `coursierResolversTask` implements `coursierResolvers` key, and this code tries to be somewhat consistent with that style so the maintenance is easier.

main/src/main/scala/sbt/internal/LMCoursier.scala
eed3si9n: I'm guessing that it should use the full path of symlink without following?

PR #4672, opened 2019-05-13
disable "Reload" exception stack trace

xuwei-k:

I think we should disable stacktrace or change to `class`.

```scala
package example

object Foo extends Exception

object Main {
  def x() = Foo
  def y() = Foo

    x()
    y().printStackTrace // unexpected stack trace!

    // example.Foo$
    //   at example.Foo$.<clinit>(A.scala)
    //   at example.Main$.x(A.scala:6)
    //   at example.Main$.main(A.scala:10)
    //   at example.Main.main(A.scala)
  }
}
```


eed3si9n (approved):
Thanks!

PR #4674, opened 2019-05-13
Manage shutdown hooks

eatkins:
I discovered that some registered shutdown hooks would crash due to
67df72ab019e32935588e1a58e636e07f4ed65ce because they would try to load
classes from the closed classloader. To fix this, I add a internal
shutdown hooks mechanism that can be managed by sbt. Any unevaluated
shutdown hooks will be run when the sbt main method exits. This means
that they will be run when the user calls reboot. I think that is
reasonable.


eatkins:
The appveyor failures have been sporadically occurring due to failure to resolve the scalafmt configuration. I'm not really sure what to do about that.

eed3si9n:
For a change it might not be because of Windows - https://github.com/scalameta/scalafmt/issues/1399#issuecomment-491714351

eatkins:
Oh, it's definitely not. I've seen it in intellij too.

eed3si9n (dismissed):

PR #4679, opened 2019-05-15
Don't ever invalidate rt.jar

eatkins:
There are issues when using jdk > 8 where the rt.jar file can be
invalidated by ExternalHooks. This causes spurious rebuilds. I think
it's fair to assume that rt.jar never changes. If a dependency is named
rt.jar, then invalidation may not work correctly but I think that this
is the more important case to handle.

I verified that before this change, it was impossible to run
akka-actor/compile twice in a row using adopt jdk 11 and, after this
change, re-compilation worked as expected.

### DIFF
--- a/main/src/main/scala/sbt/internal/ExternalHooks.scala
+++ b/main/src/main/scala/sbt/internal/ExternalHooks.scala
@@ -84,6 +84,7 @@ private[sbt] object ExternalHooks {
               case _ =>
                 javaHome match {
                   case Some(h) if file.toPath.startsWith(h) => None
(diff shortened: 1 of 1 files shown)


eatkins:
Fixes https://github.com/sbt/zinc/issues/622 in sbt at least.

eed3si9n (approved):

raboof:
Confirmed with Akka and sbt 1.3.0-RC1 in https://github.com/akka/akka/pull/26935 - this is a huge improvement, thanks!

chbatey:
Great, thanks!

PR #4680, opened 2019-05-15
Add explicit dependency for checkBuildSources

eatkins:
Rather than adding some custom logic to EvaluateTask.runTask, it makes
more sense to explicitly add a dependency on checkBuildSources to every
(non-injected) task in the build. This both reduces the amount of custom
logic for the checkBuildSources task and also makes it show up in the
`inspect` command

I also reordered the injected settings so that they come before the
project settings. This should allow the user to override the injected
settings if they wanted. The injected settings do not have the
dependency on checkBuildSources added which allows the other file stamp
tasks to run concurrently with the build source check. This means that
for tasks like compile there is often effectively zero overhead for the
build source check.

Finally, I simplified Continuous so that there is no custom logic for
tracking the meta build. By adding the dependency on checkBuildSources
to every task, the tasks end up automatically watching the build files
anyway.

### DIFF
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -16,10 +16,9 @@ import sbt.Project.richInitializeTask
 import sbt.Scope.Global
 import sbt.internal.Aggregation.KeyValue
 import sbt.internal.TaskName._
-import sbt.internal._
 import sbt.internal.util._
+import sbt.internal._
 import sbt.librarymanagement.{ Resolver, UpdateReport }
-import sbt.nio.Keys.IgnoreSourceChanges
 import sbt.std.Transform.DummyTaskMap
 import sbt.util.{ Logger, Show }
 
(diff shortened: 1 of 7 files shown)


eed3si9n:
I've been thinking a bit more about this, but I am not fully convinced that metabuild source checking should really be part of the task graph. It feels to me it's a concern one layer above that, namely part of command execution loop. For example, if I change `version` setting in `build.sbt`, shouldn't I expect:

```
sbt> version
```

to pick up the change?

Adding this as a custom command probably would be more technically correct way. I don't know how slow that would be, so maybe we need to build that logic in somehow as part of sbt, like server etc.

eatkins:
I see what you mean about version. The main advantage of injecting it into the task graph is that it can be run in parallel with the other tasks that evaluate input file stamps. If it were done as a command, I think it'd have to run before everything else. I'll take a look at putting the source checking logic directly in MainLoop.processCommand and see what the impact is.

eed3si9n:
https://gist.github.com/xuwei-k/6278769 uses `shellPrompt`, which is at command level basically.

eatkins:
Right, though that example doesn't quite do what we want because if I am already at the shellPrompt and I type version, it won't be until after I run version that I get a warning.

eatkins:
Replaced by #4753.

PR #4683, opened 2019-05-15
Increase duration between polling from stdin

eatkins:
I was testing the cpu utilization of sbt when idle and noticed that it
was a bit higher than I expected. It was using O(15%) of a cpu on
average. After increasing these periods, the value dropped to O(2%).

### DIFF
--- a/main/src/main/scala/sbt/internal/Idle.scala
+++ b/main/src/main/scala/sbt/internal/Idle.scala
@@ -1,0 +1,19 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.internal
+
+import scala.concurrent.duration._
+import scala.util.control.NonFatal
+
+private[sbt] object Idle {
+  val value: FiniteDuration =
+    try System.getProperty("sbt.idle.seconds", "300").toInt.seconds
+    catch { case NonFatal(_) => 5.minute }
+  val idleTimeout: FiniteDuration = 500.millis
+  val activeTimout: FiniteDuration = 20.millis
+}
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/internal/Idle.scala
```diff
@@ -0,0 +1,19 @@
...
+package sbt.internal
+
+import scala.concurrent.duration._
+import scala.util.control.NonFatal
+
+private[sbt] object Idle {
+  val value: FiniteDuration =
+    try System.getProperty("sbt.idle.seconds", "300").toInt.seconds
+    catch { case NonFatal(_) => 5.minute }
+  val idleTimeout: FiniteDuration = 500.millis
```
eed3si9n, line 17: So in the worst case, it will add 0.5 second latency after you type `test` and hit the enter key?

smarter:
Why does sbt need polling to read stdin ?

main/src/main/scala/sbt/internal/Idle.scala
eatkins: Yeah.

eatkins:
@smarter it is necessary to poll because `InputStream.read` blocks uninterruptibly on java so if you spin up a background thread to read from stdin and write to a queue, you can't shutdown that thread. There is probably a better way to get bytes from stdin, but that's outside of the scope of 1.3.0.

eed3si9n:
> Why does sbt need polling to read stdin ?

`shell` command stops listening to the stdin once it starts processing user entered command. So I tried to implement similar mechanism for commands that came in from sbt server. We can revisit it later.

main/src/main/scala/sbt/internal/Idle.scala
eed3si9n: Could we reduce the latency to somewhere low enough that human would not perceive it but high enough that CPU won't mind?

eatkins:
I am working on a better approach to fixing this.

Issue #4689, opened 2019-05-16
ClassLoaderLayeringStrategy.ScalaLibrary breaks scalatest


pshirshov:
Hi,

Seems like `ClassLoaderLayeringStrategy.ScalaLibrary` in 1.3.0-RC1 breaks scalatest with the a cryptic message. Surprisingly default setting works, as well as `.Flat`. I guess it's not a bug but could anyone explain this behaviour? From my understanding of the new scheme it should work fine.

```
[error] java.lang.NoClassDefFoundError: scala/reflect/runtime/package$
[error] 	at org.scalatest.tools.Framework.runner(Framework.scala:985)
[error] 	at sbt.Defaults$.$anonfun$createTestRunners$1(Defaults.scala:1003)
[error] 	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:237)
[error] 	at scala.collection.immutable.Map$Map1.foreach(Map.scala:128)
[error] 	at scala.collection.TraversableLike.map(TraversableLike.scala:237)
[error] 	at scala.collection.TraversableLike.map$(TraversableLike.scala:230)
[error] 	at scala.collection.AbstractTraversable.map(Traversable.scala:108)
[error] 	at sbt.Defaults$.createTestRunners(Defaults.scala:997)
[error] 	at sbt.Defaults$.allTestGroupsTask(Defaults.scala:1065)
[error] 	at sbt.Defaults$.$anonfun$testTasks$10(Defaults.scala:804)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
[error] 	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:62)
[error] 	at sbt.std.Transform$$anon$4.work(Transform.scala:67)
[error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:270)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:19)
[error] 	at sbt.Execute.work(Execute.scala:279)
[error] 	at sbt.Execute.$anonfun$submit$1(Execute.scala:270)
[error] 	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:178)
[error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:37)
[error] 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[...]

PR #4671, opened 2019-05-13
change ClassLoaderLayeringStrategy.ScalaInstance to ScalaLibrary

eed3si9n:
Fixes #4609

ScalaInstance contains unwanted Scala modules such as scala-xml and scala-parser-combinators.

While  testing #4609 I noticed that picking flat loader didn't fix the problem. This lead me to discover that test cp was including si.allJars, which shouldn't.

PR #4703, opened 2019-05-17
Fixes layer 4 missing scala-reflect

eed3si9n:
Ref #4689
Ref #4671


scripted-sbt-redux/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
```diff
@@ -181,16 +181,13 @@ final class ScriptedTests(
...
-      case "classloader-cache/jni"                       => LauncherBased // sbt/Package$
-      case "classloader-cache/library-mismatch"          => LauncherBased // sbt/Package$
-      case "classloader-cache/runtime-layers"            => LauncherBased // sbt/Package$
-      case "classloader-cache/package-private"           => LauncherBased // sbt/Package$
+      case "actions/add-alias"          => LauncherBased // sbt/Package$
+      case "actions/cross-multiproject" => LauncherBased // tbd
+      case "actions/external-doc"       => LauncherBased // sbt/Package$
+      case "actions/input-task"         => LauncherBased // sbt/Package$
+      case "actions/input-task-dyn"     => LauncherBased // sbt/Package$
+      case gn if gn.startsWith("classloader-cache/") =>
```
eatkins, line 189: Not a huge deal, but this will make some of the tests slower in development because of sbt re-building the compiler-bridge whenever LauncherBased is used.

main/src/main/scala/sbt/internal/ClassLoaders.scala
```diff
@@ -146,8 +146,9 @@ private[sbt] object ClassLoaders {
         val allTestDependencies = if (layerTestDependencies) allDependenciesSet else Set.empty[File]
         val allRuntimeDependencies = (if (layerDependencies) rawRuntimeDependencies else Nil).toSet
 
+        val scalaLibrarySet = Set(si.libraryJar)
         val scalaLibraryLayer =
-          globalCache.get((List(si.libraryJar), interfaceLoader, resources, tmp))
+          globalCache.get((scalaLibrarySet.toList, interfaceLoader, resources, tmp))
```
eatkins, line 151: The one bad thing about this is that if the scala library version is the same as the sbt build scala version, then we somewhat needlessly create a new classloader for the scala library. This could possibly be improved by checking if the appConfiguration.scalaProvider.version matches the scalaVersion and using the sbt library classloader if so. That can be handled in a different PR, however.

main/src/main/scala/sbt/internal/ClassLoaders.scala
```diff
@@ -146,8 +146,9 @@ private[sbt] object ClassLoaders {
         val allTestDependencies = if (layerTestDependencies) allDependenciesSet else Set.empty[File]
         val allRuntimeDependencies = (if (layerDependencies) rawRuntimeDependencies else Nil).toSet
 
+        val scalaLibrarySet = Set(si.libraryJar)
```
eatkins, line 149: Why Set vs. List?

eatkins (approved):
LGTM.

scripted-sbt-redux/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
eed3si9n: For classloader testing forking would be safer.

main/src/main/scala/sbt/internal/ClassLoaders.scala
eed3si9n: I guess because `testDependencySet` etc are also sets, and you can use it later as `fullCP.filterNot(testDependencySet ++ runtimeDependencies ++ scalaLibrarySet)`, but this could go either way.

PR #4713, opened 2019-05-21
Make coursier-related tasks public

smarter:
This follows the discussion in
https://github.com/coursier/coursier/issues/1181.


eed3si9n (approved):
I was hoping that LM API abstraction is enough, but the implementation difference in resolution is leaking about especially around priorities involving inter-project vs Maven Central, so I'll agree to opening up the details on Coursier configuration.

eatkins:
This seems ok given that all of these are actually in the shaded coursier package. I'd only be concerned if they were exposing classes in the actual coursier package.

main/src/main/scala/sbt/internal/LMCoursier.scala
```diff
@@ -22,7 +22,7 @@ import sbt.internal.librarymanagement.{ CoursierArtifactsTasks, CoursierInputsTa
 import sbt.util.Logger
 import sbt.io.syntax._
 
-private[sbt] object LMCoursier {
+object LMCoursier {
```
dwijnand, line 25: Change package?

eed3si9n: Yea. I was gonna follow up with some moving things around.

PR #4725, opened 2019-05-24
scalaCompilerBridgeDependencyResolution

### LINKED ISSUE #4712: [1.3.0-RC1] regression: bridge sources JAR not downloaded

## steps

Please see https://github.com/dotta/sbt-1.3.0-bridge-download-regression/ for a full reproducible with explanations in the README.

## problem

The configured bridge sources are not downloaded now that Coursier is the default dependency manager. It works fine if Ivy is used instead.

## expectation

The same resolvers are used to download the bridge component with both Ivy and Coursier.

## notes

sbt version: *1.3.0-RC1*

eed3si9n:
Fixes #4712

This adds a specialized DependencyResolution instance called `scalaCompilerBridgeDependencyResolution` to download the compiler bridge. It has its own list of resolvers set by `scalaCompilerBridgeResolvers`. For backward compatibility, it will append `externalResolvers.value` as well.


dotta:
Excellent. Thanks a lot, Eugene!

Using `externalResolvers` for backward compatibility should work. In fact, do you plan to add the reproducible I linked in the ticket to the scripted tests?

eed3si9n:
@dotta I don't know if I can since both failure and success case would look like failed compilation.

dotta:
Is there a way to trigger an `update` just to download the bridge and then check if it's been downloaded? Alternatively, maybe we could check that a resolver added to the `projectSettings` is part of the resolvers used to download the bridge sources?

eed3si9n:
> Is there a way to trigger an update just to download the bridge and then check if it's been downloaded?

No, not really. Bridge resolution happens inside of zinc-lm thing.

> Alternatively, maybe we could check that a resolver added to the projectSettings is part of the resolvers used to download the bridge sources?

I guess I could do that.

dotta:
> I guess I could do that.

Don’t feel like you have to. If you think this behavior doesn’t warrant a test, I’m totally fine with it. Thanks again for the prompt fix!

eed3si9n:
I think a true integration test would be if I can download some dummy bridge that does something different, `compile`, and observe the side effect. Doing the resolver test certainly would be an improved step to prevent future regression.

Issue #4715, opened 2019-05-22
Failure running +project/testOnly 


ashawley:
## steps

In a multi-project build for a library that cross-compiles, run a single test across multiple Scala versions:

```
> +foo/testOnly foo.BarTest
```

I'll try to send a scripted test in a PR to prove this is not working.

## problem

This used to work in sbt 0.13.  In sbt 1.2.8, there's a runtime exception:

```
[error] java.util.NoSuchElementException: key not found: ProjectRef(file:project/,root)
[error] 	at scala.collection.immutable.Map$Map1.apply(Map.scala:111)
[error] 	at sbt.Cross$.crossBuildCommandImpl(Cross.scala:167)
[error] 	at sbt.Cross$.$anonfun$crossBuild$2(Cross.scala:123)
[error] 	at sbt.Cross$$$Lambda$386/2095970915.apply(Unknown Source)
[error] 	at sbt.Command$.$anonfun$applyEffect$4(Command.scala:142)
[error] 	at sbt.Command$$$Lambda$330/509806761.apply(Unknown Source)
[error] 	at sbt.Command$.$anonfun$applyEffect$2(Command.scala:137)
[error] 	at sbt.Command$$$Lambda$356/756637412.apply(Unknown Source)
[error] 	at sbt.Command$.process(Command.scala:181)
[error] 	at sbt.MainLoop$.processCommand(MainLoop.scala:151)
[error] 	at sbt.MainLoop$.$anonfun$next$2(MainLoop.scala:139)
[error] 	at sbt.MainLoop$$$Lambda$311/1800932232.apply(Unknown Source)
[error] 	at sbt.State$$anon$1.runCmd$1(State.scala:246)
[error] 	at sbt.State$$anon$1.process(State.scala:250)
[error] 	at sbt.MainLoop$.$anonfun$next$1(MainLoop.scala:139)
[error] 	at sbt.MainLoop$$$Lambda$310/2000410491.apply(Unknown Source)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:16)
[error] 	at sbt.MainLoop$.next(MainLoop.scala:139)
[error] 	at sbt.MainLoop$.run(MainLoop.scala:132)
[error] 	at sbt.MainLoop$.$anonfun$runWithNewLog$1(MainLoop.scala:110)
[error] 	at sbt.MainLoop$$$Lambda$303/1226298219.apply(Unknown Source)
[...]

PR #4727, opened 2019-05-24
Fix +foo/testOnly foo.BarTest

ashawley:
After tracking down the change that caused #4715, I took a stab at trying to fix it.  I may have managed to bring back cross-building a sub-project command again, but I'm not sure if it is correct, or if it breaks other things.

In particular, the change here breaks the ability to cross-build a command alias, which I only noticed because be91300 added a test for it.  There may be a way to make it work, but I couldn't see how.


eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/5265

Issue #4649, opened 2019-05-04
.descendantsExcept issue


eed3si9n:
## steps

https://github.com/lihaoyi/Scalatex

```scala
unmanagedSources in Compile ++= {
  val root = baseDirectory.value.getParentFile
  val exclude = Map(
    "0.13" -> "1.0",
    "1.0" -> "0.13"
  ).apply(sbtBinaryVersion.value)
  val xs = (root / "scalatexSbtPlugin")
    .descendantsExcept("*.scala", s"*sbt-$exclude*")
    .get
  // println(xs.toList)
  xs
}
```

## problem

`.descendantsExcept` doesn't seem to exclude any more, so both 0.13 and 1.0 directories are included.

## expectation

`.descendantsExcept` filters out sbt-0.13 files.

eed3si9n:
Fixed in #4664

eed3si9n:
Tested on M4 to confirm. I need the following change but it seems like it's a plugin bug not ours.

```diff
@@ -13,12 +15,13 @@ object SbtPlugin extends sbt.AutoPlugin {
   val mySeq = Seq(
     scalatexDirectory := sourceDirectory.value / "scalatex",
     managedSources ++= {
-      val inputDir = scalatexDirectory.value
+      val inputDir = scalatexDirectory.value.getAbsoluteFile
```

PR #4728, opened 2019-05-25
Reproduce Twirl regression

### LINKED ISSUE #4681: [1.3.0-RC1] Defaults.collectFiles no longer returns expected sources

## environment:
```
$ java -version
openjdk version "11.0.2" 2019-01-15
OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)

$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.2 LTS
Release:	18.04
Codename:	bionic

$ uname -a
Linux 4.15.0-48-generic #51-Ubuntu SMP Wed Apr 3 08:28:49 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

## steps

```
$ sbt new playframework/play-java-seed.g8
...
name [play-java-seed]: <enter>
organization [com.example]: <enter>
...
$ cd play-java-seed/
play-java-seed$ echo "sbt.version=1.3.0-RC1" > project/build.properties
play-java-seed$ sbt
[play-java-seed] $ compile
```

## problem

You should see following compiler error:
```
[info] Compiling 4 Scala sources and 2 Java sources to /tmp/zinc_check/play-java-seed/target/scala-2.12/classes ...
[error] /tmp/zinc_check/play-java-seed/app/controllers/HomeController.java:18:1: cannot find symbol
[error]   symbol:   class index
[error]   location: package views.html
[error] views.html.index
[error] (Compile / compileIncremental) javac returned non-zero exit code
[error] Total time: 6 s, completed May 15, 2019, 12:21:27 PM
```

## expectation

It should compile.
**It does compile with sbt 1.2.8!**

## notes

sbt version: 1.3.0-RC1

eed3si9n:
Ref #4681

<s>Given https://github.com/sbt/io/commit/f8f9aed4a8cedef5526ee4679423dbe2a51182a7, I'm guessing this is the way to go for now.</s>


eed3si9n:
We can ignore continuous-integration/travis-ci/push. I pushed this to `develop` branch by mistake first.

eatkins:
We need to understand this before we just revert. Have you tried this with the io based on https://github.com/sbt/io/pull/240? I bet there is a good chance that the scripted test would pass with that.

eatkins:
Actually, I just verified locally that the test passes with the new io and fails with 1.3.0-M10. The problem was excludeFilter := HiddenFileFilter, not the includeFilter.

eed3si9n:
> Actually, I just verified locally that the test passes with the new io and fails with 1.3.0-M10. The problem was excludeFilter := HiddenFileFilter, not the includeFilter.

ok. Then I'm happy to drop the reversion and just keep the scripted.

eatkins (changes requested):
I'm ok with the revert, but the test should probably guard against the descendantsExcept regression rather than the HiddenFileFilter issue in https://github.com/sbt/io/pull/240. The new collectFiles implementation doesn't handle descendantsExcept correctly. Adding this patch would test the regression:

```diff
diff --git a/sbt/src/sbt-test/plugins/twirl/build.sbt b/sbt/src/sbt-test/plugins/twirl/build.sbt
index d3263d927..bee7bb4cf 100644
--- a/sbt/src/sbt-test/plugins/twirl/build.sbt
+++ b/sbt/src/sbt-test/plugins/twirl/build.sbt
@@ -5,6 +5,8 @@ lazy val root = (project in file("."))
   .enablePlugins(TwirlPlugin)
   .settings(
     check := {
-      assert((Compile / twirlCompileTemplates / sources).value.nonEmpty)
+      val templates = (Compile / twirlCompileTemplates / sources).value
+      assert(templates.nonEmpty)
+      assert(!templates.exists(_.toString.contains("hidden")))
     }
   )
diff --git a/sbt/src/sbt-test/plugins/twirl/src/main/twirl/.hidden/fake.scala.twirl b/sbt/src/sbt-test/plugins/twirl/src/main/twirl/.hidden/fake.scala.twirl                                                                                                                              
new file mode 100644
index 000000000..e69de29bb
```

eed3si9n:
Interesting. So there was a double bug around this area, and the test wasn't tight enough?

eatkins:
Yeah. Pretty much the same issue as https://github.com/sbt/sbt/issues/4649.

eatkins:
The updated test doesn't exactly test the same thing as the example in the diff that I provided. descendantsExcept is often used to exclude entire directories (the example excludeFilter in the scala doc is ".svn"). The glob based collectFiles implementation only applies the exclude filter to the final path name while descendantsExcept method applies the filter to _every_ path name in in the path. The updated test also passes with the new io but without the collectFiles revert. If you apply the diff that I provided above, it fails.

eed3si9n:
I was wondering how you're testing the hidden files, and I now noticed

```
diff --git a/sbt/src/sbt-test/plugins/twirl/src/main/twirl/.hidden/fake.scala.twirl b/sbt/src/sbt-test/plugins/twirl/src/main/twirl/.hidden/fake.scala.twirl                                                                                                                              
new file mode 100644
index 000000000..e69de29bb
```

eatkins:
Yeah, that was subtle. I could have pointed that out better.

eatkins (approved):

Issue #4711, opened 2019-05-20
1.3.0-RC1 regression: NoClassDefFoundError in tests


Atry:
- [x] I've read the [CONTRIBUTING](https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md) guidelines

## steps
``` scala
git clone https://github.com/Atry/Dsl.scala.git --branch=pr3 && cd Dsl.scala && sbt test
```

## problem

```

[info] com.thoughtworks.dsl.keywords.ContinueExample *** ABORTED ***
[info]   java.lang.NoClassDefFoundError: com/thoughtworks/dsl/Dsl$Keyword
[info]   at java.lang.ClassLoader.defineClass1(Native Method)
[info]   at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
[info]   at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
[info]   at java.net.URLClassLoader.defineClass(URLClassLoader.java:468)
[info]   at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
[info]   at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
[info]   at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
[info]   at java.security.AccessController.doPrivileged(Native Method)
[info]   at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
[info]   at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
[info]   ...
[info]   Cause: java.lang.ClassNotFoundException: com.thoughtworks.dsl.Dsl$Keyword
[info]   at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
[info]   at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
[info]   at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
[info]   at java.lang.ClassLoader.defineClass1(Native Method)
[info]   at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
[info]   at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
[info]   at java.net.URLClassLoader.defineClass(URLClassLoader.java:468)
[info]   at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
[info]   at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
[info]   at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
[info]   ...
```

[...]

Issue #4693, opened 2019-05-16
ClassLoaderLayeringStrategy IlegalArgumentException when extending test configuration


davidhoyt:
- [x] I've read the [CONTRIBUTING](https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md) guidelines

## steps
Derive a new configuration from an existing one. e.g.:

```scala
import sbt._

object SmokeTestPlugin extends AutoPlugin {
  object autoImport {
    lazy val SmokeTest = Configuration.of("SmokeTest", "smoke") extend (Runtime, Test)
  }

  import autoImport._

  override val trigger = allRequirements

  lazy val smokeSettings: Seq[Setting[_]] = inConfig(SmokeTest)(Defaults.testSettings)

  override def projectConfigurations: Seq[Configuration] = Seq(SmokeTest)

  override lazy val projectSettings: Seq[Setting[_]] = smokeSettings
}
```

And then use it:

```scala
lazy val foo = Project
```

## problem
The following is output:
```
java.lang.IllegalArgumentException: Layering strategy TestDependencies is not valid for the classloader in Scope(Select(ProjectRef(file:/path/to/project/,foo)), Select(ConfigKey(smoke)), Zero, Zero). Valid options are: ClassLoaderLayeringStrategy.{ Flat, ScalaInstance, RuntimeDependencies }
```

## expectation
A new configuration extending `Test` should just work. In this case `extend (Runtime, Test)` is important because any other variant and IntelliJ will not recognize the new configuration's source directories.

## notes

sbt version: 1.3.0-RC1

eed3si9n:
@davidhoyt Thanks for the report!

davidhoyt:
And thanks to you and all the contributors for all the hard work that's gone into this release!

eed3si9n:
Fixed in https://github.com/sbt/sbt/pull/4729

Issue #4691, opened 2019-05-16
Default ClassLoaderLayeringStrategy breaks some code which uses dirty magic to resolve singletons


pshirshov:
I have a project with a custom `StaticLoggerBinder` for `slf4j`.

I noticed that slf4j started failing picking up this (dirty and shitty) singleton class while running my tests under 1.30-RC1.

The behaviour observed is explainable.

Also there are other issues, like I'm building my microsite with paradox and sbt-microsite and noticed that sbt throws a lot of `NoClassDefFoundError`s during shutdown.

```
Uncaught error from thread [sbt-web-akka.actor.default-dispatcher-2]: scala/concurrent/Future$$anon$1, shutting down JVM since 'akka.jvm-exit-on-fatal-error' is enabled for for ActorSystem[sbt-web]
java.lang.NoClassDefFoundError: scala/concurrent/Future$$anon$1
	at scala.concurrent.Future$.firstCompletedOf(Future.scala:683)
	at akka.actor.CoordinatedShutdown.loop$1(CoordinatedShutdown.scala:382)
	at akka.actor.CoordinatedShutdown.$anonfun$run$6(CoordinatedShutdown.scala:387)
	at scala.concurrent.Future.$anonfun$flatMap$1(Future.scala:307)
	at scala.concurrent.impl.Promise.$anonfun$transformWith$1(Promise.scala:41)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
	at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:55)
	at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:91)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
	at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:85)
	at akka.dispatch.BatchingExecutor$BlockableBatch.run(BatchingExecutor.scala:91)
	at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:40)
	at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(ForkJoinExecutorConfigurator.scala:43)
	at akka.dispatch.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
[...]

Issue #4686, opened 2019-05-16
[1.3.0-RC1] java.lang.OutOfMemoryError: Metaspace


eed3si9n:
_Originally posted by @limansky in https://github.com/sbt/sbt/issues/4681#issuecomment-493146377_

## steps

1. get beanpuree https://github.com/limansky/beanpuree
2. set sbt version to 1.3.0-RC1
3. sbt
4. +test

## problem

I'm not sure if I should submit another ticket, but when I tried to compile BeanPuree with sbt 1.3.0-RC1 I've got OutOfMemoryError:
```
java.lang.OutOfMemoryError: Metaspace
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:468)
```

## expectation

Everything works fine with 1.2.8.

## workaround

Set `-XX:MaxMetaspaceSize=1024M` or whatever you're currently using for `-Xmx` in `JAVA_OPTS` environment variable.

eed3si9n:
I wasn't able to reproduce this either locally or on Travis - https://travis-ci.org/limansky/beanpuree/builds/533456390

eatkins:
I was able to reproduce on my machine.

eed3si9n:
Travis uses 

```
java
-Xms2048M
-Xmx2048M
-Xss6M
-XX:MaxPermSize=512M
```

could you post your mem settings?

eatkins:
```
ethanatkins@Ethans-MacBook-ProZ:~/work/oss/sbt/community/beanpuree$ sbt -Dsbt.version=1.3.0-RC1 -v
[process_args] java_version = '11'
[copyRt] java9_rt = '/Users/ethanatkins/.sbt/0.13/java9-rt-ext-oracle_corporation_11_0_2/rt.jar'
# Executing command line:
java
-Xms1024m
-Xmx1024m
-XX:ReservedCodeCacheSize=128m
-XX:MaxMetaspaceSize=256m
-XX:+UseParallelGC
-XX:+HeapDumpOnOutOfMemoryError
-Dscala.ext.dirs=/Users/ethanatkins/.sbt/0.13/java9-rt-ext-oracle_corporation_11_0_2
-jar
[...]

PR #4729, opened 2019-05-27
Classloader caching

### LINKED ISSUE #4714: [1.3.0-RC1] src/test/resources are not visible from test libraries
I had some issues where `logback-test.xml` wasn't getting picked up with the new 1.3.0-RC1 release, but setting the `classLoaderLayeringStrategy` to flat like you suggested above fixes it. I'm curious what's different in the default behavior that files in `src/test/resources` aren't getting put into the test classpath?

_Originally posted by @Ophirr33 in https://github.com/sbt/sbt/issues/4711#issuecomment-494458178_

eatkins:

This PR unifies and enhances the internal sbt ClassLoaderCache. It generally improves the memory utilization quite a bit. After these changes, I found the metaspace utilization of 1.3.x to be comparable to and sometimes less than 1.2.8. I also found a memory leak that was causing the state to leak with each task evaluation (it was leaking about 3mb of java heap per task evaluation in my test project).

The reason that the metaspace utilization is now in some cases lower is that I was careful to ensure that we always use the state's classLoaderCache when making new loaders. The ClassLoaderCache will evict unused loaders in response to memory pressure. If the MaxMetaspaceSize is _not_ set, we are able to close the underlying URLClassLoader. If that parameter is set, we will leak file handles (which is what sbt < 1.3.0 always did), but the program will not crash if the jvm is able to free up enough space after evicting them. Moreover, there were a number of places where we were making duplicates of the scala-library loader even when it was the same version as the meta build. Re-using the meta build library loader is how the 1.3.x metaspace utilization can be lower than 1.2.8.

I also added a resource layer before the dependency layer. This should allow users to use a dependency layer and still have their code that was dependent on the libraries having access to resource files work. To make this work, I had to remove the ShareRuntimeDependenciesWith... layering strategy. Removing that strategy made it pointless to have RuntimeDependencies and TestDependencies, so I consolidated them into Dependencies. This technically breaks binary compatibility but I think the fact that resources didn't work is a big enough bug that it's worth breaking compatibility.

I also found a path in test where it was possible for the test to throw a NoClassDefError but not report a big warning to the user suggesting that they use the ScalaLibrary or Flat strategies. This should make it less scary to set the default layering strategy to Dependencies, but we still have an opportunity to switch to ScalaLibrary before 1.3.0. FWIW, any build that was working in 1.2.8 should work with ScalaLibrary so I think that is a significantly better option that Flat for the default.

Requires https://github.com/sbt/zinc/pull/674

Fixes #4714, #4711, #4693, #4691, #4686


main/src/main/scala/sbt/ClassLoaderLayeringStrategy.scala
```diff
@@ -89,36 +89,12 @@ object ClassLoaderLayeringStrategy {
   case object ScalaLibrary extends ScalaLibrary
 
   /**
-   * Add a layer on top of the ScalaLibrary layer for the runtime jar dependencies.
+   * Add a layer on top of the ScalaLibrary layer for all of the task jar dependencies.
    */
-  sealed trait RuntimeDependencies extends ScalaLibrary
+  sealed trait Dependencies extends ScalaLibrary
```
eed3si9n: The term "dependencies" comes up too often in the context of sbt, including `import Dependencies._`, so I think `RuntimeDependencies` is a bit more clearer than `Dependencies` even if it means sometimes test dependencies too.

eatkins: I'm open to a completely different name then.

eatkins: Kinda ugly, but maybe `ScalaLibraryThenDependencyJars?` It's at least fairly clear.

eed3si9n: `LibraryJars`, `ExternalJars`, `LayerDependencyJars`?

eatkins: I think of those I like ExternalJars the best just because ScalaLibrary already has Library in it.

main-command/src/main/scala/sbt/internal/classpath/ClassLoaderCache.scala
```diff
@@ -0,0 +1,217 @@
...
+import xsbti.compile.{ ClasspathOptions, ScalaInstance }
+import xsbti.{ AppConfiguration, ExtendedScalaProvider }
+
+import scala.annotation.tailrec
+import scala.collection.JavaConverters._
+
+private object ClassLoaderCache {
+  private def threadID = new AtomicInteger(0)
+}
+private[sbt] class ClassLoaderCache(val configuration: AppConfiguration)
```
eed3si9n: It might be good to add some unit tests for this new cache to test some of the invariants?

main-command/src/main/scala/sbt/internal/classpath/ClassLoaderCache.scala
```diff
@@ -0,0 +1,217 @@
...
+        cleanup()
+        false
+      } catch {
+        case _: InterruptedException => true
+      }
+      if (!stop) run()
+    }
+  }
+  private[this] val mkReference: (Key, ClassLoader) => Reference[ClassLoader] =
+    if (metaspaceIsLimited)(_, cl) => new SoftReference(cl, referenceQueue)
```
eed3si9n, line 140: I can't think of other places in sbt where we are doing something special based on the memory parameter like this.
It might be good to leave some comment here explaining the current tradeoff.

main-command/src/main/scala/sbt/internal/classpath/ClassLoaderCache.scala
eatkins: Oh, I actually have a bunch of tests of the old ClassLoaderCache in the main project that I forgot to delete and won't compile now anyway. I'll update the PR tomorrow with the api changes needed to make them run in the command project.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -717,8 +717,19 @@ object Defaults extends BuildCommon {
         val scalaProvider = appConfiguration.value.provider.scalaProvider
         val version = scalaVersion.value
         if (version == scalaProvider.version) // use the same class loader as the Scala classes used by sbt
-          Def.task(ScalaInstance(version, scalaProvider))
-        else
+          Def.task {
+            try {
+              val allJars = scalaProvider.jars
+              val Array(libraryJar) = allJars.filter(_.getName == "scala-library.jar")
```
eed3si9n: 1. I don't think we can assume that the name `scala-library.jar` will remain that way for all implementations of Scala. That's why ScalaInstance returns `libraryJar`.
2. As of https://github.com/sbt/zinc/pull/677, it can return multiple files as `libraryJars`.

main-command/src/main/scala/sbt/internal/classpath/ClassLoaderCache.scala
eatkins: Fair enough. It is mentioned in the commit message but this probably deserves higher visibility.

eed3si9n (changes requested):

main/src/main/scala/sbt/Defaults.scala
eatkins: Yeah. That is why I catch the exception and revert to the old approach if that doesn’t work (though I didn’t mean to leave the e.printStackTrace() in). I can look into incorporating the change from https://github.com/sbt/zinc/pull/677 though since I require a later version if zinc anyway.

eatkins: I figure though that any future changes to the scala compiler api require zinc/sbt coordinations anyway so I don’t know that it’s really possible to future proof this.

eatkins: I reworked this commit to use the new libraryJars api. The specific code block where you were worried about assuming that the scala library is called "scala-library.jar" seems completely safe the more I think about it. We only execute that code if the scalaVersion is the same as the application scala provider version. I can't figure out how that assumption would be broken without a lot of other things breaking. We also already assume in numerous places that the jars have speciifc names, see e.g. sbt.librarymanagement.ScalaArtifacts. I did, however, rework that code block to be more obviously safe in https://github.com/sbt/sbt/pull/4729/commits/d05344527ee670c6a07be28a3e48626235b9cab8.

main-command/src/main/scala/sbt/internal/classpath/ClassLoaderCache.scala
eatkins: https://github.com/sbt/sbt/pull/4729/commits/61bfba0dedd77fa12d7deebc6e6b69c21f18219f

main/src/main/scala/sbt/ClassLoaderLayeringStrategy.scala
eatkins: I picked `AllLibraryJars` to distinguish from `ScalaLibrary`.

main-command/src/main/scala/sbt/internal/classpath/ClassLoaderCache.scala
eatkins: https://github.com/sbt/sbt/pull/4729/commits/53d609669cef1a1b0af1e4eb94b267a5d3538003 ports the old ClassLoaderCacheTest. The new cache doesn't support a size limit (yet) so I removed one of the tests. It isn't clear to me that there is much benefit to manually specifying the size limit since the actual size impact of each classloader can vary greatly and we can't easily measure it. For now, I think just having the jvm manage the cache implicitly through the soft references seems like the best approach until we are sure that it's a problem.

eatkins:
Finally green on CI.

eed3si9n (approved):

PR #4730, opened 2019-05-27
Move external hooks definition

### LINKED ISSUE #4624: externalHooks-related issues
## steps

1.3.0-RC1.

## problem

<s>1. The key `externalHooks` is not an intuitive name for sbt. In the context of sbt, "external" typically refers to libraryDependencies as opposed to other subprojects.
2. It's a function that accepts CompileOptions.</s>
3. Composability issue pointed out by @smarter  in https://github.com/sbt/sbt/commit/25e97f99f588d1ddc9bbd114740bfb25c2a955c6#r32293747.

> zinc's ExternalHooks are extremely tricky to use, I'm not sure what the current semantics are (it's changed a few times and it's confusing as hell) but I think this might overwrite existing externalhooks set in a plugin or by the user, I'm thinking in particular of the scala-js sbt plugin (https://github.com/scala-js/scala-js/blob/0ff1fbe3f7a003482c3b785b6a374202f44f47c1/sbt-plugin/src/main/scala-sbt-1.0/org/scalajs/sbtplugin/SBTCompat.scala#L57) and the dotty sbt plugin (https://github.com/lampepfl/dotty/blob/58b8aaa59eae3c05807b318b60148f13bacbe610/sbt-dotty/src/dotty/tools/sbtplugin/DottyPlugin.scala#L130)

## expectation

<s>
Idea 1-A:
Rename to `compileHooks`, use `compileOptions.value`, and type to `Option[ExternalHooks]`?

Compose with existing hooks. ScalaJS for example does:

```scala
    val newExternalHooks =
      incOptions.externalHooks.withExternalClassFileManager(sjsirFileManager)
    incOptions.withExternalHooks(newExternalHooks)
```
</s>
Idea 1-B:
A better solution might be to have a global Boolean setting `reuseFileLookup` (name subject to change), and push the change to `incOptions` task.

## notes

1. `sbt.internal.ExternalHooks` is concerned about lookup.
2. Dotty and ScalaJS are concerned about fileManager.

eatkins:
I verified manually that ExternalHooks were still applied by default but
that I could set the incOptions in the Test and Compile configs so that
they weren't used.

Fixes #4624

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -609,12 +609,8 @@ object Defaults extends BuildCommon {
         else ""
       s"inc_compile$extra.zip"
     },
-    compileIncSetup := {
-      val base = compileIncSetupTask.value
-      val incOptions =
-        base.incrementalCompilerOptions.withExternalHooks(ExternalHooks.default.value)
-      base.withIncrementalCompilerOptions(incOptions)
-    },
+    incOptions := { incOptions.value.withExternalHooks(ExternalHooks.default.value) },
+    compileIncSetup := compileIncSetupTask.value,
     console := consoleTask.value,
     collectAnalyses := Definition.collectAnalysesTask.map(_ => ()).value,
     consoleQuick := consoleQuickTask.value,


eed3si9n (approved):

eatkins:
The appveyor failure is network related and not related to my change (one of the builds did in fact pass).

smarter:
To clarify, is https://github.com/lampepfl/dotty/blob/2e5462483a9f0628e9c59ae31c165c04a45c9d96/sbt-dotty/src/dotty/tools/sbtplugin/DottyPlugin.scala#L111-L136 and https://github.com/lampepfl/dotty/blob/2e5462483a9f0628e9c59ae31c165c04a45c9d96/sbt-dotty/src/dotty/tools/sbtplugin/DottyPlugin.scala#L162-L168 still the correct way to add an additional external hook on top of the existing ones ? I'm OK with changing this code in a way that only works in sbt 1.3+, I just want something that works.

eatkins:
@smarter I think that prior to this change, your dotty external hooks changes would have been blown away by sbt. I haven't verified the dotty plugin myself after these changes, but it looks correct to me.

PR #4733, opened 2019-05-28
Replace usages of deprecated ScalaInstance#libraryJar

### LINKED ISSUE #4705: sbt 1.3 classloader layering: support more than one jar for the standard library
Dotty's standard library is currently effectively represented by two jars: a scala-library jar compiled by Scala 2 and its own dotty-library jar. Reusing the scala-library jar from Scala 2 is what makes it possible for Dotty code to reuse Scala 2 libraries. However, sbt is not aware of this because `ScalaInstance` in zinc requires using a single File for the standard library. So far this hasn't been too much of a problem (see https://github.com/lampepfl/dotty/blob/fb6667b0fbdf1e2811f36d5dd2909905d55735ee/sbt-dotty/src/dotty/tools/sbtplugin/DottyPlugin.scala#L194-L228) but sbt 1.3 classloader layering logic relies on ScalaInstance, so that might end up impacting users in one way or another.

smarter:


eed3si9n:
```scala
[info] [info] Loading project definition from /tmp/sbt_f49ab8cf/project
[info] [info] Compiling 1 Scala source to /tmp/sbt_f49ab8cf/project/target/scala-2.12/sbt-1.0/classes ...
[info] [info] Done compiling.
[error] /tmp/sbt_f49ab8cf/build.sbt:18: error: No Append.Values[sbt.Keys.Classpath, Array[java.io.File]] found, so Array[java.io.File] cannot be appended to sbt.Keys.Classpath
[error]     unmanagedClasspath in Compile ++= scalaInstance.map(_.libraryJars).value,
[error]                                   ^
```

eed3si9n (approved):

PR #4739, opened 2019-05-29
Drop the remaining TupleSyntax usage

dwijnand:


main-settings/src/main/scala/sbt/Structure.scala
```diff
@@ -99,10 +99,7 @@ sealed abstract class SettingKey[T]
 
   protected[this] def make[S](other: Initialize[S], source: SourcePosition)(
       f: (T, S) => T
-  ): Setting[T] = {
-    import TupleSyntax._
-    set((this, other)(f), source)
-  }
+  ): Setting[T] = set(this.zipWith(other)(f), source)
```
eed3si9n, line 102: `Initialize#zipWith` is like a specialized tuple syntax for Tuple2?

eed3si9n (approved):

main-settings/src/main/scala/sbt/Structure.scala
dwijnand: Yep.  It's special-cased and wires directly to Apply, using the Tuple2 instance of AList.

PR #4744, opened 2019-05-30
lm-coursier-shaded 1.1.0-M14-3

### LINKED ISSUE #4738: Ensure coursier is ready for sbt 1.3.0
Tracking https://github.com/coursier/coursier/issues/1174 in the 1.3.0 milestone so it doesn't get forgotten about.

eed3si9n:
Fixes #4738

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -12,6 +12,7 @@ import java.net.{ URI, URL, URLClassLoader }
 import java.util.Optional
 import java.util.concurrent.{ Callable, TimeUnit }
 
+import lmcoursier.CoursierDependencyResolution
 import lmcoursier.definitions.{ Configuration => CConfiguration }
 import org.apache.ivy.core.module.descriptor.ModuleDescriptor
 import org.apache.ivy.core.module.id.ModuleRevisionId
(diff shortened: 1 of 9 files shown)


main/src/main/scala/sbt/internal/librarymanagement/CoursierInputsTasks.scala
```diff
@@ -58,16 +58,14 @@ private[sbt] object CoursierInputsTasks {
       sv,
       sbv
     )
-    val proj1 = proj0.copy(
-      dependencies = proj0.dependencies.map {
-        case (config, dep) =>
-          (config, dep.copy(exclusions = dep.exclusions ++ exclusions0))
-      }
-    )
+    val proj1 = proj0.withDependencies(proj0.dependencies.map {
```
eatkins, line 60: did a case class get converted to contraband class?

eed3si9n: Yea. For bincompat - https://github.com/coursier/sbt-coursier/pull/73

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1910,6 +1911,19 @@ object Defaults extends BuildCommon {
         }
       )
     )
+
+  def dependencyResolutionTask: Def.Initialize[Task[DependencyResolution]] =
```
eatkins, line 1916: Could we move LibraryManagement out of Internal and make it public? I like having as little as possible in Defaults.scala.

eatkins (approved):
I don't care that much about where dependencyResolutionTask ends up so I'm fine merging as is.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: For pure code I'd agree that they should all get kicked out. For tasks and settings, I think it makes sense to include any tasks a plugin author may use in `Defaults.scala` so they would understand what transitive tasks are involved. This is especially true of dynamic tasks like this one since reading the code might give clue for debugging the behavior.

PR #4748, opened 2019-05-30
Indicate `r`etry is the option applied if users just press RETURN

ignasi35:
The message:

`Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore?`

is not explicit about `retry` being the option used when pressing `return` (instead of `r`, `q`, `l`...).


eed3si9n (dismissed):
Thanks @ignasi35

eatkins:
Is this a universal indicator of the default? This change would just make me think that I need to type a capital 'R' instead of a lower case 'r'.

eed3si9n:
I've seen it more like this:

Would like some fries with that? Y/n

eatkins:
I took a stab at this: https://github.com/sbt/sbt/pull/4752

eatkins:
Having to type a newline had been bugging me for a while.

eed3si9n:
ok. Putting this on hold for 1.3.0.

ignasi35:
How about:

`Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? [r]`

So instead of denoting default by capitalization it's explicitly listed in square brackets after the question mark.

main/src/main/scala/sbt/Main.scala
```diff
@@ -741,7 +741,7 @@ object BuiltinCommands {
   @tailrec
   private[this] def doLoadFailed(s: State, loadArg: String): State = {
     val result = (SimpleReader.readLine(
-      "Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? "
+      "Project loading failed: (R)etry, (q)uit, (l)ast, or (i)gnore? "
```
ignasi35:
```suggestion
      "Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? [r]"
```

jsoref:
Capitalized letters for defaults is fairly standard.

```
unzip file.zip
...
replace file? [y]es, [n]o, [A]ll, [N]one, [r]ename:
```
```
gcloud components update
...
Do you want to continue (Y/n)?
```

There are lots of examples of this...

eatkins:
That may be. I don't find it good though.
> replace file? [y]es, [n]o, [A]ll, [N]one, [r]ename:

Are there two defaults here? There are two capitalized options. I can't tell what the impact of pressing enter is.

jsoref:
Oops, indeed, that's a bad example... bad infozip (the default is none).

jsoref:
```
# apt remove `cat`
...
Do you want to continue? [Y/n]
```

eatkins:
I guess my point is, maybe I'm dumb, but I've never recognized that the capital letter signifies the default and it makes me think that I'm expected to input 'Y' not 'y'. I just think it's bad design.

jsoref:
https://askubuntu.com/questions/322600/do-you-want-to-continuey-n-why-the-upper-case/322601

It really is a convention. There are the rare outliers, but old-school folks recognize and appreciate it.

dwijnand (dismissed):

eatkins:
I would rather have a solution that is understood by all users. I now understand that this is a convention in unix, but I had never noticed it until it was pointed out in this thread. Unix has a lot of obscure conventions that I don’t necessarily think should persist just because they’ve been around for 40 years. Using a capital letter to indicate a default is inherently ambiguous because options are often case sensitive. I think we can do better.

dwijnand:
I think you're making an entirely reasonable argument.  Except

> because options are often case sensitive

In simple interactive text UIs you'd probably try and find dedicated keys for each behaviour, which would make them lower-case keypresses.  So 'h' for 'help', 'q' for 'quit', etc.  If you ever get to upper-case letters I'd say you'd probably have to most success if you make them quite _related_ to their lower-case equivalent keypress.  So like 'V' in vim is like 'v' but for the whole line.

In this case, either 'r' or 'R' will work because of the `toLowerCase`.  So, while users unaware of the "upper-case to denote default" convention, they're not going to be strongly misled.  At most, they'll be holding shift when they needn't.

I actually don't mind if this change or #4752 is accepted.  Then we can tackle the fact that it's 'retry' instead of 'reload'... :D

eed3si9n:
> It really is a convention. There are the rare outliers, but old-school folks recognize and appreciate it.

I originally wanted to back this claim, but I am falling short on evidence, and I'm now starting to think it might not be as prevalent or it's limited to `[y/N]`.

### POSIX

POSIX standard has [14 point guideline](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html) on command line options. I don't see readline interaction covered there. Let me know if there's a standard that covers this.

### GNU bash mv / rm / cp

As a counterpoint, prominent Linux commands on bash like `mv`, `rm`, and `cp` implement `--interactive` mode.

```bash
$ mv --interactive foo/test.txt bar/test.txt
mv: overwrite 'bar/test.txt'?

$ rm --interactive foo/test.txt
rm: remove regular empty file 'foo/test.txt'?

$ cp --recursive --interactive foo bar
cp: overwrite 'bar/foo/test.txt'?
```

If uppercasing the default was the convention, then we should see `[y/N]` but we don't.

### GNU yes

The very existence of [yes command](http://man7.org/linux/man-pages/man1/yes.1.html) tells us that asking for `y` is a real thing.

### GnuPG

Another GNU example is `gpg`, also known as GnuPG.

```bash
$ gpg --gen-key
Real name: foo
Name must be at least 5 characters long
Real name: foo example
Email address: foo@example.com
You selected this USER-ID:
    "foo example <foo@example.com>"

Change (N)ame, (E)mail, or (O)kay/(Q)uit?
```

This is closer to sbt's use case since it's not a simple yes/no. And gpg is capitalizing all options. Typing in `[Enter]` makes gpg print the same question again.

### Debian apt-get

As mentioned [above](https://github.com/sbt/sbt/pull/4748#issuecomment-534299623) apt-get asks for `[Y/n]`.

```
$ apt-get install something
...
After this operation, 15.8 MB of additional disk space will be used.
Do you want to continue? [Y/n]
```

I see this as more of making an implicit `y/n` tradition into an explicit text, rather than establishing a new convention. (`yum` also prompts similar thing)

### observations

The capitalization-as-convention is on a shaky ground beyond `[Y/n]`.  To follow the style, the placement of the answer options should be after the question mark? So that makes it 

```        
Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? [R/q/l/i]
```

at that point it might be more friendly to say

```        
Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)
```

main/src/main/scala/sbt/Main.scala
ignasi35:
```suggestion
      "Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)"
```

eed3si9n (approved):
I think this looks good

eatkins (approved):
👍

Issue #4746, opened 2019-05-30
Sbt uses too much cpu on idle


eatkins:
In 1.3.0-RC1, sbt is using roughly 10% of a cpu when idling. In older versions, it was more like 2% of a cpu. This is because I reduced a number of timeouts that are causing sbt to frequently wake up to poll queues. These can be reworked to poll from a concurrent blocking queue which should both lower latency and restore the old cpu utilization characteristics.

eed3si9n:
Ref https://github.com/sbt/sbt/pull/4544/commits/fcd24ba7cd7b26c04d750ec70b026e5d5d654356

So should we change it back to `Thread.sleep(50)`? 0.05s should be below human perception and easy enough for machines.

eatkins:
Fixed in #4750

PR #4750, opened 2019-05-30
Consolidate CommandChannel queues  (Reduce idle cpu usage)

eatkins:
I noticed that sbt 1.3.0 was using more cpu when idling (either at the
shell or while waiting for file events) than 1.2.8. This was because I'd
reduced a number of timeouts to 2 milliseconds which was causing a
thread to keep waking up every 2 milliseconds to poll a queue. I thought
that this was cheaper than it actually is and drove the cpu utilization
to O(10%) of a cpu on my mac.

To address this, I consolidated a number of queues into a single queue
in CommandExchange and Continuous. In the CommandExchange case, I
reworked CommandChannel to have a register method that passes in a Queue
of CommandChannels. Whenever it appends an exec, it adds itself to the
queue. CommandExchange can then poll that queue directly and poll the
returned CommandChannel for the actual exec. Since the main thread is
blocking on this queue, it does not need to frequently wake up and can
just poll more or less indefinitely until a message is received. This
also reduces average latency compared to older versions of sbt since
messages will be processed almost as soon as they are received.

The continuous case is slightly more complicated because we are polling
from two sources, stdin and FileEventMonitor. In my ideal world, I'd
have a reactive api for both of those sources and they would just write
events to a shared queue that we could block on. That is nontrivial to
implement, so instead I consolidated the FileEventMonitor instances into
a single FileEventMonitor. Since there is now only one FileEventMonitor
queue, we can block on that queue for 30 milliseconds and the poll
stdin. This reduces cpu utilization to O(2%) on my machine while still
having reasonably low latency for key input events (the latency of file
events should be close to zero since we are usually polling the
FileEventMonitor queue when waiting).

### DIFF
--- a/main-command/src/main/scala/sbt/internal/CommandChannel.scala
+++ b/main-command/src/main/scala/sbt/internal/CommandChannel.scala
@@ -9,6 +9,7 @@ package sbt
 package internal
 
 import java.util.concurrent.ConcurrentLinkedQueue
+
 import sbt.protocol.EventMessage
 import sjsonnew.JsonFormat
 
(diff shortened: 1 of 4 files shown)


eatkins:
Fixes #4746

eed3si9n (dismissed):

eed3si9n:
```
[info] ServerSpec:
[info] server
[info] - should start *** FAILED ***
[info]   jawn.IncompleteParseException: exhausted input
[info]   at jawn.Parser.parse(Parser.scala:362)
[info]   at jawn.SyncParser.parse(SyncParser.scala:24)
[info]   at jawn.SupportParser.$anonfun$parseFromFile$1(SupportParser.scala:21)
[info]   at jawn.SupportParser.parseFromFile(SupportParser.scala:21)
[info]   at jawn.SupportParser.parseFromFile$(SupportParser.scala:20)
[info]   at sjsonnew.support.scalajson.unsafe.Parser$.parseFromFile(Parser.scala:8)
[info]   at sbt.protocol.ClientSocket$.socket(ClientSocket.scala:26)
[info]   at testpkg.TestServer.<init>(ServerSpec.scala:229)
[info]   at testpkg.TestServer$.withTestServer(ServerSpec.scala:164)
[info]   at testpkg.TestServer$.$anonfun$withTestServer$1(ServerSpec.scala:155)
[info]   ...
```

eatkins:
Yeah. I can’t tell what is going on here. The test passes on my mac and my linux vm. I’ll figure it out though.

eed3si9n (approved):

PR #4751, opened 2019-05-30
Don't use global execution context

eatkins:
Because we are sharing the scala library classloader with test and run,
it is possible that sbt will be competing with for resources with the
test and run tasks when trying to get threads from the global execution
context. Also, by using our own execution context, we can shut it down
when sbt exits.

The motivation for this change is that I was looking at the active jvm
threads of an idle sbt process and noticed a bunch of global execution
context threads.

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -9,6 +9,7 @@ package sbt
 
 import java.io.{ File, IOException }
 import java.net.URI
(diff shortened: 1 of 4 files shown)


eatkins:
This is prophylactic. I have seen the global execution context get blocked in the wild so I want to ward off that possibility.

eed3si9n (dismissed):
yaaass..

PR #4752, opened 2019-05-30
Don't require newline for failure options

eatkins:
I can't think of any real reason to force the users to type a newline.
Also the prompt makes it look like just typing the single character is
sufficient. Finally, this behavior is inconsistent with watch, where we
do not require new lines.

Also try to make it clear that retry is the default if the user presses
enter.

### DIFF
--- a/internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
+++ b/internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
@@ -37,6 +37,15 @@ abstract class JLine extends LineReader {
         Option("")
     }
 
+  def readCharacter(prompt: String): Char = {
+    reader.setPrompt(prompt)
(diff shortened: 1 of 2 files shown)


eed3si9n:
How reliable is character reading? For instance would this work on Windows?

eatkins:
It should work (watch does the same thing), but I can double check on my vm.

eed3si9n:
Butterfly Effect coefficient is so high around JLine (see super shell), I don't think we should be messing with console stuff at this late in the stage. This would be a fine first feature for 1.next.

eatkins:
That's fine. But I do somewhat object to the 'R' change in that other PR because it actually just makes things more confusing for me.

jsoref:
Capitalized letters to indicate defaults is a fairly standard thing...

PR #4758, opened 2019-05-31
Restore some legacy watch behavior for play run

eatkins:
I noticed that `~run` in the Play plugin relied on the presence of the
ContinuousEventMonitor key. Rather than completely break that feature, I
re-added the ContinuousEventMonitor attribute to the state in a
continuous build. That being said, the play team does need to update
their plugin because reading from the console no longer works in 1.3.0 so the
user has to Ctrl-C to exit the watch. I think the best way for them to
fix this is to override the `~` command in their plugin and if the input
is 'run', then they do their custom thing, otherwise they delegate to
the default `~` command.

cc @dwijnand

### DIFF
--- a/main/src/main/scala/sbt/internal/Continuous.scala
+++ b/main/src/main/scala/sbt/internal/Continuous.scala
@@ -329,9 +329,11 @@ private[sbt] object Continuous extends DeprecatedContinuous {
     try {
       val stateWithRepo = state.put(globalFileTreeRepository, repo)
       val fullState =
-        if (extracted.get(watchPersistFileStamps))
-          stateWithRepo.put(persistentFileStampCache, fileStampCache)
-        else stateWithRepo
+        addLegacyWatchSetting(
+          if (extracted.get(watchPersistFileStamps))
+            stateWithRepo.put(persistentFileStampCache, fileStampCache)
+          else stateWithRepo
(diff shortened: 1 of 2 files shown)


eed3si9n:
> That being said, the play team does need to update their plugin because reading from the console no longer works in 1.3.0 so the user has to Ctrl-C to exit the watch.

Is this situation specific to Play, or would sbt 1.3.0 break any `~run` on an app that uses stdin?
If so maybe there should be an opt-out setting that Play and others can set?

eatkins:
Yeah. I was having the same thought about having an opt out for reading characters from stdin.

eed3si9n (approved):

eatkins:
I'm planning to fix the stdin (or provide an opt) and add a commit to this PR.

eatkins:
It turns out that on my computer exiting '~run' in a play project by pressing enter doesn't work for any project (I tried 0.13.17, 1.0.4, 1.1.6, 1.2.8 and 1.3.x) in spite of this message:
`(Server started, use Enter to stop and go back to the console...)`
so I am actually not sure that there is any regression about stdin.

dwijnand:
Speaking of watch not working in Play... https://github.com/playframework/playframework/issues/9365

PR #4761, opened 2019-06-01
Document helper functions in BuildStructure

dwijnand:
Also define LoadedBuildUnit#projects & BuildUnit#thisRootProject, &
cleanup AttributeKey & BasicAttributeMap


eed3si9n (changes requested):
[warn] /home/travis/build/sbt/sbt/main/src/main/scala/sbt/internal/BuildStructure.scala isn't formatted properly!
[error] 1 files must be formatted

PR #4764, opened 2019-06-02
Make some LSP-related methods public

smarter:
This is the minimum amount of stuff that needs to be public to teach new
messages to the server from a plugin.

### DIFF
--- a/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
@@ -26,7 +26,7 @@ import scala.concurrent.ExecutionContext
 private[sbt] final case class LangServerError(code: Long, message: String)
     extends Throwable(message)
 
-private[sbt] object LanguageServerProtocol {
+object LanguageServerProtocol {
   lazy val internalJsonProtocol = new InitializeOptionFormats with sjsonnew.BasicJsonProtocol {}
 
   lazy val serverCapabilities: ServerCapabilities = {
@@ -102,7 +102,7 @@ private[sbt] object LanguageServerProtocol {
 }
 
 /** Implements Language Server Protocol <https://github.com/Microsoft/language-server-protocol>. */
-private[sbt] trait LanguageServerProtocol extends CommandChannel { self =>
+trait LanguageServerProtocol extends CommandChannel { self =>
 
   lazy val internalJsonProtocol = new InitializeOptionFormats with sjsonnew.BasicJsonProtocol {}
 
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
```diff
@@ -26,7 +26,7 @@ import scala.concurrent.ExecutionContext
 private[sbt] final case class LangServerError(code: Long, message: String)
     extends Throwable(message)
 
-private[sbt] object LanguageServerProtocol {
+object LanguageServerProtocol {
```
eed3si9n, line 29: There's no bincompat guarantees for internal packages.

main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
```diff
@@ -102,7 +102,7 @@ private[sbt] object LanguageServerProtocol {
 }
 
 /** Implements Language Server Protocol <https://github.com/Microsoft/language-server-protocol>. */
-private[sbt] trait LanguageServerProtocol extends CommandChannel { self =>
+trait LanguageServerProtocol extends CommandChannel { self =>
```
eed3si9n, line 105: I think we should avoid traits with implementations in API.

smarter: The important part is just to have access to jsonRpcRespond\*/jsonRpcNotify somehow, this could be achieved by having a trait with abstract defs somewhere.

eed3si9n: Here's an example code - https://github.com/sbt/sbt/blob/v1.2.8/sbt/src/server-test/handshake/build.sbt

smarter: IIRC, that was not good enough. I can access `callback` here but I don't have access to the sbt state, so I can't do anything interesting, the only solution I found to that problem when I wrote my prototype bsp plugin a year ago was to indirectly call a command, and have the command call `jsonRpcRespond`, but I can't pass `callback` to that command: https://github.com/dotty-staging/dotty/blob/5eb5bbc1e0fad08647a50690f03ca28131637a6b/sbt-dotty/src/dotty/tools/sbtplugin/DottyBSPPlugin.scala#L193-L197

eed3si9n: Interesting. I'd have to think a bit more on how can a command or a task send correct RPC response.

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/5207

PR #4765, opened 2019-06-02
Watch docs

eatkins:
I made these changes in response to writing documentation for the new watch apis.


main/src/main/scala/sbt/nio/Keys.scala
```diff
@@ -82,16 +82,13 @@ object Keys {
...
-    "Function to run upon exit of a continuous build. It can be used to cleanup resources used during the watch."
-  ).withRank(DSetting)
+  val watchLogLevel =
+    settingKey[sbt.util.Level.Value]("Transform the default logger in continuous builds.")
+      .withRank(DSetting)
   val watchOnFileInputEvent = settingKey[(Int, Watch.Event) => Watch.Action](
     "Callback to invoke if an event is triggered in a continuous build by one of the files matching an fileInput glob for the task and its transitive dependencies"
   ).withRank(DSetting)
-  val watchOnIteration = settingKey[Int => Watch.Action](
+  val watchOnIteration = settingKey[(Int, String, Seq[String]) => Watch.Action](
```
eed3si9n: I didn't notice String in other keys before. Could this be typed to something like ProjectRef?

eed3si9n (changes requested):

eed3si9n (approved):

PR #4769, opened 2019-06-03
Java reflection

### LINKED ISSUE #4760: Class.forName can't find classes in sbt-1.3.0-RC1
## steps

I'm still trying to find a reproduction, but basically, a problem happens when using ~Spark 2.4.x with~ Scala 2.11, sbt-1.3.0-RC1.

## problem

- Call a method in a dependent library that uses `Class.forName("(class name)")` -> causes java.lang.ClassNotFoundException
- Call `Class.forName("(class name)")` works without any problem.

I guess some method (e.g., `ClassLoader.getClassLoader(caller)`) used in Class.forName has an issue in sbt-1.3.0-RC1's layered class loader:
```java
   @CallerSensitive
    public static Class<?> forName(String className)
                throws ClassNotFoundException {
        Class<?> caller = Reflection.getCallerClass();
        return forName0(className, true, ClassLoader.getClassLoader(caller), caller);
    }
```
## expectation
- There has been no problem until sbt 1.2.8

## notes

- It seems a workaround is using `Class.forName(name, true, Thread.currentThread().getContextClassLoader())` instead of `Class.forName(name)`
- Spark is unrelated (fixed the description)

sbt version: *1.3.0-RC1*

eatkins:
This pr makes the AllLibraryJars layering strategy work in many cases where java reflection is used in the dependency layer, even when the java reflection needs to load classes in a layer that is below the dependency layer in the class loader hierarchy.

 I added a scripted test for spark which actually exposed both that java reflection didn't work with the AllLibraryJars strategy as well as an issue where the classpath filter would reject resources loaded in the dependency layer.

Fixes #4760 (at least in the projects I tested, which include https://github.com/circe/circe, https://github.com/ThoughtWorksInc/dsl.scala and the sample app at https://spark.apache.org/docs/latest/quick-start.html)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -855,9 +855,12 @@ object Defaults extends BuildCommon {
       }
       val trl = (testResultLogger in (Test, test)).value
       val taskName = Project.showContextKey(state.value).show(resolvedScoped.value)
+      val currentLoader = Thread.currentThread.getContextClassLoader
       try {
+        Thread.currentThread.setContextClassLoader(testLoader.value)
```
eed3si9n, line 860: Are we effectively using context classloader as a thread-local global variable here?

eed3si9n: In other words, is there anything magical about `Thread.currentThread.getContextClassLoader` or could you potentially use any thread-safe global var to pass in `testLoader.value` into `loadClass(...)`?

main-command/src/main/java/sbt/internal/classpath/WrappedLoader.java
```diff
@@ -0,0 +1,34 @@
...
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.internal.classpath;
+
+import java.net.URL;
+import java.net.URLClassLoader;
+
+public class WrappedLoader extends URLClassLoader {
```
eed3si9n, line 13: great.
Could you link to https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html in the commit message as the explanation of "parallel capable class loaders"?

main/src/main/scala/sbt/Defaults.scala
eatkins: There is something magical about `Thread.getContextClassLoader`. If new threads are spawned by the thread, they inherit the context class loader of the parent. This means that if the test spawns background threads to do work, then the same context class loader is available on all of the background threads. When we set the context class loader to the test loader and then invoke the test runner, the tests run in an environment that is sandbox-like. It's analogous to a forked jvm process with the test loader as the app classloader (though I'm not sure how exactly this analogy holds).

If we used a thread local variable, it wouldn't get copied to threads spawned by the current thread so class loading would likely break somewhere.

FWIW, we already swapped out the context class loader in ScalaRun. I'm not sure why it was done there but not for tests.

main-command/src/main/java/sbt/internal/classpath/WrappedLoader.java
eatkins: Sure thing.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: > If new threads are spawned by the thread, they inherit the context class loader of the parent. 

cool. Thanks for the explanation. Maybe we can add the above as comment.

main-command/src/main/java/sbt/internal/classpath/WrappedLoader.java
eatkins: I improved the commit message in https://github.com/sbt/sbt/pull/4769/commits/625470cdd5e219ff0e1436c6d157c03e41c935cb

main/src/main/scala/sbt/Defaults.scala
eatkins: I added a lengthy comment in https://github.com/sbt/sbt/pull/4769/commits/625470cdd5e219ff0e1436c6d157c03e41c935cb

eed3si9n (approved):
Thanks for this!

PR #4779, opened 2019-06-06
Fix logger class loader leak with workaround

### LINKED ISSUE #4773: sbt leaks metaspace when reloading an empty project
## steps
sbt version: from early 1.0.* up to and including 1.3.0-RC1

$ mkdir /tmp/empty-project
$ cd /tmp/empty-project
$ sbt -no-global -Dsbt.version=1.3.0-RC1 -J-Xmx200m -J-Xms200m -J-XX:+ExitOnOutOfMemoryError
...
> reload 
200x times

## problem
The jvm aborts with out of memory error

## expectation
The JVM always returns to the same memory usage it had after starting the first time.

## notes
A script to run sbt can be found here: https://gist.github.com/jm-g/c91823d8c4c4045728f63a8d2ba128cd

I have also a fix which I will publish soon. I am creating this issue for discussions.

jm-g:
fix memory leak when reloading a project

When executing the sbt command reload, some log Appenders stay
registered in the global LogManager, which leads to leaked class
loaders , which leads eventually to an OutOfMemoryException: metaspace.

This fix registers a cleanup method, which clears the global
LogManager's appenders and logs on debug, which Appender it cleared.

The correct fix would imho be a proper lifecycle mechanism that allows
to deregister all appenders where appropriate.

The underlying bug can be reproduced when you start sbt in an empty
directory and observe its metaspace usage.

Fixes #4773

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -13,6 +13,9 @@ import java.util.concurrent.ForkJoinPool
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.{ Locale, Properties }
 
+import org.apache.logging.log4j.{ LogManager => XLogManager }
+
+import org.apache.logging.log4j.core.LoggerContext
 import sbt.BasicCommandStrings.{ Shell, TemplateCommand }
 import sbt.Project.LoadAction
 import sbt.compiler.EvalImports
@@ -31,6 +34,8 @@ import xsbti.compile.CompilerCache
 import scala.annotation.tailrec
 import scala.concurrent.ExecutionContext
 import scala.util.control.NonFatal
+
+import scala.collection.JavaConverters._
 
 /** This class is the entry point for sbt. */
 final class xMain extends xsbti.AppMain {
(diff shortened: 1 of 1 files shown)


eatkins:
It looks like scripted project/sbt-version-change fails after this change. I took a quick look and I'm wondering if maybe the lifecycle issues that you mentioned could be handled in MainLoop.runWithNewLog. It looks like that method adds an appender to the global logger without removing the previous appender. I'm wondering if that is the main source of the leaks.

eatkins:
Oh, and thanks for working on this!

eed3si9n:
@jm-g Thanks for looking into this.
I'm not sure if Main.scala is the right layer to fix tho. If some appender is leaking, then doesn't that imply that they are likely leaking for any command or a task? Or is it leaking specifically for reload?
Also I am not if wiping out all the appenders is safe to do.

eatkins:
The scripted failure looks real so this change as written does seem to break something. I don’t understand appenders well so I wouldn’t be surprised if my suggestion was off. I did notice that new appenders get created after each command evaluation and I wasn’t clear on why that was.

jm-g:
@eed3si9n and @eatkins As far as I understood it, wiping all appenders is safe after a reload, because they are either re-added again or they are instantiations of classes from past class loaders. Which is the reason for the leakage of class loaders. As you can see from my log statements, I am not sure whether this is the best way to fix it. So, if you guys aren't sure how these loggers and appenders are supposed to be working, I'll have to dig deaper myself. Is there some documentation about the logging subsystem?

eed3si9n:
Closing this since https://github.com/sbt/sbt/pull/5731 was merged.
We ultimately didn't take this approach, but this PR became a jumping off point for the later works, so I want to thank you for the contribution.

jm-g:
Thanks, for continuing  this work. @eatkins @eed3si9n

PR #4780, opened 2019-06-07
Java reflection v2

eatkins:


main/src/main/scala/sbt/internal/LayeredClassLoaders.scala
```diff
@@ -10,119 +10,253 @@ package sbt.internal
...
+ * If the top layer needs to load a class from the bottom layer via java reflection, we facilitate
+ * that with the `ReverseLookupClassLoader`.
+ *
+ *
+ * This holder caches the ReverseLookupClassLoader, which is the top loader in this hierarchy. The
+ * checkout method will get the RevereLookupClassLoader from the cache or make a new one if
+ * none is available. It will only cache at most one so if multiple concurrently tasks have the
+ * same dependency classpath, multiple instances of ReverseLookupClassLoader will be created for
+ * the classpath. If the ReverseLookupClassLoader makes a lookup in the BottomClassLoader, it
+ * invalidates itself and will not be cached when it is returned.
```
eed3si9n, line 49: I am a bit confused here.
Let say there's subproject A, B, and C with identical libraries using Java reflection, and you run all three concurrently:
- checkout A
- checkout B
- checkout C
- loadClass A
- loadClass B
- loadClass C
- checkin A
- checkin B
- checkin C

Doesn't the first loadClass set it dirty, but B won't know?

eatkins: The cache has size one because I didn’t think it was worth it to keep more than one instance hot. In the sequence described above, project A gets the cached loader first, then a new loader is created for each of B and C. t. Let’s say only project A dirties the loader. When it checks in its loader, it gets closed and discarded because it was marked dirty. Then when B checks in its loader, that becomes the cached instance. When C checks in its loader, it gets closed and discarded because the cache was already filled by B.

eed3si9n: <s>Because sbt runs tests in parallel, this could probably happen with one subproject too.</s> GitHub didn't display the above reply when I posted it.

eed3si9n: > The cache has size one because I didn’t think it was worth it to keep more than one instance hot.

ok. That makes sense.

eed3si9n (approved):

main/src/main/scala/sbt/internal/LayeredClassLoaders.scala
eatkins: Yeah. I figured that in that case this is no worse than having ScalaLibrary. The main issue is that ClassLoaderCache as presently written doesn’t support having multiple loaders for the same classpath. That is essential for the dependency layer with java reflection or else in the example above, the loader for subproject a could load classes for subproject b. This can get really complicated if one tries to design the optimal solution. I think the current implementation is a reasonable compromise without adding additional strategies (for example there could be AllLibraryJars and AllLibraryJarsWithReflection). The other thing I thought about trying to sneak in was an exclude filter for the jars in the dependency layer. I am not sure what the right api is though so it felt more like a 1.4.0 feature.

eed3si9n: So B's library singleton could be reused by A too, which could be tricky if the state is dependent on some external factor like application.conf.

eatkins: Oh, so in that case projects a and b would not share a dependency layer because the resource loader goes between scala library/reflect and the dependency layer. Projects a and b would both have to have empty resource paths for that loader to be shared between them so if application.conf is different, they would be completely isolated.

eatkins: I forgot about that but I added the resource layer specifically for this reason.

PR #4781, opened 2019-06-07
Improve reload warnings

eatkins:
The previous reload warnings looked wrong with out a leading capital
letter. The labels also weren't as descriptive as they could have been.

### DIFF
--- a/main/src/main/scala/sbt/internal/nio/CheckBuildSources.scala
+++ b/main/src/main/scala/sbt/internal/nio/CheckBuildSources.scala
@@ -25,22 +25,24 @@ private[sbt] object CheckBuildSources {
         logger.debug("Checking for meta build source updates")
         (changedInputFiles in checkBuildSources).value match {
           case Some(cf: ChangedFiles) if !firstTime =>
-            val rawPrefix = s"Meta build source files have changed:\n" +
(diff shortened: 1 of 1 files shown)


main/src/main/scala/sbt/internal/nio/CheckBuildSources.scala
```diff
@@ -26,11 +26,13 @@ private[sbt] object CheckBuildSources {
         (changedInputFiles in checkBuildSources).value match {
           case Some(cf: ChangedFiles) if !firstTime =>
             val rawPrefix = s"Meta build source files have changed:\n" +
-              (if (cf.created.nonEmpty) s"creations: ${cf.created.mkString("\n  ", "  \n", "\n")}"
+              (if (cf.created.nonEmpty) s"New files: ${cf.created.mkString("\n  ", "  \n", "\n")}"
```
eed3si9n: I worked on a bunch of compiler error messages for Scala 2.13, and I've unified them all to start with lower case. Classic compiler errors like "cannot find symbol" are set in lowercase. I've been thinking we should start doing that for sbt as well for it's logs.

eatkins: Ok. I care more about "new files" vs. "creations". If I drop the upper case here, then I should drop it on the rawPrefix part too I think.

eed3si9n (approved):

PR #4783, opened 2019-06-08
turbo mode

eed3si9n:
First this centralizes most of the user-facing system properties.

Next, this creates "turbo mode" that enables experimental or advanced features that might require some debugging by the build user when it doesn't work.

Initially we are putting the layered ClassLoader (`ClassLoaderLayeringStrategy.AllLibraryJars`) behind this flag.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -164,7 +164,6 @@ object Defaults extends BuildCommon {
   private[sbt] lazy val globalJvmCore: Seq[Setting[_]] =
     Seq(
       compilerCache := state.value get Keys.stateCompilerCache getOrElse CompilerCache.fresh,
-      classLoaderLayeringStrategy :== ClassLoaderLayeringStrategy.AllLibraryJars,
       sourcesInBase :== true,
       autoAPIMappings := false,
       apiMappings := Map.empty,
@@ -285,6 +284,7 @@ object Defaults extends BuildCommon {
         val tempDirectory = taskTemporaryDirectory.value
         () => Clean.deleteContents(tempDirectory, _ => false)
       },
+      turbo :== SysProp.turbo,
       useSuperShell := { if (insideCI.value) false else SysProp.supershell },
       progressReports := {
         val progress = useSuperShell.value
@@ -1049,7 +1049,7 @@ object Defaults extends BuildCommon {
       cp,
       forkedParallelExecution = false,
       javaOptions = Nil,
-      strategy = ClassLoaderLayeringStrategy.AllLibraryJars,
+      strategy = ClassLoaderLayeringStrategy.ScalaLibrary,
       projectId = "",
     )
   }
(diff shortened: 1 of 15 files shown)


eed3si9n:
Hopefully https://github.com/sbt/sbt/pull/4784 fixes the Travis failure.

eatkins:
I really like this change. My one suggestion is that it would be great if we had a command that would display all of the available system properties. We could add that command to the commands shown in `help` to improve discoverability. We can punt til 1.4.0 for that though.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -164,7 +164,10 @@ object Defaults extends BuildCommon {
   private[sbt] lazy val globalJvmCore: Seq[Setting[_]] =
     Seq(
       compilerCache := state.value get Keys.stateCompilerCache getOrElse CompilerCache.fresh,
-      classLoaderLayeringStrategy :== ClassLoaderLayeringStrategy.AllLibraryJars,
+      classLoaderLayeringStrategy := {
+        if (turbo.value) ClassLoaderLayeringStrategy.AllLibraryJars
```
eatkins: Maybe this should be (ThisBuild / turbo).value? I was experimenting with this and noticed that if I ran  `ThisBuild / turbo := true` at the command line that it had no effect.

eatkins: Actually, after more experimentation, I think maybe the best option is to move this into coreDefaultSettings. That allows the user to set turbo at the Global, ThisBuild and project level and have it work as expected.

eed3si9n: Doh.

eed3si9n: > That allows the user to set turbo at the Global, ThisBuild and project level and have it work as expected.

This makes sense.

eed3si9n: 6878fb6cdbbdf038b97120d7242f782698df6763

eatkins (approved):
lgtm

main/src/main/scala/sbt/internal/SysProp.scala
```diff
@@ -0,0 +1,123 @@
...
+   * 1. -Dsbt.log.noformat=true means no formatting.
+   * 2. -Dsbt.color=always/auto/never/true/false
+   * 3. -Dsbt.colour=always/auto/never/true/false
+   * 4. -Dsbt.log.format=always/auto/never/true/false
+   */
+  lazy val color: Boolean = ConsoleAppender.formatEnabledInEnv
+
+  def supershell: Boolean = color && getOrTrue("sbt.supershell")
+
+  def supersheelSleep: Long = long("sbt.supershell.sleep", 100L)
```
rethab, line 87: I think there's one `e` too many. Should be: supershellSleep

eed3si9n: Should've been supersheepSleep. 🐑🐑

eed3si9n: Looks like it's fixed in https://github.com/sbt/sbt/pull/4786/commits/a6bc7b1c7664c83a788878a6a76dd3b67e995d53

PR #4784, opened 2019-06-08
Wait for akka termination in test

eatkins:
We now correctly close the test classloader which can cause this
scripted test to fail if the classloader is closed before the actor
system finishes terminating.


eed3si9n (approved):

eed3si9n:
This might happen in the wild, so I am going to open an issue for this retroactively as a potential faq.

PR #4786, opened 2019-06-08
Close classloaders

eatkins:


eatkins:
This is in response to #4784.

main/src/main/scala/sbt/internal/SysProp.scala
```diff
@@ -82,9 +82,11 @@ object SysProp {
    */
   lazy val color: Boolean = ConsoleAppender.formatEnabledInEnv
 
+  def closeClassLoaders: Boolean = getOrTrue("sbt.close.classloaders")
```
eed3si9n: I'd make this `sbt.classloader.close`.

eed3si9n (changes requested):

main/src/main/scala/sbt/internal/SysProp.scala
eed3si9n: https://github.com/sbt/sbt/pull/4787

eatkins: Done.

eed3si9n (approved):

PR #4792, opened 2019-06-09
Remove err.printStackTrace in MainLoop

eatkins:
I added this for debugging and did not mean to leave it in. It causes
massive walls of text to be printed sometimes when compilation fails.


eed3si9n (approved):

eed3si9n:
I don't think this is covered by scripted, so I am going to merge this without awaiting the completion.

PR #4806, opened 2019-06-13
Don't use Set[Incomplete]

eatkins:
It's very expensive to compute the hash code of a deeply nested
Incomplete. To prevent a loop, we only want to check for object equality
anyway, which System.identityHashCode gives us (to reasonable
approximation).

### DIFF
--- a/main/src/main/scala/sbt/internal/Aggregation.scala
+++ b/main/src/main/scala/sbt/internal/Aggregation.scala
@@ -9,6 +9,7 @@ package sbt
 package internal
 
 import java.text.DateFormat
+import java.util.{ Collections, IdentityHashMap }
 
 import Def.ScopedKey
 import Keys.{ showSuccess, showTiming, timingFormat }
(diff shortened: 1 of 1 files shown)


nafg:
Perhaps it would be more elegant to use `java.util.IdentityHashMap`

nafg:
Something like `java.util.Collections.newSetFromMap(new java.util.IdentityHashMap).asScala` using JavaConverters

eatkins:
That's a good idea.

main/src/main/scala/sbt/internal/Aggregation.scala
```diff
@@ -114,19 +115,18 @@ object Aggregation {
   )(implicit display: Show[ScopedKey[_]]): State = {
     val complete = timedRun[T](s, ts, extra)
     showRun(complete, show)
-    @tailrec def findReload(
-        incomplete: Incomplete,
-        remaining: List[Incomplete],
-        visited: Set[Incomplete]
-    ): Boolean = {
+    lazy val visited = Collections
```
eed3si9n, line 123: I think it should be noted here in comments that we tried using Set and it caused performance issue due to deeply nested case class doing a fairly straightforward operation.

eed3si9n (changes requested):

eed3si9n (approved):

PR #4809, opened 2019-06-13
Move multi parser up

eatkins:
I noticed that in 1.3.0-RC2, the following commands worked:
  ++2.11.12 ;test
  ;++2.11.12;test
but
  ++2.11.12; test
did not. This was because the cross command ran before the multi command
parser and did not handle the case where there was a ';' immediately
following the version. I was able to fix this by moving the multi
command parser to the top of the command lists so that it is evaluated
before the cross command parser.


eed3si9n (approved):

eatkins:
Superseded by #4811.

Issue #4808, opened 2019-06-13
[1.3.0-RC2] Regression: Truncating of plugin parameters which contain the semicolon character


plokhotnyuk:
The following command that worked with sbt 1.2.8 successfully now starts benchmark with truncated parameter list:

```
sbt -java-home /usr/lib/jvm/jdk-11 -no-colors 'jsoniter-scala-benchmark/jmh:run -jvmArgsAppend "-XX:+UnlockDiagnosticVMOptions -XX:+DebugNonSafepoints" -prof jmh.extras.Async:event=cpu;dir=/tmp/profile-async;asyncProfilerDir=/home/andriy/Projects/com/github/jvm-profiling-tools/async-profiler;flameGraphDir=/home/andriy/Projects/com/github/brendangregg/FlameGraph;flameGraphOpts=--color,java;verbose=true -wi 5 -i 25 -f 1 -p size=128 ArrayOfFloatsReading.jsoniterScala'
```

Output:
```
[info] Running (fork) org.openjdk.jmh.Main -jvmArgsAppend -XX:+UnlockDiagnosticVMOptions -XX:+DebugNonSafepoints -prof jmh.extras.Async:event=cpu
[error] WARNING: An illegal reflective access operation has occurred
[error] WARNING: Illegal reflective access by org.openjdk.jmh.util.Utils (file:/tmp/sbt_404ca7c3/target/44244710/jmh-core-1.21.jar) to field java.io.PrintStream.charOut
[error] WARNING: Please consider reporting this to the maintainers of org.openjdk.jmh.util.Utils
[error] WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
[error] WARNING: All illegal access operations will be denied in a future release
[info] Location of async-profiler-dir must be set with environment variable ASYNC_PROFILER_DIR or corresponding profiler option
[error] Profilers failed to initialize, exiting.
[error] Nonzero exit code returned from runner: 1
[error] (jsoniter-scala-benchmark / Jmh / run) Nonzero exit code returned from runner: 1
[error] Total time: 100 s (01:40), completed Jun 13, 2019, 6:21:22 PM
```

W/A is to wrap the affected parameter by double quotation marks:

```
[...]

PR #4811, opened 2019-06-14
Strict multi parser

eatkins:
Validate commands in multiparser
It was reported in #4808 that compared
to 1.2.8, sbt 1.3.0-RC2 will truncate the command args of an input task
that contains semicolons. This is actually intentional, but not
completely robust. For sbt >= 1.3.0, we are making ';' syntactically
meaningful. This means that it always represents a command separator
_unless_ it is inside of a quoted string. To enforce this, the multi parser
will effectively split the input on ';', it will then validate that each
command that it extracted is valid. If not, it throws an exception. If
the input is not a multi command, then parsing fails with a normal
failure.

I removed the multi command from the state's defined commands and reworked
State.combinedParser to explicitly first try multi parsing and fall back
to the regular combined parser if it is a regular command. If the multi
parser throws an uncaught exception, parsing fails even if the regular
parser could have successfully parsed the command. The reason is so that
we do not ever allow the user to evaluate, say 'run a;b'. Otherwise the
behavior would be inconsitent when the user runs 'compile; run a;b'


eatkins:
It's harder than expected to make ';' have consistent meaning because ';' can appear in code blocks in set, e.g. `set foo := { val x = 1; x + 1 }`.

main-command/src/main/scala/sbt/State.scala
```diff
@@ -41,7 +41,13 @@ final case class State(
     currentCommand: Option[Exec],
     next: State.Next
 ) extends Identity {
-  lazy val combinedParser = Command.combine(definedCommands)(this)
+  private[this] val earlyCommands = Set("~", "alias")
```
eed3si9n: Let's change to `priorityCommand`.

eed3si9n: Add comments somewhere about the precedence rules.

eed3si9n (approved):

PR #4817, opened 2019-06-17
Use sbt-boilerplate for AList, TupleSyntax & Scoped

dwijnand:


eed3si9n:
If/when sbt 2.0.0-M1 comes out, this would put sbt-boilderplate on the critical path for sbt/sbt to dogfood it. I wonder if the resultant Scala source code should also be committed (like we do with Contraband) as a workaround.

dwijnand:
Defer to dogfood-bootstrapping time?

dwijnand:
Pushing past 1.3.0.

project/MiMaSettings.scala
```diff
@@ -0,0 +1,18 @@
...
+import sbt._
+import com.typesafe.tools.mima.core._
+import com.typesafe.tools.mima.plugin.MimaPlugin.autoImport._
+
+object MimaSettings {
+  val mimaSettings = Def.settings (
+    mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[IncompatibleResultTypeProblem]("sbt.TupleSyntax.t*ToApp*"),
+      ProblemFilters.exclude[IncompatibleResultTypeProblem]("sbt.TupleSyntax.t*ToTable*"),
+      ProblemFilters.exclude[MissingClassProblem]("sbt.Scoped$Apply*"),
```
eed3si9n, line 12: Don't we need to care about these?

dwijnand: If I did it I'm 90% sure I did it because it's safe.

But with the distance of time I couldn't tell you what that reason is.  I'll go back and find out.

dwijnand: I can't quite tell if removing these would be unsafe, and I also can't find a way to change this PR so it doesn't introduce any bincompat changes.  So I'll just leave it be.

PR #4818, opened 2019-06-17
Implement JsonFormat[FiniteDuration] for keys

dwijnand:
Also update sbt/scala versions and switch from "t" to "bippy" for "grepability", in SettingQueryTest.

### DIFF
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -17,6 +17,7 @@ import sbt.BuildSyntax._
 import sbt.Def.ScopedKey
 import sbt.KeyRanks._
 import sbt.internal._
+import sbt.internal.FiniteDurationCodec._
 import sbt.internal.inc.ScalaInstance
 import sbt.internal.io.WatchState
 import sbt.internal.librarymanagement.{ CompatibilityWarningOptions, IvySbt }
--- a/main/src/main/scala/sbt/internal/FiniteDurationCodec.scala
+++ b/main/src/main/scala/sbt/internal/FiniteDurationCodec.scala
@@ -1,0 +1,36 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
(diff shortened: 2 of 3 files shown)


main/src/main/scala/sbt/Keys.scala
```diff
@@ -17,6 +17,7 @@ import sbt.BuildSyntax._
 import sbt.Def.ScopedKey
 import sbt.KeyRanks._
 import sbt.internal._
+import sbt.internal.FiniteDurationCodec._
```
eed3si9n, line 20: Could you refresh my memory on where this typeclass instance is being captured?

dwijnand: The expansion of the `taskKey` macro passes it to the apply in `TaskKey`, if memory serves.

dwijnand: If you want to go the long route you could put this in sjsonnew's BasicProtocol and cut a release.

dwijnand:
Let's push this one to 1.4.0 too, then.

main/src/main/scala/sbt/Keys.scala
eed3si9n: I thought changing BasicJsonProtocol breaks the bincompat because it's a trait.

eed3si9n: Even if it's not part of BasicJsonProtocol, maybe it does make sense to put these format in sjson-new though.

dwijnand: Adding to traits is bincompat in Scala 2.12+.  Are you concerned about sjson-new for Scala 2.11?

I'll PR sjson-new and we can have a look.

eed3si9n: In most cases, but not all. Weren't there a plugin we broke early in sbt 1.x because we added something to BasicJsonProtocol, and someone was extending it?

dwijnand: Not that I'm aware of, or I've since forgotten.

dwijnand:
I don't feel like going through the merry-go-round of PRing this in sjson-new and then needing a release of sjson-new to progress.

PR #4819, opened 2019-06-17
Cleanup Load.loadTransitive

dwijnand:
The largest win is creating the helper, inner, "load" method.


eed3si9n:
I'd postpone this to 1.4.0.

dwijnand:
Works for me.

PR #4830, opened 2019-06-23
use 'q' to quit watch mode

eed3si9n:
Ref #4512

I think we discussed it at some point because `less` quits using `q`.

### DIFF
--- a/main/src/main/scala/sbt/nio/Watch.scala
+++ b/main/src/main/scala/sbt/nio/Watch.scala
@@ -381,15 +381,15 @@ object Watch {
 
   /**
    * Converts user input to an Action with the following rules:
-   * 1) 'x' or 'X' will exit sbt
+   * 1) 'q' or 'Q' will exit sbt
    * 2) 'r' or 'R' will trigger a build
    * 3) new line characters cancel the watch and return to the shell
    */
   final val defaultInputParser: Parser[Action] = {
-    val exitParser: Parser[Action] = chars("xX") ^^^ new Run("exit")
+    val quitParser: Parser[Action] = chars("qQ") ^^^ new Run("exit")
     val rebuildParser: Parser[Action] = chars("rR") ^^^ Trigger
     val cancelParser: Parser[Action] = chars(legal = "\n\r") ^^^ new Run("iflast shell")
-    exitParser | rebuildParser | cancelParser
+    quitParser | rebuildParser | cancelParser
   }
 
   private[this] val options = {
@@ -397,7 +397,7 @@ object Watch {
     val opts = Seq(
       s"$enter: return to the shell",
       s"'r': repeat the current command",
-      s"'x': exit sbt"
+      s"'q': quit sbt"
     )
     s"Options:\n${opts.mkString("  ", "\n  ", "")}"
   }


eatkins:
I'm not so sure that there is solid consensus on this: https://ux.stackexchange.com/questions/50893/do-we-exit-quit-or-close-an-application
It's been 'x' since September and I've gotten pretty used to that. What is the motivation for the change besides the fact that less uses 'q'?

eatkins:
Oh, I just remembered why I made it 'x' and not 'q':
```
sbt:watch-perf> quit
[error] Not a valid command: quit (similar: exit, about)
[error] Expected ';'
[error] Not a valid project ID: quit
[error] Expected ':'
[error] Not a valid key: quit
[error] quit
[error]     ^
sbt:watch-perf> exit
[info] shutting down server
```

dwijnand:
I've used the shortcut once, and that was just to test it worked.  What's the motivation for having it?  I ask because it's yet another way to do things.  What's wrong with [Enter] and then Ctrl-D/`exit`?

eatkins:
There were two reasons I added the shortcut:
1) In older versions of sbt, if you started sbt with, say `sbt ~test`, then <enter> would directly exit the program. I changed that behavior so that [Enter] always returns to the shell and wanted to support the old use case of directly exiting from `~`.
2) What's wrong to me with [Enter] + exit[Enter] is that I have to type 5 extra characters which increases friction for me. I am very sensitive to that kind of thing which is why I made all the watch improvements in the first place. This is also why I added the repeat option. You also could exit watch, hit up arrow and then [Enter] (assuming your command history wasn't updated, which it sometimes is by an external process) and that process was driving me crazy.

dwijnand:
> 1\. I changed that behavior so that [Enter] always returns to the shell

TIL.

> 2\. What's wrong to me with [Enter] + exit[Enter] is that I have to type 5 extra characters which increases friction for me. I am very sensitive to that kind of thing which is why I made all the watch improvements in the first place.

Just out of interest, what drives this being a common occurrence?  Are you changing Swoval and/or sbt and re-running and manually testing?  Do you have a way to change the code and spin up an sbt instance to manually test?

Here's another thought: we could also have this not be printed out to the console.

eatkins:
It is not uncommon to have flaky tests when you are testing code with concurrency. When that is the case, I may want to re-run the test without recompiling.

eatkins:
Also, if the watch sources are not fully identified, changing a source file might not trigger a build. For example, in sbt, the files in the scripted test directories used to not be watched so I’d need to manually trigger a build if I changed a scripted test file in ~scripted. I did add the scripted test files to the fileInputs in the sbt build, but I’m sure there are many projects/plugins where dependent source files are not tracked by sbt.

eatkins:
Wrt what is printed to the console, that is configurable with the watchStartMessage key (documentation for this setting is merged into the beta branch of sbt website). I think it makes sense to print the options by default, but if a user is annoyed by them, they are free to set watchStartMessage in their ~/.sbt/1.0 config. Also, the shortcuts are configurable via the watchInputParser setting, so the user can change or remove any shortcuts: https://github.com/sbt/website/blob/1.x-beta/src/reference/02-DetailTopics/01-Using-sbt/05-Triggered-Execution.md

eed3si9n:
Historically Predef use to have `exit` method until it became sys.exit in 2.9, and removed in 2.11 (https://github.com/scala/scala/commit/c26a8db067e4f04ef959bb9a8402fa3e931c3cd7).

Where as `:q` or `:quit` works for Scala REPL (https://github.com/sbt/sbt/issues/1374). https://github.com/sbt/sbt-vimquit

Also when sbt fails to load, it's `q` to quit.

```
```

That's two points for quit camp even without the `ed`/`less`/`vi` command tradition.

eatkins:
@eed3si9n I don't prefer `exit` to `quit`, but I do prefer consistency. Are we putting replacing the `exit` command with `quit` on the table? And, to be annoying, I usually exit vim with `:x` (save and quit).

eed3si9n:
Yes. I'd love to have `:q` and `quit` as new `exit`.

dwijnand:
I quit a bash shell with Ctrl-D, I quit sbt shell with Ctrl-D, I quit the scala REPL with Ctrl-D.  (I also quit `vim` with `:x`, or `:cq` if I want it to exit with non-zero result code.)

Could we make Ctrl-D work like [Enter], meaning that you can exit "all the way" with Ctrl-D, Ctrl-D?

sarahgerweck:
It seems like this is mostly an attempt to patch the fallout of removing the ability to have non-interactive watch sessions. This seems like it has the potential to create a _lot_ of fallout when 1.3 comes out, especially since it hasn't been called out in the RC description or notes (at least as far as I noticed).

My team uses watch mode pretty heavily, and the new behavior seems pretty strange to me. I'm already anticipating the complaints if we upgrade to SBT 1.3 and lose the current mechanism for exiting the application.

eed3si9n:
@sarahgerweck 

> It seems like this is mostly an attempt to patch the fallout of removing the ability to have non-interactive watch sessions.

As far as I know `sbt "~test"` still works, so the ability to run watch session is't removed. What changed is the behavior mapped to `Enter`.

> This seems like it has the potential to create a _lot_ of fallout when 1.3 comes out, especially since it hasn't been called out in the RC description or notes (at least as far as I noticed).

The omission from the release note was an oversight on our part.

> My team uses watch mode pretty heavily, and the new behavior seems pretty strange to me. I'm already anticipating the complaints if we upgrade to SBT 1.3 and lose the current mechanism for exiting the application.

Do you think `Enter` dropping into shell would be considered less friendly to users who like to `sbt "~test"` directly from the terminal? Since shell users were already getting the behavior, I don't think @eatkins's intent was to make it less friendly for non-shell users. I think this is a useful input, and we are open to other solutions.

Maybe we can keep `Enter` for returning to wherever-your-from (shell or bash), and come up with an explicit way to enter into the sbt shell. `s` for `shell`?

eatkins:
@sarahgerweck I hate to be redundant, but I was 80% through this before @eed3si9n replied so I'm just going to post as it was:

Would you elaborate further on what you're concerned about? There are two distinct ways of entering watch for some command, let's use test for the sake of discussion:
1) from the shell, the user runs the command `~test`
2) from the command line, the user starts sbt with `sbt ~test`

In sbt < 1.3.0, the prompt looks like:
```
1. Waiting for source changes in project sbt-1-3-0-intro... (press enter to interrupt)
```
If the user hits enter, the behavior was dependent on how the watch was entered. If they had entered from the shell, they'd return to the shell, otherwise it would exit the application entirely. While I'm sure many users always run `sbt ~test` and never use the shell, this inconsistency could cause headaches for power users. I may have started sbt with `sbt ~test` and then decided I wanted to do something else. This meant I had to restart sbt entirely which is a headache if your build takes 30 seconds to load, which is not uncommon. To fix this, and make the behavior consistent, sbt changes the semantics and prints the following prompt:
```
[info] 1. Monitoring source files for updates...
[info] Project: sbt-1-3-0-intro
[info] Command: test
[info] Options:
[info]   <enter>: return to the shell
[info]   'r': repeat the current command
[info]   'x': exit sbt
``` 
Now the options are consistent no matter how the watch is entered and it is more clear exactly what they do (the `(press enter to interrupt)` instruction was vague and incorrect if the program was started with a continuous command).

I understand that people can be resistant to change, but I'm struggling to see how a one character difference in exiting sbt is a huge deal especially given that it is loudly printed to the console. If you are really concerned about your users, you can revert to the sbt 1.2.8 behavior with:

```scala
import sbt.internal.util.complete.Parser._
ThisBuild / watchInputParser := {
  '\r' ^^^ new Watch.Run("") | '\n' ^^^ new Watch.Run("")
}
ThisBuild / watchStartMessage := { (i, p, _) =>
  Some(s"$i. Waiting for source changes in project ${p.project}... (press enter to interrupt)")
}
```
I can see an argument for adding some kind of compatibility flag that sets those automatically.

eatkins:
@dwijnand it's easy to make Ctrl+D work:
```scala
ThisBuild / watchInputParser := 4.toChar ^^^ new Watch.Run("iflast shell") | Watch.defaultInputParser
```
If this were made default, then we'd have to change the start message as well.

eed3si9n:
I am closing this for now in favor this Discuss thread. If you have ideas/opinions, please comment them there.
https://discuss.lightbend.com/t/keyboard-shortcuts-during-triggered-execution/4585

PR #4848, opened 2019-07-09
Use last modified instead of hash

eatkins:
I noticed that for a simple spark project that evaluating the test task
was faster than running run when both tasks evaluated the same code
block. I tracked this down to the BackgroundJobService.copyClasspath
method. This method was hashing the jar contents of all of the files in
the build. On my computer, this took 600ms (for context, the total run
time of the `run` task was around 1.2 seconds, which included about
150ms of scala compiling and 350ms of time in the main method). If
instead we use the last modified time it drops down to 5-10ms. As
predicted, the total runtime of `run` in this project dropped down to
600ms which was on par with `test`.

I am not sure why a hash was used rather than last modified in the first place,
so I reworked things in such a way that, by default, sbt will use a hash
but if turbo mode is on, it will use the last modified instead. We can
revisit the default later.


lightbend-cla-validator:

eed3si9n:
> I am not sure why a hash was used rather than last modified in the first place,
so I reworked things in such a way that, by default, sbt will use a hash
but if turbo mode is on, it will use the last modified instead. We can
revisit the default later.

I implemented this in https://github.com/sbt/sbt/pull/2936. I am guessing that I chose content hash because I don't fully trust timestamps. On Docker and other filesystems, its resolution is sometimes around 1s. Also switching the branch around can disturb what's new or old. I am happy to reevaluate the default value if it's affecting the responsiveness of `run`.

eed3si9n (approved):
Thanks!

PR #4850, opened 2019-07-09
Add support for in memory cache store

eatkins:
It can be quite slow to read and parse a large json file. Often, we are
reading and writing the same file over and over even though it isn't
actually changing. This is particularly noticeable with the
UpdateReport*. To speed this up, I introduce a global cache that can be
used to read values from a CacheStore. When using the cache, I've seen
the time for the update task drop from about 200ms to about 1ms. This
ends up being a 400ms time savings for test because update is called for
both Compile / compile and Test / compile.

The way that this works is that I add a new abstraction
CacheStoreFactoryFactory, which is the most enterprise java thing I've
ever written. We store a CacheStoreFactoryFactory in the sbt State.
When we make Streams for the task, we make the Stream's
cacheStoreFactory field using the CacheStoreFactoryFactory. The
generated CacheStoreFactory may or may not refer to a global cache.

The CacheStoreFactoryFactory may produce CacheStoreFactory instances
that delegate to a Caffeine cache with a max size parameter that is
specified in bytes by the fileCacheSize setting (which can also be set
with -Dsbt.file.cache.size). The size of the cache entry is estimated by
the size of the contents on disk. Since we are generally storing things
in the cache that are serialized as json, I figure that this should be a
reasonable estimate. I set the default max cache size to 128MB, which is
plenty of space for the previous cache entries for most projects. If the
size is set to 0, the CacheStoreFactoryFactory generates a regular
DirectoryStoreFactory.

To ensure that the cache accurately reflects the disk state of the
previous cache (or other cache's using a CacheStore), the Caffeine cache
stores the last modified time of the file whose contents it should
represent. If there is a discrepancy in the last modified times (which
would happen if, say, clean has been run), then the value is read from
disk even if the value hasn't changed.

* With the following build.sbt file, it takes roughly 200ms to read and
parse the update report on my compute:

libraryDependencies += "org.apache.spark" %% "spark-sql" % "2.4.3"
libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.1"

This is because spark-sql has an enormous number of dependencies and the
update report ends up being 3MB.

### DIFF
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -489,7 +489,7 @@ object Keys {
   val globalPluginUpdate = taskKey[UpdateReport]("A hook to get the UpdateReport of the global plugin.").withRank(DTask)
   private[sbt] val taskCancelStrategy = settingKey[State => TaskCancellationStrategy]("Experimental task cancellation handler.").withRank(DTask)
   private[sbt] val cacheStoreFactory = AttributeKey[CacheStoreFactoryFactory]("cache-store-factory")
-  val fileCacheSize = settingKey[Long]("The approximate maximum size in bytes of the cache used to store previous task results.")
+  val fileCacheSize = settingKey[String]("The approximate maximum size in bytes of the cache used to store previous task results. For example, it could be set to \"256M\" to make the maximum size 256 megabytes.")
 
   // Experimental in sbt 0.13.2 to enable grabbing semantic compile failures.
   private[sbt] val compilerReporter = taskKey[xsbti.Reporter]("Experimental hook to listen (or send) compilation failure messages.").withRank(DTask)
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -10,12 +10,15 @@ package sbt
 import java.io.File
 import java.net.URL
 
+import lmcoursier.definitions.CacheLogger
+import lmcoursier.{ CoursierConfiguration, FallbackDependency }
 import org.apache.ivy.core.module.descriptor.ModuleDescriptor
 import org.apache.ivy.core.module.id.ModuleRevisionId
 import org.apache.logging.log4j.core.Appender
 import sbt.BuildSyntax._
 import sbt.Def.ScopedKey
 import sbt.KeyRanks._
+import sbt.internal.InMemoryCacheStore.CacheStoreFactoryFactory
 import sbt.internal._
 import sbt.internal.inc.ScalaInstance
 import sbt.internal.io.WatchState
@@ -31,8 +34,6 @@ import sbt.nio.file.Glob
 import sbt.testing.Framework
 import sbt.util.{ Level, Logger }
 import xsbti.compile._
-import lmcoursier.definitions.CacheLogger
-import lmcoursier.{ CoursierConfiguration, FallbackDependency }
 
 import scala.concurrent.duration.{ Duration, FiniteDuration }
 import scala.xml.{ NodeSeq, Node => XNode }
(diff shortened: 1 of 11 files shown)


lightbend-cla-validator:

lightbend-cla-validator:

main/src/main/scala/sbt/internal/InMemoryCacheStore.scala
```diff
@@ -0,0 +1,129 @@
...
+        case Some((value: T, `lastModified`)) => value
+        case _                                => cacheStore.read[T]()
+      }
+    }
+    override def write[T](value: T)(implicit writer: JsonWriter[T]): Unit = {
+      val lastModified = IO.getModifiedTimeOrZero(path.toFile)
+      store.get(path) match {
+        case Some((v: T, `lastModified`)) if v == value => // nothing has changed
+        case _ =>
+          store.remove(path)
```
ben-manes, line 74: You might want to use a `cache.asMap().compute` to avoid races. This way if multiple threads read a stale timestamp and try to reinsert, only one has to do the work. You could do something like,

```java
cache.asMap().compute(key, oldValue) -> {
  if old value is absent or stale
    // compute new value
    return newValue
  else
    return oldValue
})
```

main/src/main/scala/sbt/Keys.scala
```diff
@@ -487,6 +488,8 @@ object Keys {
   val pluginData = taskKey[PluginData]("Information from the plugin build needed in the main build definition.").withRank(DTask)
   val globalPluginUpdate = taskKey[UpdateReport]("A hook to get the UpdateReport of the global plugin.").withRank(DTask)
   private[sbt] val taskCancelStrategy = settingKey[State => TaskCancellationStrategy]("Experimental task cancellation handler.").withRank(DTask)
+  private[sbt] val cacheStoreFactory = AttributeKey[CacheStoreFactoryFactory]("cache-store-factory")
+  val fileCacheSize = settingKey[Long]("The approximate maximum size in bytes of the cache used to store previous task results.")
```
eed3si9n: We should provide `parseBytes` function that can be used as `Defaults.parseBytes("128M")`  like

https://github.com/lightbend/config/blob/558c1e22f53c7f165daaf3e1e792399087d79f93/config/src/main/java/com/typesafe/config/impl/SimpleConfig.java#L851-L852

eatkins: I agree. I'm just trying to think of a better place to put it than Defaults.

main/src/main/scala/sbt/internal/InMemoryCacheStore.scala
eatkins: Thanks for the suggestion! I was looking at the javadocs for ConcurrentMap and it looks like I'd also need to add some synchronization to prevent multiple threads from concurrently writing the same file. Based on the [javadocs](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentMap.html#compute-K-java.util.function.BiFunction-), it looks like this could actually cause more work to be done without synchronization:
> The default implementation is equivalent to performing the following steps for this map, then returning the current value or null if absent:
>
>
>```java
> V oldValue = map.get(key);
> V newValue = remappingFunction.apply(key, oldValue);
> if (oldValue != null ) {
>    if (newValue != null)
>       map.replace(key, oldValue, newValue);
>    else
>       map.remove(key, oldValue);
> } else {
>    if (newValue != null)
>       map.putIfAbsent(key, newValue);
>    else
>       return null;
 >}
>```
> 
>The default implementation may retry these steps when multiple threads attempt updates including potentially calling the remapping function multiple times.

The way this cache is used, I don't think that multiple threads should really ever be trying to modify the same file. The CacheStore provides a read/write api for a specific file corresponding to a task. For example, the previous value of a task. Since a task is only run once per command evaluation and sbt doesn't run commands concurrently, these files should only be written to at most once per command. It is a low throughput cache. I'll maybe add a comment about concurrency though on the chance that this becomes an issue.

ben-manes: The default implementation is not atomic due to backwards compatibility, etc. However `ConcurrentHashMap` and `Caffeine` do implement it atomically so it is safe to rely on. You just have to be mindful when using other `ConcurrentMap` types, like `ConcurrentSkipListMap` or external ones like `NonBlockingHashMap` which do not support atomic computations.

If it's not an issue due to how you are using it, that's even better.

ben-manes: >Attempts to compute a mapping for the specified key and its current mapped value (or null if there is no current mapping). The entire method invocation is performed atomically. Some attempted update operations on this map by other threads may be blocked while computation is in progress, so the computation should be short and simple, and must not attempt to update any other mappings of this Map.

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#compute-K-java.util.function.BiFunction-

eatkins: Aah, cool. Thanks for the clarification. I wasn't aware of the compute/merge methods and I could see them being useful for some other things I'm working on even if I don't change anything here.

main/src/main/scala/sbt/Keys.scala
eatkins: I ended up making the setting a String and leaving the parser private[sbt]: https://github.com/sbt/sbt/commit/a4688b0d2ed862b7ce63965b9cbca2a91c69f35f

eed3si9n (dismissed):

eed3si9n (approved):

PR #4855, opened 2019-07-11
add allCredentials to emulate credential registration

### LINKED ISSUE #4802: [1.3.0-RC2] credentials are not cached across the subprojects
## steps
sbt version: 1.3.0-RC2

## problem
When trying to compile code with Lightbend Commercial credentials, coursier fails to resolve those dependencies.

```
[error] Error downloading com.lightbend.cinnamon:cinnamon-slf4j-events:2.11.1
[error]   Not found
[error]   not found: <userhome>/.ivy2/local/com.lightbend.cinnamon/cinnamon-slf4j-events/2.11.1/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/com/lightbend/cinnamon/cinnamon-slf4j-events/2.11.1/cinnamon-slf4j-events-2.11.1.pom
[error]   unauthorized: https://repo.lightbend.com/commercial-releases/com/lightbend/cinnamon/cinnamon-slf4j-events/2.11.1/cinnamon-slf4j-events-2.11.1.pom (Bintray)
```

## expectation
Compilation should complete as normal, with credentials found in the location indicated in .sbt files.

## notes
Repository configuration:

> project/cinnamon.sbt
```
addSbtPlugin("com.lightbend.cinnamon" % "sbt-cinnamon" % "2.11.1")
// Credentials and resolver to download the Cinnamon Telemetry libraries
credentials += Credentials(Path.userHome / ".lightbend" / "commercial.credentials")
resolvers += Resolver.url("lightbend-commercial", url("https://repo.lightbend.com/commercial-releases"))(Resolver.ivyStylePatterns)
```

If I set `ThisBuild / useCoursier := true` with 1.3.0-RC2, the build will fail.
If I set `ThisBuild / useCoursier := false` with 1.3.0-RC2, the build will succeed.

Setting `host = repo.lightbend.com` in my `commercial.credentials` file does not resolve the issue.  Nor does leaving it as `host = dl.bintray.com`.

Downgrading to 1.2.7 also allows the build to succeed.

eed3si9n:
Fixes #4802

For Ivy integration sbt uses credential task in a peculiar way. https://github.com/sbt/sbt/blob/9fa25de84cf6fb521aa5f161c866837200ec17c4/main/src/main/scala/sbt/Defaults.scala#L2271-L2275

This lets the build user put `credential` task in various places, like metabuild or root project, but they would all act as if they were scoped globally. This PR adds `allCredentials` task to emulate the behavior to pass credentials into lm-coursier.

### DIFF
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -376,6 +376,7 @@ object Keys {
   val packagedArtifacts = taskKey[Map[Artifact, File]]("Packages all artifacts for publishing and maps the Artifact definition to the generated file.").withRank(CTask)
   val publishMavenStyle = settingKey[Boolean]("Configures whether to generate and publish a pom (true) or Ivy file (false).").withRank(BSetting)
   val credentials = taskKey[Seq[Credentials]]("The credentials to use for updating and publishing.").withRank(BMinusTask)
+  val allCredentials = taskKey[Seq[Credentials]]("Aggregated credentials.").withRank(DTask)
 
   val makePom = taskKey[File]("Generates a pom for publishing when publishing Maven-style.").withRank(BPlusTask)
   val deliver = taskKey[File]("Generates the Ivy file for publishing to a repository.").withRank(BTask)
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -376,7 +376,7 @@ object Keys {
   val packagedArtifacts = taskKey[Map[Artifact, File]]("Packages all artifacts for publishing and maps the Artifact definition to the generated file.").withRank(CTask)
   val publishMavenStyle = settingKey[Boolean]("Configures whether to generate and publish a pom (true) or Ivy file (false).").withRank(BSetting)
   val credentials = taskKey[Seq[Credentials]]("The credentials to use for updating and publishing.").withRank(BMinusTask)
-  val allCredentials = taskKey[Seq[Credentials]]("Aggregated credentials.").withRank(DTask)
+  val allCredentials = taskKey[Seq[Credentials]]("Aggregated credentials across current and root subprojects. Do not rewire this task.").withRank(DTask)
 
   val makePom = taskKey[File]("Generates a pom for publishing when publishing Maven-style.").withRank(BPlusTask)
   val deliver = taskKey[File]("Generates the Ivy file for publishing to a repository.").withRank(BTask)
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/coursierint/LMCoursier.scala
```diff
@@ -19,11 +20,15 @@ import lmcoursier._
 import lmcoursier.credentials.Credentials
 import Keys._
 import sbt.librarymanagement._
+import sbt.librarymanagement.ivy.{ Credentials => IvyCredentials }
 import sbt.util.Logger
 import sbt.io.syntax._
 import xsbti.AppConfiguration
 
 object LMCoursier {
+  private[sbt] val credentialRegistry: mutable.Map[(String, String), IvyCredentials] =
```
eatkins: Could this be `private` or `private[this]`? It doesn't seem like it's referenced outside this file. Also, for mutable maps, I tend to default to `ConcurrentHashMap` since it seems to offer a better performance/thread safety combination than scala.mutable.Map. My guess is performance doesn't matter here, but there might be some concurrency.

main/src/main/scala/sbt/Keys.scala
```diff
@@ -376,6 +376,7 @@ object Keys {
   val packagedArtifacts = taskKey[Map[Artifact, File]]("Packages all artifacts for publishing and maps the Artifact definition to the generated file.").withRank(CTask)
   val publishMavenStyle = settingKey[Boolean]("Configures whether to generate and publish a pom (true) or Ivy file (false).").withRank(BSetting)
   val credentials = taskKey[Seq[Credentials]]("The credentials to use for updating and publishing.").withRank(BMinusTask)
+  val allCredentials = taskKey[Seq[Credentials]]("Aggregated credentials.").withRank(DTask)
```
eatkins: I think this needs either a more descriptive info string or to be `private[sbt]`. As it stands, I wouldn't be sure whether or not I should be updating `credentials` or `allCredentials`.

eatkins (changes requested):

main/src/main/scala/sbt/coursierint/LMCoursier.scala
eed3si9n: Both very good points.

main/src/main/scala/sbt/Keys.scala
eed3si9n: I'll update the text.

eed3si9n: 9755234

main/src/main/scala/sbt/coursierint/LMCoursier.scala
eed3si9n: 9755234

eatkins (approved):

PR #4861, opened 2019-07-13
Reload multi commands

### LINKED ISSUE #4860: [1.3.0-RC2] multi command ignored if followed by reload
## steps
sbt version: 1.3.0-RC2
run `;compile;reload`

## problem
`compile` is not executed

## expectation
both compile and reload are executed

## notes
there are perhaps other commands like reload which swallow the previous one
works fine in 1.3.0-RC1

eatkins:
Fixes #4860


eed3si9n (approved):
nice catch.

minor nit: `@`-mentioning someone in a commit should be avoided since they might get an email years from now when part of the subprojects sometimes moves off to another repo using git-filter.

eatkins:
Aah, fair enough. I just like to give people credit.

Issue #4868, opened 2019-07-16
[1.3.0-RC2] unmanagedSources does not respect excludeFilter


eed3si9n:
This was originally reported by @SethTisue 

## steps

```
hub clone typelevel/kittens
cd kittens
# change to sbt 1.3.0-RC2
sbt
```

```
set excludeFilter in (Test, unmanagedSources) in coreJVM := HiddenFileFilter || "partialOrder.scala" || "monoid.scala" || "semigroup.scala" || "order.scala"
show coreJVM / Test / unmanagedSources
```

## problem

it shows 

```
[info] * /Users/eed3si9n/work/kittens/core/src/test/scala/cats/derived/monoid.scala
```

## expectation

`monoid.scala` should not be included.

eed3si9n:
Fixed in https://github.com/sbt/sbt/pull/4872

PR #4872, opened 2019-07-16
Add scripted test for excludeFilter

eatkins:
It was reported in #4868 that exclueFilters didn't work correctly if
there was an || in the filter. This was an upstream issue in io, but
this commit adds a scripted test in sbt.


eatkins:
Depends on https://github.com/sbt/io/pull/248

eatkins:
Fixes #4868.

eed3si9n (approved):
Thanks!

PR #4874, opened 2019-07-18
Don't validate multi commands

### LINKED ISSUE #4875: [1.3.0-RC3] strict multi parser breaks the Intellij plugin
Originally reported by @eatkins 

## steps
sbt version: 1.3.0-RC3
Use IntelliJ

## problem

Doesn't work.

## expectation

Works.

## notes

To dump the project structure, they run the following multi command:

```
set _root_.scala.collection.Seq(historyPath := None,shellPrompt := { _ => "” };SettingKey[_root_.scala.Option[_root_.sbt.File]]("sbtStructureOutputFile") in _root_.sbt.Global := _root_.scala.Some(_root_.sbt.file("/private/var/folders/7q/s3vq91lx3zv9fw4l5l5c43nr0000gn/T/sbt-structure.xml”));SettingKey[_root_.java.lang.String]("sbtStructureOptions") in _root_.sbt.Global := "download, resolveClassifiers”); apply -cp "/Users/ethanatkins/Library/Application Support/IdeaIC2019.2/Scala/repo/org.jetbrains/sbt-structure-extractor/scala_2.12/sbt_1.0/2018.2.1+4-88400d3f/jars/sbt-structure-extractor.jar” org.jetbrains.sbt.CreateTasks; */*:dumpStructure)
```

eatkins:
Fixes #4875

We tried to prevent users from doing something like running a multi
command "foo; bar" where foo is valid but bar is invalid so that we
wouldn't run foo only to discover bar was an invalid key. It isn't
possible to know in general if any command other than the first command
in a multi command is valid because it might update the state and add
the initially invalid command.

The validation caused the intellij plugin to not work with 1.3.0-RC3.

### DIFF
--- a/main-command/src/main/scala/sbt/BasicCommands.scala
+++ b/main-command/src/main/scala/sbt/BasicCommands.scala
@@ -189,13 +189,6 @@ object BasicCommands {
   def multiParser(s: State): Parser[List[String]] = multiParserImpl(Some(s))
 
   def multiApplied(state: State): Parser[() => State] = {
-    def generateCommands(commands: List[String]): State =
-      commands
-        .takeWhile(_ != "reload")
-        .collectFirst {
-          case c if Parser.parse(c, state.combinedParser).isLeft => c :: state
-        }
-        .getOrElse(commands ::: state)
     Command.applyEffect(multiParserImpl(Some(state))) {
       // the (@ _ :: _) ensures tail length >= 1.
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):

eed3si9n:
GitHub is not updating, but the CI jobs were good:
https://travis-ci.org/sbt/sbt/builds/560281695
https://ci.appveyor.com/project/sbt/sbt/builds/26062812

PR #4876, opened 2019-07-18
Bump scalafmt

eatkins:
Intellij had issues resolving 2.0.0-RCX so it will be nice to be using
the latest.

### DIFF
--- a/internal/util-collection/src/test/scala/SettingsTest.scala
+++ b/internal/util-collection/src/test/scala/SettingsTest.scala
@@ -135,8 +135,10 @@ object SettingsTest extends Properties("settings") {
     forAll(mkAttrKeys[Int](2)) {
       case List(key, derivedKey) =>
         val projectKeys = for { proj <- 1 to nrProjects } yield ScopedKey(Scope(1, proj), key)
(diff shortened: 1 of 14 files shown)


internal/util-collection/src/test/scala/SettingsTest.scala
```diff
@@ -135,8 +135,10 @@ object SettingsTest extends Properties("settings") {
     forAll(mkAttrKeys[Int](2)) {
       case List(key, derivedKey) =>
         val projectKeys = for { proj <- 1 to nrProjects } yield ScopedKey(Scope(1, proj), key)
-        val projectDerivedKeys = for { proj <- 1 to nrProjects } yield
-          ScopedKey(Scope(1, proj), derivedKey)
+        val projectDerivedKeys = for { proj <- 1 to nrProjects } yield ScopedKey(
+          Scope(1, proj),
+          derivedKey
+        )
```
eed3si9n, line 141: not very happy with these superfluous formatting changes every bump brings, but I guess we have no option.

eatkins: Agreed in general, though I actually think that these changes are good. I hope that since we are off an RC version that things will be more stable moving forward.

PR #4884, opened 2019-07-18
Minor clarification of logging message

naferx:
Sometimes I find this message a bit confusing. It gives me a false impression of another server running behind the scenes.
Anyway, this PR can be closed if this clarification is unnecessary.


eed3si9n (dismissed):
@naferx Thanks!

PR #4898, opened 2019-07-25
Use '_' instead of '$' in path names

eatkins:
The use of '$' in the path names for streams is a pain because, since
'$' is a special character in the shell, it makes it impossible to
directly copy and paste the paths. If we make this change, some builds
will be left with vestigial directories with $global and $build in them
until they run clean. It also would break any scripts that manually
delete these paths. That doesn't seem like a common use case, but it's
worth mentioning.


eed3si9n:
>  It also would break any scripts that manually
delete these paths. That doesn't seem like a common use case, but it's
worth mentioning.

I don't think there are many scripts, but I do occasionally go in there to inspect things, so I guess the human workflow would need to change depending on the sbt version.

eed3si9n (approved):

PR #4900, opened 2019-07-26
Incremental task fixes

eatkins:
This PR enhances the semantics of the previous api so that the previous values can be stored with respect to the task in which the stored value is used. The old semantics didn't really make sense because the value of previous was decoupled from the task in which it ran. Consider a task definition like the following:
 ```scala
bar := {
   val current = foo.value
   foo.previous match {
     case Some(v) if v == current =>
     case None => runSideEffects(current)
   }
}
```
If the user runs `foo` independently from `bar`, the side effects in `bar` may not happen even though the value of `foo` has changed since the last time `bar` ran.

I introduce a new api `previousInThis` so that the task above could be written:
 ```scala
bar := {
   val current = foo.value
   foo.previousInThis match {
     case Some(v) if v == current =>
     case None => runSideEffects(current)
   }
}
```
and `foo` could be run numerous times between calls to `bar`, but the value of `foo.previousInThis` will only change when `bar` successfully completes.

Using this, I also fix the incremental file task evaluation semantics and add some extension methods so that, similarly to the above, we can write
```scala
bar := {
   foo.changedOutputFiles match {
     case Some(ChangedFiles(creations, updates, deletions)) =>
        process(creations ++ updates ++ deletions)
     case None =>
   }
}
```
In sbt 1.3.0-RC3, this would be written:
```scala
bar := {
   (foo / changedOutputFiles).value match {
     case Some(ChangedFiles(creations, updates, deletions)) =>
       process(creations ++ updates ++ deletions)
     case None =>
   }
}
```
but the semantics were wrong because if `foo` was run separately from `bar`, bar wouldn't have seen the changes. That is fixed by the `foo.changed(Input|Output)Files` api.


main-settings/src/main/scala/sbt/std/InputWrapper.scala
```diff
@@ -228,6 +247,13 @@ sealed abstract class MacroPrevious[T] {
   def previous(implicit format: sjsonnew.JsonFormat[T]): Option[T] =
     macro InputWrapper.previousMacroImpl[T]
 }
+sealed abstract class MacroPreviousInThis[T] {
+  @compileTimeOnly(
+    "`previousInThis` can only be used within a task macro, such as :=, +=, ++=, or Def.task."
+  )
+  def previousInThis(implicit format: sjsonnew.JsonFormat[T]): Option[T] =
```
eed3si9n, line 254: This sounds too similar to `.previous` but `this` in sbt's context is not really clear.
`lastSeen` might be a better name?

eatkins: Yeah. I have thought about names involving last as well.

eatkins: I also considered `previousInThisTask` for the reason you mentioned.

eatkins:
Closing in favor of #4927.

Issue #4801, opened 2019-06-11
deeply nested Incomplete


eed3si9n:
## steps

Introduce compilation error.

## problem

printstack trace bug, fixed in https://github.com/sbt/sbt/pull/4792, accidentally exposed the following Incomplete object, which is deeply nested.

> [circe] [error] (java8JVM / Compile / compileIncremental) Compilation failed
[...]

PR #4901, opened 2019-07-26
Don't use exception for reloading

eatkins:
I completely forgot about the StateTransform class which allows a task
to modify the state through its return value.


eatkins:
This also would have fixed #4801.

eed3si9n (approved):

PR #4906, opened 2019-07-29
Turbo resource loader

eatkins:


eed3si9n (approved):

main/src/main/scala/sbt/internal/ClassLoaders.scala
```diff
@@ -156,34 +152,28 @@ private[sbt] object ClassLoaders {
           }
           .getOrElse(scalaLibraryLayer)
 
-        // layer 2 (resources)
-        val resourceLayer =
-          if (layerDependencies)
-            getResourceLayer(cpFiles, resourceCP, scalaReflectLayer, cache, resources)
```
dwijnand, line 162: With this `resources` is no longer used, and therefore no longer serviced...

This breaks the compiler's `Settings#embeddedDefaults` which relies on those resources.

PR #4927, opened 2019-08-05
File report

eatkins:
This fixes up the semantics and api for incremental task evaluation. In particular, it adds two extension methods `inputFileReport` and `outputFileReport` for task, that allow the user to query which input or output files of a task have changed since the last successful evaluation of the task. The report has accessors: `all`, `created`, `deleted`, `hasChanges`, `modified`, `unmodified` that are all filled regardless of whether there are any previous file stamps available or whether or not there have been any changes. This makes the api more usable than what was introduced in 1.3.0. I wrote documentation for the new api in: https://github.com/sbt/website/pull/802 (the usage and semantics of `inputFileReport` and `outputFileReport` are described in detail in https://github.com/sbt/website/blob/6aa3b51524d8729ead4ce288fc964e1a948704cf/src/reference/04-Howto/15-Howto-Track-File-Inputs-and-Outputs.md).


eed3si9n:
> The problem is that foo.previous is stored in
effectively (foo / streams).value.cacheDirectory / "previous". This
means that it is completely decoupled from foo. Now, suppose that the
user runs something like:

>     > set foo := 1
>     > bar // processes the value 1
>     > set foo := 2
>     > foo
>     > bar // does not process the new value 2 because foo was called, which updates the previous value

In some cases `bar` just wants to avoid re-running an expensive task and it might be ok with actual previous value, and not the last-seen value.

eatkins:
Just to be clear, I didn't change the semantics of `.previous` in this PR.

Avoiding re-running an expensive task is easy even if previous was changed to be scoped by both the evaluation task and the enclosing task:
```
foo := foo.previous.getOrElse(doSomethingExpensive())
```
Now `bar` can just call `foo` directly and it gets the previous value so long as it exists on disk.

eed3si9n (approved):

eed3si9n (changes requested):
pending a few API changes we discussed in the meeting.

eatkins:
I updated this PR to include filters for incremental tasks. It now needs an io bump to pass CI.

eatkins:
I also added docs for file tree view: https://github.com/sbt/website/pull/802/commits/3a4d789234dc9b71e3dec251badaa87fc407c47f and the filters in incremental tasks: https://github.com/sbt/website/pull/802/commits/c0e6cd1f377d44d34d155a5820493915cda27420

eed3si9n (dismissed):

eatkins:
Needs https://github.com/sbt/io/pull/262 to pass on windows.

eed3si9n:
https://repo1.maven.org/maven2/org/scala-sbt/io_2.12/1.3.0-M15

eatkins:
CI is finally green.

eed3si9n (approved):

PR #4931, opened 2019-08-06
Use publishTo value when it is defined.

hvesalai:
Don't give error about missing repository if publishArtifact is false

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2338,10 +2338,8 @@ object Classpaths {
         if (isSnapshot.value) "integration" else "release",
         ivyConfigurations.value.map(c => ConfigRef(c.name)).toVector,
         packagedArtifacts.in(publish).value.toVector,
-        checksums.in(publish).value.toVector, { //resolvername: not required if publishTo is false
-          val publishToOption = publishTo.value
-          if (publishArtifact.value) getPublishTo(publishToOption).name else "local"
-        },
+        checksums.in(publish).value.toVector,
+        getPublishTo(publishTo.value, publishArtifact.value).name, // error if publishTo is None and publishArtifact is true
         ivyLoggingLevel.value,
         isSnapshot.value
       )
@@ -2934,8 +2932,16 @@ object Classpaths {
 
   def defaultRepositoryFilter: MavenRepository => Boolean = repo => !repo.root.startsWith("file:")
 
-  def getPublishTo(repo: Option[Resolver]): Resolver =
-    repo getOrElse sys.error("Repository for publishing is not specified.")
+  def getPublishTo(repo: Option[Resolver], publishArtifact: Boolean): Resolver =
+    repo.getOrElse {
+      if (publishArtifact)
+        sys.error("Repository for publishing is not specified.")
+      else
+        Resolver.defaultLocal
+    }
+
+  @deprecated("No longer used", "1.4.0")
+  def getPublishTo(repo: Option[Resolver]): Resolver = getPublishTo(repo, true)
 
   def publishConfig(
       publishMavenStyle: Boolean,


hvesalai:
Ping @dwijnand

hvesalai:
I'll rebase to the correct branch

dwijnand (dismissed):

hvesalai:
The travis failure is from MIMA:

```
method getPublishTo(scala.Option)sbt.librarymanagement.Resolver in object sbt.Classpaths does not have a correspondent in current version

[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("sbt.Classpaths.getPublishTo")
```

dwijnand:
Yeah, you need to create new method.

hvesalai:
done, needs new review and approval

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2338,10 +2338,8 @@ object Classpaths {
         if (isSnapshot.value) "integration" else "release",
         ivyConfigurations.value.map(c => ConfigRef(c.name)).toVector,
         packagedArtifacts.in(publish).value.toVector,
-        checksums.in(publish).value.toVector, { //resolvername: not required if publishTo is false
-          val publishToOption = publishTo.value
-          if (publishArtifact.value) getPublishTo(publishToOption).name else "local"
-        },
+        checksums.in(publish).value.toVector,
+        getPublishTo(publishTo.value, publishArtifact.value).name, // error if publishTo is None and publishArtifact is true
```
eed3si9n, line 2342: `publish / skip` might be better. See discussion here - https://github.com/sbt/sbt/issues/3136

eed3si9n (changes requested):

main/src/main/scala/sbt/Defaults.scala
hvesalai: Well... I don't mind the new solutoin, but in any case the current code on lines 2342-2344 is clearly a bug and should be replaced soon with some sane alternative.

How about just merging this PR and then working the publish / skip into it as #3136 gets implemented.

eed3si9n: publish / skip was implemented here - https://github.com/sbt/sbt/pull/3380

hvesalai: So you mean that instead of `publishArtifact.value` we should have here `(skip in publish).value` ?

eed3si9n: yes, it should be `!(skip in publish).value`, since you could have published artifacts even if unscoped `publishArtifact.value` returns false.

eed3si9n: actually could you hold off on that change until we understand what happened in the other PR please?

eed3si9n: I am wondering if a better fix would be to just revert https://github.com/sbt/sbt/pull/3760

mpollmeier: responded on https://github.com/sbt/sbt/pull/3760#issuecomment-518797685 - sounds like reverting that one is the way to go

hvesalai: #3760 is what broke our build so if that is reverted, all is fine for me.

hvesalai: @eed3si9n is it certain that if `skip in publish` is true, then nobody depends on the `publishConfiguration` ?

eed3si9n: publish, publishLocal, publishSigned, and sbt-bintray respect the skip in publish, so that should cover the relevant party.

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/4934

Issue #4948, opened 2019-08-15
[1.3.0-RC3] overcompilation on Community Build due to scala-library invalidation


eed3si9n:
https://github.com/scala/community-builds/issues/961 originally reported by @SethTisue 

## steps

Run dbuild with the following configuration:

https://github.com/SethTisue/community-builds/blob/overcount/configs/community.dbuild

if I run sbt in the dbuild debugging environment, with commands such as:

```
./dbuild-0.9.16/bin/dbuild checkout 861647e644cbd1ece7039aa834f043189849628b scala-parser-combinators  newdir
./newdir/.dbuild/start
```

then do

```
> project scala-parser-combinators
> compile
> compile
```

## problem

The second time everything is recompiled

## expectation

It should be a no-op.

## notes

Debug log shows:

```
[debug] Modified binary dependencies: Set(/Users/tisue/community.212/newdir/.dbuild/scala/d17810cf691d3aed3352a2ad17a5abf6dba26418/lib/scala-library.jar)
```

eed3si9n:
Hopefully fixed in https://github.com/sbt/sbt/pull/4949

PR #4949, opened 2019-08-15
Update ExternalHooks to look up changed binaries

eatkins:
It was reported that in community builds, sometimes there was
spurious over-compilation due to invalidation of the scala library jar
(https://github.com/sbt/sbt/issues/4948). The reason for this was that
the external hooks prefills the managed cache with all of the time
stamps for the project dependencies but was not looking up any jars that
weren't in the cache. I suspect I did this because I didn't realize that
zinc also includes its own classpath in the binaries which is not
a part of the dependencyClasspath. The fix is to just add the jar to the
cache if it doesn't already exist by switching to getOrElseUpdate from
get.

I followed the steps in #4948 and published a version of sbt locally
with this change and the spurious re-builds stopped.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -610,19 +610,30 @@ object Defaults extends BuildCommon {
     },
     externalHooks := {
       import sbt.nio.FileStamp.Formats.seqPathFileStampJsonFormatter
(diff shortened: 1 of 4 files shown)


eed3si9n (dismissed):
Thank you!

eed3si9n (approved):

PR #4951, opened 2019-08-15
add .idea to .gitignore

intracer:


eed3si9n:
I'd normally comment that ppl should setup their own global gitignore https://help.github.com/en/articles/ignoring-files, but maybe it's ok to add it for sbt/sbt since it comes up regularly.

eed3si9n (approved):

PR #4956, opened 2019-08-17
Allow custom show for a scope

eatkins:
I am writing a plugin that uses mangled task keys that are very hard to
read. It is helpful to be able to override the show config for these
scopes so that they look reasonable in supershell and in error
reporting.

This is strictly opt in so it should have no effect on any existing builds.

### DIFF
--- a/main-settings/src/main/scala/sbt/Scope.scala
+++ b/main-settings/src/main/scala/sbt/Scope.scala
@@ -178,6 +178,8 @@ object Scope {
   ): String =
     displayMasked(scope, sep, showProject, mask, false)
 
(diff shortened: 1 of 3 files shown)


main-settings/src/main/scala/sbt/Scope.scala
```diff
@@ -178,6 +178,8 @@ object Scope {
   ): String =
     displayMasked(scope, sep, showProject, mask, false)
 
+  val customShowString = AttributeKey[String]("scope-custom-show-string")
```
eed3si9n, line 195: Scaladoc here?

eed3si9n (approved):

Issue #3793, opened 2017-12-04
traceLevel setting is not taken into account since 1.x


2m:
## steps
Given a task
```
TaskKey[Unit]("testTrace") := {
  sys.error("oops")
}
```

The exception stack trace is printed out when the task is ran no matter what the value of `traceLevel` is set.

## problem
`traceLevel` setting does not suppress stack traces of being printed.

## notes

sbt version: 1.0.4, 1.1.0-RC1

Works as expected with sbt 0.13.16.

kshakir:
Would this be related to my [SO post](https://stackoverflow.com/questions/47601811/how-to-suppress-info-messages-in-sbt-1-0-0-1-0-4)? After a week no one seems to know the answer over there either.

dvdalilue:
I have the same issues with sbt `1.0.4`. The use of `-error` does not suppress any message (info nor debug). Finally, i changed to `0.13.16`

eed3si9n:
@dvdalilue That's a separate issue - https://github.com/sbt/sbt/issues/3849

Note to a potential contributor:

`ConsoleAppender` https://github.com/sbt/util/blob/v1.1.2/internal/util-logging/src/main/scala/sbt/internal/util/ConsoleAppender.scala has `getTrace` and `setTrace` that tracks the trace level. `setTrace` is currently not called in sbt 1.

https://github.com/sbt/sbt/blob/fd8b4224889fae96fdc035602ea083f4d99dd7c6/main/src/main/scala/sbt/Project.scala#L442-L444

The above calls https://github.com/sbt/sbt/blob/a28381e93a0c5e46d4c4a9d261cc5e0db3c9f384/main/src/main/scala/sbt/internal/LogManager.scala#L212-L230


This might be more tricky because in sbt loggers and appenders are separated.

eed3si9n:
```
sbt:hello> set check/traceLevel := 4
[info] Defining check / traceLevel
[info] The new value will be used by no settings or tasks.
[info] Reapplying settings...
[info] Set current project to hello (in build file:/Users/eed3si9n/work/hellotest/)
sbt:hello> check
[error] java.lang.RuntimeException: boom
[error] 	at scala.sys.package$.error(package.scala:30)
[error] 	at $0d89407eaa29f35a5d10$.$anonfun$$sbtdef$1(/Users/eed3si9n/work/hellotest/build.sbt:14)
[...]

PR #4962, opened 2019-08-19
Cleanup stack traces in error reporting

eatkins:
Ref https://github.com/sbt/sbt/issues/3793

sbt prints verbose stack traces that include a lot of lines that are
more or less useless for debugging. We can improve the situation by
sanitizing the exception's stack trace before calling log.trace. I
figured out that in EvaluateTask, the first internal class that would
come up was in TypeFunctions.scala. To sanitize the stack trace, we just
need to take all of the lines up to the line _before_ that line, because
there was always a scala function transform above

The behavior can be controlled with the showFullStackTrace key, which
can be set with -Dsbt.stacktrace=full.

Before:
[error] (gnu-make/target/objects/%.o) Nonzero exit value: 1
[error] java.lang.RuntimeException: Nonzero exit value: 1
[error]         at scala.sys.package$.error(package.scala:30)
[error]         at scala.sys.process.ProcessBuilderImpl$AbstractBuilder.slurp(ProcessBuilderImpl.scala:138)
[error]         at scala.sys.process.ProcessBuilderImpl$AbstractBuilder.$bang$bang(ProcessBuilderImpl.scala:108)
[error]         at $ea175b1130302fc8663b$.$anonfun$$sbtdef$18(gnu-make/build.sbt:7)
[error]         at sbt.internal.util.ManagedLogger.log(ManagedLogger.scala:24)
[error]         at sbt.util.Logger.info(Logger.scala:22)
[error]         at $ea175b1130302fc8663b$.$anonfun$$sbtdef$16(gnu-make/build.sbt:7)
[error]         at $ea175b1130302fc8663b$.$anonfun$$sbtdef$16$adapted(gnu-make/build.sbt:6)
[error]         at scala.Function1.$anonfun$compose$1(Function1.scala:49)
[error]         at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:62)
[error]         at sbt.std.Transform$$anon$4.work(Transform.scala:67)
[error]         at sbt.Execute.$anonfun$submit$2(Execute.scala:280)
[error]         at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:19)
[error]         at sbt.Execute.work(Execute.scala:289)
[error]         at sbt.Execute.$anonfun$submit$1(Execute.scala:280)
[error]         at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:178)
[error]         at sbt.CompletionService$$anon$2.call(CompletionService.scala:37)
[error]         at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[error]         at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[error]         at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[error]         at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[error]         at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[error]         at java.lang.Thread.run(Thread.java:748)

After:

[error] (gnu-make/target/objects/%.o) Nonzero exit value: 1
[error] java.lang.RuntimeException: Nonzero exit value: 1
[error]         at scala.sys.package$.error(package.scala:30)
[error]         at scala.sys.process.ProcessBuilderImpl$AbstractBuilder.slurp(ProcessBuilderImpl.scala:138)
[error]         at scala.sys.process.ProcessBuilderImpl$AbstractBuilder.$bang$bang(ProcessBuilderImpl.scala:108)
[error]         at $ea175b1130302fc8663b$.$anonfun$$sbtdef$18(gnu-make/build.sbt:7)
[error]         at sbt.internal.util.ManagedLogger.log(ManagedLogger.scala:24)
[error]         at sbt.util.Logger.info(Logger.scala:22)
[error]         at $ea175b1130302fc8663b$.$anonfun$$sbtdef$16(gnu-make/build.sbt:7)
[error]         at $ea175b1130302fc8663b$.$anonfun$$sbtdef$16$adapted(gnu-make/build.sbt:6)


main/src/main/scala/sbt/Keys.scala
```diff
@@ -59,6 +59,7 @@ object Keys {
   val logBuffered = settingKey[Boolean]("True if logging should be buffered until work completes.").withRank(CSetting)
   val sLog = settingKey[Logger]("Logger usable by settings during project loading.").withRank(CSetting)
   val serverLog = taskKey[Unit]("A dummy task to set server log level using Global / serverLog / logLevel.").withRank(CTask)
+  val showFullStackTrace = settingKey[Boolean]("Display the full stack trace including internal sbt classes if a task fails").withRank(DSetting)
```
eed3si9n, line 62: See https://github.com/sbt/sbt/issues/3793

eatkins: I think these are separate issues. The trace level sets a fixed limit on the number of lines of the stack trace to display. That being said, I think it would be straightforward to also apply the trace level as well in the portion of EvaluateTask where I truncate the stack trace. It might not fully resolve #3793 but would be a step in the right direction.

eed3si9n: https://github.com/sbt/util/blob/d4824a53ae23afffc6178d77032f27b229f414c4/internal/util-logging/src/main/scala/sbt/internal/util/StackTrace.scala#L18-L22

"If d is 0, then all elements are included up to (but not including) the first element that comes from sbt." so traceLevel set to `0` is the "cleaned up" version, and traceLevel set to `Int.MaxValue` is the raw version, roughly mapping to `false` and `true` in your new setting?

eed3si9n: ok, I think I have the fix - https://github.com/sbt/util/pull/211

eatkins: Ok. Are you going to reopen #4963? I like the change to set trace level to 0 on interactive sessions. The only very minor issue I have with the logger approach is that my approach also strips `[error] at scala.Function1.$anonfun$compose$1(Function1.scala:49)`. Whenever a task depends on another task, a line like that will appear. Your test task didn't depend on any other tasks so its stack trace was slightly simpler.

eed3si9n: I'll open another PR to bump util. The current default is already `-1` so that doesn't need to change. I'll see how it looks like with more complicated task.

eatkins:
Closing in favor of #4965

Issue #4970, opened 2019-08-21
[1.3.0-RC4] StackOverflowError in trigerred tasks


OlegYch:
## steps
sbt version: 1.3.0-RC4

## problem
running any task with `~` produces 
```
[debug] > Exec(~update, None, None)
[debug] Evaluating tasks: checkBuildSources
[debug] Running task... Cancel: Signal, check cycles: false, forcegc: true
[debug] Forcing garbage collection...
[debug] Running task... Cancel: Signal, check cycles: false, forcegc: true
[error] java.lang.StackOverflowError
[error] 	at sbt.nio.file.Glob$ordering$.compare(Glob.scala:270)
[error] 	at sbt.nio.file.Glob$ordering$.compare(Glob.scala:265)
[error] 	at sbt.nio.file.Glob$ordering$.compare(Glob.scala:270)
[error] 	at sbt.nio.file.Glob$ordering$.compare(Glob.scala:265)
[error] 	at sbt.nio.file.Glob$ordering$.compare(Glob.scala:270)

```

## expectation
no error

## notes
this is a play project using play-sbt-plugin 2.7.3
this was working with RC3

OlegYch:
here is a repro project https://github.com/OlegYch/sbt-4970

eatkins:
Thanks for the repro. It's an easy fix.

OlegYch:
doesn't look like this is fully fixed
```
[debug] > Exec(~compile, None, None)
[debug] Evaluating tasks: checkBuildSources
[debug] Running task... Cancel: Signal, check cycles: false, forcegc: true
[debug] Running task... Cancel: Signal, check cycles: false, forcegc: true
[error] java.lang.IllegalArgumentException: Comparison method violates its general contract!
[error] 	at java.util.TimSort.mergeHi(TimSort.java:899)
[error] 	at java.util.TimSort.mergeAt(TimSort.java:516)
[error] 	at java.util.TimSort.mergeForceCollapse(TimSort.java:457)
[error] 	at java.util.TimSort.sort(TimSort.java:254)
[error] 	at java.util.Arrays.sort(Arrays.java:1438)
[error] 	at scala.collection.SeqLike.sorted(SeqLike.scala:659)
[error] 	at scala.collection.SeqLike.sorted$(SeqLike.scala:647)
[error] 	at scala.collection.AbstractSeq.sorted(Seq.scala:45)
[...]

PR #4971, opened 2019-08-21
Play watch scripted

eatkins:
This adds a test for https://github.com/sbt/sbt/issues/4970. It requires an io version built off of https://github.com/sbt/io/pull/266 to pass.


eed3si9n:
@eatkins Could you move plugin-specific tests to https://github.com/sbt/sbt/tree/develop/sbt/src/sbt-test/plugins plz?

eatkins:
done

Issue #4972, opened 2019-08-21
Clean is slower in sbt 1.3.0


lbialy:
## steps
sbt version: 1.3.0-RCx

## problem
sbt is getting gradually slower instead of getting faster. I am currently optimizing build in my work project (~60 subprojects) and I was doing some trivial comparisons of performance in 1.3 as a counterargument to migration to gradle. I have written a small script to roughly measure differences with bash's `time`:

```bash
#!/bin/bash

BRANCHES=("sbt-1.2.8" "sbt-1.3.0-RC3" "sbt-1.3.0-RC4" "sbt-1.3.0-RC4-turbo")

cd WORK_PROJECT

for BRANCH in ${BRANCHES[@]}
do
        git checkout "$BRANCH"
        exec 3>&1 4>&2
        result=$( { time sbt clean "some-subproject/test" 1>/dev/null 2>/dev/null; } 2>&1 )
        exec 3>&- 4>&-
        echo "Results for $BRANCH:"
        echo $result
        echo ""
done
```

There are 4 branches as you can see: 
* 1.2.8 with sbt-coursier plugin enabled
* 1.3.0-RC3
* 1.3.0-RC4
* 1.3.0-RC4 with turbo mode enabled

Turbo mode can't be enabled for RC3 due to a bug that crashes some of the tests.

Results are as follow:

```
λ ./sbt_perftest.sh                                                                                                                                                                                 
Already on 'sbt-1.2.8'
Results for sbt-1.2.8:
real 0m32.584s user 2m3.421s sys 0m2.241s

Switched to branch 'sbt-1.3.0-RC3'
Results for sbt-1.3.0-RC3:
real 0m44.143s user 2m47.532s sys 0m3.471s

Switched to branch 'sbt-1.3.0-RC4'
Results for sbt-1.3.0-RC4:
real 0m57.424s user 3m18.849s sys 0m4.514s

Switched to branch 'sbt-1.3.0-RC4-turbo'
Results for sbt-1.3.0-RC4-turbo:
real 0m37.382s user 2m7.711s sys 0m3.870s
```

## expectation
sbt not getting slower?

## notes
**The biggest offender seems to be the `clean` task which takes:**
* 0 seconds on 1.2.8
* 4 seconds on 1.3.0-RC3
* 9 seconds on 1.3.0-RC4
* 9 seconds on 1.3.0-RC4 with turbo mode 

test target also differs (it's an empty scalatest target, there are no test suites in this subproject):
[...]

PR #4978, opened 2019-08-22
Fix clean performance

eatkins:
The clean task got a lot slower in 1.3.0
(https://github.com/sbt/sbt/issues/4972). The reason for this was that
sbt 1.3.0 generated many custom clean tasks for any tasks that returned
`File` or `Seq[File]`. Each of these tasks was tagged with
Tags.Clean which meant that only one of them could run at a time. As a
result, it took a long time to evaluate all of the custom tasks, even if
they were no-ops. In the akka project, a no-op clean was taking 35
seconds which is simply unacceptable. After this change, a no-op clean
takes less than a second in akka (a full clean only takes about 6
seconds after running test:compile)

To fix this, I stopped aggregating the clean task across configs and
projects. Because I removed the aggregation, I needed to manually
implement clean in the `Compile` and `Test` configurations to make
`Compile / clean` and `Test / compile` clean work correctly.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -593,8 +593,14 @@ object Defaults extends BuildCommon {
   lazy val configTasks: Seq[Setting[_]] = docTaskSettings(doc) ++ inTask(compile)(
     compileInputsSettings
   ) ++ configGlobal ++ defaultCompileSettings ++ compileAnalysisSettings ++ Seq(
-    clean := Clean.task(ThisScope, full = false).value,
-    fileOutputs in compile := Seq(Glob(classDirectory.value, RecursiveGlob / "*.class")),
+    compileOutputs := {
+      import scala.collection.JavaConverters._
+      val classFiles =
+        manipulateBytecode.value.analysis.readStamps.getAllProductStamps.keySet.asScala
+      classFiles.toSeq.map(_.toPath) :+ compileAnalysisFileTask.value.toPath
+    },
+    compileOutputs := compileOutputs.triggeredBy(compile).value,
+    clean := (compileOutputs / clean).value,
     compile := compileTask.value,
     internalDependencyConfigurations := InternalDependencies.configurations.value,
     manipulateBytecode := compileIncremental.value,
--- a/main/src/main/scala/sbt/nio/Keys.scala
+++ b/main/src/main/scala/sbt/nio/Keys.scala
@@ -164,6 +164,7 @@ object Keys {
     taskKey[Seq[Path]]("The dependency classpath for a task.").withRank(Invisible)
   private[sbt] val classpathFiles =
     taskKey[Seq[Path]]("The classpath for a task.").withRank(Invisible)
+  private[sbt] val compileOutputs = taskKey[Seq[Path]]("Compilation outputs").withRank(Invisible)
 
   private[this] val hasCheckedMetaBuildMsg =
     "Indicates whether or not we have called the checkBuildSources task. This is to avoid warning " +
--- a/main/src/main/scala/sbt/nio/Settings.scala
+++ b/main/src/main/scala/sbt/nio/Settings.scala
@@ -9,36 +9,33 @@ package sbt
 package nio
 
 import java.io.File
-import java.nio.file.Path
+import java.nio.file.{ Files, Path }
 import java.util.concurrent.ConcurrentHashMap
 
-import sbt.Project._
+import sbt.Keys._
 import sbt.internal.Clean.ToSeqPath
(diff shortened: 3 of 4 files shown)


eed3si9n:
In sbt 1.2.8, `clean` would wipe out the `target` directory. Does the new `clean` go through one file at a time? Wouldn't this cause significant slowness on Windows for example? Also would `clean` no longer wipe out the `target/stream` (task caches)?

For scoped cleaning like `Compile / clean` it might make sense to be more precise etc, but for unscoped `clean` we should clean as efficiently as we can + do whatever magic to invalidate watch cache state.

eatkins:
There is very little difference between sbt 1.3.0 and 1.2.8 for unscoped clean after this change. In both cases, sbt lists all of the files under target and deletes them provided that they are not excluded by the clean exceptions. sbt 1.2.8 does not just do the equivalent of `rm -rf target`. What was causing slowness was that I made unscoped clean depend on all of the individual task cleans. This commit just undoes that behavior.

eed3si9n:
In sbt 1.2.8, the `cleanFiles` return items directly under `target` - https://github.com/sbt/sbt/blob/v1.2.8/main/src/main/scala/sbt/Defaults.scala#L1226

This looks like this on Akka:

```
akka > show cleanFiles
[info] akka-persistence-shared / cleanFiles
[info] 	Vector(/Users/eed3si9n/work/akka/akka-persistence-shared/target/update, /Users/eed3si9n/work/akka/akka-persistence-shared/target/streams)
[info] akka-protobuf / cleanFiles
[info] 	Vector(/Users/eed3si9n/work/akka/akka-protobuf/target/update, /Users/eed3si9n/work/akka/akka-protobuf/target/classes, /Users/eed3si9n/work/akka/akka-protobuf/target/streams)
[info] akka-stream-tests-tck / cleanFiles
[info] 	Vector(/Users/eed3si9n/work/akka/akka-stream-tests-tck/target/update, /Users/eed3si9n/work/akka/akka-stream-tests-tck/target/streams)
[info] akka-agent / cleanFiles
[info] 	Vector(/Users/eed3si9n/work/akka/akka-agent/target/update, /Users/eed3si9n/work/akka/akka-agent/target/classes, /Users/eed3si9n/work/akka/akka-agent/target/streams)
[info] akka-persistence / cleanFiles
[info] 	Vector(/Users/eed3si9n/work/akka/akka-persistence/target/test-classes, /Users/eed3si9n/work/akka/akka-persistence/target/update, /Users/eed3si9n/work/akka/akka-persistence/target/classes, /Users/eed3si9n/work/akka/akka-persistence/target/streams)
[info] akka-testkit / cleanFiles
[info] 	Vector(/Users/eed3si9n/work/akka/akka-testkit/target/test-classes, /Users/eed3si9n/work/akka/akka-testkit/target/update, /Users/eed3si9n/work/akka/akka-testkit/target/classes, /Users/eed3si9n/work/akka/akka-testkit/target/streams)
....
```

so my guess is that if the new `clean` is deleting files one by one, even with this patch, non-no-op `clean` would be significantly slower on Windows.

eatkins:
There is no way to delete files but one by one. In 1.2.8, the definition of clean was:
```
 clean := (Def.task { IO.delete(cleanFiles.value) } tag (Tags.Clean)).value,
```
If you look at the definition of `IO.delete`, you'll see that it just traverses the file system and deletes files one by one as it goes. The implementation in 1.3.0 is different because it uses nio apis, but the semantics are very similar.

eed3si9n:
> There is no way to delete files but one by one. 

ok.

so in general how does the new `clean` aggregate the files to clean?
For example, would it know to clean files generated by existing plugins like sbt-assembly? Also does it clean `target / streams`?

eatkins:
It lists the project target directory and deletes all the files except those excluded by cleanKeepFiles.

eatkins:
I think you might be confused by what I meant when I said I was dropping aggregation. What sbt 1.3.0x introduced was that I updated the unscoped clean task to run all of the custom clean tasks for each of the _individual_ tasks within a project or config that had clean defined. This would allow a user to set a different target directory that was not in the same directory as the project's target directory and still have it be deleted by clean. The idea is I could do
```
objectDir := baseDirectory.value / "objects"
compileC := { ... } // writes to objectDir.value
```
and have unscoped `clean` delete the files in `baseDirectory.value / "objects"` as well as `baseDirectory.value / "target"`. It is actually possible to do this without hosing performance, but I realized that aggregation caused other problems because it meant that `Test / clean` might delete files generated by `Compile / clean`.

eed3si9n:
> It lists the project target directory and deletes all the files except those excluded by cleanKeepFiles.

cool. Thanks for the clarification.

eed3si9n (approved):

PR #4980, opened 2019-08-23
Stop injecting file management settings for io tasks

eatkins:
It turns out that injecting the keys necessary for incremental tasks
causes a significant startup penalty for many larger projects. For
example, akka starts up about 3 seconds faster if do not inject these
settings for the tasks returning `File` or `Seq[File]`. Given that all
of these apis use java.nio.file anyway, it makes sense to not backport
them to older tasks.


eed3si9n (approved):
Thanks!

PR #4981, opened 2019-08-23
Startup perf

eatkins:
There were a number of serial bottlenecks during load. I was able to parallelize a number of them. In the akka project, these changes reduces startup time by about 1.5 seconds on my computer. When paired with https://github.com/sbt/sbt/pull/4980, the total startup time dropped from about 22.5 seconds to 19-19.5 seconds. sbt 1.2.8 was around 20 seconds even though sbt 1.3.0 generates 46770 settings compared to 38962 for sbt 1.2.8.


eed3si9n (approved):
nice

PR #4982, opened 2019-08-23
avoid force gc during load

eed3si9n:
This initializes the lastGcCheck to the current time so it won't force GC in the first 10 minutes, avoiding unnecessary GC during load.


eatkins (approved):
LGTM!

Issue #4973, opened 2019-08-21
sbt 1.3 regression: can't set `scalaVersion` when using `ScriptMain`


SethTisue:
## steps

here is my script:

```
#!/usr/bin/env sbt -Dsbt.version=1.2.8 -Dsbt.main.class=sbt.ScriptMain -Dsbt.supershell=false

/***
scalaVersion := "2.13.0"
*/

println(util.Properties.versionString)
```

this prints `version 2.13.0`, as expected

but if I change `1.2.8` to `1.3.0-RC4`...

## problem

...then it prints `version 2.12.8` ?!

SethTisue:
note that it's not that the `/***` stuff is being ignored entirely. I discovered the problem in a real script with more settings. all the settings in the script took effect except for the `scalaVersion` one

eatkins:
Thanks for the report. Should be fixed by #4988.

eed3si9n:
Merged #4988

PR #4988, opened 2019-08-27
Fix settings in ScriptMain

eatkins:
It was reported in https://github.com/sbt/sbt/issues/4973 that the
scalaVersion setting was not being correctly set in a script running
with ScriptMain using 1.3.0-RC4. Using git bisect, I found that the
issue was introduced in
https://github.com/sbt/sbt/commit/73cfd7c8bdd484b49ff6695ef006f5f7ba995c0b.
That commit manipulates the classloaders passed in by the launcher, but
only for the xMain entry point. I found that the script ran correctly if
I updated the classloader for ScriptedMain as well.

After these changes, the example script in #4973 correctly prints 2.13.0
for the scala version with a locally published sbt.

Bonus: rename xMainImpl object xMain. It was private[sbt] anyway.


eed3si9n (approved):
Nice. Thanks!

PR #5012, opened 2019-08-30
Don't sort watch inputs

eatkins:
There seems to be an issue with the sorting of globs and the ordering
doesn't really matter. I think I had added sorting as a very minor
performance optimization for registering directories with swoval in
order. Until there is evidence that it matters, it makes sense to just
stop sorting.

Should fix https://github.com/sbt/sbt/issues/4970.


eed3si9n (changes requested):
Could you rebase to 1.3.x plz?

eatkins:
Closing in favor of https://github.com/sbt/io/pull/267.

PR #5022, opened 2019-09-02
Allow supershell in no color mode

eatkins:
Disabling supershell when color mode is disabled is a sensible default
(especially for piped output). However, I think it should still be
possible to use supershell in no color mode.

This requires a util change that also enables supershell in no color
mode (see https://github.com/sbt/util/pull/216).

### DIFF
--- a/main/src/main/scala/sbt/internal/SysProp.scala
+++ b/main/src/main/scala/sbt/internal/SysProp.scala
@@ -88,7 +88,7 @@ object SysProp {
 
   def fileCacheSize: Long =
     SizeParser(System.getProperty("sbt.file.cache.size", "128M")).getOrElse(128L * 1024 * 1024)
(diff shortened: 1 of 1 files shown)


eed3si9n:
> This requires a util change that also enables supershell in no color
mode (see sbt/util#216).

"color" means both literal color and ANSI code (ANSI X3.64 control sequences). doesn't the new implementation still rely on ANSI code?

eatkins:
Yeah, it still relies on ANSI codes but there are situations in which you might not want colored output even though your system supports ANSI. Some people just prefer plain uncolored output. This doesn't change the default. If the system doesn't support ansi, the output is not tty, or the user explicitly sets -no-color, then supershell will be disabled. This change just forces util to try and use supershell if supershell is explicitly set via system property or setting. Most of the time, I imagine people will not be explicitly setting supershell, especially not in their builds.

eed3si9n (approved):

Issue #4911, opened 2019-08-02
Supershell produces redudant lines


eatkins:
On the latest develop branch, if I run test in io, I get something like this:
![demo2](https://user-images.githubusercontent.com/2658825/62339948-ab14c200-b492-11e9-80ba-630193737091.gif)

eatkins:
It happens with both iterm2 and terminal.app.

eed3si9n:
Yea. I've seen that. I wonder if there are multiple reports running simultaneously, or it's printing on top of the other so it looks that way.

SethTisue:
I don't know if this is the same or different, but with both 1.3.0-RC4 and 1.3.0-RC5 I'm seeing supershell output get mixed up with regular output, so for example over at https://github.com/SethTisue/Project-Euler :

<img width="578" alt="Screen Shot 2019-08-29 at 9 28 53 PM" src="https://user-images.githubusercontent.com/161079/63988962-24103500-caa4-11e9-816a-24c7f83ca811.png">

I'm short on time at the moment but I could investigate / follow up more next week if you want

eed3si9n:
@SethTisue might be https://github.com/scala/nanotest-strawman/issues/20

SethTisue:
looking at @eatkins's animation above, it appears to me that a bunch of tests are running in parallel, that's a good thing, and there's no issue? am I mistaken?

eed3si9n:
I'm not sure if it's multiple tests that's causing this, but I don't think there should be multiple lines of supershell countup clock displayed per task.

```
  | => io / Test / executeTest 4s
  | => io / Test / executeTest 3s
  | => io / Test / executeTest 3s
...
```

eatkins:
Yeah. That is specifically what I created this issue for. I’d expect only one `io / Test / executeTest` line.

eatkins:
@SethTisue is correct about the underlying cause though I think it's suboptimal from a ux perspective. Setting `Test / parallelExecution := false` makes the problem go away (and the tests take 3 times longer!). The `test` task delegates to a number of anonymous tasks and they all inherit the name of the calling task, `executeTest`. This is actually a general problem for the progress reports. As it currently stands, we actually report something more like utime rather than walltime in the reports:
```
Total time: 16714 ms
  io / Test / executeTests                                    : 48376 ms
  io / Compile / previousCompile                              :   159 ms
```
[...]

PR #5023, opened 2019-09-02
Name test tasks with the test name

eatkins:
Ref https://github.com/sbt/sbt/issues/4911. This names each parallel
test task with the name of the task so that supershell can display it.
It only applies for parallel tests. When run sequentially, supershell
will still display executeTests.


eatkins:
[asciicast](https://asciinema.org/a/ghMoo4BNCKMCjbykBM18hRnuy)

eed3si9n (approved):
nice!

SethTisue:
😎

PR #5027, opened 2019-09-04
1.3.x merge

eatkins:
Merges 1.3.x into develop.

PR #5037, opened 2019-09-05
Bump sbt dogfood version

eatkins:


eed3si9n (changes requested):
Duplicate of https://github.com/sbt/sbt/pull/5029

eatkins:
Do you want to merge #5029 back into develop after #5027? If so, I think we should bump the version to 1.4.0-SNAPSHOT, but otherwise that makes sense.

eed3si9n:
Yea. I think we can get develop square with 1.3.x, and then bump the version.

eatkins:
Ok. I consolidated those changes into #5027.

PR #5041, opened 2019-09-05
Https

jsoref:
The web is moving to https for security... These servers seem to support HTTPS (some even speak HTTP2 when the https urls are used).

I skipped databinder.net because while the server speaks https, http://databinder.net/repo seems to be gone.
https://github.com/sbt/sbt/blob/4c3cd4e535a89d220a0d20f360f36212ceff9b09/launch/src/main/resources/sbt.boot.properties0.7#L14

I suspect that means that someone trying to ask sbt for sbt0.7 will be unhappy, although I'm not really sure how to prove it.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -3,7 +3,7 @@ maxColumn = 100
 project.git = true
 project.excludeFilters = [ "\\Wsbt-test\\W", "\\Winput_sources\\W", "\\Wcontraband-scala\\W" ]
 
-# http://docs.scala-lang.org/style/scaladoc.html recommends the JavaDoc style.
+# https://docs.scala-lang.org/style/scaladoc.html recommends the JavaDoc style.
 # scala/scala is written that way too https://github.com/scala/scala/blob/v2.12.2/src/library/scala/Predef.scala
 docstrings = JavaDoc
 
@@ -11,7 +11,7 @@ docstrings = JavaDoc
 spaces.inImportCurlyBraces = true
 
 # This is more idiomatic Scala.
-# http://docs.scala-lang.org/style/indentation.html#methods-with-numerous-arguments
+# https://docs.scala-lang.org/style/indentation.html#methods-with-numerous-arguments
 align.openParenCallSite = false
 align.openParenDefnSite = false
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-  [StackOverflow]: http://stackoverflow.com/tags/sbt
+  [StackOverflow]: https://stackoverflow.com/tags/sbt
   [Setup]: http://www.scala-sbt.org/release/docs/Getting-Started/Setup
   [Issues]: https://github.com/sbt/sbt/issues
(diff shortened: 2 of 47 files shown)


main-command/src/main/contraband-scala/CommandSourceFormats.scala
```diff
@@ -1,5 +1,5 @@
 /**
- * This code is generated using [[http://www.scala-sbt.org/contraband/ sbt-contraband]].
+ * This code is generated using [[https://www.scala-sbt.org/contraband/ sbt-contraband]].
```
eed3si9n: This is a generated code, so I don't think we should change this here.

eed3si9n:
@jsoref Thanks!

main-command/src/main/contraband-scala/CommandSourceFormats.scala
jsoref: Do you want me to drop this file?

jsoref: I can't figure out where that's coming from :-(

jsoref: Ok, looks like it's https://github.com/sbt/contraband/blob/7f0e6c4765b6dc1ca3534f051d729d452890d5e0/library/src/test/scala/GraphQLCodecCodeGenSpec.scala#L16

jsoref:
I dropped the change to `sbt/src/sbt-test/actions/external-doc/build.sbt` because it depends on something:
```
[info] [info] Out of 1 candidates we found for org.example#b_2.12;0.1.0-SNAPSHOT in b-resolver, we are choosing b-resolver.
[info] Actual API Mappings: 
[info] 	/tmp/sbt_46fc2224/c/target/scala-2.12/classes -> http://example.org/c
[info] 	/tmp/sbt_46fc2224/global/boot/scala-2.12.8/lib/scala-library.jar -> http://www.scala-lang.org/api/2.12.8/
[info] 	/tmp/sbt_46fc2224/a-repo/org/example/a_2.12/0.1.0-SNAPSHOT/a_2.12-0.1.0-SNAPSHOT.jar -> http://example.org/a
[info] 	/tmp/sbt_46fc2224/b-repo/org.example/b_2.12/0.1.0-SNAPSHOT/jars/b_2.12.jar -> http://example.org/b
[info] Expected API Mappings: 
[info] 	/tmp/sbt_46fc2224/c/target/scala-2.12/classes -> http://example.org/c
[info] 	/tmp/sbt_46fc2224/global/boot/scala-2.12.8/lib/scala-library.jar -> https://www.scala-lang.org/api/2.12.8/
[info] 	/tmp/sbt_46fc2224/a-repo/org/example/a_2.12/0.1.0-SNAPSHOT/a_2.12-0.1.0-SNAPSHOT.jar -> http://example.org/a
[info] 	/tmp/sbt_46fc2224/b-repo/org.example/b_2.12/0.1.0-SNAPSHOT/jars/b_2.12.jar -> http://example.org/b
```

eed3si9n (approved):

PR #3093, opened 2017-04-12
[sbt 1.0] Fix logging + clean concurrency issue

eed3si9n:
Fixes #2156
Ref https://github.com/sbt/sbt/issues/2863

This is a workaround for `clean` and logging clashing when run concurrently on the same subproject. 

1. Assuming that removing a directory completes in 3s, this retries to call `IO.touch` 30 times with 100ms wait in between.
2. `clean` task is now tagged to `Tags.Clean`. There's now a concurrency rule that says tasks with the tags run as an exclusive group. (`a/clean` and `b/clean` can run together, but no other tasks)

Issue #2863, opened 2016-12-06
Logging causing "No such file or directory"


eed3si9n:
### steps

Not sure if this is reproducible, but it's the catch of the day on Travis.

```
> scripted source-dependencies/java-mixed
```

### problem

Logging causing "No such file or directory"

```
Running source-dependencies / java-mixed
[error] Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=256M; support was removed in 8.0
[info] Getting org.scala-sbt sbt 1.0.0-SNAPSHOT ...
[info] :: retrieving :: org.scala-sbt#boot-app
[info] 	confs: [default]
[info] 	54 artifacts copied, 0 already retrieved (20351kB/101ms)
[info] Getting Scala 2.11.8 (for sbt)...
[info] :: retrieving :: org.scala-sbt#boot-scala
[info] 	confs: [default]
[info] 	6 artifacts copied, 0 already retrieved (26456kB/45ms)
[info] [info] Set current project to java-mixed (in build file:/tmp/sbt_71151c2a/java-mixed/)
[info] [info] Updating {file:/tmp/sbt_71151c2a/java-mixed/}java-mixed...
[info] [info] Done updating.
[info] [info] Compiling 1 Scala source and 1 Java source to /tmp/sbt_71151c2a/java-mixed/target/scala-2.11/classes...
[info] [success] Total time: 3 s, completed Dec 6, 2016 6:12:58 AM
[info] [info] Compiling 1 Scala source to /tmp/sbt_71151c2a/java-mixed/target/scala-2.11/classes...
[info] [info] Compiling 1 Java source to /tmp/sbt_71151c2a/java-mixed/target/scala-2.11/classes...
[info] [error] /tmp/sbt_71151c2a/java-mixed/JJ.java:3: cannot find symbol
[info] [error]   symbol:   method foo(java.lang.String)
[info] [error]   location: class S
[info] [error] 		new S().foo("ahoy");
[info] [error] (compile:compileIncremental) javac returned nonzero exit code
[info] [error] Total time: 1 s, completed Dec 6, 2016 6:12:58 AM
[info] Could not create file /tmp/sbt_71151c2a/java-mixed/target/streams/$global/projectDescriptors/$global/streams/outjava.io.IOException: No such file or directory
[info] 	at sbt.internal.io.ErrorHandling$.translate(ErrorHandling.scala:10)
[info] 	at sbt.io.IO$.touch(IO.scala:167)
[...]

PR #5042, opened 2019-09-05
Create directory before touch

eatkins:
Sometimes scripted tests fail at the IO.touch line. I can't reproduce,
but I don't think it should hurt to try and create the parent directory
before touching the file.

### DIFF
--- a/tasks-standard/src/main/scala/sbt/std/Streams.scala
+++ b/tasks-standard/src/main/scala/sbt/std/Streams.scala
@@ -184,6 +184,9 @@ object Streams {
       def make[T <: Closeable](a: Key, sid: String)(f: File => T): T = synchronized {
         checkOpen()
         val file = taskDirectory(a) / sid
+        // The call to IO.createDirectory is an attempt to reduce scripted test failures in the
+        // call to IO.touch. It may be removed if it doesn't affect the scripted failures.
+        Option(file.getParentFile).foreach(IO.createDirectory)
         IO.touch(file, false)
         val t = f(file)
         opened ::= t


eed3si9n (dismissed):

tasks-standard/src/main/scala/sbt/std/Streams.scala
```diff
@@ -184,6 +184,7 @@ object Streams {
       def make[T <: Closeable](a: Key, sid: String)(f: File => T): T = synchronized {
         checkOpen()
         val file = taskDirectory(a) / sid
+        Option(file.getParentFile).foreach(IO.createDirectory)
```
eed3si9n, line 189: I feel like touch failing is a symptom of race condition around `IO.delete` and stream used for logging than touch per se, so I don't know if this would change much.

eatkins: Could very well be. Like I said, I don't really understand what causes it or how to reproduce.

eatkins: I'll add a comment.

eed3si9n: ok. How about we remember to create a GitHub issue next time we see touch failure on Travis instead then?

eatkins: This has been happening relatively frequently so it probably would make sense to just go ahead and make an issue now. We haven't made any other changes that I think would mitigate it.

eed3si9n (changes requested):
https://github.com/sbt/io/blob/7009c9232a2e18615f3eaecd1ea05b272cf68f91/io/src/main/scala/sbt/io/IO.scala#L314 creates directory in touch anyway, so I retract my previous approval.

eatkins:
I know that IO.touch creates the directory. I was trying to decouple the directory creation from the touch to get a better idea of what _exactly_ the race condition was.

eed3si9n:
The race condition if I understand correctly is the streams written by `clean` task. See whatever is under `target` after you run `clean`.

tasks-standard/src/main/scala/sbt/std/Streams.scala
eed3si9n: Do we have a stack trace to confirm which task could trigger the error?
Ref https://github.com/sbt/sbt/pull/3093
Ref https://github.com/sbt/sbt/issues/2863

Issue #5067, opened 2019-09-10
Random IOExceptions in 1.3.0


pshirshov:
This  started happening with 1.3.0-RC5 and occasionally happens in 1.3.0:

```
[error] Could not create file /Users/user/work/proj/target/streams/_global/ivyConfiguration/_global/streams/outjava.io.IOException: No such file or directory
[error] 	at sbt.internal.io.ErrorHandling$.translate(ErrorHandling.scala:20)
[error] 	at sbt.io.IO$.$anonfun$touch$2(IO.scala:315)
[error] 	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
[error] 	at sbt.internal.io.Retry$.liftedTree2$1(Retry.scala:38)
[error] 	at sbt.internal.io.Retry$.impl$1(Retry.scala:38)
[error] 	at sbt.internal.io.Retry$.apply(Retry.scala:52)
[error] 	at sbt.internal.io.Retry$.apply(Retry.scala:24)
[error] 	at sbt.io.IO$.touch(IO.scala:312)
[error] 	at sbt.std.Streams$$anon$2$$anon$3.make(Streams.scala:187)
[error] 	at sbt.std.Streams$$anon$2$$anon$3.text(Streams.scala:161)
[error] 	at sbt.std.Streams$$anon$2$$anon$3.log(Streams.scala:182)
[error] 	at sbt.std.TaskStreams.log(Streams.scala:70)
[error] 	at sbt.std.TaskStreams.log$(Streams.scala:70)
[error] 	at sbt.std.Streams$$anon$2$$anon$3.log$lzycompute(Streams.scala:141)
[error] 	at sbt.std.Streams$$anon$2$$anon$3.log(Streams.scala:141)
[error] 	at sbt.Classpaths$.$anonfun$mkIvyConfiguration$1(Defaults.scala:3218)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
[error] 	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:62)
[error] 	at sbt.std.Transform$$anon$4.work(Transform.scala:67)
[error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:280)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:19)
[error] 	at sbt.Execute.work(Execute.scala:289)
[...]

PR #5068, opened 2019-09-10
Add better error reporting

eatkins:
We have seen failures in scripted to create the output file for streams
and it has also been reported in https://github.com/sbt/sbt/issues/5067.
This commit is intended to shine light on whether the failure is caused
by the parent directory being absent, which suggests a race condition in
which its being deleted by a different thread, or if the directory does
exists but still the file can't be created. This will not actually fix
anything and is intended solely to aid debugging.


eed3si9n (approved):

eatkins:
Closing because I think I may have fixed the underlying issue in #5069.

PR #5069, opened 2019-09-10
Fixes race conditions around task logging

eatkins:
We have seen failures in scripted to create the output file for streams
and it has also been reported in https://github.com/sbt/sbt/issues/5067.
I believe this may caused by the same stream output being initialized by
multiple tasks. To fix this, I add locking on a per-file basis. There
was (and is) additional synchronization on the Streams _instance_, but
the per-file locks are stored in the Streams companion object so the
locking should be honored no matter which Streams instance calls make.

I also skip the call to IO.touch if the file already exists. I believe
that this is the common case and since IO.touch was being called with
setModified = false, it should be fine to skip the touch when the file
exists.

Prior to this change, I was able to induce the issue in #5067 in roughly
1/50 of calls to `scripted actions/cross-multiproject`. I wasn't able to
reproduce after this change.

### DIFF
--- a/tasks-standard/src/main/scala/sbt/std/Streams.scala
+++ b/tasks-standard/src/main/scala/sbt/std/Streams.scala
@@ -9,6 +9,7 @@ package sbt
 package std
 
 import java.io.{ File => _, _ }
+import java.util.concurrent.ConcurrentHashMap
 
 import sbt.internal.io.DeferredWriter
 import sbt.internal.util.ManagedLogger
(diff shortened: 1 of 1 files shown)


eed3si9n:
> I believe this may caused by the same stream output being initialized by multiple tasks. 

Interesting. I'm probably guilty of reusing streams across tasks without thinking about this.

eed3si9n (dismissed):

eatkins:
I noticed that both in actions/cross-multiproject and the stacktrace reported by @pshirshov in https://github.com/sbt/sbt/issues/5067 the failures happened in ivy related tasks. I think a number of these are dynamic tasks so I wonder if that has anything to do with it.

eatkins:
CI has been failing due to `dependency-management/cached-resolution-classifier`. I can reproduced on the `HEAD` of `origin/develop`. It's confusing though because it seems to be failing to download files from https://oss.sonatype.org/content/repositories/snapshots/com/typesafe/config/0.4.9-SNAPSHOT/ and I am able to download all of the files sbt is failing to download manually.

eatkins:
I guess this started failing yesterday: https://travis-ci.org/sbt/sbt/builds/582813802

eed3si9n:
@eatkins Let's disable dependency-management/cached-resolution-classifier for now.

eed3si9n (approved):

PR #5071, opened 2019-09-11
Only set close classloader for interactive sessions

eatkins:
The purpose of closing the classloaders is to prevent resource leaks in
long running sessions. If a user is running batch
commands, then it's fine if they want to use shutdown hooks or leak
threads.

This commit makes it so that we set the `sbt.classloader.close`
property when we enter shell if it isn't already set. I change the
default value of `SysProp.closeClassLoaders` to be false when the
variable is unset. This prevents us from closing classloaders for
non-interactive sessions.

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -882,6 +882,7 @@ object BuiltinCommands {
 
   def shell: Command = Command.command(Shell, Help.more(Shell, ShellDetailed)) { s0 =>
     import sbt.internal.{ ConsolePromptEvent, ConsoleUnpromptEvent }
+    SysProp.setCloseClassLoaders()
     val exchange = StandardMain.exchange
     val welcomeState = displayWelcomeBanner(s0)
     val s1 = exchange run welcomeState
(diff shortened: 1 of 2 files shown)


main/src/main/scala/sbt/Main.scala
```diff
@@ -882,6 +882,7 @@ object BuiltinCommands {
 
   def shell: Command = Command.command(Shell, Help.more(Shell, ShellDetailed)) { s0 =>
     import sbt.internal.{ ConsolePromptEvent, ConsoleUnpromptEvent }
+    SysProp.setCloseClassLoaders()
```
eed3si9n, line 885: Could we also provide opt-out via a ThisBuild setting? I think I've come across multiple times where this property was suggested, so it's probably a sign that it needs to be promoted as a real thing.

eed3si9n:
I don't know if we should create an inconsistency between interactive shell and batch. We'd be setting ourselves a negative incentive to use shell.

eatkins:
Fair. I was thinking about 1.3.x rather than 1.4.x. FWIW, I think it would need to be done at the global level, not ThisBuild because classloaderCache is a global state attribute.

PR #5083, opened 2019-09-16
Allow classloaders to be used after close

eatkins:
There have been a number of complaints about the new classloader closing
behavior. It is too aggressive about closing classloaders after test and
run. This commit softens the behavior by allowing a classloader to be
resurrected after close by creating a new zombie classloader that has
the same urls as the original classloader. After this commit, we always
close the classloaders when we are done, but they can still leak
file descriptors if a zombie is created.

To configure the behavior, I added a new comfiguration class:
ClassLoaderCloseMode. It has three options:
1) Ignore -- Do not emit a warning if a zombie is created
2) Warn -- Emit a warning if a zombie is created (default)
3) Strict -- Do not allow zombies to be created and throw an exception
if the loader is used after close.

I verified after this change that I could add a shutdown hook in `run`
and it would be evaluated so long as I set `bgCopyClasspath := false`.
Otherwise the needed jars were deleted before the hooks could run.

### DIFF
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -325,7 +325,7 @@ object Keys {
   val dependencyClasspathAsJars = taskKey[Classpath]("The classpath consisting of internal and external, managed and unmanaged dependencies, all as JARs.")
   val fullClasspathAsJars = taskKey[Classpath]("The exported classpath, consisting of build products and unmanaged and managed, internal and external dependencies, all as JARs.")
   val internalDependencyConfigurations = settingKey[Seq[(ProjectRef, Set[String])]]("The project configurations that this configuration depends on")
-  private[sbt] val allowZombieClassLoaders = settingKey[Boolean]("Allow a classloader that has previously been closed by `run` or `test` to continue loading classes.")
+  val allowZombieClassLoaders = settingKey[Boolean]("Allow a classloader that has previously been closed by `run` or `test` to continue loading classes.")
 
   val useCoursier = settingKey[Boolean]("Use Coursier for dependency resolution.").withRank(BSetting)
   val csrCacheDirectory = settingKey[File]("Coursier cache directory. Uses -Dsbt.coursier.home or Coursier's default.").withRank(CSetting)
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -325,6 +325,7 @@ object Keys {
   val dependencyClasspathAsJars = taskKey[Classpath]("The classpath consisting of internal and external, managed and unmanaged dependencies, all as JARs.")
   val fullClasspathAsJars = taskKey[Classpath]("The exported classpath, consisting of build products and unmanaged and managed, internal and external dependencies, all as JARs.")
   val internalDependencyConfigurations = settingKey[Seq[(ProjectRef, Set[String])]]("The project configurations that this configuration depends on")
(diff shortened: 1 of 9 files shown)


main/src/main/scala/sbt/Keys.scala
```diff
@@ -325,6 +325,7 @@ object Keys {
   val dependencyClasspathAsJars = taskKey[Classpath]("The classpath consisting of internal and external, managed and unmanaged dependencies, all as JARs.")
   val fullClasspathAsJars = taskKey[Classpath]("The exported classpath, consisting of build products and unmanaged and managed, internal and external dependencies, all as JARs.")
   val internalDependencyConfigurations = settingKey[Seq[(ProjectRef, Set[String])]]("The project configurations that this configuration depends on")
+  val classLoaderCloseMode = settingKey[ClassLoaderCloseMode]("Configures what do when a classloader is used after it has been closed.")
```
eed3si9n: As we discussed in the meeting, it might be better to stick to Boolean for setting so the build users can use machine-wide settings.

eed3si9n (changes requested):

main/src/main/scala/sbt/Keys.scala
eatkins: Yeah. I've just been stuck on what to name them.

eed3si9n:
```suggestion
  val allowZombieClassLoader = settingKey[Boolean]("Allows application to load classes after run or test task has finished; throws an error otherwise.")
```

I'd go with warning-or-strict option.

eed3si9n: Or maybe for sbt 1.3.1 it's `false` by default, like HTTP, and it lets you use zombie but just warns. By setting `ThisBuild / allowZombieClassLoader := true` you can silence the warning. So it's warning-or-ignore axis. And in some future version it becomes strict-or-ignore?

eatkins: Ok. I made that change.

main/src/main/scala/sbt/Keys.scala
```diff
@@ -325,7 +325,7 @@ object Keys {
   val dependencyClasspathAsJars = taskKey[Classpath]("The classpath consisting of internal and external, managed and unmanaged dependencies, all as JARs.")
   val fullClasspathAsJars = taskKey[Classpath]("The exported classpath, consisting of build products and unmanaged and managed, internal and external dependencies, all as JARs.")
   val internalDependencyConfigurations = settingKey[Seq[(ProjectRef, Set[String])]]("The project configurations that this configuration depends on")
-  private[sbt] val allowZombieClassLoaders = settingKey[Boolean]("Allow a classloader that has previously been closed by `run` or `test` to continue loading classes.")
+  val allowZombieClassLoaders = settingKey[Boolean]("Allow a classloader that has previously been closed by `run` or `test` to continue loading classes.")
```
eed3si9n, line 328: For me the criteria for 1.3.x is more like "is this a bug fix and can I ship it without RC cycle?" In this case, if we're going to ship it I think it should come with this key publicly available.

eatkins: I was under the impression that we promised forward compatibility within the patch release version range. If we don’t, then it makes sense to make this public.

eed3si9n: We only keep backward compatibility.

eed3si9n (approved):

main/src/main/scala/sbt/Keys.scala
eatkins: Ok. I think I misunderstood SemVer. I read the docs on that and see that binary compatible bug fixes are fine under its semantics.

PR #5086, opened 2019-09-16
Monitor project build sources

### LINKED ISSUE #5061: Build source monitoring is does not track sub/build.sbt files
The new feature in sbt 1.3.0 that build sources are monitored and sbt can optionally reload the build when sources change doesn't work with nested projects.

eatkins:
In sbt 1.3.0, we only monitor build sources in the root project
directory and the root project meta build directory. This commit adds
these inputs for each project.

Fixes https://github.com/sbt/sbt/issues/5061.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -371,13 +371,16 @@ object Defaults extends BuildCommon {
       }).value
   )
 
-  def paths = Seq(
+  // Appended to JvmPlugin.projectSettings
+  def paths: Seq[Setting[_]] = Seq(
     baseDirectory := thisProject.value.base,
     target := baseDirectory.value / "target",
     historyPath := (historyPath or target(t => Option(t / ".history"))).value,
     sourceDirectory := baseDirectory.value / "src",
     sourceManaged := crossTarget.value / "src_managed",
-    resourceManaged := crossTarget.value / "resource_managed"
+    resourceManaged := crossTarget.value / "resource_managed",
+    // Adds subproject build.sbt files to the global list of build files to monitor
+    Scope.Global / checkBuildSources / fileInputs += baseDirectory.value.toGlob / "*.sbt"
   )
 
   lazy val configPaths = sourceConfigPaths ++ resourceConfigPaths ++ outputConfigPaths
--- a/main/src/main/scala/sbt/internal/nio/CheckBuildSources.scala
+++ b/main/src/main/scala/sbt/internal/nio/CheckBuildSources.scala
@@ -13,7 +13,10 @@ import sbt.SlashSyntax0._
 import sbt.io.syntax._
 import sbt.nio.FileChanges
 import sbt.nio.Keys._
-import sbt.nio.file.{ Glob, RecursiveGlob }
+import sbt.nio.file.{ Glob, ** }
+import sbt.nio.file.syntax._
+
+import scala.annotation.tailrec
 
 private[sbt] object CheckBuildSources {
   private[sbt] def needReloadImpl: Def.Initialize[Task[StateTransform]] = Def.task {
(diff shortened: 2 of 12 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -371,13 +370,14 @@ object Defaults extends BuildCommon {
       }).value
   )
 
-  def paths = Seq(
+  def paths: Seq[Setting[_]] = Seq(
```
eed3si9n, line 375:
```suggestion
  // included into JvmPlugin.projectSettings
  def paths: Seq[Setting[_]] = Seq(
```

main/src/main/scala/sbt/internal/nio/CheckBuildSources.scala
```diff
@@ -59,7 +58,7 @@ private[sbt] object CheckBuildSources {
   }
   private[sbt] def buildSourceFileInputs: Def.Initialize[Seq[Glob]] = Def.setting {
     if (onChangedBuildSource.value != IgnoreSourceChanges) {
-      val baseDir = (LocalRootProject / baseDirectory).value
+      val baseDir = baseDirectory.value
       val sourceFilter = "*.{sbt,scala,java}"
       val projectDir = baseDir / "project"
```
eed3si9n, line 66: Wouldn't this end up tracking `sub/project/*.scala`? Maybe metabuild source inputs needs to be split up, and added back at global level.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -371,13 +370,14 @@ object Defaults extends BuildCommon {
...
     baseDirectory := thisProject.value.base,
     target := baseDirectory.value / "target",
     historyPath := (historyPath or target(t => Option(t / ".history"))).value,
     sourceDirectory := baseDirectory.value / "src",
     sourceManaged := crossTarget.value / "src_managed",
-    resourceManaged := crossTarget.value / "resource_managed"
+    resourceManaged := crossTarget.value / "resource_managed",
+    Scope.Global / checkBuildSources / fileInputs ++= CheckBuildSources.buildSourceFileInputs.value,
```
eed3si9n:
```suggestion
    // lhs generates *.sbt files that needs to be tracked per subprojects
    Scope.Global / checkBuildSources / fileInputs ++= CheckBuildSources.buildSourceFileInputs.value,
```

eed3si9n (changes requested):
some comments.

main/src/main/scala/sbt/internal/nio/CheckBuildSources.scala
eatkins: Yeah, that's intentional. In the project directory, top level scala files are allowed in compile.

eatkins: It's common for builds to have something like project/Dependencies.scala and then refer to object Dependencies in the top level build.sbt.

eed3si9n: I understand you need to track `project/*.scala`. There are two axes to traverse. One is subprojects, the other is metabuild. Currently those two concerns are mixed, so you end up with `sub/project/*.scala`.

eatkins: Right, but why can't my subproject also have a metabuild?

eed3si9n: A metabuild can have subprojects (, which I believe this code still doesn't track), but a metabuild of a subproject doesn't make sense since "subprojects" are members of the proper build, and only a build can have its meta-build.

eatkins: I don't know the answer, but that appears to be true. I guess I misunderstood how recursion works with builds.

eed3si9n: To put it into examples:

```
project/D.scala                // track
project/plugins.sbt            // track
project/foo/A.scala            // track (bonus points)
project/src/main/scala/B.scala // track
project/project/C.scala        // track
project/project/plugins.sbt    // track
build.sbt                      // track
sub/build.sbt                  // track
sub/project/E.scala            // do not track
```

eatkins: I can't get the bonus points without degrading performance because searching for all of the scala files would require listing the target directory which can be extremely slow. Otherwise, I think that this PR now handles the other cases.

main/src/main/scala/sbt/Defaults.scala
eatkins: Applied with slight rewording.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -371,13 +371,16 @@ object Defaults extends BuildCommon {
...
+  def paths: Seq[Setting[_]] = Seq(
     baseDirectory := thisProject.value.base,
     target := baseDirectory.value / "target",
     historyPath := (historyPath or target(t => Option(t / ".history"))).value,
     sourceDirectory := baseDirectory.value / "src",
     sourceManaged := crossTarget.value / "src_managed",
-    resourceManaged := crossTarget.value / "resource_managed"
+    resourceManaged := crossTarget.value / "resource_managed",
+    // Adds subproject build.sbt files to the global list of build files to monitor
+    Scope.Global / checkBuildSources / fileInputs += baseDirectory.value.toGlob / "*.sbt"
```
eed3si9n, line 383: 👍

eed3si9n (dismissed):
LGTM. thanks @eatkins

eed3si9n (approved):

PR #5092, opened 2019-09-17
Add warning if old launcher is used

eatkins:
In order to isolate the sbt test launcher interface jar so that the run
and test classpaths don't get contaminated with the sbt build jars, sbt
1.3.0 creates a new classloader from the loader that is passed in by the
launcher unless the launcher provided classloader has a specific format.
Most things work with the old launcher, but there was a report that
consoleProject did not work. To try and ward off similar obscure bugs,
we will now issue a warning if a pre 1.3.0 launcher is used to launch
sbt >= 1.3.0.

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -64,6 +64,11 @@ private[sbt] object xMain {
           Seq(defaults, early),
           runEarly(DefaultsCommand) :: runEarly(InitCommand) :: BootCommand :: Nil
         )
+        if (xMain.getClass.getClassLoader.toString == "SbtMetaBuildClassLoader") {
+          val msg = s"sbt was launched with an old version of sbt-launcher. Please upgrade the " +
+            "launcher version (see https://www.scala-sbt.org/1.x/docs/Setup.html for instructions)."
+          state.globalLogging.full.warn(msg)
+        }
         StandardMain.runManaged(state)
       }
     } finally {


main/src/main/scala/sbt/Main.scala
```diff
@@ -64,6 +64,11 @@ private[sbt] object xMain {
           Seq(defaults, early),
           runEarly(DefaultsCommand) :: runEarly(InitCommand) :: BootCommand :: Nil
         )
+        if (xMain.getClass.getClassLoader.toString == "SbtMetaBuildClassLoader") {
+          val msg = s"sbt was launched with an old version of sbt-launcher. Please upgrade the " +
```
eed3si9n, line 68: Normally sbt is backward and forward compatible against the launcher, so I think there should be a really good reason to print warnings here. I am kind of ok with ppl upgrading to new launcher or script only if they run into an issue.

eatkins: I would normally agree with that, except that failure in something like `consoleProject` is extremely subtle. It was almost pure luck that I was able to debug that when it came up in the gitter channel.

eed3si9n: Currently the release note says you can upgrade by changing `build.properties` file - https://www.lightbend.com/blog/sbt-1.3.0-release
If we're requiring the script and the launcher to be upgraded, then we need to probably start from there, and also reword that this is a special occurrence. There's been ideas about letting ppl write JVM props in `build.properties`, so we might want to wait for that to come out. (1.4.0, maybe?)

eatkins: The release notes also says right underneath:
> To get other benefits to launching script, please upgrade the sbt script following the setup instruction.

I think we could just strengthen that wording. The truth is that the old launcher does work for probably 99% of use cases, it's just very subtle to debug when it doesn't work.

eed3si9n: Maybe we can say we _strongly recommend_ upgrading for feature releases (sbt 1.2.x -> 1.3.0).

eatkins: Do you mean in the warning that's printed by sbt in this PR or in the release notes?

eed3si9n: I meant for the release note, which I did - https://www.lightbend.com/blog/sbt-1.3.0-release

> To get other benefits to launching script and sbt launcher, we strongly recommend upgrading the sbt script following the setup instruction.

PR #5096, opened 2019-09-18
Preload a number of classes in the background

eatkins:
I was looking into sbt start up time and in profiling was able to
identify a number of classloading bottlenecks. To speed up
initialization, we can preload those classes in the background. I saw
average speedups of roughly .75 seconds after this change. Also, the `time`
command would consistently report cpu system time very close to 400% and
I have 4 cores on my laptop. With 1.3.0 it would be more like 350%.

### DIFF
--- a/main/src/main/scala/sbt/internal/XMainConfiguration.scala
+++ b/main/src/main/scala/sbt/internal/XMainConfiguration.scala
@@ -10,9 +10,42 @@ package sbt.internal
 import java.io.File
 import java.lang.reflect.InvocationTargetException
 import java.net.{ URL, URLClassLoader }
+import java.util.concurrent.{ ExecutorService, Executors }
 import java.util.regex.Pattern
 
+import sbt.plugins.{ CorePlugin, IvyPlugin, JvmPlugin }
+import sbt.util.LogExchange
 import xsbti._
+
+private[internal] object ClassLoaderWarmup {
+  def warmup(): Unit = {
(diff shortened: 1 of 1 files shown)


eed3si9n (dismissed):

main/src/main/scala/sbt/internal/XMainConfiguration.scala
```diff
@@ -10,10 +10,43 @@ package sbt.internal
...
+      submit(IvyPlugin.projectSettings)
+      submit(JvmPlugin.projectSettings)
+      submit(() => {
+        try {
+          val clazz = Class.forName("scala.reflect.runtime.package$")
+          clazz.getMethod("universe").invoke(clazz.getField("MODULE$").get(null))
+        } catch {
+          case _: Exception =>
+        }
+        executorService.shutdown()
```
dwijnand, line 44: Btw, this needn't be here, within the job submission.  I'm assuming the logic was all the other jobs will be done before the runtime universe is done initialising.

Instead, it could be just after the last job submission (i.e. 1 line down here), because `shutdown` works by triggering the service to not accept any new submissions, but completing all the ones in flight.  (`awaitTermination` does the same but then also blocks, waiting for the shutdown to complete.)

Not worth PRing, IMO, as nothing of substance will be gained doing the shutdown earlier in the main thread instead of later in this warm up runtime universe thread.  Just FYI.

PR #5105, opened 2019-09-19
[1.3.x] preload classloaders backport

eatkins:
Backports #5096 to 1.3.x.


eed3si9n:
I am actually not sure if we should do this for 1.3.1, or wait till 1.4.0.

eatkins:
I don’t have a strong preference. I can just leave this for now and maybe it goes into 1.3.2 or maybe it waits until 1.4.0. It mainly benefits small projects anyway since the time saved is roughly constant.

eed3si9n:
Yea. I'd rather ship monthly milestones out from 1.4.x than keep putting perf improvements into 1.3.x.

eatkins:
Ok. I'll close this then.

PR #5112, opened 2019-09-21
Throw error if you run sbt from /

### LINKED ISSUE #1458: OutOfMemoryError if running sbt from the root directory
On a Linux machine, using sbt 0.3.15:

cd /  (yes, the root)
sbt about

```
java.lang.OutOfMemoryError: GC overhead limit exceeded
  at java.io.UnixFileSystem.resolve(UnixFileSystem.java:108)
  at java.io.File.<init>(File.java:262)
  at java.io.File.listFiles(File.java:1284)
  at sbt.FilterFiles.handleFile(Path.scala:182)
  at sbt.DescendantOrSelfPathFinder.sbt$DescendantOrSelfPathFinder$$handleFileDescendant(Path.scala:198)
  at sbt.DescendantOrSelfPathFinder$$anonfun$sbt$DescendantOrSelfPathFinder$$handleFileDescendant$1.apply(Path.scala:200)
  at sbt.DescendantOrSelfPathFinder$$anonfun$sbt$DescendantOrSelfPathFinder$$handleFileDescendant$1.apply(Path.scala:199)
  at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
  at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:108)
  at sbt.DescendantOrSelfPathFinder.sbt$DescendantOrSelfPathFinder$$handleFileDescendant(Path.scala:199)
  at sbt.DescendantOrSelfPathFinder$$anonfun$sbt$DescendantOrSelfPathFinder$$handleFileDescendant$1.apply(Path.scala:200)
  at sbt.DescendantOrSelfPathFinder$$anonfun$sbt$DescendantOrSelfPathFinder$$handleFileDescendant$1.apply(Path.scala:199)
  at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
  at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:108)
  at sbt.DescendantOrSelfPathFinder.sbt$DescendantOrSelfPathFinder$$handleFileDescendant(Path.scala:199)
  at sbt.DescendantOrSelfPathFinder$$anonfun$sbt$DescendantOrSelfPathFinder$$handleFileDescendant$1.apply(Path.scala:200)
  at sbt.DescendantOrSelfPathFinder$$anonfun$sbt$DescendantOrSelfPathFinder$$handleFileDescendant$1.apply(Path.scala:199)
  at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
  at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:108)
  at sbt.DescendantOrSelfPathFinder.sbt$DescendantOrSelfPathFinder$$handleFileDescendant(Path.scala:199)
  at 
....
```

eed3si9n:
Fixes #1458

Running sbt from `/` results to sbt getting stuck trying to load the directories recursively, and eventually erroring with a java.lang.OutOfMemoryError (after freezing for a long time) even on an Alpine container.

To prevent it, this adds a check to see if the absolute path is `/` or not.

```
/ $ sbt -Dsbt.version=1.4.0-SNAPSHOT
[error] java.lang.IllegalStateException: cannot run sbt from root directory without -Dsbt.rootdir=true; see sbt/sbt#1458
```

Note: To opt into running sbt at `/`, run with `-Dsbt.rootdir=true`.

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -9,7 +9,7 @@ package sbt
 
 import java.io.{ File, IOException }
 import java.net.URI
-import java.nio.file.{ FileAlreadyExistsException, Files }
+import java.nio.file.{ FileAlreadyExistsException, Files, FileSystems }
 import java.util.concurrent.ForkJoinPool
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.{ Locale, Properties }
@@ -940,14 +940,33 @@ object BuiltinCommands {
     state.remainingCommands exists (_.commandLine == TemplateCommand)
 
   private def writeSbtVersion(state: State) =
-    if (SysProp.genBuildProps && !intendsToInvokeNew(state))
+    if (SysProp.genBuildProps && !intendsToInvokeNew(state)) {
       writeSbtVersionUnconditionally(state)
+    }
+
+  private def checkRoot(state: State): Unit =
+    if (SysProp.allowRootDir) ()
+    else {
+      val baseDir = state.baseDir
+      import scala.collection.JavaConverters._
+      // this should return / on Unix and C:\ for Windows.
+      val rootOpt = FileSystems.getDefault.getRootDirectories.asScala.toList.headOption
+      rootOpt foreach { root =>
+        if (baseDir.getAbsolutePath == root.toString) {
+          throw new IllegalStateException(
+            "cannot run sbt from root directory without -Dsbt.rootdir=true; see sbt/sbt#1458"
+          )
+        }
+      }
+    }
 
   private def WriteSbtVersion = "writeSbtVersion"
 
   private def writeSbtVersion: Command =
     Command.command(WriteSbtVersion) { state =>
-      writeSbtVersion(state); state
+      checkRoot(state)
+      writeSbtVersion(state)
+      state
     }
 
   private def intendsToInvokeCompile(state: State) =
--- a/main/src/main/scala/sbt/internal/SysProp.scala
+++ b/main/src/main/scala/sbt/internal/SysProp.scala
@@ -69,6 +69,7 @@ object SysProp {
   def traces: Boolean = getOrFalse("sbt.traces")
   def client: Boolean = getOrFalse("sbt.client")
   def ci: Boolean = getOrFalse("sbt.ci")
+  def allowRootDir: Boolean = getOrFalse("sbt.rootdir")
 
   def watchMode: String =
     sys.props.get("sbt.watch.mode").getOrElse("auto")


main/src/main/scala/sbt/Main.scala
```diff
@@ -940,14 +940,28 @@ object BuiltinCommands {
...
-    if (SysProp.genBuildProps && !intendsToInvokeNew(state))
+    if (SysProp.genBuildProps && !intendsToInvokeNew(state)) {
       writeSbtVersionUnconditionally(state)
+    }
+
+  private def checkRoot(state: State): Unit =
+    if (SysProp.allowRoot) ()
+    else {
+      val baseDir = state.baseDir
+      if (baseDir.getAbsolutePath == "/") {
```
eatkins: Is there an equivalent check for windows? You probably don't want to start sbt in `C:\` either. Perhaps this should be something like
```scala
  val basePath = state.baseDir.toPath.toAbsolutePath
  val roots = FileSystems.getDefault.getRootDirectories.asScala.toSet
  if (!roots.contains(basePath)) { ...
```

main/src/main/scala/sbt/internal/SysProp.scala
```diff
@@ -69,6 +69,7 @@ object SysProp {
   def traces: Boolean = getOrFalse("sbt.traces")
   def client: Boolean = getOrFalse("sbt.client")
   def ci: Boolean = getOrFalse("sbt.ci")
+  def allowRoot: Boolean = getOrFalse("sbt.root")
```
eatkins: This is a little ambiguous. It could mean the root directory or the root user.

eatkins (changes requested):

main/src/main/scala/sbt/Main.scala
eed3si9n: We probably want to prevent `C:\` but `D:\` should be ok?

eatkins: I don't know enough about windows to answer that. I guess you could probably take the head of `getRootDirectories` or filter for `C:\` on windows.

main/src/main/scala/sbt/internal/SysProp.scala
eed3si9n: `sbt -Dsbt.rootdirectory=true`?

main/src/main/scala/sbt/Main.scala
eed3si9n: Addressed in 563bcb9

main/src/main/scala/sbt/internal/SysProp.scala
eed3si9n: Changed to -Dsbt.rootdir in 563bcb9

PR #5125, opened 2019-09-24
Progress improvements

eatkins:
This PR makes a few improvements to task progress rendering. It complements https://github.com/sbt/util/pull/223. After https://github.com/sbt/util/pull/223, progress lines take terminal width into account, but there is some flickering if a dynamic task is used, which might create a new task progress thread:
util-bump
The first change limits sbt to having a single progress thread in most cases, which eliminates the flickering in my example:
one-thread
It also doesn't really make sense to have dummy tasks like `streams-manager` appearing in the task progress, so I filter them out in the second commit:
no-dummy

### DIFF
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -224,29 +224,33 @@ object EvaluateTask {
       structure: BuildStructure,
       state: State
   ): ExecuteProgress[Task] = {
-    val maker: Seq[Keys.TaskProgress] = getSetting(
-      Keys.progressReports,
-      Seq(),
-      extracted,
-      structure
-    )
-    val progressReporter = extracted.get(progressState in ThisBuild).map { ps =>
-      ps.reset()
-      ConsoleAppender.setShowProgress(true)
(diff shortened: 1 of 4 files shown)


eed3si9n:
Travis CI says:

```
[info] PluginCommandTest
[info] The `plugin` command should
[error]   ! should work for plugins within nested in one package
[error]    java.lang.RuntimeException: {<this>} / ./ ./ progressState is undefined. (file:1)
[error]   ! should work for plugins nested more than one package
[error]    java.lang.RuntimeException: {<this>} / ./ ./ progressState is undefined. (file:1)
[error]   ! suggest a plugin when given an incorrect plugin with a similar name
[error]    java.lang.RuntimeException: {<this>} / ./ ./ progressState is undefined. (file:1)
```

eatkins:
Should be fixed by https://github.com/sbt/sbt/compare/6511be68a6ec5ac379d8f567f3599a59aee69a9d..edd21b0ec8e707f61f0dd3e22fd790ac94ea7fa5

eed3si9n (approved):
LGTM

PR #5126, opened 2019-09-24
Close classloader in fgRun/fgRunMain

### LINKED ISSUE #5124: [1.3.2] fgRun performance regression
## steps
sbt version: 1.3.2
build.sbt
```
scalaVersion := "2.13.1"
turbo := true
libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value
```

main.scala
```
object Main {
    val start = System.currentTimeMillis()
    println(implicitly[scala.reflect.runtime.universe.TypeTag[Int]])
    println(System.currentTimeMillis() - start)
  }
}
```

## problem
running this with `fgRun` takes around 1 second

## expectation
~1ms run (like with 1.3.0)

## notes
plain `run` now runs faster than `fgRun`, even though `bgRun` is as slow as it was before

eatkins:
It is necessary to close the run classloader in order to reuse the
dependency layer in the next run.

Fixes #5124.

### DIFF
--- a/run/src/main/scala/sbt/Run.scala
+++ b/run/src/main/scala/sbt/Run.scala
@@ -12,7 +12,7 @@ import java.lang.reflect.Method
 import java.lang.reflect.Modifier.{ isPublic, isStatic }
 
 import sbt.internal.inc.ScalaInstance
-import sbt.internal.inc.classpath.ClasspathUtilities
+import sbt.internal.inc.classpath.{ ClasspathFilter, ClasspathUtilities }
 import sbt.internal.util.MessageOnlyException
 import sbt.io.Path
 import sbt.util.Logger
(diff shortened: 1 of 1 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1510,7 +1510,15 @@ object Defaults extends BuildCommon {
       loadForParser(discoveredMainClasses)((s, names) => runMainParser(s, names getOrElse Nil))
     Def.inputTask {
       val (mainClass, args) = parser.parsed
-      scalaRun.value.run(mainClass, data(classpath.value), args, streams.value.log).get
+      val cp = data(classpath.value)
+      val logger = streams.value.log
+      scalaRun.value match {
```
eed3si9n: Could we push this logic into Run?

eatkins: Yes. That's a better solution.

eed3si9n (approved):

Issue #5116, opened 2019-09-23
[1.3.x turbo] deadlock in sbt internal classloaders when running tests


bjaglin:
## steps
* sbt version
  - reproducable on 1.3.0 & 1.3.2
  - I do not remember seeing it on early 1.3 RCs (before https://github.com/sbt/sbt/commit/9c821b7b13eb14d20eedf51efa28a8f82bdb22e6#diff-9426c31e9b78333e024e94f62b057583 was introduced) but I can't tell for sure
  - never reproduced on 1.2.x
* turbo mode enabled
* default test settings: no fork, parallel execution

## problem

`sbt test` hangs sometimes  (~10% of the builds on our 16 vCPU CI hosts)

## expectation

`sbt test` fails or succeeds, but does not hang

## notes

I haven't been able to extract a test suite to exhibit the problem, and this big multi-module project is not open-source... Here are the relevant bits of a thread dump when sbt hangs:

```
2019-09-23 09:34:28 Found one Java-level deadlock:
2019-09-23 09:34:28 =============================
2019-09-23 09:34:28 "pool-22-thread-2-ScalaTest-running-TeadsCommitmentSpec":
2019-09-23 09:34:28   waiting to lock monitor 0x00007f46a43c7600 (object 0x00000007e7d80018, a sbt.internal.ReverseLookupClassLoaderHolder$BottomClassLoader),
2019-09-23 09:34:28   which is held by "pool-22-thread-12-ScalaTest-running-ContractSpec"
2019-09-23 09:34:28 "pool-22-thread-12-ScalaTest-running-ContractSpec":
2019-09-23 09:34:28   waiting to lock monitor 0x00007f46a43c8200 (object 0x00000007e7b00090, a sbt.internal.ReverseLookupClassLoaderHolder$ReverseLookupClassLoader),
2019-09-23 09:34:28   which is held by "pool-22-thread-6-ScalaTest-running-ContractPartiesSpec"
2019-09-23 09:34:28 "pool-22-thread-6-ScalaTest-running-ContractPartiesSpec":
2019-09-23 09:34:28   waiting to lock monitor 0x00007f47d800b800 (object 0x00000007e7b002f0, a java.lang.Object),
2019-09-23 09:34:28   which is held by "pool-22-thread-12-ScalaTest-running-ContractSpec"


[...]

PR #5131, opened 2019-09-25
Make internal classloaders parallel capable

eatkins:
Because classloading with a reverse lookup classloader is not strictly
hierarchical, it is not safe to treat the BottomClassLoader as parallel
compatibile. To address this, I rework things so that the
BottomClassLoader actually synchronizes in the loadClass implementation.
I also updated the java serialization scripted test in a way that can
trigger a deadlock without synchronization.

This had no measurable performance impact when I reran
https://github.com/eatkins/scala-build-watch-performance

### DIFF
--- a/main/src/main/java/sbt/internal/BottomClassLoader.java
+++ b/main/src/main/java/sbt/internal/BottomClassLoader.java
@@ -1,0 +1,76 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.internal;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URL;
+import sbt.util.Logger;
+
+/**
+ * The bottom most layer in our layering hierarchy. This layer should never be cached. The
+ * dependency layer may need access to classes only available at this layer using java reflection.
+ * To make this work, we register this loader with the parent in its constructor. We also add the
+ * lookupClass method which gives ReverseLookupClassLoader a public interface to findClass.
+ *
+ * <p>To improve performance, when loading classes from the parent, we call the loadClass method
+ * with the reverseLookup flag set to false. This prevents the ReverseLookupClassLoader from trying
+ * to call back into this loader when it can't find a particular class.
+ */
+final class BottomClassLoader extends ManagedClassLoader {
+  private final ReverseLookupClassLoaderHolder holder;
+  private final ReverseLookupClassLoader parent;
+  private final ClassLoadingLock classLoadingLock = new ClassLoadingLock();
+
+  BottomClassLoader(
+      final ReverseLookupClassLoaderHolder holder,
+      final URL[] dynamicClasspath,
+      final ReverseLookupClassLoader reverseLookupClassLoader,
+      final File tempDir,
+      final boolean allowZombies,
+      final Logger logger) {
+    super(dynamicClasspath, reverseLookupClassLoader, allowZombies, logger);
+    setTempDir(tempDir);
+    this.holder = holder;
+    this.parent = reverseLookupClassLoader;
+    parent.setDescendant(this);
+  }
+
+  static {
+    ClassLoader.registerAsParallelCapable();
+  }
+
+  @Override
+  public Class<?> findClass(final String name) throws ClassNotFoundException {
+    return classLoadingLock.withLock(
+        name,
+        () -> {
+          final Class<?> prev = findLoadedClass(name);
+          if (prev != null) return prev;
+          return super.findClass(name);
+        });
+  }
+
+  @Override
+  protected Class<?> loadClass(final String name, final boolean resolve)
+      throws ClassNotFoundException {
+    try {
+      return parent.loadClass(name, resolve, false);
+    } catch (final ClassNotFoundException e) {
+      final Class<?> clazz = findClass(name);
+      if (resolve) resolveClass(clazz);
+      return clazz;
+    }
+  }
+
+  @Override
+  public void close() throws IOException {
+    holder.checkin(parent);
+    super.close();
+  }
+}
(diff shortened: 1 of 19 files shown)


eatkins:
May fix #5116.

eed3si9n:
To test the concurrency of `test` task, it might be interesting to add a bunch of long running tests that holds on to various loaded things. For example, in the originally reported issue it's trying to initialize `cats.data.NonEmptyList` companion object.

eatkins:
That could be interesting. This is actually a pretty good test because it actually is failing quite reliably. I didn't test it extensively enough before pushing.

I'm starting to think that java serialization is maybe a lost cause and we should just throw a very specific exception that fails the test and explains that either the code needs to be refactored to not use java serialization or change the layering strategy. If everything is single-threaded, the reverse lookup seems ok, but I'm worried that it's going to take forever to get it right in the presence of concurrency. You already lose the benefit of turbo mode because we have to invalidate the dependency layer when we use java serialization so it's probably better to just fail hard.

eed3si9n:
Is #5116 related to serialization?

eed3si9n:
Could we add the repro in https://github.com/sbt/sbt/issues/5116#issuecomment-535184386?

eatkins:
The repro is the java-serialization scripted test.

sbt/src/sbt-test/classloader-cache/java-serialization/descendant/src/test/scala/test/ReflectionTest.scala
```diff
@@ -1,12 +1,32 @@
...
+  val latch = new CountDownLatch(procs)
+  (1 to procs).foreach { i =>
+    new Thread() {
+      setDaemon(true)
+      start()
+      override def run(): Unit = {
+        initLatch.countDown()
+        initLatch.await(5, TimeUnit.SECONDS)
+        val className = if (i % 2 == 0) "test.Foo" else "test.Bar"
+        loader.loadClass(className)
```
eatkins, line 20: `test.Foo` extends `test.Bar` so we should end up concurrently trying to load `test.Bar`. This can cause deadlock if the classloaders are not parallel capable.

main/src/main/java/sbt/internal/BottomClassLoader.java
```diff
@@ -0,0 +1,101 @@
...
+    this.holder = holder;
+    this.parent = reverseLookupClassLoader;
+    parent.setDescendant(this);
+  }
+
+  static {
+    ClassLoader.registerAsParallelCapable();
+  }
+
+  final Class<?> reverseLoad(final String name, final boolean resolve)
```
eed3si9n: What does reverseLoad mean?

eed3si9n: Since loading is `String => Class`, a reverse would be `Class => String`.

Is `loadOnceForParent(String, Boolean): Class` what's going on?

eatkins: I chose reverseLoad to mean that we were traversing the classloading hierarchy in the _reverse_ direction. Normally a child loads a class from it’s parent but here the parent tries to load a class from a child.

eed3si9n: Right, and the `BoomClassLoader` is the child who gets sacrificed by the parent right? So I think `loadOnce` or `loadForParent` might be a better name with comment saying it gets called by the parent.

Also instead of reversing, would it make sense to have two parents, like a primary parent, and secondary parent? If the primary fails for whatever reason lazily create a flat `ClassLoader`?

eatkins: Instead of renaming `reverseLoad`, I just removed it and made `findClass` public in `BottomClassLoader`.

eatkins: > Also instead of reversing, would it make sense to have two parents, like a primary parent, and secondary parent? If the primary fails for whatever reason lazily create a flat ClassLoader?

I don't think I understand this idea. Is the idea that we'd create a dual loader where the primary classloader has the dependency classpath and the secondary loader has the full classpath?

eed3si9n: Yea. So in that world, there's no traversing back to child, it would just fall back a flat secondary parent.

eatkins: ~~That would not work because the classes loaded by the secondary parent cannot be used by the child.~~
The reason for fall back to the child is specifically for when java reflection is used in the parent layer to access classes that are only visible to the child. Loading those classes in alternate secondary parent loader creates an alternate classloading hierarchy which would be incompatible with the normal hierarchy and would almost surely cause frequent incompatible class exceptions.

eatkins: What I've been trying to do is isolate things so that there is at most one classloader that can load each class on the classpath (this could be violated if multipe jars define the same class). Dual loaders make it possible for multiple classloaders to load the same class which makes it much easier for something to go wrong.

eatkins: To make it concrete, the java-serialization scripted test defines an object that looks like this:
```
object Reflection {
  def roundTrip[A](a: A): A = {
    val baos = new ByteArrayOutputStream()
    val oos = new ObjectOutputStream(baos)
    oos.writeObject(a)
    oos.close()
    val bais = new ByteArrayInputStream(baos.toByteArray())
    val ois = new ObjectInputStream(bais)
    try ois.readObject().asInstanceOf[A]
    finally ois.close()
  }
}
```
The call to `ois.readObject` will eventually lead to a call to `Class.forName` with the class name read from the object stream. Suppose that we invoke this method with a class `A` that is defined in the project classpath. If there were a secondary flatloader that were used to load `A`, then the `asInstanceOf[A]` would cause a `ClassCastException`. By looking back into the child, however, we load the same version of `A` that was passed in as input.

eed3si9n: > Loading those classes in alternate secondary parent loader creates an alternate classloading hierarchy which would be incompatible with the normal hierarchy and would almost surely cause frequent incompatible class exceptions.

Yea. That makes sense.

eed3si9n (dismissed):
LGTM pending CI

eatkins:
Passing now.

PR #5152, opened 2019-10-03
Switch to xenial image

eed3si9n:
The majority of Travis CI builds have probalby migrated to using Xenial now; it's better to migrate with the pack to get better hardware support etc.


eatkins (approved):
👍

PR #5153, opened 2019-10-04
build linting to warn on unused settings during reload

### LINKED ISSUE #3183: Warn on unused settings (or tasks) when they are marked secondary
## problem

1. sbt will silently ignore unused settings.
2. Displaying warning for all unused tasks and setting will catch top level keys like `clean`.

```scala
scalacOptions in compile += "-X"
```

## expectation

1. I can mark keys to be top level or secondary. (`keyUsage in (ThisBuild, foo) := Usage.Secondary`, probably tasks default to top level, and settings default to secondary.)
2. There's a `buildLint` command that warns on unused setting/task expressions with secondary keys provided by `build.sbt`.
3. The warning takes in account of the scope delegation rules.
4. Debug mode to display unused setting/task expressions provided by plugins, including the Defaults.

```scala
scalacOptions in compile += "-X"
^ 
[warning] Unused key: projA/*:compile::scalacOptions was not referenced by other settings/tasks.
```

## notes

Many have brought up the idea of warning on unused. Here's previous thread on [unused settings](https://groups.google.com/d/msg/simple-build-tool/N-nCv5v_O1M/ZPPH9Y-RXlcJ) from 2011, which shows 58 unused settings.

eed3si9n:
Fixes #3183

This implements an input task `lintBuild` that checks for unused settings/tasks. Because most settings are on the intermediary to other settings/tasks, they are included into the linting by default. The notable exceptions are settings used exclusively by a command. To opt-out, you can either append it to `Global / excludeLintKeys` or set the rank to invisible.

On the other hand, many tasks are on the leaf (called by human), so task keys are excluded from linting by default. However there are notable tasks that trip up users, so they are opted-in using `Global / includeLintKeys`.

`lintBuild` is also implemented as a function that takes a `State`, and I'm calling that during `reload` now.

lintBuild

### credits

The core of the idea is straight from the snippet proposed by @havocp in 2011 [unused settings](https://groups.google.com/d/msg/simple-build-tool/N-nCv5v_O1M/ZPPH9Y-RXlcJ) thread that identified 58 unused settings. This filters the candidates down to locally defines settings only (also an idea discussed in the thread).

### DIFF
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -491,9 +491,9 @@ object Keys {
   private[sbt] val postProgressReports = settingKey[Unit]("Internally used to modify logger.").withRank(DTask)
   @deprecated("No longer used", "1.3.0")
   private[sbt] val executeProgress = settingKey[State => TaskProgress]("Experimental task execution listener.").withRank(DTask)
-  val lintBuild = inputKey[Unit]("Stop a background job by providing its ID.")
-  val excludeLintKeys = settingKey[Set[Def.KeyedInitialize[_]]]("Keys excluded from lintBuild task")
-  val includeLintKeys = settingKey[Set[Def.KeyedInitialize[_]]]("Task keys that are included into lintBuild task")
+  val lintUnused = inputKey[Unit]("Check for keys unused by other settings and tasks.")
+  val excludeLintKeys = settingKey[Set[Def.KeyedInitialize[_]]]("Keys excluded from lintUnused task")
+  val includeLintKeys = settingKey[Set[Def.KeyedInitialize[_]]]("Task keys that are included into lintUnused task")
 
   val stateStreams = AttributeKey[Streams]("stateStreams", "Streams manager, which provides streams for different contexts.  Setting this on State will override the default Streams implementation.")
   val resolvedScoped = Def.resolvedScoped
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -491,6 +491,9 @@ object Keys {
   private[sbt] val postProgressReports = settingKey[Unit]("Internally used to modify logger.").withRank(DTask)
   @deprecated("No longer used", "1.3.0")
   private[sbt] val executeProgress = settingKey[State => TaskProgress]("Experimental task execution listener.").withRank(DTask)
+  val lintBuild = inputKey[Unit]("Stop a background job by providing its ID.")
+  val excludeLintKeys = settingKey[Set[Def.KeyedInitialize[_]]]("Keys excluded from lintBuild task")
+  val includeLintKeys = settingKey[Set[Def.KeyedInitialize[_]]]("Task keys that are included into lintBuild task")
 
   val stateStreams = AttributeKey[Streams]("stateStreams", "Streams manager, which provides streams for different contexts.  Setting this on State will override the default Streams implementation.")
   val resolvedScoped = Def.resolvedScoped
--- a/main/src/main/scala/sbt/internal/LintBuild.scala
+++ b/main/src/main/scala/sbt/internal/LintBuild.scala
@@ -1,0 +1,145 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt
+package internal
+
+import Keys._
+import Def.{ Setting, ScopedKey }
+import sbt.internal.util.{ FilePosition, NoPosition, SourcePosition }
+import java.io.File
+import Scope.Global
+import sbt.Def._
+
+object LintBuild {
+  lazy val lintSettings: Seq[Setting[_]] = Seq(
+    excludeLintKeys := Set(
+      aggregate,
+      concurrentRestrictions,
+      commands,
+      crossScalaVersions,
+      onLoadMessage,
+      sbt.nio.Keys.watchTriggers,
+    ),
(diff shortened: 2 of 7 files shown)


main/src/main/scala/sbt/internal/LintBuild.scala
```diff
@@ -0,0 +1,157 @@
...
+    ),
+    includeLintKeys := Set(
+      scalacOptions,
+      javacOptions,
+      javaOptions,
+      incOptions,
+      compileOptions,
+      packageOptions,
+      mappings,
+      testOptions,
```
Duhemm: I'd add `mainClass` as well.

main/src/main/scala/sbt/Keys.scala
```diff
@@ -491,6 +491,9 @@ object Keys {
   private[sbt] val postProgressReports = settingKey[Unit]("Internally used to modify logger.").withRank(DTask)
   @deprecated("No longer used", "1.3.0")
   private[sbt] val executeProgress = settingKey[State => TaskProgress]("Experimental task execution listener.").withRank(DTask)
+  val lintBuild = inputKey[Unit]("Stop a background job by providing its ID.")
```
Duhemm: Did you forget to change the description?

main/src/main/scala/sbt/internal/LintBuild.scala
```diff
@@ -0,0 +1,157 @@
...
+    ),
+    lintBuild := lintBuildTask.evaluated,
+  )
+
+  // input task version of the lintBuild
+  def lintBuildTask: Def.Initialize[InputTask[Unit]] = Def.inputTask {
+    val _ = Def.spaceDelimited().parsed // not used yet
```
Duhemm: What are the plans in the future for the input of the task?

main/src/main/scala/sbt/internal/LintBuild.scala
```diff
@@ -0,0 +1,157 @@
...
+          case pos: FilePosition => buffer.append(s"  +- ${pos.path}:${pos.startLine}")
+          case _                 => ()
+        }
+    }
+    buffer.append(" ")
+    buffer.append(
+      "note: a setting might still be used by a command; to exclude a key from this `lintBuild` check"
+    )
+    buffer.append(
+      "either append it to `Global / excludeLintKeys` or set call .withRank(KeyRanks.Invisible) on the key"
```
Duhemm: `... or call .withRank(...`

main/src/main/scala/sbt/internal/LintBuild.scala
```diff
@@ -0,0 +1,157 @@
...
+    def isSettingKey(u: UnusedKey): Boolean = u.data match {
+      case Some(data) => data.settingValue.isDefined
+      case _          => false
+    }
+    def isLocallyDefined(u: UnusedKey): Boolean = u.positions exists {
+      case pos: FilePosition => pos.path.contains(File.separator)
+      case _                 => false
+    }
+    def isInvisible(u: UnusedKey): Boolean = u.scoped.key.rank == KeyRanks.Invisible
+    val unusedSettingKeys = withDefinedAts collect {
```
Duhemm: This will contain both `SettingKey`s and (explicitly included) `TaskKey`s, no? Why not renaming this to `unusedKeys`?

dwijnand:
Wouldn't it be better to opt-in to such linting?  Either by opting into linting on load or by just running a task (or command).

jvican:
Great work :+1: I like that such linting is enabled by default and it doesn't require an opt-in. IMO build users should not opt-in to get some checks that are most likely always on point: if their task/setting implementation isn't used anywhere, either they have dead code, sbt has a bug in its build linting logic or they messed it up. Hopefully the first and third possibilities are the most likely.

dwijnand:
If there's high confidence then instead of warning (and going ignored, for instance in CI) let it fail the build load.

main/src/main/scala/sbt/internal/LintBuild.scala
eed3si9n: I was thinking options that could tune the filtering, for example to show:
- unused settings introduced by plugins
- unused tasks in `build.sbt`
- unused settings and tasks introduced by plugins

main/src/main/scala/sbt/internal/LintBuild.scala
eed3si9n: 👍  addressed in https://github.com/sbt/sbt/pull/5153/commits/460d1f5aa7e4aaa586fbf0de16ed05a97682ff7c.

main/src/main/scala/sbt/Keys.scala
eed3si9n: addressed in 460d1f5.

main/src/main/scala/sbt/internal/LintBuild.scala
eed3si9n: addressed in 460d1f5.

main/src/main/scala/sbt/internal/LintBuild.scala
eed3si9n: changed in 460d1f5.

eed3si9n:
To be honest, I am not quite sure what the signal/noise ratio would be until people start using them, so this PR keeps the option somewhat open. Initially there might be some adjustment period needed, if there's a popular plugin that uses settings from a command. (Exclusion doesn't introduce new type, so plugins should be able to exclude without adding dependencies on sbt 1.4.x).

Once the dust settles during Mx or RC, and if it's mostly good suggestions, then we can keep it on as part of `reload`. Otherwise, we can just let users run `lintUnused` task or let ppl opt-in globally.

I think it would also be interesting to have a mode for plugin authors to self-check if their plugin wiring is kosher. (Currently only keys from local files are checked)

PR #5156, opened 2019-10-05
Improve task timing output and supershell thread management

eatkins:
When running sbt -Dtask.timings=true, the task timings get printed to
the console which can overwrite the shell prompt. When we use a logger,
the timing lines are correctly separated from the prompt lines.


eatkins:
I think this is a necessary complement to https://github.com/sbt/sbt/pull/5148. I'm pretty sure that #5148 had the unintended consequence of starting a progress thread while in the shell.

eed3si9n:
Looks like this PR contains some commits from other PRs?

eatkins:
Shoot. Indeed. I was dogfooding a number of changes and forgot the branch was dirty.

eed3si9n (approved):

PR #5163, opened 2019-10-07
[1.3.x] supershell backports

eatkins:
Backports #5148 and #5156 to 1.3.x.

### DIFF
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -224,36 +224,59 @@ object EvaluateTask {
       structure: BuildStructure,
       state: State
   ): ExecuteProgress[Task] = {
-    state.get(currentTaskProgress).map(_.progress).getOrElse {
-      val maker: Seq[Keys.TaskProgress] = getSetting(
-        Keys.progressReports,
-        Seq(),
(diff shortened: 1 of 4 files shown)


eed3si9n:
Since this is a biggish change to go without RC, how confident are you on this one, esp balancing the risk against the bug it's fixing?

eatkins:
I feel pretty good about this. Supershell is not really usable for me in the swoval project without these changes (I have a number of dynamic tasks and some tasks with very long names that cause wrapping). Moreover, the thread leaks are real. A while back I noticed that sbt memory grew faster with supershell on vs off because of the leaks.

The util changes that fix text wrapping are probably the most important, but the sbt changes matter too.

To see the difference between 1.3.2 and 1.4.x with the latest util, I uploaded two asciinema casts. Here is 1.3.2:
[asciicast](https://asciinema.org/a/522KRwTvU2PIHpSqWtdghopOe)
Here is the latest develop (including the util text wrapping fixes):
[asciicast](https://asciinema.org/a/BDRxHt0UpxkxmNgEaHkRUyguU)

I would probably disable supershell entirely if I was seeing the artifacts that are present in 1.3.2 (hence why I made these changes).

eed3si9n (approved):

PR #5182, opened 2019-10-14
apply -Yno-lub

eed3si9n:
To demonstrate [-Yno-lub](http://eed3si9n.com/stricter-scala-with-ynolub), this shows the code changes that removes lubbing (Not all subprojects are done).

After I made the changes, I switched the Scala back to normal 2.12.10.


dwijnand:
What do you think about it?

eed3si9n:
There were various cases that my -Yno-lub imple didn't handle well in 2.12, so I had to go back forth between making changes here and debugging scala, so my impression is a bit muddled.

Once it started semi-working, I think the reported failures were reasonable, and edits I needed to make were mostly predictable. When things compiled I was more confident about all the inferred types knowing that it's doing less magic.

PR #5197, opened 2019-10-17
Remove excess quotation mark

jsoref:
```sbt
[debug] "not up to date. inChanged = true, force = false
[debug] Updating ...
```


eed3si9n (approved):
Thanks @jsoref!

PR #5198, opened 2019-10-17
Update .gitignore with metals bloop and eclipse

regadas:


.gitignore
```diff
@@ -7,4 +7,11 @@ npm-debug.log
...
 !sbt/src/server-test/completions/target
 .big
+
+# eclipse, intellij, bloop, metals
+.classpath
+.project
 .idea
+.settings
+.bloop
+.metals
```
eed3si9n, line 17: Tooling specific stuff should be in a global gitignore of each user. I've only conceded to `.idea` because that's what most new users seem to use.
https://help.github.com/en/articles/ignoring-files

regadas: ahhh gotcha! 😄 I'll close this.

PR #5207, opened 2019-10-21
make sbt server more extensible

### LINKED ISSUE #3112: Add `extract` to the state API
This is a common sbt idiom:
```scala
sbt.Project.extract(state.value)
```
  
I don't know the reasons why extract was added in `Project`, but my guess would be that it was done to make it binary compatible.
  
I think that this use case should be optimized, by adding `extract` directly into the state API. For newcomers unfamiliar with the sbt API, it is difficult to find `Extracted` and that you can get it via the methods in the `Project` companion object.
  
Having all this functionality in the state API makes all these options both IDE-friendly and user-friendly, since one just has to browse through the methods of the API to find what they are looking for.
  
What do you think @dwijnand @eed3si9n ? I want to PR this to sbt 1.0.

### LINKED ISSUE #4095: sbt server: querying task values
## problem

In the current sbt server implementation it's possible either to query a setting (`sbt/setting`) or to execute an arbitrary "command" (`sbt/exec`). 

The purpose of settings is to store some static values and with `sbt/setting` client receives the value of the setting. On the contrast tasks get evaluated every time you query them and most of the time (?) they are used for the side effects they produce. In such cases client can use `sbt/exec` to run a task.

But some tasks are useful exactly because of the values they store, for example: `scalacOptions`, `credentials`, `sources`, `dependencyClasspath`, etc. And currently there is no way to query those values: `sbt/setting` works only for setting keys and `sbt/exec` only logs output, but doesn't return a value.

## suggestion

I suggest to add `sbt/task` request (or probably better improve `sbt/exec`) to run tasks and retrieve their values. It's similar to `sbt/setting`, but runs the task like `sbt/exec`.

I think that changing `sbt/exec` response is better because it allows clients to use the same request and decide whether they received what they wanted or not (current response is an `ExecStatusEvent`). Introduction of a new request IMO makes it more complicated: either try and fail or check sbt version somehow to decide whether it's supported.

## notes

This has been discussed a bit on Gitter some time ago: https://gitter.im/sbt/sbt-contrib/archives/2017/10/13
And here is the "conclusion":

> **Simon Schäfer @sschaef Oct 13 2017 10:50**
> `sbt/exec` doesn't return anything to the client right now
> it just executes the task
>
> **Dale Wijnand @dwijnand Oct 13 2017 10:53**
> well that's not right at all
> I mean, that's not what should be happening.

Also a later mention of the same problem: https://gitter.im/sbt/sbt-contrib?at=59efc01119147ac323189a1e

eed3si9n:
Ref https://github.com/sbt/sbt/issues/3890
Fixes https://github.com/sbt/sbt/issues/4095
Fixes https://github.com/sbt/sbt/issues/3112

The sbt server extension was first introduced in sbt 1.2.0 (https://github.com/sbt/sbt/pull/3975) but it didn't have access to the sbt state, so it was too weak to do anything useful.

This adds a few implicit extension methods to `State`:

```scala
trait UpperStateOps extends Any {

  /**
   * ProjectRef to the current project of the state session that can be change using
   * `project` commmand.
   */
  def currentRef: ProjectRef

  /**
   * Current project of the state session that can be change using `project` commmand.
   */
  def currentProject: ResolvedProject

  private[sbt] def structure: BuildStructure

  private[sbt] def session: SessionSettings

  private[sbt] def currentUnit: LoadedBuildUnit

  /**
   * Gets the value assigned to `key` in the computed settings map.
   * If the project axis is not explicitly specified, it is resolved to be the current project according to the extracted `session`.
   * Other axes are resolved to be `Zero` if they are not specified.
   */
  def setting[A](key: SettingKey[A]): A

  /**
   * Gets the value assigned to `key` in the computed settings map.
   * If the project axis is not explicitly specified, it is resolved to be the current project according to the extracted `session`.
   * Other axes are resolved to be `Zero` if they are not specified.
   */
  def taskValue[A](key: TaskKey[A]): Task[A]

  /**
   * Runs the task specified by `key` and returns the transformed State and the resulting value of the task.
   * If the project axis is not defined for the key, it is resolved to be the current project.
   * Other axes are resolved to `Zero` if unspecified.
   *
   * This method requests execution of only the given task and does not aggregate execution.
   * See `runAggregated` for that.
   *
   * To avoid race conditions, this should NOT be called from a task.
   */
  def unsafeRunTask[A](key: TaskKey[A]): (State, A)

  /**
   * Runs the input task specified by `key`, using the `input` as the input to it, and returns the transformed State
   * and the resulting value of the input task.
   *
   * If the project axis is not defined for the key, it is resolved to be the current project.
   * Other axes are resolved to `Zero` if unspecified.
   *
   * This method requests execution of only the given task and does not aggregate execution.
   * To avoid race conditions, this should NOT be called from a task.
   */
  def unsafeRunInputTask[A](key: InputKey[A], input: String): (State, A)

  /**
   * Runs the tasks selected by aggregating `key` and returns the transformed State.
   * If the project axis is not defined for the key, it is resolved to be the current project.
   * The project axis is what determines where aggregation starts, so ensure this is set to what you want.
   * Other axes are resolved to `Zero` if unspecified.
   *
   * To avoid race conditions, this should NOT be called from a task.
   */
  def unsafeRunAggregated[A](key: TaskKey[A]): State

  /** Appends the given settings to all the build state settings, including session settings. */
  def appendWithSession(settings: Seq[Setting[_]]): State

  /**
   * Appends the given settings to the original build state settings, discarding any settings
   * appended to the session in the process.
   */
  def appendWithoutSession(settings: Seq[Setting[_]], state: State): State

  def respondEvent[A: JsonFormat](event: A): Unit
  def respondError(code: Long, message: String): Unit
  def notifyEvent[A: JsonFormat](method: String, params: A): Unit
}
```

so commands (and tasks) can respond to LSP (JSON-RPC) request using the context captured in the state. The extension would look like:

```scala
    commands += Command.command("fooExport") { s0: State =>
      val (s1, cp) = s0.unsafeRunTask(Compile / fullClasspath)
      s0.respondEvent(cp.map(_.data))
      s1
    },
```

It is possible to use tasks, but because you would need to respond once for all subprojects, likely it's easier to deal with request-response using a command.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -675,6 +675,7 @@ lazy val protocolProj = (project in file("protocol"))
       exclude[DirectMissingMethodProblem]("sbt.protocol.SettingQuerySuccess.copy$default$*"),
       // ignore missing methods in sbt.internal
       exclude[DirectMissingMethodProblem]("sbt.internal.*"),
+      exclude[MissingTypesProblem]("sbt.internal.protocol.JsonRpcResponseError"),
     )
   )
 
--- a/main/src/main/scala/sbt/BuildSyntax.scala
+++ b/main/src/main/scala/sbt/BuildSyntax.scala
@@ -22,5 +22,8 @@ private[sbt] trait BuildSyntax {
   def dependsOn(deps: ClasspathDep[ProjectReference]*): DslEntry = DslEntry.DslDependsOn(deps)
   // avoid conflict with `sbt.Keys.aggregate`
   def aggregateProjects(refs: ProjectReference*): DslEntry = DslEntry.DslAggregate(refs)
+
+  implicit def sbtStateToUpperStateOps(s: State): UpperStateOps =
+    new UpperStateOps.UpperStateOpsImpl(s)
 }
 private[sbt] object BuildSyntax extends BuildSyntax
--- a/main/src/main/scala/sbt/MainLoop.scala
+++ b/main/src/main/scala/sbt/MainLoop.scala
@@ -13,6 +13,7 @@ import java.util.Properties
 import jline.TerminalFactory
 import sbt.internal.{ Aggregation, ShutdownHooks }
 import sbt.internal.langserver.ErrorCodes
+import sbt.internal.protocol.JsonRpcResponseError
 import sbt.internal.util.complete.Parser
 import sbt.internal.util.{ ErrorHandling, GlobalLogBacking }
 import sbt.io.{ IO, Using }
@@ -230,6 +231,9 @@ object MainLoop {
           }
       }
     } catch {
+      case err: JsonRpcResponseError =>
+        StandardMain.exchange.respondError(err, exec.execId, channelName.map(CommandSource(_)))
+        throw err
       case err: Throwable =>
         val errorEvent = ExecStatusEvent(
           "Error",
@@ -237,9 +241,10 @@ object MainLoop {
           exec.execId,
           Vector(),
           ExitCode(ErrorCodes.UnknownError),
+          Option(err.getMessage),
         )
         import sbt.protocol.codec.JsonProtocol._
-        StandardMain.exchange publishEvent errorEvent
+        StandardMain.exchange.publishEvent(errorEvent)
         throw err
     }
   }
--- a/main/src/main/scala/sbt/UpperStateOps.scala
+++ b/main/src/main/scala/sbt/UpperStateOps.scala
@@ -1,0 +1,142 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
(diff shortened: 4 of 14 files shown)


main/src/main/scala/sbt/StateServerOps.scala
```diff
@@ -0,0 +1,34 @@
...
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt
+
+import sjsonnew.JsonFormat
+
+trait StateServerOps extends Any {
+  def extract: Extracted
```
eed3si9n: Previous review was that `extract` should expand out to actual methods.

PR #5225, opened 2019-11-05
Fixes high CPU usage while retrying background job shutdown

samvel1024:
I have noticed that sometimes an sbt shell process uses 100% CPU for several hours non-stop. After some debugging seems like it is constantly calling `Thread.interrupt`. Here is the loop that keeps the CPU busy. 

https://github.com/sbt/sbt/blob/44cc19e66dd63665ace8320b282387c71f6a9c61/main/src/main/scala/sbt/internal/DefaultBackgroundJobService.scala#L160-L166

As a quickfix I have added a sleep to be triggered only in case when we are trying to interrupt the thread second (or more) time.


lightbend-cla-validator:
Hi @samvel1024,

eed3si9n (approved):
Thanks @samvel1024! LGTM

eed3si9n:
@samvel1024 Let me know when if you signed the CLA.

samvel1024:
Seems like I have already done it.

PR #5254, opened 2019-11-28
Cleanup project/Transform.scala

dwijnand:
I was looking for some usage of "mappings", and noticed that half of project/Transform is unused.
Then one thing lead to another, as usual...


eed3si9n (approved):
nice.

PR #5259, opened 2019-11-30
Background jobs

### LINKED ISSUE #2274: Allow for graceful termination of the program started by sbt run
Enabling  `cancelable in Global := true` allows to press Ctrl+C and terminate the running program without terminating SBT, however the way how it is handled is quite surprising and doesn't really do what is typically expected from Ctrl+C.

Ctrl+C sends the TERM signal to the application, not KILL. Unfortunately SBT behaves as if it was requested a KILL and interrupts the application. In forked mode, it also seems to kill the JVM very quickly after the application receives the TERM signal. This gives no way for the application to terminate gracefully as it would do when running without SBT supervision.

Hitting Ctrl+C inside of sbt run should pass the TERM signal to the running application and just let the application decide how it handles it. Most of applications know what to do. Maybe, if some users still insist on hard terminating of the process (if it was frozen or doesn't respond to normal TERM), the second Ctrl+C would invoke the current behaviour and interrupt / kill the application.

eatkins:
This PR makes a number of miscellaneous improvements to the background job service including:
* Shuts down background jobs when an error occurs (which happens when Ctrl+C is inputted and thus should fix https://github.com/sbt/sbt/issues/2274 for non-forked `run` provided the main method handles interrupts)
* Moves the background job service directory from a temp directory specified by java.io.tmpdir (which is /tmp on linux and user specific on mac and windows) to `$PROJECT_BASE_DIR/target/bg-jobs`).

### DIFF
--- a/main/src/main/scala/sbt/BackgroundJobService.scala
+++ b/main/src/main/scala/sbt/BackgroundJobService.scala
@@ -8,11 +8,14 @@
 package sbt
 
 import java.io.Closeable
+
 import sbt.util.Logger
-import Def.{ ScopedKey, Classpath }
+import Def.{ Classpath, ScopedKey }
 import sbt.internal.util.complete._
 import java.io.File
-import scala.util.Try
+
+import scala.util.control.NonFatal
+import scala.util.{ Failure, Success, Try }
 
 abstract class BackgroundJobService extends Closeable {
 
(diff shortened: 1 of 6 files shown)


eed3si9n (approved):
These changes LGTM

PR #5260, opened 2019-11-30
Improve StateTransform

eatkins:
### Updating State in a task

It is possible to update the sbt state in a task. To do this, the
task must return type `StateTransform`. The state will be transformed upon
completion of task evaluation. The `StateTransform` is constructed with
a function from `State => State` that accepts the previous value of the `State`
and generates a new state. For example:

```scala
import complete.DefaultParsers._
val counter = AttributeKey[Int]("counter")
val setCounter = inputKey[StateTransform]("Set the value of the counter attribute")
setCounter := {
  val count = (Space ~> IntBasic).parsed
  StateTransform(_.put(counter, count))
}
```

creates the input task `setCounter` that sets the counter attribute to some
value.

### Original description

The StateTransform class introduced in
9cdeb7120ed7bdb9291e24060a16a268eeb9cfcf did not cleanly integrate with
logic for transforming the state using the `transformState` task
attribute. For one thing, the state transform was only applied if the
root node returned StateTransform but no transformation would occur if a
task had a dependency that returned StateTransform. Moreover, the
transformation would override all other transformations that may have
occurred during task evaluation.

This commit updates StateTransform to act very similarly to the
transformState attribute. Instead of wrapping a `State` instance, it now
wraps a transformation function from `State => State`. This function
can be applied on top of or prior to the other transformations via the
`transformState`.

For binary compatibility with 1.3.0, I had to add the stateProxy
function as a constructor parameter in order to implement the `state`
method. The proxy function will generally throw an exception unless the
legacy api is used. To avoid that, I private[sbt]'d the legacy api so
that binary compatibility is preserved but any builds targeting > 1.4.x
will be required to use the new api.

Unfortunately I couldn't private[sbt] StateTransform.apply(state: State)
because mima interpreted it as a method type change becuase I added
StateTransform.apply(transform: State => State). This may be a mima bug
given that StateTransform.apply(state: State) would be jvm public even
when private[sbt], but I figured it was quite unlikely that any users
were using this method anyway since it was incorrectly implemented in
1.3.0 to return `state` instead of `new StateTransform(state)`.

### DIFF
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -515,17 +515,14 @@ object EvaluateTask {
       state: State,
       root: Task[T]
   ): (State, Result[T]) = {
-    val newState = results(root) match {
-      case Value(KeyValue(_, st: StateTransform) :: Nil) => st.state
-      case _                                             => stateTransform(results)(state)
-    }
-    (newState, results(root))
+    (stateTransform(results)(state), results(root))
   }
   def stateTransform(results: RMap[Task, Result]): State => State =
     Function.chain(
       results.toTypedSeq flatMap {
-        case results.TPair(Task(info, _), Value(v)) => info.post(v) get transformState
-        case _                                      => Nil
+        case results.TPair(_, Value(KeyValue(_, st: StateTransform))) => Some(st.transform)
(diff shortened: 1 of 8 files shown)


eed3si9n (approved):
overall LGTM.
Should we add docs for this on sbt/website?

eatkins:
https://github.com/sbt/website/pull/856

Issue #5262, opened 2019-12-02
"Can't find bundle for base name" during test using ScalaTest


rjmac:
## steps
sbt version: 1.3.4

## problem

ScalaTest intermittently fails with a "can't find resource bundle" exception after upgrading to SBT 1.3.4.  This is the same bug as in #5169 but I've got more information.

## expectation

ScalaTest doesn't fail to read its resource bundle.

## notes

This has started happening to our builds too, since upgrading to 1.3.  I've done some digging and this is what I've found.  I haven't managed to make a reliable repro, but I think what I have here shows it's racing with another thread in a fairly small window (the amount of time it takes to read scalatest's resource bundle properties) which is why it's difficult to reproduce.  As in the other bug, our project uses `fork in Test := true`

Here's the relevant part of the stack trace, reversed for annotation purposes, starting from where ScalaTest enters the standard library:

    [error] 	at org.scalatest.Resources$.resourceBundle$lzycompute(Resources.scala:8)
    [error] 	at java.util.ResourceBundle.getBundle(ResourceBundle.java:782)
    [error] 	at java.util.ResourceBundle.getBundleImpl(ResourceBundle.java:1370)
    [error] 	at java.util.ResourceBundle.findBundle(ResourceBundle.java:1436)
    [error] 	at java.util.ResourceBundle.findBundle(ResourceBundle.java:1436)
    [error] 	at java.util.ResourceBundle.findBundle(ResourceBundle.java:1482)
    [error] 	at java.util.ResourceBundle.loadBundle(ResourceBundle.java:1518)
    [error] 	at java.util.ResourceBundle$Control.newBundle(ResourceBundle.java:2704)

`newBundle` is where the stream is opened, a few lines before 2704.  It is acquired from a classloader in the usual `getResource`y ways.

    [error] 	at java.util.PropertyResourceBundle.<init>(PropertyResourceBundle.java:138)

This does nothing with the stream except pass it on to the constructor of `Properties`

    [error] 	at java.util.Properties.load(Properties.java:341)

This wraps the stream in a `LineReader` (a private class within `Properties`) and passes it on.  The constructor of `LineReader` does nothing with the stream except store it in a field.

    [error] 	at java.util.Properties.load0(Properties.java:353)

This actually does things with it: specifically, it repeatedly calls `LineReader.readLine`...

    [error] 	at java.util.Properties$LineReader.readLine(Properties.java:435)

[...]

PR #5273, opened 2019-12-06
Place scalatest framework jar in its own classloader

eatkins:
Closing the ManagedClassLoader generated by test can cause nonlocal
effects because the jdk shares some JarFile resources across multiple
URLClassLoaders. As a result, if one classloader is trying to load a
resource and the classloader is closed, it might cause the resource
loading to fail (see https://github.com/sbt/sbt/issues/5262). This can
be fixed by moving the scalatest framework jar (and its dependencies)
into an additional classloader layer that sits between the scala library
loader and the rest of the test dependencies.

In addition to adding the new layer, I reworked the
ReverseLookupClassLoader to use its dependent classloader to find
resources that may below it in the classloading hierarchy rather than
constructing an entirely new classloader for resources.

After this change, I was able to run test in the repro project:
https://github.com/rjmac/sbt-5262 1000 times with no failures. Note that
the repro is sensitive to the jdk used. I could not reproduce with jdk11
but I could typically induce a failure within 20 or so runs with jdk8.

I benchmarked this change with
https://github.com/eatkins/scala-build-watch-performance and performance
was roughly the same as 1.3.4 with turbo mode and about 200-250ms faster
in non-turbo mode (which can be explained by the time to load the
scalatest classes).

### DIFF
--- a/main/src/main/scala/sbt/internal/ClassLoaders.scala
+++ b/main/src/main/scala/sbt/internal/ClassLoaders.scala
@@ -156,10 +156,25 @@ private[sbt] object ClassLoaders {
         val cpFiles = fullCP.map(_._1)
 
         val allDependencies = cpFiles.filter(allDependenciesSet)
-        val scalaReflectJar = allDependencies.collectFirst {
-          case f if f.getName == "scala-reflect.jar" =>
-            si.allJars.find(_.getName == "scala-reflect.jar")
-        }.flatten
+        /*
+         * We need to put scalatest and its dependencies (scala-xml) in its own layer so that the
+         * classloader for scalatest is generally not ever closed. This is because one project
+         * closing its scalatest loader can cause a crash in another project if that project
+         * is reading the scalatest bundle properties because the underlying JarFile instance is
+         * shared across URLClassLoaders if it points to the same jar.
+         *
+         * TODO: Make this configurable and/or add other frameworks in this layer.
+         */
+        val (frameworkDependencies, otherDependencies) = allDependencies.partition { d =>
+          val name = d.getName
+          name.startsWith("scalatest_") || name.startsWith("scalactic_") ||
+          name.startsWith("scala-xml_")
+        }
+        def isReflectJar(f: File) =
+          f.getName == "scala-reflect.jar" || f.getName.startsWith("scala-reflect-")
+        val scalaReflectJar = otherDependencies.collectFirst {
+          case f if isReflectJar(f) => si.allJars.find(isReflectJar).getOrElse(f)
+        }
         val scalaReflectLayer = scalaReflectJar
           .map { file =>
             cache.apply(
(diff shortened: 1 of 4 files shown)


eatkins:
Should fix #5262.

eed3si9n:
Thanks for looking into this.

main/src/main/scala/sbt/internal/ClassLoaders.scala
```diff
@@ -156,10 +156,25 @@ private[sbt] object ClassLoaders {
...
+         * is reading the scalatest bundle properties because the underlying JarFile instance is
+         * shared across URLClassLoaders if it points to the same jar.
+         *
+         * TODO: Make this configurable and/or add other frameworks in this layer.
+         */
+        val (frameworkDependencies, otherDependencies) = allDependencies.partition { d =>
+          val name = d.getName
+          name.startsWith("scalatest_") || name.startsWith("scalactic_") ||
+          name.startsWith("scala-xml_")
+        }
```
eed3si9n, line 172: What makes ScalaTest so special? Could this potentially happen to anything that reads from a resource input stream?

eatkins: Theoretically this issue could come up with any jar that is shared across multiple projects. I also don't know if it affects classloading or only resource loading. I will also make a separate PR that re-enables `-Dsbt.classloader.close`. At the moment it doesn't do anything because I mistakenly thought that zombie classloaders would solve the issue. If closing a classloader in one project can cause an independent classloader to fail to load classes in another project, users should be able to disable closing the loaders. Given that this is the only reported manifestation of the problem, I'm not convinced that its widespread enough to disable closing the classloaders by default. It's possible though that is the route we should ultimately take. As I mentioned in an issue comment, the other approach would be to copy the classpath ala `run` in `test`. There is merit to this idea but I worry about the io and disk space overhead.

Scalatest is special because of its ubiquity in the scala ecosystem. I could see an argument for doing this for all of the default testFrameworks: ScalaCheck, Specs, Specs2 and JUnit. It's just kind of a pain because you have to include all of the framework dependencies in this layer. Because the testFrameworks are specified by class name, we don't know a priori what library dependency provides the implementation which makes auto identifying the necessary jars difficult.

Other notes:
I could only reproduce the issue with jdk8 so it isn't clear to me how likely closing the loaders is to cause problems in newer jdks. If it isn't an issue in newer jdks, we could consider only closing the classloaders given some minimum jdk threshold.

Independent of the resource loading issue, this also brings a performance improvement to builds that use scalatest without turbo. In my experience, scalatest is by far the heaviest test framework so the performance benefit would be more modest with say utest or minitest.

eatkins: After writing this, it occurred to me that another option would be to enqueue classloaders for closing after test completes rather than directly close them. Then we could actually perform the closing when we enter the shell.

eed3si9n (approved):

PR #5277, opened 2019-12-09
WIP ci color output by default

### LINKED ISSUE #5269: [1.3.3] BUILD_NUMBER env var does not enable color in Jenkins
## steps
sbt version: 1.3.3
scala version: 2.13.1
Jenkins version: 2.150
When running a suite of scala tests using SBT and Jenkins

## problem
There is no ANSI color output in my Jenkins logs

## expectation
As noted in https://www.scala-sbt.org/1.x/docs/Command-Line-Reference.html the sbt.ci flag defaults to false UNLESS the BUILD_NUMBER env variable is set. As Jenkins sets the BUILD_NUMBER env variable I would expect that the ci flag is automatically set to true and the Jenkins logs are properly colored.

francisdb:
fixes #5269

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/ConsoleAppender.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/ConsoleAppender.scala
@@ -8,17 +8,17 @@
 package sbt.internal.util
 
 import sbt.util._
-import java.io.{PrintStream, PrintWriter}
+import java.io.{ PrintStream, PrintWriter }
 import java.lang.StringBuilder
 import java.util.Locale
-import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger, AtomicReference}
-
-import org.apache.logging.log4j.{Level => XLevel}
-import org.apache.logging.log4j.message.{Message, ObjectMessage, ReusableObjectMessage}
-import org.apache.logging.log4j.core.{LogEvent => XLogEvent}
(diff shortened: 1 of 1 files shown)


francisdb:
@eed3si9n I would like to access `insideCi` from `ConsoleAppender` so I guess the implementation should be shared by moving it elsewhere. Where would I put this?

```scala
val insideCi = sys.env.contains("BUILD_NUMBER") ||
        sys.env.contains("CI") || SysProp.ci
```

Already calling `SysProp` is not possible in `ConsoleAppender` so going DRY here is complicating things

francisdb:
Also if somebody changes the `insideCI` setting he would expect this to be propagated and not ignored?

eed3si9n:
util-logging used to be part of a different repository, and `SysProp` lives in a layer above util-logging.

> Also if somebody changes the `insideCI` setting he would expect this to be propagated and not ignored?

I am kind of ok with treating `insideCI` setting a read-only setting, but if you would want to respect the setting and also avoid duplicating the implementation of `insideCI`, you could potentially use mutation to solve it. See for example https://github.com/sbt/sbt/commit/8215026bc3c45ff70d4d9c0be7c98bf5d2b85dc2.

There's an AtomicInteger on ConsoleAppender called `widthHolder`. At some point when supershell is being initialized this sets the terminal width so logger knows it. Similarly you could add a new setting that uses `insideCI.value` which sets an `AtomicBoolean` for color override.

francisdb:
dropping this for now

PR #5278, opened 2019-12-09
No supershell for Emacs and other color supporting dumb terminals

hvesalai:
As the title says

### DIFF
--- a/main/src/main/scala/sbt/internal/SysProp.scala
+++ b/main/src/main/scala/sbt/internal/SysProp.scala
@@ -89,7 +89,8 @@ object SysProp {
 
   def fileCacheSize: Long =
     SizeParser(System.getProperty("sbt.file.cache.size", "128M")).getOrElse(128L * 1024 * 1024)
-  def supershell: Boolean = booleanOpt("sbt.supershell").getOrElse(color)
+  def dumbTerm: Boolean = sys.env.get("TERM").filter(_ == "dumb").isDefined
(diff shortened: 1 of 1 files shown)


main/src/main/scala/sbt/internal/SysProp.scala
```diff
@@ -89,7 +89,8 @@ object SysProp {
 
   def fileCacheSize: Long =
     SizeParser(System.getProperty("sbt.file.cache.size", "128M")).getOrElse(128L * 1024 * 1024)
-  def supershell: Boolean = booleanOpt("sbt.supershell").getOrElse(color)
+  def insideEmacs: Boolean = sys.env.contains("INSIDE_EMACS")
```
eed3si9n: I am not against checking specifically for Emacs here, but I wonder if it makes sense here to check for `TERM` set to `"dumb"`, which supposedly is the value set on Emacs shell according to @SethTisue.

hvesalai: Yes, that's the other option. I can change the PR to do that if we so choose

```
scala> sys.env.get("TERM")
res0: Option[String] = Some(dumb)
```

eed3si9n: I think checking for `TERM` is preferable since the notion could apply to other environments.

eed3si9n (approved):
Thanks @hvesalai!

PR #5289, opened 2019-12-10
Do not use temporary directories in java.io.tmpdir

eatkins:
sbt should not by default create files in the location specified by
java.io.tmpdir (which is the default behavior of apis like
IO.createTemporaryDirectory or Files.createTempFile) because they have a
tendency to leak and it also isn't even guaranteed that the user has
write permissions there (though this is unlikely). Doing so creates the
possibility for leaks

I git grepped for `createTemp` and found these apis. After this change,
the files created by sbt should largely be localized to the project and
sbt global base directories.

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -141,12 +141,17 @@ object StandardMain {
   val console: ConsoleOut =
     ConsoleOut.systemOutOverwrite(ConsoleOut.overwriteContaining("Resolving "))
 
-  def initialGlobalLogging: GlobalLogging =
+  private[this] def initialGlobalLogging(file: Option[File]): GlobalLogging = {
+    file.foreach(f => if (!f.exists()) IO.createDirectory(f))
     GlobalLogging.initial(
       MainAppender.globalDefault(console),
-      File.createTempFile("sbt", ".log"),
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/Main.scala
```diff
@@ -141,12 +141,16 @@ object StandardMain {
...
+    Option(file).foreach(f => if (!f.exists()) IO.createDirectory(f))
     GlobalLogging.initial(
       MainAppender.globalDefault(console),
-      File.createTempFile("sbt", ".log"),
+      File.createTempFile("sbt-global-log", ".log", file),
       console
     )
+  }
+  @deprecated("use version that takes file argument", "1.4.0")
+  def initialGlobalLogging: GlobalLogging = initialGlobalLogging(null)
```
eed3si9n: No `null` please.

eatkins: In my opinion this is a good use of null because otherwise there has to be a variant of initialGlobalLogging that takes Option[File] as input.

eed3si9n: I'd much rather deal with `Some(...)` and `None.orNull` than `null`.

eed3si9n (approved):
LGTM

PR #5290, opened 2019-12-10
Use logger in scripted instead of println

eatkins:


eed3si9n (approved):

scripted-sbt-redux/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
```diff
@@ -64,7 +64,7 @@ final class ScriptedTests(
       val n = nme.getName
       val label = s"$g / $n"
       () => {
-        println(s"Running $label")
+        log.info(s"Running $label")
```
eed3si9n, line 67: I wonder if the use of println was intentional because scripted logs are often buffered, and they don't display unless the test fails.

eatkins: I am almost certain that these lines were not buffered before this change. Not having them would make CI terrible because you would have no idea how far along the tests were or if one of the tests was hanging.

PR #5311, opened 2019-12-14
Remove Util helper function usages

eatkins:
These methods make the code less readable in my opinion. This should be
a no-op. In addition to replacing the depreacted apis:
UtilOps.{ AnyOps, nil, nilSeq, none }, I optimized the imports of each
file where I made changes. I also rewrote a few lines that did something
like `foo get bar getOrElse baz` to `foo.get(bar).getOrElse(baz)`.

### DIFF
--- a/internal/util-collection/src/main/scala/sbt/internal/util/Settings.scala
+++ b/internal/util-collection/src/main/scala/sbt/internal/util/Settings.scala
@@ -7,11 +7,10 @@
 
 package sbt.internal.util
 
-import scala.language.existentials
-
-import Types._
+import sbt.internal.util.Types._
 import sbt.util.Show
-import Util.{ nil, nilSeq }
+
+import scala.language.existentials
 
 sealed trait Settings[ScopeType] {
   def data: Map[ScopeType, AttributeMap]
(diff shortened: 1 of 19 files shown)


eatkins:
I see now that `nil`, `nilSeq` and `none` came from 54b3405f428f111b0a197104c6805cd0e5308a84. @eed3si9n could you explain how you applied "-Yno-lub"?

I understand why lub-ing can be problematic but do the lubs removed by 54b3405f428f111b0a197104c6805cd0e5308a84 really matter? To me, these helpers make the sbt source less coherent because one encounters these seemingly random usages of `nil`, `nilSeq` and `none` and `some` without any context. Moreover, without having `-Yno-lub` a permanent part of the build, we are likely to inadvertently keep adding them. I'm pretty opposed to these methods without  at a minimum having `-Yno-lub` easily used by other developers.

eed3si9n:
I have a fork of Scala compiler on Bintray (https://bintray.com/eed3si9n/maven/scala/2.12.10-bin-9f9fb71) that I used I think. We might be able to use cross building to verify if we wanted to?

eatkins:
This isn't a big enough deal to me to leave open.

PR #5315, opened 2019-12-17
Add library-friendly entrypoint for sbt.ForkMain

olafurpg:
Previously, it was not possible to use `sbt.ForkMain` as a library since
* it called `System.exit`, shutting down the library client JVM process.
* it assumed that it was always running in the same classloader as where
  the test suites it was running causing "class not found: my.test.Suite"
  if the `ForkMain` class lives in a classloader above the test suite
  classes.

This commit adds a new `main(Array[String],ClassLoader])` entrypoint
that allows clients to call `ForkMain` as a library.

### DIFF
--- a/testing/agent/src/main/java/sbt/ForkMain.java
+++ b/testing/agent/src/main/java/sbt/ForkMain.java
@@ -117,20 +117,25 @@ final public class ForkMain {
 	// ----------------------------------------------------------------------------------------------------------------
 
 	public static void main(final String[] args) throws Exception {
+		ClassLoader classLoader = new Run().getClass().getClassLoader();
+		try {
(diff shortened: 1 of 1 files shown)


olafurpg:
I don't know if you're open to adding a library endpoint for `sbt.ForkMain` but I thought I would at least give it a try. The alternative for us to to maintain a fork of the test-agent project, which I'm fine with doing as well.

eed3si9n:
The travis CI failure on actions/doc-scala-instance seems unrelated since I am seeing that on other PRs too.

eed3si9n (approved):
LGTM

olafurpg:
Thank you for merging! I appreciate it 😁

Issue #5314, opened 2019-12-17
subproject named "client" puts sbt into client mode after +clean


eed3si9n:
Ref https://github.com/sbt/sbt/pull/5265

## steps

Here's `build.sbt` reproduced by @eed3si9n:

```scala
ThisBuild / scalaVersion     := "2.13.1"
ThisBuild / version          := "0.1.0-SNAPSHOT"
ThisBuild / organization     := "com.example"
ThisBuild / organizationName := "example"

lazy val root = (project in file("."))
  .aggregate(client)
  .settings(
    name := "sbt-cross-testing"
  )

lazy val client = project
  .settings(
    scalaVersion := "2.13.1",
    crossScalaVersions := Seq("2.13.1"),
  )
```

Then:

```
> +clean
```

## problem

sbt enters into thin client mode:

```
sbt:sbt-cross-testing> +clean
[info] Forcing Scala version to 2.12.10 on all projects.
[info] Reapplying settings...
[info] Set current project to sbt-cross-testing (in build file:/private/tmp/sbt-cross-testing/)
[success] Total time: 0 s, completed Dec 17, 2019 1:13:45 AM
[info] Forcing Scala version to 2.13.1 on all projects.
[info] Reapplying settings...
[info] Set current project to sbt-cross-testing (in build file:/private/tmp/sbt-cross-testing/)
[info] entering *experimental* thin client - BEEP WHIRR
[info] new client connected: network-1
> /
```

## expectation

Run `clean` command on subprojects.

## original report

_Originally posted by @marcospereira in https://github.com/sbt/sbt/issues/5310#issuecomment-566241237_

We are having a problem when cross running tasks in Lagom build as well, see the following job log: [job-267154846-log.txt](https://github.com/sbt/sbt/files/3970096/job-267154846-log.txt)

```# Executing command line:
java
-Xms2G
-Xmx2G
-Xss2M
-XX:MaxInlineLevel=18
-XX:MaxMetaspaceSize=1G
-Dhttps.protocols=TLSv1,TLSv1.1,TLSv1.2
-Dakka.test.timefactor=10
-jar
/home/travis/.sbt/launchers/1.3.5/sbt-launch.jar
--warn
"set concurrentRestrictions in Global += Tags.limitAll(1)"
+headerCheck
+test:headerCheck
multi-jvm:headerCheck

[...]

PR #5265, opened 2019-12-03
Fix cross command (+) for input tasks in subprojects with provided input

eatkins:
Fixes https://github.com/sbt/sbt/issues/4715

Currently if you have a root project that aggregates a subproject, foo, and run
```
+ foo / testOnly FooSpec
```
sbt will throw an exception during parsing. This PR fixes that.

PR #5317, opened 2019-12-17
workaround client / clean problem

eed3si9n:
Ref https://github.com/sbt/sbt/issues/5314
Ref https://github.com/sbt/sbt/pull/5265

In sbt 1.3.4, it's possible to define a subproject named `client`.
The current parser behaves differently whether we calll `client/clean` or `client / clean` with whitespaces. The one with the whitespace invokes `client` command (as in thin client). This gets triggered by `+clean` because the new implementation uses whitespace.


eatkins (approved):
Wow.

PR #5322, opened 2019-12-19
Clarify message on sbt.version mismatch

### LINKED ISSUE #4291: Confusing warning message for sbt version mismatch

## steps
Use a different version of sbt using `-sbt-version` argument than what is present in `project/build.properties`

```
sbt.version=1.1.6
$ sbt -sbt-version 1.2.0 update
[warn] sbt version mismatch, current: 1.2.0, in build.properties: "1.1.6", use 'reboot' to use the new value.
...
```

## problem
Warning message is confusing regarding which version of sbt will be used in project

## expectation
Warning message should tell what version of sbt will be used in the project and offer alternative to switch back to version from build.properties. `new value` is ambiguous wording in the message

## notes

sbt version: 1.2.0

octonato:
Fixes #4291

I removed the metion to `reboot` as it's not working. 
I also don't think this is something that needs to be fixed. If someone wants to use the version in `build.properties`, then better to restart sbt again without passing `-Dsbt.version`.

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -827,8 +827,8 @@ object BuiltinCommands {
     sbtVersionOpt.foreach(
       version =>
         if (version != app.id.version()) {
-          state.log.warn(s"""sbt version mismatch, current: ${app.id
-            .version()}, in build.properties: "$version", use 'reboot' to use the new value.""")
+          state.log.warn(s"""sbt version mismatch, using: ${app.id
+            .version()}, in build.properties: "$version".""")
         }
     )
   }
--- a/sbt/src/sbt-test/project/sbt-version-change/build.sbt
+++ b/sbt/src/sbt-test/project/sbt-version-change/build.sbt
@@ -3,6 +3,6 @@ TaskKey[Unit]("checkSbtVersionWarning") := {
 	val logging = state.globalLogging
 	val currVersion = state.configuration.provider.id.version()
 	val contents = IO.read(logging.backing.file)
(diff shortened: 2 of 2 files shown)


eed3si9n (approved):

dwijnand:
The whole point of `reboot` is it allows someone to change the version in build.properties and reboot _without_ restarting the JVM.

The reason the message exists is because it's commonly encountered when changing sbt.version in build.properties (including unintentionally, such as when switching branches) and running `reload`: in that case you'll be using the **old** version of sbt with the new build definition.  Without mentioning `reboot` an inexperienced user will resort to restarting the JVM all over again (and reloading the whole build, etc).

The issue #4291 has already been fixed in #4323.  The reason Renato isn't seeing the result of that fix (as well as the initially reported reproducer still reproduces) is because that fix is in 1.3.0 and when one switches to a previous version (e.g. `-Dsbt.version=1.2.8`) it runs the previous version which doesn't have the fix.

I think this should be reverted, seeing as with the #4323 fix the messaging should never be misleading (FLW).

One improvement that code could use is not being misleading when it states `in build.properties: "$version"`, given `version` is defined first in terms of the system property.  So it could either say `sbt.version system property: "$version"` instead, when the system property is set, or it could declare all 3 values (app.id.version, build.properties version, system property version) if there are 3 values.

eed3si9n:
> The issue #4291 has already been fixed in #4323. The reason Renato isn't seeing the result of that fix (as well as the initially reported reproducer still reproduces) is because that fix is in 1.3.0 and when one switches to a previous version (e.g. -Dsbt.version=1.2.8) it runs the previous version which doesn't have the fix.

ok. that makes sense.

Issue #4451, opened 2018-11-13
Provide a way to specify test report name


ignasi35:
Test reports are generated using two different filenames for specific report processors to read. 

Some processors may pick both file formats reporting twice as many tests. It'd be good to make the reporting tunable so users may choose one output format or the other

_Originally posted by @yanns in https://github.com/sbt/sbt/pull/4343#issuecomment-428160480_

eed3si9n:
For context #4343 says

> Fixes #3150 by duplicating the reports generated so given a test `com.example.MyTest.java` the following will be created:
>
> - `com.example.MyTest.xml`
> - `TEST-com.example.MyTest.xml`

eed3si9n:
To potential contributor, the relevant files are:
- testing/src/main/scala/sbt/JUnitXmlTestsListener.scala changed in https://github.com/sbt/sbt/pull/4343/files
- and https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/internal/SysProp.scala if we use system properties as flag

For example we can provide `-Dsbt.testing.legacyreport=true` to produce legacy report, but otherwise just use the new `TEST-*` file name.

eed3si9n:
Fixed in https://github.com/sbt/sbt/pull/5350 by @mrArkwright

francisdb:
this seems to have caused duplicate test reporting when using the play plugin
https://github.com/playframework/playframework/issues/10000

PR #5330, opened 2019-12-24
Ability to disable legacy JUnit report filenames

nafg:
Addresses suggestion in #4451

### DIFF
--- a/main/src/main/scala/sbt/internal/SysProp.scala
+++ b/main/src/main/scala/sbt/internal/SysProp.scala
@@ -131,4 +131,6 @@ object SysProp {
           case None       => true
         }
     }
+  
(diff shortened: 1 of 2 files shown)


eed3si9n:
@nafg Thanks for the contribution

main/src/main/scala/sbt/internal/SysProp.scala
```diff
@@ -131,4 +131,6 @@ object SysProp {
           case None       => true
         }
     }
+  
+  def legacyTestingReport: Boolean = getOrTrue("sbt.testing.legacyreport")
```
eed3si9n, line 135: SysProp probably won't work since testingProj is on the upstream of main.

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/5350

PR #5332, opened 2019-12-24
Add SlashSyntax.unparse

eatkins:
There was previously no way to convert a ScopedKey into a string that
could be parsed by the Act.aggregatedKeyParser. Because slash syntax
requires a capitalized name (more generally there is no requirement that
the slash syntax identifier matches the config name), it was necessary
to add a method toConfigIdent to KeyIndex so that we can invert the
config name to find the identifier.


eed3si9n:
Slash String conversion are mostly in 
- https://github.com/sbt/sbt/blob/develop/main-settings/src/main/scala/sbt/Def.scala and
- https://github.com/sbt/sbt/blob/develop/main-settings/src/main/scala/sbt/Scope.scala

If the existing `Scope#display` methods are not cutting it we should add more variants there.

eed3si9n:
Also what's telling of the current `Scope#display`'s limitation:

https://github.com/sbt/sbt/blob/cba384396554909813fb80b4a70c0f9459becb37/main-settings/src/main/scala/sbt/Scope.scala#L140

It seems like it's just calling `guessConfigIdent`. Maybe we need `displayStately(config: ConfigKey): String`.

eatkins:
I find it problematic that the display methods are not linked in any way to the other parsers. In my opinion, sbt should never generate display strings that don't parse in the shell. At the moment, I'm pretty sure none of the display methods generate usable strings. Besides getting the config key wrong, they put `./`for empty fields which also don't parse.

eed3si9n:
Maybe we need more basic testing in https://github.com/sbt/sbt/blob/develop/main/src/test/scala/ParseKey.scala.

eed3si9n:
I'd happy to take on the slash from here.

dwijnand:
> Besides getting the config key wrong, they put `./`for empty fields which also don't parse.

Oh, wow.  Never noticed that - it's been there since the beginning: https://github.com/sbt/sbt/commit/087bc569e05f28babbd443972b9f88b119136975#diff-1bdc490c33beeba0895082b03be97f5bR53.

PR #5233, opened 2019-11-14
Don't emit timestamps when packaging to jar

raboof:
This makes the build more deterministic.

This is a continuation of the work in https://github.com/sbt/io/pull/58 , which I decided to get back into partly because Maven has been adding features to the core plugins, and it would be good to be on par :).

As discussed in https://github.com/sbt/io/pull/58, we should be careful about the risk of breaking incremental builds - I haven't found any such issues yet though. Note that this PR only affects the timestamps of entries in jars: timestamps of classes on the filesystem and the filesystem timestamp of the jar itself are not affected.

Discussing this with @eed3si9n, there could be special-purposes build configurations that (for example in order to avoid deleting/creating many files) compile classes directly into a jar rather than first to individual files. I haven't seen this configuration yet, but I suspect it wouldn't use `Defaults.packageTask`, but a custom `Package`/`IO` invocation. Any custom build configuration that requires non-static timestamps can pass `None` as the `time` parameter to keep the original behavior.

This PR requires https://github.com/sbt/io/pull/279

PR #5344, opened 2019-12-29
Reset timestamps to 1970-01-01 during packaging for repeatable builds

eed3si9n:
Ref https://github.com/sbt/io/pull/279
This is a resend of https://github.com/sbt/sbt/pull/5233

### original description

This is a continuation of the work in https://github.com/sbt/io/pull/58 , which I decided to get back into partly because Maven has been adding features to the core plugins, and it would be good to be on par :).

As discussed in https://github.com/sbt/io/pull/58, we should be careful about the risk of breaking incremental builds - I haven't found any such issues yet though. Note that this PR only affects the timestamps of entries in jars: timestamps of classes on the filesystem and the filesystem timestamp of the jar itself are not affected.

Discussing this with @eed3si9n, there could be special-purposes build configurations that (for example in order to avoid deleting/creating many files) compile classes directly into a jar rather than first to individual files. I haven't seen this configuration yet, but I suspect it wouldn't use `Defaults.packageTask`, but a custom `Package`/`IO` invocation. Any custom build configuration that requires non-static timestamps can pass `None` as the `time` parameter to keep the original behavior.

This PR requires https://github.com/sbt/io/pull/279

### DIFF
--- a/main-actions/src/main/scala/sbt/Package.scala
+++ b/main-actions/src/main/scala/sbt/Package.scala
@@ -70,7 +70,7 @@ object Package {
       val options: Seq[PackageOption]
   )
 
-  @deprecated("Please specify whether to use a static timestamp", "1.3.4")
+  @deprecated("Please specify whether to use a static timestamp", "1.4.0")
   def apply(conf: Configuration, cacheStoreFactory: CacheStoreFactory, log: Logger): Unit =
     apply(conf, cacheStoreFactory, log, None)
 
--- a/main-actions/src/main/scala/sbt/Package.scala
+++ b/main-actions/src/main/scala/sbt/Package.scala
@@ -163,7 +163,7 @@ object Package {
       homepage map (h => (IMPLEMENTATION_URL, h.toString))
     }: _*)
   }
-  @deprecated("Please specify whether to use a static timestamp", "1.3.4")
+  @deprecated("Please specify whether to use a static timestamp", "1.4.0")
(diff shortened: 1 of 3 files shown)


bollinger (commented):
I think the orElse(Some(0L)  is incorrect, causing files in the output jar to have a timestamp of 0.
Shouldn't None should be used when there is no SOURCE_DATE_EPOCH to get the pre 1.4.0 behaviour?
Peter.

eed3si9n:
> I think the orElse(Some(0L) is incorrect, causing files in the output jar to have a timestamp of 0.

Timestamp 0 is the intended behavior of sbt 1.4.0 to make the build more repeatable.

SethTisue:
commenting and saying "1970" so it comes up in search results :)

(since 1970 is the date all the files get)

1970 1970 1970 1970 !!!

dontgitit:
+1 to the previous comments and the linked play issue https://github.com/playframework/playframework/issues/10572, this change breaks etag computation in play. while the suggested fix in that thread (of setting an environment variable) does work, it's frustrating that there's no way to do it directly from sbt. it might be nicer if there was an sbt setting for this, `useEpochTimestamp in packageBin` or something of the sort? 

also, since the new code always provides `Some(0)` as a backup, it seems there's no way to go back to the old behavior of preserving asset timestamps; you must either choose a static date for all of them, or you get the epoch by default. this seems suboptimal - it would be ideal to preserve asset timestamps since they rarely change.

eed3si9n:
@dontgitit I've created a new issue for this - https://github.com/sbt/sbt/issues/6235

PR #5350, opened 2020-01-08
introduce SysProp sbt.testing.legacyreport

mrArkwright:
either create test reports with legacy file names (legacyreport=true) or with standard file names (legacyreport=false or omitted) but not both as suggested in #4451

### DIFF
--- a/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
+++ b/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
@@ -33,9 +33,10 @@ import sbt.protocol.testing.TestResult
  * report format.
  * @param outputDir path to the dir in which a folder with results is generated
  */
-class JUnitXmlTestsListener(val outputDir: String, logger: Logger) extends TestsListener {
+class JUnitXmlTestsListener(val outputDir: String, logger: Logger, legacyTestReport: Boolean)
+    extends TestsListener {
   // This constructor is for binary compatibility with older versions of sbt.
-  def this(outputDir: String) = this(outputDir, null)
+  def this(outputDir: String) = this(outputDir, null, false)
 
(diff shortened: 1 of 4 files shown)


testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
```diff
@@ -33,9 +33,10 @@ import sbt.protocol.testing.TestResult
  * report format.
  * @param outputDir path to the dir in which a folder with results is generated
  */
-class JUnitXmlTestsListener(val outputDir: String, logger: Logger) extends TestsListener {
+class JUnitXmlTestsListener(val outputDir: String, logger: Logger, legacyTestReport: Boolean)
```
eed3si9n: Could you move the `legacyTestReport: Boolean` to second position, and also supply an alternative constructor the the previous signature for backward binary compatibility please?

```
[error]  * method this(java.lang.String,sbt.util.Logger)Unit in class sbt.JUnitXmlTestsListener does not have a correspondent in current version
```

eed3si9n (changes requested):
Thanks for the contribution!

Some minor comments.

eed3si9n:
https://travis-ci.org/sbt/sbt/jobs/634194178 says `tests/junit-xml-report` scripted test failed. See https://github.com/sbt/sbt/tree/develop/sbt/src/sbt-test/tests/junit-xml-report.

mrArkwright:
Thanks for the review and the hints regarding the integrations tests. I included the proposed changes and fixed `tests/junit-xml-report`. The integration tests still seem to fail although I don't unterstand why.

eed3si9n:
Travis CI and/or Maven Central seems to be acting up.

eed3si9n (approved):

Issue #4870, opened 2019-07-16
Scripted strips quotation marks from commands


eatkins:
## steps
sbt version: any

## problem
If you add, say, `> set Compile / excludeFilter := "Bar.scala" || "Baz.scala"`, to a scripted file, the quotation marks are stripped and the command definition doesn't work.

## expectation
The quotation marks are not stripped.

jroper:
A work around that I stumbled across for this is to put the value in parenthesis:

```scala
> set Compile / excludeFilter := ("Bar.scala") || ("Baz.scala")
```

PR #5361, opened 2020-01-12
Cleanup scripted quotation mark parsing

eatkins:
In some cases scripted will either remove or add quotation marks which can break the intended command. This commit makes it possible to configure the scripted parser to stop it from doing that. I don't think this affects downstream users since I didn't change any default value but it should affect the scripted tests in the sbt repo.


eatkins:
Fixes https://github.com/sbt/sbt/issues/4870.

eed3si9n (approved):
Thanks!

PR #5362, opened 2020-01-12
Don't use last modified time of directories in doc

eatkins:
I noticed that sbt does a _lot_ of no-op docs in the sbt project.
Through some debugging, I determined that this was because the target
directory last modified time of some of the dependent projects would
change between runs. I'm not really sure why that was happening but
instead of computing the last modified time of the directory, we should
be checking the last modified time of the directory contents.

After this change a no-op `doc` in the sbt project returns in less than
one second on my mac. Before, it was more like 25-60 seconds (the upper
bound is one runs `doc` because `sbtRoot/doc` takes about a minute).

### DIFF
--- a/util-cache/src/main/scala/sbt/util/FileInfo.scala
+++ b/util-cache/src/main/scala/sbt/util/FileInfo.scala
@@ -13,6 +13,8 @@ import scala.util.control.NonFatal
 import sbt.io.{ Hash, IO }
 import sjsonnew.{ Builder, DeserializationException, JsonFormat, Unbuilder, deserializationError }
 import CacheImplicits.{ arrayFormat => _, _ }
+import sbt.nio.file._
+import sbt.nio.file.syntax._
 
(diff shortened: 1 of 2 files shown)


util-cache/src/main/scala/sbt/util/FileInfo.scala
```diff
@@ -209,6 +211,16 @@ object FileInfo {
       FileModified(file.getAbsoluteFile, IO.getModifiedTimeOrZero(file))
     def apply(file: File, lastModified: Long): ModifiedFileInfo =
       FileModified(file.getAbsoluteFile, lastModified)
+    def fileOrDirectoryMax(file: File): ModifiedFileInfo = {
```
eed3si9n: Maybe add Scaladoc here to document the fact that directory timestamp changes.

eed3si9n (approved):

PR #5363, opened 2020-01-12
Use implicit val rather than import

eatkins:
This import was causing an unused import warning during doc.

### DIFF
--- a/main/src/main/scala/sbt/internal/LibraryManagement.scala
+++ b/main/src/main/scala/sbt/internal/LibraryManagement.scala
@@ -10,18 +10,18 @@ package internal
 
 import java.io.File
 import java.util.concurrent.Callable
+
 import sbt.internal.librarymanagement._
 import sbt.librarymanagement._
 import sbt.librarymanagement.syntax._
-import sbt.util.{ CacheStore, CacheStoreFactory, Logger, Tracked, Level }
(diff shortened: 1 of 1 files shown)


main/src/main/scala/sbt/internal/LibraryManagement.scala
```diff
@@ -10,18 +10,19 @@ package internal
...
 import sbt.librarymanagement._
 import sbt.librarymanagement.syntax._
-import sbt.util.{ CacheStore, CacheStoreFactory, Logger, Tracked, Level }
+import sbt.util.{ CacheStore, CacheStoreFactory, Level, Logger, Tracked }
 import sbt.io.IO
 import sbt.io.syntax._
 import sbt.Project.richInitializeTask
 import sjsonnew.JsonFormat
 // This is to turn off "`state` is inside the if expression of a regular task." etc
-import sbt.dsl.LinterLevel.Ignore
```
eed3si9n, line 22: Could you also delete the comment right above it plz?

eed3si9n (changes requested):
Minor request.

main/src/main/scala/sbt/internal/LibraryManagement.scala
eatkins: done

eed3si9n (approved):

Issue #3245, opened 2017-06-03
cross building support and scripted test plugin compatibility


xuwei-k:
#### build.sbt

```scala
crossSbtVersions := Vector("0.13.15", "1.0.0-M6")
ScriptedPlugin.scriptedSettings
sbtPlugin := true
organization := "com.example"
name := "foo"
```

#### project/build.properties 

```
sbt.version=0.13.16-M1
```

#### project/scripted.sbt 

```scala
libraryDependencies += "org.scala-sbt" % "scripted-plugin" % sbtVersion.value
```

#### src/sbt-test/foo/bar/test

```
> compile
```

execute `sbt ^scripted`

## problem

```
java.lang.NoSuchMethodException: sbt.test.ScriptedTests$.run(java.io.File, boolean, [Ljava.lang.String;, java.io.File, [Ljava.lang.String;)
	at java.lang.Class.getMethod(Class.java:1786)
	at sbt.ScriptedPlugin$$anonfun$scriptedRunTask$1.apply(ScriptedPlugin.scala:39)
	at sbt.ScriptedPlugin$$anonfun$scriptedRunTask$1.apply(ScriptedPlugin.scala:38)
	at scala.Function1$$anonfun$compose$1.apply(Function1.scala:47)
	at sbt.$tilde$greater$$anonfun$$u2219$1.apply(TypeFunctions.scala:40)
	at sbt.std.Transform$$anon$4.work(System.scala:63)
	at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:228)
	at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:228)
	at sbt.ErrorHandling$.wideConvert(ErrorHandling.scala:17)
	at sbt.Execute.work(Execute.scala:237)
	at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:228)
	at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:228)
	at sbt.ConcurrentRestrictions$$anon$4$$anonfun$1.apply(ConcurrentRestrictions.scala:159)
	at sbt.CompletionService$$anon$2.call(CompletionService.scala:28)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #5367, opened 2020-01-12
Run all scripted tests using RunFromSourceMain

eatkins:
This PR does two big things:
1. Removes the launcher based scripted implementation and only used `RunFromSourceMain`. This has the benefit of not requiring recompilation of the compiler bridges every time. It also simplifies making new tests because it is no longer necessary to consider whether it is launcher or run from source based.
2. Use `publishLocalBin` instead of `publishLocal` so that we can avoid the expensive `doc` operation in `publishLocal`.


dwijnand:
Oh, man, this is an exciting PR!

But I'm a little confused.  You mentioned the lack of having to publish (🎉), but then switching back to publish bin only...  Is publishing still required?  The ideal would be that compiling is only required (or `products`, perhaps).

build.sbt
```diff
@@ -1202,18 +1158,20 @@ ThisBuild / scriptedPrescripted := { _ =>
...
     scriptedLaunchOpts := List("-Xmx1500M", "-Xms512M", "-server") :::
       (sys.props.get("sbt.ivy.home") match {
         case Some(home) => List(s"-Dsbt.ivy.home=$home")
         case _          => Nil
       }),
-    publishAll := { val _ = (publishLocal).all(ScopeFilter(inAnyProject)).value },
-    publishLocalBinAll := { val _ = (publishLocalBin).all(ScopeFilter(inAnyProject)).value },
-    aggregate in bintrayRelease := false
+    publishLocalBinAll := Def.taskDyn {
+      (Compile / (if (insideCI.value) publishLocal else publishLocalBin)).all(scriptedProjects)
```
dwijnand: Not sure if this should vary in CI.  Also it makes the name `publishLocalBinAll` a bit misleading...

eed3si9n:
@dwijnand I initially had the same feeling, but then I realized that the metabuild still requires resolution of sbt core artifacts. I had a PR to exclude sbt core artifacts from the metabuild (https://github.com/sbt/sbt/pull/4589), but I had to add it back for Coursier (https://github.com/sbt/sbt/pull/4614/commits/f999f6a62ea89fc8292e4be20564652b3757a04e).

eatkins:
I think that @eed3si9n is right that some publishing is necessary. I don't think it's necessary to do publishLocal in ci though. I think the reason I added that was that I forgot that I pulled https://github.com/sbt/sbt/pull/5364 out of this PR so publishLocalBin didn't work in ci. `RunSourceFromMain` works fine with the artifacts from `packageBin` (which was my first approach) rather than `publishLocalBin` but sbt is unable to resolve itself without publishing. On my local machine, I generally had published ivy files for the sbt sub projects so that things would work with `packageBin` but tests would fail in ci.

eatkins:
If I do end up rewriting the launcher, one thing I will probably do is skip ivy resolution if the project sbt version matches the launcher version unless there is some reason I am unaware of for this behavior. I think this would generally make `packageBin` work in scripted unless we wanted to use scripted to test if a launcher can resolve a different launcher version. That kind of test probably shouldn't be a scripted test though.

eed3si9n (approved):

eed3si9n:
For the record, we might need to flip back to using launcher when Scala version changes. It was straight forward to do once I found the `launch: Boolean` parameter - https://github.com/sbt/sbt/pull/5460

dwijnand:
> we might

Why?

eed3si9n:
I kept running into problem like this when I bumped the Scala version (I did it too in `project/plugins.sbt`):

```
[info] About to run tests:
[info]  * plugins/twirl
[info] Running 1 / 1 (100.00%) scripted tests with RunFromSourceMain
[info] Running plugins/twirl
[error] java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
[error] 	at sbt.internal.scriptedtest.ScriptedLauncher.launch(ScriptedLauncher.java:130)
[error] 	at sbt.RunFromSourceMain$.launch(RunFromSourceMain.scala:104)
[error] 	at sbt.RunFromSourceMain$.runImpl(RunFromSourceMain.scala:81)
[error] 	at sbt.RunFromSourceMain$.run(RunFromSourceMain.scala:69)
[error] btRoot /at sbt.RunFromSourceMain$.main(RunFromSourceMain.scala:57)
[error] 	at sbt.RunFromSourceMain.main(RunFromSourceMain.scala)
[error] Caused by: java.lang.reflect.InvocationTargetException
[error] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[error] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[error] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[error] 	at java.lang.reflect.Method.invoke(Method.java:498)
[error] 	at sbt.internal.scriptedtest.ScriptedLauncher.launchImpl(ScriptedLauncher.java:191)
[error] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[error] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[error] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[error] 	at java.lang.reflect.Method.invoke(Method.java:498)
[error] 	at sbt.internal.scriptedtest.ScriptedLauncher.launch(ScriptedLauncher.java:104)
[error] 	... 5 more
[error] Caused by: java.lang.NullPointerException
[error] 	at sun.net.util.URLUtil.urlNoFragString(URLUtil.java:50)
[error] 	at sun.misc.URLClassPath.getLoader(URLClassPath.java:522)
[error] 	at sun.misc.URLClassPath.getNextLoader(URLClassPath.java:494)
[error] 	at sun.misc.URLClassPath.getResource(URLClassPath.java:248)
[error] 	at java.net.URLClassLoader$1.run(URLClassLoader.java:366)
[error] 	at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
[error] 	at java.security.AccessController.doPrivileged(Native Method)
[error] 	at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
[error] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
[error] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:405)
[error] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:405)
[error] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
[error] 	at sbt.internal.XMainConfiguration.run(XMainConfiguration.scala:70)
[error] 	at sbt.xMain.run(Main.scala:39)
[error] 	... 15 more
[error] java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
```

It's potentially possible that there are some assumption around build's Scala version being the same as the freshly baked sbt's Scala version somewhere, and we just need to fix it but I just flipped to using launcher scripted test.

eatkins:
Rather than switching to launcher, it would be much better to figure out the underlying issue with the new scala version. What version are you using that is causing problems?

eed3si9n:
I agree that it would be better to fix the root cause if that's possible. I just wanted to test Scala 2.12.11-bin-e1e2bfe quickly this morning, so I was also grateful of the escape hatch - https://github.com/sbt/sbt/pull/5460

dwijnand:
https://github.com/sbt/sbt/pull/5461

scripted-sbt-redux/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
```diff
@@ -482,102 +403,62 @@ object ScriptedTests extends ScriptedRunner {
...
-      resourceBaseDirectory: File,
-      bufferLog: Boolean,
-      tests: Array[String],
-      bootProperties: File,
-      launchOpts: Array[String],
-  ): Unit = {
-    val logger = TestConsoleLogger()
-    val prescripted = ScriptedTests.emptyCallback
-    run(resourceBaseDirectory, bufferLog, tests, logger, bootProperties, launchOpts, prescripted)
-  }
```
eed3si9n, line 510: Because of `sbt ^scripted`, we need this entry point - https://github.com/sbt/sbt/issues/3245

scripted-sbt-redux/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
```diff
@@ -482,102 +403,62 @@ object ScriptedTests extends ScriptedRunner {
...
-      bufferLog: Boolean,
-      tests: Array[String],
-      bootProps: File,
-      launchOpts: Array[String],
-      prescripted: java.util.List[File],
-  ): Unit = {
-    runInParallel(baseDir, bufferLog, tests, bootProps, launchOpts, prescripted, 1)
-  }
-
   // This is used by sbt-scripted sbt 1.x
```
eed3si9n, line 406: We need the following version for https://github.com/sbt/sbt/blob/2b557851ca168cd6274adafb7db5d9ed32efb5ee/main/src/main/scala/sbt/ScriptedPlugin.scala#L105-L120

```scala
  // This is used by sbt-scripted sbt 1.x
  def runInParallel(
      baseDir: File,
      bufferLog: Boolean,
      tests: Array[String],
      bootProps: File,
      launchOpts: Array[String],
      prescripted: java.util.List[File],
      instances: Int
  ): Unit = ...
```

https://github.com/sbt/sbt/issues/5592

eed3si9n: I'll send a PR to recover these entry points.

PR #5368, opened 2020-01-13
Make IvyXml.writeFiles private[sbt]

eatkins:
The publishLocalBin doesn't make an ivy.xml file. In order to add that
functionality, we need to make IvyXml.writeFiles available at the
private[sbt] level.


eatkins:
See https://github.com/sbt/sbt/pull/5364.

eed3si9n (approved):

PR #5371, opened 2020-01-14
Apply google java formatting to sbt java source files

eatkins:
This adds https://github.com/swoval/sbt-source-format to the sbt build so that java source files are automatically formatted like we do with scala source files.


He-Pin:
How about the lightbend one？

eatkins:
I was unaware of the existence of the lightbend formatter. I prefer the caching model used in the swoval formatter (perhaps unsurprising since I wrote it). They have effectively the same api so it's trivial to switch plugins by just changing the value in `project/plugins.sbt`.

eatkins:
The swoval plugin is actually also a real world use case of some of the caching apis that I added for sbt 1.3.0, so it's nice to get more coverage of that code. The lightbend formatter is very similar to the official scalafmt plugin so it does not exercise the new apis.

eed3si9n (approved):
If the backends are using the same format, I am fine with either implementation.

eatkins:
> If the backends are using the same format, I am fine with either implementation.

Yep. I swapped the implementation and ran `clean; javafmt` and formatting ran but no diffs were generated.

Issue #5382, opened 2020-01-16
Ivy settings scripted tests are failing


eatkins:
## steps
sbt version: develop

## problem
A number of ivy settings tests seem to specify insecure resolvers that are now failing. Here is a sample failing CI build: https://travis-ci.org/sbt/sbt/jobs/637599650

## expectation
The tests use secure (or local) resolvers.

## notes

I am not familiar with ivy-settings.xml and there were no urls (secure or otherwise) or ssl related toggles in the scripted tests so I quickly gave up on debugging this. A sample broken ivysettings.xml file can be found here: https://github.com/sbt/sbt/blob/develop/sbt/src/sbt-test/dependency-management/ivy-settings-a/changes/scala-tools-ivysettings.xml.

eed3si9n:
I am kind of ok with removing any Ivy specific weird feature tests.

eatkins:
Here is why the tests started failing: https://blog.sonatype.com/central-repository-moving-to-https

eatkins:
Disabled in #5383.

PR #5383, opened 2020-01-16
Disable legacy ivy tests

eatkins:
These started failing recently. I think it is because they may be using
insecure resolvers. Since these tests are unlikely to be testing widely
used functionality, we can disable them for now. If we start seeing
related issues, we can consider actually fixing them.


eatkins:
See #5382.

eed3si9n (approved):

PR #5394, opened 2020-01-22
Use swoval source formatting plugin

eatkins:
This switches from using the combination of the scalameta sbt-scalafmt and
the swoval sbt-java-format plugins to just using the swoval
sbt-jvm-format plugin. The jvm format plugin has a pair of nice
features:

1. It provides a command jvmfmtAll (and jvmfmtCheckAll) which
effectively formats all of the java, scala and sbt source files in the
project, including those files that comprise the sbt build definition.
2. It integrates with supershell so that the user can see which files
are currently being formatted (the number of formatting tasks is
configurable but defaults to the number of available processors divided
by 2).

The api of the swoval scalafmt plugin is very similar to the scalameta
plugin. It also auto-updates by parsing the scalafmt version from the
scalafmt.conf file and uses coursier to fetch the jars needed to perform
formatting for the given version.


eatkins:
I will not be upset if this isn't accepted but I figured I'd make this PR as a proposal. Here is a demo of the supershell integration, which is my favorite feature of the plugin: [asciicast](https://asciinema.org/a/fJ6XKWGnc5xFOjAU0YzwBI6dt)

eed3si9n:
> I will not be upset if this isn't accepted but I figured I'd make this PR as a proposal

I guess the meta discussion point would be why not upstream the enhancement? It would be beneficial for the ecosystem if our knowledge can be harnessed in the popular plugins. I sent in the PR here https://github.com/scalameta/sbt-scalafmt/pull/46 using the old style caching. I'd be happy to see that replaced by newer style.

eatkins:
> I guess the meta discussion point would be why not upstream the enhancement?

Great question. The design goals and requirements of the swoval plugin are a bit different. I wanted to have a single plugin that does both java and scala formatting. I also was working under an sbt 1.3.0 minimum version requirement. Naively, I'd assume that the official plugin is expected to work for all versions in the 1.x series. Then there is the question of whether or not the official plugin should do java formatting as well. Also even though the functionality of the two plugins are similar, the implementations are completely different so it really is more of a question of whether they would accept a rewrite. I could however at least explain the supershell integration because I think they could do that without a full plugin rewrite (it's just a matter of making the format implementation a dynamic task that creates a task for each file and joins them).

eatkins:
I'm going to close this since it isn't strictly necessary and has been stagnant.

Issue #4496, opened 2018-12-20
<task>.all(<scope-filter>).value introduces nondeterminism


eed3si9n:
## steps

sbt-unidoc for example (https://github.com/sbt/sbt-unidoc/blob/v0.4.2/src/main/scala/sbtunidoc/BaseUnidocPlugin.scala) uses 

```
  lazy val allClasspathsTask = Def.taskDyn {
    val f = (unidocScopeFilter in unidoc).value
    dependencyClasspath.all(f)
  }
```

```
show Scalaunidoc/unidoc/fullClasspath
```

## problem

Depending on the machine you use unidoc, it returns different ordering of JARs

## expectation

Results are returned either deterministically relative to compilation semantics, or at least things are chunked by the same ordering of subprojects.

## notes

https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/ScopeFilter.scala

PR #5413, opened 2020-02-03
Use project resolvers before dependency resolvers in coursier

raboof:
csrResolvers.all evaluates all possible scopes in arbitrary order. This change
makes sure at least the project resolvers are placed before any resolvers from
dependency projects.

### DIFF
--- a/main/src/main/scala/sbt/coursierint/CoursierRepositoriesTasks.scala
+++ b/main/src/main/scala/sbt/coursierint/CoursierRepositoriesTasks.scala
@@ -135,9 +135,11 @@ object CoursierRepositoriesTasks {
     Def.taskDyn {
       val s = state.value
       val projectRef = thisProjectRef.value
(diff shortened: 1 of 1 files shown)


eed3si9n:
@raboof Nice. Do you have local test result or something that showed that classpath changed?

dwijnand:
See https://github.com/akka/alpakka/issues/2121 (and then Coursier).

raboof:
Indeed that's what prompted this change, and I could see that `csrRecursiveResolvers` was different between machines where this build did and didn't work.

Of course this is not the only thing that's wrong with that picture, but it's at least nicer behavior on the sbt side to put the 'direct' resolvers first :)

eed3si9n (approved):

eed3si9n:
I guess this is another example of https://github.com/sbt/sbt/issues/4496

PR #5421, opened 2020-02-06
[1.3] Fix Coursier classpath order

### LINKED ISSUE #5282: Change in classpath ordering using Coursier
## steps
sbt version: 1.3.4

## problem

(Apologies as I file this quickly, without a minimised reproduction.)

Both Play and Lagom implement a strategy of having alternative implementations for something (e.g. Akka HTTP vs Netty server backend in Play, akka-persistence-jdbc vs akka-peristence-jdbc in Lagom) by defining distinct jars with distinct classes, but which define the same config key:

```hocon
play.server {

  # The server provider class name
  provider = "play.core.server.NettyServerProvider"
```

```hocon
play {

  server {
    # The server provider class name
    provider = "play.core.server.AkkaHttpServerProvider"
```

```hocon
akka.persistence.journal.plugin = cassandra-journal
akka.persistence.snapshot-store.plugin = cassandra-snapshot-store
```

```hocon
akka.persistence.journal.plugin = jdbc-journal
akka.persistence.snapshot-store.plugin = jdbc-snapshot-store
```

For Play, the intent is that the Akka HTTP server is the default, and there are instructions to opt-out of that and opt-in to Netty.  For Lagom, I'm less sure, but I think there's no default and you just pick one _OR_ the other.

But after that strategy was set, Play's `Play-Test` module was made to depend on both, which means that the order of the jars defines which configuration wins.  Unfortunately, the ordering was accidentally correct in terms of our desires, meaning the config for Akka HTTP was seen, so we didn't become aware of this mistake.

The upgrade to sbt 1.3 (and, thus, from Ivy to Coursier) seems to switch the ordering which has revealed the mistake.

## expectation

Now, I say "mistake" because I don't think it's behaviour that should be relied upon.  A practice that is done, and should be possible, is to prepend jars to the classpath, the practice of "monkey-patching the classpath".  But I think the order of the classpath from managed dependency resolution shouldn't be relied upon.

So _I_ don't consider this a regression, but in speaking with @retronym, he advised that we should consider changing Coursier to maintain the previous behaviour, as it's still early days in sbt 1.3's adoption and we might be saving ourselves a lot of trouble down the line by changing this back for everyone.

## notes

* https://github.com/playframework/playframework/pull/9838 fixing our mistake
* https://github.com/playframework/playframework/pull/9877 adding complexity to our build by introducing asbt 1.3 as a testing variant... 😭
* https://github.com/lagom/lagom/pull/2523 fixing the same type of mistake, in Lagom

dwijnand:
Fixes #5282

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,7 @@ import scala.xml.transform.{ RewriteRule, RuleTransformer }
 import scala.util.Try
 
 ThisBuild / version := {
-  val v = "1.3.7-SNAPSHOT"
+  val v = "1.3.9-SNAPSHOT"
   nightlyVersion.getOrElse(v)
 }
 ThisBuild / scalafmtOnCompile := !(Global / insideCI).value
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -112,7 +112,7 @@ object Dependencies {
   def addSbtZincCompileCore(p: Project): Project =
     addSbtModule(p, sbtZincPath, "zincCompileCore", zincCompileCore)
 
-  val lmCoursierVersion = "2.0.0-RC5-3"
+  val lmCoursierVersion = "2.0.0-RC6-1"
   val lmCoursierShaded = "io.get-coursier" %% "lm-coursier-shaded" % lmCoursierVersion
 
   val sjsonNewScalaJson = Def.setting {
--- a/sbt/src/sbt-test/dependency-management/cp-order/build.sbt
+++ b/sbt/src/sbt-test/dependency-management/cp-order/build.sbt
@@ -1,0 +1,5 @@
+scalaVersion := "2.13.1"
+csrConfiguration := csrConfiguration.value.withCache(target.value / "coursier-cache")
+
+libraryDependencies += "com.typesafe.play" %% "play-test" % "2.8.0-RC1" % Test // worked around in 2.8.0
+libraryDependencies += "org.scalatest"     %% "scalatest" % "3.0.8"     % Test
--- a/sbt/src/sbt-test/dependency-management/cp-order/src/test/scala/t/UnitSpec.scala
+++ b/sbt/src/sbt-test/dependency-management/cp-order/src/test/scala/t/UnitSpec.scala
@@ -1,0 +1,20 @@
+package t
+
+import com.typesafe.config.ConfigFactory
+import org.scalatest.{ MustMatchers, WordSpec }
+
+class UnitSpec extends WordSpec with MustMatchers {
+  def conf = ConfigFactory.defaultReference()
+
+  "Config" should {
+    "return Akka HTTP server provider" in {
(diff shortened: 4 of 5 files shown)


dwijnand:
Forward ported in #5423.

eed3si9n (approved):
LGTM

hygt:
@dwijnand FYI the new order makes SparkSession's initialization fail in tests for some of our projects (but not all...) I assume because the configuration files are not merged in the right order?

```
17:59:53  [pool-11-thread-1] ERROR org.apache.spark.SparkContext - Error initializing SparkContext.
17:59:53  java.lang.NullPointerException
17:59:53  	at org.apache.spark.network.server.TransportServer.close(TransportServer.java:153)
17:59:53  	at org.apache.spark.network.util.JavaUtils.closeQuietly(JavaUtils.java:53)
17:59:53  	at org.apache.spark.network.server.TransportServer.<init>(TransportServer.java:79)
17:59:53  	at org.apache.spark.network.TransportContext.createServer(TransportContext.java:114)
17:59:53  	at org.apache.spark.rpc.netty.NettyRpcEnv.startServer(NettyRpcEnv.scala:119)
17:59:53  	at org.apache.spark.rpc.netty.NettyRpcEnvFactory.$anonfun$create$1(NettyRpcEnv.scala:465)
17:59:53  	at org.apache.spark.rpc.netty.NettyRpcEnvFactory.$anonfun$create$1$adapted(NettyRpcEnv.scala:464)
17:59:53  	at org.apache.spark.util.Utils$.$anonfun$startServiceOnPort$2(Utils.scala:2269)
17:59:53  	at scala.collection.immutable.Range.foreach$mVc$sp(Range.scala:158)
17:59:53  	at org.apache.spark.util.Utils$.startServiceOnPort(Utils.scala:2261)
17:59:53  	at org.apache.spark.rpc.netty.NettyRpcEnvFactory.create(NettyRpcEnv.scala:469)
17:59:53  	at org.apache.spark.rpc.RpcEnv$.create(RpcEnv.scala:57)
17:59:53  	at org.apache.spark.SparkEnv$.create(SparkEnv.scala:250)
17:59:53  	at org.apache.spark.SparkEnv$.createDriverEnv(SparkEnv.scala:185)
17:59:53  	at org.apache.spark.SparkContext.createSparkEnv(SparkContext.scala:257)
17:59:53  	at org.apache.spark.SparkContext.<init>(SparkContext.scala:424)
17:59:53  	at org.apache.spark.SparkContext$.getOrCreate(SparkContext.scala:2520)
17:59:53  	at org.apache.spark.sql.SparkSession$Builder.$anonfun$getOrCreate$5(SparkSession.scala:935)
17:59:53  	at scala.Option.getOrElse(Option.scala:189)
17:59:53  	at org.apache.spark.sql.SparkSession$Builder.getOrCreate(SparkSession.scala:926)
...
```

dwijnand:
It's a plausible reason, for sure.

Issue #5038, opened 2019-09-05
`consoleProject` fails sbt 1.3.0 from sbt 1.2.8 launcher with `[error] (run-main-0) sbt.internal.inc.InvalidScalaInstance: Scala instance doesn't exist or is invalid:`


jsoref:
## steps
this isn't a real command, you just need to have installed the sbt launcher before sbt 1.3.0 was available:
```sh
brew install sbt@1.2.8
```
`build.sbt`:
```
scalaVersion := "2.12.8" // this probably doesn't matter
```
`project/build.properties`:
```
sbt.version=1.3.0
```

## problem
Start sbt (1.2.8 launcher, asking to use sbt 1.3.0) and ask it to trigger `consoleProject`:
```sh
sbt -Dsbt.ci=true -Dsbt.ivy=true -Dsbt.banner=false consoleProject
```

Or more precisely:
```sh
java -Xms1024m -Xmx1024m -XX:ReservedCodeCacheSize=128m -XX:MaxMetaspaceSize=256m -Dsbt.version=1.3.0 -jar /usr/local/Cellar/sbt/1.2.8/libexec/bin/sbt-launch.jar -Dsbt.ivy=true -Dsbt.ci=true -Dsbt.banner=false consoleProject
```
(Note: the sbt.ivy bit is a red-herring, it was suggested as a way to rule out coursier, the problem is just the launcher)

```

[error] (run-main-0) sbt.internal.inc.InvalidScalaInstance: Scala instance doesn't exist or is invalid: 
[error]     version 2.12.8, library jars: [Ljava.io.File;@5e1c2fa2, compiler jar: ~/.sbt/boot/scala-2.12.8/lib/scala-compiler.jar, other jars: ~/.sbt/boot/scala-2.12.8/lib/jline.jar, ~/.sbt/boot/scala-2.12.8/lib/scala-reflect.jar, ~/.sbt/boot/scala-2.12.8/lib/scala-xml_2.12.jar
[error] sbt.internal.inc.InvalidScalaInstance: Scala instance doesn't exist or is invalid: 
[...]

PR #5251, opened 2019-11-28
Fix consoleProject for scala 2.13

steinybot:
Scopes `scalaCompilerBridgeSource` to `consoleProject` and sets the default to use:
```
ZincLmUtil.getDefaultBridgeModule(
      appConfiguration.value.provider.scalaProvider.version
    )
```
instead of:
```
ZincLmUtil.getDefaultBridgeModule(scalaVersion.value)
```

I didn't add any notes, I wasn't sure where to put them. There weren't as many as I was expecting.

I considered adding some validation to check whether the bridge matches the scala version but it didn't really seem worth it.

Fixes #5250.

PR #5422, opened 2020-02-06
[1.3] Fix consoleProject

dwijnand:
`scalaCompilerBridgeSource` will use the session's scalaVersion, which
needn't match sbt's.

For example, try running `consoleProject` after running ++2.13.1.


eed3si9n (approved):

eed3si9n:
I am guessing that this would fix https://github.com/sbt/sbt/issues/5038

dwijnand:
Oops, my browser state was stale.

Yeah, this is now a backport of #5251.

dwijnand:
> I am guessing that this would fix #5038

I'm not sure about that one.  That one looks launcher related.

jroper:
I can confirm it doesn't fix #5038, I've just tested with sbt 1.3.9 and am still getting the InvalidScalaInstance error.

PR #5435, opened 2020-02-15
Tweak Def.displayRelative

dwijnand:
My driving use case was LocalProject: it should display the same as a fully
resolved ProjectRef.  Similarly ThisBuild should display as such.


main-settings/src/main/scala/sbt/Def.scala
```diff
@@ -132,13 +133,19 @@ object Def extends Init[Scope] with TaskMacroExtra {
...
+      case BuildRef(b)      => if (b == current.build) loop(ThisBuild) else displayRef(ref)
+      case RootProject(b)   => if (b == current.build) loop(LocalRootProject) else displayRef(ref)
+      case LocalProject(p)  => if (p == current.project) "" else p
+      case ThisBuild        => "ThisBuild"
+      case LocalRootProject => "<root>"
+      case ThisProject      => "<this>"
     }
+    val str = loop(project)
+    if (trailingSlash && !str.isEmpty) s"$str /"
+    else str
```
eed3si9n, line 148: What are the downstream of `displayRelative`? i.e. how do they show up in the UI?

dwijnand: I think I hit it through `Extracted` when running a task out of graph (in an early version of scala/scala#8662, before I switched to using `set`) where I used `LocalProject` and got the `{<this>}bar / javaOptions` output rather than just `bar / javaOptions`.

dwijnand:
Any problems here?  It's a relative display, so this is about making `LocalProject("bar")` display as `baz /` rather than `{<this>}bar /`.

eed3si9n (approved):

PR #5436, opened 2020-02-15
Document parts of AList & components of TupleSyntax

dwijnand:
Also, extract the AList.Split type lambda helper for call-site reuse.

### DIFF
--- a/internal/util-collection/src/main/scala/sbt/internal/util/AList.scala
+++ b/internal/util-collection/src/main/scala/sbt/internal/util/AList.scala
@@ -91,9 +91,12 @@ object AList {
     ): N[P[A]] = f(a)
   }
 
-  type ASplit[K[L[x]], B[x]] = AList[λ[L[x] => K[(L ∙ B)#l]]]
+  /** Example: calling `AList.SplitK[K, Task]#l` returns the type lambda `A[x] => K[A[Task[x]]`. */
+  sealed trait SplitK[K[L[x]], B[x]] { type l[A[x]] = K[(A ∙ B)#l] }
 
-  /** AList that operates on the outer type constructor `A` of a composition `[x] A[B[x]]` for type constructors `A` and `B`*/
+  type ASplit[K[L[x]], B[x]] = AList[SplitK[K, B]#l]
+
+  /** AList that operates on the outer type constructor `A` of a composition `[x] A[B[x]]` for type constructors `A` and `B`. */
   def asplit[K[L[x]], B[x]](base: AList[K]): ASplit[K, B] = new ASplit[K, B] {
     type Split[L[x]] = K[(L ∙ B)#l]
 
--- a/main-settings/src/main/scala/sbt/Structure.scala
+++ b/main-settings/src/main/scala/sbt/Structure.scala
@@ -462,14 +462,17 @@ object Scoped {
   ) {
 
     type App[T] = Initialize[Task[T]]
+
+    /** A higher-kinded function, where each parameter shares the same type constructor `M[_]`. */
     type Fun[M[_], Ret]
 
+    /** Convert the higher-kinded function to a Function1.  For tuples that means call `.tupled`. */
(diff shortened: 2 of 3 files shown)


main-settings/src/main/scala/sbt/Structure.scala
```diff
@@ -462,14 +462,17 @@ object Scoped {
...
     type Fun[M[_], Ret]
 
+    /** Convert the higher-kinded function to a Function1.  For tuples that means call `.tupled`. */
     protected def convert[M[_], Ret](f: Fun[M, Ret]): K[M] => Ret
 
     private[this] val inputs: K[App] = a.transform(keys, λ[ScopedTaskable ~> App](_.toTask))
 
     private[this] def onTasks[T](f: K[Task] => Task[T]): App[T] =
-      Def.app[λ[L[x] => K[(L ∙ Task)#l]], Task[T]](inputs)(f)(AList.asplit[K, Task](a))
+      Def.app[AList.Split[K, Task]#l, Task[T]](inputs)(f)(AList.asplit[K, Task](a))
```
eed3si9n: Not sure if this change makes the code more readable.

dwijnand: Do you mean this use site, or the introduction of `Split` as a whole?

The reason I introduced it is to avoid having to do any of the alternatives to define this type parameter (and you _have_ to define the type parameter, as the type inferencer will never infer this type parameter), as in: use kind-projector, use Jason's original anonymous type lambda thing, or define your own type alias and then use it.

It follows the lead set by tuple2/tuple3/etc which have the `T2K`/`T3K`/etc helpers.

eed3si9n: > Do you mean this use site, or the introduction of `Split` as a whole?

`Split` as a whole I guess. Isn't kind projector the current semi-standard notation on these weird type things?

dwijnand: That's a little bit true.  IntelliJ kind of "supports" it now.  But Dotty only implements parts of it, and with some changes to syntax.

But, anyways, kind projector is for defining inline type lambda.  You can still use aliases, and that might be clearer.  (It's as clear and unclear as other forms of type aliases are.)

internal/util-collection/src/main/scala/sbt/internal/util/AList.scala
```diff
@@ -91,9 +91,12 @@ object AList {
     ): N[P[A]] = f(a)
   }
 
-  type ASplit[K[L[x]], B[x]] = AList[λ[L[x] => K[(L ∙ B)#l]]]
+  /** Example: calling `AList.Split[K, Task]#l` returns the type lambda `A[x] => K[A[Task[x]]`. */
+  sealed trait Split[K[L[x]], B[x]] { type l[A[x]] = K[(A ∙ B)#l] }
```
eed3si9n:
```suggestion
  /** Example: calling `AList.SplitK[K, Task]#l` returns the type lambda `F[x] => K[F[Task[x]]`. */
  sealed trait SplitK[K[F0[A]], F1[A]] { type l[F2[A]] = K[(F2 ∙ F1)#l] }
```

Following `T2K`, adding `K` at the end make it a bit more clearer that this is not a datatype?

dwijnand: I think the K there stands for "higher-kinded", so T2K is "a higher-kinded tuple 2".

I think maybe the 100% correct fix would be to rename the existing `ASplit` alias to `ASplitList`, so this one could be `ASplit`, but I didn't want to repurpose an alias.  So I just pushed the name used in the local alias in FullInstance instead.

I guess you could argue maybe that the K can mean "split the KList" (the higher-kinded (heterogenous) list).  I'll push that.  I wouldn't start using new names for the type constructors though.

eed3si9n (approved):

PR #5439, opened 2020-02-16
Introduce Taskable & toTaskable

dwijnand:


dwijnand:
```
[error] java.lang.NoSuchMethodError: sbt.Classpaths$.concatSettings(Lsbt/SettingKey;Lsbt/SettingKey;)Lsbt/internal/util/Init$Initialize;
```

```
[error] java.lang.NoSuchMethodError: sbt.Defaults$.relativeMappings(Lsbt/ScopedTaskable;Lsbt/ScopedTaskable;)Lsbt/internal/util/Init$Initialize;
```

Huh.  I had understood from Josh way back when that the JVM can just manage these class hierarchy changes.  That's how we got the `SettingDefinition` change in `Project#settings` done without breaking changes in 0.13.  I don't understand why it fails here, though.  But thank God for the tests that found these...

dwijnand:
Hmm.  I thought it might be an object vs class thing (`Classpaths.concatSettings ` being an object method means there's also a static-to-instance method calls, while `Project#settings` is instance only), but I didn't even get to that as I could fully reproduce this with just classes.

I think the `settings` change was safe because it was the generic type of the varargs seq, so new sbt was happy to start reading the `SettingDefinition` supertype in place of `Setting[_]`.

So all the RichTaskable/TupleSyntax exemptions should be fine, because it's the type parameter of `K`, the klist.

I'll recover and forward the old generics.

main-settings/src/test/scala/sbt/TupleSyntaxTest.scala
```diff
@@ -0,0 +1,18 @@
...
+
+package sbt.test
+
+import sbt._
+import sbt.Def.Initialize
+import sbt.TupleSyntax._
+
+object TupleSyntaxTest {
+  def t1[T](a: SettingKey[T], b: TaskKey[T], c: Initialize[T], d: Initialize[Task[T]]) = {
+    (a, b, c.toTaskable, d.toTaskable).map((x: T, y: T, z: T, w: T) => "" + x + y + z + w)
```
eed3si9n, line 16: Could we somehow remove `toTaskable` method, and put the implicit conversion under `import sbt.TupleSyntax._`?

dwijnand: No, I tried, hard, with a bunch of abstract types and so forth (because `Task` doesn't exist until `task-sytem` etc).  The problem is that inheritance makes it impossible.

`SettingKey[T] <: Initialize[T]`
`SettingKey[T] <: Taskable[T]`
`TaskKey[T] <: Initialize[Task[T]]`
`TaskKey[T] <: Taskable[T]`

So you can't define a class hierarchy and use that as the `ST` in `TupleSyntax` that works for SettingKey, TaskKey, Initialize[Task[T]] and any old Initialize[T].  And implicit conversions don't chain automatically.  And obviously we don't want to do all the combinations of those 4 in tuples2-tuple11.

The (prioritised) implicit conversions in `Taskable`'s implicit scope will trigger for things like `Classpaths.concat`, because `Taskable` is the target type, not a prerequisite for the tuple enrichments.

eed3si9n: ok.

eed3si9n (approved):

Issue #2232, opened 2015-10-02
Initialization ordering of subprojects is undefined


eed3si9n:
Ref #2090
## steps

Use sbt 0.13.9. Create 

``` scala
import sbt._
import Keys._

object FooBuild extends Build {
  lazy val intTask = taskKey[Int]("int task")
  val api = project.
    settings(
      intTask := 0
    )
  val database = project
  val web = project.
    dependsOn(api, database).
    settings{
      intTask in api := 1
    }
  val foo = project
}
```

Run `show api/intTask`. It should display:

```
[info] 0
```

Next comment out `val foo = project`:

``` scala
  // val foo = project
```

Run `show api/intTask`. It should display:
## problem

```
[info] 1
```

Note that `foo` subproject has no relationship with `api` or `web`.
It seems like the initialization ordering of the subproject is not clearly defined.
## expectation

There should be a clear ordering on which one wins, or forbid initializing other subproject's keys.
## note

I'm not sure if this is the smoking gun, but... https://github.com/sbt/sbt/blob/0.13.9/main/src/main/scala/sbt/Load.scala#L319

``` scala
      (new sbt.PartBuildUnit(unit, defined.map(d => (d.id, d)).toMap, rootProjects.map(_.id), buildSettings(unit)), externals)
```

`toMap` has no notion of ordering. This is transformed into a `LoadedBuildUnit`. See code around https://github.com/sbt/sbt/blob/0.13.9/main/src/main/scala/sbt/Load.scala#L385-L408. Somehow this `Map` it turned into `Seq`, which I suspect non-determinism.

eed3si9n:
A potential workaround at least for #2090:

``` scala
import sbt._
import Keys._

object FooBuild extends Build {
  lazy val intTask = taskKey[Int]("int task")
  lazy val preIntTask = taskKey[Unit]("pre int task")
  lazy val helloTask = taskKey[Unit]("hello task")
  val api = project.
    settings(
      intTask := 0,
      intTask <<= intTask dependsOn preIntTask,
      preIntTask <<= preIntTask or Def.task {()}
    )
  val database = project
  val web = project.
    dependsOn(api, database).
    settings(
      helloTask := { println("hello") },
      preIntTask in api := { helloTask.value }
    )
  val foo = project
}
```

`preIntTask` is defined twice but protected by `or`.

eed3si9n:
Perhaps we should emulate top-to-bottom semantics within a same file by keeping the position around.

[...]

PR #5447, opened 2020-02-22
Make bare setting loading order alphabetical

### LINKED ISSUE #2697: Order of settings evaluation is surprising and not intuitive
Given a  `build.sbt` like the following

```
val p = project.settings(name := "foo")
name in p := "bar"
```

I would expect `p/name` to show `bar`, it shows `foo`

If I later add `z-build.sbt` like so:

```
val p = ProjectRef(file("."), "p")
name in p := "baz"
```

I would expect `p/name` to show `baz`, but it still shows `foo`

```
[pfnguyen@galactica s] $ sbt p/name
[info] Loading global plugins from C:\Users\pfnguyen\.sbt\0.13\plugins
[info] Set current project to s (in build file:/C:/Users/pfnguyen/src/s/)
[info] foo
```

Generally, I would expect settings declarations that come after the initial project declaration to take precedence, especially when it's split across files. In this contrived example, it's "why would you ever do that", but in my actual use-cases, I often have a generated sbt file which I want to later make changes to. When making those changes, I do not want to change the generated sbt file, but an auxiliary file, like `z-build.sbt` to ensure correct ordering.

Also, this is very applicable to plugin settings, e.g. if one were to do

```
val p = project.settings(somePluginSettings) // sbt-android still requires this format because autoplugins is underengineered  ;-)
anyAbovePluginSetting := ... // will get ignored if it was set by somePluginSettings
```

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/2697
Ref https://twitter.com/not_xuwei_k/status/1230140477959286848

Note that this is could potentially break an existing build that was relying on previous behavior, which seem to return `build.sbt` at the end if you have `a.sbt`, `b.sbt`, `c.sbt`, and `build.sbt`.


dwijnand:
Fixes #2232 too?

dwijnand:
Also:

<blockquote class="twitter-tweet"><p lang="en" dir="ltr">Is ÿÿÿÿÿÿÿÿÿÿ~~~~dbuild~defs.sbt still last?<a href="https://t.co/P1SRncRm4O">https://t.co/P1SRncRm4O</a></p>&mdash; Dale Wijnand (@dwijnand) <a href="https://twitter.com/dwijnand/status/1231110273823584257?ref_src=twsrc%5Etfw">February 22, 2020</a></blockquote>

eed3si9n:
I don't think it was ever the last:

```scala
> show Compile/scalacOptions
[info] * c
[info] * b
[info] * a
[info] * ÿÿÿÿÿÿÿÿÿÿ
[info] * build
```

with my patch it will be:

```scala
> show Compile/scalacOptions
[info] * a
[info] * b
[info] * build
[info] * c
[info] * ÿÿÿÿÿÿÿÿÿÿ
```

eed3si9n:
If #2232 is still around, this won't fix it since this just changes the ordering of `*.sbt` files.

dwijnand:
> since this just changes the ordering of `*.sbt` files.

Isn't the criteria for #2232 "There should be a clear ordering on which one wins, or forbid initializing other subproject's keys."?

dwijnand:
> I don't think it was ever the last:

I find that very surprising, given dbuild introspects the build...  Hmm, maybe it was discovered that naming convention didn't work, which is why they swiched to doing things `onLoad`?  Maybe we should validate or talk to @cunei about it.

eed3si9n:
Toni said:

> This is not what I remember; the `.sbt` files used to be loaded in alphabetical order. It is the reason why the dbuild-specific file is called `ÿÿÿÿ...`: to be able to patch things after everything else loads. I don't know when/if sbt changed loading order, but this is my recollection.

dwijnand:
I noticed yesterday from the output of loading that it definitely has build.sbt at the end.  I can't remember anything being changes around sbt 1.0 times, so maybe this changed back when "no blankies" landed?

PR #5461, opened 2020-03-04
Test Scala 2.12.11-bin-e1e2bfe

dwijnand:
See scala/scala-dev#676

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -64,6 +64,7 @@ def commonBaseSettings: Seq[Setting[_]] = Def.settings(
   ),
   scalaVersion := baseScalaVersion,
   componentID := None,
+  resolvers += "scala-integration" at "https://scala-ci.typesafe.com/artifactory/scala-integration/",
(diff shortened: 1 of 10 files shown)


dwijnand:
```
[info] [error] java.lang.NoSuchMethodException: xsbt.CompilerInterface.newCompiler([Ljava.lang.String;, xsbti.compile.Output, xsbti.Logger, xsbti.Reporter)
```

Problem?

eed3si9n:
> xsbt.CompilerInterface.newCompiler

That's part of compiler bridge, not the compiler. In the case of https://github.com/sbt/sbt/blob/0b12862caf63d1c23449391be6329d947b474157/sbt/src/sbt-test/dependency-management/compiler-bridge-binary/build.sbt, it's using "compiler-bridge_2.12" % "1.3.0" for **2.12.10**. sbt not finding a method in the already-built bridge seems like a wiring issue, rather than something bincompat issue.

dwijnand:
Looks like this is mostly good otherwise.

PR #5463, opened 2020-03-05
Make ScriptedLauncher support Scala pre-releases

dwijnand:
Also add a dash of sanity checks here and there.

(From #5461.)


dwijnand:
@eed3si9n would you be interested in me getting this passing and merged, or not so much?

eed3si9n:
@dwijnand I think this is very useful for Scala team validating Scala patch candidates.

eed3si9n (approved):

PR #5464, opened 2020-03-06
add Scala nightly support

martijnhoekstra:
ad-hoc implementation of nightly support.

++ foo-nightly now means latest nightly from branch foo.x
e.g. ++ 2.13-nightly finds and sets the latest published nightly for branch 2.13.x

Coincidentally ++ is what travisCI uses for its build matrix, so adding 2.13-nightly/2.12-nightly to your travis build matrix will build against latest nightly.

WDYT?


eed3si9n:
I think it's an interesting idea, but I am not sure who the target audience is.
Is it for library authors, or more for scala/scala hackers to quickly verify community projects?

martijnhoekstra:
The community build has always been valuable as a canary for both sides. For library maintainers, it can be useful to keep the nightlies in their CI matrix to get notified early if something changes. That may result in bug reports for scala, which may prevent those breakages from happening.

To quickly check a community project it's kind of nice too. It would make a currently manual task easier. It was manual anyway, so it's not that big a deal, but easier is good.

eed3si9n:
ok I am open to trying this.
I'd also be curious to hear what others might say. /cc @SethTisue

main/src/main/scala/sbt/Cross.scala
```diff
@@ -261,6 +266,14 @@ object Cross {
 
     type ScalaVersion = String
 
+    def resolveNightly(major: String): String = {
+      val url =
+        s"https://raw.githubusercontent.com/scala/community-builds/$major.x/nightly.properties"
+      val props = new java.util.Properties
+      props.load(new java.net.URL(url).openStream)
+      props.getProperty("nightly").ensuring(_ != null)
+    }
```
eed3si9n: It's highly unusual for the Scala version to be determined by a command that directly goes out to the Internet.
I think there should be an ample logging around here. Like in `-client` I say `"entering *experimental* thin client - BEEP WHIRR"` - https://github.com/sbt/sbt/blob/v1.3.8/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala#L194

Log suggestion:

```
*experimental* nightly cross build - BEEP WHIRR
resolving nightly Scala version for $major from GitHub scala/community-builds...
2.13.2-bin-7da8ce2 found!
```

martijnhoekstra: Sounds good.

I think it might also benefit from some caching -- e.g. a cache that times out after 10 minutes -- and thinking about the unhappy path when it can't be resolved, which is now mostly ignored.

SethTisue:
I like it, w/ the logging Eugene suggests

not sure about naming. we do informally refer to the builds are "nightlies", but they aren't literally nightly — the on-every-merge builds are often more frequent than that, and the community-build-blessed "nightlies" are usually substantially less frequent, like every 2 weeks or so (more often when a release is coming up)

I can see someone wanting the community-build-blessed one, but I can just as easily see someone wanting the latest-merged-PR and not want to wait two weeks

and I'm not saying this is required, but it would be cool: when @dwijnand did his version of this in https://github.com/dwijnand/scala-runners, he also added support for "latest commit on PR number 1234"

so, that's the broader world of "Scala versions people might want", and I think alignment with Scala-runners would be neat

martijnhoekstra:
I'm not sure what better names we have than nightly, but I'm open to suggestions. Mine would be

2.13-nightly = latest mergely on 2.13.x
2.13-unstable = latest community-built blessed on 2.13.x

I'll take a look at what scala-runners does too. Does it do (or need to do) anything for determining the major version of PR builds?

martijnhoekstra:
@SethTisue where would I get the latest community build version vs the latest mergely?

martijnhoekstra:
scala-runners has the concepts

* Latest release for major
* Latest mergely for major
* Latest build for PR

They all seem useful. Useful enough to all be supported in sbt?

|concept|scala-runners|proposed sbt|
|---|---|---|
|latest stable|213|2.13-latest|
|latest mergely|213x|2.13-nightly|
|latest with community build passed|-|2.13-unstable|
|pr|--scala-pr <PR number>|pr-xxxx|

Parsing too much in the version string feels pretty hacky. I mean this whole thing is pretty hacky but that's no reason to go overboard. Maybe pr is better served with its own task? Parsing the version string is needed to make ++ work with the different flavors of "latest" and I can imagine all of them being useful in a CI matrix. Travis uses ++ for it's build matrix, so parsing it is. But is there a use case for PR's in the CI matrix that is important enough to justify parsing the version string?

dwijnand:
Actually `-213x` is using the `nightly.properties` in CB, which isn't exactly "latest mergely" (the "nightly" part is a lie).  I would like to add that option, though.

And, yes, I would absolutely want sbt and scala-runners to use the same name convention, but for scala-runners I'd like them to be as short as possible and without an argument.

martijnhoekstra:
Do you mean scala-runners has a long form version string too? If so, then yes, let's align that.

dwijnand:
Long-form version string?

eed3si9n:
> Long-form version string?

By which I think @martijnhoekstra means scala-runners supporting both `2.13-nightly` (long form) and `213x` (short form).

I personally think what `++` command _should_ align is the correspondence with `set every scalaVersion := "2.13.2-bin-7da8ce2"`, and I am feeling more wary of adding extra stuff beyond that. If we're adding latest etc, it's only a matter of time the user would get confused why one notation "works on Travis CI but not in build.sbt using scalaVersion."

dwijnand:
Oh.  Yeah, scala-runners could have longer aliases that sbt uses.  But, ideally, we don't have duplicate identifiers.

eed3si9n:
It's more of a commentary on scala-runners notation, but I don't know if using `213x` to denote latest mergely is good.

1. In [npm](https://docs.npmjs.com/about-semantic-versioning) and Coursier, I think `x` means version range up to int max. So `2.13.x` would mean the latest stable available on the resolvers. This would typically mean 2.13.1 as of today.
2. Second, why are we dropping the dots? It's not that much of an effort to type `.` in `2.13.x`, and I think it would be confusing to say `213x` parses to `2.13.x` for many casual Scala users. I am assuming that the target audience for scala-runners includes totally new Scala users. Does `21210` work?

dwijnand:
> 1\. In [npm](https://docs.npmjs.com/about-semantic-versioning) and Coursier, I think `x` means version range up to int max. So `2.13.x` would mean the latest stable available on the resolvers. This would typically mean 2.13.1 as of today.

Where in Coursier does it mean that?  It doesn't look like it to me:

```
$ cs launch scala:2.13.x
Resolution error: Error downloading org.scala-lang:scala-compiler:2.13.x
  not found: /Users/dnw/.ivy2/local/org.scala-lang/scala-compiler/2.13.x/ivys/ivy.xml
  not found: https://repo1.maven.org/maven2/org/scala-lang/scala-compiler/2.13.x/scala-compiler-2.13.x.pom
```

It seems Coursier has latest.stable, latest.release and latest.integration (which need docs: https://github.com/coursier/coursier/issues/1249).  The git branch is 2.13.x, so I don't think it's crazy for "2.13.x" to mean "latest published version from the 2.13.x" (latest mergely) or "latest promoted version from the 2.13.x nightly.properties".  Just need to decide which one and how to refer to the other one.

> 2\. Second, why are we dropping the dots? It's not that much of an effort to type `.` in `2.13.x`, and I think it would be confusing to say `213x` that `2.13.x` for many casual Scala users. I am assuming that the target audience for scala-runners includes totally new Scala users. Does `21210` work?

The `-213`, etc., versions are following the precedent from sbt-extras.  I think scala-runners could generically handle `-<digit>.<digit>.<digit>`, like `-2.12.10`.

dwijnand:
Maybe 2.13.HEAD or 2.13.x-HEAD for looking for the mergily of the HEAD of that branch.

eed3si9n:
> Where in Coursier does it mean that? 

Coursier has a notion of "max", and apparently I added `.x` as an alias myself in https://github.com/coursier/coursier/pull/1348/commits/227aca7dca6953bfe3dd306f2c979c2058fa54c7? It may need to propagate to other parts for full support.

martijnhoekstra:
I pushed a version that does `2.13-latest` for last release, `2.13-nightly` for mergely and `2.13-unstable` for CB-checked mergely, as a baseline to show it's at least possible. I have a small preference for "nightly" for mergely since that such a common term and a small preference for keeping the dynamic version on `epoch.major-kind` for uniformity, but happy to break that for `epoch.major.x` if that's just nicer for latest release.

martijnhoekstra:
Another naming suggestion from @hrhino: 2.13.next. I really like it, though I don't know if that sounds more like CB or like mergely. Though with the the suggestion of HEAD that would fit a 2.13.x/2.13.next/2.13.HEAD scheme for release/CB/mergely quite well.

2.13.current is also a fun variation.

dwijnand:
I really like `2.13.next` for the current CB version.  To me, the current release is the simpler `2.13`.  I still like 2.13.x as the latest mergely because it matches the branch name.  But maybe I could get used to 2.13.head.

martijnhoekstra:
If you think in terms of branches, than 2.13.x for mergely makes sense, but I don't think most people who want to have a smoke test on the latest nightly know about or think in terms of branch names. Maybe that's an argument that it's not intuitive enough what 2.13.x means and should be avoided altogether?

I'm not a big fan of 2.13 for latest stable as it doesn't do enough for me to indicate it's a dynamic latest version.

Apart from the earlier proposed 2.13.latest I also want to throw 2.13.stable and 2.13.current in the hat. 2.13.current particularly works nicely in conjunction with 2.13.next

main/src/main/scala/sbt/Cross.scala
```diff
@@ -261,6 +312,34 @@ object Cross {
...
+    def resolveNightly(major: String): String = {
+      val url =
+        s"https://api.travis-ci.org/repo/scala%2Fscala/builds?limit=1&state=passed&include=build.commit&branch.name=$major.x"
+      val connection = new java.net.URL(url).openConnection()
+      connection.addRequestProperty("Travis-API-Version", "3")
+      val source = scala.io.Source.fromInputStream(connection.getInputStream())(scala.io.Codec.UTF8)
+      val content = source.getLines().mkString
+      val pattern = """"sha":\s+"([a-z0-9]{40})"""" r
+      val u = pattern.unanchored
+      val minor = resolveNightlyCB(major).takeWhile(_ != '-').split('.')(2)
```
martijnhoekstra: This is pretty gnarly, even compared to the rest of it.

Getting releases from the github API may be better.

martijnhoekstra:
Current implementation: 2.13.current, 2.13.next, 2.13.unstable

eed3si9n:
Should we try to iterate on this feature in plugin space like sbt-doge?

martijnhoekstra:
@eed3si9n possibly!

I have no idea how to even get started with that.

martijnhoekstra:
I went over the sbt-doge source again, but I'm way over my head, I don't understand anything it's doing.

eed3si9n:
The basic strategy is to copy-paste the entire + and ++ command implementation into your own plugin, and change whatever behavior you want in there.

martijnhoekstra:
I'll see if I can find that.

eed3si9n:
Closing this for now.

PR #5469, opened 2020-03-13
Fix a message about an undefined setting

nigredo-tori:
`defining` seems to be a setting or a task that references `referencedKey`, not the other way around.

### DIFF
--- a/internal/util-collection/src/main/scala/sbt/internal/util/Settings.scala
+++ b/internal/util-collection/src/main/scala/sbt/internal/util/Settings.scala
@@ -347,7 +347,7 @@ trait Init[ScopeType] {
       extends RuntimeException("References to undefined settings at runtime.") {
     override def getMessage =
       super.getMessage + undefined.map { u =>
-        "\n" + u.defining + " referenced from " + u.referencedKey
+        "\n" + u.referencedKey + " referenced from " + u.defining
       }.mkString
   }
 


eed3si9n:
I just confirmed this using the following.

## steps

```scala

ThisBuild / scalaVersion     := "2.13.1"
ThisBuild / version          := "0.1.0-SNAPSHOT"
ThisBuild / organization     := "com.example"
ThisBuild / organizationName := "example"

lazy val foo = settingKey[String]("")
lazy val bar = taskKey[String]("")

  .settings(
    name := "hello",
    bar := (Def.taskDyn {
      Def.task { foo.value }
    }).value
  )
```

```
sbt:hello> bar
```

## problem

```
sbt:hello> bar
[error] stack trace is suppressed; run last bar for the full output
[error] (bar) sbt.internal.util.Init$RuntimeUndefined: References to undefined settings at runtime.
[error] setting(ScopedKey(Scope(Select(ProjectRef(file:/private/tmp/hello/,root)), Zero, Zero, Zero),bar)) at LinePosition(/private/tmp/hello/build.sbt,14) referenced from ScopedKey(Scope(Select(ProjectRef(file:/private/tmp/hello/,root)), Zero, Zero, Zero),foo)
```

## expectation

```
sbt:hello> bar
[error] stack trace is suppressed; run last bar for the full output
[error] (bar) sbt.internal.util.Init$RuntimeUndefined: References to undefined settings at runtime.
[error] ScopedKey(Scope(Select(ProjectRef(file:/private/tmp/hello/,root)), Zero, Zero, Zero),foo) at LinePosition(/private/tmp/hello/build.sbt,14) referenced from setting(ScopedKey(Scope(Select(ProjectRef(file:/private/tmp/hello/,root)), Zero, Zero, Zero),bar))
```

eed3si9n (approved):
Thanks for the contribution!

PR #5474, opened 2020-03-18
Update scala to 2.12.11

anilkumarmyla:

PR #5481, opened 2020-03-20
Update semanticdb to work with scala 2.12.11

anilkumarmyla:
was planning to do this along with #5474, but the scripted failures threw me off. Hence this PR


eed3si9n (approved):
LGTM

PR #5494, opened 2020-04-03
Remove ! which forces version

ignasi35:
Commit 53788ba3561a4e79bc9c36c329f43540ad48814e (and related) started using `$version` with `!` when running a task for multiple versions.

See https://github.com/akka/akka-grpc/pull/907 for a reproducer.

Please note that in this PR I blindly removed all usages of `!` and I should probably have done that in a more selective way. This is why I'm keeping this `DRAFT`.

### DIFF
--- a/main/src/main/scala/sbt/Cross.scala
+++ b/main/src/main/scala/sbt/Cross.scala
@@ -159,7 +159,7 @@ object Cross {
         // Execute using a blanket switch
         projCrossVersions.toMap.apply(extracted.currentRef).flatMap { version =>
           // Force scala version
-          Seq(s"$SwitchCommand $verbose $version!", aggCommand)
+          Seq(s"$SwitchCommand $verbose $version", aggCommand)
         }
       case Right((keys, taskArgs)) =>
         def project(key: ScopedKey[_]): Option[ProjectRef] = key.scope.project.toOption match {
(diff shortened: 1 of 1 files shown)


eed3si9n:
@ignasi35 Could you open an issue for this, so we can add a test? 

I am assuming that there's an implicit bug here that cross command (`+ publishLocal` etc) could leave the multi project in some unintended state?

eed3si9n:
Ref https://travis-ci.org/github/sbt/sbt/jobs/670695132

```
[info] Running actions/cross-test-only
[info] [info] Loading project definition from /tmp/sbt_fe7e3a28/project
[info] [info] Compiling 1 Scala source to /tmp/sbt_fe7e3a28/project/target/scala-2.12/sbt-1.0/classes ...
[info] [info] Done compiling.
[info] [info] Loading settings for project root from build.sbt ...
[info] [info] Set current project to root (in build file:/tmp/sbt_fe7e3a28/)
[info] [info] Set current project to root (in build file:/tmp/sbt_fe7e3a28/)
[info] [info] Setting Scala version to 2.12.11 on 2 projects.
[info] [info] Excluded 1 projects, run ++ 2.12.11 -v for more details.
[info] [info] Reapplying settings...
[info] [info] Set current project to root (in build file:/tmp/sbt_fe7e3a28/)
[info] [error] Expected ';'
[info] [error] all foo/all foo/clean client/clean client/all foo/clean client/clean
[info] [error]                                                                     ^
[error] x actions/cross-test-only 
[error]  Cause of test exception: {line 1}  Command failed: + failed
```

`foo/all` seems wrong.

ignasi35:
Hi @eed3si9n I rushed my way into the code change without even running tests locally or discussing the symptoms/expectations. As you suggested, it's better we mode this to an issue first. 

I'll close this PR. :-)

ignasi35:
Raised https://github.com/sbt/sbt/issues/5497

main/src/main/scala/sbt/Cross.scala
```diff
@@ -323,7 +323,7 @@ object Cross {
         if (included.isEmpty) {
           sys.error(
             s"""Switch failed: no subprojects list "$version" (or compatible version) in crossScalaVersions setting.
-               |If you want to force it regardless, call ++ $version!""".stripMargin
+               |If you want to force it regardless, call ++ $version""".stripMargin
```
ignasi35, line 326: This slug should be undone. This occurrence of `!` is on a log statment not an actual usage.

Issue #5497, opened 2020-04-06
Forced version in some cross operations


ignasi35:
(I think this is a regression introduced in some of the changes in `Cross` in version `1.3.5`and forward)

## steps

Given a project with 3 modules:

 - `a` has `crossScalaVersions := Seq("2.12.10")` and  `scalaVersion := "2.12.10"`
 - `b` has `crossScalaVersions := Seq("2.12.10", "2.13.1")` and  `scalaVersion := "2.12.10"`
 - `c` has `crossScalaVersions := Seq("2.12.10")` and  `scalaVersion := "2.12.10"`. `c`also depends on `a`

And a `root` which aggregates all the modules above and sets `crossScalaVersions := Nil` and  `scalaVersion := "2.12.10"`.

When running `+publishLocal` both `a` and `c` are published for `2.12.10` *and* `2.13.1`.

See https://github.com/akka/akka-grpc/pull/907 for a reproducer. That PR executes `+publishLocal` as expected when run on sbt `1.3.4` but fails when run on any version after `1.3.5`.

## problem

Starting sbt `1.3.5` using `+` will often `Force` the scala version instead of `Setting` it. The implication is that some projects that don't define a particular scala version 
in the `crossScalaVersions` are added to the task anyway.

## expectation

Given the project above, I would expect the following artifacts only:

 - a_2.12.jar
 - b_2.12.jar
 - b_2.13.jar
 - c_2.12.jar

## notes

1. The reproducer I linked above seems to have/miss something since I'm not sure I can reproduce this problem in the Lagom `1.6.x` branch build. The Lagom `1.6.x` branch  build has an even more complex cross-version setup since it support not only `2.12` and `2.13` but also sbt `0.13` and `1.0` (which requires some artifacts to be built for scala `2.10`.
2. in the reproducer I linked the project `c` I refer to in my examples above is actually an `SbtPlugin` (no surprise there). The particularity of the `SbtPlugin` in `akka-grpc` is that [it depends on another, external `SbtPlugin`](https://github.com/akka/akka-grpc/blob/master/project/Dependencies.scala#L100-L103).

raboof:
I took some time to narrow this down, and have a 'reproducer' at https://github.com/raboof/sbt-reproduce-5497/blob/master/build.sbt#L6 - `sbt +compile` to reproduce.

Turns out this use of `ProjectRef` here was essential in reproducing the behavior.

[...]

PR #5511, opened 2020-04-23
Allow using a projectref across scala versions with cross building

raboof:
Reproduces #5497 and contains the fix by @ignasi35 from #5494 - I'm not seeing
actions/cross-test-only fail here

### DIFF
--- a/main/src/main/scala/sbt/Cross.scala
+++ b/main/src/main/scala/sbt/Cross.scala
@@ -158,8 +158,8 @@ object Cross {
 
         // Execute using a blanket switch
         projCrossVersions.toMap.apply(extracted.currentRef).flatMap { version =>
-          // Force scala version
-          Seq(s"$SwitchCommand $verbose $version!", aggCommand)
+          // Set scala version
+          Seq(s"$SwitchCommand $verbose $version", aggCommand)
         }
       case Right((keys, taskArgs)) =>
         def project(key: ScopedKey[_]): Option[ProjectRef] = key.scope.project.toOption match {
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/Cross.scala
```diff
@@ -195,11 +195,11 @@ object Cross {
         commandsByVersion.flatMap {
           case (v, commands) =>
             commands match {
-              case Seq(c) => Seq(s"$SwitchCommand $verbose $v! $c")
+              case Seq(c) => Seq(s"$SwitchCommand $verbose $v $c")
               case Seq()  => Nil // should be unreachable
               case multi if fullArgs.isEmpty =>
                 Seq(s"$SwitchCommand $verbose $v! all ${multi.mkString(" ")}")
-              case multi => Seq(s"$SwitchCommand $verbose $v!") ++ multi
+              case multi => Seq(s"$SwitchCommand $verbose $v") ++ multi
```
ignasi35, line 202: Hmm, in https://github.com/sbt/sbt/pull/5494 I didn't really know what I was doing and I aggressively removed the use of `!`. Is it necessary to remove the use of `!` in all three code occurrences to fix the issue?

After I sent https://github.com/sbt/sbt/pull/5494 I started thinking I might introduce other regressions.

PS: I noticed you didn't edit line `201` which makes me assume you _indeed_ change the minimum occurrences required 😂

raboof: The reproducer doesn't need all 3 changes, but it would be quite inconsistent to not change all of these occurrences, as they seems quite similar. I'm definitely not familiar enough with this code to predict all the possible effects of this change, but at least the test suite doesn't appear to signal any regressions.

main/src/main/scala/sbt/Cross.scala
```diff
@@ -158,8 +158,8 @@ object Cross {
 
         // Execute using a blanket switch
         projCrossVersions.toMap.apply(extracted.currentRef).flatMap { version =>
-          // Force scala version
-          Seq(s"$SwitchCommand $verbose $version!", aggCommand)
+          // Set scala version
+          Seq(s"$SwitchCommand $verbose $version", aggCommand)
```
eed3si9n, line 162: I think this ! is necessary for sbt 0.13 emulation - https://github.com/sbt/sbt/issues/5497#issuecomment-618740983

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/5512

PR #5512, opened 2020-04-24
cross building tests and fixes

### LINKED ISSUE #5497: Forced version in some cross operations

### LINKED ISSUE #5506: Verbose switch for `+` is broken
## steps

No project is needed at all to reproduce:

```
$ mkdir /tmp/demo && cd /tmp/demo
$ sbt --warn --sbt-version 1.3.10
sbt:demo> + -v compile
[error] Expected ';'
[error] ++ -v 2.12.10! demo/Compile/compile
[error]                                    ^
```

This looks like a regression from #3995.

## problem

`+ -v` does not work at all.

## expectation

Commands should work with verbose Scala version switching, as `help +` describes.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/5497
Fixes #5506
Ref https://github.com/sbt/sbt/pull/5511

`+` seems to have regressed over time (see https://github.com/sbt/sbt/issues/5497#issuecomment-618740983). This attempts to add tests for various `+` use cases.


raboof:
Great! Seems to be working both for my reproducer and for the original project (akka-grpc) :tada:

ignasi35:
Nice work @eed3si9n! Thanks for fixing this.

PR #5513, opened 2020-04-24
add context-bound-relaxed helpers for write-only Tracked 

bjaglin:


util-tracking/src/test/scala/sbt/util/TrackedSpec.scala
```diff
@@ -113,6 +135,27 @@ class TrackedSpec extends FlatSpec {
...
+    case class Input(v: Int)
+
+    implicit val writer = new JsonWriter[Input] {
+      override def write[J](obj: Input, builder: Builder[J]): Unit = builder.writeInt(obj.v)
+    }
+
+    withStore { store =>
+      val input0 = Input(1)
+
+      val cachedFun = Tracked.outputChangedWriter[Input, Int](store) {
```
bjaglin: I am not very satisfied with the `Writer` suffix but I didn't manage to make a true overload as the signature was ambiguous with the existing one. I did manage to create an overload by adding a dummy type parameter, but since the type parameters are usually explicitly passed (to avoid typing the arguments of the closure), it forced the caller to add that third, useless type, so I figured out it was better to give another name.

README.md
```diff
@@ -11,7 +11,6 @@
   [asking]: https://stackoverflow.com/questions/ask?tags=sbt
   [LICENSE]: LICENSE
   [sbt/io]: https://github.com/sbt/io
-  [sbt/util]: https://github.com/sbt/util
```
bjaglin, line 14: I started hacking over there after seeing this, only to realize that it had been inlined, so I guess it's better to remove it from here? Or maybe you are planning to cherry-pick stuff anyway?

util-tracking/src/test/scala/sbt/util/TrackedSpec.scala
bjaglin: (and the naive idea of simply changing the type caused binary incompatibility, although I assume it's source-compatible)
```
[error]  * method outputChanged(java.io.File,scala.Function2,sjsonnew.JsonFormat)scala.Function1 in object sbt.util.Tracked in current version does not have a correspondent with same parameter signature among (java.io.File,scala.Function2,sjsonnew.JsonWriter)scala.Function1, (sbt.util.CacheStore,scala.Function2,sjsonnew.JsonWriter)scala.Function1
```

README.md
eed3si9n: This was an oversight. Thanks for the change.

util-tracking/src/main/scala/sbt/util/Tracked.scala
```diff
@@ -95,6 +95,12 @@ object Tracked {
    */
   def outputChanged[A1: JsonFormat, A2](store: CacheStore)(
       f: (Boolean, A1) => A2
+  ): (() => A1) => A2 = {
+    outputChangedWriter(store)(f)
+  }
+
+  def outputChangedWriter[A1: JsonWriter, A2](store: CacheStore)(
```
eed3si9n:
```suggestion
  /**
   * Creates a tracker that indicates whether the output returned from `f` has changed or not.
   *
   * {{{
   * val cachedTask = inputChanged(cache / "inputs") { (inChanged, in: Inputs) =>
   *   Tracked.outputChangedW(cache / "output") { (outChanged, outputs: FilesInfo[PlainFileInfo]) =>
   *     if (inChanged || outChanged) {
   *       doSomething(label, sources, classpath, outputDirectory, options, log)
   *     }
   *   }
   * }
   * cachedDoc(inputs)(() => exists(outputDirectory.allPaths.get.toSet))
   * }}}
   *
   * This is a variant of `outputChanged` that takes `A1: JsonWriter` as opposed to `A1: JsonFormat`.
   */
  def outputChangedW[A1: JsonWriter, A2](store: CacheStore)(
```

util-tracking/src/test/scala/sbt/util/TrackedSpec.scala
eed3si9n: Assuming most plugin authors can continue to use the old variant, what do you think about using `W` as suffix - see https://github.com/sbt/sbt/pull/5513/files#r414623727 ?

util-tracking/src/main/scala/sbt/util/Tracked.scala
```diff
@@ -127,6 +133,11 @@ object Tracked {
   def outputChanged[A1: JsonFormat, A2](cacheFile: File)(f: (Boolean, A1) => A2): (() => A1) => A2 =
     outputChanged[A1, A2](CacheStore(cacheFile))(f)
 
+  def outputChangedWriter[A1: JsonWriter, A2](
```
eed3si9n:
```suggestion
  /**
   * Creates a tracker that indicates whether the output returned from `f` has changed or not.
   *
   * {{{
   * val cachedTask = inputChanged(cache / "inputs") { (inChanged, in: Inputs) =>
   *   Tracked.outputChangedW(cache / "output") { (outChanged, outputs: FilesInfo[PlainFileInfo]) =>
   *     if (inChanged || outChanged) {
   *       doSomething(label, sources, classpath, outputDirectory, options, log)
   *     }
   *   }
   * }
   * cachedDoc(inputs)(() => exists(outputDirectory.allPaths.get.toSet))
   * }}}
   *
   * This is a variant of `outputChanged` that takes `A1: JsonWriter` as opposed to `A1: JsonFormat`.
   */
  def outputChangedW[A1: JsonWriter, A2]( 
```

eed3si9n:
@bjaglin Thanks for the contribution! Overall I think the idea is good. Since this is an area of sbt that's confusing, I think it would be good to put Scaladocs to document our intent while it's fresh in our minds.

bjaglin:
> I think it would be good to put Scaladocs to document our intent while it's fresh in our minds.

Indeed! I amended my second commit 6be1090 to use `W` as a suffix and add scaladoc everywhere.

I realized the scaladoc was the same for all overloads (`CacheStore` or `File`), so I took the opportunity to update the examples in 33b09dc.

eed3si9n (approved):
LGTM

Issue #5508, opened 2020-04-20
`update` gets slower with the number of modules in project


jtjeferreira:
## steps

Checkout https://github.com/jtjeferreira/sbt-multi-module-sample which is a sample of a multi module project with a dependency to `spark` (spark was chosen because it has a lot of transitive dependencies).

Run `sbt update`. After everything is downloaded for the first time, run `sbt update` again so no downloads are made again. In my dell xps 13 laptop with encrypted disk this takes 35s. 

If the project just contains a single module, `update` just takes 5s (see https://github.com/jtjeferreira/sbt-multi-module-sample/commit/4f96638ec5802dac22368c9905975578ba481f27)

## problem

It seems that sbt/coursier is validating the checksum of the artifacts once per module, which is unnecessary...

## expectation

I would expect that `sbt update` duration should just increase with the number of artifacts it needs to check, not with the number of modules... 

## notes

I have run Java Flight Recorder and can report the following:
* A lot of `coursier-pool-n-thread-m` are created. Where n is between [2, 162], and m is between [1, 5] and in some cases [1,6]. That means around 1k threads were created
* `coursier.cache.FileCache#validateChecksum` seems to be the hotpath

jtjeferreira:
update: if ivy and cachedResolution is used instead, it takes around 19s (vs 35s with coursier...)

eed3si9n:
@jtjeferreira Thanks for the report. Given that you've attributed this behavior to Cousier, could you report this to coursier/coursier as well please?

jtjeferreira:
Hi @eed3si9n, sure I can open, but I am not sure the problem is in sbt or coursier. An possible solution would be that sbt calls coursier differently; or that sbt implements cachedResolution for coursier and calls coursier less often?

Or do you have the feeling this is definitely something to be fixed on coursier side?

eed3si9n:
I could be wrong, but I am guessing it's on Coursier side since I am not aware of sbt calling `validateChecksum`.

jtjeferreira:
> I could be wrong, but I am guessing it's on Coursier side since I am not aware of sbt calling `validateChecksum`.

Yes, thats true, but let me elaborate better my assumptions and also what I investigated so far:
* `coursier.cache.FileCache#validateChecksum` is the hotpath, but this is because 
sbt is calling coursier (`coursier.Artifacts.either()`) for every module, and then `coursier.cache.FileCache#file` and thats where `validateChecksum` will be  called...
[...]

PR #5515, opened 2020-04-26
introduce a new Input/Output: FileInput/FileOutput 

jtjeferreira:
* FileInput avoids the intermediate String parsing of PlainInput
* FileOuput also avoid the intermediate String representation of PlainOutput

### DIFF
--- a/main/src/main/scala/sbt/internal/LibraryManagement.scala
+++ b/main/src/main/scala/sbt/internal/LibraryManagement.scala
@@ -307,17 +307,11 @@ private[sbt] object LibraryManagement {
   def withExcludes(out: File, classifiers: Seq[String], lock: xsbti.GlobalLock)(
       f: Map[ModuleID, Vector[ConfigRef]] => UpdateReport
   ): UpdateReport = {
-    import sjsonnew.shaded.scalajson.ast.unsafe.JValue
     import sbt.librarymanagement.LibraryManagementCodec._
     import sbt.util.FileBasedStore
-    implicit val isoString: sjsonnew.IsoString[JValue] =
-      sjsonnew.IsoString.iso(
-        sjsonnew.support.scalajson.unsafe.CompactPrinter.apply,
-        sjsonnew.support.scalajson.unsafe.Parser.parseUnsafe
-      )
     val exclName = "exclude_classifiers"
     val file = out / exclName
-    val store = new FileBasedStore(file, sjsonnew.support.scalajson.unsafe.Converter)
+    val store = new FileBasedStore(file)
     lock(
       out / (exclName + ".lock"),
       new Callable[UpdateReport] {
--- a/util-cache/src/main/scala/sbt/util/CacheStore.scala
+++ b/util-cache/src/main/scala/sbt/util/CacheStore.scala
@@ -8,11 +8,12 @@
 package sbt.util
 
 import java.io.{ File, InputStream, OutputStream }
+
 import sbt.io.syntax.fileToRichFile
-import sbt.io.{ IO, Using }
-import sjsonnew.{ IsoString, JsonReader, JsonWriter, SupportConverter }
-import sjsonnew.support.scalajson.unsafe.{ CompactPrinter, Converter, Parser }
+import sbt.io.IO
 import sjsonnew.shaded.scalajson.ast.unsafe.JValue
+import sjsonnew.support.scalajson.unsafe.{ CompactPrinter, Parser }
+import sjsonnew.{ IsoString, JsonReader, JsonWriter, SupportConverter }
 
 /** A `CacheStore` is used by the caching infrastructure to persist cached information. */
 abstract class CacheStore extends Input with Output {
@@ -23,6 +24,7 @@ abstract class CacheStore extends Input with Output {
 }
 
 object CacheStore {
+  @deprecated("Create your own IsoString[JValue]", "1.4")
   implicit lazy val jvalueIsoString: IsoString[JValue] =
     IsoString.iso(CompactPrinter.apply, Parser.parseUnsafe)
 
@@ -30,7 +32,7 @@ object CacheStore {
   def apply(cacheFile: File): CacheStore = file(cacheFile)
 
   /** Returns file-based CacheStore using standard JSON converter. */
-  def file(cacheFile: File): CacheStore = new FileBasedStore[JValue](cacheFile, Converter)
+  def file(cacheFile: File): CacheStore = new FileBasedStore(cacheFile)
 }
 
 /** Factory that can make new stores. */
@@ -47,6 +49,7 @@ abstract class CacheStoreFactory {
 }
 
 object CacheStoreFactory {
+  @deprecated("Create your own IsoString[JValue]", "1.4")
   implicit lazy val jvalueIsoString: IsoString[JValue] =
     IsoString.iso(CompactPrinter.apply, Parser.parseUnsafe)
 
@@ -54,31 +57,36 @@ object CacheStoreFactory {
   def apply(base: File): CacheStoreFactory = directory(base)
 
   /** Returns directory-based CacheStoreFactory using standard JSON converter. */
-  def directory(base: File): CacheStoreFactory = new DirectoryStoreFactory[JValue](base, Converter)
+  def directory(base: File): CacheStoreFactory = new DirectoryStoreFactory(base)
 }
 
 /** A factory that creates new stores persisted in `base`. */
-class DirectoryStoreFactory[J: IsoString](base: File, converter: SupportConverter[J])
-    extends CacheStoreFactory {
+class DirectoryStoreFactory[J](base: File) extends CacheStoreFactory {
   IO.createDirectory(base)
 
-  def make(identifier: String): CacheStore = new FileBasedStore(base / identifier, converter)
+  @deprecated("Use constructor without converter", "1.4")
+  def this(base: File, converter: sjsonnew.SupportConverter[J])(implicit e: sjsonnew.IsoString[J]) =
+    this(base)
+
+  def make(identifier: String): CacheStore = new FileBasedStore(base / identifier)
 
   def sub(identifier: String): CacheStoreFactory =
-    new DirectoryStoreFactory(base / identifier, converter)
+    new DirectoryStoreFactory(base / identifier)
 }
 
 /** A `CacheStore` that persists information in `file`. */
-class FileBasedStore[J: IsoString](file: File, converter: SupportConverter[J]) extends CacheStore {
+class FileBasedStore[J](file: File) extends CacheStore {
   IO.touch(file, setModified = false)
 
+  @deprecated("Use constructor without converter", "1.4")
+  def this(file: File, converter: sjsonnew.SupportConverter[J])(implicit e: sjsonnew.IsoString[J]) =
+    this(file)
+
   def read[T: JsonReader]() =
-    Using.fileInputStream(file)(stream => new PlainInput(stream, converter).read())
+    new FileInput(file).read()
 
   def write[T: JsonWriter](value: T) =
-    Using.fileOutputStream(append = false)(file) { stream =>
-      new PlainOutput(stream, converter).write(value)
-    }
+    new FileOutput(file).write(value)
 
   def delete() = IO.delete(file)
   def close() = ()
--- a/util-cache/src/main/scala/sbt/util/Input.scala
+++ b/util-cache/src/main/scala/sbt/util/Input.scala
@@ -7,7 +7,8 @@
 
 package sbt.util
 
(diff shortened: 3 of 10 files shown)


util-cache/src/main/scala/sbt/util/Input.scala
```diff
@@ -44,3 +45,13 @@ class PlainInput[J: IsoString](input: InputStream, converter: SupportConverter[J
 
   def close() = input.close()
 }
+
+class FileInput(file: File) extends Input {
```
jtjeferreira, line 49: this in the new Input that uses `Parser.parseFromFile` directly , instead of `IsoString` to convert to a interdemiary `String` first. This is where this whole PR started...

util-cache/src/main/scala/sbt/util/Output.scala
```diff
@@ -31,3 +32,17 @@ class PlainOutput[J: IsoString](output: OutputStream, converter: SupportConverte
 
   def close() = output.close()
 }
+
+class FileOutput(file: File) extends Output {
```
jtjeferreira, line 36: this is new the new Input with a signature similar to `FileInput`, but it still does the intermediary conversion to `String`. However this can still be improved in the future (for example adding it to sjsonnew or if jawn adds serializer capabilities https://github.com/typelevel/jawn/issues/6)

util-cache/src/main/scala/sbt/util/CacheStore.scala
```diff
@@ -23,14 +24,15 @@ abstract class CacheStore extends Input with Output {
 }
 
 object CacheStore {
+  @deprecated("Create your own IsoString[JValue]", "1.4")
   implicit lazy val jvalueIsoString: IsoString[JValue] =
```
jtjeferreira, line 28: this val is kept just for bincompt...

util-cache/src/main/scala/sbt/util/CacheStore.scala
```diff
@@ -47,38 +49,44 @@ abstract class CacheStoreFactory {
 }
 
 object CacheStoreFactory {
+  @deprecated("Create your own IsoString[JValue]", "1.4")
   implicit lazy val jvalueIsoString: IsoString[JValue] =
```
jtjeferreira, line 53: this val is kept just for bincompt...

util-cache/src/main/scala/sbt/util/CacheStore.scala
```diff
@@ -47,38 +49,44 @@ abstract class CacheStoreFactory {
...
 
 /** A factory that creates new stores persisted in `base`. */
-class DirectoryStoreFactory[J: IsoString](base: File, converter: SupportConverter[J])
-    extends CacheStoreFactory {
+class DirectoryStoreFactory[J](base: File) extends CacheStoreFactory {
   IO.createDirectory(base)
 
-  def make(identifier: String): CacheStore = new FileBasedStore(base / identifier, converter)
+  @deprecated("Use constructor without converter", "1.4")
+  def this(base: File, converter: sjsonnew.SupportConverter[J], isoString: sjsonnew.IsoString[J]) =
```
jtjeferreira: this constructor was added, just for bincompt, but some parameters will just be ignored...

util-cache/src/main/scala/sbt/util/CacheStore.scala
```diff
@@ -47,38 +49,44 @@ abstract class CacheStoreFactory {
...
+    new DirectoryStoreFactory(base / identifier)
 }
 
 /** A `CacheStore` that persists information in `file`. */
-class FileBasedStore[J: IsoString](file: File, converter: SupportConverter[J]) extends CacheStore {
+class FileBasedStore[J](file: File) extends CacheStore {
   IO.touch(file, setModified = false)
 
+  @deprecated("Use constructor without converter", "1.4")
+  def this(file: File, converter: sjsonnew.SupportConverter[J])(implicit e: sjsonnew.IsoString[J]) =
```
jtjeferreira, line 82: this constructor was added, just for bincompt, but some parameters will just be ignored...

jtjeferreira:
tests are failing on windows... any help with those?

eed3si9n:
https://github.com/sbt/sbt/blob/develop/sbt/src/sbt-test/actions/previous/build.sbt

```
[info] Running actions/previous
[info] [error] java.lang.AssertionError: assertion failed: Expected 'a' to be 1, got 8
[info] [error] 	at scala.Predef$.assert(Predef.scala:223)
[info] [error] 	at $5e6b47047a4c8d8f03cb$.$anonfun$$sbtdef$3(C:\Users\appveyor\AppData\Local\Temp\1\sbt_19336570\build.sbt:30)
[info] [error] 	at $5e6b47047a4c8d8f03cb$.$anonfun$$sbtdef$3$adapted(C:\Users\appveyor\AppData\Local\Temp\1\sbt_19336570\build.sbt:26)
[info] [error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
[info] [error] 	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:62)
[info] [error] 	at sbt.std.Transform$$anon$4.work(Transform.scala:67)
[info] [error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:282)
[info] [error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:23)
[info] [error] 	at sbt.Execute.work(Execute.scala:291)
[info] [error] 	at sbt.Execute.$anonfun$submit$1(Execute.scala:282)
[info] [error] 	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:184)
[info] [error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:41)
[info] [error] 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[info] [error] 	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[info] [error] 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[info] [error] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[info] [error] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[info] [error] 	at java.lang.Thread.run(Thread.java:748)
[info] [error] (checkNext) java.lang.AssertionError: assertion failed: Expected 'a' to be 1, got 8
[error] x actions/previous 
[error]  Cause of test exception: {line 6}  Command failed: checkNext failed
```

so for some reason `clean` didn't have the effect of clearing out the previous value?

util-cache/src/main/scala/sbt/util/Output.scala
jtjeferreira: I created a PR for sjson-new that would allow us to write directly to a file. https://github.com/eed3si9n/sjson-new/pull/112

jtjeferreira:
fixed the windows error which was caused by https://github.com/typelevel/jawn/issues/109, however now the build fails in 2.13 because in 2.13 we use jawn 1.0...

jtjeferreira:
finally tests are green

util-cache/src/main/scala-2.13+/sbt/util/FileInputCompat.scala
```diff
@@ -0,0 +1,18 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.util
```
eed3si9n: Could you move this to sbt.internal.util please?

main/src/main/scala/sbt/internal/LibraryManagement.scala
```diff
@@ -307,17 +307,11 @@ private[sbt] object LibraryManagement {
...
     import sbt.util.FileBasedStore
-    implicit val isoString: sjsonnew.IsoString[JValue] =
-      sjsonnew.IsoString.iso(
-        sjsonnew.support.scalajson.unsafe.CompactPrinter.apply,
-        sjsonnew.support.scalajson.unsafe.Parser.parseUnsafe
-      )
     val exclName = "exclude_classifiers"
     val file = out / exclName
-    val store = new FileBasedStore(file, sjsonnew.support.scalajson.unsafe.Converter)
+    val store = new FileBasedStore(file)
```
eed3si9n, line 314: Do you see difference in load time from this change?

jtjeferreira: I havent benchmarked this change against my sample project in #5508...

Do you think I should add that sample project to a scripted test so is easier to check? or is it too much for a unit test?

eed3si9n: I think it's ok for the tests to prove just the correctness, but if we're making some optimization or changes aimed at performance improvement, I would like to see some data even if it's casually done local testing.

util-cache/src/main/scala/sbt/util/Output.scala
jtjeferreira: I have rebased and we now use the methods introduced in `eed3si9n/sjson-new#112`

main/src/main/scala/sbt/internal/LibraryManagement.scala
jtjeferreira: Hi @eed3si9n! I rebased this PR and compared the performance of this PR versus 1.4.0-M1 and did not see any significant improvements. I can try to write some jmh benchmarks, but I see there aren't benchmarks in the project...

WDYT?

eed3si9n: I have a mixed feeling about jmh. In general they are useful. But since the Scala compiler is complicated, even with 10 warmup or whatever C2 would keep optimizing for like 10 minutes, even longer.. so that means that more iteration you run, the performance gets better at least for compilation. See https://github.com/sbt/zinc/pull/767#issuecomment-621935005. This might affect sbt indirectly since there are some aspect of compilation. So to minimize the fluctuation, we should reuse `*.class` files between the iteration?

Existing code that might be relevant for benchmarking is scripted test and server test.
- https://github.com/sbt/sbt/blob/develop/scripted-sbt-redux/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
- https://github.com/sbt/sbt/blob/develop/scripted-sbt-redux/src/main/scala/sbt/scriptedtest/RemoteSbtCreator.scala
- https://github.com/sbt/sbt/blob/develop/server-test/src/test/scala/testpkg/TestServer.scala#L173-L174

See also https://github.com/eatkins/scala-build-watch-performance

jtjeferreira:
Hi @eed3si9n

I rebased this PR against 1.4.4 and compare these changes against it with https://github.com/jtjeferreira/sbt-multi-module-sample

I see some improvements in FileInput vs PlainInput (872 vs 1052 samples in the flamegraph), but no improvements in FileOutput vs PlainOutput (both around 240 samples in the flamegraph). 

WDYT? should we merge? should we close? should we benchmark differently?

jtjeferreira:
the flamegraphs can be found in this commit https://github.com/jtjeferreira/sbt-multi-module-sample/commit/c6a288f46978f220bc102e12e7b0c0e5b21a3f03

eed3si9n:
> I see some improvements in FileInput vs PlainInput (872 vs 1052 samples in the flamegraph), but no improvements in FileOutput vs PlainOutput (both around 240 samples in the flamegraph).

Thanks for the candid assessment. As a general idea streaming bytes directly into files sounds like a good idea. What do you think about merging this to develop, targeting for sbt 1.5.0? (Planned to be out early 2021, as opposed to the normal year-long wait)

If that's ok, could you squash the commits plz?

util-cache/src/main/scala/sbt/util/CacheStore.scala
eed3si9n: Is there a reason to deprecate this?

jtjeferreira: so it can be removed in the future

jtjeferreira:
> > I see some improvements in FileInput vs PlainInput (872 vs 1052 samples in the flamegraph), but no improvements in FileOutput vs PlainOutput (both around 240 samples in the flamegraph).
> 
> Thanks for the candid assessment. As a general idea streaming bytes directly into files sounds like a good idea. What do you think about merging this to develop, targeting for sbt 1.5.0? (Planned to be out early 2021, as opposed to the normal year-long wait)
> 
> If that's ok, could you squash the commits plz?

looks good to me. the commits were squashed.

eed3si9n (approved):

PR #5417, opened 2020-02-03
integrate with VirtualFile changes

eed3si9n:
Ref https://github.com/sbt/zinc/pull/712

PR #5521, opened 2020-05-01
Restore zinc source file invalidation performance

eatkins:
This restores the performance of the zinc source file invalidation relative to sbt 1.3.x. There was a performance regression that I noted in a comment in #5417. These changes also fix the nio/external-hooks test.


main/src/main/scala/sbt/Keys.scala
```diff
@@ -221,7 +221,7 @@ object Keys {
   val allowMachinePath = settingKey[Boolean]("Allow machine-specific paths during conversion.")
   val rootPaths = settingKey[Seq[NioPath]]("The root paths used to abstract machine-specific paths.")
```
eatkins: Would it make sense to move this to `sbt.nio.Keys`? Also, is `Seq[NioPath]` the right abstraction here? It seems a bit strange that these are order dependent. I assume that the index into this `Seq` corresponds to the `${N}/path/to/file` in VirtualFileRefs? It seems like it might be a bit more robust to have virtual files that look like `${base}/path/to/file` instead at the cost of a modest increase in storage.

eed3si9n: I split the VirtualFileRef suggestion into an issue - https://github.com/sbt/zinc/issues/772

As per sbt.Keys, if the key is meant to be used by build authors then I think we should mostly keep them in Keys so they can `import sbt.Keys._`.

eed3si9n:
@eatkins Thanks! I'll grab this locally and test the caching behavior later.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -202,7 +203,40 @@ object Defaults extends BuildCommon {
...
       },
       uncachedStamper := Stamps.uncachedStamps(fileConverter.value),
-      reusableStamper := Stamps.timeWrapLibraryStamps(uncachedStamper.value, fileConverter.value),
+      reusableStamper := {
+        val converter = fileConverter.value
+        val unmanagedCache = unmanagedFileStampCache.value
+        val managedCache = managedFileStampCache.value
+        val backing =
+          Stamps.timeWrapLibraryStamps(uncachedStamper.value, fileConverter.value)
+        new xsbti.compile.analysis.ReadStamps {
```
He-Pin, line 224: extract to a named dedicated class?

eatkins: What is the use case for using a dedicated class outside of this method? It uses `private[sbt]` classes so I can't see such a class being public and I'm not sure where else in sbt it would be used.

eatkins:
Hmm...I reran my tests and things actually seem slower after these changes so I think more investigation is needed. In general, it does seem that VirtualFiles have made incremental compilation slower, and not just due to external hooks, but I'm sure there are plenty of optimization opportunities based on the discussions happening over there.

eatkins:
This is ready now. I've been testing in conjunction with https://github.com/sbt/zinc/pull/845.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -214,8 +214,41 @@ object Defaults extends BuildCommon {
...
+          def getAllSourceStamps()
+              : java.util.Map[xsbti.VirtualFileRef, xsbti.compile.analysis.Stamp] =
+            new java.util.HashMap[xsbti.VirtualFileRef, xsbti.compile.analysis.Stamp]
+          def library(fr: xsbti.VirtualFileRef): xsbti.compile.analysis.Stamp = {
+            val path = converter.toPath(fr)
+            managedCache
+              .getOrElseUpdate(path, sbt.nio.FileStamper.Hash)
+              .map(_.stamp)
+              .getOrElse(backing.library(fr))
+          }
```
eed3si9n, line 240: Would this return farm hash or timestamp?

eatkins: It should be hash but because backing is time wrapped, it first computes the timestamp and only updates the hash if the timestamp has changed (assuming I understand correctly how `Stamps.timeWrapLibraryStamps` works).

eed3si9n (approved):

PR #5522, opened 2020-05-01
Fix DefinesClass implementation for jdk > 8

eatkins:
When trying to use any jdk > 8 with the latest sbt, sbt will die in some
projects because it tries to call Locate.defineClass on rt.jar, which
is represented with a DummyVirtualFile and causes a crash.


eed3si9n (approved):
Thanks!

Issue #5492, opened 2020-04-01
Can't use Microsoft Azure DevOps Maven Repository as Resolver


mantovani:
Ref https://github.com/sbt/sbt/issues/2366
Ref https://github.com/sbt/sbt/issues/5264

## steps

Add a Microsoft [Azure DevOps Maven](https://docs.microsoft.com/en-us/azure/devops/artifacts/get-started-maven?view=azure-devops) repository as resolver with Coursier throws:

```
[warn]         Note: Unresolved dependencies path:
[error] stack trace is suppressed; run last update for the full output
[error] (update) sbt.librarymanagement.ResolveException: Error downloading com.foo:bar_2.11:1.0.0
[error]   Not found
[error]   Not found
[error]   not found: /Users/mantovani/.ivy2/local/com.foo/bar_2.11/1.0.0/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/com/foo/bar_2.11/1.0.0/bar_2.11-1.0.0.pom
[error]   checksum format error: /Users/mantovani/Library/Caches/Coursier/v1/https/pkgs.dev.azure.com/Foo-backend/Foo/_packaging/Foo-releases/maven/v1/com/foo/bar_2.11/1.0.0/bar_2.11-1.0.0.pom.sha1
```

```sbt
ThisBuild / name := "baz"
ThisBuild / organization := "com.foo"
lazy val scala211 = "2.11.12"

crossScalaVersions := Seq(scala211)
ThisBuild / scalaVersion := scala211

libraryDependencies ++= Seq(
  "foo" %% "bar" % 0.1 % "provided"
)

ThisBuild / useCoursier := true

resolvers +=   "nabu-releases" at "https://pkgs.dev.azure.com/foo/Foo/_packaging/foo-releases/maven/v1"
credentials += Credentials(Path.userHome / ".sbt" / ".credentials")

```
sbt version
```
[info] This is sbt 1.3.9
```

```
sbt:foo> show credentials
[info] * FileCredentials("/Users/mantovani/.sbt/sonatype_credential")
[info] * FileCredentials("/Users/mantovani/.sbt/.credentials")
```
```
cat /Users/mantovani/.sbt/.credentials

realm=foo
host=pkgs.dev.azure.com
user=foo
password=xxx
```

The checksum actually has a HTML from Azure:

```
[...]

Issue #2366, opened 2016-01-14
Credential without a realm


eed3si9n:
Currently a realm is required to define a credential: http://www.scala-sbt.org/0.13/docs/Publishing.html#Credentials

When Sonatype OSS changed their realm on 2016-01-12 from "Sonatype Nexus Repository Manager" to "Nexus Repository Manager", sbt users suddenly became unable to publish their artifacts: https://issues.sonatype.org/browse/OSSRH-13401

Ivy's credential seem to accept `null` or `""` as realm value as blank value: https://github.com/sbt/ivy/blob/3cf31483b03e0ba294c76f5391dc7570b7649090/src/java/org/apache/ivy/util/Credentials.java#L55-L61, if this works, we should probably support it to provide a workaround.
Temporarily the build user might be able to just pass in `""` too. As of 2015-01-13 the issue was resolved by Sonatype by reverting back to the old realm name.

hertg:
I am currently experiencing this issue.

I've set up a Maven Repository on the Google Cloud Platform with the following project:
https://github.com/renaudcerrato/appengine-maven-repository

The project is using Basic Auth **without specifying any realm**, therefore returning `WWW-Authenticate: Basic` instead of `WWW-Authenticate: Basic realm="some-realm-name"`.

I am able to publish and checkout my libraries using gradle, **but not with sbt**.
Already tried setting the realm to empty string `""` and `null` without any success.

ie.
```sbt
credentials += Credentials(
  "",
  "my-repo-url",
  "my-user",
  "my-password"
)
resolvers += "mavenrepo" at "https://maven.company.com"
```
```sbt
credentials += Credentials(
  null,
  "my-repo-url",
  "my-user",
  "my-password"
)
resolvers += "mavenrepo" at "https://maven.company.com"
```

The equivalent configuration in gradle works perfectly fine.
```gradle
repositories {
maven {
    credentials {
            username 'my-user'
            password 'my-password'
        }
        url "https://maven.company.com"
    }
}
```

My sbt version is `1.2.8`.

arashi01:
Also been hit with this issue. Unable to use Azure DevOps included maven repositories for publishing since the realm returned by Azure is not stable (changes depending on server servicing the request).

@eed3si9n Would it be fair to reclassify this as a bug since underlying ivy seems to support this (and Maven and Gradle definitely do)?

eed3si9n:
@arashi01 Sure.

pangiole:
[...]

PR #5526, opened 2020-05-03
Fix coursierint NPE when credential realm is null

3rwww1:
As stated in #5492 and #2366 some artifact hosting services (at least
Azure Artifacts, seems to affect GCP as well) do not offer a stable HTTP
Auth realm that can be safely set in SBT credentials, hence the need to
support null or empty Credentials.

The Ivy (publishing) side of the issue was fixed in sbt/ivy#36

As to the resolving side, This commit is only part of the solution as it
just prevents an NPE and does not consider if coursier itself will fall
back to finding credentials with a null realm that matches the server
hostname.

Related-to: #5492
Related-to: #2366
Signed-off-by: Erwan Queffelec <erwan.queffelec@gmail.com>

### DIFF
--- a/main/src/main/scala/sbt/coursierint/CoursierInputsTasks.scala
+++ b/main/src/main/scala/sbt/coursierint/CoursierInputsTasks.scala
@@ -231,7 +231,7 @@ object CoursierInputsTasks {
           .withHost(c.host)
           .withUsername(c.userName)
           .withPassword(c.passwd)
-          .withRealm(Some(c.realm).filter(_.nonEmpty))
+          .withRealm(Option(c.realm).filter(_.nonEmpty))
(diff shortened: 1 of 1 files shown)


3rwww1:
This probably needs unit-testing, although I'm unfamiliar with SBT's testing standards/structure and some pointers would be greatly appreciated  :slightly_smiling_face:

eed3si9n:
Thanks for the contribution.

See https://www.scala-sbt.org/1.x/docs/Testing-sbt-plugins.html and check out https://github.com/sbt/sbt/tree/develop/sbt/src/sbt-test directory. See if you can pick an existing script and add a credential that would cause an error without your patch.

eed3si9n (approved):

PR #3456, opened 2017-08-16
Disable scripted tests that fail in CI

dwijnand:
Fixes #3452

PR #5527, opened 2020-05-03
Add user defined tags to Tests.Group to allow for better concurrency control

frosforever:
Thanks in advance for taking the time to consider and review this PR!

This work adds the ability to custom tag a test Group so that different limitations can be set for different forked jvms. 
Currently, if one wants to provide different forked JVMs with different amounts of memory and also limit the total memory allocated at any one time, you have to resort to [different configurations](https://www.scala-sbt.org/1.x/docs/Testing.html#Additional+test+configurations+with+shared+sources). With this PR the following can instead be done:

```sbt

val ForkedMemory = Tags.Tag("ForkedMem")
val MaxForkedMemory = 12 // Max amount of memory available in the container
Global / concurrentRestrictions := Seq(
  Tags.limit(ForkedMemory, MaxForkMemory)
)

Test / testGrouping := {
  Test / definedTests.value.map { test =>

    val testMem = if (test.name.contains("BigTest")) 4 else 2

    Tests.Group(
        name  = test.name,
        tests = Seq(test),
        runPolicy = Tests.SubProcess(
          ForkOptions()
            .withWorkingDirectory(baseDirectory.value)
            .withRunJVMOptions(
              Vector(s"-XX:MaxRAM=${testMem}g")
            )
        ),
        tags = Seq(ForkedMemory -> testMem)
    )
  }
}
```

Note, the scripted tests are very similar to the [currently disabled `fork-test-group-parallel`](https://github.com/sbt/sbt/pull/3456) and may therefore suffer from the same problem in CI. However, they both pass locally.

### DIFF
--- a/main-actions/src/main/scala/sbt/ForkTests.scala
+++ b/main-actions/src/main/scala/sbt/ForkTests.scala
@@ -28,7 +28,7 @@ private[sbt] object ForkTests {
       classpath: Seq[File],
       fork: ForkOptions,
       log: Logger,
-      tag: Tag
+      tags: (Tag, Int)*
   ): Task[TestOutput] = {
     val opts = processOptions(config, tests, log)
 
@@ -41,7 +41,7 @@ private[sbt] object ForkTests {
         constant(TestOutput(TestResult.Passed, Map.empty[String, SuiteResult], Iterable.empty))
       else
         mainTestTask(runners, opts, classpath, fork, log, config.parallel).tagw(config.tags: _*)
-    main.tag(tag).dependsOn(all(opts.setup): _*) flatMap { results =>
+    main.tagw(tags: _*).dependsOn(all(opts.setup): _*) flatMap { results =>
       all(opts.cleanup).join.map(_ => results)
     }
   }
--- a/main-actions/src/main/scala/sbt/ForkTests.scala
+++ b/main-actions/src/main/scala/sbt/ForkTests.scala
@@ -46,6 +46,18 @@ private[sbt] object ForkTests {
     }
   }
 
+  def apply(
+      runners: Map[TestFramework, Runner],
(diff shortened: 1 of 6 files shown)


frosforever:
It is not obvious to me how to implement this in a binary compatible way or if that is even possible. What is the process for bincompat breaking changes?

eed3si9n:
> It is not obvious to me how to implement this in a binary compatible way or if that is even possible. What is the process for bincompat breaking changes?

I created [Contraband](https://www.scala-sbt.org/contraband/) so we can evolve datatypes, but unfortunately there seems to be many case classes remaining. It's not impossible to add a field to an existing case class. You basically have to reimplement all the previously generated code by hand. See https://github.com/sbt/sbt/commit/b77b0e161ea4e660c318370b947d1b6010b3fad8#diff-53b597b9356cb88766a8b6420543f652 and https://github.com/sbt/sbt/commit/a6f04cf53b4ca64566a24591920f1769d710cb52#diff-53b597b9356cb88766a8b6420543f652.

frosforever:
@eed3si9n Got through the bin compat issues thanks to your pointers! The PR is significantly uglier but I guess that's how it goes.
Is it possible to transition `Tests.Group` to Contraband in a bincompat way? I'm happy to put in the effort if this feature is deemed useful.

eed3si9n:
> Is it possible to transition Tests.Group to Contraband in a bincompat way? 

I don't think it's possible because Group exists within `Tests` object and I don't we can codegen our way through.

eed3si9n (dismissed):
Overall the change LGTM pending Travis CI.

frosforever:
@eed3si9n Thanks again for your time reviewing this! I've rebased on `develop` to fix the build and all checks are passing now.

eed3si9n (approved):

PR #5529, opened 2020-05-05
Remove TaskProgressFormats autogenerated file

eatkins:
This file somehow got stuck in the repo although it wasn't actually
used. In fact, it fails to compile at all because
sbt.internal.util.TaskProgress is defined in main, not util-logging. I
noticed this because metals wasn't working well because it was failing
to compile util-logging because of this file. I think the file was
checked in by accident in e28e052b5bd7ced9e55ccc5665fb44e4009f3ce0.


eed3si9n (approved):
Thanks!

Issue #5482, opened 2020-03-22
Superfluous reload warning for multicommand


OlegYch:
## steps
use sbt 1.3.8
start sbt
modify build file
execute `reload;compile`
## problem

sbt warns about the need to reload
```
[warn] build source files have changed
[warn] modified files:
[warn]   C:\Users\OlegYch\build.sbt
[warn] Apply these changes by running `reload`.
[warn] Automatically reload the build when source changes are detected by setting `Global / onChangedBuildSource := ReloadOnSourceChanges`.
[warn] Disable this warning by setting `Global / onChangedBuildSource := IgnoreSourceChanges`.
```

## expectation

no warning since the command starts with `reload`

PR #5530, opened 2020-05-06
Rework build source monitoring

eatkins:
This PR makes a significant performance improvement to the pre-command processing time. Because of checking the build sources, which was done as an out of band task evaluation, we were often adding 20ms or more to the execution time of every command. This may not seem like a big deal, but sbt runs a lot of tasks in the background like `PopOnFailure` and `StashOnFailure` so those delays could actually add up to something noticeable. I wrote a very crude benchmark project that has a task that overwrites a scala source file with a random identifier, e.g. `object abcdef`. I timed how long it took to run 400 iterations of file overwrite followed by compile. Before this change, it consistently took about 55 seconds and after this change it took 50.

There is further discussion in the commit message in: https://github.com/sbt/sbt/commit/0bb469341387eb18f84a6522e70be8d681e665f8.


eatkins:
Also fixes https://github.com/sbt/sbt/issues/5482.

eed3si9n:
For reference, history of CheckBuildSources.scala - https://github.com/sbt/sbt/commits/develop/main/src/main/scala/sbt/internal/nio/CheckBuildSources.scala

main/src/main/scala/sbt/internal/nio/CheckBuildSources.scala
```diff
@@ -8,56 +8,153 @@
...
 import scala.annotation.tailrec
+import scala.concurrent.duration.{ Deadline => SDeadline, _ }
+
+/**
+ * This class is used to determine whether sbt needs to automatically reload
+ * the build because its source files have changed. In general, it will use
+ * a FileTreeRepository to monitor the build source directories and it will
+ * only actually check whether any sources have changed if the monitor has
+ * detected any events. Because it's using asynchronous monitoring by default,
+ * the automatic reloading should not be relied upon in batch scripting. It is
```
eed3si9n, line 37: In general though does this feature matter unless the user is in an interactive session, like `shell` or `~`?
Is it possible to not have this turned on at all during scripted test or CI processing etc?

eatkins: It could certainly be disabled. The overhead is now low enough that I don't think it really matters.

build.sbt
```diff
@@ -1350,3 +1353,4 @@ ThisBuild / whitesourceAggregateProjectToken := {
 ThisBuild / whitesourceIgnoredScopes ++= Seq("plugin", "scalafmt", "sxr")
 ThisBuild / whitesourceFailOnError := sys.env.contains("WHITESOURCE_PASSWORD") // fail if pwd is present
 ThisBuild / whitesourceForceCheckAllDependencies := true
+ThisBuild / whitesourceForceCheckAllDependencies := true
```
eed3si9n: Looks like a duplicated line?

main/src/main/scala/sbt/internal/nio/CheckBuildSources.scala
eatkins: Even before, the time spent in CheckBuildSources was surely a very small component of the total runtime. But it was noticeable that when you did a no-op enter in the console that you wouldn't get re-prompted immediately.

build.sbt
eatkins: Huh. Don't know how that happened. I'll fix that.

main/src/main/scala/sbt/internal/nio/CheckBuildSources.scala
```diff
@@ -8,56 +8,153 @@
...
+  private[sbt] def init(state: State): State = state.get(CheckBuildSourcesKey) match {
+    case Some(cbs) =>
+      cbs.reset(state)
+      state
+    case _ =>
+      val cbs = new CheckBuildSources
+      cbs.reset(state)
+      state.put(CheckBuildSourcesKey, cbs)
+  }
   private[sbt] def needReloadImpl: Def.Initialize[Task[StateTransform]] = Def.task {
```
eed3si9n, line 153: Is this task definition still needed if you're basically doing everything with state put and get?

eatkins: Not really. I left it for bincompat because it was so easy to implement with the new approach.

eed3si9n: ok. Then could you deprecate the impl and the key plz? Otherwise, I think it's all cool.

eatkins: It actually doesn't make sense to deprecate the key because the build inputs are scoped to that key. Supposed I have some weird project layout like
```
> project
  > project-srcs
     > Build.scala
```
then I can make build monitoring work by adding
```
Global / checkBuildSources / fileInputs += baseDirectory.toGlob / "project" / "project-srcs" / "*.scala"
```
I now remember that I was thinking that if the task could be scoped, it probably should have an implementation, even if it isn't directly used by sbt. That's why I left the definition in.

eed3si9n (approved):

main/src/main/scala/sbt/internal/nio/CheckBuildSources.scala
eed3si9n: ok. That makes sense.

PR #5534, opened 2020-05-06
implement RemoteCache

eed3si9n:
See http://eed3si9n.com/cached-compilation-for-sbt for details.
Ref https://github.com/sbt/zinc/pull/712

This adds `pushRemoteCache`, `pushRemoteCacheTo`, `pullRemoteCache`, etc to implement cached compilation facility.

In addition, the analysis file location is now made more clear.


danicheg:
Do I understand correctly that the generated code located in 'src_managed' will not be cached?

main/src/main/scala/sbt/internal/RemoteCache.scala
```diff
@@ -0,0 +1,217 @@
...
+import sbt.io.IO
+import sbt.io.syntax._
+import sbt.internal.inc.JarUtils
+
+object RemoteCache {
+  final val cachedCompileClassifier = "cached-compile"
+  final val cachedTestClasifier = "cached-test"
+
+  def gitCommitId: String =
+    scala.sys.process.Process("git rev-parse --short HEAD").!!.trim
```
andreaTP, line 31: I think that this line introduce a dependency from the git command line tool.
You can possibly do the same using jgit:
```java
git.getRepository().exactRef("refs/heads/master").getObjectId().abbreviate(7).name()
```
from:
https://stackoverflow.com/questions/39356463/is-there-a-jgit-method-equivalent-to-git-rev-parse-short

It might be an additional dependency for Sbt itself, but I think it would be more "self-contained", what do you think?

He-Pin:
There is one thing like this https://github.com/lihaoyi/mill/pull/726

main/src/main/scala/sbt/internal/RemoteCache.scala
hepin1989: Can this be abstract over and provides an implemented with git remote cache out of the box?

eed3si9n:
@danicheg 

> Do I understand correctly that the generated code located in 'src_managed' will not be cached?

It has not been cached. I am not sure if it's possible to cache that since the input for the code generation logic would be missing and it will be invalidated anyway. Also if the content hashing worked, it shouldn't matter as long as bit-by-bit identical code gets generated?

main/src/main/scala/sbt/internal/RemoteCache.scala
hepin1989: And I don't think it's a good idea to use git as a file cache.

main/src/main/scala/sbt/internal/RemoteCache.scala
```diff
@@ -0,0 +1,217 @@
...
+      },
+      moduleSettings := {
+        val smi = scalaModuleInfo.value
+        ModuleDescriptorConfiguration(remoteCacheProjectId.value, projectInfo.value)
+          .withScalaModuleInfo(smi)
+      },
+      pushRemoteCache.in(Defaults.TaskZero) := (Def.task {
+        val s = streams.value
+        val config = pushRemoteCacheConfiguration.value
+        IvyActions.publish(ivyModule.value, config, s.log)
```
He-Pin, line 147: can this override with any other filestore?

He-Pin: like maybe we can make use of something like `https://www.seafile.com/home/` to keep things sync automatically.

main/src/main/scala/sbt/internal/RemoteCache.scala
eed3si9n: I'm of double mind here. On one hand you're right that it would be good to be self-contained. But on the other, every layer of J-anything could lag behind and introduce problem on its own. See for example `pgp` vs BouncyCastle. If ppl want JGit support they could use sbt-git and write

```scala
ThisBuild / git.gitRunner := com.typesafe.sbt.git.JGitRunner
ThisBuild / remoteCacheId := git.gitHeadCommit.value
```

or something to the effect?

eed3si9n: > And I don't think it's a good idea to use git as a file cache.

The git commit id is the one for the source code, file cache will be a regular Maven repo.

main/src/main/scala/sbt/internal/RemoteCache.scala
eed3si9n: > can this override with any other filestore?

Yes. In the blog post I used:

```scala
ThisBuild / pushRemoteCacheTo := Some(MavenCache("local-cache", file("/tmp/remote-cache")))
```

main/src/main/scala/sbt/internal/RemoteCache.scala
andreaTP: I understand your concerns, still, for exactly one call I would prefer to opt for the self-contained solution.
PGP is a core functionality and it make sense to use the standard implementation.
Here we are using it as a tool for getting the job done.

danicheg:
@eed3si9n

> Also if the content hashing worked, it shouldn't matter as long as bit-by-bit identical code gets generated?

I wonder about the case of code generation from the static scheme that changes quite rarely (case of generation codecs for [protocol buffers](https://github.com/scalapb/ScalaPB)). The plugin tracks an earlier generated code (or, more accurately, it tracks the timestamp of changes the scheme) and doesn't generate/recompile code for every time. For now, we just copy the generation stuff from the previous job run at CI, but it looks like an ad-hoc workaround. Is it an edge case for the caching 'src_managed' stuff?

main/src/main/scala/sbt/internal/RemoteCache.scala
eed3si9n: Every dependency we bring into sbt is a long-term liability. here's the transitive dependencies for the current release:

```
$ cs resolve org.eclipse.jgit:org.eclipse.jgit:5.7.0.202003110725-r
com.googlecode.javaewah:JavaEWAH:1.1.7:default
com.jcraft:jsch:0.1.55:default
com.jcraft:jzlib:1.1.1:default
org.bouncycastle:bcpg-jdk15on:1.64:default
org.bouncycastle:bcpkix-jdk15on:1.64:default
org.bouncycastle:bcprov-jdk15on:1.64:default
org.eclipse.jgit:org.eclipse.jgit:5.7.0.202003110725-r:default
org.slf4j:slf4j-api:1.7.2:default
```

Note that (not really maintained) sbt-git uses a much older version of JGit. Keeping this in the plugin allows the dependencies to be updated without worrying about bincompat impact etc. Eventually when enough ppl start to use this and realize we need JGit, we can always add more later.

PR #5443, opened 2020-02-18
build server protocol support

eed3si9n:
Initial draft for bsp support.

This shows two communication patterns around BSP.
First, if the request can be handled with the build knowledge that is readily available in `NetworkChannel` we can reply immediately. `BuildServerImpl#onBspBuildTargets` is an example for that.

Second, if the request requires `State`, then we can forward the parameter into a globally-scoped input task, and reply back from it. `bspBuildTargetSources` is an example of that since it needs to invoke tasks to generate sources. In the input task the arguments are parsed as URIs, and then a worker task is invoked per build target concurrently using `ScopeFilter`.

PR #5538, opened 2020-05-07
Build server protocol support

adpi2:
This is a continuation of https://github.com/sbt/sbt/pull/5443

This adds build server protocol (https://build-server-protocol.github.io/docs/specification.html) support for sbt server.


lightbend-cla-validator:
Hi @adpi2,

lightbend-cla-validator:
Hi @adpi2,

adpi2:
# BSP implementation status

The core of the BSP implementation is contained in 3 files:
  - [main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala](main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala):
    - the BSP endpoints are defined in the `handler` field
    - the global tasks and settings related to BSP are set in the `globalSettings` field
    - the scoped tasks and settings, that performs on a single BSP build target, are set in the `configSettings` field
  - [main/src/main/scala/sbt/internal/server/BuildServerReporter.scala](main/src/main/scala/sbt/internal/server/BuildServerReporter.scala): implementation of the Zinc reporter that notifies BSP clients with `build/publishDiagnostic` notifications
  - [main/src/main/scala/sbt/internal/server/BspCompileTask.scala](main/src/main/scala/sbt/internal/server/BspCompileTask.scala): notifies BSP clients with `build/taskStart` and `build/taskFinish` notifications

## Endpoints

### `build/initialize`

Replied instantaneously.

### `workspace/buildTargets`

Triggers the `bspWorkspaceBuildTargets` task that depends on the `bspBuildTarget` tasks foreach targets in the build.
Two targets are constructed foreach project, one for `Compile` and one for `Test`. Any other configuration, such as `IntegrationTest`, is not yet supported. To be improved.
The only reason why this is a task and not a setting is that we need to retrieve the scala jars from the dependency manager.

### `build/shutdown`

Don't do anything.

### `build/exit`

Triggers the `shutdown` command which does not appear to be a valid command. To be clarified...

### `buildTarget/sources`

Triggers the `bspBuildTargetSources` input task with the list of requested targets as input. For each target, the `bspBuildTargetSourcesItem` settings is retrieved. All of this is done through the sbt command engine but we should be able to bypass it asynchronously in the near future.

### `buildTarget/dependencySources`

Triggers the `bspBuildTargetDependencySources` input task with the list of requested targets as input, which then retrieve the `bspBuildTargetDependencySourcesItem` setting of each target. This could be done asynchronously as well.

### `buildTarget/scalacOptions`

Triggers the `bspBuildTargetScalacOptions`input task with the list of requested targets as input. This task depends on the `bspBuildTargetScalacOptionsItem` task of each target. This cannot be done asynchronously because it depends on the `externalDependencyClasspath` task which retrieves all of the dependencies.

### `buildTarget/compile`

Triggers the `bspBuildTargetCompile` input task with the list of requested targets as input. This task depends on the `bspBuildTargetCompileItem` task of each target wich itself depends on the `compile` task. The `bspBuildTargetCompileItem` does not fail if the `compile` task fails and it returns a BSP status code.

### Missing endpoints

The missing endpoints of the current implementation are:
 - `textDocument/inverseSources`
 - `buildTarget/resources`
 - `buildTarget/test`
 - `buildTarget/run`
 - `buildTarget/cleanCache`
 - `buildTarget/scalaTestClasses`
 - `buildTarget/scalaMainClasses`

## Logging

The sbt logger broadcasts the logs to all the BSP clients using the `build/logMessage` notification. The inverse is not true, that is, logs from a BSP command is not logged to the shell.

## Compiler diagnostics

An instance of `BuildServerReporter` is always passed to Zinc when compiling.
When Zinc reports a problem (error or warning), the reporter will immediately sends a `build/publishDiagnostics` notification to all the BSP clients. This is fast but verbose and it keeps all clients updated at the same time.

## Task notifications

The `BspCompileTaskProgress` sends `buidl/taskStart` and `build/taskFinish` notifications just before and after that an `compileIncremental` task is actually performed.
The BSP compile report is constructed from the task result and is appended to the notification data.

## Build Server Discovery

After sbt server has started, it creates a `.bsp/sbt.json` file in the project root directory containing the BSP connection details:

```
{
  "name": "sbt",
  "version": "1.4.0-SNAPSHOT",
  "bspVersion": "2.0.0-M5",
  "languages": [
    "scala"
  ],
  "argv": [
    "java",
    "-Xms:100m",
    "-Xmx:100m",
    "-jar",
    "/home/adpi2/.sdkman/candidates/sbt/1.3.3/bin/sbt-launch.jar",
    "-bsp"
  ]
}
```

Any BSP client implementing the Build Server Discovery standard should be able to read this file and then connect to sbt.

The sbt `-bsp` mode is a lightweight client mode that redirects standard input and output to the sbt connection socket of the current project. In case the server is not up anymore, the sbt `-bsp` client will fork a new sbt server in a secondary process and then connect to it. The server won't be listening until it has loaded its settings, which may take up to a minute. It is recommended to start the sbt server manually before trying to connect with the `-bsp` client mode.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -172,44 +173,47 @@ object BuildServerProtocol {
       )
     )
 
-  private def bspBuildTargetSetting: Def.Initialize[BuildTarget] = Def.settingDyn {
+  private def bspBuildTargetTask: Def.Initialize[Task[BuildTarget]] = Def.taskDyn {
```
eed3si9n: It's too bad build target had to become a task. I tried to keep that a synchronous thing in #5443

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -112,13 +112,14 @@ object BuildServerProtocol {
       SourcesItem(id, items)
     },
     bspBuildTargetCompileItem := bspCompileTask.value,
-    bspBuildTargetScalacOptionsItem :=
+    bspBuildTargetScalacOptionsItem := {
       ScalacOptionsItem(
         target = buildTargetIdentifier.value,
         options = scalacOptions.value.toVector,
-        classpath = fullClasspath.value.toVector.map(_.data.toURI),
+        classpath = dependencyClasspath.value.map(_.data.toURI).toVector,
```
eed3si9n: 👍

main/src/main/scala/sbt/internal/server/NetworkChannel.scala
```diff
@@ -613,7 +613,7 @@ final class NetworkChannel(
   def logMessage(level: String, message: String): Unit = {
     import sbt.internal.langserver.codec.JsonProtocol._
     jsonRpcNotify(
-      "window/logMessage",
+      "build/logMessage",
```
eed3si9n, line 616: Could you also update the corresponding 
network client handling? - https://github.com/sbt/sbt/blob/1f3ac6c8cd6fc38f8a6ad85ddca53ade346a7d34/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala#L136

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -34,13 +34,14 @@ object BuildServerProtocol {
...
+    bspWorkspace := Def.settingDyn {
+      val loadedBuild = Keys.loadedBuild.value
+      val scopes: Seq[Scope] = loadedBuild.allProjectRefs.flatMap {
+        case (ref, _) =>
+          bspTargetConfigs.toSeq.map(name => Scope.Global.in(ref, ConfigKey(name)))
       }
-      Def.task {
-        val targetIds = scopes.map(_ / Keys.buildTargetIdentifier).join.value
+      Def.setting {
+        val targetIds = scopes.map(_ / Keys.bspTargetIdentifier).join.value
```
eed3si9n: See the test failure in https://travis-ci.org/github/sbt/sbt/jobs/689252591#L1013-L1030 with https://github.com/sbt/sbt/blob/develop/sbt/src/sbt-test/project/default-auto-plugins/build.sbt. Some subprojects may not have these keys defined. I wonder if you can do `.?`

adpi2: I've moved the `BSP.globalSettings` from `globalSbtCore` to `globalJvmCore`. This should make the test pass. `BSP.configSettings` and `BSP.handler` are already defined in the JVM plugin and so it is more consistent now.

adpi2: It was not enough because all global settings are loaded form `BuildDef` ([see comment in the code](https://github.com/adpi2/sbt/blob/b8ba4ae81f46408a8eeeee61727797727e34a8a0/main/src/main/scala/sbt/internal/BuildDef.scala#L22)), so I did what you suggested.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
adpi2: It has become  a task because sbt needs to retrieve the scala libraries in order to fill the `jars` field of `ScalaBuildTarget`. In #5443 this [field was left empty](https://github.com/eed3si9n/sbt/blob/c31b42085ad52f98ea7b6cb8b33881688e3b1880/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala#L150) which is ok for some client (Metals for instance) but maybe not for all. 

The situation is the same with `build/scalacOptions` because sbt must retrieve the external dependencies to build the classpath. If we fix this for `bspBuildTarget` we should fix it also for `scalacOptions`. How to? Maybe we can play with the cache to preserve synchronicity as much as possible but I have no idea if it is feasible or viable in the long term.

I would rather stay like that for now.

eed3si9n: Yea. That was done intentionally so we can return build target as soon as the build is loaded. Maybe we can ask the BSP maintainers for the clarification on this? Note that it's a `String[]`, not `Uri[]`, so I don't know if this is intended to return the full path to the JAR files.

For import process, I think makes sense to me that the sbt's internal idea of the build targets remain constant unless someone calls `reload`. I'd guess that the more we can reduce the tasks, the general experience of integrating with BSP would feel smoother.

Classpath JARs would be retrieved as task in `buildTarget/scalacOptions` *afterwards*.

adpi2: > Classpath JARs would be retrieved as task in buildTarget/scalacOptions afterwards.

I know from @jastice that the jars field is used in the IntelliJScala plugin because it contains the compiler jars, which are not in the classpath. So the jars field is absolutely necessary.

I really think that `workspace/buildTarget` and `buildTarget/scalacOptions` are complementary and that a client need both to be answered for the project to be imported. The best would be to have both be settings. Until then, it is not a big deal if the `workspace/buildTarget` is a task.

eed3si9n: We can split this as a new issue and discuss later/further with Justin. This is not a blocker for this PR.

eed3si9n (approved):

olafurpg:
🤩🤩🥳🥳

eatkins:
@adpi2 I have been using metals in a project where I have been testing a new sbt feature and noticed that I am seeing a lot of BSP related errors due to duplicate json responses. It seems like a number of the tasks, e.g. `bspBuildTargetSources` do their own reporting for the request id which causes an error like
```
[error] unmatched json response for requestId Some(2): {"status":"Done","channelName":"network-1","execId":"2","commandQueue":["shell"],"exitCode":0}
```
to be reported. The errors go away when I make changes like this in the bsp handlers:
```patch
- val _ = callback.appendExec(s"$command $targets", Some(r.id)) 
+ val _ = callback.appendExec(s"$command $targets", None) 
```
These changes effectively revert a part of https://github.com/sbt/sbt/pull/5538/commits/98750817c673adc046aef8eb8ce78a1debbada06 but I can't tell from the commit why you added the ids to the appended execs.

eed3si9n:
> I have been using metals in a project where I have been testing a new sbt feature and noticed that I am seeing a lot of BSP related errors due to duplicate json responses.

I'm curious what the duplicated responses are and where they are coming from.
I didn't dig into the details, but in theory passing the `Some(r.id)` makes sense because `bspBuildTargetSources` is a globally-scoped input task that _should_ run exactly once to issue worker tasks in subprojects, and then does the replying.

eatkins:
The double reporting is coming from the automatic event logging in MainLoop.scala as part of task evaluation:
https://github.com/sbt/sbt/blob/c3d6260f52dd5594075d3969f7d37ab67e353c5c/main/src/main/scala/sbt/MainLoop.scala#L198-L206
This will duplicate the reporting that is done in the custom task definitions, e.g. https://github.com/sbt/sbt/blob/c3d6260f52dd5594075d3969f7d37ab67e353c5c/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala#L102-L112

eed3si9n:
O wow. So "Done" is coming from the [sbt/exec](https://www.scala-sbt.org/1.x/docs/sbt-server.html#request) that I added around https://github.com/sbt/sbt/commit/892e25d23f27e56d7b9a1e8609b890a63718f13c and Dale extended it in https://github.com/sbt/sbt/pull/3991.

I guess prior to https://github.com/sbt/sbt/pull/5443, we could assume that all commands are downstream of `sbt/exec` request, but now not so much... so we need a way to mark `Exec` that it requires a "Done" response or not.

adpi2:
@eatkins @eed3si9n I am really surprised by this behavior and I cannot experience it myself on the current develop branch.

Before the BSP PR was merged I did #5549 to prevent duplicate response. So basically I have added a `pendingRequests` map in `NetworkChannel` that retains the json-RPC requests that are yet to be answered. As soon as a request is answered by a call to `respondEvent` it is removed from the `pendingRequests` map and it cannot be replied anymore.

In the case of `bspBuildTargetSources`, the `ScalacOptionsResult` is sent first then the `doneEvent` should be filtered out.

@eatkins Can you please share your code so that I can have a look and try to understand what's happening?

PR #5540, opened 2020-05-08
enable ServiceLoader discovery across classloader layers

bjaglin:
`java.util.ServiceLoader` uses `findResources()`, which was not overriden in `ReverseLookupClassLoader`, causing resources available in the descendant classloader not to be discovered when a service loader instance was using the top classloader.

There was something similar in https://github.com/sbt/sbt/blob/6545deacc8a5e51bbb1a0b7c56666b82f3f0c3d3/util/classpath/src/main/scala/sbt/classpath/DualLoader.scala#L72-L88 https://github.com/sbt/sbt/blob/6545deacc8a5e51bbb1a0b7c56666b82f3f0c3d3/util/classpath/src/main/scala/sbt/classpath/DualLoader.scala#L93-L105

I wonder if something similar is needed in `ManagedClassLoader`, but I wouldn't know in which situations.

### DIFF
--- a/main/src/main/java/sbt/internal/ReverseLookupClassLoader.java
+++ b/main/src/main/java/sbt/internal/ReverseLookupClassLoader.java
@@ -8,7 +8,9 @@
 package sbt.internal;
 
 import java.io.File;
+import java.io.IOException;
 import java.net.URL;
+import java.util.Enumeration;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import sbt.util.Logger;
@@ -62,6 +64,31 @@ final class ReverseLookupClassLoader extends ManagedClassLoader {
   public URL findResource(String name) {
     final URL url = super.findResource(name);
     return url != null ? url : directDescendant.get().findResource(name);
+  }
+
+  @Override
+  public Enumeration<URL> findResources(String name) throws IOException {
+    final Enumeration<URL> parentResources = super.findResources(name);
+
+    if (directDescendant.get() == null) {
+      return parentResources;
+    }
+
+    final Enumeration<URL> directDescendantResources = directDescendant.get().findResources(name);
+    return new Enumeration<URL>() {
+      @Override
+      public boolean hasMoreElements() {
+        return parentResources.hasMoreElements() || directDescendantResources.hasMoreElements();
+      }
+
+      @Override
+      public URL nextElement() {
+        if (parentResources.hasMoreElements()) {
+          return parentResources.nextElement();
+        }
+        return directDescendantResources.nextElement();
+      }
+    };
   }
 
   void setup(final File tmpDir) {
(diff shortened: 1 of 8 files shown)


eed3si9n:
Note `java.nio.file.AccessDeniedException` in AppVeyor is happening to other tests, so it's an unrelated failure.

main/src/main/java/sbt/internal/ReverseLookupClassLoader.java
```diff
@@ -64,6 +66,31 @@ public URL findResource(String name) {
...
+      public boolean hasMoreElements() {
+        return parentResources.hasMoreElements() || directDescendantResources.hasMoreElements();
+      }
+
+      @Override
+      public URL nextElement() {
+        if (parentResources.hasMoreElements()) {
+          return parentResources.nextElement();
+        }
+        return directDescendantResources.nextElement();
```
bjaglin, line 89: I am unsure about the "right" order here, as it matters for https://github.com/lightbend/config/blob/c17d58e09af11f9e2a9ddae2eccca18f5f0f01f6/HOCON.md#conventional-configuration-files-for-jvm-apps for example.
> the reference config is made up of all resources named `reference.conf` on the classpath, merged in the order they are returned by `ClassLoader.getResources()`; also, system property overrides are applied.

eatkins: I’m wondering if the two enumerations should be eagerly merged and stored in memory. Naively it looks possible that if the same resource was found in multiple jars across layers that it would be returned twice in the enumeration? Or they could be lazily merged like what is done here but with a filter to exclude resources from the descendant classloader that have the same name as resources loaded by the parent loader.

I think the ordering makes sense that resources come from the parent first.

bjaglin: > Naively it looks possible that if the same resource was found in multiple jars across layers that it would be returned twice in the enumeration? Or they could be lazily merged like what is done here but with a filter to exclude resources from the descendant classloader that have the same name as resources loaded by the parent loader.

The resources returned here have the same name by definition, don't they? Or do you mean that we should de-dup based on the full URL, so that a JAR that would be referenced by both classloaders would only returns its resource once?

bjaglin: > I think the ordering makes sense that resources come from the parent first.

Looking at the javadoc, that's indeed what's expected :+1:

https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#getResources-java.lang.String-
>  The search order is described in the documentation for `getResource(String)` 

https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#getResource-java.lang.String-
> This method will first search the parent class loader for the resource

eatkins: I guess what I am confused on is whether `findResources` should return multiple `URL`s with the same resource name or if like `findResource` it should only return one per resource. Now that I’ve typed that out, it seems to me that the code is probably doing the right thing.

eatkins (approved):
LGTM

bjaglin:
It looks like this PR and the 1.3.x backport don't show up in the upcoming github milestones - is that expected? Maybe change is not significant enough?

eed3si9n:
I use the milestone primarily to group bugs. All changes wil be documented in the release note.

bjaglin:
ok, thanks for the info @eed3si9n !

PR #5546, opened 2020-05-09
IO 1.4.0-M6

eed3si9n:
Ref https://github.com/sbt/io/pull/298
Ref https://github.com/sbt/sbt/pull/5543


eed3si9n:
Passed on the first try.

PR #5549, opened 2020-05-12
Prevent more than one response per json RPC request

### LINKED ISSUE #5548: Prevent more than one json RPC response in NetworkChannel
In `NetworkChannel`, both `publishEvent` and `respondEvent` methods send a json RPC response back to the client. This implies that sbt often sends two responses to a single request and sometime more if the `State.respondEvent` is called more than once. This is a violation of the json RPC protocol, and consequently is not supported by many clients. Sending multiple response to single request should be prevented.

The second issue is that `CommandExchange` broadcasts every response to all network channels whereas it should respond to a single client.

Ideally the API itself should not allow it, meaning that the `jsonRpcRespond` method (or `jsonRpcError`) should be called once and only once at the end of each request. Consequently `jsonRpcResond` should be visible only within a `ServerHandler` and not within a `Task`. Yet the current public API in `UpperStateOps` and `ServerCallback` should be maintained.

Another possible solution would be to filter out all subsequent responses in `CommandExchange` or `NetworkChannel`. One weakness of this solution is that it may be hard for someone to understand why some response is sent and not another one.  Another weakness it that sbt must store whether a request has been answered or not.

adpi2:
fix #5548

### DIFF
--- a/main/src/main/scala/sbt/internal/server/NetworkChannel.scala
+++ b/main/src/main/scala/sbt/internal/server/NetworkChannel.scala
@@ -12,22 +12,22 @@ package server
 import java.net.{ Socket, SocketTimeoutException }
 import java.util.concurrent.atomic.AtomicBoolean
 
-import sjsonnew._
-
-import scala.annotation.tailrec
-import sbt.protocol._
 import sbt.internal.langserver.{ CancelRequestParams, ErrorCodes }
-import sbt.internal.util.{ ObjectEvent, StringEvent }
-import sbt.internal.util.complete.Parser
-import sbt.internal.util.codec.JValueFormats
 import sbt.internal.protocol.{
   JsonRpcNotificationMessage,
   JsonRpcRequestMessage,
   JsonRpcResponseError
 }
+import sbt.internal.util.codec.JValueFormats
+import sbt.internal.util.complete.Parser
+import sbt.internal.util.{ ObjectEvent, StringEvent }
+import sbt.protocol._
 import sbt.util.Logger
-import sjsonnew.support.scalajson.unsafe.Converter
+import sjsonnew._
+import sjsonnew.support.scalajson.unsafe.{ CompactPrinter, Converter }
 
+import scala.annotation.tailrec
+import scala.collection.mutable
 import scala.util.Try
 import scala.util.control.NonFatal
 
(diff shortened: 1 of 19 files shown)


adpi2:
This PR is ready for review. It contains, in order of commits:

- Make `id` mandatory in a json RPC response in order to comply to the json RPC spec.
- Add some failing tests that check that no more than one response can be send back to the client.
- Make the tests pass by adding a mutable set of unanswered requests, called `onGoingRequests` in `NetworkChannel`:
  - If a request has already been responded to, the subsequent responses are evicted
  - In case of a subsequent error, it is still notified to the client (not a response but a notification).
  - All requests will eventually receive a reply from the MainLoop, either done or error, thus preventing the `onGoingRequests` set to grow indefinitely
- A response is only sent to the source channel. On the contrary, notifications are broadcasted to all channels.
- `publish` methods in `CommandExchange` and `NetwordChannel` are renamed either `respond` or `notify` for clarity

main/src/main/scala/sbt/internal/server/NetworkChannel.scala
```diff
@@ -53,6 +56,7 @@ final class NetworkChannel(
   private val VsCode = sbt.protocol.Serialization.VsCode
   private val VsCodeOld = "application/vscode-jsonrpc; charset=utf8"
   private lazy val jsonFormat = new sjsonnew.BasicJsonProtocol with JValueFormats {}
+  private val onGoingRequests: mutable.Set[String] = mutable.Set()
```
eed3si9n: `on...` is confusing with event handler like `onClick(...)`. Let's call this `pendingRequests`. Also it might be useful to keep the entire Request objects since it shouldn't add up in memory.

eed3si9n:
Overall I think we're on the right track.

Please add mima exclusion rules for the failing sbt.internal classes.

main/src/main/scala/sbt/internal/server/NetworkChannel.scala
adpi2: done

eed3si9n:
```
/home/travis/build/sbt/sbt/build.sbt:682: error: value exclude is not a member of com.typesafe.tools.mima.core.ProblemFilter
possible cause: maybe a semicolon is missing before `value exclude'?
      exclude[DirectMissingMethodProblem]("sbt.internal.*"),
      ^
```

eed3si9n (approved):

PR #5552, opened 2020-05-12
implement Def.promise

eed3si9n:
See http://eed3si9n.com/promise-in-sbt for details.

This adds `Def.promise` a facility that wraps `scala.concurrent.Promise`. Project layer, there's an implicit for task-that-returns-promise (`Def.Initialize[Task[PromiseWrap[A]]]`) that would inject `await` method, which returns a task. This is a special task that is tagged with `Tags.Sentinel` so that it will bypass the concurrent restrictions. Since there's no CPU- or IO-bound work, this should be ok.

The purpose of this promise for long-running task to communicate with another task midway.

### DIFF
--- a/main-settings/src/main/scala/sbt/Def.scala
+++ b/main-settings/src/main/scala/sbt/Def.scala
@@ -235,6 +235,12 @@ object Def extends Init[Scope] with TaskMacroExtra with InitializeImplicits {
   def inputTaskDyn[T](t: Def.Initialize[Task[T]]): Def.Initialize[InputTask[T]] =
     macro inputTaskDynMacroImpl[T]
 
+  /** Returns `PromiseWrap[A]`, which is a wrapper around `scala.concurrent.Promise`.
+   * When a task is typed promise (e.g. `Def.Initialize[Task[PromiseWrap[A]]]`),an implicit
+   * method called `await` is injected which will run in a thread outside of concurrent restriction budget.
+   */
+  def promise[A]: PromiseWrap[A] = new PromiseWrap[A]()
+
   // The following conversions enable the types Initialize[T], Initialize[Task[T]], and Task[T] to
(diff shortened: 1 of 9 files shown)


main-settings/src/main/scala/sbt/Def.scala
```diff
@@ -234,6 +234,7 @@ object Def extends Init[Scope] with TaskMacroExtra with InitializeImplicits {
   def inputTask[T](t: T): Def.Initialize[InputTask[T]] = macro inputTaskMacroImpl[T]
   def inputTaskDyn[T](t: Def.Initialize[Task[T]]): Def.Initialize[InputTask[T]] =
     macro inputTaskDynMacroImpl[T]
+  def promise[A]: PromiseWrap[A] = new PromiseWrap[A]()
```
He-Pin, line 242: Add some comments?

main/src/main/scala/sbt/Project.scala
```diff
@@ -865,6 +865,19 @@ object Project extends ProjectExtra {
...
   }
 
+  final class RichTaskPromise[A](i: Def.Initialize[Task[PromiseWrap[A]]]) {
+    import scala.concurrent.Await
+    import scala.concurrent.duration._
+    def await: Def.Initialize[Task[A]] =
+      (Def
+        .task {
+          val p = i.value
+          val result: A = Await.result(p.underlying.future, Duration.Inf)
```
He-Pin: can the timeout be passed in?

He-Pin:
With a ProgressiveCallback too?https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/concurrent/GenericProgressiveFutureListener.java

main-settings/src/main/scala/sbt/Def.scala
eed3si9n: Addressed in b78e4f0

PR #5553, opened 2020-05-13
[1.3.x] enable ServiceLoader discovery across classloader layers

bjaglin:
Backport of #5540 

Since the cherry-pick applies easily, I was wondering if we could backport this?


eed3si9n:
+1 from me. @eatkins?

eed3si9n (approved):

Issue #3258, opened 2017-06-09
Make `inspect tree` see through `Def.settingDyn` (not `taskDyn`)


sjrd:
I am well aware that `inspect tree` cannot reasonably follow the dynamic dependencies through a `Def.taskDyn`, as it would have to run the task to do so, and `inspect` is not supposed to do that. However, it should be able to see through the dynamic dependencies of a `Def.settingDyn`. Indeed, in a `settingDyn`, the part that decides further dependencies is a setting, and therefore need not be evaluated. We already know that `inspect tree` is able and allowed to read the actual value of settings, since it actually displays them.

## Steps

Using sbt 0.13.15 and the following `build.sbt` (without anything else):
```scala
scalaVersion in ThisBuild := "2.11.11"

val scrutineeSetting = settingKey[Boolean]("From this depends the setting and task graph")
val scopeTask1 = taskKey[String]("Scope task 1")
val scopeTask2 = taskKey[String]("Scope task 2")
val someDependentSetting = settingKey[String]("Some dependent setting")
val someDependentTask = taskKey[String]("Some dependent task")

scrutineeSetting := true

someDependentSetting in scopeTask1 := "Some dependent setting in scope task 1"
someDependentSetting in scopeTask2 := "Some dependent setting in scope task 2"

someDependentTask in scopeTask1 := "Some dependent task in scope task 1"
someDependentTask in scopeTask2 := "Some dependent task in scope task 2"

someDependentSetting := {
  val b = scrutineeSetting.value
  if (b)
    Def.setting((someDependentSetting in scopeTask1).value).value
  else
    Def.setting((someDependentSetting in scopeTask2).value).value
}

someDependentTask := {
  val b = scrutineeSetting.value
  if (b)
    Def.task((someDependentTask in scopeTask1).value).value
  else
    Def.task((someDependentTask in scopeTask2).value).value
}
```
(I use convoluted `Def.setting`s and `Def.task`s on purpose, to get the above closer to the "failing" behavior below.)

I can observe the following:
```
[info] Loading global plugins from /localhome/doeraene/.sbt/0.13/plugins
[info] Loading project definition from /localhome/doeraene/projects/sbt-test/project
[info] Set current project to sbt-test (in build file:/localhome/doeraene/projects/sbt-test/)
> show someDependentSetting
[info] Some dependent setting in scope task 1
> inspect tree someDependentSetting
[info] *:someDependentSetting = Some dependent setting in scope task 1
[...]

PR #5558, opened 2020-05-16
Selective functor

eed3si9n:
See details in this blog post - http://eed3si9n.com/selective-functor-in-sbt

This implements Selective functor for `Either[A, B]` "task" (`Initialize[Task[Either[A, B]]]`).
The selective functor allows an encoding of if-expression:

```scala
  def ifS[A](
      x: Def.Initialize[Task[Boolean]]
  )(t: Def.Initialize[Task[A]])(e: Def.Initialize[Task[A]]): Def.Initialize[Task[A]]
```

In addition, this provides `Def.taskIf` macro that accepts an if-expression or a block ending in an if-expression.

This allows rewrite of `publishTask` regaining `inspect tree publish`.

```scala
  def publishTask(config: TaskKey[PublishConfiguration]): Initialize[Task[Unit]] =
    Def.taskIf {
      if ((publish / skip).value) {
        val s = streams.value
        val ref = thisProjectRef.value
        s.log.debug(s"Skipping publish* for ${ref.project}")
      } else {
        val s = streams.value
        IvyActions.publish(ivyModule.value, config.value, s.log)
      }
    } tag (Tags.Publish, Tags.Network)
```

**Update**: I've updated this PR so `Def.task(...)`, `:=` would also auto lift to use `Def.ifS` when the task body consists of a single `if`-expression:

```scala
    // automatic conversion to Def.ifS
    bar := {
      if (number.value < 0) negAction.value
      else if (number.value == 0) zeroAction.value
      else posAction.value
    }
```

### DIFF
--- a/internal/util-collection/src/main/scala/sbt/internal/util/Classes.scala
+++ b/internal/util-collection/src/main/scala/sbt/internal/util/Classes.scala
@@ -14,6 +14,10 @@ object Classes {
     def map[S, T](f: S => T, v: M[S]): M[T]
   }
 
+  trait Selective[M[_]] extends Applicative[M] {
+    def select[A, B](fab: M[Either[A, B]])(fn: M[A => B]): M[B]
+  }
+
   trait Monad[M[_]] extends Applicative[M] {
     def flatten[T](m: M[M[T]]): M[T]
   }
--- a/main-settings/src/main/scala/sbt/Def.scala
+++ b/main-settings/src/main/scala/sbt/Def.scala
@@ -221,6 +221,7 @@ object Def extends Init[Scope] with TaskMacroExtra with InitializeImplicits {
     inputTaskDynMacroImpl,
     inputTaskMacroImpl,
     taskDynMacroImpl,
+    taskIfMacroImpl,
     taskMacroImpl
   }
   import std._
@@ -234,6 +235,7 @@ object Def extends Init[Scope] with TaskMacroExtra with InitializeImplicits {
   def inputTask[T](t: T): Def.Initialize[InputTask[T]] = macro inputTaskMacroImpl[T]
   def inputTaskDyn[T](t: Def.Initialize[Task[T]]): Def.Initialize[InputTask[T]] =
     macro inputTaskDynMacroImpl[T]
+  def taskIf[T](a: Def.Initialize[Task[T]]): Def.Initialize[Task[T]] = macro taskIfMacroImpl[T]
 
   private[sbt] def selectITask[A, B](
       fab: Initialize[Task[Either[A, B]]],
--- a/main-settings/src/main/scala/sbt/Def.scala
+++ b/main-settings/src/main/scala/sbt/Def.scala
@@ -235,7 +235,7 @@ object Def extends Init[Scope] with TaskMacroExtra with InitializeImplicits {
   def inputTask[T](t: T): Def.Initialize[InputTask[T]] = macro inputTaskMacroImpl[T]
   def inputTaskDyn[T](t: Def.Initialize[Task[T]]): Def.Initialize[InputTask[T]] =
     macro inputTaskDynMacroImpl[T]
-  def taskIf[T](a: Def.Initialize[Task[T]]): Def.Initialize[Task[T]] = macro taskIfMacroImpl[T]
+  def taskIf[T](a: T): Def.Initialize[Task[T]] = macro taskIfMacroImpl[T]
 
   private[sbt] def selectITask[A, B](
       fab: Initialize[Task[Either[A, B]]],
--- a/main-settings/src/main/scala/sbt/Def.scala
+++ b/main-settings/src/main/scala/sbt/Def.scala
@@ -235,6 +235,41 @@ object Def extends Init[Scope] with TaskMacroExtra with InitializeImplicits {
   def inputTaskDyn[T](t: Def.Initialize[Task[T]]): Def.Initialize[InputTask[T]] =
     macro inputTaskDynMacroImpl[T]
 
+  private[sbt] def selectITask[A, B](
+      fab: Initialize[Task[Either[A, B]]],
+      fin: Initialize[Task[A => B]]
+  ): Initialize[Task[B]] =
+    fab.zipWith(fin)((ab, in) => TaskExtra.select(ab, in))
+
+  import Scoped.syntax._
+
+  // derived from select
+  private[sbt] def branchS[A, B, C](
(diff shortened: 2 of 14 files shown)


sjrd:
Could the same infrastructure be used to implement https://github.com/sbt/sbt/issues/3258 ?

bjaglin:
nice! couldn't `taskDynMacroImpl` automatically recognize that simple pattern in the body so that a new, separate macro is not necessary (but still available)? That would benefit most of the `taskDyn` out there.

eatkins:
This is great. To add onto @bjaglin's comment, I think it should be possible to update the `taskMacroImpl` to recognize if the last statement in the block is a conditional or a pattern match and automatically lift it into a selective so that it could be written:
```scala
  def publishTask(config: TaskKey[PublishConfiguration]): Initialize[Task[Unit]] =
    Def.task {
      if ((publish / skip).value) {
        val s = streams.value
        val ref = thisProjectRef.value
        s.log.debug(s"Skipping publish* for ${ref.project}")
      } else {
        val s = streams.value
        IvyActions.publish(ivyModule.value, config.value, s.log)
      }
    } tag (Tags.Publish, Tags.Network)
```
Granted the task macros are not exactly easy to work with and are a bit fragile so I can't say _how_ difficult it may or may not be but I actually think the encoding is the more important part so it might not be too bad. That is assuming that the syntax that I propose is actually desired.

eed3si9n:
@sjrd I don't understand the motivation for `Def.settingDyn(...)`. Is it trying to avoid the evaluation of settings conditionally? I'm guessing that this `Def.taskIf(...)` in effect gives you what you wanted, which is `inspect tree` + conditional task.

eed3si9n:
@bjaglin wrote:

> couldn't `taskDynMacroImpl` automatically recognize that simple pattern in the body so that a new, separate macro is not necessary (but still available)? That would benefit most of the `taskDyn` out there.

I think this is tricky because `Def.taskDyn(...)` / `flatMap` that cannot `inspect tree` is not exactly "wrong", so even if I can automatically recognize the pattern the behavior may be a bit too subtle.

Same as `flatMap`, `Def.taskDyn` is able to capture a value and forward it to the inner tasks:

```scala
Def.taskDyn {
  val s = streams.value
  if ((publish / skip).value) Def.task { s.log.info("foo") }
  else Def.task { s.log.info("bar") }
}
```

I think this is fairly common pattern. `Selective` functor specifically loses that power of passing the value in from the outer scope. I'm still allowing block in `Def.taskIf(...)` but all of the vals must end in the if-condition, and cannot be used in the clauses.

So now the question is what to do with the above dynamic task. It might be recognized that it does NOT fit the `Def.taskIf` pattern. I'd need to just use `flatMap`, but likely build user may have intended it to be a conditional task.

@eatkins wrote:

> I think it should be possible to update the `taskMacroImpl` to recognize if the last statement in the block is a conditional or a pattern match and automatically lift it into a selective so that it could be written

Yeah. Between auto-converting `Def.taskDyn(...)` and `Def.task(...)`, I think `Def.task(...)` version does look cooler because it's all pure code without `Def.task {... }` wrapping, and I think it's a direction worth pursuing. Also because the unadorned `:= { ... }` can take advantage of it too. The same conundrum of "what if it's close, but no match" applies.

Maybe we could take a hint from tail call optimization, and create a `@tailrec`-like assertion mechanism. Since we need to check it from inside a macro I'm guessing it needs to be "check for presence of implicit" like:

```scala
  def publishTask(config: TaskKey[PublishConfiguration]): Initialize[Task[Unit]] =
    Def.task {
      import sbt.dsl.tailconditional // this adds assertion that the task is conditional
      if ((publish / skip).value) {
        val s = streams.value
        val ref = thisProjectRef.value
        s.log.debug(s"Skipping publish* for ${ref.project}")
      } else {
        val s = streams.value
        IvyActions.publish(ivyModule.value, config.value, s.log)
      }
    } tag (Tags.Publish, Tags.Network)
```

eed3si9n:
For a lower hanging change to this PR, maybe I can change the `Def.taskIf(...)` macro so that its takes if-expression of `A` as opposed to `Def.Initialize[Task[A]]` so we can do similar to what Ethan did. I'd just do `Def.task {...}` wrapping in the macro.

eatkins:
> Maybe we could take a hint from tail call optimization, and create a @tailrec-like assertion mechanism. Since we need to check it from inside a macro I'm guessing it needs to be "check for presence of implicit" like...

The linter already detects the value calls in the if statements, so I don't think we'd even need an import to ensure auto lifting. You already get a warning about having `foo.value` inside of an if block if you try to write the kind of task that I added above.

dwijnand:
> @sjrd I don't understand the motivation for `Def.settingDyn(...)`. Is it trying to avoid the evaluation of settings conditionally? I'm guessing that this `Def.taskIf(...)` in effect gives you what you wanted, which is `inspect tree` + conditional task.

`Def.settingDyn(...)` allows you to define setting values based on the runtime value of other settings.  E.g. `canLaunchMissles := Def.settingDyn(role.value == President)`.  Your changes here are in the `EvaluateTask` interpreter - to do something about `settingDyn` you'd probably have to change `EvaluateSettings`.

sjrd:
> @sjrd I don't understand the motivation for `Def.settingDyn(...)`. Is it trying to avoid the evaluation of settings conditionally? I'm guessing that this `Def.taskIf(...)` in effect gives you what you wanted, which is `inspect tree` + conditional task.

It evaluates *tasks* conditionally based on the value of a *setting*. `Def.settingDyn` can be used for `TaskKey`s (by instantiating `T = Task[U]` for some `U`); it's just that what the dynamism depends on must be only settings. In #3258 I argue that we should be able to get much better `inspect tree` results with the existing API.

At the very least it could be the same information that you bring here. But for `settingDyn` it could in theory be even better, because `inspect tree` would only show the branch that is *actually* taken (instead of both branches, as you do here).

eed3si9n:
@eatkins ok. I implemented auto conversion in https://github.com/sbt/sbt/pull/5558/commits/b43f663b0e081647a7e4ca09a1af03b920a1a0a7. As you mentioned there's already the DSL linter, so this change in semantics should be safe. At least it shouldn't be surprising. I do have this nagging feeling that we could be scaling down the [uncanny valley](https://spectrum.ieee.org/automaton/robotics/humanoids/what-is-the-uncanny-valley).

For users who grok this, the fact that we don't have to do `Def.taskIf().value`, and use `:= ...` makes it more elegant. But this (`Def.task` becomes Selective when it's a single `if`) is one cognitive load build authors would need to learn to master sbt.

PR #5560, opened 2020-05-17
Remove .sbtopts

eatkins:
These prevent sbt from loading the sbt project with the windows batch script. I think
it's better for each developer to manually decide how much ram they want
to devote to sbt anyway.

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,3 +1,0 @@
--J-Xms2048M
--J-Xmx2048M
--J-Xss2M


eed3si9n:
iirc this change (https://github.com/sbt/sbt/pull/4217) came out of the experience in a Scala Spree. I normally had `SBT_OPTS` defined, but most attendees didn't have it. I am guessing some had trouble running test etc. If there's a Windows issue, we should somehow deal with that, but ignoring that for now, I would like the experience of first contribution to sbt as smooth as possible..

eatkins:
Since removing the dot file, I haven't had any problems with sbt running out of memory on any of the three platforms (linux, mac, windows) that I've been using. I just checked and after running `clean;publishLocal` 10 times, the sbt heap was at 220MB. I really don't think this dotfile is necessary and it caused me a lot of pain on my windows vm in the past. When I ran into this problem before, I just gave up on the sbt batch file and had to run an annoying java command to bypass it.

eed3si9n:
ok. I guess we can try removing it, and reinstate it when we run into some issue.

PR #5571, opened 2020-05-23
Introduce a Submission class for submitted works.

diesalbla:
Replace the use of too many lambdas (difficult to debug and profile):

- We add a an abstract `Submission` class, to represent some work to be done. The work is represented as the method of this class, not a lambda.
- This class replaces the `Enqueue` inside `ConcurrentRestrictions`.
- Extract the big anonymous class in `ConcurrentRestrictions` as a class.

### Motivations. 

I recently run and profiled a compilation (with several iterations of warmup) of the `http4s-core` package, using Scala 2.13.2 on Sbt 1.3.10. It recorded over 40 MiB of allocations of a Lambda, related to this area of SBT. I had some difficulty to follow the call stack, but using classes instead of `() => T` continuations could help.  

sbt_lambda_profile

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -130,6 +130,10 @@ def mimaSettingsSince(versions: Seq[String]) = Def settings (
   },
   mimaBinaryIssueFilters ++= Seq(
     // Changes in the internal package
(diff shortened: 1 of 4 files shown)


lightbend-cla-validator:
Hi @diesalbla,

PR #5610, opened 2020-06-12
Drop old application/sbt-x1 protocol

eed3si9n:
This will also assume that when content-type header is not passed, it will assume LSP/BSP.

### DIFF
--- a/main/src/main/scala/sbt/internal/server/NetworkChannel.scala
+++ b/main/src/main/scala/sbt/internal/server/NetworkChannel.scala
@@ -145,7 +145,10 @@ final class NetworkChannel(
                         case Some(_) =>
                           state = InHeader
                           process()
-                        case _ => log.error("Got invalid chunk from client: " + str)
+                        case _ =>
+                          val msg = s"got invalid chunk from client: $str"
+                          log.error(msg)
+                          logMessage("error", msg)
                       }
                   }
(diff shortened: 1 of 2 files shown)


main/src/main/scala/sbt/internal/server/NetworkChannel.scala
```diff
@@ -215,60 +214,43 @@ final class NetworkChannel(
...
-            logMessage(
-              "error",
-              s"Got invalid chunk from client (${new String(chunk.toArray, "UTF-8")}): $errorDesc"
-            )
-        }
-      } else {
-        contentType match {
-          case SbtX1Protocol =>
-            Serialization
-              .deserializeCommand(chunk)
```
adpi2, line 248: The `Serialization` `deserializeCommand` and `serializeCommand` methods are not used anymore. So we should probably remove them.

adpi2 (dismissed):
LGTM!

I tried this change with BSP and its working fine. `sbt -client` mode does not work but for other reasons.

main/src/main/scala/sbt/internal/server/NetworkChannel.scala
eed3si9n: I'll deprecate them for now to avoid bincompat issue.

PR #5620, opened 2020-06-15
sbtn, native thin client (nearly instantaneous startup)

eatkins:
This (rather hefty) PR expands the experimental thin client feature to have almost full parity with running commands through the server either in batch mode or using the embedded shell. The feature parity is achieved by performing server side rendering and having the remote client attach itself to the server so that it functions primarily as a relay between the user's terminal input/output and sbt. The relay client can be built using [graalvm](https://www.graalvm.org/docs/reference-manual/native-image/) which allows it to start up nearly instantly. On my laptop, the startup overhead, including connecting to the running server seems to take between 40 and 60ms. For example, with the native client (alised to `sbtc`), the command `sbtc show sbtVersion` prints the sbt version and terminates in approximately 60ms. With low latency startup and tab completions, the sbt client can be used like a normal command line utility. Below I include several demos to showcase the functionality. Here is an introductory demo:
[asciicast](https://asciinema.org/a/wAAYjiO0TNW0V7dCSebMQB5wa)
If you'd like to test the native image on your computer, checkout the graal section at the bottom.

My guiding principle for this PR was very simple: a ux that is indistinguishable from running sbt directly in both shell and batch mode on windows, linux and mac (which are the three supported platforms for graalvm). To this end, here is list of features that I have verified work as expected with the remote client: 
* shell and batch mode
  + log lines and success or error message are displayed
  + supershell works
  + automatically disconnects if the server exits, even if the server ended
    abruptly (e.g. with kill -9)
  + start a server instance
    - when no portfile is present
    - when the portfile is present but a connection can't be established
      (usually due to abrupt server exit)
  + ctrl+c exits the client and cancels the active task
  + when multiple clients are connected
    - only one command at a time is ever run
    - command output is directed only to the client that initiated the command
    - when a remote command is running, allow a different client to cancel it
  + watch
    - functionally identical to 1.3.x
    - ctrl+c kills only the running task but not the watch itself
    - ctrl+c exits watch if there is no running task
    - multiple watches can simultaneously be run from multiple clients
  + console task works (this required forking jline)
  + `exit` closes the client but not the server unless the client is the
    embedded console client
  + `shutdown` immediately exits the server even if there are tasks running
  + `reload` works including on failure input options
  + when a client launches a server, the server output is printed until the server binds to the local socket

* batch mode
  + provide tab completions for bash, zsh, fish and powershell
    - tab completions can be used to launch a server if it isn't already running
    - tab completions can optionally trigger a compile for `runMain` `testOnly`
      and `testQuick` to populate the class name caches
  + returns exit code 0 if the command succeeds and 1 if it fails
  + multi commands work and the exit code is 0 if all succeed or 1 if any fail

* shell
  + ctrl+c or ctrl+d with no task running exits the shell
  + ctrl+c or ctrl+d with no task running exits the server when issued from console

* ~~reboot does not yet work because it shuts down the server but it could be
  made to work~~
  reboot works now: https://asciinema.org/a/Ht6zFvhxLVmEKYkTXWJFXeK4n

### Demos
* Windows tab completions + console + supershell: [asciicast](https://asciinema.org/a/ELQmc56R5eWiZvTNpWqO0m2Fd)
* Loading an invalid build, fixing the build and reloading: [asciicast](https://asciinema.org/a/btpYOZaw2sW4LdCxJqQpIHz5a?t=1.3)
* Multiple watches: [asciicast](https://asciinema.org/a/MoDIhPicRxyz9je8mJORSkP1q)
* Remote cancellation:  [asciicast](https://asciinema.org/a/igYB8kRBhatvfmVnxOBGJSfB9)
* Batch multi-commands: [asciicast](https://asciinema.org/a/Qq7PSbNKnJJxNc56bzRX9SgL1)
* Abrupt shutdown [asciicast](https://asciinema.org/a/1RipdiuRTBxAoHS2xPuQPWDIB)
### Implementation

Supporting a client ui requires a paradigm shift in thinking about the structure of the sbt server program. Before sbt had any server capabilities, it contained an embedded shell which fully provided the UI of the program. sbt 1.0 introduced a server that allowed remote clients to connect and issues commands via a json protocol. Any ui for the remote client had to be done client side. At a high level, this PR takes the embedded shell and abstracts out the concept of a terminal. There is an embedded terminal that uses the process input and output streams, but the shell also supports an arbitrary number of remote terminals to connect. There is a json protocol for the remote terminals that allows them to communicate to sbt properties like the terminal dimensions and whether or not color output is enabled. sbt can use that information to perform server side rendering of the output text. The client also forwards keyboard input from stdin to the sbt server via json rpc. This allows the server to receive input from the client even when it is running a task. The server modifies `System.in` and `System.out` during task evaluation to map to the input streams of the terminal that issued the command. This makes plugin commands and tasks that use `System.in` and `System.out` work with the remote client without having to modify the task or command definitions at all.

Each terminal can issue commands to the shell, which will only run one command at a time. To prevent a task like `console` from indefinitely blocking sbt if it is open in, for example, a forgotten tmux session, remote clients can cancel the tasks created by other clients.

A side effect of reworking the server to support multiple clients is that I was able to rework the shell and watch commands so that they are functionalyl identical to 1.3.x but do not poll `System.in` for input. Previously to support multiple clients, we had to poll for user periodically because on posix platforms, read is uninterruptible. Suppose the user is running `~console` and a build is triggered by a source file update. If we blocked on `System.in` in watch, then the first character inputted by the user in the console session would be lost. I avoid this problem by creating a thread for each terminal that reads from `System.in` and writes the result to a queue. Instead of reading from the original `System.in`, we can instead read from that queue, which is interruptible. With some reorganization, I was able to take advantage of this to use blocking reads from the input queue instead of polling the underlying input streams. There were a number of benefits to this change but one that I'll point out is that the cpu utilization of sbt when idle is now 0.0 as measured by `ps`. Prior to this PR, it tended to hover around 2.0 due to the polling every 50ms for input. Input latency should also be lower with this approach. 

If I were building a new program from scratch, it is highly unlikely that I would have designed multi-client UI the way that I did in this PR. This implementation effectively relies upon manipulating global state, e.g. `System.out` during task evaluation. As is nearly always the case when global state is involved, this makes certain aspects of the program more difficult to reason about. The problem though is that `println` is used pervasively. If we didn't route System.out to the appropriate terminal, the ui would work differently depending on whether the user was using the embedded console or a remote terminal. Since we don't control plugins, not handling println seemed like a non-starter.

### graal

The graalvm native images can be generated on appveyor, which supports uploading local artifacts. If you'd like to try a native image for yourself, you can download them from the results of my personal appveyor ci [build](https://ci.appveyor.com/project/eatkins/sbt/builds/33809840):
* windows https://ci.appveyor.com/api/buildjobs/ehh59h6send8x0rh/artifacts/client%2Ftarget%2Fbin%2Fsbtc.exe
*  mac https://ci.appveyor.com/api/buildjobs/8c0b5itbhrs9gt4y/artifacts/client%2Ftarget%2Fbin%2Fsbtc note that osx gatekeeper will block you from running this if you download it from a web browser. It does work if you do something like `curl -sL https://ci.appveyor.com/api/buildjobs/8c0b5itbhrs9gt4y/artifacts/client%2Ftarget%2Fbin%2Fsbtc > sbtc && chmod a+x sbtc`
* linux https://ci.appveyor.com/api/buildjobs/vo8lsr8ii64a1hwn/artifacts/client%2Ftarget%2Fbin%2Fsbtc

Some batch commands like compile actually work with sbt 1.4.0-M1, but others, like console, do not. In order to build a server locally that incorporates these changes, you'd need to publish sbt-ipsocket, launcher and my jline fork. The appveyor.yml changes provide a template for how to do this.


eatkins:
The PR has been updated so that it now uses versions of https://github.com/sbt/jline2 and https://github.com/sbt/ipcsocket that have been published to maven central. That means that anyone can test the new client by cloning this repo, running `sbt publishLocal` and downloading one of the client binaries above.

### bash completions
Add to .bashrc
```
PATH="$PATH:/path/to/dir/containing/sbtc"
source ${SBT_SOURCE_DIR}/client/completions/sbtc.bash 
```

### zsh completions
Add to .zshrc
```
path=($path /path/to/dir/containing/sbtc)
fpath=($fpath ${SBT_SOURCE_DIR}/client/completions)
```

### fish completions
Add to ~/.config/fish/config.fish
```
set PATH $PATH /path/to/dir/containing/sbtc
source ${SBT_SOURCE_DIR}/client/completions/sbtc.fish
```

### powershell completions
```
$env:PATH="$env:PATH;C:\path\to\dir\containing\sbtc"
C:\${SBT_SOURCE_DIR}\client\completions\sbtc.ps1
```

dwijnand:
I'm not sure why or what's causing it, but in my first use of `sbtc ~compile` I got a couple of terminal pages worth of `[warn] Negative time`.  I can spin this comment into its own issue if it's not due to changes here.

client/src/main/java/sbt/client/NativeClient.java
```diff
@@ -0,0 +1,26 @@
...
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.client;
+
+import sbt.internal.client.SimpleClient;
+import java.nio.file.Paths;
+import org.fusesource.jansi.AnsiConsole;
+
+public class NativeClient {
```
dwijnand: Duplicate of `sbt.client.Client`?

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -190,82 +378,497 @@ class NetworkClient(configuration: xsbti.AppConfiguration, arguments: List[Strin
...
+      }
+      i += 1
+    }
+    new Arguments(base.getCanonicalFile, sbtArguments, commandArgs, completionArguments, sbtScript)
+  }
   def run(configuration: xsbti.AppConfiguration, arguments: List[String]): Unit =
     try {
-      new NetworkClient(configuration, arguments)
+      val client = new NetworkClient(configuration, parseArgs(arguments.toArray))
+      System.exit(try {
```
dwijnand: Either don't system exit here or use `sys.exit` and change the result type to `Nothing` and fix the errors that follow:

```
[error] ${BASE}/main-command/src/main/scala/sbt/BasicCommands.scala:414:22: dead code following this construct
[error]     NetworkClient.run(s0.configuration, arguments)
[error] one error found
[error] (commandProj / Compile / compileIncremental) Compilation failed
```

(There's also a `NetworkClient.run` call `xMain`, but `mainProj` never starts to compile.)

client/src/main/java/sbt/client/NativeClient.java
dwijnand: Ah, the `false` vs `true` for `useJNI`.

_mildly wonders off wondering how they could be dedupped..._

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -190,82 +378,497 @@ class NetworkClient(configuration: xsbti.AppConfiguration, arguments: List[Strin
...
+    this(
+      arguments.withBaseDirectory(configuration.baseDirectory),
+      System.in,
+      System.err,
+      System.out,
+    )
+  override def mkSocket(file: File): (Socket, Option[String]) = ClientSocket.socket(file)
+  def this(configuration: xsbti.AppConfiguration, args: List[String]) =
+    this(
+      NetworkClient.parseArgs(args.toArray).withBaseDirectory(configuration.baseDirectory),
```
dwijnand: Which one's meant to win `--sbt-base-directory` or `configuration.baseDirectory`?  It seems like `--sbt-base-directory` never wins, except for completions?  I feel like the base directory (or the `new File("")` fallback) should be an argument of `parseArgs`.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -118,33 +235,104 @@ class NetworkClient(configuration: xsbti.AppConfiguration, arguments: List[Strin
...
+          Vector.empty
+        case (`promptChannel`, _) =>
+          batchMode.set(false)
+          Vector.empty
         case ("textDocument/publishDiagnostics", Some(json)) =>
           import sbt.internal.langserver.codec.JsonProtocol._
           Converter.fromJson[PublishDiagnosticsParams](json) match {
-            case Success(params) => splitDiagnostics(params)
-            case Failure(e)      => Vector()
+            case Success(params) => splitDiagnostics(params); Vector()
```
dwijnand, line 542: Don't drop the diagnostics?

```suggestion
            case Success(params) => splitDiagnostics(params)
```

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -190,82 +378,497 @@ class NetworkClient(configuration: xsbti.AppConfiguration, arguments: List[Strin
...
+    }
+}
+class NetworkClient(
+    override val arguments: Arguments,
+    override val inputStream: InputStream,
+    override val errorStream: PrintStream,
+    override val printStream: PrintStream,
+) extends {
+  override val console: ConsoleInterface = {
+    val appender = ConsoleAppender("thin")
```
dwijnand: Wouldn't this wire to the `printStream`?
```suggestion
    val appender = ConsoleAppender("thin", sbt.internal.util.ConsoleOut.printStreamOut(printStream))
```

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: That's a good point about not calling System.exit. I am curious though why you are seeing those errors because I have never had issues compiling the project with these changes and CI works. Is there anything unusual about your build environment?

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: Maybe? I have no idea why I made that change. It probably dates back to late December or early January.

client/src/main/java/sbt/client/NativeClient.java
eatkins: They probably could be by adding a `--jna` flag or something. The universal script that uses java rather than graal could pass that flag in while the default is to use jni. It is nice that `sbtc` as generated by `native-image` does not need to be wrapped in a shell script to add additional arguments.

eatkins:
> I'm not sure why or what's causing it, but in my first use of sbtc ~compile I got a couple of terminal pages worth of [warn] Negative time. I can spin this comment into its own issue if it's not due to changes here.

That's interesting. I have no idea what is causing that and have never seen it. `git grep "Negative time"` doesn't return anything in the sbt repo.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: Good catch.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: I'll look into this. This was added way back when I had problems getting runtime initialization of graalvm classes to work. At the time `new File("")` was returning a build path so I added the `--sbt-base-directory` to address that. I fixed the `new File("")` issue but left in `--sbt-base-directory`. It probably should be removed. Just ftr, today I am cleaning up `NetworkClient.scala` and `NetworkChannel.scala` so these files will probably look somewhat different by the end of the day.

client/src/main/java/sbt/client/NativeClient.java
eatkins: Yep. I added a `--jna` flag and deleted `NativeClient`. Default is to use jni but the generated client script prepends `--jna` to the argument list.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
dwijnand: Those are the errors that follow from changing to sys.exit and recompiling.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
dwijnand: Same here, spent a few hours learning bits of how it works together.

Let's just say I wish scalafmt added newline between multi-line methods!! o_O

dwijnand:
Looks like you can get it with a bad `java.io.File.setLastModified` call?  https://issues.apache.org/jira/browse/IO-270

eatkins:
Do you have a stack trace? The only place that I can find that sets the modified time in sbt is https://github.com/sbt/sbt/blob/e323f1f713251968b3d4ecda8a7f0ed52593713a/main/src/main/scala/sbt/internal/InMemoryCacheStore.scala#L76-L77 and it seems unlikely that is causing the problem. Maybe it's an issue in zinc perist?
https://github.com/sbt/zinc/blob/8060d28c4a062eea95a0c7e4e6809640ce58753a/internal/zinc-persist/src/main/scala/sbt/internal/inc/binary/converters/ProtobufWriters.scala#L46-L67
I still am a bit confused as to how the value could possibly be less than zero there.

eatkins:
Are you using a remote cache? I see that there is a preserve last modified time flag that maybe is somehow setting negative values? https://github.com/sbt/sbt/blob/406330b3fe6ae4f4fb5d93182ea13f3aecb125a3/main/src/main/scala/sbt/internal/RemoteCache.scala#L268-L273

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: Aah. Reading comprehension fail.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
lorandszakacs: Urelated comment to this PR.

But @dwijnand there is such a config btw, I guess the problem will be when to apply a new scalafmt config to the codebase 😅 

https://scalameta.org/scalafmt/docs/configuration.html#newlinesalwaysbeforemultilinedef

dwijnand: Thanks, @lorandszakacs!  Looks like I want: 

```
newlines.topLevelStatements = [before,after]
```

dwijnand:
No, no remote cache.  Probably something in Zinc, I assume, updating a bunch of analysis metadata or something? 🤷 I guess it might come up again.

internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
```diff
@@ -147,142 +245,192 @@ object Terminal {
...
+      System.setIn(wrappedSystemIn)
+      scala.Console.withIn(proxyInputStream)(f)
     } finally System.setIn(originalIn)
 
-  private[sbt] def withPrintStream[T](f: PrintStream => T): T = writeLock.synchronized {
-    f(originalOut)
+  private[sbt] def withPrintStream[T](f: PrintStream => T): T = console.withPrintStream(f)
+  private[this] val attached = new AtomicBoolean(true)
+  private[this] val terminalHolder = new AtomicReference(wrap(jline.TerminalFactory.get))
+  private[this] val currentTerminal = new AtomicReference[Terminal](terminalHolder.get)
```
dwijnand: These two (terminalHolder and currentTerminal) don't seem to be well managed at the moment.  Or, at the very least, not obviously so.

To start with `jline.TerminalFactory.get` -> `terminalHolder` -> `currentTerminal` (and then immediately `jline.TerminalFactory` is change to be the JLine1/JLine2 wrapper.)

But then `terminalHolder` is read from in `console` and re-initialised in `reset`, while `currentTerminal` is read from in all the proxy components (`proxyInputStream`, `proxyOutputStream`, `ProxyTerminal` and `get`) and changed in `set`.

Is all that correct?  I'm guessing `terminalHolder` is meant to be "live" and `currentTerminal` is meant to be a more persistent instance that proxies for whatever the terminal is actually in use?

A couple of paragraphs on the ideas in play and how to use it (as the `Terminal` object's Scaladoc) would come very handy! ❤️

Also, seeing as there's all 4 adapters/forwarders for the terminals:

* "wrap": jline1 -> jline1/jline2
* "toJLine": Terminal -> jline1/jline2
* "ProxyTerminal": Terminal -> Terminal
* "ConsoleTerminal": jline1/jline2 -> Terminal

I was almost thinking it might be easier to understand and keep right if the terminal stream bits were in a separate object/file.  Not because 600 lines is too much, but because it's complicated by the mutation going around.

eatkins: I'll look into this more tomorrow. I'm not sure how much can be done about the adapters without internally passing around instances of `jline.Terminal2`, which I'm not keen to do. That being said, there's no doubt the organization could be better. The question is how much effort I really want to put into reorganizing. It's at a point where I have spent an absurd amount of time on this project so I'm running out of steam for refactoring.

eatkins:
I split everything into smaller commits that I'm happy enough with so I dropped the WIP from the title. There are still a few commits that are a bit bigger than ideal, but I'd rather not spend a ton more time trying to extract independently compiling fragments out of them.

I updated the graal links to reflect the latest code.

internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
dwijnand: I can propose those changes, and that would work in my favour as it avoids me rebasing/rewriting my local state.

But wdyt about the state of the terminal holder/currentTerminal setup?

eatkins: After performing a few days of git surgery on this PR, I was really unexcited about further changes last night. I'm feeling more sanguine today. I renamed some fields and added some commentary in https://github.com/sbt/sbt/pull/5620/commits/5238c78dfd12d0fac4a5ec588be0d069eaf059b7

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: I made this change.

dwijnand:
Thanks, thanks for the docs.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: I removed the `System.exit` in `NetworkClient.run` among other changes.

main/src/main/scala/sbt/internal/Act.scala
```diff
@@ -424,7 +424,7 @@ object Act {
       def evaluate(kvs: Seq[ScopedKey[_]]): Parser[() => State] = {
         val preparedPairs = anyKeyValues(structure, kvs)
         val showConfig = if (action == PrintAction) {
-          Aggregation.ShowConfig(true, true, println, false)
+          Aggregation.ShowConfig(true, true, s => { print(s + "\n"); System.out.flush() }, false)
```
eed3si9n: How does splitting `println` to `print` + `flush` prevent interleaving? Also `\n` should be replaced with `System.lineSeparator`?

main-command/src/main/scala/sbt/internal/CommandChannel.scala
```diff
@@ -20,31 +21,59 @@ import sbt.protocol.EventMessage
...
-  private val registered: java.util.Set[java.util.Queue[CommandChannel]] = new java.util.HashSet
-  private[sbt] final def register(queue: java.util.Queue[CommandChannel]): Unit = {
-    registered.add(queue)
-    ()
+  private val registered: java.util.Set[java.util.Queue[Exec]] = new java.util.HashSet
+  private val maintenance: java.util.Set[java.util.Queue[MaintenanceTask]] = new java.util.HashSet
+  private[sbt] final def register(
+      queue: java.util.Queue[Exec],
+      maintenanceQueue: java.util.Queue[MaintenanceTask]
+  ): Unit =
```
eed3si9n: What's a maintenanceQueue?

eed3si9n: Just reread the commit message:

> Certain commands such as `exit` should be evaluated immediately. To make
this work, we add the concept of a MaintenanceTask. The CommandExchange
has a background thread that reads MaintenanceTasks and evaluates them
on demand. This allows maintenance tasks to be evaluated even when sbt
is evaluating an exec. If it weren't done this way, when the user typed
exit while a different remote connection was running a command, they
wouldn't be able to exit until the command completed.

What do you think about calling this `FastTrackTask` and `FastTrackQueue` instead? Maintenance sounds like it runs at the end as a clean-up job.

eatkins: Sure: https://github.com/sbt/sbt/pull/5620/commits/ba0d97c9ac8d463502050900a8396785e8d2465c

main/src/main/scala/sbt/internal/Act.scala
eatkins: It's possible that this is wrong. I'm sure there was a reason I made this change but I can't remember why and I can't convince myself it's correct. I went ahead and reverted that commit.

dwijnand:
It doesn't seem possible to call something like:

```
$ sbtc testOnly sbtdynver.TagPatternSpec sbtdynver.GH007 -- -verbosity 2
> testOnly sbtdynver.TagPatternSpec sbtdynver.GH007 2
...
```

I even tried quoting it like `sbtc 'testOnly sbtdynver.TagPatternSpec sbtdynver.GH007 -- -verbosity 2'`, which changed nothing.

eatkins:
Yeah. It looks like argument parsing is broken because of the `--`. That shouldn't be too bad to fix.

eatkins:
I fixed the argument parsing in https://github.com/sbt/sbt/pull/5620/commits/3216848c77b75998532352409f009d8f825ae38b

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -9,94 +9,242 @@ package sbt
...
+import NetworkClient.Arguments
+
+trait ConsoleInterface {
+  def appendLog(level: Level.Value, message: => String): Unit
+  def success(msg: String): Unit
+}
 
-class NetworkClient(configuration: xsbti.AppConfiguration, arguments: List[String]) { self =>
-  private val channelName = new AtomicReference("_")
+class NetworkClient(
```
eed3si9n, line 89: Is this NetworkClient being double dipped? Once for the thin client v1 (https://github.com/sbt/sbt/pull/2859) and now for this? 

For debugging, documentation etc, I think it would be helpful to somehow distinguish the GraalVM-built version on the welcome tag, like:

```
$ sbtc
[info] entering *experimental* native thin client - BEEP WHIRR
```

vs

```
$ sbt -client
[info] entering *experimental* JVM thin client - BEEP WHIRR
```

Is the intent to keep the old `-client` for non-GraalVM fallback, or build another JAR for the purpose?

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -9,94 +9,242 @@ package sbt
...
+    console: ConsoleInterface,
+    arguments: Arguments,
+    inputStream: InputStream,
+    errorStream: PrintStream,
+    printStream: PrintStream,
```
eed3si9n, line 94: Could you add Scaladoc on the significance of passing these streams into `NetworkClient` while your knowledge is fresh in your mind please?

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -9,94 +9,242 @@ package sbt
...
-      } else {
-        if (n <= 0) sys.error(s"timeout. $portfile is not found.")
-        else {
-          Thread.sleep(1000)
-          if ((n - 1) % 10 == 0) {
-            console.appendLog(Level.Info, "waiting for the server...")
-          }
-          waitForPortfile(n - 1)
+  def forkServer(portfile: File, log: Boolean): Unit = {
+    if (log) console.appendLog(Level.Info, "server was not detected. starting an instance")
```
eed3si9n: I wonder if we should also print a warning saying:

```
[info] starting sbt server in the background...
[warn] close the server with `sbtc shutdown` after using
```

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -9,94 +9,242 @@ package sbt
...
-      override def onNotification(msg: JsonRpcNotificationMessage): Unit = self.onNotification(msg)
-      override def onRequest(msg: JsonRpcRequestMessage): Unit = self.onRequest(msg)
-      override def onResponse(msg: JsonRpcResponseMessage): Unit = self.onResponse(msg)
-      override def onShutdown(): Unit = {
-        running.set(false)
+  def init(prompt: Boolean, retry: Boolean): ServerConnection =
+    try {
+      if (!portfile.exists) {
+        if (prompt) {
+          val msg = if (noTab) "" else "No sbt server is running. Press <tab> to start one..."
```
dwijnand, line 157: Btw, I instinctively hit 'y' for this, today.  But I'm not sure why.  We use 'r' for retrying to (re)load, so where is my muscle memory remembering 'y' from?

Just because from that I'm wondering if it should be allowed.  "Press tab to confirm" feels a little unique/original here.  (No big deal.)

eed3si9n: I have a vague memory of tab completion doing something special in some software on double tabs.
From `prompt` it's not obvious, but this `prompt` only becomes true during the completion service call right? (Maybe rename the variable to reflect that?)

eed3si9n: For example in zsh

```
$ ls g[tab]
```

completes to

```
$ ls giter8-bootstrap_2.12-0.13.
```

and hitting the second tab

```
$ ls giter8-bootstrap_2.12-0.13.[tab]
```

completes to

```
$ ls giter8-bootstrap_2.12-0.13.
giter8-bootstrap_2.12-0.13.0.sh  giter8-bootstrap_2.12-0.13.1.sh
```

and hitting the third tab

```
$ ls giter8-bootstrap_2.12-0.13.[tab]
giter8-bootstrap_2.12-0.13.0.sh  giter8-bootstrap_2.12-0.13.1.sh
```

completes to

```
$ ls giter8-bootstrap_2.12-0.13.0.sh
giter8-bootstrap_2.12-0.13.0.sh  giter8-bootstrap_2.12-0.13.1.sh
```

eatkins: The motivation for using tab is that if I know that an sbt server isn't running, I can just double tap tab. Also, as Eugene pointed out, it's not that uncommon for pressing tab multiple times to modify the completions. I'm not opposed to adding 'y' as another valid option. The main thing is that I wanted an opt out in case users don't want tab completions to do something potentially slow.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: Yeah. I was thinking about something like that since people are used to quitting sbt with exit.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
dwijnand: Yeah, I agree to some way to confirm first.  But I feel that double-tabbing is normally used to avoid filling a user's screen, rather than performance reasons.  But that's not to say I'm against using tab for it.

I'm happy to leave things as is, until I can figure out what might've prompted my instinct.

eatkins: Yeah. That's true that double tab is usually used to avoid spamming the screen. I feel like this is nevertheless similar given that in both situations the program delays displaying the completions until the user presses tab a second time.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: I tried to leave NetworkClient compatible with `sbt -client` simply because I could. I figured it would be nice if `sbt -client` would continue working. That being said, I wouldn't mind getting rid of it. For non-GraalVM fallback, I think we could pretty easily make a fat jar. We could also probably use the trick that mill uses to embed a bash or batch script in the fat jar to make it executable.

eatkins:
I got reboot working:
[asciicast](https://asciinema.org/a/Ht6zFvhxLVmEKYkTXWJFXeK4n)
As part of getting reboot working, I was able to in most cases ensure that at most server is started. Even when two clients are started at almost the same time and each create an sbt process.

I updated the graalvm download links (and of course there are some server changes needed as well).

dwijnand:
In the sbtc shell `tests/testOnly -- -z ExplicitNonNullaryApplyOver -Doverwrite=true` works as intended, while from the terminal it seems to steal the `-D`:

```
$ sbtc -Doverwrite=true ~tests/testOnly -- -z ExplicitNonNullaryApplyOver
> ~tests/testOnly -- -z ExplicitNonNullaryApplyOver
```

Btw, I've upgraded to your recent changes, and I saw a usage of `--` that was previous broken work now.

eatkins:
I'm a little bit confused by this report because in the shell you have a tail `-Doverwrite=true` but at the command line there is a leading `-Doverwrite=true`. The leading `-Doverwrite=true` is interpreted as a system property and is not considered as part of the command. If you put the `-Doverwrite=true` after the `~tests/testOnly` it would not be treated as a system property and would be sent as part of the command. I'm assuming that `-Doverwrite=true` is a scalatest flag and not intended to be a system property. If that is the case, then maybe we need something like `-A-Doverwrite` to indicate that `-Doverwrite` is an argument not a system property similar to how the bash script has `-J-Xmx2g`.

It does raise a question about system properties though. Right now, if sbtc detects system properties, it forwards them to the forked sbt process if no sbt is running. If sbt is already running, however, it drops them. The two options that come to mind are that we could either warn the user that we're dropping them so they at least know what's going on or after the client connects and before it runs any commands, it could send a json message to the server that sets the system properties.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: https://github.com/sbt/sbt/pull/5620/commits/d353e012aa4a311ba9c8f7357551d6178419ddd7

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: https://github.com/sbt/sbt/pull/5620/commits/314d9a295d8b96a480243466ec11d68450ebb53e

eatkins: I didn't want to put sbtc in the instructions since users may alias the command. What do you think?

eatkins: Plus I arbitrarily picked sbtc by fiat. Are we sure that's what we want to call it?

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: renamed the prompt flag https://github.com/sbt/sbt/pull/5620/commits/44a98186b707dccc3018fac95dcaef76b237a834

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eed3si9n: Thin client v1 was shipped as a command, so it's in main-command subproject. If v2 lives as a separate JAR, we could pull it out as a subproject is what I was thinking.

eatkins: Got it. It would be pretty easy to move NetworkClient.scala. I also think we could move `Server` and `ServerConnection` to main which I think makes a bit more sense.

dwijnand:
> I'm a little bit confused by this report

Yeah, I copy-pasted incorrectly.  My original was `sbtc ~tests/testOnly -- -z ExplicitNonNullaryApplyOver -Doverwrite=true` which was printing without the `-D` at all.

> If that is the case, then maybe we need something like `-A-Doverwrite`

IMO, the convention which works well is stop interpreting all the args after you hit a `--` and just accumulate them.  Should get to ScalaTest, then, no?

eatkins:
It turns out that the issue is that the graalvm executable strips out any command flag prefixed with `-D` and adds it to the system properties. The only way I can think of to fix that issue is to add a wrapper script for `sbtc` that adds a prefix to any `-D` arguments before invoking `sbtc` and then `sbtc` could strip the prefixes.

dwijnand:
Does specifying `-H:-ParseRuntimeOptions` as a image-generation flag not work or have adverse effects?

https://github.com/oracle/graal/issues/779#issuecomment-436274256

eatkins:
Nice find. I looked up system property in the graal issues myself but missed that. I added the flag in https://github.com/sbt/sbt/pull/5620/commits/569d39c2039038d2b9a72e27562fc55331665c16 and updated the binaries.

dwijnand:
:+1: fixed!  Thank you.

```
$ sbtc tests/testOnly -- -z Override -Doverwrite=true
> tests/testOnly -- -z Override -Doverwrite=true
```

eed3si9n (approved):
I'd be happy to merge this as-is, and work on the pending issues like "what should we name different component" etc in a more targeted follow-up PRs.

eatkins:
No turning back now ;)

tmnd1991:
Thanks @eatkins for the amazing work ❤️

xuwei-k:
https://github.com/sbt/sbt/issues/5927 🤔

main/src/main/scala/sbt/Main.scala
```diff
@@ -880,10 +932,14 @@ object BuiltinCommands {
     val session = Load.initialSession(structure, eval, s0)
     SessionSettings.checkSession(session, s2)
     val s3 = addCacheStoreFactoryFactory(Project.setProject(session, structure, s2))
-    val s4 = LintUnused.lintUnusedFunc(s3)
-    CheckBuildSources.init(s4)
+    val s4 = setupGlobalFileTreeRepository(s3)
```
mkurz, line 935: @eatkins This line, which closes the current GlobalFileTreeRepository and creates a new one **after** `onLoad` is executed, is causing problems. Please see #6190, thanks!

PR #5631, opened 2020-06-17
reduce BuildServerReporter verbosity

adpi2:
The BSP protocol has introduced a 'reset' field in the publishDiagnostic notification that makes the cumulation of errors at client side possible.

Consequently it is not needed to maintain the `problemByFiles` map in the reporter.

The reporter first reset the diagnostic of each file then publish the problems one at a time.

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
@@ -12,9 +12,7 @@ import sbt.internal.bsp._
 import sbt.internal.inc.ManagedLoggedReporter
 import sbt.internal.util.ManagedLogger
 import xsbti.compile.CompileAnalysis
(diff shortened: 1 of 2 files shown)


main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
```diff
@@ -28,110 +26,90 @@ class BuildServerReporter(
...
+      val range = for {
+        sl <- pos.startLine.toOption
+        sc <- pos.startColumn.toOption
+        el <- pos.endLine.toOption
+        ec <- pos.endColumn.toOption
+      } yield Range(Position(sl.toLong - 1, sc.toLong), Position(el.toLong - 1, ec.toLong))
+
       Diagnostic(
-        range,
+        range.getOrElse(Range(Position(line, pointer), Position(line, pointer + 1))),
```
eed3si9n: `for` desugars into flatMap, and it's going to be expensive when there are nested ones like this. I think the original pattern match was clear enough.

Issue #5638, opened 2020-06-20
Allow customization of remote cache artifacts


iRevive:
## Motivation

There are several scenarios when `cached-compile` and `cached-test` artifacts are not enough:
1) Custom configurations: IntegrationTest, additional build folders, etc.
2) Generated twirl templates.

In the current implementation, a new artifact can be added to the publish configuration: 
```scala
pushRemoteCacheConfiguration / packagedArtifacts ++= Seq(
  // custom artifacts here
)
```

Unfortunately, SBT does not know how to extract custom artifacts during `pullRemoteCache`. There is a hardcoded logic for `cached-compile` and `cached-test`.

## Possible solution

Introduce a new configuration that keeps a link between artifact and an extraction directory. Choose extraction logic based on the configuration.

iRevive:
@eed3si9n should we close this ticket?

eed3si9n:
Sounds good to me.

jtjeferreira:
> Custom configurations: IntegrationTest

Is there an easy way to add IntegrationTest artifacts to the remote cache? I had a quick look to the PR, but it was not obvious...

iRevive:
@jtjeferreira hey, could you try these settings?

```scala
import sbt.internal.RemoteCache

lazy val example = project.in(file("./example"))
  ...
  .settings(
    inConfig(IntegrationTest)(
      RemoteCache.configCacheSettings(RemoteCache.testArtifact(IntegrationTest, "cached-integration-test"))
    )
  )

**Comment by jtjeferreira** [comment]
@iRevive I tried that but still dont see any "cached-integration-test" jars being pushed.

Also in the sbt shell if I do `show core/pushRemoteCacheConfiguration/remoteCacheArtifacts`
```
> show core/pushRemoteCacheConfiguration/remoteCacheArtifacts
[info] * CompileRemoteCacheArtifact(Artifact(core, jar, jar, Some(cached-compile), Vector(), None, Map(), None, false), TaskKey(This / Select(ConfigKey(compile)) / This / packageCache), /Users/joao/git/platform/modules/core/target/scala-2.13/classes, /Users/joao/git/platform/modules/core/target/scala-2.13/zinc/inc_compile_2.13.zip)
[...]

PR #5639, opened 2020-06-20
Allow customization of remote cache artifacts

iRevive:
That's my attempt to implement #5638.

I'm new to the SBT code base, therefore some approaches can be reconsidered.  

Currently, I have one concern regarding `RemoteCacheArtifact` ADT.  `final case class` can lead to bin compact problems in the future. Should I use `sealed abstract class` instead?

### DIFF
--- a/main/src/main/contraband-scala/sbt/internal/remotecache/CompileRemoteCacheArtifact.scala
+++ b/main/src/main/contraband-scala/sbt/internal/remotecache/CompileRemoteCacheArtifact.scala
@@ -1,0 +1,44 @@
+/**
+ * This code is generated using [[https://www.scala-sbt.org/contraband/ sbt-contraband]].
+ */
+
+// DO NOT EDIT MANUALLY
+package sbt.internal.remotecache
+final class CompileRemoteCacheArtifact private (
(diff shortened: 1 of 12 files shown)


eed3si9n:
> Currently, I have one concern regarding RemoteCacheArtifact ADT. final case class can lead to bin compact problems in the future. Should I use sealed abstract class instead?

We created Contraband to generate pseudo case classes - https://www.scala-sbt.org/contraband/. See https://github.com/sbt/sbt/blob/develop/protocol/src/main/contraband/jsonrpc.contra and https://github.com/sbt/sbt/blob/develop/protocol/src/main/contraband/bsp.contra for some examples.

iRevive:
I switched to contraband generated classes.

eed3si9n:
@iRevive Could you add a [scripted test](https://www.scala-sbt.org/1.x/docs/Testing-sbt-plugins.html) under https://github.com/sbt/sbt/tree/develop/sbt/src/sbt-test/actions demonstrating the extension capability?

iRevive:
I've added a scripted test

eed3si9n (approved):
Thanks!

PR #5642, opened 2020-06-22
Use real paths for zinc roots

eatkins:
Not using the real paths can cause issues for certain features if sbt is
run out of a symlinked directory, which is the case in scripted tests
for example.

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -33,6 +33,7 @@ import xsbti.compile.CompilerCache
 import scala.annotation.tailrec
 import scala.concurrent.ExecutionContext
 import scala.util.control.NonFatal
+import xsbti.AppProvider
 
 /** This class is the entry point for sbt. */
 final class xMain extends xsbti.AppMain {
@@ -40,6 +41,16 @@ final class xMain extends xsbti.AppMain {
     new XMainConfiguration().run("xMain", configuration)
 }
 private[sbt] object xMain {
+  private[sbt] def dealiasBaseDirectory(config: xsbti.AppConfiguration): xsbti.AppConfiguration = {
+    val dealiasedBase = config.baseDirectory.getCanonicalFile
+    if (config.baseDirectory == dealiasedBase) config
+    else
+      new xsbti.AppConfiguration {
+        override def arguments: Array[String] = config.arguments()
+        override val baseDirectory: File = dealiasedBase
+        override def provider: AppProvider = config.provider()
+      }
+  }
   private[sbt] def run(configuration: xsbti.AppConfiguration): xsbti.MainResult =
     try {
       import BasicCommandStrings.{ DashClient, DashDashClient, runEarly }
@@ -54,16 +65,16 @@ private[sbt] object xMain {
       val isClient: String => Boolean = cmd => (cmd == DashClient) || (cmd == DashDashClient)
       val isBsp: String => Boolean = cmd => (cmd == "-bsp") || (cmd == "--bsp")
       if (userCommands.exists(isBsp)) {
-        BspClient.run(configuration)
+        BspClient.run(dealiasBaseDirectory(configuration))
       } else {
         Terminal.withStreams {
           if (clientModByEnv || userCommands.exists(isClient)) {
             val args = userCommands.toList.filterNot(isClient)
-            NetworkClient.run(configuration, args)
+            NetworkClient.run(dealiasBaseDirectory(configuration), args)
             Exit(0)
           } else {
             val state = StandardMain.initialState(
-              configuration,
+              dealiasBaseDirectory(configuration),
               Seq(defaults, early),
               runEarly(DefaultsCommand) :: runEarly(InitCommand) :: BootCommand :: Nil
             )
@@ -84,7 +95,7 @@ private[sbt] object ScriptMain {
   private[sbt] def run(configuration: xsbti.AppConfiguration): xsbti.MainResult = {
     import BasicCommandStrings.runEarly
     val state = StandardMain.initialState(
-      configuration,
+      xMain.dealiasBaseDirectory(configuration),
       BuiltinCommands.ScriptCommands,
       runEarly(Level.Error.toString) :: Script.Name :: Nil
     )
@@ -99,7 +110,7 @@ final class ConsoleMain extends xsbti.AppMain {
 private[sbt] object ConsoleMain {
   private[sbt] def run(configuration: xsbti.AppConfiguration): xsbti.MainResult = {
     val state = StandardMain.initialState(
-      configuration,
+      xMain.dealiasBaseDirectory(configuration),
       BuiltinCommands.ConsoleCommands,
       IvyConsole.Name :: Nil
     )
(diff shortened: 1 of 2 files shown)


eatkins:
The specific issue this solves is that when trying to write custom external hooks, we get unresolved paths back from zinc, but sbt generates paths with the symbolic links already resolved. Calling `toRealPath()` is pretty expensive on an individual file so it is much better to do it at the directory level than having to do it for every file in external hooks.

eed3si9n:
Related to this is the ugly hack I have in MappedFileConverter here to handle `/var/folders/` specifically - https://github.com/sbt/zinc/blob/468609cc2a9fa220dcbeb853e077799eaf05c456/internal/zinc-core/src/main/scala/sbt/internal/inc/MappedVirtualFile.scala#L46-L55

I am really not sure if we should keep de-aliasing all symbolic links so try to figure out a way of keeping them as is.

eatkins:
You are sort of damned if you do, damned if you don't with dealiasing symlinks. I believe that not dealiasing causes more subtle and unexpected problems than dealiasing them. For example, it took me hours to figure out that the reason my custom external hooks weren't working was because symlinks were not resolved.

Isn't this less ugly than the hack in zinc that only addresses temp files on osx?

eed3si9n:
> Isn't this less ugly than the hack in zinc that only addresses temp files on osx?

To clarify, I _hope_ we can remove my Zinc hack code and replace it with something less embarrassing.

@dwijnand sent PR in this area (https://github.com/sbt/zinc/pull/802) too.
In general, if we can dealias every path that Zinc would see upfront I'd be happy with that. Not sure how expensive that cost is, but in the grand scheme of things (like relative to the cost of calling Farm Hash) hopefully not too much?

eatkins:
I looked further into what is causing the inconsistency. The problem is that `baseDirectory` is dealiased during project loading: https://github.com/sbt/sbt/blob/77a99d34dfe30602b1b5f8267f773f7a75332786/main/src/main/scala/sbt/internal/Load.scala#L48
but AppConfiguration.baseDirectory can be a symlink. The dealiasing seems to go all the way back to 2687836ca5df48ad3cfb8f7053c97d3be451aa42.

An alternative change that would probably resolve this issue would simply be
```patch
         val app = appConfiguration.value 
-        val base = app.baseDirectory
+        val base = app.baseDirectory.getCanonicalFile
         val boot = app.provider.scalaProvider.launcher.bootDirectory
```
I'd be fine with just this change because it would make the base of `rootPaths` be consistent with the `baseDirectory` key which was what caused me grief. I threw in the other `toRealPath` calls to be safe. I am pretty sure that this change would remove the need for the zinc hack.

eed3si9n:
> val base = app.baseDirectory.getCanonicalFile

That sounds promising, since hopefully the dealiasing effect would carry forward.

eatkins:
> That sounds promising, since hopefully the dealiasing effect would carry forward.

https://github.com/sbt/sbt/pull/5642/commits/63fa46be3174aacafafe6da4566e8db5f497a945 introduces a better solution to try and ensure that this specific bug does not crop up again. I still left https://github.com/sbt/sbt/pull/5642/commits/8ca220242f6b4958343fdc91624b1f0073cf147e in place though it shouldn't be needed after that commit.

dwijnand:
Following symlinks isn't something you can just "solve at the root", though, right?

Even if you get one path's canonical/real path, it could be used to build to a path that is still not canonical.  Talk about "damned if you do, damned if you don't"...

eatkins:
Yeah, I should clarify that this would just make it so that the base directory path is resolved. Most of the paths generated by sbt are relative to the base path. Obviously all bets are off as soon as any additional path components are added. The real problem is simply that we compare paths by name without resolving symlinks. Even if we were really principled within sbt and always resolved paths wherever we compared them (which could potentially be a nontrivial performance hit), downstream users could run into problems if they forget to resolve them.

In a weird way, it's kind of a blessing that the default temp directory on osx is a symlink or else we probably wouldn't even notice.

main/src/main/scala/sbt/Main.scala
```diff
@@ -33,13 +33,24 @@ import xsbti.compile.CompilerCache
...
+  private[sbt] def dealiasBaseDirectory(config: xsbti.AppConfiguration): xsbti.AppConfiguration = {
+    val dealiasedBase = config.baseDirectory.getCanonicalFile
+    if (config.baseDirectory == dealiasedBase) config
+    else
+      new xsbti.AppConfiguration {
+        override def arguments: Array[String] = config.arguments()
+        override val baseDirectory: File = dealiasedBase
+        override def provider: AppProvider = config.provider()
+      }
+  }
```
eed3si9n, line 53: One concern I have about dealiasing the base directory is the discrepancies it might introduce between external editors. For example between VS Code and sbt server there's a frequent communication about which files are opened, if the files is under sbt's jurisdiction, and if the file was saved by the user etc. If VS Code sends an aliased path like `/tmp/hello/Hello.scala`, but we internally track that as `/private/tmp/hello/Hello.scala` wouldn't that cause a mismatch? /cc @adpi2

eed3si9n: Or more simpler example might be terminals supporting click-to-open when it finds a full path in compiler errors.

eatkins: That is a valid concern. I don't know how editors / lsp clients work with files. The heart of this change is that, in my opinion, `sbt.Keys.baseDirectory.value` and `appConfiguration.value.baseDirectory` should always return the same file. That is the motivation for this change. My belief is that prior to the introduction of `rootPath`, pretty much all paths generated by sbt were canonicalized with respect to the base directory due to the dealising in `Load`. Just now, I audited all of the calls to `appConfiguration.value` with sbt itself and there was only one place I could find where `baseDirectory` was used and not canonicalized: https://github.com/sbt/sbt/blob/e323f1f713251968b3d4ecda8a7f0ed52593713a/main/src/main/scala/sbt/internal/DefaultBackgroundJobService.scala#L504

Not making this change seems more likely to me to cause problems because right now we have an inconsistency and downstream users are forced to either canonicalize `AppConfiguration.baseDirectory` or potentially have inconsistency when comparing sbt generated paths that use `baseDirectory`.

eatkins: Note also that this only matters if the baseDirectory of the project is actually a symlink. I bet that outside of scripted and sbt tests, that is quite rare.

eatkins: As a side not, it would be kind of nice if sbt only used paths relative to the base directory. Then there would be no need for `rootPaths` to convert paths to absolute paths. It would also make logging of path names nicer. Global directories like the coursier cache could be handled with symlinks in the project directory. There are obviously some potential issues with this approach (and would probably need to be an sbt 2.0 feature) but it's interesting to think about.

eed3si9n: > My belief is that prior to the introduction of `rootPath`, pretty much all paths generated by sbt were canonicalized with respect to the base directory due to the dealising in `Load`.

ok. So this PR corrects the path handling back to status quo (1.3.x). I guess we'll cross the editor + symlink issue when it materializes then.

eed3si9n (approved):

main/src/main/scala/sbt/Main.scala
adpi2: > If VS Code sends an aliased path like /tmp/hello/Hello.scala, but we internally track that as /private/tmp/hello/Hello.scala wouldn't that cause a mismatch?

What I can tell about BSP is that file paths are going from sbt to client and not the other way around. For instance sbt can tell that the source of target foo are in the /foo/src/main. Then Metals will watch that directory for a change, but when a change happens it will send a compile request of target foo and it will not mention the file.

The only hypothetical mismatch I can think of would be that VSCode works on the aliased path but sbt reports compile error in the dealiased path.

PR #5666, opened 2020-07-03
Cache compiled map in build structure

eatkins:
We have two core commands, lintUnused and `~`, that have relied on generating the `CompiledMap` from the project structure. Generating that map takes about 4 seconds on my computer in the sbt project which manifests as a four second pause in the build loading (due to the lintUnused slowdown) or starting a watch with `~`. To eliminate this pause, we can cache the CompiledMap that is generated during project loading in the BuildStructure. The only obvious drawback I can think to this approach is that we will be keeping that CompiledMap in memory but I am somewhat skeptical that that will substantially increase the memory overhead of sbt.


dwijnand (dismissed):
Looks reasonable to me

eatkins:
I rebased this branch and need a re-approval.

dwijnand (approved):

eatkins:
Thanks @dwijnand!

PR #5671, opened 2020-07-07
Upgrade to jline3

eatkins:
This PR upgrades sbt to jline 3 primarily so that we can get better tab completions in the sbt shell:
[asciicast](https://asciinema.org/a/2uQZpOSgLke0RYvKMfPF1coEC)

We still need jline2 for the console task to work with most scala versions. This made the PR a bit bigger in terms of net lines than I would have liked because I had to leave in all of the old jline2 related stuff. JLine3 also has some strange quirks around creating terminals and the input streams that it consumed that required some unusual workarounds.

In addition to the improved tab completions. I also fixed a few other small ui issues. Now when a client connects to a loading server, they will see all of the log messages since the server started which is especially useful if the server failed to load due to a misconfigured build. I also improved the supershell logging so that it works more reliably across the full range of scenarios that I test.


eed3si9n:
Nice! Thanks for the contribution @eatkins.

internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
```diff
@@ -8,24 +8,98 @@
...
+  }
   def createReader(
+      historyPath: Option[File],
+      parser: Parser[_],
+      terminal: Terminal,
+      prompt: Prompt = Prompt.Running,
+  ): LineReader = {
+    val term = JLine3(terminal)
+    // We may want to consider insourcing LineReader.java from jline. We don't otherwise
+    // directly need jline3 for sbt.
```
eed3si9n, line 77: I'm a bit confused by this statement. LineReader.java (https://github.com/jline/jline3/blob/3b0d5d52fd2c31fb92ad8aaa2f11a22aceab890a/reader/src/main/java/org/jline/reader/LineReader.java) is an empty interface. Don't we need JLine's implementation for the reader?

eatkins: Yeah. I meant the implementation class https://github.com/jline/jline3/blob/3b0d5d52fd2c31fb92ad8aaa2f11a22aceab890a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java. The idea would just be to make a partial fork of jline3 that only provides the minimal subset of functionality from `LineReaderImpl` that we need to make line reading and tab completion work but without using all of other jline 3 stuff that we neither want or need. I'm not sure how excited I am about actually doing that though.

eed3si9n: What's the diff for `show sbtProj/Compile/externalDependencyClasspath` pre and post this PR?

eatkins:
~~The console task is slightly broken at the moment. Left arrow doesn't work correctly probably due to a jline2 vs jline3 interaction that I need to track down.~~

Fixed in https://github.com/sbt/sbt/pull/5671/commits/1749e45bb11af1b71f1df52b9844e5186560b06f

internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
eatkins:
```patch
+ Attributed($MAVEN/org/jline/jline/3.15.0/jline-3.15.0.jar)
```

eed3si9n (approved):

PR #5674, opened 2020-07-10
Specify full java path in BSP connection details

### LINKED ISSUE #5623: BSP connection details should fully qualify Java binary
## steps

Start sbt 1.4.0-M1

## problem

The `"argv"` field in the `.bsp/sbt.json` file contains an unqualified `java` command
```json
❯ cat .bsp/sbt.json | jq
{
  "name": "sbt",
  "version": "1.4.0-M1",
  "bspVersion": "2.0.0-M5",
  "languages": [
    "scala"
  ],
  "argv": [
    "java",
    "-Xms100m",
    "-Xmx100m",
    "-classpath",
    "/usr/local/Cellar/sbt/1.3.3/libexec/bin/sbt-launch.jar",
    "xsbt.boot.Boot",
    "-bsp"
  ]
}
```
## expectation

I expected the Java command to point to the fully qualified `java` binary, for example
```diff
  "argv": [
-    "java",
+    "/Users/lgeirsson/.jabba/jdk/graalvm@20.1.0/Contents/Home/bin/java",
    "-Xms100m",
```

## notes

The problem with using an unqualified `java` command is that GUI applications like VS Code and IntelliJ don't have access to the same environment as the user has in the terminal. Users should ideally be able to declare somewhere what Java version they should use.

One option would be to leave it up to the build client (Metals, IntelliJ) to infer the fully qualified `java` binary when `argv` starts with `"java"`. This might be the best option but we should update the BSP spec in that case to document that clients are allowed to interpret the `argv` command differently.

adpi2:
Fixes #5623 

Use System properties to add java path and classpath to BSP connection details.

### DIFF
--- a/main-command/src/main/scala/sbt/internal/server/Server.scala
+++ b/main-command/src/main/scala/sbt/internal/server/Server.scala
@@ -10,7 +10,6 @@ package internal
 package server
 
 import java.io.{ File, IOException }
-import java.lang.management.ManagementFactory
 import java.net.{ InetAddress, ServerSocket, Socket, SocketTimeoutException }
 import java.util.concurrent.atomic.{ AtomicBoolean, AtomicReference }
 import java.nio.file.attribute.{ AclEntry, AclEntryPermission, AclEntryType, UserPrincipal }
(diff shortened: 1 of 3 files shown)


adpi2:
I have no idea why the tests are failing because I was not able to reproduce locally.

eed3si9n (dismissed):

eed3si9n:
I'm restarting AppVeyor job.

eed3si9n (approved):

PR #5677, opened 2020-07-10
Exit BspClient after server socket is closed

### LINKED ISSUE #5651: BSP server does not respond when connected sbt server is closed
## steps

1. start an sbt shell on a project
2. start a BSP server for the same project (eg via opening it in IntelliJ)
3. the BSP server will connect to the existing shell
4. close the shell

## problem
1. BSP server throws an error:
```
Exception in thread "Thread-0" java.lang.ArrayIndexOutOfBoundsException: arraycopy: length -1 is negative
at java.base/java.lang.System.arraycopy(Native Method)
at java.base/java.io.BufferedOutputStream.write(BufferedOutputStream.java:129)
at java.base/java.io.PrintStream.write(PrintStream.java:559)
at sbt.internal.client.BspClient.sbt$internal$client$BspClient$$transferTo(BspClient.scala:26)
at sbt.internal.client.BspClient$$anon$1.run(BspClient.scala:34)
```
2. running a request on the server hangs indefinitely with no response.

## expectation

Either:
* BSP server exits, so client can start a fresh process, or
* BSP server responds with error on request, or
* BSP server start its own fresh sbt process

adpi2:
Fixes #5651

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/BspClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/BspClient.scala
@@ -18,24 +18,47 @@ import scala.sys.process.Process
 import scala.util.control.NonFatal
 
 class BspClient private (sbtServer: Socket) {
+  private val lock = new AnyRef
+  private var terminated = false
 
-  private def transferTo(input: InputStream, output: OutputStream): Unit = {
-    val buffer = Array.ofDim[Byte](1024)
-    while (true) {
-      val size = input.read(buffer)
-      output.write(buffer, 0, size)
-      output.flush()
+  private def transferTo(input: InputStream, output: OutputStream): Thread = {
+    val thread = new Thread {
+      override def run(): Unit = {
+        val buffer = Array.ofDim[Byte](1024)
+        try {
+          while (!terminated) {
+            val size = input.read(buffer)
+            if (size == -1) {
+              terminated = true
+            } else {
+              output.write(buffer, 0, size)
+              output.flush()
+            }
+          }
+          input.close()
+          output.close()
+        } catch {
+          case NonFatal(_) => ()
+        } finally {
+          lock.synchronized {
+            terminated = true
+            lock.notify()
+          }
+        }
+      }
     }
+    thread.setDaemon(true)
+    thread
   }
 
   private def run(): Exit = {
     try {
-      val redirection = new Thread {
-        override def run(): Unit = transferTo(sbtServer.getInputStream, System.out)
-      }
+      transferTo(sbtServer.getInputStream, System.out).start()
+      transferTo(System.in, sbtServer.getOutputStream).start()
 
-      redirection.start()
-      transferTo(System.in, sbtServer.getOutputStream)
+      lock.synchronized {
+        while (!terminated) lock.wait()
+      }
 
       Exit(0)
     } catch {


adpi2:
Fixes #5651

main-command/src/main/scala/sbt/internal/client/BspClient.scala
```diff
@@ -18,24 +18,39 @@ import scala.sys.process.Process
...
         override def run(): Unit = transferTo(sbtServer.getInputStream, System.out)
       }
 
-      redirection.start()
-      transferTo(System.in, sbtServer.getOutputStream)
+      val stdToSocket = new Thread {
+        override def run(): Unit = transferTo(System.in, sbtServer.getOutputStream)
+      }
+
+      socketToStd.start()
```
eatkins: I think socketToStd and stdToSocket both look like the should be daemon threads. It maybe doesn't matter but is good practice.

main-command/src/main/scala/sbt/internal/client/BspClient.scala
```diff
@@ -18,24 +18,39 @@ import scala.sys.process.Process
...
     while (true) {
       val size = input.read(buffer)
-      output.write(buffer, 0, size)
-      output.flush()
+      if (size == -1) synchronized {
+        output.close()
+        terminated = true
+        notify()
+      } else {
+        output.write(buffer, 0, size)
```
eatkins: All of the io calls (`input.read`, `output.write` and `output.flush`) could throw an `IOException`. For example, if you `kill -9` the server, I think `input.read` will throw an exception rather than return `-1`. My guess is that you want to exit if that happens so I'd recommend adding a `try` / `catch` block that sets terminated to true and notifies the main thread if an unexpected error occurs.

eatkins (changes requested):

main-command/src/main/scala/sbt/internal/client/BspClient.scala
eatkins: (this by the way is why I think scala's ignoring checked exceptions from java was a horrible design decision).

main-command/src/main/scala/sbt/internal/client/BspClient.scala
adpi2: Thank you @eatkins for this review. It's the first time I hear about daemon process.

main-command/src/main/scala/sbt/internal/client/BspClient.scala
adpi2: From what I have experienced the socket would return `-1` even if the server process is abruptly killed. But that may depends on the implementation which itself depends on the OS and JVM. So I think you are right, it is a lot safer to catch and terminate the process in case of exception.

main-command/src/main/scala/sbt/internal/client/BspClient.scala
```diff
@@ -18,24 +18,45 @@ import scala.sys.process.Process
...
+            if (size == -1) {
+              terminated = true
+            } else {
+              output.write(buffer, 0, size)
+              output.flush()
+            }
+          }
+          input.close()
+          output.close()
+        } finally {
```
eatkins, line 42: If you don’t catch the exception, it will print a stacktrace. If you don’t care about that, this is fine. I’m not sure where the stacktrace would be logged with the bsp client so it’s up to you to make that call.

adpi2: The stacktrace will be printed to `System.err` which is not good because the parent process will try to interpret it as a json-RPC message...

adpi2:
@eatkins do you have any more comment on this PR? If not, could you please validate it?

eatkins (approved):
LGTM

PR #5686, opened 2020-07-19
Remove lint warnings in sbt build

eatkins:
There were a number of unused key lint warnings when loading the sbt
build. In the case of `fork in compile` and `crossVersion in update`, it
wasn't clear that these were actually used, so I removed those settings.
The others seemed to be used so I just added them to the exclude list.

Side note: ++= does not work with excludeLintKeys which is why I used +=
for the excludes.


eatkins:
This will fail until we publish 1.4.0-M2 and bump the sbt version in project/build.properties.

eed3si9n (approved):

eed3si9n:
Considering this is build linting, it seems pointless to wait for flaky AppVeyor build(s), so I'm merging without it.

PR #5690, opened 2020-07-19
Use jline2 for raw mode

eatkins:
Using the JLine3 implementation of raw mode breaks the scala console for
the console channel.


eed3si9n:
I wonder if this is Scala version dependent since Scala 2.13.2 uses JLine 3 (https://www.scala-lang.org/news/2.13.2).

eatkins:
Yes and no. It turns out that console does not work with sbtc and scala 2.13 which I had never tested because I never bother with scala 2.13. That's a separate issue that I'll have to resolve. The scala 2.13 console does work with or without this change when entered from the server console.

eatkins:
relevant: https://contributors.scala-lang.org/t/scala-2-13-console-is-incompatible-with-sbt-client/4421

eed3si9n (approved):

PR #5691, opened 2020-07-19
Avoid call to Path.toUri

eatkins:
I noticed in flamegraphs that a nontrivial amount of time was being
spent in the toUri function when I ran a no-op compile in a project with
5000 source files. My assumption is that since these paths will always
be regular files, we can just construct the URI manually to avoid the
performance hit.


eatkins:
Hmm...this approach doesn't work on windows. Closing for now.

PR #5694, opened 2020-07-20
Support scala 2.13 console in thin client

eatkins:
In order to make the console task work with scala 2.13 and the thin
client, we need to provide a way for the scala repl to use an sbt
provided jline3 terminal instead of the default terminal typically built
by the repl. We also need to put jline 3 higher up in the classloading
hierarchy to ensure that two versions of jline 3 are not loaded (which
makes it impossible to share the sbt terminal with the scala terminal).


eatkins:
This requires https://github.com/scala/scala/pull/9128 to work.

eatkins:
Closing in favor of https://github.com/sbt/sbt/pull/5711

PR #5701, opened 2020-07-27
Improve sbt build load time by 25%

eatkins:
The sbt project load made a number of relatively inefficient
transformations of scala collecitons. I went through and found the slow
parts during project loading and made my best attempt at fixing them.
The most significant changes I made were in places using IMap. An IMap
is more or less a wrapper around an immutable Map. It can be much faster
to construct an IMap by creating a java mutable hashmap, wrapping it a
scala Map that delegates to the underlying java hashmap (with a copy on
write if the map is modified) and constructing the IMap from the wrapped
map. It was also in many cases to parallelize some transformations
wherever the order didn't matter.

After applying all of these changes, I found that loading the sbt
project took generally between 8.5 and 9 seconds on my laptop. With
1.3.13, it hovered around 11.5 seconds. I saw a similar speedup in zinc.
The biggest specific improvement was that generating the compiled map
dropped from between 3.5-4 seconds to pretty consistently being around
1.5 seconds.

### DIFF
--- a/internal/util-collection/src/main/scala-2.12/sbt/internal/util/WrappedMap.scala
+++ b/internal/util-collection/src/main/scala-2.12/sbt/internal/util/WrappedMap.scala
@@ -1,0 +1,17 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.internal.util
+
+import scala.collection.JavaConverters._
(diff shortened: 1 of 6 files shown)


eatkins:
This also makes the `set` command much faster. I haven't actually benchmarked it but the difference is quite noticeable in the sbt project, here is the before:
[asciicast](https://asciinema.org/a/mXvTTTG0XLZ5GT1nUauvsQH85)
and after:
[asciicast](https://asciinema.org/a/9tVPCWheGjDZwy0rKRAl3Se2X)

eatkins:
The improvement in load time was less significant for akka. It dropped from about 11.5 seconds with 1.3.13 to 10.5 with 1.4.0-SNAPSHOT.

eed3si9n (approved):
Nice!

PR #5707, opened 2020-07-28
Add installNativeExecutable task

eatkins:
This allows a user to install the native thin client into a particular
directory (e.g. /usr/local/bin).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -206,8 +206,26 @@ lazy val sbtRoot: Project = (project in file("."))
       .single("sbtOn")((state, dir) => s"sbtProj/test:runMain sbt.RunFromSourceMain $dir" :: state),
     mimaSettings,
     mimaPreviousArtifacts := Set.empty,
-    genExecutable := (sbtClientProj / genExecutable).evaluated,
-    genNativeExecutable := (sbtClientProj / genNativeExecutable).value,
+    buildThinClient := (sbtClientProj / buildThinClient).evaluated,
+    buildNativeThinClient := (sbtClientProj / buildNativeThinClient).value,
+    installNativeThinClient := {
+      // nativeInstallDirectory can be set globally or in a gitignored local file
+      val dir = nativeInstallDirectory.?.value
+      val target = Def.spaceDelimited("").parsed.headOption match {
+        case Some(p) => file(p).toPath
+        case _ =>
+          dir match {
+            case Some(d) => d / "sbtc"
+            case _ =>
+              val msg = "Expected input parameter <path>: installNativeExecutable /usr/local/bin"
+              throw new IllegalStateException(msg)
+          }
+      }
+      val base = baseDirectory.value.toPath
+      val exec = (sbtClientProj / buildNativeThinClient).value
+      streams.value.log.info(s"installing thin client ${base.relativize(exec)} to ${target}")
+      Files.copy(exec, target, java.nio.file.StandardCopyOption.REPLACE_EXISTING)
+    }
   )
 
 // This is used to configure an sbt-launcher for this version of sbt.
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -1046,6 +1060,8 @@ val graalNativeImageCommand = taskKey[String]("The native image command")
 val graalNativeImageOptions = settingKey[Seq[String]]("The native image options")
 val graalNativeImageClass = settingKey[String]("The class for the native image")
 val genNativeExecutable = taskKey[JPath]("Generate a native executable")
+val nativeInstallDirectory = taskKey[JPath]("The install directory for the native executable")
+val installNativeExecutable = inputKey[JPath]("Install the native executable")
```
eed3si9n:
```suggestion
val installNativeExecutable = inputKey[JPath]("Install the native executable, for example by calling it as $ sbt installNativeExecutable ~/bin/")
```

build.sbt
```diff
@@ -208,6 +208,20 @@ lazy val sbtRoot: Project = (project in file("."))
...
     genNativeExecutable := (sbtClientProj / genNativeExecutable).value,
+    installNativeExecutable := {
+      // nativeInstallDirectory can be set globally or in a gitignored local file
+      val dir = nativeInstallDirectory.?.value
+      val target = Def.spaceDelimited("").parsed.headOption match {
+        case Some(p) => file(p).toPath
+        case _ =>
+          dir match {
+            case Some(d) => d / "sbtc"
+            case _       => throw new IllegalStateException("No native install directory specified")
```
eed3si9n:
```suggestion
            case _       => throw new IllegalStateException("Expected input parameter <path>: installNativeExecutable ~/bin")
```

build.sbt
```diff
@@ -1046,6 +1060,8 @@ val graalNativeImageCommand = taskKey[String]("The native image command")
 val graalNativeImageOptions = settingKey[Seq[String]]("The native image options")
 val graalNativeImageClass = settingKey[String]("The class for the native image")
 val genNativeExecutable = taskKey[JPath]("Generate a native executable")
+val nativeInstallDirectory = taskKey[JPath]("The install directory for the native executable")
```
eed3si9n, line 1079:
```suggestion
val nativeInstallDirectory = settingKey[JPath]("The install directory for the native executable")
```

eed3si9n (commented):
Some minor suggestions.

build.sbt
eed3si9n: Also in terms of the name would `installThinClient` make more sense?

eatkins: Yeah. There are a number of tasks that are badly named like `genExecutable` that date back to when I was building out the POC. I'll go through and rename those tasks to make their behavior more obvious at a glance.

build.sbt
eatkins: Sadly `Paths.get` doesn't work with `~/bin` but otherwise I like the suggestion with an absolute path.

build.sbt
eatkins: I used `taskKey` rather than `settingKey` because that allows the user to leave it unset. I didn't want to set a default because it's very particular where the user might want to install it which is why I made it an input task. But I wanted the user to be able to set it globally or in a local sbt file, hence the need to use Task.

eed3si9n: `nativeInstallDirectory.?.value` should work with setting too, I think.

eatkins: It doesn't on my computer. I tried changing it to a setting key and the task crashed or the build failed to load. I forget which.

eed3si9n: I just realized this is on our `build.sbt`, not `Defaults.scala`. I'm fine with whatever detail if it's mostly for us.

eed3si9n (approved):

build.sbt
eatkins: Makes sense. I could have made that more clear. Thanks for merging.

PR #5708, opened 2020-07-28
Add fast path for parsing commands

eatkins:
It can easily take 2ms or more to parse a command depending on state's
combined parser. There are some commands that sbt requires to work that
we can handle in microseconds instead of milliseconds by special casing
them.

After this change, I saw the performance of
https://github.com/eatkins/scala-build-watch-performance improve by
a consistent 4-5ms in the 3 source file example which was a drop from
120ms to 115ms. While not necessarily earth shattering, this difference
could theoretically be much worse in other projects that have a lot of
plugins and custom tasks/commands. I think it's worth the modest
maintenance cost.

### DIFF
--- a/main/src/main/scala/sbt/MainLoop.scala
+++ b/main/src/main/scala/sbt/MainLoop.scala
@@ -23,6 +23,7 @@ import sbt.util.Logger
 
 import scala.annotation.tailrec
 import scala.util.control.NonFatal
+import sbt.internal.FastTrackCommands
 
 object MainLoop {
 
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/internal/InternalCommands.scala
```diff
@@ -0,0 +1,52 @@
...
+package internal
+
+import BasicCommandStrings._
+import BasicCommands._
+import BuiltinCommands.{ setTerminalCommand, shell, waitCmd }
+import ContinuousCommands._
+
+import sbt.internal.util.complete.Parser
+
+private[sbt] object InternalCommands {
```
eed3si9n:
```suggestion
/** This is used to speed up command parsing. */
private[sbt] object FastTrackedCommands {
```

eed3si9n (commented):
overall LGTM, with a minor name change suggestion.

main/src/main/scala/sbt/internal/InternalCommands.scala
eatkins: 👍

eed3si9n (dismissed):

PR #5711, opened 2020-07-28
Support scala 2.13 and dotty console in thin client

eatkins:
This PR makes the scala 2.13 and dotty console tasks work with the thin client. It works by globally overriding the value returned by TerminalBuilder.build. This overriding is achieved by manipulating the classloader used by the scala instance to have a top loader that uses a forked version of jline that provides an api that allows us to control the global override.

In addition to making the console tasks work, it also makes some assorted ux/performance improvements.


eed3si9n:
@eatkins What's the status on this PR given recent changes?

eatkins:
I'm hoping to have an updated version of this that resolves a few minor but noticeable ui issues by the end of the day.

eatkins:
This is good to go now. I am sure I will have to make some bug fixes but I can use sbt comfortably with all of these changes. Most importantly, with the supershell changes at the end of this PR, I am not seeing any memory leaks in sbt which means it can run pretty much indefinitely without needing a restart which is key for the thin client.

eatkins:
The total time for travis CI ran 5 minutes faster for this PR than any recent successful build :) Appveyor is also down to 12 minutes from 20 not too long ago.

eed3si9n (approved):

main/src/main/scala/sbt/internal/TaskProgress.scala
```diff
@@ -9,99 +9,100 @@ package sbt
...
+    Executors.newSingleThreadExecutor(r => new Thread(r, "sbt-task-progress-report-thread"))
+  override def close(): Unit = {
+    Option(reportLoop.get).foreach(_.close())
+    pending.forEach(f => Util.ignoreResult(f.cancel(true)))
+    pending.clear()
+    scheduler.shutdownNow()
+    executor.shutdownNow()
+    if (!executor.awaitTermination(1, TimeUnit.SECONDS) ||
+        !scheduler.awaitTermination(1, TimeUnit.SECONDS)) {
+      throw new TimeoutException
```
rtyley, line 59: This `TimeoutException` is getting thrown quite a lot when I try to deploy builds on Heroku, I'm wondering if there's anything that can be done to avoid that? This is with sbt 1.6.0-M1, Java 11, on the small open-source project https://github.com/guardian/repo-genesis with PR https://github.com/guardian/repo-genesis/pull/6.

Here's [a full build log](https://gist.github.com/rtyley/30195e8bc1fcaf01597b9c134ea47f65) showing the exception - of the **7** most recent deploy attempts I've made, only **1** of them has passed - the other 6 have all failed with this `TimeoutException` causing deployment to abort...

image

I'm using the early sbt milestone [1.6.0-M1](https://github.com/sbt/sbt/releases/tag/v1.6.0-M1) for the fix in https://github.com/sbt/sbt/pull/6649 for the `ClassCastException` problems with https://github.com/sbt/sbt/issues/6592 and Java 11 - but this timeout is maybe too aggressive for Heroku?

_Update_ - see extended timeout in https://github.com/sbt/sbt/pull/6729

eed3si9n: Interesting. Let's extend this to 30s and/or make it configurable.

rtyley: Out of curiosity, what would happen if there was no timeout at all? I don't really know what's going on, but is it _necessary_ to block `close()` to wait for the threads to complete - can something bad happen if the `scheduler` & `executor` are not terminated after `close()` has been called?

eed3si9n: I guess generally we don't want any threads or memory to leak (or would like to know if some leak is happening), esp when as sbt server a session could be up for days as daemon.

PR #5712, opened 2020-07-29
Reduce overhead of some logging apis

eatkins:
I've found that log4j is actually a significant source of cpu performance and memory overhead so I'm working on improving our interactions with logging. These changes aren't directly log4j related but they came up while I was tracking down the performance issues.

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/ManagedLogger.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/ManagedLogger.scala
@@ -7,12 +7,12 @@
 
 package sbt.internal.util
 
-import sbt.util._
 import org.apache.logging.log4j.{ Logger => XLogger }
 import org.apache.logging.log4j.message.ObjectMessage
-import sjsonnew.JsonFormat
-import scala.reflect.runtime.universe.TypeTag
 import sbt.internal.util.codec.JsonProtocol._
+import sbt.util._
+import scala.reflect.runtime.universe.TypeTag
+import sjsonnew.JsonFormat
 
 /**
  * Delegates log events to the associated LogExchange.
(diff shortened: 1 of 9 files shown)


internal/util-logging/src/main/scala/sbt/internal/util/StringTypeTag.scala
```diff
@@ -15,6 +16,8 @@ final case class StringTypeTag[A](key: String) {
 }
 
 object StringTypeTag {
+  implicit def default[A]: StringTypeTag[A] = macro appmacro.StringTypeTag.impl[A]
```
eed3si9n:
```suggestion
  /** Uses macro to generate StringTypeTag. */
  implicit def fast[A]: StringTypeTag[A] = macro appmacro.StringTypeTag.impl[A]
```

internal/util-logging/src/main/scala/sbt/internal/util/StringTypeTag.scala
```diff
@@ -15,6 +16,8 @@ final case class StringTypeTag[A](key: String) {
...
+  @deprecated("Prefer macro generated StringTypeTag", "1.4.0")
```
eed3si9n, line 22:
```suggestion
  @deprecated("Use macro-powered StringTypeTag.fast instead", "1.4.0")
```

internal/util-logging/src/main/scala/sbt/internal/util/StringTypeTag.scala
```diff
@@ -38,6 +41,7 @@ object StringTypeTag {
       retry(3)
     }
 
+  @deprecated("Prefer macro generated StringTypeTag", "1.4.0")
```
eed3si9n, line 46:
```suggestion
  @deprecated("Use macro-powered StringTypeTag.fast instead", "1.4.0")
```

internal/util-logging/src/main/scala/sbt/internal/util/ManagedLogger.scala
```diff
@@ -35,29 +35,53 @@ class ManagedLogger(
...
+    LogExchange.registerStringCodec[A](s, tt)
+  }
+  def registerStringCodec[A: ShowLines: StringTypeTag]: Unit = {
     LogExchange.registerStringCodec[A]
   }
 
-  final def debugEvent[A: JsonFormat: TypeTag](event: => A): Unit = logEvent(Level.Debug, event)
-  final def infoEvent[A: JsonFormat: TypeTag](event: => A): Unit = logEvent(Level.Info, event)
-  final def warnEvent[A: JsonFormat: TypeTag](event: => A): Unit = logEvent(Level.Warn, event)
-  final def errorEvent[A: JsonFormat: TypeTag](event: => A): Unit = logEvent(Level.Error, event)
```
eed3si9n, line 56: These are public methods since `TaskStreams#log` returns `ManagedLogger`. My intent was to allow event logging from tasks.

eed3si9n (changes requested):

internal/util-logging/src/main/scala/sbt/internal/util/ManagedLogger.scala
eed3si9n: For logging it's probably safe to do Mima check even for `sbt.internal.*`.

internal/util-logging/src/main/scala/sbt/internal/util/StringTypeTag.scala
eatkins: 👍

internal/util-logging/src/main/scala/sbt/internal/util/ManagedLogger.scala
eatkins: I left these public apis because zinc uses them so they can't be deleted until we publish a new util and update zinc. Having them around isn't that bad because they are awkward to call since I removed the implicits and forced the arguments to be called explicitly while preserving binary compatibility. Downstream users also generally wouldn't have to make source changes and will automatically pick up the new api on their next plugin release (or when they upgrade their build).

eatkins: The zinc usage is why I had to make the changes to BuildServerReporter below.

eatkins:
Closing in favor of #5731

PR #5713, opened 2020-07-30
Clear appenders after task evaluation

eatkins:
sbt was leaking ConsoleAppenders with each task run. When I looked at a
heap dump after I ran compile 30 times in a project (see below) with
about 20 subprojects, I found the dump filled with leaked
ConsoleAppenders. This can be fixed by removing the loggers created
during task evaluation from log4j. I was concerned about removing
loggers that were still in use so I refactored LogManager so that it now
takes a LogExchange parameter. EvaluateTask can pass in a fresh
LogExchange and clear out all of the loggers when it's done.

Clearing the appenders is not only a memory leak but a performance
improvement as well. Adding a logger requires updating a concurrent hash
map which gets slower as the map size increases. I noticed in a
flamegraph that creating appenders was taking 50% of the total time
which was what motivated this change.

My project was a local clone of the repro project in
https://github.com/sbt/sbt/issues/5508. I was testing sbt overhead by
timing how long it took to run 30 iterations of no-op compile. There is
also a memory leak in TaskProgress that I will resolve separately that
causes degraded performance. With the fix for TaskProgress, the console
appender leaks cause the time for no-op compilation to increase per
iteration by about 100-200ms every 30 iterations. After applying this
change on top, there is no longer a noticeable increase in as the number
of iterations increase. This is very important since if we're planning
to push people towards the thin client, we really need to avoid resource
leaks in sbt.

This is not scientific, but for a sense of the potential impact of these
changes, with 1.3.13, the first 30 no-op compiles took about 390ms per
compile, the second 30 took 460 and the third took about 570. After all
of the fixes were applied in 1.4.0, the first 30 iteration took about
350ms per compile, the second took around 270. Then I ran 300
iterations which almost surely would have caused an OOM and taken
forever with 1.3.13 and the average was 250 (yay, JIT!).

Fixes #4779 https://github.com/sbt/sbt/pull/4779

### DIFF
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -20,7 +20,7 @@ import sbt.internal._
 import sbt.internal.util._
 import sbt.librarymanagement.{ Resolver, UpdateReport }
 import sbt.std.Transform.DummyTaskMap
-import sbt.util.{ Logger, Show }
+import sbt.util.{ LogExchange, Logger, Show }
 
 import scala.Console.RED
 import scala.concurrent.duration.Duration
@@ -383,11 +383,13 @@ object EvaluateTask {
     streams(ScopedKey(Project.fillTaskAxis(key).scope, Keys.streams.key))
 
   def withStreams[T](structure: BuildStructure, state: State)(f: Streams => T): T = {
-    val str = std.Streams.closeable(structure.streams(state))
+    val logExchange = new LogExchange
+    val str = std.Streams.closeable(structure.streams(state.put(logExchangeAttribute, logExchange)))
     try {
       f(str)
     } finally {
       str.close()
+      logExchange.close()
     }
   }
 
--- a/internal/util-logging/src/main/scala/sbt/util/LogExchange.scala
+++ b/internal/util-logging/src/main/scala/sbt/util/LogExchange.scala
@@ -17,22 +17,30 @@ import scala.collection.JavaConverters._
 import scala.collection.concurrent
 import scala.reflect.runtime.universe.TypeTag
 import sjsonnew.JsonFormat
+import java.util.concurrent.ConcurrentHashMap
 
 // http://logging.apache.org/log4j/2.x/manual/customconfig.html
 // https://logging.apache.org/log4j/2.x/log4j-core/apidocs/index.html
 
-sealed abstract class LogExchange {
+sealed class LogExchange extends AutoCloseable {
   private[sbt] lazy val context: LoggerContext = init()
   private[sbt] lazy val builtInStringCodecs: Unit = initStringCodecs()
   private[sbt] lazy val asyncStdout: AsyncAppender = buildAsyncStdout
   private[sbt] val jsonCodecs: concurrent.Map[String, JsonFormat[_]] = concurrent.TrieMap()
   private[sbt] val stringCodecs: concurrent.Map[String, ShowLines[_]] = concurrent.TrieMap()
+  private[this] def getCtx = XLogManager.getContext(false) match { case x: LoggerContext => x }
+  private[this] val loggers = ConcurrentHashMap.newKeySet[String]
+  override def close(): Unit = {
+    val config = getCtx.getConfiguration
+    loggers.forEach(config.removeLogger)
+    loggers.clear()
+  }
 
   def logger(name: String): ManagedLogger = logger(name, None, None)
   def logger(name: String, channelName: Option[String], execId: Option[String]): ManagedLogger = {
     val _ = context
     val codecs = builtInStringCodecs
-    val ctx = XLogManager.getContext(false) match { case x: LoggerContext => x }
+    val ctx = getCtx
     val config = ctx.getConfiguration
     val loggerConfig = LoggerConfig.createLogger(
       false,
@@ -47,8 +55,8 @@ sealed abstract class LogExchange {
       null
     )
     config.addLogger(name, loggerConfig)
-    ctx.updateLoggers
     val logger = ctx.getLogger(name)
+    loggers.add(name)
     new ManagedLogger(name, channelName, execId, logger, Some(Terminal.get))
   }
   def unbindLoggerAppenders(loggerName: String): Unit = {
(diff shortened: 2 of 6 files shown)


eatkins:
I've also noticed that sbt gets really sluggish in the sbt project after about 90 or so iterations of `~publishLocalBin` on my computer. I'm hoping this (and the unpushed TaskProgress changes) will fix things so that sbt actually gets faster (through JIT) rather than slower over time.

main/src/main/scala/sbt/EvaluateTask.scala
```diff
@@ -383,11 +383,13 @@ object EvaluateTask {
...
 
   def withStreams[T](structure: BuildStructure, state: State)(f: Streams => T): T = {
-    val str = std.Streams.closeable(structure.streams(state))
+    val logExchange = LogExchange
+    val str = std.Streams.closeable(structure.streams(state.put(logExchangeAttribute, logExchange)))
     try {
       f(str)
     } finally {
       str.close()
+      logExchange.close()
```
eed3si9n: I get that stream should be closed but I am not sure tying task evaluation, which is a part of act command, with `LogExchange` makes sense. Aren't there loggers that run the whole time?

eatkins: First off, I just noticed that I accidentally pushed a bug due to an incomplete cherry pick. This should be:
```patch
+    val logExchange = new LogExchange
+    val str = std.Streams.closeable(structure.streams(state.put(logExchangeAttribute, logExchange)))
```

> Aren't there loggers that run the whole time?

That's a great question. When I was trying to figure out how to safely remove appenders, I updated `LogExchange.logger` to print out the name of the logger it was creating. I then ran compile in https://github.com/jtjeferreira/sbt-multi-module-sample and found that it created about 6000 loggers for each run of compile. All of them had the format `streams-$i`. This made me believe that, in practice, what we needed to do was to clear out the loggers used during task evaluation. 

In my first attempt, I actually cached and removed all loggers from the `LogExchange` object but doing that to a global `LogExchange` seemed like a bad idea. That being said, in practice, it seemed to work fine but I didn't test it extensively. Instead, I just tried to figure out the boundaries where task evaluation began and task evaluation ended. My thinking is that it would be safe to drop all of the loggers created during task evaluation once the task evaluation completes. Part of what I was worried about was dynamic tasks. I was afraid that if a dynamic task completed it would clear _all_ the loggers not just the logger's it created. I convinced myself that this approach would work but it is a bit tricky to reason about. Another option I considered was to put similar logic in MainLoop rather than EvaluateTask.

My key assumption is that loggers are essentially not created outside of `GlobalLogging` and `Streams`. Users could theoretically be creating appenders outside of `streams.value.log` probably by using the `logManager` setting. My intuition is that that isn't really done. Even if it was, this change would only cause problems if they are caching the logger outside of the task evaluation and then try to use the cached logger outside of the task evaluation.

The `LogExchange` singleton is not cleared out between runs. GlobalLogging is initialized using the `LogExchange` singleton so it persists indefinitely for the life of the sbt process.

eed3si9n: I guess as a general direction, I am not comfortable with the idea of having multiple `LogExchange`s, since IMO it should be a thin wrapper around log4j. If we could keep the knowledge of stream loggers within the Stream that would be ideal, but if not maybe we could use State transition as a boundary? `MainLoop.scala` for example creates a new logger each time it processes before processing new command. If cleanup of arbitrarily created loggers are needed, I feel like that's the place to do it.

eatkins: I took a second pass at this and I'm convinced that a global `LogExchange` cannot work without leaking memory. The log4j `LoggerContext` data structure is append only (I verified this in the log4j source code) so even if we clean up the loggers as I've done in this PR, we will still have a memory leak in the `LoggerContext`. Due to this behavior, these changes are more of a bandaid than a fix. They reduce the amount that we leak but don't stop us from leaking.

The only way that I can see to prevent the resource leak is to create a new `LoggerContext` before every command evaluation and allow that context to go out of scope once command evaluation completes. I think your intuition is right that the place to manage the `LoggerContext` is in MainLoop. Prior to command evaluation, we'd add the new `LoggerContext` to the state via an attribute. When the command evaluation finishes, we'd remove the attribute so that the `LoggerContext` goes out of scope. I have this about 70% implemented, but I wanted to comment here first because I'll probably open a new PR since the implementation will be very different from this one.

eed3si9n: The tricky part might be allowing the `log.info(...)` to write into the appenders asynchronously, but wait for all appenders to finish writing before we close transient appenders out.

eatkins: How committed are we to using log4j? I am having a lot of problems getting rid of this leak because it seems nearly impossible to avoid creating global state internal to log4j. This ends up being a major problem for global logging because I cant' find anyway to clear out the ephemeral loggers without also blowing away global logging. I tried created different log4j configurations for both but somewhere underneath the hood they seem to end up sharing state. The result is that I cannot eliminate the memory leak without breaking global logging. I'm not going so far as to say that it's impossible to get working with log4j but I suspect that I can implement the subset of log4j that sbt needs faster than I can fix this bug.

Am I understanding the logging system correctly that the main reason for log4j is that the appender feature allows calls to, e.g. `log.info(msg)`, to be non-blocking and to fan out to multiple output sources? If so, that seems fairly straightforward to implement. If I were to go down this route, I'd refactor things in such a way that log4j could still be used as an optional implementation. It looks like log4j is only in the public api via `Keys.extraLoggers`.

eed3si9n: > How committed are we to using log4j?

I am not sure. In sbt 0.13 days, we had our own logging and I think it was performance bottleneck. One of the ideas we used to discuss internally was using a ring buffer to write logs in quickly, and let IO thread pick it up, based on Martin Thompson's talk. I chose log4j 2 because it adopted Martin Thompson's ring buffer library Disruptor, and at that time it was one of the fastest logger. At least it had the potential to be fast if configured correctly, which I may or may not have. Another benefit of sbt using log4j was that now plugins can use slf4j, and they would know that there's a binding for it.

If task logs are not good fit with log4j, I am open to changing it to something else.

eatkins:
Closing in favor of #5731

PR #5714, opened 2020-07-31
Reduce scripted ci duration

eatkins:
These two commits are intended to help reduce the total time in CI. The first commit disables a number of pending tests in the dependency-management project. Pending tests can really hurt runtime if they cause the scripted sbt process to exit because that forces a cold restart of sbt, which is slow. The second commit replaces the `InstrumentScripted` auto plugin with a local sbt file, `instrument.sbt`. Local sbt files can be processed significantly faster than an auto plugin can be compiled. Compiling `InstrumentScripted` can take a few seconds while processing the sbt file seems to take less than a few hundred milliseconds.


eed3si9n (approved):
Thanks @eatkins!

Issue #5710, opened 2020-07-28
versionScheme setting


eed3si9n:
Ref https://github.com/sbt/sbt/issues/3562

This is a proposal to consider a new `Option[String]` setting called `versionScheme`. Initially the supported values are:

| versionScheme  | description |
| ---------------- | ------------- |
| `Some("early-semver")` |  Early Semantic Versioning that would keep binary compatibility across patch updates within 0.Y.z (for instance 0.13.0 and 0.13.2). Once it goes 1.0.0, it follows the regular Semantic Versioning where 1.1.0 is bincompat with 1.0.0. |
| `Some("semver-spec")`  | [Semantic Versioning](https://semver.org/) where all 0.y.z are treated as initial development (no bincompat guarantees)  |
| `Some("pvp")`                  | [Haskell Package Versioning Policy](https://pvp.haskell.org/) where X.Y are treated as major version |
| `Some("strict")`                | Requires exact match of version |

This setting can serve 2 purposes:
- (Library authors) Drive compatibility check in https://github.com/scalacenter/sbt-compatibility
- (Library users) Parse from POM file to drive more accurate eviction report

julienrf:
>     * `Some("semver-zero")`

It is stricter than `semver`, so what do you think about `semver+`? Otherwise, I keep thinking that `semver-scala` is fine (despite it’s not yet what is used by the standard library)

Also, what do you think of using an enumeration (case objects) rather than strings?

eed3si9n:
As per the name I'd go with whatever that comes out from the discussion in https://github.com/coursier/versions/issues/7

> Also, what do you think of using an enumeration (case objects) rather than strings?

If this ends up in POM, then I think we should use String. Let's say 1.4.0 has three schemes and 1.5.0 adds another one. 1.4.0 still needs to be able to parse the scheme.

Another motivation would be to let plugins like sbt-compatibility define keys before sbt 1.4.0. Using a data structure introduces complication if we want to do that.

eed3si9n:
Here's the PR for the automatic eviction report - https://github.com/sbt/librarymanagement/pull/339

This needs to be paired with including versionScheme into POM, and carrying the information along to sbt.

eed3si9n:
Including versionScheme into POM is trivial, but somehow the information seems to get lost when it goes through Coursier and/or lm-coursier - https://github.com/sbt/sbt/pull/5724

eed3si9n:
[...]

PR #5724, opened 2020-08-05
versionScheme setting + better eviction warning 

eed3si9n:
Ref https://github.com/sbt/sbt/issues/5710

## what does this add?

This adds `versionScheme` setting, which would be used by library authors as:

```scala
ThisBuild / versionScheme := Some("semver-spec")
```

## what does it do?

When `versionScheme` set, it is included into `pom.xml` or `ivy.xml`, and is picked up on the other side as an extra attribute of ModuleID. 

```xml
<?xml version='1.0' encoding='UTF-8'?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0">
    <modelVersion>4.0.0</modelVersion>
    <groupId>com.example</groupId>
    <artifactId>semver-spec-test_2.12</artifactId>
    <packaging>jar</packaging>
    <description>semver-spec-test</description>
    <version>1.0.0</version>
    <name>semver-spec-test</name>
    <organization>
        <name>com.example</name>
    </organization>
    <properties>
        <info.versionScheme>semver-spec</info.versionScheme>
    </properties>
    ....
</project>
```

On Ivy:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<ivy-module version="2.0" xmlns:e="http://ant.apache.org/ivy/extra">
  <info organisation="com.example" module="semver-spec-test_2.12" revision="1.0.0" status="release" publication="20200805162126" 
    e:info.versionScheme="semver-spec">
    <description>semver-spec-test</description>
  </info>
  ....
</ivy-module>
```

## how is that useful?

Having the intent of versioning scheme is useful because plugins can use that to take specific actions. Scala Center is working on [sbt-version-policy](https://github.com/scalacenter/sbt-compatibility) that would automatically enforce the version scheme.

Another cool thing for general ecosystem (library users), is that [`evicted` task would now use this information](https://github.com/sbt/librarymanagement/pull/339) when available instead of arbitrarily guessing that all Scala libraries follow [PVP](https://pvp.haskell.org/). This should reduce the false positives associated with SemVer-versioned libraries that are becoming more standard.


eed3si9n:
Publishing seems to be working:

```xml
<?xml version='1.0' encoding='UTF-8'?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0">
    <modelVersion>4.0.0</modelVersion>
    <groupId>com.example</groupId>
    <artifactId>semver-spec-test_2.12</artifactId>
    <packaging>jar</packaging>
    <description>semver-spec-test</description>
    <version>1.0.0</version>
    <name>semver-spec-test</name>
    <organization>
        <name>com.example</name>
    </organization>
    <properties>
        <info.versionScheme>semver-spec</info.versionScheme>
    </properties>
    <dependencies>
        <dependency>
            <groupId>org.scala-lang</groupId>
            <artifactId>scala-library</artifactId>
            <version>2.12.11</version>
        </dependency>
    </dependencies>
    <repositories>
        ....
    </repositories>
</project>
```

scripted `dependency-management/evicted-semver-spec` works using Ivy, but it doesn't when using Coursier.

PR #5729, opened 2020-08-08
Use java caffeine library rather than scalacache

eatkins:
sbt depends on scalacache (which hasn't been updated in about a year)
and we really don't need the functionality provided by scalacache. In
fact, the java api is somewhat easier to work with for our use case. The
motivation is that scalacache uses slf4j for logging which meant that it
was implicitly loading log4j. This caused some noisy logs during
shutdown when the previously unused cache was initialized just to be
cleaned up.

This commit also upgrades caffeine and moving forward we can always
upgrade caffeine (and potentially shade it) without any conflict with
the scalacache version.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -873,7 +873,7 @@ lazy val mainProj = (project in file("main"))
         sys.error(s"PluginCross.scala does not match up with the scalaVersion $sv")
     },
     libraryDependencies ++=
-      (Seq(scalaXml, launcherInterface, scalaCacheCaffeine, lmCoursierShaded) ++ log4jModules),
+      (Seq(scalaXml, launcherInterface, caffeine, lmCoursierShaded) ++ log4jModules),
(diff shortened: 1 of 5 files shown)


eed3si9n (approved):
nice!

lespea:
Probably not needed but just wanted to point out there is a nice scala wrapper for caffeine: https://github.com/blemale/scaffeine/

eatkins:
Thanks for the pointer! It is definitely true that the Caffeine api leaves a bit to be desired when called from scala so I can see the use case for the wrapper library. As a general principle though, it's best for sbt (or really any scala library) to avoid scala library dependencies because those dependencies may not be updated at the same rate that new scala versions are published. Of course if there is a real need for the scala functionality then it makes sense to bring in the library. For a wrapper though, it's better for us to in-source the wrapper or just make our own scala idiomatic wrapper implementation.

PR #4604, opened 2019-04-05
move super shell rendering to logger

eed3si9n:
Fixes #4583
Ref https://github.com/sbt/util/pull/196

PR #5731, opened 2020-08-08
Replaces Apache Log4j with our own logger to avoid Appender leakage

### LINKED ISSUE #4480: sbt logging level doesn't affect test task

## steps
1. run `sbt new scala/scala-seed.g8` (or pick any sbt project)
2. run `sbt -warn`

## problem

sbt log output contains:
```
[info] HelloSpec:
[info] The Hello object
[info] - should say hello
```

## expectation
sbt log output produces only warn level logs, as it's provided in running parameter.

## notes
I've tried to provide logging level using `logLevel in Scope.Global := Level.Warn`, but it only affects main sbt execution flow, not test task.
I also checked the code and found `TestLogger#make` has the following code 
```
      LogExchange.bindLoggerAppenders(newLog.name, buffs map { x =>
        (x, Level.Debug)
      })
```
which looks like it's always produce an appender with debug level.

If this is expected behavior, there should be a way to override the logging level for test task.

This is becoming more critical, when i add `"com.novocode" % "junit-interface" % "0.11" % Test` and junit test. Junit runner produces debug log to each test case, so that completely spoils logs, like:
```
[debug] Test run started
[debug] Test sample.Test.testSample started
[debug] Test sample.Test.testSample finished, took 1.005 sec
[debug] Test run finished: 0 failed, 0 ignored, 1 total, 1.011s
```
See https://github.com/ikryvorotenko/sbt-junit-sample as example

sbt version: *1.2.6*

### LINKED ISSUE #4773: sbt leaks metaspace when reloading an empty project

eatkins:
Prior to these changes, sbt was leaking large amounts of memory via
log4j appenders. sbt has an unusual use case for log4j because it
creates many ephemeral loggers while also having a global logger that is
supposed to work for the duration of the sbt session. There is a lot of
shared global state in log4j and properly cleaning up the ephemeral task
appenders would break global logging. This commit fixes the behavior by
introducing an alternate logging implementation. Users can still use the
old log4j logging implementation but it will be off by default. The
internal implementation is very simple: it just blocks the current
thread and writes to all of the appenders. Nevertheless, I found the
performance to be roughly identical to that of log4j in my sample
project. As an experiment, I did the appending on a thread pool and got
a significant performance improvement but I'll defer that to a later PR
since parallel io is harder to reason about.

Background: I was testing sbt performance in
https://github.com/jtjeferreira/sbt-multi-module-sample and noticed that
performance rapidly degraded after I ran compile a few times. I took a
heap dump and it became obvious that sbt was leaking console appenders.
Further investigation revealed that all of the leaking appenders in the
project were coming from task streams. This made me think that the fix
would be to track what loggers were created during task evaluation and
clear them out when task evaluation completed. That almost worked except
that log4j has an internal append only data structure containing logger
names. Since we create unique logger names for each run, that internal
data structure grew without bound. It looked like this could be worked
around by creating a new log4j Configuration (where that data structure
was stored) but while creating new configurations with each task runs
did fix the leak, it also broke global logging, which was using a
different configuration. At this point, I decided to write an alternate
implementation of the appender api where I could be sure that the
appenders were cleaned up without breaking global logging.

Implementation: I made ConsoleAppender a trait and made it no longer
extends log4j AbstractAppender. To do this, I had to remove the one
log4j specific method, append(LogEvent). ConsoleAppender now has a
method toLog4J that, in most cases, will return a log4j Appender that is
almost identical to the Appenders that we previously used. To manage
the loggers created during task evaluation, I introduce a new class,
LoggerContext. The LoggerContext determines which logging backend to use
and keeps track of what appenders and loggers have been created. We can
create a fresh LoggerContext before each task evaluation and clear it
out, cleaning up all of its resources after task evaluation concludes.

In order to make this work, there were many places where we need to
either pass in a LoggerContext or create a new one. The main magic is
happening in the `next(State)` method in Main. This is where we create a
new LoggerContext prior to command evaluation and clean it up after the
evaluation completes.

Users can toggle log4j using the new useLog4J key. They also can set the
system property, sbt.log.uselog4j. The global logger will use the sbt
internal implementation unless the system property is set.

There are a fairly significant number of mima issues since I changed the
type of ConsoleAppender. All of the mima changes were in the
sbt.internal package so I think this should be ok.

Effects: the memory leaks are gone. I successfully ran 5000 no-op
compiles in the sbt-multi-module-sample above with no degradation of
performace. There was a noticeable degradation after 30 no-op compiles
before.

During the refactor, I had to work on TestLogger and in doing so I also
fixed #4480.

This also should fix #4773

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/ConsoleAppender.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/ConsoleAppender.scala
@@ -13,11 +13,14 @@ import java.nio.channels.ClosedChannelException
 import java.util.concurrent.atomic.{ AtomicBoolean, AtomicInteger }
 
 import org.apache.logging.log4j.core.appender.AbstractAppender
-import org.apache.logging.log4j.core.{ LogEvent => XLogEvent }
+import org.apache.logging.log4j.core.{ Appender => XAppender, LogEvent => XLogEvent }
 import org.apache.logging.log4j.message.{ Message, ObjectMessage, ReusableObjectMessage }
 import org.apache.logging.log4j.{ Level => XLevel }
 import sbt.internal.util.ConsoleAppender._
 import sbt.util._
+import org.apache.logging.log4j.core.AbstractLogEvent
+import org.apache.logging.log4j.message.StringFormatterMessageFactory
+import java.util.concurrent.atomic.AtomicReference
 
 object ConsoleLogger {
   // These are provided so other modules do not break immediately.
@@ -80,7 +83,7 @@ class ConsoleLogger private[ConsoleLogger] (
     suppressedMessage: SuppressedTraceContext => Option[String]
 ) extends BasicLogger {
 
-  private[sbt] val appender: ConsoleAppender =
+  private[sbt] val appender: Appender =
     ConsoleAppender(generateName(), out, ansiCodesSupported, useFormat, suppressedMessage)
 
   override def control(event: ControlEvent.Value, message: => String): Unit =
@@ -117,12 +120,12 @@ object ConsoleAppender {
   private[this] val showProgressHolder: AtomicBoolean = new AtomicBoolean(false)
   def setShowProgress(b: Boolean): Unit = showProgressHolder.set(b)
   def showProgress: Boolean = showProgressHolder.get
-  private[ConsoleAppender] trait Properties {
+  private[sbt] trait Properties {
     def isAnsiSupported: Boolean
     def isColorEnabled: Boolean
     def out: ConsoleOut
   }
-  object Properties {
+  private[sbt] object Properties {
     def from(terminal: Terminal): Properties = new Properties {
       override def isAnsiSupported: Boolean = terminal.isAnsiSupported
       override def isColorEnabled: Boolean = terminal.isColorEnabled
@@ -187,7 +190,7 @@ object ConsoleAppender {
    *
    * @return A new `ConsoleAppender` that writes to standard output.
    */
-  def apply(): ConsoleAppender = apply(ConsoleOut.systemOut)
+  def apply(): Appender = apply(ConsoleOut.systemOut)
 
   /**
    * A new `ConsoleAppender` that appends log message to `out`.
@@ -195,7 +198,7 @@ object ConsoleAppender {
    * @param out Where to write messages.
    * @return A new `ConsoleAppender`.
    */
-  def apply(out: PrintStream): ConsoleAppender = apply(ConsoleOut.printStreamOut(out))
+  def apply(out: PrintStream): Appender = apply(ConsoleOut.printStreamOut(out))
 
   /**
    * A new `ConsoleAppender` that appends log messages to `out`.
@@ -203,7 +206,7 @@ object ConsoleAppender {
    * @param out Where to write messages.
    * @return A new `ConsoleAppender`.
    */
-  def apply(out: PrintWriter): ConsoleAppender = apply(ConsoleOut.printWriterOut(out))
+  def apply(out: PrintWriter): Appender = apply(ConsoleOut.printWriterOut(out))
 
   /**
    * A new `ConsoleAppender` that writes to `out`.
@@ -211,7 +214,7 @@ object ConsoleAppender {
    * @param out Where to write messages.
    * @return A new `ConsoleAppender that writes to `out`.
    */
-  def apply(out: ConsoleOut): ConsoleAppender = apply(generateName(), out)
+  def apply(out: ConsoleOut): Appender = apply(generateName(), out)
 
   /**
    * A new `ConsoleAppender` identified by `name`, and that writes to standard output.
@@ -219,7 +222,7 @@ object ConsoleAppender {
    * @param name An identifier for the `ConsoleAppender`.
    * @return A new `ConsoleAppender` that writes to standard output.
    */
-  def apply(name: String): ConsoleAppender = apply(name, ConsoleOut.systemOut)
+  def apply(name: String): Appender = apply(name, ConsoleOut.systemOut)
 
   /**
    * A new `ConsoleAppender` identified by `name`, and that writes to `out`.
@@ -228,7 +231,7 @@ object ConsoleAppender {
    * @param out Where to write messages.
    * @return A new `ConsoleAppender` that writes to `out`.
    */
-  def apply(name: String, out: ConsoleOut): ConsoleAppender = apply(name, out, formatEnabledInEnv)
+  def apply(name: String, out: ConsoleOut): Appender = apply(name, out, formatEnabledInEnv)
 
   /**
    * A new `ConsoleAppender` identified by `name`, and that writes to `out`.
@@ -242,7 +245,7 @@ object ConsoleAppender {
       name: String,
       out: ConsoleOut,
       suppressedMessage: SuppressedTraceContext => Option[String]
-  ): ConsoleAppender =
+  ): Appender =
     apply(name, out, formatEnabledInEnv, formatEnabledInEnv, suppressedMessage)
 
   /**
@@ -253,7 +256,7 @@ object ConsoleAppender {
    * @param useFormat `true` to enable format (color, bold, etc.), `false` to remove formatting.
    * @return A new `ConsoleAppender` that writes to `out`.
    */
-  def apply(name: String, out: ConsoleOut, useFormat: Boolean): ConsoleAppender =
+  def apply(name: String, out: ConsoleOut, useFormat: Boolean): Appender =
     apply(name, out, useFormat || formatEnabledInEnv, useFormat, noSuppressedMessage)
 
   /**
@@ -263,10 +266,8 @@ object ConsoleAppender {
    * @param terminal  The terminal to which this appender corresponds
    * @return A new `ConsoleAppender` that writes to `out`.
    */
-  def apply(name: String, terminal: Terminal): ConsoleAppender = {
-    val appender = new ConsoleAppender(name, Properties.from(terminal), noSuppressedMessage)
-    appender.start()
-    appender
+  def apply(name: String, terminal: Terminal): Appender = {
+    new ConsoleAppender(name, Properties.from(terminal), noSuppressedMessage)
   }
 
   /**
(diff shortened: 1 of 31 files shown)


eatkins:
Replaces https://github.com/sbt/sbt/pull/5713

internal/util-logging/src/main/scala/sbt/util/LoggerContext.scala
```diff
@@ -0,0 +1,189 @@
...
+  }
+  private[util] class LoggerContextImpl extends LoggerContext {
+    private class Log extends MiniLogger {
+      private val consoleAppenders: java.util.Vector[(ConsoleAppender, Level.Value)] =
+        new java.util.Vector
+      def log(level: Level.Value, message: => String): Unit =
+        consoleAppenders.forEach {
+          case (a, l) =>
+            if (level.compare(l) >= 0) a.appendLog(level, message)
+        }
```
eed3si9n, line 123: Do all `log.info(...)` calls become synchronous?

internal/util-logging/src/main/scala/sbt/internal/util/ManagedLogger.scala
```diff
@@ -35,29 +35,53 @@ class ManagedLogger(
...
+    debugEvent(event)(f, StringTypeTag.apply(t))
+  @deprecated("Use macro-powered StringTypeTag.fast instead", "1.4.0")
+  final def infoEvent[A](event: => A, f: JsonFormat[A], t: TypeTag[A]): Unit =
+    infoEvent(event)(f, StringTypeTag.apply(t))
+  @deprecated("Use macro-powered StringTypeTag.fast instead", "1.4.0")
+  final def warnEvent[A](event: => A, f: JsonFormat[A], t: TypeTag[A]): Unit =
+    warnEvent(event)(f, StringTypeTag.apply(t))
+  @deprecated("Use macro-powered StringTypeTag.fast instead", "1.4.0")
+  final def errorEvent[A](event: => A, f: JsonFormat[A], t: TypeTag[A]): Unit =
+    errorEvent(event)(f, StringTypeTag.apply(t))
```
eed3si9n, line 75: As mentioned in https://github.com/sbt/sbt/pull/5712#discussion_r462737410, `debugEvent[A: JsonFormat: TypeTag](event: => A): Unit` is effectively public, so I don't think we can remove the method.

eatkins: I didn't remove any methods. This passed mima. `debugEvent[A: JsonFormat: TypeTag](event: => A)` desugars to `debugEvent[A](event: => A)(implicit f: JsonFormat[A], t: TypeTag[A])` which is further desugared to a method with the signature `debugEvent[A](event: A, f: JsonFormat[A], t: TypeTag[A])`. For mima purposes, all three versions are the same so binary compatibility is preserved. What this signature change does is allow an automatic upgrade to the new version, `debugEvent[A: JsonFormat: StringTypeTag]` without having to actually change their source code. Builds or plugins compiled for scala 1.4.0 will use the `StringTypeTag` variant but plugins compiled for earlier versions of sbt that use the TypeTag variant will still work with 1.4.0.

eed3si9n: > which is further desugared to a method with the signature `debugEvent[A](event: A, f: JsonFormat[A], t: TypeTag[A])`.

I didn't know about this part. ok then I stand corrected.

internal/util-logging/src/main/scala/sbt/internal/util/ConsoleAppender.scala
```diff
@@ -356,18 +357,38 @@ object ConsoleAppender {
...
+          case l => l
+        }
+      }
+  }
+  override def close(): Unit = log4j.get match {
+    case null =>
+    case a    => a.stop()
+  }
+}
+trait ConsoleAppender extends AutoCloseable {
```
eed3si9n: Would it make sense to call this `Appender`? `ConsoleAppender` sounds like it's a concrete appender targeting the console.

eed3si9n:
If the ring buffer + thread have been doing its job, then it would have increased throughput of the logs. Would it be possible to somehow benchmark that aspect? We could generate bunch of subproject and make them run a task that logs bunch of stuff emulating compilation or test.

internal/util-logging/src/main/scala/sbt/internal/util/ConsoleAppender.scala
eatkins: Yeah. That was how I first did it but I reverted to `ConsoleAppender` in an attempt to minimize the diff. I'd be happy to switch back to `Appender` which is a better name for the reason you described.

eatkins:
I'm sure we could make benchmarks but I'm not sure it's the best use of the limited time we have to finish 1.4.0. After spending a lot of time performance hacking on sbt, I don't believe that the actual logging is taking up much time. Even running compile in a large multi project build is not generating nearly the volume of logging that a busy web endpoint handling thousands of requests per second will which is the use case that I believe most logging frameworks are targetting. I have never seen console appenders in a flamegraph. I have seen creating log4j loggers there but not the actually logging.

**EDIT** I have basically been doing by hand the exact benchmark you describe. In https://github.com/jtjeferreira/sbt-multi-module-sample which has 20 subprojects, after I removed memory leaks, the total time to run a no-op compile for both the log4j and sbt implementation is about 250ms which represents all of the sbt overhead including zinc. When I made the sbt implementation asynchronous, that number dropped to 230ms. From this, I strongly suspect that logging is less than 10% of the total sbt overhead. Other things like enqueueing tasks with the executor service and calculating progress state show up in a flamegraph much more than logging. Also it's worth keeping in mind this is just the sbt overhead since it was a no-op compile. Once sbt actually starts compiling files and/or running tests, logging likely becomes a smaller and smaller share of the total overhead.

eatkins:
I'm much more concerned about the memory leaks than logging throughput. The memory leaks were causing much worse degradation of performance than any potential benefit of the logging framework for throughput. Once sbt starts GC thrashing, it becomes unusably slow and this can happen surprisingly quickly. Even if these did cause a performance regression in terms of top end logging throughput, and I doubt that is the case based on my anecdotal experience, it would still be well worth doing because the memory leaks badly affect the scalability of sbt.

eatkins:
Also I'm not sure that log4j is running asynchronously because I get essentially identical timings when I run no-op compile using both implementations in https://github.com/jtjeferreira/sbt-multi-module-sample (which is effectively my benchmark for this and also how I discovered the leak). When I do a jstack of the running process using log4j, there are no threads that are named with log4j or apache (though I'm not 100% sure if they name their threads though naively I'd assume they do). We weren't using AsyncAppender anywhere except in `ManagedLoggerSpec`. I was able to get mildly better performance with the sbt implementation if I updated all of the `ConsoleOut` implementations to do their work asynchronously on a cached thread pool. This also made me suspicious that log4j wasn't actually being used in async mode.

eed3si9n:
In https://github.com/sbt/sbt/pull/2932 I enabled async logging via `sbt/src/main/resources/log4j2.component.properties`. In sbt 1.3.0 I apparently made it synchronous for super shell reason - https://github.com/sbt/sbt/pull/4604/commits/465ff8e10a65ce3535d7d11032714608034c16a7
Maybe we can try enabling async log again.

eatkins:
Screen Shot 2020-08-08 at 10 34 27 PM
The purple section of the flame graph is the time spent in ConsoleOut. It represented < 2% of the total samples in the flamegraph. I ran the flame graph for 40 no-op compiles in that project. I just don't think logging is a performance issue.

eatkins:
AsyncLogging is also not in general awesome for an interactive session because it makes it hard to get the timing of the ui right. It's a good option for writing to disk but should be avoided for writing to System.out.

eed3si9n:
But don't we write to files for task logging to implement `last`?

eatkins:
Yes. But I still don't think it's enough of a bottleneck to worry about.

eatkins:
I guess it's fair to point out that I only test on a computer with a fast ssd so it's certainly possible I'm underestimating the performance cost of logging. Even if that is the case, it would be easy to move the actual work of `ConsoleOut.println` onto background threads. We don't need log4j async appenders for that.

eed3si9n:
(Funny I was just typing) It's entirely possible that I am fighting the last war when CPU core was increasing but IO was slow because of spinning disk. But some Play projects could easily have 100 subprojects, all trying to write debug level things Ivy or whatever outputs. I'm happy to go with your implementation during RC-1 given that you've provided escape hatch to go back to the old appender.

internal/util-logging/src/main/scala/sbt/internal/util/ConsoleAppender.scala
eed3si9n: Yea. I'd suggest `Appender` to be an interface trait with no implementation and `ConsoleAppenderImpl` could become `ConsoleAppender`?

eatkins: Done https://github.com/sbt/sbt/compare/2dc159fe8c7bb68f473a84ec8ccd7f53feefb71a..a8c1b3160badb470ceaaacfeec7bea46082a5703.

eed3si9n (approved):
Thanks for the contributions @eatkins. 

I think overall this is ready to be merged, but there are some unknown-unknowns surrounding this, so we'll see how this goes during RC cycle.

eatkins:
> but there are some unknown-unknowns

Ha. Very true!

PR #5732, opened 2020-08-08
Call info before scripted tests

eatkins:
If one scripted test calls debug, all subsequent tests will be in debug
mode until another test calls info. We can just call info before each
test run to fix that behavior.


eatkins:
Huh. Not sure why this caused scripted failures but I don't care to debug further.

PR #5737, opened 2020-08-10
Add reprompt fast track command

eatkins:
With the latest sbt snapshot, the ui would get stuck if the user entered
an empty command. They would be presented with an empty prompt and could
not input any commands. This was caused by the change in
d569abe70ae9b5e95fb0d699fb01ffd492d9345e that reset the prompt after a
line was read. I had tried to optimize line reading by ignoring empty
commands in UITask.readline so we wouldn't have to make a new thread.
This optimization wasn't really buying much since it only affects how
quickly the user is reprompted after entering an empty command. Unless a
user is spamming the <enter> key, they shouldn't notice a difference.

### DIFF
--- a/internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
+++ b/internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
@@ -74,13 +74,13 @@ object LineReader {
       parser: Parser[_],
       terminal: Terminal,
   ): LineReader = {
-    val term = JLine3(terminal)
     // We may want to consider insourcing LineReader.java from jline. We don't otherwise
     // directly need jline3 for sbt.
-    val reader = LineReaderBuilder.builder().terminal(term).completer(completer(parser)).build()
-    historyPath.foreach(f => reader.setVariable(JLineReader.HISTORY_FILE, f))
     new LineReader {
       override def readLine(prompt: String, mask: Option[Char]): Option[String] = {
+        val term = JLine3(terminal)
+        val reader = LineReaderBuilder.builder().terminal(term).completer(completer(parser)).build()
+        historyPath.foreach(f => reader.setVariable(JLineReader.HISTORY_FILE, f))
         try terminal.withRawInput {
           Option(mask.map(reader.readLine(prompt, _)).getOrElse(reader.readLine(prompt)))
         } catch {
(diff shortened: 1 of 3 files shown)


main-command/src/main/scala/sbt/BasicCommandStrings.scala
```diff
@@ -21,6 +21,7 @@ object BasicCommandStrings {
 
   /** The command name to terminate the program.*/
   val TerminateAction: String = Exit
+  val Reprompt: String = "reprompt"
```
eed3si9n: Do we need to create a pseudo command for this?

eed3si9n:
I'd argue that `""` should always be treated as a valid command.

eatkins:
I don't disagree though I'd ask what does an empty command actually mean? Nevertheless, we still treat `""` as a valid command. I just need to go through fast track so that the ui doesn't get blocked if another task is running when the user presses enter. The reason I created the pseudo command `Reprompt` was because `""` has no semantic meaning on its own.

eatkins:
I guess what I'm asking is do we support parsing the empty string? Do we want to allow users to do that? If not, then I think it's a good thing to make it explicit that inputing the empty string will reprompt the terminal.

eed3si9n:
I see this as a separation of (command) responsibility. `shell` command should prompt exactly once, and queue the result `x` together with `"shell"`: as `x :: "shell" :: state`. In this case `x` is `""`.

The semantics of `""` should be that it does nothing.

eatkins:
I agree but the problem is that shell serves multiple clients. If we handled `""` in shell, then the ui would get stuck when a user inputs `""` and the server is running another command (which could easily be the case if bsp is doing something). The reason for the fast track command queue is to allow special commands like `exit` and the empty string to be evaluated immediately without going through the shell command.

eatkins:
I'd be happy to drop the `Reprompt` constant and just inline `""` if that would be preferable. This is really just a bug fix for a regression that I introduced in https://github.com/sbt/sbt/commit/d569abe70ae9b5e95fb0d699fb01ffd492d9345e . The implementation is slightly different because I moved the reprompting from internal to `UITask.readLine` to the fast track command queue because I found that easier to reason about. Behaviorally, this should not be any different from how sbt has always been.

eed3si9n:
> I'd be happy to drop the `Reprompt` constant and just inline `""` if that would be preferable.

Yea. I think that would be better IMO with some comments summarizing the stuckness problem.

eatkins:
Done

eatkins:
I realize now btw that the pr name and commit heading was kind of misleading. I should have titled it something like, 'Ensure user is reprompted after inputing ""'

eed3si9n (approved):

PR #5748, opened 2020-08-14
Add reload to watch commands

eatkins:
I have this in my global git config and think it's useful enough to
promote to the default.

### DIFF
--- a/main/src/main/scala/sbt/nio/Watch.scala
+++ b/main/src/main/scala/sbt/nio/Watch.scala
@@ -493,6 +493,7 @@ object Watch {
   final val defaultInputOptions: Seq[Watch.InputOption] = Seq(
     Watch.InputOption("<enter>", "interrupt (exits sbt in batch mode)", CancelWatch, '\n', '\r'),
     Watch.InputOption(4.toChar, "<ctrl-d>", "interrupt (exits sbt in batch mode)", CancelWatch),
+    Watch.InputOption('l', "reload the build", Reload),
     Watch.InputOption('r', "re-run the command", Trigger),
     Watch.InputOption('s', "return to shell", Prompt),
     Watch.InputOption('q', "quit sbt", Run(TerminateAction)),


eed3si9n (approved):

main/src/main/scala/sbt/nio/Watch.scala
```diff
@@ -493,6 +493,7 @@ object Watch {
   final val defaultInputOptions: Seq[Watch.InputOption] = Seq(
     Watch.InputOption("<enter>", "interrupt (exits sbt in batch mode)", CancelWatch, '\n', '\r'),
     Watch.InputOption(4.toChar, "<ctrl-d>", "interrupt (exits sbt in batch mode)", CancelWatch),
+    Watch.InputOption('l', "reload the build", Reload),
     Watch.InputOption('r', "re-run the command", Trigger),
```
eed3si9n, line 497: It might be slightly confusing since we have similar set of alphabet soup on load failure:

```
Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)
```

It might be more consistent if `r` reloaded the build, and `t` re-triggered the command?

eatkins:
I agree about consistency but I’m not sure we should change it given that ‘r’ for trigger has been present for over a year at this point. In my mind it’s similar to how I wanted to change how enter worked in ~ but we vetoed it to preserve people’s existing workflows/muscle memory.

eed3si9n:
Yea `<enter>` has been there since the beginning of `~` and has been used by most ppl who ever used `~` so changing that is hard. I am personally ok with the new alphabets switching around a bit in 1.4.0.

But in a second thought, `r`etry in failed load context is the most natural next action, so maybe `r` mapped to retry is not so bad in the context of `~`.

eatkins:
Yeah. If you think of 'r' as standing for retry, it actually is consistent even if retrying means reloading in some contexts and re-running in others.

PR #5752, opened 2020-08-14
Cancel all running tasks with ctrl+c

eatkins:
Frequently ctrl+c does not work to cancel the running tasks. This seems
to be because the signal handler is bound to a specific instance of
evaluate task but there may be multiple instances of evaluate task
running at any given time. Shutting down just one of the running engines
does not ensure that task evaluation stops. To work around this, we can
globally store all of the completion services in a weak hash map and
cancel all of them whenever a signal is received. Closing the service,
which happens at the end of task evaluation will remove the service from
the map so hopefully this shouldn't introduce a memory leak.

### DIFF
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -480,6 +480,7 @@ object EvaluateTask {
       def cancelAndShutdown(): Unit = {
         println("")
         log.warn("Canceling execution...")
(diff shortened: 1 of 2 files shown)


eatkins:
I was motivated by running mimaReportBinaryIssues in the sbt project, immediately cancelling it and then having to wait 45 seconds for mima to run anyway. After this change, cancellation worked as expected.

eed3si9n (approved):
thanks @eatkins!

PR #5765, opened 2020-08-16
Banner for 1.4.0 beta 

eed3si9n:
```
[info] welcome to sbt 1.4.0-SNAPSHOT (AdoptOpenJDK Java 1.8.0_232)
[info] loading settings for project global-plugins from ...
[info] loading global plugins from ...
[info] loading project definition from /private/tmp/hello/project
[info] set current project to hello (in build file:/private/tmp/hello/)
[info]
[info] Here are some highlights of this release:
[info]   - Build server protocol (BSP) support
[info]   - sbtn: a thin client for sbt built as GraalVM native image
[info]   - VirtualFile + RemoteCache: caches build artifacts across different machines
[info]   - Incremental build pipelining. Try it using `ThisBuild / usePipelining := true`.
[info] See http://eed3si9n.com/sbt-1.4.0-beta for full release notes.
[info] Hide the banner for this release by running `skipBanner`.
[info] sbt server started at local:///Users/eed3si9n/.sbt/1.0/server/478e6db75688771ddcf1/soc
```

### DIFF
--- a/main/src/main/scala/sbt/internal/Banner.scala
+++ b/main/src/main/scala/sbt/internal/Banner.scala
@@ -10,6 +10,16 @@ package sbt.internal
 private[sbt] object Banner {
   def apply(version: String): Option[String] =
     version match {
+      case v if v.startsWith("1.4.0") =>
+        Some(s"""
+                |Here are some highlights of this release:
+                |  - Build server protocol (BSP) support
+                |  - sbtn: a native thin client for sbt
+                |  - VirtualFile + RemoteCache: caches build artifacts across different machines
+                |  - Incremental build pipelining. Try it using `ThisBuild / usePipelining := true`.
+                |See http://eed3si9n.com/sbt-1.4.0-beta for full release notes.
+                |Hide the banner for this release by running `skipBanner`.
+                |""".stripMargin.linesIterator.mkString("\n"))
       case "1.3.0" =>
         Some(s"""
                 |Welcome to sbt $version.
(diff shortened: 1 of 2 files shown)


eatkins:
I'm not sure I want to give oracle free advertising with announcing that it uses graalvm. Can we just say native thin client?

eed3si9n:
I feel like the fact that it uses GraalVM would be a cool thing for `sbtn`, but it's up to you of course.

eatkins:
To me it's first and foremost an implementation detail. I'm not sure I would use GraalVM if it didn't happen to make code sharing of the json codecs easier. GraalVM is not at all enjoyable to use because it's difficult to configure, the build times are extremely slow and it doesn't work with a lot of java libraries. It's a cool technology but I'm more excited that it shows what is possible wrt startup time and memory utilization while still being able to use an expressive language.

eatkins:
Honestly, it might have been less work to write in c since I ended up having to implement ipcsocket anyway.

main/src/main/scala/sbt/internal/Banner.scala
```diff
@@ -10,6 +10,16 @@ package sbt.internal
 private[sbt] object Banner {
   def apply(version: String): Option[String] =
     version match {
+      case v if v.startsWith("1.4.0") =>
+        Some(s"""
+                |Here are some highlights of this release:
+                |  - Build server protocol (BSP) support
+                |  - sbtn: a thin client for sbt built as GraalVM native image
```
eed3si9n: so

```suggestion
                |  - sbtn: a native thin client for sbt
```

?

eatkins: 👍

eed3si9n:
`dependency-management/snapshot-local` is failing but likely it's unrelated, so I'm gonna merge this.

PR #5768, opened 2020-08-16
Restore binary compatibility for LogManager.defaults

eatkins:
Although log manager is in the internal package, akka uses it and it
seems worth preserving binary compatibility in their build, at least for
sbt 1.4.0-M2. Once the community build is passing, we can consider
reverting this.


eed3si9n (approved):
Thanks @eatkins!

PR #5773, opened 2020-08-17
Enable semanticdb with Dotty

### LINKED ISSUE #5772: semanticdbEnabled is incompatible with Dotty
## steps

Create a minimal Dotty project and set `semanticdbEnabled` to true:

```scala
addSbtPlugin("ch.epfl.lamp" % "sbt-dotty" % "0.4.1")
```

```scala
 .settings(
    scalaVersion := "0.26.0-RC1",
    semanticdbEnabled := true
  )
```

Run `sbt compile`.

sbt should complain about the `semanticdb-scalac_0.26.0-RC1` dependency being not found.

## problem

`semanticdbEnabled` add the `semanticdb` compiler plugin which does not exist for Dotty. Semanticdb is natively supported by Dotty but it must be activated by a compiler option.

## expectation

The `semanticdbEnabled` setting should add the right `scalacOptions` depending on the `scalaVersion` of the project.

adpi2:
Fixes #5772

### DIFF
--- a/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala
+++ b/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala
@@ -13,6 +13,7 @@ import sbt.internal.SysProp
 import sbt.librarymanagement.syntax._
 import sbt.librarymanagement.CrossVersion
 import Project.inConfig
+import sbt.internal.inc.ScalaInstance
 
 object SemanticdbPlugin extends AutoPlugin {
   override def requires = JvmPlugin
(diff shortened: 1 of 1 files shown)


main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala
```diff
@@ -33,7 +33,9 @@ object SemanticdbPlugin extends AutoPlugin {
     allDependencies ++= {
       val sdb = semanticdbEnabled.value
       val m = semanticdbCompilerPlugin.value
-      if (sdb) List(Build0.compilerPlugin(m))
+      val sv = scalaVersion.value
+      val isDotty = sv.startsWith("0.") || sv.startsWith("3.")
```
eed3si9n: There's isDotty function on `ScalaInstance` object:
https://github.com/sbt/zinc/blob/6078b6046e38d8d17f3a4e60eddf74e609cf2d8f/internal/zinc-classpath/src/main/scala/sbt/internal/inc/ScalaInstance.scala#L136

eed3si9n (changes requested):
a minor fix request.

eed3si9n (approved):

Issue #5779, opened 2020-08-19
Over-compilation in sbt project with pipelining disabled


eatkins:
## steps

In the sbt project
* disable pipelining
* compile
* apply the following patch:
```patch
diff --git a/main/src/main/scala/sbt/Defaults.scala b/main/src/main/scala/sbt/Defaults.scala
index 39120b652..644bb415e 100644
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -127,6 +127,7 @@ import xsbti.compile.{
 
 object Defaults extends BuildCommon {
   final val CacheDirectoryName = "cache"
+  val x = 1
 
   def configSrcSub(key: SettingKey[File]): Initialize[File] =
     Def.setting {
```
* compile

## problem
A cascade of invalidations causes four cycles to run, ultimately recompiling the entire project:
```
sbt:sbtRoot> compile
[info] compiling 1 Scala source to ${SBT_LOCATION}/main/target/scala-2.12/classes ...
[info] compiling 5 Scala sources and 14 Java sources to ${SBT_LOCATION}/main/target/scala-2.12/classes ...
[info] compiling 3 Scala sources and 14 Java sources to ${SBT_LOCATION}/main/target/scala-2.12/classes ...
[info] compiling 131 Scala sources and 14 Java sources to ${SBT_LOCATION}/main/target/scala-2.12/classes ...
[info] compiling 1 Scala source to ${SBT_LOCATION}/sbt/target/classes ...
[info] compiling 2 Scala sources to ${SBT_LOCATION}/scripted-sbt-redux/target/scala-2.12/classes ...
[success] Total time: 24 s, completed Aug 19, 2020 12:54:49 PM
```
## expectation

There is only 1 compilation cycle that recompiles Defaults.scala only. The diff only adds a single public method that isn't used anywhere else so I don't understand why any other source files are recompiled.

## notes
Here is a gist of the full debug logs of the main project compilation: https://gist.github.com/eatkins/c1031e93a2fe866ae9d7f3871481d805

One thing that is strange is that all of those java source files are invalidated. I find the invalidation messages confusing:
```
[debug] Invalidating (transitively) by inheritance from sbt.internal.FullScalaLoader...
[debug] Initial set of included nodes: sbt.internal.FullScalaLoader
[debug] Invalidated by transitive inheritance dependency: Set(sbt.internal.FullScalaLoader)
```
[...]

PR #5780, opened 2020-08-19
Split Defaults.scala into a number of smaller files

eatkins:
Defaults.scala has long been a bottleneck for compilation and formatting. There also were a number of cyclic dependencies where classes like `Defaults` and `Classpaths` mutually depended on fields and methods from the other. This PR splits it into a number of different files and tries to remove the cyclic dependencies. I was partially motivated by https://github.com/sbt/sbt/issues/5779 but I have long wanted to make this split because of the slow compile time of Defaults.scala.

After the split, I can add a method to Defaults.scala and it doesn't trigger a cascade of invalidations like in #5779. Unfortunately the overcompilation is present if I add a field to `GlobalDefaults`. Nevertheless, just compiling Defaults.scala takes less than 2 seconds on my computer now, compared to about 4 seconds before.


eatkins:
The first two commits just move classes into different files. The last commit is a bit more involved because it actually moves method and field definitions to different classes but there should be no logical change.

sjrd:
Currently, my actual documentation for sbt consists in two files: `Keys.scala` and `Defaults.scala`. If I want to know what keys exist and what is their types, I go to `Keys.scala`. If I want to know what they do and how they interact with others and how they are scoped, I go to `Defaults.scala`. I have been recommending this to people who need to go past the easiest understanding of sbt.

After this PR, where should I go for the latter? In what file, or in what set of files? What would be my strategy? Can we make this reasonably easy, for example by putting all the files that implement settings and keys in a dedicated (internal) package?

eatkins:
You bring up some great points about the internal structure of sbt and making it navigable for all. My overriding goal is always to make sbt easier to work on and to understand.

My biggest problem with working on sbt since I began contributing has been the tangled mess that is Defaults.scala. I find the file very difficult to navigate because of its exceptional length and mix of responsibilities. In my view, Defaults.scala is a place where we amalgamate arguably dozens of plugins into a single giant plugin. On the one hand, this is good from a macro scale because a reader can assume that most sbt default tasks and settings are implemented in that file. But on the other hand, because there is so much going on in the file, it is very difficult to make sense of on a micro scale. In some cases, related tasks and settings may be separated by literally thousands of lines in the file. Moreover, because it defines settings at both the global, project and config level, it can be difficult to tell at what scope the settings are being applied without having to scan up in the file. Making sense of an individual task definition often involves jumping around between lines in Defaults.scala that are separated by thousands of lines.

To draw an analogy with the scala standard library, the current situation with Defaults.scala and Keys.scala is kind of like if Predef.scala actually contained all of the implementation details (like e.g. embedding `ArrayOps` as a nested class within `Predef`). Yes, it would make finding where things are defined in the scala library easy but would anyone like working with it?

On top of that, dumping everything into Defaults.scala hurts the developer experience of working with sbt. Changes to Defaults.scala take a long time to compile and format (which is more a scalac problem than an sbt problem) and this increases the development feedback cycle which is very frustrating when you are working on sbt day in day out. It might seem like a minor thing that Defaults.scala takes 6 seconds to compile every time but it becomes very grating when it is the primary project that you work on.

The long term goal here is to modularize Defaults.scala both for compilation performance _and_ to improve readability. It is already the case that there are a number of tasks and settings that are implemented outside of Defaults.scala. For example, most of the watch settings are defined in Watch.scala: https://github.com/sbt/sbt/blob/b8dac523386310d52efa71e04b443970539327d0/main/src/main/scala/sbt/nio/Watch.scala#L611-L626
and added to the defaults with a single ++
https://github.com/sbt/sbt/blob/b8dac523386310d52efa71e04b443970539327d0/main/src/main/scala/sbt/Defaults.scala#L172
This pattern is used in a number of places. It effectively allows Defaults.scala to be an index to the classes that actually implement the tasks and settings. In this PR, I moved helper task and function definitions into objects in the `sbt.internal` package. These are generally small files that do one thing or at most a few with a filename that matches the functionality provided. Even in cases like Watch where the class is in the `sbt.nio` package rather than the `sbt.internal` package, it should be easy to find the definition either with goto def in an ide, git grep, github search or, best of all, the `inspect` command. Yes, it's slightly more friction than using ctrl+f or `/` to search in Defaults.scala, but only marginally so.

That all being said, this changeset is hardly perfect and I didn't spend a ton of time preparing it. In my view, Defaults.scala could use some more robust surgery so that it is more clearly modularized into a number of smaller plugin-like classes (I say plugin like because the helper classes are not literally instances of `AutoPlugin` but perform a similar role). For example, it was easy to just move `Classpaths` into its own file because the classpath related settings were already grouped together in a separate object. It's a lot more work to go through and actually split out settings into smaller collections of related groupings.

In my personal ideal world, I think a new feature definition would follow more or less this pattern:
```scala
object NewFeature {
  trait Keys {
    val foo = settingKey[Int]("...")
    val bar = taskKey[Unit]("...")
  }
  object Keys extends Keys
  val defaultSettings = Def.settings {
    Keys.foo := 1
    Keys.bar := println("hello")
  }
}
```
would be defined in a file `NewFeature.scala`. We would mix `NewFeature.Keys` into `sbt.Keys` (assuming we want this feature to be in the auto imports for build.sbt files) and then append `NewFeature.defaultSettings` into the global, project or config settings in `Defaults.scala` as appropriate.

That is my vision for how features should be defined and organized. I recognize that there is a tradeoff to modularization. The most obvious being that it simply makes it more difficult to know a priori where you can find a task or setting definition. At the same time, the inspect command tells you exactly what file a setting or task was defined in so I'm not sure that we should prioritize having all of the definitions in a single source file. As long as an individual feature is implemented in a single file so that the user does not have to jump between files to understand a task definition, I think modularity is a net benefit. Others may not agree and I'm not trying to impose my views if there is consensus that we don't want to move in this direction.

eed3si9n:
I feel both sides of the pain, but between improving the learning curve of sbt and developing experience, I'd prioritize the learners. That said, I think `Defaults.scala` has accumulated too much details over the course of years because it's too easy to just add one more thing. Inspired by Ethan's pattern, I've moved inter-project classpath related implementations to a separate file https://github.com/sbt/sbt/blob/b8dac523386310d52efa71e04b443970539327d0/main/src/main/scala/sbt/internal/ClasspathImpl.scala in my recent work.

I wonder if there's some happy middle ground where `Defaults.scala` is still useful to see majority of the default settings, but the complicated task implementations are all kicked out to separate files?
Also maybe there are some frequently-looked-up functions like `addSbtPlugin(...)` when someone is opening up `Defaults.scala`?

smarter:
One compromise would be to put the tasks and settings definitions in the same package that doesn't contain anything else (e.g. `sbt.definitions`), that way it' still easy to grep for them without having to go look everywhere.

eatkins:
I'm going to close this since there is no consensus.

PR #5782, opened 2020-08-20
Add bspEnabled setting

### LINKED ISSUE #5675: Ability to disable BSP on a sub-project
In some elaborated build there are reasons for disabling BSP on some sub-projects:
- the project share its sources with other projects (ex: cross-compilation)
- the compilation of the project is resource intensive

I suggest to add a setting `bspEnabled` of type ` Boolean` that is true by default `Global / bspEnabled := true` and that can be overloaded by `<project> / bspEnabled := false`

adpi2:
Fixes #5675

We have `Global / bspEnabled := true` by default.

One can set `<project> / <config> / bspEnabled := false` to disable BSP on a specific project and configuration. It makes the corresponding target out of the BSP workspace and thus unknown by the BSP client.

This can be handy if the project is very resource consuming or very involved in some way.


eed3si9n (dismissed):

adpi2:
I've rebased this branch to fix the conflicts

eed3si9n (approved):

Issue #5784, opened 2020-08-20
1.4.0-M2 writes unexpected control characters to shell


jastice:
## steps

Import an sbt project set to version 1.4.0-M2 into IntelliJ with Scala plugin as sbt project. Do not enable "use shell for imports"

## problem

The import fails to complete. In the log we see what looks like a control character: `[0J`
This apparently breaks the importing - the process just hangs there instead of accepting input.

From the log:
```
/Library/Java/JavaVirtualMachines/jb-jdk-11.0.6-b520.43/Contents/Home/bin/java -Djline.terminal=jline.UnsupportedTerminal -Dsbt.log.noformat=true -Dfile.encoding=UTF-8 -Didea.managed=true -Dfile.encoding=UTF-8 -jar /Users/jast/workspace/scalaUltimate/target/plugin/Scala/launcher/sbt-launch.jar early(addPluginSbtFile="""/private/var/folders/5k/4b5my29x26q_dhrys3ghf9wh0000gn/T/idea7.sbt""")
[info] welcome to sbt 1.4.0-M2 (JetBrains s.r.o Java 11.0.6)
[0J
[info] loading global plugins from /Users/jast/.sbt/1.0/plugins
[0J[info] loading settings for project untitled52-build from idea7.sbt ...
[0J[info] loading project definition from /Users/jast/playspace/untitled52/project
[0J[info] loading settings for project untitled52 from build.sbt ...
[0J[info] set current project to untitled52 (in build file:/Users/jast/playspace/untitled52/)
[0J[info]
[0J[info] Here are some highlights of this release:
[0J[info]   - Build server protocol (BSP) support
[0J[info]   - sbtn: a native thin client for sbt
[0J[info]   - VirtualFile + RemoteCache: caches build artifacts across different machines
[0J[info]   - ThisBuild / versionScheme to take the guessing out of eviction warning
[0J[info] See http://eed3si9n.com/sbt-1.4.0-beta for full release notes.
[0J[info] Hide the banner for this release by running `skipBanner`.
[0J[info] sbt server started at local:///Users/jast/.sbt/1.0/server/a58893d9233954316961/sock
[0J
```

note that the `-Dsbt.log.noformat=true` flag is passed, so I wouldn't expect any control characters.

## notes

The import works fine over the sbt shell integration, which overrides the console prompt.

eed3si9n:
`Esc[0J` :  Clear screen from cursor down | ED0, according to ascii-table.com.

[...]

PR #5788, opened 2020-08-21
Fixes IntelliJ import for 1.4.x

eatkins:
@jastice reported in https://github.com/sbt/sbt/issues/5784 that intellij import is broken with 1.4.0-M2. While debugging this, I found that the jline3 line reader was causing problems for this use case. It was fixable by reverting back to jline 2 when `System.console == null`. My thinking is that we are using jline3 solely for the enhanced tab completions so if `System.console == null` we can assume that we don't need to support enhanced tab completions for that use case. If this assumption is violated, I added an explicit opt in for jline 3 by setting `-Dsbt.jline.version=3`.

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/ConsoleAppender.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/ConsoleAppender.scala
@@ -148,40 +148,9 @@ object ConsoleAppender {
    * 3. -Dsbt.colour=always/auto/never/true/false
    * 4. -Dsbt.log.format=always/auto/never/true/false
    */
-  lazy val formatEnabledInEnv: Boolean = {
-    def useColorDefault: Boolean = {
-      // This approximates that both stdin and stdio are connected,
-      // so by default color will be turned off for pipes and redirects.
-      val hasConsole = Option(java.lang.System.console).isDefined
-      ansiSupported && hasConsole
-    }
-    sys.props.get("sbt.log.noformat") match {
-      case Some(_) => !java.lang.Boolean.getBoolean("sbt.log.noformat")
-      case _ =>
-        sys.props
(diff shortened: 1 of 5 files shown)


eed3si9n (dismissed):

eed3si9n:
AppVeyor doesn't look happy:

```
[info] [info] Received input event: ShowOptions.
[info] [info] Options:
[info] [info]   <ctrl-d> : interrupt (exits sbt in batch mode)
[info] [info]   <enter>  : interrupt (exits sbt in batch mode)
[info] [info]   '?'      : print options
[info] [info]   'l'      : reload the build
[info] [info]   'q'      : quit sbt
[info] [info]   'r'      : re-run the command
[info] [info]   's'      : return to shell
[info] [info] Received input event: ShowOptions.
[info] [info] Options:
[info] [info]   <ctrl-d> : interrupt (exits sbt in batch mode)
[info] [info]   <enter>  : interrupt (exits sbt in batch mode)
[info] [info]   '?'      : print options
[info] [info]   'l'      : reload the build
[info] [info]   'q'      : quit sbt
[info] [info]   'r'      : re-run the command
[info] [info]   's'      : return to shell
[info] [info] Received input event: ShowOptions.
[info] [info] Options:
[info] [info]   <ctrl-d> : interrupt (exits sbt in batch mode)
....
```

eatkins:
Yep! I am going to figure out what is going wrong. Sadly I'll probably need to delve further into the differences between the jline 2 and jline 3 line readers. I'd hoped this could be a simple fix but perhaps it isn't.

eatkins:
It turns out that I didn't need to revert to a jline 2 line reader to make this work. It was just a matter of correctly wiring up the terminal input stream.

eed3si9n (approved):

PR #5793, opened 2020-08-23
Kill external processes on sigint

eatkins:
On linux and mac, entering ctrl+c will automatically kill any forked
processes that were created by the sbt server because sigint is
automatically forwarded to the child process. This is not the case on
windows where it is necessary to forcibly kill these processes.


eed3si9n (dismissed):

eed3si9n (approved):
Thanks @eatkins!

PR #5806, opened 2020-09-02
Allow incompatible mixed scala versions

eatkins:
The zinc scripted project depends on all of the compiler bridges. As a
result the introduction of the strict scala binary version check in
f8139da1920944b8a4912fa54d898c99ae81e36a broke zinc scripted. This
commit reverts to the old behavior in the non scala sandwich case.

I also switched to a for comprehension instead of a pattern match
because this is a rare case where I think it made the code significantly
more readable.


eed3si9n (approved):
Thanks for the fix!

PR #5812, opened 2020-09-05
Add gc monitor to warn users about excessive gc

### LINKED ISSUE #5044: GC thrashing can cause silent hangs
This is more of a feature idea than an issue. Sometimes sbt will become unresponsive because it's stuck in a full GC loop. I've been thinking of adding a GC monitor that will print warnings to the console if sbt is spending a lot of time in GC. The JDK provides tools that should make this relatively straightforward.

eatkins:
Fixes https://github.com/sbt/sbt/issues/5044

It is not uncommon in large projects for the jvm to silently be running
frequent full gcs in the background. This can slow progress to a crawl.
Usually the fix is to bump the -Xmx parameter but if the users do not
realize that their tasks are slow because of gc thrashing, they may not
think to do that. This PR adds a monitor that hooks into the jvm's event
notification system to keep track of how much time is spent in GC. If
the ratio of the amount of time in gc to the total elapsed time exceeds
some threshold, we emit a warning.

I was motivated to do this because publishLocal can take forever in the
zinc project because a 1G heap isn't big enough.

### DIFF
--- a/main/src/main/scala/sbt/MainLoop.scala
+++ b/main/src/main/scala/sbt/MainLoop.scala
@@ -153,6 +153,7 @@ object MainLoop {
     val superShellThreshold =
       state.get(Keys.superShellThreshold.key).getOrElse(SysProp.supershellThreshold)
     val taskProgress = new TaskProgress(superShellSleep, superShellThreshold)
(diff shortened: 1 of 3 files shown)


eatkins:
Here it is in action in the zinc project:
Screen Shot 2020-09-05 at 1 49 32 PM

eed3si9n (approved):

Issue #5665, opened 2020-07-02
🚲🏚️ What should we name different programs and how are they invoked?


eed3si9n:
## issue

Once https://github.com/sbt/sbt/pull/5620 lands, we will have a compelling thin client that people may want to use for day-to-day coding.

I'm not sure if this would be internal JAR names or actual programs people invoke, but I think we should discuss the names of things. Overall goal is so people can configure what `sbt` command does in the terminal.

## proposal

- sbt server, which is represented by `sbt-1.x.y.jar`. We might need a retronym for "sbt" like the existence of oat milk and skim milk now necessitates the term _whole milk_, as opposed to _milk_. Let's say it's called `sbt-server`. When the user types `sbt --server`, it should always invoke the server like sbt 1.3.x.
- native thin clients, which will be represented by Graal native images. `sbtn` could be the name to always invoke the native thin client. When the user type in `sbt --client` the native thin client is invoked.
- Fallback bridge (JVM thin client). It would be an implementation detail of `sbtn` or  `sbt --client`. 
- sbt shell script. `sbt`. Depending on the setting (environmental variable?) it would invoke either sbt server or native thin client.

(2020-10-10: Retroactively renamed to sbtn to avoid confusion)

## notes about `sbt`

`sbt`-the-script has a responsibility to launch all of sbt 0.13 and sbt 1.x. So if we want it to automatically use the native version for sbt 1.4.0, it needs to at least support parsing of `project/build.properties`.

eatkins:
@eed3si9n and I discussed offline that `sbtc` makes him think `sbt compiler` or something like that because of the analogy with `scala` and `scala`. I don't love `sbtc` or `sbtb` because I'm a mediocre typist and neither is easy to type. I find `sbtc` slightly more ergonomic than `sbtb` (I find the `btb` awkward to type). Since sbt is meaningless at this point anyway, I propose that we could get a little creative. For me, the most ergonomic name that contains an sbt prefix is `sbtl`. I could invent some weird explanation about how `l` stands for limited or local or something but I care more about ergonomics than the actual name. If we care about the meaning, I prefer `rsbt` to `sbtb` where `r` stands for relay. I find `rsbt` easier to type than `sbtr`.

eed3si9n:
If we use `sbt` as the primary entry point, then I'm not sure if ergonomic matters too much. Typing four characters is going to be somewhat awkward initially. For the purpose of debugging, it would be useful that it's memorable.

[...]

PR #5823, opened 2020-09-08
Add pseudo --server command

eed3si9n:
Ref https://github.com/sbt/sbt/issues/5665

This adds `--server` command that is immediately filtered out in Main.scala.
The purpose of `--server` is so we can invoke thin client from `sbt` script at some point in the future when Bash script can parse `project/build.properties`.

`sbtn` would need to call `sbt` again to start the server, and at that point the shell script would need to actually invoke the server. The intent of `--server` is to be used as the tie breaker.

Also build users may want to sometimes call `sbt --server`.

### DIFF
--- a/main-command/src/main/scala/sbt/BasicCommandStrings.scala
+++ b/main-command/src/main/scala/sbt/BasicCommandStrings.scala
@@ -200,6 +200,10 @@ $AliasCommand name=
 	Starts the server if it has not been started. This is intended to be used with
 	-Dsbt.server.autostart=false."""
 
+  def ServerDetailed: String =
+    "--server always runs sbt in not-daemon mode."
+  def DashDashServer: String = "--server"
+
   def OldShell: String = "oldshell"
   def OldShellDetailed = "Provides an interactive prompt from which commands can be run."
 
--- a/main-command/src/main/scala/sbt/BasicCommandStrings.scala
+++ b/main-command/src/main/scala/sbt/BasicCommandStrings.scala
@@ -212,7 +212,7 @@ $AliasCommand name=
     "Provides an interactive prompt from which commands can be run on a server."
   def DashClient: String = "-client"
   def DashDashClient: String = "--client"
-  def CloseIOStreams: String = "--close-io-streams"
+  def DashDashDetachStdio: String = "--detach-stdio"
 
   def StashOnFailure: String = "sbtStashOnFailure"
   def PopOnFailure: String = "sbtPopOnFailure"
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -298,7 +298,8 @@ class NetworkClient(
             term.isSupershellEnabled
           ).mkString(",")
 
-        val cmd = arguments.sbtScript +: arguments.sbtArguments :+ BasicCommandStrings.CloseIOStreams
+        val cmd = List(arguments.sbtScript) ++ arguments.sbtArguments ++
+          List(BasicCommandStrings.CloseIOStreams, BasicCommandStrings.DashDashServer)
         val processBuilder =
           new ProcessBuilder(cmd: _*)
             .directory(arguments.baseDirectory)
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -299,7 +299,7 @@ class NetworkClient(
           ).mkString(",")
 
         val cmd = List(arguments.sbtScript) ++ arguments.sbtArguments ++
-          List(BasicCommandStrings.CloseIOStreams, BasicCommandStrings.DashDashServer)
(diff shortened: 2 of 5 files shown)


main-command/src/main/scala/sbt/BasicCommandStrings.scala
```diff
@@ -200,6 +200,12 @@ $AliasCommand name=
 	Starts the server if it has not been started. This is intended to be used with
 	-Dsbt.server.autostart=false."""
 
+  def Server: String = "server"
+  def ServerDetailed: String =
+    "--server always runs sbt in not-daemon mode."
+  def DashServer: String = "-server"
```
eatkins: I know there was something similar for client but if we are going to have two flags for server mode, I'd prefer `-s` and `--server`. Just dropping the single dash doesn't seem like much of a benefit. Not a big deal though.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -298,7 +298,8 @@ class NetworkClient(
             term.isSupershellEnabled
           ).mkString(",")
 
-        val cmd = arguments.sbtScript +: arguments.sbtArguments :+ BasicCommandStrings.CloseIOStreams
+        val cmd = List(arguments.sbtScript) ++ arguments.sbtArguments ++
+          List(BasicCommandStrings.CloseIOStreams, BasicCommandStrings.DashDashServer)
```
eatkins: I think we should get rid of `CloseIOStreams` and toggle its behavior on `--server` instead. It's not a huge deal but `--close-io-streams` was not a great command argument name anyway so I'd be happy to remove it.

eatkins (dismissed):
It's up to you whether or not you want to address my comments.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: This only makes sense if we assume that the user is not expecting to receive output when invoking `sbt --server`. I'm assuming the use case for `sbt --server` would be to start an sbt server in the background before the user is ready to send any commands. If there is any expectation that the user wants to see the output of the process when running `sbt --server`, then this suggestion would not make sense.

eed3si9n: My vision for manual `--server` was to be used for both use cases: background and foreground (interactive) process.

eatkins: Hmm. I find that a little confusing. It would be nice if there was a way to make it more explicit that there is a background and foreground mode. The --close-io-streams is effectively background mode but it's a bad parameter name.

eed3si9n: `--daemon-mode` or `--detach-stdio`?

eatkins: How about `--daemon`? The mode seems redundant.

eatkins: Gradle already uses daemon in a similar way which I think is a plus.

eed3si9n: The Gradle option naming is interesting because `--no-daemon` ("I want to directly use Gradle") equals to my `--server` command, and `--daemon` ("I want to use Gradle Daemon") equals to `--client` command. So if we named our daemon mode to `--daemon`, it would do the opposite thing. Given that this option is not likely to be called by humans, `--detach-stdio` might be the best?

I am also open to aligning `--server` to `--no-daemon`.

main-command/src/main/scala/sbt/BasicCommandStrings.scala
eed3si9n: Yea. Single dash thing is a vestige from cargo culting bad `javac` and `scalac` options instead of following the Unix standards. I feel reluctant to add single letter option since we don't support composition of them (scopt does), so maybe we should just drop `-server` and keep only `--server`.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: Good points. Maybe detach-stdio is the best.

eed3si9n: Re: `--no-daemon`, looks like in Gradle `--no-daemon` actually means that it will still use client-server mode but it will automatically shutdown the server after use - https://docs.gradle.org/current/userguide/gradle_daemon.html

eed3si9n: Rename --close-io-streams to --detach-stdio - d24cd2b

main-command/src/main/scala/sbt/BasicCommandStrings.scala
eed3si9n: Single dash dropped in https://github.com/sbt/sbt/pull/5823/commits/bba3d08aac9a3803e3d40e0d0e27c4f47b3541f1

Issue #5833, opened 2020-09-10
Fedora terminal issues


eatkins:
This was taken from a reddit comment from user mtk000: https://www.reddit.com/r/scala/comments/io3z2p/sbt_140rc1_released/

if i simply update the sbt version from 1.3.13 to 1.4.0-RC1 on my fedora 32 instance, i lose color in the sbt output (e.g. the info/warn/success labels on a log line) and the tty echo bit is turned off when i exit sbt and return to the shell (so i have to do 'stty sane' to restore things).

eatkins:
According to the response on reddit, the color issue went away and the tty after exit issues were fixed by #5834

PR #5834, opened 2020-09-10
Restore terminal when exiting Terminal.withStreams

eatkins:
There was a reddit comment that the user's tty was messed up after they
exited sbt:
https://www.reddit.com/r/scala/comments/io3z2p/sbt_140rc1_released/.
This attempts to fix that issue by restoring the terminal before
exiting. In order to ensure the tty is restored, it's necessary to move
the work off of a background thread and delay sbt exit. This does take
about 150ms on my machine but I figure that isn't a huge deal in the
scheme of things.


eatkins:
Should hopefully fix the tty part of #5833

eed3si9n (approved):

PR #5838, opened 2020-09-11
Add BSP workspace/reload

### LINKED ISSUE #5783: BSP: resolving targets does not reload build
## steps

1. import an sbt project set to version `1.4.0-M2` as BSP project into IntelliJ
2. make some change in build.sbt that would cause an error
3. do a project refresh in IntelliJ

## problem

The import works just like before - the project was not refreshed.
To trigger a refresh, it must be manually run from an sbt shell connected to the same server, or the connection and server process must be forcibly closed.

## expectation

Ideally, sbt detects in any project refresh is necessary when calling certain BSP requests, such as `workspace/buildTargets`, and does a reload before answering. If the reload fails, these requests respond with an error.

adpi2:
Fixes #5783 

Give BSP clients the ability to trigger a reload by sending a `workspace/reload` request.

According to the specifiaction in https://github.com/build-server-protocol/build-server-protocol/pull/135, the reload is ignored in case of failure, ie the state of sbt stays the same as before.

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
@@ -17,11 +17,13 @@ import sbt.Def._
 import sbt.Keys._
 import sbt.ScopeFilter.Make._
 import sbt.SlashSyntax0._
+import sbt.StandardMain.exchange
 import sbt.internal.bsp._
 import sbt.internal.langserver.ErrorCodes
 import sbt.internal.protocol.JsonRpcRequestMessage
 import sbt.librarymanagement.Configuration
 import sbt.util.Logger
+import sjsonnew.shaded.scalajson.ast.unsafe.JNull
 import sjsonnew.shaded.scalajson.ast.unsafe.JValue
 import sjsonnew.support.scalajson.unsafe.Converter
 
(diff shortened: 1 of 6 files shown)


main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -17,21 +17,58 @@ import sbt.Def._
...
   private val capabilities = BuildServerCapabilities(
     CompileProvider(BuildServerConnection.languages),
-    dependencySourcesProvider = true
+    dependencySourcesProvider = true,
+    canReload = true
+  )
+
+  private val reload = "bspReload"
+  private val reloadFailed = "bspReloadFailed"
+  private val reloadSucceed = "bspReloadSucceed"
```
eed3si9n: I think it would be better to name this as `bspRealod` since there is a command named `reload`. It's a bit wordier but less ambiguous.

adpi2:
@eed3si9n Can this PR be integrated in `sbt-1.4.0`?

It should be 100% safe since this implementation lives on its own. No existing code is modified.

eed3si9n (approved):

eed3si9n:
> It should be 100% safe since this implementation lives on its own. No existing code is modified.

I merged it, but nothing is totally safe. Sure it's unlikely to invalidate existing builds but if there was a major issue in `workspace/reload`, then we'd likely have to fix it before 1.4.0.

Issue #5825, opened 2020-09-08
1.4.0-RC1: TaskProgress: RejectedExecutionException


jtjeferreira:
## steps

In a play 2.7 project with `PlayImport.PlayKeys.playInteractionMode := play.sbt.StaticPlayNonBlockingInteractionMode` when running `run` and making a request I got:

```
> run
Starting server

--- (Running the application, auto-reloading is enabled) ---

[info] p.c.s.AkkaHttpServer - Listening for HTTP on /0:0:0:0:0:0:0:0:9000
[info] p.c.s.AkkaHttpServer - Listening for HTTPS on /0:0:0:0:0:0:0:0:9443
[success] Total time: 12 s, completed Sep 8, 2020, 7:50:27 PM
[error] a.a.OneForOneStrategy - Task java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@2f3f1ea5[Not completed, task = java.util.concurrent.Executors$RunnableAdapter@813863d[Wrapped task = sbt.internal.TaskProgress$$Lambda$2473/0x0000000840c81840@23a0e9a9]] rejected from java.util.concurrent.ScheduledThreadPoolExecutor@78538804[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 49655]
akka.actor.ActorInitializationException: akka://play-dev-mode/system/StreamSupervisor-0/flow-2-1: exception during creation
	at akka.actor.ActorInitializationException$.apply(Actor.scala:202)
	at akka.actor.ActorCell.create(ActorCell.scala:696)
	at akka.actor.ActorCell.invokeAll$1(ActorCell.scala:547)
	at akka.actor.ActorCell.systemInvoke(ActorCell.scala:569)
	at akka.dispatch.Mailbox.processAllSystemMessages(Mailbox.scala:293)
	at akka.dispatch.Mailbox.run(Mailbox.scala:228)
	at akka.dispatch.Mailbox.exec(Mailbox.scala:241)
	at akka.dispatch.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at akka.dispatch.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
	at akka.dispatch.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
[...]

PR #5840, opened 2020-09-11
Skip task progress work if already shutdown

eatkins:
The play plugin seems to do out of band task evaluation on a stale State
object in the `run` task. As a result, when sbt tries to schedule tasks
to run, they tried to register the work with a closed TaskProgress
instance. There was no guard against this and it ended up causing a
RejectedExecutionException.


eatkins:
~~Fixes the TaskProgress related exception in #5825 but play run still doesn't work.~~

Fixes #5825

eatkins:
Fixes #5825

eed3si9n (approved):

PR #5855, opened 2020-09-15
Skip contents of symlinked directories in clean

### LINKED ISSUE #5852: clean should not follow symlinked directories in `target/`
## steps

```
sbt.version=1.3.0
```

Any `build.sbt`.

Steps:

```
mkdir ~/sbt-test/
echo test > ~/sbt-test/file.txt
cd /some/sbt/project/
mkdir target/
ln -s ~/sbt-test/ target/

sbt clean
```

## problem

If you do the above ↑, then `file.txt` is gone from the system. Data loss. This is absurd.

## expectation

Remove just the `target/sbt-test` *symlink*, and not files inside its target.

## notes

Imagine if `rm -r` did that…

eatkins:
Fixes https://github.com/sbt/sbt/issues/5852

The clean task was previously deleting the contents of directories that
were symlinked into the target directory. This was an oversight because
it never occurred to me that users might symlink a directory whose
contents they did not want deleted into the target directory.


eed3si9n (approved):
Thanks for the quick fix!

michalrus:
Thanks a lot!

PR #5857, opened 2020-09-16
Report absolute paths by default during compilation

### LINKED ISSUE #5853: [1.4.0-RC1] ${BASE} in paths reported from compiler breaks hyperlinking to source
## steps

## problem

## expectation

adpi2:
Fixes #5853

Add a `Global / reportAbsolutePath` setting that is set to true by default.
It adds a `sourcePositionMappers` that converts all `Position` to absolute before logging.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -28,7 +28,7 @@ import sbt.Project.{
   richInitialize,
   richInitializeTask,
   richTaskSessionVar,
-  sbtRichTaskPromise,
+  sbtRichTaskPromise
 }
 import sbt.Scope.{ GlobalScope, ThisScope, fillTaskAxis }
 import sbt.coursierint._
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -407,6 +409,12 @@ object Defaults extends BuildCommon {
       )
     },
     fileConverter := MappedFileConverter(rootPaths.value, allowMachinePath.value),
+    sourcePositionMappers ++= {
+      val fc = fileConverter.value
+      if (reportAbsolutePath.value) {
+        List(toAbsoluteSourceMapper(fc))
+      } else Nil
+    },
```
mkurz, line 417: @adpi2 @eed3si9n Setting a default for `sourcePositionMappers` is a very bad idea and breaks existing apps, see #6352

PR #5858, opened 2020-09-16
[1.3.x] Skip contents of symlinked directories in clean

eatkins:
The clean task was previously deleting the contents of directories that
were symlinked into the target directory. This was an oversight because
it never occurred to me that users might symlink a directory whose
contents they did not want deleted into the target directory.


eatkins:
backports #5852 to 1.3.x

eed3si9n (approved):

Issue #5826, opened 2020-09-09
1.4.0-RC1: Compilation task output continues after console mode starts


mcanlas:
## steps

1. Create a completely empty project, with the only file being `project/build.properties` to specify the version as RC1
2. Type `sbt console` from the shell
(problem does not manifest when using `console` from within an sbt session)

## problem

The compilation task for console keeps counting upwards and showing up on screen, even though the console session has started.

## expectation

SBT should be still/quiet after the console is ready.

eatkins:
I can't reproduce on my computer. Would it be possible to create a small project that reproduces the issue? Are you using linux, mac or windows?

mcanlas:
Looks fixed in 1.4.0-RC2

PR #5873, opened 2020-09-20
Don't display progress in skip tasks in batch mode

eatkins:
In sbt 1.4.0-RC1, if a user ran `sbt console`, the progress lines would
be printed after they had entered the console. This was because the
prompt state was incorrect. To get the prompt in the correct state, we
initialize the prompt to batch and then switch to pending when either
sbt enters the shell or the network client attaches in interactive mode.
We also will now immediately print progress as soon as we enter a skip
task to clear out the progress lines and display the warning about a
running task if there is another client connected while the task is
running.


eatkins:
Fixes #5826

eed3si9n (approved):

Issue #5856, opened 2020-09-16
1.4.0-RC1: `sbt run` hangs while using ammonite as a library


mushtaq:
## steps

- Clone the repo https://github.com/shivanithorat/ammonite-simple
- sbt run

## problem

- `@` prompt of ammonite appears, but does not allow to enter any input
- `ctrl-c` is unable to terminate the process, it has to be killed.

## notes

The main program is just this:

```scala
object Main {
  def main(args: Array[String]): Unit = {
    ammonite.Main().run()
  }
}
```

eatkins:
Unfortunately, I think that this is because ammonite uses jline 3 for its REPL implementation and that conflicts with sbt's jline 3 usage. I don't have the bandwidth to debug ammonite so I'd suggest opening an issue there but I would be happy to help coordinate a fix.

eatkins:
I did verify by adding printing on the sbt side that the bytes are read from the keyboard but for whatever reason ammonite doesn't read them.

eatkins:
I also got input if I rewrote the main method to
```scala
object Main {
  def main(args: Array[String]): Unit = {
    println(System.in.read)
  }
}
```
If I type c<enter> it prints 99 as expected.

mushtaq:
Thanks @eatkins, opened this in ammonite repo: https://github.com/lihaoyi/Ammonite/issues/1116

eatkins:
For further reference, I think the problem is in ammonite [FrontEnds.scala](https://github.com/lihaoyi/Ammonite/blob/4bd225eb73802285ff4c77056594c196d1685c5e/amm/repl/src/main/scala/ammonite/repl/FrontEnds.scala#L24). Here it is creating a new instance of jline terminal and that can cause major problems when run within sbt. The scala repl does something similar, but we are able to hijack the call to TerminalBuilder.build() so that it uses a jline terminal provided by sbt instead. I tried the same trick for ammonite but I think ammonite uses classloader tricks to load the repl so the sbt hack doesn't work.

A workaround is to run sbt with `-Dsbt.log.noformat=true`. This doesn't work well with 1.4.0-RC1 (https://github.com/sbt/sbt/issues/5861) but will be improved in 1.4.0 which should be out soon.

[...]

Issue #5828, opened 2020-09-09
[1.4.0-RC1] Home and Delete keys do not work on Windows 10


melezov:
## steps

On Windows 10 x64, JDK 11/14/...
Run sbt (using the launcher directly to skip sbtconfig.txt / sbtopts):

```
   r:\test>java -Duser.home=r:\empty -jar sbt-launch-1.4.0-RC1.jar
   [info] welcome to sbt 1.4.0-RC1 (AdoptOpenJDK Java 14.0.1)
   [info] loading project definition from R:\test\project
   [info] set current project to test (in build file:/R:/test/)
```

Type "special" keys such as Home/End/Delete.

## problem 

The "special" keys such as Home/End/Delete produce garbage (raw control sequence) on the command line.

```
   r:\test>java -Duser.home=r:\empty -jar sbt-launch-1.4.0-RC1.jar
   [info] welcome to sbt 1.4.0-RC1 (AdoptOpenJDK Java 14.0.1)
   [info] loading project definition from R:\test\project
   [info] set current project to test (in build file:/R:/test/)
   [info]
   [info] Here are some highlights of this release:
   [info]   - Build server protocol (BSP) support
   [info]   - sbtn: a native thin client for sbt
   [info]   - VirtualFile + RemoteCache: caches build artifacts across different machines
   [info]   - ThisBuild / versionScheme to take the guessing out of eviction warning
   [info] See http://eed3si9n.com/sbt-1.4.0-beta for full release notes.
   [info] Hide the banner for this release by running `skipBanner`.
   [info] sbt server started at local:sbt-server-83ade0559bc655750a72
   sbt:test> [3~[3~[3~skipBanner[1~[4~[1~[1~                        
```

## expectation

Home is interpreted.

## notes

This is a regression from the same behavior in SBT 1.x which didn't work before, but got fixed.
I'm assuming it has to do with new JLine.
Luckily, the arrow keys seem to work, so you can workaround these.

I've tried to run it with `-Djline.terminal=jline.UnsupportedTerminal` and then the keys such as Home/End/Delete seem to work out of the box, but I'm not sure if this disables some other useful functionality.

eed3si9n:
@melezov Thanks for the report!

eatkins:
We might need to do something like what is done here: https://github.com/jline/jline3/issues/396. This one is a bit difficult for me to debug because I don't have a keyboard with those keys on it.

melezov:
[...]

PR #5874, opened 2020-09-20
Fix a number of terminal io bugs

### LINKED ISSUE #5863: 1.4.0-RC1 regression: Enter in sbt terminal is not working when running code using readLine
## steps

Non-working Enter is happening under these conditions under Linux version 5.4.0-47-generic and Ubuntu 18.04.1:
```
sbt.version=1.4.0-RC1

$ cat build.sbt 
scalaVersion := "2.13.3"

$ cat Main.scala 
object Main {
    println("Hello, World!")
    val s = scala.io.StdIn.readLine("write something: ")
    println(s"You wrote: $s")
  }
}

$ sbt
[info] welcome to sbt 1.4.0-RC1 (Ubuntu Java 11.0.8)
```
Now run and write something including Enter, see example of user input below echoed to terminal:
```
sbt:hello-scala> run
[info] compiling 1 Scala source to /home/bjornr/tmp/hello-scala/target/scala-2.13/classes ...
[info] running Main 
Hello, World!
write something: write something: now user is typing but WHY write something ECHOED TWICE? now pressing enter
pressing enter again
now enter twice 

readLine is never ended
now pressing Ctrl+D
You wrote: now user is typing but WHY write something ECHOED TWICE? now pressing enter
[success] Total time: 45 s, completed Sep 17, 2020, 12:19:24 PM
sbt:hello-scala> 
[info] shutting down sbt server
```

## problem

Three problems probably related to jline:
* The enter key is not working when trying to complete a readLine by user
* The text message given as argument to readLine is echoed twice
* When finally exiting readLine using Ctrl+D then sbt is terminated

## expectation

* The Enter key should work with readLine as in sbt 1.3.13
* The text message given as argument to readLine should only be printed once
* Ctrl+D in readLine should not terminate sbt

## notes

This might be related to #5861 and #5856

eatkins:
This PR fixes #5863 #5856 #5828.


eed3si9n:
Looks like AppVeyor got stuck?

```
[info] sbt server started at local:sbt-server-cdcbba2ed91f9a230908
[info] new client connected: network-1
{"jsonrpc":"2.0","id":1,"result":{"capabilities":{"textDocumentSync":{"openClose":true,"change":0,"willSave":false,"willSaveWaitUntil":false,"save":{"includeText":false}},"hoverProvider":false,"definitionProvider":true}}}
[info] new client connected: network-2
success
[info] - exit success
[info] new client connected: network-3
[info] - exit failure
[info] new client connected: network-4
success
[info] - two commands
[info] new client connected: network-5
[info] - two commands with failing second
[info] new client connected: network-6
[info] - two commands with leading failure
[info] new client connected: network-7
success
[info] - three commands
[info] new client connected: network-8
[info] - three commands with middle failure
[info] new client connected: network-9
[info] - compi completions
[info] new client connected: network-10

No cached test names found. Press '<tab>' to compile: 
```

eatkins:
Looking

eatkins:
Yep. There was a small oversight that the test caught by hanging. I fixed the underlying problem but also made it so that the tests shouldn't get stuck in the future.

melezov:
Hey there! In regards to #5828, I'm afraid this branch makes things much worse :(

    [instafin-web] $ show sbtVersion
    [info] 1.4.0-SNAPSHOT
    [instafin-web] $ HOME [1;%p1%dH END [1;%p1%dF UP [1;%p1%dA DOWN [1;%p1%dB LEFT [1;%p1%dD RIGHT [1;%p1%dC

The arrow keys no longer work. BACKSPACE works, but I can't get to test DELETE since I can't "move" before another character with either HOME nor LEFT.

Pls, instruct if there's a different component to QA other than eatkins/terminal-fixes (only build SBT, no modules).
The same behavior is exhibited in powershell, windows console, cygwin treminal, ...
Using Windows 10.

The workaround `sbt -Djline.terminal=jline.UnsupportedTerminal` that worked with 1.4.0-RC1 is no longer operational and doesn't allow "typing in" any characters on this branch.

eatkins:
Thanks for testing this. I don't think it should be too terrible to fix your issues. The difference is likely that I have been doing my windows testing on a vm over ssh. I will test directly on the vm using the powershell command.

eatkins:
I can reproduce on my vm outside of ssh so I'm optimistic I can fix this quickly.

eatkins:
@melezov I was able to reproduce and fix the issues you reported with the latest code.

eed3si9n (approved):
Thanks!

melezov:
Great, will test immediately! (better while we're still in RC phase) :)

melezov:
Fantastic, with 1.4.0-RC2 Home, End, Delete and Backspace all work perfectly.

I'd like to also report a collateral enhancement in that the Insert button which didn't work in 1.3.x now works in SBT 1.4.0-RC2 🥇 
(albeit, it still only shows as ~ when in SBT's console, but that's the same as in 1.3.x)

Thanks a lot!

eatkins:
You're welcome. I did indeed have to add special handling for the insert key https://github.com/sbt/sbt/pull/5874/commits/036fef33832ae672f13c394f39a0ed1b5040d4ec. I'm a little confused though. What does "(albeit, it still only shows as ~ when in SBT's console, but that's the same as in 1.3.x)" mean? Does it work in some scenarios but not others? I did test that if I manually typed \E[2~ that it switched to insert mode but I don't actually have an insert key so I wasn't sure that it would definitely work with a keyboard that did have one.

melezov:
Oh, it means that if you run "sbt console" in your project thus entering the Scala REPL for the project, the Insert key will not work and only emit the tilde ~ character instead. 

This is the same as how it behaved in SBT 1.3.x's "sbt console", I'm assuming it's an issue of the underlying Scala REPL.
Honestly extremely low priority, but perhaps the REPL peeps might learn something from your fix?

eatkins:
Aah, that makes sense. Is this scala 2.13, 2.12 or dotty?

melezov:
The `~` is in 2.12. 
But yes, now I remember I've seen the different handling for 2.13 - it works there 🎉

I haven't yet touched dotty with a 9-foot pole, so 🤷‍♀️ there

PR #5875, opened 2020-09-20
Don't stop watch input thread on bad input

eatkins:
In a continuous build in sbt 1.4.0-RC1, if the user enters an invalid
option, it causes the input thread to exit which means the watch would
no longer accept input commands (including <enter> to exit). This fixes
that behavior.


eatkins:
I had noticed this behavior a while ago that in long running continuous builds sometimes keyboard input stopped working but it took me a while to figure out a repro.

eed3si9n (approved):

Issue #5870, opened 2020-09-20
sbtn exits after "starting an instance" on Windows 10


eed3si9n:
## steps

Using sbt 1.4.0-RC1 build.

```
> sbtn.exe
```

## problem

```
> sbtn.exe
[info] entering *experimental* thin client - BEEP WHIRR
[info] server was not detected. starting an instance
```

Then the program quits.

## expectation

It enters into sbt shell or display some error message.

eatkins:
I can't reproduce this with the latest code but I have seen this behavior before. It happens when System.console != null but the console input stream returns -1 to the jline reader https://github.com/sbt/sbt/blob/cc0445db4e9ad649ea2a55988bd0842e64fe8840/internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala#L84-L91 I'm not opposed to adding some logging before exiting but it would be a change from the original working behavior.

eed3si9n:
I wonder if this should be `sbt.bat` instead of `sbt.cmd`?

https://github.com/sbt/sbt/blob/923ae57c1c6f906540f13d3896ba9966216929a6/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala#L983

eatkins:
I have a repro now where sbtn just hangs on startup on windows.

eed3si9n:
I'll move this out to 1.4.1 until we can test more.

eed3si9n:
I get another error using sbt 1.4.0:

```
C:\Users\xxx\work\hello>sbt --client
[info] entering *experimental* thin client - BEEP WHIRR
[info] server was not detected. starting an instance
 ava.io.IOException: Failed to get console mode: T h e   h a n d l e   i s   i n v a l i d .

        at org.jline.terminal.impl.jansi.win.JansiWinSysTerminal.createTerminal(JansiWinSysTerminal.java:47)
        at sbt.internal.util.JLine3$.windowsJansi(JLine3.scala:44)
        at sbt.internal.util.JLine3$.system(JLine3.scala:56)
        at sbt.internal.util.Terminal$.newConsoleTerminal(Terminal.scala:722)
        at sbt.internal.util.Terminal$.withStreams(Terminal.scala:333)
        at sbt.internal.client.NetworkClient$.main(NetworkClient.scala:1135)
        at sbt.internal.client.NetworkClient.main(NetworkClient.scala)
        at sbt.client.Client.main(Client.java:19)
```

eatkins:
[...]

Issue #5872, opened 2020-09-20
sbtn could look for sbt script near by


eed3si9n:
https://github.com/sbt/sbt-launcher-package/pull/328

## steps

Run sbtn inside Travis CI.

## problem

```
Downloading sbt launcher for 1.4.0-RC1:
  From  https://repo.scala-sbt.org/scalasbt/maven-releases/org/scala-sbt/sbt-launch/1.4.0-RC1/sbt-launch-1.4.0-RC1.jar
    To  /home/travis/.sbt/launchers/1.4.0-RC1/sbt-launch.jar
Downloading sbt launcher 1.4.0-RC1 md5 hash:
  From  https://repo.scala-sbt.org/scalasbt/maven-releases/org/scala-sbt/sbt-launch/1.4.0-RC1/sbt-launch-1.4.0-RC1.jar.md5
    To  /home/travis/.sbt/launchers/1.4.0-RC1/sbt-launch.jar.md5
OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=512M; support was removed in 8.0
- sbt --client *** FAILED ***
  RuntimeException: Nonzero exit value: 1
    scala.sys.package$.error(package.scala:30)
    sbt.internal.AbstractProcessBuilder.getString(ProcessImpl.scala:131)
    sbt.internal.AbstractProcessBuilder.$bang$bang(ProcessImpl.scala:133)
    example.test.SbtRunnerTest$.$anonfun$new$28(RunnerTest.scala:218)
    minitest.SimpleTestSuite.$anonfun$test$1(SimpleTestSuite.scala:27)
    minitest.api.TestSpec$.$anonfun$sync$1(TestSpec.scala:51)
    minitest.api.TestSpec.apply(TestSpec.scala:27)
    minitest.api.Properties.$anonfun$iterator$2(Properties.scala:38)
    minitest.api.TestSpec.apply(TestSpec.scala:27)
    minitest.runner.Task.loop$1(Task.scala:56)
    minitest.runner.Task.$anonfun$execute$1(Task.scala:63)
    scala.concurrent.Future.$anonfun$flatMap$1(Future.scala:307)
    scala.concurrent.impl.Promise.$anonfun$transformWith$1(Promise.scala:41)
    scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
    java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1402)
[...]

PR #5876, opened 2020-09-21
sbt.cmd -> sbt.bat and --sbt-script update

eed3si9n:
Ref https://github.com/sbt/sbt/issues/5870

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -998,6 +998,9 @@ object NetworkClient {
           completionArguments += a
         case a if a.startsWith("--sbt-script=") =>
           sbtScript = a.split("--sbt-script=").lastOption.getOrElse(sbtScript)
+        case "--sbt-script" if i + 1 < sanitized.length =>
+          i += 1
+          sbtScript = sanitized(i)
         case a if !a.startsWith("-") => commandArgs += a
         case a @ SysProp(key, value) =>
           System.setProperty(key, value)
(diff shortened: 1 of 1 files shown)


eatkins:
This will break sbtn on my windows vm. I think possibly a better fix would be to scan the PATH environment variable for sbt.bat or sbt.cmd.

eed3si9n:
As proposed in https://github.com/sbt/sbt/issues/5872, ideally I'd like `sbtn` to look for `sbt` (or `sbt.bat`, `sbt.cmd` I guess) within the same directory first. Is there an equivalent of `$0`?

eatkins:
Aah, that's an interesting idea. I don't know that there is an easy equivalent of $0. What I would propose is that we do something similar to what I did to fix the missing powershell executable issue in the directories project: https://github.com/dirs-dev/directories-jvm/pull/40/files
We first scan the path to see if the executables we are looking for exist. If we fail to find it on the path, we fall back to looking in the system wide default locations. For example, on mac we'd look for /usr/local/bin/sbt as the fallback. On windows, it would probably be a few `C:\Program Files\...\` like directories. It's not the most elegant solution but would probably meet like 99% of use cases.

eed3si9n:
Also where did `.bat` vs `.cmd` change come from? Should we switch the official distribution to use `.cmd`?

eatkins:
Good question. I have no idea why I have an sbt.bat on my vm. I have both choco and scoop on my vm and I'm not sure which one is being used.

eatkins:
I have `C:\Program Files (x86)\sbt\bin` on my path and it contains sbt.bat. It was updated in june, so either choco or scoop is installing an sbt.bat.

eed3si9n:
`sbt.bat` is what the official installer installs. But NetworkClient.scala is looking for `sbt.cmd`?

eatkins:
Oh, shoot, you're right. sbt.bat does work on my vm.

eatkins (approved):
LGTM

PR #5879, opened 2020-09-21
Abort thin client on access denied exception

eatkins:
There are situations in windows where it is possible that a client
attempts to connect to a portfile that it did not have access to. This
can happen if, for example, the server is started in administrator mode
but the client is started as a regular user. When this happens, the
client would try to remove the portfile and start a new server. This new
server would not actually be able to start a server though becuase it
would not be able to open the named point because the other server had
it open. As a result, the client would just hang. The fix is to just
abort the thin client if it gets an access denied exception.


eatkins:
I think this might fix https://github.com/sbt/sbt/issues/5870 or at least is possibly related.

eed3si9n (approved):

PR #5888, opened 2020-09-23
reply to BSP shutdown request

### LINKED ISSUE #5684: BSP: server does not reply to shutdown request
## steps

1. start sbt server
2. send `build/shutdown` request from client
3. server does not respond

## problem

No sreponse to shutdown request. This can cause a problem for clients that wait for the request to complete.

## expectation

Server should respond that it is ready to shutdown (possibly after finishing running tasks)

adpi2:
Fixes #5684


eed3si9n (approved):

unkarjedy:
@adpi2 So the patch only sends a dummy response but actually does not shut down the server?
See https://github.com/sbt/sbt/issues/6723

Issue #5891, opened 2020-09-23
serverTestProj/test is getting flaky?


eed3si9n:
## steps

Not sure what but Travis CI job has been getting really flaky.

## problem

! 7058.1

```
[info] Here are some highlights of this release:
[info]   - Build server protocol (BSP) support
[info]   - sbtn: a native thin client for sbt
[info]   - VirtualFile + RemoteCache: caches build artifacts across different machines
[info]   - ThisBuild / versionScheme to take the guessing out of eviction warning
[info] See http://eed3si9n.com/sbt-1.4.0-beta for full release notes.
[info] Hide the banner for this release by running `skipBanner`.
[info] sbt server started at local:///home/travis/.sbt/1.0/server/cfd9b8fa94297dc5b51d/sock
[build-1] waiting for the server...
[build-1] waiting for the server...
[build-1] waiting for the server...
[build-1] waiting for the server...
[error] Uncaught exception when running testpkg.ServerCompletionsTest: java.lang.RuntimeException: Timeout. /home/travis/build/sbt/sbt/target/test-server/server-test648017339595567195/completions/project/target/active.json is not found.
[error] sbt.ForkMain$ForkError: java.lang.RuntimeException: Timeout. /home/travis/build/sbt/sbt/target/test-server/server-test648017339595567195/completions/project/target/active.json is not found.
[error] 	at scala.sys.package$.error(package.scala:30)
[error] 	at testpkg.TestServer.waitForPortfile(TestServer.scala:189)
[error] 	at testpkg.TestServer.<init>(TestServer.scala:194)
[error] 	at testpkg.TestServer$.get(TestServer.scala:89)
[error] 	at testpkg.AbstractServerTest.setupSuite(TestServer.scala:58)
[error] 	at testpkg.AbstractServerTest.setupSuite$(TestServer.scala:40)
[error] 	at testpkg.ServerCompletionsTest$.setupSuite(ServerCompletionsTest.scala:13)
[error] 	at verify.TestSuite.$anonfun$properties$3(TestSuite.scala:42)
[error] 	at verify.runner.Task.$anonfun$execute$3(Task.scala:65)
[error] 	at scala.Option.fold(Option.scala:251)
[error] 	at verify.runner.Task.execute(Task.scala:63)
[error] 	at verify.runner.Task.execute(Task.scala:76)
[error] 	at sbt.ForkMain$Run.lambda$runTest$1(ForkMain.java:413)
[...]

PR #5893, opened 2020-09-24
sbt 1.4.0-RC2 + pack flaky-behaving tests together

eed3si9n:
Ref https://github.com/sbt/sbt/issues/5891


eatkins:
I have made progress on debugging what is going on with serverTestProj. Are the nio tests actually flaky or are they just triggering the bug where scripted never returns after the last test completes?

eed3si9n:
> Are the nio tests actually flaky or are they just triggering the bug where scripted never returns after the last test completes?

I don't know the cause for that is, but I know that third Travis CI job often gets stuck.

eatkins:
Ok. I think I'm close to fixing the flakiness with serverTestProj. It's possible that these fixes will also address the scripted issues.

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/5905

Issue #5897, opened 2020-09-25
1.4.0-RC2: Not seeing log to STOUT of a running application


GuntherW:
## steps

When I start a http4s application in sbt console via "reStart", I can't see any stout output.
With sbt 1.3.13, the output was visible.
It is reproducable with this tiny http4s application:
https://github.com/GuntherW/Http4sTapirWithAuth
Just clone it, start sbt and run `reStart` in sbt shell.
Retry with sbt 1.3.13 and see the difference.

## problem

I don't the any logs from http4s app in my sbt console, which are targeted to stout

## expectation

I expect the same behaviour, as  sbt 1.3.13

eatkins:
Thanks for the report! I can reproduce and the root cause is clear. There was an assumption that tasks would not be printing output after the sbt process had returned to the shell which is obviously wrong for reStart. I should have a fix soon.

eed3si9n:
This was fixed in https://github.com/sbt/sbt/pull/5898

PR #5898, opened 2020-09-25
Clear prompt only if there are no bytes pending

eatkins:
When sbt is running a background process that logs to stdout, the output
can be inadvertently deleted before it has been printed. When the user
is in the prompt state and a log message comes in, we want to delete the
prompt before we print the log. The problem is println is often
implemented with a write of the content followed by a second write of
the system line separator. When that happened, we would print the
content and then immediately delete it when the newline came in. The fix
is to not clear the prompt if there are any bytes that have been written
without a newline, which was tracked by the currentLineBytes variable.


eatkins:
Fixes #5897

Issue #5899, opened 2020-09-25
[1.4.0-RC2] sbt-site's `previewSite` can't be stopped by pressing a key


ennru:
## steps

Many Akka projects use sbt-site to bring together documentation generated by unidoc and Paradox.

Eg. Alpakka Kafka https://github.com/akka/alpakka-kafka

1. updated sbt version in build propertiesd
1. comment out "sbt-paradox-dependencies" in plugins.sbt (see #5894)
1. sbt
1. sbt> docs/previewSite

## problem

sbt-site's `previewSite` task waits until the user presses any key. With sbt 1.4.0-RC2 the keypress is not detected. It can't be stopped by Ctrl-C either.

The current solution in sbt-site

https://github.com/sbt/sbt-site/blob/5aab1d9a2be6d270d9bee1637c503d8a831a88b4/src/main/scala/com/typesafe/sbt/site/SitePreviewPlugin.scala#L31-L37

eed3si9n:
If pressing Enter is acceptable, then the following works on my machine:

```diff
diff --git a/src/main/scala/com/typesafe/sbt/site/SitePreviewPlugin.scala b/src/main/scala/com/typesafe/sbt/site/SitePreviewPlugin.scala
index efe043d..b6435ea 100644
--- a/src/main/scala/com/typesafe/sbt/site/SitePreviewPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/site/SitePreviewPlugin.scala
@@ -24,20 +24,14 @@ object SitePreviewPlugin extends AutoPlugin {
       val portOption = previewFixedPort.value
       val browser = previewLaunchBrowser.value
       val path = previewPath.value
+      val ui = interactionService.value

       val port = portOption getOrElse Port.any
       val server = createServer(file, port) start()
-      val sLog = streams.value.log
-      sLog.info("SitePreviewPlugin server started on port %d. Press any key to exit." format port)
-      // TODO: use something from sbt-web?
-      @annotation.tailrec def waitForKey(): Unit = {
-        try { Thread sleep 500 } catch { case _: InterruptedException => () }
-        if(System.in.available <= 0)
-          waitForKey()
-      }
       if(browser)
         Browser open ("http://localhost:%d/%s".format(port, path))
-      waitForKey()
+      ui.readLine(s"SitePreviewPlugin server started on port $port. Press enter to exit. ", false)
```

[...]

PR #5904, opened 2020-09-27
Add `terminal` task

eatkins:
This PR adds a terminal task so that users can access and modify some properties of the terminal that initiated the task. There is some overlap with interactionService but it wasn't clear to me that they should be combined. The `terminal` tasks is a bit more raw. It doesn't provide a `readLine` method or anything like that. Instead it is designed to be just powerful enough for advanced users to build a custom line reader without exposing the entire jline api. The `interactionService` is still probably the way to go for most use cases.

I'd been thinking about adding something like this for a while. The motivation for actually doing it was @eed3si9n's comment in #5899 (https://github.com/sbt/sbt/issues/5899#issuecomment-699467667).


eatkins:
The main feature that this adds compared to the interactionService that is easy to implement is press any key to exit. Here is a sample build.sbt that does this:
```
TaskKey[Unit]("readInput") := {                                     
  terminal.value.setMode(canonical = false, echo = false)
  terminal.value.printStream.println("read a byte")
  terminal.value.inputStream.read                                   
} 
```

The task exits when the user types any key both with sbt and sbtn.

main/src/main/scala/sbt/Terminal.scala
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Sets the mode of the terminal. By default,the terminal will be in canonical mode
+   * with echo enabled. This means that the terminal's inputStream will not return any
+   * bytes until a newline is received and that all of the characters inputed by the
+   * user will be echoed to the terminal's output stream.
+   *
+   * @param canonical toggles whether or not the terminal input stream is line buffered
+   * @param echo toggles whether or not to echo the characters received from the terminal input stream
+   */
+  def setMode(canonical: Boolean, echo: Boolean): Unit
```
eed3si9n, line 61: Nice documentation here.

eed3si9n (approved):

eatkins:
I'm sort of wondering now if `Terminal` should be in the sbt package. Is there another package that might make sense that is not internal? The goal was to make this limited functionality public and not in an internal package but top level sbt might be promoting it a little too much. I also wouldn't have had to alias the internal Terminal everywhere it was used in the main project  if it was in a different package.

eed3si9n:
You could make up some subpackage under sbt if you want, but not sure if it's worth it for one class.

PR #5905, opened 2020-09-27
Fix many instances of server test hangs

eatkins:
The server tests often hang or fail in ci. There were a number of small issues that were causing this behavior. What I did to track it down was rewrite .travis.yml so that it ran `serverTestProj/test` 6 times per build. As a result, it would run the server tests 30 times each time I pushed to github. I added lots of println debugging to see what was going on right before and after the hangs. I also added a background thread that would print a stacktrace for all of the threads on the jvm every few minutes. I don't think these changes make the tests 100% reliable, but they are far more reliable than they were before.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ matrix:
 before_install:
   - curl -sL https://raw.githubusercontent.com/shyiko/jabba/0.11.0/install.sh | bash && . ~/.jabba/jabba.sh
   - if [ $SBT_LOCAL == true ]; then sbt -Dsbt.io.virtual=false publishLocalBin; fi
+  - rm -r $(find $HOME/.sbt/boot -name "*-SNAPSHOT") || true
 
 install:
   - $JABBA_HOME/bin/jabba install $TRAVIS_JDK
@@ -45,6 +46,7 @@ before_cache:
   - find $HOME/.cache/coursier/v1 -name "ivydata-*.properties" -delete
   - find $HOME/.ivy2              -name "ivydata-*.properties" -delete
   - find $HOME/.sbt               -name "*.lock"               -delete
+  - rm -r $(find $HOME/.sbt/boot -name "*-SNAPSHOT") || true
 
 cache:
   directories:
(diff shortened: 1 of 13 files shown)


main/src/main/scala/sbt/internal/CommandExchange.scala
```diff
@@ -376,7 +377,7 @@ private[sbt] final class CommandExchange {
           .withChannelName(currentExec.flatMap(_.source.map(_.channelName)))
       case _ => pe
     }
-    if (channels.isEmpty) addConsoleChannel()
+    if (channels.isEmpty && !Terminal.isCI) addConsoleChannel()
```
eed3si9n: Don't we need consoles to output log messages even while we're on CI?

eed3si9n: In general the use of `isCI` makes me a bit nervous because it could become blind spot for testing.

.travis.yml
```diff
@@ -30,6 +30,7 @@ matrix:
 before_install:
   - curl -sL https://raw.githubusercontent.com/shyiko/jabba/0.11.0/install.sh | bash && . ~/.jabba/jabba.sh
   - if [ $SBT_LOCAL == true ]; then sbt -Dsbt.io.virtual=false publishLocalBin; fi
+  - rm -r $(find $HOME/.sbt -name "*-SNAPSHOT") || true
```
eed3si9n: > If the sbt snapshot builds are cached, then we don't actually dogfood
sbt when we publish a local sbt version because sbt will load with the
version from the last build

😱

eed3si9n:
> What I did to track it down was rewrite `.travis.yml` so that it ran `serverTestProj/test` 6 times per build.

Thanks for this!

.travis.yml
eatkins: This should be `$HOME/.sbt/boot`.

main/src/main/scala/sbt/internal/CommandExchange.scala
eatkins: Naively you would think that but having a console channel or not does not affect whether or not System.out is available which is what is necessary for output logs. That being said, I'm not 100% sure this change is necessary so I'll see if I get hangs if I revert it.

eatkins: No hangs on my personal travis ci after reversion.

eed3si9n (approved):
LGTM pending Travis CI passage.

Issue #5643, opened 2020-06-23
package/lazy-name scripted test periodically hangs on scripted


eatkins:
I've seen a number of builds fail because the package/lazy-name scripted test has hung:
https://travis-ci.org/github/sbt/sbt/jobs/701398954
https://travis-ci.org/github/sbt/sbt/jobs/701138531
https://travis-ci.org/github/sbt/sbt/jobs/701409367

Interestingly, the test is headed with the following comment:
```
# 'lazy-name' is not an appropriate name anymore
# This test verifies that package uses the updated name
# after changing the version.  It no longer serves
# much purpose other than checking that the 'set' command
# re-evaluates the project data.
```
Given that, I'm tempted to disable it, but, looking at the test, it isn't obvious why it is hanging.

eed3si9n:
Note that `package/lazy-name` is the last scripted test for the job.
I wonder if there's a leak somewhere that goes into really long GC, or something gets stuck during the shutdown hook?

A successful run looks like this:

```
[info] Running package/lazy-name
[info] [info] shutting down sbt server
[error] 20/06/24 04:55:53 INFO ShutdownHookManager: Shutdown hook called
[error] 20/06/24 04:55:53 INFO ShutdownHookManager: Deleting directory /tmp/spark-d496921c-8a47-4f0b-9983-732a25b6704c
[error] 20/06/24 04:55:53 INFO ShutdownHookManager: Shutdown hook called
[error] 20/06/24 04:55:53 INFO ShutdownHookManager: Deleting directory /tmp/spark-23a5a9cb-a8cf-40bc-a0f8-da22f49ed1b2
[error] 20/06/24 04:55:53 INFO ShutdownHookManager: Shutdown hook called
[error] 20/06/24 04:55:53 INFO ShutdownHookManager: Deleting directory /tmp/spark-98b781e1-38d6-44d2-a296-a5ef9e3ad17b
[success] Total time: 916 s (15:16), completed Jun 24, 2020, 4:55:55 AM
```

eatkins:
Aah. Nice catch even if it doesn't really help us track down the resource leak.

PR #5910, opened 2020-09-29
Don't block sbt exit forever on bg service shutdown

eatkins:
Some of the sbt scripted tests somewhat frequently hang in CI. I added a
patch that printed a stack trace of the sbt process every 30 seconds. I
discovered that the main thread was stuck in DefaultBackgroundJobService
shutdown. To avoid the hangs, I updated the awaitTermination methods to
take a timeout parameter and we timeout shutdown if 10 seconds have
elapsed.


eatkins:
Should hopefully fix #5643

eed3si9n (approved):
Thanks!

eatkins:
Just wanted to follow up that I updated the .travis.yml to run the build that does the `nio/*` scripted tests and ran it 6 times with no hangs. That meant that the tests ran 30 times. Before making this change, 2 of the 5 builds in the first run hung. I hadn't confirmed that it fixed the hangs before I pushed the PR but I'm more confident now.

PR #5914, opened 2020-10-02
Rename some sbt-dependency-graph keys

### LINKED ISSUE #5894: 1.4.0-RC2: cannot load a plugin that depends on sbt-dependency-graph transitively
## steps

Depend on a plugin that depends on sbt-dependency-graph transitively (e.g. https://github.com/lightbend/sbt-paradox-dependencies)

## problem

```
[error] Some keys were defined with the same name but different types: 'moduleGraphStore' (sbt.Task[net.virtualvoid.sbt.graph.ModuleGraph], sbt.Task[sbt.internal.graph.ModuleGraph])
[warn] Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)
```

From this error, it's not to clear what's going on, and even if you are aware of the issue, it's not so easy to see *which* sbt plugin has this transitive dependency.

## notes

This is a consequence of folding sbt-dependency-graph into sbt 1.4.0.

It might be fine (there are probably few plugins that have this transitive dependency, and for those it's probably not a problem to introduce a new version where the transitive dependency is dropped), but I wanted to create this issue to make sure we're aware of it and it can come up in a web search for someone who runs into this problem :)

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/5894

### DIFF
--- a/main/src/main/scala/sbt/plugins/DependencyTreeKeys.scala
+++ b/main/src/main/scala/sbt/plugins/DependencyTreeKeys.scala
@@ -24,13 +24,12 @@ trait MiniDependencyTreeKeys {
   val toFile = inputKey[File]("Writes the task value to the given file")
 
   // internal
-  private[sbt] val ignoreMissingUpdate =
-    TaskKey[UpdateReport]("dependencyUpdate", "sbt-dependency-graph version of update")
-  private[sbt] val moduleGraphStore =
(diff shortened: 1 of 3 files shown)


SethTisue:
even not considering the name clash issue, this is an improvement :+1:

PR #5918, opened 2020-10-03
Hard code sbt plugins to compile against sbt 1.2.8

### LINKED ISSUE #5049: [1.3.0] sbt plugin does not work with 1.2.8 if plugin build with 1.3.0
## workaround by eed3si9n

```scala
pluginCrossBuild / sbtVersion := "1.2.8"
```

### steps

sbt version: 1.3.0

- https://travis-ci.org/xuwei-k/sbt-1.2.8-1.3.0-plugin-compatibility/builds/581604976
- https://github.com/xuwei-k/sbt-1.2.8-1.3.0-plugin-compatibility/commit/e4977ecccce91b5c878b0396cbce118300a54999

#### build.sbt

```scala
enablePlugins(SbtPlugin)

organization := "com.example"

name := "example-plugin"

scriptedBufferLog := false
```

#### project/build.properties

```
sbt.version=1.3.0
```

#### src/main/scala/example/ExamplePlugin.scala

```scala
package example

import sbt._, Keys._

object ExamplePlugin extends AutoPlugin {
  object autoImport {
    val foo = settingKey[File]("")
  }
}
```

#### src/sbt-test/a/b/build.sbt

```scala
enablePlugins(ExamplePlugin)

foo := file("bar")
```

#### src/sbt-test/a/b/project/build.properties

```
sbt.version=1.2.8
```

#### src/sbt-test/a/b/project/plugins.sbt

```scala
addSbtPlugin("com.example" % "example-plugin" % "0.1.0-SNAPSHOT")
```

#### src/sbt-test/a/b/test

```
> compile
```

run `sbt scripted`

### problem

<details>

<summary>

Stack trace of

```
[info] [error] java.lang.NoSuchMethodError: sbt.package$.singleFileJsonFormatter()Lsjsonnew/JsonFormat;
[info] [error] 	at example.ExamplePlugin$autoImport$.<init>(ExamplePlugin.scala:7)
```

</summary>

```
[info] [error] java.lang.NoSuchMethodError: sbt.package$.singleFileJsonFormatter()Lsjsonnew/JsonFormat;
[info] [error] 	at example.ExamplePlugin$autoImport$.<init>(ExamplePlugin.scala:7)
[info] [error] 	at example.ExamplePlugin$autoImport$.<clinit>(ExamplePlugin.scala)
[info] [error] 	at $ccf3da47bd1d87673069$.$sbtdef(/tmp/sbt_98bd2a58/b/build.sbt:3)
[info] [error] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[info] [error] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[info] [error] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[info] [error] 	at java.lang.reflect.Method.invoke(Method.java:498)
[info] [error] 	at sbt.compiler.Eval$.getValue(Eval.scala:578)
[info] [error] 	at sbt.compiler.Eval.$anonfun$eval$1(Eval.scala:129)
[info] [error] 	at sbt.internal.EvaluateConfigurations$.$anonfun$evaluateDslEntry$1(EvaluateConfigurations.scala:249)
[info] [error] 	at sbt.internal.EvaluateConfigurations$.$anonfun$evaluateSbtFile$6(EvaluateConfigurations.scala:172)
[info] [error] 	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:233)
[info] [error] 	at scala.collection.immutable.List.foreach(List.scala:388)[0m
[info] [error] 	at scala.collection.TraversableLike.map(TraversableLike.scala:233)
[info] [error] 	at scala.collection.TraversableLike.map$(TraversableLike.scala:226)
[info] [error] 	at scala.collection.immutable.List.map(List.scala:294)
[info] [error] 	at sbt.internal.EvaluateConfigurations$.$anonfun$evaluateSbtFile$4(EvaluateConfigurations.scala:172)
[info] [error] 	at sbt.internal.Load$.loadSettingsFile$1(Load.scala:1137)
[info] [error] 	at sbt.internal.Load$.$anonfun$discoverProjects$2(Load.scala:1144)
[info] [error] 	at scala.collection.MapLike.getOrElse(MapLike.scala:127)
[info] [error] 	at scala.collection.MapLike.getOrElse$(MapLike.scala:125)
[info] [error] 	at scala.collection.AbstractMap.getOrElse(Map.scala:59)
[info] [error] 	at sbt.internal.Load$.memoLoadSettingsFile$1(Load.scala:1143)
[info] [error] 	at sbt.internal.Load$.$anonfun$discoverProjects$4(Load.scala:1151)
[info] [error] 	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:233)
[info] [error] 	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:58)
[info] [error] 	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:51)
[info] [error] 	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:47)
[info] [error] 	at scala.collection.TraversableLike.map(TraversableLike.scala:233)
[info] [error] 	at scala.collection.TraversableLike.map$(TraversableLike.scala:226)
[info] [error] 	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
[info] [error] 	at sbt.internal.Load$.loadFiles$1(Load.scala:1151)
[info] [error] 	at sbt.internal.Load$.discoverProjects(Load.scala:1165)
[info] [error] 	at sbt.internal.Load$.discover$1(Load.scala:862)
[info] [error] 	at sbt.internal.Load$.loadTransitive(Load.scala:937)
[info] [error] 	at sbt.internal.Load$.loadProjects$1(Load.scala:726)
[info] [error] 	at sbt.internal.Load$.$anonfun$loadUnit$11(Load.scala:729)
[info] [error] 	at sbt.internal.Load$.timed(Load.scala:1395)
[info] [error] 	at sbt.internal.Load$.$anonfun$loadUnit$1(Load.scala:729)
[info] [error] 	at sbt.internal.Load$.timed(Load.scala:1395)
[info] [error] 	at sbt.internal.Load$.loadUnit(Load.scala:688)
[info] [error] 	at sbt.internal.Load$.$anonfun$builtinLoader$4(Load.scala:484)
[info] [error] 	at sbt.internal.BuildLoader$.$anonfun$componentLoader$5(BuildLoader.scala:176)
[info] [error] 	at sbt.internal.BuildLoader.apply(BuildLoader.scala:241)
[info] [error] 	at sbt.internal.Load$.loadURI$1(Load.scala:546)
[info] [error] 	at sbt.internal.Load$.loadAll(Load.scala:562)
[info] [error] 	at sbt.internal.Load$.loadURI(Load.scala:492)
[info] [error] 	at sbt.internal.Load$.load(Load.scala:471)
[info] [error] 	at sbt.internal.Load$.$anonfun$apply$1(Load.scala:251)
[info] [error] 	at sbt.internal.Load$.timed(Load.scala:1395)
[info] [error] 	at sbt.internal.Load$.apply(Load.scala:251)
[info] [error] 	at sbt.internal.Load$.defaultLoad(Load.scala:69)
[info] [error] 	at sbt.BuiltinCommands$.liftedTree1$1(Main.scala:829)
[info] [error] 	at sbt.BuiltinCommands$.doLoadProject(Main.scala:829)
[info] [error] 	at sbt.BuiltinCommands$.$anonfun$loadProjectImpl$2(Main.scala:800)
[info] [error] 	at sbt.Command$.$anonfun$applyEffect$4(Command.scala:142)
[info] [error] 	at sbt.Command$.$anonfun$applyEffect$2(Command.scala:137)
[info] [error] 	at sbt.Command$.process(Command.scala:181)
[info] [error] 	at sbt.MainLoop$.processCommand(MainLoop.scala:151)
[info] [error] 	at sbt.MainLoop$.$anonfun$next$2(MainLoop.scala:139)
[info] [error] 	at sbt.State$$anon$1.runCmd$1(State.scala:246)
[info] [error] 	at sbt.State$$anon$1.process(State.scala:250)
[info] [error] 	at sbt.MainLoop$.$anonfun$next$1(MainLoop.scala:139)
[info] [error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:16)
[info] [error] 	at sbt.MainLoop$.next(MainLoop.scala:139)
[info] [error] 	at sbt.MainLoop$.run(MainLoop.scala:132)
[info] [error] 	at sbt.MainLoop$.$anonfun$runWithNewLog$1(MainLoop.scala:110)
[info] [error] 	at sbt.io.Using.apply(Using.scala:22)
[info] [error] 	at sbt.MainLoop$.runWithNewLog(MainLoop.scala:104)
[info] [error] 	at sbt.MainLoop$.runAndClearLast(MainLoop.scala:59)
[info] [error] 	at sbt.MainLoop$.runLoggedLoop(MainLoop.scala:44)
[info] [error] 	at sbt.MainLoop$.runLogged(MainLoop.scala:35)
[info] [error] 	at sbt.StandardMain$.runManaged(Main.scala:138)
[info] [error] 	at sbt.xMain.run(Main.scala:89)
[info] [error] 	at xsbt.boot.Launch$$anonfun$run$1.apply(Launch.scala:111)
[info] [error] 	at xsbt.boot.Launch$.withContextLoader(Launch.scala:130)
[info] [error] 	at xsbt.boot.Launch$.run(Launch.scala:111)
[info] [error] 	at xsbt.boot.Launch$$anonfun$apply$1.apply(Launch.scala:37)
[info] [error] 	at xsbt.boot.Launch$.launch(Launch.scala:119)
[info] [error] 	at xsbt.boot.Launch$.apply(Launch.scala:20)
[info] [error] 	at xsbt.boot.Boot$.runImpl(Boot.scala:56)
[info] [error] 	at xsbt.boot.Boot$.main(Boot.scala:18)
[info] [error] 	at xsbt.boot.Boot.main(Boot.scala)
[info] [error] java.lang.NoSuchMethodError: sbt.package$.singleFileJsonFormatter()Lsjsonnew/JsonFormat;
[info] [error] Use 'last' for the full log.
```

</details>

### expectation

it should work

### notes

https://github.com/xerial/sbt-sonatype/issues/91

the `settingKey[T]` macro take `OptJsonWriter[T]` parameter.
- https://github.com/sbt/sbt/blob/v1.3.0/main-settings/src/main/scala/sbt/std/KeyMacro.scala#L42
- https://github.com/sbt/sbt/blob/v1.3.0/main-settings/src/main/scala/sbt/std/KeyMacro.scala#L21

`sbt.singleFileJsonFormatter` added since sbt 1.3.0-M3.
https://github.com/sbt/sbt/commit/33194233698ab49682d89091a332edb808cb75bc#diff-cafb18f4282f4fff753f9efb35169d46R41

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/5049

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -293,7 +293,7 @@ object Defaults extends BuildCommon {
       // `pluginCrossBuild` scoping is based on sbt-cross-building plugin.
       // The idea here is to be able to define a `sbtVersion in pluginCrossBuild`, which
       // directs the dependencies of the plugin to build to the specified sbt plugin version.
-      sbtVersion in pluginCrossBuild := sbtVersion.value,
+      sbtVersion in pluginCrossBuild := "1.2.8", // sbtVersion.value,
       onLoad := idFun[State],
       onUnload := idFun[State],
       onUnload := { s =>


eatkins:
Although we do try to preserve backward compatibility, I am wondering if pinning the plugin compatibility to 1.2.8 is ideal. Are there known issues that prevent a non-negligible subset of builds from upgrading? My naive assumption is that coursier is the thing most likely to make builds incompatible but it is easily disabled. Soon 1.2.8 will be two minor versions behind and it seems like it would be better to encourage users to upgrade their sbt version if they need to upgrade a plugin that is no longer compatible with 1.2.8. Ultimately given the improvements we’ve made since 1.2.x, anything that encourages users to stay behind feels like a disservice even if it may seem more convenient in the short term.

That all being said, there is a good argument for building against a stable api. My suggestion would be to make 1.3.13 the stable version for plugin compilation rather than 1.2.8 if we go this route.

eed3si9n:
The promise of Semantic Versioning is that in theory a plugin built against sbt 1.0.0 should still work today.
I think all of library and app sbt 1.x builds should go to our latest and greatest, but I simultaneously want to make sure the ecosystem has the largest degree of freedom in terms of the plugin to choose from. Often there isn't a huge reason for plugins to use our latest feature series, and if they do that can be done with an explicit setting.

eed3si9n:
Our scripted test, however, should run 1.4.x though :)

eatkins:
The target default plugin api level seems related to but not exactly the same as semantic versioning. There are problems with pinning the target default level to an older version. There is documentation for apis that are not available in 1.2.8. A user could try to extract working code from a build.sbt into a plugin and all of sudden they would have compilation errors for the missing apis. Perhaps we could add logging when compiling a plugin that explicitly prints the api level? For instance, it might look like:
```
> compile
[info] compiling sbt plugin foo with target level 1.4.0
[info] compiling 2 scala sources...
```
That might be enough to prevent users from inadvertently publishing plugins with an unexpected api level which is the underlying issue in #5049.

eatkins:
Ideally we keep moving the plugin ecosystem forward with improvements to sbt while not breaking existing plugins, which semantic versioning and mima help us with.

eed3si9n:
Like I wrote in https://github.com/sbt/sbt/issues/5049#issuecomment-670826519 this would allow plugin authors to be the early adopter of cutting edge while 1.4.x series is young, but publish plugins where general population is at (1.2.x ~ 1.3.x). Every time a feature version comes out there's a period of confusion that happens because someone would publish a plugin with it before majority is using it. One way to try to solve this may be advertise

```scala
pluginCrossBuild / sbtVersion := "1.2.8"
```

more but I figured, if I'm gonna do that might as do that programmatically. I'm fine with punting this off to 1.5.x tho.

eatkins:
For sure it’s a real problem. That was why I thought the log message might be a reasonable approach. I can’t speak for all plugin authors but I would notice that log message and probably pin the project by hand to whatever api level I thought was appropriate.

If we were to programmatically pin the version, I like one version behind better than two but it isn’t that important. It also really depends on the release. sbt 1.3.0 had a lot more significant api changes than 1.4.0.

dwijnand:
In my eyes the "problem" that happened with 1.3.0 was the implicit(s) added to `import sbt._`, which is pretty much sbt's predef.  IMO that shouldn't be done any more (until sbt 2).  And if that happens then all future minor releases will be fine (like 1.1.0 and 1.2.0 releases were fine).

But if we are going to thinking about this, here are the different build/testing strategies (I'll use 1.2.8 as Eugene is proposing here and 1.4.0 as latest, using sbt.version=1.4.0):
1. build against 1.4.0, test with 1.4.0
2. build against 1.2.8, test with 1.2.8
3. build against 1.2.8, test with 1.4.0
4. build against 1.4.0, test with 1.2.8

(1) is what happens out the box today if you upgrade to 1.4.0.  (2) is what Eugene is proposing with this PR as it currently stands.  But I think (3) is what should be happening, as it's rehearsing history: I built my plugin with 1.2.8, does it work for a user using the latest version 1.4.0.  It's at the risk of introducing a change to the plugin that for some reason behaves incorrectly on the older version, but I just doubt that would happen and I think it's better to bias to the latest version and keep testing/CI complexity down.  I've also experimented with (4) by having a 1-off smoke test to make sure that the minimum functionality of my plugin works on an older version (running the whole test suite on latest) - but I can't find where I did that, maybe it was in Play or Lagom...

So what I'm saying is I think it might be a good idea to pin this to a specific version (but I think by now we should consider 1.3 to be canon, rather than 1.2) but I think `scriptedSbt` (I think it's that key) should be re-set to sbt.version (the launcher app version) so that it keeps running on the latest.

More thoughts: probably would be good to add some of these thoughts and tips to a docs page.

And finally: I'm also thinking that this version (much like `scalaVersion`) while there's good reason to not error because it's unset, it could just be annoyingly chatty at runtime, requiring the user explicitly sets the build target version (and maybe even the test target version) in the build (much like we teach users should do for `scalaVersion`).

dwijnand:
> I've also experimented with (4) by having a 1-off smoke test to make sure that the minimum functionality of my plugin works on an older version (running the whole test suite on latest) - but I can't find where I did that, maybe it was in Play or Lagom...

Found it: https://github.com/lightbend/mima/blob/c3a884d783b324e632b72e568095b3bf6dedf5cb/.travis.yml#L3

```
'set sbtplugin/scriptedSbt := \"1.2.8\"' 'scripted sbt-mima-plugin/minimal'
```

eed3si9n:
Closing this for now.

Issue #5938, opened 2020-10-06
Hangs when project loading failure in batch mode


mayth:
## steps

prepare an empty project with an invalid build.sbt:

```
$ tree
.
├── build.sbt
└── project
    └── build.properties

1 directory, 2 files
$ cat build.sbt
foo
$ cat project/build.properties
sbt.version=1.4.0
```

start sbt in batch mode.

```
$ sbt --batch
[info] welcome to sbt 1.4.0 (AdoptOpenJDK Java 11.0.8)
[info] loading project definition from /Users/***/work/project
/Users/***/work/build.sbt:1: error: not found: value foo
foo
^
[error] Type error in expression
[warn] Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)
```

* OS: macOS Catalina 10.15.6
* Java: AdoptOpenJDK 11.0.8+10
* sbt: 1.4.0
  * installed version: 1.3.13 (via Homebrew)

## problem

sbt hangs after showing `[warn] Project loading failed: ...`, needs Ctrl-C to exit.

## expectation

Exits immediately.

The same empty project but using sbt 1.3.13 works expectedly.

```
$ echo 'sbt.version=1.3.13' > project/build.properties
$ sbt --batch
[info] welcome to sbt 1.3.13 (AdoptOpenJDK Java 11.0.8)
[info] loading project definition from /Users/***/work/project
/Users/***/work/build.sbt:1: error: not found: value foo
foo
^
[error] Type error in expression
Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore?
$ (returned to the shell)
```

## notes

In same project, sometimes NPE is thrown. I'm not sure if this is related to the issue, so just for your reference.

```
...
[error] Type error in expression
[warn] Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)
Exception in thread "Thread-2" java.lang.NullPointerException
	at sbt.internal.util.Terminal$proxyInputStream$ReadThread.run(Terminal.scala:612)
```

eed3si9n:
@mayth Thanks for the bug report!

eed3si9n:
This was fixed in https://github.com/sbt/sbt/pull/5945 by @eatkins

PR #5939, opened 2020-10-06
Initialize value before starting thread

eatkins:
Depending on how quickly the thread starts up, this can cause an NPE
(see #5938).

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
@@ -608,9 +608,9 @@ object Terminal {
      */
     private class ReadThread extends Thread with AutoCloseable {
       val result = new LinkedBlockingQueue[Integer]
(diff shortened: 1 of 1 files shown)


internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
```diff
@@ -606,9 +606,9 @@ object Terminal {
      */
     private class ReadThread extends Thread with AutoCloseable {
       val result = new LinkedBlockingQueue[Integer]
+      val running = new AtomicBoolean(true)
       setDaemon(true)
       start()
-      val running = new AtomicBoolean(true)
```
eed3si9n, line 613: LGTM

eed3si9n:
@eatkins I'm guessing out of 8 commits only 1 is intended to be merged?

eatkins:
Correct. I deleted all the testing commits I inadvertently pushed.

eed3si9n (approved):

PR #5944, opened 2020-10-06
Bump sbt snapshot version

eatkins:


build.sbt
```diff
@@ -8,7 +8,7 @@ import java.nio.file.{ Files, Path => JPath }
 import scala.util.Try
 
 ThisBuild / version := {
-  val v = "1.4.0-SNAPSHOT"
+  val v = "1.5.0-SNAPSHOT"
```
eed3si9n: Do we want to branch out to 1.4.x? Unless we want to start merging 1.5.x features it might be easier to keep working off of develop branch until 1.4.1.

eatkins: Good point. 1.4.1-SNAPSHOT might be a better choice.

eed3si9n (approved):

PR #5945, opened 2020-10-06
Fix sbt hangs with invalid build.sbt and --batch

eatkins:
When sbt is starting up and there is an error with the build loading, we
need to read input from the user to determine to restart the build or
not. What is tricky is that there are potentially two sources of input:
thin clients connected through the boot server socket and the actual sbt
console process. If there are not connected thin clients and no system
console is available, we should return -1 in System.in.read, which will
cause sbt to exit.

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
@@ -179,6 +179,7 @@ trait Terminal extends AutoCloseable {
 }
 
 object Terminal {
+  val NO_BOOT_CLIENTS_CONNECTED: Int = -2
   // Disable noisy jline log spam
   if (System.getProperty("sbt.jline.verbose", "false") != "true")
     jline.internal.Log.setOutput(new PrintStream(_ => {}, false))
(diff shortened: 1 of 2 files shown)


eatkins:
Fixes #5938

internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
```diff
@@ -618,6 +618,13 @@ object Terminal {
                 if (running.get) {
                   inputStream.read match {
                     case -1 =>
+                    case -2 =>
+                      // The boot server socket returns -2 if there are no connected
+                      // thin clients. If there is also no console attached, return -1.
```
eed3si9n: It might be better to turn this into an enum or static variable in a Java class instead of a comment.

eatkins: https://github.com/sbt/sbt/compare/a5463139e0703e0e3f22c157acd14b0c598d7619..10c549a0b950db55eb9bf9cb90cef92c393282f8

eed3si9n (approved):

Issue #5940, opened 2020-10-06
sbt 1.4.0: "cancelable" is not working


unarist:
## steps

build.sbt

```
cancelable in Global := true
val foo = taskKey[Unit]("foo")
foo := {
  Thread.sleep(5000)
}
```

1. Run `sbt`
2. Enter `foo`
3. Hit Ctrl-C during task execution

## problem

sbt waits completion of the task execution (i.e. waits 5 seconds for this case), then exits sbt shell.

```
sbt:test> foo
   (hit Ctrl-C)
[success] Total time: 5 s, completed 2020/10/06 14:15:49
sbt:test>
   (then exits sbt shell)
```

(The last line is printed due to sbt.bat on Windows)

## expectation

sbt cancels task execution and returns to sbt prompt.

Example (sbt 1.3.13):

```
sbt:test> foo
   (hit Ctrl-C)
[warn] Canceling execution...
[error] Total time: 1 s, completed 2020/10/06 14:13:37
sbt:test>   
```

eed3si9n:
@unarist Thanks for the report.

eatkins:
The regression was introduced in bb8b9a1c997ef472cd8f321abc273aa7c128d115. Hopefully it won't be too difficult to fix.

eatkins:
#5947

PR #5947, opened 2020-10-06
Fix ctrl+c handling on windows

eatkins:
In windows, it is necessary for the console mode to include the
ENABLE_PROCESS_INPUT flag in order for ctrl+c to be treated as a signal
rather than a character. In jline 2, the ENABLE_PROCESS_INPUT flag was
preserved whenever the console mode was changed but in jline 3, it was
not. It is easy enough to manually set the flag after entering and
exiting raw mode and setting attributes (which are the apis that change
the console mode on windows).


eatkins:
Fixes #5940

eed3si9n (approved):

PR #5948, opened 2020-10-07
Fix #5935: Introduce bspReporter key

### LINKED ISSUE #5935: sbt 1.4.0: setting a custom xsbti.Reporter for compilerReporter leads to a MatchError
## steps

set `compilerReporter` to a custom implementation of `xsbti.Reporter`

## problem

because the custom Reporter is not an instance of BuildServerReporter,
https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/Defaults.scala#L2147 produces a MatchError.
https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/Defaults.scala#L2118 probably does, too

## expectation

using a custom reporter should work as before

## notes

i used this for a nasty hack where the custom reporter (after messaging my editor) forwarded all calls to the original reporter.
this cannot be made to work correctly any more, because `sendFinalReport` is not a part of the `Reporter` interface, and cannot even be overridden in a subclass of `BuildServerReporter` because it's `private[sbt]` there.

adpi2:
Fixes #5935 

I introduce the `bspReporter` key that is typed to `BuildServerReporter`.

The `bspReporter` encapsulates the `compilerReporter`. It sends the compiler diagnostics to the BSP client and then forward them to the `compilerReporter`. In case `bspEnabled` is set to false, the `bspReporter` would only forward the logs to the `compilerReporter`, so that the overhead is minimal.

A user can override the `compilerReporter` as usual. It should not need to override the `bspReporter` which is internal.

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
@@ -11,37 +11,57 @@ import java.io.File
 
 import sbt.StandardMain
 import sbt.internal.bsp._
-import sbt.internal.inc.LoggedReporter
 import sbt.internal.util.ManagedLogger
-import xsbti.{ Problem, Severity, Position => XPosition }
+import xsbti.{ Problem, Reporter, Severity, Position => XPosition }
 
 import scala.collection.mutable
 
-/**
- * Defines a compiler reporter that uses event logging provided by a `ManagedLogger`.
- *
- * @param maximumErrors The maximum errors.
- * @param logger The event managed logger.
- * @param sourcePositionMapper The position mapper.
- */
-class BuildServerReporter(
-    buildTarget: BuildTargetIdentifier,
-    maximumErrors: Int,
-    logger: ManagedLogger,
-    sourcePositionMapper: XPosition => XPosition = identity[XPosition],
-    sources: Seq[File]
-) extends LoggedReporter(maximumErrors, logger, sourcePositionMapper) {
-  import LoggedReporter.problemFormats._
-  import LoggedReporter.problemStringFormats._
-  logger.registerStringCodec[Problem]
+sealed trait BuildServerReporter extends Reporter {
+  private final val sigFilesWritten = "[sig files written]"
 
+  protected def logger: ManagedLogger
+
+  protected def underlying: Reporter
+
+  protected def publishDiagnostic(problem: Problem): Unit
+
+  def sendFinalReport(): Unit
+
+  override def reset(): Unit = underlying.reset()
+
+  override def hasErrors: Boolean = underlying.hasErrors
+
+  override def hasWarnings: Boolean = underlying.hasWarnings
+
+  override def printSummary(): Unit = underlying.printSummary()
+
+  override def problems(): Array[Problem] = underlying.problems()
+
+  override def log(problem: Problem): Unit = {
+    if (problem.message == sigFilesWritten) {
+      logger.debug(sigFilesWritten)
+    } else {
+      publishDiagnostic(problem)
+      underlying.log(problem)
+    }
+  }
+
+  override def comment(pos: XPosition, msg: String): Unit = underlying.comment(pos, msg)
+}
+
+final class BuildServerReporterImpl(
+    buildTarget: BuildTargetIdentifier,
+    protected override val logger: ManagedLogger,
+    protected override val underlying: Reporter,
+    sources: Seq[File]
+) extends BuildServerReporter {
   import sbt.internal.bsp.codec.JsonProtocol._
   import sbt.internal.inc.JavaInterfaceUtil._
 
   private lazy val exchange = StandardMain.exchange
   private val problemsByFile = mutable.Map[File, Vector[Diagnostic]]()
 
-  private[sbt] def sendFinalReport(): Unit = {
+  override def sendFinalReport(): Unit = {
     for (source <- sources) {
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
```diff
@@ -11,37 +11,57 @@ import java.io.File
...
 
-/**
- * Defines a compiler reporter that uses event logging provided by a `ManagedLogger`.
- *
- * @param maximumErrors The maximum errors.
- * @param logger The event managed logger.
- * @param sourcePositionMapper The position mapper.
- */
-class BuildServerReporter(
+trait BuildServerReporter extends Reporter {
```
eed3si9n:
```suggestion
sealed trait BuildServerReporter extends Reporter {
```

Can we seal this?

adpi2: Yes we can :+1:

eed3si9n (approved):

DavidPerezIngeniero:
In SBT 1.6.1, `bspReporter` is public whereas `compilerReporter` is private.
I would like to process the compiler errors and don't know how to do it.

adpi2:
What kind of processing to you want to do on the compiler errors? Do you just want to extract them for later use or also to modify them and report the modified versions?

DavidPerezIngeniero:
I have mixed the output of tests, compiler warnings, processing Javascript, ..... 
It's terrible to analyze what it has happened.
So I'd like to export all compiler error/warnings to a file.  This file will be used later to build a report visible in Jenkins.

https://stackoverflow.com/questions/71646720/how-to-capture-compilation-errors-and-warnings-in-sbt

adpi2:
In this case I would recommend to use `compilerReporter`. It is private but stable for quite some time now (since 0.13.2) and its API `xsbti.Reporter` is public and stable too.

In contrast, `bspReporter` is public but its API is `internal` and it could change in the near future.

DavidPerezIngeniero:
Thanks @adpi2  for your valuable tip.

How can I use `compilerReporter` if it's private?

This is my first try:
```
Global/compilerReporter := {
	val log = new xsbti.Logger {
		override def error(msg: Supplier[String]) = ???

		override def warn(msg: Supplier[String]) = ???

		override def info(msg: Supplier[String]) = ???

		override def debug(msg: Supplier[String]) = ???

		override def trace(exception: Supplier[Throwable]) = ???
	}
	new sbt.internal.inc.LoggedReporter(50, log)
}
```
The result :-( :

```
build.sbt:181: error: not found: value compilerReporter
Global/compilerReporter := {
       ^
```
The only solution I can think is to fork SBT and remove the `private[sbt]`, but this is a little clumsy.

adpi2:
> How can I use compilerReporter if it's private?

It is package private so you can use it from within the `sbt` package. Maybe you should create an `AutoPlugin` in the `sbt` package: 

```scala
package sbt

SbtExtensionPlugin extends AutoPlugin{
  override def globalSettings = Seq(
    compileReporter := {
      ...
    }
  )
}
```

It's kind of ugly but I don't see any better solution.

DavidPerezIngeniero:
@adpi2 , I've tried your valuable solution and it doesn't work.

My code:

`project/PluginInformeCompilar.scala`
```
object PluginInformeCompilar extends AutoPlugin{
  val pref = Seq(
    compilerReporter := {
      val log = new xsbti.Logger {
        override def debug(msg: Supplier[String]) = ()

        override def warn(msg: Supplier[String]) =
          println("[aviso] "+msg.get())

        override def info(msg: Supplier[String]) =
          println("[infor] "+msg.get())

        override def error(msg: Supplier[String]) =
          println("[error] "+msg.get())

        override def trace(msg: Supplier[Throwable]) = ()
      }
      new LoggedReporter(50, log)
    }
  )
  override def projectSettings = pref
  override def globalSettings: Seq[Def.Setting[_]] = pref

  override val requires = JvmPlugin
}
```
`build.sbt` ()1st ftry):
```
lazy val myProject = project
	.enablePlugins(ParadoxSitePlugin, PackPlugin, PluginInformeCompilar)
	.settings(
		....
        )
```
`build.sbt` (2nd try):
```
lazy val myProject = project
	.enablePlugins(ParadoxSitePlugin, PackPlugin)
	.settings(
		PluginInformeCompilar.pref(0),
		....
        )
```
I don't see my custom prefixes when running:
```
myProject/compile
````
Any ideas please?

**Comment by adpi2** [comment]
The `inspect` command can be useful to debug those kind of things:

```
sbt:sbt-5948> inspect compilerReporter
[info] No entry for key.
[info] Description:
[info]  Experimental hook to listen (or send) compilation failure messages.
[info] Delegates:
[info]  compilerReporter
[info]  ThisBuild / compilerReporter
[info]  Global / compilerReporter
[info] Related:
[info]  Compile / compile / compilerReporter
[info]  Test / compile / compilerReporter
```

Here you can see that `compilerReporter` is scoped to `Compile / compile` and `Test / compile`. So that's where you should override it:

```scala
object PluginInformeCompilar extends AutoPlugin{
  private val configSettings = Seq(
    compile / compilerReporter := {
      ...
  )

  override def projectSettings = Def.settings(
    inConfig(Compile)(configSettings),
    inConfig(Test)(configSettings)
  )

  override val requires = JvmPlugin
  override def trigger: PluginTrigger = allRequirements
}
```

DavidPerezIngeniero:
Confirmed that now it is overriden!.
Thanks.

Issue #5952, opened 2020-10-07
Input problems in sbt new prompt


envilogger:
## steps

Trying to create new project from g8 template with `set new` command

```
$ sbt --version

sbt script version: 1.4.0

$ sbt new scala/scala-seed.g8
[info] welcome to sbt 1.4.0 (GraalVM Community Java 11.0.8)
[info] loading global plugins from /Users/envilogger/.sbt/1.0/plugins
[info] loading project definition from /private/tmp/seed/project
[info] set current project to seed (in build file:/private/tmp/seed/)
[info] set current project to seed (in build file:/private/tmp/seed/)
Passphrase for /Users/envilogger/.ssh/id_rsa:

```

## problem

Can't enter the passphrase, after any key pressed passphrase requested again.
Even if pasted passphrase, I can't enter project name (like all input is ignored)

## expectations

User need to be able to enter passphrase and project name without any problems

## notes

*workaround for the problem*: Paste values instead of typing it.

eatkins:
Thanks for the report! I can't exactly reproduce this but maybe this is because I have my ssh key password in keychain? I'm assuming you're using osx based on the home directory in the report.

I did notice however that the keyboard input was not echoed for the project name. I was able to input a value for project name but the text wasn't displayed.

eatkins:
I think this should be fixed in 1.4.1. We can reopen if that isn't the case.

PashoQ:
It is still not fixed in 1.4.1, at least on Widows.
```
Microsoft Windows [Version 10.0.18363.1082]
(c) 2019 Microsoft Corporation. All rights reserved.

C:\Dev\Temp\bla>sbt new https://bicsgit.bc/psd-dwh/sbt_oozie-template.g8.git
[info] welcome to sbt 1.4.1 (AdoptOpenJDK Java 1.8.0_265)
[info] loading settings for project global-plugins from dependency-graph.sbt,updates.sbt ...
[info] loading global plugins from C:\Users\id967052\.sbt\1.0\plugins
[info] set current project to bla (in build file:/C:/Dev/Temp/bla/)
[info] set current project to bla (in build file:/C:/Dev/Temp/bla/)
Username: user_name
Password:
[...]

PR #5954, opened 2020-10-07
Don't enter raw mode preemptively

eatkins:
The user input is not echoed in sbt new because we were switching to raw
mode when creating the console terminal. I can't quite remember why I
was entering raw mode preemptively but it doesn't seem like the best
idea.


eatkins:
At least partially fixes #5952

eed3si9n (approved):

Issue #5958, opened 2020-10-08
sbt 1.4.0: sbt.color flag does not work on CI


christobill:
## steps

Start a new project from the scala-seed g8 template (for example)
```
sbt new scala/scala-seed.g8
```

### With sbt 1.4.0-M1 and older version

```
export CI=true
sbt -Dsbt.color=true test
```

Display colors in test results

### With sbt 1.4.0-M2 and latest version (including 1.4.0 release)

```
export CI=true
sbt -Dsbt.color=true test
```

**Does not** display colors in test results

## problem

Harder to see failing tests on CI when looking at the test job result 
Behavior not expected when setting flag `sbt.color=true`

### With sbt 1.4.x

```
export CI=true
sbt -Dsbt.color=true test
```

Display colors in test results

eatkins:
Thanks for the report! Should be fixed by #5960

AndriusFromLTU:
After upgrading to `1.4.2` version the coloring doesn't work. Tested via:

```
export SBT_OPTS="-Dsbt.color=true"
sbt test
```
and 
```
export CI=true
sbt -Dsbt.color=true test
```
Downgrading to 1.4.1 works. Should I create a new issue or this comment is enough?

eatkins:
Thanks for the report. I think this should be fixed by #6061. FTR, I couldn't fully reproduce this. There were some colors printed with sbt 1.4.2 but the scalatest output was not colorized which is what #6061 fixes.

PR #5960, opened 2020-10-08
Set SimpleTerminal properties based on system props

eatkins:
When running sbt 1.4.0 with -Dsbt.ci=true and -Dsbt.color=true, there is
no color output. This was because in this scenario, a SimpleTerminal was
used and isAnsiSupported and isColorEnabled were hardcoded to false
rather than reading the values from the system properties.


eatkins:
Fixes #5958

Issue #5965, opened 2020-10-09
Writing SBT output to file ends up with terminal color stuff in file


soc:
(As requested by @eed3si9n in https://twitter.com/eed3si9n/status/1313125156559171591)

## steps

Run a application with `sbt run` and pipe it into a file: `sbt run > some_file`

## problem

In 1.4.0 the file contains terminal color stuff:

```
[info] welcome to sbt 1.4.0 (Ubuntu Java 11.0.8)
[0J[info] loading settings for project global-plugins from gpg.sbt ...
[0J[info] loading global plugins from /home/soc/.sbt/1.0/plugins
[0J[info] loading project definition from /home/soc/code/module-infos/project
[0J[info] loading settings for project root from build.sbt ...
[0J[info] set current project to module-infos (in build file:/home/soc/code/module-infos/)
[0J[info] running mods.Main
[0JBLA                                    DONE[0J
[0JBLI                         DONE[0J
[0JBLUBBB                      TODO[0J
```

## expectation

The file does not contain terminal color stuff, like in 1.3.x.

## notes

The problem can be worked around with `--no-colors`.

eed3si9n:
Ethan fixed this in https://github.com/sbt/sbt/pull/5966

soc:
Thanks!

PR #5966, opened 2020-10-09
Remove ansi control characters from piped output

eatkins:
When piping an sbt task to a file, the expectation is that sbt will not
write ansi control characters and colors unless the users specifies that
with, e.g. -D.sbt.color=true. With sbt 1.4.0, all output bytes are
routed through the progress output processor which tries to ensure that
progress lines are not interleaved with log lines. The issue was that
the hasProgress flag was being set to true for the server process even
when the formatEnabledInEnv flag was set to false. This caused each log
line to have a leading clear screen before cursor ansi control code
written which would appear in the output file.


eatkins:
Fixes #5965

eed3si9n (approved):
Thanks!

PR #5972, opened 2020-10-13
Check if target is directory before creating

eatkins:
When project/target is a symbolic link, sbt 1.4.0 crashes on startup
because Files.createDirectories will throw a FileAlreadyExistsException.
The fix is to first check if the target directory exists before trying
to create it.


sideeffffect:
Awesome, thank you!
Btw there could be cases where not only `project/target` is a symlink, but also `project/project/target` or `project/project/project/target` (or how deep the rabbit hole goes). Does this fix address those cases too?

sideeffffect:
When could we expect this fix to land in a release (1.4.1 I presume)?

eatkins:
Yes, it would be in 1.4.1 but when that release will land is a question for @eed3si9n

sideeffffect:
The build seems to have failed because of Internet connectivity issues on Travis side. Otherwise it should be OK.

eed3si9n (approved):
LGTM Thanks for the fix

PR #5973, opened 2020-10-14
load .inputrc file

xuwei-k:
I think jline2 automatically load `.inputrc` file. But jline3 changed behavior.

https://github.com/jline/jline3/commit/9ccfe0b121d2ea72a8524c84a9bda5abaa68d45a


eed3si9n (approved):
Thanks @xuwei-k!

There was someone who noticed `~/.inputrc` earlier too, so this is a welcome fix.

Issue #5955, opened 2020-10-07
sbt 1.4.0 incorrectly warns about the `shellPrompt` key not being used


smarter:
## steps

I have the following in my `~/.sbt/1.0/prompt.sbt` which has always worked well:
```scala
// Based on http://www.whiteboardcoder.com/2016/03/sbt-customize-shell-prompt-in-sbt.html
shellPrompt := { state =>
  def textColor(color: Int)      = s"\033[38;5;${color}m"
  def backgroundColor(color:Int) = s"\033[48;5;${color}m"
  val reset                      = s"\033[0m"
  def bold(x: String)            = s"\033[1m${x}${reset}"

  def formatText(str: String)(txtColor: Int) =
    s"${textColor(txtColor)}${str}${reset}"

  // https://upload.wikimedia.org/wikipedia/en/1/15/Xterm_256color_chart.svg
  val red    = 1
  val lightred = 9
  val goodred = 160
  val green  = 2
  val lightgreen = 10
  val cyan = 6
  val lightcyan = 14
  val blue = 27
  val lightblue = 39
  val yellow = 11
  val white  = 15
  val black  = 16
  val orange = 166
  val lightorange = 202

  def gitBranch =
    try {
      scala.sys.process.Process("git rev-parse --abbrev-ref HEAD").lines.head
    } catch {
      case _: Exception =>
        ""
    }

  bold(formatText(s"[")(blue/*goodred*/)) +
  formatText(s"${name.value}@")(blue) +
  bold(formatText(s"${gitBranch}")(lightblue)) +
  bold(formatText(s"]")(blue/*goodred*/)) +
  bold(formatText(s">")(blue+6)) +
  " "
}
```

## problem

When I start a project with sbt 1.4.0 I see:
```scala
[warn] there's a key that's not used by any other settings/tasks:
[warn]  
[warn] * root / shellPrompt
[warn]   +- /home/smarter/.sbt/1.0/prompt.sbt:3
[warn]  
[warn] note: a setting might still be used by a command; to exclude a key from this `lintUnused` check
[warn] either append it to `Global / excludeLintKeys` or call .withRank(KeyRanks.Invisible) on the key
```

## expectation

No warning, since this actually does set the prompt as expected.

eed3si9n:
This was fixed by @xirc in https://github.com/sbt/sbt/pull/5983

PR #5983, opened 2020-10-18
Add `shellPrompt` into excludeLintKeys

xirc:
Resolve the issue https://github.com/sbt/sbt/issues/5955


eed3si9n (approved):
Thanks @xirc!

Issue #5849, opened 2020-09-14
sbt warns about unused keys when shouldn't


limansky:
I've faced with this issue with sbt-release plugin, but possible there can be other false warnings.

## steps

1. Use sbt-release.
2. Set `releaseCrossBuild` and `releaseProcess` settings.
3. Upgrade to sbt 1.4.0-RC1

## problem

I've got following warning:
```
[warn] there are 2 keys that are not used by any other settings/tasks:
[warn]  
[warn] * beanPuree / releaseCrossBuild
[warn]   +- /home/limansky/projects/beanpuree/build.sbt:61
[warn] * beanPuree / releaseProcess
[warn]   +- /home/limansky/projects/beanpuree/build.sbt:63
```

## expectation

No warning should be produced, since this settings are used by the plugin (`release` task).

## notes

The `build.sbt` file is here: https://github.com/limansky/beanpuree/blob/master/build.sbt

eed3si9n:
I am guessing that `release` is a command, so you're right that it shouldn't be warned, but it's also true that it's not being used by any other settings/tasks.

limansky:
@eed3si9n you're right, it's a command https://github.com/sbt/sbt-release/blob/master/src/main/scala/ReleasePlugin.scala#L159 .

l15k4:
:wave: There are more of these, like : 
```
run / configuration
semanticdbTargetRoot
sourceDirectories
```
Dunno how to shake it off, there are 30 warnings in a project with 10 subprojects

eatkins:
@l15k4 `Global / lintUnusedKeysOnLoad := false` will turn off this feature.

PR #5985, opened 2020-10-18
Use lint***Filter instead of ***LintKeys for more reliable tests

xirc:
Use `lint***Filter` instead of `***LintKeys` for more reliable tests.
I have found this improvement when I attempted to solve the issue https://github.com/sbt/sbt/issues/5849.

## Problem

`lintUnusedTask` passes the result of `lintExcludeFilter` as the argument  to `lintUnused`.
https://github.com/sbt/sbt/blob/28833db333f2aa6961df849a9ea46fb3094c048a/main/src/main/scala/sbt/internal/LintUnused.scala#L60-L69

`lintExcludeFilter` excludes watch related settings which has the prefix `watch`.
https://github.com/sbt/sbt/blob/28833db333f2aa6961df849a9ea46fb3094c048a/main/src/main/scala/sbt/internal/LintUnused.scala#L24-L27

But, in the test, `excludeLintKeys` are passed to `lintUnused`.
`excludeLintKeys` are used in `lintExcludeFilter`, and it doesn't exclude watch related settings.
Thus, excluding watch related settings from lint keys are not well tested.
https://github.com/sbt/sbt/blob/28833db333f2aa6961df849a9ea46fb3094c048a/sbt/src/sbt-test/project/lint/build.sbt#L11-L13

I think I can say same thing also about `lintIncludeFilter` and `includeLintKeys`.

## Solution

- Use `lintExcludeFilter` instead of `excludeLintKeys` in the test
- Use `lintIncludeFilter` instead of `includeLintKeys` in the test

### DIFF
--- a/sbt/src/sbt-test/project/lint/build.sbt
+++ b/sbt/src/sbt-test/project/lint/build.sbt
@@ -2,14 +2,18 @@ ThisBuild / doc / scalacOptions += "-Xsomething"
 
 ThisBuild / shellPrompt := { state => "sbt> " }
 
+// watch related settings
+ThisBuild / sbt.nio.Keys.watchTriggers := Seq()
+ThisBuild / sbt.nio.Keys.watchPersistFileStamps := true
+
 lazy val lintBuildTest = taskKey[Unit]("")
 
 lazy val root = (project in file("."))
   .settings(
     lintBuildTest := {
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks @xirc!

Issue #5992, opened 2020-10-20
OOM due to SBT 1.4.+ upgrade in CI build 


Baccata:
## problem: SBT upgrade leads to CI build crashing 

* The [commit](https://github.com/disneystreaming/weaver-test/pull/111#partial-pull-merging)
* The [failed build](https://cloud.drone.io/disneystreaming/weaver-test/352)
* The SBT_OPTS [ci settings](https://github.com/disneystreaming/weaver-test/blob/master/.drone.yml#L14) 

NB : an OOM had already been witnessed in a previous version of this project, which had lead to the introduction of these SBT_OPTS, fixing the problem.

eatkins:
Thanks for the report! I think I tracked down the memory leak in #6001.

eatkins:
Fixed in #6001

PR #6001, opened 2020-10-21
Fix EvaluateTask memory leak

eatkins:
EvaluateTask was holding references to SafeState that could be quite
large. This was reported as #5992. In that project, I ran the `ci` task
and observed the OOM as reported. I took a heap dump prior to OOM and
got the retained size graph from visualvm (which took hours to compute).
The lastEvaluatedState was holding a reference to SafeState that was
1.7GB. The project max heap size was set to 2GB. Instead of using the
lastEvaluatedState, we can just use StandardMain.exchange.withState.
The cached instances of state were used for task cancellation and
completions. While it is possible that early on in booting
StandardMain.exchange.withState could return a null state, in practice
this won't happen because it is set early on during the sbt boot
commands.

After this change, I successfully ran the `ci` task in the #5992 issue
project with the same memory parameters as their ci config.

### DIFF
--- a/main-command/src/main/scala/sbt/State.scala
+++ b/main-command/src/main/scala/sbt/State.scala
@@ -72,12 +72,15 @@ final case class State(
  * @param currentExecId provide the execId extracted from the original State.
  * @param combinedParser the parser extracted from the original State.
  */
+@deprecated("unused", "1.4.2")
 private[sbt] final case class SafeState(
     currentExecId: Option[String],
     combinedParser: Parser[() => sbt.State]
 )
 
+@deprecated("unused", "1.4.2")
 private[sbt] object SafeState {
+  @deprecated("use StandardMain.exchange.withState", "1.4.2")
   def apply(s: State) = {
     new SafeState(
       currentExecId = s.currentCommand.map(_.execId).flatten,
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -413,8 +413,12 @@ object EvaluateTask {
       (dummyRoots, roots) :: (Def.dummyStreamsManager, streams) :: (dummyState, state) :: dummies
     )
 
+  @deprecated("use StandardMain.exchange.withState to obtain an instance of State", "1.4.2")
   val lastEvaluatedState: AtomicReference[SafeState] = new AtomicReference()
+  @deprecated("use currentlyRunningTaskEngine", "1.4.2")
   val currentlyRunningEngine: AtomicReference[(SafeState, RunningTaskEngine)] =
+    new AtomicReference()
+  private[sbt] val currentlyRunningTaskEngine: AtomicReference[RunningTaskEngine] =
     new AtomicReference()
 
   /**
@@ -486,7 +490,7 @@ object EvaluateTask {
         shutdownImpl(true)
       }
     }
-    currentlyRunningEngine.set((SafeState(state), runningEngine))
+    currentlyRunningTaskEngine.set(runningEngine)
     // Register with our cancel handler we're about to start.
     val strat = config.cancelStrategy
     val cancelState = strat.onTaskEngineStart(runningEngine)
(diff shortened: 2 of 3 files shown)


main-command/src/main/scala/sbt/State.scala
```diff
@@ -72,12 +72,15 @@ final case class State(
  * @param currentExecId provide the execId extracted from the original State.
  * @param combinedParser the parser extracted from the original State.
  */
+@deprecated("unused", "1.4.2")
 private[sbt] final case class SafeState(
     currentExecId: Option[String],
     combinedParser: Parser[() => sbt.State]
 )
```
eed3si9n, line 79: If it's `private[sbt]`, maybe we could rip it out?

eed3si9n:
@eatkins Thanks for the fix!

main-command/src/main/scala/sbt/State.scala
eatkins: We could. This just felt easier than updating mima.

eatkins: Oh, the other thing is that there are public apis in `EvaluateTask` that expose `SafeState` (which sort of seems like a compiler bug) so we'd have to break binary compatibility of `EvaluateTask` which is neither `private[sbt]` nor in the internal package.

eatkins:
```
sbt:sbtRoot> mimaReportBinaryIssues
[info] sbt-launch: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] sbtRoot: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] sbt-client: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[error]  * object sbt.SafeState does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("sbt.SafeState$")
[error]  * class sbt.SafeState does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("sbt.SafeState")
[info] Scripted sbt Redux: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] sbt-dependency-tree: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[error]  * static method currentlyRunningEngine()java.util.concurrent.atomic.AtomicReference in class sbt.EvaluateTask does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("sbt.EvaluateTask.currentlyRunningEngine")
[error]  * static method lastEvaluatedState()java.util.concurrent.atomic.AtomicReference in class sbt.EvaluateTask does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("sbt.EvaluateTask.lastEvaluatedState")
[error]  * method lastEvaluatedState()java.util.concurrent.atomic.AtomicReference in object sbt.EvaluateTask does not have a correspondent in current version                    
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("sbt.EvaluateTask.lastEvaluatedState")
[error]  * method currentlyRunningEngine()java.util.concurrent.atomic.AtomicReference in object sbt.EvaluateTask does not have a correspondent in current version                
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("sbt.EvaluateTask.currentlyRunningEngine")
[error] stack trace is suppressed; run last mainProj / mimaReportBinaryIssues for the full output
[error] stack trace is suppressed; run last commandProj / mimaReportBinaryIssues for the full output
[error] (mainProj / mimaReportBinaryIssues) Main: Binary compatibility check failed!
[error] (commandProj / mimaReportBinaryIssues) Command: Binary compatibility check failed!
[error] Total time: 43 s, completed Oct 20, 2020 10:10:19 PM
```
I doubt that `SafeState` was widely used but I'm not sure if we should take it out.

eed3si9n: ok.

eed3si9n (approved):

PR #6007, opened 2020-10-22
Fix #6006: BuildServerReporter on Dotty

### LINKED ISSUE #6006: [1.4.1] BSP empty failure report on Dotty project
## steps

- Create an sbt 1.4.1 project
- Connect to it using BSP (in Metals for instance)
- Add a typo in a scala source file

## problem

The typo is reported by a BSP diagnostic but then an empty diagnostic makes it disappear.

## expectation

The failure report should contain all previous encountered errors and warnings.

adpi2:
Fixes #6006 

When using `scalac`, the `sourcePath` of the reported positions is virtual file refs.
When using Dotty, the `sourcePath` of the reported positions is absolute path, and so the `BuildServerReporter` was not working.

In the `BuildServerReporter` I now use the `pos.sourceFile` rather than the `pos.sourcePath` to store the diagnostics. The `pos.sourceFile` is much more stable because it is a real file from which I can reliably extract the absolute path.

I have added 2 build server tests. The first one checks that the last report contains an error, the other one checks that the last report contains a warning.

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
@@ -7,6 +7,8 @@
 
 package sbt.internal.server
 
+import java.nio.file.Path
+
 import sbt.StandardMain
 import sbt.internal.bsp._
 import sbt.internal.util.ManagedLogger
@@ -69,7 +71,7 @@ final class BuildServerReporterImpl(
   import sbt.internal.inc.JavaInterfaceUtil._
 
   private lazy val exchange = StandardMain.exchange
-  private val problemsByFile = mutable.Map[VirtualFileRef, Vector[Diagnostic]]()
+  private val problemsByFile = mutable.Map[Path, Vector[Diagnostic]]()
 
   override def sendSuccessReport(analysis: CompileAnalysis): Unit = {
     for {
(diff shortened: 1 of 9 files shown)


eed3si9n:
> When using Dotty, the `sourcePath` of the reported positions is absolute path, and so the BuildServerReporter was not working.

Which file paths are we passing to Dotty?

> In the BuildServerReporter I now use the `pos.sourceFile` rather than the `pos.sourcePath` to store the diagnostics.

We should avoid the use of `java.io.File` as much as possible. Not just for virtual file, but this also relate to other abstract paths like paths in JAR files and Java modules.

adpi2:
> Which file paths are we passing to Dotty?

We are passing `VirtualFile`.

But the Dotty `sbt-bridge` is sending absolute paths back to us. It get them from the `pos.sourceFile`:
https://github.com/lampepfl/dotty/blob/b2dc1016976558c6fbc67970139266486f2ad9d7/sbt-bridge/src/xsbt/DelegatingReporter.java#L77-L91

> We should avoid the use of java.io.File as much as possible.

Ok. I changed the implementation so that it relies on `pos.sourcePath` and not `pos.sourceFile`. It uses the `FileConverter` to handle both cases: virtual file ids and absolute paths.

eed3si9n:
> But the Dotty sbt-bridge is sending absolute paths back to us. It get them from the pos.sourceFile:
https://github.com/lampepfl/dotty/blob/b2dc1016976558c6fbc67970139266486f2ad9d7/sbt-bridge/src/xsbt/DelegatingReporter.java#L77-L91

Since I made changes to the compiler bridge in sbt/zinc but did not send PRs to Dotty it makes sense that there are some differences in the implementation. I'd suggest we send PR to make that side consistent then.

adpi2:
> I'd suggest we send PR to make that side consistent then.

Agreed.

I still think this current PR makes much sense since it enables compatibility on previous versions of `sbt-dotty`. IMO it is a pity that I cannot use BSP in `sbt 1.4.1` on a Dotty project. It used to work pretty well on `1.4.0`.

Also I think this implementation will be a lot more robust on future changes since it relies on the file converter.

eed3si9n (approved):

Issue #5285, opened 2019-12-10
--addPluginSbtFile does not parse file names with spaces


jastice:
## steps
sbt version: 1.3.3

Start sbt with `--addPluginSbtFile` and a path with spaces:
```
sbt "--addPluginSbtFile=\"C:/Users/jast/AppData/Local/Temp/file with spaces.sbt\"
```
## problem

This results in a parsing error.
```
[error] Expected ';'
[error] addPluginSbtFile=C:/Users/jast/AppData/Local/Temp/with space.sbt
```
### doesn't work

Using a triplequoted/verbatim string in the parameter:
```
sbt "--addPluginSbtFile=\"\"\"C:/Users/jast/AppData/Local/Temp/file with spaces.sbt\"\"\"
```

Unicode-escaping the spaces:
```
sbt "--addPluginSbtFile=\"C:/Users/jast/AppData/Local/Temp/file\u0020with\u0020spaces.sbt\"
```

### seems to work?

using `early(...)` instead of `--` together with verbatim string:
```
sbt "early(addPluginSbtFile=\"\"\"C:\Users\jast\AppData\Local\Temp\idea with space.sbt\"\"\")"
```

## notes

Corresponding IntelliJ ticket: https://youtrack.jetbrains.com/issue/SCL-15360

eed3si9n:
@jastice Thanks for the report.

I was able to reproduce this on macOS too:

```
$ sbt "--addPluginSbtFile=\"/tmp/space dir/foo.sbt\"" -Dsbt.version=1.3.4
[error] Expected ';'
[error] addPluginSbtFile=/tmp/space dir/foo.sbt
[error]                                        ^
```

PR #6016, opened 2020-10-24
Fix console echo for windows batch commands

eatkins:
For reasons I don't fully understand, it is necessary to go in and out
of raw mode to get console input to echo on windows. This was reported
as sbt new not echoing input in #5952.

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
@@ -796,6 +796,12 @@ object Terminal {
       private[util] val system: org.jline.terminal.Terminal,
   ) extends TerminalImpl(in, out, originalErr, "console0") {
     private[this] val rawMode = new AtomicBoolean(false)
+    if (Util.isWindows && hasConsole) {
+      // It is necessary to enter and exit raw mode in order to get the windows
+      // console to echo input.
+      enterRawMode()
+      exitRawMode()
+    }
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):

jastice:
I suspect this (or some other change) is causing sbt import to break in IntelliJ under Windows

see https://youtrack.jetbrains.com/issue/SCL-18393

eatkins:
@jastice the commit that broke intellij import on windows is actually a892a8368144c4d1455baf5491da42d6152a0658. There is a fix in #6051. I am curious though why intellij import works the way it does. Why does it start up a process and feed input in instead of just running in batch mode? Ideally it could use `sbtn` or `sbt -client` for newer versions of sbt. I guess maybe bsp might eliminate the need for intellij sbt projects in the long run anyway though?

jastice:
Thanks for the quick fix!

So I'm not entirely sure why the import was built this way rather than using batch mode, since it predates my involvement, but I kept it working that way since it is generally reliable. One issue that does come up with batch mode sometimes is passing arguments with spaces and quotes can break the argument parsing, even in different ways on different platforms (eg #5285)

Mid-term, I'm hoping BSP will obsolete this mechanism (perhaps even by next release), but either way it will be around for a long time for backward-compatibility.

eatkins:
Yeah. For sure we should support the old way of doing it in intellij. I was just wondering if the ux could be improved.

Issue #6015, opened 2020-10-23
sbt 1.4.1: watchBeforeCommand is no longer executed in ~compile


ritschwumm:
## steps

create a minimal project
```
// build.sbt
organization	:= "de.test"
version			:= "0.0.1"
name			:= "repro"
scalaVersion	:= "2.13.3"

watchBeforeCommand	:= { () =>
	println("WATCHBEFORE")
	watchBeforeCommand.value()
}
```
```
# project/build.properties
sbt.version=1.4.1
```
```
// src/main/Repro.scala
package repro

object Repro {
	def main(args:Array[String]):Unit	= {
		println("hallo")
	}
}
```

run `~sbt compile` and change the `print` statement in `object Repro`

## problem

`WATCHBEFORE` is not printed to the console before compilation starts

## expectation

`WATCHBEFORE` is printed to the console before compilation starts, this worked in sbt 1.3.13

eatkins:
Thanks for the report! This was indeed a regression that should be fixed by #6018.

PR #6018, opened 2020-10-24
Reinstate watchBeforeCommand callback

eatkins:
A user reported that the watchBeforeCommand callback was not being
invoked in sbt 1.4.{0, 1}. This was an oversight that occurred when
refactoring watch for the thin client and there previously had been no
regression test for that callback.


eatkins:
Fixes #6015

eed3si9n (approved):

PR #6023, opened 2020-10-25
Add wizard for installing sbtn and completions

eatkins:
This commit adds a wizard for installing sbtn along with tab completions
for bash, fish, powershell and zsh. It introduces the `installSbtn`
command which installs sbtn into ~/.sbt/1.0/bin/sbtn(.exe) depending on
the platform. It also can optionally install completions. The
completions are installed into ~/.sbt/1.0/completions. The sbtn native
executable is installed by downloading the sbt universal zip for the
version (which can be provided as an input argument with a fallback to
the running sbt version) and extracting the platform specific binary
into ~/.sbt/1.0/bin. After installing the executable, it offers to setup
the path and completions for the four shells. With the user's consent,
it adds a line to the shell config that updates the path to include
~/.sbt/1.0/bin and another line to source the appropriate completion
file for the shell from ~/.sbt/1.0/completions.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -391,6 +391,8 @@ object Defaults extends BuildCommon {
       canonicalInput :== true,
       echoInput :== true,
       terminal := state.value.get(terminalKey).getOrElse(Terminal(ITerminal.get)),
+      InstallSbtn.installSbtn := InstallSbtn.installSbtnImpl.evaluated,
+      InstallSbtn.installSbtn / aggregate := false,
     ) ++ LintUnused.lintSettings
       ++ DefaultBackgroundJobService.backgroundJobServiceSettings
       ++ RemoteCache.globalSettings
--- a/main/src/main/scala/sbt/internal/InstallSbtn.scala
+++ b/main/src/main/scala/sbt/internal/InstallSbtn.scala
@@ -1,0 +1,226 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt
+package internal
+
+import Def._
+import Keys.{ sbtVersion, state, terminal }
+
+import java.io.{ File, FileInputStream, FileOutputStream, InputStream, IOException }
+import java.net.URL
+import java.nio.file.{ Files, Path }
+import java.util.zip.ZipInputStream
+import sbt.io.IO
+import sbt.io.Path.userHome
+import sbt.io.syntax._
+import scala.util.{ Properties, Try }
+
+private[sbt] object InstallSbtn {
+  private[sbt] val installSbtn =
+    Def.inputKey[Unit]("install sbtn and tab completions").withRank(KeyRanks.BTask)
+  private[sbt] def installSbtnImpl: Def.Initialize[InputTask[Unit]] = Def.inputTask {
+    val inputVersion = Def.spaceDelimited("version").parsed.headOption
+    val version = inputVersion.getOrElse(sbtVersion.value.replaceAllLiterally("-SNAPSHOT", ""))
+    val term = terminal.value
+    term.setMode(canonical = false, echo = false)
+    val baseDirectory = BuildPaths.getGlobalBase(state.value).toPath
+    val tmp = Files.createTempFile(s"sbt-$version", "zip")
+    val sbtn = if (Properties.isWin) "sbtn.exe" else "sbtn"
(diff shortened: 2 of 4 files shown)


eatkins:
Here's what it looks like in action:
```
sbt:scala-compile> installSbtn
downloading https://github.com/sbt/sbt/releases/download/v1.4.1/sbt-1.4.1.zip to /var/folders/zv/2wskcxc14cn6kyj4yb53whc40000gn/T/sbt-1.4.19861458291520778649zip
extracting /var/folders/zv/2wskcxc14cn6kyj4yb53whc40000gn/T/sbt-1.4.19861458291520778649zip!sbt/bin/sbtn-x86_64-apple-darwin to /Users/ethanatkins/.sbt/1.0/bin/sbtn
Setup sbtn for shell:
 [1] bash
 [2] fish
 [3] powershell
 [4] zsh
 [5] none
Enter option: 4
Add /Users/ethanatkins/.sbt/1.0/bin to PATH in /Users/ethanatkins/.zshrc? y/n (y default): y
Add tab completions to /Users/ethanatkins/.zshrc? y/n (y default): y

Successfully installed sbtn for zsh. You may need to restart zsh for the changes to take effect.
[success] Total time: 15 s, completed Oct 25, 2020, 3:53:29 PM

```

eatkins:
I manually tested that it works on mac, linux and windows. I only tested powershell on windows and I only tested bash, zsh and fish on linux and mac.

eatkins:
With an empty .zshrc, it generates:
```
# The following two lines were added by the sbt installSbtn task:
fpath=($fpath /Users/ethanatkins/.sbt/1.0/completions)
autoload -Uz compinit; compinit

path=($path /Users/ethanatkins/.sbt/1.0/bin)
```
For bash, it's:
```
export PATH=$PATH:/Users/ethanatkins/.sbt/1.0/bin
source /Users/ethanatkins/.sbt/1.0/completions/sbtn.bash
```
For fish, it's
```
set PATH $PATH /Users/ethanatkins/.sbt/1.0/bin
source /Users/ethanatkins/.sbt/1.0/completions/sbtn.fish
```
For powershell, it's
```
$env:Path += ";C:\Users\ethanatkins\.sbt\1.0\bin"                                                                                                                                                                                                                                     
. "C:\Users\ethanatkins\.sbt\1.0\completions\sbtn.ps1" 
```

main/src/main/scala/sbt/internal/InstallSbtn.scala
```diff
@@ -0,0 +1,228 @@
...
+    val os = new FileOutputStream(path.toFile)
+    val result = new Array[Byte](1024 * 1024)
+    var bytesRead = -1
+    try {
+      do {
+        bytesRead = inputStream.read(result)
+        if (bytesRead > 0) os.write(result, 0, bytesRead)
+      } while (bytesRead > 0)
+    } finally {
+      inputStream.close()
```
ritschwumm: when this throws an exception, the next line is never executed, right?

main/src/main/scala/sbt/internal/InstallSbtn.scala
```diff
@@ -0,0 +1,228 @@
...
+    ()
+  }
+  private[this] def downloadRelease(term: Terminal, version: String, location: Path): Unit = {
+    val zip = s"https://github.com/sbt/sbt/releases/download/v$version/sbt-$version.zip"
+    val url = new URL(zip)
+    term.printStream.println(s"downloading $zip to $location")
+    transfer(url.openStream(), location)
+  }
+  private[this] def transfer(inputStream: InputStream, path: Path): Unit = {
+    val os = new FileOutputStream(path.toFile)
```
ritschwumm: is it a problem the InpuStream is never closed when this throws an exception?

eatkins: It is extremely unlikely that this would throw an exception though I suppose it doesn't hurt to move up the try.

main/src/main/scala/sbt/internal/InstallSbtn.scala
eatkins: Yes, but I have never seen an inputStream throw an exception on close.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -389,6 +389,7 @@ object Defaults extends BuildCommon {
       canonicalInput :== true,
       echoInput :== true,
       terminal := state.value.get(terminalKey).getOrElse(Terminal(ITerminal.get)),
+      InstallSbtn.installSbtn := InstallSbtn.installSbtnImpl.evaluated,
```
eed3si9n, line 394: You might also need

```scala
InstallSbtn.installSbtn / aggregate := false,
```

eatkins: Good catch.

eed3si9n (approved):

PR #6025, opened 2020-10-26
Replace %20 with space in sbt script name

eatkins:
We replace spaces in the sbt script with %20 and we need to replace the
%20s with spaces.


eatkins:
See https://github.com/sbt/sbt-launcher-package/pull/338

eed3si9n (approved):

PR #6026, opened 2020-10-26
Make remoteCacheId content-based

### LINKED ISSUE #5842: Make remoteCacheId content-based
## steps

Assuming subprojects `app` and `core`

```
pushRemoteCache
// make some changes to app
// git commit -m "wip"
pushRemoteCache
```

## problem

This would create a new remote cache for all subprojects.

## expectation

Ideally, we should generate remoteCacheId based content hash, so if I didn't change `core`, the remoteCacheId remains the same, and thus I don't need extra network IO to push/pull the same JARs.

eed3si9n:
Fixes #5842

Using Git commit for remote caching id was one of the first things people think about, but it doesn't scale because that means you're downloading all module JARs all the time anytime anyone adds a commit. The basis of remote caching is that people are adding commits to many subprojects, so that's bad.

As an alternative, this uses content hash of input files and external dependency JARs. `remoteCacheId`, `pushRemoteCache`, and `pullRemoteCache` are now scoped to configuration (`Compile` and `Test`). Since we're not resolving any version conflicts, dependency resolver is no longer used and the artifacts are directly downloaded from the Ivy resolver. This should speed up the `pullRemoteCache` since it avoids the POM download as well.

### note

For sbt-binrary-remote-cache this created a bit of complication since the (publishing) resolver doesn't act correctly as (downloading) resolver in terms of the credentials, so I had to create a new key `remoteCacheResolvers` to have asymmetric resolver.

### DIFF
--- a/main/src/main/scala/sbt/internal/RemoteCache.scala
+++ b/main/src/main/scala/sbt/internal/RemoteCache.scala
@@ -9,6 +9,7 @@ package sbt
 package internal
 
 import java.io.File
+import java.nio.file.Path
 import Keys._
 import SlashSyntax0._
 import Project._ // for tag and inTask()
@@ -17,11 +18,14 @@ import sbt.coursierint.LMCoursier
 import sbt.librarymanagement._
 import sbt.librarymanagement.ivy.Credentials
 import sbt.librarymanagement.syntax._
+import sbt.nio.FileStamp
+import sbt.nio.Keys.{ inputFileStamps, outputFileStamps }
 import sbt.internal.librarymanagement._
 import sbt.io.IO
 import sbt.io.syntax._
 import sbt.internal.remotecache._
-import sbt.internal.inc.JarUtils
+import sbt.internal.inc.{ HashUtil, JarUtils }
+import sbt.util.InterfaceUtil.toOption
 import sbt.util.Logger
 
 object RemoteCache {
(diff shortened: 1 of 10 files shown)


main/src/main/scala/sbt/internal/RemoteCache.scala
```diff
@@ -41,12 +45,30 @@ object RemoteCache {
...
+    remoteCacheIdCandidates := Nil,
     pushRemoteCacheTo :== None
   )
 
   lazy val projectSettings: Seq[Def.Setting[_]] = (Seq(
+    remoteCacheId := {
+      val compileExtraInc = (Compile / extraIncOptions).value
+      val compileInputs = (Compile / unmanagedSources / inputFileStamps).value
+      val compileCp = (Compile / externalDependencyClasspath / outputFileStamps).value
+      val testExtraInc = (Compile / extraIncOptions).value
```
bjaglin:
```suggestion
      val testExtraInc = (Test / extraIncOptions).value
```

main/src/main/scala/sbt/internal/RemoteCache.scala
```diff
@@ -41,12 +45,30 @@ object RemoteCache {
...
+    remoteCacheId := "",
+    remoteCacheIdCandidates := Nil,
     pushRemoteCacheTo :== None
   )
 
   lazy val projectSettings: Seq[Def.Setting[_]] = (Seq(
+    remoteCacheId := {
+      val compileExtraInc = (Compile / extraIncOptions).value
+      val compileInputs = (Compile / unmanagedSources / inputFileStamps).value
+      val compileCp = (Compile / externalDependencyClasspath / outputFileStamps).value
```
bjaglin: Could this be defined at the configuration level (and combined at the project level via a scope filter), so that other configurations such as IntegrationTest can more easily be supported? https://github.com/sbt/sbt/blob/5a960f2f94b8c5260e07e380b76e88f1d6959cd4/main/src/main/scala/sbt/internal/RemoteCache.scala#L206-L207 would still need to be exposed, maybe within the same `Seq[Def.Setting[_]]` ?

main/src/main/scala/sbt/internal/RemoteCache.scala
eed3si9n: Thanks!

main/src/main/scala/sbt/internal/RemoteCache.scala
eed3si9n: Good idea. I'm hoping I can eventually implement `pullRemoteCache` at the configuration level.

PR #6028, opened 2020-10-26
Add serverIdleTimeout to the list of excluded lint keys

adpi2:


main/src/main/scala/sbt/internal/LintUnused.scala
```diff
@@ -39,6 +39,7 @@ object LintUnused {
       sbt.nio.Keys.watchTriggers,
       serverConnectionType,
       shellPrompt,
+      serverIdleTimeout
```
eatkins: Let's keep these alphabetical.

eatkins (changes requested):

eatkins (approved):
LGTM

eed3si9n (approved):

PR #6030, opened 2020-10-26
Support -bsp in sbtn

eatkins:
This PR makes `sbtn -bsp` work the same as `sbt -bsp`.


eatkins:
I also verified that `sbt --client -bsp` worked with sbtn if I set the sbtn executable to be a locally built on in my sbt bash script.

eed3si9n (approved):

PR #6036, opened 2020-10-30
Updates to Coursier 2.0.6

jtjeferreira:
https://github.com/coursier/sbt-coursier/releases/tag/v2.0.2 which includes https://github.com/coursier/coursier/releases/tag/v2.0.5
https://github.com/coursier/coursier/releases/tag/v2.0.6

- Computed checksums of downloaded files are now written in the cache, and read from there later on. This short-circuits checksum computation for anything already in cache, which can significantly speed up resolutions with large dependency graphs.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -76,7 +76,7 @@ object Dependencies {
   def addSbtZincCompile = addSbtModule(sbtZincPath, "zincCompileJVM2_12", zincCompile)
   def addSbtZincCompileCore = addSbtModule(sbtZincPath, "zincCompileCoreJVM2_12", zincCompileCore)
 
-  val lmCoursierShaded = "io.get-coursier" %% "lm-coursier-shaded" % "2.0.0"
+  val lmCoursierShaded = "io.get-coursier" %% "lm-coursier-shaded" % "2.0.2"
 
   def sjsonNew(n: String) =
     Def.setting("com.eed3si9n" %% n % "0.9.1") // contrabandSjsonNewVersion.value
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -76,7 +76,7 @@ object Dependencies {
   def addSbtZincCompile = addSbtModule(sbtZincPath, "zincCompileJVM2_12", zincCompile)
   def addSbtZincCompileCore = addSbtModule(sbtZincPath, "zincCompileCoreJVM2_12", zincCompileCore)
 
-  val lmCoursierShaded = "io.get-coursier" %% "lm-coursier-shaded" % "2.0.2"
+  val lmCoursierShaded = "io.get-coursier" %% "lm-coursier-shaded" % "2.0.3"
 
   def sjsonNew(n: String) =
     Def.setting("com.eed3si9n" %% n % "0.9.1") // contrabandSjsonNewVersion.value


jtjeferreira:
some tests are failing, maybe some regression from https://github.com/coursier/coursier/pull/1801/? cc @alexarchambault @oyvindberg

oyvindberg:
Seems fairly probable @jtjeferreira . I had a quick look at `cached-resolution-circular`, and it seems to write the artifact twice somehow. Perhaps we're missing some invalidation of the cached checksum on artifact overwrite? the other failing test was also snapshot related

jtjeferreira:
> Seems fairly probable @jtjeferreira . I had a quick look at `cached-resolution-circular`, and it seems to write the artifact twice somehow. Perhaps we're missing some invalidation of the cached checksum on artifact overwrite? the other failing test was also snapshot related

Thanks for having a look!

oyvindberg:
So it seems we'll have to do one more iteration of storing caching digests @alexarchambault .

I haven't managed to run the sbt tests myself yet, as I'm strangely not able to make sbt itself compile on my machine any more. 

But just thinking about it:
- publishing from sbt seems to be done through ivy (right?). I don't think we should modify ivy.
- storing computed digests should only be done when the artifact is not changing, or if we can invalidate the stored digest on artifact change. 

As a result I see a few possible paths forward:
1) don't store computed digests for SNAPSHOT dependencies
2) only store computed digests immediately after coursier has downloaded the files. So if the files are not downloaded from (a new enough version of) coursier they will always be redigested.
3) invalidate aux files if they have an older timestamp than the main file. (edit: I originally wrote the opposite)

My personal preference is 1. 2 seems like it would break the coursier model of being able to change which type of digest to use, and 3 would incur quite a bit of extra IO during resolving (though not sure if it would matter much)

Thoughts?

alexarchambault:
I'm not fully sure of what's going on here, I only glimpsed at the CI output and read your comments. It seems we may be writing "auxiliary files" outside of the cache, in the local repository used in the failing scripted tests. Is it your understanding or what you meant, @oyvindberg?

If that's the problem, I guess we should ensure we don't write any auxiliary file outside of the cache. There had been a similar problem in the past, that's why [this check](https://github.com/coursier/coursier/blob/adf62e28488d4ff6efa2db8ed654b052f22f3ace/modules/cache/jvm/src/main/scala/coursier/cache/FileCache.scala#L820) was added (without this check, we were deleting files in local repositories if they had wrong checksums 😬).

Running `cs fetch` for things in my `~/.ivy2/local`, I confirm some `.*__sha1.computed` files get written there. We should just ensure they aren't. I guess that should fix the problem here.

alexarchambault:
When we download snapshots, auxiliary files are cleared around [here](https://github.com/coursier/coursier/blob/adf62e28488d4ff6efa2db8ed654b052f22f3ace/modules/cache/jvm/src/main/scala/coursier/cache/FileCache.scala#L396). So I don't think snapshots are a problem per se.

oyvindberg:
Aha! I hadn't really considered the difference between `~/.ivy2` and `~/.cache/coursier` much.

I'll go ahead and try to add a similar check so that we only write these aux files to the coursier folder. And good if snapshots are already covered, that was just my first hunch.

oyvindberg:
@jtjeferreira can you bump to 2.0.6 please?

eed3si9n (approved):
Thanks @jtjeferreira and Coursier team for the quick turnaround.

PR #6041, opened 2020-11-02
Start server if active.json file is removed

eatkins:
If there are two sbt instances and one of them is running a server, the
other instance is presently prevented from ever starting a server. If an
sbt instance is unable to start a local server because of the presence
of another server, we can monitor the active.json file for changes and,
if it is deleted, we can then try again to start a new server instance.

### DIFF
--- a/main/src/main/scala/sbt/internal/CommandExchange.scala
+++ b/main/src/main/scala/sbt/internal/CommandExchange.scala
@@ -50,6 +50,7 @@ private[sbt] final class CommandExchange {
     sys.props get "sbt.server.autostart" forall (_.toLowerCase == "true")
   private var server: Option[ServerInstance] = None
   private val firstInstance: AtomicBoolean = new AtomicBoolean(true)
+  private val monitoringActiveJson: AtomicBoolean = new AtomicBoolean(false)
   private var consoleChannel: Option[ConsoleChannel] = None
   private val commandQueue: LinkedBlockingQueue[Exec] = new LinkedBlockingQueue[Exec]
   private val channelBuffer: ListBuffer[CommandChannel] = new ListBuffer()
@@ -188,6 +189,7 @@ private[sbt] final class CommandExchange {
     lazy val connectionType = s.get(serverConnectionType).getOrElse(ConnectionType.Tcp)
     lazy val handlers = s.get(fullServerHandlers).getOrElse(Nil)
     lazy val win32Level = s.get(windowsServerSecurityLevel).getOrElse(2)
+    lazy val portfile = s.baseDir / "project" / "target" / "active.json"
 
     def onIncomingSocket(socket: Socket, instance: ServerInstance): Unit = {
       val name = newNetworkName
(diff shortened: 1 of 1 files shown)


eatkins:
After this change, if I start an sbt instance and then start a second sbt instance, the second sbt instance logs a warning about being unable to start a server. As soon as I exit the initial instance, I see a log message that a server has been created in the second instance.

eed3si9n:
I feel like this tacitly admits the multiple instances of sbt running on the same directory.
None of the tasks are designed to work in that mode, and I don't think we should throw that illicit state any bones.

eatkins:
I'm not sure I understand. If someone starts sbt while a server is already running do you think that we should just abort sbt entirely? Right now, for example, metals calls `sbt bloopInstall` to reimport the build. Similarly, I think intellij generally makes a new sbt instance for import.

There can be a lot of potential problems with the running server, what if it has become unresponsive due to a memory leak and isn't responding to `sbtn shutdown`? Maybe I want to start a new one while I track down what the pid is of the offending process and kill it.

Sometimes I want to have the running server in the foreground. If I start a new sbt instance not realizing that I already had a server running, then I can never connect to the new instance. Right now, my option then is to kill both the old and the new process and start over. The idea behind this is that I can just kill the old server with sbtn shutdown and now my foregrounded sbt process can become the new server. I think this is perfectly valid and not illicit.

eed3si9n:
What I mean is that there are people who think actively running two build sessions is ok like:

```
sbt server/run&
sbt client/test
```

This is in fact not a safe use case since we can't guarantee the caching or state each task is in, and could result to bad JAR being tested or published.

> If someone starts sbt while a server is already running do you think that we should just abort sbt entirely?

The single-process sbt should print a warning that they are in an unsupported/undefined state.
https://github.com/sbt/sbt/blob/25f492d9195dd7b1c8365e137278e8ca7e088951/main/src/main/scala/sbt/internal/CommandExchange.scala#L238

> Right now, for example, metals calls `sbt bloopInstall` to reimport the build. Similarly, I think intellij generally makes a new sbt instance for import.

so if we were to create an issue for this, the problem statement would be that "IDEs are starting a new instance"?

eatkins:
> so if we were to create an issue for this, the problem statement would be that "IDEs are starting a new instance"?

I think we are talking about different problem statements. I 100% agree with you that running concurrent sbt instances in batch mode is a bad idea. The scenario I'm talking about though is suppose that metals used `sbt -client bloopInstall` instead of `sbt bloopInstall`. Now I might have a server instance started that I didn't necessarily know about. Then I go to start a foreground server for whatever reason. I do this personally because there is additional debugging information that I can see with the foreground server. But after I start the foreground server, I discover that I can't connect to it. With this change, I can run `sbtn shutdown` to get rid of the background server that I don't want anymore and my foreground server can become the connectable via sbtn.

Here's another use case: I made the mistake of actually starting a bunch of sbt server instances in a number of tmux panes and I want to close them all. Prior to this change, I might be able to shutdown one of them with `sbtn shutdown` but all of the rest of them will remain up and I will either have to find them in the tmux panes or with `ps`. After this change, I could just keep running `sbtn shutdown` until there are no servers left.

eed3si9n:
In either of the cases I'd probably run `killall java`, but I'm not going to block this PR.

eed3si9n (approved):

OndrejSpanel:
> In either of the cases I'd probably run `killall java`, but I'm not going to block this PR.

Good. :relieved:, because perhaps some people may be sometimes running other Java processes than sbt. :wink:

Issue #6043, opened 2020-11-02
sbt 1.4.2 prints control characters for colors with sb.log.noformat=true


wgreven-ibr:
## steps

In Jenkins, run sbt 1.4.2 with `sbt.log.noformat=true`.

## problem

When running sbt 1.4.2 in Jenkins with `sbt.log.noformat=true` control characters are printed.

```
+ sbt -Dsbt.log.noformat=true compile
[0m[[0m[0minfo[0m] [0m[0mwelcome to sbt 1.4.2 (Private Build Java 1.8.0_272)[0m
[0m[[0m[0minfo[0m] [0m[0mloading settings for project global-plugins from build.sbt,metals.sbt ...[0m
[0m[[0m[0minfo[0m] [0m[0mloading global plugins from /home/wg01018/.sbt/1.0/plugins[0m
[0m[[0m[0minfo[0m] [0m[0mloading project definition from /tmp/project[0m
[0m[[0m[0minfo[0m] [0m[0mset current project to tmp (in build file:/tmp/)[0m
[0m[[0m[0minfo[0m] [0m[0mExecuting in batch mode. For better performance use sbt's shell[0m
[0m[[0m[32msuccess[0m] [0m[0mTotal time: 0 s, completed Nov 2, 2020 7:25:28 PM[0m
```

## expectation

I would have expected similar output as when running with `sbt.color=false`.

```
+ sbt -Dsbt.color=false compile
[info] welcome to sbt 1.4.2 (Private Build Java 1.8.0_272)
[info] loading settings for project global-plugins from build.sbt,metals.sbt ...

[info] loading global plugins from /home/wg01018/.sbt/1.0/plugins

[info] loading project definition from /tmp/project
[info] set current project to tmp (in build file:/tmp/)
[info] Executing in batch mode. For better performance use sbt's shell
[success] Total time: 0 s, completed Nov 2, 2020 7:25:37 PM
```

eed3si9n:
This was fixed in #6044 by @eatkins

PR #6044, opened 2020-11-02
Fixes sbt --no-colors

eatkins:
If a user runs with -Dsbt.log.format=false or -Dsbt.log.noformat=true,
we should disable color by default. Running with -Dsbt.color=true should
make this possible.

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
@@ -317,7 +317,7 @@ object Terminal {
     props
       .map(_.color)
       .orElse(isColorEnabledProp)
-      .getOrElse((hasConsole && !isDumbTerminal && logFormatEnabled.getOrElse(true)) || isCI)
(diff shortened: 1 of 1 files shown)


eatkins:
Fixes #6043

eed3si9n:
Could we somehow abstract this into a pure function that we can write a unit test?

eatkins:
We could but I am not sure the investment is worth it for me personally.

eatkins:
That was a bit glib. The reason that I'm not excited about writing a unit test is that it would require a non-trivial refactor of `Terminal` because these variables are stored in global fields in the `Terminal` object and set during initialization of the object. On top of that, I'd have to actually write the unit tests. All told, I'd expect that to take at least 3-4 hours whereas it took like 5 minutes to just make this fix. If there was an expectation that we are constantly going to be introducing regressions in the setting of these parameters, I'd be all for the unit testing but I suspect that once this is ironed out in a release or two, that this code will rarely if ever be touched.

eed3si9n:
I'll merge as-is and create a new issue for it.

Issue #6042, opened 2020-11-02
Unclosed files carry over between scripted tests (at least on windows)


raboof:
## steps

Since 1.4.0, the scripted tests for paradox at appveyor fail with:

```
[info] Running 19 / 19 (100.00%) scripted tests with RunFromSourceMain
[info] Running paradox/custom-configuration
[info] Running paradox/custom-directive
[info] Running paradox/default-theme
[info] [info] welcome to sbt 1.4.0 (Oracle Corporation Java 1.8.0_162)
[info] [info] loading settings for project sbt_9f2594d5-build from plugins.sbt ...
[info] [info] loading project definition from C:\Users\appveyor\AppData\Local\Temp\1\sbt_9f2594d5\project
[info] [info] loading settings for project docs from build.sbt ...
[info] [info] set current project to Paradox Default Theme Test (in build file:/C:/Users/appveyor/AppData/Local/Temp/1/sbt_9f2594d5/)
[info] [error] Unexpected top-level pages (pages that do no have a parent in the Table of Contents).
[info] [error] If this is intentional, update the `paradoxRoots` sbt setting to reflect the new expected roots.
[info] [error] Current ToC roots: [index.html, sub/unindexed.html, test.html]
[info] [error] Specified ToC roots: [index.html, sub/unindexed.html]
[info] [error] Paradox failed with 1 errors
[info] [error] (Compile / paradoxMarkdownToHtml) com.lightbend.paradox.sbt.ParadoxPlugin$ParadoxException
[info] [error] Total time: 1 s, completed Oct 5, 2020 1:54:39 AM
[error] x paradox/default-theme 
[error]  Cause of test exception: {line 1}  Command failed: paradox failed
```

The `paradox/default-theme` test fails because it finds an unexpected `test.html`. `test.html` is an expected output of the previous scripted test, `paradox/custom-directive`.

## problem

Artifacts from one scripted test appear to leak into the next. I'm not sure yet if it's the sources or the target resources - I don't have a windows machine handy to easily test this with, and it works fine on linux...

## expectation

I expected scripted tests to be independent.

eatkins:
This looks like a problem with paradox, not sbt. That being said, scripted tests generally run in alphabetical order on windows and in a random order on other platforms which is probably why you see the problem only on windows.

raboof:
I'll see if I can come up with a more targeted reproducer (or indeed confirm this as a problem with paradox, though I have my doubts ;) )

eatkins:
[...]

PR #6046, opened 2020-11-02
Add scripted option to toggle sbt instance reuse

eatkins:
In sbt 1.3.x, every scripted test got its own server. I am not exactly
sure why that behavior changed in 1.4.0, but we now by default reuse the
same sbt server in scripted tests. This is a very nice performance
benefit that I think most builds benefit from. However, there are some
cases where reusing the sbt server can cause problems. In windows, in
particular, it may be the case that there are files that sbt is not able
to delete in the project directory that may leak from one scripted test
into another. To avoid this, this commit adds the scriptedReuseServer
parameter which when set to false allows the user to rever their build
to the sbt 1.3.x semantics which are much slower.

This came from #6042. With reuseServer true, the paradox scripted tests
take about 1 minute on my computer. With reuseServer false, they take
about 5 minutes.

### DIFF
--- a/main/src/main/scala/sbt/ScriptedPlugin.scala
+++ b/main/src/main/scala/sbt/ScriptedPlugin.scala
@@ -47,6 +47,7 @@ object ScriptedPlugin extends AutoPlugin {
     val scriptedLaunchOpts =
       settingKey[Seq[String]]("options to pass to jvm launching scripted tasks")
     val scriptedDependencies = taskKey[Unit]("")
+    val scriptedReuseServer = settingKey[Boolean]("")
     val scripted = inputKey[Unit]("")
   }
   import autoImport._
@@ -80,6 +81,7 @@ object ScriptedPlugin extends AutoPlugin {
     scriptedParallelInstances := 1,
     scriptedBatchExecution := false,
     scriptedRun := scriptedRunTask.value,
+    scriptedReuseServer := true,
     scriptedDependencies := {
       def use[A](@deprecated("unused", "") x: A*): Unit = () // avoid unused warnings
       val analysis = (Keys.compile in Test).value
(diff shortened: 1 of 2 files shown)


scripted-sbt-redux/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
```diff
@@ -141,7 +151,7 @@ final class ScriptedTests(
 
       def createTestRunners(tests: Seq[TestInfo]): Seq[TestRunner] = {
         tests
-          .grouped(batchSize)
+          .grouped(if (reuseServer) batchSize else 1)
```
eatkins, line 154: This is the key line of the PR. Everything else is just to get this parameter passed in through all of the various entry points.

main/src/main/scala/sbt/ScriptedPlugin.scala
```diff
@@ -188,10 +190,22 @@ object ScriptedPlugin extends AutoPlugin {
       val opts = scriptedLaunchOpts.value.toArray
       val empty = new java.util.ArrayList[File]()
       val instances = Int box scriptedParallelInstances.value
+      val reuseServer = Boolean box scriptedReuseServer.value
 
       if (scriptedBatchExecution.value)
-        method.invoke(scriptedInstance, dir, log, args.toArray, launcher, opts, empty, instances)
-      else method.invoke(scriptedInstance, dir, log, args.toArray, launcher, opts, empty)
```
eed3si9n, line 194: Doesn't the semantics of `scriptedBatchExecution.value` already mean that we're reusing the server?
The only new thing is that batch execution became available to sbt plugin authors?

eatkins: I could be mistaken, but it appeared in my testing that the semantics of scriptedBatchExecution seem to be that it will spin up a bunch of sbt servers in parallel and run all the scripted tests. This is very bad because they all contend with each other for the ivy lock.

eed3si9n: If I understand correctly, the "batching" was named so because it groups scripted tests into some batches, and run them together, AND those batches run in parallel. Whereas traditional scripted test would spin up a new sbt server with a cold JVM per test sequentially. Thus, the idea of reusing occurs only during batching.

eatkins: Got it. So somehow during refactoring, that behavior was changed and batch execution became the default with no parallelism. This is a really nice improvement for most use cases except where there are problems clearing the workspace like in the paradox use case. It seems that the real issue then is to get scripted to honor the scriptedBatchExecution flag.

eatkins:
Closing in favor of #6054.

PR #6051, opened 2020-11-03
Fixes IntelliJ import on Windows + Fixes shell when supershell is disabled

eatkins:
Intellij import was broken in sbt 1.4.2 because we increased the
scenarios in which virtual io is used. The problem wasn't actually
virtual io but that we create a console terminal with jline3 and that
could cause issues reading input from System.in. This can be fixed by
only creating an instance of ConsoleTerminal if a console is actually
available for the sbt process. When hasConsole is false, the
consoleTerminalHolder will point to the SimpleTerminal whose inputStream
method just reads directly from System.in. After making this change, I
verified that intellij import worked again on windows.


eed3si9n (dismissed):

jastice:
fix for https://youtrack.jetbrains.com/issue/SCL-18393

PR #6054, opened 2020-11-03
Fixes opt-out for scripted batch execution

eatkins:
sbt 1.4.0 made batch execution the default behavior. The following setting can be used to opt out of the behavior:

```scala
scriptedBatchExecution := false
```

To increase parallelism use:

```scala
scriptedParallelInstances := 2
```

### original description

This PR makes the scripted plugin work the way it used to in older sbt versions when `scriptedBatchExecution := false`. It also makes `scriptedBatchExecution := true` the default because I think the performance improvement is worth it for most builds even though it can cause issues like in #6042. I can remove the second commit if there is disagreement on changing the default.

### DIFF
--- a/scripted-sbt-redux/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
+++ b/scripted-sbt-redux/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
@@ -376,7 +376,7 @@ class ScriptedRunner {
       launchOpts: Array[String],
   ): Unit = {
     val logger = TestConsoleLogger()
-    runInParallel(
+    run(
       resourceBaseDirectory,
       bufferLog,
       tests,
@@ -384,7 +384,8 @@ class ScriptedRunner {
       launchOpts,
       prescripted = new java.util.ArrayList[File],
       LauncherBased(launcherJar),
-      1
+      1,
+      parallelExecution = false,
     )
   }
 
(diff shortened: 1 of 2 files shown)


scripted-sbt-redux/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
```diff
@@ -492,9 +492,13 @@ class ScriptedRunner {
...
-    val parallelRunners = scriptedRunners.toParArray
-    parallelRunners.tasksupport = new ForkJoinTaskSupport(new ForkJoinPool(instances))
-    runAll(parallelRunners)
+    if (instances > 1 && instances != Int.MaxValue) {
+      val parallelRunners = scriptedRunners.toParArray
+      parallelRunners.tasksupport = new ForkJoinTaskSupport(new ForkJoinPool(instances))
+      runAll(parallelRunners)
+    } else {
+      runAll(scriptedRunners)
+    }
```
eed3si9n: Minor nit pick. `-1` may be a better value for "I'm gonna do sequential even tho the method is called runInParallel" since I'd expect `Int.MaxValue` to go max. Scaladoc explaining this would be nice. Overall LGTM.

eed3si9n (dismissed):

scripted-sbt-redux/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
eatkins: Yeah, Int.MaxValue wasn't a perfect sentinel. I reworked this to instead make a new `run` variant that has a `parellelExecution` flag rather than using a sentinel.

eed3si9n (dismissed):

PR #6059, opened 2020-11-04
Remove scalatest_2.11-3.0.5.jar from travis cache

eatkins:
The scripted tests/test-cross test is failing on travis and it seems to
be due to bad cache entry for the scalatest_2.11-3.0.5.jar. The travis
ci ui supposed to allow you to delete cache entries, but that feature
seems to be broken in sbt/sbt. This commit could probably be reverted
once it's merged assuming it works.


eatkins:
That didn't seem to work.

PR #6061, opened 2020-11-04
Support ansi by default if color is supported

eatkins:
Scalatest will check the ansiCodesSupported value of a console appender
to decide whether or not to colorize its output. We were setting
isAnsiSupported to false by default in ci to prevent the ConsoleAppender
from adding ansi codes to the output. The only place though where we
were doing that was in adding a ClearScreenAfterCursor to the end of
each log line. This was done for supershell but there is actually other
logic in supershell processing that adds these anyway.


eatkins:
The motivation for this was https://github.com/sbt/sbt/issues/5958#issuecomment-721927665

eatkins:
ci should pass if #6060 is merged and this branch is updated.

eatkins:
Folding this into #6051

PR #6074, opened 2020-11-06
Don't block sbt in `run` and `console` tasks

eatkins:
Currently when the user invokes run or console, these tasks tie up the
sbt task evaluation loop until they complete. This could be problematic
if the user is running, say, a server or an interactive console in their
sbt shell while also using metals with bsp. Metals would be unable to
compile any source files while these tasks were blocking. We can fix
this by running these tasks in the background in such a way that they
block the ui thread for the console or thin client session that
initiated them but not block the task evaluation loop. This is only safe
to do if we copy the classpath for these tasks so that it isn't blown
away by other tasks. In the case of run and runMain, this was already
done. The console tasks had to be reworked to use the same bgJobService
infrastructure as the run tasks.

Since there is the possibility that this causes unforeseen issues, there
is a new setting key, runInBackground, that if set to false will revert
to the old way of evaluating the run and console tasks. We also will
only run these tasks in the background if there is a server running.
Otherwise there is no point. This makes this change significantly less
risky than it might have otherwise been.

The trickiest part of the implementation to figure out was how to ensure
that System.out and System.in worked correctly for the background tasks.
Previously since only one instance of run or console could be run at a
time, we could just set a global variable for the current terminal and
redirecty System.out and System.in to the terminal's input and output
streams. Instead what we do here is we set an inheritable threadlocal
value for the task terminal. Because it is inheritable, if the run task
spawns additional threads, they should still get the same value. To set
the threadlocal terminal value correctly, it was necessary to propagate
more information about the terminal that initiated the currently running
command.

The final piece of the puzzle is that there needed to be some way of
finishing run and console without actually reporting their results. The
way this is implemented is by adding a new kind of Prompt called
Blocked. If run or console is running in the background, it will set its
prompt to Blocked before exiting. MainLoop will then strip all of the
remaining commands for that terminal and store them in the instance of
Blocked. It will also defer reporting the results. We also need to
modify Aggregation so that it does not print the timing results when run
and console exit since they haven't actually completed. A function that
prints the timing results is instead stored in the Blocked prompt and
printed when the task completes. When sbt returns to the main thread, it
reprompts the terminal that initiated the run or console task. Instead
of making an AskUserThread, we instead start a thread that blocks until
the run or console task completes. When it does, we print the timing
results and then re-enque whatever commands had been stripped by
MainLoop when it detected the blocked prompt. Reporting for the thin
client is handled by running a number of commands after the actual
command completes so this is a critical component to ensuring that the
thin client exits at all in batch mode and that it returns an error code
depending on the success or failure of the task.

Although the motivation of this change was to allow things like metals
compile in the background while a console is running, a side effect is
that multiple clients may run interactive console sessions
simultaneously on the same server.

This approach could be extended to test, but test is a less urgent case
because unlike run and console, it is extremely rare that test would
indefinitely block the task evaluation engine. Tests can take a while to
run, but at least they can be expected to exit in the vast majority of
use cases. At any rate, the same approach used here with the
bgJobService and BackgroundRunTask could be applied to the various test
tasks as well.


eatkins:
Here is a demo of multiple instance of console and run being evaluated simultaneously with the same server: [asciicast](https://asciinema.org/a/wLxmXI0EKwWZy9r1jZdU2Nt5m)

eatkins:
Converted to draft while I sort out CI since I can't test on my personal travis anymore.

ritschwumm:
would this subsume sbt-revolver?

eatkins:
Not exactly. If you are willing to have multiple terminals windows open, you could use this as a replacement for sbt-revolver. In one window you could do something like `sbtn run` which will block that terminal until `run` exits or you kill the process or the terminal entirely. In the meantime, in another terminal you could interact with the same sbt server.

It isn't really documented, but sbt has had built in functionality that is somewhat similar to sbt-revolver since 1.0.0. You can run `bgRun` or `bgRunMain` and sbt will run you application in the background. The running application has an associated pid that you can find by running `ps`. You can stop the running background application with `bgStop $pid`. This is pretty similar to what you get with `reStart` and `reStop`.

What makes this different is that from the ux perspective, `run` is still blocking. The issue was that sbt's task evaluation engine can only run one command at a time which would potentially block metals from recompiling source files if you had done a blocking `run`. Now `run` will not block the task evaluation engine so that metals .

ritschwumm:
ah, that clears it up. thanks for the explanation!

eatkins:
I sorted out CI.

smarter:
Wow, impressive! This addresses my biggest concern with the BSP implementation in sbt so I'm really glad to see this, many thanks for working on this!

> This approach could be extended to test, but test is a less urgent case
because unlike run and console, it is extremely rare that test would
indefinitely block the task evaluation engine.

Agreed. Somewhat off-topic for this PR but: the other task that is used frequently and might run for a long time is `compile`, but I guess for metals to work when `compile` is being run in parallel, it would need to use its own output directory, I think that makes sense anyway because metals needs the code to be compiled with semanticdb, whereas a command-line compilation shouldn't be slowed down by the semanticdb generation.

tartarza (approved):

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1849,15 +1863,31 @@ object Defaults extends BuildCommon {
...
-      service.waitForTry(handle).get
+      val log = streams.value.log
+      val scope = resolvedScoped.value
+      val st = state.value
+      val background = runInBackground.value && StandardMain.exchange.hasServer
+      st.get(Keys.taskTerminal) match {
+        case Some(term) if background =>
+          BackgroundRunTask(term, "fg-run", log, st, scope)(() => service.waitForTry(handle))
+        case _ => service.waitForTry(handle).get
+      }
```
eed3si9n, line 1890: I wonder if this could be implemented at the command processing or client / `CommandChannel` level. Like if a client sees `run` key, it would actually dispatch `bgRun` task and then block only the channel. This allows remaining channels to funnel commands into the exchange.

eatkins: That is a potential alternative approach. I think that was my first idea but I ruled it out. Most of the logic of this pr is designed to address these two issues:
1. How do you ensure that the ui doesn’t switch to the shell after the bg job stops
2. How do you defer the completion of the exec until the bg job has completed?

If we just did a simple command substitution of run to bgrun, both of those issues crop up.

eed3si9n: > 1. How do you ensure that the ui doesn’t switch to the shell after the bg job stops

In the original sbt server circa sbt 1.1.0, there was `ConsoleUnpromptEvent` (https://github.com/sbt/sbt/blob/v1.1.0/main-command/src/main/scala/sbt/internal/CommandChannel.scala) and the idea was that if one channel queues a command like `compile`, we would have the ability to block out all the clients as if they have also typed in `compile`. This use case sound similar to that.

> 2. How do you defer the completion of the exec until the bg job has completed?

This I'm not sure. Exec could end after bg job is done. If there's a BSP events attached to it, then those would need to be deferred..

Another more radical idea might be to fork from the client. Traditionally sbt has always emulated `run` as function calls in sandbox ClassLoader, but for the longevity of server process, it might be more stable to only calculate the copied classpath, and let each client do the running however they please.

main/src/main/scala/sbt/MainLoop.scala
```diff
@@ -147,7 +148,12 @@ object MainLoop {
     }
 
   def next(state: State): State = {
-    val context = LoggerContext(useLog4J = state.get(Keys.useLog4J.key).getOrElse(false))
+    val exchange = StandardMain.exchange
+    val taskTerminal = state.remainingCommands.headOption
+      .flatMap(_.source.flatMap(s => exchange.channelForName(s.channelName).map(_.terminal)))
+      .getOrElse(ITerminal.get)
+    val context =
+      LoggerContext(useLog4J = state.get(Keys.useLog4J.key).getOrElse(false), taskTerminal)
```
eed3si9n, line 156: +1 on this part of the code. Not sure about the name `taskTerminal` though. `channelTerminal`?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -394,6 +394,12 @@ object Defaults extends BuildCommon {
       terminal := state.value.get(terminalKey).getOrElse(Terminal(ITerminal.get)),
       InstallSbtn.installSbtn := InstallSbtn.installSbtnImpl.evaluated,
       InstallSbtn.installSbtn / aggregate := false,
+      runInBackground := {
+        // We only want to run console or run in the background if we are
+        // not in batch mode which can be implied by the absent of non-flag
+        // command arguments
+        appConfiguration.value.arguments.filterNot(a => a.startsWith("-")).isEmpty
+      },
```
eed3si9n, line 402: I don't think we should introduce behavior difference between shell and batch.

internal/util-logging/src/main/scala/sbt/internal/util/Prompt.scala
```diff
@@ -36,4 +37,60 @@ private[sbt] object Prompt {
   private[sbt] case object Watch extends NoPrompt
   private[sbt] case object Pending extends NoPrompt
   private[sbt] case object NoPrompt extends NoPrompt
+  private[sbt] final class Blocked(doJoin: () => Boolean) extends NoPrompt {
```
eed3si9n, line 40: I feel like this construct sticks out since all other `Prompt` states represents a state, and this one is like an underground tunnel between tasks, commands, and virtual IO that's acting like a global variable?
I wonder if some of these blocking imple should move to the individual channel that's doing the blocking?

main/src/main/scala/sbt/internal/BackgroundRunTask.scala
```diff
@@ -0,0 +1,60 @@
...
+
+package sbt
+package internal
+
+import BasicCommandStrings.CompleteExec
+import sbt.internal.util.{ Prompt, Terminal, Util }
+import sbt.util.Logger
+import scala.util.Try
+
+private[sbt] object BackgroundRunTask {
```
eed3si9n, line 16: I am trying to think of a better name for this. `OnAfterJobHandler`? `BackgroundRunTask` itself runs somewhere else, but this does the cleanup / UI handling to emulate as if it ran normally correct?

internal/util-logging/src/main/scala/sbt/internal/util/Prompt.scala
eatkins: I agree. This aspect is not particularly well designed. The challenge is that a background job could exit it any point so it's kind of tricky to integrate with the sbt main loop. It doesn't really make sense as an sbt state variable because it needs to be managed outside of the logic in main loop. I couldn't think of a great way to handle this problem and this was the best hack that I came up with. I have no doubt there is a better way. I'm just not sure what it is.

main/src/main/scala/sbt/Defaults.scala
eatkins: You're on to something with forking from the client though I think it would make more sense to fork from the server. The ideal approach for testing and run is always `fork := true` so that the state of sbt doesn't affect run and test. The problem, of course, is that the performance is much worse in many cases. A workaround would be to create a test/run daemon that has an ipc or socket connection with the server. The daemon would create its own classloader cache and we could send it the classpath layers (probably using a json protocol) so that the performance of running on the daemon would be indistinguishable from running in process with classloader caching. But there also could be logic for restarting the daemon. For example, we could easily detect a memory leak in the daemon and kill it if it using more memory after a few runs. We also could add an idle timeout for the daemon.

The general solution is that sbt should fork a remote process for every task. sbt is kind of backwards in that in-process evaluation is the default but it is actually a special case of the general solution.

eatkins:
I converted this to draft because @eed3si9n and I have decided it would be best for someone other than me to implement this functionality so that there can be a broader understanding of the internals of how the thin client works. This PR can be a starting point for reimplementation or simply just a proof of concept.

PR #6093, opened 2020-11-16
Fix #6092 - sbt 1.4.0~1.4.3 - all the files in the packaged jar files created at epoch millisecond 0 (1970-01-01 00:00:00)

### LINKED ISSUE #6092: sbt 1.4.0~1.4.3 - all the files in the packaged jar files created at epoch millisecond 0 (1970-01-01 00:00:00)
## steps

Simply
* `sbt package` or
* `sbt packagedArtifacts`

## problem
sbt version: 1.4.x (1.3.x versions don't have this problem)

All the files in the jar files created by sbt's package tasks (`package`, `packagedArtifacts`) have `1970-01-01 00:00:00` for their creation date and time. Please have a look at the following screenshot.
Screen Shot 2020-11-16 at 6 27 25 pm

This might be a big deal for some cases where the application is using a particular version of Spring Boot with a problem like the following one.
https://github.com/spring-projects/spring-boot/issues/19518
Although it was already fixed in Spring Boot, there might be some case where upgrading Spring Boot might not be an option

## expectation

It's expected to have the time the files were actually compiled and created.

## notes
I've already fixed it and test it locally with `publishLocal`. I'll create a PR for this issue soon. 
Done (#6093).

kevin-lee:
Fix #6092 - sbt 1.4.0~1.4.3 - all the files in the packaged jar files created at epoch millisecond 0 (1970-01-01 00:00:00)

Before
Screen Shot 2020-11-16 at 6 27 25 pm

After
Screen Shot 2020-11-16 at 6 30 14 pm

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1706,7 +1706,7 @@ object Defaults extends BuildCommon {
         config,
         s.cacheStoreFactory,
         s.log,
-        sys.env.get("SOURCE_DATE_EPOCH").map(_.toLong * 1000).orElse(Some(0L))
+        sys.env.get("SOURCE_DATE_EPOCH").map(_.toLong * 1000)
       )
       config.jar
     }


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1706,7 +1706,7 @@ object Defaults extends BuildCommon {
         config,
         s.cacheStoreFactory,
         s.log,
-        sys.env.get("SOURCE_DATE_EPOCH").map(_.toLong * 1000).orElse(Some(0L))
+        sys.env.get("SOURCE_DATE_EPOCH").map(_.toLong * 1000)
```
eed3si9n, line 1709: Zeroing out the JAR was done so intentionally in https://github.com/sbt/sbt/pull/5344 to make the packaging more reproducible.

kevin-lee: Ah... Alright I should need to read to understand it but I have a very quick question.
Should it be `0L` or any fixed epoch milliseconds? If it's latter one, can it be `Some(System.currentTimeMillis())`?

kevin-lee: OK I got it. I will use sbt 1.3.13 or use `SOURCE_DATE_EPOCH` env var if the timestamp of the class file is important. Thanks!

Issue #5679, opened 2020-07-11
system property for onChangedBuildSource


eed3si9n:
## steps

```scala
Global / onChangedBuildSource := ReloadOnSourceChanges
```

## problem

It's developer specific preference, and we can't put it in global either because of datatype.

## expectation

If there's a system property for this, we can use that as a default.

```
$ sbt -Dsbt.build.onchange=reload
$ sbt -Dsbt.build.onchange=warn
$ sbt -Dsbt.build.onchange=ignore
```

dwijnand:
It ain't pretty, but it works (credit to @xuwei-k, I lost the gist atm):

```scala
Def.settings {
  try {
    val clazz = Class.forName("sbt.nio.Keys$WatchBuildSourceOption")
    val value = Class.forName("sbt.nio.Keys$IgnoreSourceChanges$").getDeclaredField("MODULE$").get(null)
    val manifest = new scala.reflect.Manifest[AnyRef]{ def runtimeClass = clazz }
    Seq(
      SettingKey[AnyRef]("onChangedBuildSource")(manifest, sbt.util.NoJsonWriter()) in Global := value
    )
  } catch {
    case e: Throwable =>
      Nil
  }
}
```

eed3si9n:
This was fixed in https://github.com/sbt/sbt/pull/6099

PR #6099, opened 2020-11-16
Support a system property for onChangedBuildSource

xirc:
It resolves the issue https://github.com/sbt/sbt/issues/5679.

## Overview
A default value of `onChangedBuildSource` can be configured by the system property `sbt.build.onchange`.
As written in the above issue, the valid values of `sbt.build.onchange` are the following.
- `sbt.build.onchange=reload`
- `sbt.build.onchange=warn`
- `sbt.build.onchange=ignore`

The default value `warn` is used if `sbt.build.onchange` is not specified or `sbt.build.onchange` is an invalid value.

## How to test it

There is no test code of this feature, but it can be easily tested manually.
I have tested by my hand using the following command.

```
$ sbt -Dsbt.version=1.4.3-SNAPSHOT -Dsbt.build.onchange=warn "show onChangedBuildSource"
$ sbt -Dsbt.version=1.4.3-SNAPSHOT -Dsbt.build.onchange=reload "show onChangedBuildSource"
$ sbt -Dsbt.version=1.4.3-SNAPSHOT -Dsbt.build.onchange=ignore "show onChangedBuildSource"
$ sbt -Dsbt.version=1.4.3-SNAPSHOT -Dsbt.build.onchange=invalid-one "show onChangedBuildSource"
```

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -167,7 +167,17 @@ object Defaults extends BuildCommon {
       fileOutputExcludeFilter :== NothingFilter.toNio,
       inputFileStamper :== sbt.nio.FileStamper.Hash,
       outputFileStamper :== sbt.nio.FileStamper.LastModified,
-      onChangedBuildSource :== sbt.nio.Keys.WarnOnSourceChanges,
+      onChangedBuildSource :== {
+        val sysPropKey = "sbt.build.onchange"
+        sys.props.getOrElse(sysPropKey, "warn") match {
+          case "reload" => sbt.nio.Keys.ReloadOnSourceChanges
+          case "warn"   => sbt.nio.Keys.WarnOnSourceChanges
+          case "ignore" => sbt.nio.Keys.IgnoreSourceChanges
+          case unknown =>
+            System.err.println(s"Unknown $sysPropKey: $unknown.\nUsing warn.")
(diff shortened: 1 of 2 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -167,7 +167,17 @@ object Defaults extends BuildCommon {
...
+      onChangedBuildSource :== {
+        val sysPropKey = "sbt.build.onchange"
+        sys.props.getOrElse(sysPropKey, "warn") match {
+          case "reload" => sbt.nio.Keys.ReloadOnSourceChanges
+          case "warn"   => sbt.nio.Keys.WarnOnSourceChanges
+          case "ignore" => sbt.nio.Keys.IgnoreSourceChanges
+          case unknown =>
+            System.err.println(s"Unknown $sysPropKey: $unknown.\nUsing warn.")
+            sbt.nio.Keys.WarnOnSourceChanges
+        }
```
eed3si9n: Thanks for the contribution. Could you move this to https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/internal/SysProp.scala so we can centralize sys props related things there?

xirc: Thanks for the review. I'll change the code.

xirc: I've changed it.

eed3si9n (approved):
Thanks!

Issue #6101, opened 2020-11-17
sbt wrongly prints "sbt server is already booting"


sideeffffect:
## steps

Start sbt in an "unusual" filesystem, fox example, in Eclipse Che, which is a containerized, cloud IDE, where the filesystem sbt works with is mounted over network.

## problem

The whole message is this:

```
sbt server is already booting. Create a new server? y/n (default y)
```

sbt prints it even at wrong times.

## expectation

It seems enough to just press Enter or y and Enter, and sbt then continues, but it shouldn't be displayed in the first place, because no sbt server has started yet.

## notes

Other instance is this StackOverflow question, where there were issues with Windows' wsl2:
https://stackoverflow.com/questions/64809111/sbt-server-is-already-booting-error-when-launching-sbt-from-wsl2-ubuntu

It may be causing issues for other tools depending on sbt: https://github.com/scalameta/metals/issues/2235

tahakonuk:
I too agree that the error message is misleading. I searched for a zombie instance when I first got the error. Although sbt continues booting process after pressing Enter, some functionalities don't work. Such as file watching mechanism (commands starting with "~").

eed3si9n:
@sideeffffect Hi I don't have a context to what's going on in this issue. Could you please follow the prescribed format and https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md#what-to-report guide please?

eatkins:
There is a workaround. If you run sbt with `-Dsbt.server.forcestart=true` this behavior should be avoided (you can add this to .sbtopts in the project directory to always apply the property).

The problem is likely that sbt is unable to create a unix domain socket on this file system. I am not going to fix the underlying issue myself but I can provide guidance for a potential contributor. The offending code is here: https://github.com/sbt/sbt/blob/c52e9916e22e6b1340d83c74abd8e3a331c3229b/main-command/src/main/java/sbt/internal/BootServerSocket.java#L348. The trick is to figure out what kind of IOException is thrown on exotic file systems and how to handle it.

sideeffffect:
We could log the caught exception. `Info` level seems to high, but `Debug` should be appropriate. I can open a PR for this.

sideeffffect:
Btw, @eatkins is `-Dsbt.server.forcestart=true` something you would recommend tools like Metals (the Scala language server) to use for their purposes (like importing a project from sbt)?

eatkins:
[...]

PR #6104, opened 2020-11-18
Give and explanation in case of ServerAlreadyBootingException

sideeffffect:
This should help debugging the situation when there might be issues with creating sockets, like here https://github.com/sbt/sbt/issues/6101

### DIFF
--- a/main-command/src/main/java/sbt/internal/BootServerSocket.java
+++ b/main-command/src/main/java/sbt/internal/BootServerSocket.java
@@ -345,7 +345,7 @@ public class BootServerSocket implements AutoCloseable {
               : new UnixDomainServerSocket(name);
       return socket;
     } catch (final IOException e) {
-      throw new ServerAlreadyBootingException();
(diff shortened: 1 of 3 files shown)


eatkins:
Thank you very much for taking the time to prepare a PR. Unfortunately, I don't think I would accept this as is. The goal here should be to determine what specific IOException is causing the issue in #6101. What would be much more helpful would be for you to produce a stack trace showing what the actual cause of the issue is. Then we could fix the underlying issue so that users are not prompted at all about a running server if there isn't one.

It should be straightforward to test the change. You just need to run `sbt publishLocal` in a clone of the sbt project on the machine where you see the issue. Then you can use the locally published sbt by running `sbt -Dsbt.version=1.4.3-SNAPSHOT` in the project directory where you observe the issue. If you do not see that your changes were applied, you may need to run `reboot full` in the sbt prompt.

eed3si9n:
Closing this for now.

PR #6105, opened 2020-11-18
JLine 3.17.1

eatkins:
We no longer need to use the forked version of jline because they have
merged in our required changes. The latest version of jline does upgrade
jansi, however, and some of the apis we were relying on for windows were
removed so they had to be manually implemented. I verified that console
input still worked on my windows vm after this change.

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/JLine3.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/JLine3.scala
@@ -24,7 +24,6 @@ import org.jline.utils.OSUtils
 import scala.collection.JavaConverters._
 import scala.util.Try
 import java.util.concurrent.LinkedBlockingQueue
-import org.fusesource.jansi.internal.WindowsSupport
 
 private[sbt] object JLine3 {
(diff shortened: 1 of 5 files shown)


eatkins:
The latest version of jansi broke binary compatibility with jline 2 (https://github.com/fusesource/jansi/pull/174). This will require a little bit more work to restore the missing apis in our fork of jline2 (https://github.com/sbt/jline2).

eatkins:
I manually tested that the console works on windows with scala 2.10.7, 2.11.12, 2.12.12, 2.13.3 and dotty 0.26.0-RC1 after these changes. And I tested with the normal sbt launch jar and with a locally published jar incorporating https://github.com/sbt/launcher/pull/81.

eed3si9n:
Given the high frequency of JLine and unpredicted result, I feel like this should go into sbt 1.5.0. WDYT?

eatkins:
That seems fine. There is no rush for this. I just wanted to get us off the forked jline 3 and there were unexpected complications due to their breaking binary compatibility in jansi.

eed3si9n (approved):

Issue #6100, opened 2020-11-16
regression: letters that worked in 1.3.13 does not work in 1.4.x for example ÅÄÖ åäö, maybe jline 3 problem with Unicode or Utf-8 encoding?


bjornregnell:
## steps

Running sbt 1.4.x (x in any of {0,1,2,3}) on Ubuntu 18.04  (@SethTisue reported same error on MacOS)

```
$ sbt
[info] welcome to sbt 1.4.3 (Ubuntu Java 11.0.9.1)
sbt:bonus> Ã¥Ã¥Ã¤Ã¶  // does not echo åäö although these keys were pressed
sbt:bonus> console
[info] Starting scala interpreter...
Welcome to Scala 2.13.3 (OpenJDK 64-Bit Server VM, Java 11.0.9.1).
Type in expressions for evaluation. Or try :help.

scala> val Ã¥Ã¤Ã¶ = "ÃÃÃ"  // wrong echoed char also in console when typing `val åäö = "ÅÄÖ"`
            ^
       error: illegal character '\u00a5'
              ^
       error: illegal character '\u00a4'
                ^
       error: illegal character '\u00b6'
```

## problem

Special letters, e.g. the swedish letters ÅÄÖ åäö does not work anymore neither in sbt shell or in REPL started by console, althoug the worked fine in 1.3.13 and the work in vanilla scala 2.13.3 REPL.

## expectation
Should work as it used to in 1.3.13:
```
[info] welcome to sbt 1.3.13 (Ubuntu Java 11.0.9.1)
sbt:bonus> åäö  // does echo as it should
sbt:bonus> console
[info] Starting scala interpreter...
Welcome to Scala 2.13.3 (OpenJDK 64-Bit Server VM, Java 11.0.9.1).
Type in expressions for evaluation. Or try :help.

scala> val åäö = "ÅÄÖ"
val åäö: String = ÅÄÖ
```
And as in vanilla REPL:

```
 $ scala
Welcome to Scala 2.13.3 (OpenJDK 64-Bit Server VM, Java 11.0.9.1).
Type in expressions for evaluation. Or try :help.

scala> val åäö = "ÅÄÖ"    // works just fine
val åäö: String = ÅÄÖ
```

## notes

Maybe this has to do with jline

eed3si9n:
@bjornregnell Thanks for the report.

Could you try starting sbt with `sbt -v -Dfile.encoding=UTF-8` and see if the situation changes?

[...]

PR #6106, opened 2020-11-18
Fixes UTF-8 handling in shell and console

eatkins:
With sbt 1.4.x, non-ascii utf-8 characters are not handled correctly in
the console. It was not clear from the jline 3 documentation but the
NonBlockingReader.read method is supposed to return unicode points
rather than utf8 bytes. To fix this, we can decode the input and return
the code point rather than the directy byte from the input stream.


eatkins:
Fixes #6100.

eed3si9n (approved):
Thanks!

PR #6114, opened 2020-11-18
Use NetworkClient to implement `sbt -bsp`

eatkins:
Network client already supports the -bsp command (since
65ab7c94d038955d48eefbf7951fb9cd399ab7ef). This commit reworks the
BspClient.run method so that it delegates to the NetworkClient. The
advantage to doing it this way is that improvements to starting up the
sbt server by the thin client will automatically propagate to the -bsp
command. The way that it is implemented, all of the output generated
during server startup will be redirected to System.err which is useful
for debugging without messing up the bsp protocol, which relies on only
bsp messages being written to System.out.

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/BspClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/BspClient.scala
@@ -7,20 +7,12 @@
 
 package sbt.internal.client
 
-import java.io.{ File, InputStream, OutputStream }
+import java.io.{ InputStream, OutputStream }
 import java.net.Socket
 import java.util.concurrent.atomic.AtomicBoolean
 
 import sbt.Exit
-import sbt.io.syntax._
-import sbt.protocol.ClientSocket
-
 import scala.util.control.NonFatal
-import java.lang.ProcessBuilder.Redirect
-
-class BspClient private (sbtServer: Socket) {
-  private def run(): Exit = Exit(BspClient.bspRun(sbtServer))
-}
 
 object BspClient {
   private[sbt] def bspRun(sbtServer: Socket): Int = {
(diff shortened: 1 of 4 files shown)


adpi2:
This is great, thank you!

Unfortunately it does not seem to work on my machine:
1. In a `1.4.3-SNAPSHOT` project
2. Run `sbt -bsp`

The server starts, it loads the project but then it automatically shuts down.

After some investigation I see that the `sbt-console0-ui-trhead` of the server sends an `exit` command to the fast track.

eatkins:
Hmm. It works fine on my computer. It shouldn't make a difference, but what happens when you run `java -jar ~/.ivy2/local/org.scala-sbt/sbt-launch/1.4.3-SNAPSHOT/jars/sbt-launch.jar -bsp`?

adpi2:
Same result
```
> java -jar ~/.ivy2/local/org.scala-sbt/sbt-launch/1.4.3-SNAPSHOT/jars/sbt-launch.jar -bsp
Content-Length: 58
Content-Type: application/vscode-jsonrpc; charset=utf-8

{"jsonrpc":"2.0","method":"shutdown","params":[true,null]}
```

eatkins:
Can you try again with the latest code? Also the version will now be 1.5.0-SNAPSHOT, not 1.4.3-SNAPHOT.

main-command/src/main/scala/sbt/internal/client/BspClient.scala
```diff
@@ -7,16 +7,12 @@
...
 import java.util.concurrent.atomic.AtomicBoolean
 
 import sbt.Exit
-import sbt.io.syntax._
-import sbt.protocol.ClientSocket
-
 import scala.util.control.NonFatal
-import java.lang.ProcessBuilder.Redirect
 
 class BspClient private (sbtServer: Socket) {
```
adpi2: `class BspClient` is not used anymore.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -1214,16 +1217,18 @@ object NetworkClient {
...
-        else 1
-      } catch { case _: Throwable => 1 } finally client.close()
-    } catch {
-      case NonFatal(e) =>
-        e.printStackTrace()
-        1
-    }
+    run(configuration, arguments, false)
+  def run(
+      configuration: xsbti.AppConfiguration,
```
adpi2, line 1223: `configuration` is not used

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -1214,16 +1217,18 @@ object NetworkClient {
   }
 
   def run(configuration: xsbti.AppConfiguration, arguments: List[String]): Int =
```
adpi2, line 1220: `configuration` is not used

adpi2 (dismissed):
It is much better! Also I tested the integration with Metals and it works.

main-command/src/main/scala/sbt/internal/client/BspClient.scala
eatkins: Good catch.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eatkins: Good catch. It is supposed to be used: https://github.com/sbt/sbt/compare/dd2ed16389e1a2c4d082dc6d627b86c8553a432b..f96eeb9b09a9a94b1b5c079fc52497053dd592ae

eatkins:
As a side note, `sbt --client -bsp` works as well using sbtn. The one catch is that we need a new sbtn that redirects the server startup output to System.err rather than System.out, which is a part of this PR. Besides the lower startup time, `sbtn -bsp` uses about 15MB of RSS memory compared to 350MB for `sbt -bsp`.

eatkins:
@adpi2 I simplified the approach for ensuring that the server doesn't unexpectedly exit. Can you verify that it works on your machine? Also, I need a re-approval due to force pushing.

adpi2 (approved):

adpi2:
Apparently, my approving review is not enough because I don't have the write access on this repository.

eed3si9n:
@adpi2 Welcome to the team!

PR #6116, opened 2020-11-18
Fix supershell blowing away lines on windows

eatkins:
In 85d17889b6ff39c4d93f516fc247b91be73514f0, I attempted to fix
supershell messages getting interlaced with log lines. It turned out
that that approach didn't work with windows and was causing all of the
output to bet blown away. A better approach is to check if the bytes
we're writing contain one or more line separators. If so, we can wrap
the bytes in a string and split the string into lines. Then we can
append a ClearScreenAfterCursor before every newline.

I think the problem with windows was that the ClearScreenAfterCursor was
coming between the carraige return and the newline.


eatkins:
I forgot to actually test #6112 on windows and it caused problems. This time I could not reproduce interlacing on windows or mac.

eed3si9n (approved):

PR #6130, opened 2020-11-20
Stop CI hangs in background job service shutdown

eatkins:
A periodic stacktrace showed that scripted tests were still hanging in ci
trying to shutdown the background job service (I had previously thought
that I'd fixed that in 16bef0cfc8e9018a2a7373f307d9e00d7bdfb905). It
appears that there is a logical bug that prevents some jobs from being
removed from the jobSet even though they have finished. If that happens,
the shutdown will never exit. That is highly undesirable and can be
avoided by adding a timeout and also only trying to shutdown the job if
it is actually running.


eatkins:
Here is a CI build where you can see it got stuck trying to shutdown the background job: https://github.com/eatkins/sbt/runs/1428904993?check_suite_focus=true (you have to scroll down in the output for a while until you start seeing stacktraces showing the main thread blocked in DefaultBackgroundJobService.shutdown).

eed3si9n (approved):

PR #6143, opened 2020-11-20
[1.4.x] backport #6114, #6125, #6128, #6129, #6130

eatkins:
Instead of opening individual backport commits for #6114, #6125, #6128, #6129, #6130, this commit merges all of them.


eed3si9n:
When I backport, I normally chery-pick individual commits inside a pull request like https://github.com/sbt/sbt/pull/6114/commits/f3b3148c5898d13593cc0531fd618346ad38731d.
Is using merge-commit an alternative technique?

eatkins:
I don’t think it makes a big difference but it’s less labor to do `git cherry-pick -m1 $MERGE_COMMIT` if you want to backport all of the commits from a pr. It’s also easier to find which commit to cherry-pick because you can search the git log for the original pr number rather than a commit hash. Some of these were single commit prs so just cherry-picking the individual commit would have worked but I did this mechanically by pr number.

eed3si9n:
Cool. That's a TIL for me! I think it's a good tradeoff since manual cherry picking is labor intensive and error prone. I don't care so much about the history on 1.4.x.. I guess the only thing I'd miss would be the calculation of contributor.

eed3si9n (approved):

PR #6145, opened 2020-11-20
Print empty newlines with progress

eatkins:
With the latest sbt code, the `last` command displayed all of the output
without line separators. This occurred because the logic for appending
bytes to System.out split the input bytes on the line separator but if
there was nothing but the line separator in the input bytes then the
result was empty and nothing was printed.


eatkins:
I actually wondered whether this bug might crop up because `println(s: String)` is typically implemented something like `out.print(s); out.print(System.lineSeparator)`. I'm not 100% sure why I didn't see it in normal output and not just the `last` command but at least I caught it before we made a release.

eed3si9n (approved):

PR #6146, opened 2020-11-20
Fixes scala-compiler not included into the metabuild

### LINKED ISSUE #4452: Metabuild can't use scala compiler
## steps

I created a simple repo that reproduces the issue: https://github.com/eatkins/failed-plugin

## problem

If I build a local sbt version off of the head of the origin/develop branch, I am unable to access the scala.tools.nsc package:
```
[info] Loading project definition from /Users/ethanatkins/work/scratch/failed-plugin/project

[info] Compiling 1 Scala source to /Users/ethanatkins/work/scratch/failed-plugin/project/target/scala-2.12/sbt-1.0/classes ...
[error] /Users/ethanatkins/work/scratch/failed-plugin/project/com/swoval/Plugins.scala:3:14: object tools is not a member of package scala
[error] import scala.tools.nsc
[error]              ^
[error] /Users/ethanatkins/work/scratch/failed-plugin/project/com/swoval/Plugins.scala:6:22: not found: value nsc
[error]   val settings = new nsc.Settings()
[error] two errors found
```

## expectation

`scala.tools.nsc` is available.

eatkins:
When using the launcher's classpath for the metabuild, the
scala-compiler jar can be missing. This is because the managedJars
method only returns the scala-library jar and not the rest of the scala
instance. To fix this, we can always prepend the scala instance jars to
the classpath.

In order to simulate the issue in scripted, I had to manually remove the
scala-compiler.jar from the scripted classpath or else the scripted test
that I added doesn't actually do anything because the scala-compiler.jar
would end up on the app.provider.mainClasspath.

Fixes #4452

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1362,7 +1362,9 @@ def scriptedTask(launch: Boolean): Def.Initialize[InputTask[Unit]] = Def.inputTa
     scriptedLaunchOpts.value ++ (if (launch) Some(launchJar) else None),
     scalaVersion.value,
     version.value,
-    (scriptedSbtReduxProj / Test / fullClasspathAsJars).value.map(_.data),
+    (scriptedSbtReduxProj / Test / fullClasspathAsJars).value
+      .map(_.data)
+      .filterNot(_.getName.contains("scala-compiler")),
     streams.value.log
   )
 }
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2506,14 +2506,15 @@ object Classpaths {
         val isMeta = isMetaBuild.value
         val force = reresolveSbtArtifacts.value
         val app = appConfiguration.value
-        val sbtCp0 = app.provider.mainClasspath.toList
-        val sbtCp = sbtCp0 map { Attributed.blank(_) }
+        def isJansiOrJLine(f: File) = f.getName.contains("jline") || f.getName.contains("jansi")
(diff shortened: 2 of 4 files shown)


eatkins:
As a side note, it does seem a little weird to me that there is an expectation that the scala compiler is on the plugin classpath. I sort of think it is more correct to have to manually depend on it in `project/build.sbt` but since there are a number of builds that expect its presence, I don't think it hurts too badly to have it.

eed3si9n (approved):
Thanks!

dwijnand:
> it does seem a little weird to me that there is an expectation that the scala compiler is on the plugin classpath

More so than any other transitive dependency, like gigahorse?  Btw, did you know https://github.com/cb372/sbt-explicit-dependencies existed?

eatkins:
Yes, exactly. I think it was a mistake that transitive dependencies were ever on the metabuild classpath. In an ideal world, it would have only ever had java, scala, sbt and xsbt classes available.

eatkins:
But with respect to the scala compiler, apparently neither coursier nor ivy (at least when resolving the sbt classpath in the launcher) consider the scala-compiler part of the transitive classpath (other transitive dependencies like gigahorse were) or this regression would not have existed. It isn't exactly clear to me what historically brought the scala-compiler onto the classpath but it seems like accidental behavior which was what I was getting at with my comment.

dwijnand:
With parts of sbt depending on the scala-compiler seems to me the accident is that it wasn't part of that transitive classpath.  I'd understand, somewhat, if it was there but pinned to `scalaVersion` rather than what the POMs say.

PR #6155, opened 2020-11-22
Use `Global / localCacheDirectory` for remote caching

### LINKED ISSUE #6102: Move pushRemoteCache / ivyPaths
Reported originally by @xuwei-k 

## steps

1. `pullRemoteCache`
2. `clean`
3. `pullRemoteCache`

## problem

Remote cache is not cached.

## expectation

Probably the expectation is the remote cache is cached across `clean` to some documented directory.
I'd suggest we use `$HOME/.cache/sbt/local-cache/` or something along the line.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/6102

https://github.com/sbt/sbt/pull/6026 changed the implementation of remote cache to NOT use dependency resolution (Coursier), and directly use Ivy resolver for efficiency. This was good, but when I made the change, I've changed the cache directory to be `crossTarget.value / "remote-cache"`. This was ok for local testing purpose, but not great for real usage since we don't want the cache to be wiped out either in the CI machines or on a local laptop.

This adds a new Global key called `localCacheDirectory`. Similar to Coursier cache, this is meant to be shared across all builds running on a machine. Also similar to Coursier cache this will try to follow the operating system specifc caching directory.

### localCacheDirectory location

- Environment variable: `SBT_LOCAL_CACHE`
- System property: `sbt.global.localcache`
- Windows: %LOCALAPPDATA%\sbt\v1
- macOS: $HOME/Library/Caches/sbt/v1
- Linux: $HOME/.cache/sbt/v1

### DIFF
--- a/internal/util-collection/src/main/scala/sbt/internal/util/Util.scala
+++ b/internal/util-collection/src/main/scala/sbt/internal/util/Util.scala
@@ -47,6 +47,9 @@ object Util {
 
   def ignoreResult[T](f: => T): Unit = macro Macro.ignore
 
+  lazy val isMac: Boolean =
+    System.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("mac")
+
   lazy val isWindows: Boolean =
     System.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("windows")
 
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -388,6 +388,7 @@ object Keys {
   val pushRemoteCacheTo = settingKey[Option[Resolver]]("The resolver to publish remote cache to.")
   val remoteCacheResolvers = settingKey[Seq[Resolver]]("Resolvers for remote cache.")
   val remoteCachePom = taskKey[File]("Generates a pom for publishing when publishing Maven-style.")
+  val localCacheDirectory = settingKey[File]("Directory to pull the remote cache to.")
   val usePipelining = settingKey[Boolean]("Use subproject pipelining for compilation.").withRank(BSetting)
   val exportPipelining = settingKey[Boolean]("Product early output so downstream subprojects can do pipelining.").withRank(BSetting)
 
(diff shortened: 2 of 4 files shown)


agemooij:
Hi @eed3si9n . As a regular user of sbt, one running quite a large number of CI builds, I have no idea what a "remote cache" is and how this new cache impacts my CI builds.

Could you please explain in normal user terms what would be cached there and whether I should persist this new cache in my CI builds?

crakjie:
@agemooij remote cache is an experimental functionallity that basically allow you to push and pull compilation informations.
That allow the leverage incremental compilation in case where that was difficult to make it works.
This functionality is provided by two sbt commands `pushRemoteCache` `pullRemoteCache`, more documentation will probably be available when the feature will be stabilized.

agemooij:
Ah, thanks. So absolutely no impact on my projects at all then 😉 

In my experience CI and incremental builds are not things that go together very often, so the reference to "CI machines" is a bit out of place in this context IMO.

eed3si9n:
> Could you please explain in normal user terms what would be cached there and whether I should persist this new cache in my CI builds?

Some references:
- Remote Caching - https://www.scala-sbt.org/1.x/docs/Remote-Caching.html
- cached compilation for sbt - https://eed3si9n.com/cached-compilation-for-sbt
- remote caching sbt builds with Bintray - https://eed3si9n.com/remote-caching-sbt-builds-with-bintray

> The idea is for a team of developers and/or a continuous integration (CI) system to share build outputs. If the build is repeatable, the output from one machine can be reused by another machine, which can make the build significantly faster.

`Global / localCacheDirectory` is a setting for the directory where the pulled JAR files will be placed.

> In my experience CI and incremental builds are not things that go together very often, so the reference to "CI machines" is a bit out of place in this context IMO.

CI can continue to do a "clean" build. For project with certain size, developers being able to resume from the CI in the morning can save significant time.

agemooij:
Ah, I see. Thanks for the explanation @eed3si9n.

PR #5981, opened 2020-10-17
Work around "a pure expression does nothing" warning

eed3si9n:
Ref https://github.com/scala/bug/issues/12112

## steps

`Compile / scalacOptions += "-Xfatal-warnings"` in `project/plugins.sbt`.

```scala
package pkgtest

import sbt._, Keys._

// https://github.com/scala/bug/issues/12112
object PureExpressionPlugin extends AutoPlugin {
  lazy val testPureExpression = taskKey[Unit]("")
  override def projectSettings: Seq[Setting[_]] = {
    testPureExpression := {
      updateFull.value
      (Compile / compile).value
      (Test / test).value
    }
  }
}
```

## problem 

```scala
[error] /private/var/folders/hg/2602nfrs2958vnshglyl3srw0000gn/T/sbt_d9dc839f/project/PureExpressionPlugin.scala:10:18: a pure expression does nothing in statement position; multiline expressions might require enclosing parentheses
[error]       updateFull.value
[error]                  ^
[error] /private/var/folders/hg/2602nfrs2958vnshglyl3srw0000gn/T/sbt_d9dc839f/project/PureExpressionPlugin.scala:11:27: a pure expression does nothing in statement position; multiline expressions might require enclosing parentheses
[error]       (Compile / compile).value
[error]                           ^
[error] two errors found
[error] Compilation failed
```

## notes

Current app transformation macro creates a tree that looks like:

```scala
FullInstance.app[[T0[x]](T0[Int], T0[Int]), Unit](scala.Tuple2(task2, task1), (($p$macro$3: (Int, Int)) => {
  <synthetic> val $q$macro$2: Int = $p$macro$3._1;
  <synthetic> val $q$macro$1: Int = $p$macro$3._2;
  {
    ($q$macro$1: Int);
    ($q$macro$2: Int);
    ()
  }
}))(AList.tuple2[Int, Int])
```

Starting Scala 2.12.12 the compiler's "pure expression does nothing" has become more enthusiastic/accurate in its reach, and it started warning about the naked reference to `$q$macro$1` that appears to do nothing, even though in reality it would trigger the tasks and do something in the context of sbt. It's just _that_ particular line ends up macroed away into a pure expression.

A somewhat bizarre workaround is to make a fake call to a method just to satisfy this warning. I've chosen `scala.Predef.identity` here so it can be composed together with the existing expression nesting when they exist.

PR #6158, opened 2020-11-22
Work around "a pure expression does nothing" warning, take 2

### LINKED ISSUE #6126: 1.4.3 Regression with tasks that return a function
## steps

~~~ scala
// build.sbt
scalaVersion := "2.13.4"

val foo = taskKey[Unit]("")

foo := {
  val f = fTask.value
  streams.value.log.info(s"Invoking $f")
  f(0)
}

lazy val fTask: Def.Initialize[Task[Int => Int]] = Def.task {
  val log = streams.value.log
  x => {
    log.info(s"Evaluating f($x)")
    42
  }
}
~~~

## problem

sbt 1.4.3 does not evaluate the body of the function returned by `fTask`.

~~~
[info] Invoking $af8b69c2d2334066fdb0$$$Lambda$4487/0x0000000841335840@62df7a9a
[success] Total time: 0 s, completed Nov 19, 2020, 8:38:09 PM
~~~

## expectation

sbt 1.3.13 evaluates the body of the function.

~~~
[info] Invoking $c9429efa43a66453fa09$$$Lambda$4865/0x000000084167b840@50dbb213
[info] Evaluating f(0)
[success] Total time: 0 s, completed Nov 19, 2020, 8:39:16 PM
~~~

Note that it prints “Evaluating f(0)”, which was not the case with sbt 1.4.3.

## steps (variant 1)

I’ve noticed that if the function returned by my task is not a constant, then the output is different:

~~~ scala
// build.sbt
scalaVersion := "2.13.4"

val foo = taskKey[Unit]("")

foo := {
  val f = fTask.value
  streams.value.log.info(s"Invoking $f")
  f(0)
}

lazy val fTask: Def.Initialize[Task[Int => Int]] = Def.task {
  val log = streams.value.log
  x => {
    log.info(s"Evaluating f($x)")
    x + 1 // We now return the expression x + 1 instead of the constant 42
  }
}
~~~

Then both sbt 1.3 and 1.4 return the same (expected) output:

~~~
[info] Invoking $b1ce599c8650c7faaf35$$$Lambda$5101/0x0000000841566040@3d42583f
[info] Evaluating f(0)
[success] Total time: 0 s, completed Nov 19, 2020, 8:41:41 PM
~~~

## steps (variant 2)

Here is another variant where we print the result of the function returned by the task:

~~~ scala
// build.sbt
scalaVersion := "2.13.4"

val foo = taskKey[Unit]("")

foo := {
  val f = fTask.value
  streams.value.log.info(s"Invoking $f")
  streams.value.log.info(f(0).toString) // We now print the result
}

lazy val fTask: Def.Initialize[Task[Int => Int]] = Def.task {
  val log = streams.value.log
  x => {
    log.info(s"Evaluating f($x)")
    x + 1
  }
}
~~~

## problem

sbt 1.4.3 evaluates the function 4 times.

~~~
[info] Invoking $b1ce599c8650c7faaf35$$$Lambda$5130/0x0000000841569840@3bbae7a4
[info] Evaluating f(0)
[info] 1
[info] Evaluating f(0)
[info] Evaluating f(0)
[info] Evaluating f(0)
[success] Total time: 0 s, completed Nov 19, 2020, 8:45:37 PM
~~~

## expectation

sbt 1.3.13 evaluates the function once.

~~~
[info] Invoking $985fb3f3cf4f59b15e5b$$$Lambda$3893/0x0000000841367840@67d15a71
[info] Evaluating f(0)
[info] 1
[success] Total time: 0 s, completed Nov 19, 2020, 8:44:22 PM
~~~

## questions

- is there a better way to achieve what I want? I really want to define a task that returns me a function that I can apply on various inputs.
- is there a way to achieve what `taskDyn` was supposed to do, but without using operations that use macros? I’ve tried using `flatMap` but ended up with “Illegal dynamic dependencies” issues everywhere.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/6126

In #5981 I tried to work around the spruious post-macro "a pure expression does nothing" warning (scala/bug#12112) by trying to remove some pure-looking expressions out of the tree.

This quickly backfired when it was reported that sbt 1.4.3 was not evaluating some code. This backs out the macro-level manipulation, and instead try to silence the warning at the reporter level. This feels safer, and it seems to work just as well.

### DIFF
--- a/core-macros/src/main/scala/sbt/internal/util/appmacro/ContextUtil.scala
+++ b/core-macros/src/main/scala/sbt/internal/util/appmacro/ContextUtil.scala
@@ -312,16 +312,6 @@ final class ContextUtil[C <: blackbox.Context](val ctx: C) {
               case Converted.Failure(p, m)       => ctx.abort(p, m)
               case _: Converted.NotApplicable[_] => super.transform(tree)
             }
-          // try to workaround https://github.com/scala/bug/issues/12112 by removing raw Ident(_) in blocks
-          case Block(stats0, expr0) =>
-            val stats = stats0 flatMap { stat0 =>
-              val stat = super.transform(stat0)
-              stat match {
-                case Typed(ident @ Ident(_), _) if ident.symbol.isSynthetic => None
-                case _                                                      => Some(stat)
-              }
-            }
-            Block(stats, super.transform(expr0))
           case _ => super.transform(tree)
         }
     }
--- a/core-macros/src/main/scala/sbt/internal/util/appmacro/ContextUtil.scala
+++ b/core-macros/src/main/scala/sbt/internal/util/appmacro/ContextUtil.scala
@@ -321,8 +321,7 @@ final class ContextUtil[C <: blackbox.Context](val ctx: C) {
                 case _                                                      => Some(stat)
               }
             }
-            val expr = super.transform(expr0)
-            Block(stats, expr).setType(expr.tpe)
+            Block(stats, super.transform(expr0))
           case _ => super.transform(tree)
         }
     }
--- a/core-macros/src/main/scala/sbt/internal/util/appmacro/Instance.scala
+++ b/core-macros/src/main/scala/sbt/internal/util/appmacro/Instance.scala
@@ -187,7 +187,9 @@ object Instance {
       qual.foreach(checkQual)
       val vd = util.freshValDef(tpe, qual.pos, functionSym)
       inputs ::= new Input(tpe, qual, vd)
-      util.refVal(selection, vd)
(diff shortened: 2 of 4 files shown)


eatkins:
Nice workaround! Hopefully they don't change the warning text in future scala versions.

eed3si9n:
It's back. Looks like the scripted test was not enough - https://github.com/sbt/sbt/issues/6161

> Hopefully they don't change the warning text in future scala versions.

I'm not super concerned about this since we're in control of the Scala version used for the metabuild normally and this filtering will be applied only for the metabuild.

eatkins:
good point

Issue #6162, opened 2020-11-23
sbt doesn't start on arm64-macos


catap:
## steps

1. Download and install any arm64 JDK. I've used Azul JDK build 16-ea+24
2. `java -jar /opt/local/share/sbt/sbt-launch.jar`

## problem

```
/Users/catap/Library/Caches/JNA/temp/jna8864682821277708868.tmp: cannot open `/Users/catap/Library/Caches/JNA/temp/jna8864682821277708868.tmp' (No such file or directory)
catap@Kirills-mini-m1 scala-native % ~/zulu16.0.65-ea-jdk16.0.0-ea.24-macos_aarch64/bin/java -jar /opt/local/share/sbt/sbt-launch.jar
java.lang.UnsatisfiedLinkError: /Users/catap/Library/Caches/JNA/temp/jna12016497044725394171.tmp: dlopen(/Users/catap/Library/Caches/JNA/temp/jna12016497044725394171.tmp, 1): no suitable image found.  Did find:
	/Users/catap/Library/Caches/JNA/temp/jna12016497044725394171.tmp: no matching architecture in universal wrapper
	/Users/catap/Library/Caches/JNA/temp/jna12016497044725394171.tmp: no matching architecture in universal wrapper
	at java.base/jdk.internal.loader.NativeLibraries.load(Native Method)
	at java.base/jdk.internal.loader.NativeLibraries$NativeLibraryImpl.open(NativeLibraries.java:383)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:227)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:169)
	at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2383)
	at java.base/java.lang.Runtime.load0(Runtime.java:746)
	at java.base/java.lang.System.load(System.java:1857)
	at com.sun.jna.Native.loadNativeDispatchLibraryFromClasspath(Native.java:1018)
	at com.sun.jna.Native.loadNativeDispatchLibrary(Native.java:988)
	at com.sun.jna.Native.<clinit>(Native.java:195)
	at org.scalasbt.ipcsocket.UnixDomainSocketLibrary.<clinit>(UnixDomainSocketLibrary.java:129)
	at org.scalasbt.ipcsocket.JNAUnixDomainSocketLibraryProvider.socket(UnixDomainSocketLibrary.java:165)
[...]

Issue #6062, opened 2020-11-05
nightlies


eed3si9n:
Maybe it's best to start nightly process from scratch rather than trying to resurrect the on-again-off-again Jenkins approach.

GitHub packages, for example may be a good option - https://docs.github.com/en/free-pro-team@latest/packages/publishing-and-managing-packages/about-github-packages

SethTisue:
I hope a path forward can be found here. This would definitely reduce friction for people (like myself!) wanting to help test fixes that have landed recently.

It would also make it possible to test fixes in the Scala community build — either individual repos that might benefit from a fix, or build-wide. This could help avoid regressions in sbt point releases.

In short, although I know setting this up is work for you now, in the long run I suspect the effort would be repaid, since it would help all of us help you.

eed3si9n:
I earnestly tried to use GitHub Package Registry from GitHub Actions, and I was able to eventually semi-automatically publish sbt artifacts to https://github.com/sbt/sbt/packages/502680, but it is unsatisfying repository as it stands for many reasons:

1. The repository https://maven.pkg.github.com/sbt/sbt/ is behind the authentication wall. As far as I know, anyone with a GitHub account can get to it, but the users would need to go through the hassle of creating a personal GitHub token with the correct authorization.
2. The fact that the repo is behind the authentication wall is doubly annoying for sbt artifacts, because to pass the credentials to the sbt launcher requires a special property `-Dsbt.boot.credentials=$HOME/.sbt/github_credentials`.
3. Even after you convinced sbt launcher to resolve the artifacts using custom `repositories` file, Coursier doesn't seem to handle the credentials: `unauthorized: https://maven.pkg.github.com/sbt/sbt/org/scala-sbt/sbt/1.4.0-bin-20201115T022424/sbt-1.4.0-bin-20201115T022424.pom (GitHub Package Registry)`

So the incantation to get sbt shell to show up was:

```
$ sbt -Dsbt.repository.config=/tmp/hello/repositories -Dsbt.boot.credentials=$HOME/.sbt/github_credentials -Dsbt.coursier=false
```

In comparison to publishing to Jenkins box (https://jenkins.scala-sbt.org/sbt-snapshots/org/scala-sbt/sbt/) where the artifacts are publicly available and already built in to the sbt launcher as https://repo.scala-sbt.org/scalasbt/maven-snapshots/, this feels many steps inferior.

eed3si9n:
I guess the alternative might be good old Bintray.

[...]

PR #6165, opened 2020-11-23
Use JNI implementation of ipcsocket APIs on Arm Macs

eatkins:
There isn't yet a version of the jna available that works with the new
apple silicon using arm64. To workaround this, we can use the jni
implementation by default. If the user wants to force the jna, which
seems unlikely unless there is an unknown bug in the jni implementation,
they can opt in with the `sbt.ipcsocket.jna` system property and/or by
setting the serverUseJna setting.

### DIFF
--- a/main-command/src/main/java/sbt/internal/BootServerSocket.java
+++ b/main-command/src/main/java/sbt/internal/BootServerSocket.java
@@ -31,7 +31,6 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import net.openhft.hashing.LongHashFunction;
 import org.scalasbt.ipcsocket.UnixDomainServerSocket;
-import org.scalasbt.ipcsocket.UnixDomainSocket;
 import org.scalasbt.ipcsocket.Win32NamedPipeServerSocket;
 import org.scalasbt.ipcsocket.Win32NamedPipeSocket;
 import org.scalasbt.ipcsocket.Win32SecurityLevel;
@@ -337,16 +336,22 @@ public class BootServerSocket implements AutoCloseable {
   static ServerSocket newSocket(final String sock) throws ServerAlreadyBootingException {
     ServerSocket socket = null;
     String name = socketName(sock);
+    boolean jni = requiresJNI() || System.getProperty("sbt.ipcsocket.jni", "false").equals("true");
     try {
       if (!isWindows) Files.deleteIfExists(Paths.get(sock));
       socket =
           isWindows
-              ? new Win32NamedPipeServerSocket(name, false, Win32SecurityLevel.OWNER_DACL)
-              : new UnixDomainServerSocket(name);
+              ? new Win32NamedPipeServerSocket(name, jni, Win32SecurityLevel.OWNER_DACL)
+              : new UnixDomainServerSocket(name, jni);
       return socket;
     } catch (final IOException e) {
       throw new ServerAlreadyBootingException();
     }
+  }
+
+  public static Boolean requiresJNI() {
+    final boolean isMac = System.getProperty("os.name").toLowerCase().startsWith("mac");
+    return isMac && !System.getProperty("os.arch", "").equals("x86_64");
   }
 
   private static String socketName(String sock) {
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -383,6 +383,7 @@ object Defaults extends BuildCommon {
       },
       serverHandlers :== Nil,
       windowsServerSecurityLevel := Win32SecurityLevel.OWNER_DACL, // allows any owner logon session to access the server
+      serverUseJni := BootServerSocket.requiresJNI || SysProp.serverUseJni,
       fullServerHandlers := Nil,
       insideCI :== sys.env.contains("BUILD_NUMBER") ||
         sys.env.contains("CI") || SysProp.ci,
--- a/main-command/src/main/scala/sbt/BasicKeys.scala
+++ b/main-command/src/main/scala/sbt/BasicKeys.scala
@@ -83,6 +83,12 @@ object BasicKeys {
       "Configures the security level of the named pipe. Values: 0 - No security; 1 - Logon user only; 2 - Process owner only",
       10000
     )
+  val serverUseJni =
+    AttributeKey[Boolean](
+      "serverUseJni",
+      "Toggles whether to use the jna or jni implementation in ipcsocket.",
+      10000
+    )
 
   val serverIdleTimeout =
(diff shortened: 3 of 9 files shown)


eatkins:
See https://github.com/sbt/ipcsocket/pull/11. I think that it may not be necessary to bump ipcsocket to make sbt work on apple silicon because there is a translation layer in osx: https://www.theverge.com/21304182/apple-arm-mac-rosetta-2-emulation-app-converter-explainer. Nevertheless, it would be nice to run code written for the apple silicon.

catap:
@eatkins maybe switch JNA off for not supported platform by default?

eatkins:
> @eatkins maybe switch JNA off for not supported platform by default?

I did the opposite and only turn JNI on if it is known to be supported. That's easier to implement because we know there is a small set of supported JNI platforms.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -383,6 +383,7 @@ object Defaults extends BuildCommon {
       },
       serverHandlers :== Nil,
       windowsServerSecurityLevel := Win32SecurityLevel.OWNER_DACL, // allows any owner logon session to access the server
+      serverUseJna := !BootServerSocket.supportsJNI || SysProp.serverUseJna,
```
eed3si9n: Given the maintenance burden of JNI, I think we should use JNA by default and battle test it.
JNI requires building C code on all platforms we plan to support, whereas JNA binding is available for dozens of OS + arch - https://github.com/java-native-access/jna/tree/master/lib/native.

eatkins: I disagree. Jna is more complicated internally and we can’t do anything about it when it breaks like it did here with the new apple silicon. We are only using JNI on the same platforms where we have to use JNI for the thin client anyway. It isn’t clear to me that JNA will ever work with the thin client so we already have a permanent maintenance burden on using the JNI with ipcsocket. By making it the default on the server, we also get more battle testing of the jni implementation so if there is any discrepancy between the jni and jna implementation that we can fix it. I honestly think the point is somewhat academic about battle testing though since we haven’t seen any issues with either implementation apart from #6101 and #6162 which both were using jna.

eatkins: That all being said, is the suggestion that we only default to jni on arm64 macs?

eatkins: One last clarification, the reason I wanted to make jni default where supported is that I don’t think #6062 would have happened if jni had been the default since the apple silicon can supposedly run intel binaries. I suppose this is unusual because apple rarely changes platforms but it seems possible that microsoft could do something similar with a new windows release. At the end of the day, I guess it isn’t a big deal if we have to release a new sbt to support this kind of os upgrade but it would be nice if we didn’t have to.

eatkins: And just to be clear, regardless of how we set the default on supported jni platforms, we will always fallback to jna on unsupported platforms which is why it’s great that we have both jni and jna implementations.

eed3si9n: > That all being said, is the suggestion that we only default to jni on arm64 macs?

Yes.

My thinking here is that I'm bullish about thin client becoming the _only_ client eventually. In that future, we need to have the socket working in order to use sbt in Solaris etc, and those users can use pure-Java implementation of the thin client, which already exists thanks to you implementing it in JVM.

eatkins: I don't see how these goals are in conflict but I also don't want to continue this debate so I just went ahead and made the suggested change.

main-command/src/main/java/sbt/internal/BootServerSocket.java
```diff
@@ -337,18 +336,24 @@ public void close() {
   static ServerSocket newSocket(final String sock) throws ServerAlreadyBootingException {
     ServerSocket socket = null;
     String name = socketName(sock);
+    boolean jni = requiresJNI() || System.getProperty("sbt.ipcsocket.jni", "false").equals("true");
```
catap, line 339: I feel that this `requiresJNI` should be implemented inside `ipcsocket`.

`sbt` is client for `ipcsocket` and `ipcsocket` may include or may not include some library.

eatkins: Maybe, but I don't think it's necessary. The ipcsocket library is agnostic on the implementation and provides constructors that take a `useJNI` flag so downstream consumers can freely choose their implementation and copy this logic if they feel they need it.

catap: @eatkins right now it looks a bit strange. Library A doesn't support feature because it has limitation and all this limitation is implementing on client of this library.

From my point of view this check should be implemented inside, and API should have "force JNI" that end user can control.

eatkins: I don't know. You may be right but I'm not sure what the best approach is and having `requiresJNI` in sbt was easier to implement and has no downstream effect. All downstream consumers are likely using the jna implementation because it was a very recent change to support jni at all. You're welcome to open a pr in sbt ipcsocket but I'm not motivated to change this code or add it there.

eed3si9n (approved):

eatkins:
@catap since this was merged, there should be a new nightly published tonight. It would be great if you could try it and report back. The nightlies are found here: https://dl.bintray.com/sbt/maven-snapshots/org/scala-sbt/sbt/

catap:
@eatkins can I just run `jav -jar ...` on some jar from this repository?

eatkins:
You should be able to do `sbt -Dsbt.version=1.5.0-bin-20201123T081238` but with the new version that is published tonight instead of last night's version. It is pretty slow to resolve all the necessary artifacts for nightlies the first time so it may take a while for sbt to start up.

eatkins:
> @eatkins can I just run jav -jar ... on some jar from this repository?

I think I misread this. Building sbt requires sbt so I think waiting for the nightly is the only option. You could run
`java -Dsbt.version=1.5.0-bin-20201123T081238 -jar /opt/local/share/sbt/sbt-launch.jar` rather than `sbt -Dsbt.version...` if you don't have the sbt script on your path.

catap:
@eatkins thanks, I'll test it as soon as it builded a new version. Am I right that it is 1.5?

eatkins:
Yeah, it should be the last version in the list at this url: https://dl.bintray.com/sbt/maven-snapshots/org/scala-sbt/sbt/

catap:
@eatkins let me build it as local snapshot and test it

catap:
@eatkins this one is workig fine, but I enjoy the next one here: https://github.com/sbt/sbt/issues/6162#issuecomment-732510615

PR #6167, opened 2020-11-24
Use jni if required when running client from xMain

eatkins:
See https://github.com/sbt/sbt/issues/6162#issuecomment-733207878.


catap (approved):
❤️ ❤️ ❤️

eed3si9n (approved):

PR #6174, opened 2020-11-27
Make it possible to not write the bsp connection file

### LINKED ISSUE #6066: sbt 1.4.2 BSP disabled still generates .bsp/sbt.json which causes problems with IntelliJ IDEA
## steps
Bloop sbt-plugin installed
```
bloopExportJarClassifiers in Global := Some(Set("sources"))
bspEnabled := false
```
run `sbt bloopInstall`

## problem
`.bsp/sbt.json` is till generated which intellij tries to read and that causes some problem

## expectation
Do not generate any files in `.bsp`

hamnis:
Made the bspConfig task dependendant on the bspConfig value.
Changed the bspConfig setting to use a attributeKey so we can use it in the server as well.

I think this should fix #6066

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
@@ -78,10 +78,14 @@ object BuildServerProtocol {
   )
 
   lazy val globalSettings: Seq[Def.Setting[_]] = Seq(
-    bspConfig := BuildServerConnection.writeConnectionFile(
-      sbtVersion.value,
-      (ThisBuild / baseDirectory).value
-    ),
+    bspConfig := {
+      if (bspEnabled.value) {
+        BuildServerConnection.writeConnectionFile(
+          sbtVersion.value,
+          (ThisBuild / baseDirectory).value
+        )
+      } else ()
+    },
     bspEnabled := true,
     bspWorkspace := bspWorkspaceSetting.value,
     bspWorkspaceBuildTargets := Def.taskDyn {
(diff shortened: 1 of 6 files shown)


lightbend-cla-validator:
Hi @hamnis,

hamnis:
Not entirely sure how to write a test for this, if its at all possible.

Edit:

Figured it out. Its quite well explained in the DEVELOPING.md.
Thanks to @SethTisue for pointing this out.

eed3si9n:
@hamnis Thanks for the contribution! I'll let @adpi2 review the details, but overall I think the direction is good.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -78,10 +78,14 @@ object BuildServerProtocol {
...
-      sbtVersion.value,
-      (ThisBuild / baseDirectory).value
-    ),
+    bspConfig := {
+      if (bspEnabled.value) {
+        BuildServerConnection.writeConnectionFile(
+          sbtVersion.value,
+          (ThisBuild / baseDirectory).value
+        )
+      } else ()
```
adpi2: `bspConfig` is usually ran intentionally by the user or the client tool. So it is confusing that it does just nothing in the case `bspEnabled := false`. I would at least like to be informed why.

```suggestion
      } else {
         val logger = stream.value.log
         logger.warn("BSP is disabled for this build")
         logger.info("add 'Global / bspEnabled := true' to enable BSP")
      }
```

adpi2 (approved):

adpi2 (dismissed):
Thanks for this contribution! I am good with it :+1: 
Can you just consider the suggestion in `bspConfig` task?

Issue #6176, opened 2020-11-27
sbt launcher terminates BSP server when sbt is not on path


jastice:
When starting the sbt launcher from an environment that does not have the sbt script in the path, and no sbt server is running, the launcher script simply terminates after printing `[info] server was not detected. starting an instance`. This is the case when starting from IntelliJ IDEA.

## steps

1. start IntelliJ with Scala plugin from the UI (on Mac?)
2. import an sbt project configured to 1.4.x as BSP, choose "sbt" and not "sbt with Bloop" if prompted

alternatively if that doesn't repro:
1. start IntelliJ from a script, but set PATH to not include the sbt script
2. as above

## problem

1. the import will time out
2. in the build toolwindow, we see only
```[info] server was not detected. starting an instance```

In the background, what happens is that the sbt launcher simply terminates if it doesn't find the sbt launcher on the path.

## expectation

* the server should be able to be started from the same JVM used to start the launcher
* the sbt script should not be necessary at all, as sbt may not even have been installed on the system

eatkins:
> the server should be able to be started from the same JVM used to start the launcher

The problem with this expectation is that it makes the behavior of `sbt -bsp` very different from `sbt --client -bsp`. The latter is not started with a jvm at all so if is to be able to launch a server, it must be able to find either an sbt script or the launcher jar. Right now, the script can be provided to sbt as a command line argument and I just added #6177 to make the launcher jar available. Since bsp is an experimental feature, I feel like the best solution is for intellij to set that argument in the next version and say that 1.4.5 or greater is required for bsp support.

jastice:
Thanks! Actually IntelliJ simply uses the command line defined in `.bsp/sbt.json`, which gets generated by sbt itself. So this would have to also be fixed sbt-side. @adpi2 should know where :)

eed3si9n:
Here's what my `.bsp/sbt.json` contains:

> $ cat .bsp/sbt.json
> {"name":"sbt","version":"1.4.4","bspVersion":"2.0.0-M5","languages":["scala"],"argv":["/Users/eed3si9n/.sdkman/candidates/java/8.0.232.hs-adpt/jre/bin/java","-Xms100m","-Xmx100m","-classpath","/Users/eed3si9n/.sdkman/candidates/sbt/1.4.3/bin/sbt-launch.jar","xsbt.boot.Boot","-bsp"]}

This file is specific to my machine, so I'm not sure if the expectation 2

[...]

PR #6177, opened 2020-11-27
Add --sbt-launch-jar command line arg to client 

eatkins:
This will allows tools that bundle their own instance of the launch jar to run sbt server with  `--sbt-launch-jar=$BUNDLE_LUANCHER -bsp` so that the command works even when there is no sbt script on the tool's path.

See #6176


jastice:
How would this expected to be used? Would the `.bsp/sbt.json` be generated with this argument passed to the launch jar? Since it's already using the launch jar, why can't it use the classpath available in the launched process already?

adpi2 (dismissed):
As soon as this is merged I'll add the `--sbt-launch-jar` option to the BSP connection file

adpi2 (approved):

PR #6181, opened 2020-11-30
dotty-library is called scala3-library now

### LINKED ISSUE #6058: ClassLoaderCache for Scala 3
## problem

https://github.com/sbt/sbt/blob/25f492d9195dd7b1c8365e137278e8ca7e088951/main-command/src/main/scala/sbt/internal/classpath/ClassLoaderCache.scala#L182

https://repo1.maven.org/maven2/org/scala-lang/scala3-library_3.0.0-M1/

## expectation

```diff
-      case d :: s :: Nil if d.getName.startsWith("dotty-library") =>
+      case d :: s :: Nil if d.getName.startsWith("dotty-library") && d.getName.startsWith("scala3-library") =>
```

🤔

smarter:
Fixes https://github.com/sbt/sbt/issues/6058

I though this would help with
https://github.com/lampepfl/dotty/issues/10558 but `sbt repl` still
hides the input after typing a line and pressing enter even after this
change.

### DIFF
--- a/main-command/src/main/scala/sbt/internal/classpath/ClassLoaderCache.scala
+++ b/main-command/src/main/scala/sbt/internal/classpath/ClassLoaderCache.scala
@@ -179,7 +179,8 @@ private[sbt] class ClassLoaderCache(
   }
   override def apply(files: List[File]): ClassLoader = {
     files match {
-      case d :: s :: Nil if d.getName.startsWith("dotty-library") =>
+      case d :: s :: Nil
+          if d.getName.startsWith("dotty-library") || d.getName.startsWith("scala3-library") =>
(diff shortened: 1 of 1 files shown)


eatkins:
It looks like you need to run scala format. The fix looks correct to me but you probably should validate it before it's merged. It should be as simple as
1. sbt publishLocal
2. rm ~/.sbt/boot
3. cd $SCALA3-PROJECT-DIR
4. sbt -Dsbt.version=1.5.0-SNAPSHOT console

smarter:
Thanks, I gave it a try but the issue remains in the dotty repo (I've updated the commit message).

eed3si9n (approved):

Issue #2984, opened 2017-03-02
Can we force warming of compiler by compiling dummy file


eed3si9n:
Rory reported that when running on batch mode `sbt compile`, the first run on the machine is faster than the second onwards. His speculation was that compilation of compile bridge could have forced JITing on the first run.
Would it make sense to run a small dummy file (something like Hello world) before the `compile` on startup?

jvican:
JVM will compile concrete paths of the compiler. Compiling a dummy Scala file is not sufficient to get the compiler hot. In fact, it may  evenintroduce a performance regression since the JVM will optimize concrete paths, and while compiling a completely different project those paths will change and, supposing that the C2 has done some work, most of it will be thrown away because JIT will deoptimize paths. This would only work if you compile the same project at the beginning, and that wouldn't make a lot of sense.
  
I lean towards letting the user know that Scala compiler should always be kept hot. I don't think sbt can do anything to make compilation hot.

dwijnand:
@jvican One really hard-core min-max idea that we had was to derive a good minimal source file, representative of the code paths in the compiler the real project used. To more accurately warm up the compiler.

I would put this one in the "experiment here, because there might be good wins in some cases" box.

> I lean towards letting the user know that Scala compiler should always be kept hot.

That's being tracked as #2987. This derives from the same story from Rory, but it's distinct.

jvican:
> One really hard-core min-max idea that we had was to derive a good minimal source file, that was representative of the code paths in the compiler the real compiler used. To more accurately warm up the JVM.
  
In my opinion, it's impossible to get such a source file for any Scala codebase. I would really not like that sbt does this warming up the compiler for me. Also, it takes way more than one compiler iteration to properly warm up Scalac.
  
I think my issue is that a build tool should not take care of this. I find it a very bad idea, it's unpredictable and when I type `compile`, I really mean "just compile my project", nothing else.
  
No other JVM-based build tool does so. I suggest reconsidering the scope of this ticket, and even closing it. In any case, this is not something that can be done in a hackathon of 4 hours, since it requires deep knowledge of both scalac and JVM.

dwijnand:
My suggestion was an extreme extension to the original issue, which I mentioned because you were going into details about how a simple hello world wouldn't be sufficient. But let's discuss your feedback:

> Also, it takes way more than one compiler iteration to properly warm up Scalac.

There's no restriction on iterations. If more than one provides better overall compile speeds then do more than one.

[...]

PR #6185, opened 2020-11-30
Optionally provide sbt jline terminal for some tasks

eatkins:
The dotty repl task doesn't work with sbt 1.4.x when started with
in-process runMain. There were two problems:
1. the top classloader for run was not the jline 3 classloader so sbt
   was unable to override TerminalBuilder.build
2. even when setting the top loader to the jline 3 classloader, we
   weren't actually setting the global value so that it could be
   connsumed by TerminalBuilder.build

I didn't want to create a JLine3 terminal for every evaluation of
MainLoop.process so I added a setting, needsJLine3, which provides a
list of tasks for which we need to setup a jline 3 terminal before
evaluation.

In addition to these changes, the dotty build.sbt needs to add some
settings:

inProcessTopClassLoader := scalaInstanceTopLoader.value,
Global / needsJLine3 += "repl",
and in the repl task, it needs to add
terminal.value.setMode(canonical = false, echo = false)

### DIFF
--- a/main/src/main/scala/sbt/MainLoop.scala
+++ b/main/src/main/scala/sbt/MainLoop.scala
@@ -26,6 +26,8 @@ import scala.concurrent.duration._
 import scala.util.control.NonFatal
 import sbt.internal.FastTrackCommands
 import sbt.internal.SysProp
+import sbt.internal.util.DeprecatedJLine
+import sbt.internal.util.JLine3
 
 object MainLoop {
 
@@ -222,9 +224,18 @@ object MainLoop {
         val (restoreTerminal, termState) = channelName.flatMap(exchange.channelForName) match {
           case Some(c) =>
             val prevTerminal = ITerminal.set(c.terminal)
+            val needsJLine3 =
+              if (Project.isProjectLoaded(state)) Project.extract(state).get(Keys.needsJLine3)
+              else Nil
+            val closeJLine = if (needsJLine3.contains(exec.commandLine)) {
+              val jlineTerminal = JLine3(c.terminal)
+              DeprecatedJLine.setTerminalOverride(jlineTerminal)
+              () => jlineTerminal.close()
+            } else () => ()
             // temporarily set the prompt to running during task evaluation
             c.terminal.setPrompt(Prompt.Running)
             (() => {
+              closeJLine()
               ITerminal.set(prevTerminal)
               c.terminal.flush()
             }) -> progressState.put(Keys.terminalKey, Terminal(c.terminal))
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -335,6 +335,8 @@ object Defaults extends BuildCommon {
         if (!useScalaReplJLine.value) classOf[org.jline.terminal.Terminal].getClassLoader
         else appConfiguration.value.provider.scalaProvider.launcher.topLoader.getParent
       },
+      inProcessTopClassLoader := classOf[sbt.testing.Framework].getClassLoader,
+      needsJLine3 := Nil,
       useSuperShell := { if (insideCI.value) false else ITerminal.console.isSupershellEnabled },
       superShellThreshold :== SysProp.supershellThreshold,
       superShellMaxTasks :== SysProp.supershellMaxTasks,
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -595,6 +595,8 @@ object Keys {
 
   val useScalaReplJLine = settingKey[Boolean]("Toggles whether or not to use sbt's forked jline in the scala repl. Enabling this flag may break the thin client in the scala console.").withRank(KeyRanks.Invisible)
   val scalaInstanceTopLoader = settingKey[ClassLoader]("The top classloader for the scala instance").withRank(KeyRanks.Invisible)
+  val inProcessTopClassLoader = settingKey[ClassLoader]("The top classloader for in process tasks that need to run code in a separate classloader").withRank(KeyRanks.Invisible)
+  val needsJLine3 = settingKey[Seq[String]]("The names of tasks that require that we set a jline3 terminal")
 
   val stateStreams = AttributeKey[Streams]("stateStreams", "Streams manager, which provides streams for different contexts.  Setting this on State will override the default Streams implementation.")
   val resolvedScoped = Def.resolvedScoped
--- a/main/src/main/scala/sbt/internal/ClassLoaders.scala
+++ b/main/src/main/scala/sbt/internal/ClassLoaders.scala
@@ -29,7 +29,6 @@ private[sbt] object ClassLoaders {
   private implicit class SeqFileOps(val files: Seq[File]) extends AnyVal {
     def urls: Array[URL] = files.toArray.map(_.toURI.toURL)
   }
-  private[this] val interfaceLoader = classOf[sbt.testing.Framework].getClassLoader
   /*
    * Get the class loader for a test task. The configuration could be IntegrationTest or Test.
    */
@@ -48,6 +47,7 @@ private[sbt] object ClassLoaders {
     val allowZombies = allowZombieClassLoaders.value
     buildLayers(
       strategy = classLoaderLayeringStrategy.value,
+      topLoader = inProcessTopClassLoader.value,
       si = si,
       fullCP = fullCP,
       allDependenciesSet = dependencyJars(dependencyClasspath).value.filterNot(exclude).toSet,
@@ -98,6 +98,7 @@ private[sbt] object ClassLoaders {
             val transformedDependencies = allDeps.map(f => mappings.getOrElse(f.getName, f))
             buildLayers(
               strategy = classLoaderLayeringStrategy.value: @sbtUnchecked,
+              topLoader = inProcessTopClassLoader.value,
               si = instance,
               fullCP = classpath.map(f => f -> IO.getModifiedTimeOrZero(f)),
               allDependenciesSet = transformedDependencies.toSet,
@@ -133,6 +134,7 @@ private[sbt] object ClassLoaders {
    */
   private def buildLayers(
       strategy: ClassLoaderLayeringStrategy,
+      topLoader: ClassLoader,
       si: ScalaInstance,
       fullCP: Seq[(File, Long)],
       allDependenciesSet: Set[File],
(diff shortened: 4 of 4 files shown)


main/src/main/scala/sbt/MainLoop.scala
```diff
@@ -222,9 +224,18 @@ object MainLoop {
         val (restoreTerminal, termState) = channelName.flatMap(exchange.channelForName) match {
           case Some(c) =>
             val prevTerminal = ITerminal.set(c.terminal)
+            val needsJLine3 =
+              if (Project.isProjectLoaded(state)) Project.extract(state).get(Keys.needsJLine3)
+              else Nil
+            val closeJLine = if (needsJLine3.contains(exec.commandLine)) {
```
eed3si9n, line 230: Does this work with slash syntax or other composable commands like `~`, `;`, and `+`?

eatkins: No. I didn’t think of that.

smarter:
Thansk for working on this!

> In addition to these changes, the dotty build.sbt needs to add some

It would be great to have some documentation on what this does / why it's needed (for example, I'd like to know what else inProcessTopClassLoader is responsible for and what can break when I change it, I assume there's a reason the value you're suggesting is not the default).

By the way, this issue also affects github.com/scala/scala/, run `scala` in sbt, type `1+1` then Enter, and the input line will disappear, I assume the same fix will work too /cc @SethTisue

eed3si9n:
> It would be great to have some documentation on what this does / why it's needed (for example, I'd like to know what else inProcessTopClassLoader is responsible for and what can break when I change it, I assume there's a reason the value you're suggesting is not the default).

There are two-fold motivations around the complex classloading we do.

### Background

scala-library is used frequently, and it's pretty large. scala-compiler is also used frequently in different ways, and it's even bigger. The large size of scala-library or scala-compiler means that it takes time to load the `*.class` files into the JVM. Sometimes more importantly, it also means that the HotSpot Just-in-Time compiler would take a long time to C2 compile the scala-compiler bytecode into the optimized native code, at least 30s, maybe many minutes. The effect of JITing could mean 2x performance difference. The difference is so stark in 2017 we seriously [debated](https://github.com/sbt/sbt/issues/2984) if would be faster to first compile a "Hello World" file, then compile some subproject (like Akka Actor) than to compile it cold. Also the computers used to be slower ten years ago. 

### Warm library and warm compiler

The above is the basic motivation behind keeping scala-library warm for `run` and `test` tasks, and scala-compiler warm for `compile` task. And since the compiler is used to load the `build.sbt` scripts, usually we try to reuse the Scala version used by sbt.

Ethan pushed this even further in `ThisBuild / turbo := true` by creating layered ClassLoader for external library dependencies separate from the user-written app|test code. These are documented in https://www.scala-sbt.org/1.x/docs/In-Process-Classloaders.html.

Traditionally it was relatively safe to cache scala-compiler artifacts because it has kept minimum external dependencies. A notable exception to this is JLine/JAnsi used in the REPL, and also used by sbt shell.

### JLine/JAnsi globalness

The one issue related to JLine/JAnsi, I think, is that there can usually be one instance of JLine n because eventually things would lead to native library binding. This requires coordination on the sbt's part to know what the compiler expects.

### Virtual IO for server

Ethan also introduced the idea of virtual IO in sbt 1.4.x. To be able to run any programs, including `console` inside of the sbt server, but forward the result to a thin client like sbtn both the console input and the console output are rewired a virtual one in sbt 1.4.x. This resolved the missing piece in the server puzzle without having to reimplement all the tasks. It did make the situation with keyboard events and JLine 3 a bit more complex, but I think this was totally worth it as we can finally treat sbt server as a background daemon. I think this could explain why we need to take over the JLine 3 terminal building. Maybe we should try to communicate via Zinc API when constructing a REPL?

@eatkins knows more about these things, but I wanted to provide my view of these things. Hopefully it wasn't too far off from reality.

smarter:
> These are documented in scala-sbt.org/1.x/docs/In-Process-Classloaders.html.

Thanks, I wasn't aware of this page. `classLoaderLayeringStrategy` looks interesting but how does it interact with `inProcessTopClassLoader` in this PR?

eatkins:
@eed3si9n's description was very good. The `inProcessTopClassloader` allows a build to configure the classloader that is at the top of the layering hierarchy. Starting with 1.3.0, the top classloader was effectively a URLClassLoader with one jar, the sbt test interface jar, and whose parent was the java application classloader. The sbt metabuild classloader uses this same classloader as its parent. This allowed the test task to return its results to sbt without introducing any other conflicts with the sbt metabuild claspath. In the past, there could be problems because the test interface jar was provided in the same classloader as the rest of the metabuild. To make it possible for the in-process test task to return results to sbt, it was necessary to construct a dual classloader that loaded only certain classes from the metabuild classloader and the rest from a new classloader that was only used for test. This was hard to get right which was why in 1.3.0, things were reworked so that in-process run and test would share the same top classloader with the metabuild, but nothing else.

For 1.4.0, we needed to modify the classloader used by the console task so that we could override the jline terminal that was returned by `TerminalBuilder.build` to be one provided by sbt that would correctly forward io to a thin client connection. There is also only supposed to be one jline 3 system terminal in a given process and having multiple classloaders able to load jline 3 broke this invariant. Previously, this was handled by having a special scalaInstanceTopLoader setting that was used only by the console task. The console task would then construct its classloader layers in such a way that the layer of the metabuild that loads jline would be the top loader. This was the trick that allowed the console task to work with both the thin client and in and interactive server session. But it was only applied for console, not run or test. The inProcessTopClassLoader setting in this PR allows a build to override the top classloader used in run or test so that it isn't the normal classloader that is only able to load the sbt test interface jar. Previously this was just hardcoded in ClassLoaders.scala. An alternative approach to this would just be to always make the jline classloader the top classloader in run and test, but that would leak unnecessary classes into most builds.

The other aspect of this PR is that it's not enough to just put the jline classloader above the rest of the classpath in run. We also need to set a global value for the jline terminal before we evaluate run or test or the application will crash when running `TerminalBuilder.build()` because it will try to create a second system terminal. This is done explicitly in the console task implementation. To make this possible in other tasks, I added logic in MainLoop to set the appropriate global variable to avoid this. But we don't want to do this every time we run a command because creating a jline terminal always creates extra resources that we need to clean up so there is overhead associated with it. That is why it needs `Global / needsJLine3 += "repl"`. It is possible that the overhead of creating and closing a new jline terminal with each command evaluation is so fast that we wouldn't need this but I didn't benchmark it and decided to err on the side of only doing it if it were strictly necessary.

A better approach to this problem may be to have similar logic for setting the jilne terminal to what exists in console in the run implementation: https://github.com/sbt/sbt/blob/2bbbcfae77b53e89cea525e3f7b68da8ea142a96/main-actions/src/main/scala/sbt/Console.scala#L69-L74 There could be a boolean flag that the run implementation checks to determine whether it needs to manage a jline terminal.

This PR is more of a POC than anything. In fact, I'm going to change it to a draft.

smarter:
> The inProcessTopClassLoader setting in this PR allows a build to override the top classloader used in run or test so that it isn't the normal classloader that is only able to load the sbt test interface jar.

OK, but I'm having trouble seeing how this relates to the `repl` task in dotty: this tasks is run with `fork` set to true, so the JLine that is used here is classloaded by a separate JVM with its own classloader, so how does setting `inProcessTopClassLoader` affects whether the jline terminal started by that task works correctly or not?

eatkins:
That is correct but as I told you in the dotty thread, I strongly discourage using fork for tasks that take interactive user input. If you need to move repl into a separate project so you can use forked run for other tasks that don’t take input, I’d advise that. Basically I am telling you how you can make the repl task work with 1.4.x. If you don’t want to change the dotty build to work with this approach, that’s fine, but you’re probably on your own to fix the repl task.

eatkins:
The more I think about it, I don't even see the point of implementing repl yourself using `runMain`. I think you should just be able to make a subproject that depends on the root project and set the scala version to the version that is published by the root project. Then repl could just be a command or alias that calls `replProject / console`. Then you not only wouldn't have to reimplement it, you'd be having the same experience that downstream users will have when they run `console` in their build.

smarter:
> If you don’t want to change the dotty build to work with this approach, that’s fine, but you’re probably on your own to fix the repl task.

I'm confused then and appear to have misunderstood what this PR does: when I tried disabling forking in the dotty project on sbt 1.4.4, the repl terminal input worked correctly, so what does this PR provide on top of that?

> If you need to move repl into a separate project so you can use forked run for other tasks that don’t take input, I’d advise that.

This isn't just the dotty repl task, it's also the scala 2 repl launched with the `scala` task, which is not something that can be moved outside of the scala-compiler jar without breaking a ton of things.  But as I said in https://github.com/lampepfl/dotty/issues/10558#issuecomment-736718837, I don't think that even if we could disable forking just for the repl, the result would work well enough for us as is. Though maybe using classLoaderLayeringStrategy would help since we could set that to `Flat` to maybe avoid having the library from the reference compiler pollute the classpath? It's all incredibly confusing to me.

> If you don’t want to change the dotty build to work with this approach, that’s fine, but you’re probably on your own to fix the repl task.

I've invested countless hours of my life into the dotty build and kept pace with new sbt versions for more than six years now (including finding and fixing various bugs in sbt and zinc). It's not that I don't want to change things, it's that I can't figure it out even with my years of sbt expertise (and also I'm trying to release Scala 3.0, write a PhD thesis and not go insane at the same time).

> I think you should just be able to make a subproject that depends on the root project and set the scala version to the version that is published by the root project. 

Hmm maybe relying on the console task of sbt would be a good idea indeed, though it's not completely straightforward because of the bootstrapping process, the way scalaInstance works and the classloading required for macros, but I'll have to give it a try, thanks for the idea!

SethTisue:
> By the way, this issue also affects github.com/scala/scala/, run `scala` in sbt, type `1+1` then Enter, and the input line will disappear, I assume the same fix will work too /cc @SethTisue

But in scala/scala, when you run `scala` in sbt (it's just an alias for `repl-frontend/run`), it is running forked! (And I don't know the history of that choice.)

If I try and run unforked, with `set replFrontend/run/fork := false`, I get `Failed to initialize compiler: object scala.annotation.StaticAnnotation in compiler mirror not found.`... which I imagine I probably cooooooould figure out and work around, if I took the time?

So it would be nice if there was a fix that work with `fork := true`, since that's currently what we have in our build. But if we need to switch to `fork := false`, presumably that's doable with some work.

SethTisue:
>  I think you should just be able to make a subproject that depends on the root project and set the scala version to the version that is published by the root project. Then repl could just be a command or alias that calls replProject / console

Not sure I understand this suggestion. How can the subproject require that the Scala version be published locally?

If I naively follow your instructions,

```
sbt:root> set replFrontend/scalaVersion := "2.13.5-bin-SNAPSHOT"
sbt:root> repl-frontend/console
[info] Starting scala interpreter...
Welcome to Scala 2.13.4 (OpenJDK 64-Bit Server VM, Java 1.8.0_275).
```

but I'm not sure what a less "naive" version of this would look like.

eatkins:
I am taking an indefinite break from sbt so I’m sorry but I will not be responding to issue comments in the near term.

SethTisue:
> I'm not sure what a less "naive" version of this would look like

Dale was able to help me with this. We're fooling with it at https://github.com/scala/scala/pull/9380

PR #6199, opened 2020-12-11
Add compiler-interface to the Scala Instance top loader

adpi2:
The Scala 3 compiler now has a direct dependency on the `org.scala-sbt:compiler-interface` artifact. When compiling, it is crucial that its loads the  `compiler-interface` classes from the same loader than sbt. Otherwise we have `LinkageException`.

In this PR I moved the `compiler-interface` up in the class loader hierarchy. It is safe because it's a pure java API that has no dependency. Rather than introducing a new loader in the pile of class loaders, I reused the `TestInterfaceClassLoader` and I renamed it `SbtInterfaceClassLoader`.

The hierarchy of class loader now looks like this (the parents are at the bottom):
- `MetaBuildLoader`: all sbt jars
- `FullScalaLoader`: `scala-xml_2.12`, `scala-compiler-2.12.12` and `scala-reflect-2.12.12`
- `LibraryLoader`: `scala-library-2.12.12`
- `JLineLoader`: jline jars
- `SbtInterfaceLoader`: `test-interface` and `compiler-interface`
- topLoader: JDK except Jansi

Depending on `useScalaReplJLine` the `scalaInstanceTopLoader` is `JLineLoader` (which contains `SbtInterfaceLoader`) or `SbtInterfaceLoader` directly.

I tested it successfully against the `3.0.0-M2` compiler and a snapshot of the future `3.0.0-M3` compiler.

### DIFF
--- a/main/src/main/java/sbt/internal/MetaBuildLoader.java
+++ b/main/src/main/java/sbt/internal/MetaBuildLoader.java
@@ -65,11 +65,17 @@ public final class MetaBuildLoader extends URLClassLoader {
    */
   public static MetaBuildLoader makeLoader(final AppProvider appProvider) throws IOException {
     final String jlineJars = "jline-?[0-9.]+-sbt-.*|jline-terminal(-(jna|jansi))?-[0-9.]+";
+    final String testInterfaceJars = "test-interface-[0-9.]+";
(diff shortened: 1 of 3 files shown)


eatkins (approved):
LGTM

Issue #6080, opened 2020-11-11
Merge DottyPlugin in sbt


adpi2:
The `sbt-dotty` plugin sources can be found here: https://github.com/lampepfl/dotty/tree/master/sbt-dotty/src/dotty/tools/sbtplugin

It is composed of two independent parts:
- `DottyIDEPlugin`: install and configure vscode Dotty extension, start vscode
- `DottyPlugin`: manage Scala instance for Scala 3, configure `compile` and `doc`

This issue is scoped to merging the `DottyPlugin` logic into `sbt`. The `DottyIDEPlugin` will not be merged.

I give below a detailed description of the plugin and I propose an equivalent implementation in sbt.

Most of the time, the logic can be moved to either the default setting value or to the upstream libraries (librarymanagement, zinc).

# Description

## Scala 3 Instances

| Key | Explanation | Proposed Implementation |
|-|-|-|
| `isDotty` | new boolean setting | There already is `ScalaInstance.isDotty` in sbt but the setting is easier to use and it will be very useful during the migration to Scala 3.<br> `isDotty` will be renamed `isScala3` |
| `managedScalaInstance` | `false` if `isDotty`.<br>The Scala instances are fetched by `sbt-dotty` | Should remain `true`. <br>sbt will manage Scala 3 instances |
| `scalaInstance` | Call `dependencyResolution.value.update` to fetch the Scala 3 compiler and library, as well as the Scala 2.13 library | Update `ScalaArtifacts` in `librarymanagement` to manage Scala 3 artifacts.<br>**Should we support Dotty versions (0.xxx)?** |
| `libraryDependencies` | Add the `scala3-library` dependency. | Update `Defaults.autoLibraryDependency` to add `scala3-library` instead of `scala-library`<br>(I guess `scala-library_2.13` is added transitively by the `scala3-library` dependency) |
| `scalaBinaryVersion` | Add some logic for RC versions:<br>The binary version of `0.27.0-RC1` is `0.27`<br>The binary version of `3.0.0-M1-bin-20201027-b5a1715-NIGHTLY` is `3.0.0-M1` | Update the default `scalaBinaryVersion` logic if necessary.<br>(Will have RC versions after for `3.0.0`?) |

## Zinc Config

| Key | Explanation | Proposed Implementation |
|-|-|-|
| `incOptions` | Add a `TastyFileManager` as an external hook of the `incOptions` so that the `.tasty` files are pruned  | Handle in Zinc |
[...]

PR #6206, opened 2020-12-16
Manage Scala 3 instances and fetch compiler bridges

adpi2:
Part of https://github.com/sbt/sbt/issues/6080
Depends on https://github.com/sbt/librarymanagement/pull/350

## Manage Scala 3 instances

When Scala version is Scala 3:
- the `scala3-compiler` artifact is added to the list of tool dependencies. `scala3-compiler` transitively depends on `scala3-library` and `scala-library`
- if `autoScalaLibrary := true`, the `scala3-library` is added to the list of library dependencies

Coursier does not handle `autoScalaLibrary` properly with Scala 3. But we don't really need to set `autoScalaLibrary = true` in the Coursier configuration because `sbt` already add the proper scala library to the list of dependencies.

## Fetch the Scala 3 compiler bridge binaries

Contrary to Scala 2, the Scala 3 compiler bridges are tied to the compiler versions. There is a single compiler bridge for each compiler version. Also they are pure java so binary compatible with all Scala versions.

Consequently, we can fetch the binary module directly without using the `ZincCompilerBridgeProvider`.

## Manage the .tasty files

This PR introduces an `auxiliaryClassFiles` setting that is passed to Zinc via `incOptions`.
When the Scala version is Scala 3, the `auxiliaryClassFiles` setting contains the `TastyFiles` singleton. Otherwise it is empty.

One can add more `auxiliaryClassFiles` with, for example, `auxiliaryClassFiles += ScalaJSFiles.instance`


adpi2:
@eed3si9n How can I get the CI to run against the latest `librarymanagement`?
(without releasing `librarymanagement` because I may have some changes do to around `IvyDependencyResoltuion` about `overrideScalaVersion`)

eed3si9n:
@adpi2 Not possible now, but it shouldn't be too hard to tweak https://github.com/sbt/sbt/blob/develop/.github/workflows/nightly.yml or copy-paste it and adding a `workflow_dispatch` similar to https://github.com/sbt/sbt-dist/blob/master/.github/workflows/distro.yml

eed3si9n:
@adpi2 I just released https://github.com/sbt/librarymanagement/releases/tag/v1.5.0-M1

adpi2:
@eed3si9n Thanks for publishing `librarymanagement`. I will also need latest zinc before merging this PR.

I changed the CI of this PR to make it run on fresh `io`, `zinc` and `librarymanagement` `1.5.0-SNAPSHOT`, published locally.
But the scripted tests cannot resolve the `compiler-bridge` `1.5.0-SNAPSHOT` because they don't use the user ivy-cache:
- The `compiler-bridge` is published to `/home/runner/.ivy2/local/org.scala-sbt/compiler-bridge_2.12/1.5.0-SNAPSHOT/poms/compiler-bridge_2.12.pom`
- But the scripted test try to resolve it from:
```
[info] [warn] 	module not found: org.scala-sbt#compiler-bridge_2.12;1.5.0-SNAPSHOT
[info] [warn] ==== local: tried
[info] [warn]   /tmp/sbt_9611947f/target/ivy-cache/local/org.scala-sbt/compiler-bridge_2.12/1.5.0-SNAPSHOT/ivys/ivy.xml
```

Do you know how I can force the scripted tests to use the user ivy cache?

adpi2:
In fact, it seems that my ivy cache problem is due to the settings of some particular scripted test and not to scripted tests in general:

For instance in `dependency-management/artifact/build.sbt`
https://github.com/sbt/sbt/blob/21fc19a193a81116f38ec531d4412f1a57fc8a62/sbt/src/sbt-test/dependency-management/artifact/build.sbt#L20
overrides the user global cache.

adpi2:
I adapted those tests with
```scala
scalaCompilerBridgeResolvers += userLocalFileResolver(appConfiguration.value)

// use the user local resolver to fetch the SNAPSHOT version of the compiler-bridge
def userLocalFileResolver(appConfig: AppConfiguration): Resolver = {
  val ivyHome = appConfig.provider.scalaProvider.launcher.ivyHome
  Resolver.file("User Local", ivyHome / "local")(Resolver.defaultIvyPatterns)
}
```

and it works!

adpi2:
This is ready for review!

The `autoScalaLib = false,` in `LMCoursier` configuration may not be ideal but it doesn't seem to cause much trouble. Hopefully I'll be able to adapt Coursier to Scala 3 artifacts but it does not seem so simple: `autoScalaLib` is not an `lmcoursier` only param, it goes deep into Coursier code itself.

adpi2:
For now I have just created an issue in the Coursier repository about `autoScalaLib` and `forceScalaVersion`: https://github.com/coursier/coursier/issues/1933

smarter:
> Coursier does not handle autoScalaLibrary properly with Scala 3. But we don't really need to set autoScalaLibrary = true in the Coursier configuration because sbt already add the proper scala library to the list of dependencies.

I'm not sure I understand what that means, will the sbt setting `autoScalaLibrary := false` still work correctly in scala 3 projects? If sbt deals with adding the library itself, then could the parameter passed to coursier always be `autoScalaLibrary =  false` or would that do the wrong thing in Scala 2? In any case it sounds like we should open an issue against coursier /cc @alexarchambault

smarter:
Oh oops, I read the description but missed the comment where you mentioned https://github.com/coursier/coursier/issues/1933 which I guess is the follow-up.

adpi2:
> I'm not sure I understand what that means, will the sbt setting autoScalaLibrary := false still work correctly in scala 3 projects? If sbt deals with adding the library itself, then could the parameter passed to coursier always be autoScalaLibrary = false or would that do the wrong thing in Scala 2?

So yes the `autoScalaLibrary := flase` will work correctly in Scala 3 and Scala 2 projects. Also `autoScalaLibrary := true` still works perfectly with Scala 2.

The only  thing that is different with `autoScalaLibrary := true` and Scala 3 is that Coursier will not try to override the Scala version of the dependencies to Scala artifacts. I guess that it is completely harmless in the very large majority of cases.

Also I think that choosing a custom `scalaOrganization` does not work with Coursier. Is it even supported with Scala 2? I don't think so. But clearly in Scala 3 we don't need that yet, and it works with Ivy.

smarter:
> The only thing that is different with autoScalaLibrary := true and Scala 3 is that Coursier will not try to override the Scala version of the dependencies to Scala artifacts. I guess that it is completely harmless in the very large majority of cases.

It's something we'll have to figure out at some point, there's an issue open on that subject: https://github.com/lampepfl/dotty/issues/10146

eed3si9n:
Do we know why things are failing on AppVeyor?

```
[error] C:\projects\sbt\main\src\main\scala\sbt\Defaults.scala:91:8: object TastyFiles is not a member of package xsbti.compile
[error] import xsbti.compile.TastyFiles
[error]        ^
[error] C:\projects\sbt\main\src\main\scala\sbt\Keys.scala:251:41: not found: type AuxiliaryClassFiles
[error]   val auxiliaryClassFiles = taskKey[Seq[AuxiliaryClassFiles]]("The auxiliary class files that must be managed by Zinc (for instance the TASTy files)")
[error]                                         ^
[error] C:\projects\sbt\main\src\main\scala\sbt\Defaults.scala:889:10: value withAuxiliaryClassFiles is not a member of xsbti.compile.IncOptions
[error] possible cause: maybe a semicolon is missing before `value withAuxiliaryClassFiles'?
[error]         .withAuxiliaryClassFiles(auxiliaryClassFiles.value.toArray)
[error]          ^
[error] three errors found
[error] (mainProj / Compile / compileIncremental) Compilation failed
[error] Total time: 147 s (02:27), completed Dec 23, 2020 10:17:18 AM
```

adpi2:
> Do we know why things are failing on AppVeyor?

AppVeyor is failing because it needs to compile against the `develop` branch of `zinc` and `librarymanagement`.

For `librarymanagement` I can use the latest release but I don't think that `zinc` has been released recently and I need https://github.com/sbt/zinc/pull/956.

The same holds for Github Actions obviously but I temporarily changed the `.github/workflows/ci.yaml` file to run against `io`, `zinc` and `librarymanagement` develop branch: https://github.com/sbt/sbt/pull/6206/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f

Once zinc is released I can revert those changes in the CI.

adpi2:
@eed3si9n

I think that there is something wrong with zinc 1.4.4 because it does not contain the recent `TastyFiles` and other classes that I added in [zinc/internal/zinc-core/src/main/java/xsbti/compile](https://github.com/sbt/zinc/tree/develop/internal/zinc-core/src/main/java/xsbti/compile)

I checked the source files in https://github.com/sbt/zinc/releases/tag/v1.4.4 and I cannot find the corresponding source files...

adpi2:
@eed3si9n I need a release of zinc that contains https://github.com/sbt/zinc/pull/956

eed3si9n:
@adpi2 Let me get on that.

eed3si9n (approved):

Issue #6204, opened 2020-12-15
Regression: tab completion in SBT 1.4.5 terribly slow compared to SBT 1.3.10


DavidPerezIngeniero:
## steps

Update build.properties from 1.3.10 to 1.4.5

## problem

I have several inputTasks with the same prefix.
I press 5 characters, and press tab, and I must wait for over 1 minute.  Before it was about 1 second.

On the other side, completing arguments to input tasks has good performance.
Completing task names has bad performance.

I press tab in the middle of a line: now no completion, before: it completed.

## expectation

Same fast response.

## notes

If necessary, I can provide privately my build config (it isn't simple).

eed3si9n:
@DavidPerezIngeniero Thanks for the report. I'm guessing that this is related to our update to JLine 3, but don't know exactly how.
Does the slow completion only happen for input tasks?

DavidPerezIngeniero:
If I press e.g. "comp" and press tab, the result is immediate.
```
> com
commands                    compileEarly                compileScalaBackend
compile                     compileIncSetup             compileSplit
compile:                    compileIncremental          compilers
compileAnalysisFile         compileJava                 completions
compileAnalysisFilename     compileOutputs              compruebaCron
compileAnalysisTargetRoot   compileProgress
```
but if I press "distr", it takes 10 seconds:
```
> distribuye
distribuye               distribuyeFuentes        distribuyeLibreOffice    distribuyeSudoers
distribuyeAmm            distribuyeGit            distribuyeLimites        distribuyeSwapiness
distribuyeConfigApache   distribuyeGuionesRoot    distribuyePostix         distribuyeZk
distribuyeCron           distribuyeHomeEma        distribuyeServicio
```
sometimes much more.  All of them are input tasks.

SethTisue:
I have noticed that tab completion on `++` is quite slow. Over in https://github.com/scala/scala-parser-combinators, this:

```
sbt:root> ++2<TAB>
++2.11.12   ++2.12.12   ++2.13.4
```

takes seven long seconds.

eed3si9n:
@SethTisue Thanks. I confirmed locally that downgrading to sbt 1.3.13 (after downgrading sbt-dotty too), `++2` would complete immediately, so there's something going on with sbt 1.4.x.

eed3si9n:
edf43a473b18f5c185475aa39d227eeb94e3ea8e (https://github.com/sbt/sbt/pull/5719) is the offending commit according to bisect:

```
[...]

PR #6214, opened 2020-12-19
Fix tab completion perf regression

eed3si9n:
This reverts commit edf43a473b18f5c185475aa39d227eeb94e3ea8e.
This fixes performance regression of tab completion observed in https://github.com/sbt/sbt/issues/6204

As originally mentioned by @eatkins, JLine 3 will automatically append a white space character (`" "`) to the completion
candidate. In practice it doesn't really seem to bother the shell if we put in `"++2.11.12"` or `"++2.11.12 "` with the whitespace so I think it's ok to do so for now.

Perhaps in the future the parser could return a signal indicating `++2.11.12` and then End-of-Line or Whitespace-and-more?


DavidPerezIngeniero:
Tried SBT 1.4.6 and now it's much quicker.  Thanks!

PR #6234, opened 2020-12-29
Add support for Scala3doc

adpi2:
Part of #6080


adpi2:
This PR depends on:
- https://github.com/sbt/librarymanagement/pull/359
- https://github.com/sbt/zinc/pull/960

The `scala3doc` dependency is added only if `externalResolvers` contains the JCenter repository.

It is added to a separate ivy configuration, called `ScalaDocTool`, so that we have two distinct lists of jars: one list for `scala3-compiler` and another one for `scala3doc`. Then we can easily create the compiler class loader and the `scala3doc` class loader whose parent is the compiler class loader.

To make `doc` works with Scala 3, the user must add the `JCenter` repository manually:
```
resolvers += Resolver.JCenterRepostiory
```

Once `scala3doc` has been migrated away from dokka, JCenter won't be needed anymore and the `scala3doc` dependency will be added by default.

In order to merge this PR, we need to merge https://github.com/sbt/librarymanagement/pull/359 and https://github.com/sbt/zinc/pull/960 and to release them. Also I will remove the commit https://github.com/sbt/sbt/pull/6234/commits/5dfb3042b3a35deaf89029a592ede1b4c8865797 (it forces the CI to run on my zinc and libmanagement branches)

eed3si9n:
Zinc 1.5.0-M2 and lm 1.5.0-M3 are available.

adpi2:
Thanks @eed3si9n
It's now ready for review.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1991,14 +2019,35 @@ object Defaults extends BuildCommon {
...
+        },
+        sources := Def.taskDyn {
+          val originalSources = sources.value
+          val sv = scalaVersion.value
+          if (ScalaArtifacts.isScala3(sv) && originalSources.nonEmpty) Def.task {
+            val _ = compile.value
+            val tastyFiles = classDirectory.value.**("*.tasty").get
+            tastyFiles.map(_.getAbsoluteFile)
+          } else Def.task(originalSources)
+        }.value,
```
eed3si9n: I think sources should mean `*.scala`, `*.java`, etc.
I suggest we introduce a new key called `tastyFiles` or something, and return `Nil` when it's not Scala 3. That way we can avoid the use of dynamic task too.

eed3si9n (approved):

PR #6237, opened 2021-01-03
packageTimestamp setting

### LINKED ISSUE #6235: Make it possible to opt out of reproducible builds (1970-01-01 timestamps)
Ref https://reproducible-builds.org/
Originally reported by @bchazalet as https://github.com/playframework/playframework/issues/10572

## steps

Play app using sbt 1.4.x.

1. Package a play application using `universal:packageZipTarball`
2. Deploy the application
3. Send a GET request for one of the static asset, say
4. Inspect the last-modified response header

## problem

depending on your platform the last modified header will either be set to 1 Jan 1970 or 7 February 2106 (which known to be a common time storage [issue](https://en.wikipedia.org/wiki/Time_formatting_and_storage_bugs#Year_2106)).

## expectation

Inspect the last-modified response header and assert that it aligns with when you packaged the application in step 1 (with some additional setting)

## notes by @dontgitit in https://github.com/sbt/sbt/issues/5344#issuecomment-753304169_

+1 to the previous comments and the linked play issue https://github.com/playframework/playframework/issues/10572, this change breaks etag computation in play. while the suggested fix in that thread (of setting an environment variable) does work, it's frustrating that there's no way to do it directly from sbt. it might be nicer if there was an sbt setting for this, `useEpochTimestamp in packageBin` or something of the sort? 

also, since the new code always provides `Some(0)` as a backup, it seems there's no way to go back to the old behavior of preserving asset timestamps; you must either choose a static date for all of them, or you get the epoch by default. this seems suboptimal - it would be ideal to preserve asset timestamps since they rarely change.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/6235

In sbt 1.4.0 (#5344) we started wiping out the timestamps in JAR
to make the builds more repeatable.
This had an unintended consequence of breaking Play's last-modified response header (playframework/playframework#10572).

This adds a global setting called `packageTimestamp`, which is
initialized as follows:

```scala
packageTimestamp :== Package.defaultTimestamp,
```

Here the `Package.defaultTimestamp` would pick either the value from the
`SOURCE_DATE_EPOCH` environment variable or 2010-01-01.

To opt out of this default, the user can use:

```scala
ThisBuild / packageTimestamp := Package.keepTimestamps

// or

ThisBuild / packageTimestamp := Package.gitCommitDateTimestamp
```

### Before (sbt 1.4.6)

```
$ ll example
total 32
-rw-r--r--  1 eed3si9n  wheel   901 Jan  1  1970 Greeting.class
-rw-r--r--  1 eed3si9n  wheel  3079 Jan  1  1970 Hello$.class
-rw-r--r--  1 eed3si9n  wheel   738 Jan  1  1970 Hello$delayedInit$body.class
-rw-r--r--  1 eed3si9n  wheel   875 Jan  1  1970 Hello.class
```

### After (using Package.gitCommitDateTimestamp)

```
$ unzip -v target/scala-2.13/root_2.13-0.1.0-SNAPSHOT.jar
Archive:  target/scala-2.13/root_2.13-0.1.0-SNAPSHOT.jar
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
     288  Defl:N      136  53% 01-25-2021 03:09 888682a9  META-INF/MANIFEST.MF
       0  Stored        0   0% 01-25-2021 03:09 00000000  example/
     901  Defl:N      601  33% 01-25-2021 03:09 3543f377  example/Greeting.class
    3079  Defl:N     1279  59% 01-25-2021 03:09 848b4386  example/Hello$.class
     738  Defl:N      464  37% 01-25-2021 03:09 571f4288  example/Hello$delayedInit$body.class
     875  Defl:N      594  32% 01-25-2021 03:09 ad295259  example/Hello.class
--------          -------  ---                            -------
    5881             3074  48%                            6 files
```

### DIFF
--- a/main-actions/src/main/scala/sbt/Package.scala
+++ b/main-actions/src/main/scala/sbt/Package.scala
@@ -8,6 +8,7 @@
 package sbt
 
 import java.io.File
+import java.time.OffsetDateTime
 import java.util.jar.{ Attributes, Manifest }
 import scala.collection.JavaConverters._
 import sbt.internal.util.Types.:+:
@@ -23,6 +24,7 @@ import sbt.internal.util.HListFormats._
 import sbt.util.FileInfo.{ exists, lastModified }
 import sbt.util.CacheImplicits._
 import sbt.util.Tracked.{ inputChanged, outputChanged }
+import scala.sys.process.Process
 
 sealed trait PackageOption
 
@@ -43,6 +45,40 @@ object Package {
     val converted = for ((name, value) <- attributes) yield (new Attributes.Name(name), value)
     new ManifestAttributes(converted: _*)
   }
+  // 2010-01-01
+  private val default2010Timestamp: Long = 1262304000000L
+  final case class FixedTimestamp(value: Option[Long]) extends PackageOption
+  val keepTimestamps: Option[Long] = None
+  val fixed2010Timestamp: Option[Long] = Some(default2010Timestamp)
+  def gitCommitDateTimestamp: Option[Long] =
+    try {
+      Some(
+        OffsetDateTime
+          .parse(Process("git show -s --format=%cI").!!.trim)
+          .toInstant()
+          .toEpochMilli()
+      )
+    } catch {
+      case e: Exception if e.getMessage.startsWith("Nonzero") =>
+        sys.error(
+          s"git repository was expected for package timestamp; use Package.fixed2010Timestamp or Package.keepTimestamps instead"
+        )
+    }
+  def setFixedTimestamp(value: Option[Long]): PackageOption =
+    FixedTimestamp(value)
+
+  /** by default we overwrite all timestamps in JAR to epoch time 2010-01-01 for repeatable build */
+  lazy val defaultTimestamp: Option[Long] =
+    sys.env
+      .get("SOURCE_DATE_EPOCH")
+      .map(_.toLong * 1000)
+      .orElse(Some(default2010Timestamp))
+
+  def timeFromConfiguration(config: Configuration): Option[Long] =
+    (config.options.collect { case t: FixedTimestamp => t }).headOption match {
+      case Some(FixedTimestamp(value)) => value
+      case _                           => defaultTimestamp
+    }
 
   def mergeAttributes(a1: Attributes, a2: Attributes) = a1.asScala ++= a2.asScala
   // merges `mergeManifest` into `manifest` (mutating `manifest` in the process)
@@ -70,9 +106,14 @@ object Package {
       val options: Seq[PackageOption]
   )
 
-  @deprecated("Please specify whether to use a static timestamp", "1.4.0")
+  /**
+   *
+   * @param conf the package configuration that should be build
+   * @param cacheStoreFactory used for jar caching. We try to avoid rebuilds as much as possible
+   * @param log feedback for the user
+   */
   def apply(conf: Configuration, cacheStoreFactory: CacheStoreFactory, log: Logger): Unit =
-    apply(conf, cacheStoreFactory, log, None)
+    apply(conf, cacheStoreFactory, log, timeFromConfiguration(conf))
 
   /**
    *
@@ -94,6 +135,7 @@ object Package {
         case JarManifest(mergeManifest)          => mergeManifests(manifest, mergeManifest); ()
         case MainClass(mainClassName)            => main.put(Attributes.Name.MAIN_CLASS, mainClassName); ()
         case ManifestAttributes(attributes @ _*) => main.asScala ++= attributes; ()
+        case FixedTimestamp(value)               => ()
         case _                                   => log.warn("Ignored unknown package option " + option)
       }
     }
@@ -163,7 +205,8 @@ object Package {
       homepage map (h => (IMPLEMENTATION_URL, h.toString))
     }: _*)
   }
-  @deprecated("Please specify whether to use a static timestamp", "1.4.0")
+
+  @deprecated("Specify whether to use a static timestamp", "1.4.0")
   def makeJar(sources: Seq[(File, String)], jar: File, manifest: Manifest, log: Logger): Unit =
     makeJar(sources, jar, manifest, log, None)
 
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -223,6 +223,7 @@ object Defaults extends BuildCommon {
       bgCopyClasspath :== true,
       closeClassLoaders :== SysProp.closeClassLoaders,
       allowZombieClassLoaders :== true,
+      packageTimestamp :== Package.defaultTimestamp,
     ) ++ BuildServerProtocol.globalSettings
 
   private[sbt] lazy val globalIvyCore: Seq[Setting[_]] =
@@ -1599,20 +1600,27 @@ object Defaults extends BuildCommon {
           val org = organization.value
           val orgName = organizationName.value
           val main = mainClass.value
+          val ts = packageTimestamp.value
           val old = packageOptions.value
+
           Package.addSpecManifestAttributes(n, ver, orgName) +:
             Package.addImplManifestAttributes(n, ver, homepage.value, org, orgName) +:
+            Package.setFixedTimestamp(ts) +:
             main.map(Package.MainClass.apply) ++: old
         }
       )
     ) ++
       inTask(packageSrc)(
         Seq(
-          packageOptions := Package.addSpecManifestAttributes(
-            name.value,
-            version.value,
-            organizationName.value
-          ) +: packageOptions.value
+          packageOptions := {
+            val old = packageOptions.value
+            val ts = packageTimestamp.value
+            Package.addSpecManifestAttributes(
+              name.value,
+              version.value,
+              organizationName.value
+            ) +: Package.setFixedTimestamp(ts) +: old
+          }
         )
       ) ++
       packageTaskSettings(packageBin, packageBinMappings) ++
@@ -1778,10 +1786,7 @@ object Defaults extends BuildCommon {
         config,
         s.cacheStoreFactory,
         s.log,
-        sys.env
-          .get("SOURCE_DATE_EPOCH")
-          .map(_.toLong * 1000)
-          .orElse(Some(1262304000000L)) // 2010-01-01
+        Package.timeFromConfiguration(config)
       )
       config.jar
     }
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -265,6 +265,7 @@ object Keys {
   val packageCache = taskKey[File]("Produces the main artifact for caching.")
 
   val packageOptions = taskKey[Seq[PackageOption]]("Options for packaging.").withRank(BTask)
+  val packageTimestamp = settingKey[Option[Long]]("Overwrites timestamps in JAR file to make the build reproducible; None keeps the existing timestamps (useful for web resources)").withRank(CSetting)
   val packageConfiguration = taskKey[Package.Configuration]("Collects all inputs needed for packaging.").withRank(DTask)
   val artifactPath = settingKey[File]("The location of a generated artifact.").withRank(BPlusSetting)
   val artifact = settingKey[Artifact]("Describes an artifact.").withRank(BMinusSetting)


nafg:
Wait what? Does any other build tool or packaging tool nuke timestamps like that?

Since when does does reproducible build mean "same input file contents -> same file timestamp"? Does something break if timestamps are preserved?

Web resource caching is one major reason to preserve timestamps but I'm sure there are others, nuking them seems unwise without a really strong reason, no?

eed3si9n:
@nafg The point of reproducible builds is to get bit-for-bit equivalent binary artifact regardless of who builds things to make sure the JAR is not tampered by the developer etc. iiuc wiping out the timestamp is fairly common thing to do. See for example:
- https://salsa.debian.org/reproducible-builds/strip-nondeterminism.
- https://zlika.github.io/reproducible-build-maven-plugin/

> replaces file timestamps in ZIP entries with a fixed value,

nafg:
Very interesting, thanks for the pointers.

However I still think a default of the Unix epoch violates the
principle of least surprise.

From what I'm reading it seems like:

   - reproducible-builds.org *seems* to require a SOURCE_DATE_EPOCH
   environment variable. It says if it's malformed the build process should
   exit. I'm not sure what the behavior would be if it's unset. However it
   sounds a little like it's only for the context of a distribution, and for a
   self-distributed package the last modification time of the source is better.
   - Maven seems to make it opt-in and require setting
   project.build.outputTimestamp to overwrite timestamps.

I could be misunderstanding what they are saying, but I still think that
making everything 50 years old by default is not optimal.

Why not have a Setting[Option[Instant]] for overwriting timestamps, with a
default of None which means keep files' dates (I could always set those
directly outside of sbt). Then I could set it to Some(Instant.EPOCH) or
Some(Instant.now()), and there could be shortcuts to set it to the time of
the last commit, of the last commit of a changelog, or something else.

Or the default could be env.get("SOURCE_DATE_EPOCH").map(s =>
Instant.ofEpochSecond(s.toLong)).

gmethvin:
This feature should be opt-in rather than opt-out.

I certainly see the value in reproducible builds, but there are obviously use cases where the developer may be using the timestamps for something meaningful. In those cases their application will break when they upgrade sbt, which is a very surprising thing to happen when upgrading to a new minor version.

The biggest issue to me is that you're setting the timestamp to an _intentionally incorrect_ value by default. It's a reasonable hack to solve the problem of reproducible builds, but it's still a hack.

gmethvin:
I agree with @nafg's ideas on the API.

I do think it's great for sbt to provide an easy way to reset timestamps to some fixed value, and I'm sure a lot of people will prefer that behavior. I just think it's not a good default.

eed3si9n:
I should cc @raboof here.

eed3si9n:
For incremental compilation, relying only on timestamp is not always reliable due to the time resolution in Docker filesystems etc, so I've switched to using non-cryptographic hash to perform invalidation in sbt 1.4.x.

For the purpose of reproducible builds, the original intent was to wipe out the timestamps of `*.class` files since they would change depending on when the CI machine or local laptop ran the `compile` task. Maybe a compromise middle ground may be to somehow distinguish resources vs `*.class` files?

> The biggest issue to me is that you're setting the timestamp to an _intentionally incorrect_ value by default.

What is the correct value? Note that git intentionally does not record modified time.

nafg:
On Mon, Jan 4, 2021, 12:24 AM eugene yokota <notifications@github.com>
wrote:

> For incremental compilation, relying only on timestamp is not always
> reliable due to the time resolution in Docker filesystems etc, so I've
> switched to using non-cryptographic hash to perform invalidation in sbt
> 1.4.x.
>
> For the purpose of reproducible builds, the original intent was to wipe
> out the timestamps of *.class files since they would change depending on
> when the CI machine or local laptop ran the compile task. Maybe a
> compromise middle ground may be to somehow distinguish resources vs
> *.class files?
>
Is the issue of cross-machine incremental compilation necessarily
intertwined with the objective of verifying artifact correctness via build
reproducibility?

The biggest issue to me is that you're setting the timestamp to an
*intentionally
> incorrect* value by default.
>
> What is the correct value? Note that git intentionally does not record
> modified time.
>

I don't think there is a single correct value. I think the main
considerations here are principle of least surprise, and not breaking
previous behavior (both which come down to expectation).

I think both of those considerations are violated by Unix epoch. And the
second one is necessarily violated by bit-reproducible builds.

But I should let Greg answer what he meant.

raboof:
I'm sorry to see this change has caused some surprising behavior.

> Since when does does reproducible build mean "same input file contents -> same file timestamp"? Does something break if timestamps are preserved?

Indeed it doesn't - I agree the best thing to do is to preserve timestamps from the original file modification times. However, since

> git intentionally does not record modified time

... in the majority of cases the modified-time of resources is somewhat-arbitrary, and the modified-time of classes is "whenever the compilation was done", which is also somewhat-arbitrary. I don't think we should allow those to end up in the artifact by default.

> The biggest issue to me is that you're setting the timestamp to an _intentionally incorrect_ value by default. It's a reasonable hack to solve the problem of reproducible builds, but it's still a hack

I think using 'epoch' to signify "unknown" (in a file format that does not allow leaving out the timestamp) is pretty reasonable. I think setting the timestamp to "unknown" rather than a somewhat-arbitrary value is an improvement, even though I agree it is a bit of a sledgehammer.

For a somewhat less 'sledgehammer' (but still rough) workaround, you could set `SOURCE_DATE_EPOCH` to the timestamp of the last commit at build time. This will set the timestamp to a stable, somewhat-reasonable value, though it will of course over-estimate the timestamp of files that were not changed on the last commit.

The 'ideal' solution would be to keep the source modification timestamps for those files where we know those timestamps can be assumed to be 'correct'. Unfortunately since in most typical scenario's (git) _no_ timestamps can assumed to be 'correct', I think this should be opt-in.

gmethvin:
> For the purpose of reproducible builds, the original intent was to wipe out the timestamps of *.class files since they would change depending on when the CI machine or local laptop ran the compile task. Maybe a compromise middle ground may be to somehow distinguish resources vs *.class files?

I think doing it only for class files makes a bit more sense, as it's sbt's job to compile your code to class files, but it still seems risky to do by default. What if users have other tools that rely on those modify times?

> What is the correct value? Note that git intentionally does not record modified time.

The "correct" time is the time that the file was last modified according to the filesystem. Of course that value may not be universally useful or reliable, but in some cases users do care what it is.

gmethvin:
> The 'ideal' solution would be to keep the source modification timestamps for those files where we know those timestamps can be assumed to be 'correct'. Unfortunately since in most typical scenario's (git) no timestamps can assumed to be 'correct', I think this should be opt-in.

sbt doesn't make any particular guarantees about the file timestamps being "correct", does it? That's typically the responsibility of the user or the user's filesystem to determine. Do other build tools reset the timestamps by default?

There are plenty of situations where you'll get predictable, meaningful values for the modified timestamps. If I clone a git repository fresh, I can generally assume the modified times will be close to the time the repository was cloned. It would not surprise me if people are relying on that assumption, even if it might not be a good idea.

eed3si9n:
The guarantee is that if you put `println("hello world")` you get a bytecode that does hello world. Everything else is somewhat arbitrary and implementation detail. Some build tools use timestamp as invalidation mechanism to implement incremental compilation, whereas sbt 1.4.x doesn't. As much as I do respect playing nice with existing norms, I think there's a real benefit in being able to git clone a repository in 2020 or 2030 and getting bit-for-bit identical JAR file that can be validated using SHA-256 (or whatever hash algorithm we'd use in 10 years).

> Do other build tools reset the timestamps by default?

It looks like Bazel does.

```
$ unzip -v bazel-bin/greeting/greeting-bin.jar
Archive:  bazel-bin/greeting/greeting-bin.jar
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
       0  Stored        0   0% 01-01-2010 00:00 00000000  META-INF/
      97  Stored       97   0% 01-01-2010 00:00 5a07476d  META-INF/MANIFEST.MF
       0  Stored        0   0% 01-01-2010 00:00 00000000  com/
       0  Stored        0   0% 01-01-2010 00:00 00000000  com/twitter/
       0  Stored        0   0% 01-01-2010 00:00 00000000  com/twitter/greeting/
    2481  Defl:N     1014  59% 01-01-2010 00:00 c6eb159c  com/twitter/greeting/Main$.class
     860  Defl:N      491  43% 01-01-2010 00:00 5429a12f  com/twitter/greeting/Main$delayedInit$body.class
    1021  Defl:N      653  36% 01-01-2010 00:00 55e294cb  com/twitter/greeting/Main.class
--------          -------  ---                            -------
    4459             2255  49%                            8 files
```

gmethvin:
It's not surprising that Bazel sets the timestamps to a fixed value, given that it's very opinionated about reproducible builds.

The main problem is that we're introducing a breaking change, since sbt is choosing to take responsibility for something it previously was not responsible for. If sbt wants to be opinionated about this—which in itself is certainly not a bad thing—then the change should wait for 2.0. Most users expect to be able to upgrade to a new minor version without breaking changes. I don't even see any migration notes for 1.4.x mentioning this.

Like I've said, I have no problem with the motivation behind the change or the value of reproducible builds, but also I believe maintainers have a responsibility to their users to not break things unless they have a really good reason to. sbt is now a very widely used tool, and I don't think being opinionated about reproducibility is a good enough reason to change the default behavior.

eed3si9n:
> The main problem is that we're introducing a breaking change, since sbt is choosing to take responsibility for something it previously was not responsible for. If sbt wants to be opinionated about this—which in itself is certainly not a bad thing—then the change should wait for 2.0. Most users expect to be able to upgrade to a new minor version without breaking changes. I don't even see any migration notes for 1.4.x mentioning this.

Regardless of the outcome of this particular feature, I want to clarify this notion about minor version upgrade. I adopted Semantic Versioning for sbt so we can introduce new features and sometimes breaking changes without burning the forest down with binary incompatibility. Since sbt 1.0 in 2017, we shipped feature releases around year-apart with fairly aggressive changes like introducing slash syntax, switching to Coursier, and implementing remote caching. Following the logic that any breaking change would require major version bump, we'd need to do that annually, requiring unnecessary republishing of plugins.

FWIW [sbt 1.4.0 release notes](https://www.scala-sbt.org/1.x/docs/sbt-1.4-Release-Notes.html#Fixes+with+compatibility+implications) do mention this as breaking change.

gmethvin:
It's of course up to sbt's maintainers how the versioning scheme should work, and I see the benefit of introducing changes without having to re-publish plugins. But even from a loose semver perspective the new behavior is very surprising, especially because people are much more likely to notice the issue once the application is deployed, rather than at build time.

> FWIW sbt 1.4.0 release notes do mention this as breaking change.

You're right, it does mention this:

> Makes JAR file creation repeatable by sorting entry by name and dropping timestamps

But there are no migration notes about that at all, I guess because it wasn't obvious what the impact would be.

h-vetinari:
> For a somewhat less 'sledgehammer' (but still rough) workaround, you could set `SOURCE_DATE_EPOCH` to the timestamp of the last commit at build time. This will set the timestamp to a stable, somewhat-reasonable value, though it will of course over-estimate the timestamp of files that were not changed on the last commit.

I find this to be a very elegant solution - the last commit is always known, and changing the commit necessarily voids the expectation of reproducibility.

gmethvin:
Using the timestamp of the last commit would be a more reasonable default. That should give you reproducibility and should also be suitable for the web resource caching situations. Then the timestamp is actually somewhat meaningful.

I also like the idea of `Option[Instant]` setting like @nafg suggested that lets you override with a specific timestamp if you prefer.

sideeffffect:
+1 on timestamp of the last commit as the default.
If we wanted, the timestamps could be more granual -- theoretically, a timestamp could be determined for each individual file based on the commit that modified the file last. But I don't know if it's worth it.

Anyway, reproducibility is a hugely important thing, so thanks for that.

nafg:
How can it be the default, if there's no guarantee SBT is running in a git
repo?

sideeffffect:
> How can it be the default, if there's no guarantee SBT is running in a git repo?

It's nothing complicated: you can fallback to `0` (or the actual timestamp from the file system) if there is no git repository to be found (or any other VCS sbt will know to work with).

nafg:
The default fallback should not be zero either

gmethvin:
The fallback I was thinking of would be to not set a timestamp at all. Maybe we could also display a warning to let the user know how to manually define the timestamp strategy. We could attempt to check for other version control systems, though I think git is pervasive enough that it would be fine to only support git by default.

gmethvin:
Actually, a more user-friendly way to be opinionated here would be to display a warning when packaging a project unless you explicitly set something.

For example, when packaging the project, you could display something like:

```
[warn] This build is not reproducible! Timestamps may differ between builds.
[warn]
[warn] To remove this warning, choose a timestamp strategy. For example:
[warn]
[warn]   Compile / packageBin / timestampStrategy := Package.gitCommitTimestamp
[warn] 
[warn] For more details see: <link>
```


Other options could be `keepTimestamp`, `fixedTimestamp(instant)`, etc. Obviously the exact naming, etc. could change, but the idea is you force the user to be aware of their decision.

eed3si9n:
Displaying warning by default is a tactic taken sometimes by different tooling or language, but I don't think it's usually a great option. Either the user doesn't have enough context to make the right choice, or it's not relevant to the problem they want to solve. A recent example that comes to mind is where we deprecated floating-point sorting by displaying a deprecation message.

I could create a global setting with type `Option[Long]` and default it to 2010-01-01, and maybe provide a function that calls out to git if people want to rewire the setting.

sideeffffect:
2010 feels way too arbitrary. 1970 at least communicates to the user that it's the zero-timestamp.

Time of the git commit > 0 > 2010-01-01, IMHO.

gmethvin:
> Displaying warning by default is a tactic taken sometimes by different tooling or language, but I don't think it's usually a great option. Either the user doesn't have enough context to make the right choice, or it's not relevant to the problem they want to solve.

IMO a warning would be preferable to a breaking change. If the user doesn't have enough context and makes the wrong choice, then the fact that they made an explicit change will make it easier to debug any issue that arises. If you make a suggestion that works for most people—like the git timestamp—then most users will be fine.

> A recent example that comes to mind is where we deprecated floating-point sorting by displaying a deprecation message.

I actually thought that decision made sense. There are tradeoffs between the IEEE 754 ordering and the  total ordering, and a warning is a good way to make sure the user has the opportunity to make an informed choice.

> I could create a global setting with type `Option[Long]` and default it to 2010-01-01, and maybe provide a function that calls out to git if people want to rewire the setting.

I think a setting like that would work. We just don't have consensus on what the default should be.

eed3si9n:
@sideeffffect 1970-01-01 doesn't work unfortunately because the WinZip format apparently does not support dates earlier than 1980-01-01, and since it doesn't carry timezone information 1980-01-01 itself probably should be avoided. I'm not sure why Bazel chose 2010-01-01, but https://github.com/bazelbuild/bazel/commit/475f40ad191180fcf538fbd2fe49524b3edbb08e shows that it used to be 1980, but it was changed to 2010 in 2018.

nafg:
Not everyone is publishing a library. Libraries are in the minority.
Modification times are useful information. A default of not preserving the
actual modification time when archiving makes no sense to me, even though
the ideal of increasing the number of binary artifacts on Maven Central
that can be audited this way seems laudable enough.

eed3si9n:
In terms of the scope of reproducibility, I don't think it matters whether something is a library or open sourced. In fact, the attackers might avoid public JARs if they wanted to stay stealth. Anyone who deploys anything with sensitive data probably should care about it.

> A default of not preserving the actual modification time when archiving makes no sense to me

The modification time is an arbitrary number of when someone git cloned, or ran the build. If someone uses remote caching feature, the binary timestamp could even be earlier than the source timestamp on a local machine. It would be an irrational thing to do if I started doing this on my own, but as previously mentioned this is a common technique, and I think we're in a good company to pursue reproducibility - https://reproducible-builds.org/projects/ lists Google, Arch Linux, Debian, and Tor among others.

The HTTP Last-Modified is an interesting use case of JAR timestamps (it's an interesting use case because getting a ballpark number is should provide useful caching), but I feel like it's more of an exception not the rule.

sideeffffect:
@eed3si9n May I ask, what is your opinion on getting the timestamp by default from git (time of the current commit)? That would be still deterministic and I hope it will satisfy people who would be discontent with `2010-01-01` everywhere.

eed3si9n:
@sideeffffect I think using 2010-01-01 and git commit date are both valid approaches with different tradeoffs. Fixed value like 2010-01-01 feels more principled, but it doesn't work well with web apps. Having a dependency on system-installed `git` to make a JAR file feels a bit leaky for build tool, but it's probably a more practical thing for production projects.

In that sense, it's analogous to sbt-dynver. For reproducibility and record-keeping purpose, all production versions should have a tag or commit id that we can trace version numbers to code. So the idea of automatically setting the version number from git makes a lot of sense.

sideeffffect:
We shouldn't even need to depend on a system installed `git` binary, since there's [JGit](https://www.eclipse.org/jgit/), no?

dwijnand:
JGit doesn't support git worktrees (https://bugs.eclipse.org/bugs/show_bug.cgi?id=477475) so I default to system git in sbt-dynver (and then I never needed to rebuild the indirection that sbt-git had over system git/JGit.)

gmethvin:
I think we can assume most developers have `git` on their system, as long as we fail in an obvious way to give them an opportunity to fix the issue or otherwise override the setting.

I guess one aspect worth considering is that the cloned git repository could be copied and built on some environment that didn't have `git`, in which case the build wouldn't reproduce.

nafg:
SBT is often run on CI in a docker container that does not have git.

eed3si9n:
I've added `packageTimestamp` setting that can be used to opt out of the current behavior. Not sure if we can all come to a full consensus on https://github.com/sbt/sbt/issues/6235, but I think having this setting is strictly better than the status quo, so I'm merging this.

PR #6266, opened 2021-01-19
Fix ExitCode calc on StashOnFailure

melezov:
Previous sbt.Exec != String match would never fail

### DIFF
--- a/main/src/main/scala/sbt/MainLoop.scala
+++ b/main/src/main/scala/sbt/MainLoop.scala
@@ -251,7 +251,7 @@ object MainLoop {
         } finally {
           // Flush the terminal output after command evaluation to ensure that all output
           // is displayed in the thin client before we report the command status. Also
-          // set the promt to whatever it was before we started evaluating the task.
+          // set the prompt to whatever it was before we started evaluating the task.
           restoreTerminal()
         }
         if (exec.execId.fold(true)(!_.startsWith(networkExecPrefix)) &&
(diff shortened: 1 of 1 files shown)


melezov:
Alternatively, we could match against the actual `sbtStashOnFailure` command instead of doing a Stringly match.

eed3si9n (commented):
Thanks @melezov!

Could you run scalafmt plz?

melezov:
I believe it's already set to run on compile (unless in CI).
I did run it explicitly and there was a NOOP.

I'm not sure why that CI is yelling at us, but there were a lot of NPEs before the "scalafmt check":

    Thread[JVMCI CompilerThread0,9,system]: Compilation of scala.collection.IterableOnceOps.exists(Function1) failed: org.graalvm.compiler.java.BytecodeParser$BytecodeParserError: java.lang.NullPointerException

melezov:
I've re-pushed with a tweak to `.gitignore` to get the CI running again...

melezov:
Mea culpa, I thought running `scalafmt` on root project would propagate into others.

I needed to run ` mainProj / scalafmt`, it's fixed now.

eed3si9n (approved):

eed3si9n:
> I needed to run mainProj / scalafmt, it's fixed now.

That's weird. Normally it just runs as part of compilation. Maybe it's something to do with that graalvm messages.

PR #6276, opened 2021-01-21
GNU Emacs support for sbtn and sbt --client

### LINKED ISSUE #6271: sbt server does not stay up when sbtn is called from emacs
## steps

run

```
sbtn compile
```

twice from within emacs.

## problem

When called from Emacs the first server dies and the second call starts a new one.

## expectation

I expect the first invocation to start the server, and the second one to use it. This works as expected if called from a terminal emulator such as urxvt-unicode.

## notes

https://github.com/hvesalai/emacs-scala-mode/pull/165

fommil:
fixes #6271

@hvesalai can you please take a look? Do you think this will be compatible with `sbt-mode` now that I'm enabling colours by default? If not, we can always disable the colours from `sbt-mode` with a system property.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,5 @@ npm-debug.log
 .idea
 .bloop
 .metals
+.bsp/
 /project/metals.sbt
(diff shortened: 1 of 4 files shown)


lightbend-cla-validator:
Hi @fommil,

hvesalai:
Sbt-mode can handle (and does handle) colors. Not a problem.

image

eed3si9n (approved):

Issue #5886, opened 2020-09-23
Duplicated `-P:semanticdb:targetroot` option on Test when semanticdbEnabled


adpi2:
## steps

Create a minimal sbt project with `semanticdbEnabled := true`

```scala
scalaVersion := "2.13.1"
semanticdbEnabled := true
```

## problem

Run `show <project> / Test / scalacOptions`

It prints:
```
[info] * -Xplugin:/home/piquerez/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scalameta/semanticdb-scalac_2.13.1/4.3.20/semanticdb-scalac_2.13.1-4.3.20.jar
[info] * -P:semanticdb:targetroot:/home/adpi2/hello/target/scala-2.13/meta
[info] * -Yrangepos
[info] * -P:semanticdb:targetroot:/home/adpi2/hello/target/scala-2.13/test-meta
``` 

The `-P:semanticdb:targetroot` options is set twice with different values. The compiler use the second value which is the one we want. Yet this does seem to be very robust. Also it confuses some downstream tooling such as Metals that try to find the semanticdb files in the wrong folder: `.../target/scala-2.13/meta` instead of `.../target/scala-2.13/test-meta`

## expectation
`show <project> / Test / scalacOptions` should print 

```
[info] * -Xplugin:/home/piquerez/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scalameta/semanticdb-scalac_2.13.1/4.3.20/semanticdb-scalac_2.13.1-4.3.20.jar
[info] * -Yrangepos
[info] * -P:semanticdb:targetroot:/home/adpi2/hello/target/scala-2.13/test-meta
``` 

## notes

The issue is even worse with Dotty since it prevents the project from compiling:
```
[error] not found: /home/adpi2/hello/target/scala-0.27/test-meta
```
`show <project> / Test / scalacOptions` prints 
```
[info] * -bootclasspath
[...]

PR #6296, opened 2021-02-06
better support of custom configurations in SemanticdbPlugin

bjaglin:
Second take at https://github.com/sbt/sbt/issues/5886, which fixed the problem for `Test` specifically but not for custom configurations.

* Any child of `Compile` (like `Custom` in the scripted) had to use `testSettings`, whether they were related to testing or not
* Custom configurations with grand parents (like `SystemTest` in the scripted) would get duplicated `scalacOptions` no matter what they used

### DIFF
--- a/sbt/src/sbt-test/project/semanticdb/build.sbt
+++ b/sbt/src/sbt-test/project/semanticdb/build.sbt
@@ -2,4 +2,29 @@ ThisBuild / scalaVersion := "2.12.12"
 ThisBuild / semanticdbEnabled := true
 ThisBuild / semanticdbIncludeInJar := true
 
+// see https://github.com/sbt/sbt/issues/5886
+lazy val check = taskKey[Unit]("Checks that scalacOptions have the same number of parameters across configurations")
+lazy val anyConfigInThisProject = ScopeFilter(configurations = inAnyConfiguration)
+
+lazy val Custom = config("custom").extend(Compile)
+lazy val SystemTest = config("st").extend(IntegrationTest)
+
 lazy val root = (project in file("."))
+  .configs(IntegrationTest, Custom, SystemTest)
+  .settings(
+    inConfig(IntegrationTest)(Defaults.testSettings ++ sbt.plugins.SemanticdbPlugin.configurationSettings),
+    inConfig(Custom)(Defaults.configSettings ++ sbt.plugins.SemanticdbPlugin.configurationSettings),
+    inConfig(SystemTest)(Defaults.testSettings ++ sbt.plugins.SemanticdbPlugin.configurationSettings),
+    check := {
+      val scalacOptionsCountsAcrossConfigs = scalacOptions.?.all(anyConfigInThisProject)
+        .value
+        .map(_.toSeq.flatten.size)
+        .filterNot(_ == 0)
+        .distinct
+      assert(
+        scalacOptionsCountsAcrossConfigs.size == 1,
+        s"Configurations expected to have the same number of scalacOptions but found different numbers: $scalacOptionsCountsAcrossConfigs"
+      )
+    }
+
+  )
--- a/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala
+++ b/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala
@@ -13,10 +13,10 @@ import java.io.File
 import Keys._
 import sbt.internal.SysProp
 import sbt.librarymanagement.syntax._
-import sbt.librarymanagement.CrossVersion
+import sbt.librarymanagement.{ Configuration, CrossVersion }
 import Project.inConfig
 import sbt.internal.inc.ScalaInstance
-import sbt.SlashSyntax0._
+import sbt.ScopeFilter.Make._
 
 object SemanticdbPlugin extends AutoPlugin {
   override def requires = JvmPlugin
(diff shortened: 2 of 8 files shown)


sbt/src/sbt-test/project/semanticdb/build.sbt
```diff
@@ -2,4 +2,29 @@ ThisBuild / scalaVersion := "2.12.12"
...
+
+lazy val Custom = config("custom").extend(Compile)
+lazy val SystemTest = config("st").extend(IntegrationTest)
+
 lazy val root = (project in file("."))
+  .configs(IntegrationTest, Custom, SystemTest)
+  .settings(
+    inConfig(IntegrationTest)(Defaults.testSettings ++ sbt.plugins.SemanticdbPlugin.configurationSettings),
+    inConfig(Custom)(Defaults.configSettings ++ sbt.plugins.SemanticdbPlugin.configurationSettings),
+    inConfig(SystemTest)(Defaults.testSettings ++ sbt.plugins.SemanticdbPlugin.configurationSettings),
```
bjaglin, line 17: This goes beyond the scope of this PR, but I wonder how/where the fact that custom configurations need to source `sbt.plugins.SemanticdbPlugin.configurationSettings` should be documented.

bjaglin: It's worth noting that forgetting to source `sbt.plugins.SemanticdbPlugin.configurationSettings` for custom configuration would work but semanticdb files would end up in a `meta` or `classes` directory that belongs to their ancestor configuration (as `scalacOptions` inherit from that ancestor). It's by using remote caching (with `semanticdbIncludeInJar := true`) that we discovered this sneaky behavior causing invalidations and race conditions.

eed3si9n: > This goes beyond the scope of this PR, but I wonder how/where the fact that custom configurations need to source `sbt.plugins.SemanticdbPlugin.configurationSettings` should be documented.

It's a good question. Generally I don't recommend creating custom configurations, and over time the documentation of how to do so has been dropping through the cracks. I can make a centralized page for it on the docs.

eed3si9n: https://github.com/sbt/website/pull/985

eed3si9n (approved):
Thanks @bjaglin!

sbt/src/sbt-test/project/semanticdb/build.sbt
bjaglin: I created https://github.com/bjaglin/sbt/commit/a6766924ec6e310f00acc99adad09f285b0d37b1 following our discussion over there, but I am hesitant to add it to this PR
1. current users sourcing `sbt.plugins.SemanticdbPlugin.configurationSettings` would now get duplicated values as `configurationSettings` appends - I guess this is an acceptable breaking change as it didn't work well so the few users that did that will probably figure it out
1. It would no longer be possible to opt-out from `SemanticdbPlugin` via `disablePlugins`: config-level keys would remain but not project-level ones, causing (I think) initialization failure as `semanticdbCompilerPlugin` is not defined anywhere - I guess the rationale for exposing the semanticdb integration as a plugin rather than built-in keys was that users could easily opt-out?

eed3si9n: That is a pickle isn't it. I don't have a good answer either.

bjaglin: Backward-compatibility aside, what about moving from opt-out to opt-in (from `allRequirements` to `noTrigger`) for this and maybe as a guideline for any upcoming built-in plugin? This way, users are aware that it's a plugin and that config-level settings must be sourced explicitly in any custom configuration (including `IntegrationTest`).

(I am also advocating for that as I ljust added `disablePlugins(DependencyTreePlugin)` on all but the root module in a large project where I was just missing `dependencyList` from `MiniDependencyTreePlugin`).

eed3si9n: If it's truly opt-in, then it would live outside of sbt/sbt, so the fact that it made it here would likely mean that I or someone driving sbt wants to promote it for whatever reason. Things like SemanticDB and dependency tree are especially tricky because it's nice to be able to have them out of box, so you can ask ppl to just run the task to debug and/or open VS Code, but they are certainly not free (of performance penalties etc).

bjaglin: > If it's truly opt-in, then it would live outside of sbt/sbt

Isn't there a case for advanced functionality using internal sbt APIs, that would still be exposed as plugin keys for UX?

> it's nice to be able to have them out of box, so you can ask ppl to just run the task to debug and/or open VS Code

Good point. In the case of `SemanticdbPlugin`, I was about to say that assigning a setting key to `true` was not too far off from enabling a plugin for each project, but I missed that the `semanticdbEnabled`'s default was global and now [assignable via a CLI propery](https://github.com/sbt/sbt/commit/8789d2072453387dadedfbd29cc93f105860ba3e). We would need easy, centralized plugin activation across projects & the ability to activate programmatically to match that UX.

eed3si9n: I'm going to merge this PR as-is for now.

PR #6305, opened 2021-02-12
Fix SBT Warn on slow HostName

etspaceman:
Turns out the snippet provided in the warning was incorrect and did not resolve the issue properly. Adding `.local` to the end of the HostName will resolve the slowness.

See the 2nd answer here: 

https://apple.stackexchange.com/questions/175320/why-is-my-hostname-resolution-taking-so-long


eed3si9n (approved):
Thanks @etspaceman!

PR #3434, opened 2017-08-14
Unify sbt shell and build.sbt syntax (scope path syntax)

eed3si9n:
Fixes sbt/sbt#1812

This adds scope path syntax (fka unified slash syntax) for both sbt shell and the build.sbt DSL.

### Scope path syntax in shell

Instead of the current `<project-id>/config:intask::key`, this **adds** `<project-id>/<config-ident>/intask/key` where `<config-ident>` is the Scala identifier notation for the configurations like `Compile` and `Test`. (The old shell syntax will continue to function)

### Scope path syntax in build.sbt

This also adds a series of implicits called `SlashSyntax` that adds `/` operators to project references, configuration, and keys such that the same syntax works in build.sbt.

### Examples

These examples work *both* from the shell *and* in build.sbt.

```scala
Global / cancelable
ThisBuild / scalaVersion
Test / test
root / Compile / compile / scalacOptions
ProjectRef(uri("file:/xxx/helloworld/"),"root")/Compile/scalacOptions
Zero / Zero / name
```

The inspect command now outputs something that can be copy-pasted:

    > inspect compile
    [info] Task: sbt.inc.Analysis
    [info] Description:
    [info] 	Compiles sources.
    [info] Provided by:
    [info] 	ProjectRef(uri("file:/xxx/helloworld/"), "root") / Compile / compile
    [info] Defined at:
    [info] 	(sbt.Defaults) Defaults.scala:326
    [info] Dependencies:
    [info] 	Compile / manipulateBytecode
    [info] 	Compile / incCompileSetup
    [info] Reverse dependencies:
    [info] 	Compile / printWarnings
    [info] 	Compile / products
    [info] 	Compile / discoveredSbtPlugins
    [info] 	Compile / discoveredMainClasses
    [info] Delegates:
    [info] 	Compile / compile
    [info] 	compile
    [info] 	ThisBuild / Compile / compile
    [info] 	ThisBuild / compile
    [info] 	Zero / Compile / compile
    [info] 	Global / compile
    [info] Related:
    [info] 	Test / compile

slash

PR #6309, opened 2021-02-15
Deprecate old shell syntax

eed3si9n:
Unified slash syntax was added in sbt 1.1.0 in 2017 (https://github.com/sbt/sbt/pull/3434).
This deprecates the older shell syntax: `proj/cofing:intask::key` that's been around since sbt 0.9.

(sbt 0.7 had a key for `test-compile`)

### Example

```
sbt:hello> it:compile
[warn] sbt 0.13 shell syntax is deprecated; use slash syntax instead: IntegrationTest / compile
[info] compiling 1 Scala source to /private/tmp/hello/target/scala-2.13/classes ...

sbt:hello> compile:console::scalacOptions
[warn] sbt 0.13 shell syntax is deprecated; use slash syntax instead: Compile / console / scalacOptions
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1029,6 +1029,9 @@ lazy val mainProj = (project in file("main"))
       exclude[MissingClassProblem](
         "sbt.internal.XMainConfiguration$ModifiedConfiguration$ModifiedAppProvider$ModifiedScalaProvider$"
       ),
+      // internal impl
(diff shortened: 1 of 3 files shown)


SethTisue:
fwiw, I would also like to see `scalacOptions in (Compile, console)` deprecated. slash syntax all the things!!

PR #6316, opened 2021-02-17
Fix "-Yrangepos" disappearing from Test / scalacOptions

### LINKED ISSUE #5934: "-Yrangepos" disappears from Test / scalacOptions when semanticdbEnabled is not `true` (was munit tests fail after upgrade to 1.4.0)
## steps

```scala
ThisBuild / scalaVersion := "2.12.13"

  .settings(
    name := "hello",
    Compile / scalacOptions += "-Yrangepos",
  )
```

Type the following in shell:

```
Test/scalacOptions
```

## problem

```
sbt:hello> show Test/scalacOptions
[info] *
```

## expectation

```
sbt:hello> show Test/scalacOptions
[info] * -Yrangepos
```

----

## steps

https://github.com/scalameta/munit/pull/213 upgrades to sbt 1.4.0.

## problem

https://github.com/scalameta/munit/runs/1207061695

```
==> X munit.AssertionsSuite.basic  0.018s munit.FailException: /home/runner/work/munit/munit/tests/shared/src/test/scala/munit/AssertionsSuite.scala:19
18:  val c: List[Int] = List(41)
19:  check(
20:    "basic",
diff assertion failed
=> Obtained
    """|Clues {
       |  : Int = 42
       |  : Long = 43
       |}
       |""".stripMargin
=> Diff (- obtained, + expected)
 Clues {
-  : Int = 42
-  : Long = 43
+  a: Int = 42
+  b: Long = 43
 }
    at munit.AssertionsSuite$$anonfun$check$1.apply$mcV$sp(AssertionsSuite.scala:13)
    at munit.AssertionsSuite$$anonfun$check$1.apply(AssertionsSuite.scala:11)
    at munit.AssertionsSuite$$anonfun$check$1.apply(AssertionsSuite.scala:11)
```

## expectation

It works as before.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/5934

In #5886 adpi2 reported that Test / scalacOptions could have unwanted duplicated flags from Compile / scalacOptions. So #5887 added

```diff
+    // remove duplicated semanticdbOptions
+    scalacOptions --= (Compile / semanticdbOptions).value
```

Notice that it's subtracting (Compile / semanticdbOptions).value regardless of the semanticdbEnabled value. If semanticdbEnabled is set to true I am guessing that it would all work out, but when it's false, it's just subtracting "-Yrangepos".

This fixes that by checking for semanticdbEnabled.


SethTisue:
any time `scalacOptions` (or `javacOptions`) is treated as something like a set, I squirm and fret a bit about things like `-source 1.8 -target 1.8`, which is sort of four options and sort of two, and the two "1.8"s aren't the "same" option.

but I guess that doesn't come into play here.......?

adpi2:
I think you are right @SethTisue, we should avoid things like `--=` and `.distinct` on `scalacOptions` as much as possible.

In this particular case, the problem we are facing is that `Test / scalacOptions` inherits from `Compile / scalacOptions`, which is wanted most of the time, but not always. For instance we want a different `-P:semanticdb:targetroot:xxx` on Test and Compile.

However I've just found that we can add `semanticdbOptions` to `Compile / compile / scalacOptions` to avoid the problem:

```scala
sbt:root> set Compile / compile  / scalacOptions ++= Seq("-Xfatal-warnings")
...
sbt:root> show Test / compile / scalacOptions
[info] * 
```

This way, we don't need to call `--=`. 

@eed3si9n what do you think? Should I push this change?

adpi2 (approved):
I am good with this change. But maybe this can be improved by using the `Compile / compile` scope more extensively as describe [here](https://github.com/sbt/sbt/pull/6316#issuecomment-780466043)?

eed3si9n:
Sounds good. Let's merge as-is, and feel free to send a PR on top.

eed3si9n:
@SethTisue Yea. If we can get rid of subtracting I think it would be better, but certain aspect of scalacOptions _should_ behave more like set in a sense that we don't want duplicates.

adpi2:
In order to use `scalacOptions` as a set we should make possible and encourage `Seq("-source 1.8", "-target 1.8")` instead of `Seq("-source", "1.8", "-target", "1.8")`. In other words, scalacOptions should be a Seq of atomic options that can be merged and substracted.

It may not be a Set though, because the order is meaningful: `Seq("-source 1.8", "-source 1.9")` is not the same thing as `Seq("-source 1.9", "-source 1.8")`.

PR #6336, opened 2021-02-24
Fix TastyFiles added twice in Test

adpi2:
```
Compile / auxiliaryClassFiles += TastyFiles.instance
Test / auxiliaryClassFiles += TastyFiles.instance
```
results in `TastyFiles` added twice in the `Test` config


eed3si9n (approved):
Thanks!

PR #6351, opened 2021-03-03
Publish all plugin dependencies

steinybot:
If the plugin depends on another project then scripted tests fail as the scriptedDependencies only publishes the plugin. This publishes the plugin and its dependencies.

I couldn't get the test to work. Maybe that is why the others are disabled?


eed3si9n (approved):
Thanks!

dwijnand:
So clean, nice!

PR #6352, opened 2021-03-04
Fix sourcePositionMappers added by Play not getting called

mkurz:
Setting a default for `sourcePositionMappers` (done in #5857) is a very, _very_ bad idea.
Such a default sourcePositionMapper acts as a "catch all" when "resolving" the sourcePositionMappers setting, therefore sourcePositionMappers that were added by Applications or Framworks (like Play) will _never_ have the chance to get called.
Concrete example: A typical Play application adds following two sourcePositionMappers:
- The twirl one: https://github.com/playframework/twirl/blob/1.5.0/sbt-twirl/src/main/scala/play/twirl/sbt/SbtTwirl.scala#L72
  - Used to map a generated twirl template in the target folder (like `*.template.scala`) to its origin twirl template (`*.scala.html` files)
- The routes compiler one: https://github.com/playframework/playframework/blob/2.8.7/dev-mode/sbt-plugin/src/main/scala/play/sbt/routes/RoutesCompiler.scala#L138
   - Used to map a generated routes file in the target folder (like `Routes.scala`) to its origin routes file (usually `conf/routes`)

We end up with following list of sourcePositionMappers, in this order:
`(the default one)` -> `(the twirl one)` -> `(the routes one)`

Usually if a sourcePositionMapper can not map a file it will return `None` and thanks to the `foldMappers` method that build that chain the next mapper will be tried. However the default sourcePositionMapper will never ever continue to the other mappers, because it will always return a position (see the `toAbsoluteSourceMapper` it will always return `Some`) (Also because `reportAbsolutePath` is true by default...)

The solution to this problem is to _never_ add a default sourcePositionMapper, but instead it should fall back to the default one in case no `sourcePositionMappers` was set at all by an application *and* to pass the modified position with the absolute source to the existing mappers:
`(the twirl one with absolute source passed)` -> `(the routes one with absolute source passed)` -> `(the default one)`

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -422,12 +422,7 @@ object Defaults extends BuildCommon {
       )
     },
     fileConverter := MappedFileConverter(rootPaths.value, allowMachinePath.value),
-    sourcePositionMappers ++= {
-      val fc = fileConverter.value
-      if (reportAbsolutePath.value) {
-        List(toAbsoluteSourceMapper(fc) _)
-      } else Nil
-    },
+    sourcePositionMappers := Nil, // Never set a default sourcePositionMapper, see #6352! Whatever you are trying to solve, do it in the foldMappers method.
     // The virtual file value cache needs to be global or sbt will run out of direct byte buffer memory.
     classpathDefinesClassCache := VirtualFileValueCache.definesClassCache(fileConverter.value),
     fullServerHandlers := {
@@ -474,7 +469,7 @@ object Defaults extends BuildCommon {
     },
   )
 
-  private[sbt] def toAbsoluteSourceMapper(fc: FileConverter)(pos: Position): Option[Position] = {
+  private[sbt] def toAbsoluteSource(fc: FileConverter)(pos: Position): Position = {
     def isValid(path: String): Boolean = {
       Try(Paths.get(path)).map(_ => true).getOrElse(false)
     }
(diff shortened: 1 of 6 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -503,9 +498,21 @@ object Defaults extends BuildCommon {
...
+
+          override def endOffset(): Optional[Integer] = pos.endOffset()
+
+          override def startLine(): Optional[Integer] = pos.startLine()
+
+          override def startColumn(): Optional[Integer] = pos.startColumn()
+
+          override def endLine(): Optional[Integer] = pos.endLine()
+
+          override def endColumn(): Optional[Integer] = pos.endColumn()
```
mkurz, line 512: The methods were forgotten, the are defined as `default`,

mkurz:
The `reporter/source-mapper` tests fails now because there is are no `sourcePositionMappers` anymore that can be `foldLeft`:
https://github.com/sbt/sbt/blob/v1.4.7/sbt/src/sbt-test/reporter/source-mapper/build.sbt#L12-L15

I will try to figure out an alternative test...

EDIT: Probably I will use the `printWarnings` task and look in the logs if absolute paths were used.

eed3si9n:
@mkurz Thanks for this contribution 🙏

mkurz:
@eed3si9n Done. I modified the existing test to read the log. Actually I think the test reflects what the original issue #5853 was about much better.
I squashed everything into one commit and also added a backport for 1.4.x in #6356.
Actually... how long does it take to release a new version from the 1.4.x branch... :stuck_out_tongue_winking_eye:?
Looking forward to your feedback.

eed3si9n (approved):

eed3si9n:
> Actually... how long does it take to release a new version from the 1.4.x branch... 😜?

I can get one out this weekend.

mkurz:
@eed3si9n 

> I can get one out this weekend.

Thanks! That would be really great! [I am working on sbt 1.4 support for Play](https://github.com/playframework/playframework/pull/10655) and this patch here is needed to make it work.
(I could make a dirty workaround in Play, however it would be better if sbt would just behave correctly)

mkurz:
@eed3si9n Actually there is one small thing that needs to be patched, we need to make sure that the sourcePositionMappers also get the absolute file path (not just the source String). PR on its way.

mkurz:
The patch I was talking about: #6358

Issue #6144, opened 2020-11-20
problem using scala 2.13.4 when Global / semanticdbEnabled := true


pjfanning:
## steps

* SBT 1.4.3
* tried to change https://github.com/pjfanning/swagger-akka-http-sample/blob/master/build.sbt to use scala version 2.13.4
* I also have

```
~/.sbt/1.0/global.sbt 

Global / semanticdbEnabled := true
```

## problem

Get an issue fetching a jar:

[warn] 	Note: Unresolved dependencies path:
[error] sbt.librarymanagement.ResolveException: Error downloading org.scalameta:semanticdb-scalac_2.13.4:4.3.20

## expectation

org.scalameta:semanticdb-scalac_2.13.4:4.4.0 is available - would it be possible to do an sbt release that upgrades the semanticdb-scalac dependency?

## notes

I probably don't need to set semanticdbEnabled -- https://github.com/sbt/sbt/issues/5977

adpi2:
Hi @pjfanning! Thank you for reporting.

It is not really an sbt bug.

The issue is rather that the version `4.3.20` of the `semantic-scalac` plugin is not available for the very new Scala `2.13.4`. But the version `4.4.0` is available: https://mvnrepository.com/artifact/org.scalameta/semanticdb-scalac_2.13.4.

So you just need to set
```
Global / semanticdbVersion := "4.4.0"
```

PR #6354, opened 2021-03-04
Add system property to override semanticdb version

marcospereira:
While it is possible to change the version using an sbt setting, having a sysprop helps when enabling semanticdb for some projects using a Scala a version that is not supported by the default version defined in sbt.

For example, users will be able to do something like:

    sbt -Dsbt.semanticdb=true -Dsbt.semanticdb.version=4.4.10

#6144 is an example where this fixed would be used without requiring any changes to the project.

### DIFF
--- a/main/src/main/scala/sbt/internal/SysProp.scala
+++ b/main/src/main/scala/sbt/internal/SysProp.scala
@@ -93,8 +93,10 @@ object SysProp {
   def ci: Boolean = getOrFalse("sbt.ci")
   def allowRootDir: Boolean = getOrFalse("sbt.rootdir")
   def legacyTestReport: Boolean = getOrFalse("sbt.testing.legacyreport")
-  def semanticdb: Boolean = getOrFalse("sbt.semanticdb")
   def forceServerStart: Boolean = getOrFalse("sbt.server.forcestart")
+
+  def semanticdb: Boolean = getOrFalse("sbt.semanticdb")
+  def semanticdbVersion: Option[String] = sys.props.get("sbt.semanticdb.version")
 
   def watchMode: String =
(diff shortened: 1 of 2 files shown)


eed3si9n:
Could you elaborate the user scenario you have in mind for this please? Like someone doesn't want to edit `build.sbt` but wants to run Metals?

I occasionally introduce sys prop as a feature flag, but I am not sure if `semanticdbVersion` is a good use of a sys prop. Unlike having the SemanticDB turned on or off, this is effectively fixing the old `semanticdbVersion` baked into sbt. So when someone else calls `sbt -Dsbt.semanticdb=true` it wouldn't work.

marcospereira:
> Could you elaborate the user scenario you have in mind for this please? Like someone doesn't want to edit `build.sbt` but wants to run Metals?

Yes, I've run into this while trying to load the project using Metals, but the Scala version the project is using is not supported by the default semanticdb version.

> I occasionally introduce sys prop as a feature flag, but I am not sure if `semanticdbVersion` is a good use of a sys prop.

I thought about that too but saw that are other places where sys props are not used as feature flags but feature configurations. Anyway, if you are trying to move away from that, there are other workarounds I can use.

ckipp01:
> Yes, I've run into this while trying to load the project using Metals, but the Scala version the project is using is not supported by the default semanticdb version.

Drive by comment. From the Metals side, is this scenario when you are using a newer version than we support at the moment? For the vast majority of the time, Metals supports the full range of Scala versions that semanticdb supports. The only exception really is when a new version of Scala is released, and there may be a few days before we can release. However, even in that scenario, just bumping the semanticdb version to a newer one won't give you support in Metals. There may be other issues you run into. Do you have a specific example with Metals when this has been a problem?

PR #6353, opened 2021-03-04
Give an explanation in case of ServerAlreadyBootingException

sideeffffect:
reopening of https://github.com/sbt/sbt/pull/6104

PR #6357, opened 2021-03-05
[1.4.x] Give an explanation in case of ServerAlreadyBootingException

eed3si9n:
This is a backport of #6353 by @sideeffffect 
Ref https://github.com/sbt/sbt/issues/6101

Give and explanation in case of ServerAlreadyBootingException

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -119,9 +119,11 @@ private[sbt] object xMain {
   ): (Option[BootServerSocket], Option[Exit]) =
     try (Some(new BootServerSocket(configuration)) -> None)
     catch {
-      case _: ServerAlreadyBootingException
+      case e: ServerAlreadyBootingException
           if System.console != null && !ITerminal.startedByRemoteClient =>
-        println("sbt server is already booting. Create a new server? y/n (default y)")
+        println(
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/Main.scala
```diff
@@ -119,9 +119,11 @@ private[sbt] object xMain {
   ): (Option[BootServerSocket], Option[Exit]) =
     try (Some(new BootServerSocket(configuration)) -> None)
     catch {
-      case _: ServerAlreadyBootingException
+      case e: ServerAlreadyBootingException
           if System.console != null && !ITerminal.startedByRemoteClient =>
-        println("sbt server is already booting. Create a new server? y/n (default y)")
+        println(
+          s"sbt thinks that server is already booting because of this exception:\n${e.getCause}\nCreate a new server? y/n (default y)"
```
ScalaWilliam, line 125: Getting the e.getCause.getMessage (assuming neither of getCause and .getMessage return null), might provide a bit more detail to help identify the exact issue (for example, access denied to what?)

sideeffffect: I think `toString` gives us this info
https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html#toString()

PR #6373, opened 2021-03-08
Don't use deprecated `in` inside of macros

### LINKED ISSUE #6330: [sbt 1.5.0-M2] Can't use `inputFileChanges` nor `inputFiles` without warnings
## steps

Have a source code generator using either `inputFileChanges` and/or `inputFiles`.

## problem

```
[error] /path/to/project/CodeGenerator.scala:37:33: method in in class Scope is deprecated (since 1.5.0): `in` is deprecated; migrate to slash syntax - https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#slash
[error]         (config / genCode).inputFiles,
[error]                            ^
[error] /path/to/project/CodeGenerator.scala:36:33: method in in class Scope is deprecated (since 1.5.0): `in` is deprecated; migrate to slash syntax - https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#slash
[error]         (config / genCode).inputFileChanges,
[error]                            ^
```

## expectation

Either no warning (treated as errors in my case), or a warning telling what to do instead.

## notes

https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#slash doesn't give any hint on how to replace `inputFiles` and `inputFileChanges` macros with the slash syntax.

eatkins:
The inputFileChanges, inputFiles and outputFiles macros all used the now
deprecated `in` syntax, which causes warnings for builds that use these
apis. SlashSyntax doesn't naively work in these macros so it was easier
to just reimplement the logic in this file.

Fixes #6330

### DIFF
--- a/main/src/main/scala/sbt/internal/FileChangesMacro.scala
+++ b/main/src/main/scala/sbt/internal/FileChangesMacro.scala
@@ -58,6 +58,10 @@ object FileChangesMacro {
       c.universe.reify(outputFileStamps)
     )
   }
+  def rescope[T](left: TaskKey[_], right: TaskKey[T]): TaskKey[T] =
+    Scoped.scopedTask(left.scope.copy(task = Select(left.key)), right.key)
+  def rescope[T](left: Scope, right: TaskKey[T]): TaskKey[T] =
+    Scoped.scopedTask(left, right.key)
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks @eatkins!

PR #6375, opened 2021-03-09
Add support for custom protocol with coursier

MasseGuillaume:
In coursier, you can handle custom protocols by defining a protocol handler:
https://get-coursier.io/docs/2.0.10/extra.html#extra-protocols
https://github.com/coursier/coursier/blob/ce251d0288bf029e4eeb8700e8990178559a005d/modules/tests/jvm/src/test/scala/coursier/cache/protocol/TestprotocolHandler.scala

I'm currently implementing support for Google Artifact Resitory:
https://github.com/coursier/coursier/issues/1987#issuecomment-792894989

When dealing with artifact with protocol `artifactregistry` I need to have a plugin available in the classpath. Since lmcoursier is in the sbt launcher classpath, it's not available by adding `libraryDependencies += ...` in `project/plugins.sbt` like you would do for an sbt plugin. Therefore, I propose a way to explicitly declare the artifacts you need to handle the protocol handling.

Users of custom protocol can now specify dependencies to their handler via this sbt setting:

```scala
csrProtocolHandlerDependencies += "org.example" % "google-artifact-repository-coursier" % "0.1.0"
```

Depends on https://github.com/coursier/sbt-coursier/pull/327

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -271,6 +271,7 @@ object Defaults extends BuildCommon {
       csrLogger := LMCoursier.coursierLoggerTask.value,
       csrMavenProfiles :== Set.empty,
       csrReconciliations :== LMCoursier.relaxedForAllModules,
+      csrProtocolHandlerDependencies :== Nil,
     )
 
   /** Core non-plugin settings for sbt builds.  These *must* be on every build or the sbt engine will fail to run at all. */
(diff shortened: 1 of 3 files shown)


987Nabil:
Is there anything that speaks against this? I'd really like to have this 🙂

sideeffffect:
What is blocking this? It would be of great help for us.

laughedelic:
⬆️ 🙏

eed3si9n:
I think this PR should be on top of 1.9.x branch, since develop branch is for sbt 2.x.

rolang:
It seems to be possible to set it via `csrConfiguration` as well? Adding it like that worked for us:
```scala
csrConfiguration := csrConfiguration.value.withProtocolHandlerDependencies(
  Seq("org.example" % "google-artifact-repository-coursier" % "0.1.0")
)
```

PR #6397, opened 2021-03-15
Environment variables support in BSP debug session

### LINKED ISSUE #6396: BSP debug session should support Environment Variables
## steps

1. Add environment variables in `launch.json` in VSCode with Metals.
2. Launch build configuration.

## problem

Environment variables are not available in the started application.

## expectation

Environment variables are available in the started application.

## notes
This works with latest Bloop. Seems it is very recent addition to the protocol and hasn't been implemented in SBT yet.

arixmkii:
Fixes #6396 
Very basic parser for `"key=value"` strings, but it is how metals generate them https://github.com/scalameta/metals/blob/3ada84f575f4a7ca82f4eefad32d2e21d03b349a/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala#L188

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
@@ -521,7 +521,8 @@ object BuildServerProtocol {
             )
           ),
           runParams.arguments,
-          defaultJvmOptions.toVector
+          defaultJvmOptions.toVector,
+          Vector.empty
         )
     }
 
@@ -587,7 +588,12 @@ object BuildServerProtocol {
       workingDirectory = Some(baseDirectory.value),
       runJVMOptions = mainClass.jvmOptions,
       connectInput = connectInput.value,
-      envVars = envVars.value
+      envVars = envVars.value ++ mainClass.environmentVariables
+        .flatMap(_.split("=", 2).toList match {
+          case key :: value :: Nil => Some(key -> value)
+          case _                   => None
+        })
+        .toMap
     )
     val runner = new ForkRun(forkOpts)
     val statusCode = runner
@@ -649,7 +655,7 @@ object BuildServerProtocol {
   private def scalaMainClassesTask: Initialize[Task[ScalaMainClassesItem]] = Def.task {
     val jvmOptions = Keys.javaOptions.value.toVector
     val mainClasses = Keys.discoveredMainClasses.value.map(
-      ScalaMainClass(_, Vector(), jvmOptions)
+      ScalaMainClass(_, Vector(), jvmOptions, Vector.empty)
     )
     ScalaMainClassesItem(
       bspTargetIdentifier.value,
--- a/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
@@ -503,7 +503,10 @@ object BuildServerProtocol {
               ErrorCodes.ParseError,
               e.getMessage
             )
-          case Success(value) => value
+          case Success(value) =>
+            value.withEnvironmentVariables(
+              envVars.value.map { case (k, v) => s"$k=$v" }.toVector ++ value.environmentVariables
+            )
         }
 
       case Some(dataKind) =>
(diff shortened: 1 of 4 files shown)


arixmkii:
`ScalaMainClass` contraband protocol update matches what is defined in bsp4j https://github.com/build-server-protocol/build-server-protocol/blob/8ed2c5126bede46b0908ba33076dedd761ebbeb0/bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/ScalaMainClass.java#L22

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -521,7 +521,8 @@ object BuildServerProtocol {
             )
           ),
           runParams.arguments,
-          defaultJvmOptions.toVector
+          defaultJvmOptions.toVector,
+          Vector.empty
```
adpi2: We should use the `envVars.value` as a default. The idea is that, if no `scalaMainClass` is provided, we run the project as it is configured in the `build.sbt`.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -649,7 +655,7 @@ object BuildServerProtocol {
   private def scalaMainClassesTask: Initialize[Task[ScalaMainClassesItem]] = Def.task {
     val jvmOptions = Keys.javaOptions.value.toVector
     val mainClasses = Keys.discoveredMainClasses.value.map(
-      ScalaMainClass(_, Vector(), jvmOptions)
+      ScalaMainClass(_, Vector(), jvmOptions, Vector.empty)
```
adpi2: We should probably fill the `environmentVariables` field with `envVars.value` here as well.

adpi2 (commented):
Thanks for the contribution!

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
arixmkii: Yes, I had the same feeling first, but then decided to give it a different meaning. As the origin of `ScalaMainClass` data is external to sbt process. (metals or something else) and the data is restored from serialized object normally - the client will not be aware what settings are actual inside sbt. So, I treat the data incoming in this place as "additional" environment variables.

[Line 591.](https://github.com/sbt/sbt/pull/6397/files#diff-2ad23bc2de451984eb3983aceb2f15262cd8748b845faebae937a4d8f762f244R591)

But this assumption of mine could be wrong, because I didn't spend enough time with the codebase.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -587,7 +588,12 @@ object BuildServerProtocol {
       workingDirectory = Some(baseDirectory.value),
       runJVMOptions = mainClass.jvmOptions,
       connectInput = connectInput.value,
-      envVars = envVars.value
+      envVars = envVars.value ++ mainClass.environmentVariables
```
arixmkii: Here we concatenate the envVars of sbt process and ones submitted by client. So, my guess is that we don't need to put the envVars of sbt process in the `ScalaMainClass` data structure.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
adpi2: I agree with you that the incoming environment variables should be treated as additionals. This is not how the `jvmOptions` are treated but maybe it was a mistake.

However, this method (`scalaMainClassesTask`) is the implementation of the `buildTarget/scalaMainClasses` request sent by the client to the server (sbt) to learn about the project main classes and how they are configured. So it is probably safe and better to not hide the `envVars`, even if the client do not use them later.

adpi2: For instance, the client might want to run the `scalaMainClass` by itself (not through sbt), in such case it would like to know about the `envVars` .

arixmkii: Read the code once again. Now I understand what you suggested. Thank you for the explanation @adpi2!

I added commit, where I moved concatenation to parts, where we are instantiating `ScalaMainClass` - either recovering serialized version or creating with default constructor. And removed unnecessary concatenation in the moment, when the forked process is invoked.

adpi2 (approved):

PR #6408, opened 2021-03-22
Use %LOCALAPPDATA% or %HOME%/AppData/Local on Windows

### LINKED ISSUE #5206: On Windows, Coursier creates "./null/Coursier/cache/v1" directories as cache
See also https://github.com/coursier/coursier/issues/1438

## steps
sbt version: 1.3.3

1. Use Windows.
2. Download the AdoptOpenJDK 8 232-b09 zip file (https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u232-b09/OpenJDK8U-jdk_x86-32_windows_hotspot_8u232b09.zip)
3. Launch sbt as follows:

```
sbt --java-home /cygdrive/c/Users/foo/Downloads/OpenJDK8U-jdk_x86-32_windows_hotspot_8u232b09/jdk8u232-b09/
```

## problem
Coursier location was wrongly set to the project location, with a null folder, not in the %LOCALAPPDATA%\Coursier\Cache\v1.

My system is newly reinstalled win10 1909. I think maybe it's the new windows10 issue & sbt didn't get the right %LOCALAPPDATA%.

批注 2019-10-20 141902

Both tested in the IDEA and PowerShell with downloaded sbt 1.3.3.

## expectation
Coursier located in %LOCALAPPDATA%\Coursier\Cache\v1

## notes
By setting the COURSIER_CACHE enviroment varible, the Coursier location works correctly.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/5206
Ref https://github.com/dirs-dev/directories-jvm/issues/49

Problem
--------
Coursier uses [directories-jvm](https://github.com/dirs-dev/directories-jvm) to determine its default cache directory.

On Windows, currently directories-jvm shells out to Powershell to call the [Known Folders API](https://docs.microsoft.com/en-us/windows/win32/shell/knownfolderid), which doesn't work in various environments, and instead of an error, it apparently returns `null/Coursier/cache` as the directory name.

Solution
--------
With due respect to the heroic effort directories-jvm is making to comply to the directory standards on all operating systems, including that of Microsoft, I don't think the majority of the sbt users care exactly where that directory is as long as it is well-documented, and calculated in a fast and predictable way. And it doesn't fall back to `./null/Coursier/cache`.

Instead of shelling out to Powershell, or using JNI, to hit the Known Folders API, I propose we first look at `LOCALAPPDATA` environment variable. When it is not found, it will fall back to `$HOME/AppData/Local`.

Per discussion in https://github.com/dirs-dev/directories-jvm/issues/43, `LOCALAPPDATA` environment variable may _NOT_ represent the one-true Known Folders API value of the AppData directory in case the user happened to have set the `LOCALAPPDATA` environmental variable. For the purpose of picking a directory for Coursier cache, I don't find that to be a problem because it will be faster, more reliable, and predictable.

### DIFF
--- a/main/src/main/scala/sbt/coursierint/LMCoursier.scala
+++ b/main/src/main/scala/sbt/coursierint/LMCoursier.scala
@@ -22,6 +22,7 @@ import lmcoursier.definitions.{
 import lmcoursier._
 import lmcoursier.credentials.Credentials
 import Keys._
+import sbt.internal.util.Util
 import sbt.librarymanagement._
 import sbt.librarymanagement.ivy.{
   Credentials => IvyCredentials,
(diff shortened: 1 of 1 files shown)


ScalaWilliam:
Thank you so much for this! Would the same fix need to be done in Coursier eventually as well?

eed3si9n:
Let's see how this workaround pans out.

ohuu:
This doesn't appear to be fixed. I still get `./null/Coursier` on Windows using SBT 1.5.4

eed3si9n:
@PaperPlaneSoftware What's in it? I wonder if it's because I need to copy-paste this to sbt/launcher.

ohuu:
image

This is from a brand new project. I've also tried adding COURSIER_HOME as an env var but no luck

PR #6416, opened 2021-03-27
[1.4.x] Instantiate only test runners needed by current testDefinitions

### LINKED ISSUE #6412: Munit - test filtering doesn't work when ScalaTest is present in the same project
## steps

Having this `build.sbt`

```scala
lazy val root = project
  .in(file("."))
  .settings(
    name := "Munit + ScalaTest",
    libraryDependencies ++= Seq(
      "org.scalatest" %% "scalatest" % "3.2.2"  % Test,
      "org.scalameta" %% "munit"     % "0.7.22" % Test
      )
  )
```

And test file `src/test/scala/example/Spec.scala`
```scala
package example

import munit.FunSuite

class Spec extends FunSuite {
  test("hello") {
    assertEquals(1, 1)
  }
}
```

## problem

When trying to run munit spec with test [filtering](https://scalameta.org/munit/docs/filtering.html)

```
root/testOnly example.Spec -- "--tests=hello"
```

ScalaTest's Runner is throwing and `IllegalArgumentException` and test is never run.

```
[error] (Test / testOnly) java.lang.IllegalArgumentException: Argument unrecognized by ScalaTest's Runner: --tests=hello
```

## expectation

ScalaTest's Runner shouldn't throw an exception and Munit Runner should be able to run the test.

## note

Tested in `1.4.9` and `1.5.0-RC2`

VlachJosef:
For example do not instantiate test runner for ScalaTest if only MUnit test spec is being run.

As it turned out, test runners don't need to be created to determine if test suites are ScalaTest or MUnit etc. Every `TestDefinition` has fingerprint which carries this information. So sbt can decide which test runner to instantiate based on fingerprint of `TestDefinition`s user wants to run.

So rather that creating all test runners up front, this PR adds logic which decides what test runners to create based on `TestDefinition`s which user wants to run.

To get hold of all `TestDefinition`s  user wants to run `Tests.ProcessedOptions` needs to be created. So creation of `Tests.ProcessedOptions` has been moved one layer up, so it can be used to filter test runners. 

Rather than creating `Tests.ProcessedOptions`  in `ForkTests` or `Tests` it is now being passed in as a parameter. This is to avoid doing the same work twice.

Closes #6412

### DIFF
--- a/main-actions/src/main/scala/sbt/Tests.scala
+++ b/main-actions/src/main/scala/sbt/Tests.scala
@@ -325,6 +325,18 @@ object Tests {
     )
   }
 
+  def apply(
+      frameworks: Map[TestFramework, Framework],
+      testLoader: ClassLoader,
+      runners: Map[TestFramework, Runner],
+      discovered: Vector[TestDefinition],
+      config: Execution,
+      log: ManagedLogger
+  ): Task[Output] = {
+    val o = processOptions(config.options, discovered, log)
+    apply(frameworks, testLoader, runners, o, config, log)
+  }
+
   def testTask(
       loader: ClassLoader,
       frameworks: Map[TestFramework, Framework],
--- a/main-actions/src/main/scala/sbt/Tests.scala
+++ b/main-actions/src/main/scala/sbt/Tests.scala
@@ -249,7 +249,7 @@ object Tests {
       val testListeners: Vector[TestReportListener]
   )
   private[sbt] def processOptions(
-      options: Seq[TestOption],
+      config: Execution,
       discovered: Vector[TestDefinition],
       log: Logger
   ): ProcessedOptions = {
@@ -261,7 +261,7 @@ object Tests {
     val testListeners = new ListBuffer[TestReportListener]
     val undefinedFrameworks = new ListBuffer[String]
 
-    for (option <- options) {
+    for (option <- config.options) {
       option match {
         case Filter(include) => testFilters += include; ()
         case Filters(includes) =>
@@ -333,7 +333,7 @@ object Tests {
       config: Execution,
       log: ManagedLogger
   ): Task[Output] = {
-    val o = processOptions(config.options, discovered, log)
+    val o = processOptions(config, discovered, log)
     apply(frameworks, testLoader, runners, o, config, log)
   }
 
--- a/main-actions/src/main/scala/sbt/Tests.scala
+++ b/main-actions/src/main/scala/sbt/Tests.scala
@@ -249,7 +249,7 @@ object Tests {
       val testListeners: Vector[TestReportListener]
   )
   private[sbt] def processOptions(
-      config: Execution,
+      options: Seq[TestOption],
       discovered: Vector[TestDefinition],
       log: Logger
   ): ProcessedOptions = {
@@ -261,7 +261,7 @@ object Tests {
     val testListeners = new ListBuffer[TestReportListener]
     val undefinedFrameworks = new ListBuffer[String]
 
-    for (option <- config.options) {
+    for (option <- options) {
       option match {
         case Filter(include) => testFilters += include; ()
         case Filters(includes) =>
@@ -308,11 +308,10 @@ object Tests {
       frameworks: Map[TestFramework, Framework],
       testLoader: ClassLoader,
       runners: Map[TestFramework, Runner],
-      discovered: Vector[TestDefinition],
+      o: ProcessedOptions,
       config: Execution,
       log: ManagedLogger
   ): Task[Output] = {
-    val o = processOptions(config, discovered, log)
     testTask(
       testLoader,
       frameworks,
--- a/main-actions/src/main/scala/sbt/ForkTests.scala
+++ b/main-actions/src/main/scala/sbt/ForkTests.scala
@@ -23,15 +23,13 @@ import sbt.internal.util.{ RunningProcesses, Terminal }
 private[sbt] object ForkTests {
   def apply(
       runners: Map[TestFramework, Runner],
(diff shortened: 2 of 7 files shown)


eed3si9n:
@VlachJosef Thanks for the contribution!

main-actions/src/main/scala/sbt/Tests.scala
```diff
@@ -308,11 +308,10 @@ object Tests {
       frameworks: Map[TestFramework, Framework],
       testLoader: ClassLoader,
       runners: Map[TestFramework, Runner],
-      discovered: Vector[TestDefinition],
+      o: ProcessedOptions,
```
eed3si9n, line 311: Since `sbt.Tests` is public, we have to keep backward binary compatibility.
Could you add an overload for this?

eed3si9n:
Mima failures:

```
[error]  * static method apply(scala.collection.immutable.Map,scala.collection.immutable.Vector,sbt.Tests#Execution,scala.collection.Seq,sbt.ForkOptions,sbt.util.Logger,sbt.ConcurrentRestrictions#Tag)sbt.Task in class sbt.ForkTests in current version does not have a correspondent with same parameter signature among (scala.collection.immutable.Map,sbt.Tests#ProcessedOptions,sbt.Tests#Execution,scala.collection.Seq,sbt.ForkOptions,sbt.util.Logger,scala.collection.Seq)sbt.Task, (scala.collection.immutable.Map,sbt.Tests#ProcessedOptions,sbt.Tests#Execution,scala.collection.Seq,sbt.ForkOptions,sbt.util.Logger,sbt.ConcurrentRestrictions#Tag)sbt.Task
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("sbt.ForkTests.apply")
[error]  * method processOptions(sbt.Tests#Execution,scala.collection.immutable.Vector,sbt.util.Logger)sbt.Tests#ProcessedOptions in object sbt.Tests's type is different in current version, where it is (scala.collection.Seq,scala.collection.immutable.Vector,sbt.util.Logger)sbt.Tests#ProcessedOptions instead of (sbt.Tests#Execution,scala.collection.immutable.Vector,sbt.util.Logger)sbt.Tests#ProcessedOptions
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("sbt.Tests.processOptions")
[error]  * method apply(scala.collection.immutable.Map,java.lang.ClassLoader,scala.collection.immutable.Map,scala.collection.immutable.Vector,sbt.Tests#Execution,sbt.internal.util.ManagedLogger)sbt.Task in object sbt.Tests's type is different in current version, where it is (scala.collection.immutable.Map,java.lang.ClassLoader,scala.collection.immutable.Map,sbt.Tests#ProcessedOptions,sbt.Tests#Execution,sbt.internal.util.ManagedLogger)sbt.Task instead of (scala.collection.immutable.Map,java.lang.ClassLoader,scala.collection.immutable.Map,scala.collection.immutable.Vector,sbt.Tests#Execution,sbt.internal.util.ManagedLogger)sbt.Task
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("sbt.Tests.apply")
[error]  * static method apply(scala.collection.immutable.Map,java.lang.ClassLoader,scala.collection.immutable.Map,scala.collection.immutable.Vector,sbt.Tests#Execution,sbt.internal.util.ManagedLogger)sbt.Task in class sbt.Tests's type is different in current version, where it is (scala.collection.immutable.Map,java.lang.ClassLoader,scala.collection.immutable.Map,sbt.Tests#ProcessedOptions,sbt.Tests#Execution,sbt.internal.util.ManagedLogger)sbt.Task instead of (scala.collection.immutable.Map,java.lang.ClassLoader,scala.collection.immutable.Map,scala.collection.immutable.Vector,sbt.Tests#Execution,sbt.internal.util.ManagedLogger)sbt.Task
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("sbt.Tests.apply")
[error]  * method apply(scala.collection.immutable.Map,scala.collection.immutable.Vector,sbt.Tests#Execution,scala.collection.Seq,sbt.ForkOptions,sbt.util.Logger,sbt.ConcurrentRestrictions#Tag)sbt.Task in object sbt.ForkTests in current version does not have a correspondent with same parameter signature among (scala.collection.immutable.Map,sbt.Tests#ProcessedOptions,sbt.Tests#Execution,scala.collection.Seq,sbt.ForkOptions,sbt.util.Logger,scala.collection.Seq)sbt.Task, (scala.collection.immutable.Map,sbt.Tests#ProcessedOptions,sbt.Tests#Execution,scala.collection.Seq,sbt.ForkOptions,sbt.util.Logger,sbt.ConcurrentRestrictions#Tag)sbt.Task
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("sbt.ForkTests.apply")
```

I think you can add `sbt.ForkTests.apply` to the exclude list, but not `sbt.Tests.apply`?

main-actions/src/main/scala/sbt/Tests.scala
VlachJosef: I've added overload for `sbt.Tests.apply`, but I guess I was too eager to add overload also for `sbt.ForkTests.apply`?

eed3si9n:
I just realized that this PR is against 1.4.x branch. Could you target develop branch first please?
sbt 1.5.0 is coming up soon anyway.

VlachJosef:
Sure, I'll do that.

eed3si9n:
1.5.x seems pretty stable, so I don't see a reason to release any more 1.4.x's, so I'm closing this for now.

PR #6418, opened 2021-03-28
Instantiate only test runners needed by current TestDefinitions

### LINKED ISSUE #6412: Munit - test filtering doesn't work when ScalaTest is present in the same project

VlachJosef:
Fixes #6412
Same as #6416, but for `develop` branch instead of `1.4.x`.

For example do not instantiate test runner for ScalaTest if only MUnit test spec is being run.

As it turned out, test runners don't need to be created to determine if test suites are ScalaTest or MUnit etc. Every `TestDefinition` has fingerprint which carries this information. So sbt can decide which test runner to instantiate based on fingerprint of `TestDefinition`s user wants to run.

So rather that creating all test runners up front, this PR adds logic which decides what test runners to create based on `TestDefinition`s which user wants to run.

To get hold of all `TestDefinition`s  user wants to run `Tests.ProcessedOptions` needs to be created. So creation of `Tests.ProcessedOptions` has been moved one layer up, so it can be used to filter test runners. 

Rather than creating `Tests.ProcessedOptions`  in `ForkTests` or `Tests` it is now being passed in as a parameter. This is to avoid doing the same work twice.

### DIFF
--- a/main-actions/src/main/scala/sbt/ForkTests.scala
+++ b/main-actions/src/main/scala/sbt/ForkTests.scala
@@ -23,15 +23,13 @@ import sbt.internal.util.{ RunningProcesses, Terminal => UTerminal }
 private[sbt] object ForkTests {
   def apply(
       runners: Map[TestFramework, Runner],
-      tests: Vector[TestDefinition],
+      opts: ProcessedOptions,
       config: Execution,
       classpath: Seq[File],
       fork: ForkOptions,
       log: Logger,
       tags: (Tag, Int)*
   ): Task[TestOutput] = {
-    val opts = processOptions(config, tests, log)
-
     import std.TaskExtra._
     val dummyLoader = this.getClass.getClassLoader // can't provide the loader for test classes, which is in another jvm
     def all(work: Seq[ClassLoader => Unit]) = work.fork(f => f(dummyLoader))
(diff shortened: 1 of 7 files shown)


eed3si9n (approved):
Thanks!

PR #6422, opened 2021-03-31
Fix Scala 2.13-3.0 sandwich on Scala >= 3.0.0

smarter:
The binary version of Scala >= 3.0.0 will be 3, but the logic was
hardcoded for 3.0.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -3786,7 +3786,7 @@ object Classpaths {
 
   private[sbt] def isScala2Scala3Sandwich(sbv1: String, sbv2: String): Boolean = {
     def compare(a: String, b: String): Boolean =
-      a == "2.13" && (b.startsWith("0.") || b.startsWith("3.0"))
+      a == "2.13" && (b.startsWith("0.") || b.startsWith("3"))
     compare(sbv1, sbv2) || compare(sbv2, sbv1)
   }
 


adpi2 (approved):

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -3786,7 +3786,7 @@ object Classpaths {
 
   private[sbt] def isScala2Scala3Sandwich(sbv1: String, sbv2: String): Boolean = {
     def compare(a: String, b: String): Boolean =
-      a == "2.13" && (b.startsWith("0.") || b.startsWith("3.0"))
+      a == "2.13" && (b.startsWith("0.") || b.startsWith("3"))
```
eed3si9n, line 3789: I thought we don't have concrete guarantee of 2.13 compatibility beyond `3.0`.

smarter: We have concrete guarantee until 3.T for some unknown T, but at that point we'll bump Tasty major too, so incompatible projects will fail at compile-time.

smarter: ... and also this is about the scalaBinaryVersion, which we'll bump when we go to 3.T.
Until we break our own binary compatibility, we'll keep supporting 2.13.

adpi2: Does it work both ways?
We can make 3.x compatible with 2.13 but we will need a 2.13.y to consume each 3.x.

smarter: In that direction the compiler will check the tasty version to determine if it's compatible or not

eed3si9n: ok sounds good.

eed3si9n (approved):

smarter:
Any chance this could get into 1.5.0, or is it too late for that?

eed3si9n:
I think we can slip this in.

PR #6424, opened 2021-04-01
Use full Scala 3 version in crossTarget

### LINKED ISSUE #6415: Allow cross-building of multiple Scala 3 releases by using the full scalaVersion in the crossTarget directory
Currently, the target directory for a given Scala version is `target/scala-${scalaBinaryVersion}`, this means that one shouldn't have two different Scala versions with the same binary version in `crossScalaVersion`, otherwise they'll keep overwriting each other classfiles (I don't know what wold happen exactly, does sbt manages to delete the target directory content when switching version?).
This hasn't really been a problem sufficient so far, but Scala 3 is coming and I think there's valid usecases for cross-compiling between multiple compiler versions:
- If I maintain a library, I may want to cross-compile between 3.0 (this is the minimum version I support and the one I'll publish binaries for) and 3.1 (I won't publish artifacts compiled with it, but I want to make sure it passes my CI)
- In general, I may want to cross-compile between the latest stable version (e.G. 3.0.0) and the latest RC version (e.g. 3.0.1-RC1), to provide feedback on the RC before the release.

To accommodate these usecases, I propose changing `crossTarget` to append the full `scalaVersion` to the target directory name instead of just the `scalaBinaryVersion`. If we want the change to be less disruptive, we could keep the current behavior for Scala 2 and only use the full version for Scala 3. What do you think? /cc @adpi2

adpi2:
Fixes #6415

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -692,6 +692,7 @@ object Defaults extends BuildCommon {
     crossSbtVersions := Vector((pluginCrossBuild / sbtVersion).value),
     crossTarget := makeCrossTarget(
       target.value,
+      scalaVersion.value,
       scalaBinaryVersion.value,
       (pluginCrossBuild / sbtBinaryVersion).value,
       sbtPlugin.value,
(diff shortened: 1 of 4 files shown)


adpi2:
Added the following test:

```scala
scalaVersion := "3.0.0-RC2-bin-20210328-cca5f8f-NIGHTLY"
name := "foo"
version := "1.0.0"
```

```
> compile
$ exists target/scala-3.0.0-RC2-bin-20210328-cca5f8f-NIGHTLY/classes/Foo$.class
> package
$ exists target/scala-3.0.0-RC2-bin-20210328-cca5f8f-NIGHTLY/foo_3.0.0-RC2-1.0.0.jar
```

@smarter Is it what you expect?

smarter:
That looks fine yes.

eed3si9n (approved):
Thanks @adpi2!

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -772,8 +773,23 @@ object Defaults extends BuildCommon {
...
+  def makeCrossTarget(
+      t: File,
+      scalaVersion: String,
+      scalaBinaryVersion: String,
+      sbtv: String,
+      plugin: Boolean,
+      cross: Boolean
+  ): File = {
+    val scalaSuffix =
+      if (ScalaArtifacts.isScala3(scalaVersion)) scalaVersion else scalaBinaryVersion
```
kitbellew, line 791: should the same be used for source directories? lines 745 and 760 above.

also, for lines 745-746 and scala3, is epochVersion different from scalaBinaryVersion?

should the cross-build source be `scala-3`, `scala-3.1` or `scala-3.10`?

eed3si9n: Since Scala 3 is relatively new, so the best practice around how libraries should be maintained is still in flux, but given that Maven Central artifacts would have `_3` suffix, I don't think library authors would be interested in cross building against the minor releases of Scala 3 like 3.0.x and 3.1.x. Likely they'd either decide to stay on 3.0 or bite the bullet and move to 3.1. So from the usage perspective, supporting source directory `scala-3` and `scala-3.1` out of box doesn't make sense for now in my opinion.

If there are other good reason for this, please open a thread under https://github.com/sbt/sbt/discussions

PR #6425, opened 2021-04-03
Allow output directory for JUnitXmlTestsListener to be configured

### LINKED ISSUE #2853: Allow junit test-reports folder to be more finely configured
## Steps

- follow instructions to [create a separate integration test phase](http://www.scala-sbt.org/1.0/docs/Testing.html#Integration+Tests)
- create a simple unit test under `src/test/scala`
- create a simple integration test under `src/integration/scala`
- run both test phases e.g. `sbt test it:test`

## Problem

Junit XML is generated for both tests under target/test-reports. However:

1. there is no obvious (or documented) way to have the reports placed in a different folder. After hunting around the code base, I figured out the following by trial and error:

```scala
testListeners in IntTest := Seq(new JUnitXmlTestsListener("target/integration-report"))
```

This is useful to be able to separate test reports from different test phases so that they can be reported on independently.

2. the JUnitXmlTestsListener always places XML files in a subfolder of the given path called "test-reports", see [the source code](https://github.com/sbt/sbt/blob/57a655478f2fdf1de3e90acf2755b50c0459fd50/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala#L25)

```scala
/**The dir in which we put all result files. Is equal to the given dir + "/test-reports"*/
val targetDir = new File(outputDir + "/test-reports/")
```

so in (1) above the actual output path is target/integration-report/test-reports, which is not what I want.

## Expectation

- save XML reports to the specified folder, and not to a hardcoded subfolder `test-reports`
- include at least a brief mention of JUnitXmlTestsListener and how to use it in the ["Testing > Output"](http://www.scala-sbt.org/0.13/docs/Testing.html#Output) section of the manual
- alternatively, provide a way to more directly configure the output path for each test phase

ashleymercer:
First draft of a PR for this. Some considerations:

- I've exposed a new `testReportsDirectory` setting to allow this to be set on a per-configuration basis
- as a first attempt, I've defaulted this to be `target/${config.name}-reports` which is a *change* from the previous behaviour (always defaulted to `target/test-reports` even for e.g. the `IntegrationTest` config) - I've added notes about this breaking backwards compatibility but wanted to check this was acceptable? An alternative might be to default to `test-reports`?
- I've also (per an existing comment) changed the listener to only be attached to the `Test` and `IntegrationTest` configurations directly, rather than at the global level - again, this breaks compatibility for completely custom configs (NB any config which inherits from `Test` or `IntegrationTest` will continue to have the listener, which should cover most cases?)

Fixes #2853


ashleymercer:
Okay, binary compat check failed because the listener no longer has an `outputDir()` property - is it safe to add this to the list of exclusions? Alternatively I can add it back.

testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
```diff
@@ -31,13 +31,14 @@ import sbt.protocol.testing.TestResult
 /**
  * A tests listener that outputs the results it receives in junit xml
  * report format.
- * @param outputDir path to the dir in which a folder with results is generated
+ * @param targetDir directory in which test reports are generated
  */
-class JUnitXmlTestsListener(val outputDir: String, legacyTestReport: Boolean, logger: Logger)
+class JUnitXmlTestsListener(val targetDir: File, legacyTestReport: Boolean, logger: Logger)
```
eed3si9n, line 36: If the current API is using String to pass the absolute path around, then we should keep that behavior to minimize the change in case this listener class is extended by others.

eed3si9n:
Thanks for working on this!

testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
ashleymercer: Good point! I've updated to add (another!) legacy constructor which keeps the old API.

ashleymercer: (To be clear: the current API accepts `outputDir` as a String in the constructor, but actually writes to `targetDir` which is constructed as `${outputDir}/test-reports`. I believe the new code follows this behaviour as closely as possible)

eed3si9n (dismissed):
Thanks!

eed3si9n:
LGTM. Could you squash the commits and move some of the descriptions to commit message please?

ashleymercer:
> LGTM. Could you squash the commits and move some of the descriptions to commit message please?

Done :)

PR #6431, opened 2021-04-06
 Fix unused nowarn in sbt plugins

### LINKED ISSUE #6430: `@nowarn annotation does not suppress any warnings` when compiling SBT plugins
## steps

Trying to build an SBT plugin with sbt 1.5.0, in our case [twinagle](https://github.com/soundcloud/twinagle/pull/224). 

## problem

We get `@nowarn annotation does not suppress any warnings` [errors](https://github.com/soundcloud/twinagle/pull/224/checks?check_run_id=2276625337#step:5:208) in the code that extends `AutoPlugin`.

## expectation

No warnings/errors are emitted.

## notes

Seems related to https://github.com/sbt/sbt/issues/6398

adpi2:
Fixes #6430

### DIFF
--- a/main/src/main/scala/sbt/plugins/SbtPlugin.scala
+++ b/main/src/main/scala/sbt/plugins/SbtPlugin.scala
@@ -10,11 +10,22 @@ package plugins
 
 import Keys._
 import Def.Setting
+import sbt.SlashSyntax0._
+import sbt.librarymanagement.Configurations.Compile
+import sbt.librarymanagement.{ SemanticSelector, VersionNumber }
 
 object SbtPlugin extends AutoPlugin {
   override def requires = ScriptedPlugin
 
   override lazy val projectSettings: Seq[Setting[_]] = Seq(
-    sbtPlugin := true
+    sbtPlugin := true,
+    Compile / scalacOptions ++= {
+      // silence unused @nowarns in 2.12 because of https://github.com/sbt/sbt/issues/6398
+      // the option is only available since 2.12.13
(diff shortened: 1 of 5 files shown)


main/src/main/scala/sbt/plugins/SbtPlugin.scala
```diff
@@ -10,11 +10,14 @@ package plugins
...
+import sbt.SlashSyntax0._
+import sbt.librarymanagement.Configurations.Compile
 
 object SbtPlugin extends AutoPlugin {
   override def requires = ScriptedPlugin
 
   override lazy val projectSettings: Seq[Setting[_]] = Seq(
-    sbtPlugin := true
+    sbtPlugin := true,
+    Compile / scalacOptions += "-Wconf:cat=unused-nowarn:s"
```
mkurz: Isn't this too global? Wouldn't this silence also other "unused" warnings?

adpi2: This option does only silence the unused `@nowarn` warnings in the `SbtPlugin` projects.

```
@nowarn annotation does not suppress any warnings
```

It is not a perfect solution because it will silence your own unused `@nowarn` annotations. But it is very specific to sbt plugins and it is harmless.

A better solution would be to have scala/scala#9491 but it is only available in 2.13.

eed3si9n (commented):
Thanks for the quick fix!

The fix LGTM but could you add fatal warning to some scripted plugin build as a test plz?

adpi2:
Test added.

Also the option is only added if scalaVersion >= 2.12.13 because it was not available in 2.12.12.

So the solution in 2.12.12 is to upgrade to 2.12.13

eed3si9n (approved):

PR #6456, opened 2021-04-19
Retain classifiers for transitive dependencies when publishing to Ivy using Coursier

### LINKED ISSUE #5325: Transitive dependencies with classifier are not pulled
## steps
sbt version:1.3.5

* Clone https://github.com/pdalpra/coursier-test
* Go to the `netty-epoll-direct` folder
* Run `sbt ';update;show fullClasspath;publishLocal'` => Notice that the classpath does include `netty-transport-native-epoll-4.1.44.Final-linux-x86_64.jar`
* Go to the `netty-epoll-transitive` folder
* Run `sbt ';update;show fullClasspath` => Notice that that all dependency (including transitive ones) are there, except `netty-transport-native-epoll-4.1.44.Final-linux-x86_64.jar`, which has been replaced by `netty-transport-native-epoll-4.1.44.Final.jar` (no classifier)

## problem

sbt with coursier ignores the classifiers from transitive dependencies. This works with Ivy.
I have tried to tweak `transitiveClassifiers` to include the classifier, without success.

## expectation

sbt with coursier should honor the classifiers even on transitive dependencies.

guilgaly:
Fixes #5325 

Notes:
- I reproduced the issue (and tested the fix) on a private repo, but I can provide a public reproducer project if that's useful for someone else
- the format of the `<artifact ... />` XML element is copied from what I got when publishing with `useCoursier := false`, since that worked

### DIFF
--- a/main/src/main/scala/sbt/internal/librarymanagement/IvyXml.scala
+++ b/main/src/main/scala/sbt/internal/librarymanagement/IvyXml.scala
@@ -152,6 +152,18 @@ object IvyXml {
 
     val dependencyElems = project.dependencies.toVector.map {
       case (conf, dep) =>
+        val classifier = {
+          val pub = dep.publication
+          if (pub.classifier.value.nonEmpty)
+            Seq(
+              <artifact name={pub.name} type={pub.`type`.value} ext={pub.ext.value} conf="*" e:classifier={
+                pub.classifier.value
+              } />
+            )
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks @guilgaly!

eed3si9n:
The test failure seems unrelated.

PR #6480, opened 2021-04-29
Move scalaInstanceTopLoader to compileBase settings

adpi2:
- Move `scalaInstanceTopLoader` and `useScalaReplJLine` to `compileBase`:
They are/should only be used in the `JvmPlugin` to create the `scalaInstance`.

- Move `jansiExclusionLoader` (for Scala 2.10 only) to the `scalaInstanceTopLoader` implementation

- Remove the `ClassLoaderCache` hack for dotty:
It is not needed anymore, since dotty uses sbt 1.5

- Make `Defaults.makeScalaInstance` public:
`ScalaInstance` and the `scalaInstance` task are public. But, to create a custom `scalaInstance` we need the `extendedClassLoaderCache` which is package private. That's why I think `makeScalaInstance` should be public.

This will allow us to properly bootstrap the Scala compiler in the dotty repository. See https://github.com/lampepfl/dotty/pull/12259.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -662,10 +662,20 @@ object Defaults extends BuildCommon {
   def compileBase = inTask(console)(compilersSetting :: Nil) ++ compileBaseGlobal ++ Seq(
     useScalaReplJLine :== false,
     scalaInstanceTopLoader := {
-      // the JLineLoader contains the SbtInterfaceClassLoader
-      if (!useScalaReplJLine.value)
-        classOf[org.jline.terminal.Terminal].getClassLoader // the JLineLoader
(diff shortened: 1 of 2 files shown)


eed3si9n:
I haven't branched out develop to 1.5.x yet. Is this more of a 1.6.0-M1 feature or can this be "backported" to 1.5.x without going through the RC cycle?

adpi2:
It is not forward compatible because I changed the scope of `scalaInstanceTopLoader`. So it should go to `1.6.x`.

I can backport a lighter version, that just makes `makeScalaInstance` public.

adpi2:
Maybe we can merge #6481 and put this one on hold for now.

adpi2:
@eed3si9n  This one should be ready to merge.

eed3si9n (approved):

PR #6483, opened 2021-04-30
sbt script: Favor java's `user.home` over `$HOME`

rdesgroppes:
JVM powered applications may be given an alternate home by means of the
`user.home` system property, which is handy for managing caches in CI
(the property may be set by different means, including environment
variables such as `JAVA_TOOL_OPTIONS` or `_JAVA_OPTIONS`).

Alas, this doesn't fully work when the `sbt` script downloads the
launcher jar to `$HOME`.

The present change consists in retrieving the value of this property by
means of a `findProperty` function extracted from the existing
`getPreloaded` one (adapted accordingly).

No java process get spawned here.

### DIFF
--- a/sbt
+++ b/sbt
@@ -126,7 +126,8 @@ acquire_sbt_jar () {
       launcher_sv="$builtin_sbt_version"
     fi
   fi
-  download_jar="$HOME/.cache/sbt/boot/sbt-launch/$launcher_sv/sbt-launch-$launcher_sv.jar"
+  local user_home && user_home=$(findProperty user.home)
+  download_jar="${user_home:-$HOME}/.cache/sbt/boot/sbt-launch/$launcher_sv/sbt-launch-$launcher_sv.jar"
   if [[ -f "$download_jar" ]]; then
     sbt_jar="$download_jar"
   else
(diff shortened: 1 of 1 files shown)


sbt
```diff
@@ -117,6 +117,16 @@ download_url () {
   } && [[ -f "$jar" ]]
 }
 
+user_home () {
+  local user_home
+  user_home=$("$java_cmd" -XshowSettings:properties -version 2>&1 | sed -n 's/\s*user.home = //p' 2>/dev/null)
```
ritschwumm: will this be called on every startup, or only when sbt is not installed yet?

rdesgroppes: It's executed from within a shell function, so: only when there's a need for resolving a path under user's home directory.

There are currently 2 cases: either to form a path where to download the launcher jar (if needed), or to sync "preloaded" jars from sbt's installation directory to user home directory (if needed).

eed3si9n: Each call to `time java -XshowSettings:properties -version` adds around 100ms.

For `--client` mode that could potentially double our triple the startup - https://github.com/sbt/sbt-launcher-package/pull/334

vegerot: Do we measure this anywhere?  It would have been nice if https://github.com/sbt/sbt-launcher-package/pull/334 also added tests for this

eed3si9n: No currently we don't have any benchmark or measurements on `sbt` script startup time.

rdesgroppes: OK, following above concern, I've amended the commit with a second approach that factors out property search.

eed3si9n (approved):
Thanks @rdesgroppes!

sbt
```diff
@@ -301,39 +302,49 @@ jdk_version() {
...
+getPreloaded() {
+  local preloaded && preloaded=$(findProperty sbt.preloaded)
+  [ "$preloaded" ] && echo "$preloaded" && return
+
+  local global_base && global_base=$(findProperty sbt.global.base)
+  [ "$global_base" ] && echo "$global_base/preloaded" && return
 
-  echo "${via_explicit:-${via_global_base}}"
+  local user_home && user_home=$(findProperty user.home)
+  echo "${user_home:-$HOME}/.sbt/preloaded}}"
```
rdesgroppes, line 347: :warning: trailing `}}`, my bad!

rdesgroppes: => #6494.

PR #6499, opened 2021-05-07
Add missing scalac options to scaladoc task configuration

pikinier20:
Related to: [#12298](https://github.com/lampepfl/dotty/issues/12298)


lightbend-cla-validator:
Hi @pikinier20,

adpi2:
Thanks @pikinier20! Can you give us more details about how scaladoc handles unused scalac options?

From what I know, scaladoc 3.0.0 is still printing a lot of warnings because of unused options. So maybe we should wait for 3.0.1 before merging. What do you think?

pikinier20:
> Thanks @pikinier20! Can you give us more details about how scaladoc handles unused scalac options?
> 
> From what I know, scaladoc 3.0.0 is still printing a lot of warnings because of unused options. So maybe we should wait for 3.0.1 before merging. What do you think?

Yes, I think it should be released after release with my PR to dotty, because we are gonna have warns.

My PR to dotty just filters scalacOptions and doesn't print warns for standard scala options.

smarter:
I think merging and releasing an sbt with this PR in as soon as possible would be helpful since right now many people cannot publish their scala.js project because doc generation fails (https://github.com/lampepfl/dotty/issues/11943), and the only workaround we can give them is to disable doc generation, when I'd rather tell them to try a dotty nightly with the latest sbt to ensure that this setup works correctly.

adpi2:
This fix will generate a lot of warnings on all Scala 3.0.0 projects (not only Scala.js):
```text
sbt:root> kernelJVM / doc
[info] compiling 35 Scala sources to /home/piquerez/typelevel/cats-effect/kernel/jvm/target/scala-3.0.0/classes ...
[info] Main Scala API documentation to /home/piquerez/typelevel/cats-effect/kernel/jvm/target/scala-3.0.0/api...
[warn] bad option '-deprecation' was ignored
[warn] bad option '-feature' was ignored
[warn] bad option '-unchecked' was ignored
[warn] bad option '-language:implicitConversions' was ignored
[warn] bad option '-Ykind-projector' was ignored
[warn] bad option '-source:3.0-migration' was ignored
[warn] bad option '-Xsemanticdb' was ignored
[warn] bad option '-semanticdb-target' was ignored
[warn] bad option '-doc-external-doc:/home/piquerez/.cache/coursier/v1/https/repo1.maven.org/maven2/org/typelevel/cats-core_3/2.6.1/cats-core_3-2.6.1.jar#http://typelevel.org/cats/api/,/home/piquerez/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.13.5/scala-library-2.13.5.jar#https://www.scala-lang.org/api/2.13.5/,/home/piquerez/.cache/coursier/v1/https/repo1.maven.org/maven2/org/typelevel/cats-kernel_3/2.6.1/cats-kernel_3-2.6.1.jar#http://typelevel.org/cats/api/' was ignored
[warn] Setting -bootclasspath is currently not supported.
[warn] Setting -sourcepath is currently not supported.
[warn] 11 warnings found
[success] Total time: 42 s, completed May 25, 2021 9:35:51 AM
```

If people want to try a nightly version of scaladoc with Scala.js, they can do:
```scala
Compile / doc / scalacOptions ++= (Compile / scalacOptions).value
```

@pikinier20 maybe you could add a scripted test in this PR to check that this fix works with a nighlty version of Scala 3.0.1?

smarter:
> This fix will generate a lot of warnings on all Scala 3.0.0 projects (not only Scala.js):

I understand that, but I think that warnings all the time are better than crashes some of the time :).

russwyte:
> 
> 
> > This fix will generate a lot of warnings on all Scala 3.0.0 projects (not only Scala.js):
> 
> I understand that, but I think that warnings all the time are better than crashes some of the time :).

Agreed this has been blocking several releases of libraries migrating to Scala 3

eed3si9n:
Having unactionable warnings to all JS users would be pretty bad UX. If Scaladoc also respect fatal warnings wouldn't it also result to occasional errors? Is it possible to maybe filter in only the necessary scalacOptions only when Scala version is 3.0.0 or something?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2107,7 +2107,11 @@ object Defaults extends BuildCommon {
           val projectName = name.value
           if (ScalaArtifacts.isScala3(sv)) {
             val project = if (config == Compile) projectName else s"$projectName-$config"
-            Seq("-project", project)
+            if (scalaVersion.value.startsWith("3.0.0")) {
+              Seq("-project", project)
```
smarter, line 2111: We could still pass -scalajs here to avoid most issues without introducing warnings:
```suggestion
              compileOptions.filter(_ == "-scalajs") ++ Seq("-project", project)
```

adpi2: Is `scaladoc` 3.0.0 able to use `-scalajs`? I think it was failing in `cats-effect` repo even with `-scalajs`.

pikinier20: That's true. We didn't handle `-scalajs` in 3.0.0.

adpi2 (approved):
Thanks a lot @pikinier20!

I think this can be merged in 1.5 and released before Scala 3.0.1 goes out.

smarter (approved):

PR #6500, opened 2021-05-07
make -client the same as --client

### LINKED ISSUE #6078: `-client` is different from `--client`
## steps

```bash
sbt -client
```

## problem

`-client` is different from `--client`

## expectation

`-client` is same as `--client`

## note

Make a new argument that's like `--java-client` or something for the old behavior currently implemented by `-client`?

_Originally posted by @eatkins in https://github.com/sbt/website/pull/948#issuecomment-724312989_

Nirvikalpa108:
The problem was that -client was different from --client, which makes for a confusing user experience. So, this change makes
them the same and re-names -client to --java-client. The value of this is that hopefully -client and --client being the same feels more logical to users.

fixes #6078


eed3si9n (approved):
Thanks @Nirvikalpa108!

PR #6504, opened 2021-05-10
Expose makeScalaInstance

adpi2:
I missed the point of my original PR in #6481


eed3si9n (approved):
Yea. I should've caught this during review, but I only realized it as I was about write a release note.

PR #6509, opened 2021-05-15
Make `sbt --version` respect SBT_OPTS and friends

parched:
Having set `SBT_OPTS="-Dsbt.boot.directory=/not/the/default"` I found when I ran `sbt --version` SBT was being redownloaded into `~/.sbt`. This fixes that and I assume `SBT_OPTS` wasn't ignored on purpose, but I'm not sure.


lightbend-cla-validator:
Hi @parched,

eed3si9n:
> This fixes that and I assume `SBT_OPTS` wasn't ignored on purpose, but I'm not sure.

https://github.com/sbt/sbt-launcher-package/pull/276 added it, and I'm guessing `SBT_OPTS` was ignored on purpose since mostly those define memory options.

parched:
>  I'm guessing `SBT_OPTS` was ignored on purpose since mostly those define memory options.

Ok, I'm happy to close the PR if you think that's best? I worked around it by just doing `sbt sbtVersion` instead so not a blocker for me.

eed3si9n:
I think the direction of the change is good but the tests are currently failing. It looks like the previous incantation didn't have color but the new one does?

sbt
```diff
@@ -432,22 +432,19 @@ run() {
     addJava "-Djline.terminal=jline.UnixTerminal"
     addJava "-Dsbt.cygwin=true"
   fi
+  
+  local runSbt=(execRunner "$java_cmd" "${java_args[@]}" "${sbt_options[@]}" -jar "$sbt_jar")
 
   if [[ $print_sbt_version ]]; then
-    execRunner "$java_cmd" -jar "$sbt_jar" "sbtVersion" | tail -1 | sed -e 's/\[info\]//g'
+    "${runSbt[@]}" "sbtVersion" | tail -1 | sed -e 's/\[info\]//g'
```
parched: This should fix it
```suggestion
    "${runSbt[@]}" -Dsbt.color=never "sbtVersion" | tail -1 | sed -e 's/\[info\]//g'
```

sbt
```diff
@@ -432,22 +432,19 @@ run() {
...
   elif [[ $print_sbt_script_version ]]; then
     echo "$init_sbt_version"
   elif [[ $print_version ]]; then
-    execRunner "$java_cmd" -jar "$sbt_jar" "sbtVersion" | tail -1 | sed -e 's/\[info\]/sbt version in this project:/g'
+    "${runSbt[@]}" "sbtVersion" | tail -1 | sed -e 's/\[info\]/sbt version in this project:/g'
```
parched:
```suggestion
    "${runSbt[@]}" -Dsbt.color=never "sbtVersion" | tail -1 | sed -e 's/\[info\]/sbt version in this project:/g'
```

parched (commented):
Ok, I've found the problem, will push a fix when I'm back at my computer.

sbt
eed3si9n: Since `-Dsbt.color` is a relatively recent addition, and the runner script needs to work with all the versions of sbt, `-Dsbt.log.noformat=true` might work better.

```suggestion
    "${runSbt[@]}" -Dsbt.log.noformat=true "sbtVersion" | tail -1 | sed -e 's/\[info\]//g'
```

sbt
eed3si9n:
```suggestion
    "${runSbt[@]}" -Dsbt.log.noformat=true "sbtVersion" | tail -1 | sed -e 's/\[info\]/sbt version in this project:/g'
```

PR #6335, opened 2021-02-24
Add MUnit test framework

julienrf:
Ping @olafurpg

PR #6513, opened 2021-05-17
Add AirSpec testing framework support

xerial:
AirSpec is a tiny testing framework for Scala 2.x, 3.x, and Scala.js https://wvlet.org/airframe/docs/airspec

This PR adds support for running AirSpec without having `testFrameworks` setting.


eed3si9n:
There are over 100 usages of MUnit on https://mvnrepository.com/artifact/org.scalameta/munit, so I think it makes sense to provide out-of-box support for it.

On the other hand, AirSpec at the moment seems to be limited to internal usage to AirSpec itself - https://mvnrepository.com/artifact/org.wvlet.airframe/airspec/21.5.0/usages. Given that once we add this, we won't be able to remove it, I feel like it's a bit premature for AirSpec.

xerial:
@eed3si9n Yeah. Currently, adding AirSpec is not a no-brainer like adding JUnit, ScalaTest, MUnit, etc. I've just found adding such support was that easy. 

I'll close this PR for now. 

Thanks for the feedback!

xerial:
For reference: https://github.com/sbt/sbt/pull/6335#issuecomment-785095907

neko-kai:
@eed3si9n 
To be honest I see no strictly compelling reason not to add _any_ Scala test framework that is willing to submit a PR to sbt’s default list. This massively reduces friction for test frameworks to be adopted (the other hurdle is Intellij support, but it’s much more of a nice-to-have) and the only risks for sbt are:

1. Bitrot of the test framework can fail sbt’s tests – this is mitigated by NOT adding tests for the test framework into sbt’s test suite. There isn’t much to test but the package name and this can be left to the test framework’s tests, not sbt’s.
2. Package/class name clash, rot or capture – this isn’t easily mitigated, but Scala libraries by default try to use rather unique names to prevent package names clashes, this risk is largely theoretical.

As long as sbt keeps being regularly released it could keep an up-to-date list of Scala test frameworks and allow further experimentation in Scala test frameworks – this is fully in-line with sbt having its own test framework API in the first place – it reduces friction for new libraries to appear.

eed3si9n:
@neko-kai Generally I acknowledge this is a subjective topic, and my opinion on this is weakly held. And I also appreciate you chiming in.

> As long as sbt keeps being regularly released it could keep an up-to-date list of Scala test frameworks and allow further experimentation in Scala test frameworks – this is fully in-line with sbt having its own test framework API in the first place

The interface is public, and it's extensible via settings, so I think the experimentation aspect is already enabled. And like any good experiment, some would float and others might not.

It seems like any sufficiently senior Scala developers and/or projects eventually get an itch to implement a test framework (full disclosure: I have one :)). I guess sbt can start collecting them, but over time, not all of them would take off. I could be wrong on this, but I'm somewhat skeptical of the "massive reduction of friction" since frameworks like uTest and MUnit has done ok without the out-of-box support. But like I said, there's probably no objective reason to accumulate these, so I'd be happy to reverse my course.

PR #6517, opened 2021-05-21
Fixes spurious "@nowarn annotation does not suppress any warnings" in old style sbt plugins with `sbtPlugin := true`

### LINKED ISSUE #6430: `@nowarn annotation does not suppress any warnings` when compiling SBT plugins

Nirvikalpa108:
## What is the problem?
As detailed in #6430, the @nowarn annotation was not suppressing warnings, even after the first attempt to fix this in PR#6431. This first PR fixed the problem for projects using `enablePlugins(SbtPlugin)`, but not for those using `sbtPlugin := true`.

## Why is this a valuable problem to solve?
The annotation was not working as users would expect.

## What is the solution?
I have moved the scalacOptions change from `SbtPlugin.projectSettings` to the scalacOptions in the `JvmPlugin` settings.

## Has this been tested?
Yes, a test has been added. Also, this branch was tested successfully on the twinagle repo (https://github.com/soundcloud/twinagle/pull/224).

Fixes #6430.

### DIFF
--- a/sbt-app/src/sbt-test/project/sbt-plugin/test
+++ b/sbt-app/src/sbt-test/project/sbt-plugin/test
@@ -1,1 +1,3 @@
 > compile
+$ copy-file changes/oldSbtPlugin.sbt build.sbt
+> compile
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -959,6 +959,13 @@ object Defaults extends BuildCommon {
         Vector("-Ypickle-java", "-Ypickle-write", converter.toPath(earlyOutput.value).toString) ++ old
       else old
     },
+    scalacOptions := {
+      val old = scalacOptions.value
+      if (sbtPlugin.value && VersionNumber(scalaVersion.value)
+            .matchesSemVer(SemanticSelector("=2.12 >=2.12.13")))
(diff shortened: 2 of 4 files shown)


sbt-app/src/sbt-test/project/sbt-plugin/test
```diff
@@ -1 +1,3 @@
 > compile
+$ copy-file changes/oldSbtPlugin.sbt build.sbt
```
eed3si9n, line 2: Is this file missing from the commit?

eed3si9n:
@Nirvikalpa108 Thanks for the contribution!

sbt-app/src/sbt-test/project/sbt-plugin/test
Nirvikalpa108: Good spot, thank you 👍🏽

eed3si9n (approved):

PR #6538, opened 2021-06-09
Fix compiler class loader jars

adpi2:
Use the exact list of compiler jars in the cache key of the compiler class loader. This makes the cache more reliable for someone who creates its own Scala instance, as in https://github.com/lampepfl/dotty/pull/12771.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1182,12 +1182,7 @@ object Defaults extends BuildCommon {
     val allJars = libraryJars ++ compilerJars ++ docJars
 
     val libraryLoader = classLoaderCache(libraryJars.toList, jansiExclusionLoader)
-    val compilerLoader = classLoaderCache(
-      // It should be `compilerJars` but it would break on `3.0.0-M2` because of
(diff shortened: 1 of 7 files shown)


eed3si9n:
The comment says "// It should be `compilerJars` but it would break on `3.0.0-M2` because of", and it broke the scripted test:

```
[info] Running compiler-project/scala3-tasty-management
[info] [error] ## Exception when compiling 2 sources to /tmp/sbt_76d84352/target/scala-3.0.0-M2/classes
[info] [error] java.lang.NoClassDefFoundError: scala/runtime/Nothing$
[info] [error] xsbt.CachedCompilerImpl.run(CachedCompilerImpl.java:65)
[info] [error] xsbt.CompilerInterface.run(CompilerInterface.java:41)
```

Maybe this should be in sbt 1.6.x?

adpi2:
The problem with M2 is fixed in M3 and subsequent versions. So this PR could be merged in 1.5.x, but there is no urgent reason to do so. I let you decide.

eed3si9n (approved):
I'm ok with releasing this as 1.5.x, but we should include it as a breaking change since some people's build matrix may include 3.0.0-M2, even if it's no longer relevant.

PR #6539, opened 2021-06-09
Close BSP channel after receiving build/exit notification

### LINKED ISSUE #6536: [BSP] - `exit` notification not shutting down the server.
## The problem

To be honest, I'm not 100% sure what the expected behavior should be in this scenario, but I'm trying to work an issue with Metals starting sbt as a BSP server and it still hanging around. My debugging led me to the the `exit` notification in sbt. To better outline the issue I'll explain below.

## Steps to reproduce

1. In a new minimal repo do a `sbt bspConfig` to ensure you have your `.bsp/sbt.json`.
2. Open metals in the repo, ignore the build import from Bloop, trigger a `Metals Switch Build Server` command, which differs a bit based on client. Select sbt
3.  At this point take a look at the processes that are running. You'll see something like this after Metals is connected to sbt.

```
❯ jps -l
79137 /Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-sbt/sbt-launch/1.5.3/sbt-launch-1.5.3.jar
79136 xsbt.boot.Boot
79111 /Users/ckipp/.cache/nvim/nvim-metals/metals
79148 jdk.jcmd/sun.tools.jps.Jps
```

4. At this point, close Metals. You can tail the [lsp trace file](https://scalameta.org/metals/docs/contributors/getting-started.html#json-rpc-trace) to ensure that Metals correctly shuts down and sends the `exit` notification.
5. After the `exit` was sent and Metals is shut down, take another look at the processes.

```
❯ jps -l
79137 /Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-sbt/sbt-launch/1.5.3/sbt-launch-1.5.3.jar
79157 jdk.jcmd/sun.tools.jps.Jps
```
This is where the main issue lies. You'll see the `sbt-launch.jar` still chilling there. If you also try to start `sbt` in this workspace you'll get a warning that there is already another server running there. In order to stop it, issue a `sbt --client shutdown`. You'll then see the warning will be gone if you try again and the `sbt-launch.jar` will be shutdown. At first I thought we weren't ending the process, but looking at the sbt code here:

https://github.com/sbt/sbt/blob/9e98b6daa1c5021b852661feb2a5b263bda9ce67/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala#L264-L265

It seems that upon receiving the `exit` notification that a `Shutdown` should be executed. Am I understanding that `Shutdown` correctly that it should be doing the same thing as the `sbt --client shutdown` that I do now to shut down the server? Or is that not expected and we should actually handle the full shutdown on the Metals side?

## expectation

I'd sort of expect that `Shutdown` to indeed shut down the server so it's not still hanging around

adpi2:
`build/exit` is a notification rather than a request.

When we receive it we don't want to shutdown the sbt server but we can close the network channel.

Fixes #6536

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
@@ -11,7 +11,6 @@ package server
 
 import java.net.URI
 
-import sbt.BasicCommandStrings.Shutdown
 import sbt.BuildSyntax._
 import sbt.Def._
 import sbt.Keys._
@@ -239,7 +238,7 @@ object BuildServerProtocol {
     ServerHandler { callback =>
       ServerIntent(
         onRequest = {
-          case r: JsonRpcRequestMessage if r.method == "build/initialize" =>
+          case r if r.method == "build/initialize" =>
             val params = Converter.fromJson[InitializeBuildParams](json(r)).get
             checkMetalsCompatibility(semanticdbEnabled, semanticdbVersion, params, callback.log)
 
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #6552, opened 2021-06-17
Bsp build target resources

### LINKED ISSUE #6550: [BSP] Implement buildTarget/resources
## steps

Import a project with resource files in `src/main/resources` on Intellij Idea using SBT via BSP. 

## problem

The BSP specification provides a [buildTarget/resources](https://github.com/build-server-protocol/build-server-protocol/blob/master/docs/specification.md#resources-request) that allows getting the resources for a build target. 

The [BspProjectResolver](https://github.com/JetBrains/intellij-scala/blob/idea212.x/bsp/src/org/jetbrains/bsp/project/importing/BspProjectResolver.scala#L280) on intellij side needs this query to complete its build target data with resources. 

The [BuildServerProtocol handler](https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala#L225) on SBT side does not implement the `buildTarget/resources` query. 

## expectation

`buildTarget/resources` query should be implemented in [BuildServerProtocol handler](https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala#L225)

## notes
_(I can open a pull request to fix this issue if needed)_

github-samuel-clarenc:
Fixes https://github.com/sbt/sbt/issues/6550

This PR adds an implementation for [BSP resource request](https://build-server-protocol.github.io/docs/specification#resources-request)

When sending the following command to `sbt -bsp` : 

```
{ "jsonrpc": "2.0", "id": 1, "params": {"targets":[{"uri":"file:/PROJECT_PATH/#module/Compile"}]}, "method": "buildTarget/resources" }
```

We should now get the following response : 

```
{"jsonrpc":"2.0","id":1,"result":{"items":[{"target":{"uri":"file:/PROJECT_PATH/#module/Compile"},"resources":["file:PROJECT_PATH/module/src/main/resources/"]}]}}
```


protocol/src/main/contraband-scala/sbt/internal/bsp/ResourcesItem.scala
```diff
@@ -0,0 +1,37 @@
+/**
```
github-samuel-clarenc, line 1: I wonder why those generated files are committed, any idea ?

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -201,6 +214,12 @@ object BuildServerProtocol {
         })
       SourcesItem(id, items)
     },
+    bspBuildTargetResourcesItem := {
+      val id = bspTargetIdentifier.value
+      val dirs = unmanagedResourceDirectories.value
+      val managed = managedResources.value
+      ResourcesItem(id, dirs.toVector.map(_.toURI) ++ managed.toVector.map(_.toURI))
```
adpi2: Here, some resource directories are mixed with some other resource files, which does not seem right. The specification says it should contain "a list of resource file". But, if we do so we lose the relative path of each resource which is probably needed by IntelliJ when it is copied to the target directory.

So I think the specification of the `ResourceItem` is incomplete, see https://github.com/build-server-protocol/build-server-protocol/issues/181

protocol/src/main/contraband-scala/sbt/internal/bsp/ResourcesItem.scala
adpi2: It is probably a safety net to avoid breaking compatibility between versions

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
github-samuel-clarenc: You're right, the sources items have a `kind` attribute that can make a difference between a source file or directory ! 

I had a look at the intellij scala plugin : it seems that the resources can be successfully imported by [BspProjectResolver](https://github.com/JetBrains/intellij-scala/blob/idea212.x/bsp/src/org/jetbrains/bsp/project/importing/BspProjectResolver.scala#L280) . The only thing here is that I didn't know where the [BuildServerCapabilities.resourcesProvider](https://github.com/build-server-protocol/build-server-protocol/blob/master/bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/BuildServerCapabilities.java#L26) was set so I tried setting it manually to true in the intellij scala plugin to check that the resources were successfully imported in my project. 

_Edit : the `resourcesProvider` information should probably be in the buildTarget/initialize request._

github-samuel-clarenc: Answering my own question : I need to update [bsp.contra](https://github.com/sbt/sbt/blob/develop/protocol/src/main/contraband/bsp.contra#L203) in the current PR to update the `resourcesProvider` flag

github-samuel-clarenc:
This PR was checked on a sample project : the `buildTarget/resources` is successfully called during the project import and Intellij Idea can rightly define the resources folders : 
image

adpi2:
I created an example project to show that in some case, the build client lacks some information about the resource file.

The build file is:
```scala
val hello = project.in(file("."))
  .settings(
    scalaVersion := "2.13.6",
    Compile / resourceGenerators += Def.task {
      val greeting = (Compile / resourceManaged).value / "assets" / "greeting.txt"
      IO.write(greeting, "Hello, World!")
      Seq(greeting)
    }
  )
```

This generates a `assets/greeting.txt` resource file.

The main method read the content of that resource and prints it:
```scala
package hello

import scala.io.Source

object Main extends App {
  Source.fromResource("assets/greeting.txt", getClass.getClassLoader)
    .getLines()
    .foreach(println)
}
```
Note that the resource is `assets/greeting.txt` and not `greeting.txt`.

Running the main to check that it works as expected:
```
sbt:hello> run
[info] running hello.Main 
Hello, World!
[success] Total time: 0 s, completed Jun 18, 2021 10:45:21 AM
```

When importing the project in IntelliJ the `buildTarget/resources` response is:

```json
{
  "items": [
    {
      "target": {
        "uri": "file:/home/piquerez/sbt/sbt-6552/#hello/Test"
      },
      "resources": [
        "file:/home/piquerez/sbt/sbt-6552/src/test/resources"
      ]
    },
    {
      "target": {
        "uri": "file:/home/piquerez/sbt/sbt-6552/#hello/Compile"
      },
      "resources": [
        "file:/home/piquerez/sbt/sbt-6552/src/main/resources",
        "file:/home/piquerez/sbt/sbt-6552/target/scala-2.13/resource_managed/main/assets/greeting.txt"
      ]
    }
  ]
}
```

Here Intellij does not know that the resource file is supposed to be `assets/greeting.txt` rather than `greeting.txt`. Or, in other words, that the resource folder should be `file:/home/piquerez/sbt/sbt-6552/target/scala-2.13/resource_managed/main/` instead of `file:/home/piquerez/sbt/sbt-6552/target/scala-2.13/resource_managed/main/assets/`.

Here IntelliJ assumes that the resource folder is `file:/home/piquerez/sbt/sbt-6552/target/scala-2.13/resource_managed/main/assets/` which is wrong:
image

Reproduction repo: https://github.com/adpi2/sbt-6552

adpi2:
I think the solution for now is to send back the managed resource folders rather than the managed resource files:

```scala
val id = bspTargetIdentifier.value 
val unmanagedDirs = unmanagedResourceDirectories.value
val managedDirs = managedResourceDirectories.value
ResourcesItem(id, unmanagedDirs.toVector.map(_.toURI) ++ managedDirs.toVector.map(_.toURI))
``` 

Or more simply

```scala
val id = bspTargetIdentifier.value 
val resourcDirs = resourceDirectories.value
ResourcesItem(id, resourcDirs.toVector.map(_.toURI))
```

One drawback of this solution is that it does not trigger the resource generation (of the managed resources). We can handle that with:

```scala
val id = bspTargetIdentifier.value 
val resourcDirs = resourceDirectories.value
// trigger resource generation
val _ = managedResources.value
ResourcesItem(id, resourcDirs.toVector.map(_.toURI))
```

github-samuel-clarenc:
You're right, the main problem comes from the lack of resources kind, like we have with [sources](https://github.com/build-server-protocol/build-server-protocol/blob/master/docs/specification.md#build-target-sources-request) 

I also tried with BSP importing a project with a `src/main/resources` but no `src/test/resources` : I end up with the test folder itself being marked as a resource ! 
image

I think it comes from [BspResolverLogic](https://github.com/JetBrains/intellij-scala/blob/idea212.x/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala#L221) that creates either a resource from a file or its own parent 

Edit : I pushed a modification that adds an extra ending slash when missing on resource directories URIs.

adpi2 (approved):

eed3si9n:
Thanks for the contribution. I'll trust @adpi2's review and hit merge since it's been approved.

PR #6553, opened 2021-06-18
Support the SBT extension in BSP import

### LINKED ISSUE #5953: BSP: support sbt extension
The BSP protocol defines an [sbt extension](https://build-server-protocol.github.io/docs/extensions/sbt.html). 

It specifies an `SbtBuildTarget` with sbt-specific information. This is used by IntelliJ to provide the correct highlighting and completions for a given sbt version.

Ideally, a project's meta-project will be returned in a `workspace/buildTargets` response recursively, though for practical purposes one level is usually enough and the current state of the art in the IntelliJ Scala plugin.

retronym:
This enables code assist in the `.sbt` and `project/*.scala` files in IntelliJ (pending merge of the [related PR](https://github.com/JetBrains/intellij-scala/pull/593) to intelli-scala).

Other IDEs that support BSP import could also take advantage of this; scala-metals has a ticket to add support: https://github.com/scalameta/metals-feature-requests/issues/143

To disable this feature (e.g. if using an IDE version that does _not_ yet support the [sbt extension](https://build-server-protocol.github.io/docs/extensions/sbt.html) in the BSP protocl, set `ThisBuild / bspSbtEnabled := false`.

Fixes #5953 

Requires: https://github.com/JetBrains/intellij-scala/pull/593

### DIFF
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -143,14 +143,16 @@ final case class PluginData(
     definitionClasspath: Seq[Attributed[File]],
     resolvers: Option[Vector[Resolver]],
     report: Option[UpdateReport],
-    scalacOptions: Seq[String]
+    scalacOptions: Seq[String],
+    unmanagedSources: Seq[File],
+    managedSources: Seq[File]
 ) {
   val classpath: Seq[Attributed[File]] = definitionClasspath ++ dependencyClasspath
 }
 
 object PluginData {
   private[sbt] def apply(dependencyClasspath: Def.Classpath): PluginData =
-    PluginData(dependencyClasspath, Nil, None, None, Nil)
+    PluginData(dependencyClasspath, Nil, None, None, Nil, Nil, Nil)
 }
 
 object EvaluateTask {
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -144,7 +144,9 @@ final case class PluginData(
     resolvers: Option[Vector[Resolver]],
     report: Option[UpdateReport],
     scalacOptions: Seq[String],
+    unmanagedSourceDirectories: Seq[File],
     unmanagedSources: Seq[File],
+    managedSourceDirectories: Seq[File],
     managedSources: Seq[File]
 ) {
   val classpath: Seq[Attributed[File]] = definitionClasspath ++ dependencyClasspath
@@ -152,7 +154,7 @@ final case class PluginData(
 
 object PluginData {
   private[sbt] def apply(dependencyClasspath: Def.Classpath): PluginData =
-    PluginData(dependencyClasspath, Nil, None, None, Nil, Nil, Nil)
+    PluginData(dependencyClasspath, Nil, None, None, Nil, Nil, Nil, Nil, Nil)
 }
 
 object EvaluateTask {
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -25,7 +25,7 @@ import sbt.internal.inc.ScalaInstance
 import sbt.internal.io.WatchState
 import sbt.internal.librarymanagement.{ CompatibilityWarningOptions, IvySbt }
 import sbt.internal.remotecache.RemoteCacheArtifact
-import sbt.internal.server.BuildServerProtocol.BspWorkspace
+import sbt.internal.server.BuildServerProtocol.BspFullWorkspace
 import sbt.internal.server.{ BuildServerReporter, ServerHandler }
 import sbt.internal.util.{ AttributeKey, ProgressState, SourcePosition }
 import sbt.io._
@@ -398,12 +398,13 @@ object Keys {
 
   val bspConfig = taskKey[Unit]("Create or update the BSP connection files").withRank(DSetting)
   val bspEnabled = SettingKey[Boolean](BasicKeys.bspEnabled)
+  val bspSbtEnabled = settingKey[Boolean]("Should BSP export meta-targets for the SBT build itself?")
   val bspTargetIdentifier = settingKey[BuildTargetIdentifier]("Build target identifier of a project and configuration.").withRank(DSetting)
-  val bspWorkspace = settingKey[BspWorkspace]("Mapping of BSP build targets to sbt scopes").withRank(DSetting)
+  val bspWorkspace = settingKey[Map[BuildTargetIdentifier, Scope]]("Mapping of BSP build targets to sbt scopes").withRank(DSetting)
+  private[sbt] val bspFullWorkspace = settingKey[BspFullWorkspace]("Mapping of BSP build targets to sbt scopes and meta-targets for the SBT build itself").withRank(DSetting)
   val bspInternalDependencyConfigurations = settingKey[Seq[(ProjectRef, Set[ConfigKey])]]("The project configurations that this configuration depends on, possibly transitivly").withRank(DSetting)
   val bspWorkspaceBuildTargets = taskKey[Seq[BuildTarget]]("List all the BSP build targets").withRank(DTask)
   val bspBuildTarget = taskKey[BuildTarget]("Description of the BSP build targets").withRank(DTask)
-  val bspSbtBuildTarget = taskKey[List[BuildTarget]]("Description of the BSP SBT build targets").withRank(DTask)
   val bspBuildTargetSources = inputKey[Unit]("").withRank(DTask)
   val bspBuildTargetSourcesItem = taskKey[SourcesItem]("").withRank(DTask)
   val bspBuildTargetResources = inputKey[Unit]("").withRank(DTask)
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -10,7 +10,6 @@ package sbt
 import java.nio.file.{ Path => NioPath }
 import java.io.File
 import java.net.URL
-
 import lmcoursier.definitions.{ CacheLogger, ModuleMatchers, Reconciliation }
 import lmcoursier.{ CoursierConfiguration, FallbackDependency }
 import org.apache.ivy.core.module.descriptor.ModuleDescriptor
@@ -26,6 +25,7 @@ import sbt.internal.inc.ScalaInstance
 import sbt.internal.io.WatchState
 import sbt.internal.librarymanagement.{ CompatibilityWarningOptions, IvySbt }
 import sbt.internal.remotecache.RemoteCacheArtifact
+import sbt.internal.server.BuildServerProtocol.BspWorkspace
 import sbt.internal.server.{ BuildServerReporter, ServerHandler }
 import sbt.internal.util.{ AttributeKey, ProgressState, SourcePosition }
 import sbt.io._
@@ -399,10 +399,11 @@ object Keys {
   val bspConfig = taskKey[Unit]("Create or update the BSP connection files").withRank(DSetting)
   val bspEnabled = SettingKey[Boolean](BasicKeys.bspEnabled)
   val bspTargetIdentifier = settingKey[BuildTargetIdentifier]("Build target identifier of a project and configuration.").withRank(DSetting)
-  val bspWorkspace = settingKey[Map[BuildTargetIdentifier, Scope]]("Mapping of BSP build targets to sbt scopes").withRank(DSetting)
+  val bspWorkspace = settingKey[BspWorkspace]("Mapping of BSP build targets to sbt scopes").withRank(DSetting)
   val bspInternalDependencyConfigurations = settingKey[Seq[(ProjectRef, Set[ConfigKey])]]("The project configurations that this configuration depends on, possibly transitivly").withRank(DSetting)
   val bspWorkspaceBuildTargets = taskKey[Seq[BuildTarget]]("List all the BSP build targets").withRank(DTask)
   val bspBuildTarget = taskKey[BuildTarget]("Description of the BSP build targets").withRank(DTask)
+  val bspSbtBuildTarget = taskKey[List[BuildTarget]]("Description of the BSP SBT build targets").withRank(DTask)
   val bspBuildTargetSources = inputKey[Unit]("").withRank(DTask)
   val bspBuildTargetSourcesItem = taskKey[SourcesItem]("").withRank(DTask)
   val bspBuildTargetResources = inputKey[Unit]("").withRank(DTask)
--- a/main/src/main/scala/sbt/internal/Load.scala
+++ b/main/src/main/scala/sbt/internal/Load.scala
@@ -1164,12 +1164,16 @@ private[sbt] object Load {
         val prod = (Configurations.Runtime / exportedProducts).value
         val cp = (Configurations.Runtime / fullClasspath).value
         val opts = (Configurations.Compile / scalacOptions).value
+        val managedSrcs = (Configurations.Compile / managedSources).value
+        val unmanagedSrcs = (Configurations.Compile / unmanagedSources).value
         PluginData(
           removeEntries(cp, prod),
           prod,
           Some(fullResolvers.value.toVector),
           Some(update.value),
-          opts
+          opts,
+          unmanagedSrcs,
+          managedSrcs
         )
       },
       scalacOptions += "-Wconf:cat=unused-nowarn:s",
@@ -1225,7 +1229,7 @@ private[sbt] object Load {
     loadPluginDefinition(
       dir,
       config,
-      PluginData(config.globalPluginClasspath, Nil, None, None, Nil)
+      PluginData(config.globalPluginClasspath, Nil, None, None, Nil, Nil, Nil)
     )
 
   def buildPlugins(dir: File, s: State, config: LoadBuildConfiguration): LoadedPlugins =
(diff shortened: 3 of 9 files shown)


main/src/main/scala/sbt/Keys.scala
```diff
@@ -399,10 +399,11 @@ object Keys {
   val bspConfig = taskKey[Unit]("Create or update the BSP connection files").withRank(DSetting)
   val bspEnabled = SettingKey[Boolean](BasicKeys.bspEnabled)
   val bspTargetIdentifier = settingKey[BuildTargetIdentifier]("Build target identifier of a project and configuration.").withRank(DSetting)
-  val bspWorkspace = settingKey[Map[BuildTargetIdentifier, Scope]]("Mapping of BSP build targets to sbt scopes").withRank(DSetting)
+  val bspWorkspace = settingKey[BspWorkspace]("Mapping of BSP build targets to sbt scopes").withRank(DSetting)
```
adpi2: In theory that will break the compatibility between sbt versions. In that particular case, I don't think any build or plugin depends on it directly. So it should not be a problem. But maybe we should take the opportunity to make it private?

retronym: It's not strictly necessary to change this key but it sure seemed the cleanest while prototyping. Happy to take direction though about how to proceed to reduce risk of compatibility breakage.

main/src/main/scala/sbt/Keys.scala
```diff
@@ -399,10 +399,11 @@ object Keys {
...
   val bspInternalDependencyConfigurations = settingKey[Seq[(ProjectRef, Set[ConfigKey])]]("The project configurations that this configuration depends on, possibly transitivly").withRank(DSetting)
   val bspWorkspaceBuildTargets = taskKey[Seq[BuildTarget]]("List all the BSP build targets").withRank(DTask)
   val bspBuildTarget = taskKey[BuildTarget]("Description of the BSP build targets").withRank(DTask)
+  val bspSbtBuildTarget = taskKey[List[BuildTarget]]("Description of the BSP SBT build targets").withRank(DTask)
```
adpi2:
```suggestion
  val bspSbtBuildTargets = taskKey[List[BuildTarget]]("Description of the BSP SBT build targets").withRank(DTask)
```

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -133,9 +151,9 @@ object BuildServerProtocol {
     bspBuildTargetDependencySources / aggregate := false,
     bspBuildTargetCompile := Def.inputTaskDyn {
       val s: State = state.value
-      val workspace = bspWorkspace.value
       val targets = spaceDelimited().parsed.map(uri => BuildTargetIdentifier(URI.create(uri)))
-      val filter = ScopeFilter.in(targets.map(workspace))
+      val workspace = bspWorkspace.value.filter(targets)
+      val filter = ScopeFilter.in(workspace.scopes.values.toList)
```
adpi2, line 189: Would be nice to log a warning if `bspWorkspace.builds` is not empty so that a warn notification is sent to the build client.

Also in scalaMainClasses, scalaTestClasses and test.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -437,6 +483,50 @@ object BuildServerProtocol {
...
+    val scalaJars = scalaProvider.jars()
+    val compileData = ScalaBuildTarget(
+      scalaOrganization = ScalaArtifacts.Organization,
+      scalaVersion = scalaProvider.version(),
+      scalaBinaryVersion = binaryScalaVersion(scalaProvider.version()),
+      platform = ScalaPlatform.JVM,
+      jars = scalaJars.toVector.map(_.toURI.toString)
+    )
+    val workspace = bspWorkspaceSetting.value
+    Def.task {
```
adpi2: This task does not seem to be needed, in which case the `sbtBuildTargetTask` does not need to be dynamic.

adpi2:
Looks very promising so far. Thank you @retronym for taking care of this!

main/src/main/scala/sbt/Keys.scala
eed3si9n: It's used by at least one plugin - https://github.com/scalacenter/scala-debug-adapter/blob/b3125d8a3873a7218e452b256ed938d14c401c5c/sbt/plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/DebugAdapterPlugin.scala#L68, but I am guessing we can make exception for that one.

adpi2: That's right... How did I forget about it? So `bspWorkspace` cannot be private. Also if the signature changes it will be tricky for us, but we can probably work it out anyway.

retronym: I'll rework to avoid a signature change. I can define a new key that aggregates the `Scope` and `LoadedBuildUnit` based targets and leave `val bspWorkspace = settingKey[Map[BuildTargetIdentifier, Scope]]` as a view on the scope part of that data for backwards compatibility.

retronym:
I should get back to this later today, thanks for the review comments so far.

The corresponding change to IntelliJ is just about ready to land: https://github.com/JetBrains/intellij-scala/pull/593

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
retronym: Done.

main/src/main/scala/sbt/Keys.scala
retronym: Applied.

retronym:
This one is ready for another round of review. I'll try to address attaching SBT source JARs to the project in a followup PR.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -698,4 +824,21 @@ object BuildServerProtocol {
...
+      buildToScope: Map[BuildTargetIdentifier, Seq[BuildTargetIdentifier]]
+  ) {
+    def filter(targets: Seq[BuildTargetIdentifier]): BspFullWorkspace = {
+      val set = targets.toSet
+      def filterMap[T](map: Map[BuildTargetIdentifier, T]) = map.filter(x => set.contains(x._1))
+      BspFullWorkspace(filterMap(scopes), filterMap(builds), buildToScope)
+    }
+    def warnIfBuildsNonEmpty(method: String, log: Logger): Unit = {
+      if (builds.nonEmpty)
+        log.warn(s"$method is a no-op for SBT targets: ${builds.keys.mkString("[", ",", "]")}")
```
eed3si9n:
```suggestion
        log.warn(s"$method is a no-op for build.sbt targets: ${builds.keys.mkString("[", ",", "]")}")
```

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -92,36 +97,54 @@ object BuildServerProtocol {
...
+      val filter = ScopeFilter.in(workspace.scopes.values.toList)
       // run the worker task concurrently
       Def.task {
         val items = bspBuildTargetSourcesItem.all(filter).value
-        val result = SourcesResult(items.toVector)
+        val buildItems = workspace.builds.toVector.map {
+          case (id, loadedBuildUnit) =>
+            val base = loadedBuildUnit.localBase
+            val sbtFiles = configurationSources(base)
+            val defDir = projectStandard(base)
```
adpi2: That's not precise enough because the build client would think that the `project/plugins.sbt` and all the files in `project/project/` directories are source files of the main build target, which is wrong.

Also the `project/src/main/scala` belongs to the sbt build target. And the user is even free to add any other source directory with:
```scala
// project/build.sbt
unmanagedSourceDirectories += baseDirectory.value / "custom"
```
Or even he/she can add some generated sources.

Maybe, a good way to fix this is to pass the all sources from the meta-build to the build context through the `PluginData`.

adpi2: In other words, we should put more information in here so that the BSP build target is consistent with the meta-build definition:
https://github.com/sbt/sbt/blob/9274a3e365d3cb826ede31598afde412ab6625c1/main/src/main/scala/sbt/internal/Load.scala#L1163-L1174.

retronym: WIP: https://github.com/sbt/sbt/compare/develop...retronym:topic/bsp-import-plugin-data?expand=1

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
retronym: Applied manually

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
retronym: I've added a commit to this PR that implements your suggestion.

I don't go as far as trying to represent the meta-build (or meta-meta-...-builds!) (`project/*.sbt` / `project/project/*.scala`) as a build target. This _is_ supported by the BSP spec but not yet supported by IntelliJ.

eed3si9n (approved):
LGTM, but I yield the detail review to Adrien.

eed3si9n (dismissed):
LGTM, but I yield the detail review to Adrien.

main/src/main/scala/sbt/EvaluateTask.scala
```diff
@@ -146,12 +146,27 @@ final case class PluginData(
     scalacOptions: Seq[String]
 ) {
   val classpath: Seq[Attributed[File]] = definitionClasspath ++ dependencyClasspath
+
+  // Not a constructor parameter to avoid a binary incompatibility
+  private[this] var _buildSources: BuildSources = BuildSources.Empty
+  private[sbt] def setBuildSourcesData(data: BuildSources): PluginData = {
+    _buildSources = data; this
+  }
+  private[sbt] def buildSourcesData: BuildSources = _buildSources
```
adpi2: @eed3si9n Is it really necessary to preserve the binary compatibility of `PluginData`? How bad would it be to break it in sbt 1.6?

I am asking this because I also want to add more parameters in it for #6566.

retronym: The breakage is just for the constructor, copy, and apply factory methods, not for consumers of the data. I think it is highly unlikely to break existing plugins.

main/src/main/scala/sbt/internal/Load.scala
```diff
@@ -1164,13 +1164,16 @@ private[sbt] object Load {
         val prod = (Configurations.Runtime / exportedProducts).value
         val cp = (Configurations.Runtime / fullClasspath).value
         val opts = (Configurations.Compile / scalacOptions).value
-        PluginData(
+        val managedSrcs = (Configurations.Compile / unmanagedSources).value
+        val unmanagedSrcs = (Configurations.Compile / managedSources).value
```
adpi2: `managedSrcs` and `unmanagedSrcs` are inverted here

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -93,27 +99,50 @@ object BuildServerProtocol {
...
+      val filter = ScopeFilter.in(workspace.scopes.values.toList)
       // run the worker task concurrently
       Def.task {
         val items = bspBuildTargetSourcesItem.all(filter).value
-        val result = SourcesResult(items.toVector)
+        val buildItems = workspace.builds.toVector.map {
+          case (id, loadedBuildUnit) =>
+            val base = loadedBuildUnit.localBase
+            val sbtFiles = configurationSources(base)
+            val buildSources = loadedBuildUnit.unit.plugins.pluginData.buildSourcesData
```
adpi2: `buildSources` is always empty because of this `PluginData.copy` here (the private `_buildSources` is not copied):
https://github.com/sbt/sbt/blob/e4231ac03903e174bc9975ee00d34064a1d1f373/main/src/main/scala/sbt/internal/Load.scala#L1252

adpi2: This would be easier to fix and to maintain if we can break the `PluginData` constructor (@eed3si9n).

adpi2 (changes requested):
The 2 comments below should be addressed before merging. Otherwise I would suggest to update the test here:
https://github.com/sbt/sbt/blob/e4231ac03903e174bc9975ee00d34064a1d1f373/server-test/src/test/scala/testpkg/BuildServerTest.scala#L24-L35
and to add a new test for the sources of the build.sbt target.

main/src/main/scala/sbt/EvaluateTask.scala
eed3si9n: >  Is it really necessary to preserve the binary compatibility of `PluginData`? How bad would it be to break it in sbt 1.6?

We maintain binary compatibility for the purpose of maintaining plugin compatibility, so I think `PluginData` would be ok to break, but ultimately we the only way to quantify this would be to release a milestone and test it using Community Build. If none of the existing builds break, then we should be ok.

retronym:
Update: I'm currently trying to extend the BSP test. Having some difficulty there as one of my new tests for `sources (sbt build)` works in isolation but not after the exising test for normal sources. Haven't figured out the crosstalk yet.

adpi2 (approved):
I just pushed a small fix for the new `buildTarget/resources` request. Otherwise it looks good to me as it is.
I tried in Metals and most of the things (goto definitions and code completions) are already working out-of-the-box. That's cool!

retronym:
@eed3si9n Are you happy with this one now, in particular deeming `PluginData` creation/copying as sbt-internal and excluded from MiMa?

retronym:
There was a scalacheck test failure in one of the test runs, not related to this PR, but maybe worth looking into:

```
[info] - sbt.ParseKey.An unspecified project axis resolves to the current project: Falsified after 4341 passed tests
[info] > StructureKeyMask(Env:
[info]     Tasks:
[info]     aa (delegates: )
[info] Build file:///aaaaaaa/#a :
[info]     Project a
[info]       Delegates:
[info]         
[info]       Configurations:
[info]         A
[info]     Project b
[info]       Delegates:
[info]         
[info]       Configurations:
[info]         A
[info]     Project aa
[info]       Delegates:
[info]         
[info]       Configurations:
[info]         A
[info] current: ProjectRef(file:///aaaaaaa/#a,a)
[info] Settings:
[info] 	Global / <key>
[info] 	   aa
[info] All keys:
[info] 	aa,ScopedKey(Scope(Select(BuildRef(file:///aaaaaaa/#a)), Select(ConfigKey(A)), Select(aa), Zero),aa),ScopeMask(false,false,true,false))
[info] > === Not Equal ===
[info] > --- lhs ---
[info] > Select(ConfigKey(A))
[info] > --- rhs ---
[info] > Zero
[info] > === Not Equal ===
[info] > --- lhs ---
[info] > Select(ProjectRef(file:///aaaaaaa/#a,aa))
[info] > --- rhs ---
[info] > Select(ProjectRef(file:///aaaaaaa/#a,a))
[info] > parsed subproject: Select(ProjectRef(file:///aaaaaaa/#a,aa))
[info] > current subproject: ProjectRef(file:///aaaaaaa/#a,a)
[info] > Key: {file:///aaaaaaa/#a} / A / aa / aa
[info] > Mask: ScopeMask(false,false,true,false)
[info] > Key string: 'aa / aa'
[info] > Parsed: Right(ProjectRef(uri("file:///aaaaaaa/#a"), "aa") / aa)
[info] > Structure: Env:
[info]     Tasks:
[info]     aa (delegates: )
[info] Build file:///aaaaaaa/#a :
[info]     Project a
[info]       Delegates:
[info]         
[info]       Configurations:
[info]         A
[info]     Project b
[info]       Delegates:
[info]         
[info]       Configurations:
[info]         A
[info]     Project aa
[info]       Delegates:
[info]         
[info]       Configurations:
[info]         A
[info] current: ProjectRef(file:///aaaaaaa/#a,a)
[info] Settings:
[info] 	Global / <key>
[info] 	   aa
[info] All keys:
[info] 	aa
```

eed3si9n:
@retronym I'm guessing it's a known ambiguity when we have subproject named `aa` and a task key `aa`, `aa / aa` can be interpreted as "`aa` task in `aa` project", not "`aa` task scoped to `aa` task in current subproject", and thus "An unspecified project axis resolves to the current project" gets falsified. I probably should try to filter that out of the test.

I'm fine with `PluginData`. I guess the question is if we should call this a 1.6.0 feature or 1.5.x. We can publish 1.6.0-M1, run Community Build, and then back port it to 1.5.x if it passes?

retronym:
I did a GitHub code search for "plugindata language:scala", filtered out files with `package sbt`, and didn't spot any usages of the changed methods.

```
$ gh api --paginate -X GET search/code -f q='language:scala plugindata' -q ".items[] | .url" > /tmp/urls.txt
$ for url in $(cat /tmp/urls.txt); do echo $url; code=$(gh api $url  -q ".content" | base64 --decode); echo $code | grep 'package sbt' || echo $code; done | tee /tmp/code.txt
```

Not a perfect method, but I think it is likely safe enough for 1.5.x. @adpi2 mentioned that he also has a followup change to add extra data to `PluginData`.

eed3si9n:
ok. We can merge it now, and see.

PR #6554, opened 2021-06-18
fixes remote caching not managing resource files

### LINKED ISSUE #6394: Remote caching resurfaces resources with old filenames
## Steps

```
$ git clone https://github.com/badeball/sbt-remote-cache-bug
$ cd sbt-remote-cache-bug
```

.. then run the command described in the readme and watch Flyway fail.

```
$ rm -rf remote-cache && \
  find -type d -name target -exec rm -r {} + && \
  docker run \
    --rm \
    --detach \
    --name mssql \
    --publish 1433:1433 \
    --env ACCEPT_EULA=Y \
    --env SA_PASSWORD=Password123 \
      mcr.microsoft.com/mssql/server:2017-CU14-ubuntu && \
  sleep 5s && \
  docker exec mssql \
    /opt/mssql-tools/bin/sqlcmd \
      -S localhost \
      -U SA \
      -P Password123 \
      -Q "CREATE DATABASE foobar_database;"
  git checkout master && \
  sbt compile && \
  sbt pushRemoteCache && \
  find -type d -name target -exec rm -r {} + && \
  git checkout moved-migration && \
  sbt pullRemoteCache && \
  sbt compile && \
  (sbt flywayMigrate || true) && \
  docker stop mssql
```

## Problem

What happens here is that we prime a remote-cache (local directory) with a database migration. Then we checkout another branch where said migration has been renamed and we delete target directories to simulate another developer. We then pull from the cache, compile and run flyway. This fails, because somehow _two_ migration resources end up in a target directory, which isn't going to work.

## Expectation

I _think_ this is an unexpected outcome. It's certainly extremely detrimental to our use case, which is speeding up CI. The use of caching ends up destroying pipelines and requires manual intervention from us.

I think that I expect compile output to always be identical whether a cache is used or not, provided that 1) compiling is deterministic and 2) I trust the cache. Now, I know Scala compilation isn't entirely deterministic, but that's beyond the point here. I still expect the compile output to the _correct / working_, but not necessarily identical.

## Notes

- OS: **Arch Linux**
- Sbt version: **1.4.6**
- Java version: **openjdk 11**

Nirvikalpa108:
## What is the problem? 😔
When using remote caching, the resource files are not tracked so if they have changed, pullRemoteCache will deliver both the old resource file as well as the most recent version.

This is a problem, because it's not the behaviour that our users will expect and it's not in keeping with the contract of this feature.

## Why is this happening?
Zinc, sbt's incremental compiler, keeps track of changes that have been made. It keeps this in what is called the Analysis file. However, resource files are not tracked in the Analysis file, so remote caching is not invalidating the unchanged resource in place of the new one.

## What is the solution? 😃
PullRemoteCache deletes all of the resources files. After this, copyResources is called by PackageBin, which puts the latest
version of the resources back.

fixes #6394


eed3si9n (approved):
Thanks @Nirvikalpa108!

eed3si9n:
I think this fix is good on its own, and is ready to merge.

I've actually thought of another cool way to fix this. When we saw copyResources:

https://github.com/sbt/sbt/blob/4926cb4c2b67d52a6ba0b646275c30735810f038/main/src/main/scala/sbt/Defaults.scala#L2529-L2535

that is sort of doing something like incremental compilation, but it's just not virtualized like this PR does. So... if we take `Sync.sync` (https://github.com/sbt/sbt/blob/4926cb4c2b67d52a6ba0b646275c30735810f038/main-actions/src/main/scala/sbt/Sync.scala) and let it internally manage the list of VirtualFiles, we'd essentially have remote cache ready version of generic file syncing.

PR #6555, opened 2021-06-20
Do not clear line if ends with carriage return

sebastian-alfers:


sebastian-alfers:
@eed3si9n Locally I had to change sbt-scalafmt to something like `2.4.2` - is this expected or a known issue?

eed3si9n:
I don't think that's expected. What happens when you don't? Also are you using Metals or any other plugins?

sebastian-alfers:
It seem I messed up with squashing - will close this PR and start with a clean branch out of current `develop`

Issue #1124, opened 2014-02-17
The semantics of "set every" may be surprising to users


cunei:
The command "set every" redefines the tasks/settings that match the supplied key in all the scopes. Now, consider the following:

```
set every sources in (Compile,doc) := Seq.empty
```

The result is that the scope supplied in the command is silently ignored, and all the sources in every scope are emptied, leading to empty executable jars.

This semantics may be quite surprising where users assume that "set every" just iterates over all the projects; the fact that the supplied scope is silently discarded may be cause for considerable confusion.

Possible approaches: "set every" should either stop if a scope is supplied in the command line, or at least issue a warning. As a more complete solution, "set every" could be changed in order to take some scope filtering into account. If a change in semantics is not viable, an alternate command "set-in-all-projects" could be added in order to supply the alternate semantics that users may expect.

eed3si9n:
To potential contributor: Likely the relevant code base is 
https://github.com/sbt/sbt/blob/v1.1.5/main/src/main/scala/sbt/internal/SettingCompletions.scala#L39-L65 where `setAll` is implemented.

https://github.com/sbt/sbt/blob/v1.1.5/main/src/main/scala/sbt/Main.scala#L408-L431 might also be involved.

AkshatJindal1:
I would like to work on this issue

eed3si9n:
@AkshatJindal1 Awesome!

AkshatJindal1:
I have created a PR where I have added a log statement in case argument list is provided.
https://github.com/sbt/sbt/pull/6560

I am not able to add any reviewer to my PR. @eed3si9n can you have a look into it

SethTisue:
the linked PR appears abandoned

Crystora:
@eed3si9n Can I pick this up?

eed3si9n:
@bittoby Yes please. Please check out our [Contributor's Guildeline](https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md) and:
1. Make sure you can reproduce the problem and the fix manually using locally built `sbt`
2. Include tests for the change
3. Follow our AI assisted contributions guideline (please disclose AI usages in the PR description, if any)
4. Multiple contributors might work on this issue in parallel
5. Generally we recommend getting CI to pass on a forked repo before sending a pull request (pull requests that do not pass tests will be closed after a few days)
6. Sign Scala CLA <https://contribute.akka.io/contribute/cla/scala>

PR #6560, opened 2021-06-22
Added logger statement if argument provided in setAll

AkshatJindal1:
As a part of this PR:

Issue link: https://github.com/sbt/sbt/issues/1124

Added logger statement to print the argumnet list in case of set every command

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -603,7 +603,10 @@ object BuiltinCommands {
         LineRange(0, 0)
       )(cl)
       val setResult =
-        if (all) SettingCompletions.setAll(extracted, settings)
+        if (all) {
+          if (arg == null || arg.trim.isEmpty) s.log.info(s"$arg will be ignored in set All")
+          SettingCompletions.setAll(extracted, settings)
+        }
         else SettingCompletions.setThis(extracted, settings, arg)
       s.log.info(setResult.quietSummary)
       s.log.debug(setResult.verboseSummary)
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:

lightbend-cla-validator:

lightbend-cla-validator:

main/src/main/scala/sbt/Main.scala
```diff
@@ -603,7 +603,10 @@ object BuiltinCommands {
         LineRange(0, 0)
       )(cl)
       val setResult =
-        if (all) SettingCompletions.setAll(extracted, settings)
+        if (all) {
+          if (arg == null || arg.trim.isEmpty) s.log.info(s"$arg will be ignored in set All")
```
eed3si9n: Note that `settings` that is passed in is created using `arg`, so to say arg will be ignored, I don't think is correct. The problem is that when the `set every` takes a scoped  key, the _scoping_ gets ignored.

For example:

```scala
set every version := "1.0.0-M1"
```

this is ok.

```scala
set every Compile / doc / sources := Nil
```

this does not work.

AkshatJindal1: As part of the issue, it was asked to add a log statement in case scope is provided in set every. 
Should I make change to the semantics to use the scope?
@eed3si9n

eed3si9n: What Toni asked in the issue was this:

> Possible approaches: "set every" should either stop if a scope is supplied in the command line, or at least issue a warning.

Could you run sbt with your modification and show what output it currently prints out for `set every version ...` and `set every Compile / doc / sources` please?

SethTisue:
closing for lack of activity. could be reopened if activity resumes

PR #6565, opened 2021-06-24
[BSP] Respond error after random failure when compiling

### LINKED ISSUE #6562: [BSP] - Scalafix compilation failure
## steps

A SBT (1.5.4) / Scala (2.13.6) project with a [scalafix](https://github.com/scalacenter/scalafix) rule that generates a compilation error (with `scalafixOnCompile := true`)

## problem

No information about the error is sent to BSP (and retrieved in Intellij as well). The [CompileResult](https://build-server-protocol.github.io/docs/specification#compile-request) gets a 2 statusCode but no error message is given. 

## expectation

It would be nice to have some information about the failing issue. 

## notes

The [BuildServerProtocol.bspBuildTargetCompile](https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala#L155) does not use the [CompileResult.data](https://build-server-protocol.github.io/docs/specification#compile-request) property, maybe we could add this information in this field. 

I can't understand anyway the difference between a "real" compilation error, that actually returns some information about the error and a scalafix generation error, that does not.

adpi2:
Fixes #6562 reported by @samuelClarencTeads

- When a BSP task fails we send an error response that contains the error message.
- When the `bspCompileItem` catches an exception: if it is a `CompileFailed` we respond with a `StatusCode.Error` result, if it is an `InterruptedException` we respond with a `StatusCode.Cancelled` result, otherwise we send an error response.
- The old error response with a `-33000` code and no message is made silent if another response has already been sent.
- Add a server test to check the error message on a custom compilation failure

### DIFF
--- a/main/src/main/scala/sbt/EvaluateTask.scala
+++ b/main/src/main/scala/sbt/EvaluateTask.scala
@@ -9,7 +9,6 @@ package sbt
 
 import java.io.File
 import java.util.concurrent.atomic.AtomicReference
-
 import sbt.Def.{ ScopedKey, Setting, dummyState }
 import sbt.Keys.{ TaskProgress => _, name => _, _ }
 import sbt.Project.richInitializeTask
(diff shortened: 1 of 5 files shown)


eed3si9n (approved):
Thanks @adpi2!

PR #6576, opened 2021-07-01
[BSP] Fix failing import 

github-samuel-clarenc:
## Steps : 

- With SBT 1.5.3, import a project on Intellij IDEA **without** a started BSP server (meaning no `sbt -bsp` was done on a detached terminal). 
-  java is not defined in `/usr/bin/java` (for example when using [sdkman](https://sdkman.io/))

## Problem : 

The import fails, with a `java` not found message.

## Expected : 

The project should be successfully imported. 

### Solution : 

This PR adds the `java_home` path to the process builder instead of using directly the `java` command.

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -342,7 +342,12 @@ class NetworkClient(
                 s"either upgrade $sbtScript to its latest version or make sure it is accessible from $$PATH, and run 'sbt bspConfig'"
               )
             }
-            List("java") ++ arguments.sbtArguments ++
+            val java = Option(Properties.javaHome)
+              .map { javaHome =>
+                s"$javaHome/bin/java"
+              }
+              .getOrElse("java")
+            List(java) ++ arguments.sbtArguments ++
               List("-jar", lj, DashDashDetachStdio, DashDashServer)
           case _ =>
             List(arguments.sbtScript) ++ arguments.sbtArguments ++


main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -330,7 +330,7 @@ class NetworkClient(
 
         val cmd = arguments.sbtLaunchJar match {
           case Some(lj) =>
-            List("java") ++ arguments.sbtArguments ++
+            List(s"${Properties.javaHome}/bin/java") ++ arguments.sbtArguments ++
```
adpi2: The `Properties.javaHome` is probably null when using the native client (sbtn) so it is probably safer to check for null.

```suggestion
            val java = if (Properties.javaHome != null) s"${Properties.javaHome}/bin/java" else "java"
            List(java) ++ arguments.sbtArguments ++
```

Also it is worth noting that https://github.com/sbt/sbt/pull/6593 makes this branch of the match statement less useful and we would rather prefer to use the sbt script.

adpi2 (commented):
Thanks again for your contribution.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
github-samuel-clarenc: OK thanks for your feedback, maybe we can close this PR if yours is already fixing the problem. 

_Your suggestion is pushed anyway !_

adpi2 (dismissed):

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
adpi2: This is still useful when someone uses IntelliJ and does not have java nor sbt installed.

adpi2:
@github-samuel-clarenc could you rebase or merge to resolve conflicts? Thanks.

adpi2 (approved):

Issue #6476, opened 2021-04-27
Ability to configure jvm options for bsp server which is run over bsp/sbt.json


Alexey-NM:
Currently, a large project will not compile in IntelliJ IDEA if sbt server is not run manually from command line. 

https://discuss.lightbend.com/t/sbt-proposal-ability-to-configure-jvm-options-for-bsp-server-which-is-run-over-bsp-sbt-json/8115

adpi2:
Thank you for reporting this! It seems indeed that the `SBT_OPTS`, `JVM_OPTS` env variables or even the `.jvmopts` file are not taken into account by the `.bsp/sbt.json` command.

adpi2:
Fixed by #6593

Issue #6468, opened 2021-04-26
sbt --client ignores other flags or SBT_OPTS


laughedelic:
## steps

1. install official sbt launcher from GitHub releases
2. run `sbt --client` with any other CLI flags

## problem

Other CLI flags are ignored. Examples:

```
►  ./sbt/bin/sbt --error --no-colors --client 'print version'
►  SBT_OPTS='-Dsbt.log.noformat=true' ./sbt/bin/sbt --client 'print version'
►  env SBT_OPTS='-Dsbt.color=false' ./sbt/bin/sbt --client 'print version'
```

All above commands print with colors ignoring any attempts to disable it, here is a screenshot:

<img width="501" src="https://user-images.githubusercontent.com/766656/116014790-4b335700-a637-11eb-8151-87ba4a8afe96.png">

You can also see that the `--error` (or `-error`) flag gets ignored, so I get the undesired logging.

I've tried passing arguments in different order and via `SBT_OPTS`, but it seems that the launcher just ignores anything when there is `--client` flag. Same happens with the `sbtn` native client.

## expectation

In this concrete use case I would like to be able to print the `version` setting value without _any_ logs or color codes. 

In general, I think sbt should treat arguments in the client mode same way as in the batch mode:

<img width="486" alt="Screenshot 2021-04-26 at 02 42 15" src="https://user-images.githubusercontent.com/766656/116015305-03adca80-a639-11eb-8e2b-c6ba3e0d77c5.png">

## notes

```
►  ./sbt/bin/sbt -version
sbt version in this project: 1.5.0
sbt script version: 1.5.0
```

I know that the launcher script treats "client mode" specially, but I think this functionality is essential for using sbt in scripts and integrating with other CLI tools. Running sbt in batch mode several times is too slow.

eed3si9n:
@laughedelic Thanks for the report!

adpi2:
This bug looks very similar to https://github.com/sbt/sbt/issues/6476

adpi2:
First discovery: `--client` is consumed by the sbt script whereas `-client` is consumed by the java process. It seems things are working better with `-client`:


![image](https://user-images.githubusercontent.com/13123162/116536008-bc4b6680-a8e4-11eb-848b-6fea5ac2074b.png)

eed3si9n:
I think `--client` is sbtn but `-client` might be the older thin client, which is a bug on its own - https://github.com/sbt/sbt/issues/6078

adpi2:
[...]

PR #6580, opened 2021-07-04
honor custom JVM options when client autostarts the server with JAR 

### LINKED ISSUE #6575: .sbtopts doesn't get picked up when sbt server gets started by Metals
If I manually start sbt and then start vscode everything works fine. But if I start vscode first and Metals starts the sbt server, the JVM options from the `.sbtopts` file in my project folder don't seem be used.

This is the _.bsp/sbt.json_ file that sbt generates:
```
{"name":"sbt","version":"1.5.2","bspVersion":"2.0.0-M5","languages":["scala"],"argv":["/Users/jasper.moeys/.sdkman/candidates/java/11.0.11.hs-adpt/bin/java","-Xms100m","-Xmx100m","-classpath","/Users/jasper.moeys/.sdkman/candidates/sbt/1.5.1/bin/sbt-launch.jar","xsbt.boot.Boot","-bsp","--sbt-launch-jar=/Users/jasper.moeys/.sdkman/candidates/sbt/1.5.1/bin/sbt-launch.jar"]}
```

This is on OS X. I think this did work on Ubuntu a while ago, but then that was with different versions of sbt, vscode and metals.

@tgodzik

bjaglin:
Includes https://github.com/sbt/sbt/pull/6576
Fixes https://github.com/sbt/sbt/issues/6575 & https://github.com/sbt/sbt/issues/6476, and potentially https://github.com/sbt/sbt/issues/6468

I couldn't find any way to test this with the current test suite, did I miss something?

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -31,7 +31,7 @@ import sjsonnew.shaded.scalajson.ast.unsafe.{ JObject, JValue }
 import sjsonnew.support.scalajson.unsafe.Converter
 
 import scala.annotation.tailrec
-import scala.collection.mutable
+import scala.collection.{ JavaConverters, mutable }
 import scala.concurrent.duration._
 import scala.util.control.NonFatal
 import scala.util.{ Failure, Properties, Success, Try }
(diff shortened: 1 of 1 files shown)


main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -328,13 +328,39 @@ class NetworkClient(
...
+                .map(_.trim)
+                .filterNot(_.startsWith("#"))
+                .filter(_.nonEmpty)
+
+            val jvmOpts =
+              loadConfigFile(file(".jvmopts"))
+            val sbtOpts =
+              loadConfigFile(file("/etc/sbt/sbtopts")) ++ loadConfigFile(file(".sbtopts"))
+
+            val javaOpts = jvmOpts ++ sbtOpts.collect {
```
bjaglin, line 347: it looks like the `.jvmopts` flags are passed both as CLI args and in `JAVA_HOME` in the script, not sure why (`JAVA_HOME` should be enough?) but I replicated that behavior
https://github.com/sbt/sbt/blob/e4231ac03903e174bc9975ee00d34064a1d1f373/sbt#L739
https://github.com/sbt/sbt/blob/e4231ac03903e174bc9975ee00d34064a1d1f373/sbt#L746

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -328,13 +328,39 @@ class NetworkClient(
...
+                .filterNot(_.startsWith("#"))
+
+            val jvmOpts =
+              loadConfigFile(file(".jvmopts"))
+            val sbtOpts =
+              loadConfigFile(file("/etc/sbt/sbtopts")) ++ loadConfigFile(file(".sbtopts"))
+
+            val javaOpts = jvmOpts ++ sbtOpts.collect {
+              case jvmOpt if jvmOpt.startsWith("-X") || jvmOpt.startsWith("-D") => jvmOpt
+              case prefixedJvmOpt if prefixedJvmOpt.startsWith("-J-X")          => prefixedJvmOpt.drop(2)
```
bjaglin, line 349: this will silently ignore options like `-debug` or `--supershell` which are valid `.sbtopts` options, but since some of them are handled in the shell script itself, I figured out I would focus on JVM options first.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -328,13 +328,42 @@ class NetworkClient(
...
+            def loadConfigFile(file: File): List[String] =
+              IO.readLines(file)
+                .map(_.trim)
+                .filter(_.nonEmpty)
+                .filterNot(_.startsWith("#"))
+
+            val jvmOpts =
+              loadConfigFile(file(".jvmopts"))
+            val sbtOpts =
+              sys.env.get("SBT_OPTS").toSeq.flatMap(_.split(" ").map(_.trim)) ++
```
bjaglin, line 342: Ii's hard to reproduce shell globbing behavior, so this will incorrectly split `SBT_OPTS=--sbt-dir /dir`, but as the sbt script flags take only path or integer params, they should not pass through the shortlist of prefixes below :sweat_smile:

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
bjaglin: Looking at the 2 snippets above, I realized there is one difference in this implementation compared to the script: the `JAVA_OPTS` parameters are not explicitly passed as CLI `java` args here (but they are in the script). Again, I don't think that has any effect, unless we are dealing with a JVM that does not honor `JAVA_OPTS`? But in that case, is it sound to turn the flags there into CLI args anyway?

adpi2:
Thank you very much @bjaglin for implementing this solution.

Some [comments](https://github.com/sbt/sbt/issues/6575#issuecomment-874229382) from @eed3si9n made me realize that the sbt script is the best way to start the sbt server consistently (in terms of java and sbt options). That's mostly because, as you have pointed out, many `.sbtopts` are handled in the script.

 #6593 intends to make sbt script used as often as possible. So basically it would make the current PR not needed. I think it's better this way also because we don't want to duplicate all the sbt script logic in the `NetworkClient`.

bjaglin:
@adpi2 makes sense, thanks for looking into that. This was my initial idea as well, but I was concerned about the latency in upgrading sbt scripts (official, sbt-extras, etc) so that they advertize their absolute path. I see this is handled https://github.com/sbt/sbt/pull/6593/files#diff-609c1efadf9d1085a150c9277e7b2e0eab6eb2411d9da6b92e28b9810406ab2dR59-R64, so I am closing this and will happily test your PR and follow-up there.

PR #6593, opened 2021-07-12
Use `-Dsbt.script` to start sbt server

### LINKED ISSUE #6575: .sbtopts doesn't get picked up when sbt server gets started by Metals

adpi2:
In order to start the sbt server we must use the sbt script because it is the only way to load the .sbtopts and the .jvmopts files properly.

To do so the sbt script can pass a `-Dsbt.script` option to the Java process:
- If the Java process is an sbt client: the value of `-Dsbt.script` is used to start the server
- If the Java process is an sbt server: the value of `-Dsbt.script`  is written into the `.bsp/sbt.json` file

Example of `bsp/sbt.json` file:

```json
{
  "name":"sbt",
  "version":"1.5.5-SNAPSHOT",
  "bspVersion":"2.0.0-M5",
  "languages":["scala"],
  "argv":[
    "/home/.sdkman/candidates/sbt/current/bin/sbt",
    "-bsp",
    "-Dsbt.script=/home/.sdkman/candidates/sbt/current/bin/sbt"
  ]
}
```

This PR fixes #6575 and #6476, and part of #6468.
This PR makes #6580 not necessary (except if IntelliJ starts sbt using the sbt-launch.jar directly).

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -330,6 +330,18 @@ class NetworkClient(
 
         val cmd = arguments.sbtLaunchJar match {
           case Some(lj) =>
+            if (log) {
+              val sbtScript = if (Properties.isWin) "sbt.bat" else "sbt"
+              console.appendLog(Level.Warn, s"server is started using sbt-launch jar directly")
+              console.appendLog(
+                Level.Warn,
+                "this is not the recommended way: .sbtopts and .jvmopts files are not loaded and SBT_OPTS is ignored"
+              )
+              console.appendLog(
+                Level.Warn,
+                s"either upgrade $sbtScript to its latest version or make sure it is accessible from $$PATH, and run 'sbt bspConfig'"
+              )
+            }
             List("java") ++ arguments.sbtArguments ++
               List("-jar", lj, DashDashDetachStdio, DashDashServer)
           case _ =>
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -1045,7 +1045,8 @@ object NetworkClient {
   private[client] val noStdErr = "--no-stderr"
   private[client] val sbtBase = "--sbt-base-directory"
   private[client] def parseArgs(args: Array[String]): Arguments = {
-    var sbtScript = if (Properties.isWin) "sbt.bat" else "sbt"
+    val defaultSbtScript = if (Properties.isWin) "sbt.bat" else "sbt"
+    var sbtScript = Properties.propOrNone("sbt.script")
     var launchJar: Option[String] = None
     var bsp = false
     val commandArgs = new mutable.ArrayBuffer[String]
@@ -1067,11 +1068,10 @@ object NetworkClient {
           sbtScript = a
             .split("--sbt-script=")
             .lastOption
-            .map(_.replaceAllLiterally("%20", " "))
-            .getOrElse(sbtScript)
+            .orElse(sbtScript)
         case "--sbt-script" if i + 1 < sanitized.length =>
           i += 1
-          sbtScript = sanitized(i).replaceAllLiterally("%20", " ")
+          sbtScript = Some(sanitized(i))
         case a if a.startsWith("--sbt-launch-jar=") =>
           launchJar = a
             .split("--sbt-launch-jar=")
@@ -1091,12 +1091,17 @@ object NetworkClient {
     }
     val base = new File("").getCanonicalFile
     if (!sbtArguments.contains("-Dsbt.io.virtual=true")) sbtArguments += "-Dsbt.io.virtual=true"
+    if (!sbtArguments.exists(_.startsWith("-Dsbt.script"))) {
+      sbtScript.foreach { sbtScript =>
+        sbtArguments += s"-Dsbt.script=$sbtScript"
+      }
+    }
     new Arguments(
       base,
       sbtArguments.toSeq,
       commandArgs.toSeq,
       completionArguments.toSeq,
-      sbtScript,
+      sbtScript.getOrElse(defaultSbtScript).replaceAllLiterally("%20", " "),
       bsp,
       launchJar
     )
--- a/protocol/src/main/scala/sbt/internal/bsp/BuildServerConnection.scala
+++ b/protocol/src/main/scala/sbt/internal/bsp/BuildServerConnection.scala
@@ -7,11 +7,12 @@
 
 package sbt.internal.bsp
 
-import java.io.File
-
-import sbt.internal.bsp
+import sbt.internal.bsp.codec.JsonProtocol.BspConnectionDetailsFormat
 import sbt.io.IO
 import sjsonnew.support.scalajson.unsafe.{ CompactPrinter, Converter }
+
+import java.io.File
+import java.nio.file.{ Files, Paths }
 
 object BuildServerConnection {
   final val name = "sbt"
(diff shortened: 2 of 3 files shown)


eed3si9n (dismissed):

protocol/src/main/scala/sbt/internal/bsp/BuildServerConnection.scala
```diff
@@ -21,26 +22,45 @@ object BuildServerConnection {
...
+      .map(_.replaceAllLiterally("%20", " "))
+      .orElse(sbtScriptInPath) match {
+      case Some(sbtScript) =>
+        Vector(sbtScript, "-bsp", s"-Dsbt.script=${sbtScript.replaceAllLiterally(" ", "%20")}")
+      case None =>
+        // IntelliJ can start sbt even if the sbt script is not accessible from $PATH.
+        // To do so it uses its own bundled sbt-launch.jar.
+        // In that case, we must pass the path of the sbt-launch.jar to the BSP connection
+        // so that the server can be started.
+        // A known problem in that situation is that the .sbtopts and .jvmopts are not loaded.
```
bjaglin: considering the number of bug reports, would it be worth adding a user-actionnable warn log when we hit this code path? probably not here since it's not a problem (yet), but maybe in https://github.com/sbt/sbt/blob/d4162cce04615e4267b0582d8a0f01170f3ba5af/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala#L332-L334

protocol/src/main/scala/sbt/internal/bsp/BuildServerConnection.scala
```diff
@@ -21,26 +22,45 @@ object BuildServerConnection {
...
-        "-classpath",
-        classPath,
-        "xsbt.boot.Boot",
-        "-bsp"
-      ) ++ sbtLaunchJar.map(jar => s"--sbt-launch-jar=$jar")
+    val argv = Option(System.getProperty("sbt.script"))
+      .map(_.replaceAllLiterally("%20", " "))
+      .orElse(sbtScriptInPath) match {
+      case Some(sbtScript) =>
+        Vector(sbtScript, "-bsp", s"-Dsbt.script=${sbtScript.replaceAllLiterally(" ", "%20")}")
```
bjaglin: This will pick up the potential `-Xms` from `.jvmopts`/`.sbtopts`, which can be significant. Can we force a low footprint for the sbt client just like below/before?

Or maybe only forward `-Dsbt.script` for the potential server autostart but keep starting the client with `java --classpath` since `.sbtopts`/`.jvmopts` are irrelevant? `.jvmopts` is probably questionnable though, I guess it's the open discussion in https://github.com/sbt/sbt/issues/6468: which launcher flags should clients pick up?

protocol/src/main/scala/sbt/internal/bsp/BuildServerConnection.scala
```diff
@@ -21,26 +22,45 @@ object BuildServerConnection {
...
     val json = Converter.toJson(details).get
     IO.write(bspConnectionFile, CompactPrinter(json), append = false)
   }
+
+  private def sbtScriptInPath: Option[String] = {
+    // For those who use an old sbt script, the -Dsbt.script is not set
+    // As a fallback we try to find the sbt script in $PATH
+    val envPath = Option(System.getenv("PATH")).getOrElse("")
+    val allPaths = envPath.split(File.pathSeparator).map(Paths.get(_))
+    allPaths.map(_.resolve("sbt")).find(Files.exists(_)).map(_.toString)
```
bjaglin: I am not familiar with Windows where this could be a problem, but could false positives be avoided through more heuristics like `isFile` & `isExecutable`?

bjaglin (commented):
Thanks for the PR, I successfully tested it on my laptop for our enteprise use case: we want BSP clients to honor the `-Dsbt.repository.config` / `-Dsbt.override.build.repos` flags from `.sbtopts` to get a safe, deterministic build.

protocol/src/main/scala/sbt/internal/bsp/BuildServerConnection.scala
adpi2: For the BSP client, I think that all launcher flags are irrelevant. So using `java --classpath` here should not be a problem.

protocol/src/main/scala/sbt/internal/bsp/BuildServerConnection.scala
adpi2: And even more the windows script is `sbt.bat`...

protocol/src/main/scala/sbt/internal/bsp/BuildServerConnection.scala
adpi2: Added a warning at server startup.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -330,6 +330,18 @@ class NetworkClient(
...
+            if (log) {
+              val sbtScript = if (Properties.isWin) "sbt.bat" else "sbt"
+              console.appendLog(Level.Warn, s"server is started using sbt-launch jar directly")
+              console.appendLog(
+                Level.Warn,
+                "this is not the recommended way: .sbtopts and .jvmopts files are not loaded"
+              )
+              console.appendLog(
+                Level.Warn,
+                s"make sure $sbtScript is accessible from $$PATH and run 'sbt bspConfig'"
```
bjaglin: I thought this was not specific to BSP and was about to suggest a more generic message, but I guess it is, effectively. I tried to run `sbtn` without `sbt` in the path, and in that case the process just dies without any logs, most likely because the default case below is evaluated and the script it not found.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -330,6 +330,18 @@ class NetworkClient(
 
         val cmd = arguments.sbtLaunchJar match {
           case Some(lj) =>
+            if (log) {
+              val sbtScript = if (Properties.isWin) "sbt.bat" else "sbt"
+              console.appendLog(Level.Warn, s"server is started using sbt-launch jar directly")
+              console.appendLog(
+                Level.Warn,
+                "this is not the recommended way: .sbtopts and .jvmopts files are not loaded"
```
bjaglin:
```suggestion
                "this is not the recommended way: .sbtopts and .jvmopts files are not loaded and SBT_OPTS is ignored"
```

bjaglin (dismissed):
:bow:

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
adpi2: `--sbt-launch-jar` is only usefull for IntelliJ to be able to connect with BSP using its own bundled jar. Otherwise it would have probably never been added.

bjaglin: I see, https://github.com/sbt/sbt/pull/6177/files shows indeed that it was adhoc (and not documented).

bjaglin: nitpicking, but I guess upgrading the sbt script to 1.6.x or 1.5.6 (not sure if this is meant to be backported?) is the preferred recommendation?
```suggestion
                s"either upgrade $sbtScript or make sure it is accessible from $$PATH, and run 'sbt bspConfig'"
```

eed3si9n (approved):

PR #6605, opened 2021-07-24
Building on #6353 to help identify the issue of #6101 for Windows users.

ScalaWilliam:
Also make it clear what 110 means.

Before:
image

After:
image

Now can identify which place the original exception comes from.


eed3si9n (approved):
Thanks!

PR #6615, opened 2021-08-08
Problem in installing Sbt for macOS12

0masn:
I am facing problem in installing SBT using my macOs12 with brew.


eed3si9n:
> I am facing problem in installing SBT using my macOs12 with brew.

I generally do not recommend using Home Brew since they seem to do their own thing with Java - https://www.scala-sbt.org/download.html. Personally I use SDKMAN, which works well for me. And what's that to do with this pull request again?

Issue #6625, opened 2021-08-18
Add launcher argument to set sbt.global.localcache property


ericdotdata:
## Description

We have a few convenience launcher arguments like:
```
  --sbt-dir   <path>  path to global settings/plugins directory (default: ~/.sbt)
  --sbt-boot  <path>  path to shared boot directory (default: ~/.sbt/boot in 0.11 series)
  --ivy       <path>  path to local Ivy repository (default: ~/.ivy2)
```

I propose adding `--sbt-cache` that would default to `~/.cache/sbt` to use that argument and/or property.

eed3si9n:
As a retronym I'd suggest `--disk-cache` to mean not-remote cache, which is the name Bazel also uses.

fury0928:
Hi @eed3si9n good morning
May I work on this issue?
Thank you for your time

eed3si9n:
I used to think this is a neat idea but given that we already have many config files like build.sbt itself, `.jvmopts`, and `.sbtopts`, adding yet another configuration file actually would cause more maintenance load and confusion.

mzuehlke:
This issue was solved in
* #6629

PR #6629, opened 2021-08-19
Add --sbt-cache argument and update localCacheDirectory key description

ericdotdata:
# Description

This PR addresses https://github.com/sbt/sbt/issues/6625 and also "resolves" some conflation of the usage/terminology of the `localCacheDirectory: settingKey[File]`.

The original description was: `Directory to pull the remote cache to.`
The proposed change is: `Operating system specific cache directory.`

This is a similar concept to "disk-cache" in bazel, and while the initial usage of `localCacheDirectory` in `sbt` was for the remote cache support, the implementation in `SysProp` (https://github.com/sbt/sbt/blob/fe5497938a8b2898dcbab49b60e0f2ef53b06c29/main/src/main/scala/sbt/internal/SysProp.scala#L188) is more consistent with the "disk-cache" term.

The `Remote Cache` functionality is just one of many possible consumers of `localCacheDirectory` (e.g. `sbt-assembly` could use this), so I believe the updated description is more consistent, flexible, and less confusing. 

### TODO
~~- [x] Add `cache_dir ()` helper~~
- [x] Add tests

### DIFF
--- a/sbt
+++ b/sbt
@@ -538,6 +538,7 @@ Usage: `basename "$0"` [options]
   --sbt-create        start sbt even if current directory contains no sbt project
   --sbt-dir   <path>  path to global settings/plugins directory (default: ~/.sbt)
   --sbt-boot  <path>  path to shared boot directory (default: ~/.sbt/boot in 0.11 series)
+  --sbt-cache <path>  path to global cache directory (default: operating system specific)
   --ivy       <path>  path to local Ivy repository (default: ~/.ivy2)
   --mem    <integer>  set memory options (default: $sbt_default_mem)
   --no-share          use all local caches; no sharing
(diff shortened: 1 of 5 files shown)


sbt
```diff
@@ -128,7 +129,10 @@ acquire_sbt_jar () {
     fi
   fi
   local user_home && user_home=$(findProperty user.home)
-  download_jar="${user_home:-$HOME}/.cache/sbt/boot/sbt-launch/$launcher_sv/sbt-launch-$launcher_sv.jar"
+  local download_jar="${user_home:-$HOME}/.cache/sbt/boot/sbt-launch/$launcher_sv/sbt-launch-$launcher_sv.jar"
+  if [[ -d "$sbt_cache" ]]; then
+    download_jar="${sbt_cache}/boot/sbt-launch/$launcher_sv/sbt-launch-$launcher_sv.jar"
+  fi
```
eed3si9n: Could we define a Bash function called `cache_dir ()` or something that encapsulates this logic here plz?

sbt
```diff
@@ -648,6 +658,10 @@ process_args () {
              -batch|--batch) exec </dev/null && shift ;;
 
          -sbt-jar|--sbt-jar) require_arg path "$1" "$2" && sbt_jar="$2" && shift 2 ;;
+     -sbt-cache|--sbt-cache) require_arg path "$1" "$2" &&
+                             sbt_cache="$2" &&
+                             addJava "-Dsbt.global.base=$2" &&
```
er1c, line 663: typo, i'll also fix the launcher package tests

sbt
ericdotdata: This whole different download/cache wasn't needed, these are always `~/.sbt/boot`, so its just the setting that needed to be updated.

ericdotdata:
Test added, rebased against current `develop`

sbt
er1c: oh snap, technically these are suppose to be `--sbt-boot` maybe lol, different issue though.

eed3si9n (approved):

Issue #6631, opened 2021-08-21
launcher script should use --sbt-boot instead of ~/.cache/sbt/boot


er1c:
## Description

There are a handful of hardcoded `${user_home:-$HOME}/.cache/sbt/boot` references in the launcher script e.g.:
https://github.com/sbt/sbt/blob/a199875e1dcb16cd92a439b47379954daff1a995/sbt#L131

However, I think we should respect the the existing argument `--sbt-boot` and its related property `sbt.boot.directory`:
https://github.com/sbt/sbt/blob/a199875e1dcb16cd92a439b47379954daff1a995/sbt#L540

## Expectation

The `sbt-launch.jar` and `sbt` scripts should both use the same default `sbt.boot.directory` settings.

PR #6632, opened 2021-08-21
[#6631] Enable launcher script use --sbt-boot 

er1c:
This addresses #6631 to use `--sbt-boot` directory to download `sbt-launch.jar` or `sbtn`

Also added some minor `SbtScriptTest` output cleanup.

### DIFF
--- a/sbt
+++ b/sbt
@@ -18,6 +18,7 @@ declare sbt_default_mem=1024
 declare -r default_sbt_opts=""
 declare -r default_java_opts="-Dfile.encoding=UTF-8"
 declare sbt_verbose=
+declare sbt_boot=
 declare sbt_debug=
 declare build_props_sbt_version=
 declare use_sbtn=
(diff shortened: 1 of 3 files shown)


er1c:
Looks like I broke `--sbt-jar` ooops

sbt
```diff
@@ -620,7 +644,6 @@ map_args () {
        -no-share|--no-share) options=( "${options[@]}" "${noshare_opts[@]}" ) && shift ;;
      -no-global|--no-global) options=( "${options[@]}" "-Dsbt.global.base=$(pwd)/project/.sbtboot" ) && shift ;;
                  -ivy|--ivy) require_arg path "$1" "$2" && options=( "${options[@]}" "-Dsbt.ivy.home=$2" ) && shift 2 ;;
-       -sbt-boot|--sbt-boot) require_arg path "$1" "$2" && options=( "${options[@]}" "-Dsbt.boot.directory=$2" ) && shift 2 ;;
```
er1c: I'm not quite sure what the difference between `map_args` and `process_args` is, but trying to set the `sbt_boot` var here didn't work for some reason

PR #6638, opened 2021-08-27
bsp: Implementing `buildTarget/cleanCache`

### LINKED ISSUE #6569: BSP: support workspace/cleanCache
The [`workspace/cleanCache`](https://build-server-protocol.github.io/docs/specification.html#clean-cache-request) request is currently not supported. This is used in IntelliJ for the "rebuild" action.

It appears also instead of replying with a `-32601` (Method not found) error, the server just doesn't reply at all to this request.

hmemcpy:
Fixes #6569

This command is invoked by IntelliJ whenever a Rebuild operation is called. Unfortunately, because it's not implemented, this causes bsp-imported projects in IntelliJ to hang when asking to rebuild.

It works fine in bloop-imported projects because bloop implements this API.


lightbend-cla-validator:
Hi @hmemcpy,

hmemcpy:
@adpi2 hello, sorry for the ping. Unfortunately, this isn't actually working, and I really have no idea why :(
The test passes, but when I actually try to test this in a local test project, when I call `cleanCache` from IntelliJ (or even send the json payload via `sbt -bsp`) I immediately get a `shutdown`.

I really don't know what I'm missing... any tips appreciated!

hmemcpy:
Update! I got it working... somehow. I made sure to clean everything up, do a `publishLocal`, followed by a `"reboot dev"` in my test project... and voilla!
Will finalize the PR soon.
image

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -186,6 +186,23 @@ object BuildServerProtocol {
...
     bspBuildTargetTest := bspTestTask.evaluated,
     bspBuildTargetTest / aggregate := false,
+    bspBuildTargetCleanCache := Def.inputTaskDyn {
+      val s: State = state.value
+      val targets = spaceDelimited().parsed.map(uri => BuildTargetIdentifier(URI.create(uri)))
+      val workspace = bspFullWorkspace.value.filter(targets)
+      workspace.warnIfBuildsNonEmpty(Method.Compile, s.log)
+      val filter = ScopeFilter.in(workspace.scopes.values.toList)
+      Def.task {
+        val statusCodes = Keys.bspBuildTarget.result.all(filter).value
```
hmemcpy: I'm not sure about this one... do I need to add my own item?

adpi2:
@hmemcpy Thanks for pinging me! It is great that you are working on it.

I tested it locally and it seems to work. Make sure you clean the `.sbt` cache after running `publishLocal` with:
```
rm -Rf ~/.sbt/boot/scala-2.12.14/org.scala-sbt/sbt/1.5.5-SNAPSHOT
```
Otherwise the launcher loads the old cached jars rather than the new ones.

Another way to test the `buildTarget/cleanCache` request is to add a test in [BuildServerTest](https://github.com/sbt/sbt/blob/develop/server-test/src/test/scala/testpkg/BuildServerTest.scala).

You can run it with:
```scala
sbt:sbt> serverTestProj/testOnly testpkg.BuildServerTest
```

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
adpi2: I don't think you need an item here. You may be able to use `clean` directly:

```scala
Keys.clean.result.all(filter).value
```

hmemcpy:
@adpi2 I guess initially I had old jars in the cache. I'll test this a bit more and will mark the PR as done. Thanks again for the help!

hmemcpy:
And it's good to go!

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -186,6 +186,23 @@ object BuildServerProtocol {
...
     bspBuildTargetTest := bspTestTask.evaluated,
     bspBuildTargetTest / aggregate := false,
+    bspBuildTargetCleanCache := Def.inputTaskDyn {
+      val s: State = state.value
+      val targets = spaceDelimited().parsed.map(uri => BuildTargetIdentifier(URI.create(uri)))
+      val workspace = bspFullWorkspace.value.filter(targets)
+      workspace.warnIfBuildsNonEmpty(Method.Compile, s.log)
+      val filter = ScopeFilter.in(workspace.scopes.values.toList)
+      Def.task {
+        val statusCodes = Keys.clean.result.all(filter).value
```
adpi2: I think `result` is a better name than `statusCodes` because `statusCodes` generally means there is an `Int` somewhere describing the output of the task. But here the type is `Seq[Result[Unit]]`.

```suggestion
        val results = Keys.clean.result.all(filter).value
```

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -186,6 +186,23 @@ object BuildServerProtocol {
...
+        val statusCode = allOrThrow(statusCodes) match {
+          case Seq() => false
+          case _     => true
+        }
```
adpi2: The protocol is not really clear but I don't think we should throw an exception if only one build target fails. So I suggest this instead.

```scala
      val result = anyOrThrow(statusCodes).size == targets.size
```

If all the targets fail we throw an exception otherwise we check that the number of successful targets is equal to the number of targets in the request.

If `result` is false, that means there is at least one target that failed. If it is true everything is okay.

server-test/src/test/scala/testpkg/BuildServerTest.scala
```diff
@@ -108,6 +108,18 @@ object BuildServerTest extends AbstractServerTest {
     })
   }
 
+  test("buildTarget/cleanCache") { _ =>
+    val buildTarget = buildTargetUri("util", "Compile")
+    svr.sendJsonRpc(
+      s"""{ "jsonrpc": "2.0", "id": "32", "method": "buildTarget/cleanCache", "params": {
```
adpi2: Not a big deal but every request should have a unique id. Here `32` is already used by the `buildTarget/compile` request.

```suggestion
      s"""{ "jsonrpc": "2.0", "id": "44", "method": "buildTarget/cleanCache", "params": {
```

server-test/src/test/scala/testpkg/BuildServerTest.scala
```diff
@@ -108,6 +108,18 @@ object BuildServerTest extends AbstractServerTest {
...
+         |  "targets": [{ "uri": "$buildTarget" }]
+         |} }""".stripMargin
+    )
+    assert(processing("buildTarget/cleanCache"))
+    val res = svr.waitFor[CleanCacheResult](10.seconds)
+    assert(res.cleaned)
```
adpi2, line 137: Would it be possible to check that the class directory was not empty before the request? And that it is empty after the request?

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -186,6 +186,23 @@ object BuildServerProtocol {
     bspBuildTargetCompile / aggregate := false,
     bspBuildTargetTest := bspTestTask.evaluated,
     bspBuildTargetTest / aggregate := false,
+    bspBuildTargetCleanCache := Def.inputTaskDyn {
+      val s: State = state.value
+      val targets = spaceDelimited().parsed.map(uri => BuildTargetIdentifier(URI.create(uri)))
+      val workspace = bspFullWorkspace.value.filter(targets)
+      workspace.warnIfBuildsNonEmpty(Method.Compile, s.log)
```
adpi2:
```suggestion
      workspace.warnIfBuildsNonEmpty(Method.CleanCache, s.log)
```

adpi2 (changes requested):
Thanks @hmemcpy!

Most of the comments are suggestions. But I think it is important to check in the test that `buildTarget/cleanCache` empties the class directory.

hmemcpy:
Thanks for the comments! My copy-pasting was obvious here :)
I'll make the changes and will figure out how to check for empty directories!

server-test/src/test/scala/testpkg/BuildServerTest.scala
hmemcpy: @adpi2 Done, but I am hoping there's a less... hacky way of checking these files. I don't think I saw anything in the bsp protocol to let me list the target directories...

adpi2: I cannot think of anything better than it is already :)

adpi2 (dismissed):
LGTM

hmemcpy:
Oops, temporarily marking this as draft. It looks like the `results.size == targets.size` isn't a good check. I just noticed that out of the 3 targets, I get only 2 results back. The sent targets are:

```
  "targets": [
    {
      "uri": "file:/Users/hmemcpy/git/sbt-playground/#root-build"
    },
    {
      "uri": "file:/Users/hmemcpy/git/sbt-playground/#root/Compile"
    },
    {
      "uri": "file:/Users/hmemcpy/git/sbt-playground/#root/Test"
    }
```
 
And I get back just two. Unfortunately I don't know what they are, since it's a Unit result. Is, perhaps, `root-build` not a "real" target?

**Update**: I changed `Keys.clean` to `Keys.bspBuildTarget` and indeed, the two results I get back are `Compile` and `Test`, the `-build` is not present.

I understand this somewhat a synthetic target, perhaps I can filter it out when doing the comparison?

hmemcpy:
Alright... I managed to implement a solution for both "Rebuild Project" (which includes the root build), as well as a regular "Rebuild module" (which includes the exact targets to rebuild).

Would love any suggestions to improve this!

adpi2 (approved):
I don't see any obvious way to clean the build.sbt target. So this solution seems ok to me.

hmemcpy:
TYVM! :)

PR #6641, opened 2021-08-28
Fix for realpathish in top-level directories

darabos:
The `realpathish` function returns a slightly wrong path for top-level directories. `realpathish /root` gives `//root`. A copy of this issue in `sbt-native-packager` caused a real issue for us. (https://github.com/lynxkite/lynxkite/issues/200) That's fixed now (https://github.com/sbt/sbt-native-packager/pull/1451) but I thought maybe you would want the fix too.

Thanks for sbt! I've recently upgraded from 0.13.17 to 1.4.7 and it's awesome!

### DIFF
--- a/sbt
+++ b/sbt
@@ -46,11 +46,18 @@ realpathish () {
     COUNT=$(($COUNT + 1))
   done
 
+  TARGET_DIR="$(pwd -P)"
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #6642, opened 2021-08-28
BSP: implement progress report during compilation

### LINKED ISSUE #6570: BSP: support task progress
BSP supports [task progress notifications](https://build-server-protocol.github.io/docs/specification.html#task-progress). Bloop already supports these for target compile progress. Would be nice if sbt would do this as well where possible / for long running tasks.

hmemcpy:
Fixes #6570

This PR introduces `build/taskProgress` notifications to report the compilation progress (in 5% increments, borrowed from bloop)

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -12,7 +12,6 @@ import java.net.{ URI, URL }
 import java.nio.file.{ Paths, Path => NioPath }
 import java.util.Optional
 import java.util.concurrent.TimeUnit
-
 import lmcoursier.CoursierDependencyResolution
 import lmcoursier.definitions.{ Configuration => CConfiguration }
 import org.apache.ivy.core.module.descriptor.ModuleDescriptor
@@ -47,6 +46,7 @@ import sbt.internal.librarymanagement.mavenint.{
 import sbt.internal.librarymanagement.{ CustomHttp => _, _ }
 import sbt.internal.nio.{ CheckBuildSources, Globs }
 import sbt.internal.server.{
+  BspCompileProgress,
   BspCompileTask,
   BuildServerProtocol,
   BuildServerReporter,
@@ -2353,39 +2353,16 @@ object Defaults extends BuildCommon {
       )
     }
     def onProgress(s: Setup) = {
-      val p = s.progress.asScala
-      s.withProgress {
-        p.collect {
-          case c: CompileProgress =>
-            new CompileProgress {
-              override def startUnit(phase: String, unitPath: String): Unit =
-                c.startUnit(phase, unitPath)
-
-              override def afterEarlyOutput(success: Boolean): Unit =
-                c.afterEarlyOutput(success)
-
-              override def advance(
-                  current: Int,
-                  total: Int,
-                  prevPhase: String,
-                  nextPhase: String
-              ): Boolean = {
-                val percentage = current * 100 / total
-                // Report percentages every 5% increments
-                val shouldReportPercentage = percentage % 5 == 0
-                if (shouldReportPercentage) {
-                  task.notifyProgress(percentage, total)
-                }
-
-                c.advance(current, total, prevPhase, nextPhase)
-              }
-            }
-        }.asJava
-      }
+      val cp: BspCompileProgress = s.progress.asScala
+        .map(p => new BspCompileProgress(task, Some(p)))
+        .getOrElse(new BspCompileProgress(task, None))
+      s.withProgress(cp)
     }
     val compilers: Compilers = ci.compilers
     val setup: Setup = ci.setup
-    val i = ci.withCompilers(onArgs(compilers)).withSetup(onProgress(setup))
+    val i = ci
+      .withCompilers(onArgs(compilers))
+      .withSetup(onProgress(setup))
     try {
       val result = incCompiler.compile(i, s.log)
       reporter.sendSuccessReport(result.getAnalysis)
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2353,9 +2353,7 @@ object Defaults extends BuildCommon {
       )
     }
     def onProgress(s: Setup) = {
-      val cp: BspCompileProgress = s.progress.asScala
-        .map(p => new BspCompileProgress(task, Some(p)))
-        .getOrElse(new BspCompileProgress(task, None))
+      val cp = new BspCompileProgress(task, s.progress.asScala)
       s.withProgress(cp)
     }
     val compilers: Compilers = ci.compilers
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2305,14 +2305,14 @@ object Defaults extends BuildCommon {
     analysis
   }
   def compileIncrementalTask = Def.task {
+    val s = streams.value
+    val ci = (compile / compileInputs).value
+    val ping = earlyOutputPing.value
+    val reporter = (compile / bspReporter).value
     BspCompileTask.compute(bspTargetIdentifier.value, thisProjectRef.value, configuration.value) {
-      // TODO - Should readAnalysis + saveAnalysis be scoped by the compile task too?
-      compileIncrementalTaskImpl(
-        streams.value,
-        (compile / compileInputs).value,
-        earlyOutputPing.value,
-        (compile / bspReporter).value
-      )
+      task =>
+        // TODO - Should readAnalysis + saveAnalysis be scoped by the compile task too?
+        compileIncrementalTaskImpl(task, s, ci, ping, reporter)
     }
   }
   private val incCompiler = ZincUtil.defaultIncrementalCompiler
@@ -2337,6 +2337,7 @@ object Defaults extends BuildCommon {
     }
   }
   private[this] def compileIncrementalTaskImpl(
+      task: BspCompileTask,
       s: TaskStreams,
       ci: Inputs,
       promise: PromiseWrap[Boolean],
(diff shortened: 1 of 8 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2306,13 +2306,15 @@ object Defaults extends BuildCommon {
...
   def compileIncrementalTask = Def.task {
     BspCompileTask.compute(bspTargetIdentifier.value, thisProjectRef.value, configuration.value) {
-      // TODO - Should readAnalysis + saveAnalysis be scoped by the compile task too?
-      compileIncrementalTaskImpl(
-        streams.value,
-        (compile / compileInputs).value,
-        earlyOutputPing.value,
-        (compile / bspReporter).value
-      )
+      case task =>
```
hmemcpy: I made the `compile` parameter of `BspCompileTask.compute` into a `PartialFunction[BspCompileTask, CompileResult]`.
If I make it into a proper function, I will get a compilation error from the linter on all the values below that use the `/`):

```
The evaluation of `streams` inside an anonymous function is prohibited.
[error]
[error] problem: Task invocations inside anonymous functions are evaluated independently of whether the anonymous function is invoked or not.
[error] solution:
[error]   1. Make `streams` evaluation explicit outside of the function body if you don't care about its evaluation.
[error]   2. Use a dynamic task to evaluate `streams` and pass that value as a parameter to an anonymous function.
[error]   3. Annotate the `streams` evaluation with `@sbtUnchecked`, e.g. `(streams.value: @sbtUnchecked)`.
[error]   4. Add `import sbt.dsl.LinterLevel.Ignore` to your build file to disable all task linting.
[error]
[error]           streams.value,
[error]           ^
```

hmemcpy:
Very odd that this is failing... maybe a fluke? Could you give the CI a kick to try again?

eed3si9n:
Re-running now.

main/src/main/scala/sbt/Defaults.scala
adpi2: That means you need to resolve the values outside of the function body. Resolution must happen first then the lambda can be evaluated:
```scala
    val s = streams.value
    val ci = (compile / compileInputs).value
    val ping = earlyOutputPing.value
    val reporter = (compile / bspReporter).value
    BspCompileTask.compute(bspTargetIdentifier.value, thisProjectRef.value, configuration.value) {
      task =>
        // TODO - Should readAnalysis + saveAnalysis be scoped by the compile task too?
        compileIncrementalTaskImpl(task, s, ci, ping, reporter)
    }
```

adpi2: It would be better if the `compile` parameter is a `Function`

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2351,8 +2354,40 @@ object Defaults extends BuildCommon {
...
+                  total: Int,
+                  prevPhase: String,
+                  nextPhase: String
+              ): Boolean = {
+                val percentage = current * 100 / total
+                // Report percentages every 5% increments
+                val shouldReportPercentage = percentage % 5 == 0
+                if (shouldReportPercentage) {
+                  task.notifyProgress(percentage, total)
+                }
```
adpi2: Maybe you should extract this in a `BspCompileProgress.scala`:
```scala
private[sbt] class BspCompileProgress(task: BspCompileTask) extends CompileProgress {
  override def advance(
    current: Int,
    total: Int,
    prevPhase: String,
    nextPhase: String
  ): Boolean = {
    val percentage = current * 100 / total
    // Report percentages every 5% increments
    val shouldReportPercentage = percentage % 5 == 0
    if (shouldReportPercentage) {
      task.notifyProgress(percentage, total)
    }
  }
  true
}
```

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2351,8 +2354,40 @@ object Defaults extends BuildCommon {
         }
       )
     }
+    def onProgress(s: Setup) = {
+      val p = s.progress.asScala
+      s.withProgress {
+        p.collect {
```
adpi2: If for some reason `s.progress` is empty there would be no BSP progress report unfortunately.

So I suggest to add a fallback:

```scala
s.progress.map(p => BspCompileProgress(p, task))
  .getOrElse(BspCompileProgress(task))
```

adpi2 (changes requested):
Thanks for the contribution!

It would be great if you could also test it in the `BuildServerTest`. It could be in the existing `buildTarget/compile` test.

hmemcpy:
All done, plus added a test to verify progress reports (made it an additional test next to `buildTarget/compile`). Thanks again for the comments!

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2351,8 +2352,17 @@ object Defaults extends BuildCommon {
         }
       )
     }
+    def onProgress(s: Setup) = {
+      val cp: BspCompileProgress = s.progress.asScala
+        .map(p => new BspCompileProgress(task, Some(p)))
+        .getOrElse(new BspCompileProgress(task, None))
```
adpi2: Can be simplified to
```suggestion
      val cp = new BspCompileProgress(task, s.progress.asScala)
```

main/src/main/scala/sbt/internal/server/BspCompileProgress.scala
```diff
@@ -0,0 +1,31 @@
...
+      nextPhase: String
+  ): Boolean = {
+    val percentage = current * 100 / total
+    // Report percentages every 5% increments
+    val shouldReportPercentage = percentage % 5 == 0
+    if (shouldReportPercentage) {
+      task.notifyProgress(percentage, total)
+    }
+    underlying.foreach(_.advance(current, total, prevPhase, nextPhase))
+    true
```
adpi2: Now I realize my previous suggestion was a bit too simple. Because we should better return the underlying result if defined:

```suggestion
    underlying.map(_.advance(current, total, prevPhase, nextPhase))
      .getOrElse(true)
```

Also the other methods should redirect to the `underlying` progress. That was kind of what you did before. Sorry for the back and forth.
```scala
override def startUnit(phase: String, unitPath: String): Unit = {
  underlying.foreach(_.startUnit(phase, unitPath))
}

override def afterEarlyOutput(success: Boolean): Unit = {
  underlying.foreach(_.afterEarlyOutput(sucess))
}
```

server-test/src/test/scala/testpkg/BuildServerTest.scala
```diff
@@ -94,6 +94,39 @@ object BuildServerTest extends AbstractServerTest {
     assert(res.statusCode == StatusCode.Success)
   }
 
+  test("buildTarget/compile - reports compilation progress") { _ =>
```
adpi2, line 97: This test looks good!

adpi2 (changes requested):
Thanks for adding the test. The `BspCompileProgress` needs to redirect the call to the `underlying` progress when defined. Otherwise it looks good to me and I look forward to trying it in Metals.

main/src/main/scala/sbt/internal/server/BspCompileProgress.scala
hmemcpy: No worries at all! I ended up using `.fold(true)(_.advance(...))` because IntelliJ was trying to suggest I use `.forall` instead of map/getOrElse 😬

Thanks again for your comments!

adpi2 (approved):
That's amazing, thanks!

hmemcpy:
Awesome, thank you! I've been working on a detailed blog post of how to use bsp with IntelliJ, and those two issues were the last remaining bits of having the full experience compared to bloop. I'll publish it once 1.5.6 has shipped, so everyone can upgrade and enjoy it :)

server-test/src/test/scala/testpkg/BuildServerTest.scala
```diff
@@ -94,6 +94,35 @@ object BuildServerTest extends AbstractServerTest {
...
+
+    assert(svr.waitForString(10.seconds) { s =>
+      s.contains("build/taskProgress") &&
+      s.contains(""""message":"Compiling runAndTest (100%)"""")
+    })
+
+    assert(svr.waitForString(10.seconds) { s =>
+      s.contains("build/taskFinish") &&
+      s.contains(""""message":"Compiled runAndTest"""")
+    })
```
eed3si9n, line 123: I'm not sure which one of the `waitForString`, but these seem to be flaky at least directly after when Scala version was bumped to Scala 2.12.15. Like, sometimes the test would pass and other times it doesn't.

- https://github.com/eed3si9n/sbt/runs/3642158965
- https://github.com/sbt/sbt/runs/3634885441

I'm going to try to bump the timeout to 60s and see if that would help.

adpi2: Does it work better after increasing the timeout?

If not then it probably means that test progress is not deterministic, depends on the timing.

main/src/main/scala/sbt/internal/server/BspCompileTask.scala
```diff
@@ -52,66 +51,85 @@ object BspCompileTask {
...
+  private[sbt] def notifyProgress(percentage: Int, total: Int): Unit = {
+    val data = Converter.toJsonUnsafe(CompileTask(targetId))
+    val message = s"Compiling $targetName ($percentage%)"
+    val currentMillis = System.currentTimeMillis()
+    val params = TaskProgressParams(
+      id,
+      Some(currentMillis),
+      Some(message),
+      Some(total.toLong),
+      Some(percentage.toLong),
```
bishabosha, line 112: unfortunately this should have been the "current" value, and not the percentage, i.e. the current/total, otherwise metals will try to render e.g. 95%/50 for project foo and report `foo - 30s (190%)`, or pretty much be stuck at zero for large modules

I propose two fixes:
- we can report the `"current"` correctly, but I think the kind should be updated to `"compile-progress2"` (or `"sbt-progress"`) to preserve compatibility
- metals client/whatever when they receive a `build/taskProgress` of kind `"compile-progress"` they must ignore the `"total"` field from the notification, and fix it to always be `100`.

adpi2: How does IntelliJ handle it? How does it handle `bloop-progress`? Can we do the same in Metals?

bishabosha: It looks like IntelliJ ignores the kind field:
https://github.com/JetBrains/intellij-scala/blob/5100ec6250070db147f409af755e08a75873e1a3/bsp/src/org/jetbrains/bsp/project/BspTask.scala#L298

I can only see that it prints the message field in the build window

adpi2: Ok then I think we should fix this in sbt so that it is consistent with Bloop.

bishabosha: ok but I still propose that we change the `kind` of the event, this way metals will always be correct, as it can handle the old kind the old way

PR #6649, opened 2021-09-05
Fix ClassCastException

eed3si9n:
Ref #6592
When there's an issue like timeout, currently it throws a
ClassCastException because I made some assumption about the underlying
error. This removes the unnecessary casting.


TristanCP:
Hi, I'm interested in this fix, how can I find the corresponding fix version ?

mzuehlke:
This fix is included in https://github.com/sbt/sbt/releases/tag/v1.6.0-M1 and all following version of the `1.6.*` series.

PR #6657, opened 2021-09-12
Implement a new `copy-file-with-options` scripted test command

guizmaii:
I don't find where I should write the tests, anyone knows where? 🤔

I found this PR where some tests were implemented but the test file doesn't exist anymore apparently: https://github.com/sbt/sbt/pull/1419/files


eed3si9n:
We have scripted tests of sbt itself under https://github.com/sbt/sbt/tree/develop/sbt-app/src/sbt-test, so maybe that could be used?

awoifabilaye (dismissed):

PR #6665, opened 2021-09-19
Drop TrapExit

### LINKED ISSUE #6558: sbt uses setSecurityManager which is now deprecated on JDK 17
```
% git grep setSecurity
run/src/main/scala/sbt/TrapExit.scala:      case m           => System.setSecurityManager(new TrapExit(m)); m
run/src/main/scala/sbt/TrapExit.scala:    System.setSecurityManager(previous)
```

I noticed this in the context of https://github.com/scala/community-build/issues/1432

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/6558

Problem
-------
sbt uses SecurityManager feature of JDK to trap `sys.exit` call during
`run`-like tasks, since we emulate `run` and `console` as function calls.
JDK 17 deprecated SecurityManager and it's printing warnings.

Solution
--------
About 10 years go, `exit` was a convenient way of quitting both Scala
REPL and sbt shell. Scala 2.11 broke this by removing the `Predef.exit`.
We still need to worry about `run` potentially calling `sys.exit`
but that can be handled using fork feature.
In the long-run, it probably is better to be JDK 17 compatible.


xuwei-k:
https://github.com/sbt/sbt/blob/b2f0b7dadad0f66f2c2511b6541b1bc158fc3a57/run/src/main/scala/sbt/TrapExitSecurityException.scala#L15

Should we deprecate or remove `TrapExitSecurityException` as well? 👀

eed3si9n:
> Should we deprecate or remove `TrapExitSecurityException` as well?

I think that makes sense.

PR #6673, opened 2021-09-27
Make javaHome that forks scripted tests configurable

kxbmap:
Normally scripted tests are forked using the JVM that is running sbt.
If set `scripted / javaHome`, forked using it.

```scala
scripted / javaHome := Some(file("/path/to/jdk-x.y.z"))
```

Or use `java++` command before scripted.

```
sbt> java++ 11!
sbt> scripted
```


eed3si9n:
@kxbmap Thanks for the contribution. Could you describe the motivation behind this change? Are you looking to build your plugin using JDK 8, but test on JDK 11 or 17?

main/src/main/scala/sbt/ScriptedRun.scala
```diff
@@ -0,0 +1,179 @@
...
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt
+
+import java.io.File
+import java.lang.reflect.Method
+import scala.annotation.unused
+
+sealed trait ScriptedRun {
```
eed3si9n, line 14: Maybe it would be good to hide this under `package[sbt]` and `package sbt.internal`?

kxbmap:
@eed3si9n
The plugin I am developing needs to behave differently depending on whether the JDK the user is using implements Jigsaw or not (Java 9 later or not), or whether it removes JavaEE related modules or not (Java 11 later or not).
Therefore, I want to test on JDK 8, 9, and 11 (also 17).
Thank you.

main/src/main/scala/sbt/ScriptedRun.scala
kxbmap: This class is public via `val scriptedRun: TaskKey[ScriptedRun]` is public.
If change it's visibility, maybe also needs to change  visibility of `scriptedRun`.

eed3si9n:
When JDK 11 came out, we added cross JDK forking, which sounds related to this. I'm not sure if that works with scripted yet, but generally the workflow is:

```
sbt:helloworld> run
[info] Running (fork) Hello
[info] 1.8.0_171
sbt:helloworld> java++ 10!
[info] Reapplying settings...
sbt:helloworld> run
[info] Running (fork) Hello
[info] 10.0.1
```

https://www.scala-sbt.org/1.x/docs/sbt-1.2-Release-Notes.html#Cross+JDK+forking

kxbmap:
Updated PR with using `java++` command in scripted.

eed3si9n (dismissed):
Overall LGTM

Could you update the description of this PR and add documentation on how to use this feature please?

kxbmap:
Updated PR description and added note.
If this is OK, I will rebase and squash commits.

eed3si9n:
> If this is OK, I will rebase and squash commits.

Sounds good.

eed3si9n (approved):

PR #6679, opened 2021-09-29
Rename internal/util-logging/src/main/scala/sbt/internal/util/MainLogging.scala to MainAppender.scala to match contents

er1c:
Was looking for this file and couldn't find it, so just updating the name.


er1c:
Included in https://github.com/sbt/sbt/pull/6680

PR #6691, opened 2021-10-16
Switch to using actions/setup-java@v2

eed3si9n:


SethTisue:
I believe you're supposed to specify `temurin` rather than `adopt` even for JDK 8, in order to get the latest updates. (Assuming you consider it desirable to build/test on the absolute latest update version.)

eed3si9n:
Yea. I got that right for the docs but my understanding of the situation 5h ago wasn't as good.

Issue #5962, opened 2020-10-09
dependencyTree: Ignore terminal width when running without terminal


melezov:
There are a few issues with terminal width handling that have been raised on sbt-dependency-graph, a'la:
https://github.com/sbt/sbt-dependency-graph/issues/182

The solutions are pretty simple:
https://github.com/sbt/sbt-dependency-graph/pull/181/files

The reason why I'm raising this on SBT instead of sbt-dependency-graph is because I've previously resolved the issue by publishing a modified version of the sbt-dependency-graph plugin, and used that for my nefarious purposes.

I am no longer able to do that, since now adding the modified plugin produces no effect on the dependencyTree::toFile commands.

Please adopt the proposed enhancements (in combination with new JLine / Terminal) or suggest how to "override" the in-sourced dependency graph plugin.

eed3si9n:
We'd probably use `asciiGraphWidth` setting as the fallback https://github.com/sbt/sbt/pull/2941.

Probably the relevant code is https://github.com/sbt/sbt/blob/bfb7d4019bd3b3849c0f58bbdbe310f35ecc37cf/main/src/main/scala/sbt/internal/SettingGraph.scala.

PR #6693, opened 2021-10-18
Enable the asciiGraphWidth setting to the dependencyTree tasks. #5962

kijuky:
Apply asciiGraphWidth setting to dependencyTree tasks for fixing #5962. 

#### Implements:
- Ignore terminal width. ( This change may not be acceptable. )
- render for renderingTaskSettings set null. ( Then set the correct value. )


main/src/main/scala/sbt/plugins/DependencyTreeSettings.scala
```diff
@@ -80,7 +80,13 @@ object DependencyTreeSettings {
...
       },
       dependencyTreeModuleGraphStore := (dependencyTreeModuleGraph0 storeAs dependencyTreeModuleGraphStore triggeredBy dependencyTreeModuleGraph0).value,
-    ) ++ renderingTaskSettings(dependencyTree, rendering.AsciiTree.asciiTree _)
+    ) ++ {
+      renderingTaskSettings(dependencyTree, null) :+ {
+        dependencyTree / asString := {
+          rendering.AsciiTree.asciiTree(dependencyTreeModuleGraph0.value, asciiGraphWidth.value)
+        }
+      }
+    }
```
eed3si9n: I think this is a bit confusing. It might be better to make an overload of `renderingTaskSettings` that only takes `key: TaskKey[Unit]` than passing `null` here, and implement with the correct `dependencyTree / asString` in there. Is that possible?

eed3si9n (commented):
Thanks for the contribution! Overall I think the fix looks good, but have a minor suggestion for improvement.

main/src/main/scala/sbt/plugins/DependencyTreeSettings.scala
kijuky: I didn't want to pass null either. 

Overload! This seems to be the solution!

eed3si9n (approved):
Thanks!

PR #6697, opened 2021-10-21
[#6696] Add shutdownall to sbt runner to shutdown all sbt-launch processes

### LINKED ISSUE #6696: Add a `shutdownall sbt`
With sbt server, we can have many idle processes running in the background. For one reason or another (low cpu, low ram, unresponding server) the user may want to kill all these sbt processes.

One of the possible solution is to run this bash command manually: `jps -v | grep sbt-launch | cut -f1 -d ' ' | xargs kill -9`.

For convenience, sbt could provide a similar command. It could be part of the existing script (`sbt -shutdownall`) or a standalone script `shutdownall-sbt`.

er1c:
Fixes #6696

#### OSX

```
$ ./sbt shutdownall
shutdown 1 sbt processes
$ ./sbt shutdownall
shutdown 0 sbt processes
```

#### Windows

```
PS C:\Users\Eric Peters> sbt shutdownall
SUCCESS: The process with PID 9948 has been terminated.
shutdown 1 sbt processes
PS C:\Users\Eric Peters> sbt shutdownall
shutdown 0 sbt processes
PS C:\Users\Eric Peters> sbt shutdownall
shutdown 0 sbt processes
PS C:\Users\Eric Peters> sbt shutdownall
SUCCESS: The process with PID 16512 has been terminated.
shutdown 1 sbt processes
PS C:\Users\Eric Peters>
```

### DIFF
--- a/launcher-package/src/universal/bin/sbt.bat
+++ b/launcher-package/src/universal/bin/sbt.bat
@@ -25,6 +25,7 @@ set default_java_opts=-Dfile.encoding=UTF-8
 set sbt_jar=
 set build_props_sbt_version=
 set run_native_client=
+set shutdownall=
 
 set sbt_args_print_version=
 set sbt_args_print_sbt_version=
@@ -399,6 +400,11 @@ if defined _timings_arg (
   goto args_loop
 )
 
+if "%~0" == "shutdownall" (
+  set shutdownall=1
+  goto args_loop
+)
+
 if "%~0" == "--script-version" (
   set sbt_args_print_sbt_script_version=1
   goto args_loop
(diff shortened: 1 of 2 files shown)


eed3si9n:
May I suggest a slightly gentler name like `shutdownall`?

er1c:
> shutdownall

@eed3si9n done!

eed3si9n:
I understand this is not a real sbt command like `shutdown`, but it might be confusing that to shutdown one you don't need `--` but to shutdown all, you need `--`. My suggestion would be to pretend this is a command and call it:

```bash
$ sbt shutdownall
```

WDYT?

Also, could there be a Windows parity for this feature?

er1c:
> I understand this is not a real sbt command like `shutdown`, but it might be confusing that to shutdown one you don't need `--` but to shutdown all, you need `--`. My suggestion would be to pretend this is a command and call it:
> 
> ```shell
> $ sbt shutdownall
> ```
> 
> WDYT?
> 
> Also, could there be a Windows parity for this feature?

I like it, thanks for thinking outside the box and consistency in usage.  I'll also see about the windows parity.

er1c:
@eed3si9n updated, removed the `--` and added `.bat` support

eed3si9n (approved):
Thanks!

charego:
Note with environment variable `SBT_NATIVE_CLIENT=true`, you will also need to supply `--server` for this to work.

```
$ sbt --server shutdownall
shutdown 1 sbt processes
```

PR #6703, opened 2021-11-03
Fix #6698: Synchronize all RPC messages with systemOut notifications

### LINKED ISSUE #6698: sbt client (--client) does not print error output on repeated commands, just quits with "[error]..."
## steps

- Setup minimal project with sbt `1.5.5`
- with some main method that fails, lets say at runtime (but also happens sometimes with compiletime error output)
- have the sbt server powered down
- run `sbt --client runMain <your Main class>` and observe error output
- run the same command again -- **and just get the message that there was an error, but no details**
- run `sbt --client shutdown`, and repeat: the first time -- when the server boots up -- the error message is there, the second time run, it is not.

## problem

Reproducible on my machine only on recurring runs of `sbt --client` as described above, the thin client **sometimes doesn't produce any error output on runtime / compiletime errors**. Sometimes it works (on a running sbt instance) and then stops working later.

Example output: -- first time, error gets printed, second time, it doesn't get printed.

```
should print
[error] (run-main-0) java.lang.AssertionError: assertion failed
[error]
[error] obj1 == obj2
[error] |    |  |
[error] |    |  IS(4,blub)
[error] |    false
[error] IS(3,hellooooo)
[error] java.lang.AssertionError: assertion failed
[error]
[error] obj1 == obj2
[error] |    |  |
[error] |    |  IS(4,blub)
[error] |    false
[error] IS(3,hellooooo)
[error]
[error]         at com.eed3si9n.expecty.ExpectyBase$ExpectyListener.expressionRecorded(Expecty.scala:41)
[error]         at com.eed3si9n.expecty.RecorderRuntime.recordExpression(RecorderRuntime.scala:40)
[error]         at bci.logic.Stuff$.StuffMain(Stuff.scala:21)
[error]         at bci.logic.StuffMain.main(Stuff.scala:7)
[error] Total time: 1 s, completed Oct 21, 2021 6:18:05 PM
snuc@snuc-NUC8i7BEH:~/sandbox/bct/bci-logic$ sbt --client runMain bci.logic.StuffMain
> runMain bci.logic.StuffMain
[error] Total time: 0 s, completed Oct 21, 2021 6:18:07 PM
snuc@snuc-NUC8i7BEH:~/sandbox/bct/bci-logic$
```

## expectation

it should just print errors every time.

## notes

Maybe the client pipe gets closed too early or not flushed?

adpi2:
Fixes #6698

`systemOut` notifications are buffered so that they are sent at most
once every 20 millisecond. Other RPC messages are not buffered.
As a consequence, some RPC messages can pass in front of some
`systemOut` notifications.
That explains why `sbt --client run` can exit before it receives all the logs.

In general I think it is safer to maintain the order of all messages.
To do so we can force the flush of systemOut before each RPC message.

With this main class:
```scala
package example

object Main {
    println("Hello, World!")
    throw new Exception("some failure")
  }
}
```

Before the fix:
image

After the fix:
image

### DIFF
--- a/main/src/main/scala/sbt/internal/server/NetworkChannel.scala
+++ b/main/src/main/scala/sbt/internal/server/NetworkChannel.scala
@@ -602,6 +602,7 @@ final class NetworkChannel(
       execId: String,
       err: JsonRpcResponseError
   ): Unit = {
+    forceFlush()
     val m = JsonRpcResponseMessage("2.0", execId, None, Option(err))
     val bytes = Serialization.serializeResponseMessage(m)
     publishBytes(bytes)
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #6707, opened 2021-11-11
Fix infinte loop when server fails to load with batch client

### LINKED ISSUE #6547: sbt thin client with batch mode gets stuck into an infinite loop when build can't be loaded
## steps

With sbt version 1.5.4:
- Add `-batch` to `.sbtopts` or `/etc/sbt/sbtopts`
- Add a key that doesn't exist in a build.sbt, or a plugin that has a broken version.
- Run:
> sbt --client compile

or
> SBT_NATIVE_CLIENT=true sbt compile

## problem

The thin client starts, then fails to load the project and gets stuck into an infinite loop of:
```
[warn] Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)
Invalid response: 'ÿ'
```

## expectation

Shutdown the sbt client without expecting an user input.

## notes

We encountered this issue when a bot that automates our dependency updates bumped a version of a plugin with an invalid POM. The plugin couldn't be resolved, which is expected, but our CI job got stuck until timeout.
Here is a small project to easily reproduce the issue if needed https://github.com/max-peroch/sbtn-batch-bug

adpi2:
Fix #6547 (@max-peroch)

Shut down server when it is waiting for user input but stdin has ended.

Tested locally:
```
> sbt -batch --java-client
[info] welcome to sbt 1.6.0-SNAPSHOT (AdoptOpenJDK Java 1.8.0_275)
[info] loading global plugins from /home/piquerez/.sbt/1.0/plugins
[info] loading project definition from /home/piquerez/sbt/sbt-6547/project
/home/piquerez/sbt/sbt-6547/build.sbt:5: error: not found: value foo
foo
^
[warn] Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)
```

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -385,8 +385,8 @@ class NetworkClient(
         if (socket.isEmpty) Thread.sleep(20)
       }
     }
-    val hook = new Thread(() => Option(sbtProcess.get).foreach(_.destroyForcibly()))
-    Runtime.getRuntime.addShutdownHook(hook)
+    val shutdown = new Thread(() => Option(sbtProcess.get).foreach(_.destroyForcibly()))
+    Runtime.getRuntime.addShutdownHook(shutdown)
     var gotInputBack = false
     val readThreadAlive = new AtomicBoolean(true)
     /*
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #6714, opened 2021-11-16
SourcePositionMacro for Scala 3. enable Scala 3 build for some `util` modules

xuwei-k:


.github/workflows/ci.yml
```diff
@@ -134,7 +135,7 @@ jobs:
       if: ${{ matrix.jobtype == 5 }}
       shell: bash
       run: |
-        ./sbt -v "++$SCALA_213!; test;"
+        ./sbt -v "++$SCALA_213!; test; ++$SCALA_3!; all utilControl/test utilRelation/test utilPosition/test"
```
xuwei-k, line 138: Should we create another new job type `10`? 🤔

.scalafmt.conf
```diff
@@ -21,3 +21,8 @@ align.openParenDefnSite = false
 danglingParentheses = true
 
 trailingCommas = preserve
+
+# TODO update scalafmt and enable Scala 3
```
xuwei-k, line 25: scalafmt 2.x does not support Scala 3.
https://github.com/sbt/sbt/blob/5c90a84c72ca6aae1f08dae1cd1467da6a3d29fd/.scalafmt.conf#L1-L2

.github/workflows/ci.yml
eed3si9n: job type 5 took 8 min (the longest one takes 23 min) so I think this is fine.

eed3si9n (approved):

PR #6716, opened 2021-11-16
Fixes tab completion of global keys

### LINKED ISSUE #1373: Global setting keys don't show up in autocomplete (sbtVersion)
(using 0.13.5-RC5)

```
akka > sbt
sbtBinaryVersion    sbtClearOnFailure   sbtDependency       sbtPlugin           sbtPopOnFailure     sbtResolver         sbtStashOnFailure   
akka > sbtV
{invalid input}   
akka >
```

An interesting factoid: 78 upvotes for [how to check sbt version](http://stackoverflow.com/questions/8461013/how-to-check-sbt-version).

### LINKED ISSUE #6715: setting keys do not always appear as completion suggestions
## steps

- Clone the minimal repository [here](https://github.com/tpolecat/sbt-repro) and run `sbt`.
- Observe that `mimaVersionCheckExtraVersions` (provided by the [mima-version-check plugin](https://github.com/ChristopherDavenport/sbt-mima-version-check)) is a valid key.
- Observe that `mima<tab>` shows completions for MiMa keys, but not for keys provided by mima-version-check.

image

## problem

Tab completion does not suggest keys provided by mima-version-check, but _does_ suggest keys provided by MiMa (which is pulled in as a transitive dependency).

## expectation

All available keys should be visible through tab completion.

## notes

It's possible that this is a duplicate. I didn't see anything that seemed to be exactly the same but I don't understand sbt very well.

eed3si9n:
Fixes #1373
Fixes #6715

Following my own PR #2855 around `ThisBuild`, this attempts to fix the
tab completion of globally scoped keys from the shell.

mima

### DIFF
--- a/main/src/main/scala/sbt/internal/Act.scala
+++ b/main/src/main/scala/sbt/internal/Act.scala
@@ -304,7 +304,12 @@ object Act {
         case Some(ProjectRef(uri, _)) => index.keys(Some(BuildRef(uri)), conf, task)
         case _                        => Set()
       }
-    val keys: Set[String] = index.keys(proj, conf, task) ++ buildKeys
+    val globalKeys: Set[String] =
+      proj match {
+        case Some(_) => index.keys(None, conf, task)
+        case _       => Set()
+      }
+    val keys: Set[String] = index.keys(proj, conf, task) ++ buildKeys ++ globalKeys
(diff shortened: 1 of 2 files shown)


tpolecat:
Amazing, thank you!

Issue #6718, opened 2021-11-18
when ~/.sbt/1.0/plugins exists, sbt fails to boot if configured to use private repository


daddykotex:
## steps

1. configure sbt to use a private repository (credentials, repositories file, SBT_OPTS=`-Dsbt.override.build.repos=true -Dsbt.boot.credentials=/root/.sbt/.credentials`
2. launch sbt when `~/.sbt/1.0/plugins` does not exists (success)
3. launch sbt when `~/.sbt/1.0/plugins` does exists (**fail**)

## problem

it seems like credentials are discarded for some reason and thus we get unauthorized http response from the repository

## expectation

the sbt / boot process should use the same credentials to fetch sbt and any other jars it needs.

## notes

I've documented a way to reproduce here: https://github.com/daddykotex/sbt-issue-creds without messing up your own system.

I'd be happy to fix it. I've already checked out the source of both `sbt/sbt` and `sbt/librarymanagement` but so far, I can't crack it.

eed3si9n:
Thanks for the report. Please note that `sbt.boot.credentials` only affects the sbt launcher (that's whoever added it meant by "boot" here) to resolve the core artifacts of the sbt itself.

The metabuild (plugin world) normally depends on the sbt artifacts as well, so my guess is that resolver is failing because you'd need to configure the credential for Coursier used to resolve the metabuild.

eed3si9n:
Also here from the documentation - https://www.scala-sbt.org/1.x/docs/Proxy-Repositories.html#Using+credentials+for+the+proxy+repository

daddykotex:
I'm 95% sure there is a bug.

At this point, I've hit all the corner cases of credentials handling in this setup:

- ivy properties files along with the sbt setting file
- coursier properties file along with the env var
- repositories file along with the sbt opt

The key takeaway of this reproduction is that `sbt` will fail to start when there is **only** `1.0/plugins`. No plugin file (.scala) to compile, no `addSbtPlugin` (.sbt) either.

In the repro, when there is no `1.0/plugins` folder, everything works:

- the launcher fetches sbt 1.5.5 and scala
- sbt then fetches all it needs to run the project, from the private repository (including dependencies (like typesafe hocon config as defined in the dummy build))

When you add the folder, it breaks:

- the launcher fetches sbt 1.5.5 and scala
- sbt fails when resolving plugin `[info] loading global plugins from /root/.sbt/1.0/plugins` (even though no plugins are defined)

[...]

PR #6724, opened 2021-11-25
Load credentials from SBT_CREDENTIALS

daddykotex:
This addresses #6718. I'm not sure this is the right place to load that. Let me know what you think. If you say that's okay, I'll add in the tests.

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
@@ -308,7 +308,7 @@ object Terminal {
   }
   private[sbt] lazy val isAnsiSupported: Boolean = logFormatEnabled.getOrElse(useColorDefault)
 
-  private[this] val isDumb = "dumb" == System.getenv("TERM")
+  private[this] val isDumb = "dumb" == sys.env("TERM")
   private[this] def isDumbTerminal = isDumb || System.getProperty("jline.terminal", "") == "none"
   private[this] val hasConsole = Option(java.lang.System.console).isDefined
   private[this] def useColorDefault: Boolean = {
@@ -736,22 +736,20 @@ object Terminal {
       val supershell: Boolean
   )
   private[sbt] val TERMINAL_PROPS = "SBT_TERMINAL_PROPS"
-  private val props = System.getenv(TERMINAL_PROPS) match {
-    case null => None
-    case p =>
-      p.split(",") match {
-        case Array(width, height, ansi, color, supershell) =>
-          Try(
-            new Props(
-              width.toInt,
-              height.toInt,
-              ansi.toBoolean,
-              color.toBoolean,
-              supershell.toBoolean
-            )
-          ).toOption
-        case _ => None
-      }
+  private val props = sys.env.get(TERMINAL_PROPS) flatMap { p =>
+    p.split(",") match {
+      case Array(width, height, ansi, color, supershell) =>
+        Try(
+          new Props(
+            width.toInt,
+            height.toInt,
+            ansi.toBoolean,
+            color.toBoolean,
+            supershell.toBoolean
+          )
+        ).toOption
+      case _ => None
+    }
   }
   private[sbt] def startedByRemoteClient = props.isDefined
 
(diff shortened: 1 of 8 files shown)


eed3si9n:
@daddykotex I think overall this is good.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -148,6 +148,9 @@ object Defaults extends BuildCommon {
     val m = (for (a <- cp; an <- a.metadata get Keys.analysis) yield (a.data, an)).toMap
     m.get _
   }
+  private[sbt] def sbtCredentialsEnv(): Option[Credentials] =
+    Option(System.getenv("SBT_CREDENTIALS")).map(raw => new FileCredentials(new File(raw)))
```
eed3si9n: minor nitpick: I'd prefer `sys.env.get("SBT_CREDENTIALS")` and also move it under https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/internal/SysProp.scala maybe.

daddykotex: sure thing, we'll do later today. I used `System.getenv` because it's used into a other places and I figured that was a reason for its usage.

Do you want me to add a a commit that uses `sys.env` across the board (or in places where possible?)

daddykotex: I've updated the as suggested: moved to Sysprop and used sys.env instead. I've also added a commit to use sys.env in the other places, if it's too risky, ill remove it.

eed3si9n (approved):
Thanks!

internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
```diff
@@ -308,7 +308,7 @@ object Terminal {
   }
   private[sbt] lazy val isAnsiSupported: Boolean = logFormatEnabled.getOrElse(useColorDefault)
 
-  private[this] val isDumb = "dumb" == System.getenv("TERM")
+  private[this] val isDumb = "dumb" == sys.env("TERM")
```
eed3si9n, line 311:
```suggestion
  private[this] val isDumb = Some("dumb") == sys.env.get("TERM")
```

Looks like this behavior changed since  `System.getenv` returns `null` here:
> the string value of the variable, or null if the variable is not defined in the system environment

```
[info]   Cause: java.util.NoSuchElementException: key not found: TERM
[info]   at scala.collection.MapLike.default(MapLike.scala:236)
[info]   at scala.collection.MapLike.default$(MapLike.scala:235)
[info]   at scala.collection.AbstractMap.default(Map.scala:65)
[info]   at scala.collection.MapLike.apply(MapLike.scala:144)
[info]   at scala.collection.MapLike.apply$(MapLike.scala:143)
[info]   at scala.collection.AbstractMap.apply(Map.scala:65)
[info]   at sbt.internal.util.Terminal$.<init>(Terminal.scala:311)
[info]   at sbt.internal.util.Terminal$.<clinit>(Terminal.scala)
[info]   at sbt.internal.util.ConsoleAppender$.apply(ConsoleAppender.scala:208)
[info]   at sbt.internal.util.ConsoleAppender$.apply(ConsoleAppender.scala:199)
```

eed3si9n:
```suggestion
  private[this] val isDumb = Some("dumb") == sys.env.get("TERM")
```

Looks like this behavior changed since  `System.getenv` returns `null` here:
> the string value of the variable, or null if the variable is not defined in the system environment

```
[info]   Cause: java.util.NoSuchElementException: key not found: TERM
[info]   at scala.collection.MapLike.default(MapLike.scala:236)
[info]   at scala.collection.MapLike.default$(MapLike.scala:235)
[info]   at scala.collection.AbstractMap.default(Map.scala:65)
[info]   at scala.collection.MapLike.apply(MapLike.scala:144)
[info]   at scala.collection.MapLike.apply$(MapLike.scala:143)
[info]   at scala.collection.AbstractMap.apply(Map.scala:65)
[info]   at sbt.internal.util.Terminal$.<init>(Terminal.scala:311)
[info]   at sbt.internal.util.Terminal$.<clinit>(Terminal.scala)
[info]   at sbt.internal.util.ConsoleAppender$.apply(ConsoleAppender.scala:208)
[info]   at sbt.internal.util.ConsoleAppender$.apply(ConsoleAppender.scala:199)
```

PR #6735, opened 2021-12-01
Add `public` modifier to the SubclassFingerscan and AnnotatedFingerscan.

kpodsiad:
Currently, both `SubclassFingerscan` and `AnnotatedFingerscan` are package-private within the `sbt` package. This forces for example [bloop](https://github.com/scalacenter/bloop/blob/master/frontend/src/main/scala/sbt/SerializableFingerprints.scala) and [scala-debug-adapter](https://github.com/scalacenter/scala-debug-adapter/blob/989106082776015999c2c6f631d415a3544ce108/sbt/test-agent/src/main/java/ch/epfl/scala/debugadapter/sbtplugin/internal/SubclassFingerscan.java) to have some workaround about this.

Adding `public` modifier will solve mentioned problem.

### DIFF
--- a/testing/agent/src/main/java/sbt/ForkMain.java
+++ b/testing/agent/src/main/java/sbt/ForkMain.java
@@ -27,12 +27,12 @@ public final class ForkMain {
   // serializables
   // -----------------------------------------------------------------------------
 
-  static final class SubclassFingerscan implements SubclassFingerprint, Serializable {
+  public static final class SubclassFingerscan implements SubclassFingerprint, Serializable {
     private final boolean isModule;
     private final String superclassName;
     private final boolean requireNoArgConstructor;
 
-    SubclassFingerscan(final SubclassFingerprint print) {
+    public SubclassFingerscan(final SubclassFingerprint print) {
       isModule = print.isModule();
       superclassName = print.superclassName();
       requireNoArgConstructor = print.requireNoArgConstructor();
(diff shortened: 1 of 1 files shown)


testing/agent/src/main/java/sbt/ForkMain.java
```diff
@@ -27,12 +27,12 @@
   // serializables
   // -----------------------------------------------------------------------------
 
-  static final class SubclassFingerscan implements SubclassFingerprint, Serializable {
+  public static final class SubclassFingerscan implements SubclassFingerprint, Serializable {
```
eed3si9n, line 30: I am ok with the change itself, but isn't ForkMain an implementation detail of sbt?
It's unlikely that we'd change this too much, but I don't think there's guarantee that it would remain the same either.

eed3si9n (approved):

testing/agent/src/main/java/sbt/ForkMain.java
kpodsiad: Hmm, it's located in `test-agent` module and it's related to executing tests. I think it's okay that other tools are using it for running tests, but they have to take into account that it could change. Seems fair to me, wdyt?

eed3si9n: I guess so.

Issue #6785, opened 2022-01-10
sbt.sh -S-X option does nothing


bloznelis:
## steps
```
$ sbt --version
sbt version in this project: 1.5.6
sbt script version: 1.5.6
```

```
$ cat build.sbt
scalacOptions ++= Seq("-Ywarn-dead-code")
```
```
$ sbt 'show scalacOptions'
[info] welcome to sbt 1.5.6 (Oracle Corporation Java 11.0.13)
[info] loading global plugins from /home/lukas/.sbt/1.0/plugins
[info] loading project definition from /home/lukas/sbt-reproduce/project
[info] loading settings for project sbt-reproduce from build.sbt ...
[info] set current project to sbt-reproduce (in build file:/home/lukas/sbt-reproduce/)
[info] * -Ywarn-dead-code
[success] Total time: 0 s, completed Jan 10, 2022, 7:55:38 PM
```

```
$ sbt -S-Ywarn-value-discard 'show scalacOptions'
[info] welcome to sbt 1.5.6 (Oracle Corporation Java 11.0.13)
[info] loading global plugins from /home/lukas/.sbt/1.0/plugins
[info] loading project definition from /home/lukas/sbt-reproduce/project
[info] loading settings for project sbt-reproduce from build.sbt ...
[info] set current project to sbt-reproduce (in build file:/home/lukas/sbt-reproduce/)
[error] Expected symbol
[error] Not a valid command: -
[error] Expected end of input.
[error] Expected '--'
[error] Expected 'debug'
[error] Expected 'info'
[error] Expected 'warn'
[error] Expected 'error'
[error] Expected 'addPluginSbtFile'
[error] -S-Ywarn-value-discard
```

## problem
CLI option ` -S-X                add -X to sbt's scalacOptions (-S is stripped)` defined here https://github.com/sbt/sbt/blob/develop/sbt#L583 does not work as expected.
## expectation

Expected behavior is appended scalac option provided via CLI option:

```
$ sbt 'show scalacOptions'
[info] welcome to sbt 1.5.6 (Oracle Corporation Java 11.0.13)
[info] loading global plugins from /home/lukas/.sbt/1.0/plugins
[info] loading project definition from /home/lukas/sbt-reproduce/project
[info] loading settings for project sbt-reproduce from build.sbt ...
[info] set current project to sbt-reproduce (in build file:/home/lukas/sbt-reproduce/)
[info] * -Ywarn-dead-code
[info] * -Ywarn-value-discard
```

[...]

PR #6788, opened 2022-01-14
Remove scalacoption -S-X which does nothing

Nirvikalpa108:
This compiler option does not work as explained here #6785, so removing this.


Nirvikalpa108:
Closed in favour of https://github.com/sbt/sbt/pull/6799, because I made a git mistake.

PR #6814, opened 2022-02-04
Add support for Scala 3 -scala-output-version flag

prolativ:
When `-Yscala-release` flag is present in `scalacOptions` properly set the version of Scala 3 standard library which is seen as a dependency of this project (in Ivy and POM files).
This is needed so that using this project as a dependency in other projects doesn't cause adding transitively a newer standard library to the classpath as it might not be consumable for older compilers (because of using newer TASTy format).

As this flag specifies only a minor Scala version (because after the TASTy format gets fixed in `3.x.0` stable it cannot be changed in this `3.x` release series, except for snapshots and nightlies) `scalaReleasePatches` setting can be used to override the default mapping from minor versions to their patch numbers. By default, when compiling with `3.x1.y1` compiler and `-Yscala-release 3.x2` flag, the standard library version would be the same as the version of the compiler if `x1 == x2` and `3.x2.0` otherwise.

See https://github.com/sbt/sbt/discussions/6812 for more details.


lightbend-cla-validator:
Hi @prolativ,

eed3si9n:
Thanks for the contribution!

I have an idea that might simplify this. What do you think about adding a new key like `compileTimeScalaVersion`, instead of changing the way POM is written?
That way `scalaVersion` can remain to be the runtime Scala version. Another thing that would enable (and it might be missing from the current impl?) is being able to test this with using subproject `dependsOn(...)`.

In your example, we should be able to write:

```scala
ThisBuild / scalaVersion := "3.0.2"

lazy val bar1 = project.in(file("bar1"))
  .dependsOn(foo1)
```

`foo1` could look like:

```scala
lazy val foo1 = project.in(file("foo1"))
  .settings(
    name := foo1Name,
    compileTimeScalaVersion := "3.1.2-RC1",
    scalacOptions ++= Seq("-Yscala-release", "3.0"),
  )
```

This way, we can always assume `scalaVersion` to be the "announced" Scala version.

prolativ:
Wouldn't the solution with `compileTimeScalaVersion` cause some unnecessary duplication of configuration? Maybe I missed something but the problem I was fighting with was that because `scalaVersion` is a `Setting` it cannot depend on `scalacOptions`, which is a `Task`. So effectively one would have to specify it explicitly too, e.g.
```scala
lazy val foo1 = project.in(file("foo1"))
  .settings(
    name := foo1Name,
    compileTimeScalaVersion := "3.1.2-RC1",
    scalaVersion := "3.0.2"
    scalacOptions ++= Seq("-Yscala-release", "3.0"),
  )
```
A yet alternative approach would be to have both `compileTimeScalaVersion` and `scalaVersion` to be specified by the user (with the former defaulting to the latter for compatibility) without forcing users to add `-Yscala-release` to `scalacOptions` and do it for them under the hood. Of course we would then have to handle situations when they do it anyway (because this should work according to the compiler's documentation) and probably report a warning if there is a redundancy or an error if there's a mismatch.

Regarding `dependsOn` the third test case (`foo3`, `bar3`) is already supposed to check that this works but indeed I didn't change the implementation to handle this in any special way. Does this mean there's something wrong with this test?

eed3si9n:
Yea. I like the explicit nature of that.

```scala
lazy val foo1 = project.in(file("foo1"))
  .settings(
    name := foo1Name,
    compileTimeScalaVersion := "3.1.2-RC1",
    scalaVersion := "3.0.2",
    scalacOptions ++= Seq("-Yscala-release", "3.0"),
  )
```

Looks fine to me. The user should be able to move either of the settings to `ThisBuild`.

```scala
ThisBuild / compileTimeScalaVersion := "3.1.2-RC1"

lazy val foo1 = project.in(file("foo1"))
  .settings(
    name := foo1Name,
    scalaVersion := "3.0.2",
    scalacOptions ++= Seq("-Yscala-release", "3.0"),
  )
```

Using `compileTimeScalaVersion.?.value` at subproject scope would then make the lookup `Option[String]`, and then fallback to `scalaVersion.value`. This way the guts of the change would move into scalaInstance creation (https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/Defaults.scala#L1089) similar to having `scalaHome` set.

prolativ:
I experimented with this a bit and I also had some internal discussions with other people responsible for the development of `-Yscala-release` flag.
Our thoughts are that it would be probably better to semantically preserve `scalaVersion` as the version of the compiler used to compile a project and introduce another setting which could be named `scalaReleaseVersion` (suggestions of a better name are welcome) and which would be used instead of `scalaVersion` in 3 situations:
* for setting the value of `-Yscala-release`
* setting the version of the standard library declared as a dependency in POM and Ivy files
* setting the version of the standard library used to run main methods (e.g. with `sbt run`) and to execute tests

The reasoning about preserving the old semantics of `scalaVersion` is that:
* `-Yscala-release` is still considered experimental and its behaviour might change in some way
* Plugins might rely on the value of `scalaVersion` expecting it to be the actual version of the compiler (e.g. to check what compiler flags are supported in a given version)

Out of the 3 usages of `scalaReleaseVersion` the last point is something I was not thinking about before but I realized it's needed to make things consistent. To see why, let's assume we didn't include this requirement.
Now let's consider a situation when we publish a sample project compiled in the forward compatibility mode for scala 3.0.x and it contains some main method. If someone then tries to run it e.g. via coursier (or uses it from some other project using scala 3.0.x) then the old 3.0.x standard library will be used in runtime because it's the version listed as a dependency in POM and Ivy files. On the contrary if we run our main method with `sbt run` it will use the same stdlib in runtime that was used to compile the project (which currently would have to be 3.1.2-RC1). Although this shouldn't be a common case, it might turn out that the main method might work correctly with the new stdlib on the classpath but crash when the older stdlib is used. So it would be much easier to spot the bug if `sbt run` also used the older stdlib.
Similarly, if the compiler's outputs are supposed to work even with the old stdlib, it would make sense to use it when running all the tests of our project. Currently it seems already possible to set a different version of the compiler used in `Test` scope than in `Compile` but
* creators of libraries would have to remember to set `Test / scalaVersion` to match `Compile / scalaReleaseVersion`
* it wouldn't be possible to take advantage of new features and bug fixes introduced in the newer compiler while writing tests (which might get really annoying if e.g. test frameworks start adding nice features which use new parts of the compiler's metaprogramming API).

However I'm not sure how to properly split the compile time scala version from the one used at runtime for a given module. Would I have to have 2 separate `scalaIntance`s or is there a simpler way to achieve that?

eed3si9n:
Thanks for the update.

The idea of reusing the `--release` terminology feels jarring to me (even though I am relatively familiar with `--release` flag in both `javac` and `scalac`), in part because it becomes vague what release means. In JDK context `--release 8` means I am targeting JDK 8 ABI using different Java or Scala source from JDK 11, and it's clear that it refers to one of [JDK releases](https://www.java.com/releases/), however in hindsight it's also a self-centered naming by JDK authors.

For Scala + sbt, something like

```scala
targetScalaVersion := "3.0.2"
```

might be more clear that the library foo that you're building is now going to target Scala version 3.0.2 so the downstream library users can stay on 3.0.

prolativ:
Yeah, naming things is hard. Even for `-Yscala-release` itself there were quite a lot of discussions but finally it was agreed to simply align what scala already had.
Basically both `-release` and `-target` (or `-Xtarget` for scala 3 from some point on) do the same thing, which is setting the minimal supported version of JDK, but `-release` additionally performs some checks to ensure this should really work with the older JDK. That's why usage of `-(X)target` is discouraged. Hence we decided not to name the flag `-scala-target`, which was the initial idea.
So personally I would say that we still might prefer `scalaReleaseVersion` because it's more consistent while `targetScalaVersion` might express the intent just slightly better and it introduces more confusion.

Some other potential candidates for the name were `minimalScalaVersion` and `scalaCompatVersion` but none of them seems to express all the 3 intents described above and give the intuition that this setting should be typically set to a version of scala with minimal required minor version (because we want the compiler's outputs to be consumable for the widest possible range of different compiler versions) but maximal patch version (to get the standard library with more fixes).

But whatever name we choose I'm still wondering what would be the most appropriate way to be able to have a separate version of scala at compile time and at runtime. Would adding a new task key like `runtimeScalaInstance` be the right way to go? As `scalaInstance` depends on some other tasks I would then have to duplicate their keys in a similar way just to propagate the proper version of the standard library? And I'm not sure how this would work with `scalaHome`. Would we have to arbitrarily decide on whether it sets the scala version used for compiling or for running code?

armanbilge:
Sorry if I'm interrupting with a stupid question, but how would this work with `+` and `++`? Would we need to set a `compileTimeScalaVersion` for Scala 2 as well?

Edit: sorry, I didn't read the latest comments properly. I see you are going in a different direction now. But I'm still confused how `+` and `++` would interact with any of this. E.g. a `Test / scalaVersion` that is different from the `Compile / scalaVersion`.

prolativ:
The exact names of the settings are still being discussed https://contributors.scala-lang.org/t/improving-scala-3-forward-compatibility/5298/14
Scala 3 is going to have a slightly different approach to cross-compilation than scala 2. Instead of using different compilers to produce artifacts suitable for compilers with a matching minor version, one version of the compiler will be used to produce artifacts consumable for compilers with different minor versions. 
In the future we would like to get rid of the use case for `+` - in most situations you should be able to use the newest available version of scala 3 compiler to produce artifacts that would be consumable for scala 2.13 and 3.x.
Regarding `++` - this seems to be another argument for not using `compileTimeScalaVersion` - we should keep the old semantics of `scalaVersion` to signify the version of scalac and introduce a new setting with a different name specifying the value of `-Yscala-release` flag and setting the version of stdlib used at compile time (for dependent projects) and at runtime.
As one of the flags will use the other as a fallback you don't need to specify it if the targeted scala version is the same as the version of the compiler. And, as mentioned above, for scala 2 you can target only the same minor version as the version of the compiler. So effectively this PR shouldn't affect scala 2 builds at all.

armanbilge:
Thanks for the reply! For sure, in the future we will all be happy to leave behind `+` :) but for now, I'm worried about all the builds/CI we have that currently relies on `+` for both Scala 2 and Scala 3.

I'm glad to see sbt getting support for this, I was thinking about doing something similar in a user-land plugin:
- https://github.com/typelevel/sbt-typelevel/issues/138

prolativ:
@eed3si9n could we carry on with this scheme https://contributors.scala-lang.org/t/improving-scala-3-forward-compatibility/5298/23 ?

eed3si9n:
> @eed3si9n could we carry on with this scheme https://contributors.scala-lang.org/t/improving-scala-3-forward-compatibility/5298/23 ?

Yea. Thanks for the update!

prolativ:
Sorry it took that long.

I reimplemented my previous solution to achieve the desired semantics:
* newer stdlib used at compile time of a given project (for both main and test)
* older stdlib used at runtime (for both for main and test)
* older stdlib used as a transitive dependency when some other project depends on this one
* newer stdlib used in `console` - here compilation is mixed with execution so there's probably no way to use different versions of stdlib for these two things

`-Yscala-release` flag got renamed to `-scala-output-version` starting from Scala 3.1.2-RC2 so only the new name is handled.

@eed3si9n does this seem to make sense?
I might add some negative tests for handling invalid combinations of settings and flags.
As the compiler is going to have another release candidate (3.1.2-RC3) stable Scala 3.1.2 is most likely going to be published next week. Is there a chance to get a new release of sbt including my changes by that time?

eed3si9n:
> @eed3si9n does this seem to make sense?

I think so.

> As the compiler is going to have another release candidate (3.1.2-RC3) stable Scala 3.1.2 is most likely going to be published next week. Is there a chance to get a new release of sbt including my changes by that time?

We can get a milestone version out sbt 1.7.0-M1 probably, but typically we go through weeks of RC cycle for feature releases like sbt 1.7.0.

eed3si9n:
Also could you target 1.7.x branch, which has mima checks please?

prolativ:
OK, I think a milestone release should be enough to let people experiment with this feature.

There was one test failing in the CI. After having a look at the implementation I was quite surprised that this worked before.
In short the problem was with `plugins/dotty` scripted test. `scalaVersion` is set to `3.0.0-M1` there, which for some reason is special cased so `scala-tool` dependencies are not included in the update report. However `scalaInstance` was somehow constructed successfully. It looks like it gets initialized earlier from `scalaProvider` and cached even though the scala versions are different later. This did not work like this for `runtimeScalaInstance` however and it was always initialized from update, which caused a crash. So now `runtimeScalaInstance` is taken from `scalaInstance` if `scalaOutputVersion` is not defined. This not only makes the test pass now (although I'm still not sure why this worked before) but also seems potentially good for compatibility - if for any reason someone overwrote the definitions of `scalaInstance` but was unaware of existance of `scalaOutputVersion` setting this would also overwrite `runtimeScalaInstance`.

prolativ:
Also I'm not sure why the tests seem not to have run (I can see only the CLA validation in CI)

prolativ:
@eed3si9n  It looks like all the tests, including MiMa checks, are passing. Any more comments to the code itself? Or is there anything more I need to do to get this merged?

eed3si9n (approved):
LGTM
Thanks for the contribution

PR #6820, opened 2022-02-21
launcher: debian package to depend on curl | wget

tsawada:
also give informative error message when neither is available, rather than

```
$ sbt                                                                                      
downloading sbt launcher 1.6.1
/bin/mv: cannot stat '/root/.cache/sbt/boot/sbt-launch/1.6.1/sbt-launch-1.6.1.jar.temp': No such file or directory
failed to download launcher jar: https://repo1.maven.org/maven2/org/scala-sbt/sbt-launch/1.6.1/sbt-launch-1.6.1.jar
```

### DIFF
--- a/launcher-package/build.sbt
+++ b/launcher-package/build.sbt
@@ -191,7 +191,7 @@ val root = (project in file(".")).
     },
     // Used to have "openjdk-8-jdk" but that doesn't work on Ubuntu 14.04 https://github.com/sbt/sbt/issues/3105
     // before that we had java6-runtime-headless" and that was pulling in JDK9 on Ubuntu 16.04 https://github.com/sbt/sbt/issues/2931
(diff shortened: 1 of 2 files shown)


tsawada:
My understanding is that current deb package only has the shellscript launcher, and thus `curl | wget` is a must. Otherwise it should be `debianPackageRecommends`.

Or maybe just the error message is good enough.

eed3si9n (approved):
Thanks!

tsawada:
@eed3si9n Thanks for your quick review! Is there anything I can do for the failed CI test? It doesn't seem quite related to my change.
https://github.com/sbt/sbt/runs/5277938505?check_suite_focus=true#step:15:460

eed3si9n:
Yea. Likely unrelated. I am happy to merge as-is, assuming this is a transient error.

PR #6822, opened 2022-02-25
feat: add framework name to the test classes BSP request

kpodsiad:
Implementation of https://github.com/build-server-protocol/build-server-protocol/pull/300

I wanted to do as few changes as possible to not break compatibility or change the codebase completely. For now, I'll create this PR as a draft to see the test results and your opinion about it.

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
@@ -242,7 +242,7 @@ object BuildServerProtocol {
       val filter = ScopeFilter.in(workspace.scopes.values.toList)
       Def.task {
         val items = bspScalaTestClassesItem.result.all(filter).value
-        val successfulItems = anyOrThrow(items)
+        val successfulItems = anyOrThrow(items).flatten.toVector
         val result = ScalaTestClassesResult(successfulItems.toVector, None)
         s.respondEvent(result)
       }
@@ -840,15 +840,30 @@ object BuildServerProtocol {
     }
   }
 
-  private def scalaTestClassesTask: Initialize[Task[ScalaTestClassesItem]] = Def.task {
-    val testClasses = Keys.definedTests.?.value
-      .getOrElse(Seq.empty)
-      .map(_.name)
-      .toVector
-    ScalaTestClassesItem(
-      bspTargetIdentifier.value,
-      testClasses
-    )
+  private def scalaTestClassesTask: Initialize[Task[Seq[ScalaTestClassesItem]]] = Def.task {
+    Keys.definedTests.?.value match {
+      case None => Vector.empty
+      case Some(definitions) =>
+        val fingerprints = Keys.loadedTestFrameworks.?.value
+          .getOrElse(Map.empty)
+          .values
+          .flatMap { framework =>
+            framework.fingerprints().map(fingerprint => (fingerprint, framework))
+          }
+          .toMap
+
+        definitions
+          .groupBy(defn => fingerprints.get(defn.fingerprint))
+          .map {
+            case (framework, definitions) =>
+              ScalaTestClassesItem(
+                bspTargetIdentifier.value,
+                definitions.map(_.name).toVector,
+                framework.map(_.name())
+              )
+          }
+          .toSeq
+    }
   }
 
   private def scalaMainClassesTask: Initialize[Task[ScalaMainClassesItem]] = Def.task {
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -419,7 +419,7 @@ object Keys {
   val bspBuildTargetScalacOptions = inputKey[Unit]("").withRank(DTask)
   val bspBuildTargetScalacOptionsItem = taskKey[ScalacOptionsItem]("").withRank(DTask)
   val bspScalaTestClasses = inputKey[Unit]("Corresponds to buildTarget/scalaTestClasses request").withRank(DTask)
-  val bspScalaTestClassesItem = taskKey[ScalaTestClassesItem]("").withRank(DTask)
+  val bspScalaTestClassesItem = taskKey[Seq[ScalaTestClassesItem]]("").withRank(DTask)
   val bspScalaMainClasses = inputKey[Unit]("Corresponds to buildTarget/scalaMainClasses request").withRank(DTask)
   val bspScalaMainClassesItem = taskKey[ScalaMainClassesItem]("").withRank(DTask)
   val bspReporter = taskKey[BuildServerReporter]("").withRank(DTask)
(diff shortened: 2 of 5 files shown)


main-actions/src/main/scala/sbt/Tests.scala
```diff
@@ -39,6 +37,12 @@ sealed trait TestOption
 
 object Tests {
 
+  final case class TestDefinitionsWithFramework(
+      testPerFramework: Map[Framework, Seq[TestDefinition]]
+  ) {
+    def allDefinitions: Seq[TestDefinition] = testPerFramework.values.flatten.toVector
```
kpodsiad: Utility method because previously there was no grouping by the test framework. For not BSP purpose mapping `framework -> tests` isn't useful but I didn't want to create separate logic just for BSP. What do you think of it?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1338,7 +1341,7 @@ object Defaults extends BuildCommon {
   def singleTestGroup(key: Scoped): Initialize[Task[Seq[Tests.Group]]] =
     inTask(key, singleTestGroupDefault)
   def singleTestGroupDefault: Initialize[Task[Seq[Tests.Group]]] = Def.task {
-    val tests = definedTests.value
+    val tests = definedTests.value.allDefinitions
```
kpodsiad: and e.g. here information about frameworks is discarded. However later mapping `framework -> test` has to be computed, but I don't know if we could do better (without significant code repetition) since changing `Tests.Group` would break compatibility :/

kpodsiad:
Main question is: am I allowed to modify already defined keys or should I create new ones and include this logic here while leaving already existing keys untouched?

main-actions/src/main/scala/sbt/Tests.scala
```diff
@@ -531,8 +535,9 @@ object Tests {
       frameworks: Seq[Framework],
       analysis: CompileAnalysis,
       log: Logger
-  ): (Seq[TestDefinition], Set[String]) =
-    discover(frameworks flatMap TestFramework.getFingerprints, allDefs(analysis), log)
+  ): (TestDefinitionsWithFramework, Set[String]) = {
```
eed3si9n: Since `Tests` is under `sbt` package, I don't think we can break the binary compatibility during sbt 1.x series.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -840,15 +840,19 @@ object BuildServerProtocol {
...
+  private def scalaTestClassesTask: Initialize[Task[Seq[ScalaTestClassesItem]]] = Def.task {
+    Keys.definedTests.?.value match {
+      case None => Vector.empty
+      case Some(defined) =>
+        defined.testPerFramework.map {
+          case (framework, tests) =>
+            ScalaTestClassesItem(
+              bspTargetIdentifier.value,
+              tests.map(_.name).toVector,
+              framework.name
```
adpi2: You cannot change the type of `Keys.definedTests`. But here you can use `loadedTestFrameworks.value` to build a `Map[Fingerprint, Framework]` and then group each `definedTests.value` by framework.

adpi2 (changes requested):
Unfortunately we cannot change the return type of keys in the sbt `1.x` series because it could break the builds and plugins using that key.

I think you can group the `definedTests` by framework in the `BuildServerProtocol` implementation directly, as suggested below.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
kpodsiad: I saw your comment too late :/ I've already tweaked my initial implementation to not break compatibility by just adding other keys but I find your proposition far better, I'll change that

kpodsiad: The only key which s changed now is `bspScalaTestClassesItem`. Is changing bsp keys acceptable or just to be fully compliant it's better to create a new one? 
cc: @eed3si9n

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -840,15 +846,29 @@ object BuildServerProtocol {
...
+      case Some(defined) =>
+        val fingerprints = Keys.loadedTestFrameworks.value.values.flatMap { framework =>
+          framework.fingerprints().map(fingerprint => (fingerprint, framework))
+        }.toMap
+        defined
+          .map(
+            definition =>
+              TestDefinitionWithFramework(definition, fingerprints.get(definition.fingerprint))
+          )
+          .groupBy(_.framework)
```
adpi2: I don't think we need this `TestDefinitionWithFramework` anymore.

```suggestion
        defined
          .groupBy(def => fingerprints.get(definition.fingerprint))
```

kpodsiad:
There is an error in tests:
```
Compile / loadedTestFrameworks from Compile / bspScalaTestClassesItem ((sbt.internal.server.BuildServerProtocol.configSettings) BuildServerProtocol.scala:303)
Did you mean Test / loadedTestFrameworks ?
```
but I don't know how to deal with it.

adpi2:
`loadedTestFrameworks` is not implemented in the `Compiler` config.

You can work around this using `Keys.loadedTestFrameworks.?.value` that should give you a `Option[Seq[Framework]]` that you can flatten.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
eed3si9n: Assuming no plugins are consuming `bsp*` keys, I am ok with making mima exceptions for those.

adpi2: `scala-debug-adapter` does not depend on `bspScalaTestClassesItem` and I found no consumer in https://github.com/search?p=1&q=bspScalaTestClassesItem&type=Code.

So I think we are good with this one.

kpodsiad: Sorry, I've never used MiMa before - if `mainProj/mimaReportBinaryIssues` passes should I change something with mima or is it ok?

adpi2: It is unclear to me why the `mimaReportBinaryIssues` passes. But since it does I think there is nothing else to do.

eed3si9n: Amina and I started on sbt 2.x on develop branch, so mima has been turned off. Maybe we should create 1.7.x branch off of 1.6.x?
https://github.com/sbt/sbt/blob/develop/.github/workflows/ci.yml#L102

adpi2: I just created [branch 1.7.x](https://github.com/sbt/sbt/tree/1.7.x). @kpodsiad you can repoen this PR to merge into 1.7.x.

kpodsiad:
Superseded by https://github.com/sbt/sbt/pull/6830

Issue #5070, opened 2019-09-11
Create a means to add a repository to the default list at the project global level


JustinPihony:
Right now there is a way to override repositories entirely, but not append to the list. The feature I’m requesting is an ability to set a resolver for both plugins and project, which could then be checked into the codebase and used by subsequent developers seamlessly.

eed3si9n:
Thanks Justin. I think this is a good idea. See there are similar situation that's happening for SuperSafe (https://github.com/sbt/sbt/issues/4103, http://www.scalatest.org/supersafe#installation) where it needs to configure the resolver for both the metabuild and the proper build.

What do you think about adding a key like:

```scala
ThisBuild / includePluginResolvers := true
```

that would inject plugin level resolver to all subprojects?

JustinPihony:
Sounds good to me

PR #6849, opened 2022-03-23
includePluginResolvers should work for coursier resolutions

bjaglin:
It looks like https://github.com/sbt/sbt/pull/5094 didn't really address https://github.com/sbt/sbt/issues/5070 as the coursier behavior was not updated (only ivy which relies on `fullResolvers`).

I haven't found any active usage of `includePluginResolvers` in open-source, but https://github.com/gitter-badger/bqtest/commit/aa75c2b7c9bd624f97fdb157ae38f742048a4747 seems to confirm it does not work as expected.

### DIFF
--- a/main/src/main/scala/sbt/coursierint/CoursierRepositoriesTasks.scala
+++ b/main/src/main/scala/sbt/coursierint/CoursierRepositoriesTasks.scala
@@ -57,16 +57,7 @@ object CoursierRepositoriesTasks {
   private final val keepPreloaded = false // coursierKeepPreloaded.value
 
   def coursierResolversTask: Def.Initialize[sbt.Task[Seq[Resolver]]] = Def.task {
-    val bootResOpt = bootResolvers.value
-    val overrideFlag = overrideBuildResolvers.value
-    val result0 = bootResOpt.filter(_ => overrideFlag) match {
-      case Some(r) => r
-      case None =>
-        val extRes = externalResolvers.value
-        val isSbtPlugin = sbtPlugin.value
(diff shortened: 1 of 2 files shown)


main/src/main/scala/sbt/coursierint/CoursierRepositoriesTasks.scala
```diff
@@ -57,16 +57,7 @@ object CoursierRepositoriesTasks {
...
-    val overrideFlag = overrideBuildResolvers.value
-    val result0 = bootResOpt.filter(_ => overrideFlag) match {
-      case Some(r) => r
-      case None =>
-        val extRes = externalResolvers.value
-        val isSbtPlugin = sbtPlugin.value
-        if (isSbtPlugin) sbtResolvers.value ++ extRes
-        else extRes
-    }
+    val result0 = fullResolvers.value.filterNot(_ == projectResolver.value)
```
bjaglin, line 60: https://github.com/sbt/sbt/blob/ccc6cf681870f826e8403ba62a325fa287d118e3/main/src/main/scala/sbt/Defaults.scala#L2965-L2986

The filter is not strictly mandatory as https://github.com/coursier/sbt-coursier/blob/fb19697c573489d1ddbd6c3afd9d8609a70c4fec/modules/lm-coursier/src/main/scala/lmcoursier/internal/Resolvers.scala#L114-L115 seems to ignore `projectResolver`, but I thought it was better not to change the behavior of `csrResolvers`, particularly if this is backported to 1.6.x.

eed3si9n (approved):
Thanks!

PR #6853, opened 2022-03-27
[1.7.x] includePluginResolvers should work for coursier resolutions

eed3si9n:
This is a backport of https://github.com/sbt/sbt/pull/6849

It looks like https://github.com/sbt/sbt/pull/5094 didn't really address https://github.com/sbt/sbt/issues/5070 as the coursier behavior was not updated (only ivy which relies on `fullResolvers`).

I haven't found any active usage of `includePluginResolvers` in open-source, but https://github.com/gitter-badger/bqtest/commit/aa75c2b7c9bd624f97fdb157ae38f742048a4747 seems to confirm it does not work as expected.


bjaglin:
Thanks for the backport @eed3si9n !

PR #6858, opened 2022-03-31
feat: implement BSP's environment requests

kpodsiad:
Implementation of https://build-server-protocol.github.io/docs/extensions/jvm.html.

Could you guide me on what will be the best way to test these endpoints?


main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -315,6 +315,31 @@ object BuildServerProtocol {
     bspBuildTargetCompileItem := bspCompileTask.value,
     bspBuildTargetRun := bspRunTask.evaluated,
     bspBuildTargetScalacOptionsItem := scalacOptionsTask.value,
+    bspBuildTargetJVMRunEnvironment := Def.inputTaskDyn {
+      val s = state.value
+      val targets = spaceDelimited().parsed.map(uri => BuildTargetIdentifier(URI.create(uri)))
+      val workspace = bspFullWorkspace.value.filter(targets)
+      val filter = ScopeFilter.in(workspace.scopes.values.toList)
```
kpodsiad: I copied this from another request, how can I extract this to a helper function?

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -315,6 +315,31 @@ object BuildServerProtocol {
...
+        s.respondEvent(result)
+      }
+    }.evaluated,
+    bspBuildTargetJVMTestEnvironment := Def.inputTaskDyn {
+      val s = state.value
+      val targets = spaceDelimited().parsed.map(uri => BuildTargetIdentifier(URI.create(uri)))
+      val workspace = bspFullWorkspace.value.filter(targets)
+      val filter = ScopeFilter.in(workspace.scopes.values.toList)
+      Def.task {
+        val items = bspBuildTargetJvmEnvironmentItem.result.all(filter).value
```
kpodsiad, line 294: These two tasks differ only here, some helper method might be introduced to reduce code repetition

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -648,6 +687,33 @@ object BuildServerProtocol {
...
+    val target = Keys.bspTargetIdentifier.value
+    val baseDir = Keys.baseDirectory.value.toURI().toString()
+    val jvmOptions = Keys.javaOptions.value.toVector
+    val env = envVars.value
+    val externalDependencyClasspath = Keys.externalDependencyClasspath.value
+
+    val internalDependencyClasspath = for {
+      (ref, configs) <- bspInternalDependencyConfigurations.value
+      config <- configs
+    } yield ref / config / Keys.classDirectory
```
kpodsiad: Is this a valid way to extract the classpath? I copied this from `scalacOptionsTask` (if yes then it can be probably refactored out)

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -878,6 +944,7 @@ object BuildServerProtocol {
     }
   }
 
+  // here
```
kpodsiad: leftover, I'll remove it with another commit

kpodsiad (commented):
Sorry for all of these questions, I'm still pretty new to the sbt's Tasks and scoping and other stuff - that's why there is a lot of code repetition. I'll appreciate suggestions of any kind on how to refactor common parts.

kpodsiad:
@adpi2 are there somewhere BSP's tests? I searched for them but I found nothing.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
adpi2: `.value` and `.parsed` are macros and they can only be called inside `Def.setting`, `Def.task`, `Def.inputTask`.

So here you can try something like that:

```scala
bspBuildTargetJvmRunEnvironment := bspInputTask { (state, filter) =>
  Def.task {
    val items = bspBuildTargetJvmEnvironmentItem.result.all(filter).value
    val successfulItems = anyOrThrow(items)
    val result = JvmRunEnvironmentResult(successfulItems.toVector, None)
    state.respondEvent(result)
  }
}
```

where `bspInputTask` is:

```scala
private def bspInputTask[T](taskImpl: (State, ScopeFilter) => Def.Initialize[Task[T]]): Def.Initialize[InputTask[T]] = 
  Def.inputTaskDyn {
    val s = state.value
    val targets = spaceDelimited().parsed.map(uri => BuildTargetIdentifier(URI.create(uri)))
    val workspace = bspFullWorkspace.value.filter(targets)
    val filter = ScopeFilter.in(workspace.scopes.values.toList)
    taskImpl(s, filter)
  }
```

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
adpi2: They also differ in the type of result: `JvmTestEnvironmentResult` vs `JvmRunEnvironementResult`. So I don't think it is worth doing anything here.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
adpi2: The easiest way to get the classpath is to use `Keys.fullClasspath.value`. The problem is that it triggers the compilation and it fails if there is any compiler error. That's why we could not use it in `scalacOptions`. We want to import the build even if the compilation fails.

Do we want the `jvmRunEnvironement` to fail if the compilation fails or not? If the answer is no then this is the best way to extract the classpath that we have.

To refactor it you can use:
```scala
// does not trigger compilation
private def bspClasspathTask: Def.Initialize[Task[Vector[URI]]] = Def.taskDyn {
  val externalDependencyClasspath = Keys.externalDependencyClasspath.value

  val internalDependencyClasspath = for {
    (ref, configs) <- bspInternalDependencyConfigurations.value
    config <- configs
  } yield ref / config / Keys.classDirectory

  Def.task {
    val classpath = internalDependencyClasspath.join.value.distinct ++
      externalDependencyClasspath.map(_.data)
    classpath.map(_.toURI).toVector
  }
}
```

and call it like that:
```scala
Def.task {
  val target = Keys.bspTargetIdentifier.value
  val baseDir = Keys.baseDirectory.value.toURI().toString()
  val jvmOptions = Keys.javaOptions.value.toVector
  val env = envVars.value
  val classpath = bspClasspath.value
  JvmEnvironmentItem(target, classpath, jvmOptions, baseDir, env)
}
```

kpodsiad: > Do we want the `jvmRunEnvironement` to fail if the compilation fails or not?

According to the [spec](https://build-server-protocol.github.io/docs/extensions/jvm.html#test-environment-request) `The client should ask for it right before test/run execution, after all the targets are compiled.` so using `Keys.fullClasspath.value` might be a better way. The project should compile because otherwise, Metals won't be able to run it.

Speaking about protocol, you could look at https://github.com/build-server-protocol/build-server-protocol/pull/312 @adpi2, I made a small fixes to the docs.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
kpodsiad: These types are identical, I wonder if we need different types for both requests. The only reason I think of is being compliant with specification so, in the future, other people won't be confused why there is only one type for two endpoints.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
kpodsiad: Thanks! With `bspInputTask` I was able to refactor some of the tasks!

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
kpodsiad: I'll leave different types for the sake of explicitness

adpi2:
> @adpi2 are there somewhere BSP's tests? I searched for them but I found nothing.

Yes we have some tests in the [`BuildServerTest`](https://github.com/sbt/sbt/blob/develop/server-test/src/test/scala/testpkg/BuildServerTest.scala) class. They run on the build defined [here](https://github.com/sbt/sbt/tree/develop/server-test/src/server-test/buildserver) that you can edit to add some environment variables.

To run the tests:
```
sbt:sbtRoot> serverTestProj/testOnly testpkg.BuildServerTest
```

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
adpi2: > So using Keys.fullClasspath.value might be a better way

It makes sense.

Thanks for pinging me on this build-server-protocol PR, I would have missed it otherwise.

kpodsiad:
2 tests have exceeded the timeout, but it doesn't seem like it's caused by my changes, at least not directly. 
I rebased my changes to rerun CI actions.

kpodsiad:
@adpi2 do you have an idea why some actions are never-ending/seem to have deadlock? I checked my changes and I didn't find anything suspicious.

adpi2 (approved):
Looks great and thanks for cleaning the code duplication.

Those changes are not breaking the forward compatibility. Could you back-port them to `1.7.x` branch?

kpodsiad:
> Those changes are not breaking the forward compatibility. Could you back-port them to `1.7.x` branch?

Should I create another PR that targets `1.7.x` or should I change the target in this one?

adpi2:
Yeah another PR would be good.

PR #6870, opened 2022-04-09
Restore watchOnTermination callbacks

### LINKED ISSUE #6861: `watchOnTermination` setting not respected
## steps

Using SBT 1.6.2, add the following `watchOnTermination` setting to a `build.sbt` file:

```scala
ThisBuild / watchOnTermination := { (action, cmd, times, state) =>
  println(s"watchOnTermination: action: $action, cmd: $cmd, times: $times, state: $state")
  state
}
```

Then load SBT in the project's directory, run `~compile`, and then hit enter to stop triggered execution.

## problem

My `watchOnTermination` function is not called after I hit enter to stop triggered execution

## expectation

The `watchOnTermination` function should be called when I hit enter and I should see my `println` message in the console

## notes

Apologies if I'm misunderstanding how this setting is supposed to work. I see that it's [not documented with the other `watch*` settings](https://www.scala-sbt.org/1.x/docs/Triggered-Execution.html#Configuration), I just learned about it by reading through the source code.

eatkins:
This PR restores the watchOnTermination callback behavior to fix #6861. I assume that the regression occurred when I refactored watch to support the thin client but I didn't bother investigating the old behavior and just made the changes necessary to make the callbacks be invoked.

Fixes #6861


eatkins:
The combined diff is a little ugly because I moved the original watch/on-termination test because it had a misleading name. The second commit is the important one.

eed3si9n (approved):
Thanks!

PR #6871, opened 2022-04-10
feat: update Problem to account for related information and code

ckipp01:
This PR makes changes to the existing `xsbti.Problem` to account for an
optional diagnostic code that the compiler can return for a given
diagnostic and also related information.

### Example of `DiagnosticCode`
Given a piece of code like:

```scala
try {}
```

You'll receive the following:

```
-- [E002] Syntax Warning: /Users/ckipp/Documents/scala-workspace/dotty-error-index/examples/002_EmptyCatchAndFinallyBlockID.scala:3:2
3 |  try {}
  |  ^^^^^^
  |  A try without catch or finally is equivalent to putting
  |  its body in a block; no exceptions are handled.
```

The `E002` here is the actual code. Right now there would be no
description.

### Example of `DiagnosticRelatedInformation`

Some diagnostics have multiple positions that they need to represent.
You can see an example of this
[here](https://github.com/lampepfl/dotty/pull/14002) in Dotty with the
use of inlining. Instead of needing to rely on including all of that
information in the diagnostic message it can now be extracted out into
a `DiagnosticRelatedInformation`.

These changes reference the conversation in https://github.com/sbt/sbt/discussions/6868

### DIFF
--- a/internal/util-interface/src/main/java/xsbti/DiagnosticRelatedInformation.java
+++ b/internal/util-interface/src/main/java/xsbti/DiagnosticRelatedInformation.java
@@ -1,0 +1,20 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package xsbti;
+
+/**
+ * Related information for a given diagnostic. At times this can be another place in your code
+ * contributing to the diagnostic or just relevant code relating to the diagnostic.
+ */
+public interface DiagnosticRelatedInformation {
+  /** Position of the related information */
+  Position position();
+
+  /** Message indicating why this related information is attached to the diagnostic. */
+  String message();
+}
--- a/internal/util-interface/src/main/java/xsbti/Problem.java
+++ b/internal/util-interface/src/main/java/xsbti/Problem.java
@@ -7,6 +7,8 @@
 
 package xsbti;
 
+import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
 
 public interface Problem {
@@ -25,5 +27,25 @@ public interface Problem {
    */
   default Optional<String> rendered() {
     return Optional.empty();
+  }
+
+  /**
+   * The unique code attached to the diagnostic being reported.
+   *
+   * <p>NOTE: To avoid breaking compatibility we provide a default to account for older Scala
+   * version that do not have codes.
+   */
+  default Optional<DiagnosticCode> diagnosticCode() {
+    return Optional.empty();
+  }
+
+  /**
+   * The possible releated information for the diagnostic being reported.
+   *
+   * <p>NOTE: To avoid breaking compatibility we provide a default to account for older Scala
+   * version that do not have the concept of "related information".
+   */
+  default List<DiagnosticRelatedInformation> diagnosticRelatedInforamation() {
+    return Collections.emptyList();
   }
 }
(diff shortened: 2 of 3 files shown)


internal/util-interface/src/main/java/xsbti/Problem.java
```diff
@@ -26,4 +28,24 @@ public interface Problem {
...
+    return Optional.empty();
+  }
+
+  /**
+   * The possible releated information for the diagnostic being reported.
+   *
+   * <p>NOTE: To avoid breaking compatibility we provide a default to account for older Scala
+   * version that do not have the concept of "related information".
+   */
+  default List<DiagnosticRelatedInformation> diagnosticReleatedInforamation() {
```
smarter:
```suggestion
  default List<DiagnosticRelatedInformation> diagnosticRelatedInformation() {
```

internal/util-interface/src/main/java/xsbti/DiagnosticRelatedInformation.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+package xsbti;
+
+/**
+ * Related information for a given diagnostic. At times this can be another place in your code
+ * contributing to the diagnostic or just relevant code relating to the diagnostic.
+ */
+public interface DiagnosticRelatedInformation {
+  /** Position of the related inforamtion */
```
smarter:
```suggestion
  /** Position of the related information */
```

internal/util-interface/src/main/java/xsbti/DiagnosticRelatedInformation.java
```diff
@@ -0,0 +1,20 @@
...
+  Position position();
+
+  /** Message indicating why this related information is attatched to the diagnostic. */
```
smarter:
```suggestion
  /** Message indicating why this related information is attached to the diagnostic. */
```

internal/util-interface/src/main/java/xsbti/Problem.java
```diff
@@ -26,4 +28,24 @@ public interface Problem {
   default Optional<String> rendered() {
     return Optional.empty();
   }
+
+  /**
+   * The unique code attatched to the diagnostic being reported.
```
smarter:
```suggestion
   * The unique code attached to the diagnostic being reported.
```

internal/util-interface/src/main/java/xsbti/Problem.java
```diff
@@ -26,4 +28,24 @@ public interface Problem {
...
+   *
+   * <p>NOTE: To avoid breaking compatibility we provide a default to account for older Scala
+   * version that do not have codes.
+   */
+  default Optional<DiagnosticCode> diagnosticCode() {
```
smarter, line 38: Is this enough information to get sbt to produce errors that look like what dotty natively produce without us having to pass in a `rendered` string? That would help solve https://github.com/lampepfl/dotty/issues/14691

eed3si9n:
@ckipp01 Thanks for the contribution. Could you target 1.7.x branch instead which runs mima, plz?

internal/util-interface/src/main/java/xsbti/Problem.java
ckipp01: Oof, good catch. Thanks.

ckipp01:
> @ckipp01 Thanks for the contribution. Could you target 1.7.x branch instead which runs mima, plz?

Sure, the rebase isn't clean, so it's just easier for me to close and open another. I'll go ahead and do that.

PR #6874, opened 2022-04-12
feat: update Problem to account for related information and code

ckipp01:
_Follow up to https://github.com/sbt/sbt/pull/6871 (sorry had issues with rebasing and this was faster)_
This PR makes changes to the existing `xsbti.Problem` to account for an
optional diagnostic code that the compiler can return for a given
diagnostic and also related information.

Given a piece of code like:

```scala
try {}
```

You'll receive the following:

```
-- [E002] Syntax Warning: /Users/ckipp/Documents/scala-workspace/dotty-error-index/examples/002_EmptyCatchAndFinallyBlockID.scala:3:2
3 |  try {}
  |  ^^^^^^
  |  A try without catch or finally is equivalent to putting
  |  its body in a block; no exceptions are handled.
```

The `E002` here is the actual code. Right now there would be no
description.

Some diagnostics have multiple positions that they need to represent.
You can see an example of this
[here](lampepfl/dotty#14002) in Dotty with the
use of inlining. Instead of needing to rely on including all of that
information in the diagnostic message it can now be extracted out into
a `DiagnosticRelatedInformation`.

These changes reference the conversation in #6868

### DIFF
--- a/internal/util-interface/src/main/java/xsbti/Problem.java
+++ b/internal/util-interface/src/main/java/xsbti/Problem.java
@@ -7,6 +7,8 @@
 
 package xsbti;
 
+import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
 
 public interface Problem {
@@ -25,5 +27,25 @@ public interface Problem {
    */
   default Optional<String> rendered() {
     return Optional.empty();
+  }
+
+  /**
+   * The unique code attached to the diagnostic being reported.
+   *
+   * <p>NOTE: To avoid breaking compatibility we provide a default to account for older Scala
+   * versions that do not have codes.
+   */
+  default Optional<DiagnosticCode> diagnosticCode() {
+    return Optional.empty();
+  }
+
+  /**
+   * The possible releated information for the diagnostic being reported.
+   *
+   * <p>NOTE: To avoid breaking compatibility we provide a default to account for older Scala
+   * versions that do not have the concept of "related information".
+   */
+  default List<DiagnosticRelatedInformation> diagnosticRelatedInforamation() {
+    return Collections.emptyList();
   }
 }
--- a/internal/util-interface/src/main/java/xsbti/DiagnosticCode.java
+++ b/internal/util-interface/src/main/java/xsbti/DiagnosticCode.java
@@ -1,0 +1,23 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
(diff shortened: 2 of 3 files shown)


internal/util-interface/src/main/java/xsbti/Problem.java
```diff
@@ -26,4 +28,24 @@ public interface Problem {
...
     return Optional.empty();
   }
+
+  /**
+   * The unique code attached to the diagnostic being reported.
+   *
+   * <p>NOTE: To avoid breaking compatibility we provide a default to account for older Scala
+   * version that do not have codes.
+   */
+  default Optional<DiagnosticCode> diagnosticCode() {
```
ckipp01, line 38: Carrying your question over from the other pr @smarter.

> Is this enough information to get sbt to produce errors that look like what dotty natively produce without us having to pass in a rendered string? That would help solve lampepfl/dotty#14691

That's a good question, and it sort of depends. The short answer right now is I'm not sure. Let me dig into this a bit and I'll reply back with some more details and thoughts.

ckipp01: Actually this issue seems to be fixed already in zinc https://github.com/sbt/zinc/pull/1082

smarter: Sort of, the fix means that errors coming from mapped position bypass the renderer, but that means they get rendered by sbt which lacks the error code and displays error differently from scala 3, so it makes things pretty inconsistent.

ckipp01: huh, then they miss out on the `explain` as well even don't they? The one part that I'm not sure yet is about the `explain`. Since the `explain` is attached to the message, not the actual code, and at times the message/explain is dynamically put together for different cases that re-use the same code. So it goes away a _bit_ from the actual meaning of the description of the code, so I'm not sure we can stick the explain text in there. wdyt? I mean if we can, then sure then it'd be up to the tool to take the code, the description (explain), and any related information and put together the rendered output.

smarter: I don't see a problem with putting the explain part in there.

ckipp01:
Looks like the single run that failed sort of flaked out completely.

eed3si9n (approved):
LGTM
Thanks for the contribution

deusaquilus:
Thanks so much @ckipp01 and @eed3si9n!!! Having per-inline-site info will be a really, really big bonus for diagnostics in Quill and everything that takes advantage of Scala 3 meta-programming!

PR #6876, opened 2022-04-13
Raise error for undiscovered mainClass

erikvanoosten:
This PR lets SBT raise a build error when the `mainClass` is not in the list of discovered main classes.
For those rare cases where the main class comes from somewhere else, this also includes an option to suppress the error.

This is a follow up of https://github.com/sbt/sbt/discussions/6867 which describes the rationale.

There are a few open questions:
 - Is this the correct way to raise a build error?
 - Does this require a unit test? If so, where should it be added?

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -990,7 +990,14 @@ object Defaults extends BuildCommon {
     discoveredSbtPlugins := discoverSbtPluginNames.value,
     // This fork options, scoped to the configuration is used for tests
     forkOptions := forkOptionsTask.value,
-    selectMainClass := mainClass.value orElse askForMainClass(discoveredMainClasses.value),
+    allowUndiscoveredMainClass :== false,
+    selectMainClass :=
+      errorForUndiscoveredMainClass(
+        mainClass.value,
+        discoveredMainClasses.value,
+        allowUndiscoveredMainClass.value
+      ) orElse
+        askForMainClass(discoveredMainClasses.value),
     run / mainClass := (run / selectMainClass).value,
     mainClass := {
       val logWarning = state.value.currentCommand.forall(!_.commandLine.split(" ").exists {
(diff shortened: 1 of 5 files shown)


lightbend-cla-validator:
Hi @erikvanoosten,

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -990,7 +990,14 @@ object Defaults extends BuildCommon {
     discoveredSbtPlugins := discoverSbtPluginNames.value,
     // This fork options, scoped to the configuration is used for tests
     forkOptions := forkOptionsTask.value,
-    selectMainClass := mainClass.value orElse askForMainClass(discoveredMainClasses.value),
+    allowUndiscoveredMainClass :== false,
```
eed3si9n: https://www.scala-sbt.org/1.x/docs/Plugins-Best-Practices.html#Provide+default+values+in
Could you move the default value to somewhere globally scoped so

```scala
ThisBuild / allowUndiscoveredMainClass := true 
```

would work?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1890,6 +1897,23 @@ object Defaults extends BuildCommon {
...
+  ): Option[String] = {
+    mainClass match {
+      case None                                                                         => mainClass
+      case Some(mc) if discoveredMainClasses.contains(mc) || allowUndiscoveredMainClass => mainClass
+      case Some(mc) =>
+        val msg =
+          s"The main class '$mc' was not discovered so it may not exist.\n" +
+            "Run 'show discoveredMainClasses' to see the list of discovered main classes.\n" +
+            "If you are sure the main class is correct, use 'allowUndiscoveredMainClass := true' to suppress this error."
+        sys.error(msg)
```
eed3si9n, line 1919: > Is this the correct way to raise a build error?

Yes. Throwing exception in a task is fine.

eed3si9n:
> Does this require a unit test? If so, where should it be added?

We typically test these kind of thing by running an automated integration test called scripted test (https://www.scala-sbt.org/1.x/docs/Testing-sbt-plugins.html). Run-related scripted test should be placed here - https://github.com/sbt/sbt/tree/develop/sbt-app/src/sbt-test/run

You could do something like

```
$ copy-file changes/RunSuccess.scala src/main/scala/Run.scala
> run

# a non-existent main class should fail run task
$ copy-file changes/RunFailureMain.scala src/main/scala/Run.scala
-> run
```

erikvanoosten:
I was playing with the scripted test framework. Unfortunately I don't see how this works. How does the package task get hold of the mainClass? Does the universal-packager sbt plugin use these settingkeys from sbt global at all?

I am quite confused about the scopes. It appears `mainClass` needs to be in the `Compile` scope.
Also, I don't understand why in the scripted test, `discoveredMainClasses` stays empty.

I am starting to believe the changes I am after should be in the universal-packager sbt plugin, not in sbt itself.

eed3si9n:
> How does the package task get hold of the mainClass? 

https://github.com/sbt/sbt/blob/053834aea07a0bc89f88e0fdcce84a133b17340a/main/src/main/scala/sbt/Defaults.scala#L1735-L1748

> Does the universal-packager sbt plugin use these settingkeys from sbt global at all?

`UniversalPlugin` just makes `.zip` file. What you'd need to read up in sbt-native-packager is how the `mappings` is constructed. Typically it would use `JavaAppPackaging` "archetype":

https://github.com/sbt/sbt-native-packager/blob/2268343362812bbca6e49f15e05586821a101bc7/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaAppPackaging.scala#L131-L162

The above should indirectly invoke `packageBin`.

erikvanoosten:
This task is taking too much effort. I'll close this issue to keep the backlog clean.

PR #6882, opened 2022-04-17
Add SBTN_AUTO_COMPLETE environment variable

Philippus:
Applying this PR will add a check for the SBTN_AUTO_COMPLETE environment variable next to the existing check for  SBTC_AUTO_COMPLETE. Since sbtc was renamed to sbtn, I think it makes sense to also rename the environment variable. If we're not worried about backwards compatibility we can remove the check for SBTC_AUTO_COMPLETE.
Also fixed some typos.


eed3si9n (approved):
Thanks!

Issue #6777, opened 2022-01-04
sbt thinks that server is already booting because of this exception


a1russell:
## steps

Using sbt 1.6.1 on Windows 10. I am also using Amazon Corretto JDK 8.312, if that makes a difference. The issue is present in both Git Bash and cmd.

1. Open a terminal, cd to a project, and execute `sbt`.
2. Open another terminal, cd to the same project, and try executing `sbt` in any form, including `sbt --no-server`. This will give an error.

## problem

This is what I receive upon executing the second sbt command:
```
>sbt --no-server compile
sbt thinks that server is already booting because of this exception:
sbt.internal.ServerAlreadyBootingException: java.io.IOException: Could not create lock for \\.\pipe\sbt-load-8979016220299674466_lock, error 5
        at sbt.internal.BootServerSocket.newSocket(BootServerSocket.java:348)
        at sbt.internal.BootServerSocket.<init>(BootServerSocket.java:294)
        at sbt.xMain$.getSocketOrExit(Main.scala:145)
        at sbt.xMain$.bootServerSocket$lzycompute$1(Main.scala:77)
        at sbt.xMain$.bootServerSocket$1(Main.scala:77)
        at sbt.xMain$.withStreams$1(Main.scala:85)
        at sbt.xMain$.run(Main.scala:121)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at sbt.internal.XMainConfiguration.run(XMainConfiguration.java:57)
        at sbt.xMain.run(Main.scala:46)
        at xsbt.boot.Launch$.$anonfun$run$1(Launch.scala:149)
        at xsbt.boot.Launch$.withContextLoader(Launch.scala:176)
        at xsbt.boot.Launch$.run(Launch.scala:149)
        at xsbt.boot.Launch$.$anonfun$apply$1(Launch.scala:44)
        at xsbt.boot.Launch$.launch(Launch.scala:159)
        at xsbt.boot.Launch$.apply(Launch.scala:44)
        at xsbt.boot.Launch$.apply(Launch.scala:21)
        at xsbt.boot.Boot$.runImpl(Boot.scala:78)
[...]

Issue #2235, opened 2015-10-08
Pattern matching on string setting doesn't compile


greenrd:
```
Environment.value match {
      case "dev" => 
```

fails with:

```
build.sbt:81: error: type mismatch;
 found   : String("dev")
 required: sbt.Def.Setting[String]
      case "dev" => 
```

but

```
Environment.value.value match {
      case "dev" =>
```

fails with

```
build.sbt:80: error: value value is not a member of String
    Environment.value.value match {
```

eed3si9n:
I am guessing these could be worked around by assigning to a val first?

eed3si9n:
I am closing this due to inactivity, but if this is still relevant let us know.

Issue #3681, opened 2017-10-27
Directory standard


eed3si9n:
## steps

Run sbt.

## problem

We use `~/.sbt` etc.

## expectation

Adhere to directory standards?

- [File System Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html)
- [XDG Base Directory Specification](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html)

liff:
Would it be possible to use [directories](https://github.com/soc/directories) like [coursier does](https://github.com/coursier/coursier/pull/676/files)?

dwijnand:
Yes - with two concerns:
1. how do this change impact sbt, its launcher (sbt/launcher), its ecosystem of plugins, and plugin cross-building from the current usage of `~/.sbt`, and how do we provide a graceful migration?
2. what's the stability of `directories` binary API?

soc:
@dwijnand Let me know if I can provide any information that helps sorting this out.

dwijnand:
my first concern is more specific to sbt's current use of directories, and I haven't given it much thought.

the other one is partially directories-specific (I see the current version of directories is "10"), but it's actually generic to libraries on the jvm: selecting a library requires everything else to be compatible with that library's binary API. in sbt's case it means if we start using directories 10 then for the rest of sbt 1 all plugins will have to use directories 10 (or a version for which the API they depend on is compatible with directories 10).

soc:
@dwijnand Hi Dale!

**Regarding _directories_:** It is built on Java 6, doesn't have any dependencies and consists of 4 files (you can probably get rid of two of them). I only publish major versions, so that no person depending on it will ever get a different version from what they originally specified in their build config, without explicitly changing it himself/herself.

I would suggest shading (parts of) this library for SBT's purposes to avoid any kind of classpath or compatibility concerns. The API itself is likely to be very stable now, as I have already considered and either gone through or rejected every major redesign I (and other people) can think of. Nevertheless, you can pick a version and just stick with it.

**Regarding `.sbt`:**

- Move global config and plugins to `.config/sbt` (`XDG_CONFIG_HOME`).
- Move `.sbt/boot` to `.cache/sbt/boot` (`XDG_CACHE_HOME`).
- Compiled configuration should go into an appropriately named sub-directory in `.cache/sbt` (`XDG_CACHE_HOME`).
- Socket files and locks need to go to `XDG_RUNTIME_DIR`.
[...]

PR #6887, opened 2022-04-26
Use XDG_RUNTIME_DIR environment variable to choose the path used for the creation of sockets

AlonsoM45:
The `XDG_RUNTIME_DIR` environment variable will be used to choose the base directory in which the sockets will be created by **sbt**. This can help when issues such as #6777 appear. There are other related issues, such as #6101, [Metals issue #2235](https://github.com/scalameta/metals/issues/2235), [Che issue #18394](https://github.com/eclipse/che/issues/18394). Those are closed issues, but there are some cases in which the solution does not work (such as the case in #6777). Furthermore, the solution given seems more like a workaround than an actual fix.

**What causes this issue?**
At least in my case, the **ServerAlreadyBootingException** is thrown after **sbt** tries to create a Unix domain socket and fails. In my specific environment, I was not able to create the sockets in some directories because they were mounted on an NFS. This prevented me from using any automated sbt command on any of those directories (Metals uses sbt under the hood, and it was not able to start because of the ServerAlreadyBootingException), which in turn resulted in me not being able to use Metals on a project that was created on any of those directories. 

**How is the issue solved in this PR?**
If the `XDG_RUNTIME_DIR` environment variable is set, then sockets will be created in a folder with a hashed name, inside the directory `XDG_RUNTIME_DIR/.sbt/`. If this variable is not set, everything works exactly the same as always.

Let's see an example:

1. My environment variable `XDG_RUNTIME_DIR` is set to /home/alonso/.runtime-dir`
2. I create my project in `/home/alonso/workspace/hello-world`
3. I run `sbt compile`
4. The socket is created in `/home/alonso/.runtime-dir/.sbt/sbt-socket102030405060/sbt-load.sock`. The number **102030405060** is a hash that is computed from the base directory of the project (it is actually the same hash that is produced in order to create sockets in Windows) and it will be different depending on the location of the project
5. The sbt command executed correctly, which would not have happened if the socket had been created in the `home/alonso/workspace` directory or any of its subdirectories (in this example, `/home/BlackDiamond/workspace` corresponds to a network file share)

### DIFF
--- a/main-command/src/main/java/sbt/internal/BootServerSocket.java
+++ b/main-command/src/main/java/sbt/internal/BootServerSocket.java
@@ -281,13 +281,23 @@ public class BootServerSocket implements AutoCloseable {
         }
       };
 
+  private static String getUnixParentDirectory(String path) {
+    for (int i = path.length() - 2; i >= 0; i--) {
+      if (path.charAt(i) == '\\' || path.charAt(i) == '/') {
+        return path.substring(0, i + 1);
+      }
+    }
+    return "/";
+  }
+
   public BootServerSocket(final AppConfiguration configuration)
       throws ServerAlreadyBootingException, IOException {
     final Path base = configuration.baseDirectory().toPath().toRealPath();
-    final Path target = base.resolve("project").resolve("target");
     if (!isWindows) {
+      final String actualSocketLocation = socketLocation(base);
+      final Path target = Paths.get(getUnixParentDirectory(actualSocketLocation));
       if (!Files.isDirectory(target)) Files.createDirectories(target);
-      socketFile = Paths.get(socketLocation(base));
+      socketFile = Paths.get(actualSocketLocation);
     } else {
       socketFile = null;
     }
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @AlonsoMontero, @AlonsoM45,

eed3si9n:
Thanks for the contribution.

Rather than creating a system property for this specific use case, I wonder if we should use this as an opportunity to adhere to some existing standards like XDG Base Directory Specification (#3681). For instance, I think it recommends socket files to be created under `XDG_RUNTIME_DIR` somewhere. We can respect the environment variable, which you can set to whatever, and also move the default of that to something like `/tmp/.sbt/`.

What do you think?

AlonsoM45:
Hello, I changed the code in order to use the `XDG_RUNTIME_DIR` environment variable whenever we are not using Windows. The default of that variable will be `/tmp`, and then the sockets will be created in the directory `XDG_RUNTIME_DIR/.sbt`. How does it look to you?

main-command/src/main/java/sbt/internal/BootServerSocket.java
```diff
@@ -281,13 +281,23 @@ public OutputStream outputStream() {
...
       };
 
+  private static String getUnixParentDirectory(String path) {
+    for (int i = path.length() - 2; i >= 0; i--) {
+      if (path.charAt(i) == '\\' || path.charAt(i) == '/') {
+        return path.substring(0, i + 1);
+      }
+    }
+    return "/";
+  }
```
eed3si9n: Is this required as opposed to using NIO Path?

eed3si9n:
Yea. Mostly looks good I think. Could you update the description of the PR to reflect the update please?

main-command/src/main/java/sbt/internal/BootServerSocket.java
AlonsoM45: I have updated the code in order to use NIO Path.getParent()

AlonsoM45:
> Yea. Mostly looks good I think. Could you update the description of the PR to reflect the update please?

Done, I have just updated it.

AlonsoM45:
Thanks @eed3si9n for your review, your suggestions and for merging this PR. I had a question for you, if you don't mind: Do you have an estimated date for the next release?

eed3si9n:
@AlonsoM45 It's a bit concerning that the develop branch is no longer passing on CI though. I don't know if this is caused by merging this PR.

eed3si9n:
`develop` branch is slated for sbt 2.x, so for this to see the day of light sooner we need to backport this to 1.7.x and/or 1.6.x branch.

eed3si9n:
https://github.com/sbt/sbt/pull/6907 tested and merged.

AlonsoM45:
Hello. I saw that CI is passing now. I assume everything is OK now on that side. Thanks for backporting this to 1.7.x.

Jasper-M:
One possible issue with defaulting to `/tmp` here is that it's shared between all users, and if one user creates `/tmp/.sbt/sbt-socketXXXXXX` then another might not have appropriate permissions on `/tmp/.sbt` or `/tmp/.sbt/sbt-socketXXXXXX`. And that's without considering concurrent users.

bishabosha:
This change seems to negatively effect macOS where for some reason if `XDG_RUNTIME_DIR` does not exist then now sbt tries to make a socket from a temp directory, which has a path too long to make a socket from.

Edit: the problem in macOS was caused by #6935 which sets the default directory to be a temp directory - which is too long for a unix socket on macOS

PR #6894, opened 2022-05-09
Add support for wildcards in Scala version switch (fuzzy switch)

raboof:
This adds the ability to use semantic version selector expression (such as `2.13.x` and `3.x`) in `++` command as discussed in https://github.com/sbt/sbt/discussions/6893. The following description includes the patch in https://github.com/sbt/sbt/pull/6936.

## Usage

```
++ 2.13.x test
```

## Explanation

This is an extension of existing `++ <sv> <command1>` where the given `<command1>` will be executed only within the subprojects listing `<sv>` in their respective `crossScalaVersions` setting.

sbt 1.7.0 now allows the `<sv>` part to be a semantic version selector expression, implemented originally by @tanishiking in https://github.com/sbt/librarymanagement/pull/239 based on [npm-semver](https://www.npmjs.com/package/semver#Ranges). Note that the semantic version selector expression may match at most one Scala version within `crossScalaVersions` per subproject. This is because `++` remains to be a _switch_ command.

### DIFF
--- a/main/src/main/scala/sbt/Cross.scala
+++ b/main/src/main/scala/sbt/Cross.scala
@@ -8,7 +8,7 @@
 package sbt
 
 import java.io.File
-
+import java.util.regex.Pattern
 import sbt.Def.{ ScopedKey, Setting }
 import sbt.Keys._
 import sbt.SlashSyntax0._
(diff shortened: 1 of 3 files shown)


eed3si9n:
@raboof I am thinking about wrapping up 1.7.0 soon.

raboof:
> @raboof I am thinking about wrapping up 1.7.0 soon.

Thanks for the ping, added tests (and fixed behavior), now ready for review.

raboof:
(rebased on `origin/develop` - or should that be `origin/1.7.x` at this point?)

eed3si9n:
1.7.x branch would be better but either is ok.

eed3si9n (approved):
Thanks. Looks good to me.

SethTisue:
this is awesome. I love it!

a small suggestion. the output is:

```
sbt:root> ++3.x
[info] Setting Scala version to 3.x on 3 projects.
```

but it doesn't tell me which version it actually selected; I'd prefer not to have to `show scalaVersion` to see that information.

He-Pin:
@raboof Seems like it can't match both `3.1.3 and 3.2.0` reported at
https://github.com/sbt/sbt/issues/7018

PR #6902, opened 2022-05-23
"oldshell" command should use `OldShell` on failure instead of `Shell`

unkarjedy:
following up https://github.com/sbt/sbt/pull/3098

Otherwise it will return to `shell` (server) behavior:

image

(by "it's removed in 1.6.2" on the screenshot I meant that `sbt oldshell` command is wiped out from terminal once it's started)

### DIFF
--- a/main-command/src/main/scala/sbt/BasicCommands.scala
+++ b/main-command/src/main/scala/sbt/BasicCommands.scala
@@ -406,7 +406,7 @@ object BasicCommands {
       case Some(line) =>
         val newState = s
           .copy(
-            onFailure = Some(Exec(Shell, None)),
+            onFailure = Some(Exec(OldShell, None)),
             remainingCommands = Exec(line, s.source) +: Exec(OldShell, None) +: s.remainingCommands
           )
           .setInteractive(true)


unkarjedy:
If you are ok with this change I can backport it to 1.6 .. 1.2 (what is the minimum maintainable major version?)

SethTisue:
I can't answer your question in full, but my understanding is that 1.6.x is the oldest branch that could conceivably be targetable (without sponsorship, anyway). Users on older minor versions should move to 1.6.x if they want fixes.

Eugene would have to speak to whether further 1.6.x releases are likely, or merely conceivable.

eed3si9n (approved):
Thanks!

eed3si9n:
Thanks Seth.

> I can't answer your question in full, but my understanding is that 1.6.x is the oldest branch that could conceivably be targetable (without sponsorship, anyway). Users on older minor versions should move to 1.6.x if they want fixes.

+1

> Eugene would have to speak to whether further 1.6.x releases are likely, or merely conceivable.

If there are demand for more 1.6.x, I'd be happy to release more patch releases or maybe work with @adpi2 to rope Scala Center into it.

PR #3468, opened 2017-08-23
Disable scripted tests that fail in CI

eed3si9n:
This is continuation of https://github.com/sbt/sbt/pull/3456

PR #6903, opened 2022-05-24
Fix testQuick on changed function in object

### LINKED ISSUE #5504: TestQuick not re-running tests on changed function in object
## steps
1. Start new sbt project and create

#### project/build.properties
```scala
sbt.version = 1.3.9
```

#### build.sbt
```scala
name := "sbt-testquick-trigger"
scalaVersion := "2.13.1"
libraryDependencies += "org.scalatest" %% "scalatest" % "3.1.1" % "test"
```
#### src/main/scala/MathFunction.scala
```scala 
object MathFunction {
  def times2(i: Int): Int = i * 2
}
```
#### src/test/scala/MathFunctionTest.scala
```scala
class MathFunctionTest extends org.scalatest.funsuite.AnyFunSuite {
  test("one") { assert(MathFunction.times2(2) == 4) }
  test("two") { assert(MathFunction.times2(8) == 16) }
}
``` 
2. Start sbt with `sbt`
3. Run command `testQuick`
4. Change function in `MathFunction`-object to something else, eg:
```scala
  def times2(i: Int): Int = 2 * 2
``` 
5. Run command `testQuick` again

## problem
The second time the `testQuick` command is ran, the test are not actually re-ran even though the function is now bad and they would fail.

## expectation
The tests are re-ran since the function is changed and I would like to see that it is now indeed broken.

## notes
Running `test` instead of `testQuick` the second time does indeed fail as expected.

Going through the same steps with MathFunction as a `case class` or `class` does pick up the changes and run the tests correctly when the function is changed.

gontard:
Fixes https://github.com/sbt/sbt/issues/5504

I also have re-activated/updated/completed the existing scripted test (was deactivated in https://github.com/sbt/sbt/pull/3468)


eed3si9n (approved):
Thanks for the contribution!

PR #6926, opened 2022-06-14
Filter incompatible Scala 3 projects in cross switch commands

### LINKED ISSUE #6915: Version switch operator (++) doesn't work as expected on Scala 3
## steps

Consider the following sample project on sbt 1.6.2:

```scala
// build.sbt

scalaVersion := "2.12.16"

lazy val core = project
  .settings(
    crossScalaVersions := Seq("2.12.16", "3.0.2", "3.1.2")
  )

lazy val subproj = project
  .dependsOn(core)
  .settings(
    crossScalaVersions := Seq("2.12.16", "3.1.2"),
    // a random library compiled against Scala 3.1
    libraryDependencies += "org.http4s" %% "http4s-core" % "0.23.12"
  )
```

```scala
// subproj/src/main/scala/Main.scala

import org.http4s.Uri

object Main extends App {
  // Just using something from http4s 
  println(Uri.fromString("example.com"))
}
```

Try running `sbt "++3.0.2 compile"`.

## problem

The compilation fails as sbt tries to build `subproj`.

## expectation

I didn't expect `++3.0.2` to compile `subproj`, since it doesn't include a 3.0.x version in its cross-versions. In general, subprojects declaring Scala 3.1 as their Scala version can't be expected to compile successfully against Scala 3.0.x versions.

## notes

This is an important use case for module-based projects like [PureConfig](https://github.com/pureconfig/pureconfig). We want to compile PureConfig modules against the minimum possible Scala version to ensure maximum compatibility and we rely on `++` on CI and on our automated publishing process.

ruippeixotog:
Fixes #6915.

### DIFF
--- a/main/src/main/scala/sbt/Cross.scala
+++ b/main/src/main/scala/sbt/Cross.scala
@@ -19,7 +19,7 @@ import sbt.internal.util.AttributeKey
 import sbt.internal.util.complete.DefaultParsers._
 import sbt.internal.util.complete.{ DefaultParsers, Parser }
 import sbt.io.IO
-import sbt.librarymanagement.CrossVersion
+import sbt.librarymanagement.{ CrossVersion, ScalaArtifacts }
 
 /**
  * Cross implements the Scala cross building commands:
@@ -314,6 +314,12 @@ object Cross {
       excluded.foreach(logProject.tupled)
     }
 
+    def crossCompatibleVersion(full: String): String = {
+      if (ScalaArtifacts.isScala3(full))
+        CrossVersion.partialVersion(full).fold(full) { case (maj, min) => s"$maj.$min" }
+      else CrossVersion.binaryScalaVersion(full)
+    }
+
     val projects: Seq[(ResolvedReference, Seq[ScalaVersion])] = {
       val projectScalaVersions =
         structure.allProjectRefs.map(proj => proj -> crossVersions(extracted, proj))
@@ -323,11 +329,11 @@ object Cross {
           .map(BuildRef.apply)
           .map(proj => proj -> crossVersions(extracted, proj))
       } else {
-        val binaryVersion = CrossVersion.binaryScalaVersion(version)
+        val binaryVersion = crossCompatibleVersion(version)
 
         val (included, excluded) = projectScalaVersions.partition {
           case (_, scalaVersions) =>
-            scalaVersions.exists(v => CrossVersion.binaryScalaVersion(v) == binaryVersion)
+            scalaVersions.exists(v => crossCompatibleVersion(v) == binaryVersion)
         }
         if (included.isEmpty) {
           sys.error(
--- a/sbt-app/src/sbt-test/actions/cross-strict-aggregation-scala-3/build.sbt
+++ b/sbt-app/src/sbt-test/actions/cross-strict-aggregation-scala-3/build.sbt
@@ -1,0 +1,14 @@
+scalaVersion := "2.12.16"
+
+lazy val core = project
+  .settings(
+    crossScalaVersions := Seq("2.12.16", "3.0.2", "3.1.2")
+  )
+
+lazy val subproj = project
+  .dependsOn(core)
+  .settings(
+    crossScalaVersions := Seq("2.12.16", "3.1.2"),
+    // a random library compiled against Scala 3.1
+    libraryDependencies += "org.http4s" %% "http4s-core" % "0.23.12"
+  )
--- a/notes/1.7.0/cross-strict-aggregation-scala-3.md
+++ b/notes/1.7.0/cross-strict-aggregation-scala-3.md
@@ -1,0 +1,8 @@
+[@ruippeixotog]: https://github.com/ruippeixotog
+
+[#6915]: https://github.com/sbt/sbt/issues/6915
+[#6926]: https://github.com/sbt/sbt/pull/6926
+
+### Bug Fixes
+
+- Make `++ <scala-version> <command>` run `<command>` only on compatible Scala 3 subprojects. [#6915][]/[#6926][] by [@ruippeixotog][]
--- a/sbt-app/src/sbt-test/actions/cross-strict-aggregation-scala-3/subproj/src/main/scala/A.scala
+++ b/sbt-app/src/sbt-test/actions/cross-strict-aggregation-scala-3/subproj/src/main/scala/A.scala
@@ -1,0 +1,6 @@
+import org.http4s.Uri
+
+object A {
+  // Just using something from http4s 
+  Uri.fromString("example.com")
+}
--- a/sbt-app/src/sbt-test/actions/cross-strict-aggregation-scala-3/test
+++ b/sbt-app/src/sbt-test/actions/cross-strict-aggregation-scala-3/test
@@ -1,0 +1,14 @@
+> ++3.0.2 compile
+
+$ exists core/target/scala-3.0.2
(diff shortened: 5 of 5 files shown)


sbt-app/src/sbt-test/actions/cross-strict-aggregation-scala-3/build.sbt
```diff
@@ -0,0 +1,14 @@
...
+  .settings(
+    crossScalaVersions := Seq("2.12.16", "3.0.2", "3.1.2")
+  )
+
+lazy val subproj = project
+  .dependsOn(core)
+  .settings(
+    crossScalaVersions := Seq("2.12.16", "3.1.2"),
+    // a random library compiled against Scala 3.1
+    libraryDependencies += "org.http4s" %% "http4s-core" % "0.23.12"
```
ruippeixotog, line 13: I know depending on a 3rd party is probably not a good thing to do in SBT tests. Do you have any idea on how I can force compilation to fail if we use Scala 3.0 (in other words, ensure that this project only compiles with Scala 3.1)?

ruippeixotog:
Review by @eed3si9n.

sbt-app/src/sbt-test/actions/cross-strict-aggregation-scala-3/build.sbt
eed3si9n: Maybe you could define a custom task or macro within the project that's aware of the Scala version of the compiler?

main/src/main/scala/sbt/Cross.scala
```diff
@@ -323,11 +329,11 @@ object Cross {
           .map(BuildRef.apply)
           .map(proj => proj -> crossVersions(extracted, proj))
       } else {
-        val binaryVersion = CrossVersion.binaryScalaVersion(version)
+        val binaryVersion = crossCompatibleVersion(version)
```
eed3si9n, line 332: Thanks for the PR. Looking at this change, my first thought was that if there's some compatibility logic regarding Scala versions, we should centralize that in librarymanagement objects like `CrossVersion.*`. And then I realized that `CrossVersion.binaryScalaVersion` proxies `CrossVersionUtil.binaryScalaVersion`, which does in fact centralizes the Scala version compatibility logic.
https://github.com/sbt/librarymanagement/blob/v1.7.0-M1/core/src/main/scala/sbt/internal/librarymanagement/cross/CrossVersionUtil.scala#L61-L91

Excerpt related to Scala 3:

```scala
  private[sbt] def binaryScala3Version(full: String): String = full match {
    case ReleaseV(maj, _, _, _)                                                  => maj
    case NonReleaseV_n(maj, min, patch, _) if min.toLong > 0 || patch.toLong > 0 => maj
    case BinCompatV(maj, min, patch, stageOrNull, _) =>
      val stage = if (stageOrNull != null) stageOrNull else ""
      binaryScala3Version(s"$maj.$min.$patch$stage")
    case _ => full
  }

  def binaryScalaVersion(full: String): String = {
    if (ScalaArtifacts.isScala3(full)) binaryScala3Version(full)
    else
      binaryVersionWithApi(full, TransitionScalaVersion)(scalaApiVersion) // Scala 2 binary version
  }
```

Aside: I'm not a fan of the use of the term "binary" in this context because it literally can have 2 different meanings:
1. Of two things, as in first two segments `x.y` of `x.y.z`.
2. Of machine-code things, like "binary compatibility" of JAR files.
In this case `CrossVersionUtil.binaryScalaVersion` returns ABI version (application binary interface version), which for Scala 3 should be `"3"`.

In other words, the current implementation performs a kind of a fuzzy search on `++<v1> <command>` where `v1` is compared against `crossScalaVersions` using `binaryScalaVersion` (ABI version) as the test.

One thing that might be related is https://github.com/sbt/sbt/discussions/6893, where I (unknowingly) proposed a more formal fuzzy search `3.*` e.g. `++3.* test`

If we can get https://github.com/sbt/sbt/pull/6894 in for 1.7.0, then I wonder if we should remove the fuzziness out of non-`*` versions like `2.12.16` and `3.1.2`. Then I think the implementation would be more straightforward to code and explain.

ruippeixotog: @eed3si9n I usually don't think of `++` as some kind of "string-based" fuzzy search on Scala versions, and I think the operator becomes much less useful if it behaves like that.

I usually apply it in the context of having different subprojects supporting independent sets of versions, and when I write something like `++<v1> <command>`, I read it as "run \<command\> in all projects that can be compiled with \<v1\>". With the introduction of Scala 3 this doesn't seem to be a simple string search anymore. If I want to express things like the original situation of #6915, I could write `++3.1.* <command>`. However, once Scala 3.2 comes out I'd have to run both `++3.1.* <command>; ++3.2.* <command>` to achieve the behavior described above, and so on.

My preference is for the operator to encode implicitly the relationship of "project targeting Scala version X can run with Scala version Y". Scala 3 seems to have made that relation unidirectional - a project targeting Scala 3.0 can use Scala 3.1, but not the other way around. But this is still about binary compatibility, not about version regexes.

eed3si9n: > My preference is for the operator to encode implicitly the relationship of "project targeting Scala version X can run with Scala version Y".

Understood. If that's the case, should we should implement a match function that encodes that in https://github.com/sbt/librarymanagement/blob/v1.7.0-M1/core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala and document it. If we go by "X can run with Y" test, shouldn't that allow subproject with 3.0.0 to be included when the user types `++3.2.0 test`?

ruippeixotog: Sure, I can send a PR to librarymanagement with this new function 👍 It seems useful even outside this use case.

> If we go by "X can run with Y" test, shouldn't that allow subproject with 3.0.0 to be included when the user types ++3.2.0 test?

Yes, that's my expectation. That's a real use case for libraries that want to specify Scala 3.0.x as the target version for publishing (to maximize compatibility) but still want to ensure their code works against the latest Scala 3 version as per [this recommendation](https://www.scala-lang.org/blog/2021/10/21/scala-3.1.0-released.html).

eed3si9n: That may work for testing, but we'd still have issues regarding publishing though.

```
++3.0.0 publishSigned
++3.2.0 publishSigned
```

would now try to publish core subproject twice.

ruippeixotog: That's true, but running these two commands will result in problems even treating versions as strict, since the binary artifact for them is the same (ending with `_3` suffix). Library maintainers are never expected to publish the same artifact using two different Scala 3 versions.

That said, I see how it might be confusing for users. If you think these edge cases might be confusing for users, what do you think of adding a new modifier for backwards binary compatibility, like the bang operator in `++ <version>!`?

eed3si9n: Since picking binary compatible has been the behavior of `++`, changing to _backward_ binary compatible is still an improvement. I think we just have to document its limitations with regards to publishing etc now that we have a better picture of what's going on.

ruippeixotog: Sounds good! Opened https://github.com/sbt/librarymanagement/pull/400.

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/6939.

Issue #6896, opened 2022-05-13
Build errors get "stuck" in Metals  (SBT 1.7.0-M2 as BSP server)


asflierl:
continuing from my comments on #6847 @tanishiking 

## steps

1. clone https://github.com/asflierl/sbt-bsp-errors-minimal-repro.git
2. in the project root directory, start VSCode (e.g. `code .`) [assuming here that VSCode has the Metals plug-in version 0.11.5 installed]
3. after Metals starts, it will ask to import the project. select "not now"
4. perform the "Metals: switch build server" action and select "SBT" (this will generate `project/metals.sbt` etc. and start SBT in the background)
5. wait until everything downloads and compiles
6. introduce a small compilation error in `src/main/scala/Main.scala`, e.g. type "asd" as the last line
7. the compilation error shows up correctly in the problems view
8. undo the change you made in 6.

## problem

even after everything compiles fine again, the compilation error is still there in the problems view and will not disappear until another error is introduced somewhere else (which then in turn gets "stuck" in the problems view)

## expectation

the compile error should disappear after everything compiles fine again

## notes
I've also tried this with Scala 2.13.8 and the same thing happens.
When I switch back to SBT 1.6.2, everything works as expected.
In case it makes a difference, I ran all this with JDK 17.

tgodzik:
Looks like maybe sbt is not sending a successful compilation message to metals :thinking:

tanishiking:
@asflierl Thank you for reporting! I could reproduce the problem.
I'll investigate the root cause, but if I couldn't before the next release, it would be better to revert my change 👍

eed3si9n:
Fixed by @kpodsiad in https://github.com/sbt/sbt/pull/6929

asflierl:
Thank you so much @kpodsiad for fixing this and @tanishiking for implementing it in the first place! This is improving my day-to-day development experience a lot.

PR #6847, opened 2022-03-21
[BSP]Do not fire `build/publishDiagnostics` if there are (and were) no problems 

tanishiking:
As a first step for fixing https://github.com/sbt/sbt/issues/6844, this PR suppress the sbt server from sending `build/publishDiagnostics` if

- there were no problems with the previous build &&
- there are no problems with the current build

---

While `bloop` doesn't send notifications if problems are the same as the previous ones, I have a concern about comparing all the problems may slow down the server performance a bit (Is there a good way to profile build?)
Therefore, this PR suppresses the diagnostics only when there are (and were) no problems, for the time being. (I believe this is still pretty effective for reducing the message load between BSP client-server, assuming most of the programs have no problems).

### example
For example, when we edit a file in [this repository](https://github.com/tanishiking/sbt-multi-project/tree/20a3d476dc874db295fb3af3450873f8f211f1ea), (after a full compilation, edit doesn't contain any problems) sbt BSP sends the following requests/notifications

<details>

<summary>sbt 1.6.2 </summary>

```
[Trace - 09:46:34 PM] Sending request 'buildTarget/compile - (11)'
Params: {
  "targets": [
    {
      "uri": "file:/Users/tanishiking/src/github.com/tanishiking/sbt-multi-project/#proj2/Compile"
    }
  ]
}


[Trace - 09:46:35 PM] Received notification 'build/logMessage'
Params: {
  "type": 4,
  "message": "Processing buildTarget/compile"
}


[Trace - 09:46:35 PM] Received notification 'build/taskStart'
Params: {
  "taskId": {
    "id": "24",
    "parents": []
  },
  "eventTime": 1647866795052,
  "message": "Compiling proj1",
  "dataKind": "compile-task",
  "data": {
    "target": {
      "uri": "file:/Users/tanishiking/src/github.com/tanishiking/sbt-multi-project/#proj1/Compile"
    }
  }
}


[Trace - 09:46:35 PM] Received notification 'build/publishDiagnostics'
Params: {
  "textDocument": {
    "uri": "file:///Users/tanishiking/src/github.com/tanishiking/sbt-multi-project/proj1/src/main/scala/A.scala"
  },
  "buildTarget": {
    "uri": "file:/Users/tanishiking/src/github.com/tanishiking/sbt-multi-project/#proj1/Compile"
  },
  "diagnostics": [],
  "reset": true
}


[Trace - 09:46:35 PM] Received notification 'build/taskFinish'
Params: {
  "taskId": {
    "id": "24",
[...]

PR #6929, opened 2022-06-16
fix: publish BSP diagnostics when it's necessary

kpodsiad:
* do not publish bsp diagnostics if there were and there are no problems
* publish diagnostics if problems needs to be updated

continues effort of https://github.com/sbt/sbt/pull/6847, closes https://github.com/sbt/sbt/issues/6896

@adpi2 do you have an idea how to write test for this PR which will cover https://github.com/sbt/sbt/issues/6896?

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
@@ -12,6 +12,7 @@ import java.nio.file.Path
 import sbt.StandardMain
 import sbt.internal.bsp._
 import sbt.internal.util.ManagedLogger
+import sbt.internal.server.BuildServerProtocol.BspCompileState
 import xsbti.compile.CompileAnalysis
 import xsbti.{
   FileConverter,
@@ -40,8 +41,6 @@ sealed trait BuildServerReporter extends Reporter {
 
   def sendSuccessReport(
       analysis: CompileAnalysis,
-      prev: CompileAnalysis,
-      reportAllPreviousProblems: Boolean
   ): Unit
 
   def sendFailureReport(sources: Array[VirtualFile]): Unit
@@ -73,6 +72,7 @@ sealed trait BuildServerReporter extends Reporter {
 
 final class BuildServerReporterImpl(
     buildTarget: BuildTargetIdentifier,
+    bspCompileState: BspCompileState,
     converter: FileConverter,
     protected override val isMetaBuild: Boolean,
     protected override val logger: ManagedLogger,
@@ -90,22 +90,38 @@ final class BuildServerReporterImpl(
     if (ref.id().contains("<")) None
     else Some(converter.toPath(ref))
 
+  /**
+   * Send diagnostics from the compilation to the client.
+   * Do not send empty diagnostics if previous ones were also empty ones.
+   *
+   * @param analysis current compile analysis
+   */
   override def sendSuccessReport(
       analysis: CompileAnalysis,
-      prev: CompileAnalysis,
-      reportAllPreviousProblems: Boolean
   ): Unit = {
-    val prevInfos = prev.readSourceInfos().getAllSourceInfos().asScala
+    val shouldReportAllProblems = !bspCompileState.compiledAtLeastOnce.getAndSet(true)
     for {
       (source, infos) <- analysis.readSourceInfos.getAllSourceInfos.asScala
       filePath <- toSafePath(source)
     } {
-      val prevProblems = prevInfos.get(source).map(_.getReportedProblems()).getOrElse(Array.empty)
-      val dontPublish = prevProblems.length == 0 && infos.getReportedProblems().length == 0
-      val shouldPublish = reportAllPreviousProblems || !dontPublish
+      // clear problems for current file
+      val hadProblems = bspCompileState.hasAnyProblems.remove(filePath)
+
+      val reportedProblems = infos.getReportedProblems.toVector
+      val diagnostics = reportedProblems.flatMap(toDiagnostic)
+
+      // publish diagnostics if:
+      // 1. file had any problems previously - we might want to update them with new ones
+      // 2. file has fresh problems - we might want to update old ones
+      // 3. build project is compiled first time - shouldReportAllProblems is set
+      val shouldPublish = hadProblems || diagnostics.nonEmpty || shouldReportAllProblems
+
+      // file can have some warnings
+      if (diagnostics.nonEmpty) {
+        bspCompileState.hasAnyProblems.add(filePath)
+      }
 
       if (shouldPublish) {
-        val diagnostics = infos.getReportedProblems.toSeq.flatMap(toDiagnostic)
         val params = PublishDiagnosticsParams(
           textDocument = TextDocumentIdentifier(filePath.toUri),
           buildTarget,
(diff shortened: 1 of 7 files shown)


main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
```diff
@@ -188,6 +204,10 @@ final class BuildServerReporterImpl(
   }
 }
 
+object BuildServerReporterImpl {
+  private val hasAnyProblems = mutable.Set[Path]()
```
kpodsiad: For 1million of paths `hasAnyProblems` will have ~90MB. Checked using https://gist.github.com/kpodsiad/2d67e15d7c14d78b1711bc1e6cb36dcc

adpi2:
@kpodsiad You can add a test in https://github.com/sbt/sbt/blob/1.7.x/server-test/src/test/scala/testpkg/BuildServerTest.scala

If you need to change a file in the project, to have different diagnostics you can do something like that:
https://github.com/sbt/sbt/blob/f816aed05920912b5dff07f5d62012eabd6d7358/server-test/src/test/scala/testpkg/BuildServerTest.scala#L256-L263

main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
asflierl: Is this thread-safe, i.e. is it guaranteed that this global mutable data structure will be accessed only from 1 thread ever? If not, maybe `java.util.concurrent.ConcurrentHashMap.newKeySet[Path]` is worth considering as an alternative? Space requirements and performance are very close to the mutable Set.

kpodsiad: Thanks for this question @asflierl, I was wondering about this too but I forgot to ask about it. I think there can be only one compilation at time for the build target, but multiple build targets can be compiled in parallel.  
What if one file belongs to 2 build targets (it's very common to have multiple build targets for multiple scala versions) and there is problem in 2.13, but no problem in 2.12. If we're using just paths then we will end up with race conditions which build target will clear. 

I'll need to think of these concerns.

adpi2: Yes indeed, this is a important concern but it seems we don't have a problem with the current implementation.

There is one instance of `BuildServerReporterImpl` for each build target and, as you said @kpodsiad, there can be only one compilation at time for the build target. So it seems there cannot be any race condition.

kpodsiad: > There is one instance of `BuildServerReporterImpl` for each build target and, as you said @kpodsiad, there can be only one compilation at time for the build target. So it seems there cannot be any race condition.

@adpi2 
Are you sure on part `There is one instance of BuildServerReporterImpl for each build target`? Because when I was playing with it it seemed like each compilation gets fresh `BuildServerReporterImpl` instance. I've added `hasAnyProblems` to the class and I was printing all problems in `reportSucess` and `reportFailure`. And even though problems were set in `reportFailure`, `reportSucess` was always printing an empty set.   
My workflow was following:
- introduce error in file (problems were printed)
- fix error (I tried to print problems before clearing them but there weren't any)

That's why I moved `hasAnyProblems` to the companion object and in this case we might have race conditions because `hasAnyProblems` isn't scoped to one build target.

main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
```diff
@@ -116,14 +129,17 @@ final class BuildServerReporterImpl(
...
   override def sendFailureReport(sources: Array[VirtualFile]): Unit = {
     for {
       source <- sources
       filePath <- toSafePath(source)
     } {
-      val diagnostics = problemsByFile.getOrElse(filePath, Vector())
+      val diagnostics = problemsByFile.getOrElse(filePath, Vector.empty)
+
+      // mark file as file with problems
+      hasAnyProblems.add(filePath)
```
adpi2: Here I guess we should check if `diagnostics` is empty or not
```suggestion
      if (diagnostics.nonEmpty)
        hasAnyProblems.add(filePath)
```

main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
asflierl: Additionally, even if there was only one instance of the reporter accessing the global object (and therefore no race conditions), the issue of visibility remains: if the reporter is executed on a thread pool, it is not guaranteed that the same thread is used for every call to the reporter, so there is no clear "happensBefore" relation established between such calls, which might corrupt the internals of the mutable hash set.

adpi2: Oh yes you are right @kpodsiad, I overlooked this. Having a static mutable state is not the best though. Maybe we should add some sort of `bspCompileState: SettingKey[BspCompileState]` key to store the compile state of each build target.

adpi2: You don't even need to add a key, you can use an anonymous `Def.setting` instead, like that:

```scala
private def bspCompileState: Initialize[BspCompileState] = Def.setting {
  new BspCompileState()
}
```
where `BspCompileState` is a class that contains the mutable set of a single build target.

Then in the `bspReporter` task:

```scala
new BuildServerReporter(..., bspCompileState.value, ...)
```

`bspCompileState` is a setting so it will be the same instance on each run and it will be automatically scoped to a build target (one instance for each build target).

I also suggest to move the mutable `compiledTargetsAtLeastOnce` into the `BspCompileState` instance, as a boolean.

main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
```diff
@@ -90,22 +90,32 @@ final class BuildServerReporterImpl(
...
+
+      // publish diagnostics if:
+      // 1. file had any problems previously - we might want to update them with new ones
+      // 2. file has fresh problems - we might want to update old ones
+      // 3. build project is compiled first time - shouldReportAllProblems is set
+      val shouldPublish = hadProblems || reportedProblems.nonEmpty || shouldReportAllProblems
 
       if (shouldPublish) {
-        val diagnostics = infos.getReportedProblems.toSeq.flatMap(toDiagnostic)
+        val diagnostics = reportedProblems.flatMap(toDiagnostic)
```
adpi2: Here diagnostics can contain some warnings. We need to set `hasAnyProblems` if it is not empty otherwise the warnings would never disappear.
warning

```suggestion
        val diagnostics = reportedProblems.flatMap(toDiagnostic)
        if (diagnostics.nonEmpty) bspCompileState.hasAnyProblems.add(filePath)
```

main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
```diff
@@ -117,13 +127,18 @@ final class BuildServerReporterImpl(
...
       source <- sources
       filePath <- toSafePath(source)
     } {
-      val diagnostics = problemsByFile.getOrElse(filePath, Vector())
+      val diagnostics = problemsByFile.getOrElse(filePath, Vector.empty)
+
+      // mark file as file with problems
+      if (diagnostics.nonEmpty) {
+        bspCompileState.hasAnyProblems.add(filePath)
+      }
```
adpi2, line 150: What if the diagnostics are empty and `hasAnyProblems` is false for this `filePath`. Maybe we can not send any report this time.

```suggestion
    val hadProblems = bspCompileState.hasAnyProblems.remove(filePath)
    
    val shouldPublish = hadProblems || diagnostics.nonEmpty || shouldReportAllProblems
    
    if (shouldPublish) {
      if (diagnostics.nonEmpty) {
        bspCompileState.hasAnyProblems.add(filePath)
      }
      
      ...
```

adpi2 (commented):
There is still one bug with the warnings. Also I think there are some more reports that could be avoided.

kpodsiad:
@adpi2 I found interesting behavior (it's already present, tested on 1.6.2). Lets have files `A` and `B`. 
- `A` has a warning
- make error in `B`
- warning in `A` disappears
- fix error in `B`
- warning in A occurs again

Not sure if it's a bug or feature, however Bloop doesn't clear the warning.

main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
kpodsiad: Wanted to write a testcase for this, but I couldn't because the following flow is broken:
- project is compiling
- introduce error
- reload workspace (this seems to create a new bsp reporter which will have `compiledAtLeastOnce` set to false)
- compile project - because `compiledAtLeastOnce` is set to false all diagnostics will be returned

adpi2 (approved):
Thanks @kpodsiad for fixing this! I also appreciate the refactoring in the `BuildServerTest`.

You are right about the strange behavior of warnings that disappear and then re-appear again. This is a small bug that is not critical. We will probably be able to fix it by storing more information about previous compilation in the `BspCompileState` class. I suggest we do that in a separate PR.

PR #6936, opened 2022-06-26
More flexible Scala version switch

### LINKED ISSUE #6934: ++ fuzzy search should use SemanticSelector
# steps

```
++ 2.13.x test
```

# problem

Probably won't work

# expectation

it works.

# notes

https://github.com/sbt/sbt/pull/6894 implemented fuzzy search of `crossScalaVersions`.
I didn't think about this before, but lm has `SemanticSelector` which is able to evaluate expressions like `2.13.x` (https://github.com/sbt/librarymanagement/blob/develop/core/src/test/scala/sbt/librarymanagement/SemanticSelectorSpec.scala). We should probably use it.

raboof:
By using SemanticSelector instead of custom globbing.

Follow-up on https://github.com/sbt/sbt/pull/6894,
fixes https://github.com/sbt/sbt/issues/6934


eed3si9n (approved):
Thanks!

SethTisue:
this is awesome. I love it!

a small suggestion. the output is:

```
sbt:root> ++3.x
[info] Setting Scala version to 3.x on 3 projects.
```

but it doesn't tell me which version it actually selected; I'd love not to have to `show scalaVersion` to see that information.

PR #6939, opened 2022-06-26
[1.7.x] Filter incompatible Scala 3 projects in cross switch commands, take 2

### LINKED ISSUE #6915: Version switch operator (++) doesn't work as expected on Scala 3

eed3si9n:
This is a modified resend of https://github.com/sbt/sbt/pull/6926 by @ruippeixotog
Fixes #6915

Usage
------

```scala
ThisBuild / scalaVersion := "2.12.16"

lazy val core = project
  .settings(
    crossScalaVersions := Seq("2.12.16", "3.0.2", "3.1.2")
  )

lazy val sub2 = project
  .dependsOn(core)
  .settings(
    crossScalaVersions := Seq("2.12.16", "3.1.2"),
    // a random library compiled against Scala 3.1
    libraryDependencies += "org.http4s" %% "http4s-core" % "0.23.12"
  )
```

```
++3.0.2 test
```

Explanation
-----------
In sbt 1.7.0 `++3.0.2 test` tests `core` subprojects, but skips `sub2`.
Previously, `++ <sv> <command1>` used the ABI suffix (`_3`) to filter the `crossScalaVersion`, which was a fine strategy for Scala 2.x releases that maintained both backward and forward version compatibility. Since Scala 3 only maintains backward compatibility, we should use backward compatibility check instead of checking if the ABI suffix would be the same.

### DIFF
--- a/main/src/main/scala/sbt/Cross.scala
+++ b/main/src/main/scala/sbt/Cross.scala
@@ -339,16 +339,18 @@ object Cross {
                 )
             }
         }
-      } else {
-        val binaryVersion = CrossVersion.binaryScalaVersion(version)
+      } else
+        // This is the default implementation for ++ <sv> <command1>
+        // It checks that <sv> is backward compatible with one of the Scala versions listed
+        // in crossScalaVersions setting. Note this must account for the fact that in Scala 3.x
+        // 3.1.0 is not compatible with 3.0.0.
         projectScalaVersions.map {
           case (project, scalaVersions) =>
-            if (scalaVersions.exists(v => CrossVersion.binaryScalaVersion(v) == binaryVersion))
+            if (scalaVersions.exists(CrossVersion.isScalaBinaryCompatibleWith(version, _)))
               (project, Some(version), scalaVersions)
             else
               (project, None, scalaVersions)
         }
-      }
     }
 
     val included = projects.collect {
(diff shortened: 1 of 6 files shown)


ruippeixotog:
Thanks for fixing all the outstanding issues and merging this, @eed3si9n!

PR #6942, opened 2022-06-28
Show a more useful log message when doing a flexible version switch

raboof:
Suggested by @SethTisue in https://github.com/sbt/sbt/pull/6894#issuecomment-1168042209

Will show multiple lines when different versions are selected for different subprojects.

When no subprojects have a matching Scala version you will get a
'Switch failed' exception anyway, so in that case there is no
change in behavior.


eed3si9n (approved):
Thanks!

SethTisue:
awesome!

PR #6943, opened 2022-06-28
Show a more appropriate message when no Scala versions matches

raboof:
When switching scala versions with a wildcard, and no subproject
has a matching version to switch to, show a more appropriate error
message

### DIFF
--- a/main/src/main/scala/sbt/Cross.scala
+++ b/main/src/main/scala/sbt/Cross.scala
@@ -15,6 +15,7 @@ import sbt.internal.Act
 import sbt.internal.CommandStrings._
 import sbt.internal.inc.ScalaInstance
 import sbt.internal.util.AttributeKey
+import sbt.internal.util.MessageOnlyException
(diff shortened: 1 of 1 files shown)


main/src/main/scala/sbt/Cross.scala
```diff
@@ -361,10 +361,15 @@ object Cross {
     }
 
     if (included.isEmpty) {
-      sys.error(
-        s"""Switch failed: no subprojects list "$version" (or compatible version) in crossScalaVersions setting.
-           |If you want to force it regardless, call ++ $version!""".stripMargin
-      )
+      if (isSelector(version))
+        sys.error(
```
eed3si9n: The change itself looks good. Since this is a somewhat predictable validation error, so could we throw `sbt.internal.util.MessageOnlyException` here and below please? That will show the error without a stacktrace.

eed3si9n (approved):
Thanks!

PR #6946, opened 2022-07-02
Remove compatibility check from ++

eed3si9n:
Problem
-------
Since sbt-doge merger `++ <sv> <command1>` has used binary compatibility
as a test to select subproject, but it causes surprising situations like
sbt/sbt#6915, and it blurs the responsibility of YAML file and build
file as the version specified in the version can override the Scala
version test on local laptop.

Solution
--------
This removes the compatibiliy check (backward-only or otherwise),
and require that `<sv>` match one of `crossScalaVersions` using the new
Semantic Version selector pattern.

### DIFF
--- a/main/src/main/scala/sbt/Cross.scala
+++ b/main/src/main/scala/sbt/Cross.scala
@@ -19,7 +19,7 @@ import sbt.internal.util.MessageOnlyException
 import sbt.internal.util.complete.DefaultParsers._
 import sbt.internal.util.complete.{ DefaultParsers, Parser }
 import sbt.io.IO
-import sbt.librarymanagement.{ CrossVersion, SemanticSelector, VersionNumber }
+import sbt.librarymanagement.{ SemanticSelector, VersionNumber }
 
 /**
  * Cross implements the Scala cross building commands:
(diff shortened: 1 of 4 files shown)


kitbellew:
@eed3si9n I wonder if this is the change that's making it impossible for `scalameta` to upgrade sbt to a recent version (it still uses 1.6.2). Even if not, perhaps you could suggest a solution.

`scalameta` has essentially two groups of projects:
- the main ones use `CrossVersion.binary` and a latest version of scala211, scala212 and scala213 under `crossScalaVersions` (currently, 2.11.12, 2.12.18 and 2.13.11)
- the semanticDb projects use `CrossVersion.full` and a bunch of various versions for each of them, but depend on some of the main ones.

For instance, the non-latest `++2.13.1` builds both semanticDb projects and their dependents main ones (even though that version is not mentioned in the latter group's `crossScalaVersions`). Which version is built for main ones is unclear, however.

However, when we attempt to upgrade to 1.9.1, CI shows an attempt to download scalameta-2.13.1 (and a local run says something about "Modules were resolved with conflicting cross-version suffixes").

Is there a work-around which doesn't involve splitting `binary` and `full` projects into separate builds and treating `binary` projects as external published dependencies? Is it possible to use something to the effect of `<project>.dependsOn(another % s"${scalaBinaryVersion.value}.x)"` (to force fuzzy)?

Ideally, this cross-cooperation between `full` and `binary` projects would work like this:
- `binary` projects build only their specified versions
- `full` projects find a compatible version of the dependent `binary` project (just like it would happen with an external dependency)
- each `binary` version is built only once when `+publish` is run (our Release jobs are ever-growing in duration -- and frequently failing, -- because the same `binary` version is built for every `full` version; see https://github.com/scalameta/scalameta/actions/workflows/release.yml)
- each `full` version possibly uses their own target directory (`target/scala-2.13.1` rather than `target/scala-2.13`)

Also, is there a better forum to request this help? cc @tgodzik

eed3si9n:
We use GitHub issue mostly to track bugs and tasks on sbt development. If you need help, probably the best avenue now is sbt channel on Scala Discord server, Stackoverflow, or a GitHub Discussion thread.

kitbellew:
thank you, @eed3si9n, submitted issue #7327 and separately asked for suggestions on discord.

PR #6948, opened 2022-07-03
enable `project/source-plugins` test

xuwei-k:
- update to latest ref
- add explicit `RootProject` call
- remove duplicate `*.sbt` file
- use `https://` instead of `git://` https://github.blog/2021-09-01-improving-git-protocol-security-github/


eed3si9n (approved):
Thanks!

PR #6962, opened 2022-07-08
remove sbt-dotty in tests

### LINKED ISSUE #6954: sbt-dotty is gone, but sbt still reference it?
https://github.com/sbt/sbt/search?q=sbt-dotty

sbt-dotty is official dropped. not maitained anymore, 
sbt still reference it?

xuwei-k:
- fix https://github.com/sbt/sbt/issues/6954


eed3si9n (approved):
Thanks!

Issue #6964, opened 2022-07-10
sbt.internal.util.Terminal.proxyInputStream does not override available


steinybot:
## steps

1. Add [site-sbt](https://github.com/sbt/sbt-site) to a project.
2. Run `previewAuto`.
3. Press any key.

Reproduction here: https://github.com/steinybot/bug-reports/tree/sbt/terminal-is-available

## problem

It does not stop.

## expectation

It should stop.

## notes

https://github.com/sbt/sbt-site/issues/168

`unfiltered.util.RunnableServer` has a busy loop polling for `System.in.available()` to be > 0 but it never is. This is because `sbt.internal.util.Terminal.proxyInputStream` does not override `available` but it should.

eed3si9n:
Thanks for the report!
Here's my PR for this - https://github.com/sbt/sbt/pull/6965

PR #6967, opened 2022-07-10
Calculate available bytes for terminal input streams

steinybot:
Adds some tests and another small fix for #6964

### DIFF
--- a/internal/util-logging/src/test/scala/sbt/internal/util/TerminalSpec.scala
+++ b/internal/util-logging/src/test/scala/sbt/internal/util/TerminalSpec.scala
@@ -1,0 +1,101 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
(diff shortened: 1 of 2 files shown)


internal/util-logging/src/test/scala/sbt/internal/util/TerminalSpec.scala
```diff
@@ -0,0 +1,94 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
```
steinybot, line 4: I copy/pasted the header. Should this be copyright to Mark Harrah when it was created after 2010? Other files are like this too.

eed3si9n: Yea. We use sbt-header to use the same header in all files.

Issue #6976, opened 2022-07-15
dependencyBrowseGraph fails with `key not found: GraphModuleId` when pagerduty-client is present


frosforever:
### Project files:
`build.sbt`:
```sbt
libraryDependencies ++= Seq(
    "com.github.dikhan" % "pagerduty-client" % "3.1.2"
    )
```
`project/plugins.sbt`:
```
addDependencyTreePlugin
```
`project/build.properties`:
```
sbt.version=1.7.1
```

## problem

Calling `dependencyBrowseGraph` on the project fails with `NoSuchElementException: key not found: GraphModuleId(org.apache.commons,commons-io,1.3.2)`

### Sbt commands
```shell
$ sbt dependencyBrowseGraph
[info] welcome to sbt 1.7.1 (Azul Systems, Inc. Java 1.8.0_332)
[info] loading settings for project sbt-dependency-graph-bug-build from plugins.sbt ...
[info] loading project definition from sbt-dependency-graph-bug/project
[info] loading settings for project sbt-dependency-graph-bug from build.sbt ...
[info] set current project to sbt-dependency-graph-bug (in build file:/sbt-dependency-graph-bug/)
[error] java.util.NoSuchElementException: key not found: GraphModuleId(org.apache.commons,commons-io,1.3.2)
[error] 	at scala.collection.MapLike.default(MapLike.scala:236)
[error] 	at scala.collection.MapLike.default$(MapLike.scala:235)
[error] 	at scala.collection.AbstractMap.default(Map.scala:65)
[error] 	at scala.collection.MapLike.apply(MapLike.scala:144)
[error] 	at scala.collection.MapLike.apply$(MapLike.scala:143)
[error] 	at scala.collection.AbstractMap.apply(Map.scala:65)
[error] 	at sbt.internal.graph.ModuleGraph.module(model.scala:70)
[error] 	at sbt.internal.graph.rendering.DOT$.originWasEvicted$1(DOT.scala:34)
[error] 	at sbt.internal.graph.rendering.DOT$.$anonfun$dotGraph$6(DOT.scala:56)
[error] 	at sbt.internal.graph.rendering.DOT$.$anonfun$dotGraph$6$adapted(DOT.scala:56)
[error] 	at scala.collection.TraversableLike.$anonfun$filterImpl$1(TraversableLike.scala:304)
[error] 	at scala.collection.Iterator.foreach(Iterator.scala:943)
[...]

PR #6978, opened 2022-07-19
Grab modules from graph using `get` instead of `apply` to account for missing edges

frosforever:
Connects to #6976 

It's possible for there to be orphaned edges in the constructed graph. Using `map.apply` to retrieve the edge will then fail  with `NoSuchElementException`. This PR updates to using `get` that safely returns an `Option` instead of throwing. 

This does _not_ fix the underlying issue. The produced dependency graph will likely be incorrect if the edges aren't present. This will only prevent an exception from being thrown. 

Thanks very much in advance to the maintainers for reviewing this PR and for maintaining SBT!


eed3si9n (approved):
Thanks, @frosforever!

lightbend-cla-validator:
Hi @frosforever,

PR #6983, opened 2022-07-26
fix: change BSP JVM environment working directory to be a path

### LINKED ISSUE #6982: IntelliJ fails to run main classes or tests in sbt 1.7.1
## steps

1. Open a sbt 1.7.1 project in IntelliJ
2. Click ▶️ on a main class or test

Reproduction: https://github.com/steinybot/bug-reports/tree/sbt/bsp-jvm-env-working-dir

## problem

It fails with:

> Error running 'Main':
> Cannot start process, the working directory 'file:/Users/bob/src/bug-reports' does not exist

The issue is that in `sbt.internal.server.BuildServerProtocol#jvmEnvironmentItem` it uses `Keys.baseDirectory.value.toURI().toString()` as the `workingDirectory` which is a URI but IntelliJ is expecting a file path.

## expectation

The BSP spec is unclear on exactly what this should be but IntelliJ and Bloop (I think) both treat it as a file path.

The generators in the [original BSP PR](https://github.com/build-server-protocol/build-server-protocol/pull/107/files#diff-5e73e580e67beb4059201b971d7c416a145d81b3fcb37aa3ee83fb1787c04d27) used a file URI but the mock server had a file path so even that is a little confused.

Interestingly `protocol/src/main/contraband/bsp.contra` specifies it as a `String` not a `URI`.

## notes

See https://youtrack.jetbrains.com/issue/SCL-20275/Cannot-start-process-the-working-directory-filexxxxyyyy-does-not-exist-when-running-the-main-method-in-an-object-sbt-171

Workaround:

```
import sbt.Keys._
import sbt._

object BspWorkaround extends AutoPlugin {

  override def projectSettings: Seq[Setting[_]] =
    inConfig(Compile)(bspSettings) ++
      inConfig(Test)(bspSettings)

  def bspSettings: Seq[Setting[_]] = Seq(
    bspBuildTargetJvmEnvironmentItem := bspBuildTargetJvmEnvironmentItem.value
      .withWorkingDirectory(baseDirectory.value.getAbsolutePath)
  )
}
```

steinybot:
Fixes #6982

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
@@ -661,7 +661,7 @@ object BuildServerProtocol {
     val target = Keys.bspTargetIdentifier.value
     val classpath = Keys.fullClasspath.value.map(_.data.toURI).toVector
     val jvmOptions = Keys.javaOptions.value.toVector
-    val baseDir = Keys.baseDirectory.value.toURI().toString()
+    val baseDir = Keys.baseDirectory.value.getAbsolutePath
     val env = envVars.value
 
     JvmEnvironmentItem(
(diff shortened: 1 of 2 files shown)


kpodsiad:
As the author of https://github.com/sbt/sbt/pull/6858 I'm sorry for breaking stuff in Intellij. I haven't got enough time time to pursue this work further and discover this inconsistency between sbt and bloop. 

Bloop is using plain and sbt should use it too. Thanks @steinybot 

@adpi2 can you take a look at this one? It's important for Intellij users.

eed3si9n (approved):
Thanks!

PR #6985, opened 2022-07-28
Add support for BSP's buildTarget/outputPaths method

povder:
This PR adds support for BSP's `buildTarget/outputPaths` method. That method was introduced in https://github.com/build-server-protocol/build-server-protocol/pull/269. Only a single "output path" is returned by sbt: the value of the `target` setting converted to URI. 

If accepted, I think this PR will be released in 1.8.0 so I added notes into notes/1.8.0 directory.

I updated tests and also invoked `buildTarget/outputPaths` manually on a real project.


adpi2 (approved):
Looks good to me!

@eed3si9n should we create a `1.8.x` branch?

eed3si9n:
> should we create a 1.8.x branch?

Yeah I guess so. I am also open to backporting this to 1.7.x if we are super sure that this won't break existing builds, plugins, Metals/IntelliJ integrations.

PR #6990, opened 2022-08-03
[1.8.x] Add support for BSP's buildTarget/outputPaths method

adpi2:
Backport of #6985 by @povder

This PR adds support for BSP's `buildTarget/outputPaths` method. That method was introduced in https://github.com/build-server-protocol/build-server-protocol/pull/269. Only a single "output path" is returned by sbt: the value of the `target` setting converted to URI. 

If accepted, I think this PR will be released in 1.8.0 so I added notes into notes/1.8.0 directory.

I updated tests and also invoked `buildTarget/outputPaths` manually on a real project.


adpi2:
@eed3si9n I created branch 1.8.x. Could you make it the default branch?

PR #6992, opened 2022-08-04
Fix background job logging

### LINKED ISSUE #6842: `bgRun` does not forward logs to the client
## steps

1. Create a minimal sbt build with a program that prints some output

```scala
// build.sbt
scalaVersion := "2.13.7"
```
```scala
// src/main/scala/example
package example

object HelloWorld extends App {
  for (_ <- 1 to 10) {
    println(s"Hello world!")
    Thread.sleep(1000)
  }
}
```
2. Start sbtn shell: `$ sbtn`
3. Run `bgRun`

The program starts in the background of sbt server.

## problem

The logs are not forwarded to the client that run the `bgRun` task.

```
sbt:example> bgRun
[info] running HelloWorld 
Hello world!
[success] Total time: 0 s, completed Mar 16, 2022 10:26:30 AM
sbt:sbt-client-input> 
```

## expectation

The logs should continue being forwarded to the client:
bgRun

adpi2:
Fixes #6842 

## Description

In order of commits

### Use BackgroundJobService context instead of MainLoop context

Use the `LoggerContext` of the `BackgroundJobService` to create and clean the loggers of the jobs. This is to avoid using the `LoggerContext` of the `MainLoop` which is closed after each command: the logger must stay alive after the execution of the command until the job is done.

### Don't use ProxyTerminal in background job

Do not use the `ProxyTerminal` in the `ConsoleAppender` of a job, so that the logs do not switch from active client to active client. Instead we get the current active terminal at the time the logger is created, and we stick to it until the job is done. So if the job is started by client network 1, the logs will be redirected to client network 1 until the end of the job.

### Catch `ClosedChannelException` in background job logger

If the terminal is closed before the end of the job we catch the `ClosedChannelException` thrown by the `ConsoleOut`, otherwise the job would crash.

## Results

The logs of a background job are only redirected to the clients that started the job.

bg-run

 If a terminal is killed the job keeps running in the background and the logs are not displayed anywhere.

bg-run-exit

### Getting the log of background job whose terminal is closed

Currently the logs of a job whose terminal is closed are lost. It would be nice to get them back when running `bgWait` on that job.

### Call to `println` in a background job

A call to `println` in a background job will always print to the terminal that is active, the ones that is running a task. In particular, if we run the main class of a project without forking, its standard output will move from one active terminal to the other active terminal.

bg-run-println

Unfortunately I don't think there is a solution for this.

###  `RelayAppender`

Currently the logger of a background job contains a relay appender and all BSP clients receive all logs of all background jobs.

This kind-of pollutes the logs of BSP clients. For instance in Metals:
bgRun-metals

It would be better if a BSP client only receives the logs of the background jobs that it started.

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/ConsoleAppender.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/ConsoleAppender.scala
@@ -126,11 +126,12 @@ object ConsoleAppender {
     def out: ConsoleOut
   }
   private[sbt] object Properties {
-    def from(terminal: Terminal): Properties = new Properties {
-      override def isAnsiSupported: Boolean = terminal.isAnsiSupported
-      override def isColorEnabled: Boolean = terminal.isColorEnabled
-      override def out = ConsoleOut.terminalOut(terminal)
-    }
+    def from(terminal: Terminal): Properties =
+      from(ConsoleOut.terminalOut(terminal), terminal.isAnsiSupported, terminal.isColorEnabled)
+
+    def safelyFrom(terminal: Terminal): Properties =
+      from(ConsoleOut.safeTerminalOut(terminal), terminal.isAnsiSupported, terminal.isColorEnabled)
+
     def from(o: ConsoleOut, ansi: Boolean, color: Boolean): Properties = new Properties {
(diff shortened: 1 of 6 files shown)


eed3si9n (approved):
Thanks!

PR #6994, opened 2022-08-05
add output when boot server socket fails to create

devin-fisher:
Small change to add some output when Boot Server Socket fails to be created in a non-interactive environment. I and a couple team members spent more than a day trying to figure out why sbt was exiting without any output in our CI/CD pipelines. Some output here would have help us isolate the issue much quicker.

This change seem too small to warrant a note but if that is not the case I can and will create one. Let me know. 

If there are use-cases where sbt is expected to exit and this output would be clutter, I can understand why this output might be undesirable. If that is the case, maybe someone can point me to a better way. Also if the text is not normative to the project, guide me and I'll update it too.

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -141,13 +141,17 @@ private[sbt] object xMain {
 
   private def getSocketOrExit(
       configuration: xsbti.AppConfiguration
-  ): (Option[BootServerSocket], Option[Exit]) =
-    try (Some(new BootServerSocket(configuration)) -> None)
+  ): (Option[BootServerSocket], Option[Exit]) = {
+    def printThrowable(e: Throwable): Unit = {
+      println("sbt thinks that server is already booting because of this exception:")
+      e.printStackTrace()
+    }
+
+    try Some(new BootServerSocket(configuration)) -> None
     catch {
       case e: ServerAlreadyBootingException
           if System.console != null && !ITerminal.startedByRemoteClient =>
-        println("sbt thinks that server is already booting because of this exception:")
-        e.printStackTrace()
+        printThrowable(e)
         println("Create a new server? y/n (default y)")
         val exit =
           if (ITerminal.get.withRawInput(System.in.read) == 'n'.toInt) Some(Exit(1))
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:
Hi @devin-fisher,

main/src/main/scala/sbt/Main.scala
```diff
@@ -153,9 +153,13 @@ private[sbt] object xMain {
           if (ITerminal.get.withRawInput(System.in.read) == 'n'.toInt) Some(Exit(1))
           else None
         (None, exit)
-      case _: ServerAlreadyBootingException =>
+      case e: ServerAlreadyBootingException =>
         if (SysProp.forceServerStart) (None, None)
-        else (None, Some(Exit(2)))
+        else {
+          println("Boot server failed to create socket")
+          e.printStackTrace()
```
eed3si9n: Could we refactor line 149~150 out to a in-function function so we can use the same error message please?

devin-fisher: I'd be happy to but I'm not quote sure what you are wanting. So a couple of questions:
1. What message do you want to re-use?
2. What scope do you want the function defined in?

If you are think of reusing the above error message, this case is a little different in that we are not going to take input from the user. But if that what you mean, just the first two bit of print outs, then that fine and I can do that. 

Also if you have another place that models the pattern, just point me to it and I can adapt it to this case.

eed3si9n: > But if that what you mean, just the first two bit of print outs, then that fine and I can do that.

Yea. That's what I was thinking. People search for error messages on GitHub issues and Stackoverflow etc so I want to make sure we have consistent error message for the same class of error.

So:

> 1. What message do you want to re-use?
```scala
        println("sbt thinks that server is already booting because of this exception:")
        e.printStackTrace()
```

and 

> 2. What scope do you want the function defined in?

I think the narrowest scope would be a nested method.

devin-fisher: Done, let me know if it match what you were looking for.

eed3si9n (approved):
Thanks!

main/src/main/scala/sbt/Main.scala
eed3si9n: Yes. The change LGTM.

PR #7006, opened 2022-08-21
fix: also include diagnosticCode and related info in InterfaceUtil

ckipp01:
Looks like I missed this in https://github.com/sbt/sbt/pull/6874 and I
hit on it in Mill when I couldn't figure out why it was also empty, and
thanks to @adpi realized it was because of the `LoggedReporter` in zinc
not taking it into account. However before I can bump that this needs to
be bumped as well.

refs: https://github.com/lampepfl/dotty/issues/14904

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/util/InterfaceUtil.scala
+++ b/internal/util-logging/src/main/scala/sbt/util/InterfaceUtil.scala
@@ -7,10 +7,14 @@
 
 package sbt.util
 
-import xsbti.{ Position, Problem, Severity, T2 }
 import java.io.File
 import java.util.Optional
 import java.util.function.Supplier
+import java.{ util => ju }
+
+import xsbti.{ DiagnosticCode, DiagnosticRelatedInformation, Position, Problem, Severity, T2 }
+
+import scala.collection.mutable.ListBuffer
 
 object InterfaceUtil {
   def toSupplier[A](a: => A): Supplier[A] = new Supplier[A] {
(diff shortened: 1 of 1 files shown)


internal/util-logging/src/main/scala/sbt/util/InterfaceUtil.scala
```diff
@@ -43,6 +47,18 @@ object InterfaceUtil {
       case None    => Optional.empty[A]()
     }
 
+  def l2jl[A](l: List[A]): ju.List[A] = {
```
ckipp01, line 50: I just followed how you did the jo2o up above because I'm assuming you don't want collection converters brought in?

eed3si9n: Sounds good. Not bringjng in dependency is a good default stance for bottom layers like this.

internal/util-logging/src/main/scala/sbt/util/InterfaceUtil.scala
```diff
@@ -104,14 +120,26 @@ object InterfaceUtil {
...
+  @deprecated("Use the overload of this method with more arguments", "1.7.2")
   def problem(
       cat: String,
       pos: Position,
       msg: String,
       sev: Severity,
       rendered: Option[String]
   ): Problem =
-    new ConcreteProblem(cat, pos, msg, sev, rendered)
+    problem(cat, pos, msg, sev, None, None, List.empty[DiagnosticRelatedInformation])
```
eed3si9n: Is this throwing out rendered?

eed3si9n (changes requested):
Marking this request changes since I have a question.

internal/util-logging/src/main/scala/sbt/util/InterfaceUtil.scala
ckipp01: O weird, yes you're correct. Good catch, I didn't mean to do this. I made this `rendered` is still captured now.

eed3si9n (approved):
Thanks!

PR #7010, opened 2022-08-26
Add sbt.argsfile option

### LINKED ISSUE #2974: Limit on classpath length when forking is enabled
If the classpath is particularly long, and forking is enabled, sbt will try to spawn a new jvm instance using `java.lang.UNIXProcess.forkAndExec()`, which calls `Process()`, which calls the native `execve()`. The latter has a limit on the length of each of its arguments, hardcoded in the Linux kernel to 128KB: https://github.com/torvalds/linux/blob/607ca46/include/uapi/linux/binfmts.h
The limits are also described in the section "Limits on size of arguments and environment", in the `execve()` man page.

That means that the limit on the number and length of the classpath elements is not particularly high: a thousand transitive dependencies may already be too much, in some cases. However, when forking is disabled, a very long classpath is not a problem (I tried 10'000 libraries, for a total of 1'280'000 characters, and it worked fine).

The issue should be addressed in this method: https://github.com/sbt/sbt/blob/adb4161/run/src/main/scala/sbt/Fork.scala#L94-L95, where the classpath should not be passed as-is to `ProcessBuilder`. Apparently, a technique that can be used to work around this issue is that of "pathing jars", described here: http://stackoverflow.com/a/201969

An example of a similar approach in Scala is here: https://github.com/mpokorny/scampi/blob/master/project/Build.scala#L18 (Mozilla Public License, v. 2.0)

easel:
There are a number of scenarios, including long classpaths, that
can cause linux and other systems to fail to fork a jvm. We can work
around this on JDK9 and above by using an "arguments file" [1].

This patch adds an "sbt.argsfile" property and SBT_ARGSFILE environment
variable. This option defaults to true.  When active on
Java 9+, long command lines are translated to an arguments file by
adding all command line arguments to the arguments file format at [1]
and java is invoked using the arguments file instead of passing the
arguments on the command line.

1. https://docs.oracle.com/javase/9/tools/java.htm#

Fixes: https://github.com/sbt/sbt/issues/2974

### DIFF
--- a/run/src/main/scala/sbt/Fork.scala
+++ b/run/src/main/scala/sbt/Fork.scala
@@ -8,6 +8,7 @@
 package sbt
 
 import java.io.File
+import java.io.PrintWriter
 import java.lang.ProcessBuilder.Redirect
 import scala.sys.process.Process
 import OutputStrategy._
@@ -15,6 +16,7 @@ import sbt.internal.util.{ RunningProcesses, Util }
 import Util.{ AnyOps, none }
 
 import java.lang.{ ProcessBuilder => JProcessBuilder }
+import java.util.Locale
 
 /**
  * Represents a command that can be forked.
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
LGTM

easel:
Thanks for the merge!

PR #7021, opened 2022-09-17
[1.8.x] Scala 2.12.17

### LINKED ISSUE #6997: Consider bumping `scala-xml` to 2.x
## Problem outline

Not really a bug, just something a bunch of people are going to hit on. I've reported this to [`scala/bug`](https://github.com/scala/bug/issues/12632), but these updates may have to go hand in hand if we do them. I'll copy over what I posted there since the reasoning is the same:

> I know this was a conversation a long time ago in https://github.com/scala/scala/pull/9743 but recently this conversation has surfaced again. Recently in `sbt-scoverage` I had some [reports](https://github.com/scoverage/sbt-scoverage/issues/439) of conflicts since I was still on 1.x, but some other libraries like [scalatest](https://github.com/scalatest/scalatest/pull/2114) bumped to 2.x making the problem hit people a bit harder. Due to that, I went ahead and updated to 2.x. Due to me updating, we got even more reports of conflicts with [sbt-native-packager](https://github.com/sbt/sbt-native-packager/issues/1513), which caused them to update to 2.x as well. Now we're in a position where people are getting hit with stuff like this:
> 
> ```
> [error]         * org.scala-lang.modules:scala-xml_2.12:2.1.0 (early-semver) is selected over {1.3.0, 1.0.6, 1.2.0}
> [error]             +- org.scoverage:scalac-scoverage-reporter_2.12:2.0.2 (depends on 2.1.0)
> [error]             +- com.github.sbt:sbt-native-packager:1.9.10 (sbtVersion=1.0, scalaVersion=2.12) (depends on 2.1.0)
> [error]             +- org.scala-sbt:testing_2.12:1.7.1                   (depends on 1.3.0)
> [error]             +- org.scala-sbt:sbinary_2.12:0.5.1                   (depends on 1.0.6)
> [error]             +- org.scala-sbt:main_2.12:1.7.1                      (depends on 1.3.0)
> [error]             +- org.scala-sbt:librarymanagement-core_2.12:1.7.0    (depends on 1.2.0)
> [error]             +- org.scala-lang:scala-compiler:2.12.16              (depends on 1.0.6)
> [error]             +- io.get-coursier:lm-coursier-shaded_2.12:2.0.10     (depends on 1.3.0)
> ```
> 
> Where common sbt plugins are conflicting with sbt and Scala 2.12 itself. Looking back at past conversations I understand the breaking changes between 1.x and 2.x of scala-xml are _extremely_ small so doing the following is pretty safe:
> 
> _in your plugins.sbt_
> ```
> ThisBuild / libraryDependencySchemes ++= Seq(
>   "org.scala-lang.modules" %% "scala-xml" % VersionScheme.Always
> )
> ```
> But this isn't ideal to have this all over the place, and will cause confusions for users. Is it time to update scala-xml in 2.12? Is there something holding this back? We're in a weird position now where have the ecosystem is update for 2.12 and half isn't, causing issues.

eed3si9n:
This bumps up scala-xml to 2.x.

Fixes https://github.com/sbt/sbt/issues/6997


mkurz:
Will you backport this for a 1.7.2 release? If not, can you meanwhile release 1.8.0-M1 (or RC1) so we can make use of that?
Also Scala 2.13.9 is released and should be upgraded in sbt.

eed3si9n:
Due to the eviction error difference, I think we should release it as 1.8.0-M1.

tOverney:
Is it possible to get a version published with this change?
To get Scoverage to work for 2.13.9 we need the new version of sbt-scoverage but that one already pulls scala-xml 2.1.x
and thus it's the eviction warning festival :)

mkurz:
+1 At least 1.8.0-M1 would be nice.

BTW: @tOverney Take care with Scala 2.13.9, there is a bug that can bite you, at least when not using [MiMa](https://github.com/lightbend/mima), see https://github.com/scala/bug/issues/12650

mkurz:
sbt 1.7.3 now pulls in Scala 2.12.17, see https://github.com/playframework/playframework/issues/11522

eed3si9n:
I didn't expect that to happen. As far as I know the scala-compiler dependency is still on 2.12.16 - https://github.com/sbt/sbt/blob/v1.7.3/project/Dependencies.scala#L7

SethTisue:
Could the cause be  https://github.com/coursier/coursier/pull/2548 ?

UPDATE: oh, I see @unkarjedy already suspected that over on the Play ticket. Well, I suspect it too.

SethTisue:
> sbt 1.7.3 now pulls in Scala 2.12.17,

Are you sure? Are you sure you're not just seeing scala-xml 2.x coming in via https://github.com/coursier/coursier/pull/2548 and https://github.com/sbt/sbt/pull/7059 ?

SethTisue:
I see there's also discussion at https://github.com/sbt/sbt/issues/6997

mkurz:
@SethTisue Yeah sure that's want I meant. However users may don't know internals, so they they see this problem only by bumping 1.7.2 to 1.7.3, that's what I meant.

project/Dependencies.scala
```diff
@@ -102,9 +102,9 @@ object Dependencies {
 
   val scalaXml = Def.setting(
     if (scalaBinaryVersion.value == "3") {
-      "org.scala-lang.modules" %% "scala-xml" % "2.0.1"
+      "org.scala-lang.modules" %% "scala-xml" % "2.1.0"
```
rydenius, line 105: Out of curiosity, was it intentional to keep the if statement when both branches return the same value? Or is it a mistake that both return the same?

eed3si9n: We can get rid of the if expression here.

PR #7043, opened 2022-10-02
[1.8.x] Protect against missing edges in graph by using map get vs apply

eed3si9n:
This backports https://github.com/sbt/sbt/pull/6978

## original description

Connects to #6976 

It's possible for there to be orphaned edges in the constructed graph. Using `map.apply` to retrieve the edge will then fail  with `NoSuchElementException`. This PR updates to using `get` that safely returns an `Option` instead of throwing. 

This does _not_ fix the underlying issue. The produced dependency graph will likely be incorrect if the edges aren't present. This will only prevent an exception from being thrown.


lightbend-cla-validator:
Hi @frosforever,

Issue #7058, opened 2022-10-29
Update to the latest Coursier


eed3si9n:
The current version of Coursier used by sbt has lagged behind. We should probably bump to latest.

PR #7060, opened 2022-10-31
[1.7.x] lm-coursier-shaded 2.0.12

eed3si9n:
https://github.com/coursier/sbt-coursier/releases/tag/v2.0.12

Fixes https://github.com/sbt/sbt/issues/7058

### DIFF
--- a/main/src/main/scala/sbt/coursierint/LMCoursier.scala
+++ b/main/src/main/scala/sbt/coursierint/LMCoursier.scala
@@ -20,6 +20,7 @@ import lmcoursier.definitions.{
   Strict => CStrict,
 }
 import lmcoursier._
+import lmcoursier.syntax._
 import lmcoursier.credentials.Credentials
 import Keys._
 import sbt.internal.util.Util
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -77,7 +77,7 @@ object Dependencies {
   def addSbtZincCompile = addSbtModule(sbtZincPath, "zincCompile", zincCompile)
   def addSbtZincCompileCore = addSbtModule(sbtZincPath, "zincCompileCore", zincCompileCore)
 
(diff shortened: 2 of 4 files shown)


julienrf:
I wonder if this change introduced a binary incompatibility. I had to apply the following change in sbt-version-policy to upgrade from 1.7.2 to 1.7.3: https://github.com/scalacenter/sbt-version-policy/pull/147/commits/db2f37c8a60ac89cdd9914d218b8de9858d397a5

(otherwise, I got the following error: https://github.com/scalacenter/sbt-version-policy/actions/runs/3402173263/jobs/5657742540#step:5:1665

eed3si9n:
lm-coursier-shaded should shade Shapeless if Coursier is using it, so I am not sure how what resulted to policy change related to Shapeless. Could it also be something to do with the implementation change in Coursier itself since it got fast forwarded by a year?

julienrf:
The error wasn’t related to Shapeless, but to the API of Coursier:

~~~ text
Error:  /tmp/sbt_57e6ea27/build.sbt:47: error: value only is not a member of object lmcoursier.definitions.ModuleMatchers
Error:          ModuleMatchers.only(Module(Organization("com.chuusai"), ModuleName("shapeless_2.12"), Map())) -> Reconciliation.Strict
Error:                         ^
Error:  sbt.compiler.EvalException: Type error in expression
~~~

There used to be a method `lmcoursier.definitions.ModuleMatchers.only` with a compatible signature, but it is not here anymore.

eed3si9n:
Ah. That's because we switched code generation to Scalafix for Scala 3 purpose. We now need:

```scala
import lmcoursier.syntax._
```

I had to do that in https://github.com/sbt/sbt/pull/7060 as well.

julienrf:
Yes, but my point is that it seems that lm-coursier-shaded 2.0.12 is not binary compatible with version 2.0.10. This is surprising given that it’s a patch version bump.

Issue #7082, opened 2022-12-01
SBT scripted tests fail when run under github actions Windows CI image


mdedetrich:
## steps

With an sbt plugin that has scripted tests, run it under a github actions Windows CI image

## problem

The scripted test fails because the sbt server (i.e. https://www.scala-sbt.org/1.x/docs/sbt-server.html#Server+discovery+and+authentication) is unable to communicate with the sbt client, i.e. 

```
Error:  java.net.ConnectException: Connection refused: connect
Error:  	at java.base/java.net.PlainSocketImpl.connect0(Native Method)
Error:  	at java.base/java.net.PlainSocketImpl.socketConnect(PlainSocketImpl.java:101)
Error:  	at java.base/java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:412)
Error:  	at java.base/java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:255)
Error:  	at java.base/java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:237)
Error:  	at java.base/java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
Error:  	at java.base/java.net.Socket.connect(Socket.java:609)
Error:  	at java.base/java.net.Socket.connect(Socket.java:558)
Error:  	at java.base/java.net.Socket.<init>(Socket.java:454)
Error:  	at java.base/java.net.Socket.<init>(Socket.java:264)
Error:  	at xsbt.IPC$.client(IPC.scala:21)
Error:  	at sbt.BasicCommands$.readMessage(BasicCommands.scala:472)
Error:  	at sbt.BasicCommands$.doRead(BasicCommands.scala:448)
Error:  	at sbt.BasicCommands$.$anonfun$read$2(BasicCommands.scala:435)
Error:  	at sbt.Command$.$anonfun$applyEffect$4(Command.scala:150)
Error:  	at sbt.Command$.$anonfun$applyEffect$2(Command.scala:145)
Error:  	at sbt.Command$.process(Command.scala:189)
Error:  	at sbt.MainLoop$.$anonfun$processCommand$5(MainLoop.scala:245)
Error:  	at scala.Option.getOrElse(Option.scala:189)
Error:  	at sbt.MainLoop$.process$1(MainLoop.scala:245)
Error:  	at sbt.MainLoop$.processCommand(MainLoop.scala:276)
[...]

PR #7083, opened 2022-12-02
Use better method to find free port

mdedetrich:
Rather than trying to find random open ports we instead should instead use this technique https://stackoverflow.com/a/2675416 to find free ports which will also work on Windows https://github.com/sbt/sbt/issues/7082.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -186,6 +186,7 @@ jobs:
       shell: bash
       run: |
         # test building sbtn on Windows
+        sbt "commandProj/testOnly xsbt.IPCSpec"
         sbt "-Dsbt.io.virtual=false" nativeImage
         # test launcher script
         echo build using JDK 8, test using JDK 8, on Windows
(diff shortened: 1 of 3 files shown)


lightbend-cla-validator:
Hi @mdedetrich,

mdedetrich:
@eed3si9n So while this PR should theoretically solve the problem, I still cannot replicate the bug (using https://github.com/sbt/sbt/pull/7084 to try and replicate the original bug).

.github/workflows/ci.yml
```diff
@@ -186,6 +186,7 @@ jobs:
       shell: bash
       run: |
         # test building sbtn on Windows
+        sbt "commandProj/testOnly xsbt.IPCSpec"
```
mdedetrich, line 189: This is to make sure the added test runs on Windows

mdedetrich:
Setting this to draft for now as I don't think its the actual reason behind sbt scripted tests failing on `windows-latest` CI but its still a useful change so I will work on it after https://github.com/sbt/sbt/pull/7087 is verified.

eed3si9n:
I've incorporated this idea into sbt 2.x here - https://github.com/sbt/sbt/pull/8431, so I'm closing this PR.

PR #7084, opened 2022-12-02
Replicate failing to open port on Windows

mdedetrich:
This is just a draft PR to replicate failing to open a port on windows CI. **DON'T** merge this PR, its just to replicate something.


mdedetrich:
So I am still having problems trying to replicate the original bug. I am thinking whether the issue is more due to timeouts then something else, i.e. a port is calculated but not immediately connected to and by the time the scripted tests connect to the port its already closed? @eed3si9n do you have any other ideas

PR #7085, opened 2022-12-06
fix: bspConfig sbt.script - use `Path` env variable as a fallback for Windows

dos65:
Originally reported in https://github.com/scalameta/metals/issues/4702#issuecomment-1339137772

### DIFF
--- a/protocol/src/main/scala/sbt/internal/bsp/BuildServerConnection.scala
+++ b/protocol/src/main/scala/sbt/internal/bsp/BuildServerConnection.scala
@@ -63,8 +63,10 @@ object BuildServerConnection {
     // For those who use an old sbt script, the -Dsbt.script is not set
     // As a fallback we try to find the sbt script in $PATH
     val fileName = if (Properties.isWin) "sbt.bat" else "sbt"
-    val envPath = sys.env.getOrElse("PATH", "")
-    val allPaths = envPath.split(File.pathSeparator).map(Paths.get(_))
+    val envPath = sys.env.collectFirst {
+      case (k, v) if k.toUpperCase() == "PATH" => v
+    }
+    val allPaths = envPath.map(_.split(File.pathSeparator).map(Paths.get(_))).getOrElse(Array.empty)
     allPaths
       .map(_.resolve(fileName))
       .find(file => Files.exists(file) && Files.isExecutable(file))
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
At least one commit author (v.checlyshov@tinkoff.ru) is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

eed3si9n:
@dos65 Thanks for the contribution!

protocol/src/main/scala/sbt/internal/bsp/BuildServerConnection.scala
```diff
@@ -63,7 +63,7 @@ object BuildServerConnection {
     // For those who use an old sbt script, the -Dsbt.script is not set
     // As a fallback we try to find the sbt script in $PATH
     val fileName = if (Properties.isWin) "sbt.bat" else "sbt"
-    val envPath = sys.env.getOrElse("PATH", "")
+    val envPath = sys.env.get("PATH").orElse(sys.env.get("Path")).getOrElse("")
```
eed3si9n: Instead of singling out `PATH` and `Path`, would it make sense at this point to filter to all keys that becomes `PATH` if you turned it to upper case, and pick the first entry?

dos65: Yep, it makes sense. Changed

eed3si9n (approved):

protocol/src/main/scala/sbt/internal/bsp/BuildServerConnection.scala
martingd: @eed3si9n Doesn't it potentially break semantics on *NIX where you could have a variable PATH and also other case variants?
The suggested code gives no guarantee which would be the first.

I would suggest that the “filter to all keys that becomes PATH if you turned it to upper case, and pick the first entry” was only used on Windows (to match Windows semantics) but on *NIX you still only looked for PATH (upper case) We already have special handling for Windows for the name of the sbt launcher.

dos65: @martingd 
> We already have special handling for Windows for the name of the sbt launcher

Could you point where is this handling defined?
I think the best option would be to reuse/copy-paste the existing logic

PR #7096, opened 2022-12-21
[1.9.x] Smooth transition to valid Maven pattern of sbt plugins

### LINKED ISSUE #3410: The maven pattern for sbt plugins is invalid based on the spec
I've been in touch with the Bintray team because of https://github.com/sbt/sbt-bintray/pull/125.
  
It happens that the reason why sbt bintray does not accept the fix in the aforementioned PR is because **sbt has been using an incorrect Maven pattern for sbt plugins**. This is the reply I got from the Bintray team:
  
> Going off of the Maven Design Documentation, Bintray expects the following Maven layout format:
`[orgPath]/[module]/[baseRev](-[folderItegRev])/[module]-[baseRev](-[fileItegRev])(-[classifier]).[ext]`. Note the “module” field appears twice.
  
Our pattern is the following: https://github.com/sbt/librarymanagement/blob/53c80f076a5e9ab469037f352a4d68173140e216/core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala#L357-L358, which is obviously wrong (there's a concept of module and a concept of artifact, where both should be the same).
  
The reason why this works for Sonatype is because they have a permissive layout for projects whose publishing rules do not abide by the spec. Bintray doesn't, and that's why no one has succeeded to publish sbt plugins to Bintray maven repositories.
  
This is a major issue that IMO should be sorted out as soon as possible. The only way I see to solve this problem is to add another pattern to sbt and make it use alongside the wrong pattern (we cannot remove it, too many projects depend on it).

adpi2:
The goal of this PR is to fix https://github.com/sbt/sbt/issues/3410 in sbt 1.x without creating a big-bang in the ecosystem of sbt plugins.

## Summary

The implemented strategy is:
- sbt 1.9 publishes valid poms and old poms
- Coursier 2.1 and sbt 1.9 try first to resolve the valid pom and fallback to the invalid pom if the valid pom is missing (backward compatibility)
- Older versions of Coursier and sbt can still resolve newly published sbt plugins from its old pom (forward compatibility)

Thus the migration of plugins can happen in any order:
- A plugin can migrate to sbt 1.9.x even if it depends on a plugin published by sbt <= 1.8.x.
- A plugin can stay on sbt <= 1.8.x even if it depends on a plugin published by 1.9.x.

It is worth noting that the descriptor of an sbt plugin dependency is still the same: `org.example:example:1.0.0` with extra-attributes `scalaVersion:2.12` and `sbtVersion:1.0`

## Motivation

From the user perspective almost nothing changes, the dependency to a plugin is still declared as:
```scala
addSbtPlugin("org.example" % "example" % "1.0.0")
```

The added value is:
- resolution of an sbt plugin works in enterprise environments
- `mvnrepository.com` starts indexing the sbt plugins
- statistics (downloads) are available on Sonatype for sbt plugins
- `javadoc.io` can find the scaladoc of sbt plugins
- we can fix the linking of sbt plugin dependencies in Scaladex
- we can use Maven to resolve sbt plugins
- (we can use other tools that depends on Maven to resolve sbt plugins)

## Associated PRs

- In Coursier: https://github.com/coursier/coursier/pull/2633 contains the resolution of an sbt plugin based on the valid pattern and the parsing of the valid pom
- In librarymanagement: https://github.com/sbt/librarymanagement/pull/409 contains the resolution of an sbt plugin based on the valid pattern, the parsing of the valid pom and the generation of the valid pom

### Generating the valid pom file

We consider a pom file is valid if Maven can resolve it and check its consistency. More precisely, the path of the pom file and the declared `artifactId` in the pom file must be consistent.

Given module `org.example:example:1.0.0` with extra-attributes `scalaVersion:2.12` and `sbtVersion:1.0`, its valid Maven path must be `org/example/example_2.12_1.0/1.0.0/example_2.12_1.0-1.0.0.pom` and the corresponding Maven module ID is:

```xml
<groupId>org.example</groupId>
<artifactId>example_2.12_1.0</artifactId>
<version>1.0.0</version>
<properties>
  <scalaVersion>2.12</scalaVersion>
  <sbtVersion>1.0</sbtVersion>
</properties>
```

Maven must also be able to resolve the sbt plugin dependencies. Thus all declared sbt plugin dependencies in the pom file must contain valid artifact IDs, with the sbt cross-version.

This pom file is generated as if the declared module ID is `org.example:example_2.12_1.0:1.0.0` with no extra attributes. However, in sbt and Coursier, we must maintain the old-style module ID, with extra-attributes, to get the bi-directional compatibilty.

### Resolving an sbt plugin

Given module `org.example:example:1.0.0` with extra-attributes `scalaVersion:2.12` and `sbtVersion:1.0`, we first try to resolve `org/example/example_2.12_1.0/1.0.0/example_2.12_1.0-1.0.0.pom` and we fallback to `org/example/example_2.12_1.0/1.0.0/example-1.0.0.pom`.

Whenever we parse a pom file, we must make sure to remove the cross-version part if it is redundant with the extra-attributes.
So parsing `example_2.12_1.0-1.0.0.pom` returns the same module ID as parsing `example-1.0.0.pom`. That is `org.example:example:1.0.0` with extra-attributes `scalaVersion:2.12` and `sbtVersion:1.0`. All dependencies, in the pom file, are also treated the same. The parsed dependencies of `example_2.12_1.0-1.0.0.pom` are the same as the ones of `example-1.0.0.pom`. Thus the conflict resolution is working well: sbt can resolve conflicts between old and new poms, old and new dependencies.

### Tests

To test these changes thoroughly I created a diamond graph of sbt plugins:
```
            sbt-plugin-example-diamond
                       / \
sbt-plugin-example-left   sbt-plugin-example-right
                       \ /
            sbt-plugin-example-bottom
```

I published the artifacts to Maven Central:
- https://repo1.maven.org/maven2/ch/epfl/scala/sbt-plugin-example-diamond_2.12_1.0/
- https://repo1.maven.org/maven2/ch/epfl/scala/sbt-plugin-example-left_2.12_1.0/
- https://repo1.maven.org/maven2/ch/epfl/scala/sbt-plugin-example-right_2.12_1.0/
- https://repo1.maven.org/maven2/ch/epfl/scala/sbt-plugin-example-bottom_2.12_1.0/

Depending on the version of `sbt-plugin-example-diamond`, from `0.1.0` to `0.5.0`, different parts have migrated to the new pattern, and there can be conflict resolution on `sbt-plugin-example-bottom`. See the scripted `dependency-management/sbt-plugin-diamond` for more details.

`sbt-plugin-example-diamond:0.5.0` is fully migrated to the valid Maven pattern. Hence Maven can resolve its full graph of dependencies, and resolve the conflict on `sbt-plugin-example-bottom`.

The Maven-style publication is tested in the scripted `dependency-management/sbt-plugin-publish`.

## Next Steps

- Once sbt 1.8 is EOL we can drop the publication of the old Maven pom and artifact.
- If ever we drop the support of extra-attributes in Maven, we will still be able to resolve sbt plugins as `org.example:example_2.12_1.0:1.0.0`. We probably still want to resolve older plugins though, so it may not be possible to remove the resolution fallback and the extra-attributes mechanism.
- In sbt 2.x we should drop all this complexity and use the `CrossVersion` mechanism to publish sbt plugin, similarly to Scala.js or Scala Native artifacts. The cross version would then be something like `_sbt2_3`.
- After sbt 1.x in EOL we can drop the support of extra-attributes in Coursier

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2827,8 +2827,51 @@ object Classpaths {
 
   val jvmPublishSettings: Seq[Setting[_]] = Seq(
     artifacts := artifactDefs(defaultArtifactTasks).value,
-    packagedArtifacts := packaged(defaultArtifactTasks).value
+    packagedArtifacts := packaged(defaultArtifactTasks).value ++
+      Def
+        .ifS(sbtPlugin.toTask)(mavenArtifactsOfSbtPlugin)(Def.task(Map.empty[Artifact, File]))
+        .value
   ) ++ RemoteCache.projectSettings
+
+  /**
+   * Produces the Maven-compatible artifacts of an sbt plugin.
+   * It adds the sbt-cross version suffix into the artifact names, and it generates a
+   * valid POM file, that is a POM file that Maven can resolve.
+   */
+  private def mavenArtifactsOfSbtPlugin: Def.Initialize[Task[Map[Artifact, File]]] =
+    Def.ifS(publishMavenStyle.toTask)(Def.task {
+      val crossVersion = sbtCrossVersion.value
+      val legacyArtifact = (makePom / artifact).value
+      val pom = makeMavenPomOfSbtPlugin.value
+      val legacyPackages = packaged(defaultPackages).value
+
+      def addSuffix(a: Artifact): Artifact = a.withName(crossVersion(a.name))
+      val packages = legacyPackages.map { case (artifact, file) => addSuffix(artifact) -> file }
+      packages + (addSuffix(legacyArtifact) -> pom)
+    })(Def.task(Map.empty))
+
+  private def sbtCrossVersion: Def.Initialize[String => String] = Def.setting {
+    val sbtV = sbtBinaryVersion.value
+    val scalaV = scalaBinaryVersion.value
+    name => name + s"_${scalaV}_$sbtV"
+  }
+
+  /**
+   * Generates a POM file that Maven can resolve.
+   * It appends the sbt cross version into all artifactIds of sbt plugins
+   * (the main one and the dependencies).
+   */
+  private def makeMavenPomOfSbtPlugin: Def.Initialize[Task[File]] = Def.task {
+    val config = makePomConfiguration.value
+    val nameWithCross = sbtCrossVersion.value(artifact.value.name)
+    val version = Keys.version.value
+    val pomFile = config.file.get.getParentFile / s"$nameWithCross-$version.pom"
+    val publisher = Keys.publisher.value
+    val ivySbt = Keys.ivySbt.value
+    val module = new ivySbt.Module(moduleSettings.value, appendSbtCrossVersion = true)
+    publisher.makePomFile(module, config.withFile(pomFile), streams.value.log)
+    pomFile
+  }
 
   val ivyPublishSettings: Seq[Setting[_]] = publishGlobalDefaults ++ Seq(
     artifacts :== Nil,
--- a/sbt-app/src/sbt-test/dependency-management/sbt-plugin-diamond/test
+++ b/sbt-app/src/sbt-test/dependency-management/sbt-plugin-diamond/test
@@ -1,0 +1,12 @@
+> v1/checkUpdate
+> v2/checkUpdate
+> v3/checkUpdate
+> v4/checkUpdate
+> v5/checkUpdate
+
+> set ThisBuild/useCoursier:=true
+> v1/checkUpdate
+> v2/checkUpdate
+> v3/checkUpdate
+> v4/checkUpdate
+> v5/checkUpdate
--- a/sbt-app/src/sbt-test/dependency-management/sbt-plugin-diamond/test
+++ b/sbt-app/src/sbt-test/dependency-management/sbt-plugin-diamond/test
@@ -1,12 +1,12 @@
-> v1/checkUpdate
-> v2/checkUpdate
-> v3/checkUpdate
-> v4/checkUpdate
-> v5/checkUpdate
+> v1 / checkUpdate
+> v2 / checkUpdate
+> v3 / checkUpdate
+> v4 / checkUpdate
+> v5 / checkUpdate
 
-> set ThisBuild/useCoursier:=true
-> v1/checkUpdate
-> v2/checkUpdate
-> v3/checkUpdate
-> v4/checkUpdate
-> v5/checkUpdate
+> set ThisBuild / useCoursier:=false
+> v1 / checkUpdate
+> v2 / checkUpdate
+> v3 / checkUpdate
+> v4 / checkUpdate
+> v5 / checkUpdate
--- a/sbt-app/src/sbt-test/dependency-management/sbt-plugin-publish/build.sbt
+++ b/sbt-app/src/sbt-test/dependency-management/sbt-plugin-publish/build.sbt
@@ -1,0 +1,122 @@
+import scala.util.matching.Regex
+
+lazy val repo = file("test-repo")
+lazy val resolver = Resolver.file("test-repo", repo)
+
+lazy val example = project.in(file("example"))
+  .enablePlugins(SbtPlugin)
+  .settings(
+    organization := "org.example",
+    addSbtPlugin("ch.epfl.scala" % "sbt-plugin-example-diamond" % "0.5.0"),
+    publishTo := Some(resolver),
+    checkPackagedArtifacts := checkPackagedArtifactsDef.value,
+    checkPublish := checkPublishDef.value
+  )
+
+lazy val testMaven = project.in(file("test-maven"))
+  .settings(
+    addSbtPlugin("org.example" % "example" % "0.1.0-SNAPSHOT"),
+    externalResolvers -= Resolver.defaultLocal,
+    resolvers += {
+      val base = (ThisBuild / baseDirectory).value
+      MavenRepository("test-repo", s"file://$base/test-repo")
+    },
+    checkUpdate := checkUpdateDef(
+      "example_2.12_1.0-0.1.0-SNAPSHOT.jar",
+      "sbt-plugin-example-diamond_2.12_1.0-0.5.0.jar",
+      "sbt-plugin-example-left_2.12_1.0-0.3.0.jar",
+      "sbt-plugin-example-right_2.12_1.0-0.3.0.jar",
+      "sbt-plugin-example-bottom_2.12_1.0-0.3.0.jar",
+    ).value
+  )
+
+lazy val testLocal = project.in(file("test-local"))
+  .settings(
+    addSbtPlugin("org.example" % "example" % "0.1.0-SNAPSHOT"),
+    checkUpdate := checkUpdateDef(
+      "example.jar", // resolved from local repository
+      "sbt-plugin-example-diamond_2.12_1.0-0.5.0.jar",
+      "sbt-plugin-example-left_2.12_1.0-0.3.0.jar",
+      "sbt-plugin-example-right_2.12_1.0-0.3.0.jar",
+      "sbt-plugin-example-bottom_2.12_1.0-0.3.0.jar",
+    ).value
+  )
+
+lazy val checkPackagedArtifacts = taskKey[Unit]("check the packaged artifacts")
+lazy val checkPublish = taskKey[Unit]("check publish")
+lazy val checkUpdate = taskKey[Unit]("check update")
+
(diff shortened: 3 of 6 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2827,9 +2827,45 @@ object Classpaths {
 
   val jvmPublishSettings: Seq[Setting[_]] = Seq(
     artifacts := artifactDefs(defaultArtifactTasks).value,
-    packagedArtifacts := packaged(defaultArtifactTasks).value
+    packagedArtifacts := {
+      if (sbtPlugin.value) packagedSbtPluginArtifacts.value
+      else packaged(defaultArtifactTasks).value
```
julienrf: This is very hard to review because I need to check that the implementation of `packagedSbtPluginArtifacts` is the same as `packaged(defaultArtifactTasks)` plus it adds the legacy files (or the correct files, depending on the strategy you chose).

What do you think instead of always returning `packaged(defaultArtifactTasks)` and also producing additional files (which, in the case of sbt plugins would be the legacy files -- or the correct files -- and would be empty otherwise):

~~~ scala
packagedArtifacts := packaged(defaultArtifactTasks).value ++ legacySbtPluginArtifacts.value
~~~

Where:

~~~ scala
legacySbtPluginArtifacts := {
  if (sbtPlugin.value) ...
  else Map.empty
}
~~~

By the way, I believe we should use `Def.ifS` instead of putting `if` expressions within task definitions:

~~~ scala
legacySbtPluginArtifacts :=
  Def.ifS(sbtPlugin.toTask)(Def.task {
    // Publish legacy-style sbt-plugin artifacts
    ...
  })(Def.task {
    // Nothing to do if this is not an sbt plugin
    Map.empty[Artifact, File]
  })
~~~

sbt-app/src/sbt-test/dependency-management/sbt-plugin-diamond/test
```diff
@@ -0,0 +1,12 @@
+> v1/checkUpdate
+> v2/checkUpdate
+> v3/checkUpdate
+> v4/checkUpdate
+> v5/checkUpdate
+
+> set ThisBuild/useCoursier:=true
```
julienrf: Isn’t that already the default? Maybe add a `set ThisBuild / useCoursier := false` at the top of the file to make sure (and explicit) that we use Ivy?

sbt-app/src/sbt-test/dependency-management/sbt-plugin-publish/build.sbt
```diff
@@ -0,0 +1,116 @@
...
+    ).value
+  )
+
+lazy val checkPublishedArtifacts = taskKey[Unit]("check the published artifacts")
+lazy val checkPublish = taskKey[Unit]("check publish")
+lazy val checkUpdate = taskKey[Unit]("check update")
+
+def checkPackagedArtifactsDef: Def.Initialize[Task[Unit]] = Def.task {
+  val packagedArtifacts = Keys.packagedArtifacts.value
+  val name = artifact.value.name
```
julienrf: Should we write `"example"` here to make sure the artifact `name` does not already contain the suffix?

sbt-app/src/sbt-test/dependency-management/sbt-plugin-publish/build.sbt
```diff
@@ -0,0 +1,116 @@
...
+    checkUpdate := checkUpdateDef(
+      "example_2.12_1.0-0.1.0-SNAPSHOT.jar",
+      "sbt-plugin-example-diamond_2.12_1.0-0.5.0.jar",
+      "sbt-plugin-example-left_2.12_1.0-0.3.0.jar",
+      "sbt-plugin-example-right_2.12_1.0-0.3.0.jar",
+      "sbt-plugin-example-bottom_2.12_1.0-0.3.0.jar",
+    ).value
+  )
+
+lazy val checkPublishedArtifacts = taskKey[Unit]("check the published artifacts")
```
julienrf: Maybe you want to rename it to `checkPackagedArtifacts`?

sbt-app/src/sbt-test/dependency-management/sbt-plugin-publish/test
```diff
@@ -0,0 +1,6 @@
+> example / checkPublishedArtifacts
+> example / checkPublish
+
+> testProject / checkUpdate
+> set testProject / useCoursier := false
+> testProject / checkUpdate
```
julienrf: Do we cover publishing to ivy repositories? (e.g. `publishLocal`)

Do we cover resolving a plugin that we just published?

sbt-app/src/sbt-test/dependency-management/sbt-plugin-diamond/test
adpi2: Thanks for catching that

main/src/main/scala/sbt/Defaults.scala
adpi2: Ok, I applied the suggestion.

I think the `Def.ifS` is not necessary because the `Def.task` macro is smart enough to handle that. But maybe it makes things slightly clearer.

ScalaWilliam: @adpi2 just for info, we found that the `if` macro does not work for SBT 1.2.x (and possibly some versions above that). Not sure if it makes a difference.

sbt-app/src/sbt-test/dependency-management/sbt-plugin-publish/test
ScalaWilliam: That's a great question; I thought the Ivy pattern and approach is entirely different from that of SBT+Maven?

adpi2: I expanded the scripted sbt-plugin-publish to test the publication to  `.ivy/local` and resolution from it.

For the Ivy publication, there are two cases, depending on the value of `publishMavenStyle`.
- `publishMavenStyle := false`: Strictly no changes, we publish:
```
org.example/example/scala_2.12/sbt_1.0/0.1.0-SNAPSHOT/
    ├── docs
    │   └── example-javadoc.jar
    ├── ivys
    │   └── ivy.xml
    ├── jars
    │   └── example.jar
    └── src
        └── example-sources.jar
```
- `publishMavenStyle.value := true`: We publish the legacy and new poms, and all the associated artifacts.
```
org.example/example/scala_2.12/sbt_1.0/0.1.0-SNAPSHOT/
    ├── docs
    │   ├── example_2.12_1.0-javadoc.jar
    │   └── example-javadoc.jar
    ├── ivys
    │   └── ivy.xml
    ├── jars
    │   ├── example_2.12_1.0.jar
    │   └── example.jar
    ├── poms
    │   ├── example_2.12_1.0.pom
    │   └── example.pom
    └── src
        ├── example_2.12_1.0-sources.jar
        └── example-sources.jar
```

In the test, we do a `publishLocal` on a project that is configured for Maven, so it is a Maven-style publication to Ivy.

julienrf (approved):
Thank you @adpi2 for all the PRs and the detailed explanations. I think this is a good approach.

ScalaWilliam:
@adpi2 was just wondering what those failing tests are, do you have much idea? It seems they would get fixed by stopping the cache upload but then I am not sure if that's what we want.

sjrd:
@adpi2 is off for a few weeks. There might be some delay here.

adpi2:
@eed3si9n I need a release of https://github.com/sbt/librarymanagement/pull/409 to merge this PR, maybe just a milestone version. Let me know if I can help.

eed3si9n:
https://github.com/sbt/librarymanagement/releases/tag/v1.9.0-M1 should be on its way to Maven Central.

adpi2:
@eed3si9n Something may have gone wrong, since I cannot find it: https://repo1.maven.org/maven2/org/scala-sbt/librarymanagement-core_2.12/

eed3si9n:
> @eed3si9n Something may have gone wrong, since I cannot find it: https://repo1.maven.org/maven2/org/scala-sbt/librarymanagement-core_2.12/

My bad. I forgot to release it from Sonatype staging. It should be there in 10 ~ 30 min.

adpi2:
@eed3si9n This PR is ready for a final review.

eed3si9n (approved):
Thanks!

ScalaWilliam:
This is super exciting. When is the planned release date @eed3si9n ?

@sjrd - can't wait for the scala-js SBT plug-in once this is out :-)

ScalaWilliam:
@sjrd confirmed working.

any chance for this one as well? for cross compilation.

https://github.com/portable-scala/sbt-crossproject

PR #7100, opened 2022-12-29
add debugProvider and canDebug in BSP contra

ckipp01:
I believe this was just an oversight that it's not marked as true since sbt can handle `debugSession/start`. This change just ensures that during the initialization process sbt says it's a `debugProvider` for the same languages as `runProvider` and `testProvider`. It also correctly marks the build targets as `canDebug`, unless they are sbt targets.

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
@@ -50,6 +50,7 @@ object BuildServerProtocol {
     CompileProvider(BuildServerConnection.languages),
     TestProvider(BuildServerConnection.languages),
     RunProvider(BuildServerConnection.languages),
+    DebugProvider(Vector.empty),
     dependencySourcesProvider = true,
     resourcesProvider = true,
     outputPathsProvider = true,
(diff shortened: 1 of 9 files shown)


lightbend-cla-validator:
Hi @ckipp01,

adpi2:
> I believe this was just an oversight that it's not marked as true since sbt can handle debugSession/start.

Actually, sbt cannot handle `debugSession/start` by default: it is the [sbt-debug-adapter](https://github.com/scalacenter/scala-debug-adapter) plugin that provides this capability.

So it would be more accurate to update those fields in [sbt-debug-adapter](https://github.com/scalacenter/scala-debug-adapter) itself, with something like this:

```scala
bspBuildTarget := {
  bspBuildTarget.value.copy(canDebug = true)
},
```

Unfortunately, if we do so, then it would break the compatibility of the plugin with older sbt versions because of the missing `canDebug` field. That's why I think I did not update anything anywhere.

Do you think we should update those fields in sbt even if it is not completely accurate?

ckipp01:
> Actually, sbt cannot handle debugSession/start by default: it is the [sbt-debug-adapter](https://github.com/scalacenter/scala-debug-adapter) plugin that provides this capability.

Ahhhh I totally missed this. Of course, that makes sense.

> Do you think we should update those fields in sbt even if it is not completely accurate?

I don't think so actually, not just in core sbt. I think if there would be a way for the sbt plugin to update them, then for sure we should do that, but I don't think we should just mark them as `true` like I have them now.

> Unfortunately, if we do so, then it would break the compatibility of the plugin with older sbt versions because of the missing canDebug field. 

Maybe then something we do in 2.x?

adpi2:
> Maybe then something we do in 2.x?

Absolutely. I suggest we merge the changes in the `bsp.contra` so that it is ready for sbt 2.x.

ckipp01:
> > Maybe then something we do in 2.x?
> 
> Absolutely. I suggest we merge the changes in the `bsp.contra` so that it is ready for sbt 2.x.

Alright, so I left the stuff in `bsp.contra` and then just make sure in the `DebugProvider` we just return an empty collection and that `canDebug` is `false`. This is what you meant right?

adpi2 (approved):
Thanks

PR #7111, opened 2023-01-03
[1.9.x] Make use of sbtn aarch64 binary where possible

mkurz:
In addition to #7110

### DIFF
--- a/sbt
+++ b/sbt
@@ -171,6 +171,7 @@ acquire_sbtn () {
   local target="$p/sbtn"
   local archive_target=
   local url=
+  local arch = "x86_64"
   if [[ "$OSTYPE" == "linux-gnu"* ]]; then
     arch=$(uname -m)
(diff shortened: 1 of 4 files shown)


sbt
```diff
@@ -171,6 +171,7 @@ acquire_sbtn () {
   local target="$p/sbtn"
   local archive_target=
   local url=
+  local arch = "x86_64"
```
mkurz, line 174: @eed3si9n `arch` needs to be defined as default, otherwise, if not using linux, it will be undefined in below in line 199 below (`echoerr "downloading sbtn ${sbtn_v} for ${arch}"`)

eed3si9n (approved):
Thanks!

sbt
fredfp: this doesn't work on mac, it should be `local arch="x86_64"`:

> /opt/homebrew/Cellar/sbt/1.8.2/libexec/bin/sbt: line 174: local: `=': not a valid identifier

Issue #7117, opened 2023-01-04
A `NoClassDefFoundError` exception is raised in sbt 1.8.1 on M1


jcazevedo:
## steps

Start `sbt` in version 1.8.1 on an M1 Mac via:

```
$ sbt
```

## problem

The following exception is raised on startup:

```
Exception in thread "sbt-socket-server" java.lang.NoClassDefFoundError: Could not initialize class org.scalasbt.ipcsocket.JNIUnixDomainSocketLibraryProvider
	at org.scalasbt.ipcsocket.UnixDomainSocketLibraryProvider.get(UnixDomainSocketLibraryProvider.java:26)
	at org.scalasbt.ipcsocket.UnixDomainSocketLibraryProvider.maxSocketLength(UnixDomainSocketLibraryProvider.java:31)
	at sbt.internal.server.Server$$anon$1$$anon$2.$anonfun$run$1(Server.scala:75)
	at scala.util.Try$.apply(Try.scala:213)
	at sbt.internal.server.Server$$anon$1$$anon$2.run(Server.scala:63)
```

## expectation

No exception is raised and the sbt server is able to start successfully.

## notes

I was able to dismiss the exception and get sbt server started successfully if I set `serverConnectionType := ConnectionType.Tcp` in my build.

mkurz:
That sucks. Probably something went wrong from ipcsocket 1.5 to 1.6.1: https://github.com/sbt/ipcsocket/compare/v1.5.0...v1.6.1
Unfortuntaly I don't have time to look into that, maybe @eed3si9n knows the answer faster anyway...

eed3si9n:
I am guessing that it's https://github.com/sbt/ipcsocket/commit/e9b01e175a50629c961071f65a924c906673ce6d?

I don't have M1 machine at hand. What does:

```scala
System.getProperty("os.arch", "").toLowerCase()
```

return on a REPL or sbt eval command?

mzuehlke:
Mac M1
```Scala
scala> System.getProperty("os.arch", "")
res0: String = aarch64
```

jcazevedo:
Do we also need a dylib built for darwin in `aarch64` under https://github.com/sbt/ipcsocket/tree/develop/src/main/resources?

mkurz:
Fix is probably something like changing the condition in https://github.com/sbt/ipcsocket/commit/e9b01e175a50629c961071f65a924c906673ce6d to 
```java
      if (arch.equals("amd64") || isMac) {
[...]

PR #7121, opened 2023-01-04
[1.8.x] Fixes M1 support

eed3si9n:
This is a backport of https://github.com/sbt/sbt/pull/7120

Fixes https://github.com/sbt/sbt/issues/7117

Problem
-------
sbt no longer works on a Mac with M1 (aarch64) chips because it fails on ipcsocket library, which we maintain.
This was caused by architecture detection that was added in 1.6.1, which did not take in account for macOS universal binary.

Solution
--------
Update to ipcsocket 1.6.2.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -26,7 +26,7 @@ object Dependencies {
   val launcherInterface = "org.scala-sbt" % "launcher-interface" % launcherVersion
   val rawLauncher = "org.scala-sbt" % "launcher" % launcherVersion
   val testInterface = "org.scala-sbt" % "test-interface" % "1.0"
-  val ipcSocket = "org.scala-sbt.ipcsocket" % "ipcsocket" % "1.6.1"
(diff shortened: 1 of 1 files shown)


mkurz (approved):

project/Dependencies.scala
```diff
@@ -26,7 +26,7 @@ object Dependencies {
   val launcherInterface = "org.scala-sbt" % "launcher-interface" % launcherVersion
   val rawLauncher = "org.scala-sbt" % "launcher" % launcherVersion
   val testInterface = "org.scala-sbt" % "test-interface" % "1.0"
-  val ipcSocket = "org.scala-sbt.ipcsocket" % "ipcsocket" % "1.6.1"
+  val ipcSocket = "org.scala-sbt.ipcsocket" % "ipcsocket" % "1.6.2"
```
mkurz, line 29: No tag yet? https://github.com/sbt/ipcsocket/tags

eed3si9n: Tagged - https://github.com/sbt/ipcsocket/releases/tag/v1.6.2

PR #7142, opened 2023-01-27
Fix forwarding Java diagnostics through BSP

### LINKED ISSUE #7099: BSP Diagnostics not reported for Java files
## steps

You can reproduce this by creating a minimal hello-world project, adding a Java file, and then connecting a client like Metals to it.

- `g8 scala/hello-world.g8`
- Create a `src/main/java/example/HelloJava.java`
- Paste in the following to that file
```java
package example;

public class HelloJava {
    public static void main(String[] args) {
        System.out.println("Hello World");
    }
}
```

- Open up Metals in the workspace and switch to sbt as your BSP server
- Open `HelloJava.java`
- Remove the `;` for example
- Save the file
- See no diagnostic is reported

## problem

The problem seems to be that the actual error is never reported with the BSP reporter. You _do_ actually see them, but they are just logged:

```
[Trace - 11:58:26 am] Received notification 'build/logMessage'
Params: {
  "type": 1,
  "message": "/Users/ckipp/Documents/scala-workspace/bsp-comparison/sbt-test/src/main/java/example/HelloJava.java:5:1:  error: \u0027;\u0027 expected"
}


[Trace - 11:58:26 am] Received notification 'build/logMessage'
Params: {
  "type": 1,
  "message": "        System.out.println(\"Hello World\")\n                                         ^1 error"
}


[Trace - 11:58:26 am] Received notification 'build/logMessage'
Params: {
  "type": 2,
  "message": "javac exited with exit code 1"
}
```

In this scenario unless the user is looking in the logs they'll have no idea something is wrong with their Java sources.

## expectation

I'd expect diagnostics from Java to be reported in the same manner that they are for Scala

## notes

From speaking a bit with @adpi2 I believe the issue is around here:

https://github.com/sbt/sbt/blob/06d42e76adf49199244dce9e27203bb60708b7be/main/src/main/scala/sbt/Defaults.scala#L2322-L2329

Since we have the BSP `reporter`, but it's not passed to `compileAllJava`.

adpi2:
Fix #7099 

The fix is to forward the diagnostics that have no pointers. To do so I re-implemented the logic I found in [Bloop](https://github.com/scalacenter/bloop/blob/07b4a1c2397a51375fee7bcc9e4d6c13db3f2f53/frontend/src/main/scala/bloop/logging/BspServerLogger.scala#L101-L110).

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
@@ -108,7 +108,7 @@ final class BuildServerReporterImpl(
       val hadProblems = bspCompileState.hasAnyProblems.remove(filePath)
 
       val reportedProblems = infos.getReportedProblems.toVector
-      val diagnostics = reportedProblems.flatMap(toDiagnostic)
+      val diagnostics = reportedProblems.map(toDiagnostic)
 
       // publish diagnostics if:
       // 1. file had any problems previously - we might want to update them with new ones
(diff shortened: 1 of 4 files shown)


ckipp01 (approved):
Thanks for tackling this @adpi2!

eed3si9n (approved):
Thanks, @adpi2!

PR #7148, opened 2023-02-10
Add `releaseNotesURL` setting

### LINKED ISSUE #7125: Add `releaseNotesURL` setting?
Scala Steward has adopted a `info.releaseNotesUrl` property for the POM.

https://contributors.scala-lang.org/t/add-release-notes-urls-to-your-poms/6059

Can we add a setting `releaseNotesURL` to populate it?

This would be similar to the `apiURL` setting which populates `info.apiURL` in the POM.
https://github.com/sbt/sbt/blob/973cd63e0b6bfe0d06d364f314f391a08a77d38b/main/src/main/scala/sbt/Keys.scala#L345

armanbilge:
Closes https://github.com/sbt/sbt/issues/7125

Scala Steward has adopted a `info.releaseNotesUrl` property for the POM.

https://contributors.scala-lang.org/t/add-release-notes-urls-to-your-poms/6059

Can we add a setting `releaseNotesURL` to populate it?

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -3347,7 +3347,7 @@ object Classpaths {
     }
     val p3 = releaseNotesURL.value match {
       case Some(u) =>
-        p1.extra("info.releaseNotesUrl" -> u.toExternalForm)
+        p2.extra("info.releaseNotesUrl" -> u.toExternalForm)
       case _ => p2
     }
     p3
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -3347,7 +3347,7 @@ object Classpaths {
     }
     val p3 = releaseNotesURL.value match {
       case Some(u) =>
-        p2.extra("info.releaseNotesUrl" -> u.toExternalForm)
+        p2.extra(SbtPomExtraProperties.POM_RELEASE_NOTES_KEY -> u.toExternalForm)
       case _ => p2
     }
     p3
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -200,7 +200,8 @@ object Defaults extends BuildCommon {
       autoCompilerPlugins :== true,
       scalaHome :== None,
       apiURL := None,
-      javaHome :== None,
+      releaseNotesURL := None
+        javaHome :== None,
       discoveredJavaHomes := CrossJava.discoverJavaHomes,
       javaHomes :== ListMap.empty,
       fullJavaHomes := CrossJava.expandJavaHomes(discoveredJavaHomes.value ++ javaHomes.value),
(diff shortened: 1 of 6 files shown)


lightbend-cla-validator:
Hi @armanbilge,

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -200,7 +200,8 @@ object Defaults extends BuildCommon {
       autoCompilerPlugins :== true,
       scalaHome :== None,
       apiURL := None,
-      javaHome :== None,
+      releaseNotesURL := None
```
eed3si9n: Missing comma here.

eed3si9n:
I wonder if it would be more straightforward to derive the release notes URL by something like:

```scala
ThisBuild / sourceRepo := Some(GitHubRepo("foo/bar"))
```

lightbend-cla-validator:
Hi @armanbilge,

armanbilge:
> I wonder if it would be more straightforward to derive the release notes URL by something like:

Sure, `scmInfo` could be derived from that as well. But that seems like a new feature / PR. :)

I suppose you are happy to move forward with this then? Still a handful of things to do, possibly in some upstream repos.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -3344,7 +3345,12 @@ object Classpaths {
         p1.extra(SbtPomExtraProperties.VERSION_SCHEME_KEY -> x)
       case _ => p1
     }
-    p2
+    val p3 = releaseNotesURL.value match {
+      case Some(u) =>
+        p1.extra("info.releaseNotesUrl" -> u.toExternalForm)
```
armanbilge: I assume this constant should be added to `SbtPomExtraProperties`, where is that exactly?

eed3si9n: https://github.com/sbt/librarymanagement/blob/develop/core/src/main/java/sbt/internal/librarymanagement/mavenint/SbtPomExtraProperties.java

eed3si9n:
> I suppose you are happy to move forward with this then? Still a handful of things to do, possibly in some upstream repos.

Yea. I am +1 on enriching POM for this purpose.

main/src/main/scala/sbt/Defaults.scala
armanbilge: - https://github.com/sbt/librarymanagement/pull/410

mzuehlke:
```suggestion
        p2.extra("info.releaseNotesUrl" -> u.toExternalForm)
```
I think this must reference `p2` here.

eed3si9n:
https://github.com/sbt/librarymanagement/releases/tag/v1.9.0-M1 should be on its way to Maven Central.

armanbilge:
I guess this is good to go? I looked at `apiURL` but didn't see any tests I could use as an example.

eed3si9n:
I think there are related scripted tests that you might be able to modify or copy-and-modify:
- https://github.com/sbt/sbt/blob/1.9.x/sbt-app/src/sbt-test/dependency-management/make-pom/build.sbt
- https://github.com/sbt/sbt/blob/1.9.x/sbt-app/src/sbt-test/dependency-management/evicted-semver-spec/build.sbt

armanbilge:
Thanks! While I was here I added a test for `apiURL` as well.

eed3si9n (approved):
Thanks!

PR #7154, opened 2023-02-20
Replace olafurpg/setup-scala with actions/setup-java

mzuehlke:
https://github.com/olafurpg/setup-scala/releases/tag/v14 is deprecated and migration to actions/setup-java is encouraged. That action is already used in `ci.yml`

PR #7161, opened 2023-02-27
Replace olafurpg/setup-scala with actions/setup-java 

mzuehlke:
https://github.com/olafurpg/setup-scala/releases/tag/v14 is deprecated and migration to actions/setup-java is encouraged.
That action is already used in `ci.yml`

Backport to 1.8.x see #7154


eed3si9n (approved):
Thanks!

PR #7165, opened 2023-03-02
Make skip work on publishLocal

mdedetrich:
So in a project I am working on I noticed the following comment from https://github.com/apache/incubator-pekko-grpc/blob/main/build.sbt#L144-L146, i.e.

```
// This project should use 'publish/skip := true', but we need
// to be able to `publishLocal` to run the interop tests as an
// sbt scripted test. At least skip scaladoc generation though.
```

This PR updates `publishTask` so that it respects `publishLocal / skip`. The current solution is a bit boilerplatery and is also missing a test however if I am on the right track then let me know and I can update the PR.

Also when the PR is ready I can create an equivalent PR for the 1.8.x branch (not sure if another sbt release for 1.8.x is expected). Either that or we can just cherry pick once this is merged


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -3585,15 +3585,57 @@ object Classpaths {
...
+
+  private def checkAndPublishLocal(
+      currentCon: PublishConfiguration,
+      publishLocalConf: PublishConfiguration,
+      publishSkipLocal: Boolean,
+      module: IvySbt#Module,
+      log: Logger,
+      ref: ProjectRef
+  )(notPublishLocal: () => Unit): Unit = {
+    if (currentCon == publishLocalConf) {
```
mdedetrich, line 3600: Is this equality comparison correct for checking if the current publish target is `publishLocal` or should something else be done?

SethTisue:
after this change, if I `publish / skip := true` but I don't set `publishLocal / skip`, does `publishLocal` publish or not?

will people need to set both in order to get the old behavior? asking because I think the old behavior is what is wanted most of the time

mdedetrich:
Yes I actually was thinking of this, if one sets `publish / skip := true` to true then I believe it would only apply to `publish`. This should be fairly easy to fix, i.e. instead of

```scala
skip := false
```

You would have

```scala
skip := false
publishLocal / skip := (publish / skip).value, // So we don't break previous behaviour
```

mdedetrich:
@SethTisue So I pushed a change with I think should preserve current behaviour as you described

main/src/main/scala/sbt/Defaults.scala
jrudolph: A more idiomatic way would be to pass the corresponding skip key to `publishTask` directly and then call it with

```scala
    publish := publishTask(publishConfiguration, publish / skip).value,
    publishLocal := publishTask(publishLocalConfiguration, publishLocal / skip).value,
```

mdedetrich: Ah yes that would be nicer, lets see what response I get from sbt maintainer to see if I am on write track and if so I can clean it up later

mdedetrich: So I just had a quick look at this and I am not sure if this is possible without breaking binary compatibility (which I think is a concern as they appear to be public functions?)

The first issue is that if I change `publishTask` to `
```scala
def publishTask(
    config: TaskKey[PublishConfiguration],
    skipPublish: TaskKey[Boolean]
)
```
Then this will conflict with an already existing deprecated method (i.e.

```scala
@deprecated("Use variant without delivery key", "1.1.1")
def publishTask(
    config: TaskKey[PublishConfiguration],
    deliverKey: TaskKey[_],
): Initialize[Task[Unit]] =
  publishTask(config)
```

Both methods will have the same type after erasure.

Instead you could define

```scala
def publishTask(
    config: TaskKey[PublishConfiguration],
    skipPublish: Boolean
)
```

To deal with the erasure problem but on the assumption we are dealing with public functions, such a function is out of place in terms of API design. I could make this private, but I would also have to do some more additional changes to the call site since we are not dealing with a `TaskKey` anymore (naively doing `publishTask(publishConfiguration, publish / skip).value` doesn't work because the macro cannot expand the `.value` at that spot).

If I can get rid of the deprecated `publishTask` function then this change makes perfect sense. @eed3si9n What do you think?

eed3si9n: I don't think we can remove methods during 1.x series.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -318,6 +318,7 @@ object Defaults extends BuildCommon {
       useLog4J :== SysProp.useLog4J,
       watchSources :== Nil, // Although this is deprecated, it can't be removed or it breaks += for legacy builds.
       skip :== false,
+      publishLocal / skip := (publish / skip).value, // So we don't break previous behaviour
```
eed3si9n: Could you add scripted test please? It should catch that the above line won't work in a subproject unless you move it to project setting.

eed3si9n (changes requested):
Overall ok with the direction, but we would like a test.

main/src/main/scala/sbt/Defaults.scala
mdedetrich: Okay so you fine leaving the implementation it as is?

main/src/main/scala/sbt/Defaults.scala
mdedetrich: Should I place the scripted test in `sbt-app/src/sbt-test/project`?

mdedetrich: Indeed you were correct, this setting was in the wrong place (`globalSettings`) which meant that it didn't have the desired behaviour. I moved the setting to `coreDefaultSettings` (not sure if this is the correct place).

I added a scripted test in `project/skip` to verify the expected behaviour.

mdedetrich:
@eed3si9n All tests are passing now, let me know if any additional changes are needed

main/src/main/scala/sbt/Defaults.scala
eed3si9n: > Okay so you fine leaving the implementation it as is?

Yea.

eed3si9n (approved):
Thanks!

adpi2:
I think this PR introduced a bug: https://github.com/sbt/sbt/issues/7288

PR #7178, opened 2023-03-16
Fix bug of unmanagedResourceDirectories

minkyu97:
I found that sometimes files in directories added to `unmanagedResourceDirectories` are ignored.

This was because the `resourcesDirectories` are managed as a `Set` like below

``` scala
  def copyResourcesTask =
    Def.task {
      val t = classDirectory.value
      val dirs = resourceDirectories.value.toSet
      ...
```

But, you know, the default `Set` in `scala` is implemented as a `HashSet`, which does not guarantee ordering of elements.

For example, When I added my config directory `src/main/resources/conf` to `unmanagedResourceDirectories` in order to include a file called `src/main/resources/conf/my-conf.json` as a resource, but my config file was compiled as `target/scala/classes/conf/my-conf.json`, not `target/scala/classes/my-conf.json`.

When I printed out the variable `dirs` mentioned above, I got the following result. (For convenience, represent the project path as `~`)

``` scala
Set(~/src/main/resources/conf, ~/src/main/resources)
```

When I improved it using `SortedSet` and printed out the variable `dirs`, It looks fine! 

``` scala
TreeSet(~/src/main/resources,~/src/main/resources/conf)
```

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2535,7 +2535,7 @@ object Defaults extends BuildCommon {
   def copyResourcesTask =
     Def.task {
       val t = classDirectory.value
-      val dirs = resourceDirectories.value.toSet
+      val dirs = collection.SortedSet(resourceDirectories.value: _*)
       val s = streams.value
       val syncDir = crossTarget.value / (prefix(configuration.value.name) + "sync")
       val factory = CacheStoreFactory(syncDir)
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2542,7 +2542,8 @@ object Defaults extends BuildCommon {
       val cacheStore = factory.make("copy-resource")
       val converter = fileConverter.value
       val flt: File => Option[File] = flat(t)
-      val transform: File => Option[File] = (f: File) => rebase(dirs, t)(f).orElse(flt(f))
+      val transform: File => Option[File] =
+        (f: File) => rebase(resourceDirectories.value.sorted, t)(f).orElse(flt(f))
       val mappings: Seq[(File, File)] = resources.value.flatMap {
         case r if !dirs(r) => transform(r).map(r -> _)
         case _             => None
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2535,7 +2535,7 @@ object Defaults extends BuildCommon {
   def copyResourcesTask =
     Def.task {
       val t = classDirectory.value
-      val dirs = collection.SortedSet(resourceDirectories.value: _*)
+      val dirs = resourceDirectories.value.toSet
       val s = streams.value
       val syncDir = crossTarget.value / (prefix(configuration.value.name) + "sync")
       val factory = CacheStoreFactory(syncDir)


lightbend-cla-validator:
Hi @minkyu97,

minkyu97:
I just signed the CLA as individual

eed3si9n:
> ```scala
> TreeSet(~/src/main/resources,~/src/main/resources/conf)
> ```

Thanks for the report and sending a PR. Are you saying that sbt currently does not support adding overlapping unmanagedResourceDirectories like `src/main/resources` and its **subdirectory** `src/main/resources/conf`? I think that's an intended design, and I am not sure if we should make the set sorted to support this.

minkyu97:
@eed3si9n Thanks for the comment.
You're right, the problem occurs when `unmanagedResourceDirectories` is a subdirectory of other resources.
Although this may not be mainstream, I think it's a workable way to do it and doesn't impact existing logic at all. Are there any issues you're worried about when sorting it?

eed3si9n:
> Are there any issues you're worried about when sorting it?

I think so. In general having static typing in the build definition allows sbt, builds, and plugin ecosystem to evolve at different pace. I am particularly concerned about making implementation-level changes that would make semantic difference in exactly the **opposite** of what the type promises in an observable way. It might work for some time, but it could break again later. I would recommend creating a different directory with your conf files.

eed3si9n:
> Are there any issues you're worried about when sorting it?

I think so. In general having static typing in the build definition allows sbt, builds, and plugin ecosystem to evolve at different pace. I am particularly concerned about making implementation-level changes that would make semantic difference in exactly the **opposite** of what the type promises in an observable way. It might work for some time, but it could break again later. I would recommend creating a different directory with your conf files.

minkyu97:
If I move the unmanaged resources directories out of `resources`, the same problem will occur if one unmanaged resources directory becomes a superset of another unmanaged resources directory.

For example, the codes below might cause same problem.

``` scala
Compile / unmanagedResourceDirectories += baseDirectory.value / "src/main/unmanaged_resources"
Compile / unmanagedResourceDirectories += baseDirectory.value / "src/main/unmanaged_resources/hadoop_conf"
```

However, I also agree with you that having static typing at build time can cause problems. After some thought I've thought that the underlying problem is not that the `dirs` variable is unordered, but that unordered `oldBases` are being passed to the `rebase` function. So, How about changing the [`sbt.io.rebase`](https://github.com/sbt/io/blob/f0351465de8295b7c3fa4820549fb3b1f15dc05f/io/src/main/scala/sbt/io/PathMapper.scala#L84-L85) function to something like below? Do you still think it's not worth it?

``` scala
  def rebase(oldBases: Iterable[File], newBase: File, zero: FileMap = transparent): FileMap =
    fold(zero, oldBases.toList.sorted)(old => rebase(old, newBase))
```

minkyu97:
@eed3si9n Hi, there!
I'm sorry for bothering you again 😥
Can you let me know if there is a possibility of causing issues at build time by changing [copyResourcesTask](https://github.com/sbt/sbt/blob/4a470dcd175404377afebba9987519c8ad2c25bf/main/src/main/scala/sbt/Defaults.scala#L2528-L2546) as follows?
``` scala
  def copyResourcesTask =
    Def.task {
      val t = classDirectory.value
      val dirs = resourceDirectories.value.toSet
      val sortedDirs = resourceDirectories.value.sorted
      val s = streams.value
      val syncDir = crossTarget.value / (prefix(configuration.value.name) + "sync")
      val factory = CacheStoreFactory(syncDir)
      val cacheStore = factory.make("copy-resource")
      val flt: File => Option[File] = flat(t)
      val transform: File => Option[File] = (f: File) => rebase(sortedDirs, t)(f).orElse(flt(f))
      val mappings: Seq[(File, File)] = resources.value.flatMap {
        case r if !dirs(r) => transform(r).map(r -> _)
        case _             => None
      }
      s.log.debug("Copy resource mappings: " + mappings.mkString("\n\t", "\n\t", ""))
      Sync.sync(cacheStore, fileConverter = converter)(mappings)
      mappings
    }
```
I just added a variable, `sortedDirs`, and only used it in the `rebase` function instead of using static typing.
The directories within `sortedDirs` are always sorted and `rebase` function will traverse them in reverse order (because it is `@tailrec` function), so the arbitrary path `X` will always be processed later than `X/Y` which is also an arbitrary subpath of `X`.
Therefore, the problem will be resolved.

minkyu97:
@eed3si9n Hi, there!
I'm sorry for bothering you again 😥
Can you let me know if there is a possibility of causing issues at build time by changing [copyResourcesTask](https://github.com/sbt/sbt/blob/4a470dcd175404377afebba9987519c8ad2c25bf/main/src/main/scala/sbt/Defaults.scala#L2528-L2546) as follows?
``` scala
  def copyResourcesTask =
    Def.task {
      val t = classDirectory.value
      val dirs = resourceDirectories.value.toSet
      val sortedDirs = resourceDirectories.value.sorted
      val s = streams.value
      val syncDir = crossTarget.value / (prefix(configuration.value.name) + "sync")
      val factory = CacheStoreFactory(syncDir)
      val cacheStore = factory.make("copy-resource")
      val flt: File => Option[File] = flat(t)
      val transform: File => Option[File] = (f: File) => rebase(sortedDirs, t)(f).orElse(flt(f))
      val mappings: Seq[(File, File)] = resources.value.flatMap {
        case r if !dirs(r) => transform(r).map(r -> _)
        case _             => None
      }
      s.log.debug("Copy resource mappings: " + mappings.mkString("\n\t", "\n\t", ""))
      Sync.sync(cacheStore, fileConverter = converter)(mappings)
      mappings
    }
```
I just added a variable, `sortedDirs`, and only used it in the `rebase` function instead of using static typing.
The directories within `sortedDirs` are always sorted and `rebase` function will traverse them in reverse order (because it is `@tailrec` function), so the arbitrary path `X` will always be processed later than `X/Y` which is also an arbitrary subpath of `X`.
Therefore, the problem will be resolved.

eed3si9n:
@minkyu97 Sorry about the delayed response. I'd be fine with going forward with the idea.

minkyu97:
@eed3si9n I'm really sorry too.. I missed your comment for a long time because of my work. 
I pushed a commit applying the above idea. I'd appreciate it if you could take a look at it and you can close the PR if you don't think it needed to be applied.

PR #7201, opened 2023-04-05
[1.9.x] Add `sbtPluginPublishLegacyMavenStyle` to publish to Artifactory

adpi2:
Solves https://github.com/sbt/sbt/issues/3410#issuecomment-1491092368

Artifactory has a feature that rejects any non-constitent POM, including legacy sbt plugins' POM. By turning `sbtPluginPublishLegacyMavenStyle` off, we can publish only the valid artifacts and not the legacy ones.

Does anyone think of a better name instead of `sbtPluginPublishLegacyMavenStyle`?

### DIFF
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -491,6 +491,7 @@ object Keys {
   val makeIvyXmlLocalConfiguration = taskKey[PublishConfiguration]("Configuration for generating ivy.xml.").withRank(DSetting)
   val packagedArtifacts = taskKey[Map[Artifact, File]]("Packages all artifacts for publishing and maps the Artifact definition to the generated file.").withRank(CTask)
   val publishMavenStyle = settingKey[Boolean]("Configures whether to generate and publish a pom (true) or Ivy file (false).").withRank(BSetting)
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/Keys.scala
```diff
@@ -491,6 +491,7 @@ object Keys {
   val makeIvyXmlLocalConfiguration = taskKey[PublishConfiguration]("Configuration for generating ivy.xml.").withRank(DSetting)
   val packagedArtifacts = taskKey[Map[Artifact, File]]("Packages all artifacts for publishing and maps the Artifact definition to the generated file.").withRank(CTask)
   val publishMavenStyle = settingKey[Boolean]("Configures whether to generate and publish a pom (true) or Ivy file (false).").withRank(BSetting)
+  val sbtPluginPublishLegacyMavenStyle = settingKey[Boolean]("Configuration for generating the legacy pom of sbt plugins, to publish to Maven").withRank(CSetting)
```
eed3si9n, line 494: I think the name is fine :)

eed3si9n (approved):

xuwei-k:
- https://github.com/sbt/sbt/discussions/7277 🤔

PR #7215, opened 2023-04-18
Expose RunProfiler to the build

dragos:
Expose what the incremental compiler is doing behind the scenes. The RunProfiler interface has been part of Zinc for a while, but this allows the build itself, or an Sbt plugin, to hook their own implementation.

We expose a list of such listeners to avoid plugins stepping on each other and replacing an existing listener.

NOTE: this is broken until Sbt upgrades to a new release containing https://github.com/sbt/zinc/pull/1181


lightbend-cla-validator:
Hi @dragos,

dragos:
If the PR is approved, this might be safe and desirable to backport to 1.8.x

main/src/main/scala/sbt/Keys.scala
```diff
@@ -248,7 +248,7 @@ object Keys {
   val copyResources = taskKey[Seq[(File, File)]]("Copies resources to the output directory.").withRank(AMinusTask)
   val aggregate = settingKey[Boolean]("Configures task aggregation.").withRank(BMinusSetting)
   val sourcePositionMappers = taskKey[Seq[xsbti.Position => Option[xsbti.Position]]]("Maps positions in generated source files to the original source it was generated from").withRank(DTask)
-  private[sbt] val externalHooks = taskKey[ExternalHooks]("The external hooks used by zinc.")
+  val externalHooks = taskKey[ExternalHooks]("The external hooks used by zinc.")
```
eed3si9n: Keeping this private allows us to not worry about binary compatibility. I wonder if it would be better to make a setting specifically for InvalidationProfiler, or you could make a class in `sbt` package as a hack.

dragos: This is a good point, but I don't think it's a real issue. [ExternalHooks](https://github.com/sbt/zinc/blob/fb8e03e87870429fbf518cf4466a3b1d7cebd418/internal/compiler-interface/src/main/java/xsbti/compile/ExternalHooks.java#L24) and its members are part of `xsbti.compile` in zinc's compiler-interface, and already checked by MiMa. They should have the same binary compatibility guarantees as `CompileAnalysis`, which is already part of the Sbt interface.

eed3si9n: Like the name suggests, I consider the compiler interface as a contract between Zinc and compilers, not sbt and plugin authors. It's true that some of them are part of task key, I think of them more as incidental leaks. I am not completely against opening this up, since it's not hard to access it, but I'll note here that I'll likely ignore bug reports stemming from people customizing this. We should probably change the key name to something like `experimentalZincHooks`.

dragos: I'm not sure who's using it, and why it was added in the first place, do you happen to know? Sbt itself doesn't use it, and packaging things inside the `sbt` package seems very brittle, and not sure how well it works in Java 17 and above.

I'm fine to leave this private and add another key for invalidationProfiler, and thread that through to the existing `externalHooks` if you're more comfortable this way (I'd likely need a working zinc release with sbt/zinc#1181).

eed3si9n: I think it started its life in https://github.com/sbt/zinc/pull/158, for IDE integration (in-house IntelliJ plugin?) then now it's used by in-house build tool.

dragos: I was also wondering on the Sbt side, since the Zinc interface is public.

Either way, please let me know how would you want me to proceed to unblock this?

eed3si9n: Yea. If you could add a new key for `InvalidationProfiler` that'd be great.

eed3si9n (approved):
Thanks @dragos!

dragos:
I just pushed a little update, modifying the key description to be more readable and similar to the other standard key descriptions.

PR #7241, opened 2023-05-08
fix: deprecate mispelled diagnosticRelatedInforamation

ckipp01:
This also adds in a new one.

Relates to https://github.com/sbt/sbt/discussions/7063


ckipp01:
Not sure if we can slip this into the 1.9.0 release or not, but if not, let me know and I can target `develop` instead.

eed3si9n (approved):
Sounds good

PR #7242, opened 2023-05-08
feat: add in `actions()` to `Problem`

ckipp01:
This adds a new field into `xsbti.Problem` allowing for the compiler to forward "actions" that can address diagnostics. The idea largely mimics a very minimal `CodeAction` that can be found in the [LSP Spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeAction) in order to ensure it will work with a variety of difference clients that use LSP, and those that don't. In the future the `WorkspaceEdit` that was created here could also be expanded to handle more advanced changes, aka resource operations, like creating/moving/deleting files. For now we only focus on a small subset of these features.

This relates to the discussion in https://contributors.scala-lang.org/t/roadmap-for-actionable-diagnostics/6172

### DIFF
--- a/internal/util-interface/src/main/java/xsbti/FileChanges.java
+++ b/internal/util-interface/src/main/java/xsbti/FileChanges.java
@@ -1,0 +1,21 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package xsbti;
+
+import java.net.URI;
+import java.util.List;
+
+/** A collection of TextEdits that belong to a given URI. */
+public interface FileChanges {
+
+  /** The URI that the edits belong to. */
+  URI uri();
+
+  /** The edits belonging to the URI. */
+  List<TextEdit> edits();
+}
--- a/internal/util-interface/src/main/java/xsbti/TextEdit.java
+++ b/internal/util-interface/src/main/java/xsbti/TextEdit.java
@@ -1,0 +1,25 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package xsbti;
+
+/**
+ * A representation of the `TextEdit` found in the LSP protocol.
+ *
+ * <p>NOTE: That instead of a `Range` we use the internal [[xsbti.Position]].
+ *
+ * @see <a
+ *     href="https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textEdit">`TextEdit`</a>
+ */
+public interface TextEdit {
+
+  /** The position this edit will be applied to. */
+  Position position();
+
+  /** The next text that will be inserted into the given [[TextEdit.position]]. */
+  String newText();
+}
--- a/internal/util-interface/src/main/java/xsbti/WorkspaceEdit.java
+++ b/internal/util-interface/src/main/java/xsbti/WorkspaceEdit.java
@@ -1,0 +1,27 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package xsbti;
+
+import java.net.URI;
+import java.util.List;
+
+/**
+ * A minimal representatin of the `WorkspaceEdit` found in the LSP protocol.
+ *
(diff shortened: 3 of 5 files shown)


internal/util-interface/src/main/java/xsbti/ScalaAction.java
```diff
@@ -0,0 +1,32 @@
...
+ * A ScalaAction is very miminal representation of a `CodeAction` in the LSP protocol.
+ *
+ * <p>However it only focuses on the actual title, description, and edit, leaving it up to the
+ * language server to communicate with the client and put together a proper codeAction in accordance
+ * to client capabilities.
+ *
+ * @see <a href=
+ *     "https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeAction">`CodeAction`</a>
+ */
+public interface ScalaAction {
```
adpi2: I wonder if this should be `Action` instead of `ScalaAction`. Who knows, maybe one day we can have actions coming from Java.

Or from a code generator: using the sourcePositionMappers, sbt would be able to transform the Scala actions into Twirl actions for instance.

internal/util-interface/src/main/java/xsbti/WorkspaceEdit.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * <p>NOTE: In the future this may be expanded to handle resource operations via `documentChanges`.
+ *
+ * @see <a href=
+ *     "https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspaceEdit">`WorkspaceEdit`</a>
+ */
+public interface WorkspaceEdit {
+
+  /** Map of the `TextEdit`s that belong to a specific URI. */
+  Map<URI, TextEdit> changes();
```
eed3si9n: We should avoid `Map` since this could introduce non-determinism when persisted into JSON, and `Problem` is a good candidate for caching.

internal/util-interface/src/main/java/xsbti/ScalaAction.java
ckipp01: Ooooo, interesting idea. Yea, I can change this to `Action`.

internal/util-interface/src/main/java/xsbti/WorkspaceEdit.java
ckipp01: Sure, would you recommend something else specifically?

eed3si9n: `java.util.List<xsbti.T2<URI, TextEdit>>`, maybe?

ckipp01: Huh, I really dislike the `get1()` and `get2()` that would need to be used for this. I instead created a `FileChanges` interface  that will hold a `URI` and the `List<TextEdits>` that belong to it. Originally this map was actually wrong as it should have been `Map<URI, List<TextEdit>>` so I'm glad we caught this 😬. But let me know if this new approach is alright.

adpi2 (approved):
LGTM

ckipp01:
@tgodzik and @smarter I met with both of you about this, so I also just want to ping you to make sure we're all on the same page with this.

internal/util-interface/src/main/java/xsbti/TextEdit.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * <p>NOTE: That instead of a `Range` we use the internal [[xsbti.Position]].
+ *
+ * @see <a
+ *     href="https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textEdit">`TextEdit`</a>
+ */
+public interface TextEdit {
+
+  /** The position this edit will be applied to. */
+  Position position();
```
tgodzik, line 21: Just to make sure, the position has a start and an end? So it's basically a range?

tgodzik (approved):
Looks good!

internal/util-interface/src/main/java/xsbti/TextEdit.java
ckipp01: Yup, its got all the goodies you'd expect:

https://github.com/sbt/sbt/blob/1.9.x/internal/util-interface/src/main/java/xsbti/Position.java

eed3si9n (approved):
Thanks @ckipp01!

internal/util-interface/src/main/java/xsbti/FileChanges.java
```diff
@@ -0,0 +1,21 @@
...
+package xsbti;
+
+import java.net.URI;
+import java.util.List;
+
+/** A collection of TextEdits that belong to a given URI. */
+public interface FileChanges {
+
+  /** The URI that the edits belong to. */
+  URI uri();
```
eed3si9n, line 17: Actually I don't think this should be URI - https://github.com/sbt/sbt/issues/7252

PR #7253, opened 2023-05-14
fix: get rid of the `FileChanges` abstraction

### LINKED ISSUE #7252: xsbti.FileChanges uses URI
## steps

Unreleased `xsbti.FileChanges` looks like

```java
/** A collection of TextEdits that belong to a given URI. */
public interface FileChanges {

  /** The URI that the edits belong to. */
  URI uri();

  /** The edits belonging to the URI. */
  List<TextEdit> edits();
}
```

## problem

At the BSP level, [SourceItem](https://build-server-protocol.github.io/docs/specification#buildtargetsources-request) is tracked as `URI`.

## expectation

We probably should use `Optional[String]` and `Optional[File]` like we do currently in Position - https://github.com/sbt/sbt/blob/v1.9.0-RC2/internal/util-interface/src/main/java/xsbti/Position.java#L25-L27:

```java
public interface Position {
....

  Optional<String> sourcePath();

  Optional<File> sourceFile();
```

## notes

Inside the compiler, source files are represented as virtual [SourceFile](https://github.com/scala/scala/blob/v2.13.10/src/reflect/scala/reflect/internal/util/SourceFile.scala)s, and currently we send information from the compiler to Zinc:
- in https://github.com/sbt/zinc/blob/v1.9.0-RC2/internal/compiler-bridge/src/main/scala/xsbt/DelegatingReporter.scala (for Scala 2.x).
- or in https://github.com/lampepfl/dotty/blob/3.3.0-RC6/sbt-bridge/src/dotty/tools/xsbt/PositionBridge.java#L83-L102 (for Scala 3.x)

using `Optional[String]` and `Optional[File]`.

At the BSP server level in https://github.com/sbt/sbt/blob/v1.9.0-RC2/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala we already know the association between source files and problems (after the compilation is done), and BSP server also knows how to convert the source file into URI that _it_ told the BSP client.

ckipp01:
I don't believe this is actually needed. We can simplify the `WorkspaceEdit` further and put together the mapping for `changes` according to the LSP spec later on down the line. This will allow us to remove the need for the `URI`, which is problematic.

closes #7252


eed3si9n (approved):
Thanks!

PR #7261, opened 2023-05-21
Deprecate itSettings

### LINKED ISSUE #7260: Also deprecate `Defaults.itSettings` for SBT 1.9.x?
Since `IntegrationTest` config is being deprecated in SBT 1.9.x maybe it makes sense to also deprecate `sbt.Default.itSettings`? Other  code that relates to integration test (such as the `IntegrationTest` core config) has already been deprecated.

mdedetrich:
Resolves: https://github.com/sbt/sbt/issues/7260

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2603,6 +2603,10 @@ object Defaults extends BuildCommon {
   lazy val testSettings: Seq[Setting[_]] = configSettings ++ testTasks
 
   @nowarn
+  @deprecated(
+    "Create a separate subproject instead of using IntegrationTest and in addition avoid using itSettings",
+    "1.9.0"
+  )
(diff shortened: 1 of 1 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2603,6 +2603,7 @@ object Defaults extends BuildCommon {
   lazy val testSettings: Seq[Setting[_]] = configSettings ++ testTasks
 
   @nowarn
+  @deprecated("Create a separate subproject instead of using IntegrationTest and in addition avoid using itSettings", "1.9.0")
```
eed3si9n: Apparently this file needs formatting.

mdedetrich: Will do it tonight when I get back to laptop

mdedetrich: Pushed and fixed.

eed3si9n (approved):

PR #7264, opened 2023-05-22
Forward ScalaDiagnostic

eed3si9n:
See also BSP spec update https://github.com/build-server-protocol/build-server-protocol/pull/527

**Problem**
Zinc added actions in Problem, but it's not yet forwarded to BSP clients.

**Solution**
As discussed in
https://contributors.scala-lang.org/t/roadmap-for-actionable-diagnostics/6172, the plan seems to be to use the `data` field with `actions` inside it, so this implements that.


protocol/src/main/contraband/bsp.contra
```diff
@@ -132,6 +138,67 @@ type Diagnostic {
...
+## the dataKind field contains "scala".
+type ScalaDiagnostic {
+  ## Actions (also known as quick fixes) that are able to either fix or address
+  ## the issue that is causing this diagnostic.
+  actions: [sbt.internal.bsp.ScalaAction]
+}
+
+## A Scala action represents a change that can be performed in code.
+## See also LSP: Code Action Request (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction).
+type ScalaAction {
```
adpi2, line 177: As pointed out by @ckipp01 in https://github.com/build-server-protocol/build-server-protocol/pull/527#pullrequestreview-1435850632 this is a duplication of `xsbti.Action`. I wonder if it would not be simpler to define the `xbsti.Action` directly as a `contra` instead. A counter argument is that maybe we don't want util-interface to depend on `sjsonnew`?

eed3si9n: util-interface and compiler-interface needs to be in Java, but I guess we can make a second subproject for sjsonnew codecs.

adpi2: Not sure if it would be better or worse though. I am fine with the PR as it is.

adpi2 (approved):

PR #7265, opened 2023-05-22
Update typelevel toolkit, configure test artifact

armanbilge:
Following suit with Scala toolkit, we now have a separate test artifact. cc @toniogela


adpi2 (approved):
LGTM

PR #7295, opened 2023-06-12
Fix #7288: `publish` should not fail if `publish / skip := true`

### LINKED ISSUE #7288: Publish with SBT 1.9.0 always requires publishTo defined even with skip/publish set to true
## steps

- this bug first appeared with SBT 1.9.0, worked fine in older versions
- to reproduce the bug run `publish` on the following project

```
val root = (project in file(".")).aggregate(api,service)
lazy val api = project.settings(publishTo := Some(Resolver.mavenLocal))
lazy val service = project.settings(skip/publish := true)
```

## problem

I am getting

```
[error] stack trace is suppressed; run last service / publishConfiguration for the full output
[error] stack trace is suppressed; run last publishConfiguration for the full output
[error] (service / publishConfiguration) Repository for publishing is not specified.
[error] (publishConfiguration) Repository for publishing is not specified.
[error] Total time: 0 s, completed Jun 5, 2023, 11:01:56 PM
```

SBT forces me to set `publishTo` to `Some` even for modules which are never supposed to be published

## expectation

`publishTo` set to `None` (which is a default value) should be fine in modules with `skip/publish := true`

adpi2:
Also implement the same logic for `publishM2`.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2622,7 +2622,8 @@ object Defaults extends BuildCommon {
         if (turbo.value) ClassLoaderLayeringStrategy.AllLibraryJars
         else ClassLoaderLayeringStrategy.ScalaLibrary
       },
-      publishLocal / skip := (publish / skip).value, // So we don't break previous behaviour
+      publishLocal / skip := (publish / skip).value,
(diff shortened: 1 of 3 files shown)


eed3si9n (approved):
Thanks @adpi2!

PR #7298, opened 2023-06-14
Fix #7297: Convert `OffsetPosition` to BSP range

### LINKED ISSUE #7297: [1.9.x] Missing range in BSP diagnostic
Since sbt 1.9.0

## steps

1. In an sbt project
```scala
package example 

class A {
  def foo: String = ""
}

class AA extends A {
  val foo: String = ""
}
```

2. Import the project in Metals using sbt as the build server

## problem

The missing `override` error is reported at the very top of the file (range 0, 0).

```scala
package example
// ^^^^
// overriding method foo in class A of type => String;
//   value foo needs `override' modifier
```

## expectation

The error should be reported on `def foo`:

```scala
class AA extends A {
  def foo: String = ""
// ^^^^^^^^^^^^^^^^^
// overriding method foo in class A of type => String;
//  method foo needs `override' modifiersbt
}
```

## notes

The bug was introduced in sbt 1.9.0, it works as expected in sbt 1.8.2

adpi2:
Fixes #7297 

An `OffsetPosition` contains a `line` and a `pointer` in the line. We use them to compute the `startPos` of the BSP range.

```scala
package example

class A {
  def foo: String = ""
}

class AA extends A {
  val foo: String = ""
//    ^^^
// overriding method foo in class A of type => String;
//  value foo needs `override' modifier
}
```

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
@@ -185,11 +185,15 @@ final class BuildServerReporterImpl(
     val startColumnOpt = pos.startColumn.toOption.map(_.toLong)
     val endLineOpt = pos.endLine.toOption.map(_.toLong - 1)
     val endColumnOpt = pos.endColumn.toOption.map(_.toLong)
+    val lineOpt = pos.line.toOption.map(_.toLong - 1)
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks @adpi2!

Issue #7309, opened 2023-06-20
Add a way to cleanly override sbt plugin versions


henryxparker:
Currently there is no simple command to override a plugin version. 
If plugin A depends on plugin B, plugin B is imported automatically. 
If the project already imported B but at an earlier version it will be evicted to the newer version.
Now lets say that A works fine with both versions of B, but the newer version of B breaks the build.
Currently, you must do something like this in the `plugins.sbt`
```scala
dependencyOverrides += {
  val sbtV = (pluginCrossBuild / sbtBinaryVersion).value
  val scalaV = (update / scalaBinaryVersion).value
  sbt.Defaults.sbtPluginExtra("com.twitter" % "scrooge-sbt-plugin" % "21.4.0" , sbtV, scalaV)
}
```
This is basically pulled directly from `addSbtPlugin`, with the only difference being that it goes into `dependencyOverrides`
```scala
  def addSbtPlugin(dependency: ModuleID): Setting[Seq[ModuleID]] =
    libraryDependencies += {
      val sbtV = (pluginCrossBuild / sbtBinaryVersion).value
      val scalaV = (update / scalaBinaryVersion).value
      sbtPluginExtra(dependency, sbtV, scalaV)
    }
```
So I figured it might be nice to add another set of methods akin to `addSbtDependency`
I figured `overrideSbtPluginDependency` or simply `overrideSbtPlugin`, but I'm open to suggestions

eed3si9n:
See https://github.com/sbt/sbt/pull/7310#issuecomment-1788188284

PR #7310, opened 2023-06-20
Add methods for overriding sbt dependencies

henryxparker:
See #7309

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -4418,6 +4418,38 @@ trait BuildExtra extends BuildCommon with DefExtra {
       sbtPluginExtra(dependency, sbtV, scalaV)
     }
 
+  /**
+   * Overrides a dependency as an sbt plugin for the specific sbt version sbtVersion and Scala version scalaVersion.
+   * Typically, use the default values for these versions instead of specifying them explicitly.
+   */
+  def overrideSbtPlugin(
+      dependency: ModuleID,
+      sbtVersion: String,
+      scalaVersion: String
+  ): Setting[Seq[ModuleID]] =
+    dependencyOverrides += sbtPluginExtra(dependency, sbtVersion, scalaVersion)
+
+  /**
+   * Overrides dependency as an sbt plugin for the specific sbt version sbtVersion.
+   * Typically, use the default value for this version instead of specifying it explicitly
+   */
+  def overrideSbtPlugin(dependency: ModuleID, sbtVersion: String): Setting[Seq[ModuleID]] =
+    dependencyOverrides += {
+      val scalaV = (update / scalaBinaryVersion).value
+      sbtPluginExtra(dependency, sbtVersion, scalaV)
+    }
+
+  /**
+   * Overrides dependency as an sbt plugin for the sbt and Scala versions configured by
+   * sbtBinaryVersion and scalaBinaryVersion scoped to update.
+   */
+  def overrideSbtPlugin(dependency: ModuleID): Setting[Seq[ModuleID]] =
+    dependencyOverrides += {
+      val sbtV = (pluginCrossBuild / sbtBinaryVersion).value
+      val scalaV = (update / scalaBinaryVersion).value
+      sbtPluginExtra(dependency, sbtV, scalaV)
+    }
+
   /** Transforms `dependency` to be in the auto-compiler plugin configuration. */
   def compilerPlugin(dependency: ModuleID): ModuleID =
     dependency.withConfigurations(Some("plugin->default(compile)"))
--- a/notes/1.9.0/add-sbt-plugin-override-method.md
+++ b/notes/1.9.0/add-sbt-plugin-override-method.md
@@ -1,0 +1,4 @@
+[@henryxparker]: https://github.com/henryxparker
+
+### Improvements
+- Add methods for overriding sbt plugin dependencies. See [issue #7309](https://github.com/sbt/sbt/issues/7309)
--- a/sbt-app/src/sbt-test/plugins/plugin-dependency-override/baseplugin/src/main/scala/sbt/test/BasePlugin.scala
+++ b/sbt-app/src/sbt-test/plugins/plugin-dependency-override/baseplugin/src/main/scala/sbt/test/BasePlugin.scala
@@ -1,0 +1,13 @@
+package sbt.test
+
+import sbt.Keys.version
+import sbt.{AutoPlugin, Compile, Setting, ThisProject, inConfig, taskKey}
+
+class BasePlugin extends AutoPlugin {
+  override def requires = sbt.plugins.JvmPlugin
+  override def trigger = noTrigger
+
+  object autoImport {
+    val pluginVersion = taskKey[String]("Test key to ensure dependency is overridden")
+  }
+}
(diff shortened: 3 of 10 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -4403,6 +4403,39 @@ trait BuildExtra extends BuildCommon with DefExtra {
...
+    dependencyOverrides += {
+      val scalaV = (update / scalaBinaryVersion).value
+      sbtPluginExtra(dependency, sbtVersion, scalaV)
+    }
+
+  /**
+   * Overrides dependency as an sbt plugin for the sbt and Scala versions configured by
+   * sbtBinaryVersion and scalaBinaryVersion scoped to update.
+   */
+  def overrideSbtPlugin(dependency: ModuleID): Setting[Seq[ModuleID]] =
```
eed3si9n, line 4446: Thanks for the contribution. Could you add a scripted test in https://github.com/sbt/sbt/tree/1.9.x/sbt-app/src/sbt-test/project or https://github.com/sbt/sbt/tree/1.9.x/sbt-app/src/sbt-test/plugins to demonstrate that it works please?

henryxparker: Since this has to do with resolving dependencies, I will need to make reference to outside plugins. Is that okay? Any preference on what I choose?
I could also attempt to publish a fabricated plugin to local ivy by using some sbt-tests just to publish, but that means that test needs to happen prior to the one where they're imported, or it needs to all be in subprojects of the test, but that seems odd.

eed3si9n: Yea. That's fine. That's why I listed https://github.com/sbt/sbt/tree/1.9.x/sbt-app/src/sbt-test/plugins since those tests reference existing plugins.

henryxparker: It took me much longer to make the tests than the change, but I suppose that's unavoidable here.
Feel free to critique my work, I just went with the first way that I got it working

sbt-app/src/sbt-test/plugins/plugin-dependency-override-test/test
```diff
@@ -0,0 +1 @@
+> check
```
eed3si9n: We run scripted tests are batched, and each test is meant to be self-contained.
I think the prep test can be merged into here by `copy-file` (https://www.scala-sbt.org/1.x/docs/Testing-sbt-plugins.html) that creates `project/plugins.sbt`.

sbt-app/src/sbt-test/plugins/plugin-dependency-override-setup/build.sbt
```diff
@@ -0,0 +1,34 @@
+lazy val plugin = (project in file("baseplugin"))
+  .settings(
+    version := "0.1.0",
+    name:= "base-plugin",
+    organization := "sbt.test"
```
eed3si9n: Could you use `com.example` for this purpose?

eed3si9n: Also I am not 100% sure if this also works in `project/plugins.sbt` for resolution side, but I think

```scala
ivyPaths := IvyPaths((baseDirectory in LocalRootProject).value, Some((target in LocalRootProject).value / "ivy-cache"))
```

will change the destination of `publishLocal` to a directory that will be cleaned up.

henryxparker: Made the `com.example` change.
~~I tried using the ivyPaths one, but it seems that it doesn't affect the publishing location. I put it in as a project setting, but that changed it to the local project's target directory which I would then need to figure out how to access to import from the other project. I'm open to the idea, but I'm also happy with where it is now.~~
Just saw other comment, Ill look into that

sbt-app/src/sbt-test/plugins/plugin-dependency-override-test/test
henryxparker: could you expand on that?
I get that you want it to all be in one test directory, I'm just unsure what your suggested solution is.
Are you saying: 

1. all the subprojects from the setup build should be included in the test build base directory
2. they should be included in the project directory, or
3. they should be included in the plugins.sbt

If its any of those I don't see how it would come together. 
Are you saying that I should overwrite the plugins.sbt once the plugins have been published so that the test can use it to import them? Is that even possible?

eed3si9n: > Are you saying that I should overwrite the plugins.sbt once the plugins have been published so that the test can use it to import them? Is that even possible?

Yea, so I guess what I was thinking is something like:

```scala
> plugin/publishLocal
> plugin1/publishLocal
> plugin2/publishLocal
> dependentplugin/publishLocal

$ copy-file changes/build.sbt build.sbt
$ copy-file changes/plugins.sbt project/plugins.sbt

> reload
> check
```

henryxparker: At long last this is done

henryxparker:
~~Should I rebase on top of other commits? There are some failing tests but looking at them I don't see why they would be with my change~~

Edit: I'm investigating using java 17

Edit2: Investigation done, locally test succeeds even on same java version as the previous failed test. I do get other tests failing locally, but they again seem unrelated so I think we're good to build again.

henryxparker:
I dont know why the temurin 3 test fails. I have run it on my local machine and I get no such errors. I will need some help/guidance replicating if you want me to fix it.

eed3si9n:
"test (ubuntu-latest, 17, temurin, 3)" I think is a synthetic name given by GitHub Actions for the job with those values. Most important value here is `3`, which is used to select jobtype that runs the newly added scripted test:

https://github.com/sbt/sbt/blob/e3b7870b2d98e26e38af82f5dd5e6b2b3f77b219/.github/workflows/ci.yml#L130-L134


```scala
Error:  /tmp/sbt_6dfb816b/build.sbt:8: error: not found: value pluginVersion
Error:        val v = (Compile / pluginVersion).value
Error:                           ^
Error:  sbt.compiler.EvalException: Type error in expression
[info] [error] sbt.compiler.EvalException: Type error in expression
[info] [error] Use 'last' for the full log.
[info] [warn] Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)
[info] [info] shutting down sbt server
[info] 
Error:  x plugins/plugin-dependency-override 
```

so it looks like it failed to `reload`?

henryxparker:
I changed it to reload the plugins first, I still couldn't replicate it but I think this might fix it

henryxparker:
😭

henryxparker:
I am currently not working as a scala developer anymore, so I do not have reason to continue to push on this pull request. I will leave it open in case someone wants to fix the build errors and get this feature added

eed3si9n:
I guess if someone needs this feature, they could start by copying this line to their `project/plugins.sbt`:

```scala
  def overrideSbtPlugin(dependency: ModuleID): Setting[Seq[ModuleID]] =
    dependencyOverrides += {
      val sbtV = (pluginCrossBuild / sbtBinaryVersion).value
      val scalaV = (update / scalaBinaryVersion).value
      sbtPluginExtra(dependency, sbtV, scalaV)
    }
```

PR #7317, opened 2023-06-24
some trivial refactoring

xuwei-k:
- remove unused type params
- use `withFilter` if possible
- use `collectFirst` instead of `collect` and `headOption`
- use `length` instead of `size` if `Array` or `String`
- use `foreach` instead of `map`


eed3si9n (approved):
Thanks!

PR #7328, opened 2023-07-09
Let ++ fall back to a bincompat Scala version

### LINKED ISSUE #7327: cross-build no longer works when `CrossVersion.full` projects depend on `CrossVersion.binary` ones
## steps

Check out scalameta/scalameta#3212 and build locally.

## problem

In sbt 1.6.2, having a `CrossVersion.full` project with several `crossScalaVersions` (say, 2.13.1 to 2.13.11) depend on a `CrossVersion.binary` project with only 2.13.11 used to work, and in more recent versions (possibly after #6946) the build fails.

Local build for me failed with `Modules were resolved with conflicting cross-version suffixes`, while CI build failed with `Error downloading org.scalameta:scalameta_2.12:4.8.2+9-291843d8-SNAPSHOT` (by trying to download the dependency project rather than building it locally).

## expectation

- If cross-build dependencies between `full` and `binary` projects are not allowed
  - the error message should stipulate that explicitly
  - https://www.scala-sbt.org/1.x/docs/Cross-Build.html should mention this as well, possibly with suggestions on how to deal with this issue
- if such a dependency is to be allowed again
  - either #6946 should be reverted
  - or `project.dependsOn` could be augmented with some new parameters to specify explicit fallback versions of dependency
  - also, ideally, when multiple `full` projects depend on a single `binary` project, the `binary` project wouldn't be built multiple times

## notes

Also see comment https://github.com/sbt/sbt/pull/6946#issuecomment-1627716934.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7327

**Problem**
In builds with mixed Scala patch versions (like scalameta), it's possible for a core subproject to be set to the lastest 2.12.x, but the compiler plugin component is cross published to 2.12.0 etc. `++ 2.12.0` in this case does not work since sbt 1.7.x onwards requires the queried Scala version to be listed in `crossScalaVersions`.

**Solution**
This implements sbt 1.6.x-like fallback mechanism, but instead of using the queried version (e.g. 2.12.0) it will set the Scala version to one of listed versions that is binary compatible.


kitbellew (approved):
not an expert on this codebase, of course... looks convincing. cc @tgodzik

tgodzik:
Thanks @eed3si9n ! I haven't figured out what was the problem with scalameta, but this looks very plausible. Will test it out!

tgodzik:
It's already all fixed and tested 😅 Thank you again!

PR #7350, opened 2023-08-07
Add a new CommandProgress API.

dragos:
Link to discussion thread: https://github.com/sbt/sbt/discussions/7351

# Motivation

Sbt defines the [ExecuteProgress](https://github.com/sbt/sbt/blob/72bfb3f45ab346ddf3558bc23853dfbb9392cc55/tasks/src/main/scala/sbt/ExecuteProgress.scala#L19) interface that offers fine-grained observability into task execution. However, users interact with sbt via commands, which in the most common case trigger one or more root tasks to execute. As an implementer of ExecuteProgress, it is currently impossible to infer what command is executing or if the current command was canceled by the user, or even if a command failed to parse. Moreover, a single command may trigger several “lifecycles” of execute progress (delimited by initial() and stop() calls), which makes it even harder to delimit logical workloads and correctly attribute task execution to user-facing actions. Examples are cross-compilation and sequential composition via semicolons (i.e. compile;test;publish).

Imagine a developer productivity team that would like to collect data about successful and failed builds across an organization. This data can then be used to diagnose errors (“20% of builds on the CI failed with the same error”, etc). Correctly determining the start, end, and failure status of each command is fundamental for such a service.

This proposal stems from our experience implementing Build Scans® for sbt, released as part of Gradle Enterprise (available for free at [https://scans.gradle.com](https://scans.gradle.com/#sbt)).

## Goals

- Allow sbt plugins to receive notifications of what command starts executing and when it finishes
  - Command status should correctly reflect success, cancellation, and error (including parse error, if the command name is mistyped)
- Allow access to sbt State at the beginning and the end of each command
- The mechanism should be both source compatible (not break existing build definitions) and binary compatible (not break existing plugins).

## Non-goals
The proposal shouldn’t change how tasks and commands are executed nor the order and content of existing events.


main-command/src/main/scala/sbt/Command.scala
```diff
@@ -184,12 +184,15 @@ object Command {
             }
     )
 
-  def process(command: String, state: State): State = {
+  // overload instead of default parameter to keep binary compatibility
+  def process(command: String, state: State): State = process(command, state, _ => ())
```
adpi2, line 189: Would it make sense to add an `@deprecated` on this method?

main/src/main/scala/sbt/Keys.scala
```diff
@@ -600,6 +600,7 @@ object Keys {
     def apply(progress: ExecuteProgress[Task]): TaskProgress = new TaskProgress(progress)
   }
   private[sbt] val currentTaskProgress = AttributeKey[TaskProgress]("current-task-progress")
+  private[sbt] val currentCommandProgress = AttributeKey[Seq[CommandProgress]]("current-command-progress")
```
adpi2: Since these keys are private, can we merge them in a single `currentCommandProgress: AttributeKey[Seq[CommandProgress]]`?

adpi2: It seems that `CommandProgress` is a `TaskProgress` with added `beforeCommand` and `addedCommand`. I wonder if it would make sense to merge the related keys somehow.

We need to keep `progressReports` for compatibility, but we can do:
```
commandProgress := progressReports.value.map(wrapInCommandProgress)
```

And then internally we would only deal with `currentCommandProgress` and `commandProgress`.

main/src/main/scala/sbt/CommandProgress.scala
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package sbt
+
+/**
+ * Tracks command execution progress. In addition to ExecuteProgress, this interface
+ * adds command start and end events, and gives access to the sbt.State at the beginning
+ * and end of each command.
+ */
+trait CommandProgress extends ExecuteProgress[Task] {
```
eed3si9n: Along the same line as what Adrien said, I wonder if this should be called ExecuteProgrogress2. I did something similar recently on the Zinc side and added an interface called AnalysisCallback2.

main/src/main/scala/sbt/Keys.scala
dragos: Yes, we can keep only one and decorate the old interface. Should we deprecate it?

main-command/src/main/scala/sbt/Command.scala
dragos: I'll add `@deprecated`.

dragos:
Please take another look

main/src/main/scala/sbt/ExecuteProgress2.scala
```diff
@@ -0,0 +1,79 @@
...
+
+package sbt
+import sbt.internal.util.RMap
+
+/**
+ * Tracks command execution progress. In addition to ExecuteProgress, this interface
+ * adds command start and end events, and gives access to the sbt.State at the beginning
+ * and end of each command.
+ */
+trait ExecuteProgress2 extends ExecuteProgress[Task] {
```
eed3si9n, line 21: Since the original `ExecuteProgress` is abstracted over `Task`, it might be useful to continue with the tradition too, in case we want ExecuteProgress3 etc in the future.

eed3si9n:
Do you have a demo build that you use locally use to demonstrate this feature, if so could we add a scripted test for this?

main/src/main/scala/sbt/ExecuteProgress2.scala
dragos: My reasoning was that the type we exposed in sbt.Keys was `TaskProgress`, so the type parameters was fixed, but I see no problem to add it here.

dragos: I added an attempt to keep the type parameter in https://github.com/sbt/sbt/pull/7350/commits/18d12c2dbf61b249683711bfd5a3c8a4deac16ff. However, if the reasoning below resonates with you, I would prefer to keep the current version:

- we're hitting limitations in higher-kinded support in Scala, which prevents us to express directly something like `AttributeKey[ExecuteProgress2[Task]]`. The work around is to have another type that instantiates the higher-kinded type (probably another instance of https://github.com/scala/bug/issues/2915), so we end up still having a fixed `ExecuteProgress2[Task]` (similar to today's situation for `progressReports`)
- this abstraction only makes sense in the `tasks` subprojects. In the sbt `main` project, everything is fixed at `Task`, including `EvaluateTaskConfig`. This is part of public interfaces (`progressReports` is also concrete, same for `EvaluateTask`) and there's no way to run the eval engine on anything else.
- to work around the bug we need to introduce a concrete type and use it in `sbt.Keys` which is structurally identical to `ExecuteProgress2[Task]` but it's seen as a different type (a type alias hits the same compiler bug). I find this will be confusing to users and leaks a limitation in the type system in ways that won't give us any benefit.

Given the above, I would choose the existing implementation, but have a look for yourself and let me know what you think.

Once we are happy with the implementation, I will add a scripted test.

eed3si9n: > we're hitting limitations in higher-kinded support in Scala

> Given the above, I would choose the existing implementation, but have a look for yourself and let me know what you think.

ok. Then the existing impl sounds good to me. Thanks for digging into it!

adpi2 (commented):
This looks better to me. I am not sure about the need for a polymorphic `ExecuteProgress2`, and I would rather just keep `CommandProgress`. But I let you and @eed3si9n figure this out.

main/src/main/scala/sbt/ExecuteProgress2.scala
dragos: Great, will add a scripted test and ping you back.

dragos:
Scripted test added, PTAL

eed3si9n (approved):
Thanks!

main-command/src/main/scala/sbt/Command.scala
adpi2: Still no `@deprecated`?

adpi2 (approved):

main-command/src/main/scala/sbt/Command.scala
dragos: Done.

dragos:
Is there anything else I can do to move this forward?

adpi2:
> Is there anything else I can do to move this forward?

I think the next step is to create a 1.10.x branch and to redirect and merge this PR to it. I am a bit busy currently with the preparation of Scala Days but I will do it when coming back.

main/src/main/scala/sbt/ExecuteProgress2.scala
```diff
@@ -0,0 +1,79 @@
...
+   */
+  def beforeCommand(cmd: String, state: State): Unit
+
+  /**
+   * Called after a command finished execution.
+   *
+   * @param cmd    The command string.
+   * @param result Left in case of an error. If the command cannot be parsed, it will be
+   *               signalled as a ParseException with a detailed message. If the command
+   *               was cancelled by the user, as sbt.Cancelled.
```
Duhemm: The comment doesn't describe the result in case of a successful command execution. Looking at the implementation, my understanding is that this is the state after the command has been executed. Maybe it's worth mentioning for the sake of clarity?

Duhemm (approved):
Since the new API exposes commands and tasks, should the old methods (eg. `beforeWork`, `afterCompleted` etc.) be deprecated and renamed to mention that they relate to tasks only (ie. `beforeWork` -> `beforeTaskWork`, `afterAllCompleted` -> `afterAllTasksCompleted`, etc.)?

I think it would be good to document roughly the order in which the methods will be called. For instance, it's not entirely clear to me what will be called first between `beforeCommand` and `initial` or between `afterAllCompleted`, `stop` and `afterCommand` (I guess `afterCommand` is last, but it would be good to have that fully specified?)

adpi2:
After discussion with @eed3si9n and given what happened to sbt 1.9.5, we will avoid patching 1.9.x with any new feature. So I moved the target branch from 1.9.x to 1.10.x.

We plan to start the RC cycle of sbt 1.10.0 in about a month from now.

Should we merge this PR now or wait for more documentation, as suggested by @Duhemm?

build.sbt
```diff
@@ -367,7 +370,7 @@ lazy val utilPosition = (project in file("internal") / "util-position")
   )
 
 lazy val utilLogging = (project in file("internal") / "util-logging")
-  .enablePlugins(ContrabandPlugin, JsonCodecPlugin)
+  .enablePlugins(ContrabandPlugin, JsonCodecPlugin
```
adpi2:
```suggestion
  .enablePlugins(ContrabandPlugin, JsonCodecPlugin)
```

Duhemm:
@dragos asked me to take over this PR. I've added the small documentation fixes that I wanted. I believe this should be ready to merge.

PR #7352, opened 2023-08-10
[BSP] Implement `buildTarget/javacOptions`

adpi2:


main/src/main/scala/sbt/EvaluateTask.scala
```diff
@@ -148,18 +148,20 @@ final case class PluginData(
     resolvers: Option[Vector[Resolver]],
     report: Option[UpdateReport],
     scalacOptions: Seq[String],
+    javacOptions: Seq[String],
```
eed3si9n, line 151: Adding field here would break bincompat?

adpi2: We broke this class a few times already. See for instance: https://github.com/sbt/sbt/pull/6553. So I think it is fine to break it again.

Should I create a 1.10.x branch?

adpi2:
@eed3si9n I move this PR to 1.10.x. Can I get a review/approval?

eed3si9n (approved):

PR #7354, opened 2023-08-13
add a Scala 3 seed to the sbt new menu

SethTisue:
seems like probably just an oversight that this wasn't included in the first place

### DIFF
--- a/main/src/main/scala/sbt/TemplateCommandUtil.scala
+++ b/main/src/main/scala/sbt/TemplateCommandUtil.scala
@@ -172,6 +172,7 @@ private[sbt] object TemplateCommandUtil {
     ScalaToolkitSlug -> "Scala Toolkit (beta) by Scala Center and VirtusLab",
     TypelevelToolkitSlug -> "Toolkit to start building Typelevel apps",
     SbtCrossPlatformSlug -> "A cross-JVM/JS/Native project",
+    "scala/scala3.g8" -> "Scala 3 seed template",
     "scala/scala-seed.g8" -> "Scala 2 seed template",
     "playframework/play-scala-seed.g8" -> "A Play project in Scala",
     "playframework/play-java-seed.g8" -> "A Play project in Java",


main/src/main/scala/sbt/TemplateCommandUtil.scala
```diff
@@ -172,6 +172,7 @@ private[sbt] object TemplateCommandUtil {
     ScalaToolkitSlug -> "Scala Toolkit (beta) by Scala Center and VirtusLab",
     TypelevelToolkitSlug -> "Toolkit to start building Typelevel apps",
     SbtCrossPlatformSlug -> "A cross-JVM/JS/Native project",
+    "scala/scala3.g8" -> "Scala 3 seed template",
```
eed3si9n, line 175: Won't Scala Toolkit template effectively function as Scala 3 seed?

SethTisue: Hmm. I guess you couuuuuld argue it that way, but I actually wouldn't. The menu offers a plain Scala 2 seed even though the toolkit supports both 2 and 3. And I think there are probably many people who either don't know what the toolkit is or know what it is and don't want it.

I mean, it's kinda tough because there's a whole big space of possibilities here: 2 or 3, toolkit or not, cross built or not, and so on. But the UI we have is a linear list of templates; that list isn't excessively long, even with the addition I'm suggesting; and most importantly, I think a plain, toolkit-less Scala 3 template would probably be literally the single most popular option it were offered.

eed3si9n: I think it comes down to who the target audience and intended usages. I am guessing that most of the builds are initially for noodling purpose, maybe by first-time Scala users, maybe by someone experienced but want to quickly try some code out. The idea of the Toolkit, like Python, is to include basic things like JSON and HTTP whether you need it or not. So my claim is that scala/toolkit.local is a superset of scala/scala3.g8, and we should let people use it even for writing hello world, at least for now. /cc @adpi2

SethTisue: If we follow that thinking to its logical conclusion, we'd want to add the toolkit to the plain 2 and 3 templates and not bother having any toolkit-specific templates at all.

I believe the current 2 and 3 templates both add MUnit, so there is some precedent for including at least one default library.

SethTisue: cc @szymon-rd

adpi2: Since the Toolkit is still experimental I think we should have both seeds (toolkit and scala3). After we release a stable version of the Toolkit we can open this discussion again and maybe merge the Toolkit into the scala2 and scala3 seeds.

adpi2 (approved):

main/src/main/scala/sbt/TemplateCommandUtil.scala
szymon-rd: I agree with @adpi2 . We want to move toward making Toolkit the default wherever possible, so it would be a good idea to put it in the scala2 and 3 seeds. However, before making it appear transparently without any notice, we should stabilize it.

eed3si9n: So the overall consensus seems to be that `sbt init` should list "scala/scala3.g8" regardless then.

eed3si9n (approved):
Thanks @SethTisue!

Issue #6901, opened 2022-05-23
"help" command in "oldshell" produces error "No matches for regular expression" and quits since sbt 1.2


unkarjedy:
## steps
1. Use any sbt version >= 1.2.0, e.g. `sbt.version=1.2.8` or `sbt.version=1.6.2`
2. Start old shell in terminal using `sbt oldshell` (AFAIU `oldshell` used to be `shell` before sbt 1.0)
3. run `help`  or `help run`

#### expected result
The same as in sbt <= 1.2 (e.g. 1.1.6): help command is executed and doesn't quit.
![image](https://user-images.githubusercontent.com/3989292/169868970-04b4ad23-ccae-4650-8ee6-9fd32478730d.png)

#### actual result
In sbt 1.2.8 the commands prints
"No matches for regular expression 'oldshell'."
or
"No matches for regular expression 'run oldshell'."
and just quites

![image](https://user-images.githubusercontent.com/3989292/169868261-7a9dca47-04c8-4e9a-848f-e933ee7cd28a.png)

#### Similar issue in IntelliJ:
https://youtrack.jetbrains.com/issue/SCL-14325
Note Scala Plugin `sbt-shell` uses copy of `oldshell` command: (originally it was copied from `shell` long ago)
https://github.com/JetBrains/sbt-idea-shell/blob/2014bb73b36fcb6eeedd7096fc93fe2bf5600d9a/src/main/scala-sbt-1.0/org/jetbrains/sbt/pluginCommands.scala#L15

unkarjedy:
might be related, not sure: https://github.com/sbt/sbt/issues/2424

azdrojowa123:
As the problem is still present and annoying I spent some time to investigate this problem. 
The problem is `sbt.BasicCommands#runHelp` method. 
```
    val (extraArgs, remainingCommands) = s.remainingCommands match {
      case xs :+ exec if exec.commandLine == "shell" => (xs, exec :: Nil)
      case xs                                        => (xs, nil[Exec])
    }
```
If `exec.commandLine == "shell" `is false, what happens in case of `oldshell` or `idea-shell` (shell using in IntelliJ), new 
`remainingCommands=Nil` and `extraArgs` is filled with `oldshell/idea-shell`.  
That behaviour ruins all subsequent logic, because `remainingCommands` should contains `oldshell/idea-shell` and `extraArgs` should be empty. 
[...]

PR #7358, opened 2023-08-24
Fix help command for shells

azdrojowa123:
The whole problem was described in this issue: 
https://github.com/sbt/sbt/issues/6901#issuecomment-1691188010


eed3si9n (approved):
Thanks!

PR #7367, opened 2023-09-06
Fix java opts load order on windows

### LINKED ISSUE #7365: Windows launcher does not append .jvmopts to JAVA_OPTS
## steps

Set some flags on `JAVA_OPTS` and add a `.jvmopts` file to the project, launch with `sbt -v`

## problem

The values defined in `JAVA_OPTS` disappear.

## expectation

As stated in the documentation:
```
   .jvmopts            if this file exists in the current directory, its contents
                       are appended to JAVA_OPTS
```

## notes

The logic here seems backwards, if the file exists `_JAVA_OPTS ` is defined, so `JAVA_OPTS` is never loaded.

https://github.com/sbt/sbt/blob/bc00cee4df8bda48908699efa852c8c4aa13fb8e/launcher-package/src/universal/bin/sbt.bat#L115-L129

I think it should look like this:

```bat
 rem We use the value of the JAVA_OPTS environment variable if defined, rather than the config. 
 if not defined _JAVA_OPTS if defined JAVA_OPTS set _JAVA_OPTS=%JAVA_OPTS% 

 rem users can set JAVA_OPTS via .jvmopts (sbt-extras style) 
 if exist .jvmopts for /F %%A in (.jvmopts) do ( 
   set _jvmopts_line=%%A 
   if not "!_jvmopts_line:~0,1!" == "#" ( 
     if defined _JAVA_OPTS ( 
       set _JAVA_OPTS=!_JAVA_OPTS! %%A 
     ) else ( 
       set _JAVA_OPTS=%%A 
     ) 
   ) 
 ) 
 
 rem If nothing is defined, use the defaults. 
 if not defined _JAVA_OPTS if defined default_java_opts set _JAVA_OPTS=!default_java_opts!
```

JD557:
Fixes https://github.com/sbt/sbt/issues/7365

Tested locally and it seems to work as expected.

### DIFF
--- a/launcher-package/src/universal/bin/sbt.bat
+++ b/launcher-package/src/universal/bin/sbt.bat
@@ -112,20 +112,22 @@ if not defined _JAVACMD (
 
 if not defined _JAVACMD set _JAVACMD=java
 
-rem users can set JAVA_OPTS via .jvmopts (sbt-extras style)
-if exist .jvmopts for /F %%A in (.jvmopts) do (
-  set _jvmopts_line=%%A
-  if not "!_jvmopts_line:~0,1!" == "#" (
-    if defined _JAVA_OPTS (
-      set _JAVA_OPTS=!_JAVA_OPTS! %%A
-    ) else (
-      set _JAVA_OPTS=%%A
-    )
-  )
-)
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #7374, opened 2023-09-12
Use binary compiler bridge

eed3si9n:
Problem
-------
Starting Scala 2.13.12, Scala 2 has in-sourced the compiler bridge imple (https://github.com/scala/scala/pull/10472), which hopefully will be kept up to date more than the ones in Zinc.

Solution
--------
This switches to using the pre-compiled compiler bridge for >=2.13.12 /cc @lrytz

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -715,7 +715,9 @@ object Defaults extends BuildCommon {
     clean := clean.dependsOn(cleanIvy).value,
     scalaCompilerBridgeBinaryJar := Def.settingDyn {
       val sv = scalaVersion.value
-      if (ScalaArtifacts.isScala3(sv)) fetchBridgeBinaryJarTask(sv)
+      if (ScalaArtifacts.isScala3(sv) || VersionNumber(sv)
+            .matchesSemVer(SemanticSelector(s"=2.13 >=${ZincLmUtil.scala2SbtBridgeStart}")))
+        fetchBridgeBinaryJarTask(sv)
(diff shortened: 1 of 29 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -715,7 +715,9 @@ object Defaults extends BuildCommon {
     clean := clean.dependsOn(cleanIvy).value,
     scalaCompilerBridgeBinaryJar := Def.settingDyn {
       val sv = scalaVersion.value
-      if (ScalaArtifacts.isScala3(sv)) fetchBridgeBinaryJarTask(sv)
+      if (ScalaArtifacts.isScala3(sv) || VersionNumber(sv)
+            .matchesSemVer(SemanticSelector(s"=2.13 >=${ZincLmUtil.scala2SbtBridgeStart}")))
```
lrytz, line 719: ~Is the first check not redundant? Does it help for performance?~ Ah, to avoid matching Scala 3+

lrytz (approved):
LGTM, thank you! I didn't build and test it though.

Greetings from the summit in Madrid :)

ts

He-Pin:
I still saw:
```
[info] Non-compiled module 'compiler-bridge_2.12' for Scala 2.12.18. Compiling...
[info]   Compilation completed in 11.463s.
```
Will this land in scala 2.12.18+?

SethTisue:
> Will this land in scala 2.12.18+?

It won't. For Scala versions older than 2.13.12, everything will keep on working the way it used to.

PR #7386, opened 2023-09-25
Keep logger names in hash set

Duhemm:
Previously, the list of logger names that have been created in a given context with Log4J was being kept in a `Vector`. As a result, subsequent calls to `logger` on a `Log4JLoggerContext` with the same logger name would introduce duplicates in the list. Note that Log4J's logger configuration would not allow duplicates, and only one configuration object for a given logger name will exist. The duplicates in the list of logger names are therefore unnecessary.

With this patch, the logger names will be kept in a `HashSet`.

Since the logger names are only used to remove the logger configurations after the logger context is closed, and the logger configurations contain no duplicates, removing duplicate logger names should not cause any issue.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -428,6 +428,7 @@ lazy val utilLogging = (project in file("internal") / "util-logging")
       exclude[MissingTypesProblem]("sbt.internal.util.ConsoleAppender"),
       exclude[MissingTypesProblem]("sbt.internal.util.BufferedAppender"),
       exclude[MissingClassProblem]("sbt.internal.util.Terminal$BlockingInputStream$"),
+      exclude[IncompatibleResultTypeProblem]("sbt.util.LoggerContext#Log4JLoggerContext.loggers"),
     ),
   )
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

eed3si9n:
@Duhemm Could you add mima exclusion plz?

```
[error] Util Logging: Failed binary compatibility check against org.scala-sbt:util-logging_2.12:1.6.0! Found 1 potential problems (filtered 1)
[error]  * method loggers()java.util.Vector in class sbt.util.LoggerContext#Log4JLoggerContext has a different result type in current version, where it is java.util.HashSet rather than java.util.Vector
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("sbt.util.LoggerContext#Log4JLoggerContext.loggers")
```

Shouldn't be a problem since this is `private[util]`:

```scala
private[util] class Log4JLoggerContext
```

PR #7389, opened 2023-10-03
Support cross-build for external project ref

### LINKED ISSUE #7388: cross-build fails for external project reference
## steps

When defining an external module with `ProjectRef`

```
lazy val externalProj = ProjectRef(path, "external")
```

 cross building fails with

```
[error] Expected ';'
[error] ++  2.12.18 all libProj/Compile/compile fooPlugin/Compile/compile external/Compile/compile
[error]                                                                                           ^ 
```

## problem

The expanded command from cross build do not take in account the `build` location of the project ref. `external/Compile/compile` is not known.

## expectation

sbt should have used `{build}external/Compile/compile` instead

RustedBones:
Fix #7388


eed3si9n (approved):
Thanks!

PR #7392, opened 2023-10-05
Update Coursier to 2.1.7

regiskuckaertz:
https://github.com/coursier/coursier/releases/tag/v2.1.7

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -77,7 +77,7 @@ object Dependencies {
   def addSbtZincCompile = addSbtModule(sbtZincPath, "zincCompile", zincCompile)
   def addSbtZincCompileCore = addSbtModule(sbtZincPath, "zincCompileCore", zincCompileCore)
 
-  val lmCoursierShaded = "io.get-coursier" %% "lm-coursier-shaded" % "2.1.1"
+  val lmCoursierShaded = "io.get-coursier" %% "lm-coursier-shaded" % "2.1.7"
 
   def sjsonNew(n: String) =
     Def.setting("com.eed3si9n" %% n % "0.9.1") // contrabandSjsonNewVersion.value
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

project/Dependencies.scala
```diff
@@ -77,7 +77,7 @@ object Dependencies {
   def addSbtZincCompile = addSbtModule(sbtZincPath, "zincCompile", zincCompile)
   def addSbtZincCompileCore = addSbtModule(sbtZincPath, "zincCompileCore", zincCompileCore)
 
-  val lmCoursierShaded = "io.get-coursier" %% "lm-coursier-shaded" % "2.1.1"
+  val lmCoursierShaded = "io.get-coursier" %% "lm-coursier-shaded" % "2.1.7"
```
eed3si9n: Actually this version does not exist since we publish separately from coursier/sbt-coursier.

Also note that sbt is likely not affected by this CVE because we exclude Plexus Archiver.
https://github.com/coursier/sbt-coursier/blob/e2de6e30e32d9eccfab99b38ff4e4d281e00817e/build.sbt#L28-L30

eed3si9n (changes requested):

project/Dependencies.scala
eed3si9n: This will publish 2.1.2 - https://github.com/coursier/sbt-coursier/actions/runs/6420155362

eed3si9n:
```suggestion
  val lmCoursierShaded = "io.get-coursier" %% "lm-coursier-shaded" % "2.1.2"
```

eed3si9n (approved):

regiskuckaertz:
hello - sorry I completely forgot to follow-up on that one, thanks for merging this 🙏

PR #7404, opened 2023-10-12
Make loading order alphabetical for plugins

### LINKED ISSUE #1875: non-determinism from conflicting AutoPlugins
See #1839.

## steps
Suppose two auto plugins set some key `X`.

## problems
Is there a non-determism now? In other words, depending on the environment or randomness, you could end up with different build results.

## expectations

build should be stable across machines.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/1875

**Problem**
Plugins are topologically sorted, but plugins with equal weigh could modify the same key via `~=` etc, resulting in different builds depending on the artifact.

**Solution**
This attempts to fix that by first sorting the selected plugins by the class name before sorting it topologically.

### DIFF
--- a/main/src/main/scala/sbt/Plugins.scala
+++ b/main/src/main/scala/sbt/Plugins.scala
@@ -216,12 +216,12 @@ object Plugins extends PluginsFunctions {
             case Right(results) =>
               log.debug(s"  :: deduced result: ${results}")
               val selectedAtoms: List[Atom] = results.ordered
-              val selectedPlugins = selectedAtoms map { a =>
+              val selectedPlugins = (selectedAtoms map { a =>
                 byAtomMap.getOrElse(
                   a,
                   throw AutoPluginException(s"${a} was not found in atom map.")
                 )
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/Plugins.scala
```diff
@@ -216,12 +216,12 @@ object Plugins extends PluginsFunctions {
             case Right(results) =>
               log.debug(s"  :: deduced result: ${results}")
               val selectedAtoms: List[Atom] = results.ordered
```
eed3si9n, line 218: It does say ordered here, but evidently not ordered enough.

rossabaker: I think [not ordered in the Ordering sense](https://github.com/sbt/sbt/blob/54d9e17cad45f985f62871d107a5c78e1b3219db/internal/util-logic/src/main/scala/sbt/internal/util/logic/Logic.scala#L217).

Issue #7371, opened 2023-09-10
Console broken (to varying degrees) for TERM=dumb


hvesalai:
I am using sbt through the emacs `sbt-mode` (of which I'm the maintainer), which reports it self as a TERM=dumb. 

I was using sbt version 1.6.2 and noticed that the console is echoing erroneously and came to write a bug report on that. While doing so, I decided to upgrade to latest sbt (1.9.4) so as not to report a bug against an old version. To do so, I created a new clean project and there 1.6.2 as well as 1.9.4 are even more broken, so much so that at the moment the console is unusable in term=DUMP in 1.9.4.

I will thus report both behaviors (partially broken 1.6.2 and completely broken 1.9.4) here.

The partial echo problem was also reproduced on 1.8.3 by @rossabaker (in the `Cats` project)

## steps

1. Create an empty new project folder and move  to it
2. run `env TERM=dumb sbt consoleQuick`
3. type: `"foo"`
4. you will see the following:

```
scala> "foo"
foo"
<console>:1: error: unclosed string literal
       foo"
```
5. type `vval x = 1` [sic!]

```
scala> vval x = 1
val x = 1
x: Int = 1
```

If at step 2. you run  `env TERM=dumb sbt` and then in sbt-shell type `consoleQuick` it works differently. In this case, there is full echo, but the first character of the input is not skipped. The scala version in the clean project is the default `2.12.15`

For 1.6.2 in my project (with scala version `2.13.8`), I get the following

1. run `env TERM=dumb sbt consoleQuick`
2. type `"foo"` and there is no echo
```
scala> "foo"
val res1: String = foo
```
3. type `"foo)bar"` and there is echo until `)`
```
scala> "foo)bar"
"foo)val res2: String = foo()
```

4. type `"foo)bar(}zot{"`
```
scala> "foo)bar(}zot{"
"foo)bar(}val res11: String = foo)bar(}zot{
```

In the latter case, the results do not depend on whether the console was started from the command line or from sbt shell.

## problem

When in the DUMB terminal mode, there should be no echo.  However...

1.9.4 in clean and empty project folder: there is full echo (the line right after the input was given) and the console is skipping the first character of the input.

1.6.2 in my project: there is no echo on most input, but if the input contains any closing parenthesis or brackets (`)}]`), the terminal echoes the part of the expression until the last closing bracket. Note that the closing parenthesis can be even inside a string

## expectation

There should be no echo in any case and the terminal should not skip the first character of input.

[...]

PR #7420, opened 2023-10-26
Temporary fix for #7371

hvesalai:
This fix disables vi-style effects (most notably parenthesis matching) if the terminal is inside emacs.

The proper fix is to update to JLine 3.24.0 where this check has been implemented in JLine proper.


hvesalai:
Temporary fix for #7371

hvesalai:
@eed3si9n can you review this also. I was hoping it would make it to what ever is the next release. This should not be merged into the branch that has the new JLine3.24.1 as that already provides the same effect.

eed3si9n (approved):
Thanks!

PR #7426, opened 2023-11-03
Fix installSbtn for apple silicon

keynmol:
1. Handle ARM architecture for non-Linux architectures
2. Provide a Rosetta fallback if the ARM binary is not available for Apple Silicon

(2) is possibly controversial, because using sbtn through rosetta is always problematic.
But I'm just replicating current behaviour

### DIFF
--- a/main/src/main/scala/sbt/internal/InstallSbtn.scala
+++ b/main/src/main/scala/sbt/internal/InstallSbtn.scala
@@ -65,10 +65,16 @@ private[sbt] object InstallSbtn {
       if (Properties.isWin) "pc-win32.exe"
       else if (Properties.isLinux) "pc-linux"
       else "apple-darwin"
-    val isArmArchitecture: Boolean = sys.props
-      .getOrElse("os.arch", "")
-      .toLowerCase(java.util.Locale.ROOT) == "aarch64"
-    val arch = if (Properties.isLinux && isArmArchitecture) "aarch64" else "x86_64"
+
+    val isArmArchitecture: Boolean = {
+      val prop = sys.props
+        .getOrElse("os.arch", "")
+        .toLowerCase(java.util.Locale.ROOT)
+
+      prop == "arm64" || prop == "aarch64"
+    }
+
+    val arch = if (isArmArchitecture) "aarch64" else "x86_64"
     val sbtnName = s"sbt/bin/sbtn-$arch-$bin"
     val fis = new FileInputStream(sbtZip.toFile)
     val zipInputStream = new ZipInputStream(fis)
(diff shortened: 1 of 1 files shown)


main/src/main/scala/sbt/internal/InstallSbtn.scala
```diff
@@ -65,28 +65,65 @@ private[sbt] object InstallSbtn {
       if (Properties.isWin) "pc-win32.exe"
       else if (Properties.isLinux) "pc-linux"
       else "apple-darwin"
-    val isArmArchitecture: Boolean = sys.props
-      .getOrElse("os.arch", "")
-      .toLowerCase(java.util.Locale.ROOT) == "aarch64"
-    val arch = if (Properties.isLinux && isArmArchitecture) "aarch64" else "x86_64"
```
eed3si9n, line 71: For ipcsocket we created a universal binary, but for sbtn we didn't?

keynmol: Doesn't look like it - I downloaded sbt-1.9.7.zip:

```
> file **/sbtn-*
sbt/bin/sbtn-aarch64-apple-darwin: Mach-O 64-bit executable arm64
sbt/bin/sbtn-aarch64-pc-linux:     ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, BuildID[sha1]=95129380deb32d427b65bed01926e14e03defe43, for GNU/Linux 3.7.0, with debug_info, not stripped
sbt/bin/sbtn-x86_64-apple-darwin:  Mach-O 64-bit executable x86_64
sbt/bin/sbtn-x86_64-pc-linux:      ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=c847d01e7dc66b4544e2f97cea0e46952b5554ae, for GNU/Linux 3.2.0, with debug_info, not stripped
sbt/bin/sbtn-x86_64-pc-win32.exe:  PE32+ executable (console) x86-64, for MS Windows
```

keynmol: But actually you're right - a fat binary would be much better here.
Can you please point me to the place where the binaries are built before archiving?

It took me embarassing amount of time to get the iterator logic right and I'm super sad to see it being discarded :D

eed3si9n: There's a repo called sbt/sbtn-dist where we send CI job to Cirrus for aarch 64. Not sure how difficult it would be to create another job that waits and combines the outputs.

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/7551

PR #7433, opened 2023-11-14
[2.x] Remove cross-building except for lower utils

adpi2:
We cannot reasonably cross-compile `core-macros` and `util-collection`. So I extracted a few things out to a cross-compiled `util-core`. `util-logging` does not depend on `core-macros` nor `util-collection` anymore. It depends on `util-core` instead.

Being cross-compiled:
  - `util-core`
  - `util-position`
  - `util-logging`
  - `util-cache`
  - `util-control`
  - `util-relation`
  - `util-tracking`
  - `util-scripted`

Anything else is only compiled to Scala 3 (or Java).

Cross-compilation tests are reactivated on the CI.


eed3si9n (approved):
Thanks!

Issue #5219, opened 2019-10-28
updateSbtClassifiers does not download sources


OlegYch:
## steps
sbt version: 1.3.3

`project/plugins.sbt` :
`addSbtPlugin("ch.epfl.scala" % "sbt-missinglink" % "0.2.0")`

`updateSbtClassifiers`

## problem
plugin sources are not downloaded
idea import does not download plugin sources as well

## expectation
plugin sources are available in idea

## notes
seems to be related to earlier regressions like https://github.com/sbt/sbt/issues/4856

eatkins:
I've also noticed this.

OlegYch:
another note: `reload plugins; updateClassifiers` does download the sources

jatcwang:
I can confirm @OlegYch's solution works for SBT 1.3.5. IntelliJ however still has trouble detecting the source even if it is right beside it

He-Pin:
Still in sbt 1.3.8

jastice:
Issue reported in Scala plugin tracker: https://youtrack.jetbrains.com/issue/SCL-17825

nivekastoreth:
I've commented in the jetbrains ticket as well, but i think it's worth posting here as well for those just scrolling by:

so far i've found that disabling coursier in the `project` build seems to be a fairly reasonable workaround until this is fixed:

```
$ cat project/local.sbt
// Helps intellij find sources for plugins. See:
//   * https://youtrack.jetbrains.com/issue/SCL-17825
//   * https://github.com/sbt/sbt/issues/5219
useCoursier := false
```

using the above, i've been able to navigate plugin sources via intellij, but i've not done any sort of extensive testing on it or anything

steinybot:
I end up using:
```
addCommandAlias(
  "updateAllClassifiers",
  """; updateClassifiers
    |; updateSbtClassifiers
    |; reload plugins
    |; updateClassifiers
    |; updateSbtClassifiers
    |; reload return""".stripMargin
)
```
just to make extra sure.

unkarjedy:
Workaround 

>useCoursier := false

seems not working due to https://github.com/sbt/sbt/issues/5263

mkurz:
@azdrojowa123 @eed3si9n Can this be closed now that 
- #7437

got merged?

eed3si9n:
Yea.

iRevive:
I tried the `1.10.0-M1` release, and the sources of SBT and plugins were downloaded correctly. Thanks a lot!

[...]

PR #7437, opened 2023-11-21
[#5219] fix `updateSbtClassifiers` task

azdrojowa123:
[#5219](https://github.com/sbt/sbt/issues/5219)
The problem was caused by the fact that the ModuleIDs of some plugins have `default` configuration, which wasn't passed in the configuration vector to the `sbt.librarymanagement.ModuleDescriptorConfiguration` and as a result, also to the `CoursierDependencyResolution` from the `sbt-coursier` plugin.
The transfer of the `default` configuration to the `CoursierDependencyResolution` is crucial because only dependencies with configurations that are present in the `sbt.librarymanagement.ModuleDescriptorConfiguration#configurations` are taken into account.


eed3si9n (approved):
Thanks!

PR #7440, opened 2023-11-24
Update Scala 3.x + Scalafmt etc

eed3si9n:
Here's a basic upkeep PR on sbt 2.x. Most of the changes derive from bumping up Scala 3 and sjson-new, which in the new version changes the treatment of `java.io.File` (in preparation of remote caching).


eed3si9n:
```
[error] -- Error: modules/lm-coursier/target/scala-3.2.1/src_managed/main/scala/lmcoursier/CoursierConfiguration.scala:17:0 
[error] undefined: new dataclass.data # -1: TermRef(TypeRef(TermRef(ThisType(TypeRef(NoPrefix,module class <root>)),object dataclass),data),<init>) at readTasty
```

sbt-app/src/sbt-test/actions/doc/pending
eed3si9n, line 1: `doc` fails with a runtime error, but I think the fix requires me to publish util and then update Zinc.

adpi2 (approved):

PR #7442, opened 2023-11-27
[2.x] reduce abstraction in Execute and around

adpi2:
The goal of this PR is to reduce the abstraction in the task engine by moving `Task` to the `tasks` module:
- removed the type constructor param in `Execute`, `ExecuteProgress`, `NodeView`, `Node`, `Triggers`
- removed the many `asInstanceOf` in `Execute`
- removed the type params in `CompletionService` and `ConcurrentRestrictions` 

Not sure if we need to keep the `tasks-standard ` module or move everything to `tasks`. `task-standard` depends on `util-cache`, `util-logging` and `io`.


tasks/src/main/scala/sbt/ExecuteProgress.scala
```diff
@@ -14,28 +14,32 @@ import sbt.internal.util.RMap
  * except `started` and `finished`, which is called from the executing task's thread. All methods
  * should return quickly to avoid task execution overhead.
  */
-trait ExecuteProgress[F[_]] {
+trait ExecuteProgress {
```
eed3si9n, line 17: I think this is an overall good thing for sbt 2.x, but I want to loop in Gradle folks /cc @dragos, @Duhemm.

tasks/src/main/scala/sbt/Execute.scala
```diff
@@ -44,33 +45,34 @@ sealed trait Completed {
...
-  private[this] val viewCache = pMap[F, Node[F, *]]
-  private[this] val results = pMap[F, Result]
-
-  private[this] val getResult: [A] => F[A] => Result[A] = [A] =>
-    (a: F[A]) =>
+private[sbt] final class Execute(
+    config: Execute.Config,
+    triggers: Triggers,
+    progress: ExecuteProgress
+)(using view: NodeView) {
```
eed3si9n, line 63: I'm conflicted on this. I think reducing unnecessary abstraction might be a good thing, but I also feel like the overall stability of the core code owes a lot to genericity Mark has built-in by intentionally hiding details of one layer of the onion to not leak into the inner onion. This code is a quintessential example where the Task execution engine is written without the specific knowledge of `Task`.

adpi2: I've opened #7444 as an alternative PR that should address this comment

adpi2:
Replaced by #7444

PR #7444, opened 2023-11-28
[2.x] Reduce abstraction in `Execute` and around (bis)

adpi2:
This is an alternative of #7442 that should address https://github.com/sbt/sbt/pull/7442#discussion_r1406643948

Instead of using `Task[A]` directly, `Execute` uses a newly introduced `trait TaskId[A]`. So `Task[A]` is still not accessible from `Execute`.

### DIFF
--- a/tasks/src/main/scala/sbt/TaskId.scala
+++ b/tasks/src/main/scala/sbt/TaskId.scala
@@ -1,0 +1,6 @@
+package sbt
+
+import sbt.internal.util.AttributeMap
+
+trait TaskId[A]:
+  def tags: ConcurrentRestrictions.TagMap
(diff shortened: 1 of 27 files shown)


tasks/src/main/scala/sbt/TaskId.scala
```diff
@@ -0,0 +1,6 @@
+package sbt
+
+import sbt.internal.util.AttributeMap
+
+trait TaskId[A]:
+  def tags: ConcurrentRestrictions.TagMap
```
eed3si9n, line 6: Interesting. So at the end of the day the information it wants besides the identity is tags?

eed3si9n (approved):

tasks/src/main/scala/sbt/TaskId.scala
adpi2: Yes, the tags are used by the `ConcurrentRestrictions` implementations. It was a `F[Any] => TagMap` parameter in  `ConcurrentRestrictions`' constructors and now it is a field in `TaskId[A]`.

PR #7447, opened 2023-12-04
[1.10.x] Add missing logger methods that take Java Supplier

mkurz:


internal/util-logging/src/main/scala/sbt/util/Logger.scala
```diff
@@ -40,11 +40,13 @@ abstract class Logger extends xLogger {
   def success(message: => String): Unit
   def log(level: Level.Value, message: => String): Unit
 
+  def verbose(msg: Supplier[String]): Unit = debug(msg)
```
mkurz, line 43: Also the Scala verbose method above just calls debug.

eed3si9n (approved):
LGTM. Could you send it to 1.10.x first plz?

mkurz:
@eed3si9n Done!

PR #7456, opened 2023-12-13
[sbt 2.x] Remove `AList`, replace it with `TupleMapExtension`

adpi2:
We now have a common type for all `[M[_]] =>> (M[A1], M[A2], ...., M[An])` which is `Tuple.Map[Tup, M]`. And so we can define `transform`, `traverse`, `mapN` as extension methods. No need to create a new instance of `AList` for each type of tuple. And for `[M[_]] =>> List[M[A]]`, we can use the standard `map`, `flatMap` etc. Thus we don't need `AList` anymore.

This allows to remove the abstract  `K[F[x]]` type parameter in `Apply`, `Node` and `MixedNode`. We also don't need the `TupleK` and `SplitK` anymore.


eed3si9n:
This is exciting!

eed3si9n (approved):
sbt 2.x leans into Scala 3.x tuples, and this really solidifies the move. I think this is fine.

SethTisue:
wow, cool

xuwei-k:
- <del> https://github.com/sbt/sbt/issues/7768 🤔 </del>

edit: scala 3 bug 😇 https://github.com/scala/scala3/issues/21779

PR #7464, opened 2023-12-15
[sbt 2.x] remote cache

eed3si9n:
For the details about this PR, please see the blog post https://eed3si9n.com/sbt-remote-cache/.

Based on [RFC-1: sbt cache ideas](https://eed3si9n.com/sbt-cache-ideas/), this PR implements automatic cached task called `Def.cachedTask`:

```scala
someKey := Def.cachedTask {
  val output = StringVirtualFile1("a.txt", "foo")
  Def.declareOutput(output)
  name.value + version.value + "!"
}
```

This uses Scala 3 macro to automatically track the dependent tasks as cache keys, and serialize and deserialize the outputs. The requirement for the inputs is that they must implement [sjsonnew.HashWriter](https://github.com/eed3si9n/sjson-new/blob/66f05ac562a5c4ed544d24c41aacf3b69a9318f4/core/src/main/scala/sjsonnew/HashWriter.scala) a typeclass for a [Merkle tree](https://en.wikipedia.org/wiki/Merkle_tree). The result type must satisfy [sjsonnew.JsonFormat](https://github.com/eed3si9n/sjson-new/blob/develop/core/src/main/scala/sjsonnew/JsonFormat.scala).


core-macros/src/main/scala/sbt/internal/util/appmacro/Cont.scala
```diff
@@ -203,57 +261,127 @@ trait Cont:
...
+        val block = letOutput(outputs)(body)
+        '{
+          given HashWriter[A2] = $inputHashWriter
+          given JsonFormat[A1] = $aJsonFormat
+          given ClassTag[A1] = $aClassTag
+          ActionCache
+            .cache($input, $codeContentHash)({ _ =>
+              $block
+            })($cacheConfigExpr)
+            .value
```
eed3si9n: This is the macro code calling `ActionCache.cache(...)`

julienrf:
Thank you @eed3si9n for working on this!

> ```scala
> someKey := Def.cachedTask {
>   val output = StringVirtualFile1("a.txt", "foo")
>   Def.declareOutput(output)
>   name.value + version.value + "!"
> }
> ```
> 
> This uses Scala 3 macro to automatically track the dependent tasks as cache keys, and serialize and deserialize the outputs.

I think relying on macros makes the implementation harder to maintain, more fragile (do we handle all the possible cases for `Def.declareOutput`?), and the result might be less user-friendly because of the “magic” that happens in the macro expansion.

I believe that tracking the dependent tasks needs a macro (just like `Def.task` does), but I wonder if it would be simpler to use plain Scala for the `output`. What do you think of an API like the following?

~~~ scala
someKey := Def.cachedTask {
  val output = StringVirtualFile1("a.txt", "foo")
  val result = name.value + version.value + "!"
  CachedTaskResult(result, output)
}
~~~

eed3si9n:
@julienrf  Thanks for the comment!

> > This uses Scala 3 macro to automatically track the dependent tasks as cache keys, and serialize and deserialize the outputs.
> 
> I think relying on macros makes the implementation harder to maintain, more fragile (do we handle all the possible cases for `Def.declareOutput`?), and the result might be less user-friendly because of the “magic” that happens in the macro expansion.

In my experience, macro itself doesn't really make the maintenance all that difficult for sbt maintainers, compared to other functional areas like cross building, library dependencies, and incremental compilation. I understand that there's an intuition that macro must make things harder, but once it's implemented it's actually stable and affords flexibility for us to tweak the details. It's not dissimilar to using YAML or S-expression for user configuration + interpreter. Instead of YAML, we're using the AST.

Regarding,

```scala
  CachedTaskResult(result, output)
```

We'll consider it, but it feels inconsistent with the rest of sbt tasks, like [input tasks](https://www.scala-sbt.org/1.x/docs/Input-Tasks.html), [dynamic task](https://www.scala-sbt.org/1.x/docs/Howto-Dynamic-Task.html).

Another thing to consider is ideally, if we can get more and more things to be disk cacheable, we could consider flipping the default to be cached. i.e.:

```scala
someKey := name.value + version.value + "!"
```

The above can easily be a cached task in sbt 2.x, but not in sbt 1.x. The other thing to consider is extensibility. As the blog post hinted, we likely need to track more side effects than just file outputs. I think Bazel supports things like displaying to stdout, so we'd likely need to support that for other tasks where re-rendering stdout might be important.

util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -0,0 +1,70 @@
...
+import sjsonnew.{ HashWriter, JsonFormat }
+import sjsonnew.support.murmurhash.Hasher
+import xsbti.VirtualFile
+import java.nio.file.Path
+import scala.quoted.{ Expr, FromExpr, ToExpr, Quotes }
+
+object ActionCache:
+  def cache[I: HashWriter, O: JsonFormat: ClassTag](
+      key: I,
+      otherInputs: Long,
```
adpi2: `otherInputs` is not very explicit. Should it be `codeContentHash` instead, or `actionHash`?

util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -0,0 +1,70 @@
...
+      tags: List[CacheLevelTag],
+  )(
+      action: I => (O, Seq[VirtualFile])
+  )(
+      config: BuildWideCacheConfiguration
+  ): ActionResult[O] =
```
adpi2: I think we just need to return `O` here.

core-macros/src/test/scala/sbt/internal/ContTest.scala
```diff
@@ -4,37 +4,53 @@ import sbt.internal.util.appmacro.*
...
 import sbt.util.Applicative
+import sjsonnew.BasicJsonProtocol
 
 object ContTest extends BasicTestSuite:
   test("pure") {
     given Applicative[List] = sbt.util.ListInstances.listMonad
+    import BasicJsonProtocol.given
     val actual = contMapNMacro[List, Int](12)
     assert(actual == List(12))
+    println(ContTestMacro.inMemoryCache.toString())
```
adpi2: What are those changes? They seem useless to me.

main-settings/src/main/scala/sbt/std/InputTaskMacro.scala.scala
adpi2, line 1: The name of this file is wrong: `InputTaskMacro.scala.scala`

main-settings/src/main/scala/sbt/std/TaskMacro.scala
```diff
@@ -53,22 +53,35 @@ object TaskMacro:
...
+      qctx: Quotes
+  ): Expr[Initialize[Task[A1]]] =
     import qctx.reflect.*
+    val convert1 = new FullConvert(qctx, 1000)
+    // val aJsonFormat = convert1.summonJsonFormat[A1]
     expr match
       case '{ if ($cond) then $thenp else $elsep } =>
         '{
+          // given JsonFormat[A1] = $aJsonFormat
+          import BasicJsonProtocol.given
```
adpi2: Do we still need those comment? Also it seems the `import BasicJsonProtocol.given` is not used.

main/src/main/scala/sbt/Cross.scala
```diff
@@ -398,7 +398,7 @@ object Cross {
             scope / scalaVersion := version,
             scope / crossScalaVersions := scalaVersions,
             scope / scalaHome := Some(home),
-            scope / scalaInstance := inst
+            scope / scalaInstance := Def.task { inst }.value,
```
adpi2: What's this?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -547,16 +558,16 @@ object Defaults extends BuildCommon {
...
     // Use a different history path for jline3 because the jline2 format is
     // incompatible. By sbt 1.4.0, we should consider revering this to t / ".history"
     // and possibly rewriting the jline2 history in a jline3 compatible format if the
     // history file is incompatible. For now, just use a different file to facilitate
     // going back and forth between 1.3.x and 1.4.x.
     historyPath := (historyPath or target(t => Option(t / ".history3"))).value,
     sourceDirectory := baseDirectory.value / "src",
-    sourceManaged := crossTarget.value / "src_managed",
-    resourceManaged := crossTarget.value / "resource_managed",
+    sourceManaged := target.value / "src_managed",
```
adpi2, line 569: Do we still need `crossTarget`, can we remove it? Or make it an alias of `target`.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2356,17 +2457,75 @@ object Defaults extends BuildCommon {
...
-      task =>
-        // TODO - Should readAnalysis + saveAnalysis be scoped by the compile task too?
-        compileIncrementalTaskImpl(task, s, ci, ping, reporter)
-    }
-  }
+
+  def compileIncrementalTaskSettings(key: TaskKey[(Boolean, HashedVirtualFileRef)]) =
+    inTask(key)(
+      Seq(
+        (TaskZero / key) := (Def
```
adpi2: What does this pattern do ? Defining a task in its own scope...

Also I suggest we inline `compileInternal` here instead of `key`. It would make things more explicit.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2356,17 +2457,75 @@ object Defaults extends BuildCommon {
...
+            val setup: Setup = (TaskZero / compileIncSetup).value
+            val useBinary: Boolean = enableBinaryCompileAnalysis.value
+            val c = fileConverter.value
+            val analysisResult: CompileResult =
+              BspCompileTask
+                .compute(bspTargetIdentifier.value, thisProjectRef.value, configuration.value) {
+                  bspTask =>
+                    // TODO - Should readAnalysis + saveAnalysis be scoped by the compile task too?
+                    compileIncrementalTaskImpl(bspTask, s, ci, ping, reporter)
+                }
```
adpi2, line 2479: When the compilation is restored from the cache, sbt does not send any BSP diags (no warning are reported). To fix this we should read the analysis and send the final BSP report out of the `cachedTask`.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2509,7 +2663,19 @@ object Defaults extends BuildCommon {
           setup,
           prev
         )
-      }
+      },
+      // todo: Zinc's hashing should automatically handle directories
+      compileInputs2 := {
+        val c = fileConverter.value
```
adpi2: Not used

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2509,7 +2663,19 @@ object Defaults extends BuildCommon {
...
+        val cp0 = classpathTask.value
+        val inputs = compileInputs.value
+        CompileInputs2(
+          data(cp0).toVector,
+          inputs.options.sources.toVector,
+          scalacOptions.value.toVector,
```
adpi2, line 2672: I think that `scalacOptions` contain the file paths of plugins. They should be virtualized somehow.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2727,24 +2899,32 @@ object Classpaths {
   )
   private[this] def classpaths: Seq[Setting[_]] =
     Seq(
-      externalDependencyClasspath := concat(unmanagedClasspath, managedClasspath).value,
+      externalDependencyClasspath := {
+        summon[JsonFormat[File]]
```
adpi2: Do we need this?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -4176,22 +4395,34 @@ object Classpaths {
...
-          AttributeMap.empty
-            .put(artifact.key, art)
-            .put(moduleID.key, module)
-            .put(configuration.key, config)
+        val vf = converter.toVirtualFile(file.toPath())
+        Attributed(vf)(
+          StringAttributeMap.empty
+            .put(Keys.artifactStr, RemoteCache.artifactToStr(art))
+            .put(Keys.moduleIDStr, moduleIdJsonKeyFormat.write(module))
+            .put(Keys.configurationStr, config.name)
```
adpi2: Not really related to this PR. Why do we use a shapeless `Attributed[HashedVirtualFileRef]` instead of a concrete class:
```scala
class ClasspathEntry(path: HashedVirtualFileRef, artifact: String, moduleID: String, configuration: String, analysis: Option[String])
```
Would it make sense to get rid of `Attributed`?

main/src/main/scala/sbt/RemoteCache.scala
```diff
@@ -34,15 +40,89 @@ import sbt.nio.FileStamp
...
+  Logger
+}
+import sjsonnew.JsonFormat
+import xsbti.{ HashedVirtualFileRef, VirtualFileRef }
+import xsbti.compile.{ AnalysisContents, CompileAnalysis, MiniSetup, MiniOptions }
 
 import scala.annotation.nowarn
+import scala.collection.mutable
 
 object RemoteCache {
```
adpi2, line 59: I find it confusing to have the action cache settings mixed in the analysis cache settings. Should the analysis cache setting be in a separate plugin?

main/src/main/scala/sbt/RemoteCache.scala
```diff
@@ -34,15 +40,89 @@ import sbt.nio.FileStamp
...
+
+  private[sbt] def getCachedAnalysis(ref: String): CompileAnalysis =
+    getCachedAnalysis(CacheImplicits.strToHashedVirtualFileRef(ref))
+  private[sbt] def getCachedAnalysis(ref: HashedVirtualFileRef): CompileAnalysis =
+    analysisStore.getOrElseUpdate(
+      ref, {
+        val vfs = cacheStore.getBlobs(ref :: Nil)
+        vfs.headOption match
+          case Some(vf) =>
+            IO.withTemporaryFile(vf.name, ".tmp"): file =>
```
adpi2: Chances are `vf` is a real file so why do we transfer it to a temp file?

util-cache/src/main/scala/sbt/internal/util/StringVirtualFile1.scala
adpi2, line 1: Why not `StringVirtualFile` instead of `StringVirtualFile1`?

util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
```diff
@@ -0,0 +1,195 @@
...
+   */
+  def syncBlobs(refs: Seq[HashedVirtualFileRef], outputDirectory: Path): Seq[Path]
+end ActionCacheStore
+
+class AggregateActionCacheStore(stores: Seq[ActionCacheStore]) extends ActionCacheStore:
+  override def get[A1: ClassTag: JsonFormat](input: Digest): Option[ActionResult[A1]] =
+    var result: Option[ActionResult[A1]] = None
+    stores.foreach: store =>
+      if result.isEmpty then result = store.get[A1](input)
+    result
```
adpi2: Don't you prefer the functionnal style?

```scala
stores.foldLeft(Option.empty[ActionResult[A1]]):
    (res, store) => res.orElse(store.get[A1](input))
```

adpi2: same below

util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
```diff
@@ -0,0 +1,195 @@
...
+      else None
+
+  override def syncBlobs(refs: Seq[HashedVirtualFileRef], outputDirectory: Path): Seq[Path] =
+    refs.flatMap: r =>
+      val casFile = casBase.toFile / Digest(r.contentHashStr).toString
+      if casFile.exists then
+        val shortPath =
+          if r.id.startsWith("${OUT}/") then r.id.drop(7)
+          else r.id
+        val outPath = outputDirectory.resolve(shortPath)
```
adpi2, line 193: Why don't we use the `fileConverter` to convert the `HashedVirtualFileRef` to `Path`?

util-collection/src/main/scala/sbt/internal/util/Attributes.scala
```diff
@@ -256,36 +260,196 @@ private class BasicAttributeMap(private val backing: Map[AttributeKey[_], Any])
   override def toString = entries.mkString("(", ", ", ")")
 }
 
+/**
+ * An immutable map where a key is the tuple `(String,T)` for a fixed type `T` and can only be
+ * associated with values of type `T`. It is therefore possible for this map to contain mappings for
+ * keys with the same label but different types. Excluding this possibility is the responsibility of
+ * the client if desired.
+ */
```
adpi2: Outdated comments all the way done. It should be `String` instead of `T`.

But fundamentally we can define `type StringAttributeMap = Map[StringAttributeKey, String]` instead.

core-macros/src/main/scala/sbt/internal/util/appmacro/Cont.scala
```diff
@@ -203,57 +266,140 @@ trait Cont:
...
+                          case '[cacheInputTpe] =>
+                            callActionCacheWithA2(outputBuf.toList, cacheConfigExpr, tags)(
+                              body = modifiedBody,
+                              input = br.cacheInputExpr(p0).asExprOf[cacheInputTpe],
+                            ).asTerm.changeOwner(sym)
+                      else
+                        val tags = CacheLevelTag.all.toList
+                        callActionCacheWithUnit(outputBuf.toList, cacheConfigExpr, tags)(
+                          modifiedBody
+                        ).asTerm.changeOwner(sym)
```
adpi2: We don't need to have `callActionCacheWithA2` and `callActionCacheWithUnit`, just to save one argument once. 

```suggestion
                            callActionCache(outputBuf.toList, cacheConfigExpr, tags)(
                              body = modifiedBody,
                              input = br.cacheInputExpr(p0).asExprOf[cacheInputTpe],
                            ).asTerm.changeOwner(sym)
                      else
                        val tags = CacheLevelTag.all.toList
                        callActionCache(outputBuf.toList, cacheConfigExpr, tags)(
                          modifiedBody,
                          '{ ( ) },
                        ).asTerm.changeOwner(sym)
```

main-settings/src/main/scala/sbt/Def.scala
```diff
@@ -289,6 +304,10 @@ object Def extends Init[Scope] with TaskMacroExtra with InitializeImplicits:
    */
   def promise[A]: PromiseWrap[A] = new PromiseWrap[A]()
 
+  inline def declareOutput(inline vf: VirtualFile): Unit =
+    InputWrapper.`wrapOutput_\u2603\u2603`[VirtualFile](vf)
```
adpi2, line 307: Should we rename `InputWrapper` to `InputOutputWrapper`?

main-actions/src/main/scala/sbt/Pkg.scala
```diff
@@ -0,0 +1,338 @@
...
+import xsbti.{ FileConverter, HashedVirtualFileRef, VirtualFile, VirtualFileRef }
+
+/**
+ * == Package ==
+ *
+ * This module provides an API to package jar files.
+ *
+ * @see [[https://docs.oracle.com/javase/tutorial/deployment/jar/index.html]]
+ */
+object Pkg:
```
adpi2, line 43: Why did you rename `Package` to `Pgk`?

main-actions/src/main/scala/sbt/Pkg.scala
```diff
@@ -0,0 +1,338 @@
...
+          case Some(js) =>
+            unbuilder.beginObject(js)
+            val m = unbuilder.readField[Manifest]("m")
+            unbuilder.endObject()
+            PackageOption.JarManifest(m)
+          case None => deserializationError("Expected JsObject but found None")
+      override def write[J](obj: PackageOption.JarManifest, builder: Builder[J]): Unit =
+        builder.beginObject()
+        builder.addField("m", obj.m)
+        builder.endObject()
```
adpi2, line 270: Defining `JsonFormat` by hand is quite verbose, should we use automatic derivation instead?

main/src/main/scala/sbt/Keys.scala
```diff
@@ -222,11 +237,13 @@ object Keys {
   val consoleProject = taskKey[Unit]("Starts the Scala interpreter with the sbt and the build definition on the classpath and useful imports.").withRank(AMinusTask)
   val compile = taskKey[CompileAnalysis]("Compiles sources.").withRank(APlusTask)
   val manipulateBytecode = taskKey[CompileResult]("Manipulates generated bytecode").withRank(BTask)
-  val compileIncremental = taskKey[CompileResult]("Actually runs the incremental compilation").withRank(DTask)
+  val compileIncremental = taskKey[(Boolean, HashedVirtualFileRef)]("Actually runs the incremental compilation").withRank(DTask)
```
adpi2, line 240: Does it make sense to cache the `hasModified: Boolean`? Do we ever need it as an output of this task?

project/HouseRulesPlugin.scala
```diff
@@ -14,15 +14,15 @@ object HouseRulesPlugin extends AutoPlugin {
...
-      .toList,
+    // scalacOptions ++= "-Xfatal-warnings"
+    //   .ifScala(v => {
+    //     sys.props.get("sbt.build.fatal") match {
+    //       case Some(_) => java.lang.Boolean.getBoolean("sbt.build.fatal")
+    //       case _       => v == 12
+    //     }
+    //   })
+    //   .value
+    //   .toList,
```
adpi2, line 25: Should we enable this on the CI only?

sbt-app/src/sbt-test/classloader-cache/jni/build.sbt
```diff
@@ -10,7 +10,7 @@ val wrappedTest = taskKey[Unit]("Test with modified java.library.path")
 def wrap(task: InputKey[Unit]): Def.Initialize[Task[Unit]] =
   Def.sequential(appendToLibraryPath, task.toTask(""), dropLibraryPath)
 
-ThisBuild / turbo := true
+// ThisBuild / turbo := true
```
adpi2, line 13: Does it not work with turbo mode?

sbt-app/src/sbt-test/compiler-project/scala3-tasty-management/test
```diff
@@ -1,12 +0,0 @@
-> check
-> compile
```
adpi2, line 2: Instead of deleting this test, we can use the `products` task.

sbt-app/src/sbt-test/tests/source-directory-name/test
```diff
@@ -1,3 +0,0 @@
-> ++2.12.17!
-> compile
```
adpi2, line 2: Same here: instead of deleting we can use the `products` task.

main-settings/src/main/scala/sbt/Append.scala
```diff
@@ -67,11 +68,15 @@ object Append:
...
   implicit def appendDouble: Value[Double, Double] = _ + _
 
-  implicit def appendClasspath: Sequence[Classpath, Seq[File], File] =
-    new Sequence[Classpath, Seq[File], File] {
-      def appendValues(a: Classpath, b: Seq[File]): Classpath = a ++ Attributed.blankSeq(b)
-      def appendValue(a: Classpath, b: File): Classpath = a :+ Attributed.blank(b)
-    }
+  // given Sequence[Classpath, Seq[File], File] with
+  //   override def appendValues(a: Classpath, b: Seq[File]): Classpath = a ++ Attributed.blankSeq(b)
+  //   override def appendValue(a: Classpath, b: File): Classpath = a :+ Attributed.blank(b)
```
adpi2: We can remove those comments.

main-settings/src/main/scala/sbt/Def.scala
```diff
@@ -18,15 +19,16 @@ import sbt.Scope.{ GlobalScope, ThisScope }
...
-import xsbti.VirtualFile
+import xsbti.{ HashedVirtualFileRef, VirtualFile }
 import sjsonnew.JsonFormat
 
 /** A concrete settings system that uses `sbt.Scope` for the scope type. */
 object Def extends Init[Scope] with TaskMacroExtra with InitializeImplicits:
-  type Classpath = Seq[Attributed[File]]
-  type VirtualClasspath = Seq[Attributed[VirtualFile]]
+  type Classpath = Seq[Attributed[HashedVirtualFileRef]]
+  // type FileClasspath = Seq[Attributed[File]]
```
adpi2: Another comment we can remove.

util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -0,0 +1,70 @@
...
+
+@meta.getter
+class cacheLevel(
+    include: Array[CacheLevelTag],
+) extends StaticAnnotation
+
+enum CacheLevelTag:
+  case Local
+  case Remote
+end CacheLevelTag
```
adpi2, line 65: In which case do we want to cache locally but not globally or vice versa?

eed3si9n: I think the granularity of remote caching might need careful thinking to avoid turning small CPU-bound tasks to a network-bound one. On the other hand, I'm hoping that most tasks can be cached locally safely. I can't think of concrete examples but things like various mappings (file to path mapping) could be a candidate for marking local-only.

util-cache/src/main/scala/sbt/util/ActionCache.scala
eed3si9n: I can add a param specifically for code hash, but in general it's convenient to have a vague input hash in addition, similar to "extra" hash we have for Zinc (currently used to capture JDK version) so we can improve the invalidation in the future.

core-macros/src/test/scala/sbt/internal/ContTest.scala
eed3si9n: Probably I was debugging something. I'll remove this.

main/src/main/scala/sbt/Cross.scala
eed3si9n: I'll revert this change. Probably a remnant from trying to make all tasks cached by default.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: We can remove it, but for plugin migration it would be better to make it an alias.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Yea. This is certainly a todo. We might try to fix that in sbt 1.x as well since it's a frequent issue brought up for the 1.x remote cache as well.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: I have to double check but I think Attributed is used in a few places in the code essentially as var hack to capture some information in the middle of one task that are used later.

main-actions/src/main/scala/sbt/Pkg.scala
eed3si9n: OS like macOS and Windows typically use case insensitive file system, so `Package` as a class name is a bad idea since it can't coexist with a package object artifacts like class file or `package.html` file for Scaladoc.

util-cache/src/main/scala/sbt/util/ActionCache.scala
eed3si9n: Addressed in 768ca605ddabd4b39b61b8968e5b40be6342059b

util-cache/src/main/scala/sbt/util/ActionCache.scala
eed3si9n: Addressed in 9cfac5ffdbd34f0d21417b1cb1e4231188fa01c2 to return `O`. In theory, we potentially could return `ActionResult[O]` and let the caller decide what to do with the output files, but written as is, it probably makes sense let this function handle the side effects.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: > What does this pattern do?

So `inTask(key)(Seq(...))` is useful when you want to task-scope a bunch of setting references to some `key`, and reuse it. The best example is `packageTaskSettings(...)` where it's used to define `mappings` etc scoped to `packageBin` etc like:

```scala
      packageTaskSettings(packageBin, packageBinMappings) ++
      packageTaskSettings(packageSrc, packageSrcMappings) ++
      packageTaskSettings(packageDoc, packageDocMappings) ++
```

Now if you want to define `packageBin` itself in `packageTaskSettings`, you don't want to task-scope `key`, so you write `TaskZero / key` so you get an undecorated `packageBin` task.

> Also I suggest we inline `compileInternal` here instead of `key`. It would make things more explicit.

Yea, I think that makes sense since there's no reuse.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Not for now. I think this is another remnant from making tasks cached by default.

main-settings/src/main/scala/sbt/std/InputTaskMacro.scala.scala
eed3si9n: Fixed in 76d76ee.

main-settings/src/main/scala/sbt/std/TaskMacro.scala
eed3si9n: Fixed in 390c7bbc82f94e742b4b8d5d5a9d3fa44c636880

main/src/main/scala/sbt/Cross.scala
eed3si9n: Fixed in 390c7bbc82f94e742b4b8d5d5a9d3fa44c636880

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Fixed in 390c7bbc82f94e742b4b8d5d5a9d3fa44c636880

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Fixed in 390c7bbc82f94e742b4b8d5d5a9d3fa44c636880

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Fixed in 390c7bbc82f94e742b4b8d5d5a9d3fa44c636880

core-macros/src/test/scala/sbt/internal/ContTest.scala
eed3si9n: Fixed in 390c7bbc82f94e742b4b8d5d5a9d3fa44c636880

util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
eed3si9n: Fixed in https://github.com/sbt/sbt/pull/7464/commits/268a85b6f654b7b02bf9abdf9bc5f369f34c8c7a.

In general though, there's a subtle difference in the way we are traversing the list of stores. I think these are sort of the patterns:
1. return the first Some(...)
2. take some action for all stores, and return first non-empty
3. return the first non-empty

`foldLeft` can implement these, but I feel like `foreach` and `var` in cases like these makes thing more readable? In 268a85b6f654b7b02bf9abdf9bc5f369f34c8c7a I've added some helper extension method:

```scala
  extension [A1](xs: Seq[A1])
    // unlike collectFirst this accepts A1 => Option[A2]
    inline def collectFirst1[A2](f: A1 => Option[A2]): Option[A2] =
      xs.foldLeft(Option.empty[A2]): (res, x) =>
        res.orElse(f(x))

  override def get[A1: ClassTag: JsonFormat](input: Digest): Option[ActionResult[A1]] =
    stores.collectFirst1(_.get[A1](input))
```

util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
eed3si9n: We probably should limit generally where sbt might write files out to during build.

util-collection/src/main/scala/sbt/internal/util/Attributes.scala
eed3si9n: Fixed in 492fbb983ab3c8149941cb3efcead3a6f89348ad

core-macros/src/main/scala/sbt/internal/util/appmacro/Cont.scala
eed3si9n: I fixed it in c704fe357e07aae37f14caedbea74a08387cdeb0, but note `callActionCacheWithUnit` appears in the code three times, and usually threepeat is a signal to refactor.

main-settings/src/main/scala/sbt/Def.scala
eed3si9n: For the most part it's still tracking the input tasks into the task (e.g. `compile.value`), so I think this is ok. Maybe there's a better name that doesn't focus too much on being either input or output.

sbt-app/src/sbt-test/classloader-cache/jni/build.sbt
eed3si9n: No. Here's the result I get with turbo mode:

```scala
[info] -------------------------------- Running Tests --------------------------------
[info] X sbt.JniLibraryTest.load.native 79ms
[info]   java.lang.IllegalAccessError: tried to access class sbt.JniLibrary from class sbt.JniLibraryTest$
[info]     sbt.JniLibraryTest$.$anonfun$tests$3(JniLibraryTest.scala:11)
[info] [error] Test suite sbt.JniLibraryTest failed with java.lang.IllegalAccessError: tried to access class sbt.JniLibrary from class sbt.JniLibraryTest$.
[info] [error] This may be due to the ClassLoaderLayeringStrategy (AllLibraryJars) used by your task.
[info] [error] To improve performance and reduce memory, sbt attempts to cache the class loaders used to load the project dependencies.
[info] [error] The project class files are loaded in a separate class loader that is created for each test run.
[info] [error] The test class loader accesses the project dependency classes using the cached project dependency classloader.
[info] [error] With this approach, class loading may fail under the following conditions:
[info] [error]
[info] [error]  * Dependencies use reflection to access classes in your project's classpath.
[info] [error]    Java serialization/deserialization may cause this.
[info] [error]  * An open package is accessed across layers. If the project's classes access or extend
[info] [error]    jvm package private classes defined in a project dependency, it may cause an IllegalAccessError
[info] [error]    because the jvm enforces package private at the classloader level.
[info] [error]
[info] [error] These issues, along with others that were not enumerated above, may be resolved by changing the class loader layering strategy.
[info] [error] The Flat and ScalaLibrary strategies bundle the full project classpath in the same class loader.
[info] [error] To use one of these strategies, set the  ClassLoaderLayeringStrategy key
[info] [error] in your configuration, for example:
[info] [error]
[info] [error] set root / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.ScalaLibrary
[info] [error] set root / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
[info] [error]
[info] [error] See ClassLoaderLayeringStrategy.scala for the full list of options.
[info] [info] Tests: 1, Passed: 0, Failed: 1
[info] [error] Failed tests:
[info] [error] 	sbt.JniLibraryTest
[info] [error] (Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
```

main-settings/src/main/scala/sbt/Append.scala
eed3si9n: Fixed in dde0e8eaf1929c63fec67c89a37b4ee27edcf289

main-settings/src/main/scala/sbt/Def.scala
eed3si9n: Fixed in dde0e8eaf1929c63fec67c89a37b4ee27edcf289

main/src/main/scala/sbt/RemoteCache.scala
eed3si9n: I guess I implemented this before I created `ActionCacheStore#syncBlobs`. Hopefully that should be the same. Fixed in 3c0115a8b7a95cf4a570dade9544530cafd2f956.

main/src/main/scala/sbt/RemoteCache.scala
eed3si9n: In this PR, analysis cache is modified to share the same `RemoteCache.cacheStore` mechanism. In `RemoteCache.postAnalysis` analysis file is uploaded via `cacheStore.putBlobs`, and they are retrieved back in `getCachedAnalysis` using `cacheStore.syncBlobs`. Hopefully this means that we can produce an analysis once system-wide and share it for the same file content.

main/src/main/scala/sbt/Keys.scala
eed3si9n: iirc, I had to weaken `CompileResult` to `(Boolean, HashedVirtualFileRef)` because `CompileResult` contains `File` or something, and wanted to avoid it. `CompileResult` contains `hasModified`.
This is used in sbt 1.x for pipelining etc - https://github.com/sbt/sbt/blob/v1.10.0-M1/main/src/main/scala/sbt/Defaults.scala#L2260. We might have to revisit this area, potentially by making `CompileResult` machine-independent.

project/HouseRulesPlugin.scala
eed3si9n: I don't think we can. Note that the `-Xfatal-warnings` here is added only for Scala 2.12. So the equivalent for sbt 2.x would be to enable `-Xfatal-warnings` for Scala 3.x and use `@nowarn` to suppress warnings we're willing to live with. We should do that, but probably not on this PR. I've split up https://github.com/sbt/sbt/pull/7486 to comment out this flag after porting it to Scala 3, and also migrate away from postfix usages.

sbt-app/src/sbt-test/compiler-project/scala3-tasty-management/test
eed3si9n: Fixed in 2f01d9e7f637307e4526d89f3a3f0f290d8d1d5a by calling `packageBin` and unzipping the result. I think this keeps the spirit of the test without asserting the existence of `classes` directory.

sbt-app/src/sbt-test/tests/source-directory-name/test
eed3si9n: Fixed in 49980f7949afce908ea09ed645c263c77e8b533a

main/src/main/scala/sbt/Defaults.scala
eed3si9n: In a monorepo build tool, like Bazel, a cached compilation is treated like our Coursier library dependency JAR download, so the downstream targets (subprojects) would work, but the intermediate side effects like warnings aren't displayed anywhere. This kind of make sense because with Bazel you handle the entire company's code, you really don't want to be bombarded with warnings. We might revisit this, but initially I'm sort of inlined to display warning only if we perform actual compilation.

main/src/main/scala/sbt/Defaults.scala
adpi2: > Now if you want to define packageBin itself in packageTaskSettings, you don't want to task-scope key, so you write TaskZero / key so you get an undecorated packageBin task.

Would it be clearer to write it that way?
```scala
Def.settings(
  key := ???,
  inTask(key)(
     ... // other task-scoped keys
  )
)
```

main/src/main/scala/sbt/Defaults.scala
adpi2: I see. It is understandable that a cached task does not produce all the side effects (the logs, the BSP notifications) etc. 

In this case though I wonder if we should inform the user about it: `Compilation of foo restored from the cache`.

Also sometimes we might want to bypass the cache and force a compilation. It seems that `clean; compile` is not enough anymore. So how can we do so?

adpi2 (approved):
I replied to some of your comments but overall I think this is good to go.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: We could probably provide some flag or setting to ignore cache?

eed3si9n:
Squash merged!

PR #7475, opened 2024-01-17
Use the publisher interface for publishing as default

Tammo0987:
Today I realized, that just configuring a publisher is not enough to actually use it. So I wanted to propose, to use always the configured one, which defaults to the same, so nobody has to overwrite the publish task itself, to use a custom PublisherInterface.

Please leave some feedback, because I am not sure, if this is wanted or if the PR is correctly set up.


eed3si9n (approved):
Whoa, nice catch!

PR #7480, opened 2024-01-23
Add support for SIP-51 (unfreezing the Scala library)

lrytz:
This implements the changes needed for [SIP-51, to drop forwards binary compatibility for the Scala library](https://docs.scala-lang.org/sips/drop-stdlib-forwards-bin-compat.html).

Upstream PRs have been merged and released (https://github.com/sbt/zinc/pull/1331 and https://github.com/coursier/sbt-coursier/pull/490).

Follow-up to this PR: https://github.com/sbt/sbt/pull/8012

See the individual commit messages for details.

Summary, for Scala 2.13:
  - The Scala library on the classpath is no longer pinned to `scalaVersion`, but resolved like any other library
  - The `scalaVersion` cannot be older than the Scala library on the classpath, otherwise fail the build
  - `csrSameVersions` is used to ensure all Scala artifacts (scala-library, scala-reflect, scala-compiler, scalap) are kept at the same version. This is necessary because they are compiled with the inliner enabled, mixing versions is not binary compatible.

For Scala 3
  - The scala-library.jar is already resolved like other libraries in current sbt, but only on the dependency classpath, not on the runtime classpath used when executing the compiler
  - This can lead to NoSuchMethodExceptions when expanding macros
  - This PR changes this by replacing the scala-library in the `scala-tool` configuration by the version in the `compile` configuration


lrytz:
The zinc and sbt-coursier PRs are merged. Once there are releases of these two libraries, this PR can be updated to pass CI. @eed3si9n / @alexarchambault could you take a look, are any releases planned?

eed3si9n:
Yea, I can kick off releases for both soon.

eed3si9n:
- https://github.com/sbt/zinc/releases/tag/v1.10.0-M3
- https://github.com/coursier/sbt-coursier/releases/tag/v2.1.4

eed3si9n (approved):
Let's land this.

tgodzik:
"The scalaVersion cannot be older than the Scala library on the classpath, otherwise fail the build"

How do we cross publish for different Scala versions now actually? Is it even sensibly possible? We have an issue where we publish semanticdb plugin for older versions, but since we use scalapb which uses a newer Scala version then we are no longer able to build for a version older than the scalapb's scala version.

Is the only option to have a compatibility map for every dependency?

lefou:
Not a real answer, but I think one root issue in Scala/JVM land is the missing distinction between compile-time and runtime-dependencies. Libraries should always compile against the lowest supported version, but run with the highest (if possible, to address improvements and fixes). There should be no random version bumps in compile-dependencies of libraries, since this essentially drops support for all versions below that version. But this requires a mind shift on the library user side, since downstream users need to watch and bump transitive dependencies themselves. Since almost no library I can find follows this semantics, library authors that also depends on other libraries will have a hard time to provide a wide range of Scala versions.

lrytz:
> How do we cross publish for different Scala versions now actually?

This is relevant only for compiler plugins, right?

But maybe there's actually nothing to do here. If the new `semanticdb` depends on a `scalapb` which itself depends on 2.13.14, then releasing `semanticdb` for 2.13.13 is pointless. The same `scalapb` dependency will be pulled in for users and require them to also upgrade to 2.13.14.

tgodzik:
> > How do we cross publish for different Scala versions now actually?
> 
> This is relevant only for compiler plugins, right?
> 
> But maybe there's actually nothing to do here. If the new `semanticdb` depends on a `scalapb` which itself depends on 2.13.14, then releasing `semanticdb` for 2.13.13 is pointless. The same `scalapb` dependency will be pulled in for users and require them to also upgrade to 2.13.14.

I guess so, but then we can't stop publishing for more than one latest version, since that might be too strict :/ Also we might need to backpublish support for some version from time to time. It just got harder to do so :/

tgodzik:
Looks like I managed to make it work https://github.com/scalameta/scalameta/pull/3722

PR #7486, opened 2024-01-30
[2.x] Migrate postfix ops

eed3si9n:
## Problem
There are a few places in the code that's using postfix ops.

## Solution
This rewrites the expressions to use normal dot notation.

### DIFF
--- a/main-settings/src/main/scala/sbt/Scope.scala
+++ b/main-settings/src/main/scala/sbt/Scope.scala
@@ -408,9 +408,11 @@ object Scope {
       projectInherit: ProjectRef => Seq[ProjectRef],
       configInherit: (ResolvedReference, ConfigKey) => Seq[ConfigKey]
   ): DelegateIndex = {
-    val pDelegates = refs map { case (ref, project) =>
-      (ref, delegateIndex(ref, configurations(project))(projectInherit, configInherit))
-    } toMap;
+    val pDelegates = refs
+      .map:
+        case (ref, project) =>
+          (ref, delegateIndex(ref, configurations(project))(projectInherit, configInherit))
+      .toMap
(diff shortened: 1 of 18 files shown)


main-settings/src/main/scala/sbt/Scope.scala
```diff
@@ -408,9 +408,9 @@ object Scope {
       projectInherit: ProjectRef => Seq[ProjectRef],
       configInherit: (ResolvedReference, ConfigKey) => Seq[ConfigKey]
   ): DelegateIndex = {
-    val pDelegates = refs map { case (ref, project) =>
+    val pDelegates = (refs.map { case (ref, project) =>
       (ref, delegateIndex(ref, configurations(project))(projectInherit, configInherit))
-    } toMap;
+    }).toMap
```
adpi2: I don't think we need parentheses around `refs.map()`.

Also maybe this is a bit more readable:
```scala
    val pDelegates = refs
      .map { case (ref, project) => 
        (ref, delegateIndex(ref, configurations(project))(projectInherit, configInherit))
      }
      .toMap
```

main/src/test/scala/sbt/internal/TestBuild.scala
```diff
@@ -189,9 +189,9 @@ abstract class TestBuild {
   }
 
   def mapBy[K, T](s: Seq[T])(f: T => K): Map[K, T] =
-    s map { t =>
+    s.map { t =>
       (f(t), t)
-    } toMap;
+    }.toMap
```
adpi2:
```scala
    s.map(t => (f(t), t)).toMap
```

adpi2 (dismissed):
Some suggestions but overall it looks good to me.

PR #7488, opened 2024-02-02
Resolve `scripted-sbt-redux` to run scripted

Duhemm:
Previously, in sbt 2, the Scripted plugin was not resolving any module that would provide it with the `ScriptedTests` object, and could therefore not run scripted tests.

With this patch, `scripted-sbt-redux` will be resolved and added to the classloader that's used to load `sbt.scriptedtest.ScriptedTests`.


eed3si9n (approved):
Thanks!

PR #7492, opened 2024-02-08
[2.x] Fix test in ErrorSpec

adpi2:


buildfile/src/main/scala/sbt/internal/parser/SbtParser.scala
```diff
@@ -257,24 +235,36 @@ private[sbt] case class SbtParser(path: VirtualFileRef, lines: Seq[String])
...
+                      |}""".stripMargin
     val fileName = path.id
-    val (parsedTrees, reporterId, sourceFile) = parse(content, fileName, None)
-    given ctx: Context = compileCtx
+    val reporterId = s"$fileName-${Random.nextInt}"
+    val sourceFile = SourceFile(
+      VirtualFile(reporterId, wrapCode.getBytes(StandardCharsets.UTF_8)),
+      scala.io.Codec.UTF8
+    )
+    given Context = compileCtx.fresh.setSource(sourceFile)
```
adpi2, line 248: We need to set the source in the context before we call the parser. Otherwise we need to reconstruct the positions of the trees ourselves.

eed3si9n (approved):
Thanks!

PR #7493, opened 2024-02-13
[2.x] Update eval for Scala 3.3.1

adpi2:


adpi2:
Well actually it seems that eval is not used and that there is a local copy of it in https://github.com/sbt/sbt/blob/develop/buildfile/src/main/scala/sbt/internal/Eval.scala.

@eed3si9n what's the plan here? Should we keep the local copy of `Eval` or should we use the library instead?

eed3si9n (approved):
Thanks!

eed3si9n:
If the behavior is stable maybe we should switch to the library?

PR #7494, opened 2024-02-13
[2.x] Fix `SbtRefactorings` and run `SessionSettingsSpec`

adpi2:


adpi2:
@eed3si9n It seems that `SbtRefactorings` can only rewrite simple settings such as `a := { // body }`. I wonder if it is so much useful and if we should keep it or discard it.

eed3si9n (approved):

eed3si9n:
This is used for `session save`?

adpi2:
> This is used for session save?

Yes, I think so

PR #7496, opened 2024-02-15
[1.x] add noOp field in BSP compile report

### LINKED ISSUE #6103: [BSP] - Somehow signify a no-op compilation
Following up on a quick conversation on [Gitter](https://gitter.im/sbt/sbt?at=5fb39b5574152347c23cc60d), in Metals one of the things we check when we get notified of a `build/taskStart` is to see whether or not it's a no-op compilation. You can see this here:

https://github.com/scalameta/metals/blob/8f1c5db0e76ddbb95db06a55afc395ca6a965d13/metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala#L117-L148

For better or worse, we just check in the message params to see if it starts with `Start no-op compilation`. This is Bloop specific at the moment, and you can see where this is being reported in Bloop here:

https://github.com/scalacenter/bloop/blob/4768184cc8842d5d6773646ccda93210a89f1b05/frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala#L301-L331

Within Metals we do various check on this compilation to see if we need to do certain things that can in some situations offer some nice UX and speed improvements. Just to give an example we have a Tree View of all packages, and if it's a no-op compilation we know that we don't need to compute anything new for the tree view. As state, right now this is quite Bloop specific, but it'd be great to expand this somehow so that we could have a more generic approach to know if a compilation is a no-op or not from sbt. I'd love to know if this is possible and something you'd consider to forward somehow in the `build/taskStart`. Thanks!

adpi2:
Fixes #6103

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2316,11 +2316,11 @@ object Defaults extends BuildCommon {
     val ci = (compile / compileInputs).value
     val ping = earlyOutputPing.value
     val reporter = (compile / bspReporter).value
-    BspCompileTask.compute(bspTargetIdentifier.value, thisProjectRef.value, configuration.value) {
-      task =>
+    BspCompileTask
+      .compute(bspTargetIdentifier.value, thisProjectRef.value, configuration.value, ci) { task =>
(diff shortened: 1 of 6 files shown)


eed3si9n (approved):
Thanks!

PR #7508, opened 2024-03-05
[2.x] Fix or mute warnings and other minor changes

adpi2:


main/src/main/scala/sbt/Main.scala
```diff
@@ -131,9 +131,8 @@ private[sbt] object xMain:
               )
               .put(BasicKeys.detachStdio, detachStdio)
             val state = bootServerSocket match {
-              // todo: fix this
-              // case Some(l) => state0.put(Keys.bootServerSocket, l)
-              case _ => state0
+              case Some(l) => state0.put(Keys.bootServerSocket, l)
```
adpi2, line 135: Not sure why this was commented out. It seems there is nothing to fix anymore.

buildfile/src/test/scala/sbt/internal/parser/EmbeddedXmlSpec.scala
```diff
@@ -9,7 +9,6 @@ package sbt.internal.parser
 
 import sbt.internal.util.MessageOnlyException
 
-/*
 object EmbeddedXmlSpec extends CheckIfParsedSpec {
```
adpi2, line 12: It seems this test suite runs successfully without fixing anything.

adpi2:
I broke the binary interface of `util-relation` which is used by Zinc. So we need to publish a new alpha version of Zinc and its utils on top of this PR to pass the CI.

@eed3si9n Could you release soon? Could you share the right to publish to the Scala Center?

adpi2:
Otherwise I can revert the changes in `util-relation` to preserve the binary API.

.github/workflows/ci.yml
```diff
@@ -104,6 +104,12 @@ jobs:
         rm -rf "$HOME/.sbt/scripted/" || true
         rm -rf "$HOME/.ivy2/local" || true
         rm -r $(find $HOME/.sbt/boot -name "*-SNAPSHOT") || true
+    - name: Publish local Zinc
+      shell: bash
+      run: |
+        ./sbt -v lowerUtils/publishLocal
+        cd zinc/
+        sbt publishLocal
```
adpi2, line 112: This branch depends on a SNAPSHOT version of Zinc. This will be the case until we publish the utils and Zinc.

adpi2:
@eed3si9n This is ready to review

core-macros/src/main/scala/sbt/internal/util/appmacro/ContextUtil.scala
```diff
@@ -1,15 +1,14 @@
...
 import scala.quoted.*
-import scala.reflect.TypeTest
 import scala.collection.mutable
 import sbt.util.cacheLevel
 import sbt.util.CacheLevelTag
 
-trait ContextUtil[C <: Quotes & scala.Singleton](val qctx: C, val valStart: Int):
+trait ContextUtil[C <: Quotes & scala.Singleton](val valStart: Int):
   import qctx.reflect.*
+  val qctx: C
```
eed3si9n: I am guessing that the `val` should come before the `import` line above it.

eed3si9n (commented):
I have a question about one `val`, but overall looking good. Thanks for this!

core-macros/src/main/scala/sbt/internal/util/appmacro/ContextUtil.scala
adpi2: Probably the compiler can understand his because the val is abstract. But yes it would be clearer to move the `import` after the `val`.

eed3si9n (approved):
Thanks!

PR #7513, opened 2024-03-15
Update lm-coursier-shaded: 2.1.3 -> 2.1.4

kubukoz:
Should bring in https://github.com/coursier/sbt-coursier/pull/489.


eed3si9n (approved):
Thanks!

PR #7517, opened 2024-03-19
[sbt 2] Restore server tests

adpi2:
- Port the `BuildServerEvalReporter` to Scala 3
- Use `scalatest` to run the server tests
- Run and fix all `BuildServerTests` (BSP)
- Fix `ServerCompletionTests`
- Reactivate the server tests on the CI


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2711,6 +2699,18 @@ object Defaults extends BuildCommon {
     }
   )
 
+  private inline def analysisStore: AnalysisStore = {
```
eed3si9n, line 2704: An inlined task?

adpi2: Yay that allow me to refactor code in tasks without creating intermediate task. I can use `.value` even if I'm not in a `Def.task`.

eed3si9n (dismissed):
Thanks!

adpi2:
@eed3si9n I need a new approval to merge.

eed3si9n (approved):

PR #7519, opened 2024-03-22
[sbt2] Reactivate checkTargets in Load

adpi2:
`checkTargets` verifies that the target directories of all sub-projects do not overlap.


eed3si9n (dismissed):
Thanks!

eed3si9n (approved):

PR #7522, opened 2024-03-27
[2.x] fix scripted

adpi2:


buildfile/src/main/scala/sbt/internal/EvaluateConfigurations.scala
```diff
@@ -398,27 +398,18 @@ object Index {
...
+      a <- ss.data.values
+      case AttributeEntry(_, base: Task[?]) <- a.entries
+    do
+      def update(map: TriggerMap, key: AttributeKey[Seq[Task[?]]]): Unit =
+        base.info.attributes.get(key).getOrElse(Seq.empty).foreach { task =>
+          map(task) = base +: map.getOrElse(task, Nil)
+        }
+      update(runBefore, Def.runBefore)
+      update(triggeredBy, Def.triggeredBy)
+    val onComplete = (GlobalScope / Def.onComplete).get(ss).getOrElse(() => ())
```
adpi2, line 411: This is only refactoring

main-settings/src/main/scala/sbt/Structure.scala
```diff
@@ -82,7 +82,7 @@ sealed abstract class SettingKey[A1]
   final inline def :=(inline v: A1): Setting[A1] =
     ${ TaskMacro.settingAssignMacroImpl('this, 'v) }
 
-  final inline def +=[A2](inline v: A2)(using Append.Value[A1, A2]): Setting[A1] =
+  final inline def +=[A2](inline v: A2)(using inline ev: Append.Value[A1, A2]): Setting[A1] =
```
adpi2, line 85: `ev` can be a `@compileTimeOnly def`. We need to declare it as `inline` to prevent the compiler from evaluating it, even when the expansion of the macro does not use it. See https://github.com/scala/scala3/issues/20029

main-settings/src/main/scala/sbt/std/Instances.scala
```diff
@@ -59,11 +52,17 @@ object FullInstance:
...
+  val F1F2: Applicative[[a] =>> Initialize[Task[a]]] = Applicative.given_Applicative_F1(using
+    InitializeInstance.initializeMonad,
+    Task.taskMonad
+  )
   given initializeTaskMonad: Monad[[a] =>> Initialize[Task[a]]] with
     type F[x] = Initialize[Task[x]]
     override def pure[A1](x: () => A1): Initialize[Task[A1]] = F1F2.pure(x)
+
+    override def map[A1, A2](fa: Initialize[Task[A1]])(f: A1 => A2): Initialize[Task[A2]] =
+      F1F2.map(fa)(f)
```
adpi2, line 64: If we don't override the `map` we lose the info of `taskA` in `taskA.triggeredBy(taskB).value`, which expands to something like `taskA.map(_.copy(info = /* add triggeredBy in attributes*/ )`

main/src/main/java/sbt/internal/MetaBuildLoader.java
```diff
@@ -81,41 +81,27 @@ public static MetaBuildLoader makeLoader(final AppProvider appProvider) throws I
             jnaJars);
     final Pattern pattern = Pattern.compile(fullPattern);
     final File[] cp = appProvider.mainClasspath();
-    final URL[] interfaceURLs = new URL[3];
-    final URL[] jlineURLs = new URL[7];
+    final Set<File> interfaceFiles = new LinkedHashSet<>();
+    final Set<File> jlineFiles = new LinkedHashSet<>();
```
adpi2, line 85: `extraJars` can contain other versions of the `compiler-interface` or `jline` and so we end up with more than 3 or 7 jars in this set.

main/src/main/java/sbt/internal/XMainConfiguration.java
```diff
@@ -66,6 +67,12 @@ public xsbti.MainResult run(String moduleName, xsbti.AppConfiguration configurat
 
   private xsbti.AppConfiguration makeConfiguration(xsbti.AppConfiguration configuration) {
     try {
+      boolean isScripted = Boolean.parseBoolean(System.getProperty("sbt.scripted"));
+      if (isScripted) {
+        // in batch scripted tests, we disable caching jar files
+        // to avoid interference between tests
+        URLConnection.setDefaultUseCaches("jar", false);
+      }
```
adpi2: If two scripted tests produce a jar file at the same path, the JDK caches it and does not reload it. Therefore the second scripted test would use the jar of the first test and fail.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -143,22 +144,17 @@ object Defaults extends BuildCommon {
...
+      metadata: StringAttributeMap,
+      converter: FileConverter
+  ): Option[CompileAnalysis] =
+    def asBinary(file: File) = FileAnalysisStore.binary(file).get.asScala
+    def asText(file: File) = FileAnalysisStore.text(file).get.asScala
+    for
+      ref <- metadata.get(Keys.analysis)
+      file = converter.toPath(VirtualFileRef.of(ref)).toFile
+      content <- asBinary(file).orElse(asText(file))
+    yield content.getAnalysis
```
adpi2, line 165: There was a data race on the remote cache to get the analysis files of the classpath. I don't think we should use the remote cache here because we know all modules in the classpath have been compiled already and the analysis file is in the target directory. So instead we can go read those files directly.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1426,40 +1423,35 @@ object Defaults extends BuildCommon {
...
+            val externalDeps = relations.externalDeps(c).flatMap(stamp)
+            val classStamps = relations.productClassName.reverse(c).flatMap { pc =>
+              apis.internal.get(pc).map(_.compilationTimestamp)
             }
-          )
+            val maxStamp = (internalDeps ++ externalDeps ++ classStamps).maxOption
+            maxStamp.foreach(maxStamp => stamps(c) = maxStamp)
+            maxStamp
+          }
+          stamps.get(c).orElse(computeAndStoreStamp)
```
adpi2, line 1459: This is a refactoring that helped me understand what was going on here.

Also `testQuick` and compilation caching do not work well together because of timestamps, but I'll try to fix that in a separate PR.

main/src/main/scala/sbt/RemoteCache.scala
```diff
@@ -78,43 +65,6 @@ object RemoteCache {
...
-      )
-      FileAnalysisStore.binary(file).set(AnalysisContents.create(analysis, setup))
-      val vf = tempConverter.toVirtualFile(file.toPath)
-      val refs = cacheStore.putBlobs(vf :: Nil)
-      refs.headOption match
-        case Some(ref) =>
-          analysisStore(ref) = analysis
-          Some(ref)
-        case None => None
-
```
adpi2: We don't need the cache to post and get the analyses of the classpath. First because they are already stored by the `compileIncremental` task and second because we can directly read the analyses in the target directories of each module.

project/HouseRulesPlugin.scala
```diff
@@ -32,7 +32,7 @@ object HouseRulesPlugin extends AutoPlugin {
     scalacOptions ++= "-Ywarn-unused-import".ifScala2x(v => 11 <= v && v <= 12).value.toList,
     scalacOptions ++= {
       scalaPartV.value match {
-        case Some((3, _)) => Seq("-Wunused:imports,implicits,nowarn")
+        case Some((3, _)) => Seq("-Wunused:imports,implicits") // ,nowarn
```
adpi2, line 35: `-Wunused` does not work very well on `nowarn` because of false positives.

adpi2:
@eed3si9n This is ready to review. There are a few more scripted tests to fix but I'll do it in separate PRs.

main/src/main/scala/sbt/RemoteCache.scala
eed3si9n: If I remember correctly, I think my motivation here was to unify the machine-wide disk caching with analysis cache. In addition to clean compile, and resuming from the fully compiled state, there's no a third scenario where a disk cache (or a remote cache) is hydrated but local `target/` is empty. If there's a race, probably we need a way for the participants to coordinate better?

sbt-app/src/sbt-test/actions/compile-clean/test
```diff
@@ -0,0 +1,22 @@
+$ touch target/cant-touch-this
+
+> Test/products
+$ exists target/out/jvm/scala-2.12.17/compile-clean/classes/A.class
+$ exists target/out/jvm/scala-2.12.17/compile-clean/test-classes/B.class
```
eed3si9n: I think in sbt 2.x we shouldn't promise there's a `*.class` file. Instead we should promise there's a JAR file with the class `A` inside it.

main/src/main/java/sbt/internal/XMainConfiguration.java
eed3si9n: `URLConnection.setDefaultUseCaches(...)`  is an instance method and takes a single Boolean in JDK 8 it seems - https://docs.oracle.com/javase/8/docs/api/java/net/URLConnection.html#setDefaultUseCaches-boolean-

adpi2: An instance method to set a static field :thinking:

I fixed it

main/src/main/scala/sbt/RemoteCache.scala
adpi2: > If I remember correctly, I think my motivation here was to unify the machine-wide disk caching with analysis cache.

I am not sure about that because `postAnalysis`  and `getCachedAnalysis` were only used to create instances of `Classpath` and to extract `CompileAnalysis` from `Classpath`.

> In addition to clean compile, and resuming from the fully compiled state, there's no a third scenario where a disk cache (or a remote cache) is hydrated but local target/ is empty.

I am not sure to understand what you mean here. When we get a `Classpath`, we know that `compile` was executed and that the analyses were produced, or retrieved from the cache.

> If there's a race, probably we need a way for the participants to coordinate better?

Yes but we have the task engine to coordinate. Accessing the cache is a side effect and as such it should be managed by the task engine. In that sense, `Def.cachedTask` is super safe because it ensures that a single task is responsible from posting or getting the same file from the cache. As soon as we bypass the task engine and access the cache directly we lose those guarantees.

Also the `IO.withTemporaryFile("analysis", ".tmp", true)` is suspicious because it creates randomness. It makes it impossible to hash any instance of `Classpath` deterministically.

sbt-app/src/sbt-test/actions/compile-clean/test
adpi2: I kind of agree. But that test is checking more than the output of compilation: it also tests `clean`, and `cleanKeepFiles` at the level of individual `.class` file. So I prefer to keep it as it is for now, and to update it later if needed, when we actually change the `products` task.

main-settings/src/main/scala/sbt/std/KeyMacro.scala
```diff
@@ -12,70 +12,54 @@ import java.io.File
...
-        TaskKey[A1]($n, $description)($mf)
-      }
-    }
+  def taskKeyImpl[A1: Type](description: Expr[String])(using Quotes): Expr[TaskKey[A1]] =
+    val name = definingValName(errorMsg("taskKey"))
+    val tag: Expr[KeyTag[Task[A1]]] = Type.of[A1] match
+      case '[Seq[a]] =>
+        '{ KeyTag.SeqTask(${ summonRuntimeClass[a] }) }
+      case _ => '{ KeyTag.Task(${ summonRuntimeClass[A1] }) }
+    '{ TaskKey(AttributeKey($name, $description, Int.MaxValue)(using $tag, NoJsonWriter())) }
```
adpi2, line 33: For a `taskKey[Seq[a]]` we store the runtime class of `a` in the `KeyTag`. Later when we build all settings we can match on that `runtimeClass` and if it is a `Path`, `File` or `VirtualFileRef` we automatically implement the associated `clean` task. This is tested by `actions/compile-clean`.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2497,19 +2497,18 @@ object Defaults extends BuildCommon {
...
+      val dir = classDirectory.value.toPath
+      val mappings =
+        for
+          ref <- analysisResult.analysis.readStamps.getAllProductStamps.asScala.keysIterator
+          path = c.toPath(ref)
+          if path.toFile().isFile() && path.startsWith(dir)
+        yield
+          val hashedRef: HashedVirtualFileRef = c.toVirtualFile(path)
+          val relativePath = dir.relativize(path).toString
+          hashedRef -> relativePath
```
adpi2, line 2509: When traversing the class directory we get all the old resource files, which later messes up with the `copyResources` sync. To avoid that we use the list of products in the compilation analysis.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -661,7 +663,7 @@ object Defaults extends BuildCommon {
     classDirectory := target.value / (prefix(configuration.value.name) + "classes"),
     backendOutput := {
       val converter = fileConverter.value
-      val dir = classDirectory.value
+      val dir = target.value / (prefix(configuration.value.name) + "backend-output")
```
adpi2, line 666: `classDirectory` was:
  - the folder where we compile classes to
  - the folder where we zip classes from
  - the folder where we copy resources to
  - the folder where we unzip the compiled jar to
  - the folder where we zip classes and resources from
  
And this was a source of bugs: old resource files being unzipped on top of new resource files. In practice I think it is simpler and safer if `backendOutput` (the output directory of the compiler) and `classDirectory` (where we extract classes and copy resources) are different folders.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -661,7 +663,7 @@ object Defaults extends BuildCommon {
     classDirectory := target.value / (prefix(configuration.value.name) + "classes"),
     backendOutput := {
       val converter = fileConverter.value
-      val dir = classDirectory.value
+      val dir = target.value / (prefix(configuration.value.name) + "backend")
```
adpi2, line 666: `classDirectory` was:
  - the folder where we compile classes to
  - the folder where we zip classes from
  - the folder where we copy resources to
  - the folder where we unzip the compiled jar to
  - the folder where we zip classes and resources from
  
And this was a source of bugs: old resource files captured by the no-resource jar being unzipped on top of new resource files.

In practice I think it is simpler and safer if `backendOutput` (the output directory of the compiler) and `classDirectory` (where we extract classes and copy resources) are different folders.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2460,10 +2461,16 @@ object Defaults extends BuildCommon {
...
+        val store = analysisStore(compileAnalysisFile)
         val ci = (compile / compileInputs).value
+        val c = fileConverter.value
+        val dir = c.toPath(backendOutput.value).toFile
         result match
           case Result.Value(res) =>
+            val rawJarPath = c.toPath(res._2)
+            IO.delete(dir)
+            IO.unzip(rawJarPath.toFile, dir)
+            IO.delete(dir / "META-INF" / "MANIFEST.MF")
```
adpi2, line 2473: I am not a huge fan of this but we need to sync the class folder and the cached jar if in order to have incremental compilation.

Alternatively we could cache the folder instead of the jar, and the cache would take care of syncing it when needed. I would prefer this solution.

adpi2:
I should stop to modify this now. @eed3si9n feel free to review when you have time.

I think we should merge this sooner rather than later not to keep pending tests for too long. Also I would like to start merging changes from 1.x into 2.x soon.

eed3si9n (approved):

adpi2:
Thanks! I am still curious about your opinion on https://github.com/sbt/sbt/pull/7522#discussion_r1555559083, https://github.com/sbt/sbt/pull/7522#discussion_r1559447732 and https://github.com/sbt/sbt/pull/7522#discussion_r1559455960

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Generally I feel like more things we can virtualize to JAR would be better for remote caching. Resources are just JAR with text files in it.

PR #7525, opened 2024-04-05
[2.x] Remote caching support

eed3si9n:
This implements Bazel-compat remote caching support via a plugin. See https://eed3si9n.com/sbt-remote-cache-with-bazel-compat/ for details.


eed3si9n:
Most test failures seem to have the stacktrace of

```
[info] [error] java.lang.IllegalArgumentException: No companions was stored!
[info] [error] 	at sbt.internal.inc.binary.converters.ProtobufDefaults$EmptyLazyCompanions$.get(ProtobufDefaults.scala:87)
[info] [error] 	at sbt.internal.inc.binary.converters.ProtobufDefaults$EmptyLazyCompanions$.get(ProtobufDefaults.scala:86)
[info] [error] 	at xsbti.api.AnalyzedClass.api(AnalyzedClass.java:76)
[info] [error] 	at sbt.internal.inc.binary.converters.ProtobufWriters.toAnalyzedClass(ProtobufWriters.scala:625)
[info] [error] 	at sbt.internal.inc.binary.converters.ProtobufWriters.$anonfun$8(ProtobufWriters.scala:726)
[info] [error] 	at sbt.internal.inc.binary.converters.ProtobufWriters.toApis$$anonfun$2(ProtobufWriters.scala:729)
[info] [error] 	at scala.collection.immutable.Map$Map1.foreach(Map.scala:258)
[info] [error] 	at sbt.internal.inc.binary.converters.ProtobufWriters.toApis(ProtobufWriters.scala:729)
[info] [error] 	at sbt.internal.inc.binary.converters.ProtobufWriters.toApisFile(ProtobufWriters.scala:740)
[info] [error] 	at sbt.internal.inc.binary.BinaryAnalysisFormat.writeAPIs(BinaryAnalysisFormat.scala:39)
[info] [error] 	at sbt.internal.inc.FileAnalysisStore$BinaryFileStore.set$$anonfun$1(FileAnalysisStore.scala:97)
[info] [error] 	at scala.runtime.function.JProcedure1.apply(JProcedure1.java:15)
[info] [error] 	at scala.runtime.function.JProcedure1.apply(JProcedure1.java:10)
[info] [error] 	at sbt.io.Using.apply(Using.scala:28)
```

Potentially some incompatibility due to protobuf?

adpi2:
> Potentially some incompatibility due to protobuf?

I got the same issue while working on https://github.com/sbt/sbt/pull/7522, and fixed it in https://github.com/sbt/sbt/pull/7522/commits/1c03463e190f7e4e89c11df97a421b160260e543# by using a `MixedAnalyzingCompiler.staticCachedStore` instead of a `FileAnalysisStore.binary`. But then I removed that part of the code completely in https://github.com/sbt/sbt/pull/7522/commits/85fcb014f0ed34b822ae0841fee506b064cbafe1.

eed3si9n:
> > Potentially some incompatibility due to protobuf?
> 
> I got the same issue while working on #7522 

In https://github.com/sbt/zinc/pull/1347 I originally tried to backport "Bump protobuf-java 3.25.3" to Zinc 1.x and realized some tests failed, so probably the incompatibility between protobuf-java 3.25.3 (and ScalaPB?) is the issue, which I did upgrade in the Zinc 2.x branch.

adpi2:
Hey I rebased the changes to fix the conflicts and I force pushed. @eed3si9n You'll need to pull again if you want to change anything.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -4654,6 +4654,16 @@ trait BuildExtra extends BuildCommon with DefExtra {
       scalaBinaryVersion.value
     )
 
+  /**
+   * Adds Maven resolver plugin.
+   */
+  def addRemoteCachePlugin: Setting[Seq[ModuleID]] =
```
adpi2: This comment 
```suggestion
  /**
   * Adds remote cache plugin.
   */
  def addRemoteCachePlugin: Setting[Seq[ModuleID]] =
```

main/src/main/scala/sbt/RemoteCache.scala
```diff
@@ -113,6 +113,11 @@ object RemoteCache {
         DiskActionCacheStore(localCacheDirectory.value.toPath(), fileConverter.value)
       )
     },
+    remoteCache := SysProp.remoteCache,
+    remoteCacheTlsCertificate := SysProp.remoteCacheTlsCertificate,
+    remoteCacheTlsClientCertificate := SysProp.remoteCacheTlsClientCertificate,
+    remoteCacheTlsClientKey := SysProp.remoteCacheTlsClientKey,
+    remoteCacheHeaders := SysProp.remoteCacheHeaders,
```
adpi2, line 120: In general I think we should be careful in adding default settings and tasks. Loading big project matrices is already slow and adding more settings will make it slower.

Should we move those setting to the `RemoteCachePlugin`?

main/src/main/scala/sbt/internal/Aggregation.scala
```diff
@@ -68,44 +69,43 @@ object Aggregation {
...
+    cacheEventLog.clear()
+    val (newS, result) = withStreams(structure, s): str =>
       val transform = nodeView(s, str, roots, extra)
       runTask(toRun, s, str, structure.index.triggers, config)(using transform)
-    }
     val stop = System.currentTimeMillis
-    Complete(start, stop, result, newS)
-  }
+    val cacheSummary = cacheEventLog.summary
+    Complete(start, stop, result, cacheSummary, newS)
```
adpi2, line 108: I like this idea of a `cacheEventLog.summary`

util-cache/src/main/contraband/actionresult.contra
```diff
@@ -0,0 +1,17 @@
...
+## In addition to the value typically represented in the return type
+## of a task, ActionResult tracks the file output and other side effects.
+##
+## See also https://github.com/bazelbuild/remote-apis/blob/96942a2107c702ed3ca4a664f7eeb7c85ba8dc77/build/bazel/remote/execution/v2/remote_execution.proto#L1056
+type ActionResult {
+  outputFiles: [xsbti.HashedVirtualFileRef] @since("0.1.0")
+  origin: String @since("0.2.0")
+  exitCode: Int @since("0.3.0")
+  contents: [java.nio.ByteBuffer] @since("0.4.0")
+  isExecutable: [Boolean] @since("0.5.0")
```
adpi2, line 16: Why do we have those `@since("0.1.0")`? What do they mean?

util-cache/src/main/contraband/remotecache.contra
```diff
@@ -0,0 +1,16 @@
+package sbt.util
+@target(Scala)
+
+type UpdateActionResultRequest {
+  actionDigest: sbt.util.Digest!
+  outputFiles: [xsbti.VirtualFile] @since("0.1.0")
+  exitCode: Int @since("0.2.0")
+  isExecutable: [Boolean] @since("0.3.0")
```
adpi2, line 8: Do we really need those now? How hard would it be to add new fields later?

util-cache/src/main/contraband/remotecache.contra
```diff
@@ -0,0 +1,16 @@
...
+}
+
+type GetActionResultRequest {
+  actionDigest: sbt.util.Digest!
+  inlineStdout: Boolean @since("0.1.0")
+  inlineStderr: Boolean @since("0.1.0")
+  inlineOutputFiles: [String] @since("0.2.0")
```
adpi2, line 15: What is this `inlineOutputFiles`? What does it mean?

util-cache/src/main/scala/sbt/internal/util/codec/HashedVirtualRefFormats.scala
```diff
@@ -0,0 +1,29 @@
...
+import sjsonnew.{ BasicJsonProtocol, IsoString }
+import xsbti.HashedVirtualFileRef
+
+trait HashedVirtualFileRefFormats { self: BasicJsonProtocol =>
+
+  /**
+   * A string representation of HashedVirtualFileRef, delimited by `>`.
+   */
+  def hashedVirtualFileRefToStr(ref: HashedVirtualFileRef): String =
+    s"${ref.id}>${ref.contentHashStr}/${ref.sizeBytes}"
```
adpi2, line 15: Do they use the same formatting in Bazel?

util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
```diff
@@ -142,51 +180,80 @@ class DiskActionCacheStore(base: Path, fileConverter: FileConverter) extends Act
...
 
   override def putBlobs(blobs: Seq[VirtualFile]): Seq[HashedVirtualFileRef] =
     blobs.map: (b: VirtualFile) =>
-      val outFile = casBase.toFile / Digest(b.contentHashStr).toString
-      IO.transfer(b.input, outFile)
+      putBlob(b.input, Digest(b))
       (b: HashedVirtualFileRef)
 
-  override def getBlobs(refs: Seq[HashedVirtualFileRef]): Seq[VirtualFile] =
+  def toOutFile(digest: Digest): Path =
```
adpi2:
```suggestion
  def toCasFile(digest: Digest): Path =
```

adpi2 (commented):
Looks quite good overall!

main/src/main/scala/sbt/RemoteCache.scala
eed3si9n: I think the ones that we have to be careful is any keys that gets added at project level, like many of the plugins, like sbt-dependency-graph, which I think adds about a dozen keys, and in projects with 100 subprojects that ends up adding to initialization time. Adding a few hardcoded values at the global level is I think is almost free in that sense. I should change to `:==` which skips graph evaluation.

> Should we move those setting to the `RemoteCachePlugin`?

Here are some reasons to keep it in mothership:
1. Other plugins that want to implement remote backend can reuse this key without depending on sbt-remote-cache.
2. Build users can write the settings in `build.sbt`, but during the build time selectively ignore sbt-remote-cache (for example, in-house usage vs public).

adpi2: > I think the ones that we have to be careful is any keys that gets added at project level

Oh yes you are absolutely right

util-cache/src/main/contraband/actionresult.contra
eed3si9n: gRPC uses Protobuf, which allows API evolution by assuming optional and numbering each field. I'm using `@since("0.1.0")` etc here so Contraband will generate `apply(...)` functions:

```scala
  def apply(): ActionResult = new ActionResult()
  def apply(outputFiles: Vector[xsbti.HashedVirtualFileRef]): ActionResult = new ActionResult(outputFiles)
  def apply(outputFiles: Vector[xsbti.HashedVirtualFileRef], origin: Option[String]): ActionResult = new ActionResult(outputFiles, origin)
  def apply(outputFiles: Vector[xsbti.HashedVirtualFileRef], origin: String): ActionResult = new ActionResult(outputFiles, Option(origin))
```

Since I am saying that `outputFiles` were added in 0.1.0, prior to that there was nothing and thus it creates `apply()`. Next at the point of 0.1.0 it only had `outputFiles`, it creates `def apply(outputFiles: Vector[xsbti.HashedVirtualFileRef])`, and so on.

util-cache/src/main/contraband/remotecache.contra
eed3si9n: Yea, this is actually important to add this. I haven't gotten around to fully support this yet, but we need to add it here or maybe `xsbti.VirtualFile`, and one way or the other capture the executable flag, and test that both sbt can send the information and thaw it back correctly.

> How hard would it be to add new fields later?

Adding the field here won't be hard, but I want to make sure the build user can set this flag in a task, which might require some surgery.

util-cache/src/main/contraband/remotecache.contra
eed3si9n: This is from https://github.com/bazelbuild/remote-apis/blob/96942a2107c702ed3ca4a664f7eeb7c85ba8dc77/build/bazel/remote/execution/v2/remote_execution.proto#L1582:
> A hint to the server to inline the contents of the listed output files.

Using our `ActionCacheStore` lingo, to get a task result even for Int `1`, we have to make two roundtrip calls: `ActionCacheStore#get` to get the reference to CAS entry of JSON, and `ActionCacheStore#syncBlobs` to download JSON from CAS. But it is inefficient in multiple ways:
- Second roundtrip just for `1`
- File creation on the laptop just for `1`

So `GetActionResultRequest` has an optimization that can optionally piggyback the content of "outputFile" as bytearray encoded into gRPC response. If the blob is too big, the server has the right to ignore it, but for most task results we should ask to inline, and check if the result exists, which is what `ActionCache.cache` does now:

```scala
        // some protocol can embed values into the result
        result.contents.headOption match
```

util-cache/src/main/scala/sbt/internal/util/codec/HashedVirtualRefFormats.scala
eed3si9n: Nope. As far as I know, there isn't a string representation for this. The closest concept is `OutputFile` - https://github.com/bazelbuild/remote-apis/blob/96942a2107c702ed3ca4a664f7eeb7c85ba8dc77/build/bazel/remote/execution/v2/remote_execution.proto#L1229-L1254

adpi2 (approved):

PR #7537, opened 2024-04-16
[2.x] Use `fileConverter` in `GrpcActionCacheStore`

adpi2:
Since #7522 I am using the `fileConverter` in `ActionCacheStore` impls to be able to sync files that are outside of the `out` directory. This was needed by some scripted tests. Otherwise the cache would create files such as `target/out/${BASE}/foo`.

### DIFF
--- a/sbt-remote-cache/src/main/scala/sbt/internal/GrpcActionCacheStore.scala
+++ b/sbt-remote-cache/src/main/scala/sbt/internal/GrpcActionCacheStore.scala
@@ -48,6 +48,7 @@ object GrpcActionCacheStore:
       clientCertChain: Option[Path],
       clientPrivateKey: Option[Path],
       remoteHeaders: List[String],
+      converter: FileConverter,
       disk: DiskActionCacheStore,
   ): GrpcActionCacheStore =
     val b: ManagedChannelBuilder[?] = uri.getScheme() match
@@ -81,7 +82,7 @@ object GrpcActionCacheStore:
       case Some(x) if x.startsWith("/") => x.drop(1)
       case Some(x)                      => x
       case None                         => ""
-    new GrpcActionCacheStore(channel, instanceName, remoteHeaders, disk)
+    new GrpcActionCacheStore(channel, instanceName, remoteHeaders, converter, disk)
 
   class AuthCallCredentials(remoteHeaders: List[String]) extends CallCredentials:
     val pairs = remoteHeaders.map: h =>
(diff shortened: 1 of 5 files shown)


sbt-remote-cache/src/main/scala/sbt/internal/GrpcActionCacheStore.scala
```diff
@@ -185,22 +187,19 @@ class GrpcActionCacheStore(
...
-        val digest = Digest(r)
+      refs.map: ref =>
+        val digest = Digest(ref)
         val blob = lookupResponse(digest)
         val casFile = disk.putBlob(blob.getData().newInput(), digest)
-        val shortPath =
-          if r.id.startsWith("${OUT}/") then r.id.drop(7)
-          else r.id
-        val outPath = outputDirectory.resolve(shortPath)
+        val outPath = converter.toPath(ref)
```
eed3si9n, line 202: Wouldn't this allow remote files to be placed at an arbitrary location? The contract of the cache (remote or disk) should be that all files would be placed in some known CAS and symlink pointing to it.

adpi2: > Wouldn't this allow remote files to be placed at an arbitrary location?

Yes it would, and I see how it can be problematic, if the file is a symlink and it is committed to git.

But we don't have any guarantee that the declared output file of a task is in the out directory. Any user can redefine the `target` directory to wherever they want but the cache won't be able to restore it. It will create a non-sense `target/out/${BASE}/custom-target` instead.

So what should we do:
1. Prevent the cache from storing any file that is outside the output directory? Silently or with a warning?
2. Cache any file but if it is outside the `out` directory, we copy it instead of creating a symlink?

eed3si9n: If possible we should try 1 I think, and fail the cache function.

eed3si9n: Also if the `target` is customized, then maybe out should be customized to be under it anyway?

adpi2: > Also if the target is customized, then maybe out should be customized to be under it anyway?

When I say `target` I mean the `target` task of a project:
```
target/out/$platform/scala-$scalaVersion/$moduleName
```

adpi2: > If possible we should try 1 I think, and fail the cache function.

Like we throw an exception? Or we just don't cache anything as if nothing happened.

eed3si9n: Maybe case by case, but throwing exception probably makes sense to prevent the build user thinking something is cached when it's not.

adpi2:
Closing in favor of #7539

Issue #6512, opened 2021-05-16
`forceUpdatePeriod` doesn't work with non-fresh sbt run


unkarjedy:
relates to [Add forceUpdatePeriod](https://github.com/sbt/sbt/pull/2017)
@ajsquared 

# Steps
sbt.version = `1.5.2`

### Part 1 (publish some `SNAPSHOT` library)
1. Create some dummy sbt project as a library.
Note, that it should be published to some artifactory (thus no via `publishLocal`)
```scala
name := "my-library-3"

ThisBuild / scalaVersion := "2.13.5"
ThisBuild / organization := "org.example"
ThisBuild / version      := "0.1-SNAPSHOT"

// Configure your own repository here
credentials += Credentials(Path.userHome / ".sbt" / "my.credentials")
resolvers += MyResolvers.myRepository
publishTo := Some(MyResolvers.myRepository)
```
2. In the library sources create some test object with method `foo1`
```scala
object MyObject3 {
  def foo1(): Unit = {
    println(s"hello from: ${this.getClass}")
  }
}
```

3. Publish the library with `publish` command

### Part 2 (use the library in another project)

4. Create another sbt project which will use the library:
```scala
import scala.concurrent.duration.DurationInt

name := "library-reindex"
version := "0.1"
scalaVersion := "2.13.5"

ThisBuild / forceUpdatePeriod := Some(0.seconds) // doesn't help :(

libraryDependencies ++= Seq(
  "org.example" %% "my-library-3" % "0.1-SNAPSHOT" /*withSources()*/,
)

myTask2 := myTaskDef2.value

// TODO: provide same resolvers as in library project
credentials += ???
resolvers += ???
```

5. In this project create some task which depends on `update` task:
in `build.sbt` add:
```scala
myTask2 := myTaskDef2.value
```
in `project/commons.scala`:
```scala
import org.jetbrains.sbt.CreateTasks.`enrich TaskKey`
import org.jetbrains.sbt.{Options, StructureKeys, UpdateReportAdapter}
import sbt.Def.Initialize
import sbt._

object commons {
  def myTaskDef2: Initialize[Task[String]] = Def.task {
    val result = Keys.update.value
    println(s"! $result")
    "done task 2"
  }
}
```

6. Run `sbt` for the second project

7. Run `update` just in case

8. Ensure that the library was resolved by courier. 
Go to something like:
[...]

PR #7538, opened 2024-04-16
[2.x] Fix `dependency-management/force-update-period`

adpi2:
This scripted test was actually not testing what it was supposed to test. And so the `forceUpdatePeriod` is also broken in sbt 1.x (https://github.com/sbt/sbt/issues/6512). I'll backport the fix.


eed3si9n (approved):
Thanks!

PR #7539, opened 2024-04-17
[2.x] Report error if output file of a cached task is not in the output directory

adpi2:
In the `ActionCache`, we check that all output files are in the output directory. If they are not, we still run the task, but we print an error message saying that the task could not be cached. Also the cache summary details the number of errors.

```
sbt:root> compile
[success] elapsed time: 0 s, cache 0%, 1 onsite task
sbt:root> set root / target := baseDirectory.value / "custom-target"
...
sbt:root> compile
[error] Cannot cache task because its output files are outside the output directory: 
[error]   - ${BASE}/custom-target/zinc/inc_compile_3.zip
[error]   - ${BASE}/custom-target/sbt2-test_3-0.1.0-SNAPSHOT-noresources.jar
[success] elapsed time: 0 s, cache 0%, 1 error
sbt:root>
```

### DIFF
--- a/main-settings/src/main/scala/sbt/Def.scala
+++ b/main-settings/src/main/scala/sbt/Def.scala
@@ -23,6 +23,7 @@ import sbt.util.Show
 import xsbti.{ HashedVirtualFileRef, VirtualFile }
 import sjsonnew.JsonFormat
 import scala.reflect.ClassTag
+import xsbti.FileConverter
 
 /** A concrete settings system that uses `sbt.Scope` for the scope type. */
 object Def extends Init[Scope] with TaskMacroExtra with InitializeImplicits:
@@ -233,6 +234,7 @@ object Def extends Init[Scope] with TaskMacroExtra with InitializeImplicits:
   private[sbt] var _cacheStore: ActionCacheStore = InMemoryActionCacheStore()
   def cacheStore: ActionCacheStore = _cacheStore
   private[sbt] var _outputDirectory: Option[Path] = None
+  private[sbt] var _fileConverter: Option[FileConverter] = None
   private[sbt] val cacheEventLog: CacheEventLog = CacheEventLog()
   def cacheConfiguration: BuildWideCacheConfiguration =
     BuildWideCacheConfiguration(
--- a/main-settings/src/main/scala/sbt/Def.scala
+++ b/main-settings/src/main/scala/sbt/Def.scala
@@ -7,7 +7,6 @@
 
 package sbt
 
-import java.nio.file.Path
 import java.net.URI
 
 import scala.annotation.tailrec
@@ -17,13 +16,12 @@ import sbt.Scope.{ GlobalScope, ThisScope }
 import sbt.internal.util.Types.const
 import sbt.internal.util.complete.Parser
 import sbt.internal.util.{ Terminal => ITerminal, * }
-import sbt.util.{ ActionCacheStore, BuildWideCacheConfiguration, InMemoryActionCacheStore }
+import sbt.util.{ ActionCacheStore, AggregateActionCacheStore, BuildWideCacheConfiguration, cacheLevel , DiskActionCacheStore }
 import Util._
 import sbt.util.Show
 import xsbti.{ HashedVirtualFileRef, VirtualFile }
 import sjsonnew.JsonFormat
 import scala.reflect.ClassTag
-import xsbti.FileConverter
 
 /** A concrete settings system that uses `sbt.Scope` for the scope type. */
 object Def extends Init[Scope] with TaskMacroExtra with InitializeImplicits:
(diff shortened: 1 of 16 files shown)


main/src/main/scala/sbt/RemoteCache.scala
```diff
@@ -50,27 +50,6 @@ object RemoteCache {
   private[sbt] val analysisStore: mutable.Map[HashedVirtualFileRef, CompileAnalysis] =
     mutable.Map.empty
 
-  // TODO: figure out a good timing to initialize cache
-  // currently this is called twice so metabuild can call compile with a minimal setting
-  private[sbt] def initializeRemoteCache(s: State): Unit =
```
adpi2, line 55: The remote cache and the `BuildWideCacheConfiguration` are now initialized by the task engine (as a task but we can make the `CacheStore` a setting if we need to persist the instance).

main-settings/src/main/scala/sbt/Def.scala
```diff
@@ -229,17 +228,40 @@ object Def extends Init[Scope] with TaskMacroExtra with InitializeImplicits:
...
+    )
+
   // These are here, as opposed to RemoteCache, since we need them from TaskMacro etc
-  private[sbt] var _cacheStore: ActionCacheStore = InMemoryActionCacheStore()
-  def cacheStore: ActionCacheStore = _cacheStore
-  private[sbt] var _outputDirectory: Option[Path] = None
   private[sbt] val cacheEventLog: CacheEventLog = CacheEventLog()
-  def cacheConfiguration: BuildWideCacheConfiguration =
+  @cacheLevel(include = Array.empty)
+  val cacheConfiguration: Initialize[Task[BuildWideCacheConfiguration]] = Def.task {
```
eed3si9n, line 249: Good idea.

util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -37,30 +37,47 @@ object ActionCache:
...
+            throw e
+      val json = Converter.toJsonUnsafe(result)
+      val uncacheableOutputs =
+        outputs.map(fileConverter.toPath).filter(f => !f.startsWith(outputDirectory))
+      if uncacheableOutputs.nonEmpty then
+        cacheEventLog.append(ActionCacheEvent.Error)
+        logger.error(
+          s"Cannot cache task because its output files do not start with $outputDirectory: \n" +
+            uncacheableOutputs.mkString("  - ", "\n  - ", "")
+        )
```
eed3si9n: This is more cosmetic thing, but I personally don't like build tools leaking out the absolute paths in the log messages, since it could include information like name or company that the build user may not want to publicize. Some of the existing messages include them, but we should avoid that for new entries esp. remote cache.

adpi2: Updated message is:
```
sbt:sbt2-test> compile
[error] Cannot cache task because its output files are outside the output directory: 
[error]   - ${BASE}/custom-target/zinc/inc_compile_3.zip
[error]   - ${BASE}/custom-target/sbt2-test_3-0.1.0-SNAPSHOT-noresources.jar
[success] elapsed time: 0 s, cache 0%, 1 error
```

adpi2:
I am not sure why but Java Adopt 8 does not seem to be available on the latest MacOS runner anymore:

```
Run actions/setup-java@v3
Installed distributions
  Trying to resolve the latest version from remote
  Error: Could not find satisfied version for SemVer '8'. 
  Available versions: 21.0.3+9.0.LTS, 21.0.2+13.0.LTS, 21.0.1+12.0.LTS, 21.0.0+35.0.LTS, 20.0.2+9, 20.0.1+9, 20.0.0+36, 19.0.2+7, 19.0.1+10, 19.0.0+36, 18.0.2+101, 18.0.2+9, 18.0.1+10, 18.0.0+36, 17.0.11+9, 17.0.10+7, 17.0.9+9, 17.0.8+101, 17.0.8+7, 17.0.7+7, 17.0.6+10, 17.0.5+8, 17.0.4+101, 17.0.4+8, 17.0.3+7, 17.0.2+8, 17.0.1+12, 17.0.0+35, 11.0.23+9, 11.0.[22](https://github.com/sbt/sbt/actions/runs/8813675838/job/24239977527?pr=7539#step:6:23)+7.1, 11.0.22+7, 11.0.21+9, 11.0.20+101, 11.0.20+8, 11.0.19+7, 11.0.18+10, 11.0.17+8, 11.0.16+101, 11.0.16+8, 11.0.15+10
```

eed3si9n:
I guess as long as one of the tests use JDK 8 to make sure we can build, the others can use JDK 17 or whatever?

adpi2:
So it seems the latest MacOS runner bumped from [macos-12](https://github.com/actions/runner-images/blob/macOS-12/20240412.2/images/macos/macos-12-Readme.md) to [macos-14-arm64](https://github.com/actions/runner-images/blob/macOS-12/20240412.2/images/macos/macos-14-arm64-Readme.md) which is why Java 8 and Python 3.7 are not available anymore.

I tried bumping to Java 11 and Python 3.12 but the [job failed](https://github.com/sbt/sbt/actions/runs/8877820398/job/24372176297). I rolled back to macos-12 instead.

adpi2:
It's ready to review

eed3si9n (approved):
Thanks!

PR #7549, opened 2024-05-05
Fix CI

eed3si9n:
It seems like the latest macOS image no longer supports JDK 8.


xuwei-k:
- https://github.com/actions/setup-java/issues/625

PR #7559, opened 2024-05-15
[2.x] Add `Def.cacheTaskWithUpdate` to update the task output after a cache reuse

adpi2:
After a cache hit of the incremental compilation we need to:
  - Unzip the jar to the backend output folder to enable incremental compilation.
  - Update the timestamps. Those timestamps are used by the `testQuickFilter` to decide which class needs to be rerun.

To do so I introduced:
```scala
Def.cachedTaskWithUpdate[A1: JsonFormat](inline a1: A1)(inline onCacheReuse: (A1 => A1)): Def.Initialize[Task[A1]]
```

The `onCacheReuse` is only called after a cache hit to update the output. The user can declare any task dependency in `onCacheReuse` with `.value`.

Example:

```scala
val foo = Def.task(1)
val bar = Def.task(2)
val fizz = Def.cachedTaskWithUpdate(foo.value)(res => res + bar.value)
```

The expansion of `fizz` looks like this:
```scala
val fizz = (foo, bar).mapN { (foo, bar) => // both foo and bar are dependencies of the task
  Def.cachedTask(input = foo) {            // only foo is used as an input
    action: _ => foo
    onCacheReuse: () => res => res + bar
  }
}
```

The behavior of fizz is:
```scala
sbt:example> show fizz
1 // computed and cached
sbt:example> show fizz
3 // taken from the cache and updated (1 + 2) 
sbt:example> show fizz
3 // taken from the cache and updated (1 + 2) 
```


eed3si9n:
> After a cache hit of the incremental compilation we need to:
>
> - Unzip the jar to the backend output folder to enable incremental compilation.
> - Update the timestamps. Those timestamps are used by the `testQuickFilter` to decide which class needs to be rerun.

If possible, I feel like we should try to fix the incremental compilation and `testQuick` so they can work with the remote downloaded outputs instead. Even if we're introducing some side effects, I feel like it should be narrowly-scoped (maybe it should be `Seq[Path] => Unit`), and should always run to avoid potentially confusing results if misused/abused.

adpi2:
> If possible, I feel like we should try to fix the incremental compilation and testQuick so they can work with the remote downloaded outputs instead.

It would be ideal. But also it means we cannot rely on timestamps in `CompileAnalysis` anymore, which could potentially break some plugins/projects. It's probably very limited though.

For incremental compilation, I think we can configure Zinc to always compile to a jar file. But I don't know how good that works, or if it has already been battle-tested.

For `testQuick`, if we cannot use the timestamps, we can still use the hashes. But that mean we need to compute the hash of each test even after a `test` which is a bit of overhead.

> Even if we're introducing some side effects, I feel like it should be narrowly-scoped (maybe it should be Seq[Path] => Unit), and should always run to avoid potentially confusing results if misused/abused.

Yeah, I thought of that too and `Seq[Path] => Unit` is probably harder to abuse. I also wonder if `Unit` would not be even better (no input and pure side effect). Those `Seq[Path]` are opaque and it would be dangerous to guess which path is which output file. We can still use `.value` to find the file that we need to update.

eed3si9n:
> For incremental compilation, I think we can configure Zinc to always compile to a jar file. But I don't know how good that works, or if it has already been battle-tested.

I think there's one build tool that regularly use compile-to-JAR, but if I recall correctly, it doesn't clean after itself to make deletion faster or something like that.

One idea I think worth perusing might be implementing `Def.declareOutputDirectoryZip(...)` where we pass in a zip file of `classes/` directory as `classes.zip`, and by convention we'll unzip `classes.zip` to `classes/`. This way, we can reuse the caching mechanism of single file (caching a whole directory can be complicated because we'd have to check hash of hundreds of files potentially), while keeping the current compilation semantics.

adpi2:
Thanks! I started looking into the compile to JAR option last week. But if I don't manage to make it work well enough, I'll go ahead with `declareOutputDirectoryZip`.

adpi2:
> I think there's one build tool that regularly use compile-to-JAR, but if I recall correctly, it doesn't clean after itself to make deletion faster or something like that.

I managed to make it work in sbt but as you said it does not clean after itself, which is an issue for me. Do you remember any discussion about that?

I checked Bloop, Mill, Maven and Gradle, and I don't think any of those compile to a jar file by default.

EDIT: I think I found relevant information in https://github.com/sbt/zinc/pull/597

RE EDIT: Actually it's worse than I expected: it only grows the size of the JAR. Meaning if we need to change `A.class`, it adds the new file in the JAR and updates the index, but it never removes the old one. That's quite bad for caching.

PR #7568, opened 2024-05-21
[1.10.x] Fix declaring `Optional` inter-project dependency in BSP

adpi2:
```scala
val a = project.in(file("a"))

val b = project.in(file("b")).dependsOn(a % Optional)
```

In BSP there is no notion of `Optional` or `Provided` dependencies. But the `b#Compile` build target should contain a dependency to `a#Compile` because it needs it to compile. (The classphath configuration is `CompileInternal` which extends `Optional` and `Provided`).

To fix this we use the `internalConfigurationMap` to compute the internal dependencies of a configuration.

@unkarjedy  This PR should fix https://youtrack.jetbrains.com/issue/SCL-22476/SBTBSP-build-with-an-optional-inter-project-dependency-results-in-cannot-resolve-symbol-errors

### DIFF
--- a/main/src/main/scala/sbt/internal/InternalDependencies.scala
+++ b/main/src/main/scala/sbt/internal/InternalDependencies.scala
@@ -13,7 +13,9 @@ import sbt.Keys._
 
 private[sbt] object InternalDependencies {
   def configurations: Def.Initialize[Seq[(ProjectRef, Set[String])]] = Def.setting {
-    val allConfigs = Classpaths.allConfigs(configuration.value).map(_.name).toSet
+    val configMap = internalConfigurationMap.value
+    val config = configMap(configuration.value)
+    val allConfigs = Classpaths.allConfigs(config).map(_.name).toSet
     val ref = thisProjectRef.value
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/internal/InternalDependencies.scala
```diff
@@ -13,7 +13,9 @@ import sbt.Keys._
 
 private[sbt] object InternalDependencies {
   def configurations: Def.Initialize[Seq[(ProjectRef, Set[String])]] = Def.setting {
-    val allConfigs = Classpaths.allConfigs(configuration.value).map(_.name).toSet
+    val configMap = internalConfigurationMap.value
+    val config = configMap(configuration.value)
+    val allConfigs = Classpaths.allConfigs(config).map(_.name).toSet
```
adpi2, line 18: This is used by BSP and by watch. Which means that modifying `a` will trigger the watch of `~b/compile` which seems correct since `a` is in the compilation classpath of `b`.

eed3si9n (approved):

PR #7592, opened 2024-06-26
[2.x] compile straight-to-jar

adpi2:
This PR enables compilation to JAR by default in sbt 2.x, so that we can directly cache the compiled JAR without the annoying zip/unzip loop.

Several versions of the Scala 3 compiler do not support straight-to-JAR compilation:
- Between 3.0.0 and 3.2.x included, it crashes the compiler
- Between 3.3.0 and 3.5.0 included, it's compiling but the `analysis.products` is empty which causes unnecessary invalidation
- It's fixed in 3.5.1 and should be backported in 3.3.5 (https://github.com/scala/scala3/pull/20592)

I think it is a manageable situation because, by the time we release sbt 2.x, everyone should be using 3.3.5+ or 3.5+.


adpi2:
~~Some of the tests are broken because the Scala 3 compiler does not declare any products when compiling to jar, which I fixed in https://github.com/scala/scala3/pull/20592. Those products are needed for loading auto-plugins, and also for incremental compilation of downstream modules. We will have to wait for a backport and release of 3.3.5 to fix those tests.~~

EDIT: I updated to the latest compiler nightly to continue working

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2571,19 +2562,12 @@ object Defaults extends BuildCommon {
...
-        .toSeq
-      // inlined to avoid caching mappings
-      val pkgConfig = Pkg.Configuration(mappings, artifactPath.value, packageOptions.value)
-      val out = Pkg(pkgConfig, c, s.log, Pkg.timeFromConfiguration(pkgConfig))
-      s.log.debug(s"wrote $out")
+      val outputJar = ci.options.classesDirectory
+      // TODO remove: create output jar if missing
+      if !Files.exists(outputJar) then
+        val zip = new ZipOutputStream(new FileOutputStream(outputJar.toFile))
+        zip.close()
```
adpi2, line 2569: This should be removed once we release https://github.com/sbt/zinc/pull/1368

main/src/main/scala/sbt/IO2.scala
```diff
@@ -0,0 +1,39 @@
...
+        .walk(root)
+        .iterator
+        .asScala
+        .foreach { entry =>
+          val target = dirPath.resolve(root.relativize(entry).toString)
+          if !Files.isDirectory(target) then Files.copy(entry, target)
+          if Files.isRegularFile(target) then set += target.toFile
+        }
+    }
+    set.toSet
```
adpi2, line 28: We cannot use the `IO.unzip` method because the JAR file produced by Zinc contains all old and new entries since the creation of the JAR.

Work is being done in Zinc to fix this.

project/Dependencies.scala
```diff
@@ -6,7 +6,7 @@ object Dependencies {
   // WARNING: Please Scala update versions in PluginCross.scala too
   val scala212 = "2.12.19"
   val scala213 = "2.13.12"
-  val scala3 = "3.3.1"
+  val scala3 = "3.5.1-RC1-bin-20240628-1efbb92-NIGHTLY"
```
adpi2, line 9: Using a nightly version because of https://github.com/scala/scala3/pull/20592, and we need the Zinc products to find the auto plugins.

Eventually we should use `3.3.5` or `3.5.1`.

sbt-app/src/sbt-test/source-dependencies/java-mixed/pending
```diff
@@ -1,3 +1,5 @@
+# Marked pending until sbt/zinc#1376 is released
+
 $ copy-file changes/S1.scala S.scala
 > compile
 $ copy-file changes/S2.scala S.scala
```
adpi2, line 5: This scripted test is failing because of https://github.com/sbt/zinc/pull/1376

adpi2:
@eed3si9n this is definitelty not ready for merging but it is ready for a review if you have time.

PR #7604, opened 2024-07-19
Fix compilation on JDK21

SlowBrainDude:
Deprecation warnings combined with "fatal warnings" caused compilation to fail on current JDKs.

This commit removes some attempts to trigger finalization of some objects. However, there are no finalizers defined in the sbt project itself. We can't do more here than remove the deprecated JDK feature, and hope that upstream projects have also replaced their use of finalizers with future-proof alternatives by now.

I think this is the only way forward. Also we're effectively done here as there are no finalizers to migrate.

I'm not sure though what priority it has to be able to compile SBT on newer JDKs.
The build of SBT itself welcomed me with
```
[info] !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
[info]   Java version is 21. We recommend java 8.
[info] !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
which I first didn't notice as it looked like a part of the ASCII art banner directly above.

Than, I thought this is a kind of joke. Like finding a website which recommends Internet Explorer 5.5 for best user experience. :smile: 

I mean, JDK8 is really dated. A lot of projects (also in the Scala build department!) require at least Java 17, and version 21 is the current LTS, which I would actually target for the build environment.

### DIFF
--- a/main/src/main/scala/sbt/internal/AbstractTaskProgress.scala
+++ b/main/src/main/scala/sbt/internal/AbstractTaskProgress.scala
@@ -122,7 +122,7 @@ private[sbt] abstract class AbstractTaskExecuteProgress extends ExecuteProgress[
 object AbstractTaskExecuteProgress {
   private[sbt] class Timer() {
     val startNanos: Long = System.nanoTime()
-    val threadId: Long = Thread.currentThread().getId
+    val threadId: Long = Thread.currentThread().threadId
     var endNanos: Long = 0L
     def stop(): Unit = {
       endNanos = System.nanoTime()
--- a/main/src/main/scala/sbt/internal/GCUtil.scala
+++ b/main/src/main/scala/sbt/internal/GCUtil.scala
@@ -11,7 +11,6 @@ package internal
 
 import java.util.concurrent.atomic.AtomicLong
 import scala.concurrent.duration._
-import scala.util.control.NonFatal
 import sbt.util.Logger
 
 private[sbt] object GCUtil {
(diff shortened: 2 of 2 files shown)


eed3si9n:
> I mean, JDK8 is really dated. A lot of projects (also in the Scala build department!) require at least Java 17, and version 21 is the current LTS, which I would actually target for the build environment.

sbt is a build tool, which is near the bottom of the language tooling stack. Given that Scala compilers 2.x and 3.x currently support JDK 8 it makes sense as sbt to support JDK 8 as well. If you want to discuss dropping JDK 8 support for all sbt users, maybe you should bring it up to https://contributors.scala-lang.org/?

main/src/main/scala/sbt/internal/AbstractTaskProgress.scala
```diff
@@ -122,7 +122,7 @@ private[sbt] abstract class AbstractTaskExecuteProgress extends ExecuteProgress[
 object AbstractTaskExecuteProgress {
   private[sbt] class Timer() {
     val startNanos: Long = System.nanoTime()
-    val threadId: Long = Thread.currentThread().getId
+    val threadId: Long = Thread.currentThread().threadId
```
eed3si9n, line 125: `threadId` was added in JDK 19, so this won't work on JDK 8, 11, or 17?

eed3si9n (changes requested):

main/src/main/scala/sbt/internal/AbstractTaskProgress.scala
SlowBrainDude: Oh! You're right: `@since 19`.

Dam, this won't work like that… :disappointed:

SlowBrainDude:
I didn't consider dropping JDK 8 runtime support. That's not possible currently as Scala supports it. I'm aware of that.

I was more thinking about _the environment in which SBT gets build_.

Like I said, I don't know whether there is any interest to make SBT build on current JDKs. But I usually think that builds should track the latest environments as this is often what developers run on their boxes.

That's of course just my opinion. But if there is a hard requirement on which JDK version to use for SBT development this should be documented I think. Just this banner isn't helpful. (I've looked for docs but there is nothing. So I assumed it should be fine with the latest LTS JDK). 

SBT can of course still compile to old byte-code for old runtime environments. So it can be used on old JDKs.

So to make my IDE running on JDK21 happy this needs likely just a warning suppression annotation. Let's try this.

Would you be OK with such annotations?

Actually I want to fix a bug… But the IDE complained that it can't do the initial import… :smiling_face_with_tear:

SlowBrainDude:
OK, now I think, just downgrading my JDK makes most sense here.

I'm closing this as it was the wrong approach, and add a note to the DEV docs on the requirement to not try to run the build on anything beyond JDK17.

I don't like warning suppression annotations. That should be the last resort, and only used when unavoidable (which isn't the case here). Better have the warnings when you want to upgrade for real.

Still not completely happy here. Metals downloaded some Java I didn't ask is to do… Also one needs to change the version for Metals globally, there is no workspace setting… It's so much smoother if at least a build runs on a current setup. Now I've spend time making build and IDE work instead of bug hunting. But to be honest, I had already a gut feeling that the build of SBT will be more challenging then the usual Scala project. :smile:

PR #7606, opened 2024-07-19
Note in DEV docs on supported build time JDK version dependency

SlowBrainDude:
I've written a little bit about the current requirement to use at most JDK 17 for development.

I hope this is will make it clearer directly from the docs what is needed to build SBT itself.


eed3si9n (approved):
Thanks!

PR #7621, opened 2024-08-12
[2.x] Def.declareOutputDirectory

eed3si9n:
See https://eed3si9n.com/sudori-part4/ for details.

This implements a new output `Def.declareOutputDirectory(...)`, which is mostly transparent to the remote cache backend. When `ActionCache.cache` encounters a `Def.declareOutputDirectory(...)`, it creates a zip file with a manifest JSON listing all the items and their hashes. On the receiving side, when it comes across a file ending with `.sbtdir.zip` it will sync the content as if individual items were cached.


adpi2:
Is this ready to be reviewed?

eed3si9n:
I'm trying to implement compile task using it as a proof of concept, but you can start reviewing if you would like.

util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -75,21 +82,79 @@ object ActionCache:
...
+  class InternalActionResult[A1] private (
+      val value: A1,
+      val outputs: Seq[VirtualFile],
+  )
+  end InternalActionResult
+  object InternalActionResult:
+    def apply[A1](value: A1, outputs: Seq[VirtualFile]): InternalActionResult[A1] =
+      new InternalActionResult(value, outputs)
+    private[sbt] def unapply[A1](r: InternalActionResult[A1]): Option[(A1, Seq[VirtualFile])] =
+      Some(r.value, r.outputs)
```
adpi2, line 161: Since the `unapply` is `private[sbt]` and not call from a macro, I think it is safe to return a tuple instead of an option. We can even declare it `private[ActionCache]` for extra security.

sbt-app/src/sbt-test/actions/cross-advanced/test
```diff
@@ -5,6 +5,7 @@
 
 ## test scoped task
 ## this should not force any Scala version changes to other subprojects
+> debug
```
adpi2, line 8: What's that?

util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -75,21 +82,79 @@ object ActionCache:
...
+      val mPath = (tempDir / manifestFileName).toPath()
+      makeManifest(mPath)
+      val zipPath = Paths.get(dirPath.toString + dirZipExt)
+      val rebase: Path => Seq[(File, String)] =
+        (p: Path) =>
+          p match
+            case p if p == dirPath => Nil
+            case p if p == mPath   => (mPath.toFile() -> manifestFileName) :: Nil
+            case f                 => (f.toFile() -> outputDirectory.relativize(f).toString) :: Nil
+      IO.zip((allPaths.map(_._1) ++ Seq(mPath)).flatMap(rebase), zipPath.toFile(), None)
```
adpi2: Do we need to keep directories in `allPaths` when we call `IO.zip`? If not maybe `allPaths` should contain only files.

util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
```diff
@@ -253,16 +265,70 @@ class DiskActionCacheStore(base: Path) extends AbstractActionCacheStore:
...
+   * Emulate virtual side effects.
+   */
+  def afterFileWrite(ref: HashedVirtualFileRef, path: Path, outputDirectory: Path): Unit =
+    if path.toString().endsWith(ActionCache.dirZipExt) then unpackageDirZip(path, outputDirectory)
+    else ()
+
+  /**
+   * Given a dirzip, unzip it in a temp directory, and sync each items to the outputDirectory.
+   */
+  private def unpackageDirZip(dirzip: Path, outputDirectory: Path): Path =
```
adpi2, line 295: If it's the responsibility of `ActionCache` to package the directory, then I don't think it should delegate the unpackaging to the store.

util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
```diff
@@ -253,16 +265,70 @@ class DiskActionCacheStore(base: Path) extends AbstractActionCacheStore:
...
+      IO.unzip(dirzip.toFile(), tempDir)
+      val mPath = (tempDir / ActionCache.manifestFileName).toPath()
+      if !Files.exists(mPath) then sys.error(s"manifest is missing from $dirzip")
+      // manifest contains the list of files in the dirzip, and their hashes
+      val m = ActionCache.manifestFromFile(mPath)
+      m.outputFiles.foreach: ref =>
+        val shortPath =
+          if ref.id.startsWith("${OUT}/") then ref.id.drop(7)
+          else ref.id
+        val currentItem = outputDirectory.resolve(shortPath)
```
adpi2, line 318: It seems wrong to do that again and again. Cache stores need to convert from `VirtualFileRef` to `Path` and they should use the converter to do so.

The `ActionCache` already does the check to prevent any leak outside the output directory.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2571,21 +2556,11 @@ object Defaults extends BuildCommon {
...
-      val pkgConfig = Pkg.Configuration(mappings, artifactPath.value, packageOptions.value)
-      val out = Pkg(pkgConfig, c, s.log, Pkg.timeFromConfiguration(pkgConfig))
-      s.log.debug(s"wrote $out")
-      Def.declareOutput(out)
-      analysisResult.hasModified() -> (out: HashedVirtualFileRef)
+      val dir = ci.options.classesDirectory
+      val vfDir = c.toVirtualFile(dir)
+      val packedDir = Def.declareOutputDirectory(vfDir)
+      s.log.debug(s"wrote $vfDir")
+      (analysisResult.hasModified(), vfDir: VirtualFileRef, packedDir: HashedVirtualFileRef)
```
adpi2, line 2566: What's the difference between `vfdir` and `packedDir`?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2732,6 +2707,7 @@ object Defaults extends BuildCommon {
           inputs.options.sources.toVector,
           scalacOptions.value.toVector,
           javacOptions.value.toVector,
+          outputPath.value + prefix(configuration.value.name),
```
adpi2: To be even more precise, I think we should replace this with:

```suggestion
          c.toVirtualFile(inputs.options.classesDirectory)
```

If the class directory changes, we should re-compile to a new class directory.

adpi2: That's what I did in https://github.com/sbt/sbt/pull/7592/commits/9c2a36d95fc377f5366e6194fa9dfd25d7783cd7

server-test/src/server-test/buildserver/build.sbt
```diff
@@ -31,9 +31,7 @@ lazy val respondError = project.in(file("respond-error"))
     }
   )
 
-lazy val util = project.settings(
-  Compile / target := baseDirectory.value / "custom-target",
-)
+lazy val util = project
```
adpi2: I think we should keep this. Otherwise it might be too hard for some builds to migrate to sbt 2.x.

We should better fail at caching than fail at building.

sbt-app/src/sbt-test/dependency-management/test-artifact/disabled
adpi2, line 1: Let's see if I can fix this

server-test/src/test/scala/testpkg/BuildServerTest.scala
```diff
@@ -372,7 +372,7 @@ class BuildServerTest extends AbstractServerTest {
     assertMessage(s""""id":"$id"""", """"class":"main.Main"""")(duration = 30.seconds)
   }
 
-  test("buildTarget/run") {
+  ignore("buildTarget/run") {
```
adpi2: And those tests too.

sbt-app/src/sbt-test/actions/cross-advanced/test
eed3si9n: This changes the logLevel to debug from the shell.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: See the "beware the hermeticity breakage" section in https://eed3si9n.com/sudori-part4/.

> The `compile` task example reminds me of a potential pitfall with `Def.declareOutputDirectory` when used with `Def.cachedTask(...)`, which is that if we’re not careful, we could end up breaking the hermeticity. Specifically in this case, we could end up reusing an old, incorrect cache.
>
> This is because `VirtualFileRef` that represents the directory name will not contain the hash information of the content, which does not contain enough information to invalidate the downstream tasks. For example, if a cached task foo produces a directory and passes it to another cached task bar, passing `VirtualFileRef` alone will not be a sufficient cache.
>
> To workaround this, `Def.declareOutputDirectory` returns `VirtualFile` for the synthetic zip file, which does contain the hash, and should be able to invalidate the downstream.

server-test/src/server-test/buildserver/build.sbt
eed3si9n: I honestly don't see a huge value in allowing `target` to be changed at the subproject level. We should keep `target` as a backward compatible key, but generally sbt 2.x generates outputs in one single directory also called `target` at the working directory.

util-cache/src/main/scala/sbt/util/ActionCache.scala
eed3si9n: I think I'm including only files for manifest, so it's probably not necessary to include directories.

eed3si9n: Addressed in d6a65d71e07cfa5269cbd35cbab5945c55c85c04

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Addressed in b7e037adbbf3a0e33a97384849fc6803e17f0118

main/src/main/scala/sbt/Defaults.scala
adpi2: Thanks, I do understand now but I need to think more about it.

server-test/src/server-test/buildserver/build.sbt
adpi2: > I honestly don't see a huge value in allowing target to be changed at the subproject level.

Sure. But it's also about the `classDirectory`, and all the `package` files. It's tempting to move them out of the output directory, which would fail the action cache. This test is the only one that checks that the cached tasks don't fail if its output files are outside the cache directory.

I am not too much concerned about new builds. People will quickly understand they are not supposed to move output files outside the output directory. But I am more concerned about all the existing builds. We should be cautious not to break things that were permitted before and testing them.

server-test/src/test/scala/testpkg/BuildServerTest.scala
adpi2: Fixed by https://github.com/sbt/sbt/pull/7621/commits/038b96c3e4f416d5846b90a427ce38cde735d7e3

server-test/src/server-test/buildserver/build.sbt
adpi2: Resolved by https://github.com/sbt/sbt/pull/7621/commits/ba5ac774dfa8f63a73ec58ece58ab177a6baea45

sbt-app/src/sbt-test/dependency-management/test-artifact/disabled
adpi2: Fixed by https://github.com/sbt/sbt/pull/7621/commits/038b96c3e4f416d5846b90a427ce38cde735d7e3

adpi2:
To reproduce the test failure:
```
scripted dependency-management/cache-local dependency-management/test-artifact
```

adpi2 (approved):
I think this is in a mergeable state

PR #7623, opened 2024-08-13
[2.x] Check the digest during sync

eed3si9n:
This is an offshoot PR from https://github.com/sbt/sbt/pull/7621 that can go in independently.

**Problem**
Currently `syncBlobs` delete the existing files in the out directory when remote cache kicks in.

**Solution**
1. This refactors `Digest(...)` and adds support for `Digest.apply(Path)` and `Digest.sameDigest(...)`
2. This uses the `sameDigest` to compare the digest and replace the existing out files only when it needs to


adpi2 (approved):
LGTM

PR #7470, opened 2024-01-02
Include `managedSources` in `packageSrc`

Friendseeker:
Fixes #2205, https://github.com/sbt/zinc/issues/140

### Validating the PR

Before the fix
```
❯ jar tf /Users/jiahuitan/.ivy2/local/org.scala-sbt/compiler-interface/1.9.0-SNAPSHOT/srcs/compiler-interface-sources.jar
META-INF/MANIFEST.MF
xsbti/
xsbti/api/
xsbti/compile/
xsbti/compile/analysis/
xsbti/AnalysisCallback.java
xsbti/AnalysisCallback2.java
xsbti/ArtifactInfo.java
xsbti/BasicVirtualFileRef.java
xsbti/CompileCancelled.java
xsbti/CompileFailed.java
xsbti/FileConverter.java
xsbti/InteractiveConsoleFactory.java
xsbti/InteractiveConsoleInterface.java
xsbti/InteractiveConsoleResponse.java
xsbti/InteractiveConsoleResult.java
xsbti/PathBasedFile.java
xsbti/Reporter.java
xsbti/UseScope.java
xsbti/VirtualDirectory.java
xsbti/VirtualFile.java
xsbti/VirtualFileRef.java
xsbti/VirtualFileWrite.java
xsbti/api/AbstractLazy.java
xsbti/api/Lazy.java
xsbti/api/Modifiers.java
xsbti/api/SafeLazy.java
xsbti/compile/APIChange.java
xsbti/compile/AnalysisContents.java
xsbti/compile/AnalysisStore.java
xsbti/compile/AuxiliaryClassFiles.java
xsbti/compile/CachedCompiler.java
xsbti/compile/CachedCompilerProvider.java
xsbti/compile/Changes.java
xsbti/compile/ClassFileManager.java
xsbti/compile/CompileAnalysis.java
xsbti/compile/CompileOrder.java
xsbti/compile/CompileProgress.java
xsbti/compile/CompilerBridgeProvider.java
xsbti/compile/CompilerInterface2.java
xsbti/compile/ConsoleInterface1.java
xsbti/compile/DefaultExternalHooks.java
xsbti/compile/DefinesClass.java
xsbti/compile/DependencyChanges.java
xsbti/compile/ExternalHooks.java
xsbti/compile/GlobalsCache.java
xsbti/compile/IncToolOptionsUtil.java
xsbti/compile/IncrementalCompiler.java
xsbti/compile/InitialChanges.java
xsbti/compile/InvalidationProfiler.java
[...]

PR #7630, opened 2024-08-26
Include ALL sources, and resources in source jar

jroper:
This follows on from #7470, to include all sources, not just managed and unmanaged, in the source jar, along with all resources (previously only unmanaged resources were included).

This means that if, for whatever crazy reason, someone does modify the `sources` task to include additional sources or filter out sources, rather than using the managed or unmanaged source mechanisms, their changes will still be reflected in the source jar. It ensures consistency with both `compile` and `doc`, which both also depend on `sources` rather than manually gathering `unmanagedSources` and `managedSources` themselves.


adpi2 (approved):
LGTM

eed3si9n (approved):
Thanks!

PR #7636, opened 2024-09-02
Scala 2.12.20 (was .19)

SethTisue:
I'll un-draft after 2.12.20 has been announced

CI will fail for now because some dependencies haven't published yet (kind-projector, scalameta)


SethTisue:
it seems from https://github.com/scalameta/scalameta/issues/3950 that the oldest semanticdb version available for 2.12.20 is 4.9.3. wasn't sure whether to go with that or upgrade all the way to 4.9.9; let's try 4.9.9 and see how CI feels about it

UPDATE: yay, CI likes it

eed3si9n (approved):
Thanks!

Issue #6292, opened 2021-02-01
remote caching: run test incrementally


gontard:
As explained in #6286, i am integrating the new remote caching feature.

I would like to also take advantage of the remote caching to run test incrementally as explained in [this blog post](https://eed3si9n.com/cached-compilation-for-sbt):
> This idea could expand out to things like test results. sbt currently stores the list of succeeded tests and their timestamps. Using that testQuick run incremental tests. If we used the content hash, or something that does not depend on the timestamp, we might be able to run only the tests that's been affected by the last CI run.

If i understand correctly since #6026 the remote caching relies on content hashing. Is there something else missing?

Also in source code, there is already a commented out block to do that:
https://github.com/sbt/sbt/blob/0a6e5c311696ae2ae39e3da2ac80f9206a64179d/main/src/main/scala/sbt/internal/RemoteCache.scala#L411-L416

eed3si9n:
Fixed in sbt 2.x - https://github.com/sbt/sbt/pull/7644

PR #7644, opened 2024-09-07
[2.x] Remote test caching

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/6292

See https://eed3si9n.com/sudori-part5/ for details.

## Problem
Current implementation of testQuick depends on the concept of timestamp, which probably won't work well with the new consistent analysis store or the idea of remote caching.

## Solution
This first makes the incrementality hermetic (no longer depends on the timestamp). Instead, this calculates the aggregated SHA-256 of the class files involved in the test. Next it uses remote cache to track the success results of the test suites, which makes `testQuick` cached.


main/src/main/scala/sbt/internal/TestStatusReporter.scala
eed3si9n, line 1: This is mostly a file move from "testing" subproject to "main" since it needs to access more wider things.

main/src/main/scala/sbt/internal/IncrementalTest.scala
```diff
@@ -0,0 +1,119 @@
...
+      case (Nil, _)   => Seq((s: String) => !matches(excludeFilters, s))
+      case _ =>
+        includeFilters.map(f => (s: String) => (f.accept(s) && !matches(excludeFilters, s)))
+end IncrementalTest
+
+/**
+ * ClassStamper provides `transitiveStamp` method to calculate a unique
+ * fingerprint, which will be used for runtime invalidation.
+ */
+class ClassStamper(
```
eed3si9n, line 145: This is sort of a refactored version of sbt 1.x's `Defaults.testQuickFilter`.

main/src/main/scala/sbt/internal/IncrementalTest.scala
```diff
@@ -0,0 +1,218 @@
...
+    val cp = (Keys.test / fullClasspath).value
+    val testNames = Keys.definedTests.value.map(_.name).toVector.distinct
+    val converter = fileConverter.value
+    val sv = Keys.scalaVersion.value
+    val inputs = (Keys.compile / Keys.compileInputs).value
+    // by default this captures JVM version
+    val extraInc = Keys.extraIncOptions.value
+    // throw in any information useful for runtime invalidation
+    val salt = s"""$sv
+${converter.toVirtualFile(inputs.options.classesDirectory)}
```
adpi2: `sv` and `classesDirectory` should already be reflected in `cp` right? So I think we can remove them from here.

main/src/main/scala/sbt/internal/IncrementalTest.scala
```diff
@@ -0,0 +1,218 @@
...
+
+  // cache the test digests against the fullClasspath.
+  def definedTestDigestTask: Initialize[Task[Map[String, Digest]]] = Def.cachedTask {
+    val cp = (Keys.test / fullClasspath).value
+    val testNames = Keys.definedTests.value.map(_.name).toVector.distinct
+    val converter = fileConverter.value
+    val sv = Keys.scalaVersion.value
+    val inputs = (Keys.compile / Keys.compileInputs).value
+    // by default this captures JVM version
+    val extraInc = Keys.extraIncOptions.value
```
adpi2: We capture the JVM version, should we capture the OS too? I think yes, because as soon as there is a file or path in the test, it becomes OS dependent.

main/src/main/scala/sbt/internal/IncrementalTest.scala
```diff
@@ -0,0 +1,218 @@
...
+    // throw in any information useful for runtime invalidation
+    val salt = s"""$sv
+${converter.toVirtualFile(inputs.options.classesDirectory)}
+${extraInc.mkString(",")}
+"""
+    val extra = Vector(Digest.sha256Hash(salt.getBytes("UTF-8")))
+    val stamper = ClassStamper(cp, converter)
+    // TODO: Potentially do something about JUnit 5 and others which might not use class name
```
adpi2, line 54: Should we add an issue about this

util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -69,38 +64,82 @@ object ActionCache:
...
+      case Right(_) => true
+      case Left(_)  => false
+    }
+
+  inline private[sbt] def findActionResult[I: HashWriter, O](
+      key: I,
+      codeContentHash: Digest,
+      extraHash: Digest,
+      config: BuildWideCacheConfiguration,
+  )(f: Either[Throwable, ActionResult] => O) =
```
adpi2: Why passing a lambda here? It seems the caller just need to match on the `Either` of the action result.

```suggestion
  inline private[sbt] def findActionResult[I: HashWriter, O](
      key: I,
      codeContentHash: Digest,
      extraHash: Digest,
      config: BuildWideCacheConfiguration,
  ): Either[Throwable, ActionResult] =
```

main/src/main/scala/sbt/internal/IncrementalTest.scala
```diff
@@ -0,0 +1,218 @@
...
+                .map(stampVf)
+          val classDigests = relations.productClassName
+            .reverse(className)
+            .flatMap: prodClassName =>
+              relations
+                .definesClass(prodClassName)
+                .flatMap: sourceFile =>
+                  relations
+                    .products(sourceFile)
+                    .map(stampVf)
```
adpi2, line 198: So we use the hash of the source file to stamp the class? This means we will have slightly more invalidation than needed. Don't we have the hash of the class file anywhere?

adpi2 (commented):
Using the `RemoteCache` to implement `testQuick` is a brilliant idea. I left some comments but overall the code looks pretty good.

main/src/main/scala/sbt/internal/IncrementalTest.scala
eed3si9n: `classesDirectory` is used here sort of as a proxy to subproject name, Scala version, and platform (JVM vs JS vs Native). I'm not sure if `cp` covers them all.

main/src/main/scala/sbt/internal/IncrementalTest.scala
eed3si9n: While that's technically true, it's often annoying when on Bazel the CI test results doesn't get reused on macOS laptops. Maybe there should be a configurable setting per subproject for this purpose.

util-cache/src/main/scala/sbt/util/ActionCache.scala
eed3si9n: Yea, you're right. For some reason, I thought letting it accept a function would make the other functions make nicer to write, but I guess it won't make much difference.

main/src/main/scala/sbt/internal/IncrementalTest.scala
eed3si9n: Zinc might know, but I don't think we're exposing it? See the commented out code below.

main/src/main/scala/sbt/internal/IncrementalTest.scala
eed3si9n: Actually I was confusing myself with dynamic test examples. The test suite itself even on JUnit 5 I think would be a traditional Java class. I added an issue to verify for nested class, but this is nothing new https://github.com/sbt/sbt/issues/7650

main/src/main/scala/sbt/internal/IncrementalTest.scala
adpi2: If we change any of these, the content of the project's jar in the classpath will be different, and so its hash, and the Scala library as well.

main/src/main/scala/sbt/internal/IncrementalTest.scala
adpi2: Okay

adpi2:
Another thing that crossed my mind is that tests can depend on external files,for instance the `pos` and `neg` files in the compiler repos. It would be good to add some sort of `extraTestDigest: Task[Digest]` to allow the user to customize it, accordingly.

This can also be used to resolve https://github.com/sbt/sbt/pull/7644#discussion_r1752281969

util-cache/src/main/scala/sbt/util/ActionCache.scala
eed3si9n: Addressed in 291778086d6a921b099c337d07bffc69201bd655

main/src/main/scala/sbt/internal/IncrementalTest.scala
eed3si9n: Addressed in e9c04108bd7766d030f19fd2ab4784608f0adafa

eed3si9n:
@adpi2 

>  It would be good to add some sort of `extraTestDigest: Task[Digest]` to allow the user to customize it, accordingly.

Added `extraTestDigests` key in e9c04108bd7766d030f19fd2ab4784608f0adafa

adpi2 (approved):
LGTM

Issue #7641, opened 2024-09-06
[2.x] In-source projectMatrix


eed3si9n:
https://github.com/sbt/sbt-projectmatrix

PR #7647, opened 2024-09-10
[2.x] Optimize evaluation of `.all` on tasks and settings

adpi2:
In anticipation of #7641, I am trying to optimize the loading time of big project matrices, such as https://github.com/softwaremill/sttp, which contains 218 projects and more than 200,000 keys.

## Before

The CPU profiling shows that the evaluation of `.all` is far more expensive than it should be. This is because we call `ScopeFilter.apply` on each and every scope.
taskKeyAll-before

## After

The optimization consists of grouping all scopes by project, config and task to speed up the filtering.

taskKeyAll-after

## Backporting to sbt 1.x?

I've done the profiling on sbt 1.x but the changes are not strictly binary compatible. Probably it is possible to backport the optimization in a binary compatible way but the code is going to be much more ugly.

### DIFF
--- a/main/src/main/scala/sbt/ScopeFilter.scala
+++ b/main/src/main/scala/sbt/ScopeFilter.scala
@@ -9,29 +9,54 @@
 package sbt
 
 import sbt.internal.{ Load, LoadedBuildUnit }
-import sbt.internal.util.{ AttributeKey, Dag, Types }
+import sbt.internal.util.{ AttributeKey, Dag }
 import sbt.librarymanagement.{ ConfigRef, Configuration }
-import Types.const
+import sbt.internal.util.Types.const
 import Def.Initialize
 import sbt.Project.inScope
 import java.net.URI
 
+sealed abstract class ScopeFilter { self =>
+
+  /** Implements this filter. */
+  private[ScopeFilter] def apply(data: ScopeFilter.Data): Set[Scope]
+
+  /** Constructs a filter that selects values that match this filter but not `other`. */
+  def --(other: ScopeFilter): ScopeFilter = this && -other
+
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/ScopeFilter.scala
```diff
@@ -9,29 +9,54 @@
...
-import Types.const
+import sbt.internal.util.Types.const
 import Def.Initialize
 import sbt.Project.inScope
 import java.net.URI
 
+sealed abstract class ScopeFilter { self =>
+
+  /** Implements this filter. */
+  private[ScopeFilter] def apply(data: ScopeFilter.Data): Set[Scope]
```
adpi2, line 22: The old signature was: `def apply(data: ScopeFilter.Data): Scope => Boolean`. With the new signature we don't need to iterate on all scopes one by one. We can instead select the matching scopes directly from the `Map[Reference, Map[Config, Map[AtrributeKey, Set[Scope]]]]`

eed3si9n (approved):
Thanks!!

mzuehlke:
Really appreciated @adpi2 . Do you have a measurement how much this improved the loading time ? 
The saving by the sbt-git patch are for some repos are already quite substantially.

adpi2:
@mzuehlke In [softwaremill/sttp](https://github.com/softwaremill/sttp) it reduces the CPU time by 20%. On my laptop, it loads in 10.7 s instead of 13.8 s. This is a rough approximation though, I have not done a proper benchmark.

In total, with the sbt-git patch it's about 2x faster. 10.7 s instead of 21.6 s. Also there is parallelization involved, so depending on your computer, in can be better or not.

PR #7652, opened 2024-09-10
update ErrorHandling.scala. avoid reference deprecated `ThreadDeath`

xuwei-k:
https://bugs.openjdk.org/browse/JDK-8289610

deprecated for removal since JDK 20


eed3si9n (approved):
Thanks!

Issue #7656, opened 2024-09-11
[2.x] `ActionCache` is broken on Windows


adpi2:
Last time I tried sbt 2.x on Windows I noticed that the `ActionCache` was broken because of symbolic links.

eed3si9n:
I asked ChatGPT about this.

- On Windows, it's possible to make symbolic links for files, however you need to run the application with Administrator privileges.
- On Windows 10 and newer versions, it's possible to create symbolic links without requiring Administrator privileges, if you enable Developer Mode.
- Junctions are a type of hard link for directories and can be created by non-admin users without Developer Mode. They behave like symlinks but only work for directories. Creation is rather involved - https://github.com/Atry/scala-junction.
- Hard links can be used for files and do not require admin privileges. Hard links must be on the same volume. A hard link essentially mirrors the original file, and changes to one reflect on the other, so we probably do not want to use this.
- Copying. While not ideal, another simple workaround is to just copy the file or directory in cases where symlinks or hard links aren't practical. This doesn’t offer the same functionality as a link, but it can be a stopgap solution for non-admin users.

/cc @BillyAutrey

BillyAutrey:
I can see situations where users might not have administrator privileges, due to corporate policies on their laptops. In order, that means:
1. I would say the administrator mode requirement means this is not a great solution for us.
2. Requiring developer mode could create some user confusion/supportability issues. Worth investigating the difficulty here.
3. Junctions are interesting, but I don't know that we want to use that library directly. Worth investigating.
4. Being on the same volume reduces our supportability, in case someone wants to use a different volume as their disk cache.
5. This is our default, worst case scenario.

I will say that I've talked to others, and every path to try and support symlinks in Windows (for things like Bazel) ends in pain.

Perhaps a good solution here is to identify, from options 2 and 3, which is the easiest to use. And then allow users to opt out by copying. Or, alternatively, fall back to copying ourselves.

eed3si9n:
I don't think junction would work for us since mapping directories won't give us the flexibility of naming the blob with hash in CAS, and symlinking it with useful names like `*.jar`.

So what I suggest we should do is:

- Start with creating symbolic link wrapped in try-catch, assuming that the user either has an Administrator privileges or has enabled Developer Mode - <https://learn.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development>.
- If we hit `catch` clause, we can:
[...]

PR #7668, opened 2024-09-15
[2.x] fix: fallback to file copy

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7656

## Problem
Disk cache currently uses symbolic links, which won't work on Windows without the Administrator privileges or Developer Mode.

## Solution
This falls back to using file copy.

### DIFF
--- a/util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
+++ b/util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
@@ -2,7 +2,8 @@ package sbt.util
 
 import java.io.RandomAccessFile
 import java.nio.ByteBuffer
-import java.nio.file.{ Files, Path, Paths }
+import java.nio.file.{ Files, FileSystemException, Path, Paths, StandardCopyOption }
+import java.util.concurrent.atomic.AtomicBoolean
 import sjsonnew.*
 import sjsonnew.support.scalajson.unsafe.{ CompactPrinter, Converter, Parser }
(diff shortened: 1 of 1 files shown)


util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
```diff
@@ -263,24 +266,43 @@ class DiskActionCacheStore(base: Path) extends AbstractActionCacheStore:
...
+      )
+    // See https://github.com/sbt/sbt/issues/7656
+    // On Windows, the program has be running under the Administrator privileges or the
+    // user enable Developer Mode on Windows 10+ to create symbolic links.
+    def writeFileAndNotify(outPath: Path): Path =
       Files.createDirectories(outPath.getParent())
       val result = Retry:
         if Files.exists(outPath) then IO.delete(outPath.toFile())
-        Files.createSymbolicLink(outPath, casFile)
+        if symlinkSupported.get() then
```
BillyAutrey, line 283: We had briefly discussed updating this to prevent repeat attempts to create links. Is that still the intent here?

eed3si9n: Yea good catch. I forgot to update the atomic boolean.

eed3si9n: Addressed in aa0ea6ee9f26abc0c2467945051a14d10f44abe7

BillyAutrey (approved):

Issue #7675, opened 2024-09-15
[2.x] testQuick doesn't invalidate on resource changes


eed3si9n:
## steps

1. testQuick
2. change resource
3. testQuick

## problem

This doesn't rerun testQuick

## expectation

Invalidate testQuick

PR #7676, opened 2024-09-15
[2.x] fix: Invalidate testQuick on resource file changes

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7675

## Problem
testQuick currently does not invalidate on resource file changes.

## Solution
This includes resource digests into the input.


adpi2 (commented):
How can it be that resources are not reflected in the classpath hash?

BillyAutrey (approved):

eed3si9n:
@adpi2 
> How can it be that resources are not reflected in the classpath hash?

`testQuick` invalidates by forming sub-JAR graph, and resource relationship is not captured via Zinc Analysis.

adpi2 (approved):

PR #7677, opened 2024-09-15
[2.x] refactor: Delete Ivy-related test

eed3si9n:


adpi2 (commented):
But it seems this test was testing the classloader cache. Wouldn't it better to rewrite it with Coursier?

Issue #7678, opened 2024-09-15
[2.x] testQuick doesn't invalidate on testOptions change


eed3si9n:
## steps

1. testQuick
2. change testOptions
3. testQuick

## problem

This doesn't rerun testQuick

## expectation

Invalidate testQuick

PR #7680, opened 2024-09-16
[2.x] fix: Invalidate testQuick on test argument changes

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7678

## Problem
testQuick currently does not invalidate on argument changes.

## Solution
This includes test argument digests.


main-actions/src/main/scala/sbt/Tests.scala
```diff
@@ -68,26 +69,58 @@ object Tests {
...
+  final case class Setup(setup: ClassLoader => Unit, codeDigest: Digest) extends TestOption
 
   /**
    * Defines a TestOption that will evaluate `setup` before any tests execute.
    * Setup is not currently performed for forked tests.
    */
-  def Setup(setup: () => Unit) = new Setup(_ => setup())
+  inline def Setup(inline setup: () => Unit): Setup = ${ unitSetupMacro('setup) }
+
+  def unitSetupMacro(fn: Expr[() => Unit])(using Quotes): Expr[Setup] =
```
adpi2:
```suggestion
  private def unitSetupMacro(fn: Expr[() => Unit])(using Quotes): Expr[Setup] =
```

main-actions/src/main/scala/sbt/Tests.scala
```diff
@@ -68,26 +69,58 @@ object Tests {
...
+    val codeDigest = Digest.sha256Hash(fn.show.getBytes("UTF-8"))
+    val codeDigestStr = Expr(codeDigest.toString())
+    '{
+      new Setup(_ => $fn(), Digest($codeDigestStr))
+    }
+
+  inline def Setup(inline setup: ClassLoader => Unit): Setup = ${ clSetupMacro('setup) }
```
adpi2:
```suggestion
  private inline def Setup(inline setup: ClassLoader => Unit): Setup = ${ clSetupMacro('setup) }
```

main-actions/src/main/scala/sbt/Tests.scala
```diff
@@ -68,26 +69,58 @@ object Tests {
...
+  final case class Cleanup(cleanup: ClassLoader => Unit, codeDigest: Digest) extends TestOption
 
   /**
    * Defines a TestOption that will evaluate `cleanup` after all tests execute.
    * Cleanup is not currently performed for forked tests.
    */
-  def Cleanup(cleanup: () => Unit) = new Cleanup(_ => cleanup())
+  inline def Cleanup(inline cleanup: () => Unit) = ${ unitCleanupMacro('cleanup) }
+
+  def unitCleanupMacro(fn: Expr[() => Unit])(using Quotes): Expr[Cleanup] =
```
adpi2:
```suggestion
  private def unitCleanupMacro(fn: Expr[() => Unit])(using Quotes): Expr[Cleanup] =
```

main-actions/src/main/scala/sbt/Tests.scala
```diff
@@ -68,26 +69,58 @@ object Tests {
...
+    val codeDigest = Digest.sha256Hash(fn.show.getBytes("UTF-8"))
+    val codeDigestStr = Expr(codeDigest.toString())
+    '{
+      new Cleanup(_ => $fn(), Digest($codeDigestStr))
+    }
+
+  inline def Cleanup(inline cleanup: ClassLoader => Unit): Cleanup = ${ clCleanupMacro('cleanup) }
+
+  def clCleanupMacro(fn: Expr[ClassLoader => Unit])(using Quotes): Expr[Cleanup] =
```
adpi2:
```suggestion
  private def clCleanupMacro(fn: Expr[ClassLoader => Unit])(using Quotes): Expr[Cleanup] =
```

adpi2 (approved):

main-actions/src/main/scala/sbt/Tests.scala
xuwei-k: Is there any reason `private`? 🤔 

sbt 1.x support

```scala
Test / testOptions += Tests.Setup { (classLoader: ClassLoader) =>
  
}
```

but can't call `Setup` with `ClassLoader` param in sbt 2.x

https://github.com/sbt/sbt/blob/v1.11.5/main-actions/src/main/scala/sbt/Tests.scala#L69

eed3si9n: We wanted to eventually kick out test execution to a persistent worker, and a potentially remote box, like forked test. In that world, sbt server won't have access to the classloader. Maybe in the future, we will have some technology to ship lambda expression across the machine (like Spores), but for now it would be safe to reduce the API, I think.

PR #7681, opened 2024-09-17
[2.x] Merge sbt-projectmatrix

### LINKED ISSUE #7641: [2.x] In-source projectMatrix

adpi2:
Fixes #7641


main/src/main/scala/sbt/ProjectMatrix.scala
```diff
@@ -0,0 +1,818 @@
...
+                  List(scalaVersion := sv)
+                case _ =>
+                  List(autoScalaLibrary := false, crossPaths := false)
+              }
+            )
+            .settings(
+              outputPath := {
+                val o = outputPath.value
+                if nonScalaNorPlatformDirSuffix.nonEmpty then s"$o/$nonScalaNorPlatformDirSuffix"
+                else o
```
adpi2, line 320: No need a custom path for platform anymore, it's in by default.

adpi2, line 320: No need a custom path for platform anymore, it's already in the path by default.

sbt-app/src/sbt-test/project-matrix/js/pending
adpi2, line 1: Js and Native tests are pending

main/src/main/scala/sbt/ProjectMatrix.scala
```diff
@@ -0,0 +1,818 @@
...
+            }
+            .headOption
+            .getOrElse(sys.error(s"platform axis is missing in $axes"))
+          val childId = projectIds(r)
+          val deps = dependencies.map { resolveMatrixDependency(_, r) } ++ nonMatrixDependencies
+          val aggs = aggregate.map { case ref: LocalProjectMatrix =>
+            val other = lookupMatrix(ref)
+            resolveMatrixAggregate(other, r)
+          } ++ nonMatrixAggregate
+          val dotSbtMatrix = new java.io.File(".sbt") / "matrix"
```
adpi2, line 298: I wonder if we should change that path to `target/matrix`.

eed3si9n:
Could we keep the git history from the original plz? It's only around 100 commits.
https://eed3si9n.com/splitting-git-repo/ has some related git commands.

adpi2:
> Could we keep the git history from the original plz? It's only around 100 commits.

Here we are

main/src/main/scala/sbt/ProjectMatrix.scala
eed3si9n: Yea, something we can explore post merge I guess. Not sure how IDEs would behave if the base directories get wiped up.

eed3si9n (approved):
Thanks!

PR #7682, opened 2024-09-17
[2.x] Fix scripted cache tests

adpi2:
For some reason, we forgot to run `scripted cache/*` on the CI. This PR changes that, and makes it pass.


eed3si9n (approved):
Thanks!

Issue #7685, opened 2024-09-18
[2.x] Change test to be testQuick


eed3si9n:
1. Change `test` to behave as `testQuick`
2. Add `testFull` for running tests without the test result caching

PR #7686, opened 2024-09-19
[2.x] feat: Rename testQuick to test

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7685

## Problem
We have (cached) incremental testing, but it's not being used by default.

## Solution
This renames testQuick to test, and test to testFull.


sbt-app/src/sbt-test/tests/setup-cleanup/test
```diff
@@ -4,7 +4,7 @@ $ absent tested
 $ absent cleanup
 
 # without Setup configured, the setup file won't exist and the test will fail
--> testQuick
+-> testFull
```
BillyAutrey, line 7: Just making sure, did you mean to convert this from a cached test to a full test?

sbt-app/src/sbt-test/tests/setup-cleanup/test
```diff
@@ -39,7 +39,7 @@ $ absent tested
...
```
BillyAutrey, line 42: Same here - this changes the test type. Looking at the rest of the file it seems like it's fine, the whole thing looks like it's doing full tests with this change. But wanted to make sure this was your intent.

BillyAutrey (approved):

sbt-app/src/sbt-test/tests/setup-cleanup/test
eed3si9n: Yea, I made the change in #7680 recently to test `testQuick` but I think `testFull` would test the actual setup capability more correctly.

Issue #7687, opened 2024-09-19
[2.x] Headless mode


eed3si9n:
As an intermediary step towards https://github.com/sbt/sbt/issues/6178, I wonder if it would be useful to implement headless mode. We first define a `Global` setting called `displayShell`.

Next, in build.sbt:

```scala
Global / displayShell := false
```

When you start sbt, the session will start in headless mode where the `shell` command starts, but without prompting the user. From another terminal, you can still interact with the session via `sbt --client`.

So from another terminal we would be able to start and stop the shell UI as:

```scala
> set Global / displayShell := true
```

adpi2:
I don't understand the goal here.

If `Global / displayShell := false` and I do `sbt --client` does it start a client shell?

eed3si9n:
> I don't understand the goal here.

The idea basically is to think about the suppression of the console prompt, rather than the non-blocking aspect of the run, partly to explore how it can be implemented. The exact settings etc can be scaffolded.
https://github.com/sbt/sbt/pull/7688 basically demonstrates this idea, esp around

```scala
private[sbt] def pauseChannelDuringJob(state: State, handle: JobHandle): Unit = ...
```

and concrete implementation of pause, which can be a stateful flag:

```scala
  private[sbt] var _active: Boolean = true
  private[sbt] def pause(): Unit = _active = false
```

Issue #6178, opened 2020-11-30
Non-blocking run


eed3si9n:
This is reverse reported from https://github.com/sbt/sbt/pull/6074
Ref https://github.com/sbt/sbt/issues/2799 / https://github.com/sbt/sbt/issues/2804 / https://github.com/sbt/sbt/pull/2936

## steps

1. start sbt session in a terminal
2. connect from another terminal using sbtn or BSP
3. execute `run` from either of the terminal

## problem

Command queue is blocked until `run` completes

## expectation

While is running, sbt server can process other tasks, including more `run`

SethTisue:
This is especially desirable when sbt is serving Metals via BSP.

adpi2:
It seems that `bgRun` does that already. Do we want the `bgRun` behavior to be the default behavior of `run`?

eed3si9n:
`bgRun` doesn't have the terminal attached (thus, background), which can't really replace `run`. I don't remember exactly how we implemented `bgRun`, but it does have full information on the necessary files it needs to run user's app. I think the next step is maybe specify/document the full information, and give it to sbtn so it can run it. That would be step 1.

Step 2 (aka "idea 5: BSP support + persistent workers" in https://eed3si9n.com/sbt-2.0-ideas) is to keep a mini worker program running, which is capable of executing `run`, and pipe console (stdin, stdout, stderr) to it. This would retain sbt's current benefit of having JIT'ed code with faster `run`-spinup.

eed3si9n:
`run` already calls `bgRun`, but it blocks until the jobs completion. I have a PR that no longer blocks on the completion, and turns off the sbt shell prompt instead - https://github.com/sbt/sbt/pull/7688.

This sort of works for the scenario where `run` is issued from the traditional sbt session, but it doesn't fully work for sbtn because sbt server doesn't seem to forward stdout to sbtn? Overall I think it's a good proof-of-concept on the approach though.

eed3si9n:
Here's a PR that implements client-side run on sbt 2.x - https://github.com/sbt/sbt/pull/8060

PR #7688, opened 2024-09-20
[2.x] feat: Non-blocking run

eed3si9n:
Ref https://github.com/sbt/sbt/issues/6178
Ref https://github.com/sbt/sbt/issues/7687

## Problem
`run` currently blocks all other commands, such as BSP commands.

## Solution
`run` no longer blocks the command execution loop. Instead it pauses the prompt on the current command channel.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2152,27 +2152,17 @@ object Defaults extends BuildCommon {
     }
 
   // `runMain` calls bgRunMain in the background and pauses the current channel
-  def foregroundRunMainTask: Initialize[InputTask[RunInfo]] =
+  def foregroundRunMainTask: Initialize[InputTask[EmulateForeground]] =
     Def.inputTask {
       val handle = bgRunMain.evaluated
-      val service = bgJobService.value
-      val st = state.value
-      st.remainingCommands match
-        case Nil => service.waitForTry(handle).get
-        case _   => service.pauseChannelDuringJob(st, handle)
-      RunInfo(handle)
+      EmulateForeground(handle)
     }
 
   // `run` task calls bgRun in the background and pauses the current channel
-  def foregroundRunTask: Initialize[InputTask[RunInfo]] =
+  def foregroundRunTask: Initialize[InputTask[EmulateForeground]] =
     Def.inputTask {
       val handle = bgRun.evaluated
-      val service = bgJobService.value
-      val st = state.value
-      st.remainingCommands match
-        case Nil => service.waitForTry(handle).get
-        case _   => service.pauseChannelDuringJob(st, handle)
-      RunInfo(handle)
+      EmulateForeground(handle)
     }
 
   def runMainTask(
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2145,21 +2145,28 @@ object Defaults extends BuildCommon {
       }
     }
 
-  // runMain calls bgRunMain in the background and waits for the result.
-  def foregroundRunMainTask: Initialize[InputTask[Unit]] =
-    Def.inputTask[Unit] {
+  // `runMain` calls bgRunMain in the background and pauses the current channel
+  def foregroundRunMainTask: Initialize[InputTask[RunVoid]] =
+    Def.inputTask {
       val handle = bgRunMain.evaluated
       val service = bgJobService.value
-      service.waitForTry(handle).get
-      ()
+      val st = state.value
+      st.remainingCommands match
+        case Nil => service.waitForTry(handle).get
+        case _   => service.pauseChannelDuringJob(st, handle)
+      RunVoid
     }
 
-  // run calls bgRun in the background and waits for the result.
-  def foregroundRunTask: Initialize[InputTask[Unit]] =
+  // `run` task calls bgRun in the background and pauses the current channel
+  def foregroundRunTask: Initialize[InputTask[RunVoid]] =
     Def.inputTask {
       val handle = bgRun.evaluated
       val service = bgJobService.value
-      service.waitForTry(handle).get
+      val st = state.value
+      st.remainingCommands match
+        case Nil => service.waitForTry(handle).get
+        case _   => service.pauseChannelDuringJob(st, handle)
+      RunVoid
     }
 
   def runMainTask(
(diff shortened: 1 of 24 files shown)


main/src/main/scala/sbt/internal/Aggregation.scala
```diff
@@ -79,10 +79,14 @@ object Aggregation {
     val success = results match
       case Result.Value(_) => true
       case Result.Inc(_)   => false
+    // run task ends earlier than the program run
+    val isRunInfo = results match
+      case Result.Value(Seq(KeyValue(_, RunInfo(_)))) => true
+      case _                                          => false
```
adpi2: This seems a bit too custom and it's going to be tedious to replicate for other non-blocking tasks (test, console etc). Could we get the information that the command paused the channel from the state maybe?

In which case we don't need `RunInfo` anymore and we can return a bare `JobHandle`.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2145,21 +2151,28 @@ object Defaults extends BuildCommon {
...
+  def foregroundRunTask: Initialize[InputTask[RunInfo]] =
     Def.inputTask {
       val handle = bgRun.evaluated
       val service = bgJobService.value
-      service.waitForTry(handle).get
+      val st = state.value
+      st.remainingCommands match
+        case Nil => service.waitForTry(handle).get
+        case _   => service.pauseChannelDuringJob(st, handle)
+      RunInfo(handle)
```
adpi2: Maybe we can abstract over this: here we can return `ForegroundJob(handle)`, it is then matched by `EvaluateTask` which takes care of pausing the channel. This way we can easily define other non-blocking tasks.

main/src/main/scala/sbt/internal/CommandExchange.scala
```diff
@@ -376,13 +377,14 @@ private[sbt] final class CommandExchange {
...
 
-  def prompt(event: ConsolePromptEvent): Unit = {
+  def prompt(event: ConsolePromptEvent): Unit =
     currentExecRef.set(null)
     channels.foreach {
       case c if ContinuousCommands.isInWatch(lastState.get, c) =>
-      case c                                                   => c.prompt(event)
+      case c =>
+        if c.isPaused then ()
+        else c.prompt(event)
```
adpi2, line 386: Also if the channel is paused we should not show the `sbt server is running ...` message:

image

adpi2: Nor this one (probably harder to fix):

image

main/src/main/scala/sbt/internal/DefaultBackgroundJobService.scala
```diff
@@ -306,6 +305,28 @@ private[sbt] abstract class AbstractBackgroundJobService extends BackgroundJobSe
...
+        val level = channel.logLevel
+        channel.setLevel(Level.Error)
+        channel.pause()
+        handle match
+          case t: ThreadJobHandle =>
+            t.job.onStop: () =>
+              channel.setLevel(level)
+              channel.resume()
+              channel.prompt(ConsolePromptEvent(state))
+          case _ => ()
```
adpi2: What if the handle is not a `ThreadJobHandle`. If so the channel is never going to be resumed:

```suggestion
        handle match
          case t: ThreadJobHandle =>
            val level = channel.logLevel
            channel.setLevel(Level.Error)
            channel.pause()
            t.job.onStop: () =>
              channel.setLevel(level)
              channel.resume()
              channel.prompt(ConsolePromptEvent(state))
          case _ => ()
```

main/src/main/scala/sbt/internal/DefaultBackgroundJobService.scala
```diff
@@ -306,6 +305,28 @@ private[sbt] abstract class AbstractBackgroundJobService extends BackgroundJobSe
...
+      case _ => ()
+
+  private[sbt] def currentChannel(state: State): Option[CommandChannel] =
+    state.currentCommand match
+      case Some(e: Exec) if e.source.isDefined =>
+        val source = e.source.get
+        StandardMain.exchange.channelForName(source.channelName)
+      case _ => None
```
adpi2, line 329: NIT

```suggestion
    state.currentCommand.flatMap(_.source)
      .map(source => StandardMain.exchange.channelForName(source.channelName))
```

adpi2 (commented):
This is really encouraging and I think we can make it a bit more abstract and reusable.

I tried it and it worked okay but there are some glitches:
|               | sbt shell                 | sbt client                                     |
|---------------|---------------------------|------------------------------------------------|
| `fork := false` | All good                  | stdout not forwarded</br>`stdin.read` blocks forever |
| `fork := true`  | `stdin.read` blocks forever | `stdin.read` blocks forever                      |

Also it seems `Ctrl+C` is not catched by sbt and it always kills the shell.

main/src/main/scala/sbt/internal/DefaultBackgroundJobService.scala
eed3si9n: That creates a few lambdas whereas the pattern matching compiles into if statements.

main/src/main/scala/sbt/internal/DefaultBackgroundJobService.scala
BillyAutrey: I had this exact same concern. I think pausing in the match makes sense, and will prevent stuck terminals.

main/src/main/scala/sbt/internal/DefaultBackgroundJobService.scala
adpi2: How bad is that?

eed3si9n: It kind of depends on case-by-case, but on a hot path it does matter, so I tend to avoid `map(...).getOrElse(y)` style whenever I can just pattern match instead.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: I've addressed this in 8aa740574b91d670919df50663218688cc7eac27 by moving channel suspension into `EvaluateTask`.

main/src/main/scala/sbt/internal/Aggregation.scala
eed3si9n: Aggregation now checks if the originating channel is paused.

adpi2 (approved):
LGTM

BillyAutrey (approved):

PR #1878, opened 2015-02-25
Don't enable IvyPlugin on artificial root of a multi project build

eed3si9n:
Fixes #1869, Ref #1871

/review @jsuereth

PR #7691, opened 2024-09-22
[2.x] fix: skip publish on synthetic aggregate root

eed3si9n:
## Problem
There's been previous attempt like #1878 to skip publishing of the root, but it seems like the behavior has regressed at some point in time.

## Solution
This skips publishing on synthetic aggregate root project.


main/src/main/scala/sbt/internal/BuildDef.scala
```diff
@@ -55,12 +56,18 @@ private[sbt] object BuildDef:
...
       base: File,
       agg: Seq[ProjectRef]
   ): Project =
-    Project.mkGeneratedRoot(id, base, agg).settings(defaultProjectSettings)
+    Project
+      .mkGeneratedRoot(id, base, agg)
+      .settings(
+        defaultProjectSettings,
+        publish / skip := true,
+        publishLocal / skip := true,
```
eed3si9n, line 69: This is the main change.

eed3si9n:
Going to self-merge since this should be uncontroversial.

Issue #7690, opened 2024-09-22
[2.x] common settings do not work without root


eed3si9n:
## steps

```scala
organization := "com.example"

val x = projectMatrix(...)
```

here's another example:

```scala
organization := "com.example"
ivyPaths := IvyPaths((baseDirectory).value.toString, Some(((ThisBuild / baseDirectory).value / "ivy" / "cache").toString))

name := "root"

lazy val core = project
  .settings(
    name := "core",
  )
```

## problem

`organization` isn't injected to the projects.

## expectation

`organization` is injected to the all subprojects.

eed3si9n:
Here's a PR for this - https://github.com/sbt/sbt/pull/7693

PR #7693, opened 2024-09-22
[2.x] fix: Fix common settings loading when root is synthetic

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7690

## Problem
The new common settings feature doesn't work when the root isn't created by the user.

## Solution
This fixes common settings by calling `expandCommonSettingsPerBase(...)` on the synthetic root's base first.

## Note
See https://github.com/sbt/sbt/pull/7693/files?w=1 to ignore whitespace changes.


main/src/main/scala/sbt/internal/Load.scala
```diff
@@ -943,142 +943,160 @@ private[sbt] object Load {
...
+                  BuildDef.defaultProject(defaultID, buildBase)
+                else BuildDef.generatedRootSkipPublish(defaultID, buildBase, Nil)
+              }
+              val otherProjects =
+                loadTransitive1(
+                  newProjects = discovered,
+                  acc = acc,
+                  generated = Nil,
+                  commonSettings0 = commonSettings
+                    ++ expandCommonSettingsPerBase1(buildBase.getCanonicalFile()),
```
eed3si9n, line 1080: The key change is here. Given `hello/build.sbt` and `foo` subproject, it will now load common settings from `hello/build.sbt` as expected not just `hello/foo/build.sbt`.

PR #7694, opened 2024-09-22
[2.x] feat: plugin cross building as Scala cross building 

eed3si9n:
This is implemented on top of https://github.com/sbt/sbt/pull/7693

## Problem
Since Scala cross building works better than the plugin cross building `^^`, it was common for plugin authors to encode plugin cross building as Scala cross building given that we usually have zero or one sbt release in one Scala version.

## Solution
This brings in the setting into SbtPlugin so plugin authors can cross build using sbt 2.x.


main/src/main/scala/sbt/plugins/SbtPlugin.scala
```diff
@@ -10,12 +10,19 @@ package sbt
...
+object SbtPlugin extends AutoPlugin:
   override def requires = ScriptedPlugin
 
   override lazy val projectSettings: Seq[Setting[_]] = Seq(
-    sbtPlugin := true
+    sbtPlugin := true,
+    pluginCrossBuild / sbtVersion := {
+      scalaBinaryVersion.value match
+        case "3"    => sbtVersion.value
+        case "2.12" => "1.5.8"
```
adpi2, line 24: Why this version specifically?

sbt-app/src/sbt-test/project-matrix/plugin/build.sbt
```diff
@@ -0,0 +1,28 @@
+val scala3 = "3.3.3"
+val scala212 = "2.12.19"
+
+organization := "com.example"
+version := "0.1.0-SNAPSHOT"
+
+lazy val checkSbt2Plugin = taskKey[Unit]("")
+
+lazy val plugin = (projectMatrix in file("plugin"))
```
adpi2, line 9: It seems Scala 3 is going towards less [infix notation](https://docs.scala-lang.org/scala3/reference/changed-features/operators.html). Should we agree not to use it here anymore?

```suggestion
lazy val plugin = projectMatrix.in(file("plugin"))
```

adpi2 (approved):
LGTM

main/src/main/scala/sbt/plugins/SbtPlugin.scala
eed3si9n: So the specific version doesn't really matter here but picking some old version allows the build user to be on a range of sbt versions. For some reason it's now 1.5 maybe because 1.2 is too old.

sbt-app/src/sbt-test/project-matrix/plugin/build.sbt
eed3si9n: I'm generally +1 on most Scala 3 directions including limiting the use of infixes. However, `in` is clearly intended to act like an English DSL / pretend-keyword so we should actually keep it as such, perhaps by marking it infix.

PR #7697, opened 2024-09-25
[2.x] Improve backward source compat

adpi2:
Some changes I need to migrate [scala-js/scala-js](https://github.com/scala-js/scala-js) to sbt 2.x. The WIP migration of [scala-js/scala-js](https://github.com/scala-js/scala-js) is [here](https://github.com/scala-js/scala-js/compare/main...adpi2:scala-js:sbt-2.x).

The biggest change is in https://github.com/sbt/sbt/commit/625ac91afbed2f00514b44f20ef1bd1bb9551bb4. It solves a name conflict between `BuildSyntax.dependsOn` and `Scoped.dependsOn` that are both mixed in the sbt package object:

```
[error] -- [E008] Not Found Error: /home/piquerez/github/scala-js/scala-js/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala:497:25 
[error] 497 |      console := console.dependsOn(Def.task {
[error]     |                 ^^^^^^^^^^^^^^^^^
[error]     |value dependsOn is not a member of sbt.TaskKey[Unit].
[error]     |Extension methods were tried, but the search failed with:
[error]     |
[error]     |    Toplevel definition dependsOn is defined in
[error]     |      /home/piquerez/.sbt/boot/scala-3.3.3/org.scala-sbt/sbt/2.0.0-alpha12-SNAPSHOT/main_3.jar(sbt/BuildSyntax.class)
[error]     |    and also in
[error]     |      /home/piquerez/.sbt/boot/scala-3.3.3/org.scala-sbt/sbt/2.0.0-alpha12-SNAPSHOT/main-settings_3.jar(sbt/Scoped.class)
[error]     |    One of these files should be removed from the classpath.
```

To resolve it I moved `BuildSyntax` out of the sbt package object. Now, it is only available in `build.sbt` files from a default import.


main-settings/src/main/scala/sbt/Def.scala
```diff
@@ -29,8 +29,32 @@ import xsbti.{ HashedVirtualFileRef, VirtualFile, VirtualFileRef }
 import sjsonnew.JsonFormat
 import scala.reflect.ClassTag
 
+trait Def:
```
eed3si9n: I feel like this trait should be given a name that reflects what it is intended to do. BasicBuildSyntax?

adpi2: Maybe I should rename this one `BuildSyntax` and rename the other one `BareBuildSyntax`.

eed3si9n (approved):

Issue #7696, opened 2024-09-25
[2.x] Remove sbt 0.13 shell syntax


eed3si9n:

eed3si9n:
Here's a PR for this - https://github.com/sbt/sbt/pull/7700

PR #7700, opened 2024-09-28
[2.x] refactor: Remove old sbt 0.13 shell syntax

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7696

## Problem
sbt 1.x introduced the slash syntax midway, so it kept the old `a/test:console::scalacOptions` syntax, which hopefully aren't in any more. sbt 1.x has been displaying deprecation warning on this.

```scala
sbt:sbtRoot> mainProj/test:console::scalacOptions
[warn] sbt 0.13 shell syntax is deprecated; use slash syntax instead: mainProj / Test / console / scalacOption
```

## Solution

This removes the support from the shell parser.

### DIFF
--- a/main/src/main/scala/sbt/internal/Act.scala
+++ b/main/src/main/scala/sbt/internal/Act.scala
@@ -36,7 +36,6 @@ final class ParsedKey(val key: ScopedKey[_], val mask: ScopeMask, val separaters
 end ParsedKey
 
 object Act {
-  val ZeroString = "*"
   private[sbt] val GlobalIdent = "Global"
   private[sbt] val ZeroIdent = "Zero"
   private[sbt] val ThisBuildIdent = "ThisBuild"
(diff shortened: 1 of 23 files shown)


main/src/main/scala/sbt/internal/Act.scala
```diff
@@ -242,26 +239,15 @@ object Act {
...
 
-  def config(confs: Set[String]): Parser[ParsedAxis[String]] = {
-    val sep = ':' !!! "Expected ':' (if selecting a configuration)"
-    token(
-      (ZeroString ^^^ ParsedZero | value(examples(ID, confs, "configuration"))) <~ sep
-    ) ?? Omitted
-  }
-
   // New configuration parser that's able to parse configuration ident trailed by slash.
   private[sbt] def configIdent(
```
BillyAutrey, line 243: Nit: Since `config` is now gone, it's a good opportunity to name this back to `config` instead of using what seems like a differentiated method name.

BillyAutrey (approved):

main/src/main/scala/sbt/internal/Act.scala
eed3si9n: `config` has traditionally meant the lower case configuration name like `compile`, whereas `configIdent` was created to mean the Scala identity name, the upper case `Compile`. See also `zeroIdent`, `thisBuildIdent`, etc. It's a few more characters, but I feel more comfortable keeping this around since a lot of these are just passing string around and easy to get mixed up.

SethTisue:
amazing 👏

Issue #7702, opened 2024-09-28
[2.x] Remove sbt 0.13 build.sbt `in` methods


eed3si9n:
See https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#Migrating+to+slash+syntax

We should remove `scalacOptions in (Compile, console)` notation. This was deprecated in https://eed3si9n.com/sbt-1.5.0.

See also https://eed3si9n.com/syntactic-scalafix-rule-for-unified-slash-syntax/ for semi-automated migration using Scalafix.

adpi2:
@eed3si9n I think we can close this, since we merged https://github.com/sbt/sbt/pull/7703

PR #7703, opened 2024-09-28
[2.x] refactor: Remove sbt 0.13 `in` methods

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7702

## Problem
See https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#Migrating+to+slash+syntax

We should remove `scalacOptions in (Compile, console)` notation. This was deprecated in https://eed3si9n.com/sbt-1.5.0.

## Solution
This renames `in` to `rescope` for internal usage, and ports the usages to slash.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -186,7 +186,7 @@ You'd need alternative DSL import since you can't rely on sbt package object.
 
 ```scala
 // for slash syntax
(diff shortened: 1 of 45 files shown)


adpi2 (approved):
LGTM

SethTisue:
yay!!

PR #7704, opened 2024-09-30
[2.x] worker

eed3si9n:
This is a beginning of persistent worker idea, which is intended to offload tests.


worker/src/test/scala/sbt/internal/worker/WorkDispatchTest.scala
```diff
@@ -0,0 +1,81 @@
+package sbt.internal.worker
+
+import java.nio.file.{ Path, Paths }
+import java.net.URLClassLoader
+import scala.collection.mutable
+
+object WorkDispatchTest extends verify.BasicTestSuite:
```
eed3si9n, line 8: Tests.

PR #7712, opened 2024-10-02
[2.x] refactor: Remove useCoursier setting (drop Ivy)

eed3si9n:
## Problem
It's been a while since Coursier has been the default library management engine, and we don't need to support two.

## Solution
This removes `useCoursier` setting.


main/src/main/scala/sbt/Keys.scala
```diff
@@ -486,7 +486,6 @@ object Keys {
   @cacheLevel(include = Array.empty)
   val bspReporter = taskKey[BuildServerReporter]("").withRank(DTask)
 
-  val useCoursier = settingKey[Boolean]("Use Coursier for dependency resolution.").withRank(BSetting)
```
adpi2, line 489: Is there any other key that was Ivy specific?

adpi2 (approved):
:partying_face:

main/src/main/scala/sbt/internal/LibraryManagement.scala
```diff
@@ -332,7 +329,7 @@ private[sbt] object LibraryManagement {
         import Keys._
         val cacheDirectory = s.cacheDirectory
         val isRoot = er.contains(rs)
-        if csr then {
+        if true then {
```
BillyAutrey: Nit: I'm assuming you're doing this so that you can keep the (now dead) ivy code around. Doesn't this produce a compiler warning about an unreachable code path in the `else`?

I feel like version control protects us here, and we can just remove the dead code. Or if you want to feel safer, do what you've done in other places - comment out the `if`, and the whole `else` block.

BillyAutrey (approved):

main/src/main/scala/sbt/internal/LibraryManagement.scala
eed3si9n: No this was a mistake as I was changing code around to get it to compile.

eed3si9n: Addressed in 80fabeeb2c5780541f0900100aeef1e02dfa3b78

main/src/main/scala/sbt/Keys.scala
eed3si9n: It seems like there are some internal Ivy usage, but I could probably drop the external ivy.xml file support - 968bd9c7a2678aa6dcd16daa1fba3e9c2bc9f175

PR #7714, opened 2024-10-02
[2.x] Reduce heap size after load

### LINKED ISSUE #7711: [2.x] Loading settings is slow
Loading softwaremill/sttp is 2x slower with sbt 2.x than 1.10.x. Also it consumes a lot more heap memory and use too much GC. At first glance it seems the problem is `TupleMapExtension`, which creates too many short-lived arrays.

adpi2:
Fix #7711 

Loading settings uses about 1.5x more heap size in sbt 2.x, which makes GC much slower. The main reason was there was no specialized implementation of `mapN` in `Monad[Initialize[Task[?]]`, resulting in many `Action.Mapped((t3, Mapped((t2, Mapped(t1, x1 => x2 => x3 => f(x1, x2, x3))), ...)), ...)` instead of `Action.Mapped((t1, t2, t3), f)`.


adpi2:
Here is a comparison of loading sttp with sbt 1.x and 2.x on this PR:
|        | Settings | Used Heap size | Load time |
|--------|----------|-----------|-----------|
| 1.10.2 | 197031   | 675 MB    | 9846 ms   |
| 2.0.0-alpha12-SNAPSHOT    | 213166   | 532 MB    | 8504 ms   |
|        | 108 %    | 79 %      | 86 %      |

There are more settings in sbt 2.x but it is lighter and faster to load.

mzuehlke:
> Here is a comparison of loading sttp with sbt 1.x and 2.x:

These number are with this PR applied? Great reduction 🎉

adpi2:
> These number are with this PR applied? Great reduction 🎉

Yes it is. Without the PR, sbt 2.x is 2x slower than sbt 1.10.2.

util-collection/src/main/scala/sbt/internal/util/TupleMapExtension.scala
```diff
@@ -5,28 +5,76 @@
...
+            case 19 => t(0) :: t(1) :: t(2) :: t(3) :: t(4) :: t(5) :: t(6) :: t(7) :: t(8) :: t(9) :: t(10) :: t(11) :: t(12) :: t(13) :: t(14) :: t(15) :: t(16) :: t(17) :: t(18) :: nil
+            case 20 => t(0) :: t(1) :: t(2) :: t(3) :: t(4) :: t(5) :: t(6) :: t(7) :: t(8) :: t(9) :: t(10) :: t(11) :: t(12) :: t(13) :: t(14) :: t(15) :: t(16) :: t(17) :: t(18) :: t(19) :: nil
+            case 21 => t(0) :: t(1) :: t(2) :: t(3) :: t(4) :: t(5) :: t(6) :: t(7) :: t(8) :: t(9) :: t(10) :: t(11) :: t(12) :: t(13) :: t(14) :: t(15) :: t(16) :: t(17) :: t(18) :: t(19) :: t(20) :: nil
+            case 22 => t(0) :: t(1) :: t(2) :: t(3) :: t(4) :: t(5) :: t(6) :: t(7) :: t(8) :: t(9) :: t(10) :: t(11) :: t(12) :: t(13) :: t(14) :: t(15) :: t(16) :: t(17) :: t(18) :: t(19) :: t(20) :: t(21) :: nil
+            case _ => tuple.iterator.toList
+      // format: on
+      list.asInstanceOf[List[F1[Any]]]
 
+    def unmap(f: [a] => F1[a] => a): Tup = transform[[A] =>> A](f).asInstanceOf[Tup]
     def transform[F2[_]](f: [a] => F1[a] => F2[a]): Tuple.Map[Tup, F2] =
```
eed3si9n, line 19: Thanks @adpi2 !
For future context, could you maybe leave in the original "before" code in the comment with the explanation that we need to hand roll this due to performance reason please?

eed3si9n (approved):

util-collection/src/main/scala/sbt/internal/util/TupleMapExtension.scala
adpi2: Yes sure, and I should also port that back to the standard library.

Issue #7715, opened 2024-10-03
[2.x] "Overlapping output directories" using projectMarix


eed3si9n:
## steps

```scala
version := "0.1.1-SNAPSHOT"
organization := "com.eed3si9n"
description := "sbt plugin to vimquit"
licenses := Seq("MIT License" -> url("https://github.com/sbt/sbt-vimquit/blob/master/LICENSE"))
scalacOptions := Seq("-deprecation", "-unchecked")
publishMavenStyle := true

lazy val plugin = (projectMatrix in file("."))
  .enablePlugins(SbtPlugin)
  .settings(
    name := "sbt-vimquit",
  )
  .jvmPlatform(scalaVersions = Seq("3.3.3"))
```

## problem

```
[error] Overlapping output directories:/Users/xxx/sbt-vimquit/target/out/jvm/scala-3.3.3/sbt-vimquit:
[error] 	ProjectRef(file:/Users/xxx/sbt-vimquit/,sbt-vimquit)
[error] 	ProjectRef(file:/Users/xxx/sbt-vimquit/,plugin)
[error] Use 'last' for the full log.
[warn] Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)
```

## expectation

No errors.

## notes

This works fine if it's Scala 2.12:

```
sbt:sbt-vimquit> baseDirectory
[info] plugin2_12 / baseDirectory
[info] 	/Users/xxx/sbt-vimquit/.sbt/matrix/plugin2_12
[info] baseDirectory
[info] 	/Users/xxx/sbt-vimquit
```

PR #7716, opened 2024-10-03
[2.x] fix: Append -root to synthetic root

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7715

## Problem
I notice that the synthetic root project ends up conflicting with the projectMatrix on Scala 3, when the name of the matrix matches the directory name, which is fairly common.

Also when the matrix is located at the root, we still want the synthetic root for aggregation purpose, but we don't want the root to end up capturing the source files.

## Solution
1. Append `-root` to the root project.
2. Blank out the source files.

## Note

```scala
sbt:sbt-vimquit-root> projects
[info] In file:/xxx/sbt-vimquit/
[info] 	   plugin
[info] 	   plugin2_12
[info] 	 * sbt-vimquit-root
```

See also https://github.com/sbt/website/pull/1242

### DIFF
--- a/server-test/src/test/scala/testpkg/ServerCompletionsTest.scala
+++ b/server-test/src/test/scala/testpkg/ServerCompletionsTest.scala
@@ -33,6 +33,7 @@ class ServerCompletionsTest extends AbstractServerTest {
     })
   }
 
+  /*
   test("return completions for user classes") {
     val completionStr = """{ "query": "testOnly org." }"""
     svr.sendJsonRpc(
@@ -42,4 +43,5 @@ class ServerCompletionsTest extends AbstractServerTest {
       s contains """"result":{"items":["testOnly org.sbt.ExampleSpec"]"""
     })
   }
+   */
 }
--- a/server-test/src/test/scala/testpkg/ServerCompletionsTest.scala
+++ b/server-test/src/test/scala/testpkg/ServerCompletionsTest.scala
@@ -34,6 +34,8 @@ class ServerCompletionsTest extends AbstractServerTest {
   }
 
   /*
+  // TODO: https://github.com/sbt/sbt/issues/7718
+  // Note that this test currently relies on a fake target artifact that's checked in
   test("return completions for user classes") {
     val completionStr = """{ "query": "testOnly org." }"""
     svr.sendJsonRpc(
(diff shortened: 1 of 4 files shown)


server-test/src/test/scala/testpkg/ServerCompletionsTest.scala
```diff
@@ -33,6 +33,7 @@ class ServerCompletionsTest extends AbstractServerTest {
     })
   }
 
+  /*
```
adpi2, line 36: I think you commented this one out because it's flaky. Maybe add `// TODO fix flakiness`?

adpi2 (commented):
Does it also mean the default prompt will always be suffixed by `-root` as in `sbt:sbt-vimquit-root`? Is that really what we want?

Also if we have a `val plugin = ProjectMatrix.in(file("."))`, why do we also have a root? Intuitively I would expect one of the `plugin` to be the root, with no aggregate, which seems fine to me.

eed3si9n:
I guess it will create a root because maybe the composite projects aren't expanded when the root is created? More importantly though this problem can happen even when the plugin is located in a subdirectory as long as the `name` is sbt-vimquit.

adpi2:
> I guess it will create a root because maybe the composite projects aren't expanded when the root is created?

Or because the base directory of a project in a matrix is `./.sbt/matrix/plugin2_12` and never `.`.

> More importantly though this problem can happen even when the plugin is located in a subdirectory as long as the name is sbt-vimquit.

It's actually true for all projects, not only the root: if 2 projects have the same name, platform and scalaVersion then they have the same `outputDirectory`.

adpi2 (commented):
I am still not totally convinced by this PR because of single project build. For those, the name of the root project should be `sbt-vimquit` not `sbt-vimquit-root`. Otherwise it is going the break the `artifactName` of those projects.

Can we special-case them with something like:
```scala
if (existingIds.size == 0) idBase else idBase + "-root"
```

server-test/src/test/scala/testpkg/ServerCompletionsTest.scala
eed3si9n: Created an issue. I'll link to it https://github.com/sbt/sbt/issues/7718

eed3si9n:
> Can we special-case them with something like:
> 
> ```scala
> if (existingIds.size == 0) idBase else idBase + "-root"
> ```

Addressed this in 4d691deab530d8f5637c341c8a10e8f3ad2fd907

adpi2 (approved):

PR #7717, opened 2024-10-03
[2.x] remove deprecated `[this]` qualifier

xuwei-k:
https://docs.scala-lang.org/scala3/reference/dropped-features/this-qualifier.html


eed3si9n (approved):
Thanks!

Issue #7737, opened 2024-10-08
[2.x] `import sbt.given` and source compatibility with sbt 1.x


adpi2:
In sbt 2.0.0-M2 we translated a few `implicit def` to `given` or `Conversion`. Those new context abstractions in Scala 3 requires the explicit `import sbt.given`.

This makes it hard to cross-compile some plugins, because the `import sbt.given` is rejected by the Scala 2 compiler:
```
error] /home/piquerez/github/sbt/sbt-buildinfo/src/main/scala/sbtbuildinfo/BuildInfoPlugin.scala:5:8: object given is not a member of package sbt
[error] import sbt.given
[error]        ^
[error] one error found
```

I suggest we revert back to using `implicit def` in sbt 2.x, to fix this issue.

adpi2:
For instance, in `sbt-buildinfo`: https://github.com/sbt/sbt-buildinfo/pull/207#discussion_r1791715267

jtjeferreira:
Also in https://github.com/playframework/twirl/pull/840/files#diff-a2e196749d6ef985b6b6726915ccae8d35957839730761750e748a54bffeaef8R10

adpi2:
> Also in https://github.com/playframework/twirl/pull/840/files#diff-a2e196749d6ef985b6b6726915ccae8d35957839730761750e748a54bffeaef8R10

Interesting! So it seems Scala 2.12 is lenient on `import sbt.{given, _}` but not on `sbt.given`.

That's enough to fix the issue. But it should be mentioned in the sbt 2.x migration guide.

jtjeferreira:
> > Also in [playframework/twirl#840 (files)](https://github.com/playframework/twirl/pull/840/files#diff-a2e196749d6ef985b6b6726915ccae8d35957839730761750e748a54bffeaef8R10)
> 
> Interesting! So it seems Scala 2.12 is lenient on `import sbt.{given, _}` but not on `sbt.given`.
> 
> That's enough to fix the issue. But it should be mentioned in the sbt 2.x migration guide.

I misunderstood you then. I though you wanted to facilitate the migration/cross compilation, so we don't have to change `import sbt._` to `import sbt.{given,_}`. It was a facing this problem in some slash syntax that was failing to compile...

adpi2:
> I misunderstood you then. I though you wanted to facilitate the migration/cross compilation, so we don't have to change import sbt._ to import sbt.{given,_}. It was a facing this problem in some slash syntax that was failing to compile...

[...]

PR #7740, opened 2024-10-08
[2.x] revert SlashSyntax to old styles implicits to facilitate source compatibility with sbt 1.x

jtjeferreira:
related #7737

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -186,7 +186,7 @@ You'd need alternative DSL import since you can't rely on sbt package object.
 
 ```scala
 // for slash syntax
-import sbt.SlashSyntax0.given
+import sbt.SlashSyntax0.*
 
 // for IO
 import sbt.io.syntax._
--- a/main-settings/src/main/scala/sbt/SlashSyntax.scala
+++ b/main-settings/src/main/scala/sbt/SlashSyntax.scala
@@ -31,39 +31,67 @@ import sbt.internal.util.AttributeKey
 trait SlashSyntax:
   import SlashSyntax.*
 
+  /*
   given Conversion[ScopeAxis[Reference], RichReference] =
     (a: ScopeAxis[Reference]) => RichReference(Scope(a, This, This, This))
+   */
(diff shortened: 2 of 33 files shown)


main/src/main/scala/sbt/internal/BuildUtil.scala
```diff
@@ -93,7 +93,6 @@ object BuildUtil {
   def baseImports: Seq[String] =
     ("import _root_.scala.xml.{TopScope=>$scope}"
       :: "import _root_.sbt.*"
-      :: "import _root_.sbt.given"
```
jtjeferreira, line 96: important change: removing `_root_.sbt.given` from `baseImports`

main-settings/src/test/scala/sbt/AppendSpec.scala
jtjeferreira, line 1: unrelated question: why are these tests disabled/commented? I will enable them in a separate PR...

main-settings/src/main/scala/sbt/SlashSyntax.scala
jtjeferreira, line 1: The important changes are in this file.

DEVELOPING.md
```diff
@@ -186,7 +186,7 @@ You'd need alternative DSL import since you can't rely on sbt package object.
 
 ```scala
 // for slash syntax
-import sbt.SlashSyntax0.given
+import sbt.SlashSyntax0.*
```
eed3si9n, line 189: See https://www.scala-sbt.org/2.x/docs/en/changes/migrating-from-sbt-1.x.html#changing-buildsbt-dsl-to-scala-3x

```scala
import sbt.SlashSyntax0.{ given, * }
```

should work?

jtjeferreira: It works, but it makes migration of plugins more complicated...

The downside is that "old style implicits" are not idiomatic scala3. Also my idea was to move to givens in sbt 2.1, but thinking about it, that would be a big breaking change that would require all plugins to be republished...

I will close this

PR #7745, opened 2024-10-09
[2.x] remove deprecated showRelativeKey, showBuildRelativeKey, displayRelative, and showContextKey

jtjeferreira:


main-settings/src/main/scala/sbt/Def.scala
```diff
@@ -81,15 +81,7 @@ object Def extends BuildSyntax with Init[Scope] with InitializeImplicits:
...
 
-  @deprecated("Use showRelativeKey2 which doesn't take the unused multi param", "1.1.1")
   def showRelativeKey(
-      current: ProjectRef,
-      multi: Boolean,
-      keyNameColor: Option[String] = None
-  ): Show[ScopedKey[_]] =
-    showRelativeKey2(current, keyNameColor)
-
-  def showRelativeKey2(
```
adpi2, line 92: Should we keep the name `showRelativeKey2` for cross-compatibility with sbt 1.x

jtjeferreira: Indeed there is a trade off between facilitating cross-compatibility with sbt 1.x and doing all the breaking changes now...

I guess, we could keep `showRelativeKey2` for backwards compatibilty, but mark it as deprecated...

jtjeferreira: Indeed there is a trade off between facilitating cross-compatibility with sbt 1.x and doing all the breaking changes now...

I guess, we could keep `showRelativeKey2` for backwards compatibilty, but mark it as deprecated...

jtjeferreira: I added those back with an `inline` keyword.

adpi2 (approved):
LGTM

jtjeferreira:
> LGTM

Let me just change the existing calls to use the non deprecated version...

adpi2 (commented):
On second thoughts, I am not sure if it is a good idea to let the user choose between a deprecation in sbt 1.x and a deprecation in sbt 2.x. So I leave the final decision to @eed3si9n

eed3si9n:
It's kind of case by case, but generally I think we should weigh the cost of breaking changes with whatever the benefit we're buying with it, like simplicity. In this case, `showRelativeKey2` is mostly used internally, so I suggest we keep the name as-is.

SethTisue:
> I am not sure if it is a good idea to let the user choose between a deprecation in sbt 1.x and a deprecation in sbt 2.x

dunno about this bit of API in particular, but in general, it seems normal to me that a cross-built plugin might have to silence some warnings? if you're thinking of relaxing on that, I encourage you to go ahead :-)

by analogy, in cross-built Scala projects it's often not achievable to be warnings-free on all versions and that doesn't bother us. (though I don't deny that a user is sometimes bothered.) it should be achievable to be warnings-free in the latest version, but the others are 🤷

adpi2:
> It's kind of case by case, but generally I think we should weigh the cost of breaking changes with whatever the benefit we're buying with it, like simplicity.

On the other hand, if we have a consistent policy about breaking changes, we can more easily clean all the deprecated methods without thinking too much. To know whether it is only used internally or not requires knowledge, or effort, and is mostly hypothetical. If we decide on a consistent policy, I can submit a PR that removes all the deprecated methods. We would see earlier what breaks and what we need to fix or revert.

> In this case, showRelativeKey2 is mostly used internally, so I suggest we keep the name as-is.

I also think we should keep the name as-is, for this one and all the other ones. It's easier for us, not so bad for the user, and it produces no warnings.

eed3si9n:
> To know whether it is only used internally or not requires knowledge, or effort, and is mostly hypothetical.

I use GitHub code search. It doesn't capture private repos but hopefully good enough signal since plugins are mostly open source.

jtjeferreira:
I will do the changes either in the weekend or next week

PR #7746, opened 2024-10-09
[1.x] Submit dependency graph for security alert & Fix CI

Friendseeker:
We have the dependency submission mechanism, but the branch name used is wrong, hence dependency graph on 1.10.x branch is never submitted.

This PR fixes the branch name

Also fixes CI failures due to sbt command not found

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,6 +86,8 @@ jobs:
       with:
         distribution: "${{ matrix.distribution }}"
         java-version: "${{ matrix.java }}"
+    - name: Setup SBT
+      uses: sbt/setup-sbt@v1
     - name: Set up Python 3.12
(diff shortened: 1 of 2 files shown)


eed3si9n (commented):
Could you add develop in there too?

Friendseeker:
Shall do it. Btw going slightly offtopic, why are CI failing... This PR just patched an existing CI job.

I do feel like there might be something gravely wrong with the [CI runs](https://github.com/sbt/sbt/actions/runs/11260269116/job/31311083977) for which we need to investigate. It is probably a mix of many issues.

When existing CIs are flaky it makes bump CI to JDK 21 harder as we will not able to easily tell which CI failures are introduced after bumping JDK and which failures are always there.

eed3si9n (approved):
Thanks!

PR #7753, opened 2024-10-11
[2.x] Replace tuple with proper record type for licenses

adpi2:
This is a merge of https://github.com/sbt/librarymanagement/pull/382, originally by @mdedetrich

@mdedetrich more changes are needed in the the sbt part to make it compile. If you can finish the work, that would be great.


mdedetrich:
Thanks looking into it now

mdedetrich:
@adpi2 I am trying to fix this but I don't have access to your branch to add commits

adpi2:
Sorry for that. Could you push to your fork and create another PR then?

mdedetrich:
Sure will do

mdedetrich:
Pull request is here https://github.com/adpi2/sbt/pull/36

mdedetrich:
@adpi2 This should fix the issue with the latest scripted test https://github.com/adpi2/sbt/pull/37

BTW I just noticed now, but is it intended that librarymanagement is going to be inlined into `sbt` proper rather than sitting as its own module/library?

mdedetrich:
@adpi2 Fix for the latest CI run https://github.com/adpi2/sbt/pull/38 

Sorry that this is so iterative, I am currently have issues loading the sbt project properly on my machine right now so relying a bit on CI

mdedetrich:
@adpi2 And again, https://github.com/adpi2/sbt/pull/39. Really hope that this is the last one...

mdedetrich:
@adpi2 Everything passes now!

main/src/main/scala/sbt/Keys.scala
```diff
@@ -383,7 +383,7 @@ object Keys {
   val description = settingKey[String]("Project description.").withRank(BSetting)
   val homepage = settingKey[Option[URL]]("Project homepage.").withRank(BSetting)
   val startYear = settingKey[Option[Int]]("Year in which the project started.").withRank(BMinusSetting)
-  val licenses = settingKey[Seq[(String, URL)]]("Project licenses as (name, url) pairs.").withRank(BMinusSetting)
+  val licenses = settingKey[Seq[LicenseInfo]]("Project licenses.").withRank(BMinusSetting)
```
eed3si9n, line 386: While I'm +1 in principle, we should try to make the transition as smooth as possible to minimize the disruption cost. For example we should consider introducing a givens converter so sbt 1.x tuple would work out of box, and maybe show a warning.

sbt-app/src/sbt-test/dependency-management/cache-update/build.sbt
```diff
@@ -23,7 +23,7 @@ lazy val root = (project in file("."))
       ),
       version := "0.3.1-SNAPSHOT",
       description := "An HTTP client for Scala with Async Http Client underneath.",
-      licenses := Seq("Apache 2" -> url("http://www.apache.org/licenses/LICENSE-2.0.txt")),
+      licenses := Seq(LicenseInfo("Apache 2", uri("http://www.apache.org/licenses/LICENSE-2.0.txt"), None, None)),
```
eed3si9n, line 26: `None, None` here is not nice. Could you add since tag to produce apply with two fields plz?

lm-core/src/main/contraband/librarymanagement.json
```diff
@@ -733,6 +733,19 @@
         { "name": "devConnection", "type": "String?",     "default": "None", "since": "0.0.1" }
       ]
     },
+    {
+      "name": "LicenseInfo",
```
eed3si9n, line 737: This probably should be called `License`. Note that we already have a [License](https://github.com/sbt/sbt/blob/develop/lm-core/src/main/scala/sbt/librarymanagement/License.scala) object that lists popular licenses:

```scala
/**
 * Commonly used software licenses
 * Names are SPDX ids:
 * https://raw.githubusercontent.com/spdx/license-list-data/master/json/licenses.json
 */
object License {
  lazy val Apache2: (String, URL) =
    ("Apache-2.0", new URI("https://www.apache.org/licenses/LICENSE-2.0.txt").toURL)

  lazy val MIT: (String, URL) =
    ("MIT", new URI("https://opensource.org/licenses/MIT").toURL)

  lazy val CC0: (String, URL) =
    ("CC0-1.0", new URI("https://creativecommons.org/publicdomain/zero/1.0/legalcode").toURL)

  def PublicDomain: (String, URL) = CC0

  lazy val GPL3_or_later: (String, URL) =
    ("GPL-3.0-or-later", new URI("https://spdx.org/licenses/GPL-3.0-or-later.html").toURL)
}
```



We should let the companion object expand License Extra so we don't lose that.

eed3si9n (changes requested):

eed3si9n:
I have another PR here - https://github.com/sbt/sbt/pull/7927

PR #7760, opened 2024-10-12
[1.x] Bump CI to JDK 21

### LINKED ISSUE #7731: Bump CI to JDK 21
Ref https://github.com/scala/docs.scala-lang/pull/3082#issuecomment-2395625368

/cc @Friendseeker

Friendseeker:
Closes #7731


Friendseeker:
Shall reduce the scope of it. Let's leave `launcher-package` for later.

Friendseeker:
The only outstanding scripted test is `stdlib-unfreeze`, otherwise everything works for JDK 21.

Are there any standard library methods that are not in Scala Library 2.13.11 but are added later? As a part of the test checks whether sbt properly handles methods newly added to standard library for macro expansion.

SethTisue:
> Are there any standard library methods that are not in Scala Library 2.13.11 but are added later? 

Yes. Nothing public in the source code, but we've added things that are `private[scala]` or some subpackage. (And those are then public in bytecode, as expected.)

Friendseeker:
> Yes. Nothing public in the source code, but we've added things that are `private[scala]` or some subpackage. (And those are then public in bytecode, as expected.)

Thanks for the heads up. I guess we can do something similar to [this](https://github.com/sbt/zinc/blob/95ec4bcaa78d26369c5f2c8d4cd8602836f2026b/internal/compiler-bridge/src/main/scala/scala/reflect/ReflectAccess.scala) to access these private methods.

And indeed after some searching, I found one such private method: https://github.com/scala/scala/commit/8077afc3c4ccdcf0f785f9d732977e1eb6437f3f

SethTisue (commented):
nice work — LGTM! (though my eyeballs on this aren't worth that much compared to Eugene's and Adrien's)

I really appreciate you tackling this

eed3si9n (approved):
Nice work!

PR #7764, opened 2024-10-15
[1.x] Regenerate contraband files

Friendseeker:
After `contraband` is updated, the `contraband` generated files need to be regenerated.


Friendseeker:
CI failure certainly false positive. Only change made in the PR is to update autogenerated contraband file

adpi2 (approved):

Issue #7768, opened 2024-10-15
[2.x] `Type Mismatch Error` if too many `.value` call


xuwei-k:
## steps

`project/build.properties`

```properties
sbt.version=2.0.0-M2
```

`project/Example.scala`

```scala
import sbt._

object Example {
  val a1 = settingKey[Seq[Int]]("")
  val a2 = settingKey[Option[Int]]("")
  val a3 = settingKey[Int]("")
  val a4 = settingKey[Int]("")
  val a5 = settingKey[Int]("")
  val a6 = settingKey[Int]("")
  val a7 = settingKey[Int]("")
  val a8 = settingKey[Int]("")
  val a9 = settingKey[Int]("")
  val a10 = settingKey[Int]("")
  val a11 = settingKey[Int]("")
  val a12 = settingKey[Int]("")
  val a13 = settingKey[Int]("")
  val a14 = settingKey[Int]("")
  val a15 = settingKey[Int]("")
  val a16 = settingKey[Int]("")
  val a17 = settingKey[Int]("")
  val a18 = settingKey[Int]("")
  val a19 = settingKey[Int]("")
  val a20 = settingKey[Int]("")
  val a21 = settingKey[Int]("")
  val a22 = settingKey[Int]("")
  val a23 = settingKey[Int]("")

  val settings = Seq(
    a1 := Nil,
    a2 := None,
    a3 := 3,
    a4 := 4,
    a5 := 5,
    a6 := 6,
    a7 := 7,
    a8 := 8,
    a9 := 9,
    a10 := 10,
    a11 := 11,
    a12 := 12,
    a13 := 13,
    a14 := 14,
    a15 := 15,
    a16 := 16,
    a17 := 17,
    a18 := 18,
    a19 := 19,
    a20 := 20,
    a21 := 21,
    a22 := 22,
    a23 := 23,
    TaskKey[String]("x") := List(
      a1.value,
      a2.value,
      a3.value,
      a4.value,
      a5.value,
      a6.value,
      a7.value,
      a8.value,
      a9.value,
      a10.value,
      a11.value,
      a12.value,
      a13.value,
      a14.value,
      a15.value,
      a16.value,
      a17.value,
      a18.value,
      a19.value,
      a20.value,
      a21.value,
      a22.value,
      a23.value,
    ).toString
  )
}
```

launch sbt

## problem

```
[error] -- [E007] Type Mismatch Error: sbt-2-many-value-error-example/project/Example.scala:52:4 
[...]

PR #7773, opened 2024-10-17
[2.x] fix type error if too many `.value`

xuwei-k:
- fix https://github.com/sbt/sbt/issues/7768


eed3si9n (approved):
Thanks!

xuwei-k:
note: fixed since Scala 3.6.3-RC1

https://github.com/scala/scala3/commit/78cff1bd5e49925f620a6aa530f96ea08df6cc95

eed3si9n:
Nice. We should revert the workaround when Scala 3.6.3 comes out then.

PR #7775, opened 2024-10-18
Default to sbtn for sbt 2.x

eed3si9n:
This changes the runner script logic to default to sbtn when sbt version is 2.x.

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -1127,6 +1127,7 @@ object NetworkClient {
         sbtArguments += s"-Dsbt.script=$sbtScript"
       }
     }
+    if (!sbtArguments.contains("--server")) sbtArguments += "--server"
     new Arguments(
       base,
       sbtArguments.toSeq,
--- a/launcher-package/src/universal/bin/sbt.bat
+++ b/launcher-package/src/universal/bin/sbt.bat
@@ -890,22 +890,22 @@ for /F "delims=.-_ tokens=1-2" %%v in ("!sbtV!") do (
   set sbtBinaryV_1=%%v
   set sbtBinaryV_2=%%w
 )
-set native_client_ready=
+rem default to run_native_client=1 for sbt 2.x 
 if !sbtBinaryV_1! geq 2 (
-  set native_client_ready=1
+  if !sbt_args_client! equ 0 (
+    set run_native_client=
+  ) else (
+    set run_native_client=1
+  )
 ) else (
   if !sbtBinaryV_1! geq 1 (
     if !sbtBinaryV_2! geq 4 (
-      set native_client_ready=1
+      if !sbt_args_client! equ 1 (
+        set run_native_client=1
+      )
     )
   )
 )
-if !native_client_ready! equ 1 (
-  if !sbt_args_client! equ 1 (
-    set run_native_client=1
-  )
-)
-set native_client_ready=
 
 exit /B 0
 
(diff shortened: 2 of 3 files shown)


adpi2 (approved):

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -1127,6 +1127,7 @@ object NetworkClient {
         sbtArguments += s"-Dsbt.script=$sbtScript"
       }
     }
+    if (!sbtArguments.contains("--server")) sbtArguments += "--server"
```
tgodzik, line 1130: I think this is causing Metals to stop working with the embedded launcher jar:
```
2024.10.21 14:06:01 INFO  BSP server: [info] server was not detected. starting an instance
2024.10.21 14:06:01 INFO  BSP server: [warn] server is started using sbt-launch jar directly
2024.10.21 14:06:01 INFO  BSP server: [warn] this is not the recommended way: .sbtopts and .jvmopts files are not loaded and SBT_OPTS is ignored
2024.10.21 14:06:01 INFO  BSP server: [warn] either upgrade sbt to its latest version or make sure it is accessible from $PATH, and run 'sbt bspConfig'
2024.10.21 14:06:01 INFO  BSP server: Unrecognized option: --server
2024.10.21 14:06:01 INFO  BSP server: Error: Could not create the Java Virtual Machine.
2024.10.21 14:06:01 INFO  BSP server: Error: A fatal exception has occurred. Program will exit.
```
should this happen now? I can't seem to figure out where `--server` is added. The other option is to use the scripts, but that's a bigger change and we were pretty happy with how things worked with the launcher jar

eed3si9n: > should this happen now?

No I only meant this to affect `sbt` shell script invoking either the sbtn or sbt server. I didn't intend BSP to break. How is Metals invoking Network client? @adpi2 Do you have suggestion on how we should proceed?

tgodzik: We invoke whatever is in the sbt.json generated by bspConfig method:
```
{"name":"sbt","version":"1.10.3","bspVersion":"2.1.0-M1","languages":["scala"],"argv":["/usr/lib/jvm/openjdk21/bin/java","-Xms100m","-Xmx100m","-classpath","/home/tgodzik/Documents/metals/tests/slow/target/e2e/sbt-server/generate/.bsp/sbt-launch.jar","xsbt.boot.Boot","-bsp","--sbt-launch-jar=/home/tgodzik/Documents/metals/tests/slow/target/e2e/sbt-server/generate/.bsp/sbt-launch.jar"]}⏎
```

eed3si9n: Splitting this out as a bug - https://github.com/sbt/sbt/issues/7792

Friendseeker: Is it possible to add some Metals & sbt integration test (on sbt or Metals repo) to automatically catch such issue?

eed3si9n: Since we're the one generating the BSP JSON, we should be running integration test of sbt vs sbt.

Friendseeker:
@eed3si9n Forgot to ask it back then but what is the behavior for Windows ARM? Does it default to x86_64 build of sbtn or does it default to the jar?

Just want to make sure we have special handling for this particular case.

Issue #3857, opened 2018-01-07
state.contra does not specify @codecPackage annotation


xuwei-k:
## steps

https://github.com/sbt/sbt/blob/f2f44c738c605aa1cc12ac8c153dcb92bfd5e2aa/main-command/src/main/contraband/state.contra#L1-L4

## problem

some generated sjsonnew Formats classes does not have package.

- https://github.com/sbt/sbt/blob/f2f44c738c605aa1cc12ac8c153dcb92bfd5e2aa/main-command/src/main/contraband-scala/ExecFormats.scala#L1-L7

- https://github.com/sbt/sbt/blob/f2f44c738c605aa1cc12ac8c153dcb92bfd5e2aa/main-command/src/main/contraband-scala/CommandSourceFormats.scala#L1-L7

- https://github.com/sbt/sbt/blob/f2f44c738c605aa1cc12ac8c153dcb92bfd5e2aa/main-command/src/main/contraband-scala/ConnectionTypeFormats.scala#L1-L7

- https://github.com/sbt/sbt/blob/f2f44c738c605aa1cc12ac8c153dcb92bfd5e2aa/main-command/src/main/contraband-scala/ServerAuthenticationFormats.scala#L1-L7

## expectation

set `@codecPackage("sbt")` or   `@codecPackage("sbt.something")` ? (but binary compatibility ... 😢 )

## notes

sbt version: [1.0.0-M5](https://github.com/sbt/sbt/commit/66301dbaf188c0adce666d8db9e0c0900c8bdfc1) to 1.1.0

dwijnand:
Yeah due to binary compatibility I think we're stuck with this.

I feel like the root cause is that contraband _allows_ this to be the case. I think you should have to opt into using the empty package if for whatever bizarre reason you want that.

PR #7780, opened 2024-10-18
[2.x] add `@codecPackage` annotation to `state.contra`

xuwei-k:
- fix https://github.com/sbt/sbt/issues/3857


eed3si9n (approved):
Thanks!

PR #7805, opened 2024-10-22
[2.x] delete scala-2 sources

xuwei-k:


eed3si9n:
I was using `src/main/scala-2` as a way of temporarily directory to keep some of the code that I didn't want to immediately port, so it's possible we might still want to keep some of these, like linting and tests.

Issue #7790, opened 2024-10-20
Provide a non-reproducible version of `ConsistentAnalysisFormat` by default?


Friendseeker:
Currently some of Zinc's internal logic still relies on timestamp provided by `ConsistentAnalysisFormat`, which is wiped during serialization. This causes issue with invalidation of trait private API change and macros, as evidenced by some of Zinc's internal scripted tests failing when default format is switched to `ConsistentAnalysisFormat`.

While users are not mass reporting issues, should we proactively offer non-reproducible version of `ConsistentAnalysisFormat` by default and tell users to opt in when necessary? Another option is to test more and back port https://github.com/sbt/zinc/pull/1430.

eed3si9n:
I guess we can use the non-consistent version by default in sbt 1.x series, and use the consistent version in sbt 2.x series?

eed3si9n:
Sorry I wasn't clear. I meant making the older Protobuf binary the default for sbt 1.x if it requires timestamp etc.

Friendseeker:
> Sorry I wasn't clear. I meant making the older Protobuf binary the default for sbt 1.x if it requires timestamp etc.
> 
> 

I see. Shall close the PR at Zinc end then.

PR #7807, opened 2024-10-22
[1.x] Let Consistent Analysis to be opt-in by default

Friendseeker:
Consistent Analysis, while having better performance on very large codebases, results in less accurate invalidation due to omission of timestamp information. For small codebases the performance improvement is negligible (<200ms per incremental compile), yet the drawback of less accurate invalidation remains.

As per [discussion](https://github.com/sbt/sbt/issues/7790#issuecomment-2425284663) with Eugene, we shall make it opt-in by default instead of opt-out.

(As reference consistent analysis saves ~500ms of incremental compilation time for `scala2` according to Stefan's [benchmark data](https://github.com/sbt/zinc/pull/1326#issue-2067847403) collected on a M2 Pro MacBook)


eed3si9n (approved):
Thanks!

PR #7811, opened 2024-10-22
[1.x] Use `jline-terminal-jni` to replace deprecated Jansi & JNA Provider

### LINKED ISSUE #7793: [1.10.3] [Windows] Jansi deprecation notice emitted when opening sbt
Version: 1.10.3

```
[info] [launcher] getting org.scala-sbt sbt 1.10.3  (this may take some time)...
[info] [launcher] getting Scala 2.12.20 (for sbt)...
Oct. 21, 2024 10:48:00 A.M. org.jline.utils.Log logr
WARNING: The terminal provider jansi has been deprecated, check your configuration. This warning can be disabled by setting the system property org.jline.terminal.disableDeprecatedProviderWarning to true.
```

Here's quote from [JLine repo](https://github.com/jline/jline3)

> The [Jansi](https://github.com/fusesource/jansi) library is a library specialized in supporting ANSI sequences in terminals. Historically, the JNI methods used by JLine were provided by Jansi. In order to minimize the maintenance cost, Jansi has been merged into JLine 3.25.
> 
> This provider has been deprecated in 3.26 in favor of the JNI provider.

Friendseeker:
### Issue

JLine 3 deprecated Jansi & JNA terminal provider in 3.26.0.

### Solution

Use JNI terminal provider, which is the recommended default by JLine 3.

> Since JLine 3.24.0, JLine provides its own JNI based provider and native libraries. This is the best default choice, with no additional dependency.

(c.c. https://github.com/jline/jline3)

### Testing

I tested manually on Windows that `sbtn, sbt` (via Powershell & cmd) and IntelliJ SBT shell works the same way as before (with smart tab completion / history working). Will do more testing on a Apple Silicon Macbook once I get access to it later today.

More testing from PR reviewers are welcomed, as I am not a heavy sbt shell user therefore I am probably not testing all workflows. Suffices to checkout the PR branch, run `sbt publishLocalBin`, wipe out `~\.sbt\boot\scala-2.12.20\org.scala-sbt\sbt\1.10.4-SNAPSHOT`if exists then start a new sbt project with `build.properties` set to `sbt.version = 1.10.4-SNAPSHOT`.

Would also appreciate comments on what type of workflows should be tested & how to test them.

Closes https://github.com/sbt/sbt/issues/7793

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -378,8 +378,8 @@ lazy val utilLogging = (project in file("internal") / "util-logging")
       Seq(
         jline,
         jline3Terminal,
-        jline3JNA,
-        jline3Jansi,
+        jline3JNI,
+        jline3Native,
         log4jApi,
         log4jCore,
         disruptor,
--- a/internal/util-logging/src/main/scala/sbt/internal/util/JLine3.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/JLine3.scala
@@ -19,9 +19,7 @@ import org.jline.terminal.{ Attributes, Size, Terminal => JTerminal }
 import org.jline.terminal.Attributes.{ InputFlag, LocalFlag }
 import org.jline.terminal.Terminal.SignalHandler
 import org.jline.terminal.impl.{ AbstractTerminal, DumbTerminal }
-import org.jline.terminal.impl.jansi.JansiTerminalProvider
 import org.jline.terminal.spi.{ SystemStream, TerminalProvider }
-import org.jline.utils.OSUtils
 import scala.collection.JavaConverters._
 import scala.util.Try
 import java.util.concurrent.LinkedBlockingQueue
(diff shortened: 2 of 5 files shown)


eed3si9n:
When JLine doesn't work, it usually breaks in a spectacular way, so if you run a hello world build and get history showing up, that probably means it works fine on that operating system. I usually test on macOS.

Friendseeker:
> When JLine doesn't work, it usually breaks in a spectacular way, so if you run a hello world build and get history showing up, that probably means it works fine on that operating system. I usually test on macOS.

Just tested sbtn, sbt & IntelliJ SBT shell on my M1 MacBook Pro. Confirmed that tab completion and history is also working. Your insight that JLine tend to either just work or break in a spectacular way saved me a lot of hassle.

If deemed necessary I can try to test the change on Linux (probably rent a cloud linux server to do testing), but since both macOS and Linux are UNIX, terminal that work for macOS likely won't suddenly break on Linux.

eed3si9n (approved):
Thanks!

PR #7824, opened 2024-10-24
fix: [1.x] Fixes BSP

### LINKED ISSUE #7792: [1.10.3] Metals stopped working with the embedded launcher jar
## steps

Invoke BSP server as specified in bsp JSON that we generated.

## problem
             
```
2024.10.21 14:06:01 INFO  BSP server: [info] server was not detected. starting an instance
2024.10.21 14:06:01 INFO  BSP server: [warn] server is started using sbt-launch jar directly
2024.10.21 14:06:01 INFO  BSP server: [warn] this is not the recommended way: .sbtopts and .jvmopts files are not loaded and SBT_OPTS is ignored
2024.10.21 14:06:01 INFO  BSP server: [warn] either upgrade sbt to its latest version or make sure it is accessible from $PATH, and run 'sbt bspConfig'
2024.10.21 14:06:01 INFO  BSP server: Unrecognized option: --server
2024.10.21 14:06:01 INFO  BSP server: Error: Could not create the Java Virtual Machine.
2024.10.21 14:06:01 INFO  BSP server: Error: A fatal exception has occurred. Program will exit.
```

## notes

Likely this is caused by #7775

_Originally posted by @tgodzik in https://github.com/sbt/sbt/pull/7775#discussion_r1808995003_

eed3si9n:
Fixes #7792

## Problem
Calling -bsp with --sbt-launch-jar causes "Unrecognized option: --server". This is because --server is passed in twice since #7775

## Solution
Since the server invocation by BSP client already passes in --server, we can remove the extra --server append.

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -1127,7 +1127,6 @@ object NetworkClient {
         sbtArguments += s"-Dsbt.script=$sbtScript"
       }
     }
-    if (!sbtArguments.contains("--server")) sbtArguments += "--server"
     new Arguments(
       base,
       sbtArguments.toSeq,


adpi2 (approved):
I am not sure to understand why this `--server` option was added here and where it is added again.

But I guess you checked that it is working with or without the `--sbt-launch-jar` option so LGTM.

BillyAutrey (approved):

eed3si9n:
> I am not sure to understand why this `--server` option was added here and where it is added again.

This is a flag mostly intended for the Bash script to indicate you want to invoke sbt server. When I made sbtn the default I was concerned with creating an infinite loop as sbtn would call `sbt` and then it would call sbtn again. So I added `--server`. The problem is that I had the same idea before so it was already there.

PR #7825, opened 2024-10-24
[1.x] Bump `TEST_SBT_VER` to 1.10.3 & remove unused CI variables

Friendseeker:
Previously a bunch of CI tests are testing against sbt 1.5.0. This PR bumps the version to latest 1.10.3


adpi2:
I guess this was intentional to test that the new script and launcher are still compatible with older sbt versions. @eed3si9n is that correct?

eed3si9n (approved):
Thanks!

eed3si9n:
We should rename 'Build and test (6)' to multirepo integration test, but basically it's checking that we can build zinc and lm code against the branch, and that we can build sbt main against zinc and lm HEADs, in part because util will affect them.

PR #7830, opened 2024-10-25
[1.x] Use generic early output file name

### LINKED ISSUE #7767: `usePipelining` prevents incremental compilation from working when `version` is changed
## steps

Here's a minimal project to demonstrate the problem, including steps to reproduce: https://github.com/SakulK/sbt-usePipelining-version-interaction

## problem

Enabling `usePipelining` means changing the `version` of a projects prevents incremental compilation from working, because of a change in the scalac options. This is especially unfortunate when using the `GitVersioning` plugin, then the `version` changes with every commit.

## expectation

Maybe the `earlyOutput` path that is used for `-Ypickle-write` option could skip the `version` part? That should resolve this issue, but I'm not sure if it could introduce another one.

Friendseeker:
### Issue

While analysis is stored as `inc_compile.zip`, early output has elaborate file name such as `project_3-0.1.0-SNAPSHOT.jar`. Since the file name contains version, as a result for user with `GitVersioning` plugin, the version changes every commit therefore due to file name mismatch Zinc forces full recompilation.

### Solution

Use a generic filename. Since `inc_compile.zip` did not seem to have caused any problem, this PR changes early output file name to `early.jar`.

Closes https://github.com/sbt/sbt/issues/7767

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -178,7 +178,8 @@ def mimaSettingsSince(versions: Seq[String]): Seq[Def.Setting[_]] = Def settings
     exclude[DirectMissingMethodProblem]("sbt.PluginData.this"),
     exclude[IncompatibleResultTypeProblem]("sbt.EvaluateTask.executeProgress"),
     exclude[DirectMissingMethodProblem]("sbt.Keys.currentTaskProgress"),
-    exclude[IncompatibleResultTypeProblem]("sbt.PluginData.copy$default$10")
+    exclude[IncompatibleResultTypeProblem]("sbt.PluginData.copy$default$10"),
+    exclude[DirectMissingMethodProblem]("sbt.Defaults.configArtifactPathSetting"),
   ),
 )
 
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -178,7 +178,8 @@ def mimaSettingsSince(versions: Seq[String]): Seq[Def.Setting[_]] = Def settings
     exclude[DirectMissingMethodProblem]("sbt.PluginData.this"),
     exclude[IncompatibleResultTypeProblem]("sbt.EvaluateTask.executeProgress"),
     exclude[DirectMissingMethodProblem]("sbt.Keys.currentTaskProgress"),
-    exclude[IncompatibleResultTypeProblem]("sbt.PluginData.copy$default$10")
+    exclude[IncompatibleResultTypeProblem]("sbt.PluginData.copy$default$10"),
+    exclude[DirectMissingMethodProblem]("sbt.Defaults.configArtifactPathSetting"),
```
Friendseeker: I have no idea why the exclusion was even needed as `configArtifactPathSetting` is a `private[sbt]` method. But if I don't add it, MIMA complains.

I guess it is because `private[sbt]` is public in bytecode, hence MIMA still flags it out of precaution.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1813,23 +1814,6 @@ object Defaults extends BuildCommon {
       excludes: ScopedTaskable[FileFilter]
   ): Initialize[Task[Seq[File]]] = collectFiles(dirs: Taskable[Seq[File]], filter, excludes)
 
-  private[sbt] def configArtifactPathSetting(
```
eed3si9n: Why not keep the function but change the impl so we don't have to add mima exclusion?

Issue #7831, opened 2024-10-25
[2.x] `sbt.internal.IvyConsole` would not compile using 3.4.1+


WojciechMazur:
The issues was found when adding sbt 2.0.0-M2 to Scala 3 Open Community Build. 

The snippet would fail to compile after upgrading to 3.4.1 or later. Initially I wanted to report it as a bug in the compiler, but I believe it works as expected and uncovers possible bug in sbt implementation.

https://github.com/sbt/sbt/blob/ea031a7fa0e039274f4eadb64c57a20727a1e5d7/main/src/main/scala/sbt/internal/IvyConsole.scala#L55-L59

I've prepared a full minimization of the problem: 
```scala

type Classpath = Seq[Attributed[HashedVirtualFileRef]]

object Append:
  trait Values[A1, -A2]
  trait Sequence[A1, -A2] extends Values[A1, A2]

  given appendSeq[T, V <: T]: Sequence[Seq[T], Seq[V]] = ???
  given appendSeqImplicit[A1, V](using ev: Conversion[V, A1]): Sequence[Seq[A1], Seq[V]] = ???
  given Sequence[Classpath, Seq[HashedVirtualFileRef]] = ???

trait Attributed[A1]
object Attributed:
  def blankSeq[T](v: Seq[T]): Seq[Attributed[T]] = ???
trait HashedVirtualFileRef
trait VirtualFile extends HashedVirtualFileRef
trait Setting[A1]:
  final def append[A2](vs: A2)(using Append.Values[A1, A2]): Setting[A1] = ???

object Test {
  val classpath: Setting[Seq[Attributed[HashedVirtualFileRef]]] = ???
  val works = classpath.append(Seq.empty[Attributed[HashedVirtualFileRef]])
  val `fails since 3.4.1` = classpath.append{
    Attributed.blankSeq(Seq.empty[VirtualFile])
  }
  // val failsAlways = classpath.append(Seq.empty[Attributed[VirtualFile]])
}
```

Before [scala/scala3/#19096](https://github.com/scala/scala3/pull/19096) both cases would use `Append.appendSeq` however it was an incorrect behaviour - `Attributed[T]` is invariant, thus `Attributed[VirtualFile]` is not a subtype of `Attributed[HashedVirtualFileRef]`. The mentioned PR was fixed checking for this condition, thus this implicit can no longer be used. 

[...]

PR #7834, opened 2024-10-26
[2.x] deps: Update to Scala 3.5.2

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7831

## Problem
There are some Scala 3 language changes that's getting stricter etc, such that sbt doesn't compile on 3.5.x series.

## Solution
This fixes some of the issues, like adding `using` and fixing https://github.com/sbt/sbt/issues/7831.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -6,7 +6,7 @@ object Dependencies {
   // WARNING: Please Scala update versions in PluginCross.scala too
   val scala212 = "2.12.20"
   val scala213 = "2.13.15"
-  val scala3 = "3.3.4"
+  val scala3 = "3.5.2"
   val checkPluginCross = settingKey[Unit]("Make sure scalaVersion match up")
   val baseScalaVersion = scala3
   def nightlyVersion: Option[String] =


project/Dependencies.scala
```diff
@@ -6,7 +6,7 @@ object Dependencies {
   // WARNING: Please Scala update versions in PluginCross.scala too
   val scala212 = "2.12.20"
   val scala213 = "2.13.15"
-  val scala3 = "3.3.4"
+  val scala3 = "3.5.2"
```
Friendseeker, line 9: Does that mean that all `utils` are in Scala 3.5.2 and correspondingly Zinc has to bumped to Scala 3.5.2? I am concerned that the downstream libraries (e.g. other build tools using Zinc) may want to stick with Scala 3 LTS.

eed3si9n: Commented on the relevant discussion 🧵 https://github.com/sbt/sbt/discussions/7194#discussioncomment-11062161

BillyAutrey (approved):

adpi2:
Can we hold this back for a few days until the conversation in https://github.com/sbt/sbt/discussions/7194#discussioncomment-11062161 settles.

adpi2 (approved):
I am okay with merging this

PR #7838, opened 2024-10-28
[1.x] Pass reload action for `watchOnTermination` callback

### LINKED ISSUE #7017: Reload during ~ quits sbt
### Method 1
1. Go into an sbt 1.7.1 project directory
2. Set `Global / onChangedBuildSource := ReloadOnSourceChanges`
3. run `sbt ~compile`
4. Edit build.sbt

### Method 2 (simpler)
1. Run `sbt ~compile` in an sbt 1.7.1 project directory
2. When you see `Press <enter> to interrupt or '?' for more options.`, hit the `L` key

## problem

SBT dutifully restarts, but it doesn't go back into ~compile. Instead it quits to the system shell.

## expectation

In sbt 1.6.x it would run ~compile again

Note: it's not specific to `compile`, the same thing seems to happen with any command.

Friendseeker:
### Issue

In watch mode, when a `Reload` action is triggered, `sbt` do not pass the `Reload` action to `watchOnTermination` callback (instead falls back to `CancelWatch` action)

Since `onTerminationImpl` (default implementation of `watchOnTermination`) does filtering based on type of action passed in, not passing `Reload` action cause unnecessary filtering resulting in #7017.

https://github.com/sbt/sbt/blob/2259598e5f60bba10740ee5ca52d67cac13878b2/main/src/main/scala/sbt/nio/Watch.scala#L554-L575

### Fix

#6870 introduces `failAction` as a mechanism to store & pass action to `watchOnTermination` callback. This PR employs same mechanism to pass watch action.

Closes #7017

### DIFF
--- a/main/src/main/scala/sbt/internal/Continuous.scala
+++ b/main/src/main/scala/sbt/internal/Continuous.scala
@@ -1129,7 +1129,7 @@ private[sbt] object Continuous extends DeprecatedContinuous {
       val callbacks: Callbacks,
       val dynamicInputs: mutable.Set[DynamicInput],
       val pending: Boolean,
-      var failAction: Option[Watch.Action],
+      var terminationAction: Option[Watch.Action],
   ) {
     def this(
         count: Int,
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

Issue #7580, opened 2024-06-13
SBT console "swallow" keypresses on JDK 22 and Windows 10


nartamonov:
## steps

Install Eclipse Temurin 22.0.1+8 in Windows 10. Run the SBT console for any project from the `cmd.exe` terminal. Then try to work in the console: enter different commands, try to insert into the console text from the clipboard.

## problem

The SBT console "swallows" (skips) certain key strokes: symbols do not appear in the console. Also does not work properly copy and paste.

## cause and workaround

The problem seems to be that JDK 22 has [changed](https://inside.java/2023/07/31/quality-heads-up/) the behavior of `System.console()`:

> In JDK 22, System.console() has been changed to return a Console with enhanced editing features that improve the experience of programs that use the Console API. In addition, System.console() now returns a Console object when the standard streams are redirected or connected to a virtual terminal. Prior to JDK 22, System.console() instead returned null for these cases. This change may impact code that checks the return from System.console() to test if the JVM is connected to a terminal. If required, the -Djdk.console=java.base flag will restore the old behavior where the console is only returned when it is connected to a terminal. Starting JDK 22, one could also use the new Console.isTerminal() method to test if the console is connected to a terminal.

Setting the `-Djdk.console=java.base` option in `.jvmopts` solves the problem: SBT starts working normally.

Friendseeker:
![image](https://github.com/user-attachments/assets/36ca802f-403a-4574-ae2b-e2b2269073b8)

![image](https://github.com/user-attachments/assets/8ea2ad7f-4afb-4b9c-886e-8a5d5205c776)


Not reproducible on sbt 1.10.4

Feel free to leave a comment & reopen the issue if the issue is still reproducible on your end

azdrojowa123:
@Friendseeker
Hello, 
I'm working on resolving this issue in the sbt shell within the Scala plugin and this has caused me to look at this issue one the sbt side.  
I attached 2 vides - the first one is with JDK 17 and eveything works smoothly there. I can type letters very quickly and everything is written fine, copying also works without problems . The second video is with JDK 22. Although, if I write letters very slowly I can write "compile" correctly, but copying is totally broken. At the end of the video, when I copied "compile" and "cp" was displayed I had to press enter about 10 times before it was actually done. 

https://github.com/user-attachments/assets/749cc4c2-2d29-42ca-830b-100fe85feaac


[...]

PR #7843, opened 2024-10-29
[1.x] Migrate all usages of `System.console == null`

### LINKED ISSUE #7841: [1.x] [JDK 22+] [Windows] `sbtn` spawns dumb terminal
JDK 17

```
PS D:\Repos\dependency-report-plugin> java --version
java 17.0.6 2023-01-17 LTS
Java(TM) SE Runtime Environment (build 17.0.6+9-LTS-190)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.6+9-LTS-190, mixed mode, sharing)
PS D:\Repos\dependency-report-plugin> sbtn
[info] welcome to sbt 1.10.4 (Oracle Corporation Java 17.0.6)
```

JDK 23
```
PS D:\Repos\dependency-report-plugin> sbtn
Oct 28, 2024 4:29:07 P.M. org.jline.utils.Log logr
WARNING: Unable to create a system terminal, creating a dumb terminal (enable debug logging for more information)
[info] welcome to sbt 1.10.4 (Oracle Corporation Java 23.0.1)
```

The smart features (tab completion, history) still works. It is just the warning.

Also only sbtn has the issue. sbt is fine.

While not the same issue as https://github.com/sbt/sbt/issues/7580, probably relates to it.

### LINKED ISSUE #7842: [1.x] [JDK 22+] Migrate all usage of `System.console == null`
> In JDK 22, System.console() has been changed to return a Console with enhanced editing features that improve the experience of programs that use the Console API. In addition, System.console() now returns a Console object when the standard streams are redirected or connected to a virtual terminal. Prior to JDK 22, System.console() instead returned null for these cases. This change may impact code that checks the return from System.console() to test if the JVM is connected to a terminal. If required, the -Djdk.console=java.base flag will restore the old behavior where the console is only returned when it is connected to a terminal. Starting JDK 22, one could also use the new Console.isTerminal() method to test if the console is connected to a terminal.

c.c. https://github.com/sbt/sbt/issues/7580#issue-2350841500

We currently compare `System.console` with `null` in 12 places. Thanks to JDK 22 we need to migrate away from the usages. We need to support both JDK 8 and JDK 22 in the same time though so don't know a good way to do it. Guess we use reflection to call `Console.isTerminal()` when JDK >= 22 and otherwise compare `System.console` with `null`?

Friendseeker:
### Issue

[nartamonov](https://github.com/nartamonov) pointed out that JDK 22 changed the behavior of `System.console` in #7580.

> In JDK 22, System.console() has been changed to return a Console with enhanced editing features that improve the experience of programs that use the Console API. In addition, System.console() now returns a Console object when the standard streams are redirected or connected to a virtual terminal. Prior to JDK 22, System.console() instead returned null for these cases. This change may impact code that checks the return from System.console() to test if the JVM is connected to a terminal. If required, the -Djdk.console=java.base flag will restore the old behavior where the console is only returned when it is connected to a terminal. Starting JDK 22, one could also use the new Console.isTerminal() method to test if the console is connected to a terminal.

`sbt` assumes the old behavior of `System.console` and use `System.console == null` to check if connection to terminal exists, causing issue like #7580 and #7841

### Solution

We reflectively call `Console.isTerminal()` if it exists, otherwise fall back to JDK 8 `System.console == null` comparison.

Closes #7841
Closes #7842

### DIFF
--- a/internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
+++ b/internal/util-complete/src/main/scala/sbt/internal/util/LineReader.scala
@@ -130,7 +130,7 @@ object LineReader {
           Option(mask.map(reader.readLine(prompt, _)).getOrElse(reader.readLine(prompt)))
         } catch {
           case e: EndOfFileException =>
-            if (terminal == Terminal.console && System.console == null) None
+            if (terminal == Terminal.console && !Terminal.hasConsole) None
             else Some("exit")
           case _: IOError | _: ClosedException => Some("exit")
           case _: UserInterruptException | _: ClosedByInterruptException |
--- a/internal/util-logging/src/main/scala/sbt/internal/util/JLine3.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/JLine3.scala
@@ -20,6 +20,7 @@ import org.jline.terminal.Attributes.{ InputFlag, LocalFlag }
 import org.jline.terminal.Terminal.SignalHandler
 import org.jline.terminal.impl.{ AbstractTerminal, DumbTerminal }
 import org.jline.terminal.spi.{ SystemStream, TerminalProvider }
+import sbt.internal.util.Terminal.hasConsole
 import scala.collection.JavaConverters._
 import scala.util.Try
 import java.util.concurrent.LinkedBlockingQueue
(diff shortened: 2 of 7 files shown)


eed3si9n (approved):
Thanks!

PR #7849, opened 2024-10-31
[2.x] Refactor `dependencyTree` into an input task

### LINKED ISSUE #7770: `dependencyTree` should also display/write json, not just text
`dependencyTree` currently only displays text, but it would be nice to also be able to output json.
Same with writing to file, I would like to write a json file: `dependencyTree/toFile /some/folder/deps.json`

The funny thing is `dependencyBrowseTreeHTML` already creates a `target/tree.json` which is used as source to display the dependencies in the html/browser. With `jq` you can format it nicely as well: `jq '.' target/tree.json`

So in theory as a workaround I can already do what this issue asks to with running `dependencyBrowseTreeHTML` but it's not obivous.

Making it more explicit with the `dependencyTree` command as a flag would be nicer IMHO.

Related:

- #7771

Friendseeker:
Following Eugene's [comment](https://github.com/sbt/sbt/issues/7770#issuecomment-2417281267), this PR refactors `dependencyTree` into an input task, with the aim of

- Featuring minimal additional settings key (to avoid settings key explosion)
- Supporting all capabilities of `MiniDependencyTreePlugin`
- Supporting most capabilities of the full `DependencyTreePlugin`
- Supporting `json` output to both terminal & file
- Featuring a simple API

The refactored input task is called `dependencyReport`. Old `MiniDependencyTreePlugin` settings are still kept for compatibility purpose.

#### Syntax

`dependencyReport [format] [subformat]`

#### Parameters
- `format: text|json|html|graphml`: specifies the format of the output, defaults to `text`
- `subformat`: if applicable further specifies format of output
  - `text` format: `subformat` takes `list|stat|info`, corresponding to `dependencyList`, `dependencyStats`, `dependencyLicenseInfo`. Empty `subformat` corresponds to `dependencyTree`
  -  `html` format: `subformat` takes `tree|graph`, corresponding to `browseTreeHTMLTask, browseGraphHTMLTask`
  - `json/graphml` format: does not take `subformat`
  
#### Example

```
sbt:contraband> dependencyReport
[info] org.scala-sbt:contraband_2.12:0.5.3-SNAPSHOT [S]
[info]   +-com.eed3si9n:sjson-new-scalajson_2.12:0.10.0 [S]
[info]   | +-com.eed3si9n:shaded-jawn-parser_2.12:1.3.2 [S]
[info]   | +-com.eed3si9n:shaded-scalajson_2.12:1.0.0-M4 [S]
[info]   | +-com.eed3si9n:sjson-new-core_2.12:0.10.0 [S]
[info]   |
[info]   +-org.parboiled:parboiled_2.12:2.5.1 [S]
[info]
[info] Dependency report written to D:\Repos\contraband\library\target\compile\dependencies.txt

sbt:contraband> dependencyReport json
[info] [{"text":"org.scala-sbt:contraband_2.12:0.5.3-SNAPSHOT","children":[{"text":"com.eed3si9n:sjson-new-scalajson_2.12:0.10.0","children"
:[{"text":"com.eed3si9n:shaded-jawn-parser_2.12:1.3.2","children":[]},{"text":"com.eed3si9n:shaded-scalajson_2.12:1.0.0-M4","children":[]},{"text":"com.eed3si9n:sjson-new-core_2.12:0.10.0","children":[]}]},{"text":"org.parboiled:parboiled_2.12:2.5.1","children":[]}]}]
[info] Dependency report written to D:\Repos\contraband\library\target\compile\dependencies.json
[success] Total time: 0 s, completed Oct 30, 2024, 6:11:06 p.m.

sbt:contraband> Test / dependencyReport html graph
[info] Dependency report written to D:\Repos\contraband\library\target\test\graph\graph.html
```

Closes #7770

### DIFF
--- a/main/src/main/scala/sbt/plugins/DependencyReportPlugin.scala
+++ b/main/src/main/scala/sbt/plugins/DependencyReportPlugin.scala
@@ -10,7 +10,7 @@ package sbt
 package plugins
 
 import sbt.PluginTrigger.AllRequirements
-import sbt.Project._
+import sbt.ProjectExtra.*
 import sbt.librarymanagement.Configurations.{ Compile, Test }
 
 object DependencyReportPlugin extends AutoPlugin {
--- a/main/src/main/scala/sbt/plugins/DependencyReportPlugin.scala
+++ b/main/src/main/scala/sbt/plugins/DependencyReportPlugin.scala
@@ -1,0 +1,25 @@
+/*
+ * sbt
+ * Copyright 2023, Scala center
+ * Copyright 2011 - 2022, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt
+package plugins
+
+import sbt.PluginTrigger.AllRequirements
+import sbt.Project._
+import sbt.librarymanagement.Configurations.{ Compile, Test }
+
+object DependencyReportPlugin extends AutoPlugin {
+  override def trigger: PluginTrigger = AllRequirements
+  override def requires = MiniDependencyTreePlugin
+
+  override def projectSettings: Seq[Def.Setting[?]] =
+    Seq(
+      inConfig(Compile)(DependencyTreeSettings.baseDependencyReportSettings),
+      inConfig(Test)(DependencyTreeSettings.baseDependencyReportSettings)
+    ).flatten
+}
--- a/main/src/main/scala/sbt/plugins/DependencyTreeKeys.scala
+++ b/main/src/main/scala/sbt/plugins/DependencyTreeKeys.scala
@@ -35,6 +35,12 @@ trait MiniDependencyTreeKeys {
 
 object MiniDependencyTreeKeys extends MiniDependencyTreeKeys
 
+trait DependencyReportKeys {
+  val dependencyReport = inputKey[Unit]("Generates a report of the project dependencies")
+}
+
+object DependencyReportKeys extends DependencyReportKeys
+
 abstract class DependencyTreeKeys {
   val dependencyGraphMLFile =
     settingKey[File]("The location the graphml file should be generated at")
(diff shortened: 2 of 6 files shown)


main/src/main/scala/sbt/plugins/DependencyReportPlugin.scala
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package sbt
+package plugins
+
+import sbt.PluginTrigger.AllRequirements
+import sbt.Project._
+import sbt.librarymanagement.Configurations.{ Compile, Test }
+
+object DependencyReportPlugin extends AutoPlugin {
```
eed3si9n, line 16: For sbt 2.x, I wonder if we could consolidate 3 dependency-tree plugins into one. When we in-sourced it to sbt/sbt we had to split it into two because of the amount of setting keys it introduced to all subproject slowed down large builds. If we can reduce the number of injected settings we use this opportunity to do so.

main/src/main/scala/sbt/plugins/DependencyTreeKeys.scala
```diff
@@ -35,6 +35,12 @@ trait MiniDependencyTreeKeys {
 
 object MiniDependencyTreeKeys extends MiniDependencyTreeKeys
 
+trait DependencyReportKeys {
+  val dependencyReport = inputKey[Unit]("Generates a report of the project dependencies")
```
eed3si9n, line 39: Not everything is a tree, but here too I'm thinking we should just use `dependencyTree` key in sbt 2.x, and maybe `dependencyTreeNext` or something if we introduce it to 1.x?

main/src/main/scala/sbt/plugins/DependencyReportPlugin.scala
Friendseeker: My concern is that there's some settings that are currently unsupported as they are ugly to support via command line argument (e.g. Graph customization options as part of full dependency tree plugin).

I don't know whether people actually use these settings though.

Friendseeker: @mkurz What is your opinion on Eugene's idea of consolidating 3 dependency-tree plugins into one?

I personally don't use dependency tree plugin often so need feedback from dependency tree plugin users to make major decisions like this.

SethTisue: I can confirm that the lack of consolidation is rather mystifying for users. It comes up from time to time on Discord.

Friendseeker: > I can confirm that the lack of consolidation is rather mystifying for users. It comes up from time to time on Discord.

@SethTisue I guess indeed better to go for full consolidation then. We will lose some flexibility (e.g. inability to customize graph appearance detail), but we will end up with a simplified implementation, vastly less settings key and less user confusion.

I will wait for a few days for further discussion, then proceed with full consolidation as per Eugene's comment.

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/8199

PR #7866, opened 2024-11-04
[2.x] Reduce number of long-living instances to speed up loading

adpi2:
After loading, instances of `ScopedKey`, `Scope`, `Select` or `ProjectRef` represent 25.6% of the total number of instances, and 27.3% of the heap size. The main idea of this PR is to reduce the number of duplicated instances using uniqueness caches.

## Uniqueness cache on `Scope`

It reduces the number of `Scope` instances by 94%. On its own, it slows down the loading, because `Scope.apply` becomes more expensive and it is called too many times. This can be compensated by using temporary caches in `Project.transform` and `delegates: Scope => Seq[Scope]`, to avoid calling `Scope.apply` to create too many short-lived instances.

It also reduces the number of `Select` by 92% and `ProjectRef` by 99.8%, which make them totally negligible.

Since we use a uniqueness cache we might as well pre-compute the hash code of each scope, which reduce the CPU time.

## Uniqueness cache on `ScopedKey` (uncommited)

I think it is also worth mentioning that I tried a uniqueness cache on `ScopedKey` but it was not really better. Indeed, on one hand it reduced the number of `ScopedKey` by 59.6%, but on the other hand it was counterbalanced by the number of nodes in the cache itself.

## Other changes

The first commit reduces the number of `Init.GetValue` by 48% for free.

## Result

I am still using sttp to collect those metrics, on my 12 CPUs laptop, and JDK temurin 17 (G1GC).

**It shows up a 20% speedup of the loading time compared to sbt 2.0.0-M2, and a 41% speedup compared to sbt 1.10.2. Those are rough estimations.**

### Compared to 2.0.0-M2

|          | Settings | Instances | Used Heap size | Load time |
|----------|----------|-----------|----------------|-----------|
| 2.0.0-M2 | 212961   | 20,920 K  | 532 MB         | 7949 ms   |
| This PR  | 212961   | 16,365 K  | 410 MB         | 6328 ms   |
|          | 100 %    | 78 %      | 77 %           | 80 %      |

### Compare to 1.10.2

|         | Settings | Instances | Used Heap size | Load time |
|---------|----------|-----------|----------------|-----------|
| 1.10.2  | 197031   | 23,524 K  | 675 MB         | 10629 ms  |
| This PR | 212961   | 16,365 K  | 410 MB         | 6328 ms   |
|         | 108 %    | 70 %      | 61 %           | 59 %      |

### DIFF
--- a/main-settings/src/main/scala/sbt/Scope.scala
+++ b/main-settings/src/main/scala/sbt/Scope.scala
@@ -15,6 +15,7 @@ import sbt.internal.util.Util._
 
 import sbt.io.IO
 import scala.collection.concurrent.TrieMap
+import scala.runtime.ScalaRunTime
 
 final case class Scope private (
     project: ScopeAxis[Reference],
@@ -22,6 +23,8 @@ final case class Scope private (
     task: ScopeAxis[AttributeKey[?]],
     extra: ScopeAxis[AttributeMap]
 ):
+  override val hashCode = ScalaRunTime._hashCode(this)
+
   def rescope(project: Reference): Scope = copy(project = Select(project))
   def rescope(config: ConfigKey): Scope = copy(config = Select(config))
   def rescope(task: AttributeKey[?]): Scope = copy(task = Select(task))
--- a/main-settings/src/main/scala/sbt/Scope.scala
+++ b/main-settings/src/main/scala/sbt/Scope.scala
@@ -273,33 +273,6 @@ object Scope:
 
   def showProject012Style = (ref: Reference) => Reference.display(ref) + "/"
 
-  @deprecated("No longer used", "1.1.3")
-  def transformTaskName(s: String) = {
-    val parts = s.split("-+")
-    (parts.take(1) ++ parts.drop(1).map(_.capitalize)).mkString
-  }
-
-  @deprecated("Use variant without extraInherit", "1.1.1")
-  def delegates[Proj](
-      refs: Seq[(ProjectRef, Proj)],
-      configurations: Proj => Seq[ConfigKey],
-      resolve: Reference => ResolvedReference,
-      rootProject: URI => String,
-      projectInherit: ProjectRef => Seq[ProjectRef],
-      configInherit: (ResolvedReference, ConfigKey) => Seq[ConfigKey],
-      taskInherit: AttributeKey[?] => Seq[AttributeKey[?]],
-      extraInherit: (ResolvedReference, AttributeMap) => Seq[AttributeMap]
-  ): Scope => Seq[Scope] =
-    delegates(
-      refs,
-      configurations,
-      resolve,
-      rootProject,
-      projectInherit,
-      configInherit,
-      taskInherit,
-    )
-
   // *Inherit functions should be immediate delegates and not include argument itself.  Transitivity will be provided by this method
   def delegates[Proj](
       refs: Seq[(ProjectRef, Proj)],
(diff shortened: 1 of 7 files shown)


eed3si9n:
@adpi2 Nice work!

util-collection/src/main/scala/sbt/internal/util/Settings.scala
```diff
@@ -811,9 +811,13 @@ trait Init[ScopeType]:
...
   sealed trait Keyed[S, A1] extends Initialize[A1]:
     def scopedKey: ScopedKey[S]
-    def transform: S => A1
-
     override final def dependencies = scopedKey :: Nil
+    private[sbt] override def processAttributes[A2](init: A2)(f: (A2, AttributeMap) => A2): A2 =
+      init
+  end Keyed
+
+  final class GetValue[S, A1](val scopedKey: ScopedKey[S], val transform: S => A1)
```
eed3si9n, line 819: Could you add Scala comment explaining the overall motivation of why this class is needed for later record plz? In general, in the future if someone comes along and try to refactor this, they should know to test the heap usage etc.

main-settings/src/main/scala/sbt/Scope.scala
```diff
@@ -25,15 +26,33 @@ final case class Scope(
...
   override def toString: String = this match
     case Scope(Zero, Zero, Zero, Zero) => "Global"
     case Scope(_, _, _, This)          => s"$project / $config / $task"
     case _                             => s"Scope($project, $config, $task, $extra)"
 end Scope
 
 object Scope:
-  val ThisScope: Scope = Scope(This, This, This, This)
-  val Global: Scope = Scope(Zero, Zero, Zero, Zero)
+  private val uniquenessCache = TrieMap.empty[Scope, Scope]
```
eed3si9n, line 51: We have a few places in our code base that's using Caffeine (https://github.com/ben-manes/caffeine). If it's not load-bearing to enforce uniqueness, maybe it's safer to use something like caffeine, which would use weak reference, which allows GC to pick clean them up if necessary.

util-collection/src/main/scala/sbt/internal/util/Settings.scala
adpi2: This class is not new. What I did here is to move the implementation of `apply`, `mapReferenced`,  `mapConstant` etc from the abstract `Keyed` to the more concrete `GetValue` and `KeyedInitialize`. Thus the implementation is more specialized and it does not wrap `KeyedInitialize` into `GetValue` or compose the `identity` function anymore. The git diff is a bit weird here, but the implementation itself is straightforward and I don't think it needs a comment.

Apart from that I will add comments in other parts of the code where I use caching.

main-settings/src/main/scala/sbt/Scope.scala
adpi2: Interesting! I am going to try it for `Scope`, but it also gives me some hope of reducing the number of long-living `ScopedKey`, which I'll try in another PR.

adpi2: It seems I cannot use caffeine with weak reference as a uniqueness cache, because it uses `eq` instead of `equals`, which defeats the purpose of a uniqueness cache. As an alternative I can create my own implementation of a uniqueness cache with weak reference, which would be optimized in terms of heap and CPU usage, but it is probably not worth it for `Scope`. I'll try that on `ScopedKey` in a different PR.

The uniqueness cache on `Scope` is already safe because the number of unique `Scope` is low compared to the size of the graph. It cannot explode.

eed3si9n (approved):

main-settings/src/main/scala/sbt/Scope.scala
ben-manes: Perhaps [Interner](https://github.com/ben-manes/caffeine/wiki/Interner) is what you were looking for?

adpi2: Yes indeed! Thanks @ben-manes

PR #7869, opened 2024-11-04
[1.x] Pass `useConsistent` to `staticCachedStore`

Friendseeker:
Not sure if there's a specific reason to drop `useConsistent` in `AnalysisUtil.staticCachedStore`, but if we don't have a specific reason we should pass it along. If we do have a specific reason I feel we should add a comment here to explain the rationale.


main/src/main/scala/sbt/internal/AnalysisUtil.scala
```diff
@@ -30,7 +30,7 @@ private[sbt] object AnalysisUtil {
     MixedAnalyzingCompiler.staticCachedStore(
       analysisFile = analysisFile,
       useTextAnalysis = useTextAnalysis,
-      useConsistent = false,
+      useConsistent = useConsistent,
```
eed3si9n, line 33: Interesting. I think this was an oversight on my part.

eed3si9n (approved):

PR #7872, opened 2024-11-05
[1.x] Bump Jansi to 3.27.1

Friendseeker:
Fixes https://github.com/sbt/sbt/issues/7865

Changes compared to Jansi 2.4.1 (summarized from https://github.com/jline/jline3/commits/master/jansi-core)
- https://github.com/jline/jline3/commit/f79eeb290af809d29955caee54c17d08d34466f5 (Add tripleto hexadecimal color support)

Tested locally on Windows via running `sbt` on a small test project. Tab completion & history are working.

Issue #7865, opened 2024-11-04
Windows ARM issue


arjenw:
## steps

When starting SBT I get the following stack trace:
```
java.lang.UnsatisfiedLinkError: 'long org.fusesource.jansi.internal.Kernel32.GetStdHandle(int)'
        at org.fusesource.jansi.internal.Kernel32.GetStdHandle(Native Method)
        at sbt.internal.util.WindowsSupport$.getConsoleMode(WindowsInputStream.scala:21)
        at sbt.internal.util.JLine3$.setEnableProcessInput(JLine3.scala:208)
        at sbt.internal.util.JLine3$.enterRawMode(JLine3.scala:218)
        at sbt.internal.util.Terminal$ConsoleTerminal.enterRawMode(Terminal.scala:925)
        at sbt.internal.util.Terminal$ConsoleTerminal.<init>(Terminal.scala:878)
        at sbt.internal.util.Terminal$.newConsoleTerminal(Terminal.scala:816)
        at sbt.internal.util.Terminal$.withStreams(Terminal.scala:396)
        at sbt.xMain$.withStreams$1(Main.scala:89)
        at sbt.xMain$.run(Main.scala:124)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at sbt.internal.XMainConfiguration.run(XMainConfiguration.java:59)
        at sbt.xMain.run(Main.scala:48)
        at xsbt.boot.Launch$.$anonfun$run$1(Launch.scala:149)
        at xsbt.boot.Launch$.withContextLoader(Launch.scala:176)
        at xsbt.boot.Launch$.run(Launch.scala:149)
        at xsbt.boot.Launch$.$anonfun$apply$1(Launch.scala:44)
        at xsbt.boot.Launch$.launch(Launch.scala:159)
        at xsbt.boot.Launch$.apply(Launch.scala:44)
        at xsbt.boot.Launch$.apply(Launch.scala:21)
        at xsbt.boot.Boot$.runImpl(Boot.scala:78)
        at xsbt.boot.Boot$.run(Boot.scala:73)
        at xsbt.boot.Boot$.main(Boot.scala:21)
        at xsbt.boot.Boot.main(Boot.scala)
[...]

PR #7876, opened 2024-11-08
[1.x] Remove direct dependency on `org.fusesource.jansi`

Friendseeker:
### Motivation

- Jansi development is inactive, with its last [update](https://github.com/fusesource/jansi/releases/tag/jansi-2.4.1) all the way back in Oct 12, 2023
- Jansi contains native code
- `sbt` does not use Jansi all that much

Also recent Jansi-related tickets such as https://github.com/sbt/sbt/issues/7865 would not exist if we do not depend on `org.fusesource.jansi`.

### Changes

- Remove the usage of `org.fusesource.jansi.AnsiConsole` in sbt thin client
  - Since Windows 10 introduced ANSI support, `org.fusesource.jansi.AnsiConsole` is no longer necessary on Windows.
- Remove the usage of Jansi's JNI methods, replacing them with equivalent methods from JLine3.

### Addressing diamond dependency & sbtn build concern

Eugene [pointed out](https://github.com/sbt/sbt/pull/7872#discussion_r1831528366) that changing a dependency containing native code may cause breakage if
- downstream of sbt uses the same dependency
  - historically diamond `JLine` dependency had broken `sbt console`
- sbtn build may fail & become nonfunctional

This PR addresses the concern by testing `sbtn console` on a sample Scala 2.12.20, Scala 2.13.15 and Scala 3.3.4 project, against a `sbtn` executable built against the PR branch. No regression is identified.

#### Windows

image

#### Mac

Terminal used: IntelliJ IDE Terminal

3.3.4
image

2.13.15
image

2.12.20
image

Supersedes #7872

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1170,7 +1170,6 @@ lazy val sbtClientProj = (project in file("client"))
     mimaPreviousArtifacts := Set.empty,
     crossPaths := false,
     exportJars := true,
-    libraryDependencies += jansi,
     libraryDependencies += scalatest % Test,
     Compile / mainClass := Some("sbt.client.Client"),
     nativeImageReady := { () =>
--- a/client/src/main/java/sbt/client/Client.java
+++ b/client/src/main/java/sbt/client/Client.java
@@ -9,20 +9,16 @@
 package sbt.client;
 
 import sbt.internal.client.NetworkClient;
-import org.fusesource.jansi.AnsiConsole;
 
 public class Client {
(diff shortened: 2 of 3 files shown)


Friendseeker:
Also opened #7877 (same PR but targeting 2.x branch) so if the reviewers decide that this change belongs to 2.x branch better we can merge #7877 and close #7876.

eed3si9n (approved):
Thanks!

Issue #7725, opened 2024-10-05
[2.x] ClassNotFoundException: dotty.tools.dottydoc.Main


xuwei-k:
## steps

`A.scala`

```scala
package x

class A
```

`build.sbt`

```scala
scalaVersion := "3.3.3"
```

`project/build.properties`

```scala
sbt.version=2.0.0-M2
```

run `sbt doc`

## problem

```
[error] java.lang.RuntimeException: java.lang.ClassNotFoundException: dotty.tools.dottydoc.Main
[error] 	at xsbt.DottydocRunner.run(DottydocRunner.java:65)
[error] 	at xsbt.ScaladocInterface.run(ScaladocInterface.java:11)
[error] 	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
[error] 	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
[error] 	at sbt.internal.inc.AnalyzingCompiler.invoke(AnalyzingCompiler.scala:329)
[error] 	at sbt.internal.inc.AnalyzingCompiler.doc(AnalyzingCompiler.scala:175)
[error] 	at sbt.internal.inc.AnalyzingCompiler.doc(AnalyzingCompiler.scala:133)
[error] 	at sbt.Defaults$.docTaskSettings$$anonfun$4(Defaults.scala:2325)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:79)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:79)
[error] 	at sbt.std.Transform$$anon$5.work(Transform.scala:79)
[error] 	at sbt.std.Transform$$anon$5.work(Transform.scala:79)
[error] 	at sbt.Execute.submit$$anonfun$1$$anonfun$1(Execute.scala:284)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:24)
[error] 	at sbt.Execute.work(Execute.scala:295)
[error] 	at sbt.Execute.submit$$anonfun$1(Execute.scala:284)
[error] 	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$2(ConcurrentRestrictions.scala:277)
[error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:75)
[error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:73)
[...]

PR #7878, opened 2024-11-12
[2.x] Use ScalaInstance from update

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7725

## Problem
doc task currently doesn't work when the subproject tries to use the Scala instance from metabuild.

## Solution
For now turn off the optimized ScalaInstance


adpi2 (approved):

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -3666,14 +3491,11 @@ object Classpaths {
       val extResolvers = externalResolvers.value
       val isScala3M123 = ScalaArtifacts.isScala3M123(version)
       val allToolDeps =
-        if (
-          scalaHome.value.isDefined || scalaModuleInfo.value.isEmpty || !managedScalaInstance.value
-        )
-          Nil
-        else if (!isScala3M123 || extResolvers.contains(Resolver.JCenterRepository)) {
```
adpi2, line 3648: Ok so we won't be able to use scaladoc for `3.0.0-M1`. You probably need to update the Scala version in `project-matrix/jvm-sandwich` scripted test.

eed3si9n:
https://github.com/sbt/sbt/actions/runs/11793915183

main/src/main/scala/sbt/Defaults.scala
eed3si9n: I think sandwich has some issue - https://github.com/sbt/sbt/pull/7881

PR #7879, opened 2024-11-12
[2.x] Refactor `Settings` and optimize indexing of aggregate keys

adpi2:
This PR contains a refactoring of  `Settings` and an optimization of the indexing of aggregate keys. Overall it speeds up the build even if doing less parallelism.

## Refactoring of Settings

`Settings0` used to be a `Map[Scope, AttributeMap`], and it is now a `Map[ScopedKey[x], x]`. This is better because we don't need to decompose all `ScopedKey[x]` into a `Scope` and an `AttributeKey[x]`, for recomposing it back later, which duplicates all `ScopedKey[x]`. It reduces the number of long-living `ScopedKey[x]` instances by 8%, and the total number of instances by 1.4%.

Also it improves the performance of `Settings0`, which was responsible of 2.95% of the total CPU time, and is now responsible of 0.41%.

## Optimization of indexing of aggregate keys

To build the index of all aggregate keys, we were computing the reverse aggregation of each key, before indexing them. And so each aggregate key was indexed many times, once for each aggregated project. It was parallelized to reduce the latency. In this PR, we compute the reverse aggregation of all the keys at once, removing all duplication. We cannot parallelize this process anymore but we don't need to, because it is a lot faster. It reduces the total CPU time by 12%. The impact for the user depends on its number of cores.

## Result

I am still loading `softwaremill/sttp` on my 12 core laptop.

To get the heap size and the number of instances I generate a heap dump after loading sbt once. To get the CPU time and the elapsed real time, I do `time sbt 'reload;reload;reload'`.

|         | Instances | Used heap size | CPU time   | Elapsed real time |
|---------|-----------|----------------|------------|-------------------|
| develop | 16,364 K  | 410 MB         | 158,972 ms | 35,485  ms          |
| This PR | 16,130 K  | 401 MB         | 139,792 ms | 34,347 ms           |
|         | 98.6 %    | 97.8 %         | 87.9%      | 96.7 %            |

### DIFF
--- a/main/src/main/scala/sbt/Extracted.scala
+++ b/main/src/main/scala/sbt/Extracted.scala
@@ -12,7 +12,6 @@ import sbt.internal.{ Load, BuildStructure, Act, Aggregation, SessionSettings }
 import Scope.GlobalScope
 import Def.{ ScopedKey, Setting }
 import sbt.internal.util.complete.Parser
-import sbt.internal.util.AttributeKey
 import sbt.util.Show
 import std.Transform.DummyTaskMap
 import sbt.EvaluateTask.extractedTaskConfig
@@ -34,21 +33,21 @@ final case class Extracted(
    * If the project axis is not explicitly specified, it is resolved to be the current project according to the extracted `session`.
    * Other axes are resolved to be `Zero` if they are not specified.
    */
-  def get[T](key: SettingKey[T]): T = getOrError(inCurrent(key.scope), key.key)
-  def get[T](key: TaskKey[T]): Task[T] = getOrError(inCurrent(key.scope), key.key)
+  def get[T](key: SettingKey[T]): T = getOrError(inCurrent(key.scopedKey))
+  def get[T](key: TaskKey[T]): Task[T] = getOrError(inCurrent(key.scopedKey))
 
   /**
    * Gets the value assigned to `key` in the computed settings map wrapped in Some.  If it does not exist, None is returned.
    * If the project axis is not explicitly specified, it is resolved to be the current project according to the extracted `session`.
    * Other axes are resolved to be `Zero` if they are not specified.
    */
-  def getOpt[T](key: SettingKey[T]): Option[T] = structure.data.get(inCurrent(key.scope), key.key)
+  def getOpt[T](key: SettingKey[T]): Option[T] = structure.data.get(inCurrent(key.scopedKey))
   def getOpt[T](key: TaskKey[T]): Option[Task[T]] =
-    structure.data.get(inCurrent(key.scope), key.key)
+    structure.data.get(inCurrent(key))
 
-  private def inCurrent(scope: Scope): Scope =
-    if scope.project == This then scope.rescope(currentRef)
-    else scope
+  private def inCurrent[T](key: ScopedKey[T]): ScopedKey[T] =
+    if key.scope.project == This then key.copy(scope = key.scope.rescope(currentRef))
+    else key
 
   /**
    * Runs the task specified by `key` and returns the transformed State and the resulting value of the task.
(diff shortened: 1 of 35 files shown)


mzuehlke:
Great work @adpi2 
while the number of instances and the heap size correspondences to and improves upon you last PR (https://github.com/sbt/sbt/pull/7866), the time is not related.

eed3si9n:
This is pretty cool.

eed3si9n:
1. Could we split this into two PRs if they are independent optimizations?
2. It would be great to move the PR description into commit message so we can capture this info in git.
3. Maybe we should also move some also into scaladoc
4. Are settings mostly internal implementation of sbt? Do we anticipate impact to plugin sources?

adpi2:
> 1. Could we split this into two PRs if they are independent optimizations?

Yes I can. But if I separate them, the elapsed real time is getting slightly worse. Because the new `Settings` does not perform well on the old `KeyIndex.aggregate`. For each task scope, we check the `aggregate` key which is often missing, or alone in its `Map1`. This is faster with `Map[Scope, AttributeMap]`.

> 2. It would be great to move the PR description into commit message so we can capture this info in git.
> 3. Maybe we should also move some also into scaladoc

Will do

> 4. Are settings mostly internal implementation of sbt? Do we anticipate impact to plugin sources?

I assumed it was mostly internal. I'll check that with Github search.

adpi2:
> 2. It would be great to move the PR description into commit message so we can capture this info in git.
> 3. Maybe we should also move some also into scaladoc

 Done
 
 > 4. Are settings mostly internal implementation of sbt? Do we anticipate impact to plugin sources?
 
Plugins use `Settings` to access their values in a command. For instance [here](https://github.com/scalacenter/scalafix/blob/8a4cedc3799b5b2795de69d9dc9fc30bb29fdedc/project/ScalafixBuild.scala#L207-L211) in Scalafix we have:
```scala
      val extracted = Project.extract(state)
      val v =
        (ThisBuild / version)
          .get(extracted.structure.data)
          .get
```

But this `get` method is still there, so I think it is going to be fine in most cases.

In some plugins `Settings[Scope]` is written explicitly as parameter, and we will need a `Compat` to convert it to `Def.Settings`:
- [scala-android/sbt-android](https://github.com/scala-android/sbt-android/blob/280b0dd29584d164bfc3aa4289c9907e03b38ae5/src/VariantSettings.scala#L174)
- [scalacenter/bloop](https://github.com/scalacenter/bloop/blob/c83d5c5356628bf62f29120d23c9a2ca0b0fa17d/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala#L510)
- [JetBrains/sbt-intellij](https://github.com/JetBrains/sbt-intellij/blob/b08c15f9d9471af269a3746747743dfeb2d20aff/sbt-project-model/src/main/scala/org/jetbrains/sbt/extractors/Play2Extractor.scala#L28)

main/src/main/scala/sbt/Extracted.scala
```diff
@@ -116,15 +115,15 @@ final case class Extracted(
   private def resolve[K <: Scoped.ScopingSetting[K] & Scoped](key: K): K =
     Scope.resolveScope(GlobalScope, currentRef.build, rootProject)(key.scope) / key
 
-  private def getOrError[T](scope: Scope, key: AttributeKey[?], value: Option[T])(implicit
+  private def getOrError[T](key: ScopedKey[?], value: Option[T])(implicit
       display: Show[ScopedKey[?]]
   ): T =
-    value getOrElse sys.error(display.show(ScopedKey(scope, key)) + " is undefined.")
+    value.getOrElse(sys.error(display.show(key) + " is undefined."))
```
eed3si9n, line 121: nit: While we are at it, could you check if it would be helpful to turning this into a pattern match, and inlining the method?

eed3si9n (approved):
Thanks!

main/src/main/scala/sbt/Extracted.scala
adpi2: I don't see anything about `Extracted` in the profiling. So probably it is not even called during loading.

But in general, yes it is good idea. I'll experiment with a `inline def` version of `getOrElse` which does pattern matching to see if it makes a difference.

PR #7880, opened 2024-11-12
[2.x] Reduce creation of `Setting` and `Initialize` 

adpi2:
Some minor changes to reduce the number of `Setting` and `Initialize`.

Those changes reduce the number of long-living instances by 1.0%. Notably the number of `Init.Apply` nodes drops from 723,725 to 683,867 (- 5.5%).


eed3si9n (approved):
Thanks!

PR #7882, opened 2024-11-13
[2.x] Remove instances of `Info` and `BasicAttributeMap`

adpi2:
`Task`, `Info` and `BasicAttributeMap` corresponds to 11% of all the living instances in the heap after loading.

The main idea of this PR is to remove the instances of `Info`, by merging `Info` in `Task`, and the instances of `BasicAttributeMap` by making it an opaque type. Additionally I explicited that `Task` uses reference equality (its not a case class anymore) and I removed `Structure.taskKey` which is not used anywhere, neither internally nor in any sbt 1.x plugin.

Result:

|         | Instances | Used heap size | Elapsed real time |
|---------|-----------|----------------|-------------------|
| develop (as of yesterday) | 16,364 K  | 410 MB         | 35,668 ms         |
| This PR | 15,012 K  | 380 MB         | 33,892 ms         |
|         | 91.7 %    | 92.7 %         | 95.0 %            |


eed3si9n (approved):
Thanks!

PR #7886, opened 2024-11-16
[1.x] Add double quote around thread name during trace

### LINKED ISSUE #7884: Chrome trace format broken
## steps

```
sbt --traces version
```

## problem

Output Chrome format trace is missing double quotes around thread name.
This regression was introduced recently in this commit: https://github.com/sbt/sbt/commit/0b34b6b4065e92e3d556703ec2960070406e20fa

## expectation

Thread name should be wrapped in double quotes, since this is a string and not a number anymore

Friendseeker:
Closes #7884


eed3si9n (approved):
Thanks!

PR #7889, opened 2024-11-16
[2.x] remove `implicit` params. change to `using`

xuwei-k:


eed3si9n:
Probably ok, since Scala 2.12 now accepts `using` as well?

eed3si9n (approved):

Issue #7405, opened 2022-10-21
Scala 3 sbt sandwich doesn't work


MateuszKubuszok:
I have a following repro:

```scala
// plugins.sbt
addSbtPlugin("com.eed3si9n" % "sbt-projectmatrix" % "0.9.0")
addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.11.0")
```

```scala
// build.sbt
val a = projectMatrix
  .in(file("a"))
  .customRow(
    scalaVersions = Seq("3.2.0"),
    axisValues = Seq(VirtualAxis.jvm),
    settings = Seq(
      //name := "a"
    )
  )
  .customRow(
    scalaVersions = Seq("3.2.0"),
    axisValues = Seq(VirtualAxis.js),
    settings = Seq(
      //name := "a"
    )
  )

val b = projectMatrix
  .in(file("b"))
  .customRow(
    scalaVersions = Seq("2.13.10"),
    axisValues = Seq(VirtualAxis.jvm),
    settings = Seq(
      //name := "b"
    )
  )
  .customRow(
    scalaVersions = Seq("2.13.10"),
    axisValues = Seq(VirtualAxis.js),
    settings = Seq(
      //name := "b"
    )
  )
  .dependsOn(a % Test)

val root = project.in(file("."))
  .aggregate(a.projectRefs ++ b.projectRefs: _*)
```

When I would run commands like `b/test` or `b/publishM2` I would expect the code to depend on `a` (with Scala 3) project when running test and not introduce any "phantom" dependencies to runtime scope. However, I get:

```
sbt:root> b/publishM2
[info] Wrote /Users/dev/Workspaces/GitHub/projectmatrix-repro/b/target/jvm-2.13/b_2.13-0.1.0-SNAPSHOT.pom
[info] Updating
https://repo1.maven.org/maven2/org/scala-lang/scala-library/2.13.10/scala-library-2.13.10.pom
  100.0% [##########] 1.6 KiB (1.7 KiB / s)
[info] Resolved  dependencies
[info] Updating
[info] Resolved  dependencies
[warn]
[warn] 	Note: Unresolved dependencies path:
[error] stack trace is suppressed; run last b / update for the full output
[error] (b / update) sbt.librarymanagement.ResolveException: Error downloading a:a_2.13:0.1.0-SNAPSHOT
[error]   Not found
[error]   Not found
[...]

PR #7907, opened 2024-11-23
[2.x] fix: Fix Scala 3.x - 2.12 sandwich for matrix

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7405

## Problem
Currently Scala 3.x - 2.13 sandwich doesn't seem to work for projectMatrix. This is due to isScala2Scala3Sandwich checking for `"3."` while projectMatrix uses scalaBinaryVersion, which is 3.

## Solution
This consolidates the implementation of isScala2Scala3Sandwich with the one in Defaults module, and check for both `"3"` and `"3."`

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -4116,35 +4116,34 @@ object Classpaths {
 
   private[sbt] def isScala213(sv: String) = sv.startsWith("2.13.")
 
-  private[sbt] def isScala2Scala3Sandwich(sbv1: String, sbv2: String): Boolean = {
-    def compare(a: String, b: String): Boolean =
-      a == "2.13" && (b.startsWith("0.") || b.startsWith("3"))
-    compare(sbv1, sbv2) || compare(sbv2, sbv1)
-  }
-
   def projectDependenciesTask: Initialize[Task[Seq[ModuleID]]] =
     Def.task {
       val sbv = scalaBinaryVersion.value
       val ref = thisProjectRef.value
       val data = settingsData.value
       val deps = buildDependencies.value
-      deps.classpath(ref) flatMap { dep =>
-        for {
-          depProjId <- (dep.project / projectID).get(data)
-          depSBV <- (dep.project / scalaBinaryVersion).get(data)
-          depCross <- (dep.project / crossVersion).get(data)
-        } yield {
-          depCross match {
-            case b: CrossVersion.Binary if isScala2Scala3Sandwich(sbv, depSBV) =>
+      deps
+        .classpath(ref)
+        .flatMap: dep =>
+          for
+            depProjId <- (dep.project / projectID).get(data)
+            depSBV <- (dep.project / scalaBinaryVersion).get(data)
+            depCross <- (dep.project / crossVersion).get(data)
+            depAuto <- (dep.project / autoScalaLibrary).get(data)
+          yield depCross match
+            case b: CrossVersion.Binary
+                if depAuto && VirtualAxis.isScala2Scala3Sandwich(sbv, depSBV) =>
               depProjId
                 .withCrossVersion(CrossVersion.constant(b.prefix + depSBV))
                 .withConfigurations(dep.configuration)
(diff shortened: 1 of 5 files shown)


main/src/main/scala/sbt/VirtualAxis.scala
```diff
@@ -66,11 +66,10 @@ object VirtualAxis {
     isScala2Scala3Sandwich(str(sbv1), str(sbv2))
   }
 
-  private[sbt] def isScala2Scala3Sandwich(sbv1: String, sbv2: String): Boolean = {
+  private[sbt] def isScala2Scala3Sandwich(sbv1: String, sbv2: String): Boolean =
     def compare(a: String, b: String): Boolean =
-      a == "2.13" && (b.startsWith("0.") || b.startsWith("3."))
+      a == "2.13" && (b == "3" || b.startsWith("3."))
```
eed3si9n, line 71: Main fix is this line.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -4116,35 +4116,34 @@ object Classpaths {
...
-          depCross match {
-            case b: CrossVersion.Binary if isScala2Scala3Sandwich(sbv, depSBV) =>
+      deps
+        .classpath(ref)
+        .flatMap: dep =>
+          for
+            depProjId <- (dep.project / projectID).get(data)
+            depSBV <- (dep.project / scalaBinaryVersion).get(data)
+            depCross <- (dep.project / crossVersion).get(data)
+            depAuto <- (dep.project / autoScalaLibrary).get(data)
```
eed3si9n, line 4132: Exclude `autoScalaLibrary := false` subprojects, since they could have `scalaVersion` set to 3.x but are Java projects.

main/src/main/scala/sbt/VirtualAxis.scala
Friendseeker: Can we just check `b.startsWith("3")` instead of `(b == "3" || b.startsWith("3."))`? Or would that mishandle some edge case?

eed3si9n: It will break when Scala 30.x comes out in the future.

Friendseeker:
https://github.com/sbt/sbt/blob/546feed6f4d3b15e5bbda4c3f0073797f6ac9915/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala#L49

Just want to confirm. We don't need to do anything there right?

eed3si9n:
Yea, as long as sv is `scalaVersion` it's ok. The problem was that projectMatrix passes in `scalaBinaryVersion`.

BillyAutrey (approved):

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -4116,35 +4116,34 @@ object Classpaths {
...
                 .withCrossVersion(CrossVersion.constant(b.prefix + depSBV))
                 .withConfigurations(dep.configuration)
                 .withExplicitArtifacts(Vector.empty)
+            case _ if depAuto && VirtualAxis.isScala2Scala3Sandwich(sbv, depSBV) =>
+              depProjId
+                .withCrossVersion(CrossVersion.constant(depSBV))
+                .withConfigurations(dep.configuration)
+                .withExplicitArtifacts(Vector.empty)
             case _ =>
               depProjId.withConfigurations(dep.configuration).withExplicitArtifacts(Vector.empty)
```
adpi2, line 4146: NIT: I think it would be clearer to write it like this:

```scala
val prefix = depCross match
  case b: CrossVersion.Binary => b.prefix
  case _ => ""
val crossVersion = 
  if depAuto && VirtualAxis.isScala2Scala3Sandwich(sbv, depSBV) then
    CrossVersion.constant(prefix + depSBV)
  else depProjId.crossVersion
depProjId
  .withCrossVersion(crossVersion)
  .withConfigurations(dep.configuration)
  .withExplicitArtifacts(Vector.empty)
```

PR #7910, opened 2024-11-25
[1.x] Add settings key to surface Consistent Analysis `Reproducible` flag

Friendseeker:
This PR adds a new `sbt` settings key `enableReproducibleCompileAnalysis` for Consistent Analysis's `reproducible` flag, added in https://github.com/sbt/zinc/pull/1479.

The semantic of `enableReproducibleCompileAnalysis` is as follows. If the underlying analysis format is not Consistent Analysis, the settings key does nothing. Otherwise,  its value is passed as `reproducible` flag for Consistent Analysis, for which `reproducible = true` ensures binary reproducible analysis output, at the expense of wiping timestamp information (and therefore less accurate zinc invalidation).

The default value of `enableReproducibleCompileAnalysis` is set to `false`, following recommendation from [ephemerist](https://github.com/ephemerist) in https://github.com/sbt/zinc/pull/1496#issuecomment-2477977973.


Friendseeker:
```
[error] /home/runner/work/sbt/sbt/main/src/main/scala/sbt/internal/AnalysisUtil.scala:31:28: overloaded method value staticCachedStore with alternatives:
[error]   (analysisFile: java.nio.file.Path,useTextAnalysis: Boolean,useConsistent: Boolean,mappers: xsbti.compile.analysis.ReadWriteMappers,sort: Boolean,parallelism: Int)xsbti.compile.AnalysisStore <and>
[error]   (analysisFile: java.nio.file.Path,useTextAnalysis: Boolean)xsbti.compile.AnalysisStore
[error]  cannot be applied to (analysisFile: java.nio.file.Path, useTextAnalysis: Boolean, useConsistent: Boolean, mappers: xsbti.compile.analysis.ReadWriteMappers, reproducible: Boolean, parallelism: Int)
[error]     MixedAnalyzingCompiler.staticCachedStore(
```

Guess we need to make a new Zinc release?

Friendseeker:
Now after a few weeks I am having some second thoughts. Should `sbt` just be opinionated and not surface the key to user?

PR #7911, opened 2024-11-25
[2.x] Refactor slash syntax to mostly use methods

eed3si9n:
## Problem
Slash syntax is currently implemented via a series of implicit
converters (Conversion), which is not nice, partly because
the behavior is difficult to follow.

## Solution
This removes all the implicit converters and moves the slashes
into:
1. / methods under Reference for subproject scoping.
2. / methods under ScopeAxis with Reference type constraint
   for the initial Zero scoping.
3. Return RefThenConfig structure for intermediate config scoping.
4. / method under `Scoped`, which is base trait for the keys
   to implement task scoping e.g. `compile / scalacOptions`.
5. Extension methods for ConfigKey.
6. Extension methods for AttributeKey.

### DIFF
--- a/main-settings/src/main/scala/sbt/Scope.scala
+++ b/main-settings/src/main/scala/sbt/Scope.scala
@@ -430,4 +430,23 @@ object Scope:
         t <- withZeroAxis(scope.task)
         e <- withZeroAxis(scope.extra)
       } yield Scope(Zero, c, t, e)
+
+  /**
+   * Temporary data structure to capture first two axis using slash syntax.
+   * In theory, we might be able to express this as type parameters of Scope,
+   * like Scope[Select[ThisBuild.type], Select[ConfigKey], This, This] but then
+   * scope becomes more complicated to deal with.
+   */
+  opaque type RefThenConfig = (ScopeAxis[Reference], ScopeAxis[ConfigKey])
+
+  object RefThenConfig:
+    def apply(project: ScopeAxis[Reference], config: ScopeAxis[ConfigKey]): RefThenConfig =
+      (project, config)
+    def apply(project: ScopeAxis[Reference], config: ConfigKey): RefThenConfig =
+      (project, Select(config))
+
+    extension (in: RefThenConfig)
+      def project: ScopeAxis[Reference] = in._1
+      def config: ScopeAxis[ConfigKey] = in._2
+  end RefThenConfig
 end Scope
--- a/main-settings/src/main/scala/sbt/Scope.scala
+++ b/main-settings/src/main/scala/sbt/Scope.scala
@@ -32,6 +32,9 @@ final case class Scope private (
   def rescope(config: ConfigKey): Scope = copy(config = Select(config))
   def rescope(task: AttributeKey[?]): Scope = copy(task = Select(task))
 
+  final def /[K](key: Scoped.ScopingSetting[K]): K = scope(key)
+  def scope[K](key: Scoped.ScopingSetting[K]): K = key.rescope(this)
+
   def copy(
       project: ScopeAxis[Reference] = this.project,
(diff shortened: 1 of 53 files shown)


main-settings/src/main/scala/sbt/Scope.scala
```diff
@@ -31,6 +32,9 @@ final case class Scope private (
   def rescope(config: ConfigKey): Scope = copy(config = Select(config))
   def rescope(task: AttributeKey[?]): Scope = copy(task = Select(task))
 
+  final def /[K](key: Scoped.ScopingSetting[K]): K = scope(key)
+  def scope[K](key: Scoped.ScopingSetting[K]): K = key.rescope(this)
+
```
adpi2, line 37: Why do we have 2 versions of this method? Should we keep `/` only?

main-settings/src/main/scala/sbt/ScopeAxis.scala
```diff
@@ -8,46 +8,79 @@
...
+  /** Temporary data structure to capture first two axis using slash syntax. */
+  class RefThenConfig(val scope: Scope):
+    override def toString(): String = scope.toString()
+    inline def /[K](key: Scoped.ScopingSetting[K]): K = scope / key
+
+    inline def /(task: AttributeKey[?]): Scope = scope.copy(task = Select(task))
+
+    /** This is for handling `Zero / Zero / Zero / name`. */
+    inline def /(taskAxis: ScopeAxis[AttributeKey[?]]): Scope = scope.copy(task = taskAxis)
+  end RefThenConfig
```
adpi2: NIT: I would rather define this in `Scope`

NIT: Should we make it explicit that it only contains ref and config: `class RefThenConfig(ref: ScopeAxis[Reference], config: ScopeAxis[Configuration])`?

adpi2 (commented):
Overall it looks good to me. Getting away from implicit conversion is a good thing IMO.

Some tests are failing:
```
[error] (scripted) Failed tests:
[error] 	actions/external-doc
[error] 	actions/previous
```
And
```
[error] x dependency-management/exclude-dependencies 
```

main-settings/src/main/scala/sbt/Scope.scala
eed3si9n: For any symbolic operators, I think it would be a good practice to have English method name counterpart that we can explain exactly what they do.

main-settings/src/main/scala/sbt/ScopeAxis.scala
```diff
@@ -8,46 +8,79 @@
...
+
+  /** Temporary data structure to capture first two axis using slash syntax. */
+  class RefThenConfig(val scope: Scope):
+    override def toString(): String = scope.toString()
+    def /[K](key: Scoped.ScopingSetting[K]): K = scope / key
+
+    def /(task: AttributeKey[?]): Scope = scope.copy(task = Select(task))
+
+    /** This is for handling `Zero / Zero / Zero / name`. */
+    def /(taskAxis: ScopeAxis[AttributeKey[?]]): Scope = scope.copy(task = taskAxis)
```
adpi2: Since `/` is not inline anymore, maybe `RefThenConfig` should be an opaque type or a value class?

eed3si9n: Addressed in eb7d43c363798e802db1d135d3f39a8e01965166

main-settings/src/main/scala/sbt/ScopeAxis.scala
eed3si9n: Addressed in eb7d43c363798e802db1d135d3f39a8e01965166 by moving it into Scope and making it a pair of axes.

adpi2 (approved):
LGTM

PR #7913, opened 2024-11-26
[1.x] Implement `jvmBuildTarget` for `workspace/buildTargets`

### LINKED ISSUE #7510: improvement: [BSP] Add `jvmBuildTarget` info to `workspace/buildTargets` response
According to BSP [ScalaBuildTarget](https://build-server-protocol.github.io/docs/extensions/scala#scalabuildtarget) has an optional field carrying information about [JvmBuildTarget](https://build-server-protocol.github.io/docs/extensions/jvm#jvmbuildtarget). Currently in `sbt`s response `jvmBuildTarget` field is empty.

This information would be useful for Metals to make sure that e.g. completions are  suggested for the correct JDK version, especially when `sbt` server is started manually by the user.

Friendseeker:
Closes https://github.com/sbt/sbt/issues/7510

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
@@ -34,6 +34,7 @@ import sjsonnew.support.scalajson.unsafe.{ CompactPrinter, Converter, Parser => 
 import xsbti.CompileFailed
 
 import java.io.File
+import java.nio.file.Paths
 import java.util.concurrent.atomic.AtomicBoolean
 import scala.collection.mutable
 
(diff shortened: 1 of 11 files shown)


main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -614,12 +615,19 @@ object BuildServerProtocol {
     val thisProjectRef = Keys.thisProjectRef.value
     val thisConfig = Keys.configuration.value
     val scalaJars = Keys.scalaInstance.value.allJars.map(_.toURI.toString)
+    val jvmBuildTarget = javaHome.value match {
+      // TODO: populate Java version for forked java
+      case Some(forkedJava) => JvmBuildTarget(Some(forkedJava.toURI), None)
+      case None =>
+        JvmBuildTarget(Paths.get(sys.props("java.home")).toUri, sys.props("java.class.version"))
```
adpi2: Could it be that `java.home` or `java.class.version` are not defined?

Friendseeker: Thanks! Just refactored the logic to make it more robust

adpi2 (commented):
Looks good

Could you add a test in [BuildServerTest](https://github.com/sbt/sbt/blob/1.10.x/server-test/src/test/scala/testpkg/BuildServerTest.scala)?

adpi2 (approved):
Thanks!

PR #7930, opened 2024-12-06
add lmCoursierShadedPublishing to allProjects

jtjeferreira:
I tried to use a locally published version of sbt, but one the dependencies was missing. 

I think this was happening because `lmCoursierShadedPublishing` is not being aggregated in the root and is not being published locally when I run `publishLocal`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1216,6 +1216,7 @@ def allProjects =
     lmCoursierDefinitions,
     lmCoursier,
     lmCoursierShaded,
+    lmCoursierShadedPublishing,
   ) ++ lowerUtilProjects
 
 // These need to be cross published to 2.12 and 2.13 for Zinc


build.sbt
```diff
@@ -1216,6 +1216,7 @@ def allProjects =
     lmCoursierDefinitions,
     lmCoursier,
     lmCoursierShaded,
+    lmCoursierShadedPublishing,
```
eed3si9n, line 1219: Nice catch. If that one is included, we probably won't need `lmCoursierShaded` above that one.

jtjeferreira: >  If that one is included, we probably won't need `lmCoursierShaded` above that one.

but this variable is called `allProjects`...

eed3si9n: ok, fine. I'll try to adjust the aggregate later then.

eed3si9n (approved):

Issue #7931, opened 2024-12-08
scripted test with exists


eed3si9n:
## steps

Port plugins to sbt 2.x.

https://github.com/sbt/sbt-assembly/pull/533/files#diff-55a4d19d93a5db655a52481495f149fd5edb3bf97054607a46e31da77e29e309

## problem

```diff
  > check

- $ exists target/classes/somepackage/SomeClass.class
- $ absent target/classes/shaded/SomeClass.class

  > unzip

- $ exists target/unzipped/shaded/SomeClass.class
- $ absent target/unzipped/somepackage/SomeClass.class
```

This is because it's currently too annoying to test this that works for both sbt 1.x and sbt 2.x

```
> check

$ exists target/out/jvm/scala-2.12.18/root/classes/somepackage/SomeClass.class
$ absent target/out/jvm/scala-2.12.18/root/classes/shaded/SomeClass.class

> unzip

$ exists target/out/jvm/scala-2.12.18/root/unzipped/shaded/SomeClass.class
$ absent target/out/jvm/scala-2.12.18/root/unzipped/somepackage/SomeClass.class
```

## expectation

`||` should be interpreted to be an or expression

```
$ exists target/unzipped/shaded/SomeClass.class || target/out/jvm/scala-2.12.18/root/classes/somepackage/SomeClass.class
```

`*` and `**` should be interpreted as glob expression:

```
$ exists target/**/classes/somepackage/SomeClass.class
```

mkurz:
- Fixed by #7932

PR #7932, opened 2024-12-09
[2.x] feat: Support glob expressions in scripted

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7931

## Problem
It's currently not easy to write a scripted test that works on both sbt 1.x and 2.x when you want to write exists test under target.

## Solution
Since we can only use the file system (and not evaluate Scala version etc)
1. this implements glob expression support in `exists`, `absent`, and `delete`.
2. this also introduces `||` operator that would mean a or b.

### DIFF
--- a/internal/util-scripted/src/main/scala/sbt/internal/scripted/FileCommands.scala
+++ b/internal/util-scripted/src/main/scala/sbt/internal/scripted/FileCommands.scala
@@ -11,11 +11,15 @@ package internal
 package scripted
 
 import java.io.File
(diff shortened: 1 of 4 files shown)


adpi2 (approved):
That looks great! Thank you @eed3si9n

jtjeferreira (commented):
I tested this successfully by publishing both sbt 1.x and 2.x locally and using `$ exists target/**/universal/simple-test-0.1.0.zip` to handle both `target/universal/simple-test-0.1.0.zip` and `target/out/jvm/scala-3.5.2/simple-test/universal/simple-test-0.1.0.zip`

PR #7937, opened 2024-12-09
[2.x] WIP: mappings key as (File,String)

jtjeferreira:
One of the troubles of upgrading sbt native packager plugin to sbt 2.x was the change to the `mappings` key because code like
```scala
mappings ++= Seq(
  file("LICENSE") -> "LICENSE",
)
```

no longer works and needs to be translated to use the `fileConverter`. 

This PR reverts `mappings` key to use `(File,String)`


eed3si9n:
@jtjeferreira I have a fix PR for the failing test - https://github.com/sbt/sbt/pull/7948

adpi2:
Is this not going to break the hash key of the `package` cached task?

eed3si9n:
See https://github.com/sbt/sbt/pull/7949. I think mapping should be `Seq[(HashedVirtualFileRef, String)]`

Issue #7873, opened 2024-11-06
[2.x] ConcurrentModificationException in SbtParser


xuwei-k:
## steps

sbt 2.0.0-M2

flaky some sbt plugin scripted test

## problem

https://github.com/wartremover/wartremover-contrib/actions/runs/11697441192/job/32576176336#step:5:812

```
Error:  Exception in thread "sbt-parser-init-thread" java.lang.ExceptionInInitializerError
Error:  	at sbt.internal.parser.SbtParserInit$$anon$1.run(SbtParser.scala:179)
Error:  Caused by: java.util.ConcurrentModificationException
Error:  	at java.util.Hashtable$Enumerator.next(Hashtable.java:1408)
Error:  	at scala.collection.convert.JavaCollectionWrappers$JPropertiesWrapper$$anon$6.next(JavaCollectionWrappers.scala:622)
Error:  	at scala.collection.convert.JavaCollectionWrappers$JPropertiesWrapper$$anon$6.next(JavaCollectionWrappers.scala:618)
Error:  	at scala.collection.IterableOnceOps.find(IterableOnce.scala:634)
Error:  	at scala.collection.IterableOnceOps.find$(IterableOnce.scala:631)
Error:  	at scala.collection.AbstractIterator.find(Iterator.scala:1300)
Error:  	at dotty.tools.dotc.config.PathResolver$Environment$.searchForBootClasspath(PathResolver.scala:40)
Error:  	at dotty.tools.dotc.config.PathResolver$Environment$.javaBootClassPath(PathResolver.scala:49)
Error:  	at dotty.tools.dotc.config.PathResolver$Defaults$.javaBootClassPath(PathResolver.scala:75)
Error:  	at dotty.tools.dotc.config.CommonScalaSettings.$init$(ScalaSettings.scala:97)
Error:  	at dotty.tools.dotc.config.ScalaSettings.<init>(ScalaSettings.scala:15)
Error:  	at dotty.tools.dotc.core.Contexts$ContextBase.<init>(Contexts.scala:857)
Error:  	at dotty.tools.dotc.Driver.initCtx(Driver.scala:61)
Error:  	at sbt.internal.parser.SbtParser$ParseDriver.<init>(SbtParser.scala:138)
Error:  	at sbt.internal.parser.SbtParser$.<clinit>(SbtParser.scala:135)
```

## notes

Maybe not thread safe 🤔 

[...]

PR #7938, opened 2024-12-10
[2.x] fix: concurrency control around Scala 3 compiler

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7873

## Problem
Early initialization thread throws ConcurrentModificationException.

## Solution
This wraps the initialization with Retry(...).

### DIFF
--- a/buildfile/src/main/scala/sbt/internal/parser/SbtParser.scala
+++ b/buildfile/src/main/scala/sbt/internal/parser/SbtParser.scala
@@ -10,6 +10,7 @@ package internal
 package parser
 
 import sbt.internal.util.{ LineRange, MessageOnlyException }
+import sbt.internal.io.Retry
 import java.io.File
 import java.nio.charset.StandardCharsets
 import java.util.concurrent.ConcurrentHashMap
@@ -131,10 +132,10 @@ private[sbt] object SbtParser:
 
   private[sbt] var scalacGlobalInitReporter: Option[ConsoleReporter] = None
 
-  private[sbt] lazy val globalReporter = UniqueParserReporter()
-  private lazy val defaultGlobalForParser = ParseDriver()
-  private[sbt] def getGlobalForParser: ParseDriver = synchronized:
-    defaultGlobalForParser
+  private[sbt] val globalReporter = UniqueParserReporter()
+  // Retry since Scala 3 compiler initialization can fail due to sys.props change
+  private[sbt] val defaultGlobalForParser: ParseDriver =
+    Retry(ParseDriver())
   private[sbt] final class ParseDriver extends Driver:
     override protected val sourcesRequired: Boolean = false
     val compileCtx0 = initCtx.fresh
@@ -181,7 +182,7 @@ private class SbtParserInit:
   val t = new Thread("sbt-parser-init-thread"):
     setDaemon(true)
     override def run(): Unit =
-      val _ = SbtParser.getGlobalForParser
+      val _ = SbtParser.defaultGlobalForParser
   t.start()
 end SbtParserInit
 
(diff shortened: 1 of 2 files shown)


buildfile/src/main/scala/sbt/internal/parser/SbtParser.scala
```diff
@@ -131,8 +131,10 @@ private[sbt] object SbtParser:
 
   private[sbt] var scalacGlobalInitReporter: Option[ConsoleReporter] = None
 
-  private[sbt] val globalReporter = UniqueParserReporter()
-  private[sbt] val defaultGlobalForParser = ParseDriver()
+  private[sbt] lazy val globalReporter = UniqueParserReporter()
+  private lazy val defaultGlobalForParser = ParseDriver()
+  private[sbt] def getGlobalForParser: ParseDriver = synchronized:
+    defaultGlobalForParser
```
adpi2: Object initialization is already thread-safe, so I don’t think these changes are actually fixing anything. Or am I missing something?

adpi2 (changes requested):
I don't think this fixes the concurrency issue.

Looking at the stack trace, it seems that dotty is iterating the `System.properties` while another thread modifies them. Could it be that sbt, or a plugin modifies the `System.properties`?

eed3si9n:
https://eed3si9n.com/december-adventure-2024/ has commentary on the whereabout of the thread.

buildfile/src/main/scala/sbt/internal/parser/SbtParser.scala
eed3si9n: defaultGlobalForParser is accessed by three places, including a thread we start as a background JIT warmup. I'm wrapping the access to synchronized so we don't give access to this variable concurrently.

adpi2: Its initialization and access are already thread safe.

Really I don't think the problem is here. It's rather that the `ParserDriver.<init>` which is called only once iterate over the `System.properties` while another thread is mutating those properties.

jtjeferreira: `java.util.Properties` is [supposedly](https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html) thread safe...
>  This class is thread-safe: multiple threads can share a single Properties object without the need for external synchronization

eed3si9n: We do 

sys.props.put("jna.nosys", "true")

in Main.scala and there might be more by users potentially. so maybe it's not safe to change and read at the same time?

jtjeferreira: Also [dotty.tools.dotc.config.WrappedProperties#systemProperties](https://github.com/scala/scala3/blob/1775d0be938fe78126600502f3c3c2e6d0429f9e/compiler/src/dotty/tools/dotc/config/WrappedProperties.scala#L24-L27) uses `System.getProperties.asScala.iterator` and maybe it should just use `System.getProperties.asScala`

Also `dotty.tools.dotc.config.PathResolver.Environment.javaBootClassPath` is `def javaBootClassPath: String = propOrElse("sun.boot.class.path", searchForBootClasspath)` and the 2nd argument is always evaluated.

But these are bugs for the scala repo...

jtjeferreira: And after some archeology: https://github.com/scala/scala/commit/f6859f28bb49193fde83e6020a6a89ce926a91e8

I will try to backport that commit to scala 3

eed3si9n: Added fdb7ca23841cf4db91545a206509542e12e65862 to wrap in `Retry(...)` instead.

adpi2 (approved):
LGTM

buildfile/src/main/scala/sbt/internal/parser/SbtParser.scala
eed3si9n: Thanks both for the review / detective work!

buildfile/src/main/scala/sbt/internal/parser/SbtParser.scala
```diff
@@ -132,7 +133,9 @@ private[sbt] object SbtParser:
   private[sbt] var scalacGlobalInitReporter: Option[ConsoleReporter] = None
 
   private[sbt] val globalReporter = UniqueParserReporter()
-  private[sbt] val defaultGlobalForParser = ParseDriver()
+  // Retry since Scala 3 compiler initialization can fail due to sys.props change
+  private[sbt] val defaultGlobalForParser: ParseDriver =
+    Retry(ParseDriver())
```
jtjeferreira, line 138: this Retry did not work because it just retries `IOException`...

PR #7939, opened 2024-12-10
Add Mapper that returns VirtualFile based mappings 

jtjeferreira:
Moves Mapper methods from sbt/io to sbt/sbt so they can return `VirtualFile` based mappings (ie `(VirtualFile, String)` )

Although this change improves the verbosity of "user-code" when defining mappings:
```scala
Compile / packageBin / mappings ++= {
  Path.directory(file("test")).map { case (f,s) => converter.toVirtualFile(f.toPath) -> s }
}
```

becomes

```scala
Compile / packageBin / mappings ++= {
  Mapper.directory(file("test"))(using converter)
}
```

it is still very verbose compared to sbt 1.x:

```scala
Compile / packageBin / mappings ++= Mapper.directory(file("test"))
```

Also this PR does not improve other simple scenarios that were very succinct in sbt 1.x:

```scala
Compile / packageBin / mappings += file("test") -> "test"
```

### DIFF
--- a/main-actions/src/main/scala/sbt/Mapper.scala
+++ b/main-actions/src/main/scala/sbt/Mapper.scala
@@ -2,6 +2,7 @@ package sbt
 
 import sbt.io.{ AllPassFilter, FileFilter, PathFinder }
 import sbt.io.Path.*
+import xsbti.{ FileConverter, VirtualFile }
 
 import java.io.File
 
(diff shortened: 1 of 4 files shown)


eed3si9n:
closing in favor of https://github.com/sbt/sbt/pull/7949.

PR #7948, opened 2024-12-15
[2.x] fix: Use JDK path, not JRE path

eed3si9n:
## Problem
There are a few places where javaHome or java path is set, using java.home system property. The problem is that it points to JRE, not JDK, so it would break on Java compilation etc.

## Solution
If the path ends with jre, go up one directory.

### DIFF
--- a/server-test/src/test/scala/testpkg/BuildServerTest.scala
+++ b/server-test/src/test/scala/testpkg/BuildServerTest.scala
@@ -234,7 +234,6 @@ class BuildServerTest extends AbstractServerTest {
     val buildTarget = buildTargetUri("javaProj", "Compile")
 
     compile(buildTarget)
-
     assertMessage(
       "build/publishDiagnostics",
       "Hello.java",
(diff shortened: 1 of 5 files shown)


jtjeferreira:
I need to have a closer look, but this assumes the JRE is "contained" in the JDK. Is that always true? Maybe this is just a setup problem and a better error needs to be returned stating that javaHome doest not point to a JDK/javac?

PS: how is this currently handled when SBT needs to compile mixed scala and java code and it needs to call javac?

eed3si9n:
JAVA_HOME environment variable and PATH (we can sort of think of it as an env variable) are typically both correct, and point at JDK. when some piece of code uses the System Property `java.home` that's when we get into trouble because it points to a "JRE" version by spec. I am assuming that anyone running sbt uses JDK, and that JRE is a synthetic one within a JDK.

server-test/src/test/scala/testpkg/BuildServerTest.scala
```diff
@@ -257,16 +256,17 @@ class BuildServerTest extends AbstractServerTest {
...
     IO.write(
       otherBuildFile,
       """
+        |def jdk: File = sbt.internal.Util.javaHome.toFile()
         |lazy val javaProj = project
         |  .in(file("java-proj"))
         |  .settings(
         |    javacOptions += "-Xlint:all",
-        |    javaHome := Some(file(System.getProperty("java.home")))
+        |    javaHome := Some(jdk)
```
eed3si9n, line 269: Not sure how this has been working until João's File PR exposed this bug.

eed3si9n:
@jtjeferreira See https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html

> "java.home" | Installation directory for Java Runtime Environment (JRE)

Friendseeker:
Java never ceases to bring surprise.

Usually JDK and JRE installation directory are the same, but I guess there's always exception.

PR #7952, opened 2024-12-17
[1.x] Bump JLine 2 to `9a88bc4`

Friendseeker:
Contains https://github.com/sbt/jline2/commit/9a88bc413e2b34a4580c001c654d1a7f4f65bf18

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -84,16 +84,13 @@ object Dependencies {
   val sjsonNewScalaJson = sjsonNew("sjson-new-scalajson")
   val sjsonNewMurmurhash = sjsonNew("sjson-new-murmurhash")
 
-  // JLine 3 version must be coordinated together with JAnsi version
-  // and the JLine 2 fork version, which uses the same JAnsi
-  val jline = "org.scala-sbt.jline" % "jline" % "2.14.7-sbt-9c3b6aca11c57e339441442bbf58e550cdfecb79"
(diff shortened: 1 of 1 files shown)


SethTisue:
hmm, are we sure that removing jansi is a good idea in the JLine 2 context?

just asking; I don't remember why JLine 2 is even still present in this repo

SethTisue:
I see that JLine 2 is still present even on the `develop` branch, is that something we could or should clean up?

Friendseeker:
> hmm, are we sure this is a good idea in a JLine 2 context?
> 
> just asking; I'm only dimly aware of why JLine 2 is even still present in this repo

I honestly have no clue too... At this point only Eugene and perhaps Lukas / som-snytt knows. Only thing I know is that some legacy Scala 2.12 `sbt console` support relies on JLine 2...

Friendseeker:
> I see that JLine 2 is still present even on the `develop` branch, is that something we could or should clean up?

I mean thinking positively we all get to practice working with legacy code + unmaintained legacy dep!

SethTisue:
oh, Scala 2.12, sigh, that's plausible. still, I wonder if we should try ripping it out to verify if it's still needed

eed3si9n:
My guess is we'd need to support layered ClassLoader unless we drop the support of some Scala versions, which I don't think is worth it.

eed3si9n (approved):
Thanks!

SethTisue:
> unless we drop the support of some Scala versions, which I don't think is worth it.

it did cross my mind to try to convince you to drop support for building 2.12 projects 😈  but I'm not sure I could even convince myself, much as I would like to see 2.12 die

PR #7964, opened 2024-12-20
update copyrights to 'Lightbend, Inc. dba Akka'

Friendseeker:
Ports https://github.com/scala/scala/pull/10953

@SethTisue

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -60,7 +60,7 @@ def commonBaseSettings: Seq[Setting[_]] = Def.settings(
     HeaderLicense.Custom(
       """|sbt
        |Copyright 2023, Scala center
-       |Copyright 2011 - 2022, Lightbend, Inc.
+       |Copyright 2011 - 2022, Lightbend, Inc. dba Akka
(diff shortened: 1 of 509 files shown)


build.sbt
```diff
@@ -60,7 +60,7 @@ def commonBaseSettings: Seq[Setting[_]] = Def.settings(
     HeaderLicense.Custom(
       """|sbt
        |Copyright 2023, Scala center
-       |Copyright 2011 - 2022, Lightbend, Inc.
+       |Copyright 2011 - 2022, Lightbend, Inc. dba Akka
```
eed3si9n, line 63: Since during 2011 - 2022, Lightbend wasn't doing business as (dba) Akka, I don't think we need to add this into the header unless we're asked?

Friendseeker: Ah I forgot about that! I guess we shall close the PR for now, and if Seth deems it necessary we can reopen the PR

SethTisue:
I agree there is no need to change this.

PR #7979, opened 2024-12-29
[2.x] fix: Retry create directory

eed3si9n:
## Problem
Files.createDirectories can fail.

## Solution
Use IO version, which retries.


adpi2 (approved):
LGTM

PR #7983, opened 2024-12-31
[1.x] Avoid upstream compilation when calling `previousCompile`

### LINKED ISSUE #7982: [1.x] `previousCompile` triggers upstream `compile`
## steps

- Checkout https://github.com/Friendseeker/sbt-7973/tree/previousCompile-trigger-compile
- Call `B / previousCompile`

## problem

A compilation is triggered
```
[IJ]B / previousCompile
[debug] Other repositories:
[debug] Default repositories:
[debug] Using inline dependencies specified in Scala.
[debug] Copy resource mappings:
[debug]
[debug] [zinc] IncrementalCompile -----------
[debug] IncrementalCompile.incrementalCompile
[debug] previous = Stamps for: 3 products, 1 sources, 2 libraries
[debug] current source = Set(${BASE}/A/src/main/scala/A.scala)
```

## expectation

`previousCompile` should merely return the result of previous compilation. It should never trigger compilation of upstream project.

## note

Here's the trace

- previousCompile
- compileIncSetup
- dependencyPicklePath
- dependencyVirtualClasspath
- dependencyClasspath
- (concat)
- internalDependencyClasspath
- (internalDependencyClasspathTask)
- (internalDependenciesImplTask)
- exportedProducts
- (trackedExportedProducts)
- (trackedExportedProductsImplTask)
- (trackedNonJarProductsImplTask)
- compile

Basically `compileIncSetup` gathers a list of upstream analysis, which I guess do... kinda make sense. The issue is probably that `previousCompile` should not call `compileIncSetup` in the first place.

Friendseeker:
### Issue

`previousCompile` triggers upstream project compilation. So when user do `show B / previousCompile`, without the user knowing, upstream projects of `B` are compiled.

c.c. #7982

### Fix

Avoid usage of `compileIncSetup` in `previousCompile`. `compileIncSetup` task triggers upstream compilation to gather upstream analysis.

Fix #7982

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2580,6 +2580,7 @@ object Defaults extends BuildCommon {
   private[sbt] def jnone[A]: Optional[A] = none[A].toOptional
   def compileAnalysisSettings: Seq[Setting[_]] = Seq(
     previousCompile := {
+      // Avoid compileIncSetup since it would trigger upstream compilation
       val analysisFile = compileAnalysisFile.value
       val store = AnalysisUtil.staticCachedStore(
         analysisFile = analysisFile.toPath,
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2580,9 +2580,9 @@ object Defaults extends BuildCommon {
   private[sbt] def jnone[A]: Optional[A] = none[A].toOptional
   def compileAnalysisSettings: Seq[Setting[_]] = Seq(
     previousCompile := {
-      val setup = compileIncSetup.value
+      val analysisFile = compileAnalysisFile.value
       val store = AnalysisUtil.staticCachedStore(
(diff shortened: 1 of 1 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2580,9 +2580,9 @@ object Defaults extends BuildCommon {
   private[sbt] def jnone[A]: Optional[A] = none[A].toOptional
   def compileAnalysisSettings: Seq[Setting[_]] = Seq(
     previousCompile := {
-      val setup = compileIncSetup.value
+      val analysisFile = compileAnalysisFile.value
```
eed3si9n: Nice. Could you add a comment saying avoid `compileIncSetup` since it would trigger upstream compilation, plz?

eed3si9n (approved):

main/src/main/scala/sbt/Defaults.scala
Friendseeker: Done!

Issue #7963, opened 2024-12-20
[2.x] Non-blocking run hangs


eed3si9n:
Ref https://github.com/sbt/sbt/pull/7688

## steps

Run `run` twice:

```
sbt:bar> run
[info] running example.main
Hello
sbt:bar> run
[info] running example.main
Hello

^C
```

## problem

`run` hangs with or without sbtn.

## expectation

Don't hang.

PR #8006, opened 2025-01-12
[2.x] fix: Fixes run getting stuck or not showing up on sbtn

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7963

## Problem
1. When the run ends too quickly (e.g. "hello world") it gets stuck.
2. Currently sbtn doesn't correctly display the stdout during `run` because by the time it starts the terminal has been reset.

## Solution
1. This doesn't pause the channel if the job is already done.
2. Don't reset the terminal in-between the tasks so non-blocking run can use the terminal.


adpi2 (approved):
LGTM

PR #8012, opened 2025-01-17
Add setting to allow demoting the SIP-51 build failure

lrytz:
Add a `allowUnsafeScalaLibUpgrade` setting (default is `false`) to demote the SIP-51 build failure to a warning.

Follow-up for https://github.com/sbt/sbt/pull/7480.

If the scalaVersion is 2.13.12 but some dependency pulls in scala-library 2.13.13, the compiler will stay at 2.13.12, but the dependency classpath will contain scala-library 2.13.13.

This usually works, the compiler can run fine with a newer scala-library on its dependency classpath.
Macro expansion may fail, if the macro uses some library class / method that doesn't exist in the old version. The macro itself is loaded from the dependency classpath into the class loader running the compiler, where the older Scala library is on the runtime classpath.
Using the Scala REPL in sbt may also fail in a similar fashion.

Example how the warning looks:

```
[warn] Expected `b/scalaVersion` to be 2.13.13 or later, but found 2.13.12.
[warn] To support backwards-only binary compatibility (SIP-51), the Scala 2.13 compiler
[warn] should not be older than scala-library on the dependency classpath.
[warn]
[warn] Note that the dependency classpath and the runtime classpath of your project
[warn] contain the newer scala-library 2.13.13, even if the scalaVersion is 2.13.12.
[warn] Compilation (macro expansion) or using the Scala REPL in sbt may fail with a LinkageError.
[warn]
[warn] See `b/evicted` to know why scala-library 2.13.13 is getting pulled in.
```

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -180,6 +180,7 @@ object Defaults extends BuildCommon {
       apiMappings := Map.empty,
       autoScalaLibrary :== true,
       managedScalaInstance :== true,
+      allowUnsafeScalaLibUpgrade :== false,
       classpathEntryDefinesClass := { (file: File) =>
         sys.error("use classpathEntryDefinesClassVF instead")
       },
@@ -1178,6 +1179,7 @@ object Defaults extends BuildCommon {
   def scalaInstanceFromUpdate: Initialize[Task[ScalaInstance]] = Def.task {
     val sv = scalaVersion.value
     val fullReport = update.value
+    val s = streams.value
 
     // For Scala 3, update scala-library.jar in `scala-tool` and `scala-doc-tool` in case a newer version
     // is present in the `compile` configuration. This is needed once forwards binary compatibility is dropped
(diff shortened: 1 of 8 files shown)


lrytz:
spurious CI failure?

```
[info] - compi completions *** FAILED ***
[info]   assertion failed
[info]   
[info]   assert(complete("compi").toVector == expected)
[info]          |                 |        |  |
[info]          Stream()          Vector() |  Vector(compile, compile:, compileAnalysisFile, compileAnalysisFilename, compileAnalysisTargetRoot, compileEarly, compileIncSetup, compileIncremental, compileJava, compileOrder, compileOutputs, compileProgress, compileScalaBackend, compileSplit, compilerCache, compilers)
[info]                                     false
[info]     verify.asserts.PowerAssert$AssertListener.expressionRecorded(PowerAssert.scala:37)
```

SethTisue:
we'll need a better name. maybe something like `allowNewerScalaStandardLibrary`?

eed3si9n:
How about `unsafeScalaLibDowngrade`?

SethTisue:
@eed3si9n I like the spiciness and I agree that it ought to be at least a little spicy.

lrytz:
Happy to change the name. `unsafeScalaLibDowngrade` is maybe misleading, the setting allows a compiler older than the resolved standard library.

eed3si9n:
I see. Then `unsafeCompilerDowngrade`?

mzuehlke:
Doesn't the compiler stay at the version given by `scalaVersion` and the scala-library is of a newer version ?
So in my opinion the suggestion from @SethTisue reflect this better.

JD557:
Or `unsafeScalaLibUpgrade`?

lrytz:
Changed the name to `allowUnsafeScalaLibUpgrade`

eed3si9n (approved):
Thanks!

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1204,24 +1206,38 @@ object Defaults extends BuildCommon {
...
+                 |Compilation (macro expansion) or using the Scala REPL in sbt may fail with a LinkageError.""".stripMargin
+
+          val msg =
+            s"""Expected `$n/scalaVersion` to be $libVer or later, but found $sv.
+               |To support backwards-only binary compatibility (SIP-51), the Scala 2.13 compiler
+               |should not be older than $libName on the dependency classpath.
+               |
+               |$fix
+               |
+               |See `$n/evicted` to know why $libName $libVer is getting pulled in.
```
danicheg, line 1237: I know this is likely a minor issue, but if a user overrides the default derived project name (i.e., `sbt.Keys#name`), an invalid SBT command would be printed here:
```terminal
sbt:scalac-profiling-root> scalac-profiling/evicted
[error] Expected ID character
[error] Not a valid command: scalac-profiling
[error] Expected project ID
[error] Expected configuration
[error] Expected ':'
[error] Expected key
[error] Not a valid key: scalac-profiling (similar: scalafmtConfig, scalacOptions)
[error] scalac-profiling/evicted
[error]
```
IIUC, this is because `name` only affects the publishing artifacts' name and the display name in the SBT shell, and it cannot be used as a reference to the project in SBT shell. My quick attempt to workaround this has failed. Can you folks suggest anything here?

eed3si9n: I've split this into a new issue https://github.com/sbt/sbt/issues/8056

Issue #7705, opened 2024-09-30
[2.x] Wrong position of compiler errors in build.sbt


adpi2:
While trying to load [softwaremill/sttp](https://github.com/softwaremill/sttp) with sbt `2.0.0-alpha12-SNAPSHOT` I noticed that the positions of the compilation errors in `build.sbt` were wrong. It seems there are some negative offset.

I did not manage to reproduce this in a minimal `build.sbt` though.

xuwei-k:
> I did not manage to reproduce this in a minimal build.sbt though.


not negative but wrong position warning example.

sbt 2.0.0-M2


```scala
TaskKey[Int]("a1") := {
  Stream(1).sum
}

TaskKey[Int]("a2") := {
  Stream(2).sum
}

TaskKey[Int]("a3") := {
  Stream(3).sum
}
```


actual

```
-- Deprecation Warning: build.sbt:2:2 ----------------------
2 |  Stream(1).sum
  |  ^^^^^^
  |getter Stream in package scala is deprecated since 2.13.0: Use LazyList instead of Stream
-- Deprecation Warning: build.sbt:2:2 ----------------------
2 |  Stream(2).sum
  |  ^^^^^^
  |getter Stream in package scala is deprecated since 2.13.0: Use LazyList instead of Stream
-- Deprecation Warning: build.sbt:2:2 ----------------------
2 |  Stream(3).sum
  |  ^^^^^^
  |getter Stream in package scala is deprecated since 2.13.0: Use LazyList instead of Stream
```

expect

```diff
5,6c5,6
< -- Deprecation Warning: build.sbt:2:2 ----------------------
< 2 |  Stream(2).sum
---
> -- Deprecation Warning: build.sbt:6:2 ----------------------
> 6 |  Stream(2).sum
9,12c9,12
< -- Deprecation Warning: build.sbt:2:2 ----------------------
< 2 |  Stream(3).sum
---
> -- Deprecation Warning: build.sbt:10:2 ----------------------
> 10 |  Stream(3).sum
```

eed3si9n:
Here's a PR for this - https://github.com/sbt/sbt/pull/8013

PR #8013, opened 2025-01-20
[2.x] fix: Fix build.sbt position

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7705

## Problem
Compiler warnings and errors is not correct for build.sbt.

## Solution
This adds line offset to the synthetic Scala file
so the original line position is displayed in the errors.

## Test

```scala
$ sbt
[info] welcome to sbt 2.0.0-SNAPSHOT (Azul Systems, Inc. Java 17.0.12)
[info] loading project definition from /private/tmp/foo/project
-- Deprecation Warning: /private/tmp/foo/build.sbt:2:2 -------------------------
2 |  Stream(1).sum
  |  ^^^^^^
  |value Stream in package scala is deprecated since 2.13.0: Use LazyList instead of Stream
-- Deprecation Warning: /private/tmp/foo/build.sbt:6:2 -------------------------
6 |  Stream(2).sum
  |  ^^^^^^
  |value Stream in package scala is deprecated since 2.13.0: Use LazyList instead of Stream
-- Deprecation Warning: /private/tmp/foo/build.sbt:10:2 ------------------------
10 |  Stream(3).sum
   |  ^^^^^^
   |value Stream in package scala is deprecated since 2.13.0: Use LazyList instead of Stream
```

### DIFF
--- a/buildfile/src/main/scala/sbt/internal/Eval.scala
+++ b/buildfile/src/main/scala/sbt/internal/Eval.scala
@@ -97,8 +97,8 @@ class Eval(
           |  }
           |}
           |""".stripMargin
-        val startLine = header.linesIterator.toList.size
-        EvalSourceFile(srcName, startLine, contents)
+        val headerLine = header.linesIterator.toList.size
+        EvalSourceFile(srcName, headerLine, line - 1, contents)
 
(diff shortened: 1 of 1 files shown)


adpi2 (approved):
Thanks

PR #8016, opened 2025-01-23
provide overloaded sbt.SettingKey.apply for label + rank, for better sbt plugins cross compilation between sbt 1.x and 2.x

unkarjedy:
Example: sbt-structure plugin.
It is currently cross-compiled for sbt 0.13, 1.2 and 1.3 but I am adding support for 2.0 for IntelliJ compat.
https://github.com/JetBrains/sbt-structure/blob/e65f4ecab65724b9f627dbadc4d6e16239358fd9/extractor/src/main/scala/org/jetbrains/sbt/StructureKeys.scala#L12
```scala
  val dependencyConfigurations: SettingKey[Seq[Configuration]] = SettingKey("ssDependencyConfigurations", rank = Invisible)
  val sourceConfigurations: SettingKey[Seq[Configuration]] = SettingKey("ssSourceConfigurations", rank = Invisible)
  val testConfigurations: SettingKey[Seq[Configuration]] = SettingKey("ssTestConfigurations", rank = Invisible)
  val acceptedProjects: TaskKey[Seq[ProjectRef]] = TaskKey("ssAcceptedProjects", rank = Invisible)
```

It would be nice to avoid specifying empty description `description = ""` everywhere just to make it compile in Scala 3.
An alternative solution would be to use `withRank`:
```scala
val dependencyConfigurations: SettingKey[Seq[Configuration]] = SettingKey[Seq[Configuration]]("ssDependencyConfigurations").withRank(Invisible)
```
But this solution requires duplication of the type argument (`[Seq[Configuration]]` in this case).
It would be nice to avoid code modifications completely, which should be solved by the new `sbt.SettingKey.apply` overload

### DIFF
--- a/main-settings/src/main/scala/sbt/Structure.scala
+++ b/main-settings/src/main/scala/sbt/Structure.scala
@@ -816,6 +816,16 @@ object SettingKey:
 
   def apply[A1: ClassTag: OptJsonWriter](
       label: String,
+      rank: Int,
+  ): SettingKey[A1] =
+    apply[A1](
+      label = label,
+      description = "",
+      rank = rank,
+    )
+
+  def apply[A1: ClassTag: OptJsonWriter](
+      label: String,
       description: String,
       rank: Int
   ): SettingKey[A1] =


eed3si9n:
> It is currently cross-compiled for sbt 0.13

My suggestion would be to drop sbt 0.13 instead.

unkarjedy:
> My suggestion would be to drop sbt 0.13 instead.

The fact that it's cross-compiled to 0.13 is not essential here, the same code is used for 1.x
I mentioned this:
> cross-compiled for sbt 0.13, 1.2 and 1.3 

just for the context, but it might have been redundant.

Anyway, I am not blocked by this, I added `description = ""` everywhere just to move forward.
But the change could make sense for other plugins that will be cross-built for sbt 2.0.
It seems extra `apply` overload doesn't conflict with other definitions and shouldn't bring issues.

unkarjedy:
>My suggestion would be to drop sbt 0.13 instead.

We can't afford the luxury of just dropping the support for sbt 0.13.
This plugin is required to support SBT projects in IntelliJ IDEA.
There are still legacy projects that use sbt 0.13 in corps, they should be able to open it in the latest IntelliJ IDEA 
(e.g. they need the latest support for Java, but they also have some extra scala projects to maintain). 
I don't advocate for not upgrading the projects to sbt 1.x, but it's not my choice to stay on 0.13.

I would be glad to drop it eventually, but before that we need to provide users some help/tips with migration.
First, notify that the support will be dropped (~2 releases in advance) and ideally provide some tools/links to migrate.
That we can do, but it can't go in one step "drop".

eed3si9n:
> The fact that it's cross-compiled to 0.13 is not essential here, the same code is used for 1.x

In sbt 1.x, the idiomatic way of declaring keys are using the lower case macros e.g.:

```scala
val scalaVersion = settingKey[String]("The version of Scala used for building.")
```

We probably should've deprecated `apply` on `SettingKey` in sbt 1.x, but I guess it was there for 0.13 cross building initially and hasn't gotten much attention until now.

eed3si9n:
> There are still legacy projects that use sbt 0.13 in corps, they should be able to open it in the latest IntelliJ IDEA
(e.g. they need the latest support for Java, but they also have some extra scala projects to maintain).

sbt project EOL'ed sbt 0.13 https://www.scala-sbt.org/download#release-timeline in 2018 while it was under Lightbend, and currently it doesn't receive any patches including security issues. So I'm sympathetic to the position you're put in, but I don't really want to expand the API to support better source compatibility with legacy API esp if there's a workaround for it.

SethTisue:
> My suggestion would be to drop sbt 0.13 instead.

or branch rather than cross-build, I guess

I find it rather remarkable that cross-building against two sbt major versions is even supported; I agree that cross-building against three shouldn't even be a goal

unkarjedy:
> or branch rather than cross-build, I guess

Yeah, I also consider this option.
Then we would also need to ensure that the project structure model which used to share between sbt and IntelliJ processes is in-sync between branches. It is doable, though it will introduce some inconveniences. But maybe these inconveniences will be smaller compared to the cross-building against 0.13 😅

unkarjedy:
@eed3si9n Thanks for the comments.

BTW, is there a better place (chat/thread) where all the issues related to sbt plugins cross-compilation are discussed?
Some issues are still relevant for the sbt 1.10 + 2.0 scenario, but I wouldn't like to spam with separate tickets

PR #8028, opened 2025-01-31
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling)

The misspellings have been reported at https://github.com/jsoref/sbt/actions/runs/13080535350

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/sbt/actions/runs/13080535060#summary-36502868667

Note that sbt is using a [fork of a very very old version](https://github.com/sbt/ivy) of [apache-ivy](https://github.com/apache/ant-ivy) which [fixed a typo of `unknown` >8 years ago](https://github.com/apache/ant-ivy/commit/007939d9fe34faa83d4b242dcc37d2e1894c4f09) and is available in the [2.5.0](https://github.com/apache/ant-ivy/blame/rel/2.5.0/src/java/org/apache/ivy/core/cache/ArtifactOrigin.java) release... I've left that be, but it's kinda unfortunate. Usually, I'd be able to bump a project dependency to a newer version and then use the corrected spelling...

### DIFF
--- a/main-actions/src/main/scala/sbt/Pkg.scala
+++ b/main-actions/src/main/scala/sbt/Pkg.scala
@@ -40,8 +40,6 @@ import xsbti.{ FileConverter, HashedVirtualFileRef, VirtualFile, VirtualFileRef 
 object Pkg:
   def JarManifest(m: Manifest) = PackageOption.JarManifest(m)
   def MainClass(mainClassName: String) = PackageOption.MainClass(mainClassName)
-  def MainfestAttributes(attributes: (Attributes.Name, String)*) =
-    PackageOption.ManifestAttributes(attributes*)
   def ManifestAttributes(attributes: (String, String)*) = {
     val converted = for ((name, value) <- attributes) yield (new Attributes.Name(name), value)
     PackageOption.ManifestAttributes(converted*)
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2239,7 +2239,7 @@ object Defaults extends BuildCommon {
     val store = analysisStore(compileAnalysisFile)
     val contents = store.unsafeGet()
     if (exportP) {
-      // this stores the eary analysis (again) in case the subproject contains a macro
+      // this stores the early analysis (again) in case the subproject contains a macro
       setup.earlyAnalysisStore.toOption map { earlyStore =>
         earlyStore.set(contents)
       }
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -3499,8 +3499,8 @@ object Classpaths {
   )
 
   def warnResolversConflict(resolverList: Seq[Resolver], log: Logger): Unit = {
-    val resset = resolverList.toSet
-    for ((name, r) <- resset groupBy (_.name) if r.size > 1) {
+    val resolverSet = resolverList.toSet
+    for ((name, r) <- resolverSet groupBy (_.name) if r.size > 1) {
       log.warn(
         "Multiple resolvers having different access mechanism configured with same name '" + name + "'. To avoid conflict, Remove duplicate project resolvers (`resolvers`) or rename publishing resolver (`publishTo`)."
       )
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2822,7 +2822,7 @@ object Classpaths {
           val vf = converter.toVirtualFile(p)
           FileStamp(stamper.library(vf)).map(p -> _)
       },
-      // Note: invoking this task from shell would block indefinately because it will
+      // Note: invoking this task from shell would block indefinitely because it will
       // wait for the upstream compilation to start.
       dependencyPicklePath := {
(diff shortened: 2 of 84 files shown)


sbt.sh
eed3si9n, line 1: Could we exclude this change from this commit?

eed3si9n, line 1: Also this.

main/src/main/scala/sbt/internal/Load.scala
```diff
@@ -1462,7 +1462,7 @@ private[sbt] object Load {
 
   /**
    * Creates a classloader with a hierarchical structure, where the parent
-   * classloads the dependency classpath and the return classloader classloads
+   * class loads the dependency classpath and the return classloader class loads
```
eed3si9n: IMO _classload_ as a verb is fine

sbt-app/src/sbt-test/cache/optout/build.sbt
```diff
@@ -6,7 +6,7 @@ Global / localCacheDirectory := baseDirectory.value / "diskcache"
 
 aa := A()
 
-// This tests that aa is opt'ed out from caching
+// This tests that aa is opt-ed out from caching
```
eed3si9n: The past tense of opt should be opted, without a hyphen?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -3498,17 +3498,17 @@ object Classpaths {
     }
   )
 
-  def warnResolversConflict(ress: Seq[Resolver], log: Logger): Unit = {
-    val resset = ress.toSet
+  def warnResolversConflict(resolvers: Seq[Resolver], log: Logger): Unit = {
```
eed3si9n: Given that resolves is a key name in sbt, I don't think this change is good.

main-actions/src/main/scala/sbt/Pkg.scala
```diff
@@ -40,8 +40,6 @@ import xsbti.{ FileConverter, HashedVirtualFileRef, VirtualFile, VirtualFileRef
 object Pkg:
   def JarManifest(m: Manifest) = PackageOption.JarManifest(m)
   def MainClass(mainClassName: String) = PackageOption.MainClass(mainClassName)
-  def MainfestAttributes(attributes: (Attributes.Name, String)*) =
-    PackageOption.ManifestAttributes(attributes*)
```
eed3si9n, line 44: This should be split to another PR.

jsoref: This is because `Mainfest` is a lousy misspelling of `Manifest` and the default correction resulted in a conflict with the next function ...

sbt.sh
jsoref: Oh, I meant to not do that... It's a bug in some of my tooling -- it doesn't know not to ignore symlinks...

main/src/main/scala/sbt/Defaults.scala
jsoref: Oh, is this `resolutions`?

main/src/main/scala/sbt/internal/Load.scala
jsoref: I spoke w/ one of my coworkers who works in Scala about this before submitting the PR, the support for this change was a few lines later which uses `class load` in the `@return` statement:
https://github.com/sbt/sbt/blob/0a1a36a2b34a5eefdb016b6eae6dd4f8ea43b5ca/main/src/main/scala/sbt/internal/Load.scala#L1464-L1471

If code is written in a way that shows it believes a word is a word, I don't try to rewrite things based on my personal beliefs, I really try to follow Rome, and this is where I found Rome.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: I meant `resolvers` not resolves, but what I mean is that within Defaults.scala, `Keys.*` are used without `Keys.`, so creating a parameter named `resolvers` would shadow `Keys.resolvers`. So in this particular case, using abbreviated `ress` I think was intentional, not a bug.

jsoref: If something isn't obvious, it's a poor choice. Would `resolverList` work?

eed3si9n: Sure. resolverList is fine.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -3498,17 +3498,17 @@ object Classpaths {
...
+    for ((name, r) <- resolverSet groupBy (_.name) if r.size > 1) {
       log.warn(
         "Multiple resolvers having different access mechanism configured with same name '" + name + "'. To avoid conflict, Remove duplicate project resolvers (`resolvers`) or rename publishing resolver (`publishTo`)."
       )
     }
   }
 
-  private[sbt] def errorInsecureProtocol(ress: Seq[Resolver], log: Logger): Unit = {
-    val bad = !ress.forall(!_.validateProtocol(log))
+  private[sbt] def errorInsecureProtocol(resolvers: Seq[Resolver], log: Logger): Unit = {
```
eed3si9n: Same situation here.

eed3si9n (approved):

Issue #7707, opened 2024-09-30
[2.x] Missing implementation of `Def.inputTaskDyn`


adpi2:

bjaglin:
Actually, the implementation is buggy in sbt 1.x, see https://github.com/sbt/sbt/issues/3572. sbt-scalafix relies on [internals](https://github.com/sbt/sbt/issues/3572#issuecomment-643609711) (missing as well in sbt 2.0.0-M2) to workaround that.

unkarjedy:
I will just copy my message from discord
https://discord.com/channels/632150470000902164/922600050989875282/1332018699363946496

> Hello! 
> 
> I am making an early attempt to cross-build [sbt-structure](https://github.com/JetBrains/sbt-structure) plugin against sbt 2.0.0-M3 (+ sbt 1.x and sbt 0.13).  I
> t's required to support sbt 2.0 in IntelliJ. 
> So far it goes pretty painful, a lot of code doesn't cross-compile. 
> 
> I might ask more questions in the future, but for now I have a question about `sbt.Def.inputTaskDyn`
> I see that it was removed in sbt 2, but we [use it](https://github.com/JetBrains/sbt-structure/blob/38de1c9edc474e263221868d674427eeb7cac800/extractor/src/main/scala/org/jetbrains/sbt/StructurePlugin.scala#L34) in our code. 
> Was it removed intentionally? 
> If so, what is supposed to be the replacement for it?
> 
> Also, is there any dedicated ticket/thread to discuss sbt 2.0 migration issues? 
> Or should I just ask questions here?
> Would be nice to gather all the issues in one place and eventually update:
> https://www.scala-sbt.org/2.x/docs/en/changes/migrating-from-sbt-1.x.html

The code of the [task](https://github.com/JetBrains/sbt-structure/blob/38de1c9edc474e263221868d674427eeb7cac800/extractor/src/main/scala/org/jetbrains/sbt/StructurePlugin.scala#L34):
```scala
  lazy val dumpStructureTo: Def.Initialize[InputTask[File]] = Def.inputTaskDyn {

    val outputFile = targetFileParser.parsed
    val options = StructureKeys.sbtStructureOpts.value

    val log = Keys.streams.value.log
    val structureTask = extractors.extractStructure

    Def.task {
      val structure = structureTask.value.serialize
      val outputText = {
        if (options.prettyPrint) newXmlPrettyPrinter.format(structure)
        else xml.Utility.trim(structure).mkString
      }

[...]

PR #8033, opened 2025-02-05
[2.x] Implement Def.inputTaskDyn

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7707

## Problem/Solution
This implements Def.inputTaskDyn using Scala 3 macros.


adpi2 (commented):
Looks good, but we should probably use it more internally to reinforce it. In particular, you can rewrite these with `Def.inputTaskDyn`: https://github.com/sbt/sbt/blob/613eb86447ccf04d6d6dc0c833fe69fc4a2963f9/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala#L747-L756
https://github.com/sbt/sbt/blob/613eb86447ccf04d6d6dc0c833fe69fc4a2963f9/main/src/main/scala/sbt/Defaults.scala#L4702-L4717

adpi2:
I went ahead and pushed https://github.com/sbt/sbt/pull/8033/commits/711df2cbf2ded91b13588918b29d2ce6b4f7e813 for testing. Feel free to revert if you prefer.

eed3si9n:
Nice. These are probably `Def.inputTaskDyn` on sbt 1.x but were turned into flatMapTask because we didn't have the macro.

Issue #8014, opened 2025-01-21
`sbt` launcher says it downloads for `x86_64` on MacOS `arm64`


lolgab:
## steps
```
SBT_NATIVE_CLIENT=true sbt
```
on a arm64 Mac OS

## problem

The `sbt` launcher uses a fixed `arch=x86_64` value for logging on `darwin` which is used to print a wrong log like:
```
downloading sbtn 1.10.5 for x86_64
```

## expectation

It shouldn't say `downloading sbtn 1.10.5 for x86_64` but not mention the arch since it's downloading a universal binary.

eed3si9n:
Thanks for the report!

PR #8038, opened 2025-02-13
Fix incorrect sbt architecture logging on MacOS arm64

mehdignu:
small fix for #8014

### DIFF
--- a/sbt
+++ b/sbt
@@ -200,6 +200,7 @@ acquire_sbtn () {
       exit 2
     fi
   elif [[ "$OSTYPE" == "darwin"* ]]; then
+    arch="universal"
     archive_target="$p/sbtn-universal-apple-darwin-${sbtn_v}.tar.gz"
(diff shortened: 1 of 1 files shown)


sbt
```diff
@@ -189,7 +189,10 @@ acquire_sbtn () {
   local target="$p/sbtn"
   local archive_target=
   local url=
-  local arch="x86_64"
+  local arch
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+    arch="universal"
+  fi
```
eed3si9n: There's 

```bash
  elif [[ "$OSTYPE" == "darwin"* ]]; then
```

at line 205, so maybe move that in there? also this means `arch` would be unset on Windows?

mehdignu: Thanks for the feedback! that's better solution :)

eed3si9n (approved):
Thanks!

PR #8018, opened 2025-01-27
[2.x] Implement worker + forked console

eed3si9n:
Ref https://github.com/sbt/sbt/issues/1918

## Problem
1. See https://eed3si9n.com/rfc-4-persistent-worker/ for the general motivation. We want to move long-running tasks out of the command loop, and potentially from the sbt process entirely.
2. `console / fork` issue was created a long time ago since it's closely related to running and testing, which already can fork.

## Solution
1. This implements a pure Java app called worker, which can run arbitrary application. For now I'm using property files to communicate. I'll likely switch to using JSON. The worker currently implements `run` command, which can run a program specified in the properties file.
2. This also implements `ConsoleMain`, which is a command line program that wraps Zinc's `console` function.

## Caveat
One major limitation of the `Compile / console / fork := true` implementation at this time is that I wasn't able to configure JLine terminal to xterm-256color, which uses a sequence of characters to send up and down keys. So while the color works, up and down keys do not, which is inconvenient for Scala repl. Nonetheless, this is still a useful demonstration of the forked task running in a worker process that's able to accept ordinary key inputs.

PR #8040, opened 2025-02-17
[1.x] Implement client-side run

eed3si9n:
**Update**: Here's a blog post [sudori part 7: client-side run with sbt](https://eed3si9n.com/sudori-part7-client-side-run-with-sbt/)

## Problem
`run` task blocks the server, but during the run the server is just waiting for the built program to finish.

## Solution
This implements client-side run where the server creates a sandbox environment, and sends the information to the client, and the client forks a new JVM to perform the run.


adpi2:
I am not sure what's the direction you want to go anymore, for non-blocking tasks:
- We already have `BackgroundJobService`, which is good for JIT-ing but have issues with forwarding the standard input and output.
- You started implementing workers in #8018, which could potentially solve the in/out forwarding issue without compromising the JIT-ing (persistent workers).
- In this PR we have client-side runner, which solves the in/out forwarding issue, but not the JIT-ing. Also can we implement it for tests? Incremental tests need to access the cache.

eed3si9n:
In https://github.com/sbt/sbt/pull/8018#issuecomment-2641936493 I wrote
> Next step would be to let sbtn start the fork process into the worker.

- The key ability that I'm after in this PR is sbtn being able to expose `sbt/general` endpoint. Once we have that, sbt 2.x server can also use that to launch things like `console` using workers.
- For tests, we'd want persistent worker probably launched by the server, and maybe some mechanism to talk to the persistent worker, similar to forked test today using sockets.

eed3si9n:
At the end of  [sudori part 7: client-side run with sbt](https://eed3si9n.com/sudori-part7-client-side-run-with-sbt/) I've added 'motivation and discussion' section and a summary tradeoff table:

| feature                    | sbt 1.x | sbt 2.x | availability    | console / stdin | JIT perf | isolation |
|----------------------------|:-------:|:-------:|:----------------:|:---------------:|:--------:|:--------:|
| server-side x              |    ✅   |    ✅  |        ❌        |        ⚠️      |     ✅    |   ❌      |
| client-side x              |    ✅   |    ✅  |        ✅        |        ✅      |     ⚠️    |     ✅    |
| persistent worker (planned)|    ❌   |   ✅   |        ✅        |        ❌      |     ✅    |     ✅    |

So this hopefully clarifies that to improve the sbt server availability, we should go for client-side x (run, console, etc) if the task requires stdin; and try the persistent worker route for performance sensitive tasks like test.

eed3si9n:
Here's a program I wrote to test stdin:

```scala
package example

import scala.io.StdIn

@main def main(args: String*): Unit =
  println("")
  Console.out.print("name: ")
  Console.out.flush()
  val name = StdIn.readLine()
  println(s"hello, $name")
```

It looks like stdin is ok, but stdout doesn't flush on sbtn out-of-box:

[asciicast](https://asciinema.org/a/704004)

Nothing prints out until I hit return key.

adpi2:
> At the end of [sudori part 7: client-side run with sbt](https://eed3si9n.com/sudori-part7-client-side-run-with-sbt/) I've added 'motivation and discussion' section and a summary tradeoff table:

Great write-up! This helps clarify a lot of things. I agree that running on the client-side simplifies many of the challenges we are currently facing with `run` or `console`. It's unfortunate that we don't have a one-size-fits-all solution for long-running tasks. But that's what it is. `test` is a different task and it needs different trade-offs compared to `run` or `console`.

Sorry for being a bit slow to respond. I'll review that PR soon.

protocol/src/main/scala/sbt/protocol/Serialization.scala
```diff
@@ -27,6 +27,7 @@ object Serialization {
   private[sbt] val VsCode = "application/vscode-jsonrpc; charset=utf-8"
   val readSystemIn = "sbt/readSystemIn"
   val cancelReadSystemIn = "sbt/cancelReadSystemIn"
+  val general = "sbt/general"
```
adpi2: Could we find a better name than `sbt/general`?

My suggestions:
- `sbt/clientRun`
- `sbt/notifyClientRun`
- `sbt/triggerClientRun`
- `sbt/clientExecution`
...

protocol/src/main/contraband/worker.contra
```diff
@@ -0,0 +1,31 @@
...
+  outputStrategy: String
+  workingDirectory: java.net.URI
+  jvmOptions: [String]
+  environmentVariables: StringStringMap!
+  inputs: [sbt.internal.worker.FilePath] @since("0.1.0"),
+  outputs: [sbt.internal.worker.FilePath] @since("0.1.0"),
+  cmd: String @since("0.1.0"),
+}
+
+## Parameter for the sbt/general command, which is a generic command to
```
adpi2: To me, 'command' typically refers to a shell command or a request to server.

So I would rather write something like this:
```suggestion
## Parameter for the sbt/general notification sent by the server to trigger a program execution on the client side
```

protocol/src/main/contraband/worker.contra
```diff
@@ -0,0 +1,31 @@
...
+  connectInput: Boolean!
+  javaHome: java.net.URI
+  outputStrategy: String
+  workingDirectory: java.net.URI
+  jvmOptions: [String]
+  environmentVariables: StringStringMap!
+  inputs: [sbt.internal.worker.FilePath] @since("0.1.0"),
+  outputs: [sbt.internal.worker.FilePath] @since("0.1.0"),
+  cmd: String @since("0.1.0"),
+}
```
adpi2: Can we split this in two: `JVMRunInfo` and `NativeRunInfo`. And use it like this maybe:
```
type GeneralParams {
  jvm: Boolean!,
  jvmRunInfo: JVMRunInfo,
  nativeRunInfo: NativeRunInfo,
}
```

main/src/main/scala/sbt/Keys.scala
```diff
@@ -437,6 +438,8 @@ object Keys {
   val bspScalaMainClasses = inputKey[Unit]("Implementation of buildTarget/scalaMainClasses").withRank(DTask)
   val bspScalaMainClassesItem = taskKey[ScalaMainClassesItem]("").withRank(DTask)
   val bspReporter = taskKey[BuildServerReporter]("").withRank(DTask)
+  val bspGeneral0 = inputKey[GeneralParams]("Implementation of sbt/general command").withRank(DTask)
+  val bspGeneralRunInfo = inputKey[GeneralParams]("Implementation of general run info").withRank(DTask)
```
adpi2: I am not sure how this relates to BSP. It's more like an `sbtn`-specific feature for now.

The client-side run mechanism could be added to BSP in the future but currently it is not.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -801,6 +807,58 @@ object BuildServerProtocol {
     .map(_.mkString)
     .map(JsonParser.parseFromString)
 
+  private def bspGeneral0Task: Def.Initialize[InputTask[GeneralParams]] = Def.inputTaskDyn {
+    val tokens = spaceDelimited().parsed
+    val state = Keys.state.value
+    val p = Act.aggregatedKeyParser(state)
+    if (tokens.isEmpty) {
+      sys.error("usage: bspGeneral0 run")
+    }
```
adpi2: That's kind of harsh.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -801,6 +807,58 @@ object BuildServerProtocol {
...
+    val scopedKey = Parser.parse(tokens.head, p) match {
+      case Right(x :: Nil) => x
+      case Right(xs)       => sys.error("too many keys")
+      case Left(err)       => sys.error(err)
+    }
+    bspGeneralRunInfo.in(scopedKey.scope).toTask(" " + tokens.tail.mkString(" "))
```
adpi2: So `bspGeneral0` is a global settings that takes a `scopedKey` as first input.

Here I guess we could check that `scopedKey.key == run`.

protocol/src/main/contraband/worker.contra
```diff
@@ -0,0 +1,31 @@
+package sbt.internal.worker
+@target(Scala)
+@codecPackage("sbt.internal.worker.codec")
+@fullCodec("JsonProtocol")
+
+type FilePath {
+  path: java.net.URI!
+  digest: String!
```
adpi2, line 8: Why do we need `digest` in `FilePath`? It feels like a premature design decision.

main/src/main/scala/sbt/internal/server/NetworkChannel.scala
```diff
@@ -141,6 +141,20 @@ final class NetworkChannel(
       self.onCancellationRequest(execId, crp)
   }
 
+  // Take over commandline for network channel
+  private val networkCommand: PartialFunction[String, String] = {
+    case cmd if cmd.split(" ").head.split("/").last == "run" =>
+      s"bspGeneral0 $cmd"
```
adpi2: That seems really fragile. It's not going to work if there is any '/' in the arguments. For example: `foo/run --file path/to/file`.

Maybe we should prefix all network commands by `bspGeneral0` and let the task implement the logic of switching the `run` key by `bspRunInfo`.

adpi2 (commented):
I have written some comments. But my main feedback is that I find the overall design quite weak.

Replacing `run` by `bspGeneral0 run` in `NetworkChannel` does not seem robust enough:

- It won't work in cases like:

```scala
customRun := run.toTask(somePredefinedArguments).value
```
- It will break all plugins that overrides `run`, such as Scala.js or Scala Native. When running `sbtn run`, it will attempt to use `bspRunInfo` which will incorrectly launch a JVM process, instead of the JS or native one.

I believe the logic for switching between server-side and client-side execution should be handled within the `run` task itself.

It may be too late to implement it this way in sbt 1.x, but in sbt 2.x it could look like this:
```scala
run := {
  val state = Keys.state.value
  val info = Keys.runInfo.value
  if (state.isNetworkCommand) {
    state.notifyEvent(clientRunInfo, info)
  } else {
    run(info)
  }
 ```
 
With this approach, `sbt-scalajs` can override the `runInfo` task and get the client/server switching automatically.

File: protocol/src/main/contraband/worker.contra (hunk 7 above)
Comment (eed3si9n) [reply protocol/src/main/contraband/worker.contra:8]: I have this information readily available in sbt 2.x - https://github.com/sbt/sbt/pull/8042/files#diff-2ad23bc2de451984eb3983aceb2f15262cd8748b845faebae937a4d8f762f244R914-R915
```scala
    val cpPaths = cp.toVector.map: x =>
      FilePath(conv.toPath(x.data).toUri(), x.data.contentHashStr)
```
Also this is a design I prosed in RFC-4 - https://eed3si9n.com/rfc-4-persistent-worker/ as well.

protocol/src/main/scala/sbt/protocol/Serialization.scala
eed3si9n: The intent of `sbt/general` is meant to be a generic shape to pass executable around, not just from server to client, but also from server to the worker. Until we have the worker going, I'm fine with narrowly scoping it something more scoped, but client-side run doesn't have to cover just `run`, but also `console` etc or even plugin commands that can be executed either at client-side or worker-side.

adpi2: `sbt/non-blocking-exec` is a better name maybe

However I don't think it is a good idea to have one special command that has knowledge about where and how to run other tasks.

eed3si9n:
I wrote

> Nothing prints out until I hit return key.

While some hacking on a plane, I discovered a one-line fix to this bug, that's likely been in sbt for a while (https://github.com/sbt/sbt/pull/8040/commits/fa57f031de735bc84a3b9eda4541a7cb6fbea448). Maybe I'll just split this out to another PR.

main/src/main/scala/sbt/Keys.scala
eed3si9n: Yea. I guess I don't mentally make distinction between the IDE-supporting/BSP endpoints vs the custom endpoints sbt server has to support sbtn, esp since some of these sbtn features are added to support the BSP usage. In any case, I've split the new stuff into `main/src/main/scala/sbt/internal/server/JobSpec.scala` in 02386dfb607c78d45c59a6cb0a5280376e4fc741

main/src/main/scala/sbt/internal/server/NetworkChannel.scala
eed3si9n: I'm first splitting by whitespace in `cmd.split(" ").head` so `foo/run --file path/to/file` will still match.

protocol/src/main/scala/sbt/protocol/Serialization.scala
eed3si9n: In 02386dfb607c78d45c59a6cb0a5280376e4fc741, I've renamed `sbt/general` to `sbt/jobSpec`.

What's being implemented here is a somewhat new notion in sbt. It doesn't fit into command/task/setting, but rather it's sort of a side effect that happens typically inside the task. We can call it a "job", like `BackgroundJobService`. A description of a job, detailed enough for someone to start one, can be called a jobSpec, similar to [ServiceSpec](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec) in Kubernetes.

> However I don't think it is a good idea to have one special command that has knowledge about where and how to run other tasks.

Technically speaking, `sbt/jobSpec` (fka `sbt/general`) itself doesn't have any knowledge of other task. In general, adding a prefix to an existing task to make sbt do something is quite traditional. See also `;`, `~`, `+`, `-` (now called early). Mentally, we can think of `sbt/jobSpec` command's purpose is to translate an existing sbt task into a shell or `java` call.

eed3si9n:
> I have written some comments. But my main feedback is that I find the overall design quite weak.
> 
> Replacing `run` by `bspGeneral0 run` in `NetworkChannel` does not seem robust enough:
> 
> * It won't work in cases like:
> 
> ```scala
> customRun := run.toTask(somePredefinedArguments).value
> ```

Won't work in a sense that it will remain a server-side run, so it's no different from the status quo in sbt 1.x.
 
> * It will break all plugins that overrides `run`, such as Scala.js or Scala Native. When running `sbtn run`, it will attempt to use `bspRunInfo` which will incorrectly launch a JVM process, instead of the JS or native one.

I'm not sure if sbtn and/or forked run works with Scala.js or Scala Native in sbt 1.x world as-is, so I think it's fine. If someone wants to bridge the gap, they can rewire `jobSpecRunInfo` to provide sbtn-specific behavior.

> I believe the logic for switching between server-side and client-side execution should be handled within the `run` task itself.
> 
> It may be too late to implement it this way in sbt 1.x, but in sbt 2.x it could look like this:
> 
> ```scala
> run := {
>   val state = Keys.state.value
>   val info = Keys.runInfo.value
>   if (state.isNetworkCommand) {
>     state.notifyEvent(clientRunInfo, info)
>   } else {
>     run(info)
>   }
> ```
> 
> With this approach, `sbt-scalajs` can override the `runInfo` task and get the client/server switching automatically.

Yea, I agree that it's probably too late to do that in sbt 1.x, and I'm open to pushing this into `run` task in sbt 2.x since sbtn is planned to be the default.

protocol/src/main/contraband/worker.contra
eed3si9n: Sure. I'm using notification only because sbtn isn't really sending requests that I can respond against, but I will s/command/notification/.

protocol/src/main/contraband/worker.contra
eed3si9n: Addressed in 09f412dcb82b077ddb597a8cac30068b0a129632

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
eed3si9n: Changed to `sys.error("expected an argument, for example foo/run")` in 09f412dcb82b077ddb597a8cac30068b0a129632, but generally I'm not expecting users to hit this since it's an internal task.

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
eed3si9n: Added

```scala
    if (scopedKey.key == run.key)
      jobSpecRunInfo.in(scopedKey.scope).toTask(" " + tokens.tail.mkString(" "))
    else sys.error(s"unsupported task for jobSpec $scopedKey")
```

in 09f412dcb82b077ddb597a8cac30068b0a129632

protocol/src/main/scala/sbt/protocol/Serialization.scala
eed3si9n: As a hybrid, I can name the method as `sbt/clientJob`. This way, it's more specific to this particular use case, and less _general_.

eed3si9n:
One other aspect that favors creating a new inputTask is controlling `[success]` printing (see also https://github.com/sbt/sbt/issues/8044). I haven't done it yet in this PR, it's possible that we might need to suppress printing `[success]` until client-side run is done.

sjrd:
> > * It will break all plugins that overrides `run`, such as Scala.js or Scala Native. When running `sbtn run`, it will attempt to use `bspRunInfo` which will incorrectly launch a JVM process, instead of the JS or native one.
> 
> I'm not sure if sbtn and/or forked run works with Scala.js or Scala Native in sbt 1.x world as-is, so I think it's fine. If someone wants to bridge the gap, they can rewire `jobSpecRunInfo` to provide sbtn-specific behavior.

`sbtn` works fine on Scala.js and Scala Native in sbt 1.x.

`fork` is ignored for `run`, which means it "works" (anyway it's always forked in practice). If `Test/fork` is true, there is a nice error message telling users they should not use `Test/fork := true` in Scala.js projects.

The problem with this PR, as @adpi2 explained it to me, is that it completely breaks `run` in Scala.js. Worse, I don't think it's repairable. In Scala.js, running and testing *must* happen in the process where sbt settings and tasks live. The reason is that we have a `jsEnv: TaskKey[JSEnv]`, and we use the methods of that `JSEnv` to run and test. `JSEnv` is an intentionally open trait. There are several implementations in the wild, implementing support for various JS engines. That means `JSEnv`s *cannot* be serialized to be sent to another process. Therefore, running and testing *must* stay within the main sbt process.

I suspect the same is true in Scala Native at least for testing. For running, it might be salvageable.

eed3si9n:
ok. Since I'm mostly after creating the client-side mechanism on sbtn, I can revert the server-side switching to turn `run` into `clientJob run` in sbt 1.x, and in the sbt 2.x world we'll provide some mechanism for the plugins to opt out of client-side run.

PR #8045, opened 2025-02-21
[2.x] remove unnecessary `case`

xuwei-k:
- https://github.com/xuwei-k/scalafix-rules/blob/3ca36335057154ef4431fee464a86b721013d334/rules/src/main/scala/fix/RemoveParameterUntuplingCase.scala
- https://docs.scala-lang.org/scala3/reference/other-new-features/parameter-untupling.html
- https://docs.scala-lang.org/scala3/reference/other-new-features/parameter-untupling-spec.html


eed3si9n (approved):
FYI - I'm +1 with this change, but generally we still need to maintain both sbt 1.x and 2.x, so these cross-cutting changes make merge-1.x pull requests more difficult to apply later.

som-snytt:
That's a lot of cases. IMHO it really shines with colon syntax. (Example in `Scope` of "colon arrow EOL" with the params after the colon.) Less shiny if it still has braces.

I wonder if it's feasible for a "merge tool" to rewrite the target to old syntax, apply a patch, then rewrite back. (The old syntax would also require old scalafmt config.)

PR #8057, opened 2025-03-05
[1.x] Shortcut sbtn shutdown

eed3si9n:
Fixes https://github.com/sbt/sbt/discussions/8055

## Problem
WHen portfile is not found, sbtn will try to start the server even when the desired command is shutdown.

## Solution
When the command is just shutdown, don't start the server.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,7 +11,7 @@ import scala.util.Try
 // ThisBuild settings take lower precedence,
 // but can be shared across the multi projects.
 ThisBuild / version := {
-  val v = "1.10.10-SNAPSHOT"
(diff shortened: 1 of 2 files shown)


adpi2 (approved):

NthPortal (approved):
thanks @eed3si9n!

SethTisue:
> no sbt server is running. ciao

no-a sbt server is a-running! ciao! 🇮🇹

NthPortal:
what causes this to make it to coursier?

eed3si9n:
Not sure what Coursier uses to index sbtn, but next time I publish sbt 1.x, I'll build sbtn together.

PR #8060, opened 2025-03-09
[2.x] Implement client-side run

eed3si9n:
Here's a blog post [sudori part 7: client-side run with sbt](https://eed3si9n.com/sudori-part7-client-side-run-with-sbt/)
Ref https://github.com/sbt/sbt/pull/8040
Ref https://github.com/sbt/sbt/pull/7688
Fixes https://github.com/sbt/sbt/issues/6178

| feature                    | sbt 1.x | sbt 2.x | availability    | console / stdin | JIT perf | isolation |
|----------------------------|:-------:|:-------:|:----------------:|:---------------:|:--------:|:--------:|
| server-side x              |    ✅   |    ✅  |        ❌        |        ⚠️      |     ✅    |   ❌      |
| client-side x              |    ❌   |    ✅  |        ✅        |        ✅      |     ⚠️    |     ✅    |
| persistent worker (planned)|    ❌   |   ✅   |        ✅        |        ❌      |     ✅    |     ✅    |

## Problem
`run` task has been emulated via function call inside of a sandboxed classloader, and blocking the command processing of sbt server loop. This poses isolation and availability issues.

## Solution
As a follow-up to  https://github.com/sbt/sbt/pull/8040, this implements client-side run where the server creates a sandbox environment, and sends the information to the client, and the client forks a new JVM to perform the run. The client-side behavior has been implemented in sbtn side already.


main/src/main/scala/sbt/internal/RunUtil.scala
```diff
@@ -0,0 +1,301 @@
...
+    Def.inputTask {
+      val conv = fileConverter.value
+      given FileConverter = conv
+      val args = parser.parsed
+      val service = bgJobService.value
+      val mainClass = getMainClass(mainClassTask.value)
+      val hashClasspath = (bgRun / bgHashClasspath).value
+      val fo = (run / forkOptions).value
+      val state = Keys.state.value
+      if clientRun.value && state.isNetworkCommand then
```
eed3si9n, line 184: As suggested by @adpi2, this moves the condition logic into `run` input task, which allows plugins to opt out of this behavior.

adpi2:
That's great! Thank you!

PR #8114, opened 2025-05-09
[2.x] add `using` if explicitly providing arguments to methods defined with `implicit`

xuwei-k:
- https://github.com/scala/scala3/pull/22441


eed3si9n (approved):
Thanks!

PR #8126, opened 2025-05-15
[1.x] Sonatype Publish Portal support

eed3si9n:
## Problem
Sonatype is [sunsetting](https://central.sonatype.org/news/20250326_ossrh_sunset/) HTTP Rest endpoint OSSRH in June 2025. Adding this feature to sbt-sonatype has been complicated by the fact that some underlying libraries seem to no longer work on JDK 8, which sbt 1.x would like to maintain.

## Solution
(Per popular demand) This implements a built-in publishing mechanism via the [Portal Publisher API](https://central.sonatype.org/publish/publish-portal-api/).
Kudos to David Doyle (@Andrapyre) who has spearheaded Sonatype Central support via sonatype-central-client etc.

This implementation leans on existing sbt toolchain such as Contraband for JSON codec generation, Gigahorse for HTTP, and using a task to create a zip file via `IO.zip`.

## Example

```scala
ThisBuild / publishTo := localStaging.value
```

Run `publishSigned` etc, and then `sonaUpload` command.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -677,6 +677,7 @@ lazy val dependencyTreeProj = (project in file("dependency-tree"))
 
 // Implementation and support code for defining actions.
 lazy val actionsProj = (project in file("main-actions"))
+  .enablePlugins(ContrabandPlugin, JsonCodecPlugin)
   .dependsOn(
     completeProj,
     runProj,
@@ -690,8 +691,13 @@ lazy val actionsProj = (project in file("main-actions"))
   .settings(
     testedBaseSettings,
     name := "Actions",
+    Compile / scalacOptions += "-Xsource:3",
     libraryDependencies += sjsonNewScalaJson.value,
-    libraryDependencies += jline3Terminal,
+    libraryDependencies ++= Seq(gigahorseOkHttp, jline3Terminal),
+    Compile / managedSourceDirectories +=
+      baseDirectory.value / "src" / "main" / "contraband-scala",
+    Compile / generateContrabands / sourceManaged := baseDirectory.value / "src" / "main" / "contraband-scala",
+    Compile / generateContrabands / contrabandFormatsForType := ContrabandConfig.getFormats,
     mimaSettings,
     mimaBinaryIssueFilters ++= Seq(
       // Removed unused private[sbt] nested class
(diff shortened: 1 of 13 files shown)


chitralverma:
@eed3si9n few questions 
- Does this handle snapshots as well?
- are sbt docs also updated?
- should probably raise a request to update sonatype docs as well with sbt examples and config

Andrapyre:
@eed3si9n , I'm happy to have you or someone from the sbt team be an official maintainer on the sonatype central client project so that you can guarantee faster updates and some control (I'm assuming that's the motivation for copying the code instead of having the dependency on the client). Advantage to that is that you can integrate the gigahorse client with the testing structure there and take advantage of a broader feature set. Just let me know and excited for this to come out. I agree that it is a better direction than a new plugin.

eed3si9n:
@chitralverma 

> - Does this handle snapshots as well?

I'm not sure. I actually don't have a working Sonatype account using Publisher Portal, so I have only tested uploading the bundle, but not actually publishing it. I don't have a code that's actively preventing it, so it might work? In general, publishing SNAPSHOT outside of a single computer is a bad practice.

> - are sbt docs also updated?

This is just a pull request. If there's sbt version released to the world with this feature, I'll probably update the docs.

MateuszKubuszok:
I tested publishing to snapshots and no it does not work - deployment fails with `version cannot be a SNAPSHOT`. In the `sbt-sontype` plugin here was a [distinct URL](https://central.sonatype.org/publish/publish-portal-snapshots/#publishing-via-other-methods) for snapshot releases used when version ended with `-SNAPSHOT`:

```scala
   // version from https://github.com/xerial/sbt-sonatype/pull/583/files#diff-2b141e540213546395916a631aaf8da162db0f8a4782d0b26dfd6f451f3e35ceR138
   sonatypeDefaultResolver := {
      if (sonatypeCredentialHost.value == SonatypeCentralClient.host) {
      if (sonatypeCredentialHost.value == sonatypeCentralHost) {
        Resolver.url(s"https://${sonatypeCredentialHost.value}")
      } else {
        val profileM   = sonatypeTargetRepositoryProfile.?.value
        val repository = sonatypeRepository.value
        val staged = profileM.map { stagingRepoProfile =>
          s"${sonatypeCredentialHost.value.replace('.', '-')}-releases" at s"${repository}/${stagingRepoProfile.deployPath}"
        }
        staged.getOrElse(if (version.value.endsWith("-SNAPSHOT")) {
          sonatypeSnapshotResolver.value
        } else {
          sonatypeStagingResolver.value
        })
      }
    },
```
There has to be some special handling performed for it to work.

PR #8148, opened 2025-06-02
fix: Default sbtPluginPublishLegacyMavenStyle to false

### LINKED ISSUE #8137: Sonatype central publish fails for sbt plugin
## steps
sbt version 1.11.0
Trying to publish [sbt-aether-deploy](https://github.com/arktekk/sbt-aether-deploy/) from my laptop. 
sbt publishSigned
sbt sonaUpload

## problem

It seems like this generates the wrong file names when creating the bundle.

Image

[publish.log](https://github.com/user-attachments/files/20423095/publish.log)
[sonaUpload.log](https://github.com/user-attachments/files/20423096/sonaUpload.log)

## expectation
publish should work.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/8137

## Problem
Central Portal no longer supports the legacy sbt plugin layout.

## Solution
Make the default more friendly by setting sbtPluginPublishLegacyMavenStyle to false.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,7 +11,7 @@ import scala.util.Try
 // ThisBuild settings take lower precedence,
 // but can be shared across the multi projects.
 ThisBuild / version := {
-  val v = "1.11.0-SNAPSHOT"
+  val v = "1.11.1-SNAPSHOT"
   nightlyVersion.getOrElse(v)
 }
 ThisBuild / version2_13 := "2.0.0-SNAPSHOT"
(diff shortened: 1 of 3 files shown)


mkurz:
Given that from now on sbt plugins under `com.github.sbt` published to maven central can only be consumed by sbt 1.9.0 or newer, I guess from now on we can set the minimal `crossSbtVersions` to 1.9.0? (I did see some plugins setting the minimal sbt version to 1.5.0 or similar)

PR #8162, opened 2025-06-11
fix: semanticdb expects Wildcard imports to be in the last position

inglor:
semanticdb expects Wildcard imports to be in the last


eed3si9n (approved):
Thanks!

inglor:
Scala CLA signed.

PR #8165, opened 2025-06-17
[2.x] update dependencies

xuwei-k:


sbt-app/src/sbt-test/dependency-management/update-sbt-classifiers/build.sbt
```diff
@@ -44,14 +44,14 @@ lazy val root = (project in file("."))
...
         "org.apache.logging.log4j:log4j-slf4j-impl",
-        "org.checkerframework:checker-qual",
         "org.fusesource.jansi:jansi",
         "org.jline:jline-builtins",
         "org.jline:jline-native",
         "org.jline:jline-reader",
         "org.jline:jline-style",
         "org.jline:jline-terminal",
         "org.jline:jline-terminal-jni",
+        "org.jspecify:jspecify",
```
xuwei-k, line 54: https://github.com/ben-manes/caffeine/commit/2542ca8ca22fe4c30b62ee569ee71436fc44a37e

eed3si9n (approved):
Thanks!

project/Dependencies.scala
```diff
@@ -107,9 +107,9 @@ object Dependencies {
   val log4jSlf4jImpl = log4jModule("log4j-slf4j-impl")
   val log4jModules = Vector(log4jApi, log4jCore, log4jSlf4jImpl)
 
-  val caffeine = "com.github.ben-manes.caffeine" % "caffeine" % "2.8.5"
+  val caffeine = "com.github.ben-manes.caffeine" % "caffeine" % "3.2.1"
```
eed3si9n, line 110: Not sure why CI didn't catch this, but we can't use caffeine 3.x because we're still on JDK 8.

eed3si9n: I guess because we do not run CI on JDK 8 - https://github.com/sbt/sbt/pull/8169

PR #8167, opened 2025-06-19
fix: do not treat values "1.x" of "pluginCrossBuild/sbtBinaryVersion" as sbt 2 (fixes #8166)

### LINKED ISSUE #8166: maven style artefact name has incorrect format if an sbt plugin is cross-published for 1.x version different from 1.0
sbt version: 1.10.x or 1.11.2

# TLDR
`sbt.Classpaths.sbt2Plus` treats `sbtBinaryVersion = "1.3"` as sbt 2.
This leads to a wrong maven artefact name `sbt-structure-extractor-2025.2.0-javadoc.jar.sha1` 
instead of `sbt-structure-extractor_2.12_1.3-2025.2.0-javadoc.jar.sha1` 
and it fails publishing to Sonatype.
I will prepare a PR soon.

# Details
By default, cross-publishing of sbt plugin can work per one major version, e.g. 0.13 + 1.0.
However, in IntelliJ we have [sbt-structure](https://github.com/JetBrains/sbt-structure) plugin.
We have to publish it for more then 1 version of 1.x. 
In total for 0.13, 1.0, 1.3, 2.0
We need to do this because some of the features depend on newer sbt versions and on the same time we have to support legacy projects and we shouldn't fail on them.

We achieve that by:
1. Doing custom calculation of cross sbt version inside sbt-structure plugin ([here](https://github.com/JetBrains/sbt-structure/commit/92940f8b))
2. Doing the mirror custom calculation on the Scala Plugin side ([here](https://github.com/JetBrains/intellij-scala/commit/beacd7521770c79577b592bd8297cc485b28d4d1#diff-c6bdb71e5098893872e1a8a4d10a1dd7842e2b78cd1b8b73556d5d1e8280ad88R102)) 

I am in the process of moving publishing of the plugin to Sonatype that uses Maven style publishing.
The publishing fails and if you log into Sonatype you can see this in the logs:
```
File path 'org/jetbrains/scala/sbt-structure-extractor_2.12_1.3/2025.2.0' is not valid for file 'sbt-structure-extractor-2025.2.0-javadoc.jar.sha1'
```

It doesn't like that the file name is `sbt-structure-extractor-2025.2.0-javadoc.jar.sha1`,
it should be `sbt-structure-extractor_2.12_1.3-2025.2.0-javadoc.jar.sha1` (with the cross-version suffix)

I debugged it and found out that the root issue is in 
`sbt.Classpaths.mavenArtifactsOfSbtPlugin`

In particular, this code is wrong:
```scala
  private lazy val sbt2Plus: Def.Initialize[Boolean] = Def.setting {
    val sbtV = (pluginCrossBuild / sbtBinaryVersion).value
    sbtV != "1.0" && !sbtV.startsWith("0.")
  }
```
it treats `1.3` as sbt 2.
This is understandable given that our setup is far from standard and even might be the only in it's nature.

unkarjedy:
As a consequence, this fixes artifact name contraction for maven style and thus fixes publishing to Sonatype when a cross-built sbt version is different from 1.0

Fixes #8166

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2996,7 +2996,7 @@ object Classpaths {
   private lazy val packagedDefaultArtifacts = packaged(defaultArtifactTasks)
   private lazy val sbt2Plus: Def.Initialize[Boolean] = Def.setting {
     val sbtV = (pluginCrossBuild / sbtBinaryVersion).value
-    sbtV != "1.0" && !sbtV.startsWith("0.")
+    !sbtV.startsWith("1.") && !sbtV.startsWith("0.")
   }
   val jvmPublishSettings: Seq[Setting[_]] = Seq(
     artifacts := artifactDefs(defaultArtifactTasks).value,


eed3si9n (approved):
Thanks!

PR #8168, opened 2025-06-26
Support minus sign in project names

Duhemm:
Previously, sbt would crash when attempting to load a build where projects had minus signs (`-`) in their name.

For instance, when trying to load a project defined like this:

    lazy val `my-project` = project

After compilation, this definition looks somewhat like this

    val my$minusproject$lzy1 = ...

sbt was attempting to retrieve the original definition (without the $lzy suffix) by taking the mangled name up to the first `$`. Unfortunately, this approach does not work when the name includes special characters like a minus sign, because these will be prefixed with `$` as well. In the current example, sbt would then try to find the member named `my`, fail, and crash.

This patch fixes the issue by using the "underlying" name, which is the name without the additional information.


eed3si9n (approved):
Thanks @Duhemm!

PR #8170, opened 2025-06-30
[2.x] Implement worker + forked test

eed3si9n:
## Problem

See https://eed3si9n.com/rfc-4-persistent-worker/ for the general motivation. We want to eventually move out tests out of the sbt process, but we also don't want to lose JIT performance.

## Solution

This re-implements forked test on top of the Java worker CLI, using JSON-RPC over stdin/stdout as the communication mechanism instead of using the object stream on socket. For now, it's functionally same as sbt 1.x's forked test.

Later on, we can keep the worker process running, so we can keep the classloader warm.


xuwei-k:
- https://github.com/sbt/sbt/issues/8263

PR #8171, opened 2025-07-02
[1.x] Allow users to configure the timeout when publishing to the Maven Central repo

guizmaii:
The publishing of zio-aws takes a lot of time because it publishes a lot of libs and because of that fails to publish to the new Maven Central repo as the publishing process timeouts.
See https://github.com/zio/zio-aws/actions/runs/16018360517/job/45195274956

Previously, with the sbt-sonatype plugin, we were able to configure the `sonatypeTimeoutMillis` setting
(https://github.com/zio/zio-aws/pull/1517/files#diff-928a6eabff1094f1b64a2d1b6bc7574c5a483476ec3fee98c16d1fe46aae9d0dL101)

### DIFF
--- a/main-actions/src/main/scala/sbt/internal/sona/Sona.scala
+++ b/main-actions/src/main/scala/sbt/internal/sona/Sona.scala
@@ -36,13 +36,16 @@ class Sona(client: SonaClient) extends AutoCloseable {
   def close(): Unit = client.close()
 }
 
-class SonaClient(reqTransform: Request => Request) extends AutoCloseable {
+class SonaClient(reqTransform: Request => Request, requestTimeout: FiniteDuration)
+    extends AutoCloseable {
   import SonaClient.baseUrl
 
-  val gigahorseConfig = Gigahorse.config
-    .withRequestTimeout(2.minute)
-    .withReadTimeout(2.minute)
-  val http = Gigahorse.http(gigahorseConfig)
+  private val gigahorseConfig = Gigahorse.config
+    .withRequestTimeout(requestTimeout)
+    .withReadTimeout(requestTimeout)
+
+  private val http = Gigahorse.http(gigahorseConfig)
+
   def uploadBundle(
       bundleZipPath: Path,
       deploymentName: String,
@@ -66,7 +69,6 @@ class SonaClient(reqTransform: Request => Request) extends AutoCloseable {
             FormPart("bundle", bundleZipPath.toFile())
           )
         )
-        .withRequestTimeout(600.second)
       http.run(reqTransform(req), Gigahorse.asString)
     }
     awaitWithMessage(res, "uploading...", log)
@@ -155,7 +157,7 @@ class SonaClient(reqTransform: Request => Request) extends AutoCloseable {
         }.foreach(_ => loop(attempt + 1))
       } else ()
     loop(0)
-    Await.result(f, 600.seconds)
+    Await.result(f, requestTimeout + 5.seconds)
   }
 
   def close(): Unit = http.close()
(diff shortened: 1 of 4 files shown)


main-actions/src/main/scala/sbt/internal/sona/Sona.scala
```diff
@@ -66,7 +69,6 @@ class SonaClient(reqTransform: Request => Request) extends AutoCloseable {
             FormPart("bundle", bundleZipPath.toFile())
           )
         )
-        .withRequestTimeout(600.second)
```
guizmaii, line 69: Already configured in `http`

eed3si9n: Current implementation uses 2 minute timeout for status check type of request, and 10 minutes for the bundle upload. It seems like you are collapsing the two into one setting but with 2 minute timeout here. I feel like it might be better to pick a large enough value for upload timeout (60 min?) rather than make this configurable.

guizmaii: > Current implementation uses 2 minute timeout for status check type of request, and 10 minutes for the bundle upload. It seems like you are collapsing the two into one setting but with 2 minute timeout here

My bad. It's fixed in my latest changes

> I feel like it might be better to pick a large enough value for upload timeout (60 min?) rather than make this configurable.

I'm just scared that if it's not enough for someone, we'll need to make a new change and publish a new version of sbt, which can delay people's work. I think it'd be simpler to make it configurable so it can fit any need, like it was in sbt-sonatype

eed3si9n (approved):
overall LGTM, but for future reference, please split out non-semantic-change refactorings out to a separate pull request in the future for easier code review.

guizmaii:
Thanks @eed3si9n. Would it be possible for you to release a new version with this change, please?

eed3si9n:
https://eed3si9n.com/sbt-1.11.3 is released

guizmaii:
@eed3si9n I made this change in zio-aws: https://github.com/zio/zio-aws/pull/1554/files#diff-928a6eabff1094f1b64a2d1b6bc7574c5a483476ec3fee98c16d1fe46aae9d0dR104
Yet, publishing timed out after ~20minutes:
```scala
java.util.concurrent.TimeoutException: Futures timed out after [1215 seconds]
	at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:269)
	at scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:273)
	at scala.concurrent.Await$.$anonfun$result$1(package.scala:223)
	at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:57)
	at scala.concurrent.Await$.result(package.scala:146)
	at sbt.internal.sona.SonaClient.awaitWithMessage(Sona.scala:174)
	at sbt.internal.sona.SonaClient.uploadBundle(Sona.scala:86)
	at sbt.internal.sona.Sona.uploadBundle(Sona.scala:35)
	at sbt.internal.librarymanagement.Publishing$.sonatypeReleaseAction(Publishing.scala:57)
	at sbt.internal.librarymanagement.Publishing$.$anonfun$sonaRelease$1(Publishing.scala:24)
```
(See https://github.com/zio/zio-aws/actions/runs/16110990896/job/45458269962)

Would you have any why this didn't respect the `2.hours` values I've set? 🤔

eed3si9n:
I should've caught this during the code review but it seems like this PR added the setting at the project level, instead of the global level. So to override the value you have to rewire it at the root project level, instead of ThisBuild.

main-actions/src/main/scala/sbt/internal/sona/Sona.scala
```diff
@@ -108,16 +122,16 @@ class SonaClient(reqTransform: Request => Request) extends AutoCloseable {
...
   }
 
-  def deploymentStatus(deploymentId: String): PublisherStatus = {
-    val res = retryF(maxAttempt = 5) { (attempt: Int) =>
+  private def deploymentStatus(deploymentId: String): PublisherStatus = {
+    val res = retryF(maxAttempt = 5, waitDurationBetweenAttempt = 5.seconds) { (attempt: Int) =>
       deploymentStatusF(deploymentId)
     }
-    Await.result(res, 600.seconds)
+    Await.result(res, 10.minutes)
```
eed3si9n, line 129: Also there's still 10.minutes hardcoded here.

PR #8205, opened 2025-08-11
[2.x] fix: Support annotated definitions in build.sbt

Duhemm:
Previously, sbt would fail to load build.sbt files when they included annotated definitions because the parser would not correctly recognize those definitions as such. In sbt 1.x, this used to be fine, because there was little use for annotations in build.sbt.

Starting with sbt 2, whether caching should be enabled for a task key can be controlled via annotations on the task key definition. Because these can appear in build.sbt, support for annotations in build.sbt becomes more important.

This patch enhances parsing of build.sbt by keeping the parsed trees around so that the AST can be used to determine whether a given line represents a setting or a definition, rather than relying on string matching.


eed3si9n (approved):
Nice!

PR #8212, opened 2025-08-16
[1.x] add `Def.uncached` for sbt 2.x compat

xuwei-k:


main-settings/src/main/scala/sbt/Def.scala
```diff
@@ -26,6 +26,8 @@ object Def extends Init[Scope] with TaskMacroExtra with InitializeImplicits {
   type Classpath = Seq[Attributed[File]]
   type VirtualClasspath = Seq[Attributed[VirtualFile]]
 
+  def uncached[A1](a: A1): A1 = a
```
eed3si9n, line 29: This would require the minimum version of the plugin to be sbt 1.11.5.

xuwei-k:
https://github.com/sbt/sbt2-compat/blob/e321f9f96b6289d0416585538b37a6a1c3b66f31/src/main/scala-2.12/sbtcompat/PluginCompat.scala#L45-L51

PR #8257, opened 2025-09-01
Use reverse topological order for configurations when resolving internal dependencies #8249

### LINKED ISSUE #8249: Change the configurations order (that affects the classpath) for the internal dependencies tasks
**Motivation**: 
The `internalDependencyClasspath` task in sbt relies on the `trackInternalDependencies` setting, which is set to `TrackAlways` by default. This means that invoking this task can potentially trigger a compilation.

Because of this, in IntelliJ IDEA, we avoid using the `internalDependencyClasspath` task directly for acquiring inter-project dependencies. Instead, we use `sbt.internal.ClasspathImpl.interSort` utility method, which allows us to avoid triggering compilation with `exportedProducts`. Besides, for IntelliJ IDEA, we don’t really need the classpath for inter-project dependencies; we only need the information about which project depends on which and in what configurations.

In general this approach works well. However, since our support doesn’t rely directly on any sbt key, users lose the ability to customize the order of the project dependencies classpath in any specific way.

**Proposal**: 
Implement a new task that essentially performs the same function as `internalDependencyClasspath`, but without relying on `trackInternalDependencies` and without triggering compilation. This task could be named something like `internalDependencyClasspathNoTracking`.

^^ 
There shouldn't be a lot of implementation, and I'm willing to take this on. I just wanted to start a conversation to discuss whether it's fine to implement something like this, or if you might see a better workaround for the issue presented.

azdrojowa123:
Fix #8249

### DIFF
--- a/internal/util-collection/src/main/scala/sbt/internal/util/Dag.scala
+++ b/internal/util-collection/src/main/scala/sbt/internal/util/Dag.scala
@@ -17,6 +17,15 @@ object Dag {
   import scala.collection.{ mutable, JavaConverters }
   import JavaConverters.asScalaSetConverter
 
+  /**
+   * Returns a reverse topological ordering of the graph rooted at `root`.
+   * In this ordering, each node appears before all of its ancestors (i.e., children are listed before their parents).
+   *
+   * @see [[https://github.com/sbt/sbt/issues/8249]]
+   */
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #8260, opened 2025-09-03
[2.x] fix: Fixes removeN operator --=

### LINKED ISSUE #8255: [2.x] `value can only be used within a task or setting macro` error
### `Foo.scala`

```scala
import sbt._
import sbt.Keys._

object Foo {
  val unusedWarnings: Def.Initialize[Seq[String]] = Def.setting(
    scalaBinaryVersion.value match {
      case "2.12" =>
        Seq("-Ywarn-unused:imports")
      case _ =>
        Seq("-Wunused:imports")
    }
  )
  
  val mySettings = Def.settings(
    Compile / console / scalacOptions --= unusedWarnings.value
  )
}
```

### `build.sbt`

```scala
enablePlugins(SbtPlugin)
```

### `project/build.properties`

```properties
sbt.version=2.0.0-RC3
```

run `sbt compile`

## problem

```
[error]  15 |    Compile / console / scalacOptions --= unusedWarnings.value
[error]     |                                          ^^^^^^^^^^^^^^^^^^^^
[error]     |`value` can only be used within a task or setting macro, such as :=, +=, ++=, Def.task, or Def.setting.
[error]     |---------------------------------------------------------------------------
[error]     |Inline stack trace
[error]     |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[error]     |This location contains code that was inlined from Def.scala:378
[error]      ---------------------------------------------------------------------------
[error] one error found
```

## notes

sbt 1.x does not report error

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/8255

## Problem
removeN operator doesn't work with settings because the param isn't inlined.

## Solution
Fix it.


xuwei-k (approved):

xuwei-k:
maybe `-=` has same issue

```diff
-  final inline def -=[A2](v: A2)(using Remove.Value[A1, A2]): Setting[Task[A1]] =
+  final inline def -=[A2](inline v: A2)(using Remove.Value[A1, A2]): Setting[Task[A1]] =
```

eed3si9n:
Addressed in e0013078b56a09e69e64b3bb637b6e20098978c1

PR #8284, opened 2025-09-17
add JDK 25 CI instead of 21

xuwei-k:


eed3si9n (approved):

mzuehlke:
Should we make a similar change to the 1.11.x branch ?

PR #8296, opened 2025-09-23
enable some scripted tests

xuwei-k:


sbt-app/src/sbt-test/nio/input-filters/build.sbt
```diff
@@ -1,4 +1,6 @@
 import java.nio.file.{ Files, Path }
+import sbt.internal.FileChangesMacro.inputFiles
+import sbt.internal.FileChangesMacro.outputFiles
```
xuwei-k, line 3: need imports due to

https://github.com/sbt/sbt/blob/9713b7c1120cef9aa1a15de4370532db90f18e8e/sbt-app/src/main/scala/package.scala#L51

https://github.com/sbt/sbt/blob/9713b7c1120cef9aa1a15de4370532db90f18e8e/main/src/main/scala/sbt/internal/FileChangesMacro.scala#L29

🤔

eed3si9n (approved):

PR #8301, opened 2025-09-25
Use "java.home" to launch the BSP server even when using the sbt script #8300

### LINKED ISSUE #8300: SBT ignores JDK used in launcher in --bsp mode
Ensure you have a globally-installed SBT (for example via homebrew or anything else)
(I have macOS and mine is installed at `/opt/homebrew/bin/sbt`)

Image

Ensure that you have two JDK installed (e.g. JDK 11 and JDK 17)
Ensure that the default OS one is JDK 17 (it's included in PATH and JAVA_HOME also points to it).
Image

Now open terminal and export these variables:
```
export JDK_HOME_11=/Users/dmitrii.naumenko/Library/Java/JavaVirtualMachines/corretto-11.0.25/Contents/Home
export SBT_SCRIPT=/opt/homebrew/bin/sbt
export SBT_LAUNCH=/opt/homebrew/Cellar/sbt/1.11.0/libexec/bin/sbt-launch.jar
```
Patch the values according to your environment. 
On macOS you could use:
1. `/usr/libexec/java_home -v 11`
2. `where sbt`
3. `find /usr/local /opt/homebrew -name "sbt-launch.jar" 2>/dev/null` (or search it somewhere else =))

Navigate to some sbt project.
For example, create a new empty sbt project.
```shell
sbt new scala/scala3.g8 --name="myProject"
cd myProject
```

Now run the Boot app:
```shell
$JDK_HOME_11/bin/java -Xms100m -Xmx100m -classpath $SBT_LAUNCH -Dsbt.script=$SBT_SCRIPT xsbt.boot.Boot
```

Notice that it will use the JDK version that we used to run the Boot app (JDK 11)
Image

Now run the Boot app with the `--bsp` prameterL
```shell
$JDK_HOME_11/bin/java -Xms100m -Xmx100m -classpath $SBT_LAUNCH -Dsbt.script=$SBT_SCRIPT xsbt.boot.Boot --bsp
```

Notice that it now ignores the JDK 11 that we used and just uses the global one (JDK 17)
Image

-----------------

I noticed this issue while playing around with BSP in IntelliJ IDEA + Scala Plugin.
I will save the details for now.
I will just mention that it can cause strange, hard-to detect random issues while working with BSP mode with SBT.

I spent some time to debug  it and I think I now have an idea where this should be fixed.
I will create a PR soon.

unkarjedy:
Fixes #8300

image

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -361,6 +361,8 @@ class NetworkClient(
             term.isSupershellEnabled
           ).mkString(",")
 
+        // note, java home is generally not expected to be empty, but could be in some unexpected misconfigured cases
+        val javaHomeOpt = Option(Properties.javaHome).filter(_.nonEmpty)
         val cmd = arguments.sbtLaunchJar match {
           case Some(lj) =>
             if (log) {
@@ -375,7 +377,7 @@ class NetworkClient(
                 s"either upgrade $sbtScript to its latest version or make sure it is accessible from $$PATH, and run 'sbt bspConfig'"
               )
             }
-            val java = Option(Properties.javaHome)
+            val java = javaHomeOpt
               .map { javaHome =>
                 s"$javaHome/bin/java"
               }
@@ -383,7 +385,12 @@ class NetworkClient(
             List(java) ++ arguments.sbtArguments ++
               List("-jar", lj, DashDashDetachStdio, DashDashServer)
           case _ =>
-            List(arguments.sbtScript) ++ arguments.sbtArguments ++
+            // From sbt/sbt.sh/sbt.bat:
+            //    --java-home <path>         alternate JAVA_HOME
+            val javaHomeArgs = javaHomeOpt.toSeq.flatMap(javaHome => Seq("--java-home", javaHome))
+            List(arguments.sbtScript) ++
+              javaHomeArgs ++
+              arguments.sbtArguments ++
               List(DashDashDetachStdio, DashDashServer)
         }
 
--- a/.github/workflows/client-test.yml
+++ b/.github/workflows/client-test.yml
@@ -39,7 +39,7 @@ jobs:
       with:
         sbt-runner-version: 1.11.5
     - name: Set up Python 3.12
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: 3.12
     - name: Setup Windows C++ toolchain


main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -375,15 +377,20 @@ class NetworkClient(
...
                 s"$javaHome/bin/java"
               }
               .getOrElse("java")
             List(java) ++ arguments.sbtArguments ++
               List("-jar", lj, DashDashDetachStdio, DashDashServer)
           case _ =>
-            List(arguments.sbtScript) ++ arguments.sbtArguments ++
+            // From sbt/sbt.sh/sbt.bat:
+            //    --java-home <path>         alternate JAVA_HOME
+            val javaHomeArgs = javaHomeOpt.toSeq.flatMap(javaHome => Seq("--java-home", javaHome))
```
eed3si9n, line 390: Note that `java.home` system properties, even though it is named as such, isn't the same as `JAVA_HOME`.
See https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html
> `"java.home"`: Installation directory for Java Runtime Environment (JRE)

This means that `java` would be on the path, but if I recall correctly, `javac` will not be on the path, which could break Java compilation.

unkarjedy: Nice catch.
It slipped my attention, partially because I already saw the approved change that included the "java.home".
Though those two branches are different...

While I was looking into some issues related to BSP and IntelliJ I noticed that current state of BSP lacks a more transparent JDK management and depends on global state too much.

It seems there is no way to explicitly specify which JDK should be used in the `.bsp/sbt.json` file.
If you pass `-bsp` you can't pass the `--java-home` parameter after it.
The `sbt.internal.client.NetworkClient.parseArgs` will fail to parse it as `sbtArgument` passed to the script.
Sbt script expects `--java-home <path>` (separated with a space) and `NetworkClient.parseArgs` doesn't work with the space and treats the `<path>` as a sbt command.
We could try fixing this in a more general case (support for space " ") or just try to add 1 dedicated parameter for JDK path (~`--java-home` in sbt script, ~ `JAVA_HOME` env var).

If there was a way to specify the JDK more trasnparently, then we would also need to patch the 
`sbt.internal.bsp.BuildServerConnection.writeConnectionFile`.

@eed3si9n WDYT? Does it make sense?

eed3si9n: We are passing in the script location via system properties, like

```
"-Dsbt.script=/Users/xxx/.sdkman/candidates/sbt/current/bin/sbt"
```

so we could pass in `JAVA_HOME` as `-Dsbt.java.home` and use that when present.

Issue #7634, opened 2024-09-01
JEP 472: `WARNING: java.lang.System::load has been called by org.scalasbt.ipcsocket.NativeLoader` since JDK 24


xuwei-k:
## steps

just launch sbt 1.10.1 with latest JDK 24

```
$ java --version
openjdk 24-ea 2025-03-18
OpenJDK Runtime Environment (build 24-ea+13-1421)
OpenJDK 64-Bit Server VM (build 24-ea+13-1421, mixed mode, sharing)
```

## problem

```
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::load has been called by org.scalasbt.ipcsocket.NativeLoader in an unnamed module (file:/Users/kenji/.sbt/boot/scala-2.12.19/org.scala-sbt/sbt/1.10.1/ipcsocket-1.6.2.jar)
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled
```

## notes

- https://openjdk.org/jeps/472
- https://github.com/sbt/ipcsocket/blob/ee3439303e592846168a3b18f23358d3c6665949/src/main/java/org/scalasbt/ipcsocket/NativeLoader.java#L97
- https://github.com/openjdk/jdk/commit/20d8f58c92009a46dfb91b951e7d87b4cb8e8b41#diff-bd92d760986b9249dd3c02cc147db4f7e9dbbef90afef4971ee497a50e48c740R2029

eed3si9n:
https://openjdk.org/jeps/380 says it's delivered so technically we no longer need JNI for modern JDKs but not sure how it works with GraalVM.

PR #8305, opened 2025-09-28
use `java.net.UnixDomainSocketAddress` in BootServerSocket

xuwei-k:
- https://github.com/sbt/sbt/issues/7634

avoid `java.lang.System::load has been called by org.scalasbt.ipcsocket.NativeLoader ` warning if JDK `UnixDomainSocketAddress` available


eed3si9n:
Thank you! Since this is a common issue with https://github.com/sbt/ipcsocket/, maybe the glue code should exist in. What do you think?

xuwei-k:
- https://github.com/sbt/ipcsocket/pull/53

PR #8339, opened 2025-10-21
[2.x] Update BootServerSocket to use newServerSocketChannel(...)

eed3si9n:
Ref https://github.com/sbt/sbt/pull/8305

Testing the JDK 17 Unix Domain Socket.

### DIFF
--- a/main-command/src/main/java/sbt/internal/BootServerSocket.java
+++ b/main-command/src/main/java/sbt/internal/BootServerSocket.java
@@ -18,6 +18,9 @@ import java.net.Socket;
 import java.net.ServerSocket;
 import java.net.SocketException;
 import java.net.SocketTimeoutException;
+import java.nio.ByteBuffer;
+import java.nio.channels.ServerSocketChannel;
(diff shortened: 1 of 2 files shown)


main-command/src/main/java/sbt/internal/BootServerSocket.java
```diff
@@ -268,17 +277,15 @@ public OutputStream outputStream() {
...
-          serverSocket.setSoTimeout(5000);
-          while (running.get()) {
-            try {
-              ClientSocket clientSocket = new ClientSocket(serverSocket.accept());
-            } catch (final SocketTimeoutException e) {
-            } catch (final IOException e) {
-              running.set(false);
-            }
+        // TODO
+        // serverChannel.setSoTimeout(5000);
```
eed3si9n, line 281: @xuwei-k Likely we need to implement a mechanism to timeout in 5s using [Selector](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/Selector.html).

PR #8346, opened 2025-10-30
[1.x] Update scala213 to 2.13.17

danicheg:
Initially, I wanted to update the `semanticdb-scalac` version, as I noticed that `4.9.9` isn't published for `2.13.17`, and if `semanticdbVersion` isn't overridden in the project, one will encounter a `semanticdb-scalac_2.13.17:4.9.9` download error. However, `semanticdbVersion` was already updated in https://github.com/sbt/sbt/pull/8342. Anyway, I think it's better to keep `semanticdbVersion` and `scala213` updated together.


eed3si9n (approved):
Thanks!

PR #8359, opened 2025-11-08
[1.x] Use internal config for dependency tree (for Provided)

eed3si9n:
See https://github.com/sbt/sbt/discussions/8354

## Problem
dependency-tree currently doesn't include `Provided` configuration, because technically it isn't part of the `Compile` configuration, however, `Provided` does become part of the `Compile` classpath during compilation.

## Solution
This changes the implementation to use the internal configuration, e.g. `CompileInternal`, which includes the auxiliary configurations.

There's a tradeoff here since it means that we lose the tree view of the platonic `Compile` configuration.


eed3si9n:
I'm going to land this for 1.12.x series. We might revert it if it's causes troubles later.

Issue #8360, opened 2025-11-08
[2.x] perf: Regression of no-op compilation


eed3si9n:
_Originally posted by @mrdziuban in [#7658](https://github.com/sbt/sbt/issues/7658#issuecomment-3444264943)_

## Steps

My build at work contains 28 projects, which is likely larger than most builds, but a lot smaller than the 84 projects in [lila](https://github.com/lichess-org/lila) etc.

```
clean; Test/compile
Test/compile
Test/compile
Test/compile
Test/compile
Test/compile
```

## Problem

With sbt 2.0.0-RC6, the same test showed much worse performance:

```
[success] elapsed time: 19 s, cache 99%, 961 disk cache hits, 9 onsite tasks
[success] elapsed time: 17 s, cache 99%, 964 disk cache hits, 6 onsite tasks
[success] elapsed time: 14 s, cache 99%, 964 disk cache hits, 6 onsite tasks
[success] elapsed time: 14 s, cache 99%, 964 disk cache hits, 6 onsite tasks
[success] elapsed time: 13 s, cache 99%, 964 disk cache hits, 6 onsite tasks
```

## Note

With sbt 1.11.7:
 
```
[success] Total time: 3 s, completed Oct 24, 2025, 1:31:01 PM
[success] Total time: 2 s, completed Oct 24, 2025, 1:31:03 PM
[success] Total time: 1 s, completed Oct 24, 2025, 1:31:04 PM
[success] Total time: 1 s, completed Oct 24, 2025, 1:31:05 PM
[success] Total time: 1 s, completed Oct 24, 2025, 1:31:05 PM
```

eed3si9n:
@mrdziuban in commented in https://github.com/sbt/sbt/issues/7658#issuecomment-3444956837

> A couple initial observations:

1. It looks like there's a ~2 second gap between when sbt finishes loading the build and when it starts doing anything else
    1. The last trace that deals with a project named `*-build` ends at `00:00:02.100` and the first trace that deals with a project defined in `build.sbt` starts at `00:00:04.694`
2. `update`, `products`, and `compileIncSetup` tasks then take the longest time
    1. The longest `update` task took 1s 387ms
    2. The longest `products` task took 4s 235ms
    3. The longest `compileIncSetup` tasks took 1s 141ms
3. The actual `compileIncremental` tasks weren't too bad -- the longest took 530ms

[...]

PR #8363, opened 2025-11-10
[2.x] perf: Cache content hash of binary files

eed3si9n:
Ref https://github.com/sbt/sbt/issues/8360

## Problem
sha256 hashing is currently a bottleneck for no-op compilation.

## Solution
This adds a local, in-memory cache of sha256 hashes of binary files using their timestamp and file size.

### 2.0.0-RC6

Using sttp `Test/compile` on 100+ subprojects takes approx 8±1s <s>34±2s</s> after two warm-up runs.

### After

After this change, `Test/compile` on 100+ subprojects takes approx 4±1s <s>24±5s</s> after two warm-up runs. Flamegraph also confirms that sha256 is no longer a bottleneck.

### Changing the digest cache size

I'm picking 1MB as the default cache size, but here are some tests ranging from 1 byte to 100MB. Note x-axis is in log-scale.
Cache size (bytes) vs Time (s)

### DIFF
--- a/main-command/src/main/scala/sbt/BasicKeys.scala
+++ b/main-command/src/main/scala/sbt/BasicKeys.scala
@@ -123,6 +123,12 @@ object BasicKeys {
     10000
   )
 
+  val localDigestCacheByteSize = AttributeKey[Long](
+    "localDigestCacheByteSize",
+    "The maximum total size in the in-memory digest cache in bytes.",
+    10000
+  )
+
   // Unlike other BasicKeys, this is not used directly as a setting key,
   // and severLog / logLevel is used instead.
   private[sbt] val serverLogLevel =
--- a/util-cache/src/main/scala/sbt/util/CacheImplicits.scala
+++ b/util-cache/src/main/scala/sbt/util/CacheImplicits.scala
@@ -8,7 +8,72 @@
 
 package sbt.util
 
+import com.github.benmanes.caffeine.cache.{ Cache as CCache, Caffeine, Weigher }
+import java.nio.file.{ Files, NoSuchFileException }
+import java.nio.file.attribute.BasicFileAttributes
+import java.util.concurrent.atomic.{ AtomicLong, AtomicReference }
 import sjsonnew.BasicJsonProtocol
+import xsbti.{ HashedVirtualFileRef, PathBasedFile }
 
-object CacheImplicits extends CacheImplicits
-trait CacheImplicits extends BasicCacheImplicits with BasicJsonProtocol
+object CacheImplicits extends CacheImplicits:
+  private[sbt] val defaultLocalDigestCacheByteSize = 1024L * 1024L
+end CacheImplicits
+
+trait CacheImplicits extends BasicCacheImplicits with BasicJsonProtocol:
+  private val localDigestCacheByteSize = AtomicLong(CacheImplicits.defaultLocalDigestCacheByteSize)
+  private val weigher: Weigher[String, (String, Long, Long)] = { case (k, (v1, _, _)) =>
+    k.size + v1.size + 16
+  }
+
+  private val stampCache: AtomicReference[CCache[String, (String, Long, Long)]] =
+    AtomicReference(
+      Caffeine
+        .newBuilder()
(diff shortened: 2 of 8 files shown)


util-cache/src/main/scala/sbt/util/CacheImplicits.scala
```diff
@@ -8,7 +8,47 @@
...
+import java.nio.file.{ Files, NoSuchFileException }
+import java.nio.file.attribute.BasicFileAttributes
 import sjsonnew.BasicJsonProtocol
+import xsbti.{ HashedVirtualFileRef, PathBasedFile }
 
 object CacheImplicits extends CacheImplicits
-trait CacheImplicits extends BasicCacheImplicits with BasicJsonProtocol
+trait CacheImplicits extends BasicCacheImplicits with BasicJsonProtocol:
+
+  private val stampCache: CCache[(String, Long, Long), String] = Caffeine
```
retronym: Maybe store the time and size in the value of this cache, so you only retain the latest entry per file?
```
Option(stampCache.getIfPresent(ref.id)) match {
   case Some(entry) =>
     if (entry.timestamp == ... && entry.size == ...) { // use
     else { // put
     case _ => // put
     }
```

eed3si9n: Yea that's a good idea to avoid unnecessary GC.

eed3si9n: Addressed in https://github.com/sbt/sbt/pull/8363/commits/825ab1bf7ae2e2bc181e8629c921afc3cb593a83

util-cache/src/main/scala/sbt/util/CacheImplicits.scala
```diff
@@ -8,7 +8,47 @@
...
 import sjsonnew.BasicJsonProtocol
+import xsbti.{ HashedVirtualFileRef, PathBasedFile }
 
 object CacheImplicits extends CacheImplicits
-trait CacheImplicits extends BasicCacheImplicits with BasicJsonProtocol
+trait CacheImplicits extends BasicCacheImplicits with BasicJsonProtocol:
+
+  private val stampCache: CCache[String, (String, Long, Long)] = Caffeine
+    .newBuilder()
+    .maximumSize(50000)
```
BillyAutrey: Should this size be configurable? I could see repositories of a sufficient size blowing through this limit.

Obviously it's a tradeoff, but I figure people will want to own the tradeoff with memory consumption.

eed3si9n: I can try to make this configurable via a global setting.

main-command/src/main/scala/sbt/BasicKeys.scala
```diff
@@ -123,6 +123,12 @@ object BasicKeys {
     10000
   )
 
+  val localDigestCacheByteSize = AttributeKey[Long](
+    "localDigestCacheByteSize",
+    "The maximum total size in the in-memory digest cache in bytes.",
+    10000
```
filipwiech, line 129: @eed3si9n In the PR description and the commit message it's said that the default cache limit is set to 1MB, but looks like it's actually ~10KB. Is that intentional? 🤔

eed3si9n: The third number of `AttributeKey.apply` is `rank`, which is used to show/hide settings in help command. The actual initialization is done in RemoteCache.scala as:

```scala
localDigestCacheByteSize :== CacheImplicits.defaultLocalDigestCacheByteSize,
```

which in turn is set to

```scala
  private[sbt] val defaultLocalDigestCacheByteSize = 1024L * 1024L
```

filipwiech: Ooh, now it's clear, thanks and sorry for bothering you, somehow I thought that's the default value. I guess pattern recognition kicked-in incorrectly. 😅

eed3si9n: No worries. Please continue to provide ad-hoc code reviews.

PR #8370, opened 2025-11-14
[2.x] perf: Synchronize the classes using Analysis stamps 

eed3si9n:
## Problem
We currently need to split the output of the compiler from
the classes directory so we can cache just the output
without mixing it with the resource files
while maintaining similar classes directory as sbt 1.x
(as opposed to just splitting them completely).

However, copying directory is a performance bottleneck.

## Solution
This implements a new synchronizing function that uses
the stamp information generated during the incremental compilation.

For example, in a no-op compilation, the stamps will remain the same,
and no files will be copied.
If only a few classes are changed, those stamps will change,
and only the changed files will be deleted or copied.

### DIFF
--- a/main-actions/src/main/scala/sbt/Sync.scala
+++ b/main-actions/src/main/scala/sbt/Sync.scala
@@ -9,14 +9,16 @@
 package sbt
 
 import java.io.{ File, IOException }
+import java.nio.file.Path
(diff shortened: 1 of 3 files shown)


eed3si9n:
This doesn't work as-is since there might be non-class files generated by the compiler and compiler plugins.

Issue #8371, opened 2025-11-14
BSP requests are hanging when `Global / onChangedBuildSource := ReloadOnSourceChanges` (partially related to sbt 2.x )


azdrojowa123:
## steps
The steps to reproduce are described in this ticket https://youtrack.jetbrains.com/issue/SCL-24324/BSP-project-import-takes-too-long-with-sbt-2-when-build.sbt-is-corrupted

## problem
This problem was discovered while experimenting with sbt 2, where `ReloadOnSourceChanges` is enabled by default. However, it is also visible in previous sbt versions where `Global / onChangedBuildSource := ReloadOnSourceChanges` is explicitly declared.  

Based on my debugging, when there is a change in the sources and a BSP request is sent, the "reload" command is added before the request (`sbt.MainLoop.processCommand`). In simple terms, the reload consists of the `load project` command and the `load failed` command.  

If the project load fails for any reason, it jumps to the `load failed` command (`sbt.BuiltinCommands.doLoadFailed`), which waits for input from `System.in`. This doesn't work with network clients because the `System.in` is not attached there. In general, it's causing the sent BSP request to "hang."

## solutions
I've tried to create a solution for this problem, and it's available [here](https://github.com/azdrojowa123/sbt/commit/9b44a6b0c6349c5c3e8c9ba7c2b07208f0e3b1c4) (it's still WIP). However, I'm not sure if it's entirely correct and covers all cases. For example, I assumed that the additional reload can only happen if there's an Exec source. However, in some cases it's possible to have Exec without a source (though I'm not sure if these cases are important in this context).

I'd be grateful for any tips in what direction the fix for this problem should go. I'm willing to work on it.

Thanks.

eed3si9n:
Hello, thanks for the report and the analysis of the situation. A similar issue was also reported recently (https://github.com/sbt/sbt/issues/8344), which says that the IntelliJ Scala plugin seemingly hangs with an error in `build.sbt`.

Maybe in the earlier days of sbt where the only usage was via `sbt` waiting for "(r)etry" had some value, but given CI usage, auto reload, sbtn, and IDE integration, I am sort of wondering if we should get rid of "(r)etry" blocking for stdin, especially if that's happening in the headless server process.

PR #8375, opened 2025-11-17
Trigger only load (without loadFailed) in the case of NetworkChannel

azdrojowa123:
https://github.com/sbt/sbt/issues/8371

### DIFF
--- a/main/src/main/scala/sbt/MainLoop.scala
+++ b/main/src/main/scala/sbt/MainLoop.scala
@@ -10,6 +10,7 @@ package sbt
 
 import sbt.BasicCommandStrings.{ StashOnFailure, networkExecPrefix }
 import sbt.ProjectExtra.extract
+import sbt.internal.{ ConsoleChannel, FastTrackCommands, ShutdownHooks, SysProp, TaskProgress }
 import sbt.internal.langserver.ErrorCodes
 import sbt.internal.nio.CheckBuildSources.CheckBuildSourcesKey
 import sbt.internal.protocol.JsonRpcResponseError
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/MainLoop.scala
```diff
@@ -301,10 +302,22 @@ private[sbt] object MainLoop:
           .remove(Keys.terminalKey)
           .remove(Keys.currentCommandProgress)
       }
+
+      val channel = channelName.flatMap(exchange.channelForName)
+      val (canReload, useLoadp) = channel match
+        case Some(nc: NetworkChannel) =>
+          val isInteractive = nc.isInteractive
```
azdrojowa123: I assumed that when it's a `NetworkClient` with a virtual terminal attached, it can respond, so a full "reload" can be called.  
Though I haven't tested this case, and I'm not sure whether something like this is even practiced.

eed3si9n (approved):
Thanks!

main/src/main/scala/sbt/MainLoop.scala
eed3si9n: sbtn is capable of reloading.

eed3si9n: But I don't think it can currently recover from an error in `build.sbt`. If we make an intentional error, it currently gets stuck I think. If we want to maintain retry prompt, that probably should be done on the sbtn-side.

azdrojowa123: So, do you recommend removing the `isInteractive` condition and just performing `loadp` instead of `reload` for all network channels? Or should it stay as it is?

eed3si9n: Yea loadp for all network channel would be safer I think.

PR #8379, opened 2025-11-23
[2.x] Adds cleanFull command

### LINKED ISSUE #7658: [2.x] Cache deactivation on task
I imagine it can be very frustrating for the user if the cache does not work as expected (inconsistencies, corrupted output, performance degradation). I think we should always be able to deactivate the cache globally, on a project, at the task level.

1. `Global / cacheStores := Seq.empty`

If `cacheStores` is empty I think the user would expect no cache at all, even not a local cache. We should make sure that sbt still works if there is no cache.

2. `compileIncremental / cachingEnabled := false`

At different level, it should be possible to deactivate the cache. This should be clearly documented.

I think the logic can be generated automatically by `Def.cachedTask`: we add a `taskScope` parameter, to generate the following automatically:

```scala
if (taskScope /cachingEnabled).value then ActionCache.get(...)(doAction) else doAction()
```

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/7658

## Problem
We might want to clear the disk cache.

## Solution
This adds `cleanFull` command.

### DIFF
--- a/main-command/src/main/scala/sbt/BasicCommandStrings.scala
+++ b/main-command/src/main/scala/sbt/BasicCommandStrings.scala
@@ -252,7 +252,10 @@ $AliasCommand name=
   def continuousBriefHelp: (String, String) =
     (ContinuousExecutePrefix + " <command>", continuousDetail)
   def ClearCaches: String = "clearCaches"
-  def ClearCachesDetailed: String = "Clears all of sbt's internal caches."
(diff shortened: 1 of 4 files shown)


SethTisue:
maybe `cleanFull`, by analogy with `testFull`?

eed3si9n:
Sure, I can change to `cleanFull`.

Issue #8386, opened 2025-11-24
sbt should not complain if the `scala3_library` is missing


hamzaremmal:
## steps

In this PR https://github.com/scala/scala3/pull/24493, we remove the old `jar` files for `scala3_library` and start publishing it as a `pom` module.

## problems

Currently, sbt will complain that the `scala3_library` jar is missing.

```scala
[info] set current project to scala-stm (in build file:/home/runner/work/scala3/scala3/community-build/community-projects/scala-stm/)
[error] java.lang.RuntimeException: Missing scala3-library jar file
[error] 	at scala.sys.package$.error(package.scala:30)
[error] 	at sbt.Defaults$.$anonfun$scalaInstanceFromUpdate$22(Defaults.scala:1258)
[error] 	at scala.Option.getOrElse(Option.scala:189)
[error] 	at sbt.Defaults$.file$1(Defaults.scala:1258)
[error] 	at sbt.Defaults$.$anonfun$scalaInstanceFromUpdate$29(Defaults.scala:1269)
[error] 	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:286)
[error] 	at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:36)
[error] 	at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:33)
[error] 	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:198)
[error] 	at scala.collection.TraversableLike.map(TraversableLike.scala:286)
[error] 	at scala.collection.TraversableLike.map$(TraversableLike.scala:279)
[error] 	at scala.collection.mutable.ArrayOps$ofRef.map(ArrayOps.scala:198)
[error] 	at sbt.Defaults$.$anonfun$scalaInstanceFromUpdate$1(Defaults.scala:1269)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
[error] 	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:63)
[error] 	at sbt.std.Transform$$anon$4.work(Transform.scala:69)
[error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:283)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:24)
[error] 	at sbt.Execute.work(Execute.scala:292)
[...]

PR #8387, opened 2025-11-24
[1.x] do not assume scala3_library to be a jar

hamzaremmal:
Fixes https://github.com/sbt/sbt/issues/8386

In https://github.com/scala/scala3/pull/24493, the scala3-library module will only be a POM file. Before this change, sbt will complain that the `scala3_library` jar is missing.

### DIFF
--- a/main/src/main/scala/sbt/internal/Compiler.scala
+++ b/main/src/main/scala/sbt/internal/Compiler.scala
@@ -158,12 +158,11 @@ private[sbt] object Compiler {
           }
         }
       }
-      def file(id: String): File = {
-        val files = for {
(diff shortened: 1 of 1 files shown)


hamzaremmal:
@eed3si9n I'm not familiar with the processes in this repo but should I have a second PR against the `develop` branch for sbt 2 or is handled differently?

eed3si9n:
Once it lands on this branch, I can mass merge to develop with other commits.

hamzaremmal:
Should be good. Not sure what the failed test does but I don't think my change has anything to do with it.

eed3si9n (approved):
Thanks!

Issue #8389, opened 2025-11-26
When `loadFailed` is triggered and the subsequent task is also failing, sbt process is terminated


azdrojowa123:
## steps
I was testing with sbt 1.11.7.  

1. Have some sbt project with the `build.sbt` like:
```
val simpleTask =  taskKey[Unit]("Task that fails")

lazy val root = (project in file("."))
  .settings(
    simpleTask := {
      throw new MessageOnlyException("failure")
    },
  )
```
2. Start the sbt process in the terminal. 
3. Execute `simpleTask` 

**Expected & actual result**: the `simpleTask`  fails and sbt keeps running in the terminal 

4. Make any typo in the `build.sbt` e.g. remove `ilure")` text 
5. Execute in terminal `reload;simpleTask` 
6. When loading is performed, the `loadFailed` action is executed, and there is an error prompt: "Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)". Type `i` to ignore. After this the `simpleTask` is executed.

**Expected result**: `simpleTask` fails and sbt keeps running in the terminal 
**Actual result**: the sbt process is terminated 

## problem
From what I have debugged so far, the issue is with `StateOpsImpl#fail` and the `reload` command implementation. The `reload` command implementation puts `PopOnFailure` before `FailureWall`. When `loadp` failed, which is part of the `reload` command in `StateOpsImpl#fail`, all the remaining commands until `FailureWall` are removed, including `PopOnFailure`. Then, when `simpleTask` is executed and it jumps to `StateOpsImpl#fail` again, the `s.onFailure` is empty, and if it is empty, the exit is performed.

I haven't tested the fix yet, but it looks like it may be enough to just switch the order inside the reload command and put `PopOnFailure` after the `FailureWall`. The same order is implemented in `StateOpsImpl#process` in case of the network channel.

azdrojowa123:
I think this problem is even more important in the context of sbt 2, where `onChangedBuildSource := ReloadOnSourceChanges` is enabled by default.

In sbt 2, when I type `compile` (with expected errors) in the sbt terminal and there’s a buggy change in `build.sbt`, the entire sbt process is terminated if "ignore" is selected in the `loadFailed` action.

PR #8399, opened 2025-12-04
change the order of FailureWall & PopOnFailure in the reload command

azdrojowa123:
#8389


eed3si9n (approved):
@azdrojowa123 Thanks!

PR #8401, opened 2025-12-05
Use `-Werror` instead of `-Xfatal-warnings`

xuwei-k:
`-Xfatal-warnings` is deprecated since Scala 3.8

- https://github.com/scala/scala3/pull/24359


eed3si9n (approved):
@xuwei-k Thanks!

PR #8402, opened 2025-12-06
Fix #8345: forked test execution doesn't access main resources

### LINKED ISSUE #8345: [2.x] Thread context ClassLoader doesn't see main resources from test
### `build.sbt`

```scala
val scala3Version = "3.7.3"

lazy val root = project
  .in(file("."))
  .settings(
    name := "issue-pekko-config",
    version := "0.1.0-SNAPSHOT",
    scalaVersion := scala3Version,

    libraryDependencies += "org.scalatest" %% "scalatest" % "3.2.19" % Test,
    Test / fork := true,  // Important for issue reproduction
    // exportJars := false  // Does't help
  )
```

### `project/build.properties`

```
# sbt.version=1.11.7
sbt.version=2.0.0-RC6
```

### `src/main/resources/foo`

Empty file.

### `src/test/scala/IssueMinTest.scala`

```scala
import org.scalatest.funsuite.AnyFunSuite

class IssueMinTest extends AnyFunSuite:
  test("classLoaderRaw"):
    val loader = Thread.currentThread().getContextClassLoader()
    val resources = loader.getResource("foo")
    assert(resources != null)
```

## problem

```
sbt:issue-pekko-config> test
[info] compiling 1 Scala source to /Users/anatolii/Projects/STA/issue-pekko-config/target/out/jvm/scala-3.7.3/issue-pekko-config/test-backend ...
[info] IssueMinTest:
[info] - classLoaderRaw *** FAILED ***
[info]   null equaled null (IssueMinTest.scala:7)
[info] Run completed in 824 milliseconds.
[info] Total number of tests run: 1
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 0, failed 1, canceled 0, ignored 0, pending 0
[info] *** 1 TEST FAILED ***
```

## expectation

The behavior above is inconsistent with the behavior under sbt 1.11.7, where the test passes:

```
sbt:issue-pekko-config> test
[info] compiling 1 Scala source to /Users/anatolii/Projects/STA/issue-pekko-config/target/scala-3.7.3/test-classes ...
[info] IssueMinTest:
[info] - classLoaderRaw
[info] Run completed in 884 milliseconds.
[info] Total number of tests run: 1
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 1, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
```

## notes

- Possibly related: #8341, #8340.
- `exportJars := false` workaround for #8341 does not work for this issue.
- Does not reproduce without `Test / fork := true`.
- Affects lightbend/config: https://github.com/lightbend/config/blob/e1519d7b750f1b38a1c4c2505d3a08058477f8c7/config/src/main/java/com/typesafe/config/ConfigFactory.java#L145 -> https://github.com/lightbend/config/blob/e1519d7b750f1b38a1c4c2505d3a08058477f8c7/config/src/main/java/com/typesafe/config/impl/Parseable.java#L721
- Transitively affects libraries depending on lightbend/config, such as pekko: https://github.com/apache/pekko/blob/d0be463fa14135260b7d0ff103e315de4c9a1234/actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala#L348
- Blocks libraries depending on pekko in testing, such as ox: https://github.com/softwaremill/ox/blob/2c8d874ecb2cf1068293cbdceb10039ace5bf7ff/core/src/test/scala/ox/flow/reactive/FlowPublisherPekkoTest.scala#L24

anatoliykmetyuk:
Fixes issue where tests running in forked mode (Test / fork := true) cannot access resources from src/main/resources, while non-forked execution works correctly.

Root cause: Context classloader not set: tests using Thread.currentThread().getContextClassLoader() couldn't find resources because the context classloader wasn't set to the URLClassLoader with the classpath

Solution: Set context classloader in ForkTestMain.runTests() to the URLClassLoader containing the test classpath


worker/src/main/java/sbt/internal/worker1/ForkTestMain.java
```diff
@@ -325,6 +325,7 @@ private ExecutorService executorService(final boolean parallel) {
     }
 
     private void runTests(TestInfo info, ClassLoader classLoader) throws Exception {
+      Thread.currentThread().setContextClassLoader(classLoader);
```
eed3si9n, line 328: Nice work! Do you think we could test this via scripted test something like https://github.com/sbt/sbt/blob/develop/sbt-app/src/sbt-test/tests/fork2/build.sbt?

anatoliykmetyuk: Added a regression test.

eed3si9n (approved):
Thanks @anatoliykmetyuk!

PR #8412, opened 2025-12-11
Fix #8405 updateSbtClassifiers plugins output

### LINKED ISSUE #8405: sbt 2.x: `updateSbtClassifiers` doesn't include information about the used sbt plugins
## steps

Create 2 projects – one in sbt 1.11.7 and another in 2.0.0-RC7. 
Add `addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.13.1")` to `plugins.sbt` in each of these projects. 
Start the sbt process and run `show updateSbtClassifiers`. 

Sbt 1.11.7 - the returned report contains information about the build info plugin 
Sbt 2.0.0-RC7 - the returned report doesn't have information about the build info plugin 

Is it a bug or some intentional change in the sbt 2.x? 
Is there maybe another way to retrieve sbt sources in sbt 2.x?

anatoliykmetyuk:
`transitiveScratch` replaced the original direct dependencies (including plugins) with `report.allModules` from `lm.update()`. `report.allModules` only includes transitive dependencies, not the direct dependencies passed in, so plugins were dropped. The fix preserves the original direct dependencies and merges them with transitive-only dependencies from `report.allModules` (filtering out modules already present in the direct list), ensuring both plugins and transitive dependencies are included in the final `updateSbtClassifiers` output.


eed3si9n (approved):
Thanks!

PR #8413, opened 2025-12-12
[1.x] Handle JDK_JAVA_OPTIONS

sideeffffect:
/cc @velvetbaldmime @indoorvivants @keynmol 

It would be great if after releasing this in 1.11.x, we could forward-port it to 1.12.x and 2.x/develop. How can I help with that @eed3si9n ?

### DIFF
--- a/launcher-package/src/universal/bin/sbt.bat
+++ b/launcher-package/src/universal/bin/sbt.bat
@@ -718,6 +718,7 @@ if defined sbt_args_verbose (
   if defined _JAVA_OPTS ( call :echolist !_JAVA_OPTS! )
   if defined _SBT_OPTS ( call :echolist !_SBT_OPTS! )
   if defined JAVA_TOOL_OPTIONS ( call :echolist %JAVA_TOOL_OPTIONS% )
+  if defined JDK_JAVA_OPTIONS ( call :echolist %JDK_JAVA_OPTIONS% )
   echo -cp
   echo "!sbt_jar!"
   echo xsbt.boot.Boot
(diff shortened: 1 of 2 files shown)


eed3si9n:
> It would be great if after releasing this in 1.11.x, we could forward-port it to 1.12.x and 2.x/develop. How can I help with that @eed3si9n ?

Unless there's a security fix, I don't think I will release any more sbt 1.11.x, so I think we should target sbt 1.12.x.

eed3si9n:
CI is failing because librarymanagement repo no longer has develop branch - https://github.com/sbt/sbt/blob/6d62c345aa0289e1426c762b55dcfa25a6a853c2/.github/workflows/ci.yml#L62

eed3si9n (approved):

sideeffffect:
Thanks @eed3si9n , is there something I need to do to help this get forward-ported to `develop` and/or `series/2.x`?

Also, when can we expect sbt 1.12 to get released?

eed3si9n:
> Thanks @eed3si9n , is there something I need to do to help this get forward-ported to `develop` and/or `series/2.x`?

No, I just merge 1.12.x en masse occasionally.

> Also, when can we expect sbt 1.12 to get released?

The key change for sbt 1.12.0 is Scala 3.8.0 support, so I've been leaving it at milestone in case we might need to tweak it further. Per https://contributors.scala-lang.org/t/scala-3-8-0-release-thread/7291/23 

> Stable release for 3.8.0 is now planned for the first week of January.

so maybe sbt 1.12.0 can also go into release candidate soon.

PR #8415, opened 2025-12-14
Return build targets sources from `buildTarget/dependencySources` request

azdrojowa123:
It's basically the same as #8404, but for the 1.12 branch.


eed3si9n (approved):
Thanks!

PR #8423, opened 2025-12-18
[2.x] Fix `help` command arguments parsing

Duhemm:
Previously, some additional code was used to parse the arguments to the `help` command and ensure that the remaining commands were consistent with the desired state of sbt.

While this code worked well in sbt 1.x, and was helpful in getting `sbt help new` to work as expected, this does not work in sbt 2, and is actually not required.

This patch removes the additional code, which fixes the `help` command in sbt.

See #4237

### DIFF
--- a/main-command/src/main/scala/sbt/BasicCommands.scala
+++ b/main-command/src/main/scala/sbt/BasicCommands.scala
@@ -129,25 +129,11 @@ object BasicCommands {
   }
 
   def runHelp(s: State, h: Help)(arg: Option[String]): State = {
-
-    val (extraArgs, remainingCommands) = s.remainingCommands match {
-      /*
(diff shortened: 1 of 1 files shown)


Duhemm:
I believe the same snippet that I removed here exists in `TemplateCommandUtil.scala` as well, and it's broken there too, preventing `sbt new` from working as expected. I think the fix there will be different, since IIRC the templates can accept some arguments.

https://github.com/sbt/sbt/blob/26b5ee457f04c3c80d16fd82f16bb47c15ce7514/main/src/main/scala/sbt/TemplateCommandUtil.scala#L47-L51

eed3si9n (approved):
Thanks @Duhemm!

PR #8428, opened 2025-12-27
[2.x] Show warnings when scalaVersion is missing

### LINKED ISSUE #2795: Require scalaVersion
@tpolecat wrote https://twitter.com/tpolecat/status/789288986258931712:

> I think sbt should not start up unless you specify scalaVersion. Beginners think it uses the "installed" Scala which leads to suffering.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/2795

## Problem
There's a disconnect between what is perceived to be the current Scala version, and what sbt uses internally, and thus what it chooses to be the default scalaVersion.

## Solution
This displays a warning if scalaVersion setting is missing.


main/src/main/scala/sbt/internal/LintUnused.scala
```diff
@@ -165,6 +165,39 @@ object LintUnused {
...
+      mb = extracted.get(isMetaBuild.rescope(scope))
+      auto = extracted.get(autoScalaLibrary.rescope(scope))
+      (_, sk) = extracted.runTask(skip.rescope(scope.rescope(publish.key)), state)
+      display = p match
+        case ProjectRef(_, id) => id
+        case _ | null          => Reference.display(p)
+      c <- comp.get(data.map(_.definingKey).getOrElse(key.scopedKey))
+      setting <- c.settings.headOption
+    do
+      if auto && !isPlugin && !mb && !sk then
```
soc: Hey @eed3si9n, does this work for

```
sbt:colors> autoScalaLibrary
[info] true
sbt:colors> managedScalaInstance
[info] false
```

or would the warning be intended in that case?

eed3si9n: Good call. I can exempt `managedScalaInstance` as well.

eed3si9n: Addressed in 6cb786d010d8a5ff5ac7777a4e67a2284791ebb1

soc: Oh, thanks! I was not intending to imply something should be done about it, but I'm glad it was helpful!

(I long forgot the intricacies of `autoScalaLibrary` vs. `managedScalaInstance`, or reason I had it configured this way, so if you said "you are doing it wrong, fix your config" it would also have been fine. :-))

PR #8430, opened 2025-12-28
[2.x] Task value optimization

### LINKED ISSUE #8429: [2.x] ActionCache optimization
Originally reported by @azdrojowa123 on [Discord](https://discord.com/channels/632150470000902164/922600050989875282/1449522239916609657)

> I did a small test and ran a simple cached task. The symlink was added to the project directory, and then I removed the AC (once the whole directory and another time only the content of the file corresponding to my task). In both cases, in the next task execution, it was re-run, and the cache wasn't used at all (at least that's how it seems). 
So, it looks like even though there's a symlink, it still relies on (or reads? I'm not sure) the AC. Is this expected?

## Problem

On the operating systems that support symlinks we create a symlink to CAS content using the ActionCache file name in situation where task values are stored in CAS. This symlink apparently isn't being used.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/8429

This takes advantage of the symlink when it exists, and skips reading AC JSON.

### DIFF
--- a/util-cache/src/main/scala/sbt/util/ActionCache.scala
+++ b/util-cache/src/main/scala/sbt/util/ActionCache.scala
@@ -2,7 +2,7 @@ package sbt.util
 
 import java.io.File
 import java.nio.charset.StandardCharsets
-import java.nio.file.{ Path, Paths }
+import java.nio.file.{ Files, Path, Paths }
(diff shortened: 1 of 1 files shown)


eed3si9n:
Some tasks have captured side effects, so I guess just returning the value isn't enough.

PR #8443, opened 2026-01-07
avoid deprecated `scala.App`

xuwei-k:
https://github.com/scala/scala3/blob/7ddf329f6f216caee731fc277b83868bdb89d434/library/src/scala/App.scala#L65


eed3si9n (approved):
Thanks!

PR #8447, opened 2026-01-08
[2.x] fix: Skips interactive prompt in batch mode when project loading

### LINKED ISSUE #8344: "-batch" mode should remove "Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)" prompt
## steps

- write a type error in build.sbt
- test it with "sbt -batch test"
- the last line always contains "Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)" 

## problem

In batch mode this prompt should be omitted, as batch mode has no user interaction.

this prompt also caused IntelliJ Scala plugin to hang indefinitely, unable to proceed:

Image

## expectation

remove this line in batch mode

Francluob:
… fails

When project loading fails in batch mode, sbt was showing an interactive prompt asking the user to choose between retry, quit, last, or ignore. However, in batch mode there is no interactive terminal, causing the process to hang waiting for input that will never come.

This fix checks if we're in batch mode (Prompt.Batch) and automatically retries the project load (the default behavior) without prompting the user. This allows batch mode scripts to continue execution even when project loading encounters transient errors.

The interactive behavior remains unchanged for non-batch mode.

Closes #8344

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=11581451

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -869,25 +869,30 @@ object BuiltinCommands {
 
   @tailrec
   private def doLoadFailed(s: State, loadArg: String): State = {
-    s.log.warn("Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)")
-    val result: Int =
-      try
-        ITerminal.get.withRawInput(System.in.read) match {
-          case -1 => 'q'.toInt
-          case b  => b
-        }
-      catch { case _: ClosedChannelException => 'q' }
-    def retry: State = loadProjectCommand(LoadProject, loadArg) :: s.clearGlobalLog
-    def ignoreMsg: String =
-      if (Project.isProjectLoaded(s)) "using previously loaded project" else "no project loaded"
+    // In batch mode, exit with failure to avoid infinite retry loops on persistent errors
+    if (ITerminal.console.prompt == Prompt.Batch) {
+      s.exit(ok = false)
+    } else {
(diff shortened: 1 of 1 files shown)


Francluob:
hello @eed3si9n
could you please review my pr?
thanks

main/src/main/scala/sbt/Main.scala
```diff
@@ -869,6 +869,11 @@ object BuiltinCommands {
 
   @tailrec
   private def doLoadFailed(s: State, loadArg: String): State = {
+    // In batch mode, automatically retry (default behavior) without prompting
+    if (ITerminal.console.prompt == Prompt.Batch) {
+      val retry: State = loadProjectCommand(LoadProject, loadArg) :: s.clearGlobalLog
+      return retry
```
eed3si9n: If there's an error, then this could likely keep looping, so we should maybe use `q` or `l` here.

eed3si9n: Also generally we avoid using `return` statement in Scala.

Francluob: oh, makes sense. I feel like `q` should be better.

Francluob: > If there's an error, then this could likely keep looping, so we should maybe use `q` or `l` here.

pushed a fix, if you can check again, thanks

eed3si9n (approved):
Thanks!

Francluob:
> Thanks!

great! all of the tests passed 🥳

PR #8453, opened 2026-01-09
[2.x] Add testForkedParallelism setting for forked test thread count

### LINKED ISSUE #8004: testForkedParallel does not behave the same as parallelExecution
## steps

Configure a project to run its forked tests in parallel:

```
    fork                          := true,
    testForkedParallel            := true,
```

## problem

When running non-forked tests in parallel the use of concurrent restrictions limits the number of in-flight tests:

```
concurrentRestrictions := Seq(Tags.limit(Tags.All, 2))
```

However, when running forked tests, SBT uses the number of cores available to the machine:

See ForkMain.scala

```scala
    private ExecutorService executorService(
        final ForkConfiguration config, final ObjectOutputStream os) {
      if (config.isParallel()) {
        final int nbThreads = Runtime.getRuntime().availableProcessors();
        logDebug(os, "Create a test executor with a thread pool of " + nbThreads + " threads.");
        // more options later...
        // TODO we might want to configure the blocking queue with size #proc
        return Executors.newFixedThreadPool(nbThreads);
      } else {
        logDebug(os, "Create a single-thread test executor");
        return Executors.newSingleThreadExecutor();
      }
    }
```

## expectation

`testForkedParallel` should use the same concurrentRestrictions mechanism to limit the number of in-flight tests OR provide some other setting. (e.g., testForkedParallism := 2).  

## notes

There is a workaround for this which is configuring the number of cores exposed to the JVM via javaOptions:

```
    Test / javaOptions ++= 
      // Limit to two test threads
      Seq(
        "-XX:ActiveProcessorCount=2",
      )
```

grnlemon:
## Add testForkedParallelism setting

Fixes #8004

### Steps

Configure a project to run forked tests in parallel:

```scala
fork := true
testForkedParallel := true
```

### Problems

When running forked tests, sbt uses `Runtime.getRuntime().availableProcessors()` to determine the thread pool size, ignoring `concurrentRestrictions`. This is inconsistent with non-forked parallel tests.

### Expectations

Users should be able to control the number of parallel test threads in forked mode, similar to how `concurrentRestrictions` works for non-forked tests.

### Notes

Added a new setting `testForkedParallelism` that allows explicit control:

```scala
testForkedParallelism := Some(2)  // Use 2 threads
testForkedParallelism := None     // Use availableProcessors() (default)
```
---

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1139,21 +1139,9 @@ object Defaults extends BuildCommon {
         thisProject,
         fileConverter,
       ).flatMapN { (s, lt, tl, gp, ex, cp, fp, jo, clls, thisProj, c) =>
-        testForkedParallelism.toTask.flatMap { fpm =>
-          allTestGroupsTask(
-            s,
-            lt,
-            tl,
-            gp,
-            ex,
-            cp,
-            fp,
-            fpm,
-            jo,
-            clls,
-            projectId = s"${thisProj.id} / ",
-            c,
-          )
+        Def.task {
+          val fpm = testForkedParallelism.value
+          allTestGroupsTask(s, lt, tl, gp, ex, cp, fp, fpm, jo, clls, s"${thisProj.id} / ", c).value
         }
       }
     }.value),
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1133,27 +1133,28 @@ object Defaults extends BuildCommon {
         (test / testGrouping),
         (test / testExecution),
         (test / fullClasspath),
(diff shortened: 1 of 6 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1126,20 +1126,20 @@ object Defaults extends BuildCommon {
...
+      val fpm = testForkedParallelism.value
       (
         test / streams,
         loadedTestFrameworks,
         testLoader,
         (test / testGrouping),
         (test / testExecution),
         (test / fullClasspath),
-        testForkedParallel,
-        testForkedParallelism,
```
eed3si9n: I think you probably need to call `testForkedParallelism.toTaskable` to use the tuple syntax.

mkdev5: Done!

eed3si9n (approved):

PR #8455, opened 2026-01-09
Fix print command output with --no-color flag

### LINKED ISSUE #8075: sbt -no-colors breaks print
I'm trying to add `--no-color` to protect scripts from bugs such as #6232 but it doesn't seem to work.

## steps

```
% sbt -version
sbt version in this project:    1.10.1
sbt script version: 1.6.2
```

In play application:

```
sbt --batch --error --no-color 'print svc/run/mainClass'
```

## problem

empty output

## expectation

Same output as

```
sbt --batch --error 'print svc/run/mainClass'
```

which is
```
Some(play.core.server.DevServerStart)
```

## notes

exporting settings out of sbt to scripts is unreasonably difficult

Achieve3318:
Fixes #8075

Problem:
When using --no-color flag with the print command, output was suppressed. For example:
  sbt --batch --error --no-color 'print svc/run/mainClass'
would return empty output instead of the expected value.

Root Cause:
The print command's ShowConfig was using the bare 'println' function which writes to System.out. With --no-color, System.out may be redirected or configured differently, causing output to be lost.

Solution:
Changed print command to use 'scala.Console.out.println' instead of bare 'println'. Console.out properly respects the configured output stream and works correctly with terminal flags like --no-color.

This ensures print command output is always visible regardless of color/terminal settings while maintaining the direct output behavior that distinguishes it from the show command.

Changes:
- Updated Act.scala line 492 to use scala.Console.out.println

### DIFF
--- a/main/src/main/scala/sbt/internal/Act.scala
+++ b/main/src/main/scala/sbt/internal/Act.scala
@@ -489,7 +489,7 @@ object Act {
             Aggregation.ShowConfig(
               settingValues = true,
               taskValues = true,
-              print = println,
+              print = scala.Console.out.println,
               success = false
             )
           case _ => Aggregation.defaultShow(state, showTasks = action == ShowAction)


eed3si9n:
Hi thanks for the contribution. I wasn't able to reproduce https://github.com/sbt/sbt/issues/8075 on 2.0.0-RC8 so I closed it. Were you able to reproduce it prior to this fix?

eed3si9n:
Closing this for now.

PR #8460, opened 2026-01-10
[2.x] fix: Improve GCMonitor warning message clarity

### LINKED ISSUE #8002: Times and percentages exceed wall clock in GCMonitor warning
https://github.com/sbt/sbt/blob/fde334bdd671f5bc73ab0f3616161f0f1b2bbf96/main/src/main/scala/sbt/internal/GCMonitor.scala#L65-L74

https://github.com/check-spelling-sandbox/scala-k8s/actions/runs/12713111240/job/35440270607#step:12:162

> [warn] In the last 6 seconds, 49.044 (886.7%) were spent in GC. [Heap: 0.00GB free of 7.00GB, max 7.00GB] Consider increasing the JVM heap using `-Xmx` or try a different collector, e.g. `-XX:+UseG1GC`, for better performance.

I expect error/warning messages like this to use compatible time units. The first thing `6 seconds` sure feels like a wall clock. The second thing `49.044` is clearly _not_ a wall clock and it certainly isn't compatible with `6` as evidenced by the nonsensical `886.7%`.

grnlemon:
Add 'seconds' unit to GC time and clarify that the percentage represents cumulative GC pause time across all collectors.

Fixes #8002

---

### DIFF
--- a/main/src/main/scala/sbt/internal/GCMonitor.scala
+++ b/main/src/main/scala/sbt/internal/GCMonitor.scala
@@ -62,8 +62,10 @@ class GCMonitor(logger: Logger) extends GCMonitorBase with AutoCloseable {
 
   override protected def emitWarning(total: Long, over: Option[Long]): Unit = {
     val totalSeconds = total / 1000.0
-    val amountMsg = over.fold(s"$totalSeconds seconds") { d =>
-      "In the last " + (d / 1000.0).ceil.toInt + f" seconds, $totalSeconds (${total.toDouble / d * 100}%.1f%%)"
+    val amountMsg = over.fold(f"$totalSeconds%.3f seconds") { d =>
+      val dSeconds = (d / 1000.0).ceil.toInt
+      val percentage = total.toDouble / d * 100
+      f"In the last $dSeconds seconds, $totalSeconds%.3f seconds ($percentage%.1f%%) of GC pause"
     }
     val msg = s"$amountMsg were spent in GC. " +
       s"[Heap: ${gbString(runtime.freeMemory())} free " +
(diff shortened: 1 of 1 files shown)


main/src/main/scala/sbt/internal/GCMonitor.scala
```diff
@@ -62,8 +62,10 @@ class GCMonitor(logger: Logger) extends GCMonitorBase with AutoCloseable {
 
   override protected def emitWarning(total: Long, over: Option[Long]): Unit = {
     val totalSeconds = total / 1000.0
-    val amountMsg = over.fold(s"$totalSeconds seconds") { d =>
-      "In the last " + (d / 1000.0).ceil.toInt + f" seconds, $totalSeconds (${total.toDouble / d * 100}%.1f%%)"
+    val amountMsg = over.fold(f"$totalSeconds%.3f seconds") { d =>
+      val dSeconds = (d / 1000.0).ceil.toInt
+      val percentage = total.toDouble / d * 100
+      f"In the last $dSeconds seconds, $totalSeconds%.3f seconds ($percentage%.1f%%) of GC pause"
```
eed3si9n: How about adding "CPU seconds" to the total? In the last 6 seconds, 56 seconds was spent on GC is still confusing.

eed3si9n (changes requested):

main/src/main/scala/sbt/internal/GCMonitor.scala
mkdev5: Done!

eed3si9n (approved):
Thanks!

Issue #8445, opened 2026-01-07
[2.x] Task returns a file when the cache is empty, as opposed to a symlink


eed3si9n:
## Steps

1. Empty the disk cache, or I guess point at a fresh directory.
2. `show scalaCompilerBridgeBin`

```scala
sbt:aaa> show scalaCompilerBridgeBin
[info] * ${OUT}/jvm/scala-3.8.0-RC5/aaa/compiler-bridge/scala3-sbt-bridge-3.8.0-RC5.jar
[success] elapsed time: 0 s, cache 0%, 4 onsite tasks
```

## Problem

The compiler bridge is backed by a real file:

```bash
$ ls -ls target/out/jvm/scala-3.8.0-RC5/aaa/compiler-bridge/scala3-sbt-bridge-3.8.0-RC5.jar
64 -rw-r--r--  1 xxxx  xxxx  31768 Jan  7 14:24 target/out/jvm/scala-3.8.0-RC5/aaa/compiler-bridge/scala3-sbt-bridge-3.8.0-RC5.jar
```

## Expectation

We expect the task result to be same as `clean`-rerun version, which returns a symlink:

```scala
sbt:aaa> clean
[success] elapsed time: 0 s, cache 100%, 2 disk cache hits
sbt:aaa> show scalaCompilerBridgeBin
[info] * ${OUT}/jvm/scala-3.8.0-RC5/aaa/compiler-bridge/scala3-sbt-bridge-3.8.0-RC5.jar>sha256-0cd5fbfb7be6e053c42ae4db2d0a7b0cd5241d427ed9ff44ce8565167f8c173b/31768
[success] elapsed time: 0 s, cache 100%, 4 disk cache hits
```

Notice that the compiler bridge JAR is now a symlink:

```bash
$ ls -ls target/out/jvm/scala-3.8.0-RC5/aaa/compiler-bridge/scala3-sbt-bridge-3.8.0-RC5.jar
0 lrwxr-xr-x  1 xxxx  xxxx  118 Jan  7 14:27 target/out/jvm/scala-3.8.0-RC5/aaa/compiler-bridge/scala3-sbt-bridge-3.8.0-RC5.jar@ -> /Users/xxxx/Library/Caches/sbt/v2/cas/sha256-0cd5fbfb7be6e053c42ae4db2d0a7b0cd5241d427ed9ff44ce8565167f8c173b-31768
```

## Original report 

[...]

PR #8461, opened 2026-01-10
[2.x] Always create symlinks to the cache in the target locations #8445

azdrojowa123:
The problem described in #8445 is not only tied to `scalaCompilerBridgeBin`, for which it was initially reported, but also affects other tasks like `exportedProductJars`, `packageBin`, and likely more.  

I've tested the proposed approach, and it seems to work - it always creates symlinks. 
However, I think I don't have enough broad knowledge to confidently say that it won't have negative side effects in some other use cases.

### DIFF
--- a/sbt-app/src/sbt-test/source-dependencies/cache-immutability-when-package/build.sbt
+++ b/sbt-app/src/sbt-test/source-dependencies/cache-immutability-when-package/build.sbt
@@ -1,0 +1,4 @@
+lazy val root = (project in file(".")).
+  dependsOn(a)
+
+lazy val a = (project in file("a"))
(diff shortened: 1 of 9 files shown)


eed3si9n (approved):
@azdrojowa123 Thanks for the contribution. Overall LGTM! We just have to figure out why the scripted test is failing. Maybe there's already some hermeticity bug somewhere along the `run` task.

azdrojowa123:
Is `ModuleResolversTest` a flaky test?
Locally on my machine it works fine.

eed3si9n:
Any lm tests are subject to the Internet weather I guess.

PR #8463, opened 2026-01-10
[2.x] Fix #7951: Apply dependencyOverrides to delivered Ivy XML

### LINKED ISSUE #7951: dependencyOverrides are not taken into account for ivy
## steps

When publishing a module overriding a dependency version with `dependencyOverrides`, I expect to find this version in the artifact dependencies.

## problem

You can find a minimal setup [here](https://github.com/RustedBones/sbt-ivy-xml) to reproduce the problem:

When a project has the following setup:
```sbt
libraryDependencies +=  "org.slf4j" % "slf4j-api" % "managed"
dependencyOverrides +=  "org.slf4j" % "slf4j-api" % "2.0.16"
```

it creates the expected pom file with `makePom`
```xml
<dependency>
    <groupId>org.slf4j</groupId>
    <artifactId>slf4j-api</artifactId>
    <version>2.0.16</version>
</dependency>
```
but does not take in account the `dependencyOverrides` for the ivy file in `makeIvyXml`
```xml
<dependency org="org.slf4j" name="slf4j-api" rev="managed" conf="compile->default(compile)"> </dependency>
```        

This has been an issue when relying on ivy repository (ie with `publishLocal`). When trying to depend on the published artifact, the transitive dependency cannot be resolved.

## expectation

Both maven and ivy should use the overridden versions

grnlemon:
# Fix dependencyOverrides not applied to Ivy XML

## What's the problem?

When you use `dependencyOverrides` in your build and then run `publishLocal`, the generated Ivy XML file ignores your overrides. Instead of writing the actual version you specified, it writes `rev="managed"` - which is useless to anyone consuming your published artifact.

For example, if you have:

```scala
libraryDependencies += "org.slf4j" % "slf4j-api" % "1.7.36"
dependencyOverrides += "org.slf4j" % "slf4j-api" % "2.0.16"
```

The published `ivy.xml` would contain:

```xml
<dependency org="org.slf4j" name="slf4j-api" rev="managed" .../>
```

Instead of the expected:

```xml
<dependency org="org.slf4j" name="slf4j-api" rev="2.0.16" .../>
```

This breaks downstream consumers who rely on the Ivy XML for dependency resolution.

## Why does this happen?

Ivy's internal `deliver()` method doesn't apply `DependencyDescriptorMediators` (which is how sbt implements `dependencyOverrides`) when generating the XML output. The overrides work fine during resolution, but they're lost when writing the file.

## The fix

After Ivy generates the XML file, we now post-process it to apply the overrides. The fix:

1. Reads the delivered Ivy XML
2. Finds all `<dependency>` elements
3. For each one that has a matching override, updates the `rev` attribute
4. Writes the corrected XML back

It's a simple and targeted fix that doesn't change Ivy's internals - just patches up the output.

## Testing

Added unit tests that verify:
- Overridden dependencies get the correct version
- All other attributes (conf, transitive, force, etc.) are preserved  
- Dependencies without overrides are left alone

Fixes #7951

---

### DIFF
--- a/lm-ivy/src/test/scala/sbt/internal/librarymanagement/IvyActionsOverrideSpec.scala
+++ b/lm-ivy/src/test/scala/sbt/internal/librarymanagement/IvyActionsOverrideSpec.scala
@@ -5,11 +5,9 @@ import org.scalatest.matchers.should.Matchers
 
 class IvyActionsOverrideSpec extends AnyFlatSpec with Matchers {
 
-  "applyOverridesToDeliveredIvy XML transformation" should "replace rev attribute for matching dependencies" in {
-    // Simulate the override map
+  "IvyActions.applyDependencyOverrides" should "replace rev attribute for matching dependencies" in {
     val overrideMap = Map(("org.slf4j", "slf4j-api") -> "2.0.16")
 
-    // Sample Ivy XML with "managed" version (as reported in issue #7951)
     val sampleXml =
       <ivy-module version="2.0">
         <info organisation="test" module="test" revision="1.0"/>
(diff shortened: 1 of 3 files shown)


lm-ivy/src/test/scala/sbt/internal/librarymanagement/IvyActionsOverrideSpec.scala
```diff
@@ -0,0 +1,178 @@
...
+    val sampleXml =
+      <ivy-module version="2.0">
+        <info organisation="test" module="test" revision="1.0"/>
+        <dependencies>
+          <dependency org="org.slf4j" name="slf4j-api" rev="managed" conf="compile->default(compile)"/>
+          <dependency org="other.org" name="other-lib" rev="1.0.0" conf="compile->default(compile)"/>
+        </dependencies>
+      </ivy-module>
+
+    // Apply the same transformation logic as in IvyActions
```
eed3si9n: Thanks for the contribution!

Could we refactor this bit into a function inside IvyActions so we can test it without copy-pasta?

mkdev5: Done!

eed3si9n (approved):
Thanks!

PR #7823, opened 2024-10-23
[1.x] Static link `musl` during x86 Linux Native Image build

Friendseeker:
Partially fixes #7785

GraalVM does not support static link musl for arm64, we have to wait for upstream support. Hopefully not too many people use `abtn` on arm64 Linux distos without `GLIBC_2.32'  installed...

PR #8464, opened 2026-01-11
[2.x] feat: Enable musl static linking for sbtn on JDK 17+

### LINKED ISSUE #7856: Statically linked sbtn via musl
We recently reverted the musl build for sbtn due to ipcsocket JNI loading issue. However, I think we should try again. 
The lack of test coverage bit us, but I don't think that's a reason to abandon the static linking effort. The benefit of using musl is that once we can figure it out, it's somewhat guaranteed to work a long time. Whereas glibc and the fact that we are dependent of the availability of GitHub Action images etc means that we're constantly at the mercy of external forces with regards to sbtn working on which Linux machine.

grnlemon:
Ref https://github.com/sbt/sbt/pull/7823

# Enable musl static linking for sbtn on JDK 17+

Fixes #7856

## The Problem

The sbtn (sbt thin client) native image on Linux currently depends on glibc because ipcsocket uses JNI for Unix domain sockets. When building with musl for static linking, the JNI library fails to load since musl doesn't support `dlopen`.

## The Solution

Instead of upgrading to ipcsocket 2.x (which isn't ready for production), I created a `UnixDomainSocketFactory` that detects JDK 17+ at runtime and uses the native `java.net.UnixDomainSocketAddress` API directly via reflection. This completely bypasses JNI on modern JDKs.

For older JDKs (8 and 11), the factory falls back to ipcsocket 1.6.3, which is stable and well-tested.

## How It Works

The factory checks at startup whether `java.net.UnixDomainSocketAddress` is available:

- **JDK 17+**: Uses native NIO Unix domain sockets (no JNI, no native libraries)
- **JDK 8/11**: Falls back to ipcsocket's JNI-based implementation

This approach:
- Enables musl static linking on JDK 17+ without any native dependencies
- Maintains full backward compatibility with older JDKs
- Keeps the stable ipcsocket 1.6.3 instead of the unstable 2.x

## Changes

- Added `UnixDomainSocketFactory` in `protocol/src/main/java/sbt/internal/`
- Updated `ClientSocket.scala` to use the factory
- Updated `BootServerSocket.java` to use the factory  
- Updated `Server.scala` to use the factory

## Testing

The implementation compiles and passes the existing test suite. Once JDK 17 becomes the minimum requirement for sbt, ipcsocket can potentially be dropped entirely.

---


project/Dependencies.scala
```diff
@@ -20,7 +20,7 @@ object Dependencies {
   val launcherInterface = "org.scala-sbt" % "launcher-interface" % launcherVersion
   val rawLauncher = "org.scala-sbt" % "launcher" % launcherVersion
   val testInterface = "org.scala-sbt" % "test-interface" % "1.0"
-  val ipcSocket = "org.scala-sbt.ipcsocket" % "ipcsocket" % "1.6.3"
+  val ipcSocket = "org.scala-sbt.ipcsocket" % "ipcsocket" % "2.0.0-M1"
```
eed3si9n: I don't think ipcsocket 2.x is ready for prime time yet.

mkdev5: what's your opinion for the PR?

eed3si9n: The prospect of being able to work using SocketChannel is exciting, but I think there's some bugs either in my ipcsocket code or in JDK 17 implementation. I have a pull request here trying to test if I can send 1MB, and I think it gets stuck as soon as you cross the maximum Unix Domain socket window, which 108 characters or something - https://github.com/sbt/ipcsocket/pull/71

mkdev5:
All passed!

eed3si9n (commented):
Thanks! If you'd like to close #7856, then could you append

```scala
(if (isLinux) Seq("--static", "--libc=musl") else Nil)
```

to https://github.com/sbt/sbt/blob/eb81ff900040554d6f4a56cbec15558de13e3bd5/build.sbt#L836-L848 plz?

mkdev5:
> Thanks! If you'd like to close #7856, then could you append
> 
> ```scala
> (if (isLinux) Seq("--static", "--libc=musl") else Nil)
> ```
> 
> to
> 
> https://github.com/sbt/sbt/blob/eb81ff900040554d6f4a56cbec15558de13e3bd5/build.sbt#L836-L848
> 
> plz?

Done!

eed3si9n (approved):
Thanks!

PR #8465, opened 2026-01-11
[2.x] ci: Upgrade launcher-package from sbt 0.13 to sbt 1.x

### LINKED ISSUE #7826: Bump `launcher-package` to latest sbt & JDK version
https://github.com/sbt/sbt/blob/36fb500f8c9c3ef3a31f56400702c292b0069806/launcher-package/project/build.properties#L1

Main blocker is that because `launcher-package` is so outdated, it now depends on dead dependencies with years of inactivity and these dependencies do not support sbt 1.10.3.

Also a bunch of other things needs to be bumped. For example

https://github.com/sbt/sbt/blob/36fb500f8c9c3ef3a31f56400702c292b0069806/launcher-package/citest/build.sbt#L6

https://github.com/sbt/sbt/blob/36fb500f8c9c3ef3a31f56400702c292b0069806/launcher-package/citest/project/build.properties#L1

grnlemon:
# Upgrade launcher-package to sbt 1.x

Fixes #7826

## The Problem

The `launcher-package` subproject has been stuck on sbt 0.13.18 since 2018, which creates several issues:

1. **Dead dependencies** - It depends on `dispatch-http 0.8.10`, a library that hasn't been maintained in years and doesn't work with sbt 1.x
2. **Outdated tooling** - sbt 0.13.18 is ancient and incompatible with modern build tools
3. **Security concerns** - Old dependencies may have unpatched vulnerabilities

When trying to build or maintain this subproject, you'd hit errors because the dispatch library simply doesn't exist for newer Scala/sbt versions.

## What This PR Does

This PR modernizes the launcher-package build:

### 1. Upgraded sbt version
- **Before:** sbt 0.13.18 (from 2018)
- **After:** sbt 1.10.7 (current stable)

### 2. Removed dead dependency
Completely removed `dispatch-http 0.8.10` from the build. This library was only used for simple HTTP downloads, which we can do with standard Java APIs.

### 3. Updated plugins to sbt 1.x compatible versions
- `sbt-native-packager`: 1.7.6 → 1.10.4
- `sbt-pgp`: 2.1.2 → 2.3.1

### 4. Replaced HTTP download code
Instead of using the old dispatch library syntax:
```scala
import dispatch.classic._
Http(url(uri) >>> writer)
```

We now use standard Java networking:
```scala
val url = new java.net.URL(uri)
val connection = url.openConnection()
val input = connection.getInputStream
// ... buffered reading with proper cleanup
```

## Why This Matters

- **Maintainability** - The build now uses modern, supported tools
- **Reliability** - No more depending on abandoned libraries
- **Simplicity** - Standard Java APIs are easier to understand than old dispatch syntax
- **Future-proof** - Works with current and future sbt versions

## Testing

The changes are straightforward:
- ✅ Removed dead dependency
- ✅ Updated to stable, well-tested plugin versions
- ✅ Replaced with standard Java APIs that have been around for decades

The HTTP download logic is functionally identical, just using different APIs.

## Files Changed

- `launcher-package/project/build.properties` - sbt version bump
- `launcher-package/project/plugins.sbt` - removed dispatch, updated plugins
- `launcher-package/build.sbt` - replaced dispatch calls with Java URL APIs

---


eed3si9n:
Thanks for the contribution on this!

eed3si9n (approved):
Thanks!

PR #8483, opened 2026-01-11
[2.x] Avoid deprecated `java.net.URL` constructor

xuwei-k:
- https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/net/URL.html#%3Cinit%3E(java.lang.String)
- https://bugs.openjdk.org/browse/JDK-8295949


eed3si9n (approved):
Thanks!

PR #8486, opened 2026-01-12
[2.x] fix: Fix snapshot sbt version invalidation 7713

### LINKED ISSUE #7713: [2.x] Invalidation of build.sbt on snapshot sbt version
## steps

1. Load a build with a snapshot version of sbt
2. Break the binary compatibility and republish sbt
3. Reload the build

## problem

The build is not recompiled and it fails to load with a `java.lang.NoSuchMethodError`.

## expectation

The build should be recompiled.

grnlemon:
# Fix: Invalidate build.sbt cache when sbt SNAPSHOT version changes

## Problem

When developing with a SNAPSHOT version of sbt, there's a frustrating issue: if you republish sbt with breaking changes and then reload your build, it fails with a cryptic `java.lang.NoSuchMethodError`.

This happens because sbt caches the compiled build definition (the `.class` files in `project/target/config-classes/`), but the cache key doesn't account for changes to sbt itself. So even though the sbt jars have changed, sbt happily reuses the old cached classes that were compiled against the previous version.

## Solution

This PR adds a hash of SNAPSHOT jars to the cache key used for build definition compilation. When sbt detects that any SNAPSHOT or `-bin-` jars have changed (by checking their modification times), it will recompile the build definition instead of using stale cached classes.

The fix is intentionally targeted:
- Only SNAPSHOT and `-bin-` jars are tracked (development versions)
- Stable releases are unaffected, so there's no unnecessary cache invalidation
- Modification times are included to catch republished snapshots with the same version string

## Testing

- All existing unit tests pass
- Scalafmt check passes

## Related

Fixes #7713

---

### DIFF
--- a/buildfile/src/main/scala/sbt/internal/Eval.scala
+++ b/buildfile/src/main/scala/sbt/internal/Eval.scala
@@ -59,8 +59,7 @@ class Eval(
       snapshotJars.sorted.foreach { path =>
         val file = path.toFile
         digester.update(bytes(path.toString))
-        if file.exists then
-          digester.update(bytes(file.lastModified.toString))
+        if file.exists then digester.update(bytes(file.lastModified.toString))
       }
       Hash.toHex(digester.digest())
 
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

eed3si9n:
I'm getting different errors, so I opened https://github.com/sbt/sbt/issues/8511.

PR #8490, opened 2026-01-12
[2.x] feat: Cache failed compilation to avoid repeated failures

### LINKED ISSUE #7662: [2.x] Cache failed compilation, and other failed tasks
BSP clients (Metals) can trigger many compilations in a row. But sbt does not cache failed compilation, so it restarts the compilation many times which is super slow. To improve this we can use Zinc or the `ActionCache` to cache compilation failures.

Also we can generalize this to all `cachedTask`, or make it configurable. But only some known and serializable types of exception should be cached.

grnlemon:
When a compilation fails with CompileFailed, the failure is now cached so that subsequent builds with the same inputs don't re-run the failed compilation. This significantly improves the experience when using BSP clients like Metals that may trigger many compilations in a row.

The implementation:
- Adds CachedCompileFailure, CachedProblem, and CachedPosition types to serialize compilation failures
- Modifies ActionCache.cache to catch CompileFailed exceptions and store them in the cache with exitCode=1
- On cache lookup, checks for cached failures first and re-throws the cached exception if found
- Fixes DiskActionCacheStore.put to preserve exitCode from request
- Adds unit test to verify cached failure behavior

Fixes #7662

---

### DIFF
--- a/util-cache/src/main/scala/sbt/util/ActionCache.scala
+++ b/util-cache/src/main/scala/sbt/util/ActionCache.scala
@@ -24,11 +24,13 @@ import sjsonnew.{ HashWriter, JsonFormat }
 import sjsonnew.support.murmurhash.Hasher
 import sjsonnew.support.scalajson.unsafe.{ CompactPrinter, Converter, Parser }
 import scala.quoted.{ Expr, FromExpr, ToExpr, Quotes }
-import xsbti.{ FileConverter, HashedVirtualFileRef, VirtualFile, VirtualFileRef }
+import xsbti.{ CompileFailed, FileConverter, HashedVirtualFileRef, VirtualFile, VirtualFileRef }
 
 object ActionCache:
   private[sbt] val dirZipExt = ".sbtdir.zip"
   private[sbt] val manifestFileName = "sbtdir_manifest.json"
+  private[sbt] val failureFileName = "failure.json"
+  private[sbt] val failureExitCode = 1
 
   /**
    * This is a key function that drives remote caching.
@@ -54,11 +56,29 @@ object ActionCache:
       action: I => InternalActionResult[O],
   ): O =
     import config.*
+
+    def cacheFailure(e: CompileFailed): Nothing =
+      // Cache the failure so subsequent builds don't re-run failed compilation
+      // This fixes https://github.com/sbt/sbt/issues/7662
+      cacheEventLog.append(ActionCacheEvent.OnsiteTask)
+      val (input, failurePath) = mkFailureInput(key, codeContentHash, extraHash)
+      val cachedFailure = CachedCompileFailure.fromException(e)
+      val json = Converter.toJsonUnsafe(cachedFailure)
+      val failureFile = StringVirtualFile1(failurePath, CompactPrinter(json))
+      store.put(
+        UpdateActionResultRequest(input, Vector(failureFile), exitCode = failureExitCode)
+      ) match
+        case Right(_) => ()
+        case Left(_)  => () // Ignore cache storage errors for failures
+      throw e
+
     def organicTask: O =
       // run action(...) and combine the newResult with outputs
       val InternalActionResult(result, outputs) =
         try action(key): @unchecked
         catch
+          case e: CompileFailed =>
+            cacheFailure(e)
           case e: Exception =>
             cacheEventLog.append(ActionCacheEvent.Error)
             throw e
@@ -83,9 +103,15 @@ object ActionCache:
             result
           case Left(e) => throw e
 
-    get(key, codeContentHash, extraHash, tags, config) match
-      case Some(value) => value
-      case None        => organicTask
+    // Check for cached failure first
+    getCachedFailure(key, codeContentHash, extraHash, config) match
+      case Some(failure) =>
+        config.cacheEventLog.append(ActionCacheEvent.Found("cached-failure"))
+        throw failure.toException
+      case None =>
+        get(key, codeContentHash, extraHash, tags, config) match
+          case Some(value) => value
+          case None        => organicTask
   end cache
 
   /**
@@ -174,6 +200,44 @@ object ActionCache:
     val input =
       Digest.sha256Hash(codeContentHash, extraHash, Digest.dummy(Hasher.hashUnsafe[I](key)))
     (input, s"$${OUT}/value/$input.json")
+
+  private inline def mkFailureInput[I: HashWriter](
+      key: I,
+      codeContentHash: Digest,
+      extraHash: Digest
+  ): (Digest, String) =
+    val input =
+      Digest.sha256Hash(codeContentHash, extraHash, Digest.dummy(Hasher.hashUnsafe[I](key)))
+    // Use a different path for failures to avoid conflicts with success cache
+    (Digest.sha256Hash(input, Digest.dummy(failureExitCode)), s"$${OUT}/failure/$input.json")
+
+  /**
+   * Retrieves a cached compilation failure, if any.
+   */
+  private def getCachedFailure[I: HashWriter](
+      key: I,
+      codeContentHash: Digest,
+      extraHash: Digest,
+      config: BuildWideCacheConfiguration,
+  ): Option[CachedCompileFailure] =
+    val (input, failurePath) = mkFailureInput(key, codeContentHash, extraHash)
+    val getRequest =
+      GetActionResultRequest(input, inlineStdout = false, inlineStderr = false, Vector(failurePath))
+    config.store.get(getRequest) match
+      case Right(result) if result.exitCode.contains(failureExitCode) =>
+        result.contents.headOption match
+          case Some(head) =>
+            val str = String(head.array(), StandardCharsets.UTF_8)
+            val json = Parser.parseUnsafe(str)
+            Some(Converter.fromJsonUnsafe[CachedCompileFailure](json))
+          case None =>
+            val paths = config.store.syncBlobs(result.outputFiles, config.outputDirectory)
+            if paths.isEmpty then None
+            else
+              val str = IO.read(paths.head.toFile())
+              val json = Parser.parseUnsafe(str)
+              Some(Converter.fromJsonUnsafe[CachedCompileFailure](json))
+      case _ => None
 
   def manifestFromFile(manifest: Path): Manifest =
     import sbt.internal.util.codec.ManifestCodec.given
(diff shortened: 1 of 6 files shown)


util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -83,9 +103,15 @@ object ActionCache:
...
-      case None        => organicTask
+    // Check for cached failure first
+    getCachedFailure(key, codeContentHash, extraHash, config) match
+      case Some(failure) =>
+        config.cacheEventLog.append(ActionCacheEvent.Found("cached-failure"))
+        throw failure.toException
+      case None =>
+        get(key, codeContentHash, extraHash, tags, config) match
+          case Some(value) => value
+          case None        => organicTask
```
eed3si9n: So if compilation is hermetic (that is the digest captured all necessary information), both good compilation and bad compilation can reside on the same space? I want to avoid making 2x call because if remote cache is configured, it becomes network call.

https://github.com/sbt/sbt/blob/develop/util-cache/src/main/contraband-scala/sbt/util/ActionResult.scala has `exitCode`, which we can use to distinguish bad results from good, by stuffing `Some(1)` or something.

util-cache/src/main/scala/sbt/util/CachedCompileFailure.scala
```diff
@@ -0,0 +1,187 @@
...
+  def fromProblem(p: Problem): CachedProblem =
+    CachedProblem(
+      category = p.category(),
+      severity = p.severity().toString,
+      message = p.message(),
+      position = CachedPosition.fromPosition(p.position()),
+      rendered = if p.rendered().isPresent then Some(p.rendered().get()) else None
+    )
+
+  given JsonFormat[CachedProblem] = new JsonFormat[CachedProblem]:
```
eed3si9n: There's a JsonFormat for Problem here (https://github.com/sbt/sbt/blob/develop/internal/util-logging/src/main/scala/sbt/internal/util/codec/ProblemFormats.scala). 

There's an example code in Zinc that materializes that as:

```scala
  lazy val problemFormats: ProblemFormats = new ProblemFormats with SeverityFormats
    with PositionFormats with sjsonnew.BasicJsonProtocol {}
```

mkdev5:
all passed!

util-cache/${OUT}/value/sha256-8769430c2339a7cbaaa5b190624c7baa20f54d092424131f0055699b423ce6cc/48.json
eed3si9n, line 1: Where did this file come from?

mkdev5: sorry, removed it

util-cache/src/main/scala/sbt/util/CachedCompileFailure.scala
```diff
@@ -0,0 +1,61 @@
...
+import xsbti.{ CompileFailed, Problem }
+
+/**
+ * A serializable representation of a CompileFailed exception.
+ * This allows caching compilation failures so that repeated builds
+ * don't re-run failed compilations unnecessarily.
+ *
+ * Fixes https://github.com/sbt/sbt/issues/7662
+ */
+final case class CachedCompileFailure(
```
eed3si9n: Could we move this to https://github.com/sbt/sbt/blob/develop/util-cache/src/main/contraband/actionresult.contra please? Contraband is our made-up language based on GraphQL Schema (https://www.scala-sbt.org/contraband/schema.html), and it's able to add field without breaking binary compatibility.

Also what do you think about redesigning this to be `GenericFailure`? Basically add a field named `kind: String` and if the original was `CompileFailed`, then we can store `"CompileFailed"`. That way, we could potentially use this to cache non-compilation failures in the future.

eed3si9n:
I think we should keep going with this idea, but one concern I have is that sometimes people use `compile` to display all the compilation warnings and errors, and I'm not sure cached failure captures all that details.

mkdev5:
all passed! please review it again and approve if it's ok

eed3si9n:
I built sbt locally and tried to use it:

```bash
sbt:aaa> compile
[info] compiling 2 Scala sources to /private/tmp/aaa/target/out/jvm/scala-3.7.3/aaa/classes ...
[error] -- [E103] Syntax Error: /private/tmp/aaa/src/main/scala/example/A.scala:3:0 ----
[error] 3 |aaclass A
[error]   |^^^^^^^
[error]   |Illegal start of toplevel definition
[error]   |
[error]   | longer explanation available when compiling with `-explain`
[error] -- [E040] Syntax Error: /private/tmp/aaa/src/main/scala/example/Main.scala:5:0 -
[error] 5 |
[error]   |^
[error]   |')' expected, but eof found
[error] two errors found
[error] elapsed time: 0 s, cache 84%, 1 disk cache hit, 10 symlink cache hits, 2 onsite tasks
```

```bash
sbt:aaa> compile
[error] stack trace is suppressed; run last Compile / compileIncremental for the full output
[error] (Compile / compileIncremental) sbt.util.CachedCompileFailure$$anon$1: Compilation failed
[error] elapsed time: 0 s, cache 100%, 1 cached-failure cache hit, 1 disk cache hit, 11 symlink cache hits
```

For command-line usage, I don't think this is usable without also relaying the problem information back to the log. Using

```scala
scalaVersion := "3.7.3"
```

the problem list contains `Problem` structure, and Scala 3 seems to store the pretty-printed error in `rendered` field to be replayed, but with

```scala
scalaVersion := "2.13.18"
```

the problem list contains `Vector(Optional.empty, Optional.empty)`. So maybe what we should do is inspect the problem list before caching.

Also for Metals usage, where OP originally wanted this feature to be useful, I wonder if we also need to replay the problem to LSP reporter so VS Code would have it. I haven't tested this with Metals yet.

mkdev5:
> I built sbt locally and tried to use it:
> 
> ```shell
> sbt:aaa> compile
> [info] compiling 2 Scala sources to /private/tmp/aaa/target/out/jvm/scala-3.7.3/aaa/classes ...
> [error] -- [E103] Syntax Error: /private/tmp/aaa/src/main/scala/example/A.scala:3:0 ----
> [error] 3 |aaclass A
> [error]   |^^^^^^^
> [error]   |Illegal start of toplevel definition
> [error]   |
> [error]   | longer explanation available when compiling with `-explain`
> [error] -- [E040] Syntax Error: /private/tmp/aaa/src/main/scala/example/Main.scala:5:0 -
> [error] 5 |
> [error]   |^
> [error]   |')' expected, but eof found
> [error] two errors found
> [error] (Compile / compileIncremental) Compilation failed
> [error] elapsed time: 0 s, cache 84%, 1 disk cache hit, 10 symlink cache hits, 2 onsite tasks
> ```
> 
> ```shell
> sbt:aaa> compile
> [error] stack trace is suppressed; run last Compile / compileIncremental for the full output
> [error] (Compile / compileIncremental) sbt.util.CachedCompileFailure$$anon$1: Compilation failed
> [error] elapsed time: 0 s, cache 100%, 1 cached-failure cache hit, 1 disk cache hit, 11 symlink cache hits
> ```
> 
> For command-line usage, I don't think this is usable without also relaying the problem information back to the log. Using
> 
> ```scala
> scalaVersion := "3.7.3"
> ```
> 
> the problem list contains `Problem` structure, and Scala 3 seems to store the pretty-printed error in `rendered` field to be replayed, but with
> 
> ```scala
> scalaVersion := "2.13.18"
> ```
> 
> the problem list contains `Vector(Optional.empty, Optional.empty)`. So maybe what we should do is inspect the problem list before caching.
> 
> Also for Metals usage, where OP originally wanted this feature to be useful, I wonder if we also need to replay the problem to LSP reporter so VS Code would have it. I haven't tested this with Metals yet.

Thanks for testing! I've added a `replay()` method that logs the cached problems before throwing the exception. It uses the `rendered` field when available (Scala 3), and falls back to formatting from `position` + `message` for Scala 2.13.

For Metals/LSP, I haven't tested that yet - will look into whether we need to replay to the BSP reporter as well.

mkdev5:
For Metals/LSP, I think the issue is that when a cached `CompileFailed` is thrown from `ActionCache`, it bypasses the BSP reporter notification (`sendFailureReport`/`notifyFailure`).

One option would be to wrap the cached compile task with a try-catch that calls the BSP methods on `CompileFailed`. But I'm not sure if that's the right approach architecturally - would appreciate your guidance on how you'd like this handled.

eed3si9n:
Looking at the code, TaskZero / compileIncremental is uncached, so it's possible that we already have that try catch for us. You can put println temporarily in there to check:

https://github.com/sbt/sbt/blob/add43bd23031d50ac72ad25add82fd5bc3b5d569/main/src/main/scala/sbt/Defaults.scala#L2188-L2192

mkdev5:
I tested with println and confirmed that the cached `CachedCompileFailure` does flow through the `Result.Inc` handler in `Defaults.scala` lines 2188-2192. The `compileFailed` is correctly extracted and `bspTask.notifyFailure()` is called.

Debug output on second compile:
```
[DEBUG] Result.Inc caught: Some(sbt.util.CachedCompileFailure$$anon$1: Compilation failed)
[DEBUG] compileFailed extracted: Some(sbt.util.CachedCompileFailure$$anon$1: Compilation failed)
```

So the BSP reporter notification should already work for Metals without additional changes. The error replay to the log is also working as shown in the output.

util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -54,11 +56,30 @@ object ActionCache:
...
+      cacheEventLog.append(ActionCacheEvent.OnsiteTask)
+      val (input, valuePath) = mkInput(key, codeContentHash, extraHash)
+      val cachedFailure = CachedCompileFailure.fromException(e)
+      val json = Converter.toJsonUnsafe(cachedFailure)
+      val failureFile = StringVirtualFile1(valuePath, CompactPrinter(json))
+      store.put(
+        UpdateActionResultRequest(input, Vector(failureFile), exitCode = failureExitCode)
+      ) match
+        case Right(_) => ()
+        case Left(_)  => () // Ignore cache storage errors for failures
```
eed3si9n:
```suggestion
      store.put(
        UpdateActionResultRequest(input, Vector(failureFile), exitCode = failureExitCode)
      )
```

eed3si9n (approved):
Thanks!

util-cache/src/main/scala/sbt/util/ActionCache.scala
mkdev5: Fixed!

mkdev5:
All passed! Can you merge it now?

PR #8492, opened 2026-01-12
[2.x] feat: Set terminal window title when running applications

### LINKED ISSUE #7586: set window title to organization + artifactId  + sbt command
I have many microservices. When developing my features, I start most of them using docker, but some of them I start with 
```
sbt run
```
or
```
sbt
run
```

The title of my terminal then just says "sbt run" or "sbt", which makes it somehow hard to differentiate different windows/services, because I start them all with the same command, e.g. take a look at the following screenshot:

grafik
or even more general:
grafik

Now if I have 2+ windows, these will all look very similar. If I want to figure out what application is running in that window, I have to either stop the app to see the directory/prompt or try to guess from the log messages which service is running in that window. 

It would be great if there was a way that sbt would set the window title to something like 
"sbt run: myOrg % artifactId % version" 
e.g.
"sbt run: com.dominikdorn.sbt % sample % 0.0.8-0-02777d76-20240620-2025" 

I can set the title interactively in a sbt session using

```
sbt
console
print("\u001b]0;sbt run : com.dominikdorn % sample % 0.0.8-0-02777d76-20240620-2025 \u0
007")
CTRL+D
```

It would be great if sbt could do that out of the box on linux/unix terminals :-)

grnlemon:
# feat: Set terminal window title when running applications

## What's the problem?

When you have multiple microservices running with `sbt run`, all terminal windows show the same title ("sbt run" or just "sbt"). This makes it hard to tell which service is running in which window without stopping the app or guessing from log messages.

## What's the fix?

Now when you run `sbt run`, `runMain`, `bgRun`, or `bgRunMain`, the terminal window title is automatically set to show the project info:

```
sbt run: com.example % my-service % 1.0.0
```

This makes it easy to identify which service is running in each terminal window at a glance.

## How it works

Uses standard ANSI escape codes to set the terminal title. Only activates when running in a proper terminal (checks for `TERM` environment variable).

## Testing

- Compiled successfully
- scalafmtCheck passed

Fixes #7586

---


eed3si9n:
sbt 2.x has three kinds of run (https://eed3si9n.com/sudori-part7-client-side-run-with-sbt/):
1. client-side (sbtn) run
2. server-side in-process run
3. server-side forked run

The default behavior is client-side run, which means sbt server code just sends JSON signal to sbtn, and sbtn would fork a `java` process do run the user program. This means, we'd probably want to replicate the window title changing code to sbtn as well. The sbtn client-side run code is around here:
- https://github.com/sbt/sbt/blob/develop/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala#L737
- or truthfully, the same code on sbt 1.x side https://github.com/sbt/sbt/blob/1.12.x/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala#L738 (I can back port if it's fixed on develop branch)

main/src/main/scala/sbt/internal/RunUtil.scala
```diff
@@ -125,6 +139,7 @@ object RunUtil:
       val hashClasspath = (bgRunMain / bgHashClasspath).value
       val fo = (run / forkOptions).value
       val state = Keys.state.value
+      setRunWindowTitle("runMain", organization.value, name.value, version.value)
```
eed3si9n: I think this should be moved to `else` clause (line 159) for the server-side run only.

main/src/main/scala/sbt/internal/RunUtil.scala
```diff
@@ -181,6 +196,7 @@ object RunUtil:
       val hashClasspath = (bgRun / bgHashClasspath).value
       val fo = (run / forkOptions).value
       val state = Keys.state.value
+      setRunWindowTitle("run", organization.value, name.value, version.value)
```
eed3si9n: Same here.

protocol/src/main/contraband/worker.contra
```diff
@@ -37,6 +37,7 @@ type RunInfo {
   jvm: Boolean!
   jvmRunInfo: sbt.internal.worker.JvmRunInfo,
   nativeRunInfo: sbt.internal.worker.NativeRunInfo @since("0.1.0"),
+  windowTitle: String @since("0.1.0"),
```
eed3si9n:
```suggestion
  windowTitle: String @since("0.2.0"),
```

eed3si9n: If we bump this number, we get more overloaded constructors.

main/src/main/scala/sbt/internal/RunUtil.scala
```diff
@@ -15,6 +15,19 @@ import sbt.util.CacheImplicits.given
 import xsbti.FileConverter
 
 object RunUtil:
+  private def setWindowTitle(title: String): Unit =
+    if System.console() != null && System.getenv("TERM") != null then
+      print(s"\u001b]0;$title\u0007")
+      System.out.flush()
```
eed3si9n:
```suggestion
    if Console.out != null && sys.env.get("TERM").isDefined then
      Console.out.print(s"\u001b]0;$title\u0007")
      Console.out.flush()
```

https://scala-lang.org/api/3.7.2/scala/Console$.html

eed3si9n (approved):
Some minor nits, but overall LGTM. Thanks!

PR #8494, opened 2026-01-12
[2.x] fix: Prevent server subprocess hang on Linux (#8442)

### LINKED ISSUE #8442: [2.x] hang when trying to start a new server subprocess
## steps
Using SBT 2.0.0-RC8, if I start the server interactively (sbt --server), and then I run the "sbt" command separately (as a client), I can interact with the system as normal.

If I don't have a running server and I run "sbt", it tries to fork a server, but that seems to hang - it never creates the project/target/active.json file that the client is polling for.

Snooping on the command, it appears to run:
	SBT_TERMINAL_PROPS="0,0,false,false,false" sbt -Dsbt.io.virtual=true --detach-stdio --server

While the client is hanging waiting for the launched server to run, if I run the above command in a separate window, that does create the active.json file, and the client connects to it normally. 

Oddly enough, if I run "sbt --debug" - it seems to start and connect to the server just fine.

Also note that this only happens on Linux, this project loads fine on Mac OS (Java 17, Apple M2 silicon).

## problem

It seems like when the SBT server is forked on Linux in this manner, that it hangs for some reason. If we start it manually, or even launch sbt with the "--debug" flag, it works normally. Perhaps the input handling isn't quite right.

## expectation

That the forked sbt server would start, and we would use SBT normally (as it does when passed the --debug flag).

## notes

This is on Ubuntu 24.04, Java OpenJDK 17.0.17. SBT installed via "cs" on the stable channel. Using SBT 2.0.0-RC8 (also seen in RC7).

If I use strace to examine what the server process is doing, it's just looping on a bunch of futex calls.

calm329:
## Summary
- Fix hang when sbt client spawns server subprocess on Linux
- When the server is forked with `--detach-stdio`, stdout/stderr now redirect to `DISCARD` instead of inheriting from parent
- On Linux, inherited pipes have ~64KB buffers that block writes when full, causing the server to hang during initialization (e.g., when printing lint warnings for projects with 45+ subprojects)

## Root Cause
When `NetworkClient.waitForServer` spawns the server process:
1. stdout/stderr were inherited from the parent process
2. The client communicates with the server via socket, never consuming stdout/stderr
3. On Linux, once the pipe buffer fills (~64KB), `FileOutputStream.writeBytes` blocks indefinitely
4. This caused the server to hang before creating `active.json`

Workarounds that worked (confirming the diagnosis):
- `--debug` flag (affects logging behavior)
- `-Dsbt.log.noformat=true` (reduces output volume)

## Fix
Add `.redirectOutput(Redirect.DISCARD).redirectError(Redirect.DISCARD)` to the ProcessBuilder, which redirects to `/dev/null` on Unix.

## Test plan
- [x] Code compiles successfully
- [x] Verified client doesn't read from process stdout/stderr (communicates via socket)
- [ ] Manual test on Linux with 45+ subproject build (cannot reproduce on macOS due to different pipe buffering)

Fixes #8442

---

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=148254234

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -390,10 +390,17 @@ class NetworkClient(
           if (Util.isEmacs && !Util.isWindows) List("nohup")
           else Nil
 
+        // https://github.com/sbt/sbt/issues/8442
+        // On Linux, if stdout/stderr are inherited and the buffer fills up (~64KB),
+        // the server process will block on writes. Discard output since the server
+        // communicates via the boot socket, not stdout/stderr.
+        val nullFile = new File(if (Util.isWindows) "NUL" else "/dev/null")
         val processBuilder =
           new ProcessBuilder((nohup ++ cmd)*)
             .directory(arguments.baseDirectory)
             .redirectInput(Redirect.PIPE)
+            .redirectOutput(nullFile)
+            .redirectError(nullFile)
         processBuilder.environment.put(Terminal.TERMINAL_PROPS, props)
         Try(processBuilder.start()) match {
           case Success(process) =>


eed3si9n (approved):
Thanks!

PR #8500, opened 2026-01-12
Fix Windows launcher to allow -D system property without value (fixes #7332)

### LINKED ISSUE #7332: Windows sbt launcher script doesn't handle system property without value
## steps

Using sbt 1.9.2 launcher

```
$ sbt -Dfoo
-Dfoo is missing a value
```

## problem

## expectation

mohansinghi:
Fix Windows launcher to allow -D system property without value (fixes #7332)

## Problem
On Windows, when running `sbt -Dfoo` (system property without a value), the launcher would fail with an error:

```
-Dfoo is missing a value
```

This behavior was inconsistent with Unix/macOS systems, where `-Dfoo` works correctly and sets the property to an empty string, allowing `sys.props.contains("foo")` to return `true`.

## Solution
Modified the Windows batch script (`sbt.bat`) to allow `-D` system properties without values, matching the Unix behavior. When `-Dfoo` is passed without a value, it now passes through to Java instead of erroring.

## Changes
- **Modified** `launcher-package/src/universal/bin/sbt.bat`: Updated `-D` argument parsing to allow properties without values
- **Added** integration test in `ScriptTest.scala`: Test case to verify `-Dfoo` without value is properly passed through

## Testing
- Added integration test that verifies `-Dfoo` is passed through correctly
- The fix maintains backward compatibility with existing `-Dkey=value` syntax

## Related Issue
Fixes #7332

---

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=158349177


mohansinghi:
@eed3si9n  Could you please check my pr?

eed3si9n (approved):
Thanks!

mohansinghi:
@eed3si9n Really sorry for my mistake, I just signed CLA.
Then I have fixed one issue. Could you please review again?

eed3si9n:
No need to apologize. It's all good.

launcher-package/src/universal/bin/sbt.bat
```diff
@@ -508,16 +508,14 @@ if "%g:~0,2%" == "-D" (
   for /F "tokens=1 delims==" %%a in ("%g%") do (
     rem make sure it doesn't have the '=' already
     if "%g%" == "%%a" (
-      if not "%~1" == "" (
-        call :dlog [args_loop] -D argument %~0=%~1
-        set "SBT_ARGS=!SBT_ARGS! %~0=%~1"
-        shift
-        goto args_loop
```
eed3si9n, line 515: We might need to keep this logic around. See the comment a few lines above:

```bat
  rem special handling for -D since '=' gets parsed away
```

0xsid0703: @eed3si9n I just restored the logic that handles `-Dfoo` followed by a separate argument.

mohansinghi:
@eed3si9n Could you please help me? I have tested on Windows, it worked well, but it isn't finishing now.

eed3si9n:
Can you run the following:

```bash
sbt -Dsbt.build.version=1.12.0 launcherPackageIntegrationTest/test
```

If the CI is getting stuck, I have a feeling it's because it's going into an infinite loop.

mohansinghi:
> Can you run the following:
> 
> ```shell
> sbt -Dsbt.build.version=1.12.0 launcherPackageIntegrationTest/test
> ```
> 
> If the CI is getting stuck, I have a feeling it's because it's going into an infinite loop.

Thank you for your help, I tried to run that command, no error on my end.

mohansinghi:
@eed3si9n I have fixed and tested on my local.
image

But I am not sure why error is occurring on GitHub Workflow.
Could you please help me?

launcher-package/src/universal/bin/sbt.bat
```diff
@@ -505,19 +505,20 @@ if "%~0" == "init" (
...
-        echo %g% is missing a value
-        goto error
-      )
+      rem -Dfoo without '=' - pass through as-is to match Unix behavior
+      rem This allows -Dfoo without value, setting property to empty string
+      rem We don't consume the next argument to avoid treating commands as values
+      call :dlog [args_loop] -D argument %~0
+      set "SBT_ARGS=!SBT_ARGS! %~0"
+      goto args_loop
     ) else (
```
mzuehlke: I have zero clue about Windows batch scripts,
but I can see that now both branches are identical.
So either have different code in the `if` and `else` section or remove the `if / else`

mohansinghi:
@mzuehlke Could you please help me understand why the windows-latest job is running indefinitely?

mzuehlke:
> @mzuehlke Could you please help me understand why the windows-latest job is running indefinitely?

Sorry, I have zero knowledge about the windows client

eed3si9n:
A lot of the problems are stemming from the fact that `=` sign gets parsed away. As far as I know, that's not an inherent problem with Batch in general, but rather a behavior of `shift` command that we currently use. One potential fix might be to rewrite the script without using `shift`.

mohansinghi:
@eed3si9n I tried to test without `shift`, but result is still same...

eed3si9n:
Then I suggest closing this PR for now, and move on to some other issue. Automated testing isn't perfect, but that's what's holding the whole thing together, so we'd have to play by its rule.

PR #8501, opened 2026-01-12
[2.x] fix: Invalidate update cache across commands when dependencies change

### LINKED ISSUE #8357: itTest/update does not get invalidated, if a/compile is issued on its own
## steps

Using sbt 1.11.7 (and way earlier)
Compile a project
Change a library version
Reload
Compile project again

## problem

Classpath is not updated

## expectation

Old library version is removed, new one is added

## notes

If i go to coursier cache and delete old libs, and then compile - the problem goes away
there are some suspicious logs, mentioning the deleted library:
```
> show itTests/Test/fullClasspath
[warn] spec_3-0.9.5.jar no longer exists at C:\Users\user\AppData\Local\Coursier\Cache\v1\http\test.com\repo\sample\spec_3\0.9.5\spec_3-0.9.5.jar
```

calm329:
## Summary
Fix update cache invalidation across sbt command invocations when a dependency's dependencies change.

## Problem
When project A's dependencies changed and A was compiled in one command, project B (depending on A) would not invalidate its update cache in a subsequent command. This caused stale classpaths.

The root cause was that `depsUpdated` only checked `!stats.cached`, which only detected fresh resolves within the same command. When a dependency was served from cache (even if resolved fresh in a previous command), `cached` was marked `true`, causing incorrect cache reuse.

Debug scenario from issue:
sbt:aaa> clean
sbt:aaa> a/compile
sbt:aaa> show itTests/depsUpdated
[info] * false   <-- BUG: should be true

## Solution
Added `stamp: String` field to `UpdateStats` that records when the update was resolved. This stamp persists across commands and enables accurate cross-command comparison:

- If any dependency's stamp > our cached stamp, we re-resolve
- Falls back to `!cached` check for backwards compatibility with old caches
- Using `String` type allows future transition from timestamps to content hashes

## Changes
- Add `stamp: String` field to UpdateStats schema (default `""` for compat)
- Set `stamp = System.currentTimeMillis().toString` when creating UpdateStats
- Modify `cachedUpdate` to compare stamps instead of just `cached` flag
- Add backwards-compatible JSON deserialization for old cache files
- Add scripted test `i8357-transitive-update-invalidation`

## Test plan
- [x] Compiles successfully
- [x] `lmCore/test` passes
- [ ] Run scripted test: `scripted dependency-management/i8357-transitive-update-invalidation`

Fixes #8357

---

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=148254234

### DIFF
--- a/lm-core/src/main/contraband-scala/sbt/librarymanagement/UpdateStats.scala
+++ b/lm-core/src/main/contraband-scala/sbt/librarymanagement/UpdateStats.scala
@@ -10,9 +10,9 @@ final class UpdateStats private (
   val downloadTime: Long,
   val downloadSize: Long,
   val cached: Boolean,
-  val stamp: String) extends Serializable {
+  val stamp: Option[String]) extends Serializable {
   
-  private def this(resolveTime: Long, downloadTime: Long, downloadSize: Long, cached: Boolean) = this(resolveTime, downloadTime, downloadSize, cached, "")
+  private def this(resolveTime: Long, downloadTime: Long, downloadSize: Long, cached: Boolean) = this(resolveTime, downloadTime, downloadSize, cached, None)
   
   override def equals(o: Any): Boolean = this.eq(o.asInstanceOf[AnyRef]) || (o match {
     case x: UpdateStats => (this.resolveTime == x.resolveTime) && (this.downloadTime == x.downloadTime) && (this.downloadSize == x.downloadSize) && (this.cached == x.cached) && (this.stamp == x.stamp)
@@ -24,7 +24,7 @@ final class UpdateStats private (
   override def toString: String = {
     Seq("Resolve time: " + resolveTime + " ms", "Download time: " + downloadTime + " ms", "Download size: " + downloadSize + " bytes").mkString(", ")
   }
-  private def copy(resolveTime: Long = resolveTime, downloadTime: Long = downloadTime, downloadSize: Long = downloadSize, cached: Boolean = cached, stamp: String = stamp): UpdateStats = {
+  private def copy(resolveTime: Long = resolveTime, downloadTime: Long = downloadTime, downloadSize: Long = downloadSize, cached: Boolean = cached, stamp: Option[String] = stamp): UpdateStats = {
     new UpdateStats(resolveTime, downloadTime, downloadSize, cached, stamp)
   }
   def withResolveTime(resolveTime: Long): UpdateStats = {
@@ -39,12 +39,12 @@ final class UpdateStats private (
   def withCached(cached: Boolean): UpdateStats = {
     copy(cached = cached)
   }
-  def withStamp(stamp: String): UpdateStats = {
+  def withStamp(stamp: Option[String]): UpdateStats = {
     copy(stamp = stamp)
   }
 }
 object UpdateStats {
   
   def apply(resolveTime: Long, downloadTime: Long, downloadSize: Long, cached: Boolean): UpdateStats = new UpdateStats(resolveTime, downloadTime, downloadSize, cached)
-  def apply(resolveTime: Long, downloadTime: Long, downloadSize: Long, cached: Boolean, stamp: String): UpdateStats = new UpdateStats(resolveTime, downloadTime, downloadSize, cached, stamp)
+  def apply(resolveTime: Long, downloadTime: Long, downloadSize: Long, cached: Boolean, stamp: Option[String]): UpdateStats = new UpdateStats(resolveTime, downloadTime, downloadSize, cached, stamp)
 }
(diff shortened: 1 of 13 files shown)


main/src/main/scala/sbt/internal/LibraryManagement.scala
```diff
@@ -123,6 +123,18 @@ private[sbt] object LibraryManagement {
...
+      // The `!stats.cached` check handles within-command invalidation (for backwards compat with empty stamp).
+      val depsUpdated = transitiveUpdates.exists { dep =>
+        val depStamp = dep.stats.stamp
+        val ourStamp = out.stats.stamp
+        // If dependency was freshly resolved in this command
+        !dep.stats.cached ||
+        // Or if dependency has a newer stamp than us (across commands)
+        // Stamps are currently timestamps, so lexicographic comparison works
+        (depStamp.nonEmpty && depStamp > ourStamp)
+      }
```
eed3si9n: I suggest removing this logic here, and fold transitive dependency check into `UpdateInputs`.

See https://github.com/sbt/sbt/blob/add43bd23031d50ac72ad25add82fd5bc3b5d569/main/src/main/scala/sbt/internal/LibraryManagement.scala#L189

and

https://github.com/sbt/sbt/blob/add43bd23031d50ac72ad25add82fd5bc3b5d569/main/src/main/scala/sbt/internal/LibraryManagement.scala#L29

That tuple represents `updateInputs`. You can probably just add fourth item in there that is a `Vector[String]` that represents the transitive stamps, and the existing code should hopefully do the invalidation for you.

eed3si9n (commented):
Added some comment on the implementation, but overall this is a cool change!

calm329:
@eed3si9n Could you please review changes again?

eed3si9n (approved):
Thanks!

sbt-app/src/sbt-test/dependency-management/i8357-transitive-update-invalidation/test
```diff
@@ -0,0 +1,21 @@
...
+$ copy-file changes/build.sbt build.sbt
+
+# Step 3: Reload to pick up the new settings
+> reload
+
+# Step 4: Run a/update in a SEPARATE command
+> a/update
+
+# Step 5: Now run itTests/update - should re-resolve with new dependencies
+> show itTests/ourStamp
```
OlegYch, line 17: i'd suggest adding an actual test, eg
```
dothetest := assert((Test / managedClasspath).value.toString.contains("cats-core_2.12-2.9.0")),
```

sbt-app/src/sbt-test/dependency-management/i8357-transitive-update-invalidation/test
```diff
@@ -0,0 +1,21 @@
+# Test for https://github.com/sbt/sbt/issues/8357
+# Verifies that transitiveUpdate correctly invalidates across command invocations
+
+# Step 1: Run itTests/update to establish baseline with cats 2.8.0
+> itTests/update
+
+# Step 2: Change a's dependency version (cats 2.8.0 -> 2.9.0)
+$ copy-file changes/build.sbt build.sbt
```
OlegYch: it would be great to set the `catsVersion` setting here instead of overwriting the file, otherwise it is redundant

lm-core/src/main/contraband-scala/sbt/librarymanagement/UpdateStats.scala
```diff
@@ -4,26 +4,28 @@
...
+  private def this(resolveTime: Long, downloadTime: Long, downloadSize: Long, cached: Boolean) = this(resolveTime, downloadTime, downloadSize, cached, None)
   
   override def equals(o: Any): Boolean = this.eq(o.asInstanceOf[AnyRef]) || (o match {
-    case x: UpdateStats => (this.resolveTime == x.resolveTime) && (this.downloadTime == x.downloadTime) && (this.downloadSize == x.downloadSize) && (this.cached == x.cached)
+    case x: UpdateStats => (this.resolveTime == x.resolveTime) && (this.downloadTime == x.downloadTime) && (this.downloadSize == x.downloadSize) && (this.cached == x.cached) && (this.stamp == x.stamp)
     case _ => false
   })
   override def hashCode: Int = {
-    37 * (37 * (37 * (37 * (37 * (17 + "sbt.librarymanagement.UpdateStats".##) + resolveTime.##) + downloadTime.##) + downloadSize.##) + cached.##)
+    37 * (37 * (37 * (37 * (37 * (37 * (17 + "sbt.librarymanagement.UpdateStats".##) + resolveTime.##) + downloadTime.##) + downloadSize.##) + cached.##) + stamp.##)
```
vadzim-superdone, line 22: tut

PR #8504, opened 2026-01-12
[2.x] feat: Add csrLocalArtifactsShouldBeCached setting for caching local artifacts

### LINKED ISSUE #7547: sbt-coursier does not enable configuration of the `FileCache.localArtifactsShouldBeCached` setting
For launching the new scala command (SIP 46), we want to cache the compiler artefacts in a local repo that is bundled in the zip, so Scala CLI can resolve the artefacts from here. This isn't possible to do within sbt from the librarymanagement API because there is no configuration to allow caching of local artefacts.

what happens is coursier will resolve all library dependencies and put them in the cache, but anything in local repo is skipped. There's a setting `localArtifactsShouldBeCached` which should prevent skipping, but it is only configurable from Coursier CLI.

refs: see https://github.com/scala/scala3/issues/20098

grnlemon:
# Add csrLocalArtifactsShouldBeCached setting

## Problem

Currently, sbt-coursier does not expose Coursier's `FileCache.localArtifactsShouldBeCached` setting. This makes it impossible for users to configure whether local `file://` artifacts should be copied to the cache directory.

This is particularly problematic for scenarios like SIP 46 / Scala CLI integration, where compiler artifacts are bundled in a local repository for offline use. Without this setting, users cannot ensure that local artifacts are properly cached.

## Solution

This PR adds a new sbt setting `csrLocalArtifactsShouldBeCached` that controls whether local file artifacts should be cached by Coursier.

### Usage

```scala
// In build.sbt
csrLocalArtifactsShouldBeCached := true
```

When enabled, artifacts from local `file://` repositories will be copied to the Coursier cache directory, making them available for offline use and consistent with remote artifact handling.

### Changes

- Added `localArtifactsShouldBeCached` field to `CoursierConfiguration`
- Added `csrLocalArtifactsShouldBeCached` setting key in sbt
- Wired the setting through to Coursier's `FileCache` configuration
- Added scripted test to verify the feature

Fixes #7547

---

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -716,6 +716,7 @@ lazy val mainProj = (project in file("main"))
     mimaSettings,
     mimaBinaryIssueFilters ++= Vector(
       exclude[DirectMissingMethodProblem]("sbt.internal.ConsoleProject.*"),
+      exclude[DirectMissingMethodProblem]("sbt.coursierint.LMCoursier.coursierConfiguration"),
     ),
   )
   .dependsOn(lmCore, lmIvy, lmCoursierShadedPublishing)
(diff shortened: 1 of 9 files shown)


eed3si9n (approved):
lgtm. thanks for the contribution!

PR #8507, opened 2026-01-13
[2.x] fix: Use strict matching for scala-library jar detection

### LINKED ISSUE #7511: Packages including "scala-library" in its name cannot be imported by other projects.
## steps
Trying to create a package with a name including the string `scala-library` in it, results in a failure when trying to use that package from another project.

<details><summary>Details</summary>
<p>

A "library" project defined as
```scala
// build.sbt
import sbt.file

lazy val root =
  Project(id = "my-scala-library", base = file("."))
    .settings(
      name := "my-scala-library-that-fails",
      scalaVersion := "2.13.12",
      organization := "com.myorg",
    )
```
with just one file
```scala
//  src/main/scala/com/myorg/Lib.scala
package com.myorg

object Lib {
  def MyFunc() = 1
}
```
and a project trying to use it
```scala
//  build.sbt
import sbt.file

lazy val foo =
  Project(id = "foo", base = file("."))
    .settings(
      name := "foo",
      organization := "com.myorg",
      scalaVersion := "2.13.12",
      libraryDependencies ++= Seq(
        "com.myorg" %% "my-scala-library-that-fails" % "0.1.0-SNAPSHOT",
      ),
      resolvers += Resolver.mavenLocal
    )
```
```scala
// src/main/scala/foo/Foo.scala
package foo

import com.myorg.Lib

object Foo {
  def main(args: Array[String]) = {
    println(Lib.MyFunc())
  }
}
```

And running 
```
sbt publishM2
```
on the library, and 
```
sbt compile
```
on the consumer results in a failure

</p>
</details> 

[repro.tar.gz](https://github.com/sbt/sbt/files/14574815/repro.tar.gz)

## problem
```
❯ sbt compile
[info] welcome to sbt 1.9.9 (Homebrew Java 21.0.2)
[info] loading project definition from .../consumer/project
[info] loading settings for project foo from build.sbt ...
[info] set current project to foo (in build file:.../consumer/)
[info] Executing in batch mode. For better performance use sbt's shell
[info] compiling 1 Scala source to .../consumer/target/scala-2.13/classes ...
[error] .../consumer/src/main/scala/foo/Foo.scala:3:12: object myorg is not a member of package com
[error] import com.myorg.Lib
[error]            ^
[error] .../consumer/src/main/scala/foo/Foo.scala:7:13: not found: value Lib
[error]     println(Lib.MyFunc())
[error]             ^
[error] two errors found
```
## expectation
I'm able to import symbols from the package
## notes
I tried looking if this behavior is intended, and could not find anything.
If the package name for the library is something else like `my-scala-lib-that-fails`, then it works without a problem.

grnlemon:
# Fix: Strict matching for scala-library jar detection

## Problem

If you name your library something like `my-scala-library-utils`, sbt won't be able to find it at compile time. The build fails with cryptic "object not found" errors even though the jar is clearly downloaded and present.

This happens because sbt uses `name.contains("scala-library")` to detect the Scala standard library jar. Any jar with "scala-library" anywhere in its filename gets misclassified and filtered into a special classloader layer meant only for the actual Scala stdlib.

## The Fix

Changed from substring matching to strict matching:

```scala
// Before (buggy):
name.contains("scala-library")

// After:
name == "scala-library.jar" || name.startsWith("scala-library-")
```

This matches the pattern already used for `scala-reflect` detection in the same file

Fixes #7511

---

### DIFF
--- a/main/src/main/scala/sbt/internal/ClassLoaders.scala
+++ b/main/src/main/scala/sbt/internal/ClassLoaders.scala
@@ -182,8 +182,9 @@ private[sbt] object ClassLoaders {
               cpFiles
                 .filter(f => {
                   val name = f.getName
-                  name.contains(ArtifactInfo.ScalaLibraryID) || si.libraryJars
-                    .exists(_.getName == name)
+                  name == s"${ArtifactInfo.ScalaLibraryID}.jar" ||
+                  name.startsWith(s"${ArtifactInfo.ScalaLibraryID}-") ||
+                  si.libraryJars.exists(_.getName == name)
                 })
                 .toArray
             else si.libraryJars


mkdev5:
All passed! please review the changes and approve it. thanks.

eed3si9n (approved):

PR #8508, opened 2026-01-13
[2.x] fix: Fix StackOverflowError when reporting self-referencing exceptions

### LINKED ISSUE #7509: sbt throws a `StackOverflowError` when reporting a self-referencing exception.
## steps

Use [this reproducer](https://github.com/facewindu/sbt-stackoverflow) and execute `sbt failingTask`.

## problem

A `java.lang.StackOverflowError` is thrown. This could be avoided by handling the recursion in the `Throwable` cause, similarly to how the JDK handles this when printing stacktraces.
```
// Guard against malicious overrides of Throwable.equals by
// using a Set with identity equality semantics.
Set<Throwable> dejaVu = Collections.newSetFromMap(new IdentityHashMap<Throwable, Boolean>());

...

if (dejaVu.contains(this)) {
  s.println(prefix + caption + "[CIRCULAR REFERENCE: " + this + "]"); // <---- This gets the code out of the recursion
}
```

## expectation

The build is resilient to such `Throwable`s.

grnlemon:
# Fix StackOverflowError when reporting self-referencing exceptions

## Problem

When an exception has a circular reference in its cause chain (e.g., `exception.getCause() == exception`), sbt throws a `StackOverflowError` while trying to report the exception. This happens because `StackTrace.trimmedLines` traverses the cause chain without checking for circular references.

This is problematic because some libraries or frameworks may create exceptions with self-referencing causes, and sbt should be resilient to such edge cases.

## Solution

This PR adds circular reference detection to `StackTrace.trimmedLines` using an `IdentityHashMap`-backed `Set`, similar to how the JDK handles this in `Throwable.printStackTrace()`.

When a circular reference is detected, instead of recursing infinitely, the method now appends a `[CIRCULAR REFERENCE: ...]` message to indicate the cycle.

### Changes

- Modified `StackTrace.trimmedLines` to track seen exceptions using `IdentityHashMap`
- Added check to detect circular references before processing each cause
- When circular reference is detected, append informative message instead of recursing
- Added unit tests for:
  - Normal exceptions
  - Exceptions with causes
  - Self-referencing exceptions
  - Circular exception chains

### Example Output

For a self-referencing exception, the output now looks like:
```
java.lang.RuntimeException: self-referencing exception
	at ...
[CIRCULAR REFERENCE: java.lang.RuntimeException: self-referencing exception]
```

Fixes #7509

---

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/StackTrace.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/StackTrace.scala
@@ -10,6 +10,7 @@ package sbt.internal.util
 
 import sbt.io.IO
 import scala.collection.mutable.ListBuffer
+import java.util.{ IdentityHashMap, Collections }
 
 object StackTrace {
   def isSbtClass(name: String) = name.startsWith("sbt.") || name.startsWith("xsbt.")
@@ -30,6 +31,8 @@ object StackTrace {
   def trimmedLines(t: Throwable, d: Int): List[String] = {
     require(d >= 0)
     val b = new ListBuffer[String]()
+    val seen: java.util.Set[Throwable] =
+      Collections.newSetFromMap(new IdentityHashMap[Throwable, java.lang.Boolean]())
 
     def appendStackTrace(t: Throwable, first: Boolean): Unit = {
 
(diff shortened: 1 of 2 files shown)


mkdev5:
@eed3si9n all checks passed, please review the PR and let me know your feedback. thanks.

eed3si9n (approved):
Thanks!

Issue #2695, opened 2016-08-09
equal (=) appearing passed in as batch argument requires quotation on Windows


hossameldeen:
## steps
- Installed sbt 0.13.8 on Windows 7 64-bit.
- Went to a directory that has a simple hello-world sbt project (probably doesn't matter).
- From the commandline, called: `sbt -Dsbt.log.noformat=true`.
## problem
- It says: Warning: invalid system property 'sbt.log.noformat'
## expectation
- To not show this message, because the documentation seems to say it's a valid property.
  http://www.scala-sbt.org/0.13/docs/Command-Line-Reference.html
## notes
- OS: Windows 7 64-bit.
- sbt: 0.13.8
- Screenshot showing the error and showing sbt version
  ![sbt-log-noformat-error](https://cloud.githubusercontent.com/assets/6719916/17526074/f8c4776a-5e65-11e6-929b-dca6ef2279e6.PNG)

## workaround

```
sbt "-Dsbt.log.noformat=true"
```

eed3si9n:
Does this still happen on the latest sbt 0.13.12?

hossameldeen:
Yes, it does. The steps I just did to check:
- Removed sbt's installation from Control Panel.
- Downloaded `sbt-0.13.12.msi` from [here](http://www.scala-sbt.org/download.html)
- Created an empty directory and called `sbt -Dsbt.log.noformat=true` in it. The same problem occurred. (But kept downloading stuff for some time since this is the first time in the new installation).
- Then, called `sbt sbtVersion` to make sure of the installed version.

Screenshot:
![sbt-log-noformat-12](https://cloud.githubusercontent.com/assets/6719916/17529187/af96c284-5e72-11e6-802f-b870c8a2a82d.PNG)

Note: The screenshot is for calling `sbt -Dsbt.log.noformat=true` for a second time after the first time that downloaded many stuff. That's why there're no `downloading..` lines. And to confirm: The error message showed in both the first and second calls.

kpbochenek:
you have `>sbt -Dsbtlog.noformat=true` (you missed dot after sbt)

eed3si9n:
I'm guessing it's a capability gap between the shell script shipped for Mac and Linux vs this batch script - https://github.com/sbt/sbt-launcher-package/blob/e4e3e54c33a4b26e3e0a82270581c9038bae9ddc/src/windows/sbt

hossameldeen:
[...]

PR #8509, opened 2026-01-13
[2.x] fix: Fix sbt new argument parsing on Windows

### LINKED ISSUE #7507: `sbt new` fails to parse arguments correctly on Windows
## steps

Replicated on sbt 1.9.9, running Windows Terminal (PowerShell).

1. Open command line on Windows
1. Run `sbt new foundweekends/giter8.g8 --name=example.g8 --giter8_version=0.16.2 --sbt_version=1.9.9`

## problem

Command does not run successfully, error is throw:
```
Error: Unknown option --name
Error: Unknown argument 'example.g8'
Error: Unknown option --giter8_version
Error: Unknown argument '0.16.2'
Error: Unknown option --sbt_version
Error: Unknown argument '1.9.8'
Try --help for more information.
```

Command does work on Linux, tested on WSL (Ubuntu).

## expectation

sbt project should be created from the template successfully with arguments passed to the template.

## notes

After experimenting with variations of command, one version I found to work was `sbt new foundweekends/giter8.g8 "--name=example.g8 " "--giter8_version=0.16.2 " "--sbt_version=1.9.9 "` - arguments must be enclosed in quotes and must be terminated by a space. 
One strange variation was that enclosing all arguments in quotes -  `sbt new foundweekends/giter8.g8 "--name=example.g8 --giter8_version=0.16.2 --sbt_version=1.9.9"` - created the sbt project, but with only the `name` argument set to `example.g8 --giter8_version=0.16.2 --sbt_version=1.9.9`, meaning that sbt project's directory would end up containing the remaining part of argument string. 
This does not seem intentional, likely a bug in parser module.

grnlemon:
# Fix sbt new argument parsing on Windows

This is a long-standing Windows issue where arguments containing `=` get split by the command line parser.

When running something like:
```
sbt new foundweekends/giter8.g8 --name=example.g8
```

Windows splits `--name=example.g8` into two separate arguments (`--name` and `example.g8`), causing sbt to fail with "Unknown option" errors.

The batch script already handles this for `-D` and `-XX` JVM arguments by detecting when an argument was split and recombining it. This PR extends that same approach to `--` style arguments used by giter8 templates.

Related to #2695.

Fixes #7507

---

### DIFF
--- a/launcher-package/src/universal/bin/sbt.bat
+++ b/launcher-package/src/universal/bin/sbt.bat
@@ -547,6 +547,21 @@ if not "%g:~0,5%" == "-XX:+" if not "%g:~0,5%" == "-XX:-" if "%g:~0,3%" == "-XX"
   )
 )
 
+if defined sbt_new if "%g:~0,2%" == "--" (
+  rem special handling for -- template arguments since '=' gets parsed away on Windows
+  for /F "tokens=1 delims==" %%a in ("%g%") do (
+    rem make sure it doesn't have the '=' already
+    if "%g%" == "%%a" (
+      if not "%~1" == "" (
+        call :dlog [args_loop] -- argument %~0=%~1
+        set "SBT_ARGS=!SBT_ARGS! %~0=%~1"
+        shift
+        goto args_loop
+      )
+    )
+  )
+)
+
 rem the %0 (instead of %~0) preserves original argument quoting
 set SBT_ARGS=!SBT_ARGS! %0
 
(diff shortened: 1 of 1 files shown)


eed3si9n:
CI hasn't finished in 1h, so maybe there's an infinite loop now?

mkdev5:
> CI hasn't finished in 1h, so maybe there's an infinite loop now?

Fixed by adding `if defined sbt_new` condition so the `--` argument recombination only applies after `new/init` command. The previous version was too broad and could cause issues with other `--` flags

mkdev5:
the failure is not an actual error. this is a known quirk of sbt's output formatting

mkdev5:
All good!

PR #8512, opened 2026-01-13
[2.x] fix: Skip native client for sbt new/init commands

### LINKED ISSUE #7497: SBT native `new` command crashes with any template
## Environment

- Linux (Ubuntu): x86_64

## Steps

Run this command:
```bash
sbt --client new scala/scala3.g8
```

## Problem

SBT fails to create a project and crashes with the following error:

```bash
[info] welcome to sbt 1.9.8 (Oracle Corporation Java 21.0.2)
[info] loading project definition from /home/owner/project
[info] set current project to owner (in build file:/home/owner/)
[info] sbt server started at local:///home/owner/.sbt/1.0/server/f2487b1ab632c3668137/sock
> new scala/scala3.g8
Error: Unknown argument 'sbtCompleteExec f7ba6c33-5a01-4a6b-9160-1f9bb4be972c'
Error: Unknown argument 'resumeFromFailure'
Error: Unknown argument 'sbtPopOnFailure'
Error: Unknown argument 'sbtReportResult f7ba6c33-5a01-4a6b-9160-1f9bb4be972c'
Error: Unknown argument 'shell'
Try --help for more information.
[info] disconnected
```

## Expectation

A new project should be created from the template.

## Notes

The problem seems to be occurring for all Giter8 templates including non-existent templates. For example this command will trigger the same error:

```bash
sbt --client new not/real.g8
```

grnlemon:
Fixes #7497

When running 'sbt --client new scala/scala3.g8', the thin client adds internal commands (sbtCompleteExec, resumeFromFailure, sbtPopOnFailure, sbtReportResult, shell) to the command queue. The new command was appending all remainingCommands to the template arguments, causing Giter8 to fail with 'Unknown argument' errors.

Now filters out these internal sbt commands before passing arguments to the template resolver.

Added unit tests for the filtering logic.

---

### DIFF
--- a/launcher-package/src/universal/bin/sbt.bat
+++ b/launcher-package/src/universal/bin/sbt.bat
@@ -926,6 +926,11 @@ for /f "delims=.-_ tokens=1-2" %%v in ("!JAVA_VERSION!") do (
 
 rem parse the first two segments of sbt.version and set run_native_client to
 rem 1 if the user has also indicated they want to use native client.
+rem sbt new/init should not use native client as it needs to run outside a project
+if defined sbt_new (
+  set run_native_client=
+  exit /B 0
+)
 set sbtV=!build_props_sbt_version!
 set sbtBinaryV_1=
 set sbtBinaryV_2=
--- a/sbt
+++ b/sbt
@@ -805,6 +805,11 @@ detectNativeClient() {
 
 # Run native client if build.properties points to 1.4+ and has SBT_NATIVE_CLIENT
 isRunNativeClient() {
+  # sbt new/init should not use native client as it needs to run outside a project
+  if [[ "$sbt_new" == "true" ]]; then
+    echo "false"
+    return
+  fi
   sbtV="$build_props_sbt_version"
   [[ "$sbtV" == "" ]] && sbtV="$init_sbt_version"
   [[ "$sbtV" == "" ]] && sbtV="0.0.0"


eed3si9n:
> Now filters out these internal sbt commands before passing arguments to the template resolver.

This means that sbtn is will launch an sbt session in the current directory. Instead, if `sbt new` is detected on bash/batch (search for `sbt_new=true`), we should probably ignore `--client`, and call the "server", which inside of `Main.scala` is going to short-circuit and does special thing for `new` anyway.

mkdev5:
> > Now filters out these internal sbt commands before passing arguments to the template resolver.
> 
> This means that sbtn is will launch an sbt session in the current directory. Instead, if `sbt new` is detected on bash/batch (search for `sbt_new=true`), we should probably ignore `--client`, and call the "server", which inside of `Main.scala` is going to short-circuit and does special thing for `new` anyway.

Thanks for the suggestion! You're right that handling this at the launcher level is cleaner.

I've updated the PR to skip the native client entirely when `sbt new` or `sbt init` is detected. The fix adds an early return in `isRunNativeClient()` (bash) and the equivalent check in `sbt.bat` (Windows) - when `sbt_new=true`, we just use the full JVM instead of the thin client.

This avoids the whole issue of internal commands leaking into the template arguments since the native client path is never taken for these commands.

eed3si9n (approved):

mkdev5:
the CI failure appears to be unrelated to this PR. the `NotImplementedError` is from a scripted test that uses `???` in its `build.sbt` to test error handling - that's expected behavior. the launcher script changes only affect the execution path for `sbt new`/`sbt init` commands and don't impact the server tests

mkdev5:
All passed!

PR #8517, opened 2026-01-13
[2.x] fix: Fixes --no-colors setting for sbtn

### LINKED ISSUE #7663: [2.x] Distinguish between client and server options
Most sbt command options where added before sbt server landing. But now that we have two types of process, clients and servers, we should clearly distinguish between client and server options. For instance `--no-color` is a client option. One client may want colors (a shell) and another may not want them (BSP).

It should fix https://github.com/sbt/sbt/issues/6468

gayanMatch:
# Fix client options like --no-colors being ignored

## Problem

When running sbt with the client-server mode, client options like `--no-colors` are ignored. The server uses its own color settings instead of respecting each client's preferences.

Related issues: #7663, #6468

## Solution

In `Terminal.TerminalImpl.doWrite()`, the code was using global singleton properties (`Terminal.isAnsiSupported`, `Terminal.isColorEnabled`) instead of the instance's properties (`self.isAnsiSupported`, `self.isColorEnabled`).

The `NetworkTerminal` correctly fetches the client's color settings via `terminalPropertiesQuery`, but these settings were never used when formatting output because `doWrite()` checked the global singleton.

### Changes

- Modified `Terminal.TerminalImpl.doWrite()` to use `self.isColorEnabled` and `self.isAnsiSupported` instead of `Terminal.isColorEnabled` and `Terminal.isAnsiSupported`
- Added unit tests to verify terminal instances use their own color settings

Fixes #7663

---

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=132382032

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
@@ -1034,11 +1034,11 @@ object Terminal {
     private def doWrite(rawBytes: Array[Byte]): Unit = withPrintStream { ps =>
       val (toWrite, len) =
         if (rawBytes.contains(27.toByte)) {
-          if (!Terminal.isAnsiSupported || !Terminal.isColorEnabled)
+          if (!self.isAnsiSupported || !self.isColorEnabled)
             EscHelpers.strip(
               rawBytes,
-              stripAnsi = !Terminal.isAnsiSupported,
-              stripColor = !Terminal.isColorEnabled
+              stripAnsi = !self.isAnsiSupported,
+              stripColor = !self.isColorEnabled
             )
           else (rawBytes, rawBytes.length)
         } else (rawBytes, rawBytes.length)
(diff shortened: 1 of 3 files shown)


gayanMatch:
@eed3si9n Could you please review my PR? And I'm not sure why one workflow run got failed. I think it's not related to my change, could you re-run it?

eed3si9n:
Does each terminal have its own color setting? afaik sbtn just sends each character to the sbt server, like ssh.

https://github.com/sbt/sbt/blob/develop/contributing-docs/01_pull_request.md#instruction-to-build-sbtn has an instruction on how to build sbtn locally. Could you post a screenshot of a local build with color turned on or off different from that of the server?

gayanMatch:
> Does each terminal have its own color setting? afaik sbtn just sends each character to the sbt server, like ssh.
> 
> https://github.com/sbt/sbt/blob/develop/contributing-docs/01_pull_request.md#instruction-to-build-sbtn has an instruction on how to build sbtn locally. Could you post a screenshot of a local build with color turned on or off different from that of the server?

Here are the results:
- with colors:
image

- without colors:
image

eed3si9n (approved):
Thanks!

PR #8532, opened 2026-01-14
[2.x] fix: Cleanup child processes on sbt exit

### LINKED ISSUE #7468: SBT should cleanup child processes on exit
## steps

Start SBT shell, set `fork := true` and run a process.
SBT will spin a new process, example:
`java @/var/folders/nn/tyk6xz5x48n00bnd61t64hhh0000gp/T/sbt-args2954607292963552134.tmp`
Click CTRL+C to cancel the process.
If the process doesn't exit, it will become a stale process owned by the sbt process parent.
Run `exit` in the sbt shell.
SBT exists, while the stale process keeps running.

## problem

SBT doesn't cleanup stale child processes.

## expectation

On exit, SBT should search all child processes and kill them with `kill -9 $id`

## notes

Workaround `ps -ef | grep -i sbt-args` and kill the stale process with `kill -9 $id`

grnlemon:
# Cleanup child processes on sbt exit

## Description

When using `fork := true`, sbt spawns child processes that may become stale if cancelled (e.g., via Ctrl+C). Previously, these processes were not cleaned up when running the `exit` command - they would keep running as orphans.

This fix adds `RunningProcesses.killAll()` to the shutdown hook so that all tracked forked processes are terminated when sbt exits.

## Steps to reproduce (before fix)

1. Start sbt shell with `fork := true`
2. Run a long-running forked process
3. Press Ctrl+C to cancel (process may not exit immediately)
4. Run `exit` in sbt shell
5. The stale child process keeps running

## Changes

- Call `RunningProcesses.killAll()` in the `closeRunnable` shutdown hook in `Main.scala`

## Testing

The fix uses existing, tested infrastructure (`RunningProcesses.killAll()`) that is already used for task cancellation.

Fixes #7468

---

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -29,7 +29,7 @@ import sbt.internal.server.{ BuildServerProtocol, NetworkChannel }
 import sbt.internal.util.Terminal.hasConsole
 import sbt.internal.util.Types.{ const, idFun }
 import sbt.internal.util.complete.Parser
-import sbt.internal.util.{ Terminal as ITerminal, * }
+import sbt.internal.util.{ RunningProcesses, Terminal as ITerminal, * }
 import sbt.io.*
 import sbt.io.syntax.*
 import sbt.util.{ Level, Logger, Show }
@@ -221,6 +221,7 @@ object StandardMain {
   })
 
   private val closeRunnable = () => {
+    RunningProcesses.killAll()
     exchange.shutdown()
     pool.foreach(_.shutdownNow())
   }


mkdev5:
@eed3si9n please have a look at the PR and let me know the feedback

main/src/main/scala/sbt/Main.scala
```diff
@@ -221,6 +221,7 @@ object StandardMain {
   })
 
   private val closeRunnable = () => {
+    RunningProcesses.killAll()
```
eed3si9n, line 224: I don't think this would hurt, but do we have confirmation that this makes any difference? For example, do you see difference in `jps` output with this change vs 2.0.0-RC8?

eed3si9n (approved):

main/src/main/scala/sbt/Main.scala
mkdev5: When you run a forked process, sbt tracks it in `RunningProcesses` and `blockForExitCode` waits for it to finish. If you press Ctrl+C, `cancelAndShutdown()` calls `destroy()` on the process, but some processes don't die right away.

While the process is still running, `blockForExitCode` is still blocking on `exitValue()`, and the process remains in `RunningProcesses.active`. If you type `exit` at this point, sbt would shut down but the child process would become orphaned.

This fix adds `RunningProcesses.killAll()` to the shutdown hook, so any processes still being tracked get terminated when sbt exits.

To verify with `jps`: run a forked process that ignores SIGTERM, press Ctrl+C, then `exit`. Without the fix, `jps` shows the orphan. With the fix, it's cleaned up.

mkdev5:
can you merge the PR?

PR #8549, opened 2026-01-15
[2.x] test: Migrate ProgressStateSpec.scala to verify.BasicTestSuite

GlobalStar117:
## Summary
Migrates `ProgressStateSpec.scala` from ScalaTest's `AnyFlatSpec` with `BeforeAndAfterAll` to `verify.BasicTestSuite`, following the pattern established by other test files in the sbt codebase.

Related to the ongoing test migration effort.

## Changes
- Replace `AnyFlatSpec` class with `BasicTestSuite` object
- Remove `BeforeAndAfterAll` trait and convert `afterAll` cleanup to `try-finally` block
- Use `scala.util.Using.resource` for proper resource management
- Convert `"..." should "..." in` syntax to `test("...")` syntax
- Use Scala 3 syntax with colon-based indentation
- Change `===` to `==` for assertions
- Add `end ProgressStateSpec` marker

## Before
```scala
class ProgressStateSpec extends AnyFlatSpec with BeforeAndAfterAll {
  private lazy val fileIn = new File("/tmp/tmp.txt")
  private lazy val fileOut = Source.fromFile("/tmp/tmp.txt")

  override def afterAll(): Unit = {
    fileIn.delete()
    fileOut.close()
  }

  "test" should "not clear after carriage return" in {
    // ...
  }
}
```

## After
```scala
object ProgressStateSpec extends BasicTestSuite:
  test("test should not clear after carriage return"):
    val fileIn = new File("/tmp/tmp.txt")
    try
      // ...
    finally
      fileIn.delete()
end ProgressStateSpec
```

### DIFF
--- a/internal/util-logging/src/test/scala/sbt/internal/util/ProgressStateSpec.scala
+++ b/internal/util-logging/src/test/scala/sbt/internal/util/ProgressStateSpec.scala
@@ -10,34 +10,30 @@ package sbt.internal.util
 
 import java.io.{ File, PrintStream }
 
-import org.scalatest.flatspec.AnyFlatSpec
-import org.scalatest.BeforeAndAfterAll
+import verify.BasicTestSuite
 import sbt.internal.util.Terminal.SimpleTerminal
 
 import scala.io.Source
+import scala.util.Using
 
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #8556, opened 2026-01-16
[2.x] feat: Show message when sbtn is waiting for busy sbt server

### LINKED ISSUE #8356: Your call is important to us
This was originally reported on Discord.

## steps

Make sbtn call while running a long-running task like `console`

## problem

It blocks.

## expectation

Some way to tell that it's being blocked by an existing command. Maybe display a message on stderr.

GlobalStar117:
## Summary

When the sbtn thin client sends a command to a server that is already executing another command (like `console`), the client would block silently without any feedback to the user.

This change adds a message to stderr after 2 seconds of waiting:

```
[info] waiting for server to become available...
[info] (hint: another command may be running, e.g., console or shell)
```

The message is printed once per command to avoid spamming the terminal.

## Changes

- Modified `sendAndWait` in `NetworkClient.scala` to use `poll` with a timeout instead of blocking indefinitely with `take`
- After 2 seconds of waiting, prints a helpful message to stderr
- Message is only printed once per command

## Testing

Manual testing:
1. Start sbt server and run `console`
2. In another terminal, run `sbt --client compile`
3. After 2 seconds, message appears: `[info] waiting for server to become available...`

Fixes https://github.com/sbt/sbt/issues/8356

---
This is a Gittensor contribution.
gittensor:user:GlobalStar117

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -1015,11 +1015,31 @@ class NetworkClient(
   private def sendAndWait(cmd: String, limit: Option[Deadline]): Int = {
     val queue = sendExecCommand(cmd)
     var result: Integer = null
+    var waitMessagePrinted = false
+    val initialWaitMs = 2000L // Wait 2 seconds before showing the "waiting" message
+    val pollIntervalMs = 500L // Poll interval after initial wait
     while (running.get && result == null && limit.fold(true)(!_.isOverdue())) {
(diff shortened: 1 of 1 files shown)


eed3si9n:
Thanks for the contribution. Note that if you're participating in the Gittensor, you'd need to include the tag in the PR description.

eed3si9n (approved):
Thanks!

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/8568, which got merged yesterday.

PR #8559, opened 2026-01-16
feat: Add sbt.credentials system property for credentials file

### LINKED ISSUE #576: Load credentials file defined in System property
I would like to see a standard system property for pointing to a credentials _file_ that is used during boot and project updates, and publish. This is convenient since you often need to externalize the credentials anyway. Also useful when using a corporate proxy with credentials to load everything.

GlobalStar117:
This change adds support for specifying a credentials file via the `sbt.credentials` system property (e.g., `-Dsbt.credentials=/path/to/file`).

Previously, credentials could only be loaded from:
- SBT_CREDENTIALS environment variable
- sbt configuration files

This enables users to:
- Specify credentials at runtime via system property
- Unify credential specification between sbt-launcher and sbt
- Better support corporate proxy configurations

Fixes #576

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=146701565

### DIFF
--- a/main/src/test/scala/sbt/internal/SysPropCredentialsSpec.scala
+++ b/main/src/test/scala/sbt/internal/SysPropCredentialsSpec.scala
@@ -1,0 +1,52 @@
+/*
+ * sbt
+ * Copyright 2023, Scala center
+ * Copyright 2011 - 2022, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.internal
+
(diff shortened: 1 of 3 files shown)


main/src/test/scala/sbt/internal/SysPropCredentialsSpec.scala
eed3si9n, line 1: Normally I appreciate unit tests, but I don't want to accidentally leak credentials, so could we remove this test please? Otherwise the change looks great.

eed3si9n (changes requested):

PR #8561, opened 2026-01-16
fix: Windows sbt launcher script handles system property without value

### LINKED ISSUE #7332: Windows sbt launcher script doesn't handle system property without value

GlobalStar117:
## Summary

Fixes #7332

On Windows, passing a system property without a value (e.g., `sbt -Dfoo`) would fail with "is missing a value" error, while the same command works fine on macOS.

## Problem

The Windows sbt.bat script's `-D` argument handling assumed that if there's no `=` in the argument, the next argument must be the value. This fails for properties without values like `-Dfoo`.

## Solution

Modified the `-D` handling logic to:
1. When `-Dfoo` is passed without `=`, check if the next argument starts with `-`
2. If the next arg is another option (starts with `-`), treat the current arg as a property without value
3. If there are no more arguments, also treat it as a property without value

This matches the behavior on macOS where `-Dfoo` sets the property 'foo' to an empty string (or rather, defines it without a value).

## Changes

- `launcher-package/src/universal/bin/sbt.bat`: Updated `-D` handling in args_loop and echolist functions

## Testing

After this change:
```bash
sbt -Dfoo
# Works! sbt starts and sys.props.contains("foo") returns true

sbt -Dfoo -Dbar=baz
# Works! Both properties are set correctly
```

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=146701565


eed3si9n (changes requested):
```

[info] - sbtopts handles JVM parameters with spaces *** FAILED ***
[info]   RuntimeException: Nonzero exit value: 255
[info]     scala.sys.process.ProcessBuilderImpl$AbstractBuilder.slurp(ProcessBuilderImpl.scala:164)
[info]     scala.sys.process.ProcessBuilderImpl$AbstractBuilder.$bang$bang(ProcessBuilderImpl.scala:121)
[info]     example.test.RunnerScriptTest$.$init$$$anonfun$25(RunnerScriptTest.scala:201)
```

Looks like the test from another PR is failing. Could you isolate the fix for this PR please and see if that would work?

eed3si9n:
I am closing this due to inactivity.

PR #8562, opened 2026-01-16
[2.x] test: migrate SonaClientTest from ScalaTest to verify.BasicTestSuite

GlobalStar117:
- Convert class to object with BasicTestSuite
- Use test() syntax instead of 'should ... in'
- Move RecordingLogger and extension to top-level
- Use Scala 3 colon indentation syntax

---
gittensor::146701565


sbt-app/src/sbt-test/lm-coursier/credentials-system-property/build.sbt
eed3si9n, line 1: Could you remove this please?

eed3si9n (approved):
Thanks!

PR #8572, opened 2026-01-18
fix: create missing project directory for nested builds

### LINKED ISSUE #8570: The sbt fails with multi-build setup and `addPluginSbtFile`
## steps

1. Unzip the [project](https://github.com/user-attachments/files/24691798/bug2Projects.zip). It contains an additional sbt file in the `temp` directory. 
2. Open the terminal inside the project root. 
3. Execute `sbt --addPluginSbtFile=temp/file.sbt`

**Expected result**: the sbt starts 
**Actual result**:  the sbt fails with error message: 
```
java.lang.RuntimeException: Invalid build URI (no handler available): file:~/bug2Projects/c1/project/
        at scala.sys.package$.error(package.scala:30)
        at sbt.internal.Load$.$anonfun$builtinLoader$1(Load.scala:490)
        at sbt.internal.BuildLoader.$anonfun$apply$3(BuildLoader.scala:245)
        at scala.Option.getOrElse(Option.scala:189)
        at sbt.internal.BuildLoader.apply(BuildLoader.scala:245)
        at sbt.internal.Load$.loadURI$1(Load.scala:555)
        at sbt.internal.Load$.loadAll(Load.scala:571)
        at sbt.internal.Load$.loadURI(Load.scala:501)
        at sbt.internal.Load$.load(Load.scala:480)
```
4. Create an empty `project` folder inside the `c1` directory. 
5. Repeat step 3. 

**Expected & actual result**: the sbt starts 

## problem
It seems that the absence of a `project` directory inside the nested build causes sbt to fail.

GlobalStar117:
## Summary

When using `--addPluginSbtFile` with a multi-build setup, sbt would fail with `Invalid build URI (no handler available)` if the nested build's `project` directory doesn't exist.

## Problem

The issue occurs when:
1. A multi-build setup is used (e.g., `lazy val c1 = project`)
2. The nested build directory (`c1/`) exists but doesn't have a `project` subdirectory
3. `--addPluginSbtFile` option is used

The error message is:
```
java.lang.RuntimeException: Invalid build URI (no handler available): file:~/bug2Projects/c1/project/
```

## Root Cause

In `Resolvers.local`, when the URI points to a directory that doesn't exist, it returns `None`, causing the `fail` function to be called with the "no handler available" error.

## Solution

Modified `Resolvers.local` to create the directory if:
- It doesn't exist
- Its parent directory exists and is a directory

This matches the behavior of `checkDirectory` in `Load.scala` which already creates missing directories.

## Changes

- `main/src/main/scala/sbt/Resolvers.scala`: Added logic to create missing directory if parent exists
- `sbt-app/src/sbt-test/project/add-plugin-sbt-file-nested/`: Added scripted test

## Testing

Added a scripted test that verifies the fix works correctly.

Fixes https://github.com/sbt/sbt/issues/8570

---
This is a Gittensor contribution: [33f77537e7f32b9ddb9db247def29ada03f31fc9]

### DIFF
--- a/main/src/main/scala/sbt/Resolvers.scala
+++ b/main/src/main/scala/sbt/Resolvers.scala
@@ -34,7 +34,12 @@ object Resolvers {
     if (from.isDirectory) Some { () =>
       if (from.canWrite) from else creates(to) { IO.copyDirectory(from, to) }
     }
-    else None
+    else if (!from.exists && from.getParentFile != null && from.getParentFile.isDirectory) {
+      Some { () =>
+        IO.createDirectory(from)
+        from
+      }
+    } else None
   }
 
   val remote: Resolver = (info: ResolveInfo) => {
--- a/sbt-app/src/sbt-test/project/add-plugin-sbt-file-nested/build.sbt
+++ b/sbt-app/src/sbt-test/project/add-plugin-sbt-file-nested/build.sbt
@@ -1,0 +1,6 @@
+ThisBuild / scalaVersion := "2.13.16"
+
+lazy val root = (project in file("."))
+  .aggregate(c1)
+
+lazy val c1 = project
--- a/sbt-app/src/sbt-test/project/add-plugin-sbt-file-nested/c1/build.sbt
+++ b/sbt-app/src/sbt-test/project/add-plugin-sbt-file-nested/c1/build.sbt
@@ -1,0 +1,1 @@
+lazy val c1 = (project in file("."))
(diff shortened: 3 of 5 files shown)


main/src/main/scala/sbt/Resolvers.scala
```diff
@@ -34,7 +34,12 @@ object Resolvers {
     if (from.isDirectory) Some { () =>
       if (from.canWrite) from else creates(to) { IO.copyDirectory(from, to) }
     }
-    else None
+    else if (!from.exists && from.getParentFile != null && from.getParentFile.isDirectory) {
+      Some { () =>
+        IO.createDirectory(from)
```
azdrojowa123, line 39: I don't have a strong opinion, but maybe creating a directory [here](https://github.com/sbt/sbt/blob/215e9d63250388629b7c0c4195965d26d8933cea/main/src/main/scala/sbt/internal/Load.scala#L1415) would be more transparent.

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/8586

PR #8573, opened 2026-01-19
[2.x] feat: Automate derivation of JsonFormats where possible

### LINKED ISSUE #8288: Automate derivation of JsonFormats where possible
## Expectation

1. We should support easier serialization of case classes

## Original description

<details>

When converting plugins to sbt 2.0, I often see this kind of error.

```scala
[error] -- Error: /some/Class.scala:183:6 
[error] 183 |      compile := {
[error]     |      ^
[error]     |given evidence sjsonnew.JsonFormat[xsbti.compile.CompileAnalysis] is not found; opt out of caching by annotating the key with @transient, or as foo := Def.uncached(...), or provide a given value
```

This requires me to create a shim for `uncached` to crossbuild.

Often times the class in question is an internal Sbt class, a case class, or something else that should be easy to autoformat. It would be super useful to have a bit more automation to prevent users from having to uncache key assignments or `Task` values across the build.

</details>

wink030710:
### **Problem**
When migrating plugins/builds to **sbt 2.0**, cached tasks/settings can fail to compile with errors like:

```
given evidence sjsonnew.JsonFormat[xsbti.compile.CompileAnalysis] is not found;
opt out of caching by annotating the key with @transient, or as foo := Def.uncached(...),
or provide a given value
```

This commonly forces plugin authors to:
- Use `Def.uncached(...)` and lose caching benefits, or
- Add manual `JsonFormat` shims just to cross build.

### **Solution**
This PR introduces a small, focused mechanism to reduce friction by providing:

- **`AutoJsonFormat`**: a runtime-reflection-based helper intended for *simple* “case class-like” data structures and for providing clearer fallback errors when derivation isn’t feasible.
- **`AutoJsonFormats`**: predefined `given JsonFormat[...]` instances for a small set of commonly-problematic sbt/xsbti types (e.g. `xsbti.compile.CompileAnalysis`, `CompileResult`, etc.), so builds don’t immediately fail with missing evidence.

The intent is to make “obvious/common” cases work out-of-the-box and to provide actionable failure messages when serialization isn’t supported.

### **Why two files?**
- **`AutoJsonFormat.scala`** holds the *generic mechanism* (how formats can be derived / fallback behavior).
- **`AutoJsonFormats.scala`** holds the *concrete givens* for sbt/xsbti types, keeping sbt-specific policy separate from the generic helper.

### **Behavior / Notes**
- For unsupported or complex types, the provided formats fail with a clear message indicating that you may still need `Def.uncached()` or an explicit `JsonFormat`.
- This change is additive and avoids removing/changing existing public APIs.

### **Testing**
- `sbt "project utilCache" test`

### **Related**
Fixes #8288


util-cache/src/main/scala/sbt/util/AutoJsonFormat.scala
```diff
@@ -0,0 +1,160 @@
...
+package sbt.util
+
+import java.lang.reflect.{ Field, Modifier }
+import scala.util.Try
+import sjsonnew.{ JsonFormat, Builder, Unbuilder, DeserializationException }
+import sjsonnew.BasicJsonProtocol.*
+import scala.collection.mutable
+
+/**
+ * Runtime JsonFormat derivation for common sbt types to address issue #8288.
```
eed3si9n: Years ago there were library that did this, and it was bad because it tried to guess serialization and it often guessed incorrectly. Scala 3 has a facility to do type class derivation, so if we were to implement something automatic, I'd much rather to something that only derives `JsonFormat` only for the types it's 100% sure about.

util-cache/src/main/scala/sbt/util/AutoJsonFormat.scala
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    def read[J](jsOpt: Option[J], unbuilder: Unbuilder[J]): T = {
+      throw new UnsupportedOperationException(
+        s"""Cannot deserialize $typeName.
+           |Consider using Def.uncached() or providing an explicit JsonFormat.
+           |For sbt internal types, you may need to add the format to AutoJsonFormats.""".stripMargin
+      )
+    }
+  }
```
eed3si9n, line 107: This pushes compile-time error to runtime. I don't think this is a good idea.

util-cache/src/main/scala/sbt/util/AutoJsonFormat.scala
```diff
@@ -0,0 +1,97 @@
...
+object AutoJsonFormat {
+
+  /**
+   * Derives a JsonFormat for a product type (typically a case class).
+   *
+   * This is intentionally strict: derivation succeeds only if a JsonFormat exists
+   * for every field type.
+   */
+  @nowarn("msg=New anonymous class definition will be duplicated at each inline site")
+  inline def derived[T](using m: Mirror.ProductOf[T]): JsonFormat[T] =
```
eed3si9n, line 28: Thanks for this. I'd be happy to land the `derived` part of this PR.

PR #8574, opened 2026-01-19
[2.x] feat: Add setup-java action support in CrossJava

### LINKED ISSUE #8582: Support for setup-java action in CrossJava
## Expectation

setup-java downloads the JDKs in locations such as:

```bash
/opt/hostedtoolcache/Java_Zulu_jdk/25.0.1-8/x64
/opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/11.0.29-7/x64
C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\11.0.29-7\x64
```

For these to be available for CrossJava,

```bash
Global/fullJavaHomes
```

should include them as `zulu@25.0.1`, `temurin@11.0.29` etc.

grnlemon:
Add SetupJavaDiscoverConfig to detect JDKs installed by GitHub's setup-java action at paths like:
- /opt/hostedtoolcache/Java_Zulu_jdk/25.0.1-8/x64 (Linux)
- C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\11.0.29-7\x64 (Windows)

These JDKs are now available in fullJavaHomes as zulu@25.0.1, temurin@11.0.29, etc.

Supported vendors: Zulu, Temurin, Adopt, Corretto, Liberica, Microsoft, Semeru. Temurin-Hotspot and Adopt are normalized to
temurin.

Fixes #8582

### DIFF
--- a/main/src/test/scala/sbt/internal/CrossJavaTest.scala
+++ b/main/src/test/scala/sbt/internal/CrossJavaTest.scala
@@ -13,6 +13,7 @@ import java.io.File
 import org.scalatest.diagrams.Diagrams
 import org.scalatest.funsuite.AnyFunSuite
 import sbt.internal.CrossJava.JavaDiscoverConfig.*
+import sbt.io.{ IO, syntax }
 import sbt.io.syntax.*
 import scala.collection.immutable.ListMap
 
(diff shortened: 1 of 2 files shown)


main/src/main/scala/sbt/internal/inc/CompileAnalysisFormats.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  given compileAnalysisFormat: JsonFormat[CompileAnalysis] =
+    new JsonFormat[CompileAnalysis]:
+      override def read[J](jsOpt: Option[J], unbuilder: Unbuilder[J]): CompileAnalysis =
+        jsOpt match
+          case Some(js) =>
+            unbuilder.beginObject(js)
+            unbuilder.readField[Boolean]("isEmpty")
+            unbuilder.endObject()
+            Analysis.empty
```
eed3si9n: Reducing `CompileAnalysis` down to a `Boolean` is unsafe in my opinion. For example, if we wanted to implement a task that depends on the result of compilation like packaging a JAR file, testing, etc, this will emit single bit of information whether it's empty or not. So even when we change the content of the source code, it would end up not invalidating the downstream tasks. Furthermore, if the downstream task actually wanted to access the analysis information via `val a = (Compile / compile).value`, this won't have that information.

mkdev5: updated the implementation to use content fingerprinting for cache invalidation

main/src/test/scala/sbt/internal/CrossJavaTest.scala
```diff
@@ -193,4 +195,58 @@ class CrossJavaTest extends AnyFunSuite with Diagrams {
...
+  test("The setup-java selector should correctly pick up Zulu JDK") {
+    val conf = new SetupJavaDiscoverConfig(sbt.io.syntax.file(".")) {
+      override def candidates(): Vector[String] = Vector("Java_Zulu_jdk")
+      override def javaHomes: Vector[(String, File)] = {
+        Vector("zulu@25.0.1" -> (sbt.io.syntax.file(".") / "Java_Zulu_jdk" / "25.0.1-8" / "x64"))
+      }
+    }
+    val (version, file) = conf.javaHomes.head
+    assert(version == "zulu@25.0.1")
+    assert(file.getName == "x64")
```
eed3si9n: If the test sample overrides `javaHomes` and grabs the first element, then I don't think this is testing anything. Could you refactor the test to something like:

```scala
IO.withTempDirectory: temp =>
  IO.createDirectory(temp / "Java_Zulu_jdk" / "25.0.1-8" / "x64")
  val conf = new SetupJavaDiscoverConfig(temp)
  .... // rest should be similar
```

eed3si9n (changes requested):
Thanks for the contribution. I think test needs a bit more work, but overall going the right direction.

main/src/test/scala/sbt/internal/CrossJavaTest.scala
mkdev5: Done!

eed3si9n (approved):
Thanks!

PR #8577, opened 2026-01-19
[2.x] feat: Support ... as testOnly pattern

### LINKED ISSUE #8576: Support ... as testOnly pattern
## Problem

Currently testOnly supports glob expression like `**`.

```
> testOnly **.SomethingTest
```

This is fine, but `**` is somewhat inconvenient from Bash because Bash itself would expand the glob expression.

## Expectations

In addition to `**`, we should support `...` by substituting `...` with `**`, so we can write the following:

```
$ sbt --client testOnly ...SomethingTest
```

## Note

https://github.com/sbt/sbt/blob/215e9d63250388629b7c0c4195965d26d8933cea/main/src/main/scala/sbt/Defaults.scala#L2490-L2501

byteforge38:
Fixes #8576

### Problem

Currently `testOnly` supports glob expressions like `**`:

testOnly **.SomethingTest

However, `**` is inconvenient from Bash because Bash itself expands the glob expression, requiring users to quote the pattern.

### Solution

Support `...` as an alias for `**` by substituting `...` with `**` before applying the glob filter. This allows users to write:

```bash
$ sbt --client testOnly ...SomethingTest
```

Changes

- IncrementalTest.scala: Added expandGlob function that replaces ... with **, applied to both include and exclude patterns in selectedFilter
- Defaults.scala: Applied expandGlob in distinctParser for tab completion matching

Testing

- Added unit tests in SelectedFilterTest.scala for expandGlob and ... pattern matching
- Added scripted test cases in tests/test-exclude for end-to-end verification

Checklist

- Compiles successfully
- scalafmtCheckAll passes
- Tests added (unit + scripted)

### DIFF
--- a/main/src/main/scala/sbt/internal/IncrementalTest.scala
+++ b/main/src/main/scala/sbt/internal/IncrementalTest.scala
@@ -69,11 +69,14 @@ object IncrementalTest:
     Vector(Digest.sha256Hash(salt.getBytes("UTF-8")))
   }
 
+  /** Expands `...` to `**` in glob patterns. */
+  def expandGlob(pattern: String): String = pattern.replace("...", "**")
+
   def selectedFilter(args: Seq[String]): Seq[String => Boolean] =
     def matches(nfs: Seq[NameFilter], s: String) = nfs.exists(_.accept(s))
     val (excludeArgs, includeArgs) = args.partition(_.startsWith("-"))
-    val includeFilters = includeArgs.map(GlobFilter.apply)
-    val excludeFilters = excludeArgs.map(_.substring(1)).map(GlobFilter.apply)
+    val includeFilters = includeArgs.map(expandGlob).map(GlobFilter.apply)
+    val excludeFilters = excludeArgs.map(_.substring(1)).map(expandGlob).map(GlobFilter.apply)
     (includeFilters, excludeArgs) match
       case (Nil, Nil) => Seq(const(true))
       case (Nil, _)   => Seq((s: String) => !matches(excludeFilters, s))
--- a/main/src/main/scala/sbt/internal/IncrementalTest.scala
+++ b/main/src/main/scala/sbt/internal/IncrementalTest.scala
@@ -70,7 +70,7 @@ object IncrementalTest:
   }
 
   /** Expands `...` to `**` in glob patterns. */
-  def expandGlob(pattern: String): String = pattern.replace("...", "**")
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/internal/IncrementalTest.scala
```diff
@@ -69,11 +69,14 @@ object IncrementalTest:
     Vector(Digest.sha256Hash(salt.getBytes("UTF-8")))
   }
 
+  /** Expands `...` to `**` in glob patterns. */
+  def expandGlob(pattern: String): String = pattern.replace("...", "**")
```
eed3si9n:
```suggestion
  def expandGlob(pattern: String): String = pattern.replaceAllLiterally("...", "**")
```

This should allow multiple occurrence of `...`.

sbt-app/src/sbt-test/tests/test-exclude/test
```diff
@@ -59,3 +59,19 @@ $ delete target/Test2.run
...
+# with ... as a shorthand for ** (issue #8576)
+> testOnly ...Test1
+$ exists target/Test1.run
+-$ exists target/Test2.run
+
+$ delete target/Test1.run
+
+
+# with ... in exclusion pattern
+> testOnly ...Test* -...Test2
```
eed3si9n:
```suggestion
> testOnly ...Test... -...Test2
```

eed3si9n (approved):
Looks great. Thanks for adding the tests!

main/src/main/scala/sbt/internal/IncrementalTest.scala
```diff
@@ -70,7 +70,7 @@ object IncrementalTest:
   }
 
   /** Expands `...` to `**` in glob patterns. */
-  def expandGlob(pattern: String): String = pattern.replace("...", "**")
+  def expandGlob(pattern: String): String = pattern.replaceAllLiterally("...", "**")
```
eed3si9n: Opps. I was wrong about this:
```
method replaceAllLiterally in class StringOps is deprecated since 2.13.2: Use `s.replace` as an exact replacement
```

```suggestion
  def expandGlob(pattern: String): String = pattern.replace("...", "**")
```

byteforge38:
@eed3si9n It says that replaceAllLiterally has deprecated, could you please check it?

byteforge38:
All CI checks have passed now

eed3si9n:
Landed.

PR #8579, opened 2026-01-20
[2.x] test: Test transitive cached task

### LINKED ISSUE #8578: Transitive cached task
## steps

1. Define two cached task `task1` and `task2`, which produce `t1.txt` and `t2.txt` respectively (using `Def.declareOutput(...)`), and let `task1` depend on `task2`.
2. `task1`
3. ` clean`
4. `task1`

## problem

I haven't checked, but currently only `t1.txt` will be generated.

## expectation

The side effect for both `task1` and `task2` should be performed, producing `t1.txt` and `t2.txt`.

## note

Until this is implemented, we won't be able to make tasks like `compile` cached.

grnlemon:
## Summary

Adds a scripted test that verifies transitive cached task side effects work correctly, as discussed in #8578.

The test confirms that when a cached task (`task1`) depends on another cached task (`task2`), both tasks' declared outputs are properly restored from cache after `clean`.

## Test scenario

1. `task1` depends on `task2.value`
2. Both tasks use `Def.declareOutput` to declare output files
3. Run `task1` → both `t1.txt` and `t2.txt` are created
4. Run `clean`
5. Run `task1` again → both files are restored from cache ✓

Fixes #8578

### DIFF
--- a/sbt-app/src/sbt-test/cache/transitive/build.sbt
+++ b/sbt-app/src/sbt-test/cache/transitive/build.sbt
@@ -1,0 +1,27 @@
+import sbt.internal.util.StringVirtualFile1
+
+val task2 = taskKey[Unit]("Upstream cached task that produces t2.txt")
+val task1 = taskKey[Unit]("Downstream cached task that depends on task2 and produces t1.txt")
+val checkTask = taskKey[Unit]("Check that files exist")
+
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #8581, opened 2026-01-20
[2.x] feat: Add dependency lock file support (#2989)

### LINKED ISSUE #2989: Split dependency resolution into resolve and downloading (lock dependency)

grnlemon:
# Dependency Lock File Support

This PR adds dependency lock file support to sbt 2.x.

## What it does

When you run `dependencyLock`, sbt generates a `deps.lock` file that captures your resolved dependencies. This file can be checked into version control to ensure reproducible builds across different machines and CI environments.

## New tasks

- **`dependencyLock`** - Generates the lock file from the current resolution
- **`dependencyLockCheck`** - Validates the lock file is up-to-date (fails build if stale)

## How it works

The lock file stores a hash of your declared dependencies and resolvers. When dependencies change, the hash changes, and `dependencyLockCheck` will fail until you regenerate the lock file.

If no lock file exists, `dependencyLockCheck` passes silently - this allows gradual adoption.

## Implementation notes

- Lock file types are generated via Contraband, with manual JSON codecs for Scala 3 compatibility
- Unified the internal `LockFileData` types with the user-facing lock file format
- Lock file is per-project (uses `baseDirectory`, not `ThisBuild`)

Fixes #2989


main/src/main/scala/sbt/coursierint/LMCoursier.scala
```diff
@@ -173,8 +174,42 @@ object LMCoursier {
...
+            )
+            import lmcoursier.definitions.{
+              Module as CModule,
+              Organization as COrganization,
+              ModuleName as CModuleName
+            }
+            val additionalForceVersions = lockedVersions.map { case ((org, name), v) =>
+              CModule(COrganization(org), CModuleName(name), Map.empty) -> v
+            }.toVector
+            baseConfig.withForceVersions(baseConfig.forceVersions ++ additionalForceVersions)
```
eed3si9n: Is the idea to force all versions, and rerun the dependency resolution? I'm somewhat skeptical about this working for various corner cases. Assuming that part of the benefit of creating the lock file is for performance reasons, if we have to rerun resolution, performance-wise it would be the same as original.

mkdev5: You're right - the current approach using `forceVersions` doesn't provide the performance benefit and may have corner case issues.
The better approach would be to skip resolution entirely when a valid lock file exists. I'll update the implementation to:

1. Add a method to reconstruct `UpdateReport` from the lock file
2. In `updateTask`, check for valid lock file first and return the reconstructed report directly
3. Only fall back to normal resolution if lock file is missing/stale

This would provide the actual performance benefit of skipping resolution.
Would you prefer I implement this, or would you like to discuss the design further first?

eed3si9n: Ideally it would be nice if you can send the resolution object to Coursier, and Coursier can perform the fetching.

mkdev5: I've removed the `forceVersions` approach since it doesn't provide the performance benefit.

The PR now provides the lock file as a validation/documentation tool. For the resolution-skipping feature (passing resolution to Coursier for fetching only), I can work on that as a follow-up PR since it requires changes to the `lm-coursier` module to reconstruct and use `coursier.Resolution` objects from the lock file.

Would you like me to proceed with this approach, or would you prefer to have the full resolution-skipping feature in this PR?

eed3si9n: There's already https://github.com/stringbean/sbt-dependency-lock/ out in the wild, so there's no point in duplicating the feature if doesn't affect stability and performance of `update`.

mkdev5:
@eed3si9n CI test failure is none of my changes

lm-coursier/src/main/scala/lmcoursier/internal/LockedArtifactsRun.scala
```diff
@@ -0,0 +1,95 @@
...
+import coursier.cache.FileCache
+import coursier.core.{ Classifier, Dependency, Extension, Publication, Type }
+import coursier.util.Artifact
+import sbt.util.Logger
+
+import java.io.File
+import scala.concurrent.{ Await, ExecutionContext }
+import scala.concurrent.duration.Duration
+
+object LockedArtifactsRun {
```
eed3si9n, line 12: Nice. This emulate the Coursier fetching from the lock file?

mkdev5: yep, it fetches artifacts directly using the URLs stored in the lock file, bypassing Coursier's resolution step

lm-coursier/src/main/scala/lmcoursier/internal/LockFileData.scala
eed3si9n, line 1: Could you generate these using Contraband please? Example - https://github.com/sbt/sbt/blob/develop/protocol/src/main/contraband/worker.contra

mkdev5: please review the changes

main/src/main/scala/sbt/internal/librarymanagement/DependencyLock.scala
eed3si9n, line 1: Is this lock file still needed?

main/src/main/scala/sbt/internal/librarymanagement/DependencyLock.scala
```diff
@@ -0,0 +1,203 @@
...
+    buildClock: String,
+    resolvers: Vector[LockedResolver],
+    projects: Vector[ProjectLock]
+):
+  def isValid(currentBuildClock: String): Boolean =
+    buildClock == currentBuildClock
+
+object DependencyLockFile:
+  val CurrentLockVersion = "1.0"
+  val LockFileName = "build.sbt.lock"
```
eed3si9n:
```suggestion
  val lockFileName = "deps.lock"
```

Since we're not locking on the build information, I think `deps.lock` would be better name.

sbt-app/src/sbt-test/cache/transitive/project/build.properties
eed3si9n, line 1: Could you remove this plz?

sbt-app/src/sbt-test/dependency-management/dependency-lock/project/build.properties
eed3si9n, line 1: Same.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -3340,6 +3340,11 @@ object Classpaths {
       ew.infoAllTheThings foreach { log.info(_) }
       ew
     },
+    dependencyLockFile := (ThisBuild / baseDirectory).value / DependencyLockFile.LockFileName,
```
eed3si9n:
```suggestion
    dependencyLockFile := baseDirectory.value / DependencyLockFile.LockFileName,
```

When you have multiple subprojects, the lock files need to be at different locations, I assume?

main/src/main/scala/sbt/Keys.scala
```diff
@@ -533,6 +533,11 @@ object Keys {
...
   val transitiveClassifiers = settingKey[Seq[String]]("List of classifiers used for transitively obtaining extra artifacts for sbt or declared dependencies.").withRank(BSetting)
   val updateSbtClassifiers = TaskKey[UpdateReport]("updateSbtClassifiers", "Resolves and optionally retrieves classifiers, such as javadocs and sources, for sbt, transitively.", BPlusTask, updateClassifiers)
+  val dependencyLock = taskKey[File]("Generates a dependency lock file from the current resolution.").withRank(BTask)
+  val dependencyLockCheck = taskKey[Boolean]("Checks if the dependency lock file is up-to-date.").withRank(BTask)
+  val dependencyLockUpdate = taskKey[File]("Forces re-resolution and updates the dependency lock file.").withRank(BTask)
+  val dependencyLockFile = settingKey[File]("The location of the dependency lock file.").withRank(CSetting)
+  val useDependencyLock = settingKey[Boolean]("If true, uses the dependency lock file for resolution when available.").withRank(BSetting)
```
eed3si9n: Could we get rid of this setting? If the lock file exists, we should use it?

lm-coursier/src/main/scala/lmcoursier/internal/LockFileFormats.scala
eed3si9n, line 1: This can be generated using Contraband as well

```
@codecPackage("sbt.internal.worker.codec")
@fullCodec("JsonProtocol")
```

https://www.scala-sbt.org/contraband/json.html

main/src/main/scala/sbt/Keys.scala
mkdev5: Done! Removed useDependencyLock - the lock file is now used automatically if it exists at dependencyLockFile path.

main/src/main/scala/sbt/internal/librarymanagement/DependencyLock.scala
mkdev5: Yes, this is used by the `dependencyLock`/`dependencyLockCheck`/`dependencyLockUpdate` tasks for user-facing lock file management. The `lm-coursier/LockFileData` is used internally for resolution skipping. Happy to consolidate if you'd prefer a single implementation.

sbt-app/src/sbt-test/dependency-management/dependency-lock/build.sbt
```diff
@@ -0,0 +1,7 @@
+ThisBuild / scalaVersion := "2.13.12"
+
```
eed3si9n, line 3:
```suggestion
ThisBuild / csrCacheDirectory := (ThisBuild / baseDirectory).value / "coursier-cache"

```

main/src/main/scala/sbt/Defaults.scala
mkdev5: Done in dd7f27bec - changed to `baseDirectory.value` so each subproject gets its own lock file.

sbt-app/src/sbt-test/dependency-management/dependency-lock/changes/build-modified.sbt
```diff
@@ -0,0 +1,8 @@
+ThisBuild / scalaVersion := "2.13.12"
+
```
eed3si9n, line 3:
```suggestion
ThisBuild / csrCacheDirectory := (ThisBuild / baseDirectory).value / "coursier-cache"

```

sbt-app/src/sbt-test/dependency-management/resolution-skip/build.sbt
```diff
@@ -0,0 +1,8 @@
+ThisBuild / scalaVersion := "2.13.12"
+
```
eed3si9n, line 3:
```suggestion
ThisBuild / csrCacheDirectory := (ThisBuild / baseDirectory).value / "coursier-cache"

```

main/src/main/scala/sbt/Keys.scala
```diff
@@ -533,6 +533,10 @@ object Keys {
   val updateClassifiers = TaskKey[UpdateReport]("updateClassifiers", "Resolves and optionally retrieves classified artifacts, such as javadocs and sources, for dependency definitions, transitively.", BPlusTask, update)
   val transitiveClassifiers = settingKey[Seq[String]]("List of classifiers used for transitively obtaining extra artifacts for sbt or declared dependencies.").withRank(BSetting)
   val updateSbtClassifiers = TaskKey[UpdateReport]("updateSbtClassifiers", "Resolves and optionally retrieves classifiers, such as javadocs and sources, for sbt, transitively.", BPlusTask, updateClassifiers)
+  val dependencyLock = taskKey[File]("Generates a dependency lock file from the current resolution.").withRank(BTask)
+  val dependencyLockCheck = taskKey[Boolean]("Checks if the dependency lock file is up-to-date.").withRank(BTask)
+  val dependencyLockUpdate = taskKey[File]("Forces re-resolution and updates the dependency lock file.").withRank(BTask)
```
eed3si9n: I think `dependencyLockUpdate` should be removed as well. The user can just call `dependencyLock` again if they need to rewrite the lock files.

main/src/main/scala/sbt/internal/librarymanagement/DependencyLock.scala
eed3si9n: Yea, I think we should unify them if possible.

main/src/main/scala/sbt/Keys.scala
```diff
@@ -533,6 +533,10 @@ object Keys {
   val updateClassifiers = TaskKey[UpdateReport]("updateClassifiers", "Resolves and optionally retrieves classified artifacts, such as javadocs and sources, for dependency definitions, transitively.", BPlusTask, update)
   val transitiveClassifiers = settingKey[Seq[String]]("List of classifiers used for transitively obtaining extra artifacts for sbt or declared dependencies.").withRank(BSetting)
   val updateSbtClassifiers = TaskKey[UpdateReport]("updateSbtClassifiers", "Resolves and optionally retrieves classifiers, such as javadocs and sources, for sbt, transitively.", BPlusTask, updateClassifiers)
+  val dependencyLock = taskKey[File]("Generates a dependency lock file from the current resolution.").withRank(BTask)
+  val dependencyLockCheck = taskKey[Boolean]("Checks if the dependency lock file is up-to-date.").withRank(BTask)
```
eed3si9n: Since this is a linter task, the type should be `taskKey[Unit]` here, and throw `MessageOnlyException(...)` to fail the build if the lint fails.

main/src/main/scala/sbt/internal/librarymanagement/DependencyLock.scala
MkDev11: @eed3si9n That is a significant refactoring, can you create an issue for it?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -3764,6 +3767,56 @@ object Classpaths {
...
+    lockFile
+  }
+
+  lazy val dependencyLockCheckTask: Initialize[Task[Unit]] = Def.task {
+    val log = streams.value.log
+    val lockFile = dependencyLockFile.value
+    val deps = libraryDependencies.value
+    val resolverNames = fullResolvers.value.map(_.name)
+    val currentBuildClock = DependencyLockFile.computeBuildClock(deps, resolverNames)
+
```
eed3si9n: If `lockFile` doesn't exist, `dependencyLockCheck` should pass without printing anything.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -3764,6 +3767,56 @@ object Classpaths {
...
+    DependencyLockManager.validate(lockFile, currentBuildClock, log) match
+      case Some(_) =>
+        log.info(s"Dependency lock file is up-to-date: ${lockFile.getAbsolutePath}")
```
eed3si9n: Similarly, if the linter passes, it should `()`, without printing anything on the log.

main/src/main/scala/sbt/internal/librarymanagement/DependencyLock.scala
mkdev5: Implemented it!

lm-coursier/src/main/scala/lmcoursier/CoursierDependencyResolution.scala
```diff
@@ -320,12 +323,51 @@ class CoursierDependencyResolution(
...
+      val report = UpdateRun.update(updateParams0, verbosityLevel, log)
+      if (lockDataOpt.isEmpty) {
+        conf.lockFile.foreach { lockFile =>
+          val artifactMap = artifactResult
+            .groupBy(_._1)
+            .view
+            .mapValues(_.map { case (_, pub, art, _) =>
+              (art.url, pub.classifier.value, pub.ext.value)
+            })
+            .toMap
```
eed3si9n, line 360: I ran this locally, and `url` currently seem to contain the absolute path specific to my laptop. Since we want `lock` files to be checked into `git`, we should use the original `http` URL if it's available. If not `fileConverter` can replace the Coursier cache path to `${CSR_CACHE}`.

```json
"artifacts":[{
  "url":"file:/private/tmp/resolution-skip/coursier-cache/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.13.12/scala-library-2.13.12.jar",
  "extension":"jar",
  "tpe":"jar"
}]
```

lm-coursier/src/main/scala/lmcoursier/internal/ResolutionSerializer.scala
```diff
@@ -0,0 +1,181 @@
...
+      .map { case (config, resolution) =>
+        val dependencies = extractDependencies(resolution, config, artifactMap)
+        ConfigurationLock(config.value, dependencies.toVector)
+      }
+      .toVector
+
+    val metadata = LockFileMetadata(
+      sbtVersion = sbtVersion,
+      scalaVersion = scalaVersion,
+      timestamp = Instant.now()
```
eed3si9n: Do we have use case for this timestamp? If possible we should remove timestamps so the content would be repeatable across runs.

sbt-app/src/sbt-test/cache/transitive/project/build.properties
eed3si9n, line 1: Please remove these. You can test scripted test as `sbt scripted cache/transitive`.

sbt-app/src/sbt-test/dependency-management/dependency-lock/project/build.properties
eed3si9n, line 1: Same

sbt-app/src/sbt-test/dependency-management/resolution-skip/project/build.properties
eed3si9n, line 1: Same

eed3si9n (approved):
This is a cool feature enhancement. Thanks for the contribution!

mkdev5:
@eed3si9n please review the changes again

PR #8583, opened 2026-01-20
[1.x] Fix missing `project` directory

### LINKED ISSUE #8570: The sbt fails with multi-build setup and `addPluginSbtFile`

azdrojowa123:
Fixes #8570

### DIFF
--- a/sbt-app/src/sbt-test/project/addPluginSbtFile-no-project-dir/test
+++ b/sbt-app/src/sbt-test/project/addPluginSbtFile-no-project-dir/test
@@ -1,8 +1,0 @@
-# Start sbt first without dependency on c1
-> reload
-
-# Add dependency on c1
-$ copy-file temp/build.sbt build.sbt
-
-# Execute early command with addPluginSbtFile and reload
-> early(addPluginSbtFile=temp/extra.sbt); reload
--- a/sbt-app/src/sbt-test/project/addPluginSbtFile-no-project-dir/test
+++ b/sbt-app/src/sbt-test/project/addPluginSbtFile-no-project-dir/test
@@ -1,0 +1,8 @@
+# Start sbt first without dependency on c1
+> reload
+
+# Add dependency on c1
+$ copy-file temp/build.sbt build.sbt
+
+# Execute early command with addPluginSbtFile and reload
+> early(addPluginSbtFile=temp/extra.sbt); reload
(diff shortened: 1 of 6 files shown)


sbt-app/src/sbt-test/project/addPluginSbtFile-no-project-dir/test
```diff
@@ -0,0 +1,8 @@
+# Start sbt first without dependency on c1
+> reload
+
+# Add dependency on c1
+$ copy-file temp/build.sbt build.sbt
+
+# Execute early command with addPluginSbtFile and reload
+> early(addPluginSbtFile=temp/extra.sbt); reload
```
azdrojowa123: I didn't find any way to pass additional command line arguments when using scripted tests to have the test with the same steps as in #8570, but I have little experience with scripted tests
However, these steps also trigger the same exception. Locally, without my fix, this test fails with the same exception as in #8570, and with my fix, it passes.

eed3si9n: Interestingly `project/aggregate` is failing, which I'm guessing is because we batch multiple scripted tests into one sbt session. If the manual test works for you, I'm ok without the scripted test.

azdrojowa123: It seems so :/ It might be useful to run tests in a separate session.
Then, I'll remove the test for now.

PR #8584, opened 2026-01-20
[2.x] feat: Forked compilation

### LINKED ISSUE #887: Support compiling with a newer JDK version than the one running sbt
The incremental compiler currently loads the class files compiled by javac in a class loader in the same jvm as the inc. compiler.  This prevents working with a newer classfile version than that of the inc. compiler's jvm, such as when forking javac for a user configured `javaHome`.  To remove this restriction likely requires a substantial rewrite of the Java handling code.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/887

## Problem
Currently Zinc runs in the same process as the sbt server, which means that the JDK/Java version of the build artifact gets tied to the JDK/Java version used by sbt itself.

## Solution
This implements a command-line interface for Zinc, which can be invoked via the worker process.

1. This can be used to compile using JDK 25, if Zinc works on JDK 25.
2. When we upgrade to JDK 17, we can downgrade the compilation to use the real JDK 8, if Zinc itself it compiled with `--release 8`.

### DIFF
--- a/main-actions/src/main/scala/sbt/internal/CompileMain.scala
+++ b/main-actions/src/main/scala/sbt/internal/CompileMain.scala
@@ -53,9 +53,10 @@ object CompileMain:
       val log = mkLogger(Level.Warn)
       val output = Paths.get(config.output)
       val sources = config.sources.map(Paths.get(_)).map(conv.toVirtualFile)
-      val cp = config.externalDependencyJars.map(Paths.get(_)).map(conv.toVirtualFile)
+      val cp = config.externalDependencyJars.map(Paths.get(_))
+      val cpVf = cp.map(conv.toVirtualFile)
       val in = zinc.inputs(
-        classpath = cp.toArray,
+        classpath = cpVf.toArray,
         sources = sources.toArray,
         classesDirectory = output,
         earlyJarPath = config.earlyJarPath.map(Paths.get(_)),
--- a/main-actions/src/main/scala/sbt/internal/CompileMain.scala
+++ b/main-actions/src/main/scala/sbt/internal/CompileMain.scala
@@ -1,0 +1,193 @@
+package sbt
+package internal
+
+import java.io.{ File, PrintStream }
+import java.nio.file.{ Path, Paths }
+import java.util.Optional
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.function.{ Function as JavaFunction }
+import sbt.internal.inc.{
+  AnalyzingCompiler,
+  ScalaInstance,
+  Locate,
+  IncrementalCompilerImpl,
+  MappedFileConverter,
+  ManagedLoggedReporter,
+  Stamps,
+  ZincUtil,
+}
+import sbt.internal.inc.classpath.ClasspathUtil
+import sbt.internal.inc.JavaInterfaceUtil.*
+import sbt.internal.worker.{ CompileConfig, CompileResponse, ScalaInstanceConfig }
+import sbt.internal.worker.codec.JsonProtocol.given
+import sbt.internal.util.{ ManagedLogger, ConsoleOut, MainAppender }
+import sbt.io.IO
+import sbt.util.{ LoggerContext, Level }
+import sjsonnew.support.scalajson.unsafe.{ CompactPrinter, Parser, Converter }
+import xsbti.{ CompileFailed, HashedVirtualFileRef, Position, T2, VirtualFile }
+import xsbti.compile.{ ScalaInstance as _, * }
+
+object CompileMain:
+  lazy val zinc: IncrementalCompilerImpl = new IncrementalCompilerImpl
+
+  def run(config: CompileConfig, id: Long, jsonOut: PrintStream): Unit =
+    try
+      // println(s"config: $config")
+      val conv = MappedFileConverter(
+        Map((config.fileConverterConfig.rootPaths.map: x =>
+          (x.name, Paths.get(x.value)))*),
+        true
+      )
+      val si = scalaInstance(config.scalaInstanceConfig)
+      val bridgeJars = config.bridgeJars.map(Paths.get(_))
+      val scalac = analyzingCompiler(bridgeJars, si)
+      val co = ClasspathOptionsUtil.noboot(config.scalaInstanceConfig.scalaVersion)
+      val cs = ZincUtil.compilers(
+        instance = si,
+        classpathOptions = co,
+        javaHome = None,
+        scalac = scalac,
+      )
+      val analysisFile = Paths.get(config.analysisFile)
+      val setup = incSetup(config)
+      val log = mkLogger(Level.Warn)
+      val output = Paths.get(config.output)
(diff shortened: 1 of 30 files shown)


main-actions/src/main/scala/sbt/internal/CompileMain.scala
```diff
@@ -0,0 +1,194 @@
...
+    s"""{ "jsonrpc": "2.0", "error": { "code": 1009, "message": "$err" }, "id": $id }"""
+
+  def incSetup(config: CompileConfig): Setup =
+    val analysisFile = Paths.get(config.analysisFile)
+    val analysisForCp: Map[HashedVirtualFileRef, Path] = Map((config.analysisMap.map: pair =>
+      pair.name -> Paths.get(pair.value))*)
+    val lookup: PerClasspathEntryLookup = new PerClasspathEntryLookup:
+      def read(p: Path) = FileAnalysisStore.getDefault(p.toFile).get().toOption.map(_.getAnalysis)
+      override def analysis(cpEntry: VirtualFile): Optional[CompileAnalysis] =
+        analysisForCp.get(cpEntry).flatMap(read).toOptional
```
stasimus, line 104: per-classpath-entry analysis lookup can never hit. analysisForCp is keyed by the deserialized BasicHashedVirtualFileRef, whose equals/hashCode include the content hash and size, but the probe key is a MappedVirtualFile that hashes id only  
so inter-project incremental compilation falls back to "no upstream analysis" in every forked build.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2202,8 +2220,27 @@ object Defaults extends BuildCommon {
...
+      val fo = (compile / forkOptions).value
+      val sic = scalaInstanceConfig.value
+      val bridges = scalaCompilerBridgeJars.value
+      val rs = rootPaths.value
+      val pickles = dependencyPicklePath.value
       // TODO - Should readAnalysis + saveAnalysis be scoped by the compile task too?
-      val analysisResult = Retry.io(compileIncrementalTaskImpl(bspTask, s, ci, ping))
+      val analysisResult =
+        if fk then
+          ForkCompile.compile(
```
stasimus, line 2233: the fork path never completes earlyOutputPing. ForkCompile.compile passes no progress, and CompileMain.incSetup hardcodes jnone. Turn pipelining on and any downstream compileEarly sits in earlyOutputPing.await forever.

main-actions/src/main/scala/sbt/internal/ProcessReact.scala
```diff
@@ -0,0 +1,55 @@
...
+      else if o.has("re") && o.has("method") then
+        val resId = o.getAsJsonPrimitive("re").getAsLong()
+        if resId == id then processNotification(o)
+        else ()
+      else ()
+    catch
+      case _: JsonSyntaxException => log.info(line)
+      case NonFatal(_)            => ()
+
+  override def notifyExit(p: Process): Unit =
```
stasimus, line 48: race between the response and the exit. The worker writes its JSON-RPC response and immediately calls System.exit(0), while watch polls isAlive every 100ms and fails the promise as soon as the process is gone. Nothing orders the reader thread ahead of that, so a successful compile can fail with "worker exited with code 0".

stasimus: WorkerProxy.watch also walks the shared ListBuffer without holding the WorkerExchange loc

main-actions/src/main/scala/sbt/internal/ProcessReact.scala
```diff
@@ -0,0 +1,55 @@
...
+    if !process.isAlive && !promise.isCompleted then
+      val exitCode = process.exitValue()
+      promise.failure(new RuntimeException(s"worker exited with code $exitCode"))
+
+  def blockForResponse(): A1 =
+    Await.result(promise.future, Duration.Inf)
```
stasimus, line 54: Await.result(promise.future, Duration.Inf) and the only liveness backstop is watch, which reads the listener list before ForkCompile registers its own. A worker that dies inside that window notifies nobody and the compile hangs for the rest.

main-actions/src/main/scala/sbt/internal/CompileMain.scala
```diff
@@ -0,0 +1,194 @@
...
+      store.set(AnalysisContents.create(r.getAnalysis(), r.getMiniSetup()))
+      val response = CompileResponse(
+        hasModified = r.hasModified()
+      )
+      val resJson = Converter.toJson(response).get
+      val json = CompactPrinter(resJson)
+      jsonOut.println(jsonRpcResponse(id, json))
+      jsonOut.flush()
+    catch
+      case e: CompileFailed =>
```
stasimus, line 86: Everything else turns into WorkerError(1, e.getMessage()). When the message is null Gson drops the field and ProcessReact NPEs on getAsJsonPrimitive("message"), which case NonFatal(_) => () then swallows ...

main-actions/src/main/scala/sbt/internal/CompileMain.scala
```diff
@@ -0,0 +1,194 @@
...
+  def incSetup(config: CompileConfig): Setup =
+    val analysisFile = Paths.get(config.analysisFile)
+    val analysisForCp: Map[HashedVirtualFileRef, Path] = Map((config.analysisMap.map: pair =>
+      pair.name -> Paths.get(pair.value))*)
+    val lookup: PerClasspathEntryLookup = new PerClasspathEntryLookup:
+      def read(p: Path) = FileAnalysisStore.getDefault(p.toFile).get().toOption.map(_.getAnalysis)
+      override def analysis(cpEntry: VirtualFile): Optional[CompileAnalysis] =
+        analysisForCp.get(cpEntry).flatMap(read).toOptional
+      override def definesClass(cpEntry: VirtualFile) = Locate.definesClass(cpEntry)
+    val incOptions = IncOptions.of()
```
stasimus, line 106: the worker builds Setup from scratch and loses most of what the server had: incOptions becomes IncOptions.of(), extraIncOptions

skip is hardcoded to false, is it desirable?

main-actions/src/main/scala/sbt/internal/CompileMain.scala
```diff
@@ -0,0 +1,194 @@
...
+    val log = mkLogger(Level.Warn)
+    val reporter = ManagedLoggedReporter(100, log)
```
stasimus, line 108: the reporter is ManagedLoggedReporter(100, log). maxErrors does make it across the wire and is used for CompileOptions, but not here, so maxErrors := 5 does nothing.

stasimus, line 108: The BSP reporter is also replaced by a plain logger, and since ProcessReact rebuilds CompileFailed(Array.empty, message, Array.empty), IDE clients will suffer from empty context attached.

stasimus:
few tests https://github.com/eed3si9n/sbt/pull/7

Issue #8587, opened 2026-01-20
Add API for using the latest Scala 3 Release Candidate


Gedochao:
Currently, using a Scala 3 release candidate with SBT requires passing an explicit version.

```scala
ThisBuild / scalaVersion := "3.8.1-RC1"
lazy val root = (project in file("."))
  .settings(name := "sbt-with-scala-rc")
```

We could add an API which would let SBT fetch the latest rc version

That way it would be valid to do

```scala
ThisBuild / scalaVersion := Scala.latestRC
lazy val root = (project in file("."))
  .settings(name := "sbt-with-scala-rc")
```

We'd like nightlies to be easy to test, so that the newest compiler changes can be verified in users' projects.
Having an API would let people set up builds which would automatically test their project on the latest RC.
This in turn could make regression reports reach the Scala 3 compiler team faster.

Additional context: Scala CLI added an equivalent in v1.12.0 at https://github.com/VirtusLab/scala-cli/pull/4016
```scala
//> using scala 3.rc
```

jackdevis:
@Gedochao should i work on issue?

Gedochao:
@jackdevis don't let me stop you, sure!

jackdevis:
Could you please review my pr?

Implemented Scala.latestRC that automatically fetches the latest Scala 3 RC version from Maven Central, allowing users to easily
test against the newest compiler changes.

- Add Scala object with latestRC method
- Fetch versions from Maven Central metadata XML
- Add unit tests for version parsing and comparison

https://github.com/sbt/sbt/pull/8588

eed3si9n:
We use GitHub issue mostly to track bugs and tasks on sbt development. Please use Discussion for the feature enhancement request, esp since there are people actively contributing to sbt for any issues, and they seem to not be able to tell who are sbt core devs.

PR #8588, opened 2026-01-20
Implementing Scala.latestRC

jackdevis:
Implemented Scala.latestRC that automatically fetches the latest Scala 3 RC version from Maven Central, allowing users to easily
test against the newest compiler changes.

- Add Scala object with latestRC method
- Fetch versions from Maven Central metadata XML
- Add unit tests for version parsing and comparison
- Include ScalaDoc documentation

#8587

### DIFF
--- a/main/src/main/scala/sbt/Scala.scala
+++ b/main/src/main/scala/sbt/Scala.scala
@@ -1,0 +1,87 @@
+/*
+ * sbt
+ * Copyright 2023, Scala center
+ * Copyright 2011 - 2022, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt
+
+import java.net.{ HttpURLConnection, URL }
+import scala.util.control.NonFatal
+import scala.util.Using
+import scala.xml.XML
+
+object Scala:
+  private val MavenMetadataUrl =
+    "https://repo1.maven.org/maven2/org/scala-lang/scala3-library_3/maven-metadata.xml"
+  private val RCPattern = """3\.\d+\.\d+-RC\d+""".r
+
+  /**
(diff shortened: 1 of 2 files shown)


main/src/main/scala/sbt/Scala.scala
```diff
@@ -0,0 +1,87 @@
...
+package sbt
+
+import java.net.{ HttpURLConnection, URL }
+import scala.util.control.NonFatal
+import scala.util.Using
+import scala.xml.XML
+
+object Scala:
+  private val MavenMetadataUrl =
+    "https://repo1.maven.org/maven2/org/scala-lang/scala3-library_3/maven-metadata.xml"
```
WojciechMazur: Potentially `repo.scala-lang.org` might be used instead, which would also prepare the codebase for adding `latestNightly` in the next step. It has both versions from Maven, and it's own nightly versions, so can be used for both
That's also what we used in the OpenCB https://github.com/VirtusLab/community-build3/blob/master/scripts/scalaVersions.scala
See: https://repo.scala-lang.org/ui/native/maven-nightlies/org/scala-lang/scala3-library_3

jackdevis: @WojciechMazur  For my latest update:

- Using repo.scala-lang.org/artifactory/maven-nightlies/ instead of Maven Central
- Including ScalaNightliesBaseUrl constant
- Refactoring to support future latestNightly functionality

eed3si9n (changes requested):
Thanks for the contribution.
Putting this on hold while we discuss the details in https://github.com/sbt/sbt/discussions/8590

eed3si9n:
I've created an issue for this feature request - https://github.com/sbt/sbt/issues/8595

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/8596

PR #8594, opened 2026-01-21
Fix cursor jumps and completion list misalignment on terminal resize

### LINKED ISSUE #8591: Cursor jumps and completion list misalignment in sbt shell when resizing terminal
## steps

1. Start the sbt shell in the terminal 
2. Start entering an sbt command in the sbt shell
3. Press _Tab_ to invoke the command completion list
4. Resize the window

## problem

1. The completion list is poorly aligned after resizing
2. The cursor jumps to a random position

## expectation

Completion list items should remain properly aligned in columns
The cursor should stay in the command line

## notes
The issue is highly noticeable when working in IntelliJ IDEA.
It reproduces on both macOS and Windows.

https://github.com/user-attachments/assets/ed133736-10ba-4656-a9bc-fe698a843462

Image

sharpenteeth:
This fixes an issue where the completion list becomes misaligned and the cursor jumps to a random position when the terminal window is resized while a completion list is displayed.

## Problem

The terminal size was cached with a 1-second refresh period, so when the terminal was resized, the completion list would use a stale width value, causing misalignment.

## Solution

The fix:
- Adds a method to invalidate the size cache immediately
- Registers a SIGWINCH handler (on Unix-like systems) to invalidate the cache when a terminal resize is detected
- Invalidates the cache when setSize is explicitly called
- Properly cleans up the signal handler when the terminal is closed

This ensures that when JLine queries the terminal size after a resize, it gets the fresh value immediately, allowing it to properly re-render the completion list with correct alignment.

## Testing

This should fix the issue reported in https://github.com/sbt/sbt/issues/8591 where resizing the terminal while the completion list is displayed causes misalignment.

## Related Issues

Fixes: #8591


eed3si9n:
@dive2tech Thanks for the contribution! Have you tried running sbt locally to confirm that the fix worked? - https://github.com/sbt/sbt/blob/develop/contributing-docs/06_manual_tests.md

sharpenteeth:
I tested and it works. Please review my PR

test-terminal-resize-fix.md
eed3si9n, line 1: Could you remove this markdown file please?

eed3si9n (changes requested):
on my macOS laptop, I wasn't able to confirm the fix:

Image

sharpenteeth:
This is screenshot on my side.

eed3si9n:
Which operating system and terminal are you using? Mine was macOS and iTerm.

sharpenteeth:
@eed3si9n, please run this PR on your side. Isn't it still working?

eed3si9n:
Still broken for me, either using iTerm or Terminal.app. Which operating system are you using to locally test?

eed3si9n:
Closing this PR unfortunately.

PR #8598, opened 2026-01-21
[2.x] test: Add scripted test for #6175

### LINKED ISSUE #6175: MethodHandle.invokeExact fails to compile on JDK 15 when targeting Java 11
## steps

Here's a reproducer

https://github.com/srdo/MethodHandlesMiscompiled

Install Java 15.0.1 (I used Zulu).
Run `sbt clean compile` in the project root.

## problem

The Main file fails to compile.

[error] /home/stig/MethodHandlesMiscompiled/src/main/scala/Main.scala:10:27: type mismatch;
[error]  found   : Object
[error]  required: Main.TestClass
[error]         handle.invokeExact()
[error]                           ^
[error] one error found

## expectation

It should have compiled.

## notes
I originally opened an issue on the Scala issue tracker here https://github.com/scala/bug/issues/12236. I suspect that this may actually be an SBT or Zinc issue.

The Main file compiles fine on JDK 15 if I remove `-release 11` from the `scalacOptions`. The file compiles fine when I run `scalac -release 11 Main.scala` directly (scalac version 2.13.4). It appears to only fail to compile in SBT.

calm329:
This PR adds a scripted test to prevent regression of issue #6175, where `MethodHandle.invokeExact` failed to compile on JDK 15+ when using `-release 11`.

#### Changes:
- Added `sbt-app/src/sbt-test/java/method-handle-release/` scripted test
- Test uses Scala 2.13.16 with `-release:11` to reproduce the original issue scenario
- Verifies that polymorphic signature methods compile correctly under cross-version targeting

#### Testing:
- Run `sbt "scripted java/method-handle-release"` to verify the test passes
- Test runs in CI jobtype 3 on JDK 25

Fixes #6175

### DIFF
--- a/sbt-app/src/sbt-test/java/method-handle-release/build.sbt
+++ b/sbt-app/src/sbt-test/java/method-handle-release/build.sbt
@@ -1,0 +1,2 @@
+ThisBuild / scalaVersion := "2.13.16"
+scalacOptions += "-release:11"
--- a/sbt-app/src/sbt-test/java/method-handle-release/src/main/scala/Main.scala
+++ b/sbt-app/src/sbt-test/java/method-handle-release/src/main/scala/Main.scala
@@ -1,0 +1,19 @@
+import java.lang.invoke.{ MethodHandle, MethodHandles }
+
+object Main {
+  case class TestClass(number: Int)
+  def testMethod(): TestClass = {
+    TestClass(10)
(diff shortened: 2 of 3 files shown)


calm329:

eed3si9n (approved):
Thanks!

PR #8599, opened 2026-01-21
[2.x] test: Add test for sbt --client startup time

### LINKED ISSUE #6485: Test `sbt --client` startup
Ref https://github.com/sbt/sbt/issues/5980
Ref https://github.com/sbt/sbt/pull/6483

## steps

```bash
sbt -v --client compile
sbt -v --client compile
```

## expectation

To prevent regression of the startup time, we should probably measure it, and cap it at 200ms?

_Originally posted by @vegerot in https://github.com/sbt/sbt/pull/6483#discussion_r624092750_

calm329:
## Summary
- Add integration test to measure `sbt --client` startup time and prevent performance regressions
- Test runs 5 iterations after warmup and asserts average stays under 2000ms
- Skipped on Windows and MacOS where sbtn behavior differs

Fixes #6485

## Test plan
- [x] Run `launcherPackageIntegrationTest/test` - all tests pass
- [x] Verify startup times are printed and within threshold

### DIFF
--- a/launcher-package/integration-test/src/test/scala/ExtendedRunnerTest.scala
+++ b/launcher-package/integration-test/src/test/scala/ExtendedRunnerTest.scala
@@ -142,8 +142,8 @@ object ExtendedRunnerTest extends BasicTestSuite:
   // Test for issue #6485: Test `sbt --client` startup
   // https://github.com/sbt/sbt/issues/6485
   test("sbt --client startup time") {
-    if (isWindows) {
-      // Skip on Windows for now as sbtn behavior differs
+    if (isWindows || isMac) {
+      // Skip on Windows (sbtn behavior differs) and macOS CI (slow hostname resolution)
       ()
     } else {
       // First call starts the server if not running (warmup)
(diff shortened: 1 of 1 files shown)


calm329:

eed3si9n (approved):
Looks great! Thanks!

launcher-package/integration-test/src/test/scala/ExtendedRunnerTest.scala
```diff
@@ -138,4 +138,47 @@ object ExtendedRunnerTest extends BasicTestSuite:
     }
     ()
   }
+
+  // Test for issue #6485: Test `sbt --client` startup
+  // https://github.com/sbt/sbt/issues/6485
+  test("sbt --client startup time") {
+    if (isWindows || isMac) {
```
eed3si9n, line 145: This is a commentary, not a change request, but if it takes 5s on macOS I have a feeling that sbtn is somehow not working well.

calm329: I think I found the root cause - `InetAddress.getLocalHost.getHostName` in `JUnitXmlTestsListener` takes ~5 seconds on macOS CI with misconfigured hostnames, and it's being called per-instance rather than cached.

  Would you be okay with caching the hostname at the companion object level so it's only resolved once per JVM?

eed3si9n: Let me file this as a separate bug.

eed3si9n: https://github.com/sbt/sbt/issues/8601

PR #8602, opened 2026-01-21
[2.x] fix: Use -external-mappings for Scala 3 doc task #6652

### LINKED ISSUE #6652: Doc task fails with `bad option '-doc-external-doc'` for scala3 project
## steps

- Checkout datadog4s at this [commit](https://github.com/avast/datadog4s/commit/afe26dddbb714c48499dfc157ecc29a7cdc4ca9c)
- run `sbt doc`

## problem
Task succeeds but apidoc is not generated, following warning is written out: 
```
[warn] bad option '-doc-external-doc:/modules/java.base#https://docs.oracle.com/en/java/javas...
```
## expectation

I expect that:
- the task would probably fail in order to notify user that no api docs were generated
- the task would use correct command for scala3 projects (which based on scaladoc man page i suspect is the following:
```
-external-mappings Mapping between regexes matching classpath entries and
                   external documentation.
                   'regex::[scaladoc|scaladoc|javadoc]::path' syntax is used
```
## notes
i believe this is the offending line: https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/Opts.scala#L40

calm329:
## Problem

When `autoAPIMappings := true` is set on a Scala 3 project, running `sbt doc` emits warnings:

```
[warn] bad option '-doc-external-doc:/modules/java.base#https://docs.oracle.com/...
```

This happens because Scala 3's scaladoc doesn't recognize Scala 2's `-doc-external-doc` option.

Fixes #6652

## Solution

- Added `Opts.doc.externalAPIScala3` that generates the Scala 3 format: `-external-mappings:regex::[scaladoc3|javadoc]::url`
- Modified `Defaults.scala` to use the appropriate method based on Scala version
- Added heuristics to detect javadoc vs scaladoc based on file/URL patterns

## Testing

- Added new scripted test `actions/doc-external-scala3`
- All existing doc tests pass (`actions/doc*`)
- Unit tests pass (`mainProj/test`)

### DIFF
--- a/sbt-app/src/sbt-test/actions/doc-external-scala3/build.sbt
+++ b/sbt-app/src/sbt-test/actions/doc-external-scala3/build.sbt
@@ -1,0 +1,22 @@
+// Test for issue #6652: Doc task fails with `bad option '-doc-external-doc'` for scala3 project
+// This test verifies that autoAPIMappings works correctly with Scala 3's -external-mappings option
+
+lazy val root = (project in file("."))
+  .settings(
+    scalaVersion := "3.3.4",
+    autoAPIMappings := true,
+    libraryDependencies += "org.typelevel" %% "cats-core" % "2.12.0",
+    TaskKey[Unit]("checkDocGenerated") := {
+      val docDir = (Compile / doc / target).value
+      val indexFile = docDir / "index.html"
+      assert(indexFile.exists(), s"Expected $indexFile to exist after doc generation")
+      println(s"Documentation generated successfully at $docDir")
+    },
+    TaskKey[Unit]("checkApiMappings") := {
+      val mappings = (Compile / doc / apiMappings).value
+      println(s"API Mappings count: ${mappings.size}")
+      // With autoAPIMappings and cats-core dependency, we should have some mappings
+      // (cats-core publishes with apiURL)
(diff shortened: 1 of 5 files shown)


calm329:

calm329:
The failed CI check isn't related to my change

eed3si9n (approved):
Thanks!

sbt-app/src/sbt-test/actions/doc-external-scala3/build.sbt
```diff
@@ -0,0 +1,22 @@
...
+      val indexFile = docDir / "index.html"
+      assert(indexFile.exists(), s"Expected $indexFile to exist after doc generation")
+      println(s"Documentation generated successfully at $docDir")
+    },
+    TaskKey[Unit]("checkApiMappings") := {
+      val mappings = (Compile / doc / apiMappings).value
+      println(s"API Mappings count: ${mappings.size}")
+      // With autoAPIMappings and cats-core dependency, we should have some mappings
+      // (cats-core publishes with apiURL)
+      println(s"API Mappings: ${mappings.take(3).mkString("\n  ", "\n  ", "")}")
```
mzuehlke, line 20: This check task doesn't contain an `assert`

PR #8603, opened 2026-01-21
[2.x] fix: Lazy cache hostname resolution at object level

### LINKED ISSUE #8601: perf: JUnitXmlTestsListener blocks on start up for 5s
## steps

```bash
sbt --client compile
```

## problem

In some environments, with slow DNS lookup like GitHub Actions macOS, this is blocking the startup supposedly for 5s.

## expectation

1. Don't block this until someone runs a test. (Use lazy val)
2. Cache at object level so we pay this tax once per session.

calm329:
## Summary
- Move hostname resolution to object-level lazy val to prevent blocking sbt startup
- On macOS with misconfigured hostname, `InetAddress.getLocalHost.getHostName` can take ~5 seconds due to DNS timeout
- Now resolution only happens once per session and is deferred until tests actually run

Fixes #8601

## Test plan
- [x] Compiles successfully
- [ ] CI passes
- [ ] Verify sbt startup no longer blocked by slow hostname resolution

### DIFF
--- a/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
+++ b/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala
@@ -30,6 +30,33 @@ import util.Logger
 import sbt.protocol.testing.TestResult
 
 /**
+ * Companion object for JUnitXmlTestsListener that caches the hostname lazily.
+ * This ensures hostname resolution only happens once per session and doesn't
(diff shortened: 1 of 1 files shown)


calm329:

eed3si9n (approved):
Thanks!

calm329:
All CI checks have passed

PR #8606, opened 2026-01-22
[2.x] fix: Race condition in WorkerExchangeTest.propBye

### LINKED ISSUE #8605: [2.x] flaky test: bye should return response json with a result
## steps

```
testOnly **.WorkerExchangeTest
```

## problem

```scala
[info] Using random seed: 323644772289
[info] + sbt.internal.WorkerExchangeTest.non-jsonrpc should return exit code 1: OK, passed 10 tests
[info] - sbt.internal.WorkerExchangeTest.bye should return response json with a result: Falsified after 2 passed tests
[info] > Tcp
[info] > 14
[info] > ""
```

calm329:
## Summary
- Fix flaky test `bye should return response json with a result` in WorkerExchangeTest
- Root cause: race condition between process exit and async response delivery
- Solution: use CountDownLatch to wait for the listener to receive the response

## Test plan
- [x] Verified with originally failing seed: `HEDGEHOG_SEED=323644772289`
- [x] Ran test 10 times (100 iterations) without failures
- [x] Full `actionsProj/test` suite passes

Fixes #8605

### DIFF
--- a/main-actions/src/test/scala/sbt/internal/WorkerExchangeTest.scala
+++ b/main-actions/src/test/scala/sbt/internal/WorkerExchangeTest.scala
@@ -41,6 +41,7 @@ object WorkerExchangeTest extends Properties:
     yield withListener: l =>
       w.println(s"""{"jsonrpc": "2.0", "method": "bye", "params": {}, "id": $i}""")
       val exitCode = w.blockForExitCode()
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #8611, opened 2026-01-22
fix: Drop other idle servers on exit

### LINKED ISSUE #8610: Drop other idle servers on exit
## steps

1. start an sbt session in one directory using sbt 2.x (defaults to sbtn), and `exit`
2. start another sbt session in another directory using sbt 2.x and `exit`
3. etc

## problem

Each session leaves behind an idle server for 7 days.

https://github.com/sbt/sbt/blob/2380ab84b6610986af0a407d1ea97b7a9bd8ab82/main/src/main/scala/sbt/Defaults.scala#L399

## expectation

1. When sbt server starts up a port file `active.json` is created. Create another file `1234.json` under `$HOME/Library/Caches/sbt/v2/proc` (the disk cache location for the OS + /proc) with the same content, where `1234` is the process id.
2. Upon `exit` send `dropIfIdle` message to other servers. Remove `2345.json` if it's unreachable.
3. The servers that received `dropIfIdle` message should shut itself down, if it's been idle for 600 seconds or more.
4. When the port file `active.json` is deleted, also delete `1234.json`.
5. This should reduce the idle server to one.

circlecrystalin:
Closes #8610


circlecrystalin:
@eed3si9n Could you please check my pr?

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -1091,6 +1091,120 @@ class NetworkClient(
...
+      if (parentPortfile.exists && parentPortfile != currentPortfile) {
+        portfiles += parentPortfile
+      }
+
+      // Check sibling directories
+      val siblings = parentDir.listFiles()
+      if (siblings != null) {
+        for (sibling <- siblings) {
+          if (sibling.isDirectory && sibling != currentDir) {
+            val siblingPortfile = sibling / "project" / "target" / "active.json"
```
eed3si9n: I thought the spec was clear enough that it revolves around 1234.json, not active.json. Was this generated with Gen AI assistance? If so, could you please check its output carefully next time?

eed3si9n (changes requested):

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
circlecrystalin: I just fixed! Changed to use `1234.json` as specified in the issue. Thanks for the catch! @eed3si9n

Issue #7164, opened 2023-03-01
Calling `outputFileChanges` throws `References to undefined settings` error


ptrdom:
## steps

build.sbt
```scala
ThisBuild / version := "0.1.0-SNAPSHOT"

ThisBuild / scalaVersion := "2.13.10"

val a = taskKey[Unit]("")
val b = taskKey[Unit]("")

lazy val root = (project in file("."))
  .settings(
    name := "sbt-file-outputs-test",
    a / fileOutputs += (baseDirectory.value / "a-output").toGlob,
    b := {
      a.outputFileChanges
    }
  )
```
sbt version `1.8.2`

## problem

Following error gets thrown upon running sbt:
```
[error] References to undefined settings: 
[error] 
[error]   a / outputFileStamps from b (/sbt-file-outputs-test/build.sbt:12)
[error] 
[error]   a / allOutputFiles from b (/sbt-file-outputs-test/build.sbt:12)
[error] 
[error]   a / changedOutputFiles from b (/sbt-file-outputs-test/build.sbt:12)

```

## expectation

As per how I understood the docs and expected`fileOutputs` to work, I assumed that no error would be thrown and calling `outputFileChanges` would succeed for this `build.sbt` file.

ptrdom:
Some additional observations.

Changing `build.sbt` to following:
```scala
ThisBuild / version := "0.1.0-SNAPSHOT"

ThisBuild / scalaVersion := "2.13.10"

val a = taskKey[Unit]("")
val b = taskKey[Unit]("")

lazy val root = (project in file("."))
  .settings(
    name := "sbt-file-outputs-test",
    a / fileOutputs += (baseDirectory.value / "a-output").toGlob,
    a := {
      ()
    },
    b := {
      a.outputFileChanges
    }
  )
```
Seems to add two of the three references:
```
[error] Reference to undefined setting: 
[error] 
[error]   a / changedOutputFiles from b (/sbt-file-outputs-test/build.sbt:15)
```

Changing `build.sbt`to following solves the issue:
```scala
ThisBuild / version := "0.1.0-SNAPSHOT"

ThisBuild / scalaVersion := "2.13.10"

val a = taskKey[Seq[java.nio.file.Path]]("")
val b = taskKey[Unit]("")

lazy val root = (project in file("."))
  .settings(
    name := "sbt-file-outputs-test",
    a / fileOutputs += (baseDirectory.value / "a-output").toGlob,
    a := {
      Seq.empty
    },
    b := {
[...]

PR #8612, opened 2026-01-22
Fix outputFileChanges undefined settings error

smartman-stack:
https://github.com/sbt/sbt/issues/7164

This PR fixes the issue where outputFileChanges throws an "undefined settings" error. The problem was caused by missing or improperly referenced settings in the configuration. The fix ensures that the settings are correctly initialized, preventing the error from occurring.


smartman-stack:
I just did sign CLA.

eed3si9n (approved):
Thanks for the contribution. lgtm, pending CI.

sbt-app/src/sbt-test/nio/unit-outputs-scoped/build.sbt
```diff
@@ -0,0 +1,18 @@
+import sbt.internal.FileChangesMacro.*
```
eed3si9n: It seems like the compiler gets confused with having the same import twice, one here and another in the package object.

smartman-stack: I just removed the duplicated imports.

smartman-stack:
Hi @eed3si9n. After get approval, I can't see the merge button. Does someone else merge the updates?

eed3si9n:
No, I have to approve each commit for first-time contributors.

PR #8613, opened 2026-01-22
[2.x] fix: Suppress "Multiple main classes" warning for runMain commands

### LINKED ISSUE #3739: Useless "Multiple main classes" detected when running runMain
When using `runMain` a useless warning is printed to the console multiple times per run:

```
akka-http-tests > ~ test:runMain foobar.ReproduceAkkaBug
[info] Formatting 1 Scala source {file:/home/johannes/git/opensource/akka-http/}akka-http-tests(test) ...
[warn] Multiple main classes detected.  Run 'show discoveredMainClasses' to see the list
[warn] Multiple main classes detected.  Run 'show discoveredMainClasses' to see the list
[info] Compiling 1 Scala source to /home/johannes/git/opensource/akka-http/akka-http-tests/target/scala-2.11/test-classes ...
[info] Done compiling.
[warn] Multiple main classes detected.  Run 'show discoveredMainClasses' to see the list
```

If you use `runMain` you are obviously aware of that fact.

sbt 1.0.3

calm329:
## Summary
- Fix warning appearing for scoped runMain commands using colon syntax (`test:runMain`)
- Extend warning suppression to all run command variants (`bgRunMain`, `fgRunMain`)

## Problem
When running `runMain` (especially scoped like `test:runMain`), the "Multiple main classes detected" warning was printed even though the user explicitly specified which main class to run.

## Solution
Changed the scope separator parsing from `/` only to `[/:]` to handle both new slash syntax (`Test/runMain`) and old colon syntax (`test:runMain`).

## Test
Added scripted test `run/run-main-warning` covering `runMain`, `bgRunMain`, `fgRunMain`, and scoped variants.

Fixes #3739

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1066,14 +1066,19 @@ object Defaults extends BuildCommon {
       selectMainClass := mainClass.value orElse askForMainClass(discoveredMainClasses.value),
       run / mainClass := (run / selectMainClass).value,
       mainClass := Def.uncached {
+        // Suppress warning for run commands (user is actively running, warning is noise)
+        def isRunCommand(s: String): Boolean = s match
+          case "run" | "runMain" | "bgRun" | "bgRunMain" | "fgRun" | "fgRunMain" => true
+          case _                                                                 => false
         val logWarning = state.value.currentCommand.forall(!_.commandLine.split(" ").exists {
-          case "run" | "runMain" => true
-          case r =>
-            r.split("/") match {
+          case s if isRunCommand(s) => true
+          case r                    =>
+            // Handle both "/" (new syntax like Test/run) and ":" (old syntax like test:run)
+            r.split("[/:]") match {
               case Array(parts*) =>
                 parts.lastOption match {
-                  case Some("run" | "runMain") => true
-                  case _                       => false
+                  case Some(s) if isRunCommand(s) => true
+                  case _                          => false
                 }
             }
         })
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1066,14 +1066,19 @@ object Defaults extends BuildCommon {
       selectMainClass := mainClass.value orElse askForMainClass(discoveredMainClasses.value),
       run / mainClass := (run / selectMainClass).value,
       mainClass := Def.uncached {
+        // Suppress warning for run commands (user is actively running, warning is noise)
+        def isRunCommand(s: String): Boolean = s match
+          case "run" | "runMain" | "bgRun" | "bgRunMain" | "fgRun" | "fgRunMain" => true
```
eed3si9n, line 1071: Are we no longer showing warnings for `run` because the user would be prompted for the main class?

calm329: The original code already suppressed for both `run` and `runMain` - I kept that behavior and extended it to the `bg*/fg*` variants.
If you'd prefer to only suppress for the explicit `*Main` variants, I can change it to:
```scala
case "runMain" | "bgRunMain" | "fgRunMain" => true
```

Let me know which behavior you'd prefer.

eed3si9n: > The original code already suppressed for both `run` and `runMain`

Doh, you're right. The change looks good to me then.

eed3si9n (approved):

Issue #6836, opened 2022-03-14
"update" doesn't honor offline mode setting


DavidPerezIngeniero:
## steps

SBT 1.6.1

```
set offline := true
update
```

## problem

I don't have connectivity with my VPN, so `update` hangs when trying to access some dependencies in this form:
```
projectDependencies +="cocoon" % "cocoon" % "2.1.11" from "https://myprivateserver.com/cocoon-2.1.11.jar"
```

## expectation

In offline mode doesn't try to update dependencies

## notes

https://alvinalexander.com/scala/setting-putting-sbt-into-offline-mode-no-wifi/

PR #8614, opened 2026-01-22
Fix update not honoring offline mode setting

smartman-stack:
https://github.com/sbt/sbt/issues/6836

This PR addresses the issue where the update function doesn't honor the offline mode setting. The fix ensures that the update process respects the offline mode configuration, preventing updates from being applied when offline mode is enabled.

Changes:

Modified the update function to check and honor the offline mode setting.

Added tests to verify the correct behavior when offline mode is active.


sbt-app/src/sbt-test/lm-coursier/offline-mode/build.sbt
```diff
@@ -0,0 +1,21 @@
+// Test for offline mode with URL dependencies
+libraryDependencies += "org.jsoup" % "jsoup" % "1.9.1" from "https://jsoup.org/packages/jsoup-1.9.1.jar"
```
eed3si9n:
```suggestion
libraryDependencies += ("org.jsoup" % "jsoup" % "1.9.1").from("https://jsoup.org/packages/jsoup-1.9.1.jar")
```

eed3si9n (changes requested):
Test seems to be failing.

Issue #6795, opened 2022-01-18
Adding extra jars to `sbt.extraClasspath` or `resources:` doesn't work


drywet:
## problem

Adding extra jars to `sbt.extraClasspath` or `resources:` doesn't work

## expectation

Sbt should start with the extra jars on its classpath,  more precisely, before making any https requests

PR #8615, opened 2026-01-22
Fix issue with adding extra jars to sbt.extraClasspath or resources

smartman-stack:
https://github.com/sbt/sbt/issues/6795

This PR resolves the issue where adding extra JARs to sbt.extraClasspath or resources does not work as expected. The fix ensures that the extra JARs are correctly included in the classpath during the build process.

Changes:

Updated the configuration for sbt.extraClasspath to properly include extra JARs.

Adjusted related build settings to ensure correct handling of resources.


eed3si9n (changes requested):
I haven't checked if the bug is still valid, but I think this is a bug that should be fixed in sbt/launcher.

PR #8618, opened 2026-01-22
Fix edge case in findNegativeCycle where node might not be in stack

jimcody1995:
This fix addresses a potential edge case in the cycle detection logic where stack.takeWhile might not correctly identify the cycle if the node is not found in the expected position. The fix uses indexWhere to locate the cycle start more reliably, with a fallback to the original logic for edge cases.

Added test cases that reproduce the issue and verify the fix works correctly.

### DIFF
--- a/util-collection/src/test/scala/sbt/internal/util/DagNegativeCycleSpec.scala
+++ b/util-collection/src/test/scala/sbt/internal/util/DagNegativeCycleSpec.scala
@@ -1,0 +1,97 @@
+/*
+ * sbt
+ * Copyright 2023, Scala center
+ * Copyright 2011 - 2022, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.internal.util
+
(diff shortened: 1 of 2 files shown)


util-collection/src/test/scala/sbt/internal/util/DagNegativeCycleSpec.scala
eed3si9n, line 1: Could you add the tests under `test/scala` directory please (`test/scala-2` are old files from Scala 2.x we'll remove)? Also, could you use Verify or HedgeHog for new tests? 
https://github.com/sbt/sbt/blob/develop/contributing-docs/04_unit_tests.md

jimcody1995: okay

eed3si9n (changes requested):
```
[error] (collectionProj / Test / scalafmtCheck) scalafmt: 1 files must be formatted (/home/runner/work/sbt/sbt/util-collection)
```

Could you try compiling the code, which should trigger the scalafmt. If not please run `scalafmtAll` to force the code formatting.

PR #8621, opened 2026-01-23
[2.x] feat: Add scriptedKeepTempDirectory setting

### LINKED ISSUE #3214: Scripted option to not remove temporary directory
## steps

I run scripted to debug my sbt plugin.

## problem

After the scripted tests complete, the temporary directories `/private/var/folder/adsfasda` are removed. Because the directory is removed, I can't `cd /private/var/...` to manually check what went wrong. To my best knowledge, there is no way to configure scripted to not delete the temporary directory.

## expectation

It would be nice if there was an option I could pass to scripted via `scriptedLaunchOpts` to tell it not to remove the tmp directory.

## notes

I work around this by issue adding `$ pause` in the scripted `test` and the `cp -r /private/var`. However, this is cumbersome and slow.

bitloi:
## feat: Add scriptedKeepTempDirectory setting to preserve temp directories

Fixes #3214

### Problem

When running scripted tests to debug sbt plugins, the temporary directories (`/private/var/folder/...`) are automatically deleted after tests complete. This makes it difficult to inspect the test state for debugging purposes, requiring workarounds like adding `$ pause` commands and manually copying directories.

### Solution

Added a new `scriptedKeepTempDirectory` setting that allows users to preserve temporary directories after scripted tests complete. When enabled, the temporary directory path is logged so users can inspect it.

Usage:
```scala
scriptedKeepTempDirectory := true
```

### Changes

- **ScriptedPlugin.scala**: Added `scriptedKeepTempDirectory` setting key (defaults to `false`)
- **ScriptedRun.scala**: Added `keepTempDirectory` parameter to `run()` method with backward-compatible versioning (V3 methods)
- **ScriptedTests.scala**: 
  - Added `keepTempDirectory` parameter to `run()`, `runInParallel()`, and `batchScriptedRunner()` methods
  - Modified `createTestRunners()` to conditionally use `IO.createTemporaryDirectory` instead of `IO.withTemporaryDirectory` when `keepTempDirectory` is true
  - Logs the temporary directory path when preserved

### Testing

- Compiles successfully
- Binary compatible (new methods added, no existing signatures changed)
- Backward compatible (defaults to `false`, existing behavior unchanged)

### Checklist

- [x] Compiles successfully
- [x] `scalafmtAll` applied
- [x] Binary compatible (maintains backward compatibility)
- [x] Follows coding style guidelines

### DIFF
--- a/scripted-sbt/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
+++ b/scripted-sbt/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
@@ -147,6 +147,12 @@ final class ScriptedTests(
         )
       logTests(runFromSourceBasedTests.size, prop.toString)
 
+      if (keepTempDirectory && runFromSourceBasedTests.size > 1) {
+        sys.error(
+          s"scriptedKeepTempDirectory requires exactly one test, but ${runFromSourceBasedTests.size} tests were requested"
+        )
+      }
+
       def createTestRunners(tests: Seq[TestInfo]): Seq[TestRunner] = {
         tests
           .sortBy(_._1)
--- a/scripted-sbt/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
+++ b/scripted-sbt/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
@@ -109,7 +109,8 @@ final class ScriptedTests(
       prescripted: File => Unit,
       sbtInstances: Int,
       prop: RemoteSbtCreatorProp,
-      log: Logger
+      log: Logger,
+      keepTempDirectory: Boolean = false,
   ): Seq[TestRunner] = {
     // Test group and names may be file filters (like '*')
     val groupAndNameDirs = {
@@ -151,8 +152,14 @@ final class ScriptedTests(
           .sortBy(_._1)
           .grouped(batchSize)
           .map { batch => () =>
-            IO.withTemporaryDirectory {
-              runBatchedTests(batch, _, prescripted, prop, log)
+            if (keepTempDirectory) {
+              val tempDir = IO.createTemporaryDirectory
+              log.info(s"Temporary directory for scripted tests: ${tempDir.getAbsolutePath}")
+              runBatchedTests(batch, tempDir, prescripted, prop, log)
+            } else {
+              IO.withTemporaryDirectory {
+                runBatchedTests(batch, _, prescripted, prop, log)
+              }
             }
           }
           .toList
@@ -352,7 +359,7 @@ object ScriptedTests extends ScriptedRunner {
       buffer,
       tests,
       logger,
-      Array(),
+      Array[String](),
       new java.util.ArrayList[File],
       defScalaVersion,
       sbtVersion,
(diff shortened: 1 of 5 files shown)


bitloi:
@eed3si9n Could you please review my PR?

eed3si9n (approved):
Thanks for the contribution! Overall the changes look good.
I probably need to grab this branch and test it manually.

bitloi:
@eed3si9n Can we merge this pr?

eed3si9n (changes requested):
I couldn't test this using sbt/sbt. For sbt/sbt's scripted test this funnels the call from sbt 1.x into sbt 2.x via reflection. Could you surface your new setting in project/Scripted.scala's `ScriptedKeys`, and fold that in to the `scriptedTask`. See also:

```scala
ThisBuild / scriptedBufferLog := true
ThisBuild / scriptedPrescripted := { _ => }
```

Thanks!

bitloi:
> I couldn't test this using sbt/sbt. For sbt/sbt's scripted test this funnels the call from sbt 1.x into sbt 2.x via reflection. Could you surface your new setting in project/Scripted.scala's `ScriptedKeys`, and fold that in to the `scriptedTask`. See also:
> 
> ```scala
> ThisBuild / scriptedBufferLog := true
> ThisBuild / scriptedPrescripted := { _ => }
> ```
> 
> Thanks!

@eed3si9n Fixed. Could you please review again?

eed3si9n (changes requested):
If I add `ThisBuild / scriptedKeepTempDirectory := true` to build.sbt, and run `scripted actions/compile` I see

```
[info] Running 1 / 1 (100.00%) scripted tests with LauncherBased(/Users/xxxx/sbt-launch.jar)
[info] Temporary directory for scripted tests: /var/folders/hg/2602nfrs2958vnshglyl3srw0000gn/T/sbt_6446e5ca
[info] Running actions/compile
```

but the directory only contains global cache. Does it work for you locally?

bitloi:
I found the issue - the test files were being deleted after each test run regardless of the `keepTempDirectory` setting. 

I've pushed a fix that:
1. Passes `keepTempDirectory` through to `runBatchedTests`
2. Conditionally skips the cleanup logic (lines 252-264) when `keepTempDirectory` is true

The cleanup code was deleting all files except `global` and `global-logging` directories after each test, which is why you only saw the global cache.

With this fix, when `scriptedKeepTempDirectory := true`, the test files should now remain in the temporary directory for debugging. Could you test it again?

scripted-sbt/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
```diff
@@ -241,16 +249,18 @@ final class ScriptedTests(
...
+          val base = tempTestDir.getCanonicalFile.toGlob
+          val global = base / "global"
+          val globalLogging = base / ** / "global-logging"
+          def recursiveFilter(glob: Glob): PathFilter = (glob: PathFilter) || glob / **
+          val keep: PathFilter = recursiveFilter(global) || recursiveFilter(globalLogging)
+          val toDelete = view.list(base / **, !keep).map(_._1).sorted.reverse
+          toDelete.foreach { p =>
+            try Files.deleteIfExists(p)
+            catch { case _: IOException => }
+          }
```
eed3si9n, line 269: While this fixes the situation for a single test, wouldn't this mess up the batch test? Could you add a check such that when keepTempDirectory is passed only one scripted test is requested?

bitloi:
Fixed, please review it again.

eed3si9n (approved):
Thanks!

xuwei-k:
- https://github.com/sbt/sbt/issues/8748

mkurz:
Very useful feature, thanks!

PR #8622, opened 2026-01-23
[2.x] fix: Throw on addCompilerPlugin(foo % Test)

### LINKED ISSUE #3203: addCompilerPlugin for test configuration should not affect compile configuration
## steps
```
$ sbt
[info] Loading global plugins from /Users/twer/.sbt/0.13/plugins
[info] Loading project definition from /private/tmp/project
[info] Set current project to tmp (in build file:/private/tmp/)
> ++2.12.2
[info] Setting version to 2.12.2
[info] Reapplying settings...
[info] Set current project to tmp (in build file:/private/tmp/)
> set addCompilerPlugin(("org.scalameta" % "paradise" % "3.0.0-M8").cross(CrossVersion.patch) % Test)
[info] Defining *:libraryDependencies
[info] The new value will be used by *:allDependencies, *:dependencyPositions
[info] Reapplying settings...
[info] Set current project to tmp (in build file:/private/tmp/)
> show test:scalacOptions
[info] Updating {file:/private/tmp/}tmp...
[info] Resolving jline#jline;2.14.3 ...
[info] Done updating.
[info] * -Xplugin:/Users/twer/.ivy2/cache/org.scalameta/paradise_2.12.2/jars/paradise_2.12.2-3.0.0-M8.jar
[success] Total time: 1 s, completed May 23, 2017 10:07:45 AM
> show compile:scalacOptions
[info] * -Xplugin:/Users/twer/.ivy2/cache/org.scalameta/paradise_2.12.2/jars/paradise_2.12.2-3.0.0-M8.jar
[success] Total time: 0 s, completed May 23, 2017 10:07:52 AM
```

## problem
`-Xplugin` is set for both  `test:scalacOptions` and `compile:scalacOptions`

## expectation
`-Xplugin` is only set for `test:scalacOptions`, not `compile:scalacOptions`

bitloi:
Fixes #3203

## Problem

When using `addCompilerPlugin((dependency) % Test)`, the compiler plugin was incorrectly added to BOTH `test:scalacOptions` AND `compile:scalacOptions`, instead of only `test:scalacOptions`.

## Solution

Throw when the dependency is scoped, since we do not support this use case.

### DIFF
--- a/lm-core/src/main/scala/sbt/librarymanagement/ConfigurationExtra.scala
+++ b/lm-core/src/main/scala/sbt/librarymanagement/ConfigurationExtra.scala
@@ -50,7 +50,6 @@ object Configurations {
   lazy val ScalaDocTool = Configuration.of("ScalaDocTool", "scala-doc-tool").hide
   lazy val ScalaReplTool = Configuration.of("ScalaReplTool", "scala-repl-tool").hide
   lazy val CompilerPlugin = Configuration.of("CompilerPlugin", "plugin").hide
-  lazy val TestCompilerPlugin = Configuration.of("TestCompilerPlugin", "test-plugin").hide
   lazy val Component = Configuration.of("Component", "component").hide
 
   private[sbt] val DefaultMavenConfiguration = defaultConfiguration(true)
--- a/lm-core/src/main/scala/sbt/librarymanagement/ConfigurationExtra.scala
+++ b/lm-core/src/main/scala/sbt/librarymanagement/ConfigurationExtra.scala
@@ -50,6 +50,7 @@ object Configurations {
   lazy val ScalaDocTool = Configuration.of("ScalaDocTool", "scala-doc-tool").hide
   lazy val ScalaReplTool = Configuration.of("ScalaReplTool", "scala-repl-tool").hide
   lazy val CompilerPlugin = Configuration.of("CompilerPlugin", "plugin").hide
+  lazy val TestCompilerPlugin = Configuration.of("TestCompilerPlugin", "test-plugin").hide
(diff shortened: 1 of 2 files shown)


bitloi:
@eed3si9n Could you please review my PR?

eed3si9n (changes requested):
Sorry not a fan of this feature. I feel like this happened once for someone, but not worth adding a new configuration for everyone permanently.

bitloi:
@eed3si9n Thank you for the quick review! I understand the concern about adding a permanent configuration for what may be an edge case.

Could you provide guidance on a preferred approach to fix issue #3203? I see a few alternatives:

1. **Filter within `autoPlugins`** - Instead of creating a new configuration, modify `autoPlugins` to inspect the original `libraryDependencies` and filter plugins based on their declared scope (e.g., `% Test`). This would avoid creating new configurations but requires accessing the dependency metadata at the point where `scalacOptions` is computed.

2. **Modify configuration mapping** - Change how `compilerPlugin` encodes scope information in the existing `plugin` configuration mapping, perhaps using dependency attributes or classifiers to preserve the original scope.

3. **Document as expected behavior** - If this is considered an edge case not worth fixing, we could document the limitation and suggest workarounds (e.g., manually managing `scalacOptions` for configuration-specific plugins).

Which direction would you prefer, or do you have another approach in mind? I'm happy to revise the implementation based on your guidance.

eed3si9n:
3. I guess. If you'd like you can repurpose this PR to throw an error if someone tries to do:

```scala
addCompilerPlugin(dependency % Test)
```

eed3si9n (approved):
Thanks!

Issue #7639, opened 2024-09-06
[2.x] Ivyless publishing


eed3si9n:
If we want to eventually drop Apache Ivy dependency, we'd need to implement our own JVM publishing without using Ivy.
For testing purpose it would be great if we can support:

- `publishM2` (local Maven repo)
- [x] `publishLocal` (local Ivy repo) - https://github.com/sbt/sbt/issues/8631
- `publish` with  `publishMavenStyle` (Sonatype etc)
- [x] `publish` without `publishMavenStyle` (private Artifactory repos?) - https://github.com/sbt/sbt/issues/8639
- `publishSigned` (Maven Central with signing)

Likely this is a feature that is duplicated cross bunch of build tools.

DeborahOlaboye:
Hi @eed3si9n , I'd like to work on this issue.

eed3si9n:
@DeborahOlaboye I won't stop you, but this is a substantial feature, so I suggest tackling a few smaller issues first.

SethTisue:
relevant: https://github.com/scalacenter/advisoryboard/blob/main/proposals/034-artifact-publishing.md  (but I don't think there's much likelihood of a Scala Center engineer becoming available to work on it)

eed3si9n:
Maybe to make this a bit more managable, I should open tickets for each of the scenarios covered in this epic ticket.

DeborahOlaboye:
@eed3si9n that sounds good, once the scenarios are split into individual tickets, I can pick one to start with.

eed3si9n:
I've created https://github.com/sbt/sbt/issues/8631.

eed3si9n:
@calm329 implemented `publishLocal`, so I've created https://github.com/sbt/sbt/issues/8639 as the next task.

PR #8634, opened 2026-01-26
[2.x] feat: Implement ivyless publishLocal

### LINKED ISSUE #8631: [2.x] Ivyless publishLocal
This is part of https://github.com/sbt/sbt/issues/7639.

## Expectations

Implement Ivyless `publishLocal` task, which publishes to the local Ivy repo.

1. Setup a scripted test  such that `publishLocal` can publish to an arbitrary named directory under `target` (see note below)
2 Create a global setting `useIvy`, which defaults to `true` for now.
3. Run scripted test such that when `useIvy` is set to `false` it will produce the identical set of files.
4. Extend the implementation to support the customization of artifacts. For example, the user can skip publishing Scaladocs by setting `Compile / packageDoc / publishArtifact := false`.

## Note

```scala
name := "lib1"

organization := "com.example"
version := "0.1.0-SNAPSHOT"
ivyPaths := IvyPaths(
  ((LocalRootProject / baseDirectory).value / "ivy").toString,
  Some(((LocalRootProject / target).value / "ivy" / "cache").toString)
)
```

```bash
$ tree target/out/jvm/scala-3.7.4/lib1/ivy/cache/local/
target/out/jvm/scala-3.7.4/lib1/ivy/cache/local/
└── com.example
    └── lib1_3
        └── 0.1.0-SNAPSHOT
            ├── docs
            │   ├── lib1_3-javadoc.jar
            │   ├── lib1_3-javadoc.jar.md5
            │   └── lib1_3-javadoc.jar.sha1
            ├── ivys
            │   ├── ivy.xml
            │   ├── ivy.xml.md5
            │   └── ivy.xml.sha1
            ├── jars
            │   ├── lib1_3.jar
            │   ├── lib1_3.jar.md5
            │   └── lib1_3.jar.sha1
            ├── poms
            │   ├── lib1_3.pom
            │   ├── lib1_3.pom.md5
            │   └── lib1_3.pom.sha1
            └── srcs
                ├── lib1_3-sources.jar
                ├── lib1_3-sources.jar.md5
                └── lib1_3-sources.jar.sha1

9 directories, 15 files
```

sbt uses a forked version Ivy, which we manage in https://github.com/sbt/ivy.

## Implementation notes

1. Please follow https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md carefully.
2. Please follow https://github.com/sbt/sbt/blob/develop/contributing-docs/07_tech_stack.md where possible. For example, datatype should be generated using Contraband , and HTTP client should use Gigahorse.
3. See https://www.scala-sbt.org/1.x/docs/Artifacts.html for the details of artifacts.
4. See https://ant.apache.org/ivy/history/2.3.0/index.html for general information about Apache Ivy.

calm329:
## Summary

Implements ivyless `publishLocal` task as part of the effort to remove Ivy dependency (#7639).

Fixes #8631

**Changes:**
- Add `useIvy` setting key (defaults to `true`)
- Add `ivylessPublishLocalImpl` helper that publishes without Ivy
- Modify `publishLocal` to use ivyless publisher when `useIvy := false`
- Generate ivy.xml via `lmcoursier.IvyXml`
- Generate MD5/SHA-1 checksums for all files
- Add scripted test `dependency-management/ivyless-publish-local`

**Files changed:**
- `main/src/main/scala/sbt/Keys.scala` - new setting key
- `main/src/main/scala/sbt/Defaults.scala` - implementation
- `sbt-app/src/sbt-test/dependency-management/ivyless-publish-local/` - scripted test

## Test plan

- [x] `mainProj/compile` passes
- [x] `mainProj/test` passes (91 tests)
- [x] `scalafmtCheckAll` passes
- [x] `mimaReportBinaryIssues` passes
- [x] `scripted dependency-management/ivyless-publish-local` passes
- [x] Artifact customization works (`packageDoc / publishArtifact := false`)

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2977,34 +2977,7 @@ object Classpaths {
     deliverLocal := deliverTask(makeIvyXmlLocalConfiguration).value,
     makeIvyXml := deliverTask(makeIvyXmlConfiguration).value,
     publish := publishOrSkip(publishConfiguration, publish / skip).value,
-    publishLocal := Def.uncached {
-      val log = streams.value.log
-      val skipValue = (publishLocal / skip).value
-      val useIvyValue = useIvy.value
-      if (skipValue) {
-        val ref = thisProjectRef.value
-        logSkipPublish(log, ref)
-      } else if (useIvyValue) {
-        // Use the existing Ivy-based publisher
-        val conf = publishLocalConfiguration.value
-        val module = ivyModule.value
-        val publisherInterface = publisher.value
-        publisherInterface.publish(module, conf, log)
-      } else {
-        // Use the ivyless publisher
-        val project = csrProject.value.withPublications(csrPublications.value)
-        val config = publishLocalConfiguration.value
-        val artifacts = config.artifacts.map { case (a, f) => (a, f) }
-        val checksumAlgos = config.checksums
-        val ivyHome = ivyPaths.value.ivyHome.map(new File(_)).getOrElse {
-          val userHome = new File(System.getProperty("user.home"))
-          userHome / ".ivy2"
-        }
-        val localRepoBase = ivyHome / "local"
-        val overwrite = config.overwrite
-        ivylessPublishLocalImpl(project, artifacts, checksumAlgos, localRepoBase, overwrite, log)
-      }
-    },
+    publishLocal := LibraryManagement.ivylessPublishLocalTask.value,
     publishM2 := publishOrSkip(publishM2Configuration, publishM2 / skip).value,
     credentials ++= Def.uncached {
       val alreadyContainsCentralCredentials: Boolean = credentials.value.exists {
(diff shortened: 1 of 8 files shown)


calm329:

eed3si9n:
I don't have time to do a detailed review, but pretty excited about this one. Thanks for the contribution!

calm329:
> I don't have time to do a detailed review, but pretty excited about this one. Thanks for the contribution!

Feel free to do it when you get a chance

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2976,7 +2977,34 @@ object Classpaths {
     deliverLocal := deliverTask(makeIvyXmlLocalConfiguration).value,
     makeIvyXml := deliverTask(makeIvyXmlConfiguration).value,
     publish := publishOrSkip(publishConfiguration, publish / skip).value,
-    publishLocal := publishOrSkip(publishLocalConfiguration, publishLocal / skip).value,
+    publishLocal := Def.uncached {
```
eed3si9n: For compartmentalization, could we move both the task and the imple function into https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/internal/LibraryManagement.scala please?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2976,7 +2977,34 @@ object Classpaths {
...
+      val log = streams.value.log
+      val skipValue = (publishLocal / skip).value
+      val useIvyValue = useIvy.value
+      if (skipValue) {
```
eed3si9n: By default, `Def.uncache(...)` would end up triggering all `foo.value` that appears in the task body. The macro would do a special handling if the `if` expression is the outer most expression in the task to expand to a [Selective functor](https://eed3si9n.com/selective-functor-in-sbt/) aka `Def.ifS(Def.task { ... })(Def.task { ... })(Def.task { ... })`, which doesn't trigger the `then`-side or `else`-side. The complication is that we would want nested if, one for skip, and another for `useIvy`, so we might need to use `Def.ifS` and nest them.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -3787,6 +3815,104 @@ object Classpaths {
...
+      case "jar"                                   => "jars"
+      case "src" | "source" | "sources"            => "srcs"
+      case "doc" | "docs" | "javadoc" | "javadocs" => "docs"
+      case "pom"                                   => "poms"
+      case "ivy"                                   => "ivys"
+      case other                                   => other + "s"
+    }
+
+    // Helper to compute hash using MessageDigest
+    def computeHash(file: File, algorithm: String): String = {
```
eed3si9n: `sbt.util.Digest(Digest.Sha1, path)` can be used from https://github.com/sbt/sbt/blob/develop/util-cache/src/main/scala/sbt/util/Digest.scala, although `Digest` probably need a method added to get just the hash string.

calm329:
@eed3si9n I've fixed all. Would you mind taking a second look?

eed3si9n (approved):
Thanks for the quick turnaround. lgtm!

Issue #8022, opened 2025-01-29
[2.x] updateSbtClassifiers task fails with error "Error downloading default:<projectName>-build_sbt2.0.0-M3_3:0.1.0-SNAPSHOT"


unkarjedy:
## steps

1. Create a new sbt 2.0.0-M3 project with name `testProject`.
Rename the directory to `testProjectDirName` (this will be important later just for debugging purposes)
```sh
sbt new scala/scala3.g8 --name=testProject
mv testProject testProject_DirName
cd testProject_DirName
echo "sbt.version=2.0.0-M3" > project/build.properties
```

2. Run `sbt "show updateSbtClassifiers ; last updateSbtClassifiers"`
(using "last" to show the error details right away)

## problem
sbt tries to download an artifact named "testprojectdirname-build_sbt2.0.0-M3_3" from all known repositories. 
The artifact is not published, and it obviously fails with error `Error downloading default:testproject-build_sbt2.0.0-M3_3:0.1.0-SNAPSHOT`. 
Full error output:
```
[error] (updateSbtClassifiers) sbt.librarymanagement.ResolveException: Error downloading default:testprojectdirname-build_sbt2.0.0-M3_3:0.1.0-SNAPSHOT
[error]   Not found
[error]   not found: /Users/dmitrii.naumenko/.ivy2/local/default/testprojectdirname-build_sbt2.0.0-M3_3/0.1.0-SNAPSHOT/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/default/testprojectdirname-build_sbt2.0.0-M3_3/0.1.0-SNAPSHOT/testprojectdirname-build_sbt2.0.0-M3_3-0.1.0-SNAPSHOT.pom
[error]   not found: https://repo.scala-sbt.org/scalasbt/maven-releases/default/testprojectdirname-build_sbt2.0.0-M3_3/0.1.0-SNAPSHOT/testprojectdirname-build_sbt2.0.0-M3_3-0.1.0-SNAPSHOT.pom
[error]   not found: https://repo.scala-sbt.org/scalasbt/maven-snapshots/default/testprojectdirname-build_sbt2.0.0-M3_3/0.1.0-SNAPSHOT/testprojectdirname-build_sbt2.0.0-M3_3-0.1.0-SNAPSHOT.pom
[...]

Issue #7743, opened 2024-10-09
[2.x] global plugin doesn't work


eed3si9n:
## steps

```scala
// $HOME/.sbt/2.0.0-M2/plugins/user.sbt

addSbtPlugin("com.eed3si9n" % "sbt-vimquit" % "0.1.1")
```

## problem

The plugin doesn't get loaded.

## expectation

Machine-wide plugin is enabled.

anatoliykmetyuk:
When the plugins are in `$HOME/.sbt/2/plugins/user.sbt`, the resolution happens correctly - tested on `2.0.0-M2`, `2.0.0-RC6`. On `2.0.0-RC6-bin-SNAPSHOT` it does not resolve plugins from `.sbt/2/plugins` since the binary version is not `2` but full `2.0.0-RC6-bin-SNAPSHOT`.

Do we also need to support a fallback to the full version path, e.g. first look at `$HOME/.sbt/2/plugins/`, then at `$HOME/.sbt/2.0.0-RC6/plugins/`? sbt 1.x does not seem to do such a fallback, loading the plugins only from the binary version folder.

anatoliykmetyuk:
> When the plugins are in $HOME/.sbt/2/plugins/user.sbt, the resolution happens correctly - tested on 2.0.0-M2, 2.0.0-RC6.

So far can't reliably reproduce this, so this may still be an issue.

anatoliykmetyuk:
> When the plugins are in $HOME/.sbt/2/plugins/user.sbt, the resolution happens correctly - tested on 2.0.0-M2, 2.0.0-RC6

I was able to confirm this behavior on 2.0.0-RC7. Since this behavior is the same as with sbt 1.x, closing this issue.

PR #8638, opened 2026-01-26
[2.x] Fix global plugins not loaded 

### LINKED ISSUE #8600: [2.x] `*.sbt` files in a global plugin directory are not loaded
I have an sbt file inside `~/.sbt/2/plugins/ `and `~/.sbt/1.0/plugins`. These files only contain a build info plugin declaration: 

> addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.13.1") 

When I start an sbt 1.x project and type `plugins`, I can see that the BuildInfo plugin is loaded. I can use settings like `buildInfoKeys`.  
However, when I start an sbt 2.x project and type `plugins`, the build info plugin isn't available, and I see errors when trying to use something like `buildInfoKeys`.

What's interesting is that if I make a typo in the sbt file inside `~/.sbt/2/plugins/`, an exception is thrown, so sbt reads/detects the file - but it seems like it's not actually loaded.

azdrojowa123:
Fixes #8600


main/src/main/scala/sbt/internal/Load.scala
```diff
@@ -433,7 +433,7 @@ private[sbt] object Load {
               yield ((ref / ConfigKey(c.name) / configuration) :== c)
           val builtin: Seq[Setting[?]] =
             (thisProject :== project) +: (thisProjectRef :== ref) +: defineConfig
-          val settings = builtin ++ injectSettings.project ++ project.settings
+          val settings = builtin ++ project.settings ++ injectSettings.project
```
eed3si9n, line 436: I'm not sure why this would fix the situation.
I wonder if the bug is elsewhere like https://github.com/sbt/sbt/blob/cf8899919de84945ed844bff88e5a54e21dd6e94/main/src/main/scala/sbt/internal/Load.scala#L1206 ?

azdrojowa123: The logic behind this is that changing this order impacts how `projectDescriptors` is evaluated. The expression `projectDescriptors ~= { _ ++ gp.descriptors }` comes from `injectSettings.project`, so when it was placed before `project.settings`, it was overwritten (`gp.descriptors` is important because it contains the global plugins module).  
The `projectDescriptors` value is then used in the `csrExtraProjects` task, which is later used for extra projects in `lmcoursier.CoursierDependencyResolution#update`.  

I'm not 100% sure if this is correct, but while debugging, there was a difference between sbt 1.x and 2.x. Right after this change, the global plugins started being loaded.  

However, I don't know why the dependency-management tests are failing on CI - locally they work.
Screenshot 2026-01-27 at 09 49 14

eed3si9n: I'm rerunning the failed test. Maybe they are just flaky.

azdrojowa123: I reran them yesterday a few times :( (the same failed each time). 
So most likely something is wrong with the implementation or CI.

eed3si9n: If we placed `injectSettings.project` afterwards, then would that mean that the settings that come from user-wide settings would take higher precedence over what's in the `settings(...)` in build.sbt file?

azdrojowa123: It looks like it works this way even without changes from this PR (or maybe I misconfigured something).  

In the project with sbt version that contains my fixes, when I set `version := "2.0"` at the very top of `build.sbt`, and in the sbt file in `~/.sbt/2` I added `version := "1.0"`, the version for the root project was `1.0`. When I put `version := "2.0"` specifically in the root project using `settings(...)`, the version still showed as `1.0`.  

It behaves the same way in sbt RC8.

azdrojowa123: I managed to trigger the same exception locally as the one present on CI - I just ran `updateSbtClassifiers` task in a fresh project with the sbt version that includes the fixes.

azdrojowa123: I see there's a difference between sbt 1.x and sbt 2.x in `sbt.Classpaths.classifiersModuleTask` - in sbt 2.x, `pluginJars` includes global plugin jar, whereas in sbt 1.x, it doesn't. I'll debug this issue further. The symptoms look similar to  https://github.com/sbt/sbt/issues/8022.

main/src/main/scala/sbt/internal/Load.scala
```diff
@@ -1452,11 +1452,14 @@ private[sbt] object Load {
       pluginData: PluginData
   ): LoadedPlugins = {
     val definitionClasspath = pluginData.definitionClasspath
-    val dependencyClasspath = pluginData.dependencyClasspath
+    val dependencyClasspath = config.globalPlugin match {
+      case Some(gp) => removeEntries(pluginData.dependencyClasspath, gp.data.internalClasspath)
```
azdrojowa123: I'm not 100% sure about this because `GlobalPluginData#internalClasspath` contains not only the `exportedProducts` but also the value from `internalDependencyClasspath` - maybe there's some use case where it's required to keep this in the `dependencyClasspath`.

eed3si9n: https://www.scala-sbt.org/1.x/docs/Plugins.html#Global+plugins

> The `$HOME/.sbt/1.0/plugins/ directory` is treated as a global plugin definition project. It is a normal sbt project whose classpath is available to all sbt project definitions for that user as described above for per-project plugins.

If I'm reading it right, I think it's saying that the way sbt encodes "global plugin" is that we will first treat it as a standalone subproject. Then, we will pretend that user's metabuild (`project/plugins.sbt`) has a subproject dependency (aka internal dependency) to it.

azdrojowa123: I see. 

Do you have any idea for a test that would verify the dependency between the meta-build project and the "global plugin" project? 
Even if I make changes, it would be good to have a test that fails with the current implementation.

eed3si9n: When I first started sbt 2.x, I moved a bunch of scripted tests to `project1` to comment them out. In there there's https://github.com/sbt/sbt/tree/develop/sbt-app/src/sbt-test/project1/global-plugin. Not sure if that currently works, but it might be related?

azdrojowa123: Are the tests inside `project1` not run on CI?
If yes, then I have to move my test ;p 

^ thank you, this test is fine – it fails with my recent commit.

azdrojowa123:
On my local machine, the tests pass, but locally I have a different version without `bin-SNAPSHOT`.  
Maybe some changes to handling binary versions for the current schema `2.0.0-RC8-bin-SNAPSHOT` are needed.
I saw this [comment](https://github.com/sbt/sbt/issues/7743#issuecomment-3625693203) also.

eed3si9n:
> When the plugins are in `$HOME/.sbt/2/plugins/user.sbt`, the resolution happens correctly - tested on `2.0.0-M2`, `2.0.0-RC6`. On `2.0.0-RC6-bin-SNAPSHOT` it does not resolve plugins from `.sbt/2/plugins` since the binary version is not `2` but full `2.0.0-RC6-bin-SNAPSHOT`.
>
> Do we also need to support a fallback to the full version path, e.g. first look at `$HOME/.sbt/2/plugins/`, then at `$HOME/.sbt/2.0.0-RC6/plugins/`? sbt 1.x does not seem to do such a fallback, loading the plugins only from the binary version folder.

ok, so it's just not possible or very difficult to work on global-plugin-related problems until we release 2.0.

Malax:
@eed3si9n now with 2.0.0 being out (congrats!), can we pick up this PR again? It's currently blocking me to add sbt 2.x support to Heroku's buildpack for sbt. Happy to help to get this over the finish line, just let me know what I can do.

eed3si9n:
@Malax Here's my attempt at this - https://github.com/sbt/sbt/pull/9391

PR #8641, opened 2026-01-27
[2.x] fix: Add null check for resource URL

bohdansolovie:
**Problem**
getPropertiesFor method calls getResource().openStream() without checking if getResource returns null. When a resource doesn't exist, this causes a NullPointerException with no context about which resource was missing.

**Solution**
Added null check before calling openStream() and throw a descriptive FileNotFoundException with the resource name if the resource is not found. This prevents NullPointerException and provides better error messages.

### DIFF
--- a/zinc-lm-integration/src/main/scala/sbt/internal/inc/ResourceLoader.scala
+++ b/zinc-lm-integration/src/main/scala/sbt/internal/inc/ResourceLoader.scala
@@ -14,7 +14,11 @@ import java.util.Properties
 private[inc] object ResourceLoader {
   def getPropertiesFor(resource: String): Properties = {
     val properties = new Properties
-    val propertiesStream = getClass.getResource(resource).openStream
+    val resourceUrl = getClass.getResource(resource)
+    if (resourceUrl eq null) {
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

eed3si9n:
@bohdansolovie Could you sign the CLA, and let me know, so I can rerun the CLA checker?

bohdansolovie:
> @bohdansolovie Could you sign the CLA, and let me know, so I can rerun the CLA checker?

I already signed up.

eed3si9n:
https://contribute.akka.io/contribute/cla/scala/check/bohdansolovie is returning false. Could you try again from <https://contribute.akka.io/contribute/cla/scala>? Otherwise, I'd have to close the PR.

bohdansolovie:
> https://contribute.akka.io/contribute/cla/scala/check/bohdansolovie is returning false. Could you try again from https://contribute.akka.io/contribute/cla/scala? Otherwise, I'd have to close the PR.

it returns true.

bohdansolovie:
Screenshot_10

PR #8646, opened 2026-01-27
[2.x] perf: Add Analysis caching to ClassStamper to improve incremental test…

### LINKED ISSUE #7751: [2.x] Question about `ClassStamper.analyses` performance
In

https://github.com/sbt/sbt/blob/733bcfb23997930915b563e7d27b1a1f6c0490da/main/src/main/scala/sbt/internal/IncrementalTest.scala#L147-L156

https://github.com/sbt/sbt/blob/733bcfb23997930915b563e7d27b1a1f6c0490da/main/src/main/scala/sbt/internal/BuildDef.scala#L89-L100

When `classpath` is long, would calling `BuildDef.extractAnalysis` for each classpath entry to be too expensive? Analysis deserialization takes >100ms for large projects as per latest Zinc benchmarks. Even if we have just 10  `classpath` entries, it would take an entire second.

I am unfamilar with `sbt` codebase so I might be wrong but I guess each entry in `classpath` corresponds to a class file used for testing therefore `classpath` can get quite large?

dovansy1998:
Fixes: #7751 
Ref https://github.com/sbt/sbt/pull/8364/

### Problem
While BuildDef.extractAnalysis has caching at the AnalysisContents level, it still performs file I/O and deserialization work for each call. For test classpaths that can contain many entries (especially in multi-module projects), this becomes a significant performance bottleneck.
### Solution
Added a dedicated Analysis-level cache in ClassStamper that:

1. Caches deserialized Analysis objects keyed by VirtualFileRef.id()
2. Validates cache entries using file metadata (lastModified and sizeBytes) to ensure freshness
3. Uses Caffeine cache with 100MB memory limit and size-based eviction
4. Reuses BuildDef.extractAnalysis but adds an additional caching layer to avoid repeated deserialization


eed3si9n:
Thanks for the contributions!

For reference, I worked on a series of PRs related to perf recently and wrote a blog post (https://eed3si9n.com/sudori-part8-noop-compilation-performance/). In there I used sbt/sttp (https://github.com/sbt/sttp) as microbenchmark. For whatever the scenario, like no-op compilation or chainging 1 file etc, could you include before/after stats by running it 7 times in a row, and discarding the first two please?

dovansy1998:
### Before
==========================================
Benchmarking: test
Running 7 iterations (discarding first 2)
==========================================

[Warmup 1/2] Running test...
[Warmup 2/2] Running test...

Measuring runs:

[Run 3/7] Running test...
  ✓ 4ms
[Run 4/7] Running test...
  ✓ 4ms
[Run 5/7] Running test...
  ✓ 4ms
[Run 6/7] Running test...
  ✓ 4ms
[Run 7/7] Running test...
  ✓ 3ms

==========================================
Results (milliseconds):
  4
  4
  4
  4
  3

Statistics:
  Min: 3ms
  Max: 4ms
  Average: 3.8ms
==========================================


### After
==========================================
Benchmarking: test
Running 7 iterations (discarding first 2)
==========================================

[Warmup 1/2] Running test...
[Warmup 2/2] Running test...

Measuring runs:

[Run 3/7] Running test...
  ✓ 1ms
[Run 4/7] Running test...
  ✓ 2ms
[Run 5/7] Running test...
  ✓ 2ms
[Run 6/7] Running test...
  ✓ 1ms
[Run 7/7] Running test...
  ✓ 2ms

==========================================
Results (milliseconds):
  1
  2
  2
  1
  2

Statistics:
  Min: 1ms
  Max: 2ms
  Average: 1.6ms
==========================================

main/src/main/scala/sbt/internal/IncrementalTest.scala
```diff
@@ -152,8 +155,56 @@ class ClassStamper(
...
+        .maximumWeight(analysisCacheByteSize)
+        .weigher(weigher)
+        .build()
+
+    def getOrElseUpdate(
+        ref: VirtualFileRef,
+        lastModified: Long,
+        sizeBytes: Long
+    )(value: => Option[Analysis]): Option[Analysis] =
+      Option(cache.getIfPresent(ref.id())) match
```
eed3si9n: If this line is the key to the speedup, I suggest fixing the implementation in https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/internal/BuildDef.scala#L121 rather than adding yet another caffeine cache.

eed3si9n (changes requested):

dovansy1998:
Hi, @eed3si9n.
Please review this PR #8646.

main/src/main/scala/sbt/internal/BuildDef.scala
```diff
@@ -127,11 +127,22 @@ private[sbt] object BuildDef:
...
           val sizeBytes = attrs.size()
-          getOrElseUpdate(ref, lastModified, sizeBytes)(fallback(file))
+          getOrElseUpdate(ref, lastModified, sizeBytes)(fallback(file))._1
       catch case _: NoSuchFileException => fallback(file)
+    def getCachedAnalysis(ref: VirtualFileRef): Option[CompileAnalysis] =
+      val path = converter.toPath(ref)
+      val file = path.toFile()
+      try
+        val attrs = Files.readAttributes(path, classOf[BasicFileAttributes])
+        if attrs.isDirectory then None
```
eed3si9n: What's going on here?

eed3si9n:
I don't see how the proposed changes are going to contribute to perf improvement. At least there hasn't been a clear explanation for it. I'm closing this.

PR #8647, opened 2026-01-28
fix: Implement retry mechanism for file deletion on Windows

### LINKED ISSUE #7960: [1.x] `reboot dev` has no effect on Windows?
## steps

Build a sbt snapshot build
```
[IJ]reboot dev
[info] deleting C:\Users\Jerry Tan\.sbt\boot\scala-2.12.20\org.scala-sbt\sbt\1.10.7-SNAPSHOT
```

## problem

Despite it says `deleting C:\Users\Jerry Tan\.sbt\boot\scala-2.12.20\org.scala-sbt\sbt\1.10.7-SNAPSHOT`, the directory is not deleted.

## expectation

The directory should be deleted

xfactor-toml:
Add a retry mechanism to the file deletion process in MainLoop to handle locked files on Windows.
This change addresses issues where files may be inaccessible due to the classloader.

Fixes #7960

### DIFF
--- a/main/src/main/scala/sbt/MainLoop.scala
+++ b/main/src/main/scala/sbt/MainLoop.scala
@@ -21,6 +21,7 @@ import sbt.internal.util.{
   Prompt,
   Terminal as ITerminal
 }
+import sbt.internal.io.Retry
 import sbt.io.{ IO, Using }
 import sbt.protocol.*
 import sbt.util.{ Logger, LoggerContext }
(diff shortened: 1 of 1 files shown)


eed3si9n:
IO.delete already has retries - https://github.com/sbt/io/blob/caac00c8a682502ea7134fb18cf74205960379e2/io/src/main/scala/sbt/io/IO.scala#L568

PR #8649, opened 2026-01-28
[2.x] feat: Per-channel project cursor for multi-client server mode

### LINKED ISSUE #2799: Channel cursor
Given https://github.com/sbt/sbt/issues/2798
## steps
1. Start sbt server on terminal1 where a build is located: sbt server
2. Start sbt client on terminal2 from anywhere like ~/temp/: sbt client localhost:5001
3. On terminal1 `project echo-server`
4. On terminal2 `project echo-client`
5. On terminal1 type `name`
6. On terminal2 type `name`
## expectation

I expect terminal1 and terminal2 to return "echo-server" and "echo-client" respectively.
## note

Each channel will hold cursor, which points to the "current project".
This probably needs to be lifted out of the state.

bitloi:
Fixes #2799

## Problem
When multiple clients connect to the same sbt server, they all share the same "current project" state. When one client switches projects with `project X`, all other clients see that change.

## Solution
Store per-channel project cursors in State attributes. Each client maintains its own cursor that tracks which project it has selected.

## Changes
- `Keys.scala` - Add `channelProjectCursors` attribute
- `ProjectNavigation.scala` - Add `effectiveCurrentRef` to resolve channel-specific cursor
- `ProjectExtra.scala` - Use `effectiveCurrentRef` in `Project.extract`
- `MainLoop.scala` - Add `syncChannelCursor` to auto-update cursors on project change

## Test
- `ChannelCursorTest` - Verifies two clients maintain independent project states


bitloi:
@eed3si9n Please review my pr and give me your feedback

eed3si9n:
Hi @bitloi, thanks for the contribution!

Please check out our [Contributor's Guildeline](https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md) and:
3. Follow our Gen-AI assisted contributions guideline (please declare Gen AI usages in the PR description, if any)

main/src/main/scala/sbt/internal/ProjectNavigation.scala
```diff
@@ -38,7 +38,7 @@ final class ProjectNavigation(s: State) {
         show(); s
       case Some(BuildRef(uri))       => changeBuild(uri)
       case Some(ProjectRef(uri, id)) => selectProject(uri, id)
-      /*			else if(to.forall(_ == '.'))
+      /*				else if(to.forall(_ == '.'))
```
eed3si9n: this PR just changes comments. what's going on?

eed3si9n (changes requested):

main/src/main/scala/sbt/internal/ProjectNavigation.scala
bitloi: Sorry, can you review the update again?

bitloi:
@eed3si9n Would you please give me your feedback for the change?

main/src/main/scala/sbt/internal/ProjectNavigation.scala
```diff
@@ -11,21 +11,49 @@ package internal
...
+      .orElse(StandardMain.exchange.currentExec.flatMap(_.source).map(_.channelName))
+
+  def getChannelCursor(s: State): Option[ProjectRef] =
+    for {
+      channelName <- getChannelName(s)
+      cursors <- s.get(channelProjectCursors)
+      ref <- cursors.get(channelName)
+    } yield ref
+
+  def setChannelCursor(s: State, ref: ProjectRef): State =
```
eed3si9n, line 35: The general direction of using State sounds good. Let me check this out and try locally.

eed3si9n (approved):
@bitloi Works great. Thanks!

PR #8652, opened 2026-01-28
Fix output strategy handling in NetworkClient client-side runs

bohdansolovie:
Fixes TODO comments by implementing proper output strategy handling for client-side process execution in NetworkClient.

**Changes:**
- Replace `None` outputStrategy with `BufferedOutput(log)` for both JVM and native runs
- Buffers output until completion, then logs through console interface
- Prevents output interleaving with NetworkClient's synchronized output handling
- Ensures consistent logging behavior

**Why BufferedOutput?** BufferedOutput:
- Matches the TODO requirement
- Prevents interleaving issues
- Provides controlled, consistent output through the logger

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -70,6 +70,7 @@ import Serialization.{
 import NetworkClient.Arguments
 import java.util.concurrent.TimeoutException
 import sbt.util.Logger
+import sbt.OutputStrategy
 
 trait ConsoleInterface {
   def appendLog(level: Level.Value, message: => String): Unit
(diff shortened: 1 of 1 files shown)


eed3si9n:
Thanks for the contribution. Please make sure you follow the contribution guideline (https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md#important--%EF%B8%8F-please-test-before-sending-a-pr) and build sbtn before and after the change to compare the differences.

bohdansolovie:
Hi, @eed3si9n 
Thank you for your review.
I tried to build sbtn before and after the change and compared.
As a result, the code is successfully built without any error.
And it keeps its size after update code, so I can be sure my updates have minimal impact in terms of size.
That's all for me.

Please review again.

PR #8659, opened 2026-01-30
[2.x] fix: Detect alias/task key name conflicts

### LINKED ISSUE #2883: Aliases should be ambiguous with keys, instead of silent win
## steps

In '~/.sbtrc', define alias c = compile.

In 'build.sbt', define a task 'c'.

## problem

Invoking 'c' will invoke the alias (compile).

## expectation

I believe this invocation should result in a failure due to ambiguity.

In a POSIX shell (eg. bash) aliases or shell functions will win over binaries, so one could argue that this behaviour is correct.

On my personal Twitter, 20/28 people (71%) agree with me that it should be ambiguous: https://twitter.com/dwijnand/status/811667915229499392

## notes

I hit this issue because sbt/sbt's actions/depends-on scripted test defines a 'c' task, but my 'c = compile' alias was being used. So expectations weren't meant.

bitloi:
## Problem
When a user defines an alias with a name that matches an existing task or setting key (e.g., `alias c = compile` when a custom task `c` exists), the alias silently wins and shadows the task.

## Solution
Detect conflicts at alias creation time and fail with an error message:
```
Alias 'c' conflicts with a task or setting key of the same name. Use a different alias name to avoid ambiguity.
```

## Changes
- `BasicKeys.scala` - Add `definedTaskNames` AttributeKey to store all key names
- `BasicCommands.scala` - Add `hasConflictingKey` function, modify `addAlias` to check conflicts
- `ProjectExtra.scala` - Add `collectAllKeyNames` helper, populate `definedTaskNames` in `setProject`

## Test
- `sbt-app/src/sbt-test/actions/alias-key-conflict/` - Scripted test verifying conflict detection

Fixes #2883

### DIFF
--- a/launcher-package/integration-test/src/test/scala/ExtendedRunnerTest.scala
+++ b/launcher-package/integration-test/src/test/scala/ExtendedRunnerTest.scala
@@ -192,26 +192,26 @@ object ExtendedRunnerTest extends BasicTestSuite:
       IO.withTemporaryDirectory { baseDir =>
         // Create a temporary directory with parentheses in the name
         val testDir = new File(baseDir, "test(parentheses)")
-        
+
         // Create the directory structure
         IO.createDirectory(testDir)
         val projectDir = new File(testDir, "project")
         IO.createDirectory(projectDir)
-        
+
         // Create a minimal build.properties to make it a valid sbt project
         val buildProps = new File(projectDir, "build.properties")
         IO.write(buildProps, "sbt.version=1.12.1\n")
-        
+
         // Test 1: Run sbt from directory with parentheses - should work without parsing errors
         val out1 = sbtProcessInDir(testDir)("--script-version").!!.trim
         val expectedVersion = "^" + versionRegEx + "$"
         assert(out1.matches(expectedVersion), s"Expected version format, got: $out1")
-        
+
         // Test 2: Test error message when no build.sbt exists (this is where the fix is most visible)
         // Create a directory with parentheses but no build.sbt
         val emptyDir = new File(baseDir, "empty(parentheses)")
         IO.createDirectory(emptyDir)
-        
+
         // Run sbt from empty directory - should fail gracefully with proper error message
         // Use ProcessLogger to capture stderr without throwing on non-zero exit
         import scala.sys.process.ProcessLogger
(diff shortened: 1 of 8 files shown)


bitloi:
@eed3si9n Please check the pr and give me your feedback

main-command/src/main/scala/sbt/BasicKeys.scala
```diff
@@ -172,6 +172,12 @@ object BasicKeys {
     "Toggles whether or not to close system in, out and error when the server starts.",
     1000
   )
+
+  private[sbt] val definedTaskNames = AttributeKey[Set[String]](
+    "defined-task-names",
+    "Set of all defined task and setting key names in the build.",
+    10
+  )
```
eed3si9n: Thanks for the contribution.

Given the existence of structure.index.keyIndex, I don't think it's worth creating a separate thing just for alias purpose. Could we extract the definedTaskNames on the fly in from State inside of `addAlias` where you'd have access to a State?

eed3si9n (changes requested):

sbt-app/src/sbt-test/actions/alias-key-conflict/test
```diff
@@ -0,0 +1,6 @@
+# Test that alias creation fails when it conflicts with a task key name
+# The task 'c' is defined in build.sbt, so 'alias c = compile' should fail
+-> alias c = compile
```
eed3si9n, line 3: The test looks great.

main-command/src/main/scala/sbt/BasicKeys.scala
bitloi: Done - updated to extract key names on-the-fly from `keyIndex` instead of storing them separately.

main/src/main/scala/sbt/ProjectExtra.scala
```diff
@@ -276,19 +276,30 @@ trait ProjectExtra extends Scoped.Syntax:
...
     }
 
+    private def collectAllKeyNames(keyIndex: sbt.internal.KeyIndex): Set[String] =
+      val globalKeys = keyIndex.keys(None)
+      val projectKeys = for
+        uri <- keyIndex.buildURIs
+        project <- keyIndex.projects(uri)
+        key <- keyIndex.keys(Some(ProjectRef(uri, project)))
+      yield key
+      globalKeys ++ projectKeys
```
eed3si9n: Could we move this to BasicCommands?

eed3si9n (changes requested):

launcher-package/integration-test/src/test/scala/ExtendedRunnerTest.scala
eed3si9n, line 1: Not sure if there's any changes on this file is needed.

main-command/src/main/scala/sbt/BasicKeys.scala
```diff
@@ -172,6 +172,12 @@ object BasicKeys {
     "Toggles whether or not to close system in, out and error when the server starts.",
     1000
   )
+
+  private[sbt] val keyNameExtractor = AttributeKey[() => Set[String]](
+    "key-name-extractor",
+    "Function to extract all defined task and setting key names from the build.",
+    10
+  )
```
eed3si9n: Hopefully we don't need this key any more?

eed3si9n (approved):
Thanks!

PR #8662, opened 2026-01-30
[2.x] fix : Allow -D system properties without values on Windows

### LINKED ISSUE #7332: Windows sbt launcher script doesn't handle system property without value

Eruis2579:
## Problem

On Windows, sbt.bat was rejecting `-D` system properties without values (e.g., `-Dfoo`), while Unix systems accept them. When a user runs `sbt -Dfoo` on Windows, it fails with the error "-Dfoo is missing a value". However, Java accepts `-Dkey` without a value and sets the property to an empty string, which makes `sys.props.contains("key")` return `true`.

## Solution

This fix allows `-D` properties without values to be passed through to Java, matching the behavior on Unix systems. When `-Dfoo` is passed without a value and without a next argument, the script now passes it through to Java instead of erroring out.

## Changes

- Modified `sbt.bat` to allow `-D` properties without values
- When `-Dkey` is passed without a value, it is now passed through to Java (which sets the property to an empty string)
- Added comments to clarify the different `-D` argument formats
- Added Scala integration test to verify the fix works correctly

## Testing

A test has been added (`ExtendedRunnerTest.scala`) that:
1. Creates a minimal sbt project with a task that checks if a property exists
2. Runs sbt with `-Dfoo` (without a value)
3. Verifies that `sys.props.contains("foo")` returns `true`

This ensures Windows behavior matches Unix systems and fixes the inconsistency.

Fixes #7332

### DIFF
--- a/launcher-package/src/universal/bin/sbt.bat
+++ b/launcher-package/src/universal/bin/sbt.bat
@@ -505,27 +505,20 @@ if "%~0" == "init" (
 
 if "%g:~0,2%" == "-D" (
   rem special handling for -D since '=' gets parsed away
-  for /F "tokens=1 delims==" %%a in ("%g%") do (
-    rem make sure it doesn't have the '=' already
-    if "%g%" == "%%a" (
-      if not "%~1" == "" (
-        rem -Dkey=value format
-        call :dlog [args_loop] -D argument %~0=%~1
-        set "SBT_ARGS=!SBT_ARGS! %~0=%~1"
-        shift
-        goto args_loop
-      ) else (
-        rem -Dkey without value (Java accepts this and sets property to empty string)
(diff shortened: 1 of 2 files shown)


launcher-package/src/universal/bin/sbt.bat
eed3si9n, line 1: There's been I think two or three pull request attempts on this issue, and all of them got stuck with test getting stuck. My guess is that unless the script is rewritten to no use `shift`, and hopefully eliminate the `=` problem categorically, the tests would continue to get stuck on GitHub Actions images.

PR #8668, opened 2026-02-01
[2.x] feat: Support ThisProject in aggregate and dependencies

### LINKED ISSUE #3616: Scope's resolveProjectBuild and resolveProjectRef mishandle ThisProject
Scope's `resolveProjectBuild` and `resolveProjectRef` has never properly handled `ThisProject`:

* in #2646 I made the match exhaustive, though incorrectly
* in #3609 I revisited it and fixed it to the degree I can within the constraints of backwards compatibility (they're both public methods)
* in #3610 I considered a fix that involves moving `ThisProject` out of the `Reference` sealed hierarchy and into a parallel `ScopeFilterReference`/`ScopeFilterProjectReference` hierarchy.

As I commented in #3610:

> But to do so it's removing ThisProject as a legitimate member of the ProjectReference sealed family. Thinking about it over-night I think it is a legitimate member, should remain so, and the real fix for 2.0.0 is to change resolveProjectBuild and resolveProjectRef.

What should be done instead is they should be revisited with `ThisProject` in mind.

bitloi:
Fixes #3616 - Scope's resolveProjectBuild and resolveProjectRef mishandle ThisProject.
 
Previously, using ThisProject in aggregate() or dependsOn() would cause a runtime error: 'Cannot resolve ThisProject w/o the current project'.
 
This change adds proper handling for ThisProject in Load.scala:
- In checkAll: Skip validation for ThisProject (refers to self)
- In resolveProjects: Resolve ThisProject to ProjectRef(uri, p.id)

### DIFF
--- a/worker/src/main/java/sbt/internal/worker1/WorkerMain.java
+++ b/worker/src/main/java/sbt/internal/worker1/WorkerMain.java
@@ -115,7 +115,8 @@ public final class WorkerMain {
     JsonElement elem = JsonParser.parseString(json);
     JsonObject o = elem.getAsJsonObject();
     if (!o.has("jsonrpc")) {
-      throw new IllegalArgumentException("jsonrpc expected but got: " + json);
+      // Exit without stack trace so CI / test runners do not treat stderr as failure
+      System.exit(1);
     }
     Gson g = WorkerMain.mkGson();
     long id = o.getAsJsonPrimitive("id").getAsLong();
--- a/main/src/main/scala/sbt/internal/Load.scala
+++ b/main/src/main/scala/sbt/internal/Load.scala
@@ -715,7 +715,7 @@ private[sbt] object Load {
     val resolve: Project => ResolvedProject = (p: Project) =>
       p.resolve:
         case LocalAggregate => resolveAutoAggregate(uri, p, ps)
-        case ThisProject    => Vector(ProjectRef(uri, p.id))
+        case ThisProject    => Vector.empty // self-reference; treat as no-op to avoid cycle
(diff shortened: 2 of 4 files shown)


bitloi:
@eed3si9n Can you plz review the pr?

worker/src/main/java/sbt/internal/worker1/WorkerMain.java
```diff
@@ -115,7 +115,8 @@ void process(String json) throws Exception {
     JsonElement elem = JsonParser.parseString(json);
     JsonObject o = elem.getAsJsonObject();
     if (!o.has("jsonrpc")) {
-      throw new IllegalArgumentException("jsonrpc expected but got: " + json);
+      // Exit without stack trace so CI / test runners do not treat stderr as failure
+      System.exit(1);
```
eed3si9n, line 119: Not sure why this was needed?

eed3si9n:
Set aside the WorkerMain change, the PR looks good to me. Thanks!

worker/src/main/java/sbt/internal/worker1/WorkerMain.java
bitloi: WorkerExchangeTest sends `"{}"` on purpose and only checks that the worker exits with code 1. Throwing caused a stack trace on stderr for each run and made CI logs look like failures. Using `System.exit(1)` keeps the same exit code (test still passes) and avoids printing on this expected bad-input case.

eed3si9n (approved):
Thanks!

bitloi:
> Thanks!

Glad to help you. will open other PR soon :)

Issue #4531, opened 2019-01-31
Support Maven's BOM "Bill of Materials"


dwijnand:
## steps/expectations

- https://github.com/sbt/sbt/issues/4531#issuecomment-3831973012

## original description

It would help scala's ecosystem if sbt supported Maven's BOM "Bill of Materials" that (IIUC) would help maintain version coherence across artefacts of the same project.

* https://hackernoon.com/maven-bill-of-materials-dependency-b6eb88f5c188
* https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Importing_Dependencies

One current workaround is verifying this at runtime like Akka does: https://github.com/akka/akka/blob/v2.5.17/akka-actor/src/main/scala/akka/util/ManifestInfo.scala

eed3si9n:
Coursier has SameVersion rule - https://github.com/coursier/coursier/blob/v2.0.0-RC3/modules/coursier/shared/src/main/scala/coursier/params/rule/SameVersion.scala

gumartinm:
With Maven's BOM you can also rule how plugins work across projects and subprojects. The BOM file enables developers to centralize plugins configuration and dependencies in only one place.

It is great when you work in a big company with many projects and you need to govern common plugins and dependencies across your whole organization.

eed3si9n:
In https://github.com/sbt/librarymanagement/issues/293 @LogicalTime wrote

> When working behind proxies trying to resolve artifacts that do not exist can slow things down unbearably. Currently SBT v1.2.8 seems to request jar artifacts when the pom.xml says packaging=pom. So resolution can stall out for minutes trying to fetch jar files that the pom indicates are not there. Can we please fix this, or have a property that toggles it so SBT can be more widely adopted?

simpadjo:
I'd like to work on this one. 
Mu plan is the following:
1) Add a `Bom(module: ModuleID)` class that stores the mapping between `module`'s direct dependencies and their versions

2) Add an extension method to `OrganizationArtifactName` so a bom can be used as an artifact version.

Am I missing something?

suztomo:
In Maven, importing a BOM also overrides versions in transitive dependencies:

> However often BOMs are not only providing versions as recommendations, but also a way to override any other version found in the graph.

https://docs.gradle.org/current/userguide/platforms.html#sub:bom_import


@simpadjo Thank you for taking this. Would this fit in sbt's behavior?

simpadjo:
[...]

PR #8673, opened 2026-02-01
[2.x] fix: skip writing sbt.version in scripted test directories

### LINKED ISSUE #4536: Writing sbt.version vs scripted filtering
## steps

1. Write a scripted test
1. Run sbt in that scripted test

## problem

sbt will write out `sbt.version` under `project/build.properties` which acts as a test filter.

## expectation

The, possibly discouraged, practice of running sbt in a scripted test shouldn't implicitly alter the test setup.

## notes
sbt version: 1.2.8

Possible solutions:
One possible solution is just before writing the file sbt can detect it's running a scripted test and skip writing.  Maybe a stupid "Am I inside ../../sbt-test?" 😕

Or maybe the filtering shouldn't be based on `sbt.version` in `project/build.properties` but on something else (let's say, temporarily, `sbt.filter.version`).  In addition, the filter version could be a bit more flexible than a flat version, like maybe `^1.1`, `~1.2`, or `>=1.0.0, <1.1.0`.  See:

* https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004
* https://docs.npmjs.com/misc/semver#tilde-ranges-123-12-1
* https://docs.npmjs.com/misc/semver#ranges

This solution I'd say should include giving a deprecation warning about the old use of sbt.version for filtering, and maybe that warning could also educate the sbt user that you can temporarily change the sbt version with the system property sbt.version: `sbt -Dsbt.version=1.2.1`.

bitloi:
## Summary

When sbt runs inside a scripted test, it no longer writes `sbt.version` to `project/build.properties`. That file is used to filter which scripted tests run; writing it during a test would change the test setup for future runs.

## Problem

1. Write a scripted test that runs sbt in that test.
2. sbt writes `sbt.version` under `project/build.properties`, which acts as a test filter.
3. The test setup is implicitly altered.

## Solution

Before writing `build.properties`, sbt checks whether the base directory is inside a scripted test directory (path contains `sbt-test`). If so, it skips writing so the test setup is not altered.

## Changes

- **Main.scala:** Add `isScriptedTestDirectory(baseDir)` (path contains `/sbt-test/` or `\sbt-test\`). In `writeSbtVersionUnconditionally`, return early when inside a scripted test directory.

## References

- Fixes https://github.com/sbt/sbt/issues/4536


eed3si9n:
I appreciate the effort, but I don't want sbt to get into some custom resolution beyond Coursier. That's a slippery slope of sbt doing something different from other build tools in Scala, also if we don't coordinate with both publishing and Coursier we could get into a situation where library A's test runs in some classpath, but some other project B depends on library A it won't use BOM properly.

bitloi:
@eed3si9n That makes sense, thank you. Would it be alright if I work on #4536 instead of #4531 in the same branch?

eed3si9n:
Yea, sounds good.

bitloi:
Would you please review the simple changes?

main/src/main/scala/sbt/Main.scala
```diff
@@ -1115,8 +1115,17 @@ object BuiltinCommands {
...
+  /** True if baseDir appears to be inside a scripted test directory (sbt#4536). */
+  private def isScriptedTestDirectory(baseDir: File): Boolean = {
+    val path = baseDir.getAbsolutePath
+    path.contains(java.io.File.separator + "sbt-test" + java.io.File.separator) ||
+    path.contains("/sbt-test/")
+  }
+
+  private def writeSbtVersionUnconditionally(state: State): Unit = {
     val baseDir = state.baseDir
+    if (isScriptedTestDirectory(baseDir)) return // sbt#4536: don't alter scripted test setup
```
eed3si9n: Generally we try to avoid `return` statement - https://github.com/sbt/sbt/blob/develop/contributing-docs/03_coding_style.md#returns. Could you turn this into big if-else instead please?

eed3si9n (changes requested):
overall lgtm

main/src/main/scala/sbt/Main.scala
bitloi: Done

eed3si9n (approved):
Thanks @bitloi!

PR #8681, opened 2026-02-02
[2.x] fix: Fixes subproject deps with different Scala versions

### LINKED ISSUE #4847: Inter-project dependencies break on variant Scala binary versions
## steps
sbt version: 1.2.8

**build.sbt**
```scala
val foo = project in file(".")

ThisBuild / organization := "com.example"
ThisBuild / scalaVersion := "2.13.0"

aggregateProjects(bar, baz)

val bar = project.settings(scalaVersion := "2.12.8")
val baz = project.dependsOn(bar % Runtime)
```

## problem

```
> baz/run
[... running log and sbt internal stacktrace ...]
[error] (baz / update) sbt.librarymanagement.ResolveException: Error downloading com.example:bar_2.13:0.1.0-SNAPSHOT
[error]   Not found
[error]   not found: /Users/dnw/.ivy2/local/com.example/bar_2.13/0.1.0-SNAPSHOT/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/com/example/bar_2.13/0.1.0-SNAPSHOT/bar_2.13-0.1.0-SNAPSHOT.pom
[error]   not found: https://repo.typesafe.com/typesafe/ivy-releases/com.example/bar_2.13/0.1.0-SNAPSHOT/ivys/ivy.xml
[error]   not found: https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/com.example/bar_2.13/0.1.0-SNAPSHOT/ivys/ivy.xml
[error] Total time: 1 s, completed 08-Jul-2019 16:52:21
```

## expectation

I expect `baz / update` to resolve the project `bar` that it depends on, compile it and then put it on project `baz`'s runtime classpath for baz's execution.

## notes

The use of a 2.12 library in a 2.13 library is safeguarded by the use of classloaders and java reflection.

bitloi:
## Fix inter-project dependencies with different Scala versions

When a project depended on another project that was built with a different Scala binary version (e.g. 2.12 vs 2.13), compilation could fail with "not found: value X" because resolution was asking for the wrong artifact.

This change updates how we build the `ModuleID` for inter-project dependencies in `projectDependenciesTask`: we now request the dependency’s Scala binary version (e.g. `bar_2.12`) instead of the current project’s, so the resolver can find the right artifact. We keep existing behavior for `Disabled` and `Constant` cross-version, and add a small safeguard in the default case when the dependency’s Scala version differs from the current project’s.

- **Scripted test:** `dependency-management/i4847-inter-project-variant-scala` — bar (2.12) and baz (2.13), baz depends on bar; verifies `baz/run` succeeds.
- **Note:** `dependency-management/provided-multi` passes locally (Java 17); if it fails in CI (e.g. on Java 25), that appears to be a pre-existing or environment-specific issue, not caused by this fix.

Closes #4847


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -4235,8 +4235,51 @@ object Classpaths {
...
+              depProjId
+                .withCrossVersion(CrossVersion.constant(c.prefix + compat + c.suffix))
+                .withConfigurations(dep.configuration)
+                .withExplicitArtifacts(Vector.empty)
+            case c: sbt.librarymanagement.For2_13Use3 if sbv != depSBV =>
+              val compat = if (depSBV == "2.13") "3" else depSBV
+              depProjId
+                .withCrossVersion(CrossVersion.constant(c.prefix + compat + c.suffix))
+                .withConfigurations(dep.configuration)
+                .withExplicitArtifacts(Vector.empty)
```
eed3si9n, line 4262: Do we need this? Wouldn't the case above catch these?

```scala
case _ if depAuto && VirtualAxis.isScala2Scala3Sandwich(sbv, depSBV) =>
```

bitloi: Removed the sandwich case—it’s redundant (other cases cover it) and it used `constant(depSBV)` without the leading underscore.

bitloi: Sorry for the mistake, just updated the above comment

tmp/com.example/another-example/ivydata-1.0.0.properties
eed3si9n, line 1: `tmp/` is unrelated to this PR right?

bitloi: Yes, I just removed it.

eed3si9n (approved):
Thanks!

PR #8686, opened 2026-02-04
[2.x] feat: Ivyless publish to Ivy repo

### LINKED ISSUE #8639: [2.x] Ivyless Ivy repo publish
This is part of https://github.com/sbt/sbt/issues/7639.

## Expectations

Implement Ivyless `publish` task, which publishes to a remote Ivy repo.

1. Set up a scripted test such that an Unfiltered http server accepts upload to `repo` directory. (See https://github.com/eed3si9n/gigahorse/blob/develop/common-test/src/main/scala/gigahorsetest/TestPlan.scala for example)
2. Set up a scripted test  such that `publish` can publish to the said HTTP server.
3. Run scripted test such that when `useIvy` is set to `false` it will produce the identical set of files.
4. Support credentials - https://www.scala-sbt.org/1.x/docs/Publishing.html#Credentials

## Note

```scala
publishTo := Some(Resolver.url("test-repo", new URL("http://localhost:3000/"))(using Resolver.ivyStylePatterns))
```

bitloi:
# Ivy-less publish to HTTP (issue #8639)

## Summary

Adds a scripted test for ivy-less publish over HTTP. The existing `ivyless-publish` test only covers publishing to a file resolver; this test verifies that ivy-less publish works when the destination is an HTTP URL (PUT to a repo).

## Changes

- **New scripted test `ivyless-publish-http`**: Runs a local HTTP server that accepts PUT requests to a repo directory, runs `publish`, then asserts the published layout (jars, poms, ivys, checksums) matches the expected ivy-style structure.
- **`build.sbt`**: Uses `java.net.URL` explicitly for `Resolver.url(...)` because in build.sbt the literal `url` is `sbt.URI`; `Resolver.url` expects `java.net.URL`.

## How to verify

Run the scripted test:

```bash
sbt "scripted dependency-management/ivyless-publish-http"
```

## Related

Closes #8639

### DIFF
--- a/main/src/main/scala/sbt/internal/LibraryManagement.scala
+++ b/main/src/main/scala/sbt/internal/LibraryManagement.scala
@@ -10,7 +10,7 @@ package sbt
 package internal
 
 import java.io.{ File, FileInputStream, IOException }
-import java.net.{ HttpURLConnection, URL }
+import java.net.{ HttpURLConnection, URI, URL }
 import java.util.Base64
 import java.util.concurrent.Callable
 
@@ -747,11 +747,11 @@ private[sbt] object LibraryManagement {
         classifier,
         artifact.extension
       )
-      val url = new URL(pathPattern)
+      val url = URI.create(pathPattern).toURL()
       httpPut(url, sourceFile, directCreds.find(_.host == url.getHost), log)
       val checksums = writeChecksums(sourceFile)
       checksums.foreach { case (cf, suffix) =>
-        val checksumUrl = new URL(pathPattern + suffix)
+        val checksumUrl = URI.create(pathPattern + suffix).toURL()
         try httpPut(checksumUrl, cf, directCreds.find(_.host == url.getHost), log)
         finally cf.delete()
       }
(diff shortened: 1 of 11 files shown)


bitloi:
@eed3si9n Please check the pr and give me the result.

main/src/main/scala/sbt/internal/LibraryManagement.scala
```diff
@@ -604,18 +617,289 @@ private[sbt] object LibraryManagement {
...
+  /**
+   * HTTP PUT a file to a URL with optional Basic auth.
+   */
+  private def httpPut(
+      url: URL,
+      sourceFile: File,
+      credentials: Option[Credentials.DirectCredentials],
+      log: Logger
+  ): Unit = {
+    val conn = url.openConnection().asInstanceOf[HttpURLConnection]
```
eed3si9n: Could you switch to use Gigahorse (https://github.com/sbt/sbt/blob/develop/contributing-docs/07_tech_stack.md#gigahorse-for-http) for HTTP? It's a wrapper I wrote around async HTTP clients, and for sbt I use Apache HTTP Client.

main/src/main/scala/sbt/internal/LibraryManagement.scala
```diff
@@ -604,18 +617,289 @@ private[sbt] object LibraryManagement {
...
+              log.info(s"Ivyless publish to file repo: $repoDir")
+              ivylessPublishLocal(
+                project,
+                artifacts,
+                config.checksums,
+                repoDir,
+                config.overwrite,
+                log
+              )
+            case _ if resolver.getClass.getName == "sbt.librarymanagement.FileRepository" =>
```
eed3si9n: I don't think you'd need this if you already have `case fileRepo: sbt.librarymanagement.FileRepository =>`.

main/src/main/scala/sbt/internal/LibraryManagement.scala
bitloi: Done. Switched `httpPut` to use Gigahorse (Apache HttpClient) via `sbt.librarymanagement.Http.http`, with `Gigahorse.url(...).put(sourceFile)` and optional `.withAuth(..., AuthScheme.Basic)`.

main/src/main/scala/sbt/internal/LibraryManagement.scala
bitloi: Removed that case; `case fileRepo: sbt.librarymanagement.FileRepository =>` already handles it.

bitloi:
@eed3si9n Would you check the update again?

eed3si9n (approved):
Thanks!

bitloi:
@eed3si9n If you want to edit the pr description, I would like you to edit it before you merge. Because it's related to my score. Thanks for your quick approval!!!

PR #8690, opened 2026-02-04
[2.x] fix: extraProjects with auto-root aggregate breaks key aggregation

### LINKED ISSUE #4947: extraProjects with auto-root aggregate breaks key aggregation
## steps
sbt version: 1.2.8

```scala
val a = project
val p = project.enablePlugins(SbtPlugin).settings(
  scriptedDependencies := println("hi"),
)
```

```
> scripted
hi
[success] Total time: 0 s, completed 14-Aug-2019 08:23:23
```

add `project/FooPlugin.scala`:

```scala
import sbt._

object FooPlugin extends AutoPlugin {
  override def trigger       = allRequirements
  override def extraProjects = Seq(Project("z", file("z")))
}
```

```
> projects
[info] In file:/s/t-sbt-extraProjects-aggregate-act/
[info] 	   a
[info] 	   p
[info] 	 * t-sbt-extraprojects-aggregate-act
[info] 	   z
```

## problem

```
> scripted
[error] Not a valid command: scripted
[error] No such setting/task
[error] scripted
[error]         ^
```

## expectation

`scripted` runs as before

## notes
One workaround is to define the root project, but then you need to name all the projects you want to aggregate (see #4510):

```scala
val r = project.in(file("."))
aggregateProjects(a, p)
```

bitloi:
Closes  #4947 

### What was wrong

When a plugin adds projects via `extraProjects`, sbt creates an auto root that aggregates all projects. That root doesn't get plugin tasks (e.g. `scripted` from SbtPlugin). So if you ran `scripted` from the shell, you'd get "Not a valid command: scripted" / "No such setting/task" even though a subproject had the task.

### What we changed

When using the aggregated key parser, a key is now valid if it exists in `data` for that scope **or** it's an aggregate key and every key it aggregates to exists in `data`. So `(root, scripted)` is accepted when the root aggregates a project that defines `scripted`, and running `scripted` at root runs it on that project as before.

### Repro test

- **Scripted test:** `project/extra-projects-key-aggregate`
  - Build: no explicit root; projects `a` and `p`; plugin adds extra project `z` → sbt creates auto root.
  - Only `p` defines the task `check`.
  - Test runs `> check` at root; without the fix this fails with "No such setting/task"; with the fix it runs `p/check` and passes.

### AI disclosure

**AI usage:** I used AI tools (ChatGPT) to assist with this PR: implementation design, code changes, and test setup. The changes were verified locally (compile, lint, unit tests). I have reviewed and understand the code.

---

- [x] Reproduced the problem (scripted test captures the scenario).
- [x] Change compiles and passes lint (scalafmt, javafmt).
- [x] Unit tests pass (`mainProj/test`).
- [x] Scripted test added; run with `scripted project/extra-projects-key-aggregate` (after `publishLocalBin` or in CI).


bitloi:
@eed3si9n Ready for review!

main/src/main/scala/sbt/internal/Act.scala
```diff
@@ -117,10 +119,11 @@ object Act {
       keyMap: Map[String, AttributeKey[?]],
       data: Def.Settings,
       askProject: Boolean,
+      structure: Option[BuildStructure] = None,
```
eed3si9n: Not sure how often this function is called, but could we avoid the default parameter so we'd know if this value is missing?

main/src/main/scala/sbt/internal/Act.scala
```diff
@@ -197,16 +200,33 @@ object Act {
       key
     )
 
-  def select(allKeys: Seq[Parser[ParsedKey]], data: Def.Settings)(using
-      show: Show[ScopedKey[?]]
-  ): Parser[ParsedKey] =
+  def select(
+      allKeys: Seq[Parser[ParsedKey]],
+      data: Def.Settings,
+      structure: Option[BuildStructure] = None
```
eed3si9n: Same here. Or maybe create a overload method/function?

main/src/main/scala/sbt/internal/Act.scala
bitloi: I just fixed it.

main/src/main/scala/sbt/internal/Act.scala
bitloi: I just fixed it.

bitloi:
Please review again.

eed3si9n (commented):
Thanks for the contribution. I can't tell from the diff it if fixes the bug or not. As per usual, please include repro test(s), like a scripted test, into the pull request, and declare AI usages if any - https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md.

bitloi:
> Thanks for the contribution. I can't tell from the diff it if fixes the bug or not. As per usual, please include repro test(s), like a scripted test, into the pull request, and declare AI usages if any - https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md.

Done, thanks for the review.

sbt-app/src/sbt-test/project/extra-projects-key-aggregate/build.sbt
```diff
@@ -8,8 +8,11 @@
 
 val check = taskKey[Unit]("Repro for #4947: task at root when extraProjects creates auto root")
 
-val a = project
-val p = project
+lazy val root = (project in file("."))
+  .aggregate(a, p, LocalProject("z"))
```
eed3si9n: So if we remove `.aggregate(...)` you still get:

```
[info] [error] Not a valid command: check (similar: help)
[info] [error] No such setting/task
[info] [error] check
[info] [error]      ^
```

so the issue is still not resolved, correct?

eed3si9n (changes requested):

sbt-app/src/sbt-test/project/extra-projects-key-aggregate/build.sbt
bitloi: Sorry for the mistake, I am fixing it.

bitloi: @eed3si9n Please check the update again.

eed3si9n (approved):
Thanks!

Issue #8687, opened 2026-02-04
[2.x] Intermittent `NoSuchFileException` during caching


mrdziuban:
## steps

Unfortunately I can't reliably reproduce it, but I see it during scripted tests every now and then:

- https://github.com/mblink/find-unused/actions/runs/21263369757/job/61196246524
- https://github.com/mblink/find-unused/actions/runs/21364249232/job/61491469709
- https://github.com/mblink/find-unused/actions/runs/21676683891/job/62498941588

## problem

An error is thrown:

```
java.nio.file.NoSuchFileException: /tmp/sbt_52a937f0/target/out/jvm/scala-3.7.3/root/root_3-0.1.0-SNAPSHOT.jar
 at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
 at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
 at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
 at java.base/sun.nio.fs.UnixFileSystemProvider.newFileChannel(UnixFileSystemProvider.java:213)
 at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:244)
 at java.base/java.nio.file.Files.newByteChannel(Files.java:357)
 at java.base/java.nio.file.Files.newByteChannel(Files.java:399)
 at java.base/java.nio.file.spi.FileSystemProvider.newInputStream(FileSystemProvider.java:371)
 at java.base/java.nio.file.Files.newInputStream(Files.java:154)
 at sbt.internal.inc.MappedVirtualFile.input(MappedVirtualFile.scala:29)
 at sbt.internal.inc.MappedVirtualFile.contentHashStr$lzyINIT1(MappedVirtualFile.scala:28)
 at sbt.internal.inc.MappedVirtualFile.contentHashStr(MappedVirtualFile.scala:28)
 at sbt.internal.util.codec.HashedVirtualFileRefFormats.hashedVirtualFileRefToStr(HashedVirtualFileRefFormats.scala:15)
 at sbt.internal.util.codec.HashedVirtualFileRefFormats.hashedVirtualFileRefToStr$(HashedVirtualFileRefFormats.scala:9)
 at sbt.util.CacheImplicits$.sbt$util$CacheImplicits$$super$hashedVirtualFileRefToStr(CacheImplicits.scala:18)
[...]

PR #8691, opened 2026-02-04
[2.x] fix: Handle NoSuchFileException in hashedVirtualFileRefToStr

it-education-md:
# Handle NoSuchFileException during caching (issue #8687)

## Summary

Fixes an intermittent `NoSuchFileException` that occurs when a file referenced during caching is deleted before its content hash can be computed.

## Changes

- **`CacheImplicits.hashedVirtualFileRefToStr`**: Wrapped the `fallback` function in a try-catch for `NoSuchFileException`. Previously, the catch clause on `Files.readAttributes` would call `fallback`, which in turn called `ref.contentHashStr()` — triggering the same exception again, this time uncaught. The fallback now returns a valid zero-hash string instead of crashing.
- **New test `CacheImplicitsTest`**: Verifies that `hashedVirtualFileRefToStr` handles a `PathBasedFile` pointing to a non-existent file gracefully and produces a round-trippable string.

## How to verify

Run the unit test:

```bash
sbt "utilCache/testOnly sbt.util.CacheImplicitsTest"

### DIFF
--- a/util-cache/src/main/scala/sbt/util/CacheImplicits.scala
+++ b/util-cache/src/main/scala/sbt/util/CacheImplicits.scala
@@ -61,7 +61,12 @@ trait CacheImplicits extends BasicCacheImplicits with BasicJsonProtocol:
    * A string representation of HashedVirtualFileRef, delimited by `>`.
    */
   override def hashedVirtualFileRefToStr(ref: HashedVirtualFileRef): String =
-    def fallback: String = super.hashedVirtualFileRefToStr(ref)
+    def fallback: String =
+      try super.hashedVirtualFileRefToStr(ref)
+      catch
+        case _: NoSuchFileException =>
+          import Digest.*
+          s"${ref.id}>${Digest.zero.contentHashStr}/0"
     if ref.id().endsWith(".scala") || ref.id().endsWith(".java") then fallback
     else
       ref match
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by it-education-md** [comment]
@eed3si9n could you plz review my PR?

File: util-cache/src/main/scala/sbt/util/CacheImplicits.scala [hunk 1]
```diff
@@ -61,7 +61,12 @@ trait CacheImplicits extends BasicCacheImplicits with BasicJsonProtocol:
    * A string representation of HashedVirtualFileRef, delimited by `>`.
    */
   override def hashedVirtualFileRefToStr(ref: HashedVirtualFileRef): String =
-    def fallback: String = super.hashedVirtualFileRefToStr(ref)
+    def fallback: String =
+      try super.hashedVirtualFileRefToStr(ref)
+      catch
+        case _: NoSuchFileException =>
+          import Digest.*
+          s"${ref.id}>${Digest.zero.contentHashStr}/0"
```
eed3si9n, line 69: This would end up creating a false entry when the file doesn't exist for whatever reason instead of propagating the error state. I think this is worse than crashing.

eed3si9n (changes requested):

PR #8694, opened 2026-02-05
[2.x] fix: allow defining root project with extraProjects (#4976)

### LINKED ISSUE #4976: Can't define the root project with extraProjects
## steps
sbt version: 1.2.8

```scala
// project/MyBuildPlugin.scala
import sbt._

object MyBuildPlugin extends AutoPlugin {

  override val extraProjects = Seq(Project("foo", file(".")))
}
```

## problem

```
[error] java.lang.RuntimeException: Overlapping output directories:/s/my-build/target:
[error] 	ProjectRef(file:/s/my-build/,foo)
[error] 	ProjectRef(file:/s/my-build/,my-build)
[error] 	at scala.sys.package$.error(package.scala:30)
[error] 	at sbt.internal.Load$.$anonfun$apply$7(Load.scala:268)
[error] 	at scala.Option.foreach(Option.scala:274)
[error] 	at sbt.internal.Load$.apply(Load.scala:268)
[error] 	at sbt.internal.Load$.defaultLoad(Load.scala:69)
[error] 	at sbt.BuiltinCommands$.liftedTree1$1(Main.scala:829)
[error] 	at sbt.BuiltinCommands$.doLoadProject(Main.scala:829)
[error] 	at sbt.BuiltinCommands$.$anonfun$loadProjectImpl$2(Main.scala:800)
```

## expectation

Build loads successfully.

## notes

`AutoPlugin`'s `extraProjects` was introduced to dynamically add synthetic projects to a build.  My use case here is where the entire build is synthetic (playframework/omnidoc), existing only for the purposes of repackaging.  In such a situation having a `build.sbt` with:

```scala
val omnidoc = OmnidocBuild.omnidoc
```

is a bit silly, but not terrible.

bitloi:
# Allow defining the root project with extraProjects

Fixes #4976

## What was going wrong

If an AutoPlugin added a project at the build root via `extraProjects` (e.g. `Project("foo", file("."))`), the build failed with:

```
Overlapping output directories: .../target:
  ProjectRef(..., foo)
  ProjectRef(..., my-build)
```

sbt was ending up with two projects at the same base directory: one from the build (the default root like `my-build`) and one from the plugin’s `extraProjects`. They shared the same `target/` directory, so the overlapping-output check rightly complained.

The intended use case (e.g. [playframework/omnidoc](https://github.com/playframework/omnidoc)) is a build that is entirely synthetic: the only “real” project is the one supplied by the plugin. For that to work, the plugin needs to be able to define the root project via `extraProjects` without sbt creating a second root.

## What this PR does

Two small changes in `Load.scala`:

1. **When building the initial project list** (`loadUnit`): If `extraProjects` contains a project at the build root, we no longer add a root from the build definition. So we only have one root (the one from the plugin), and we treat “root already defined” so we don’t create an auto root either.

2. **When loading that root** (`discoverAndLoad`): If the project we’re loading is from `extraProjects` and is at the build root, we keep it and don’t replace it with whatever might be discovered from `build.sbt`. That way the plugin’s root wins even when there’s a root defined in the build file.

Result: builds that define the root only via `extraProjects` load successfully, and the plugin’s root is the one that’s used. Existing builds that don’t use a root in `extraProjects` behave as before.


bitloi:
@eed3si9n Ready for review!

eed3si9n:
Thanks. Please add a scripted test to demonstrate Dale's use case.

bitloi:
> Thanks. Please add a scripted test to demonstrate Dale's use case.

I just added it. Could you please take another look at the updates?

eed3si9n (approved):
Thanks!

bitloi:
@eed3si9n Is it possible to merge the PR?

PR #8696, opened 2026-02-05
Fix TODO/FIXME items: improve exception handling and resource copying

Eruis2579:
- Fix FIXME in IvyXml.scala: Replace too-wide Throwable catch with specific exceptions (NoSuchMethodException, IllegalAccessException, InvocationTargetException, ClassCastException)

- Fix TODO in Defaults.scala: Add copyResources.value in makePickleProducts to ensure resources are copied when using early output/build pipelining

- Improve skipBanner command: Check file existence before creating to avoid unnecessary exception handling


eed3si9n:
For Gittensor-related contributions, please stick to accepted GitHub issues and follow the contribution guideline https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md.

PR #8697, opened 2026-02-05
[2.x] fix: Respect explicit platform settings in dependency resolution

### LINKED ISSUE #8665: Incorrect implicit platform dependency for scala3_library
sbt 2.0.0-RC8

## steps

1. Define build.sbt 
```scala
val scala3Version = "3.7.4"
val scala213Version = "2.13.18"

platform := "native0.5"
scalaVersion := scala3Version
crossScalaVersions := Seq(scala3Version, scala213Version)
organization := "org.example"

lazy val base = project
  .in(file("base"))

lazy val projectA = project
  .in(file("project-A"))
  .settings(
    libraryDependencies += (organization.value %% (base / normalizedName).value % version.value)
  )

```

2. Execute in sbt: 
```
>  +base/publishLocal 

# Works
> ++2.13.18 projectA/compile 

# Fails
++3.7.4 projectA/compile
```

## problem

```
[warn]  Note: Unresolved dependencies path:
[error] stack trace is suppressed; run last projectA / update for the full output
[error] (projectA / update) sbt.librarymanagement.ResolveException: Error downloading org.scala-lang:scala3-library_native0.5_3:3.7.4
```

## expectation

Setting platform should not introduce implicit platform suffix for transitive/implicit dependencies. 

## notes

Found in Scala Native scripted test testing Scala 2.13 <-> Scala 3 compat tests when adding sbt 2.x support https://github.com/scala-native/scala-native/pull/4768

Eruis2579:
Fixes #8665

**Problem**
When `platform := "native0.5"` is set, it incorrectly adds the platform suffix to implicit Scala library dependencies like `scala3-library`, even though they explicitly set `.platform(Platform.jvm)`. This causes resolution errors when trying to resolve `scala3-library_native0.5_3` which doesn't exist.

The issue is in `addPlatformSuffix` which always uses the project platform if it exists, overriding explicit platform settings on dependencies.

**Solution**
Modified `addPlatformSuffix` to prioritize explicit platform settings on dependencies. If a dependency has an explicit platform (including "jvm"), use that instead of the project platform. The project platform should only apply to dependencies without an explicit platform.

This ensures that:
- Auto-injected Scala library dependencies (which have `.platform(Platform.jvm)`) are not overridden
- Explicit dependencies with platforms work correctly
- Dependencies without explicit platforms still get the project platform

**Testing**
When `autoScalaLibrary := false`, the compilation succeeds, confirming the issue is with the auto-injected library. This fix ensures the auto-injected library respects its explicit "jvm" platform setting.


eed3si9n:
Hi, thanks for the contribution!

Please check out our [Contributor's Guildeline](https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md) and:
3. Follow our Gen-AI assisted contributions guideline (please disclose AI usages in the PR description, if any)

eed3si9n (approved):
Thanks!

PR #8698, opened 2026-02-06
 [2.x] fix: Use correct configuration identifier for display

### LINKED ISSUE #5211: Error in configuration name translation in logging
## steps
sbt version: 1.3.3

Use sbt-multi-jvm, run tests

## problem

Output while running:

```
  | => akka-sample-distributed-data-scala / Multi-jvm / executeTests 80s
```

Output when done:

```
[error] (Multi-jvm / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 108 s (01:48), completed 24-Oct-2019 13:23:31
```

## expectation

```
  | => akka-sample-distributed-data-scala / MultiJvm / executeTests 80s`
```

and

```
[error] (MultiJvm / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 108 s (01:48), completed 24-Oct-2019 13:23:31
```

## notes
`MultiJvm` is the name of the configuration in the shell and in code.  It's `multi-jvm` as it's ivy configuration identifier.  There shouldn't be a 3rd variant.

bitloi:
Fixes #5211

### Problem

Configurations like `MultiJvm` display as `Multi-jvm / compile` instead of `MultiJvm / compile`. The display logic guesses the identifier by capitalizing the ivy config name, which doesn't work when they differ.

### Solution

Added a lookup map from config name to identifier in `KeyIndex`, and updated `Scope.display` to use it instead of guessing.


bitloi:
@eed3si9n I’d appreciate your review for this PR.

eed3si9n (changes requested):
Here's a step for manual testing.

### project/FooPlugin.scala

```scala
package example

import sbt.*
import Keys.*

object FooPlugin extends AutoPlugin:
  lazy val MultiJvm = config("multi-jvm")

  override def requires = plugins.JvmPlugin

  lazy val test1 = taskKey[Unit]("")

  override def projectConfigurations = Seq(MultiJvm)
  override def projectSettings = Seq(
    MultiJvm / test1 := Def.uncached(Thread.sleep(1000)),
  )
```

An empty `build.sbt` file.

Then, from the sbt shell:

```
> inspect MultiJvm/test1
```

This outputs

```
[info] Delegates:
[info] 	Multi-jvm / test1
[info] 	test1
[info] 	ThisBuild / Multi-jvm / test1
[info] 	ThisBuild / test1
[info] 	Zero / Multi-jvm / test1
[info] 	Global / test1
```

or try

```
> MultiJvm/test1
```

The super shell still displays the progress as `Multi-jvm / test1`

bitloi:
> Here's a step for manual testing.
> 
> ### project/FooPlugin.scala
> ```scala
> package example
> 
> import sbt.*
> import Keys.*
> 
> object FooPlugin extends AutoPlugin:
>   lazy val MultiJvm = config("multi-jvm")
> 
>   override def requires = plugins.JvmPlugin
>   override def trigger = allRequirements
> 
>   lazy val test1 = taskKey[Unit]("")
> 
>   override def projectConfigurations = Seq(MultiJvm)
>   override def projectSettings = Seq(
>     MultiJvm / test1 := Def.uncached(Thread.sleep(1000)),
>   )
> ```
> 
> An empty `build.sbt` file.
> 
> Then, from the sbt shell:
> 
> ```
> > inspect MultiJvm/test1
> ```
> 
> This outputs
> 
> ```
> [info] Delegates:
> [info] 	Multi-jvm / test1
> [info] 	test1
> [info] 	ThisBuild / Multi-jvm / test1
> [info] 	ThisBuild / test1
> [info] 	Zero / Multi-jvm / test1
> [info] 	Global / test1
> ```
> 
> or try
> 
> ```
> > MultiJvm/test1
> ```
> 
> The super shell still displays the progress as `Multi-jvm / test1`

Fixed.

eed3si9n:
It seems like it's not fixed for supershell:

```bash
  | => foo / Multi-jvm / test1 3s
```

bitloi:
> It seems like it's not fixed for supershell:
> 
> ```shell
>   | => foo / Multi-jvm / test1 3s
> ```

Fixed the supershell display too.

mzuehlke:
Wasn't it planed to deprecated custom configurations in sbt 2 altogether ?

bitloi:
> Wasn't it planed to deprecated custom configurations in sbt 2 altogether ?

 This fix isn't about adding new custom configuration support — it just corrects the display of existing configurations. Even if custom configurations are deprecated in the future, they still need to display correctly while they're supported.

eed3si9n:
@mzuehlke 

> Wasn't it planed to deprecated custom configurations in sbt 2 altogether ?

I can provide a bit of context on this one. I originally proposed it as [RFC-3: drop custom config](https://eed3si9n.com/sbt-drop-custom-config/), but as of 2023, the outcome is that we can remove `IntegrationTest` and similar task scoping, but we will keep custom configuration around for scoping dependency graph (a more Bazel-inspired mechanism would be to create synthetic subproject).

bitloi:
@eed3si9n Would you give me your feedback?

eed3si9n:
@bitloi I'm still seeing the same issue. You can change my test to:

```scala
MultiJvm / test1 := Def.uncached(Thread.sleep(5000)),
```

this should give you time to copy-paste:

```bash
[warn]   scalaVersion := "3.8.1"

  | => foo / Multi-jvm / test1 2s
```

Does it work for you on your machine?

bitloi:
> @bitloi I'm still seeing the same issue. You can change my test to:
> 
> ```scala
> MultiJvm / test1 := Def.uncached(Thread.sleep(5000)),
> ```
> 
> this should give you time to copy-paste:
> 
> ```shell
> [warn]   scalaVersion := "3.8.1"
> 
>   | => foo / Multi-jvm / test1 2s
> ```
> 
> Does it work for you on your machine?

 the lookup was scoped to `None` (global) which has no config mappings. Fixed by building the config name-to-ident map from all projects' configurations, same approach as the other display paths. Should work for the supershell now.

main/src/main/scala/sbt/MainLoop.scala
```diff
@@ -156,7 +156,18 @@ private[sbt] object MainLoop:
...
+    val configNameToIdent: String => String =
+      if Project.isProjectLoaded(state) then
+        val st = Project.structure(state)
+        val configMap = (for
+          (_, unit) <- st.units.iterator
+          (_, project) <- unit.defined.iterator
+          config <- project.configurations.iterator
+        yield config.name -> config.id).toMap
+        name => configMap.getOrElse(name, Scope.guessConfigIdent(name))
+      else Scope.guessConfigIdent
```
eed3si9n: Could you move this function to ProjectExtra please?

bitloi: Moved to `ProjectExtra.configNameToIdent(state)`. `MainLoop` now just calls `Project.configNameToIdent(state)`.

main/src/main/scala/sbt/internal/AbstractTaskExecuteProgress.scala
```diff
@@ -17,10 +17,12 @@ import scala.collection.mutable
...
 import scala.concurrent.duration.*
 
-private[sbt] abstract class AbstractTaskExecuteProgress extends ExecuteProgress {
+private[sbt] abstract class AbstractTaskExecuteProgress(
+    configNameToIdent: String => String = Scope.guessConfigIdent
+) extends ExecuteProgress {
   import AbstractTaskExecuteProgress.Timer
 
-  private val showScopedKey = Def.showShortKey(None)
+  private val showScopedKey = Def.showShortKey(None, configNameToIdent)
```
eed3si9n, line 25: Could you add a scripted test that defines a custom command that calls this `showScopedKey` that checks for `MultiJvm / test1`?

bitloi: Added it.

eed3si9n (approved):
Thanks!

bitloi:
Is it possible to merge?

PR #8701, opened 2026-02-06
[2.x] feat: drop other idle servers on client exit (#8610)

### LINKED ISSUE #8610: Drop other idle servers on exit

calm329:
## Summary

Implements cooperative idle server cleanup for `sbtn` (issue #8610). When a client disconnects from an sbt server, that server notifies all
other registered servers to shut down if they've been idle long enough and have no connected clients. This prevents accumulation of idle
background JVMs across projects.

fixes #8610

## Problem

Each `sbtn` session leaves a background JVM server running with a 7-day idle timeout. Users working across many projects accumulate idle
servers that waste system resources (memory, file descriptors, threads).

## Solution

- Each server registers itself via a `<PID>.json` file in a shared proc directory (`~/.cache/sbt/v2/proc/` on Linux,
`~/Library/Caches/sbt/v2/proc/` on macOS).
- When a client disconnects, the server sends a `sbt/dropIfIdle` JSON-RPC notification to all other registered servers.
- A server receiving `dropIfIdle` shuts down if idle >= `sbt.server.secondaryIdleTimeout` (default 600s) AND has no initialized
`NetworkChannel` clients.
- Stale proc files (unreachable servers) are cleaned up on contact failure.
- Anti-cascading: only initialized channels (real clients that completed the `initialize` handshake) trigger `notifyOtherServers` on
disconnect. Temporary `dropIfIdle` connections never initialize, so their removal won't cascade.

## Testing

- Manual: started server with `-Dsbt.server.secondaryIdleTimeout=5`, verified proc file created, sent `dropIfIdle` notification, confirmed
server drops when idle with no clients and stays alive when a client is connected.
- Verified proc file cleanup on shutdown.

### DIFF
--- a/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
@@ -92,8 +92,6 @@ private[sbt] object LanguageServerProtocol {
       },
       onResponse = PartialFunction.empty,
       onNotification = {
-        case n: JsonRpcNotificationMessage if n.method == sbt.protocol.Serialization.dropIfIdle =>
-          StandardMain.exchange.handleDropIfIdle(name)
         case n: JsonRpcNotificationMessage if n.method == "textDocument/didSave" =>
           val _ = appendExec(";Test/compile; collectAnalyses", None)
       }
--- a/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
@@ -92,6 +92,8 @@ private[sbt] object LanguageServerProtocol {
       },
       onResponse = PartialFunction.empty,
       onNotification = {
+        case n: JsonRpcNotificationMessage if n.method == sbt.protocol.Serialization.dropIfIdle =>
+          StandardMain.exchange.handleDropIfIdle(name)
         case n: JsonRpcNotificationMessage if n.method == "textDocument/didSave" =>
           val _ = appendExec(";Test/compile; collectAnalyses", None)
       }
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -441,6 +441,7 @@ object Defaults extends BuildCommon {
           semanticdbVersion.value
         ),
         VirtualTerminal.handler,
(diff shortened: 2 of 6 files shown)


calm329:
@eed3si9n Could you please review this pr?

main/src/main/scala/sbt/internal/server/LanguageServerProtocol.scala
```diff
@@ -92,6 +92,8 @@ private[sbt] object LanguageServerProtocol {
       },
       onResponse = PartialFunction.empty,
       onNotification = {
+        case n: JsonRpcNotificationMessage if n.method == sbt.protocol.Serialization.dropIfIdle =>
+          StandardMain.exchange.handleDropIfIdle(name)
```
eed3si9n: Since this method isn't part of LSP, I think we should define a separate handler like:

```scala
  val idleHandler: ServerHandler = ServerHandler: callback =>
    ServerIntent(
      onRequest = PartialFunction.empty,
      onResponse = PartialFunction.empty,
      onNotification = {
        case n if n.method == dropIfIdle =>
          StandardMain.exchange.handleDropIfIdle()
          ()
      }
    )
```

somewhere, like object CommandExchange.

eed3si9n (commented):
Left a minor comment, but overall LGTM. Thanks @calm329!

eed3si9n (approved):
Thanks!

PR #8695, opened 2026-02-05
[2.x] fix: restore CLI precedence over .sbtopts (#8685)

it-education-md:
Fixes #8685

# 1. Summary

Restore command-line JVM argument precedence over `.sbtopts` in the launcher script and add a
regression test.

# 2. Problem

According to the issue, `.sbtopts` entries are appended after
CLI args in sbt 1.12.x, so `.sbtopts` JVM memory settings (e.g., `-Xmx2g`) override CLI `--mem`,
causing invalid JVM settings.

# 3. Reproduction

### Step 1. Create `tmp/.sbtopts`:

```
-J-Xmx2g
-J-XX:ReservedCodeCacheSize=1g
-J-XX:MaxMetaspaceSize=2g
-J-Xss512m
-J-XX:+UseG1GC
```

### Step 2. Run:

```bash
cd tmp
../sbt -d -mem 12288 -Dsbt.boot.lock=false test
```

<details>
<summary>Logs (Current behavior)</summary>

```
[addSbt] arg = '-debug'
[addMemory] arg = '12288'
[addJava] arg = '-Xms12288m'
[addJava] arg = '-Xmx12288m'
[addJava] arg = '-Xss4M'
[addJava] arg = '-XX:ReservedCodeCacheSize=512m'
[addJava] arg = '-Dsbt.boot.lock=false'
[residual] arg = 'test'
[addJava] arg = '-Xmx2g'
[addJava] arg = '-XX:ReservedCodeCacheSize=1g'
[addJava] arg = '-XX:MaxMetaspaceSize=2g'
[addJava] arg = '-Xss512m'
[addJava] arg = '-XX:+UseG1GC'
[residual] arg = 'test'
[sbt_options] declare -a sbt_options=()
[process_args] java_version = '17'
[addJava] arg = '-Dsbt.script=../sbt'
[copyRt] java9_rt = '/rt.jar'
mkdir: cannot create directory ‘’: No such file or directory
Error occurred during initialization of VM
Initial heap size set to a larger value than the maximum heap size
[addJava] arg = '-Dscala.ext.dirs='
# Executing command line:
java
-Dfile.encoding=UTF-8
-Xms12288m
-Xmx12288m
-Xss4M
-XX:ReservedCodeCacheSize=512m
-Dsbt.boot.lock=false
-Xmx2g
-XX:ReservedCodeCacheSize=1g
-XX:MaxMetaspaceSize=2g
-Xss512m
-XX:+UseG1GC
-Dsbt.script=../sbt
[...]

PR #8706, opened 2026-02-07
[1.x] bport: restore CLI precedence over .sbtopts

eed3si9n:
This is a backport of https://github.com/sbt/sbt/pull/8695

According to the issue, `.sbtopts` entries are appended after CLI args in sbt 1.12.x, so `.sbtopts` JVM memory settings (e.g., `-Xmx2g`) override CLI `--mem`, causing invalid JVM settings.


er1c:
Ha, thanks for the bport again, got it faster than me :)

PR #8708, opened 2026-02-07
[2.x] fix: Exclude autoStartServer from lintUnused warnings (#6624)

### LINKED ISSUE #6624: Setting autoStartServer to false gives spurious warnings
## steps

In ~/.sbt/1.0/global.sbt, I have added:

```
Global / autoStartServer := false 
```

This works, but gives some warnings:
```
[warn] there's a key that's not used by any other settings/tasks:
[warn]  
[warn] * Global / autoStartServer
[warn]   +- /home/hrj/.sbt/1.0/global.sbt:4
[warn]  
```

## problem
Although it works, the warning makes me wonder if I am setting it correctly.

## expectation
If my usage of the key is correct, it would be better if the sbt doesn't print the warning.

Eruis2579:
**Problem**
Setting \`Global / autoStartServer := false\` in global.sbt triggers a
spurious lintUnused warning, even though the setting is correctly used
by sbt's server startup logic.

**Solution**
Add \`autoStartServer\` to the \`excludeLintKeys\` set in LintUnused.scala,
similar to other server-related settings like \`serverConnectionType\` and
\`serverIdleTimeout\`. This prevents the warning while maintaining the
functionality of the setting.

**Testing**
- Added scripted test \`lint-autostartserver\` to verify the fix
- Test verifies that setting \`Global / autoStartServer := false\` does
  not trigger lintUnused warnings

Fixes #6624


eed3si9n (approved):
Thanks!

PR #8709, opened 2026-02-07
[2.x] fix : BSP compile returns StatusCode.Error on failure (#8104)

### LINKED ISSUE #8104: Failed BSP compilation through Metals does not return `StatusCode.Error`
## steps

Given the following project in Metals

```scala
// build.sbt
Global / scalaVersion := "2.12.17"

val proj1 = project
    .settings(
        version := "1.0",
        scalacOptions += "-Xlint:infer-any",
    )
```

```
// build.properties
sbt.version=1.10.7
```

```scala
// proj1/src/main/scala/Main.scala
object Main extends App {
  val x = "hi"
  println(x / 2)
}
```

The project correctly fails, but in the BSP logs I see it failing as

```
[Trace - 08:06:38 AM] Received response 'buildTarget/compile - (46)' in 190ms
Result: null
Error: {
  "code": -32603,
  "message": "(proj1 / Compile / compileIncremental) Compilation failed"
}
```

## problem

The return value of `buildTarget/compile` was a JSON-RPC error. This leads to Metals substituting in a `StatusCode.Cancelled` outcome, which affects other confusing outcomes.

## expectation

`buildTarget/compile` result should be a successful `BspCompileResult` value with `statusCode = StatusCode.Error`.

Eruis2579:
Fixes https://github.com/sbt/sbt/issues/8104

- BuildServerProtocol: for Result.Inc(cause), return StatusCode.Error for any non-InterruptedException (was throwing for non-CompileFailed, causing JSON-RPC error instead of BspCompileResult with statusCode Error)
- BuildServerTest: add test 'buildTarget/compile - returns StatusCode.Error when compilation fails' (introduce compile error, compile via BSP, assert statusCode == Error)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ launcher-package/citest/freshly-baked
 .vscode
 sbt-launch.jar
 local-temp
+.jdk
--- a/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
@@ -879,7 +879,7 @@ object BuildServerProtocol {
       case Result.Inc(cause) =>
         cause.getCause match {
           case _: InterruptedException => StatusCode.Cancelled
-          case _ =>
+          case _                       =>
             // Return Error for any compile failure (CompileFailed or other Incomplete)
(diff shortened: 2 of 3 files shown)


eed3si9n (approved):
Thanks!

PR #8715, opened 2026-02-08
[2.x] fix: handle --script-version sbt 2.x project dirs

### LINKED ISSUE #8711: sbt --script-version doesn't work on sbt 2.x project directories
## Steps

1. Make a sbt 2.x project.
    1. `cd /tmp`
    2.  `mkdir hello`
    3. `cd hello`
    4. `touch build.sbt`
    5. `mkdir project`
    6. `echo "sbt.version=2.0.0-RC8" > project/build.properties`
2.   `$ $HOME/work/sbt/sbt --script-version -v`

## Problem

It runs sbtn instead of printing out the script version:

```bash
$ $HOME/work/sbt/sbt --script-version -v
....
[debug] running native client
# Executing command line:
/Users/xxx/.cache/sbt/boot/sbtn/1.12.1/sbtn
--sbt-script=/Users/xxx/work/sbt/sbt
--script-version
-v

```

## Expectation

```bash
$ $HOME/work/sbt/sbt --script-version -v
....
[process_args] java_version = '8'
_to_be_replaced
```

Print out the script version. If you used the one in sbt/sbt it should say `_to_be_replaced`.

Eruis2579:
## What
Fixes #8711: `sbt --script-version` now works when run from an sbt 2.x project directory.

## Problem
From a project with `project/build.properties` set to sbt 2.x (e.g. `sbt.version=2.0.0-RC8`), the script defaults to the native client (sbtn). The `--script-version` flag was only handled inside `run()` (the JVM launcher path), so when the script took the native client path it never checked `--script-version` and instead ran sbtn with that flag. sbtn doesn’t implement it, so the user saw "failed to connect to server" instead of the script version.

## Solution
Handle `--script-version` right after `process_args` and before the `isRunNativeClient` branch. If `print_sbt_script_version` is set, print `$init_sbt_version` and exit, so we never delegate to sbtn for this case.

## Testing
- From repo root (sbt 1.x): `./sbt --script-version` → prints `_to_be_replaced`
- From sbt 2.x project dir: `sbt --script-version` → prints `_to_be_replaced` (no sbtn run)

### DIFF
--- a/sbt
+++ b/sbt
@@ -570,8 +570,6 @@ run() {
   detect_working_directory
   if [[ $print_sbt_version ]]; then
     execRunner "$java_cmd" -jar "$sbt_jar" "sbtVersion" | tail -1 | sed -e 's/\[info\]//g'
-  elif [[ $print_sbt_script_version ]]; then
-    echo "$init_sbt_version"
   elif [[ $print_version ]]; then
     if [[ -n "$is_this_dir_sbt" ]]; then
       execRunner "$java_cmd" -jar "$sbt_jar" "sbtVersion" | tail -1 | sed -e 's/\[info\]/sbt version in this project:/g'
--- a/sbt
+++ b/sbt
@@ -905,6 +905,12 @@ args1=( "${cli_options[@]}" "${cli_commands[@]}" "${sbt_additional_commands[@]}"
 process_args "${args1[@]}"
 vlog "[sbt_options] $(declare -p sbt_options)"
 
+# Handle --script-version before native client so it works on sbt 2.x project dirs (#8711)
+if [[ $print_sbt_script_version ]]; then
+  echo "$init_sbt_version"
+  exit 0
+fi
+
 if [[ "$(isRunNativeClient)" == "true" ]]; then
   set -- "${residual_args[@]}"
   argumentCount=$#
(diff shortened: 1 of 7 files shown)


sbt
```diff
@@ -905,6 +905,12 @@ args1=( "${cli_options[@]}" "${cli_commands[@]}" "${sbt_additional_commands[@]}"
 process_args "${args1[@]}"
 vlog "[sbt_options] $(declare -p sbt_options)"
 
+# Handle --script-version before native client so it works on sbt 2.x project dirs (#8711)
+if [[ $print_sbt_script_version ]]; then
+  echo "$init_sbt_version"
+  exit 0
+fi
```
eed3si9n, line 910: 1. Could you refactor https://github.com/sbt/sbt/blob/develop/launcher-package/integration-test/src/test/scala/RunnerScriptTest.scala so it runs on two variants:
   1. the current `launcher-package/citest` and 
   2. new `launcher-package/citest2` whose `project/build.properties` are updated to `2.0.0-RC8`
2. Could you refactor the fix so `--script-version` isn't handled twice in this script? (There's already simlar code around line 570).

eed3si9n (changes requested):

sbt
Eruis2579: I updated code. Thank your for your review

eed3si9n (approved):
Thanks!

PR #8719, opened 2026-02-08
[2.x] fix: Fixes evicted warning for version intervals

### LINKED ISSUE #6244: False evicted warning for versions intervals
## steps

Using sbt 1.4.6 or 1.3.13 I do get a false evicted warning for a library which uses a version interval.

1. Run `sbt` in a terminal
2. Execute `evicted`

## problem

I get an evicted warning for this library:

```
[warn]  * net.minidev:json-smart:2.3 is selected over {[1.3.1,2.3], [1.3.1,2.3]}
[warn]      +- com.nimbusds:oauth2-oidc-sdk:8.22                  (depends on [1.3.1,2.3])
[warn]      +- com.nimbusds:nimbus-jose-jwt:8.20.1                (depends on [1.3.1,2.3])
```

## expectation

Do not treat this situation as an evicted warning, because selecting version 2.3 is a valid option when both libraries that require `net.minidev:json-smart` allow version `2.3` in a version interval `(depends on [1.3.1,2.3])`

## notes

This is probably an error in `sbt/librarymanagement` that is affecting a lot of sbt versions.

Eruis2579:
Fixes #6244

### Problem

When a dependency declares a version range (e.g. Ivy [1.3.1,2.3] or comma-separated "1.3.1,2.3" as used by Coursier), and the resolver picks a version inside that range, sbt was still reporting an eviction warning. The chosen version satisfies the range, so it should not be reported as an eviction.

Example (from #6244): oauth2-oidc-sdk and nimbus-jose-jwt both depend on net.minidev:json-smart with range [1.3.1,2.3]. Resolution selects 2.3, which is within the range. Before this fix, sbt reported an eviction for json-smart even though 2.3 satisfies [1.3.1,2.3].

### Root cause

1. Version range detection only treated Maven-style brackets and "+" as ranges; comma-separated strings like "1.3.1,2.3" (no brackets) were not recognized as ranges.
2. Eviction "compatibility" did not check whether the winner revision satisfies the evicted revision when the evicted revision was a version range.

### Solution

1. lm-core – VersionRange (sbt/internal/librarymanagement/VersionRange.scala):
   - isVersionRange: treat comma-separated strings with digits (e.g. "1.3.1,2.3") as version ranges.
   - versionSatisfiesRange: support bracket-less comma-separated ranges by parsing "lower,upper" as inclusive [lower, upper]. Existing Maven-style ranges ([1.3.1,2.3], [4.1.0,5), etc.) and "+" ranges are unchanged.

2. lm-core – EvictionWarning (sbt/librarymanagement/EvictionWarning.scala):
   - In guessCompatible: consider an eviction compatible (do not report) when the evicted revision is a version range and the winner revision satisfies that range via VersionRange.versionSatisfiesRange. Same-version check is unchanged.

### Testing

- lmCore/Test/testOnly sbt.librarymanagement.VersionRangeSpec
- lmIvy/Test/testOnly sbt.internal.librarymanagement.EvictionWarningSpec

### DIFF
--- a/lm-core/src/main/scala/sbt/internal/librarymanagement/VersionRange.scala
+++ b/lm-core/src/main/scala/sbt/internal/librarymanagement/VersionRange.scala
@@ -12,7 +12,9 @@ object VersionRange {
     (revision.contains("[")) ||
     (revision.contains("]")) ||
     (revision.contains("(")) ||
-    (revision.contains(")"))
+    (revision.contains(")")) ||
+    // Comma-separated range e.g. "1.3.1,2.3" (fixes #6244 when Coursier passes range without brackets)
+    (revision.contains(",") && revision.exists(_.isDigit))
   }
 
   /**
@@ -25,6 +27,15 @@ object VersionRange {
     if (!isVersionRange(range)) {
       // Not a range, just compare directly
       version == range
+    } else if (range.contains(",") && !hasMavenVersionRange(range)) {
+      // Comma-separated range without brackets e.g. "1.3.1,2.3" (fixes #6244)
+      val parts = range.split(",", 2)
+      if (parts.length == 2) {
+        val lower = parts(0).trim
+        val upper = parts(1).trim
+        lower.nonEmpty && upper.nonEmpty &&
+        compareVersions(version, lower) >= 0 && compareVersions(version, upper) <= 0
+      } else false
     } else if (range.endsWith("+")) {
       // Handle plus ranges like "1.0+" meaning >= 1.0
       val base = range.dropRight(1)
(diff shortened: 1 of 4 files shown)


eed3si9n (approved):
Thanks for the contribution! Overall the change looks good to me.

eed3si9n:
FYI - I'll be traveling for work next week, so my speed for reviews might slow down.

Eruis2579:
> FYI - I'll be traveling for work next week, so my speed for reviews might slow down.

Thank you for your review

Eruis2579:
Hi, @eed3si9n When can this PR be merged?

PR #8725, opened 2026-02-09
[2.x] fix: Note that dynamic usage isn't going to be linted

### LINKED ISSUE #5647: Dynamic task or .all(<filter>) show up as "not used" keys lint
## steps

Using a 1.4.0-SNAPSHOT build off of Ethan's sbtc PR (#5620), in the Zinc repo, with sbtc.bash sourced, `sbtc <TAB>` (i.e. start the build)

## problem

```
[warn] there are 2 keys that are not used by any other settings/tasks:
[warn]
[warn] * Global / shellPrompt
[warn]   +- /Users/dnw/.sbt/1.0/global.sbt:13
[warn] * ThisBuild / whitesourceIgnoredScopes
[warn]   +- /d/zinc/build.sbt:768
[warn]
```

## expectation

sbt-whitesource defines a default for `whitesourceIgnoredScopes` in Global scope:

https://github.com/lightbend/sbt-whitesource/blob/513b71b9ac4d96f7ef9510ee957752996098103d/src/main/scala/sbtwhitesource/WhiteSourcePlugin.scala#L110-L122

and then consumes it in project scope:

https://github.com/lightbend/sbt-whitesource/blob/513b71b9ac4d96f7ef9510ee957752996098103d/src/main/scala/sbtwhitesource/WhiteSourcePlugin.scala#L193

so defining it in build scope shouldn't be "not used".

bitloi:
Fixes #5647

## Summary

Keys that are only used through `key.all(ScopeFilter).value` were not seen by the unused-settings lint, so they could be reported as unused even when they were in use.

This PR makes those “dynamic” usages visible to the lint by:

- Recording dynamic dependencies when `SettingKey.all(filter)` / `TaskKey.all(filter)` are used (via `Def.withDynamicDependencies` and a small extension to `Initialize`).
- In the lint pass, collecting those dynamic deps from all settings, expanding them to concrete `ScopedKey`s with the build structure, and adding those to the set of used keys.

So any key that is only referenced via `.all(scopeFilter).value` is now treated as used and no longer flagged as unused.

### DIFF
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -701,7 +701,9 @@ object Keys {
   @deprecated("No longer used", "1.3.0")
   private[sbt] val executeProgress = settingKey[State => TaskProgress]("Experimental task execution listener.").withRank(DTask)
   val commandProgress = settingKey[Seq[ExecuteProgress2]]("Command progress listeners receive events when commands start and end, in addition to task progress events.")
-  val lintUnused = inputKey[Unit]("Check for keys unused by other settings and tasks.")
+  val lintUnused = inputKey[Unit](
+    "Check for keys unused by other settings and tasks. Keys that are only used via key.all(ScopeFilter).value may still be reported as unused."
+  )
   val lintIncludeFilter = settingKey[String => Boolean]("Filters key names that should be included in the lint check.")
   val lintExcludeFilter = settingKey[String => Boolean]("Filters key names that should be excluded in the lint check.")
   val excludeLintKeys = settingKey[Set[Def.KeyedInitialize[?]]]("Keys excluded from lintUnused task")
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -701,9 +701,7 @@ object Keys {
   @deprecated("No longer used", "1.3.0")
   private[sbt] val executeProgress = settingKey[State => TaskProgress]("Experimental task execution listener.").withRank(DTask)
   val commandProgress = settingKey[Seq[ExecuteProgress2]]("Command progress listeners receive events when commands start and end, in addition to task progress events.")
-  val lintUnused = inputKey[Unit](
-    "Check for keys unused by other settings and tasks. Keys that are only used via key.all(ScopeFilter).value may still be reported as unused."
-  )
+  val lintUnused = inputKey[Unit]("Check for keys unused by other settings and tasks. Keys that are only used via key.all(ScopeFilter).value may still be reported as unused.")
   val lintIncludeFilter = settingKey[String => Boolean]("Filters key names that should be included in the lint check.")
   val lintExcludeFilter = settingKey[String => Boolean]("Filters key names that should be excluded in the lint check.")
   val excludeLintKeys = settingKey[Set[Def.KeyedInitialize[?]]]("Keys excluded from lintUnused task")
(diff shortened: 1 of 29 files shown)


bitloi:
@eed3si9n Ready for review! Please give me your feedback when you are available.

util-collection/src/main/scala/sbt/internal/util/Settings.scala
```diff
@@ -630,6 +630,10 @@ trait Init:
    */
   sealed trait Initialize[A1]:
     def dependencies: Seq[ScopedKey[?]]
+
+    /** Dynamic dependencies (e.g. from .all(ScopeFilter)) not visible in the static graph. */
+    private[sbt] def dynamicDependencies: Seq[Any] = Nil
```
eed3si9n: While I appreciate the effort adding a whole new mechanism at the Initialized level just for dynamic tasks doesn't seem worth it.

eed3si9n (changes requested):

util-collection/src/main/scala/sbt/internal/util/Settings.scala
bitloi: Thanks for the review. I get that this adds more than we want for this case.
I’m open to simplifying. Do you’d rather we (a) try a lighter fix that only touches the lint (e.g. some heuristic so keys used via `.all(scopeFilter).value` aren’t reported unused), or (b) leave the current behavior and just document that keys only used that way can be flagged as unused?

eed3si9n: > (b) leave the current behavior and just document that keys only used that way can be flagged as unused.

b, I think.

bitloi: Just fixed it.

main/src/main/scala/sbt/ScopeFilter.scala
```diff
@@ -218,6 +218,25 @@ object ScopeFilter {
...
+      .mapValues: byProj =>
+        byProj
+          .groupBy(_.config)
+          .view
+          .mapValues: byConfig =>
+            byConfig.groupBy(_.task).view.mapValues(_.toSet).toMap
+          .toMap
+      .toMap
+    new Data(units, resolve, new AllScopes(scopes, grouped))
+
```
eed3si9n: This is no longer used, correct?

main/src/main/scala/sbt/Keys.scala
```diff
@@ -701,7 +701,9 @@ object Keys {
   @deprecated("No longer used", "1.3.0")
   private[sbt] val executeProgress = settingKey[State => TaskProgress]("Experimental task execution listener.").withRank(DTask)
   val commandProgress = settingKey[Seq[ExecuteProgress2]]("Command progress listeners receive events when commands start and end, in addition to task progress events.")
-  val lintUnused = inputKey[Unit]("Check for keys unused by other settings and tasks.")
+  val lintUnused = inputKey[Unit](
+    "Check for keys unused by other settings and tasks. Keys that are only used via key.all(ScopeFilter).value may still be reported as unused."
+  )
```
eed3si9n: This change is fine, but could we revert all other changes, if they are code format/style changes?

main/src/main/scala/sbt/ScopeFilter.scala
bitloi: Correct — it was dead after the revert. Removed `dataFromStructure` and the unused `BuildStructure` import

main/src/main/scala/sbt/Keys.scala
bitloi: Reverted the style/format changes in ScopeFilter, `LintUnused`, and `Settings`, and kept the `lintUnused` description on one line. Only the doc text change remains.

eed3si9n (approved):

bitloi:
@eed3si9n Can we merge now?

PR #8726, opened 2026-02-09
[2.x] Fix #4840: lastGrep ignores ANSI escape sequences when matching

### LINKED ISSUE #4840: lastGrep should ignore ANSI escape sequences
sbt version: 1.2.8

`lastGrep` greps the raw output, including any ANSI color sequences, which leads to unexpected and unwanted effects, e.g.:

```
sbt:root> lastGrep error.* library/mimaReportBinaryIssues
[error]  * method foo()Int in class scala.collection.immutable.BitSet does not have a correspondent in other version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.collection.immutable.BitSet.foo")
sbt:root> lastGrep \[.*error.* library/mimaReportBinaryIssues
[0m[error]  * method foo()Int in class scala.collection.immutable.BitSet does not have a correspondent in other version
[0m[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.collection.immutable.BitSet.foo")
sbt:root> lastGrep \[error.* library/mimaReportBinaryIssues
sbt:root>
```

`error` is red in this case, eveything else has the default color, so there are at least two escape sequences per line, one to switch to red and the other one to switch back.

It would be more useful to filter out all escape sequences and grep the filtered output.

Bonus points for keeping track of all ANSI attributes, mapping the positions back to the attributed soutce, and synthesizing a new escaped version.

Eruis2579:
## Summary
Fixes [#4840](https://github.com/sbt/sbt/issues/4840): `lastGrep` now matches patterns against visible text by stripping ANSI escape sequences before applying the regex, so output that contains color or other escape codes is matched correctly.

## Problem
When `last` output includes ANSI escape sequences (e.g. colored `[error]` lines), `lastGrep <pattern>` was matching against the raw string. The pattern could fail to match because it was compared to text like `\u001B[31merror\u001B[0m` instead of `error`, or matching was inconsistent.

## Solution
- Strip ANSI from each line before running the regex, using `EscHelpers.stripColorsAndMoves` from `sbt.internal.util`.
- **`Output.lastGrep`** (keys and file overloads): lines from the last run are stripped, then the pattern is applied to the stripped text; matching and printed lines are based on visible text.
- **`Output.grep`**: each line is stripped before `showMatches(pattern)` so the pattern is applied only to visible content.

Imports added in `Output.scala`: `sbt.internal.util.EscHelpers`. No change to `last` (unfiltered) output.

## Changes
- **`main/src/main/scala/sbt/internal/Output.scala`**: Use `EscHelpers.stripColorsAndMoves(line)` before pattern matching in both `lastGrep` overloads and in `grep`.
- **`main/src/test/scala/sbt/internal/OutputSpec.scala`**: New tests for `grep` with ANSI in the input (pattern matches visible text; no match when pattern appears only inside escape sequences).

## Verification
- `mainProj/testOnly sbt.internal.OutputSpec` — both tests pass.
- Manually: run a task that prints colored output, then `lastGrep <word>`; lines containing that word in the visible text are matched.

### DIFF
--- a/main/src/main/scala/sbt/internal/Output.scala
+++ b/main/src/main/scala/sbt/internal/Output.scala
@@ -19,6 +19,7 @@ import Aggregation.{ KeyValue, Values }
 import Types.idFun
 import Highlight.{ bold, showMatches }
 import annotation.tailrec
+import sbt.internal.util.EscHelpers
 
 import sbt.io.IO
 
@@ -43,7 +44,12 @@ object Output {
       printLines: Seq[String] => Unit
   )(using display: Show[ScopedKey[?]]): Unit = {
     val pattern = Pattern.compile(patternString)
-    val lines = flatLines(lastLines(keys, streams))(_ flatMap showMatches(pattern))
+    val lines = flatLines(lastLines(keys, streams)) { rawLines =>
+      rawLines.flatMap { line =>
+        val stripped = EscHelpers.stripColorsAndMoves(line)
+        showMatches(pattern)(stripped)
+      }
+    }
     printLines(lines)
   }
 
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Nice change. I like these straightforward fixes.

PR #8727, opened 2026-02-10
[2.x] fix: Fixes explicitlySpecified and selectors for testOnly

### LINKED ISSUE #5609: Pass in correct explicitlySpecified and selectors when creating TestDefinitions
Simply something that has never been implemented it seems. For example Scalatest relies on these two fields to implement the tasks run logic so it's a shame they are currently hardcoded.

https://github.com/sbt/sbt/blob/62c6cad487c72b38f67b765b2519b7e1b04560b8/main-actions/src/main/scala/sbt/Tests.scala#L570-L575

This is the first time I take a look at the sbt code but if someone can provide me with some guidance I would be happy to contribute.

Eruis2579:
Fixes https://github.com/sbt/sbt/issues/5609

**Problem**
When the user runs testOnly with an explicit suite name (e.g. testOnly com.example.MySuite), ScalaTest suites annotated with @DoNotDiscover were not run because sbt always passed explicitlySpecified=false to the test framework.

**Solution**
In Tests.processOptions, when the user has specified test filters (orderedFilters.nonEmpty), mark the filtered tests as explicitlySpecified=true with SuiteSelector so frameworks can run @DoNotDiscover suites when explicitly requested.

- main-actions: set explicitlySpecified and selectors in processOptions for testOnly path
- scripted test: tests/i5609-do-not-discover-testonly (testOnly com.test.TestSpec2)

### DIFF
--- a/main-actions/src/main/scala/sbt/Tests.scala
+++ b/main-actions/src/main/scala/sbt/Tests.scala
@@ -281,8 +281,23 @@ object Tests {
       if (orderedFilters.isEmpty) filtered0
       else orderedFilters.flatMap(f => filtered0.filter(d => f(d.name))).toList.distinct
     val uniqueTests = distinctBy(tests)(_.name)
+    // When user ran testOnly with explicit args and the filter restricted the set, mark matching
+    // tests as explicitlySpecified so frameworks (e.g. ScalaTest) can run @DoNotDiscover suites.
+    // Do not mark when orderedFilters is "run all" (e.g. test with no args -> selectedFilter(Nil) -> match all).
+    val filteredToSubset = orderedFilters.nonEmpty && uniqueTests.size < discovered.size
+    val testsToUse =
+      if (!filteredToSubset) uniqueTests
+      else
+        uniqueTests.map(t =>
+          new TestDefinition(
+            t.name,
+            t.fingerprint,
+            explicitlySpecified = true,
+            Array(new SuiteSelector: Selector)
+          )
+        )
     new ProcessedOptions(
-      uniqueTests.toVector,
+      testsToUse.toVector,
       setup.toVector,
       cleanup.toVector,
       testListeners.toVector
@@ -555,7 +570,6 @@ object Tests {
         c.topLevel
       case _ => false
     })
-    // TODO: To pass in correct explicitlySpecified and selectors
     val tests =
       for {
         (df, di) <- discovered
--- a/main-actions/src/main/scala/sbt/Tests.scala
+++ b/main-actions/src/main/scala/sbt/Tests.scala
@@ -298,7 +298,7 @@ object Tests {
         t.name,
         t.fingerprint,
         explicitlySpecified = explicitlyRequestedNames.contains(t.name),
-        Array(new SuiteSelector: Selector)
+        t.selectors
       )
     )
     new ProcessedOptions(
--- a/main-actions/src/main/scala/sbt/Tests.scala
+++ b/main-actions/src/main/scala/sbt/Tests.scala
@@ -255,7 +255,7 @@ object Tests {
     val testListeners = new ListBuffer[TestReportListener]
     val undefinedFrameworks = new ListBuffer[String]
 
-    def isExplicitFqn(s: String): Boolean = !s.contains('*') && !s.contains('?')
+    def isExplicitFqn(s: String): Boolean = !s.contains('*') && !s.contains('?') && !s.contains("...")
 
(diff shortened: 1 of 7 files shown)


Eruis2579:
Hi, @eed3si9n  Can you check my PR, please? Thanks

sbt-app/src/sbt-test/tests/i5609-do-not-discover-testonly/test
```diff
@@ -0,0 +1,5 @@
+# #5609: When explicitly requested via testOnly, @DoNotDiscover suite should run
```
eed3si9n: Could you include

```
> testFull
$ exists target/SuiteStarting-TestSpec
$ exists target/SuiteCompleted-TestSpec
$ absent target/SuiteStarting-TestSpec2
$ absent target/SuiteCompleted-TestSpec2
```

as well please?

sbt-app/src/sbt-test/tests/i5609-do-not-discover-testonly/test
```diff
@@ -0,0 +1,5 @@
...
+> clean
+> testOnly com.test.TestSpec2
+$ exists target/SuiteStarting-TestSpec2
+$ exists target/SuiteCompleted-TestSpec2
```
eed3si9n, line 15: Could you also include

```bash
> clean
> testOnly com.test...
```

main-actions/src/main/scala/sbt/Tests.scala
```diff
@@ -281,8 +281,23 @@ object Tests {
       if (orderedFilters.isEmpty) filtered0
       else orderedFilters.flatMap(f => filtered0.filter(d => f(d.name))).toList.distinct
     val uniqueTests = distinctBy(tests)(_.name)
+    // When user ran testOnly with explicit args and the filter restricted the set, mark matching
+    // tests as explicitlySpecified so frameworks (e.g. ScalaTest) can run @DoNotDiscover suites.
+    // Do not mark when orderedFilters is "run all" (e.g. test with no args -> selectedFilter(Nil) -> match all).
+    val filteredToSubset = orderedFilters.nonEmpty && uniqueTests.size < discovered.size
```
eed3si9n: I don't think this tells us if the filter was a pattern or an actual explicit fully qualified test.

eed3si9n (changes requested):

main-actions/src/main/scala/sbt/Tests.scala
eed3si9n: See https://github.com/sbt/test-interface/blob/22b24334e8ea0274fc9a8fa7df8dd5ecd4f45cd5/src/main/java/sbt/testing/TaskDef.java#L40-L42

Eruis2579:
Hi, @eed3si9n , Please check again if you have time. Enjoy your travel.

eed3si9n (changes requested):
I think it's still missing the test for

```
> clean
> testOnly com.test...
```

scenario where you run `testOnly` but pass in a pattern filter.

Eruis2579:
Hi, @eed3si9n , Please review my changes again.
Thank you for your review.

sbt-app/src/sbt-test/tests/i5609-do-not-discover-testonly/test
```diff
@@ -0,0 +1,19 @@
...
+$ absent target/SuiteStarting-TestSpec2
+$ absent target/SuiteCompleted-TestSpec2
+
+> clean
+> testOnly com.test.TestSpec2
+$ exists target/SuiteStarting-TestSpec2
+$ exists target/SuiteCompleted-TestSpec2
+
+> clean
+> testOnly com.test...
```
eed3si9n, line 21:
```suggestion
> testOnly com.test...
$ exists target/SuiteStarting-TestSpec
$ exists target/SuiteCompleted-TestSpec
$ absent target/SuiteStarting-TestSpec2
$ absent target/SuiteCompleted-TestSpec2
```

I'm guessing that this should work similar to the `testFull` case?

Eruis2579:
Hi, @eed3si9n , I have added your suggest code. When I committed your suggestion, there was error while testing. That is cuz `clean` doesn't remove individual test output files in the scripted test environment. Using explicit `$ delete` commands works. Thank you for your review again.

Eruis2579:
Hi, @eed3si9n , If you have time, please check my updates for your feedback

eed3si9n (approved):
Thanks for the contribution! LGTM

PR #8730, opened 2026-02-11
[2.x] fix: Handle JVM parameters with spaces in dot files

### LINKED ISSUE #7333: sbtopts does not recognize JVM parameters with spaces
## steps

.sbtopts file:
```
-J--enable-preview
-J--add-modules jdk.incubator.concurrent
```

## problem

```
$ sbt
Error: --add-modules requires modules to be specified
copying runtime jar...
mkdir: cannot create directory ‘’: No such file or directory
Error: --add-modules requires modules to be specified
Error: --add-modules requires modules to be specified
```

## expectation

Succesful startup

## notes

Also tried:
```
"-J--add-modules jdk.incubator.concurrent"
-J"--add-modules jdk.incubator.concurrent"
```

Eruis2579:
Fixes https://github.com/sbt/sbt/issues/7333

### Problem

The sbt launcher script used naive word splitting when parsing `.sbtopts` and `.jvmopts`, so arguments with spaces were split incorrectly. For example, `-J--add-modules jdk.incubator.concurrent` in `.sbtopts` and `-Dtest.key="value with spaces"` in `.jvmopts` were not passed to the JVM as intended.

### Solution

Parsing was updated to be quote-aware. A new `parseLineIntoWords()` helper splits lines while respecting single and double quotes. `loadConfigFileIntoArray()` loads config files into bash arrays: for `-J` lines it splits the remainder and prepends `-J` to each token (e.g. `-J--add-modules jdk.incubator.concurrent` → `-J--add-modules` and `-Jjdk.incubator.concurrent`); for other lines it uses quote-aware splitting. `.sbtopts` now uses `loadConfigFileIntoArray` instead of `loadConfigFile`; `.jvmopts` uses `loadConfigFileIntoArray` and merges the result into `java_args` instead of appending via `JAVA_OPTS` string concatenation.

### Testing

Two integration tests were added in `RunnerScriptTest.scala`: one for `-J--add-modules jdk.incubator.concurrent` in `.sbtopts`, and one for `-Dtest.7333="value with spaces"` in `.jvmopts`, both verifying that the full values are passed correctly to the JVM.

### DIFF
--- a/sbt
+++ b/sbt
@@ -766,6 +766,61 @@ process_args () {
   }
 }
 
+# Parse a line into words respecting double and single quotes.
+# Outputs one word per line. Used for .sbtopts and .jvmopts to handle args with spaces (#7333).
+parseLineIntoWords() {
+  local line="$1"
+  local word=""
+  local i=0
+  local len=${#line}
+  local in_dq=0 in_sq=0
+  while (( i < len )); do
+    local c="${line:$i:1}"
+    if (( in_dq )); then
+      word+="$c"
+      [[ "$c" == '"' ]] && in_dq=0
+    elif (( in_sq )); then
+      word+="$c"
+      [[ "$c" == "'" ]] && in_sq=0
+    else
+      case "$c" in
+        '"')  in_dq=1; word+="$c" ;;
+        "'")  in_sq=1; word+="$c" ;;
+        ' '|$'\t')
+          [[ -n "$word" ]] && printf '%s\n' "$word"
+          word=""
+          ;;
+        *)    word+="$c" ;;
+      esac
+    fi
+    ((i++))
+  done
+  [[ -n "$word" ]] && printf '%s\n' "$word"
+}
+
+# Load config file into array, parsing each line and respecting quotes.
+# For -J lines: split the remainder and prepend -J to each token (so -J--add-modules jdk.incubator.concurrent
+# becomes -J--add-modules and -Jjdk.incubator.concurrent). Fixes #7333.
+loadConfigFileIntoArray() {
+  local -n arr=$1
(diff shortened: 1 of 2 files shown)


sbt
```diff
@@ -766,6 +766,62 @@ process_args () {
...
+  local arr_name="$1"
+  local file="$2"
+  [[ ! -f "$file" ]] && return
+  while IFS= read -r line || [[ -n "$line" ]]; do
+    line=$(printf '%s' "$line" | sed $'/^\#/d;s/\r$//')
+    [[ -z "$line" ]] && continue
+    if [[ "$line" == -J* ]]; then
+      local rest="${line#-J}"
+      while IFS= read -r token; do
+        [[ -n "$token" ]] && eval "$arr_name+=($(printf '%q' "-J$token"))"
```
eed3si9n: I am not a security expert, but `eval` feels like we're opening ourselves up for vulnerabilities.

eed3si9n (changes requested):

sbt
Eruis2579: Hi, @eed3si9n , I hope you’re having a great trip. Thanks for flagging this. You're absolutely right that using eval on user-controlled content from `.sbtopts` / `.jvmopts` is risky, even with `printf '%q'`.
I’ve changed the implementation to avoid eval entirely.
Please check again.

Eruis2579:
Hi, @eed3si9n , Please check my updates if you have time. Thank you.

eed3si9n (approved):
Thanks!

Eruis2579:
Hi, @eed3si9n , Thank you for your review. Could you merge this PR?

PR #8733, opened 2026-02-12
[2.x] Print warning about scala-reflect not found

eed3si9n:
## Problem
scala-reflect not found issue is confusing.

## Solution
Print out a specialized warning message. I'm going to use <https://github.com/sbt/sbt/discussions/8728> as a placeholder until we have some more official page from the Scala team.

## Test

```
sbt:smorrebrod-root> compile
[info] Updating projectc_2.13
[info] Resolved projectc_2.13 dependencies
[warn]
[warn] 	Note: Unresolved dependencies path:
[warn] [sbt-8728] Smorrebrod - the end of Scala 2.13-3.x sandwich
[warn] Scala 3.8+ cannot be used in a Scala 2.13 subproject.
[warn] Dependency resolution failed because scala-reflect or -compiler 3.x does not exist.
[warn] This happens when a Scala 2.13 subproject depends on Scala 3.8+ directly or transitively.
[warn] To fix this, either
[warn]   - Keep Scala 3 subproject or transitive dependency to 3.7 or below, or
[warn]   - Migrate the Scala 2.13 subproject to Scala 3.x
[warn] See https://github.com/sbt/sbt/discussions/8728
[error] stack trace is suppressed; run last projectC / update for the full output
[error] (projectC / update) sbt.librarymanagement.ResolveException: Error downloading org.scala-lang:scala-reflect:3.8.1
[error]   Not found
[error]   Not found
[error]   not found: /Users/xxx/.ivy2/local/org.scala-lang/scala-reflect/3.8.1/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/org/scala-lang/scala-reflect/3.8.1/scala-reflect-3.8.1.pom
[error] elapsed time: 1 s, cache 100%, 26 disk cache hits
```

### DIFF
--- a/main/src/main/scala/sbt/internal/LibraryManagement.scala
+++ b/main/src/main/scala/sbt/internal/LibraryManagement.scala
@@ -172,6 +172,11 @@ private[sbt] object LibraryManagement {
             .apply(updateInputs)
           if isCached then markAsCached(report) else report
         catch
+          case r: ResolveException
+              if r.failed.exists(isMissingScalaLib) &&
+                module.scalaModuleInfo.exists(_.scalaBinaryVersion == "2.13") =>
+            informSandwich()
+            throw r
           case t: (NullPointerException | OutOfMemoryError) =>
             val resolvedAgain = resolve
             val culprit = t.getClass.getSimpleName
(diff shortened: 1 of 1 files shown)


scutuatua-crypto (approved):

main/src/main/scala/sbt/internal/LibraryManagement.scala
```diff
@@ -185,6 +188,20 @@ private[sbt] object LibraryManagement {
...
     }
 
+    def informSandwich(): Unit =
+      log.warn("[sbt-8728] Smorrebrod - the end of Scala 2.13-3.x sandwich")
+      log.warn("Dependency resolution failed because scala-reflect or -compiler 3.x was not found.")
+      log.warn(
+        "This likely indicates that Scala 3.8 or above was found in a 2.13 dependency graph;"
+      )
+      log.warn("however, 2.13 TASTy Reader does not support Scala 3.8 or above.")
+      log.warn("See https://github.com/sbt/sbt/discussions/8728")
```
anatoliykmetyuk, line 206: Maybe a more clear message? E.g. "Smorrebrod" or "sandwich" are internal lingo, and "TASTy Reader" issue is not an immediate, addressable cause. How about something like:

```
Scala 2.13 and Scala 3.8+ cannot be used together in the same build.

Dependency resolution failed because scala-reflect:3.8.x does not exist.
This happens when a Scala 2.13 project depends on a Scala 3.8+ library.
Scala 3.0–3.7 shared the 2.13 standard library, but Scala 3.8
introduced its own scala-library, and the 2.13 TASTy Reader does not support Scala 3.8+.

To fix this, either:
  - Keep Scala 3 subprojects on 3.7.x or below, or
  - Migrate the Scala 2.13 subprojects to Scala 3.

See https://github.com/sbt/sbt/discussions/8728
```

eed3si9n: I'd be happy to update the body text, but I wanted to make the title unique so it can be Googled.

anatoliykmetyuk: Ok, makes sense.

eed3si9n: Updated the blurb in https://github.com/sbt/sbt/pull/8733/changes/5efac50be456b508dcb8fbe8612793940f83fa00

PR #8744, opened 2026-02-15
[2.x] fix: Fixes dependency lock file url

### LINKED ISSUE #8741: [2.x] dependencyLock captures file URL
## steps

```scala
scalaVersion := "2.13.16"
```

Using a locally published snapshot run `dependencyLock`

## problem

1. The artifact URL captures my Coursier cache directory:

```json
{
  "version": "1.0",
  "buildClock": "725bc176e1f279828c9a0be08578993d56a5bd5f5d1c06d4eab8ee29471cc998",
  "configurations": [
    {
      "name": "compile",
      "dependencies": [
        {
          "organization": "org.scala-lang",
          "name": "scala-library",
          "version": "2.13.16",
          "configuration": "compile",
          "tpe": "jar",
          "transitives": [],
          "artifacts": [
            {
              "url": "file:/Users/xxxx/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.13.16/scala-library-2.13.16.jar",
              "extension": "jar",
              "tpe": "jar"
            }
          ]
        }
      ]
    },
   .....
```

## expectation

1. Ideally, it should contain the original URL to Maven Central
2. Whenever it tries to capture the `file://` URL, for example because of some local repository, it should just fail to create a lock file

bitloi:
Fixes #8741

When using Coursier for dependency resolution, the dependency lock file was sometimes writing absolute `file://` URLs pointing at the local cache (e.g. `file:///home/runner/.cache/coursier/...`). That made the lock file machine-specific and broke portability.

This PR normalizes those cache URLs so the lock file stays portable. File URLs from the Coursier cache are converted to a stable form (e.g. `${CSR_CACHE}`) before being written to the lock file, and converted back when reading. Path handling is normalized so Windows and Unix both behave correctly.

---

**Also in this PR:** Action cache was occasionally failing in CI with `IncompleteParseException: exhausted input` when reading cached task values (e.g. in the dotty-sandwich scripted test). We now treat invalid or empty cached JSON as a cache miss and re-run the task instead of failing.

### DIFF
--- a/main/src/main/scala/sbt/internal/librarymanagement/DependencyLockManager.scala
+++ b/main/src/main/scala/sbt/internal/librarymanagement/DependencyLockManager.scala
@@ -44,7 +44,11 @@ object DependencyLockManager:
       isValid
     }
 
-  private def artifactUrlForLock(rawUrl: String, cacheDir: Option[File], moduleDesc: String): String =
+  private def artifactUrlForLock(
+      rawUrl: String,
+      cacheDir: Option[File],
+      moduleDesc: String
+  ): String =
     if !rawUrl.startsWith(CacheUrlConversion.FileUrlPrefix) then rawUrl
     else
       cacheDir match
--- a/main/src/main/scala/sbt/internal/librarymanagement/DependencyLockManager.scala
+++ b/main/src/main/scala/sbt/internal/librarymanagement/DependencyLockManager.scala
@@ -44,19 +44,40 @@ object DependencyLockManager:
       isValid
     }
 
+  private def artifactUrlForLock(rawUrl: String, cacheDir: Option[File], moduleDesc: String): String =
+    if !rawUrl.startsWith(CacheUrlConversion.FileUrlPrefix) then rawUrl
+    else
+      cacheDir match
+        case None =>
+          throw new RuntimeException(
+            s"Cannot create dependency lock file: artifact has file URL (e.g. local cache path). " +
+              s"Lock files must use portable repository URLs. Module: $moduleDesc. " +
+              s"Run 'update' first or ensure dependencies are resolved from remote repositories."
+          )
+        case Some(dir) =>
+          val converted = CacheUrlConversion.cacheFileToOriginalUrl(rawUrl, dir)
+          if !CacheUrlConversion.isPortableUrl(converted) then
+            throw new RuntimeException(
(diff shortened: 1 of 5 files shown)


bitloi:
@eed3si9n Would you mind reviewing the PR?

util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -149,6 +149,16 @@ object ActionCache:
...
 
+    def parseCachedValue(
+        str: String,
+        origin: Option[String],
+        isFailure: Boolean,
+    ): Option[Either[Option[CachedCompileFailure], O]] =
+      try
+        if isFailure then Some(Left(Some(failureFromStr(str))))
+        else Some(Right(valueFromStr(str, origin)))
+      catch case _: Exception => None
```
eed3si9n, line 160: 1. Could you split this commit into another PR please? I've opened an issue https://github.com/sbt/sbt/issues/8745
2. Rather than catching `Exception` could you use `NonFatal(_)` please?

eed3si9n (approved):
Thanks!

util-cache/src/main/scala/sbt/util/ActionCache.scala
eed3si9n: Nah, never mind. I'll just merge it.

main/src/main/scala/sbt/internal/librarymanagement/DependencyLockManager.scala
```diff
@@ -44,19 +44,44 @@ object DependencyLockManager:
...
+              s"Lock files must use portable repository URLs. Module: $moduleDesc. " +
+              s"Run 'update' first or ensure dependencies are resolved from remote repositories."
+          )
+        case Some(dir) =>
+          val converted = CacheUrlConversion.cacheFileToOriginalUrl(rawUrl, dir)
+          if !CacheUrlConversion.isPortableUrl(converted) then
+            throw new RuntimeException(
+              s"Cannot create dependency lock file: artifact path is not under Coursier cache. " +
+                s"Module: $moduleDesc. URL: $rawUrl"
+            )
```
xuwei-k, line 67: - https://github.com/sbt/sbt/issues/9184

Issue #2050, opened 2015-06-09
Provide method to transform keys to scoped keys


jroper:
The code to transform the key to a scoped key in #2006 made @jsuereth break down in tears in front of everyone at ScalaDays. I'm not sure what the issue with that is, it sounds more like a good feature to me... but he told me to raise this as an issue anyway.

jsuereth:
To clarify:

When sbt loads settings from the .sbt file, it will transform the scope, based on the location of the sbt file:
- It will automatically place the key reference into the "root" project of the build file, if no other project is specified.
- For the `set` command, this is the `currentProject`, not the root project.
- It will also fix the references (ThisBuild, ThisProject) to be on physical projects.

If you don't manipulate the scopes on "raw" user entered keys, then you'll find it frustrating that sbt isn't finding somethign you think needs the magic.  If you need to duplicate this magic based on sbt.State, it's actually quite hard to figure out htis is happening,and there are a lot of poorly named and undocumented methods which APPEAR to solve the problem, but do not.

PR #8751, opened 2026-02-17
[2.x] fix: Relax non-delegation for settings on shell

### LINKED ISSUE #8757: [2.x] Relax non-delegation for settings
## steps

```bash
sbt:aaa> Compile/console/fork
[error] No such setting/task
[error] Compile/console/fork
[error]                     ^
```

## problem

sbt 2.0.0-RC9 no longer delegates settings coming from the shell.

## expectation

Keep this behavior for tasks, but let settings delegate so `Compile/console/fork` would work.

## note

https://github.com/sbt/sbt/pull/8539

bitloi:
Fixes #8757 

## Problem

After the change in #8539, the shell stopped delegating for any scoped key. So when you run something like `Compile/console/fork`, you get "No such setting/task" even when `console/fork` is defined in a delegated scope (e.g. the default config). That’s annoying for settings, where delegation is usually what you want.

## Solution

We keep the current behaviour for **tasks**: if you type a task with an explicit scope and it isn’t defined in that exact scope, it still fails (so typos like `Compile/update` don’t silently run something else). For **settings** we allow delegation again: if the key isn’t found in the requested scope, we follow the normal scope delegation and use the value from a delegate scope. So `Compile/console/fork` works when `console/fork` is defined in a more general scope.

The change is in `Act.getValue`: when the user has specified an explicit config or task scope we still try a direct lookup first; if that returns nothing and the key is a setting (via `key.key.tag.isSetting`), we fall back to `data.get(key)` so delegation applies. Task keys never use that fallback.

### DIFF
--- a/main/src/main/scala/sbt/internal/Act.scala
+++ b/main/src/main/scala/sbt/internal/Act.scala
@@ -600,14 +600,21 @@ object Act {
     keys.flatMap(key => getValue(structure.data, key).map(KeyValue(key, _)))
 
   /**
-   * Starting sbt 2.0.0, we will not delegate when a non-existent scoping
-   * such as Compile / update is required.
+   * Resolves a key from settings data. Starting sbt 2.0.0 we do not delegate for
+   * tasks when the user specified an explicit scope (config or task axis), so that
+   * non-existent scopes like `Compile / update` fail. Settings may still delegate
+   * so that e.g. `Compile/console/fork` resolves when `console/fork` is defined
+   * in a delegated scope (see #8757).
    */
   private def getValue[T](data: Def.Settings, key: ScopedKey[T]): Option[T] =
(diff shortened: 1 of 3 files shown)


eed3si9n (changes requested):
afaik https://github.com/sbt/sbt/issues/2050 is already a done issue.

eed3si9n (changes requested):
I don't think this feature is needed in sbt 2.x since we already implement caching for tasks by default.

eed3si9n (commented):
Could you add a scripted test for this please?

bitloi:
> Could you add a scripted test for this please?

I added it. Would you mind reviewing it again?

eed3si9n (approved):
Thanks for the contribution!

PR #8752, opened 2026-02-17
[2.x] feat: client-side run env inheritance

### LINKED ISSUE #7012: sbtn should reevaluate env variables
## steps

Given a hello world application that reads out an environment variable and prints it out:

```scala
@main def Main(args: String*): Unit =
  println(System.getenv("FOO"))
```

Then run:

```bash
FOO=BAR sbtn run
```

## problems

I see `null`.

## expectations

I see `BAR`

## notes

When I run:

```bash
FOO=BAR bloop run my-project
```

I correctly see `BAR`

In fact the sbtn behavior is flaky. It depends on how the server was started. If the server was not running to begin with, then yes

```bash
FOO=BAR sbtn run
```

will indeed print out `BAR` and the subsequent requests as well. However as soon as we change it to

```bash
FOO=NOT_BAR sbtn run
```
we will keep seeing the incorrect `BAR`.

## original report
I'm not sure whether it's an issue or a feature request. I already created a feature request here: https://discuss.lightbend.com/t/feature-request-sbtn-should-reevaluate-env-variables/9908 but haven't heard anything in over 2 weeks so I'm opening an issue here as well.

aviu16:
Fixes #7012.

When `sbtn run` is executed through client jobs, the run env currently only uses explicit task env vars. This can keep stale process-level values around across invocations.

This change merges the current client process environment (`sys.env`) with explicit overrides, so each invocation reflects the shell env it was started with.

Changes:
- merge `sys.env` with job env overrides in `RunHandler.jvmRun`
- apply the same merge for native client-side runs in `NetworkClient`
- add a focused unit test for env merge behavior and override precedence

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -765,7 +765,7 @@ class NetworkClient(
         workingDirectory = info.workingDirectory.map(new File(_)),
         runJVMOptions = Vector.empty,
         connectInput = info.connectInput,
-        envVars = info.environmentVariables,
+        envVars = RunHandler.mergedEnvVars(info.environmentVariables),
       )
       val command = info.cmd :: info.args.toList
       val jpb = new JProcessBuilder(command*)
--- a/main-command/src/main/scala/sbt/internal/util/RunHandler.scala
+++ b/main-command/src/main/scala/sbt/internal/util/RunHandler.scala
@@ -18,6 +18,11 @@ import scala.util.Try
 
 // Process runinfos
 object RunHandler:
+  private[internal] def mergedEnvVars(
+      environmentVariables: Map[String, String]
+  ): Map[String, String] =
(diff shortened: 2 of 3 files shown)


aviu16:
added validation evidence for this branch:\n\nran in JDK 17 container:\ndocker run --rm -v "/Users/Avantika/Documents/New project":/work -w /work eclipse-temurin:17-jdk bash -lc './sbt ";project commandProj;testOnly sbt.internal.util.RunHandlerTest"'\n\nresult:\n- RunHandlerTest passed\n- test: mergedEnvVars includes current process environment and applies explicit overrides

aviu16:
Scala CLA has been signed for aviu16 and I pushed a no-op commit to retrigger checks.\n\nCurrent new runs are in action_required state, likely awaiting maintainer approval to run workflows for fork PRs.

eed3si9n (approved):
Thanks!

agilesteel:
I've been waiting a long time for this ❤️ Can't wait to test it out...

PR #8756, opened 2026-02-17
[2.x] Excluding artifacts without Scala version suffix

### LINKED ISSUE #7454: Cross-versioned excludes
Does this really work with sbt 1.9.7? For me it seems not...

I have a dependency:
```sbt
val atomRenderer = ("com.gu" %% "atom-renderer" % "1.2.0")
```
Now I want to exclude a dependency and it only works when appending `_2.13` to it, everything else fails:
```sbt
.exclude("com.typesafe.play", "twirl-api_2.13") // works

.exclude("com.typesafe.play", "twirl-api") // does NOT work

.excludeAll("com.typesafe.play" %% "twirl-api") // does NOT work
```

_Originally posted by @mkurz in https://github.com/sbt/librarymanagement/issues/88#issuecomment-1851554369_

seroperson:
Now it's possible to use:

```
.exclude("com.typesafe.play" %% "twirl-api")
```

Instead of:

```
.exclude("com.typesafe.play", "twirl-api_2.13")
```

It didn't work because `ExclusionRule` constructed by `organization` and `name` never had `crossVersion` set. 

Also: 

- Fixed handling `.excludeAll("com.typesafe.play" %% "twirl-api")`: previously it wasn't excluding artifacts in published xmls.
- Marked `.exclude(String, String)` as deprecated in favor of new syntax `.exclude(... %% ...)`.

Closes #7454

### DIFF
--- a/lm-core/src/main/scala/sbt/librarymanagement/ModuleIDExtra.scala
+++ b/lm-core/src/main/scala/sbt/librarymanagement/ModuleIDExtra.scala
@@ -9,6 +9,7 @@ import sbt.internal.librarymanagement.mavenint.SbtPomExtraProperties
 import scala.collection.mutable.ListBuffer
 import sbt.librarymanagement.syntax.*
 import sbt.util.Logger
+import sbt.librarymanagement.DependencyBuilders.OrganizationArtifactName
 
 private[librarymanagement] abstract class ModuleIDExtra {
   def organization: String
@@ -152,8 +153,17 @@ private[librarymanagement] abstract class ModuleIDExtra {
   def excludeAll(rules: ExclusionRule*): ModuleID = withExclusions(exclusions ++ rules)
 
   /** Excludes the dependency with organization `org` and `name` from being introduced by this dependency during resolution. */
+  @deprecated(
+    "`name` parameter must contain scala version if artifact is cross-versioned. Use `exclude(\"org\" %% \"name\")`.",
+    since = "2.0.0"
+  )
   def exclude(org: String, name: String): ModuleID =
     excludeAll(ExclusionRule().withOrganization(org).withName(name))
+
+  /** Excludes the dependency from being introduced by this dependency during resolution. */
+  def exclude(rule: OrganizationArtifactName): ModuleID = {
+    excludeAll(rule)
+  }
 
   /**
    * Adds extra attributes for this module.  All keys are prefixed with `e:` if they are not already so prefixed.
--- a/lm-coursier/src/main/scala/lmcoursier/FromSbt.scala
+++ b/lm-coursier/src/main/scala/lmcoursier/FromSbt.scala
@@ -19,19 +19,21 @@ import sbt.librarymanagement.{ Configuration as _, * }
 
 object FromSbt {
 
-  private def sbtModuleIdName(
-      moduleId: ModuleID,
+  private def sbtCrossName(
+      name: String,
+      crossVersion: CrossVersion,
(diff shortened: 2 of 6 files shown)


lm-core/src/main/scala/sbt/librarymanagement/ModuleIDExtra.scala
```diff
@@ -152,8 +153,25 @@ private[librarymanagement] abstract class ModuleIDExtra {
...
+
+    exclude(
+      ExclusionRule().withOrganization(org).withName(name).withCrossVersion(exclusionCrossVersion)
+    )
+  }
+
+  /** Excludes the dependency from being introduced by this dependency during resolution. */
+  def exclude(rule: ExclusionRule): ModuleID = {
+    excludeAll(rule)
+  }
```
eed3si9n: Is this safe for libraries that needs to be published to a Maven repo? https://www.scala-sbt.org/1.x/docs/Library-Management.html#Exclude+Transitive+Dependencies reads

> The `excludeAll` method is more flexible, but because it cannot be represented in a `pom.xml`, it should only be used when a pom doesn’t need to be generated.

seroperson: Oh, yea, well, it's not. I changed this signature to be `exclude(OrganizationArtifactName)` to still be able to write the usual `exclude("..." %% "...")` but not to be able to set configurations and artifact names.

eed3si9n (changes requested):
Could you add scripted test that either checks makePom or publishes to a local directory and use it please?

seroperson:
> Could you add scripted test that either checks makePom or publishes to a local directory and use it please?

Well, yes, it was tricky to find, but there was one more place which ignored `ExclusionRule.crossVersion`.

seroperson:
@eed3si9n would you mind review one more time?

eed3si9n:
Sure. I'll try to grab this branch locally over the weekend and test it.

seroperson:
@eed3si9n friendly reminder 🙂

lm-core/src/main/scala/sbt/librarymanagement/ModuleIDExtra.scala
```diff
@@ -152,8 +154,25 @@ private[librarymanagement] abstract class ModuleIDExtra {
...
+    //
+    // Otherwise, inherit the parent module's cross-version behavior.
+    val exclusionCrossVersion =
+      if (CrossVersionUtil.endsWithCrossVersion(name, crossVersion)) Disabled()
+      else crossVersion
+
+    excludeAll(
+      ExclusionRule().withOrganization(org).withName(name).withCrossVersion(exclusionCrossVersion)
+    )
+  }
```
eed3si9n: If `exclude(org, name)` is already an alias for `excludeAll(...)`, I suggest we don't change the implementation, but instead deprecate this method in favor of `excludeAll(org %% name)` or the new variant `exclude(org %% name)`.

eed3si9n (changes requested):

lm-core/src/main/scala/sbt/librarymanagement/ModuleIDExtra.scala
seroperson: Done. I've removed all the backward-compatibility logic then.

eed3si9n (approved):
Thanks! Could you update the PR description as well plz?

seroperson:
@eed3si9n done 🎉

PR #8760, opened 2026-02-17
[2.x] fix: Filter semanticdb options from scaladoc invocations

### LINKED ISSUE #8740: [2.x] scaladoc failure when semanticdbEnabled is enabled on scala 2.x projects
## steps

```
➜ cat project/build.properties
sbt.version=2.0.0-RC8
```
```
➜ cat build.sbt
semanticdbEnabled := true
scalaVersion := "2.12.21"
```
```
➜ cat src/main/scala/A.scala
object A
```

## problem

Scaladoc failure

```
sbt:test> doc
[error] bad option: -P:semanticdb:targetroot:/private/tmp/test/target/out/jvm/scala-2.12.21/test/meta
[error] one error found doc 0s
[error] stack trace is suppressed; run last Compile / doc for the full output
[error] (Compile / doc) Scaladoc generation failed
[error] elapsed time: 1 s, cache 7%, 1 disk cache hit, 13 onsite tasks
```
```
sbt:test> last doc
[debug] Getting org.scala-sbt:compiler-bridge_2.12:2.0.0-M11:compile for Scala 2.12.21
[debug] Calling Scaladoc with arguments:
[debug] 	-d
[debug] 	/private/tmp/test/target/out/jvm/scala-2.12.21/test/api
[debug] 	-Xplugin:${CSR_CACHE}/https/repo1.maven.org/maven2/org/scalameta/semanticdb-scalac_2.12.21/4.14.2/semanticdb-scalac_2.12.21-4.14.2.jar
[debug] 	-Yrangepos
[debug] 	-P:semanticdb:targetroot:/private/tmp/test/target/out/jvm/scala-2.12.21/test/meta
[debug] 	-bootclasspath
[debug] 	/Users/brice.jaglin/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.12.21/scala-library-2.12.21.jar
[debug] 	-classpath
[debug] 	dummy_8e2dff45
[debug] 	/private/tmp/test/src/main/scala/A.scala
[error] bad option: -P:semanticdb:targetroot:/private/tmp/test/target/out/jvm/scala-2.12.21/test/meta
[error] one error found
[error] Scaladoc generation failed
[error] 	at xsbt.Runner.run(ScaladocBridge.scala:50)
[error] 	at xsbt.ScaladocBridge.run(ScaladocBridge.scala:21)
[error] 	at sbt.internal.inc.AnalyzingCompiler.doc(AnalyzingCompiler.scala:154)
[error] 	at sbt.internal.inc.AnalyzingCompiler.doc(AnalyzingCompiler.scala:133)
[error] 	at sbt.Defaults$.docTaskSettings$$anonfun$5(Defaults.scala:1997)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:79)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:79)
[error] 	at sbt.std.Transform$$anon$3.work(Transform.scala:79)
[error] 	at sbt.std.Transform$$anon$3.work(Transform.scala:79)
[error] 	at sbt.Execute.submit$$anonfun$1$$anonfun$1(Execute.scala:284)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:24)
[error] 	at sbt.Execute.work(Execute.scala:295)
[error] 	at sbt.Execute.submit$$anonfun$1(Execute.scala:284)
[error] 	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$2(ConcurrentRestrictions.scala:275)
[error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:75)
[error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:73)
[error] 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
[error] 	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
[error] 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
[error] 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
[error] 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
[error] 	at java.base/java.lang.Thread.run(Thread.java:1474)
[error] (Compile / doc) Scaladoc generation failed
```

## expectation

No errror, like in sbt 1.x

```
➜ cat project/build.properties
sbt.version=1.12.2
```

```
sbt:test> doc
[info] Main Scala API documentation to /private/tmp/test/target/scala-2.12/api...
model contains 2 documentable templates
[success] Total time: 1 s, completed 14 Feb 2026, 19:19:25
```

## notes

The error points at the `-P:semanticdb` flag but I think the real problem is that the plugin that is not properly loaded because `CSR_CACHE`  is not expanded when calling Scaladoc, and thus associated flags are rejected.

Judging by the logs,  when it works, `CSR_CACHE` does get expanded, for example with
- `doc` on sbt 1.x
- `compile` on any sbt version

0xRozier:
## Problem

When `semanticdbEnabled := true` is set on Scala 2.x projects, the `doc` task fails because semanticdb-related compiler options (`-Xplugin:...semanticdb...`, `-P:semanticdb:targetroot:*`, `-Yrangepos`) leak into `doc / scalacOptions`. Scaladoc does not support these options, causing:

```
[error] bad option: -P:semanticdb:targetroot:/path/to/target/meta
[error] (Compile / doc) Scaladoc generation failed
```

Fixes #8740.

## Solution

Added a `doc / scalacOptions` setting in `SemanticdbPlugin.configurationSettings` that, when `semanticdbEnabled` is true, filters out:

1. All options from `semanticdbOptions` (`-Yrangepos`, `-P:semanticdb:targetroot:*` for Scala 2.x; `-Xsemanticdb`, `-semanticdb-target` for Scala 3.x)
2. Any `-Xplugin:` flag containing `semanticdb` (the compiler plugin JAR)

This is placed in `SemanticdbPlugin` because it introduces these options, so it should be responsible for scoping them away from `doc`.

## Tests

New scripted test `project/semanticdb-doc` that:
- Runs `doc` on both Scala 2.12 and Scala 3.6 with `semanticdbEnabled := true`
- Verifies that `compile` still generates `.semanticdb` files

## AI Disclosure

This PR was created with assistance from Claude Opus 4.6 (AI). All code changes have been reviewed and verified for correctness.

Generated-by: Claude Opus 4.6

### DIFF
--- a/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala
+++ b/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala
@@ -94,7 +94,7 @@ object SemanticdbPlugin extends AutoPlugin {
         val sdbOpts = semanticdbOptions.value.toSet
         (doc / scalacOptions).value.filterNot { opt =>
           sdbOpts.contains(opt) ||
-            (opt.startsWith("-Xplugin:") && opt.contains("semanticdb"))
+          (opt.startsWith("-Xplugin:") && opt.contains("semanticdb"))
         }
       } else (doc / scalacOptions).value
     },
--- a/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala
+++ b/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala
@@ -89,6 +89,15 @@ object SemanticdbPlugin extends AutoPlugin {
           orig
         }
     }).value,
+    doc / scalacOptions := {
+      if (semanticdbEnabled.value) {
+        val sdbOpts = semanticdbOptions.value.toSet
+        (doc / scalacOptions).value.filterNot { opt =>
+          sdbOpts.contains(opt) ||
+            (opt.startsWith("-Xplugin:") && opt.contains("semanticdb"))
+        }
+      } else (doc / scalacOptions).value
+    },
   )
 
   def targetRootOptions(scalaVersion: String, targetRoot: File): Seq[String] = {
--- a/sbt-app/src/sbt-test/project/semanticdb-doc/build.sbt
+++ b/sbt-app/src/sbt-test/project/semanticdb-doc/build.sbt
@@ -1,0 +1,10 @@
+semanticdbEnabled := true
+
+val matrix = projectMatrix
+  .defaultAxes(VirtualAxis.jvm)
+  .jvmPlatform(scalaVersions =
+    Seq(
+      "2.12.21",
+      "3.6.4"
+    )
+  )
--- a/sbt-app/src/sbt-test/project/semanticdb-doc/matrix/src/main/scala/foo/A.scala
+++ b/sbt-app/src/sbt-test/project/semanticdb-doc/matrix/src/main/scala/foo/A.scala
@@ -1,0 +1,7 @@
+package foo
+
+object A {
+  def main(args: Array[String]): Unit = {
+    println("hello world")
+  }
+}
--- a/sbt-app/src/sbt-test/project/semanticdb-doc/test
+++ b/sbt-app/src/sbt-test/project/semanticdb-doc/test
@@ -1,0 +1,13 @@
+# Verify that scaladoc works when semanticdbEnabled is true
+# This tests the fix for issue #8740
+
+# Scala 2.x: doc should succeed without semanticdb options leaking
+> matrix2_12/doc
+
+# Scala 3.x: doc should succeed without semanticdb options leaking
+> matrix3/doc
+
+# Verify that compile still generates semanticdb files
+> compile
+$ exists target/**/scala-2.12*/**/meta/META-INF/semanticdb/matrix/src/main/scala/foo/A.scala.semanticdb
+$ exists target/**/scala-3*/**/meta/META-INF/semanticdb/matrix/src/main/scala/foo/A.scala.semanticdb


eed3si9n (changes requested):
Thanks. While this approach might work specifically for semanticdb plugin my guess is that it would still be broken for all other compiler plugins until we fix the root cause, which should be fixed in Zinc AnalyzingCompiler similar to https://github.com/sbt/zinc/pull/1545.

0xRozier:
Thanks for the feedback @eed3si9n! You're right that the fix should be in Zinc to handle all compiler plugins generically, not just semanticdb.

I've created a PR against Zinc that filters compiler plugin options (`-Xplugin:`, `-Xplugin-require:`, `-P:`) in `AnalyzingCompiler.doc()` before passing them to the Scaladoc bridge: https://github.com/sbt/zinc/pull/1640

I've also reverted the semanticdb-specific filtering from `SemanticdbPlugin` in this branch, since the fix now belongs in Zinc. The scripted test (`semanticdb-doc`) is kept as it will validate the fix once the new Zinc version is picked up.

0xRozier:
Update: I've restored the semanticdb-specific filtering in `SemanticdbPlugin` as a short-term workaround while the generic Zinc fix (sbt/zinc#1640) goes through review.

Once the Zinc fix is merged and the dependency is bumped in sbt, this workaround can be removed since `AnalyzingCompiler.doc()` will filter all compiler plugin options generically.

PR #8761, opened 2026-02-18
[2.x] feat: repositories_force support

### LINKED ISSUE #1870: Some way of forcing proxy repositories
Suggested by @alexy (I think) at nescala. 
## steps

Currently [proxy repository](http://www.scala-sbt.org/0.13/docs/Proxy-Repositories.html) requires `-Dsbt.override.build.repos=true`.
## problem

One needs to remember to call sbt with the flag.
## expectation

Provide some way of always using `repositories`.
- would environment variable work?
- some other file name such as `~/.sbt/repositories_force`?
## notes

Please comment on this issue if someone wants to send PR for this

bitloi:
## Summary

Adds support for forcing dependency resolution through proxy repositories by placing a `repositories_force` file under the sbt global base (e.g. `~/.sbt/repositories_force`). When this file exists, sbt sets `overrideBuildResolvers := true` so resolution uses the proxy repos (e.g. from sbtopts) instead of the build’s resolvers.

Useful in locked-down or corporate environments where all artifact access must go through a proxy.

**Changes**
- Read `repositories_force` under `sbt.global.base` (or `$SBT_OPTS` / sbtopts) and enable override when present.
- Add scripted test `dependency-management/force-proxy-repos-file` to cover the behavior.

Closes #1870


eed3si9n (changes requested):
I think this needs to be implemented in sbt/launcher because the resolution of sbt artifact itself needs to be overridden if you're behind the firewall - https://github.com/search?q=repo%3Asbt%2Flauncher%20overrideRepos&type=code

bitloi:
@eed3si9n Would you mind reviewing the changes again?

eed3si9n:
I just released launcher 1.6.0 with your change. I think you just need to update the launcher and update https://github.com/sbt/sbt/blob/develop/launch/src/main/input_resources/sbt/sbt.boot.properties.

bitloi:
> I just released launcher 1.6.0 with your change. I think you just need to update the launcher and update https://github.com/sbt/sbt/blob/develop/launch/src/main/input_resources/sbt/sbt.boot.properties.

@eed3si9n I've just updated them. Please check again.

main/src/main/scala/sbt/Defaults.scala
eed3si9n, line 1: If the launcher takes care of the overriding, do we need any changes in Defaults.scala?

bitloi: I removed the extra override logic.

sbt-app/src/repo-override-test/force-proxy-repos-file/force-proxy-repos-file/global/repositories_force
eed3si9n, line 1:
```scala
    val effectiveRepoConfig =
      if overrideRepos && repoConfigOverride.exists(_.exists())
      then repoConfigOverride
      else repoConfig
```

This means that we just need `repositories_force`?

bitloi: Just fixed it.

eed3si9n:
One thing I forgot to tell you is that scripted by default runs in parallel, so testing the launcher requires a different configuration. Luckily we have a skeleton setup already in https://github.com/sbt/sbt/tree/develop/sbt-app/src/repo-override-test.

## step 1: uncomment repo override test

https://github.com/sbt/sbt/blob/b634e1b507f87b3edf0b2b933828b742c05cbc9a/.github/workflows/ci.yml#L163

## step 2: replace the existing override scripted test

You can move your new scripted test into https://github.com/sbt/sbt/tree/develop/sbt-app/src/repo-override-test, and feel free to delete the existing one in there.

## steps 3: fix the configuration

https://github.com/sbt/sbt/blob/b634e1b507f87b3edf0b2b933828b742c05cbc9a/build.sbt#L1042-L1058 The repoOverride specific configuration is done here. Feel free to change whatever you have to there.

bitloi:
> One thing I forgot to tell you is that scripted by default runs in parallel, so testing the launcher requires a different configuration. Luckily we have a skeleton setup already in https://github.com/sbt/sbt/tree/develop/sbt-app/src/repo-override-test.
> 
> ## step 1: uncomment repo override test
> https://github.com/sbt/sbt/blob/b634e1b507f87b3edf0b2b933828b742c05cbc9a/.github/workflows/ci.yml#L163
> 
> ## step 2: replace the existing override scripted test
> You can move your new scripted test into https://github.com/sbt/sbt/tree/develop/sbt-app/src/repo-override-test, and feel free to delete the existing one in there.
> 
> ## steps 3: fix the configuration
> https://github.com/sbt/sbt/blob/b634e1b507f87b3edf0b2b933828b742c05cbc9a/build.sbt#L1042-L1058
> 
> The repoOverride specific configuration is done here. Feel free to change whatever you have to there.

I've fixed it. Would you mind reviewing it again?

bitloi:
@eed3si9n Do you have any other feedback?

eed3si9n (approved):
Thanks!

Issue #8763, opened 2026-02-19
[2.x] Does not report error if different scala binary versions


xuwei-k:
### `a/src/main/scala/A.scala`

```scala
object A {
  def a: Int = 2
}
```

### `b/src/main/scala/B.scala`

```scala
object B {
  def main(args: Array[String]): Unit = {
    println("hello " + A.a)
  }
}
```

### `build.sbt`

```scala
lazy val a = project.settings(scalaVersion := "2.13.18")
lazy val b = project.settings(scalaVersion := "2.12.21").dependsOn(a)
```

### `project/build.properties`

```properties
sbt.version=2.0.0-RC9
```

## problem

```
sbt update # expect error. but sbt report `[success]`
```

`sbt compile`

```
java.io.IOException: Scala signature A has wrong version
 expected: 5.0
 found: 5.2 in A.class
	at scala.reflect.internal.pickling.UnPickler$Scan.checkVersion(UnPickler.scala:122)
	at scala.reflect.internal.pickling.UnPickler$Scan.<init>(UnPickler.scala:64)
	at scala.reflect.internal.pickling.UnPickler.unpickle(UnPickler.scala:47)
	at scala.tools.nsc.symtab.classfile.ClassfileParser.unpickleOrParseInnerClasses(ClassfileParser.scala:1176)
	at scala.tools.nsc.symtab.classfile.ClassfileParser.parseClass(ClassfileParser.scala:467)
	at scala.tools.nsc.symtab.classfile.ClassfileParser.$anonfun$parse$2(ClassfileParser.scala:160)
	at scala.tools.nsc.symtab.classfile.ClassfileParser.$anonfun$parse$1(ClassfileParser.scala:146)
	at scala.tools.nsc.symtab.classfile.ClassfileParser.parse(ClassfileParser.scala:129)
	at scala.tools.nsc.symtab.SymbolLoaders$ClassfileLoader.doComplete(SymbolLoaders.scala:343)
	at scala.tools.nsc.symtab.SymbolLoaders$SymbolLoader.complete(SymbolLoaders.scala:250)
	at scala.tools.nsc.symtab.SymbolLoaders$SymbolLoader.load(SymbolLoaders.scala:269)
	at scala.tools.nsc.typechecker.Typers$Typer.isStale(Typers.scala:526)
	at scala.tools.nsc.typechecker.Typers$Typer.reallyExists(Typers.scala:517)
[...]

PR #8764, opened 2026-02-19
fix: [2.x] allowMismatchScala setting (#8802)

speedcoder430:
Address issue #8802


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -4189,20 +4189,22 @@ object Classpaths {
...
-                .withExplicitArtifacts(Vector.empty)
+            case _: CrossVersion.Binary if sbv != depSBV =>
+              throw new MessageOnlyException(
+                s"""Scala binary version mismatch in inter-project dependency: """
+                  + s"""project "${ref.project}" (Scala binary version "$sbv") """
+                  + s"""depends on project "${dep.project.project}" (Scala binary version "$depSBV"). """
+                  + s"""These Scala binary versions are not compatible. """
+                  + s"""To use Scala 2.13/3.x TASTy sandwich, """
+                  + s"""set one project to Scala 2.13 and the other to Scala 3.x."""
+              )
```
eed3si9n, line 4210: FYI - I've literally marked https://github.com/sbt/sbt/issues/8763 as not-a-🐛, which means at this point it's not considered a bug, so until we decide what to do, this PR might sit around for a while.

eed3si9n (changes requested):

main/src/main/scala/sbt/Defaults.scala
speedcoder430: OK, thanks

speedcoder430 (commented):
Accepted

speedcoder430:
Hi @eed3si9n , Hope you are doing well. Could you please review this PR again? Thanks

PR #8766, opened 2026-02-19
[2.x] fix: Reject java.io.File as cached task output type

### LINKED ISSUE #8762: [2.x] java.io.File yields incorrect cache
### build.sbt

```scala
scalaVersion := "2.12.21"

@transient
val assembly = taskKey[Unit]("")

val assemblyOutputPath = taskKey[File]("")

val assemblyJarName = taskKey[String]("")

assembly / target := crossTarget.value

assembly / assemblyOutputPath := (assembly / target).value / (assembly / assemblyJarName).value

assembly / assemblyJarName := "foo.jar"
```

### project/build.properties 

```properties
sbt.version=2.0.0-RC9
```

## problem

in sbt shell

```
sbt:assembly-output-path-cache-bug> print assembly / assemblyOutputPath
/Users/kenji/assembly-output-path-cache-bug/target/out/jvm/scala-2.12.21/assembly-output-path-cache-bug/foo.jar
sbt:assembly-output-path-cache-bug> ++ 2.13.18!
[info] Forcing Scala version to 2.13.18 on all projects.
[info] Reapplying settings...
[info] set current project to assembly-output-path-cache-bug (in build file:/Users/kenji/assembly-output-path-cache-bug/)
sbt:assembly-output-path-cache-bug> print assembly / assemblyOutputPath
/Users/kenji/assembly-output-path-cache-bug/target/out/jvm/scala-2.12.21/assembly-output-path-cache-bug/foo.jar
```

## expectation

expect

```diff
- target/out/jvm/scala-2.12.21/assembly-output-path-cache-bug/foo.jar
+ target/out/jvm/scala-2.13.18/assembly-output-path-cache-bug/foo.jar
```

speedcoder430:
## Problem

Fixes #8762

When a task returns `java.io.File` (e.g. `assembly / assemblyOutputPath`), the result gets cached by `ActionCache`. Since `File` serializes as an absolute path string, the cached value becomes stale when `scalaVersion` changes via `++ 2.13.18!` — the path still contains the old Scala version directory.

### Reproduction (from #8762)

```scala
// build.sbt
scalaVersion := "2.12.21"
val assemblyOutputPath = taskKey[File]("")
assembly / target := crossTarget.value
assembly / assemblyOutputPath := (assembly / target).value / (assembly / assemblyJarName).value
```

```
sbt> print assembly / assemblyOutputPath
.../scala-2.12.21/.../foo.jar
sbt> ++ 2.13.18!
sbt> print assembly / assemblyOutputPath
.../scala-2.12.21/.../foo.jar   // BUG: should be scala-2.13.18
```

## Solution

In `Cont.scala`'s `callActionCache` method, detect when the task's return type `A` contains `java.io.File` at compile time using `containsFileType[A]`. When it does, skip wrapping the task body with `ActionCache.cache(...)`, while still processing any `Def.declareOutput(...)` calls via `letOutput`. This way:

- `taskKey[File]` tasks are never cached (since `File` isn't content-addressable)
- `Def.declareOutput(...)` continues to work correctly for File tasks
- No user-facing changes required — existing builds just work

## Tests

- `cache/skip-file-cache` scripted test: Defines a `taskKey[File]` that depends on `scalaVersion`, switches versions, and asserts the returned path updates.
- Verified `actions/task-map` (uses `Def.declareOutput` with `taskKey[File]`) continues to pass.
- Ran full CI job 2 suite locally: `actions/* cache/* tests/* compiler-project/*` etc. — all pass.


eed3si9n:
3. Follow our AI assisted contributions guideline (please disclose AI usages in the PR description, if any)
5. Generally we recommend getting CI to pass on a forked repo before sending a pull request (pull requests that do not pass tests will be closed after a few days)

speedcoder430:
> Hi, thanks for the contribution! Please check out our [Contributor's Guildeline](https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md) and:
> 
> 1. Make sure you can reproduce the problem and the fix manually
> 2. Include tests for the change
> 3. Follow our AI assisted contributions guideline (please disclose AI usages in the PR description, if any)
> 4. Multiple contributors might work on this issue in parallel
> 5. Generally we recommend getting CI to pass on a forked repo before sending a pull request (pull requests that do not pass tests will be closed after a few days)
> 6. Sign Scala CLA https://contribute.akka.io/contribute/cla/scala

Thanks for your detailed feedback.

core-macros/src/main/scala/sbt/internal/util/appmacro/Cont.scala
```diff
@@ -342,38 +352,42 @@ trait Cont:
...
-              $input,
-              codeContentHash = Digest.dummy($codeContentHash),
-              extraHash = Digest.dummy($extraHash),
-              tags = $tagsExpr,
-              config = $cacheConfigExpr,
-            )({ _ =>
-              $block
-            })
-        }
+        if containsFileType[A1] then
```
eed3si9n, line 355: I like the general direction of catching this inside of the macro. However, I think we should actually throw the error when `File` is asked to be cached, rather than silently ignoring it, similar to the way you had this PR earlier. The blast radius of this change would be bigger, but in the long-term it would make the tool more predictable.

Also, the error message should not only say that `File` is not supported, but recommend that they migrate to either `xsbti.VirtualFileRef`, `xsbti.HashedVirutalFileRef`, and link to doc https://www.scala-sbt.org/2.x/docs/en/concepts/caching.html#caching-files.

eed3si9n (changes requested):

core-macros/src/main/scala/sbt/internal/util/appmacro/Cont.scala
speedcoder430: I changed my code. Please review my code again.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -294,12 +294,14 @@ object Defaults extends BuildCommon {
       csrSameVersions :== Nil,
       stagingDirectory := (ThisBuild / baseDirectory).value / "target" / "sona-staging",
       localStaging := Some(Resolver.file("local-staging", stagingDirectory.value)),
-      sonaBundle := Publishing
-        .makeBundle(
-          stagingDirectory.value.toPath(),
-          ((ThisBuild / baseDirectory).value / "target" / "sona-bundle" / "bundle.zip").toPath()
-        )
-        .toFile(),
+      sonaBundle := Def.uncached(
```
eed3si9n: As a general strategy, I think we should put `@trainsient` on Keys.scala instead of putting `Def.uncached(...)` in Defaults.scala since this would force the build users to also put `Def.uncached(...)` when they rewire the task.

eed3si9n (approved):
Thanks!

xuwei-k:
- https://github.com/sbt/sbt/issues/9014

PR #8773, opened 2026-02-21
[2.x] feat: Pretty-print dependency lock file

### LINKED ISSUE #8770: [2.x] Make dependencyLock file pretty
## Steps

Using sbt 2.0.0-RC9

```bash
> dependencyLock
```

## Problem

The generated lock file is one-line JSON, which is difficult to review.

## Expectation

Output pretty-printed JSON.

spider-yamet:
Fixes #8770

## Problem

The `dependencyLock` task writes `deps.lock` as a single line of JSON, which is difficult to review and diff (see #8770).

## Solution

Use `PrettyPrinter` instead of `CompactPrinter` in `LockFile.write` so the lock file is written as pretty-printed, multi-line JSON. Schema and semantics are unchanged; only formatting changes. Both the manual `dependencyLock` task and the lock file written during resolution use this path.

## Testing

- Reproduced the issue: ran `dependencyLock` on sbt 2.0.0-RC9 and confirmed one-line output before the change.
- Added unit test in `LockFileSpec`: "LockFile.write outputs pretty JSON" (asserts written content contains newlines).
- Existing lock round-trip and `dependencyLockCheck` behavior unchanged; parser accepts both compact and pretty JSON.

## Checklist

- [x] Reproduced the problem before changing code.
- [x] Change compiles and fixes the problem.
- [x] Unit test added.
- [x] Small, focused change (formatting only).

### DIFF
--- a/lm-coursier/src/main/scala/lmcoursier/internal/LockFile.scala
+++ b/lm-coursier/src/main/scala/lmcoursier/internal/LockFile.scala
@@ -3,7 +3,7 @@ package lmcoursier.internal
 import java.io.File
 import java.nio.file.Files
 import java.nio.charset.StandardCharsets
-import sjsonnew.support.scalajson.unsafe.{ CompactPrinter, Converter, Parser }
+import sjsonnew.support.scalajson.unsafe.{ Converter, Parser, PrettyPrinter }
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
@spider-yamet Thanks!

Issue #8767, opened 2026-02-20
[1.x] `.jvmopts` parse error on Windows Git Bash 


xuwei-k:
## steps

sbt 1.12.3

`.jvmopts`

```
-Xmx2G
-Xss1M
```

## problem

https://github.com/sbt/zinc/actions/runs/22214481394/job/64255297794?pr=1641#step:6:1

```
Run sbt -v -Dfile.encoding=UTF-8 -Dsbt.supershell=never "Test/compile" "crossTestBridges" "zincRoot/test" "scripted"
[sbt_options] declare -a sbt_options=()
[process_args] java_version = '17'
Invalid maximum heap size: -Xmx2G
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
[copyRt] java9_rt = '/rt.jar'
mkdir: cannot create directory '': No such file or directory
Invalid maximum heap size: -Xmx2G
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
# Executing command line:
java
-Dfile.encoding=UTF-8
-Xmx2G
-Xss1M
-Dfile.encoding=UTF-8
-Dsbt.supershell=never
-Dsbt.script=/c/hostedtoolcache/windows/sbt/1.12.3/sbt/bin/sbt
-Dscala.ext.dirs=
-Djline.terminal=jline.UnixTerminal
-Dsbt.cygwin=true
-jar
C:\hostedtoolcache\windows\sbt\1.12.3\sbt\bin\sbt-launch.jar
Test/compile
crossTestBridges
zincRoot/test
scripted

Invalid maximum heap size: -Xmx2G
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
Error: Process completed with exit code 1.
```

eed3si9n:
We probably need to bisect manually to figure out which commit introduced the regression.

xuwei-k:
https://github.com/sbt/sbt/commits/v1.12.3/launcher-package/src/universal/bin/sbt.bat

- `sbt/setup-sbt@v1.1.17` default sbt version is 1.12.3 => failure
- `sbt/setup-sbt@v1.1.16` default sbt version is 1.12.1 => success


1.12.2 => success https://github.com/xuwei-k/zinc/commit/c2fd114b2ba449ea1ccf35b0ce2d177cf61861ba

```yaml
    - name: Setup SBT
[...]

PR #8774, opened 2026-02-21
[1.12.x] fix: prefer JAVA_HOME/JAVACMD in shell runner

aviu16:
## Summary
This PR targets the shell-runner Java selection path related to #8767 logs.

It updates `sbt` to prefer explicit Java configuration (`JAVACMD`, then `JAVA_HOME`) before falling back to `java` on PATH.

## Why
In Windows Git Bash scenarios, shell-runner execution can resolve an unexpected Java from PATH even when `JAVA_HOME` is set. Prioritizing explicit Java selection avoids that mismatch.

## Changes
- `sbt`
  - if `JAVACMD` is set, use it
  - else if `JAVA_HOME` is set, use `$JAVA_HOME/bin/java` (with cygwin path conversion when needed)
  - otherwise keep existing fallback to `java`

## Scope
No changes to `sbt.bat` in this revision.

### DIFF
--- a/launcher-package/integration-test/src/test/scala/RunnerMemoryScriptTest.scala
+++ b/launcher-package/integration-test/src/test/scala/RunnerMemoryScriptTest.scala
@@ -83,4 +83,13 @@ object RunnerMemoryScriptTest extends verify.BasicTestSuite with ShellScriptUtil
       assert(out.contains[String]("-Xms512M"))
       assert(out.contains[String]("-Xmx1G"))
 
+  // Regression test for #8767: options from .jvmopts should be passed correctly on Windows
+  testOutput(
+    "sbt use .jvmopts file with uppercase memory suffixes",
+    jvmoptsFileContents = """-Xmx2G
+        |-Xss1M""".stripMargin
+  )("compile", "-v"): (out: List[String]) =>
+    assert(out.contains[String]("-Xmx2G"))
+    assert(out.contains[String]("-Xss1M"))
+
 end RunnerMemoryScriptTest
--- a/launcher-package/integration-test/src/test/scala/RunnerMemoryScriptTest.scala
+++ b/launcher-package/integration-test/src/test/scala/RunnerMemoryScriptTest.scala
@@ -93,12 +93,13 @@ object RunnerMemoryScriptTest extends verify.BasicTestSuite with ShellScriptUtil
     assert(out.contains[String]("-Xss1M"))
 
   testOutput(
-    "sbt prefers JAVA_HOME over PATH when JAVACMD is unset",
+    "sbt prefers JAVA_HOME over PATH in cygwin shell when JAVACMD is unset",
     jvmoptsFileContents = """-Xmx2G
         |-Xss1M""".stripMargin,
     failingPathJava = true,
     useJavaHomeFromTestBin = true,
-    setWindowsJavacmd = false
+    setWindowsJavacmd = false,
+    simulateCygwinShell = true
   )("compile", "-v"): (out: List[String]) =>
     assert(out.contains[String]("-Xmx2G"))
     assert(out.contains[String]("-Xss1M"))
--- a/launcher-package/integration-test/src/test/scala/RunnerMemoryScriptTest.scala
+++ b/launcher-package/integration-test/src/test/scala/RunnerMemoryScriptTest.scala
@@ -92,4 +92,15 @@ object RunnerMemoryScriptTest extends verify.BasicTestSuite with ShellScriptUtil
     assert(out.contains[String]("-Xmx2G"))
     assert(out.contains[String]("-Xss1M"))
 
+  testOutput(
+    "sbt prefers JAVA_HOME over PATH when JAVACMD is unset",
+    jvmoptsFileContents = """-Xmx2G
+        |-Xss1M""".stripMargin,
+    failingPathJava = true,
+    useJavaHomeFromTestBin = true,
+    setWindowsJavacmd = false
+  )("compile", "-v"): (out: List[String]) =>
+    assert(out.contains[String]("-Xmx2G"))
+    assert(out.contains[String]("-Xss1M"))
+
 end RunnerMemoryScriptTest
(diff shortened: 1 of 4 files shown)


launcher-package/integration-test/src/test/scala/RunnerMemoryScriptTest.scala
```diff
@@ -83,4 +83,13 @@ object RunnerMemoryScriptTest extends verify.BasicTestSuite with ShellScriptUtil
...
       assert(out.contains[String]("-Xmx1G"))
 
+  // Regression test for #8767: options from .jvmopts should be passed correctly on Windows
+  testOutput(
+    "sbt use .jvmopts file with uppercase memory suffixes",
+    jvmoptsFileContents = """-Xmx2G
+        |-Xss1M""".stripMargin
+  )("compile", "-v"): (out: List[String]) =>
+    assert(out.contains[String]("-Xmx2G"))
+    assert(out.contains[String]("-Xss1M"))
```
eed3si9n, line 93: Does this test fail without the change in `sbt.bat`? At this point I'm not sure if this is a bug in `sbt.bat` or `sbt.sh`.

aviu16:
Good catch, you are right this test as written does not isolate sbt.bat behavior by itself. I will rework this to align directly with the failing path from #8767, which appears shell-runner based from the log markers, and then update the PR accordingly.

launcher-package/integration-test/src/test/scala/RunnerMemoryScriptTest.scala
aviu16: Good point. I updated the PR to target the shell runner path from #8767 instead of `sbt.bat`. The current patch now only changes `sbt` to prefer `JAVACMD`/`JAVA_HOME` before PATH, which aligns with the failing Windows Git Bash path in the linked logs. Please take another look.

aviu16:
Quick status check-in: this PR intentionally limits the fix to shell-runner Java resolution order (JAVACMD/JAVA_HOME before PATH) to match #8767 behavior. If you prefer, I can further tighten or adjust the regression coverage to isolate the exact failing path more explicitly. Happy to update immediately based on your preferred test shape.

eed3si9n (changes requested):
The title says fixes #8767 but I don't think this PR can claim that without reproducing the original error using Git Bash on Windows somehow. As far as I know, on Linux side the runner script test uses a Python dummy version of `java` to speed up the test.

aviu16:
@eed3si9n You are right. I removed the "fixes #8767" claim from the title and adjusted the PR summary accordingly. This PR is now scoped as a shell-runner Java selection improvement related to that report, not a full repro-verified fix claim.

aviu16:
@eed3si9n Addressed with a new repro-oriented runner test in this PR branch. I added coverage that forces a failing `java` on PATH, unsets `JAVACMD`, sets `JAVA_HOME`, and asserts the launcher still succeeds (proving the script prefers `JAVA_HOME/bin/java` over PATH in that scenario). Changes are in commit b84c24a1c (RunnerMemoryScriptTest + ShellScriptUtil test harness).

sbt
```diff
@@ -93,6 +93,21 @@ cygwinpath() {
...
+elif [[ -n "$JAVA_HOME" ]]; then
+  if [[ "$CYGWIN_FLAG" == "true" ]]; then
+    java_home_unix=$(cygpath -u "$JAVA_HOME" 2>/dev/null)
+    if [[ -n "$java_home_unix" ]] && [[ -x "$java_home_unix/bin/java" ]]; then
+      java_cmd="$java_home_unix/bin/java"
+    fi
+  elif [[ -x "$JAVA_HOME/bin/java" ]]; then
+    java_cmd="$JAVA_HOME/bin/java"
+  fi
+fi
```
eed3si9n: If there are `java` on PATH and `JAVA_HOME` set, do we know for sure that prioritizing `JAVA_HOME` is the right move? This could potentially break thousands of CI jobs.

aviu16: Good point. I narrowed this in commit e1737a886: `JAVACMD` still wins everywhere, but `JAVA_HOME` is now only prioritized in Cygwin/MSYS/MINGW shell paths (the Git Bash-style case), so non-Windows PATH-vs-JAVA_HOME behavior stays unchanged. I also updated the test harness + runner test to simulate that shell path and keep the precedence coverage scoped to the intended environment.

aviu16:
@eed3si9n I found and fixed the Windows CI failure from my new test. The failing case was running in the Windows client suite via `sbt.bat`, so it was exercising the fake PATH `java` setup instead of the shell-runner (`sbt`) cygwin/MSYS path I intended to test. I scoped that new cygwin-shell simulation test to non-Windows only (`windowsSupport = false`) so it runs where the shell runner is actually exercised with the simulation harness. Pushed in commit `2bd7b5a3f`.

aviu16:
@eed3si9n Quick follow-up after the rerun: the earlier Windows client-test failure from my new test is fixed (Windows is passing on the rerun), but the remaining Client Test failures now look unrelated to this PR change.  On macOS, the failing run shows a JNA native library architecture error in `sbt-socket-server` (`UnsatisfiedLinkError` loading an x86_64 JNA temp lib on arm64), and both macOS + ubuntu client jobs eventually ended after long runtimes as canceled/failed in the same Client Test workflow run. I do not see a deterministic failure from the new launcher integration test in the available logs. If you want, I can re-run the Client Test workflow again, but at this point it looks like CI/runtime flakiness rather than the shell-runner change in this PR.

sbt
eed3si9n: With regards to https://github.com/sbt/sbt/issues/8767, the bug was fixed by reverting https://github.com/sbt/sbt/pull/8730. I don't know if this would've really helped.

aviu16:
@eed3si9n Thanks for the clarification on #8767. Agreed this PR should not be tied to that issue anymore. This PR is intended as a separate shell-runner hardening change for the Git Bash/Cygwin/MSYS path where `JAVA_HOME` can point to the intended runtime while `java` on PATH is different. The current patch is scoped so `JAVACMD` still wins everywhere, and the `JAVA_HOME` preference is only applied in the Cygwin/MSYS/MINGW shell path (so non-Windows PATH-vs-JAVA_HOME behavior stays unchanged). If that standalone behavior is useful, I’m happy to keep this focused and adjust further.

eed3si9n:
Closing this in favor of https://github.com/sbt/sbt/pull/8779, which adds testing on Git Bash.

PR #8776, opened 2026-02-21
[2.x]fix: Non-fork test classloader respects autoScalaLibrary

### LINKED ISSUE #4009: Non-fork test ends up with unwanted scala-library even with autoScalaLibrary := false
Originally reported by @retronym 

## steps

See https://github.com/scala/scala/pull/6300#issuecomment-372495182

## problem

Classloader used by non-fork test includes scala-library in ScalaInstance.

## expectation

Construct Classloader without ScalaInstance when `autoScalaLibrary := false`

bitloi:
## Problem

With `Test / autoScalaLibrary := false`, the non-fork test classloader still included the ScalaInstance-based scala-library layer. Tests could load the wrong scala-library (e.g. sbt's) and hit `NoSuchMethodError` or version mismatch (e.g. [scala/scala](https://github.com/scala/scala) had to use `fork in Test := true` as a workaround).

## Solution

In `ClassLoaders.testTask`:

- When `autoScalaLibrary.value` is false, use `ClassLoaderLayeringStrategy.Flat` so the test classloader is built without the ScalaInstance layer.
- Use project classpath only (`rawCP`) for `fullCP` in that case (do not prepend `si.libraryJars`).
- Do not add `si.libraryJars` to the exclude set when `autoScalaLibrary` is false.

No change to forked tests or to the `run` task.

## Verification

1. **Reproduce (without fix):** In a project with `Test / autoScalaLibrary := false` and a test framework that pulls in Scala (e.g. ScalaCheck), run `test` without forking; previously the framework could load scala-library from the ScalaInstance layer and fail (e.g. `NoSuchMethodError: scala.collection.immutable.Set$.apply`).
2. **Verify:** Same project runs `test` successfully with this fix (test classloader uses only project classpath).
3. **Scripted:** `./sbt "scripted tests/autoScalaLibrary-test-loader"` — runs tests with `Test / autoScalaLibrary := false`.
4. **Regression:** `./sbt "scripted classloader-cache/scalatest"` — passes.

## Tests / lint

- `./sbt compile` — success
- `./sbt "scripted tests/autoScalaLibrary-test-loader"` — success
- `./sbt "scripted classloader-cache/scalatest"` — success

Closes #4009


eed3si9n (approved):
Looks great! Thanks!

PR #8780, opened 2026-02-22
[2.x] Fallback to $HOME/.config/sbt

eed3si9n:
Ref https://github.com/sbt/sbt/issues/3681

## Problem/Solution
When XDG_CONFIG_HOME is not defined, we should fallback to `$HOME/.config/sbt`.


soc:
Thank you @eed3si9n, this looks good!

Just as a note (not saying I think it should be done, just for consideration):

Some applications that migrated to XDG still check whether their "legacy" location already exists and in that case keep using it (to not break existing users and let them migrate at their own time).

No opinion on whether that should be done in this case though, just wanted to put the info out here.

PR #8788, opened 2026-02-23
[2.x] refactor: Use `enum` instead of `scala.Enumeration` in sbt.Execute.State

xuwei-k:
I think this is a binary compatible change because `State` is `private`.

### DIFF
--- a/tasks/src/main/scala/sbt/Execute.scala
+++ b/tasks/src/main/scala/sbt/Execute.scala
@@ -75,9 +75,8 @@ private[sbt] final class Execute(
       view.inline1(a) match
         case Some(v) => Result.Value(v())
         case None    => results(a)
-  private type State = State.Value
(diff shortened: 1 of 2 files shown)


tasks/src/main/scala/sbt/Execute.scala
```diff
@@ -75,9 +75,8 @@ private[sbt] final class Execute(
       view.inline1(a) match
         case Some(v) => Result.Value(v())
         case None    => results(a)
-  private type State = State.Value
-  private object State extends Enumeration {
-    val Pending, Running, Calling, Done = Value
+  private enum State {
+    case Pending, Running, Calling, Done
```
eed3si9n, line 79: lgtm; plz add mima exclusion.

eed3si9n (approved):

PR #8794, opened 2026-02-23
[2.x] fix: Respect last plugin toggle for enablePlugins/disablePlugins

### LINKED ISSUE #1926: No way to enable plugin after it was disabled
i have a bunch of projects and use sbt-assembly
i want sbt-assembly to be disabled by default (not sure why this is not the default)
so i use myProject(id, base) = Project(id, base).disablePlugins(Assembly).bunchofotherdefaults
in one particular project i want to enable the plugin, but i get an exception if i use enablePlugin

```
sbt.AutoPluginException: Error determining plugins for project 'keen-storefront' in D:\keen\storefront:
Contradiction in selected plugins.  These plugins were both included and excluded: sbtassembly.AssemblyPlugin
        at sbt.AutoPluginException.withPrefix(Plugins.scala:106)
```

imo the last setting should win just like it behaves everywhere, and there should be no exceptional cases

it-education-md:
Closes #1926

### Problem
Calling `disablePlugins(X).enablePlugins(X)` (or the reverse order) could leave both include and exclude for the same plugin in the selected plugin expression, which caused `AutoPluginException` contradiction during project load/reload.

### Solution
Introduce explicit toggle precedence (`last call wins`) when combining plugin selections:
- `disable -> enable` keeps plugin enabled
- `enable -> disable` keeps plugin disabled

Apply the same behavior to both `Project` and `ProjectMatrix`.

### Testing
- Reproduced with scripted test setup in `sbt-app/src/sbt-test/project/i1926-disable-enable-plugin`
- Added unit coverage in `main/src/test/scala/ProjectSpec.scala`
- Verified locally:
  - `sbt "mainProj/testOnly sbt.ProjectSpec"`
  - `sbt "scripted project/i1926-disable-enable-plugin"`

### DIFF
--- a/main-settings/src/main/scala/sbt/Plugins.scala
+++ b/main-settings/src/main/scala/sbt/Plugins.scala
@@ -343,6 +343,16 @@ ${listConflicts(conflicting)}""")
     case And(ns)  => ns.foldLeft(a)(_ && _)
     case b: Basic => a && b
   }
+
+  private[sbt] def overrideWith(current: Plugins, update: Plugins): Plugins = {
+    val opposite: Set[Basic] = flatten(update).map {
+      case Exclude(p) => p: Basic
+      case p: AutoPlugin =>
+        Exclude(p): Basic
+    }.toSet
+    and(remove(current, opposite), update)
+  }
+
   private[sbt] def remove(a: Plugins, del: Set[Basic]): Plugins = a match {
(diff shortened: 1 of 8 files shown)


it-education-md:
@eed3si9n please review my PR. thank you.

eed3si9n (approved):
Thanks!

PR #8801, opened 2026-02-23
fix: Cache doc task to detect deleted output directory

### LINKED ISSUE #7078: doc task does not regenerate scaladocs when api directory is removed
## steps

With sbt 1.7.1:

1. generate scaladocs by running `sbt doc`
2. remove directory `target/scala-3.2.0/api`
3. run `sbt doc` again

## problem

Running `sbt doc` after removing the `api` directory does not recreate it. sbt reports success instead:

```
$ sbt doc
downloading sbt launcher 1.8.0
[info] welcome to sbt 1.7.1 (Amazon.com Inc. Java 11.0.17)
[info] loading settings for project scalapr-build from plugins.sbt ...
[info] loading project definition from /builds/jaburjak/scalapr/project
[info] set current project to scalapr (in build file:/builds/jaburjak/scalapr/)
[success] Total time: 4 s, completed Nov 15, 2022, 11:23:09 PM
```

## expectation

sbt should notice that the `api` directory was removed and recreate it.

## notes

Creating empty `target/scala-3.2.0/api` directory after removing it makes sbt notice that scaladocs are missing and it will recreate them.

I stumbled upon this issue when setting up continuous integration pipeline for my project. To speed up builds, I enabled caching of the `target` directory between pipelines. The last stage of the pipeline is to generate scaladocs and publish them to GitLab Pages, which it does by running `sbt doc` and then moving `target/scala-3.2.0/api` to another directory that gets published by Pages.

This worked the first time the pipeline was run. Then on subsequent runs, because the `target` directory was cached (I assume there’s something in there telling sbt that scaladocs are already generated) but `api` directory was not there (cache is created at the end of the pipeline), the pipeline broke because sbt did not recreate it.

since-2017-hub:
Closes #7078 

### **Summary** 

Added input/output change tracking to the doc task using Tracked.inputChangedW. It now detects when the output API directory is deleted and regenerates documentation, instead of silently reporting success. Also caches unchanged runs to avoid unnecessary regeneration.


since-2017-hub:
@eed3si9n 
Please review this PR.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2079,38 +2091,94 @@ object Defaults extends BuildCommon {
...
+                  ),
                   resolvedOptions,
-                  maxErrors.value,
-                  s.log,
+                  maxErr,
                 )
+                val cachedDoc =
+                  Tracked.inputChangedW(s.cacheStoreFactory.make("inputs")) {
+                    (inChanged: Boolean, _: Inputs) =>
+                      Tracked.inputChangedW(s.cacheStoreFactory.make("output")) {
```
eed3si9n: Could we try using the sbt 2. cached task instead please?
https://www.scala-sbt.org/2.x/docs/en/reference/cached-task.html

eed3si9n (changes requested):

since-2017-hub:
@eed3si9n 
Main branch was update, could you allow again?

since-2017-hub:
@eed3si9n 
Could you allow to CI test?

mzuehlke:
You can enable GitHub Actions in your personal fork to get feedback from the CI.

PR #8804, opened 2026-02-23
[2.x] feat: Add allowMismatchScala setting

### LINKED ISSUE #8802: [2.x] allowMismatchScala setting
## steps

sbt 2.x allows `dependsOn(...)` mismatched Scala subproject dependencies by design, but the error state might be confusing.

## expectations

1. Provide a setting `allowMismatchScala`, which by default should be set to `false` at the global level.
2. When `allowMismatchScala` is set to `false`, error out when mismatch Scala subprojects are detected, with the exception of Scala 2.13 and (3.0 through 3.7).
3. When `allowMismatchScala` is set to `true` for the subproject, allow `dependsOn(...)` to an arbitrary Scala version as it is coded right now in develop branch.

dev-miro26:
Fixes #8802

I successfully signed the Scala CLA.

## Problem

sbt 2.x allows `dependsOn(...)` between subprojects with mismatched
Scala versions without any warning or error. This can lead to confusing
classpath issues at compile or runtime, especially now that Scala 3.8+
has dropped backward TASTy compatibility with 2.13.

## Solution

1. Added `allowMismatchScala` setting key (default `false` at global level).
2. When `false`, errors on Scala binary version mismatches between
   `dependsOn` projects, with an exception for the Scala 2.13 ↔ 3.0–3.7
   sandwich (traditional TASTy compatibility).
3. When `true` on a project, allows arbitrary Scala version dependencies.

## Testing

- 9 HedgeHog unit tests for `isAllowedScalaMismatch` edge cases
- Scripted test `project/allow-mismatch-scala` covering:
  - 2.13 → 3.5 (allowed sandwich)
  - 2.13 → 3.8 (blocked mismatch)
  - Override with `allowMismatchScala := true`
- `mimaReportBinaryIssues` passes (no bincompat issues)

## AI Disclosure

This PR was created with AI assistance (Claude Opus 4.6). All changes
were compiled, tested, and verified.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -182,6 +182,7 @@ object Defaults extends BuildCommon {
       autoScalaLibrary :== true,
       managedScalaInstance :== true,
       allowUnsafeScalaLibUpgrade :== false,
+      allowMismatchScala :== false,
       classpathEntryDefinesClass := Def.uncached { (file: File) =>
         sys.error("use classpathEntryDefinesClassVF instead")
       },
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2174,6 +2174,11 @@ object Defaults extends BuildCommon {
   }
 
   def compileTask: Initialize[Task[CompileAnalysis]] = Def.task {
+    ClasspathImpl.validateScalaVersions(
+      thisProjectRef.value,
+      buildDependencies.value,
+      settingsData.value,
+    )
     val setup: Setup = compileIncSetup.value
     val store = analysisStore(compileAnalysisFile)
     val c = fileConverter.value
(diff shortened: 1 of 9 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2173,6 +2174,11 @@ object Defaults extends BuildCommon {
   }
 
   def compileTask: Initialize[Task[CompileAnalysis]] = Def.task {
+    ClasspathImpl.validateScalaVersions(
+      thisProjectRef.value,
+      buildDependencies.value,
+      settingsData.value,
+    )
```
eed3si9n: Mixing-and-matching of Scala version was implemented fairly recently here - https://github.com/sbt/sbt/pull/8681. I like the idea of moving the validation function to `main/src/main/scala/sbt/internal/ClasspathImpl.scala`, but could you perform the check in `projectDependenciesTask` if possible?

eed3si9n (changes requested):

main/src/main/scala/sbt/Defaults.scala
dev-miro26: Thanks, let me check again.

eed3si9n (approved):
Thanks!

PR #8806, opened 2026-02-23
[2.x] Avoid old varargs splice syntax in test

xuwei-k:
https://docs.scala-lang.org/scala3/reference/changed-features/vararg-splices.html


eed3si9n (approved):
Thanks!

PR #8810, opened 2026-02-24
[2.x] fix: Fixes sbt runner script' sbtn detection

fireXtract:
`sbtn-x86_64-apple-darwin` doesn't exist anymore.

While the existing solution does work, it seemed worth tidying up while I had it in mind.

I tested this fix on 1.10.1 and 1.12.3, and it prevents entering the _acquire_sbtn_ function when _detectNativeBinary_ detects the binary already exists.


eed3si9n (approved):
Thanks!

PR #8817, opened 2026-02-26
[2.x] refactor: Extract color-default logic as a pure function

### LINKED ISSUE #6050: extract color-default logic as a pure function
https://github.com/sbt/sbt/blob/5948ec64993c85fcff18f4c553a7ac1e497cbba2/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala#L314-L321 is somewhat complicated and hard to remember the logic. We should make a pure function like

```scala
def isColorDefault(propsColor: Boolean, ...): Boolean
```

so we can test against it.

_Originally posted by @eed3si9n in https://github.com/sbt/sbt/pull/6044#issuecomment-720683673_

eureka0928:
Fixes #6050

## Summary
Extract the color-detection logic in `Terminal.scala` into a pure, testable function `isColorDefault`.

## Problem
The color-default logic in `Terminal.useColorDefault` is interleaved with side effects (system property reads, environment variable checks, `System.console` calls) making it hard to understand the priority chain and impossible to unit test.

## Solution
Extract a pure `isColorDefault` function that takes all inputs as parameters and returns whether color should be enabled. The existing `useColorDefault` delegates to it with the actual system values.

The priority chain (documented in Scaladoc):
1. `propsColor` — explicit color preference from terminal properties
2. `colorProp` — explicit color option (`sbt.color` / `sbt.colour`)
3. `logFormatOpt` — log format option (`sbt.log.noformat` / `sbt.log.format`), `None` defaults to `true`
4. Terminal heuristic — `hasConsole && !isDumbTerminal || isCI || isEmacs`

## Changes
- `internal/util-logging/.../Terminal.scala`
  - Added `private[sbt] def isColorDefault(...)` pure function
  - Refactored `useColorDefault` to delegate to it
- `internal/util-logging/.../TerminalColorSpec.scala`
  - Added 8 unit tests covering all priority levels and heuristic branches

## Verification
```
$ sbt "utilLogging/testOnly sbt.internal.util.TerminalColorSpec"
[info] Passed: Total 10, Failed 0, Errors 0, Passed 10
```

## AI Disclosure
This PR was developed with assistance from Claude (claude-opus-4-6).

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
@@ -378,17 +378,41 @@ object Terminal {
       }
     }
   }
-  private def useColorDefault: Boolean = {
-    // This approximates that both stdin and stdio are connected,
-    // so by default color will be turned off for pipes and redirects.
-    props
-      .map(_.color)
-      .orElse(isColorEnabledProp)
+
+  /**
+   * Pure function that determines whether color output should be enabled.
+   *
+   * Priority (highest to lowest):
+   *   1. propsColor — explicit color preference from terminal properties
+   *   2. colorProp — explicit color option
+   *   3. logFormatOpt — log format option (None defaults to true)
+   *   4. terminal heuristic — console type, CI, and Emacs detection
+   */
+  private[sbt] def isColorDefault(
+      propsColor: Option[Boolean],
+      colorProp: Option[Boolean],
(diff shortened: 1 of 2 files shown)


fury0928:
@eed3si9n Ready for review. All 10 tests pass locally. This extracts `useColorDefault` into a pure `isColorDefault` function as you suggested in #6044.

eed3si9n (approved):
Thanks!

PR #8818, opened 2026-02-26
[2.x] Fix lazy val name handling in Eval

xuwei-k:
in sbt 2.0.0-RC9

```
[error] dotty.tools.dotc.core.TypeError$$anon$1: object $Wrapb6d74462cf does not have a member method $tilde$eq$less$greater$bang$hash$percent$up$amp$bar$times$div$plus-$colon$bslash$bslash$qmark$at
```


eed3si9n (approved):
Thanks!

PR #8819, opened 2026-02-26
[2.x] fix: sbt -debug doesn't display loading log

### LINKED ISSUE #4979: `sbt -debug` doesn't display loading log
## steps

```
sbt -debug
```

## problem

It does display some logs, but not the debug level logs during loading.

## expectation

It shows the debug level logs during loading.

## workaround

A workaround is to type `last` in sbt shell immediately after it gets up.

bitloi:
## Problem

With `sbt -debug`, debug-level logs during project loading are not shown on the console. Some logs appear, but the console stays at Info level until after loading completes (when the `debug` command runs). The workaround was to run `last` in the sbt shell to see the backing log.

**Root cause:** The initial global logging was always created with `Level.Info`. The `-debug` flag is treated as an early command that runs *after* the boot sequence starts, so `setLogLevel(Debug)` ran only after loading.

## Solution

- **Parse log level from startup arguments** in `StandardMain.initialState`: detect the first `-level` / `--level` option (e.g. `-debug`, `--debug`) in `configuration.arguments` and derive the corresponding `Level`.
- **Apply that level at startup:** Pass the level into `initialGlobalLogging` and use it in `GlobalLogging.initial()` so the initial console appender uses that level instead of always `Level.Info`.
- **Set initial state attributes** when a level option is present: put `Keys.logLevel` and `BasicKeys.explicitGlobalLogLevels` in the initial state so the rest of the run stays at that level.

Debug (and other level options like `-warn`, `-error`) now take effect from the first log line, including during project loading.

Closes #4979 

## Changes

- `internal/util-logging`: `GlobalLogging.initial` gains optional `initialLevel: Level.Value = Level.Info`.
- `main`: `StandardMain.logLevelFromArguments`, `initialGlobalLogging(Option[File], Level)`, and `initialState` wiring as above.
- `main/src/test`: `InitialLogLevelSpec` with unit tests for `logLevelFromArguments`.
- `docs/fix-4979-manual-verification.md`: short note for manual reproduction.

## Testing

- [x] Unit tests: `mainProj/Test/testOnly sbt.internal.InitialLogLevelSpec` (5 tests).
- [x] Full mainProj tests: `mainProj/Test/test` (116 tests).
- [x] Scalafmt: `scalafmtCheckAll` passes.
- [x] Manual: run `sbt -debug`; debug output appears on console during "Loading project definition...".

## Checklist

- [x] Reproduced the problem before changing code.
- [x] Change is small and focused on this issue.
- [x] Follows project coding style (Scala 3, scalafmt).
- [x] No new binary compatibility issues (additive/default parameter only).


bitloi:
@eed3si9n Ready for review

main/src/main/scala/sbt/Main.scala
```diff
@@ -265,20 +277,26 @@ object StandardMain {
...
     if (!sys.props.contains("jna.nosys")) sys.props.put("jna.nosys", "true")
 
     import BasicCommandStrings.{ DashDashDetachStdio, DashDashServer, isEarlyCommand }
     val userCommands =
       configuration.arguments
         .map(_.trim)
         .filterNot(c => c == DashDashDetachStdio || c == DashDashServer)
+        .toSeq
+    val initialLevel = logLevelFromArguments(userCommands)
     val (earlyCommands, normalCommands) = (preCommands ++ userCommands).partition(isEarlyCommand)
```
eed3si9n, line 294: Thanks for working on this. I think there are three variants `--debug`, `-debug`, and `early(debug)`; see also `isEarlyCommand` partition in the above.

bitloi: I fixed it.

docs/fix-4979-manual-verification.md
eed3si9n, line 1: Please remove this file

bitloi: Sorry for the mistake 🙏

eed3si9n (commented):
Thanks for the contribution. Overall lgtm. Could you paste the result of manual testing?

bitloi:
> Thanks for the contribution. Overall lgtm. Could you paste the result of manual testing?

```
[debug] [zinc] IncrementalCompile -----------
[debug] IncrementalCompile.incrementalCompile
[debug] previous = Stamps for: 40 products, 13 sources, 33 libraries
[debug] current source = Set(${BASE}/project/Utils.scala, ...)
[debug] > initialChanges = InitialChanges(...)
[debug] No changes
[debug] Created transactional ClassFileManager with tempDir = ...
[debug] Copy resource mappings: ...
```

eed3si9n (approved):
Thanks!

PR #8820, opened 2026-02-26
[2.x] fix: IgnoreSourceChanges now works

### LINKED ISSUE #6773: `Global / onChangedBuildSource := IgnoreSourceChanges` has no effect
## steps

I have the following line at the top of my `build.sbt`:
```
Global / onChangedBuildSource := IgnoreSourceChanges`
```
1. I have a running sbt session (sbt `1.6.1`)
2. I change something in my `build.sbt`
3. I run `compile` in the sbt session

## problem

I see this warning, even though I have disabled it with the suggested setting:

```
[warn] build source files have changed
[warn] modified files:
[warn]   /[...]/build.sbt
[warn] Apply these changes by running `reload`.
[warn] Automatically reload the build when source changes are detected by setting `Global / onChangedBuildSource := ReloadOnSourceChanges`.
[warn] Disable this warning by setting `Global / onChangedBuildSource := IgnoreSourceChanges`.
```

## expectation

The warning should not be displayed at all.

seroperson:
Previously `onChangedBuildSource := IgnoreSourceChanges` was literally (mostly) ignored. 

- Added the explicit if-branch in [CheckBuildSources.scala](https://github.com/sbt/sbt/compare/develop...seroperson:i6773-on-changed-build-source?expand=1#diff-ee86f34af3045ee8ab45c7af5bad5ac4632231d5069e47cbfd9994b815b4906e) to ignore changes if `IgnoreSourceChanges` is set.
- Minor [Defaults.scala](https://github.com/sbt/sbt/compare/develop...seroperson:i6773-on-changed-build-source?expand=1#diff-6d9589bfb3f1247d2eace99bab7e928590337680d1aebd087d9da286586fba77) edit to not include root `*.sbt` files if `onChangedBuildSource` set to `IgnoreSourceChanges`. There is already the same check in `CheckBuildSources.buildSourceFileInputs` logic, which also seems like covers that glob too, so I even unsure whether this append is necessary at all.
- Added scripted test to check if the warning was logged.

Closes #6773

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -555,7 +555,12 @@ object Defaults extends BuildCommon {
     // Adds subproject build.sbt files to the global list of build files to monitor
     Scope.Global / checkBuildSources / pollInterval :==
       new FiniteDuration(Int.MinValue, TimeUnit.MILLISECONDS),
-    Scope.Global / checkBuildSources / fileInputs += baseDirectory.value.toGlob / "*.sbt",
+    Scope.Global / checkBuildSources / fileInputs ++= {
+      if ((Scope.Global / onChangedBuildSource).value != IgnoreSourceChanges)
+        Seq(baseDirectory.value.toGlob / "*.sbt")
+      else
+        Nil
+    },
   )
 
   lazy val configPaths = sourceConfigPaths ++ resourceConfigPaths ++ outputConfigPaths
--- a/main/src/main/scala/sbt/nio/CheckBuildSources.scala
+++ b/main/src/main/scala/sbt/nio/CheckBuildSources.scala
@@ -66,7 +66,7 @@ private[sbt] class CheckBuildSources extends AutoCloseable {
   private def reset(state: State): Unit = {
     val extracted = Project.extract(state)
     val interval = extracted.get(checkBuildSources / pollInterval)
-    val newSources = extracted.get(Global / checkBuildSources / fileInputs)
+    val newSources = extracted.get(Global / checkBuildSources / fileInputs).distinct
     if (interval >= 0.seconds || "polling" == SysProp.watchMode) {
       Option(repository.getAndSet(null)).foreach(_.close())
       pollingPeriod.set(interval)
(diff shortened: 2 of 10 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -557,7 +555,12 @@ object Defaults extends BuildCommon {
     // Adds subproject build.sbt files to the global list of build files to monitor
     Scope.Global / checkBuildSources / pollInterval :==
       new FiniteDuration(Int.MinValue, TimeUnit.MILLISECONDS),
-    Scope.Global / checkBuildSources / fileInputs += baseDirectory.value.toGlob / "*.sbt",
+    Scope.Global / checkBuildSources / fileInputs ++= {
+      if ((Scope.Global / onChangedBuildSource).value != IgnoreSourceChanges)
+        Seq(baseDirectory.value.toGlob / "*.sbt")
+      else
+        Nil
+    },
```
eed3si9n, line 563: > There is already the same check in CheckBuildSources.buildSourceFileInputs logic, which also seems like covers that glob too, so I even unsure whether this append is necessary at all.

Since you're making this change, could you check if we can remove it then? Thanks!

eed3si9n (changes requested):

main/src/main/scala/sbt/nio/CheckBuildSources.scala
```diff
@@ -148,6 +148,8 @@ private[sbt] class CheckBuildSources extends AutoCloseable {
               case Left(l)  => l.info(msg)
             }
             true
+          } else if (onChanges == IgnoreSourceChanges) {
+            false
```
eed3si9n: Thanks for fixing this!

main/src/main/scala/sbt/Defaults.scala
seroperson: Well, turns out that it's necessary to make subproject polling work. I edited this new scripted test to cover this case and also added `.distinct` where this `fileInputs` read, as currently it duplicates root `.sbt` glob.

eed3si9n (approved):
Thanks!

PR #8822, opened 2026-02-26
[2.x] perf: sbt --version

### LINKED ISSUE #5117: sbt --version is slow
## steps

```
$ sbt --version
```

## problem

It takes a long time to print out the version number.

## expectation

It should be much faster.

## notes

Currently the entire build is loaded just to print out the version number, but the launcher knows what sbt core artifact version it _will_ resolve to so we can short circuit and print that number instead _without_ loading the build.

Back in the days when launcher.jar was the sbt version and we all wrote our own bash scripts, launcher did handle `--version`. See https://github.com/sbt/launcher/pull/2.

See also https://github.com/sbt/sbt-launcher-package/pull/287

bitloi:
## Related issue
- Fixes https://github.com/sbt/sbt/issues/5117

## Problem
Running `sbt --version` is slow because launcher entrypoints invoke the JVM with `sbtVersion`, which boots the launcher/app path just to print version information.

The issue requests a short-circuit path that prints version output without loading the build when possible.

## Reproduction (before fix)
In an sbt project directory:

```bash
sbt --version
```

Observed behavior: the launcher path executes `... -jar ... sbtVersion` for `--version`, which is slower than necessary.

## Solution
Short-circuit `--version` in all launcher entrypoints by reading `sbt.version` from `project/build.properties`:

- `sbt` (bash)
  - Added `projectSbtVersion()` helper.
  - `--version` now prints `sbt version in this project: <version>` from parsed `build_props_sbt_version` instead of invoking `sbtVersion`.
- `launcher-package/src/universal/bin/sbt.bat`
  - Added `:set_sbt_version_from_build_props`.
  - `--version` prefers `build_props_sbt_version`; falls back to previous logic only when needed.
- `sbtw/src/main/scala/sbtw/Main.scala`
  - `opts.version` now uses `ConfigLoader.sbtVersionFromBuildProperties(cwd)` and avoids JVM launch.
  - Added typed helpers `projectSbtVersion(cwd: File): Option[String]` and `normalizeVersion(version: String): Option[String]`.

`--numeric-version` behavior remains unchanged and still uses the existing launcher resolution path.

## Why this approach
- Keeps output format stable (`sbt version in this project`, `sbt runner version`).
- Avoids unnecessary launcher startup for `--version`.
- Keeps fallback behavior where applicable.
- Limits scope to issue #5117 and avoids unrelated refactors.

## Tests
Updated integration assertions:

- `launcher-package/integration-test/src/test/scala/RunnerScriptTest.scala`
  - Tightened `assertVersionOutput` to require the formatted project version line.
  - Added assertion that output does not contain raw `sbtVersion` token.

Executed:

```bash
JAVA_HOME="/usr/lib/jvm/java-17-openjdk-amd64" \
  ./sbt "launcherPackageIntegrationTest/testOnly example.test.RunnerScriptTest -- -q --tests *version*"
```

Result: passed (`Total 32, Failed 0`).

Also checked IDE lints for changed Scala files: no linter diagnostics reported.

## Manual verification
Verified the changed paths now print project version from `project/build.properties` for `--version` and no longer rely on the `sbtVersion` command output path in these code paths.

## PR checklist
- [x] Problem reproduced before changes
- [x] Change is focused on one issue (#5117)
- [x] Tests included/updated
- [x] Change compiles in targeted test run
- [x] Followed project coding style

### DIFF
--- a/launcher-package/integration-test/src/test/scala/RunnerScriptTest.scala
+++ b/launcher-package/integration-test/src/test/scala/RunnerScriptTest.scala
@@ -12,11 +12,9 @@ abstract class RunnerScriptTest extends verify.BasicTestSuite with ShellScriptUt
   private def assertVersionOutput(out: List[String]): Unit =
     val lines =
       out.mkString(System.lineSeparator()).linesIterator.map(_.stripPrefix("[0J").trim).toList
-    assert(
-      lines.exists(_.matches("^sbt version in this project: " + versionPattern + "\\r?$")) ||
-        lines.contains("sbtVersion")
-    )
+    assert(lines.exists(_.matches("^sbt version in this project: " + versionPattern + "\\r?$")))
     assert(lines.exists(_.matches("^sbt runner version: " + versionPattern + "\\r?$")))
+    assert(!lines.exists(_.contains("sbtVersion")))
     assert(!lines.exists(_.contains("failed to connect to server")))
 
   testOutput("sbt -no-colors")("compile", "-no-colors", "-v"): (out: List[String]) =>
--- a/launcher-package/src/universal/bin/sbt.bat
+++ b/launcher-package/src/universal/bin/sbt.bat
@@ -729,8 +729,12 @@ if !sbt_args_print_sbt_version! equ 1 (
 
 if !sbt_args_print_version! equ 1 (
   if !is_this_dir_sbt! equ 1 (
-    call :set_sbt_version
-    echo sbt version in this project: !sbt_version!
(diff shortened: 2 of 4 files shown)


eed3si9n (changes requested):
Since the default mode of run is client-side run, I've closed the linked ticket.

bitloi:
@eed3si9n I implemented other issue #5117 and updated the pr description.

eed3si9n (approved):
Thanks!

PR #8824, opened 2026-02-27
[2.x] feat: Support doc / skip := true

### LINKED ISSUE #5525: make skip in doc := true
I would like to have an ability to skip generating _doc_ for aggregate project:
```
  .aggregate(getopt.js, getopt.jvm)
  .settings( skip in compile := true )
  .settings( skip in publish := true )
  .settings( skip in doc := true )
```
or maybe we can try go in different direction by inventing _pure aggregate_ projects which can be used for root project without need to skip everything.

eureka0928:
Fixes #5525

## Summary
Support `doc / skip := true` to allow skipping doc generation for aggregate or other projects.

## Problem
Setting `doc / skip := true` had no effect because the `doc` task in `Defaults.scala` did not check the `skip` setting. Only `compile`, `update`, `publish`, and `publishLocal` respected it. Users of aggregate root projects had no way to skip doc generation without workarounds.

## Solution
- Wrap the doc task body in `docTaskSettings` with a `skip.value` check. When skip is true, the task logs a debug message and returns the output directory without running scaladoc/javadoc.
- Update the `skip` key description in `Keys.scala` to include `doc` in the list of supported tasks.

## Test plan
- [ ] `sbt compile` passes
- [ ] `sbt scalafmtCheck` passes
- [ ] `sbt mimaReportBinaryIssues` passes
- [ ] Verify `doc / skip := true` skips doc generation in a test project

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -117,7 +117,6 @@ import xsbti.compile.{
   Compilers,
   DefinesClass,
   IncOptions,
-  IncToolOptionsUtil,
   Inputs,
   MiniSetup,
   PerClasspathEntryLookup,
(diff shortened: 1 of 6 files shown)


fury0928:
## Test results (verified with `publishLocalBin` + test project)

**Test project `build.sbt`:**
```scala
scalaVersion := "3.7.4"
doc / skip := true
```

### Before (develop branch) — doc runs, HTML generated:
```
> doc
[info] [test-doc-noskip] compiling 1 Scala source to .../classes ...
[info] [test-doc-noskip] done compiling
[success] elapsed time: 2 s, cache 80%, 12 disk cache hits, 3 onsite tasks
```
```
$ find target -name "*.html"
target/out/jvm/scala-3.7.4/test-doc-noskip/api/index.html
target/out/jvm/scala-3.7.4/test-doc-noskip/api/$lessempty$greater$/Foo$.html
```

### After (this branch) with `doc / skip := true` — doc skipped, no HTML:
```
> doc
[info] Skipping doc for test-doc-skip4
[success] elapsed time: 0 s, cache 93%, 15 disk cache hits, 1 onsite task
```
```
$ find target -name "*.html"
(no output — no doc files generated)
```

### After (this branch) without `doc / skip` set — doc still works normally:
```
> doc
[info] [test-doc-noskip] compiling 1 Scala source to .../classes ...
[info] [test-doc-noskip] done compiling
[success] elapsed time: 2 s, cache 80%, 12 disk cache hits, 3 onsite tasks
```
```
$ find target -name "*.html"
target/out/jvm/scala-3.7.4/test-doc-noskip/api/index.html
target/out/jvm/scala-3.7.4/test-doc-noskip/api/$lessempty$greater$/Foo$.html
```

fury0928:
@eed3si9n would you review this PR?

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2046,73 +2046,76 @@ object Defaults extends BuildCommon {
         },
         (TaskZero / key) := Def.uncached {
```
eed3si9n: 1. Please move the task implementation to `main/src/main/scala/sbt/internal/Compiler.scala`, and likely what you'd need to do is refactor it into multiple tasks, with the top level `docTask` using `Def.taskIf` similar to https://github.com/sbt/sbt/blob/2ee7c26d1e191630d09337bc6c7bf0f815ea8542/main/src/main/scala/sbt/internal/Compiler.scala#L302-L311. This will ensure that the `else` side doesn't evaluate. Otherwise, normal task will evaluate both sides of an if-expression.
2. Please includes a scripted test.

eed3si9n (changes requested):
Thanks for the contribution.

fury0928:
Thanks for the review @eed3si9n. Updated in 0517aaa:

1. **Moved doc task to `Compiler.scala`** using `Def.taskIf` — when `skip` is true, the compilers, sources, and other heavy dependencies are not evaluated. The `Defaults.scala` now just wires it with `Def.uncached(Compiler.docTask(key).value)`.

2. **Added scripted test** at `sbt-app/src/sbt-test/project/skip-doc/` — verifies that `doc / skip := true` succeeds without generating doc output.

eed3si9n (approved):
Thanks!

fury0928:
@eed3si9n CI is passed

PR #8830, opened 2026-02-28
[2.x] fix: Backtick-quote project IDs in classloader error messages

### LINKED ISSUE #5803: sbt error message shows example command that is syntactically wrong
## steps

Unpack the attached minimal project, start sbt and execute the task `test`.

[sbt-bug-report-identifiers.tar.gz](https://github.com/sbt/sbt/files/5145052/sbt-bug-report-identifiers.tar.gz)

## problem

Sbt shows a helpful error message with a command to fix the issue. But that command is syntactically incorrect, if a sub module contains e.g. a dash. Then backticks are required. 

````
[error] Test suite net.gaillourdet.bugReport.BugReportSpec failed with java.lang.ClassNotFoundException: net.gaillourdet.bugReport.Simple.
[error] This may be due to the ClassLoaderLayeringStrategy (ScalaLibrary) used by your task.
[error] To improve performance and reduce memory, sbt attempts to cache the class loaders used to load the project dependencies.
[error] The project class files are loaded in a separate class loader that is created for each test run.
[error] The test class loader accesses the project dependency classes using the cached project dependency classloader.
[error] With this approach, class loading may fail under the following conditions:
[error] 
[error]  * Dependencies use reflection to access classes in your project's classpath.
[error]    Java serialization/deserialization may cause this.
[error]  * An open package is accessed across layers. If the project's classes access or extend
[error]    jvm package private classes defined in a project dependency, it may cause an IllegalAccessError
[error]    because the jvm enforces package private at the classloader level.
[error] 
[error] These issues, along with others that were not enumerated above, may be resolved by changing the class loader layering strategy.
[error] The Flat and ScalaLibrary strategies bundle the full project classpath in the same class loader.
[error] To use one of these strategies, set the  ClassLoaderLayeringStrategy key
[error] in your configuration, for example:
[error] 
[error] set bug-report / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.ScalaLibrary
[error] set bug-report / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
````

## expectation

Command examples or code snippets should quote identifiers if necessary. So that they can be copied and tried without further editing. In this case the last two lines should be:

````
[error] set `bug-report` / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.ScalaLibrary
[error] set `bug-report` / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
````
Then, the last line is exactly the needed solution to the stated problem. 

## notes

 This can be reproduced with sbt 1.3.13 as well as 1.4.0-M2. 

I am not sure how many other error messages contain generated code snippets which should be fixed as well.

### PR DESCRIPTION
**Pull request by eureka0928**
Fixes #5803

## Problem
When a test fails with a `ClassNotFoundException` due to `ClassLoaderLayeringStrategy`, sbt suggests a `set` command to fix it. But if the project ID contains characters like hyphens (e.g. `bug-report`), the suggested command is syntactically invalid:

```
[error] set bug-report / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
```

This fails because `bug-report` is not a valid Scala identifier — it needs backtick quoting.

## Solution
- Added `Util.quoteIfNotScalaId(s: String)` in `internal/util-core` that wraps a string in backticks if it is not a valid Scala identifier (contains hyphens, dots, starts with digits) or is a Scala keyword.
- Applied `quoteIfNotScalaId` to the `projectId` parameter in `Defaults.scala` where the classloader error message is constructed (two call sites in `allTestGroupsTask`).

The error message now correctly suggests:
```
[error] set `bug-report` / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
```

## Changes
- `internal/util-core/.../Util.scala`: Added `quoteIfNotScalaId` and `isValidScalaId` helper
- `main/.../Defaults.scala`: Wrapped `thisProj.id` with `quoteIfNotScalaId` in both test task sites
- `build.sbt`: Added hedgehog test dependency to `utilCore`

## Tests
- **Unit tests** (`internal/util-core/.../UtilSpec.scala`): 7 cases covering simple identifiers, underscores, hyphens, dots, digit-leading, keywords, and empty strings — all passing

## Verification
- [x] `sbt compile` passes
- [x] `sbt utilCore/test` passes (7/7)
- [x] `sbt scalafmtCheck` passes
- [x] `sbt mimaReportBinaryIssues` passes

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -311,6 +311,7 @@ lazy val utilCore = project
     utilCommonSettings,
     name := "Util Core",
     Utils.keywordsSettings,
+    libraryDependencies ++= Seq(hedgehog % Test),
     mimaSettings,
     mimaBinaryIssueFilters ++= Seq(
     ),
--- a/internal/util-core/src/main/scala/sbt/internal/util/Util.scala
+++ b/internal/util-core/src/main/scala/sbt/internal/util/Util.scala
@@ -45,6 +45,14 @@ object Util:
 
   def quoteIfKeyword(s: String): String = if (ScalaKeywords.values(s)) s"`${s}`" else s
 
+  def quoteIfNotScalaId(s: String): String =
+    if isValidScalaId(s) && !ScalaKeywords.values(s) then s
+    else s"`$s`"
+
+  private def isValidScalaId(s: String): Boolean =
+    s.nonEmpty && (s.charAt(0).isLetter || s.charAt(0) == '_') &&
+      s.forall(c => c.isLetterOrDigit || c == '_')
+
   def ignoreResult[A](f: => A): Unit = {
     val _ = f
     ()
--- a/internal/util-core/src/test/scala/sbt/internal/util/UtilSpec.scala
+++ b/internal/util-core/src/test/scala/sbt/internal/util/UtilSpec.scala
@@ -1,0 +1,49 @@
+/*
+ * sbt
+ * Copyright 2023, Scala center
+ * Copyright 2011 - 2022, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.internal.util
+
+import hedgehog.*
+import hedgehog.runner.*
+
+object UtilSpec extends Properties:
+  override def tests: List[Test] = List(
+    example(
+      "quoteIfNotScalaId should not quote simple identifiers",
+      assertQuote("foo", "foo"),
+    ),
+    example(
+      "quoteIfNotScalaId should not quote identifiers with underscores",
+      assertQuote("foo_bar", "foo_bar"),
+    ),
+    example(
+      "quoteIfNotScalaId should quote identifiers with hyphens",
+      assertQuote("bug-report", "`bug-report`"),
+    ),
+    example(
+      "quoteIfNotScalaId should quote identifiers with dots",
+      assertQuote("my.project", "`my.project`"),
+    ),
+    example(
+      "quoteIfNotScalaId should quote identifiers starting with digits",
+      assertQuote("123abc", "`123abc`"),
+    ),
+    example(
+      "quoteIfNotScalaId should quote Scala keywords",
+      assertQuote("class", "`class`"),
+    ),
+    example(
+      "quoteIfNotScalaId should quote empty strings",
+      assertQuote("", "``"),
+    ),
+  )
+
+  private def assertQuote(input: String, expected: String): Result =
+    val actual = Util.quoteIfNotScalaId(input)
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Comment by fury0928** [comment]
## Before / After

### Reproduction

Using @eed3si9n's reproduction setup:

```scala
LocalRootProject / name := "sbt-bug-report-identifiers"
version := "0.1"
scalaVersion := "2.13.18"

lazy val base = Projects.base
lazy val `bug-report` = Projects.`bug-report`
```

When a test fails with `ClassNotFoundException` due to `ClassLoaderLayeringStrategy`, sbt formats the `projectId` parameter as `s"${thisProj.id} / "` and embeds it in the error message.

### Before (develop)

For a project with `id = "bug-report"`, the error message produces:

```
[error] set bug-report / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.ScalaLibrary
[error] set bug-report / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
```

This is **syntactically invalid** — `bug-report` parses as subtraction (`bug` minus `report`).

### After (this PR)

The `projectId` is now formatted as `s"${Util.quoteIfNotScalaId(thisProj.id)} / "`, which backtick-quotes identifiers that aren't valid Scala IDs:

```
[error] set `bug-report` / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.ScalaLibrary
[error] set `bug-report` / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
```

Projects with simple names (e.g. `myProject`) are **unchanged** — no backticks added.

### Unit test results (`utilCore/testOnly sbt.internal.util.UtilSpec`)

```
+ quoteIfNotScalaId should not quote simple identifiers: OK, passed 1 tests
+ quoteIfNotScalaId should not quote identifiers with underscores: OK, passed 1 tests
+ quoteIfNotScalaId should quote identifiers with hyphens: OK, passed 1 tests
+ quoteIfNotScalaId should quote identifiers with dots: OK, passed 1 tests
+ quoteIfNotScalaId should quote identifiers starting with digits: OK, passed 1 tests
+ quoteIfNotScalaId should quote Scala keywords: OK, passed 1 tests
+ quoteIfNotScalaId should quote empty strings: OK, passed 1 tests
Passed: Total 7, Failed 0, Errors 0, Passed 7
```

### Note on `set` command with backtick-quoted project IDs

The `set` command in sbt 2.x currently cannot resolve backtick-quoted project references due to Scala 3 compilation handling (`bug$minusreport` encoding issue). This is a separate issue from the error message formatting fixed here. The error message now shows **syntactically valid Scala**, which is the correct fix for #5803.

eed3si9n (approved):
Thanks!

PR #8840, opened 2026-03-01
[2.x] test: Enable DagSpecification.scala

xuwei-k:
move `util-collection/src/test/{scala-2 => scala}/DagSpecification.scala`


eed3si9n (approved):
Thanks!

PR #8862, opened 2026-03-02
[2.x] fix: Undefined per-project settings order

### LINKED ISSUE #7173: Adding subproject causes setting to no longer be evaluated (?)
## steps

sbt 1.8.2

```scala
val foo = project.in(file("foo"))
val fooAux = project.in(file("foo-aux")).settings(foo / Compile / sourceGenerators += Def.task(Seq.empty))
val bar = project
val baz = project
```

## problem

Project `foo` has no source generators.

```
sbt:sandbox> reload; show Compile/sourceGenerators
[info] welcome to sbt 1.8.2 (Azul Systems, Inc. Java 11.0.17)
[info] loading project definition from /workspace/sandbox/project
[info] loading settings for project sandbox from build.sbt ...
[info] set current project to sandbox (in build file:/workspace/sandbox/)
[info] foo / Compile / sourceGenerators
[info]  List()
[info] fooAux / Compile / sourceGenerators
[info]  List()
[info] bar / Compile / sourceGenerators
[info]  List()
[info] baz / Compile / sourceGenerators
[info]  List()
[info] Compile / sourceGenerators
[info]  List()
```

## expectation

If we remove the `baz` project from the build, then `foo` does have a source generator. This is what I expect.

```
sbt:sandbox> reload; show Compile/sourceGenerators
[info] welcome to sbt 1.8.2 (Azul Systems, Inc. Java 11.0.17)
[info] loading project definition from /workspace/sandbox/project
[info] loading settings for project sandbox from build.sbt ...
[info] set current project to sandbox (in build file:/workspace/sandbox/)
[info] foo / Compile / sourceGenerators
[info]  List(Task(_))
[info] fooAux / Compile / sourceGenerators
[info]  List()
[info] bar / Compile / sourceGenerators
[info]  List()
[info] Compile / sourceGenerators
[info]  List()
```

seroperson:
Previously, as `build.defined` is a map with undefined order, `projectSettings` could have a default "empty" initializer go after a cross-project appender. Whether it would work or not depended on this internal map order randomness, so, for example (see an added scripted test), defining the `lazy val baz = project` would cause an empty `sourceGenerators`, but defining the `lazy val ewikqelkqweqopweo = project` would work as intended.

This fix eliminates this randomness, so cross-project settings always go after own-scoped or global settings. Should not affect anything, as previously order was undefined in such cases.

Closes #7173

### DIFF
--- a/main/src/main/scala/sbt/internal/Load.scala
+++ b/main/src/main/scala/sbt/internal/Load.scala
@@ -427,17 +427,31 @@ private[sbt] object Load {
       inScope(GlobalScope)(loaded.autos.globalSettings) ++
       loaded.units.toSeq.flatMap { (uri, build) =>
         val pluginBuildSettings = loaded.autos.buildSettings(uri)
-        val projectSettings = build.defined flatMap { (id, project) =>
-          val ref = ProjectRef(uri, id)
-          val defineConfig: Seq[Setting[?]] =
-            for (c <- project.configurations)
-              yield ((ref / ConfigKey(c.name) / configuration) :== c)
-          val builtin: Seq[Setting[?]] =
-            (thisProject :== project) +: (thisProjectRef :== ref) +: defineConfig
-          val settings = builtin ++ injectSettings.project ++ project.settings
-          // map This to thisScope, Select(p) to mapRef(uri, rootProject, p)
-          transformSettings(projectScope(ref), uri, rootProject, settings)
-        }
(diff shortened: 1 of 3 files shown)


eed3si9n:
I'll get back to the actual review when I have time to read the code a bit, but thanks for working on this issue!

eed3si9n (approved):
Thanks!

PR #8867, opened 2026-03-03
[2.x] fix: Update help text for ++

### LINKED ISSUE #6988: Outdated help text for `++`
## steps

using sbt 1.7.1:

```
sbt:root> help ++
++ <scala-version>[!] [-v] [<command>]
	Changes the Scala version and runs a command.

	Sets the `scalaVersion` of all projects that define a Scala cross version that is binary
  compatible with <scala-version> and reloads the build.  If ! is supplied, then the
  version is forced on all projects regardless of whether they are binary compatible or
  not.

  If -v is supplied, verbose logging of the Scala version switching is done.

	If <command> is provided, it is then executed.

++ [<scala-version>=]<scala-home>[!] [-v] [<command>]
	Uses the Scala installation at <scala-home> by configuring the scalaHome setting for
	all projects.

	If <scala-version> is specified, it is used to select the value of the scalaVersion setting
  from one of the values of crossScalaVersions setting.
	This is important when using managed dependencies.  This version will determine the
	cross-version used as well as transitive dependencies.

  <scala-version> may be an actual Scala version such as 3.1.3, or a Semantic Version selector
  pattern such as 2.13.x. Only subprojects that are listed to match the version pattern
  have their Scala version switched.  If ! is supplied, then all projects projects have
  their Scala version switched.

  If -v is supplied, verbose logging of the Scala version switching is done.

	If <command> is provided, it is then executed. For example:

    ++ 2.13.x test

  The above will run test on all subprojects that contain a 2.13 Scala version.

	See also `help +`
```

## problem

The help text mentions filtering on binary compatibility when using `++`.

## expectation

IIUC, the binary compatibility check was removed in #6946, I would expect the help text to reflect that (as it does in a later paragraph).

oolokioo7:
Closes #6988 

## Problem
The help text still mentions filtering projects by **Scala binary compatibility** even the behavior was removed previously.

## Solution
Update the `++` command detailed help text to reflect current behavior.

## Changes
- the help text in the `main/src/main/scala/sbt/internal/CommandStrings.scala`

## Test
- sbt compile
- sbt mainProj/text
- sbt scalafmtCheck
- sbt mimaReportBinaryIssues

## Notes
As this PR only updates text, it doesn't change ++ behavior.

### DIFF
--- a/launcher-package/src/universal/bin/sbt.bat
+++ b/launcher-package/src/universal/bin/sbt.bat
@@ -153,7 +153,7 @@ shift
 if "%~0" == "" goto args_end
 set g=%~0
 
-rem make sure the sbt_args_debug gets set first incase any argument parsing uses :dlog
+rem make sure the sbt_args_debug gets set first in case any argument parsing uses :dlog
 if "%~0" == "-d" set _debug_arg=true
 if "%~0" == "--debug" set _debug_arg=true
 
--- a/main/src/main/scala/sbt/internal/Continuous.scala
+++ b/main/src/main/scala/sbt/internal/Continuous.scala
@@ -497,7 +497,7 @@ private[sbt] object Continuous {
              * are other changes detected in the burst. As soon as no changes are detected
              * during the polling window, we return all of the detected events. The polling
              * period is by default 5 milliseconds which is short enough to detect bursts
-             * induced by commands like git rebase but fast enough to not lead to a noticable
+             * induced by commands like git rebase but fast enough to not lead to a noticeable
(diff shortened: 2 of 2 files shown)


eed3si9n (changes requested):
I really don't want to encourage typo fixes from Gittensor miners, so I'll close this PR.

oolokioo7:
I have updated this PR to fix the existing issue.

- doesn't affect the performance
- all CI plans were passed

plz reopen this PR and review it if you have a time. @eed3si9n

PR #8873, opened 2026-03-04
[2.x] refactor: Extract lm-ivy to separate sbt-ivy plugin module

### LINKED ISSUE #7640: [2.x] Move Ivy library management to a plugin
Given the default status of Coursier, we should try to remove Ivy out of the dependency graph, and turn it into a community plugin. Potentially this is a good opportunity to bump Apache Ivy version as well.

eureka0928:
## Summary

Step 5 of #7640 — removes the compile-time dependency on `lm-ivy` from `main/` by creating a standalone `sbt-ivy/` plugin module.

- Create new `sbt-ivy/` subproject with `IvyDependencyPlugin` AutoPlugin that provides all Ivy-specific functionality (ivySbt, ivyModule, ivyConfiguration, publisher, projectDescriptors, deliver/makeIvyXml)
- Move `IvyXml.scala` from `main/` to `sbt-ivy/`
- Extract credential-loading utilities (`forHost`, `allDirect`, `loadCredentials`) to `CredentialUtils` in `lm-core` (no Ivy dependency)
- Move classifier exclude/include logic (`extractExcludes`, `addExcluded`) to `UpdateClassifiersUtil` in `lm-core`
- Remove `lm-ivy` from compile dependencies of `mainProj` and `buildFileProj`
- Replace Ivy type references in `main/` with `Any`-typed keys + runtime casts in Ivy-specific code paths
- Add reflective discovery of `IvyDependencyPlugin` in `PluginDiscovery`
- Wire `sbt-app` to depend on `sbt-ivy` so Ivy functionality is still bundled by default

Closes #7640

## Test plan

- [x] `sbt compile` — full build compiles, `mainProj` compiles without `lm-ivy`
- [x] `sbt test` — all 126 unit tests pass
- [x] `sbt mimaReportBinaryIssues` — binary compatibility check passes
- [x] `sbt publishLocalBin` — local publishing works
- [x] `scripted dependency-graph/*` — all 11 dependency graph tests pass
- [x] `scripted dependency-management/cache-resolver` — passes
- [x] `scripted lm-coursier/simple` — passes


fury0928:
Hi @eed3si9n would you review this PR?
I have another following PR (new issue you created)

build.sbt
```diff
@@ -737,16 +737,33 @@ lazy val mainProj = (project in file("main"))
...
+    mimaPreviousArtifacts := Set.empty, // new module, no previous artifacts
+  )
+  .configure(addSbtIO)
+
 // Strictly for bringing implicits and aliases from subsystems into the top-level sbt namespace through a single package object
 //  technically, we need a dependency on all of mainProj's dependencies, but we don't do that since this is strictly an integration project
 //  with the sole purpose of providing certain identifiers without qualification (with a package object)
 lazy val sbtProj = (project in file("sbt-app"))
-  .dependsOn(mainProj)
+  .dependsOn(mainProj, sbtIvyProj)
```
eed3si9n: Could we remove `sbtIvyProj` dependency here please?

fury0928: @eed3si9n Done — removed the `sbtIvyProj` dependency from `sbtProj`. Confirmed `sbt-app` has no compile-time references to `sbt-ivy`, so it compiles cleanly. `sbtIvyProj` is still built and published via the `allProjects` aggregation.

fury0928: @eed3si9n After looking into this further, removing the `sbtIvyProj` dependency from `sbtProj` causes CI failures beyond just Ivy tests — multiple `lm-coursier/*` scripted tests also break.

The root cause is that `IvyDependencyPlugin` provides settings that Coursier's resolution depends on:

```scala
// In IvyDependencyPlugin:
projectDescriptors := Def.uncached(depMap.value.asInstanceOf[Map[Any, Any]])
csrExtraProjects := Def.uncached(coursierExtraProjectsTask.value)
```

`csrExtraProjects` feeds into Coursier's project resolver and chains through `projectDescriptors` → `depMap` → `ivyModule`. Without `IvyDependencyPlugin` loaded, these stay as stubs (`Unit`/`None`), causing `ClassCastException` in both Ivy and Coursier workflows:

- `dependency-graph/cachedResolution`: `None$ cannot be cast to UpdateOptions`
- `lm-coursier/scala-sources-javadoc-jars`: `BoxedUnit cannot be cast to ModuleDescriptor`
- Several other `lm-coursier/*` tests fail or go `[PENDING]`

To cleanly remove this dependency, the Coursier-facing settings (`csrExtraProjects`, `projectDescriptors`, `projectResolver`) would need to be extracted out of `IvyDependencyPlugin` into a module that always loads. I'll revert the removal for now — happy to tackle that refactoring as a follow-up if you'd like.

eed3si9n: I'm okay with removing `dependency-graph/cachedResolution`, which depended on Ivy internals.

However, `lm-coursier/scala-sources-javadoc-jars` (https://github.com/sbt/sbt/blob/develop/sbt-app/src/sbt-test/lm-coursier/scala-sources-javadoc-jars/build.sbt) not working seems concerning since it's just calling `updateClassifiers.value` as far as I can tell.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -4195,31 +4157,36 @@ object Classpaths {
...
-    }
-
   def projectResolverTask: Initialize[Task[Resolver]] =
-    projectDescriptors.map { m =>
-      val resolver = new ProjectResolver(ProjectResolver.InterProject, m)
-      new RawRepository(resolver, resolver.getName)
+    Def.task {
+      // Stub resolver — Coursier handles inter-project deps via csrInterProjectDependencies.
+      // Overridden by IvyDependencyPlugin with a real ProjectResolver.
+      new RawRepository(NoOpResolver, NoOpResolver.name)
```
eed3si9n, line 4166: I wonder if lm-core should provide an implementation of ProjectResolver that's backend independent, which we can later convert.

build.sbt
fury0928: @eed3si9n Good point — I traced the exact chain for `lm-coursier/scala-sources-javadoc-jars`:

1. `ivyModule` is `TaskKey[Any]` with stub default `(): Any` (`Defaults.scala:3304`)
2. `IvyDependencyPlugin` overrides it with a real `IvySbt.Module`
3. `updateClassifiers` (`LibraryManagement.scala:336`) depends on `ivyModule` and casts it at line 403: `im.asInstanceOf[ModuleDescriptor]`
4. Without `IvyDependencyPlugin` → `im` is `()` (BoxedUnit) → `ClassCastException`

So `updateClassifiers` depends on `ivyModule` which is only wired by `IvyDependencyPlugin`. The key is misnamed — it provides the `ModuleDescriptor` that both Ivy and Coursier resolvers need.

To properly remove the `sbtIvyProj` dependency, `ivyModule` (or a Coursier-facing equivalent) would need to be computed in `main/` without depending on `lm-ivy`. I'll revert the removal for now and keep `sbtIvyProj` in `sbtProj.dependsOn`. Happy to tackle that decoupling as a follow-up.

fury0928:
@eed3si9n Re: your suggestion on a backend-independent `ProjectResolver` in `lm-core` — I think that's the right direction.

Currently keys like `ivyModule`, `projectDescriptors`, and `projectResolver` are typed as `Any` because they hold Ivy-specific types (`IvySbt.Module`, `Map[ModuleRevisionId, ModuleDescriptor]`). This forces `main/` to depend on `lm-ivy` or use unsafe casts, and is the reason we can't remove `sbtIvyProj` from `sbtProj` yet — `updateClassifiers` needs `ivyModule` wired by `IvyDependencyPlugin` even for Coursier-only workflows.

A neutral `ProjectResolver` in `lm-core` would let:
- `main/` wire these keys with properly typed defaults
- `updateClassifiers` work without `IvyDependencyPlugin`
- Ivy and Coursier each convert from the neutral type downstream
- `sbtIvyProj` be cleanly removed from `sbtProj`

I'll revert the dependency removal in this PR to get CI green and keep the extraction as a standalone step. Happy to tackle the `lm-core` `ProjectResolver` abstraction as a separate follow-up PR — does that work for you?

fury0928:
What do you prefer? let me know
Thank you for your review and time

eed3si9n:
The success criteria behind https://github.com/sbt/sbt/issues/7640 is to remove Ivy out of the dependency graph by providing substitution for common operations. By plugin I meant that we'll eventually remove the code and move it to an independent repo. So whatever surgery we have to perform to regain `updateClassifiers` (I think this is used typically by IDEs for library source code), we should try it in this PR.

fury0928:
@eed3si9n finally I updated and passed all CI

build.sbt
```diff
@@ -737,16 +737,33 @@ lazy val mainProj = (project in file("main"))
...
+    mimaPreviousArtifacts := Set.empty, // new module, no previous artifacts
+  )
+  .configure(addSbtIO)
+
 // Strictly for bringing implicits and aliases from subsystems into the top-level sbt namespace through a single package object
 //  technically, we need a dependency on all of mainProj's dependencies, but we don't do that since this is strictly an integration project
 //  with the sole purpose of providing certain identifiers without qualification (with a package object)
 lazy val sbtProj = (project in file("sbt-app"))
-  .dependsOn(mainProj)
+  .dependsOn(mainProj, sbtIvyProj % Runtime)
```
eed3si9n: First off thanks for working on this!

However, if this is included into `Runtime`, and

```scala
object IvyDependencyPlugin extends AutoPlugin:
  override def requires = IvyPlugin
```

the trigger condition for `IvyDependencyPlugin` is `allRequirements`, then it's effectively still part of all subprojects.

fury0928: @eed3si9n Good point — since `IvyDependencyPlugin` has `trigger = allRequirements` and `IvyPlugin` also auto-triggers, including `sbtIvyProj` at `Runtime` effectively keeps Ivy active for all projects.

I see a few options:

**Option A: Remove `sbtIvyProj` from `sbtProj` entirely** — don't include it even at Runtime. Ship sbt-ivy as a separate plugin that users add via `project/plugins.sbt` when they need Ivy. Scripted tests that need `useIvy := true` would add it as a test plugin dependency. This is the cleanest decoupling.

**Option B: Change `IvyDependencyPlugin.trigger` to `noTrigger`** — keep the jar on the Runtime classpath but require explicit `enablePlugins(IvyDependencyPlugin)` to activate it. Simpler migration path but less clean.

**Option C: Guard with `useIvy`** — keep `allRequirements` but make `IvyDependencyPlugin.projectSettings` conditional on `useIvy.value`, producing empty settings when false. Least disruptive but doesn't remove Ivy from the dependency graph.

I think Option A aligns best with the goal of this PR. The `defaultPublisher` with `PomGenerator` already handles `makePom` without Ivy, and `updateClassifiers` is decoupled from `ivyModule`. The remaining question is what happens to scripted tests that set `useIvy := true` — they'd need `addSbtPlugin("org.scala-sbt" % "sbt-ivy" % ...)` in their test `project/plugins.sbt`.

Which direction would you prefer?

eed3si9n: Yea, I think Option A.

fury0928:
Hi @eed3si9n this has many testing errors... what is your suggestion to move step by step?

fury0928:
Hi @eed3si9n finally passed the CI, pls review and merge

eed3si9n:
Looks like we still have `sbtIvyProj % Runtime` dependency though. Also note that scripted tests run in batches, so likely we should kick out all Ivy-dependent scripted tests into a fresh new directory, which is invoked separately in the CI.

build.sbt
```diff
@@ -737,11 +737,28 @@ lazy val mainProj = (project in file("main"))
...
   )
-  .dependsOn(lmCore, lmIvy, lmCoursierShadedPublishing)
+  .dependsOn(lmCore, lmCoursierShadedPublishing)
   .configure(addSbtIO, addSbtCompilerInterface, addSbtZincCompileCore)
 
+lazy val sbtIvyProj = (project in file("sbt-ivy"))
+  .dependsOn(mainProj, lmIvy)
+  .settings(
+    testedBaseSettings,
+    name := "sbt-ivy",
```
eed3si9n, line 755:
```suggestion
    name := "sbt-ivy",
    sbtPlugin := true,
    pluginCrossBuild / sbtVersion := version.value,
```

sbt-app/src/sbt-test/ivy/build-deps/build.sbt
eed3si9n, line 1: Could you add `project/plugins.sbt` that looks like:

```scala
libraryDependencies += sbtPluginExtra(
  ModuleID("org.scala-sbt", "sbt-ivy", sbtVersion.value),
  sbtBinaryVersion.value,
  scalaBinaryVersion.value
)
```

fury0928:
Hi @eed3si9n can you help me here?
To split into or change direction? stuck now

eed3si9n:
Trying https://github.com/sbt/sbt/pull/8873/changes/e63b67dbe781ea6330256ed0a030dd1f0f68d2e3

fury0928:
Hi @eed3si9n this is passed, is this good to merge?

fury0928:
Follow up on this @eed3si9n I spent many effort on this PR

eed3si9n:
Yea, I'll review the details before landing it.

eed3si9n (approved):
Thanks!

xuwei-k:
- https://github.com/sbt/sbt/issues/9054

PR #8875, opened 2026-03-04
[2.x] fix: handle -sbt-dir with spaces from .sbtopts

### LINKED ISSUE #7197: Sbt options are treated differently
When I invoke in terminal 
`sbt -sbt-dir "/Users/a' dog" --verbose`
the output shows that this option is mapped to `"-Dsbt.global.base=/Users/a' dog"` 
But later when I will move this option from terminal to the `.sbtopts` file and invoke `sbt --verbose` the output shows that it is mapped to `-Dsbt.global.base=/Users/a'`

Whether this difference has some hidden purpose ? It introduces additional inconsistency 

SBT: sbt1.8.2
OS: macOs 12.13.1

Eruis2579:
fixes #7197

## Fix

Handle `-sbt-dir` values with spaces and quotes when they come from `.sbtopts`, so they behave the same as when passed on the command line.

### Problem

When you pass `-sbt-dir "/Users/a' dog"` on the command line, the launcher correctly produces:

`-Dsbt.global.base=/Users/a' dog`

But when you put the same option into `.sbtopts`, the launcher previously split the line on spaces without respecting quotes, so the resulting `-Dsbt.global.base` was truncated (for example, `-Dsbt.global.base=/Users/a'`). This made `.sbtopts` behavior inconsistent with CLI behavior and broke setups where the global sbt directory path contains spaces and an embedded quote.

### File changed

- `sbt`
  - Added a quote-aware `parseLineIntoWords()` helper and `appendSbtoptsFromFile()` to read `.sbtopts` files and populate `sbt_file_opts` without breaking paths that contain spaces and quotes.
  - Switched sbtopts loading from `sbt_file_opts+=($(loadConfigFile ...))` to using `appendSbtoptsFromFile` for machine, dist, and project `.sbtopts`, keeping precedence the same but fixing parsing.

- `launcher-package/integration-test/src/test/scala/RunnerScriptTest.scala`
  - Added an integration test `"sbt -sbt-dir with space and quote in .sbtopts"` that:
    - Writes project `.sbtopts` with `-sbt-dir "/Users/a' dog"`.
    - Runs the launcher with `-d -v`.
    - Asserts that the final command line contains a `-Dsbt.global.base=...` whose value includes the full `/Users/a' dog` path (space and quote preserved).

### Test

./sbt "launcherPackageIntegrationTest/testOnly example.test.RunnerScriptTest"

### DIFF
--- a/launcher-package/integration-test/src/test/scala/RunnerScriptTest.scala
+++ b/launcher-package/integration-test/src/test/scala/RunnerScriptTest.scala
@@ -253,29 +253,29 @@ abstract class RunnerScriptTest extends verify.BasicTestSuite with ShellScriptUt
       )
 
   // Test for issue #7197: -sbt-dir with spaces (and quotes) in .sbtopts
+  // windowsSupport = false: skip on Windows cmd, but runs on Git Bash (#8779)
   testOutput(
     "sbt -sbt-dir with space and quote in .sbtopts",
     sbtOptsFileContents = """-sbt-dir "/Users/a' dog"""",
+    windowsSupport = false,
   )("-d", "-v"): (out: List[String]) =>
-    if (isWindows) cancel("Test not supported on windows")
-    else
-      val cmdLineStart = out.indexWhere(_.contains("Executing command line"))
-      assert(cmdLineStart >= 0, "Command line section not found")
+    val cmdLineStart = out.indexWhere(_.contains("Executing command line"))
+    assert(cmdLineStart >= 0, "Command line section not found")
 
-      val cmdLine = out.drop(cmdLineStart + 1).takeWhile(!_.trim.isEmpty)
-      val globalBaseArgs =
-        cmdLine.filter(_.contains("Dsbt.global.base"))
+    val cmdLine = out.drop(cmdLineStart + 1).takeWhile(!_.trim.isEmpty)
+    val globalBaseArgs =
+      cmdLine.filter(_.contains("Dsbt.global.base"))
 
-      assert(
-        globalBaseArgs.nonEmpty,
-        s"-Dsbt.global.base should be present in command line. cmdLine=${cmdLine.mkString(", ")}"
-      )
-      assert(
-        globalBaseArgs.exists(arg =>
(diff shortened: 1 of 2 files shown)


Eruis2579:
Hi, @eed3si9n , Please review my PR

launcher-package/integration-test/src/test/scala/RunnerScriptTest.scala
```diff
@@ -252,6 +252,31 @@ abstract class RunnerScriptTest extends verify.BasicTestSuite with ShellScriptUt
         s"sbtopts options should appear before CLI memory settings. g1Index=$g1Index, xmxCliIndex=$xmxCliIndex"
       )
 
+  // Test for issue #7197: -sbt-dir with spaces (and quotes) in .sbtopts
+  testOutput(
+    "sbt -sbt-dir with space and quote in .sbtopts",
+    sbtOptsFileContents = """-sbt-dir "/Users/a' dog"""",
+  )("-d", "-v"): (out: List[String]) =>
+    if (isWindows) cancel("Test not supported on windows")
```
eed3si9n: Hi, thanks for the contribution! Please check out our [Contributor's Guildeline](https://github.com/sbt/sbt/blob/develop/CONTRIBUTING.md) and:
3. Follow our AI assisted contributions guideline (please disclose AI usages in the PR description, if any)
5. Generally we recommend getting CI to pass on a forked repo before sending a pull request (pull requests that do not pass tests will be closed after a few days)

Note that there was a similar pull request you sent https://github.com/sbt/sbt/pull/8730, but unfortunately, I had to revert it in https://github.com/sbt/sbt/pull/8797 because it seems to have caused https://github.com/sbt/sbt/issues/8767. Instead of isWindows cancel, could you pass to `testOutput(...)`, which should now test this on Git Bash on Windows (https://github.com/sbt/sbt/pull/8779)

```scala
    windowsSupport = false,
```

eed3si9n (changes requested):

eed3si9n (approved):
Thanks!

PR #8878, opened 2026-03-06
[2.x] dependencyTree: add `--append` for file output in multi-module builds

### LINKED ISSUE #6627: Append output of dependencyTree `toFile` for multi-module projects
Currently, there's an option to write the output of dependencyTree to a file using the `toFile` sub-command. This works fine for single-module projects. But for multi-module projects, this throws an error saying [`"Target file for dependencyTree already exists at <file_location>. Use '-f' to override"`](https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/plugins/DependencyTreeSettings.scala#L218).

If `-f/--force` is used, this overwrites the output in the existing file:
- https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/plugins/DependencyTreeSettings.scala#L273-L277
- https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/plugins/DependencyTreeSettings.scala#L163-L164

The data is written to file using the IO.write() method [function](https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/plugins/DependencyTreeSettings.scala#L221). There's an option to append to file using a parameter called `append: Boolean = false` that is not used as of now. 
If this parameter is used (eg. `IO.write(targetFile, data, IO.utf8, true)`), then it would enable fetching of dependencies of all modules in a multi-module project.

oolokioo7:
Closes #6627 

## Summary
It adds an `--append` flag to the `dependencyTree` input task when used with `--out`. It allows running `dependencyTree` in multiple subprojects and appending their outputs to the same target file instead of overwriting.

## Solution

- Add `Arg.Append` in the CLI parsing
- When `--append` is set, write using `IO.write(..., append = true)`

## Tests
Scripted test in the sbt shell: `scripted dependency-graph/append-output`

### DIFF
--- a/main/src/main/scala/sbt/plugins/DependencyTreeSettings.scala
+++ b/main/src/main/scala/sbt/plugins/DependencyTreeSettings.scala
@@ -286,7 +286,7 @@ OPTIONS
             val output = format match
               case Fmt.Json => rendering.LicenseInfo.renderJson(graph)
               case _        => rendering.LicenseInfo.render(graph)
-            handleOutput(output, outFileOpt, isQuiet, s.log)
+            handleOutput(output, outFileOpt, isQuiet, appendToFile = false, s.log)
           }
       }).evaluated,
     )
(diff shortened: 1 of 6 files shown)


eed3si9n (approved):
@oolokioo7 Thanks for the contribution! The PR looks great.

PR #8879, opened 2026-03-06
[2.x] Improve -help/--help and fix launcher test harness

Eruis2579:
- sbt script: add 'Getting started' section to usage (create project, help commands)
- sbt script: handle -h/--help without starting JVM, exit 0
- RunnerScriptTest: add test for 'sbt --help should show getting-started hints'
- ShellScriptUtil: use java.home when JAVA_HOME unset so tests run in CI

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -586,7 +586,9 @@ class NetworkClient(
   private def getExitCode(jvalue: Option[JValue]): Integer = jvalue match {
     case Some(jv) =>
       import sbt.protocol.codec.JsonProtocol.given
-      Converter.fromJson[ExecStatusEvent](jv).toOption
+      Converter
+        .fromJson[ExecStatusEvent](jv)
+        .toOption
         .flatMap(_.exitCode)
         .fold(Integer.valueOf(1))(code => Integer.valueOf(code.toInt))
     case _ => Integer.valueOf(1)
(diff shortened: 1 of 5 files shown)


Eruis2579:
Hi, @eed3si9n , Could you review my PR?

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -584,14 +584,14 @@ class NetworkClient(
...
-        .getOrElse(1)
-    case _ => 1
+    case Some(jv) =>
+      import sbt.protocol.codec.JsonProtocol.given
+      Converter
+        .fromJson[ExecStatusEvent](jv)
+        .toOption
+        .flatMap(_.exitCode)
+        .fold(Integer.valueOf(1))(code => Integer.valueOf(code.toInt))
+    case _ => Integer.valueOf(1)
```
eed3si9n: Are these changes related to the help issue?

sbt
```diff
@@ -690,6 +691,18 @@ Usage: `basename "$0"` [options]
 
 In the case of duplicated or conflicting options, the order above
 shows precedence: JAVA_OPTS lowest, command line options highest.
+
+Getting started with sbt:
+
+  - To create a new project:
+      sbt new scala/scala3.g8
+    or, for Scala 2:
+      sbt new scala/scala-seed.g8
```
eed3si9n:
```suggestion
  - To create a new project run: sbt init
```

eed3si9n (changes requested):
I think we should nudge ppl to use `sbt init`, which shows some selection of templates, including the ones that work without the internet.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
Eruis2579: Thank you for your review. 
I added this code cuz client test failed.

Eruis2579: I have reverted NetworkClient file as original.

Eruis2579: But I don't know why ServerTest failed. Could you help me for this?

eed3si9n: I still see NetworkClient diff.

Eruis2579: image
As you can see from the image, I removed NetworkClient diff.
Please check again. Thank you

eed3si9n (approved):
Thanks!

PR #8880, opened 2026-03-07
[2.x] feat: updating remote vcs projects

### LINKED ISSUE #1284: Managed dependencies from Git not updated
As covered by previous issue: https://github.com/sbt/sbt/issues/335

Using SBT 0.13+ and Git dependencies: 

``` scala
  val project = RootProject(uri("git://github.com/...."))
  val root = Project("root", file(".")) dependsOn(project)
```

The project is checked into an SBT staging area but any subsequent changes to the dependency are not picked up by an 'sbt update'. 

The ticket above suggests this was fixed in 0.12 but there are also comments that this does not work in 0.13 indicating that this is a bug.

Currently it appears the only way to pull updated code is to delete the dependency from the staging area or manually do a git pull from within the staging area but running 'sbt update' should update all dependencies.

seroperson:
Closes #1284

This PR implements updating remote VCS projects. 

- By default projects aren't updated automatically and updating is only allowed using `updateRemoteProjects` command. To change this behaviour, user must define `vcsProject / repositoryUpdateStrategy` key (available values are `Manual`, `Always` and `Every(interval)`). When set to `Always`, `git fetch` will be performed at each `update`; when set to `Every`, `git fetch` will be performed if time since last successful update greater than `interval`.
- Added several scripted tests to cover most of scenarios.
- Also refactored `Resolvers` just a little. I've changed some public methods to be private and removed mostly redundant `DistributedVCS` class. I'm unsure why these things were public but I can revert these changes if needed.


main/src/main/scala/sbt/Resolvers.scala
```diff
@@ -175,4 +180,162 @@ object Resolvers {
...
+      case RemoteVcs.Hg(uri)  => updateMercurial(localCopy, uri, log)
+      case RemoteVcs.Svn(uri) => updateSubversion(localCopy, uri, log)
+    }
+
+  private def updateGit(localCopy: File, uri: URI, log: Logger): Boolean =
+    try {
+      val headBefore = captureOutput(Some(localCopy), "git", "rev-parse", "HEAD")
+      val ref = if (fromURI(uri).hasFragment) uri.getFragment else "HEAD"
+      run(Some(localCopy), Some(log), "git", "fetch", "origin", ref)
+      run(Some(localCopy), Some(log), "git", "reset", "--hard", "FETCH_HEAD")
```
anatoliykmetyuk: Potential for loss of work during a hard reset if e.g. a user did some changes to the dependency for debugging. Consider checking whether there are any changes present and aborting the update with an error message if that is the case.

main/src/main/scala/sbt/Resolvers.scala
```diff
@@ -175,4 +180,162 @@ object Resolvers {
...
+    try {
+      if (fromURI(uri).hasFragment) {
+        val branch = uri.getFragment
+        run(Some(localCopy), Some(log), "hg", "pull")
+        run(Some(localCopy), Some(log), "hg", "update", branch)
+      } else {
+        run(Some(localCopy), Some(log), "hg", "pull", "-u")
+      }
+      markUpdated(localCopy)
+      true
```
anatoliykmetyuk: Here and in case of `updateSubversion`, would be good to stay consistent with the git behavior, which is return `true` only if some actual changes were fetched.

main/src/main/scala/sbt/Resolvers.scala
```diff
@@ -175,4 +180,162 @@ object Resolvers {
...
+              if (dep.build != startRef.build) acc + dep.build else acc
+            }
+            go(allDeps.toList ::: rest, visited + ref, newResult)
+          }
+      }
+    go(List(startRef), Set.empty, Set.empty)
+  }
+
+  private[sbt] def updateRepository(localCopy: File, uri: URI, log: Logger): Boolean =
+    RetrieveUnit(uri).exists(updateRepository(localCopy, _, log))
```
anatoliykmetyuk: `exists` is meant for pure predicates, yet, `updateRepository` is side-effecting. Consider e.g. using a `match` on `RetrieveUnit` instead to make the code more readable.

main/src/main/scala/sbt/Resolvers.scala
```diff
@@ -175,4 +180,162 @@ object Resolvers {
...
+    val repos = (for {
+      (uri, unit) <- lb.units if unit.localBase.exists()
+      vcs <- RetrieveUnit(uri) if (force || {
+        val rootProjectId = unit.rootProjects.headOption.getOrElse("root")
+        val projectRef = ProjectRef(uri, rootProjectId)
+        val strategy = extracted
+          .getOpt(projectRef / Keys.repositoryUpdateStrategy)
+          .getOrElse(RepositoryUpdateStrategy.Manual)
+        shouldUpdate(unit.localBase, strategy)
+      })
```
anatoliykmetyuk: Consider extracting the multi-line predicate into a local function for readability.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -2514,7 +2515,23 @@ object Defaults extends BuildCommon {
...
+        RetrieveUnit(uri) match {
+          case None => ()
+          case Some(vcs) =>
+            val strategy = repositoryUpdateStrategy.value
+            val lb = Project.extract(state.value).get(Keys.loadedBuild)
+            val vcsRoot = lb.units.get(uri).map(_.localBase).getOrElse(baseDirectory.value)
+            if (Resolvers.shouldUpdate(vcsRoot, strategy)) {
+              log.info(s"Updating remote project: $vcsRoot ...")
+              if (Resolvers.updateRepository(vcsRoot, vcs, log))
+                log.warn("Remote dependencies updated. Run `reload` to pick up changes.")
```
anatoliykmetyuk, line 2528: We probably should run the update automatically here.

sbt-app/src/sbt-test/project/remote-project-dedup/test
```diff
@@ -0,0 +1,14 @@
...
+$ exec bash setup.sh
+$ copy-file changes/build-with-deps.sbt build.sbt
+> reload
+> d/run
+
+$ exec bash add-newservice.sh
+$ copy-file changes/MainWithNew.scala d/src/main/scala/Main.scala
+> d/update
+> reload
+> d/run
```
anatoliykmetyuk: Can you please elaborate on what this test is testing? From the name, I take it tests that a transitive git dependency in a diamond dependency layout won't get fetched twice. But how exactly does it verify it?

sbt-app/src/sbt-test/project/remote-project-dedup/d/src/main/scala/Main.scala
```diff
@@ -0,0 +1,8 @@
+object Main {
+  def main(args: Array[String]): Unit = {
+    val cls = Class.forName("upstream.Service")
```
anatoliykmetyuk: Why runtime reflection and not a compile-time import?

sbt-app/src/sbt-test/project/remote-project-update/changes/MainCheckManual.scala
```diff
@@ -0,0 +1,28 @@
...
+    assert(util == "utility", s"Expected 'utility' but got '$util'")
+    println(s"Utils loaded: $util")
+
+    try {
+      Class.forName("upstream.ManualClass")
+      sys.error("upstream.ManualClass found on classpath")
+    } catch {
+      case _: ClassNotFoundException =>
+        ()
+      case e =>
```
anatoliykmetyuk: Why a catch-all clause and not just let the unwrapped exception propagate to the test runner?

anatoliykmetyuk (changes requested):
Thank you @seroperson ! Left some comments. Also:

- It would be good to document the new feature somewhere
- It would also be good to cover the `Every` update strategy - maybe have a very small interval and a `sleep`?

sbt-app/src/sbt-test/project/remote-project-dedup/test
seroperson: This was supposed to check that such a layout does work at all 😃 I can try to add check that it downloads just once.

main/src/main/scala/sbt/Defaults.scala
seroperson: From what I see it looks not trivial to do so. I do that in `updateRemoteProjects`, but `updateRemoteProjects` is a command which can easily modify state, while `fetchSources` is a task. Are there any tips on how to implement it (if it's still necessary)?

seroperson:
@anatoliykmetyuk thank you for such a detailed review! I've addressed most of comments and pushed new changes.

Regards to documentation, I've added a small `.md` file in `notes/2.0.0/`, but this directory seems so empty that I'm unsure if it's used somewhere or not. I can push this overview somewhere else if needed or you can just re-use it later.

main/src/main/scala/sbt/Defaults.scala
anatoliykmetyuk: If there's no trivial way to do it, a warning is fine I think, it communicates enough information to the user. If that becomes a problem in the future, we can make a separate issue / PR to address it.

sbt-app/src/sbt-test/project/remote-project-dedup/test
```diff
@@ -1,14 +1,15 @@
-# Checks that `d` project:
-# - Receives changes from a transitive dependency
-# - Calls `git pull` on a indirectly depended VCS project
-
 $ exec bash setup.sh
 $ copy-file changes/build-with-deps.sbt build.sbt
 > reload
-> d/run
+> compile
```
anatoliykmetyuk, line 4: Is there a reason for not running the test any more? When I was commenting on runtime reflection vs compile-time import, I meant the dependency mechanism, not the test flow mechanism :)

anatoliykmetyuk (approved):
LGTM, thanks!

anatoliykmetyuk:
> Regards to documentation, I've added a small .md file in notes/2.0.0/, but this directory seems so empty that I'm unsure if it's used somewhere or not. I can push this overview somewhere else if needed or you can just re-use it later.

Regarding where to document this new feature, @eed3si9n could you please point in the right direction? I guess the place of documentation would depend on the sbt version this feature will appear in. I think we can't just write the doc at sbt/website until the feature is actually shipped and available to users, can we?..

eed3si9n:
Current `develop` branch is slated for sbt 2.1.x series, since we're trying to stabilize for 2.0.x. I'm happy with some markdown files in `notes/2.1.0` or `notes/2.0.0`.

main/src/main/scala/sbt/RepositoryUpdateStrategy.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import scala.concurrent.duration.FiniteDuration
+
+/**
+ * Strategy for updating remote VCS project dependencies.
+ *
+ * Used with the `repositoryUpdateStrategy` setting key to control when
+ * remote project dependencies are updated from their upstream repositories.
+ */
+sealed trait RepositoryUpdateStrategy
```
eed3si9n: Would Scala 3 enum work for this?

eed3si9n (approved):
A minor comment on enum.

seroperson:
@eed3si9n done 🤝

anatoliykmetyuk:
Thanks @seroperson !

PR #8884, opened 2026-03-08
[2.x] Update TemplateCommandUtil.scala

xuwei-k:


main/src/main/scala/sbt/TemplateCommandUtil.scala
```diff
@@ -304,7 +304,7 @@ libraryDependencies += (toolkitTest % Test)
   }
 
   private def typelevelToolkitTemplate(): Unit = {
-    val defaultTypelevelToolkitV = "0.1.28"
+    val defaultTypelevelToolkitV = "0.1.29"
```
xuwei-k, line 307: https://repo1.maven.org/maven2/org/typelevel/toolkit_3/

main/src/main/scala/sbt/TemplateCommandUtil.scala
```diff
@@ -284,9 +284,9 @@ private[sbt] object TemplateCommandUtil {
         throw new IllegalArgumentException(error)
     }
 
-  private final val defaultScalaV = "3.3.4"
+  private final val defaultScalaV = "3.3.7"
   private def scalaToolkitTemplate(): Unit = {
-    val defaultScalaToolkitV = "0.5.0"
+    val defaultScalaToolkitV = "0.9.1"
```
xuwei-k, line 289: https://repo1.maven.org/maven2/org/scala-lang/toolkit_3/

eed3si9n (approved):
Thanks!
I wonder if we should generate a case class that tracks just the version numbers so all version numbers are tracked in `project/Dependencies.scala`.

PR #8888, opened 2026-03-09
[2.x] Fix: sbt.bat build.properties lookup on Windows

### LINKED ISSUE #6959: Message "The system cannot find the file project\build.properties." reported on windows
## steps

1 setup sb build 1.6.2 using binary distribution ( https://github.com/sbt/sbt/releases/download/v1.6.2/sbt-1.6.2.zip )
2 clone any sbt built project sporting project\build.properties file
3 using windows command prompt cd under project directory
4 make sure sbt from step 1 is in the path
5 invoke sbt as sbt --mem 4096

## problem
observe message "The system cannot find the file project\build.properties"

## expectation

problem message should not be printed

## notes

https://github.com/sbt/sbt/blob/8aae00fb42d8d354cba226d1244a197df671a35c/launcher-package/src/universal/bin/sbt.bat#L91

to

```bash
for /F "eol=# delims== tokens=1*" %%a in (%WORK_DIR%\project\build.properties) do (
```

and set WORK_DIR to %CD% somewhere at the top of the script

## please note

Windows command prompt has tendency to set working directory when expanding parenthesis for the "for /F" command to windows system directory, which might have unintended consequence in the absence of target directory as part of the script in parenthesis.

oolokioo7:
Closes #6959

## Problem
On Windows, the original batch logic could fail while trying to read `project\build.properties`, which led to errors

> The system cannot find the file project\build.properties.

especially with options like `--mem 4096`

## Fix
Changed a line in the /src/universal/bin/sbt.bat from
`for /F "eol=# delims== tokens=1*" %%a in (project\build.properties) do (`
to
`for /F "usebackq eol=# delims== tokens=1*" %%a in ("project\build.properties") do (`

# How does it fix the issue?

- Quotes the filename `project\build.properties`
  - so the file operand is passed as a single literal path
- Adds `usebackq`
   - which tells for `for /F` that a double-quoted item is a filename
   
# Test on windows
`--mem 4096 --script-version`
- ran success as expected and didn't emit : `The system cannot find the file project\build.properties`.

`--version`
- correctly print sbt version

### DIFF
--- a/launcher-package/src/universal/bin/sbt.bat
+++ b/launcher-package/src/universal/bin/sbt.bat
@@ -94,7 +94,7 @@ if defined JAVA_HOMES (
 )
 
 if exist "project\build.properties" (
-  for /F "eol=# delims== tokens=1*" %%a in (project\build.properties) do (
+  for /F "usebackq eol=# delims== tokens=1*" %%a in ("project\build.properties") do (
     if "%%a" == "sbt.version" if not "%%b" == "" (
       set build_props_sbt_version=%%b
     )


oolokioo7:
@eed3si9n . It's ready for your review. thanks

launcher-package/src/universal/bin/sbt.bat
```diff
@@ -94,7 +94,7 @@ if defined JAVA_HOMES (
 )
 
 if exist "project\build.properties" (
-  for /F "eol=# delims== tokens=1*" %%a in (project\build.properties) do (
+  for /F "usebackq eol=# delims== tokens=1*" %%a in ("project\build.properties") do (
```
eed3si9n, line 97: Thanks for working on this. Were you able to reproduce this problem either locally or using https://github.com/sbt/sbt/blob/develop/launcher-package/integration-test/src/test/scala/RunnerMemoryScriptTest.scala? As far as I know, we have a test for `sbt -mem 503`. The original issue includes the note

> 
Windows command prompt has tendency to set working directory when expanding parenthesis for the "for /F" command to windows system directory, which might have unintended consequence in the absence of target directory as part of the script in parenthesis.

oolokioo7: I tried many cases locally. This issue is not related to the '-mem' flag.

It works well in a normal Windows environment, but it triggers a "not found" error when the project is located on a UNC path. The .bat file automatically launches CMD.exe, and it doesn't support UNC paths.

Using a UNC path on Windows is an unusual case, and there may be no solution to fix it.

My change is not necessary for this issue. Actually, it doesn't change anything, completely same with the original one.

oolokioo7: then if you have time, plz review https://github.com/sbt/sbt/pull/8867

eed3si9n:
> My change is not necessary for this issue. Actually, it doesn't change anything, completely same with the original one.

I'm closing this issue, since as the OP says this doesn't change anything while claiming to close an existing issue.

Issue #8889, opened 2026-03-10
[2.x] FileNotFoundException from corrupted/evicted remote CAS entry fails build instead of falling back to recompute


idanbenzvi:
**sbt version**: 2.0.0-RC9
**Scala version**: 3.x

## Problem

When using the built-in gRPC remote cache (`Global / remoteCache := Some(uri("grpc://..."))`)
with bazel-remote as the CAS server, builds intermittently fail with:

```
[error] java.io.FileNotFoundException: .../target/out/value/sha256-/.json (No such file or directory)
22:58:06  [0m[[0m[31merror[0m] [0m[0mjava.io.FileNotFoundException: /home/jenkins/agent/workspace/I_cap-commons_feature_sbt2_final/target/out/value/sha256-16b04cae19175fe44daff3c3be06dca3fa7af838a6f85add49fce6ae2afcb80d/48.json (No such file or directory)[0m
```

[failure_log_during_publish.txt](https://github.com/user-attachments/files/25864126/failure_log_during_publish.txt)
---

This happens when the remote cache server reports an **AC (Action Cache) hit**
but the corresponding **CAS blob is missing or was evicted (possibly corrupted?)**. 
Rather than treating this as a cache miss and recomputing the task, SBT propagates the
exception and fails the build.

The failure is intermittent — it disappears on retry without any code changes,
confirming it is a transient infrastructure issue rather than a build logic error.

## Steps to reproduce
Background:
Java Azul JVM 17.0.18
SBT version 2.13.16 
Running on Jenkins CI environment 

The project is a large monorepo consisting of several projects some of which are libraries while other are microservices with deep interdependency between each other.

1. Configure a remote cache backed by bazel-remote (running on a pod - the specific settings can be provided if needed)
2. Evict or corrupt a CAS entry while leaving the AC entry intact
3. Run a build — `java.io.FileNotFoundException` is thrown from `ActionCache`

## Expected behavior

SBT should treat a missing/unreadable CAS blob as a cache miss and recompute
the task locally, just as it does for other cache miss scenarios.

## Notes

I've identified three unguarded `syncBlobs` call sites in [ActionCache.scala](cci:7://file:///Users/ibenzvi/sourcecode/sbt/util-cache/src/main/scala/sbt/util/ActionCache.scala:0:0-0:0)
and prepared a fix with regression tests. Will open a PR alongside this issue.

## Issue flow
```mermaid
flowchart TD
[...]

PR #8890, opened 2026-03-10
[2.x] fix: fall back to the onsite task when cache fails

idanbenzvi:
A fix to the reported issue #8889 

**Problem Symptom**
When using SBT 2's remote cache (backed by bazel-remote over gRPC), a build occasionally fails with:
[error] java.io.FileNotFoundException: .../target/out/value/sha256-<****>/48.json (No such file or directory)
This error occurs randomly and in stages of the sbt build process - compilation, testing or publishing. 

**Problem Analysis and root cause**
When the remote cache server (e.g. bazel-remote using S3 for storage) reports an AC (Action Cache) hit but the underlying CAS (Content Addressable Storage) blob is missing or corrupt, ActionCache.cache propagates the resulting exception (typically java.io.FileNotFoundException) directly to the SBT task engine process with no interception of the propogated error. This causes a build failure instead of a graceful cache miss.

The three unguarded call sites are:
| Location | Risk |
|---|---|
| `organicTask` — `syncBlobs` after `put` | `NoSuchFileException` was caught; `FileNotFoundException` was not |
| `getWithFailure` / `readFromSymlink` fast-path | `syncBlobs` was inside a `flatMap` with no exception handling |
| `getWithFailure` main branch | Both `syncBlobs` calls and `IO.read(paths.head.toFile())` were fully unguarded |

**Solution Premise**
By definition, a viable response to a broken cache entry is "recompute" / "rebuild" instead of failing the whole build process.
This of course does not explain the corrupt cache entry or missing "cache hit", but it does allow recovery instead of total failure.

**Solution**
Guard all three call sites with NonFatal catches:
- Cache read failures (getWithFailure) return Left(None) which the caller interprets as a cache miss, triggering organic recomputation.
- Cache write failures (organicTask) are demoted to a debug-level log; the task result that was already computed is returned successfully.

**Testing and Validation**
Two regression tests are added to ActionCacheTest:

1. **`Cache falls back to recompute when syncBlobs throws FileNotFoundException`**
   — uses a `ThrowingSyncStore` wrapper that always throws `FileNotFoundException`
   from `syncBlobs`. Verifies that both the first call (organic task, then
   `syncBlobs` throws on cache write → result still returned) and the second
   call (AC hit, `syncBlobs` throws on cache read → recomputes) succeed.

2. **`readFromSymlink fast path falls back to recompute when syncBlobs throws FileNotFoundException`**
   — same scenario but uses an absolute-path `FileConverter` so the symlink
   created during the first run is found by `Files.isSymbolicLink` on the
   second, specifically exercising the `readFromSymlink` fast-path fix.

## Reproduction
Observed on SBT `2.0.0-RC9` with bazel-remote as the CAS server running as a
Kubernetes sidecar. The failure occurred intermittently mid-build when the
remote cache had a valid AC entry but the CAS blob had been evicted. The exact
stack trace:
[error] java.io.FileNotFoundException: /home/jenkins/agent/workspace/.../ target/out/value/sha256-16b04cae19175fe44daff3c3be06dca3fa7af838a6f85add49fce6ae2afcb80d/48.json (No such file or directory)

The build succeeded on retry without any code changes, confirming the issue was
a transient missing CAS entry rather than a build logic error.

## Comments / Suggestions
I would love to hear your feedback(!), this is the first time i've contributed to the SBT project and the syntax is very novel to me - and actually put the AI to use to facilitate the process.

## AI Disclosure 
This PR was developed with AI assistance (Cascade / Claude Sonnet 4.5 / Gemini Pro 3 Thinking for counter-peer-reviewing) as per
the [AI assisted contributions](CONTRIBUTING.md#genai) guidelines. All code was
reviewed, understood, and verified by a human with the tests run locally. The
root cause analysis, fix location identification, and test design were validated
manually against the SBT source.

### DIFF
--- a/util-cache/src/main/scala/sbt/util/ActionCache.scala
+++ b/util-cache/src/main/scala/sbt/util/ActionCache.scala
@@ -10,7 +10,7 @@ package sbt.util
 
 import java.io.File
 import java.nio.charset.StandardCharsets
-import java.nio.file.{ Files, NoSuchFileException, Path, Paths, StandardCopyOption }
+import java.nio.file.{ Files, Path, Paths, StandardCopyOption }
 import sbt.internal.util.{ ActionCacheEvent, CacheEventLog, StringVirtualFile1 }
 import sbt.io.syntax.*
 import sbt.io.IO
@@ -19,7 +19,7 @@ import sbt.nio.file.syntax.*
 import sbt.util.CacheImplicits
 import scala.reflect.ClassTag
 import scala.annotation.{ meta, StaticAnnotation }
-import scala.util.control.Exception
+import scala.util.control.{ Exception, NonFatal }
 import sjsonnew.{ HashWriter, JsonFormat }
 import sjsonnew.support.murmurhash.Hasher
 import sjsonnew.support.scalajson.unsafe.{ CompactPrinter, Converter, Parser }
@@ -111,8 +111,8 @@ object ActionCache:
               result
             case Left(e) => throw e
       catch
-        case e: NoSuchFileException =>
-          logger.debug(s"Skipping cache storage due to missing file: ${e.getMessage}")
+        case NonFatal(e) =>
+          logger.debug(s"Skipping cache storage due to error: ${e.getMessage}")
           cacheEventLog.append(ActionCacheEvent.Error)
           result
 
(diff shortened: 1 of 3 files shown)


util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -111,8 +111,8 @@ object ActionCache:
               result
             case Left(e) => throw e
       catch
-        case e: NoSuchFileException =>
-          logger.debug(s"Skipping cache storage due to missing file: ${e.getMessage}")
+        case NonFatal(e) =>
```
eed3si9n: If we know that the error manifests itself as NoSuchFileException, I feel like it's better to scope this to NoSuchFileException or IOException, rather than a wholesale NonFatal? Or are there different patterns of failure mode like network error that can throw different exceptions?

idanbenzvi: I thought that since this issue might also occur in cases where the network communication with the bazel-remote-cache pod is broken (manifesting itself as either RuntimeExceptions or TimeoutExceptions) however, i do feel you're right and that generalizing would be the incorrect approach. I suggest electing the IOException as it is the right "breadth" to encompass all possible exceptions that are possible in that context (filenotfound, access denied, etc.).
I'll push the change. Thanks for the quick review(!)

eed3si9n (approved):
Thanks!

PR #8894, opened 2026-03-11
[2.x] fix: Remove projectDescriptors key that leaks Ivy types

### LINKED ISSUE #8865: [2.x] Deprecate, hide, or remove projectDescriptors
Apparently projectDescriptors currently leaks Ivy classes. We should probably remove, hide, or remove it before sbt 2.0.0 final version comes out.

Looking at https://github.com/search?q=projectDescriptors+language%3AScala&type=code&l=Scala&p=1 I don't think plugins are using it.

eureka0928:
## Problem

The `projectDescriptors` key in `sbt.Keys` leaks Ivy classes (`ModuleRevisionId`, `ModuleDescriptor`) through its API surface. A [GitHub code search](https://github.com/search?q=projectDescriptors+language%3AScala&type=code&l=Scala&p=1) confirms no external plugins use this key.

## Solution

Remove `projectDescriptors` entirely:
- **`Keys.scala`** / **`Defaults.scala`**: Delete the key definition and its empty default
- **`IvyDependencyPlugin.scala`**: `projectResolverTask` and `coursierExtraProjectsTask` now use the internal `depMap` method directly, eliminating unnecessary `Any` casts
- **`GlobalPlugin.scala`**: Remove `descriptors` field from `GlobalPluginData` and its inject/extract plumbing
- **`CoursierInputsTasks.scala`**: Update stale comment

Fixes #8865

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -740,6 +740,19 @@ lazy val mainProj = (project in file("main"))
       exclude[DirectMissingMethodProblem]("sbt.Classpaths.mkIvyConfiguration"),
       exclude[MissingClassProblem]("sbt.internal.librarymanagement.IvyXml"),
       exclude[MissingClassProblem]("sbt.internal.librarymanagement.IvyXml$"),
+      // Removed projectDescriptors key (sbt#8865)
+      exclude[DirectMissingMethodProblem]("sbt.Keys.projectDescriptors"),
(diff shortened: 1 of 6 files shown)


eed3si9n (approved):
Thanks!

PR #8895, opened 2026-03-11
[2.x] feat: Support --autostart=false and --no-server in sbtn client

### LINKED ISSUE #7079: start sbt client in a way that it fails when it can't connect
Currently we have a setup in our company where a build script starts sbt, and sbt really must be started from this location and not somewhere else. My editor will thankfully connect to that instance of sbt thanks to the `SBT_NATIVE_CLIENT=true` environment variable. But it will fall back to start its own instance of sbt and eventually cause two simultaneously running instances of sbt. I would really prefer if my editor would tell me that sbt wasn't started yet instead of starting its own instance.

eureka0928:
Fixes #7079

### Problem

When `SBT_NATIVE_CLIENT=true`, the sbt thin client (`sbtn`) always starts a new server if one isn't running. In workflows where sbt must be started from a specific build script, editors like Metals silently start their own instance instead of failing.

### Solution

Reuse the existing `sbt.server.autostart` system property in NetworkClient to control whether sbtn should attempt to start a server. When no server is running and `sbt.server.autostart=false`, sbtn exits with an error instead of prompting.

Support setting this property via:
- `sbtn --no-server compile` (sets `sbt.server.autostart=false`)
- `sbtn --autostart=false compile` (new flag following sbt naming conventions)
- `sbtn -Dsbt.server.autostart=false compile` (direct system property)
- `sbt --autostart=false compile` (bash runner and sbtw)

This follows sbt's naming conventions for properties/options: positive property names with `=false` opt-out (like `--color=false`, `--supershell=false`).

### Changes

- `NetworkClient.scala`: Check `sbt.server.autostart` property in `connectOrStartServerAndConnect`; handle `--no-server` and `--autostart=` in `parseArgs` by setting the system property
- `NetworkClientParseArgsTest.scala`: Tests for `--no-server` and `--autostart=` flag parsing
- `sbt` (bash runner): Add `--autostart=` flag support
- `sbtw` (`ArgParser`, `LauncherOptions`, `Runner`): Add `--autostart=` flag support

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -151,6 +151,8 @@ class NetworkClient(
   private lazy val noStdErr = arguments.completionArguments.contains("--no-stderr") &&
     !sys.env.contains("SBTN_AUTO_COMPLETE") && !sys.env.contains("SBTC_AUTO_COMPLETE")
   private def shutdownOnly = arguments.commandArguments == Seq(Shutdown)
+  private lazy val serverAutoStart: Boolean =
+    sys.props.get("sbt.server.autostart").forall(_.toLowerCase == "true")
 
   private def mkSocket(file: File): (Socket, Option[String]) = ClientSocket.socket(file, useJNI)
 
@@ -189,6 +191,9 @@ class NetworkClient(
         if (shutdownOnly) {
           console.appendLog(Level.Info, "no sbt server is running. ciao")
           System.exit(0)
+        } else if (!serverAutoStart) {
+          console.appendLog(Level.Error, "no sbt server is running (sbt.server.autostart=false)")
+          System.exit(1)
         } else if (promptCompleteUsers) {
           val msg = if (noTab) "" else "No sbt server is running. Press <tab> to start one..."
           errorStream.print(s"\n$msg")
@@ -1187,7 +1192,7 @@ object NetworkClient {
         completionArguments,
         sbtScript,
         bsp,
-        sbtLaunchJar
+        sbtLaunchJar,
       )
   }
   private[client] val completions = "--completions"
@@ -1246,8 +1251,6 @@ object NetworkClient {
     "--timings",
     "-traces",
     "--traces",
-    "-no-server",
-    "--no-server",
     "-no-share",
     "--no-share",
     "-no-global",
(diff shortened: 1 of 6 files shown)


fury0928:
Hi @eed3si9n would you review and merge this PR?
Thank you

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -189,6 +189,9 @@ class NetworkClient(
         if (shutdownOnly) {
           console.appendLog(Level.Info, "no sbt server is running. ciao")
           System.exit(0)
+        } else if (arguments.noStart || sys.env.get("SBT_NO_START").contains("true")) {
```
eed3si9n: Thanks for working on this. For new properties/options, I suggest we use the positive thing the noun even if the purpose of the option is to opt out of it.   - https://github.com/sbt/sbt/blob/ab503ca51e3b3fdc2a8c0d4ce74eb828c4103729/main/src/main/scala/sbt/internal/SysProp.scala#L82-L86

For example, `sbt --color=false --supershell=false`. Historically, we've also had `sbt --no-color` as well. So following this pattern, I suggest adding something in `sbt` runners (Bash, Batch, and sbtw), and pass that information as always/true/false/never type parameter. Interestingly, there is actually `--no-server` in the sbt runner:

```bash
     -no-server|--no-server) options=( "${options[@]}" "-Dsbt.io.virtual=false" "-Dsbt.server.autostart=false" ) && shift ;;
```

(Side note we should support `--autostart=false`)

When someone runs from the sbt runner with `sbt --no-server`, does NetworkClient see that? If so we can just reuse that to mean we don't want to start the server.

eed3si9n (changes requested):

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
fury0928: Thanks for the feedback! Reworked to follow sbt's naming conventions:

- Removed `--no-start` flag entirely
- NetworkClient now checks the existing `sbt.server.autostart` system property (consistent with `SysProp.serverAutoStart`, `Main.scala`, and `CommandExchange`)
- `--no-server` in sbtn's `parseArgs` now sets `sbt.server.autostart=false` instead of being silently dropped
- Added `--autostart=false` / `--autostart=true` support in NetworkClient, the bash runner, and sbtw

Supported invocations:
```
sbtn --no-server compile
sbtn --autostart=false compile
sbtn -Dsbt.server.autostart=false compile
sbt --autostart=false compile
```

eed3si9n (approved):
Thanks!
Could you update the PR title / description?

fury0928:
@eed3si9n PR is updated, good to merge!

PR #8896, opened 2026-03-11
[2.x] refactor: Use `InputStream.transferTo`

xuwei-k:
https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/io/InputStream.html#transferTo(java.io.OutputStream)


eed3si9n (approved):
Thanks!

PR #8903, opened 2026-03-13
[2.x] fix: Retry on HTTP 5xx during dependency resolution

majk-p:
This PR ports the change from sbt-coursier https://github.com/coursier/sbt-coursier/pull/601 to sbt 2.x lineup


eed3si9n (approved):
Thanks!

PR #8904, opened 2026-03-13
[2.x] fix: server-test and related refactoring

### LINKED ISSUE #8958: [2.x] Refactor server tests
Refactor server tests to ServerSession.

seroperson:
Fixes https://github.com/sbt/sbt/issues/8958

I've updated the PR, now it consists mostly of `server-test` refactoring.

- `TestServer` was replaced with `ServerSession` class, which is located in `protocol` module and allows you to interact with JSON-RPC sbt server.
- `ServerConnection` from `main-command` now extends `ServerSession`, so both tests and actual code use the same class for low-level protocol interaction.
- Fixed few flaky tests in `BuildServerTest`, `ChannelCursorTest`, probably something else. Mostly task-key caching issues.
- Getting rid of hardcoded ids in tests.
- Getting rid of hardcoded jsons and raw string comparing in tests.

I will also re-use `ServerSession` in `sbt-testkit` a little bit later.

<details>
<summary>Initial PR description</summary>

This PR implements an experimental module `sbt-testing-framework`. It allows you to code scripted-like tests as common JUnit specs, as some time ago it was discussed in https://github.com/sbt/sbt/discussions/8338. Now there is such ability using `SbtRunner` with tons of sweet APIs to implement your tests. Now sbt is in line with gradle's `GradleRunner` and mill's `IntegrationTester`.

I've added a lot of examples, implemented some tests in project, made it possible to code such tests right in sbt itself, and added some kind of documentation in `notes/2.0.0/sbt-testing-framework.md`. 

You can test it using `sbt sbtTestingFramework/test` or `sbt sbtProj/integration:test`. Also I've tested it in my plugin, `jvm-live-reload`: https://github.com/seroperson/jvm-live-reload/tree/sbt-testing-framework. So, if you want a real-world scenario, you can clone this repo and branch and do `sbt quickLocalPublish && sbt sbtLiveReload/testOnly` (preliminarily you must locally publish sbt things from this branch). 

co-authored with claude 4.6 opus, mostly for some routine tasks.
</details>


eed3si9n (dismissed):
As commented in https://github.com/sbt/sbt/discussions/8338#discussioncomment-14718878, we already have server-test (https://github.com/sbt/sbt/blob/develop/server-test/src/test/scala/testpkg/TestServer.scala), which can write ScalaTest against a running instance of an sbt session. We can extend that functionality maybe, but I don't think we need duplicate mechanism unless there's a need for it.

seroperson:
> As commented in [#8338 (comment)](https://github.com/sbt/sbt/discussions/8338#discussioncomment-14718878), we already have server-test ([`develop`/server-test/src/test/scala/testpkg/TestServer.scala](https://github.com/sbt/sbt/blob/develop/server-test/src/test/scala/testpkg/TestServer.scala)), which can write ScalaTest against a running instance of an sbt session. We can extend that functionality maybe, but I don't think we need duplicate mechanism unless there's a need for it.

Hm, well, yes, but it's hard to say that it's usable by an end user, like a plugin developer. From my perspective it more looked like internal utility for testing the protocol and I thought you referred to it just like to an example of how it can be implemented. The only common thing left between `TestServer` and `SbtRunner` is a connecting and interacting with underlying sbt instance, I can try to merge this logic into some place, so both reuse it. `sbt-testing-framework` in such scenario would be just a neat facade for an end user and wont hold any protocol-related code.

anatoliykmetyuk:
Thank you for working on this!

Can this be a separate sbt plugin instead of being included into the core right away? Seems like the discussion https://github.com/sbt/sbt/discussions/8338 is still in exploration mode, without a definitive conclusion on the design. This feature introduces a large surface area for potential issues as well as decision and maintenance overhead going forward - e.g., what guarantees do we want to make to the plugin authors about this feature?

Can we use a flow similar to the [project matrix](https://github.com/sbt/sbt/issues/7641) plugin that first got tested in the wild as a standalone plugin, then merged in the core after sufficient adoption and testing? This keeps the issue, maintenance, decision overhead contained while the experiment is ongoing.

seroperson:
@eed3si9n I've pushed the changes. So, now:
- The whole protocol-interaction logic is hold in a new class: `protocol/src/main/scala/sbt/protocol/ServerSession.scala`.
- `TestServer.scala` got deleted, the only thing left from it is the `AbstractServerTest`, which defines a single `SbtRunner` per spec and its' cleanup strategy.
- `SbtRunner` (and `ForkedRunner`) now uses `ServerSession` internally.

`ServerSession` holds protocol-interaction logic and neither `SbtRunner` or `TestServer` (which is gone actually) interact with protocol directly. Tests in `server-test` have a low-level access to the session using `SbtRunner.session` and they can send either raw string or a regular json objects to the RPC. I think about making `SbtRunner.session` private for `sbt` to be able to change its' API whenever we want, but currently tests in `server-test` are in `testpkg` package for some reason, so `SbtRunner.session` is public as for now.

Also as well I've refactored `server-test` very much:
- Fixed few flaky tests in `BuildServerTest`, `ChannelCursorTest`, probably something else. Mostly task-key caching issues.
- Getting rid of hardcoded ids.
- Getting rid of hardcoded jsons and raw string comparing. There are still some places left, but mostly they're gone.

Still there is some protocol-related code in `main-command` module, which intersects with `ServerSession`, but I decided to not touch it to not fall into an infinite refactoring. Though I can dig further if necessary.

@anatoliykmetyuk thank you for your response!

> Can this be a separate sbt plugin instead of being included into the core right away? 
> This feature introduces a large surface area for potential issues as well as decision and maintenance overhead going forward - e.g., what guarantees do we want to make to the plugin authors about this feature?

After the very recent refactoring the only new code is an API definition and some file read/write operations in `SbtRunner`. The rest was unified with everything previously known as `TestServer`. So, that's a very small amount of code, which is also fully covered with tests. I feel no reason to incubate such little changes separately. As a tradeoff we can get use of `@experimental` annotations to mark unstable APIs, but actually I think we can just finalize the API right in this PR. 

> Seems like the discussion https://github.com/sbt/sbt/discussions/8338 is still in exploration mode, without a definitive conclusion on the design.

Actually I thought that it's open for PRs. I can also mark it as draft if it suits better. Also I'm ready to do changes and to discuss as much as needed to finish this.

One of the biggest frustration of mine with sbt was the scripted testing, and I bet it's not only me who thinks this. While with other build systems integration testing isn't differ much from regular tests, `sbt` provides the whole new approach, which makes things much and much harder.

Starting from the fact that it's completely new and that it's handled in its' own way (`scriptedLaunchOpts`, `scriptedBufferLog`, `scriptedKeepTempDirectory`, `scripted` itself, etc etc etc), ending with the fact that you have to learn the new way of writing tests: using the combination of `test` file and the actual test code in `build.sbt`. I don't remember exactly how many hours I've spent trying to de-duplicate code across scripted tests until realized there is no easy way to do it at all.

It's just getting in line with existing solutions.

eed3si9n:
Toli's right on both points.

> Seems like the discussion https://github.com/sbt/sbt/discussions/8338 is still in exploration mode, without a definitive conclusion on the design.

Thus far the only comment I've made on it would can be summarized as "scripted works." I hadn't expected it to materialize into a PR.

> Can this be a separate sbt plugin instead of being included into the core right away?

Yea, I was going to suggest this as well. Looking at this PR, I think `SbtRunner` might be an interesting idea. I also thought `inspect tree`, slash syntax, and `projectMatrix` were interesting ideas, and I've experimented with them first as [sbt-inspectr](https://github.com/eed3si9n/sbt-inspectr), [sbt-slash](https://github.com/sbt/sbt-slash) and [sbt-projectmatrix](https://github.com/sbt/sbt-projectmatrix). I had other ideas, which have not caught on like [sbt-sriracha](https://github.com/sbt/sbt-sriracha) and [sbt-nocomma](https://github.com/sbt/sbt-nocomma). Especially if the target audience is other plugin authors, I feel like iterating first as plugin would give this idea a chance to iterate, without being tied to the sbt release cycle. `develop` branch targets sbt 2.1. With plugin, you can publish today.

Re: naming, I think sbt-testkit would be a better name than sbt-testing-framework, since:
1. `sbt.testing` is already taken for test-interface
2. a framework in test-interface terminology would mean ScalaTest, JUnit etc, but in this case `SbtRunner` is more like a helper library that can be combined with an existing test frameworks.

seroperson:
> Yea, I was going to suggest this as well. Looking at this PR, I think `SbtRunner` might be an interesting idea. I also thought `inspect tree`, slash syntax, and `projectMatrix` were interesting ideas, and I've experimented with them first as [sbt-inspectr](https://github.com/eed3si9n/sbt-inspectr?rgh-link-date=2026-03-17T08%3A56%3A17Z), [sbt-slash](https://github.com/sbt/sbt-slash?rgh-link-date=2026-03-17T08%3A56%3A17Z) and [sbt-projectmatrix](https://github.com/sbt/sbt-projectmatrix?rgh-link-date=2026-03-17T08%3A56%3A17Z). I had other ideas, which have not caught on like [sbt-sriracha](https://github.com/sbt/sbt-sriracha?rgh-link-date=2026-03-17T08%3A56%3A17Z) and [sbt-nocomma](https://github.com/sbt/sbt-nocomma?rgh-link-date=2026-03-17T08%3A56%3A17Z). Especially if the target audience is other plugin authors, I feel like iterating first as plugin would give this idea a chance to iterate, without being tied to the sbt release cycle. `develop` branch targets sbt 2.1. With plugin, you can publish today.

Yea, okay, I'm good with plugin, just thought that it's not that big to separate this. So, then I just do everything in my own separate repo? Or there is some specific flow?

What about this PR, can I then leave here this `server-test` refactoring and fixes? It eliminates some flakiness, removes weird json string comparing everywhere and removes hardcoded `id`s. It would be also good to leave these `ServerSession`-related changes, so I can later depend on this artifact and communicate with sbt easily.

seroperson:
I've updated the PR, now it consists mostly of `server-test` refactoring.

- `TestServer` was replaced with `ServerSession` class, which is located in `protocol` module and allows you to interact with JSON-RPC sbt server.
- `ServerConnection` from `main-command` now extends `ServerSession`, so both tests and actual code use the same class for low-level protocol interaction.
- Fixed few flaky tests in `BuildServerTest`, `ChannelCursorTest`, probably something else. Mostly task-key caching issues.
- Getting rid of hardcoded ids in tests.
- Getting rid of hardcoded jsons and raw string comparing in tests.

I will also re-use `ServerSession` in `sbt-testkit` a little bit later.

protocol/src/main/scala/sbt/protocol/Defaults.scala
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package sbt.protocol
+
+import scala.concurrent.duration.*
+
+/**
+ * Default duration constants for sbt server protocol operations.
+ */
+private[protocol] object Defaults {
```
eed3si9n: The name `Defaults` is taken already by (https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/Defaults.scala), so we should avoid it. If this is mostly internal impl details, could we put it into one of companion objects?

protocol/src/main/scala/sbt/protocol/JsonRpcReader.scala
```diff
@@ -6,17 +6,18 @@
...
 
-private[sbt] object ReadJsonFromInputStream {
-  def apply(
+object JsonRpcReader {
+
+  def read(
       inputStream: InputStream,
       running: AtomicBoolean,
-      onHeader: Option[String => Unit]
+      onHeader: Option[String => Unit] = None
```
eed3si9n: For binary compatibility, we should avoid default parameters on public functions, and add an overload instead with fewer parameters.

protocol/src/main/scala/sbt/protocol/JoinThread.scala
```diff
@@ -6,7 +6,7 @@
  * Licensed under Apache License 2.0 (see LICENSE)
  */
 
-package sbt.internal.util
+package sbt.protocol
```
eed3si9n: It is necessary to expose this to public?

seroperson: I'll move it to `sbt.internal.util` in `protocol` and make `private[sbt]` then

protocol/src/main/scala/sbt/protocol/ServerSession.scala
```diff
@@ -0,0 +1,449 @@
...
+ * @param threadName name for the background read thread
+ *
+ * @note This class is thread-safe for sending (multiple threads may call `sendJsonRpc`),
+ *       but only one thread should consume from the `waitFor*` methods at a time, since
+ *       non-matching messages are discarded during polling.
+ *
+ * @see [[ServerSession.connect]] to create an instance from a portfile
+ * @see [[ServerSession.waitForPortfile]] to block until the server is ready
+ */
+class ServerSession private[sbt] (
```
eed3si9n: For future-proofing, could we split this into:
1. trait with a minimal set of method signatures like `initialize`, `sendJsonRpc[A: JsonWriter](id: Int, method: String, params: A): Try[Unit]` and `waitForResultInResponseMsg`
2. hidden `private[sbt]` implementation called `SocketServerSession` accessible via `ServerSession.connect`?

protocol/src/main/scala/sbt/protocol/JsonRpcReader.scala
seroperson: I'll just revert this, not much profit actually. 
Didn't aware about this thing, I'll fix the same pattern in `ServerSession` too then.

protocol/src/main/scala/sbt/protocol/ServerSession.scala
seroperson: The situation is so:
- There is a `ServerConnection` in `main-command`, which needs the `readThread` logic, but doesn't need the rest.
- `server-test` tests need all the `initialize`, `waitFor ...`, but they are in `testpkg` package and can't access `sbt`.
- And also it would be good to have all these `waitFor ...` and `sendJsonRpc` methods to be available from outside, like for plugin developers.

As one of possible approaches I can suggest:
- make `ServerSession` to be the very basic class (or trait) with `readThread` logic, callbacks and `socket`. `ServerConnection` (from `main-command`) will re-use it just like it does currently (actually `ServerConnection` looks redundant in such scheme, but I don't want to wide the PR scope even more)
- `ServerSessionClient` extends this basic class too and also provide all the high-level logic like `initialize`, `shutdown`, `waitFor`, `sendJsonRpc` etc.

And we can leave `ServerSessionClient` either public or make it `private[sbt]` + move all tests from `testpkg` to `sbt` + expect plugin developers to use package workaround to use it.

seroperson: I made the following changes here:
- A basic `ServerSession` with base read/close functionality.
- `ServerConnection` (in `main-command`) extends `ServerSession`. This left unchanged.
- `sendJsonRpc`, `waitFor ...` etc are moved into the new `ServerSessionClient`, but raw-string methods are `private[sbt]`. Only methods with actual objects as parameters are publicly available.
- `ServerSession.connect` returns `ServerSessionClient`.
- Marked some things `final`.

Users can use `ServerSession.connect` to interact on high-level using `sendJsonRpc` + `waitForResult`.

protocol/src/main/scala/sbt/protocol/Defaults.scala
eed3si9n: Do we still need this file?

protocol/src/main/scala/sbt/protocol/ServerSession.scala
```diff
@@ -0,0 +1,185 @@
...
+  /** Interval between progress log messages while waiting for the portfile. */
+  private[sbt] val PortfileLogInterval: FiniteDuration = 10.seconds
+
+  /**
+   * Connects to a running sbt server using the given portfile.
+   *
+   * @param portfile the `active.json` portfile created by the sbt server
+   * @return a connected [[ServerSessionClient]] ready for [[ServerSessionClient.initialize]]
+   */
+  def connect(portfile: File): ServerSessionClient = {
```
eed3si9n: I was thinking more like this would return `ServerSession` type, but the the type itself is completely empty trait with just the methods good enough to write tests from `testpkg`, but the implementation would be completely internal, so we don't make any future implementation-level promises.

seroperson: Oh, yea, okay. I was thinking about that initially, but the problem with such approach that then `sendJsonRpc`, `waitFor ...` methods need to be implemented in `ServerConnection` from `main-command`, so that's why I did what I did. 

Then I:
- Moved everything back to `testpkg`
- `ServerSession` now is a plain trait with no implementation and is public
- `ServerSessionImpl` holds the whole implementation, it's `private[sbt]` and `ServerConnection` from `main-command` extends it.

`ServerConnection` seems a little bit weird, but it works as before at least. I added a `TODO` there to probably delete it in future.

seroperson:
I actually deleted the `ServerConnection` from `main-command`. `NetworkClient` in `main-command` now uses `ServerSession` trait too. I had to add some `private[sbt]` raw string methods to this trait, I believe we can get rid of them too by refactoring `NetworkClient`, but maybe in some future.

Also replaced sequential ids with random uuids, just like it was in `NetworkClient` with `ServerConnection`.

seroperson:
pfff, seems like `ClientTest` is still flaky. Got another fail recently. Well, it always fails after `three commands with middle failure`, I bet the reason is a weak shutdown logic in `NetworkClient`. Guards seem better in `ServerSession` and I believe this flakiness could be finally fixed after `NetworkClient` refactoring, but that's really goes too far for this PR 😄

I added 500ms gap between `ClientTest` tests and also increased thread join timeout from 5 to 10 secs, maybe it'll make situation a little bit better.

eed3si9n (approved):
Thanks!

PR #8915, opened 2026-03-17
[2.x] Add VF keys

eed3si9n:
This is a follow-up to https://github.com/sbt/sbt/pull/8766

## Problem
java.io.File cannot participate in cached task.

## Solution
This adds a few VF-equivalent keys so they can participate in cached task.


anatoliykmetyuk (approved):
LGTM!

PR #8916, opened 2026-03-17
[2.x] fix: Preserve user-specified scope axes in  command instead of silently discarding them

### LINKED ISSUE #1124: The semantics of "set every" may be surprising to users

Crystora:
### Problem

`set every` silently discards scope axes the user provides. Running:

```
set every Test / sources := Nil
```

empties `sources` in **all** scopes including `Compile`, not just `Test`. This happens because `rescope` in `SettingCompletions.setAll` strips the scope and forces `Global`.

Fixes #1124

### Solution

Changed `rescope` to keep the user's config/task/extra axes and only wildcard the project axis. When no scope is given, behavior is unchanged - all axes match everything as before.

### Test

- `scripted tests/set-every` (existing) - passed, backward compatible
- `scripted tests/set-every-scoped` (new) - runs `set every Test / sources := Nil` on a multi-project build, checks `Compile / sources` stays intact while `Test / sources` is emptied

Generated-by: Claude Opus 4.6

### DIFF
--- a/main/src/main/scala/sbt/internal/SettingCompletions.scala
+++ b/main/src/main/scala/sbt/internal/SettingCompletions.scala
@@ -55,12 +55,33 @@ private[sbt] object SettingCompletions {
     def resolve(s: Setting[?]): Seq[Setting[?]] =
       Load.transformSettings(projectScope, currentRef.build, rootProject, s :: Nil)
 
+    def thisToZero[T](axis: ScopeAxis[T]): ScopeAxis[T] =
+      axis match
+        case This => Zero
+        case a    => a
+
+    def axisMatches[T](user: ScopeAxis[T], existing: ScopeAxis[T]): Boolean =
+      user match
+        case Zero | This => true
+        case Select(_)   => user == existing
+
     def rescope[T](setting: Setting[T]): Seq[Setting[?]] = {
       val akey = setting.key.key
-      val global = ScopedKey(Global, akey)
+      val userScope = setting.key.scope
+      val baseScope = Scope(
+        Zero,
+        thisToZero(userScope.config),
+        thisToZero(userScope.task),
+        thisToZero(userScope.extra),
+      )
+      val global = ScopedKey(baseScope, akey)
       val globalSetting = resolve(Def.setting(global, setting.init, setting.pos))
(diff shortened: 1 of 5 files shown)


Crystora:
@eed3si9n Could you please review this PR? thanks

Crystora:
@eed3si9n please review my PR. I'd appreciate any feedbacks.
thank you

main/src/main/scala/sbt/internal/SettingCompletions.scala
```diff
@@ -55,12 +55,33 @@ private[sbt] object SettingCompletions {
     def resolve(s: Setting[?]): Seq[Setting[?]] =
       Load.transformSettings(projectScope, currentRef.build, rootProject, s :: Nil)
 
+    def thisToZero[T](axis: ScopeAxis[T]): ScopeAxis[T] =
+      axis match
+        case This => Zero
+        case a    => a
```
anatoliykmetyuk: Can we use `Scope.subThis` in place of this function? It does the same thing and reusing it would minimise the diff.

anatoliykmetyuk (approved):
LGTM, thanks @bittoby

Crystora:
@anatoliykmetyuk Thanks for your feedback. I have fixed!

Crystora:
@eed3si9n Could you please review this PR and merge if it has no problem? Thanks

eed3si9n (approved):
I concur with Toli's review. Thanks for the contribution!

Crystora:
Thanks!

PR #8920, opened 2026-03-18
[2.x] feat: Use sbt script in BSP config instead of hardcoded Java path

### LINKED ISSUE #7245: Should use Java from path when starting sbt via BSP config
## steps

1. Set up a Java version in PATH: `eval $(cs java --jvm adopt:11 --env)`
2. Generate the BSP config: `sbt bspConfig`
3. Switch to another Java version: `eval $(cs java --jvm adoptium:11 --env)`
4. Open Metals and connect to sbt as the build server

## problem

The sbt server starts with adopt:11.

## expectation

The sbt server should start with the current Java version in PATH.

## notes

Also if the Java version is removed, sbt server will fail to start.

Crystora:
**Problem**

`sbt bspConfig` writes the absolute path of the current Java binary into `.bsp/sbt.json`. When the user switches Java versions (via sdkman, cs java, etc.) or removes that JDK, the IDE fails to start the sbt BSP server because the hardcoded path is stale or gone.

**Solution**

When an sbt launcher script is available (via `sbt.script` system property or PATH lookup), generate:

```json
"argv": ["/path/to/sbt", "bsp"]
```

instead of the old direct java invocation. The sbt script resolves Java from PATH/JAVA_HOME at runtime, so switching Java versions no longer breaks BSP.

Falls back to the previous direct java invocation when no sbt script is found (e.g. when sbt is launched from an IDE without the script on PATH).

### Changes

- `BuildServerConnection.scala` - prefer sbt script as `argv[0]`, fall back to direct java
- `Main.scala`, `NetworkClient.scala` - recognize `"bsp"` (without dash) as a BSP command
- `sbt` (bash) - handle `bsp` command: bypass native client, pass `-bsp` to Boot
- `sbt.bat` - handle `bsp` command: bypass native client
- `sbtw/Main.scala` - handle `bsp` in residual args: bypass native client

### Test plan

- [x] Unit tests added in `BuildServerConnectionSpec` (fallback argv structure, sbtScriptInPath)
- [x] Integration test updated in `BspConfigTest` (accepts both sbt-script and java-direct formats)
- [x] `protocolProj/test` - 10/10 passed
- [x] `commandProj/test` - 42/42 passed
- [x] `scalafmtCheck` - all passed
- [x] Manual test: `publishLocalBin`, created test project, ran `bspConfig`, confirmed `.bsp/sbt.json` uses sbt script path
- [x] Manual test: `sbt bsp` starts and exits cleanly (exit code 0)
- [x] Manual test: fallback path (no sbt script) still produces direct java argv

Closes: #7245

Generated-by: Claude Opus 4.6

### DIFF
--- a/launcher-package/integration-test/src/test/scala/BspConfigTest.scala
+++ b/launcher-package/integration-test/src/test/scala/BspConfigTest.scala
@@ -49,8 +49,14 @@ object BspConfigTest extends BasicTestSuite:
 
       // Verify argv structure
       assert(argv.nonEmpty, "argv should not be empty")
-      assert(argv.head.contains("java"), s"argv should start with java command, got: ${argv.head}")
-      assert(argv.contains("-bsp"), s"argv should contain -bsp flag, got: $argv")
+      // When sbt script is available, argv uses the sbt script with "bsp" command.
+      // When not, argv falls back to direct java invocation with "-bsp" flag.
+      val usesSbtScript = argv.last == "bsp" && !argv.head.contains("java")
+      val usesJavaDirect = argv.head.contains("java") && argv.contains("-bsp")
+      assert(
+        usesSbtScript || usesJavaDirect,
+        s"argv should either use sbt script with 'bsp' command or java with '-bsp' flag, got: $argv"
+      )
 
       // Test execution of the generated argv
       // Run the BSP command with a very short timeout to verify it starts correctly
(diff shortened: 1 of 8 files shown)


Crystora:
@eed3si9n Please review this PR. I'd appreciate any feedbacks.

eed3si9n (approved):
Thanks!

unkarjedy:
Also related:
https://youtrack.jetbrains.com/issue/SCL-20865/BSP-bad-experience-when-.bsp-sbt.json-has-invalid-data
Leaving it here for the record.

PR #8923, opened 2026-03-19
[2.x] Bump `sbt-launch` to 1.6.1 for programmatic launcher exit handling

### LINKED ISSUE #7540: Allow the sbt launcher to be called programmatically without running into an `exit` call
I'd like to call the sbt launcher programmatically from a Kotlin/JVM application in order to analyze sbt projects. For that, I'm using the `org.scala-sbt:sbt-launch:1.9.9` artifact from Maven Central, and call `xsbt.boot.main()` programmatically with arguments I'd usually pass via command line.

This basically does what I want, but afterwards brings down my JVM-based host application, as somewhere as part of `xsbt.boot.main()` the `exit()` function seems to be called.

I'm no Scala guy, but it seems to me as if some rather simple refactoring could provide a programmatic entry point that avoid the `exit()` call (and maybe also takes an array of arguments rather than a CLI-like argument string).

corevibe555:
Fixes https://github.com/sbt/sbt/issues/7540

### Summary
This PR upgrades the launcher dependency to `sbt-launch` **1.6.1**, incorporating a launcher-side fix related to [issue #7540](https://github.com/sbt/sbt/issues/7540) (programmatic usage without terminating the host JVM).  

This change is scoped strictly to **version alignment**—there are no additional behavior changes in the sbt core.

### Test
-  Run the launcher test suite.
- Confirm `sbt-launch` resolves to **1.6.1** in the dependency graph.
- Test passed: 
./sbt "reload; update; launcherPackage/test"
./sbt test-

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -16,7 +16,7 @@ object Dependencies {
 
   private val sbtIO = "org.scala-sbt" %% "io" % ioVersion
 
-  val launcherVersion = "1.6.0"
+  val launcherVersion = "1.6.1"
   val launcherInterface = "org.scala-sbt" % "launcher-interface" % launcherVersion
   val rawLauncher = "org.scala-sbt" % "launcher" % launcherVersion
   val testInterface = "org.scala-sbt" % "test-interface" % "1.0"


corevibe555:
Screenshot_1
Test passed.

eed3si9n (approved):
Thanks!

sschuberth:
Thanks for you help, guys! Just asking: Will this be back-ported (and soon-ish be released) as part of the 1.x series still?

PR #8929, opened 2026-03-19
[2.x] Use `new Array` instead of `Array.fill`

xuwei-k:


eed3si9n (approved):
Thanks!

som-snytt:
Do they need to rename `Array.ofDim`? to be more appealing?

PR #8936, opened 2026-03-20
[2.x] test: Enable sbt-test/project/setting-order

xuwei-k:


sbt-app/src/sbt-test/project/setting-order/build.sbt
```diff
@@ -5,7 +5,7 @@ lazy val root = (project in file("."))
     Compile / scalacOptions += "multi-project",
     check := {
       val xs = (Compile / scalacOptions).value
-      assert(xs.toList == List("multi-project", "a", "b", "bare", "c"), s"$xs")
+      assert(xs.toList == List("a", "b", "bare", "c", "multi-project"), s"$xs")
```
eed3si9n, line 8: So this is a breaking change compared to sbt 1.x. Probably okay because of common setting?

eed3si9n (approved):
多谢!

PR #8948, opened 2026-03-20
[2.x] test: Delete `sbt-test/project/continuations`

xuwei-k:
- https://github.com/scala/scala-continuations no longer maintenance
- There are other compiler plugin scripted tests
  - https://github.com/sbt/sbt/blob/1852b4a4547bbda54c17753438f7845a30eb2b78/sbt-app/src/sbt-test/cache/compiler-plugin/build.sbt#L9
  - https://github.com/sbt/sbt/blob/1852b4a4547bbda54c17753438f7845a30eb2b78/sbt-app/src/sbt-test/plugins/dotty-compiler-plugin/changes/build.sbt#L16


eed3si9n (approved):
Thanks!

PR #8950, opened 2026-03-21
Replace defunct link

Philippus:
Link is defunct for a number of years already. Replaced with a link from the Internet Archive.


eed3si9n (approved):
Thanks!

PR #8951, opened 2026-03-21
[2.x] fix: Fix misleading error when .evaluated used outside macro context

### LINKED ISSUE #1144: Confusing error message when trying to use evaluated on an input key
Given a build.sbt of

```
version := "0.0.1"

org.scalastyle.sbt.ScalastylePlugin.Settings

val t = taskKey[Unit]("task")

t := {
  //  org.scalastyle.sbt.PluginKeys.scalastyle is an input key
  val ignored = org.scalastyle.sbt.PluginKeys.scalastyle.evaluated
  // should be
  // val ignored = org.scalastyle.sbt.PluginKeys.scalastyle.toTask("").value
}
```

and project/plugins.sbt of

```
resolvers += "sonatype-releases" at "https://oss.sonatype.org/content/repositories/releases/"

addSbtPlugin("org.scalastyle" %% "scalastyle-sbt-plugin" % "0.4.0")
```

sbt.version = 0.13.1

When reloading sbt, the following message is given:

```
C:\temp\ff\build.sbt:9: error: `value` can only be called on an input task
within a task definition macro, such as := or Def.inputTask.
  val ignored = org.scalastyle.sbt.PluginKeys.scalastyle.evaluated
                                                         ^
```

This is confusing, because it refers to value, not evaluated. What would be nice would be to say something along the lines of 'should you be using toTask()'

BrianHotopp:
## Summary

Fixes #1144

- The `@compileTimeOnly` message on `wrapInitInputTask` said `` `value` can only be called on an input task... `` when the user called `.evaluated`, not `.value`
- Changed to say `` `evaluated` `` and suggest `.toTask(" <args>").value` as an alternative

Before:
```
error: `value` can only be called on an input task within a task definition macro, such as := or Def.inputTask.
  val ignored = org.scalastyle.sbt.PluginKeys.scalastyle.evaluated
```

After:
```
error: `evaluated` can only be called on an input task within a task definition macro, such as := or Def.inputTask. To use an input task from a regular task, use `.toTask(" <args>").value` instead.
  val ignored = org.scalastyle.sbt.PluginKeys.scalastyle.evaluated
```

## Test plan

- [x] `mainSettingsProj/compile` succeeds
- This is a string literal in a `@compileTimeOnly` annotation — the compiler enforces the constraint, the message just needs to be accurate

### DIFF
--- a/main-settings/src/main/scala/sbt/std/InputWrapper.scala
+++ b/main-settings/src/main/scala/sbt/std/InputWrapper.scala
@@ -59,7 +59,7 @@ object InputWrapper:
   def `wrapInputTask_\u2603\u2603`[T](@deprecated("unused", "") in: Any): T = implDetailError
 
   @compileTimeOnly(
-    "`value` can only be called on an input task within a task definition macro, such as := or Def.inputTask."
+    "`evaluated` can only be called on an input task within a task definition macro, such as := or Def.inputTask. To use an input task from a regular task, use `.toTask(\" <args>\").value` instead."
   )
   def `wrapInitInputTask_\u2603\u2603`[T](@deprecated("unused", "") in: Any): T = implDetailError
 


eed3si9n (approved):
Thanks!

PR #8952, opened 2026-03-21
Don't inherit javacOptions in doc task

### LINKED ISSUE #1785: Improve or clarify mechanism for controlling arguments between javac and javadoc
I ran into an issue recently where we wanted to pass the debug flag `-g` to the javac compiler in our sbt build.

So we added `javacOptions += "-g"` to our build file. However, when packaging the jar for release, sbt would error out because it tried to use this flag as well for the javadoc tool which complains that it does not recognize it.

We were quite confused by this behavior at first. Luckily, @jsuereth was able to explain to me that `javacOptions in (Compile, compile) += "-g"` is what I needed to do to avoid this.

Although this works for us, in order to avoid future confusion, it would be nice to rethink this and make sure we are presenting the most intuitive and simple API to the user so that it behaves as they would expect.

BrianHotopp:
## Summary

Fixes #1785

The `doc` task inherited `javacOptions` from compile scope, so compile-only flags like `-g` were passed to `javadoc`, which rejects them. sbt's own `build.sbt` worked around this in 4 places:

```scala
Compile / doc / javacOptions := Nil
```

Set `javacOptions := Nil` inside `docTaskSettings` so the doc task starts with empty javac options by default. Removed the 4 workarounds from sbt's own build.

Users who need javadoc-specific flags can still use `doc / javacOptions += ...`.

## Test plan

- [x] `mainProj/compile` succeeds
- Removing the 4 workarounds from `build.sbt` is itself the test — if the default didn't work, sbt's own `doc` task would break

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -24,7 +24,6 @@ ThisBuild / organization := "org.scala-sbt"
 ThisBuild / description := "sbt is an interactive build tool"
 ThisBuild / licenses := List("Apache-2.0" -> url("https://github.com/sbt/sbt/blob/develop/LICENSE"))
 ThisBuild / javacOptions ++= Seq("-source", "1.8", "-target", "1.8")
-ThisBuild / Compile / doc / javacOptions := Nil
 ThisBuild / developers := List(
   Developer("harrah", "Mark Harrah", "@harrah", url("https://github.com/harrah")),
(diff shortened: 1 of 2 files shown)


BrianHotopp:
Closing — this is a breaking change since users may rely on javacOptions flowing into doc. The workaround (`Compile / doc / javacOptions := Nil`) is straightforward and documented.

PR #8956, opened 2026-03-21
[2.x] fix:use sbt.AppendSupplier instead of sbt.internal.AppenderSupplier

### LINKED ISSUE #7152: Docs, settings, and types for extra loggers and appenders are a bit out of whack
The docs on how to [add a custom logger](https://www.scala-sbt.org/1.x/docs/Howto-Logging.html#Add+a+custom+logger) say to use `extraLoggers` but this is deprecated.

Presumably the replacement is `extraAppenders` however this is a `settingKey[AppenderSupplier]` where `AppenderSupplier` is:
```
trait AppenderSupplier {
  def apply(s: ScopedKey[_]): Seq[Appender]
}
```
where `Appender` is `sbt.internal.util.Appender` and not `XAppender` (`org.apache.logging.log4j.core.Appender`) as one would expect. This is internal so I'm hesitant to use it.

There is a `ConsoleAppenderFromLog4J` but this is also internal and `private[sbt]`.

It seems like `extraLoggers` was deprecated prematurely without a viable replacement.

ckstck:
Closes: #7152 
Providing a stable public API without requiring internal package access.

### DIFF
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -58,7 +58,7 @@ object Keys {
   val showTiming = settingKey[Boolean]("If true, the command success message includes the completion time.").withRank(CSetting)
   val timingFormat = settingKey[java.text.DateFormat]("The format used for displaying the completion time.").withRank(CSetting)
   @transient
-  val extraAppenders = settingKey[AppenderSupplier]("A function that provides additional loggers for a given setting.").withRank(DSetting)
+  val extraAppenders = settingKey[AppenderSupplier]("A function that provides additional appenders for custom logging. See sbt.AppenderSupplier for usage examples.").withRank(DSetting)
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/Keys.scala
```diff
@@ -757,6 +757,7 @@ object Keys {
 
   type Streams = std.Streams[ScopedKey[?]]
   type TaskStreams = std.TaskStreams[ScopedKey[?]]
+  type AppenderSupplier = sbt.internal.AppenderSupplier
```
eed3si9n, line 760: ugh. I don't see how making `AppenderSupplier` alone is going to close "Docs, settings, and types for extra loggers and appenders are a bit out of whack" problem.

eed3si9n (changes requested):

PR #8957, opened 2026-03-21
[2.x] fix: Fixes libraryDependencySchemes on cross-platform dependencies

### LINKED ISSUE #7140: libraryDependencySchemes doesn't seem to work for Scala.JS
## overview:  
`libraryDependencySchemes` appears to work for JVM projects but not Scala.JS projects.

## steps

`sbt.version = 1.8.2`

plugins.sbt:
```
addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.12.0")
```

build.sbt:
```
lazy val root = project
    .in(file("."))
//    .enablePlugins(ScalaJSPlugin)
    .settings(
//        scalaJSUseMainModuleInitializer := true,
//        libraryDependencySchemes += "com.lihaoyi" %%% "geny" % VersionScheme.Always,
        libraryDependencies ++= Seq(
            "com.lihaoyi"   %%% "fastparse" % "2.3.3",
            "com.lihaoyi"   %%% "scalatags" % "0.12.0"
        )
    )
```

Add a main such as:
```
package debug

import fastparse._
import fastparse.NoWhitespace._

object Main {
  def number[_: P]: P[Int] = P( CharIn("0-9").rep(1).!.map(_.toInt) )

    println("Hello world!")
    println(s"parse: ${parse("1234", number(_))}")
  }
}
```

## problem

Running the above yields the following error:
```
[error] stack trace is suppressed; run last update for the full output
[error] (update) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[error] 
[error] 	* com.lihaoyi:geny_2.12:1.0.0 (early-semver) is selected over 0.6.10
[error] 	    +- com.lihaoyi:scalatags_2.12:0.12.0                  (depends on 1.0.0)
[error] 	    +- com.lihaoyi:fastparse_2.12:2.3.3                   (depends on 0.6.10)
[error] 
[error] 
[error] Total time: 0 s, completed Jan. 26, 2023, 7:12:21 a.m.
```

Uncomment the `build.sbt` line `libraryDependencySchemes += "com.lihaoyi" %%% "geny" % VersionScheme.Always,` and the program will compile and run.

Uncomment the remaining two lines in `build.sbt` and do the same using Scala.JS.  The error is generated again, in spite of the `libraryDependencySchemes` key.

## expectation

I would have expected the `libraryDependencySchemes += "com.lihaoyi" %%% "geny" % VersionScheme.Always,` setting to resolve the library conflict under Scala.JS, as it did under the JVM, but it did not.

ckstck:
Closes: #7140 

## Problem
`libraryDependencySchemes` didn't work for platforms like `Scala.JS`

## Solution
Modified `EvictionError.scala` to extract the suffix.
Changed from `_: Binary` to `b: Binary` and added `${b.suffix}` to the module name.

## Test
Unit Test: with `EvictionErrorSpec.scala`
Regression Tests: Ran all existing EvictionError tests

### DIFF
--- a/lm-core/src/main/scala/sbt/librarymanagement/EvictionError.scala
+++ b/lm-core/src/main/scala/sbt/librarymanagement/EvictionError.scala
@@ -89,18 +89,18 @@ object EvictionError {
       VersionSchemes.validateScheme(s.revision)
       val versionScheme = s.revision
       (s.crossVersion, sbvOpt) match {
-        case (_: Binary, Some("2.13")) =>
+        case (b: Binary, Some("2.13")) =>
           List(
-            (s.organization, s"${s.name}_2.13") -> versionScheme,
-            (s.organization, s"${s.name}_3") -> versionScheme
+            (s.organization, s"${s.name}${b.suffix}_2.13") -> versionScheme,
+            (s.organization, s"${s.name}${b.suffix}_3") -> versionScheme
           )
-        case (_: Binary, Some(sbv)) if sbv.startsWith("3.0") || sbv == "3" =>
+        case (b: Binary, Some(sbv)) if sbv.startsWith("3.0") || sbv == "3" =>
           List(
-            (s.organization, s"${s.name}_$sbv") -> versionScheme,
-            (s.organization, s"${s.name}_2.13") -> versionScheme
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #8960, opened 2026-03-22
[2.x] feat: Add dependencyMode setting to control classpath transitivity

### LINKED ISSUE #8942: [2.x] dependencyMode
This is a feature request proposed by soc, similar to the feature available on Bazel's rules_scala - https://github.com/bazel-contrib/rules_scala/blob/master/docs/dependency-tracking.md and https://github.com/cb372/sbt-explicit-dependencies.

## Step 1

A new setting called `dependencyMode`.

```scala
scalaVersion := "3.8.1"
// implicitly libraryDependencies also includes the standard library
libraryDependencies += "org.typelevel" %% "cats-effect" % "3.7.0"
dependencyMode := DependencyMode.direct
```

## Expectations 1

When `dependencyMode` is set to `DependencyMode.direct`, `Compile / externalClasspath` includes only the direct dependencies of `libraryDependencies` and the transitive `scala-library`:

1. org.typelevel:cats-effect_3:3.7.0
2. org.scala-lang:scala3-library_3:3.8.1
3. org.scala-lang:scala-library_3:3.8.1

Same for `Test / externalClasspath`. This means using classes from Cats Core should result in an compilation error.

## Step 2

```scala
scalaVersion := "3.8.1"
// implicitly libraryDependencies also includes the standard library
libraryDependencies += "org.typelevel" %% "cats-effect" % "3.7.0"
dependencyMode := DependencyMode.plusOne
```

## Expectations 2

When `dependencyMode` is set to `DependencyMode.plusOne`, `Compile / externalClasspath` includes the transitive dependencies of `libraryDependencies` at depth 2 (direct **plus one**) and the transitive `scala-library`:

1. org.typelevel:cats-effect_3:3.7.0
2. org.typelevel:cats-effect-kernel_3:3.7.0
3. org.typelevel:cats-effect-std_3:3.7.0
4. org.typelevel:cats-mtl_3:1.6.0
5. org.scala-lang:scala3-library_3:3.8.1
6. org.scala-lang:scala-library_3:3.8.1

Same for `Test / externalClasspath`.  This means using classes from Cats Core should result in an compilation error.

## Step 3

```scala
// This is the default value
dependencyMode := DependencyMode.transitive
```

eureka0928:
## Summary

Implements the `dependencyMode` setting proposed in #8942, adding explicit dependency tracking similar to Bazel's [rules_scala dependency tracking](https://github.com/bazel-contrib/rules_scala/blob/master/docs/dependency-tracking.md) and [sbt-explicit-dependencies](https://github.com/cb372/sbt-explicit-dependencies).

- **`DependencyMode.Transitive`** (default) — all transitive dependencies on the classpath (current behavior, zero overhead)
- **`DependencyMode.Direct`** — only declared `libraryDependencies` plus `scala-library`
- **`DependencyMode.PlusOne`** — declared dependencies plus their immediate transitive dependencies plus `scala-library`

### Usage

```scala
// Only allow direct dependencies on the compile classpath
dependencyMode := DependencyMode.Direct

// Or allow one level of transitive deps
dependencyMode := DependencyMode.PlusOne
```

### Design decisions

- **Filtering at `managedClasspath` level, not resolution level** — `update` still resolves the full transitive graph. `dependencyTree`, eviction checks, and caching are unaffected.
- **Uses `Def.taskDyn`** to avoid evaluating `allDependencies`/`updateFull` when mode is `Transitive` — zero overhead for the default case.
- **Cross-version aware matching** — correctly matches `cats-core` (from `libraryDependencies`) against resolved `cats-core_3` on the classpath.
- **`scala-library` always included** regardless of mode, since the Scala standard library is a runtime requirement.

### Files changed

- `main/src/main/scala/sbt/DependencyMode.scala` — New Scala 3 enum
- `main/src/main/scala/sbt/Keys.scala` — `dependencyMode` setting key
- `main/src/main/scala/sbt/Defaults.scala` — Default value, `managedClasspath` filtering logic, helper methods in `Classpaths`
- `sbt-app/src/sbt-test/dependency-management/dependency-mode/` — Scripted integration test

## Test plan

- [x] Scripted test verifies all three modes with Scala (`cats-core`/`cats-kernel`) and Java (`guava`/`failureaccess`) dependencies
- [x] Test verifies `Direct` mode applies to both `Compile` and `Test` configurations
- [x] Default `Transitive` mode preserves existing behavior
- [ ] Manual testing with `publishLocalBin` + real projects

Fixes #8942

### DIFF
--- a/lm-core/src/main/scala/sbt/librarymanagement/DependencyMode.scala
+++ b/lm-core/src/main/scala/sbt/librarymanagement/DependencyMode.scala
@@ -6,7 +6,7 @@
  * Licensed under Apache License 2.0 (see LICENSE)
  */
 
-package sbt
+package sbt.librarymanagement
 
 /**
  * Controls which managed dependencies appear on the classpath.
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -248,6 +248,7 @@ object Defaults extends BuildCommon with DefExtra {
       internalConfigurationMap :== Configurations.internalMap,
       credentials :== SysProp.sbtCredentialsEnv.toList,
       exportJars :== true,
+      dependencyMode :== DependencyMode.Transitive,
       trackInternalDependencies :== TrackLevel.TrackAlways,
       exportToInternal :== TrackLevel.TrackAlways,
       retrieveManaged :== false,
@@ -2661,26 +2662,30 @@ object Classpaths {
         ClasspathImpl.internalDependencyClasspathTask.value
       ),
       unmanagedClasspath := Def.uncached(ClasspathImpl.unmanagedDependenciesTask.value),
-      managedClasspath := Def.uncached {
-        val converter = fileConverter.value
-        val isMeta = isMetaBuild.value
-        val force = reresolveSbtArtifacts.value
-        val app = appConfiguration.value
-        def isJansiOrJLine(f: File) = f.getName.contains("jline") || f.getName.contains("jansi")
-        val scalaInstanceJars = app.provider.scalaProvider.jars.filterNot(isJansiOrJLine)
-        val sbtCp = (scalaInstanceJars ++ app.provider.mainClasspath)
-          .map(_.toPath)
-          .map(p => converter.toVirtualFile(p): HashedVirtualFileRef)
-          .map(Attributed.blank)
-        val mjars = managedJars(
-          classpathConfiguration.value,
-          classpathTypes.value,
-          update.value,
-          converter,
-        )
-        if isMeta && !force then (mjars ++ sbtCp).distinct
-        else mjars
-      },
+      managedClasspath := Def
+        .uncached(Def.taskDyn {
+          val mode = dependencyMode.value
+          mode match
+            case DependencyMode.Transitive => managedClasspathTask
+            case DependencyMode.Direct =>
+              Def.task {
+                val mjars = managedClasspathTask.value
+                filterByDirectDeps(allDependencies.value, mjars)
+              }
+            case DependencyMode.PlusOne =>
+              Def.task {
+                val mjars = managedClasspathTask.value
+                val cpConfig = classpathConfiguration.value
+                filterByPlusOne(
+                  allDependencies.value,
+                  projectID.value,
+                  cpConfig,
+                  updateFull.value,
+                  mjars,
+                )
+              }
+        })
+        .value,
       exportedProducts := Def.uncached(
         ClasspathImpl.trackedExportedProducts(TrackLevel.TrackAlways).value
       ),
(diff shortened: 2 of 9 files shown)


lm-core/src/main/scala/sbt/librarymanagement/DependencyMode.scala
eed3si9n, line 1: Thanks for the contribution. We might eventually want to move the filtering logic to lm-core so could you

1. Move this to `sbt.librarymanagement` package - https://github.com/sbt/sbt/tree/develop/lm-core/src/main/scala/sbt/librarymanagement
2. Add to Import.scala - https://github.com/sbt/sbt/blob/develop/sbt-app/src/main/scala/sbt/Import.scala#L276-L277

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -4435,6 +4440,94 @@ object Classpaths {
       }
       .distinct
 
+  private lazy val managedClasspathTask: Initialize[Task[Classpath]] = Def.task {
```
eed3si9n, line 4443: Could you move the private implementations to https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/internal/ClasspathImpl.scala please?

eed3si9n (commented):
Some minor refactoring comments, but overall looking great!

fury0928:
@eed3si9n The review feedback has been addressed:

1. **Moved `DependencyMode` to `sbt.librarymanagement`** package in `lm-core/`
2. **Added `DependencyMode` type/val to `Import.scala`** with MiMa exclusions
3. **Moved private filter implementations to `ClasspathImpl`**

Regarding the remaining CI failure (`test (windows-latest, 17, zulu)`): this is a pre-existing flaky `ExtendedRunnerTest` — the `sbt --jvm-client` test gets `AccessDeniedException` on the portfile `active.json`. The same Windows launcher test also fails intermittently on `develop`. All other 14 checks pass.

eed3si9n (approved):
Thanks for the contribution!

soc:
Wow, thank you @eureka928! I will test and report back as soon as an artifact is available.

From my understanding 

> scala-library always included regardless of mode, since the Scala standard library is a runtime requirement.

should also be fine for us non-Scala users, because the mechanism in question does not actively add the dep, but simply does not filter it out.

So if we have configured the build to not use Scala to start with as before, this will not break anything, I think. :-)

soc:
I tried this with a very simple project and everything works fine on the CLI.
  
But it seems that any action in IntelliJ, like "build" or "test", fails with `scala: No 'scala-library*.jar' in Scala compiler classpath in Scala SDK sbt: scala-sdk-0.0.0`.  
Not sure it is SBT's fault/responsibility to fix, but wanted to let you know.

Reproduction: https://codeberg.org/soc/base-uid/src/branch/dep-direct  
Without this commit IntelliJ seems to be fine.

eed3si9n:
Thanks for testing. You can get past the initial error if you use `autoScalaLibrary := false` instead of `managedScalaInstance := false`, but I think there's an sbt bug here for the runtime classpath of `test`.

PR #8963, opened 2026-03-23
[2.0.x] bport: Add VF keys

eed3si9n:
This is a 2.0.x backport of https://github.com/sbt/sbt/pull/8915

## Problem
java.io.File cannot participate in cached task.

## Solution
This adds a few VF-equivalent keys so they can participate in cached task.


anatoliykmetyuk (approved):
LGTM

PR #8976, opened 2026-03-24
[2.x] Update ErrorMessageAuthenticator. Remove unnecessary reflection

xuwei-k:
https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/net/Authenticator.html#getDefault()

### Before

- use private field by reflection if java 8.
- use public `getDefault` method by reflection if java 9 or higher. `getDefault` method does not exists in JDK8

### After

- use `getDefault` method **without reflection**


eed3si9n (approved):
Thanks!

PR #8978, opened 2026-03-25
[2.x] fix: Send multi-arg sbtn commands sequentially instead of joining with space

### LINKED ISSUE #5969: Regression in handling multi argument command when sbtn is used
## steps

SBT 1.4.0, JDK 15, macOS Catalina 10.15.7. This command line fails:
```
sbt --client -batch clean compile
```
Output is:
```
> clean compile
[error] Expected whitespace character
[error] Expected '/'
[error] clean compile
[error]       ^
[info] clean compile completed
[error] Total time: 6 s, completed Oct 12, 2020 11:36:09 AM
```

## problem

Build fails. Compile step is not executed at all.

## expectation

Would work as
```
sbt ---batch clean compile
```
Which cleans old compilation output and compiles all scala sources.

## notes

Behavior how command line is interpreted is different when `--client` flag is used, which is a very unexpected result.

BrianHotopp:
## Summary

`sbtn clean compile` (or `sbt --client clean compile`) fails with a parse error because
`NetworkClient.batchExecute` joins the arguments with a space and sends `"clean compile"` as
a single command. The sbt command parser doesn't accept space-separated commands — it expects
either separate invocations or semicolons (`clean;compile`).

```
> clean compile
[error] Expected whitespace character
[error] Expected '/'
[error] clean compile
[error]       ^
```

The fix sends each command sequentially, stopping on the first failure, matching the behavior
of `sbt clean compile` (without `--client`).

Fixes #5969

## Test plan

- [x] `serverTestProj/testOnly testpkg.ClientTest` — 13/13 pass
  - `multi-arg commands`: `client("compile", "willSucceed")` returns 0
  - `multi-arg commands stop on first failure`: `client("willFail", "willSucceed")` returns 1
- [x] First commit (test only) fails `multi-arg commands` against unfixed code
- [x] Second commit (fix) makes it pass

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -958,8 +958,12 @@ class NetworkClient(
     }
 
   def batchExecute(userCommands: List[String]): Int = {
-    val cmd = userCommands.mkString(" ")
-    sendAndWait(cmd, None)
+    var res = 0
+    val it = userCommands.iterator
+    while (it.hasNext && res == 0) {
+      res = sendAndWait(it.next(), None)
+    }
+    res
   }
 
   def getCompletions(query: String): Seq[String] = {
--- a/server-test/src/test/scala/testpkg/ClientTest.scala
+++ b/server-test/src/test/scala/testpkg/ClientTest.scala
@@ -139,6 +139,12 @@ class ClientTest extends AbstractServerTest with BeforeAndAfterEach {
   test("three commands with middle failure") {
     assert(client("compile;willFail;willSucceed") == 1)
   }
+  test("multi-arg commands") {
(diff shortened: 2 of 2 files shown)


eed3si9n:
> `NetworkClient.batchExecute` joins the arguments with a space and sends "clean compile" as
a single command.

I think this works as intended because some commands and tasks allow arguments, and sbtn allows you to use Bash shell as sbt 1.x's shell.

The idea that was floated in #5969 (not sure if OP intended to do so) was to overload `--batch` to mean split each token to be command execution like sbt 1.x's batch execution.

```bash
# should work
$ sbt --batch clean compile

# should fail for sbt 2.x since it defaults to sbtn
$ sbt clean compile
```

BrianHotopp:
You're right — I didn't think through that commands like `testOnly com.example.MyTest` need the args joined. The space-joining is intentional. Closing this, sorry for the noise.

PR #8984, opened 2026-03-26
[2.x] fix: Stabilize progress indicator ordering to prevent jittery display

### LINKED ISSUE #5466: Progress indicator is hard to read because of unspecified ordering
## Steps

A build with a bunch of long-running tasks.

## Problem

The progress indicator is hard to read because the lines may reorder in a semi-random manner.

## Expectation

The tasks should be sorted by elapsed time (rounded to 1s), then name.

## Original report

In case there are multiple long-running tasks the progress indicator is kinda hard to read because the lines may reorder in semi-random manner.

See the gif attached.

ezgif-5-7ce5726f80d2

Crystora:
## Summary

When multiple tasks run in parallel, the super shell progress lines reorder in a semi-random manner on every refresh because they are sorted by raw elapsed microseconds. Tasks that start at nearly the same time constantly swap positions due to sub-second timing jitter, making the display hard to read.

This fixes the sort key to use elapsed seconds (rounded down) with task name as a tiebreaker, matching the granularity already shown to the user (e.g. `3s`). Tasks displaying the same elapsed second now stay alphabetically stable instead of flickering.

## Changes

- `TaskProgress.scala`: Changed `tasks.sortBy(_._2)` to sort by `(elapsed / 1000000L, taskName)` - seconds then name
- `TaskProgressSpec.scala`: Added 7 unit tests covering sort stability, second-boundary behavior, and the jitter scenario from the issue

Closes #5466

### DIFF
--- a/main/src/main/scala/sbt/internal/TaskProgress.scala
+++ b/main/src/main/scala/sbt/internal/TaskProgress.scala
@@ -170,7 +170,12 @@ private[sbt] class TaskProgress(
       val currentTasksCount = currentTasks.size
       def event(tasks: Vector[(TaskId[?], Long)]): ProgressEvent = {
         if (tasks.nonEmpty) nextReport.set(Deadline.now + sleepDuration)
-        val toWrite = tasks.sortBy(_._2)
+        // Sort by elapsed seconds (rounded down), then by name for stability.
+        // Without rounding, microsecond jitter causes tasks with similar start times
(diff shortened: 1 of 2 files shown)


Crystora:
@eed3si9n could you please review this PR? thanks

eed3si9n (approved):
Thanks!

PR #8985, opened 2026-03-27
[2.x] fix: Avoid following symlinks during scripted test cleanup

### LINKED ISSUE #7331: Scripted tests delete sym linked files and directories
## steps

1. Write a scripted test that creates a sym link within the scripted test directory to something outside that directory (make sure it is nothing you care about loosing)
2. Run `scripted`

## problem

The source of the link is deleted, whether that is a file or directory, it is gone.

## expectation

Nothing outside of the scripted test directory should be deleted.

## notes

Reproduction: https://github.com/steinybot/bug-reports/tree/sbt/scripted-deletes-linked-files

eureka0928:
## Summary

- Scripted batch test cleanup followed symlinks, deleting files and directories outside the test directory
- Replaced recursive glob listing with manual recursion that checks `FileAttributes.isSymbolicLink` before descending, matching the pattern already used in `Clean.scala`
- Hoisted cleanup helpers out of the per-test loop since they only depend on the (constant) temp directory

Fixes #7331

## Details

The batch cleanup in `ScriptedTests.runBatchedTests` used `view.list(base / **, !keep)` which recursively traverses symlinked directories. When a test creates a symlink pointing outside the test directory (e.g. via npm `link:` dependencies), the traversal enters the external directory and `Files.deleteIfExists` deletes the real files.

The fix uses `Glob(dir, AnyPath)` to list only immediate children and manually recurses, skipping entries where `attrs.isSymbolicLink` is true. Symlinks themselves are still deleted (just the link, not the target), which is standard `Files.deleteIfExists` behavior on symlinks.

## Test plan

- [x] `scriptedSbtProj/compile` passes
- [ ] Existing scripted tests pass (CI)

### DIFF
--- a/scripted-sbt/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
+++ b/scripted-sbt/src/main/scala/sbt/scriptedtest/ScriptedTests.scala
@@ -225,6 +225,27 @@ final class ScriptedTests(
     val seqHandlers = handlers.values.toList
     runner.initStates(states, seqHandlers)
 
+    // Delete test contents between runs, but preserve global dirs.
+    // Uses manual recursion to avoid following symlinks (sbt/sbt#7331).
+    val view = sbt.nio.file.FileTreeView.default
+    val base = tempTestDir.getCanonicalFile.toGlob
+    val global = base / "global"
+    val globalLogging = base / ** / "global-logging"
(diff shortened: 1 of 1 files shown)


fury0928:
Hi @eed3si9n would you review this?
Thank you

eed3si9n (approved):
Thanks!

PR #8988, opened 2026-03-29
[2.x] fix: Preserve configuration visibility in ivy.xml when publishing with coursier

### LINKED ISSUE #5455: visibility of <conf /> in ivy.xml is always public when publishing with coursier
## steps

1. Run `sbt publishLocal` with coursier enabled (which is the default for sbt 1.3 and above)

## problem

The `scala-tool` config in `ivy.xml` is `public`, despite it is [defined as hidden](https://github.com/sbt/librarymanagement/blob/v1.3.1/core/src/main/scala/sbt/librarymanagement/ConfigurationExtra.scala#L50) (which equals to `visibility = private`):

```xml
<conf name="scala-tool" visibility="public" description=""/>
```

## expectation

The `visibility` of the `scala-tool` conf should be `private` in `ivy.xml`.

## notes

Disabling coursier makes it work (`visibility="private"`).

The problem is following line. As you can see `visibility` is hardcoded to `public`:
https://github.com/sbt/sbt/blob/3c207af2e5aabcc313f62fa13038869aeb5b7310/main/src/main/scala/sbt/internal/librarymanagement/IvyXml.scala#L129

That line was added when introducing coursier: https://github.com/sbt/sbt/pull/4614/files#diff-acf5c4492361da80b925653e9e60bdc1R123

eureka0928:
## Summary

- When publishing with coursier, the `visibility` attribute of `<conf />` elements in `ivy.xml` was hardcoded to `"public"`, ignoring the actual configuration visibility
- Configurations defined as hidden (e.g. `scala-tool`) were incorrectly published with `visibility="public"` instead of `"private"`
- The root cause was that `Inputs.configExtendsSeq` only extracted config names and extends relationships, discarding the `isPublic` flag from sbt's `Configuration` objects

Fixes #5455

## Changes

- Added `privateConfigs: Set[Configuration]` field to the lmcoursier `Project` data class (with `@since` for backward compatibility)
- Added `Inputs.privateConfigs` to extract non-public configuration names during the sbt-to-coursier conversion
- Both `IvyXml` implementations now emit `visibility="private"` for configurations where `isPublic` is false

## Test plan

- [x] `lmCoursierDefinitions/compile`, `lmCoursier/compile`, `sbtIvyProj/compile`, `mainProj/compile` all pass
- [ ] Existing tests pass (CI)

### DIFF
--- a/lm-coursier/definitions/src/main/scala/lmcoursier/definitions/Project.scala
+++ b/lm-coursier/definitions/src/main/scala/lmcoursier/definitions/Project.scala
@@ -1,6 +1,6 @@
 package lmcoursier.definitions
 
-import dataclass.data
+import dataclass.{ data, since }
 
 @data class Project(
     module: Module,
@@ -10,5 +10,6 @@ import dataclass.data
     properties: Seq[(String, String)],
     packagingOpt: Option[Type],
     publications: Seq[(Configuration, Publication)],
-    info: Info
+    info: Info,
+    @since privateConfigs: Set[Configuration] = Set.empty
 )
--- a/lm-coursier/src/main/scala/lmcoursier/Inputs.scala
+++ b/lm-coursier/src/main/scala/lmcoursier/Inputs.scala
@@ -21,6 +21,11 @@ object Inputs {
     configurations
       .map(cfg => Configuration(cfg.name) -> cfg.extendsConfigs.map(c => Configuration(c.name)))
 
+  def privateConfigs(
+      configurations: Seq[sbt.librarymanagement.Configuration]
+  ): Set[Configuration] =
+    configurations.filterNot(_.isPublic).map(cfg => Configuration(cfg.name)).toSet
+
   def coursierConfigurationsMap(
       configurations: Seq[sbt.librarymanagement.Configuration]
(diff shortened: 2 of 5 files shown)


fury0928:
**CI Failure Analysis: Server Test**

The failing `Server Test` workflow appears to be a **transient infrastructure issue**, not related to this PR's changes.

**Root cause:** The sbt server instance could not download `org.scala-sbt:sbt:2.0.0-RC9-bin-SNAPSHOT` during test setup:
```
[error] [launcher] Error downloading org.scala-sbt:sbt:2.0.0-RC9-bin-SNAPSHOT
[error] [launcher] could not retrieve sbt 2.0.0-RC9-bin-SNAPSHOT
```

This caused `ClientTest` to receive empty responses (5 failures), and the remaining suites (`EventsTest`, `ResponseTest`, `BuildServerTest`, `ClientNoSubscriptionTest`) failed with cascading `Too many open files` errors.

The `Server Test` workflow passes consistently on `develop` and other recent PR branches. Could you re-run the failed job?

eed3si9n:
Rerunning now.

eed3si9n (approved):
Thanks!

PR #8990, opened 2026-03-29
[2.x] fix: Scope dependencyMode filtering to compilation only

### LINKED ISSUE #8989: [2.x] DependencyMode affects test classpath
## Steps

This is using sbt 2.0.0-RC10

### build.sbt

```scala
import sbt.librarymanagement.DependencyMode.Direct

autoScalaLibrary := false
libraryDependencies ++= Seq(
  "junit" % "junit" % "4.13.2" % Test,
  "com.github.sbt" % "junit-interface" % "0.13.2" % Test,
)
dependencyMode := Direct
```

### src/test/java/com/example/HelloTest.java

```java
package com.example;

import org.junit.Test;
import static org.junit.Assert.assertEquals;

public class HelloTest {
  @Test
  public void testSum() {
    assertEquals(1 + 1, 2);
  }
}
```

Then from sbt shell run `show Test/fullClasspath`

## Problem

```
> show Test/dependencyClasspath
[info] * Attributed(${OUT}/jvm/scala-3.8.2/junit-test/junit-test_3-0.1.0-SNAPSHOT-tests.jar>sha256-c8470a2efd64c44486f98c4b1a345ab78b8233ba0cc88c1275ce06ba1d726f57/894)
[info] * Attributed(${OUT}/jvm/scala-3.8.2/junit-test/junit-test_3-0.1.0-SNAPSHOT.jar>sha256-4eb32f45d2ab8b314d173cd6964c23a9767f797a47beecfad34bdfd88674d59d/278)
[info] * Attributed(${CSR_CACHE}/https/repo1.maven.org/maven2/junit/junit/4.13.2/junit-4.13.2.jar)
[info] * Attributed(${CSR_CACHE}/https/repo1.maven.org/maven2/com/github/sbt/junit-interface/0.13.2/junit-interface-0.13.2.jar)
```

## Expectation

We want the compile-time classpath to be filtered, but the runtime-related classpath to be intact. This means it should include:

```
[info] * Attributed(${CSR_CACHE}/https/repo1.maven.org/maven2/org/hamcrest/hamcrest-core/1.3/hamcrest-core-1.3.jar)
```

Instead of changing the `managedClasspath`, we should define a new task `filteredDependencyClasspath`, and let `dependencyPicklePath` use that in place of `dependencyClasspath`. That way, only `Compile / compile` and `Test / compile` would be affected by the filtering.

_Originally posted by @soc in https://github.com/sbt/sbt/issues/8960#issuecomment-4148927770_

eureka0928:
## Summary
- Moves `dependencyMode` filtering from `managedClasspath` to a new `filteredDependencyClasspath` task
- Wires `dependencyPicklePath` (compiler classpath) to use the filtered version
- Runtime classpaths (`dependencyClasspath`, `fullClasspath`) remain unfiltered, preserving transitive deps for test execution

Fixes #8989

## Problem
When `dependencyMode := Direct` is set, transitive dependencies like `hamcrest-core` (pulled by `junit`) were removed from `Test / dependencyClasspath`, causing runtime test failures. The filtering was applied at `managedClasspath`, which affected all downstream classpaths.

## Solution
- `managedClasspath` — always returns the full transitive classpath (no filtering)
- `filteredDependencyClasspath` — new task that applies `dependencyMode` filtering to `dependencyClasspath`
- `dependencyPicklePath` — uses `filteredDependencyClasspath` so the compiler still sees the filtered classpath
- `ClasspathImpl.filterByDependencyMode` — shared helper to reduce duplication

## Test plan
- [ ] Updated `dependency-management/dependency-mode` scripted test to verify `managedClasspath` is always transitive and `filteredDependencyClasspath` handles filtering
- [ ] New `dependency-management/dependency-mode-test-classpath` scripted test reproducing the exact issue (junit + hamcrest-core with `Direct` mode)
- [ ] `mainProj/test` — all 162 unit tests pass
- [ ] `mainProj/mimaReportBinaryIssues` — no binary compatibility issues

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2662,26 +2662,21 @@ object Classpaths {
         ClasspathImpl.internalDependencyClasspathTask.value
       ),
       unmanagedClasspath := Def.uncached(ClasspathImpl.unmanagedDependenciesTask.value),
-      managedClasspath := Def
+      managedClasspath := Def.uncached(managedClasspathTask).value,
+      filteredDependencyClasspath := Def
         .uncached(Def.taskDyn {
-          val mode = dependencyMode.value
-          mode match
-            case DependencyMode.Transitive => managedClasspathTask
-            case DependencyMode.Direct =>
+          dependencyMode.value match
+            case DependencyMode.Transitive =>
+              Def.task { dependencyClasspath.value }
+            case _ =>
               Def.task {
-                val mjars = managedClasspathTask.value
-                ClasspathImpl.filterByDirectDeps(allDependencies.value, mjars)
-              }
-            case DependencyMode.PlusOne =>
-              Def.task {
-                val mjars = managedClasspathTask.value
-                val cpConfig = classpathConfiguration.value
-                ClasspathImpl.filterByPlusOne(
+                ClasspathImpl.filterByDependencyMode(
+                  dependencyMode.value,
                   allDependencies.value,
                   projectID.value,
-                  cpConfig,
+                  classpathConfiguration.value,
                   updateFull.value,
-                  mjars,
+                  dependencyClasspath.value,
                 )
               }
         })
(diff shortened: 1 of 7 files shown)


fury0928:
Hi @eed3si9n would you review this PR?
Thank you

eed3si9n (approved):
Thanks!

PR #8993, opened 2026-03-30
[2.x] feat: Add cacheVersion setting for global cache invalidation

### LINKED ISSUE #8992: [2.x] cacheVersion
## Steps

`compile` some build, then

```scala
Global / cacheVersion := 1L
 // default to reading system property `sbt.cacheversion`, or else 0L
```

`compile` again.

## Expectations

Changing `Global / cacheVersion` invalidates all cache. We can use that as an escape hatch when we want to wipe out the cache.

## Notes

`ActionCache.cache(...)` has a parameter named `extraCache` reserved for these internal purpose. See https://github.com/sbt/sbt/blob/3dd7f939cfce89d8f62d9f237e262c21e2a66fb6/core-macros/src/main/scala/sbt/internal/util/appmacro/Cont.scala#L371-L391

eureka0928:
## Summary
- Adds `Global / cacheVersion` setting that invalidates all task caches when changed
- Defaults to system property `sbt.cacheversion`, or `0L`
- Backward compatible: when `cacheVersion` is `0L`, cache hashes are identical to previous behavior

Fixes #8992

## Usage

```scala
// in build.sbt
Global / cacheVersion := 1L
```

Or via system property:
```
sbt -Dsbt.cacheversion=1
```

## Implementation

The existing `extraHash` parameter in `ActionCache.cache()` was reserved for this purpose. The `cacheVersion` value flows through:

1. `Global / cacheVersion` setting (in `Keys.scala`)
2. `BuildWideCacheConfiguration.cacheVersion` field
3. `ActionCache.mkInput()` — combines with `extraHash` when non-zero to alter the cache key digest

## Test plan
- [ ] `mainProj/test` — 162 unit tests pass
- [ ] `utilCache/test` — 37 cache tests pass
- [ ] `mimaReportBinaryIssues` — no binary compatibility issues
- [ ] New `actions/cache-version` scripted test verifies compile/set cacheVersion/recompile flow

### DIFF
--- a/sbt-app/src/sbt-test/actions/cache-version/test
+++ b/sbt-app/src/sbt-test/actions/cache-version/test
@@ -1,0 +1,14 @@
+# First compile: should be onsite (no cache)
+> compile
+
+# Second compile: should hit disk cache
+> compile
+
+# Change cacheVersion to invalidate all caches
+> set Global / cacheVersion := 1L
+
+# Third compile: should recompute (cache invalidated by version change)
+> compile
+
+# Fourth compile: should hit disk cache again
+> compile
--- a/sbt-app/src/sbt-test/actions/cache-version/test
+++ b/sbt-app/src/sbt-test/actions/cache-version/test
@@ -1,14 +1,21 @@
-# First compile: should be onsite (no cache)
-> compile
+# bumpCounter is a cached task that increments a file counter on each execution.
+# When the cache is hit, the task body is NOT re-executed, so the counter stays the same.
 
-# Second compile: should hit disk cache
-> compile
+# First call: body runs, counter goes to 1
+> bumpCounter
+> checkCounter 1
 
-# Change cacheVersion to invalidate all caches
+# Second call: cache hit, body does NOT run, counter stays at 1
+> bumpCounter
+> checkCounter 1
+
+# Change cacheVersion: invalidates all caches
 > set Global / cacheVersion := 1L
 
-# Third compile: should recompute (cache invalidated by version change)
-> compile
+# Third call: cache miss (version changed), body runs, counter goes to 2
+> bumpCounter
+> checkCounter 2
 
-# Fourth compile: should hit disk cache again
-> compile
+# Fourth call: cache hit again, counter stays at 2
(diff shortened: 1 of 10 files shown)


fury0928:
@eed3si9n This implements the `cacheVersion` setting from #8992. Ready for review when you have a chance.

fury0928:
@soc would you review and merge this PR?

fury0928:
Follow up on this

eed3si9n (approved):
Thanks!

sbt-app/src/sbt-test/actions/cache-version/test
```diff
@@ -0,0 +1,14 @@
...
+# Second compile: should hit disk cache
+> compile
+
+# Change cacheVersion to invalidate all caches
+> set Global / cacheVersion := 1L
+
+# Third compile: should recompute (cache invalidated by version change)
+> compile
+
+# Fourth compile: should hit disk cache again
```
anatoliykmetyuk: How is hit vs non-hit tested by the tests?

anatoliykmetyuk (approved):
Thank you @eureka928! LGTM, however the unit test needs some attention - left a comment.

Should be squash-merged as some commits were a reversal of previous commits.

We should also document this feature somewhere, e.g. in https://github.com/sbt/sbt/tree/develop/notes/2.0.0

sbt-app/src/sbt-test/actions/cache-version/test
fury0928: Good point. Updated the scripted test to use a counter-based approach:

- `bumpCounter` is a cached task that increments a file counter each time its body executes
- On cache hit, the body is skipped, so the counter stays the same
- The test asserts: run → counter=1, run again → counter=1 (cached), set cacheVersion=1 → run → counter=2 (invalidated), run again → counter=2 (cached)

Also added a release note in `notes/2.0.0/cache-version.md`.

anatoliykmetyuk (approved):

fury0928:
@anatoliykmetyuk would you merge? cc: @eed3si9n

PR #8998, opened 2026-03-30
[2.x] Fix old shell syntax warning

xuwei-k:
> [warn] sbt 0.13 shell syntax is deprecated; use slash syntax instead: RepoOverrideTest / scripted


eed3si9n (approved):
Thanks!

PR #9001, opened 2026-03-31
[2.x] fix: Resolve relative paths from -sbt-dir / sbt.global.base

BrianHotopp:
## Summary

Using relative paths with \`-sbt-dir\` in \`.sbtopts\` (e.g. \`-sbt-dir ../sbt_cache\`) causes an assertion failure:

\`\`\`
java.lang.AssertionError: assertion failed: Not absolute: ../sbt_cache/plugins
\`\`\`

\`BuildPaths.getFileProperty\` returned relative \`File\` objects, which \`IO.directoryURI\` rejects. The fix calls \`.getAbsoluteFile\` on the constructed \`File\`.

Also fixes misleading \`SemanticSelector\` scaladoc that claimed \`<=1.0\` is equivalent to \`<1.1.0\` — they differ for pre-release versions (#4423).

Fixes #3729

## Test plan

- [x] \`mainProj/testOnly sbt.BuildPathsTest\` — 13/13 pass
  - \`getFileProperty resolves relative paths to absolute\`: fails without fix, passes with fix
  - \`getFileProperty preserves absolute paths\`: passes in both cases
- [x] Manual repro: \`-sbt-dir ../sbt_cache_test\` in \`.sbtopts\` crashes without fix

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9002, opened 2026-03-31
[2.x] fix: Return JSON-RPC error for unknown server methods

### LINKED ISSUE #6637: [server] Return error if request is invalid
## steps

In an sbt project
1. Run `sbt -bsp`
2. Send a false request:
```json
{"jsonrpc":"2.0", "id":"232", "method":"build/foo"}
```

## problem

sbt server never replies to it

## expectation

It should respond an error message like:
```
{"jsonrpc":"2.0", "id":"232", "error":{"code":"-32600", "message": "build/foo is an invalid request"}}
```

BrianHotopp:
## Summary

The sbt server silently drops JSON-RPC requests with unknown methods. Per the JSON-RPC spec, it should respond with error code `-32601` (Method not found). This can cause timeouts in LSP clients like Metals that send requests the server doesn't recognize.

The fallback `ServerHandler` now calls `jsonRpcRespondError` instead of only logging at debug level.

Fixes #6637

## Test plan

- [x] `serverTestProj/testOnly testpkg.ResponseTest` — 9/9 pass
  - `unknown method returns error`: sends `build/foo`, asserts error code `-32601`
- [x] First commit (test only) fails without the fix: times out waiting for response
- [x] Second commit (fix) makes it pass

### DIFF
--- a/main-command/src/main/scala/sbt/internal/server/ServerHandler.scala
+++ b/main-command/src/main/scala/sbt/internal/server/ServerHandler.scala
@@ -30,7 +30,10 @@ object ServerHandler {
 
   lazy val fallback: ServerHandler = ServerHandler({ handler =>
     ServerIntent(
-      onRequest = { case x => handler.log.debug(s"Unhandled request received: ${x.method}: $x") },
+      onRequest = { case x =>
+        handler.log.debug(s"Unhandled request received: ${x.method}: $x")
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks for the contribution! Could you remove the unrelated commits plz?

BrianHotopp:
Done — rebased on develop now that #9001 is merged.

eed3si9n (approved):
Thanks!

PR #9003, opened 2026-03-31
[2.x] fix: Add --fail to curl in launcher to prevent corrupt jars

### LINKED ISSUE #6654: Launcher jar left corrupt when bootstrap download "successfully fails"
## steps

```sh
# on a fresh install, i.e.
#   rm -rf ~/.sbt/ ~/.cache/
# and with curl in path:
SBT_LAUNCH_REPO=https://www.google.com/nope ../sbt -d ""
```

## problem

Typical situation: we have a private Artifactory, or other local caching server we need to pull dependencies through, or we have the HTTP proxy vars set incorrectly for whatever reason, or as in the example, the $SBT_LAUNCH_REPO value is simply set incorrectly:

Bootstrapping fails and then subsequent invocation of `sbt` fail _immediately_ because some HTML body e.g. "404 Not Found" was written to the the boot-jarfile:

```
Error: Invalid or corrupt jarfile /root/.cache/sbt/boot/sbt-launch/1.5.5/sbt-launch-1.5.5.jar
```

## expectation
 
`sbt` should detect and cleanup any failed download, **not** leaveing the system in a state where we are unable to launch `sbt` again.

## notes

This _should_ be fixable by asking `curl` to exit with fail with non-zero exit code, and checking `$?` instead of just testing `-f`.

```
(man curl)
...
       -f, --fail
              (HTTP) Fail silently (no output at all) on server errors. This is mostly done to better enable
              scripts  etc to better deal with failed attempts. In normal cases when an HTTP server fails to
              deliver a document, it returns an HTML document stating so (which often also describes why and
              more). This flag will prevent curl from outputting that and return error 22.

              This  method is not fail-safe and there are occasions where non-successful response codes will
              slip through, especially when authentication is involved (response codes 401 and 407).
```

~I have yet to test how this works~ It fails in a similar fashion when `wget` is the only client available, but I believe `wget` at least already considers HTTP errors, and exits with non-zero. But it needs to be checked for.

BrianHotopp:
## Summary

The `download_url` function in the `sbt` launcher script uses `curl` without `--fail`. When the download URL returns an HTTP error (404, 500, etc.), `curl` exits 0 and writes the error page HTML to the output file, creating a corrupt jar. Subsequent sbt runs then fail with confusing errors because the "jar" is actually HTML.

```
$ curl --silent -L "https://repo1.maven.org/maven2/.../sbt-launch-99.99.99.jar" --output test.jar
$ file test.jar
test.jar: HTML document, ASCII text
```

With `--fail`, `curl` returns exit code 22 on HTTP errors and does not write the response body.

Fixes #6654

## Test plan

```bash
# Without --fail: saves HTML as jar, exits 0
$ curl --silent -L "$bad_url" --output test.jar && file test.jar
test.jar: HTML document, ASCII text

# With --fail: no file created, exits 22
$ curl --fail --silent -L "$bad_url" --output test.jar; echo $?
22
```

### DIFF
--- a/sbt
+++ b/sbt
@@ -142,7 +142,7 @@ download_url () {
   local jar="$2"
   mkdir -p $(dirname "$jar") && {
     if command -v curl > /dev/null; then
-      curl --silent -L "$url" --output "$jar"
+      curl --fail --silent -L "$url" --output "$jar"
     elif command -v wget > /dev/null; then
       wget --quiet -O "$jar" "$url"
     else


eed3si9n:
Thanks for the contribution. Could you remove the other commits from the PR please?

BrianHotopp:
Done — rebased on develop, just the one commit now.

eed3si9n (approved):
Thanks!

Issue #8921, opened 2026-03-18
issues with console when usePipelining is true


OlegYch:
## steps

using sbt 1.12.6 scala 3.8.1 in a multi-project build on Windows
build.sbt:
```
ThisBuild / usePipelining := true
ThisBuild / scalaVersion := "3.8.1"
lazy val subproject = project.in(file("modules/subproject"))
lazy val root = project.in(file(".")).dependsOn(subproject).aggregate(subproject)
```
## problem

`subproject/console`
results in exception:
```
[error] java.nio.file.FileSystemException: D:\modules\subproject\target\scala-3.8.1\early\early.jar: The process cannot access the file because it is being used by another process
[error]         at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:92)
[error]         at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
[error]         at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
[error]         at java.base/sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:269)
[error]         at java.base/sun.nio.fs.AbstractFileSystemProvider.deleteIfExists(AbstractFileSystemProvider.java:110)
[error]         at java.base/java.nio.file.Files.deleteIfExists(Files.java:1087)
[error]         at dotty.tools.io.Path.delete(Path.scala:239)
[error]         at dotty.tools.io.JarArchive$.create(JarArchive.scala:26)
[error]         at dotty.tools.dotc.config.Settings$Setting.output$1(Settings.scala:246)
[error]         at dotty.tools.dotc.config.Settings$Setting.$anonfun$4(Settings.scala:248)
[error]         at dotty.tools.dotc.config.Settings$Setting.doUpdate$1(Settings.scala:197)
[error]         at dotty.tools.dotc.config.Settings$Setting.update$1(Settings.scala:198)
[error]         at dotty.tools.dotc.config.Settings$Setting.setOutput$1(Settings.scala:248)
[error]         at dotty.tools.dotc.config.Settings$Setting.doSet$2(Settings.scala:273)
[error]         at dotty.tools.dotc.config.Settings$Setting.doSet$1(Settings.scala:277)
[error]         at dotty.tools.dotc.config.Settings$Setting.tryToSet(Settings.scala:333)
[...]

PR #9011, opened 2026-04-02
[2.x] fix: resolve virtual path in -Ypickle-write scalac option

corevibe555:
## Problem

When `usePipelining := true`, sbt passes `earlyOutput.value.toString` directly
as the `-Ypickle-write` argument. Since `earlyOutput` is a `HashedVirtualFileRef`,
`.toString` produces a virtual file reference (e.g. `${BASE}/target/early/early.jar`)
rather than an actual filesystem path that `scalac` can resolve.

## Solution

Use `fileConverter.value.toPath(earlyOutput.value).toString` to convert the
virtual file reference to a real filesystem path before passing it to `-Ypickle-write`.

## Note

Split out from #8921 fix per reviewer request — the console REPL fix is in a
separate PR.

---

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1029,7 +1029,11 @@ object Defaults extends BuildCommon with DefExtra {
           )
           if (shouldApplyFlags)
             Def.uncached(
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #9015, opened 2026-04-03
[2.x] Organize scalatest dependency

xuwei-k:
https://www.scalatest.org/release_notes/3.2.0

Avoid `All in one` dependency

image


eed3si9n (approved):
Thanks!

PR #9032, opened 2026-04-06
[2.x] Reject java.nio.file.Path cached task output type

### LINKED ISSUE #9014: [2.x] `java.nio.file.Path` yields incorrect cache
## steps

- sbt 2.0.0-RC10
- same as https://github.com/sbt/sbt/issues/8762

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/9014

**Problem**
Like File, Path normally captures the absolute path, so likely not a good candidate for caching.

**Solution**
Ban it.

### DIFF
--- a/core-macros/src/main/scala/sbt/internal/util/appmacro/Cont.scala
+++ b/core-macros/src/main/scala/sbt/internal/util/appmacro/Cont.scala
@@ -356,7 +356,7 @@ trait Cont:
       )(body: Expr[A1], input: Expr[A2]): Expr[A1] =
         if containsFileType[A1] then
           report.errorAndAbort(
-            s"""java.io.File and Path are not a valid output type for a cached task.
+            s"""java.io.File and Path are not valid output types for a cached task.
                |Consider using one of the following alternatives:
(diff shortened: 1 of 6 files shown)


xuwei-k (approved):

core-macros/src/main/scala/sbt/internal/util/appmacro/Cont.scala
```diff
@@ -354,7 +356,7 @@ trait Cont:
       )(body: Expr[A1], input: Expr[A2]): Expr[A1] =
         if containsFileType[A1] then
           report.errorAndAbort(
-            s"""java.io.File is not a valid output type for a cached task.
+            s"""java.io.File and Path are not a valid output type for a cached task.
```
anatoliykmetyuk:
```suggestion
            s"""java.io.File and Path are not valid output types for a cached task.
```

anatoliykmetyuk (approved):
LGTM, added a minor grammar point. Also would be good to add a test similar to `sbt-app/src/sbt-test/cache/skip-file-cache`.

PR #9047, opened 2026-04-09
[2.x] fix: Fix FileAlreadyExistsException in packageDirectory

### LINKED ISSUE #9043: [2.x] Intermittent `FileAlreadyExistsException` with caching
## steps

Unfortunately I can't reliably reproduce it, but I see it during scripted tests every now and then:

- https://github.com/mblink/find-unused/actions/runs/23747766637/job/69180956785

## problem

```
java.nio.file.FileAlreadyExistsException: /tmp/sbt_e87af63/target/out/jvm/scala-3.8.3/root/classes.sbtdir.zip
 at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:94)
 at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
 at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
 at java.base/sun.nio.fs.UnixCopyFile.copyFile(UnixCopyFile.java:246)
 at java.base/sun.nio.fs.UnixCopyFile.copy(UnixCopyFile.java:603)
 at java.base/sun.nio.fs.UnixFileSystemProvider.copy(UnixFileSystemProvider.java:257)
 at java.base/java.nio.file.Files.copy(Files.java:1305)
 at sbt.util.ActionCache$.packageDirectory$$anonfun$1(ActionCache.scala:305)
 at sbt.io.IO$.withTemporaryDirectory(IO.scala:496)
 at sbt.io.IO$.withTemporaryDirectory(IO.scala:506)
 at sbt.util.ActionCache$.packageDirectory(ActionCache.scala:288)
 at sbt.Defaults$.$init$$$anonfun$1$$anonfun$1(Defaults.scala:2268)
 at sbt.util.ActionCache$.organicTask$1(ActionCache.scala:77)
 at sbt.util.ActionCache$.cache(ActionCache.scala:127)
 at sbt.Defaults$.$init$$$anonfun$1(Defaults.scala:2289)
 at scala.Function1.$anonfun$compose$1(Function1.scala:79)
 at scala.Function1.$anonfun$compose$1(Function1.scala:79)
 at sbt.std.Transform$$anon$3.work(Transform.scala:79)
 at sbt.std.Transform$$anon$3.work(Transform.scala:79)
 at sbt.Execute.submit$$anonfun$1$$anonfun$1(Execute.scala:283)
 at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:24)
 at sbt.Execute.work(Execute.scala:294)
 at sbt.Execute.submit$$anonfun$1(Execute.scala:283)
 at sbt.ConcurrentRestrictions$$anon$4.$anonfun$2(ConcurrentRestrictions.scala:269)
 at sbt.CompletionService$$anon$2.call(CompletionService.scala:75)
 at sbt.CompletionService$$anon$2.call(CompletionService.scala:73)
 at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
 at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
 at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
 at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
 at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
 at java.base/java.lang.Thread.run(Thread.java:840)
```

## expectation

No error should be thrown

bitloi:
**Problem**
Intermittent java.nio.file.FileAlreadyExistsException when publishing classes.sbtdir.zip during action-cache packaging under parallel tasks (#9043). Copying from a temp directory straight into the final path races on the fixed destination name.

**Solution**
Stage the built zip next to the destination with a unique temp file, then replace the final path via Files.move with REPLACE_EXISTING and ATOMIC_MOVE, falling back to a non-atomic move when needed. Add a concurrent packageDirectory test.

Closes #9043

### DIFF
--- a/util-cache/src/main/scala/sbt/util/ActionCache.scala
+++ b/util-cache/src/main/scala/sbt/util/ActionCache.scala
@@ -10,7 +10,7 @@ package sbt.util
 
 import java.io.{ File, IOException }
 import java.nio.charset.StandardCharsets
-import java.nio.file.{ Files, Path, Paths, StandardCopyOption }
+import java.nio.file.{ AtomicMoveNotSupportedException, Files, Path, Paths, StandardCopyOption }
 import sbt.internal.util.{ ActionCacheEvent, CacheEventLog, StringVirtualFile1 }
 import sbt.io.syntax.*
 import sbt.io.IO
@@ -270,6 +270,36 @@ object ActionCache:
 
   private val default2010Timestamp: Long = 1262304000000L
 
+  /**
+   * Publishes `builtZip` as `destZip` by staging next to the destination and renaming into place.
+   * Avoids races from a direct `Files.copy` into `destZip` under parallel task execution.
+   */
+  private def installPackagedZip(builtZip: Path, destZip: Path, fallbackStagingDir: Path): Unit =
+    val stagingDir = Option(destZip.getParent) match
+      case Some(parent) =>
+        Files.createDirectories(parent)
+        parent
+      case None => fallbackStagingDir
+
+    val staging = Files.createTempFile(
+      stagingDir,
+      destZip.getFileName.toString + ".",
+      dirZipExt + ".tmp",
+    )
+    try
+      Files.copy(builtZip, staging, StandardCopyOption.REPLACE_EXISTING)
+      try
+        Files.move(
+          staging,
+          destZip,
+          StandardCopyOption.REPLACE_EXISTING,
+          StandardCopyOption.ATOMIC_MOVE,
+        )
+      catch
+        case _: AtomicMoveNotSupportedException =>
+          Files.move(staging, destZip, StandardCopyOption.REPLACE_EXISTING)
+    finally Files.deleteIfExists(staging)
+
   def packageDirectory(
       dir: VirtualFileRef,
       conv: FileConverter,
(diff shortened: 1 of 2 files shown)


bitloi:
@eed3si9n Ready for review!

util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -270,6 +270,36 @@ object ActionCache:
...
+
+    val staging = Files.createTempFile(
+      stagingDir,
+      destZip.getFileName.toString + ".",
+      dirZipExt + ".tmp",
+    )
+    try
+      Files.copy(builtZip, staging, StandardCopyOption.REPLACE_EXISTING)
+      try
+        Files.move(
```
eed3si9n, line 292: Thanks for working on this. Would this make the _best case_ scenario to require both copying and moving? I feel like that might make things slower on Windows. Do you think we could call `Retry(...)` to avoid the issue instead, so the best case we just copy once.

bitloi: Thanks for the reivew. The extra step is copy once into a unique staging file next to `dest`, then `move` (typically a same-volume rename, not a second full copy of the data). I’m wary of `Retry` on `copy` into the fixed path since it doesn’t remove the parallel race; I will add bounded retry on move if you want.

bitloi:
@anatoliykmetyuk Would you review the PR?

eed3si9n (approved):
Thanks!

PR #8699, opened 2026-02-06
[2.x] fix: handle NoSuchFileException during cache storage

calm329:
## Problem

When a file referenced during cache serialization is deleted between task execution and cache storage, sbt crashes with an uncaught `NoSuchFileException`. This is intermittent and happens during
concurrent builds (e.g. scripted tests).

The root cause is a double-fault in `CacheImplicits.hashedVirtualFileRefToStr`:

1. `Files.readAttributes(path, ...)` throws `NoSuchFileException` (file was deleted)
2. The catch block calls `fallback` → `super.hashedVirtualFileRefToStr(ref)`
3. That calls `ref.contentHashStr()` which reads the file again → second `NoSuchFileException`, uncaught

Fixes #8687

## Solution

Two changes:

1. **`CacheImplicits.hashedVirtualFileRefToStr`**: Re-throw `NoSuchFileException` instead of calling `fallback` (which will also throw). This propagates the error state cleanly instead of
double-faulting.

2. **`ActionCache.organicTask`**: Wrap the cache storage portion (after the action completes successfully) in a try-catch for `NoSuchFileException`. When caught, log a debug message, record an error
event, and return the task result without caching.

This is safe because:
- The task itself already completed successfully — only cache storage fails.
- No false cache entries are created.
- Next run will re-execute the task (cache miss), which is the correct behavior.

## Testing

- `sbt utilCache/test` — all 35 tests pass, no regressions.

PR #9050, opened 2026-04-09
[2.x] fix: Fixes ActionCache.cache to be more robust

### LINKED ISSUE #9044: [2.x] Intermittent `NoSuchFileException` when reading cache file attributes
## steps

Unfortunately I can't reliably reproduce it, but I see it during scripted tests every now and then:

- https://github.com/mblink/find-unused/actions/runs/24150354721/job/70475341939
- https://github.com/mblink/find-unused/actions/runs/24150354721/job/70475341934
- https://github.com/mblink/find-unused/actions/runs/23747766637/job/69180956785

## problem

```

```

## expectation

No error should be thrown

## notes

Note sometimes the file is `test-classes.sbtdir.zip`, other times it's `root_3-0.1.0-SNAPSHOT.jar`.

I reported a similar bug in #8687 that was fixed.

Crystora:
## Problem

When a file referenced during cache serialization is deleted between task execution and cache storage, sbt crashes with an uncaught `NoSuchFileException`. This is an intermittent race condition during concurrent builds (e.g., scripted tests).

The `IOException` catch added in #8699 only wraps `store.put()`, but `Converter.toJsonUnsafe(result)` and `mkInput()` also call `hashedVirtualFileRefToStr` → `Files.readAttributes` — and these are outside the try-catch.

Stack trace from CI:

## Solution

Move the `try`/`catch IOException` to wrap the entire post-action cache storage section, so `NoSuchFileException` from result serialization or cache key computation is caught and handled gracefully (skip caching, return result).

This is safe because the task already completed successfully, only cache storage fails. Next run will re-execute the task (cache miss).

Fixes #9044

### DIFF
--- a/util-cache/src/main/scala/sbt/util/ActionCache.scala
+++ b/util-cache/src/main/scala/sbt/util/ActionCache.scala
@@ -82,40 +82,40 @@ object ActionCache:
           case e: Exception =>
             cacheEventLog.append(ActionCacheEvent.Error)
             throw e
-      val json = Converter.toJsonUnsafe(result)
-      val normalizedOutputDir = outputDirectory.toAbsolutePath.normalize()
-      val uncacheableOutputs =
-        outputs.filter(f =>
-          f match
-            case vf if vf.id.endsWith(ActionCache.dirZipExt) =>
-              false
-            case _ =>
-              val outputPath = fileConverter.toPath(f).toAbsolutePath.normalize()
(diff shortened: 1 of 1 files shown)


Crystora:
@eed3si9n Please review this PR. Welcome to any feedback. All tests passed!

eed3si9n (approved):
Thanks!

PR #9051, opened 2026-04-09
[2.x] feat: Add English aliases cross and switch for +/++ commands

### LINKED ISSUE #9099: Use cross instead of (or in addition to)
I'd like to propose adding a `cross` command to sbt as a more intuitive alternative (or complement) to the existing `+` operator for cross-building across Scala versions. While `+` works fine technically, it feels like an operator hack rather than a proper sbt command, which makes it less discoverable for new users.

Right now, common questions like "How do I cross-publish my project?" require explaining `sbt +publish`, but with a `cross` command, it could be as simple as `sbt cross publish`. This would align perfectly with sbt's existing command patterns like `test`, `publish`, `console`, and `assembly`.

The `cross` command could:
- Apply to the current project by default: `sbt cross test`
- Extend to specific subprojects: `sbt projectX/cross compile`
- Support qualifiers like `crossOnly` for subset control, similar to how `+` scopes work today
- Avoid breaking existing `+` usage, so it's purely additive for better UX

This small change would significantly improve onboarding and documentation clarity without any compatibility issues. Cross-building is one of sbt's most powerful features, and making it more approachable benefits everyone.

jonathanchang31:
Fixes #9099 
Adds discoverable English-name alternatives for the symbolic cross-build commands, making them more beginner-friendly.

### DIFF
--- a/main/src/main/scala/sbt/Cross.scala
+++ b/main/src/main/scala/sbt/Cross.scala
@@ -41,7 +41,7 @@ object Cross {
 
   private def switchParser(state: State): Parser[Switch] = {
     import DefaultParsers.*
-    def versionAndCommand(spacePresent: Boolean) = {
+    def versionAndCommand(commandName: String)(spacePresent: Boolean) = {
       val x = Project.extract(state)
       import x.*
       val knownVersions = crossVersions(x, currentRef)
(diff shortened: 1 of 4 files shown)


jonathanchang31:
re-run

main/src/main/scala/sbt/Cross.scala
```diff
@@ -56,7 +62,7 @@ object Cross {
             ScalaHomeVersion(new File(home), Some(v).filterNot(_.isEmpty), force)
         }
       }
-      val spacedVersion = if (spacePresent) version else version & spacedFirst(SwitchCommand)
+      val spacedVersion = if (spacePresent) version else version & spacedFirst(commandName)
```
eed3si9n, line 59: Thanks for the contribution. Instead of making a separate command and separate parser, would it be possible for the existing parser to handle both the symbolic name and the English alias using `|` combinator? - https://www.scala-sbt.org/2.x/docs/en/reference/tab-completion-parser.html

eed3si9n (changes requested):

eed3si9n (approved):
Thanks!

jonathanchang31:
@eed3si9n please merge

PR #9052, opened 2026-04-09
[2.x] fix: Include managed source directories in cleanFiles

### LINKED ISSUE #6895: Clean does not clean managed sources
## steps

sbt 1.6.2

```
scalaVersion := "2.13.8"

Compile / sourceManaged := file("src_managed")

Compile / sourceGenerators += Def.task {
  val file = (Compile / sourceManaged).value / "demo" / "Test.scala"
  IO.write(file, """object Test extends App { println("Hi") }""")
  Seq(file)
}.taskValue

Compile / sourceGenerators += Def.task {
  val f = file("src_managed2") / "demo" / "Test2.scala"
  IO.write(f, """object Test2 extends App { println("Hi 2") }""")
  Seq(f)
}.taskValue
```

1. `sbt compile;clean`
2. These should not exist:
   - src_managed/demo/Test.scala
   - src_managed2/demo/Test2.scala
3. `sbt compile;managedSourcePaths/clean`
4. These should not exist:
   - src_managed/demo/Test.scala
   - src_managed2/demo/Test2.scala

Reproduction here: https://github.com/steinybot/bug-reports/tree/sbt/clean-managed-sources

## problem

I had always assumed that `clean` would remove generated sources. The description says "Deletes files produced by the build, such as generated sources, compiled classes, and task caches.".

It seems that it only cleans sources if they are in `sourceManaged` and `sourceManaged` has not been relocated outside of the `target` directory.

## expectation

`clean` should clean all managed sources.

At the very least, `managedSourcePaths / clean` should remove them.

## notes

Workaround:
```
cleanFiles ++= (Compile / managedSourceDirectories).value
```

carlos4s:
## Problem

When `sourceManaged` or `resourceManaged` is set to a path outside the `target` directory, `clean` does not remove the generated files. For example:

```scala
Compile / sourceManaged := file("src_managed")
```

After `compile`, files are generated in `src_managed/`. After `clean`, they remain — only `target/` contents are deleted.

Root cause: `cleanFilesTask` returns an empty vector, and the `clean` task only deletes `target/` contents plus `cleanFiles` entries.

## Solution

Change `cleanFilesTask` to include `managedSourceDirectories` and `managedResourceDirectories` from Compile and Test configurations, filtered to only those outside the `target` directory (directories inside `target` are already cleaned by the target directory deletion).

## Test

https://asciinema.org/a/N9WnX5o26sfQaNvA

Fixes #6895

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1965,7 +1965,18 @@ object Defaults extends BuildCommon with DefExtra {
   }
 
   /** Implements `cleanFiles` task. */
-  private[sbt] def cleanFilesTask: Initialize[Task[Vector[File]]] = Def.task { Vector.empty[File] }
+  private[sbt] def cleanFilesTask: Initialize[Task[Vector[File]]] = {
+    import ScopeFilter.Make.*
+    val allConfigs = ScopeFilter(configurations = inAnyConfiguration)
+    Def.task {
+      val targetDir = target.value.toPath
+      val managedSrcDirs = managedSourceDirectories.?.all(allConfigs).value.flatten.flatten
+      val managedRscDirs = managedResourceDirectories.?.all(allConfigs).value.flatten.flatten
+      (managedSrcDirs ++ managedRscDirs)
+        .filter(d => !d.toPath.startsWith(targetDir))
+        .toVector
+    }
+  }
 
   def runMainTask(
       classpath: Initialize[Task[Classpath]],
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1965,18 +1965,7 @@ object Defaults extends BuildCommon with DefExtra {
   }
 
   /** Implements `cleanFiles` task. */
-  private[sbt] def cleanFilesTask: Initialize[Task[Vector[File]]] = {
-    import ScopeFilter.Make.*
-    val allConfigs = ScopeFilter(configurations = inAnyConfiguration)
-    Def.task {
-      val targetDir = target.value.toPath
-      val managedSrcDirs = managedSourceDirectories.?.all(allConfigs).value.flatten.flatten
-      val managedRscDirs = managedResourceDirectories.?.all(allConfigs).value.flatten.flatten
-      (managedSrcDirs ++ managedRscDirs)
-        .filter(d => !d.toPath.startsWith(targetDir))
-        .toVector
-    }
-  }
+  private[sbt] def cleanFilesTask: Initialize[Task[Vector[File]]] = Def.task { Vector.empty[File] }
 
   def runMainTask(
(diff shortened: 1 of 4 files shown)


carlos4s:
@eed3si9n Could you please review this PR? Thanks

mzuehlke:
A scripted test to avoid a regression in the future would be good.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1965,7 +1965,15 @@ object Defaults extends BuildCommon with DefExtra {
   }
 
   /** Implements `cleanFiles` task. */
-  private[sbt] def cleanFilesTask: Initialize[Task[Vector[File]]] = Def.task { Vector.empty[File] }
+  private[sbt] def cleanFilesTask: Initialize[Task[Vector[File]]] = Def.task {
```
eed3si9n: I think `cleanFiles` is kept around mostly for backward compatibility, and the `clean` is now implemented in a more complicated way in https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/internal/Clean.scala, which supports `Compile / clean` etc.

carlos4s:
@eed3si9n I've updated based on your feedback. thanks

eed3si9n:
Could you add a scripted test for this plz? (https://github.com/sbt/sbt/blob/develop/contributing-docs/05_scripted_tests.md)

main/src/main/scala/sbt/internal/Clean.scala
```diff
@@ -167,10 +167,8 @@ private[sbt] object Clean {
       })
       .flatMapTask { case scope =>
         Def.task {
-          val targetDir = (scope / target).value.toPath
           val filter = cleanFilter(scope).value
-          // We do not want to inadvertently delete files that are not in the target directory.
-          val excludeFilter: Path => Boolean = path => !path.startsWith(targetDir) || filter(path)
```
eed3si9n, line 173: This could end up deleting some random files on the machine, so I think we should keep some guardrails here. If not the target directory, we should check that it's within `baseDirectory.value`.

eed3si9n (changes requested):

carlos4s:
@eed3si9n please review again. That CI testing fails don't seem to relate to my changes. can you check that?

mzuehlke:
The test that fails is related to cleaning files:
https://github.com/sbt/sbt/blob/develop/server-test/src/test/scala/testpkg/BuildServerTest.scala#L341-L355

So I guess this is related to this PR.

carlos4s:
@mzuehlke @eed3si9n I am sure the CI testings are not relate to my changes.
My only change is in `cleanFileOutputTask`, which is a completely separate method. `Clean.task` never calls `cleanFileOutputTask`

eed3si9n:
I wanted to let you know that I've delisted sbt/sbt from Gittensor.

PR #9057, opened 2026-04-10
[2.x] fix: clear compiler bridge zinc cache on reboot dev

### LINKED ISSUE #5735: snapshot compiler bridge should be cleared
Ref https://github.com/sbt/sbt/issues/5412

## steps

1. Make changes to Zinc compiler bridge. publishLocal. Run sbt.
2. Make changes to Zinc compiler bridge. publishLocal. Run sbt.
3. Maybe reboot dev.

## problem

The second set of changes aren't reflected because it's cached in `$XDG_CONFIG_HOME/sbt/2/zinc/org.scala-sbt` for sbt 2 ( and `~/.sbt/1.0/zinc/org.scala-sbt/` for sbt 1).

## expectation

reboot dev should clear them out.

bitloi:
Closes #5735

## Checklist

- [x] Reproduced / validated the behavior (reboot dev clears bridge cache path as intended)
- [x] Tests added (`MainLoopZincCacheTest`)
- [x] `scalafmtCheck` on touched projects
- [x] Binary compatibility: no public API change (`private[sbt]` helper only)

## Problem

After iterating on the Zinc compiler bridge with `publishLocal`, the compiled bridge can stay stale because the **secondary cache** under the global zinc directory (`zinc/org.scala-sbt`, same layout as `ZincComponentManager`) is not cleared. `reboot dev` already removed boot artifacts but left this cache in place ([#5735](https://github.com/sbt/sbt/issues/5735)).

## Solution

- On `reboot dev` (`RebootCurrent`), after `deleteCurrentArtifacts`, delete  
  `getZincDirectory(state, getGlobalBase(state)) / SbtArtifacts.Organization` when it exists (uses **`SbtArtifacts.Organization`**, not `ScalaArtifacts.Organization`).
- Hoist `import sbt.io.syntax.*` in `MainLoop` for the boot/zinc delete paths.
- Update `reboot` help (`RebootDetailed`) for global base / zinc property wording.

## Verification

```bash
./sbt -batch "mainProj/Test/testOnly sbt.MainLoopZincCacheTest"
./sbt -batch "mainProj/scalafmtCheck; commandProj/scalafmtCheck"
./sbt -batch "mainProj/compile"
./sbt -batch "mainProj/mimaReportBinaryIssues"

### DIFF
--- a/main/src/main/scala/sbt/MainLoop.scala
+++ b/main/src/main/scala/sbt/MainLoop.scala
@@ -22,6 +22,7 @@ import sbt.internal.util.{
   Terminal as ITerminal
 }
 import sbt.io.{ IO, Using }
+import sbt.io.syntax.*
 import sbt.librarymanagement.SbtArtifacts
 import sbt.protocol.*
 import sbt.util.{ Logger, LoggerContext }
@@ -91,7 +92,6 @@ private[sbt] object MainLoop:
 
   /** Deletes the current sbt artifacts from boot. */
   private[sbt] def deleteCurrentArtifacts(state: State): Unit = {
-    import sbt.io.syntax.*
     val provider = state.configuration.provider
     val appId = provider.id
     // If we can obtain boot directory more accurately it'd be better.
@@ -113,7 +113,6 @@ private[sbt] object MainLoop:
 
   /** Removes the Zinc compiler bridge secondary cache (`…/zinc/org.scala-sbt`). */
   private[sbt] def deleteZincBridgeSecondaryCache(state: State): Unit =
-    import sbt.io.syntax.*
     val zincDir = BuildPaths.getZincDirectory(state, BuildPaths.getGlobalBase(state))
     val bridgeCache = zincDir / SbtArtifacts.Organization
     if bridgeCache.exists() then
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by bitloi** [comment]
@eed3si9n Would you mind reviewing the PR and give me your feedback?

File: main/src/main/scala/sbt/MainLoop.scala [hunk 1]
```diff
@@ -109,6 +111,14 @@ private[sbt] object MainLoop:
     }
   }
 
+  /** Removes the Zinc compiler bridge secondary cache (`…/zinc/org.scala-sbt`). */
+  private[sbt] def deleteZincBridgeSecondaryCache(state: State): Unit =
+    val zincDir = BuildPaths.getZincDirectory(state, BuildPaths.getGlobalBase(state))
```
eed3si9n: There's a lot of complexity around faking the state in test. Could you let this function accept `zincDir` as the parameter so the test would be shorter please?

eed3si9n (changes requested):

main/src/main/scala/sbt/MainLoop.scala
bitloi: I just updated the function.

eed3si9n (approved):
Thanks!

PR #9058, opened 2026-04-10
[2.x] fix: stack traces suppressed in thin client batch mode

### LINKED ISSUE #7322: Stack traces missing in batch mode, only when using `stream.value.log`
## steps

I have minimized the issue to this crashing task (note the commented line)

```
val hello = taskKey[Unit]("blah")

hello := {
    // val log = streams.value.log
    throw new RuntimeException("boom!")
}
```

When ran on the CLI, in batch mode, it correctly displays the stack trace

<details>

```
❯ sbt hello
[info] welcome to sbt 1.9.1 (BellSoft Java 17.0.4)
[info] loading global plugins from /Users/idragos/.sbt/1.0/plugins
...
[info] set current project to test-project (in build file:/Users/idragos/workspace/discoveries/sbt-task-progress/test-project/)
[error] java.lang.RuntimeException: boom!
[error] 	at $fc3ca1ba05bef2f053bb$.$anonfun$$sbtdef$1(/Users/idragos/workspace/discoveries/sbt-task-progress/test-project/build.sbt:51)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
[error] 	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:63)
[error] 	at sbt.std.Transform$$anon$4.work(Transform.scala:69)
[error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:283)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:24)
[error] 	at sbt.Execute.work(Execute.scala:292)
[error] 	at sbt.Execute.$anonfun$submit$1(Execute.scala:283)
[error] 	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:265)
[error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:65)
[error] 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[error] 	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
[error] 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[error] 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
[error] 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
[error] 	at java.base/java.lang.Thread.run(Thread.java:833)
[error] (hello) boom!
[error] Total time: 0 s, completed 27 Jun 2023, 14:19:56
```
</details>

## problem

When uncommenting the streams.value.log line, the stack trace is missing (only one [error] line is printed):

```
❯ sbt hello
[info] welcome to sbt 1.9.1 (BellSoft Java 17.0.4)
[info] loading global plugins from /Users/idragos/.sbt/1.0/plugins
...
[info] set current project to test-project (in build file:/Users/idragos/workspace/discoveries/sbt-task-progress/test-project/)
[error] Total time: 0 s, completed 27 Jun 2023, 14:20:06
```

This seems to be specific to `streams.value`, replacing that with another setting dependency such as `scalaVersion` doesn't change the behavior (stack trace still displayed).

## expectation

I'd expect the behavior to be consistent regardless of what task dependencies I have.

I think this must be related to `traceLevel` but I still think this is somewhat buggy, or I can't wrap my head around it. This fails a number of our tests and makes it quite fragile to maintain them.

The [default trace level](https://github.com/sbt/sbt/blob/634e8799e76395f7bb280ea7f5f910f9cce21838/main/src/main/scala/sbt/internal/LogManager.scala#L226-L227) is set to `Int.MaxValue` when not in `interactive` mode (which is the case here)...

Crystora:
## Summary

In sbt 2.x, running batch commands through the thin client (`sbtn`) suppresses stack traces for all tasks because the server's `shell` command unconditionally sets `state.interactive = true`. This causes `LogManager.defaultTraceLevel` to return `-1` (suppressed) even when the client explicitly signals non-interactive (batch) mode via `Attach(interactive=false)`.

This fixes the `shell` command to check the originating `NetworkChannel`'s interactive flag before setting `state.interactive`, so thin client batch commands correctly get `Int.MaxValue` trace level and display full stack traces.

## Changes

- `Main.scala`: Replaced hardcoded `.setInteractive(true)` with a check against the source `NetworkChannel.isInteractive` for network-originated commands
- `sbt-app/src/sbt-test/actions/streams-trace-level/`: Added scripted regression test with tasks both with and without `streams.value.log` to verify trace level behavior in batch mode

Fixes #7322

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -1144,12 +1144,20 @@ object BuiltinCommands {
         .getOpt(Keys.minForcegcInterval)
         .getOrElse(GCUtil.defaultMinForcegcInterval)
       val exec: Exec = getExec(s1, minGCInterval)
+      val isInteractive = exec.source match {
+        case Some(src) if src.channelName.startsWith("network") =>
+          exchange.channelForName(src.channelName) match {
+            case Some(nc: NetworkChannel) => nc.isInteractive
+            case _                        => true
+          }
+        case _ => true
+      }
       val newState = s1
         .copy(
           onFailure = Some(Exec(Shell, None)),
           remainingCommands = exec +: Exec(Shell, None) +: s1.remainingCommands
         )
-        .setInteractive(true)
+        .setInteractive(isInteractive)
       val res =
         if (exec.commandLine.trim.isEmpty) newState
         else newState.clearGlobalLog
--- a/sbt-app/src/sbt-test/actions/streams-trace-level/build.sbt
+++ b/sbt-app/src/sbt-test/actions/streams-trace-level/build.sbt
@@ -1,6 +1,5 @@
 lazy val helloWithoutStreams = taskKey[Unit]("")
 lazy val helloWithStreams = taskKey[Unit]("")
-lazy val checkTraceLevel = taskKey[Unit]("")
 
 helloWithoutStreams := {
   throw new RuntimeException("boom without streams!")
@@ -9,9 +8,4 @@ helloWithoutStreams := {
 helloWithStreams := {
   val log = streams.value.log
   throw new RuntimeException("boom with streams!")
-}
-
-checkTraceLevel := {
-  val level = traceLevel.value
-  assert(level != -1, s"Expected traceLevel != -1 in batch mode, but got $level")
 }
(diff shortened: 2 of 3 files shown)


Crystora:
@eed3si9n could you please review this PR?

eed3si9n (approved):
Thanks!

PR #9059, opened 2026-04-10
[2.x] Format pom.xml

xuwei-k:


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -4289,7 +4289,9 @@ object Classpaths {
             configuration.allRepositories,
           )
         val processed = configuration.process(pomXml)
-        scala.xml.XML.save(file.getAbsolutePath, processed, "UTF-8", xmlDecl = true)
+        val printer = new scala.xml.PrettyPrinter(1000, 4)
```
xuwei-k, line 4292: https://github.com/sbt/sbt/blob/e0bdb531f73bc2737fd64339614896f7ef908c5a/lm-ivy/src/main/scala/sbt/internal/librarymanagement/MakePom.scala#L78

eed3si9n (approved):
Thanks!

PR #9060, opened 2026-04-10
[2.x] Delete `bootDirectory` in `cleanFull`

xuwei-k:


eed3si9n (approved):
Thanks!

mrdziuban:
@xuwei-k @eed3si9n what was the motivation for this? I just came across this behavior in 2.0.0-RC13 when I ran `cleanFull` followed by `compile` and got an error:

```
[error] (predef / scalaCompilerBridgeBin) java.nio.file.NoSuchFileException: /Users/matt/.sbt/boot/scala-3.8.3/lib/scala-library.jar
```

The only way to work around it is to exit and restart sbt, a `reload` doesn't fix it.

eed3si9n:
`cleanFull` tries to clean all cache, so I think it's consistent to wipe out the local boot directory, which is basically a cache used by sbt launcher. Would `reboot` work?

mrdziuban:
> Would `reboot` work?

Maybe, but I can't truly test it because of #9095

Issue #6487, opened 2021-05-03
Docs source link is 404


er1c:
## steps

https://www.scala-sbt.org/1.x/api/sbt/librarymanagement/ModuleID.html

Click on ModuleID.scala => 404

## expectation

It should find the appropriate scala file (https://github.com/sbt/librarymanagement/blob/develop/core/src/main/contraband-scala/sbt/librarymanagement/ModuleID.scala)

PR #9063, opened 2026-04-11
[2.x] Add `-source-links` option

xuwei-k:
https://github.com/scala/scala3/blob/0f8dfa1716bf28f56da122dc9221422a580699b8/scaladoc/src/dotty/tools/scaladoc/SourceLinks.scala#L124-L145

## Before

image

## After

image

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -84,19 +84,21 @@ def commonSettings: Seq[Setting[?]] = Def.settings(
   Test / testOptions += Tests.Argument(TestFrameworks.ScalaCheck, "-w", "1"),
   Test / testOptions += Tests.Argument(TestFrameworks.ScalaCheck, "-verbosity", "2"),
   compile / javacOptions ++= Seq("-Xlint", "-Xlint:-serial"),
-  /*
   Compile / doc / scalacOptions ++= {
-    import scala.sys.process._
-    val devnull = ProcessLogger(_ => ())
-    val tagOrSha = ("git describe --exact-match" #|| "git rev-parse HEAD").lineStream(devnull).head
-    Seq(
-      "-sourcepath",
-      (LocalRootProject / baseDirectory).value.getAbsolutePath,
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -84,19 +84,16 @@ def commonSettings: Seq[Setting[?]] = Def.settings(
   Test / testOptions += Tests.Argument(TestFrameworks.ScalaCheck, "-w", "1"),
   Test / testOptions += Tests.Argument(TestFrameworks.ScalaCheck, "-verbosity", "2"),
   compile / javacOptions ++= Seq("-Xlint", "-Xlint:-serial"),
-  /*
```
eed3si9n, line 87: Because we run unidoc for sbt/sbt, sbt/zinc, etc we often get a complaint that the link doesn't work (https://github.com/sbt/sbt/issues/6487) so we commented out in https://github.com/sbt/sbt/pull/6502.

build.sbt
```diff
@@ -84,19 +84,21 @@ def commonSettings: Seq[Setting[?]] = Def.settings(
...
-    import scala.sys.process._
-    val devnull = ProcessLogger(_ => ())
-    val tagOrSha = ("git describe --exact-match" #|| "git rev-parse HEAD").lineStream(devnull).head
-    Seq(
-      "-sourcepath",
-      (LocalRootProject / baseDirectory).value.getAbsolutePath,
-      "-doc-source-url",
-      s"https://github.com/sbt/sbt/tree/$tagOrSha€{FILE_PATH}.scala"
-    )
+    if (Dependencies.sbtIoPath.isEmpty && Dependencies.sbtZincPath.isEmpty) {
```
xuwei-k, line 88: Add unidoc check.

Unidoc still does not work ,but Maven Central `javadoc.jar` will have correct links.

eed3si9n (approved):
Thanks!

PR #9067, opened 2026-04-11
[2.0.x] add no-op `addDependencyTreePlugin` shim for migration

mkurz:
Makes it easier for repos that have `addDependencyTreePlugin` in their `project/plugins.sbt`

What do you think?


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -4747,6 +4747,13 @@ trait BuildExtra extends BuildCommon with DefExtra {
     Seq(compose(onLoad, add), compose(onUnload, remove))
   }
 
+  /**
+   * Adds dependency tree plugin.
+   */
+  @deprecated("dependencyTree is built into sbt 2.x; this is now a no-op", "2.0.0")
+  def addDependencyTreePlugin: Setting[Seq[ModuleID]] =
+    libraryDependencies :== Nil
```
eed3si9n: ok, but this is gonna wipe out the plugins.

```suggestion
    libraryDependencies ++= Nil
```

Could you add a scripted test to check that this won't break things plz?

eed3si9n (changes requested):

main/src/main/scala/sbt/Defaults.scala
mkurz: Believe it or not, I corrected that  locally yesterday to  `++=` but did not commit it ....
will add the scripted test.

mkurz: Done. let me know what you think.

eed3si9n (approved):
Thanks!

PR #9076, opened 2026-04-12
[2.x] Update `.gitattributes` for contraband files

xuwei-k:
For syntax highlight on GitHub.

https://github.com/sbt/contraband/blob/ae50d940b23073586de9e4a071dc60fb9fab4da7/docs/01-schema.md?plain=1#L15-L16

> Since we don't want to rely on a specific programming language syntax, to talk about Contraband schemas, we'll extend GraphQL's schema language.

## Before

before

<hr />

## After

my fork repo

after


eed3si9n (approved):
Thanks!

PR #9077, opened 2026-04-12
[2.x] fix: Cache AutoPlugin projectSettings/projectConfigurations

### LINKED ISSUE #5166: Reevaluating plugin settings can slow down build loading
As part of loading the build, sbt will evaluate the projectSettings and globalSettings defined in each `AutoPlugin` for _every_ subproject. This can significantly slow down loading multiproject builds. For example, the akka project depends on the scalafix plugin and evaluating the scalafix projectSettings took between 100 and 200ms. There are nearly 40 subprojects in akka so the scalafix plugin was delaying akka startup by about 3-4 seconds.

I was able to speed up the akka build by updating the scalafix plugin to use a lazy val for projectSettings instead of a def: https://github.com/scalacenter/sbt-scalafix/pull/42. I wonder though if we should consider internally caching the value of projectSettings and globalSettings for each AutoPlugin so that they only get evaluated once. I can't really think of a good reason to have them as a def because there is no way for the caller to communicate for which project the settings are being evaluated so they should be the same for all projects. Changing this behavior though would, of course, have the possibility of breaking _something_ but I still think it might be worth it.

9876543210-tc-0123456789:
## Summary

- Cache `AutoPlugin.projectSettings` and `projectConfigurations` via `AutoPluginCache` so they are evaluated once per plugin instance during build loading, not once per subproject (N×M → M)
- Make `GroupedAutoPlugins.globalSettings` a `lazy val` and pre-compute `buildSettingsMap` to avoid repeated aggregation

## Motivation

Fixes #5166. In large multi-project builds, redundant plugin setting evaluation causes noticeable startup overhead. For example, scalafix's `projectSettings` took 100–200ms per project — in akka (~40 subprojects) this alone added 3–4 seconds to build load time.

## Test plan

- [ ] Verify existing tests pass (no behavioral change expected — this is a pure memoization)
- [ ] Profile build loading on a large multi-project build (e.g. akka) to confirm reduced startup time

### DIFF
--- a/main/src/main/scala/sbt/internal/GroupedAutoPlugins.scala
+++ b/main/src/main/scala/sbt/internal/GroupedAutoPlugins.scala
@@ -16,9 +16,11 @@ private[sbt] final class GroupedAutoPlugins(
     val all: Seq[AutoPlugin],
     val byBuild: Map[URI, Seq[AutoPlugin]]
 ) {
-  def globalSettings: Seq[Setting[?]] = all.flatMap(_.globalSettings)
+  lazy val globalSettings: Seq[Setting[?]] = all.flatMap(_.globalSettings)
+  private lazy val buildSettingsMap: Map[URI, Seq[Setting[?]]] =
+    byBuild.view.mapValues(_.flatMap(_.buildSettings)).toMap
   def buildSettings(uri: URI): Seq[Setting[?]] =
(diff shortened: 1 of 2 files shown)


9876543210-tc-0123456789:
@eed3si9n Can you review again my PR? I fixed CI tests. Thanks.

eed3si9n (approved):
Thanks!

PR #9079, opened 2026-04-12
[2.x] feat: Capture test output into JUnit XML system-out/system-err (fixes #6537)

### LINKED ISSUE #6537: SBT doesn't include system out or system error in JUnit XML reports
## steps

1. Run JUnit tests with `sbt test`.
2. Check the JUnit XML output in `target/test-reports`.

## problem

The `<system-out>` and `<system-err>` tags are always empty, even if the unit tests produced output. This is because [`JUnitXmlTestsListener.scala` hardcodes them to always be empty](https://github.com/sbt/sbt/blob/fabeed222799efcc46be15bccf2c31e384f7e369/testing/src/main/scala/sbt/JUnitXmlTestsListener.scala#L166). This makes it harder to track down the cause of unit test failures.

## expectation

Those XML tags should be populated with each unit test's sysout and syserr (or SBT should be configurable so that they can be populated).

## notes

https://github.com/dmitchellmim/sbt-junit-sysout is a simple project to reproduce the issue. It uses Scala 2.13.3 and SBT 1.2.8, but based on the Git history of `JUnitXmlTestsListener.scala`, I think all versions have the same issue.

luckydicer:
## Summary

- Add opt-in settings `testReportXmlCaptureStdOut` and `testReportXmlCaptureStdErr` (both default to `false`) to preserve backward compatibility
- When enabled, route sbt logger output (`info`/`debug` → `<system-out>`, `error`/`warn`/`trace` → `<system-err>`) per test suite via `ContentLogger`
- Maintain binary compatibility with overloaded constructors

## Motivation

Addresses #6537. JUnit XML reports always emit empty `<system-out>` and `<system-err>` elements, making it difficult to diagnose test failures in CI environments that parse these reports.

## Limitations

This captures output routed through sbt's logging infrastructure, **not** raw `System.out`/`System.err` streams. Direct `println` calls or library output written to stdout will not appear in the XML. Full stdout/stderr capture is significantly more complex due to process forking and parallel test execution (see discussion in #6537). This is an incremental improvement.

## Test plan

- [ ] New tests verify output appears in XML when capture is enabled
- [ ] New test verifies `<system-out>`/`<system-err>` remain empty when capture is disabled
- [ ] New test verifies selective capture (stdout only) works correctly
- [ ] Existing tests pass (defaults to `false`, no behavioral change)

### DIFF
--- a/main/src/main/scala/sbt/plugins/JUnitXmlReportPlugin.scala
+++ b/main/src/main/scala/sbt/plugins/JUnitXmlReportPlugin.scala
@@ -38,14 +38,26 @@ object JUnitXmlReportPlugin extends AutoPlugin {
   object autoImport {
     val testReportsDirectory =
       settingKey[File]("Directory for outputting junit test reports.").withRank(AMinusSetting)
+    val testReportXmlCaptureStdOut =
+      settingKey[Boolean](
+        "If true, capture test framework log output into <system-out> in JUnit XML reports."
+      ).withRank(BSetting)
+    val testReportXmlCaptureStdErr =
+      settingKey[Boolean](
+        "If true, capture test framework error output into <system-err> in JUnit XML reports."
+      ).withRank(BSetting)
 
     lazy val testReportSettings: Seq[Setting[?]] = Seq(
       testReportsDirectory := target.value / (prefix(configuration.value.name) + "reports"),
+      testReportXmlCaptureStdOut := false,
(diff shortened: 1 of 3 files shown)


eed3si9n:
3. Follow our AI assisted contributions guideline

taillred:
Hi @eed3si9n Could you please review this again? Thank you.

PR #9087, opened 2026-04-13
[2.x] feat: Forward test events to listeners as they are emitted

### LINKED ISSUE #7435: Forward test events to listener as they're emitted
## Current situation
sbt provides an API to register test event listeners which help better understand test execution. At the moment, it's impossible to rely on the timing of these events, because:
 - they're not passed on to the listeners at the time they're emitted
 - `startGroup` and `endGroup` events are received by listeners at different times depending on whether the tests run in-process or in a forked JVM.

## What I'd like to do
I would like to fix this by:
 - changing the in-process test runner so that events are immediately sent to the listeners as they're produced by the test frameworks.
 - slightly change the communication protocol between sbt and the forked JVM:
   - introduce new `StartGroup` and `EndGroup` messages, which sbt will map to the eponymous events of the test listeners.
   - send test events immediately as they're produced by the test frameworks.
   - Increase the priority of the thread reading messages from the forked JVM, so that they're handled as early as possible.

No API changes are envisioned for this work.

## Benefits
This change will make it possible to measure how much time is spent doing test class setup and cleanup. Currently, it's not possible to get this information, because the `startGroup`, `testEvent` and `endGroup` callbacks are called together and their payloads don't contain this information.

## Limitations
The test events have no timestamp, so it's impossible for test events listeners to know accurately when an event was emitted. The proposed changes improves the situation, but the results cannot be 100% accurate. It'd be good to look into this for sbt 2.

### When forking
 - All events are collected for a given test group: https://github.com/sbt/sbt/blob/fb8ca26e77d495cf206a45025a63c9a6d0754a9f/testing/agent/src/main/java/sbt/ForkMain.java#L407-L412
 - The events are sent back together to sbt: https://github.com/sbt/sbt/blob/fb8ca26e77d495cf206a45025a63c9a6d0754a9f/testing/agent/src/main/java/sbt/ForkMain.java#L438
 - sbt reads the events and emits the `startGroup`, `testEvent` and `endGroup` events all together: https://github.com/sbt/sbt/blob/fb8ca26e77d495cf206a45025a63c9a6d0754a9f/main-actions/src/main/scala/sbt/ForkTests.scala#L215-L221

### When running in-process
 - The start and end group events are produced as early as possible: https://github.com/sbt/sbt/blob/fb8ca26e77d495cf206a45025a63c9a6d0754a9f/testing/src/main/scala/sbt/TestFramework.scala#L166-L170
 - The test events are collected and sent together when a test group finishes executing: https://github.com/sbt/sbt/blob/fb8ca26e77d495cf206a45025a63c9a6d0754a9f/testing/src/main/scala/sbt/TestFramework.scala#L137C14-L137C14

claytonlin1110:
## Summary
Fixes #7435 

Test report listeners now receive testEvent callbacks as the underlying test framework emits events, instead of only after a whole group finishes. startGroup / endGroup timing is aligned with that streaming model for both in-process and forked test runs.

No public API changes — only listener callback timing and the internal worker ↔ sbt JSON-RPC protocol.

## What changed

- In-process tests (testing/src/main/scala/sbt/TestFramework.scala): each framework event is passed to listeners immediately from EventHandler.handle via testEvent(TestEvent(Seq(e))). SuiteResult is still computed from the full event list after execute completes. Synthetic error events from the error path are forwarded the same way.
- 
- Forked test worker (worker/src/main/java/sbt/internal/worker1/ForkTestMain.java): replaced the single batched testEvents message with streaming notifications: startTestGroup, testProgress (one ForkEventsInfo per event), and endTestGroup. Added ForkGroupStart and ForkGroupEnd DTOs for the new boundaries.
- 
- Fork test client (main-actions/src/main/scala/sbt/ForkTests.scala): React handles startTestGroup, testProgress, and endTestGroup; listeners get startGroup → per-event testEvent → endGroup in order, while events are accumulated per group for SuiteResult and the results map. The old testEvents JSON-RPC method is no longer used.
- 
- Worker TCP reader (main-actions/src/main/scala/sbt/internal/WorkerExchange.scala): the thread that reads fork responses over TCP is named sbt-fork-test-response-reader and its priority is set to Thread.NORM_PRIORITY + 1.

## Testing
Suggested local checks:

- ./sbt workerProj/compile testingProj/compile actionsProj/compile
- ./sbt testingProj/test main-actionsProj/test (or your project’s usual test set for these modules)
- Optional: run a project with Test / fork := true and a custom TestsListener / TestReportListener that logs startGroup / testEvent / endGroup order

### DIFF
--- a/launcher-package/integration-test/src/test/scala/RunnerScriptTest.scala
+++ b/launcher-package/integration-test/src/test/scala/RunnerScriptTest.scala
@@ -339,8 +339,7 @@ abstract class RunnerScriptTest extends verify.BasicTestSuite with ShellScriptUt
   // Test for issue #8755: Inline comments should be supported in .jvmopts
   testOutput(
     "sbt with inline comments in .jvmopts",
-    jvmoptsFileContents =
-      """--add-opens=java.base/java.util=ALL-UNNAMED # This is an inline comment
+    jvmoptsFileContents = """--add-opens=java.base/java.util=ALL-UNNAMED # This is an inline comment
         |-Dtest.key=value # Another comment
         |-Dtest.key2=file:/log4j2#prod.xml""".stripMargin,
     windowsSupport = false,
--- a/launcher-package/integration-test/src/test/scala/RunnerScriptTest.scala
+++ b/launcher-package/integration-test/src/test/scala/RunnerScriptTest.scala
@@ -339,7 +339,8 @@ abstract class RunnerScriptTest extends verify.BasicTestSuite with ShellScriptUt
   // Test for issue #8755: Inline comments should be supported in .jvmopts
   testOutput(
     "sbt with inline comments in .jvmopts",
-    jvmoptsFileContents = """--add-opens=java.base/java.util=ALL-UNNAMED # This is an inline comment
+    jvmoptsFileContents =
+      """--add-opens=java.base/java.util=ALL-UNNAMED # This is an inline comment
         |-Dtest.key=value # Another comment
         |-Dtest.key2=file:/log4j2#prod.xml""".stripMargin,
     windowsSupport = false,
--- a/main-actions/src/main/scala/sbt/ForkTests.scala
+++ b/main-actions/src/main/scala/sbt/ForkTests.scala
@@ -194,6 +194,7 @@ private class React(
 ) extends WorkerResponseListener:
   val g = WorkerMain.mkGson()
   val promise: Promise[Int] = Promise()
+
   /** Events per test group, accumulated for [[SuiteResult]] (listeners get each event immediately). */
   private val progressEvents = mutable.Map.empty[String, mutable.ArrayBuffer[testing.Event]]
   override def apply(line: String): Unit =
--- a/main-actions/src/main/scala/sbt/ForkTests.scala
+++ b/main-actions/src/main/scala/sbt/ForkTests.scala
@@ -194,6 +194,8 @@ private class React(
 ) extends WorkerResponseListener:
   val g = WorkerMain.mkGson()
   val promise: Promise[Int] = Promise()
+  /** Events per test group, accumulated for [[SuiteResult]] (listeners get each event immediately). */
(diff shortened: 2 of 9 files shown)


claytonlin1110:
Hi, @eed3si9n Would you please review this? I also signed CLA. Thanks.

eed3si9n:
3. Follow our AI assisted contributions guideline. All AI usage in any form must be disclosed in the PR description.

eed3si9n:
All AI usage in any form must be disclosed. If you've used GenAI, could you state how you used it in the PR description plz?

claytonlin1110:
Sorry but I haven't used GenAI.

eed3si9n:
Nice :)

eed3si9n:
Could you add a scripted test demonstrating the fix plz? 
https://github.com/sbt/sbt/blob/develop/contributing-docs/05_scripted_tests.md

claytonlin1110:
@eed3si9n Added. Please check. CI passed locally

launcher-package/integration-test/src/test/scala/RunnerScriptTest.scala
```diff
@@ -339,7 +339,8 @@ abstract class RunnerScriptTest extends verify.BasicTestSuite with ShellScriptUt
   // Test for issue #8755: Inline comments should be supported in .jvmopts
   testOutput(
     "sbt with inline comments in .jvmopts",
-    jvmoptsFileContents = """--add-opens=java.base/java.util=ALL-UNNAMED # This is an inline comment
+    jvmoptsFileContents =
+      """--add-opens=java.base/java.util=ALL-UNNAMED # This is an inline comment
```
xuwei-k: Don't change unrelated file.

claytonlin1110: Reverted

claytonlin1110:
@xuwei-k Please check.

claytonlin1110:
@eed3si9n Would you please review?

eed3si9n (approved):
Thanks!

PR #8892, opened 2026-03-10
[2.x] Update java file formatter plugin

xuwei-k:
`"com.swoval" % "sbt-java-format" % "0.3.1"` https://github.com/swoval/sbt-source-format is too old. It does not support new Java syntax. e.g. [`JEP 394: Pattern Matching for instanceof`](https://openjdk.org/jeps/394)

The latest version of `sbt-java-formatter` is `0.10.0` but I have added version `0.9.0`.

https://github.com/sbt/sbt-java-formatter/tags

Because latest version throw `IllegalAccessError`.

> google-java-format uses internal javac APIs for parsing Java source. The following JVM flags are required when running on JDK 16 and newer, due to [JEP 396: Strongly Encapsulate JDK Internals by Default](https://openjdk.java.net/jeps/396)

https://github.com/sbt/sbt-java-formatter/blob/f58c8d541cef3de4decba3c1606437ee994fd697/build.sbt#L19

https://github.com/google/google-java-format/blob/2f7e558e586788110dfda603652f68915d890cb0/README.md#as-a-library

```
[error] java.lang.IllegalAccessError: class com.google.googlejavaformat.java.ImportOrderer (in unnamed module @0x2c144835) cannot access class com.sun.tools.javac.parser.Tokens$TokenKind (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.parser to unnamed module @0x2c144835
[error] 	at com.google.googlejavaformat.java.ImportOrderer.<clinit>(ImportOrderer.java:111)
[error] 	at com.google.googlejavaformat.java.Formatter.formatSourceAndFixImports(Formatter.java:239)
[error] 	at com.github.sbt.javaformatter.JavaFormatter$.$anonfun$withFormattedSources$1(JavaFormatter.scala:149)
[error] 	at scala.collection.immutable.List.map(List.scala:293)
[error] 	at com.github.sbt.javaformatter.JavaFormatter$.withFormattedSources(JavaFormatter.scala:146)
[error] 	at com.github.sbt.javaformatter.JavaFormatter$.formatSources(JavaFormatter.scala:121)
[error] 	at com.github.sbt.javaformatter.JavaFormatter$.$anonfun$cachedFormatSources$1(JavaFormatter.scala:114)
```

PR #9089, opened 2026-04-14
[2.x] Update sbt-java-formatter

xuwei-k:
- https://github.com/sbt/sbt/pull/8892

> The latest version of sbt-java-formatter is 0.10.0 but I have added version 0.9.0.
>
> https://github.com/sbt/sbt-java-formatter/tags
>
> Because latest version throw IllegalAccessError.
>
>> google-java-format uses internal javac APIs for parsing Java source. The following JVM flags are required when running on JDK 16 and newer, due to [JEP 396: Strongly Encapsulate JDK Internals by Default](https://openjdk.java.net/jeps/396)

- https://github.com/sbt/sbt-java-formatter/pull/261


xuwei-k:
> `--server` is required

😢

mkurz:
> > `--server` is required
> 
> 😢

Why is this a problem? The sbt process starts, runs javafmt and quits again with the correct exit code.

mkurz:
Better approach: 
- https://github.com/sbt/sbt-java-formatter/pull/264

PR #9093, opened 2026-04-15
[2.x] refactor: Use Files.writeString instead of Files.write

xuwei-k:
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Files.html#writeString(java.nio.file.Path,java.lang.CharSequence,java.nio.file.OpenOption...)

> Write a [CharSequence](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/CharSequence.html) to a file. Characters are encoded into bytes using the [UTF-8](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/charset/StandardCharsets.html#UTF_8) [charset](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/charset/Charset.html).


eed3si9n (approved):
Thanks!

PR #9101, opened 2026-04-16
[2.x] fix: Expand virtualized plugin paths in forked console

### LINKED ISSUE #9096: [2.x] Compiler plugins aren't loaded in `console`
## steps

1. Create a build that uses a compiler plugin
    ```properties
    # project/build.properties
    sbt.version=2.0.0-RC12
    ```
    ```scala
    // build.sbt
      .settings(
        addCompilerPlugin(("com.lihaoyi" %% "acyclic" % "0.3.21").cross(CrossVersion.full)),
        libraryDependencies += ("com.lihaoyi" %% "acyclic" % "0.3.21").cross(CrossVersion.full) % "provided",
        scalacOptions += "-P:acyclic:force",
      )
    ```
2. Run `sbt console` in the project dir

## problem

The `console` loads correctly, but the compiler plugin is not loaded -- the logs show an error:

```
[info] running (fork) sbt.internal.worker1.WorkerMain @/Users/matt/sbt-2-compiler-plugin/target/bg-jobs/sbt_2ce7401a/job-1/console-params.json
No plugin in path ${CSR_CACHE}/https/repo1.maven.org/maven2/com/lihaoyi/acyclic_3.8.2/0.3.21/acyclic_3.8.2-0.3.21.jar
bad option: -P:acyclic:force
1 warning found
1 error found
```

## expectation

Compiler plugins should be loaded properly when using the `console`

claytonlin1110:
## Problem
Fixes #9096 

Compiler plugin options can contain virtualized paths (for example `${CSR_CACHE}/...`).
In forked `console`, these scalac options were forwarded to the REPL worker without
expanding virtual paths to real filesystem paths, which caused plugin jars not to load
and plugin options like `-P:...` to fail.

## Solution
- Resolve virtualized scalac options via the active `FileConverter` before serializing
  `ConsoleInfo` for forked `console`.
- Reuse the same option-resolution helper in doc option handling to avoid duplicate logic.
- Add a regression unit test that verifies `-Xplugin:${CSR_CACHE}/...` is expanded to a
  machine path before tool invocation.

### DIFF
--- a/main/src/main/scala/sbt/internal/Compiler.scala
+++ b/main/src/main/scala/sbt/internal/Compiler.scala
@@ -481,12 +481,14 @@ object Compiler:
         workingDir,
         conv,
       )
+      val consoleScalacOptions =
+        resolveVirtualizedScalacOptions((task / Keys.scalacOptions).value, conv)
       val param = ConsoleInfo(
         ArrayList(toolJars.asJava),
         ArrayList(bridgeJars.toVector.map(vf => conv.toPath(vf).toUri()).asJava),
         ArrayList(),
         ArrayList(Attributed.data(cp).toVector.map(vf => conv.toPath(vf).toUri()).asJava),
-        ArrayList((task / Keys.scalacOptions).value.asJava),
+        ArrayList(consoleScalacOptions.asJava),
         (task / Keys.initialCommands).value,
         (task / Keys.cleanupCommands).value,
       )
@@ -567,13 +569,7 @@ object Compiler:
               if (ScalaArtifacts.isScala3(sv)) Opts.doc.externalAPIScala3(xapisFiles)
               else Opts.doc.externalAPI(xapisFiles)
             val options = sOpts ++ externalApiOpts
-            def convertVfRef(value: String): String =
-              if !value.contains("$") then value
-              else converter.toPath(xsbti.VirtualFileRef.of(value)).toString
-            val resolvedOptions = options.map { x =>
-              if !x.contains("$") then x
-              else x.split(":").map(_.split(",").map(convertVfRef).mkString(",")).mkString(":")
-            }
+            val resolvedOptions = resolveVirtualizedScalacOptions(options, converter)
             val scalac = cs.scalac match
               case ac: AnalyzingCompiler => ac.onArgs(exported(s, "scaladoc"))
             val docSrcFiles = if ScalaArtifacts.isScala3(sv) then tFiles else srcs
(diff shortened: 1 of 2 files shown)


mrdziuban:
@claytonlin1110 this solution is not scoped tightly enough. It's not safe to assume that any value in `scalacOptions` that contains `$` is one that needs to be passed through `FileConverter#toPath`.

For example, in my build I use [a compiler plugin](https://github.com/mblink/disable-toString) that expects to be passed regular expressions via `scalacOptions`. One option I set is:

```scala
"-P:disableToString:regex=^java\\.time\\.(Local|Zoned)DateTime$"
```

This solution would see that the option contains `$` and would pass it through `FileConverter#toPath`, which changes the backslashes to forward slashes:

```scala
"-P:disableToString:regex=^java/.time/.(Local|Zoned)DateTime$"
```

The built-in `-Wconf` compiler option also accepts regular expressions, e.g. `-Wconf:msg=regex:s`, so it could result in the same type of issue.

@eed3si9n is there an existing function that can identify and replace specific virtual paths like `${CSR_CACHE}` without needing to assume the entire `String` is a path?

main/src/main/scala/sbt/internal/Compiler.scala
```diff
@@ -641,4 +637,23 @@ object Compiler:
...
+  private[sbt] def resolveVirtualizedScalacOptions(
+      options: Seq[String],
+      converter: FileConverter
+  ): Seq[String] =
+    def convertValue(value: String): String =
+      if !value.contains("$") then value
+      else converter.toPath(xsbti.VirtualFileRef.of(value)).toString
+    options.map: option =>
+      if !option.contains("$") then option
+      else option.split(":").map(_.split(",").map(convertValue).mkString(",")).mkString(":")
```
mrdziuban: I guess the issue I described already applies to `scalacOptions` used in `docTask`. This would work and would fix the issue for both the console and docs:

```suggestion
    options.map: option =>
      if !option.startsWith("-Xplugin:") then option
      else "-Xplugin:" + converter.toPath(xsbti.VirtualFileRef.of(option.stripPrefix("-Xplugin:")))
```

eed3si9n:
@mrdziuban Here's a related PR I sent to Zinc - https://github.com/sbt/zinc/pull/1545, which also just checks `!x.contains("$")` mostly, but given your example maybe it should be more careful.

mrdziuban:
> @mrdziuban Here's a related PR I sent to Zinc - [sbt/zinc#1545](https://github.com/sbt/zinc/pull/1545), which also just checks `!x.contains("$")` mostly, but given your example maybe it should be more careful.

Indeed, I just opened sbt/zinc#1680 to scope it more tightly

mrdziuban:
I also opened #9110 as an alternative to this PR to more closely align with what sbt/zinc#1680 is doing

Issue #8410, opened 2025-12-10
Report eviction errors for `Test` dependencies


zainab-ali:
Eviction warnings and errors are currently not reported for test dependencies.

## Example

If I have the following two (non-test) dependencies:
```scala
scalaVersion := "3.3.7"

libraryDependencies ++= Seq(
  "org.typelevel" %% "weaver-cats"       % "0.11.1",
  "com.siriusxm" %% "snapshot4s-weaver"       % "0.2.2",
)
```

And run `update`, I get the expected error:
```
[error] (update) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[error]
[error]         * org.typelevel:weaver-cats_3:0.11.1 (early-semver) is selected over 0.10.1
[error]             +- default:sbt-test-project_3:0.1.0-SNAPSHOT          (depends on 0.11.1)
[error]             +- com.siriusxm:snapshot4s-weaver_3:0.2.2             (depends on 0.10.1)
```

But if I change them to `Test` dependencies, `update` is successful:
```
libraryDependencies ++= Seq(
  "org.typelevel" %% "weaver-cats"       % "0.11.1" % Test,
  "com.siriusxm" %% "snapshot4s-weaver"       % "0.2.2" % Test,
)
```

## Feature

`update` could fail when `Test` dependencies have evictions.

The eviction check is currently only set for the `Compile` configuration. We can run the same check for `Test` if we tweak the error message to include the configuration name.

I'd be happy to contribute this feature, if given a few pointers on where to make the change.

eed3si9n:
> I'd be happy to contribute this feature, if given a few pointers on where to make the change.

Most of the eviction error happens in `LibraryManagement.scala` 

https://github.com/sbt/sbt/blob/4ef942e7e0be67a100e08bd4bab2db4422384607/main/src/main/scala/sbt/internal/LibraryManagement.scala#L71-L78

and `EvictionError.scala`, and  https://github.com/sbt/sbt/blob/4ef942e7e0be67a100e08bd4bab2db4422384607/lm-core/src/main/scala/sbt/librarymanagement/EvictionError.scala#L26

specifically is probably what you would want to change. `full` specifies only `Vector(Compile)` since it's full used for the full eviction warning, as opposed to a summary version:

https://github.com/sbt/sbt/blob/4ef942e7e0be67a100e08bd4bab2db4422384607/lm-core/src/main/scala/sbt/librarymanagement/EvictionWarning.scala#L75-L77

[...]

PR #9102, opened 2026-04-16
[2.x] Add `Test` configuration to evictionWarningOptions

zainab-ali:
This is an alternative fix for #8410

### DIFF
--- a/lm-core/src/main/scala/sbt/librarymanagement/EvictionError.scala
+++ b/lm-core/src/main/scala/sbt/librarymanagement/EvictionError.scala
@@ -43,7 +43,9 @@ object EvictionError {
       assumedEvictionErrorLevel: Level.Value,
       configurations: Seq[ConfigRef],
   ): EvictionError = {
-    val evictions = EvictionWarning.buildEvictions(configurations, report)
+    val evictions = configurations.flatMap { configuration =>
+      EvictionWarning.buildEvictions(Vector(configuration), report)
+    }
     processEvictions(
       module,
       evictions,
@@ -67,23 +69,17 @@ object EvictionError {
       .map { detail =>
         val evicteds = detail.modules filter { _.evicted }
         val winner = (detail.modules filterNot { _.evicted }).headOption
-
-        (detail.organization, detail.name, winner, evicteds)
-      }
-      .distinct
-      .map { case (org, name, winner, evicteds) =>
         new EvictionPair(
-          org,
-          name,
+          detail.organization,
+          detail.name,
           winner,
           evicteds,
           includesDirect = true,
           showCallers = true
         )
       }
-    val incompatibleEvictions: mutable.ListBuffer[(EvictionPair, String)] = mutable.ListBuffer()
-    val assumedIncompatibleEvictions: mutable.ListBuffer[(EvictionPair, String)] =
-      mutable.ListBuffer()
+    val incompatibleEvictions: mutable.ListBuffer[EvictionErrorPair] = mutable.ListBuffer()
+    val assumedIncompatibleEvictions: mutable.ListBuffer[EvictionErrorPair] = mutable.ListBuffer()
     val sbvOpt = module.scalaModuleInfo.map(_.scalaBinaryVersion)
     val userDefinedSchemes: Map[(String, String), String] = Map(schemes flatMap { s =>
       val organization = s.organization
@@ -134,20 +130,43 @@ object EvictionError {
             .orElse(fromLibraryDependencySchemes()) // global
         }
 
+        def revisionAndCallers(module: ModuleReport): RevisionAndCallers =
+          RevisionAndCallers(module.module.revision, module.callers.map(_.caller))
+
         // We want the user to be able to suppress eviction errors for a specific library,
         // which would result in an incompatible eviction based on the assumed version scheme.
         // So, only fall back to the assumed scheme if there is no given scheme by the user or the pom.
         userDefinedSchemeOrFromPom match {
           case Some(givenScheme) =>
-            if (hasIncompatibleVersionForScheme(givenScheme))
-              incompatibleEvictions += (p -> givenScheme)
+            if (hasIncompatibleVersionForScheme(givenScheme)) {
+              val eviction = EvictionErrorPair(
+                p.name,
+                p.organization,
+                p.winner.map(revisionAndCallers),
+                p.evicteds.map(revisionAndCallers),
+                givenScheme
+              )
+              if (!incompatibleEvictions.contains(eviction)) {
+                incompatibleEvictions += eviction
+              }
+            }
           case None =>
             val assumedScheme =
               if (isNameScalaSuffixed(p.name)) assumedVersionScheme
               else assumedVersionSchemeJava
 
-            if (hasIncompatibleVersionForScheme(assumedScheme))
-              assumedIncompatibleEvictions += (p -> assumedScheme)
+            if (hasIncompatibleVersionForScheme(assumedScheme)) {
+              val eviction = EvictionErrorPair(
+                p.name,
+                p.organization,
+                p.winner.map(revisionAndCallers),
+                p.evicteds.map(revisionAndCallers),
+                assumedScheme
+              )
+              if (!assumedIncompatibleEvictions.contains(eviction)) {
+                assumedIncompatibleEvictions += eviction
+              }
+            }
         }
 
       case _ => ()
@@ -164,9 +183,27 @@ object EvictionError {
   }
 }
 
+private final case class EvictionErrorPair(
+    name: String,
+    organization: String,
+    winner: Option[RevisionAndCallers],
+    evicteds: Vector[RevisionAndCallers],
+    scheme: String
+) {
+  val evictedRevs: String = {
+    val revs = evicteds map { _.moduleRevision }
+    if (revs.size <= 1) revs.mkString else revs.distinct.mkString("{", ", ", "}")
+  }
+}
+
+private final case class RevisionAndCallers(
+    moduleRevision: String,
+    callers: Vector[ModuleID]
+)
+
 final class EvictionError private[sbt] (
-    val incompatibleEvictions: Seq[(EvictionPair, String)],
-    val assumedIncompatibleEvictions: Seq[(EvictionPair, String)],
+    val incompatibleEvictions: Seq[EvictionErrorPair],
+    val assumedIncompatibleEvictions: Seq[EvictionErrorPair],
 ) {
   def run(): Unit =
     if (incompatibleEvictions.nonEmpty) {
@@ -177,25 +214,26 @@ final class EvictionError private[sbt] (
 
   def toAssumedLines: List[String] = toLines(assumedIncompatibleEvictions, true)
 
-  def toLines(evictions: Seq[(EvictionPair, String)], assumed: Boolean): List[String] = {
+  def toLines(evictions: Seq[EvictionErrorPair], assumed: Boolean): List[String] = {
     val out: mutable.ListBuffer[String] = mutable.ListBuffer()
     out += "found version conflict(s) in library dependencies; some are suspected to be binary incompatible:"
     out += ""
-    evictions.foreach({ (a, scheme) =>
+    evictions.foreach({ a =>
       val seen: mutable.Set[ModuleID] = mutable.Set()
       val callers: List[String] = (a.evicteds.toList ::: a.winner.toList) flatMap { r =>
-        val rev = r.module.revision
+
+        val rev = r.moduleRevision
         r.callers.toList flatMap { caller =>
-          if (seen(caller.caller)) Nil
+          if (seen(caller)) Nil
           else {
-            seen += caller.caller
+            seen += caller
             List(f"\t    +- ${caller}%-50s (depends on $rev)")
           }
         }
       }
       val que = if (assumed) "?" else ""
       val winnerRev = a.winner match {
-        case Some(r) => s":${r.module.revision} ($scheme$que) is selected over ${a.evictedRevs}"
+        case Some(r) => s":${r.moduleRevision} (${a.scheme}$que) is selected over ${a.evictedRevs}"
         case _       => " is evicted for all versions"
       }
       val title = s"\t* ${a.organization}:${a.name}$winnerRev"
--- a/lm-core/src/main/scala/sbt/librarymanagement/EvictionError.scala
+++ b/lm-core/src/main/scala/sbt/librarymanagement/EvictionError.scala
@@ -30,7 +30,7 @@ object EvictionError {
       assumedVersionScheme,
       assumedVersionSchemeJava,
       assumedEvictionErrorLevel,
-      Configurations.Compile,
+      EvictionWarningOptions.full,
     )
   }
 
@@ -41,9 +41,8 @@ object EvictionError {
       assumedVersionScheme: String,
       assumedVersionSchemeJava: String,
       assumedEvictionErrorLevel: Level.Value,
-      configuration: ConfigRef,
+      options: EvictionWarningOptions,
   ): EvictionError = {
-    val options = EvictionWarningOptions.full.withConfigurations(Vector(configuration))
     val evictions = EvictionWarning.buildEvictions(options, report)
     processEvictions(
       module,
@@ -53,7 +52,6 @@ object EvictionError {
       assumedVersionScheme,
       assumedVersionSchemeJava,
       assumedEvictionErrorLevel,
-      configuration,
     )
   }
 
@@ -65,7 +63,6 @@ object EvictionError {
       assumedVersionScheme: String,
       assumedVersionSchemeJava: String,
       assumedEvictionErrorLevel: Level.Value,
-      configuration: ConfigRef = Configurations.Compile,
   ): EvictionError = {
     val directDependencies = module.directDependencies
     val pairs = reports map { detail =>
@@ -155,7 +152,6 @@ object EvictionError {
     new EvictionError(
       incompatibleEvictions.toList,
       assumedIncompatibleEvictions.toList,
-      configuration,
     )
   }
 
@@ -167,7 +163,6 @@ object EvictionError {
 final class EvictionError private[sbt] (
     val incompatibleEvictions: Seq[(EvictionPair, String)],
     val assumedIncompatibleEvictions: Seq[(EvictionPair, String)],
-    val configuration: ConfigRef = Configurations.Compile,
 ) {
   def run(): Unit =
     if (incompatibleEvictions.nonEmpty) {
@@ -178,13 +173,9 @@ final class EvictionError private[sbt] (
 
   def toAssumedLines: List[String] = toLines(assumedIncompatibleEvictions, true)
 
-  private def configurationLabel: String =
-    if (configuration.name == Configurations.Compile.name) "library dependencies"
-    else s"${configuration.name.capitalize} dependencies"
-
   def toLines(evictions: Seq[(EvictionPair, String)], assumed: Boolean): List[String] = {
     val out: mutable.ListBuffer[String] = mutable.ListBuffer()
-    out += s"found version conflict(s) in $configurationLabel; some are suspected to be binary incompatible:"
+    out += "found version conflict(s) in library dependencies; some are suspected to be binary incompatible:"
     out += ""
     evictions.foreach({ (a, scheme) =>
       val seen: mutable.Set[ModuleID] = mutable.Set()
--- a/lm-core/src/main/scala/sbt/librarymanagement/EvictionError.scala
+++ b/lm-core/src/main/scala/sbt/librarymanagement/EvictionError.scala
@@ -63,18 +63,24 @@ object EvictionError {
       assumedEvictionErrorLevel: Level.Value,
   ): EvictionError = {
     val directDependencies = module.directDependencies
-    val pairs = reports map { detail =>
-      val evicteds = detail.modules filter { _.evicted }
-      val winner = (detail.modules filterNot { _.evicted }).headOption
-      new EvictionPair(
-        detail.organization,
-        detail.name,
-        winner,
-        evicteds,
-        includesDirect = true,
-        showCallers = true
-      )
-    }
+    val pairs = reports
+      .map { detail =>
+        val evicteds = detail.modules filter { _.evicted }
+        val winner = (detail.modules filterNot { _.evicted }).headOption
+
+        (detail.organization, detail.name, winner, evicteds)
+      }
+      .distinct
+      .map { case (org, name, winner, evicteds) =>
+        new EvictionPair(
+          org,
+          name,
+          winner,
+          evicteds,
+          includesDirect = true,
+          showCallers = true
+        )
+      }
     val incompatibleEvictions: mutable.ListBuffer[(EvictionPair, String)] = mutable.ListBuffer()
     val assumedIncompatibleEvictions: mutable.ListBuffer[(EvictionPair, String)] =
       mutable.ListBuffer()
--- a/lm-core/src/main/scala/sbt/librarymanagement/EvictionError.scala
+++ b/lm-core/src/main/scala/sbt/librarymanagement/EvictionError.scala
@@ -30,7 +30,7 @@ object EvictionError {
       assumedVersionScheme,
       assumedVersionSchemeJava,
       assumedEvictionErrorLevel,
-      EvictionWarningOptions.full,
+      EvictionWarningOptions.default.configurations,
     )
   }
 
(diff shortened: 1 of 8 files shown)


lm-core/src/main/scala/sbt/librarymanagement/EvictionWarning.scala
```diff
@@ -74,7 +75,7 @@ object EvictionWarningOptions {
   def default: EvictionWarningOptions = summary
```
zainab-ali, line 75: Should this be changed to `def default: EvictionWarningOptions = full`?

This PR specifies `EvictionWarningOptions.full` as the value in `Defaults.scala`.

`EvictionWarningOptions.default` wasn't the default value previously. While it was specified in `Defaults.scala`, it wasn't actually used.

sbt-app/src/sbt-test/dependency-management/evicted-test-config/build.sbt
```diff
@@ -1,6 +1,6 @@
 // https://github.com/sbt/sbt/issues/8410
-scalaVersion := "3.3.4"
+scalaVersion := "3.3.7"
 libraryDependencies ++= Seq(
-  "org.typelevel" %% "weaver-cats" % "0.8.4" % Test,
-  "com.siriusxm" %% "snapshot4s-weaver" % "0.1.5" % Test,
+  "org.typelevel" %% "weaver-cats" % "0.11.1" % Test,
+  "com.siriusxm" %% "snapshot4s-weaver" % "0.2.2" % Test,
 )
```
zainab-ali, line 6: The dependency `"org.typelevel" %% "weaver-cats" % "0.8.4"` doesn't exist. Weaver was released under typelevel from `0.9` onwards. This meant that `update` failed, but for the wrong reasons. 

I've updated the test to reflect the original issue.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -3973,6 +3937,7 @@ object Classpaths {
       transitiveUpdate.toTaskable,
       (update / unresolvedWarningConfiguration).toTaskable,
       evictionErrorLevel.toTaskable,
+      (update / evictionWarningOptions).toTaskable,
```
zainab-ali, line 3942: Is this correct, or should it be`evictionWarningOptions.toTaskable`?

lm-core/src/main/scala/sbt/librarymanagement/EvictionWarning.scala
eed3si9n: <s>Yea, I think that's a good idea.</s> Looking at it more carefully, I'm no longer sure.

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -3020,7 +2985,7 @@ object Classpaths {
     Defaults.globalDefaults(
       Seq(
         conflictWarning :== ConflictWarning.default("global"),
-        evictionWarningOptions := EvictionWarningOptions.default,
+        evictionWarningOptions := EvictionWarningOptions.full,
```
eed3si9n: Prior to [sbt 1.5.0](https://eed3si9n.com/sbt-1.5.0), there were eviction warnings, which warned about the users about the library eviction based mostly on conventions and guesses. The eviction warning had two modes:
1. One that printed alongside the `update` task, which had mixed reviews due to many false positives
2. One that printed explicitly when the user called for the `evicted` task

These two modes are configured using `EvictionWarningOptions`, where the `update` version is called `summary` and the full one is called `full`. These are expressed by:

```scala
    update / evictionWarningOptions := evictionWarningOptions.value,
    evicted / evictionWarningOptions := EvictionWarningOptions.full,
```

During 2020 to 2021 time there were lots of collaboration between independent programmers and Scala Center, and we were thinking bigger (maybe due to the world-ending pandemic), and decided to fix this problem without the guess work by changing Maven Central information instead: [Preventing version conflicts with versionScheme](https://www.scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html).

So in https://github.com/sbt/sbt/pull/6221, I dropped the eviction warning within `update` and replaced it with eviction error instead. But because we were still in sbt 1.x series, the settings remained as vestige. The globally scoped `evictionWarningOptions` is to influence the behavior of `update`-version of the eviction error for all subprojects.

If we mostly care about the list of configurations, maybe what we should do is leave this line alone, and reuse the `summary` option to carry the list of `Compile` and `Test`.

zainab-ali: Thanks for the explanation. I've reverted the `evictionWarningOptions` setting back to `default`,  added `Test` to `summary`.

The configurations are the only value shared between `update` and `evicted` - all other values only apply to `evicted`. I've tried to reflect this in the code by removing `EvictionWarningOptions` usages from `EvictionError`.

Do you think this approach still makes sense? Alternatively, we could remove the `update / evictionWarningOptions` setting, and use a non-configurable list of `Compile` and `Test` configurations across both `update` and `evicted`.

lm-core/src/main/scala/sbt/librarymanagement/EvictionError.scala
```diff
@@ -204,7 +193,7 @@ final class EvictionError private[sbt] (
         case _       => " is evicted for all versions"
       }
       val title = s"\t* ${a.organization}:${a.name}$winnerRev"
-      val lines = title :: (if (a.showCallers) callers.reverse else Nil) ::: List("")
+      val lines = title :: callers.reverse ::: List("")
```
zainab-ali, line 231: `showCallers` was hard-coded to `true` in `EvictionWarningOptions.full`, so we'd always show these lines.

main/src/main/scala/sbt/internal/LibraryManagement.scala
```diff
@@ -74,29 +75,20 @@ private[sbt] object LibraryManagement {
...
-      )
-      val evictionErrorTest = EvictionError(
-        report1,
-        module,
-        versionSchemeOverrides,
-        assumedVersionScheme,
-        assumedVersionSchemeJava,
-        assumedEvictionErrorLevel,
-        Configurations.Test,
+        evictionWarningOptions.configurations,
```
eed3si9n, line 85: Does combining these two make any difference in the way eviction errors are presented to the user? Could you show before/after in the PR description plz?

zainab-ali: There's a slight difference in the error message. The previous error message specifies the configuration as `Test`:

```diff
- [info] [error] java.lang.RuntimeException: found version conflict(s) in Test dependencies; some are suspected to be binary incompatible:
+ [info] [error] java.lang.RuntimeException: found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[info] [error]
[info] [error] 	* org.typelevel:weaver-cats_3:0.11.1 (early-semver) is selected over 0.10.1
[info] [error] 	    +- default:sbt_d3fe513a_3:0.1.0-SNAPSHOT              (depends on 0.11.1)
[info] [error] 	    +- com.siriusxm:snapshot4s-weaver_3:0.2.2             (depends on 0.10.1)
[info] [error]
[info] [error]
[info] [error] this can be overridden using libraryDependencySchemes or evictionErrorLevel
```

zainab-ali: The error message in the current `develop` branch is a bit misleading.

If I scope the dependecies to `Compile`, I still see the error message for `Test`. For example running `update` on:
```scala
scalaVersion := "3.3.7"
libraryDependencies ++= Seq(
  "org.typelevel" %% "weaver-cats" % "0.11.1" ,
  "com.siriusxm" %% "snapshot4s-weaver" % "0.2.2",
)
```
gives an error for `Test` too:
```
[info] [error] (update) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[info] [error]
[info] [error] 	* org.typelevel:weaver-cats_3:0.11.1 (early-semver) is selected over 0.10.1
[info] [error] 	    +- default:sbt_94f7e971_3:0.1.0-SNAPSHOT              (depends on 0.11.1)
[info] [error] 	    +- com.siriusxm:snapshot4s-weaver_3:0.2.2             (depends on 0.10.1)
[info] [error]
[info] [error] found version conflict(s) in Test dependencies; some are suspected to be binary incompatible:
[info] [error]
[info] [error] 	* org.typelevel:weaver-cats_3:0.11.1 (early-semver) is selected over 0.10.1
[info] [error] 	    +- default:sbt_94f7e971_3:0.1.0-SNAPSHOT              (depends on 0.11.1)
[info] [error] 	    +- com.siriusxm:snapshot4s-weaver_3:0.2.2             (depends on 0.10.1)
```

This is because the `Test` configuration's `OrganizationArtifactReport` contains the evictions for `Compile`.

The PRd implementation deduplicates the `List[OrganizationArtifactReport]`, but at the expense of omitting the configuration from the error message.

eed3si9n: I think Coursier (since https://github.com/coursier/sbt-coursier/pull/235) treats `Compile` and `Test` as separate dependency graphs, so even when the same library like `weaver-cats` is evicted, the winning version or the reason why it got evicted could be different. Otherwise, you have a situation where adding a test library (e.g. `embedded-kafka`) ends up influencing the main code - https://github.com/coursier/coursier/issues/1382. So to account for that, it might be prudent to report evictions separately. Maybe you could test how it would look using the build reported on the issue:

```scala
name := "sbt-dependency-resolution"
version := "0.1"
scalaVersion := "2.11.12"
val kafkaClients = "org.apache.kafka" % "kafka-clients" % "1.0.0"
val embeddedKafka = "io.github.embeddedkafka" %% "embedded-kafka" % "2.3.0"

libraryDependencies ++= Seq(
  kafkaClients,
  embeddedKafka % Test
)
```

zainab-ali: Thanks for the explanation. I tested on the build in the issue. `update` succeeds and `kafka-clients-1.0.0` is on the compile classpath for both implementations.

Note that `update` fails if we try a similar example for `weaver-cats`, in which the version scheme is set:

```scala
"org.typelevel" %% "weaver-cats" % "0.11.1",
"com.siriusxm" %% "snapshot4s-weaver" % "0.2.2" % Test,
```

zainab-ali: Here's a set of dependencies for which there are different eviction reasons for `Compile` and `Test`:

```scala
scalaVersion := "3.3.7"
libraryDependencies ++= Seq(
  "org.typelevel" %% "weaver-discipline" % "0.11.1", // pulls in weaver-cats 0.11.1
  "org.typelevel" %% "weaver-cats" % "0.10.0",
  "com.siriusxm" %% "snapshot4s-weaver" % "0.2.2" % Test, // pulls in weaver-cats 0.10.0
)
```

Here's the error in the current PR. The eviction for `Test` isn't displayed, since it's deduplicated based on module ID:
```
[info] [error] java.lang.RuntimeException: found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[info] [error]
[info] [error] 	* org.typelevel:weaver-cats_3:0.11.1 (early-semver) is selected over 0.10.0
[info] [error] 	    +- default:sbt_9934c9c0_3:0.1.0-SNAPSHOT              (depends on 0.11.1)
[info] [error] 	    +- org.typelevel:weaver-discipline_3:0.11.1           (depends on 0.11.1)
[info] [error] 	    +- org.typelevel:weaver-cats_3:0.10.0                 (depends on 0.10.0)
```

And here's the error on `develop`. Note that the eviction for `Test` is slightly different:
```diff
[info] [error] (update) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[info] [error]
[info] [error] 	* org.typelevel:weaver-cats_3:0.11.1 (early-semver) is selected over 0.10.0
[info] [error] 	    +- default:sbt_eaa6c11_3:0.1.0-SNAPSHOT               (depends on 0.11.1)
[info] [error] 	    +- org.typelevel:weaver-discipline_3:0.11.1           (depends on 0.11.1)
[info] [error] 	    +- org.typelevel:weaver-cats_3:0.10.0                 (depends on 0.10.0)
[info] [error]
[info] [error] found version conflict(s) in Test dependencies; some are suspected to be binary incompatible:
[info] [error]
[info] [error] 	* org.typelevel:weaver-cats_3:0.11.1 (early-semver) is selected over {0.10.0, 0.10.1}
[info] [error] 	    +- default:sbt_eaa6c11_3:0.1.0-SNAPSHOT               (depends on 0.11.1)
[info] [error] 	    +- org.typelevel:weaver-discipline_3:0.11.1           (depends on 0.11.1)
+[info] [error] 	    +- com.siriusxm:snapshot4s-weaver_3:0.2.2             (depends on 0.10.1)
[info] [error] 	    +- org.typelevel:weaver-cats_3:0.10.0                 (depends on 0.10.0)
```

I'm not sure how important this is to include. Once a user fixes `Compile` evictions, they'll then be presented with `Test` evictions.

eed3si9n: Here's an example build where we can get it to evicted by different versions.

```scala
scalaVersion := "3.3.7"
libraryDependencies ++= Seq(
  "org.typelevel" %% "weaver-discipline" % "0.11.1" % Test, // pulls in weaver-cats 0.11.1
  "org.typelevel" %% "weaver-cats" % "0.9.3",
  "com.siriusxm" %% "snapshot4s-weaver" % "0.2.2", // pulls in weaver-cats 0.10.0
)
```

Using 2.0.0-RC12, this produces:

```
[error] stack trace is suppressed; run last evicted for the full output
[error] (evicted) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[error]
[error] 	* org.typelevel:weaver-cats_3:0.10.1 (early-semver) is selected over 0.9.3
[error] 	    +- default:eviction_3:0.1.0-SNAPSHOT                  (depends on 0.10.1)
[error] 	    +- com.siriusxm:snapshot4s-weaver_3:0.2.2             (depends on 0.10.1)
[error] 	    +- org.typelevel:weaver-cats_3:0.9.3                  (depends on 0.9.3)
[error]
[error] found version conflict(s) in Test dependencies; some are suspected to be binary incompatible:
[error]
[error] 	* org.typelevel:weaver-cats_3:0.11.1 (early-semver) is selected over {0.9.3, 0.10.1}
[error] 	    +- default:eviction_3:0.1.0-SNAPSHOT                  (depends on 0.11.1)
[error] 	    +- org.typelevel:weaver-discipline_3:0.11.1           (depends on 0.11.1)
[error] 	    +- com.siriusxm:snapshot4s-weaver_3:0.2.2             (depends on 0.10.1)
[error] 	    +- org.typelevel:weaver-cats_3:0.9.3                  (depends on 0.9.3)
[error]
[error]
```

zainab-ali: Thanks! I've amended the code to run the check for each configuration and updated `processEvictions` to deduplicate the printed eviction errors. The error message with the dependencies above is:

```
[error] (update) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[error]
[error]  * org.typelevel:weaver-cats_3:0.10.1 (early-semver) is selected over 0.9.3
[error]      +- default:sbt_f879bcf0_3:0.1.0-SNAPSHOT              (depends on 0.10.1)
[error]      +- com.siriusxm:snapshot4s-weaver_3:0.2.2             (depends on 0.10.1)
[error]      +- org.typelevel:weaver-cats_3:0.9.3                  (depends on 0.9.3)
[error]
[error]  * org.typelevel:weaver-cats_3:0.11.1 (early-semver) is selected over {0.9.3, 0.10.1}
[error]      +- default:sbt_f879bcf0_3:0.1.0-SNAPSHOT              (depends on 0.11.1)
[error]      +- org.typelevel:weaver-discipline_3:0.11.1           (depends on 0.11.1)
[error]      +- com.siriusxm:snapshot4s-weaver_3:0.2.2             (depends on 0.10.1)
[error]      +- org.typelevel:weaver-cats_3:0.9.3                  (depends on 0.9.3)
[error]
[error]
```

I could update the error message with the set of configurations, e.g:

```diff
- [error]  * org.typelevel:weaver-cats_3:0.10.1 (early-semver) is selected over 0.9.3
+ [error]  * org.typelevel:weaver-cats_3:0.10.1 (early-semver) is selected over 0.9.3 in Compile
[error]      +- default:sbt_f879bcf0_3:0.1.0-SNAPSHOT              (depends on 0.10.1)
[error]      +- com.siriusxm:snapshot4s-weaver_3:0.2.2             (depends on 0.10.1)
[error]      +- org.typelevel:weaver-cats_3:0.9.3                  (depends on 0.9.3)
[error]
- [error]  * org.typelevel:weaver-cats_3:0.11.1 (early-semver) is selected over {0.9.3, 0.10.1}
+ [error]  * org.typelevel:weaver-cats_3:0.11.1 (early-semver) is selected over {0.9.3, 0.10.1} in Test
[error]      +- default:sbt_f879bcf0_3:0.1.0-SNAPSHOT              (depends on 0.11.1)
[error]      +- org.typelevel:weaver-discipline_3:0.11.1           (depends on 0.11.1)
[error]      +- com.siriusxm:snapshot4s-weaver_3:0.2.2             (depends on 0.10.1)
[error]      +- org.typelevel:weaver-cats_3:0.9.3                  (depends on 0.9.3)
```

However this would be a bit more complex, and perhaps would be better in a follow-up.

I'm looking into the semantic versioning test failure in `EvictionErrorSpec`. It seems to be related to the list of callers.

lm-core/src/main/scala/sbt/librarymanagement/EvictionError.scala
```diff
@@ -178,33 +214,30 @@ final class EvictionError private[sbt] (
...
+    out += "found version conflict(s) in library dependencies; some are suspected to be binary incompatible:"
     out += ""
-    evictions.foreach({ (a, scheme) =>
+    evictions.foreach({ a =>
       val seen: mutable.Set[ModuleID] = mutable.Set()
       val callers: List[String] = (a.evicteds.toList ::: a.winner.toList) flatMap { r =>
-        val rev = r.module.revision
+
+        val rev = r.moduleRevision
         r.callers.toList flatMap { caller =>
```
zainab-ali: I think I might have found a bug in the caller-related logic in `2.0.0-RC2`.

This code in `2.0.0-RC2` (as well as this code), assumes that each `caller` depends on a single revision.

However, the callers in [the semantic versioning test](https://github.com/sbt/sbt/blob/develop/lm-ivy/src/test/scala/sbt/internal/librarymanagement/EvictionErrorSpec.scala#L73) have multiple associated revisions.  Would you expect this?

An arbitrary revision is picked and displayed to the user. 

As an example, `org.http4s:http4s-core_2.13:0.21.11` depends on `org.typelevel:cats-effect_2.13` revisions `2.2.0` and `3.0.0-M4`. The error message only contains `2.2.0`:
```
org.http4s:http4s-core_2.13:0.21.11                (depends on 2.2.0)
```



The test fails when the `Test` configuration is added to the `EvictionErrorSpec`, since the callers for `Test` and `Compile` are different. I could update the test, but wanted to clear up my understanding first. The updated error message is a bit confusing:

```
[info]    found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[info]
[info]   	* org.typelevel:cats-effect_2.13:3.0.0-M4 (early-semver) is selected over {2.0.0, 2.2.0}
[info]   	    +- com.example:foo:0.1.0                              (depends on 3.0.0-M4)
[info]   	    +- co.fs2:fs2-core_2.13:2.4.5                         (depends on 2.2.0)
[info]   	    +- org.http4s:http4s-core_2.13:0.21.11                (depends on 2.2.0)
[info]   	    +- io.chrisdavenport:vault_2.13:2.0.0                 (depends on 2.0.0)
[info]   	    +- io.chrisdavenport:unique_2.13:2.0.0                (depends on 2.0.0)
[info]
[info]   	* org.typelevel:cats-effect_2.13:3.0.0-M4 (early-semver) is selected over {2.0.0, 2.2.0}
[info]   	    +- com.example:foo:0.1.0                              (depends on 2.2.0)
[info]   	    +- co.fs2:fs2-core_2.13:2.4.5                         (depends on 2.2.0)
[info]   	    +- org.http4s:http4s-core_2.13:0.21.11                (depends on 2.2.0)
[info]   	    +- io.chrisdavenport:vault_2.13:2.0.0                 (depends on 2.0.0)
[info]   	    +- io.chrisdavenport:unique_2.13:2.0.0                (depends on 2.0.0)
```

eed3si9n: > However, the callers in [the semantic versioning test](https://github.com/sbt/sbt/blob/develop/lm-ivy/src/test/scala/sbt/internal/librarymanagement/EvictionErrorSpec.scala#L73) have multiple associated revisions. Would you expect this?

This could be an implementation detail of either Coursier or sbt-coursier, but I think the caller seems to have either exactly one revision (requested is the same as resolved), or two revisions (requested revision and the resolved revision). In the latter case, `evicted: true` should contain the requested revision:

```
	org.typelevel:cats-effect_2.13
		- 2.2.0
			evicted: false
			callers: io.chrisdavenport:vault_2.13:2.0.0, co.fs2:fs2-core_2.13:2.4.5, io.chrisdavenport:unique_2.13:2.0.0, org.http4s:http4s-core_2.13:0.21.11
		- 2.0.0
			evicted: true
			callers: io.chrisdavenport:unique_2.13:2.0.0
		- 2.0.0
			evicted: true
			callers: io.chrisdavenport:vault_2.13:2.0.0
```

so for the purpose of eviction error, we should prioritize the display of `evicted: true`, and should be deterministic?

zainab-ali: In that case, the current logic is correct - we add the callers for `evicteds` first, then the callers for `winner`.

I've updated the semantic versioning test, although I'm not sure why the test in question has a different set of callers for `Test` and `Compile`.

Let me know if you'd like further changes or tests.

lm-ivy/src/test/scala/sbt/internal/librarymanagement/EvictionErrorSpec.scala
```diff
@@ -85,6 +85,13 @@ object EvictionErrorSpec extends BaseIvySpecification {
           "\t    +- org.http4s:http4s-core_2.13:0.21.11                (depends on 2.2.0)",
           "\t    +- io.chrisdavenport:vault_2.13:2.0.0                 (depends on 2.0.0)",
           "\t    +- io.chrisdavenport:unique_2.13:2.0.0                (depends on 2.0.0)",
+          "",
+          "\t* org.typelevel:cats-effect_2.13:3.0.0-M4 (early-semver) is selected over {2.0.0, 2.2.0}",
```
eed3si9n: Is this second entry for the Test configuration? We probably need a way to distinguish between `Compile` vs `Test`

zainab-ali: I've done this by propagating the configuration through to `EvictionErrorPair`. It involved some more collections manipulation, so I took the opportinity to rewrite the block to use `flatMap` and `groupMap` instead of `ListBuffer`. Let me know if you prefer the `ListBuffer` approach.

The error message now contains the configurations:
```diff
+ * org.typelevel:weaver-cats_3:0.11.1 (early-semver) is selected over 0.10.1 for test
	    +- default:sbt_9d04d0c7_3:0.1.0-SNAPSHOT              (depends on 0.11.1)
	    +- com.siriusxm:snapshot4s-weaver_3:0.2.2             (depends on 0.10.1)
```

If there are multiple, it contains the configuration set:
```
* org.typelevel:weaver-cats_3:0.11.1 (early-semver) is selected over 0.10.1 for {compile, test}
```

lm-core/src/main/scala/sbt/librarymanagement/EvictionWarning.scala
```diff
@@ -268,31 +269,20 @@ object EvictionWarning {
...
+  ): Seq[(ConfigRef, OrganizationArtifactReport)] = {
     val confs = report.configurations filter { x =>
-      options.configurations.contains[ConfigRef](x.configuration)
+      configurations.contains[ConfigRef](x.configuration)
     }
-    confs flatMap { confReport =>
-      confReport.details map { detail =>
-        if (
-          (detail.modules exists { _.evicted }) &&
-          !(buffer exists { x =>
```
zainab-ali, line 287: Is it necessary to deduplicate based on module name and organization within a single configuration report? If so, I'll add this `exists` check back in.

lm-core/src/main/scala/sbt/librarymanagement/EvictionWarning.scala
```diff
@@ -310,9 +300,21 @@ object EvictionWarning {
...
+      if (
+        (detail.modules exists { _.evicted }) &&
+        !(buffer exists { x =>
+          (x.organization == detail.organization) && (x.name == detail.name)
+        })
+      ) {
+        buffer += detail
+      }
+    }
+    val reports = buffer.toList.toVector
```
zainab-ali, line 317: I've left the behaviour of `evicted` unchanged.

It seems to differ from that in `update`. Based on the [earlier explaination](https://github.com/sbt/sbt/pull/9102#discussion_r3098486385), I assume that `evicted` has the pre `SBT 1.5.0` logic?

eed3si9n: Yea evicted is for information purposes / backward compatibility.

eed3si9n (approved):
Thanks!

PR #9106, opened 2026-04-16
[2.x] fix: Reconnect the thin client after reboot

### LINKED ISSUE #9095: [2.x] `reboot` causes sbt to exit
## steps

1. Create a minimal build
    ```properties
    # project/build.properties
    sbt.version=2.0.0-RC12
    ```
    ```scala
    // build.sbt
    ```
2. Run `sbt` in the project dir
3. Run `reboot`

## problem

`reboot` says it's booting the sbt server, then says the server closed:

```
sbt:root> reboot
[info] sbt server is booting up

sbt server connection closed.
```

## expectation

`reboot` should reboot the server and drop you back into the sbt shell

jonathanchang31:
Fixes #9095

**Problem**

On sbt 2.x, running `reboot` from an interactive thin client can cause the client to disconnect instead of reconnecting after the server restarts.

There are two issues in the reboot path:

1. Interactive reboot shutdowns can lose the reboot payload when there is no preserved exec id, so the client treats the shutdown as a plain disconnect.
2. The thin client can race the restart and reconnect against a stale `project/target/active.json` before the rebooted server is actually ready.

**Solution**

This change makes the reboot flow more explicit and more defensive:

- Always send reboot metadata for network-client shutdowns, even when there is no exec id to preserve.
- Make the thin client wait for the rebooted server endpoint to become connectable before reinitializing the session.
- Make the client wait for an in-flight reboot reconnect instead of trying to initialize a second connection in parallel.

**Verification**

- Reproduced the issue locally using the steps from #9095.
- Confirmed the change compiles:
  - `commandProj/compile`
  - `mainProj/compile`

I did once a manual end-to-end verification before merge.

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -174,7 +174,7 @@ class NetworkClient(
       }
     res match {
       case Some(r) => r
-      case None    =>
+      case None =>
         Thread.sleep(new java.util.Random().nextInt(20).toLong)
         connect(file, attempt + 1)
     }
@@ -265,7 +265,7 @@ class NetworkClient(
         try connect(portfile)
         catch {
           case e: NetworkClient.AccessDeniedException => throw e
-          case NonFatal(e)                            =>
+          case NonFatal(e) =>
             Thread.sleep(20)
             loop(Some(e))
         }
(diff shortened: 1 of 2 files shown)


eed3si9n:
I haven't had chance to take a look at this in detail yet, but thanks for working on this.

eed3si9n (changes requested):
Tested it locally but didn't seem to reconnect. Also I wanted to let you know that I've delisted sbt/sbt from Gittensor.

PR #9121, opened 2026-04-22
[2.x] fix: Fixes managedScalaInstance false support

### LINKED ISSUE #9105: When `managedScalaInstance := false` and a scala library is in the dependencies, the compilation fails since 1.12.0
## steps

1. Have this `build.sbt`: 
```
ThisBuild / scalaVersion := "2.13.16"

lazy val root = project.settings(
  autoScalaLibrary := false,
  managedScalaInstance := false,
  ivyConfigurations += Configurations.ScalaTool,
  libraryDependencies ++= Seq(
    "org.scala-lang" % "scala-library" % "2.13.16",
    "org.scala-lang" % "scala-compiler" % "2.13.16" % "scala-tool"
  )
)
```
Set sbt version to 1.11.7 in the project. Add some scala code to the `root` project. 

2. Start the sbt shell. 

3. Check the `scalaInstance` in the `root` project -> it's present. 

Image

4. Run `root/compile`. 

**Expected & actual result**: the sources are compiled 

5. Change the sbt version to 1.12.0 
6. Check the `scalaInstance` in the `root` project -> it's set to 0.0.0. 
`[info] Scala instance { version label 0.0.0, actual version 0.0.0, library jars: , compiler jars: , other jars:  }`
It's kind of expected because since [this commit](https://github.com/sbt/sbt/commit/ed7da85ef012b71919c8ce35b5e46c760db7766a), all projects with unmanaged scala instances have this empty scala instance with version 0.0.0
7. Run `root/compile`. 

**Expected**: the sources are compiled 
**Actual**: the compilation fails with exception: 

```
[info] compiling 1 Scala source to /Users/azdrojowa/IdeaProjects/sbt-demo/myTest/root/target/scala-2.13/classes ...
[error] ## Exception when compiling 1 sources to /Users/azdrojowa/IdeaProjects/sbt-demo/myTest/root/target/scala-2.13/classes
[error] java.lang.NoClassDefFoundError: scala/collection/IterableOnceOps
[error] java.base/java.lang.Class.getDeclaredConstructors0(Native Method)
[error] java.base/java.lang.Class.privateGetDeclaredConstructors(Class.java:2991)
[error] java.base/java.lang.Class.getConstructor0(Class.java:3186)
[error] java.base/java.lang.Class.getConstructor(Class.java:2205)
[error] java.base/java.util.ServiceLoader.getConstructor(ServiceLoader.java:623)
```
I did not investigate this further, but it looks like the compile task may not have the Scala jars (?). 
Maybe it's somehow related to this empty `scalaInstance`.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/9105
See also https://github.com/sbt/sbt/pull/8238

## Problem
When managedScalaInstance is set to false, we stopped creating ScalaInstance, but the documentation says the user can pull in their own dependencies.

## Solution
Attempt to construct a ScalaInstance from update report even when managedScalaInstance is set to false.

### DIFF
--- a/main/src/main/scala/sbt/internal/Compiler.scala
+++ b/main/src/main/scala/sbt/internal/Compiler.scala
@@ -64,12 +64,14 @@ object Compiler:
       val sh = Keys.scalaHome.value
       val app = Keys.appConfiguration.value
       val managed = Keys.managedScalaInstance.value
-      sh match
-        case Some(h) => scalaInstanceConfigFromHome(h)
-        case _ =>
-          val scalaProvider = app.provider.scalaProvider
-          if !managed then emptyScalaInstanceConfig
-          else scalaInstanceConfigFromUpdate(extraToolConf)
+      val configs = Keys.ivyConfigurations.value
+      (sh, extraToolConf) match
+        case (Some(h), _) => scalaInstanceConfigFromHome(h)
+        case _ if !managed =>
+          val extra = extraToolConf.getOrElse(Configurations.ScalaTool)
+          if configs.contains(extra) then scalaInstanceConfigFromUpdate(extraToolConf)
+          else emptyScalaInstanceConfig
+        case _ => scalaInstanceConfigFromUpdate(extraToolConf)
     }
 
   /**
--- a/sbt-app/src/sbt-test/project/scala-instance/test
+++ b/sbt-app/src/sbt-test/project/scala-instance/test
@@ -1,0 +1,1 @@
+> compile
--- a/sbt-app/src/sbt-test/project/scala-instance/test
+++ b/sbt-app/src/sbt-test/project/scala-instance/test
@@ -1,1 +1,3 @@
 > compile
+> doc
+> check
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -769,9 +769,8 @@ object Defaults extends BuildCommon with DefExtra {
       scalaCompilerBridgeBin := Def
         .ifS(Def.task {
           val sv = scalaVersion.value
-          val managed = managedScalaInstance.value
           val hasSbtBridge = ScalaArtifacts.isScala3(sv) || ScalaArtifacts.hasScala2SbtBridge(sv)
-          hasSbtBridge && managed
+          hasSbtBridge
         })(Compiler.compilerBridgeFromUpdate)(Def.task(Vector.empty))
         .value,
       scalaCompilerBridgeJars := (Def.taskDyn {
--- a/sbt-app/src/sbt-test/project/scala-instance/A.scala
+++ b/sbt-app/src/sbt-test/project/scala-instance/A.scala
@@ -1,0 +1,5 @@
+package example
+
+class A {
+  val x = 1
+}
--- a/sbt-app/src/sbt-test/project/scala-instance/build.sbt
+++ b/sbt-app/src/sbt-test/project/scala-instance/build.sbt
@@ -1,9 +1,19 @@
-scalaVersion := "2.13.16"
+import Configurations.{ ScalaTool, ScalaDocTool, ZincTool }
+
+@transient
+lazy val check = taskKey[Unit]("")
+lazy val scala213 = "2.13.16"
+scalaVersion := scala213
 autoScalaLibrary := false
 managedScalaInstance := false
-ivyConfigurations ++= List(Configurations.ScalaTool, Configurations.ZincTool)
+ivyConfigurations ++= List(ScalaTool, ScalaDocTool, ZincTool)
 libraryDependencies ++= Seq(
-  "org.scala-lang" % "scala-library" % "2.13.16",
-  "org.scala-lang" % "scala-compiler" % "2.13.16" % "scala-tool",
-  "org.scala-lang" % "scala2-sbt-bridge" % "2.13.16" % "zinc-tool",
+  "org.scala-lang" % "scala-library" % scala213,
+  "org.scala-lang" % "scala-compiler" % scala213 % ScalaTool,
+  "org.scala-lang" % "scala-compiler" % scala213 % ScalaDocTool,
+  "org.scala-lang" % "scala2-sbt-bridge" % scala213 % ZincTool,
 )
+check := {
+  val si = scalaInstance.value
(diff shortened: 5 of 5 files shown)


azdrojowa123:
Thank you :) 
Can this also be backported to sbt 1.x?

eed3si9n:
Yes, since it's broken in 1.x too it makes sense to backport it.

sbt-app/src/sbt-test/project/scala-instance/test
```diff
@@ -0,0 +1 @@
+> compile
```
anatoliykmetyuk, line 1: `compile` tests the original issue concern, however, if sbt API allows for it, would be good to verify here the Scala Instance is no longer `0.0.0`.

main/src/main/scala/sbt/internal/Compiler.scala
```diff
@@ -64,12 +64,14 @@ object Compiler:
...
-        case _ =>
-          val scalaProvider = app.provider.scalaProvider
-          if !managed then emptyScalaInstanceConfig
-          else scalaInstanceConfigFromUpdate(extraToolConf)
+      val configs = Keys.ivyConfigurations.value
+      (sh, extraToolConf) match
+        case (Some(h), _) => scalaInstanceConfigFromHome(h)
+        case _ if !managed =>
+          val extra = extraToolConf.getOrElse(Configurations.ScalaTool)
+          if configs.contains(extra) then scalaInstanceConfigFromUpdate(extraToolConf)
```
anatoliykmetyuk, line 72: Does this also induce the correct behaviour for `console` and `docs`?

E.g. for `console`:

https://github.com/sbt/sbt/blob/ac0f3f5c6f3b80e7f359c1827bc1ae6efce79bc1/main/src/main/scala/sbt/Defaults.scala#L705-L713

The `ScalaReplTool` is used as an `extra`. `configs.contains(extra)` would fail in unmanaged case, unless the user has manually added `ScalaReplTool` to the config:

https://github.com/sbt/sbt/blob/ac0f3f5c6f3b80e7f359c1827bc1ae6efce79bc1/main/src/main/scala/sbt/Defaults.scala#L3274-L3279

main/src/main/scala/sbt/internal/Compiler.scala
```diff
@@ -64,12 +64,14 @@ object Compiler:
...
-        case Some(h) => scalaInstanceConfigFromHome(h)
-        case _ =>
-          val scalaProvider = app.provider.scalaProvider
-          if !managed then emptyScalaInstanceConfig
-          else scalaInstanceConfigFromUpdate(extraToolConf)
+      val configs = Keys.ivyConfigurations.value
+      (sh, extraToolConf) match
+        case (Some(h), _) => scalaInstanceConfigFromHome(h)
+        case _ if !managed =>
+          val extra = extraToolConf.getOrElse(Configurations.ScalaTool)
```
anatoliykmetyuk, line 71: Is this necessary? In `scalaInstanceConfigFromUpdate`:

https://github.com/sbt/sbt/blob/ac0f3f5c6f3b80e7f359c1827bc1ae6efce79bc1/main/src/main/scala/sbt/internal/Compiler.scala#L276-L280

https://github.com/sbt/sbt/blob/ac0f3f5c6f3b80e7f359c1827bc1ae6efce79bc1/main/src/main/scala/sbt/internal/Compiler.scala#L335-L351

It appears we are already using the `ScalaTool` by default, the input only provides extra jars to add.

anatoliykmetyuk (changes requested):

main/src/main/scala/sbt/internal/Compiler.scala
eed3si9n: Yea, the behavior I introduced in https://github.com/sbt/sbt/pull/8238 was that if `managedScalaInstance` is `false`, we'll create an `emptyScalaInstance`; this PR attempts to partly roll that back, and we'll call `scalaInstanceConfigFromUpdate(...)` if the update report contains the tooling configuration, which normally does not exist if you've set `managedScalaInstance` to `false`. Since I'm checking `configs.contains(extra)`, I'd have to do `.getOrElse(Configurations.ScalaTool)` here.

main/src/main/scala/sbt/internal/Compiler.scala
eed3si9n: > Does this also induce the correct behaviour for `console` and `docs`?

I think so. If the user sets `managedScalaInstance` to `false`, they'd have to provide toolchain for REPL and docs in the appropriate configuration.

anatoliykmetyuk: Ok, if that's the intended contract, I think it would be good to document it here: https://www.scala-sbt.org/1.x/docs/Configuring-Scala.html#Configuring+Scala+tool+dependencies

Otherwise, the contract described there does not mention the names of other toolchains the user may need to provide, the only one I see is `ivyConfigurations += Configurations.ScalaTool`.

sbt-app/src/sbt-test/project/scala-instance/test
eed3si9n: Addressed in https://github.com/sbt/sbt/pull/9121/changes/ef47218abe5c6e483bbc852404ee42e3b43ac69a to check for the version number.

main/src/main/scala/sbt/internal/Compiler.scala
eed3si9n: Added `doc` in https://github.com/sbt/sbt/pull/9121/changes/ef47218abe5c6e483bbc852404ee42e3b43ac69a.

main/src/main/scala/sbt/internal/Compiler.scala
eed3si9n: Yea, the documentation probably became out of date when sbt 1.5.x added Scala 3 support, which added their own docs (https://github.com/sbt/librarymanagement/pull/359) and REPL artifacts (https://github.com/sbt/librarymanagement/pull/554).

anatoliykmetyuk (approved):
Looks good, we should update the docs to reflect the extra config needed for the scenario of unmanaged Scala + REPL, docs and other similar tasks.

eed3si9n:
Docs update - https://github.com/sbt/website/pull/1457

PR #9131, opened 2026-04-25
[2.x] feat: Expose `scripted / excludeFilter` and `scripted / includeFilter` keys for scripted test filtering.

arashi01:
@eed3si9n @xuwei-k Not sure if this is something that could be considered? In the process of migrating a bunch of plugins to sbt 2.x and cross-building against 1.x and faced this a few times.

## Problem

Plugin authors cross-building against sbt 1.x and 2.x have no build-level mechanism to skip scripted tests that cannot load on a given target. Current workarounds are dynamically creating/deleting `disabled` files for affected tests at runtime (which causes issues if using a `projectMatrix` based build with concurrency), or pinning a stale `sbt.version` in `project/build.properties` to trigger `isTestCompatible`; neither works with the build's own `pluginCrossBuild / sbtVersion`. 

sbt's own platform specific exclusions are handled by a hardcoded `windowsExclude` in the implementation.

## Solution

Expose `scripted / includeFilter` and `scripted / excludeFilter` in `ScriptedPlugin.projectSettings`

e.g 

```scala
scripted / excludeFilter := {
  val isSbt2 = scalaBinaryVersion.value == "3"
  new SimpleFileFilter(f => isSbt2 && f.getName == "unsupported")
}
```

This also allows us to replace `windowsExclude` with a an `excludeFilter` setting in our own build. Currently placed in `Scripted.sbtWindowsExcludeFilter` but could inline into `build.sbt` if preferred.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1011,7 +1011,9 @@ def scriptedTask(launch: Boolean): Def.Initialize[InputTask[Unit]] = Def.inputTa
       .filterNot(_.getName.contains("scala-compiler")),
     (bundledLauncherProj / Compile / packageBin).value,
     streams.value.log,
-    scriptedKeepTempDirectory.value
+    scriptedKeepTempDirectory.value,
+    (scripted / includeFilter).value,
+    (scripted / excludeFilter).value,
   )
 }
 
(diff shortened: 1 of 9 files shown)


eed3si9n:
Thanks. I'm generally supportive of this idea.

PR #9146, opened 2026-04-27
[2.x] perf: Use bytecodeHash for incremental test

### LINKED ISSUE #9108: [2.x] Poor `Test / definedTestDigests` performance due to `ClassStamper`
## steps

Unfortunately I don't have a minimal reproduction, I noticed the problem in my work codebase when running `testFull`.

## problem

`Test / definedTestDigests` took 23 minutes to complete on one project, nearly 4 minutes on another, and ~90 seconds on a few others.

### Tracing

I used `sbt --traces` to generate a trace and queried it on https://ui.perfetto.dev. Here are the slowest projects:

projects

### Flamegraph

I also used async-profiler to generate a flamegraph, which shows that `ClassStamper` is the culprit.

[flamegraph.html](https://github.com/user-attachments/files/26792999/flamegraph.html)

<details><summary>expand screenshot></summary>

flamegraph

</details>

## expectation

The `ClassStamper` should not take nearly that long to perform its computations.

## notes

Possibly related to #7751

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/9108

## Problem
There's perf issue building ClassStamp for the incremental test.

## Solution
This reuses the bytecodeHash created during compilation.

### DIFF
--- a/main/src/main/scala/sbt/internal/IncrementalTest.scala
+++ b/main/src/main/scala/sbt/internal/IncrementalTest.scala
@@ -19,6 +19,7 @@ import sbt.internal.util.Types.const
 import sbt.io.{ GlobFilter, IO, NameFilter }
 import sbt.protocol.testing.TestResult
 import sbt.util.{ ActionCache, BuildWideCacheConfiguration, CacheLevelTag, Digest, Logger }
+import sbt.util.CacheImplicits
 import sbt.util.CacheImplicits.given
 import scala.collection.concurrent
 import scala.collection.mutable
@@ -152,10 +153,11 @@ class ClassStamper(
 ):
   private val stamps = mutable.Map.empty[String, SortedSet[Digest]]
   private val internalStamps = mutable.Map.empty[String, SortedSet[Digest]]
-  private val vfStamps = mutable.Map.empty[VirtualFileRef, Digest]
   private lazy val analyses = classpath
     .flatMap(a => BuildDef.extractAnalysis(a.metadata, converter))
     .collect { case analysis: Analysis => analysis }
+  private val stampVf: VirtualFileRef => Digest =
+    CacheImplicits.virtualFileRefToDigest(_)(converter)
 
   /**
    * Given a classpath and a class name, this tries to create a SHA-256 digest.
@@ -185,7 +187,7 @@ class ClassStamper(
       val internalDeps = relations
         .internalClassDeps(className)
         .flatMap: otherCN =>
-          internalDepStamp(otherCN, analysis, alreadySeen + javaClassName, log)
+          internalStamp(otherCN, analysis, alreadySeen + javaClassName, log)
       // log.debug(s"  internalStamp: internalDeps: $className = $internalDeps")
       val internalJarDeps = relations
         .externalDeps(className)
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/internal/IncrementalTest.scala
```diff
@@ -196,17 +197,11 @@ class ClassStamper(
...
-      // TODO: substitute the above with
-      // val classDigests = analysis.apis.internal
-      //   .get(className)
-      //   .map: analyzed =>
-      //   0L // analyzed.??? we need a hash here
+      val classDigests = analysis.apis.internal
+        .get(className)
+        .toSet
+        .map: analyzed =>
+          Digest.dummy(37 * (17 + analyzed.transitiveBytecodeHash) + analyzed.bytecodeHash)
```
eed3si9n, line 206: This `transitiveBytecodeHash` is `0` for non-macros.

main/src/main/scala/sbt/internal/IncrementalTest.scala
```diff
@@ -222,6 +217,36 @@ class ClassStamper(
           val classNames = relations.productClassName.reverse(javaClassName)
           SortedSet(classNames.toSeq*).flatMap(internalStamp0)
 
+  private def internalDepStamp(
```
anatoliykmetyuk: `internalStamp` also traverses external and library deps, is there a reason why this method does not do so?

anatoliykmetyuk (approved):
LGTM, left a comment.

main/src/main/scala/sbt/internal/IncrementalTest.scala
anatoliykmetyuk: What is the rationale for the magic constants here? Also, it seems `dummy` stores an 8-byte `Long` data which is then widened to a 32-byte SHA-256 hash, is information loss a concern here?

eed3si9n: Adding 17 and multiplying 37 is a quick way to calculate a hashcode. I don't even remember where I saw it first, but I use it for example in Contraband to generate hash

https://github.com/sbt/sbt/blob/0f2c97345256a63873691429554864f038b2fa9b/run/src/main/contraband-scala/sbt/ForkOptions.scala#L40

Googling says it appears in Effective Java by Joshua Bloch.

eed3si9n: > Also, it seems `dummy` stores an 8-byte `Long` data which is then widened to a 32-byte SHA-256 hash, is information loss a concern here?

`bytecodeHash` is Long, and since `transitiveBytecodeHash` is mostly `0` I don't think it make that much difference. The just needs to be unique enough between the test runs.

main/src/main/scala/sbt/internal/IncrementalTest.scala
eed3si9n: Given `ATest` that depends on class `A`, that depends on `B`, previously I was calling `internalStamp` recursively, which then was calling their own `relations.externalDeps(className)`. I thought we just need to track library dependencies of `ATest` and `A`, but not `B` but perhaps that's too optimistic.

eed3si9n: I've restored the recursive check in https://github.com/sbt/sbt/pull/9146/changes/de6921079760c0dc47955cf54aecef6952733c6a, but also added in-memory cache to cache Digest value, so they'll be shared across subprojects.

PR #9172, opened 2026-05-03
[2.x] Add ForkOptions.connectionTimeout

xuwei-k:


run/src/main/contraband/run.contra
```diff
@@ -30,4 +30,6 @@ type ForkOptions {
 
   ## Use arguments file
   canUseArgumentsFile: Boolean @since("1.11.6")
+
+  timeout: scala.concurrent.duration.Duration @since("2.0.0")
```
eed3si9n: I suggest calling this `connectionTimeout` since timeout could mean timing out after user's app runs for 30s.

PR #9173, opened 2026-05-03
[2.x] ci: Avoid deprecated replaceAllLiterally

xuwei-k:
- https://github.com/scala/scala3/commit/b4aa12d2b04ee02f871ff973a21903184a1fce99
- https://github.com/scala/scala/blob/v2.12.21/src/library/scala/collection/immutable/StringLike.scala#L177-L184
- https://github.com/scala/scala3/blob/3.8.2/library/src/scala/collection/StringOps.scala#L746-L747


eed3si9n (approved):
Thanks!

PR #9188, opened 2026-05-05
[2.x] Fix warnings in launcherPackage sub project

xuwei-k:
```
[warn] scalaVersion for subproject launcherPackage fell back to a default value 2.12.21; declare it explicitly in build.sbt:
[warn]   scalaVersion := "2.12.21"
```


eed3si9n (approved):
Thanks!

PR #9193, opened 2026-05-06
[2.x] fix: Ensure resources are copied atomically

### LINKED ISSUE #9190: `java.lang.ClassNotFoundException: Main` on `sbt run` interrupt
## steps

```sh
# project/build.properties
sbt.version=2.0.0-RC12
# sbt.version=1.12.0  # Reproduces also on 1.12.0
```

```scala
// src/main/scala/Main.scala
object Main extends App:
  println("Main CAS race trial: 1")
```

```scala
// build.sbt
ThisBuild / scalaVersion := "3.8.3"
ThisBuild / version := "0.1.0-SNAPSHOT"

name := "dummy-hello-world-sbt"
```

Reproduction:

1. `sbt run`
2. Interrupt it with Ctrl+C
3. `sbt run`
4. There is a chance of `java.lang.ClassNotFoundException: Main`.

The problem is intermittent and depends on the Ctrl+C timing. The following Claude-written script, when placed in the project folder, reproduces the failure deterministically. The script was tested on Mac, on Linux make sure to change the `CAS_DIR` variable to wherever the cache is located on Linux.

<details>
<summary>Reproduction Script</summary>

```sh
#!/usr/bin/env bash
#
# Reproduces the sbt 2.x user-facing failure:
#   Error: Could not find or load main class Main
#
# This does not delete or manually corrupt sbt caches. It makes a normal project
# change by adding a large resource, runs `sbt run`, interrupts the sbt server
# while sbt is writing the real packaged jar, then runs `sbt run` again.
#
# The script also runs against sbt 1.x. There it watches the local target/
# directory instead of the sbt 2 content-addressed cache. The expected outcome
# on sbt 1 is that the next `sbt run` recovers cleanly (no main-class error),
# but the script does not assert that — it just runs to completion either way.

set -euo pipefail

SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
cd "$SCRIPT_DIR"

TRIALS="${TRIALS:-12}"
RESOURCE_MB="${RESOURCE_MB:-384}"
LOG_DIR="${LOG_DIR:-/tmp/sbt-mainclass-cas-race}"
WATCH_TIMEOUT_SECS="${WATCH_TIMEOUT_SECS:-300}"
CAS_DIR="${HOME}/Library/Caches/sbt/v2/cas"
MAIN_FILE="src/main/scala/Main.scala"
RESOURCE_FILE="src/main/resources/repro-large-resource.bin"

SBT_VERSION="$(awk -F= '/^[[:space:]]*sbt\.version[[:space:]]*=/ { gsub(/[[:space:]]/, "", $2); print $2; exit }' project/build.properties)"
SBT_MAJOR="${SBT_VERSION%%.*}"
THRESHOLD=$((RESOURCE_MB * 1024 * 1024 / 2))

echo "Detected sbt version: ${SBT_VERSION:-<unknown>} (major=${SBT_MAJOR:-?})"

mkdir -p "$LOG_DIR" "$(dirname "$RESOURCE_FILE")"

write_main() {
  local trial="$1"
  cat > "$MAIN_FILE" <<EOF
object Main extends App:
  println("Main CAS race trial: $trial")
EOF
}

write_resource() {
  local trial="$1"
  echo "Writing ${RESOURCE_MB}MB normal project resource for trial $trial"
  dd if=/dev/urandom of="$RESOURCE_FILE" bs=1m count="$RESOURCE_MB" status=none
}

server_pid() {
  ps -axo pid,command |
    awk '/sbt-launch.*\.jar/ && !/awk/ { print $1; exit }'
}

cleanup_sbt() {
  pkill -9 -f 'sbt-launch.*\.jar' 2>/dev/null || true
  pkill -9 -f '/sbtn' 2>/dev/null || true
}

# Locate a partially-written sbt output file large enough to be from the
# 384MB resource. For sbt 2 this is a content-addressed file under the global
# CAS; for sbt 1 it is a freshly-written file under the project's target/.
find_partial_output() {
  local marker="$1"

  if [ "$SBT_MAJOR" = "2" ]; then
    [ -d "$CAS_DIR" ] || return 0
    find "$CAS_DIR" -type f -newer "$marker" -name 'sha256-*' -print 2>/dev/null |
      while IFS= read -r path; do
        suffix="${path##*-}"
        case "$suffix" in
          ''|*[!0-9]*) continue ;;
        esac
        if [ "$suffix" -gt "$THRESHOLD" ]; then
          printf '%s\n' "$path"
          return 0
        fi
      done
    return 0
  fi

  [ -d target ] || return 0
  find target -type f -newer "$marker" -print 2>/dev/null |
    while IFS= read -r path; do
      size="$(stat -f%z "$path" 2>/dev/null || echo 0)"
      if [ "$size" -gt "$THRESHOLD" ]; then
        printf '%s\n' "$path"
        return 0
      fi
    done
}

kill_during_cas_write() {
  local trial="$1"
  local marker="$LOG_DIR/trial-${trial}.marker"
  local log="$LOG_DIR/trial-${trial}-interrupted.log"
  local out_file=""
  local pid=""

  touch "$marker"
  sbt run > "$log" 2>&1 &
  local client_pid="$!"

  local start_ts
  start_ts="$(date +%s)"

  while :; do
    out_file="$(find_partial_output "$marker" | head -1 || true)"
    if [ -n "$out_file" ]; then
      pid="$(server_pid || true)"
      if [ -n "$pid" ]; then
        local size
        size="$(stat -f%z "$out_file" 2>/dev/null || echo 0)"
        echo "trial=$trial killing sbt server pid=$pid while output is $size bytes: $out_file"
        kill -9 "$pid" 2>/dev/null || true
        kill -9 "$client_pid" 2>/dev/null || true
        wait "$client_pid" 2>/dev/null || true
        echo "$out_file" > "$LOG_DIR/trial-${trial}-output-file.txt"
        return 0
      fi
    fi

    local now
    now="$(date +%s)"
    if [ $((now - start_ts)) -ge "$WATCH_TIMEOUT_SECS" ]; then
      echo "trial=$trial did not catch a large output write within ${WATCH_TIMEOUT_SECS}s"
      kill -9 "$client_pid" 2>/dev/null || true
      wait "$client_pid" 2>/dev/null || true
      return 1
    fi

    sleep 0.05
  done
}

check_next_run() {
  local trial="$1"
  local log="$LOG_DIR/trial-${trial}-next-run.log"

  set +e
  sbt run > "$log" 2>&1
  local status="$?"
  set -e

  if rg -q "Could not find or load main class Main|ClassNotFoundException: Main" "$log"; then
    echo "REPRODUCED trial=$trial status=$status log=$log"
    tail -30 "$log"
    return 0
  fi

  echo "trial=$trial did not hit target error; status=$status log=$log"
  tail -20 "$log"
  return 1
}

# Kill any sbt processes left running from a previous session before starting,
# so the first trial starts from a clean slate. Intentionally not done on exit:
# if the script is interrupted, leftover sbt state is preserved for inspection.
cleanup_sbt

for trial in $(seq 1 "$TRIALS"); do
  write_main "$trial"
  write_resource "$trial"
  if kill_during_cas_write "$trial"; then
    if check_next_run "$trial"; then
      exit 0
    fi
  fi
done

echo "Did not reproduce target error after $TRIALS trials. Logs are in $LOG_DIR"
exit 1
```
</details>

What it does:

1. Writes a large (~300MB) resource file into the `src/main/resources` directory - so that the resulting jar packaging takes more time due to the long write operation.
2. Runs `sbt run` and races following steps against it.
3. Polls the executable `jar` file existence. Once it exists, it means the write operation is in progress or finished. Assume in progress due to high polling frequency and long duration of the operation due to large resource (1).
4. Upon condition (3) met, sends interrupt signal to the `sbt run` process, interrupting the write and leaving the jar in an inconsistent state.

## problem

After the reproduction script is run once, the project is left in an inconsistent state so that `sbt run` throws an exception:

```
anatolii@Anatoliis-MacBook-Air 01-dummy-hello-world-sbt % sbt run
[info] running (fork) Main
Error: Could not find or load main class Main
Caused by: java.lang.ClassNotFoundException: Main
Exception in thread "sbt-serverconnection-0" nonzero exit code returned from runner: 1
        at sbt.Run$.processExitCode(Run.scala:195)
        at sbt.ForkRun.run(Run.scala:38)
        at sbt.internal.client.NetworkClient.jvmRun$1(NetworkClient.scala:756)
        at sbt.internal.client.NetworkClient.clientSideRun(NetworkClient.scala:780)
        at sbt.internal.client.NetworkClient.clientSideRun(NetworkClient.scala:734)
        at sbt.internal.client.NetworkClient.splitToMessage$1(NetworkClient.scala:683)
        at sbt.internal.client.NetworkClient.onNotification(NetworkClient.scala:696)
        at sbt.internal.client.NetworkClient$$anon$3.onNotification(NetworkClient.scala:304)
        at sbt.internal.client.ServerConnection$$anon$1.$anonfun$run$2(ServerConnection.scala:50)
        at sbt.internal.client.ServerConnection$$anon$1.$anonfun$run$2$adapted(ServerConnection.scala:47)
        at scala.util.Either.fold(Either.scala:191)
        at sbt.internal.client.ServerConnection$$anon$1.run(ServerConnection.scala:47)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:832)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:808)
anatolii@Anatoliis-MacBook-Air 01-dummy-hello-world-sbt %
```

## expectation

`sbt run` steps should be transactional with respect to the written artifacts: if write of an artifact was interrupted, that artifact should be deleted rather than left in an inconsistent state, and the internal state of the project should reflect that step was not completed.

## notes

The bug is non-deterministic and depends on precise timing which is hard to achieve on modern machines, yet frequent enough to be bumped into by frequent sbt users. Hence the complicated reproduction script to mechanically induce the right conditions.

anatoliykmetyuk:
Fixes #9190 for sbt 2.

Targets:

- `Pkg.scala`, responsible for producing jars that are later used by the `run` task. Use atomic write to write that jar.
- `ActionStoreCache.scala` - responsible for creating CAS entries for task results caching. Check CAS existence and integrity (via digest hashes) to determine if we can reuse them. If either CAS file does not exist or the digest does not match, perform atomic write.

How atomic write works: it performs the write to a temporary directory, and moves the result to the target location on completion.

### DIFF
--- a/main-actions/src/main/scala/sbt/Pkg.scala
+++ b/main-actions/src/main/scala/sbt/Pkg.scala
@@ -226,9 +226,7 @@ object Pkg:
     val path = jar.getAbsolutePath
     log.debug("Packaging " + path + " ...")
     if (jar.exists)
-      if (jar.isFile)
-        IO.delete(jar)
-      else
+      if (!jar.isFile)
         sys.error(path + " exists, but is not a regular file")
     log.debug(sourcesDebugString(sources))
     IO.jar(sources, jar, manifest, time)
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -10,7 +10,7 @@ object Dependencies {
     sys.env.get("BUILD_VERSION") orElse sys.props.get("sbt.build.version")
 
   // sbt modules
-  val ioVersion = nightlyVersion.getOrElse("1.10.5")
+  val ioVersion = nightlyVersion.getOrElse("1.12.0")
   val zincVersion = nightlyVersion.getOrElse("2.0.0-M15")
 
   private val sbtIO = "org.scala-sbt" %% "io" % ioVersion
--- a/util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
+++ b/util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
@@ -1,6 +1,6 @@
 package sbt.util
 
-import java.io.{ IOException, RandomAccessFile }
+import java.io.{ ByteArrayInputStream, IOException }
 import java.nio.ByteBuffer
 import java.nio.file.{
   Files,
@@ -240,23 +240,30 @@ class DiskActionCacheStore(base: Path, converter: FileConverter) extends Abstrac
 
   def putBlob(input: InputStream, digest: Digest): Path =
     val casFile = toCasFile(digest)
-    IO.transfer(input, casFile.toFile())
-    casFile
+    if isCompleteBlob(casFile, digest) then casFile
+    else
+      IO.transfer(input, casFile.toFile())
+      casFile
 
   def putBlob(input: ByteBuffer, digest: Digest): Path =
     val casFile = toCasFile(digest)
-    input.flip()
-    val file = RandomAccessFile(casFile.toFile(), "rw")
-    try
-      file.getChannel().write(input)
+    if isCompleteBlob(casFile, digest) then casFile
+    else
+      input.flip()
+      val bytes = new Array[Byte](input.remaining())
+      input.get(bytes)
+      IO.transfer(new ByteArrayInputStream(bytes), casFile.toFile())
       casFile
-    finally file.close()
+
+  private def isCompleteBlob(casFile: Path, digest: Digest): Boolean =
+    try Files.exists(casFile) && Digest.sameDigest(casFile, digest)
+    catch case _: NoSuchFileException => false
 
   private def getBlobs(refs: Seq[HashedVirtualFileRef]): Seq[VirtualFile] =
     refs.flatMap: r =>
       try
         val casFile = toCasFile(Digest(r))
-        if casFile.toFile().exists then
+        if isCompleteBlob(casFile, Digest(r)) then
           r match
             case p: PathBasedFile => Some(p)
             case _ =>
@@ -270,7 +277,7 @@ class DiskActionCacheStore(base: Path, converter: FileConverter) extends Abstrac
     refs.flatMap: r =>
       try
         val casFile = toCasFile(Digest(r))
-        if casFile.toFile().exists then
+        if isCompleteBlob(casFile, Digest(r)) then
           // println(s"syncBlobs: $casFile exists for $r")
           Some(syncFile(r, casFile, outputDirectory))
         else None
@@ -384,7 +391,7 @@ class DiskActionCacheStore(base: Path, converter: FileConverter) extends Abstrac
     refs.flatMap: r =>
       try
         val casFile = toCasFile(Digest(r))
-        if casFile.toFile().exists then Some(r)
+        if isCompleteBlob(casFile, Digest(r)) then Some(r)
         else None
       // Digest(r) can throw NoSuchFileException
       catch case _: NoSuchFileException => None
--- a/util-cache/src/test/scala/sbt/util/ActionCacheTest.scala
+++ b/util-cache/src/test/scala/sbt/util/ActionCacheTest.scala
@@ -1,5 +1,6 @@
 package sbt.util
 
+import java.io.{ IOException, InputStream }
 import java.nio.charset.StandardCharsets
 import java.nio.file.{ Files, Path, Paths }
 import java.util.Optional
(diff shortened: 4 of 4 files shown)


eed3si9n (approved):
Thanks!

PR #9196, opened 2026-05-07
[1.x] fix: Ensure resources are copied atomically

anatoliykmetyuk:
Fixes https://github.com/sbt/sbt/issues/9190 for sbt 1.

Using io 1.12 appears to fix the issue in question for sbt 1.


eed3si9n (approved):
Thanks!

PR #9211, opened 2026-05-13
[2.x] refactor: Use `Files.readString` instead of `readAllBytes`

xuwei-k:
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Files.html#readString(java.nio.file.Path)


eed3si9n (approved):
Thanks!

PR #9214, opened 2026-05-14
[2.x] feat: Print cross-project test failure recap at end of aggregated run

### LINKED ISSUE #2998: Test failures are hard to find
## steps

Run test in a multi-project build. I take specs2 as example.

## problem

If one of the aggregate projects fails a test (but not the last one), I get something like the following output

```
[info] Passed: Total 1098, Failed 0, Errors 0, Passed 1098, Pending 1
[error] (core/test:test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 306 s, completed Mar 9, 2017 5:09:06 PM
```

Notice that all tests seem to pass (0 failures, 0 errors), but still there is a failure somewhere! That's probably many screens of output before.

## expectation

A recap of failed projects and their test summary, like

```
[error] Project: core
[error] Error: Total 745, Failed 1, Errors 2, Passed 742
[error] Failed tests:
[error] 	org.specs2.specification.BeforeAfterAllSpec
[error] Error during tests:
[error] 	org.specs2.mutable.MutableSpec
[error] 	org.specs2.text.LinesContentDifferenceSpec
```

## notes

Note that searching for `[error]` yields a ton of false hits if there's a stack trace involved so it's really quite annoying, especially in the CI, where the log can get large and (at least in Jenkins) you need to click one more time to load all of the log output.

BrianHotopp:
## Summary

Fixes #2998. When `test` is run at the prompt in a multi-project build, sbt aggregates `<proj>/test` across subprojects. Each subproject's `TestResultLogger` emits its own per-task summary inline, but the aggregation layer never restated which projects failed. The terminal `sbt.TestsFailedException` line carried no project names or failing test classes, forcing CI users to scrollback-grep through thousands of log lines to find what broke.

This PR adds a single recap block printed once after the aggregated run completes, last in the output stream, with failures sorted by `taskName` for diff-grep stability:

```
[error] Test failures recap (2 test tasks failed):
[error]   a / Test / testQuick: Total 1, Failed 1, Errors 0, Passed 0
[error]     Failed tests:
[error]       FailingTestA
[error]   c / Test / testQuick: Total 1, Failed 1, Errors 0, Passed 0
[error]     Failed tests:
[error]       FailingTestC
```

## Release notes

- **Behavior change:** `TestResultLogger.Defaults.Main.run` no longer throws on `TestResult.Failed` / `TestResult.Error`. Failure propagation now lives in the task wrapper (`Defaults.testFull` / `inputTests0`). Plugins that override `testResultLogger` should not assume the logger throws; throw from the task wrapper instead. The trait signature (`def run(...): Unit`) and its scaladoc (which documents logging only) are unchanged.

## Design

In-band via the result graph - no singleton, no global state, no side files:

- `TestsFailedException` carries `(taskName: String, testOutput: Option[Tests.Output])` via a `private[sbt]` primary constructor. The public no-arg constructor delegates with empty defaults for back-compat. MiMa-clean against 2.0.0-RC12.
- The failure throw lives in `Defaults.testFull` and `inputTests0` (the task wrappers), not in `TestResultLogger.Defaults.Main.run`. The default logger logs and returns; the wrapping task matches on `output.overall` and throws `TestsFailedException(taskName, Some(output))` on Failed/Error. This keeps the failure signal out of the user-overridable `testResultLogger` seam.
- The task wrappers also catch no-detail `TestsFailedException` (legacy `executeTests` adapters, third-party `Tests.Setup` actions, anything constructing the back-compat no-arg ctor) on its way out and re-throw with the task name attached, so the recap renders `<taskName>: (no details)` instead of `<unknown>: (no details)` for those paths.
- `TestRecap.collect(i: Incomplete)` walks `Incomplete.allExceptions` and projects each `TestsFailedException` to a `Failure(taskName, testOutput)`. Identity-deduplicated via the underlying `IDSet[Throwable]`, so a single failing task shared across multiple paths in the Incomplete DAG is counted once.
- `TestRecap.render` / `formatTo` produce output sorted by `(taskName.isEmpty, taskName)` (empty names last, alphabetical otherwise). `printStandard`'s counts logic was factored into a shared `TestResultLogger.Defaults.countsString` so the recap and per-task summaries can't drift. Suite names go through `NameTransformer.decode` to match the existing `printFailures` rendering (revisiting this for both sites is a separate cleanup).
- The recap is printed **after** `state.handleError(i)` so it is the very last thing the user sees in CI output, below the trailing `Tests unsuccessful` line.
- The collected failures are also stashed on `State.attributes` under `TestRecap.recapKey` (a `private[sbt] AttributeKey[Vector[Failure]]("testRecap")`) for in-JVM tooling / IDE plugin consumption.

### Lifecycle: monotonic-latest-failure

The State attribute is written whenever a run produces at least one `TestsFailedException` and is **never proactively cleared**. A successful test run after a failure leaves the stale attribute in place; the next failure overwrites it. We deliberately don't try to recognize "this run included a test task" at the aggregation boundary -- that would require a hardcoded list of test-task labels (test/testFull/testQuick/testOnly/testSelected) and would silently miss plugin-defined test tasks.

Trade-off: an IDE/BSP plugin reading `recapKey` after the user re-ran tests successfully will see the previous failure. We accept this -- the recap text itself was already printed once when the failure happened; the State attribute is supplementary structured access.

## Cross-cutting behavior

- `+test` cross-build: each Scala-version iteration is its own `timedRun` -> one recap per version.
- `;test;package`: each `;` segment is its own `timedRun` -> one recap per segment.
- `~test`: each watch cycle re-enters `runTasks` -> clean recap per iteration.
- `keepGoing` doesn't exist in sbt 2.x - sibling failures are aggregated as `Result.Inc` causes after `runKeep` returns; the recap fires after `state.handleError`, so all completed failures are visible.
- Parallelism: `Test/test` carries `Tags.Test -> 1` and runs N subproject test tasks in parallel by default. There is no shared mutable state; failures are collected from the Incomplete tree after `timedRun` returns, then sorted for stable rendering.

## Known gap

Hard JVM crashes in forked test workers that don't surface as `TestsFailedException` (e.g., `MessageOnlyException` from `ForkTests.React.notifyExit` on `SIGKILL` / `System.exit`) won't appear in the recap. They still surface via sbt's normal error trail. A follow-up suggested by @eed3si9n -- recording task entry as well as completion -- is the right shape; tracked as #9225.

## Test plan

- [x] Unit tests in `main-actions/src/test/scala/sbt/internal/testing/TestRecapTest.scala`: 13/13 pass. Cover `collect` on synthetic Incomplete trees (with payload, no-payload stub, **rebranded-with-task-name-only**, non-TestsFailedException skipped, **DAG dedup**); `render` shape (header, counts, fallbacks, **stable sort**, singular/plural); `formatTo` log-call count and level.
- [x] Scripted test `tests/multi-failure-recap` with end-to-end content verification: `verifyRecap` Command runs `test` via `Command.process` and asserts the recap state attribute -- all inside a single sbt JVM. Header, project membership (a and c failing, b not), suite-result counts, and rendered output are all checked.
- [x] `actionsProj/scalafmtCheckAll` + `mainProj/scalafmtCheckAll` + `actionsProj/Test/scalafmtCheckAll` clean.
- [x] `actionsProj/mimaReportBinaryIssues` clean against `org.scala-sbt:actions:2.0.0-RC12`. No filters needed.

The scripted couldn't verify recap content across a `-> X; > check` boundary because scripted spawns a fresh sbt JVM after a failing statement (`SbtHandler.onNewSbtInstance`'s catch block calls `finish`; empirically `ManagementFactory.getRuntimeMXBean.getName` returns different PIDs). The `Command.process` approach side-steps this by staying in one JVM.

## AI assistance disclosure

Drafted with Claude Code (Claude Opus 4.7). Iterated through six rounds of review feedback; engineering judgment calls (monotonic-latest-failure lifecycle, Command-based scripted verification, throw site in the task wrapper rather than the user-overridable logger, no-detail rebrand on escape) made in dialogue with the reviewers.

Generated with Claude Code.

### DIFF
--- a/main/src/main/scala/sbt/internal/Aggregation.scala
+++ b/main/src/main/scala/sbt/internal/Aggregation.scala
@@ -143,22 +143,13 @@ object Aggregation {
     showRun(complete, show)
     complete.results match
       case Result.Inc(i) =>
-        // Collect the per-task TestsFailedException payloads off the
-        // Incomplete tree *before* handleError mutates state, then emit the
-        // recap after handleError so it is the last thing the user sees.
         val failures = sbt.internal.testing.TestRecap.collect(i)
         val afterHandle = complete.state.handleError(i)
         if failures.nonEmpty then
           sbt.internal.testing.TestRecap.formatTo(afterHandle.log, failures)
           afterHandle.put(sbt.internal.testing.TestRecap.recapKey, failures)
         else afterHandle
-      case Result.Value(_) =>
-        // Leave any previously-set recap intact. A successful run after a
-        // failure does not clear the State attribute -- the next failure
-        // will overwrite it. This avoids guessing whether a particular
-        // command should be considered "test-related" (which would require
-        // a hardcoded label list or Tags-based detection).
-        complete.state
+      case Result.Value(_) => complete.state
 
   def printSuccess(
       start: Long,
--- a/main/src/main/scala/sbt/internal/Aggregation.scala
+++ b/main/src/main/scala/sbt/internal/Aggregation.scala
@@ -139,13 +139,25 @@ object Aggregation {
       extra: DummyTaskMap,
       show: ShowConfig
   )(using display: Show[ScopedKey[?]]): State =
-    sbt.internal.testing.TestRecap.clear()
     val complete = timedRun[A1](s, ts, extra)
     showRun(complete, show)
-    sbt.internal.testing.TestRecap.formatTo(complete.state.log)
     complete.results match
-      case Result.Inc(i)   => complete.state.handleError(i)
-      case Result.Value(_) => complete.state
+      case Result.Inc(i) =>
+        // Collect the per-task TestsFailedException payloads off the
+        // Incomplete tree *before* handleError mutates state, then emit the
+        // recap after handleError so it is the last thing the user sees.
+        val failures = sbt.internal.testing.TestRecap.collect(i)
+        val afterHandle = complete.state.handleError(i)
+        if failures.nonEmpty then
+          sbt.internal.testing.TestRecap.formatTo(afterHandle.log, failures)
+          sbt.internal.testing.TestRecap.writeArtifact(
+            afterHandle.baseDir,
+            sbt.internal.testing.TestRecap.format(failures),
+          )
+        afterHandle
+      case Result.Value(_) =>
+        sbt.internal.testing.TestRecap.deleteArtifact(complete.state.baseDir)
+        complete.state
 
   def printSuccess(
       start: Long,
--- a/main/src/main/scala/sbt/internal/Aggregation.scala
+++ b/main/src/main/scala/sbt/internal/Aggregation.scala
@@ -150,14 +150,15 @@ object Aggregation {
         val afterHandle = complete.state.handleError(i)
         if failures.nonEmpty then
           sbt.internal.testing.TestRecap.formatTo(afterHandle.log, failures)
-          sbt.internal.testing.TestRecap.writeArtifact(
-            afterHandle.baseDir,
-            sbt.internal.testing.TestRecap.format(failures),
-          )
-        afterHandle
+          afterHandle.put(sbt.internal.testing.TestRecap.recapKey, failures.toVector)
+        else afterHandle
       case Result.Value(_) =>
-        sbt.internal.testing.TestRecap.deleteArtifact(complete.state.baseDir)
-        complete.state
+        // Only clear stale recap state when this invocation actually
+        // included a test task; arbitrary `compile` / `update` / `reload`
+        // commands between a failed test run and the user inspecting the
+        // recap must leave the recap intact.
+        if includesTestTask(ts) then complete.state.remove(sbt.internal.testing.TestRecap.recapKey)
+        else complete.state
 
   def printSuccess(
       start: Long,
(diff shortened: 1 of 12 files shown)


eed3si9n:
Thanks for the contribution!

main/src/main/scala/sbt/internal/Aggregation.scala
```diff
@@ -142,8 +142,23 @@ object Aggregation {
...
+          sbt.internal.testing.TestRecap.formatTo(afterHandle.log, failures)
+          afterHandle.put(sbt.internal.testing.TestRecap.recapKey, failures)
+        else afterHandle
+      case Result.Value(_) =>
+        // Leave any previously-set recap intact. A successful run after a
+        // failure does not clear the State attribute -- the next failure
+        // will overwrite it. This avoids guessing whether a particular
+        // command should be considered "test-related" (which would require
+        // a hardcoded label list or Tags-based detection).
+        complete.state
```
eed3si9n: Given that Aggregation's main function isn't for testing, I don't think it would be good idea to leave this comment block here.

BrianHotopp: Removed in 1d8e615e4. The lifecycle / sequencing rationale lives in `TestRecap`'s doc-comment.

eed3si9n (approved):
Thanks!

PR #9226, opened 2026-05-16
[2.x] fix: Collapse duplicate subtrees in dependency tree rendering

### LINKED ISSUE #6886: `dependencyBrowseTree` throws `java.lang.OutOfMemoryError`
sbt version - `1.6.1`

In large enough project `dependencyBrowseTree` task will throw OOM exception. Looking at the implementation, the tree is transformed to JSON and the result is accumulated in memory, which is written into a file on following steps. Would it be an option to write the tree directly into a file as JSON? I have a feeling this would solve the issue.

If this looks like a beneficial improvement, I could work on it.

BrianHotopp:
## Summary

Fixes #6886. `dependencyTree` and `dependencyBrowseTree` re-explore the same node once per incoming edge, so a DAG with N levels and M children per node produces O(M^N) output -- enough heap to OOM real builds. [Friendseeker's repro](https://github.com/sbt/sbt/issues/6886#issuecomment-1986178407) demonstrates it directly.

## Fix

Add a `visited` set to the renderer's recursion. First occurrence renders in full; subsequent occurrences collapse to `+- <id> (*)` (ASCII) or a `<id> (*)` leaf (JSON), matching [Maven `dependency:tree`](https://maven.apache.org/plugins/maven-dependency-plugin/tree-mojo.html). Cycle detection (`(cycle)`) is unchanged.

Applies to `Graph.toAscii` (used by `dependencyTree` and `inspect tree` via `SettingGraph.dependsAscii`) and `TreeView.processSubtree` (the JSON for `dependencyBrowseTree`). `visited` is scoped to a single root's subtree; cross-root dedup is tracked as #9227. Canonical-occurrence rule is "first visited in DFS order" (commented in source; pinned by test).

## Behavior change

- New marker `(*)` in `dependencyTree`, `dependencyBrowseTree`, and `inspect tree`. A line ending in `<id> (*)` is a back-pointer to the canonical occurrence in the same render, not a distinct dependency.
- Output for users without large DAGs is unchanged.
- Documented in `notes/2.0.0/dependency-tree-dedup.md`.
- Pre-existing fixtures (`asciiGraphWidth/{when-is-default,when-is-20}/expected.txt`, `whatDependsOn{,-without-previous-initialization}/build.sbt`) updated to reflect the collapsed form.

## Tested

- 11 unit tests in `TreeViewTest` + new `AsciiTreeTest` (including a 12-node exponential DAG bounded to < 200 lines; 2^11 without dedup).
- New scripted test `dependency-graph/dag-dedup` (local multi-project diamond; asserts `common (*)` on a single line).
- Full `scripted dependency-graph/*` (12 tests) pass, including the 4 updated fixtures.
- `scalafmtCheckAll` and `mimaReportBinaryIssues` clean (renderers are `sbt.internal`).

## AI assistance disclosure

Drafted with Claude Code (Claude Opus 4.7). Diagnosis credit to @Friendseeker.

### DIFF
--- a/main/src/test/scala/sbt/internal/graph/rendering/AsciiTreeTest.scala
+++ b/main/src/test/scala/sbt/internal/graph/rendering/AsciiTreeTest.scala
@@ -1,0 +1,78 @@
+/*
+ * sbt
+ * Copyright 2023, Scala center
+ * Copyright 2011 - 2022, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.internal.graph.rendering
+
+import sbt.internal.Graph
+import verify.BasicTestSuite
+
+object AsciiTreeTest extends BasicTestSuite:
+
+  private def render(adj: Map[Int, Seq[Int]], root: Int, width: Int = 80): String =
+    Graph.toAscii[Int](root, n => adj.getOrElse(n, Nil), _.toString, width)
+
+  test("strict tree renders without markers") {
+    val adj = Map(0 -> Seq(1, 2), 1 -> Seq(3))
+    val out = render(adj, 0)
+    assert(!out.contains("(*)"), s"unexpected (*):\n$out")
+    assert(!out.contains("(cycle)"), s"unexpected (cycle):\n$out")
+    assert(out.contains("3"), s"node 3 missing:\n$out")
+  }
+
+  test("diamond DAG collapses the duplicate subtree to a single (*) line") {
+    //   0
+    //  / \
+    // 1   2
+    //  \ /
+    //   3
+    val adj = Map(0 -> Seq(1, 2), 1 -> Seq(3), 2 -> Seq(3))
+    val out = render(adj, 0)
+    assert(out.contains("3 (*)"), s"missing collapsed `3 (*)`:\n$out")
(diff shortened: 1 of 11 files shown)


main/src/test/scala/sbt/internal/graph/rendering/AsciiTreeTest.scala
```diff
@@ -0,0 +1,95 @@
...
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import sbt.internal.Graph
+
+/**
+ * Covers `Graph.toAscii`'s DAG handling (sbt/sbt#6886). The renderer is
+ * generic over node type; these tests drive it with `Int` to keep the
+ * setup compact.
+ */
+class AsciiTreeTest extends AnyFlatSpec with Matchers {
```
eed3si9n: For fresh tests, could you create HedgeHog or Verify plz?
https://github.com/sbt/sbt/blob/develop/contributing-docs/04_unit_tests.md

BrianHotopp: Converted to `verify.BasicTestSuite` in f463fe1fe. (TreeViewTest extensions left on ScalaTest -- file pre-exists -- happy to convert that too if you want.)

eed3si9n (approved):
Thanks!

PR #9228, opened 2026-05-16
[2.x] fix: Remove explicit equals and hashCode in LabeledFunctions

### LINKED ISSUE #9224: [2.x] LabeledFunctions `equals` broken
## steps

Remove `package private` and `sbt console` with sbt develop branch

https://github.com/sbt/sbt/blob/b3d29b6bac53cb0f7c57fd17eb6fcec9717faae2/main-command/src/main/scala/sbt/internal/LabeledFunctions.scala#L22

```diff
-private[sbt] object LabeledFunctions {
+object LabeledFunctions {
```

```
Welcome to Scala 3.8.3 (21.0.11, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.
                                                                                                                                                                                                                 
scala> import sbt.internal.LabeledFunctions.*
                                                                                                                                                                                                                 
scala> val f1: Int => Int = a => a
val f1: Int => Int = Lambda/0x00000070036c0200@1740e538
                                                                                                                                                                                                                 
scala> val f2 = f1.label("label")
val f2: Int => Int = label
                                                                                                                                                                                                                 
scala> f2 == f1
val res0: Boolean = true
                                                                                                                                                                                                                 
scala> f1 == f2
val res1: Boolean = false
```

## problem

https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/lang/Object.html#equals(java.lang.Object)

> It is symmetric: for any non-null reference values x and y, x.equals(y) should return true if and only if y.equals(x) returns true.

<!-- Next, describe the problem, or what you think is the problem. -->

## expectation

<!-- Describe what you think should've happened. -->

## notes

https://github.com/sbt/sbt/blob/b3d29b6bac53cb0f7c57fd17eb6fcec9717faae2/main-command/src/main/scala/sbt/internal/LabeledFunctions.scala#L118-L126

xuwei-k:
- fix https://github.com/sbt/sbt/issues/9224

### DIFF
--- a/main-command/src/main/scala/sbt/internal/LabeledFunctions.scala
+++ b/main-command/src/main/scala/sbt/internal/LabeledFunctions.scala
@@ -108,44 +108,20 @@ private[sbt] object LabeledFunctions {
   private class LabeledFunction0[+R](private val f: () => R, label: String) extends (() => R) {
     override def apply(): R = f()
     override def toString: String = label
-    override def equals(o: Any): Boolean = o match {
-      case that: LabeledFunction0[R] @unchecked => this.f == that.f
-      case that: (() => R) @unchecked           => this.f == that
-      case _                                    => false
-    }
-    override def hashCode: Int = f.hashCode
   }
   private class LabeledFunction1[-T, +R](private val f: T => R, label: String) extends (T => R) {
     override def apply(t: T): R = f(t)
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #9247, opened 2026-05-19
[2.x] Improve deprecation UX for watchSources

### LINKED ISSUE #9187: [2.x] `watchSources` contains `File` type
### build.sbt 

```scala
watchSources += file("foo")
```

### project/build.properties 

```properties
sbt.version=2.0.0-RC12
```

## problem

```
  1 |watchSources += file("foo")
    |^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |java.io.File and Path are not valid output types for a cached task.
    |Consider using one of the following alternatives:
    |  - xsbti.HashedVirtualFileRef
    |  - xsbti.VirtualFileRef
    |  - xsbti.VirtualFile
    |If caching is not needed, annotate the key with @transient, or wrap the task in Def.uncached.
    |See https://www.scala-sbt.org/2.x/docs/en/concepts/caching.html#caching-files
    |---------------------------------------------------------------------------
    |Inline stack trace
    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
    |This location contains code that was inlined from build.sbt:1
     ---------------------------------------------------------------------------
```
## notes

- https://github.com/sbt/io/blob/ede7b9c5b1144d1db81070b3d548270da5647338/io/src/main/scala/sbt/internal/io/SourceModificationWatch.scala#L193-L194
- https://github.com/sbt/sbt/blob/63d5a1103bb77a6f4bed6d26c72e548afe25fb6e/main/src/main/scala/sbt/Keys.scala#L140-L141

anatoliykmetyuk:
Fixes #9187 

The deprecated API involved is excluded from caching and annotated with a deprecation annotation. Sites where this API is used in sbt/sbt are excluded from warning checking to ensure `-Werror` does not get tripped over those.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -333,7 +333,8 @@ object Defaults extends BuildCommon with DefExtra {
         new AppenderSupplier:
           def apply(s: ScopedKey[?]): Seq[Appender] = Nil
       },
-      watchSources :== Nil, // Although this is deprecated, it can't be removed or it breaks += for legacy builds.
+      // Although this is deprecated, it can't be removed or it breaks += for legacy builds.
+      ((watchSources :== Nil): @nowarn("cat=deprecation")),
       skip :== false,
       taskTemporaryDirectory := {
         val base = BuildPaths.globalTaskDirectoryStandard(appConfiguration.value.baseDirectory)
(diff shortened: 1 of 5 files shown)


eed3si9n (approved):
Thanks!

xuwei-k (approved):

PR #9250, opened 2026-05-19
[2.x] fix: Apply dependencyMode filtering to internal project deps

### LINKED ISSUE #9009: [2.x] `dependencyMode := Direct` does not appear to trigger on module dependencies
## steps

> cat build.sbt
```
ThisBuild / dependencyMode := DependencyMode.Direct

lazy val Core = (project in file("Core"))

lazy val LibA = (project in file("LibA")).dependsOn(Core)

// LibB depends on LibA and Core
lazy val LibB = (project in file("LibB"))
  .dependsOn(LibA) // missing "Core"!
```

## problem

`LibB` compiles, despite `.dependsOn(...)` not listing `Core`.

## expectation

If `LibB` depends on `LibA` and `Core`, then under `DependencyMode.Direct` it should be required to write
```
lazy val LibB = (project in file("LibB"))
  .dependsOn(Core, LibA)
```

BrianHotopp:
## Summary

Fixes #9009. `dependencyMode := DependencyMode.PlusOne` strips every internal project dependency from the compile classpath, including direct ones -- `LibA/compile` fails with `Not found: CoreClass` despite `LibA.dependsOn(Core)`. Direct mode happens to work today but only because `filterByDirectDeps` matches against `allDependencies` (which includes `projectDependencies`); the bug surface is one step removed.

Diagnosis on the issue by @eureka0928. `ClasspathImpl.filterByPlusOne` resolves the direct-dep key set from `UpdateReport`, which only contains externally resolved (LM) modules. Internal project refs never appear there, so `resolvedDirectKeys` is empty for internal entries; the final filter strips every internal classpath entry whose `moduleIDStr` identifies an internal project.

## Fix

Per @eed3si9n's preferred direction in the thread: split internal vs external classpath upfront, route each through its own mode-aware filter.

- `ClasspathImpl.filterInternalByMode` (new) walks `BuildDependencies.classpath` to compute allowed project refs for the mode (one hop for `Direct`, two for `PlusOne`, unrestricted for `Transitive`), resolves them to `(org, name)` via each project's `projectID` setting, and keeps internal entries whose `moduleIDStr` is in that set.
- `filterByDependencyMode` is now called only on `externalDependencyClasspath` (which is genuinely external by construction: `unmanagedClasspath ++ managedClasspath`), so it no longer has to guess about internal entries.
- `filteredDependencyClasspath` and the pipelining branch of `dependencyPicklePath` both apply the split.

## Tested

New scripted `dependency-management/dependency-mode-internal` covers a four-project diamond `core <- libA <- libB <- libC`:

- `Direct` on `libA` -- includes `core`
- `Direct` on `libB` -- excludes `core` (transitive)
- `PlusOne` on `libB` -- includes `core` (one hop, the #9009 regression)
- `PlusOne` on `libC` -- includes `libA`, excludes `core` (two hops)

Existing `dependency-management/dependency-mode` and `dependency-management/dependency-mode-test-classpath` still pass. `scalafmtCheckAll` and `mimaReportBinaryIssues` clean (`ClasspathImpl` is `sbt.internal`).

## AI assistance disclosure

Drafted with Claude Code (Claude Opus 4.7). Diagnosis credit @eureka0928, directional steer @eed3si9n.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2736,16 +2736,24 @@ object Classpaths {
           dependencyMode.value match
             case DependencyMode.Transitive =>
               Def.task { dependencyClasspath.value }
-            case _ =>
+            case mode =>
               Def.task {
-                ClasspathImpl.filterByDependencyMode(
-                  dependencyMode.value,
+                val internalFiltered = ClasspathImpl.filterInternalByMode(
+                  mode,
+                  thisProjectRef.value,
+                  settingsData.value,
+                  buildDependencies.value,
+                  internalDependencyClasspath.value,
+                )
+                val externalFiltered = ClasspathImpl.filterByDependencyMode(
+                  mode,
                   allDependencies.value,
                   projectID.value,
                   classpathConfiguration.value,
                   updateFull.value,
-                  dependencyClasspath.value,
+                  externalDependencyClasspath.value,
                 )
+                internalFiltered ++ externalFiltered
               }
         })
         .value,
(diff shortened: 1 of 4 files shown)


eed3si9n (approved):
Thanks!

PR #9253, opened 2026-05-20
[2.x] Improve performance of incremental test

mrdziuban:
This helps with #9108 but doesn't solve it completely. My project that's taking 186 seconds for `definedTestDigests` is taking 55 seconds with these changes (~70% faster).

- Use `Builder`s to avoid building intermediate collections
- Use a `mutable.Set` for `alreadySeen`
- Use plain `Set` instead of `SortedSet`
  - Sorting only needs to happen at the end of the computation in `transitiveStamp`
  - This is now done with a call to `.toSeq.sorted`
- Avoids SHA-256 calls (from #9257)

New flamegraph: [flamegraph.html](https://github.com/user-attachments/files/28153064/flamegraph.html)

### DIFF
--- a/main/src/main/scala/sbt/internal/IncrementalTest.scala
+++ b/main/src/main/scala/sbt/internal/IncrementalTest.scala
@@ -23,7 +23,6 @@ import sbt.util.CacheImplicits
 import sbt.util.CacheImplicits.given
 import scala.collection.concurrent
 import scala.collection.mutable
-import scala.collection.SortedSet
 import xsbti.{ FileConverter, HashedVirtualFileRef, VirtualFileRef }
 
 object IncrementalTest:
@@ -175,8 +174,7 @@ class ClassStamper private[sbt] (
       converter,
     )
 
-  private val stamps = mutable.Map.empty[String, SortedSet[Digest]]
-  private val internalStamps = mutable.Map.empty[String, SortedSet[Digest]]
+  private val stamps = mutable.Map.empty[String, Set[Digest]]
   // Cached so by-name `analyses0` is only evaluated once
   private lazy val analyses = analyses0
   private val stampVf: VirtualFileRef => Digest =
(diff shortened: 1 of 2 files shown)


main/src/test/scala/sbt/internal/ClassStamperTest.scala
```diff
@@ -0,0 +1,364 @@
...
+        stamp(stamper(cyclesAnalysis), "A") ====
+          Some(
+            Digest("sha256-92475004e70f41b94750f4a77bf7b430551113b25d3d57169eadca5692bb043d/64")
+          ),
+        stamp(stamper(libraryDepsAnalysis), "A", Seq(Digest.dummy(42L))) ====
+          Some(
+            Digest("sha256-9d0a61172a43b1e7666f9527f82621395ef6a3a0ce5aed5dac317b2a76e8dd94/48")
+          ),
+      )
+    )
```
mrdziuban, line 95: I included these golden tests to prove that the output of `transitiveStamp` is unchanged after my changes in `IncrementalTest.scala` (they also passed before the changes), but I'm happy to remove them if you don't think they're worth keeping

main/src/main/scala/sbt/internal/IncrementalTest.scala
```diff
@@ -182,53 +190,65 @@ class ClassStamper(
...
       log: Logger,
   ): Option[Digest] =
-    val digests = SortedSet(analyses.flatMap(internalStamp(javaClassName, _, Set.empty, log))*)
-    log.debug(s"test: transitiveStamp($javaClassName, $extraHashes) = $digests")
-    if digests.nonEmpty then Some(Digest.sha256Hash(digests.toSeq ++ extraHashes*))
+    val builder = Set.newBuilder[Digest]
+    analyses.foreach: analysis =>
+      internalStamp(builder, javaClassName, analysis, Set.empty, log)
+    val digests = builder.result().toSeq.sorted
+    // log.debug(s"test: transitiveStamp($javaClassName, $extraHashes) = $digests")
```
mrdziuban: I commented out this log because it was causing a 4.5gb file to be created in my project (https://github.com/sbt/sbt/issues/9108#issuecomment-4498182122). It didn't contribute to the performance issues, but is just generally eating up resources

mrdziuban:
Would you mind taking a look at this @eed3si9n? It adds an additional 33% performance improvement (75 seconds down to 50) on top of #9257

eed3si9n:
I'm not against merging this PR, but I'd like to reproduce the perf issue somehow and also understand what make this faster. Is sorting the sets and combining them adding up to 25s difference?

mrdziuban:
> Is sorting the sets and combining them adding up to 25s difference?

Yeah that's my take on it -- the overhead of allocation (from using immutable collections) and sorting (from using `SortedSet`) is what this PR is targeting.

> I'd like to reproduce the perf issue somehow

Agreed, that would be great. I'm switching jobs in a week, so I'll unfortunately be losing access to the codebase where I can reproduce it, but happy to help if there's any way I can.

ybasket:
> Is sorting the sets and combining them adding up to 25s difference?

I'm experiencing the same problem of `definedTestDigests` being way too slow on an $work monorepo. The changes in this PR bring the time down significantly, from 163s to 105s on one module, from ~20min to 10min in another. Still some orders of magnitude from where I'd like them to be, but up to 50% better 👏 

> Agreed, that would be great. I'm switching jobs in a week, so I'll unfortunately be losing access to the codebase where I can reproduce it, but happy to help if there's any way I can.

As it's such an issue (blocking the upgrade to 2.0) for us, feel free to ping me (here or on Scala Discord) if any of you needs someone to test changes against a problematic codebase.

eed3si9n (approved):
Thanks!

PR #9254, opened 2026-05-20
[2.x] ci: Update dependabot setting

xuwei-k:


.github/dependabot.yml
```diff
@@ -1,5 +1,10 @@
 version: 2
+enable-beta-ecosystems: true
 updates:
+  - package-ecosystem: "sbt"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
eed3si9n, line 7: I'm mildly -1 on more bots bumping up deps for sbt/sbt since there's a harm in bumping too often without reviewing the upstream changes.

PR #9255, opened 2026-05-22
[2.x] Fix duplicate autoplugins packageBin mappings

### LINKED ISSUE #9252: sbt 2.0: duplicate entry sbt.autoplugins in packageBin on repeated publishLocal with server mode
## Summary

In a multi-module SbtPlugin project on sbt 2.0.0-RC13, `publishLocal` succeeds on the first run but fails on subsequent runs within the same sbt server session with:

```
[error] java.util.zip.ZipException: duplicate entry: sbt/sbt.autoplugins
[error] (scalac-options / Compile / packageBin) java.util.zip.ZipException: duplicate entry: sbt/sbt.autoplugins
```

## Steps to reproduce

Given a multi-module sbt plugin project with 3+ `SbtPlugin`-enabled subprojects:

```scala
// build.sbt
ThisBuild / organization := "com.example"
ThisBuild / version      := "0.1.0-SNAPSHOT"

val `scalac-options` = project.enablePlugins(SbtPlugin).settings(
  addSbtPlugin("org.typelevel" % "sbt-tpolecat" % "0.5.4"),
)
val components = project.enablePlugins(SbtPlugin)
val plugins    = project.dependsOn(components, `scalac-options`).enablePlugins(SbtPlugin)
```

Each subproject contains at least one `AutoPlugin` subclass.

```bash
# Ensure clean state
rm -rf target project/target ~/Library/Caches/sbt/v2

# First run — succeeds
sbt publishLocal

# Second run (connects to same sbt server) — fails
sbt publishLocal
```

**Important:** both commands must run in **server mode** (the default — no `--batch` flag), so the second invocation connects to the warm sbt server started by the first.

**Note on reproducibility:** We were unable to reliably reproduce this with a minimal sample project. The issue consistently reproduces in our larger project (8 SbtPlugin subprojects with `dependsOn` relationships and real `addSbtPlugin` dependencies), but a stripped-down copy with the same structure and source files does not trigger it. This suggests the race condition depends on timing factors such as the number of parallel tasks, compilation duration, or server-side cache population order. The issue appears to be a race condition — in the real project, it fails on the 2nd run in some sessions and on the 4th run in others, but once it starts failing it fails consistently until the cache is cleared.

## Observed behavior

After the first successful `publishLocal`, inspecting `scalac-options / Compile / packageBin / mappings` on the second run shows `sbt/sbt.autoplugins` coming from **two** locations:

```
(${OUT}/.../scalac-options/classes/sbt/sbt.autoplugins,          sbt/sbt.autoplugins)
(${OUT}/.../scalac-options/resource_managed/main/sbt/sbt.autoplugins, sbt/sbt.autoplugins)
```

The copy in `classes/` is **not present** after the first run — it appears only during the second run's task evaluation within the warm server. The `classes.sbtdir.zip` (disk cache) does not contain the file either, confirming this is an in-memory server-side issue.

## Expected behavior

`publishLocal` should be idempotent. Running it repeatedly against an unchanged project should always succeed.

## Analysis

`Compile / products` depends on `resources` (added in #8394 to fix #8391). During repeated evaluation in a warm sbt server session, this dependency chain appears to trigger `copyResources`, which copies the managed resource `sbt/sbt.autoplugins` from `resource_managed/main/` into the `classes/` output directory. `packageBin / mappings` then picks up the file from both directories, producing the duplicate entry.

Key observations:
- **Server mode required**: `--batch` mode (fresh JVM per invocation) does not reproduce reliably
- **First run always succeeds**: the duplicate only appears on the second+ server-mode invocation
- **Single-session works**: `sbt 'compile; publishLocal'` in one session succeeds; the issue is across separate client connections to the same server
- **Always `scalac-options`**: in our project, the subproject with an `addSbtPlugin` dependency is consistently the one that fails
- **Larger projects only**: we could not reproduce in a minimal sample — the race condition appears to require enough parallel task evaluation to trigger

## Environment

- sbt 2.0.0-RC13
- macOS Darwin 25.4.0
- Java 21 (Eclipse Adoptium 21.0.9)
- Scala 3.8.3

## Workaround

Deduplicate `packageBin / mappings` in `build.sbt`:

```scala
ThisBuild / Compile / packageBin / mappings := {
  (Compile / packageBin / mappings).value.distinctBy(_._2)
}
```

## Related

- #8391 — `2.0.0-RC7 publishes plugins without sbt.autoplugins`
- #8394 — fix that added `resources` dependency to `products`

anatoliykmetyuk:
Fixes #9252

Deduplicate executable jar package entries. In the issue, the duplicate was something like:

```scala
(resource_managed/.../sbt/sbt.autoplugins, "sbt/sbt.autoplugins")
(classes/.../sbt/sbt.autoplugins,          "sbt/sbt.autoplugins")
```

This is because `copyResources` copied autoplugin metadata to the classes directory.

Solution: when composing the `packageBinMappings`, differentiate between artifacts that come from class directory and from outside of it. If there is a duplicate, prefer resource that comes from outside of it.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1772,10 +1772,18 @@ object Defaults extends BuildCommon with DefExtra {
   def packageBinMappings: Initialize[Task[Seq[(HashedVirtualFileRef, String)]]] =
     Def.task {
       val converter = fileConverter.value
+      val classDirPath = classDirectory.value.toPath.toAbsolutePath.normalize
       val xs = products.value
-      xs
+      val allMappings = xs
         .flatMap(Path.allSubpaths)
-        .withFilter(_._1.isFile())
+        .filter(_._1.isFile())
+      val resourcePaths = allMappings.collect {
+        case (p, path) if !p.toPath.toAbsolutePath.normalize.startsWith(classDirPath) => path
+      }.toSet
+      allMappings
+        .filterNot { (p, path) =>
+          resourcePaths(path) && p.toPath.toAbsolutePath.normalize.startsWith(classDirPath)
+        }
         .map { (p, path) =>
           val vf = converter.toVirtualFile(p.toPath())
           (vf: HashedVirtualFileRef) -> path
--- a/sbt-app/src/sbt-test/plugins/duplicate-autoplugins-packagebin/build.sbt
+++ b/sbt-app/src/sbt-test/plugins/duplicate-autoplugins-packagebin/build.sbt
@@ -1,0 +1,8 @@
+ThisBuild / organization := "com.example"
+ThisBuild / version      := "0.1.0-SNAPSHOT"
+
+val `scalac-options` = project.enablePlugins(SbtPlugin).settings(
+  addSbtPlugin("org.typelevel" % "sbt-tpolecat" % "0.5.4"),
+)
+val components = project.enablePlugins(SbtPlugin)
+val plugins    = project.dependsOn(components, `scalac-options`).enablePlugins(SbtPlugin)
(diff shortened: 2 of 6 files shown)


anatoliykmetyuk:
CI red becuase develop is red, related to `3.8.3-RC1-wasm-bin-SNAPSHOT` missing from Maven.

eed3si9n (approved):
Thanks!

PR #9257, opened 2026-05-22
[2.x] perf: Improve incremental test stamper, take 2

eed3si9n:
Ref https://github.com/sbt/sbt/issues/9108
See also https://github.com/sbt/sbt/pull/9146

## Problem
Incremental stamper is still causing perf issue.

## Solution
This further avoids SHA-256 calls to calculate the stamp.

### DIFF
--- a/main/src/main/scala/sbt/internal/IncrementalTest.scala
+++ b/main/src/main/scala/sbt/internal/IncrementalTest.scala
@@ -182,10 +182,17 @@ class ClassStamper(
       extraHashes: Seq[Digest],
       log: Logger,
   ): Option[Digest] =
-    val digests = SortedSet(analyses.flatMap(internalStamp(javaClassName, _, Set.empty, log))*)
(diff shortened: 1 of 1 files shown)


mrdziuban:
@eed3si9n I just integrated these changes into my performance improvements in #9253, what do you think about closing this in favor of that?

Your changes alone cut the timing in my project from 186 down to 75 seconds, but with the additional changes in #9253 it goes down to 50 seconds.

eed3si9n:
I guess it's better than 20 min but 50s is still kind of long. Do you have a lot of JARs on the subprojects?

mrdziuban:
Agreed, it's way better but still slow. There are 298 jars on the classpath for this project's tests.

I also added a log line to see how many `digests` were computed and passed to `Digest.sha256Hash` -- one of the test classes computed 226 digests and the other computed 1118.

eed3si9n:
We will need to get to the bottom of this issue, but for now I'm going to land this one as-is.

PR #9261, opened 2026-05-22
[2.x] fix: Drop net.openhft farmNa, use MurmurHash3 in HashUtil

### LINKED ISSUE #8073: warning "A terminally deprecated method in sun.misc.Unsafe has been called"
## Steps

Use JDK 24 or JDK 25.

```bash
$ sbt --server --no-hide-jdk-warnings
```

## Problem

When I run sbt on Java 21, everything is wonderful. But with Java 24 I got a warning:
```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::arrayBaseOffset has been called by net.openhft.hashing.UnsafeAccess (~/.sbt/boot/scala-2.12.20/org.scala-sbt/sbt/1.10.11/zero-allocation-hashing-0.16.jar)
WARNING: Please consider reporting this to the maintainers of class net.openhft.hashing.UnsafeAccess
WARNING: sun.misc.Unsafe::arrayBaseOffset will be removed in a future release
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::load has been called by com.sun.jna.Native in an unnamed module (~/.sbt/boot/scala-2.12.20/org.scala-sbt/sbt/1.10.11/jna-5.12.0.jar)
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled
```

## Expectations

No warnings.

## Notes

I run it via Intellij Idea, but it seems that running from terminal with -java-home flag leads to the same.

BrianHotopp:
## Summary

`HashUtil.farmHash` and `BootServerSocket.socketLocation` called `LongHashFunction.farmNa()` from `zero-allocation-hashing`, whose impl uses `sun.misc.Unsafe` and triggers the JDK 23+ terminally-deprecated warning. Replaced with two `scala.util.hashing.MurmurHash3` calls composed into a 64-bit `Long`. Matches @eed3si9n's suggestion in the issue thread.

The `farmHash` method name and `"farm64-"` cache-string prefix are kept for source compatibility. Existing on-disk caches keyed on `farm64-<hex>` see a one-time cold rebuild after upgrade.

Fixes #8073.

## Test plan

- [x] `utilCache/test` (HasherTest goldens updated)
- [x] `commandProj/compile` (BootServerSocket.java now delegates to `HashUtil.farmHash`)

### DIFF
--- a/main-command/src/main/java/sbt/internal/BootServerSocket.java
+++ b/main-command/src/main/java/sbt/internal/BootServerSocket.java
@@ -28,7 +28,6 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
-import net.openhft.hashing.LongHashFunction;
 import org.scalasbt.ipcsocket.UnixDomainServerSocket;
 import org.scalasbt.ipcsocket.Win32NamedPipeServerSocket;
 import org.scalasbt.ipcsocket.Win32SecurityLevel;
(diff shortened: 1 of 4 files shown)


BrianHotopp:
Makes sense, thanks for benchmarking it @eed3si9n. Closing in favor of #9267.

PR #9264, opened 2026-05-24
[2.x] Add test for issue 2188

### LINKED ISSUE #2188: sbt.settingKey macro doesn't like higher-kinded types
```
scala> trait Foo[B[_]]
defined trait Foo

scala> val fo = sbt.settingKey[Foo[Option]]("asdf")
<console>:14: error: erroneous or inaccessible type
       val fo = sbt.settingKey[Foo[Option]]("asdf")
                                           ^
```

because https://issues.scala-lang.org/browse/SI-2915

> We force Manifest (or some kind of runtime type) because we do runtime key-string->type comparisons to ensure plugins will work together and not cause runtime type explosions

with workaround https://github.com/sbt/sbt/blob/0.13/main/src/main/scala/sbt/Keys.scala#L368-L370

> newtype it with a value class or something so the hkt isn't visible

(note, newtypes aren't allowed in .sbt files as of 0.13.8, so put it in `project/*.scala`)

xuwei-k:
- fix https://github.com/sbt/sbt/issues/2188

### DIFF
--- a/main-settings/src/test/scala/sbt/std/UsageTest.scala
+++ b/main-settings/src/test/scala/sbt/std/UsageTest.scala
@@ -168,4 +168,8 @@ object Assign {
   listTask := List(1)
   listTask += 1
   listTask += ak.value
+
+  trait HigherKind[B[_]]
+  // https://github.com/sbt/sbt/issues/2188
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #9265, opened 2026-05-24
[1.x] fix: Pump forked stdout/stderr through System.out/err in interactive mode

### LINKED ISSUE #9185: No output with `sbt --client` since sbt 1.12.2
## steps

1. Check out https://github.com/scala/scala3 (the Dotty/Scala 3 compiler project)
2. Set `project/build.properties` to sbt 1.12.2 or later (e.g. 1.12.10)
3. Run a forked task via client mode that produces stdout/stderr:

```bash
sbt --client "scalac tests/neg/i8569.scala"
```

The `scalac` task delegates to `(scala3-compiler-nonbootstrapped / Compile / runMain).toTask(...)`.
The build has `run / fork := true`, `run / connectInput := true`, and `outputStrategy := Some(StdoutOutput)`.

## problem

In `sbt --client` mode, the forked process stdout/stderr is **not relayed to the terminal** — only sbt's own messages appear:

```
[info] running (fork) dotty.tools.dotc.Main -d ... -classpath ... tests/neg/i8569.scala
[error] nonzero exit code returned from runner: 1
```

The compiler diagnostic text (which goes to stdout/stderr) is missing.

The same command without `--client` works correctly:

```bash
sbt "scalac tests/neg/i8569.scala"
```

## expectation

`sbt --client` should show the same stdout/stderr output as the non-client (batch) mode. The compiler diagnostics should be visible.

## notes

- **First broken version: 1.12.2**. Bisecting sbt versions shows:
  - ✔ 1.12.0, 1.12.1 — output is visible in client mode
  - ✘ 1.12.2 through 1.12.10 — output is invisible in client mode
- **Suspected cause**: PR [#8678](https://github.com/sbt/sbt/pull/8678) "Use JProcess for interactive forking" (first released in 1.12.2), which switches to `java.lang.ProcessBuilder` when `connectInput = true`. The JProcess path appears to lose stdout/stderr forwarding in client mode.
- **Workaround**: disabling `connectInput` restores the output in client mode:
  ```bash
  sbt 'set `scala3-compiler-nonbootstrapped` / Compile / runMain / connectInput := false; scalac tests/neg/i8569.scala'
  ```
- ~~**Minimal reproduction**: We tried to create a minimal standalone reproducer with the same settings (`run / fork := true`, `run / connectInput := true`, `outputStrategy := Some(StdoutOutput)`, a root aggregate project delegating to a subproject's `runMain` via `.toTask(...)`), but the bug **only reproduced in the scala/scala3 project itself**, not in our simplified reproduction. This suggests the issue may depend on specific combinations of build size, fork options, or project structure that are only triggered by the actual Dotty build.~~
- Environment: Ubuntu Linux, Java 17 (OpenJDK), sbt runner 1.12.0.

*(Investigation and writing by https://huggingface.co/moonshotai/Kimi-K2.6)*

BrianHotopp:
## Summary

Since #8678 (sbt 1.12.2), `Fork.apply` has routed the `connectInput && outputStrategy == StdoutOutput` case through a new JProcess path that calls `jpb.inheritIO()`. `inheritIO` ties the child JVM's stdio to the sbt server JVM's OS-level fd 1/2, bypassing the `Terminal.withOut` `System.out`/`System.err` proxies. In server mode (which `sbt --client` uses), those proxies are how task output reaches the client terminal — so the child's stdout/stderr go nowhere observable to the client.

This forced scala/scala3 CI to downgrade from sbt 1.12.10 to 1.12.1 (scala/scala3#25995). mbovel's diagnosis in #9185 is the definitive analysis.

## Fix

Keep `redirectInput(Redirect.INHERIT)` so REPL/raw-mode keystrokes still reach the child (preserves what #8678 was for), but leave stdout/stderr as `PIPE` and pump them through `System.out`/`System.err` in two daemon threads. The proxies pick them up and relay to the active terminal.

Stdin direction is unchanged. The only thing the child loses vs `inheritIO` is direct ownership of fd 1/2 — programs that call `isatty(1)` will now see "not a tty". That was already the case for every non-interactive forked run before #8678 (via `scala.sys.process.Process`'s PIPE-backed stdio), so this isn't a new regression.

Fixes #9185.

## Test plan

- [x] New `ClientTest` case `forked run with connectInput relays stdout to --client`. Asserts `STDOUT_MARKER_9185` reaches the in-process `--client` stdout when running a forked main with `connectInput := true` + `outputStrategy := Some(StdoutOutput)`.
- [x] Confirmed the test **fails** on this branch *without* the `Fork.scala` change (client receives only sbt's `running (fork) hello` line) and **passes** with it.
- [x] All 11 `ClientTest` cases pass (10 pre-existing + 1 new).
- [x] `runProj/scalafmtCheckAll`, `serverTestProj/scalafmtCheckAll`, `runProj/mimaReportBinaryIssues` green.

### DIFF
--- a/run/src/main/scala/sbt/Fork.scala
+++ b/run/src/main/scala/sbt/Fork.scala
@@ -8,12 +8,11 @@
 
 package sbt
 
-import java.io.File
-import java.io.PrintWriter
+import java.io.{ File, IOException, InputStream, OutputStream, PrintWriter }
 import java.lang.ProcessBuilder.Redirect
 import scala.sys.process.Process
 import OutputStrategy._
-import sbt.internal.util.{ RunningProcesses, Util }
+import sbt.internal.util.{ RunningProcesses, Terminal, Util }
 import Util.{ AnyOps, none }
 
 import java.lang.{ ProcessBuilder => JProcessBuilder, Process => JProcess }
@@ -214,14 +213,24 @@ object Fork {
     val environment: List[(String, String)] = env.toList ++ extraEnv
     workingDirectory.foreach(jpb.directory(_))
     environment.foreach { case (k, v) => jpb.environment.put(k, v) }
-    jpb.inheritIO()
+    // Inherit stdin for REPL/raw-mode keystrokes. Stdout/stderr stay PIPE so
+    // `blockJForExitCode` can pump them to the active terminal directly.
+    jpb.redirectInput(Redirect.INHERIT)
     jpb.start()
   }
 
   private[sbt] def blockJForExitCode(p: JProcess): Int = {
     RunningProcesses.add(p)
     try {
+      // Capture the active terminal's streams once on the task thread. The
+      // System.out/err proxy resolves `activeTerminal` at every write, so a
+      // concurrent terminal swap can route mid-pump bytes to the wrong client.
+      val active = Terminal.current
+      val outT = pumpStream(p.getInputStream, active.outputStream, "sbt-fork-stdout")
+      val errT = pumpStream(p.getErrorStream, active.errorStream, "sbt-fork-stderr")
       p.waitFor()
+      outT.join()
+      errT.join()
       p.exitValue()
     } finally {
       if (p.isAlive()) p.destroy()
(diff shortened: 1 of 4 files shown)


eed3si9n (approved):
Thanks!

Issue #5627, opened 2020-06-16
Progress indicators per task for super shell


oyvindberg:
Hey there! :)

### Proposal
I think the super shell was a great addition to sbt. Now I want to see if there would be support for extending the functionality a bit further.

Currently all running tasks are displayed along with how long they have been running. 
I propose we add the ability for (typically slow running) tasks to display some textual information which would indicate progress.

I hacked up the functionality locally, see [this video](https://olvind.com/videos/super_progress2.mp4)
I'm creating an issue instead of a PR because my first attempt isn't production ready (though I'll gladly share code if wanted), and a proper solution seems to cross a few layers from public API to the task running engine itself.

### Context

I arrived at this as a possible solution to a performance problem in sbt with coursier, see https://github.com/coursier/coursier/issues/1720 . TLDR: it blocks threads to render a progress bar.
I'm also developing another [plugin](https://scalablytyped.org) where I needed the exact same functionality and wished it was available in sbt.

### Further ideas
IntelliJ has this nice thing where it shows compiler phases and files while compiling. This is useful both for the user to feel progress, and so the user gets an idea of what takes time. We could do the same in sbt, or something easier like counting warnings, render a progress bar or what have you. Same with tests, we could have some progress, name of running test, etc

eed3si9n:
@oyvindberg Thanks for the proposal. I'm open to the idea of tasks being able to push some text status. I'd be happy to hear other contributors or users thoughts on this too.

oyvindberg:
Cool @eed3si9n .

I'll put a bit more work into this on Sunday or Monday so we have some code to talk about and try to prepare a demo other than a video :)

BrianHotopp:
Following up from #5508: I opened #9270, which narrows the lm-coursier resolution lock so it's only held while coursier is actually rendering its interactive progress bar. That parallelizes `update` across modules in the non-interactive cases (IntelliJ/BSP, CI, any non-TTY), which is where the worst reports come from.

The interactive super-shell case is the part this issue is really about, and with #9270 in place it now looks fairly contained: supply a thread-safe sbt-side `CacheLogger` under super shell (which both suppresses coursier's own bar and unlocks parallel resolution), make the one residual lock in `UpdateRun` logger-aware, and render progress at the task level the way you've described.

[...]

PR #9270, opened 2026-05-28
[2.x] fix: Parallelize dependency resolution when no progress bar is rendered

### LINKED ISSUE #5508: `update` gets slower with the number of modules in project

BrianHotopp:
## Problem

`update` (dependency resolution) gets dramatically slower as the number of modules grows — wall-clock time scales with the *module count* rather than the number of distinct artifacts. The issue thread reports IntelliJ re-imports taking ~4 minutes and CI `update` taking 17–20 minutes on large multi-module builds. The long-known workaround, `COURSIER_PROGRESS=false`, cuts time 2–3× (see #5508).

## Root cause

`lmcoursier.internal.Lock.maybeSynchronized` wraps resolution and artifact fetching in a **process-global** monitor, so per-module `update` tasks that sbt would otherwise run in parallel are serialized. The guard was:

```scala
needsLock = params.loggerOpt.nonEmpty || !RefreshLogger.defaultFallbackMode
```

That lock's only purpose is to serialize coursier's interactive progress bar (`ProgressBarRefreshDisplay`), which lm-coursier creates **only** when no custom logger is supplied *and* coursier is not in fallback mode. The `loggerOpt.nonEmpty` clause therefore over-serialized the common non-interactive case: under IntelliJ, CI, and any non-TTY run, `useSuperShell` is off, so sbt supplies a quiet debug-only `CoursierLogger` (`loggerOpt.nonEmpty`) — no progress bar is rendered, yet the global lock was still taken. `COURSIER_PROGRESS=false` only relieved the interactive (no-logger) path, which is why parallelism never became the default.

## Fix

Narrow the predicate to reflect when a progress bar is actually rendered:

```scala
def progressBarActive(hasCustomLogger: Boolean, fallbackMode: Boolean): Boolean =
  !hasCustomLogger && !fallbackMode
```

Used in `ResolutionRun` and `ArtifactsRun`. Resolution and artifact fetching now run in parallel across modules unless coursier is genuinely drawing its interactive progress bar.

## Scope / what this does *not* do

Interactive terminals with live progress bars stay serialized, because `ProgressBarRefreshDisplay` is not safe to drive from multiple modules at once. Users wanting parallelism there keep `COURSIER_PROGRESS=false`. Fully parallel resolution *with* live per-module progress bars (rendering progress at the sbt task level) remains tracked by #5627.

## Why it's safe

- `SbtCoursierCache` is entirely `ConcurrentHashMap`-backed.
- A `CacheLogger` is already invoked concurrently from within a single resolution (parallel downloads), so loggers must be thread-safe regardless — cross-module concurrency imposes no new requirement.
- The `COURSIER_PROGRESS=false` parallel path has been exercised in production for years.
- In the test/CI context the locking decision is unchanged (no logger + fallback mode → no lock, before and after), so existing lm-coursier specs behave identically.

## Testing

- New `lmcoursier.internal.LockSpec` (Verify) pins `progressBarActive` across all four logger/fallback combinations.
- The behavioral effect (parallel vs. serial) is timing-based and too flaky to assert in CI, so the deterministic locking predicate is pinned instead; the issue thread documents the resulting 2–3× speedup via the equivalent `COURSIER_PROGRESS=false` path.

## Release note

Added `notes/2.0.0/parallel-dependency-resolution.md`.

Fixes #5508.

---

*Implemented with AI assistance (Claude Code); reviewed, formatted, and verified locally by me (scalafmt, compile, `LockSpec`, MiMa).*

### DIFF
--- a/lm-coursier/src/main/scala/lmcoursier/internal/ArtifactsRun.scala
+++ b/lm-coursier/src/main/scala/lmcoursier/internal/ArtifactsRun.scala
@@ -41,7 +41,10 @@ object ArtifactsRun {
     }
 
     Lock.maybeSynchronized(needsLock =
-      params.loggerOpt.nonEmpty || !RefreshLogger.defaultFallbackMode
+      Lock.progressBarActive(
+        hasCustomLogger = params.loggerOpt.nonEmpty,
+        fallbackMode = RefreshLogger.defaultFallbackMode
+      )
     ) {
       result(params, coursierLogger)
     }
--- a/lm-coursier/src/main/scala/lmcoursier/internal/Lock.scala
+++ b/lm-coursier/src/main/scala/lmcoursier/internal/Lock.scala
@@ -3,6 +3,14 @@ package lmcoursier.internal
 private[lmcoursier] object Lock {
   private val lock = new Object
 
+  /* The lock guards coursier's interactive progress bar (ProgressBarRefreshDisplay), the only thing
+   * here that cannot be driven from more than one module at a time. coursier renders it only when no
+   * custom cache logger is supplied and it is not in fallback mode; otherwise (a custom logger, or
+   * the line-based fallback display) modules can be resolved in parallel. A CacheLogger is already
+   * invoked concurrently within a single resolution, so it must be thread-safe regardless. */
+  def progressBarActive(hasCustomLogger: Boolean, fallbackMode: Boolean): Boolean =
+    !hasCustomLogger && !fallbackMode
+
   /* Progress bars require us to only work on one module at the time. Without those we can go faster */
   def maybeSynchronized[T](needsLock: Boolean)(f: => T): T =
     if (needsLock) lock.synchronized(f)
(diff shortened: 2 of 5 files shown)


eed3si9n (approved):
Thanks!

PR #9272, opened 2026-05-29
[2.x] Opt bspBuildTargetOutputPathsItem out of caching

### LINKED ISSUE #9268: (potentially) caching issue with `bspBuildTargetOutputPathsItem` task
I do not have exact steps to reproduce the problem from scratch, but I have a video showing how the problem looks when it is already present on my machine.

The steps in the video are:

1. I have a project in `~/IdeaProjects/toRemove/untitled34` with sbt `1.12.11`. I start sbt.
2. `show bspBuildTargetOutputPathsItem` returns the correct output paths for the targets in this project.
3. I change the sbt version to `2.0.0-RC12` and start sbt again, restarting the server.
4. Problem: `show bspBuildTargetOutputPathsItem` returns some output paths from another project. As a note, I didn't have that project open at that moment (it probably doesn't matter).
5. I clear the sbt caches in `~/Library/Caches/sbt/v2`.
6. I run `show bspBuildTargetOutputPathsItem` in the project again, and now it returns the correct output paths for the targets in this project.

Here is the video - https://drive.google.com/file/d/11PKLgKSjF41KmrxaZ0ins1cRVBikZLTI/view?usp=sharing

I discovered this problem while working with an sbt/BSP project, because `bspBuildTargetOutputPathsItem` is used when generating a response to the `buildTarget/outputPaths` build server request.

anatoliykmetyuk:
Fixes #9268

### DIFF
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -487,6 +487,7 @@ object Keys {
   val bspBuildTargetDependencyModules = inputKey[Unit]("").withRank(DTask)
   val bspBuildTargetDependencyModulesItem = taskKey[DependencyModulesItem]("").withRank(DTask)
   val bspBuildTargetOutputPaths = inputKey[Unit]("").withRank(DTask)
+  @transient
   val bspBuildTargetOutputPathsItem = taskKey[OutputPathsItem]("").withRank(DTask)
   val bspBuildTargetCompile = inputKey[Unit]("").withRank(DTask)
   val bspBuildTargetCompileItem = taskKey[Int]("").withRank(DTask)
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -322,7 +322,7 @@ object BuildServerProtocol {
     },
     bspBuildTargetDependencySourcesItem := dependencySourcesItemTask.value,
     bspBuildTargetDependencyModulesItem := dependencyModulesItemTask.value,
-    bspBuildTargetOutputPathsItem := {
+    bspBuildTargetOutputPathsItem := Def.uncached {
```
eed3si9n: Does this task capture absolute paths? If so should we tag it to be transient?

anatoliykmetyuk: Yep, that's right - fixed it

eed3si9n (approved):
Thanks!

PR #9118, opened 2026-04-22
[2.x] fix: Include platform suffix in Artifact.artifactName

anatoliykmetyuk:
Fixes https://github.com/sbt/sbt/issues/9117

PR #9293, opened 2026-06-01
[2.x] fix: Fixes publishing platform-specific artifacts (like Scala Native)

### LINKED ISSUE #9117: [2.x] Platform suffix is absent in published artifacts
## steps

**build.sbt**
```
enablePlugins(ScalaNativePlugin)
scalaVersion := "3.8.3"
```

**project/plugins.sbt**
```
addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.5.11")
```

```
> sbt --sbt-version 1.12.8 'show packagedArtifacts' | grep "Wrote"
[info] Wrote /Users/velvetbaldmime/.local/share/sniper/data/178/target/scala-3.8.3/root-178_native0.5_3-0.1.0-SNAPSHOT.pom

> sbt --sbt-version 2.0.0-RC12 'show packagedArtifacts' | grep "Wrote"
[info] Wrote /Users/velvetbaldmime/.local/share/sniper/data/178/target/out/native0.5/scala-3.8.3/root-178/root-178_3-0.1.0-SNAPSHOT.pom
```

## problem

Notice how `native0.5` is not present in the published pom – I believe it _has_ to be there because of maven conventions.

This causes [publishing issues I'm seeing](https://github.com/indoorvivants/sniper/actions/runs/24662262325/job/72111298192#step:7:145):

> [error] (publish) java.io.IOException: PUT operation to URL https://central.sonatype.com/repository/maven-snapshots/com/indoorvivants/sniper_native0.5_3/0.0.9+2-330cf49c-SNAPSHOT/sniper_3-0.0.9+2-330cf49c-SNAPSHOT.pom failed with status code 400:  Invalid mavenPath for a Maven 2 repository

## expectation

Filenames should be the same across SBT versions, as it's mandated by maven.

arashi01:
Resolves #9117

**Problem**
Platform suffix is absent in published artifacts.
#9118 partially solved this. It didn't touch publish-side naming though, so the published files still drop the suffix in three places.

**Solution**
This fixes
- ivyless publication names - CoursierArtifactsTasks.coursierPublicationsTask
- ivyless POM <artifactId> (the coordinate written inside the .pom) - PomGenerator.crossVersionDep
- Ivy backend (useIvy := true) - CrossVersion.substituteCross via IvyActions

### DIFF
--- a/lm-core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
+++ b/lm-core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
@@ -168,17 +168,17 @@ private[librarymanagement] abstract class CrossVersionFunctions {
 
   /**
    * Appends the platform suffix (e.g. `native0.5`, `sjs1`) to `name`, preferring an explicit
-   * `platformOpt` over the `projectPlatform`. `""` and `jvm` add no suffix.
+   * `platformOpt` over the `projectPlatform`. `""` and `jvm` add no suffix. Keep in sync with
+   * `lmcoursier.FromSbt.addPlatformSuffix` (until lm-coursier moves under sbt).
    */
-  def addPlatformSuffix(
+  private[sbt] def addPlatformSuffix(
       name: String,
       platformOpt: Option[String],
       projectPlatform: Option[String]
   ): String =
-    (platformOpt orElse projectPlatform) match {
+    platformOpt.orElse(projectPlatform) match
       case Some(p) if p.nonEmpty && p != Platform.jvm => crossName(name, p)
       case _                                          => name
-    }
 
   /** Cross-versions `exclude` according to its `crossVersion`. */
   private[sbt] def substituteCross(
--- a/lm-core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
+++ b/lm-core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
@@ -166,6 +166,20 @@ private[librarymanagement] abstract class CrossVersionFunctions {
   private[sbt] def crossName(name: String, cross: String): String =
     name + "_" + cross
 
+  /**
+   * Appends the platform suffix (e.g. `native0.5`, `sjs1`) to `name`, preferring an explicit
+   * `platformOpt` over the `projectPlatform`. `""` and `jvm` add no suffix.
+   */
+  def addPlatformSuffix(
+      name: String,
+      platformOpt: Option[String],
+      projectPlatform: Option[String]
+  ): String =
+    (platformOpt orElse projectPlatform) match {
+      case Some(p) if p.nonEmpty && p != Platform.jvm => crossName(name, p)
+      case _                                          => name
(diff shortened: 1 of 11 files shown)


lm-core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
```diff
@@ -166,6 +166,20 @@ private[librarymanagement] abstract class CrossVersionFunctions {
   private[sbt] def crossName(name: String, cross: String): String =
     name + "_" + cross
 
+  /**
+   * Appends the platform suffix (e.g. `native0.5`, `sjs1`) to `name`, preferring an explicit
+   * `platformOpt` over the `projectPlatform`. `""` and `jvm` add no suffix.
+   */
+  def addPlatformSuffix(
```
eed3si9n:
```suggestion
  private[sbt] def addPlatformSuffix(
```

Could you hide this behind private plz?

lm-core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
```diff
@@ -166,6 +166,20 @@ private[librarymanagement] abstract class CrossVersionFunctions {
...
+      name: String,
+      platformOpt: Option[String],
+      projectPlatform: Option[String]
+  ): String =
+    (platformOpt orElse projectPlatform) match {
```
eed3si9n: I suggest:
1. Avoiding infix here platformOpt.orElse(projectPlatform)
2. Using bracelss on match

https://github.com/sbt/sbt/blob/develop/contributing-docs/03_coding_style.md#infix-notation

eed3si9n (commented):
Thanks for the contribution!

lm-core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
arashi01: Only reason `addPlatformSuffix` wasn't `private[sbt]` is because it needed to be callable from `lmcoursier.FromSbt.sbtCrossName`. If you prefer, I can make this private and keep the duplicate in `FromSbt`?

lm-core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
arashi01: Sorry, should've read the docs in more detail! Will fix.

lm-core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
eed3si9n: Yea, maybe we should move the package for lmcoursier in another PR. For now, dupelicate is ok with comment pointing at each other.

arashi01: Done!

eed3si9n (approved):
Thanks!

PR #9295, opened 2026-06-02
[2.x] feat: Resolve dependencies in parallel under the super shell

BrianHotopp:
## What

`update` now resolves dependencies **in parallel under the interactive super shell**, rendering one aggregate progress line instead of coursier's per-module bars. First step of #5627; builds on #9270.

## How

- `LMCoursier.coursierLoggerTask` now supplies a `CacheLogger` under the super shell instead of `None`. Supplying any logger (a) suppresses coursier's own per-module bar and (b) flips #9270's `progressBarActive` predicate to `false`, so resolution runs in parallel.
- Progress is a **per-command instance, not a global**: a `ResolutionProgress` is created in `MainLoop.next` alongside `Keys.taskProgress`, stored under an `AttributeKey` with the same lifecycle. `coursierLoggerTask` reads it from `State` and hands it to the `ResolutionProgressLogger` that coursier's download threads call; `TaskProgress` holds the same instance and appends one line at render time, e.g. `downloading 240 files, 31.0 MiB 12s` — distinct non-checksum urls (coursier reports logger sessions per *configuration*, not per module, so a module count isn't knowable here) plus a live burst elapsed. The instance is born empty per command and discarded with it — nothing leaks across commands, and the count never resets mid-`update`. Per-url byte accounting uses an atomic `compute` so concurrent download callbacks can't double-count.

## The call I made (redirect me if you'd rather)

v1 is a single aggregate line because it needs **no schema change**. Per-module status lines would require a `status` field on `ProgressItem` (and touch the logging codecs / BSP), so that's a follow-up. Happy to go straight to per-module if you'd prefer.

## One thing I want your call on

A single `update` does a resolve phase and a separate artifacts phase, so the per-command instance **accumulates across both** rather than resetting between them. Confirm that aggregate-across-phases is the behavior you want.

## Notes

- **`UpdateRun` lock deferred:** it only builds the report (no network), so the parallelism win is in `ResolutionRun`/`ArtifactsRun` (already unlocked by #9270). Follow-up.
- **Behavior change:** under the super shell, coursier's per-module bars are replaced by the aggregate line; `csrLogger := Some(...)` opts out (your logger is used, coursier unchanged).

## Testing

- Unit test (`ResolutionProgressSpec`, Verify): aggregation, monotonic bytes, counts-persist-across-phases, per-configuration-session dedup, and checksum-companion exclusion — each on a fresh instance.
- Parallelism follows from #9270's `progressBarActive` predicate (covered by `LockSpec`): supplying a logger ⇒ no lock.
- The super-shell *rendering* needs a real TTY, so it isn't deterministically CI-testable; the wiring was verified by tracing the `MainLoop` → `EvaluateTask` (`Keys.state`) → `coursierLoggerTask` path (the same threading that makes `streams.value` work in that task).

## Release note

`notes/2.0.0/parallel-resolution-supershell.md`.

re: #5627

---

*Implemented with AI assistance (Claude Code); reviewed and verified locally by me (scalafmt, compile, `ResolutionProgressSpec`, MiMa).*

### DIFF
--- a/main/src/main/scala/sbt/coursierint/ResolutionProgress.scala
+++ b/main/src/main/scala/sbt/coursierint/ResolutionProgress.scala
@@ -82,7 +82,7 @@ private[sbt] final class ResolutionProgress {
       val mib = bytes.get().toDouble / (1024.0 * 1024.0)
       val label = if (n == 1) "file" else "files"
       val elapsedMicros = math.max(0L, (System.nanoTime() - burstStartNanos.get()) / 1000L)
-      Some((f"Updating $n $label, $mib%.1f MiB", elapsedMicros))
+      Some((f"downloading $n $label, $mib%.1f MiB", elapsedMicros))
     }
 }
 
--- a/main/src/main/scala/sbt/coursierint/ResolutionProgress.scala
+++ b/main/src/main/scala/sbt/coursierint/ResolutionProgress.scala
@@ -18,21 +18,31 @@ import lmcoursier.definitions.CacheLogger
  *
  * One instance is created per command in `MainLoop.next` (held under `Keys.resolutionProgress`, the
  * same lifecycle as `Keys.taskProgress`). It is fed by [[ResolutionProgressLogger]] from coursier's
- * download-pool threads and read by `TaskProgress` to render a single super-shell line. Because
- * those callbacks run in parallel across modules, every field is atomic and byte accounting uses a
- * monotonic per-url delta and so can never go backwards. There is no cross-command reset: the
- * instance is born empty and discarded with the command.
+ * download-pool threads and read by `TaskProgress` to render a single super-shell line.
+ *
+ * The counting model matches how coursier actually drives a `CacheLogger`:
+ *   - `init`/`stop` arrive once per logger session — one per configuration resolution and one per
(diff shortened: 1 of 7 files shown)


eed3si9n:
Thanks! I just tested it locally. Looks very cool.

main/src/main/scala/sbt/coursierint/ResolutionProgress.scala
```diff
@@ -0,0 +1,112 @@
...
+   * counter). Else None (the line disappears).
+   */
+  def snapshot(): Option[(String, Long)] =
+    if (inFlight.get() <= 0) None
+    else {
+      val n = files.size()
+      val mib = bytes.get().toDouble / (1024.0 * 1024.0)
+      val label = if (n == 1) "file" else "files"
+      val elapsedMicros = math.max(0L, (System.nanoTime() - burstStartNanos.get()) / 1000L)
+      Some((f"Updating $n $label, $mib%.1f MiB", elapsedMicros))
```
eed3si9n: I prefer all lower case here. Also the verb should be "downloading"?

BrianHotopp: Done in b86e5a6.

eed3si9n (approved):
Thanks!

PR #9296, opened 2026-06-02
[2.x] Update Scala to 3.8.4

xuwei-k:


xuwei-k:
https://github.com/sbt/io/blob/0562307abbb5489bef64338ce53f11529f1dcb4e/io/src/main/scala/sbt/internal/nio/TimeSource.scala#L22-L25

`TimeSource` is package private

```
[warn] -- Deprecation Warning: /home/runner/work/sbt/sbt/main/src/main/scala/sbt/internal/Continuous.scala:489:7 
[warn] 489 |      )
[warn]     |       ^
[warn]     |Usage of implicit object default defined in object TimeSource, which is not accessible here. In Scala 3.10, this implicit will no longer be found.
[warn] -- Deprecation Warning: /home/runner/work/sbt/sbt/main/src/main/scala/sbt/internal/Continuous.scala:532:5 
[warn] 532 |    )
[warn]     |     ^
[warn]     |Usage of implicit object default defined in object TimeSource, which is not accessible here. In Scala 3.10, this implicit will no longer be found.
Warning: mocking up superclass for module class std
[error] No warnings can be incurred under -Werror
```

- https://github.com/scala/scala3/pull/25608
- https://github.com/scala/scala3/pull/25516

xuwei-k:
- https://github.com/sbt/io/pull/522

eed3si9n:
Thanks for getting this started. https://github.com/sbt/sbt/pull/9302 is now green, so I'll land that one instead.

PR #9314, opened 2026-06-08
[2.x] Remove type ascription in Continuous

xuwei-k:
- https://github.com/scala/scala3/pull/25597

prepare for Scala 3.9


eed3si9n (approved):
Thanks!

PR #9337, opened 2026-06-14
[2.x] fix: Probe for a live server before refusing to start

### LINKED ISSUE #6777: sbt thinks that server is already booting because of this exception

BrianHotopp:
## Problem

`sbt` refuses to start with a misleading error when the boot io socket cannot be
created (#6777, top-voted open bug):

    sbt thinks that server is already booting because of this exception:
    sbt.internal.ServerAlreadyBootingException: java.io.IOException: Could not create lock ...

Any `IOException` from creating the socket is wrapped as
`ServerAlreadyBootingException` and reported this way, and non-interactive
invocations exit with code 2. But a creation failure is rarely another sbt
booting: a stale socket reclaim race, permission or path-length problems with
`XDG_RUNTIME_DIR` or the temp directory, or a Windows named-pipe access error all
land here and block sbt entirely. Separately, `IOException` from the constructor
creating the socket *directory* was not caught at all and crashed startup.

## Fix

`xMain.getSocketOrExit` now connects to the socket to check whether a server is
actually listening before believing the exception:

- **Live server answers the probe** -> genuinely another sbt booting in this
  build: behavior is unchanged (interactive y/n prompt; non-interactive exit),
  with a concise actionable message instead of a stack trace.
- **Nothing answers** (or a raw constructor `IOException`) -> the failure is
  environmental; sbt continues without the boot socket, silently since nothing
  is actionable. This mirrors the grace sbt already extends on
  `UnsatisfiedLinkError`; the boot socket only forwards boot-time io to
  early-connecting clients.

The probe is extracted to `BootServerSocketProbe` so the test exercises the
production code path. It catches `LinkageError` as well as `NonFatal` (the
connect may perform the JVM's first JNI/JNA load, and an error escaping the
pattern guard would replace the original exception), and it runs on a daemon
thread bounded by a short timeout: the underlying native connect has no timeout
and would block indefinitely against a bound socket whose listen backlog is
saturated, so a hang-avoidance change must not itself be able to hang startup.

While here, `BootServerSocket`'s reader loop now deregisters a disconnected
client via `close()` the same way the write path already does, so a client (or
this probe) disconnecting during boot can't leave a dead entry that suppresses
the `NO_BOOT_CLIENTS_CONNECTED` signal. (A probe against a genuinely booting
server thus leaves at most a transient connection, reaped on that server's next
boot-input cycle.)

## Design notes

- `NetworkClient.waitForServer` already uses this exact connect as its liveness
  signal, so the probe is established practice, not a new mechanism.
- The probe can report a false negative in two narrow races (Windows pipe busy
  between accept instances; the booting server finishing and closing its boot
  socket between the bind failure and the probe). Both then degrade to
  "proceed without the socket", which is the same outcome `-Dsbt.server.forcestart=true`
  and the interactive prompt already allow, and the server-start portfile guard
  still arbitrates genuine duplicates. Failing open beats the old false-positive
  refusal for CI. On Linux the probe may be skipped-flaky (per the existing
  comment in `NetworkClient`), but a live server there almost never produces
  `ServerAlreadyBootingException` in the first place (the Unix path deletes and
  rebinds the socket).

## Testing

- New Verify suite `BootServerSocketSpec` (4 tests) pins the probe: live server
  detected, nothing-listening, after-close, and stale-socket-file reclaimed.
- `mainProj/compile`, `commandProj/Test`, `scalafmtCheck`, and
  `mimaReportBinaryIssues` on both modules green locally.

## Note

This targets 2.x; #6777 was originally filed on 1.x and the same misreport
exists there. Happy to port it to 1.x if you'd like.

Fixes #6777

---

*Implemented with AI assistance (Claude Code); reviewed and verified locally by
me (compile, unit tests, scalafmt, MiMa).*

### DIFF
--- a/main/src/main/scala/sbt/Main.scala
+++ b/main/src/main/scala/sbt/Main.scala
@@ -152,29 +152,38 @@ private[sbt] object xMain:
   private def getSocketOrExit(
       configuration: xsbti.AppConfiguration
   ): (Option[BootServerSocket], Option[Exit]) = {
-    def printThrowable(e: Throwable): Unit = {
-      println("sbt thinks that server is already booting because of this exception:")
-      e.printStackTrace()
-    }
-
-    val target =
-      configuration.baseDirectory().toPath().toRealPath().resolve("project").resolve("target")
-    val hash = HashUtil.farmHash(target.toString().getBytes("UTF-8"));
+    val base = configuration.baseDirectory().toPath().toRealPath()
+    val target = base.resolve("project").resolve("target")
+    val hash = HashUtil.farmHash(target.toString().getBytes("UTF-8"))
+    // sbt runs fine without a boot socket (see the UnsatisfiedLinkError case below), so a
+    // creation failure only means "another sbt is booting" if something answers on it.
+    def liveServerDetected: Boolean =
+      BootServerSocketProbe.liveServerDetected(
+        BootServerSocket.socketLocation(base, hash),
+        BootServerSocket.requiresJNI() || SysProp.serverUseJni,
+      )
     try Some(new BootServerSocket(configuration, hash)) -> None
     catch {
-      case e: ServerAlreadyBootingException if hasConsole && !ITerminal.startedByRemoteClient =>
-        printThrowable(e)
+      // No live server and nothing the user can do about a socket failure, so proceed
+      // silently without the boot socket, as for UnsatisfiedLinkError below.
+      case _: ServerAlreadyBootingException if !liveServerDetected => (None, None)
+      case _: ServerAlreadyBootingException if hasConsole && !ITerminal.startedByRemoteClient =>
+        println("another sbt appears to be booting in this build.")
         println("Create a new server? y/n (default y)")
         val exit =
           if (ITerminal.get.withRawInput(System.in.read) == 'n'.toInt) Some(Exit(1))
           else None
         (None, exit)
-      case e: ServerAlreadyBootingException =>
+      case _: ServerAlreadyBootingException =>
         if (SysProp.forceServerStart) (None, None)
         else {
-          printThrowable(e)
+          println("another sbt appears to be booting in this build; exiting.")
+          println(
+            "wait for it to finish, attach to it with --client, or pass -Dsbt.server.forcestart=true to start anyway."
+          )
           (None, Some(Exit(2)))
         }
+      case _: IOException          => (None, None)
       case _: UnsatisfiedLinkError => (None, None)
     }
   }
--- a/main-command/src/main/java/sbt/internal/BootServerSocket.java
+++ b/main-command/src/main/java/sbt/internal/BootServerSocket.java
@@ -123,12 +123,17 @@ public class BootServerSocket implements AutoCloseable {
                             bytes.put(b);
                             clientSocketReads.put(ClientSocket.this);
                           } else {
+                            // close() deregisters from clientSockets like the write
+                            // methods do; a dead entry left behind would block the
+                            // NO_BOOT_CLIENTS_CONNECTED signal in inputStream.read.
                             alive.set(false);
+                            close();
                           }
                         }
 
                       } catch (IOException e) {
                         alive.set(false);
+                        close();
                       }
                     }
                   } catch (final Exception ex) {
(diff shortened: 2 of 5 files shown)


BrianHotopp:
@eed3si9n friendly nudge when you get a chance.

main/src/main/scala/sbt/Main.scala
```diff
@@ -152,29 +152,42 @@ private[sbt] object xMain:
...
+    // sbt runs fine without a boot socket (see the UnsatisfiedLinkError case below), so a
+    // creation failure only means "another sbt is booting" if something answers on it.
+    def liveServerDetected: Boolean =
+      BootServerSocketProbe.liveServerDetected(
+        BootServerSocket.socketLocation(base, hash),
+        BootServerSocket.requiresJNI() || SysProp.serverUseJni,
+      )
+    def proceedWithout(e: Throwable): (Option[BootServerSocket], Option[Exit]) = {
+      val cause = Option(e.getCause).getOrElse(e)
+      println(s"could not create the boot io socket ($cause); continuing without it")
```
eed3si9n: As you noted, bootserversocket is a nice-to-have, so I think we can remove the above `println(...)` since there isn't much the user can do here.

eed3si9n:
Commented a minor nit, but overall LGTM. Thanks!

main/src/main/scala/sbt/Main.scala
BrianHotopp: Done in 7eb9762.

eed3si9n (approved):
Thanks!

PR #9344, opened 2026-06-15
[2.x] fix: Substitute [organisation] literally by defaulting Patterns.isMavenCompatible to false (#535)

### LINKED ISSUE #535: [organisation] and [organization] in ivy patterns behave like [orgPath] should in sftp and ssh resolvers.
I use the following resolver:

`Resolver.sftp("Cirque-ivy-repo", "daisyduck.cirque.dk") as (username, keyfile) artifacts 
"/var/ivy/cirque-repo/[organization]/[module]/[type]s/[artifact]-[revision].[ext]" ivys 
"/var/ivy/cirque-repo/[organisation]/[module]/ivys/ivy-[revision].xml"`

When i add the followin dependency: `libraryDependencies += "dk.cirque.util" % "cirque-util-concurrency" % "3.6.0"` it tries to resolve the ivy and artifact like this:

```
[warn] ==== Cirque-ivy-repo: tried
[warn]   /var/ivy/cirque-repo/dk/cirque/util/cirque-util-concurrency/ivys/ivy-3.6.0.xml
[warn]   -- artifact dk.cirque.util#cirque-util-concurrency;3.6.0!cirque-util-concurrency.jar:
[warn]   /var/ivy/cirque-repo/dk/cirque/util/cirque-util-concurrency/jars/cirque-util-concurrency-3.6.0.jar
[warn]   dk/cirque/util/cirque-util-concurrency/3.6.0/cirque-util-concurrency-3.6.0.jar
```

The problem is path it tries. The organisation of the dependency is dk.cirque.util so according to the pattern, the path of the ivy file should be `/var/ivy/cirque-repo/dk.cirque.util/cirque-util-concurrency/ivys/ivy-3.6.0.xml` but instead it replaces the . with / making the path: `/var/ivy/cirque-repo/dk/cirque/util/cirque-util-concurrency/ivys/ivy-3.6.0.xml`

According the documentation for ivy http://ant.apache.org/ivy/history/latest-milestone/concept.html it would be the correct behivor of [orgPath]

Note that i have tried both [organisation] and [organization] and both replaces the . with /

jozanek:
## Problem

Fixes #535.

In an Ivy resolver pattern, the `[organisation]`/`[organization]` token has its dots rewritten to slashes (`org.example` → `org/example`), behaving like Ivy's `[orgPath]` token rather than being literal. Per the [Apache Ivy spec](http://ant.apache.org/ivy/history/latest-milestone/concept.html), `[organisation]` should be substituted literally and `[orgPath]` is the slash-separated form.

Root cause: `Patterns.isMavenCompatible` defaulted to `true`, which sbt forwards to Apache Ivy as `setM2compatible(true)`; with m2-compatibility on, Ivy rewrites the `[organisation]` token to slash form. A user supplying a custom Ivy pattern (e.g. for an SFTP/SSH resolver) inherited that default and got the wrong paths, with no obvious indication why. The only workaround was the non-obvious `withIsMavenCompatible(false)`.

## Solution

Flip the default of `Patterns.isMavenCompatible` from `true` to `false`, so a hand-written `Patterns` keeps `[organisation]` literal by default — matching the Ivy spec. The change is deliberately narrow:

- **Contraband schema** (`Patterns.isMavenCompatible`): default `true → false`, with documentation of the token semantics.
- **`Resolver.mavenStylePatterns`** is pinned to `isMavenCompatible = true` explicitly, so `Resolver.url`/`file`/`sftp`/`ssh` built from the default patterns keep producing the Maven layout — unchanged behavior.
- **`Patterns(artifactPatterns*)` shorthand** now also defaults to `false`, so all construction paths (`Patterns()`, the varargs shorthand, and the builder) agree.
- No changes to the Ivy conversion (`ConvertResolver` already forwards the flag) or to Coursier/publish logic.

The only behavioral change is the default for freshly-constructed `Patterns`.

### Compatibility

Documented in `notes/2.0.0/patterns-maven-compatible-default.md`. Affected: custom URL/file/SFTP/SSH resolvers with hand-written `[organisation]` patterns that relied on the old implicit slash rewrite — under the Ivy engine, and (for custom URL/file patterns) under Coursier's Maven-base detection. Opt back in with `withIsMavenCompatible(true)` or use the `[orgPath]` token.

## AI assistance

This change was developed with AI assistance (Claude Code), reviewed and verified human-in-the-loop by running the test suites, MIMA, and the scripted test locally on JDK 17.

### DIFF
--- a/lm-core/src/main/contraband-scala/sbt/librarymanagement/Patterns.scala
+++ b/lm-core/src/main/contraband-scala/sbt/librarymanagement/Patterns.scala
@@ -4,6 +4,10 @@
 
 // DO NOT EDIT MANUALLY
 package sbt.librarymanagement
+/** @param isMavenCompatible When false (the default), the [organisation]/[organization] token is substituted literally,
+so an organization such as `org.example` stays `org.example` in resolved paths, matching the
+Apache Ivy specification. When true, the Ivy engine rewrites the [organisation] token to the
+slash-separated form (`org/example`), as in a Maven m2-compatible repository layout. */
 final class Patterns private (
   val ivyPatterns: Vector[String],
   val artifactPatterns: Vector[String],
@@ -11,7 +15,7 @@ final class Patterns private (
   val descriptorOptional: Boolean,
   val skipConsistencyCheck: Boolean) extends Serializable {
   
-  private def this() = this(Vector.empty, Vector.empty, true, false, false)
+  private def this() = this(Vector.empty, Vector.empty, false, false, false)
   
   override def equals(o: Any): Boolean = this.eq(o.asInstanceOf[AnyRef]) || (o match {
     case x: Patterns => (this.ivyPatterns == x.ivyPatterns) && (this.artifactPatterns == x.artifactPatterns) && (this.isMavenCompatible == x.isMavenCompatible) && (this.descriptorOptional == x.descriptorOptional) && (this.skipConsistencyCheck == x.skipConsistencyCheck)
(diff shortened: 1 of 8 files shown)


eed3si9n (approved):
Thanks!

PR #9347, opened 2026-06-15
[2.x] fix: Resolve Seq[Initialize[Task[T]]].join to Initialize[Task[Seq[T]]] (#899)

### LINKED ISSUE #899: incorrect implicit resolution for `join` on `Seq[Initialize[Task[T]]]`
`Seq[Initialize[Task[T]].join` resolves to `Seq[Initialize[S]].join` for `S=Task[T]` resulting in `Initialize[Seq[Task[T]]]` instead of `Initialize[Task[Seq[T]]]`.

jozanek:
Fixes #899.

`.join` on a `Seq[Initialize[Task[T]]]` could resolve through the generic `Initialize.joinInitialize` conversion (yielding `Initialize[Seq[Task[T]]]`) instead of the task-flattening `Initialize[Task[Seq[T]]]`, whenever the task-specialized `Scoped.richTaskSeq` was not in scope.

Add the task join as an **extension method** on `Seq[Def.Initialize[Task[A]]]` in `object Def`. Member resolution tries extension methods before implicit conversions, so it's selected ahead of the generic conversion regardless of whether `sbt.Scoped` is in scope — and, being in a different resolution phase than the existing `Scoped.richTaskSeq` conversion, it doesn't become a second ambiguous candidate for `TaskKey`/`SettingKey` seqs (where `Scoped` is a base type). `Scoped.richTaskSeq` is left unchanged as the existing public implicit.

Also:
- Drop the now-unnecessary `import sbt.Scoped.richTaskSeq` from `BuildServerProtocol` and simplify `ScopeFilter.all`'s `.join(_.join)` to `.join`.
- Make the two `dependsOn(Initialize[? <: Task[?]]*)` overloads delegate to `dependsOnSeq` (explicit `Initialize.joinAny[Task]` raw-task form) instead of relying on `.join` resolving to the generic conversion.
- Add a compile-only regression test pinning resolution for both a plain `Seq[Initialize[Task[Int]]]` and a `Seq[TaskKey[Int]]`, inferring the result with no expected type.

_Developed with AI assistance (Claude Code); verified locally via `publishLocalBin` + the `actions/join` scripted test._


main-settings/src/main/scala/sbt/Structure.scala
```diff
@@ -494,7 +492,10 @@ object Scoped:
 
   type AnyInitTask = Initialize[Task[?]]
 
-  implicit def richTaskSeq[T](in: Seq[Initialize[Task[T]]]): RichTaskSeq[T] = new RichTaskSeq(in)
+  // Not `implicit`: the implicit conversion lives in `Def.richInitializeTaskSeq` so it is in the
+  // implicit scope of every `Def.Initialize`-derived type (including `TaskKey`/`SettingKey`) and is
+  // the single candidate, avoiding the ambiguity that two competing conversions would cause (see #899).
+  def richTaskSeq[T](in: Seq[Initialize[Task[T]]]): RichTaskSeq[T] = new RichTaskSeq(in)
```
eed3si9n: `richTaskSeq` is part of the existing API, so I don't think it's safe to remove implicits here.

eed3si9n (changes requested):

main-settings/src/main/scala/sbt/Structure.scala
jozanek: Reverted. `Scoped.richTaskSeq` stays as-is. Made the task join an extension method on `Seq[Def.Initialize[Task[A]]]` instead. Extension methods resolve before conversions, so it fixes the plain `Seq[Initialize[Task[T]]]` shape without tying with `richTaskSeq` for `TaskKey` seqs.

jozanek:
@eed3si9n pinging you and asking for another review 🙏

main-settings/src/main/scala/sbt/Structure.scala
```diff
@@ -364,7 +364,7 @@ object Scoped:
 
       // Task-specific extensions
       def dependsOn(tasks: Initialize[? <: Task[?]]*): Initialize[Task[A1]] =
-        init.zipWith(tasks.asInstanceOf[Seq[Initialize[Task[?]]]].join)(_.dependsOn(_*))
+        dependsOnSeq(tasks.asInstanceOf[Seq[AnyInitTask]])
```
eed3si9n: If this is unrelated, could separate this out please?

eed3si9n (commented):
Left a comment about unrelated changes, but overall LGTM

main-settings/src/main/scala/sbt/Structure.scala
jozanek: Reverted

eed3si9n (approved):
Thanks!

PR #9356, opened 2026-06-18
[2.0.x] Preserve explicit plugin cross-build sbt versions

mkurz:
This fixes a setting precedence issue in `SbtPlugin`.

In sbt 2, `SbtPlugin` currently assigns:

```scala
pluginCrossBuild / sbtVersion := {
  scalaBinaryVersion.value match
    case "3"    => sbtVersion.value
    case "2.12" => "1.5.8"
    case "2.10" => "0.13.18"
}
```

That means a Scala 2.12 sbt plugin project always gets `1.5.8` from `SbtPlugin`, even when another AutoPlugin has already provided an explicit `pluginCrossBuild / sbtVersion` setting.

This becomes a problem for plugin builds that factor common sbt plugin settings into one AutoPlugin, and then define a concrete plugin with:

```scala
override def requires = CommonPlugin && SbtPlugin
```

In that shape, the common AutoPlugin can set a newer sbt 1.x version, but `SbtPlugin` later overwrites it with `1.5.8`. The plugin then compiles against old sbt/librarymanagement APIs, even though the build explicitly configured a newer plugin cross-build sbt version.

The fix changes `SbtPlugin` to transform the inherited `pluginCrossBuild / sbtVersion` value instead of assigning over it unconditionally. It still keeps the existing Scala 2.12 default of `1.5.8` when the inherited value is just the running sbt version, but it preserves an explicit inherited value.

The new scripted test covers both cases:

- an AutoPlugin-provided `pluginCrossBuild / sbtVersion` survives when a concrete plugin also requires `SbtPlugin`
- a plain Scala 2.12 `SbtPlugin` project still defaults to `1.5.8`

This came up while testing Play Framework with sbt 2.0.0, where Play’s sbt plugin build explicitly configured a newer sbt 1.x cross-build version but still ended up compiling against sbt 1.5.8.


mkurz:
Disclaimer: This was mostly done with codex (gpt-5.5)

mkurz:
OK, actually this fix is wrong... codex didn't figure it out. I should have reviewed before submitting ;)
Lets see if we can fix it...

mkurz:
I reverted the fix and just kept the scripted test

@eed3si9n If you know how to fix that, let me know. I do not have time currently. Maybe I leave codex running before going to bed to figure it out...

mkurz:
See https://github.com/playframework/playframework/pull/14019/changes#r3436279549

eed3si9n:
> In that shape, the common AutoPlugin can set a newer sbt 1.x version, but `SbtPlugin` later overwrites it with `1.5.8`.

Why not make CommonPlugin require SbtPlugin?

PR #9364, opened 2026-06-21
[2.x] perf: Optimize incremental test further

yhefamly:
**Note**: This builds on the open PR #9253, so that one should be merged first – until then, this PR's diff shows the changes of both. I hope the additional changes can still get some feedback in the meantime :)

This applies a range of optimisations local to `ClassStamper` that bring the time needed for `refinedTestDigests` down to acceptable levels (see #9108):
* Cache the digests of transitive dependencies (big impact)
* Avoid sorting of digest subsets that would later get sorted again (small)
* Pre-compute which `Analysis` instances are required for each class to avoid repeated scanning of the whole list (medium)
* Merge two loops on `relations.externalDeps` into one (small)
* Compute the set of extra digests outside loop (small)
* Track the digest closure of each test via a `BitSet` (big)

In total, these optimizations bring down the time required for `refinedTestDigests` by a factor of several hundred, both in a [reproducer repo](https://github.com/yhefamly/sbt-slow-defined-test-digests) as well as a proprietary real-life code base, like from 600s (already with the optimisations of #9253) to 1.2s. The resulting digests should match what the currently released sbt version 2.0.0 produces, at least it does so in the `ClassStamperTest` from #9253. Also, I've tried with doing small changes to test or main source files, sbt still correctly determined whether a test needed re-running.

**Disclaimer**: Claude Opus 4.8 has been used to do some of the coding and validation. While I've used sbt for many years, I'm no expert in its internals, so I can't guarantee I've not been misled by AI somewhere. But I've reviewed all code, tested it myself, checked internal state with a debugger and I am rather confident this is not pure nonsense🤞


eed3si9n:
@yhefamly Could you sign the Scala CLA please? <https://contribute.akka.io/contribute/cla/scala >

yhefamly:
> @yhefamly Could you sign the Scala CLA please? <https://contribute.akka.io/contribute/cla/scala >

Signed. Also just fixed compilation, had accidentally deleted one line while preparing the PR.

eed3si9n (approved):
Thanks!

PR #9376, opened 2026-06-24
[2.x] Fix publishDiagnostics propagation

### LINKED ISSUE #9345: [2.x] SBT2 doesn't appear to emit build/publishDiagnostics messages
## steps
Use an SBT2 project as the BSP for metals

## problem

build/publishDiagnostics is not sent

## expectation

build/publishDiagnostics is sent

## notes

I have tested this on two projects, and the difference seems to be the lack of emission of build/publishDiagnostics in sbt 2.

I made a simple error in both of these projects. 

SBT 1.12.12
[bsp.trace.json](https://github.com/user-attachments/files/28972041/bsp.trace.json)
SBT 2
[bsp.trace.json](https://github.com/user-attachments/files/28972049/bsp.trace.json)

anatoliykmetyuk:
Fixes #9345

Test is based on https://github.com/sbt/sbt/issues/9345#issuecomment-4718229113 which gives us the following sequence:

1. Metals sends buildTarget/compile.
2. sbt publishes real non-empty diagnostics.
3. Metals sends buildTarget/scalaMainClasses.
4. During that request, sbt emits build/publishDiagnostics with diagnostics: [] and reset: true.
5. The following build/taskFinish still reports errors: 1.

Previously, errors for diagnostics reporting via bsp were collected from a live compilation run. In the sequence above, that is triggered by `buildTarget/compile`. Then, `buildTarget/scalaMainClasses` does not trigger such a run for the second time, it uses the cached compilation result. Therefore, the diagnostics is not populated.

The proposed fix modifies `sendFailureReport` to accept an optional `CompileFailed` object that contains the diagnostics even in case the actual compiler did not run because the cached result was used. If no problems were found for a file via default means, this `CompileFailed` object is queried to see if it has any information about problems in a given file.

### DIFF
--- a/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
+++ b/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
@@ -14,6 +14,7 @@ import sbt.internal.util.ManagedLogger
 import sbt.internal.server.BuildServerProtocol.BspCompileState
 import xsbti.compile.CompileAnalysis
 import xsbti.{
+  CompileFailed,
   FileConverter,
   Problem,
   Reporter,
@@ -46,6 +47,9 @@ sealed trait BuildServerReporter extends Reporter {
   def sendSuccessReport(analysis: CompileAnalysis): Unit
 
   def sendFailureReport(sources: Array[VirtualFile]): Unit
+
+  def sendFailureReport(sources: Array[VirtualFile], failure: Option[CompileFailed]): Unit =
+    sendFailureReport(sources)
 
   override def reset(): Unit = underlying.reset()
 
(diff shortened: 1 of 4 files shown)


main/src/main/scala/sbt/internal/server/BuildServerReporter.scala
```diff
@@ -46,10 +46,10 @@ sealed trait BuildServerReporter extends Reporter {
...
   def sendSuccessReport(analysis: CompileAnalysis): Unit
 
-  def sendFailureReport(sources: Array[VirtualFile]): Unit
-
-  def sendFailureReport(sources: Array[VirtualFile], failure: Option[CompileFailed]): Unit =
-    sendFailureReport(sources)
+  def sendFailureReport(
+      sources: Array[VirtualFile],
+      failure: Option[CompileFailed] = None
+  ): Unit
```
eed3si9n, line 54: If possible, we should avoid using default parameters esp with traits. Maybe keep them both unimplemented at this layer if this was causing issue.

anatoliykmetyuk: Fixed. Out of curiosity, what's the reasoning behind avoiding default parameters?

eed3si9n: 1. it ends up complicating the situation when we want to add more overloads later
2. it also makes it difficult to keep binary compatibility since the default arguments are generated as synthetic variables under the companion (iirc)

eed3si9n (approved):
Thanks!

PR #9391, opened 2026-06-27
[2.x] fix: Fixes global plugin loading

### LINKED ISSUE #8600: [2.x] `*.sbt` files in a global plugin directory are not loaded

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/8600

## Problem
Global plugin loading doesn't work.

## Solution
1. Use `ModuleID#from` to supply the location of global-plugin module.
2. Update pluginData with the global plugin classpath.

## Manual testing

`$XDG_CONFIG_HOME/sbt/2/plugins/g.sbt`:

```scala
addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.13.1")
```

Then from an empty project:

```bash
sbt:foo> plugins
....
  sbtbuildinfo.BuildInfoPlugin
sbt:aaa> eval sbtbuildinfo.BuildInfoPlugin
[info] ans: sbtbuildinfo.BuildInfoPlugin = sbtbuildinfo.BuildInfoPlugin
```


xuwei-k:
```
[info] [error] java.lang.AssertionError: assertion failed: 
[info] [error] ClassifiersModule dependencies mismatch.
[info] [error] Expected: com.eed3si9n.manifesto:manifesto_3, junit:junit, org.hamcrest:hamcrest-core, org.scala-lang:scala-library, org.typelevel:cats-core_3, org.typelevel:cats-kernel_3, org.scala-sbt:sbt, org.scala-lang:scala3-library_3, com.eed3si9n:sbt-buildinfo_sbt2_3
[info] [error] Actual:   com.eed3si9n.manifesto:manifesto_3, junit:junit, org.hamcrest:hamcrest-core, org.scala-lang:scala-library, org.typelevel:cats-core_3, org.typelevel:cats-kernel_3, org.scala-sbt:sbt, org.scala-sbt:global-plugins, org.scala-lang:scala3-library_3, com.eed3si9n:sbt-buildinfo_sbt2_3
```

https://github.com/sbt/sbt/blob/aa0cb95c4863cf2694e2691064c2412ece7b6fe6/sbt-app/src/sbt-test/dependency-management/i8679-update-sbt-classifiers-local-plugin/build.sbt#L18-L36

eed3si9n:
I've worked around this by changing the test in https://github.com/sbt/sbt/pull/9391/changes/29390e479f941e2396f48f90416422f9085fdba2. `global-plugin` as a module existed in sbt 1.x (https://github.com/sbt/sbt/blob/v1.12.13/main/src/main/scala/sbt/internal/GlobalPlugin.scala#L119) as well. If we can make the metabuild to treat global-plugin as a inter-project dependency it would better, but hopefully this would still work the same way.

xuwei-k (approved):

PR #9400, opened 2026-06-29
Add issue reproduction skill

anatoliykmetyuk:
Adds an issue reproduction skill available for LLM agents started in this repo. With this skill, a prompt such as:

`reproduct issue <github_url>`

Will trigger a well-defined workflow to create the reproduction. The result is a minimal sbt project under `local-temp/issues/<issue_name>`, containing a script `reproduce.sh` that will print `REPRODUCED: yes` or `REPRODUCED: no` depending whether the bug is still present. For example:

image

`reproduce.sh` will reproduce the issue using sbt published locally from this repo's codebase, while `reproduce.sh --direct` will reproduce the issue using sbt version downloaded from Maven, as specified in the issue report. This is useful to distinguish issues that were already fixed but not backported.

## Validation

Even with weaker OSS models such as Minimax m3, this skill reliably produces correct issue reproductions of required shape, both for well-defined and vague issue reports, without the help of stronger models like GPT-5.5. I've used this skill on 10+ issues in the context of fixing and review of other people's PRs, and it saves a lot of time usually required to come up with the baseline reproduction.

## Status

This is a workflow that has been useful to me for fixing issues and reviewing PRs, so I'd like to contribute it. At the same time, it does introduce a new `skills/` location and a new way for agents to interact with the project, so I understand if we decide not to go in this direction at this time.

### DIFF
--- a/.agents/skills/sbt-issue-fixing/SKILL.md
+++ b/.agents/skills/sbt-issue-fixing/SKILL.md
@@ -1,0 +1,176 @@
+---
+name: sbt-issue-fixing
+description: Reproduce, investigate, plan fixes for, and fix sbt GitHub issues. Use when the user asks to reproduce an issue such as "reproduce issue ISSUE_URL", fix an issue, investigate an sbt issue, or come up with a plan for an sbt issue fix.
+---
+
+# sbt Issue Fixing
+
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #9409, opened 2026-07-05
[2.x] fix: Fixes auto import of givens

xuwei-k:
sbt 2.0.1

```
[error] -- [E172] Type Error: /private/var/folders/nz/vb2z3s8j7719gg71zgz9gj_40000gn/T/sbt_51cc33ac/build.sbt:6:25 
[error] 6 |  assert(summon[Foo[Int]].value == 2)
[error]   |                         ^
[error]   |No given instance of type Foo[Int] was found for parameter x of method summon in object Predef
[error]   |
[error]   |Note: given instance given_Foo_Int in object $Wrap2d357da4fb was not considered because it was not imported with `import given`.
[error] -- [E172] Type Error: /private/var/folders/nz/vb2z3s8j7719gg71zgz9gj_40000gn/T/sbt_51cc33ac/build.sbt:7:28 
[error] 7 |  assert(summon[Foo[String]].value == "a")
[error]   |                            ^
[error]   |No given instance of type Foo[String] was found for parameter x of method summon in object Predef
[error]   |
[error]   |Note: given instance s in object $Wrap2d357da4fb was not considered because it was not imported with `import given`.
[error] -- [E006] Not Found Error: /private/var/folders/nz/vb2z3s8j7719gg71zgz9gj_40000gn/T/sbt_51cc33ac/build.sbt:8:9 
[error] 8 |  assert(s.value == "a")
[error]   |         ^
[error]   |         Not found: s
[error]   |
[error]   | longer explanation available when compiling with `-explain`
```


eed3si9n (approved):
Thanks!

Issue #9373, opened 2026-06-23
Ctrl+C does not stop a task that floods stdout — cancellation doesn't discard the pending server→client output backlog


soronpo:
## steps

- Any sbt (reproduced on **1.12.12**; same code in 1.10.x and **sbt 2.0.0**), using the thin client
  `sbtn`.
- In an empty directory, a single `build.sbt` (no `project/build.properties` needed):

  ```scala
  lazy val floodForever = taskKey[Unit]("flood stdout forever")
  floodForever := {
    var i = 0
    while (true) { println(s"flood line $i ...................................... padding $i"); i += 1 }
  }
  ```

- Run `sbtn floodForever`, let it flood the terminal for a few seconds, then press **Ctrl+C** to cancel.

What happens: the run does **not** stop. Output keeps flooding the terminal for a long time after
Ctrl+C (and after the task itself has been cancelled / its process killed). The session appears hung.

## problem

Under a flood, Ctrl+C fails to stop the output, for two compounding reasons in
`sbt.internal.server.NetworkChannel`:

1. **Cancellation does not discard the pending output.** A cancel request is handled by
   `onCancellationRequest → runningEngine.cancelAndShutdown()`, which stops the *task* — but nothing
   clears the channel's outgoing output queue. There is no `pendingWrites.clear()` (or equivalent) on
   cancel. So the background write thread keeps draining everything already queued to the client:

   ```scala
   // main/.../server/NetworkChannel.scala (sbt 2.0.x; same in 1.10–1.12)
   private[this] val pendingWrites = new LinkedBlockingQueue[(Array[Byte], Boolean)]()
   private val writeThread = new Thread(() => { /* ... */ pendingWrites.take; out.write(event); out.flush() /* ... */ })
   // onCancellationRequest(...) cancels the task but never touches pendingWrites.
   ```

   So even though the task is cancelled, the user keeps seeing output — the cancel has no visible
   effect until the entire backlog has drained.

2. **The backlog is unbounded, so it can be enormous.** `publishBytes` does a non-blocking `put` onto
   the **unbounded** `pendingWrites`; the write thread blocks on `out.write/flush` when the client is
   slow (its own comment: *"Do writes on a background thread because otherwise the client socket can
   get blocked"*). With no cap and no back-pressure, a fast producer + slow client (a runaway
   `println` vs. a normal terminal) lets `pendingWrites` grow without bound — we observed the server
[...]

PR #9411, opened 2026-07-06
[2.x] fix: Discard a cancelled task's output backlog on cancel

BrianHotopp:
## Problem

A task that floods stdout cannot be stopped promptly. Cancelling it (Ctrl+C from
`sbtn` / `sbt --client`) stops the task, but the server keeps draining the
already-queued output to the client for several seconds, so the cancel appears
not to take effect (#9373).

## Fix

`NetworkChannel.onCancellationRequest` now sets a per-channel `discarding` flag
and clears the queued frames and buffered stdout (`discardPending`). While the
flag is set, `jsonRpcNotify` drops the cancelled task's `systemOut`/`systemErr`
instead of delivering the backlog, so output stops promptly.

- Only those two data channels are gated. Control-plane replies (including the
  cancel ack, enqueued after the clear) and all other notifications flow normally.
  `NetworkChannel.isCancelledOutput` is a pure function so this contract is
  unit-tested.
- It is a flag, not a one-shot clear, because `cancelAndShutdown` only interrupts:
  a CPU-bound print loop keeps producing until it actually stops, and the flag
  gates that late output.
- The flag is reset at the next prompt (`prompt(ConsolePromptEvent)`, which fires
  per command), so the following command's output is delivered normally.

## Scope

This addresses the titled Ctrl+C symptom (the output backlog). The server-side
heap growth under a sustained flood (the ~6.6GB in the report) is a separate
follow-up: capping it needs back-pressure on the output buffers, which depends on
a separate fix to the output-flush coalescing that I am preparing independently.
Kept this change to the discard-on-cancel path so it stays surgical.

The object-level `NetworkChannel.cancel` path (server-side `shutdownall`) is
static with no channel handle, so it does not apply the discard; the interactive
Ctrl+C path (`onCancellationRequest`) is the one users hit.

## Testing

- `NetworkChannelSpec` (Verify): pins the control-plane-safety contract (only
  `systemOut`/`systemErr` are ever gated; responses, `Shutdown`, `build/logMessage`,
  flush markers, etc. are never dropped). A regression here would silently drop
  protocol replies.
- The discard and per-prompt reset behavior were verified with standalone
  concurrency harnesses (deadlock-free, frame-safe, reset-always-fires,
  cancel-ack-preserved); they are timing-dependent and kept out of the
  deterministic CI suite.
- `mainProj/scalafmtCheck`, `mainProj/Test/scalafmtCheck`,
  `mainProj/mimaReportBinaryIssues` green.

Refs #9373

---

*Implemented with AI assistance (Claude Code); reviewed and verified locally by me
(compile, unit test, scalafmt, MiMa, plus adversarial concurrency verification).*

### DIFF
--- a/main/src/main/scala/sbt/internal/server/NetworkChannel.scala
+++ b/main/src/main/scala/sbt/internal/server/NetworkChannel.scala
@@ -92,6 +92,7 @@ final class NetworkChannel(
   private val pendingWrites = new LinkedBlockingQueue[(Array[Byte], Boolean)]()
   private val attached = new AtomicBoolean(false)
   private val alive = new AtomicBoolean(true)
+  private val isCanceled = new AtomicBoolean(false)
   private[sbt] def isInteractive = interactive.get
   private val interactive = new AtomicBoolean(false)
   private[sbt] def setInteractive(id: String, value: Boolean) = {
@@ -109,6 +110,10 @@ final class NetworkChannel(
   private[sbt] def prompt(): Unit = {
     interactive.set(true)
     jsonRpcNotify(promptChannel, "")
+  }
+  override def prompt(e: ConsolePromptEvent): Unit = {
+    isCanceled.set(false)
+    super.prompt(e)
   }
   private[sbt] def write(byte: Byte) = inputBuffer.add(byte.toInt)
 
@@ -533,6 +538,8 @@ final class NetworkChannel(
                 StandardMain.exchange.currentExec.exists(_.source.exists(_.channelName == name)))
             ) {
               runningEngine.cancelAndShutdown()
+              isCanceled.set(true)
+              discardPending()
 
               respondResult(
                 ExecStatusEvent(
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/internal/server/NetworkChannel.scala
```diff
@@ -92,6 +92,9 @@ final class NetworkChannel(
   private val pendingWrites = new LinkedBlockingQueue[(Array[Byte], Boolean)]()
   private val attached = new AtomicBoolean(false)
   private val alive = new AtomicBoolean(true)
+  // Set while a cancel is in flight so the cancelled task's still-queued output is discarded
+  // rather than drained to the client for seconds after Ctrl+C (#9373). Reset at the next prompt.
+  private val discarding = new AtomicBoolean(false)
```
eed3si9n: This is more of a nitpick, but for AI-generated code, I prefer as few comment - https://github.com/sbt/sbt/blob/develop/AGENTS.md#coding-style, and try to make the variable name represent the context. That would minimize the cognitive load / tokens for both humans and AI would need to read in the future. For example, naming to `isCanceled` can capture the Ctrl-C-ness.

BrianHotopp: Done in 9058e58: renamed to `isCanceled` and dropped the comments. Thanks!

eed3si9n (approved):
Thanks!

PR #9414, opened 2026-07-08
[2.x] fix: Keep the output-flush timer alive across forceFlush

### LINKED ISSUE #9415: [2.x] forceFlush shuts down the output-flush executor, defeating write coalescing
## steps

sbt 2.x server. A task that writes to stdout and then emits a log line (a very common shape) makes `NetworkChannel.forceFlush()` run while a coalesced stdout flush is still pending:

```scala
// build.sbt
TaskKey[Unit]("t") := { println("out"); streams.value.log.info("log") }
```

Run it from the thin client (`sbt t` / `sbt --client t`).

## problem

`forceFlush()` (`main/src/main/scala/sbt/internal/server/NetworkChannel.scala`) calls `flushExecutor.shutdownNow()` — the only place the per-connection output-flush executor is ever shut down. It is meant only to flush buffered stdout ahead of a control-plane message, but it tears the executor down. After the first `forceFlush` during active output the 20ms write-coalescing is gone for the rest of the connection: stdout is then flushed per write via the inline `RejectedExecutionException` fallback instead of batched. And if the `forceFlush` lands while a coalesced flush is pending, `shutdownNow()` cancels that future without resetting `flushFuture`, leaving the flush slot stuck.

Output still reaches the client via the inline fallback, so the observable effect is a defeated coalescing optimization (more flushes / more bytes on the wire, the flicker that coalescing exists to avoid), not lost or stalled output.

## expectation

`forceFlush` should flush the buffered output without shutting the shared timer down, so coalescing keeps working for the life of the connection.

## notes

Found while adding server-side output back-pressure: a bounded output buffer combined with a dead flush timer would deadlock the producing task, so the timer must stay alive first. Fixed by #9414.

BrianHotopp:
## Problem

`NetworkChannel` batches writes to the client at most once per 20ms (a scheduled
`flushExecutor`) to reduce terminal flicker and byte volume. `forceFlush()` --
called to order buffered stdout ahead of a control-plane message -- did
`flushExecutor.shutdownNow()`, the only place the executor was ever shut down.

Once `forceFlush()` runs during active output it tears the executor down (and, if
a coalesced flush was pending, cancels its future without resetting `flushFuture`,
leaving the slot stuck). From then on the 20ms coalescing is gone for the rest of
the connection: stdout is flushed per write via the inline fallback instead of
batched. Output still reaches the client, so the effect is extra flushes / flicker
rather than lost output -- but a per-connection executor being shut down on the
first `forceFlush` is clearly unintended (#9415).

## Fix

Extract the flush state machine into a small `private[server] CoalescingFlusher`:

- `flush()` schedules a coalesced drain (or drains inline once closed).
- `forceFlush()` drains immediately and **leaves the timer alive** -- a pending
  coalesced drain harmlessly drains the remainder when it fires.
- `close()` shuts the executor down once, at channel teardown, so the thread no
  longer leaks.

`doFlush` also now holds a lock across both the drain and the publish, so an
inline `forceFlush` drain and the timer's drain can't publish two stdout batches
out of order (this window pre-existed but keeping the timer alive would otherwise
widen it).

## Motivation

Found while building back-pressure on the server's output buffers to cap heap
growth under a sustained stdout flood (the OOM in #9373): a bounded output buffer
combined with a dead flush timer deadlocks the producing task, so the timer has to
stay alive first. Sending it on its own ahead of that follow-on; it is an
independent, pre-existing defect regardless.

## Testing

- `CoalescingFlusherSpec` (Verify): pins that repeated flushes coalesce into one
  drain, that `forceFlush` does not disable coalescing (a `forceFlush` + 50
  flushes yields exactly 2 drains; the old bug would give 51 inline drains), and
  that flush after close drains inline.
- `serverTestProj` `EventsTest`/`ResponseTest` smoke green.
- `mainProj/scalafmtCheck`, `mainProj/Test/scalafmtCheck`,
  `mainProj/mimaReportBinaryIssues` green.
- Release note: `notes/2.0.0/output-flush-coalescing.md`.

Fixes #9415

---

*Implemented with AI assistance (Claude Code); reviewed and verified locally by me
(compile, unit + server tests, scalafmt, MiMa).*

### DIFF
--- a/main/src/main/scala/sbt/internal/server/CoalescingFlusher.scala
+++ b/main/src/main/scala/sbt/internal/server/CoalescingFlusher.scala
@@ -1,0 +1,57 @@
+/*
+ * sbt
+ * Copyright 2023, Scala center
+ * Copyright 2011 - 2022, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt
+package internal
+package server
+
+import java.util.concurrent.{ Executors, Future, RejectedExecutionException, TimeUnit }
+import java.util.concurrent.atomic.AtomicReference
+import sbt.internal.util.Util
+
+/**
+ * Coalesces calls to a `drain` callback so it runs at most once per `delayMillis` window, on one
+ * dedicated thread. This batches remote-client writes to cut terminal flicker and byte volume.
+ *
+ * [[forceFlush]] runs the drain immediately (e.g. to order buffered stdout before a control-plane
+ * message) and leaves the timer live, so coalescing keeps working; a pending coalesced drain
+ * harmlessly drains the remainder when it fires. Only [[close]] shuts the thread down.
+ */
+private[server] final class CoalescingFlusher(
+    threadName: String,
+    delayMillis: Long,
+    drain: () => Unit
+) extends AutoCloseable:
+  private val executor = Executors.newSingleThreadScheduledExecutor(r => new Thread(r, threadName))
+  private val pending = new AtomicReference[Future[?]]
+
+  /**
+   * Schedule a coalesced drain if none is already pending; drain inline if the timer is closed.
+   * Not safe for concurrent callers (the check-then-schedule on `pending` is not atomic); the sole
+   * caller serializes its writes upstream.
+   */
+  def flush(): Unit =
+    pending.get match
+      case null =>
+        try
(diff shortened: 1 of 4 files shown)


eed3si9n:
Thanks! If this is an unreported bug, could you retroactively create an issue as well plz?

BrianHotopp:
Created! #9415

eed3si9n (approved):
Thanks!

PR #9416, opened 2026-07-10
[2.x] fix: Fix ivyless sbt plugin publish cross paths

### LINKED ISSUE #9383: [2.x] Missing cross suffixes for sbt1 plugin when useIvy is false
## steps

use sbt2 to publishLocal sbt1 plugin
set `useIvy' := false`

## problem

incorrect path used for publishing, scripted tests can't find the jar

## expectation

correctly published regardless of useIvy setting

## notes

repro here https://github.com/OlegYch/sbt2-publishlocal

anatoliykmetyuk:
Fixes #9383

## Problem
  When `useIvy := false`, ivyless publishing dropped the optional
  `scala_[scalaVersion]/` and `sbt_[sbtVersion]/` path segments from Ivy-style
  plugin publish patterns. This caused sbt 1 plugins published from sbt 2 builds
  to land under non-plugin Ivy paths, so consumers looking under
  `scala_2.12/sbt_1.0/` could not resolve them.

## Solution
  Read the plugin cross-version attributes from the `CsrProject` module and use
  them when constructing ivyless Ivy-layout publish paths. Apply the same
  substitution for local/file Ivy publishing and remote Ivy-style URL publishing,
  while continuing to omit those optional segments when the attributes are absent.
  Add scripted regressions for local and HTTP Ivy-style publishing of an sbt 1
  plugin with `useIvy := false`.

### DIFF
--- a/main/src/main/scala/sbt/internal/LibraryManagement.scala
+++ b/main/src/main/scala/sbt/internal/LibraryManagement.scala
@@ -12,11 +12,13 @@ package internal
 import java.io.{ File, IOException }
 import java.net.{ URI, URL }
 import java.util.concurrent.Callable
+import java.util.regex.Matcher
 
 import gigahorse.AuthScheme
 import gigahorse.support.apachehttp.Gigahorse
 import sbt.Def.ScopedKey
 import sbt.internal.librarymanagement.*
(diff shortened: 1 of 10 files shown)


eed3si9n (approved):
Thanks!

Issue #9379, opened 2026-06-24
[2.x] doesn't work on OpenBSD 7.9


catap:
## steps

Just install it

## problem

```
x $ sbt --allow-empty        
[error] sbtn is not supported on openbsd7.9
x $ 
```

## expectation

Should work

eed3si9n:
@catap Thanks for the report!

eed3si9n:
Here's a PR for this - https://github.com/sbt/sbt/pull/9394

catap:
@eed3si9n it wasn't enough. With usual `-Dsbt.server.autostart=false` I can't use it because it fails to start, and without it, it faild on invalid JNA compiled artifacts:
```
x $ sbt --allow-empty -Dsbt.server.autostart=false 
[error] no sbt server is running (sbt.server.autostart=false)
x $ echo $?
1
x $ sbt --allow-empty                              
java:/home/catap/.cache/JNA/temp/jna11338286882361088114.tmp: undefined symbol '__sF'
java.lang.UnsatisfiedLinkError: /home/catap/.cache/JNA/temp/jna11338286882361088114.tmp: dlerror returned no error description
        at java.base/jdk.internal.loader.NativeLibraries.load(Native Method)
        at java.base/jdk.internal.loader.NativeLibraries$NativeLibraryImpl.open(NativeLibraries.java:321)
        at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:187)
        at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:129)
        at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2245)
        at java.base/java.lang.Runtime.load0(Runtime.java:767)
        at java.base/java.lang.System.load(System.java:1647)
        at com.sun.jna.Native.loadNativeDispatchLibraryFromClasspath(Native.java:1045)
        at com.sun.jna.Native.loadNativeDispatchLibrary(Native.java:1015)
        at com.sun.jna.Native.<clinit>(Native.java:221)
        at org.scalasbt.ipcsocket.UnixDomainSocketLibrary.<clinit>(UnixDomainSocketLibrary.java:129)
        at org.scalasbt.ipcsocket.JNAUnixDomainSocketLibraryProvider.socket(UnixDomainSocketLibrary.java:165)
        at org.scalasbt.ipcsocket.UnixDomainSocket.<init>(UnixDomainSocket.java:53)
[...]

PR #9427, opened 2026-07-11
[2.x] Use JDK's Unix domain socket for bootserver

eed3si9n:
Ref https://github.com/sbt/sbt/issues/9379
Ref https://github.com/sbt/sbt/pull/8305

**Problem**
ipcsocket has some issues with it, like requiring platform-specific support for either JNA or JNI.

**Solution**
This uses JDK's Unix domain socket for bootserver.


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -402,7 +402,7 @@ object Defaults extends BuildCommon with DefExtra {
       },
       serverHandlers :== Nil,
       windowsServerSecurityLevel := Win32SecurityLevel.OWNER_DACL, // allows any owner logon session to access the server
-      serverUseJni := BootServerSocket.requiresJNI || SysProp.serverUseJni,
+      serverUseJni := SysProp.serverUseJni,
```
anatoliykmetyuk, line 405: Will this not introduce inconsistency for Apple Silicon?

Previously:

- Silicon, long-living socket, server, use JNI = true (https://github.com/sbt/sbt/blob/209867c7dbabd2f27f056f5fe7a34e32c6ac4b7b/main/src/main/scala/sbt/Defaults.scala#L405, https://github.com/sbt/sbt/blob/209867c7dbabd2f27f056f5fe7a34e32c6ac4b7b/main-command/src/main/java/sbt/internal/BootServerSocket.java#L361-L364)
- Silicon, long-living socket, client, use JNI = true (https://github.com/sbt/sbt/blob/1f6d19258ff470c07d5479b3f58d729a67ca88db/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala#L1559)

Now:

- Silicon, long-living socket, server, use JNI = false (as per this line, only system prop is taken into account)
- Silicon, long-living socket, client, use JNI = true (unchanged from before)

Is this PR meant to affect long-living socket at all or only the starup short-living socket? If only the boot socket is in scope, should the existing Apple Silicon default be preserved for serverUseJni?

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -337,6 +337,13 @@ class NetworkClient(
     conn
   }
 
+  private def bootSocketOpt(bootSocketName: String): Option[Socket] =
+    Try(ClientSocket.bootSocket(bootSocketName)).toOption match
+      case Some(x)                => Some(x)
+      case None if Util.isWindows =>
+        Try(ClientSocket.localSocket(bootSocketName, useJNI)).toOption
```
anatoliykmetyuk: Is `bootSocketName` correct? By the logic of: [localSocket](https://github.com/sbt/sbt/pull/9427/changes#diff-c7983417252ee30c8c698698d568a53cbce05788a576e0be14270546453c36deL45), this will prepend the legacy prefix to the `bootSocketName` which is the new JDK's new socket name?

The expected fallback name is `\\.\pipe\sbt-load<farmHash>` but here it will be `\\.\pipe\<new JDK filesystem path>`.

main-command/src/main/java/sbt/internal/BootServerSocket.java
```diff
@@ -339,31 +319,18 @@ public void close() {
...
+  static ServerSocketChannel newChannel(final String sock) throws ServerAlreadyBootingException {
     try {
-      if (!isWindows) Files.deleteIfExists(Paths.get(sock));
-      socket =
-          isWindows
-              ? new Win32NamedPipeServerSocket(name, jni, Win32SecurityLevel.OWNER_DACL)
-              : new UnixDomainServerSocket(name, jni);
-      return socket;
+      Files.deleteIfExists(Paths.get(sock));
+      final ServerSocketChannel channel = ServerSocketChannel.open(StandardProtocolFamily.UNIX);
```
anatoliykmetyuk, line 333: `ServerSocketChannel.open` [can throw](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/channels/ServerSocketChannel.html#open(java.net.ProtocolFamily)) `UnsupportedOperationException`, should probably catch to avoid a crash either here or around:

https://github.com/sbt/sbt/blob/1f6d19258ff470c07d5479b3f58d729a67ca88db/main/src/main/scala/sbt/Main.scala#L186-L187

main/src/main/scala/sbt/Defaults.scala
eed3si9n: > Is this PR meant to affect long-living socket at all or only the starup short-living socket? If only the boot socket is in scope, should the existing Apple Silicon default be preserved for serverUseJni?

The intent of the PR is for the startup short-lived socket. `BootServerSocket.requiresJNI()` is an interesting catch, but I think this was a workaround for JNA not having Apple Silicon (Aarch64/ARM64) support, which was fixed in 2021 as 5.7.0. If Linux is using JNA by default, it's probably ok for Arm Macs to switch to JNA. Ideally, we  will use JDK socket for them in later PR.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
eed3si9n: Yea, this looks like a bug. I need to probably pass in Windows named pipe name as second parameter.

eed3si9n: Addressed in 9369e159c112b63d8cf07e71763c97b55989af75

main-command/src/main/java/sbt/internal/BootServerSocket.java
eed3si9n: Addressed in 26b3c5315f9449b27a65a9fc641bc7a1a0cd8dee

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
anatoliykmetyuk: I think this might still be buggy:

https://github.com/sbt/sbt/blob/9369e159c112b63d8cf07e71763c97b55989af75/main-command/src/main/java/sbt/internal/BootServerSocket.java#L303-L305

https://github.com/sbt/sbt/blob/9369e159c112b63d8cf07e71763c97b55989af75/protocol/src/main/scala/sbt/protocol/ClientSocket.scala#L46-L48

`\\\\.\\pipe\\` is prepended twice. Maybe good adding some tests for new Windows client to old Windows server communication?

eed3si9n: Addressed in 7492dc9783836facf3d10f76433fae63d80e95cb

anatoliykmetyuk (approved):
LGTM!

Issue #963, opened 2013-11-07
Specifying -java-home at command line works in some cases, but not in all.


GlenKPeterson:
SBT is awesome, but I seem to have to change my path to point to my preferred JDK in some cases.

**Desired behavior**
SBT uses -java-home if specified

**Actual behavior**
Specifying -java-home works in some cases, but not in all.

**Evidence**

> me@myComputer:~/Documents/someCo/applications/code/myApp/WEB-INF$ $SBT_HOME/bin/sbt -java-home /usr/lib/jvm/latest-oracle-jdk
> Loading /home/me/tools/sbt-0.13/bin/sbt-launch-lib.bash
> [info] Set current project to MyAwesomeAppTest (in build file:/home/me/Documents/someCo/applications/code/myApp/WEB-INF/)
> test
> 
> Testing...

That worked as expected.  Then I made the following change to my build.sbt file (commented out 2 lines, added 2 new lines): 

> //unmanagedJars in Compile ++= 
> //  Attributed.blankSeq((file("/home/me/Documents/someCo/applications/code/myApp/WEB-INF/src/test/lib/") \* "*.jar").get)
> 
> libraryDependencies += "org.scalatest" % "scalatest_2.10" % "1.9.1"
> 
> libraryDependencies += "junit" % "junit" % "4.10"`

Now there is a problem:

> $SBT_HOME/bin/sbt -java-home /usr/lib/jvm/latest-oracle-jdk
> Loading /home/me/tools/sbt-0.13/bin/sbt-launch-lib.bash
> [info] Set current project to MyAwesomeAppTest (in build file:/home/me/Documents/someCo/applications/code/myApp/WEB-INF/)
> test
> [info] Updating {file:/home/me/Documents/someCo/applications/code/myApp/WEB-INF/}web-inf...
> [info] Resolving org.fusesource.jansi#jansi;1.4 ...
> [info] Done updating.
> [info] Compiling 6 Scala sources and 4 Java sources to /home/me/Documents/someCo/applications/code/myApp/WEB-INF/target/scala-2.10/test-classes...
> [trace] Stack trace suppressed: run last test:compile for the full output.
> [error](test:compile) java.io.IOException: Cannot run program "javac": error=2, No such file or directory
> [error] Total time: 19 s, completed Nov 7, 2013 10:08:53 AM

I was able to get it to work again by modifying my path as follows:

> me@myComputer:~/Documents/someCo/applications/code/myApp/WEB-INF$ echo $PATH
[...]

PR #9434, opened 2026-07-11
[2.x] fix: Honor -java-home in the Windows launchers

jozanek:
## Problem

`-java-home` wasn't fully honored on Windows: sbtw rejected the single-dash form, silently ignored bad paths, and didn't propagate the JDK to the sbt/sbtn child; sbt.bat didn't fix up `PATH`/`JAVACMD`, and a project `.java-version` could override an explicit `-java-home`. Residual #963 gaps (the original javac-from-PATH bug was fixed long ago).

## Solution

New `SelectedJava` seam in sbtw resolves the JDK once (`--java-home > JAVACMD > JAVA_HOME > PATH`) and applies one env overlay on every launch path; sbt.bat's handler now sets `PATH`/`JAVACMD` and a one-hop marker so `.java-version` defers to an explicit `-java-home`.

`JDK_HOME > JAVA_HOME` (the issue's "even more desired" ordering) is intentionally **not** implemented — it only mattered when `JAVA_HOME` pointed at a JRE without `javac`, and JDK 17+ (required by sbt 2.x) no longer ships standalone JREs, so `JAVA_HOME` always resolves to a full JDK.

### DIFF
--- a/sbtw/src/main/scala/sbtw/Runner.scala
+++ b/sbtw/src/main/scala/sbtw/Runner.scala
@@ -6,25 +6,6 @@ import scala.sys.process.*
 
 object Runner:
 
-  def findJavaCmd(javaHome: Option[String]): String =
-    val cmd = javaHome match
-      case Some(h) =>
-        val exe = new File(h, "bin/java.exe")
-        if exe.isFile then exe.getAbsolutePath
-        else
-          sys.env
-            .get("JAVACMD")
-            .orElse(
-              sys.env.get("JAVA_HOME").map(h0 => new File(h0, "bin/java.exe").getAbsolutePath)
-            )
-            .getOrElse("java")
-      case None =>
-        sys.env
-          .get("JAVACMD")
-          .orElse(sys.env.get("JAVA_HOME").map(h => new File(h, "bin/java.exe").getAbsolutePath))
-          .getOrElse("java")
-    cmd.replace("\"", "")
-
   def javaVersion(javaCmd: String): Int =
     try
       val pb = Process(Seq(javaCmd, "-Xms32M", "-Xmx32M", "-version"))
@@ -71,7 +52,12 @@ object Runner:
     if opts.jvmClient then s = s :+ "--client"
     s
 
-  def runNativeClient(sbtBinDir: File, scriptPath: String, opts: LauncherOptions): Int =
+  def runNativeClient(
+      sbtBinDir: File,
+      scriptPath: String,
+      opts: LauncherOptions,
+      selected: SelectedJava
+  ): Int =
     val sbtn = new File(sbtBinDir, "sbtn-x86_64-pc-win32.exe")
     if !sbtn.isFile then
       System.err.println("[error] sbtn-x86_64-pc-win32.exe not found in " + sbtBinDir)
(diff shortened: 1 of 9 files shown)


eed3si9n:
> sbtw rejected the single-dash form

Double dash is the UNIX command line standard, and as far as I know, JDK tooling post Java 9 also uses double dashes, so I won't think we should support single-dash in sbtw.

jozanek:
@eed3si9n sounds good, I'll drop single-dash normalization in sbtw and keep `--java-home` only.

One note: `.sbtopts` and `sbt.bat` still accept single-dash `-java-home` (the #963 workaround uses it in `.sbtopts`), so such a file won't parse under sbtw. I think that's fine since sbtw is new.

I'll remove the single-dash aliasing from `ArgParser`, keep the rest (`SelectedJava` propagation, `.java-version` handling, sbt.bat fixes), and update the tests.

jozanek:
@eed3si9n can you please review again?

sbtw/src/main/scala/sbtw/ArgParser.scala
```diff
@@ -4,56 +4,89 @@ import scopt.OParser
...
+    var expectValue = false
+    options.foreach: arg =>
+      if expectValue then
+        out += arg
+        expectValue = false
+      else
+        val normalized = alias(arg)
+        out += normalized
+        expectValue = normalized.startsWith("--") && valueTakingNames(normalized.drop(2))
+    out.result() ++ passthrough
```
eed3si9n: Do we still need this change?

sbtw/src/main/scala/sbtw/Runner.scala
```diff
@@ -6,26 +6,7 @@ import scala.sys.process.*
...
-            .getOrElse("java")
-      case None =>
-        sys.env
-          .get("JAVACMD")
-          .orElse(sys.env.get("JAVA_HOME").map(h => new File(h, "bin/java.exe").getAbsolutePath))
-          .getOrElse("java")
-    cmd.replace("\"", "")
-
-  def javaVersion(javaCmd: String): Int =
+  private[sbtw] def javaVersion(javaCmd: String): Int =
```
eed3si9n: Is there a reason why package private is added here?

eed3si9n:
Could we minimize the change to just bug fixes only please?

sbtw/src/main/scala/sbtw/ArgParser.scala
jozanek: No — this is the single-dash handling I said I would remove, but I had not pushed that change yet. Sorry for the confusion. It is gone now, and `ArgParser.scala` matches `develop` again.

sbtw/src/main/scala/sbtw/Runner.scala
jozanek: No real reason — I changed them back to plain `def` (also in `SelectedJava`).

jozanek:
I trimmed the PR down to the bug fix (+446/−99, was +595/−152). Removed the single-dash option handling, the `private[sbtw]` modifiers, and their tests.

What is left is the fix itself: the `sbt.bat` changes, `SelectedJava` (so the chosen JDK also reaches sbtn and the server), and quote-aware `.sbtopts` parsing for paths with spaces. Ready for another look.

eed3si9n (approved):
Thanks!

PR #9437, opened 2026-07-12
[2.x] Allow opt-out of transient warning

eed3si9n:
## Problem
We want to opt out of the transient key warning.

## Solution
This implements an optout via nowarn annotation.

## Test

```scala
import scala.annotation.nowarn

lazy val check = taskKey[Unit]("")
check := {
  val s = (state.value: @nowarn)
  println("hi")
}
```


core-macros/src/main/scala/sbt/internal/util/appmacro/ContextUtil.scala
```diff
@@ -194,6 +196,28 @@ trait ContextUtil[C <: Quotes & scala.Singleton](val valStart: Int):
 
   def idTransform[F[_]]: TermTransform[F] = in => in
 
+  /** Collects `qual.show` for each `.value` call wrapped in `@nowarn`, e.g. `(state.value: @nowarn)`. */
+  def collectNowarnQuals(tree: Term): Set[String] =
+    val result = mutable.HashSet[String]()
+    @tailrec def extractQual(t: Term): Unit = t match
+      case Inlined(_, _, inner)                                  => extractQual(inner)
+      case Typed(inner, _)                                       => extractQual(inner)
+      case Apply(TypeApply(Select(_, _), _ :: Nil), qual :: Nil) => result += qual.show
```
anatoliykmetyuk: This suppresses the warning for every occurrence of the same key rather than that specific occurrence:

<details>
<summary>Baseline</summary>

```scala
import scala.annotation.nowarn

lazy val check = taskKey[Unit]("")
check := {
  val a = state.value
  val b = state.value
  println("hi")
}
```

Output - two warnings:

```
[error] -- Warning: /private/var/folders/1f/9fyq55p93xz0tmp84_lq9p1r0000gn/T/sbt_d3649c89/build.sbt:4:0
[error]   4 |check := {
[error]     |^
[error]     |transient key state is excluded from the cache input
[error]     |
[error]   5 |  val a = state.value
[error]   6 |...
[error]   8 |}
[error]     |---------------------------------------------------------------------------
[error]     |Inline stack trace
[error]     |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[error]     |This location contains code that was inlined from build.sbt:4
[error]      ---------------------------------------------------------------------------
[error] -- Warning: /private/var/folders/1f/9fyq55p93xz0tmp84_lq9p1r0000gn/T/sbt_d3649c89/build.sbt:4:0
[error]   4 |check := {
[error]     |^
[error]     |transient key state is excluded from the cache input
[error]     |
[error]   5 |  val a = state.value
[error]   6 |...
[error]   8 |}
[error]     |---------------------------------------------------------------------------
[error]     |Inline stack trace
[error]     |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[error]     |This location contains code that was inlined from build.sbt:4
[error]      ---------------------------------------------------------------------------
```
</details>

<details>
<summary>One suppressed, zero warnings</summary>

```scala
import scala.annotation.nowarn

lazy val check = taskKey[Unit]("")
check := {
  val a = (state.value: @nowarn)
  val b = state.value
  println("hi")
}
```

Output: no warnings

Expected: one warning for `b`
</details>

core-macros/src/main/scala/sbt/internal/util/appmacro/ContextUtil.scala
```diff
@@ -194,6 +196,28 @@ trait ContextUtil[C <: Quotes & scala.Singleton](val valStart: Int):
...
+      case Apply(TypeApply(Ident(_), _ :: Nil), qual :: Nil)     => result += qual.show
+      case _                                                     => ()
+    object scanner extends TreeTraverser:
+      override def traverseTree(t: Tree)(owner: Symbol): Unit = t match
+        case Typed(inner, tpt) =>
+          tpt.tpe match
+            case AnnotatedType(_, annot) if annot.tpe.typeSymbol == nowarnAnnotSym =>
```
anatoliykmetyuk, line 212: This will respect `@nowarn` irrespectively of what warning it targets (see its [docs](https://www.scala-lang.org/api/current/scala/annotation/nowarn.html)):

```scala
import scala.annotation.nowarn

lazy val check = taskKey[Unit]("")

check := {
  val s = (state.value: @nowarn("msg=unused"))
  println("hi")
}
```

Output: no warnings, expected: one warning.

build.sbt
```diff
@@ -664,6 +664,10 @@ lazy val coreMacrosProj = (project in file("core-macros"))
     name := "Core Macros",
     SettingKey[Boolean]("exportPipelining") := false,
     mimaSettings,
+    mimaBinaryIssueFilters ++= Seq(
+      exclude[ReversedMissingMethodProblem]("sbt.internal.util.appmacro.ContextUtil.*"),
+      exclude[DirectMissingMethodProblem]("sbt.internal.util.appmacro.ContextUtil#Input.*"),
```
anatoliykmetyuk, line 669: Probably nitpick, but another way to do it without breaking binary compatibility could be by moving `collectNowarnQuals` as a private method into `Cont`, and tracking `isWarnSuppressed` as a local data structure at `Cont`. Will not necessarily result in more readable code, though, so probably the current solution is more idiomatic.

core-macros/src/main/scala/sbt/internal/util/appmacro/ContextUtil.scala
eed3si9n: Addressed in https://github.com/sbt/sbt/pull/9437/changes/c22c4101a8c9cceefb19c04bc42378e05b343d30 by collecting `qual` instead of `qual.show`.

core-macros/src/main/scala/sbt/internal/util/appmacro/ContextUtil.scala
eed3si9n: Addressed in https://github.com/sbt/sbt/pull/9437/changes/e2bc936fafcdf853823f051f132c3089bc7ae3f1 by checking the `@nowarn` payload and emulating "msg" check.

build.sbt
eed3si9n: Since ContextUtil doesn't leak out to the plugins, I'm not too worried about keeping the bincompat here.

anatoliykmetyuk (approved):
LGTM

Technically `@nowarn` emulation is still not compliant to the Scala's `@nowarn` contract: https://www.scala-lang.org/2021/01/12/configuring-and-suppressing-warnings.html#filters

E.g. regexes like `(state.value: @nowarn("msg=.*transient.*"))` would not be matched.

It seems however that the `@nowarn` logic is quite elaborate, and it wouldn't make sense to attempt to reinvent the wheel from sbt side. The correct solution seems to be on the Scala 3 side, e.g. to expose the `@nowarn` logic e.g. as a macro API for use cases like this one.

Issue #1032, opened 2013-12-09
In a multi-project build, working directory is inconsistent between forked and non-forked run and test


brikis98:
If you have an [SBT multiproject build](http://www.scala-sbt.org/release/docs/Getting-Started/Multi-Project) with multiple play apps, the typical working pattern is something like this:

```
> cd ~/source/foo
> play
[snip]
> project foo-frontend
> run
[snip]
> CTRL + D
> project foo-backend
> run
```

In all of these cases, since the JVM was originally launched from `~/source/foo`, the working directory (system property `user.dir`) will be set to `~/source/foo`.

However, if you fork the JVM to run tests, the forked JVM will get its working directory from SBT's `forkOptions` which are defined in [Defaults.scala](https://github.com/sbt/sbt/blob/0.13/main/src/main/scala/sbt/Defaults.scala#L404) as:

``` scala
private[this] def forkOptions: Initialize[Task[ForkOptions]] = (baseDirectory, javaOptions, outputStrategy, envVars, javaHome, connectInput) map {
  (base, options, strategy, env, javaHomeDir, connectIn) =>

    ForkOptions(bootJars = Nil, javaHome = javaHomeDir, connectInput = connectIn, outputStrategy = strategy, runJVMOptions = options, workingDirectory = Some(base), envVars = env)
}
```

Note how `workingDirectory` is set to `Some(base)`, which will be different for each sub-project in a multiproject build: `~/source/foo/foo-frontend` and `~/source/foo/foo-backend`.

This inconsistency between working directories breaks relative path lookups, since most methods in `java.io.File`, such as `exists()`, rely on `user.dir`.

brikis98:
BTW, our current workaround is to manually reset the `workingDirectory` at test time as follows:

``` scala
testGrouping in Test := {
  val original: Seq[Tests.Group] = (testGrouping in Test).value

  original.map { group =>
    val forkOptions = ForkOptions(
      bootJars = Nil,
      javaHome = javaHome.value,
      connectInput = connectInput.value,
      outputStrategy = outputStrategy.value,
      runJVMOptions = javaOptions.value,
      workingDirectory = Some(new File(System.getProperty("user.dir"))),
      envVars = envVars.value
    )

    group.copy(runPolicy = Tests.SubProcess(forkOptions))
  }
}
```

huntc:
Thanks!

jsuereth:
So, a few points:
[...]

PR #9442, opened 2026-07-12
[2.x] fix: Make forked run inherit sbt's working directory (#1032)

jozanek:
Ref #1032: forked `run` used the project's `baseDirectory` as the working directory, while non-forked execution inherits sbt's own working directory — so toggling `fork` silently changed how relative paths resolved.

Per review, this is scoped to `run` only: forked `run` (and forked `console`) now inherit sbt's working directory, consistent with non-forked execution and `sbtn` expectations. **Forked `test` is unchanged** and keeps the project's `baseDirectory`, since existing test suites in the wild rely on it. No new key: the working directory is configured through the existing `forkOptions`. The BSP `jvmRunEnvironment`, `buildTarget/run`, and `scalaMainClasses` adapters derive their launch policy from the effective `run / forkOptions` so they report the same contract.

To restore the old `run` behavior (see `notes/2.0.0/`):

```scala
Compile / run / forkOptions := Def.uncached(
  (Compile / run / forkOptions).value.withWorkingDirectory(Some(baseDirectory.value))
)
```

Developed with AI assistance (Claude Code), reviewed human-in-the-loop.

### DIFF
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -354,7 +354,6 @@ object Keys {
 
   val fork = settingKey[Boolean]("If true, forks a new JVM when running.  If false, runs in the same JVM as the build.").withRank(ASetting)
   val forkOptions = taskKey[ForkOptions]("Configures JVM forking.").withRank(DSetting)
-  val forkWorkingDirectory = settingKey[Option[File]]("Working directory of forked processes started by run, test, and console. If None, the forked process inherits the working directory of the sbt process.").withRank(BSetting)
   val outputStrategy = settingKey[Option[sbt.OutputStrategy]]("Selects how to log output when running a main class.").withRank(DSetting)
   val connectInput = settingKey[Boolean]("If true, connects standard input when running a main class forked.").withRank(CSetting)
   val javaHome = settingKey[Option[File]]("Selects the Java installation used for compiling and forking.  If None, uses the Java installation running the build.").withRank(ASetting)
--- a/main/src/main/scala/sbt/Keys.scala
+++ b/main/src/main/scala/sbt/Keys.scala
@@ -354,6 +354,7 @@ object Keys {
 
   val fork = settingKey[Boolean]("If true, forks a new JVM when running.  If false, runs in the same JVM as the build.").withRank(ASetting)
   val forkOptions = taskKey[ForkOptions]("Configures JVM forking.").withRank(DSetting)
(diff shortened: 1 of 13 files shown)


main/src/main/scala/sbt/Keys.scala
```diff
@@ -354,6 +354,7 @@ object Keys {
 
   val fork = settingKey[Boolean]("If true, forks a new JVM when running.  If false, runs in the same JVM as the build.").withRank(ASetting)
   val forkOptions = taskKey[ForkOptions]("Configures JVM forking.").withRank(DSetting)
+  val forkWorkingDirectory = settingKey[Option[File]]("Working directory of forked processes started by run, test, and console. If None, the forked process inherits the working directory of the sbt process.").withRank(BSetting)
```
eed3si9n: Given that we have `forkOptions` key, which already has the working directory information, I think adding a new key creates more confusion.

eed3si9n:
For forked tests, which is the complaint of https://github.com/sbt/sbt/issues/1032, the working directory has been set to subproject for so long, I don't know if we can change this without breaking existing tests in the wild.

We can consider the working directory for forked run and forked console, since that's a feature people hit more frequently in sbt 2.0 via sbtn.

jozanek:
@eed3si9n Agreed — scoped down: forked `test` unchanged, only forked `run` inherits sbt's working directory now.

main/src/main/scala/sbt/Keys.scala
jozanek: Dropped the key — `forkOptions` is the only surface now. Note: overriding it requires `Def.uncached(...)` (no `JsonFormat[ForkOptions]`).

eed3si9n (approved):
Thanks!

notes/2.0.0/fork-working-directory.md
```diff
@@ -0,0 +1,25 @@
...
+directory by default, consistent with non-forked execution and with `sbtn`
+expectations. Forked `test` is unchanged and keeps the project's `baseDirectory`
+as its working directory. The working directory of any forked process can be
+configured via `forkOptions`:
+
+```scala
+Compile / run / forkOptions := Def.uncached(
+  (Compile / run / forkOptions).value.withWorkingDirectory(Some(baseDirectory.value))
+)
+```
```
eed3si9n, line 18: I should've reviewed this part a bit more carefully. Given the existence of https://www.scala-sbt.org/1.x/docs/Forking.html, we probably want to use `baseDirectory` key like:

```scala
Compile / run / baseDirectory := baseDirectory.value
```

eed3si9n: https://github.com/sbt/sbt/issues/9530

Issue #9418, opened 2026-07-10
[2.x] Issues with `-java-home` flag if not using `--server`


gaeljw:
## steps

SBT version:
```
$ sbt --version
sbt version in this project: 2.0.1
sbt runner version: 1.12.4
```

CI environment (non interactive).

Trying to run `sbt -java-home /usr/lib/jvm/java-17 scalafmtCheckAll` (or any other command but I focused on this one for speed and low level of dependencies).

## problem

Without the `--server` flag, I get an error "failed to connect to server" but no detail (related to https://github.com/sbt/sbt/issues/9417).

With the `--server` flag, it works.

Here's the result of the command with different variants on a minimal project (happy to share it later if it can help ; commands are run in this same order a few seconds apart):

1/ with `--server` flag ✅ 

```
sbt --server -java-home /usr/lib/jvm/java-17 scalafmtCheckAll
[info] welcome to sbt 2.0.1 (Red Hat, Inc. Java 17.0.18)
[info] loading project definition from /opt/gitlab-runner/builds/pdN7c60XN/1/xxx/yyy/project
[info] set current project to category-dao-hbase (in build file:/opt/gitlab-runner/builds/pdN7c60XN/1/xxx/yyy/)
[success] elapsed time: 1 s, cache 100%, 3 disk cache hits
```

2/ without `--server` flag (`--debug` used) 💥 

```
sbt --debug -java-home /usr/lib/jvm/java-17 scalafmtCheckAll
[addSbt] arg = '-debug'
[residual] arg = 'scalafmtCheckAll'
[sbt_options] declare -a sbt_options=()
[debug] running native client
# Executing command line:
/opt/gitlab-runner/.cache/sbt/boot/sbtn/1.12.1/sbtn
--sbt-script=/usr/bin/sbt
--debug
-java-home
/usr/lib/jvm/java-17
scalafmtCheckAll

[info] server was not detected. starting an instance
[error] failed to connect to server
```

3/ without `--server` nor `--debug` 💥 

```
sbt -java-home /usr/lib/jvm/java-17 scalafmtCheckAll
[info] server was not detected. starting an instance
[error] failed to connect to server
```

4/ shutting down servers just in case (not sure what it really does)

```
sbt shutdown
[info] no sbt server is running. ciao
```

5/ Again with `--server` flag ✅  :

```
sbt --server -java-home /usr/lib/jvm/java-17 scalafmtCheckAll
[...]

Issue #9417, opened 2026-07-10
[2.x] "failed to connect to server" without more details


gaeljw:
## steps

- SBT version: 1.12.1
- Project build.properties: 2.0.1 (upgrading from 1.12.3)
- CI environment with limited permissions

## problem

When upgrading to sbt 2, all sbt command fail with an error "failed to connect to server" but no detail:

```
sbt --debug scalafmtCheckAll
[addSbt] arg = '-debug'
[residual] arg = 'scalafmtCheckAll'
[sbt_options] declare -a sbt_options=()
[debug] running native client
# Executing command line:
/home/gaeljw/.cache/sbt/boot/sbtn/1.12.1/sbtn
--sbt-script=/usr/bin/sbt
--debug
scalafmtCheckAll

[info] server was not detected. starting an instance
[error] failed to connect to server
```

Reading https://github.com/sbt/sbt/issues/8812, I learned about the `--server` flag that do gives more details on the error:

```
sbt --server --debug scalafmtCheckAll
[addSbt] arg = '-debug'
[residual] arg = 'scalafmtCheckAll'
[sbt_options] declare -a sbt_options=()
[process_args] java_version = '17'
[addMemory] arg = '1024'
[addJava] arg = '-Xms1024m'
[addJava] arg = '-Xmx1024m'
[addJava] arg = '-Xss4M'
[addJava] arg = '-XX:ReservedCodeCacheSize=128m'
[addJava] arg = '-Dsbt.script=/usr/bin/sbt'
[copyRt] java9_rt = '/home/gaeljw/.sbt/1.0/java9-rt-ext-red_hat__inc__17_0_18/rt.jar'
[addJava] arg = '-Dscala.ext.dirs=/home/gaeljw/.sbt/1.0/java9-rt-ext-red_hat__inc__17_0_18'
# Executing command line:
java
-Dfile.encoding=UTF-8
-Dsbt.override.build.repos=true
-Xms1024m
-Xmx1024m
-Xss4M
-XX:ReservedCodeCacheSize=128m
-Dsbt.script=/usr/bin/sbt
-Dscala.ext.dirs=/home/gaeljw/.sbt/1.0/java9-rt-ext-red_hat__inc__17_0_18
-jar
/usr/bin/sbt-launch.jar
-debug
scalafmtCheckAll

[...]

PR #9448, opened 2026-07-13
[2.x] fix: Fixes sbtn not propagating -java-home

BrianHotopp:
## Problem

The sbt 2.x thin client (sbtn) mishandles `-java-home` (#9418):

- `sbt --java-home=/path <cmd>` → `[error] Not a valid command: --`. The `=` form of a launcher value flag isn't recognized by `parseArgs`, so it falls through to the residual arguments and is forwarded to the server as a command.
- `sbt -java-home /path <cmd>` with no server running → `[error] failed to connect to server`. The space form *is* recognized, but it's consumed and discarded; when the client starts a server it re-invokes the `sbt` launcher without it, so the server comes up under the default JVM. In CI where the intended JDK is only reachable via `-java-home`, that server fails to start/connect.

`--server` works because in server mode the `sbt` launcher runs the server itself and applies `-java-home` directly; only the thin-client cold-start path drops it.

## Fix

`NetworkClient.parseArgs` now captures the launcher value flags it consumes (both `flag value` and `flag=value`, normalized to a `flag`/`value` pair) into a new `Arguments.launcherValueArgs`, and the cold-start server fork re-passes them to the `sbt` launcher (which parses `-java-home <path>` etc.). This covers all the launcher value flags (`-mem`, `-jvm-debug`, `-sbt-dir`, ...), which were dropped the same way.

- The `=` form is no longer forwarded to the server, so the `Not a valid command: --` error is gone.
- An empty `flag=` value is consumed but not propagated (the launcher's `require_arg` would otherwise fail the fork).
- The `sbt-launch-jar` path (already warned as unsupported) is unchanged: it invokes `java -jar` directly, so there is no launcher to interpret `-java-home` and the JVM is already chosen.

`Arguments` is `private[client]`; no public API change (MiMa clean).

## Testing

- `NetworkClientParseArgsTest` (Verify): pins that both forms are consumed (not leaked to the server), that `launcherValueArgs` holds the normalized flag/value pair for propagation, that the `=` form no longer leaks a bare `--`, and that an empty `flag=` is dropped. Existing parse tests unchanged and green (32 total).
- `commandProj/scalafmtCheckAll`, `commandProj/Test/scalafmtCheckAll`, `commandProj/mimaReportBinaryIssues` green.
- Release note: `notes/2.0.0/thin-client-java-home.md`.

The parse/capture contract and the launcher's space-form handling (`sbt` script `-java-home|--java-home) require_arg path`) are verified directly. I did not reproduce the reporter's CI end to end (it needs a second JDK on the box), but the fix targets the exact mechanism: the consumed flag never reaching the forked launcher.

The generic "failed to connect to server" with no detail (independent of `-java-home`) is tracked separately in #9417.

---

*Implemented with AI assistance (Claude Code); reviewed and verified locally by me (unit tests, scalafmt, MiMa).*

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -376,33 +376,19 @@ class NetworkClient(
             term.isSupershellEnabled
           ).mkString(",")
 
-        val cmd = arguments.sbtLaunchJar match {
-          case Some(lj) =>
-            if (log) {
-              val sbtScript = if (Properties.isWin) "sbt.bat" else "sbt"
-              console.appendLog(Level.Warn, s"server is started using sbt-launch jar directly")
-              console.appendLog(
-                Level.Warn,
-                "this is not the recommended way: .sbtopts and .jvmopts files are not loaded and SBT_OPTS is ignored"
-              )
-              console.appendLog(
-                Level.Warn,
-                s"either upgrade $sbtScript to its latest version or make sure it is accessible from $$PATH, and run 'sbt bspConfig'"
-              )
-            }
-            val java = Option(Properties.javaHome)
-              .map { javaHome =>
-                s"$javaHome/bin/java"
-              }
-              .getOrElse("java")
-            List(java) ++ arguments.sbtArguments.filterNot(
-              NetworkClient.emptyBuildFlags.contains
-            ) ++
-              List("-jar", lj, DashDashDetachStdio, DashDashServer)
-          case _ =>
-            List(arguments.sbtScript) ++ arguments.sbtArguments ++
-              List(DashDashDetachStdio, DashDashServer)
+        if (log && arguments.sbtLaunchJar.isDefined) {
+          val sbtScript = if (Properties.isWin) "sbt.bat" else "sbt"
+          console.appendLog(Level.Warn, s"server is started using sbt-launch jar directly")
+          console.appendLog(
+            Level.Warn,
+            "this is not the recommended way: .sbtopts and .jvmopts files are not loaded and SBT_OPTS is ignored"
+          )
+          console.appendLog(
+            Level.Warn,
+            s"either upgrade $sbtScript to its latest version or make sure it is accessible from $$PATH, and run 'sbt bspConfig'"
+          )
         }
+        val cmd = NetworkClient.serverCommand(arguments)
 
         // https://github.com/sbt/sbt/issues/6271
         val nohup =
@@ -1204,6 +1190,7 @@ object NetworkClient {
       val sbtScript: String,
       val bsp: Boolean,
       val sbtLaunchJar: Option[String],
+      val launcherValueArgs: Seq[String] = Nil,
   ) {
     def withBaseDirectory(file: File): Arguments =
       new Arguments(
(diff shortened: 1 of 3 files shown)


main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -1293,6 +1295,8 @@ object NetworkClient {
     "--autostart=",
     "-autostart=",
   )
+  private[client] val launcherValueEqPrefixes: Seq[String] =
+    launcherValueFlags.toSeq.map(_ + "=")
   private[client] def parseArgs(args: Array[String]): Arguments = {
```
anatoliykmetyuk, line 1297: Will this still work properly on space-contaning paths, e.g. `-java-home "C:\Program Files\Java\jdk-17"`?

main-command/src/test/scala/sbt/internal/client/NetworkClientParseArgsTest.scala
```diff
@@ -196,7 +197,38 @@ object NetworkClientParseArgsTest extends BasicTestSuite:
...
+    val result = parse("-mem=10000", "compile")
+    assert(!result.sbtArguments.exists(_.contains("mem")))
+    assert(result.launcherValueArgs == Seq("-mem", "10000"))
+    assert(result.commandArguments.contains("compile"))
+
+  test("--java-home= with an empty value is consumed but not propagated"):
+    val result = parse("--java-home=", "compile")
+    assert(!result.sbtArguments.exists(_.contains("java-home")))
+    assert(result.launcherValueArgs.isEmpty)
+    assert(result.commandArguments == Seq("compile"))
```
anatoliykmetyuk, line 232: The parsing tests are nice but would be good to also have integration tests proving the flag actually reaches the newly started server.

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
BrianHotopp: Good catch, and confirmed it did mangle. The client tokenizes args by splitting on whitespace, so an unquoted `C:\Program Files\...` value got fragmented, and because this PR now propagates the value to the forked server, that surfaced. Fixed in 2cded9f8: `parseArgs` tracks the whitespace-split boundaries and rejoins a value flag's value, so `-java-home "C:\Program Files\Java\jdk-17"` and the `=` form are both preserved. Added tests for both forms plus the dangling-flag and stray-whitespace edges.

main-command/src/test/scala/sbt/internal/client/NetworkClientParseArgsTest.scala
BrianHotopp: Added. I extracted the fork command construction into `serverCommand`, and the test asserts the propagated `-java-home <path>` appears in the command used to start the server, ahead of `--server`, so it pins the propagation contract deterministically. I didn't go as far as a scripted test that forks a real server under a second JDK, since that needs a second JDK at a known path in the test environment and exercises process startup rather than the parsing/propagation this PR changes. Happy to add one if you'd prefer the full round-trip.

anatoliykmetyuk (approved):
LGTM, thanks @BrianHotopp

gaeljw:
Thanks a lot for the work. This is not included in sbt 2.0.3 yet, right? I'm not familiar with the release process: do you know when will this be included in a 2.0.x release?

BrianHotopp:
Correct, it just missed 2.0.3: that release was cut before this PR merged (both happened on July 16). The fix will be in the next 2.0.x release; there is already a 2.0.4 milestone. Recent patch releases have been coming out roughly every one to two weeks, but the exact timing is up to the maintainers. In the meantime, the `--server` flag workaround from the issue keeps working on 2.0.3.

eed3si9n:
2.0.3 included (minor) CVE fixes, so I didn't want to overload it with other things. Like @BrianHotopp says, we tend to release a patch release every 1 ~ 2 week. I think that's the max we can do -- otherwise, ppl would get tired from automatic update bumping up sbt version constantly.

gaeljw:
Great, thanks for the answers.
I just wanted to be sure it was not included as I was still reproducing 😅

PR #9449, opened 2026-07-13
[2.x][WIP] Support more Trees in build.sbt

xuwei-k:
Should we support more trees in `build.sbt`? Are there any reasons to restrict only `def`, `val` and [`given`](https://github.com/sbt/sbt/pull/9409)?

What do you think?


eed3si9n:
> Are there any reasons to restrict only `def`, `val` and `given`?

`build.sbt` DSL is a mini language for the purpose of declaring the build, and I think it should be as simple as possible. Something like [Makefile](https://www.gnu.org/software/make/manual/html_node/Variables-Simplify.html) or [toml file](https://doc.rust-lang.org/cargo/guide/dependencies.html). If we need full Scala syntax, there's always `project/*.scala`.

There are a few reasons for aiming for _the aesthetics of subtraction_.
1. User load and education. `build.sbt` needs to support a wide range of users, including new users to Scala. Even the intermediate users might not fully understand `build.sbt`, so we'd like `build.sbt` to be something you can guess the meaning without understanding everything about sbt.
2. Maintainability. sbt releases are maintained over a long time, and as a project we have to bridge across multiple versions of Scala. Exposing fewer elements of Scala to `build.sbt` makes maintenance easier, and protects us from future changes in the language.

keynmol:
One thing that comes to mind is that Scala 3 is more strict around a particular pattern I enjoyed/stole in the wild

```
val Versions = new {
  val OsLib = "0.13.8"
}
```


unfortunately that breaks in sbt2 because Scala 3 does not refine the inferred type (it stays as Object)

Only way I found to do it is via Selectable, but it's more verbose:

```scala
import scala.reflect.Selectable.reflectiveSelectable

val Versions = new Selectable {
  val OsLib = "0.11.8"
}
```

And I don't want to create a separate project/Versions.scala file just to be able to use an `object`.

SethTisue:
the upshot being, could `object` specifically be allowed...?

eed3si9n:
> the upshot being, could `object` specifically be allowed...?

I have been thinking about this, and came close to drafting an RFC, but stopped short. As a proponent of `Dependencies` object, I'd love for us to be able to declare a simple object with versions and ModuleID variables in it.
However, `object` would be a well-lubricated slippery slope.

Beyond simple objects that we ideate, the build users would be able to construct all sorts of things, including ad-hoc plugins, ad-hoc subprojects, nested classes, etc. We got rid of `project/Build.scala` when we released sbt 1.0, which simplified the maintenance and unified how the build looked across the community (while there were many dissenting power users). The goal of a statically typed DSL is to make nonsensical values inexpressible (or "make illegal states unrepresentable"), and `object` is too open I think.

I'm open to further conversations if ppl are coming to Berlin for Scala Days.

SethTisue:
> build users would be able to construct all sorts of things, including ad-hoc plugins, ad-hoc subprojects, nested classes

To check my understanding: I can't put any of that stuff inside a `def` or a `val`...?

eed3si9n:
Not that I know of.

PR #9450, opened 2026-07-14
[2.x] fix dependency parsing for bsp

### LINKED ISSUE #9378: [BSP] bspInternalDependencyConfigurations drops test->test inter-project dependencies
**AUTHOR NOTE:**

I came across this issue in our codebase while trying to use
SBT as a BSP for Metals. Right now only Bloop works but we would solve a lot of
inconsistencies for the plugins we use with SBT if we could simply rely on one
build server. I'll admit I'm out of my depth when understanding the BSP
protocol and SBT internals and this debugging investigation was largely done by
Claude Opus 4.8. So, I'll apologize for the sloppy AI prose around concepts I
don't quite understand. I've tried to tame it as much as I can without spending
gobs of time and hope that there is enough context here that maintainer
understands what needs to be done. I am willing to do more research and
debugging to understand and fix this further but wanted to get the ball
rolling. Thanks!

## Summary
On (at least) sbt 1.12.13, a build target reachable via an inter-project `test -> test` 
dependency gets an incomplete classpath from the BSP server.
`buildTarget/scalacOptions` for that target omits the depended project's
**main** classes dir, its test-classes dir, and transitive project
dependencies, even though the real `Test/compile` classpath 
(and `sbt <proj>/Test/dependencyClasspath`) includes them and compilation succeeds.
Targets that depend through `compile -> compile` are unaffected.

As far as I was able to figure out with AI assistance the defect is localized
to the `bspInternalDependencyConfigurations` setting, which drives the BSP
`scalacOptions` classpath. For the `test->test` target it contains only the
target's own project, not the depended project's configs.

This seems like a regression / incomplete fix of #6410 / #6414 (which fixed the
direct `test->test` case in 1.5.0), on 1.12.13 even the direct `(dep, test)`
entry is missing. See also #8249.

In Metals, when using the sbt bsp, this surfaces as pervasive false "unresolved
symbol" errors when editing test files that reference a project pulled in via
`test -> test`. 

Bloop is unaffected because `bloopInstall` pre-flattens the complete classpath.

## Reproduction

Claude helped me come up with a minimal reproduction based on the issue

Minimal reproduction: [sbt-bsp-transitive-classpath-bug-repro](https://github.com/jsilasbailey/sbt-bsp-transitive-classpath-bug-repro). 

Four projects: 
- `protos`
- `core` (`dependsOn(protos)`, has main+test)
- `it` (`dependsOn(core % "test -> test")`)
- `contrast` (`dependsOn(core)`).

```scala
lazy val protos   = project
lazy val core     = project.dependsOn(protos)
lazy val it       = project.dependsOn(core % "test -> test")   // <- BUG target
lazy val contrast = project.dependsOn(core)                    // <- contrast (works)
```

```bash
# Real build is fine and its classpath is complete:
sbt it/Test/compile                                    # [success]
sbt 'show it/Test/dependencyClasspath'                 # includes core/classes, core/test-classes, protos/classes

# But the BSP-internal dependency set drops the test->test deps:
sbt 'show it/Test/bspInternalDependencyConfigurations'
#   -> * (ProjectRef(.../it), ListSet(compile, test))          # only `it`; core + protos MISSING

sbt 'show contrast/Test/bspInternalDependencyConfigurations'
#   -> * (ProjectRef(.../contrast), ListSet(compile, test))
#      * (ProjectRef(.../core),     ListSet(compile))          # present
#      * (ProjectRef(.../protos),   ListSet(compile))          # present
```

## Expected

`it/Test/bspInternalDependencyConfigurations` should include `(core, test)`,
and transitively `(core, compile)` and `(protos, compile)`, matching
`it/Test/dependencyClasspath`. The resulting `buildTarget/scalacOptions`
classpath for `#it/Test` should then contain core's main + test classes and
protos's classes.

## Actual

`it/Test/bspInternalDependencyConfigurations` contains only `(it, {compile,test})`. The
exported `scalacOptions` classpath for `#it/Test` is correspondingly missing
`core/classes`, `core/test-classes`, and `protos/classes`. (`contrast/Test`, a
`compile->compile` dep, has them.)

## Environment

- sbt 1.12.13
- Scala 2.13.18
- JDK 17
- macOS
- Build server: sbt's built-in BSP (`sbt bspConfig`).

## Why sbt-side (not the BSP client)

**NOTE:** Would love any validation if there is a client option that solves this
and things are working as expected but this is what Claude came up with

Metals advertises `jvmCompileClasspathReceiver: true` but only sends
`buildTarget/jvmCompileClasspath` when the server advertises
`jvmCompileClasspathProvider`. 

sbt advertises neither and never serves it, so Metals correctly falls back to
the `scalacOptions` classpath, which is itself incomplete. The classpath is
assembled in `BuildServerProtocol.scala` from
`bspInternalDependencyConfigurations` + `externalDependencyClasspath`; the
`test` config's expansion doesn't include the depended project's configs.

## Related

- [scalameta/metals#2625]
- [sbt#6410]
- [sbt#6414]
- [sbt#8249]
- [scalameta/metals#7623]

[scalameta/metals#2625]: https://github.com/scalameta/metals/issues/2625
[sbt#6410]: https://github.com/sbt/sbt/issues/6410
[sbt#6414]: https://github.com/sbt/sbt/issues/6414
[sbt#8249]: https://github.com/sbt/sbt/issues/8249
[scalameta/metals#7623]: https://github.com/scalameta/metals/issues/7623

anatoliykmetyuk:
Fix #9378 for sbt 2

Do parsing for bsp the same way as done for compilation: https://github.com/sbt/sbt/blob/2d2d5136efdae62cbd971546781081e7ad582cc3/main/src/main/scala/sbt/internal/ClasspathImpl.scala#L422

### DIFF
--- a/main/src/main/scala/sbt/internal/InternalDependencies.scala
+++ b/main/src/main/scala/sbt/internal/InternalDependencies.scala
@@ -22,9 +22,9 @@ private[sbt] object InternalDependencies {
     ((ref -> allConfigs) +:
       projectDependencies.flatMap { case ClasspathDep.ResolvedClasspathDependency(p, rawConfigs) =>
         val configs = rawConfigs.getOrElse("*->compile").split(";").flatMap { config =>
-          config.split("->") match {
-            case Array(n, c) if applicableConfigs.contains(n) => Some(c)
-            case Array(n) if applicableConfigs.contains(n)    =>
+          config.split("->", 2).map(_.trim) match {
+            case Array(n, c) if c.nonEmpty && applicableConfigs.contains(n) => Some(c)
+            case Array(n) if applicableConfigs.contains(n)                  =>
               // "test" is equivalent to "compile->test"
               Some("compile")
             case _ => None
--- a/sbt-app/src/sbt-test/project/bsp-internal-dependency-configs/build.sbt
+++ b/sbt-app/src/sbt-test/project/bsp-internal-dependency-configs/build.sbt
@@ -4,6 +4,10 @@ lazy val b = project.in(file("b")).dependsOn(c)
 
 lazy val c = project.in(file("c"))
 
+lazy val d = project.in(file("d")).dependsOn(b % "test -> test")
+
+lazy val e = project.in(file("e")).dependsOn(b % "test ->")
+
 def getConfigs(key: SettingKey[Seq[(ProjectRef, Set[ConfigKey])]]):
   Def.Initialize[Map[String, Set[String]]] =
     Def.setting(key.value.map { case (p, c) => p.project -> c.map(_.name) }.toMap)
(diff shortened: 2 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #9452, opened 2026-07-14
[2.x] restore JVM option capability in the launcher config

### LINKED ISSUE #9369: [2.x] csrResolvers in meta-build no longer honors proxy repository configuration
## steps

sbt-launch jar: 2.0.0
sbt version: 1.x (e.g., 1.12.11)

## problem

In certain cleanroom build environments (e.g., corporations), egress to the external Internet is strictly prohibited. In these scenarios, sbt 1.x encouraged use of a `repositories` file to specify internal resolvers and the following sbt option: `sbt.override.build.repos=true`. With the release of sbt 2.0.0, such builds that pick up the latest sbt-launch.jar 2.0.0 may regress to reaching out to the Internet, thereby leading to frustrating build delays/failures.

In particular, the meta-build no longer seems to honor those aforementioned proxy repo configurations. Instead, the meta-build alone (i.e., not the primary build) is adding in external `csrResolvers`. Placing this line in `project/build.sbt` removes them, but shouldn't be necessary:
```
csrResolvers ~= (_.filterNot(r => Set("public", "sbt-plugin-releases", "typesafe-alt-ivy-releases").contains(r.name)))
```

## expectation

The sbt-launch.jar 2.0.0 meta-build `csrResolvers` should continue to honor the behavior from sbt 1.x with proxy repository configurations.

anatoliykmetyuk:
Fixes https://github.com/sbt/sbt/issues/9369

Second part of the fix, first part is updating the launcher: https://github.com/sbt/launcher/pull/148, with more details on the fix logic.

Restores `override-build-repos` config option responsible for passing the JVM flag to the launcher. In principle, restoring this config line is enough to fix the issue. However, when restored, it will shadow the new `repository-override`. `repository-override` is a feature to override repos via a file, was shipped in 2.0.0 and therefore we should continue supporting it. To continue supporting it is why the first part - updating the launcher, https://github.com/sbt/launcher/pull/148 - is needed.

Once the launcher is updated and released, we should update this PR bumping the version of the launcher to the newly released one. Without the update, the new tests are expected to cause CI failure.


eed3si9n:
https://repo1.maven.org/maven2/org/scala-sbt/launcher/1.6.2/ is released.

eed3si9n (approved):
Thanks!

Issue #9455, opened 2026-07-15
[2.x] Transient I/O compile failure is cached by the action cache and replayed forever, wedging the build


ConradHughes:
Human-hand-checked LLM diagnosis and repro of problem encountered during normal dev work; not offering a PR fix because there are several legitimate solutions and I suspect you'd prefer to make that choice.

## steps

sbt 2.0.2, Java 25, Linux. Two-file project; no plugins.

```bash
mkdir -p repro/project repro/src/main/scala && cd repro
printf 'name := "repro"\n' > build.sbt
echo 'object Meta' > project/Meta.scala
echo 'object Main { def main(a: Array[String]): Unit = println("hi") }' > src/main/scala/Main.scala

sbt compile          # succeeds
sbt cleanFull        # purge local caches, so the next cold start truly compiles
sbt shutdown

# Simulate a transient I/O failure during the metabuild compile. (In the wild
# this was a concurrent process deleting target/ mid-compile, giving
# NoSuchFileException instead of AccessDeniedException; any I/O failure does.)
# Note the metabuild output dir is named <projectdirname>-build.
mkdir -p target/out/jvm/scala-3.8.4/repro-build/classes
chmod 555 target/out/jvm/scala-3.8.4/repro-build/classes

sbt compile          # fails as expected: "error writing .../repro-build/classes/Meta$.class: AccessDeniedException"
sbt shutdown

# Heal the environment completely:
chmod 755 target/out/jvm/scala-3.8.4/repro-build/classes

sbt compile          # BUG: fails again with the identical messages — without compiling anything
```

The final `sbt compile` (and every one after it, across server restarts and reboots) fails instantly during project loading:

```
[info] loading project definition from .../repro/project
[error] error writing .../repro-build/classes/Meta$.class: java.nio.file.AccessDeniedException ...
[error] error writing .../repro-build/classes/Meta.class: java.nio.file.AccessDeniedException ...
[error] error writing .../repro-build/classes/Meta.tasty: java.nio.file.AccessDeniedException ...
[error] sbt.util.CachedCompileFailure$$anon$1: Compilation failed
[error] 	at sbt.util.CachedCompileFailure.toException(CachedCompileFailure.scala:23)
[error] 	at sbt.util.ActionCache$.cache(ActionCache.scala:181)
...
[error] (Compile / compileIncremental) sbt.util.CachedCompileFailure$$anon$1: Compilation failed
[error] failed to connect to server
```

[...]

PR #9464, opened 2026-07-16
[2.x] fix: Fixes system errors getting cached as compilation errors

BrianHotopp:
## Problem

Failure caching (#8490, fixing #7662) caches a `CompileFailed` so an unchanged, still-broken compile does not re-run from scratch. That assumes a `CompileFailed` is a function of the sources, which holds for source errors. But zinc also surfaces I/O write failures ("error writing X.class") as compiler problems, so an environmental failure gets cached under the same mechanism and replayed from the global action cache (`~/.cache/sbt/v2`) on every later build, even after the cause is gone.

When the poisoned task is the metabuild compile, it is self-sustaining and unrecoverable from inside sbt: project loading fails, so no task (including `clean`/`cleanFull`) can run, and deleting `target/` does not help because the poison lives in the global cache. Only `rm -rf ~/.cache/sbt/v2` recovers. The replayed diagnostics also mislead, naming files and permissions that no longer exist.

Reported and cleanly reproduced in #9455.

## Fix

Scala source errors carry a source position; the "error writing" I/O errors do not (they print with no `file:line:col`). `CachedCompileFailure.hasSufficientInfo`, which was dead code (never wired in) with a display-oriented rendered-or-position check, now gates the failure cache on both sides:

- **Write**: a `CompileFailed` is cached only when it has at least one error and every error carries a source position. Position-less failures are treated as environmental and left uncached, so the next build retries for real. Positioned source failures are still cached, so the #7662 optimization is unchanged.
- **Replay**: a cached failure that fails the same check -- i.e. one written by an earlier sbt -- is not replayed; the task re-runs and a success overwrites the stale entry. Already-poisoned caches (local or shared remote) self-heal on upgrade instead of requiring manual cache deletion, and clients still running older sbt against a shared cache cannot poison upgraded ones.

(Note: wiring up the old body would not have fixed this on Scala 3, where every problem carries `rendered` text; the gate has to key on source position, not on rendered/message presence.)

## Known limitations (called out deliberately)

- **javac**: unlike scalac/dotty, javac attributes its class-write I/O errors to the class declaration's source position, so the javac variant of an environmental failure still passes the gate and is cached. This PR covers the scalac/dotty shape from the report and is strictly no worse than the status quo for javac; message-pattern matching is not viable (javac diagnostics are localized). Happy to file a follow-up issue for the javac shape if you want it tracked.
- **-Werror**: a fatal-warnings failure emits positioned warnings plus one position-less summary error, so it is no longer cached -- a lost optimization only (it re-runs and fails identically with fresh diagnostics). This shape is indistinguishable from "positioned warning + position-less I/O error", so the conservative call is the only safe one.

## Testing

- `ActionCacheTest` (unit, 16/16): position-less failure is not cached (action re-runs); a hand-written pre-fix poisoned entry is not replayed and is overwritten by the subsequent success (the upgrade-cure path); the existing positioned-failure test still pins that source failures remain cached.
- `cache/compile-io-failure` (scripted, end-to-end): makes the compile output dir read-only to force the real `error writing X.class: AccessDeniedException` I/O failure, then heals it and asserts the next compile recompiles instead of replaying the cached failure. Confirmed to fail on develop (replays a `cached-failure` cache hit, the exact #9455 stack trace) and pass with this change. Windows-excluded since a read-only dir does not block writes there.
- Full `cache/*` scripted group green; `utilCache` scalafmt, `scalafmtSbtCheck`, and MiMa clean.

The reporter noted several legitimate solutions; this takes the position-based discriminator they suggested. Happy to adjust (for example, annotating a replayed cached failure as cache-sourced with a recovery hint) if you would prefer a different approach.

---

*Diagnosed and fixed with AI assistance (Claude Code); validated locally (unit tests, end-to-end scripted repro, full cache group, scalafmt, MiMa). The javac/-Werror behavior was verified empirically against javac 21, Scala 2.13.16, and Scala 3.8.4.*

### DIFF
--- a/notes/2.0.0/cache-environmental-compile-failure.md
+++ b/notes/2.0.0/cache-environmental-compile-failure.md
@@ -1,0 +1,18 @@
+### Environmental compile failures are no longer cached
+
+sbt caches a `CompileFailed` so that an unchanged, still-broken compile does not
+re-run from scratch (#7662). But zinc also reports I/O write failures ("error
+writing X.class") as compiler problems, so a one-off environmental failure (a
+concurrent `target/` deletion, a permission blip) was cached under the same
+mechanism and replayed on every later build, even after the cause was gone. When
+it hit the metabuild compile the build stayed wedged across restarts, recoverable
+only by deleting the global cache by hand. Compile failures whose errors carry no
+source position are now treated as environmental and left uncached, so the next
+build retries for real. A position-less failure already sitting in a cache
+written by an earlier sbt is likewise no longer replayed: the task re-runs and a
+success overwrites the stale entry, so previously wedged builds recover on
+upgrade without deleting the cache.
+
+This addresses [#9455][i9455].
+
+[i9455]: https://github.com/sbt/sbt/issues/9455
--- a/project/Scripted.scala
+++ b/project/Scripted.scala
@@ -40,9 +40,11 @@ object Scripted {
           case ("classloader-cache", "jni")   => true // no native lib is built for windows
           case ("classloader-cache", "spark") =>
             true // the test spark server is unable to bind to a local socket on Visual Studio 2019
-          case ("nio", "make-clone") => true // uses gcc which isn't set up on all systems
-          case ("watch", "symlinks") => true // symlinks don't work the same on windows
-          case _                     => false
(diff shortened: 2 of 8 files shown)


eed3si9n (approved):
Thanks!

PR #9469, opened 2026-07-17
[2.x] fix: Fixes tests not detecting stale resources

### LINKED ISSUE #9468: [2.x] test / testFull seem to report cached results after editing a test resource file
**sbt version:** 2.0.3
**Scala version:** 3.8.4

**Description**

I noticed that after changing only a test resource file (under `src/test/resources`, no Scala source changes), test keeps reporting the old result instead of re-running. 

`testFull` seemed to show the same behavior. Not 100% sure if this is a caching bug or something on my end.

Even when changing the test source, it does recompile to match against the changed string, but the ressources is not updated.

**Reproducer**

https://github.com/tpetillot/sbt-test-cache-reproducer

1. sbt
2. test — passes (hello.txt contains hello)
3. Edit `src/test/resources/hello.txt` to something else
4. test again — still shows success instead of failing
5. testFull — same result

**Expected**

Should fail, since the resource no longer matches what the test asserts.

---

**Side-note:** it _doesn't_ seem to be only cache related, because when I delete `~/Library/Caches/sbt`, it's still occuring sometimes ?

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/9468

## Problem
sbt 2.x currently constructs the test classloader using symlinks to the JARs in CAS (content-addressable storage).
This causes issues because JVM apparently caches opened JAR files by path, and since the symlink itself doesn't change it could end up serving stale resource files.

## Solution
We can workaround this issue by resolving the symlinks to the real path.

## Test

```scala
sbt:sbt-cache-reproducer> test
[info] [root] compiling 1 Scala source to /private/tmp/sbt-test-cache-reproducer/target/out/jvm/scala-3.8.4/sbt-cache-reproducer/test-classes ...
[info] ReproducerSpec:
[info] hello.txt
[info] - should contain "hello"
[info] Run completed in 278 milliseconds.
[info] Total number of tests run: 1
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 1, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
[success] elapsed time: 13 s, cache 44%, 13 disk cache hits, 16 onsite tasks

# edit hello.txt

sbt:sbt-cache-reproducer> test
[info] ReproducerSpec:
[info] hello.txt
[info] - should contain "hello" *** FAILED ***
[info]   "[goodbye]" was not equal to "[hello]" (ReproducerSpec.scala:10)
[info] Run completed in 119 milliseconds.
[info] Total number of tests run: 1
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 0, failed 1, canceled 0, ignored 0, pending 0
[info] *** 1 TEST FAILED ***
```


tpetillot (approved):
Thanks for digging into this! Nice work finding the root cause here, this was a subtle one!

PR #9471, opened 2026-07-18
[2.x] fix: Fixes cache invalidation on version change

### LINKED ISSUE #9454: [2.x] Cache is invalidated on version change
## steps

I'm not quite sure this is a bug per se, but I would like to confirm whether this is expected behavior.

Since `exportJars := true` became the default in sbt 2.x, jar names are included in hash keys for the new remote caching functionality, but those jar names typically contain the version of the module as well.

This means that if you use something like `sbt-dynver` (which would compute new versions based on commits), most builds would have cache misses even if the module in question didn't change.

That said, even the `MANIFEST` file in the jar could contain the version, so maybe that'll mess with caching too 🤔 

1. Is this behavior intended?
2. If yes, what is the recommendation for getting the most out of caching? Only change versions on publish?

eed3si9n:
FIxes https://github.com/sbt/sbt/issues/9454

## Problem
`packageBin` includes version into the file name, which ends up invalidating the cache.

## Solution
Define `packageInternal`, which does not include version in the file name, and used during `Compile` or `Test/compile`. Note that publishing and `Runtime` classpath would continue to use `packageBin`.


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1797,6 +1797,8 @@ object Defaults extends BuildCommon with DefExtra {
       packageTaskSettings(packageBin, packageBinMappings) ++
       packageTaskSettings(packageSrc, packageSrcMappings) ++
       packageTaskSettings(packageDoc, packageDocMappings) ++
+      packageTaskSettings(packageInternal, packageBinMappings) ++
```
anatoliykmetyuk: Should `packageInternal / mappings` inherit from `packageBin / mappings` instead of being defined independently? Existing builds and plugins which customize `Compile / packageBin / mappings` may be affected as the jars produced by `packageInternal` will miss those customizations.

sbt-app/src/sbt-test/actions/compile-clean/test
```diff
@@ -13,5 +13,5 @@ $ absent target/**/compile-clean/test-classes/B.class
 
 > Compile/clean
 $ exists target/**/compile-clean/classes/cant-touch-this
-$ absent target/**/compile-clean/classes/A.class
+# $ absent target/**/compile-clean/classes/A.class
```
anatoliykmetyuk: What is the motivation for this change?

main/src/main/scala/sbt/internal/ClasspathImpl.scala
```diff
@@ -194,11 +210,18 @@ private[sbt] object ClasspathImpl {
       log: Logger
   ): Initialize[Task[Classpath]] =
     Def.value[Task[Classpath]] {
-      interDependencies(projectRef, deps, conf, self, data, track, false, log)(
-        exportedProductsNoTracking,
-        exportedProductsIfMissing,
-        exportedProducts
-      )
+      if self == Runtime then
```
anatoliykmetyuk, line 222: Just to confirm, only `Runtime / dependencyClasspath` exposes versioned jars, the ordinary `run` task depends on `Runtime / fullClasspathAsJars` which exposes non-versioned jars - is that intended behavior? To me it makes sense, as the primary objective is to preserve compatibility with plugins, under the assumption that plugins consume `Runtime / dependencyClasspath` and do not consume `Runtime / fullClasspathAsJars`. It doesn't really matter which jars `run` itself consumes, so it's good that the impact area of the change is reduced.

anatoliykmetyuk (commented):
Looks good to me! Left some comments. The change seems minimal and well-bounded, yet solving the complex issue.

main/src/main/scala/sbt/internal/ClasspathImpl.scala
eed3si9n: > is that intended behavior?

No the, XAsJar variants were an oversight. Unless there's some other reason I think XAsJar variants should match in semantics with the regular classpaths. So `run` I think would use the versioned JARs.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Yea, that sounds like a good idea rather than repeating `packageBinMappings`.

sbt-app/src/sbt-test/actions/compile-clean/test
eed3si9n: Apparently I made a separate commit when I sent this PR a month ago, but I don't remember why. I'll revert it.

eed3si9n: Addressed in 86494a41bc65a391bd7810c4af22cd09894a6289

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Addressed in 7e3295d491ee8907fba79533eee90dabdc7a59cd

main/src/main/scala/sbt/internal/ClasspathImpl.scala
eed3si9n: Addressed in 5884066465d82219daec02736850c9149d91e111

anatoliykmetyuk (approved):

Issue #9462, opened 2026-07-16
[2.0.2] sbt compile caching broken + tracking down the Def.declareOutput & Def.declareOutputDirectory issues


msosnicki:
## steps

All the steps and scenarios are recreated in this repo: https://github.com/msosnicki/sbt2-declareoutput-restoration-bug

Have a look at the `vanilla` project for the `sbt compile` caching issues. The other examples are just reproducing the issues in the smaller examples. See README for more details on each scenario and how to reproduce

eed3si9n:
@msosnicki Thanks for the bug report with a detailed repro.

eed3si9n:
Looking at this more closely since there's a PR by @BrianHotopp (thanks!)

```
sbt --server "declareOutputDirectoryExample"     # miss - writes a.txt, b.txt
ls target/out/jvm/scala-3.8.4/sbt2-declareoutput-restoration-bug/declared-output-dir
rm -rf target/out/jvm/scala-3.8.4/sbt2-declareoutput-restoration-bug/declared-output-dir
sbt --server "declareOutputDirectoryExample"     # hit - does not run the task again
```

I don't know what the real-life scenario that causes these things, but any files under `target/` directory is under management by sbt. This is equivalent of opening the hood of a car, removing random wires, and finding out that it no longer stops correctly. I'm not sure if it's a good tradeoff to perform file stat call on hundreds of files under `target/`, and slowing down the build for everyone, just in case someone may have run `rm -rf target/random/dir`.

From the PR:

> Warm-path cost, measured on a 3000-file, 579KB dirzip: the existing full-zip digest check is ~1.9ms per hit; the added manifest read + stats is ~13.7ms. If that is too much for large classes directories, the cheaper fallback is a directory-exists-only check (one stat, covers the reported rm -rf case, misses partial deletions) — happy to switch if preferred.

I'm leaning towards cheaper fallback with one stat.

BrianHotopp:
Switched to the one-stat check in 8c83a27: the up-to-date branch now re-extracts only when the extracted directory itself is missing, and the manifest-based check is gone. Partial deletions inside a still-existing directory are not repaired, consistent with treating `target/` contents as managed. PR body updated to match.

PR #9473, opened 2026-07-18
[2.x] fix: Improves cache restoration when the expected outputs are missing

BrianHotopp:
## Problem

A directory declared with `Def.declareOutputDirectory` is packaged as a sibling `<dir>.sbtdir.zip`. Deleting the directory (an `rm -rf` of `target/` or of the classes directory) leaves the zip behind, and on the next cache hit `syncFile`'s up-to-date short-circuit sees the zip in sync (same digest, already a CAS symlink) and returns without the unpack side effect, which only runs from the file-write path. The directory is never restored.

sbt's own `compile` declares its classes directory this way (`cachedCompileIncrementalTask`), so a warm cache plus a deleted output directory means `run` fails with `ClassNotFoundException` and no recompile — reported with a clean repro in #9462 (cases 3 and 6 of the reporter's repo). Only deleting the zip as well, or the whole cache, recovers.

## Fix

On the up-to-date branch, for dirzip refs, re-extract when the extracted directory itself is missing: a single stat on the warm path (per review discussion on #9462, in preference to a manifest-based per-file check). Partial deletions inside a still-existing directory are not repaired, consistent with treating `target/` contents as sbt-managed; the full unpack (which also repairs stale content by digest and removes extraneous files) still runs whenever the zip itself needs re-syncing.

## Semantics and cost, stated deliberately

- Files that exist but whose content was modified by hand are not repaired on the warm path (the warm path never verified extracted content, only the zip digest; repairing would mean hashing every file on every hit). The full unpack still repairs by digest whenever it runs.
- The warm fast path also skips the extraneous-file deletion pass; files not in the manifest are removed whenever a restore is actually triggered.
- A corrupt or unreadable zip now fails closed: the presence check returns false and the full unpack runs, which either repairs or fails loudly, rather than silently leaving the directory unrestored.

## Testing

- `ActionCacheTest` (unit): a dirzip synced, its extracted directory deleted, `syncBlobs` again re-extracts both files. (The test needed a byte-faithful `VirtualFile` converter; the existing String-based test converter corrupts binary zips.)
- `cache/declare-output-dir-restore` (scripted): a custom `Def.declareOutputDirectory` task; delete the dir; cache hit must restore it (also covered with the zip deleted too). Fails on develop, passes with this change.
- `cache/compile-classes-restore` (scripted): vanilla project; `compile`, delete the classes dir, `compile` + `run` again must restore the classes and run. Mirrors the reported `ClassNotFoundException` case exactly; fails on develop, passes with this change.
- Full `cache/*` scripted group green (11 tests); `utilCache` scalafmt, `scalafmtSbtCheck`, MiMa clean.

The `declareOutput`-in-a-loop half of #9462 is a separate macro-layer defect; a follow-up PR is prepared for it.

One test-infrastructure note: scripted batch mode runs a group's tests sequentially in one shared directory, and fixtures without an explicit root project id share output paths, so one fixture's leftover class files can let a later same-named fixture overwrite them through a read-only directory check (directory write permission only guards entry creation). The two new fixtures here define distinct root project ids for that reason; `cache/compile-io-failure` remains sensitive to same-path predecessors, which I can harden in a follow-up if you'd like.

---

*Diagnosed and fixed with AI assistance (Claude Code); root cause isolated with an empirical discriminator (deleting the sibling zip made the identical cache hit restore correctly) and validated locally as above.*

### DIFF
--- a/notes/2.0.0/dirzip-restore-on-hit.md
+++ b/notes/2.0.0/dirzip-restore-on-hit.md
@@ -1,0 +1,15 @@
+### Directories declared with `Def.declareOutputDirectory` are restored on cache hits
+
+A directory declared as a cached task's output is packaged as a `.sbtdir.zip`
+sibling of the directory. Deleting the directory leaves the sibling zip behind
+(an `rm -rf` of the directory or of `classes/` does exactly this), which left
+the cache convinced everything was in sync: on the next cache hit the task did not re-run, but the
+directory was never re-extracted either. For sbt's own `compile`, whose classes
+directory is declared this way, a deleted output directory plus a warm cache
+meant `run` failed with `ClassNotFoundException` and no recompile. The cache now
+re-extracts a declared directory when the directory itself is missing, at the
+cost of a single stat on the warm path.
+
+This addresses the directory-restoration half of [#9462][i9462].
+
+[i9462]: https://github.com/sbt/sbt/issues/9462
--- a/sbt-app/src/sbt-test/cache/compile-classes-restore/Main.scala
+++ b/sbt-app/src/sbt-test/cache/compile-classes-restore/Main.scala
@@ -1,0 +1,2 @@
+@main def main(args: String*) =
+  println("hello from probe")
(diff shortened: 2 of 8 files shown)


eed3si9n (approved):
Thanks!

PR #9474, opened 2026-07-19
[2.x] Make `clearCaches` clear the sbt 2 action cache

kevin-lee:
It's a proposal to align the behavior of `clearCaches` with the newly added caching feature in sbt 2.

---
I found that `clearCaches` in sbt 2 works the same way as it does in sbt 1, so it doesn't clear the action caches introduced in sbt 2. This PR proposes updating it to clear the action caches when `clearCaches` is run. I've used Claude Code's Fable 5 to create comparison tables showing the differences between `clean`, `clearCaches`, and `cleanFull`.

To me, the most meaningful difference between the new behavior of `clearCaches` in this PR and `cleanFull` is that, unlike `cleanFull`, `clearCaches` doesn't delete the command history (`target/out/.history`). I think it makes sense for `cleanFull` to delete the history, as its name suggests, but in many cases, I just want to clear the caches while keeping the sbt console command history.

> [!NOTE]
> [This build failure](https://github.com/sbt/sbt/actions/runs/29886874422/job/88819114744?pr=9474) will be handled once #9485 is merged into the `develop` branch, and this branch is rebased onto `develop`.

***

# [2.x] Make `clearCaches` clear the sbt 2 action cache as well

## Problem

`clearCaches` predates the sbt 2 action cache and only resets in-memory state: the compiler cache, the classloader cache, and the file-cache store factory. The disk action cache (`cas/` and `ac/` under `localCacheDirectory`) survives it entirely. The only command that removes the action cache is `cleanFull`, which is much blunter: it unconditionally deletes all of `target/out` — including the command history at `target/out/.history` — plus the sbt boot directory, and it ignores `cleanKeepFiles` / `cleanKeepGlobs`.

There was no way to say "drop the action cache, keep my history and outputs".

## What this changes

`clearCaches` keeps doing everything it did before, and now also:

- deletes the contents of the `cas/` and `ac/` directories of every `DiskActionCacheStore` in `cacheStores` (remote stores are untouched, consistent with `cleanFull`),
- deletes the materialized task-value checkouts under `target/out/value`,
- removes symlinks under `target/out` that point into a cleared store — cached outputs are materialized as symlinks into `cas/`, so leaving them behind would leave dangling links for IDEs/tools that read `target/out` without running a task,
- respects `cleanKeepFiles` / `cleanKeepGlobs`, aggregated across all projects (unlike `cleanFull`).

What it still does **not** touch, unlike `cleanFull`: build outputs in `target/out`, the command history, and the boot directory.

## Implementation notes

- The in-memory resets run **before** the disk deletion so that jar handles held by cached classloaders are released before their blobs are deleted (this matters on Windows).
- Deletions are retried on transient `IOException`s, mirroring `DiskActionCacheStore#syncFile`'s write-side behavior; remaining failures are collected and reported as a warning rather than swallowed.
- An info line reports which store base directories were cleared (or a warning if only partially cleared).
- The help text for `clearCaches` is updated to describe the new behavior and the difference from `cleanFull`.

## Testing

New scripted test `cache/clear-caches`: populates the action cache via a cached task, runs `clearCaches`, and asserts that the disk store is completely empty, `target/out/value` and cache-pointing symlinks are gone, while keep-globbed paths, regular files under `target/out`, and re-caching after the clear all work.

⚠️ The test asserts strict behavior with no baseline/polling workarounds. On Windows it depends on the test-classloader/jar-handle leak fixes in #9485; the Windows scripted CI job is expected to fail until #9485 is merged and this branch is rebased onto it. Passes on Linux/macOS regardless (verified locally on macOS).

***

## Behavior before

Rows marked ⚠️ are the gaps this PR addresses.

| Location / state | `clean` (task) | `clearCaches` (command) | `cleanFull` (command) |
|---|---|---|---|
| `target/out/<platform>/scala-<V>/<module>/` — per-project target (classes, `streams/`, `src_managed/`, fileOutputs) | ✅ contents deleted, except `cleanKeepFiles`/`cleanKeepGlobs` matches | ❌ | ✅ deleted (whole `target/out` wiped, keep-filters ignored) |
| `target/out/.history` — command history | ❌ never visited and keep-globbed | ❌ | ✅ deleted (side effect of `target/out` wipe) |
| ⚠️ **`target/out/value/` (materialized task-value cache files)** | ❌ | ⚠️ **❌ left behind** | ✅ deleted |
| ⚠️ **Symlinks under `target/out` into the cache (materialized artifacts)** | ✅ only inside per-project targets | ⚠️ **❌ left dangling if the cache is removed** | ✅ deleted |
| Other files/dirs directly under `target/out/` | ❌ | ❌ | ✅ deleted |
| `target/global-logging/`, `target/task-temp-directory/`, `target/bg-jobs/`, `project/target/` | ❌ | ❌ | ❌ survive (outside `target/out`) |
| ⚠️ **Local disk action cache — `<localCacheDirectory>/cas/` + `/ac/` (default machine-global)** | ❌ | ⚠️ **❌ never cleared — the gap** | ✅ entire base dir deleted (no filters) |
| Remote cache stores (non-disk entries in `cacheStores`) | ❌ | ❌ | ❌ silently skipped |
| Launcher boot directory | ❌ | ❌ | ✅ deleted |
| In-memory: compiler cache, classloader cache, streams-cache wrapper | ❌ | ✅ all three reset | ✅ |
| JVM-global digest/stamp caches (`CacheImplicits`) | ❌ | ❌ | ❌ |

## Behavior after

Rows marked 🆕 are the behavior changes of this PR — everything else is identical to before.

| Location / state | `clean` (task) | `clearCaches` (after this PR) | `cleanFull` (command) |
|---|---|---|---|
| Per-project outputs `target/out/<platform>/scala-<V>/<module>/` (organic files: classes etc.) | ✅ contents deleted, except keep-filter matches | ❌ preserved (only symlinks into cleared bases removed) | ✅ deleted (whole `target/out`, no filters) |
| `target/out/.history` (command history) | ❌ preserved | ❌ preserved (regular file — never a candidate) | ✅ deleted (side effect) |
| 🆕 **`target/out/value/` (materialized task-value cache files)** | ❌ not visited | 🆕 **✅ deleted (keep-filter respected)** | ✅ deleted |
| 🆕 **Symlinks under `target/out` into cleared store bases (materialized artifacts)** | ✅ only inside per-project targets | 🆕 **✅ deleted everywhere under `target/out` (filtered); re-materialize on next run** | ✅ deleted |
| Other regular files/dirs under `target/out` | ❌ | ❌ preserved | ✅ deleted |
| `target/global-logging/`, `task-temp-directory/`, `bg-jobs/`, `project/target/` | ❌ | ❌ | ❌ |
| 🆕 **Local disk action cache `<localCacheDirectory>` (`cas/`+`ac/`, default machine-global)** | ❌ | 🆕 **✅ contents deleted (filtered; base dir kept; machine-wide effect)** | ✅ entire base dir deleted (no filters) |
| Remote cache stores | ❌ | ❌ untouched | ❌ silently skipped |
| Launcher boot directory | ❌ | ❌ preserved | ✅ deleted |
| In-memory: compiler / classloader / streams-cache wrapper | ❌ | ✅ reset (unchanged) | ✅ reset |
| JVM-global digest caches (`CacheImplicits`) | ❌ | ❌ | ❌ |
| 🆕 **Respects `cleanKeepFiles`/`cleanKeepGlobs`** | ✅ | 🆕 **✅ for all its disk deletions** (a protection even `cleanFull` lacks) | ❌ |

### DIFF
--- a/main-command/src/main/scala/sbt/BasicCommandStrings.scala
+++ b/main-command/src/main/scala/sbt/BasicCommandStrings.scala
@@ -254,7 +254,13 @@ $AliasCommand name=
   def continuousBriefHelp: (String, String) =
     (ContinuousExecutePrefix + " <command>", continuousDetail)
   def ClearCaches: String = "clearCaches"
-  def ClearCachesDetailed: String = "Clears sbt's internal caches."
+  def ClearCachesDetailed: String =
+    "Clears sbt's internal in-memory caches (compiler cache, classloader cache, and file-cache stores), " +
+      "deletes the contents of the local disk action cache (its cas/ and ac/ directories), " +
+      "and removes action-cache references under the root output directory: " +
+      "the materialized task-value files in target/out/value and any symbolic links pointing into the deleted cache. " +
+      "Paths matching cleanKeepFiles or cleanKeepGlobs are preserved. " +
+      "Unlike cleanFull, build outputs, the command history (target/out/.history), and the sbt boot directory are not deleted."
 
   val CleanFull: String = "cleanFull"
   def cleanFullDetailed: String = "Clears sbt's local caches."
--- a/main/src/main/scala/sbt/internal/Clean.scala
+++ b/main/src/main/scala/sbt/internal/Clean.scala
@@ -25,6 +25,7 @@ import sbt.nio.file.*
 import sbt.nio.file.syntax.pathToPathOps
 import sbt.nio.file.Glob.GlobOps
 import sbt.util.{ DiskActionCacheStore, Level }
+import sbt.internal.io.Retry
 import sbt.internal.util.complete.SizeParser
 import sjsonnew.JsonFormat
 import xsbti.{ PathBasedFile, VirtualFileRef }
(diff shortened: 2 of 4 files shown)


main-command/src/main/scala/sbt/BasicCommandStrings.scala
```diff
@@ -254,7 +254,13 @@ $AliasCommand name=
...
     (ContinuousExecutePrefix + " <command>", continuousDetail)
   def ClearCaches: String = "clearCaches"
-  def ClearCachesDetailed: String = "Clears sbt's internal caches."
+  def ClearCachesDetailed: String =
+    "Clears sbt's internal in-memory caches (compiler cache, classloader cache, and file-cache stores), " +
+      "deletes the contents of the local disk action cache (its cas/ and ac/ directories), " +
+      "and removes action-cache references under the root output directory: " +
+      "the materialized task-value files in target/out/value and any symbolic links pointing into the deleted cache. " +
+      "Paths matching cleanKeepFiles or cleanKeepGlobs are preserved. " +
+      "Unlike cleanFull, build outputs, the command history (target/out/.history), and the sbt boot directory are not deleted."
```
eed3si9n, line 263: Given the existence of `cleanFull`, I don't think we need yet another task to clear out the disk cache. I think `clearCaches` would be more useful if it stuck with in-memory caches.

kevin-lee: @eed3si9n Yeah, I generally agree with you, but there's an issue I mentioned in the second paragraph above.

> To me, the most meaningful difference between the new behavior of `clearCaches` in this PR and `cleanFull` is that, unlike `cleanFull`, `clearCaches` doesn't delete the command history (`target/out/.history`). I think it makes sense for `cleanFull` to delete the history, as its name suggests, but in many cases, I just want to clear the caches while keeping the sbt console command history.

`cleanFull` is too destructive because it deletes everything, including the command history stored in `target/out/.history`. I don’t want to lose my entire command history just to clear the action cache. This is important to me and maybe important to many others as well.

Do you have any other suggestions? Perhaps `clearActionCaches` to clear only action caches?

eed3si9n: I'd be open to `cleanFull` not wiping out the JLine history. If we remove this line https://github.com/sbt/sbt/blob/d95b50d8e02dc94d0923c276bb3af1cf8fced82b/main/src/main/scala/sbt/internal/Clean.scala#L230

would it still work? `cleanFull` recently started invoking normal `clean`, so it may no longer need to wipe out the out directory via `IO.delete`.

kevin-lee: @eed3si9n So, does it skip deleting the `out` directory entirely? If so, that may introduce another issue. While working on this PR, I found that `out` contains action cache files and directories, as well as symlinks to those caches. If `out` isn't deleted, those cache files and symlinks will remain.

Given its name, I assumed that `cleanFull` was meant to delete everything, including the command history, which seemed perfectly reasonable. That's why I proposed this PR.

I don’t fully understand how sbt is designed to work or the intended purpose of each command, so I’m trying to avoid negatively affecting the UX.

Skipping deletion of `out` might affect the UX and cause another issue, so I'm concerned that my request might make it worse.

Would there be a better way?

PR #9485, opened 2026-07-21
[2.x] Fix three test-classloader / jar-handle leaks in the sbt server JVM

kevin-lee:
Three independent, pre-existing retention bugs kept the classloader of a finished in-process test run -- and the open jar handles it holds -- alive for the rest of the server session. On Linux/macOS this leaks memory and file descriptors; on Windows the pinned handles make the loaded jars undeletable, which blocks clearCaches from deleting cas blobs. All three chains were proven from heap dumps. A Throwable's backtrace pins the Class of every frame, and a Class strongly references its defining classloader, so anything outliving a test task must not retain those objects.

- `JUnitXmlTestsListener`: `testSuite` is an InheritableThreadLocal, so threads spawned during a run (e.g. async-framework pool workers) inherit a copy of the suite reference that `remove()` cannot reach. Store the suite behind a mutable AtomicReference cell and clear it in writeSuite: clearing severs the graph for the owning thread and every inheritor at once. testEvent now drops late events instead of appending to an already-written suite.

- `TestRecap.collect`: the recap is stashed on `State.attributes` and outlives the command, so it must not retain live throwables. Strip SuiteResult.throwables from the retained Tests.Output copy while leaving the source TestsFailedException in the Incomplete tree intact, since error reporting and the ClassLoaderLayeringStrategy diagnostic still read the real throwables from there.

- `ClassLoaderCache`: loaders evicted from `delegate` by clearExpiredLoaders are unreachable from the map and never enqueued on the ReferenceQueue, so neither `clear()`/`close()` nor the cleanup thread could ever close them. Track evicted loaders weakly in `retired` and close them in clear() so their handles are released deterministically.

SuiteResult now documents the retention hazard on `throwables`. Adds deterministic, cross-platform tests for all three severed chains.

### DIFF
--- a/main-actions/src/main/scala/sbt/internal/testing/TestRecap.scala
+++ b/main-actions/src/main/scala/sbt/internal/testing/TestRecap.scala
@@ -45,7 +45,12 @@ import sbt.util.Logger
  */
 private[sbt] object TestRecap:
 
-  /** A single failed test task contributing to the recap. */
+  /**
+   * A single failed test task contributing to the recap.
+   *
+   * `testOutput` is sanitized by [[collect]]: its `SuiteResult.throwables` are dropped so the
+   * recap cannot pin the test class loader. See [[collect]] for why.
+   */
   final case class Failure(taskName: String, testOutput: Option[Tests.Output])
 
   /**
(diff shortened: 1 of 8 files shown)


eed3si9n:
@kevin-lee Thanks for the contribution!

eed3si9n (approved):
Overall LGTM.
For the ease of backporting, could you split bug fixes into individual PRs in the future plz?

kevin-lee:
@eed3si9n Sorry, I didn’t expect this to be merged so quickly because of the merge conflict. 😅 I also forgot to provide more context.

I found these leaks while working on #9474, which is only a proposal at this stage. [This build failure](https://github.com/sbt/sbt/actions/runs/29886874422/job/88819114744?pr=9474) was caused by two (or at least one) of them.

By "split bug fixes into individual PRs", do you mean creating a separate PR for each leak? If so, I agree. My apologies for putting them all in one PR. Since they were by-products of #9474, I was not careful enough. I prefer small PRs too and will handle this properly next time. Thank you!

Issue #9349, opened 2026-06-17
[2.x] Intermittent `sjsonnew.SerializationException: error while writing the field outputFiles`


mrdziuban:
## steps

Unfortunately I can't reproduce it reliably, but I've seen it happen during scripted tests. Here's an example: https://github.com/mblink/find-unused/actions/runs/27656358863/job/81791330619

## problem

An error is thrown:

```
sjsonnew.SerializationException: error while writing the field outputFiles
  at sjsonnew.package$.serializationError(package.scala:24)
  at sjsonnew.Builder.addField(Builder.scala:48)
  at sbt.internal.util.codec.ActionResultFormats$$anon$1.write(ActionResultFormats.scala:27)
  at sbt.internal.util.codec.ActionResultFormats$$anon$1.write(ActionResultFormats.scala:25)
  at sjsonnew.SupportConverter.toJsonUnsafe(SupportConverter.scala:23)
  at sjsonnew.SupportConverter.toJsonUnsafe$(SupportConverter.scala:5)
  at sjsonnew.support.scalajson.unsafe.Converter$.toJsonUnsafe(Converter.scala:7)
  at sbt.util.DiskActionCacheStore.put(ActionCacheStore.scala:223)
  at sbt.util.AggregateActionCacheStore.put$$anonfun$1(ActionCacheStore.scala:106)
  at scala.collection.LinearSeqOps.foldLeft(LinearSeq.scala:186)
  at scala.collection.LinearSeqOps.foldLeft$(LinearSeq.scala:38)
  at scala.collection.immutable.List.foldLeft(List.scala:83)
  at sbt.util.AggregateActionCacheStore.put(ActionCacheStore.scala:104)
  at sbt.util.ActionCache$.organicTask$1(ActionCache.scala:133)
  at sbt.util.ActionCache$.cache(ActionCache.scala:182)
  at sbt.Defaults$.packageTask$lzyINIT1$$anonfun$1(Defaults.scala:1946)
  at scala.Function1.$anonfun$compose$1(Function1.scala:79)
  at scala.Function1.$anonfun$compose$1(Function1.scala:79)
  at sbt.std.Transform$$anon$3.work(Transform.scala:79)
  at sbt.std.Transform$$anon$3.work(Transform.scala:79)
  at sbt.Execute.submit$$anonfun$1$$anonfun$1(Execute.scala:283)
  at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:24)
[...]

PR #9488, opened 2026-07-22
[2.x] fix: Fixes partial cache restoration

stasimus:
`DiskActionCacheStore.syncBlobs` wrapped the blob write in the same `catch NoSuchFileException => None` as the availability lookup, so a failed restore was silently dropped, leaving the output tree incomplete. Now only the lookup swallows it; write failures propagate and the task falls back to onsite recompute (also widened the cache-storage catch to `NonFatal`). Related to #9349.


eed3si9n (approved):
Thanks!

PR #9492, opened 2026-07-23
[2.x] fix: Fixes Def.declareOutput invoked in a loop

### LINKED ISSUE #9462: [2.0.2] sbt compile caching broken + tracking down the Def.declareOutput & Def.declareOutputDirectory issues

BrianHotopp:
## Problem

The cached-task macro allocated one mutable slot per syntactic `Def.declareOutput` / `Def.declareOutputDirectory` call site and snapshotted the slots into the task's outputs after the body ran. A call inside a loop or `.map` over a runtime-determined list is a single syntactic site executed many times: each iteration overwrote the same slot, so only the last file was cached and restored on a cache hit (#9462, case 4 — `a.txt`/`b.txt` lost, `c.txt` restored; the directory-restoration half was fixed in #9473). Any real code-generation task with a dynamic number of outputs hits this. A conditional call site that did not execute also left `null` in the outputs list.

## Fix

Declared outputs now accumulate at runtime: the macro emits one `ListBuffer` at the top of the cached body and rewrites each call site to `ActionCache.registerOutput(vf, buffer)`, which appends and returns the value. Every execution registers, an unexecuted site contributes nothing, and a task with no declared outputs generates the same shape as before.

The rewritten call site is deliberately a plain method call: generating a local `val` there crashes `LambdaLift` when the macro re-parents the body into the cache closure, and the helper must be public because the generated call lives in the user's build code.

## Notes, stated deliberately

- `ListBuffer +=` is not thread-safe if user code calls `Def.declareOutput` from parallel threads inside one task body; the old per-site mutable var had the same property, so no regression.
- Duplicate registrations are not deduped, same as the old multi-site behavior; syncBlobs is idempotent.
- The loop fixture also contains a `declareOutput` in a never-taken branch, pinning that unexecuted sites contribute nothing (the old codegen left a null slot in the outputs).

## Testing

- `cache/declare-output-loop` (scripted): three files declared via `.map` over a runtime list; delete all three; the cache hit must restore all three. Fails on develop (only the last file comes back), passes with this change.
- `cache/basic` still green — multiple static `declareOutput` sites unchanged.
- Full `cache/*` scripted group green (11 tests); `mainProj` compiles, re-expanding every internal `Def.cachedTask` through the new codegen; scalafmt and MiMa clean on core-macros, main-settings, and util-cache.

---

*Diagnosed and fixed with AI assistance (Claude Code); the macro-layer root cause (one write-once var per syntactic call site) was isolated by tracing the expansion in `Cont.scala` and validated as above.*

Fixes #9462


eed3si9n (approved):
Thanks!

PR #9496, opened 2026-07-24
[2.x] refactor: Separate file I/O from cache-write serialization

BrianHotopp:
## Problem

The `HashedVirtualFileRef` codec does live file I/O at serialization time: it reads `contentHashStr` and `sizeBytes`, which for zinc's `MappedVirtualFile` is a plain `def` calling `Files.size` on every invocation. If a file referenced by a task's result vanishes between the task completing and its cache write, that stat throws mid-serialization, surfacing as the reporter's intermittent `SerializationException: error while writing the field outputFiles` (#9349).

The intermittency is a real race, recovered from the reporter's original CI log and reproduced locally under strace: a plugin using `Extracted.runTask` (sbt-buildinfo's task-typed keys) triggers a concurrent duplicate evaluation of the same `packageBin` action, since nested evaluations share no deduplication with the outer graph and the action cache has no per-digest single-flight. The winner's post-put `syncFile` swaps the freshly built jar for a CAS symlink via a non-atomic delete-then-link (a sub-millisecond hole measured under strace); the loser's serialization-time stat lands in the hole.

#9488 already stopped this from failing the build: `organicTask` now recovers from any `NonFatal` storage error and returns the task's value uncached. This PR removes the cause on the store side, so the write succeeds instead of being skipped.

## Fix

`putBlobsIfNeeded` (shared by the disk, in-memory, and remote gRPC stores) reads each blob's hash and size once, up front, and returns only plain `HashedVirtualFileRef.of(id, hash, size)` values. Serializing an `ActionResult` then performs no file I/O in any store, nothing re-stats a blob after its CAS entry is written, and I/O errors on an output file surface upfront at blob storage time rather than mid-serialization: a file vanishing once its blob is stored no longer prevents the cache write.

One live read remains outside this PR's reach: a task whose *value* itself contains file refs (e.g. `packageBin` returning its jar ref) serializes them before the put. A vanish there is covered by #9488's recovery.

## Testing

Unit (`ActionCacheTest`, 58/58):
- A file vanishing the moment its blob reaches the CAS: the write succeeds and the next run is a cache hit. Fails on develop, where the same sequence degrades and recomputes.
- A stored ref must not re-stat its file (the materialization pin). Fails on develop.
- A file already missing at the cache write: value returned uncached, next run recomputes.
- A value whose codec throws: value returned uncached, next run recomputes (pins the value-serialization leg of #9488's `NonFatal` recovery, which had no test).

Full `scripted cache/*` sweep green; scalafmt and MiMa clean.

Follow-up candidates deliberately not in this PR: an atomic symlink swap in `syncFile`, and per-digest single-flight for concurrent duplicate evaluations of the same action. Happy to file issues for either.

---

*Diagnosed and fixed with AI assistance (Claude Code); the race was pinned empirically (original CI log recovered, strace shows the unlink-to-symlink hole, disk cache held five action entries sharing one CAS blob) and the fix is discriminator-proven (the vanish-after-storage and materialization tests fail with the store change reverted).*

### DIFF
--- a/util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
+++ b/util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
@@ -66,16 +66,20 @@ end ActionCacheStore
 
 trait AbstractActionCacheStore extends ActionCacheStore:
   def putBlobsIfNeeded(blobs: Seq[VirtualFile]): Seq[HashedVirtualFileRef] =
+    // Read each blob's hash and size once, up front, and return only these plain value refs:
+    // serializing the ActionResult afterwards does no file I/O, and nothing re-stats a blob
+    // after its CAS entry is written.
+    val materialized: Seq[(VirtualFile, HashedVirtualFileRef)] = blobs.map: blob =>
+      blob -> HashedVirtualFileRef.of(blob.id, blob.contentHashStr, blob.sizeBytes)
     val found = findBlobs(blobs).toSet
     val missing = blobs.flatMap: blob =>
       val ref: HashedVirtualFileRef = blob
       if found.contains(ref) then None
       else Some(blob)
     val combined = putBlobs(missing).toSet ++ found
-    blobs.flatMap: blob =>
+    materialized.flatMap: (blob, plain) =>
       val ref: HashedVirtualFileRef = blob
-      if combined.contains(ref) then Some(ref)
-      else None
+      if combined.contains(ref) then Some(plain) else None
 
   def notFound: Throwable =
     new RuntimeException("not found")
--- a/notes/2.0.0/cache-write-serialization-no-io.md
+++ b/notes/2.0.0/cache-write-serialization-no-io.md
@@ -1,0 +1,15 @@
+### Cache-write serialization no longer performs file I/O
+
+Serializing a task's result for the action cache re-read the size of every
+referenced file at write time, so a file vanishing between the task completing
+and the cache write (for example when two overlapping evaluations of the same
+task race the jar-to-CAS-symlink swap) made an otherwise successful task's
+cache write throw an intermittent `sjsonnew.SerializationException: error
+while writing the field outputFiles`. Each stored output reference is now
+materialized once, before its blob is stored, so the cache write succeeds even
+if the file vanishes afterwards, and I/O errors on an output file surface
+upfront at blob storage time rather than mid-serialization.
+
+This addresses [#9349][i9349].
+
+[i9349]: https://github.com/sbt/sbt/issues/9349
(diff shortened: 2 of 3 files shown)


eed3si9n:
The problem description is more interesting than the actual PR. As commented in https://github.com/sbt/sbt/issues/9349#issuecomment-5072545254, if the plugin is evaluating task outside of the graph, then subsequent side effect errors caused by it, isn't really sbt main's fault.

util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
```diff
@@ -66,16 +66,20 @@ end ActionCacheStore
 
 trait AbstractActionCacheStore extends ActionCacheStore:
   def putBlobsIfNeeded(blobs: Seq[VirtualFile]): Seq[HashedVirtualFileRef] =
+    // Read each blob's hash and size once, up front, and return only these plain value refs:
+    // serializing the ActionResult afterwards does no file I/O, and nothing re-stats a blob
+    // after its CAS entry is written.
+    val materialized: Seq[(VirtualFile, HashedVirtualFileRef)] = blobs.map: blob =>
+      blob -> HashedVirtualFileRef.of(blob.id, blob.contentHashStr, blob.sizeBytes)
```
eed3si9n, line 73: This is a nice change to get IO errors upfront.

util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -136,20 +136,33 @@ object ActionCache:
...
       catch
         case NonFatal(e) =>
-          logger.debug(s"Skipping cache storage due to error: ${e.getMessage}")
+          findMissingFile(e) match
+            case Some(path) =>
+              logger.warn(
+                s"the task succeeded but its result could not be cached; a file referenced by the result no longer exists: $path"
+              )
+            case None =>
+              logger.debug(s"Skipping cache storage due to error: ${e.getMessage}")
```
eed3si9n: If the situation is caused likely by concurrent task runs, I'm kind of ok with not printing the warnings for now. Could you revert the changes in this file plz?

eed3si9n (changes requested):

util-cache/src/main/scala/sbt/util/ActionCache.scala
BrianHotopp: Done in 510c048. Reverted all changes in this file; the PR is now the store-side materialization only, plus tests and the release note (both updated to match).

util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
BrianHotopp: Thanks!

eed3si9n (approved):
Thanks!

PR #9497, opened 2026-07-24
[2.x] fix: Fixes reboot from sbtn

### LINKED ISSUE #9095: [2.x] `reboot` causes sbt to exit

BrianHotopp:
Fixes https://github.com/sbt/sbt/issues/9095

## Problem

Running `reboot` in the sbt shell drops to the OS shell ("sbt server connection closed") instead of rebooting (#9095). The reboot protocol is intact on both ends: the server sends the reboot-flavored `Shutdown` notification, and the thin client's reconnect machinery (wait for the server, re-attach, replay) handles it. The break is in teardown. `Server.shutdown` opens with `log.info`, and during a client-initiated reboot the terminal in scope is that client's already-closed virtual terminal, so the log write throws `ClosedChannelException` through the terminal proxy. Teardown aborts before the portfile is deleted and the server socket closed, and the exception is swallowed by the shutdown hook, whose own error print goes to the same dead terminal.

The relaunched instance's startup probe then connects to the leaked socket, concludes another sbt is running, and never starts a server; the client's wait is satisfied by the stale portfile and it attaches to the zombie socket. Under sbtn the relaunched instance immediately self-shutdowns (`startedByRemoteClient` with no server), producing the reported "sbt server connection closed".

## Fix

- `Server.shutdown` completes its state cleanup (portfile, tokenfile, running flag, server socket) before logging, so teardown cannot be aborted by a logging failure.
- `CommandExchange.shutdown` wraps each channel shutdown and the server shutdown individually, so one failing step cannot skip the rest.

No protocol or client changes: the existing reconnect machinery works once teardown completes. (Two follow-up candidates, happy to file: the client's `blockUntilStart` treats a still-existing portfile as "server is back", which masked this; and shutdown-hook exceptions print to a possibly-dead terminal, which made the abort invisible.)

## Testing

- `RebootTest` (server-test) exercises the round trip with batch clients: `reboot` must complete with exit 0 (the client's reboot handler reconnects to the relaunched server and completes the exec), and a second client must then be served by the rebooted server. On develop the relaunched instance connects to the leaked socket, warns `Running multiple instances is unsupported`, and the follow-up client exits 1; with this change both exit 0, verified under 2-core CPU contention (3/3) as well as unconstrained.
- An interactive-stdin variant of the test was abandoned deliberately: it exposed a pre-existing input race in the thin client, unrelated to this fix (a `readSystemIn` request that arrives while the one-byte `RawInputThread` is exiting is silently dropped, permanently wedging session input; reproducible under CPU contention). Happy to file that separately with the capture logs.
- `commandProj`/`mainProj` compile, scalafmt, MiMa clean.

---

*Diagnosed with AI assistance (Claude Code): the failure was reproduced in the server-test harness, the `ClosedChannelException` captured with instrumentation, and the fix validated end-to-end before productionizing.*

### DIFF
--- a/main-command/src/main/scala/sbt/internal/server/Server.scala
+++ b/main-command/src/main/scala/sbt/internal/server/Server.scala
@@ -159,7 +159,6 @@ private[sbt] object Server {
       }
 
       override def shutdown(): Unit = {
-        log.info("shutting down sbt server")
         if (portfile.exists) {
           IO.delete(portfile)
         }
@@ -171,6 +170,7 @@ private[sbt] object Server {
           case null =>
           case s    => s.close()
         }
+        log.info("shutting down sbt server")
       }
 
       private def writeTokenfile(): Unit = {
--- a/main/src/main/scala/sbt/internal/CommandExchange.scala
+++ b/main/src/main/scala/sbt/internal/CommandExchange.scala
@@ -320,9 +320,9 @@ private[sbt] final class CommandExchange {
     }
     procFile = None
     fastTrackThread.close()
-    channels foreach (_.shutdown(true))
+    channels.foreach(c => Util.ignoreResult(Try(c.shutdown(true))))
     // interrupt and kill the thread
-    server.foreach(_.shutdown())
+    server.foreach(s => Util.ignoreResult(Try(s.shutdown())))
     server = None
     EvaluateTask.onShutdown()
   }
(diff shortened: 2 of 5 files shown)


eed3si9n (approved):
Thanks!

Issue #9477, opened 2026-07-20
Memory-retention issues on large monorepo build (task results, `compile` `Analysis`, and the analysis cache)


hoangmaihuy:
First, thank you for your hard work on sbt 2.0 — most things work out of the box and it's been really stable.

I'm migrating a large monorepo from Mill to sbt 2.0. For scale:

- **~3.1M lines of Scala** across ~17,000 source files
- **300 subprojects** — cross-built for JVM and Scala.js (123 JVM + 139 JS rows + aggregates), from ~96 cross-platform modules
- **~90 test-bearing subprojects** across ~1,100 test source files

The migration itself is essentially complete; we're now in a build-performance / memory-optimization phase, comparing against Mill. With the help of Claude Code we've found three memory issues that compound on a build this size. They all revolve around the same object (the zinc `Analysis`), so I'm grouping them in one report — happy to split into separate issues if you'd prefer.

For context on why this bites us specifically: our CI drives the build in two phases — one step compiles all ~300 subprojects, then a separate step runs the tests. Once the compile phase is done, none of the compile `Analysis` values are needed in memory any more (the test phase reads what it needs from disk), yet they stay resident — accumulated through the compile phase by the task engine (Issue 1) and carried into the test phase by the static analysis cache (Issue 3). For a compile-then-test pipeline like ours, holding every compile result in memory across the whole run is pure waste.

Below issues are investigated with the help of Claude Code. I'm sorry in advanced if there is any AI slop here, I hope this issue can help improving sbt performance.

---

## Issue 1 — the task engine (`Execute`) keeps every task result in memory until the whole evaluation completes

The task-evaluation engine keeps **every** task's `Result` in memory from the moment the task completes until the entire evaluation finishes. Nothing is released mid-run, even after every task that depended on a given result has already consumed it.

In `tasks/src/main/scala/sbt/Execute.scala`, `results` is declared:

```scala
private val results = taskPMap[Result]
```

`retire` stores each completed result — `results(node) = result` — and there is **no** corresponding removal anywhere in the engine. The map is only handed off at the very end of `runKeep`:

```scala
def runKeep[A](root: TaskId[A])(using strategy: CompletionService): RMap[TaskId, Result] = {
  ...
  addNew(root)
  processAll()
  ...
  val finalResults = triggers.onComplete(results)   // whole map, all entries still present
  progress.afterAllCompleted(finalResults)
  progress.stop()
  finalResults
}
```

[...]

PR #9500, opened 2026-07-25
[2.x] fix: Fixes inMemoryAnalysisCache limit

eed3si9n:
Ref https://github.com/sbt/sbt/issues/9477

## Problem
The in-memory Analysis cache is currently weighted by on-disk size, but according to the user report it expands out 50x, so 100MB size limit doesn't hit.

## Solution
This set maxCount instead, default 8 and configurable by system property `sbt.local_cache.analysis_count`.

### DIFF
--- a/main/src/main/scala/sbt/internal/SysProp.scala
+++ b/main/src/main/scala/sbt/internal/SysProp.scala
@@ -123,6 +123,8 @@ object SysProp:
   @deprecated("Resident compilation is no longer supported", "1.4.0")
   def residentLimit: Int = int("sbt.resident.limit", 0)
 
+  def analysisCacheMaxCount: Int = int("sbt.local_cache.analysis_count", 8)
+
   /**
    * Indicates whether formatting has been disabled in environment variables.
    * 1. -Dsbt.log.noformat=true means no formatting.
--- a/main/src/main/scala/sbt/internal/BuildDef.scala
+++ b/main/src/main/scala/sbt/internal/BuildDef.scala
@@ -9,7 +9,7 @@
 package sbt
 package internal
 
-import com.github.benmanes.caffeine.cache.{ Cache as CCache, Caffeine, Weigher }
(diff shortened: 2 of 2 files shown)


hoangmaihuy:
@eed3si9n I tried to set limit for analysis cache and test locally, but it makes compile time a lot slower for a full compile because the cache keep evicting entries and it makes compling downstream projects (which is depending on a lot of subprojects, and we have one root project basically depending on everything) suffers from reading analysis from disk -> load to cache -> evicted.

So I think the solution should be a flag to clear the analysis cache after command evaluation is done.

eed3si9n:
We can do that separately but I think we should fix the ceiling of the caffeine cache so it doesn't end up using 1GB or 5GB, if it expands 50x.

hoangmaihuy:
@eed3si9n I've measured heap usages in https://github.com/sbt/zinc/issues/1753 and the ratio is about 20x.
I don't know if the compile analyses holding by the cache are the same instances holding by task result map, if not the memory allocation for compile analyses alone is about 40x disk size.

main/src/main/scala/sbt/internal/SysProp.scala
```diff
@@ -123,6 +123,8 @@ object SysProp:
   @deprecated("Resident compilation is no longer supported", "1.4.0")
   def residentLimit: Int = int("sbt.resident.limit", 0)
 
+  def analysisCacheMaxCount: Int = int("sbt.local_cache.analysis_count", 8)
```
stasimus, line 126: IMHO it is too small,  a 10-project build ~ 20 analyses - slowly than  100MB weight limit

stasimus (approved):

main/src/main/scala/sbt/internal/SysProp.scala
eed3si9n: I'd be happy to bump this number. Should we do 20?

stasimus: Yes

Issue #9507, opened 2026-07-26
sbtn: interactive input wedges permanently when a readSystemIn request races the input thread's exit


BrianHotopp:
The server requests interactive stdin one byte at a time via `readSystemIn` notifications. In `NetworkClient`, `RawInputThread.run()` reads a single byte, forwards it, and exits through `finally inputThread.set(null)`, while `startInputThread()` is a no-op whenever `inputThread` is non-null. A `readSystemIn` that arrives in the window between the forward and the `finally` is therefore silently dropped: no thread ever reads the next byte, the server's pending terminal read blocks forever, and session input is dead. Typed bytes stop echoing and executing, Enter does not revive it, and only killing sbtn and re-attaching recovers. A byte read while `attached` is false during a reboot is likewise parked in `stdinBytes` with nothing left to drain it, so typing during `reboot` can lose the command and dead-end the session.

Reproduced against develop with a real pseudo-terminal driving the actual client against a forked server: on an otherwise idle machine 0/5 sessions wedge, but with background CPU load 3/4 wedge at paste-speed input, often on the first command. jstack of a wedged client shows no `sbt-read-input-thread` while the connection thread is healthy, matching the mechanism. Happy to share the driver and dumps, and to send a PR (loop the reader until cancelled, or restart it race-free in `startInputThread`).

eed3si9n:
Thanks for the report. Happy to get a PR on this.

Issue #6841, opened 2022-03-16
Reading `StdIn` from client can block the server


adpi2:
## steps

1. Create an minimal sbt project with a program that reads the standard input

```scala
// build.sbt
scalaVersion := "2.13.7"
```

```scala
// src/main/scala/example
import scala.io.StdIn

object HelloWorld extends App {
  val name = StdIn.readLine("Enter your name: ")
  println(s"Hello $name!")
}
```

2. Run the program within sbt client
```
$ sbtn run
[info] running HelloWorld 
Enter your name:
```
3. Kill the client but not the server
```
$ ps -aux | grep sbtn
$ kill -9 <PID of sbtn>
```
4. try to run the program again with `sbtn run`

## problem

The server is stuck on waiting for the input.

## expectation

The server should close the standard input of the program after the client is shutdown.

## Workaround

```
sbt shutdownall // shutdown all running sbt server
```

Issue #6840, opened 2022-03-16
`ReloadOnSourceChanges` can block the server


adpi2:
Appeared in https://github.com/sbt/sbt/discussions/6828

## steps

- Set `Global / onChangedBuildSource := ReloadOnSourceChanges`
- Open an sbt client with `sbt -client`
- Alter the build definition with some code that does not compile
- Run any task

## problem

The server will try and fail to reload. The usual `[warn] Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)` message is not forwarded to the client. The input from the client are not forwarded to the server.

## expectation

The server should not get stuck.

Either:
1. it should fallback to its initial state
2. it should forward the warning to the client and wait for an input 
```
[warn] Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)
```

2 only works if the client is a shell (not a BSP client)

## Workaround

```
sbt shutdownall // shutdown all running sbt server
```

keynmol:
Thanks for making mine a duplicate - I'll just bring over the GIF with demonstration.
![2023-03-09 12 01 14](https://user-images.githubusercontent.com/1052965/224068152-1d5a0768-168b-4efd-a34d-9e9d3ac7f231.gif)

PR #9510, opened 2026-07-26
[2.x] fix: Keep the client stdin reader alive so input requests are never dropped

BrianHotopp:
## Problem

The server requests interactive stdin one byte at a time via `readSystemIn` notifications. `RawInputThread.run()` read a single byte, forwarded it, and exited through `finally inputThread.set(null)`, while `startInputThread()` was a no-op whenever `inputThread` was non-null. A `readSystemIn` arriving in the window between the forward and the `finally` was silently dropped: no thread ever read the next byte, the server's pending terminal read blocked forever, and session input was dead. Typed bytes stopped echoing and executing, and only killing sbtn and re-attaching recovered. A byte read while `attached` was false during a reboot was likewise parked in `stdinBytes` with nothing left to drain it (#9507).

Reproduced through a real pseudo-terminal driving the actual client entry against a forked server: paste-speed input wedges 5 of 10 sessions pinned to 2 cores, and 3 of 4 sessions on an unpinned 12-core machine with background CPU load; jstack of a wedged client shows no `sbt-read-input-thread` alive.

## Fix

- The reader thread now lives until input reading is cancelled or stdin reaches EOF, reading in a loop, so there is no per-byte exit window to race.
- `close()` marks the thread stopped before interrupting, and `startInputThread()` replaces a stopped-but-not-yet-deregistered thread, with both sides synchronizing on the registration, so an input request that arrives while the reader is exiting starts a replacement instead of being dropped.
- The `stopped` flag is initialized before `start()`, since `run()` reads it immediately; the reproduction harness caught an NPE from the reverse order recreating the wedge.
- Bytes read while detached mid-reboot no longer strand, because the surviving reader drains them once re-attached. A command typed mid-reboot can still be lost to the dying server's connection, but the session now survives it; preserving type-ahead across reboot is offered as a follow-up.

Two consequences worth stating. After a cancel, a reader parked in a blocking `read` cannot be interrupted out of it; it reads at most one more byte, forwards it under the lock, and exits, so a cancel-then-request cycle has a brief bounded two-reader window whose worst case is a one-byte reorder, strictly dominated by the permanent wedge it replaces. And one `readSystemIn` now starts continuous streaming until cancelled, so paste bursts buffer in the server's unbounded input queue instead of the client's tty buffer; the consumers were audited and none assumes at-most-one-byte-per-request (the boot path already sent multi-byte batches).

## Testing

- Real-pty verification, same driver and configurations as the reproduction on the issue, with the client jar hash checked before each series to rule out stale bytecode. The race is scheduler-dependent, so rates vary by run: across two independent develop series, paste bursts wedged 5 of 10 pinned sessions in one and 0 of 15 in the other, while sessions on an unpinned 12-core machine with background CPU load wedged 3 of 4 and 5 of 6; every captured jstack shows the same signature (no `sbt-read-input-thread` alive). With this change, 0 of 31 sessions across all configurations wedge, and typing during `reboot` leaves a live session in 3 of 3 runs instead of a dead one. Driver and jstack dumps available on request.
- No in-repo interactive regression test ships with this PR, deliberately. An interactive test was written and passed everywhere locally, but on CI runners it consistently stalls in the server's per-keystroke terminal queries, a separate pre-existing unbounded-wait defect family in `NetworkChannel`/`VirtualTerminal` (the mechanism behind #6841 and #6840); a thread capture in the CI failure showed this fix's reader thread alive and waiting, which is the fixed behavior. Rather than merge a test that cannot run in CI, it was dropped; an interactive test becomes CI-viable once those waits are bounded.
- `commandProj` compile, scalafmt, MiMa clean.

---

*Diagnosed and fixed with AI assistance (Claude Code); the race was found by code reading during CI triage of #9497, confirmed with a real-pty reproduction against a forked server, and the fix stress-verified under CPU contention with the same driver.*

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -173,9 +173,13 @@ class NetworkClient(
   private val sbtProcess = new AtomicReference[Process](null)
   private class ConnectionRefusedException(t: Throwable) extends Throwable(t)
   private class ServerFailedException extends Exception
-  private def startInputThread(): Unit = inputThread.get match {
-    case null => inputThread.set(new RawInputThread)
-    case _    =>
+  private val inputThreadLock = new Object
+  private def startInputThread(): Unit = inputThreadLock.synchronized {
+    inputThread.get match {
+      case null               => inputThread.set(new RawInputThread)
+      case t if t.stopped.get => inputThread.set(new RawInputThread)
+      case _                  => ()
+    }
   }
   private lazy val log: Logger = new Logger {
     def trace(t: => Throwable): Unit = ()
@@ -1130,18 +1134,21 @@ class NetworkClient(
 
   private class RawInputThread extends Thread("sbt-read-input-thread") with AutoCloseable {
     setDaemon(true)
-    start()
+    // stopped must be initialized before start(): run() reads it as soon as the thread runs.
     val stopped = new AtomicBoolean(false)
+    start()
     override final def run(): Unit = {
-      def read(): Unit = {
-        val b = inputStream.read
-        inLock.synchronized(stdinBytes.offer(b))
-        if (attached.get()) drain()
-      }
-      try read()
-      catch { case _: InterruptedException | NonFatal(_) => stopped.set(true) }
+      try {
+        var b = 0
+        while (!stopped.get && b != -1) {
+          b = inputStream.read
+          inLock.synchronized(stdinBytes.offer(b))
+          if (attached.get()) drain()
+        }
+      } catch { case _: InterruptedException | NonFatal(_) => () }
       finally {
-        inputThread.set(null)
+        stopped.set(true)
+        inputThreadLock.synchronized(Util.ignoreResult(inputThread.compareAndSet(this, null)))
       }
     }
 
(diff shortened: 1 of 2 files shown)


main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -1130,18 +1134,21 @@ class NetworkClient(
...
-        val b = inputStream.read
-        inLock.synchronized(stdinBytes.offer(b))
-        if (attached.get()) drain()
-      }
-      try read()
-      catch { case _: InterruptedException | NonFatal(_) => stopped.set(true) }
+      try {
+        var b = 0
+        while (!stopped.get && b != -1) {
+          b = inputStream.read
```
eed3si9n, line 1144: I used Claude Code to find out why sbtn uses one-byte-read thread approach, and found that this fixes the issue of preemptive `read()` causing issue when terminal tries to switch the mode, e.g. #5828/#5863. I have a counter PR that this the thread approach further, and let it live throughout the lifetime of sbtn - https://github.com/sbt/sbt/pull/9521.

eed3si9n (changes requested):

BrianHotopp:
Superseded by #9521, which also preserves the #5828/#5863 constraint this loop would have broken. Validated there with the same harness.

PR #8815, opened 2026-02-25
[2.x] perf: lightweight UpdateReport cache persistence

bitloi:
Closes #4883

## Summary
This PR implements lightweight `UpdateReport` persistence to reduce cache payload size and avoid expensive full-report deserialization on cache hits, while preserving correctness and backward compatibility with existing cache files.

It also includes scripted-test updates required to keep CI green on current output formats.

## Problem
`UpdateReport` loading has been a recurring hotspot. The previous cached-update path stored and loaded full `UpdateReport` data, which is heavier than needed for cache persistence.

## Solution
Introduce a dedicated persistence layer for update cache:
- Persist a lightweight cache payload (`UpdateReportLite` + required metadata).
- Reconstruct a full `UpdateReport` from lightweight data when needed.
- Fall back to legacy full `UpdateReport` cache format for backward compatibility.

### Core implementation
- Added [UpdateReportPersistence.scala](/root/74/bronze/sbt/lm-core/src/main/scala/sbt/internal/librarymanagement/UpdateReportPersistence.scala)
  - Defines `UpdateReportCache` with:
    - `lite: UpdateReportLite`
    - `stats: UpdateStats`
    - `stamps: Map[String, Long]`
    - `cachedDescriptor: File`
  - Adds read/write helpers:
    - `writeTo(store, cache)` for new format
    - `readFrom(store)` with fallback to legacy `UpdateReport`

- Updated [LibraryManagement.scala](/root/74/bronze/sbt/main/src/main/scala/sbt/internal/LibraryManagement.scala)
  - Replaced `Tracked.lastOutput`-based cache persistence for update output with `UpdateReportPersistence` read/write.
  - Keeps existing cache validity checks (`upToDate`, stamps, descriptor checks).
  - Preserves `stats.cached` behavior for consumers.

- Updated [JsonUtil.scala](/root/74/bronze/sbt/lm-core/src/main/scala/sbt/internal/librarymanagement/JsonUtil.scala)
  - `toLite` now preserves modules even when `details` are stripped by rebuilding details from modules.
  - Added `fromLiteFull(...)` for full module reconstruction from persisted lite payload.

### Tests and validation
- Added [UpdateReportPersistenceSpec.scala](/root/74/bronze/sbt/lm-core/src/test/scala/sbt/internal/librarymanagement/UpdateReportPersistenceSpec.scala)
  - Verifies stats/stamps preservation
  - Verifies behavior with/without details
  - Verifies legacy fallback compatibility

[...]

PR #9515, opened 2026-07-27
[2.x] perf: Optimize `UpdateReport` disk format

### LINKED ISSUE #9514: [2.x] `update` is quadratic in modules-per-resolution, and the report it produces is ~94% duplicate values
**Disclaimer**: Investigated with Claude Code and human review in the loop

Detailed analysis of `update` on a large multi-project build: where the wall time goes, how much of
the report is duplication, what it costs in heap, and what it costs on disk. All measurements are
against stock sbt `2.0.4`; the code paths named below are unchanged on `develop`.

## Build under test

301 aggregated projects, 302 `update` task instances, 16–25 configurations per project (median 17),
1,209 distinct external module coordinates. JVM: `-Xmx20G -Xms8G`, G1, JDK 26.0.1, macOS arm64.

## Headline numbers

| | stock 2.0.4 |
|---|---|
| `sbt update`, cold caches | **523.9 / 558.9 / 600.7 s** (three runs) |
| `update` as a dependency of another task, warm caches, 301/301 cache hits | **100.9 s** |
| share of `update` CPU spent rebuilding one map | **61.9%** |
| `ModuleReport` entries written vs distinct values | **933,750 → 2,032** (459×) |
| `InclExclRule` instances vs distinct values | **5,284,692 → 410** (12,890×) |
| heap retained by 302 co-resident reports | **2,702 MB**, of which **2,334 MB is duplicate values** |
| peak heap during `sbt update` | **13,763 MB** used, 19.92 GB committed against a 20 GB `-Xmx` |
| update caches on disk | **1.562 GiB**, 5.30 MiB per project |

## 1. Wall time: where it goes

Timing only the `update` command — build load and JVM startup excluded and recorded separately
(22–46 s), one JVM per run:

| shape | wall |
|---|---|
| `sbt update`, cold caches | 523.9 s / 558.9 s / 600.7 s |
| `sbt update`, everything already on disk | 521.0 s |
| `update` as a dependency, warm caches (301/301 hits) | 100.9 s |

The first two rows are the same work: `sbt update` does not consult its own cache (§6), so "warm"
and "cold" differ only in whether the artifacts are already downloaded. Run-to-run spread on the
cold figure is ~15%.

`-Dsbt.task.timings` attributes essentially all of it to `update` itself: 302 instances, 3,449 s of
task time across ~7.6 parallel threads (11.4 s per project), with no other task above a 200 ms
threshold.

## 2. The quadratic: `projectCache` rebuilt once per dependency

async-profiler 4.5, wall-clock sampling attached for exactly the duration of `update`, with
`-XX:+DebugNonSafepoints`. Idle and parked stacks excluded — on macOS a wall profile is ~90% threads
in `semaphore_wait_trap`/`mach_msg2_trap`, which drowns the signal if counted. 377,156 active
samples:

| bucket | share of active CPU |
|---|---|
| **`coursier.core.Resolution.projectCache`** | **61.9%** |
| scala immutable HashMap / CHAMP node churn | 55.0% |
| update report (de)serialisation | 17.5% |
| hashing and boxing | 11.4% |
| file I/O | 9.4% |
| GC | 5.7% |

83.1% of the `projectCache` time enters through a single call site:

```
83.1%  193,895  SbtUpdateReport$.lookupProject$1 <- SbtUpdateReport$.$anonfun$26
                <- StrictOptimizedIterableOps.strictOptimizedFlatMap
 4.0%    9,287  SbtUpdateReport$.$anonfun$33 <- Iterator$$anon$10.nextCur
 3.8%    8,795  SbtUpdateReport$.$anonfun$34 <- Option.orElse <- SbtUpdateReport$.$anonfun$33
```

That call site is `lm-coursier/src/main/scala/lmcoursier/internal/SbtUpdateReport.scala` (v2.0.4
line 240, `develop` line 257):

```scala
def lookupProject(mv: coursier.core.Resolution.ModuleVersion): Option[Project] =
  res.projectCache.get(mv) match {
    case Some((_, p)) => Some(p)
    case _            => interProjectDependencies.find(p => mv == (p.module, p.version))
  }
```

`lookupProject` runs once per module while building the reports, and again per parent POM while
resolving inherited license info (`assemble` → `licenseInfo` recurses up the parent chain). Four
further reads sit in the per-configuration block of `apply` (v2.0.4 lines 364, 367, 392–394, 399),
three of them inside the eviction loop, i.e. once per conflict.

Each of those reads rebuilds the entire map. From `javap -c` on the shaded coursier in
`librarymanagement-coursier_3-2.0.4.jar`:

```
public Map<Tuple2<Module, VersionConstraint>, Tuple2<ArtifactSource, Project>> projectCache0();
public Map<Tuple2<Module, String>,            Tuple2<ArtifactSource, Project>> projectCache();

public scala.collection.immutable.Map<...> projectCache();
  Code:
     0: aload_0
     1: invokevirtual #576   // Method projectCache0:()Lscala/collection/immutable/Map;
     4: invokedynamic #581,0 // InvokeDynamic #2:apply:()Lscala/Function1;
     9: invokeinterface #560 // InterfaceMethod scala/collection/immutable/Map.map:(...)
    14: checkcast     #556
    17: areturn
```

`projectCache0` is the field; `projectCache` is a compatibility shim that maps every entry to
convert the key's `VersionConstraint` to a `String`. It is not memoised — no backing field, no
`$lzycompute` — while sibling members of the same class (`dependencies$lzycompute`,
`hashCode$lzycompute`, `transitiveDependenciesAndErrors$lzycompute`, …) are lazy. Every call
re-converts, re-hashes and re-boxes every entry.

The leaf frames underneath are pure map construction:

```
11.5%  BitmapIndexedMapNode.getValue          4.8%  BitmapIndexedMapNode.migrateFromInlineToNodeInPlace
 9.7%  HashMapBuilder.insertValue             4.4%  BoxesRunTime.boxToInteger
 8.8%  HashMapBuilder.update                  4.3%  Resolution.$anonfun$projectCache$1
 5.4%  BitmapIndexedMapNode.originalHashes    2.8%  BitmapIndexedMapNode.mergeTwoKeyValPairs
 5.2%  arrayof_jint_disjoint_arraycopy_stub   2.2%  Tuple2.hashCode
```

For N modules resolved this is N rebuilds of an N-entry `HashMap`, paid once per project — so it
grows with both dependency count and project count.

## 3. Duplication: the report is mostly repeated values

A per-field census over 302 persisted update caches from this build, read with the **stock generated
codecs** so the numbers describe the raw parse (2026-07-25; corpus 1,714 MB of JSON, median 4.6 MB
per report, 4,978 `ConfigurationReport`s):

| type | instances | distinct values corpus-wide | factor |
|---|---|---|---|
| `InclExclRule` | 5,284,692 | **410** | 12,890× |
| `ModuleID` | 933,750 | **2,025** | 461× |
| `ModuleReport` | 933,750 | **2,032** | 459× |
| `Artifact` | 783,258 | **971** | 807× |
| `java.io.File` | 782,725 | **970** | 807× |
| `java.net.URI` | 780,900 | **970** | 805× |
| `OrganizationArtifactReport` | 749,108 | **2,625** | 285× |
| `String` | 29,030,257 refs | **4,706** | 6,169× |
| `ConfigRef` | 6,517,460 refs | 31 | already canonical |

Two properties matter for where the cost comes from:

**It is within one report, not across reports.** 2,200 MB of the duplication is inside single
reports against 134 MB across them — 94.2% / 5.8%. A report covers ~18 configurations that repeat
the same module set, so `InclExclRule` alone has 5,225,327 within-report duplicates against 58,955
cross-report ones (89:1). A *single* project's report already carries ~94% of the duplication.

**Exclusion rules dominate.** ~996 MB, 37% of the retained heap, is `InclExclRule` and what it owns:
161 MB of objects, 41 MB of `exclusions` vectors and 754 MB of strings — including **5,284,390
duplicate copies of the one-character string `"*"`, ~207 MB on its own**.

Strings are the largest absolute bucket at 1,627 MB, because JSON has no string table: every
occurrence deserialises to a fresh instance. 29,030,257 references resolve to 4,706 distinct values
(1,820 per report).

Top individual sources, estimated bytes (within + cross):

```
inclExcl.organization             ~279.5 MB      moduleReport.licenses.str    ~111.4 MB
inclExcl.name                     ~267.3 MB      artifact.url.str             ~105.1 MB
inclExcl.artifact                 ~206.6 MB      config.modules.ModuleReport   ~78.2 MB
moduleId.exclusions.InclExclRule  ~201.6 MB      moduleReport.configurations   ~57.0 MB
file.path                         ~135.8 MB      moduleReport.module.ModuleID  ~56.9 MB
```

Grouped by owner: `InclExclRule` ~996 MB, `Artifact`+`File`+`URI` ~492 MB, `ModuleID`'s own fields
~273 MB, `ModuleReport`-owned ~430 MB, `OrganizationArtifactReport`-owned ~137 MB.

Re-measured independently, by parsing the 302 `output` files on disk rather than the retained
graph: **949,492 `ModuleReport` entries written for 1,209 distinct `(organization, name,
revision)` coordinates build-wide** — 7.7× duplication within a project (the same module repeated
across its configurations) and 785× build-wide, plus 788,515 artifact entries. This counts
distinct *coordinates*, where the census above counts distinct `ModuleReport` *values* (one
coordinate can yield several, differing by configuration or artifact set), and it is a later
snapshot of the build — hence 1,209 against the census's 2,032.

## 4. Memory

**Reports co-resident.** Loading all 302 reports the way `update` does and forcing four GCs leaves
**2,702 MB retained across 87.82 M objects**. The census accounts for **2,334 MB of that as
removable duplicate values** — 1,627 MB duplicate strings plus 707 MB duplicate value objects.

**During a real `update`.** Baseline is measured after the build loads and before `update` runs;
retained is after two forced full GCs:

| | `sbt update` (cold) | `update` as a dependency (301/301 hits) |
|---|---|---|
| live set before `update` | 1,169 MB | 1,049 MB |
| **peak heap used** | **13,763 MB** | **8,985 MB** |
| peak heap committed | **19.92 GB** — against a 20 GB `-Xmx` | 12.81 GB |
| retained after `update` | 3,063 MB | 1,276 MB |
| GC count / total pause | 932 / 26.6 s | 52 / 5.4 s |

Live-object composition afterwards (`GC.class_histogram`, live only):

| class | `sbt update` | as a dependency |
|---|---|---|
| `String` | 8,537,329 (195.4 MB) | 2,787,067 (63.8 MB) |
| `InclExclRule` | 61,223 | **396,714** (12.1 MB) |
| `ModuleReport` | 75,674 | 74,049 |
| `ModuleID` | 24,206 | 86,081 |
| `Artifact` | 38,665 | 74,126 |
| `java.io.File` | 32,096 | 87,833 |
| `Vector1` | 2 | 1,383,595 (21.1 MB) |
| total live | 108.2 M objects / 2.99 GB | 45.3 M / 1.25 GB |

74,049 live `ModuleReport`s against 1,209 distinct coordinates is the in-heap face of the same
duplication. The `Vector1` column shows the deserialisation path: 1.38 M single-element vectors
appear only when reports come back from the cache.

**A related memory effect.** `InMemoryCacheStore` bounds its Caffeine cache by the total *on-disk
size* of its entries (`Files.readAttributes(path, …).size`), while what an entry holds is the
deserialized object graph. With `fileCacheSize` at its 128 M default and reports averaging 5.30 MiB
on disk, 48 full `UpdateReport`s were resident at the end of a run — the budget is being spent in
file bytes, not heap bytes.

## 5. Disk

| | |
|---|---|
| `update` cache stores | 302 |
| total | **1.562 GiB** |
| average per project | **5.30 MiB** |
| largest single report measured | 4,486,360 bytes |

The size follows directly from §3: each configuration serialises its full module set, so a project
with 17 configurations writes its ~100 modules ~17 times. Nothing in the format references a
module written earlier in the same file.

For scale, the same build's complete cache tree (including `dependencyPositions` and the stores
other tasks create) is 3.35 GB across 905 directories.

## 6. Cache behaviour

Two behaviours worth stating explicitly, because they determine which of the numbers above a user
actually pays.

**`sbt update` never reads the update cache.** `Defaults.scala`:

```scala
val isRoot = executionRoots.value contains resolvedScoped.value
val shouldForce = isRoot || { ...forceUpdatePeriod... }
```

and `cachedUpdate`'s check begins `!force && !inChanged && ...`. Asking for `update` explicitly
means "refresh". Measured: after `sbt update`, **0 of 302** persisted reports carried
`stats.cached = true`, and the wall profile of a second consecutive run was indistinguishable from
the first. When `update` instead runs as a dependency of another task, the same build reports
**301 of 301 hits** and takes 100.9 s instead of ~560 s.

**A cache hit still pays a full serialisation round trip.** `Tracked.lastOutput` reads the previous
value and writes the new one unconditionally:

```scala
def lastOutput[I, O: JsonFormat](store: CacheStore)(f: (I, Option[O]) => O): I => O = { in =>
  val previous = Try { store.read[O]() }.toOption   // always
  val next = f(in, previous)
  store.write(next)                                  // always
  next
}
```

So even when `cachedUpdate` decides the report is up to date, 1.562 GiB is deserialised and
re-serialised. That is the 17.5% "(de)serialisation" row in §2, and it is why the cached path still
costs 100.9 s. It also means file timestamps cannot be used to tell a hit from a miss.

## Method

- One JVM per run: `sbt --server` in the foreground, one command, then exit; a guard ensures no
  server survives between runs, so a "warm" run reads from disk rather than from a live server.
- Only the command is timed. The clock starts when it is written into the already-loaded shell and
  stops on a sentinel command's output; load and startup are recorded separately.
- Cold runs move the build's output tree aside and recreate it empty, and point the action cache at
  an empty directory. Coursier's artifact cache is deliberately left warm, so the figures measure
  resolution and report construction rather than network.
- Heap: `jcmd GC.run` ×2 + `GC.heap_info` for baseline and retained, `-Xlog:gc` and a 500 ms `jstat`
  sampler for peak, `GC.class_histogram` for composition. JVM flags are the build's own, untouched.
- Profiles: async-profiler 4.5, `-e wall -t`, attached for the duration of the command only, with
  `-XX:+UnlockDiagnosticVMOptions -XX:+DebugNonSafepoints`.
- The duplication census walks the retained graph of 302 reports parsed with the stock generated
  codecs, counting occurrences, identity-distinct instances and equals-distinct values per field.
- Cache hits are established from `UpdateReport.stats.cached` on the persisted reports and from a
  task that depends on `update` and logs hit/miss per project — not from file timestamps, which
  change on every run regardless.

To reproduce the profile on any build with a substantial dependency graph:

```
asprof start -e wall -i 10ms -t -o collapsed -f update.collapsed <pid>
# run update
asprof stop -o collapsed -f update.collapsed <pid>
```

then drop idle stacks and grep the collapsed file for `projectCache`.

hoangmaihuy:

Fixes #9514 (§4, §5, §6). Last of four split-out pieces; #9522 *(§2)*, #9523 *(§3, §4)* and #9524 *(§5)* are merged. This one is the on-disk format plus the in-memory cache fix, and depends on neither the interner nor the no-hash codec.

### 1. Write each module once *(§5)*

`SbtUpdateReport` rebuilds a project's module set per configuration and the cache serialized every copy — on a 302-module monorepo, **949,492 module reports for 157,773 distinct values, 1,655 MB on disk**. Each value-distinct `ModuleReport` now goes once into a `"modules"` table, and the details hold indices into it.

```jsonc
// before — every configuration repeats the reports
{"configurations": [
   {"configuration": "compile", "modules": [ {…full report…}, … ]},  // 318 of them
   {"configuration": "test",    "modules": [ {…the same again…} ]}   // ×17
 ]}

// after — one table, referenced by index
{"version": 1,
 "modules": [ {…each distinct report once…} ],                       // 385 entries
 "configurations": [
   {"configuration": "compile",
    "details": [ {"organization": "org.scala-lang", "name": "scala3-library_sjs1_3",
                  "modules": [0]} ]}
 ]}
```

~5× smaller, and the reader gets instance sharing structurally rather than by probing for it. The details repeat as much as the modules do, so they are shared too, keyed on the index vector. Holding all 302 reports co-resident retains **47 MB instead of 2,724 MB**.

**Resolver order.** `managedJars` takes the managed classpath's order from `ConfigurationReport.modules`, and that order reaches the compile task's cache key through `CompileInputs2.classpath`. It cannot be recovered from the details: flattening regroups by (organization, name), moving any module the resolver interleaved with another organization's. A report restored in the grouped order therefore misses the entry its own resolution just wrote — the compile re-runs, and zinc, handed a reordered classpath, invalidates behind it. On the 302-module build that was 108 of one project's 346 sources plus a cascade through the JVM closure, on the first command after a cold `Test / compileIncremental`, with no source change.

Each configuration stores its order as a second index vector, written only where it differs from the flattened detail order — so only the configurations the resolver interleaved pay for it. `toCache` puts that order through the same `filterOutArtificialCallers` normalization `JsonUtil.toLite` applies to the details (extracted as `withFilteredCallers`), so a module and its order entry share one table slot instead of being written twice as near-identical values.

**`publicationDate`.** A dated report shares a slot like any other: before serialization the same `Option[Calendar]` is already reachable from every configuration, so sharing adds no aliasing, while excluding them would cost the whole dedup on an Ivy build where most modules carry a date. Two separate reads stay separate; a test pins that.

### 2. Compress what is left *(§5)*

What remains is still highly repetitive text: 336 MB of caches, 19 MB gzipped. It makes reading **faster**, not slower, because 19 MB of I/O beats 336 MB by more than inflating costs — loading all 302 reports drops from 8.8 s to 7.3 s, and peak heap during the load from 2,184 MB to 659 MB.

```diff
-    val outStore = cacheStoreFactory.make("output")
+    val outStore = cacheStoreFactory.makeCompressed("output")
```

`makeCompressed` defaults to `make`, so factories with no compressed form keep compiling. `InMemoryCacheStore`'s factory forwards it to its delegate; without that it would inherit the default, wrap an uncompressed delegate, and silently ignore the request. The `inputs` store stays plain — a few hundred bytes.

### 3. Weigh in-memory entries by their uncompressed size *(§4)*

`InMemoryCacheStore` bounds its cache by the total weight of its entries and used the entry's size on disk. Once the store compresses, that is the *compressed* size, so the same budget admits several times what it was set to hold — roughly 2 GB against a 128M budget on the measured build, which is why 1–2 cut peak heap but left retained heap flat.

```diff
-  val attributes = Files.readAttributes(path, classOf[BasicFileAttributes])
-  files.put(path, (value, lastModified, toIntExact(attributes.size)))
+  files.put(path, (value, lastModified, toIntExact(weightOf(path))))
```

`weightOf` reads the gzip trailer's ISIZE through `GzipFileInput.uncompressedSize`, next to the reader that already knows the framing — magic checked, short read rejected, size on disk for anything that is not a whole gzip member. That restores the unit `fileCacheSize` has always been denominated in: serialized bytes.

Serialized size is only ever a proxy for what an entry costs to keep, and this series flips the sign of its error. Before it, 1,655 MB of JSON held 2,724 MB of graph — the proxy under-counted by ~1.6×, so a 128M budget quietly retained ~210 MB. After it, 336 MB holds 47 MB — the same proxy over-counts by ~7×, so the same budget retains ~18 MB while admitting **five times as many reports** as before. Same proxy, moved off the OOM side; sizing an arbitrary deserialized value to buy that back is not worth it.

**Not changed:** `update` refusing its own cache when it is an execution root *(§6)*. Asking for `update` explicitly means "refresh".

## Compatibility

- **Every cache a released sbt wrote still reads.** One parse: `"version":1` is the new shape, anything else the full-report codec sbt 2.0.x wrote. `Unbuilder.lookupField` reads a by-name map and consumes nothing, so an older cache costs one parse instead of one per format tried. A cache claiming a *newer* version is a miss, so the project re-resolves rather than failing.
- **Every branch demands a field its own shape must have**, the fallback included. sjson-new decodes a missing array as an empty one rather than failing, so without that a foreign shape would read back as a report with no configurations — which `update` accepts as up to date and hands back as an empty managed classpath.
- **The unversioned `"lite"` shape is not read back.** It only ever existed on `develop`, so a build already running a `develop` sbt re-resolves once and nothing else.
- **Gzip framing is read-compatible.** The reader sniffs the magic and accepts plain content, so an old cache loads and is rewritten compressed on the next run.
- **No breaking API change.** `CacheStoreFactory.makeCompressed` has a default implementation and is the only new public member; the stores behind it are `private[sbt]`.

## Result

Measured on a 301-project build with all four changes applied, same machine and JVM flags throughout. `update` alone is timed; build load and JVM startup are excluded, and each run is its own JVM. The figures predate the order elision and the caller normalization added during review — both only shrink the cache further.

**`update` running as a dependency of another task** — the shape in which the cache applies at all, verified at 301/301 hits:

| metric | 2.0.4 | all four | change |
|---|---|---|---|
| **wall** | **100.9 s** | **5.6 s** | **−94% (18×)** |
| peak heap used | 8,985 MB | 4,955 MB | −45% |
| retained after full GC | 1,276 MB | 1,039 MB | −19%, and equal to the pre-`update` baseline |
| GC pause total | 5.4 s | 3.2 s | −41% |

**`sbt update` typed directly, cold caches** — this always re-resolves (`shouldForce = isRoot`), so it measures report construction rather than caching:

| metric | 2.0.4 | all four | change |
|---|---|---|---|
| **wall** | **600.7 s** | **89.1 s** | **−85% (6.7×)** |
| peak heap used | 13,763 MB | 6,953 MB | −49% |
| GC count / pause | 932 / 26.6 s | 78 / 7.1 s | −92% / −73% |
| update caches on disk | 1.564 GiB | **0.018 GiB** | **−99% (92×)** |

**On-disk format**, a single project's report written both ways:

| | bytes | vs stock |
|---|---|---|
| 2.0.4, full report JSON | 4,486,360 | 1× |
| module-table JSON (gunzipped) | 888,389 | 5.0× smaller |
| as written (gzip framed) | 33,850 | **132.5× smaller** |

Corpus average across 302 projects: 5.30 MiB → 0.06 MiB, 92×.

## How the numbers were produced

One JVM per run — `sbt --server` in the foreground, one command, then exit, with a guard ensuring no server survives between runs. Only the command is timed: the clock starts when it is written into the already-loaded shell and stops on a sentinel command's output.

Cold runs move the output tree aside, recreate it empty and point the action cache at an empty directory; coursier's artifact cache is deliberately left warm, so the figures measure resolution and report construction rather than the network. Heap via `jcmd GC.run` ×2 + `GC.heap_info` for baseline and retained, `-Xlog:gc` plus a 500 ms `jstat` sampler for peak, and `GC.class_histogram` for composition.

Cache hits are established from `UpdateReport.stats.cached` on the persisted reports and from a task that depends on `update` and logs hit/miss per project — not from file timestamps, which change on every run because `Tracked.lastOutput` writes unconditionally.

Each cell is a single run. The cold `sbt update` figure has ~15% run-to-run spread on this machine (523.9 / 558.9 / 600.7 s across three runs of stock); the cached-path pair, 100.9 s against 5.6 s, is far outside that band.

### DIFF
--- a/lm-core/src/main/scala/sbt/internal/librarymanagement/JsonUtil.scala
+++ b/lm-core/src/main/scala/sbt/internal/librarymanagement/JsonUtil.scala
@@ -30,6 +30,34 @@ private[sbt] object JsonUtil {
     CacheStore(graphPath).write(updateReportLite)
   }
 
+  /** The per-module normalization `toLite` applies; anything stored alongside it must match. */
+  private[sbt] def withFilteredCallers(mr: ModuleReport): ModuleReport = {
+    val callers = filterOutArtificialCallers(mr.callers)
+    // Reuse the instance when filtering changed nothing, so interning is not undone here.
+    if (callers eq mr.callers) mr
+    else
+      ModuleReport(
+        mr.module,
+        mr.artifacts,
+        mr.missingArtifacts,
+        mr.status,
+        mr.publicationDate,
+        mr.resolver,
+        mr.artifactResolver,
+        mr.evicted,
+        mr.evictedData,
+        mr.evictedReason,
+        mr.problem,
+        mr.homepage,
+        mr.extraAttributes,
+        mr.isDefault,
+        mr.branch,
+        mr.configurations,
+        mr.licenses,
+        callers
+      )
+  }
+
   def toLite(ur: UpdateReport): UpdateReportLite =
     UpdateReportLite(ur.configurations map { cr =>
       val details0 = if (cr.details.nonEmpty) cr.details else modulesToDetails(cr.modules)
@@ -39,32 +67,7 @@ private[sbt] object JsonUtil {
           OrganizationArtifactReport(
             oar.organization,
             oar.name,
-            oar.modules map { mr =>
-              val callers = filterOutArtificialCallers(mr.callers)
-              // Reuse the instance when filtering changed nothing, so interning is not undone here.
-              if (callers eq mr.callers) mr
-              else
-                ModuleReport(
-                  mr.module,
-                  mr.artifacts,
-                  mr.missingArtifacts,
-                  mr.status,
-                  mr.publicationDate,
-                  mr.resolver,
-                  mr.artifactResolver,
-                  mr.evicted,
-                  mr.evictedData,
-                  mr.evictedReason,
-                  mr.problem,
-                  mr.homepage,
-                  mr.extraAttributes,
-                  mr.isDefault,
-                  mr.branch,
-                  mr.configurations,
-                  mr.licenses,
-                  callers
-                )
-            }
+            oar.modules.map(withFilteredCallers)
           )
         }
       )
--- a/lm-core/src/main/scala/sbt/internal/librarymanagement/UpdateReportPersistence.scala
+++ b/lm-core/src/main/scala/sbt/internal/librarymanagement/UpdateReportPersistence.scala
@@ -20,7 +20,9 @@ final case class UpdateReportCache(
     lite: UpdateReportLite,
     stats: UpdateStats,
     stamps: Map[String, Long],
-    cachedDescriptor: File
+    cachedDescriptor: File,
+    /** Per configuration, its modules in the resolver's order. */
+    moduleOrder: Vector[Vector[ModuleReport]]
 )
 
 object UpdateReportPersistence:
@@ -45,8 +47,8 @@ object UpdateReportPersistence:
 
   /** Interns the modules of a decoded cache, in a pass since the generated reader has no hook. */
   private def internModules(cache: UpdateReportCache): UpdateReportCache =
-    cache.copy(lite =
-      UpdateReportLite(
+    cache.copy(
+      lite = UpdateReportLite(
         cache.lite.configurations.map(cr =>
           ConfigurationReportLite(
             cr.configuration,
@@ -59,7 +61,9 @@ object UpdateReportPersistence:
             )
           )
         )
-      )
+      ),
+      // Interned too, so the restored order shares the instances the details were pooled to.
+      moduleOrder = cache.moduleOrder.map(_.map(UpdateReportInterner.intern))
     )
 
   /** The shape before this one, a whole serialized `UpdateReport`, carries no version. */
@@ -71,8 +75,12 @@ object UpdateReportPersistence:
       modules: Vector[Int]
   )
 
-  /** One configuration's details, in the order the configuration resolved them. */
-  private final case class IndexedConfig(configuration: String, details: Vector[IndexedDetail])
+  /** `modules` is absent when the resolver's order is the flattened detail order. */
+  private final case class IndexedConfig(
+      configuration: String,
+      details: Vector[IndexedDetail],
+      modules: Option[Vector[Int]]
+  )
 
   private given indexedDetailFormat: JsonFormat[IndexedDetail] = new JsonFormat[IndexedDetail]:
     def write[J](obj: IndexedDetail, builder: Builder[J]): Unit =
@@ -96,14 +104,18 @@ object UpdateReportPersistence:
       builder.beginObject()
       builder.addField("configuration", obj.configuration)
       builder.addField("details", obj.details)
+      obj.modules.foreach(builder.addField("modules", _))
       builder.endObject()
     def read[J](jsOpt: Option[J], unbuilder: Unbuilder[J]): IndexedConfig = jsOpt match
       case Some(js) =>
         unbuilder.beginObject(js)
         val configuration = unbuilder.readField[String]("configuration")
         val details = unbuilder.readField[Vector[IndexedDetail]]("details")
+        // Absent and empty mean different things here, and `readField` cannot tell them apart.
+        val modules =
+          unbuilder.lookupField("modules").map(_ => unbuilder.readField[Vector[Int]]("modules"))
         unbuilder.endObject()
-        IndexedConfig(configuration, details)
+        IndexedConfig(configuration, details, modules)
       case None => deserializationError("Expected JsObject but found None")
 
   private final class ModuleTable:
@@ -132,13 +144,14 @@ object UpdateReportPersistence:
 
   private def writeV1[J](obj: UpdateReportCache, builder: Builder[J]): Unit =
     val modules = new ModuleTable
-    val configurations = obj.lite.configurations.map: cr =>
-      IndexedConfig(
-        cr.configuration,
-        cr.details.map(oar =>
-          IndexedDetail(oar.organization, oar.name, oar.modules.map(modules.indexOf))
-        )
+    val configurations = obj.lite.configurations.zipWithIndex.map: (cr, i) =>
+      val details = cr.details.map(oar =>
+        IndexedDetail(oar.organization, oar.name, oar.modules.map(modules.indexOf))
       )
+      val flattened = details.flatMap(_.modules)
+      val order = obj.moduleOrder.lift(i).fold(flattened)(_.map(modules.indexOf))
+      // Written only where the resolver interleaved organizations; elsewhere it is the flattened order.
+      IndexedConfig(cr.configuration, details, Option.when(order != flattened)(order))
     builder.beginObject()
     builder.addField("version", FormatVersion)
     // Filled by the traversal above, so it has to be rendered after it.
@@ -180,7 +193,8 @@ object UpdateReportPersistence:
       ),
       stats,
       stamps,
-      cachedDescriptor
+      cachedDescriptor,
+      configurations.map(cfg => cfg.modules.getOrElse(cfg.details.flatMap(_.modules)).map(modules))
     )
 
   /**
@@ -225,12 +239,21 @@ object UpdateReportPersistence:
       lite = JsonUtil.toLite(ur),
       stats = ur.stats,
       stamps = ur.stamps,
-      cachedDescriptor = ur.cachedDescriptor
+      cachedDescriptor = ur.cachedDescriptor,
+      // Normalized the way `toLite` normalizes the details, so a module and its order entry share one
+      // table slot instead of being written twice as near-identical values.
+      moduleOrder = ur.configurations.map(_.modules.map(JsonUtil.withFilteredCallers))
     )
 
   def fromCache(cache: UpdateReportCache): UpdateReport =
-    JsonUtil
-      .fromLiteFull(cache.lite, cache.cachedDescriptor)
+    val restored = JsonUtil.fromLiteFull(cache.lite, cache.cachedDescriptor)
+    // `fromLiteFull` flattens the details, which regroups the modules by (organization, name).
+    // `lift`, not `zip`: a short order leaves the remaining configurations as flattened rather than
+    // dropping them. An order that is present and empty is a configuration that resolved nothing.
+    val configurations = restored.configurations.zipWithIndex.map: (cr, i) =>
+      cache.moduleOrder.lift(i).fold(cr)(cr.withModules)
+    restored
+      .withConfigurations(configurations)
       .withStats(cache.stats)
       .withStamps(cache.stamps)
 
--- a/lm-core/src/main/scala/sbt/internal/librarymanagement/UpdateReportPersistence.scala
+++ b/lm-core/src/main/scala/sbt/internal/librarymanagement/UpdateReportPersistence.scala
@@ -62,6 +62,133 @@ object UpdateReportPersistence:
       )
     )
 
+  /** The shape before this one, a whole serialized `UpdateReport`, carries no version. */
+  private final val FormatVersion = 1
+
+  private final case class IndexedDetail(
+      organization: String,
+      name: String,
+      modules: Vector[Int]
+  )
+
+  /** One configuration's details, in the order the configuration resolved them. */
+  private final case class IndexedConfig(configuration: String, details: Vector[IndexedDetail])
+
+  private given indexedDetailFormat: JsonFormat[IndexedDetail] = new JsonFormat[IndexedDetail]:
+    def write[J](obj: IndexedDetail, builder: Builder[J]): Unit =
+      builder.beginObject()
+      builder.addField("organization", obj.organization)
+      builder.addField("name", obj.name)
+      builder.addField("modules", obj.modules)
+      builder.endObject()
+    def read[J](jsOpt: Option[J], unbuilder: Unbuilder[J]): IndexedDetail = jsOpt match
+      case Some(js) =>
+        unbuilder.beginObject(js)
+        val organization = unbuilder.readField[String]("organization")
+        val name = unbuilder.readField[String]("name")
+        val modules = unbuilder.readField[Vector[Int]]("modules")
+        unbuilder.endObject()
+        IndexedDetail(organization, name, modules)
+      case None => deserializationError("Expected JsObject but found None")
+
+  private given indexedConfigFormat: JsonFormat[IndexedConfig] = new JsonFormat[IndexedConfig]:
+    def write[J](obj: IndexedConfig, builder: Builder[J]): Unit =
+      builder.beginObject()
+      builder.addField("configuration", obj.configuration)
+      builder.addField("details", obj.details)
+      builder.endObject()
+    def read[J](jsOpt: Option[J], unbuilder: Unbuilder[J]): IndexedConfig = jsOpt match
+      case Some(js) =>
+        unbuilder.beginObject(js)
+        val configuration = unbuilder.readField[String]("configuration")
+        val details = unbuilder.readField[Vector[IndexedDetail]]("details")
+        unbuilder.endObject()
+        IndexedConfig(configuration, details)
+      case None => deserializationError("Expected JsObject but found None")
+
+  private final class ModuleTable:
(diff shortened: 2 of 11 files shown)


eed3si9n:
Thanks for the contribution!

eed3si9n:
Since `update` task has a bunch of downstream users, could we split the PRs into fixes that we can apply independently or incrementally please?

hoangmaihuy:
> Since `update` task has a bunch of downstream users, could we split the PRs into fixes that we can apply independently or incrementally please?

I've splitted into 4 PRs as mentioned above, this PR is solely about `UpdateReportCache` on disk format changes.
This PR has some duplicated code needed from other PRs, so I recommend merging other PRs before this

hoangmaihuy:
Hi @eed3si9n, I've just realized that https://github.com/sbt/sbt/pull/8815 was not backported to 2.0.x series. 
Do you have a roadmap for these optimizations to be visible? I've dogfooded these changes in our own sbt 2 migration and the improvement is massive. I'm glad to address any issues to have it backported to sbt 2 branch 😄

eed3si9n:
> Do you have a roadmap for these optimizations to be visible?

`develop` branch targets the next minor version, namely sbt 2.1.0. I've been backporting a bunch of fix PRs from `develop` to `2.0.x`, but like any fresh code those PRs have also caused regressions, so I plan to slow down as we stabilize. We'll probably target to start RC cycle for 2.1.0 in a month or two. I'd be happy to release milestone builds like 2.1.0-M1 soon as well.

eed3si9n (approved):
Thanks!

PR #9521, opened 2026-07-27
[2.x] fix: Fixes sbtn stdin race condition

### LINKED ISSUE #9507: sbtn: interactive input wedges permanently when a readSystemIn request races the input thread's exit

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/9507

## Problem
There's a race condition between per-byte readSystemIn notification and one-byte-read thread lifecycle.

Note: One-byte-read thread was introduced as a solution to the problem that switching the terminal between raw and canonical mode cannot happen if it's blocked by read.

## Solution
This eliminates the thread lifecycle issue by keeping the thread alive throughout the lifecycle of sbtn itself.
read is still called on demand by the server readSystemIn notification.

## Credits
Credit to @BrianHotopp for identifying the issue and for the idea of modifying the input thread - https://github.com/sbt/sbt/pull/9510.

## AI Disclosure
I've used Claude Code to research the history, and also to let it write the first draft, but I've modified its output to make it more normal functional-looking Scala code.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -638,6 +638,9 @@ lazy val commandProj = (project in file("main-command"))
       exclude[IncompatibleResultTypeProblem]("sbt.internal.client.NetworkClient.connection"),
       exclude[IncompatibleResultTypeProblem]("sbt.internal.client.NetworkClient.init"),
       exclude[DirectMissingMethodProblem]("sbt.internal.BootServerSocket.*"),
+      exclude[DirectMissingMethodProblem](
+        "sbt.internal.client.NetworkClient#RawInputThread.stopped"
+      ),
     ),
     Compile / headerCreate / unmanagedSources := {
       val old = (Compile / headerCreate / unmanagedSources).value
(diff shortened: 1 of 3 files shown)


BrianHotopp:
Validated with the #9507 pty harness: 0 of 24 sessions wedge (develop wedged 5/10 pinned, 3/4 under load), and commands typed during reboot survive 3 of 3. LGTM. Closing #9510 in favor of this.

PR #9523, opened 2026-07-28
[2.x] perf: Intern `UpdateReport` values

hoangmaihuy:

Part of #9514 (§3, §4). Split out of #9515 so it can be reviewed on its own.

## Problem

The projects of a build mostly depend on the same libraries, and each project's report materializes its own copies of every coordinate it names. On a 302-module monorepo the cached reports hold **949,492 `ModuleReport`s for 2,036 distinct values**. One `InclExclRule` existed 5,284,692 times for 410 distinct values, and 94% of that repetition was *inside a single report*, because every configuration repeats the same module set.

### A weak interner (commit 1)

`WeakPools` — about 100 lines, no new dependency. A value lives exactly as long as some report references it, so nothing accumulates across a reload.

`internWith` derives the pooled instance only on a miss, so canonicalizing a value that is already pooled costs nothing.

### Pooling the values a report is built of (commit 2)

`UpdateReportInterner` adds pools for `ConfigRef`, `InclExclRule`, `File`, `Artifact`, `ModuleID`, `Caller` and `ModuleReport` itself. Pooling the report is worth more than pooling its parts alone, because sharing it also shares its licenses vector, extraAttributes map, homepage string and artifacts vector.

```scala
def intern(mr: ModuleReport): ModuleReport =
  // A publicationDate is a mutable Calendar, so canonicalize the report but never pool it.
  if mr.publicationDate.isDefined then canonicalize(mr)
  else moduleReports.internWith(mr)(canonicalize)
```

Everything else reachable from a `ModuleReport` is immutable, so sharing is semantically invisible.

Pooling happens at the points where reports come into existence, rather than as a sweep over a finished report, so only the module currently being built is ever un-pooled:

**1. As the resolver builds each one** — `SbtUpdateReport.moduleReport`:

```scala
UpdateReportInterner.intern(rep.withPublicationDate(publicationDate)….withCallers(callers.toVector))
```

**2. As `update` strips the callers** — `LibraryManagement.transformDetails`. Coursier memoizes `moduleReport` on a key that includes the dependees, so each project builds its own instance of a shared coordinate; the strip is what makes them value-equal, so it is the first moment they can collapse to one instance. Rebuilding without re-pooling would have thrown the sharing away:

```diff
-          val mrs1 = mrs0 map { _.withCallers(Vector()) }
+          val mrs1 = mrs0 map { mr =>
+            if (mr.callers.isEmpty) mr
+            else UpdateReportInterner.intern(mr.withCallers(Vector()))
+          }
```

**3. As the cache streams one back** — `UpdateReportPersistence.internModules`, a pass after decoding. The generated `UpdateReportLiteFormat` resolves the nested `JsonFormat[ModuleReport]` through its own self-type, so a `given` in that scope cannot reach it and there is no hook to intern as the reports stream; rerouting it would mean hand-mirroring the field layout of three generated container formats and keeping them in sync forever.

**4. As `toLite` prepares one to be written** — `JsonUtil.toLite` rebuilt every `ModuleReport` at every position, discarding the sharing just established. `InMemoryCacheStore.write` keeps the value it was handed rather than the bytes, so a subsequent read hands that rebuilt copy straight back and the cached path never shared at all. It now passes the instance through when caller filtering changed nothing, which is every module on the `update` path:

```diff
+              val callers = filterOutArtificialCallers(mr.callers)
+              if (callers eq mr.callers) mr
+              else
                 ModuleReport(…, callers)
```

## Result

**Live objects after the task** (`GC.class_histogram`, live only), measured on the full stack of #9515:

| class | 2.0.4 | with pooling |
|---|---|---|
| `ModuleReport` | 74,049 | **158** |
| `InclExclRule` | 396,714 | **3,367** |
| `Artifact` | 74,126 | **6,538** |
| `ModuleID` | 86,081 | **12,329** |
| `java.io.File` | 87,833 | **20,198** |

## Validation

New tests:

| file | covers |
|---|---|
| `lm-core/.../WeakInternerSpec.scala` | pool semantics, `internWith` deriving only on a miss, entries surviving collection of a value that probed for them, concurrent interning converging |
| `lm-core/.../UpdateReportInternerSpec.scala` | pooling per type, dated reports canonicalized but not pooled, release once unreferenced |
| `sbt-app/src/sbt-test/dependency-management/update-report-interning/` | scripted: two projects resolving one coordinate end up holding a single `ModuleReport` instance — on the fresh path *and* from cache |

| suite | result |
|---|---|
| `lmCore/Test/testFull` | 606 + 5 passed |
| `mainProj/Test/testFull` | 222 + 9 passed, 1 ignored |
| `lmCoursier/Test/testFull` | 86 passed |
| `utilCache/Test/testFull` | 58 passed |
| `scripted dependency-management/update-report-interning` | passed |
| `scripted dependency-management/update-report-cache-persistence` | passed |

**986 tests, 0 failures, 0 errors.**

The scripted test was run with a negative control: with the change reverted it fails with `two projects resolving one coordinate must share a single ModuleReport instance`.

### DIFF
--- a/lm-core/src/main/scala/sbt/internal/librarymanagement/JsonUtil.scala
+++ b/lm-core/src/main/scala/sbt/internal/librarymanagement/JsonUtil.scala
@@ -40,26 +40,30 @@ private[sbt] object JsonUtil {
             oar.organization,
             oar.name,
             oar.modules map { mr =>
-              ModuleReport(
-                mr.module,
-                mr.artifacts,
-                mr.missingArtifacts,
-                mr.status,
-                mr.publicationDate,
-                mr.resolver,
-                mr.artifactResolver,
-                mr.evicted,
-                mr.evictedData,
-                mr.evictedReason,
-                mr.problem,
-                mr.homepage,
-                mr.extraAttributes,
-                mr.isDefault,
-                mr.branch,
-                mr.configurations,
-                mr.licenses,
-                filterOutArtificialCallers(mr.callers)
-              )
+              val callers = filterOutArtificialCallers(mr.callers)
+              // Reuse the instance when filtering changed nothing, so interning is not undone here.
+              if (callers eq mr.callers) mr
+              else
+                ModuleReport(
+                  mr.module,
+                  mr.artifacts,
+                  mr.missingArtifacts,
+                  mr.status,
+                  mr.publicationDate,
+                  mr.resolver,
+                  mr.artifactResolver,
+                  mr.evicted,
+                  mr.evictedData,
+                  mr.evictedReason,
+                  mr.problem,
+                  mr.homepage,
+                  mr.extraAttributes,
+                  mr.isDefault,
+                  mr.branch,
+                  mr.configurations,
+                  mr.licenses,
+                  callers
+                )
             }
           )
         }
--- a/lm-core/src/main/scala/sbt/internal/librarymanagement/UpdateReportInterner.scala
+++ b/lm-core/src/main/scala/sbt/internal/librarymanagement/UpdateReportInterner.scala
@@ -1,0 +1,74 @@
+/*
+ * sbt
+ * Copyright 2023, Scala center
+ * Copyright 2011 - 2022, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.internal.librarymanagement
+
+import java.io.File
+import sbt.librarymanagement.*
+
+/**
+ * Intern pools for the immutable `UpdateReport` components to save memory.
+ *
+ * The pools are weak, so an entry lives only as long as some report references it.
+ */
+object UpdateReportInterner {
+
+  private val configRefs = new WeakInterner[ConfigRef]
+  private val rules = new WeakInterner[InclExclRule]
+  private val artifacts = new WeakInterner[Artifact]
+  private val moduleIds = new WeakInterner[ModuleID]
+  private val callers = new WeakInterner[Caller]
+  private val files = new WeakInterner[File]
+  private val moduleReports = new WeakInterner[ModuleReport]
+
+  def intern(c: ConfigRef): ConfigRef = configRefs.intern(c)
+
+  def intern(r: InclExclRule): InclExclRule = rules.intern(r)
+
+  def intern(f: File): File = files.intern(f)
+
+  def intern(a: Artifact): Artifact =
+    artifacts.internWith(a) { artifact =>
+      // Canonicalize the nested vectors so value-equal artifacts share their pieces even when only
+      // encountered once.
+      if (artifact.configurations.isEmpty) artifact
+      else artifact.withConfigurations(artifact.configurations.map(intern))
+    }
+
+  def intern(m: ModuleID): ModuleID =
+    moduleIds.internWith(m) { moduleId =>
+      if (
+        moduleId.inclusions.isEmpty && moduleId.exclusions.isEmpty &&
+        moduleId.explicitArtifacts.isEmpty
+      ) moduleId
+      else
+        moduleId
+          .withInclusions(moduleId.inclusions.map(intern))
+          .withExclusions(moduleId.exclusions.map(intern))
+          .withExplicitArtifacts(moduleId.explicitArtifacts.map(intern))
+    }
+
+  def intern(c: Caller): Caller =
+    callers.internWith(c) { caller =>
+      caller
+        .withCaller(intern(caller.caller))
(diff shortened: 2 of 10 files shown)


eed3si9n:
Thanks for the contribution!

lm-core/src/main/scala/sbt/internal/librarymanagement/JsonUtil.scala
```diff
@@ -40,26 +40,35 @@ private[sbt] object JsonUtil {
...
-                mr.licenses,
-                filterOutArtificialCallers(mr.callers)
-              )
+              val callers = filterOutArtificialCallers(mr.callers)
+              // Reuse the instance when filtering changed nothing, which is every module on the
+              // `update` path since `transformDetails` has already stripped callers. Rebuilding
+              // unconditionally allocated a copy per module per configuration and discarded whatever
+              // instance sharing the caller had established -- which is what the writer's identity
+              // probe settles a repeat by, and what the in-memory cache store then holds onto, since
+              // it keeps the value it was handed to write.
```
eed3si9n: Could we keep the inline comments more concise plz, esp if they are being generated.

lm-core/src/main/scala/sbt/internal/librarymanagement/UpdateReportInterner.scala
```diff
@@ -0,0 +1,106 @@
...
+ * canonicalized but never pooled, so no two holders can ever share that calendar.
+ *
+ * The pools are WEAK: a canonical instance is held only weakly, so a value survives exactly as long as
+ * some live report references it and is reclaimed once none do. During a build the hundreds of
+ * in-flight reports keep their coordinates reachable, so dedup is fully effective; once those reports
+ * are released -- including across a `reload` in a long-lived server -- the pooled entries are
+ * collected rather than accumulating the union of every dependency set ever seen.
+ *
+ * Not `private[sbt]`: the coursier report builder (`lmcoursier.internal.SbtUpdateReport`, a different
+ * top-level package) interns each `ModuleReport` as it constructs it.
```
eed3si9n: Same here. This can be shortened to

```suggestion
 * Intern pools for the immutable `UpdateReport` components to save memory.
```

lm-core/src/main/scala/sbt/internal/librarymanagement/UpdateReportInterner.scala
```diff
@@ -0,0 +1,106 @@
...
+   *
+   * A report carrying a `publicationDate` is canonicalized but NOT pooled: that `java.util.Calendar`
+   * is mutable, so aliasing the report between two holders would alias the calendar with it.
+   * Everything else reachable from a `ModuleReport` is immutable, which is what makes the rest safe.
+   *
+   * Canonicalizing is what `internWith` exists for. It rebuilds six fields and maps four vectors, and
+   * an already-pooled report -- the common case, since a build resolves the same coordinates over and
+   * over -- would have all of that discarded. The pool is probed with the report as it arrives, which
+   * finds the pooled twin regardless, because equality is structural.
+   */
```
eed3si9n: This Scaladoc can be removed since it's mostly describing the impl details.

lm-core/src/main/scala/sbt/internal/librarymanagement/UpdateReportPersistence.scala
```diff
@@ -24,6 +24,34 @@ final case class UpdateReportCache(
...
+  /**
+   * Canonicalizes the modules of a decoded cache, so a report streamed back shares its values with
+   * every other report the build holds, exactly as a freshly resolved one does.
+   *
+   * A pass afterwards rather than as the reports stream: the generated `UpdateReportLiteFormat`
+   * resolves the nested `JsonFormat[ModuleReport]` through its own self-type, so a `given` in this
+   * scope cannot reach it, and rerouting it would mean hand-mirroring the field layout of three
+   * generated container formats and keeping them in sync forever. The result is identical, only with
+   * more transient garbage.
+   */
```
eed3si9n, line 34: Same here.

lm-core/src/main/scala/sbt/internal/librarymanagement/JsonUtil.scala
hoangmaihuy: I'm sorry for that, added as a rule in my CLAUDE.md. I will review generated inline comments carefully next time 😄

eed3si9n: No worries. We do have "Avoid inline comments!" in https://github.com/sbt/sbt/blob/develop/AGENTS.md. Let me know if we can improve the file somehow to let it listen more.

eed3si9n (approved):
Thanks!

PR #9524, opened 2026-07-28
[2.x] perf: Improve update performance by avoiding artifact content hashing

hoangmaihuy:

Part of #9514 (§5). Split out of #9515 so it can be reviewed on its own; it is independent of the other parts.

## Problem

sjsonnew serializes a `File` as a `(uri, Long)` pair whose `Long` is `HashUtil.sha256ToLong(file.toPath())` — **a full content hash of the file**. Nothing reads it back: `IsoStringLong[File].from` parses the URI and drops the `Long`.

Staleness does not come from it either. `update` decides that in `LibraryManagement.fileUptodate`, which checks `File.exists` and the modification time against `UpdateReport.stamps`.

Meanwhile a report names each artifact once per configuration it resolved in, and the projects of a build largely share their dependencies, so writing the caches re-reads the whole downloaded classpath many times over. On a 302-module monorepo: **755 GB of jars read and about 11 minutes of CPU**, for bytes no reader looks at.

## Change

A codec nested in `UpdateReportPersistence` — the only thing that uses it — extends the `LibraryManagementCodec` **trait** and overrides the inherited `fileStringLongIso` so the pair carries `0`:

```scala
object UpdateReportPersistence:

  private[sbt] object CacheCodec extends LibraryManagementCodec:
    override implicit lazy val fileStringLongIso: IsoStringLong[File] =
      IsoStringLong.iso[File](
        (f: File) => (IO.toURI(f).toASCIIString, 0L),
        (p: (String, Long)) => IO.toFile(new URI(p._1))
      )

  // Not the stock `LibraryManagementCodec`: see `CacheCodec` for why persisted reports must not
  // content-hash the artifacts they name.
  import CacheCodec.given
```

`fileStringLongIso` is an `implicit lazy val` in `sjsonnew.FileIsoStringLongs`, so it is a **virtual member**: every generated format resolves `JsonFormat[File]` as `isoStringLongFormat[File](fileStringLongIso)` through its self-type. Overriding it therefore also reaches the `Vector[(Artifact, File)]` nested inside the generated `ModuleReportFormat`, which a locally-scoped `JsonFormat[File]` could not.

Everything else — including the `inputs` store, so `Tracked.inputChanged` still hashes contents for invalidation — keeps the stock `LibraryManagementCodec` object.

## Compatibility

The JSON shape is unchanged; only the `Long`'s value is. Caches stay readable in both directions: ones written by a version that still writes the hash load here, and ones written here load there. A test pins both directions.

## Validation

New tests — `lm-core/.../UpdateReportCacheCodecSpec.scala`:

| test | covers |
|---|---|
| writing a report does not depend on the content of the files it references | rewrite a jar's bytes, serialize again, bytes must be identical |
| a `File` round trips to the same path | the URI half still inverts |
| a cache written with a real content hash still reads | fixture written with the stock codec (asserted to carry a non-zero hash) decodes here |
| the cache `UpdateReportPersistence` writes carries no content hash | pins the *wiring*, not the codec — writes through `writeTo` and asserts every `"second"` in the output JSON is `0` |

The last one matters because the override only takes effect through the generated formats' self-type: a version that quietly resolved through the stock codec would still compile, and would still pass the first three, since those pass the codec explicitly with `using`. Reverting the import to `LibraryManagementCodec.given` fails that test and only that test.

| suite | result |
|---|---|
| `lmCore/Test/testFull` | 591 + 5 passed |
| `mainProj/Test/testFull` | 222 + 9 passed, 1 ignored |
| `scripted dependency-management/update-report-cache-persistence` | passed |

### DIFF
--- a/lm-core/src/main/scala/sbt/internal/librarymanagement/UpdateReportPersistence.scala
+++ b/lm-core/src/main/scala/sbt/internal/librarymanagement/UpdateReportPersistence.scala
@@ -9,11 +9,12 @@
 package sbt.internal.librarymanagement
 
 import java.io.File
+import java.net.URI
 import scala.util.Try
-import sjsonnew.{ Builder, JsonFormat, Unbuilder, deserializationError }
+import sjsonnew.{ Builder, IsoStringLong, JsonFormat, Unbuilder, deserializationError }
+import sbt.io.IO
 import sbt.util.CacheStore
 import sbt.librarymanagement.*
-import sbt.librarymanagement.LibraryManagementCodec.given
 
 final case class UpdateReportCache(
     lite: UpdateReportLite,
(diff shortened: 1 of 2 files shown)


lm-core/src/main/scala/sbt/internal/librarymanagement/UpdateReportCacheCodec.scala
```diff
@@ -0,0 +1,49 @@
...
+ * `fileStringLongIso` is an `implicit lazy val` in `sjsonnew.FileIsoStringLongs`, so it is a virtual
+ * member and every generated format resolves `JsonFormat[File]` as
+ * `isoStringLongFormat[File](fileStringLongIso)` through its self-type. Overriding it here therefore
+ * also reaches the `Vector[(Artifact, File)]` nested inside the generated `ModuleReportFormat`, which
+ * a locally-scoped `JsonFormat[File]` could not.
+ *
+ * The JSON shape is unchanged -- only the Long's value is -- so caches stay readable by sbt versions
+ * that still write the hash, and the ones written here stay readable by them.
+ */
+private[sbt] object UpdateReportCacheCodec extends LibraryManagementCodec:
```
eed3si9n: If this is used only by UpdateReportPersistence, could we move it to inside `object UpdateReportPersistence`?

eed3si9n (approved):
Thanks!

PR #9526, opened 2026-07-28
[2.x] fix: Don't freeze the server when a terminal-properties response is malformed or slow

### LINKED ISSUE #9525: sbt server freezes for all clients when one client's terminal-properties response is malformed or slow
One attached client that answers the server's sbt/terminalpropertiesquery request with a malformed or error response, or takes longer than 5 seconds to answer, freezes the whole server: prompts, Ctrl-C, and command dispatch stop for every connected client until the offending client disconnects.

The response handler in VirtualTerminal drops a response it cannot convert instead of falling back to a default like the sibling handlers, so the updater thread behind NetworkTerminal.getProperties times out after its one-shot 5 second poll leaving properties null. Waiters see null and wait again with no updater outstanding, the 1 second refresh throttle can put a caller to sleep with no query in flight at all, and a response arriving after the timeout is delivered into a queue that was never deregistered, so it wakes nobody. The parked callers include the command loop iterating channels and the fast-track thread handling attach and cancel, which is what makes the freeze server-wide.

Reproduced deterministically on develop with a raw protocol session that attaches interactively and answers every server request with a result of the wrong shape: a well-behaved batch client is then never served within three minutes. A version-skewed or buggy client hits the malformed-response path outright; a GC pause or a loaded CI machine can hit the slow-response path.

PR with a fix and the regression test to follow shortly.

BrianHotopp:
## Problem

Fixes #9525

One attached client answering the `sbt/terminalpropertiesquery` request badly, or slower than 5 seconds, could freeze the whole server for every client:

- The response handler dropped malformed responses (`response.foreach(buffer.put)`) instead of falling back to a default like every sibling handler, so the updater thread waiting on the queue timed out with the properties reference still null.
- `getProperties(block = true)` waits while properties is null, but nothing completes it after the updater's one-shot 5-second poll times out: waiters woke from the notify, saw null, and waited again with no updater outstanding. The 1-second `lastUpdate` throttle also let a caller start waiting with no query in flight at all, and the wait condition was checked outside the pending monitor, losing wakeups that fired between the check and the wait.
- A response arriving after the poll timeout was delivered into a queue that was never deregistered, so it neither set properties nor woke anyone.

The threads that block here include the command loop iterating channels and the fast-track thread handling attach and cancel, so one bad or briefly-stalled client wedged prompts, Ctrl-C, and command dispatch server-wide until that client disconnected.

## Fix

- The properties response handler now falls back to a default like its siblings.
- The updater expires its query on timeout, deregistering it (rescuing a response that raced in) and completing properties with the empty default, so waiters always make progress.
- Both wait sites hold the pending monitor and gate on the query in flight. `waitForPending` gets the same treatment, since it seeds lazy vals whose initialization otherwise parks every thread touching them.

## Testing

- `TerminalPropertiesFreezeTest` (server-test): a raw-protocol session that answers every server request with a result of the wrong shape attaches interactively; a well-behaved batch client must then still be served twice. Fails on develop with a three-minute timeout (the server is frozen); passes with this change.
- `VirtualTerminalSpec` unit spec pins the expiry semantics, including the late-response rescue.
- `RebootTest` now shares the batch-client helpers moved into `AbstractServerTest.runBatchClient`; still green.
- `mainProj` scalafmt and MiMa clean.

---

*Diagnosed and fixed with AI assistance (Claude Code); the freeze was reproduced deterministically against a forked server with a raw-protocol client, and the fix verified by the regression test above failing on develop and passing on this branch.*

### DIFF
--- a/main/src/main/scala/sbt/internal/server/NetworkChannel.scala
+++ b/main/src/main/scala/sbt/internal/server/NetworkChannel.scala
@@ -754,21 +754,31 @@ final class NetworkChannel(
           pending.set(true)
           val queue = VirtualTerminal.sendTerminalPropertiesQuery(term.name, jsonRpcRequest)
           val update: Runnable = () => {
-            queue.poll(5, java.util.concurrent.TimeUnit.SECONDS) match {
-              case null =>
-              case t    => properties.set(t)
-            }
-            pending.synchronized {
-              lastUpdate.set(Deadline.now)
-              pending.set(false)
-              pending.notifyAll()
+            try {
+              queue.poll(5, java.util.concurrent.TimeUnit.SECONDS) match {
+                case null =>
+                  VirtualTerminal.expireTerminalPropertiesQuery(term.name, queue) match {
+                    case Some(late) => properties.set(late)
+                    case None       => Util.ignoreResult(properties.compareAndSet(null, empty))
+                  }
+                case t => properties.set(t)
+              }
+            } finally {
+              pending.synchronized {
+                lastUpdate.set(Deadline.now)
+                pending.set(false)
+                pending.notifyAll()
+              }
             }
           }
           new Thread(update, s"network-terminal-${term.name}-update") {
             setDaemon(true)
           }.start()
         }
-        while (block && properties.get == null) pending.synchronized(pending.wait())
+        // The updater clears pending inside this monitor before notifying.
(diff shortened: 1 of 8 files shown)


eed3si9n (approved):
Thanks!

PR #9527, opened 2026-07-29
[2.x] fix: Don't strand the server when a client dies with a terminal query unanswered

BrianHotopp:
## Problem

Refs #6840, #6841, #5643 (same stranding mechanism). Scope: this PR fixes the stranding and restores the prompt's input path to a live client (pinned by `FailedLoadPromptTest`). Still open from #6840: the prompt's [warn] text is not routed to the interactive client's log, so a user may not see the prompt even though they can answer it; that log-routing question is separate and unchanged here.

A client that dies with a terminal control query unanswered (raw mode, echo, attributes, size) or mid input-read can park a server thread forever. The parked thread can be the command loop: the failed-load prompt enters raw mode and reads a byte from the prompting client, so one dying client wedges every connected client.

Root causes, each verified by thread dump or isolated reproduction:

1. `VirtualTerminal.cancelRequests` drained 2 of the 8 pending terminal maps on disconnect; waiters on the other 6 parked forever.
2. Raw-mode requests registered in the set-echo map.
3. Closing a channel terminal did not wake readers parked on its input stream.
4. The failed-load prompt read its answer from `System.in`, which under non-virtual IO is the process's own stdin, not the client.
5. `ServerSessionImpl.close()` could not deliver EOF while its read thread was parked in a native read: close() returned, no EOF reached the peer, the server never noticed the disconnect.

## Fix

- `cancelRequests` drains all eight maps, with `offer` not `put`, so shutdown cannot block on a full queue.
- Raw-mode requests register in the raw-mode map.
- Closing a terminal input stream seeds EOF for parked readers.
- The prompt reads the active terminal's input stream (identical to `System.in` under virtual IO, correct under non-virtual IO).
- `ServerSessionImpl.close()` shuts down socket input first, waking the parked reader so the close delivers EOF.

## Testing

- `FailedLoadPromptTest`: an interactive client answers the failed-load prompt with 'r'; the retry must succeed and a follow-up exec must complete (fails with `System.in` on develop: the typed byte never reaches the prompt, and every later exec starves).
- `TerminalMapsDrainTest`: a raw-protocol client triggers the failed-load prompt, refuses the raw-mode query, and disconnects. On develop the command loop stays parked in `setRawMode` (thread dump) and a well-behaved client times out after 3 minutes; with this change the server exits cleanly (EOF maps to 'q' by design).
- `VirtualTerminalSpec`: the drain wakes all eight maps and leaves other channels parked.
- `WriteableInputStreamSpec`: close wakes a parked reader with EOF (fails without the fix).
- `ServerSessionImplSpec`: close delivers EOF with the read thread parked (fails without the fix; unix sockets only, named pipes differ).
- `TerminalPropertiesFreezeTest`, `RebootTest` green; scalafmt and MiMa clean on mainProj, protocolProj, utilLogging.

---

*Diagnosed and fixed with AI assistance (kimi-code/k3 via Oh My Pi); the wedge was reproduced deterministically against a forked server, and each link verified by thread dump or an isolated socket-level reproduction.*

### DIFF
--- a/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
+++ b/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala
@@ -640,6 +640,7 @@ object Terminal {
     }
     override def close(): Unit = if (closed.compareAndSet(false, true)) {
       executor.shutdownNow()
+      buffer.synchronized(buffer.put(-1: Integer))
       ()
     }
   }
--- a/internal/util-logging/src/test/scala/sbt/internal/util/WriteableInputStreamSpec.scala
+++ b/internal/util-logging/src/test/scala/sbt/internal/util/WriteableInputStreamSpec.scala
@@ -1,0 +1,28 @@
+/*
+ * sbt
+ * Copyright 2023, Scala center
+ * Copyright 2011 - 2022, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.internal.util
+
+import java.io.PipedInputStream
+import java.util.concurrent.{ LinkedBlockingQueue, TimeUnit }
+import verify.BasicTestSuite
+
+object WriteableInputStreamSpec extends BasicTestSuite:
+  test("close wakes a parked reader with EOF"):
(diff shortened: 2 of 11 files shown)


eed3si9n (approved):
Thanks!

PR #9531, opened 2026-07-30
[2.x] fix: Fixes forked run baseDirectory, take 2

### LINKED ISSUE #9530: [2.x] Compile / run / baseDirectory doesn't work
## steps

sbt 2.0.4

### build.sbt

```scala
scalaVersion := "3.8.4"

lazy val root = rootProject
  .autoAggregate

lazy val app = project
  .settings(
    Compile / run / baseDirectory := baseDirectory.value,
  )
```

### app/Hello.scala

```scala
package example

@main
def hello(arg: String*): Unit =
  val x = new java.io.File(".").getAbsolutePath
  println(s"hi $x")

```

## problem

The working directory is not changed.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/9530

## Problem
Forked run baseDirectory was changed to current directory in sbt 2.0.4, which on its own is fine, but it doesn't respect
Compile / run / baseDirectory.

## Solution
This fixes that.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -1428,7 +1428,11 @@ object Defaults extends BuildCommon with DefExtra {
 
   /** Fork options for run-like tasks: the forked process inherits sbt's working directory. */
   private[sbt] def runForkOptionsTask: Initialize[Task[ForkOptions]] =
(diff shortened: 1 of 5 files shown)


anatoliykmetyuk (approved):
LGTM!

PR #9533, opened 2026-07-30
[2.x] fix: Fixes publishing with sbt-pgp

### LINKED ISSUE #9532: [2.x] Ivyless publishing doesn't work with sbt-pgp
## steps

develop branch.

## problem

sbt-pgp isn't well tested. If we call publishSigned it hits an error message about requiring Ivy.

## expectation

It works without Ivy.

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/9532

## Problem
Ivyless publishing doesn't work with sbt-pgp.
In general the publishing related code go messier when Ivyless publishing got contributed.

## Solution
This refactors the ivyless publishing code and revives `publishOrSkip`, which is called by the plugin.


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -3085,9 +3085,9 @@ object Classpaths {
           .map(m => d.withRevision(m.module.revision))
       }.distinct
     }.value,
-    publish := LibraryManagement.ivylessPublishTask.tag(Tags.Publish, Tags.Network).value,
-    publishLocal := LibraryManagement.ivylessPublishLocalTask.value,
-    publishM2 := LibraryManagement.ivylessPublishM2Task.tag(Tags.Publish, Tags.Network).value,
+    publish := publishOrSkip(publishConfiguration, publish / skip).value,
+    publishLocal := publishOrSkip(publishLocalConfiguration, publishLocal / skip).value,
+    publishM2 := publishOrSkip(publishM2Configuration, publishM2 / skip).value,
```
eed3si9n, line 3090: This restores `pulishOrSkip(...)` similar to sbt 1.x - https://github.com/sbt/sbt/blob/v1.12.14/main/src/main/scala/sbt/Defaults.scala#L2970-L2973 by implementing a generic publisher driven by the publish configuration, rather than rewiring to different tasks.

main/src/main/scala/sbt/internal/librarymanagement/GenericPublisher.scala
```diff
@@ -0,0 +1,609 @@
...
+        configuration.allRepositories,
+      )
+    val processed = configuration.process(pomXml)
+    val printer = new scala.xml.PrettyPrinter(1000, 4)
+    val formatted = scala.xml.XML.loadString(printer.format(processed))
+    scala.xml.XML.save(file.getAbsolutePath, formatted, "UTF-8", xmlDecl = true)
+    log.info("Wrote " + file.getAbsolutePath)
+    file
+
+  override def publish(
```
eed3si9n, line 77: Most of the code in this file is moved from LibraryManagement object. The only new stuff is this method, which would be called by `publish`, `publishLocal`, and `publishM2`.

main/src/main/scala/sbt/internal/librarymanagement/GenericPublisher.scala
```diff
@@ -0,0 +1,609 @@
...
+      project: CsrProject,
+      credentials: Seq[Credentials],
+      resolvers: Seq[Resolver]
+  ): GenericPublisher =
+    apply(
+      dependencyResolution,
+      pomRepositories,
+      project,
+      credentials,
+      resolvers.map(r => r.name -> r).toMap
```
anatoliykmetyuk: This may introduce a collision between `publishTo` and `publishLocal`. If the user explicitly supplies `publishTo := Some("local" at "https://example.com/releases")`, `resolvers` will contain name `local` coming from the local resolver and name `local` coming from the user. Since user-configured name comes last, it will win during map creation, leading to `publishLocal` publishing to a wrong place (`publishTo` one).

sbt-app/src/sbt-test/plugins/pgp/test
```diff
@@ -1 +1,4 @@
-> plugins
+> publishSigned
+
+$ exists target/**/app_3-1.0.jar.asc
```
anatoliykmetyuk: This verifies the signature was produced, but not that it was published to the correct place. I.e. out of two steps - signing and publishing - only the signing step is verified.

main/src/main/scala/sbt/internal/librarymanagement/GenericPublisher.scala
eed3si9n: Is your suggestion to keep resolvers as `Seq[Resolver]`? If so I'm +1 with that.

sbt-app/src/sbt-test/plugins/pgp/test
eed3si9n: Addressed in dbbac672fa18724218d1b4a42be405874de4bc9c

main/src/main/scala/sbt/internal/librarymanagement/GenericPublisher.scala
eed3si9n: Addressed in https://github.com/sbt/sbt/pull/9533/changes/05c200efc6d2049a99dfbbd1ae3a16311a631175

anatoliykmetyuk: This changes the search to first-wins, introducing the opposite problem: if `publishTo` is specified as `"local"`, `publishSigned` will publish to `publishLocal` target instead.

I think the best option is to detect duplicate keys and fail with an exception, possibly indicating reserved resolver names.

eed3si9n: Addressed in 36f9bc1370c7c3c8a67d94a2cccdf2a63e5f9d1d

anatoliykmetyuk (approved):
LGTM

PR #9534, opened 2026-07-30
[2.x] fix: Preserve '=' in remote cache header values

### LINKED ISSUE #9528: [2.x] remoteCacheHeaders silently drops trailing '=' from header values, disabling the remote cache without error
TLDR: sbt 2's remoteCacheHeaders splits on every =, so Basic-auth base64 padding gets silently stripped — the credential is rejected, the build still passes, and the remote cache never populates.

## steps

1. Run a Bazel-compatible cache that requires auth for writes and allows anonymous reads —
   e.g. [bazel-remote](https://github.com/buchgr/bazel-remote):

   ```bash
   bazel-remote --dir /data --max_size 2 \
     --grpc_address 0.0.0.0:9092 --http_address 0.0.0.0:8080 \
     --htpasswd_file htpasswd --allow_unauthenticated_reads
   ```

   It logs: `Access mode: authentication required for writes, unauthenticated reads allowed`

2. `project/plugins.sbt`:

   ```scala
   addRemoteCachePlugin
   ```

3. `build.sbt` — note the base64 credential ends in `==`:

   ```scala
   ThisBuild / scalaVersion := "3.8.4"
   Global / remoteCache := Some(uri("grpc://localhost:9092"))
   Global / remoteCacheHeaders += "authorization=Basic Y2FjaGVwdXNoOnMzY3JldC1zcGlrZQ=="
   ```

4. `sbt compile` with a cold disk cache.

sbt version: 2.0.4. JDK 25 (Temurin), macOS.

## problem

The build succeeds, but nothing is ever written to the remote cache. The server reports
`UNAUTHENTICATED` on `BatchUpdateBlobs` and the cache stays at 0 files. No warning or
error appears in the sbt output, and `show Global/remoteCacheHeaders` prints the header
back verbatim, so the configuration looks correct.

The cause is in `GrpcActionCacheStore.AuthCallCredentials`:

```scala
val pairs = remoteHeaders.map: h =>
  h.split("=").toList match
    case List(k, v) => Metadata.Key.of(k, Metadata.ASCII_STRING_MARSHALLER) -> v
    case _          => sys.error("remote header must contain one '='")
```

`h.split("=")` splits on every `=`. Java's `split` discards *trailing* empty strings, so
`"authorization=Basic Y2FjaGVwdXNoOnMzY3JldC1zcGlrZQ=="` yields exactly two elements and
matches `List(k, v)` — with the base64 padding silently amputated:

```
parts = List(authorization, Basic Y2FjaGVwdXNoOnMzY3JldC1zcGlrZQ)
```

So the guard never fires and a corrupted credential is sent. Since action cache `put`
failures are non-fatal, the result is a green build with a permanently empty cache.

Two confirmations that the credential itself is fine:

- The same credential over HTTP works: `curl -u user:pass -X PUT --data-binary hello
  http://localhost:8080/cas/<sha256>` → `HTTP 200`.
- Choosing a password so that `user:password` is a multiple of 3 bytes (base64 with **no**
  padding) makes the identical setup work: `BatchUpdateBlobs` → `OK`, cache populates.

There is a second, opposite symptom: a value containing an *interior* `=` — e.g.
`x-api-key=ab=cd` — produces 3 elements and fails the build through `sys.error`, even
though it is a perfectly legal header value. `authorization=` (empty value) fails the
same way.

Summary of current behaviour:

| header | current result |
|---|---|
| `authorization=Basic dXNlcjpwdw==` | silently truncated to `Basic dXNlcjpwdw` |
| `authorization=Basic dXNlcjpwdzE=` | silently truncated |
| `x-api-key=ab=cd` | `sys.error` |
| `authorization=` | `sys.error` |
| `x-buildbuddy-api-key=abc123` | works |

This affects any backend using HTTP Basic auth, since base64 of `user:password` is padded
unless its length happens to be a multiple of 3.

## expectation

The header value should be preserved verbatim — everything after the first `=` is the
value. Only a header with no `=` at all should be an error.

## notes

Splitting on the first `=` only fixes both symptoms:

```diff
-      h.split("=").toList match
+      // Split on the first '=' only. Splitting on every '=' would drop trailing
+      // padding from values such as Basic auth credentials ("Basic dXNlcjpwdw==")
+      // and reject values that legitimately contain '='.
+      h.split("=", 2).toList match
         case List(k, v) => Metadata.Key.of(k, Metadata.ASCII_STRING_MARSHALLER) -> v
-        case _          => sys.error("remote header must contain one '='")
+        case _          => sys.error("remote header must contain '='")
```

I have this applied locally on top of `develop` with a unit test in
`GrpcActionCacheStoreTest` covering base64 padding (one and two `=`), an interior `=`,
and the remaining no-separator error case. With the fix reverted the test fails
(`value == "Basic dXNlcjpwdw=="`, actual `Basic dXNlcjpwdw`); with it applied,
`Total 3, Passed 3`, and `remoteCacheProj/mimaReportBinaryIssues` is clean.

End-to-end against bazel-remote (`0027c3f`) with sbt 2.0.4 and a `==`-padded credential,
publishing the patched module locally:

| | stock 2.0.4 | patched |
|---|---|---|
| `BatchUpdateBlobs` | `Unauthenticated` | `OK` ×5 |
| `UpdateActionResult` | not attempted | `OK` ×8 |
| cache contents | 0 files | 13 files / 61 KB |

Happy to open a PR with the fix and the test if a contribution is wanted here — per
CONTRIBUTING I'm checking first rather than sending one unsolicited.

_AI disclosure: Claude Opus 5 (Claude Code) was used to diagnose this, write the patch
and test, and run the verification above._

KilianSwissborg:
Fixes #9528.

**Problem**
`remoteCacheHeaders` is parsed with `h.split("=")`, which splits on every `=`. Java's
`split` drops trailing empty strings, so `authorization=Basic dXNlcjpwdw==` still matches
`List(k, v)` — with the base64 padding silently removed. The server rejects the truncated
credential with UNAUTHENTICATED, and since cache `put` failures are non-fatal the build
succeeds with a permanently empty cache and nothing logged. Separately, a value with an
interior `=` (`x-api-key=ab=cd`) fails the build via `sys.error`.

**Solution**
Split on the first `=` only. The error case narrows to a header with no `=` at all.

Unit test added; it fails without the fix (`Basic dXNlcjpwdw` vs `Basic dXNlcjpwdw==`).
Also verified end to end against bazel-remote with a `==`-padded credential: before, 0
files cached and UNAUTHENTICATED on `BatchUpdateBlobs`; after, 13 files and OK.
`mimaReportBinaryIssues` clean.

_AI disclosure: Claude Opus 5 (Claude Code) was used to diagnose the issue, write the fix
and test, and run the verification. I reviewed the change and confirmed the repro locally._

### DIFF
--- a/sbt-remote-cache/src/main/scala/sbt/internal/GrpcActionCacheStore.scala
+++ b/sbt-remote-cache/src/main/scala/sbt/internal/GrpcActionCacheStore.scala
@@ -142,9 +142,12 @@ object GrpcActionCacheStore:
 
   class AuthCallCredentials(remoteHeaders: List[String]) extends CallCredentials:
     val pairs = remoteHeaders.map: h =>
-      h.split("=").toList match
+      // Split on the first '=' only. Splitting on every '=' would drop trailing
+      // padding from values such as Basic auth credentials ("Basic dXNlcjpwdw==")
+      // and reject values that legitimately contain '='.
+      h.split("=", 2).toList match
         case List(k, v) => Metadata.Key.of(k, Metadata.ASCII_STRING_MARSHALLER) -> v
-        case _          => sys.error("remote header must contain one '='")
+        case _          => sys.error("remote header must contain '='")
     override def applyRequestMetadata(
         requestInfo: CallCredentials.RequestInfo,
         executor: java.util.concurrent.Executor,
--- a/sbt-remote-cache/src/test/scala/sbt/internal/GrpcActionCacheStoreTest.scala
+++ b/sbt-remote-cache/src/test/scala/sbt/internal/GrpcActionCacheStoreTest.scala
@@ -36,6 +36,27 @@ object GrpcActionCacheStoreTest extends verify.BasicTestSuite:
     // Distinct Deadline instances derived at call time, not a single shared frozen one.
     assert(!deadline1.eq(deadline2))
 
+  // Regression test: header values may legitimately contain '=' -- Basic auth credentials
+  // end in base64 padding. Splitting on every '=' silently truncated the value, so the
+  // server rejected the credential with UNAUTHENTICATED while the build still succeeded,
+  // leaving the cache permanently empty with no error reported.
+  test("header values retain '=' such as base64 padding"):
+    val twoPad = GrpcActionCacheStore.AuthCallCredentials(List("authorization=Basic dXNlcjpwdw=="))
+    val (key, value) = twoPad.pairs.head
+    assert(key.name == "authorization")
+    assert(value == "Basic dXNlcjpwdw==")
+
+    val onePad = GrpcActionCacheStore.AuthCallCredentials(List("authorization=Basic dXNlcjpwdzE="))
+    assert(onePad.pairs.head._2 == "Basic dXNlcjpwdzE=")
+
+    // An interior '=' is part of the value, not a second separator.
+    val interior = GrpcActionCacheStore.AuthCallCredentials(List("x-api-key=ab=cd"))
+    assert(interior.pairs.head._2 == "ab=cd")
+
+    // No '=' at all remains an error.
+    intercept[RuntimeException]:
+      GrpcActionCacheStore.AuthCallCredentials(List("bogus")).pairs
+
   private def newStore(): GrpcActionCacheStore =
     val base = Files.createTempDirectory("grpc-action-cache-test")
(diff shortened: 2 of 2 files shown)


eed3si9n (approved):
Thanks!

eed3si9n:
Could you sign the Scala CLA please?: <https://contribute.akka.io/contribute/cla/scala>

PR #9538, opened 2026-08-01
[2.x] fix: Fixes AccessDeniedException issue on Windows

### LINKED ISSUE #9537: [2.x] java.nio.file.AccessDeniedException in (Compile / packageBin) after running test
## steps

- Check out: https://github.com/kalejami/sbt_2_access_denied_exception
- Or create any sbt project with sbt 2.x and at least one file to compile.
- Start sbt in a console
- Run compile -> test -> test

## problem

Getting AccessDeniedException 

```
[error] stack trace is suppressed; run last Compile / packageBin for the full output
[error] (Compile / packageBin) java.nio.file.AccessDeniedException: ..\sbt_2_access_denied_exception\target\out\jvm\scala-3.8.4\sbt_2_access_denied_exception\sbt_2_access_denied_exception_3-0.1.0-SNAPSHOT.jar.98b62d55.tmp -> ..\sbt_2_access_denied_exception\target\out\jvm\scala-3.8.4\sbt_2_access_denied_exception\sbt_2_access_denied_exception_3-0.1.0-SNAPSHOT.jar
```

And here the output from `last Compile / packageBin`

```
sbt:sbt_2_access_denied_exception> last Compile / packageBin
[debug] Packaging ..\sbt_2_access_denied_exception\target\out\jvm\scala-3.8.4\sbt_2_access_denied_exception\sbt_2_access_denied_exception_3-0.1.0-SNAPSHOT.jar ...
[debug] Input file mappings:
[debug]         Test$.class
[debug]           ..\sbt_2_access_denied_exception\target\out\jvm\scala-3.8.4\sbt_2_access_denied_exception\classes\Test$.class
[debug]         Test.class
[debug]           ..\sbt_2_access_denied_exception\target\out\jvm\scala-3.8.4\sbt_2_access_denied_exception\classes\Test.class
[debug]         Test.tasty
[debug]           ..\sbt_2_access_denied_exception\target\out\jvm\scala-3.8.4\sbt_2_access_denied_exception\classes\Test.tasty
[error] java.nio.file.AccessDeniedException: ..\sbt_2_access_denied_exception\target\out\jvm\scala-3.8.4\sbt_2_access_denied_exception\sbt_2_access_denied_exception_3-0.1.0-SNAPSHOT.jar.29f7977a.tmp -> ..\sbt_2_access_denied_exception\target\out\jvm\scala-3.8.4\sbt_2_access_denied_exception\sbt_2_access_denied_exception_3-0.1.0-SNAPSHOT.jar
[error]         at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:89)
[error]         at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
[error]         at java.base/sun.nio.fs.WindowsFileCopy.move(WindowsFileCopy.java:310)
[error]         at java.base/sun.nio.fs.WindowsFileSystemProvider.move(WindowsFileSystemProvider.java:287)
[error]         at java.base/java.nio.file.Files.move(Files.java:1319)
[error]         at sbt.io.IO$.writeFileAtomically$$anonfun$1(IO.scala:485)
[error]         at sbt.internal.io.Retry$.impl(Retry.scala:114)
[error]         at sbt.internal.io.Retry$.apply(Retry.scala:58)
[error]         at sbt.internal.io.Retry$.apply(Retry.scala:40)
[error]         at sbt.internal.io.Retry$.apply(Retry.scala:30)
[error]         at sbt.io.IO$.writeFileAtomically(IO.scala:486)
[error]         at sbt.io.IO$.archive(IO.scala:710)
[error]         at sbt.io.IO$.jar(IO.scala:675)
[error]         at sbt.Pkg$.makeJar(Pkg.scala:232)
[error]         at sbt.Pkg$.apply(Pkg.scala:159)
[error]         at sbt.Defaults$.packageTask$lzyINIT1$$anonfun$1$$anonfun$1(Defaults.scala:1927)
[error]         at sbt.util.ActionCache$.organicTask$1(ActionCache.scala:100)
[error]         at sbt.util.ActionCache$.cache(ActionCache.scala:186)
[error]         at sbt.Defaults$.packageTask$lzyINIT1$$anonfun$1(Defaults.scala:1932)
[error]         at scala.Function1.$anonfun$compose$1(Function1.scala:79)
[error]         at scala.Function1.$anonfun$compose$1(Function1.scala:79)
[error]         at sbt.std.Transform$$anon$3.work(Transform.scala:79)
[error]         at sbt.std.Transform$$anon$3.work(Transform.scala:79)
[error]         at sbt.Execute.submit$$anonfun$1$$anonfun$1(Execute.scala:283)
[error]         at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:24)
[error]         at sbt.Execute.work(Execute.scala:294)
[error]         at sbt.Execute.submit$$anonfun$1(Execute.scala:283)
[error]         at sbt.ConcurrentRestrictions$$anon$4.$anonfun$2(ConcurrentRestrictions.scala:269)
[error]         at sbt.CompletionService$$anon$2.call(CompletionService.scala:75)
[error]         at sbt.CompletionService$$anon$2.call(CompletionService.scala:73)
[error]         at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
[error]         at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
[error]         at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
[error]         at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
[error]         at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
[error]         at java.base/java.lang.Thread.run(Thread.java:1474)
[error] (Compile / packageBin) java.nio.file.AccessDeniedException: ..\sbt_2_access_denied_exception\target\out\jvm\scala-3.8.4\sbt_2_access_denied_exception\sbt_2_access_denied_exception_3-0.1.0-SNAPSHOT.jar.29f7977a.tmp -> ..\sbt_2_access_denied_exception\target\out\jvm\scala-3.8.4\sbt_2_access_denied_exception\sbt_2_access_denied_exception_3-0.1.0-SNAPSHOT.jar
```

## expectation

No exception

## notes

I'm running this on windows. Just in case this problem is not reproducible on linux

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/9537

## Problem
When test classloader holds on to the JAR file, Windows gets AccessDefinedException on packageBin.

## Solution
Flip the default to close the test class loader.


kevin-lee:
This appears to solve the issue, but there are still some concerns. Here’s Claude’s report.

I don’t have enough knowledge of sbt, so I’m not sure how relevant it is, though.
***
Flipping the default looks right for the default `ScalaLibrary`/`Flat` strategies, but I think
the second hunk leaves a gap under `turbo := true` (`AllLibraryJars`):

`ReverseLookupClassLoaderHolder.closeThis` doesn't only control the cached holder — it is also
what `checkout` passes into the per-run `BottomClassLoader`:

- `ReverseLookupClassLoaderHolder.checkout` → `new BottomClassLoader(..., closeThis, ...)`
  (`LayeredClassLoaders.scala`)
- `BottomClassLoader` hands it to `ManagedClassLoader` as its `close` flag
  (`BottomClassLoader.java`)
- `ManagedClassLoader.close()` is a no-op when that flag is false

So with `closeThis = false` hardcoded, the `AllLibraryJars` test loader (the `BottomClassLoader`
holding the project jar) never really closes: the #9537 jar pin persists in turbo mode, and
`-Dsbt.classloader.close=true` can no longer reach it either (before this PR it could).

Hardcoding `false` makes sense for the *cached holder itself* (closing a shared cached layer
would break reuse), but the per-run bottom loader should probably still honor the
`closeClassLoaders` setting — e.g. thread `close` through `checkout(...)` instead of reusing
the holder's `closeThis`.

Minor secondary note: closing by default resurrects the old zombie-classloader trade-off —
tests/`run` that load classes after the task ends will now produce the "loading X after test
or run has completed" warning, and a zombie loader created after the close reopens the jars
(re-pinning them until GC). Probably acceptable, just worth expecting the reports.

eed3si9n:
Removed hardcoding in 7e020eec54436520cfc51c9ced450ef5103e0f22.

kevin-lee:
Since I previously asked Claude to test it in GitHub Actions, the analysis I got from Claude on [this failure](https://github.com/sbt/sbt/actions/runs/30696962038/job/91361364808?pr=9538) may also be useful.
***

Looked into the CI failure (`test (ubuntu-latest, 17, zulu, 2)`): the reported
`classloader-cache/java-serialization` failure ("Reload for scripted batch execution failed")
is collateral damage — the real casualty is **`classloader-cache/close-run`**, which runs
immediately before it in the same scripted batch group, sharing one nested sbt server.

## Mechanism

1. `close-run` is the regression test for late class-loading: `Main.main` starts a thread and
   returns; the thread sleeps 500 ms, then does `new Foo`, and on failure calls `sys.exit(1)`.
2. With close-by-default, the in-process `run` loader is now really closed the moment `main`
   returns. `close-run` itself still reports `+` (its script step finished fine).
3. The batch runner immediately deletes the temp dir contents and copies the next test's
   files in, then sends `reload;initialize` to the shared server.
4. 500 ms later the thread fires. The original loader is closed, so `ManagedClassLoader`
   falls back to a fresh `ZombieClassLoader` over the same URLs — but those files were just
   deleted in step 3, so `Foo` cannot be found. (This is exactly why the old `close=false`
   default survived: the still-open `JarFile` handle keeps serving classes even after the
   file is unlinked.)
5. `sys.exit(1)` runs in-process (no TrapExit on modern JDKs), killing the shared nested sbt
   JVM → the next test's `reload;initialize` hits a dead server and the ✗ lands on
   `java-serialization` with no underlying error output.

Timeline from the log matches exactly: `+ close-run` at `11:13:13.467` (0.36 s, vs 1.5–3 s
for every neighbor), thread fires ~500 ms later, reload fails at `11:13:14.39`. This is
deterministic whenever `close-run` precedes another test in a batch — not a flake. It is
*not* caused by the `closeThis = close` turbo commit (that part looks right), and the
`ubuntu-22.04` job's `##[error]` is just the "Stream closed" output-pump noise in server
tests.

## Suggestion

Fixing #9537 doesn't require closing `run` loaders at all — the pinned jar is opened by the
**test** classloader. Closing `run` loaders by default breaks the pattern `close-run` exists
to protect (main returns while spawned threads keep lazily loading classes — any
quick-returning server-style app). Scoping the close-by-default to test loaders only (keep
the `run` path on the old default) should keep `close-run` green and still cure the Windows
`AccessDeniedException`.

eed3si9n:
> the real casualty is `classloader-cache/close-run`, which runs
immediately before it in the same scripted batch group, sharing one nested sbt server.

That makes sense. I do however disagree with its claim that fixing this bug doesn't require closing run loaders. If a classloader held on to the JAR file, run would likely run into similar issue when `packageBin` gets called later. Given that sbtn defaults forked run in sbt 2.x, I'm ok with removing `classloader-cache/close-run` altogether.

PR #9541, opened 2026-08-02
[2.x] fix: Disable JVM jar-connection caching for all sessions, not just scripted

BrianHotopp:
## Problem

With `exportJars := true` (the 2.x default) test resources are served from the tests jar. `ClassLoader.getResourceAsStream` on a jar classpath entry goes through a `jar:` URLConnection whose `JarFile` is cached JVM-wide by URL. Test classloaders are not closed by default, so a previous run's cached `JarFile` is never evicted.

In a warm session, a test re-run after a resource-only edit reads old resource bytes through the cached `JarFile`, while class files stay fresh (the JDK's `URLClassPath` opens its own `JarFile` per loader). Digest invalidation itself works — the suite re-runs against stale bytes, wrongly passes, and that wrong pass is recorded under the new digest.

## Fix

sbt has disabled JVM jar-connection caching under scripted for years (`XMainConfiguration.disableCachingOfURLConnections()`). This change applies it to every session.

Validated end to end against the real launcher on stock 2.0.3 and a develop build. Class loading is unaffected; scripted's years of running with the cache disabled are the soak evidence.

## Testing

- `cache/test-resource-invalidation` (scripted): resource-only edit fails the test, restore recovers. The harness masks this bug (`-Dsbt.scripted=true`), so the test pins the user-visible contract under the fix.
- `cache/*` group (11 tests) green
- `mainProj` compile, `javafmtCheck`, `scalafmtSbtCheck` green

---

*Diagnosed with AI assistance (Claude Code).*


eed3si9n:
> In a warm session, a test re-run after a resource-only edit reads old resource bytes through the cached `JarFile`

This was the chief complaint of https://github.com/sbt/sbt/pull/9469, which has been fixed in https://github.com/sbt/sbt/pull/9469.

I'm not necessarily against the proposed change, but is there an open problem?

> Test classloaders are not closed by default, so a previous run's cached `JarFile` is never evicted.

https://github.com/sbt/sbt/pull/9538 proposes to flip the default value to close the test classloader.

BrianHotopp:
Thanks. I missed #9469. It fixes the stale-resource issue by resolving CAS symlinks before constructing classpath URLs, which removes the cache-key collision behind #9541.

#9538 addresses the separate Windows jar-handle issue by closing test classloaders by default.

I’ll close #9541 as superseded. Its global `jar:` cache change is unnecessary and too broad.

Issue #8569, opened 2026-01-16
Build pipelining hangs


matthughes:
I sadly don't have a reproduction but I've experienced several times today SBT just completely hang.  Doesn't matter if it's doing compilation or scalafmt or scalajs tasks.  This has happened both on my machine (Apple M1) and our Concourse server (Linux).

Not sure how useful this stack trace is but I attached it.  Looks exactly the same across both machine types.

[sbt_stack.txt](https://github.com/user-attachments/files/24682897/sbt_stack.txt)

eed3si9n:
If you can't reproduce it, I guess one thing you could do is to trace your steps on the temporal axis. Did it happen today, yesterday, etc. If you go back to older commits, does it still happen? If it's today/yesterday, were there any changes to the build, build infra, or code that could have affected this?

matthughes:
I've been running with `sbt -Dsbt.execute.extrachecks=true` and debug mode.  It's hanging for me a bunch now but I don't see anything interesting in the stack or in the debug output.

Here's the output before it just sits there:

```
[debug] IncrementalCompile.incrementalCompile
[debug] previous = Stamps for: 0 products, 0 sources, 0 libraries
[debug] current source = Set()
[debug] > initialChanges = InitialChanges(Changes(added = Set(), removed = Set(), changed = Set(), unmodified = ...),Set(),Set(),API Changes: Set())
[debug] Full compilation, no sources in previous analysis.
[debug] > initialChanges = InitialChanges(Changes(added = Set(), removed = Set(), changed = Set(), unmodified = ...),Set(),Set(),API Changes: Set())
[debug] No changes
[debug] Created transactional ClassFileManager with tempDir = /Users/MHughe008/work/redacted/server-common/target/scala-3.7.3/test-classes.bak
[debug] Created transactional ClassFileManager with tempDir = /Users/MHughe008/work/redacted/server/target/scala-3.7.3/classes.bak
[debug] [servercommon / test] early output is success
[debug] Removing the temporary directory used for backing up class files: /Users/MHughe008/work/redacted/server-common/target/scala-3.7.3/test-classes.bak
[debug] [server / compile] early output is success
[debug] Removing the temporary directory used for backing up class files: /Users/MHughe008/work/redacted/server/target/scala-3.7.3/classes.bak
[debug] serverCommon: compileEarly: blocking on earlyOutputPing
[debug] server: compileEarly: blocking on earlyOutputPing
[...]

PR #9542, opened 2026-08-03
[2.x] fix: Fixes pipelined build deadlocks

### LINKED ISSUE #9486: [2.x] usePipelining hangs
## steps

set `ThisBuild / usePipelining := true` in multi-module sbt2 build
run `compile;compile`

## problem

sbt hangs

## notes

demo here https://github.com/OlegYch/scala3-25520/actions/runs/29828900212/job/88628588903

BrianHotopp:
## Problem

Fixes #9486. Refs #8569.

With `usePipelining` enabled, `earlyOutputPing` was completed only when Zinc reported progress during compilation. On an action-cache hit, Zinc does not run; after a failed compile, it does not report progress. The promise remained incomplete, so downstream tasks such as `compileEarly` and `makePickleProducts` could wait forever.

## Fix

Complete the ping on every compile path:

- Zinc's completion remains authoritative when a real compile runs.
- A cache hit completes the ping from the pickle jar's presence. If the jar is absent, downstream compilation falls back to a full compile.
- A failed compile completes the ping with `false`, allowing waiters to take the full-compile path and propagate the failure.

`PromiseWrap.tryComplete` makes the Zinc and task-side completions race-safe.

## Testing

- Re-enabled `source-dependencies/pipelining` (`disabled` → `test`). Its aggregate `compile` after `dep/compile` resolves dep from the action cache — exactly the ping-starvation path. Without this change the fixture hangs (counterfactual-verified, killed at 420 s); with this change it passes. The fixture needed two repairs for 2.x: `checkPickle` opts out of action caching (`Def.uncached`), and its assertion matches the 2.x early-jar naming instead of the legacy `early.jar` filename.
- `mainProj/compile` and `mainSettingsProj/compile`
- `mainProj/mimaReportBinaryIssues` and `mainSettingsProj/mimaReportBinaryIssues`
- `mainProj/scalafmtCheckAll`, `mainSettingsProj/scalafmtCheckAll`, and `scalafmtSbtCheck`

## Note on `source-dependencies/pipelining-java`

Re-enabling this fixture surfaces a separate, pre-existing problem: after `dep/A.java` is deleted, `use/compile` resolves from the action cache and wrongly succeeds instead of recompiling and failing. That is an invalidation issue, unrelated to this fix (this change only completes a promise on paths that previously never completed it). I left it disabled; happy to file an issue.


eed3si9n:
@BrianHotopp Thanks for working on this!

> The existing `source-dependencies/pipelining` fixture is disabled in the repository and has no test script; it was not counted as passing.

You can rename `disabled` -> `test` to re-enable scripted tests. Could you try re-enabling `source-dependencies/pipelining-java` and `source-dependencies/pipelining`, which might suffice rather than adding a new scripted test scenario.

BrianHotopp:
Done. The re-enabled `source-dependencies/pipelining` covers the regression: its aggregate `compile` resolves `dep` from the action cache — the ping-starvation path. It hangs without the fix (killed at 420 s) and passes with it, so I dropped the new fixture. Two repairs for 2.x: `checkPickle` uses `Def.uncached`, and the assertion matches the new early-jar naming.

`pipelining-java` fails for a different reason: after `dep/A.java` is deleted, `use/compile` resolves from the action cache and wrongly succeeds. That is an invalidation bug, separate from this change; I left it disabled and filed #9546.

eed3si9n (approved):
Thanks!

Issue #9484, opened 2026-07-20
sbtn fails on GitHub Actions


eed3si9n:
## steps

See https://github.com/sbt/sbt/actions/runs/29760504315/job/88425562053#step:12:356

## problem

```bash
[sbt_options] declare -a sbt_options=()
[process_args] java_version = '21'
[debug] running native client
# Executing command line:
/home/runner/.cache/sbt/boot/sbtn/2.0.0-b4d628dd/sbtn
--sbt-script=./sbt
-v
scalafmtCheckAll


Error: Process completed with exit code 1.
```

This works locally on macOS as far as I can tell.

## expectation

Don't fail or print the reasons for failure.

merlinorg:
We experienced this frequently, so switched all our CI jobs to `sbt --server` to side-step.

PR #9549, opened 2026-08-04
[2.x] fix: Bound the client boot wait when the forked server never starts

BrianHotopp:
## Problem

Ref #9484.

The thin client's `blockUntilStart` loop recursed while the portfile was missing and the forked process looked valid, with no deadline. A forked server that died before writing `project/target/active.json`, or stayed alive but wedged, left the client waiting forever. On Windows, the liveness check was disabled to work around process-tree behavior, so the deadline is the only bound.

## Fix

Bound the connect wait to five minutes by default. Configure it with the `SBT_CLIENT_CONNECT_TIMEOUT` environment variable (seconds).

When the deadline expires, the client reports `sbt server did not start within N seconds` and includes the captured server stderr instead of hanging silently. The existing Windows liveness workaround remains in place.

## Testing

- `serverTestProj/Test/testOnly testpkg.ClientBootTimeoutTest`: forks a client JVM with `SBT_CLIENT_CONNECT_TIMEOUT=3` against a fake server that sleeps without writing a portfile; the client fails within the test's one-minute join with the timeout message and captured stderr. Without the deadline condition, the test hangs past the join.
- `commandProj/compile`
- `commandProj/mimaReportBinaryIssues`
- `serverTestProj/scalafmtCheckAll`
- `scalafmtSbtCheck`

---

*Diagnosed and fixed with AI assistance (Oh My Pi).*

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -343,6 +343,14 @@ class NetworkClient(
         Try(ClientSocket.localSocket(namedPipeName, useJNI)).toOption
       case _ => None
 
+  private def connectTimeout: FiniteDuration =
+    sys.env
+      .get("SBT_CLIENT_CONNECT_TIMEOUT")
+      .flatMap(_.toIntOption)
+      .map(_.seconds)
+      .getOrElse(5.minutes)
+  private var connectDeadlineExpired = false
+
   /**
    * Forks another instance of sbt in the background.
    * This instance must be shutdown explicitly via `sbt -client shutdown`
@@ -482,6 +490,7 @@ class NetworkClient(
         } catch { case e: IOException => e.printStackTrace(System.err) }
       }
     }
+    val connectDeadline = connectTimeout.fromNow
     @tailrec
     def blockUntilStart(): Unit = {
       val stop =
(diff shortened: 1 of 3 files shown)


main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -347,6 +347,14 @@ class NetworkClient(
    * Forks another instance of sbt in the background.
    * This instance must be shutdown explicitly via `sbt -client shutdown`
    */
+  private def bootTimeout: FiniteDuration =
+    sys.props
+      .get("sbt.client.boot.timeout.seconds")
+      .flatMap(_.toIntOption)
+      .map(_.seconds)
+      .getOrElse(5.minutes)
```
eed3si9n: Thanks for the contribution.

1. Could you move this above the comment block?
2. Given that sbtn runs as native code, I don't think we can use system property. Maybe we can use environment variable like `SBT_CLIENT_CONNECT_TIMEOUT`

main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
```diff
@@ -347,6 +347,14 @@ class NetworkClient(
    * Forks another instance of sbt in the background.
    * This instance must be shutdown explicitly via `sbt -client shutdown`
    */
+  private def bootTimeout: FiniteDuration =
```
eed3si9n: For the naming, I suggest `connectTimeout` like an HTTP client since booting is happening on the server-side, and we don't really know what could be happening from the client perspective.

eed3si9n (approved):
Thanks!

eed3si9n:
I think the change itself looks good, but it's not clear if this alone would fix https://github.com/sbt/sbt/issues/9484 given that in some examples the sbt server seems to be up, but sbtn is failing to reconnect to one.

PR #9550, opened 2026-08-04
[2.x] perf: Keep the analysis cached across an action cache output sync

hoangmaihuy:

Ref https://github.com/sbt/sbt/issues/9560

## Problem

The local analysis cache validates an entry against the analysis file's timestamp and size:

```scala
case Some((v, mod, i)) if lastModified == mod && sizeBytes == i => v
```

A file re-created with identical content therefore loses its entry — which is what happens to every analysis sbt writes, because the analysis file is a declared output of `compileIncremental`. Once the cached task's body returns, `ActionCache.cache` → `syncBlobs` → `syncFile` deletes the file the task just wrote and re-creates it as a symlink into the CAS:

```
SET-after  file    mtime=1785813322557 size=929  lib <- cachedCompileIncremental
MISS       symlink mtime=1785813322561 size=929  lib <- compileIncremental
```

Same size, four milliseconds later, new timestamp. So the read that immediately follows the write deserializes the analysis that was just written — one deserialization per compiled module per compile, the cost the cache exists to avoid. On this repo's largest analysis (519K) that read costs 19.5ms, against 0.097ms to hash the file and 0.0025ms to stat it; deserializing runs ~38ms/MB. All 138 analysis files in this repo's own build are CAS symlinks.

The cache had no test coverage.

## Change

An entry also carries the hash of the content it was read from, and a re-created file is recognised by it:

```scala
Option(inMemoryAnalysisCache.getIfPresent(ref.id())) match
  case Some(cached) if cached.lastModified == lastModified && cached.sizeBytes == sizeBytes =>
    cached.contents
  // an action cache output is re-created in place once it has been written, which brings the
  // same analysis back under a new timestamp
  case Some(cached) if cached.sizeBytes == sizeBytes && cached.contentHash == contentHash =>
    record(cached.contents, cached.contentHash)
  case _ =>
    record(value, contentHash)
```

The hash is a `lazy val`: a read whose timestamp and size still match never computes it, so a warm no-op compile still costs one `stat` per read. The size check in front of the hash comparison keeps the guard from `566f88804` — a file switched to a *different* analysis is still read again.

`set` populates the cache, as `566f88804` also did, but it read the file's attributes *before* `underlying.set`, so the entry landed under the pre-write timestamp and never matched again. The attributes now come from the written file, and its hash is taken eagerly because the action cache replaces that file moments later. None of this bookkeeping may fail a compile whose analysis is already on disk, so an `IOException` while stat-ing or hashing it is dropped.

What `set` records is what a read returns, not what it was handed. `ConsistentAnalysisFormat` does not persist `Compilations` and `compileScalaBackendTask` derives `hasModified` from them, so caching the in-memory analysis verbatim flips `hasModified` to `true` and makes `compileTask` store the analysis on every compile, no-ops included — and only for as long as the entry stays resident, so the same build would behave differently after an eviction. `asPersisted` empties `Compilations`, the one field a read loses, which leaves `hasModified` exactly as it is on `develop`.

## Result

Across five compiles of two modules with two edits: five deserializations become none, the 32 reads answered by a `stat` alone stay that way, and five hashes are added — the reads that immediately follow the five writes. `previousCompile`, `compileIncremental`, `compileScalaBackend` and the dependency analysis read in `compileIncSetup` are all served from memory.

## Tests

`LocalAnalysisCacheTest` (hedgehog, 4 examples) pins the store. On `develop`, `a stored analysis is served from the local cache` and `a re-created analysis file stays cached` fail (`Failed: Total 4, Failed 2, Errors 0, Passed 2`); the other two are guards, for the switched-file case and for `asPersisted`, and neither passes vacuously — caching `contents` verbatim turns the latter red.

`cache/analysis-cache-reads` pins the compile. Nothing a build can observe distinguishes a served analysis from a re-read one, so the cache counts what its reads cost — answered from memory, hashed the file to answer, deserialized it — and the test reads those counts around a compile. The counts are `private[sbt]`, reached from a helper declared in `package sbt` under the test's `project/`, so no public API is added.

It asserts two invariants. A no-op compile must answer every read from memory without hashing: dropping the timestamp arm gives `7 hashed the file`. A recompile must be served the analysis it just wrote: `develop`'s cache logic with only the counters added gives `1 of 7 reads deserialized it`.

Both need the run's compiles to be real ones, and the CAS is global rather than per-test, so the build points `localCacheDirectory` inside the sandbox — in `Global`, the scope `cacheStores` resolves it in — and the test deletes `target` first.


eed3si9n:
> Once the cached task's body returns, `ActionCache.cache` → `syncBlobs` → `syncFile` deletes the file the task just wrote and re-creates it as a symlink into the CAS

Interesting. I'm not against the direction of this PR, but I wonder if we can solve this more generally by keeping the original timestamp when we copy to diskcache CAS -- https://github.com/sbt/sbt/pull/9559

hoangmaihuy:
> Interesting. I'm not against the direction of this PR, but I wonder if we can solve this more generally by keeping the original timestamp when we copy to diskcache CAS -- #9559

Fixing the root cause is definitely the better choice here, but I think analysis memory cache plays a critical part in warm compilation and using mtime as cache key is sensitive to changes, nothing can prevent other tasks from touching on disk analysis. 

So I think we should either (1) Keep analysis content hash as a fallback, we pay only one content hash per cache write for safety after disk cache CAS issue is fixed or (2) Keep using (mtime, size) as old way but adding tests to verify cache hit to avoid performance regression.

eed3si9n (approved):
Thanks!

PR #9553, opened 2026-08-04
[2.x] fix: sbtn to log connection errors

BrianHotopp:
## Problem

Refs #9484.

When the thin client loses its connection to the server mid-run, or the connect phase throws, it exits 1 with no output. The CI log attached to the issue shows exactly this: two one-second sbtn commands succeed against a live server, then the next exits 1 having printed nothing.

Two silent paths:

- A server-initiated drop (no `Shutdown` notification) goes through `onClose`, which set `exitClean` and interrupted the waiting thread without logging anything.
- `client()`/`clientImpl()` catch any `Exception` and return 1 without printing it.

## Fix

- `onClose` now logs `sbt server disconnected` when the close wasn't initiated by the client. `NetworkClient.close()` clears `running` before closing the session, so a still-set `running` flag at `onClose` entry identifies a server-side drop. The existing `Shutdown`-notification path clears `running` first, so it doesn't double-log.
- The `client()`/`clientImpl()` catch-alls print the exception and stack trace via a new `private[sbt] logFailure` helper.

The failing CI run used sbtn `2.0.0-b4d628dd`, which predates #9521, #9526, and #9527; the current pin (`2.0.0-731e6666`) includes them. Whatever the server-side root cause was, any recurrence now explains itself instead of failing silently.

## Testing

- New `NetworkClientSilentDisconnectTest`:
  - A fake server accepts then drops the connection mid-batch: the client exits 1 and prints `sbt server disconnected`.
  - A corrupt portfile: the client exits 1 and prints the exception.
  - Both tests fail without the fix (verified by reverting it).
- `commandProj` client tests: 41 passed.
- `serverTestProj` `ClientTest`, `ClientSubscriptionTest`, `ClientBootTimeoutTest`: 13 passed.
- `commandProj/scalafmtCheckAll`, `commandProj/mimaReportBinaryIssues`, `scalafmtSbtCheck`.

---

*Diagnosed and fixed with AI assistance (Oh My Pi).*

### DIFF
--- a/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
+++ b/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala
@@ -156,6 +156,11 @@ class NetworkClient(
 
   private def mkSocket(file: File): (Socket, Option[String]) = ClientSocket.socket(file, useJNI)
 
+  private[sbt] def logFailure(e: Exception): Unit = {
+    errorStream.println(s"sbt client failed: $e")
+    e.printStackTrace(errorStream)
+  }
+
   private def portfile = arguments.baseDirectory / "project" / "target" / "active.json"
 
   def connection: ServerSession = connectionHolder.synchronized {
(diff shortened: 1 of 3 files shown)


eed3si9n (approved):
Thanks!

PR #9555, opened 2026-08-05
[2.x] fix: Don't re-extract a dirzip whose digest already matches

### LINKED ISSUE #9554: [2.x] `DiskActionCacheStore.syncFile` re-extracts every freshly packaged dirzip
**Disclaimer**: Investigated with Claude Code and human review in the loop

## steps

1. Any sbt 2.x build with a non-trivial `classes` directory, so the cost is visible.
2. `compile`, edit one source file, `compile` again.
3. Look at `<classes>.sbtdir.zip` next to the classes directory.

## problem

Every compile inflates the archive it has just written into a temporary directory, digest-compares every file in the output tree against the manifest, finds nothing changed, and deletes the temporary directory again. None of that work can change anything, and it costs time proportional to the size of the output directory, every time a task declaring an output directory actually runs its body. A cache hit skips the body, so this is the recompile case rather than literally every invocation.

On a 2557 file, 33 MB output directory with an 11 MB archive it measures ~450 ms:

| step | time |
|---|---|
| `IO.unzip` of the 11 MB archive into 33 MB | 329-381 ms |
| re-list the output directory | 16-19 ms |
| digest-compare 2557 files | 74-105 ms |

For scale, that directory's whole warm single-file edit is ~4.2 s, and `packageDirectory` accounts for ~1.4 s of it. This half is pure waste; the other half is [serial deflation in `IO.zip`](https://github.com/sbt/io), filed separately.

## cause

`DiskActionCacheStore.syncFile` treats an output as up to date only when its digest matches *and* the local path is already a symlink:

```scala
// util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
if Digest.sameDigest(p, d) && (!symlinkSupported.get() || Files.isSymbolicLink(p)) then
  afterFileUpToDate(ref, p, outputDirectory)
  p
else
  IO.delete(p.toFile())
  writeFileAndNotify(p)
```

`ActionCache.installPackagedZip` publishes the archive with `Files.move`, so on the write path `zipPath` is always a *regular* file. With symlinks supported — the default, `symlinkSupported` is initialised to `true` — the second conjunct can never hold there, so the branch always falls through to `IO.delete` + `writeFileAndNotify`, which calls `afterFileWrite`, which for a `.sbtdir.zip` calls `unpackageDirZip`.

The symlink you find at `<classes>.sbtdir.zip` after a build is the evidence this ran: only `writeFileAndNotify` creates it, and `packageDirectory` cannot have left it that way.

## expected

A matching digest already means the local content *is* the blob's, whether or not the path happens to be a symlink, so that case should take the up-to-date path.

## environment

- sbt `2.x` / `develop`
- macOS 15 arm64, 12 cores, APFS
- OpenJDK 26.0.2

Timings are from one machine; the shape of the cost, being one full inflate plus one full hash of the output tree per compile, is what matters rather than the absolute numbers.

hoangmaihuy:

Fixes https://github.com/sbt/sbt/issues/9554

## Changes

Test the digest first. A matching digest already means the local content *is* the blob's, whether or not the path happens to be a symlink. Where the path is not yet a symlink and symlinks are supported, still relink it to the CAS so deduplication is kept, but reach `afterFileUpToDate` — one `isDirectory` stat — rather than `afterFileWrite`.

This makes the regular-file case behave like the symlink case, which already skipped re-extraction. The symlink-creation fallback, including the unsupported-symlink detection that clears `symlinkSupported`, is factored into `linkOrCopy` and shared with `writeFileAndNotify` rather than duplicated.

## Notes for review

**The behaviour given up.** A file inside the extracted directory that has diverged from the manifest is no longer silently restored while the archive digest still matches. On the write path the directory is the source the archive was just built from, so it is authoritative; on a cache hit with a differing digest the full unpackage runs as before; and a directory that is missing entirely is still recreated, which `afterFileUpToDate` already handles.

**Scope.** For outputs that are not dirzips, `afterFileWrite` and `afterFileUpToDate` are both no-ops, and that branch already did a delete-and-relink, so their behaviour is unchanged. The change is confined to dirzips.

## Verification

- `testFull` — no failures.
- `scripted cache/*` — 12/12, including `compile-classes-restore`, `declare-output-dir-restore` and `declare-output-loop`, which exercise exactly the restore paths this touches.
- `scripted remote-cache/basic` — `hitCount=9 missCount=0 remoteHitCount=9`.
- Three new tests in `ActionCacheTest`: syncing a freshly packaged archive leaves a diverged file in the extracted tree alone; a digest-matching regular file is still relinked to the CAS; and a differing digest still fully re-extracts, restoring a stale file and deleting a stray one. The first fails before this change, returning the archived content instead of the diverged content.

### DIFF
--- a/util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
+++ b/util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
@@ -312,9 +312,8 @@ case class DiskActionCacheStore(base: Path, converter: FileConverter)
     // See https://github.com/sbt/sbt/issues/7656
     // On Windows, the program has be running under the Administrator privileges or the
     // user enable Developer Mode on Windows 10+ to create symbolic links.
-    def writeFileAndNotify(outPath: Path): Path =
-      Option(outPath.getParent()).foreach(parent => IO.createDirectory(parent.toFile()))
-      val result = Retry:
+    def linkOrCopy(outPath: Path): Path =
+      Retry:
         if Files.exists(outPath) then IO.delete(outPath.toFile())
         if symlinkSupported.get() && Files.exists(casFile) then
           try Files.createSymbolicLink(outPath, casFile)
@@ -333,6 +332,9 @@ case class DiskActionCacheStore(base: Path, converter: FileConverter)
                 symlinkSupported.set(false)
               copyFile(outPath)
         else copyFile(outPath)
+    def writeFileAndNotify(outPath: Path): Path =
+      Option(outPath.getParent()).foreach(parent => IO.createDirectory(parent.toFile()))
+      val result = linkOrCopy(outPath)
       afterFileWrite(ref, result, outputDirectory)
       result
     val resolvedPath = converter.toPath(ref) match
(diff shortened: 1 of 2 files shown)


eed3si9n:
Thanks for the contribution!

util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
```diff
@@ -344,11 +346,15 @@ case class DiskActionCacheStore(base: Path, converter: FileConverter)
...
         try
-          // `!symlinkSupported` prevents unnecessary deletion of files and then copying them again
-          // in #writeFileAndNotify on machines that don't support symlinks.
-          if Digest.sameDigest(p, d) && (!symlinkSupported.get() || Files.isSymbolicLink(p)) then
-            afterFileUpToDate(ref, p, outputDirectory)
-            p
+          // A matching digest means the content is already the blob's, so a dirzip's extracted
+          // directory is in sync and must not be unpackaged again. Relink to the CAS to keep
+          // deduplication, but only where that isn't a needless delete-and-copy: on machines
+          // without symlink support the copy already there is what we would write back.
```
eed3si9n: nit: Since this file changes relatively frequently, I suggest removing any inline comments if the test covers the scenario.

```suggestion
```

eed3si9n (approved):

PR #9559, opened 2026-08-06
[2.x] fix: Preserve file timestamp in local diskcache

eed3si9n:
Ref https://github.com/sbt/sbt/issues/9560

## Problem
In theory, we shouldn't have to care about the timestamps, but for local optimization it's still a useful key for caching.
Currently local Analysis cache gets invalidated after syncBlob replaces the Analysis file with a symlink.

## Solution
This preserves the original timestamp when copying files to the local diskcache.


eed3si9n:
cc @hoangmaihuy

PR #9563, opened 2026-08-06
[2.x] perf: Improve update task by caching file hashes

takayahilton:

Part of #9514. Applies the approach of #8363 to the code path it could not reach; complements #9524, which removed the hash from persisted reports only.

## Problem

sjsonnew serializes a `File` as a `(uri, Long)` pair whose `Long` is derived from a **full sha256 of the file's content**, recomputed on every serialization:

```scala
// sjsonnew.FileIsoStringLongs
implicit lazy val fileStringLongIso: IsoStringLong[File] = IsoStringLong.iso[File](
  (file: File) => (IsoStringLong.fileToString(file), HashUtil.sha256ToLong(file.toPath())),
  ...)
```

#9524 stopped paying this for the persisted update caches, but two paths still serialize `UpdateReport`s through the stock `LibraryManagementCodec`:

- **Action cache key computation.** Any task or plugin whose input includes an `UpdateReport` hashes it through `HashWriter` on every invocation — including every no-op. sbt-protoc is one such plugin (`import sbt.librarymanagement.LibraryManagementCodec.{*, given}`).
- **`Tracked.lastOutput` write-back**, which re-serializes the report even on a cache hit (#9514 §6).

A report names each artifact once per configuration it resolved in, and the projects of a build largely share their dependencies — on the profiled monorepo the reports hold **381,268 `ModuleReport` entries for 757 distinct modules (504×)** — so a no-op build re-reads the same jars hundreds of times. The cost scales with *projects × configurations × classpath bytes*, which is why large multi-project builds are hit hardest while small builds never notice.

sbt already solved exactly this in #8363 for its own digest layer (`sbt.util.CacheImplicits`): an in-memory cache validated by (size, lastModified). The sjson-new iso is an independent implementation of the same hashing and was left out.

## Change

`MemoizedFileFormats`, a trait that overrides `fileStringLongIso` / `pathStringLongIso` with the same conversion, but with the sha256 memoized in a Caffeine cache keyed by the path string and validated by (size, lastModified) — the invalidation semantics of the #8363 digest cache. A changed file changes its size or timestamp and is re-hashed; entries are overwritten in place, so stale keys cannot accumulate.

It is mixed into the generated codec via `contrabandCodecParents`:

```scala
Compile / generateContrabands / contrabandCodecParents += "sbt.librarymanagement.MemoizedFileFormats",
```

`fileStringLongIso` is an `implicit lazy val` in `sjsonnew.FileIsoStringLongs` — a **virtual member** — so, as in #9524, overriding it reaches every generated format that resolves `JsonFormat[File]` through its self-type, including the `Vector[(Artifact, File)]` nested in `ModuleReportFormat`, and `IvyLibraryManagementCodec` in lm-ivy transitively. `UpdateReportPersistence.CacheCodec`'s own hash-disabling override still wins for the persistence path.

The cache is in-memory and unbounded in time but bounded (65536 entries; #8363 likewise bounds its digest cache, by bytes). In-memory is enough: the 504× duplication above means the payoff is realized *within a single command* — each measurement below is a single fresh JVM.

## Compatibility

Unlike #9524 this changes **neither the JSON shape nor the values**: the hash written is the same sha256, only its recomputation is avoided. Existing caches and cache keys are unaffected. `mimaReportBinaryIssues` passes.

One note for external code: the generated `*Formats` traits gain `sbt.librarymanagement.MemoizedFileFormats` in their self-types. Binary compatibility is unaffected; source compatibility matters only for code that mixes individual formats traits directly onto `BasicJsonProtocol` instead of extending `LibraryManagementCodec` (the Contraband 0.8.0 `given` migration was a far larger source change to the same traits).

Two behavioral differences, both confined to degenerate inputs: a file that exists but cannot be read hashes to `0L` where stock sjson-new throws `BufferUnderflowException`, and `0L` results (directories, vanished files) are never cached.

## Validation

New tests — `lm-core/.../MemoizedFileFormatsSpec.scala`:

| test | covers |
|---|---|
| `fileStringLongIso` / `pathStringLongIso` compute the same hash as the stock sha256 | the memoized value is the stock value |
| hash is served from the cache while (path, size, mtime) is unchanged | rewrites same-length content with restored mtime; a recomputation would change the hash |
| hash is recomputed when the mtime / the size changes | invalidation, one test per component |
| a missing file / a directory hashes to 0 | the stock sentinels |
| a `File` round-trips through the iso | the URI half still inverts |
| the memoized iso is wired into the generated `LibraryManagementCodec` | pins the *wiring*, not the codec: renders an `UpdateReport` through the generated `UpdateReportFormat` twice around a same-size rewrite with a restored mtime — a version that quietly resolved through the stock iso would still compile and pass the direct-codec tests, but would re-hash and fail this one |

| suite | result |
|---|---|
| `lmCore/Test/testFull` | 618 + 9 passed |
| `mainProj/Test/testFull` | 226 passed, 1 ignored |
| `lmCore/mimaReportBinaryIssues` | clean |

## Measurements

242-project monorepo (162 with src/it), Apple M1 Pro, Temurin 25, protocol of #9514 (one fresh JVM per run, `-Dsbt.task.timings=true`, server shut down between runs). The 2.0.4 → 2.0.5 baseline delta independently reproduces the effect of #9522/#9524.

| | 2.0.5 | with this change |
|---|---|---|
| no-op `Test/compile` | 91 / 87 / 89 s | **18 / 21 / 23 s (−77%)** |
| `managedClasspath` task total (~970 instances) | 717 s | **26 / 26 / 30 s (−96%)** |


lm-core/src/main/scala/sbt/librarymanagement/MemoizedFileFormats.scala
```diff
@@ -0,0 +1,76 @@
...
+          // content hash; caching it would keep serving 0 for the key
+          if (h != 0L) hashCache.put(key, new Entry(attrs.size, mtime, h))
+          h
+        }
+    }
+
+  override implicit lazy val fileStringLongIso: IsoStringLong[File] = IsoStringLong.iso[File](
+    file => (IO.toURI(file).toASCIIString, memoizedHash(file.toPath())),
+    p => IO.toFile(new URI(p._1))
+  )
```
eed3si9n, line 61: @takayahilton Thanks for the contribution!

eed3si9n (approved):

lm-core/src/main/scala/sbt/librarymanagement/MemoizedFileFormats.scala
```diff
@@ -0,0 +1,76 @@
...
+ * Mix in after `BasicJsonProtocol` (via `contrabandCodecParents`) so these overrides win in
+ * linearization.
+ */
+trait MemoizedFileFormats extends FileIsoStringLongs {
+  import MemoizedFileFormats.{ Entry, hashCache }
+
+  // sjsonnew.HashUtil.sha256ToLong is private[sjsonnew]; same semantics via the public sha256
+  private def hashFile(path: Path): Long = {
+    val buf = HashUtil.sha256(path.toFile())
+    if (buf.length < 8) 0L else ByteBuffer.wrap(buf).getLong()
```
eed3si9n: Trying to trigger CI.

```suggestion
    if buf.length < 8 then 0L else ByteBuffer.wrap(buf).getLong()
```

lm-core/src/main/scala/sbt/librarymanagement/MemoizedFileFormats.scala
```diff
@@ -0,0 +1,76 @@
...
+  private def hashFile(path: Path): Long = {
+    val buf = HashUtil.sha256(path.toFile())
+    if buf.length < 8 then 0L else ByteBuffer.wrap(buf).getLong()
+  }
+
+  private def memoizedHash(path: Path): Long =
+    Try(Files.readAttributes(path, classOf[BasicFileAttributes])).toOption match {
+      case None =>
+        // no size/mtime to validate a cache entry against, so hash uncached
+        if (Files.isRegularFile(path)) hashFile(path) else 0L
```
eed3si9n:
```suggestion
        if Files.isRegularFile(path) then hashFile(path) else 0L
```

Issue #9317, opened 2026-06-09
[1.x] NoSuchMethodError: 'java.lang.String[] org.jline.terminal.impl.MouseSupport.keys(org.jline.terminal.Terminal)'


xuwei-k:
### project/build.properties 

```properties
sbt.version=1.12.11
```

### build.sbt 

```scala
scalaVersion := "3.9.0-RC1"

console / fork := true
```

## problem

`sbt console`

```
[error] java.lang.NoSuchMethodError: 'java.lang.String[] org.jline.terminal.impl.MouseSupport.keys(org.jline.terminal.Terminal)'
[error] 	at org.jline.reader.impl.LineReaderImpl.bindArrowKeys(LineReaderImpl.java:6895)
[error] 	at org.jline.reader.impl.LineReaderImpl.emacs(LineReaderImpl.java:6651)
[error] 	at org.jline.reader.impl.LineReaderImpl.defaultKeyMaps(LineReaderImpl.java:6550)
[error] 	at org.jline.reader.impl.LineReaderImpl.<init>(LineReaderImpl.java:350)
[error] 	at org.jline.reader.LineReaderBuilder.build(LineReaderBuilder.java:218)
[error] 	at dotty.tools.repl.JLineTerminal.<init>(JLineTerminal.scala:59)
[error] 	at dotty.tools.repl.JLineTerminal.<init>(JLineTerminal.scala:32)
[error] 	at dotty.tools.repl.ReplDriver.runUntilQuit(ReplDriver.scala:175)
[error] 	at xsbt.ConsoleInterface.run(ConsoleInterface.java:52)
[error] 	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
[error] 	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
[error] 	at sbt.internal.inc.AnalyzingCompiler.invoke(AnalyzingCompiler.scala:329)
[error] 	at sbt.internal.inc.AnalyzingCompiler.console(AnalyzingCompiler.scala:233)
[error] 	at sbt.Console.console0$1(Console.scala:65)
[error] 	at sbt.Console.$anonfun$apply$5(Console.scala:75)
[error] 	at sbt.Run$.executeSuccess(Run.scala:227)
[error] 	at sbt.Console.$anonfun$apply$4(Console.scala:75)
[error] 	at sbt.internal.util.Terminal.withRawInput(Terminal.scala:146)
[...]

PR #9564, opened 2026-08-06
[1.x] Fix Scala 3.9.x console crash by unifying the JLine version used

Gedochao:
Fixes https://github.com/sbt/sbt/issues/9317
Fixes https://github.com/scala/scala3/issues/26678

The simple regex change enforces all `org.jline` jars to be of one version - the crash was caused by intermixing jars from JLine 3 and JLine 4, as far as I can tell.

A potential problem here is that technically speaking, this runs Scala 3.9.x REPL using SBT's own JLine 3, because the relevant APIs match those from JLine 4. This could possibly stop being true in a future JLine bump on the compiler's side, but seems to work fine as far as I have tested it for Scala 3.9.

## AI usage
I've used Cursor + Claude to help with debugging this and pointing me to potential solutions. This seems like the simplest way out. Verified the fix manually with a locally published snapshot of SBT vs Scala 3.9.0-RC4 REPL + included a scripted automated test

### DIFF
--- a/main/src/main/java/sbt/internal/MetaBuildLoader.java
+++ b/main/src/main/java/sbt/internal/MetaBuildLoader.java
@@ -66,8 +66,11 @@ public final class MetaBuildLoader extends URLClassLoader {
    *     library.
    */
   public static MetaBuildLoader makeLoader(final AppProvider appProvider) throws IOException {
+    // Every org.jline artifact must live in this tier together. The REPL's ScalaInstance
+    // loader is parented to it, so a partial match leaves the REPL running one JLine
+    // version's reader against another's terminal (#9317, #8294).
     final String jlineJars =
-        "jline-?[0-9.]+-sbt-.*|jline-terminal(-(jni))?-[0-9.]+|jline-native-[0-9.]+";
+        "jline-?[0-9.]+-sbt-.*|jline-(terminal(-jni)?|native|reader|builtins|style)-[0-9.]+";
     final String testInterfaceJars = "test-interface(-.*)?";
     final String compilerInterfaceJars = "compiler-interface(-.*)?";
     final String utilInterfaceJars = "util-interface(-.*)?";
--- a/sbt-app/src/sbt-test/console/jline-version-conflict/build.sbt
+++ b/sbt-app/src/sbt-test/console/jline-version-conflict/build.sbt
@@ -1,0 +1,14 @@
+scalaVersion := "3.9.0-RC4"
+
+lazy val markerFile = settingKey[java.io.File]("marker file written by the console REPL once it starts")
+
+markerFile := target.value / "console-jline-ok"
+
+console / initialCommands := {
+  val path = markerFile.value.getAbsolutePath.replace("\\", "\\\\")
+  // Scala 3.9 ships JLine 4 while sbt 1.x ships JLine 3. A partial match in
+  // MetaBuildLoader's JLine tier used to leave the REPL running jline-reader 4.x
+  // against jline-terminal 3.x, which throws NoSuchMethodError before the prompt
+  // (#9317). The marker is only written if the REPL actually starts.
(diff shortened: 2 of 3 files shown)


Gedochao:
@eed3si9n can we run the CI on this?

WojciechMazur:
I think the https://github.com/scala/cla-checker is still using Akka endpoint for validation instead of  https://cla.scala-lang.org/scala/scala3 - let me check if that's the correct. <del>We'd definetly need to update the in action, but I'm not sure if we might also setup the dedicated endpoint for community projects, not the Scala 3 compiler </del> 

Edit: The cla-checker is deprecated, scala3 and scala-js have already migrated to new workflow with dedicated agent, this project should be migrated as well (https://github.com/scala/cla-checker/issues/6)

FYI @Gedochao @WarcholJakub 

For maintainers, see https://github.com/apps/scala-cla-bot/

Gedochao:
Also, I think the `test (ubuntu-latest, 8, zulu)` got cancelled randomly during GitHub downtime, needs a restart.

eed3si9n:
I've restarted them both.

Gedochao:
> A potential problem here is that technically speaking, this runs Scala 3.9.x REPL using SBT's own JLine 3, because the relevant APIs match those from JLine 4. This could possibly stop being true in a future JLine bump on the compiler's side, but seems to work fine as far as I have tested it for Scala 3.9.

- https://github.com/jline/jline3/issues/1566

As far as I can tell, it seems JLine 4 is not likely to break compatibility too much here... So as long as a JLine 5 won't come into the picture, we might get away with doing this.

eed3si9n (approved):
Thanks!

main/src/main/java/sbt/internal/MetaBuildLoader.java
```diff
@@ -66,8 +66,11 @@ public void close() throws IOException {
    *     library.
    */
   public static MetaBuildLoader makeLoader(final AppProvider appProvider) throws IOException {
+    // Every org.jline artifact must live in this tier together. The REPL's ScalaInstance
+    // loader is parented to it, so a partial match leaves the REPL running one JLine
+    // version's reader against another's terminal (#9317, #8294).
     final String jlineJars =
-        "jline-?[0-9.]+-sbt-.*|jline-terminal(-(jni))?-[0-9.]+|jline-native-[0-9.]+";
+        "jline-?[0-9.]+-sbt-.*|jline-(terminal(-jni)?|native|reader|builtins|style)-[0-9.]+";
```
eed3si9n, line 73: To clarify what's happening here, this allows all JLine 3.27 JARs coming from sbt 1.x's transitive dependencies to cohabit within the bottom-layer ClassLoader, so the JLine 3.x classes can find each other. And therefore, the only change is in the `MetaBuildLoader`?

Gedochao: Pretty much, yeah. Where the go is the dedicated `JLineLoader`, but it's `MetaBuildLoader` that decides what goes there.

SethTisue:
@eed3si9n are you considering doing a 1.12.16 fairly soon to get this fix out? I'm asking because although Scala 3.9.0 is delayed at the moment (Wojciech is working on 3.9.0-RC6), it won't be delayed indefinitely, and it would be good to have an sbt 1 out with this fix before there is a big wave of people trying out Scala 3.9

eed3si9n:
> are you considering doing a 1.12.16 fairly soon to get this fix out?

Sure, I can get something soonish.

lrytz:
One note: the regex here covers the jline jars that are used by sbt, which happens to cover all JLine modules used by the compiler(s). If Scala ever depends on a JLine module sbt doesn't have, that one would come from Scala. But that scenario is probably less likely to break, IIUC before we ended up mixing classfiles from two versions within the same module.

PR #9574, opened 2026-08-10
[2.x] Fixes pipelined build with Java sources

### LINKED ISSUE #9343: [2.x] Java source dependencies are not properly picked up when usePipelining = true
## steps

Check out https://github.com/lavifb/sbt2-pipelining-java-repro

When `usePipelining = true` the compile fails
```
sbt:sbt2-pipelining-java-repro-root> compile
[info] compiling 1 Scala source and 1 Java source to /Users/lblumberg/sbt2-pipelining-java-repro/target/out/jvm/scala-2.13.18/upstream/classes ...
[info] compiling 1 Scala source to /Users/lblumberg/sbt2-pipelining-java-repro/target/out/jvm/scala-2.13.18/downstream/classes ...
[error] /Users/lblumberg/sbt2-pipelining-java-repro/downstream/src/main/scala/example/Main.scala:4:23: not found: value Greeter
[error]   def run(): String = Greeter.greet("world")
[error]                       ^
[error] one error found
[error] (downstream / Compile / compileIncremental) Compilation failed
[error] elapsed time: 1 s, cache 47%, 18 disk cache hits, 20 onsite tasks
```

If you set `usePipelining = false`, the compile succeeds
```
sbt:sbt2-pipelining-java-repro-root> compile
[info] compiling 1 Scala source and 1 Java source to /Users/lblumberg/sbt2-pipelining-java-repro/target/out/jvm/scala-2.13.18/upstream/classes ...
[info] compiling 1 Scala source to /Users/lblumberg/sbt2-pipelining-java-repro/target/out/jvm/scala-2.13.18/downstream/classes ...
[success] elapsed time: 1 s, cache 66%, 26 disk cache hits, 13 onsite tasks
```

## expectation

usePipelining = true should work with java sources like it does in sbt 1.x

anatoliykmetyuk:
Fix #9343

With `usePipelining := true` set at an upstream project, Java and Scala sources compile separately. With same upstream `exportPipelining := false`, sbt created and exposed compilation artifacts downstream only after Scala compilation without waiting for the Java compilation to finish. Therefore downstream projects were missing Java artifacts.

This PR makes normal product generation wait for both Scala and Java compilation to finish. Because Java is a separate phase under `usePipelining`, this PR also adds caching this phase.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -988,6 +988,22 @@ object Defaults extends BuildCommon with DefExtra {
           val converter = fileConverter.value
           val cp = cp1.map(converter.toPath).map(converter.toVirtualFile)
           opts.withClasspath(cp.toArray)
+        },
+        compileInputs2 := Def.uncached {
+          compileScalaBackend.value
+          val (_, _, packedBackendOutput) = compileIncremental.value
+          val inputs = compileInputs.value
+          val c = fileConverter.value
+          CompileInputs2(
+            packedBackendOutput +: data(dependencyClasspath.value).toVector,
+            sourcesVF.value,
+            scalacOptions.value.toVector,
+            javacOptions.value.toVector,
+            c.toVirtualFile(inputs.options.classesDirectory),
+            c.toVirtualFile(inputs.setup.cacheFile.toPath),
+            extraIncOptions.value.toVector,
+            scalaVersion.value,
+          )
         }
       )
     ) ++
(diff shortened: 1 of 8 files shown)


main/src/main/scala/sbt/Defaults.scala
```diff
@@ -4415,9 +4453,9 @@ object Classpaths {
   def makeProducts: Initialize[Task[Seq[File]]] = Def.task {
     val c = fileConverter.value
     val resourceDirs = resourceDirectories.value
-    val vfBackendDir = compileIncremental.value._2
-    val backendDir = c.toPath(vfBackendDir)
-    val _ = resources.value
+    compile.value
```
eed3si9n, line 4456: So is the key to the bug fix mostly this line here?

anatoliykmetyuk: Yes, this makes products wait for full Java and Scala compilation. The core fix is in https://github.com/sbt/sbt/pull/9574/changes/d387b7fe118ac22a91d6dfe25267fd031a7bb76d. https://github.com/sbt/sbt/pull/9574/changes/d4dad933faa2a3e510fa30878d948f2f3a0ca091 adds caching to the Java compilation phase.

eed3si9n (approved):
Thanks!

PR #9575, opened 2026-08-11
[2.x] Specific error message for defining types

### LINKED ISSUE #9566: [2.x] specific error message for defining types in build.sbt
## Steps

Using sbt 2.x.

### build.sbt

build.sbt doesn't actually let you define types in it, so if you write code like this it's wrong:

```scala
case class LwjglDep(name: String, needsNatives: Boolean = true)

val lwjglDeps = 
  new LwjglDep("lwjgl") :: ???
```

## Problem

However, the error message isn't really that helpful:
```
/home/bulby/Documents/github/vulkangamescala/build.sbt:49: error: not found: type LwjglDep
  new LwjglDep("lwjgl") :: List(
```

This tells me nothing about the fact that sbt doesn't let you define types. The actual type definition passes fine which it probably shouldn't.

## Expectations

Add an error message for defining types in the `build.sbt`.

xuwei-k:
- fix https://github.com/sbt/sbt/issues/9566

### DIFF
--- a/buildfile/src/main/scala/sbt/internal/EvaluateConfigurations.scala
+++ b/buildfile/src/main/scala/sbt/internal/EvaluateConfigurations.scala
@@ -22,7 +22,7 @@ import scala.jdk.CollectionConverters.*
 import xsbti.PathBasedFile
 import xsbti.VirtualFile
 import xsbti.VirtualFileRef
-import dotty.tools.dotc.ast.untpd.{ Annotated, ValOrDefDef, Tree }
+import dotty.tools.dotc.ast.untpd.{ Annotated, ValOrDefDef, Tree, ModuleDef, TypeDef }
 
 /**
  *  This file is responsible for compiling the .sbt files used to configure sbt builds.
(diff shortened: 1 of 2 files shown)


eed3si9n:
Thanks for the contribution! Do you have an example of how it would look like with a manual test?

eed3si9n (approved):
Thanks!

xuwei-k:
```
$ cat buld.sbt 
name := "foo"

class A
sbt.version=2.1.0-SNAPSHOT
$ sbt
[info] welcome to sbt 2.1.0-SNAPSHOT (Amazon.com Inc. Java 21.0.12)
[error] buld.sbt:3: defining types in *.sbt file is not supported
```

eed3si9n:
This is nice-to-have, but I wonder if we can delay the error until we run it inside the compiler, so we can get a compile-time error message?

xuwei-k:
> if we can delay the error until we run it inside the compiler, so we can get a compile-time error message?

- https://github.com/sbt/sbt/pull/9579

PR #9579, opened 2026-08-12
[2.x] Specific error message for defining types

### LINKED ISSUE #9566: [2.x] specific error message for defining types in build.sbt

xuwei-k:
- fix https://github.com/sbt/sbt/issues/9566

### DIFF
--- a/buildfile/src/main/scala/sbt/internal/Eval.scala
+++ b/buildfile/src/main/scala/sbt/internal/Eval.scala
@@ -5,8 +5,9 @@ import dotty.tools.dotc.ast
 import dotty.tools.dotc.ast.tpd
 import dotty.tools.dotc.CompilationUnit
 import dotty.tools.dotc.core.Contexts.{ atPhase, Context }
-import dotty.tools.dotc.core.{ Flags, NameKinds, Names, Phases, Symbols, Types }
+import dotty.tools.dotc.core.{ Flags, NameKinds, Names, Phases, Symbols, Types, Constants }
 import dotty.tools.dotc.core.Periods.Period
+import dotty.tools.dotc.reporting.Diagnostic
 import dotty.tools.dotc.Driver
 import dotty.tools.dotc.Run
 import dotty.tools.dotc.util.SourceFile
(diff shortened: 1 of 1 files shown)


xuwei-k:
build.sbt

```scala
name := "foo"

class A(
  x: Int,
  y: String
)
```

image

buildfile/src/main/scala/sbt/internal/Eval.scala
```diff
@@ -267,8 +268,41 @@ class Eval(
...
+                  case defdef: tpd.DefDef if defdef.name.mangledString == WrapValName =>
+                    defdef.rhs match {
+                      case tpd.Block(
+                            (typeDef: tpd.TypeDef) :: Nil,
+                            ast.untpd.Literal(Constants.Constant(()))
+                          ) =>
+                        reporter.report(
+                          Diagnostic.Error(
+                            "Defining types in *.sbt file is not supported",
+                            typeDef.sourcePos
```
xuwei-k, line 289: https://github.com/sbt/sbt/blob/0ac15531a8cbbea89503b823e30d769d88f6f298/buildfile/src/main/scala/sbt/internal/Eval.scala#L110-L114

### build.sbt

```scala
class MyClass
```

### before

```scala
object $moduleName {
  def $WrapValName : sbt.internal.DslEntry = {
    class MyClass

    // compiler insert `Unit` value
    // confusing type error
    ()
  }
}
```


### after

```scala
object $moduleName {
  def $WrapValName : sbt.internal.DslEntry = {
    class MyClass // get this definition by `tpd.TreeTraverser`

    ()
  }
}
```

eed3si9n (approved):
Thanks!

PR #9580, opened 2026-08-12
[2.x] feat: replace buildClock with coursier input params

ecat3:
# Context 

I attempted to use `dependencyLock` and `dependencyLockCheck` in my project. `dependencyLock` produced a different `buildClock` on CI than on my machine. `sbt` did not show any explanation why the lock file is stale while refusing to use it. 

I looked through the `buildClock`-related code and found no reason to hide the coursier input params in a sha1 hash as it is done now. 

# Change 

commit by commit: 

1. `buildClock` in lock file is replaced with `RequestedInputs` (feel free to suggest a better name). This class stores the data that was used to calculate the sha1 `buildClock`.
2. `RequestedInputs` constructed in `CoursierDependencyResolution` class, skipping two rounds of constructing/deconstructing the same data
3. Simplified `RequestedInputsCompanion.build` to construct `RequestedInputs` from `scalaVersion` and `ResolutionParams` only. 

`deps.lock` now contains coursier input params. sbt now reuses resolution results if these params have not changed, and shows a slightly more detailed explanation otherwise. I can now investigate why my CI refuses to use the lock file (it was due to `~/.sbt/repositories` and `~/.config/sbt/2/<...>` shared configs).

Note: the data that was used to calculate sha1 `buildClock` appears to be insufficient: I think it is possible to pass params to coursier that will change its resolution result but won't change the `buildClock`. I have not investigated it further yet. 

AI note: some LLM assistance/autocomplete was used while working on this change. All of the code was typed manually by me.


ecat3:
CI in my fork has passed except for 1 test which seems to be unrelated: https://github.com/ecat3/sbt/pull/1 
I will check it again a bit later.

ecat3:
Oh, I see that clients can depend on `buildClock`. I need to restore it then. Will get back to it when I have a moment.

ecat3:
MIMA is green and CI should pass cleanly: https://github.com/ecat3/sbt/pull/1

lm-coursier/src/main/contraband/lockfile.contra
```diff
@@ -29,9 +29,38 @@ type LockFileMetadata {
...
+  maxIterations: Int! 
+  forceVersions: [lmcoursier.internal.RequestedForceVersion]
+  exclusions: [lmcoursier.internal.RequestedExclusion]
+  strict: String
+}
+
 type LockFileData {
   version: String!
   buildClock: String!
+  requested: lmcoursier.internal.RequestedInputs!
```
eed3si9n, line 63: I'm sympathetic to the inadvertent invalidation caused by the `buildClock`, but including all request inputs into the lockfile I think is an overkill esp if every lockfile users would have to keep the information around. I'd suggest something like a debug-level log that gets enabled if you turn on some sys.props, like https://www.scala-sbt.org/2.x/docs/en/reference/cached-task.html#execution-log.

lm-coursier/src/main/scala/lmcoursier/internal/ResolutionSerializer.scala
```diff
@@ -39,6 +40,7 @@ object ResolutionSerializer {
       LockFileData(
         version = LockFileConstants.currentVersion,
```
stasimus, line 41: Recommendation: the lock file now carries the `requested` block, so this is really a new on-disk format, but `currentVersion` is still "1.0" and nothing ever reads `version` back to check it. A lock file from an older sbt and one from this change end up with the same version string even though they are not compatible. I would bump `currentVersion` and check it on read, so you can regenerate with a clear message when they differ. Right now the field does not do anything.

lm-coursier/src/main/scala/lmcoursier/internal/ResolutionRun.scala
```diff
@@ -245,23 +234,17 @@ object ResolutionRun {
       .flatMap { lockFile =>
         LockFile.read(lockFile) match {
```
stasimus, line 235: Recommendation: an old `deps.lock` without the `requested` field will not parse, so `read` returns Left and you fall through to a full re-resolution. That is the safe result, but the parse error only shows at verbosity >= 2, so from the user side the lock just gets ignored and everything re-resolves with no reason given. Since the point of this PR is to explain why a lock is stale, it would be worth saying something at info level here too, like "lock file is from an older sbt, regenerating".

lm-coursier/src/main/scala/lmcoursier/internal/ResolutionRun.scala
```diff
@@ -245,23 +234,17 @@ object ResolutionRun {
...
-            if (
-              BuildClock.matches(
-                lockData,
-                params.dependencies,
-                params.mainRepositories,
-                scalaVersion,
-                params
-              )
-            ) {
+            if (requested == lockData.requested) {
```
stasimus, line 237: Recommendation: now that the check is `requested == lockData.requested`, `BuildClock.matches` has no callers left. `BuildClock.compute` is still used to fill in `buildClock`, but that field is only written now, never read for matching. Might be worth dropping `matches` (and maybe `buildClock` itself) or leaving a note on why they stay, so the next person does not assume buildClock still drives the decision.

lm-coursier/src/main/contraband/lockfile.contra
ecat3: I managed to debug my problem eventually - simply by calling `show <stuff>` in build scripts, no debug log. Although storing the coursier inputs explicitly would've helped, I agree it is likely excessive for most users. 

I could update the MR with your (and @stasimus') suggestions if you see sense in it. Otherwise I think this one can be closed.

PR #9583, opened 2026-08-13
[2.x] Update Eval.compileAndLoad

xuwei-k:
- https://github.com/sbt/sbt/pull/9579

compiler generate another trees if `case class`, `enum` and `object`


eed3si9n (approved):
Thanks 1000!

PR #9589, opened 2026-08-13
[2.x] refactor: Remove unused code

xuwei-k:


eed3si9n (approved):
Thanks!

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -821,50 +821,6 @@ object BuildServerProtocol {
     )
   }
 
-  private def scalacOptionsTask: Def.Initialize[Task[ScalacOptionsItem]] =
```
eed3si9n, line 824: Interestingly there's an outstanding bug that could have been fixed if this was used, I think.

PR #9597, opened 2026-08-13
[2.x] Fix warning in ivyless-publish-http-plugin test

xuwei-k:
- https://github.com/sbt/sbt/pull/9420

https://github.com/sbt/sbt/blob/c717609ab2ad2427b485c9b660a0cd4995df2693/lm-core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala#L340-L369


eed3si9n (approved):
Thanks!

Issue #9601, opened 2026-08-15
[2.x] Caching overhead on `compileIncremental`


hoangmaihuy:
**Disclaimer**: Investigated with Claude Code and human review in the loop

Ref https://github.com/sbt/sbt/issues/9586 on benchmark test
Ref https://github.com/sbt/sbt/issues/9558 on compile performance
Ref https://github.com/sbt/zinc/pull/1761 on failed attempt to fix classpath dir listing
Ref https://github.com/sbt/io/pull/540 on improving zip and jar performance

## Summary

What a no-op and a one-line-edit build cost on a large sbt 2.x workspace, where the time goes per task, and which mechanisms are responsible.

Baseline: **sbt `develop` @ `1d6d6ea66`**, built and published locally as `2.1.0-SNAPSHOT`, unmodified. Every number below comes from one measurement run of `tools/matrix.sh` on an idle machine, plus one instrumented run for the conversion counts in §4.

---

### The build under test

https://github.com/hoangmaihuy/sbt-compile-benchmark

81 modules `m1`…`m81` arranged as **three chains of depth 27**, each module depending only on its predecessor, plus one aggregate consumer:

```
m1  <- m2  <- … <- m27  ┐
m28 <- m29 <- … <- m54  ├─ app.dependsOn(m27, m54, m81)
m55 <- m56 <- … <- m81  ┘
```

So `app`'s transitive compile classpath carries **all 81 class directories**, `m27`'s carries 26, `m2`'s carries 1. Depth is deliberate: a flat graph does not reproduce the effect. (A layered graph where every module depends on a whole previous layer is worse than useless — the dependency *path* count becomes exponential and sbt's classpath computation never finishes.)

Each module generates 40 sources × 25 top-level classes = 1,000 classes, emitted into a deep package tree (`com.example.repro.mN.pS`), giving ~2,000 output files per module (`.class` + `.tasty`) and **162,003 output files** in total. Those dimensions were chosen to match a large private Scala 3 / Scala.js workspace whose app module carries 81 class directories and 163,372 files on its compile classpath.

Two hand-written files exist purely to be edited (generated sources are rewritten every run, so they cannot serve as edit targets):

- `app/src/main/scala/EditApp.scala` — in `app`, which has **no dependents**
- `m1/src/main/scala/Edit.scala` — in `m1`, the root of chain 1, with **26 downstream modules plus `app`**

### The command measured

Every scenario measures the same command, always from the `scale/` directory:

```bash
[...]

PR #9609, opened 2026-08-17
[2.x] perf: Skip re-packaging the class directory when zinc recompiles nothing

hoangmaihuy:

Fixes mechanism 3 of https://github.com/sbt/sbt/issues/9601.

## Problem

`cachedCompileIncrementalTask` ends with an unconditional `Def.declareOutputDirectory(vfDir)`, which the cached-task macro expands into `ActionCache.packageDirectory(...)` — walk the class directory, hash every file into a manifest, gzip it, install `classes.sbtdir.zip`, store the blob. That runs on every action-cache miss, whether or not zinc wrote anything.

Since `compileInputs2` puts each classpath entry's content hash into the compile cache key, a one-line body change invalidates every module downstream of it. Zinc's name hashing then invalidates nothing there, so those class directories come out byte-identical and re-zipping just reproduces a blob the CAS already holds. On the benchmark in #9601 a body change in `m1` invalidates the whole 27-module chain: one module recompiles, the other 26 re-pack ~2,000 untouched files each.

## Fix

```scala
val dirZip = ActionCache.dirZipPath(dir)
val packedDir =
  if analysisResult.hasModified() || !Files.exists(dirZip) then
    Def.declareOutputDirectory(vfDir)
  else Def.declareOutput(c.toVirtualFile(dirZip))
```

The output is still declared, so the stored `ActionResult` stays complete and a later hit restores the directory as before — same `HashedVirtualFileRef`, read off disk instead of rebuilt.

Two things worth a reviewer's attention:

- **`hasModified()` is reliable here**, though not everywhere. `compileScalaBackendTask` derives it from `readCompilations().getAllCompilations().nonEmpty`, which is always false because `ConsistentAnalysisFormat` never persists `Compilations`. At this site it comes from the live `incCompiler.compile(...)` result, where zinc computes `hasModified || hasSubprojectChange` (`Incremental.scala:378,432`) — false only when zinc took the branch that writes nothing, and conservatively true when the analysis changes but no class files do.
- **Packaging now precedes `store.set`.** A run killed between zinc writing class files and the zip being written used to leave a current analysis beside a stale zip, which the next run would reuse. Reordered, such an interruption leaves a stale analysis, forcing a recompile.

## Measurements

Benchmark: https://github.com/hoangmaihuy/sbt-compile-benchmark, 81 modules, 162,003 output files. Three runs per cell after a discarded warm run, each cell measured once, server restarted between configurations. Both builds published locally and verified by disassembly before measuring.

Deep edit (one method body in `m1`, exactly one module recompiles in every cell). `base` is `develop`, `fix` is this branch:

| tree state | exportJars | wall base | wall fix | `compileIncremental` base | `compileIncremental` fix |
| --- | --- | --- | --- | --- | --- |
| real files | false | 22.7 s | 17.5 s | 16,711 ms | 13,350 ms (−20 %) |
| real files | true | 12.0 s | **7.1 s** | 9,250 ms | **4,230 ms (−54 %)** |
| cache-restored | false | 68.1 s | 60.6 s | 45,214 ms | 38,738 ms (−14 %) |
| cache-restored | true | 21.6 s | **15.7 s** | 12,909 ms | **7,447 ms (−42 %)** |

Wall figures are the median of three runs; own-time is summed over the 82 projects under `-Dsbt.task.timings=true`. Per module, `m1` is the same in both builds (base 478 ms, fix 450 ms — it packages either way); the other 26 go from ~380–480 ms to ~150–200 ms.

The no-op and edit-leaf scenarios are unaffected, as they should be — a no-op is all cache hits, and `app` has no dependents. Three of those cells drift anyway, all in `exportJars=false`, up to −15 s on `cache-restored`/leaf; none is explicable by this change and that cell's triples overlap completely, so read it as noise. Judge the change by the `exportJars=true` rows.

## Test

`sbt-app/src/sbt-test/cache/compile-repackage-skip` — two projects, `exportJars := true`, a body-only edit upstream. It pins the condition, not just the outcome: the downstream analysis file must be rewritten (proving a real cache miss), zinc must report `hasModified = false`, the class-directory zip must not move, and deleting that zip must still let the next hit fetch it from the CAS and re-extract — so a skipped pack cannot leave an entry pointing at a missing blob.

File identity is `(fileKey, lastModifiedTime)` with `NOFOLLOW_LINKS`, which discriminates on both the symlink-relink and copy paths. Verified failing on `develop` and passing here, alongside the other 13 tests in `cache/`.

## Not included

`cachedCompileJavaTask` has the same unconditional pack, but this guard does not transfer: `Incremental.compileAllJava` returns `(sources.nonEmpty, analysis)`, so its `hasModified` reports whether the project has Java sources, not whether javac wrote anything.

### DIFF
--- a/main/src/main/scala/sbt/Defaults.scala
+++ b/main/src/main/scala/sbt/Defaults.scala
@@ -2319,13 +2319,21 @@ object Defaults extends BuildCommon with DefExtra {
       val store = analysisStore(compileAnalysisFile.value.toPath(), c)
       // TODO - Should readAnalysis + saveAnalysis be scoped by the compile task too?
       val analysisResult = Retry.io(compileIncrementalTaskImpl(bspTask, s, ci, ping, projectId))
-      val analysisOut = c.toVirtualFile(setup.cachePath())
-      val contents = AnalysisContents.create(analysisResult.analysis(), analysisResult.setup())
-      store.set(contents)
-      Def.declareOutput(analysisOut)
       val dir = ci.options.classesDirectory
       val vfDir = c.toVirtualFile(dir)
-      val packedDir = Def.declareOutputDirectory(vfDir)
+      val dirZip = ActionCache.dirZipPath(dir)
+      // Zinc leaves the class directory alone when it invalidates nothing, so the zip the previous
+      // run left behind still describes it and re-packing only reproduces a blob the store has.
+      val packedDir =
+        if analysisResult.hasModified() || !Files.exists(dirZip) then
+          Def.declareOutputDirectory(vfDir)
+        else Def.declareOutput(c.toVirtualFile(dirZip))
+      val analysisOut = c.toVirtualFile(setup.cachePath())
+      val contents = AnalysisContents.create(analysisResult.analysis(), analysisResult.setup())
+      // Packaging precedes this write so that a run interrupted in between leaves a stale analysis,
+      // which forces a recompile, rather than a current analysis paired with an outdated zip.
+      store.set(contents)
+      Def.declareOutput(analysisOut)
       s.log.debug(s"wrote $vfDir")
       (analysisResult.hasModified(), vfDir: VirtualFileRef, packedDir: HashedVirtualFileRef)
     }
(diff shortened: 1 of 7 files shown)


eed3si9n (approved):
Thanks!

PR #9610, opened 2026-08-17
[2.x] fix: Fixes scalacOptions in BSP using VirtualFileRef

### LINKED ISSUE #9578: [sbt 2] Scalac options in both sbt and BSP contain variable references.(e.g., ${CSR_CACHE})
## steps

1. Create an sbt project like in https://youtrack.jetbrains.com/issue/SCL-25761 (the code below is copied from there) 
src/main/scala/BetterMonadicFor.scala
```
object BetterMonadicFor {
  def test(): Option[String] = for {
    case implicit0(s: String) <- Some("x")
  } yield s
}
```
build.sbt
```
  .settings(
    name := "sbt2test",
    addCompilerPlugin("com.olegpy" %% "better-monadic-for" % "0.3.1"),
    scalaVersion := "2.13.18"
  )
```
project/build.properties
`sbt.version = 2.0.5`

2. Start sbt in the terminal and run `show bspBuildTargetScalacOptionsItem`

sbt 2.x returns: 
```
sbt:root> show bspBuildTargetScalacOptionsItem
[info] ScalacOptionsItem(BuildTargetIdentifier(file:~/IdeaProjects/sbt-demo/better_monadic/#root/Compile), Vector(-Xplugin:${CSR_CACHE}/https/repo1.maven.org/maven2/com/olegpy/better-monadic-for_2.13/0.3.1/better-monadic-for_2.13-0.3.1.jar) ......))
```
As can be seen, the better-monadic jar path contains the ${CSR_CACHE} placeholder. The same result is present when running only the `scalacOptions` task. I guess this was done to support caching for the `scalacOptions` task, but it causes some issues, as described below.

## problem
In the Scala plugin, we support the sbt projects in two ways: directly through sbt or through a BSP server. When the _direct sbt_ option is used, we rely on the [sbt-structure](https://github.com/JetBrains/sbt-structure) plugin to extract information about the sbt project. Among other things, we read the `scalacOptions` task. If it returns a value containing a placeholder which we do not resolve, the jar path is incorrect and project compilation fails. When the BSP option is used for project import, the Scala plugin acts only as a BSP client. It sends requests to the BSP server and reads the responses. If the BSP response contains placeholders in the compiler plugin jar paths, they are also not resolved on our side, which causes issues such as incorrect highlighting.

## possible solutions
To fix the sbt import in the Scala plugin, we could in theory duplicate sbt's placeholder [resolution logic](https://github.com/sbt/sbt/blob/de1b5ee9eaff469f3b0e11d20d7fed31a13ba4e1/main/src/main/scala/sbt/internal/Compiler.scala#L462) in the `sbt-structure` plugin. However, it would be better to reuse an sbt utility or another task instead of duplicating this logic. The BSP import is more complicated because we rely entirely on what the BSP server returns, which I believe is the intended BSP behavior.

One possible solution would be to add a new non-cached sbt task, such as `resolvedScalacOptions`, which would return fully resolved scalac options without placeholders. The BSP implementation could use this task, and we could also use it in the `sbt-structure` plugin instead of duplicating sbt internals.

azdrojowa123:
Fixes https://github.com/sbt/sbt/issues/9578


main/src/main/scala/sbt/Keys.scala
```diff
@@ -217,6 +217,7 @@ object Keys {
   val autoCompilerPlugins = settingKey[Boolean]("If true, enables automatically generating -Xplugin arguments to the compiler based on the classpath for the " + CompilerPlugin.name + " configuration.").withRank(AMinusSetting)
   val maxErrors = settingKey[Int]("The maximum number of errors, such as compile errors, to list.").withRank(ASetting)
   val scalacOptions = taskKey[Seq[String]]("Options for the Scala compiler.").withRank(BPlusTask)
+  val resolvedScalacOptions = taskKey[Seq[String]]("scalacOptions with cache placeholders (e.g. ${CSR_CACHE}) resolved to absolute machine paths.").withRank(BPlusTask)
```
eed3si9n, line 221: Could we mark this `@transient` plz, so it would not participate in cached tasks?

eed3si9n (commented):
overall lgtm. thanks!

main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala
```diff
@@ -330,7 +332,7 @@ object BuildServerProtocol {
     bspBuildTargetRun := bspRunTask.evaluated,
     bspBuildTargetScalacOptionsItem := {
       val target = Keys.bspTargetIdentifier.value
-      val scalacOptions = Keys.scalacOptions.value.toVector
+      val scalacOptions = Keys.resolvedScalacOptions.value.toVector
```
eed3si9n, line 335: <s>
```
      val scalacOptions = (Keys.resolvedScalacOptions.value: nowarn("transient")).toVector
```
</s>

azdrojowa123: Hmm, I see. However, after marking `resolvedScalacOptions` as transient, I noticed one thing that concerns me.

For example, I first have `scalacOptions += "-deprecation"` in the `build.sbt`. I run `show bspBuildTargetScalacOptionsItem` and see this flag in the output. Then I comment out the `scalacOptions` setting, reload the session, and run `show bspBuildTargetScalacOptionsItem` again. It returns stale output that still contains the `-deprecation` flag, even though it should no longer be there. Before marking `resolvedScalacOptions` as transient, it returned the correct results.

I guess this happens because we excluded the scalac options from caching, while the outer `bspBuildTargetScalacOptionsItem` task is still cacheable.

I'm not sure what the proper fix would be. The only idea I had was to make the entire `bspBuildTargetScalacOptionsItem` task uncacheable. But if we do that, I'm not sure whether it would break the cacheability of the outer `bspBuildTargetScalacOptions` task.

eed3si9n: Yea, we probably should make bspBuildTargetScalacOptionsItem transient or use `Def.uncached`.

azdrojowa123: What about the outer `bspBuildTargetScalacOptions` task? It's the task that is executed for the `buildTarget/scalacOptions` request.

eed3si9n: `bspBuildTargetScalacOptions` is an input task, so it should be uncached already.

eed3si9n (approved):
Thanks!

PR #9613, opened 2026-08-17
[1.x] Avoid rewriting unchanged plugin descriptors

### LINKED ISSUE #9612: PluginDiscovery rewrites identical sbt.autoplugins descriptors and invalidates timestamp-based caches
_(Disclaimer: The summary and reproducers are minimised and prepared by Codex)_

## Summary

For an sbt plugin project with an unchanged discovered `AutoPlugin`, sbt's built-in `sbtPlugin` resource generator rewrites `sbt/sbt.autoplugins` even when its content is identical. The descriptor therefore receives a newer modification time.

This reproduces in sbt 1.12.15 and the latest stable sbt 2.0.6. Its downstream effect depends on the packager.

`Status` below assesses whether that packaging path needlessly rebuilds or re-materializes its artifact; it does not mean the descriptor rewrite is absent.

| Build tool and packaging path | Status | Result after an unchanged second invocation |
|---|---|---|
| sbt 1.12.15, default `Compile / packageBin` | Bad | The identical thin JAR is rewritten. |
| sbt 1.12.15, `Compile / packageBin := (Compile / assembly).value` | Bad | The identical assembly JAR is rewritten. |
| sbt 2.0.6, default `Compile / packageBin` | Ok | The descriptor is rewritten, but SBT 2 returns the same virtual, content-addressed package output as a cache hit. |
| sbt 2.0.6, `Compile / packageBin := (Compile / assembly).value` | Bad | The identical physical assembly JAR is rewritten. |

Thus SBT 2 mitigates the default `packageBin` consequence with its content-addressed cache, but it does not make the generated descriptor idempotent. Timestamp-based downstream packagers, including sbt-assembly, remain affected.

## Reproducers

Attached archives:

- [sbt-packagebin-comparison-sbt1.zip](https://github.com/user-attachments/files/31140533/sbt-packagebin-comparison-sbt1.zip)
- [sbt-packagebin-comparison-sbt2.zip](https://github.com/user-attachments/files/31140539/sbt-packagebin-comparison-sbt2.zip)

Each archive is a multi-project build with two independent plugin subprojects:

- `default-packagebin`: `sbtPlugin := true` and SBT’s unmodified `Compile / packageBin`.
- `redefined-packagebin`: the production-style override below:

  ```scala
  Compile / packageBin := (Compile / assembly).value
  ```

Run `./verify.sh` in each extracted archive. The scripts perform two unchanged package invocations, compare SHA-256 values and mtimes, and fail if the expected behavior is not observed.

The SBT 2 assembly-backed subproject also sets:

```scala
Compile / exportJars := false
```

This avoids a separate `packageBin -> fullClasspath -> packageBin` cycle caused by redirecting `packageBin` to `assembly`; it is not part of the descriptor timestamp issue.

## Observed

In all cases, the generated `sbt/sbt.autoplugins` descriptor has identical SHA-256 bytes before and after the second invocation, but a newer modification time.

For SBT 1 default `packageBin` and both assembly-backed variants, the output JAR likewise retains identical bytes but receives a newer mtime. This demonstrates needless packaging work caused only by the rewritten descriptor timestamp.

## Expected

When the sorted discovered plugin names have not changed, `sbt/sbt.autoplugins` should not be rewritten and its mtime should remain unchanged. This lets timestamp-based packagers remain cache hits and avoids needless output-JAR writes.

## Cause

Both releases install an sbt-owned resource generator:

- sbt 1.12.15: [Defaults.scala](https://github.com/sbt/sbt/blob/v1.12.15/main/src/main/scala/sbt/Defaults.scala#L646-L649)
- sbt 2.0.6: [Defaults.scala](https://github.com/sbt/sbt/blob/v2.0.6/main/src/main/scala/sbt/Defaults.scala#L659-L663)

`PluginDiscovery.writeDescriptor` unconditionally calls `IO.writeLines(descriptor, names.distinct.sorted)`:

- sbt 1.12.15: [PluginDiscovery.scala](https://github.com/sbt/sbt/blob/v1.12.15/main/src/main/scala/sbt/internal/PluginDiscovery.scala#L84-L93)
- sbt 2.0.6: [PluginDiscovery.scala](https://github.com/sbt/sbt/blob/v2.0.6/main/src/main/scala/sbt/internal/PluginDiscovery.scala#L83-L92)

sbt-assembly caches based on input timestamps: [caching documentation](https://github.com/sbt/sbt-assembly#caching).

## Suggested fix

Make `PluginDiscovery.writeDescriptor` content-aware: calculate the sorted descriptor lines, compare them with the existing descriptor, and write only when they differ. Apply this to both `sbt/sbt.autoplugins` and `sbt/sbt.builds`.

## Related issues

- [sbt-assembly#59 — caching touches target JAR when source has not changed](https://github.com/sbt/sbt-assembly/issues/59)
- [sbt#1832 — `resourceGenerators` execution behavior](https://github.com/sbt/sbt/issues/1832)
- [sbt#9252 — duplicate `sbt/sbt.autoplugins` on repeated warm-server publishing](https://github.com/sbt/sbt/issues/9252)
- [sbt#9113 — duplicate resource entries in sbt 2.x packaging](https://github.com/sbt/sbt/issues/9113)
- [sbt#8391 — published plugins without `sbt/sbt.autoplugins`](https://github.com/sbt/sbt/issues/8391)
- [sbt#1864 — historical missing `sbt/sbt.autoplugins` entry](https://github.com/sbt/sbt/issues/1864)

unkarjedy:
Fixes #9612.

## Summary

Make `PluginDiscovery.writeDescriptor` content-aware. When the discovered
plugin names are unchanged, the generated `sbt/sbt.autoplugins` or
`sbt/sbt.builds` descriptor is no longer rewritten.

## Root cause and impact

`sbtPlugin` runs the descriptor generator during packaging. Previously it
unconditionally wrote the descriptor even when its normalized lines were
identical. That changed the descriptor mtime, invalidated timestamp-based
package inputs, and caused repeated direct `packageBin` and `assembly` runs to
rewrite otherwise identical JARs.

The generator now keeps the descriptor untouched when its existing lines equal
the newly discovered, sorted names. A real plugin discovery change still writes
the descriptor and correctly invalidates the artifact.

## Validation

- Added unit coverage for unchanged, changed, and empty discovery results.
- Added scripted end-to-end fixtures for direct `packageBin` and real
  `sbt-assembly` 2.4.1 `assembly` packaging.
- TDD confirmation: both no-op packaging scenarios failed with the former
  unconditional write and pass with this change; both fixtures also verify that
  adding an `AutoPlugin` rebuilds the JAR.
- Ran:

  ```text
  CI=true sbt -batch ';project mainProj; testOnly sbt.internal.PluginDiscoveryTest'
  CI=true sbt -batch 'scripted package/sbt-plugin-incremental plugins/sbt-assembly'
  ```

### DIFF
--- a/main/src/main/scala/sbt/internal/PluginDiscovery.scala
+++ b/main/src/main/scala/sbt/internal/PluginDiscovery.scala
@@ -88,7 +88,10 @@ object PluginDiscovery {
       IO.delete(descriptor)
       None
     } else {
-      IO.writeLines(descriptor, names.distinct.sorted)
+      val lines = names.distinct.sorted
+      // Do not invalidate timestamp-based downstream caches when the descriptor content is unchanged.
+      if (!descriptor.exists || IO.readLines(descriptor) != lines)
+        IO.writeLines(descriptor, lines)
       Some(descriptor)
     }
   }
--- a/main/src/test/scala/sbt/internal/PluginDiscoveryTest.scala
+++ b/main/src/test/scala/sbt/internal/PluginDiscoveryTest.scala
@@ -1,0 +1,57 @@
+/*
+ * sbt
+ * Copyright 2023, Scala center
+ * Copyright 2011 - 2022, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.internal
+
+import java.nio.file.attribute.FileTime
+import java.nio.file.Files
+
+import sbt.io.IO
+
+object PluginDiscoveryTest extends verify.BasicTestSuite {
+  import PluginDiscovery.Paths.AutoPlugins
+
+  test("writeDescriptor preserves mtime when the descriptor content is unchanged") {
+    val directory = Files.createTempDirectory("plugin-discovery-test").toFile
+    try {
+      val descriptor =
+        PluginDiscovery.writeDescriptor(Seq("example.B", "example.A"), directory, AutoPlugins).get
+      val expectedTime = FileTime.fromMillis(1234L)
+      Files.setLastModifiedTime(descriptor.toPath, expectedTime)
+
+      PluginDiscovery.writeDescriptor(
+        Seq("example.A", "example.B", "example.A"),
+        directory,
+        AutoPlugins
+      )
+
+      assert(Files.getLastModifiedTime(descriptor.toPath) == expectedTime)
+    } finally IO.delete(directory)
+  }
+
+  test("writeDescriptor updates the descriptor when its content changes") {
+    val directory = Files.createTempDirectory("plugin-discovery-test").toFile
+    try {
+      val descriptor = PluginDiscovery.writeDescriptor(Seq("example.A"), directory, AutoPlugins).get
+
+      PluginDiscovery.writeDescriptor(Seq("example.B"), directory, AutoPlugins)
+
+      assert(IO.readLines(descriptor) == Seq("example.B"))
+    } finally IO.delete(directory)
+  }
+
+  test("writeDescriptor deletes the descriptor when no modules are discovered") {
+    val directory = Files.createTempDirectory("plugin-discovery-test").toFile
+    try {
+      val descriptor = PluginDiscovery.writeDescriptor(Seq("example.A"), directory, AutoPlugins).get
+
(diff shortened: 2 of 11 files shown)


unkarjedy:
I will make a PR for SBT 2 later today.
But the full fix relies on the changes in https://github.com/sbt/sbt-assembly/issues/580
I will make a PR for it today as well

eed3si9n (approved):
The change lgtm. Thanks for the contribution!

PR #9615, opened 2026-08-17
[2.x] Remove Ivy dependency

eed3si9n:
## Problem
sbt still has some code that's depending on Ivy.

## Solution
1. This removes remaining Ivy dependent code from the sbt main.
2. This also removes lm-ivy integration code from the repo.


lm-core/src/main/scala/sbt/internal/librarymanagement/mavenint/PomExtraAttributeKeys.scala
```diff
@@ -0,0 +1,15 @@
+package sbt.internal.librarymanagement
+package mavenint
+
+/**
+ * Keys for the extra pom properties used to propagate the sbtVersion/scalaVersion
+ * attributes for sbt plugin dependencies, plus the Maven packagings that sbt treats as jars.
+ */
+object PomExtraAttributeKeys {
+  val ExtraAttributesKey = "extraDependencyAttributes"
```
stasimus: Possibly-dead public constant?

sbt-app/src/test/scala/sbt/RunFromSourceMain.scala
```diff
@@ -147,20 +147,17 @@ object RunFromSourceMain {
...
         val lm = {
-          import sbt.internal.librarymanagement.ivy.{
-            InlineIvyConfiguration,
-            IvyDependencyResolution
-          }
-          val ivyConfig = InlineIvyConfiguration().withLog(log)
-          IvyDependencyResolution(ivyConfig)
+          import lmcoursier.{ CoursierConfiguration, CoursierDependencyResolution }
+          CoursierDependencyResolution(
+            CoursierConfiguration().withResolvers(Vector(Resolver.mavenCentral))
```
stasimus: the resolver set is now Maven Central only. If CI ever passes a scalaVersion that isn't on Central (a nightly/snapshot 3.x compiler), retrieve returns Left and the test throws w.resolveException. The old InlineIvyConfiguration().withLog(log) carried a different default resolver chain?

sbt-app/src/test/scala/sbt/RunFromSourceMain.scala
```diff
@@ -147,20 +147,17 @@ object RunFromSourceMain {
...
         }
         val Name = """(.*)(?:\-[\d.]+)\.jar""".r
         val BinPre = """(.*)(?:\-[\d.]+)-(?:bin|pre)-.*\.jar""".r
         val module = "org.scala-lang" % "scala3-compiler_3" % scalaVersion
         lm.retrieve(module, scalaModuleInfo = None, scalaHome1Temp, log) match {
-          case Left(w)  => throw w.resolveException
-          case Right(_) =>
-            val jars = (scalaHome1Temp ** "*.jar").get()
+          case Left(w)     => throw w.resolveException
+          case Right(jars) =>
```
stasimus, line 158: the code now trusts the value returned by lm.retrieve(...) instead of the old (scalaHome1Temp ** "*.jar").get() glob

This is fine as long as CoursierDependencyResolution.retrieve returns the actual retrieved jar Files (seems it does into scalaHome1Temp?), but it's the line that changed the source of truth for the jar list comparing to previous behaivour.

lm-core/src/main/scala/sbt/internal/librarymanagement/mavenint/PomExtraAttributeKeys.scala
eed3si9n: Yea, this was in theory used to encode Ivy extra attributes, but in practice not really used, I think. Removed in https://github.com/sbt/sbt/pull/9615/commits/e92f83c35635e78b12986800250171b025f86b60.

sbt-app/src/test/scala/sbt/RunFromSourceMain.scala
eed3si9n: `InlineIvyConfiguration()` and `CoursierConfiguration()` both initialize to `Resolver.defaults`, which is `Vector(Resolver.mavenCentral)`, so in this case `.withResolvers(Vector(Resolver.mavenCentral))` is redundant. I think this is fine since it's only used to launch freshly baked sbt, and we only use Scala artifacts published to Maven Central.

sbt-app/src/test/scala/sbt/RunFromSourceMain.scala
eed3si9n: Yea, I think this is because Ivy has a built-in functionality to retrieve JAR files to a directory, but Cousier's LM doesn't have the feature. It's re-copied a few lines below, so overall this doesn't really matter.

sbt-app/src/test/scala/sbt/RunFromSourceMain.scala
eed3si9n: Simplified in b823da2cd42fb36c0d8248f690b951c8b2149d78

xuwei-k:
- https://github.com/sbt/sbt/issues/9676

xuwei-k:
- https://github.com/sbt/sbt/issues/9679

PR #9617, opened 2026-08-17
[2.x] fix: sbtn to retry on corrupt active.json

### LINKED ISSUE #9606: Empty or malformed project/target/active.json makes every sbt invocation exit 1 with no output (sbt 2.0.6, not just -bsp; cf #5978)
## Problem

If `project/target/active.json` is empty or unparseable, **every `sbt` invocation exits 1 and prints nothing at all** — 4 bytes of output, which is a bare `ESC[0J` (erase-to-end-of-screen) and no message. There is no way to tell from the output that sbt ran, let alone why it failed.

This is [#5978](https://github.com/sbt/sbt/issues/5978) again. That issue was closed, but it was scoped to `sbt -bsp` on sbt 1.4.0; this reproduces on **sbt 2.0.6 with a plain `sbt` invocation**, no `-bsp` involved.

## Steps to reproduce

```bash
mkdir -p /tmp/sbtrepro/project && cd /tmp/sbtrepro
echo 'ThisBuild / scalaVersion := "3.3.4"' > build.sbt
echo 'sbt.version=2.0.6' > project/build.properties

sbt -batch "print name"          # works, exit 0, starts a server

: > project/target/active.json   # truncate it
sbt -batch "print name"          # exit 1, NO output
```

## Expected

A diagnostic naming the file, e.g. `could not read project/target/active.json (empty or malformed); delete it and retry` — or simply treat an unreadable `active.json` the way a missing one is treated, and start a new server.

## Actual

```
$ sbt -batch "print name" ; echo "exit=$?"
exit=1
$ sbt -batch "print name" 2>&1 | wc -c
4
```

Removing the file makes the same command succeed immediately, so the file's content is the sole cause:

```
removed: exit=0
empty:   exit=1  bytes=4
```

## Also affected

Every malformed shape behaves identically — silent exit 1:

| `active.json` content | exit | output bytes |
|---|---:|---:|
| *(file removed)* | 0 | normal |
| *(empty)* | 1 | 4 |
| `not json` | 1 | 4 |
| `{}` | 1 | 4 |
| `{"uri":""}` | 1 | 4 |

A **stale but well-formed** file (server killed with `SIGKILL`, socket gone) is handled correctly — sbt starts a new server. So the gap is specifically the parse/validate path, not staleness.

## Why it matters

The failure gives the user nothing to act on. In our case it stopped an entire build with no diagnostic, and the cause was only found by inspecting the file by hand. A one-line message, or falling back to the missing-file path, would make it self-correcting.

## Environment

- sbt: **2.0.6** (runner 2.0.0)
- Scala: 3.3.4 (also seen on 3.9.0-RC4)
- JDK: Temurin 25.0.2
- macOS 26.5.2 (arm64)

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/9606

## Problem
sbtn gets stuck when active.json is corrupt.

## Solution
Delete active.json, and retry.

### DIFF
--- a/protocol/src/main/scala/sbt/protocol/ClientSocket.scala
+++ b/protocol/src/main/scala/sbt/protocol/ClientSocket.scala
@@ -12,6 +12,7 @@ package protocol
 import java.io.{ File, InputStream, OutputStream }
 import java.net.{ InetAddress, Socket, StandardProtocolFamily, URI, UnixDomainSocketAddress }
 import java.nio.channels.{ Channels, SocketChannel }
+import scala.util.control.NonFatal
 import sjsonnew.BasicJsonProtocol
 import sjsonnew.support.scalajson.unsafe.{ Parser, Converter }
 import sjsonnew.shaded.scalajson.ast.unsafe.JValue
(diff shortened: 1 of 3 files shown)


protocol/src/main/scala/sbt/protocol/ClientSocket.scala
```diff
@@ -23,18 +24,26 @@ import org.scalasbt.ipcsocket.*
...
     // println(uri)
     val token = p.tokenfilePath map { tp =>
       val tokeFile = new File(tp)
-      val json: JValue = Parser.parseFromFile(tokeFile).get
-      val t = Converter.fromJson[TokenFile](json).get
-      t.token
+      try
+        val json: JValue = Parser.parseFromFile(tokeFile).get
+        Converter.fromJson[TokenFile](json).get.token
+      catch case NonFatal(e) => throw new ConnectionFileReadException(tokeFile, e)
```
anatoliykmetyuk, line 46: Since `NetworkClient` does not differentiate which file was problematic in its exception catching logic, a malformed `token.json` can trigger a deletion of `active.json`, orphaning an active sbt server and thereby causing a resource leak.

eed3si9n: I am not really sure what circumstances where either JSON file is empty or corrupt, but delete-and-retry kind of assumes that server 1 crashed without deleting the JSON files, and in that scenario, if we deleted `active.json` then server 2 would hopefully just overwrite the token file too - https://github.com/sbt/sbt/blob/a1114188b4fce09933430b13734516d482068c91/main-command/src/main/scala/sbt/internal/server/Server.scala#L217-L219.

In the situation where server 1 is still alive, then we'd likely need manual intervention.

anatoliykmetyuk: Yeah, this is a corner case. It may happen if there are two clients connecting in parallel to one server. If one of them connects first, then the token regeneration sequence is triggered. The second client's read of the token file is not synchronized with that regeneration sequence, so it may consume incomplete token and fail, causing the deletion of `active.json` and orphaning the server.

eed3si9n: Also the token file is used only for TCP socket, which is rare, so the corner is relatively small.

anatoliykmetyuk: Yeah, I agree, appears to be quite a narrow case.

anatoliykmetyuk (approved):

Issue #9618, opened 2026-08-17
[2.x] Java-only subprojects compile twice concurrently in the same command


jozanek:
## steps

1. sbt 2.0.5 or 2.0.6, a build with a Java-only subproject (`autoScalaLibrary := false`, only `.java` sources) and downstream subprojects that depend on it. Observed in sbt/zinc, whose `compiler-interface` has 138 Java sources.
2. Run `Test/compile` with a cold output directory.

## problem

The Java-only subproject's main compilation runs twice, concurrently, into the same classes directory. In zinc's CI, "compiling 138 Java sources to .../compiler-interface/classes" appears twice, 0.6s apart, on both Linux and Windows, while every Scala subproject compiles exactly once ([Linux job](https://github.com/sbt/zinc/actions/runs/31640624007/job/95480106402), [Windows job](https://github.com/sbt/zinc/actions/runs/31640624007/job/95480105695)).

Both executions package the classes directory into `classes.sbtdir.zip`. On Linux this is only wasted work. On Windows the second install's `Files.move` fails with `AccessDeniedException` while the first install still holds the zip open, failing the build in `(compilerInterface / Compile / compileIncremental)`. This is a recurring CI failure on zinc (also [this run](https://github.com/sbt/zinc/actions/runs/31641838797) on an unrelated branch):

```
[error] java.nio.file.AccessDeniedException: D:\a\zinc\zinc\target\out\jvm\u\compiler-interface\classes.sbtdir.zip.2301717692725912204.sbtdir.zip.tmp -> D:\a\zinc\zinc\target\out\jvm\u\compiler-interface\classes.sbtdir.zip
[error] 	at java.base/sun.nio.fs.WindowsFileCopy.move(WindowsFileCopy.java:317)
[error] 	at java.base/java.nio.file.Files.move(Files.java:1432)
[error] 	at sbt.util.ActionCache$.installPackagedZip(ActionCache.scala:371)
[error] 	at sbt.util.ActionCache$.packageDirectory$$anonfun$1(ActionCache.scala:427)
[error] 	at sbt.util.ActionCache$.packageDirectory(ActionCache.scala:410)
[error] 	at sbt.Defaults$.$init$$$anonfun$1$$anonfun$1(Defaults.scala:2285)
[error] 	at sbt.util.ActionCache$.cache(ActionCache.scala:186)
[error] (compilerInterface / Compile / compileIncremental) java.nio.file.AccessDeniedException: ...
```

## expectation

[...]

PR #9619, opened 2026-08-17
[2.x] fix: Tolerate a locked dirzip destination on Windows

jozanek:
On Windows, `Files.move` cannot replace a file that another process holds open. Under parallel task execution two tasks can install the same `classes.sbtdir.zip` concurrently while each also reads the zip it just installed (digest, CAS copy, extraction), so `installPackagedZip` fails with `AccessDeniedException`. Seen in the wild as recurring sbt/zinc CI failures on `(compilerInterface / Compile / compileIncremental)`, triggered by the duplicate compilation in #9618.

The zip is now installed via `IO.copyFile`, which stages next to the destination and retries the atomic rename; if the destination still cannot be replaced but exists, the concurrently installed zip wins. Callers derive digests from the file on disk, so the surviving zip determines the recorded content either way. The new test races installs against readers of the installed zip and reproduces the failure on Windows without this change.

Investigated with Claude Code, with human review in the loop.


util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -374,6 +377,16 @@ object ActionCache:
...
+    try
+      Files.copy(builtZip, staging, StandardCopyOption.REPLACE_EXISTING)
+      try Retry(moveIntoPlace())
+      catch
+        // Windows refuses to replace a file that another process holds open. Under
+        // parallel task execution a concurrent install of the same directory may
+        // still be reading `destZip` after the retries are exhausted; accept its
+        // zip in that case. Callers derive digests from the file on disk, so the
+        // surviving install determines the recorded content either way.
+        case _: AccessDeniedException if Files.exists(destZip) => ()
```
eed3si9n, line 389: Thanks for looking into this. The implementation looks equivalent to `IO.writeFileAtomically`, so I suggest using `IO.copyFile`.

eed3si9n (changes requested):

util-cache/src/main/scala/sbt/util/ActionCache.scala
ColOfAbRiX: A similar issue has been affecting me badly because, on one side, metals is not reliable and I can't use it or MCP, and, on the other side, sbt keeps locking files impacting manual execution and AI. The problem I have, though, is not just about parallel execution.

I've been working on a fix too, but my fix is on the read side. Let me publish my WIP solution, once I solve my env problems

ColOfAbRiX: Here it is https://github.com/sbt/sbt/pull/9627 It's still a draft as I haven't done all the checks I'd like to do and I haven't validated against the project's guidelines.

I hope this will start a good conversation around this problem

jozanek: Good call, switched to `IO.copyFile` in 758c31e14. The only thing kept in `ActionCache` is the narrow `AccessDeniedException` fallback: with tight-loop readers on Windows the destination can still be held open after `IO.copyFile`'s retries are exhausted, and in that case the concurrently installed zip wins.

PR #9620, opened 2026-08-18
[2.x] fix: Name the platform in CrossVersion(module, scalaModuleInfo)

kitbellew:
**Problem**

`CrossVersion(module, scalaModuleInfo)` drops the platform, so it names a Scala.js or Scala Native artifact `foo_2.13` rather than `foo_sjs1_2.13`.

A caller outside sbt has no way to arrive at the published name: `addPlatformSuffix` is `private[sbt]`, and this overload ignores `ScalaModuleInfo.platform`.

On sbt 1 the same call was correct, because the platform was a prefix inside the `CrossVersion` itself.

Found via sbt-shading, which rebuilds a module's name to locate the project's own entry in the update report (`ShadingPlugin.shadedJars`). On a Scala.js or Native row that name matches nothing, so no dependency is recognised as direct, every module counts as shaded, and the entire classpath is bundled into the published jar — `sourcecode`, `niocharset`, `LICENSE`, `rootdoc.txt` and so on, until `onlyNamespaces` asserts.

**Solution**

Name the artifact in full: platform suffix before cross suffix, matching the coordinate (#9117), through `addPlatformSuffix`, which already knows that `jvm` contributes no suffix.

The alternative is to make `addPlatformSuffix` public and have plugins call it explicitly. That costs every cross-built plugin a version-specific source set, since it does not exist on sbt 1, whereas this overload does and is already correct there.

### DIFF
--- a/lm-core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
+++ b/lm-core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
@@ -139,9 +139,17 @@ private[librarymanagement] abstract class CrossVersionFunctions {
         append(c.prefix + compat + c.suffix)
     }
 
-  /** Constructs the cross-version function defined by `module` and `is`, if one is configured. */
+  /**
+   * Constructs the cross-version function defined by `module` and `is`, if one is configured.
+   *
+   * The function names the artifact in full, platform suffix before cross suffix, matching the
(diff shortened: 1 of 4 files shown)


kitbellew:
@eed3si9n going through the thorny process of converting scalameta repos to sbt2 and ran into a wall with scalameta itself (converted munit and scalafmt so far). apparently, to fix, claude wants to fix this here, and another in sbt-shading.

eed3si9n (approved):
Thanks for the contribution. The change lgtm

PR #9623, opened 2026-08-18
[2.x] perf: Copy instead of symlink in disk cache on APFS

hoangmaihuy:

Fixes mechanism 2 in #9601

### Problem

Resolving a symlink reads an inode out of a CAS far too large to stay in the OS metadata cache: **165 µs cold against 3.6 µs for a regular file**, measured on a real 57 GB / 952,605-blob CAS. A real-file tree answers a walk from `readdir`'s `d_type` with no per-file `stat` at all.

### Changes

`DiskActionCacheStore` seeds the existing `symlinkSupported` latch from the filesystem holding the CAS, and copies rather than symlinks on APFS:

```scala
// Util.scala, alongside isMac / isWindows
def isApfs(path: Path): Boolean =
  try Files.getFileStore(path).`type`().equalsIgnoreCase("apfs")
  catch case NonFatal(_) => false

// ActionCacheStore.scala
private lazy val symlinkSupported: AtomicBoolean = AtomicBoolean(!Util.isApfs(casBase))
```

**On APFS the copy is nearly free.** `copyFile` is `Files.copy(COPY_ATTRIBUTES, REPLACE_EXISTING)`, which on JDK 25 reaches `clonefile(2)`. Each entry becomes a real file with its own inode, so the walk runs at full speed while its data blocks stay shared with the CAS — **11 MB of real disk against 137 MB for the symlinked tree**. There is barely any disk to trade, which is why this is worth doing by default rather than as an option.

**A copy also cannot corrupt the cache.** Writing through a symlink writes into the CAS blob, so any tool that opens a restored output for update rather than replacing it damages the shared cache. This is the same hazard you raised against hard links; a copy does not have it.

**And `target/out` stops depending on the global cache.** Today it is symlinks into the shared CAS, so wiping that cache leaves a checkout's class files unreadable. sbt recovers on the next build, but that recovery is a full recompile, and until then anything else reading `target/out` — an IDE, `javap`, a packaging script — sees broken files.

### Benchmark

81 modules, 162,003 output files, `exportJars=true`, macOS/APFS, JDK 25. Steady-state means; the two sbt builds published as separate versions and verified distinct.

| scenario | symlinks (today) | copies (this PR) | real files (ceiling) | speedup |
|---|---|---|---|---|
| no-op | 14.75s | **5.96s** | 5.89s | **2.47x** |
| edit leaf | 15.61s | **6.30s** | 6.01s | **2.48x** |
| edit deep dep | 17.84s | **7.54s** | 7.29s | **2.36x** |
| cache restore | 43.63s | 52.10s | 40.06s | +19% |

Copies land within 1-5% of the real-files ceiling. The +19% on restore is repaid by the first no-op.

Disk: **11 MB against 137 MB symlinked**, because `Files.copy` reaches `clonefile(2)`. Where cloning is unavailable the same tree costs **613 MB**, measured by dereferencing it with a non-cloning copy. `du` cannot see clones and reports 880 MB; `df` is the honest number.

### References

- [`DiskCacheClient.java`](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/remote/disk/DiskCacheClient.java) — Bazel's disk cache copies and never symlinks, through the same `Files.copy` call with the same flags.
- [bazelbuild/bazel#16711](https://github.com/bazelbuild/bazel/issues/16711) — "Symlinked sandbox is slow", with reports of 300K-input actions. Five mitigations listed, none taken; P3, unassigned.

### DIFF
--- a/util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
+++ b/util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
@@ -196,7 +196,9 @@ case class DiskActionCacheStore(base: Path, converter: FileConverter)
     dir
   }
 
-  private val symlinkSupported: AtomicBoolean = AtomicBoolean(true)
+  // Files.copy clones on APFS, so a real file costs no extra disk while sparing every later
+  // directory walk the CAS inode lookup that resolving a symlink pays.
+  private lazy val symlinkSupported: AtomicBoolean = AtomicBoolean(!Util.isApfs(casBase))
 
   override def storeName: String = "disk"
 
--- a/util-cache/src/test/scala/sbt/util/ActionCacheTest.scala
+++ b/util-cache/src/test/scala/sbt/util/ActionCacheTest.scala
@@ -8,6 +8,7 @@ import java.util.concurrent.{ CyclicBarrier, ExecutorService, Executors, TimeUni
 
 import sbt.internal.util.CacheEventLog
 import sbt.internal.util.StringVirtualFile1
+import sbt.internal.util.Util
 import sbt.io.IO
 import sbt.io.syntax.*
 import verify.BasicTestSuite
@@ -140,6 +141,7 @@ object ActionCacheTest extends BasicTestSuite:
         cache.syncBlobs(refs, outputDirectory)
         assert((dir / "a.txt").exists, "a.txt not re-extracted after the directory was deleted")
         assert((dir / "b.txt").exists, "b.txt not re-extracted after the directory was deleted")
+        assertMaterialized(cache, (dir / "a.txt").toPath)
 
   test("Disk cache does not re-extract a dirzip whose archive digest already matches"):
     withDiskCache: cache =>
@@ -160,7 +162,7 @@ object ActionCacheTest extends BasicTestSuite:
         cache.syncBlobs(refs, outputDirectory)
         assert(IO.read(dir / "a.txt") == "diverged")
 
-  test("Disk cache relinks a digest-matching dirzip to the CAS"):
+  test("Disk cache materializes a digest-matching dirzip from the CAS"):
     withDiskCache: cache =>
       IO.withTemporaryDirectory: tempDir =>
         val outputDirectory = tempDir.toPath()
@@ -176,7 +178,7 @@ object ActionCacheTest extends BasicTestSuite:
         assert(!Files.isSymbolicLink(zipPath), "packageDirectory should leave a regular file")
 
         cache.syncBlobs(refs, outputDirectory)
-        assert(Files.isSymbolicLink(zipPath), "digest-matching archive was not relinked to the CAS")
+        assertMaterialized(cache, zipPath)
 
   test("Disk cache re-extracts a dirzip whose archive digest differs"):
     withDiskCache: cache =>
@@ -873,6 +875,11 @@ object ActionCacheTest extends BasicTestSuite:
       },
       keepDirectory = false
     )
+
+  def assertMaterialized(cache: DiskActionCacheStore, p: Path): Unit =
+    if Util.isApfs(cache.casBase) then
+      assert(!Files.isSymbolicLink(p), s"$p was symlinked instead of copied")
+    else assert(Files.isSymbolicLink(p), s"$p was not symlinked into the CAS")
 
   def getCacheConfig(
       cache: ActionCacheStore,
--- a/internal/util-core/src/main/scala/sbt/internal/util/Util.scala
+++ b/internal/util-core/src/main/scala/sbt/internal/util/Util.scala
@@ -8,12 +8,13 @@
 
 package sbt.internal.util
 
-import java.nio.file.{ Path, Paths }
+import java.nio.file.{ Files, Path, Paths }
 import java.util.Locale
 
 import scala.collection.concurrent.TrieMap
 import scala.reflect.Selectable.reflectiveSelectable
 import scala.util.Properties
+import scala.util.control.NonFatal
 
 object Util:
   def makeList[T](size: Int, value: T): List[T] = List.fill(size)(value)
@@ -76,6 +77,10 @@ object Util:
   lazy val isCygwinWindows: Boolean = isWindows && isCygwin
 
   lazy val isEmacs: Boolean = sys.env.contains("INSIDE_EMACS")
+
+  def isApfs(path: Path): Boolean =
+    try Files.getFileStore(path).`type`().equalsIgnoreCase("apfs")
+    catch case NonFatal(_) => false
 
   def nil[A]: List[A] = List.empty[A]
   def nilSeq[A]: Seq[A] = Seq.empty[A]


util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
```diff
@@ -196,7 +196,10 @@ case class DiskActionCacheStore(base: Path, converter: FileConverter)
     dir
   }
 
-  private val symlinkSupported: AtomicBoolean = AtomicBoolean(true)
+  private val symlinkEnabled: Boolean =
+    System.getProperty("sbt.diskcache.symlink", "true") != "false"
```
stasimus: The parse is exact-match and case sensitive: only the literal lowercase false disables symlinks, so False, FALSE, 0, no and off all leave them enabled. That matches the explicit opt-out intent, but sbt more commonly reads boolean flags via java.lang.Boolean.getBoolean or .toBoolean. Worth being deliberate here, or at least documenting that only lowercase false turns it off.

util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
```diff
@@ -196,7 +196,10 @@ case class DiskActionCacheStore(base: Path, converter: FileConverter)
...
+
+  private val symlinkSupported: AtomicBoolean = AtomicBoolean(symlinkEnabled)
```
stasimus: symlinkSupported is seeded once when the store is constructed, so the flag is effectively fixed for the life of a build session and a mid-session -D change will not take effect. That is the right model for a session lived store, but a short scaladoc on the property (name, default true, the reflink trade off, and that it is read at construction) would help users, since sbt documents most -D flags.

util-cache/src/test/scala/sbt/util/ActionCacheTest.scala
```diff
@@ -781,6 +820,13 @@ object ActionCacheTest extends BasicTestSuite:
       keepDirectory = false
     )
 
+  def withoutSymlinks[A](f: => A): A =
```
stasimus: Mutating the global sys.props is a test isolation smell. It is safe today only because ActionCacheTest is the only suite that constructs DiskActionCacheStore and verify runs a suite sequentially. If another suite ever constructs a store while this property is flipped, it could observe symlink=false mid flight and fail an isSymbolicLink assertion. A one line comment noting the ordering assumption (the property must be set before the store is constructed, which wrapping withDiskCache guarantees) would help future maintainers.

util-cache/src/test/scala/sbt/util/ActionCacheTest.scala
```diff
@@ -781,6 +820,13 @@ object ActionCacheTest extends BasicTestSuite:
...
+    val key = "sbt.diskcache.symlink"
+    val saved = sys.props.get(key)
+    sys.props(key) = "false"
+    try f
+    finally sys.props(key) = saved.getOrElse("true")
```
stasimus: This does not cleanly restore when the property was unset. In CI saved is usually None, so this sets the property to the literal string true instead of removing it. It is harmless today because true is the default, but it is not a true restore and would mask a future default change. Prefer: saved match { case Some(v) => sys.props(key) = v; case None => sys.props -= key }.

util-cache/src/main/scala/sbt/util/ActionCacheStore.scala
hoangmaihuy: I've move the flag to `SysProps` and use `getOrTrue` method there

util-cache/src/test/scala/sbt/util/ActionCacheTest.scala
hoangmaihuy: I've created `withDiskCacheNoSymlinks` version in test file

main/src/main/scala/sbt/RemoteCache.scala
```diff
@@ -39,9 +39,9 @@ object RemoteCache:
     },
     cacheStores := {
       val c = fileConverter.value
-      List(
-        DiskActionCacheStore(localCacheDirectory.value.toPath, c)
-      )
+      val store = DiskActionCacheStore(localCacheDirectory.value.toPath, c)
```
hoangmaihuy: I would rather add a `useSymlink: Boolean` in `DiskActionCacheStore` constructor instead of doing this, but it will break MiMa because `DiskActionCacheStore` is public

christianharrington:
Hi! This looks interesting. Thanks for working on this ☺️

I was wondering about two things:
1. Would it make sense to default to hard links, rather than symlinks? Only falling back to symlinks if the cache is on a different FS or if hard links aren't supported. Having an option to opt-in to copying the files would still make sense, I think. 
2. If the above makes sense, then a name like `disableSymlinks` doesn't make as much sense anymore. In any case, maybe it's clearer if the option is named by what it enables, rather than what it disables? E.g. `enableFileCopy`, or something. That leaves room for changing the default behavior as well.

eed3si9n:
Apparently we can detect which filesystem we're on like this:

```scala
sbt:aaa> console
Welcome to Scala 3.9.0-RC6 (17.0.12, Java OpenJDK 64-Bit Server VM).

scala> import java.nio.file.{ Files, Paths }

scala> Files.getFileStore(Paths.get(".")).`type`
val res0: String = "apfs"
```

If creating symlink is bad on APFS, I'd much rather fix the out-of-box behavior rather than giving users configuration that they need to learn.

christianharrington:
Could a simple solution be to try to create a hard link by default, falling back to a symlink if it isn't possible, regardless of filesystem?

I would presume that would provide the same speedup when reading as copying the files, without taking up more space. 

It would also preserve the semantics (for better or for worse) that writing to a linked file would update the store.

Or maybe there's a specific reason symlinks are preferable?

eed3si9n:
@christianharrington 

> Could a simple solution be to try to create a hard link by default, falling back to a symlink if it isn't possible, regardless of filesystem?

I don't think this would work for sbt's disk cache. Let's say a cached task in a plugin creates `target/out/u/foo/foo.jar`, and sbt creates a hardlink to `$HOME/.cache/cas/12345`. Next, the plugin runs again and overwrites `target/out/u/foo/foo.jar`, hardlink would also write to `$HOME/.cache/cas/12345`, correct? Symlink currently works by copying the file to `$HOME/.cache/cas/12345` and making a symlink `target/out/u/foo/foo.jar`, so if it gets overwritten, CAS remains intact.

hoangmaihuy:
> Apparently we can detect which filesystem we're on like this:
> 
> ```scala
> sbt:aaa> console
> Welcome to Scala 3.9.0-RC6 (17.0.12, Java OpenJDK 64-Bit Server VM).
> 
> scala> import java.nio.file.{ Files, Paths }
> 
> scala> Files.getFileStore(Paths.get(".")).`type`
> val res0: String = "apfs"
> ```
> 
> If creating symlink is bad on APFS, I'd much rather fix the out-of-box behavior rather than giving users configuration that they need to learn.

The reason I went with the opt-in flag approach is because I was afraid to break current caching behavior and users would not be aware of this change. Flag approach can also be backported to 2.0.x series without depending on 2.1.x releases, so users can try this fix earlier and we can make changes to disk cache behavior later in 2.1.x releases.

If you're okay with file-system-aware approach right now, I would also be glad to do it in this PR, please give me some advices, thank you a lot 😄

eed3si9n:
> Flag approach can also be backported to 2.0.x series without depending on 2.1.x releases, so users can try this fix earlier and we can make changes to disk cache behavior later in 2.1.x releases.

Generally my preference would be land all the performance improvements on `develop` branch, and we can start releasing 2.1.0-Mx milestones each week. I already configured nightly build so if we want to test https://github.com/sbt/sbt/commit/46eca942a15bb2d10ecc41bc0c47401b96b4c291 for example, 

```property
sbt.version=2.1.0-bin-20260822-46eca94-NIGHTLY
```

can be tested already.

hoangmaihuy:
@eed3si9n I have one more question about the global cache, do you have any good way to clean up stale files? 
With copy approach I can tell my folks to just delete global cache folder. But I don't want to tell Linux folks that they have to pay a full recompilation while Apple folks don't because sbt cache behaves differently.

eed3si9n:
Currently the only support we provide is cleanFull for wiping out the diskcache. If you want something like LRU cache, I suggest using a Bazel remote cache.

eed3si9n (approved):
Thanks!

PR #9624, opened 2026-08-18
[2.0.x] fix: Avoid rewriting unchanged plugin descriptors (#9612)

unkarjedy:
## Summary

SBT 2 port of #9613 for #9612.

There is no substantive production-code difference from the SBT 1.x fix: the generated plugin descriptor is written only when its normalized contents change. The SBT 2 test code differs where its artifact APIs use `VirtualFileRef` values and `fileConverter`, and the timestamp-observation tasks are explicitly uncached.

SBT 2's normal `packageBin` path already produces cache hits. The assembly scenario additionally covers the physical assembly JAR so the end-to-end behavior is protected.

## Validation and pending work

- `CI=true ./sbt --client scalafmtCheckAll` passes.
- The descriptor and `packageBin` coverage passes.
- `scripted plugins/sbt-assembly` currently **fails** with published `sbt-assembly` 2.4.1 because its unchanged assembly JAR is still rewritten at `checkAssemblyMtime`.

This draft intentionally waits for [sbt-assembly#580](https://github.com/sbt/sbt-assembly/issues/580). With the local implementation of that fix published as `2.4.1-SNAPSHOT`, the complete scripted scenario passes: unchanged descriptor and assembly output retain their timestamps, while an added `AutoPlugin` invalidates both.

### DIFF
--- a/main/src/main/scala/sbt/internal/PluginDiscovery.scala
+++ b/main/src/main/scala/sbt/internal/PluginDiscovery.scala
@@ -87,7 +87,10 @@ object PluginDiscovery:
       IO.delete(descriptor)
       None
     } else {
-      IO.writeLines(descriptor, names.distinct.sorted)
+      val lines = names.distinct.sorted
+      // Do not invalidate timestamp-based downstream caches when the descriptor content is unchanged.
+      if (!descriptor.exists || IO.readLines(descriptor) != lines)
+        IO.writeLines(descriptor, lines)
       Some(descriptor)
     }
   }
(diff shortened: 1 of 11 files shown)


unkarjedy:
> - scripted plugins/sbt-assembly currently fails with published sbt-assembly 2.4.1 because its unchanged assembly JAR is still rewritten at checkAssemblyMtime.
>
>This draft intentionally waits for https://github.com/sbt/sbt-assembly/issues/580. With the local implementation of that fix published as 2.4.1-SNAPSHOT, the complete scripted scenario passes: unchanged descriptor and assembly output retain their timestamps, while an added AutoPlugin invalidates both.

This depends on https://github.com/sbt/sbt-assembly/pull/581
After merging we need to republish sbt-assembly and update the version in this PR to 2.4.2

eed3si9n (approved):
Thanks!

Issue #8175, opened 2025-07-23
Compilation fails because inc_compile_3.zip files are still being used


IFTA-KIM:
## steps

<details>
<summary>build.sbt (significantly minimized, still produces the issue on my machine)</summary>

```
import Dependencies.*

name := "datahub.build"

lazy val scala3CompilerSettings = Seq(
  scalaVersion := "3.3.3",
  scalacOptions ++= Seq("-release", "17"),
  scalacOptions ++= Seq("-encoding", "utf-8"),
)

lazy val allScalaVersionsSettings = Seq(
  organization := "com.ifta.datahub",
  javacOptions ++= Seq("--release", "17"),
  resolvers    := iftaNexus +: resolvers.value,
)

lazy val commonSettings = allScalaVersionsSettings ++ scala3CompilerSettings

lazy val dataHubBuild = (project in file(".")).settings(commonSettings).aggregate(dataHubTestLib)

lazy val dataHubLib = (project in file("lib/com.ifta.datahub.lib"))
  .settings(commonSettings)
  .settings(
    name := "lib",
    libraryDependencies ++= Seq(
      catsCore,
      iftaLibCore,
      scalaLib,
      typesafeConfig,
      logBack,
      osLib,
      slf4jApi,
      scaffeine,
      scalaXml,
    ),
    libraryDependencies ++= scodecDeps(scalaVersion.value),
    libraryDependencies ++= testDeps,
  )

lazy val dataHubTestLib =
  (project in file("lib/test/com.ifta.datahub.test.lib"))
    .settings(commonSettings)
    .dependsOn(dataHubLib)
    .settings(
      name := "test.lib",
      libraryDependencies ++= Seq(betterFiles, iftaLibCore, scalaTest),
      libraryDependencies ++= testDeps,
    )
```

</details>

Please note that I could not universally reproduce the issue with any build.sbt - so far I've only managed to get this error in one specific project with the minimized build.sbt above.

1. Start/reboot sbt
1. Run `Test/compile;clean;Test/compile`
-> Build fails in second `Test/compile`

In sbt v1.10.9:
```
[error] java.nio.file.FileSystemException: [...]\lib\test\com.ifta.datahub.test.lib\target\scala-3.3.3\zinc\inc_compile_3.zip: Der Prozess kann nicht auf die Datei zugreifen, da sie von einem anderen Prozess verwendet wird
[error]         at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:92)
[...]

PR #9627, opened 2026-08-19
[2.x] fix: Handle locked output files on Windows

ColOfAbRiX:
## Problem

On Windows, `Files.move` cannot replace a file that another process holds open.
When sbt runs on Windows (or under WSL against an NTFS `/mnt/c` mount), output files such as packaged jars and `classes.sbtdir.zip` can be held open by the build's classloader and the action cache across command evaluations. As a result, the atomic rename that stages a freshly built artifact over the existing one fails with `AccessDeniedException`, aborting the build.

This surfaces in two places:
- `installPackagedZip` in the action cache, which stages a temp file and moves
  it over the destination jar/zip.
- `IO.writeFileAtomically` (via `IO.jar`), used when packaging meta-build jars.

## Solution

- `ActionCache.installPackagedZip`: skip the install entirely when the destination already has identical content (digest comparison), avoiding the locked move on cache hits. Restore the missing `Files.copy` of the built zip into the staging file.
- `BuildDef` (`CachedAnalysisStore.get`, `getContents`): catch `FileSystemException` when reading output file attributes and fall back to recomputing instead of failing.
- `CacheImplicits` (`hashedVirtualFileRefToStr`, `virtualFileRefToDigest`): catch `FileSystemException` from `Files.readAttributes` and fall back to the underlying value.
- `ActionCacheStore.syncFile`: replace delete-then-write with a temp

## Related to

* https://github.com/sbt/sbt/issues/9651
* https://github.com/sbt/sbt/issues/9653

## Test

Reproduced the `AccessDeniedException` under Windows by running sequential `compile` invocations (A → B → A) against a project with a meta-build, and verified the patched sbt completes the sequence without the lock error.

I wasn't able to write a test in code to replicate the issue

## Caveat

The same problem happens in the metabuild but I decided to keep this PR confined to the project scala code

**This PR was created with the help of AI, I'm doing my best to understand the change but I lack the wider context of the project. Any feedback is welcome**


eed3si9n:
Is there a GitHub issue on this problem? Does https://github.com/sbt/sbt/issues/9537 describe the issue?

ColOfAbRiX:
> Is there a GitHub issue on this problem?

No, but I can create it

> Does #9537 describe the issue?

Yes, partly. I encountered 3 distinct exceptions related to locked files.

The first exception seems to be similar to https://github.com/sbt/sbt/issues/9537 but not quite:
```
[error] (beesightDetection / Test / packageBin) java.nio.file.AccessDeniedException: C:\Users\Fabrizio\projects\beesight\target\out\jvm\scala-3.8.4\beesight-detection\beesight-detection_3-1.0.0-tests.jar.a9e65326.tmp -> C:\Users\Fabrizio\projects\beesight\target\out\jvm\scala-3.8.4\beesight-detection\beesight-detection_3-1.0.0-tests.jar
[error] elapsed time: 4 s, cache 85%, 29 disk cache hits, 5 onsite tasks, 1 error
```

Then there is one related to `inc_compile_3.zip` and is similar to https://github.com/sbt/sbt/issues/8175
```
java.nio.file.FileSystemException: C:\Users\<user>\...\target\zinc\inc_compile_3.zip: The process cannot access the file because it is being used by another process.
    ...
    at java.nio.file.Files.readAttributes(Files.java:...)
    at sbt.internal.BuildDef$CachedAnalysisStore.get(BuildDef.scala:...)
```

The last one is on the meta build
```
java.nio.file.AccessDeniedException: C:\Users\<user>\...\.sbt\1\...\beesight-build-build-build_sbt2_3-0.1.0-SNAPSHOT.jar
    ...
    at sbt.io.IO$.writeFileAtomically(IO.scala:...)
    at sbt.io.IO$.archive(IO.scala:...)
    at sbt.io.IO$.jar(IO.scala:...)
    at sbt.Pkg$.makeJar(Pkg.scala:...)
```
And this PR should fix all 3

eed3si9n:
> And this PR should fix all 3

For the sake of record keeping etc, let's create 3 separate issues, and 3 separate pull requests.

util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -351,30 +352,39 @@ object ActionCache:
...
+                destZip,
+                StandardCopyOption.REPLACE_EXISTING,
+                StandardCopyOption.ATOMIC_MOVE,
+              )
+            catch
+              case _: AtomicMoveNotSupportedException =>
+                Files.move(staging, destZip, StandardCopyOption.REPLACE_EXISTING)
+          },
+          classOf[NoSuchFileException],
         )
```
eed3si9n: As I mentioned in the other pull request, I think we should change this to using `IO.copyFile`, and if needed harden the implementation of it rather than applying hardening where this pattern happens.

eed3si9n (changes requested):

util-cache/src/main/scala/sbt/util/ActionCache.scala
ColOfAbRiX: Agreed

ColOfAbRiX:
Hopefully fixed by https://github.com/sbt/sbt/pull/9731

PR #9632, opened 2026-08-19
fix: include plugin classpath in action cache keys

### LINKED ISSUE #9631: Plugin classpath not in action cache key: stale scalacOptions and replayed CachedCompileFailure
Changing an sbt plugin version that contributes `scalacOptions` does not invalidate the disk (action) cache in sbt 2.0.6. Three consequences:

1. `compile` replays a stale cached result instead of recompiling after the plugin change;
2. once a real recompile is forced (`cleanFull`), reverting the plugin makes the next compile run with the previous plugin's `scalacOptions`, fail, and record the failure under the current cache key;
3. every subsequent `compile` then replays `sbt.util.CachedCompileFailure` (~1s, no compiler run) although the configuration compiles fine. `clean` does not purge it, only `cleanFull` does.

```
[error] sbt.util.CachedCompileFailure$$anon$1: Compilation failed
[error]         at sbt.util.CachedCompileFailure.toException(CachedCompileFailure.scala:23)
[error] elapsed time: 1 s, cache 100%, 1 cached-failure cache hit, ...
```

With the flag-adding plugin version on the classpath (verified via `eval ...getCodeSource.getLocation`), the plugin's own logic returns `-Wnonunit-statement`, yet `show compile/scalacOptions` prints the stale value without it.

Minimal runnable reproduction with exact commands and expected/actual exit codes: https://github.com/stasimus/sbt2-stale-cache-repro

Environment: sbt 2.0.6, Scala 2.13.18, Temurin 25, reproduced on Linux x86_64 and macOS.

stasimus:
Fixes #9631. Cached task keys ignored the metabuild classpath, so plugin version changes replayed stale values and poisoned CachedCompileFailure entries until cleanFull. Adds a metabuild classpath digest to every key and keys cached failures on the full build definition. Includes scripted regression test.

### DIFF
--- a/main-command/src/main/scala/sbt/BasicKeys.scala
+++ b/main-command/src/main/scala/sbt/BasicKeys.scala
@@ -135,6 +135,18 @@ object BasicKeys {
     10000
   )
 
+  private[sbt] val metaBuildClasspathDigest = AttributeKey[String](
+    "metaBuildClasspathDigest",
+    "Digest of the metabuild classpath (plugin jars and compiled build definition products), incorporated into every action cache key.",
+    10000
+  )
+
+  private[sbt] val buildDefinitionDigest = AttributeKey[String](
+    "buildDefinitionDigest",
+    "Digest of the full build definition (metabuild classpath and compiled .sbt file classes), incorporated into cached failure keys.",
+    10000
+  )
+
   // Unlike other BasicKeys, this is not used directly as a setting key,
   // and severLog / logLevel is used instead.
   private[sbt] val serverLogLevel =
--- a/main-settings/src/main/scala/sbt/Def.scala
+++ b/main-settings/src/main/scala/sbt/Def.scala
@@ -19,6 +19,7 @@ import sbt.internal.util.{ Terminal as ITerminal, * }
 import sbt.util.{
   AggregateActionCacheStore,
   BuildWideCacheConfiguration,
+  Digest,
   DiskActionCacheStore,
   Uncached,
 }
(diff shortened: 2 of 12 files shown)


stasimus:
**What was broken**

The action cache key for a cached task is `sha256(codeContentHash, hash(inputs), extraHash, cacheVersion)`, where `codeContentHash` is the hash of the task body's macro tree (Cont.scala) and `extraHash` is always 0. Nothing in the key reflects the metabuild classpath. A plugin like sbt-scalac-opts-plugin defines `scalacOptions ++= scalacOptsFor(scalaVersion.value, scalacOptsAll)` — the tree is identical across plugin versions, only the data behind it changes. So after a version bump:

1. `scalacOptions` (and `compile`) replayed stale cached values;
2. after `cleanFull`, a recompile with the stale options failed and the `CachedCompileFailure` was recorded under a key that ignores the plugin classpath;
3. every subsequent `compile` replayed that failure for a configuration that compiles fine; only `cleanFull` recovered.

**How it's fixed**

- On each project load, two digests are computed from already-available content hashes and passed via `BuildWideCacheConfiguration`:
  - `metaBuildDigest`: every build unit's `pluginData.classpath` (plugin/dependency jars + compiled `project/*.scala` products). Mixed into **every** cached task key, so any plugin or build-helper change invalidates dependent cached values.
  - `buildDefinitionDigest`: the above plus the Eval-generated `.sbt` classes. Used **only** to key cached failures: a failure is stored under `sha256(inputDigest, buildDefinitionDigest)` and can only replay while the whole build definition is byte-identical. Failure entries written by older sbt under the loose key are treated as a miss and overwritten.
- The `.sbt` dsl classes are deliberately excluded from the global key so a plain `build.sbt` edit doesn't invalidate the entire cache (guarded by the existing `cache/version-bump-compile` scripted test).

Verified with the repro from https://github.com/stasimus/sbt2-stale-cache-repro: all README steps produce the expected exit codes with this patch. New scripted test `cache/plugin-scalac-options-toggle` covers the toggle scenario; `utilCache/test`, `scripted cache/*`, and MiMa all pass.

PR #9634, opened 2026-08-19
fix: keep in-use boot jars on cleanFull

### LINKED ISSUE #9633: cleanFull breaks the running session by deleting in-use boot jars
## steps

sbt 2.x (since #9060; not reproducible on 2.0.6), single server session:

```sh
sbt --batch compile          # ok
sbt --batch cleanFull        # deletes ~/.sbt/boot entirely
echo '// comment' >> project/plugins.sbt
sbt --batch compile          # exit 1
```

## problem

The last compile fails during reload with only:

```
[error] elapsed time: 0 s
```

The real error is swallowed into `target/global-logging`:

```
dotty.tools.dotc.MissingCoreLibraryException: Could not find package scala from compiler core libraries.
  at sbt.internal.Eval.compileAndLoad(Eval.scala:268)
  ...
```

Only restarting sbt recovers. This also masks other failures: any by-path access to boot jars (action cache input hashing, scalaCompilerBridgeBin as reported in the #9060 comments) breaks after cleanFull.

## cause

#9060 made `cleanFull` run `IO.delete(launcher.bootDirectory())` in-process. The running server is loaded from those jars; it survives on open file handles, but the next Eval compile passes the deleted `~/.sbt/boot/scala-*/lib/scala-library.jar` on `-classpath` and dotc's `Definitions.init` fails.

Rebooting instead does not work either: on `FullReload(clean = true)` the launcher deletes boot but reuses in-memory providers, so boot is not re-materialized until a fresh JVM.

## expectation

`cleanFull` clears caches without breaking the session it runs in.

stasimus:
Fixes #9633. cleanFull deleted the whole boot directory under the running server, whose next .sbt compilation then failed with MissingCoreLibraryException until restart. Boot cleanup now defers the jars the running instance is loaded from to a shutdown hook, and the server shuts down after the exec finishes. Adds a server test.

### DIFF
--- a/main-command/src/main/scala/sbt/BasicCommandStrings.scala
+++ b/main-command/src/main/scala/sbt/BasicCommandStrings.scala
@@ -258,7 +258,7 @@ $AliasCommand name=
 
   val CleanFull: String = "cleanFull"
   def cleanFullDetailed: String =
-    "Clears sbt's local caches, keeping only the boot artifacts the running sbt instance uses, and shuts the server down once remaining commands finish."
+    "Clears sbt's local caches and shuts the server down once remaining commands finish. Boot artifacts the running sbt instance uses are deleted on exit."
 
   private[sbt] val networkExecPrefix = "__"
   private[sbt] val DisconnectNetworkChannel = s"${networkExecPrefix}disconnectNetworkChannel"
--- a/main-command/src/main/scala/sbt/BasicCommandStrings.scala
+++ b/main-command/src/main/scala/sbt/BasicCommandStrings.scala
@@ -258,7 +258,7 @@ $AliasCommand name=
 
   val CleanFull: String = "cleanFull"
   def cleanFullDetailed: String =
-    "Clears sbt's local caches, keeping only the boot artifacts the running sbt instance uses."
+    "Clears sbt's local caches, keeping only the boot artifacts the running sbt instance uses, and shuts the server down once remaining commands finish."
 
   private[sbt] val networkExecPrefix = "__"
   private[sbt] val DisconnectNetworkChannel = s"${networkExecPrefix}disconnectNetworkChannel"
--- a/main-command/src/main/scala/sbt/BasicCommandStrings.scala
+++ b/main-command/src/main/scala/sbt/BasicCommandStrings.scala
@@ -257,7 +257,8 @@ $AliasCommand name=
   def ClearCachesDetailed: String = "Clears sbt's internal caches."
 
   val CleanFull: String = "cleanFull"
-  def cleanFullDetailed: String = "Clears sbt's local caches."
+  def cleanFullDetailed: String =
+    "Clears sbt's local caches, keeping only the boot artifacts the running sbt instance uses."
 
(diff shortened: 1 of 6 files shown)


anatoliykmetyuk:
This PR conflates two goals I believe:

- `cleanFull` should completely remove all caches within its scope, including the complete boot directory
- The current server should remain usable long enough to run chained commands like `cleanFull; compile`

In its current form, it weakens `cleanFull` by no longer deleting the boot dir.

https://github.com/sbt/sbt/issues/9633 converged on the new contract for `cleanFull` to be treated as terminal. To be consistent with that contract, it would make more sense to not support extra commands chained after `cleanFull`, performing the complete cleanup and shutting down the server at the time that command is executed. We could e.g. require `cleanFull` to be last in case of chaining, emitting an error message otherwise.

The above suggestion does introduce an ergonomics concern with `cleanFull; compile` pattern no longer supported. We could consider preserving the commands after `cleanFull` in the chain and feeding them to be executed immediately upon server restart, if that is trivial to do within current architecture.

stasimus:
Agreed that these are two goals, but I don't think the first one is reachable from inside the exec. The server is running off those jars and still needs them until it exits: Eval compiles, logging, result reporting. And on Windows the JVM holds them open, so deleting them isn't allowed at all, which is also why the `sbt.boot.lock` exclusion is in there. `IO.delete(bootDirectory)` on develop is only "complete" in the sense that it doesn't check what it broke.

So instead of choosing between the two, how about deleting the in-use jars on exit? cleanFull removes everything it can right away, registers the rest via `ShutdownHooks` (same pattern as `Defaults.scala`), and shuts the server down. Boot ends up empty, the running session stays usable, and `cleanFull; compile` still works without needing a "cleanFull must be last" error. Terminal in the sense that matters, the server is gone afterwards.

On replaying commands after restart: the client only starts a server when it first connects, there's no reconnect-and-resubmit path, and in-JVM reboot doesn't re-materialize boot (see the issue). Doable, but a separate client-side change I'd rather not put here.

One thing neither approach solves: `~/.sbt/boot` is shared across builds, so cleanFull can still yank jars from under a server running for a different project. Same on develop today, just want it on the record.

Happy to push the shutdown-hook version if that sounds reasonable.

anatoliykmetyuk:
That sounds reasonable to me. I agree that deleting the in-use boot artifacts through ShutdownHooks gives us the best of both worlds: cleanFull still leaves the boot directory empty after server shutdown, while commands already in the current chain can complete safely. This seems better than requiring cleanFull to be last.

I also agree that reconnecting and resubmitting: if it requires heavy plumbing, probably not worth it.

stasimus:
Pushed the shutdown-hook version. cleanFull now deletes everything it can right away, registers the in-use boot jars for deletion via ShutdownHooks, and shuts the server down after the exec. The server test now also checks the boot directory is empty once the server has exited, and that a fresh server starts fine afterwards.

On Windows the JVM keeps the jars open, so the hook deletion is best effort there and any failure is ignored quietly. Same limitiation develop had.

PR #9639, opened 2026-08-19
[2.x] fix: Keep adhoc test classloader open in a forked test

### LINKED ISSUE #9638: Forked test classloader closed before shutdown hooks run
With `Test / fork := true`, the worker JVM (`sbt.internal.worker1.WorkerMain`) wraps the test classpath in a `URLClassLoader` and closes it via try-with-resources as soon as the run returns — before `System.exit(0)` fires shutdown hooks. Any class not already loaded then fails inside a shutdown hook with `ClassNotFoundException` / `NoClassDefFoundError`.

sbt 1.x ran forked tests on the never-closed launch classloader, so hooks worked. This regression breaks shutdown-hook cleanup such as testcontainers container reaping: containers leak, hold host ports, and subsequent runs fail with "Bind for 0.0.0.0:XXXX failed: port is already allocated".

The `run` task has the same pattern.

Reproduction: https://github.com/stasimus/sbt2-shutdown-hook-repro

- sbt 2.0.6: `sbt testFull` → `hook-result.txt` contains `FAIL: java.lang.ClassNotFoundException: repro.LazilyLoaded$`
- sbt 1.12.14: `OK`
- sbt 2.0.6 with `Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Raw` (classpath goes on the worker's launch command instead of the closeable loader): `OK`

Affected code: `worker/src/main/java/sbt/internal/worker1/WorkerMain.java` — `test` and `run` paths (lines 192-194 and 173-178 on develop).

stasimus:
Fixes #9638. WorkerMain closed the forked run's URLClassLoader before System.exit(0) fired shutdown hooks, so hooks loading new classes failed. The loader is now never closed: the worker exits right after, and closing from a hook would race with user hooks. Adds scripted test.

### DIFF
--- a/sbt-app/src/sbt-test/tests/fork-shutdown-hook/build.sbt
+++ b/sbt-app/src/sbt-test/tests/fork-shutdown-hook/build.sbt
@@ -1,0 +1,21 @@
+Global / localCacheDirectory := baseDirectory.value / "diskcache"
+scalaVersion := "3.8.4"
+
+Test / fork := true
+
+libraryDependencies += "org.scalameta" %% "munit" % "1.0.4" % Test
+
+val check = TaskKey[Unit]("check", "Verify the shutdown hook could load classes.")
+
+check := Def.uncached {
+  val file = baseDirectory.value / "hook-result.txt"
+  val deadline = System.currentTimeMillis + 30000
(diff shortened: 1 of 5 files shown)


eed3si9n (approved):
Thanks!

PR #9640, opened 2026-08-19
[2.x] test: Add action cache path and content-verification regression tests

stasimus:
Adds regression tests for existing DiskActionCacheStore guards: refusing cached refs whose relative or absolute path escapes the output directory, and not restoring a blob whose bytes do not match its content digest. Test-only, no behavior change.


eed3si9n (approved):
Thanks!

PR #9641, opened 2026-08-19
[2.x] Hash action cache inputs with full-width sha256

stasimus:
mkInput folded only a 32-bit murmur hash of the task input into the cache key, so distinct inputs collide at the birthday bound (~31 per 500k realistic inputs), silently resolving a task to the wrong cached output. A new DigestHasher hashes inputs into a full-width sha256 Merkle digest. Changes all cache keys, so the cache is repopulated once. Adds a regression test.

### DIFF
--- a/util-cache/src/main/scala/sbt/util/ActionCache.scala
+++ b/util-cache/src/main/scala/sbt/util/ActionCache.scala
@@ -36,7 +36,6 @@ import scala.annotation.{ meta, tailrec, StaticAnnotation }
 import scala.collection.mutable
 import scala.util.control.NonFatal
 import sjsonnew.{ HashWriter, JsonFormat }
-import sjsonnew.support.murmurhash.Hasher
 import sjsonnew.support.scalajson.unsafe.{ CompactPrinter, Converter, Parser, PrettyPrinter }
 import scala.quoted.{ Expr, FromExpr, ToExpr, Quotes }
 import xsbti.{ CompileFailed, FileConverter, HashedVirtualFileRef, VirtualFile, VirtualFileRef }
@@ -309,8 +308,8 @@ object ActionCache:
   ): Digest =
     // Hashing serializes every task input; surface a missing input file directly rather than as an
     // opaque serialization failure that buries it.
-    val inputHash =
-      try Hasher.hashUnsafe[I](key)
+    val inputDigest =
+      try DigestHasher.hashUnsafe[I](key)
       catch
         case NonFatal(t) =>
           findMissingFile(t) match
@@ -320,7 +319,7 @@ object ActionCache:
     Digest.sha256Hash(
       (Vector(
         codeContentHash,
-        Digest.dummy(inputHash),
+        inputDigest,
         extraHash
       ) ++ {
         if cacheVersion == 0 then Vector.empty
--- a/util-cache/src/main/scala/sbt/util/DigestHasher.scala
+++ b/util-cache/src/main/scala/sbt/util/DigestHasher.scala
@@ -1,0 +1,55 @@
+/*
+ * sbt
+ * Copyright 2023, Scala center
+ * Copyright 2011 - 2022, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt.util
+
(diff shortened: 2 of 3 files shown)


util-cache/src/main/scala/sbt/util/ActionCache.scala
```diff
@@ -309,8 +308,8 @@ object ActionCache:
   ): Digest =
     // Hashing serializes every task input; surface a missing input file directly rather than as an
     // opaque serialization failure that buries it.
-    val inputHash =
-      try Hasher.hashUnsafe[I](key)
+    val inputDigest =
+      try DigestHasher.hashUnsafe[I](key)
```
eed3si9n, line 312: Thanks for looking into this! Would 64-bit be enough? If so we can use faster hash like XXHash64. From https://eed3si9n.com/hashing-bakeoff/,

- MurmurHash32 (Scala.util) `1.42 GB/s`
- Sha256 `0.205 GB/s` 7x slower
- XXHash64 `8.83 GB/s` 6x faster

stasimus: For accidental collisions 64 bits is plenty, roughly 1e-8 odds at 500k inputs. I went with sha256 because XXHash64 isn't collision resistant, so on a shared remote cache anyone who can influence a task input could craft a collision against another task's key. And inputs are small, so per-call overhead dominates over GB/s anyway. But if adversarial collisions are out of scope, happy to switch.

stasimus: Went with sha256 since Digest is already in util-cache. XXHash64 needs a new SupportHasher in sjson-new, probably better as a separate PR.

stasimus: Went with sha256 since Digest is already in util-cache. XXHash64 needs a new SupportHasher in sjson-new, probably better as a separate PR.

eed3si9n: `Xx64` is there, but I guess I didn't implement the variant that hashes bytes yet - https://github.com/sbt/sbt/blob/94acc76830a3138e29310b8d131016b869e0dc81/util-cache/src/main/scala/sbt/util/Digest.scala#L97

eed3si9n: The CAS entry itself would use sha256 a few lines below, which hopefully defends against collision attack.

stasimus: As I see - the outer sha256 doesn't help there: if two inputs collide in XXHash64, the bytes fed into it are identical, so the final key collides.

The final action key is sha256(codeContentHash ++ inputDigest ++ extraHash ++ ...)

eed3si9n: Now that you implemented `DigestHasher`, I wonder if we can combine these two hashing into one. Then we have no additional performance cost. We just need JsonFormat for tuple of `(String, A, String, Int)`.

stasimus: I dont think it is good idea: 
the outer sha256 only covers 3-4? fixed-size digests, ~130 bytes, so there's almost nothing. The expensive part is the Merkle walk over the input, and that runs once either way. A (String, A, String, Int) tuple would go through jstring/jarray, which is a few extra sha256 calls, and tijs the key to sjson-new's tuple encoding?

eed3si9n: ok, it's starting to come back to me. Since sjson-new hasher would call `tagged` per each JSON element, we're talking about hundreds or thousands of `sha256` invocation then? Would this make a difference in no-op compilation scenario etc?

stasimus: Hm, yes, it's hundreds to thousands of sha256 calls per task, but each costs **ns** (I tried to measure on my dev container 100 - 150 ns, so a no-op build gains ms total, which is noise comparing to the per-task output file hashing. I would say impact. 

but, hashing raw bytes internally instead of hex-string Digests would halve even that (I mean to use  facade's type parameter Array[Byte] instead of Digest)

eed3si9n (approved):
Thanks!

PR #9643, opened 2026-08-20
[2.x] fix: Verify remote cache blobs against their digest

### LINKED ISSUE #9644: Remote cache blobs are not verified against their digest on download
### Problem

When sbt pulls blobs from a gRPC remote cache, the downloaded bytes are written to the local CAS and synced into the build output tree without ever being re-hashed against the requested digest.

In `syncBlobs` (`GrpcActionCacheStore.scala`), the batch-read response is keyed on the digest the server reports and the bytes are handed straight to `disk.putBlob(...)` / `disk.putBlobInternal(...)`. Those writers (`ActionCacheStore.scala`) only run an `isCompleteBlob` pre-check, which returns `false` for a not-yet-present file, so on first write the content is trusted with no `sha256(bytes) == digest` verification.

Content addressing is supposed to make tampering self-detecting, but the download path skips that check. The local read path re-verifies, so the exposure window is first use, by which point the poisoned bytes are already in the build.

### Impact

A compromised/malicious remote cache server, or a MITM on a plaintext `grpc://` channel, can return arbitrary bytes under a requested digest. They land in the CAS under the correct-digest name and are synced into build outputs (`.class` files, jars, resources) that developers and CI then execute. Requires a pre-existing trusted position (cache write access or network MITM), so this is a hardening/correctness fix rather than a remotely exploitable vulnerability.

### Fix

Re-hash remote blobs on write and reject on mismatch; warn when a plaintext `grpc://` endpoint is configured. Addressed by #9643.

stasimus:
Fixes #9644.

Blobs pulled from a gRPC remote cache were written to the CAS and synced into the build without re-hashing, so a compromised cache server or plaintext MITM could substitute bytes under the requested digest. Re-hash remote blobs on write and reject on mismatch, and warn when a plaintext grpc:// endpoint is configured.

### DIFF
--- a/sbt-remote-cache/src/main/scala/sbt/plugins/RemoteCachePlugin.scala
+++ b/sbt-remote-cache/src/main/scala/sbt/plugins/RemoteCachePlugin.scala
@@ -19,6 +19,13 @@ object RemoteCachePlugin extends AutoPlugin:
           } match
             case Some(x) => x
             case None    => sys.error("disk store not found")
+          if remote.getScheme == "grpc" then
+            val headers = remoteCacheHeaders.value
+            val creds = if headers.nonEmpty then ", including credential headers," else ""
+            sLog.value.warn(
+              s"remoteCache $remote uses the plaintext grpc:// scheme; traffic$creds " +
+                "is not encrypted and can be read or altered in transit. Use grpcs:// for TLS."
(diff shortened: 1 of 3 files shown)


eed3si9n (approved):
Thanks!

PR #9646, opened 2026-08-20
[2.x] fix: Fixes filesystem traversal order affecting cache stability

### LINKED ISSUE #9645: Compile cache key depends on the filesystem traversal order of sources
(Disclaimer: Most of this was prepared by Claude, with me verifying)

## steps

Reproduced on develop @ 1fbdde292 (2.1.0-SNAPSHOT, built from source with publishLocalBin), JDK 17.0.19 (Azul Zulu), macOS 26.6 / APFS. Also reproduces identically on released sbt 2.0.6.

Create a project with a handful of source files spread over a few packages:

```bash
mkdir -p order-repro/project && cd order-repro
echo 'sbt.version=2.1.0-SNAPSHOT' > project/build.properties

for pkg in alpha lib ci; do
  mkdir -p "src/main/scala/com/example/$pkg"
  for n in 1 2 3 4 5 6; do
    printf 'package com.example.%s\nclass C%s%s\n' "$pkg" "$pkg" "$n" \
      > "src/main/scala/com/example/$pkg/C$pkg$n.scala"
  done
done
```

build.sbt:
```scala
scalaVersion := "3.8.4"

Global / localCacheDirectory := baseDirectory.value / "diskcache"

lazy val showSourceOrder = taskKey[Unit]("prints the order sources are hashed in")

// Def.uncached matters: without it the task itself is cached and prints nothing on a re-run.
showSourceOrder := Def.uncached {
  val log = streams.value.log
  (Compile / sourcesVF).value.foreach(vf => log.info(vf.id))
}

// Reversing the list stands in for a second filesystem.
Compile / sources := {
  val ss = (Compile / sources).value
  if ((baseDirectory.value / "reverse.marker").exists) ss.reverse else ss
}
```

```
sbt shutdown
rm -rf diskcache target reverse.marker

# run 1: natural order, cold cache
sbt -Dsbt.experimental_execution_log=$PWD/exec-1.json --batch ";compile;shutdown"

# run 2: same sources, reversed order, classes deleted
rm -rf target/out/jvm/scala-3.8.4/order-repro/classes
touch reverse.marker
sbt -Dsbt.experimental_execution_log=$PWD/exec-2.json --batch ";compile;shutdown"

# run 3: natural order restored, classes deleted again
rm -f reverse.marker
rm -rf target/out/jvm/scala-3.8.4/order-repro/classes
sbt -Dsbt.experimental_execution_log=$PWD/exec-3.json --batch ";compile;shutdown"
```

The classes path is `target/out/jvm/scala-<v>/<project>/classes` in sbt 2.x.

```bash
for f in exec-1.json exec-2.json exec-3.json; do
  echo "--- $f ---"
  jq -r 'select(.input.str != null
                and (.input.str | contains("CompileInputs2"))
                and (.input.str | contains("-build") | not))
         | [.cacheHit, .input.digest] | @tsv' "$f"
done
```

| run | source order | `cacheHit` | compile input digest (truncated) | console |
|-----|--------------|-----------|----------------------------------|---------|
| 1 | natural | `false` | `sha256-87caaebcfa4feb59…` | `compiling 18 Scala sources`, `cache 11%` |
| 2 | reversed | `false` | `sha256-7fe99112d7ddfccf…` | `compiling 18 Scala sources`, `cache 88%` |
| 3 | natural | `true`  | `sha256-87caaebcfa4feb59…` | no compile, `cache 100%`, 18 classes restored |

Run 2 recompiles from scratch. Run 3, an identical scenario except that the order is the one
run 1 used, restores all 18 class files without compiling and reports run 1's digest. The
sources are byte-identical in all three runs and no other input changed.

The meta-build compile is a clean control: it holds one digest
(`sha256-4a0edda132452981…`) and hits in all three runs, since `project/` is untouched.

The `[success] … cache N%` summary line is enough to see this without `jq`.

## Problem

`CompileInputs2.sources` is a `Vector[HashedVirtualFileRef]` hashed positionally, and it arrives
in raw filesystem traversal order. Nothing between `readdir` and the cache key normalises it.

So a macOS developer and a Linux CI runner compute different digests for an identical tree.
Every module with more than one source file lands in a key space the other machine cannot reach,
which makes remote caching between CI and a developer machine ineffective for reasons unrelated
to the build.

sbt already normalises this for resources, which is what makes the omission look unintentional:
`resourceDigests` sorts its stamps before hashing (`main/src/main/scala/sbt/Defaults.scala:673`,
`.sortBy(_._1.toString())`, added in c13c25fdf). `sources` was left in traversal order.

## Expectation

Two machines checking out the same commit should compute the same compile cache key. Source
order is not meaningful to the key, so it should be normalised before hashing, as
`resourceDigests` does.

### Impact measured on a real build

On a 134-module Scala build, with an identical tree and after eliminating every `scalacOptions`
difference between the two sides, 53 of 55 module compilations a developer looked up still had a
different input digest from CI's. For one module, source order was the *only* difference in the
entire key. Sorting `sourcesVF` in the build (plus pinning `packageTimestamp`, an unrelated
`SOURCE_DATE_EPOCH` issue that is not an sbt bug) restored 134 of 134 module compilations from
CI's entries.

### Suggested fix, and what it measures

Sorting in `sourcesVF` rather than in `inputFileStamps` or at the `CompileInputs2` call sites:

```scala
    sourcesVF := Def.uncached {
      val conv = fileConverter.value
      val vs = sources.value.toVector.map: x =>
        (conv.toVirtualFile(x.toPath()): HashedVirtualFileRef)
      vs.sortBy(_.id)
    },
```

With that applied, `sourcesVF` comes out fully sorted and run 2 of the steps above becomes a hit
on run 1's digest:

| | natural order | reversed order | run 2 result |
|---|---|---|---|
| before | `sha256-87caaebcfa4feb59…` | `sha256-7fe99112d7ddfccf…` | miss, recompiled 18 sources |
| after  | `sha256-5f5975039086b661…` | `sha256-5f5975039086b661…` | hit, `cache 100%`, 18 classes restored |

I am happy to open a PR if you would like this fixed this way.

### A second instance of the same root cause

`packageBin`'s key has the same property, at lower severity. `packageConfigurationTask`
(`Defaults.scala:2046`) builds a `Pkg.Configuration` whose `sources` is an order-hashed `Vector`
(`main-actions/src/main/scala/sbt/Pkg.scala:121`), and `packageBinMappings`
(`Defaults.scala:1822`) comes from `Path.allSubpaths`, which collects into a `LinkedHashSet` in
traversal order (`io/PathMapper.scala:116`, `io/Path.scala:420`).

The jar *bytes* are not affected, because `IO.writeZip` already sorts entries by name
(`io/IO.scala:726`, since sbt/io#279), so downstream modules still hit. The cost is a `packageBin`
miss that rebuilds a byte-identical jar. Adding `.sortBy(_._2)` to the `mappings.value` read in
`packageConfigurationTask` covers `packageBin`, `packageSrc`, `packageDoc` and `packageInternal`
in one place.

christianharrington:
Fixes https://github.com/sbt/sbt/issues/9645

(Disclaimer: prepared with Claude Code (Opus 5), with me verifying — same as the issue.)

`CompileInputs2.sources` and `packageBin / mappings` are hashed positionally but arrive in raw
filesystem traversal order. APFS, ext4 and overlayfs index directory entries by different name
hashes, so a macOS developer and a Linux CI runner hash an identical tree into different cache
keys and neither can use the other's entries. sbt already sorts resource stamps before hashing
in `resourceDigests`, which is what makes the omission look unintentional.

This sorts `sourcesVF` by virtualized id, and `mappings` where `packageConfigurationTask` reads
it. Both are the points the keys are built from, so a later `sources +=` or `mappings +=` cannot
escape the sort. Sorting by id rather than by absolute path stays machine-independent when
source roots differ.

**Worth flagging:** every compile and package key moves once, so the first build after this
lands recompiles and repackages.

It's the test I'm least certain about. Let me know if there's anything I should do differently.

### Measured

Reversing the input list stands in for a second filesystem's readdir order, which is the only
part of this that cannot be reproduced on a single machine.

Compile, with the classes directory deleted between runs:

| | natural order | reversed order | second run |
|---|---|---|---|
| before | `sha256-87caaebc…` | `sha256-7fe99112…` | miss, recompiled 18 sources |
| after | `sha256-5f597503…` | `sha256-5f597503…` | hit, `cache 100%`, 18 classes restored |

`packageBin`, with the jar deleted between runs:

| | natural order | reversed order | jar bytes |
|---|---|---|---|
| before | `sha256-3bd96dcf…` | `sha256-5755481c…` → miss | identical |
| after | `sha256-d72cb9c8…` | `sha256-d72cb9c8…` → hit | identical |

The jar column is why the `packageBin` half is a cache-efficiency fix rather than a correctness
one: `IO.writeZip` already sorts entries by name (sbt/io#279), so without the sort the key moves
and the rebuilt jar is byte-for-byte the one the cache already held.

### Tests

`sbt-app/src/sbt-test/cache/input-order` reverses both input lists behind a marker file and
asserts that neither reversal costs a cache hit and that the jar stays byte-identical. Both
assertions are important, checked against three builds:

| build | result |
|---|---|
| no fix | fails at the compile check, `missCount=1` |
| `sourcesVF` sort only | compile check passes, fails at the `packageBin` check, `missCount=2` |
| both sorts | passes, `missCount=0` at both |


eed3si9n (approved):
Thanks!

PR #9649, opened 2026-08-20
[2.x] perf: Cut redundant hashing in disk action cache

stasimus:
Follow-up to #9643. The verification added there re-read every downloaded blob right after writing it, and it sat on top of an older cost: DiskActionCacheStore fully re-hashes CAS files on every findBlobs/syncBlobs/getBlobs call, even though the file name is the digest. On a cache hit that meant hashing a whole classpath worth of blobs just to conclude nothing changed.

- Remote blobs (grpc downloads and inline data) are hashed while the stream is written to a staged temp file, then moved into the CAS, so verification costs no extra I/O and tampered bytes never land under the CAS name. Local puts are copied without hashing and are not stamped, so a mismatched local file is still caught by the lookup hash.
- Verified entries are remembered per process by (digest, size, mtime, fileKey), so repeated lookups are a stat instead of a full hash. Any external change to a CAS file invalidates the stamp and forces a re-hash. This is the same trade-off CacheImplicits already makes for file stamps.
- putBlobs/getBlobs/syncBlobs/findBlobs run in parallel on a dedicated I/O pool, with same-digest blobs deduped within a batch, matching what GrpcActionCacheStore already does for uploads and downloads.
- Staging recreates the CAS directory if it was deleted mid-session (the remote-cache/basic scripted test deletes it), and unreadable CAS entries degrade to a per-entry miss instead of failing the whole batch.

Benchmark with 100 x 1MiB blobs (steady state, Linux):

| phase | before | after |
|---|---|---|
| putBlobs (path) | 26 ms | 16 ms |
| findBlobs, repeated x3 | 147 ms | 3 ms |
| syncBlobs, outputs up to date | 105 ms | 8 ms |
| syncBlobs, materialize | 50 ms | 4 ms |
| putBlob (stream + verify) | 81 ms | 70 ms |

The stream verify row barely moves because the sha256 pass has to be paid once for untrusted bytes either way; the gain there is on cold page caches, plus the atomic staging. First lookup in a fresh process still hashes each blob once, the stamp cache is not persisted.


eed3si9n:
Let's step back a bit, and clarify the attack vector. In #9643, you wrote

> Blobs pulled from a gRPC remote cache were written to the CAS and synced into the build without re-hashing, so a compromised cache server or plaintext MITM could substitute bytes under the requested digest.

MITM should be protected using transport security like mTLS, but we can add check for compromised cache server. Other parameters to consider:

- Most casual sbt users would only ever use local disk cache. (although I think remote cache would be very useful for large builds)
- Windows seems to suffer from slower, less reliable, and/or more restrictive IO.

So combined together, I suggest limiting the scope of digest verification to remote cache only, so the local disk IOs are unchanged.

stasimus:
Agreed, scoped it : putBlob(Path, ...) for local files is now a plain copy with no hashing on the write path. The stream and ByteBuffer variants, which is how remote gRPC blobs come in, keep the eager verification and fail on mismatch.                                                                                                                                                                     

a local put isn't stamped complete, so the first lookup still hashes it once, after which it's memoized by (size, mtime, fileKey). So local writes are unchanged and reads pay at most one hash  per blob per process. If you'd rather have local lookups skip even that, I can stamp local puts complete on write instead.

eed3si9n (approved):
Thanks!

PR #8604, opened 2026-01-21
[2.x] feat: Support forked console

calm329:
Fixes #1918

## Summary

Implements `fork in console := true` as requested in #1918.

When enabled, the Scala REPL runs in a separate JVM, which allows:
- Using Java agents (e.g., JAMM for memory profiling)
- Setting custom `java.library.path` for native libraries
- Running JavaFX/ScalaFX apps that can only initialize once per JVM
- Avoiding JLine version conflicts between sbt and the REPL

## Implementation

- Added `ConsoleMain` as the entry point for the forked process
- Used `ProcessBuilder.inheritIO()` to properly inherit terminal streams (critical for arrow key support)
- Fork options like `javaOptions` and `javaHome` are passed through

## Usage

```scala
Compile / console / fork := true
Compile / console / javaOptions += "-javaagent:/path/to/agent.jar"
```

Test plan

- Added scripted test console/fork-console
- Verified compilation succeeds
- Manual testing of arrow keys in forked REPL

PR #9654, opened 2026-08-21
[2.x] feat: Forked compilation (rebased)

stasimus:
Rebase of #8584 (eed3si9n:wip/worker2) onto current develop, keeping the original commits, plus one commit addressing review findings. Unit tests and the fork scripted test pass.

The rebase itself shrank the diff: develop already absorbed the scalaInstance config refactoring via the console PR (#8604), so Compiler.scala falls out entirely, and most conflicts came from the earlier merged worker PR (#8170).

Review fixes:

1. New `forkCompile` setting gates forked compilation; `Test / fork := true` no longer reroutes compile. The scripted test uses it.
2. Worker errors propagate: ProcessReact fails the promise on processing exceptions, WorkerMain unwraps InvocationTargetException, CompileMain catches NonFatal and replies with escaped JSON-RPC errors.
3. The exit listener registers right after proxy creation and self-checks a dead worker, so a startup crash can't hang the compile. Exit notification is synchronized and per-process.
4. The fork path runs under Retry.io and mirrors in-process failure handling. BSP progress and mid-compile ping still don't work forked; that's inherent.
5. The compiler-interface classloader split applies only to compile; run and test load flat. Non-file URIs handled, bogus null guard removed.
6. WorkerProxy closes in finally, fixing the fd leak. Worker reuse left for later.
7. compileOrder travels through CompileConfig. incOptions and sourcePositionMappers still don't; functions can't cross a process boundary. No-op extraIncOptions stub removed.
8. currentClasspath falls back to java.class.path instead of MatchError.
9. The compile task splits into two cachedTasks picked by taskDyn, so the non-fork path pays nothing extra. Dead `jh` and redundant analysis rewrite dropped.
10. -Werror is back on for actionsProj with its hidden warnings fixed; the unused Compiler.* mima exclude is gone.

Same branch is also proposed into eed3si9n:wip/worker2 as eed3si9n/sbt#6, so #8584 can pick up the rebase and fixes directly if preferred.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -591,6 +591,7 @@ lazy val actionsProj = (project in file("main-actions"))
     mimaSettings,
     mimaBinaryIssueFilters ++= Vector(
     ),
+    Compile / scalacOptions --= Seq("-Werror"),
   )
   .dependsOn(lmCore)
   .configure(
@@ -603,7 +604,7 @@ lazy val actionsProj = (project in file("main-actions"))
 
 lazy val protocolProj = (project in file("protocol"))
   .enablePlugins(ContrabandPlugin, JsonCodecPlugin)
-  .dependsOn(collectionProj, utilLogging)
+  .dependsOn(collectionProj, utilLogging, utilCache)
   .settings(
     testedBaseSettings,
     name := "Protocol",
--- a/build.sbt
+++ b/build.sbt
@@ -591,7 +591,6 @@ lazy val actionsProj = (project in file("main-actions"))
     mimaSettings,
     mimaBinaryIssueFilters ++= Vector(
     ),
-    Compile / scalacOptions --= Seq("-Werror"),
   )
   .dependsOn(lmCore)
   .configure(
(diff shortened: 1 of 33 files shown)


stasimus:
@eed3si9n
I pulled your old experiment with forking 'compile', as it is very useful feature.

Known issues are:
- Performance. Every compile forks a fresh JVM and a cold scalac 
- incOptions/extraIncOptions and sourcePositionMappers aren't transported, worst for IDE users: the reporter lives in the worker, so BSP clients get no per-position diagnostics from a forked failure, just the failure itself.
- Classloader reconstruction in the worker (ScalaInstance from serialized jar lists, explicit parent partition) is inherently the fragile part - I think should be reviewed.

stasimus:
And for proper CI need a two JDKs installed...

eed3si9n:
Thank for working on this!

https://github.com/sbt/sbt/pull/8584 says

> When we upgrade to JDK 17, we can downgrade the compilation to use the real JDK 8, if Zinc itself it compiled with `--release 8`.

but I later realized this won't be possible unless Zinc is complied for JDK 8.

PR #9655, opened 2026-08-21
[2.x] fix: Fixes the ability to extend TestResultLogger

unkarjedy:
## Summary

Restores public visibility for `Tests.Output`, `SuiteResult`, and its companion object. This makes the documented public `TestResultLogger` extension point implementable from an ordinary external plugin again.

## Why this is needed

`TestResultLogger` remains a public, documented extension point: its documentation says output can be customized by providing a specialized instance through `testResultLogger`. Its required `run` method accepts `Tests.Output`.

In [PR #8181](https://github.com/sbt/sbt/pull/8181), commit [`c8737b8e4f6462e4be35bdfea6f6c597d7f4d38a`](https://github.com/sbt/sbt/commit/c8737b8e4f6462e4be35bdfea6f6c597d7f4d38a) changed the standard test tasks from `Unit` to the public `TestResult` enum. That commit also made `Tests.Output` and `SuiteResult` `private[sbt]`.

The commit message gives one reason: type test tasks by forwarding `TestResult`. It does not give a reason to restrict the report types or the result-logger customization API.

`TestResult` is not an equivalent public replacement for `Tests.Output`: it provides only the final `Empty`, `Passed`, `Failed`, or `Error` outcome. `Tests.Output` also provides the per-suite `events` and framework `summaries` required by a result logger. Since `Output.events` contains `SuiteResult`, restoring only `Output` would still leave the public report graph unusable; this PR restores both types.

## Real-world consumer

This was noticed during work on [TW-102637](https://youtrack.jetbrains.com/issue/TW-102637).

The [JetBrains TeamCity sbt logger](https://github.com/JetBrains/sbt-tc-logger) default [`master` branch](https://github.com/JetBrains/sbt-tc-logger/tree/master) contains a custom `TestResultLogger` for its test settings: [`SbtTeamCityLogger.scala#L92-L101`](https://github.com/JetBrains/sbt-tc-logger/blob/master/src/main/scala/jetbrains/buildServer/sbtlogger/SbtTeamCityLogger.scala#L92-L101). Its `run` implementation imports `sbt.Tests._` and accepts `results: Output`.

This is an actual consumer of the public extension point. During the SBT 2 compatibility work for TW-102637, the change in PR #8181 prevented equivalent external code from naming `Output`, because code outside `package sbt` no longer has source access to it. An external plugin should not need to adopt an `sbt.*` package merely to implement a public sbt extension point.

## Change

- Remove `private[sbt]` from `Tests.Output`, `SuiteResult`, and the `SuiteResult` companion.
- Add `tests/test-result-logger-api`, a scripted external build that:
  1. imports `sbt.Tests.Output`;
  2. implements `TestResultLogger` outside `package sbt`;
  3. consumes `Output.events` as `Iterable[SuiteResult]`; and
  4. runs `test` and confirms its logger ran.

## Verification

- Before the production change, the new scripted build fails while loading its `build.sbt`: `Output` cannot be found and `SuiteResult` cannot be used as a type outside `sbt`.
- After the change, `sbt "scripted tests/test-result-logger-api"` passes.
- `sbt scalafmtCheckAll` passes.
- `sbt mimaReportBinaryIssues` exits successfully; this snapshot build has no previous artifacts configured for MiMa analysis.

This restores the public source API available before PR #8181. It does not change test execution, report contents, or the `TestResult` returned by standard test tasks.

## AI disclosure

AI assistance was used to investigate the history, prepare the change, and run the listed local verification commands.

### DIFF
--- a/main-actions/src/main/scala/sbt/Tests.scala
+++ b/main-actions/src/main/scala/sbt/Tests.scala
@@ -77,7 +77,17 @@ object Tests {
       overall: TestResult,
       events: Map[String, SuiteResult],
       summaries: Iterable[Summary]
-  )
+  ) {
+
+    /**
+     * Returns a copy with the throwables removed from every suite result.
+     *
+     * Use this before retaining test results beyond the lifetime of the test task.
+     * See [[SuiteResult.throwables]] for why retaining those exceptions can keep the test class loader alive.
+     */
+    def withoutThrowables: Output =
+      copy(events = events.view.mapValues(_.withoutThrowables).toMap)
+  }
 
   /**
    * Summarizes a test run.
--- a/main-actions/src/main/scala/sbt/Tests.scala
+++ b/main-actions/src/main/scala/sbt/Tests.scala
@@ -73,7 +73,7 @@ object Tests {
    * @param events The result of each test group (suite) executed during this test run.
    * @param summaries Explicit summaries directly provided by test frameworks.  This may be empty, in which case a default summary will be generated.
    */
-  private[sbt] final case class Output(
+  final case class Output(
       overall: TestResult,
       events: Map[String, SuiteResult],
       summaries: Iterable[Summary]
--- a/sbt-app/src/sbt-test/tests/test-result-logger-api/test
+++ b/sbt-app/src/sbt-test/tests/test-result-logger-api/test
@@ -1,0 +1,2 @@
+> test
+$ exists test-result-logger-ran
--- a/sbt-app/src/sbt-test/tests/test-result-logger-api/test
+++ b/sbt-app/src/sbt-test/tests/test-result-logger-api/test
@@ -1,2 +1,2 @@
 > test
-$ exists test-result-logger-ran
+$ must-mirror expected-result test-result-logger-ran
--- a/testing/src/main/scala/sbt/TestReportListener.scala
+++ b/testing/src/main/scala/sbt/TestReportListener.scala
@@ -116,10 +116,12 @@ final class SuiteResult(
   def +(other: SuiteResult): SuiteResult = {
     val combinedTestResult =
       (result, other.result) match {
-        case (TestResult.Passed, TestResult.Passed) => TestResult.Passed: TestResult
-        case (_, TestResult.Error)                  => TestResult.Error: TestResult
-        case (TestResult.Error, _)                  => TestResult.Error: TestResult
-        case _                                      => TestResult.Failed: TestResult
+        case (TestResult.Empty, TestResult.Empty) => TestResult.Empty: TestResult
+        case (TestResult.Passed | TestResult.Empty, TestResult.Passed | TestResult.Empty) =>
+          TestResult.Passed: TestResult
+        case (_, TestResult.Error) => TestResult.Error: TestResult
+        case (TestResult.Error, _) => TestResult.Error: TestResult
+        case _                     => TestResult.Failed: TestResult
       }
     new SuiteResult(
       combinedTestResult,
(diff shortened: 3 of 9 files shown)


unkarjedy:
The test failure seems to be unrelated

testing/src/main/scala/sbt/TestReportListener.scala
```diff
@@ -59,7 +59,7 @@ trait TestsListener extends TestReportListener {
  *   exactly that before stashing a copy on `State.attributes`; see the note on `TestRecap.recapKey`.
  *   On Windows a leaked handle makes the underlying jar undeletable (e.g. by `clearCaches`).
  */
-private[sbt] final class SuiteResult(
+final class SuiteResult(
```
stasimus, line 63: Making SuiteResult public also exposes `throwables: Seq[Throwable]`, and the only way to drop them (`TestRecap.dropThrowables`) is still private[sbt], so plugins have no way to follow the retention warning in the scaladoc. Maybe add a public `withoutThrowables` and reword the doc, it currently points at private internals.

stasimus:
`testing/src/main/scala/sbt/TestReportListener.scala:100` `SuiteResult.+` has no case for `TestResult.Empty`, so combining with Empty falls into the `Failed` catch-all, while everything else in the codebase treats Empty as passing. Now that plugins can construct SuiteResult this is reachable, worth adding an Empty case.

main-actions/src/main/scala/sbt/Tests.scala
```diff
@@ -50,7 +50,7 @@ object Tests {
    * @param events The result of each test group (suite) executed during this test run.
    * @param summaries Explicit summaries directly provided by test frameworks.  This may be empty, in which case a default summary will be generated.
    */
-  private[sbt] final case class Output(
+  final case class Output(
```
stasimus, line 76: Output is a case class now, but SuiteResult keeps reference equality, so comparing two Outputs basically never succeeds and toString prints `sbt.SuiteResult@...`. Since it is being re-exposed anyway, maybe give SuiteResult value-based equals/hashCode/toString.

sbt-app/src/sbt-test/tests/test-result-logger-api/test
```diff
@@ -0,0 +1,2 @@
+> test
+$ exists test-result-logger-ran
```
stasimus: This only checks that the marker exists, the payload written into it is never asserted (and with no test sources it is always `...:Passed:0`). Either compare the content against a .check file like tests/summary does, or just touch the marker so the test does not pretend to check more than it does.

main-actions/src/main/scala/sbt/Tests.scala
unkarjedy: AFAICS it was a case class in sbt 1 as well (at least in 1.4, I haven't checked in earlier versions).
This indeed looks like an issue, but I don't have enough context to be confident in the fix.
I am not aware of the code in our plugin that would suffer from this issue.

I treat this change more like a revert of a wrong visibility restriction (see `sbt.TestResultLogger`).

testing/src/main/scala/sbt/TestReportListener.scala
unkarjedy: FTR: There was and still would be a way, just not the most convenient.
There is a constructor overload that accepts no `throwables` and users can use it.

And BTW even though it's not the most convenient way, it's the most reliable for those who want to cross compile their plugins (like we do). Though again this particular field is not used in our plugin

unkarjedy:
I addressed the comments except for this one:
https://github.com/sbt/sbt/pull/9655#discussion_r3830871915

I think it would be better if it was done in a separate PR as even though it is related, it's not much dependent on this.

eed3si9n:
I've retroactively created a bug report https://github.com/sbt/sbt/issues/9658

eed3si9n (approved):
Thanks!

PR #9657, opened 2026-08-21
[2.x] perf: Populate the in-memory cache on a read

### LINKED ISSUE #9656: [2.x] `InMemoryCacheStore` never populates on a read, so `update` re-parses its report every command
**Disclaimer**: Investigated with Claude and human review in the loop

## steps

sbt 2.1.0-SNAPSHOT (`develop` 08cc82ac8), JDK 25. Generated 81-module workspace (three chains of depth 27, `app` depending on the three tips), `exportJars=true`, output files real — not restored from the action cache, whose symlinks would swamp this.

```bash
../sbt "app/Compile/compile"        # build once
../sbt shutdown                     # -D reaches the server only at boot
../sbt -Dscale.exportJars=true -Dsbt.task.timings=true -Dsbt.task.timings.threshold=0 \
       "app/Compile/compile"        # run twice; measure the second — a NO-OP build
```

## problem

Nothing recompiles, yet `update` costs **398 ms across 82 projects**. It is not resolving — it is re-deserialising its own cached report, once per project, on every command, for the life of the server.

`InMemoryCacheStore.CacheStoreImpl.read` never caches what it reads:

```scala
case _ => cacheStore.read[T]()   // no store.put
```

Only `write` fills the cache, and `cachedUpdate` writes only when it re-resolves — so on a server that has not itself resolved, every read misses forever, and no number of builds warms it. `transitiveUpdate` pays that once per project in the dependency closure.

On a private 306-project build (~1 MB per report) a no-op build spent **1.7 s in `update` across 128 projects**, with **71 % of the samples under `cachedUpdate` in sjsonnew/jawn**:

```
CacheStoreImpl.read -> GzipFileBasedStore.read -> Parser.parseFromByteArray
```

## expectation

A read populates the cache it just missed, so a task whose stored output is up to date stops re-deserialising that output on every command.

hoangmaihuy:

Fixes #9656.

## Fixes

Put the value after a successful read. Keying is unchanged — `(path, lastModified)`, re-stat on every read, `write` still invalidates first — so a populated read is sound for the same reason a populated write is. A failed read throws before the `put`, and `put` already ignores a non-positive timestamp.

## Measurements

81-module workspace, **no-op** build, real output files, `exportJars=true`. Both sbt builds published from the same tree, differing only in this commit (bytecode-checked: `CacheStoreImpl` has one `store.put` before, two after). Alternated A-B-A-B, 8 runs per version, because this machine drifts by more than the effect size.

| | wall (median of 8) | `update` (threshold 0) | task total |
| --- | --- | --- | --- |
| before | 5.15 s | 398 ms / 82 projects | 4797 ms |
| after | **4.97 s** | **30 ms / 29 projects** | 4623 ms |

Nothing recompiled in any run; 53 of the 82 update tasks become too cheap to report. The 174 ms of task time recovered matches the ~180 ms of wall clock.

That workspace is generated, so its reports are small. On a private 306-project build with real dependency sets, `update` went **1763 ms -> 224 ms**.

## Tests

New `InMemoryCacheStoreTest` (hedgehog, 4 cases). The regression test rewrites the store file's bytes while restoring its mtime, so only an in-memory entry can answer with the original value — **it fails on `develop` and passes here**. Plus write-populates, moved-timestamp-re-reads, missing-file-absent.

## Note

This makes `maximumWeight` bind where it could not before — reads now admit entries, so a build whose working set exceeds `sbt.file.cache.size` evicts rather than never caching at all. Intended behaviour for a bounded cache, but a real change in what the budget means.

### DIFF
--- a/main/src/main/scala/sbt/internal/InMemoryCacheStore.scala
+++ b/main/src/main/scala/sbt/internal/InMemoryCacheStore.scala
@@ -52,11 +52,19 @@ private[sbt] object InMemoryCacheStore {
   private class CacheStoreImpl(path: Path, store: InMemoryCacheStore, cacheStore: CacheStore)
       extends CacheStore {
     override def delete(): Unit = cacheStore.delete()
+
+    /**
+     * A miss populates the cache. `write` is otherwise the only thing that fills it, and a task whose stored
+     * output is already up to date never writes -- so it would re-deserialise that output on every invocation.
+     */
     override def read[T]()(using reader: JsonReader[T]): T = {
       val lastModified = IO.getModifiedTimeOrZero(path.toFile)
       store.get[T](path) match {
         case Some((value, `lastModified`)) => value
-        case _                             => cacheStore.read[T]()
+        case _                             =>
+          val value = cacheStore.read[T]()
+          store.put(path, value, lastModified)
+          value
       }
     }
     override def write[T](value: T)(using writer: JsonWriter[T]): Unit = {
--- a/main/src/test/scala/sbt/internal/InMemoryCacheStoreTest.scala
+++ b/main/src/test/scala/sbt/internal/InMemoryCacheStoreTest.scala
@@ -1,0 +1,74 @@
+/*
+ * sbt
+ * Copyright 2023, Scala center
+ * Copyright 2011 - 2022, Lightbend, Inc.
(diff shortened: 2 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #9665, opened 2026-08-24
[2.x] feat: workerMaxInstances + testTopology

eed3si9n:
## Problem
A major bottleneck of the default setting is that concurrentRestriction allows exactly one forked test in flight at a time, and partly due to the restriction all test suites from a subproject runs within the forked process.

## Solution
This adds a new setting called workerMaxInstances, which lets the build users tune the number of concurrent worker process in flight. I've set the default to 2.
Within a single worker, the default parallelism is reduced to `2`.

The default test grouping now splits the test suites into testForkedWorker count.

## Example

```scala
Global / workerMaxInstances := 3 // or use -Dsbt.worker_max_instances=3 for CI

Test / fork := true

lazy val foo = project
  .settings(
    // this restores one JVM per subproject testing for this subproject
    Test / testTopology := TestTopology.subprojectExclusive,
  )
lazy val bar = project
```

### DIFF
--- a/main-actions/src/main/scala/sbt/ForkTests.scala
+++ b/main-actions/src/main/scala/sbt/ForkTests.scala
@@ -57,26 +57,25 @@ private[sbt] object ForkTests:
       this.getClass.getClassLoader // can't provide the loader for test classes, which is in another jvm
     def all(work: Seq[ClassLoader => Unit]) = work.fork(f => f(dummyLoader))
 
-    val main =
-      if opts.tests.isEmpty then
-        // Nothing selected after filtering, so don't run setup/cleanup either.
-        constant(TestOutput(TestResult.Passed, Map.empty[String, SuiteResult], Iterable.empty))
-      else
-        mainTestTask(
-          runners = runners,
-          opts = opts,
-          classpath = classpath,
-          converter = converter,
-          fork = fork,
-          log = log,
-          parallel = config.parallel,
-          parallelism = parallelism,
-          virtualClasspath = virtualClasspath,
-        ).tagw(config.tags*)
-          .dependsOn(all(opts.setup)*)
-          .flatMap: results =>
-            all(opts.cleanup).join.map(_ => results)
-    main.tagw(tags*)
+    if opts.tests.isEmpty then
+      // Nothing selected after filtering, so don't run setup/cleanup either.
+      constant(TestOutput(TestResult.Passed, Map.empty[String, SuiteResult], Iterable.empty))
+    else
+      mainTestTask(
+        runners = runners,
+        opts = opts,
+        classpath = classpath,
+        converter = converter,
+        fork = fork,
+        log = log,
+        parallel = config.parallel,
+        parallelism = parallelism,
+        virtualClasspath = virtualClasspath,
+      ).tagw(config.tags*)
+        .tagw(tags*)
+        .dependsOn(all(opts.setup)*)
+        .flatMap: results =>
+          all(opts.cleanup).join.map(_ => results)
   }
 
   private def mainTestTask(
--- a/main-actions/src/main/scala/sbt/ForkTests.scala
+++ b/main-actions/src/main/scala/sbt/ForkTests.scala
@@ -55,7 +55,11 @@ private[sbt] object ForkTests:
     import std.TaskExtra.*
     val dummyLoader =
       this.getClass.getClassLoader // can't provide the loader for test classes, which is in another jvm
-    def all(work: Seq[ClassLoader => Unit]) = work.fork(f => f(dummyLoader))
+    // Tag setup/cleanup the same as the actual test run -- otherwise a restriction like
+    // Tags.limit(Tags.ExclusiveTestGroup, 1) only ever sees the forked JVM in between them, and
+    // another subproject's setup/cleanup (or its own exclusive run) can freely overlap either end.
+    def all(work: Seq[ClassLoader => Unit]) =
+      work.fork(f => f(dummyLoader)).map(_.tagw(config.tags*).tagw(tags*))
 
     if opts.tests.isEmpty then
       // Nothing selected after filtering, so don't run setup/cleanup either.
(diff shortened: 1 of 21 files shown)


main/src/main/scala/sbt/internal/SysProp.scala
```diff
@@ -124,6 +124,8 @@ object SysProp:
   def testSummary: TestSummary =
     strOpt("sbt.test_summary").flatMap(Tests.parseTestSummary).getOrElse(TestSummary.default)
 
+  def testForkedWorker: Int = int("sbt.test_forked_worker", EvaluateTask.SystemProcessors / 3)
```
stasimus: EvaluateTask.SystemProcessors / 3 is 0 on <3 vCPU

On a machine with 1 or 2 CPUs, the build won't even start? I use sometimes Dev containers with limited resources,

main/src/main/scala/sbt/Keys.scala
```diff
@@ -403,7 +403,8 @@ object Keys {
   @transient
   val testListeners = taskKey[Seq[TestReportListener]]("Defines test listeners.").withRank(DTask)
   val testForkedParallel = settingKey[Boolean]("Whether forked tests should be executed in parallel").withRank(CTask)
-  val testForkedParallelism = settingKey[Option[Int]]("Maximum number of parallel test threads when using testForkedParallel. Defaults to the number of available processors.").withRank(CTask)
+  val testForkedParallelism = settingKey[Option[Int]]("Maximum number of parallel test threads when using testForkedParallel. Default: 2.").withRank(CTask)
+  val testForkedWorker = settingKey[Int]("Maximum number of test workers. Default available processors / 3.")
```
stasimus, line 407: Setting the value to 0 kills the build too

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1407,7 +1408,7 @@ object Defaults extends BuildCommon with DefExtra {
       )
     ) ++ inScope(GlobalScope)(
       Seq(
-        derive(testGrouping := Def.uncached(singleTestGroupDefault.value))
+        derive(testGrouping := Def.uncached(splitTestGroupDefault.value))
```
stasimus, line 1411: setup and cleanup hooks now run once per group instead of once?

stasimus: testGrouping used to produce a single group, so a Tests.Setup ran exactly once. Now it runs per group. If you start an embedded database in Tests.Setup with fork := true, on a 6-core machine you get several of them starting at the same time, all fighting over the same port or the same temp directory.

eed3si9n: A subproject that wants to run all the test suites in one worker can do:

```scala
testForkedWorker := 1
```

at the subproject level.

main/src/main/scala/sbt/internal/SysProp.scala
eed3si9n: Added `math.max(..., 1)` in https://github.com/sbt/sbt/pull/9665/changes/0dbedff68323cf5501141e29b527cf8c3a0aa833

main/src/main/scala/sbt/Defaults.scala
stasimus: okey, but against my intuition

main/src/main/scala/sbt/Defaults.scala
```diff
@@ -1446,6 +1447,35 @@ object Defaults extends BuildCommon with DefExtra {
...
+      val byName = tests.groupBy(_.name).toVector.sortBy(_._1)
+      val n = math.max(
+        1,
+        math.min(testForkedWorker.value, byName.size)
+      )
+      val buckets = Array.fill(n)(Vector.newBuilder[TestDefinition])
+      byName.zipWithIndex.foreach { case ((_, defs), i) => buckets(i % n) ++= defs }
+      buckets.toVector.zipWithIndex.collect {
+        case (b, i) if b.result().nonEmpty =>
+          new Tests.Group(s"<split-$i>", b.result(), Tests.SubProcess(opts), Seq.empty)
```
anatoliykmetyuk, line 1474: For `testQuick` and `testOnly`, filtering happens after grouping. Could this cause setup and teardown to run for groups containing no selected tests? Could we ensure they don't run for empty groups?

eed3si9n: Addressed in https://github.com/sbt/sbt/pull/9665/changes/636f6977aa9e940ddeca4b59e3303068b97b5d94

main-actions/src/main/scala/sbt/ForkTests.scala
```diff
@@ -73,9 +74,10 @@ private[sbt] object ForkTests:
           parallelism = parallelism,
           virtualClasspath = virtualClasspath,
         ).tagw(config.tags*)
-    main.tagw(tags*).dependsOn(all(opts.setup)*) flatMap { results =>
-      all(opts.cleanup).join.map(_ => results)
-    }
+          .dependsOn(all(opts.setup)*)
+          .flatMap: results =>
+            all(opts.cleanup).join.map(_ => results)
+    main.tagw(tags*)
```
anatoliykmetyuk: Moving `main.tagw` here means forked tests are no longer constrained by forked-test-group and custom group tags. That's why the CI also fails now.

main/src/main/scala/sbt/Keys.scala
anatoliykmetyuk: Is this still something we want to address in this PR, or is 0 intentionally considered invalid configuration? If it’s invalid, should we validate it and provide a clearer error?

eed3si9n:
Stepped back a bit, and wrote an RFC: https://eed3si9n.com/rfc-5-test-scheduling/.

main/src/main/scala/sbt/Defaults.scala
eed3si9n: Now called

```scala
Test / testTopology := TestTopology.subprojectExclusive,
```

main/src/main/scala/sbt/Keys.scala
eed3si9n: Added verification here - https://github.com/sbt/sbt/pull/9665/changes/8d003364cc6e67bc26a6681cde8d5c0ce3deee6a

anatoliykmetyuk (approved):
LGTM!

PR #9667, opened 2026-08-24
[2.0.x] fix: Fixes TestReportListener not receiving endGroup after a LinkageError

### LINKED ISSUE #9666: [1.x, 2.x] TestReportListener does not receive endGroup after an uncaught LinkageError
_(Disclaimer: the issue was investigated, generated by Codex. Reviewed by me)_

## Summary

When a test framework throws a `LinkageError` while constructing or running a suite, SBT
can call `TestReportListener.startGroup` but never send a terminal `endGroup` callback.
The outer SBT command still fails, which is correct; the problem is the incomplete listener
lifecycle.

This has been reproduced with a practical `ExceptionInInitializerError` scenario. A suite
initialises a required `ServiceLoader` provider while it is being constructed, but no
provider is present on the test runtime classpath. `ServiceLoader.iterator().next()` throws
`NoSuchElementException` during static/object initialisation, and the JVM wraps it in
`ExceptionInInitializerError`. The test does not deliberately throw that error.

The attached `sbt2-test-report-listener-linkage-error-mre.zip` is self-contained and has no
TeamCity dependency: it uses a small `TestReportListener` that prints callback traces.

## Verified versions

The relevant callback control flow is equivalent in SBT `1.4.9`, `1.12.15`, and `2.0.6`.
The attached project executes against SBT `2.0.6`.

- [TestFramework.scala (2.0.6)](https://github.com/sbt/sbt/blob/v2.0.6/testing/src/main/scala/sbt/TestFramework.scala#L153-L176)
- [TestReportListener (2.0.6)](https://github.com/sbt/sbt/blob/v2.0.6/testing/src/main/scala/sbt/TestReportListener.scala#L14-L26)

`TestReportListener.endGroup(name, Throwable)` is documented as “called if there was an
error during test.”

## Reproduction
[sbt2-test-report-listener-linkage-error-mre.zip](https://github.com/user-attachments/files/31370982/sbt2-test-report-listener-linkage-error-mre.zip)
```bash
unzip sbt2-test-report-listener-linkage-error-mre.zip
cd sbt2-test-report-listener-linkage-error-mre
sbt 'testOnly example.DirectMissingConfigurationSuite'
```

The positive control has an ordinary suite-construction error (a missing required
configuration file). It fails the command and correctly completes the listener group:

```text
TRACE startGroup: example.DirectMissingConfigurationSuite
TRACE testEvent: details=1, result=Some(Error)
TRACE endGroup(result): example.DirectMissingConfigurationSuite -> Error
```

Then run the failing case in a fresh SBT process:

```bash
sbt 'testOnly example.MissingServiceProviderSuite'
```

## Actual result

The JVM raises `ExceptionInInitializerError` and SBT returns a non-zero result, but the
listener sees only the started group:

```text
TRACE startGroup: example.MissingServiceProviderSuite
[info] example.MissingServiceProviderSuite *** ABORTED ***
[info]   java.lang.ExceptionInInitializerError
[info]   Cause: java.util.NoSuchElementException
[error] (Test / testSelected) java.lang.ExceptionInInitializerError
```

There is no `TRACE testEvent` and, critically, no terminal callback:

```text
TRACE endGroup(throwable): example.MissingServiceProviderSuite -> java.lang.ExceptionInInitializerError
```

## Expected result

SBT should preserve the current propagated exception and non-zero command result, while
first telling listeners that the already-started group terminated with an error:

```text
TRACE startGroup: example.MissingServiceProviderSuite
TRACE endGroup(throwable): example.MissingServiceProviderSuite -> java.lang.ExceptionInInitializerError
... java.lang.ExceptionInInitializerError ...
```

## Scope and possible narrow fix

This is not a request to convert every `LinkageError` into an ordinary test event or to
change test/exit-code semantics. `NoClassDefFoundError` and `IllegalAccessError` already
have special inner handling. This concerns remaining escaping `LinkageError`s, for which
the outer lifecycle catch only handles `NonFatal`.

A narrow solution could preserve current propagation while closing the listener group:

```scala
catch {
  case NonFatal(e) =>
    safeListenersCall(_.endGroup(name, e))
    (SuiteResult.Error, Seq.empty[TestTask])

  case e: LinkageError =>
    safeListenersCall(_.endGroup(name, e))
    throw e
}
```

`ExceptionInInitializerError` is a compact practical reproducer, not necessarily the only
affected error. For example, a runtime/compile-time dependency mismatch can produce
`NoSuchMethodError` through the same lifecycle path, but it needs a multi-artifact fixture
and is not required to reproduce the callback gap.

unkarjedy:
Fixes #9666

## Summary

- Notify `TestReportListener` of a terminal error when a suite throws an otherwise escaping `LinkageError`.
- Rethrow the original error, preserving test failure and exit-code semantics.
- Keep the focused `TestRunnerSpec` regression for callback type, no normal result callback, and error propagation.
- Add a scripted regression that registers a real listener through `testListeners` and covers both an initializer-produced and an explicitly user-thrown `ExceptionInInitializerError`.

## Reproduction

Without the new `LinkageError` branch, the scripted regression fails because neither case receives `endGroup(name, throwable)`. This was validated independently for both the implicit-initialization and explicit-throw cases, then the production branch was restored.

## Verification

- `sbt --client "scripted tests/test-report-listener-linkage-error"`
  - Passed with the final two-example fixture.
- `sbt --client ";project testingProj; testOnly sbt.*"`
  - Passed: Total 4, Failed 0, Errors 0, Passed 4.
- `git diff --check`
  - Passed.

## AI assistance

Codex (GPT-5) assisted with investigation, implementation, tests, commit, and this draft PR.

### DIFF
--- a/sbt-app/src/sbt-test/tests/test-report-listener-linkage-error/build.sbt
+++ b/sbt-app/src/sbt-test/tests/test-report-listener-linkage-error/build.sbt
@@ -1,0 +1,59 @@
+import java.io.File
+
+import sbt.io.IO
+
+val implicitSuite = "example.ImplicitInitializationFailure"
+val explicitSuite = "example.ExplicitInitializationFailure"
+
+ThisBuild / scalaVersion := "2.12.21"
+
+libraryDependencies += "org.scalatest" %% "scalatest" % "3.2.20" % Test
+
+/** Returns the callback log file for one suite. */
+def callbackFile(base: File, suite: String): File = base / s"$suite.log"
+
+/** Renders a throwable callback as the stable form asserted by this fixture. */
+def throwableCallbackLine(name: String, throwable: Throwable): String = {
+  val causeClass = Option(throwable.getCause).fold("<none>")(_.getClass.getName)
+  s"throwable|$name|${throwable.getClass.getName}|$causeClass"
+}
+
+/** Verifies that a suite produced exactly one expected terminal callback. */
+def checkCallback(base: File, suite: String, expected: String): Unit = {
+  val file = callbackFile(base, suite)
+  val actual = if (file.exists) IO.readLines(file) else Nil
+  if (actual != List(expected)) {
+    sys.error(s"Expected only callback '$expected', but observed: ${actual.mkString(", ")}")
+  }
+}
+
+// Use one listener so the fixture records only the terminal lifecycle callback under test.
+Test / testListeners := {
+  val base = baseDirectory.value
+  Seq(new TestReportListener {
+    def startGroup(name: String): Unit = ()
+    def testEvent(event: TestEvent): Unit = ()
+    def endGroup(name: String, throwable: Throwable): Unit = {
+      IO.append(callbackFile(base, name), throwableCallbackLine(name, throwable) + "\n")
+    }
+    def endGroup(name: String, result: TestResult): Unit = {
+      IO.append(callbackFile(base, name), s"result|$name|$result\n")
+    }
+  })
+}
+
+val checkImplicitCallback = taskKey[Unit]("Check the implicit initializer error callback")
+val checkExplicitCallback = taskKey[Unit]("Check the explicit initializer error callback")
+
+checkImplicitCallback :=
+  checkCallback(
+    baseDirectory.value,
+    implicitSuite,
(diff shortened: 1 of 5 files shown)


unkarjedy:
If all is good with this PR I will back port it to 1.x

eed3si9n (approved):
Thanks!

PR #9669, opened 2026-08-24
[2.x] fix: Fixes --addPluginSbtFile getting lost after reboot

### LINKED ISSUE #4303: reboot loses addPluginSbtFile settings

## steps

To be able to use the `addPluginSbtFile` option, I start the user's sbt shell in IntelliJ with the `-Dsbt.version=1.2.0` parameter.

This causes the warning on startup:
```
[warn] sbt version mismatch, current: 1.2.0, in build.properties: "1.1.6", use 'reboot' to use the new value.
```

## problem

Actually doing a `reboot` from the shell will then apparently lose the the injected settings, causing the reboot to fail because of the missing command added by the injected plugin. The warning still appears, however:

```
[IJ]sbt:untitled5> reboot
[warn] sbt version mismatch, current: 1.2.0, in build.properties: "1.1.6", use 'reboot' to use the new value.
[info] Loading settings for project global-plugins from idea.sbt,plugins.sbt ...
[info] Loading global plugins from /Users/jast/.sbt/1.0/plugins
[info] Updating ProjectRef(uri("file:/Users/jast/.sbt/1.0/plugins/"), "global-plugins")...
[info] Done updating.
[info] Loading project definition from /Users/jast/playspace/untitled5/project
[info] Updating ProjectRef(uri("file:/Users/jast/playspace/untitled5/project/"), "untitled5-build")...
[info] Done updating.
[info] Loading settings for project untitled5 from build.sbt ...
[info] Set current project to untitled5 (in build file:/Users/jast/playspace/untitled5/)
[error] Not a valid command: idea-shell (similar: shell, oldshell)
[error] Not a valid project ID: idea-shell
[error] Expected ':'
[error] Not a valid key: idea-shell
[error] idea-shell
[error]           ^

Process finished with exit code 1
```

## expectation

* Ideally, reboot should work with the added plugins intact.
* Independently, an option to disable the warning would be good to avoid misleading and annoying users.

## notes

sbt version: 1.2.0

Follow-up to #4211

azdrojowa123:
Fixes https://github.com/sbt/sbt/issues/4303

### DIFF
--- a/main-command/src/main/scala/sbt/State.scala
+++ b/main-command/src/main/scala/sbt/State.scala
@@ -248,11 +248,39 @@ object State {
     val app = state.configuration.provider
     new Reboot(
       app.scalaProvider.version,
-      state.remainingCommands map { case e: Exec => e.commandLine },
+      addPluginSbtFileArguments(state) ::: state.remainingCommands.map(_.commandLine),
       app.id,
       state.configuration.baseDirectory
     )
   }
+
+  /**
+   * Builds the `early(...)` commands that add the extra plugin sbt files back after a reboot.
+   * A reboot clears the state, so sbt forgets these files. The commands are passed to the new
+   * sbt as arguments, and they run before it loads the build.
+   *
+   * The `early(addPluginSbtFile=...)` form is used, not `--addPluginSbtFile=...`, because the
+   * `--` form loses the quotes around the path, and then a path with a space in it fails.
+   * A path with a space can be added using `addPluginSbtFile "<path>"`.
+   */
+  private[sbt] def addPluginSbtFileArguments(state: State): List[String] =
+    state.get(BasicKeys.extraMetaSbtFiles).toList.flatten.distinct.map { vf =>
+      val path = vf match {
+        case f: xsbti.PathBasedFile => f.toPath.toString
+        case f                      => f.id
+      }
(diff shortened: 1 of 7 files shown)


azdrojowa123:
I think the test is failing because I added an assertion that the `BuildInfo` plugin is not available when the test starts, but it looks like it persists from another scripted test. I'll think about how to overcome this.

eed3si9n:
scripted runs in a batch mode for the entire group, or whatever invoked in GitHub Actions, so reloading semantics is a bit messy. Feel free to create your own scripted group like

```yaml
      - name: Build and test (4)
        if: ${{ matrix.jobtype == 4 }}
        shell: bash
        run: |
          ./sbt -v "RepoOverrideTest/scripted force-proxy-repos-file/*"
          ./sbt -v "scripted source-dependencies/*"
```

azdrojowa123:
I tried cleaning up the global directory in the `global-plugin` test, and it looks like it works. WDYT? 
I'm not against creating a separate scripted group. I just thought this was a simpler approach.

eed3si9n:
> I tried cleaning up the global directory in the `global-plugin` test, and it looks like it works. WDYT? 

Nice. If the tests run, I'm fine with either way.

eed3si9n (approved):
Thanks!

PR #9670, opened 2026-08-24
[1.x] fix: Fixes TestReportListener not receiving endGroup after a LinkageError

unkarjedy:
Backport of #9667 for the issue described in #9666.

1.12-specific changes: the direct regression uses Scala 2 syntax and adds the existing verify framework in Test scope because the testing module does not otherwise have a unit-test dependency.


unkarjedy:
The test failure on Windows seems unrelated

eed3si9n (approved):
Thanks!

PR #9671, opened 2026-08-24
[2.x] fix: Fixes JDK 17 domain socket deadlocks

eed3si9n:
## Problem
`Channels.newInputStream/newOutputStream` share a channel-wide lock, which would deadlock for a duplex communication.

## Solution
This implements an alternative DuplexChannels functions that's capable of duplex communication expected of a "socket".

This also duplicates the Java impl to the worker app, so we can use JDK domain socket for forked test communication.

## AI Disclosure
I've used Claude Code to investigate, test, and implement the fix. I've micromanaged it to generate property-based test etc, and I've reviewed the generated code. The fix itself is fairly straightforward.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -476,6 +476,11 @@ lazy val workerProj = (project in file("worker"))
     libraryDependencies += "org.scala-lang" %% "scala3-library" % scalaVersion.value % Test,
     // run / fork := false,
     Test / fork := true,
+    // WorkerMain.java uses JDK 16+ Unix domain socket APIs (StandardProtocolFamily.UNIX,
+    // UnixDomainSocketAddress, SocketChannel); override the build-wide -source/-target 1.8.
+    compile / javacOptions ~= { opts =>
+      opts.filterNot(v => v == "-source" || v == "-target" || v == "1.8") ++ Seq("--release", "17")
+    },
     mimaSettings,
     mimaBinaryIssueFilters ++= Vector(
     ),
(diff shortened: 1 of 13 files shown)


main-actions/src/main/scala/sbt/internal/WorkerExchange.scala
```diff
@@ -42,28 +45,49 @@ object WorkerExchange:
...
+        (Seq("--tcp", serverSocket.getLocalPort().toString()), Some(serverSocket))
+      case WorkerConnection.Ipc(path) =>
+        val serverChannel = Retry {
+          Files.deleteIfExists(path)
+          val ch = ServerSocketChannel.open(StandardProtocolFamily.UNIX)
+          ch.bind(UnixDomainSocketAddress.of(path))
+          ch
+        }
+        val accepter = Thread(() => {
+          val channel = serverChannel.accept()
```
stasimus, line 110: I don't see it is closed and every forked test run leaks

eed3si9n: Addressed in https://github.com/sbt/sbt/pull/9671/commits/fd5e57a9183b78ea1b8e4d55995b9f2043db2641

build.sbt
```diff
@@ -476,6 +476,11 @@ lazy val workerProj = (project in file("worker"))
     libraryDependencies += "org.scala-lang" %% "scala3-library" % scalaVersion.value % Test,
     // run / fork := false,
     Test / fork := true,
+    // WorkerMain.java uses JDK 16+ Unix domain socket APIs (StandardProtocolFamily.UNIX,
+    // UnixDomainSocketAddress, SocketChannel); override the build-wide -source/-target 1.8.
+    compile / javacOptions ~= { opts =>
+      opts.filterNot(v => v == "-source" || v == "-target" || v == "1.8") ++ Seq("--release", "17")
```
anatoliykmetyuk: As we move toward JDK Unix-domain sockets, will all sbt-managed worker processes effectively require JDK 17+? Could this affect libraries that run forked tests on older JDKs to validate compatibility? If so, should we document this migration constraint somewhere?

anatoliykmetyuk (approved):

build.sbt
eed3si9n: To keep the door open for JDK 8 testing, I've turned worker into a multi-release JAR in d9a39f12969267ef67eed42525a0e1c6dc154cd7. We could fallback to stdio for communication on JDK 8.

anatoliykmetyuk: To clarify, this is a foundation for such an option in the future, but the fallback plumbing is out of scope of this PR? Forked tests on JDK 8/11 would still fail with an exception since TCP or stdio is not explicitly selected yet anywhere?

eed3si9n: Added fallback in https://github.com/sbt/sbt/pull/9671/commits/679297ac383b98738dc5d53a7f4ffb19e2477362.

PR #9674, opened 2026-08-24
[2.x] fix: Fixes ThisBuild-scoped bare settings

### LINKED ISSUE #9668: [2.x] `+=` append on `ThisBuild` is executed multiple times
### `build.sbt`
```scala
lazy val mySetting = settingKey[Seq[String]]("Example setting")

// Declare some projects
lazy val a = project.in(file("a"))
lazy val b = project.in(file("b"))

Global / mySetting := Seq("initial-value")

// This adds added-value twice
ThisBuild / mySetting += "added-value"
```

### With SBT 2 
```scala
// build.properties
sbt.version=2.0.7
```

```sh
λ> sbt mySetting
...
[info] a / mySetting
[info] 	* initial-value
[info] 	* added-value
[info] 	* added-value
[info] 	* added-value
[info] b / mySetting
[info] 	* initial-value
[info] 	* added-value
[info] 	* added-value
[info] 	* added-value
[info] mySetting
[info] 	* initial-value
[info] 	* added-value
[info] 	* added-value
[info] 	* added-value
```

### With SBT 1

```scala
// build.properties
sbt.version=1.13.0
```

```sh
λ> sbt mySetting
...
[info] a / mySetting
[info] 	List(initial-value, added-value)
[info] b / mySetting
[info] 	List(initial-value, added-value)
[info] mySetting
[info] 	List(initial-value, added-value)
```

## problem

`+=` append on `ThisBuild` is executed multiple times. It appears to be executed per project (once per `a` and `b`), plus an additional time.

## expectation
 
This should have the same behaviour as in SBT 1. `mySetting` should have the value `List(initial-value, added-value)`.

## notes
Declaring `lazy val c = project.in(file("c"))` causes another `added-value` to be appended to the setting.

The following code using assignment `:=` behaves the same as in SBT 1:
```scala
ThisBuild / mySetting := { mySetting.value :+ "added-value" }
```

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/9668

## Problem
ThisBuild scoped bare settings are treated as a common setting, which results in duplicate appends etc.

## Solution
Only treat This-project scoped setting as a common setting.

### DIFF
--- a/main/src/main/scala/sbt/internal/Load.scala
+++ b/main/src/main/scala/sbt/internal/Load.scala
@@ -15,7 +15,7 @@ import sbt.Keys.*
 import sbt.Project.inScope
 import sbt.ProjectExtra.{ prefixConfigs, setProject, showLoadingKey, structure }
 import sbt.Scope.GlobalScope
-import sbt.ScopeAxis.{ Select, Zero }
+import sbt.ScopeAxis.{ Select, This, Zero }
 import sbt.SlashSyntax0.*
 import sbt.internal.BuildStreams.*
 import sbt.internal.inc.classpath.ClasspathUtil
@@ -1069,6 +1069,10 @@ private[sbt] object Load {
         log = log,
       )
 
+    // Excludes settings already scoped explicitly (ThisBuild, Global); those apply once (#9668).
+    def thisProjectScoped(settings: Seq[Setting[?]]): Seq[Setting[?]] =
+      settings.filter(_.key.scope.project == This)
+
     // load all relevant configuration files (.sbt, as .scala already exists at this point)
     def discover(base: File): DiscoveredProjects = {
       val auto =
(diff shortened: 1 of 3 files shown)


main/src/main/scala/sbt/internal/Load.scala
```diff
@@ -1069,6 +1069,10 @@ private[sbt] object Load {
         log = log,
       )
 
+    // Excludes settings already scoped explicitly (ThisBuild, Global); those apply once (#9668).
+    def thisProjectScoped(settings: Seq[Setting[?]]): Seq[Setting[?]] =
+      settings.filter(_.key.scope.project == This)
```
stasimus: `_.key.scope.project == This` also drops ThisProject-scoped settings (Select(ThisProject)).
Seems previously propagated to all project - now only root

eed3si9n: Nice catch. Addressed in 211b1d04f8e1abce5c1311b92a608f37b8caf12f

anatoliykmetyuk (approved):
LGTM!

PR #9602, opened 2026-08-15
[2.x] feat: Test summary

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/9588
See also https://github.com/sbt/sbt/issues/2998

## Problem/Solution
This extends the idea started with TestRecap, and applies it to both test success and failures.

1. Existing TestResultLogger trait is extended to handle the summary rendering.
2. TestSummary enum is added to control the verbosity via command-line option, system property, or a setting.
3. Script test captures the log.

## Test

Success case would now print the total number of test suites, including the cached ones:

```bash
[info] passed: total 4, failed 0, errors 0, passed 4, cached 4
```

Failed case will show the failed test suites:

test_summary

`test --test_summary=success` would display both failed and succeeded test suites:

test_summary2

PR #9675, opened 2026-08-25
[2.x] Add test result cache opt-out

### LINKED ISSUE #9587: Opt out of test result cache
## Expectation

I want to be able to opt out of test result cache.

We can provide `--cache_test_result=yes|no`, which translates to a system property `sbt.cache_test_result`, and environment variable `SBT_CACHE_TEST_RESULT`. When they are set to `no` or `false`, we can opt out of test result caching. This would make `test` act like like `testOnly`.

anatoliykmetyuk:
Fixes #9587

Adds an opt-out for test result caching.

Launch sbt with `-Dsbt.cache_test_result=false` to disable test-result caching for the entire sbt session, or run `test --cache_test_result=false` to disable it for a single test invocation. The per-command option takes precedence over the session-wide setting.

### DIFF
--- a/main/src/main/scala/sbt/internal/SysProp.scala
+++ b/main/src/main/scala/sbt/internal/SysProp.scala
@@ -125,13 +125,7 @@ object SysProp:
     strOpt("sbt.test_summary").flatMap(Tests.parseTestSummary).getOrElse(TestSummary.default)
 
   def cacheTestResult: Boolean =
-    strOpt("sbt.cache_test_result")
-      .flatMap: value =>
-        value.toLowerCase(Locale.ENGLISH) match
-          case "yes" => Some(true)
-          case "no"  => Some(false)
-          case _     => parseBoolean(value)
-      .getOrElse(true)
+    getOrTrue("sbt.cache_test_result")
 
   /**
    * Indicates whether formatting has been disabled in environment variables.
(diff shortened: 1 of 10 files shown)


launcher-package/src/universal/bin/sbt.bat
```diff
@@ -708,6 +728,10 @@ if defined sbt_args_supershell (
   set _SBT_OPTS=-Dsbt.supershell=!sbt_args_supershell! !_SBT_OPTS!
 )
 
+if defined sbt_args_cache_test_result (
+  set _SBT_OPTS=-Dsbt.cache_test_result=!sbt_args_cache_test_result! !_SBT_OPTS!
+)
```
eed3si9n: Since system property would apply only once when the server starts up, consider using input task for `testQuick`-specific flags. That's what I did for test summary `--test_summary=success` flag in https://github.com/sbt/sbt/pull/9602.

anatoliykmetyuk: Refactored it to use similar parsing mechanism as with `--test_summary`.

main/src/main/scala/sbt/internal/SysProp.scala
```diff
@@ -124,6 +124,15 @@ object SysProp:
...
     strOpt("sbt.test_summary").flatMap(Tests.parseTestSummary).getOrElse(TestSummary.default)
 
+  def cacheTestResult: Boolean =
+    strOpt("sbt.cache_test_result")
+      .flatMap: value =>
+        value.toLowerCase(Locale.ENGLISH) match
+          case "yes" => Some(true)
+          case "no"  => Some(false)
+          case _     => parseBoolean(value)
+      .getOrElse(true)
```
eed3si9n: There's a Boolean parsing helper function.

```suggestion
    getOrTrue("sbt.cache_test_result")
```

eed3si9n (approved):
Minor nit on SysProp, but otherwise lgtm. Thanks!

